PR #5387, opened 2022-11-08
test(per-partition-limit): added new test

soyacz:
It is now possible to limit read rates and writes rates into a partition with a new WITH per_partition_rate_limit clause for the CREATE TABLE and ALTER TABLE statements. This is useful to prevent hot-partition problems when high rate reads or writes are bogus (for example, arriving from spam bots)
Feature introduced in:
https://github.com/scylladb/scylladb/commit/dab56b82fae5e36f7aa2ca6700d8cfa5baa2b515

Each cql request goes through per-partition limiter creating a risk of latency degradation.

Test scenario is about measuring basic performance metrics (latency, op/s) with and without per-partition-limit set and verifying the impact. Also measure influence of hot-partition scenario on these metrics.

### DIFF
--- a/test-cases/features/per-partition-limit.yaml
+++ b/test-cases/features/per-partition-limit.yaml
@@ -1,0 +1,23 @@
+test_duration: 180
+# using scylla-bench instead cassandra-stress as c-s rapidly retries on failure (e.g. due per-partition limit error)
+# causing overloading cluster due flood of failing requests
+
+stress_cmd_w: "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=1000000 -clustering-row-count=1 -concurrency=50 -connection-count=10 -rows-per-request=1 -consistency-level=quorum -iterations=0 -timeout=300s -duration=10m"
+stress_cmd_r: "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1000000 -clustering-row-count=1 -concurrency=50 -connection-count=10 -rows-per-request=1 -consistency-level=quorum -iterations=0 -timeout=300s -duration=10m"
+
+# stress that hits only several partitions to hit per-partition limit, ignoring errors as we expect them
+stress_cmd: [
+  "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=20 -clustering-row-count=1 -concurrency=100 -connection-count=100 -rows-per-request=10 -consistency-level=quorum -iterations=0 -timeout=300s -duration=10m -partition-offset=1000001 -error-at-row-limit 0",
+  "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=20 -clustering-row-count=1 -concurrency=100 -connection-count=100 -rows-per-request=10 -consistency-level=quorum -iterations=0 -timeout=300s -duration=10m -partition-offset=1000001 -error-at-row-limit 0"
+]
+
+
+n_db_nodes: 3
+n_loaders: 1
+n_monitor_nodes: 1
+
+instance_type_db: 'i3.large'
+instance_type_loader: 'c5.large'
+instance_type_monitor: 't2.small'
+
+user_prefix: 'per-part-limit'
--- a/jenkins-pipelines/features-per-partition-limit.jenkinsfile
+++ b/jenkins-pipelines/features-per-partition-limit.jenkinsfile
@@ -1,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    backend: 'aws',
+    region: 'eu-west-1',
+    test_name: 'per_partition_limit_test.PerPartitionLimitTest.test_per_partition_limit',
+    test_config: 'test-cases/features/per-partition-limit.yaml',
+
+    timeout: [time: 180, unit: 'MINUTES']
+)
--- a/per_partition_limit_test.py
+++ b/per_partition_limit_test.py
@@ -1,0 +1,119 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2022 ScyllaDB
+import logging
+from dataclasses import dataclass
+
+from sdcm.stress_thread import CassandraStressThread
+from sdcm.tester import ClusterTester
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class PerfResult:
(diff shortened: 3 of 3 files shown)


soyacz:
Please review size of the cluster and load - I'm not sure if I set it correctly (maybe we should measure with higher load?)
Staging test job: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/per-partition-limit-test/3/

soyacz:
Some cluster utilisation graphs for test run:
image
we can see 10m blocks on it:
1. base write
2. base read
3. base write + 20 partitions heavy load
4. base read + 20 partitions heavy load
5. limited write
6. limited read
7. limited write + 20 partitions heavy load
8. limited read + 20 partitions heavy load

Results (in logs) are showing no latency degradation and improvement of latency and op/s when additional heavy load for limited partition set is ongoing.

soyacz:
Need to fix email report as it failing sending results.

fruch:
> Some cluster utilisation graphs for test run: image we can see 10m blocks on it:
> 
> 1. base write
> 2. base read
> 3. base write + 20 partitions heavy load
> 4. base read + 20 partitions heavy load
> 5. limited write
> 6. limited read
> 7. limited write + 20 partitions heavy load
> 8. limited read + 20 partitions heavy load
> 
> Results (in logs) are showing no latency degradation and improvement of latency and op/s when additional heavy load for limited partition set is ongoing.

I'm not sure this causing a hot partition situation, change the monitor view to be per instance, and then par shard.
we should expect to see in the case of hot partition, one node, or one shard that gets most of the traffic.

i.e. the 20 partitions traffic should be the majority of the traffic. I would start doing just those on their own and analyze/measure it first.

looking at the issue that initiate this: https://github.com/scylladb/scylladb/issues/4703
you should be looking at the P99 latency, and not the throughput 

also I would separate it to multiple keyspaces, what you call base load should be on a different keyspace, and not limited


@piodul can you share the scylla-bench command you were using for testing ?

piodul:
> @piodul can you share the scylla-bench command you were using for testing ?

Unfortunately I can't find them right now (I'm not sure I still have them, but I'll try to look later), but the things I for sure did differently were:

- I used larger concurrency over the scylla-bench invocation that was meant to cause overload (something like 1k or 2k).
- I overloaded only one partition (so that only one shard gets higher load).

Also, make sure that scylla-bench you use is shard-aware because rate limiting works best in that scenario.

soyacz:
latency graphs:
image
we can see that latency during hot-partitions load without feature enabled is much higher than when we limit rate and is quite steady.

test-cases/features/per-partition-limit.yaml
```diff
@@ -0,0 +1,23 @@
+test_duration: 180
+# using scylla-bench instead cassandra-stress as c-s rapidly retries on failure (e.g. due per-partition limit error)
+# causing overloading cluster due flood of failing requests
+
+stress_cmd_w: "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=1000000 -clustering-row-count=1 -concurrency=50 -connection-count=10 -rows-per-request=1 -consistency-level=quorum -iterations=0 -timeout=300s -duration=10m"
+stress_cmd_r: "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1000000 -clustering-row-count=1 -concurrency=50 -connection-count=10 -rows-per-request=1 -consistency-level=quorum -iterations=0 -timeout=300s -duration=10m"
```
piodul, line 6: AFAIK scylla-bench, by default, ignores all errors and doesn't limit the number of errors received. We probably want to stop these after receiving an error, right? In that case, maybe we should set `-error-at-row-limit 1` here, or the new `-error-limit 1`.

piodul:
> image In the test normal load latency and throughput is compared and how hot partition load affects it. I take results from s-b as they are more accurate than on the graphs.

Did that test run had 1 or 20 hot partitions? (With 20, I would expect the load to be more balanced)

> 
> > i.e. the 20 partitions traffic should be the majority of the traffic. I would start doing just those on their own and analyze/measure it first.
> 
> concurrency and connection count for hot partitions traffic is higher than for normal load in this test - I assume this make it to be majority. We can see it on graphs I think.

Having larger concurrency than the normal load IMO makes sense, and that is what I did in my tests.

> > looking at the issue that initiate this: [scylladb/scylladb#4703](https://github.com/scylladb/scylladb/issues/4703) you should be looking at the P99 latency, and not the throughput
> 
> I added throughput verification to test also - this feature should also bring benefit in throughput under hot partition load. My mistake I didn't show graphs for it (fixed in comment above).

I agree that we should verify both latency and throughput. We have fixed concurrency here, so according to Little's law both throughput and latency (_average_ latency, at least, but ofc we are interested in p99 more) should become better.

piodul (dismissed):
The test LGTM, only left some small nitpicks.

I initially misread the first message about the results and thought that per partition limiting didn't bring any improvement, so I suggested some changes. However, if the test shows improvements then I think the current parameters are perfectly fine.

I'm not sure whether having a single overloaded partition or multiple of them (so that all shards are loaded) is better, but I think both should be fine and should show that per-partition rate limiting increases the useful throughput of overloaded shards.

Increasing the load might not make sense, because you are using much weaker machines than I did.

To sum up, I think that the current parameters are fine.

test-cases/features/per-partition-limit.yaml
fruch: SCT is doing -error-at-row-limit 1000
If nothing else is defined.

But I agreed setting to fail on any error is reasonable here

fruch:
> 
> > I'm not sure this causing a hot partition situation, change the monitor view to be per instance, and then par shard. we should expect to see in the case of hot partition, one node, or one shard that gets most of the traffic.
> 
> image
> In the test normal load latency and throughput is compared and how hot partition load affects it. I take results from s-b as they are more accurate than on the graphs.
> 
> > 
> > i.e. the 20 partitions traffic should be the majority of the traffic. I would start doing just those on their own and analyze/measure it first.
> 
> concurrency and connection count for hot partitions traffic is higher than for normal load in this test - I assume this make it to be majority. We can see it on graphs I think.
> 
> > 
> > looking at the issue that initiate this: [scylladb/scylladb#4703](https://github.com/scylladb/scylladb/issues/4703) you should be looking at the P99 latency, and not the throughput
> 
> I added throughput verification to test also - this feature should also bring benefit in throughput under hot partition load. My mistake I didn't show graphs for it (fixed in comment above).
> 
> > 
> > also I would separate it to multiple keyspaces, what you call base load should be on a different keyspace, and not limited
> > 
> 
> What will be benefit of that?
> 

It might help looking on each load seperatly,

But from the looks of those graphs is seems like it's clearer enough to see the results.

fruch (dismissed):

fruch:
@soyacz you might want to rebase to have the CI passing.

soyacz:
rebased, please reapprove so we can merge.

piodul (approved):

fruch (approved):

soyacz:
@roydahan I believe this should go to 2022.2 and 5.1 - do you approve?

roydahan:
Yes, and to run it "officially" for 2022.2 at least once. (you can use 2022.2.0-rc6).

soyacz:
Done. 
Test run: https://jenkins.scylladb.com/view/enterprise-2022.2/job/enterprise-2022.2/job/features-/job/features-per-partition-limit/4/
It failed in jenkins, but in reality it passed - error is due of flakiness of s-b in gathering results (showing +inf for op/s).

PR #5388, opened 2022-11-08
fix(pylint): remove duplication for cluster_driver

soyacz:
There's code duplication when getting cluster driver punishing us by pylint.


fruch:
#5389 is better right now, since this code isn't used in on of those places

fruch:
@soyacz I have that fix in one of my branch, I forgot to branch it off into a PR.

the main problem from my POV, that those code duplicate are not visible on every run of CI, and I never figured out why.
that mean that every time there a duplicate we miss that, cause pylint isn't always hitting it.

soyacz:
I didn't verify it is not needed - thanks for your fix (it is hitting on my other branch)

PR #5395, opened 2022-11-09
improvement(mv-si longevity): explicitly set STCS

fgelcer:
for OSS it shouldn't change anything, as STCS is the default, but for enterprise ones, now the default is ICS, hence to have both being tested, this one will be explicitly set to use STCS, while the ICS job does use explicitly ICS in it.

### DIFF
--- a/data_dir/c-s_profile_3si_5queries.yaml
+++ b/data_dir/c-s_profile_3si_5queries.yaml
@@ -19,7 +19,7 @@ table_definition: |
     userdata blob,
     last_access timeuuid,
     PRIMARY KEY(userid)
-  );
+  ) WITH compaction = {'class': 'SizeTieredCompactionStrategy'};
 
 extra_definitions:
   - CREATE INDEX IF NOT EXISTS  users_last_name_ind ON sec_index.users (last_name)
--- a/data_dir/c-s_profile_4mv_5queries.yaml
+++ b/data_dir/c-s_profile_4mv_5queries.yaml
@@ -19,7 +19,7 @@ table_definition: |
     userdata blob,
     last_access timeuuid,
     PRIMARY KEY(userid)
-  );
+  ) WITH compaction = {'class': 'SizeTieredCompactionStrategy'};
 
 extra_definitions:
   - CREATE MATERIALIZED VIEW mview.users_by_first_name AS SELECT userid, first_name, email FROM mview.users WHERE first_name IS NOT NULL and userid IS NOT NULL PRIMARY KEY (first_name, userid);


fruch:
@fgelcer maybe if ICS is the default now in enterprise, how about removing the duplicated job for ICS ? instead of keep both ?

fgelcer:
> @fgelcer maybe if ICS is the default now in enterprise, how about removing the duplicated job for ICS ? instead of keep both ?

this change is exactly solving the duplication, so, when this PR will be merged, it won't be "duplicated", but a completely different test.

fgelcer:
> > > @fgelcer maybe if ICS is the default now in enterprise, how about removing the duplicated job for ICS ? instead of keep both ?
> > 
> > 
> > this change is exactly solving the duplication, so, when this PR will be merged, it won't be "duplicated", but a completely different test.
> 
> I was asking why no test only with ICS for enterprise, and the other type would be covered in OSS ?

for now, the majority of tests are running on enterprise only, apart from the automatically triggered ones... so we still want to test STCS, hence this PR... if we decide to run for next release this job on OSS, we should skip running it on enteprise too (but only ICS)

yarongilor (commented):
What about he 2 additional yamls of:
c-s_profile_2mv_2queries.yaml, c-s_profile_2si_2queries.yaml?

yarongilor:
What about a different suggestion - instead of having this waste of duplication - why not having both STCS and ICS in the same test for enterprise?

fgelcer:
> What about he 2 additional yamls of: c-s_profile_2mv_2queries.yaml, c-s_profile_2si_2queries.yaml?

it is being used only to read, not to create tables, so IMO we don't need to change it

fgelcer:
> For ICS job do we have these with ICS?

your question isn't very clear, but for ICS test we have it's own files, setting the CS to be ICS

yarongilor:
> c-s_profile_2mv_2queries.yaml

i see it has also a `create table` query, that doesn't specify a compaction like `WITH compaction = {'class': 'SizeTieredCompactionStrategy'};`

fgelcer:
> > c-s_profile_2mv_2queries.yaml
> 
> i see it has also a `create table` query, that doesn't specify a compaction like `WITH compaction = {'class': 'SizeTieredCompactionStrategy'};`

it is not called in any of the relevant `prepare` commands session, so the `create table` is not called for this file

fgelcer:
ping @roydahan @yarongilor

yarongilor:
> > > c-s_profile_2mv_2queries.yaml
> > 
> > 
> > i see it has also a `create table` query, that doesn't specify a compaction like `WITH compaction = {'class': 'SizeTieredCompactionStrategy'};`
> 
> it is not called in any of the relevant `prepare` commands session, so the `create table` is not called for this file

this sounds confusing to me - why would the yaml include `table_definition` and 
```
insert:
  partitions: fixed(1)
  batchtype: UNLOGGED
```
if they are not in use?

fruch:
> 
> > > > c-s_profile_2mv_2queries.yaml
> > > 
> > > 
> > > i see it has also a `create table` query, that doesn't specify a compaction like `WITH compaction = {'class': 'SizeTieredCompactionStrategy'};`
> > 
> > it is not called in any of the relevant `prepare` commands session, so the `create table` is not called for this file
> 
> this sounds confusing to me - why would the yaml include `table_definition` and 
> ```
> insert:
>   partitions: fixed(1)
>   batchtype: UNLOGGED
> ```
> if they are not in use?

It would be used if precreate would be disabled.

fgelcer:
@roydahan , this is the PR we spoke while working on 2022.2, as this test originally used the defaults, making us test twice ICS on enterprise, hence in this test we explicitly set it as STCS, so in enterprise we will have both tests.
please merge so we can start testing 2023.1

roydahan (approved):

PR #5398, opened 2022-11-10
feature(pipeline_test_duration): Generic test_duration for longevity

aleksbykov:
QA-Proj: https://github.com/scylladb/qa-tasks/issues/854

Set generic test duration from job settings

    Set generic test duration from job settings
    
    Introduce 2 new sct parameters: prepare_duration and stress_duration
    
    If stress duration is set in job pipeline,then test duration is
    calculated as prepare_duration + stress_duration + spare_time(=10min)
    
    in sct the test_duration also is counted as prepare_duration + stress_duration
    if stress duration > 0 or used old parameter test_duration to run test
    as before.


sdcm/stress_thread.py
```diff
@@ -106,6 +106,10 @@ def create_stress_cmd(self, node, keyspace_idx):
             self.profile = cs_profile
             self.keyspace_name = keyspace_name
 
+        if 'duration' in stress_cmd:
```
fruch, line 109: should look for `duration=` the word duration is too general, can be in the name of the c-s use profile filename for example.

sdcm/stress_thread.py
```diff
@@ -106,6 +106,10 @@ def create_stress_cmd(self, node, keyspace_idx):
...
+            duration = self.loader_set.params.get("test_duration") - 20
```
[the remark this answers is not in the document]
fruch: I guess we always need the stress command smaller in some degree from then the test duration.
for example to calculate start the prepare stage, but it can't be a fixed number, and I'm not sure we can easily calculate it.

the only thing I can think of, is to lookup the history of the prepare, for that job, and reduce that.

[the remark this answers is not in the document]
fgelcer: problem is, i think 20 minutes would not be enough for most cases...

roydahan:
As general approach here is what I think we can do:
1. Assume a fixed number of "prepare_time" (the time it takes us to load data in all longevities that we have prepare_cmd.
(We can choose the longest time we know right now and possibly add it as a parameter to the pipeline to be override).

2. Get the duration as parameter and translate it basically to "stress_duration".

3. test timeout (test_duration) will be "prepare_time" + "stress_duration" + "spare_time" (if we need).

roydahan:
For this PR, let's take the 1TB longevity, longevity-large-partition-200k_pks-4days, gemini-with-nemesis-3h as examples and change them to use this new duration type.

sdcm/stress_thread.py
aleksbykov: fixed

vars/longevityPipeline.groovy
```diff
@@ -38,6 +38,14 @@ def call(Map pipelineParams) {
                description: 'Availability zone',
                name: 'availability_zone')
 
+            string(defaultValue: "",
+               description: 'Test duration for test run in minutes',
+               name: 'stress_duration')
+
+            string(defaultValue: "",
+               description: 'Test duration for test run in minutes',
```
soyacz, line 46: fix description

sdcm/sct_config.py
```diff
@@ -171,7 +171,16 @@ class SCTConfiguration(dict):
...
              """),
-
+        dict(name="prepare_duration", env="SCT_PREPARE_DURATION", type=int,
+             help="""
+                  Prepare test duration (min). Parameter used to run prepare dataset before
+                  main payload.
+             """),
+        dict(name="stress_duration", env="SCT_STRESS_DURATION", type=int,
+             help="""
+                  Prepare test duration (min). Parameter used to run prepare dataset before
```
soyacz, line 181: fix description

sdcm/sct_config.py
```diff
@@ -171,7 +171,16 @@ class SCTConfiguration(dict):
                   Test duration (min). Parameter used to keep instances produced by tests
                   and for jenkins pipeline timeout and TimoutThread.
              """),
-
+        dict(name="prepare_duration", env="SCT_PREPARE_DURATION", type=int,
+             help="""
+                  Prepare test duration (min). Parameter used to run prepare dataset before
```
soyacz, line 176: this description is misleading. Can you rephrase it?

aleksbykov: @soyacz , can you check, is it become more clear?

aleksbykov:
@roydahan @fruch @soyacz @fgelcer @vponomaryov Can you take a look?

sdcm/sct_config.py
```diff
@@ -171,7 +171,17 @@ class SCTConfiguration(dict):
                   Test duration (min). Parameter used to keep instances produced by tests
                   and for jenkins pipeline timeout and TimoutThread.
              """),
-
+        dict(name="prepare_duration", env="SCT_PREPARE_DURATION", type=int,
+             help="""
+                  Time in minutes, which is required to run prepare stress commands
+                  defined in prepare_*_cmd for dataset generation, and is used in
+                  test duration calculation
```
fgelcer, line 178: is it used for `test duration calculation` only?

sdcm/sct_config.py
```diff
@@ -171,7 +171,17 @@ class SCTConfiguration(dict):
...
+             """),
+        dict(name="stress_duration", env="SCT_STRESS_DURATION", type=int,
+             help="""
+                  Time in minutes, Time of execution for stress commands from stress_cmd parameters
+                  and is used in test duration calculation
```
fgelcer, line 183: same here, will it be used to `test duration calculation` only?

sdcm/tester.py
```diff
@@ -300,7 +300,15 @@ def __init__(self, *args, **kwargs):  # pylint: disable=too-many-statements,too-
         self.log.debug("IP used for SSH connections is '%s'",
                        ip_ssh_connections)
         self.test_config.set_ip_ssh_connections(ip_ssh_connections)
-        self._duration = self.params.get(key='test_duration')
+
+        self._stress_duration: int = self.params.get('stress_duration')
+        self._prepare_duration: int = self.params.get('prepare_duration')
+        if self._stress_duration > 0:
```
fgelcer, line 306: you don't need this check here, because if this is the default (0) it won't add anything, but if it has any other value, you will sum it up...
and if someone creates this parameter with negative, this is operator error, but perhaps could be enforced in the `sct_config` setting for this parameter

fgelcer: or perhaps it should be `if not self._duration`

sdcm/tester.py
```diff
@@ -1779,8 +1787,12 @@ def run_stress_cassandra_thread(
             stats_aggregate_cmds=True, keyspace_name=None, compaction_strategy='', stop_test_on_failure=True, params=None, **_):
         # pylint: disable=too-many-locals
         # stress_cmd = self._cs_add_node_flag(stress_cmd)
+
         if duration:
             timeout = self.get_duration(duration)
+        elif self._stress_duration > 0 and 'duration=' in stress_cmd:
```
fgelcer, line 1793:
```suggestion
        elif self._stress_duration and 'duration=' in stress_cmd:
```

sdcm/tester.py
```diff
@@ -1815,6 +1827,9 @@ def run_stress_thread_bench(self, stress_cmd, duration=None, round_robin=False,
 
         if duration:
             timeout = self.get_duration(duration)
+        elif self._stress_duration > 0 and 'duration=' in stress_cmd:
```
fgelcer, line 1830:
```suggestion
        elif self._stress_duration and 'duration=' in stress_cmd:
```

sdcm/tester.py
```diff
@@ -1966,6 +1981,12 @@ def run_cdclog_reader_thread(self, stress_cmd, duration=None, stress_num=1, pref
     def run_gemini(self, cmd, duration=None):
         if duration:
             timeout = self.get_duration(duration)
+        elif self._stress_duration > 0 and '--duration' in cmd:
```
fgelcer, line 1984:
```suggestion
        elif self._stress_duration and '--duration' in cmd:
```

test-cases/longevity/longevity-large-partition-8h.yaml
```diff
@@ -5,20 +5,20 @@ max_partitions_in_test_table: 400
...
-                     "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=100 -clustering-row-count=5555 -partition-offset=301 -clustering-row-size=uniform:1024..2048 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=30s"
+                     # "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=100 -clustering-row-count=5555 -partition-offset=201 -clustering-row-size=uniform:1024..2048 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=30s",
+                     # "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=100 -clustering-row-count=5555 -partition-offset=301 -clustering-row-size=uniform:1024..2048 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=30s"
```
fgelcer, line 9: either remove dead code, or uncomment.. i believe it was used during debugging this new PR

test-cases/longevity/longevity-large-partition-8h.yaml
```diff
@@ -5,20 +5,20 @@ max_partitions_in_test_table: 400
...
-                     "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=25 -clustering-row-count=5555 -partition-offset=76 -clustering-row-size=uniform:1024..2048 -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -iterations 1 -validate-data"
+                     # "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=25 -clustering-row-count=5555 -partition-offset=51 -clustering-row-size=uniform:1024..2048 -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -iterations 1 -validate-data",
+                     # "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=25 -clustering-row-count=5555 -partition-offset=76 -clustering-row-size=uniform:1024..2048 -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -iterations 1 -validate-data"
```
fgelcer, line 15: same here, dead code

test-cases/longevity/longevity-large-partition-8h.yaml
```diff
@@ -34,8 +34,8 @@ post_prepare_cql_cmds: "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELEC
 #                  "scylla-bench -workload=uniform -mode=read -replication-factor=3 -partition-count=1000 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -rows-per-request=10 -consistency-level=quorum -timeout=30s -retry-number=30 -retry-interval=80ms,1s -partition-offset=1001 -concurrency=100 -connection-count=100 -duration=6420m"
 #                 ]
 
-n_db_nodes: 5
-n_loaders: 4
+n_db_nodes: 3
+n_loaders: 2
```
fgelcer, line 38: was it changed in purpose, or just for testing this PR?

test-cases/longevity/longevity-large-partition-8h.yaml
```diff
@@ -34,8 +34,8 @@ post_prepare_cql_cmds: "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELEC
 #                  "scylla-bench -workload=uniform -mode=read -replication-factor=3 -partition-count=1000 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -rows-per-request=10 -consistency-level=quorum -timeout=30s -retry-number=30 -retry-interval=80ms,1s -partition-offset=1001 -concurrency=100 -connection-count=100 -duration=6420m"
 #                 ]
```
fgelcer, line 35: remove dead code

sdcm/test_config.py
```diff
@@ -22,6 +22,7 @@
 
 class TestConfig(metaclass=Singleton):  # pylint: disable=too-many-public-methods
     TEST_DURATION = 60
+    TEST_WARMUP_TEARDOWN = 60
```
fgelcer, line 25: why do we need this here, if we already have in `getJobTimeouts.groovy`:

```
Integer testStartupTimeout = 20
Integer testTeardownTimeout = 40
```

vars/getJobTimeouts.groovy
```diff
@@ -2,17 +2,24 @@
...
+        export SCT_CONFIG_FILES=${test_config}
+        ./docker/env/hydra.sh output-conf -b "${params.backend}"
+        """
+        def testData = sh(script: cmd, returnStdout: true).trim()
+        println(testData)
+        testData = testData =~ /test_duration: (\d+)/
+        testDuration = testData[0][1].toInteger()
+    } else {
+        stressDuration = params.stress_duration.toInteger()
+        prepareDuration = params.prepare_duration.toInteger()
```
fgelcer, line 19: what will happen if these parameters don't exist, or are set to anything that not integer? shouldn't we consider it? something like:

```suggestion
        stressDuration = params.stress_duration.toInteger() ?: 0
        prepareDuration = params.prepare_duration.toInteger() ?: 0
```

vars/runSctTest.groovy
```diff
@@ -27,6 +27,14 @@ def call(Map params, String region, functional_test = false, Map pipelineParams
     export SCT_CONFIG_FILES=${test_config}
     export SCT_COLLECT_LOGS=false
 
+    if [[ -n "${params.stress_duration ? params.stress_duration : ''}" ]] ; then
```
fgelcer, line 30:
```suggestion
    if [[ -n "${params.stress_duration ? params.stress_duration : '0'}" ]] ; then
```

vars/runSctTest.groovy
```diff
@@ -27,6 +27,14 @@ def call(Map params, String region, functional_test = false, Map pipelineParams
...
+        export SCT_STRESS_DURATION=${params.stress_duration}
+    fi
+
+    if [[ -n "${params.prepare_duration ? params.prepare_duration : ''}" ]] ; then
```
fgelcer, line 34:
```suggestion
    if [[ -n "${params.prepare_duration ? params.prepare_duration : '0'}" ]] ; then
```

sdcm/sct_config.py
aleksbykov: Also it used in gemini as warmup period. Fixed description

sdcm/sct_config.py
aleksbykov: change description. is it clear now?

fgelcer: yes

sdcm/sct_config.py
fgelcer: thank you

vars/getJobTimeouts.groovy
```diff
@@ -2,17 +2,24 @@
...
-    #!/bin/bash
-    export SCT_CLUSTER_BACKEND="${params.backend}"
-    export SCT_CONFIG_FILES=${test_config}
-    ./docker/env/hydra.sh output-conf -b "${params.backend}"
-    """
-    def testData = sh(script: cmd, returnStdout: true).trim()
-    println(testData)
-    testData = testData =~ /test_duration: (\d+)/
-    testDuration = testData[0][1].toInteger()
+    if (params.stress_duration == "") {
```
soyacz, line 5: in case `params.stress_duration` is not defined in pipeline at all (we have other pipelines that use this function, if not all of them), I think this might be considered `false` and fail in else on integer conversion. May I be right?

vars/getJobTimeouts.groovy
soyacz: in case it does not exist, running this else condition should not happen at all.

vars/runSctTest.groovy
soyacz: but this change will make `if (params.stress_duration == "") {` from `vars/getJobTimeouts.groovy` to go to else condition which is wrong.

sdcm/tester.py
```diff
@@ -300,7 +300,15 @@ def __init__(self, *args, **kwargs):  # pylint: disable=too-many-statements,too-
         self.log.debug("IP used for SSH connections is '%s'",
                        ip_ssh_connections)
         self.test_config.set_ip_ssh_connections(ip_ssh_connections)
-        self._duration = self.params.get(key='test_duration')
+
+        self._stress_duration: int = self.params.get('stress_duration')
```
soyacz, line 304: will it be `int` if `stress_duration` is not set at all? Shouldn't we set default to 0 in sct_configuration?

soyacz (changes requested):
What happens when we shorten preparation time and not all partitions from prepare step will be inserted? Won't then stress process exit with error when keys are missing?
Please verify that and refer other comments from @fgelcer 
Let's move this PR to closing up.

vars/getJobTimeouts.groovy
aleksbykov: @soyacz is right. Parameter defined as job parameter in LongevityPipeline. so it always will be set at least with default value ''. So if it is not configured , this if condition will be skipped. and old behavior will be used.
in case that this parameter will set other than integer, groovy will raise an exception and job will not be run. I don't see, that it is need to add any additional checks on groovy and do code more complex.

in case that these parameters will be set as negative integers, i will add check to sct_config validations

vars/runSctTest.groovy
aleksbykov: it won't because the getJobTimeout is running earlier and will not affect on runSctTest, but there is no need to set this to 0, because if stress_parameter is not set in pipeline parameter, we will not export appropriate SCT env, and will use default value which is 0 and means that old behavior will be used (test_duration and stress command duration defined inside command)

vars/runSctTest.groovy
aleksbykov: Same as above, if it is not set we don't need to export env variable, checks for values other than interger, will failed job earlier on time calculation. Additional check will be add to sct_config validator

sdcm/tester.py
aleksbykov: Changed

sdcm/tester.py
aleksbykov: fixed

sdcm/tester.py
aleksbykov: fixed

sdcm/tester.py
aleksbykov: fixed

test-cases/longevity/longevity-large-partition-8h.yaml
aleksbykov: removed

test-cases/longevity/longevity-large-partition-8h.yaml
aleksbykov: return

sdcm/tester.py
aleksbykov: yes, in this case it will taken from test_default which is 0 by default.

aleksbykov:
> What happens when we shorten preparation time and not all partitions from prepare step will be inserted? Won't then stress process exit with error when keys are missing? Please verify that and refer other comments from @fgelcer Let's move this PR to closing up.

yes it could be a case. While we don't have any statistics on how log prepare stress could run, i set default as 5h (for most longevities it is quite enough). prepare_duration will be use only for counting the total test duration for that moment.

aleksbykov:
> @aleksbykov how long did the prepare take? IIRC it should take around 2 hours.

I checked the logs. It took 3h

soyacz (approved):

sdcm/sct_config.py
```diff
@@ -1775,6 +1785,18 @@ def __init__(self):
                 raise ValueError("For PasswordAuthenticator authenticator authenticator_user and authenticator_password"
                                  " have to be provided")
 
+        # 13) validate stress and prepare duration:
+        if stress_duration := self.get('stress_duration'):
```
vponomaryov, line 1789: Why the `prepare_duration` depends on the setting of the `stress_duration`?

sdcm/sct_config.py
```diff
@@ -1775,6 +1785,18 @@ def __init__(self):
...
+            try:
+                self['stress_duration'] = abs(int(stress_duration))
+            except ValueError:
+                raise ValueError(f'Configured stress duration for generic test duratinon have to be \
```
vponomaryov, line 1793: The message is not correct. Your code will transform negative int value into positive using the `abs` function.

sdcm/sct_config.py
```diff
@@ -172,7 +172,17 @@ class SCTConfiguration(dict):
                   Test duration (min). Parameter used to keep instances produced by tests
                   and for jenkins pipeline timeout and TimoutThread.
              """),
-
+        dict(name="prepare_duration", env="SCT_PREPARE_DURATION", type=int,
```
vponomaryov, line 175: Can we name it `stress_prepare_duration`?
It sounds less ambiguous to me.
Also, it becomes more familiar with the second new option - `stress_duration`.

sdcm/tester.py
```diff
@@ -374,6 +374,19 @@ def __init__(self, *args, **kwargs):  # pylint: disable=too-many-statements,too-
         time.sleep(0.5)
         InfoEvent(message=f"TEST_START test_id={self.test_config.test_id()}").publish()
 
+    def _init_test_duration(self):
+        self._stress_duration: int = self.params.get('stress_duration')
+        if self._stress_duration:
```
vponomaryov, line 379: Same here. Why the `prepare_duration` depends on the `stress_duration`?
Why I cannot use one without another? And why the `prepare` one gets ignored silently if second new option is not set?

test-cases/longevity/longevity-large-partition-8h.yaml
```diff
@@ -15,10 +15,10 @@ prepare_verify_cmd: ["scylla-bench -workload=sequential -mode=read -replication-
...
-             "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=25 -clustering-row-count=10    -partition-offset=451 -clustering-row-size=uniform:5120..8192   -concurrency=10  -connection-count=10  -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 10",
-             "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=25 -clustering-row-count=1     -partition-offset=476 -clustering-row-size=uniform:8192..10240  -concurrency=10  -connection-count=10  -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 10",
+stress_cmd: ["scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=25 -clustering-row-count=10000 -partition-offset=401 -clustering-row-size=uniform:10..1024     -concurrency=10  -connection-count=10  -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 0 -duration=540m",
```
vponomaryov, line 18: This case actually shows the newly introduced limitation.
I remember that in some cases we have different durations for different stress commands.
And with this change we won't be able to set explicit different durations. Right?

vars/getJobTimeouts.groovy
```diff
@@ -4,26 +4,38 @@ List<Integer> call(Map params, String region){
...
-    if [[ -n "${params.azure_region_name ? params.azure_region_name : ''}" ]] ; then
-        export SCT_AZURE_REGION_NAME=${groovy.json.JsonOutput.toJson(params.azure_region_name)}
-    fi
-    ./docker/env/hydra.sh output-conf -b "${params.backend}"
-    """
+        if [[ -n "${params.azure_region_name ? params.azure_region_name : ''}" ]] ; then
+            export SCT_AZURE_REGION_NAME=${groovy.json.JsonOutput.toJson(params.azure_region_name)}
+        fi
+        ./docker/env/hydra.sh output-conf -b "${params.backend}"
+        """
```
vponomaryov, line 22: Why do we need this indentation change?

vars/getJobTimeouts.groovy
```diff
@@ -4,26 +4,38 @@ List<Integer> call(Map params, String region){
...
     def testData = sh(script: cmd, returnStdout: true).trim()
     println(testData)
-    testData = testData =~ /test_duration: (\d+)/
-    testDuration = testData[0][1].toInteger()
+    if (params.stress_duration == "") {
```
vponomaryov, line 25: Same here. The `prepare_duration` depends on the `stress_duration` one. Why?

vponomaryov (changes requested):

sdcm/sct_config.py
aleksbykov: The stress duration set  duration of stress commands. But before that we run prepare commands, which build the dataset.
if stress_duration doestn' set (old behavior), the test_duration already include time which is need to run prepare dataset commands.
But if we set the stress duration, then new test duration should be counted based on provided stress duration and time for building dataset. But different job use different dataset which is generated for different time. One of advantage is that if prepare stress cmd finished ealier, then we don't wait full time. So after discussion we decide to use most longest time for prepare command is ~5h. in future we may want to configure prepare time. and based on it we should count test_duration.

So for new behavior, if stress_duration is set we additionally have to get time for prepare stress commands if it is not set, then default will be used, and new test duration will be calculated.

sdcm/tester.py
aleksbykov: answered above

test-cases/longevity/longevity-large-partition-8h.yaml
aleksbykov: Yes, you right. for common case, all stress commands will be run with same stress duration. 
For specific case, when stress commands should be run with different time( why? ) this will not work, and it should be run like previously without setting stress_duration parameter

vars/getJobTimeouts.groovy
aleksbykov: will check that, looks like IDE do it itself.

vars/getJobTimeouts.groovy
aleksbykov: the same as above

sdcm/sct_config.py
vponomaryov: I understood the logic of the implementation. I just disagree that it is correct way for doing it.
The `stress prepare` and `main stress` commands run only in separate time frames and both combine the main test run spent time excluding the provisioning part.
So, from my point of view, both are equal and it should be possible to handle them separately.

vars/getJobTimeouts.groovy
vponomaryov: If you want to keep it, then need to make it a separate commit.
If you don't care, then need to remove this code change.

test-cases/longevity/longevity-large-partition-8h.yaml
vponomaryov: > ( why? )

https://github.com/scylladb/scylla-cluster-tests/pull/6080#discussion_r1192468836

vars/getJobTimeouts.groovy
fruch: @aleksbykov please remove this change, it's not related to the commit

roydahan:
@aleksbykov what's the status of this PR?

vars/runSctTest.groovy
fruch: this is fine and follow the logic how we use env variables across the pipeline

vars/runSctTest.groovy
fruch: this is fine and follow the logic how we use env variables across the pipeline

vars/getJobTimeouts.groovy
```diff
@@ -4,26 +4,38 @@ List<Integer> call(Map params, String region){
...
+        testDuration = testData[0][1].toInteger()
+    } else {
+        stressDuration = params.stress_duration.toInteger()
+        try {
+            prepareDuration = params.prepare_duration.toInteger()
+        } catch (e) {
+            testData = testData =~ /prepare_duration: (\d+)/
+            prepareDuration = testData[0][1].toInteger()
+        }
+        Integer stressEndup = 10
```
fruch, line 36: 10min ? I think it need at least 30min, which is the max time we wait for nemesis to finish

sdcm/sct_config.py
fruch: @aleksbykov you said yes, but you didn't change it..

sdcm/sct_config.py
fruch: I agree with @vponomaryov those two should work independently

sdcm/sct_config.py
```diff
@@ -1775,6 +1785,18 @@ def __init__(self):
...
+        if stress_duration := self.get('stress_duration'):
+            try:
+                self['stress_duration'] = abs(int(stress_duration))
+            except ValueError:
+                raise ValueError(f'Configured stress duration for generic test duratinon have to be \
+                                 positive integer number in minutes. Current value: {stress_duration}') from ValueError
+            try:
+                self['prepare_duration'] = abs(int(self.get('prepare_duration')))
+            except ValueError:
+                self['prepare_duration'] = 300  # 5h
```
fruch, line 1798: why do we have default here in the code ? if the value is set incorrect, we should fail.

sdcm/sct_config.py
aleksbykov: in cases the stress_duration will be set as negative int (-50) it will be converted to positive without errors, but if value will some string, then exception will inform that it should be set as positive integer

vars/getJobTimeouts.groovy
aleksbykov: it is time to wait while stress command wait to finished. It is configured in hour sct code for stress commands

sdcm/sct_config.py
aleksbykov: set raise exception

vars/getJobTimeouts.groovy
aleksbykov: done

sdcm/sct_config.py
aleksbykov: done

sdcm/sct_config.py
aleksbykov: fixed

sdcm/sct_config.py
vponomaryov: The `string` value will be transformed into the `integer` value by the `int()` function before processing by the `abs()` one.

vars/runSctTest.groovy
```diff
@@ -27,6 +27,14 @@ def call(Map params, String region, functional_test = false, Map pipelineParams
     export SCT_CONFIG_FILES=${test_config}
     export SCT_COLLECT_LOGS=false
 
+    if [[ -n "${params.stress_duration ? params.stress_duration : ''}" ]] ; then
+        export SCT_STRESS_DURATION=${params.stress_duration}
+    fi
+
+    if [[ -n "${params.prepare_stress_duration ? params.prepare_stress_duration : ''}" ]] ; then
+        export SCT_prepare_stress_duration=${params.prepare_stress_duration}
```
vponomaryov, line 35: Wrong env var name

vars/getJobTimeouts.groovy
vponomaryov: This is still actual having the same change be done in the `sct_config.py`.

sdcm/tester.py
vponomaryov: This is still actual having the same change be done in the `sct_config.py`.

vponomaryov (changes requested):

vars/runSctTest.groovy
aleksbykov: fixed

aleksbykov:
https://github.com/orgs/scylladb/teams/qa-maintainers could you review?
@Roy could you also review?

aleksbykov:
Current implementation:
1. if stress_duration is not set, then used test_duration and job run as is
2. if stress_duration is set, then check if not set prepare_stress_duration, use default value, else get value count new test_duration  as prepare_stress_duration + stress_duration
These implemenation could bring errors for test, if prepare_stress_duration configures small, and prepare took more time and then test_duration exceed earlier than main stress command finished. Basic workaround don't set prepare_stress_duration and use default value (5h), If prepare stress commands fiinished earlier, test continue and will not wait full prepare_stress_duration time.

Another approach as @vponomaryov suggested, is provided to independent variables:
`If we set only prepare step limit then we raise error if it is exceeded. 
If we set both - raise error if any of them exceeds it's limit.
So, it turns out that the "test_duration" is redundant actually, we should have separate timeouts for each of the stages`

roydahan:
> 

In general, I don't know why we even need a duration for prepare because prepare is not limited with time, it's limited by data it has to write.
However, I'm fine with the current implementation and the current default limit of 5h, it will help us to learn where our prepare is too long and may need to be adjusted.

roydahan:
@scylladb/qa-maintainers if approved (not sure since all approvers were dismissed), let's re-approve and merge.
I would like start using it next week.

fgelcer:
> > 
> 
> In general, I don't know why we even need a duration for prepare because prepare is not limited with time, it's limited by data it has to write. However, I'm fine with the current implementation and the current default limit of 5h, it will help us to learn where our prepare is too long and may need to be adjusted.

i didn't check the implementation, but cutting off part of the prepare based on time, other than number of rows would cause tests to fail on read, if the key we are trying to access doesn't exist

sdcm/sct_config.py
fruch: remove all this validation, @vponomaryov is correct that it's already gonna happen automatically

sdcm/tester.py
fruch: having it based `stress_duration` is o.k., since we want the default behavior to be left alone, and there no point of using `prepare_duration` on it's own

fruch:
> > > 
> > 
> > 
> > In general, I don't know why we even need a duration for prepare because prepare is not limited with time, it's limited by data it has to write. However, I'm fine with the current implementation and the current default limit of 5h, it will help us to learn where our prepare is too long and may need to be adjusted.
> 
> i didn't check the implementation, but cutting off part of the prepare based on time, other than number of rows would cause tests to fail on read, if the key we are trying to access doesn't exist

only if we have validation turned on, otherwise it won't fail.
there's an assumption that 5h would be enough for any prepare, we won't know if it's true until we'll try all the cases.

test-cases/longevity/longevity-large-partition-8h.yaml
fruch: @vponomaryov it's really not clear why anyone would want not all the stress command to be with the same duration.
it's kind of contradicts what longevity are usually for, on other cases like feature specific we might still do those kind of things.

vars/getJobTimeouts.groovy
fruch: o.k.

fruch (changes requested):
Remove unneeded validation code, and I think we can merge it, and give it a go

roydahan:
The time to optimize this PR has passed.
@aleksbykov is on vacation and I'm going to merge it as is.

Unneeded validations or nipticks can be improved later.

fgelcer:
> The time to optimize this PR has passed. @aleksbykov is on vacation and I'm going to merge it as is.
> 
> Unneeded validations or nitpicks can be improved later.

@roydahan , as it is not yet merged, once you do, please run both triggers:
https://jenkins.scylladb.com/job/scylla-master/job/tier1-custom-time-trigger/
https://jenkins.scylladb.com/job/scylla-master/job/tier2-custom-time-trigger/

roydahan:
Sadly I need to merge PR without any approval.
@aleksbykov and @scylladb/qa-maintainers we should improve the cooperation on our PRs.
It's not acceptable that I ask 3 weeks ago to prioritize it and we drag our feet for so long.

fruch:
Seems like all pipelines (beside longevity) are broken cause of this change:

https://jenkins.scylladb.com/job/scylla-master/job/rolling-upgrade/job/rolling-upgrade-ubuntu20.04-test/141/
```
java.lang.NullPointerException: Cannot invoke method toInteger() on null object
```

this might the cause, that it's not handling null, and assuming empty string:

```diff
diff --git a/vars/getJobTimeouts.groovy b/vars/getJobTimeouts.groovy
index 20e1ed5bc..8ed0e2279 100644
--- a/vars/getJobTimeouts.groovy
+++ b/vars/getJobTimeouts.groovy
@@ -22,7 +22,7 @@ List<Integer> call(Map params, String region){
     """
     def testData = sh(script: cmd, returnStdout: true).trim()
     println(testData)
-    if (params.stress_duration == "") {
+    if (params.stress_duration == "" || params.stress_duration == null) {
         testData = testData =~ /test_duration: (\d+)/
         testDuration = testData[0][1].toInteger()
     } else {
```

roydahan:
We will handle it tomorrow.

fgelcer:
@scylladb/qa-maintainers , can you please backport this PR to `branch-2023.1`?

fruch:
> @scylladb/qa-maintainers , can you please backport this PR to `branch-2023.1`?

why do we need it in 2023.1 ? 

also there was a few follow up on this one that would need to be backported.

fgelcer:
> > @scylladb/qa-maintainers , can you please backport this PR to `branch-2023.1`?
> 
> why do we need it in 2023.1 ?

because we want to enhance the patch releases on all enterprise supported versions (we will use SCT branch-2023.1 for it).

> also there was a few follow up on this one that would need to be backported.

from the search i did, only #6422 came up... of course if you have more, please let's backport all of them

fruch:
@fgelcer and what about all the followups to the tests that were done cause of that ?

PR #5402, opened 2022-11-11
Revert "feature(cassandra-stress): move to run on docker base image"

fruch:
Reverts scylladb/scylla-cluster-tests#5146

Seem like recent in recent scylla images (that moved to newer ubuntu), cassandra-stress stopped working,
and I've found that moving to newer docker version solves the issue. 
since it would take a while to fix properly, I'm reverting this series. 

the error looks like this:
```
Stress command completed with bad status 1: You must set the CLASSPATH var
```


fruch:
I've found the root cause, we have different docker version on our loader AMI, between regions !!!!

fgelcer:
> I've found the root cause, we have different docker version on our loader AMI, between regions !!!!
> 
> so our `us-east-1` image breaks, while `eu-west-1` works (this on using in provision test)
> 
> So I'll re-trigger the weekly with a different region first, before reverting.

that does make sense, because some time ago we lost part of our loaders, and while copying them again (and re-installing) we possibly ended up with different versions

fruch:
@scylladb/qa-maintainers 

Seems like running on different region didn't do the trick

I've decided to revert

PR #5416, opened 2022-11-14
fix(k8s-eks): make i4i instance types be supported for the Scylla nodes

vponomaryov:
Using `i3` nodes we have local NVME disks be attached to the `/dev/md0` path, 
but in the `i4i` ones it is `/dev/nvme0n1p1`. 
So, make the `pv-setup` on EKS consider both cases.

### DIFF
--- a/sdcm/k8s_configs/eks/scylla-node-prepare.yaml
+++ b/sdcm/k8s_configs/eks/scylla-node-prepare.yaml
@@ -235,7 +235,7 @@ spec:
 
             # Create directories for each Scylla cluster on each K8S node which will host PVs
             while true; do
-              if [[ -z $(mount | grep  "/dev/md0") ]]; then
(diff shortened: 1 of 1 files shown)


fruch:
Are you sure ?

/dev/md0 is the name of the raid device the  scripts should be creating, maybe on i3 there was multiple disks and we were creating raid0 on all of them, and i4i has only one device?

fruch:
> Are you sure ?
> 
> /dev/md0 is the name of the raid device the scripts should be creating, maybe on i3 there was multiple disks and we were creating raid0 on all of them, and i4i has only one device?

change the code in scylla again, and yes we are not using raid if there only one disk.
there's a flag to force it to create raid with only on disk

fruch (approved):
LGTM

Issue #5418, opened 2022-11-13
coredump_on_timeout for nodetool command is ignored - SCT will trigger coredump for every nodetool timeout!


yarongilor:
## Installation details

Kernel Version: 5.15.0-1022-aws
Scylla version (or git commit hash): `2022.2.0~rc4-20221106.f5714e0db12f` with build-id `f4a927b8a00fbcd8d48640835192aeaa7968b1f2`
Relocatable Package: http://downloads.scylladb.com/unstable/scylla-enterprise/enterprise-2022.2/relocatable/2022-11-06T15:44:05Z/scylla-enterprise-x86_64-package.tar.gz
Cluster size: 6 nodes (r5b.2xlarge)

Scylla Nodes used in this run:
- longevity-100gb-4h-2022-2-db-node-b805f708-6 (54.216.111.37 | 10.4.3.43) (shards: 7)
- longevity-100gb-4h-2022-2-db-node-b805f708-5 (34.246.189.190 | 10.4.0.93) (shards: 7)
- longevity-100gb-4h-2022-2-db-node-b805f708-4 (54.75.3.2 | 10.4.1.46) (shards: 7)
- longevity-100gb-4h-2022-2-db-node-b805f708-3 (63.35.237.246 | 10.4.1.180) (shards: 7)
- longevity-100gb-4h-2022-2-db-node-b805f708-2 (176.34.150.92 | 10.4.1.44) (shards: 7)
- longevity-100gb-4h-2022-2-db-node-b805f708-1 (3.252.186.100 | 10.4.2.249) (shards: 7)

OS / Image: `ami-00253925d2028dcef` (aws: eu-west-1)

Test: `longevity-100gb-4h-ebs-gp3-test`
Test id: `b805f708-075e-43cf-b7d9-fad1e63afce7`
Test name: `enterprise-2022.2/longevity/longevity-100gb-4h-ebs-gp3-test`
Test config file(s):

- [longevity-100gb-4h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/519e3909374cd87064c140c4badb2acc05805682/test-cases/longevity/longevity-100gb-4h.yaml)

## Issue description

**>>>>>>>**
scenario:
1. load was running for an hour, before any nemesis started.
2. node-1 got a coredump.
**<<<<<<<**
coredump details:
```
[...]

PR #5420, opened 2022-11-15
fix(cluster.py): Generate a scylla core dump according to boolean flag

yarongilor:
When run_nodetool command gets an exception it ignored coredump_on_timeout parameter.
	This is fixed and also added an event before the actual node kill starts.
	Refs: #5418

Fix: #5418 

### DIFF
--- a/defaults/severities.yaml
+++ b/defaults/severities.yaml
@@ -95,7 +95,7 @@ TestFrameworkEvent: CRITICAL
 ElasticsearchEvent: ERROR
 SpotTerminationEvent: CRITICAL
 ScyllaRepoEvent: WARNING
-InfoEvent: NORMAL
+InfoEvent: ERROR
 ThreadFailedEvent: ERROR
 CoreDumpEvent: ERROR
 TestResultEvent: ERROR
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -98,7 +98,7 @@ from sdcm.utils.version_utils import SCYLLA_VERSION_RE, get_gemini_version, get_
 from sdcm.sct_events import Severity
 from sdcm.sct_events.base import LogEvent
 from sdcm.sct_events.health import ClusterHealthValidatorEvent
-from sdcm.sct_events.system import TestFrameworkEvent, INSTANCE_STATUS_EVENTS_PATTERNS
+from sdcm.sct_events.system import TestFrameworkEvent, INSTANCE_STATUS_EVENTS_PATTERNS, InfoEvent
 from sdcm.sct_events.grafana import set_grafana_url
 from sdcm.sct_events.database import SYSTEM_ERROR_EVENTS_PATTERNS, ScyllaHelpErrorEvent
 from sdcm.sct_events.nodetool import NodetoolEvent
@@ -2448,9 +2448,12 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
                 self.remoter.run('cp -r %s %s' % (f, new_full_path), ignore_status=True)
         return root_dir
 
-    def generate_coredump_file(self, restart_scylla=True):
-        self.log.info('Generate scylla core')
-        self.remoter.run("sudo pkill -f --signal 3 /usr/bin/scylla")
+    def generate_coredump_file(self, restart_scylla=True, node_name: str = None, message: str = None):
+        message = message or "Generating a Scylla core dump by SIGQUIT..\n"
+        if node_name:
+            message += f"Node: {node_name}"
+        InfoEvent(severity=Severity.ERROR, message=message)
+        self.remoter.sudo("pkill -f --signal 3 /usr/bin/scylla")
         self.wait_db_down(timeout=600)
         if restart_scylla:
             self.log.debug('Restart scylla server')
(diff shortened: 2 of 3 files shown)


sdcm/cluster.py
```diff
@@ -2551,8 +2551,8 @@ def run_nodetool(self, sub_cmd, args="", options="", timeout=None,
             except Exception as details:  # pylint: disable=broad-except
                 if isinstance(details, RetryableNetworkException):
                     details = details.original
-                if details.__class__.__name__.endswith("CommandTimedOut"):
-                    self.generate_coredump_file()
+                if coredump_on_timeout and details.__class__.__name__.endswith("CommandTimedOut"):
+                    self.generate_coredump_file(node=self.name)
```
fgelcer, line 2555: here you should add a message, to override the default message in the `GeneratingCoreDumpEvent` class...
it should be something like, `sending SIGQUIT because command reached timeout` to have much more detailed print before we create these coredumps on timeout

sdcm/cluster.py
```diff
@@ -2448,8 +2448,8 @@ def prepare_files_for_archive(self, fileslist):
                 self.remoter.run('cp -r %s %s' % (f, new_full_path), ignore_status=True)
         return root_dir
 
-    def generate_coredump_file(self, restart_scylla=True):
-        self.log.info('Generate scylla core')
+    def generate_coredump_file(self, restart_scylla=True, node: Any = None):
+        GeneratingCoreDumpEvent(node=node).publish()
```
fruch, line 2452: What's wrong with using InformationalEvent

sdcm/cluster.py
```diff
@@ -2448,8 +2448,8 @@ def prepare_files_for_archive(self, fileslist):
                 self.remoter.run('cp -r %s %s' % (f, new_full_path), ignore_status=True)
         return root_dir
 
-    def generate_coredump_file(self, restart_scylla=True):
-        self.log.info('Generate scylla core')
+    def generate_coredump_file(self, restart_scylla=True, node: Any = None):
```
fruch, line 2451: node isn't Any, it should be BaseNode, otherwise there's no meaning for adding type annotations here

sdcm/cluster.py
yarongilor: Roy asked for an error event. and it makes sense since some error must have happened to the node in this scenario,

sdcm/cluster.py
yarongilor: perhaps it better be either a BaseNode or an STR?

sdcm/cluster.py
fruch: first, unittest are failing with this change:
```
ValueError: There is no max severity configured for GeneratingCoreDumpEvent
```

What's wrong with:

```
InformationalEvent(severity=Severity.ERROR)
```

adding an event to sct would normally mean there more then one place it's going to called from
and there more specific information it's going to hold.

in this case you just want an error event raised, with a clear message, I think `InformationalEvent` is enough

roydahan: InformationalEvent(severity=Severity.ERROR) is good enough.
We don't need a new type of event.

yarongilor: according to @fruch , he meant using `InfoEvent` (not InformationalEvent). but then calling `InfoEvent(severity=Severity.ERROR)` may not be sufficient in order to override severity.

yarongilor: fixed

sdcm/cluster.py
yarongilor: fixed

yarongilor:
Comments are addressed. if unit-test will fail then there might be a small update needed for the event severity.

fruch (approved):
LGTM

sdcm/cluster.py
```diff
@@ -2448,8 +2448,11 @@ def prepare_files_for_archive(self, fileslist):
...
         return root_dir
 
-    def generate_coredump_file(self, restart_scylla=True):
-        self.log.info('Generate scylla core')
+    def generate_coredump_file(self, restart_scylla=True, node_name: str = None, message: str = None):
+        message = message or "Generating a Scylla core dump by SIGQUIT..\n"
+        if node_name:
+            message += f"Node: {node_name}"
+        InfoEvent(severity=Severity.ERROR, message=message)
         self.remoter.run("sudo pkill -f --signal 3 /usr/bin/scylla")
```
fgelcer, line 2456: if you are already changing, please us `self.remoter.sudo("pkill -f --signal 3 /usr/bin/scylla")` instead

fgelcer (changes requested):
just a small change required

sdcm/cluster.py
yarongilor: fixed

fgelcer (approved):
LGTM

defaults/severities.yaml
```diff
@@ -95,7 +95,7 @@ TestFrameworkEvent: CRITICAL
 ElasticsearchEvent: ERROR
 SpotTerminationEvent: CRITICAL
 ScyllaRepoEvent: WARNING
-InfoEvent: NORMAL
+InfoEvent: ERROR
```
roydahan, line 98: Why is this needed?

fruch (approved):
LGTM

defaults/severities.yaml
fruch: Since it's using this event with error severity

This file set the maximum severity for each type. (Not that I remember why we need it, i.e. one more way of filtering event I guess)

roydahan: The maximum severity?
Where do we use it?

Issue #5413, opened 2022-11-14
c-s thread freezes during `disrupt_no_corrupt_repair` nemesis


ilya-rarov:
## Installation details

Kernel Version: 5.15.0-1021-aws
Scylla version (or git commit hash): `2022.2.0~rc3-20221009.994a5f0fbb4c` with build-id `756ea8d62c25ed4acdf087054e11b3d07596a117`
Relocatable Package: http://downloads.scylladb.com/unstable/scylla-enterprise/enterprise-2022.2/relocatable/2022-10-09T10:39:11Z/scylla-enterprise-x86_64-package.tar.gz
Cluster size: 4 nodes (i3.4xlarge)

Scylla Nodes used in this run:
- longevity-tls-1tb-7d-2022-2-db-node-ebfcbd22-7 (18.202.221.40 | 10.4.1.184) (shards: 14)
- longevity-tls-1tb-7d-2022-2-db-node-ebfcbd22-6 (54.75.36.52 | 10.4.2.135) (shards: 14)
- longevity-tls-1tb-7d-2022-2-db-node-ebfcbd22-5 (52.212.53.52 | 10.4.0.135) (shards: 14)
- longevity-tls-1tb-7d-2022-2-db-node-ebfcbd22-4 (18.202.235.185 | 10.4.0.49) (shards: 14)
- longevity-tls-1tb-7d-2022-2-db-node-ebfcbd22-3 (52.18.57.81 | 10.4.1.100) (shards: 14)
- longevity-tls-1tb-7d-2022-2-db-node-ebfcbd22-2 (34.243.172.17 | 10.4.3.250) (shards: 14)
- longevity-tls-1tb-7d-2022-2-db-node-ebfcbd22-1 (54.171.228.251 | 10.4.3.237) (shards: 14)

OS / Image: `ami-0b9c9dd9d3af4cec6` (aws: eu-west-1)

Test: `longevity-1tb-7days-test`
Test id: `ebfcbd22-452c-4d2d-b0cd-6b5944e952ca`
Test name: `enterprise-2022.2/longevity/longevity-1tb-7days-test`
Test config file(s):

- [longevity-1TB-5days-authorization-and-tls-ssl.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/1b668169316e7b75f2db649c8df5b8ef1f8f2805/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml)

## Issue description
[...]

PR #5422, opened 2022-11-15
fix(disrupt_no_corrupt_repair): wait for schema agreement

ilya-rarov:
C-S thread creating keyspaces triggered in `_prepare_test_table` may fail to write data and freeze if the nodes do not get to schema agreement quickly enough (due to Scylla's problem with schema management).
To avoid it we will wait for schema agreement each time after calling `_prepare_test_table`.

This workaround should fix https://github.com/scylladb/scylla-cluster-tests/issues/5413

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1442,6 +1442,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         for i in range(10):
             self.log.debug('Prepare test tables if they do not exist')
             self._prepare_test_table(ks=f'drop_table_during_repair_ks_{i}', table='standard1')
+            self.cluster.wait_for_schema_agreement()
 
         self.log.debug("Start repair target_node in background")
         with ThreadPoolExecutor(max_workers=1, thread_name_prefix='NodeToolRepairThread') as thread_pool:


sdcm/nemesis.py
```diff
@@ -1434,8 +1434,21 @@ def disrupt_kill_scylla(self):
     def disrupt_no_corrupt_repair(self):
         # prepare test tables and fill test data
         for i in range(10):
+            ks_name = f"drop_table_during_repair_ks_{i}"
+            test_keyspaces = self.cluster.get_test_keyspaces()
```
fruch, line 1438: I would rather have this call outside of the loop.

In cases we have lots of keyspaces it could be quite slow and problematic to call it so many times

fruch: Also did you considered add this option of precreate the keyspaces into `_prepare_test_table`

fruch: And also this change only create keyspace, but creation of the
 table is also schema change we might want also to have  the table created in advance

sdcm/nemesis.py
```diff
@@ -1434,8 +1434,21 @@ def disrupt_kill_scylla(self):
...
+
+            # If a keyspace doesn't exist, create it
+            if ks_name not in test_keyspaces:
+                self.log.debug("Going to create a new keyspace `%s'...", ks_name)
+                self.tester.create_keyspace(
```
fruch, line 1443: Check the default timeout inside this implemention, we might want to enlarge it to some degree,
So we won't timeout on those calls

roydahan: Why not just run "CREATE KEYSPACE IF NOT EXIST" ?

sdcm/nemesis.py
```diff
@@ -1434,8 +1434,21 @@ def disrupt_kill_scylla(self):
...
+                    keyspace_name=ks_name,
+                    replication_factor=self.tester.reliable_replication_factor,
+                )
+
             self.log.debug('Prepare test tables if they do not exist')
-            self._prepare_test_table(ks=f'drop_table_during_repair_ks_{i}', table='standard1')
+            self._prepare_test_table(ks=ks_name, table='standard1')
+
+            time.sleep(60)
```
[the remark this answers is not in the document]
roydahan: Probably in both cases, or "wait for schema agreement" if we have such method.

ilya-rarov:
> LGTM, but @ilya-rarov , what is the status with it?

@fgelcer I created a draft PR because I triggered a [reproducer](https://jenkins.scylladb.com/view/enterprise-2022.2/job/enterprise-2022.2/job/longevity/job/Reproducers/job/longevity-1tb-7days-test/2/) and waited for it to finish.

sdcm/nemesis.py
fgelcer: > Check the default timeout inside this implementation, we might want to enlarge it to some degree, So we won't timeout on those calls
i don't see any timeout set in that function, so it should be implicit somewhere...

> Why not just run "CREATE KEYSPACE IF NOT EXIST" ?
this is exactly what it is doing underneath, but we still need to create the tables

fgelcer:
> > LGTM, but @ilya-rarov , what is the status with it?
> 
> @fgelcer I created a draft PR because I triggered a [reproducer](https://jenkins.scylladb.com/view/enterprise-2022.2/job/enterprise-2022.2/job/longevity/job/Reproducers/job/longevity-1tb-7days-test/2/) and waited for it to finish.

@ilya-rarov , that run you mention above (build 2) failed after 14h... why is that? also, build 3 failed after some 40 seconds.. most likely setup failed?

sdcm/nemesis.py
ilya-rarov: @fruch  Initially, I was thinking about changing `_prepare_test_table`, but it is used by several nemeses, not only by `disrupt_no_corrupt_repair`, so after a short discussion with @roydahan I made the change only in `disrupt_no_corrupt_repair`.
If we also need to create the tables in advance for `disrupt_no_corrupt_repair` then I guess I have no better option than change `_prepare_test_table`. Following your advice, I could introduce a new flag argument for `_prepare_test_table` and create objects in advance there only if it is `True`.

fruch: that's what I was point to, to extend the API of that call.

sdcm/nemesis.py
fruch: > > Check the default timeout inside this implementation, we might want to enlarge it to some degree, So we won't timeout on those calls
> i don't see any timeout set in that function, so it should be implicit somewhere...
> 
> > Why not just run "CREATE KEYSPACE IF NOT EXIST" ?
> this is exactly what it is doing underneath, but we still need to create the tables

we could do  `IF NOT EXIST` for tables as well (as I've suggested, we should create the tables)
and for the little amount of data we put in, we might as well generate the data with the python code, and skip c-s.

using c-s in this case, was a quick way to generate new ks/table with small amount of data.
we can always to switch to a different method/tool, i.e. the tool in use isn't the important part (and we don't care about speed/performance in those)

roydahan: 1. If we do "IF NOT EXIST" we don't need to query all tables and check that the table isn't there.

2. I don't mind to write the data without c-s, as long as it's quick enough.
In this case, since I don't want to start a saga of testing, the fix should be very simple.

For loop with create keyspace and create table, nothing else.

sdcm/nemesis.py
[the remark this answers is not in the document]
ilya-rarov: Added call of `wait_for_schema_agreement()` between the creation of objects and the start of the c-s thread and also between the keyspace loop iterations

sdcm/nemesis.py
ilya-rarov: Removed ks and table creation from the nemesis and added a new argument for `_prepare_test_table` method - `pre_create_objects` which is `True` only while calling the method from this nemesis (and `False` by default).

sdcm/nemesis.py
ilya-rarov: Moved ks and table creation to `_prepare_test_table` method

sdcm/nemesis.py
```diff
@@ -1725,13 +1727,44 @@ def disrupt_nodetool_cleanup(self):
             for keyspace in test_keyspaces:
                 node.run_nodetool(sub_cmd="cleanup", args=keyspace)
 
-    def _prepare_test_table(self, ks='keyspace1', table=None):
+    def _prepare_test_table(self,
+                            ks: str = 'keyspace1',
+                            table: str | None = None,
+                            pre_create_objects: bool = False) -> None:
         ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=self.target_node)
         table_exist = f'{ks}.{table}' in ks_cfs if table else True
```
fgelcer, line 1735: we can drop the following lines:
```
ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=self.target_node)
        table_exist = f'{ks}.{table}' in ks_cfs if table else True
```

it because both create keyspace and create table they use `IF NOT EXISTS`, so in this case we can get drop querying, because it may slow the whole prepare/create part

sdcm/nemesis.py
```diff
@@ -1725,13 +1727,44 @@ def disrupt_nodetool_cleanup(self):
...
+            if pre_create_objects:
+                self.log.debug("Going to create a new keyspace `%s'...", ks)
+                self.tester.create_keyspace(
+                    keyspace_name=ks,
+                    replication_factor=self.tester.reliable_replication_factor,
+                )
+
+                self.cluster.wait_for_schema_agreement()
+
+                table_name = table or "standard1"
```
fgelcer, line 1751: instead of having it here, we could have this default set in the function signature

sdcm/nemesis.py
```diff
@@ -1435,7 +1435,9 @@ def disrupt_no_corrupt_repair(self):
         # prepare test tables and fill test data
         for i in range(10):
             self.log.debug('Prepare test tables if they do not exist')
-            self._prepare_test_table(ks=f'drop_table_during_repair_ks_{i}', table='standard1')
+            self._prepare_test_table(ks=f'drop_table_during_repair_ks_{i}', table='standard1', pre_create_objects=True)
+
+            self.cluster.wait_for_schema_agreement()
```
fgelcer, line 1440: you can drop this, as you are already doing it inside `_prepare_test_table`

fgelcer (dismissed):

sdcm/nemesis.py
ilya-rarov: Don't we need to wait for schema agreement after we processed the previous keyspace (populated the table with data) and before we are going to create a new one?

sdcm/nemesis.py
ilya-rarov: Yes, I understand what `IF NOT EXISTS` does, but the change you suggest affects the logic of `_prepare_test_table`.
Now we do not start the C-S thread if both the keyspace and the table exist.
If I remove these lines the C-S will be triggered each time, regardless of the objects' presence. If the objects exist it will at least try to write data to them each time.
I am not sure how It affects all the nemeses using  `_prepare_test_table` - this is not only `NoCorruptRepairMonkey`, but also `SwitchBetweenPasswordAuthAndSaslauthdAuth`, `LoadAndStreamMonkey`, `RefreshBigMonkey`, `RefreshMonkey`, `TruncateMonkey`.
What do you think?

sdcm/nemesis.py
fgelcer: you call `wait_for_schema_agreement` already twice inside `_prepare_test_table`
so you don't have to call it again in here, but confirm that in the flow in `_prepare_test_table`

sdcm/nemesis.py
ilya-rarov: Yes, I could, but in this case, I would need to change how this method is called in `disrupt_truncate`.
We have there: `self._prepare_test_table(ks=keyspace_truncate)`, I would need to change it to `self._prepare_test_table(ks=keyspace_truncate, table=None)` in order to not affect anything.
Should I do this?

fgelcer: but  this is exactly as having it in the method signature... if `table` will be `None`, then it will be assigned `standard1`, but if you pass anything else, then it will be `table`...
if the function call will be `_prepare_test_table(ks: str = 'keyspace1', table: str = 'standard1' ....)`

sdcm/nemesis.py
fgelcer: if you create the keyspace and table using `IF NOT EXISTS`, then there is no need to query if they exist before, and if you populate the tables using CQL command through Python, then you also don't need to use `c-s` for it... 
all these nemeses will most likely want to have content in their tables, but not sure they need something specific in the keyspace or table...

ilya-rarov: @fgelcer just to be on the same page with you.
Let's say we already have the keyspace and table before calling `_prepare_test_table`
In this case, the code
```python
        ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=self.target_node)
        table_exist = f'{ks}.{table}' in ks_cfs if table else True

        test_keyspaces = self.cluster.get_test_keyspaces()
        # If keyspace or table doesn't exist, create it
        if ks not in test_keyspaces or not table_exist:
            ...
```
prevents populating this table with data, right?
If I remove this code (because now I am relying on `IF NOT EXISTS` and don't need extra querying), then I will populate the table no matter whether it existed before (empty or not) or has just been created.
It looks like a change of `_prepare_test_table` algorithm that may affect other nemeses, doesn't it?

Sorry for asking dumb questions, but now I am just trying to understand which way to proceed.
From my understanding, initially, I was supposed to create a simple workaround for `disrupt_no_corrupt_repair` nemesis.
But now it looks like I am about to significantly change `_prepare_test_table`. After this change `_prepare_test_table` will behave a bit differently, which means this workaround impacts all the nemeses using it. In my opinion, this impact needs to be tested.
I can do that, no problem.
I just want to clarify the goal I need to reach working on this fix.

sdcm/nemesis.py
```diff
@@ -1731,24 +1732,54 @@ def disrupt_nodetool_cleanup(self):
...
+            session.default_consistency_level = ConsistencyLevel.QUORUM
+
+            query = f"INSERT INTO {ks}.{table} " \
+                    f"(key,{','.join(columns)}) " \
+                    f"VALUES ({('?,' * (number_of_columns + 1)).rstrip(',')});"
+            insert_query = session.prepare(query=query)
+
+            for _ in range(number_of_rows):
+                values = [uuid.uuid1().hex for _ in range(number_of_columns + 1)]
+                session.execute(query=insert_query, parameters=values)
```
fruch, line 1776: you might want to use `execute_concurrent_with_args`
https://github.com/scylladb/python-driver/blob/master/cassandra/concurrent.py#L231

should work a bit quicker.

fgelcer: that would be great

roydahan:
I wanted a quick for no_corrupt_repair.
How the hell we ended up with refactoring _prepare_test_table?

If we refactor this (which I don't think is necessary, unless we want to hide all the schema issues of schema agreement), we need to test ALL nemesis that use this function.
e.g. "load_and_stream", "nodetool_refresh".

roydahan:
I think that even doing this would be good enough:

diff --git a/sdcm/nemesis.py b/sdcm/nemesis.py
index 56ef1c72..041b3938 100644
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1417,6 +1417,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         for i in range(10):
             self.log.debug('Prepare test tables if they do not exist')
             self._prepare_test_table(ks=f'drop_table_during_repair_ks_{i}', table='standard1')
+            self.cluster.wait_for_schema_agreement()

ilya-rarov:
> I think that even doing this would be good enough:

I see the problem happens inside one call of  `_prepare_test_table` (in particular, it is related to C-S thread). So adding a delay between the calls of `_prepare_test_table` won't help, IMHO.

sdcm/nemesis.py
ilya-rarov: As agreed, the only thing we need in this workaround is to wait for the schema agreement in `disrupt_no_corrupt_repair`.
The proposed change is not needed.

sdcm/nemesis.py
ilya-rarov: As agreed, the only thing we need in this workaround is to wait for the schema agreement in `disrupt_no_corrupt_repair`.
The proposed change is not needed.

sdcm/nemesis.py
ilya-rarov: As agreed, the only thing we need in this workaround is to wait for the schema agreement in `disrupt_no_corrupt_repair`.
I kept it here, but removed it from `_prepare_test_table`.

sdcm/nemesis.py
ilya-rarov: As agreed, the only thing we need in this workaround is to wait for the schema agreement in `disrupt_no_corrupt_repair`.
The proposed change is not needed.

fruch:
> > I think that even doing this would be good enough:
> 
> I see the problem happens inside one call of `_prepare_test_table` (in particular, it is related to C-S thread). So adding a delay between the calls of `_prepare_test_table` won't help, IMHO.

It might help if there a schema agreement, before the c-s starts, the schema change c-s should be faster
anyhow we can't know if it would help or not, since we don't really have reproducer of the issue.

Anyhow seems like we might have similar thing happening quite regularly on the azure test (i.e. test we probably never stabilized performance wise) 
https://70f106c98484448dbc4705050eb3f7e9.us-east-1.aws.found.io:9243/goto/b934f8eea89935b0474aa161736f35eb

fruch (approved):
It's basically a guess, but seem like it would hurt... so I'm o.k. with it.

fgelcer:
@ilya-rarov , maybe now we want to re-run the failed job on `2022.2` to be see it not failing in `disrupt_no_corrupt_repair`

PR #5424, opened 2022-11-16
fix(sct_events/setup.py): Filter LDAP connection resets

k0machi:
Per scylladb/scylladb#11969 this error is considered harmless as it happens during shutdown and doesn't affect the user

### DIFF
--- a/sdcm/sct_events/setup.py
+++ b/sdcm/sct_events/setup.py
@@ -64,6 +64,10 @@ def start_events_device(log_dir: Optional[Union[str, Path]] = None,
     EventsSeverityChangerFilter(new_severity=Severity.WARNING,
                                 event_class=DatabaseLogEvent.DATABASE_ERROR,
                                 regex='cdc - Could not update CDC description table with generation').publish()
+    EventsSeverityChangerFilter(new_severity=Severity.WARNING,
+                                event_class=DatabaseLogEvent.DATABASE_ERROR,
+                                regex='ldap_connection - Seastar read failed: std::system_error '
+                                '(error system:104, read: Connection reset by peer)').publish()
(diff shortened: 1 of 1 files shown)


sdcm/sct_events/setup.py
```diff
@@ -67,6 +67,8 @@ def start_events_device(log_dir: Optional[Union[str, Path]] = None,
     DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line='Rate-limit: supressed').publish()
     DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line='Rate-limit: suppressed').publish()
     DbEventsFilter(db_event=DatabaseLogEvent.WARNING, line='abort_requested_exception').publish()
+    DbEventsFilter(db_event=DatabaseLogEvent.DATABASE_ERROR, line='ldap_connection - Seastar read failed: '
```
fgelcer, line 70: it could be a signal of an issue, when seeing these messages without any nemesis, but we should IMHO filter it while during nemesis.

k0machi: The problem is that it happens during any nemesis that restart/shuts down a node, which could be problematic to keep track of.

fgelcer: it is not a very widely used test case, nor i'm sure about the impact of this error (maybe users won't be able to create new connections?), but IIRC it started happening because of a bogus nemesis...
i would suggest to not filter it out completely, but maybe lower its priority to be `WARNING` instead of completely ignoring it... WDYT?

k0machi: Hmm, good idea, will lower it to a `WARNING` instead.

k0machi: Done

fgelcer (approved):
LGTM

fruch (approved):
LGTM

PR #5430, opened 2022-11-16
feature(k8s-functional): population context for pytest.mark.skipif

fruch:
add context parameters, into `pytest.mark.skipif` condition, so we can use versions, or any information from within params to be able to skip tests (with not need of more extra fixtures for that)

### DIFF
--- a/functional_tests/scylla_operator/conftest.py
+++ b/functional_tests/scylla_operator/conftest.py
@@ -26,6 +26,10 @@ from packaging import version
 from functional_tests.scylla_operator.libs.auxiliary import ScyllaOperatorFunctionalClusterTester, sct_abs_path
 from sdcm.cluster_k8s import ScyllaPodCluster
 from sdcm.utils import version_utils
+from sdcm.test_config import TestConfig
+from sdcm.cluster_k8s.mini_k8s import LocalKindCluster
+from sdcm.localhost import LocalHost
+from sdcm.tester import ClusterTester
 
 
 TESTER: Optional[ScyllaOperatorFunctionalClusterTester] = None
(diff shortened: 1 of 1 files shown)


vponomaryov (commented):
It looks like combination of lots of existing fixtures in one place.
Don't see benefit doing so.
Also, it won't cover case of all the fixtures we have.
For example "skip_if_scylla_not_semver".

Also, it tries to init a cluster object anyway, and it will be done anyway as part of the existing fixtures such as "tester".

So, personally, I don't think it makes sense to have this kind of special "fixture".

fruch:
@soyacz what do you think on this one ?

I liked the idea that the usage of it is clear, and you can match it to your needs, instead of writing specific fixtures to each case you need to skip.

soyacz:
> @soyacz what do you think on this one ?
> 
> I liked the idea that the usage of it is clear, and you can match it to your needs, instead of writing specific fixtures to each case you need to skip.

I like flexibility and not needing to write specific fixtures all the time.

I just wonder, why making this so complicated? We could get `operator_version` from config - in case of latest, return some arbitrary big number (e.g. 100.0.0). I think when we use latest, we don't care about specific version - it's anyway greater than anything we know, so it should return `True` to any `>` test and return `False` for any `<` test. Correct me if you know other case.
As to params object, we can get it straight from `init_and_verify_sct_config` function.
We'll be able to drop most of the logic then.

functional_tests/scylla_operator/conftest.py
```diff
@@ -229,3 +233,35 @@ def skip_based_on_operator_version(request: pytest.FixtureRequest, tester: Scyll
...
+            self._move_kubectl_config()
+            self.localhost = LocalHost(user_prefix=self.params.get("user_prefix"), test_id=test_config.test_id())
+
+    obj = HelmOnly()
+
+    kluster = LocalKindCluster('', user_prefix=obj.params.get("user_prefix"), params=obj.params)
+    operator_version = kluster._scylla_operator_chart_version
+
+    return dict(
+        scylla_version=version.LegacyVersion(obj.params.get('scylla_version')),
```
soyacz, line 263: what it returns for `latest`  `scylla_version`?

fruch:
I'd rather not introduce those "big" we never gonna get to magic numbers.

There's a version under test, it should have a version and we should be able to retire without spinning up resources (i.e. we don't need a full running k8s cluster to get the version of software we are working with.

I agree it is currently complicated since the logic of getting the version is coupled tightly with the k8s cluster implementation.

vponomaryov:
> > @soyacz what do you think on this one ?
> > I liked the idea that the usage of it is clear, and you can match it to your needs, instead of writing specific fixtures to each case you need to skip.
> 
> I like flexibility and not needing to write specific fixtures all the time. Also telling the reason next is clearer.
> 
> I just wonder, why making this so complicated? We could get `operator_version` from config - in case of latest, return some arbitrary big number (e.g. 100.0.0). I think when we use latest, we don't care about specific version - it's anyway greater than anything we know, so it should return `True` to any `>` test and return `False` for any `<` test. Correct me if you know other case. As to params object, we can get it straight from `init_and_verify_sct_config` function. We'll be able to drop most of the logic then.

`Latest` tag will refer to completely different versions depending on the operator helm chart repo we use.
So, currently, the `latest` operator tag in `stable` repo is `v1.7.4` which is pretty old already and won't work on the latest K8S versions.

PR #5434, opened 2022-11-18
fix(cluster.py): properly parse cpuset node info

vponomaryov:
For the moment BaseNode's property called `cpuset` only supports one or two ranges of cores.
It will fail with IndexError if only one core is used. 
It will show wrong number of cores having more than 2 ranges (17+ cores instance type).

So, make it parse the possible values correctly and add unit tests coverage for it.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -528,33 +528,41 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
 
     @property
     def cpuset(self):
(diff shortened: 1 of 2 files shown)


vponomaryov:
It is needed for the K8S multitenant setup where we may configure just 1 core per Scylla pod/node or another combination of cores such as "1-7,9" or "1,9".

It may be backported to other branches but knowing that it is urgent only for K8S then I don't see any need to backport it.

fruch (approved):
LGTM

PR #5436, opened 2022-11-20
fix(docker): fix `reuse_cluster` to find instance

fruch:
code was looking for docker instance by prefix, but the api need to have `*` at the end for it to actully find any instances

### DIFF
--- a/sdcm/cluster_docker.py
+++ b/sdcm/cluster_docker.py
@@ -251,7 +251,7 @@ class DockerCluster(cluster.BaseCluster):  # pylint: disable=abstract-method
 
     def _get_nodes(self):
         containers = ContainerManager.get_containers_by_prefix(self.node_prefix)
-        for node_index, container in sorted((int(c.labels["NodeIndex"]), c) for c in containers):
+        for node_index, container in sorted(((int(c.labels["NodeIndex"]), c) for c in containers), key=lambda x: x[0]):
             LOGGER.debug("Found container %s with name `%s' and index=%d", container, container.name, node_index)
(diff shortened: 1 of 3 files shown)


sdcm/cluster_docker.py
```diff
@@ -250,8 +250,8 @@ def _create_nodes(self, count, enable_auto_bootstrap=False):
         return new_nodes
 
     def _get_nodes(self):
-        containers = ContainerManager.get_containers_by_prefix(self.node_prefix)
-        for node_index, container in sorted((int(c.labels["NodeIndex"]), c) for c in containers):
+        containers = ContainerManager.get_containers_by_prefix(f'{self.node_prefix}*')
```
fgelcer, line 253: shouldn't this be only the prefix, and the `get_containers_by_prefix` adds the `*` after the prefix??

fgelcer (approved):
not sure i have the full picture here, but if the function `get_containers_by_prefix` gets a `prefix`, it should not be followed by `*` IIUC

sdcm/cluster_docker.py
fgelcer: but approved, because it will work

fruch: yeah the `get_containers_by_prefix` is the actual place need to be fixed.
(but this is the only usage of it across the code, might as well change it to `get_containers_by_name`)

fgelcer: `get_containers_by_name` would be another function, that searches for specific and exact names

fgelcer (approved):
LGTM

PR #5443, opened 2022-11-21
test(limit): perf test for limit streaming

soyacz:
Added stream_io_throughput_mb_per_sec limit to 250gb perf test with nemesis to see the impact on latency and nemesis duration.

### DIFF
--- a/jenkins-pipelines/feature-limit-streaming-io.jenkinsfile
+++ b/jenkins-pipelines/feature-limit-streaming-io.jenkinsfile
@@ -1,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+perfRegressionParallelPipeline(
+    backend: "aws",
+    test_name: "performance_regression_test.PerformanceRegressionTest",
+    test_config: "test-cases/features/limit-streaming-io.yaml",
+    sub_tests: ["test_latency_write_with_nemesis", "test_latency_read_with_nemesis", "test_latency_mixed_with_nemesis"],
+
+    timeout: [time: 1600, unit: "MINUTES"]
+)
--- a/sdcm/provision/scylla_yaml/scylla_yaml.py
+++ b/sdcm/provision/scylla_yaml/scylla_yaml.py
@@ -172,6 +172,7 @@ class ScyllaYaml(BaseModel):  # pylint: disable=too-few-public-methods
         'org.apache.cassandra.scheduler.NoScheduler',
         'org.apache.cassandra.scheduler.RoundRobinScheduler'
     ] = "org.apache.cassandra.scheduler.NoScheduler"
+    stream_io_throughput_mb_per_sec: int = 0
 
     # pylint: disable=no-self-argument,no-self-use
     @validator("request_scheduler", pre=True, always=True)
--- a/test-cases/features/limit-streaming-io.yaml
+++ b/test-cases/features/limit-streaming-io.yaml
@@ -1,0 +1,44 @@
+# test for https://github.com/scylladb/scylladb/commit/1f21c1ecc8e90d2bced38c019b1de0b3f651ddf1
+# based on perf-regression-latency-250gb-with-nemesis
+# For full result, compare latency and time of cluster operations with performance test results.
+# Latency drop during cluster operations should be lower when streaming limit is set but as a drawback these operations will
+# take longer to complete.
+# Results from previous runs documented in: https://docs.google.com/document/d/1tCyYgwGWww0qVfu-lf80ytesw0Q1uYpgikPz2kEhamQ/edit#
+test_duration: 880
+prepare_write_cmd: ["cassandra-stress write no-warmup cl=ALL n=62500000 -schema 'replication(factor=3)' -port jmx=6868 -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=1..62500000",
+                    "cassandra-stress write no-warmup cl=ALL n=62500000 -schema 'replication(factor=3)' -port jmx=6868 -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=62500001..125000000",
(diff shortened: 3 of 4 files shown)


fruch (approved):
LGTM

fgelcer (approved):
LGTM

soyacz:
@fruch @fgelcer  I'm not sure if you are aware - this is a push to perf-v12 branch - let me know if I should follow master->backport to perf-v12

fruch:
> @fruch @fgelcer  I'm not sure if you are aware - this is a push to perf-v12 branch - let me know if I should follow master->backport to perf-v12

Yes it should go via master

fgelcer:
> > @fruch @fgelcer  I'm not sure if you are aware - this is a push to perf-v12 branch - let me know if I should follow master->backport to perf-v12
> 
> Yes it should go via master

also, IMHO we shouldn't backport this kind of feature to `branch-perf-v12`
if you need it, you can use `branch-perf-v14` (if already created, it should be), or until it is created, use master

roydahan:
1. Push it to master, not perf-v12.
2. Move it out from performance test and add it as a feature test.
We're not testing here a performance regression and comparing it across versions.
3. We may want to test in the same test, different limits.
e.g. limit to 250, run a while, change to 500, run for a while, increase, run...

roydahan (changes requested):
See my comment above.

soyacz:
> 1. Push it to master, not perf-v12.

done

> 2. Move it out from performance test and add it as a feature test.
>    We're not testing here a performance regression and comparing it across versions.

done

> 3. We may want to test in the same test, different limits.
>    e.g. limit to 250, run a while, change to 500, run for a while, increase, run...

That is not trivial - I'd need to modify the test to update configs in the middle, maybe adding more nemesis cycles for comparison. And still it's not clear for me how this test should look like then.
Shall I invest time in it?
I think easier is to just run test with different settings multiple times.

fruch:
> > 1. Push it to master, not perf-v12.
> 
> done
> 
> > 2. Move it out from performance test and add it as a feature test.
> >    We're not testing here a performance regression and comparing it across versions.
> 
> done
> 
> > 3. We may want to test in the same test, different limits.
> >    e.g. limit to 250, run a while, change to 500, run for a while, increase, run...
> 
> That is not trivial - I'd need to modify the test to update configs in the middle, maybe adding more nemesis cycles for comparison. And still it's not clear for me how this test should look like then.
> Shall I invest time in it?
> I think easier is to just run test with different settings multiple times.
> 

Since for other feature of limit, seems like you'll need the extract out the nemesis running out of this performance test.

Should be something roughly like this:

* start stress
* configure limit
* start nemesis thread with limiting it to one round 
* Wait for nemesis thread to end
* configure new limit
* start nemesis thread with limiting it to one round 
* wait for stress to end

soyacz:
> Should be something roughly like this:
> 
> * start stress
> * configure limit
> * start nemesis thread with limiting it to one round
> * Wait for nemesis thread to end
> * configure new limit
> * start nemesis thread with limiting it to one round
> * wait for stress to end

I see one problem with that - we don't know how long it will take, so stress thread may be too long.
Maybe start/stop stress thread for each limit and get the results?

fgelcer:
> I see one problem with that - we don't know how long it will take, so stress thread may be too long. Maybe start/stop stress thread for each limit and get the results?

we (you) took care of it in the perf with nemesis jobs... "gracefully" stopping stress command when the nemesis cycle was done

fruch:
> > I see one problem with that - we don't know how long it will take, so stress thread may be too long. Maybe start/stop stress thread for each limit and get the results?
> 
> we (you) took care of it in the perf with nemesis jobs... "gracefully" stopping stress command when the nemesis cycle was done

It good enough if you build ontop getting the results out of monitor, and not from stress command summary, you can't kill the command and get the summary.

soyacz:
> > 
> > we (you) took care of it in the perf with nemesis jobs... "gracefully" stopping stress command when the nemesis cycle was done
> 
> It good enough if you build ontop getting the results out of monitor, and not from stress command summary, you can't kill the command and get the summary.

I found `latency_calculator_decorator` which does the job already by taking latency from monitor. That's why we can kill stress command without loosing results (anyway we're interested in latency during specific nemesis and stress command spans across multiple nemesis - so we don't care much about c-s results).

But `latency_calculator_decorator` collects info only for latency - is it worth to develop something similar for throughput for this test?

fruch:
> 
> > > 
> > > we (you) took care of it in the perf with nemesis jobs... "gracefully" stopping stress command when the nemesis cycle was done
> > 
> > It good enough if you build ontop getting the results out of monitor, and not from stress command summary, you can't kill the command and get the summary.
> 
> I found `latency_calculator_decorator` which does the job already by taking latency from monitor. That's why we can kill stress command without loosing results (anyway we're interested in latency during specific nemesis and stress command spans across multiple nemesis - so we don't care much about c-s results).
> 
> But `latency_calculator_decorator` collects info only for latency - is it worth to develop something similar for throughput for this test?

Yes, let's see how we can share the implemention and make it easy to track any other metrics

soyacz:
> 1. Push it to master, not perf-v12.

done

> 2. Move it out from performance test and add it as a feature test.
>    We're not testing here a performance regression and comparing it across versions.

done

> 3. We may want to test in the same test, different limits.
>    e.g. limit to 250, run a while, change to 500, run for a while, increase, run...

This could be done in separate task. @roydahan if you think it is worth the time please create one and assign.


@fgelcer please review and approve if you find it ok.

fgelcer (dismissed):
LGTM

fruch:
@soyacz please fix the unit-test `ScyllaYamlTest::test_scylla_yaml`

soyacz:
> @soyacz please fix the unit-test `ScyllaYamlTest::test_scylla_yaml`

fixed

fgelcer (approved):
LGTM

fruch:
> > 1. Push it to master, not perf-v12.
> 
> done
> 
> > 2. Move it out from performance test and add it as a feature test.
> >    We're not testing here a performance regression and comparing it across versions.
> 
> done
> 
> > 3. We may want to test in the same test, different limits.
> >    e.g. limit to 250, run a while, change to 500, run for a while, increase, run...
> 
> This could be done in separate task. @roydahan if you think it is worth the time please create one and assign.

@soyacz please create the follow up task with all the details, and we'll merge this

> 
> 
> @fgelcer please review and approve if you find it ok.

soyacz:
> @soyacz please create the follow up task with all the details, and we'll merge this

followup task:
- https://github.com/scylladb/qa-tasks/issues/950

soyacz:
@fruch if there's nothing more to do, let's merge and close it.

fruch (approved):
LGTM

PR #5444, opened 2022-11-21
fix(upgrade): increase load time during upgrade/rollback

juliayakovlev:
Increase load time during upgrade/rollback for rolling upgrade SLA test

### DIFF
--- a/configurations/rolling-upgrade-with-sla.yaml
+++ b/configurations/rolling-upgrade-with-sla.yaml
@@ -1,7 +1,7 @@
 # workloads
 stress_before_upgrade: "cassandra-stress write cl=QUORUM n=60000000 -schema 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=1000 -pop seq=1..60000000 -col 'n=FIXED(10) size=FIXED(512)' -log interval=5"
-stress_during_entire_upgrade: [ "cassandra-stress read cl=QUORUM duration=30m -schema 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native <sla credentials 0> -rate threads=200 -pop seq=1..30000000 -col 'n=FIXED(10) size=FIXED(512)' -log interval=5",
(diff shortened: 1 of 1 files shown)


configurations/rolling-upgrade-with-sla.yaml
```diff
@@ -11,4 +11,9 @@ upgrade_sstables: false
 n_loaders: 2
 round_robin: true
 
+user_prefix: 'rolling-upgrade-sla'
+
+authorization_in_upgrade: ""
```
fruch, line 16: Let's have a PR to remove those "features"..

I think @roydahan would agree that we don't need that anymore

fruch (dismissed):

configurations/rolling-upgrade-with-sla.yaml
roydahan: What does it mean?

juliayakovlev:
@roydahan @fruch 
Do we plan to run upgrade test with user on 5.1 branch? 
I ask this because service level in versions 5.0/5.1 does not support "shares" parameter, so I need to add support in the code for it

roydahan:
We can add roles and users to the existing rolling upgrades tests we have.
Regardless of SLA.

Would be good to test it for 5.1, but I don’t expect any issue there (I hope).

configurations/rolling-upgrade-with-sla.yaml
fruch: it's a time bomb amos left us, from the days we started enabling auth in scylla :)

we are breaking authorization in rollbacks, but we just first noticed it

fruch:
> We can add roles and users to the existing rolling upgrades tests we have. Regardless of SLA.
> 
> Would be good to test it for 5.1, but I don’t expect any issue there (I hope).

@juliayakovlev create a task for it, and me and @roydahan would prioritize it

juliayakovlev:
> > We can add roles and users to the existing rolling upgrades tests we have. Regardless of SLA.
> > Would be good to test it for 5.1, but I don’t expect any issue there (I hope).
> 
> @juliayakovlev create a task for it, and me and @roydahan would prioritize it

https://github.com/scylladb/qa-tasks/issues/899

fruch (approved):
LGTM

PR #5450, opened 2022-11-22
fix(s-b): unpin scylla bench version in tests

soyacz:
It looks that pinning old scylla-bench version is not bringing benefits in large partition tests. Newest version was found passing so we no longer need to pin it in tests.

### DIFF
--- a/test-cases/longevity/longevity-counters-multidc.yaml
+++ b/test-cases/longevity/longevity-counters-multidc.yaml
@@ -22,7 +22,3 @@ server_encrypt: true
 internode_encryption: 'dc'
 
 use_legacy_cluster_init: false
-
-# Temporarily downgrade scylla_bench to a stable version
-stress_image:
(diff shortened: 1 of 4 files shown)


yarongilor (commented):
LGTM, i just recall we had some s-b issues in newer s-b versions with the counters-multi-dc test.
we can consider verifying they are not reproduced:

https://github.com/scylladb/scylla-bench/issues/92
https://github.com/scylladb/scylla-bench/issues/89

roydahan (approved):

soyacz:
@roydahan do we backport it to 2022.2/5.1?

roydahan:
It's a good question.
On one hand, we finished the cycle, and it would be good to keep it with the same versions we tested with.
OTOH, the reason we wanted to change is because of failures we had during the runs.

I tend to backport only to 2022.2, and possibly run 2-3 of the jobs we changed here.

PR #5452, opened 2022-11-23
feature(k8s-ingress): tag load ingress balancers

fruch:
since we are now creating the new AWS resources
we should tag them correctly for cleanups and billing

### DIFF
--- a/sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.svc.yaml
+++ b/sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.svc.yaml
@@ -3,6 +3,11 @@ kind: Service
 metadata:
   name: haproxy-kubernetes-ingress
   namespace: haproxy-controller
+  annotations:
+    # AWS based labeling base on those instructions:
+    # https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/annotations/#aws-resource-tags
+    service.beta.kubernetes.io/aws-load-balancer-additional-resource-tags: $TAGS
+
 spec:
   selector:
     app.kubernetes.io/name: haproxy-ingress
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -545,6 +545,7 @@ def clean_cloud_resources(tags_dict, dry_run=False):
     clean_instances_aws(tags_dict, dry_run=dry_run)
     clean_elastic_ips_aws(tags_dict, dry_run=dry_run)
     clean_test_security_groups(tags_dict, dry_run=dry_run)
+    clean_load_balancers_aws(tags_dict, dry_run=dry_run)
     clean_clusters_gke(tags_dict, dry_run=dry_run)
     clean_orphaned_gke_disks(dry_run=dry_run)
     clean_clusters_eks(tags_dict, dry_run=dry_run)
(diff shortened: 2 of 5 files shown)


fruch:
Still in draft since we need to setup permission for using the tagging api (works for my user, but not for all of dev users)
```
01:08:52  botocore.exceptions.ClientError: An error occurred (AccessDeniedException) when calling the GetResources operation: User: arn:aws:iam::***:user/** is not authorized to perform: tag:GetResources because no identity-based policy allows the tag:GetResources action
```

sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.svc.yaml
```diff
@@ -3,6 +3,9 @@ kind: Service
 metadata:
   name: haproxy-kubernetes-ingress
   namespace: haproxy-controller
+  annotations:
```
vponomaryov, line 6: Adding the link to the docs you mention in the task for this PR would be very useful:

https://kubernetes-sigs.github.io/aws-load-balancer-controller/v2.4/guide/service/annotations/#aws-resource-tags

sdcm/utils/common.py
```diff
@@ -911,6 +912,81 @@ def clean_test_security_groups(tags_dict, dry_run=False):
...
+    :param dry_run: if True, wouldn't delete any resource
+    :return: None
+    """
+
+    # don't if `post_behavior_k8s_cluster` was set to not destroy
+    if "NodeType" in tags_dict and tags_dict.get("NodeType") != "k8s":
+        return
+
+    assert tags_dict, "tags_dict not provided (can't clean all instances)"
+    aws_instances = list_load_balancers_aws(tags_dict=tags_dict, group_as_region=True)
```
vponomaryov, line 972: aws_instances?

sdcm/utils/common.py
```diff
@@ -911,6 +912,81 @@ def clean_test_security_groups(tags_dict, dry_run=False):
...
+            LOGGER.info("There are no ELBs to remove in AWS region %s", region)
+            continue
+        client = boto3.client('elb', region_name=region)
+        for elb in elb_list:
+            arn = elb.get('ResourceARN')
+            LOGGER.info("Going to delete '%s'", arn)
+            if not dry_run:
+                try:
+                    response = client.delete_load_balancer(LoadBalancerName=arn.split('/')[1])
+                    LOGGER.debug("Done. Result: %s\n", response)
```
vponomaryov, line 985: Maybe better to make all such calls be parallel/concurrent?
None of load balancers depends on any other.

fruch: since we have only one per test (for now, later we'll have one per region), it's doesn't make a lot of difference.

vponomaryov: How fast is it? If seconds, then ok.

fruch: we I didn't delete a lot of those, main one or two, and yest it was quick (it's not like deleting a node pool or eks cluster)

sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.svc.yaml
fruch: added

Issue #5278, opened 2022-09-15
all stress commands are by default retrying 3 times if the stress is timing out


fruch:
While testing the docker based loader (c-s) with 5000 tables use case, I've run into a stress command timing out (quite rare in regular situations, since the main load and the test duration are equal. 
in the 5000 tables, we have 55m user-profile stress, and we add 10min to it

I've seen the timeout of a specific command happen 3 time, in the same thread (i.e. with same test marker and same unique profile file) 
```
< t:2022-09-14 09:27:59,007 f:remote_libssh_cmd_runner.py l:66   c:RemoteLibSSH2CmdRunner p:ERROR > Command did not complete within 3900 seconds!
< t:2022-09-14 09:27:59,007 f:remote_libssh_cmd_runner.py l:66   c:RemoteLibSSH2CmdRunner p:ERROR > 
< t:2022-09-14 09:27:59,007 f:remote_libssh_cmd_runner.py l:66   c:RemoteLibSSH2CmdRunner p:ERROR > Command: 'sudo  docker exec c2132b228d144657b11f26c432c541232ab76712b21b4b6f7a6d808458dc77a9 /bin/sh -c \'echo TAG: loader_idx:0-cpu_idx:0-keyspace_idx:1; STRESS_TEST_MARKER=P27XL1TRVGASBD537G2W; cassandra-stress user profile=/tmp/templated_tables_mvmvtfs6ky.yaml cl=QUORUM \'"\'"\'ops(insert=1, read1=5)\'"\'"\' duration=55m -rate threads=2 -errors skip-unsupported-columns ignore -node 10.4.1.215,10.4.0.236,10.4.2.78,10.4.3.245,10.4.1.68,10.4.3.237\''
```

after drilling down into the remoter code, seem like our default `retry=1` has a hidden retry=3 on case it's retrievable exception, which timeout if one of them.

We might consider setting the `retry=0` on the stress commands, there's no real reason we ever want to retry.

roydahan:
We should do it only for stress commands but to test it heavily (shouldn't go into 5.1).

fruch:
PR to address it:
https://github.com/scylladb/scylla-cluster-tests/pull/5454

PR #5454, opened 2022-11-23
fix(loaders): disable retry on main loader run commands

fruch:
we never want to retry a stress comamnd, regardless what
the exception was, if we retry midst the stress command
we would start it from begining (which is an outcome we never want)

list of exception we retry is defined in `CommandRunner._is_error_retryable`

Fixes: #5278


vponomaryov (commented):
Both commits look ok to me, but they are unrelated to each other.
So, why are they part of the same PR?

fruch:
> Both commits look ok to me, but they are unrelated to each other. So, why are they part of the same PR?

I'm testing them together, I'll have them separated once they would work

vponomaryov (approved):

PR #5459, opened 2022-11-23
improvement(config param): support None for service_level_shares

juliayakovlev:
This commit must for https://github.com/scylladb/scylla-cluster-tests/pull/5451. 
OSS does not support 'shares' attribute for Service Level. But it's possible to create Service Level without 'shares'. 
Add option to define shares as None and ignore this attribute in time SL creation. 
'service_level_shares' config parameter define how many roles should be created for load running


fruch:
@juliayakovlev why no just use yaml `null` ? it would give you the exact same effect

juliayakovlev:
> @juliayakovlev why no just use yaml `null` ? it would give you the exact same effect

Thanks, got it. Close PR

PR #5466, opened 2022-11-23
improvement(k8s): increase load in K8S jobs

vponomaryov:
Having load 2k op/s (2 laoders * 1k op/s) for 3 nodes with i3.4xlarge nodes is too small.
So, increase it to be 5k op/s per loader.
On GKE nodes are slower than on EKS, but must be able to handle 10k op/s too.

### DIFF
--- a/test-cases/scylla-operator/longevity-scylla-operator-3h-add-remove-rack.yaml
+++ b/test-cases/scylla-operator/longevity-scylla-operator-3h-add-remove-rack.yaml
@@ -1,5 +1,5 @@
 test_duration: 300
-stress_cmd: ["cassandra-stress write cl=QUORUM duration=180m -schema 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=100 throttle=1000/s -pop seq=1..10000000 -log interval=5"
+stress_cmd: ["cassandra-stress write cl=QUORUM duration=180m -schema 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=100 throttle=5000/s -pop seq=1..10000000 -log interval=5"
              ]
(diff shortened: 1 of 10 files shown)


fruch:
Why do we throttle to begin with ? In VMs we usually don't throttle

For the typical use case we want ~70% CPU utilization in steady state (the spare are operations like streaming/repair and such)

In the case of one pod per k8s node, we should be quite close to how VMs are working.

vponomaryov:
> Why do we throttle to begin with ? In VMs we usually don't throttle
> 
> For the typical use case we want ~70% CPU utilization in steady state (the spare are operations like streaming/repair and such)
> 
> In the case of one pod per k8s node, we should be quite close to how VMs are working.

Then we need to measure it first and, moreover, measure EKS and GKE separately, because their values will be much different.

fruch:
> > Why do we throttle to begin with ? In VMs we usually don't throttle
> > For the typical use case we want ~70% CPU utilization in steady state (the spare are operations like streaming/repair and such)
> > In the case of one pod per k8s node, we should be quite close to how VMs are working.
> 
> Then we need to measure it first and, moreover, measure EKS and GKE separately, because their values will be much different.

so let do that, to make sure we we know how much load we put in out tests.

roydahan:
@vponomaryov is this PR still relevant?
Where does it stand?

roydahan:
@vponomaryov is this PR still relevant?
Where does it stand?

vponomaryov:
> @vponomaryov is this PR still relevant? Where does it stand?

The measurement part hasn't been scheduled for implementation so far.

Personally, I don't see any problem if we do both things - merge this change to increase load from `"nano"` to be about `"small"`.
And then, anytime later we measure it for different backends and different instance types and configure appropriately.

fruch (approved):

PR #5470, opened 2022-11-24
improvement(stall threshold): upgrade test to have bigger threshold

fgelcer:
we are recently seen many reactor stalls reported as `ERROR` events, making part of the tests to fail, but in most of the cases, we do expect some of these stalls during init of new nodes, or taking nodes down, hence we decided to increase the "tolerable" stall to 2000ms.


roydahan (dismissed):

defaults/test_default.yaml
```diff
@@ -13,6 +13,7 @@ manager_scylla_backend_version: '2022'
 experimental: true
 round_robin: false
 
+reactor_stall_tolerable_threshold: 500
```
roydahan, line 16: So what was the value before that?

fgelcer: the value was `500`, but hardcoded in the `sdcm/sct_events/database.py` file

BTW, i missed changing it in there, and now it is fixed

fgelcer:
giving it a try [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/PR5470_control_reactor_stall_tolerable/2/)

sdcm/sct_events/database.py
```diff
@@ -66,7 +64,7 @@ class DatabaseLogEvent(LogEvent, abstract=True):
 
 # pylint: disable=too-few-public-methods
 class ReactorStalledMixin(Generic[T_log_event]):
-    tolerable_reactor_stall: int = TOLERABLE_REACTOR_STALL
+    tolerable_reactor_stall: int = os.environ.get('SCT_REACTOR_STALL_TOLERABLE_THRESHOLD')
```
aleksbykov, line 67: Does it work? i think it wouldn't, because we didn't export config parameters to os.env. 
i think it is better to get it from node.parent_cluster.params["reacto_stall_tolerable_threshold"]

fgelcer: @aleksbykov , running it in here with very small threshold and detector:
https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/PR5470_control_reactor_stall_tolerable/4/

fgelcer:
> @fgelcer this change still needed?

i think it is still desirable, but did not have time to deal with it recently...

fruch:
It's nice to have feature, but we don't have this issue in rolling upgrade tests anymore

Closing this for now

Issue #5473, opened 2022-11-24
Recent master runs aren't getting reported to Argus


fruch:
Recent runs in master doesn't seems to be reported in Argus

https://jenkins.scylladb.com/job/scylla-master/job/longevity/job/longevity-10gb-3h-gce-test/285/
https://jenkins.scylladb.com/job/scylla-master/job/longevity/job/longevity-10gb-3h-gce-test/286/

argus.log:
```
< t:2022-11-23 17:17:54,833 f:argus_test_run.py l:244  c:sdcm.argus_test_run  p:INFO  > Preparing Test Details...
< t:2022-11-23 17:17:54,835 f:argus_test_run.py l:258  c:sdcm.argus_test_run  p:INFO  > Preparing Resource Setup...
< t:2022-11-23 17:17:54,835 f:argus_test_run.py l:278  c:sdcm.argus_test_run  p:INFO  > Initializing TestRun...
< t:2022-11-23 17:21:21,283 f:argus_test_run.py l:236  c:sdcm.argus_test_run  p:WARNING > Returning MagicMock from ArgusTestRun.get() as we are unable to acquire Argus connection
```

fruch:
looking at the logs...
```
< t:2022-11-23 17:17:56,584 f:tester.py       l:370  c:LongevityTest        p:WARNING > Unable to set up Argus connection: Unable to connect to any servers
< t:2022-11-23 17:17:56,584 f:tester.py       l:371  c:LongevityTest        p:DEBUG > Error details:  < t:2022-11-23 17:17:56,584 f:tester.py       l:371  c:LongevityTest        p:DEBUG > Error details:
< t:2022-11-23 17:17:56,584 f:tester.py       l:371  c:LongevityTest        p:DEBUG > Traceback (most recent call last):
< t:2022-11-23 17:17:56,584 f:tester.py       l:371  c:LongevityTest        p:DEBUG >   File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 365, in argus_test_run
< t:2022-11-23 17:17:56,584 f:tester.py       l:371  c:LongevityTest        p:DEBUG >     argus_test_run = ArgusTestRun.from_sct_config(test_id=UUID(self.test_id),
< t:2022-11-23 17:17:56,584 f:tester.py       l:371  c:LongevityTest        p:DEBUG >   File "/home/ubuntu/scylla-cluster-tests/sdcm/argus_test_run.py", line 279, in from_sct_config
[...]

PR #5482, opened 2022-11-27
fix(argus): Switch to Argus version that works `AddressTranslator`

fruch:
* Since we are using public addresses, we should move to using `AddressTranslator` and not count the driver to be able to connect to one node only
* added a unittest that should cover the connectivity to argus is working to avoid similar things happing unnoticed again

Ref: https://github.com/scylladb/scylla-cluster-tests/pull/5476
Ref: https://github.com/scylladb/scylla-cluster-tests/issues/5473
Ref: https://github.com/scylladb/argus/pull/185


fruch:
@k0machi currently in draft until we'll have an argus release that supports this.

unit_tests/test_argus.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+@pytest.mark.integration
+@pytest.mark.need_network
+@mock.patch.dict(os.environ, {
+    "JOB_NAME": "test_connectivity",
+})
+@pytest.mark.sct_config(files='test-cases/PR-provision-test-docker.yaml')
+def test_connectivity(params):
+    argus_test_run = ArgusTestRun.from_sct_config(test_id='test_connectivity', sct_config=params)
```
k0machi, line 19: `test_id` should be a UUID, and you can't really re-use a single one as that will lead to multiple rows with same test id (since id is not the primary key)

fruch: I don't really use it. i.e. I'm not calling save.

I just want to make sure it's connecting, not fill the DB with garbage.

fruch: other suggestions on how to achieve that goal are welcomed

fruch: also now the test is failing:
```
09:32:59          if cls._config is None:
09:32:59  >           cls._config = ArgusConfig(**KeyStore().get_argusdb_credentials(), keyspace_name="argus")
09:32:59  E           TypeError: Config.__init__() got an unexpected keyword argument 'address_mapping'
```

since the `ArgusConfig` is a bit unfriendly for extensions, using it like this.
cause of this I'm force to have multiple versions of the configuration (since this configuration is shared across SCT branches)

k0machi: > I don't really use it. i.e. I'm not calling save.
> 
> I just want to make sure it's connecting, not fill the DB with garbage.

Oh, right, then ignore that comment. Yeah that should be enough to just test the connectivity as connection is established as soon as the ArgusDatabase class is instantiated.

PR #5486, opened 2022-11-27
feature(monitor): move to newer images

fruch:
move to formal monitoring images, and skip installation of docker
and such if such images are identifed.
(also mean that now by default we will be using an ubuntu image)

Testing:
- [ ] AWS - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity/job/longevity-10gb-3h/85/
- [ ] GCE - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-10gb-3h-gce-test/24/

### DIFF
--- a/defaults/aws_config.yaml
+++ b/defaults/aws_config.yaml
@@ -8,7 +8,7 @@ instance_type_loader: 'c5.xlarge'
 instance_type_monitor: 't3.large'
 # manual on creating loader AMI's: see docs/new_loader_ami.md
 ami_id_loader: 'scylla-qa-loader-ami-v19'
-ami_id_monitor: 'ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20221206' # ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20221206 Canonical, Ubuntu, 22.04 LTS, amd64 jammy image build on 2022-12-06
+ami_id_monitor: 'scylla-monitor-4.2'
 
 availability_zone: 'a'
 root_disk_size_monitor: 50  # GB, remove this field if default disk size should be used
--- a/defaults/gce_config.yaml
+++ b/defaults/gce_config.yaml
@@ -5,7 +5,7 @@ gce_project: '' # empty mean default one, can be overwritten to use different on
 
 gce_image_db: '' # so we can use `scylla_version` as needed
 gce_image_loader: 'https://www.googleapis.com/compute/v1/projects/centos-cloud/global/images/family/centos-7'
-gce_image_monitor: 'https://www.googleapis.com/compute/v1/projects/centos-cloud/global/images/family/centos-7'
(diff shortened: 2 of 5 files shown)


fruch:
testing in:
https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity/job/longevity-10gb-3h/80

defaults/gce_config.yaml
```diff
@@ -5,7 +5,7 @@ gce_project: '' # empty mean default one, can be overwritten to use different on
 
 gce_image_db: '' # so we can use `scylla_version` as needed
 gce_image_loader: 'https://www.googleapis.com/compute/v1/projects/centos-cloud/global/images/family/centos-7'
-gce_image_monitor: 'https://www.googleapis.com/compute/v1/projects/centos-cloud/global/images/family/centos-7'
+gce_image_monitor: 'https://www.googleapis.com/compute/alpha/projects/skilled-adapter-452/global/images/scylla-monitoring-cloud-ubuntu-4-0-2-2022-09-19t14-39-35z'
```
fruch, line 8: should try it with v1 as well (even that this is the url google is returning for this image)

defaults/aws_config.yaml
```diff
@@ -10,22 +10,22 @@ instance_type_monitor: 't3.large'
 regions_data:
   us-east-1: # US East (N. Virginia)
     ami_id_loader: 'ami-00c1a16b5136fbb7c' # Loader dedicated AMI v19
-    ami_id_monitor: 'ami-02eac2c0129f6376b' # Official CentOS Linux 7 x86_64 HVM EBS ENA 1901_01
+    ami_id_monitor: 'ami-01a7efd0776f9a4fe' # scylla_monitoring_2022-09-19T14-12-11Z
```
roydahan, line 13: Which version is it of monitor?
I would like it to be clear so every time we have new monitor version (e.g. 4.1).

@amnonh will send a PR that change the branch name and the images IDs.
(We would be nicer if we could read these IDs from a centralized location like we have in scylla-version).

fruch: I think it's 4.2,

but we take the image for the cached dependencies, not for the monitoring version inside.

we still need to preserve the option to install the monitoring from clean image, cause we manger tests.

fruch:
@roydahan for taking the image in GCE, we have some permission issue to handle first, since the image in one project and we running in a different project

defaults/aws_config.yaml
roydahan: Installing monitoring will be based on cleanOS.
Wherever we do that (e.g new atifact test for monitor or Azure or manager test).

The rest should use the image as is without installing anything.

This is not high priority at the moment, and you don't need to work on it right now.
(This entire PR).

fruch:
> @roydahan for taking the image in GCE, we have some permission issue to handle first, since the image in one project and we running in a different project

permission issue with GCP is handled and we can use those images across project

defaults/aws_config.yaml
fruch: Start a thread with @amnonh so next release would have better names to indicate their branch name

vponomaryov (approved):
LGTM assuming it is tested and works.

fruch:
there a bigger task for handling new monitor images with their own user-script

PR #5494, opened 2022-11-28
feature(k8s-serverless): cloud connectivity support for cqlsh

fruch:
* add the option to use the cloud bundle if available for
  connecting with cqlsh

* added a functional test to cover it
 
### DIFF
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -775,6 +775,7 @@ def test_operator_managed_tls(db_cluster: ScyllaPodCluster, tmp_path: path.Path)
         log.debug(file)
         log.debug(file.read_text())
 
+    db_cluster.nodes[0].refresh_ip_address()
     cluster = Cluster(contact_points=[db_cluster.nodes[0].cql_ip_address], port=db_cluster.nodes[0].CQL_SSL_PORT)
     ssl_context = ssl.SSLContext(protocol=ssl.PROTOCOL_SSLv23)
     ssl_context.verify_mode = ssl.VerifyMode.CERT_REQUIRED  # pylint: disable=no-member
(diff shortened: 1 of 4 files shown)


functional_tests/scylla_operator/test_functional.py
```diff
@@ -849,3 +849,15 @@ def test_cloud_bundle_connectivity_scylla_bench(tester):
 
     # TODO: add verification that the output say it's using the cloud bundle
     #  (need to add that to log output in scylla-bench)
+
+
+@pytest.mark.required_operator("v1.8.0")
+@pytest.mark.requires_tls
+def test_cloud_bundle_connectivity_cqlsh(db_cluster: ScyllaPodCluster):
+
+    assert db_cluster.connection_bundle_file, "cloud bundle wasn't found"
```
vponomaryov, line 858: Not sure that we should fail the test in this case. It is kind of precondition.

fruch: it can fail, and then we fallback to use the cqlsh direct, and it would work o.k. since the clear ports are still open.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -775,6 +775,7 @@ def test_operator_managed_tls(db_cluster: ScyllaPodCluster, tmp_path: path.Path)
         log.debug(file)
         log.debug(file.read_text())
 
+    db_cluster.nodes[0].refresh_ip_address()
```
vponomaryov, line 778: Why it was added?

fruch: see this failure:

https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-5494/9/execution/node/117/log/

could be that now it's not needed, and that run was before merging the ip tracker.

fruch: ooh, the ip tracker isn't related to k8s-local. so it was just wild goose chase...
I guess it can be removed

PR #5503, opened 2022-11-29
feature(stress-thread): set timeout based on duration

KnifeyMoloko:
Sometimes we face freeze of c-s thread blocking the test flow.

This commit adds automatic timeout to stress threads (c-s, sb and gemini) when duration in stress cmd is specified.
Timeout is calculated as this:
warmup + duration + 600 (all in seconds).

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -75,7 +75,7 @@ from sdcm.sct_events.setup import start_events_device, stop_events_device
 from sdcm.sct_events.system import InfoEvent, TestFrameworkEvent, TestResultEvent, TestTimeoutEvent
 from sdcm.sct_events.file_logger import get_events_grouped_by_category, get_logger_event_summary
 from sdcm.sct_events.events_analyzer import stop_events_analyzer
-from sdcm.stress_thread import CassandraStressThread
+from sdcm.stress_thread import CassandraStressThread, get_timeout_from_stress_cmd
 from sdcm.gemini_thread import GeminiStressThread
 from sdcm.utils.log_time_consistency import DbLogTimeConsistencyAnalyzer
 from sdcm.utils.threads_and_processes_alive import gather_live_processes_and_dump_to_file, \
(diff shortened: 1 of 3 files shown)


sdcm/tester.py
```diff
@@ -1519,7 +1519,11 @@ def run_stress_cassandra_thread(
             self, stress_cmd, duration=None, stress_num=1, keyspace_num=1, profile=None, prefix='', round_robin=False,
             stats_aggregate_cmds=True, keyspace_name=None, stop_test_on_failure=True, **_):
         # stress_cmd = self._cs_add_node_flag(stress_cmd)
-        timeout = self.get_duration(duration)
+        if duration:
+            timeout = self.get_duration(duration)
+        else:
+            timeout = get_timeout_from_stress_cmd(stress_cmd) or self.get_duration(duration)
```
fgelcer, line 1525: this might be a problem, because in case we do write without `duration`, but expect to write a specific number of keys, in big data set tests, it can take hours, so the 10 min we give here will kill these tests

KnifeyMoloko: If this is a problem, should we have bumped into it on master?

fruch: it's default to test duration which is long enough.

but if it's really stuck, test duration is also too big, since in rolling upgrade to are execute in the midst of the test.

KnifeyMoloko: @fruch It might be even too long for longevities. We use `test_duration` as the `timeout` for getting the stress futures, for the remoter cmd and for the results verification (all in the `CassandraStressThread` class). If any of these get's stuck or is slower, we're guaranteed to end the test with a critical test timeout. Or maybe I'm missing something here?

fgelcer, line 1525: could you just use:

```suggestion
        timeout = get_timeout_from_stress_cmd(stress_cmd) or self.get_duration(duration)
```

fgelcer (commented):
well, if we already give the test's whole timeout, this would improve in the cases where we do have `duration` in the stress commands, hence LGTM
just a little question to get it improved, if possible

soyacz (approved):

roydahan:
I don't understand when this timeout is needed.
Every test should have a test_duration and the test timeout is already the timeout for the entire test, isn't it?
If not, in which cases it's not the case?

Also, why is it targeted to branch-5.0?

fruch:
> I don't understand when this timeout is needed. Every test should have a test_duration and the test timeout is already the timeout for the entire test, isn't it? If not, in which cases it's not the case?
> 
> Also, why is it targeted to branch-5.0?

this is a change already in master, @KnifeyMoloko wanted to backport it, but I'm not yet convinced it's really needs to be backported, and that this change is going to solve any specific issue.

roydahan:
Ok, So I don't think we need it for 5.0 for now.
I'll close it and if needed in 5.0 by any chance we will get back to it.

PR #5505, opened 2022-11-29
test(compaction_limit): added new feature test

soyacz:
Test measuring impact on ops/latency when compaction limit is set to
different values. In case of ops/latency drop in relation to base
results (no limits applied) raises `ErrorEvent`.

refs: https://github.com/scylladb/qa-tasks/issues/256


soyacz:
test job: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/compaction-throughput-limit-test/9/

soyacz:
Test shows that enforcing limit increases latency and reduces op/s - maybe disk load is too heavy for these instances and we don't comply to recommendations of limit to be x16 higher than throughput. I'll retest with 2xlarge machines to see how it behaves. 
Results can be checked with:
`hydra investigate show-monitor e251a9d4-89fe-47fd-9a4b-960348239e49`
Example graph (can see 3 runs: no limits, 85% limit, 65% limit)
image

soyacz:
I tested with i3.xlarge machines - there's visible improvement when limiting compaction throughput to 60%. Still looks disk is heavily loaded - I'll retry with smaller disk footprint.

image
image

jenkins-pipelines/features-per-partition-limit.jenkinsfile
```diff
@@ -6,7 +6,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 longevityPipeline(
```
roydahan, line 6: I think this file isn't related to this PR, is it?

roydahan: I see now that it covers two features (not only the one in the title)

soyacz: I reused some code from per-partition limit test - so I moved it to different file&class.

soyacz:
I think there's nothing more to do regarding this test. Feature looks working correctly. Of course it must be used with caution.
Graphs from recent test for limits: 
no limit, 283MB/s, 182MB/s, 81MB/s, 4MB/s:
throughput_latency
compaction_throughput

@fruch @roydahan please review

fruch:
> I think there's nothing more to do regarding this test. Feature looks working correctly. Of course it must be used with caution. Graphs from recent test for limits: no limit, 283MB/s, 182MB/s, 81MB/s, 4MB/s: throughput_latency compaction_throughput
> 
> @fruch @roydahan please review

I would argue that 70% and 45% are less interesting, and would like to see more between 20% and %1, where we actually see it in action, and can deduce something about what's the optimal limit for this use-case (i.e. specific cassnadra-stress command)

fruch (approved):
LGTM

roydahan (approved):

PR #5515, opened 2022-12-03
feature(k8s): implement Scylla pods IP change watcher

vponomaryov:


vponomaryov:
It is tested here:
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-upgrade-platform-k8s-eks/122/consoleFull
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-12h-multitenant-eks/11/consoleFull

The feature itself works as it can see in the jobs above. 
The multitenant job has bunch of other repeating failures, which are unrelated to the new feature. Those to be investigated.
As part of this PR, I planned also to add unit tests.

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+            LOGGER.warning("'_open_stream()': failed to open stream:\n%s", exc)
+            # NOTE: following is workaround for the error 401 which may happen due to
+            #       some config data corruption during the forced socket connection failure
+            self._k8s_core_v1_api = KubernetesOps.core_v1_api(self.k8s_kluster.get_api_client())
+            raise ConnectionError(str(exc)) from None
+
+    @retrying(n=3600, sleep_time=1,  allowed_exceptions=(
+        ProtocolError, IncompleteRead, ReadTimeoutError, TimeoutError, ValueError))
+    def _read_stream(self) -> Iterator[str]:
+        while True:
```
fruch, line 682: how this thread is being stopped ? if we are in an endless loop ?

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+    @retrying(n=987654, sleep_time=10, allowed_exceptions=(Exception, ))
+    def run(self) -> None:
+        try:
+            for line in self._read_stream():
+                self._process_line(line)
+        finally:
+            if hasattr(self.watcher, 'close'):
+                self.watcher.close()
+
+    def stop(self, timeout=None) -> None:
```
fruch, line 857: this doesn't do what's written on the cover, i.e. it joins, and doesn't stop anything.

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+
+        Per-pod callbacks get run first, then per-namespace ones.
+        """
+        if not callbacks:
+            LOGGER.warning(
+                "No callbacks are provided. Nothing to register. "
+                "namespace='%s', pod_name='%s'",
+                namespace, pod_name)
+            return
+        elif callable(callbacks):
```
fruch, line 822: if you support both a list, and one item, it should marked in the type annotations as such

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+            try:
+                func(*args, **kwargs)
+            except Exception as exc:  # pylint: disable=broad-except
+                LOGGER.warning("Callback call failed %s: %s", suffix, str(exc))
+
+        data = self.mapper_dict.get(namespace, {})
+        for dict_key in (pod_name, '__each__'):
+            callbacks = data.get(dict_key, {}).get('callbacks', [])
+            if dict_key == '__each__':
+                # NOTE: when 'add_pod_name_as_kwarg' is True it means
```
fruch, line 774: why the order is important ?

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+        #       single namespace.
+
+        def process_callback(callback, namespace, pod_name=None, add_pod_name_as_kwarg=False):
+            suffix = f"(namespace={namespace}, pod_name={pod_name})"
+            func = callback[0]
+            args = callback[1]
+            kwargs = {} | callback[2]
+            if add_pod_name_as_kwarg:
+                kwargs['pod_name'] = pod_name
+            # TODO: make following be 'debug' message
```
fruch, line 763: don't forget this

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+                    self._call_callbacks(namespace, name)
+                    break
+                else:
+                    break
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.warning(
+                "Failed to parse following line: %s\nerr: %s", line, exc)
+
+    def _call_callbacks(self, namespace, pod_name) -> None:
+        # TODO: run callbacks in parallel on per-namespace basis and serially inside a
```
fruch, line 753: how much time our callback takes ? 
refresh_ip shouldn't be that long, and update ip in monitor should be too long as well, i.e. it's less then a few seconds, long before the scylla process is ready

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+        self.watcher = None
+        self._k8s_core_v1_api = KubernetesOps.core_v1_api(self.k8s_kluster.get_api_client())
+
+    @retrying(n=3600, sleep_time=1, allowed_exceptions=(ConnectionError, ))
+    def _open_stream(self, cache={}) -> None:  # pylint: disable=dangerous-default-value
+        try:
+            now = time.time()
+            if cache.get("last_call_at", 0) + 5 > now:
+                # NOTE: sleep for some time if it is called too often
+                time.sleep(5)
```
fruch, line 663: wouldn't `sleep_time=5` in the retrying decorator would be enough ?

upgrade_test.py
```diff
@@ -1058,10 +1058,6 @@ def test_kubernetes_platform_upgrade(self):
         upgrade_version, scylla_pool = self.k8s_cluster.upgrade_kubernetes_platform(
             pod_objects=self.db_cluster.nodes,
             use_additional_scylla_nodepool=True)
-        # NOTE: refresh IP addresses of the Scylla nodes because they get changed in current case
-        for db_node in self.db_cluster.nodes:
-            db_node.refresh_ip_address()
-        self.monitors.reconfigure_scylla_monitoring()
```
fruch, line 1064: lets remove all the hacks, also from nemesis, as we talked

sdcm/utils/k8s.py
```diff
@@ -609,6 +614,251 @@ def stop(self, timeout=None):
...
+                for address in subset.get('addresses', []):
+                    current_ip_address = address['ip']
+                    if current_ip_address == self.mapper_dict[namespace][name]['current_ip']:
+                        continue
+                    old_ip_candidate = self.mapper_dict[namespace][name]['current_ip']
+                    self.mapper_dict[namespace][name]['current_ip'] = current_ip_address
+                    if not old_ip_candidate:
+                        break
+                    self.mapper_dict[namespace][name]['old_ips'].append(old_ip_candidate)
+                    LOGGER.info(
```
fruch, line 741: let's leave this info

fruch (changes requested):
overall look good

few questions and clarification, the main concern was with the stop.

sdcm/utils/k8s.py
vponomaryov: At first, we haven't been stopping any of the threads running in K8S. Why? Because we call this code in the "cleanup" stage where it is not initiated. So, all the additional threads we have are not stopped explicitly.
At second, we must not stop this thread somewhere in the middle. And it is totally ok to be closed automatically with the main python process.

In a newer version I have in my dev branch I added `_termination_event` attribute, but it is not really needed.
Due to the nature of this thread, it will hang. It will hang due to the endless timeout which we need because it is needed by the design - `wait for updates`.

sdcm/utils/k8s.py
vponomaryov: Read above. Not needed, but I will keep it close to what you expect. The thread will always hang on `self.watcher.close()`.

sdcm/utils/k8s.py
vponomaryov: ok, will do

sdcm/utils/k8s.py
vponomaryov: Because we must refresh IP addresses before reconfiguring the monitoring. If we do it vice-versa then the monitoring update becomes nonsense.

sdcm/utils/k8s.py
vponomaryov: It is here to not forget it

sdcm/utils/k8s.py
vponomaryov: It takes seconds. But this todo is not urgent, because workability of the feature as a whole is much more needed.

sdcm/utils/k8s.py
vponomaryov: It is not the reason for a `retry`. It must not be a retry from my POW.

upgrade_test.py
vponomaryov: Did it in my latest dev update which has not been uploaded here yet. 
But remember, if this thread gets stuck, we won't have any callbacks be running.
So, it is kind of risky to remove it from my POW.

sdcm/utils/k8s.py
vponomaryov: It has no comment above to do something with it. So, it is intended to stay as is.

sdcm/utils/k8s.py
fruch: o.k.

upgrade_test.py
fruch: I still think we should remove those, and not do those extra steps again and again.

sdcm/utils/k8s.py
vponomaryov: During the K8S platform upgrade, which longs for minutes, without this sleep we would send tons of requests to the unavilable API.

sdcm/utils/k8s.py
vponomaryov: Done

sdcm/utils/k8s.py
vponomaryov: Done

upgrade_test.py
vponomaryov: Done

vponomaryov:
The latest state of this PR has successfully passed here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-upgrade-platform-k8s-eks/124/consoleFull

sdcm/utils/k8s.py
fruch: What happen when we'll need to handle multiple cluster, I still think we need ability to stop this

sdcm/utils/k8s.py
fruch: Well this exact note is what missing here

fruch (approved):
LGTM

Issue #5517, opened 2022-12-04
artifacts-docker-test Failed with `AssertionError: '' != 'en_US.UTF-8'`


yaronkaikov:
https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-docker-test/582/ Failed with the following error message:
```
06:50:06  2022-12-04 04:49:29.047: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=7c51347b-3ab9-44fc-9624-4f05ee3f0832, source=ArtifactsTest.test_scylla_service (artifacts_test.ArtifactsTest) [check locale settings in the docker image]() message=Traceback (most recent call last):
06:50:06  File "/jenkins/slave/workspace/scylla-master/artifacts/artifacts-docker-test/scylla-cluster-tests/artifacts_test.py", line 342, in test_scylla_service
06:50:06  self.verify_docker_locale_settings()
06:50:06  File "/jenkins/slave/workspace/scylla-master/artifacts/artifacts-docker-test/scylla-cluster-tests/artifacts_test.py", line 248, in verify_docker_locale_settings
06:50:06  self.assertEqual(configured_value, expected_value)
06:50:06  AssertionError: '' != 'en_US.UTF-8'
06:50:06  + en_US.UTF-8

```

Probably related to changes done in https://github.com/scylladb/scylladb/pull/12122
SCT side should be modified as well

PR #5518, opened 2022-12-04
fix(docker artifacts): fix locale verification

fgelcer:
introduced on commit https://github.com/scylladb/scylladb/pull/12122, we don't have `LANGUAGE`, and changed to the minimal Ubuntu image not having `en_US.UTF-8`.

Fixes: #5517

### DIFF
--- a/artifacts_test.py
+++ b/artifacts_test.py
@@ -234,9 +234,9 @@ class ArtifactsTest(ClusterTester):  # pylint: disable=too-many-public-methods
     def verify_docker_locale_settings(self) -> None:
         run = self.node.remoter.run
         expected_locale_settings = {
-            "LC_ALL": "en_US.UTF-8",
-            "LANG": "en_US.UTF-8",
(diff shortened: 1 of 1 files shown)


fgelcer:
only problem here would be if we run on older docker images, from master branch, and that will fail

yaronkaikov:
@fgelcer  maybe adding both options for a while is better approach? But why do we actually need to verify those ?

fgelcer:
> @fgelcer maybe adding both options for a while is better approach? But why do we actually need to verify those ?

it was introduced by 0324942753d8e4a5b56963853a5cda8eb583707d and the tasks that generated this job says:
`#9570 - The locale is now set to use UTF-8 encoding in the container image, so cqlsh works correctly when displaying non-ascii data.`

roydahan (approved):

fgelcer:
closing in favor of #5519.
if decided the alternative one is not good, we can close that one, and re-open this one

PR #5519, opened 2022-12-04
fix(docker artifacts): fix locale verification

fgelcer:
introduced on commit scylladb/scylladb#12122,
we don't have `LANGUAGE`, and changed to the minimal Ubuntu image not having `en_US.UTF-8`.

Fixes: scylladb#5517
this is an alternative PR for #5518

### DIFF
--- a/artifacts_test.py
+++ b/artifacts_test.py
@@ -234,9 +234,9 @@ class ArtifactsTest(ClusterTester):  # pylint: disable=too-many-public-methods
     def verify_docker_locale_settings(self) -> None:
         run = self.node.remoter.run
         expected_locale_settings = {
-            "LC_ALL": "en_US.UTF-8",
-            "LANG": "en_US.UTF-8",
-            "LANGUAGE": "en_US:en",
+            "LC_ALL": ["en_US.UTF-8", "C.UTF-8"],
+            "LANG": ["en_US.UTF-8", "C.UTF-8"],
+            "LANGUAGE": ["en_US:en", ""],
         }
 
         locale_settings_raw = run("locale").stdout.split(sep="\n")
@@ -245,7 +245,7 @@ class ArtifactsTest(ClusterTester):  # pylint: disable=too-many-public-methods
 
         for setting_key, expected_value in expected_locale_settings.items():
             configured_value = locale_settings.get(setting_key)
-            self.assertEqual(configured_value, expected_value)
+            self.assertIn(configured_value, expected_value, f'value for key {setting_key} differs')
 
     def verify_docker_latest_match_release(self) -> None:
         for product in typing.get_args(ScyllaProduct):


fruch (approved):
LGTM

fgelcer:
running [here](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mark/job/artifacts_docker/6/)

but it seems like it is broken, as we expected to have `C.UTF-8` for `LC_ALL`, but we got `''` for some reason

```
AssertionError: '' not found in ['en_US.UTF-8', 'C.UTF-8'] : value for key LC_ALL differs
```

checking the commit on https://github.com/scylladb/scylladb/pull/12122, i see that `locale-gen` was removed completely, and it probably is the cause of it.

@syuu1228 , please advise

syuu1228:
> running [here](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mark/job/artifacts_docker/6/)
> 
> but it seems like it is broken, as we expected to have `C.UTF-8` for `LC_ALL`, but we got `''` for some reason
> 
> ```
> AssertionError: '' not found in ['en_US.UTF-8', 'C.UTF-8'] : value for key LC_ALL differs
> ```
> 
> checking the commit on [scylladb/scylladb#12122](https://github.com/scylladb/scylladb/pull/12122), i see that `locale-gen` was removed completely, and it probably is the cause of it.
> 
> @syuu1228 , please advise

Sorry, I verified LC_ALL is set to C.UTF-8 but it was `docker exec` shell, and the issue is occuring on SSH shell, because way of set locale is different between them.
I will post fix very soon.

fgelcer:
> Sent a PR: [scylladb/scylladb#12197](https://github.com/scylladb/scylladb/pull/12197)

@syuu1228 , is it enough? don't we need to do similar configuration for `LANG` too?

fruch:
@syuu1228 

running it with `5.2.0-dev-0.20221207.47a8fad2a2bd`
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mark/job/artifacts_docker/7/

it's failing

fruch:
> @fruch Why do we even need to verify this? are we actually using it somewhere?

I don't know, I just want the latest docker image to get updated.

yaronkaikov:
> > @fruch Why do we even need to verify this? are we actually using it somewhere?
> 
> I don't know, I just want the latest docker image to get updated.

@fruch So let's remove it maybe?

fruch:
test was added to cover:
https://github.com/scylladb/scylladb/commit/546e4adf9e1d560cbb99932925b13e30200c0d3a

if it's not relevant anymore (and cqlsh still works under that image), I think it's can be removed

fgelcer:
@fruch , i think it is a mistake to close this PR, as we may face issues again with CQL commands encode and not verifying it my hit us again

fgelcer:
> > @fruch , i think it is a mistake to close this PR, as we may face issues again with CQL commands encode and not verifying it my hit us again
> 
> The check that was removed doesn't verify any encoding issues in cqlsh.

nothing was removed, but rather we added an additional verification for the locale setting (to support both the next configuration and old)
removing it, we can hit the issue we had with CQL reported by Felipe IIRC

fruch:
> > > @fruch , i think it is a mistake to close this PR, as we may face issues again with CQL commands encode and not verifying it my hit us again
> > 
> > The check that was removed doesn't verify any encoding issues in cqlsh.
> 
> nothing was removed, but rather we added an additional verification for the locale setting (to support both the next configuration and old)
> removing it, we can hit the issue we had with CQL reported by Felipe IIRC

I still argue the check should be using CQLSH with problematic encoding, and not checking environment variables

PR #5527, opened 2022-12-05
Shell fixes

enaydanov:
Add group ID for Docker container user:

```shell
$ docker run --help
...
  -u, --user string                    Username or UID (format: <name|uid>[:<group|gid>])
...
```

This make it to set a proper group for files created by processes inside a container.

Also, to set the shell properly in sh step in Jenkinsfile, the shebang line must be exactly the first line of a script. Otherwise, system shell will be used (with -xe args.)

Despite we live with it from the very beginning it hits us sometimes.

### DIFF
--- a/docker/env/hydra.sh
+++ b/docker/env/hydra.sh
@@ -12,7 +12,7 @@ SCT_DIR=$(dirname "${SCT_DIR}")
 VERSION=v$(cat "${DOCKER_ENV_DIR}/version")
 HOST_NAME=SCT-CONTAINER
 RUN_BY_USER=$(python3 "${SCT_DIR}/sdcm/utils/get_username.py")
-USER_ID=$(id -u "${USER}")
+USER_ID=$(id -u "${USER}"):$(id -g "${USER}")
 HOME_DIR=${HOME}
 
 CREATE_RUNNER_INSTANCE=""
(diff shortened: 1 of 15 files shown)


fruch:
@enaydanov can you point to where exactly do we hit this, that we need this ?

fruch (approved):
LGTM

enaydanov:
> > > @enaydanov can you point to where exactly do we hit this, that we need this ?
> > 
> > 
> > Some of our builders have non-BASH system shell. And if some script uses bashisms (like [[, for example) then it will make very unexpected results. An email without results, for example.
> 
> How's the group id is related to non-BASH ?
> 
> also why not align all the builders to be configured exactly the same ? (guessing this change won't fix the issue, maybe only show them)

Ah, you are about this change. This thing is more about being consistent: we're using same user ID inside the container to have all files we write inside the container have correct ownership but completely forgot about the group.

So, if you do 

```
$ hydra 'bash -c "touch blah"
...
$ ls -l blah
-rw-r--r-- 1 ubuntu root 0 Dec  6 06:12 blah
```

After my change it will have correct group ID.

My previous comment was about fixing shebang lines. 

Again:

This is incorrect:

```
sh """
#!/bin/bash  <-- it's not a shebang line, it's just a comment
...
"""
```

and this is correct:

```
sh """#!/bin/bash  <-- this is shebang line and following script will be run using bash
...
"""
```

PR #5530, opened 2022-12-06
fix(get_nodes_status): print more information about exception

ilya-rarov:
When the exception does not have `.args`, the statement `f"{exc}"` returns nothing, and the reason why `nodetool status` failed is not printed in the log.
Changed printing to display the exception class name together with the `.args`.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2646,11 +2646,11 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
                         if node_ip:
                             LOGGER.error("Get nodes statuses. Failed to find a node in cluster by IP: %s", node_ip)
 
-        except Exception as exc:  # pylint: disable=broad-except
+        except Exception as error:  # pylint: disable=broad-except
             ClusterHealthValidatorEvent.NodeStatus(
                 severity=Severity.WARNING,
                 node=self.name,
-                message=f"Unable to get nodetool status from `{self.name}': {exc}",
+                message=f"Unable to get nodetool status from `{self.name}': {error=}",
             ).publish()
         return nodes_status
 


sdcm/cluster.py
```diff
@@ -2646,11 +2646,11 @@ def get_nodes_status(self):
                         if node_ip:
                             LOGGER.error("Get nodes statuses. Failed to find a node in cluster by IP: %s", node_ip)
 
-        except Exception as exc:  # pylint: disable=broad-except
+        except Exception as error:  # pylint: disable=broad-except
             ClusterHealthValidatorEvent.NodeStatus(
                 severity=Severity.WARNING,
                 node=self.name,
-                message=f"Unable to get nodetool status from `{self.name}': {exc}",
+                message=f"Unable to get nodetool status from `{self.name}': {error=}",
```
fruch, line 2653: how this is related to this change ?

sdcm/nemesis.py
```diff
@@ -176,6 +176,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     kubernetes: bool = False        # flag that signal that nemesis run with k8s cluster
     limited: bool = False           # flag that signal that nemesis are belong to limited set of nemesises
     has_steady_run: bool = False    # flag that signal that nemesis should be run with perf tests with steady run
+    cloud_sanity: bool = False      # flag signaling that nemesis should be run as part of siren-tests sanity longevity
```
fruch, line 179: why no set those from the siren-test code ? 
i.e. this is a filter that is used by the siren-test code, isn't more logical that the flags would be set by siren-test, and not in SCT ?

sdcm/cluster.py
ilya-rarov: This is something I hit into while looking at the logs. This is not related to the main purpose of the change, so I put it into a separate commit...

sdcm/nemesis.py
fruch: maybe just something like that in siren-test sct-plugin ?
```pyhton
from sdcm.nemesis import StopStartMonkey

StopStartMonkey.cloud_sanity = True
```

it's just need to be a property ontop of the class, the nemesis code doesn't really care how it got there.

I'm just saying that the labels should be defined in one place, not scattered between both. i.e. if I want to understand what part of `cloud_sanity` I should be looking at one code base, and not multiple of them.

ilya-rarov: I moved all the things related to the nemesis filter to `siren-tests`. Here I left only the commit for the exception printing fix.
Renamed the PR

sdcm/cluster.py
ilya-rarov: I updated the PR, so I believe this discussion is not relevant anymore

fruch (approved):
LGTM

PR #5535, opened 2022-12-07
test(longevity): new test of MV synchronous updates

juliayakovlev:
Add new longevity test of MV synchronous updates feature

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -84,7 +84,7 @@ from sdcm.utils.common import (
     normalize_ipv6_url,
     download_dir_from_cloud,
     generate_random_string,
-    prepare_and_start_saslauthd_service,
+    prepare_and_start_saslauthd_service
 )
 from sdcm.utils.ci_tools import get_test_name
 from sdcm.utils.distro import Distro
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -2234,7 +2234,7 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         self.log.debug('MV create statement: {}'.format(query))
         session.execute(query)
 
-    def _wait_for_view(self, scylla_cluster, session, key_space, view):
+    def _wait_for_view(self, scylla_cluster, session, key_space, view, wait_attempts=20):
         self.log.debug("Waiting for view {}.{} to finish building...".format(key_space, view))
 
         def _view_build_finished(live_nodes_amount):
@@ -2244,11 +2244,11 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
             self.log.debug('View build status result: {}'.format(result))
             return len([status for status in result if status[0] == 'SUCCESS']) >= live_nodes_amount
 
-        attempts = 20
+        attempts = wait_attempts
         nodes_status = scylla_cluster.get_nodetool_status()
         live_nodes_amount = 0
-        for dc in nodes_status.itervalues():
-            for ip in dc.itervalues():
+        for dc in nodes_status.values():
+            for ip in dc.values():
                 if ip['state'] == 'UN':
                     live_nodes_amount += 1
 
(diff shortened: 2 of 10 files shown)


sdcm/cluster.py
```diff
@@ -84,7 +84,7 @@
     normalize_ipv6_url,
     download_dir_from_cloud,
     generate_random_string,
-    prepare_and_start_saslauthd_service,
+    prepare_and_start_saslauthd_service
```
fruch, line 87: please don't remove those, it's on purpose

sdcm/utils/common.py
```diff
@@ -2803,3 +2851,47 @@ def __enter__(self):
     def __exit__(self, exit_type, value, _traceback):
         # remove the temporary folder as `sudo` to cover the case when the folder owner was changed during test
         self.node.remoter.sudo(f'rm -rf {self.folder_name}')
+
+
+def get_keyspace_partition_ranges(node, keyspace: str):
```
fruch, line 2856: Can we add unittest for those functions ?

sdcm/utils/common.py
```diff
@@ -2803,3 +2851,47 @@ def __enter__(self):
...
+
+
+def describering_parsing(describering_output):
+    def _list2dic(attr_list, heads_to_dict):
+        res = {}
+        for ind, attr in enumerate(heads_to_dict):
+            res[attr] = attr_list[ind].strip()
+        return res
+
+    found_attributes = re.findall(r'^\s*start_token:(-?\d+), end_token:(-?\d+), endpoints:\[([\d\., ]+)\], '
```
fruch, line 2885: please add an example, where you use regex

sdcm/utils/common.py
fruch: scratch that, now I see the unittest

sdcm/tester.py
```diff
@@ -2234,7 +2234,7 @@ def create_materialized_view(self, ks_name, base_table_name, mv_name, mv_partiti
         self.log.debug('MV create statement: {}'.format(query))
         session.execute(query)
 
-    def _wait_for_view(self, scylla_cluster, session, key_space, view):
+    def _wait_for_view(self, scylla_cluster, session, key_space, view, wait_attempts=20):
```
roydahan, line 2237: How do you know that 20 attempts is long enough?

juliayakovlev: 20 attemts is hardcoded in the code:
https://github.com/scylladb/scylla-cluster-tests/blob/892a54f2e802c9189be1cdf55cf18d2f4d2f84d4/sdcm/tester.py#L2257

I change it to be a flexible

juliayakovlev:
Work on this PR will be continue after https://github.com/scylladb/scylla-cluster-tests/pull/5782 will be completed

roydahan:
Is this something we still planning?

juliayakovlev:
> Is this something we still planning?

I think, it should be finished

fruch:
> > Is this something we still planning?
> 
> I think, it should be finished

I could be finished by the team responsible for MV, 

@juliayakovlev if can summarize a bit what is the purpose/goal of this PR, so whom should pick up, could decide on it's priority and plan accordingly

roydahan:
This is already in a PR stage so I prefer not throwing it on someone else.
If there is further work, we can move that further work to someone else.

fruch:
> This is already in a PR stage so I prefer not throwing it on someone else. If there is further work, we can move that further work to someone else.

it's almost 1y old PR, and the major part of it is to test it. (which wasn't done)

anyhow it's part of https://github.com/scylladb/qa-tasks/issues/839, which still have other sub-tasks of it open
whom gonna be responsible for the the whole feature, should consider which part he need and if this PR fits or not.

roydahan:
In case it's a responsibility that need to be transitioned, @juliayakovlev should work with @temichus who now own MVs and help with the transition.
Here it's a PR, so worth not losing the work that was already done and explain what's left.

fruch:
closing this one, if would be needed, someone would need to redo it anyhow

Issue #5538, opened 2022-12-07
disrupt_rebuild_streaming_err fails with: list index out of range


yarongilor:
```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3855, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3357, in disrupt_rebuild_streaming_err
    self.start_and_interrupt_rebuild_streaming()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3320, in start_and_interrupt_rebuild_streaming
    self._destroy_data_and_restart_scylla()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/group_common_events.py", line 243, in inner_func
    return func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1003, in _destroy_data_and_restart_scylla
    file_for_destroy = self._choose_file_for_destroy(tables)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 67, in inner
    return func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 940, in _choose_file_for_destroy
    ks_cf_for_destroy = random.choice(ks_cfs)  # expected value as: 'keyspace1.standard1'
  File "/usr/local/lib/python3.10/random.py", line 378, in choice
    return seq[self._randbelow(len(seq))]
IndexError: list index out of range
```

## Installation details

Kernel Version: 5.15.0-1026-aws
Scylla version (or git commit hash): `5.2.0~dev-20221201.c5121cf27365` with build-id `d25d298da80af2e947d1728190d8f4f26978e44a`

Cluster size: 4 nodes (i3en.2xlarge)

Scylla Nodes used in this run:
- longevity-twcs-48h-master-db-node-41fa60db-9 (34.227.61.71 | 10.12.8.50) (shards: 7)
- longevity-twcs-48h-master-db-node-41fa60db-8 (54.172.72.90 | 10.12.10.152) (shards: 7)
- longevity-twcs-48h-master-db-node-41fa60db-7 (54.227.114.85 | 10.12.9.165) (shards: 7)
[...]

PR #5540, opened 2022-12-07
fix(nemesis.py): fail nemesis if non-empty table not found

yarongilor:
fix(nemesis.py): fail nemesis if non-empty table not found
  
          If failed to find non-empty user tables, cannot run any
          data deletion operation and should fail nemesis due to table query issues.
          Fixes: #5538

https://github.com/scylladb/scylla-cluster-tests/issues/5538

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -50,6 +50,7 @@ from cassandra.cluster import Cluster as ClusterDriver  # pylint: disable=no-nam
 from cassandra.cluster import NoHostAvailable  # pylint: disable=no-name-in-module
 from cassandra.policies import RetryPolicy
 from cassandra.policies import WhiteListRoundRobinPolicy, HostFilterPolicy, RoundRobinPolicy
+from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
 from argus.db.cloud_types import ResourceState, CloudInstanceDetails, CloudResource
 from argus.db.db_types import NemesisStatus
@@ -3490,16 +3491,18 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
 
         return list(regular_table_names - materialized_view_table_names)
 
-    def is_table_has_data(self, session, table_name: str) -> bool:
+    def is_table_has_data(self, session, table_name: str) -> (bool, Optional[Exception]):
         """
-        Return True if `table_name` has data in it
+        Return True if `table_name` has data in it.
+        Checking if a result page with any data is received.
         """
-        current_rows = 0
         try:
-            current_rows = session.execute(f"SELECT * FROM {table_name} LIMIT 1").current_rows
+            result = session.execute(SimpleStatement(f"SELECT * FROM {table_name}", fetch_size=10))
+            return bool(len(result.one())), None
+
         except Exception as exc:  # pylint: disable=broad-except
             self.log.warning(f'Failed to get rows from {table_name} table. Error: {exc}')
-        return bool(current_rows)
+            return False, exc
 
     def get_all_tables_with_cdc(self, db_node: BaseNode) -> List[str]:
         """Return list of all tables with enabled cdc feature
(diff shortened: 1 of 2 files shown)


fruch:
@yarongilor how do you get into this situation ? that you run nemesis but doesn't have any data on your cluster ? 

why are you running nemesis on empty cluster ? what's the point ?

fruch (changes requested):
we should fix to propagate the actual errors, and not hide it and skip the nemesis

yarongilor:
> we should fix to propagate the actual errors, and not hide it and skip the nemesis

@fruch , does this approach mean we should remove the former check of `if not ks_cfs` as well?

fruch:
> > we should fix to propagate the actual errors, and not hide it and skip the nemesis
> 
> @fruch , does this approach mean we should remove the former check of `if not ks_cfs` as well?

`is_table_has_data` is boolean, should change to return boolean, and the error is there is one.
then the caller of this function can use propagate those error, if no table with data was found.

and in general we shouldn't be supporting running nemesis without any test data

yarongilor:
> > > we should fix to propagate the actual errors, and not hide it and skip the nemesis
> > 
> > 
> > @fruch , does this approach mean we should remove the former check of `if not ks_cfs` as well?
> 
> `is_table_has_data` is boolean, should change to return boolean, and the error is there is one. then the caller of this function can use propagate those error, if no table with data was found.
> 
> and in general we shouldn't be supporting running nemesis without any test data

fixed.

roydahan:
I don't think we should abort the nemesis, but to find the root cause and solve it.

fruch:
> I don't think we should abort the nemesis, but to find the root cause and solve it.

this PR would help exposing the failures, when they happens. till now they were not surfacing them, and the failures weren't clear

yarongilor:
> I don't think we should abort the nemesis, but to find the root cause and solve it.

@roydahan , commit description and PR title are rewritten.

sdcm/cluster.py
```diff
@@ -3490,16 +3491,22 @@ def execute_cmd(cql_session, entity_type):
...
+        Return True if `table_name` has data in it.
+        Checking if a result page with any data is received.
         """
-        current_rows = 0
+        error = None
         try:
-            current_rows = session.execute(f"SELECT * FROM {table_name} LIMIT 1").current_rows
+            query = f"SELECT * FROM {table_name}"
+            statement = SimpleStatement(query, fetch_size=10)
+            for paged_result in session.execute(statement):
```
yarongilor, line 3503: @roydahan , @fruch , i added a usage of paged-query here, (following https://github.com/scylladb/scylladb/issues/12268) since code is already modified.
There are other occurrences of non-paged queries in SCT that can be addressed in a separate PR.

sdcm/cluster.py
```diff
@@ -3490,16 +3491,22 @@ def execute_cmd(cql_session, entity_type):
...
+                return bool(len(paged_result.current_rows)), None
```
fruch, line 3504: there's no need for a loop, we care only for the first result.
```suggestion
            result = session.execute(SimpleStatement(f"SELECT * FROM {table_name}", fetch_size=10))
            return bool(len(result.one()), None
```

sdcm/cluster.py
fruch: did you tested it, and it solved the issue ? now that nemesis in that test case is working ?

sdcm/cluster.py
yarongilor: fixed and tested in:
https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/twcs-paged-query/3/

sdcm/cluster.py
yarongilor: https://github.com/scylladb/scylla-cluster-tests/pull/5555

sdcm/cluster.py
yarongilor: Test ran ok and nemesis passed ( 70 times)
```
ubuntu@ip-10-12-1-17:~/sct-results/latest$ grep sdcm.nemesis.RebuildStreamingErrMonkey: sct.log | grep duration | grep -i -v "'error':" -c
73
```

fruch (approved):
LGTM

PR #5547, opened 2022-12-08
fix(k8s-dynamic-loader): support mounting to the same dest

fruch:
when we have multiple stress command executed on the same loader but mounting to the same path

like in the cloud connectivty test, when we reuse the same loader but run multiple stress command that all of them try to copy the cloud bundle to the same path

and failed like the following:
```
LocalCmdRunner:base.py:222 The Pod "sct-loaders-0-pod-2" is invalid:
spec.containers[0].volumeMounts[1].mountPath: Invalid value:
"/tmp/tmp6cznrkfk.yaml": must be unique
```

now we identify there's a duplicate, and only update to the newest configmap

### DIFF
--- a/sdcm/remote/kubernetes_cmd_runner.py
+++ b/sdcm/remote/kubernetes_cmd_runner.py
@@ -603,11 +603,16 @@ class KubernetesPodRunner(KubernetesCmdRunner):
             for i, _ in enumerate(obj['spec']['containers']):
                 if 'volumeMounts' not in obj['spec']['containers'][i]:
                     obj['spec']['containers'][i]['volumeMounts'] = []
-                obj['spec']['containers'][i]['volumeMounts'].append({
-                    'name': cm_name,
-                    'mountPath': dst,
-                    'subPath': filename,
-                })
+                mount_list = obj['spec']['containers'][i]['volumeMounts']
+                existing_mount = [mount for mount in mount_list if mount['mountPath'] == dst]
+                if existing_mount:
+                    existing_mount[0]['name'] = cm_name
+                else:
+                    mount_list.append({
+                        'name': cm_name,
+                        'mountPath': dst,
+                        'subPath': filename,
+                    })
 
         # Add new modifier function to the list of pod template's modifiers
         self.template_modifiers.append(add_file_mount)


fruch:
@vponomaryov question (that I missed during the review), doesn't seem we have a code that clears this configmap, if we gonna use the dynamic loader on scylla-cloud, it might be an issue, filling a k8s cluster that isn't get killed after each test...

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -603,11 +603,16 @@ def add_file_mount(obj):
             for i, _ in enumerate(obj['spec']['containers']):
                 if 'volumeMounts' not in obj['spec']['containers'][i]:
                     obj['spec']['containers'][i]['volumeMounts'] = []
-                obj['spec']['containers'][i]['volumeMounts'].append({
-                    'name': cm_name,
-                    'mountPath': dst,
-                    'subPath': filename,
-                })
+                mount_list = obj['spec']['containers'][i]['volumeMounts']
```
vponomaryov, line 606: The current solution is wrong because it creates redundant resources in case of duplication and spams the pod configuration.

The proper solution would be to create a list of mounted files, look at it first, and if not found, then do all the 3 steps we have here:
- create config map
- attach it as a volume
- mount that volume to a file path.

And, if found (already mounted), do nothing.
Also, such approach assumes usage of a lock for whole method per class instance.

vponomaryov (changes requested):

vponomaryov:
> @vponomaryov question (that I missed during the review), doesn't seem we have a code that clears this configmap, if we gonna use the dynamic loader on scylla-cloud, it might be an issue, filling a k8s cluster that isn't get killed after each test...

K8S allows to delete all resources in a bunch by deleting it's namespace. So, just delete namespace.
And, moreover, even without this bug it must be done so. Test runs should not affect each other anyhow by the left resources.
For example, there is a limit for the pod number in a namespace. The dynamic loaders is designed to keep loaders.
So, the approach to reuse K8S cluster again and again assume usage of the temporary namespaces which get deleted after a test run.

vponomaryov:
Created a replacement PR with the proposed solution here: https://github.com/scylladb/scylla-cluster-tests/pull/5548

fruch:
> Created a replacement PR with the proposed solution here: https://github.com/scylladb/scylla-cluster-tests/pull/5548
> 

Close in favor of #5548

PR #5548, opened 2022-12-09
fix(k8s): make dynamic loader 'send_files' work correctly

vponomaryov:
For now, if we run more than 1 loader command on the same dynamic loader pod class and call `send_files` method in both cases with a same destination mount path then we get duplication failure.

To make it work start using a mapping object where we store paths.

New behaviour steps:
- Return from the method not doing anything if pair src/dst already exists.
- Raise ValueError if 'dst' exists with some different 'src'.


vponomaryov:
This is the solution for the problem that is described in another PR: https://github.com/scylladb/scylla-cluster-tests/pull/5547

Workability is tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/200/consoleFull

fruch (approved):

PR #5549, opened 2022-12-11
fix(k8s-dynamic-loader): stop assuming src/dest filename is identical

fruch:
in cases like:
```
  cmd_runner.send_files('/somedir/fileA.xml', '/tmp/fileB.xml')
```

would be failing the mount, since we configmap was create with `fileA.xml` as key, and the mount would look for `fileB.xml` as the key in the configmap

### DIFF
--- a/sdcm/remote/kubernetes_cmd_runner.py
+++ b/sdcm/remote/kubernetes_cmd_runner.py
@@ -588,7 +588,7 @@ class KubernetesPodRunner(KubernetesCmdRunner):
 
             # Create configMap from the 'src' file
             self.kluster.kubectl(
-                f"create configmap {cm_name} --from-file={src}",
+                f"create configmap {cm_name} --from-file={filename}={src}",
                 namespace=self.namespace)
 
             # Create modifier function for the pod template


vponomaryov (approved):

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -588,7 +588,7 @@ def send_files(self, src, dst, delete_dst=False, preserve_symlinks=False, verbos
 
             # Create configMap from the 'src' file
             self.kluster.kubectl(
-                f"create configmap {cm_name} --from-file={src}",
+                f"create configmap {cm_name} --from-file={filename}={src}",
```
soyacz, line 591: if the pattern with two `=` signs in `-from-file={filename}={src}` is ok (I don't know such details yet for k8s, and this looks strange) then I approve.

soyacz (approved):

sdcm/remote/kubernetes_cmd_runner.py
vponomaryov: Yes, it is correct. See this: https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#define-the-key-to-use-when-creating-a-configmap-from-a-file

fruch: one was doing `kubectl help create configmap` to figure it out:
```
      --from-file=[]: Key file can be specified using its file path, in which case file basename will be used as
configmap key, or optionally with a key and file path, in which case the given key will be used.  Specifying a directory
will iterate each named file in the directory whose basename is a valid configmap key.
...
Usage:
  kubectl create configmap NAME [--from-file=[key=]source] [--from-literal=key1=value1] [--dry-run=server|client|none]
[options]
```

Issue #5526, opened 2022-12-05
Rolling-upgrade c-s commands before significantly slower after move to docker based loaders


fruch:
@fgelcer noticed this in https://github.com/scylladb/scylla-cluster-tests/pull/5524#issuecomment-1336502023
> @roydahan , true that, but in [this job](https://jenkins.scylladb.com/job/scylla-master/job/rolling-upgrade/job/rolling-upgrade-ubuntu20.04-test/), until November 5th tests took 4.5 hours, and since, more than 5 hours (close to 6 hours, or more)

and @roydahan analysis:

> First, for ubuntu20 job that you gave as an example, we used scylla image as the base line, so the total time is shorter. (BTW!! we need to change now the base of ubuntu20 to be clean image and ubuntu22 to be based on image).
> 
> For ubuntu18, I see that in Nov 5th, it took 5h37m, so it's still much lower than what we have now. From a deep analysis I did [here](https://docs.google.com/spreadsheets/d/1abgVTEJlyN1yl8DkmoEj-NohM9jJ3m5nRlGsdkb1YBg/edit?usp=sharing), I can think of 3 options:
> 
> 1. Severe performance degredation in Scylla
> 2. Regression in c-s tool we use.
> 3. Slow instances.
> 
> The details that supports my analysis:
> 
> Failed test cassandra-stress results:
> 
> ```
> egrep -e 'Total operation time' -e 'Total partitions' -e 'Op rate' sct.log
> Op rate                   :    3,412 op/s  [insert: 3,423 op/s]
> Total partitions          :  5,000,000 [insert: 5,000,000]
> Total operation time      : 00:24:25
> Op rate                   :    7,814 op/s  [WRITE: 7,814 op/s]
> Total partitions          : 20,100,200 [WRITE: 20,100,200]
> Total operation time      : 00:42:52
> Op rate                   :    6,579 op/s  [WRITE: 6,579 op/s]
> Total partitions          : 10,100,200 [WRITE: 10,100,200]
> Total operation time      : 00:25:35
> Op rate                   :    9,548 op/s  [READ: 9,548 op/s]
> Total partitions          : 10,100,200 [READ: 10,100,200]
> Total operation time      : 00:17:37
> Op rate                   :    7,343 op/s  [READ: 7,343 op/s]
> Total partitions          :  4,415,382 [READ: 4,415,382]
> Total operation time      : 00:10:01
[...]

PR #5553, opened 2022-12-12
fix(rolling-upgrade): enable use_prepared_loaders

fruch:
Since we were seen issue with the docker based loader,
and unclear performence and timing effect on the rolling
upgrade test, we should be using the fallback to VMs now

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/5526


sdcm/cluster.py
```diff
@@ -468,6 +469,18 @@ def system_log(self):
...
 
+    @property
+    def cassandra_stress_version(self):
+        if not self._cassandra_stress_version:
+            result = self.remoter.run(cmd="cassandra-stress version", ignore_status=True, verbose=True)
+            match = re.search("Version: (.*)", result.stdout)
+            if match:
+                self._cassandra_stress_version = match.group(1)
+            else:
+                self.log.error("C-S version not found!")
```
soyacz, line 480: add `result.stdout` to error message for quicker issue investigation.

sdcm/cluster.py
```diff
@@ -4812,6 +4875,36 @@ def _parse_cs_summary(lines):
             return {}
         return results
 
+    @staticmethod
+    def _parse_cs_results(lines):
```
soyacz, line 4879: where is it used?

fruch: it's not, it a revert of a commit which I've deleted it. (the revert if cause of different reason)

sdcm/cluster.py
fruch: I need to remove this, it's not actually being used, it's back cause it's part of a revert I'm doing in this PR

the revert is to return the code that install c-s on-top of the loaders, (it was remove as part of switching to docker based loaders)

sdcm/cluster.py
fruch: I'll remove it from the revert

sdcm/cluster.py
fruch: I'll remove it from the revert

roydahan (approved):

PR #5556, opened 2022-12-12
fix(logcollector): change order of collection

fruch:
since getting sct.log is more importent then
trying to collect melformed sstables, we should
first start with the must have collector, and then move to the less frequent used.

since in some cases we've seen the collection of sstables were stack, probably since it trying to read all of the events and it was a 3day job with lots of events.

### DIFF
--- a/sdcm/logcollector.py
+++ b/sdcm/logcollector.py
@@ -1254,18 +1254,22 @@ class Collector:  # pylint: disable=too-many-instance-attributes,
         self.sct_set = []
         self.pt_report_set = []
         self.cluster_log_collectors = {
-            SSTablesCollector: self.db_cluster,
             ScyllaLogCollector: self.db_cluster,
-            MonitorLogCollector: self.monitor_set,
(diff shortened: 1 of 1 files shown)


sdcm/logcollector.py
```diff
@@ -1254,12 +1254,12 @@ def __init__(self, test_id=None, test_dir=None, params=None):
         self.sct_set = []
         self.pt_report_set = []
         self.cluster_log_collectors = {
-            SSTablesCollector: self.db_cluster,
             ScyllaLogCollector: self.db_cluster,
+            SCTLogCollector: self.sct_set,
```
vponomaryov, line 1258: Please, add also "k8s log collector"  to be after the SCT and before the monitoring. It is also fast and valueable to have earlier.

roydahan (approved):

sdcm/logcollector.py
vponomaryov: Actually, the monitor log collection stayed before as it was. Ok, merged anyway...

PR #5562, opened 2022-12-13
Export nemesis

roydahan:


unit_tests/test_nemesis.py
```diff
@@ -223,3 +223,34 @@ def test_use_disabled_monkey():
     collected_disrupt_methods_names = {disrupt.__name__ for disrupt in sisyphus.disruptions_list}
 
     assert 'disrupt_toggle_table_gc_mode' in collected_disrupt_methods_names
+
+
+def test_list_all_available_nemesis(export=False):
+    import pickle
```
fruch, line 229: I don't think pickle is needed here, just use json or yaml

unit_tests/test_nemesis.py
```diff
@@ -223,3 +223,34 @@ def test_use_disabled_monkey():
...
+
+    tester = FakeTester()
+
+    tester.params["nemesis_seed"] = '1'
+    tester.params["nemesis_selector"] = []
```
fruch, line 234: as I've said we probably want to add the flag to the data, so selector would also work

unit_tests/test_nemesis.py
```diff
@@ -223,3 +223,34 @@ def test_use_disabled_monkey():
...
+    sisyphus = SisyphusMonkey(tester, None)
+
+    collected_disrupt_methods_names = {disrupt.__name__ for disrupt in sisyphus.disruptions_list}
+
+    assert len(collected_disrupt_methods_names) == 67
+
+    if export:
+        # Export the list of nemesis to a static file for later usage
+        with open("test_data/nemesis_list", "wb") as fp:
```
fruch, line 243: the location of this file, shouldn't be in the unittest test data, but in `data_dir`

also we should consider having the path as part of the `SisyphusMonkey` class, and can be overwritten, for example for siren-tests, which have a bit different list  and flags, since it has more Nemesis

unit_tests/test_nemesis.py
```diff
@@ -223,3 +223,34 @@ def test_use_disabled_monkey():
...
+
+    tester.params["nemesis_seed"] = '1'
+    tester.params["nemesis_selector"] = []
+    sisyphus = SisyphusMonkey(tester, None)
+
+    collected_disrupt_methods_names = {disrupt.__name__ for disrupt in sisyphus.disruptions_list}
+
+    assert len(collected_disrupt_methods_names) == 67
+
+    if export:
```
fruch, line 241: instead of parameter for I would turn it into a hydra command:

```
hydra regenerate-nemesis-list
```

unit_tests/test_nemesis.py
```diff
@@ -223,3 +223,34 @@ def test_use_disabled_monkey():
...
+
+    tester = FakeTester()
+
+    tester.params["nemesis_seed"] = '1'
+    tester.params["nemesis_selector"] = []
+    sisyphus = SisyphusMonkey(tester, None)
+
+    collected_disrupt_methods_names = {disrupt.__name__ for disrupt in sisyphus.disruptions_list}
+
+    assert len(collected_disrupt_methods_names) == 67
```
fruch, line 239: I wouldn't state the numbers in this test, just make this test fail if the generated and the static data doesn't match

unit_tests/test_nemesis.py
```diff
@@ -223,3 +223,34 @@ def test_use_disabled_monkey():
...
+
+        # Print also a human readable file
+        nemesis_name = ["{}\n".format(i) for i in collected_disrupt_methods_names]
+        with open('test_data/nemesis_list.txt', 'w') as fp:
+            fp.writelines(nemesis_name)
+
+    with open('test_data/nemesis_list', 'rb') as fp:
+        static_nemesis_list = pickle.load(fp)
+
+    assert static_nemesis_list == collected_disrupt_methods_names
```
fruch, line 254: I would use set on both ends, the outcome of the diff would be a bit clearer.

data_dir/nemesis.yml
```diff
@@ -0,0 +1,253 @@
+- disrupt_abort_repair:
```
k0machi, line 1: How would `nemesis.yml` file map with `nemesis_classes.yml`? Or is there no need for that?

roydahan: There isn't a need for that, so I decided to keep them decoupled.

roydahan: You need the nemesis_classes.

roydahan:
@fruch please review.
I have a side branch to create a command that implements "hydra generate-report" but it's a bit ugly now, so for now without it.

k0machi (approved):

fruch:
> @fruch please review.
> I have a side branch to create a command that implements "hydra generate-report" but it's a bit ugly now, so for now without it.
> 

Looks o.k., but seems like still missing the ability to use the exported file,
I.e. merging/backporting this alone doesn't do anything.

roydahan:
> > @fruch please review.
> > I have a side branch to create a command that implements "hydra generate-report" but it's a bit ugly now, so for now without it.
> 
> Looks o.k., but seems like still missing the ability to use the exported file, I.e. merging/backporting this alone doesn't do anything.

Right.
There going to be 2 follow up PRs on top of it.
One from @k0machi for the pipeline generator for all nemesis.
Another one from me to allow sisyphus read from this static file.

fruch (approved):
LGTM

PR #5565, opened 2022-12-14
feature(chaos-mesh): added installation of chaos-mesh

soyacz:
We want to use chaos-mesh for chaos testing k8s clusters.

This PR adds:
* installation of chaos-mesh in k8s cluster for future use.
* example experiment (pod failure) and functional test using it

### DIFF
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -36,6 +36,7 @@ from sdcm.utils.k8s import (
     convert_cpu_value_from_k8s_to_units,
     HelmValues,
 )
+from sdcm.utils.k8s.chaos_mesh import PodFailureExperiment
 
 from functional_tests.scylla_operator.libs.helpers import (
     get_scylla_sysctl_value,
@@ -860,3 +861,13 @@ def test_cloud_bundle_connectivity_cqlsh(db_cluster: ScyllaPodCluster):
 
     assert not res.stderr
     assert '(1 rows)' in res.stdout
+
+
+def test_can_recover_from_fatal_pod_termination(db_cluster):
+    target_node = db_cluster.nodes[-1]
+    experiment = PodFailureExperiment(pod=target_node, duration="60s")
+    experiment.start()
+    db_cluster.k8s_cluster.kubectl(
+        f"wait --timeout=1m --for=condition=Ready=false pod {target_node.name} -n {db_cluster.namespace}")
+    experiment.wait_until_finished()
+    db_cluster.wait_for_nodes_up_and_normal(nodes=[target_node], verification_node=target_node)
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -2808,3 +2808,11 @@ class RemoteTemporaryFolder:
     def __exit__(self, exit_type, value, _traceback):
         # remove the temporary folder as `sudo` to cover the case when the folder owner was changed during test
         self.node.remoter.sudo(f'rm -rf {self.folder_name}')
+
+
+duration_pattern = re.compile(r'(?P<hours>[\d]*)h|(?P<minutes>[\d]*)m|(?P<seconds>[\d]*)s')
+
+
+def time_period_str_to_seconds(time_str: str) -> int:
+    """Transforms duration string into seconds int. e.g. 1h -> 3600, 1h22m->4920 or 10m->600"""
+    return sum([int(g[0] or 0) * 3600 + int(g[1] or 0) * 60 + int(g[2] or 0) for g in duration_pattern.findall(time_str)])
--- a/sdcm/utils/k8s/__init__.py
+++ b/sdcm/utils/k8s/__init__.py
@@ -557,7 +557,9 @@ class HelmContainerMixin:
                                  use_devel: bool = False,
                                  debug: bool = True,
                                  values: 'HelmValues' = None,
-                                 namespace: Optional[str] = None) -> str:
+                                 namespace: Optional[str] = None,
+                                 atomic: bool = False,
+                                 timeout: Optional[str] = None) -> str:
         command = [operation_type, target_chart_name, source_chart_name]
         prepend_command = []
         if version:
@@ -566,7 +568,15 @@ class HelmContainerMixin:
             command.extend(("--devel",))
         if debug:
             command.extend(("--debug",))
-
+        if atomic:
+            # if set, the installation process deletes the installation on failure
+            # or upgrade process rolls back changes made in case of failed upgrade
+            # waits for operation completion - so should specify timeout.
+            command.extend(("--atomic",))
+            if timeout is None:
+                LOGGER.warning("No timeout provided for atomic operation, default of '5m' is used.")
+            else:
+                command.extend(("--timeout", timeout))
         return self.helm(
             kluster,
             *command,
--- a/sdcm/utils/k8s/chaos_mesh.py
+++ b/sdcm/utils/k8s/chaos_mesh.py
@@ -1,8 +1,50 @@
-import logging
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2022 ScyllaDB
 
+import ast
+import json
+import logging
+from datetime import datetime
+from enum import Enum
+from tempfile import NamedTemporaryFile
+import time
+
+import yaml
+from botocore.utils import deep_merge
+
+from sdcm.utils.common import time_period_str_to_seconds
 from sdcm.utils.k8s import HelmValues, get_helm_pool_affinity_values
 
 LOGGER = logging.getLogger(__name__)
+
+
+class ChaosMeshException(Exception):
+    pass
+
+
+class PodChaosException(ChaosMeshException):
+
+    def __init__(self, msg: str, k8s_cluster: "KubernetesCluster", podchaos_name: str, namespace: str):
+        super().__init__(msg)
+        self.message = msg
+        self.details = k8s_cluster.kubectl(f"describe podchaos {podchaos_name} -n {namespace}").stdout
+
+    def __str__(self):
+        return f"{self.message}. Details:\n" + self.details
+
+
+class PodChaosTimeout(PodChaosException):
+    pass
 
 
 class ChaosMesh:  # pylint: disable=too-few-public-methods
@@ -44,3 +86,114 @@ class ChaosMesh:  # pylint: disable=too-few-public-methods
             timeout="30m"
         )
         LOGGER.info("chaos-mesh installed successfully on %s k8s cluster.", self._k8s_cluster.k8s_scylla_cluster_name)
+
+
+class ExperimentStatus(Enum):
+    STARTING = 0
+    RUNNING = 1
+    PAUSED = 2
+    FINISHED = 3
+    ERROR = 4
+    UNKNOWN = 5
+
+
+class PodChaosExperiment:
+    """Base class for all PodChaos experiments."""
+    API_VERSION = "chaos-mesh.org/v1alpha1"
+
+    def __init__(self, pod: "BasePodContainer", name: str, timeout: int = 0):
+        self._k8s_cluster = pod.parent_cluster.k8s_cluster
+        self._name = name
+        self._namespace = pod.parent_cluster.namespace
+        self._experiment = {
+            "apiVersion": self.API_VERSION,
+            "metadata": {
+                "name": self._name,
+                "namespace": self._namespace
+            },
+            "spec": {
+                "mode": "one",
+                "selector": {
+                    "labelSelectors": {
+                        "statefulset.kubernetes.io/pod-name": pod.name
+                    }
+                }
+            }
+        }
+        self._timeout: int = timeout
+        self._end_time: int = 0
+
+    def start(self):
+        """Starts experiment. Does not wait for finish."""
+        LOGGER.debug("Starting a pod-failure experiment %s", self._name)
+        assert self._k8s_cluster, "K8s cluster hasn't been configured for this experiment."
+        with NamedTemporaryFile(suffix=".yaml", mode="w") as experiment_config_file:
+            yaml.dump(self._experiment, experiment_config_file)
+            experiment_config_file.flush()
+            self._k8s_cluster.apply_file(experiment_config_file.name)
+        LOGGER.info("pod-failure experiment '%s' has started", self._name)
+        self._end_time = time.time() + self._timeout
+
+    def get_status(self) -> ExperimentStatus:
+        """Gets status of podchaos experiment."""
+        result = self._k8s_cluster.kubectl(
+            f"get podchaos {self._name}  -n {self._namespace} -o jsonpath='{{.status.conditions}}'", verbose=False)
+        condition = {cond["type"]: ast.literal_eval(cond["status"]) for cond in json.loads(result.stdout)}
+        if not condition["Selected"] and condition["Paused"]:
+            return ExperimentStatus.ERROR
+        if not condition["Selected"] and not condition["Paused"] and not condition["AllRecovered"] and not condition["AllInjected"]:
+            return ExperimentStatus.STARTING
+        if condition["Selected"] and not condition["Paused"] and not condition["AllRecovered"] and condition["AllInjected"]:
+            return ExperimentStatus.RUNNING
+        if condition["AllRecovered"] and condition["Paused"]:
+            return ExperimentStatus.PAUSED
+        if condition["AllRecovered"] and not condition["Paused"]:
+            return ExperimentStatus.FINISHED
+        LOGGER.warning("Unknown experiment status: %s", condition)
+        return ExperimentStatus.UNKNOWN
+
+    def wait_until_finished(self):
+        """Waits given timeout seconds for experiment to finish.
+
+        In case of experiment status being an error or timeout occurred, raises an exception."""
+        LOGGER.debug("waiting until '%s' experiment ends...", self._name)
+        assert self._end_time, "Experiment was not started. Use 'start()' method before waiting."
+        while time.time() < self._end_time:
+            status = self.get_status()
+            if status == ExperimentStatus.FINISHED:
+                LOGGER.debug("'%s' experiment ended.", self._name)
+                return
+            elif status == ExperimentStatus.ERROR:
+                raise PodChaosException(msg="Experiment status error",
+                                        k8s_cluster=self._k8s_cluster,
+                                        podchaos_name=self._name,
+                                        namespace=self._namespace)
+            time.sleep(2)
+        raise PodChaosTimeout(msg="Timeout when waiting for ChaosMesh experiment to complete.",
+                              k8s_cluster=self._k8s_cluster,
+                              podchaos_name=self._name,
+                              namespace=self._namespace)
+
+
+class PodFailureExperiment(PodChaosExperiment):
+    """
+    This experiment works by replacing container image with dummy image.
+    Then it waits for specified duration and rolls back image config.
+    """
+
+    def __init__(self, pod: "BasePodContainer", duration: str):
+        """Injects fault into a specified Pod to make the Pod unavailable for a period of time.
+
+        'duration' is str type in k8s notation. E.g. 10s, 5m
+        """
+        # timeout based on duration + 10 seconds margin
+        timeout = time_period_str_to_seconds(duration) + 10
+        super().__init__(
+            pod=pod, name=f"pod-failure-{pod.name}-{datetime.now().strftime('%d-%H.%M.%S')}", timeout=timeout)
+        deep_merge(self._experiment, {
+            "kind": "PodChaos",
+            "spec": {
+                "action": "pod-failure",
+                "duration": duration,
+            }
+        })
--- a/sdcm/utils/k8s/chaos_mesh.py
+++ b/sdcm/utils/k8s/chaos_mesh.py
@@ -1,0 +1,46 @@
+import logging
+
+from sdcm.utils.k8s import HelmValues, get_helm_pool_affinity_values
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ChaosMesh:  # pylint: disable=too-few-public-methods
+    NAMESPACE = "chaos-mesh"
+    VERSION = "2.5.0"
+    HELM_SETTINGS = {
+        'dashboard.create': True,
+        'dnsServer.create': True
+    }
+
+    def __init__(self, k8s_cluster: "KubernetesCluster"):
+        self._k8s_cluster = k8s_cluster
+
+    def initialize(self) -> None:
+        """Installs chaos-mesh on k8s cluster and prepares for future k8s chaos testing."""
+        if self._k8s_cluster.kubectl(f"get ns {self.NAMESPACE}", ignore_status=True).ok:
+            LOGGER.info("Chaos Mesh is already installed. Skipping installation.")
+            return
+        LOGGER.info("Installing chaos-mesh on %s k8s cluster...", self._k8s_cluster.k8s_scylla_cluster_name)
+        self._k8s_cluster.helm("repo add chaos-mesh https://charts.chaos-mesh.org")
+        self._k8s_cluster.helm('repo update')
+        self._k8s_cluster.kubectl(f"create namespace {self.NAMESPACE}")
+        aux_node_pool_affinity = get_helm_pool_affinity_values(
+            self._k8s_cluster.POOL_LABEL_NAME, self._k8s_cluster.AUXILIARY_POOL_NAME)
+        scylla_node_pool_affinity = get_helm_pool_affinity_values(
+            self._k8s_cluster.POOL_LABEL_NAME, self._k8s_cluster.SCYLLA_POOL_NAME)
+        self._k8s_cluster.helm_install(
+            target_chart_name="chaos-mesh",
+            source_chart_name="chaos-mesh/chaos-mesh",
+            version=self.VERSION,
+            use_devel=False,
+            namespace=self.NAMESPACE,
+            values=HelmValues(self.HELM_SETTINGS | {
+                "chaosDaemon": scylla_node_pool_affinity,
+                "controllerManager": aux_node_pool_affinity,
+                "dnsServer": aux_node_pool_affinity
+            }),
+            atomic=True,
+            timeout="30m"
+        )
+        LOGGER.info("chaos-mesh installed successfully on %s k8s cluster.", self._k8s_cluster.k8s_scylla_cluster_name)
--- a/test-cases/scylla-operator/functional.yaml
+++ b/test-cases/scylla-operator/functional.yaml
@@ -30,3 +30,5 @@ k8s_functional_test_dataset: MULTI_COLUMNS_DATA
 
 # enable by default
 k8s_enable_tls: true
+
+k8s_use_chaos_mesh: true
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -337,3 +337,5 @@
 | **<a href="#user-content-max_deviation" name="max_deviation">max_deviation</a>**  | Max relative difference between best and current throughput,<br>if current throughput larger then best on max_rel_diff, it become new best one | N/A | SCT_MAX_DEVIATION
 | **<a href="#user-content-n_stress_process" name="n_stress_process">n_stress_process</a>**  | Number of stress processes per loader | N/A | SCT_N_STRESS_PROCESS
 | **<a href="#user-content-stress_process_step" name="stress_process_step">stress_process_step</a>**  | add/remove num of process on each round | N/A | SCT_STRESS_PROCESS_STEP
+| **<a href="#user-content-use_hdr_cs_histogram" name="use_hdr_cs_histogram">use_hdr_cs_histogram</a>**  | Enable hdr histogram logging for cs | N/A | SCT_USE_HDR_CS_HISTOGRAM
+| **<a href="#user-content-k8s_use_chaos_mesh" name="k8s_use_chaos_mesh">k8s_use_chaos_mesh</a>**  | enables chaos-mesh for k8s testing | N/A | SCT_K8S_USE_CHAOS_MESH
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -84,6 +84,7 @@ from sdcm.utils.k8s import (
 )
 from sdcm.utils.decorators import log_run_info, retrying
 from sdcm.utils.decorators import timeout as timeout_wrapper
+from sdcm.utils.k8s.chaos_mesh import ChaosMesh
 from sdcm.utils.remote_logger import get_system_logging_thread, CertManagerLogger, ScyllaOperatorLogger, \
     KubectlClusterEventsLogger, ScyllaManagerLogger, KubernetesWrongSchedulingLogger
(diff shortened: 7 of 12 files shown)


sdcm/chaos_mesh.py
```diff
@@ -0,0 +1,62 @@
+import json
```
fruch, line 1: also there a directory for k8s related code `sdcm.cluster_k8s`, should be in there, I think...

sdcm/chaos_mesh.py
```diff
@@ -0,0 +1,62 @@
...
+            target_chart_name="chaos-mesh",
+            source_chart_name="chaos-mesh/chaos-mesh",
+            version=self.VERSION,
+            use_devel=False,
+            namespace=self.NAMESPACE,
+            values=HelmValues(self.HELM_SETTINGS),
+            atomic=True,
+            timeout="20m"
+        )
+        self._wait_for_chaos_mesh_readiness(timeout=20)
```
fruch, line 44: the time out is in seconds ? or minutes ?, bare timeout across the code is in seconds. 
while `readiness_timeout` used inside the function is in minutes.

sdcm/chaos_mesh.py
```diff
@@ -0,0 +1,62 @@
...
+    def initialize(self, k8s_cluster: KubernetesCluster) -> None:
+        """Installs chaos-mesh on k8s cluster and prepares for future k8s chaos testing."""
+        if self._k8s_cluster is not None:
+            LOGGER.warning("Chaos Mesh is already initialized. It should have not been initialized twice.")
+            return
+        self._k8s_cluster = k8s_cluster
+        self._install()
+
+    def _install(self):
+        if self._k8s_cluster.kubectl("kubectl get ns chaos-mesh", ignore_status=True).ok:
```
fruch, line 27: namespace isn't enough, we should wait

sdcm/chaos_mesh.py
```diff
@@ -0,0 +1,62 @@
...
+from sdcm.cluster_k8s import KubernetesCluster
+from sdcm.utils.k8s import HelmValues, KubernetesOps
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ChaosMesh:  # pylint: disable=too-few-public-methods
+    NAMESPACE = "chaos-mesh"
+    VERSION = "2.5.0"
+    HELM_SETTINGS = {'dashboard.create': True, 'dnsServer.create': True}
```
fruch, line 13: dashboard I understand

what's the dns server for ?

fruch:
> still under testing, I'll let know when ready for merge.
> 
> I wonder if we should wait for installation complete or move it to background (and later try if up and running before running some experiment). What's your opinion?

that's depends, on how long it takes to install.
anyhow we will need to wait for it, before using it.

sdcm/chaos_mesh.py
soyacz: it's enablement for dns server faults - I can disable it for now as it's not needed yet.

sdcm/chaos_mesh.py
soyacz: I followed pattern from readiness function - it's in minutes. I'll rename it to be clear.

sdcm/chaos_mesh.py
```diff
@@ -0,0 +1,62 @@
...
+            kluster=self._k8s_cluster,
+            total_pods=total_pods,
+            readiness_timeout=timeout,
+            namespace="chaos-mesh"
+        )
+        LOGGER.info("chaos-mesh pods are running.")
+
+
+# used as singleton - import 'chaos_mesh' instead of ChaosMesh
+chaos_mesh = ChaosMesh()
```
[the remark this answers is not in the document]
soyacz: I was thinking about it, but decided to do it in setUp directly as I'd need to implement it in all `get_cluster_k8s...` functions - while it should not depend on any k8s backend. Also it is more related to testing than k8s cluster and hiding it somewhere in deployment for me was not looking good.

Btw. Is this possible for scylla to communicate across multiple clusters? Even if we support multiregion, won't it be in one k8s cluster spreading across regions?

Btw. this is a very lazy way of creating singleton and in case of multiple clusters can be refactored easily.

sdcm/chaos_mesh.py
soyacz: I was looking there, and files there are related more to k8s backends, not test related stuff.
@vponomaryov  what is your opinion?

sdcm/chaos_mesh.py
soyacz: number of pods for chaos-mesh varies per configuration and number of k8s nodes. It's not easy to predict this number and use this function.
Here I assumed install procedure always succeeds, so verifying existence of chaos-mesh ns would be enough.

fruch: If it's hard to predict, how do you do that in the end of this function ? I.e. you should have the information, anyhow doesn't the documentation of chaos mesh suggests a different way to wait for it to be ready ?

Also assuming namespace available, mean it's fully installed, is that cause of the atomic variable?

vponomaryov: At first, we can check that all the pods we have in the namespace are ready and running. There must not be dead pods.
At second, helm chart must support setting number of replicas, so, lets configure it explicitly.
At third. We can read info about replicas number and use it for checking pods readiness.

My choice would be either 2nd or 3rd variant.

But some of them must be done here.

sdcm/chaos_mesh.py
vponomaryov: Please, define the config hierarchically. So, it will be easier to update it later with minimal code line changes.

sdcm/chaos_mesh.py
vponomaryov: I would place it at `sdcm/utils/k8s.py`.
Almost all the utility code for K8S is there. 
This one is not that big to be somewhere else.

It is definitely not needed in the main `sdcm` catalog.

sdcm/chaos_mesh.py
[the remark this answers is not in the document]
vponomaryov: It does depend on the K8S clusters. The helm chart installs it directly to a K8S cluster.
At some milestone of serverless we will have multi-region support which assumes having multiple K8S clusters.

I would expect to have chaos-mesh agent be in each K8S cluster.

So, in summary, having direct relation to each separate K8S cluster we can already make it be part of the appropriate class.

vponomaryov (changes requested):

sdcm/chaos_mesh.py
soyacz: In the end, in `_wait_for_chaos_mesh_readiness` I get the number of scheduled pods and wait for that number of pods to be ready. But this approach may fail in case of pods to be scheduled later, after I get number of pods - I was afraid of this scenario and decided to use `atomic` helm argument. And I find this 'atomic' approach a correct one, as we utilize helm function to wait for chart to be ready, instead of a bit hacky `wait_for_pods_readiness` function.

sdcm/chaos_mesh.py
soyacz: For now it is small, but in the next steps we'll be adding various experiments and their execution code. Things will grow and keeping them in separate module will ease development.

I'll move it to sdcm/utils/chaos-mesh.py

sdcm/chaos_mesh.py
soyacz: what do you mean by 'hierarchically'?

sdcm/chaos_mesh.py
soyacz: > If it's hard to predict, how do you do that in the end of this function ? I.e. you should have the information, anyhow doesn't the documentation of chaos mesh suggests a different way to wait for it to be ready ?
> 
> Also assuming namespace available, mean it's fully installed, is that cause of the atomic variable?

In the end, in _wait_for_chaos_mesh_readiness I get the number of scheduled pods and wait for that number of pods to be ready. But this approach may fail in case of pods to be scheduled later, after I get number of pods - I was afraid of this scenario and decided to use atomic helm argument. And I find this 'atomic' approach a correct one, as we utilize helm function to wait for chart to be ready, instead of a bit hacky wait_for_pods_readiness function.


> At first, we can check that all the pods we have in the namespace are ready and running. There must not be dead pods. At second, helm chart must support setting number of replicas, so, lets configure it explicitly. At third. We can read info about replicas number and use it for checking pods readiness.
> 
> My choice would be either 2nd or 3rd variant.
> 
> But some of them must be done here.

I don't think chaos-mesh's helm chart support replicas - there's one daemon pod for each k8s node and some pods existence depends on configuration (like dashboard). I think current approach of making sure all the pods in chaos-mesh namespace are in running state is enough. Do you know the case when it will not work?

sdcm/chaos_mesh.py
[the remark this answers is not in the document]
soyacz: I'm reluctant to grow k8s cluster classes. I'd rather keep things as pluggable modules. 
In case of multiple clusters, we could easily modify `ChaosMesh` class to install chaos-mesh on all provided k8s clusters and remember on which one was installed.

Later when running experiments, we could just provide cluster on which we want to run it as argument.

With this approach we may stop modifying our classes indefinitely and focus of adding new ones with single responsibilities - which is more maintainable way.

sdcm/chaos_mesh.py
vponomaryov: Need to specify the affinity rules for the `chaos-mesh` helm chart. 
After it we will know for sure the number of K8S nodes where it must land. And, as a result, we know for sure the number of pods.

sdcm/chaos_mesh.py
vponomaryov: > what do you mean by 'hierarchically'?

```
HELM_SETTINGS = {
    'dashboard.create': True,
    'dnsServer.create': True,
}
```

sdcm/chaos_mesh.py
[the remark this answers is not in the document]
vponomaryov: Saying be part of a class I meant following:
```
from sdcm.utils.chaos_mesh import ChaosMesh

class KubernetesCluster():
    def __init__(self):
        self.chaos_mesh = ChaosMesh()
```

soyacz:
Staging job on eks:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/operator-functional-eks/

sdcm/chaos_mesh.py
vponomaryov: I don't mind having it be split into `sdcm/utils/k8s/foo.py`. But, it's place %somewhere% there.

sdcm/chaos_mesh.py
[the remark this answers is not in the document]
soyacz: > Saying be part of a class I meant following:
> 
> ```
> from sdcm.utils.chaos_mesh import ChaosMesh
> 
> class KubernetesCluster():
>     def __init__(self):
>         self.chaos_mesh = ChaosMesh()
> ```

Yes, It looks easy, but it breaks good programming principles:
1. Single responsibility principle - KubernetesCluster class will be responsible for chaos-mesh
2. It creates dependencies inside own code - tight coupling
3.  Later, when using it, we'll do `self.cluster.k8s_cluster.chaos_mesh.run_experiment(...)` which also breaks some good design principle (Demeter law)

I'm opting for such approach:
`chaos_mesh.run_experiment(self.cluster.k8s_cluster, experiment)`

sdcm/chaos_mesh.py
soyacz: I think this is handled by chart itself - somehow it installed daemons automatically on all the nodes without setting it.
I'll do more testing, if works as is, maybe we don't need to specify it?

vponomaryov: It installs on all the nodes. We do not need to have it on all the nodes. We are going to apply it only on Scylla ones.
Actually, it is even easier -> Specify affinity rule for Scylla nodes and use "n_db_nodes" option for getting number of expected pods.
There are a lot of examples for adding affinity rules in the `sdcm/cluster_k8s/__init__.py`.

sdcm/chaos_mesh.py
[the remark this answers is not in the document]
soyacz: Still, because it's installed in k8s cluster doesn't mean it must be part of `KubernetesCluster` class - we can inject instance of it to chaos-mesh and have loose coupling.
I just try to avoid the mess like we have in `Cluster` class where we do everything. Easy implementation is not equal easy maintenance (over time rather opposite).

sdcm/chaos_mesh.py
soyacz: I'll move it under sdcm/utils/k8s/

sdcm/chaos_mesh.py
soyacz: I don't see the benefit here - these pods are very light, installation of this chart with wait for pods to be up took 38 seconds on EKS. Adding affinity logic adds complexity. And still 'n_db_nodes' is not everything, there are chaos-controller-managers and optional dashboard (which I presume will be disabled soon - I keep it just for playing with it)

sdcm/utils/chaos_mesh.py
```diff
@@ -0,0 +1,62 @@
...
+from sdcm.cluster_k8s import KubernetesCluster
+from sdcm.utils.k8s import HelmValues, KubernetesOps
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ChaosMesh:  # pylint: disable=too-few-public-methods
+    NAMESPACE = "chaos-mesh"
+    VERSION = "2.5.0"
+    HELM_SETTINGS = {'dashboard.create': True, 'dnsServer.create': True}
```
vponomaryov, line 13: It must have `affinity` field with the rule for the Scylla node pools.
Instance of the `KubernetesCluster` has `POOL_LABEL_NAME` and `SCYLLA_POOL_NAME` attributes which must be used in the another K8S utility like in the following example:

```
from sdcm.utils.k8s import get_helm_pool_affinity_values

HELM_SETTINGS = get_helm_pool_affinity_values(k8s_kluster.POOL_LABEL_NAME, k8s_kluster.SCYLLA_POOL_NAME)
HELM_SETTINGS |= {
    'dashboard.create': True,
    'dnsServer.create': True,
}
```

sdcm/chaos_mesh.py
vponomaryov: @soyacz adding affinity rules is not `adding complexity`, it is adding `required configuration`.
We `must` do it exactly because we should avoid what you are `ok` with -> creation of lots of redundant pods.
Always think about scale. There always limitations. We should avoid any redundant pod because it is not only resources, but also garbage in logs we gather later.

And, setting of affinity rules is pretty easy and we already have lots of such cases in the K8S code.

Here is what you need:
```
from sdcm.utils.k8s import get_helm_pool_affinity_values

HELM_SETTINGS = get_helm_pool_affinity_values(k8s_kluster.POOL_LABEL_NAME, k8s_kluster.SCYLLA_POOL_NAME)
HELM_SETTINGS |= {
    'dashboard.create': True,
    'dnsServer.create': True,
}
```

sdcm/utils/chaos_mesh.py
soyacz: I tried setting affinity like this - didn't work, anyway chaos-mesh daemon pods got installed on all the nodes.

I propose to park it in followup task if you agree.

soyacz: ok, nevermind, I made it working - just needed to nest it under `chaosDaemon`.
Thanks @vponomaryov

soyacz: There are also other resources like `controllerManager`, `dnsServer` (if we decide to use dns faults), `dashboard` (but I think we won't be using it). What pool use for them? auxiliary-pool?

soyacz:
@vponomaryov @fruch
Actually we don't need to keep any state related to chaos-mesh I think, so `ChaosMesh` class is not needed.
I added alternative solution: see sdcm/utils/k8s/chaos_mesh_alt.py 

what do you think?

sdcm/cluster_k8s/mini_k8s.py
```diff
@@ -234,6 +234,7 @@ def local_kubectl_version(self):  # pylint: disable=no-self-use
         #   Client Version: v1.18.5
         return LOCALRUNNER.run("kubectl version --client --short").stdout.rsplit(None, 1)[-1][1:]
 
+    @retrying(n=3, sleep_time=10, message="Pulling docker image")
```
fruch, line 237: how this is related to the PR ? do you call it here ?

sdcm/utils/k8s/chaos_mesh_alt.py
```diff
@@ -0,0 +1,59 @@
...
+    LOGGER.info("chaos-mesh installed successfully.")
+
+
+@dataclass
+class Experiment:
+    """This is where definition of experiments will be crated,
+    different experiments will be inheriting from this class for type checks and common params"""
+
+
+def run_experiment(k8s_cluster: KubernetesCluster, experiment: Experiment) -> None:
```
fruch, line 50: doesn't an experiment have some id or name ?

so we can get it back, get it's status, wait for it to end ?

fruch:
> @vponomaryov @fruch Actually we don't need to keep any state related to chaos-mesh I think, so `ChaosMesh` class is not needed. I added alternative solution: see sdcm/utils/k8s/chaos_mesh_alt.py
> 
> what do you think?

I'm o.k. that it would be totally classless

but having a class that keeps the reference to the k8s_cluster, seem a  bit more friendly to me (and repeating in all calls the actual cluster, it would be exactly the same one):
```
chaos_mesh = ChaosMesh(self.k8s_cluster)
experiment = chaos_mesh.run_experiment(...)

chaos_mesh.get_experiment_status(experiment)
chaos_mesh.wait_for_experiment(experiment)
```

sdcm/utils/chaos_mesh.py
vponomaryov: Yes, `auxiliary-pool` / `default-pool`.

sdcm/cluster_k8s/mini_k8s.py
soyacz: It's not related, I had issues with it and wanted to fix it. I'll move it to different commit/PR

sdcm/utils/k8s/chaos_mesh_alt.py
soyacz: yes, it has name, but I don't know complete answers yet.
We can return some id/name if it will be required and make waiting functions if necessary.

soyacz:
Ok, if you both wish to have it as part of `k8s_cluster` class - let it be this way. I'm not forcing it.

Please re-review @fruch @vponomaryov

test-cases/scylla-operator/functional.yaml
```diff
@@ -30,3 +30,6 @@ k8s_functional_test_dataset: MULTI_COLUMNS_DATA
 
 # enable by default
 k8s_enable_tls: true
+
+# for PR testing purposes only, don't merge
```
vponomaryov, line 34: Left over from the draft state?

vponomaryov (changes requested):

vponomaryov (commented):
From code point of view it looks ok.
But how to try it out? Does it work?

soyacz:
For now it just installs - in job's log we can see it get installed. I'm preparing code to inject some first fault. But maybe do it in separate PR?

vponomaryov:
> For now it just installs - in job's log we can see it get installed. I'm preparing code to inject some first fault. But maybe do it in separate PR?

Without possibility to try it out we don't really know whether it works or not.
And if it doesn't work why then to install it at all?

Separate commit in this PR with the usage of the feature will also be ok, IMHO.

fruch:
> > For now it just installs - in job's log we can see it get installed. I'm preparing code to inject some first fault. But maybe do it in separate PR?
> 
> Without possibility to try it out we don't really know whether it works or not.
> And if it doesn't work why then to install it at all?
> 
> Separate commit in this PR with the usage of the feature will also be ok, IMHO.

Let's do a simple functional test to demonstrate at least on activation of it.

And I then in following PRs replaceing some nemesis, like MemoryStressNemesis

fruch (dismissed):
LGTM, but let add one functional test, to see at least on experiment in action

soyacz:
> LGTM, but let add one functional test, to see at least on experiment in action

In progress - I added new commit with example experiment. Running it here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/operator-functional-eks/5/

soyacz:
Looks like it works, but still something wrong with the test - looks like cluster didn't recover properly after pod failure failing next test.

fruch:
> Looks like it works, but still something wrong with the test - looks like cluster didn't recover properly after pod failure failing next test.

it's a problem not related to you code changes, this test is broken on EKS, (it also failed in the weekly run).
and we'll need to fix it (I think the address of the node we are assuming there isn't directly available, and we need to pick the service address. (or the remove this test, since there are already other tests covering connectivity, and this test is a bit too specific and brittle to work.

soyacz:
> it's a problem not related to you code changes, this test is broken on EKS, (it also failed in the weekly run). and we'll need to fix it (I think the address of the node we are assuming there isn't directly available, and we need to pick the service address. (or the remove this test, since there are already other tests covering connectivity, and this test is a bit too specific and brittle to work.

Thanks @fruch for the info. Anyway I did improve logic for final part of the new test (now waits not just for the pod, but rather for cluster to be healthy - which is more appropriate way).

@vponomaryov please review how it works.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -860,3 +861,11 @@ def test_cloud_bundle_connectivity_cqlsh(db_cluster: ScyllaPodCluster):
...
     assert not res.stderr
     assert '(1 rows)' in res.stdout
+
+
+def test_can_recover_from_fatal_pod_termination(db_cluster):
+    target_node = db_cluster.nodes[-1]
+    experiment = PodFailureExperiment(pod=target_node, duration="30s")
+    experiment.start()
+    experiment.wait_until_finished(60)
+    db_cluster.wait_for_nodes_up_and_normal(nodes=[target_node], verification_node=target_node)
```
vponomaryov, line 871: This is test and we should check more things here. For now, it doesn't make sure that anything really happened.
It must fail if everything was just idle.
For example, using scylla image with non-semver compatible docker tag any update to the Scylla spec won't make any effect and your test must catch it, but it won't.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -860,3 +861,11 @@ def test_cloud_bundle_connectivity_cqlsh(db_cluster: ScyllaPodCluster):
 
     assert not res.stderr
     assert '(1 rows)' in res.stdout
+
+
+def test_can_recover_from_fatal_pod_termination(db_cluster):
+    target_node = db_cluster.nodes[-1]
+    experiment = PodFailureExperiment(pod=target_node, duration="30s")
```
vponomaryov, line 868: Why experiment? It sounds to me more from the your task point of view, but not test point of view.

In our case the `fault injection` is more suitable term here, IMHO.

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -44,3 +67,106 @@ def initialize(self) -> None:
...
+    FINISHED = 3
+    ERROR = 4
+    UNKNOWN = 5
+
+
+class PodChaosExperiment:
+    """Base class for all PodChaos experiments."""
+    API_VERSION = "chaos-mesh.org/v1alpha1"
+
+    def __init__(self, pod: "BasePodContainer", name: str):
```
vponomaryov, line 85: Requiring a `pod` here looks wrong. You use `db_cluster` object only out of it. So, why not using `db_cluster` param directly?

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -44,3 +67,106 @@ def initialize(self) -> None:
...
+        LOGGER.info("pod-failure experiment '%s' has started", self._name)
+
+    def get_status(self) -> ExperimentStatus:
+        """Gets status of podchaos experiment."""
+        result = self._k8s_cluster.kubectl(
+            f"get podchaos {self._name}  -n {self._namespace} -o jsonpath='{{.status.conditions}}'", verbose=False)
+        condition = {cond["type"]: ast.literal_eval(cond["status"]) for cond in json.loads(result.stdout)}
+        if not condition["Selected"] and condition["Paused"]:
+            return ExperimentStatus.ERROR
+        if not condition["Selected"] and not condition["Paused"] and not condition["AllRecovered"] and not condition["AllInjected"]:
```
vponomaryov, line 114: It may be shorter:
```
if not (condition["Selected"] or condition["Paused"] or condition["AllRecovered"] or condition["AllInjected"]):
```

The same is valid also in other conditions.

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -44,3 +67,106 @@ def initialize(self) -> None:
...
+            if status == ExperimentStatus.FINISHED:
+                LOGGER.debug("'%s' experiment ended.", self._name)
+                return
+            elif status == ExperimentStatus.ERROR:
+                raise PodChaosException(msg="Experiment status error",
+                                        k8s_cluster=self._k8s_cluster,
+                                        podchaos_name=self._name,
+                                        namespace=self._namespace)
+            time.sleep(2)
+        raise PodChaosException(msg="Timeout when waiting for ChaosMesh experiment to complete.",
```
vponomaryov, line 142: Looks like it must be more specific type of an exception -> `PodChaosTimeout`

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -44,3 +67,106 @@ def initialize(self) -> None:
...
+class PodFailureExperiment(PodChaosExperiment):
+    """
+    This experiment works by replacing container image with dummy image.
+    Then it waits for specified duration and rolls back image config.
+    """
+
+    def __init__(self, pod: "BasePodContainer", duration: str):
+        """Injects fault into a specified Pod to make the Pod unavailable for a period of time.
+
+        'duration' is str type with compatible with chaos-mesh notation. E.g. 10s, 5m
```
vponomaryov, line 157: It is not chaos-mesh notation. It is also valid for the whole K8S.

vponomaryov (changes requested):

vponomaryov:
Also, commit messages are not reports to a manager, it must not be past tense. 
The commit message and it's change are the instructions to code base to behave differently.
So, it must be in imperative mood.

fruch:
> Also, commit messages are not reports to a manager, it must not be in past tense. 
> The commit message and it's change are the instructions to code base to behave differently.
> So, it must be in imperative mood.

As long as the information I would expect is in the commit, it can be present progressive, past simple or greek from my POV.

It should be clear, and explain the reasons.

functional_tests/scylla_operator/test_functional.py
fruch: Experiment is the chaos-mesh terminology, I think it's o.k. to stick with that.

sdcm/utils/k8s/chaos_mesh.py
fruch: I seem like it's used for targeting, I would expect the targeting code to be in one place.

When we target a specific pod, or when we target the whole cluster.
I think the most common case, is targeting a single pod, and we should built is so we don't need to repeat those again and again

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -0,0 +1,172 @@
...
+        if condition["Selected"] and not condition["Paused"] and not condition["AllRecovered"] and condition["AllInjected"]:
+            return ExperimentStatus.RUNNING
+        if condition["AllRecovered"] and condition["Paused"]:
+            return ExperimentStatus.PAUSED
+        if condition["AllRecovered"] and not condition["Paused"]:
+            return ExperimentStatus.FINISHED
+        LOGGER.warning("Unknown experiment status: %s", condition)
+        return ExperimentStatus.UNKNOWN
+
+    def wait_until_finished(self, timeout: int):
```
fruch, line 125: in `PodFailureExperiment` we specify the duration, it would be nice if the code could use the information from the experiment to figure out what's the timeout should be. 

and not force who write it to keep track, and do the calculation of how much extra time it's gonna give this function.

functional_tests/scylla_operator/test_functional.py
fruch: why this test should be about the non-semver "feature" ? 

I agree it doesn't verify the experiment was working, i.e. before it's waiting, it can check the state of the pod is failing/fetching image. 

and then wait for the experiment to be over, and test the state of the cluster/pod is back to normal.

I think it's enough for this test, keep in mind it's just a playground, before we start calling this code from Nemesis.

sdcm/utils/k8s/chaos_mesh.py
soyacz: I rather had all conditionals to be in the same form, where for each condition we use `and` or not use condition at all when has no impact on given status. For me it is more readable. I don't care about extra few characters.

sdcm/utils/k8s/chaos_mesh.py
soyacz: true, fixed

functional_tests/scylla_operator/test_functional.py
vponomaryov: We should not use each new terminology ` as-is` from each new lib we import.
There is well-known `adapter design pattern` for such things.

sdcm/utils/k8s/chaos_mesh.py
vponomaryov: The proposed variant is also simple.

functional_tests/scylla_operator/test_functional.py
soyacz: I added verification that pod loses `Ready` status

functional_tests/scylla_operator/test_functional.py
soyacz: I think `PodFailureExperiment` is a good name - first part shows that we are going to fail the pod, and the second `experiment` refers to chaos-mesh nomenclature (which is not bad). So it's clear we're going to fail the pod with chaos-mesh.

In case we will find a better tool for this purpose, we will anyway create different classes, that may need different approach, parameters, etc. and adapter might not fit here and create additional unnecessary work.

So my proposal is to introduce adapter when it is needed and continue with this approach.

sdcm/utils/k8s/chaos_mesh.py
soyacz: but is different from the others - so will require brain creating new pattern to compile and use taking cycles, memory and context switch.

And finally I don't see it as important - this is just implementation detail.

soyacz:
@fruch @vponomaryov please re-review

sdcm/utils/common.py
```diff
@@ -2803,3 +2803,11 @@ def __enter__(self):
     def __exit__(self, exit_type, value, _traceback):
         # remove the temporary folder as `sudo` to cover the case when the folder owner was changed during test
         self.node.remoter.sudo(f'rm -rf {self.folder_name}')
+
```
soyacz, line 2806: this function is moved from `stress_thread.py` as I needed it and fits better here

sdcm/utils/k8s/chaos_mesh.py
fruch: not critical change from my POV

fruch (approved):
LGTM

soyacz:
@fruch it passed functional test by provision-test. Merge?

vponomaryov (commented):
Forgot to post review comments long time ago >.<

sdcm/utils/k8s/__init__.py
```diff
@@ -566,7 +568,15 @@ def _helm_install_or_upgrade(self,
...
         if debug:
             command.extend(("--debug",))
-
+        if atomic:
+            # if set, the installation process deletes the installation on failure
+            # or upgrade process rolls back changes made in case of failed upgrade
+            # waits for operation completion - so should specify timeout.
+            command.extend(("--atomic",))
+            if timeout is None:
+                LOGGER.warning("No timeout provided for atomic operation, default of '5m' is used.")
```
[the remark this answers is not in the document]
soyacz: I think we can drop it then

PR #5567, opened 2022-12-15
fix(test): pin scylla-bench to fix multidc counter test

soyacz:
We face the issue with disconnecting from cluster after simple table modifications in scylla-bench (counters multidc test): https://github.com/scylladb/scylla-bench/issues/114

It was proven it behaves correctly when s-b is in version 0.1.3. This commit pins s-b for couters multidc test.

### DIFF
--- a/test-cases/longevity/longevity-counters-multidc.yaml
+++ b/test-cases/longevity/longevity-counters-multidc.yaml
@@ -22,3 +22,7 @@ server_encrypt: true
 internode_encryption: 'dc'
 
 use_legacy_cluster_init: false
+
+# pin s-b version to mitigate https://github.com/scylladb/scylla-bench/issues/114
+stress_image:
+  scylla-bench: 'scylladb/hydra-loaders:scylla-bench-v0.1.3'


vponomaryov:
@soyacz in the old version you compare with we do use retries. In new one we have concrete limit. So, consider please usage of more retries or bigger intervals for this job.

fruch:
@soyacz do have a job run with this pinned ? (running with master ?)

soyacz:
this is the job with pinned: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/reproducer-counters-multidc-raft/2/

soyacz:
> @soyacz in the old version you compare with we do use retries. In new one we have concrete limit. So, consider please usage of more retries or bigger intervals for this job.

I can try with retries instead of pinning

soyacz:
I tried and with setting ` -timeout 15s -retry-interval=80ms,5s -retry-number=20` failed again after table modification.

soyacz:
ping @fruch

roydahan:
We need to fix that in s-b, but this PR should be merged meanwhile.

roydahan (approved):

PR #5569, opened 2022-12-15
feature(k8s-cloud-bundle): enable for multi tenant test

fruch:

### DIFF
--- a/test-cases/scylla-operator/longevity-scylla-operator-3h-multitenant.yaml
+++ b/test-cases/scylla-operator/longevity-scylla-operator-3h-multitenant.yaml
@@ -30,4 +30,6 @@ nemesis_seed: ['385', '543']
 space_node_threshold: [64424, 64423]
 
 user_prefix: 'longevity-scylla-operator-3h-multitenant'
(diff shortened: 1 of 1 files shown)


test-cases/scylla-operator/longevity-scylla-operator-3h-multitenant.yaml
```diff
@@ -30,4 +30,6 @@ nemesis_seed: ['385', '543']
 space_node_threshold: [64424, 64423]
 
 user_prefix: 'longevity-scylla-operator-3h-multitenant'
-cluster_health_check: false
+cluster_health_check: true
+
+k8s_enable_tls: true
```
vponomaryov, line 35: Do we want to have only this variant?
Maybe we want to create second job and then just choose what we want to run?

fruch: Yes we want only this variant, we are targeting for what serverless is, and it would be only with it.

in case we need to compare we can rerun without it

vponomaryov (approved):

PR #5579, opened 2022-12-20
improvement(k8s): split K8S node diruptive nemesis to separate methods

vponomaryov:
Changes:
- Split `disrupt_terminate_and_replace_node_kubernetes` into following:
  - `disrupt_drain_kubernetes_node_then_replace_scylla_node`
  - `disrupt_terminate_kubernetes_host_then_replace_scylla_node`
- Split `disrupt_terminate_decommission_add_node_kubernetes` into following:
  - `disrupt_drain_kubernetes_node_then_decommission_and_add_scylla_node`
  - `disrupt_terminate_kubernetes_host_then_decommission_and_add_scylla_node`
- Rename `disrupt_replace_node_kubernetes` method -> `disrupt_replace_scylla_node_on_kubernetes`.
- Rename `OperatorNodeTerminateAndReplace` class -> `DisruptKubernetesNodeThenReplaceScyllaNode`.
- Rename `OperatorNodeTerminateDecommissionAdd` class -> `DisruptKubernetesNodeThenDecommissionAndAddScyllaNode`.

Reason:
Above is done because we were using 3 different disruption methods as one. It is not correct, we should treat them as separate ones.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -56,7 +56,7 @@ from sdcm.cluster import (
     NodeSetupTimeout,
     NodeStayInClusterAfterDecommission,
 )
-from sdcm.cluster_k8s import PodCluster, ScyllaPodCluster
+from sdcm.cluster_k8s import PodCluster
 from sdcm.cluster_k8s.mini_k8s import LocalKindCluster
 from sdcm.db_stats import PrometheusDBStats
 from sdcm.log import SDCMAdapter
@@ -112,8 +112,10 @@ LOGGER = logging.getLogger(__name__)
 NEMESIS_LOCK = Lock()
 NEMESIS_RUN_INFO = {}
 EXCLUSIVE_NEMESIS_NAMES = (
-    "disrupt_terminate_and_replace_node_kubernetes",
-    "disrupt_terminate_decommission_add_node_kubernetes",
+    "disrupt_drain_kubernetes_node_then_replace_scylla_node",
+    "disrupt_terminate_kubernetes_host_then_replace_scylla_node",
+    "disrupt_drain_kubernetes_node_then_decommission_and_add_scylla_node",
+    "disrupt_terminate_kubernetes_host_then_decommission_and_add_scylla_node",
 )
 
 
@@ -1175,12 +1177,6 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         InfoEvent(message="FinishEvent - New Node is up and normal").publish()
         return new_node
 
-    def _get_kubernetes_node_break_methods(self):
-        if isinstance(self.cluster, ScyllaPodCluster) and self.cluster.node_terminate_methods:
-            return self.cluster.node_terminate_methods
-        raise UnsupportedNemesis(f"Backend {self.cluster.params.get('cluster_backend')} "
-                                 f"does not support node termination")
-
     @decorate_with_context(ignore_ycsb_connection_refused)
     def _terminate_cluster_node(self, node):
         self.cluster.terminate_node(node)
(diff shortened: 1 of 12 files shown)


sdcm/nemesis.py
```diff
@@ -1337,56 +1333,61 @@ def _check_pod_stats(node):
...
-            self.set_target_node()
-            InfoEvent(f'OperatorNodeTerminateAndReplace ({node_terminate_method_name}) {self.target_node}').publish()
-            if not self._is_it_on_kubernetes():
-                raise UnsupportedNemesis('OperatorNodeTerminateAndReplace is supported only on kubernetes')
+    def disrupt_drain_kubernetes_node_then_replace_scylla_node(self):
+        self._disrupt_kubernetes_then_replace_scylla_node('drain_k8s_node')
 
-            self._disrupt_terminate_and_replace_node_kubernetes(self.target_node, node_terminate_method_name)
+    def disrupt_terminate_kubernetes_host_then_replace_scylla_node(self):
+        # NOTE: affected by the https://github.com/scylladb/scylla-operator/issues/643
```
fruch, line 1340: two issues referenced here, which one is the reason we are skipping ?

sdcm/nemesis.py
```diff
@@ -1337,56 +1333,61 @@ def _check_pod_stats(node):
...
         self.log.info('Mark %s (uid=%s) to be replaced', node, old_uid)
         node.wait_for_svc()
         node.mark_to_be_replaced()
         self._kubernetes_wait_till_node_up_after_been_recreated(node, old_uid=old_uid)
 
+    def disrupt_drain_kubernetes_node_then_decommission_and_add_scylla_node(self):
+        self._disrupt_kubernetes_then_decommission_and_add_scylla_node('drain_k8s_node')
+
+    def disrupt_terminate_kubernetes_host_then_decommission_and_add_scylla_node(self):
+        # NOTE: affected by the https://github.com/scylladb/scylla-operator/issues/643
```
fruch, line 1364: Is this note still relevant? That issue still reproduce?

test-cases/scylla-operator/longevity-scylla-operator-3h-terminate-replace.yaml
```diff
@@ -7,7 +7,11 @@ k8s_n_scylla_pods_per_cluster: 3
 n_loaders: 2
 n_monitor_nodes: 1
 
-nemesis_class_name: 'OperatorNodeTerminateAndReplace'
+# TODO: use 'TerminateKubernetesHostThenReplaceScyllaNode' when
```
fruch, line 10: If we have 3 different nemesis should we have 3 different cases ? Or one nemesis set that would run them all ?

fruch (commented):
In general looks o.k.

But I don't want to skip because a new issue opened, I'd rather wait for some feedback on it.

sdcm/nemesis.py
vponomaryov: The one used in the exception. Second appears also in case of decommission and it doesn't harm there.
So, the duplicated PVs bug is kind of cosmetic for now, as I can see.

sdcm/nemesis.py
vponomaryov: Whole change is actual. The duplicated PVs bug is still existing, but it is harmless as I can see. THe decommission nemesis has it too, but passes ok.

test-cases/scylla-operator/longevity-scylla-operator-3h-terminate-replace.yaml
vponomaryov: The very first patch-set to this PR did exactly as you propose in the last statement. 
It was making unit tests fail.

Why?
Because creation of a nemesis with 3 methods causes `SisyphusMonkey` one ignore it's methods for itself when a nemesis filter is provided.

The cause for it is very strange and obfuscated logic for the nemesis methods handling.

Just imagine: SisyphusMonkey class, when is provided with nemesis filter, parses `string` variants of other nemesis classes definitions and looks for the "self.disrupt_foo(" stuff. Find `first` line and returns. It will pass if I have `comment` with such a string in class definition. Then, it returns only first occurrence. All other are ignored.

sdcm/nemesis.py
fruch: If it's harmless bug, let's drop this comment it doesn't help us.

test-cases/scylla-operator/longevity-scylla-operator-3h-terminate-replace.yaml
fruch: There a list of classes that should be ignored in that scan.

fruch: https://github.com/scylladb/scylla-cluster-tests/blob/1af3cc0f45c0d508b778376c260872fb39f89038/sdcm/nemesis.py#L4876

vponomaryov: > There a list of classes that should be ignored in that scan.

If it is ignored that these methods won't be picked up by the `SisyphusMonkey` complex nemesis. But I would expect it to be picked up.
The current logic, without changes, doesn't solve the current need.

vponomaryov: Ok, there is a way. Create A simple nemesis class for each of the methods and 2 complex ones. In that case we will be able to get new methods in both cases - only them and as part of the `SisyphusMonkey` one.

But the problem with it is that we add redundant code.

vponomaryov:
> In general looks o.k.
> 
> But I don't want to skip because a new issue opened, I'd rather wait for some feedback on it.

That one skipped method makes cluster lose one node forever. It will break any test run with the `SisyphusMonkey` nemesis.
So, it `must` be skipped.

vponomaryov:
@fruch 

Skipping the disruption method that is skipped here is the current approach in the `master` branch.
This PR is the `dependency` for another change which implements the following task: https://github.com/scylladb/qa-tasks/issues/935

I propose to merge this one if we want to have such separation in general.

fruch:
@vponomaryov please rebase this one

since it introduces new nemesis, now there are tests and nemesis list to update.

fruch:
> @fruch
> 
> Skipping the disruption method that is skipped here is the current approach in the `master` branch. This PR is the `dependency` for another change which implements the following task: [scylladb/qa-tasks#935](https://github.com/scylladb/qa-tasks/issues/935)
> 
> I propose to merge this one if we want to have such disruption methods separation in general.

We do want this, but still I don't like the fact we are skipping some parts, before we have a clue what is the cause and when/if it's going to be fixed

vponomaryov:
> > @fruch
> > Skipping the disruption method that is skipped here is the current approach in the `master` branch. This PR is the `dependency` for another change which implements the following task: [scylladb/qa-tasks#935](https://github.com/scylladb/qa-tasks/issues/935)
> > I propose to merge this one if we want to have such disruption methods separation in general.
> 
> We do want this, but still I don't like the fact we are skipping some parts, before we have a clue what is the cause and when/if it's going to be fixed

Honestly, I don't understand what is that `before` you are referring to here.
This PR doesn't change the number of nemesis we run and/or skip.
It only splits complex nemesis into several simpler ones.

fruch:
> > > @fruch
> > > Skipping the disruption method that is skipped here is the current approach in the `master` branch. This PR is the `dependency` for another change which implements the following task: [scylladb/qa-tasks#935](https://github.com/scylladb/qa-tasks/issues/935)
> > > I propose to merge this one if we want to have such disruption methods separation in general.
> > 
> > We do want this, but still I don't like the fact we are skipping some parts, before we have a clue what is the cause and when/if it's going to be fixed
> 
> Honestly, I don't understand what is that `before` you are referring to here.
> This PR doesn't change the number of nemesis we run and/or skip.
> It only splits complex nemesis into several simpler ones.

Please rebase.

The before, is that I don't like the fact we are skipping important case while we have zero clue of the cause.

Last time that it was skipped it was forgotten about for almost a whole year, we need to push it forward, our work doesn't end in just opening an issue.

vponomaryov:
> > > > @fruch
> > > > Skipping the disruption method that is skipped here is the current approach in the `master` branch. This PR is the `dependency` for another change which implements the following task: [scylladb/qa-tasks#935](https://github.com/scylladb/qa-tasks/issues/935)
> > > > I propose to merge this one if we want to have such disruption methods separation in general.
> > > 
> > > 
> > > We do want this, but still I don't like the fact we are skipping some parts, before we have a clue what is the cause and when/if it's going to be fixed
> > 
> > 
> > Honestly, I don't understand what is that `before` you are referring to here.
> > This PR doesn't change the number of nemesis we run and/or skip.
> > It only splits complex nemesis into several simpler ones.
> 
> Please rebase.
> 
> The before, is that I don't like the fact we are skipping important case while we have zero clue of the cause.
> 
> Last time that it was skipped it was forgotten about for almost a whole year, we need to push it forward, our work doesn't end in just opening an issue.

Forgotten by whom? The case is still broken.

At first, what do you propose to do?
At second, the discussion about the `skip/unskip` is unrelated to this particular change.

soyacz:
Maybe when doing this split, we could unify some naming, e.g. put `k8s` or `kubernetes` after word `disrupt_`? e.g: `disrupt_kubernetes_replace_scylla_node`

fruch:
> Maybe when doing this split, we could unify some naming, e.g. put `k8s` or `kubernetes` after word `disrupt_`? e.g: `disrupt_kubernetes_replace_scylla_node`

those are gonna by label as such in following work, why do we need to prefix the names ?

fruch:
> > > > > @fruch
> > > > > Skipping the disruption method that is skipped here is the current approach in the `master` branch. This PR is the `dependency` for another change which implements the following task: [scylladb/qa-tasks#935](https://github.com/scylladb/qa-tasks/issues/935)
> > > > > I propose to merge this one if we want to have such disruption methods separation in general.
> > > > 
> > > > 
> > > > We do want this, but still I don't like the fact we are skipping some parts, before we have a clue what is the cause and when/if it's going to be fixed
> > > 
> > > 
> > > Honestly, I don't understand what is that `before` you are referring to here.
> > > This PR doesn't change the number of nemesis we run and/or skip.
> > > It only splits complex nemesis into several simpler ones.
> > 
> > 
> > Please rebase.
> > The before, is that I don't like the fact we are skipping important case while we have zero clue of the cause.
> > Last time that it was skipped it was forgotten about for almost a whole year, we need to push it forward, our work doesn't end in just opening an issue.
> 
> Forgotten by whom? The case is still broken.

Forgotten by us, in
https://github.com/vponomaryov/scylla-cluster-tests/commit/678eeb6c973e9c0d18f2ff42b7a48aaaea2f93c6
we replaced the issues, and kept on 

now we facing a new issue, and we add it as skipped in a different manner.

> 
> At first, what do you propose to do? At second, the discussion about the `skip/unskip` is unrelated to this particular change.

it's related  to the PR, since we are introducing a new skip here.

Let's see what we can do to push forward resolving that issue issue, don't want to wait a year an a half, assuming it would fix itself.

sdcm/nemesis.py
```diff
@@ -1337,56 +1333,59 @@ def _check_pod_stats(node):
...
-            self.set_target_node()
-            InfoEvent(f'OperatorNodeTerminateAndReplace ({node_terminate_method_name}) {self.target_node}').publish()
-            if not self._is_it_on_kubernetes():
-                raise UnsupportedNemesis('OperatorNodeTerminateAndReplace is supported only on kubernetes')
+    def disrupt_drain_kubernetes_node_then_replace_scylla_node(self):
+        self._disrupt_kubernetes_then_replace_scylla_node('drain_k8s_node')
 
-            self._disrupt_terminate_and_replace_node_kubernetes(self.target_node, node_terminate_method_name)
+    def disrupt_terminate_kubernetes_host_then_replace_scylla_node(self):
+        raise UnsupportedNemesis("https://github.com/scylladb/scylla-operator/issues/1124")
```
fruch, line 1340: now that we know it's working as we expect in serverless, let's not skip it for serverless

fruch (changes requested):
- remove the skip from disrupt_terminate_kubernetes_host_then_replace_scylla_node for serverless
- rebase this PR

soyacz:
> > Maybe when doing this split, we could unify some naming, e.g. put `k8s` or `kubernetes` after word `disrupt_`? e.g: `disrupt_kubernetes_replace_scylla_node`
> 
> those are gonna by label as such in following work, why do we need to prefix the names ?

I was just thinking, when we already change names, so maybe put some order to naming. It's just suggestion.

vponomaryov:
> > > Maybe when doing this split, we could unify some naming, e.g. put `k8s` or `kubernetes` after word `disrupt_`? e.g: `disrupt_kubernetes_replace_scylla_node`
> > 
> > 
> > those are gonna by label as such in following work, why do we need to prefix the names ?
> 
> I was just thinking, when we already change names, so maybe put some order to naming. It's just suggestion.

The current approach has logic. First part is required `disrupt_` prefix, then goes `action type` and then the `object` for action.

fruch (approved):
LGTM

vponomaryov:
@fruch it requires your attention

fruch (approved):
LGTM

PR #5582, opened 2022-12-21
fix(aws_instance_profile): split profiles per node type

fruch:
Split the configuration, so we can select specific profile for each type of node, and not use the same one for all of the nodes.

### DIFF
--- a/defaults/aws_config.yaml
+++ b/defaults/aws_config.yaml
@@ -37,7 +37,7 @@ ami_db_scylla_user: 'scyllaadm'
 ami_loader_user: 'centos'
 # centos7 is used for monitor
 ami_monitor_user: 'centos'
-aws_instance_profile_name: 'qa-scylla-manager-backup-instance-profile'
+aws_instance_profile_name_db: 'qa-scylla-manager-backup-instance-profile'
 
 ami_id_db_scylla: ''
 ami_id_db_oracle: ''
--- a/sdcm/cluster_aws.py
+++ b/sdcm/cluster_aws.py
@@ -121,6 +121,14 @@ class AWSCluster(cluster.BaseCluster):  # pylint: disable=too-many-instance-attr
     def calculate_spot_duration_for_test(self):
         return floor(self.test_config.TEST_DURATION / 60) * 60 + 60
 
+    @property
+    def instance_profile_name(self) -> str | None:
+        if self.node_type == "scylla-db":
+            return self.params.get('aws_instance_profile_name_db')
+        if self.node_type == "loader":
+            return self.params.get('aws_instance_profile_name_loader')
+        return None
+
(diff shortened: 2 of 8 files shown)


defaults/aws_config.yaml
```diff
@@ -37,7 +37,7 @@ ami_db_scylla_user: 'scyllaadm'
 ami_loader_user: 'centos'
 # centos7 is used for monitor
 ami_monitor_user: 'centos'
-aws_instance_profile_name: 'qa-scylla-manager-backup-instance-profile'
+aws_instance_profile_name_db: 'qa-scylla-manager-backup-instance-profile'
```
soyacz, line 40: shouldn't add also value for `aws_instance_profile_name_loader` profile?

sdcm/sct_config.py
```diff
@@ -672,8 +672,11 @@ class SCTConfiguration(dict):
         dict(name="extra_network_interface", env="SCT_EXTRA_NETWORK_INTERFACE", type=boolean,
              help="if true, create extra network interface on each node"),
 
-        dict(name="aws_instance_profile_name", env="SCT_AWS_INSTANCE_PROFILE_NAME", type=str,
-             help="This is the name of the instance profile to set on all instances"),
+        dict(name="aws_instance_profile_name_db", env="SCT_AWS_INSTANCE_PROFILE_NAME_DB", type=str,
+             help="This is the name of the instance profile to set on all db instances"),
+
+        dict(name="aws_instance_profile_name_loader", env="SCT_AWS_INSTANCE_PROFILE_NAME_LOADER", type=str,
+             help="This is the name of the instance profile to set on all db instances"),
```
soyacz, line 679: should refer to loader

sdcm/cluster_aws.py
```diff
@@ -121,6 +121,14 @@ def __str__(self):
     def calculate_spot_duration_for_test(self):
         return floor(self.test_config.TEST_DURATION / 60) * 60 + 60
 
+    @property
+    def instance_profile_name(self) -> str | None:
+        if self.node_type == "scylla-db":
+            return self.params.get('aws_instance_profile_name_db')
+        if self.node_type == "loader":
+            return self.params.get('aws_instance_profile_name_loader')
+        return None
```
soyacz, line 130: what about monitor?
Shouldn't be there some default for other cases?

defaults/aws_config.yaml
fruch: no, since by default we don't add to it.

sdcm/cluster_aws.py
fruch: we don't have a use case yet that we need it for monitor.

when we will, we'll add this configuration and code to handle it.

sdcm/sct_config.py
fruch: fixed

soyacz (approved):

roydahan (approved):

Issue #5586, opened 2022-12-22
Latest hydra docker image v1.28 has too old kubectl package


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: any

### Logs
* test_id: b757321e-caca-4f78-88a3-3aa439fcdd01
* job log: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-12h-multitenant-eks/25

### Description

PR https://github.com/scylladb/scylla-cluster-tests/pull/5348 introduced new hydra version `v1.28` and built image, for some reason, includes too old `kubectl` package. The one, that was before merge of this PR: https://github.com/scylladb/scylla-cluster-tests/pull/5529 (which is in the history of the first one)

Or in short: all `K8S` backends are `broken` now in CI using `master` branch of the SCT which will fail with following error:
```
2022-12-22 13:15:19,110 f:base.py         l:146  c:LocalCmdRunner       p:ERROR > 
    Error executing command: "kubectl --cache-dir=/home/ubuntu/sct-results/20221222-124731-728559/.kube/http-cache --namespace=scylla-2 wait --timeout=18m --all --for=jsonpath='{.status.phase}'=Running pod"; Exit status: 1
2022-12-22 13:15:19,118 f:base.py         l:150  c:LocalCmdRunner       p:DEBUG > 
    STDERR: error: unrecognized condition: "jsonpath={.status.phase}=Running"
```

### Steps to Reproduce

1. Run hydra docker image v1.28
2. Type in the container `kubectl version` command
3. You will see `Client Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.2", GitCommit:"8b5a19147530eaac9476b0ab82980b4088bbc1b2", GitTreeState:"clean", BuildDate:"2021-09-15T21:38:50Z", GoVersion:"go1.16.8", Compiler:"gc", Platform:"linux/amd64"}` answer. But must be `v1.23.14`

**Expected behavior:** `kubectl` package must be of the `v1.23.14` version

**Actual behavior:** `kubectl` package is `v1.22.2`

PR #5587, opened 2022-12-22
fix(hydra): use properly built image with actual requirements

vponomaryov:
Latest, for now, hydra image `v1.28` was built without regards to the changes made as part of the `v1.26` version. 
So, use fixed version of it called `v1.28-fixed-kubectl-version`.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/5586

### DIFF
--- a/docker/env/version
+++ b/docker/env/version
@@ -1,1 +1,1 @@
-1.28
+1.28-fixed-kubectl-version-v2
--- a/requirements.in
+++ b/requirements.in
@@ -36,6 +36,7 @@ parameterized==0.8.1
 pylint==2.11.1  # Needed for pre-commit hooks
 autopep8==1.5.7  # Needed for pre-commit hooks
 kubernetes==18.20.0
+packaging==21.3
 ldap3==2.9.1
 google-api-python-client==2.24.0
 google-cloud-storage==2.4.0
--- a/requirements.txt
+++ b/requirements.txt
@@ -24,9 +24,9 @@ astroid==2.8.6 \
     --hash=sha256:5f6f75e45f15290e73b56f9dfde95b4bf96382284cde406ef4203e928335a495 \
     --hash=sha256:cd8326b424c971e7d87678609cf6275d22028afd37d6ac59c16d47f1245882f6
     # via pylint
-attrs==22.1.0 \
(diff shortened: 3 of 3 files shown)


vponomaryov:
Oh no. Now another failure appeared:
```
E   AttributeError: module 'packaging.version' has no attribute 'LegacyVersion'
```

Looks like it became too new.

roydahan (approved):

vponomaryov:
> Oh no. Now another failure appeared:
> 
> ```
> E   AttributeError: module 'packaging.version' has no attribute 'LegacyVersion'
> ```
> 
> Looks like it became too new.

Done by nailing down the `packaging` package. It's newest version is not compatible with our code.

Please review it and merge because all the K8S backends are broken now without this fix.

fruch (approved):

soyacz:
I don't see any kubectl related change in the commit. @vponomaryov how does kubectl version is set?
I wonder if there is a way to avoid such issues in future?

fruch:
> I don't see any kubectl related change in the commit. @vponomaryov how does kubectl version is set?
> I wonder if there is a way to avoid such issues in future?

The problem here is that the code doing it is in place, but the image wasn't rebuilt to reflect that, we doesn't have something that can verify that currently.

Since building the images is a manual process, it's hard to notice and avoid those errors

soyacz:
I don't think it is hard, could be automated:
1. create pre-commit hook that verifies if requirements file changed - if so, force require change in `hydra_version` file.
2. Set jenkins job that run on each master commit. It would build&publish new hydra image if `hydra_version` file changed.

feasible?

fruch:
> I don't think it is hard, could be automated:
> 1. create pre-commit hook that verifies if requirements file changed - if so, force require change in `hydra_version` file.
> 2. Set jenkins job that run on each master commit. It would build&publish new hydra image if `hydra_version` file changed.
> 
> feasible?

You are missing the part that if you changed the version, CI can't run on the PR since you won't have an image.

Also in this case the requirements file didn't change the Dockerfile changed.

Anyhow I would suggest opening a new issue and discuss it there.

soyacz:
issue open: https://github.com/scylladb/scylla-cluster-tests/issues/5602

Issue #5590, opened 2022-12-23
Test runs using K8S backends and C-S loaders fail with AttributeError and TypeError exceptions


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: any

### Logs
* test_id: 0f35e5d4-34ce-4b03-80ee-7c0fdf15b120
* job log: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-12h-multitenant-eks/31

### Description

After merge of the https://github.com/scylladb/scylla-cluster-tests/pull/5348 and fix for it's first breakage to K8S backends https://github.com/scylladb/scylla-cluster-tests/pull/5587 we get following errors running any K8S test run with C-S loader:
```
Exception ignored in: <function RemoteCmdRunnerBase.__del__ at 0x7fe37f14c310>                         
Traceback (most recent call last):                                                                     
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 138, in __del__            
    self.stop()                                                                                        
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 128, in stop               
    self._close_connection()                                                                           
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 131, in _close_connection
    if self.connection:                                                                                
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 74, in connection          
    connection = self._create_connection()                                                             
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_libssh_cmd_runner.py", line 44, in _create_connection
    host=self.hostname,                                                                                
AttributeError: 'RemoteLibSSH2CmdRunner' object has no attribute 'hostname'

...

2022-12-23 16:54:04.245: (ThreadFailedEvent Severity.ERROR) period_type=one-time event_id=0f136839-a456-42cd-a08a-5cb0b407cdc1: message=RemoteCmdRunnerBase.__init__() got an unexpected keyword argument 'kluster'Traceback (most recent call last):                                                                  
  File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/decorators.py", line 26, in wrapper       
[...]

PR #5348, opened 2022-10-11
Latency hdrhistogram

aleksbykov:
A bunch of commits which add the opportunity to run c-s command with/without docker with hdr histogram logging and 
collectiong histograms for various periods

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5591, opened 2022-12-23
fix(k8s): make CSHDRFileLogger context manager be compatible with K8S

vponomaryov:
Recently merged PR https://github.com/scylladb/scylla-cluster-tests/pull/5348 added 2 more context manager objects in the `CassandraStressThread` class.
But the problem is that one of these new objects uses explicitly only SSH-based remoter not allowing to use K8S-based ones. 
So, detect that we use K8S and enable proper remoter for that context manager.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/5590


soyacz (approved):

fruch (approved):
LGTM

PR #5596, opened 2022-12-25
fix(sct_config): make .get(None) work

fruch:
Since a recent change in instance profiles, we are doing calls of `params.get(None)`, assuming it would work like a regular dict and would return None, but it was failing like this:

```
File ".../instance_parameters_builder.py", line 74, in IamInstanceProfile
   if profile := self.params.get(self._INSTANCE_PROFILE_PARAM_NAME):
File ".../sdcm/sct_config.py", line 1808, in get
   if '.' in key:
TypeError: argument of type 'NoneType' is not iterable
```

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -1800,12 +1800,12 @@ class SCTConfiguration(dict):
 
         return environment_vars
 
-    def get(self, key: str):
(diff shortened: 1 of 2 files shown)


roydahan (approved):

fgelcer (approved):
LGTM

PR #5609, opened 2022-12-28
fix(test_operator_managed_tls): add WhiteList when connecting to cluster

fruch:
since the address we connect to the cluster are diffrent then the addresses used by scylla to publish it's nodes.
we need to whitelist the extrenal addresses we use other wise the driver would remove all the connections, since they are not accessable, and we'll fail with connction refused.

## Testing
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/functional-eks/2/

### DIFF
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -22,7 +22,12 @@ import path
 
 import pytest
 import yaml
-from cassandra.cluster import Cluster  # pylint: disable=no-name-in-module
+from cassandra.cluster import (  # pylint: disable=no-name-in-module
+    Cluster,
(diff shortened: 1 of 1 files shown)


soyacz (approved):
LGTM
Is this is only relevant to our test, not something user must be aware of?

fruch:
> LGTM Is this is only relevant to our test, not something user must be aware of?

serverless user shouldn't care about this

other operator users, it's depends if they are using it from within the k8s cluster our from the outside of it.

if it's from the outside, they probably would need on of those tricks (cqlsh does it by default, and there are other options like AddressTranslator)

cause it's not this trivial the serverless is what's should be closing the gaps for users of scylla-cloud

Issue #5455, opened 2022-11-23
[k8] MemoryStress nemesis fails to install stress-ng


fruch:
MemoryStress stress nemesis trying to install with snap that doesn't work on scylla docker

```
(DisruptionEvent Severity.ERROR) period_type=end event_id=75af9b75-7e3e-4a09-b1d6-1b888e764bbb duration=10s: nemesis_name=MemoryStress target_node=Node sct-cluster-2-us-east1-b-us-east1-2 [10.0.3.245 | 10.0.1.54] (seed: False) errors=Encountered a bad command exit code!

Command: 'snap install stress-ng'

Exit code: 1

Stdout:



Stderr:

error: cannot communicate with server: Post "http://localhost/v2/snaps/stress-ng": dial unix /run/snapd.socket: connect: no such file or directory
```

@vponomaryov have you seen this ?

maybe we should consider using some other means like
https://chaos-mesh.org/blog/how-to-efficiently-stress-test-pod-memory/

vponomaryov:
> @vponomaryov have you seen this ?
No, I haven't.

And how should it be reproduced?

fruch:
> > @vponomaryov have you seen this ?
> 
> No, I haven't.
> 
> And how should it be reproduced?

according to the dashboard:
https://70f106c98484448dbc4705050eb3f7e9.us-east-1.aws.found.io:9243/goto/b997de2089a8bcea7889c78b9ac79123

running with any scylla image bigger then 5.0, it would break (I'm guessing the image changed a bit, and some stuff was removed)

fruch:
we probably just need to replace it with using https://chaos-mesh.org/, it has the same capability, without messing around with the software of the image under test.

PR #5613, opened 2022-12-29
feature(chaos-mesh): use chaos-mesh for memory stress in k8s clusters

soyacz:
Memory stress nemesis does not work on k8s:
- #5455

This commit adds support for chaos-mesh's memory stress experiment.
Also made some bug fixing: additional experiment failure detection.
Code unification to ease adding new experiments.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -91,8 +91,9 @@ from sdcm.utils.decorators import timeout as timeout_decor
 from sdcm.utils.docker_utils import ContainerManager
 from sdcm.utils.k8s import (
     convert_cpu_units_to_k8s_value,
-    convert_cpu_value_from_k8s_to_units,
+    convert_cpu_value_from_k8s_to_units, convert_memory_value_from_k8s_to_units,
 )
+from sdcm.utils.k8s.chaos_mesh import MemoryStressExperiment
 from sdcm.utils.ldap import SASLAUTHD_AUTHENTICATOR
 from sdcm.utils.replication_strategy_utils import temporary_replication_strategy_setter, \
     NetworkTopologyReplicationStrategy, ReplicationStrategy, SimpleReplicationStrategy
@@ -409,6 +410,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     def _is_it_on_kubernetes(self) -> bool:
         return isinstance(getattr(self.tester, "db_cluster", None), PodCluster)
+
+    def _is_chaos_mesh_initialized(self) -> bool:
+        return self.tester.k8s_cluster.chaos_mesh.initialized
 
     # pylint: disable=too-many-arguments,unused-argument
     def get_list_of_methods_by_flags(
@@ -3532,12 +3536,38 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self.disrupt_grow_shrink_cluster()
         InfoEvent(message="Finished grow_shrink disruption").publish()
 
+    def _k8s_disrupt_memory_stress(self):
+        """Uses chaos-mesh experiment based on https://github.com/chaos-mesh/memStress"""
+        if not self._is_chaos_mesh_initialized:
+            raise UnsupportedNemesis(
+                "Chaos Mesh is not installed. Set 'k8s_use_chaos_mesh' config option to 'true'")
+        memory_limit = self.tester.k8s_cluster.calculated_memory_limit
+        # If a container's memory usage increases too quickly the OOM killer is invoked
+        # so reduce ramp to ~2GB/s: time_to_reach = memory (in GB) /2
+        time_to_reach_secs = int(convert_memory_value_from_k8s_to_units(memory_limit)/2)
+        duration = 100 + time_to_reach_secs
+        self.log.info('Try to allocate 90% available memory')
+        experiment = MemoryStressExperiment(pod=self.target_node, duration=f"{duration}s",
+                                            workers=1, size="90%", time_to_reach=f"{time_to_reach_secs}s")
+        experiment.start()
+        experiment.wait_until_finished()
+
+        self.log.info('Try to allocate 100% total memory')
+        experiment = MemoryStressExperiment(pod=self.target_node, duration=f"{duration}s",
+                                            workers=1, size="100%", time_to_reach=f"{time_to_reach_secs}s")
+        experiment.start()
+        experiment.wait_until_finished()
+
     def disrupt_memory_stress(self):
         """
         Try to run stress-ng to preempt allocated memory of scylla process,
         we don't monitor swap usage in /proc/$scylla_pid/status, just make sure
         no coredump, serious db error occur during the heavy load of memory.
+
         """
+        if self._is_it_on_kubernetes():
+            self._k8s_disrupt_memory_stress()
+            return
         if self.target_node.distro.is_rhel_like:
             self.target_node.install_epel()
             self.target_node.remoter.sudo('yum install -y stress-ng')
--- a/sdcm/utils/k8s/chaos_mesh.py
+++ b/sdcm/utils/k8s/chaos_mesh.py
@@ -16,6 +16,7 @@ import json
 import logging
 from datetime import datetime
 from enum import Enum
+from json import JSONDecodeError
 from tempfile import NamedTemporaryFile
 import time
 
@@ -32,18 +33,19 @@ class ChaosMeshException(Exception):
     pass
 
 
-class PodChaosException(ChaosMeshException):
+class ChaosMeshExperimentException(ChaosMeshException):
 
-    def __init__(self, msg: str, k8s_cluster: "KubernetesCluster", podchaos_name: str, namespace: str):
+    # pylint: disable=too-many-arguments
+    def __init__(self, msg: str, experiment: "ChaosMeshExperiment"):
         super().__init__(msg)
-        self.message = msg
-        self.details = k8s_cluster.kubectl(f"describe podchaos {podchaos_name} -n {namespace}").stdout
+        self.message = f"{msg}. Search debug log about {experiment.name}"
+        LOGGER.debug(experiment.describe())
 
     def __str__(self):
-        return f"{self.message}. Details:\n" + self.details
+        return self.message
 
 
-class PodChaosTimeout(PodChaosException):
+class ChaosMeshTimeout(ChaosMeshExperimentException):
     pass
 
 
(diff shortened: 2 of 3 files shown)


soyacz:
staging job: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/memory-stress-eks/

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -71,14 +71,16 @@ def initialize(self) -> None:
             self._k8s_cluster.POOL_LABEL_NAME, self._k8s_cluster.AUXILIARY_POOL_NAME)
         scylla_node_pool_affinity = get_helm_pool_affinity_values(
             self._k8s_cluster.POOL_LABEL_NAME, self._k8s_cluster.SCYLLA_POOL_NAME)
+        chaos_daemon_settings = scylla_node_pool_affinity | {
+            "runtime": "containerd", "socketPath": "/run/containerd/containerd.sock"}
```
fruch, line 76: is this correct for both EKS and kind ?

soyacz: actually it is not - I added code to detect runtime and adapt accordingly.

sdcm/nemesis.py
```diff
@@ -3537,7 +3538,19 @@ def disrupt_memory_stress(self):
         Try to run stress-ng to preempt allocated memory of scylla process,
         we don't monitor swap usage in /proc/$scylla_pid/status, just make sure
         no coredump, serious db error occur during the heavy load of memory.
+
+        For k8s cluster uses chaos-mesh experiment based on https://github.com/chaos-mesh/memStress
         """
+        if self._is_it_on_kubernetes():
```
fruch, line 3544: Also check if it's enabled in configuration...

sdcm/nemesis.py
```diff
@@ -3537,7 +3538,19 @@ def disrupt_memory_stress(self):
...
+            self.log.info('Try to allocate 120% available memory, the allocated memory will be swaped out')
+            experiment = MemoryStressExperiment(pod=self.target_node, duration="100s", workers=12, size="10%")
```
fruch, line 3546: Why are we using 12 workers here ? with stress-ng we didn't use multiple processes/threads, why do we need it here ? does it make any difference ?

Also why on each command we are using a different number of workers ?

fruch (commented):
Looks good,

One question about workers 
And one configuration check missing

sdcm/nemesis.py
soyacz: I just followed some examples. But apparently it is not working correctly:
image

I'll try with different settings

soyacz: Using approach like with typical nodes (getting available memory from `/proc/meminfo`) is not working correctly - all the memory is not occupied:
image

My guess is that scylla container gets additional memory as needed, so getting `/proc/meminfo` value before scylla occupies all available memory leads to this error..
I'll try with percentage option for this experiment (but it accepts only up to 100%, so 120% case won't be covered) and update here.

soyacz: with 100% limit got similar result, but I found we have something like `k8s_cluster.calculated_memory_limit` which should be right - testing it.

soyacz: result with using `1.2 * k8s_cluster.calculated_memory_limit`
image
Still not all the memory is occupied - I will try few more ways.

soyacz: I tried with memory stress ramp up over 30s, similar result:
image

soyacz: Ok, I think I've found a reason for this behavior: I've read that k8s memory management kills a pod when memory usage raises to quickly. This relates to that 90% mem stress shows higher mem usage as it raises slower than for 120% stress.
When I prolonged stress ramp up to 60s, result looks like this:
image

Still not perfect, but it's getting there.

sdcm/nemesis.py
```diff
@@ -3537,7 +3538,38 @@ def disrupt_memory_stress(self):
...
+
+        For k8s cluster uses chaos-mesh experiment based on https://github.com/chaos-mesh/memStress
         """
+        if self._is_it_on_kubernetes():
+            if not self.tester.k8s_cluster.chaos_mesh.initialized:
+                raise UnsupportedNemesis(
+                    "Chaos Mesh is not installed. Set 'k8s_use_chaos_mesh' config option to 'true'")
+            self.log.info('Try to allocate 120% available memory, the allocated memory will be swaped out')
+            memory_limit = self.tester.k8s_cluster.calculated_memory_limit
+            memory, suffix = int(memory_limit[:-2]), memory_limit[-2:]
```
vponomaryov, line 3550: At first, need to use existing utilities.
For this case we have `convert_memory_value_from_k8s_to_units` function.

At second, I think such conversion must be part of the `MemoryStressExperiment` class itself.
Consumer must not be bothered with such details.
So, consumer should just specified the multiplier.
Example:
```
experiment = MemoryStressExperiment(
    pod=self.target_node,
    duration="160s",
    workers=4,
    memory_multiplier=1.2,
    time_to_reach="60s",
)
```

sdcm/nemesis.py
```diff
@@ -3537,7 +3538,38 @@ def disrupt_memory_stress(self):
...
+            for _ in range(80):
+                mem_stats = self.target_node.remoter.run("cat /sys/fs/cgroup/memory/memory.limit_in_bytes "
+                                                         "/sys/fs/cgroup/memory/memory.max_usage_in_bytes "
+                                                         "/sys/fs/cgroup/memory/memory.usage_in_bytes", verbose=False).stdout
+                LOGGER.info('mem limit/max_used/usage:\n %s', mem_stats)
+                time.sleep(2)
+            experiment.wait_until_finished()
+            self.log.info('Try to allocate 90% total memory, the allocated memory will be swaped out')
+            experiment = MemoryStressExperiment(pod=self.target_node, duration="160s",
+                                                workers=1, size=f"{memory*0.9}{suffix}", time_to_reach="60s")
```
vponomaryov, line 3563: Same here.

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -99,16 +107,18 @@ class ExperimentStatus(Enum):
     UNKNOWN = 5
 
 
-class PodChaosExperiment:
-    """Base class for all PodChaos experiments."""
+class ChaosMeshExperiment:
```
vponomaryov, line 110: Is it really going to be suitable for both types of backends? K8S and VMs?

sdcm/utils/k8s/chaos_mesh.py
```diff
@@ -166,36 +183,70 @@ def wait_until_finished(self):
...
         deep_merge(self._experiment, {
-            "kind": "PodChaos",
             "spec": {
                 "action": "pod-failure",
                 "duration": duration,
             }
         })
+
+
+class MemoryStressExperiment(ChaosMeshExperiment):
```
vponomaryov, line 220: It is K8S- specific, so, I think, it must contain something specific to K8S in it's name.

vponomaryov (changes requested):

sdcm/utils/k8s/chaos_mesh.py
fruch: The name change here, is that experiments would have a base class,

From where you see it should support VMs ?

sdcm/utils/k8s/chaos_mesh.py
fruch: It's a good question, in theory something like memory stress we could use the chaos mesh command line, with the same parameters.

Again we might have different classes, we might implement it in the same class.

For now, we should limit it to only k8s.

We didn't yet tried the commandline in VM to understand of it's a path we want to follow.

sdcm/nemesis.py
soyacz: MemoryStressExperiment supports `%`: we can provide `90%` or in absolute units, so I'll stay with `%` support.
The reason I do the tricks here with memory limits read is due not supporting values >100% by chaos-mesh. 

I tried to keep the test as much as possible to VM based test. Although, I'm not sure if it is relevant, as from what I've read, K8s is rather not using swappiness (from the other hand, when reading `/sys/fs/cgroup/memory/memory.swappiness` it returns 10), is it?. And another question - will k8s allow memory to exceed the limit?
Maybe we can adapt the test and use 100% instead 120% for k8s?

sdcm/utils/k8s/chaos_mesh.py
soyacz: this class inherits from `ChaosMeshExperiment`, chaos-mesh tool is designed only for K8s and this is simple basic information that must be understood by everyone using SCT. Is there special reason that we need to put `k8s` prefix for such basics?

fruch: If there an equivalent one targeted for VM, then yes we probably need some prefix.
 
But currently there isn't, so I don't think it's needed at this stage

soyacz: `Experiment` is term from chaos-mesh. When we develop function that works on all backends, we'll need to change naming completely.

sdcm/nemesis.py
fruch: the way we configured the pods with limit/request identical, k8s doesn't let it use more memory then available.

swap is k8s is a recently new addition (only introduce in 1.22): 
https://kubernetes.io/blog/2021/08/09/run-nodes-with-swap-alpha/

I think doing 100% for now it enough, I guess later when swap is a bit more commonly used in k8s, chaos mesh would support higher then 100%, or some other configuration values to related to swap

sdcm/nemesis.py
soyacz: resutlts when getting 1gb/s ramp up:
image
Still I can see oomKill's in logs.  What I saw from VM based test results, this nemesis kills the scylla too. So, do we care? Can I leave it as it is?

fruch: oomKill that kills scylla process/pod ? that's good, i.e. the system behaves as expected, we should be aware that we'll need to add code to wait for it to recover, before moving on.

in VM we have a few issues that if we do that long enough, scylla doesn't crash but gets unresponsive, and main c-s load fails.  in some cases we only have reactor stalls.
I don't think we got OOM killed

sdcm/utils/k8s/chaos_mesh.py
vponomaryov: The name is not specific to K8S, so, one may think it is suitable for any backend.

soyacz: ChaosMesh is specific to chaos-mesh which is available only on K8s.

sdcm/nemesis.py
soyacz: maybe we should switch order of memory stress size: first make it 90% and then go to 100% wouldn't it be more natural?

soyacz:
@fruch @vponomaryov it's ready for review.
I changed a bit the test flow - first 90% memory stress, then 100% - I just find it more relevant.

Last test is running, when passes I'll recover back longevity test yaml to original state.

soyacz:
ping @fruch @vponomaryov

sdcm/utils/k8s/chaos_mesh.py
fruch: The name is fine for now, 
if we ever use ChaosMesh on VMs, we can revisit the names.

sdcm/utils/k8s/chaos_mesh.py
fruch: I think we can leave the name with prefixing for now, if it would be come relevant we could revisit this.

fruch (approved):
LGTM

vponomaryov (approved):

PR #5614, opened 2022-12-29
Export nemesis v2

roydahan:

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -1043,6 +1043,12 @@ def run_test(argv, backend, config, logdir):
                   failfast=False, buffer=False, catchbreak=True, testLoader=SctLoader())
 
 
+@cli.command('export-nemesis', help="Run SCT test using unittest")
+def export_nemesis():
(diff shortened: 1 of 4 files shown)


fruch:
@roydahan 
we didn't read the docs:
https://docs.python.org/3/library/multiprocessing.html#multiprocessing.set_start_method

seems like we should be calling this only once, so maybe put it in a session fixture

roydahan:
Marking to myself to get back to it

fruch:
@roydahan 

let's write down what are the expectations, and let someone work on it.

I don't see any value in this PR anymore

roydahan:
I think the hydra CLI is something that @pehala is already doing (or already did).
Regarding the MacOS workarounds, I need to check what's needed now.

fruch:
> I think the hydra CLI is something that @pehala is already doing (or already did). Regarding the MacOS workarounds, I need to check what's needed now.

for macos, I know you don't need the changes here, we have a github action that test it

PR #5617, opened 2022-12-30
feature(azure): add azure to cloud report

soyacz:
Currently we don't monitor Azure cloud resources usage. This leads to not needed VM's running without notice and generating costs.

This commit adds Azure backend to cloud usage report.

refs: 
- https://github.com/scylladb/qa-tasks/issues/919

### DIFF
--- a/sdcm/utils/cloud_monitor/resources/instances.py
+++ b/sdcm/utils/cloud_monitor/resources/instances.py
@@ -105,6 +105,9 @@ class AzureInstance(CloudInstance):
 
     def __init__(self, instance: VirtualMachine, resource_group: str):
         tags = instance.tags or {}
+        creation_time = tags.get("creation_time", None)
+        if creation_time:
+            creation_time = datetime.fromisoformat(creation_time).replace(tzinfo=timezone.utc)
         super().__init__(
             cloud="azure",
             name=instance.name,
@@ -114,9 +117,7 @@ class AzureInstance(CloudInstance):
             lifecycle=InstanceLifecycle.SPOT if instance.priority == "Spot" else InstanceLifecycle.ON_DEMAND,
             instance_type=instance.hardware_profile.vm_size,
             owner=tags.get("RunByUser", NA),
-            # azure is not providing vm creation time - for machines that don't have creation_time, set default in the past
-            create_time=datetime.fromisoformat(
-                tags.get("creation_time", "2022-12-20T12:00:00")).replace(tzinfo=timezone.utc),
+            create_time=creation_time,
             keep=tags.get("keep", ""),
             project=resource_group
         )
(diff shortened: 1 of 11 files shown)


sdcm/utils/cloud_monitor/resources/instances.py
```diff
@@ -96,6 +100,32 @@ def get_keep_alive_gce_instance(instance):
...
+        super().__init__(
+            cloud="azure",
+            name=instance.name,
+            instance_id=resource_group,
+            region_az=instance.location,
+            state="running",
+            lifecycle=InstanceLifecycle.SPOT if instance.priority == "Spot" else InstanceLifecycle.ON_DEMAND,
+            instance_type=instance.hardware_profile.vm_size,
+            owner=tags.get("RunByUser", NA),
+            # azure is not providing vm creation time - for machines that don't have creation_time, set default in the past
```
fgelcer, line 117: shouldn't we add the creation time as a label of the instance in this case?

fgelcer (approved):
just a little question, but LGTM

sdcm/utils/cloud_monitor/resources/instances.py
fruch: I think we are doing that, but this is a fallback here, for cases it's an instance wasn't create by SCT (or lost it's tagging for any reason)

I would rather add the needed code to support, create_time=None, and not spoofing a date like that, it would only cause confusion.

sdcm/utils/cloud_monitor/resources/static_ips.py
```diff
@@ -84,6 +98,15 @@ def get_gce_static_ips(self):
                 eip.owner = cloud_instances_by_name[eip.used_by].owner
         self.all.extend(self["gce"])
 
+    def get_azure_static_ips(self):
+        # all azure ip's are external resources and we pay for it
+        query_bits = ["Resources", "where type =~ 'Microsoft.Network/publicIPAddresses'",
```
fruch, line 103: we have code that create static ips for Azure ? in which test case it's being used ?

fruch: "all azure ip's are external" - we don't have internal only addresses in Azure ?

soyacz: before creating VM, external IP must be provided (we could skip it, but then VM wouldn't be accessible from outside).
This external IP lives own life - deleting vm may not delete IP address.

Comment in that terms is not correct, should be:
` all azure public ip's are external resources and we pay for it`
I'll fix it

sdcm/utils/cloud_monitor/resources/instances.py
soyacz: yes, this is fallback for resources not created by SCT.

I'll add code to support create_time=None

soyacz: ok, added - see separate commit for this.

fgelcer (approved):
LGTM

fruch (approved):
LGTM

Issue #5618, opened 2023-01-01
SnapshotOperations nemesis fails on `_validate_snapshot`


fruch:
Since #5606 was merge we started getting those failures
```
2022-12-31 22:15:13.955: (DisruptionEvent Severity.ERROR) period_type=end event_id=4c2b25fe-c4be-498d-b3eb-6488323804b9 duration=30s: nemesis_name=SnapshotOperations target_node=Node longevity-200gb-48h-verify-limited--db-node-a84c3f29-5 [54.174.64.46 | 10.12.10.94] (seed: False) errors=Snapshot content not as expected.
```

since now we are skipping:
system_schema.dropped_columns
system.truncated

and we should skip them, only if we are looking for tables with data.

PR #5619, opened 2023-01-01
fix(get_any_ks_cf_list): skip broken tables when filter_empty_tables

fruch:
Since on cases we want all of the tables (i.e. for comparing with snapshots) we also need to include those system tables, and not skip them.

Fixes: #5618

Ref: https://github.com/scylladb/scylladb/issues/7186
Ref: https://github.com/scylladb/scylladb/issues/12239

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3475,13 +3475,13 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
                 if filter_out_cdc_log_tables and getattr(row, column_names[1]).endswith(cdc.options.CDC_LOGTABLE_SUFFIX):
                     continue
 
-                if table_name in ["system_schema.dropped_columns", "system.truncated"]:
-                    # skipping those cause of some scylla issues on system tables
-                    # https://github.com/scylladb/scylladb/issues/7186
-                    # https://github.com/scylladb/scylladb/issues/12239
-                    continue
-
                 if is_column_type and filter_empty_tables:
+                    if table_name in ["system_schema.dropped_columns", "system.truncated"]:
+                        # skipping those cause of some scylla issues on system tables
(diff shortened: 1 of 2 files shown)


fgelcer (approved):
LGTM

sdcm/cluster.py
```diff
@@ -3475,13 +3475,13 @@ def execute_cmd(cql_session, entity_type):
                 if filter_out_cdc_log_tables and getattr(row, column_names[1]).endswith(cdc.options.CDC_LOGTABLE_SUFFIX):
                     continue
 
-                if table_name in ["system_schema.dropped_columns", "system.truncated"]:
-                    # skipping those cause of some scylla issues on system tables
-                    # https://github.com/scylladb/scylladb/issues/7186
```
juliayakovlev, line 3480: This issue was closed because of it a part of https://github.com/scylladb/scylladb/issues/7175.
So actually it is still open.

sdcm/cluster.py
```diff
@@ -3475,13 +3475,13 @@ def execute_cmd(cql_session, entity_type):
...
-                    # https://github.com/scylladb/scylladb/issues/12239
```
juliayakovlev, line 3481: same here

sdcm/cluster.py
fruch: I'm not sure I'm follwoing, I know those are still open, I've just move them below.

the issue with them is only when we try reading from those tables

juliayakovlev: Ignore it. I am after New Year celebration

juliayakovlev (approved):

sdcm/cluster.py
fruch: It's probably me that didn't drunk enough :)

PR #5621, opened 2023-01-01
fix(nemesis): remove exception prints from `execute_disrupt_method`

fruch:
since now we raise the `UnsupportedNemesis`, there no need to print it again like this in error level, we have a warning for it

```
...
   File ".../sdcm/nemesis.py", line 1647, in execute_disrupt_method
     disrupt_method()
   File ".../sdcm/nemesis.py", line 3848, in wrapper
   File ".../sdcm/nemesis.py", line 2848, in disrupt_network_random_interruptions
     raise UnsupportedNemesis("for this nemesis to work,
                               you need to set `extra_network_interface: True`")
 sdcm.nemesis.UnsupportedNemesis: for this nemesis to work,
                                  you need to set `extra_network_interface: True`
```

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1645,12 +1645,6 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self.metrics_srv.event_start(disrupt_method_name)
         try:
             disrupt_method()
-        except Exception as exc:  # pylint: disable=broad-except
-            error_msg = "Exception in random_disrupt_method %s: %s", disrupt_method_name, exc
-            self.log.error(error_msg, exc_info=True)
-            self.error_list.append(error_msg)
-            raise
-        else:
             self.log.info("<<<<<<<<<<<<<Finished random_disrupt_method %s" % disrupt_method_name)
         finally:
             self.metrics_srv.event_stop(disrupt_method_name)


sdcm/nemesis.py
```diff
@@ -1645,12 +1645,6 @@ def execute_disrupt_method(self, disrupt_method):
         self.metrics_srv.event_start(disrupt_method_name)
         try:
             disrupt_method()
-        except Exception as exc:  # pylint: disable=broad-except
-            error_msg = "Exception in random_disrupt_method %s: %s", disrupt_method_name, exc
-            self.log.error(error_msg, exc_info=True)
-            self.error_list.append(error_msg)
-            raise
-        else:
```
fgelcer, line 1653: while removing it, you are not raising it anymore... is that intentional?

fruch: I'm not catching any exception, so they would be raised automatically

fgelcer (approved):
LGTM

sdcm/nemesis.py
```diff
@@ -1645,12 +1645,6 @@ def execute_disrupt_method(self, disrupt_method):
         self.metrics_srv.event_start(disrupt_method_name)
         try:
             disrupt_method()
-        except Exception as exc:  # pylint: disable=broad-except
```
soyacz, line 1648: how about just adding:
```
except UnsupportedNemesis:
    raise
except Exception as exc:
...
```
so we don't change current logic and keep `UnsupportedNemesis` more quiet.

fruch: there's no logic here, `error_list` isn't really being used (and the same error already gets appended else where)
and the debug print is just adding more confusion, i.e. same error printed again and again.

fruch: on only reason for adding `except UnsupportedNemesis:` is for documentation, so people won't remove it

soyacz (approved):

vponomaryov (approved):

PR #5622, opened 2023-01-01
fix(k8s-ingress): align configuration to production

fruch:
After reviewing the production configuration, we
should be aligned to how it's use in there for our testing


sdcm/tester.py
```diff
@@ -1562,7 +1562,7 @@ def get_cluster_k8s_eks(self):  # pylint: disable=too-many-statements
         self.k8s_cluster.deploy_node_pool(
             eks.EksNodePool(
                 name=self.k8s_cluster.AUXILIARY_POOL_NAME,
-                num_nodes=2,
+                num_nodes=3,
```
fruch, line 1565: I guess we should have configuration for this, since in non serverless runs, we don't need this extra node.

soyacz: There's more about it  - I suppose in larger deployments we'll need to scale it out/up. I'll be verifying it in many-clients-test (and if needed I'll fix it).

vponomaryov: I agree that we should have a config option for it if we want to have more than 1 value here.

sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.deploy.yaml
```diff
@@ -22,6 +22,15 @@ spec:
...
       serviceAccountName: haproxy-kubernetes-ingress
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+            - podAffinityTerm:
+                labelSelector:
+                  matchLabels:
+                    app.kubernetes.io/name: haproxy-ingress
+                topologyKey: kubernetes.io/hostname
+              weight: 1
```
vponomaryov, line 33: Why do you need it?

sdcm/cluster_k8s/__init__.py
```diff
@@ -1025,10 +1025,34 @@ def deploy_ingress_controller(self, pool_name: str = None):
...
+            if self.params.get("cluster_backend") == "k8s-eks":
+                cpu_limit = 2
+                memory_limit = 4
+                cpu_request = 1
+                memory_request = 2
+
+                cpu_limit = convert_cpu_units_to_k8s_value(cpu_limit)
+                memory_limit = convert_memory_units_to_k8s_value(memory_limit)
+                cpu_request = convert_cpu_units_to_k8s_value(cpu_request)
+                memory_request = convert_memory_units_to_k8s_value(memory_request)
```
vponomaryov, line 1038: JFYI:
If we set `request` and `limit` value as the same then we get `dedicated` resources and better performance.
If we set `request` and `limit` to different values like in this change then we get `context switch` situation and it worse from perf point of view.

Considering that one more `defaul` node gets created I would expect to have the first variant here - with better perf.

vponomaryov (changes requested):

sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.deploy.yaml
fruch: I don't know, but that was part of the diff we had with the cloud team configuration...

sdcm/cluster_k8s/__init__.py
fruch: And again this is how the cloud team configuration is, and I wanted that our default would be identical.

sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.deploy.yaml
vponomaryov: It is garbage in our case. We set our own affinity rules making it be scheduled onto the "default" node pool.
Remove it please.

sdcm/cluster_k8s/__init__.py
vponomaryov: I don't mind. Let it be so.

sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.deploy.yaml
fruch: we'll remove

sdcm/tester.py
fruch: create a configuration for it

sdcm/k8s_configs/ingress-controller/10_ingress-default-backend.deploy.yaml
```diff
@@ -22,5 +22,17 @@ spec:
...
+            memory: 50Mi
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  app.kubernetes.io/name: haproxy-ingress
+              topologyKey: kubernetes.io/hostname
+            weight: 1
```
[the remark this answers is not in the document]
fruch: removed

sdcm/k8s_configs/ingress-controller/10_haproxy-kubernetes-ingress.deploy.yaml
fruch: removed

sdcm/tester.py
```diff
@@ -1562,7 +1562,7 @@ def get_cluster_k8s_eks(self):  # pylint: disable=too-many-statements
         self.k8s_cluster.deploy_node_pool(
             eks.EksNodePool(
                 name=self.k8s_cluster.AUXILIARY_POOL_NAME,
-                num_nodes=2,
+                num_nodes=self.params.get('k8s_n_auxiliary_nodes'),
```
vponomaryov, line 1565: Why only EKS? GKE also creates node pools of the same types

vponomaryov (changes requested):

sdcm/tester.py
fruch: added it also to GKE

vponomaryov (dismissed):
LGTM

sdcm/sct_config.py
```diff
@@ -905,6 +905,9 @@ class SCTConfiguration(dict):
         dict(name="k8s_use_chaos_mesh", env="SCT_K8S_USE_CHAOS_MESH", type=boolean,
              help="""enables chaos-mesh for k8s testing"""),
 
+        dict(name="k8s_n_auxiliary_nodes", env="SCT_K8S_N_AUXILIARY_NODES", type=int,
+             help="Number of of nodes in auxiliary pool"),
```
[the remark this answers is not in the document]
fruch: fixed

docs/configuration_options.md
```diff
@@ -210,6 +210,7 @@
 | **<a href="#user-content-k8s_enable_tls" name="k8s_enable_tls">k8s_enable_tls</a>**  | Defines whether the we enable the operator serverless options | N/A | SCT_K8S_ENABLE_TLS
 | **<a href="#user-content-k8s_connection_bundle_file" name="k8s_connection_bundle_file">k8s_connection_bundle_file</a>**  | Serverless configuration bundle file | N/A | SCT_K8S_CONNECTION_BUNDLE_FILE
 | **<a href="#user-content-k8s_use_chaos_mesh" name="k8s_use_chaos_mesh">k8s_use_chaos_mesh</a>**  | enables chaos-mesh for k8s testing | N/A | SCT_K8S_USE_CHAOS_MESH
+| **<a href="#user-content-k8s_n_auxiliary_nodes" name="k8s_n_auxiliary_nodes">k8s_n_auxiliary_nodes</a>**  | Number of of nodes in auxiliary pool | N/A | SCT_K8S_N_AUXILIARY_NODES
```
vponomaryov, line 213: That one is not fixed, but ok ))

PR #5624, opened 2023-01-01
improvement(longevity-large-partition-8h): create mview after prepare

ShlomiBalalis:


fruch:
@ShlomiBalalis please add some context, as to why we need to do this change, and why it needs to be backported.

fgelcer (changes requested):
> @ShlomiBalalis please add some context, as to why we need to do this change, and why it needs to be backported.

agree, more context is needed for such changes

ShlomiBalalis:
I first made this change to create a specific scenario for a 2022.1.4 issue. Afterwards, @roydahan requested that I'll push it to master as well.

fruch:
> I first made this change to create a specific scenario for a 2022.1.4 issue. Afterwards, @roydahan requested that I'll push it to master as well.

regardless it should be communicated with more details (even if @roydahan would see this PR, he won't necessarily remember what it's for, even if he asked for it)

roydahan:
And I don't :)

roydahan:
But IIUC it's adding a MV for large partition table (which could be enough for the commit message).

fruch:
I'm still missing the *why* are we doing this, is this trying to overcome some bug ? It's trying to expose some issue ?

scylla-qabot (approved):

PR #5625, opened 2023-01-01
feature(sct-builders): move to use gcp launch template

fruch:
* added `GceBuilder` to create templates and configure them in jenkins

* move the default labels for GCE backend to use those templates

## Testing

- [x] longevity -https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-10gb-3h-gce-test/22
- [x] artifact - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ubuntu2204-test/27/
- [ ] ~~artifact docker~~ docker would run on AWS
- [x] test multiple artifacts runs sharing the same builder (4 as we did in AWS)

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -84,11 +84,12 @@ from sdcm.monitorstack import (restore_monitoring_stack, get_monitoring_stack_se
                                kill_running_monitoring_stack_services)
 from sdcm.utils.log import setup_stdout_logger
 from sdcm.utils.aws_region import AwsRegion
+from sdcm.utils.aws_builder import AwsCiBuilder, AwsBuilder
 from sdcm.utils.gce_region import GceRegion
+from sdcm.utils.gce_builder import GceBuilder
 from sdcm.utils.aws_peering import AwsVpcPeering
 from sdcm.utils.get_username import get_username
 from sdcm.utils.sct_cmd_helpers import add_file_logger, CloudRegion, get_test_config, get_all_regions
-from sdcm.utils.aws_builder import configure_jenkins_builders
 from sdcm.send_email import get_running_instances_for_email_report, read_email_data_from_file, build_reporter, \
     send_perf_email
 from sdcm.parallel_timeline_report.generate_pt_report import ParallelTimelinesReportGenerator
@@ -1607,7 +1608,23 @@ def create_es_index(name: str, doc_type: str, mapping_file: str) -> None:
     create_index(index_name=name, doc_type=doc_type, mappings=mapping_data)
 
 
-cli.add_command(configure_jenkins_builders)
+@cli.command("configure-jenkins-builders", help="Configure all required jenkins builders for SCT")
+@cloud_provider_option
+@click.option("-r", "--regions", type=CloudRegion(), default=[], help="Cloud regions", multiple=True)
+def configure_jenkins_builders(cloud_provider, regions):
+    add_file_logger()
+    logging.basicConfig(level=logging.INFO)
+
+    match cloud_provider:
+        case 'aws':
+            AwsCiBuilder(AwsRegion('eu-west-1')).configure_auto_scaling_group()
+            AwsBuilder.configure_in_all_region(regions=regions)
+        case 'gce':
+            GceBuilder.configure_in_all_region(regions=regions)
+        case 'azure':
+            raise NotImplementedError("configure_jenkins_builders doesn't support Azure yet")
+
+
 cli.add_command(sct_ssh.ssh)
 cli.add_command(sct_ssh.tunnel)
 cli.add_command(sct_ssh.copy_cmd)
--- a/sdcm/utils/gce_builder.py
+++ b/sdcm/utils/gce_builder.py
@@ -1,0 +1,239 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2023 ScyllaDB
+import sys
+import logging
+from typing import Any
+from functools import cached_property
+
+import click
+import requests
+from google.oauth2 import service_account
+from google.cloud import compute_v1
+from google.api_core.extended_operation import ExtendedOperation
+import google.api_core.exceptions
+
+from sdcm.utils.gce_region import GceRegion
+from sdcm.utils.gce_utils import SUPPORTED_PROJECTS, SUPPORTED_REGIONS
+from sdcm.utils.context_managers import environment
+from sdcm.sct_runner import GceSctRunner
+from sdcm.keystore import KeyStore
+
+LOGGER = logging.getLogger(__name__)
+
+JENKINS_CONFIG_FORMAT = """
+// just modify this config other code just logic
+config = [
+    name: "{name}",
+    region: "{region_name}",
+    projectId: "{project_id}",
+    credentialsId: "{project_id}",
+    instanceCapStr: "{max_instances}",
+    labels: "{jenkins_labels}",
+]
+
+import com.google.jenkins.plugins.computeengine.ComputeEngineCloud
+import com.google.jenkins.plugins.computeengine.InstanceConfiguration
+import com.google.jenkins.plugins.computeengine.SshConfiguration
+import hudson.plugins.sshslaves.SSHConnector
+import hudson.plugins.sshslaves.verifiers.NonVerifyingKeyVerificationStrategy
+import hudson.model.*
+import jenkins.model.Jenkins
+
+ComputeEngineCloud gcpCloud = new ComputeEngineCloud(config.name, config.projectId, config.credentialsId, config.instanceCapStr)
+gcpCloud.setNoDelayProvisioning(true)
+
+InstanceConfiguration instanceConfiguration = new InstanceConfiguration()
+instanceConfiguration.setNumExecutorsStr("4")
+instanceConfiguration.setOneShot(false)
+instanceConfiguration.setNamePrefix("builders")
+instanceConfiguration.setDescription("SCT Jenkins builders")
+instanceConfiguration.setRegion("https://www.googleapis.com/compute/v1/projects/{project_id}/regions/{region_name}")
+instanceConfiguration.setZone("https://www.googleapis.com/compute/v1/projects/{project_id}/zones/{region_name}-b")
+instanceConfiguration.setLabelString(config.labels)
+instanceConfiguration.setMode(Node.Mode.NORMAL)
+instanceConfiguration.setTemplate("https://www.googleapis.com/compute/v1/projects/{project_id}/global/instanceTemplates/{name}")
+instanceConfiguration.setRetentionTimeMinutesStr("6")
+instanceConfiguration.setLaunchTimeoutSecondsStr("300")
+instanceConfiguration.setRunAsUser("jenkins")
+instanceConfiguration.setSshConfiguration(SshConfiguration.builder().customPrivateKeyCredentialsId("6e17c350-0a47-4a59-b0b9-7e64f1c35a65").build())
+
+gcpCloud.addConfiguration(instanceConfiguration)
+
+// get Jenkins instance
+Jenkins jenkins = Jenkins.get()
+
+// clear old configuration
+jenkins.clouds.removeAll {{ it.name == "gce-${{config.name}}" }}
+
+res = jenkins.clouds.add(gcpCloud)
+
+// save current Jenkins state to disk
+jenkins.save()
+"""
+
+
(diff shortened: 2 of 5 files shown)


sdcm/utils/cdc/gce_builder.py
```diff
@@ -0,0 +1,250 @@
...
+                region.configure()
+
+
+configure_jenkins_builders: click.Command
+
+
+@click.command("configure-jenkins-builders", help="Configure all required jenkins builders for SCT")
+@click.option("-r", "--regions", type=CloudRegion(cloud_provider='gce'),
+              default=[], help="Cloud regions", multiple=True)
+def configure_jenkins_builders(regions):
```
fruch, line 244: still need to align this code with the AWS configure_jenkins_builders,
to make it one command ?  or multiple commands ?

fgelcer: this could be perhaps an interface for each different backend, so the caller can just use the function regardless the backend in use.. similar to the unified provision.. and @KnifeyMoloko is still pending to finish the GCE part

sdcm/utils/cdc/gce_builder.py
```diff
@@ -0,0 +1,250 @@
+# This program is free software; you can redistribute it and/or modify
```
soyacz, line 1: why this file is in `cdc` folder?

sdcm/utils/cdc/gce_builder.py
```diff
@@ -0,0 +1,250 @@
...
+jenkins.clouds.removeAll {{ it.name == "gce-${{config.name}}" }}
+
+res = jenkins.clouds.add(gcpCloud)
+
+// save current Jenkins state to disk
+jenkins.save()
+"""
+
+
+class GceBuilder:
```
soyacz, line 86: please add some doc related this class responsibility

sdcm/utils/cdc/gce_builder.py
fruch: yes that the intended direction, once I have it working as I expect

Any this is used only when configuration need to be altered, hopefully it won't happen that often.

sdcm/utils/cdc/gce_builder.py
fruch, line 1: how in *** sake it's in the cdc directory ?

sdcm/utils/cdc/gce_builder.py
```diff
@@ -0,0 +1,250 @@
...
+    def configure(self):
+        self.create_launch_template()
+        self.add_cloud_configuration_to_jenkins()
+
+    @classmethod
+    def configure_in_all_region(cls, regions=None):
+        for project in SUPPORTED_PROJECTS:
+            os.environ['SCT_GCE_PROJECT'] = project
+
+            regions = regions or ['us-east1', 'us-west1']
```
soyacz, line 232: let's use `SUPPORTED_REGIONS` const

sdcm/utils/cdc/gce_builder.py
```diff
@@ -0,0 +1,250 @@
...
+            for region_name in regions:
+                region = cls(GceRegion(region_name))
+                region.configure()
+
+
+configure_jenkins_builders: click.Command
+
+
+@click.command("configure-jenkins-builders", help="Configure all required jenkins builders for SCT")
+@click.option("-r", "--regions", type=CloudRegion(cloud_provider='gce'),
```
soyacz, line 242: type here is actually string.
Let's add info that if not supplied, configures in all supported regions.

sdcm/utils/cdc/gce_builder.py
```diff
@@ -0,0 +1,250 @@
...
+        self.add_cloud_configuration_to_jenkins()
+
+    @classmethod
+    def configure_in_all_region(cls, regions=None):
+        for project in SUPPORTED_PROJECTS:
+            os.environ['SCT_GCE_PROJECT'] = project
+
+            regions = regions or ['us-east1', 'us-west1']
+            for region_name in regions:
+                region = cls(GceRegion(region_name))
```
soyacz, line 234: let's not name `region` for `GceBuilder` class - it's really confusing. Maybe `region_builder`?

sdcm/utils/cdc/gce_builder.py
```diff
@@ -0,0 +1,250 @@
...
+
+            self.wait_for_extended_operation(operation, "instance template creation")
+        except google.api_core.exceptions.Conflict:
+            pass
+
+        return template_client.get(project=self.region.project, instance_template=self.name)
+
+    def configure(self):
+        self.create_launch_template()
+        self.add_cloud_configuration_to_jenkins()
```
soyacz, line 225: these methods (actually all of them except `configure`) should be private. Let's make the api clear

sdcm/utils/cdc/gce_builder.py
fruch: moved it to be shared with the AWS command

sdcm/utils/cdc/gce_builder.py
fruch: moved it to the right place

sdcm/utils/cdc/gce_builder.py
fruch: this code moved/removed

sdcm/utils/cdc/gce_builder.py
fruch: fixed

sdcm/utils/cdc/gce_builder.py
fruch: changed to private

sdcm/utils/cdc/gce_builder.py
fruch: added the description

sdcm/utils/cdc/gce_builder.py
fruch: change to use that list

sct.py
```diff
@@ -1594,7 +1595,23 @@ def generate_parallel_timelines_report(logdir: str | None, test_id: str | None)
     pt_report_generator.generate_full_report()
 
 
-cli.add_command(configure_jenkins_builders)
+@cli.command("configure-jenkins-builders", help="Configure all required jenkins builders for SCT")
+@cloud_provider_option
+@click.option("-r", "--regions", type=CloudRegion(), default=[], help="Cloud regions", multiple=True)
```
vponomaryov, line 1600: Are you sure that exactly instance of the `CloudRegion` class must be used here?
Maybe it must be exactly class itself?

sct.py
```diff
@@ -1594,7 +1595,23 @@ def generate_parallel_timelines_report(logdir: str | None, test_id: str | None)
...
+def configure_jenkins_builders(cloud_provider, regions):
+    add_file_logger()
+    logging.basicConfig(level=logging.INFO)
+
+    match cloud_provider:
+        case 'aws':
+            AwsCiBuilder(AwsRegion('eu-west-1')).configure_auto_scaling_group()
```
vponomaryov, line 1607: Don't we want to use first region from the `regions` var instead of hardcoded `eu-west-1`?

sdcm/utils/gce_builder.py
```diff
@@ -0,0 +1,239 @@
...
+instanceConfiguration.setDescription("SCT Jenkins builders")
+instanceConfiguration.setRegion("https://www.googleapis.com/compute/v1/projects/{project_id}/regions/{region_name}")
+instanceConfiguration.setZone("https://www.googleapis.com/compute/v1/projects/{project_id}/zones/{region_name}-b")
+instanceConfiguration.setLabelString(config.labels)
+instanceConfiguration.setMode(Node.Mode.NORMAL)
+instanceConfiguration.setTemplate("https://www.googleapis.com/compute/v1/projects/{project_id}/global/instanceTemplates/{name}")
+instanceConfiguration.setRetentionTimeMinutesStr("6")
+instanceConfiguration.setLaunchTimeoutSecondsStr("300")
+instanceConfiguration.setRunAsUser("jenkins")
+instanceConfiguration.setSshConfiguration(SshConfiguration.builder().customPrivateKeyCredentialsId("6e17c350-0a47-4a59-b0b9-7e64f1c35a65").build())
```
vponomaryov, line 67: Shouldn't the creds ID be in some config? Especially when you say above that here we have `just logic`?

sdcm/utils/gce_builder.py
```diff
@@ -0,0 +1,239 @@
...
+// save current Jenkins state to disk
+jenkins.save()
+"""
+
+
+class GceBuilder:
+    """
+    This class is for configuration our Jenkins setup for GCE
+
+    It created a launch template based on sct-runner image, and adds configuration needed in Jenkins to use it
```
vponomaryov, line 88: s/created/creates/

sdcm/utils/gce_builder.py
```diff
@@ -0,0 +1,239 @@
...
+    """
+
+    def __init__(self, region: GceRegion, number=1):
+        self.region = region
+        self.number = number
+        self.jenkins_info = KeyStore().get_json("jenkins.json")
+        self.runner = GceSctRunner(region_name=self.region.region_name,
+                                   availability_zone="a")
```
vponomaryov, line 96: Why we want to stick to a specific az? Is it a problem to allow to set it?

sdcm/utils/gce_builder.py
```diff
@@ -0,0 +1,239 @@
...
+        template.properties.metadata = metadata
+        template_client = compute_v1.InstanceTemplatesClient(credentials=self.credentials)
+        try:
+            operation = template_client.insert(
+                project=self.region.project, instance_template_resource=template
+            )
+
+            self._wait_for_extended_operation(operation, "instance template creation")
+        except google.api_core.exceptions.Conflict:
+            pass
```
vponomaryov, line 223: Maybe we want to print some useful message in this case?

sdcm/utils/gce_builder.py
```diff
@@ -0,0 +1,239 @@
...
+        return template_client.get(project=self.region.project, instance_template=self.name)
+
+    def configure(self):
+        self._create_launch_template()
+        self._add_cloud_configuration_to_jenkins()
+
+    @classmethod
+    def configure_in_all_region(cls, regions=None):
+        for project in SUPPORTED_PROJECTS:
+            os.environ['SCT_GCE_PROJECT'] = project
```
vponomaryov, line 234: Are you sure that usage of this env var won't trigger any problem like before with cleanup?
Maybe better to specify it somehow else?

sct.py
fruch: yes I'm sure, this is how this class is being used

sct.py
fruch: no, this is the CI builder, and we want them in specific region

fruch: i.e. this the ASG used for SCT CI

sdcm/utils/gce_builder.py
fruch: it's only cause the key we need doesn't have meaningful name in Jenkins
we have keys similar to this one scattered in our pipelines as well (see `b8a774da-0e46-4c91-9f74-09caebaea261`)

it doesn't belongs into sct configuration, the only place I might thing of is our s3 bucket.
anyhow this would be tied to what's configured in our Jenkins server, regardless of where we'll put it.

sdcm/utils/gce_builder.py
fruch: it's pointless, we pass it along only since `GceSctRunner` mandate it.
we using this class only to extract the image name of the sct-runner to put in the launch template

sdcm/utils/gce_builder.py
fruch: this mean it exists, if it's exist we don't recreate it, and move on to the next step.
not very critical thing to log, also we don't really have a log for those hydra actions.

sdcm/utils/gce_builder.py
fruch: it's less prone for this error since it's doing just on specific call across all of the project,
the cleanup had multiple steps that wasn't taking into account.

I would be nice to use the same machinery

but this is a manual action someone would be doing, to update/rebuild the needed configuration, not something happening on each sct job, so the price of failure is lower, and someone would be attending to it straight away

sdcm/utils/gce_builder.py
fruch: fixed

sdcm/utils/gce_builder.py
fruch: fixed

sdcm/utils/gce_builder.py
vponomaryov: o.k.

sdcm/utils/gce_builder.py
vponomaryov: I didn't talk about `log`. I talked about `print` like you have above.
I understand the code, I wanted to highlight the fact that simple printed message that it already exists may be useful. It may be valueable to understand it, especially when we update the code fixing something.

fruch: also probably would want to replace the `print` with `click.secho`... o.k. I'll change

vponomaryov (dismissed):
LGTM

soyacz (dismissed):

PR #5585, opened 2022-12-21
feature(sct-builders): move builder to auto scaling group

fruch:
Since we are not using those builder all the time, we are starting to create them as each time we start a job.

* very small machine that are only use to spin off the test
* in some cases it should be able to run the test (artifact test or SCT CI)
* this is creating the ASGs and launch templates in all regions we operate
* configure those ASGs in jenkins
* switch all monitors to ubuntu jammy, since the old centos images
  we use aren't available anymore across all regions

Testing:
- [x] CI - both integration test / docker backend
- [x] longevity - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/3/
- [x] longevity - run on new region `ca-central-1` - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/10/
- [x] test artifact test - docker - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-docker-test/17/
- [x] show monitor job - test and switch to same as CI
- [x] cloud report - 
- [ ] azure -
- [ ] azure artifact test -

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5630, opened 2023-01-02
fix(integration-test): fixes to stabilize integration tests

fruch:
during working on #5585, we needed to run those test on a weaker machines, and those
are the result of that work
  
### DIFF
--- a/unit_tests/test_version_utils.py
+++ b/unit_tests/test_version_utils.py
@@ -397,7 +397,7 @@ def test_scylla_version_grouped_regexp(full_version, version, date, commit_id):
 @pytest.mark.parametrize("version, expected", (
     ("4.6.rc2-20220102.e8a1cfb6f", "scylladb/scylla:4.6.rc2"),
     ("5.0.1-0.20220719.b177dacd3", "scylladb/scylla:5.0.1"),
-    ("5.1", "scylladb/scylla:latest"),
+    ("5.1", "scylladb/scylla:5.1"),
     ("5.1.0~rc1", "scylladb/scylla:5.1.0-rc1"),
     ("5.0.1", "scylladb/scylla:5.0.1"),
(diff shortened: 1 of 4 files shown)


soyacz (approved):
LGTM

unit_tests/test_version_utils.py
```diff
@@ -397,7 +397,7 @@ def test_scylla_version_grouped_regexp(full_version, version, date, commit_id):
 @pytest.mark.parametrize("version, expected", (
     ("4.6.rc2-20220102.e8a1cfb6f", "scylladb/scylla:4.6.rc2"),
     ("5.0.1-0.20220719.b177dacd3", "scylladb/scylla:5.0.1"),
-    ("5.1", "scylladb/scylla:latest"),
+    ("5.1", "scylladb/scylla:5.1"),
```
vponomaryov, line 400: Need to add following:

```
("5.2", "scylladb/scylla:latest"),
```

vponomaryov (changes requested):

unit_tests/test_version_utils.py
fruch: why ? for it to fail once they would start the 5.2 release ?
I think this is a bit pointless

vponomaryov: I know that we will hit the same problem again, but it is coverage of the functionality.
We either test that the next version is refering to "latest" one or not.

fruch: right now 5.2 doesn't exist, we already have a test covering that.

I'm don't understand what adding this new test would help us ?

vponomaryov: It is for testing mapping of next unavilable version yet and `latest` tag.
If you think we do not need then ok.

vponomaryov (approved):

Issue #5636, opened 2023-01-03
Stop using Docker-shim in our K8S runs with 1.24+ versions


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: any
* **scylla**: any
<!--- other version that might apply -->

### Logs
* test_id: e9d89612-f8d1-4e5d-afc2-48550f8f7f66
* job log: https://jenkins.scylladb.com/view/scylla-operator/job/scylla-operator/job/operator-1.8/job/upgrade/job/upgrade-operator-k8s-eks/4
<!--- other logs that might apply -->

### Description

K8S loaders fail with the following error when we use K8S 1.24:

```
MountVolume.SetUp failed for volume "docker-socket" : hostPath type check failed: /var/run/docker.sock is not a socket file                                                                                                                                                                                                                                                
```

The reason for it is [official removal of docker-shim support from K8S 1.24+ versions](https://kubernetes.io/blog/2022/02/17/dockershim-faq/#can-i-still-use-docker-engine-in-kubernetes-1-23).

So, we should stop using it ASAP.

Fast solution is to switch all loaders to be `dynamic`.

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:** K8S loaders must work ok on any K8S version.

**Actual behavior:** it fails on the attempt to mount docker socket which is unavailable on K8S 1.24+ versions.

fruch:
How come my run didn't failed on this ?

https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-scylla-operator-3h-gke/8/

vponomaryov:
> How come my run didn't failed on this ?
> 
> https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-scylla-operator-3h-gke/8/

`k8s_loader_run_type: dynamic`

fruch:
> > How come my run didn't failed on this ?
> > 
> > https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-scylla-operator-3h-gke/8/
> 
> `k8s_loader_run_type: dynamic`

I've forgotten we default to it in GKE.

And my own test for EKS failed the same, 
[...]

PR #5637, opened 2023-01-03
fix(k8s): make static loaders work on K8S v1.24+

vponomaryov:
Starting with K8S `v1.24` the docker-shim support was removed. 
So, stop using docker for our static loaders because we are going to switch to new K8S versions.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/5636

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4789,8 +4789,7 @@ class BaseLoaderSet():
         return next(self._loader_cycle)
 
     def kill_stress_thread(self):
-        if self.nodes and self.nodes[0].is_kubernetes() and self.params.get(
-                'k8s_loader_run_type') != 'static':
+        if self.nodes and self.nodes[0].is_kubernetes():
             for node in self.nodes:
                 node.remoter.stop()
         else:
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -1686,7 +1686,7 @@ class BasePodContainer(cluster.BaseNode):  # pylint: disable=too-many-public-met
 
     def _init_remoter(self, ssh_login_info):
         self.remoter = KubernetesCmdRunner(kluster=self.parent_cluster.k8s_cluster,
-                                           pod=self.name,
+                                           pod_name=self.name,
                                            container=self.parent_cluster.container,
                                            namespace=self.parent_cluster.namespace)
 
--- a/sdcm/k8s_configs/loaders/sts.yaml
+++ b/sdcm/k8s_configs/loaders/sts.yaml
@@ -31,21 +31,4 @@ spec:
             requests:
               cpu: ${POD_CPU_LIMIT}
               memory: ${POD_MEMORY_LIMIT}
-          # TODO: those mounts/securityContext should be remove once we'll stop using bare docker for loaders
-          securityContext:
-            privileged: true
-          volumeMounts:
-            - mountPath: /usr/bin/docker
-              name: docker-binary
-            - mountPath: /var/run/docker.sock
-              name: docker-socket
-      volumes:
-        - name: docker-binary
-          hostPath:
-            path: /usr/bin/docker
-            type: File
-        - name: docker-socket
-          hostPath:
-            path: /var/run/docker.sock
-            type: Socket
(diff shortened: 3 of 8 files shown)


fruch:
I understand removing the docker mounts.

But I'm not following how the rename of pod to pod_name is related.

and why we need to change all the logic in loaders classes, while we are still not removing the static option.

That means static would still work, but only for c-s ?

vponomaryov:
> But I'm not following how the rename of pod to pod_name is related.
It is done to make the same code be working for both loader run types.
Look at changes at
- sdcm/stress_thread.py
- sdcm/scylla_bench_thread.py

> and why we need to change all the logic in loaders classes, while we are still not removing the static option.

Where do you the change of logic in loader classes? Static option gets fixed to be what it should have been all the time.

> That means static would still work, but only for c-s ?

It will work as it was before - C-S and S-B. Both are fixed.
I tested only C-S, yes, but S-B must work too. If not, then some small typo may be in some place.

fruch:
> > But I'm not following how the rename of pod to pod_name is related.
> 
> It is done to make the same code be working for both loader run types. Look at changes at
> 
> * sdcm/stress_thread.py
> * sdcm/scylla_bench_thread.py
> 
> > and why we need to change all the logic in loaders classes, while we are still not removing the static option.
> 
> Where do you see the change of logic in loader classes? Static option gets fixed to be what it should have been all the time.
> 
> > That means static would still work, but only for c-s ?
> 
> It will work as it was before - C-S and S-B. Both are fixed. I tested only C-S, yes, but S-B must work too. If not, then some small typo may be in some place.

I can't warp my head how s-b on static is going to work, we need a different docker image..

vponomaryov:
> > > But I'm not following how the rename of pod to pod_name is related.
> > 
> > 
> > It is done to make the same code be working for both loader run types. Look at changes at
> > 
> > * sdcm/stress_thread.py
> > * sdcm/scylla_bench_thread.py
> > 
> > > and why we need to change all the logic in loaders classes, while we are still not removing the static option.
> > 
> > 
> > Where do you see the change of logic in loader classes? Static option gets fixed to be what it should have been all the time.
> > > That means static would still work, but only for c-s ?
> > 
> > 
> > It will work as it was before - C-S and S-B. Both are fixed. I tested only C-S, yes, but S-B must work too. If not, then some small typo may be in some place.
> 
> I can't warp my head how s-b on static is going to work, we need a different docker image..

Ok, I forgot that we do not install it anymore explicitly.
Then yes, you are right. The S-B with fail due to absence of the expected binary.

fruch:
> > > > But I'm not following how the rename of pod to pod_name is related.
> > > 
> > > 
> > > It is done to make the same code be working for both loader run types. Look at changes at
> > > 
> > > * sdcm/stress_thread.py
> > > * sdcm/scylla_bench_thread.py
> > > 
> > > > and why we need to change all the logic in loaders classes, while we are still not removing the static option.
> > > 
> > > 
> > > Where do you see the change of logic in loader classes? Static option gets fixed to be what it should have been all the time.
> > > > That means static would still work, but only for c-s ?
> > > 
> > > 
> > > It will work as it was before - C-S and S-B. Both are fixed. I tested only C-S, yes, but S-B must work too. If not, then some small typo may be in some place.
> > 
> > I can't warp my head how s-b on static is going to work, we need a different docker image..
> 
> Ok, I forgot that we do not install it anymore explicitly.
> Then yes, you are right. The S-B with fail due to absence of the expected binary.
> 
> So, for running `s-b` commands from nemesis which do it we may want to reuse there code from dynamic loaders.
> It must be small change.
> 
> **Upd**: as I can see, only one nemesis is affected - `disrupt_truncate_large_partition`.
> So, following options are available here:
> - temporary install `s-b` on the static loaders pods
> - temporary skip this nemesis for K8S.
> - reuse dynamic loader class there and test it
> - create additional class for handling separate `s-b` calls like we have for `helm` or `gcloud`.


Let's move to use dynamic loader for EKS longevity, and maybe only the perf tests would still use static

And open a task for using dynamic and static simultaneously - it's not urgent.

Testing a release is more urgent

fruch (approved):
LGTM

PR #5638, opened 2023-01-03
refactor(tester): add support for truncate timeout

KnifeyMoloko:
Truncate queries can now have an optional USING TIMEOUT parameter. This commit adds the timeout as an optional argument to the truncate_cf() method in our Tester class.

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -37,6 +37,7 @@ from invoke.exceptions import UnexpectedExit, Failure
 
 from cassandra.concurrent import execute_concurrent_with_args  # pylint: disable=no-name-in-module
 from cassandra import ConsistencyLevel
+from cassandra.cluster import Session  # pylint: disable=no-name-in-module
 
 from argus.db.db_types import TestStatus, PackageVersion
 from sdcm import nemesis, cluster_docker, cluster_k8s, cluster_baremetal, db_stats, wait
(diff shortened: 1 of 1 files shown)


fruch:
This on it's own seems harmless

But the code that is going to use it, needs to be aware of the version it's supported in (both OSS and enterprise), especially if it's going to be using as part of upgrade sequence, where we use older version by design.

KnifeyMoloko:
> This on it's own seems harmless
> 
> But the code that is going to use it, needs to be aware of the version it's supported in (both OSS and enterprise), especially if it's going to be using as part of upgrade sequence, where we use older version by design.

I agree, but at the same time I don't think there's significant risk here. At the moment we're not using this method in rolling-upgrades (we run a bunch of pre-written queries as validation there, but not using this method) and if a future refactor includes it and runs with an older version, it will be an easy debug, since it's a CQL SyntaxError. 
This commit is more so that in the future anyone wanting to use truncate can be made aware of the timeout option via code.

fgelcer (approved):
LGTM

@KnifeyMoloko , did you try changing this on some calls to use this timeout and see what we get?

fruch:
@fgelcer @KnifeyMoloko 

If there no immediate work going to be done ontop of it, from my POV it can way.

I rather no merge unused code (we have plenty of that already)

once someone is gonna need it, he could do the changes needed for his work. (or take this change as is)

roydahan:
What I don't understand is what timeout do we use for it.
Also, What I want to see here is that a test that usually reproduce issue 6309 (Truncate timeout) doesn't reproduce it anymore when using large enough timeout.
And then, set large enough timeout as the default and hopefully even close 6309.

roydahan (approved):

fgelcer:
> What I don't understand is what timeout do we use for it. Also, What I want to see here is that a test that usually reproduce issue 6309 (Truncate timeout) doesn't reproduce it anymore when using large enough timeout. And then, set large enough timeout as the default and hopefully even close 6309.

should we create a task from this comment?

PR #5643, opened 2023-01-04
fix(k8s-eks-pv-setup): find right local store device name

fruch:
code was assuming all device name staring with nvme are local store and on some instances (i3en family) also the root device can be nvme. insted we should be asking the instance metadate api for the correct information

### DIFF
--- a/sdcm/k8s_configs/eks/scylla-node-prepare.yaml
+++ b/sdcm/k8s_configs/eks/scylla-node-prepare.yaml
@@ -233,9 +233,13 @@ spec:
             kubectl config set-context scylla --cluster=scylla --user=qa@scylladb.com
             kubectl config use-context scylla
 
+            # TODO: only for debugging, remove later
+            DEVICES=`curl -s http://169.254.169.254/latest/meta-data/block-device-mapping`
+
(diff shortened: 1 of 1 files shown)


sdcm/k8s_configs/eks/scylla-node-prepare.yaml
```diff
@@ -233,9 +233,13 @@ spec:
             kubectl config set-context scylla --cluster=scylla --user=qa@scylladb.com
             kubectl config use-context scylla
 
+            # TODO: only for debugging, remove later
+            DEVICES=`curl -s http://169.254.169.254/latest/meta-data/block-device-mapping`
+
+            EPHEMERAL_DEV_NAME=`curl -s http://169.254.169.254/latest/meta-data/block-device-mapping/ephemeral0`
```
soyacz, line 239: you can take also this value from `lsblk` if that helps you

fruch:
@soyacz got any progress with this one ? seem like also `i4i.4xlarge` is affected, which is very weird since we used it for the 14 tenant use case...

soyacz:
> @soyacz got any progress with this one ? seem like also `i4i.4xlarge` is affected, which is very weird since we used it for the 14 tenant use case...

I've forgot about it. I'll take a look today

vponomaryov:
@fruch , @soyacz 
https://github.com/scylladb/scylla-cluster-tests/pull/5666

Issue #5641, opened 2023-01-04
test_deploy_helm_with_default_values is failing on EKS, since we can't use EBS


fruch:
Since we move to K8S v1.24, we don't have EBS working, and functional test `test_deploy_helm_with_default_values` is
failing as waiting for pods to be up, but they can't since the default configuration is trying to use EBS:

```
>           raise UnexpectedExit(result)
E           invoke.exceptions.UnexpectedExit: Encountered a bad command exit code!
E
E           Command: 'kubectl --cache-dir=/home/ubuntu/sct-results/20230104-111022-805064/.kube/http-cache --namespace=t-default-values wait --timeout=1m --all --for=condition=Ready pod'
E
E           Exit code: 1
E
E           Stdout:
E
E
E
E           Stderr:
E
E           error: timed out waiting for the condition on pods/t-default-values-scylla-us-east-1-us-east-1a-0
/usr/local/lib/python3.10/site-packages/invoke/runners.py:509: UnexpectedExit
```

fruch:
@vponomaryov what do you think, should we skip until then ? or maybe we could alter the configuration to use local disk as the rest of the functional test ?

fruch:
> > @vponomaryov what do you think, should we skip until then ? or maybe we could alter the configuration to use local disk as the rest of the functional test ?
> 
> We must have some working default storage class. We will be able to use it in this test and minio too.

I'm not sure I understood your answer, current  situation is that the default storage isn't working for us in EKS (when using 1.23+), that's a given (under we'll figure how to address it), question is now, show we disable or bypass it to avoid confusion.

vponomaryov:
> > > @vponomaryov what do you think, should we skip until then ? or maybe we could alter the configuration to use local disk as the rest of the functional test ?
> > 
> > 
> > We must have some working default storage class. We will be able to use it in this test and minio too.
> 
> I'm not sure I understood your answer, current situation is that the default storage isn't working for us in EKS (when using 1.23+), that's a given (under we'll figure how to address it), question is now, show we disable or bypass it to avoid confusion.

As fast solution I am ok to skip the test, but we should make the default storage class exist and work.

PR #5644, opened 2023-01-04
test(k8s-functional): skip `deploy_helm_with_default_values` for EKS

fruch:
Since we moved in EKS to k8s version 1.24, the default EBS based storage isn't working should be resolved in https://github.com/scylladb/qa-tasks/issues/952

meanwhile we'll skip this test for EKS

Fixes: #5641

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -3,8 +3,8 @@
 // trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
 def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 
-def target_backends = ['aws', 'gce', 'docker', 'k8s-local-kind-aws', 'azure']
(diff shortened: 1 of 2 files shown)


fruch:
> looks lical-kind-aws test failed due:
> 
> PodDisruptionBudget - Kubernetes API introduced this resource in the initial release of version 1.0. The API remained fairly stable until version 1.21, when it was graduated and promoted to policy/v1 from policy/v1beta1. policy/v1beta1 is currently deprecated, and will be removed in version 1.25.
> 
> we switched to 1.25 recently didn't we?

we switched, but we still defaults to operator 1.7, which won't work on k8s 1.24+
we can run those `provision` test with latest operator (it can be done manually from jenkins rebuild), 
once 1.8 would be release, we'll change the default to 1.8.

vponomaryov (approved):
LGTM

PR #5649, opened 2023-01-05
fix(removenode_nemesis): Rerun removenode if operation rejected

aleksbykov:
if removing node is alive or marked as alive in gossiper, removenode
operation have to be rejected

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2911,7 +2911,8 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
         1. Terminate node
         2. Run full repair
-        3. Nodetool removenode
+        3. Nodetool removenode, if removenode rejected, because removing node is UN in gossiper,
+           repeat operation in 5 second
         4. Add new node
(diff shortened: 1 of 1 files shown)


sdcm/nemesis.py
```diff
@@ -2947,12 +2948,17 @@ def disrupt_remove_node_then_add_node(self):  # pylint: disable=too-many-branche
             # terminate node
             self._terminate_cluster_node(node_to_remove)
 
+        @retrying(n=3, sleep_time=5, message="Removing node from cluster...")
```
fgelcer, line 2951: maybe 15 seconds won't be enough in tests where the whole cluster + network will be very busy?

fgelcer: also, shouldn't you here be catching the specific exception you are raising in case of failure below?

roydahan (approved):

sdcm/nemesis.py
aleksbykov: The run_nodetool don't raise specific exception, So i left generic exception also

Issue #5646, opened 2023-01-05
[GKE] dynamic loader is failing after ~10 hours


fruch:
## Issue description

loader pods are getting stopped after a long run duration
```
< t:2023-01-05 00:49:24,614 f:kubernetes_cmd_runner.py l:405  c:sdcm.remote.kubernetes_cmd_runner p:WARNING > 'process_is_finished': stopping 'sct-loaders-0-pod-2' pod because stream to it cannot be established having alive pod.
```

seems like the `POD_COUNTER_TO_LIVE = 300` limit is too low when we are running such a long test

@vponomaryov  until we could figure why or logs api are getting stopped that often in GKE, maybe we should raise this number higher ? maybe based on the test duration ? maybe only for GKE ? 

## Installation details

Kernel Version: 5.15.0-1020-gke
Scylla version (or git commit hash): `2022.1.3-20220922.539a55e35` with build-id `d1fb2faafd95058a04aad30b675ff7d2b930278d`
Relocatable Package: http://downloads.scylladb.com/unstable/scylla-enterprise/enterprise-2022.1/relocatable/2022-09-22T13:36:03Z/scylla-enterprise-x86_64-package.tar.gz
Operator Image: scylladb/scylla-operator:1.8.0-rc.0
Operator Helm Version: 1.8.0-rc.0
Operator Helm Repository: https://storage.googleapis.com/scylla-operator-charts/latest
Cluster size: 4 nodes (n1-highmem-16)

Scylla Nodes used in this run:
**No resources left at the end of the run**

OS / Image: `N/A` (k8s-gke: us-east1)

Test: `longevity-scylla-operator-basic-12h-gke`
Test id: `7a41565f-b96a-45f4-b0be-6aa3191808fd`
Test name: `scylla-operator/operator-1.8/gke/longevity-scylla-operator-basic-12h-gke`
Test config file(s):

- [longevity-scylla-operator-basic-12h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/6d14a191cb6d7d355d821b7178e8049138827dc5/test-cases/scylla-operator/longevity-scylla-operator-basic-12h.yaml)

<details>
<summary>
Logs and commands
</summary>

- Restore Monitor Stack command: `$ hydra investigate show-monitor 7a41565f-b96a-45f4-b0be-6aa3191808fd`
[...]

PR #5651, opened 2023-01-06
feature(k8s): run dynamic loader pods in parallel for the static ones

vponomaryov:
After removal of the docker dependency for the static loader run type in K8S we are not able anymore to run utility loads of other types. So, add such possibility by detecting different load type in the static loader runner and reuse dynamic one in such cases.

### DIFF
--- a/sdcm/remote/kubernetes_cmd_runner.py
+++ b/sdcm/remote/kubernetes_cmd_runner.py
@@ -28,6 +28,7 @@ from urllib3.exceptions import (
 )
 
 from sdcm.cluster import TestConfig
+from sdcm import sct_abs_path
 from sdcm.utils.k8s import KubernetesOps
 from sdcm.utils.common import (
     deprecation,
@@ -40,6 +41,10 @@ from .base import RetryableNetworkException
 from .remote_base import RemoteCmdRunnerBase, StreamWatcher
 
 LOGGER = logging.getLogger(__name__)
+
+
+def is_scylla_bench_command(command):
+    return all((str_part in command for str_part in ("scylla-bench", " -workload=", " -mode=")))
 
 
 class KubernetesRunner(Runner):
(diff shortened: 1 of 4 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-eks-add-remove-rack/14/consoleFull
There were enabled following stuff:
- static loader run type
- main load `cassandra-stress` 
- `TruncateLargeParititionMonkey` nemesis that uses `scylla-bench`

Result: Passed
For fast overview search for this line: `command in a separate dynamic loader pod`

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -42,6 +43,10 @@
 LOGGER = logging.getLogger(__name__)
 
 
+def is_scylla_bench_command(command):
```
fruch, line 46: the logic should be backward
we should identify if it's cassnadra-stress of not.

anyhow we should write code that maps between a command and it's stress tool name, and configuration for docker image

we already have `DOCKER_IMAGE_PARAM_NAME`, we should at least use those.

I don't like that we are limiting this to support only scylla-bench

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -120,15 +125,44 @@ class KubernetesCmdRunner(RemoteCmdRunnerBase):
...
+                "K8S_LOADER_NAME": self.pod_name,
+                # TODO: place some values for below env vars when those become required
+                #       in the dynamic loader config.
+                # "POD_CPU_LIMIT": "foo",
+                # "POD_MEMORY_LIMIT": "bar",
+            },
+        )
+
+    def run(self, cmd, **kwargs):  # pylint: disable=arguments-differ
+        if is_scylla_bench_command(cmd) and hasattr(self, "dynamic_remoter") and hasattr(
```
soyacz, line 158: Some loader specific (or even more detailed: scylla-bench) code is leaking to `KubernetesCmdRunner` class which is quite generic.
We should avoid patching generic classes with low level details. Better make `BasePodContainer` code to create proper remoter class instead patching here.

vponomaryov: I disagree. The thing you are proposing to implement requires lots of changes in places where we call this code.
The solution proposed here is one-place change, it is small and solves the problem.

sdcm/remote/kubernetes_cmd_runner.py
vponomaryov: At first, we have been supporting only `scylla-bench` without this PR. I haven't change the support matrix.
At second, I disagree that we `should` do anything more than really needed here, because it makes everything work as it was without `docker`.
I am ok to refactor stuff you propose only in a separate change when we are not blocked right now.

fruch: Which use case it is blocking directly?

vponomaryov: https://github.com/scylladb/scylla-cluster-tests/issues/5646

sdcm/remote/kubernetes_cmd_runner.py
soyacz: I hoping it does not require lot's of changes - `KubernetesCmdRunner` changed required input parameters and there's only one place use was updated. 
If it's not like I think, I understand that in this case it requires more refactor and you decided to go simple way.

vponomaryov: It is not about initialization. It is about calling the `node.remoter.run(...)` method.
After implementation of your proposal, it will be different method, for example `node.remoter.run_dynamic(...)`.
And, it may be called `interface change`, and, obviously, it requires changes in lots of places.

vponomaryov:
@fruch I removed changing of the default value as we agreed on the call

fruch (approved):
LGTM

Issue #5566, opened 2022-12-14
GKE clusters are not getting cleaned by clean-resources


fruch:
5/8 of last weekly runs of GKE job, didn't cleaned up their cluster

https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/gke/job/longevity-scylla-operator-3h-gke/

```
20:47:45  Post behavior destroy for k8s_cluster. Clean resources
20:47:48  There are no instances to remove in AWS region eu-west-1

...

20:48:02  There are no ELBs to remove in AWS region us-west-2
20:48:05  There are no clusters to remove in GKE
20:48:09  Found following orphaned GKE disks: {}
20:48:25  There are no clusters to remove in EKS
20:48:30  There are no instances to remove in GCE
20:48:38  There are no instances to remove in GCE
20:48:49  ssh-agent started already:
20:48:49  		SSH_AUTH_SOCK=/tmp/ssh-C5zgjPV9rZSV/agent.88
20:48:49  		SSH_AGENT_PID=89
20:49:00  There are no resources to clean in Docker
20:49:00  Resources for {'TestId': 'ae47e6db-28d8-4d7a-b3f9-ce858e2c8d08'} have cleaned
20:49:00  Cleaning SSH agent
20:49:00  Agent pid 1070671 killed
20:49:00  + echo 'Finished cleaning resources.'
20:49:00  Finished cleaning resources.
```

manually doing the cleanup (from my laptop) base on the test_id worked

until we figure this out, I've disabled the weekly job

vponomaryov:
Found the root cause, will work on the solution.

fruch:
Damn, I've commented on the wrong issue

https://github.com/scylladb/scylla-cluster-tests/issues/5516#issuecomment-1374540448

I think we need to backport os.env context manager from python 3.11

fruch:
I've confused it with chdir context manager the was introduced in 3.11


We can use something like this:

https://gist.github.com/igniteflow/7267431?permalink_comment_id=4314590#gistcomment-4314590

vponomaryov:
2 solutions available as a fix for it:
Israel's: https://github.com/scylladb/scylla-cluster-tests/pull/5657
My: https://github.com/scylladb/scylla-cluster-tests/pull/5661

PR #5657, opened 2023-01-08
fix(k8s-gke): fix cleanup to not leave env variaible behind

fruch:
leave this enviremnt variables behind, cause us to miss cleaning up some of the GKE cluster, since the code was looking for them in the wrong project

Fixes: #5566

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -94,6 +94,7 @@ from sdcm.send_email import get_running_instances_for_email_report, read_email_d
 from sdcm.parallel_timeline_report.generate_pt_report import ParallelTimelinesReportGenerator
 from sdcm.utils.aws_utils import AwsArchType
 from sdcm.utils.gce_utils import SUPPORTED_PROJECTS
+from sdcm.utils.context_managers import environment
 from utils.build_system.create_test_release_jobs import JenkinsPipelines  # pylint: disable=no-name-in-module,import-error
 from utils.get_supported_scylla_base_versions import UpgradeBaseVersion  # pylint: disable=no-name-in-module,import-error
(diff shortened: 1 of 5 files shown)


soyacz (approved):
Best would be not to use env variables at all in this places (but this would require some refactor of other functions like getting credentials) - but it is out of scope here. We should avoid using env variables in low level functions.

vponomaryov:
Following code change is enough to fix the bug:

```
diff --git a/sdcm/utils/common.py b/sdcm/utils/common.py
index 0b3db197..09d82af4 100644
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -546,8 +546,10 @@ def clean_cloud_resources(tags_dict, dry_run=False):
     clean_elastic_ips_aws(tags_dict, dry_run=dry_run)
     clean_test_security_groups(tags_dict, dry_run=dry_run)
     clean_load_balancers_aws(tags_dict, dry_run=dry_run)
-    clean_clusters_gke(tags_dict, dry_run=dry_run)
-    clean_orphaned_gke_disks(dry_run=dry_run)
+    for project in SUPPORTED_PROJECTS:
+        os.environ['SCT_GCE_PROJECT'] = project
+        clean_clusters_gke(tags_dict, dry_run=dry_run)
+        clean_orphaned_gke_disks(dry_run=dry_run)
     clean_clusters_eks(tags_dict, dry_run=dry_run)
     for project in SUPPORTED_PROJECTS:
         os.environ['SCT_GCE_PROJECT'] = project
```

The proper solution would be to refactor the cleanup completely to not even do redundant API calls which we do a lot in the clean up process.

PR #5658, opened 2023-01-08
feature(k8s-eks): install/configure aws-ebs-csi-driver

fruch:
Since we move to k8s v1.24+ we need to have the aws-ebs-csi-driver
installed otherwise the default gp2 storage class won't be working
and we need it for minio

* install the `aws-ebs-csi-driver` addon
* create and configure a role to it with eksctl
* add the gp3 storage class option in case we'll need to use,
  it also has example of using tags for the volumes

## Testing
- [x] functional test - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/functional-eks/7/
- [x] longevity - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-scylla-operator-3h-eks/26/


sdcm/cluster_k8s/eks.py
```diff
@@ -334,6 +343,25 @@ def deploy(self):
         self.create_eks_cluster()
         LOGGER.info("Patch kubectl config")
         self.patch_kubectl_config()
+        LOGGER.info("Create storage class")
+        self.create_ebs_storge_class()
+        LOGGER.info("Create role for ebs-csi-driver")
+        self.create_role_for_ebs_csi_driver()
+
+    def create_ebs_storge_class(self):
+        if self.params.get('k8s_scylla_disk_class') in SUPPORTED_EBS_STORAGE_CLASSES:
```
vponomaryov, line 352: Need also to have `else` clause where we write warning message that we didn't create any storage class.
But, maybe, we want always to create a default one? Any, even if `gp2`?

fruch: gp2 is the default one and we don't need to create it. this is not going to change according to EKS docs.

anyhow the title of this PR is a bit misleading, since what we want in this PR is to make the default gp2 working again,
gp3 is less of an issue (it's just copied it from the branch I was trying to work on gp3 initially

I'm still trying to make the installation of ebs-csi-driver to work as expected...

vponomaryov: RIght, the commit with the `gp3` mentioning is totally measling because it's part with the IAM stuff is needed in general, always and not exactly for the GP3.

sdcm/cluster_k8s/eks.py
```diff
@@ -334,6 +345,41 @@ def deploy(self):
...
+        wait_for(lambda: self.ebs_csi_driver_status == 'ACTIVE', step=60, throw_exc=True, timeout=600,
+                 text='Waiting till aws-ebs-csi-driver become operational')
+        LOCALRUNNER.run(
+            f'eksctl utils associate-iam-oidc-provider --region={self.region_name} --cluster={self.short_cluster_name} --approve')
+        LOCALRUNNER.run(f'eksctl create iamserviceaccount --name ebs-csi-controller-sa --namespace kube-system '
+                        f'--cluster {self.short_cluster_name} --attach-policy-arn arn:aws:iam::aws:policy/service-role/AmazonEBSCSIDriverPolicy '
+                        f'--approve --role-name EKS_EBS-{self.short_cluster_name} --region {self.region_name} '
+                        f'--tags {tags} --override-existing-serviceaccounts')
+
+        self.kubectl("rollout restart deployment ebs-csi-controller", namespace="kube-system")
```
vponomaryov, line 382: Why the restart is triggered here? Shouldn't it just start working after eksctl commands executions?

vponomaryov (changes requested):

sdcm/cluster_k8s/eks.py
```diff
@@ -334,6 +345,41 @@ def deploy(self):
         self.create_eks_cluster()
         LOGGER.info("Patch kubectl config")
         self.patch_kubectl_config()
+        LOGGER.info("Create storage class")
+        self.create_ebs_storge_class()
+
+    def create_ebs_storge_class(self):
```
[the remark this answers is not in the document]
fruch: I'm failing to understand how create a new storage class is to with cpu/memory resources
or you referring to the resources used by ebs-csi-driver ?

sdcm/cluster_k8s/eks.py
fruch: I would have thought so as well, but the `eksctl` command just put the configuration in place, and the pods still failing with permission issue.

AWS instructions doesn't state you need the restart, but in the instruction doing the process without eksctl they do mention this step as mandatory

sdcm/cluster_k8s/eks.py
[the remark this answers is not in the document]
vponomaryov: Yes, `ebs-csi-driver` has resources requests for CPU and memory and for now it's appearance increases the probability that the requests for scylla pods may not be satisfied because they are calculated without regards for requests of the new pods.

sdcm/tester.py
```diff
@@ -1572,7 +1572,8 @@ def get_cluster_k8s_eks(self):  # pylint: disable=too-many-statements
                 disk_size=40,
                 role_arn=self.k8s_cluster.nodegroup_role_arn,
                 k8s_cluster=self.k8s_cluster),
-            wait_till_ready=False)
+            wait_till_ready=True)
```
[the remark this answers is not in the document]
fruch: moved

sdcm/cluster_k8s/eks.py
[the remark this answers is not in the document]
fruch: csi driver add 3 pods with such configuration for each node:
```
#     resources:
#       limits:
#         cpu: 100m
#         memory: 256Mi
#       requests:
#         cpu: 10m
#         memory: 40Mi
```

you think we should make the `cpu: 0.51` more cause of this new 30m requests ?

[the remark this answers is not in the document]
fruch: o.k. changed

sdcm/cluster_k8s/eks.py
vponomaryov: Ok

sdcm/utils/common.py
```diff
@@ -992,6 +997,82 @@ def clean_load_balancers_aws(tags_dict, dry_run=False):
...
+
+        tag_mappings = itertools.chain.from_iterable(
+            page['ResourceTagMappingList']
+            for page in paginator.paginate(TagFilters=tag_filter, ResourceTypeFilters=['cloudformation:stack'])
+        )
+        cloudformation_stacks[region] = list(tag_mappings)
+        if verbose:
+            LOGGER.info("%s: done [%s/%s]", region, len(list(cloudformation_stacks.keys())), len(aws_regions))
+
+    ParallelObject(aws_regions, timeout=100).run(get_stacks, ignore_exceptions=False)
```
vponomaryov, line 1031: For the moment we have 16 elements in the `aws_regions` list.
By default, ParallelObject, will run everything with concurrency `6`.
So, I think you may want to provide here additional option -> `num_workers=len(aws_regions)`.

fruch: this is a change we probably want across all of those functions, not related to this specific one.

sdcm/cluster_k8s/__init__.py
```diff
@@ -143,11 +143,19 @@
 }
 
 # Other common resources which get deployed on each scylla node such as 'kube-proxy'
-# EKS: between 100m-200m CPU
+# EKS: between 100m-230m CPU
```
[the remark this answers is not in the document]
fruch: not sure how we get to those numbers... i.e. where the x2 comes from

sdcm/cluster_k8s/__init__.py
```diff
@@ -143,11 +143,19 @@
...
+# EKS: ebs-csi-driver has 3 pods with:
+#     resources:
+#       limits:
+#         cpu: 100m
+#         memory: 256Mi
+#       requests:
+#         cpu: 10m
+#         memory: 40Mi
```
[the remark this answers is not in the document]
fruch: removed

sdcm/cluster_k8s/__init__.py
```diff
@@ -143,11 +143,19 @@
...
 # GKE: between 200m-300m CPU and 250Mi RAM
 # Above numbers are "explicit" reservations. So, reserve a bit more for other common pods.
 COMMON_CONTAINERS_RESOURCES = {
-    'cpu': 0.51,
+    'cpu': 0.5130,
```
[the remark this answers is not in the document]
fruch: changed

sdcm/utils/common.py
fruch: changed it

vponomaryov (approved):
LGTM

PR #5670, opened 2023-01-11
test(rolling-upgrade): set yaml entries in upgade procedure for raft

KnifeyMoloko:
Adds a new flag, `consistent_cluster_management` to the upgraded nodes' scylla.yaml and tidies up the way in which we change scylla.yaml contents in rolling_upgrades. It also removes the legacy `enable_3_1_0_compatibility_mode` flag setting. A new continuous event is added to allow for easier debugging of the scylla.yaml updates.

Staging job with test runs: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/maciej/job/rolling-upgrade-raft/

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -103,7 +103,7 @@ from sdcm.sct_events.base import LogEvent
 from sdcm.sct_events.health import ClusterHealthValidatorEvent
 from sdcm.sct_events.system import TestFrameworkEvent, INSTANCE_STATUS_EVENTS_PATTERNS, InfoEvent, SoftTimeoutEvent
 from sdcm.sct_events.grafana import set_grafana_url
-from sdcm.sct_events.database import SYSTEM_ERROR_EVENTS_PATTERNS, ScyllaHelpErrorEvent
+from sdcm.sct_events.database import SYSTEM_ERROR_EVENTS_PATTERNS, ScyllaHelpErrorEvent, ScyllaYamlUpdateEvent
 from sdcm.sct_events.nodetool import NodetoolEvent
 from sdcm.sct_events.decorators import raise_event_on_failure
 from sdcm.utils.auto_ssh import AutoSshContainerMixin
@@ -1540,7 +1540,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             yield new_scylla_yaml
             diff = old_scylla_yaml.diff(new_scylla_yaml)
             if not diff:
-                LOGGER.debug("%s: scylla.yaml hasn't been changed", self)
+                ScyllaYamlUpdateEvent(node_name=self.name,
+                                      message=f"ScyllaYaml has not been changed on node: {self.name}")
                 return
             scylla_yaml.clear()
             scylla_yaml.update(
@@ -1551,7 +1552,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
                     explicit=['partitioner', 'commitlog_sync', 'commitlog_sync_period_in_ms', 'endpoint_snitch']
                 )
             )
-            LOGGER.debug("%s: scylla.yaml will be updated to:\n%s", self, scylla_yaml)
+            ScyllaYamlUpdateEvent(node_name=self.name, message=f"ScyllaYaml has been changed on node: {self.name}. "
+                                                               f"Diff: {diff}")
 
     def remote_manager_yaml(self):
         return self._remote_yaml(path=SCYLLA_MANAGER_YAML_PATH)
--- a/sdcm/provision/scylla_yaml/scylla_yaml.py
+++ b/sdcm/provision/scylla_yaml/scylla_yaml.py
@@ -58,6 +58,7 @@ class ScyllaYaml(BaseModel):  # pylint: disable=too-few-public-methods
     rpc_interface: str = "eth1"
     rpc_interface_prefer_ipv6: bool = False
     seed_provider: List[SeedProvider] = [SeedProvider(class_name='org.apache.cassandra.locator.SimpleSeedProvider')]
+    consistent_cluster_management: bool = False
     compaction_throughput_mb_per_sec: int = 0
     compaction_large_partition_warning_threshold_mb: int = 1000
     compaction_large_row_warning_threshold_mb: int = 10
--- a/sdcm/sct_events/database.py
+++ b/sdcm/sct_events/database.py
@@ -381,6 +381,13 @@ class ScyllaSysconfigSetupEvent(ScyllaDatabaseContinuousEvent):
         super().__init__(node=node, severity=severity, publish_event=True)
 
 
+class ScyllaYamlUpdateEvent(InformationalEvent):
+    def __init__(self, node_name: str, message: Optional[str] = None, diff: dict | None = None,
+                 severity=Severity.NORMAL, **__):
+        super().__init__(severity=severity)
+        self.message = message or f"Updating scylla.yaml contents on node: {node_name}. Diff: {diff}"
+
+
 SCYLLA_DATABASE_CONTINUOUS_EVENTS = [
     ScyllaServerStatusEvent,
     BootstrapEvent,
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -159,6 +159,12 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
         new_scylla_repo = self.params.get('new_scylla_repo')
         new_version = self.params.get('new_version')
         upgrade_node_packages = self.params.get('upgrade_node_packages')
+        scylla_yaml_updates = {
+            "consistent_cluster_management": True
+        }
+
+        if self.params.get('test_sst3'):
+            scylla_yaml_updates.update({"enable_sstables_mc_format": True})
 
         InfoEvent(message='Upgrading a Node').publish()
         node.upgrade_system()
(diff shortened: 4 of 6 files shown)


sdcm/provision/scylla_yaml/scylla_yaml.py
```diff
@@ -58,6 +58,7 @@ def set_endpoint_snitch(cls, endpoint_snitch: str):
     rpc_interface: str = "eth1"
     rpc_interface_prefer_ipv6: bool = False
     seed_provider: List[SeedProvider] = [SeedProvider(class_name='org.apache.cassandra.locator.SimpleSeedProvider')]
+    consistent_cluster_management: bool = False
```
fgelcer, line 61: for now, is it the default in `scylla.yaml`?

upgrade_test.py
```diff
@@ -236,14 +236,11 @@ def upgrade_node(self, node, upgrade_sstables=True):
...
                         r'-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" '.format(scylla_pkg))
-        if self.params.get('test_sst3'):
-            node.remoter.run("echo 'enable_sstables_mc_format: true' |sudo tee --append /etc/scylla/scylla.yaml")
-        if self.params.get('test_upgrade_from_installed_3_1_0'):
-            node.remoter.run("echo 'enable_3_1_0_compatibility_mode: true' |sudo tee --append /etc/scylla/scylla.yaml")
         check_reload_systemd_config(node)
         # Current default 300s aren't enough for upgrade test of Debian 9.
         # Related issue: https://github.com/scylladb/scylla-cluster-tests/issues/1726
         node.run_scylla_sysconfig_setup()
+        self._update_scylla_yaml_on_node(node)
```
fgelcer, line 243: `_update_scylla_yaml_on_node` is not a good name, because it is too general, and doesn't show that it is something very specific...
true that there is another (conditional) change too, but at least we need some documentation explaining what is being updated IMHO

fgelcer (commented):
1 small question, and a little concern... but LGTM other than that

upgrade_test.py
fgelcer: also, should you not revert these changes when you rollback?

KnifeyMoloko: Agreed, it's not the best name. I changed it around again: left the generic name and added the updates dict as a parameter to it. That way we can build the updates dict in the `upgrade_node` method and just pass them to this more generic method.

sdcm/provision/scylla_yaml/scylla_yaml.py
KnifeyMoloko: It's the default in our `ScyllaYaml` class, since it would refuse unrecognized options being added.

fruch: ScyllaYaml doesn't support seeing value to None ?

i.e. it shouldn't appear by default

Since the default in scylla might change in following versions, I don't want SCT to mask that

fgelcer (approved):
LGTM

upgrade_test.py
```diff
@@ -159,6 +159,12 @@ def upgrade_node(self, node, upgrade_sstables=True):
         new_scylla_repo = self.params.get('new_scylla_repo')
         new_version = self.params.get('new_version')
         upgrade_node_packages = self.params.get('upgrade_node_packages')
+        scylla_yaml_updates = {
+            "consistent_cluster_management": "true"
```
fruch, line 163: should this be 
```suggestion
            "consistent_cluster_management": True

File: upgrade_test.py [hunk 4]
```diff
@@ -236,14 +242,11 @@ def upgrade_node(self, node, upgrade_sstables=True):
...
                         r'-o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold" '.format(scylla_pkg))
-        if self.params.get('test_sst3'):
-            node.remoter.run("echo 'enable_sstables_mc_format: true' |sudo tee --append /etc/scylla/scylla.yaml")
-        if self.params.get('test_upgrade_from_installed_3_1_0'):
-            node.remoter.run("echo 'enable_3_1_0_compatibility_mode: true' |sudo tee --append /etc/scylla/scylla.yaml")
         check_reload_systemd_config(node)
         # Current default 300s aren't enough for upgrade test of Debian 9.
         # Related issue: https://github.com/scylladb/scylla-cluster-tests/issues/1726
         node.run_scylla_sysconfig_setup()
+        self._update_scylla_yaml_on_node(node_to_update=node, updates=scylla_yaml_updates)
```
fruch, line 249: why here and not where `test_sst3` was before ?

otherwise the call there for `check_reload_systemd_config` seems kind of pointless

upgrade_test.py
```diff
@@ -501,6 +504,11 @@ def shuffle_nodes_and_alternate_dcs(nodes: List[BaseNode]) -> List[BaseNode]:
 
         return [x for x in chain.from_iterable(zip_longest(*dc_nodes.values())) if x]
 
+    @staticmethod
+    def _update_scylla_yaml_on_node(node_to_update: BaseNode, updates: dict):
+        with ScyllaYamlUpdateEvent(node=node_to_update.name), node_to_update.remote_scylla_yaml() as scylla_yaml:
```
fruch, line 509: I can understand why we'll want an event
but i'm not sure i'm following why it's need to be `ScyllaDatabaseContinuousEvent`, i.e. do we cat about it's start and end time ?

sdcm/sct_events/database.py
```diff
@@ -381,6 +381,12 @@ def __init__(self, node: str, severity=Severity.NORMAL, **__):
         super().__init__(node=node, severity=severity, publish_event=True)
 
 
+class ScyllaYamlUpdateEvent(ScyllaDatabaseContinuousEvent):
+    def __init__(self, node: str, message: Optional[str] = None, severity=Severity.NORMAL, **__):
+        super().__init__(node=node, severity=severity)
+        self.message = message or f"Updating scylla.yaml contents on node: {node}"
```
fruch, line 387: I'm guessing if we could also add the actual change it might be a big more useful

upgrade_test.py
KnifeyMoloko: I think this only matters if we touch the unit definitions for `systemd`. `scylla.yaml` is read when the service is restarted, so I put the update method before the restart.

upgrade_test.py
KnifeyMoloko: Good catch, I think it should.

sdcm/sct_events/database.py
KnifeyMoloko: Agreed, adding the content of the updates to the message.

upgrade_test.py
```diff
@@ -501,6 +504,11 @@ def shuffle_nodes_and_alternate_dcs(nodes: List[BaseNode]) -> List[BaseNode]:
 
         return [x for x in chain.from_iterable(zip_longest(*dc_nodes.values())) if x]
 
+    @staticmethod
+    def _update_scylla_yaml_on_node(node_to_update: BaseNode, updates: dict):
+        with ScyllaYamlUpdateEvent(node=node_to_update.name, update=updates), node_to_update.remote_scylla_yaml() as scylla_yaml:
```
soyacz, line 509: maybe add this to `remote_scylla_yaml()` to always send event on yaml update?
This is quite important action, so would be worth doing it always?
there's a diff extracted in `sdcm.cluster_k8s.KubernetesCluster.manage_file_in_scylla_config_map` and `sdcm.cluster.BaseNode.remote_scylla_yaml` (maybe some other place would need update, need to check) - so this diff could be published as event.
@KnifeyMoloko @fruch @fgelcer what do you think?

upgrade_test.py
```diff
@@ -159,6 +159,12 @@ def upgrade_node(self, node, upgrade_sstables=True):
         new_scylla_repo = self.params.get('new_scylla_repo')
         new_version = self.params.get('new_version')
         upgrade_node_packages = self.params.get('upgrade_node_packages')
+        scylla_yaml_updates = {
```
soyacz, line 162: when doing rollbacks, shouldn't we recover settings?

KnifeyMoloko: We kinda do, at least on reinstalls: https://github.com/scylladb/scylla-cluster-tests/blob/1d3514749a4dd9679a291c02351ba76bfbbbf7bf/upgrade_test.py#L229

That being said, we don't reuse `scylla.yaml` as it was initially, but append feature flags to it when necessary. I already removed one update for the `enable_3_1_0_compatibility_mode` flag with this PR, but I don't see us avoiding these changes to `scylla.yaml`.

upgrade_test.py
KnifeyMoloko: Seems reasonable to me.

fgelcer:
@KnifeyMoloko , were the last change requested by @soyacz implemented?

fgelcer (dismissed):
LGTM

sdcm/sct_events/database.py
```diff
@@ -381,6 +381,13 @@ def __init__(self, node: str, severity=Severity.NORMAL, **__):
         super().__init__(node=node, severity=severity, publish_event=True)
 
 
+class ScyllaYamlUpdateEvent(ScyllaDatabaseContinuousEvent):
```
soyacz, line 384: I don't consider it as continuous event - changing yaml file is quite atomic.

sdcm/cluster.py
```diff
@@ -1539,19 +1539,19 @@ def remote_scylla_yaml(self) -> ContextManager[ScyllaYaml]:
...
-            scylla_yaml.update(
-                new_scylla_yaml.dict(
-                    exclude_none=True, exclude_unset=True, exclude_defaults=True,
-                    # NOTE: explicit fields included into yaml no matter what,
-                    #  they are needed for nodetool to operate properly
-                    explicit=['partitioner', 'commitlog_sync', 'commitlog_sync_period_in_ms', 'endpoint_snitch']
+            with ScyllaYamlUpdateEvent(node=self.name, diff=diff):
+                if not diff:
+                    LOGGER.debug("%s: scylla.yaml hasn't been changed", self)
+                    return
```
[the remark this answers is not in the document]
KnifeyMoloko: We should still get some feedback if the change did not go through (i.e. if the diff was empty). I reworked it a bit to have an `InformationEvent` be created on either success or failure, with a different message.

sdcm/sct_events/database.py
KnifeyMoloko: So we have 2 votes to 1 on the Info vs. Continuous event debate :) I concede and moved the event to being a `InformationalEvent`

fgelcer (approved):
LGTM

soyacz (approved):
One small opinion, otherwise LGTM.

fgelcer:
ping @fruch

upgrade_test.py
fruch: Since those flags weren't used for a while, I guess it doesn't make much difference.

But we'll see once all the upgrade tests would run

fruch (approved):

fgelcer:
@roydahan , should it be backported to 5.2, no?

roydahan:
1. I'm not sure since I didn't review this PR.
2. IIUC it adds the flag "consistent_cluster_management: True" which means "enable raft" to all our upgrade tests, is that right?
3. If so, it means that this PR isn't good enough because now we don't cover the upgrade path that doesn't enable this flag.
4. When we test upgrade from 5.1 we have the experimental flag as "true", does it mean that raft is enabled there or we need to enable it explicitly?
5. I guess that you cleared it with product or anyone that we don't test upgrade from "--raft-experimental" to this flag.

So, if we have the answers to all of the above, I can answer your question.

PR #5672, opened 2023-01-11
Revert "improvement(k8s): remove redundant explicit Scylla IP mapping refreshes"

vponomaryov:
This reverts commit bf001fc97afc03aa6b0aaba7fe6d7575c7f5b436.

The reason is that GKE cluster may work about hour after pod IPs change not providing any update to the 'endpoint' resources which we listen to. So, return back the manual IP refreshes due to the GKE's slowliness.

### DIFF
--- a/functional_tests/scylla_operator/libs/helpers.py
+++ b/functional_tests/scylla_operator/libs/helpers.py
@@ -196,37 +196,42 @@ def verify_resharding_on_k8s(db_cluster: ScyllaPodCluster, cpus: Union[str, int,
     # Wait for the start of the resharding.
     # In K8S it starts from the last node of a rack and then goes to previous ones.
     # One resharding with 100Gb+ may take about 3-4 minutes. So, set 5 minutes timeout per node.
-    for node, liveness_probe_failures, resharding_start, resharding_finish in nodes_data:
-        assert wait_for(
-            func=lambda: list(resharding_start),  # pylint: disable=cell-var-from-loop
-            step=1, timeout=600, throw_exc=False,
-            text=f"Waiting for the start of resharding on the '{node.name}' node.",
-        ), f"Start of resharding hasn't been detected on the '{node.name}' node."
-        resharding_started = time.time()
-        log.debug("Resharding has been started on the '%s' node.", node.name)
+    try:
+        for node, liveness_probe_failures, resharding_start, resharding_finish in nodes_data:
+            assert wait_for(
+                func=lambda: list(resharding_start),  # pylint: disable=cell-var-from-loop
+                step=1, timeout=600, throw_exc=False,
+                text=f"Waiting for the start of resharding on the '{node.name}' node.",
+            ), f"Start of resharding hasn't been detected on the '{node.name}' node."
+            resharding_started = time.time()
+            log.debug("Resharding has been started on the '%s' node.", node.name)
 
-        # Wait for the end of resharding
-        assert wait_for(
-            func=lambda: list(resharding_finish),  # pylint: disable=cell-var-from-loop
-            step=3, timeout=600, throw_exc=False,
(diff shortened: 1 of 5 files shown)


fruch:
1h and the resource doesn't update, doesn't sound like slowness in my playbook, more of something functional not working, or not working as we expect.

Do we have CNI driver in GKE as in EKS ?

vponomaryov:
> 1h and the resource doesn't update, doesn't sound like slowness in my playbook, more of something functional not working, or not working as we expect.

If you refer to the python implementation made in SCT, then I don't think so.
If you refer to the GKE's functionality then it is exactly what I am referring to.

> Do we have CNI driver in GKE as in EKS ?
`CNI` is interface, it is used anyway. 
In our case we use the `default` GKE's implementation for it, not custom one like `calico` or any other.

fruch:
> > 1h and the resource doesn't update, doesn't sound like slowness in my playbook, more of something functional not working, or not working as we expect.
> 
> If you refer to the python implementation made in SCT, then I don't think so. If you refer to the GKE's functionality then it is exactly what I am referring to.

I was referring to GKE, would such a thing be documented/written anywhere ? 

> 
> > Do we have CNI driver in GKE as in EKS ?
> 
> We use the `default` GKE's implementation, not custom one like `calico` or any other.
> 
> In theory, the change of the CNI driver may help.

fruch:
> Testing it here: https://jenkins.scylladb.com/view/scylla-operator/job/scylla-operator/job/operator-1.8/job/upgrade/job/upgrade-platform-k8s-gke/4/consoleFull

this run failed, and only re-run passed, is there related to this or not ?

also is there a place we can open the issue we are seeing in GKE ? with us no seeing ip changes ? if not please open it on SCT with all the details, we will need to chase it down at some time (I think we might again seeing some rate limiting on our api calls, again without digging into the logs)

vponomaryov:
> > Testing it here: https://jenkins.scylladb.com/view/scylla-operator/job/scylla-operator/job/operator-1.8/job/upgrade/job/upgrade-platform-k8s-gke/4/consoleFull
> 
> this run failed, and only re-run passed, is there related to this or not ?

The details were described in the argus. The build#4 failure is unrelated to this change.

> also is there a place we can open the issue we are seeing in GKE ? with us no seeing ip changes ? if not please open it on SCT with all the details, we will need to chase it down at some time (I think we might again seeing some rate limiting on our api calls, again without digging into the logs)

API rate limits must not be the cause for the current problem because our solution is to keep 1 stream connection open and listen for updates.
We do not have such problem in EKS at all. So, obviously, GKE does it's work worse in this case.
For now, I think it should create SCT bug with details we have already that can be used for further investigation.

fruch:
> > > Testing it here: https://jenkins.scylladb.com/view/scylla-operator/job/scylla-operator/job/operator-1.8/job/upgrade/job/upgrade-platform-k8s-gke/4/consoleFull
> > 
> > 
> > this run failed, and only re-run passed, is there related to this or not ?
> 
> The details were described in the argus. The build#4 failure is unrelated to this change.
> 
> > also is there a place we can open the issue we are seeing in GKE ? with us no seeing ip changes ? if not please open it on SCT with all the details, we will need to chase it down at some time (I think we might again seeing some rate limiting on our api calls, again without digging into the logs)
> 
> API rate limits must not be the cause for the current problem because our solution is to keep 1 stream connection open and listen for updates. We do not have such problem in EKS at all. So, obviously, GKE does it's work worse in this case. For now, I think, we should create SCT bug with details we have already that can be used for further investigation.

o.k.let get that opened

fruch (approved):
LGTM

vponomaryov:
> o.k.let get that opened

https://github.com/scylladb/scylla-cluster-tests/issues/5679

PR #5673, opened 2023-01-11
fix(argus_test_run.py): Add random seconds to the run start_time

k0machi:
The reasoning for this fix is that since the primary key for sct reports is (build_id, start_time), two runs running in parallel could potentially start on the same second, as that's the level of precision we're currently using. This should fix missing rolling upgrade runs in the future.

### DIFF
--- a/sdcm/argus_test_run.py
+++ b/sdcm/argus_test_run.py
@@ -14,7 +14,8 @@
 import logging
 import unittest.mock
 from uuid import UUID
-from datetime import datetime
+from datetime import datetime, timedelta
+from random import randint
 
 from argus.db.db_types import TestStatus
 from argus.db.testrun import TestRunWithHeartbeat, TestRunInfo, TestDetails, TestResources, TestLogs, TestResults, \
@@ -235,9 +236,12 @@ class ArgusTestRun:
         config_files = sct_config.get("config_files")
         started_by = get_username()
 
+        # start time is a part of primary key and getting same start time will cause an overwrite
+        safe_start_time = datetime.utcnow().replace(microsecond=0) + timedelta(seconds=(randint(10, 60)))
+
         details = TestDetails(scm_revision_id=get_git_commit_id(), started_by=started_by, build_job_url=job_url,   # pylint: disable=no-value-for-parameter
                               yaml_test_duration=sct_config.get("test_duration"),
-                              start_time=datetime.utcnow().replace(microsecond=0),
+                              start_time=safe_start_time,
                               config_files=config_files, packages=[])
 
         LOGGER.info("Preparing Resource Setup...")


sdcm/argus_test_run.py
```diff
@@ -235,9 +236,12 @@ def from_sct_config(cls, test_id: UUID, sct_config: SCTConfiguration) -> TestRun
         config_files = sct_config.get("config_files")
         started_by = get_username()
 
+        # start time is a part of primary key and getting same start time will cause an overwrite
+        safe_start_time = datetime.utcnow().replace(microsecond=0) + timedelta(seconds=(randint(10, 60)))
```
fruch, line 240: it doesn't sound that safe to me... it's still can get into that situation cause it's random. 

why aren't we fixing this on the server side ? cause we need to change the schema ?

why are we zeroing the microseconds ? leaving them in sound a bit more safer...

fgelcer: i agree that fixing it on the server side would have bigger benefit, but perhaps, because of kind of "low" parallelism in these jobs, it won't be a problem the current situation, until a proper fix is done in the server side...

fgelcer: other way to mitigate it, is to delay the start of parallel jobs inside the same build by, let's say, 1 second, as it won't make much difference in the end result, and will most certainly mitigate this issue

k0machi: > it doesn't sound that safe to me... it's still can get into that situation cause it's random.

I agree, there is still a possibility of getting runs to random the same interval while also starting at the same time, or worse, getting a random interval on one run enough to intersect with start time of another run.

> why aren't we fixing this on the server side ? cause we need to change the schema ?

This will be fixed serverside (and in fact kinda is fixed already since timestamp is calculated serverside and is not submitted in the new client), but since sct is still interacting directly with db this is a temporary fix for that.

> why are we zeroing the microseconds ? leaving them in sound a bit more safer...

It seems that scylla (at least by default) doesn't accept microsecond precision in the timestamps, I'm zeroing them to resolve that, but if there's a way to make scylla to accept it that would be better.

fruch: much better, 

also when do we plan to switch to the new client api ?

soyacz (commented):
why can't we use test_id instead start time?

k0machi:
> why can't we use test_id instead start time?
We used that before, but that lead to issues with filtering and slow performance when compiling dashboard data. Build Id allows for better grouping - every run is grouped to the test by just existing, and timestamp automatically provides a nice sorting mechanism as a clustering key.

Either way, this fix is not needed anymore - I decided to finally migrate SCT to the REST client here: #5682.

Issue #5671, opened 2023-01-11
GKE evicts valueable pods from Scylla K8S node when enospc nemesis is applied


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: any

### Logs
* test_id: d457af3c-14d9-454a-a7ca-02f379e03c70
* job log: https://jenkins.scylladb.com/job/scylla-operator/job/operator-1.8/job/gke/job/longevity-scylla-operator-basic-12h-gke/6/consoleFull

### Description

When we run longevity job on the GKE (K8S) backend and trigger the `disrupt_nodetool_enospc` nemesis then 
we get the following situation change:

1) All our pods get killed on that affected Scylla K8S node:
```
NAMESPACE         NAME                                                             READY   STATUS      RESTARTS      AGE     IP            NODE                                                  NOMINATED NODE   READINESS GATES
...
default           static-local-volume-provisioner-ds-kt4b4                         0/1     Evicted     0             4m57s   <none>        gke-longev-scylla-operat--scylla-pool-ba70b468-89xk   <none>           <none>
...
default           xfs-formatter-g66b5                                              0/2     Evicted     0             4m58s   <none>        gke-longev-scylla-operat--scylla-pool-ba70b468-89xk   <none>           <none>
...
scylla            sct-cluster-us-east1-b-us-east1-2                                0/2     Pending     0             6h7m    <none>        <none>                                                <none>           <none>
```

2) The affected scylla pod becomes unschedulable with the following status:
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: "2023-01-10T22:45:52Z"
    message: '0/8 nodes are available: 1 node(s) had untolerated taint {node.kubernetes.io/disk-pressure:
      }, 1 node(s) had volume node affinity conflict, 2 Insufficient cpu, 2 Insufficient
      memory, 4 node(s) didn''t match Pod''s node affinity/selector. preemption: 0/8
      nodes are available: 2 No preemption victims found for incoming pod, 6 Preemption
      is not helpful for scheduling.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
  phase: Pending
  qosClass: Guaranteed
```

And, the `enospc` effect fails to be reverted and makes the whole test run be broken till it's end:

```
[...]

PR #5676, opened 2023-01-12
fix(k8s-gke): skip 'disrupt_nodetool_enospc' due to bug

vponomaryov:
The `disrupt_nodetool_enospc` nemesis doesn't work on the `GKE` backend due to the following bug: https://github.com/scylladb/scylla-cluster-tests/issues/5671

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1556,6 +1556,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             # Moreover, it will cause host system's /var hosting disk go out of free space too
             # which may cause unpredictable failures.
             raise UnsupportedNemesis('disrupt_nodetool_enospc is not supported on local K8S clusters')
+        if self.cluster.params.get('cluster_backend') == 'k8s-gke':
+            raise UnsupportedNemesis(
+                'disrupt_nodetool_enospc is skipped due to this: '
+                'https://github.com/scylladb/scylla-cluster-tests/issues/5671')
 
         if all_nodes:
             nodes = self.cluster.nodes


vponomaryov:
Tested here: https://jenkins.scylladb.com/job/scylla-operator/job/operator-1.8/job/gke/job/longevity-scylla-operator-basic-12h-gke/7 (see argus for comments)

fruch (approved):
LGTM

fruch:
@vponomaryov I've merged. but I don't like the fact we didn't got yet to the bottom of it.

vponomaryov:
> @vponomaryov I've merged. but I don't like the fact we didn't got yet to the bottom of it.

What exactly you propose to do? Eviction of pods is not a bug, it is design mechanism.
In theory, we can try to reconfigure the guts of GKE for some extent, but it is not what will be common users do.

fruch:
> > @vponomaryov I've merged. but I don't like the fact we didn't got yet to the bottom of it.
> 
> What exactly you propose to do? Eviction of pods is not a bug, it is designed mechanism. In theory, we can try to reconfigure the guts of GKE for some extent, but it is not what common users will do.

user would if it would be part of the instructions

soyacz:
> > @vponomaryov I've merged. but I don't like the fact we didn't got yet to the bottom of it.
> 
> What exactly you propose to do? Eviction of pods is not a bug, it is designed mechanism. In theory, we can try to reconfigure the guts of GKE for some extent, but it is not what common users will do.

Maybe GKE has different defaults and we must tune it to match EKS? If we cannot assign 100% disk (or even 90%) then during STCS compaction users may not get what they expect.

PR #5678, opened 2023-01-12
Update branch 2021.1 to use latest s-b by default

fruch:
we did update branch 2022.1 and 2022.2, and asymmetric test case was working on them
we want that case back and operational also on 2021.1 

## Testing 
- [ ] run the `asymmetric-cluster-3h` on this PR - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-large-partition-asymmetric-cluster-3h-test/17/

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -156,6 +156,6 @@ mgmt_docker_image: ''
 k8s_deploy_monitoring: false
 
 scylla_rsyslog_setup: false
-scylla_bench_version: v0.1.6
+scylla_bench_version: v0.1.14
 
 service_level_shares: [1000]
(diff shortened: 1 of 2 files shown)


fgelcer (commented):
we are branching 2023.1, and if we release it within a month, 2021.1 will be deprecated... i'm not sure we want to do this change in there, other than if we have blockers, but i don't think that is the case.
i suggest closing this PR and work on newer branches

fruch:
> we are branching 2023.1, and if we release it within a month, 2021.1 will be deprecated... i'm not sure we want to do this change in there, other than if we have blockers, but i don't think that is the case. i suggest closing this PR and work on newer branches

we are doing this as for @roydahan request to make this use case working across all the current branches.

one month, we'll see, my bet is at least 3 month

roydahan (approved):

Issue #5677, opened 2023-01-12
keyspace number is reset after first list element of stress_cmd 


soyacz:
in case we set `keyspace_num > 1` and we have multiple `stress_cmd` commands, example test config:
```
round_robin: true
keyspace_num: 50
stress_cmd: ["cassandra-stress write cl=QUORUM duration=240m -schema 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=4 -pop seq=1..20971520 -col 'n=FIXED(2) size=FIXED(256)' -log interval=5",
             "cassandra-stress read  cl=QUORUM duration=240m -schema 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=4 -pop seq=1..20971520 -col 'n=FIXED(2) size=FIXED(256)' -log interval=5"
             ]
```
then, stress commands for read always use `keyspace1`:
```
< t:2023-01-11 15:28:50,682 f:stress_thread.py l:271  c:sdcm.stress_thread   p:INFO  > cassandra-stress write cl=QUORUM duration=240m -schema keyspace=keyspace1 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=4 -pop seq=1..20971520 -col 'n=FIXED(2) size=FIXED(256)' -log interval=5 -node 10.0.3.89,10.0.3.49,10.0.2.93,10.0.2.110 -errors skip-unsupported-columns
< t:2023-01-11 15:29:10,837 f:stress_thread.py l:271  c:sdcm.stress_thread   p:INFO  > cassandra-stress read  cl=QUORUM duration=240m -schema keyspace=keyspace1 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=4 -pop seq=1..20971520 -col 'n=FIXED(2) size=FIXED(256)' -log interval=5 -node 10.0.3.89,10.0.3.49,10.0.2.93,10.0.2.110 -errors skip-unsupported-columns
[...]

PR #5684, opened 2023-01-13
fix(LoaderUtilsMixin): fix bugs and small refactor

soyacz:
Fix for #5677 along with small refactor:
- removed `_parse_stress_cmd` as it wasn't parsing, it was hiding
- removed redundant code

### DIFF
--- a/sdcm/utils/loader_utils.py
+++ b/sdcm/utils/loader_utils.py
@@ -29,19 +29,6 @@ SERVICE_LEVEL_NAME_TEMPLATE = 'sl%s_%s'
 class LoaderUtilsMixin:
     """This mixin can be added to any class that inherits the 'ClusterTester' one"""
 
-    def _parse_stress_cmd(self, stress_cmd, params):
-        # Due to an issue with scylla & cassandra-stress - we need to create the counter table manually
-        if 'counter_' in stress_cmd:
-            self._create_counter_table()
-
-        if 'compression' in stress_cmd:
-            if 'keyspace_name' not in params:
-                compression_prefix = re.search('compression=(.*)Compressor', stress_cmd).group(1)
-                keyspace_name = "keyspace_{}".format(compression_prefix.lower())
-                params.update({'keyspace_name': keyspace_name})
-
-        return params
-
     def _create_counter_table(self):
         """
         workaround for the issue https://github.com/scylladb/scylla-tools-java/issues/32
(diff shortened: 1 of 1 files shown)


sdcm/utils/loader_utils.py
```diff
@@ -87,29 +74,30 @@ def _run_all_stress_cmds(self, stress_queue, params):
...
-                stress_queue.append(self.run_stress_thread(**params))
+                stress_queue.append(self.run_stress_thread(**stress_params))
 
             time.sleep(10)
 
-            # Remove "user profile" param for the next command
-            if 'profile' in params:
-                del params['profile']
-
-            if 'keyspace_name' in params:
```
fruch, line 110: How this is a fix ?

The multiple keyspaces would be broken because of this

sdcm/utils/loader_utils.py
```diff
@@ -87,29 +74,30 @@ def _run_all_stress_cmds(self, stress_queue, params):
...
             self.log.debug('stress cmd: {}'.format(stress_cmd))
             if stress_cmd.startswith('scylla-bench'):
                 stress_queue.append(self.run_stress_thread(stress_cmd=stress_cmd,
                                                            stats_aggregate_cmds=False,
                                                            round_robin=self.params.get('round_robin')))
-            elif stress_cmd.startswith('cassandra-harry'):
-                stress_queue.append(self.run_stress_thread(**params))
             else:
-                stress_queue.append(self.run_stress_thread(**params))
+                stress_queue.append(self.run_stress_thread(**stress_params))
```
fruch, line 97: Where is stress_params coming from ?

fruch (commented):
I don't understand what is the nature of this fix, it's not clear it's handling the case that code that was remove is being handled

sdcm/utils/loader_utils.py
soyacz: line 77.
I'm copying params (btw. name is misleading, as we use params for sct config options, while here params refer to stress parameters) so in case of modification (line 88) does not affect all the next stress commands in list

sdcm/utils/loader_utils.py
soyacz: this is actual fix - removing `keyspace_name` from params was causing the issue. 
Now in case of stress params modifications we anyway use copy of original params for next iteration

soyacz:
> I don't understand what is the nature of this fix, it's not clear it's handling the case that code that was remove is being handled

because  `keyspace_name` could be modified during iteration there was deletion of this key - this was actually a bug, as it was always dropping this key. Now, each iteration operates on a copy of original params and can modify accordingly without affection of other iterations.

vponomaryov (approved):
LGTM

fruch (approved):

PR #5695, opened 2023-01-17
fix(k8s): make the Scylla IP change watcher thread be more stable

vponomaryov:
In the `ScyllaPodsIPChangeTrackerThread` thread we set `read` timeout be equal to `3h`.
And it means that if a stream read hangs for any reason then we get next bunch of info out of it only in 3 hours.
And hanging of this kind of stream happens much faster than 3 hours on GKE backend.
    
Knowing that in a common case a Scylla pod gets up in 3-4 minutes, make the timeout be about it - 3m.
    
Mostly, it will lead to the redundant reconnections having alive stream,
but 1 redundant API call in 3 minutes is much better than unpredictable errors caused by the wrong IP mapping.


vponomaryov:
Tested here:
EKS: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-upgrade-platform-k8s-eks/125/consoleFull
GKE: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-upgrade-platform-k8s-gke/109/consoleFull

soyacz (approved):
LGTM

fruch (approved):
LGTM

Issue #5688, opened 2023-01-15
terminate-and-replace: replace operation took more than 8 hours 


yarongilor:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Adding a new node during terminate-and-replace.
Adding the new node exceeded the test timeout of 8 hours.
The scenario is:
1. Terminate node-5. (< t:2023-01-14 13:03:31,872 f:nemesis.py      l:1007 c:sdcm.nemesis         p:DEBUG > sdcm.nemesis.SisyphusMonkey: Set current_disruption -> TerminateAndReplaceNode Node longevity-tls-1tb-7d-2022-2-db-node-97b904bb-5 [16.170.155.37 | 10.0.2.72] (seed: True)
)
2. Add a new node node-6: longevity-tls-1tb-7d-2022-2-db-node-97b904bb-6
3.
``` 
< t:2023-01-14 13:15:35,733 f:db_log_reader.py l:113  c:sdcm.db_log_reader   p:DEBUG > 2023-01-14T13:15:35+00:00 longevity-tls-1tb-7d-2022-2-db-node-97b904bb-6     !INFO | scylla[7733]:  [shard  0] storage_service - Starting to bootstrap...
```
5.  Init is not finished in 8 hours so test aborted this operation: 
```
< t:2023-01-14 21:15:41,433 f:nemesis.py      l:1113 c:sdcm.nemesis         p:WARNING > sdcm.nemesis.SisyphusMonkey: TestConfig of the 'Node longevity-tls-1tb-7d-2022-2-db-node-97b904bb-6 [13.49.224.84 | 10.0.3.2] (seed: False)' failed, removing it from list of nodes
< t:2023-01-14 21:15:41,433 f:nemesis.py      l:1115 c:sdcm.nemesis         p:WARNING > sdcm.nemesis.SisyphusMonkey: Node will not be terminated. Please terminate manually!!!
< t:2023-01-14 21:15:41,434 f:decorators.py   l:69   c:sdcm.utils.decorators p:DEBUG >     wait.wait_for(func=self.db_up, step=60, text=text, timeout=timeout, throw_exc=True)
< t:2023-01-14 21:16:41,490 f:nemesis.py      l:1102 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.SisyphusMonkey: Adding new node to cluster...

```
6. Init finished only 15 minutes later:
```
[...]

PR #5700, opened 2023-01-17
fix(cluster.py): Increase timeout of replace-node init to 9 hours

yarongilor:
In ics-1tb longevity, adding a new node duration took more than 8 hours.
Thus we want to increase replace-node timeout to 9 hours.
Setting a soft-timeout of 9 hours and a hard-timeout of extra 3 hours.
Refs: #11960 , https://github.com/scylladb/scylla-cluster-tests/issues/5688

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -55,6 +55,7 @@ from sdcm.cluster import (
     NodeSetupFailed,
     NodeSetupTimeout,
     NodeStayInClusterAfterDecommission,
+    HOUR_IN_SEC,
 )
 from sdcm.cluster_k8s import PodCluster, ScyllaPodCluster
 from sdcm.cluster_k8s.mini_k8s import LocalKindCluster
@@ -78,7 +79,7 @@ from sdcm.sct_events.group_common_events import (ignore_alternator_client_errors
                                                  ignore_ycsb_connection_refused, decorate_with_context)
 from sdcm.sct_events.loaders import CassandraStressLogEvent
 from sdcm.sct_events.nemesis import DisruptionEvent
-from sdcm.sct_events.system import InfoEvent
+from sdcm.sct_events.system import InfoEvent, SoftTimeoutEvent
 from sdcm.sla.sla_tests import SlaTests
 from sdcm.utils import cdc
 from sdcm.utils.common import (get_db_tables, generate_random_string,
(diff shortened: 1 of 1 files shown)


sdcm/nemesis.py
```diff
@@ -1171,8 +1172,10 @@ def _add_and_init_new_cluster_node(self, old_node_ip=None, timeout=MAX_TIME_WAIT
         self.monitoring_set.reconfigure_scylla_monitoring()
         self.set_current_running_nemesis(node=new_node)  # prevent to run nemesis on new node when running in parallel
         new_node.replacement_node_ip = old_node_ip
+        hard_timeout = soft_timeout + HOUR_IN_SEC * 3  # Setting a hard-timeout with an extra 3 hours
```
fruch, line 1175: Would be a bit more useful if the caller of the function can set both soft and hard

fruch: Also can we limit the change to the replace operation, and not change the timing for all of the nemesis that add a node ?

fruch (changes requested):

sdcm/nemesis.py
yarongilor: added `hard_limit`.

As for limiting the change - there is no option and there's no need doing it. and, in addition, the operation is not what matters here but the data-set size.
So if a new nemesis wish a different value - it could easily be set by the parameters.

fruch: That not exactly accurate, since currently we have RBO on specific operation and not on all of them.

Anyhow keeping a very large timeouts, basically mean that if we are gonna have an issue that is starts slowing, we'll never gonna notice it in small data set, and we'll also figure it out really really late, like now when we wanted to release.

yarongilor: ok, fixed - it is now increased only for replace-node.

sdcm/nemesis.py
```diff
@@ -1159,7 +1160,8 @@ def add_ldap_configuration_to_node(node):
             raise LdapNotRunning("LDAP server was supposed to be running, but it is not")
 
     @retrying(n=3, sleep_time=60, allowed_exceptions=(NodeSetupFailed, NodeSetupTimeout))
```
yarongilor, line 1162: @fruch , @fgelcer , following new usage of hard-timeout - we can consider removing this retry.

fruch: we shouldn't consider this, for `replace_node` case we shouldn't be doing retries

fgelcer:
@roydahan @fruch , do we really want to do this?
after an issue with replace node timing out, we have a longer replace node operation, and the issue is marked as regression here --> https://github.com/scylladb/scylla-enterprise/issues/2600

sdcm/nemesis.py
fgelcer: @yarongilor ?

roydahan:
No, we don't want this.

PR #5707, opened 2023-01-20
fix(k8s): allow to use custom cassandra-stress image in dynamic loaders

vponomaryov:
For the moment, using `dynamic` loader run type on K8S, we allow to use only the same Scylla image as the test target Scylla one when our stress tool is set to be `cassandra-stress`.

This limitation disallows us to test latest stable Scylla versions with the serverless feature of the `scylla-operator`.

So, fix it by considering the `stress_image.cassandra-stress` option first and only if it is not set then use the test target Scylla version.


fruch (approved):
LGTM

PR #5708, opened 2023-01-20
fix(k8s-eks): use 3 auxiliary nodes on EKS by default

vponomaryov:
If turned out that existing `2 auxiliary nodes` is not enough when we enable `scylla-manager` and `serverless` features in addition to all other pods which get installed always.

So, make `EKS` use `3 auxiliary nodes` by default to avoid scheduling problems.

### DIFF
--- a/defaults/k8s_eks_config.yaml
+++ b/defaults/k8s_eks_config.yaml
@@ -43,7 +43,7 @@ n_monitor_nodes: 1
 n_loaders: 1
 n_db_nodes: 4
 k8s_n_scylla_pods_per_cluster: 3
-k8s_n_auxiliary_nodes: 2
+k8s_n_auxiliary_nodes: 3
 
 # TODO: add '--abort-on-seastar-bad-alloc' arg to the 'append_scylla_args' option when
 #       following operator bug gets fixed: https://github.com/scylladb/scylla-operator/issues/991
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -1566,9 +1566,9 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         self.k8s_cluster.deploy_node_pool(
             eks.EksNodePool(
                 name=self.k8s_cluster.AUXILIARY_POOL_NAME,
+                # NOTE: It should have at least 5 vCPU to be able to hold all the pods
                 num_nodes=self.params.get('k8s_n_auxiliary_nodes'),
                 instance_type="t3.large",
-                # It should have at least 3 vCPU to be able to hold all the pods
                 disk_size=40,
                 role_arn=self.k8s_cluster.nodegroup_role_arn,
                 k8s_cluster=self.k8s_cluster),


vponomaryov:
Proof that we run out of available resources using only 2 auxiliary nodes of the `t3.large` type is here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-eks-terminate-replace/12/consoleFull

K8S event from it:

```
scylla-manager   73s   Warning   FailedScheduling   pod/scylla-manager-manager-dc-manager-rack-0   default-scheduler   \
    0/7 nodes are available: 2 Insufficient cpu, 5 node(s) didn't match Pod's node affinity/selector. preemption: \
    0/7 nodes are available: 2 No preemption victims found for incoming pod, 5 Preemption is not helpful for scheduling.   \
11m   3   scylla-manager-manager-dc-manager-rack-0.173bd41ac5ae9ecb
```

defaults/k8s_eks_config.yaml
```diff
@@ -43,7 +43,7 @@ n_monitor_nodes: 1
 n_loaders: 1
 n_db_nodes: 4
 k8s_n_scylla_pods_per_cluster: 3
-k8s_n_auxiliary_nodes: 2
+k8s_n_auxiliary_nodes: 3
```
fruch, line 46: this mean all the jobs not using sni_proxy would have one extra unused node.

I would argue for building a server configuration, that would hold those parameter need for a server job.

fruch (approved):
LGTM

defaults/k8s_eks_config.yaml
vponomaryov: > this mean all the jobs not using sni_proxy would have one extra unused node.

False. It will be used, it will not be loaded that much as it is now.

> I would argue for building a server configuration, that would hold those parameter need for a server job.

I didn't understand this statement at all.

fruch: > > this mean all the jobs not using sni_proxy would have one extra unused node.
> 
> False. It will be used, it will not be loaded that much as it is now. It directly influences workability of all the pods on the default node pool. Faster work of mgmt and other side-effects...

if it wasn't for serverless, we wouldn't have enlarged this number, right ? 
If it was o.k. for us running the cases we had with pool of 2 nodes, we can keep working with that

> 
> > I would argue for building a server configuration, that would hold those parameter need for a server job.
> 
> I didn't understand this statement at all.

We should make a configuration for serverless, with the relevant numbers.

soyacz:
What do you think of dropping `k8s_n_auxiliary_nodes` param and make a function that based on other params (like if we use manager, sni proxy, etc.) that returns proper value for given run?

It also is important when we need to scale SNIProxy pods for larger runs. Making automatic would take responsibility off from tester who does not know what number to pick for a new test.

defaults/k8s_eks_config.yaml
vponomaryov: It is strange to me to read it when we have this:

Screenshot from 2023-01-23 09-57-31

vponomaryov:
> What do you think of dropping `k8s_n_auxiliary_nodes` param and make a function that based on other params (like if we use manager, sni proxy, etc.) that returns proper value for given run?
> 
> It also is important when we need to scale SNIProxy pods for larger runs. Making automatic would take responsibility off from tester who does not know what number to pick for a new test.

Not only number of nodes matters, also instance type matters.
Also , SNI proxy replicas may change.
So, the automatic way may be implemented, but it is also doesn't guarantee us to be more efficient than just setting 3 simple instance types by default.

defaults/k8s_eks_config.yaml
fruch: when at the beginning it was one single value, it made sense to no complicate it with multiple configuration files
now that we see more and more values we would need to set for the sake of serverless/ingress configuration.

revisiting this idea might be in order.

altering all the defaults to match what we need for serverless, for functional test I wouldn't mind much.
but now it's effects all of the EKS based test, and it's a bit more problematic, and cost for longevity, we should think how to keep that at bay (k8s setup is moe expensive anyhow, for various reasons, and serverless gonna be more cause of loadblancers)

Issue #5495, opened 2022-11-28
Fullscan thread isn't stopping on test failures


fruch:
Test failed cause of critical error:
```
16:55:45  < t:2022-11-28 14:55:45,363 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:ERROR > sdcm.tester.CriticalTestFailure: Critical Error has failed the test
```

for 25min later the scan thread is still working (and printing like crazy info level logs)

looking at the code, seems like we now have two loops, the one in `run` and one inside `run_next_scan_operation`
`operation_queue` is filled up with (OPERATION_QUEUE_MAXSIZE = 10000) items to execute

and then execute all of them without a sleep in the middle, and without checking `nemesis_termination_event`

also nothing is filling `operation_queue` again, on on a longer test case it might finish it's work, and wouldn't do anything else.

```
    def run_next_scan_operation(self):
        try:
            while not self.operation_queue.empty():
                scan_op = self.operation_queue.queue.pop()
                self.log.info("Going to run fullscan operation %s", scan_op.__class__.__name__)
                scan_op.run_scan_operation()

            self.log.debug("Fullscan operations queue depleted.")

        except Exception as exc:  # pylint: disable=broad-except
            self.log.warning("Encountered exception while performing a fullscan operation:\n%s", exc)

    def run(self):
        end_time = time.time() + self.fullscan_params.duration
        self.wait_until_user_table_exists()

        while time.time() < end_time and not self.fullscan_params.db_cluster.nemesis_termination_event.is_set():
            self.fullscan_stats.read_pages = self.generator.choice([100, 1000, 0])
            self.run_next_scan_operation()
            time.sleep(self.fullscan_params.interval)
        self.log.info("Fullscan stats:\n%s", self.fullscan_stats.get_stats_pretty_table())
```

test_id: 14bd9a3c-5dfc-4f90-9bd6-8e3e1450cdd7
Job:
https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-large-partition-asymmetric-cluster-3h-test/12/

fruch:
> we are seeing it again in the following tests id: `c125b61b-82a7-4b68-8009-1cbf938e3051` `b71d3e31-f4fc-45e5-bcd1-8349145ab76c`

it's not related to this issue, i.e. this issue was about the new implementation

I would suggestion opening a new one with more specific info  ( and not just test_id )

KnifeyMoloko:
[...]

PR #5710, opened 2023-01-20
refactor(fullscan): introducing back the Fullscan refactoring, Fixing: #5495

temichus:
Fixing: https://github.com/scylladb/scylla-cluster-tests/issues/5495
removed Queue populating logic. Now operation objects are created
on the fly(as they are quite light)

the added improvement according to comments from the original PR to use
parameters getting as params.get(key, default=None)

minor logic fixes
use termination_event as a parameter instead of hardcode nemesis_termination_event
fix exception in get_stats_pretty_table() in case of stat.exceptions is
not empty
fix AGG_COUNT_ALL command format and it's use in
FullScanAggregatesOperation

### DIFF
--- a/sdcm/scan_operation_thread.py
+++ b/sdcm/scan_operation_thread.py
@@ -6,12 +6,13 @@ import re
 import tempfile
 import threading
 import time
+import traceback
 from abc import abstractmethod
 from dataclasses import dataclass, field, fields
 from string import Template
 from typing import Optional, Type, NamedTuple, Literal
 
-from cassandra import ConsistencyLevel
+from cassandra import ConsistencyLevel, OperationTimedOut
 from cassandra.cluster import ResponseFuture, ResultSet  # pylint: disable=no-name-in-module
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 from prettytable import PrettyTable
@@ -24,16 +25,19 @@ from sdcm.sct_events.database import FullScanEvent, FullPartitionScanReversedOrd
     FullScanAggregateEvent
 from sdcm.utils.common import get_table_clustering_order, get_partition_keys
 
-ERROR_SUBSTRINGS = ("timed out", "unpack requires", "timeout")
-TIMEOUT_VALUES = [2, 4, 8, 30, 120, 300]
+ERROR_SUBSTRINGS = ("timed out", "unpack requires", "timeout", 'Host has been marked down or removed')
+FULL_SCAN_TIMEOUT_VALUES = [2, 4, 8, 30, 120, 300]
+AGGREGATION_TIMEOUT_VALUES = [15, 30, 60, 120, 240, 480, 920, 1800]
 BYPASS_CACHE_VALUES = [" BYPASS CACHE", ""]
 LOCAL_CMD_RUNNER = LocalCmdRunner()
 
 
 # pylint: disable=too-many-instance-attributes
+
+
 @dataclass
 class FullScanParams:
-    mode: Literal['random', 'table', 'partition', 'aggregate']
+    mode: Literal['random', 'table', 'partition', 'aggregate', 'table_and_aggregate']
     termination_event: threading.Event
     db_cluster: [BaseScyllaCluster, BaseCluster] = None
     ks_cf: str = "random"
@@ -48,6 +52,7 @@ class FullScanParams:
     validate_data: bool = False
     include_data_column: bool = None
     rows_count: int = 5000
+    aggregate_operation_limit: int = 60*30  # 30 min by default can be degree job by job
 
 
 @dataclass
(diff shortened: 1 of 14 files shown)


fruch:
@temichus I think the title of such a PR should be that it introducing back the Fullscan refactoring

Also it should state (so the issue would close automatically when this get merged):
Fixing: #5495

the original change affected quite a number of jobs, we should select a few of them, and show working run of them before merging such a change.

fruch:
> @temichus I think the title of such a PR should be that it introducing back the Fullscan refactoring
> 
> Also it should state (so the issue would close automatically when this get merged): Fixing: #5495
> 
> the original change affected quite a number of jobs, we should select a few of them, and show working run of them before merging such a change.

Also I would consider making sure we have unittest and integration test for Fullscan code, that have quite lots of logic backed into it now.

sdcm/tester.py
```diff
@@ -2044,7 +2044,7 @@ def verify_gemini_results(self, queue):
                          'errors': stats['errors']})
         return stats
 
-    def run_fullscan_thread(self, ks_cf='random', interval=1, duration=None):
+    def run_fullscan_thread(self, fullscan_params: dict, thread_name: str):
```
soyacz, line 2047: why not to use `FullScanParams` class?

sdcm/scan_operation_thread.py
```diff
@@ -1,163 +1,300 @@
...
-        self.user = kwargs.get("user", None)
-        self.password = kwargs.get("password", None)
-        self._thread = threading.Thread(daemon=True, name=self.__class__.__name__, target=self.run)
-
-    def wait_until_user_table_exists(self, db_node, table_name: str = 'random', timeout_min: int = 20):
-        text = f'Waiting until {table_name} user table exists'
-        if table_name.lower() == 'random':
-            wait.wait_for(func=lambda: len(self.db_cluster.get_non_system_ks_cf_list(db_node)) > 0, step=60,
-                          text=text, timeout=60 * timeout_min, throw_exc=True)
+        nemesis_seed = self.fullscan_params.db_cluster.params.get("nemesis_seed")
```
soyacz, line 129: there's possibility that nemesis_seed is not set (not always using SisyphusMonkey), specify some random default.

longevity_test.py
```diff
@@ -40,12 +40,10 @@ def _get_tombstone_gc_verification_params(self) -> dict:
...
-    def _get_scan_operation_params(self, scan_operation: str) -> dict:
-        params = {}
-        if scan_operation_params := self.params.get(scan_operation):
-            params = json.loads(scan_operation_params)
-            self.log.info('Scan operation %s params are: %s', scan_operation, params)
-        return params
+    def _get_scan_operation_params(self) -> list[dict]:
+        params = self.params.get("run_fullscan", {})
+        self.log.info('Scan operation params are: %s', params)
+        return [json.loads(item) for item in params]
```
soyacz, line 46: this could return `FullScanParams` instance

sdcm/scan_operation_thread.py
```diff
@@ -1,163 +1,300 @@
...
+
+        self.scan_operation_instance_map = {
+            "random": lambda: self.generator.choice(
+                [FullScanOperation, FullPartitionScanOperation, FullScanAggregatesOperation]),
+            "table": lambda: FullScanOperation,
+            "partition": lambda: FullPartitionScanOperation,
+            "aggregate": lambda: FullScanAggregatesOperation
+        }
+
+    def run_next_scan_operation(self):
```
soyacz, line 142: why as public?

sdcm/scan_operation_thread.py
```diff
@@ -1,163 +1,300 @@
...
+            time.sleep(self.fullscan_params.interval)
+        self.log.info("Fullscan stats:\n%s", self.fullscan_stats.get_stats_pretty_table())
+
+    def start(self):
+        self._thread.start()
+
+    def join(self, timeout=None):
+        return self._thread.join(timeout)
+
+    def wait_until_user_table_exists(self, timeout_min: int = 20):
```
soyacz, line 170: why public?

sdcm/scan_operation_thread.py
soyacz: yes, I meant that but I was not precise.

longevity_test.py
```diff
@@ -77,11 +92,11 @@ def test_custom_time(self):
         self.run_pre_create_keyspace()
         self.run_pre_create_schema()
 
-        if fullscan_params := self._get_scan_operation_params(scan_operation='run_fullscan'):
-            self.run_fullscan_thread(ks_cf=fullscan_params['ks_cf'], interval=fullscan_params['interval'])
-
-        if full_partition_scan_params := self._get_scan_operation_params(scan_operation='run_full_partition_scan'):
-            self.run_full_partition_scan_thread(**full_partition_scan_params)
+        if scan_operation_params := self._get_scan_operation_params():
```
soyacz, line 95: this if seems redundant. In case of empty scan operation params list, for loop skips anyway

sdcm/scan_operation_thread.py
```diff
@@ -1,163 +1,300 @@
...
-        if table_name.lower() == 'random':
-            wait.wait_for(func=lambda: len(self.db_cluster.get_non_system_ks_cf_list(db_node)) > 0, step=60,
-                          text=text, timeout=60 * timeout_min, throw_exc=True)
+        nemesis_seed = self.fullscan_params.db_cluster.params.get("nemesis_seed")
+        self.generator = random.Random(int(nemesis_seed)) if nemesis_seed else random.Random()
+        self._thread = threading.Thread(daemon=True, name=f"{self.__class__.__name__}_{thread_name}", target=self.run)
+        self.termination_event = fullscan_params.termination_event
+
+        self.scan_operation_instance_map = {
+            "random": lambda: self.generator.choice(
```
soyacz, line 135: why do we need this `lambda`?

sdcm/scan_operation_thread.py
```diff
@@ -337,63 +482,123 @@ def _compare_output_files(self):
...
+
+        if self._validate_fullscan_result(cmd_result):
+            self.scan_event.message = f"{type(self).__name__} operation ended successfully"
+        else:
+            self.scan_event.severity = Severity.WARNING
+            self.scan_event.message = f"{type(self).__name__} operation failed."
+
+        return cmd_result
+
+    def _validate_fullscan_result(self, cmd_result: ResultSet):
```
soyacz, line 553: how about returning a failure message only?
In case no message -> all good, operation successful
In case of message -> add this message to operation failed event

and in case of failing to find `Dispathing forward request...` also return this message so we know operation failure reason

soyacz: ok, I see, this comes from reverted commit so I cannot refer to it easily.
Generally, I was thinking of:
function `_validate_fullscan_result` in `FullScanAggregatesOperation` returns the status of operation (`False`, for failure and 'True` for success) but gives no information about the reason. When investigating, one must go into logs to check the reason.

I propose `_validate_fullscan_result` to return the reason of failure as string (empty string or None for success).
This message should be appended to the warning event.
We can consider setting `self.scan_event.severity` inside  `_validate_fullscan_result` and dropping some logic from `execute_query` - so we don't have to return anything.

soyacz:
@temichus mark it as ready for review when it's ready.

soyacz:
@temichus remove Draft from PR

unit_tests/test_scan_operation_thread.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def get_event_log_file(events):
+    if (log_file := Path(events.temp_dir, "events_log", "events.log")).exists():
+        return log_file.read_text(encoding="utf-8").rstrip().split('\n')
+    return ""
+
+
+@pytest.fixture(scope='function', autouse=True)
+def cleanup_fife(events):
```
fruch, line 62: why do we need this cleanup ? 

and if we do, why not removing the files ? why clearing the content ?

and what's fife ?

fruch (dismissed):
LGTM,

few nipticks about the unittest

fruch (approved):

PR #5711, opened 2023-01-20
improvement(k8s): use C-S image with serverless feature support

vponomaryov:
Specifying a new enough Scylla image that supports the 'serverless' feature for the 'cassandra-stress' loaders will allow us to test the feature against any Scylla version.

With this configuration we will only need to set 'true' value for the 'k8s_enable_tls' option when we want
to test the 'serverless' feature.

### DIFF
--- a/defaults/k8s_gke_config.yaml
+++ b/defaults/k8s_gke_config.yaml
@@ -64,3 +64,9 @@ append_scylla_args: '--abort-on-lsa-bad-alloc 1 --abort-on-internal-error 1 --ab
 docker_image: ''
 backup_bucket_backend: 's3'
 backup_bucket_location: 'minio-bucket'
+
+# NOTE: Define a separate image for the 'cassandra stress' that supports 'serverless' feature.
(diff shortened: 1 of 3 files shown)


fruch (approved):
LGTM

defaults/k8s_gke_config.yaml
```diff
@@ -64,3 +64,9 @@ append_scylla_args: '--abort-on-lsa-bad-alloc 1 --abort-on-internal-error 1 --ab
 docker_image: ''
 backup_bucket_backend: 's3'
 backup_bucket_location: 'minio-bucket'
+
+# NOTE: Define a separate image for the 'cassandra stress' that supports 'serverless' feature.
+#       It will allow to test 'serverless' feature against any stable Scylla release
+#       with old 'cassandra-stress' binary.
+stress_image:
+  cassandra-stress: 'scylladb/scylla-nightly:5.2.0-rc0-0.20230119.34ab98e1be2e'
```
fruch, line 72: one might argue we don't need this on GKE, since we don't have serverless there...

PR #5713, opened 2023-01-22
fix(k8s_local): ingress maxconn value is too high

fruch:
Since we added some of the producation configuration to the ingress haproxy, the values for `maxconn` is too high for the local kind setup and fails with the following error:

```
[ALERT]    (286) : config : Cannot compute the automatic maxsslconn because
 global.maxconn is already too high for the global.memmax value (31 MB).
 The absolute maximum possible value without SSL is 780, but 70000 was found
 and SSL is in use.
```

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -1036,17 +1036,22 @@ class KubernetesCluster(metaclass=abc.ABCMeta):  # pylint: disable=too-many-publ
                 memory_limit = convert_memory_units_to_k8s_value(memory_limit)
                 cpu_request = convert_cpu_units_to_k8s_value(cpu_request)
                 memory_request = convert_memory_units_to_k8s_value(memory_request)
+                haproxy_maxconn = 70000
             else:
-                cpu_limit = '100m'
-                memory_limit = '50M'
+                # TODO: measure these values in performance testing and update if needed.
+                cpu_limit = '400m'
+                memory_limit = '200M'
                 cpu_request = '100m'
                 memory_request = '50M'
+                # that's the default, and trigger haproxy auto calculation of it.
+                haproxy_maxconn = 0
 
             environ = {
                 "POD_CPU_LIMIT": cpu_limit,
                 "POD_MEMORY_LIMIT": memory_limit,
                 "POD_CPU_REQUEST": cpu_request,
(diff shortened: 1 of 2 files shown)


sdcm/cluster_k8s/__init__.py
```diff
@@ -1036,17 +1036,20 @@ def deploy_ingress_controller(self, pool_name: str = None):
                 memory_limit = convert_memory_units_to_k8s_value(memory_limit)
                 cpu_request = convert_cpu_units_to_k8s_value(cpu_request)
                 memory_request = convert_memory_units_to_k8s_value(memory_request)
+                haproxy_maxconn = 70000
             else:
-                cpu_limit = '100m'
-                memory_limit = '50M'
+                cpu_limit = '400m'
+                memory_limit = '200M'
```
vponomaryov, line 1042: Where these numbers come from? Is it calculated?

sdcm/cluster_k8s/__init__.py
```diff
@@ -1036,17 +1036,20 @@ def deploy_ingress_controller(self, pool_name: str = None):
...
                 cpu_request = '100m'
                 memory_request = '50M'
+                haproxy_maxconn = 0
```
vponomaryov, line 1045: Why `0`?

fruch: that's the default, and would make it to calculate it on it's own

sdcm/cluster_k8s/__init__.py
fruch: mostly trial and error, before we didn't had limits, only request
I've just it's just request x 4.

sdcm/cluster_k8s/__init__.py
vponomaryov: Please, add a comment above this line saying it will trigger an autocalculation.
Because it is unclear and not only I may have this question.

sdcm/cluster_k8s/__init__.py
vponomaryov: Please, add a comment above something like following:
```
# TODO: measure these values in performance testing and update if needed.
```

vponomaryov (dismissed):
LGTM

soyacz (approved):

soyacz:
Merging, as we need it

PR #5722, opened 2023-01-24
improvement(parse_version): align to expected version format 

fgelcer:
the package complains about our Scylla versions
when we have the full version format including
commit hash and date (e.g. `5.2.0~rc0-0.20230119.34ab98e1be2e`)
with tilde or dash.
this change will give to the `parse_version` function
only `major.minor.micro` for the sake of comparisons,
as everything else is ignored anyway.

### DIFF
--- a/sdcm/utils/version_utils.py
+++ b/sdcm/utils/version_utils.py
@@ -21,6 +21,7 @@ from collections import namedtuple
 from urllib.parse import urlparse
 from functools import lru_cache, wraps
 from itertools import count
+from pkg_resources import parse_version
 
 import yaml
 import boto3
@@ -115,6 +116,11 @@ FILE_REGEX_DICT = {
 # The variable "type" indices the type of URL (Debian or Yum)
 # The variable "urls" contains all urls using the Scylla pattern
 RepositoryDetails = namedtuple("RepositoryDetails", ["type", "urls"])
+
+
+def parse_scylla_version(version_to_parse: str) -> version.Version:
+    version_format = re.compile(r'(\d+\.\d)|(\.\d+)')
+    return parse_version(version_format.search(version_to_parse)[0])
 
 
 @lru_cache(maxsize=1024)
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -29,11 +29,11 @@ from cassandra import InvalidRequest
 from cassandra.util import sortedset, SortedSet  # pylint: disable=no-name-in-module
 from cassandra import ConsistencyLevel
 from cassandra.protocol import ProtocolException  # pylint: disable=no-name-in-module
-from pkg_resources import parse_version
 
 from sdcm.tester import ClusterTester
 from sdcm.utils.decorators import retrying
 from sdcm.utils.cdc.options import CDC_LOGTABLE_SUFFIX
+from sdcm.utils.version_utils import parse_scylla_version
 
 
 LOGGER = logging.getLogger(__name__)
@@ -3141,7 +3141,7 @@ class FillDatabaseData(ClusterTester):
 
     @property
     def parsed_scylla_version(self):
-        return parse_version(self.db_cluster.nodes[0].scylla_version.replace('~', '-'))
+        return parse_scylla_version(self.db_cluster.nodes[0].scylla_version)
 
     @property
     def is_enterprise(self) -> bool:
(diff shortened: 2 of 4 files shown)


fgelcer:
or perhaps we could remove the whole `if` statement, as anything older than `4.6` isn't supported anymore, no?

fgelcer:
> > or perhaps we could remove the whole `if` statement, as anything older than `4.6` isn't supported anymore, no?
> 
> As I said before we should replace all the calls to `parse_version`, keeping patching like that, one by one you'll end with it next year.
> 
> Regardless if you'll remove one if or other

the problem is, this `parse_version` complains for both dash and tilde, and i see it being used in a manager test (but won't be affected, as it uses only the `major` part of the parsed version, and in the upgrade test (in multiple places, and also in the `fill_db_data`

i will change this patch for that, instead the replace i already used

fgelcer:
> > or perhaps we could remove the whole `if` statement, as anything older than `4.6` isn't supported anymore, no?
> 
> As I said before we should replace all the calls to `parse_version`, keeping patching like that, one by one you'll end with it next year.
> 
> Regardless if you'll remove one if or other

@fruch , changed

fruch (changes requested):
Seems like this change is breaking the unit tests

sdcm/utils/version_utils.py
```diff
@@ -117,6 +118,12 @@ class ScyllaFileType(Enum):
 RepositoryDetails = namedtuple("RepositoryDetails", ["type", "urls"])
 
 
+@property
+def parse_scylla_version(version_to_parse: str) -> version.Version:
```
vponomaryov, line 122: Maybe it should handle also `latest` tag/version?

soyacz, line 122: we use also `SCYLLA_VERSION_RE` and `SCYLLA_VERSION_GROUPED_RE` from this file.
How about places that use it?

fgelcer: it is called only when we have already a version selected... it doesn't matter to us what we are passing to the pipeline, prior to having an image selected, but this is for version comparisons only, when we need to know if a feature we want to test is already in that version we are running on

fgelcer: and BTW, now there is a change that will get either `digit.digit` or `digit.digit.digit`, so `5.2.latest` or `5.3.dev` will be covered by it now

fgelcer: they are being used in different places, to save the exact Scylla[-enterprise] version, including `build-id`, rather than this change, that will parse only the major and minor from the versions, for the sake of knowing if some features are supported in the current test setup or not

soyacz (commented):
parse_scylla_version - would be good to add unit test to this function. Should be easy and show examples what is handled and how.

fgelcer:
> parse_scylla_version - would be good to add unit test to this function. Should be easy and show examples what is handled and how.

it is already being tested, under the function https://github.com/scylladb/scylla-cluster-tests/blob/664ed61e0444191632937ba4337ecabb54638a71/utils/get_supported_scylla_base_versions.py#L142

it is being called by unittest (that i have already enhanced here -> https://github.com/fgelcer/scylla-cluster-tests/blob/9ca55bf304ed9ddea6694c926a0c987796183311/unit_tests/test_base_version.py)

soyacz:
> > parse_scylla_version - would be good to add unit test to this function. Should be easy and show examples what is handled and how.
> 
> it is already being tested, under the function
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/664ed61e0444191632937ba4337ecabb54638a71/utils/get_supported_scylla_base_versions.py#L142
> 
> it is being called by unittest (that i have already enhanced here -> https://github.com/fgelcer/scylla-cluster-tests/blob/9ca55bf304ed9ddea6694c926a0c987796183311/unit_tests/test_base_version.py)



> > parse_scylla_version - would be good to add unit test to this function. Should be easy and show examples what is handled and how.
> 
> it is already being tested, under the function
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/664ed61e0444191632937ba4337ecabb54638a71/utils/get_supported_scylla_base_versions.py#L142
> 
> it is being called by unittest (that i have already enhanced here -> https://github.com/fgelcer/scylla-cluster-tests/blob/9ca55bf304ed9ddea6694c926a0c987796183311/unit_tests/test_base_version.py)

but how about case `5.2.0~rc0-0.20230119.34ab98e1be2e` is it covered?

fgelcer:
> but how about case `5.2.0~rc0-0.20230119.34ab98e1be2e` is it covered?

@soyacz , right now `5.2.0` is not yet promoted, so, even if i added a test for it right now, once it will be released, we will need to do changes in this test as well, but it is not any different from `5.1.0`, as this version, and its base version (`5.0.x`) have the same format

the `5.2.0~rc0` is now under test in the rolling upgrade test itself, and it seems to not be a problem anymore

PR #5725, opened 2023-01-25
improvement(kernel traces): Enable print_kernel_callstack by default

yarongilor:
The Reactor-Stall trace might be less accurate than in kernel traces.
	It would be helpfull to analyse reactor-stall issues if these traces
	are enabled for all tests by default.

Refs: https://github.com/scylladb/scylladb/issues/12623

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -58,7 +58,7 @@ use_cloud_manager: false
 cloud_prom_bearer_token: ''
 
 backtrace_decoding: true
-print_kernel_callstack: false
+print_kernel_callstack: true
 
 update_db_packages: ''
 


fgelcer:
why is it a draft? move it to `ready to review` and let's get it merged

roydahan:
Wait, why?
We disabled it by default on purpose since it was polluting the log.

fgelcer:
IIRC we have never enabled it to all, but there was a recent issue, and the investigation is showing a problem in the kernel, so Avi requested to enable it to all jobs...
it will add few hundred new prints to the logs, and IMHO it should not be a problem, but if it will be, we can enable it only for specific jobs

roydahan:
We did and we didn't have a parameter for it, then it started causing issues and you introduced the parameter.

roydahan:
@fgelcer / @yarongilor  are we automatically decoding kernel_callstack?

fgelcer:
> @fgelcer / @yarongilor are we automatically decoding kernel_callstack?

no, we are not... for now, we are saving the `kallsyms` map for every boot to decode later manually

roydahan:
ok, so let's merge and hope that it's indeed not polluting the logs as was claimed in https://github.com/scylladb/scylladb/issues/12623

roydahan (approved):

fruch:
@fgelcer backported to v13

fgelcer:
thank you

fgelcer:
v14 already has this change, so no more backports are needed at this point

PR #5731, opened 2023-01-26
fix(chaos-mesh): fix installation for gke

soyacz:
looks chaos-mesh does not work on GKE as
 chaos-daemons are not deployed.

Added `tolerations` settings for chaos-daemons to use scylla

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5732, opened 2023-01-26
fix(chaos-mesh): make admission-controller not block it on GKE

vponomaryov:

### DIFF
--- a/sdcm/k8s_configs/chaos-mesh/gke-auth-workaround.yaml
+++ b/sdcm/k8s_configs/chaos-mesh/gke-auth-workaround.yaml
@@ -1,0 +1,33 @@
+# NOTE: on GKE it is not enough just to install the 'chaos-mesh' helm chart
+#       Following workaround is needed to be applied:
+#       https://chaos-mesh.org/docs/faqs/#q-the-default-administrator-google-cloud-user-account-is-forbidden-to-create-chaos-experiments-how-to-fix-it
+
+kind: ClusterRole
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: role-cluster-manager-pdmas
+rules:
+  - apiGroups: ['']
+    resources: ['pods', 'namespaces']
+    verbs: ['get', 'watch', 'list']
+  - apiGroups:
+      - chaos-mesh.org
+    resources: ['*']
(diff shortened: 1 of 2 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-gke/21/consoleFull
The failures that exist there happen after the fixed step here - `admission controller checks`.

soyacz:
I did fix in another way: https://github.com/scylladb/scylla-cluster-tests/pull/5731

fruch:
> I did fix in another way: #5731

@soyacz I think @vponomaryov PR is better in this situation, i.e. it's following the FAQ of chaos mesh.

disabling the auth on the dashboard is not recommand:
https://chaos-mesh.org/docs/manage-user-permissions/#enable-or-disable-permission-authentication

yes, we are not using the dashboard, but it's there (and might be exposed externally ?)

soyacz:
> > I did fix in another way: #5731
> 
> @soyacz I think @vponomaryov PR is better in this situation, i.e. it's following the FAQ of chaos mesh.
> 
> disabling the auth on the dashboard is not recommand: https://chaos-mesh.org/docs/manage-user-permissions/#enable-or-disable-permission-authentication
> 
> yes, we are not using the dashboard, but it's there (and might be exposed externally ?)

Without dashboard you can create experiments only with kubectl - which requires credentials. 
I think by default dashboard is not protected anyway - I don't remember need for playing with these tokens like described in docs.

fruch:
> > > I did fix in another way: #5731
> > 
> > 
> > @soyacz I think @vponomaryov PR is better in this situation, i.e. it's following the FAQ of chaos mesh.
> > disabling the auth on the dashboard is not recommand: https://chaos-mesh.org/docs/manage-user-permissions/#enable-or-disable-permission-authentication
> > yes, we are not using the dashboard, but it's there (and might be exposed externally ?)
> 
> Without dashboard you can create experiments only with kubectl - which requires credentials. I think by default dashboard is not protected anyway - I don't remember need for playing with these tokens like described in docs.

when in installed with kind, I don't really care, but when it's installed on EKS/GKE we need to verify it's not open to the whole world without us known about it

fruch (approved):
LGTM

Issue #5506, opened 2022-11-29
Azure: disrupt_rebuild_streaming_err timeouts looking for `Rebuild starts`


fruch:
## Installation details

Kernel Version: 5.15.0-1023-azure
Scylla version (or git commit hash): `5.2.0~dev-20221129.8bc0af9e34fa` with build-id `64a68bc4306bd55bc726a81b066ae3350e79d108`

Cluster size: 6 nodes (Standard_L8s_v3)

Scylla Nodes used in this run:
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-8 (23.96.49.29 | 10.0.0.14) (shards: 7)
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-7 (52.152.233.218 | 10.0.0.14) (shards: 7)
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-6 (172.173.160.18 | 10.0.0.10) (shards: 7)
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-5 (20.231.122.14 | 10.0.0.9) (shards: 7)
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-4 (20.172.155.115 | 10.0.0.8) (shards: 7)
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-3 (4.246.181.200 | 10.0.0.7) (shards: 7)
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-2 (4.246.181.198 | 10.0.0.6) (shards: 7)
- longevity-10gb-3h-master-db-node-54aa0b5e-eastus-1 (20.172.155.182 | 10.0.0.5) (shards: 7)

OS / Image: `/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/scylla-images/providers/Microsoft.Compute/images/scylla-5.2.0-dev-x86_64-2022-11-29T13-26-23` (azure: eastus)

Test: `longevity-10gb-3h-azure-test`
Test id: `54aa0b5e-c3a5-4bcb-b657-6b83d804a11d`
Test name: `scylla-master/longevity/longevity-10gb-3h-azure-test`
Test config file(s):

- [longevity-10gb-3h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/f8224bdc2a2c0d512fbe00b7b4870591c5d3414a/test-cases/longevity/longevity-10gb-3h.yaml)

## Issue description

[...]

PR #5734, opened 2023-01-27
Fix streaming err raciness

KnifeyMoloko:
This PR introduces commits related to the issue https://github.com/scylladb/scylla-cluster-tests/issues/5506 and should fix the failures we've been seeing in relation to `DecommissionStreamingErr` and other `*StreamingErr` nemesis. The failures came about after a `ParallelObject` timed out. This `ParallelObject` is used to run 2 functions in parallel - a trigger function and a log watching function that should interrupt the process started by the trigger function. Timeouts have been caused by the fact that the trigger functions use `nodetool` remoter commands and since the process started by these commands gets interrupted, these functions did not return to the caller before the `ParallelObject`'s own timeout kicked in. 
This PR's commits address this and an additional potential source of racy behavior by: a) moving the log follower to be initiated before the trigger function is called and b) by introducing timeouts to the `nodetool` remoter calls.

Jenkins staging job with the nemesis in question: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/maciej/job/2022.1-10gb-gce/

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -27,7 +27,7 @@ import time
 import traceback
 import json
 
-from typing import Any, List, Optional, Type, Tuple, Callable, Dict, Set, Union
+from typing import Any, List, Optional, Type, Tuple, Callable, Dict, Set, Union, Iterable
 from functools import wraps, partial
 
 from collections import defaultdict, Counter, namedtuple
@@ -3193,7 +3193,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self._wait_all_nodes_un()
 
     def _call_disrupt_func_after_expression_logged(self,
-                                                   expression: str,
+                                                   log_follower: Iterable[str],
                                                    disrupt_func: Callable,
                                                    disrupt_func_kwargs: dict = None,
                                                    sleep: int = 1,
@@ -3206,7 +3206,6 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         arguments after a <delay>.
         """
         start_time = time.time()
-        target_node_logs = self.target_node.follow_system_log(patterns=[expression])
 
         with DbEventsFilter(db_event=DatabaseLogEvent.RUNTIME_ERROR,
                             line="This node was decommissioned and will not rejoin",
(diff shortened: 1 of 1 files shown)


fgelcer (approved):
LGTM, and i have a small comment/question:
shouldn't the `nodetool` commands timeout be a variable, set in a "centralised" in the call itself other than hard coded to each `nodetool` call?

KnifeyMoloko:
> LGTM, and i have a small comment/question: shouldn't the `nodetool` commands timeout be a variable, set in a "centralised" in the call itself other than hard coded to each `nodetool` call?

Not really, at least in my opinion. The different `nodetool` commands can run for widely different amounts of time and event then same command can be expected to run much longer in some scenarios (more data used, heavier load), so I think it's better to use timeout as it is used here.

vponomaryov (approved):
LGTM

sdcm/nemesis.py
```diff
@@ -3260,21 +3259,25 @@ def decommission_post_action():
...
+            log_follower=log_follower,
             disrupt_func=self.reboot_node,
             disrupt_func_kwargs={"target_node": self.target_node, "hard": True, "verify_ssh": True},
             delay=0
         )
+
         ParallelObject(objects=[trigger, watcher], timeout=1200).call_objects()
         if new_node := decommission_post_action():
-            new_node.run_nodetool("rebuild")
+            new_node.run_nodetool("rebuild", timeout=300, retry=3)
```
fruch, line 3278: @fgelcer @vponomaryov, I think this change wasn't related, and isn't needed. i.e. assuming rebuild would always finish in 5min is quite off...

fgelcer: how much do you think we should add here? wait forever isn't a good option neither

fruch: waiting for ever is better option then 5min, that is failing on the release.
in general we shouldn't be setting such limits without some solid data out of previous runs and from multiple test case.

anyhow in this case, it wasn't even the intend fix/goal.

vponomaryov: I agree with @fgelcer , before it was not set, and now it is explicit.
@fruch which value for timeout do you propose?

fruch: I propose to set it back to no timeout:
https://github.com/scylladb/scylla-cluster-tests/pull/5808

Assuming a timeout here without doing the work of checking/validating it with prior data is pointless and causing confusion, like it is right now during the release, @roydahan was a about to disable this nemesis cause it's similar to a different issue with have with rebuild.

you think putting a timeout here is important enough, great put it on a task/issue and we'll assign someone to do the work, slipping such a thing in PR aiming to fix something else is something we should avoid (and we even backported it, adding more confusion on older releases as well)

roydahan: Of course we need to revert it.
How did we approve it?
Rebuild will not finish in 5 minutes for most of our cases.

PR #5735, opened 2023-01-28
fix(k8s): make flush and reshard logic be more stable

vponomaryov:
We have `flush and reshard` nemesis and functional K8S test. Both are identical in steps.
One of the steps it so check that all scylla containers didn't have additional restarts and are successful from a first attempt.

It is not correct assumption. We may get 1 or 2 restarts which do not influence a Scylla cluster workability.
From a user point of view he waits about the same time before a new Scylla member becomes up and running.

So, remove redundant logic for checking restart count which will make the logic be more stable.

### DIFF
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -828,53 +828,19 @@ def test_can_recover_from_fatal_pod_termination(db_cluster):
 def test_nodetool_flush_and_reshard(db_cluster: ScyllaPodCluster):
     target_node = db_cluster.nodes[0]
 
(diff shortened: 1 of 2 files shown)


vponomaryov:
Tested here:
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-eks-nodetool-flush-and-reshard/23/consoleFull
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-functional-eks/4/consoleFull

fruch (approved):
LGTM

PR #5742, opened 2023-01-31
improvement(sct_config): Refer to AMI images by name instead of AMI ID

temichus:
task: https://github.com/scylladb/qa-tasks/issues/268

added functionality to automaticaly convert image name  to ami_ids per region
added unit tests or converting logic
remove unused logic "assume multi dc by n_db_nodes set size" becuse when run it, "_check_multi_region_params" is fail

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -108,7 +108,6 @@ SCT_RUNNER_HOST = get_sct_runner_ip()
 
 LOGGER = setup_stdout_logger()
 
-
 click_completion.init()
 
 
@@ -580,10 +579,12 @@ def list_resources(ctx, user, test_id, get_all, get_all_running, verbose):
               help="architecture of the AMI (default: x86_64)")
 def list_images(cloud_provider: str, branch: str, version: str, region: str, arch: AwsArchType):
     add_file_logger()
-    branch_fields = ["Backend", "Name", "ImageId", "CreationDate", "BuildId", "Arch", "ScyllaVersion"]
-    #  TODO: align branch and version fields once scylla-pkg#2995 is resolved
     version_fields = ["Backend", "Name", "ImageId", "CreationDate"]
-
+    version_fields_with_tag_name = version_fields + ["NameTag"]
+    #  TODO: align branch and version fields once scylla-pkg#2995 is resolved
+    branch_specific_fields = ["BuildId", "Arch", "ScyllaVersion"]
+    branch_fields = version_fields + branch_specific_fields
+    branch_fields_with_tag_name = version_fields_with_tag_name + branch_specific_fields
     if version and branch:
         click.echo("Use --version or --branch, not both.")
         return
@@ -595,7 +596,7 @@ def list_images(cloud_provider: str, branch: str, version: str, region: str, arc
             case "aws":
                 rows = get_ami_images_versioned(region_name=region, arch=arch, version=version)
                 click.echo(
-                    create_pretty_table(rows=rows, field_names=version_fields).get_string(
+                    create_pretty_table(rows=rows, field_names=version_fields_with_tag_name).get_string(
                         title=f"AWS Machine Images by Version in region {region}")
                 )
             case "gce":
@@ -620,7 +621,7 @@ def list_images(cloud_provider: str, branch: str, version: str, region: str, arc
                 region = region or "eu-west-1"
                 ami_images = get_ami_images(branch=branch, region=region, arch=arch)
                 click.echo(
-                    create_pretty_table(rows=ami_images, field_names=branch_fields).get_string(
+                    create_pretty_table(rows=ami_images, field_names=branch_fields_with_tag_name).get_string(
                         title=f"AMI Machine Images for {branch} in region {region}"
                     )
                 )
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -40,6 +40,7 @@ from sdcm.utils.common import (
     get_branched_gce_images,
     get_scylla_ami_versions,
     get_scylla_gce_images_versions,
+    convert_name_to_ami_if_needed,
 )
 from sdcm.utils.version_utils import (
     get_branch_version,
@@ -1550,6 +1551,8 @@ class SCTConfiguration(dict):
         'stress_cmd_lwt_dc', 'stress_cmd_lwt_ue', 'stress_cmd_lwt_uc', 'stress_cmd_lwt_ine',
         'stress_cmd_lwt_d', 'stress_cmd_lwt_u', 'stress_cmd_lwt_i'
     ]
+    ami_id_params = ['ami_id_db_scylla', 'ami_id_loader', 'ami_id_monitor', 'ami_id_db_cassandra', 'ami_id_db_oracle']
+    aws_supported_regions = ['eu-west-1', 'eu-west-2', 'us-west-2', 'us-east-1', 'eu-north-1', 'eu-central-1']
 
     def __init__(self):
         # pylint: disable=too-many-locals,too-many-branches,too-many-statements
@@ -1589,15 +1592,20 @@ class SCTConfiguration(dict):
         region_names = self.region_names
 
         if cluster_backend in ['aws', 'aws-siren', 'k8s-eks']:
-            for region in region_names:
-                if region not in regions_data:
-                    raise ValueError(f"{region} isn't supported, use: {list(regions_data.keys())}")
+            if regions_data:
+                for region in region_names:
+                    if region not in regions_data:
+                        raise ValueError(f"{region} isn't supported, use: {list(regions_data.keys())}")
 
-                for key, value in regions_data.get(region, {}).items():
-                    if key not in self.keys():
-                        self[key] = value
-                    elif len(self[key].split()) < len(region_names):
-                        self[key] += " {}".format(value)
+                    for key, value in regions_data.get(region, {}).items():
+                        if key not in self.keys():
+                            self[key] = value
+                        elif len(self[key].split()) < len(region_names):
+                            self[key] += " {}".format(value)
+            else:
+                for region in region_names:
+                    if region not in self.aws_supported_regions:
+                        raise ValueError(f"{region} isn't supported, use: {self.aws_supported_regions}")
 
         # 3) overwrite with environment variables
         anyconfig.merge(self, env)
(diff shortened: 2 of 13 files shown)


sdcm/sct_config.py
```diff
@@ -1606,17 +1599,13 @@ def __init__(self):
...
-            if num_of_db_nodes_sets > num_of_regions:
-                for region in list(regions_data.keys())[:num_of_db_nodes_sets]:
-                    for key, value in regions_data[region].items():
-                        if key not in self.keys():
-                            self[key] = value
-                        else:
-                            self[key] += " {}".format(value)
+        # 5) overwrite AMIs
+        for key in self.ami_id_params:
+            print(key)
```
fruch, line 1604: Remember to remove debug prints

sdcm/sct_config.py
```diff
@@ -1606,17 +1599,13 @@ def __init__(self):
         add_severity_limit_rules(self.get("max_events_severities"))
         print_critical_events()
 
-        # 5) assume multi dc by n_db_nodes set size
-        if 'aws' in cluster_backend:
```
fruch, line 1610: I thought this is a check

Now I see this is some kind of old hack to get data out of region_data even that we already did it...

sct.py
```diff
@@ -581,8 +581,11 @@ def list_resources(ctx, user, test_id, get_all, get_all_running, verbose):
 def list_images(cloud_provider: str, branch: str, version: str, region: str, arch: AwsArchType):
     add_file_logger()
     branch_fields = ["Backend", "Name", "ImageId", "CreationDate", "BuildId", "Arch", "ScyllaVersion"]
+    branch_fields_with_name_tag = [
```
fruch, line 584: Why adding a new list ? And extending the current list ?

sdcm/utils/common.py
```diff
@@ -1796,14 +1796,60 @@ def get_ami_images(branch: str, region: str, arch: AwsArchType) -> list:
...
+
+
+def convert_name_to_ami_if_needed(ami_id_param: str, region_names: list[str],) -> str:
+    """
+    convert image name in ami_id param to ami_ids
+    Firstly trying to find name in 'tag:Name' - for ScyllaDB images case
+    then in 'name' - for ScyllaDB images case for public images like Ubuntu
+
+    usage examlpe
+    convert_name_to_ami_if_needed('scylla-5.1.4-x86_64-2023-01-22T17-58-39Z' ,
```
fruch, line 1820: Nitpick

The examples are not passing regions as list, but as string...

unit_tests/test_config.py
```diff
@@ -679,6 +679,42 @@ def test_22_get_none():
         val = conf.get(None)
         assert val is None
 
+    def test_23_convert_ami_multi_region(self):
+        # test converting with all supported regions
+        os.environ['SCT_CLUSTER_BACKEND'] = 'aws'
+        os.environ['SCT_AMI_ID_DB_SCYLLA'] = 'scylla-5.1.4-x86_64-2023-01-22T17-58-39Z'
```
fruch, line 685: Let's test with the loader AMIs

The specific scylla version would be gone in a year or two...

Also let's mark those tests as integration tests

fruch (commented):
Looks good, few comments. and it's good to go

sct.py
fruch: o.k.

fruch (approved):
LGTM

sct.py
```diff
@@ -581,8 +581,11 @@ def list_resources(ctx, user, test_id, get_all, get_all_running, verbose):
...
+        "Backend", "Name", "ImageId", "NameTag", "CreationDate", "BuildId", "Arch", "ScyllaVersion"]
     #  TODO: align branch and version fields once scylla-pkg#2995 is resolved
     version_fields = ["Backend", "Name", "ImageId", "CreationDate"]
+    version_fields_with_name_tag = ["Backend", "Name", "ImageId", "NameTag", "CreationDate"]
```
vponomaryov, line 588: I would expect something like following for the current case:
```
version_fields = ["Backend", "Name", "ImageId", "CreationDate"]
version_fields_with_name_tag = version_fields + ["NameTag"]
```

The idea is that we should avoid duplicating the literals definitions.
Using proposed variant we will change a tag name just in one place when we need to. 
Using the current approach from the PR it requires change of 2 places among longer strings. It is more error prone coding for future updates.

Moreover, definition of additional lists is not really needed. It may be used just in place:

```
version_fields = ["Backend", "Name", "ImageId", "CreationDate"]
...
# aws
create_pretty_table(rows=rows, field_names=version_fields + ["NameTag"]).get_string(...)
...
# gce
create_pretty_table(rows=rows, field_names=version_fields).get_string(...)
...
```

So, please, reuse literals.

sdcm/utils/common.py
```diff
@@ -1796,14 +1796,60 @@ def get_ami_images(branch: str, region: str, arch: AwsArchType) -> list:
...
+                                   'eu-west-1', 'eu-west-2', 'us-west-2', 'us-east-1', 'eu-north-1', 'eu-central-1'])
+    returns: 'ami-026e72e4e468afa7b ami-01b8d743224353ffe ami-03f8756d29f0b5f21
+              ami-06878d265978313ca ami-03df6dea56f8aa618 ami-0039da1f3917fa8e3'
+    """
+    param_values = ami_id_param.split()
+    if len(param_values) == 1 and not param_values[0].startswith("ami-"):
+        ami_list: list[str] = []
+        for region_name in region_names:
+            ec2_resource = boto3.resource('ec2', region_name=region_name)
+            search_filter = [{'Name': 'tag:Name', 'Values': [param_values[0],]}]
```
vponomaryov, line 1835: This block of code (`lines 1834-1846`) has duplication. Better to remove it.
Example:

```
ec2_resource = boto3.resource('ec2', region_name=region_name)
for tag_name in ("tag:Name", "name"):
    if images := ec2_resource.images.filter(Filters=[{'Name': tag_name, 'Values': [param_values[0]]}])
        ami_list.append(sorted(images, key=lambda x: x.creation_date, reverse=True)[0].image_id)
        break
else:
    raise ValueError(
        f"Can't convert name '{ami_id_param}' to AMI_id, no image found in region {region_name}")
```

sdcm/utils/common.py
```diff
@@ -1796,14 +1796,60 @@ def get_ami_images(branch: str, region: str, arch: AwsArchType) -> list:
...
+    returns: 'ami-042cf1bc21e30ce60 ami-0370fbb289d8310d2 ami-09808043b7fcdb244
+              ami-00c1a16b5136fbb7c ami-0226185e7d7951b6a ami-0bc7811c417b0eb09'
+
+     convert_name_to_ami_if_needed('ubuntu/images/hvm-ssd/ubuntu-jammy-22.04-amd64-server-20221206' ,
+                                   'eu-west-1', 'eu-west-2', 'us-west-2', 'us-east-1', 'eu-north-1', 'eu-central-1'])
+    returns: 'ami-026e72e4e468afa7b ami-01b8d743224353ffe ami-03f8756d29f0b5f21
+              ami-06878d265978313ca ami-03df6dea56f8aa618 ami-0039da1f3917fa8e3'
+    """
+    param_values = ami_id_param.split()
+    if len(param_values) == 1 and not param_values[0].startswith("ami-"):
```
vponomaryov, line 1831: If we use a name like `ami-for-loader` then we will get incorrect behavior here.
Better to use regex here to detect the ami ID.

Then, it is not safe to assume that multiple values are only ami IDs. Need to verify it and look for images if we detect that `name` is provided instead of `ID`.

vponomaryov (changes requested):

sdcm/utils/common.py
fruch: I'm fine with not safe guarding from name prefix with `ami-`, we didn't checking guarding for that before either, it would fail when trying to create an instance with that, and that's fine.

as for supporting mix of names and id, I find it less relevant to the use cases where we need the names, i.e. loader/monitor images, the whole point is that we want to stop specifying AMI ids per region, and again there was no validation before, and it would fail on creation, and that's fine.

vponomaryov: @fruch 
It is not only about failures, it is also about flexibility. If we want to use not latest ID then we specify it explicitly.
For the moment it is not supported.

fruch: > @fruch It is not only about failures, it is also about flexibility. If we want to use not latest ID then we specify it explicitly. For the moment it is not supported.

can you elaborate on which case isn't supported now ? specifying ami id would still work, specifying it by name would now work, what's not going to work ?

vponomaryov: `'some-name-first-element ami-fooid` and other `multiple` combinations

fruch: I'm o.k. with define that anything that start with `ami-` we treat as ami id. 
complicating this for this case isn't a waste of time from my POV.

I think a note about it in the configuration option such suffice, if one would run into this he shoule be able to change the name quite easily. and I don't think any formal AMI of anything would ever start like that...

vponomaryov: Ok, if it is intentional, let it be so.

sct.py
```diff
@@ -108,6 +108,13 @@
 
 LOGGER = setup_stdout_logger()
 
+# #################### constants for hydra list-images output ####################
+VERSION_FIELDS = ["Backend", "Name", "ImageId", "CreationDate"]
+VERSION_FIELDS_WITH_TAG_NAME = VERSION_FIELDS + ["NameTag"]
+#  TODO: align branch and version fields once scylla-pkg#2995 is resolved
+BRANCH_SPECIFIC_FIELDS = ["BuildId", "Arch", "ScyllaVersion"]
+BRANCH_FIELDS = VERSION_FIELDS + BRANCH_SPECIFIC_FIELDS
+BRANCH_FIELDS_WITH_TAG_NAME = VERSION_FIELDS_WITH_TAG_NAME + BRANCH_SPECIFIC_FIELDS
```
vponomaryov, line 117: This is used in one specific function.
Why it was decided to place it here?
I really fail to see any gain in such a move.

fruch (approved):

vponomaryov (approved):
LGTM

PR #5743, opened 2023-01-31
Add k8ssetmonkey nemesis

vponomaryov:


vponomaryov:
Must be used in conjunction with the following PR if multitenancy is enabled: https://github.com/scylladb/scylla-cluster-tests/pull/5744

fruch:
@vponomaryov lets also add a test case that uses 2 tenants and this nemesis, as we listed on the Test Plan (after that we should do the same with 14, but there we can mix it, and do once cluster with it, and the other clusters with the FreeTier)

fruch (approved):
LGTM

vponomaryov:
> @vponomaryov lets also add a test case that uses 2 tenants and this nemesis, as we listed on the Test Plan (after that we should do the same with 14, but there we can mix it, and do once cluster with it, and the other clusters with the FreeTier)

https://github.com/scylladb/scylla-cluster-tests/pull/5746

soyacz (approved):

PR #5741, opened 2023-01-31
Fix redundant cluster rollout after nodetool_flush_and_reshard test

soyacz:
In `test_nodetool_flush_and_reshard` test cluster configuration is
recovered to original state (cpu count change). But, it looks it is not
recovering to original cluster state because cpu limits value gets wrong
type - instead string, float is supplied.

This fix recovers proper cpu limit value. So there's no need to
additionally rollout cluster after the test.

Also added deepdiff lib as:
Sometimes it would be good to see differences in dictionaries.
E.g. when we want to know what spec parameter of k8s cluster changed
between tests so we decide to rollout it.

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5747, opened 2023-01-31
fix(hydra): fix tab completion

fruch:
for some time the tab completion we were using wasn't working removed the package that we were using, and switch to using click core tab completion support.

also upgrade fabric/paramiko to get rid of the annying warning, which was part of the breakage of completion as well:
```
CryptographyDeprecationWarning: Blowfish has been deprecated
```

currently we only support bash (even that click supports also zsh and fish, and one can manual setup them if needed)

### DIFF
--- a/docker/env/version
+++ b/docker/env/version
@@ -1,1 +1,1 @@
-1.34-deepdiff
+1.35-fix_completion
--- a/requirements.in
+++ b/requirements.in
@@ -1,7 +1,7 @@
 PyYAML==5.4.1
-paramiko==2.10.1
-fabric==2.6.0
-invoke==1.6.0
+paramiko==3.0.0
+fabric==3.0.0
+invoke==2.0.0
 boto3==1.26.49
 boto3-stubs[s3,ec2,dynamodb,pricing]==1.26.49
 awscli==1.27.49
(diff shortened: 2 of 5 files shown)


fgelcer:
typo in the commit message `hydar` instead of `hydra`

soyacz:
did you test with `ssh_transport: fabric`? 

Also have in mind I also try to publish new hydra version https://github.com/scylladb/scylla-cluster-tests/pull/5741 - I think mine is more safe (just adding deepdiff pkg) and should be merged before this one.

fruch:
> did you test with `ssh_transport: fabric`?

now I did, even that no test case is using it

> 
> Also have in mind I also try to publish new hydra version #5741 - I think mine is more safe (just adding deepdiff pkg) and should be merged before this one.

I know, both are harmless.
first one get reviews would win :)

vponomaryov (dismissed):
LGTM

soyacz:
> > did you test with `ssh_transport: fabric`?
> 
> now I did, even that no test case is using it
> 
> > Also have in mind I also try to publish new hydra version #5741 - I think mine is more safe (just adding deepdiff pkg) and should be merged before this one.
> 
> I know, both are harmless. first one get reviews would win :)

I won :grin:

roydahan:
Hallelujah for this PR.

soyacz (dismissed):

vponomaryov (approved):
LGTM

PR #5749, opened 2023-02-01
New performance index

aleksbykov:
Performance tests require new branch based on latest master. such branch should become branch-perf-v14
For that there is a decision to create new index for performance which will store new results including the hdr histograms data 
and new performance results for latest version.

for that new index will be created using  new sct subcommand create-es-index which allow to set up appropriate mappings for field. "test_details.start_time" will now have date type and will allow to build valid dashboards in kibana

Also hdr histogram data will be stored in es documents.

### DIFF
--- a/docs/hydra-create-es-index.md
+++ b/docs/hydra-create-es-index.md
@@ -1,0 +1,36 @@
+# Create ES index
+
+ES index stores all documents. ES document could have fields with different data type. The mapping of document properties and field types described in es mapping settings. ES could detect the field type dynamically and generate mapping when first document is written to ES index. After mapping is added to index, it could be changed.
+
+Detected type for document properties is not always correct.
+
+Example, Performance tests use test_details.start_time field to store date when test was start executed. But data type was set in mapping as str/long for different indexes. This doesn't allow to build Kibana dashboards based on start_time.
+
+To avoid this issues, ES Index could be create with specified mappings schema before any document added to the index. Any addition settings and formants could be configured for certain field in future documents.
+
+Sct creates ES Index using the Class name of maing test,
+
+```python
+class PerformanceTest(ClusterTester):
+	pass
+
+class LongevityTest(ClusterTester):
+	pass
+```
+
+Once test is configured to store results and is run, SCT create new index, if it is not yet exists, with name `performancetest` or `longevitytest`, but field mapping will be done dynamically and by fields of first inserted document.
+
+To be sure that certain document property has required data type, index could be created with mapping settings provided in file with new sct commands
+
+# SCT commands
+
+1. Create file with mappings with any editor you prefer.
+2. Run sct to create new index in ES:
+```bash
+# hydra create-es-index --name <index_name> --doc-type <doc_type_in_index> --filepath <full_path_to_file_with_mapping>
+hydra create-es-index --name newtestindex --doc-type test_stats --filepath "configration/es_mappings/newmapping.json"
+
+```
+3. sct test automatically will write new document to new index if its class name will `NewTestIndex
+-or-
+4. you can use ES client to write document to index `newtestindex`
--- a/performance_regression_test.py
+++ b/performance_regression_test.py
@@ -17,6 +17,8 @@
 import os
 import time
 
+from enum import Enum
+
 import yaml
 
 from sdcm.tester import ClusterTester, teardown_on_exception
@@ -25,8 +27,20 @@ from sdcm.sct_events.filters import EventsSeverityChangerFilter
 from sdcm.sct_events.loaders import CassandraStressEvent
 from sdcm.sct_events.system import HWPerforanceEvent
 from sdcm.utils.decorators import log_run_info
+from sdcm.utils.csrangehistogram import CSHistogramTagTypes
 
 KB = 1024
+
+
+class PerformanceTestWorkload(Enum):
+    WRITE = "write"
+    READ = "read"
+    MIXED = "mixed"
+
+
+class PerformanceTestType(Enum):
+    THROUGHPUT = "throughput"
+    LATENCY = "latency"
 
 
 class PerformanceRegressionTest(ClusterTester):  # pylint: disable=too-many-public-methods
@@ -253,6 +267,8 @@ class PerformanceRegressionTest(ClusterTester):  # pylint: disable=too-many-publ
             self.db_cluster.add_nemesis(nemesis=self.get_nemesis_class(), tester_obj=self)
             self.db_cluster.start_nemesis(interval=interval)
         results = self.get_stress_results(queue=stress_queue)
+
+        self.build_histogram(PerformanceTestWorkload.READ, PerformanceTestType.LATENCY)
         self.update_test_details()
         self.display_results(results, test_name='test_latency' if not nemesis else 'test_latency_with_nemesis')
         self.check_regression()
@@ -267,6 +283,7 @@ class PerformanceRegressionTest(ClusterTester):  # pylint: disable=too-many-publ
             self.db_cluster.add_nemesis(nemesis=self.get_nemesis_class(), tester_obj=self)
             self.db_cluster.start_nemesis(interval=self.params.get('nemesis_interval'))
         results = self.get_stress_results(queue=stress_queue)
+        self.build_histogram(PerformanceTestWorkload.WRITE, PerformanceTestType.LATENCY)
         self.update_test_details()
         self.display_results(results, test_name='test_latency')
         self.check_regression()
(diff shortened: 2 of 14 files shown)


test-cases/performance/perf_staging_index_throughput.yaml
```diff
@@ -0,0 +1,29 @@
+test_duration: 300
```
roydahan, line 1: Is this one just for your testing?

roydahan:
@fruch please review.
@temichus you can review and get yourself familiar with it.

sdcm/utils/es_index.py
```diff
@@ -0,0 +1,38 @@
...
+    return status
+
+
+def get_mapping(mapping_filepath: str) -> Any:
+    filepath = Path(mapping_filepath)
+    if not filepath.exists():
+        LOGGER.error("Provided mapping file %s does not exists", mapping_filepath)
+        return None
+    with open(filepath, "r", encoding="utf-8") as mapping_fp:
+        mapping = json.load(mapping_fp)
```
fruch, line 37: you can use `filepath.read_text()`, and all of the checks before aren't needed

sdcm/utils/es_index.py
```diff
@@ -0,0 +1,38 @@
...
+        LOGGER.error("Create index failed with error %s", exc)
+
+    return status
+
+
+def get_mapping(mapping_filepath: str) -> Any:
+    filepath = Path(mapping_filepath)
+    if not filepath.exists():
+        LOGGER.error("Provided mapping file %s does not exists", mapping_filepath)
+        return None
```
fruch, line 35: why return None ? and not raise exception ?

sdcm/utils/es_index.py
```diff
@@ -0,0 +1,38 @@
...
+    es_client = ES()
+    doc_type = kwargs.get("doc_type")
+    status = False
+    try:
+        es_client.indices.create(index=index_name)
+        es_client.indices.put_mapping(index=index_name, doc_type=doc_type, body=mappings,  # pylint: disable=unexpected-keyword-arg
+                                      include_type_name=True)
+        status = True
+    except Exception as exc:  # pylint: disable=broad-except
+        LOGGER.error("Create index failed with error %s", exc)
```
fruch, line 26: why ignore the errors, and not raising them up ?

performance_regression_test.py
```diff
@@ -252,6 +266,8 @@ def run_read_workload(self, nemesis=False):
             self.db_cluster.add_nemesis(nemesis=self.get_nemesis_class(), tester_obj=self)
             self.db_cluster.start_nemesis(interval=interval)
         results = self.get_stress_results(queue=stress_queue)
+
+        self.build_histogram(PerformanceTestWorkload.READ, PerformanceTestType.LATENCY)
```
vponomaryov, line 270: Shouldn't it be used only when `use_hdr_cs_histogram` is enabled?
The same for all other usages of HDR feature.

sct.py
```diff
@@ -1594,6 +1596,25 @@ def generate_parallel_timelines_report(logdir: str | None, test_id: str | None)
...
+def create_es_index(name: str, doc_type: str, mapping_file: str) -> None:
+    add_file_logger()
+
+    mapping_data = get_mapping(mapping_file)
+    if mapping_data is None:
+        click.secho(f"File {mapping_file} with mapping  not found")
+        sys.exit(1)
+    status = create_index(index_name=name, doc_type=doc_type, mappings=mapping_data)
+    if not status:
+        click.secho(f"Index {name} was not created")
```
fruch, line 1613: I'm not sure reason of the failure would be shown like this, just raise the exception, and whom is using it would need to handle it.

sct.py
```diff
@@ -1594,6 +1596,25 @@ def generate_parallel_timelines_report(logdir: str | None, test_id: str | None)
     pt_report_generator.generate_full_report()
 
 
+@cli.command("create-es-index", help="Create ElasticSearch index with mapping ")
+@click.option("-n", "--name", envvar='SCT_ES_INDEX_NAME', help="ES index name")
+@click.option("-dt", "--doc-type", envvar='SCT_ES_DOC_TYPE', default="")
+@click.option("-f", "--mapping-file", envvar='SCT_MAPPING_FILEPATH', type=click.Path(exists=True),
+              help="Full path to es index mapping file")
+def create_es_index(name: str, doc_type: str, mapping_file: str) -> None:
```
fruch, line 1604: Can you elaborate somewhere like in `/docs` about the process in which someone would use this command ?

sdcm/db_stats.py
```diff
@@ -619,7 +619,7 @@ def get_test_details(self):
         test_details['job_url'] = get_job_url()
         test_details['start_host'] = platform.node()
         test_details['test_duration'] = self.params.get(key='test_duration')
-        test_details['start_time'] = time.time()
+        test_details['start_time'] = int(time.time())
```
fruch, line 622: if this is going to be use as the timestamp of the data, why not use `datetime.utcnow()` ?

performance_regression_test.py
```diff
@@ -40,6 +40,7 @@ class PerformanceRegressionTest(ClusterTester):  # pylint: disable=too-many-publ
     start_ops = 10_000
     throttle_step = 10_000
     max_ops = 200_000
+    base_index = "stagingperftest"
```
fruch, line 57: you might want to have some degree of control here, where would not staging results would go  ?

fruch:
@aleksbykov have you build any dashboard ontop of this new index results ?

test-cases/performance/perf_staging_index_throughput.yaml
aleksbykov: yes, it will be reverted

sct.py
aleksbykov: Added docs/hydra-create-es-index.md

sdcm/utils/es_index.py
aleksbykov: fixed

sdcm/utils/es_index.py
aleksbykov: fixed

sdcm/db_stats.py
aleksbykov: in many places where calculating timeintervals and in regression stats calculate used time.time() for end timestamp. To avoid time conversion in different places, i set such way
It could be done in another pr
or maybe next constructiion could be used:
`int(time.mktime(datetime.datetime.utcnow().utctimetuple()))`

sct.py
aleksbykov: removed

sdcm/utils/es_index.py
aleksbykov: fixed

performance_regression_test.py
aleksbykov: Thanks, replace this logic to minize affect on other tests. Added new config parameter to setup specific es_index

aleksbykov:
> @aleksbykov have you build any dashboard ontop of this new index results ?

@fruch , yes, with @roydahan 's help. but i think they are not working now, because i kill most of staging indexes.

performance_regression_test.py
aleksbykov: Added check that flag is enabled

aleksbykov:
@roydahan @fruch @vponomaryov could you take a look one more time. Staging job is running to be sure that nothing is broken

roydahan:
I would like to try merging this today and creating branch-v14.

sdcm/db_stats.py
fruch: O.k.

fruch (approved):

docs/hydra-create-es-index.md
```diff
@@ -0,0 +1,36 @@
...
+1. Create file with mappings with any editor you prefer.
+2. Run sct to create new index in ES:
+```bash
+# hydra create-es-index --name <index_name> --doc-type <doc_type_in_index> --filepath <full_path_to_file_with_mapping>
+hydra create-es-index --name newtestindex --doc-type test_stats --filepath "configration/es_mappings/newmapping.json"
+
+```
+3. sct test automatically will write new document to new index if its class name will `NewTestIndex
+-or-
+4. you can use ES client to write document to index `newtestindex`
```
fgelcer, line 36: this part is not clear to me... but in any case, you have here once `NewTestIndex` with capitalized name, and once `newtestindex` like you write in the command in the line above

fruch:
> @fruch @roydahan @aleksbykov 
> 
> I worry about future modifications. 
> What if we decided that we need to set up additional appropriate mappings for existing fields like "test_details.start_time"
> what will we do in this case? if just creating a new Index, I think it makes sense to take more time for this task and make mapping for all important fields.  
>  
> 

Other fields are strings or numbers, ES automatically creates mapping for them.

Changing a type of a mapping might be problematic and would force us to create a new index, which is part of the reason why we are moving away from index's based on the class names.

Once we'll want to do such migration of an Index, we'll need to work on how (assuming we want to keep the data)

fgelcer:
> > @fruch @roydahan @aleksbykov
> > I worry about future modifications.
> > What if we decided that we need to set up additional appropriate mappings for existing fields like "test_details.start_time"
> > what will we do in this case? if just creating a new Index, I think it makes sense to take more time for this task and make mapping for all important fields.
> 
> Other fields are strings or numbers, ES automatically creates mapping for them.

in the past, i had latency results reported as strings (probably an issue when first sending the results to be saved on ES the 1st time), but sometimes it was number, making it impossible to assume and compare results

> 
> Changing a type of a mapping might be problematic and would force us to create a new index, which is part of the reason why we are moving away from index's based on the class names.
> 
> Once we'll want to do such migration of an Index, we'll need to work on how (assuming we want to keep the data)

roydahan (approved):

PR #5752, opened 2023-02-01
fix(nodetool_status): assert rack is specified

soyacz:
In case rack is not specified, drivers cannot work properly and show warnings. In azure when we don't specify availability zone, then it's not provided by metadata api. Recently support was provided to return region (location) as rack when AZ is not specified.

This fix forces `get_nodetool_status` to guard rack is always provided by `nodetool status` revealing issue of not specified rack.
refs: https://github.com/scylladb/scylladb/issues/12185

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4057,6 +4057,7 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
                             node_ip = node.ip_address
                 node_info["load"] = node_info["load"].replace(" ", "")
                 status[dc_name][node_ip] = node_info
+                assert node_info["rack"], "Rack is not defined for node {}".format(node_info)
         return status
 
     @staticmethod
--- a/unit_tests/test_cluster.py
+++ b/unit_tests/test_cluster.py
@@ -501,28 +501,9 @@ class TestNodetoolStatus(unittest.TestCase):
         node = NodetoolDummyNode(resp=resp)
         db_cluster = DummyScyllaCluster([node])
 
-        status = db_cluster.get_nodetool_status()
-
-        assert status == {'eastus': {'10.0.0.4': {'host_id': 'ed6af9a0-8c22-4813-ac9b-6fbeb462b687',
-                                                  'load': '431KB',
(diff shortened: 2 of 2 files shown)


vponomaryov:
Is it correct to say that having this fix we may need to update some of our configurations among hundreds we have?

soyacz:
> Is it correct to say that having this fix we may need to update some of our configurations among hundreds we have?

Why?
I don't anticipate any changes to configurations. We must be sure not to backport this one where we don't have https://github.com/scylladb/scylladb/pull/12274

fruch:
> > Is it correct to say that having this fix we may need to update some of our configurations among hundreds we have?
> 
> Why? I don't anticipate any changes to configurations. We must be sure not to backport this one where we don't have [scylladb/scylladb#12274](https://github.com/scylladb/scylladb/pull/12274)

I think what @vponomaryov suggests, that it's hard to say if this assert is correct across all the setups we create in SCT, 
i.e. we'll need to test this in GCP/Azure/AWS to verify this assertion is correct

vponomaryov:
Yes, we have too many different combinations to be sure everywhere everything will be ok.
But also following is not really clear to me:

If it depends on a pretty recent fix then we will get endless assertion errors on some stable scylla versions like `5.0.5`?
Right? Or am I missing something here?

soyacz:
> Yes, we have too many different combinations to be sure everywhere everything will be ok. But also following is not really clear to me:
> 
> If it depends on a pretty recent fix then we will get endless assertion errors on some stable scylla versions like `5.0.5`? Right? Or am I missing something here?

in case of upgrade test for Azure - yes - if we enable such then it will be failing.
Otherwise, rack is always (and always been) set - so this assert should pass.

This PR is proposal for catching errors with racks - but we lived without it since always so we can abandon this PR and live like we lived. I don't anticipate regressions in this area and can be actually easy to spot when looking into logs (just people must show enough curiosity about warnings in logs).
I'm also open to any other suggestions how to prevent issues like https://github.com/scylladb/scylladb/issues/12185

fruch:
@soyacz so agreed this is too risky ?

PR #5755, opened 2023-02-01
fix(Unit tests): fix EventsUtilsMixin teardown

temichus:
ignore errors like FileNotFoundError on EventsUtilsMixin teardown in shutil.rmtree 
not  best solution but fix the current error at the end of unit tests

### DIFF
--- a/unit_tests/conftest.py
+++ b/unit_tests/conftest.py
@@ -36,9 +36,11 @@ from unit_tests.lib.fake_region_definition_builder import FakeDefinitionBuilder
 from unit_tests.lib.fake_remoter import FakeRemoter
 
 
-@pytest.fixture(scope='session')
+@pytest.fixture(scope='module')
 def events():
-    mixing = EventsUtilsMixin()
+    class LocalMixing(EventsUtilsMixin):
+        pass
+    mixing = LocalMixing()
     mixing.setup_events_processes(events_device=True, events_main_device=False, registry_patcher=True)
     yield mixing
 
--- a/unit_tests/test_scan_operation_thread.py
+++ b/unit_tests/test_scan_operation_thread.py
@@ -13,7 +13,6 @@ from unittest.mock import MagicMock, patch
 import pytest
 from cassandra import OperationTimedOut
 from unit_tests.test_cluster import DummyDbCluster, DummyNode
-from unit_tests.lib.events_utils import EventsUtilsMixin
 from sdcm.scan_operation_thread import ScanOperationThread, FullScanParams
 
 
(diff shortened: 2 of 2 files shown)


fruch:
is this part of you looking into failure like this:
https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-5753/3/execution/node/36/log/

fruch:
> > is this part of you looking into failure like this: https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-5753/3/execution/node/36/log/
> 
> this fixed here [ec7bf68](https://github.com/scylladb/scylla-cluster-tests/commit/ec7bf68a476721fab1ae18f548bb4e3b3bf81ac5)

You slipped that under the radar, I didn't noticed it...
next it it would be better to put such a thing on a PR of it's own.

I look at the failing tests, and it looked like some instability, that isn't related to the PR, so I decided to merge.

fruch:
Anyhow about this fix, I'm not sure I'm following how we are getting into this situation, since we are having multiple teardowns clearing the same directory ?, since a new event mixing is being used in the module level ?

if that's the case this change is just masking the fact the directory was cleared, I have a feeling the introduction of a module level fixture for event wasn't needed.

anyhow if we thing it should be on a module level, we can just change the existing one to module level, but also making sure each one would be working on it's own directory

fruch:
> > Anyhow about this fix, I'm not sure I'm following how we are getting into this situation, since we are having multiple teardowns clearing the same directory ?, since a new event mixing is being used in the module level ?
> > if that's the case this change is just masking the fact the directory was cleared, I have a feeling the introduction of a module level fixture for event wasn't needed.
> > anyhow if we thing it should be on a module level, we can just change the existing one to module level, but also making sure each one would be working on it's own directory
> 
> not sure how but this tests [391d6c5#diff-02d17379611f0b69d65077cf8ddf154d40106e8c9e76d371b450b8deb0a59576](https://github.com/scylladb/scylla-cluster-tests/commit/391d6c52b2113df8ab59b80cef8e97617d17e29b#diff-02d17379611f0b69d65077cf8ddf154d40106e8c9e76d371b450b8deb0a59576) that run before scan operation test somehow drop evens and events did not come even after 10 sec (on local machine) so i decided to switch to module one
> 
> 2nd issue that cls.temp_dir on class level. so that means that EventsUtilsMixin can not be used twice.
> 
> Tomorrow, can move global events on the module level and check if it works

Seems like this would be safe enough if you still think you need it at the module scope

```diff
diff --git a/unit_tests/conftest.py b/unit_tests/conftest.py
index c207cd339..63a212799 100644
--- a/unit_tests/conftest.py
+++ b/unit_tests/conftest.py
@@ -36,9 +36,11 @@ from unit_tests.lib.fake_region_definition_builder import FakeDefinitionBuilder
 from unit_tests.lib.fake_remoter import FakeRemoter


-@pytest.fixture(scope='session')
+@pytest.fixture(scope='module')
 def events():
-    mixing = EventsUtilsMixin()
+    class LocalMixing(EventsUtilsMixin):
+        pass
+    mixing = LocalMixing()
     mixing.setup_events_processes(events_device=True, events_main_device=False, registry_patcher=True)
     yield mixing
```

fruch (approved):
LGTM

Issue #5751, opened 2023-02-01
Division by zero error in benchmark code when running on Azure/GCP


soyacz:
```
< t:2023-01-30 06:56:43,791 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:ERROR > exception=division by zero
< t:2023-01-30 06:56:43,793 f:grafana.py      l:80   c:sdcm.sct_events.grafana p:DEBUG > GrafanaEventAggregator start a new time window (90 sec)
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR > Exception in setUp. Will call tearDown < t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR > Exception in setUp. Will call tearDown
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR > Traceback (most recent call last):
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR >   File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 153, in wrapper
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR >     return method(*args, **kwargs)
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR >   File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 116, in inner
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR >     res = func(*args, **kwargs)
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR >   File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 741, in setUp
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR >     self.init_resources()
< t:2023-01-30 06:56:43,789 f:tester.py       l:160  c:sdcm.tester          p:ERROR >   File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 1720, in init_resources
[...]

PR #5764, opened 2023-02-02
fix(node_benchmark): correctly count average if metric is missing

aleksbykov:
Fixes: #5751. If any node performance tool doesn't correclty store result in es document, next counting Average() value could lead to exception DivisionByZero, because final list is empty for failed metrics.

use list with single element [0.0] to avoid such collisions.

### DIFF
--- a/sdcm/utils/benchmarks.py
+++ b/sdcm/utils/benchmarks.py
@@ -240,9 +240,9 @@ class ScyllaClusterBenchmarkManager(metaclass=Singleton):
                 LOGGER.warning(
                     "Failed to generate comparable result for the following item:\n%s"
                     "\nException:%s", item, exc)
-        eps = [item.sysbench_eps for item in results]
-        read_bw = [item.cassandra_fio_read_bw for item in results]
-        write_bw = [item.cassandra_fio_write_bw for item in results]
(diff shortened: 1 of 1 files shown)


fgelcer (dismissed):
LGTM, but it would be nice if instead of writing `Fix for issue #5751` you would use `Fixes: #5751`, so once this PR will be merged, it will close the issue automatically

soyacz (approved):
this will prevent from getting DivisionByZero exceptions but will show broken values.
Anyway we should fix other issues related this code (like fixing the reason why we don't get these values).
Besides, why `0.0` and not `float('nan')`?

aleksbykov:
> Besides, why `0.0` and not `float('nan')`
in this case sum() will fail and class ComparableResult use 0.0

roydahan (approved):

roydahan:
Please see why checks are failing and we can merge.

fgelcer:
> Please see why checks are failing and we can merge.

something terminated the instance running the checks:
```
pylint...................................................................Cannot contact i-07159072a45c149e2: java.lang.InterruptedException
/bin/bash: line 1:    10 Killed                  ./sct.py pre-commit
```

i have re-run it

PR #5766, opened 2023-02-05
fix(pipelines): enlarge timeout for get duration stage

fruch:
seems like 2min isn't enough cause of recent changes to use a weaker meachines for jenkins builders, and addition of more AWS api calls during configuration read (to translate name to ami ids)

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5771, opened 2023-02-06
fix(perfrefressionpipeline): Enlarge timeout for stages

aleksbykov:
New performance jobs marked as failed because stages GetDuration and Create SCT runner exceed default time.

ex: https://jenkins.scylladb.com/view/master/job/scylla-master/job/scylla-master-perf-regression-throughput-shard-aware-test/12
Increase duration for stages

### DIFF
--- a/vars/perfRegressionParallelPipeline.groovy
+++ b/vars/perfRegressionParallelPipeline.groovy
@@ -107,7 +107,7 @@ def call(Map pipelineParams) {
                 }
                 steps {
                     catchError(stageResult: 'FAILURE') {
-                        timeout(time: 5, unit: 'MINUTES') {
+                        timeout(time: 10, unit: 'MINUTES') {
                             script {
                                 wrap([$class: 'BuildUser']) {
                                     dir('scylla-cluster-tests') {
@@ -161,7 +161,7 @@ def call(Map pipelineParams) {
                                             catchError(stageResult: 'FAILURE') {
                                                 wrap([$class: 'BuildUser']) {
                                                     dir('scylla-cluster-tests') {
-                                                        timeout(time: 5, unit: 'MINUTES') {
+                                                        timeout(time: 10, unit: 'MINUTES') {
                                                             createSctRunner(params, runnerTimeout, builder.region)
                                                         }
                                                     }


fgelcer:
it was recently increased in multiple jobs to 5 min (what we had in this pipeline originally), and IMHO this won't solve all the problems we have with it:"
https://github.com/scylladb/scylla-cluster-tests/pull/5766

we need to make this function work much smarted, so it will finish within few seconds

fruch:
> it was recently increased in multiple jobs to 5 min (what we had in this pipeline originally), and IMHO this won't solve all the problems we have with it:" #5766
> 
> we need to make this function work much smarted, so it will finish within few seconds

it can't finish in few seconds, it's the first command pulling the hydra image

fgelcer:
> it can't finish in few seconds, it's the first command pulling the hydra image

not true, the 1st stage that should get the hydra image is during checkout...
in [this run](https://jenkins.scylladb.com/view/scylla-5.2/job/scylla-5.2/job/longevity/job/longevity-large-partition-200k-pks-4days-aws-i4i-test/), get test duration took 13 seconds

fruch:
> > it can't finish in few seconds, it's the first command pulling the hydra image
> 
> not true, the 1st stage that should get the hydra image is during checkout...

checkout of git, doesn't pull any docker images

> in [this run](https://jenkins.scylladb.com/view/scylla-5.2/job/scylla-5.2/job/longevity/job/longevity-large-partition-200k-pks-4days-aws-i4i-test/), get test duration took 13 seconds

and yes it can take 13 seconds, but once we have 4 jobs doing it on the same time (pulling the same image), it can slow down considerably. if it's works within 5min, we'lll leave it as is. if it won't we'll need to consider using 1 executor per builder

fgelcer (approved):
> > > it can't finish in few seconds, it's the first command pulling the hydra image
> > 
> > 
> > not true, the 1st stage that should get the hydra image is during checkout...
> 
> checkout of git, doesn't pull any docker images
> 
> > in [this run](https://jenkins.scylladb.com/view/scylla-5.2/job/scylla-5.2/job/longevity/job/longevity-large-partition-200k-pks-4days-aws-i4i-test/), get test duration took 13 seconds
> 
> and yes it can take 13 seconds, but once we have 4 jobs doing it on the same time (pulling the same image), it can slow down considerably. if it's works within 5min, we'lll leave it as is. if it won't we'll need to consider using 1 executor per builder

we can try it, although again, i'm not sure this is the right solution, but it could be a workaround for now

PR #5770, opened 2023-02-05
fix(artifacts offline): remove unecessary jobs

fgelcer:
in the function used to create release Jenkins jobs, we didn't make any filter between artifacts and the offline artifacts, as many of them don't fit the
offline artifacts tests concept, as images and
web install.
this PR is addressing that, excluding few regex
of job names used for the regular artifacts,
to not create them as artifacts offline.

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5772, opened 2023-02-06
fix(artifacts offline generator): re-introduce amazon2

fgelcer:
there was a function skipping few keywords when
creating the jobs for release in Jenkins.
happens that, it had some issues, and didn't really filter out some important ones that should not be
created, and removed some, that we could run without any problems (like `artifacts-amazon2`)

this PR enhances what #5770 already started,
and removed the bogus "filter".

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -1373,15 +1373,9 @@ def create_test_release_jobs(branch, username, password, sct_branch, sct_repo):
 
     server.create_directory(
         name='artifacts-offline-install', display_name='SCT Artifacts Offline Install Tests')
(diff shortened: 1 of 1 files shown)


fgelcer:
ran it in [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/create-release-jobs/6/) (see that, because the path where these jobs were supposed to be created didn't exist, we can only see in the log the list of jobs that were supposed to be created)

fruch (approved):
LGTM

roydahan (approved):

PR #5776, opened 2023-02-07
fix(csrangehistogram): Increase max value for hdr histogram

aleksbykov:
CS could write to hdr.log value for some intervals higher value, than was configured. Python HDR histogram lib doesn't support autorescale feature (like in java lib), which allow to rescale histogram during merging.
Increase max value for HDR Histogram which allow
to store from 1 nanosecnds up to 24 hours

### DIFF
--- a/sdcm/utils/csrangehistogram.py
+++ b/sdcm/utils/csrangehistogram.py
@@ -51,7 +51,7 @@ HistorgramSummary = make_dataclass("HistorgramSummary",
 
 class CSHistogram(HdrHistogram):
     LOWEST = 1
-    HIGHEST = 12 * 3600 * 1000 * 1000
+    HIGHEST = 24 * 3600_000_000_000
     SIGNIFICANT = 3
 
     def __init__(self, *args, **kwargs):


sdcm/utils/csrangehistogram.py
```diff
@@ -51,7 +51,7 @@ class HistorgramSummaryBase:
 
 class CSHistogram(HdrHistogram):
     LOWEST = 1
-    HIGHEST = 12 * 3600 * 1000 * 1000
+    HIGHEST = 24 * 3600_1000_1000_1000
```
fgelcer, line 54: although this is an integer, i'm not sure it is doing exactly what you expected:
```
>>> 3600_1000_1000_1000
3600100010001000
```

did you mean to use 
```suggestion
    HIGHEST = 24 * 3600_000_000_000
```

?

aleksbykov: fixed

fgelcer (approved):
LGTM

fgelcer (approved):
LGTM

roydahan (approved):

PR #5784, opened 2023-02-08
fix(data validator): rows were not deleted

juliayakovlev:
Fix `sdcm.utils.data_validator.LongevityDataValidator.validate_deleted_rows`. Handle case in validation of deleted rows when rows were not deleted. Print relevant message

### DIFF
--- a/sdcm/utils/data_validator.py
+++ b/sdcm/utils/data_validator.py
@@ -790,6 +790,11 @@ class LongevityDataValidator:
                 ).publish()
             else:
                 LOGGER.debug('Validation deleted rows finished successfully')
+        elif len(actual_result) == self.rows_before_deletion:
+            DataValidatorEvent.DeletedRowsValidator(
+                severity=Severity.WARNING,
+                message="Rows were not deleted. Maybe need to increase dataset for delete."
+            ).publish()
         else:
             LOGGER.warning('Deleted row were not found. May be issue #6181. '
                            'Actual dataset length: {}, Expected dataset length: {}'.format(len(actual_result),


sdcm/utils/data_validator.py
```diff
@@ -790,6 +790,8 @@ def validate_deleted_rows(self, session, during_nemesis=False):
                 ).publish()
             else:
                 LOGGER.debug('Validation deleted rows finished successfully')
+        elif len(actual_result) == self.rows_before_deletion:
```
vponomaryov, line 793: Probably may be useful to create a `warning event`?

vponomaryov (changes requested):
Technically, it is not `fix`. It is `improvement`.
So, need to update commit and PR topic.

fruch:
@juliayakovlev is there a context to this change ? i.e. it was happening in some case ? it was part of some issue investigation ?

sdcm/utils/data_validator.py
fruch: it's equivalent, no one would read any of them.

I'm guessing this is intend to whom has a deletion case, and assume he'll read the log/events to look it up

I would argue that the vaildator need some configuration, so the test case can set which validation to do, and which are not needed.

juliayakovlev:
> @juliayakovlev is there a context to this change ? i.e. it was happening in some case ? it was part of some issue investigation ?

@fruch 
Yes, I faced it when investigated a job run and understood that I did not cover this case. I got message "Deleted row were not found. May be issue #6181." and length of actual and expcted dataset was same. The issue [#6181](https://github.com/scylladb/scylladb/issues/6181) result is length of actual > length of expected.
So this message is not actual

sdcm/utils/data_validator.py
juliayakovlev: A debug message changed to event

fruch (approved):
LGTM

juliayakovlev:
> @juliayakovlev commit topic? s/fix/improvement/ ?

I wanted to do it as "improvement" first. But actually I missed this case (actual == expected). So I think it's "fix"

vponomaryov:
> > @juliayakovlev commit topic? s/fix/improvement/ ?
> 
> I wanted to do it as "improvement" first. But actually I missed this case (actual == expected). So I think it's "fix"

The result doesn't change something specifically. In one case we have a log message in another new event.
A test run that faces this part of code won't get a behavior changed as of now.
So, it is `improvement` which improves/eases possible debugging.

PR #5787, opened 2023-02-08
fix(jenkins pipeline): Increase timeout of the 'Get test duration' stage

ShlomiBalalis:
The stage times out often and its clear that 5 minutes is too short of a timeout. As such, I increased the timeout to 10 minutes across all of the pipelines.

### DIFF
--- a/vars/jepsenPipeline.groovy
+++ b/vars/jepsenPipeline.groovy
@@ -93,7 +93,7 @@ def call(Map pipelineParams) {
             stage('Get test duration') {
                 steps {
                     catchError(stageResult: 'FAILURE') {
-                        timeout(time: 5, unit: 'MINUTES') {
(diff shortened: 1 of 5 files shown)


ShlomiBalalis:
@fgelcer Are there any more branches that are vulnerable to this issue?

fgelcer (approved):
LGTM

fgelcer:
> @fgelcer Are there any more branches that are vulnerable to this issue?

AFAIK only master (and the ones were FF'ed from master, meaning 5.2 and 2023.1)
all other branches should be safe

fruch:
even that #5788 seems to be the root cause, I'm still merging this one, since now this stage would be doing much more pulling of the hydra image then it was doing before, and we should give it enough slack.

fruch (approved):
LGTM

PR #5790, opened 2023-02-08
fix(unittest): switch to newer amis

fruch:
some amis are old and were removed, we should
stop using them on the unittest

### DIFF
--- a/internal_test_data/complex_test_case_with_version.yaml
+++ b/internal_test_data/complex_test_case_with_version.yaml
@@ -43,7 +43,7 @@ gce_n_local_ssd_disk_monitor: 0
 
 db_type: scylla
 instance_type_db: 'i3.4xlarge'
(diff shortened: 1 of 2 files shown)


fgelcer (approved):
LGTM

soyacz:
why do we care about the specific ami? Wouldn't it serve the purpose if we just verified it starts with `ami-`?

fruch:
> > why do we care about the specific ami? Wouldn't it serve the purpose if we just verified it starts with `ami-`?
> 
> we have test there there check the logic of pulling specific version, for that we need a specific version. the verification of the exact id isn't the actual issue, the issue is that we don't find a version (not that amis doesn't match)

Anyhow I've move most of those test into integration test suite, so they are not mandatory for each PR anymore
i.e. they are not unittest once they communicate and depend on something external to the code, like that

PR #5792, opened 2023-02-09
fix(parse_scylla_version): better check the version

fruch:
it was failing like:
```
File ".../sdcm/fill_db_data.py", line 3173, in version_non_frozen_udt_support
return self.parsed_scylla_version >= parse_scylla_version(version_with_support)
File ".../sdcm/utils/version_utils.py", line 123, in parse_scylla_version
return parse_version(version_format.search(version_to_parse)[0])
TypeError: 'NoneType' object is not subscriptable
```

since the version passed to is was not matching the regex

```python
NON_FROZEN_SUPPORT_ENTERPRISE_MIN_VERSION = '2020'
```

now it checking it better and rasining clear exception why it was failing

### DIFF
--- a/sdcm/utils/version_utils.py
+++ b/sdcm/utils/version_utils.py
@@ -120,7 +120,9 @@ RepositoryDetails = namedtuple("RepositoryDetails", ["type", "urls"])
 
 def parse_scylla_version(version_to_parse: str) -> version.Version:
     version_format = re.compile(r'(\d+\.\d)|(\.\d+)')
-    return parse_version(version_format.search(version_to_parse)[0])
+    major_version = version_format.search(version_to_parse)
+    assert major_version, f"version_to_parse: '{version_to_parse}' isn't supported scylla version string"
+    return parse_version(major_version[0])
 
 
 @lru_cache(maxsize=1024)
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -45,7 +45,7 @@ class FillDatabaseData(ClusterTester):
 
     """
     NON_FROZEN_SUPPORT_OS_MIN_VERSION = '4.1'  # open source version with non-frozen user_types support
-    NON_FROZEN_SUPPORT_ENTERPRISE_MIN_VERSION = '2020'  # enterprise version with non-frozen user_types support
+    NON_FROZEN_SUPPORT_ENTERPRISE_MIN_VERSION = '2020.1'  # enterprise version with non-frozen user_types support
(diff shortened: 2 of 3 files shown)


sdcm/utils/version_utils.py
```diff
@@ -120,7 +120,9 @@ class ScyllaFileType(Enum):
 
 def parse_scylla_version(version_to_parse: str) -> version.Version:
     version_format = re.compile(r'(\d+\.\d)|(\.\d+)')
-    return parse_version(version_format.search(version_to_parse)[0])
+    major_version = version_format.search(version_to_parse)
+    assert major_version, f"version_to_parse: '{version_to_parse}' isn't supported scylla version string"
+    return parse_version(major_version[0])
```
fgelcer, line 125: nitpick: 
`major_version` is only the 1st item of the tuple created by `version_format.search(version_to_parse)` IIRC
so it should be something like `(2022, 1, 0)`

fgelcer (approved):
LGTM

sdcm/utils/version_utils.py
fruch: the logic is to extract the major version, the reason the first index is taken on line 125, is cause it can be None.

fruch: if you think the version should include the minor as well, then this code was broken to begin with...

roydahan (dismissed):

roydahan:
failing checks, please fix and merge.

sdcm/utils/version_utils.py
fgelcer: not true, because `re.compile(r'(\d+\.\d)|(\.\d+)')` will return `x.y` and it is major and minor versions

so, when you have as a version `2021.1.8`, the result will be `major=2021`, `minor=1` and `micro=8`

we can consider as our major versions the whole `x.y`, but this is not the most correct

fruch: that's not the terminology we use here in scylla for our versions:

2022.1 is a major version
2022.1.8 is a minor version

fgelcer: > that's not the terminology we use here in scylla for our versions:
> 
> 2022.1 is a major version 2022.1.8 is a minor version

wrong, `2022` is major version, `1` is minor version, `8` is build/micro version

as we have now 2022.1 and 2022.2, although we may call them as "version" or "major version", from version convention, what you said isn't accurate

fgelcer:
@fruch @roydahan , we need this on 5.2 and 2023.1 to have upgrade tests passing

PR #5794, opened 2023-02-09
fix(pipeline): dont fallback to builder when no sct runner

soyacz:
When sct-runner is not created due some error, longevity test continues using builder machine. This causes issues on Azure when builder is located in external network and uses private ips. Also with new builders which are weak are not capable of running longevity.

This commit removes exception catch from `Create SCT Runner` step, failing whole pipeline early.

### DIFF
--- a/vars/longevityPipeline.groovy
+++ b/vars/longevityPipeline.groovy
@@ -192,13 +192,11 @@ def call(Map pipelineParams) {
             }
             stage('Create SCT Runner') {
                 steps {
-                    catchError(stageResult: 'FAILURE') {
-                        script {
-                            wrap([$class: 'BuildUser']) {
-                                dir('scylla-cluster-tests') {
-                                    timeout(time: 5, unit: 'MINUTES') {
-                                        createSctRunner(params, runnerTimeout , builder.region)
-                                    }
+                    script {
+                        wrap([$class: 'BuildUser']) {
+                            dir('scylla-cluster-tests') {
+                                timeout(time: 5, unit: 'MINUTES') {
+                                    createSctRunner(params, runnerTimeout , builder.region)
                                 }
                             }
                         }


soyacz:
example run: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/27/
the drawback is that we won't see anything in Argus - so we can miss this failure.

fruch:
> example run: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/27/ the drawback is that we won't see anything in Argus - so we can miss this failure.

same as any other stage, like "get duration" timing out

vars/longevityPipeline.groovy
```diff
@@ -192,13 +192,11 @@ def call(Map pipelineParams) {
...
             stage('Create SCT Runner') {
                 steps {
-                    catchError(stageResult: 'FAILURE') {
-                        script {
-                            wrap([$class: 'BuildUser']) {
-                                dir('scylla-cluster-tests') {
-                                    timeout(time: 5, unit: 'MINUTES') {
-                                        createSctRunner(params, runnerTimeout , builder.region)
-                                    }
+                    script {
```
vponomaryov, line 195: Doing so it will try to run all other stages.
I think we should do the same actions in ccase of error like in the `Preparation` stage:

```
                            currentBuild.result = 'FAILURE'
                            error('SCT runner creation failed')
```

fruch:
> > example run: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/27/ the drawback is that we won't see anything in Argus - so we can miss this failure.
> 
> same as any other stage, like "get duration" timing out

please create an issue for report to Argus more places.

fgelcer (approved):
should we be doing this change only for Azure?

on second thought, if we failed to create a runner, we could indeed stop the test earlier...

vars/longevityPipeline.groovy
fruch: run job show it does not run the next steps

soyacz: > Doing so it will try to run all other stages. I think we should do the same actions in case of error like in the `Preparation` stage:
> 
> ```
> currentBuild.result = 'FAILURE'
> error('SCT runner creation failed')
> ```

But won't `catchError` will catch it and move forward?
Current solution just stops the pipeline (excluding 'Post Actions` step which does always some cleanup and sends email anyway)

vponomaryov: From another side, yes, doesn't really make any difference. Fast test end still is the result.

PR #2823, opened 2020-11-03
improvement(remove default from get method): Remove 'default' parameter from params.get() method

juliayakovlev:
Because of most of defaults are stored in the default yamls
and to prevent adding defaults in code it's better to remove
this parameter.

## PR pre-checks (self review)

- [x] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [x] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5797, opened 2023-02-09
refactor(full aggregation scan): rework full aggregation scan according  to task https://github.com/scylladb/qa-tasks/issues/1004

temichus:
according to qa-task https://github.com/scylladb/qa-tasks/issues/1004 the following changes were made in full aggregation scan operations 1 remove randomizing from the aggregation scan
2 as a result - make more ERROR evens instead of WARNING
3 put a big timeout for the aggregation scan(30min by default) 
4 detect different failure reasons
5 update unit tests. coverage of sdcm/scan_operation_thread.py: 88%

### DIFF
--- a/unit_tests/test_scan_operation_thread.py
+++ b/unit_tests/test_scan_operation_thread.py
@@ -11,7 +11,7 @@ import os
 from threading import Event
 from unittest.mock import MagicMock, patch
 import pytest
-from cassandra import OperationTimedOut
+from cassandra import OperationTimedOut, ReadTimeout
 from unit_tests.test_cluster import DummyDbCluster, DummyNode
 from sdcm.scan_operation_thread import ScanOperationThread, FullScanParams
 
@@ -109,7 +109,6 @@ def test_scan_positive(mode, events, cluster):  # pylint: disable=redefined-oute
         assert "MockCqlConnectionPatient" in all_events[1]
 
 
-########################################################################################################################
 class ExecuteOperationTimedOutMockCqlConnectionPatient(MockCqlConnectionPatient):
     def execute(*args, **kwargs):
         # pylint: disable=unused-argument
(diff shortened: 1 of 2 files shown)


longevity_test.py
```diff
@@ -42,7 +42,7 @@ def _get_tombstone_gc_verification_params(self) -> dict:
         return params
 
     def _get_scan_operation_params(self) -> list[FullScanParams]:
-        params = self.params.get("run_fullscan", {})
+        params = self.params.get("run_fullscan", [])
```
fgelcer, line 45: i think it should not work, since #2823

unit_tests/test_scan_operation_thread.py
```diff
@@ -151,8 +150,49 @@ def test_scan_negative_operation_timed_out(mode, severity, timeout, execute_mock
     assert "Severity.NORMAL" in all_events[0] and "period_type=begin" in all_events[0]
     assert f"Severity.{severity}" in all_events[1] and "period_type=end" in all_events[1]
 
+#######################################################################################################################
```
fgelcer, line 153: IMHO we don't need this

longevity_test.py
fruch: Guess which commit brought it back

https://github.com/scylladb/scylla-cluster-tests/blame/73dc21fe6b2fdd387986f9515bcdca9945002bc8/sdcm/sct_config.py#L1829

fgelcer: @temichus , please revert this change (from the commit above) and use as we currently do in the entire SCT

fgelcer: i always forget about this was removed from `self.params`

fruch:
@temichus do you have references to jobs running with those changes ?

fruch (approved):

PR #5799, opened 2023-02-09
feature(pre-commit): replace pylint with ruff

fruch:
Ruff is rust base linter for python that work incredibly fast
since pylint speed is causing use issue on SCT CI
we should start using ruff,
for now it uses the pylint set of rules it has implemented
and we can slow extend its configuration to use more
of the available rules set it has

Ref: https://github.com/charliermarsh/ruff
Ref: https://www.youtube.com/watch?v=jeoL4qsSLbE


add_new_dc_test.py
```diff
@@ -2,7 +2,7 @@
 from typing import Tuple, List
 
 from cassandra import ConsistencyLevel
-from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
```
soyacz, line 5: why removing these disable statements?

fruch: since they are pylint related, and I suggesting at moving to a different tool

cdc_replication_test.py
```diff
@@ -41,7 +39,7 @@ def mode_str(mode: Mode) -> str:
     return {
         Mode.DELTA: 'delta',
         Mode.PREIMAGE: 'preimage',
-        Mode.POSTIMAGE: 'postimage'
+        Mode.POSTIMAGE: 'postimage',
```
soyacz, line 42: can we make ruff behave like pylint and ignore these?

add_new_dc_test.py
soyacz: but some checks looks like valid checks. Now they will be missed?

cdc_replication_test.py
fruch: We could, but those are auto fixable, and something @vponomaryov for example was asking in lots of PRs

add_new_dc_test.py
fruch: those comments doesn't define a check, they override it
are you asking if we have the equivalent in ruff, I don't think so, since it's doesn't import the code some as pylint can.

soyacz: yes, we override because pylint check would cause to fail - now ruff will omit some checks (this one is just example), but we don't know if they don't support it in the future and cause fail then.

fruch: Then once we enable certain checks of new features in ruff we will add what is need, leaving the pylint comments would be much help (since ruff doesn't read them)

add_new_dc_test.py
```diff
@@ -90,7 +90,7 @@ def add_node_in_new_dc(self) -> BaseNode:
         self.monitors.reconfigure_scylla_monitoring()
 
         status = self.db_cluster.get_nodetool_status()
-        assert len(status.keys()) == 2, f"new datacenter was not registered. Cluster status: {status}"
+        assert len(status.keys()) == 2, f"new datacenter was not registered. Cluster status: {status}"  # noqa: PLR2004
```
vponomaryov, line 93: According to our situation in SCT, we may want to disable "PLR2004" rule in config.

fruch: my idea was to skip it for current code, and make sure it's enable from now on..

but yes, this generate a big amount of changes

fruch: I've ignored it in the configuration for now

add_new_dc_test.py
fruch: I decided not to remove the pylint comment for no, I'm leaving them as is

fruch:
@roydahan @soyacz @vponomaryov @juliayakovlev @k0machi 

since this gonna affect the daily work of all of us, it would help if you can take it for a walk, and see how it work for you.

vponomaryov:
@fruch I think we should skip rules making us replace `single-quote` with `double quote` symbols and `adding trailing comma`.
Why? Because it is all over our codebase.

fruch:
> @fruch I think we should skip rules making us replace `single-quote` with `double quote` symbols and `adding trailing comma`.
> Why? Because it is all over our codebase.
> 
> Then, it would be much easier to try it out.

You suggesting it for a way to try it out only ? to see how fast it is ?

Or skip it since it's making too many changes to the code ?

vponomaryov:
> > @fruch I think we should skip rules making us replace `single-quote` with `double quote` symbols and `adding trailing comma`.
> > Why? Because it is all over our codebase.
> > Then, it would be much easier to try it out.
> 
> You suggesting it for a way to try it out only ? to see how fast it is ?
> 
> Or skip it since it's making too many changes to the code ?

I have no doubts that it is fast.
It is about second assumption, but not completely it.
WIth much fewer changes it will be possible to
- See whole set of changes we need to do without those 2 already mentioned
- Apply it for `trying out` anytime on almost any master branch state. Now it is already out of compatibility.

soyacz:
> @fruch I think we should skip rules making us replace `single-quote` with `double quote` symbols and `adding trailing comma`.

I'm opting also to skip this - we won't get too much from this change but it will cause lot's of troubles when backporting or when checking 'git blame'

fruch:
> > @fruch I think we should skip rules making us replace `single-quote` with `double quote` symbols and `adding trailing comma`.
> 
> I'm opting also to skip this - we won't get too much from this change but it will cause lot's of troubles when backporting or when checking 'git blame'

we have lot of other changes that would affect git blame, cause of formatting has change, those would be a drop in the see.

and changing the tool/formatter does have it's price, backporting would be hard, regregless of we remove one or two rules out.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -75,75 +72,82 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
...
                     image = container["image"].split("/")[-1]
                     if image.startswith("scylla-operator:") and image.split(":")[-1] != expected_operator_tag:
-                        pods_with_wrong_image_tags.append({
-                            "namespace": namespace,
-                            "pod_name": pod["metadata"]["name"],
-                            "container_name": container["name"],
-                            "image": image,
-                        })
+                        pods_with_wrong_image_tags.append(
+                            {
```
vponomaryov, line 82: I find it counterproductive to use more new lines than it is really needed like here.
If we have bunch of single-element parent wrappers it doesn't mean it is useful to use new line for entering and closing of each of them.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -75,75 +72,82 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
...
-        'exposeOptions': {
-            'nodeService': {'type': 'Headless'},
-            'broadcastOptions': {key: {'type': 'PodIP'} for key in ('nodes', 'clients')},
-        },
-        'developerMode': True,
-        'fullnameOverride': cluster_name,
-        'scyllaImage': {
-            'repository': k8s_cluster.params.get('docker_image'),
-            'tag': k8s_cluster.params.get('scylla_version'),
+    values = HelmValues(
```
vponomaryov, line 112: Same here.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -194,9 +212,9 @@ def get_pod_names_and_ips(cluster_name: str, namespace: str):
             for current_namespace in pods_data.keys():
                 for pod_name, pod_ip in pods_data[current_namespace].items():
                     cqlsh_cmd = "SELECT JSON peer, data_center, rack, rpc_address FROM system.peers;"
-                    cqlsh_results = kubectl(
-                        f"exec {pod_name} -- /bin/cqlsh -e \"{cqlsh_cmd}\"",
-                        namespace=current_namespace).stdout.split("---\n")[-1].split("\n")
+                    cqlsh_results = (
+                        kubectl(f'exec {pod_name} -- /bin/cqlsh -e "{cqlsh_cmd}"', namespace=current_namespace).stdout.split("---\n")[-1].split("\n")
+                    )
```
vponomaryov, line 217: It created too long line.
One more case of counter-productivity.
Probably there should be some config option for setting the limit for line length it tries to apply.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -208,16 +226,14 @@ def get_pod_names_and_ips(cluster_name: str, namespace: str):
             need_to_collect_logs = False
         finally:
             if need_to_collect_logs:
-                KubernetesOps.gather_k8s_logs(
-                    logdir_path=logdir, kubectl=kubectl, namespaces=[namespace, namespace2])
+                KubernetesOps.gather_k8s_logs(logdir_path=logdir, kubectl=kubectl, namespaces=[namespace, namespace2])
```
vponomaryov, line 229: The same here.
It was on 2 lines because of the length.

vponomaryov (commented):
Tried it.
Works really fast.
But it is very big questions which rules we really want to enable.
Personally, I find lots of changes not useful (single-quotes to double-quotes transformation) or counter-productive (long line and much more lines with the same readability).

pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+lint.ignore = ["E501", "PLR2004", "COM812", ]
+
+target-version = "py310"
+
+extend-exclude = [
+    "docker/alternator-dns",
+]
+
+force-exclude = true
+line-length = 160
```
vponomaryov, line 13: Here it is.
The default is 88 according to the official docs [1].
160 is too big for opening 2 vertical windows on a screen with 1920 width.
My eyes can afford me +-120 symbols in this case.
Another bigger font size is 112.
So, if we could have here at max 112 or 110 symbols. It would be great.

[1] https://docs.astral.sh/ruff/settings/#line-length

pyproject.toml
```diff
@@ -0,0 +1,14 @@
+[tool.ruff]
+lint.select = ["PL", "YTT", "BLE", "I", "UP", "A", "COM", "T10", "EXE", "TID", "F401", "F402", "F821"]
+
+lint.ignore = ["E501", "PLR2004", "COM812", ]
+
+target-version = "py310"
+
+extend-exclude = [
+    "docker/alternator-dns",
```
vponomaryov, line 9: Why is it exclusion?
I think we should ad to exclusion only those files which cannot/shouldnot change.
Such as imported `as-is` from some projects or designed to satisfy some tools/libraries which don't know about ruff preferences...

functional_tests/scylla_operator/test_functional.py
vponomaryov: Thanks for the links.

pyproject.toml
fruch: I was playing around with this figure for the exact same reason.

it was 120 with autopep8, but it wasn't that strict, like for comment

pyproject.toml
fruch: the file originates from: https://github.com/scylladb/alternator-load-balancing/blob/master/dns/dns-loadbalancer.py

so we didn't wanted to do many changes for, so it was excluded from linting and auto formatting.
so it can be comparabled, but never had any changes, and not sure what's the plans for it

vponomaryov: The comment about the source of the file actually should have been in the file itself or text file around it.
Example: https://github.com/scylladb/scylla-cluster-tests/blob/d92d7cc71665c6155bea717ca7082723861d7183/sdcm/utils/version_utils.py#L48

pyproject.toml
```diff
@@ -0,0 +1,14 @@
...
+ignore = ["E501", "PLR2004"]
+
+target-version = "py311"
+
+extend-exclude = [
+    "docker/alternator-dns",
+]
+
+force-exclude = true
+line-length = 240
```
vponomaryov: `240`? After request to `decrease` the value having `160`?

pyproject.toml
```diff
@@ -0,0 +1,19 @@
...
+lint.ignore = ["E501", "PLR2004", "COM812", ]
+
+target-version = "py310"
+
+extend-exclude = [
+    "docker/alternator-dns",
+]
+respect-gitignore = true
+force-exclude = true
+line-length = 112
```
fruch, line 13: @vponomaryov 

look at the end results, on the commit that does the formatting, before it's not really used by the linter

vponomaryov: Confusing to see 2 different definitions in scope of single PR, but ok.

pyproject.toml
fruch: it's now 112

functional_tests/scylla_operator/test_functional.py
fruch: line length is now 112

fruch:
@soyacz @vponomaryov 

I've remove the `double quote`and  `trailing comma`, but still the list of change is quite huge.
I think it's acceptable, and a one off that would happen on any major change of formatter/linter.

pyproject.toml
fruch: I've removed it from `exclude`

we don't really update this file ever, we can format it as we want.

pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+lint.select = ["PL", "YTT", "BLE", "I", "UP", "F401", "F402", "F821", "T10", "EXE", "TID"]
+
+lint.ignore = ["E501", "PLR2004", "COM812", ]
+
+target-version = "py310"
+
+respect-gitignore = true
+force-exclude = true
+line-length = 112
+isort.split-on-trailing-comma=false
```
vponomaryov, line 11: Testing the violation of the lint rules I see following:
```
ruff.....................................................................Failed
- hook id: ruff
- files were modified by this hook

warning: The top-level linter settings are deprecated in favour of their counterparts in the `lint` section. Please update the following options in `pyproject.toml`:
  - 'isort' -> 'lint.isort'
Found 1 error (1 fixed, 0 remaining).
```

vponomaryov:
The precommit fails:
```
17:28:11  docker/alternator-dns/dns_server.py:1:1: EXE001 Shebang is present but file is not executable
17:28:11  Found 1 error.
```

performance_regression_operator_multi_tenant_test.py
```diff
@@ -38,35 +33,32 @@ def run_fstrim_on_all_db_nodes(self):
...
-            stress_queue, params = [], {
-                'prefix': 'preload-',
-            }
+            tenant.create_test_stats(sub_type='write-prepare', doc_id_with_timestamp=True)
+            stress_queue, params = (
+                [],
+                {
+                    'prefix': 'preload-',
+                },
+            )
```
vponomaryov, line 53: This change looks odd.
Better to make either one line with still both options on it or 2 lines - one per option.
But not `this`.

performance_regression_operator_multi_tenant_test.py
```diff
@@ -18,14 +16,11 @@
...
-# pylint: disable=too-many-public-methods
 class PerformanceRegressionOperatorMultiTenantTest(MultiTenantTestMixin, PerformanceRegressionTest):
     load_iteration_timeout_sec = 7200
 
-    def create_test_stats(self, *args, **kwargs):  # pylint: disable=unused-argument
-        self.log.info(
-            "Suppress the test class stats creation. "
-            "Leave it for the per-DB classes.")
+    def create_test_stats(self, *args, **kwargs):
+        self.log.info("Suppress the test class stats creation. " "Leave it for the per-DB classes.")
```
vponomaryov, line 23: Redundant split for 2 str objects.

performance_regression_lwt_test.py
```diff
@@ -38,84 +37,90 @@ class PerformanceRegressionLWTTest(PerformanceRegressionTest):
...
-            'groups': ['MIXED']
+            'groups': ['MIXED'],
         },
     ]
 
     @log_run_info
-    @retrying(n=3, sleep_time=15, allowed_exceptions=(UnexpectedExit, Failure))  # retrying since SSH can fail with 255
+    @retrying(
+        n=3, sleep_time=15, allowed_exceptions=(UnexpectedExit, Failure)
+    )  # retrying since SSH can fail with 255
```
vponomaryov, line 107: I think we should keep decorator be one liner and move the comment above it.

vponomaryov:
@fruch 
Added more comments after looking at several python files.
There area lot of cases where transformation is ugly because the structure doesn't get changed.

So, multiplying this observation onto the 400+ files , which is insane to review as part of a single PR, I think we should not merge such a big change.

**Proposal:**
I think, better would be to add `ruff` as second linter where almost all python files are exlcuded except bunch of fixed ones.
And then we can extend the coverage fixing files in bunches.
Something like 40 commit with 10 python files...

@soyacz WDYT?

fruch:
> @fruch 
> Added more comments after looking at several python files.
> There are a lot of cases where transformation is ugly because the structure doesn't get changed.
> 
> So, multiplying this observation onto the 400+ files , which is insane to review as part of a single PR, I think we should not merge such a big change.
> 
> **Proposal:**
> I think, better would be to add `ruff` as second linter where almost all python files are exlcuded except bunch of fixed ones.
> And then we can extend the coverage fixing files in bunches.
> Something like 40 commit with 10 python files...
> 
> @soyacz WDYT?

We can try black, but it's gonna be a similar experience, it's a very opinionated formatter, and people might not like the choices it would do.

But that's the deal with those formatters, you trust them to eliminate discussions about format.

As for splitting, it might be doable, but gonna be a bit tedious, like we did in dtest when introduced pylint

soyacz:
> @fruch Added more comments after looking at several python files. There are a lot of cases where transformation is ugly because the structure doesn't get changed.
> 
> So, multiplying this observation onto the 400+ files , which is insane to review as part of a single PR, I think we should not merge such a big change.
> 
> **Proposal:** I think, better would be to add `ruff` as second linter where almost all python files are exlcuded except bunch of fixed ones. And then we can extend the coverage fixing files in bunches. Something like 40 commit with 10 python files...
> 
> @soyacz WDYT?

I think we should not format with ruff for now.
My proposal would to leave autopep6 and replace pylint only. And put in pyproject.toml:
```
[tool.ruff]
lint.select = ["PL", "YTT", "BLE", "I", "UP", "F401", "F402", "F821", "T10", "EXE", "TID"]

lint.ignore = ["E501", "PLR2004", "COM812", "BLE001", "UP031"]

target-version = "py310"

respect-gitignore = true
force-exclude = true
line-length = 112

[tool.ruff.lint.pylint]
max-args = 20
max-statements = 180
max-branches = 81
```
and fix manually errors (only around 200 and some are valid ones)

fruch:
> > @fruch Added more comments after looking at several python files. There are a lot of cases where transformation is ugly because the structure doesn't get changed.
> > So, multiplying this observation onto the 400+ files , which is insane to review as part of a single PR, I think we should not merge such a big change.
> > **Proposal:** I think, better would be to add `ruff` as second linter where almost all python files are exlcuded except bunch of fixed ones. And then we can extend the coverage fixing files in bunches. Something like 40 commit with 10 python files...
> > @soyacz WDYT?
> 
> I think we should not format with ruff for now. My proposal would to leave autopep6 and replace pylint only. And put in pyproject.toml:
> 
> ```
> [tool.ruff]
> lint.select = ["PL", "YTT", "BLE", "I", "UP", "F401", "F402", "F821", "T10", "EXE", "TID"]
> 
> lint.ignore = ["E501", "PLR2004", "COM812", "BLE001", "UP031"]
> 
> target-version = "py310"
> 
> respect-gitignore = true
> force-exclude = true
> line-length = 112
> 
> [tool.ruff.lint.pylint]
> max-args = 20
> max-statements = 180
> max-branches = 81
> ```
> 
> and fix manually errors (only around 200 and some are valid ones)

I've remove the formatter change, and I'll see how we introduce it in smaller batches

I'm not extending the number of pylint, just to minimize the change, we it the same figured as we had before., also BLE001 (it's the only one in BLE)

upgrade_test.py
```diff
@@ -605,7 +605,7 @@ def _update_scylla_yaml_on_node(node_to_update: BaseNode, updates: dict):
         with node_to_update.remote_scylla_yaml() as scylla_yaml:
             scylla_yaml.update(updates)
 
-    def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-statements
```
vponomaryov, line 608: The `pylint: disable=too-many-locals,too-many-statements` becomes redundant and must be removed.

sdcm/cluster_k8s/eks.py
```diff
@@ -41,6 +43,7 @@
 )
 from sdcm.remote import LOCALRUNNER
 from sdcm.utils.aws_utils import (
+    EksClusterCleanupMixin,
```
vponomaryov, line 46: Previous commit mistakenly removed the import of the `EksClusterCleanupMixin` class.
Now it is added...

docker/alternator-dns/dns_server.py
```diff
@@ -1,14 +1,14 @@
 #!/usr/bin/python3
-import sys
```
vponomaryov, line 2: This is not in exclusions anymore?

fruch: I've removed it

sdcm/cluster_aws.py
```diff
@@ -77,7 +77,7 @@ class AWSCluster(cluster.BaseCluster):  # pylint: disable=too-many-instance-attr
     Cluster of Node objects, started on Amazon EC2.
     """
 
-    def __init__(self, ec2_ami_id, ec2_subnet_id, ec2_security_group_ids,  # pylint: disable=too-many-arguments
+    def __init__(self, ec2_ami_id, ec2_subnet_id, ec2_security_group_ids,  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/cluster_aws.py
```diff
@@ -825,7 +825,7 @@ def ena_support(self) -> bool:
 
 class ScyllaAWSCluster(cluster.BaseScyllaCluster, AWSCluster):
 
-    def __init__(self, ec2_ami_id, ec2_subnet_id, ec2_security_group_ids,  # pylint: disable=too-many-arguments
+    def __init__(self, ec2_ami_id, ec2_subnet_id, ec2_security_group_ids,  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/cluster_aws.py
```diff
@@ -1032,7 +1032,7 @@ def __init__(self, ec2_ami_id, ec2_subnet_id, ec2_security_group_ids,  # pylint:
 
 class MonitorSetAWS(cluster.BaseMonitorSet, AWSCluster):
 
-    def __init__(self, ec2_ami_id, ec2_subnet_id, ec2_security_group_ids,  # pylint: disable=too-many-arguments
+    def __init__(self, ec2_ami_id, ec2_subnet_id, ec2_security_group_ids,  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/cluster_azure.py
```diff
@@ -169,7 +169,7 @@ def configure_remote_logging(self) -> None:
 
 
 class AzureCluster(cluster.BaseCluster):   # pylint: disable=too-many-instance-attributes
-    def __init__(self, image_id, root_disk_size,  # pylint: disable=too-many-arguments, too-many-locals
+    def __init__(self, image_id, root_disk_size,  # pylint: disable=too-many-arguments, too-many-locals  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/cluster_gce.py
```diff
@@ -241,7 +241,7 @@ class GCECluster(cluster.BaseCluster):  # pylint: disable=too-many-instance-attr
     """
     _gce_service: compute_v1.InstancesClient
 
-    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments
+    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/cluster_gce.py
```diff
@@ -530,7 +530,7 @@ def add_nodes(self, count, ec2_user_data='', dc_idx=0, rack=0, enable_auto_boots
 
 class ScyllaGCECluster(cluster.BaseScyllaCluster, GCECluster):
 
-    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments
+    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/cluster_gce.py
```diff
@@ -566,7 +566,7 @@ def _wait_for_preinstalled_scylla(node):
 
 class LoaderSetGCE(cluster.BaseLoaderSet, GCECluster):
 
-    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments
+    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/cluster_gce.py
```diff
@@ -597,7 +597,7 @@ def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_s
 
 class MonitorSetGCE(cluster.BaseMonitorSet, GCECluster):
 
-    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments
+    def __init__(self, gce_image, gce_image_type, gce_image_size, gce_network, gce_service, credentials,  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/db_stats.py
```diff
@@ -607,17 +607,16 @@ def get_setup_details(self):
         for key, value in test_params:
             if key in exclude_details or (isinstance(key, str) and key.startswith('stress_cmd')):  # pylint: disable=no-else-continue
                 continue
+            elif is_gce and key in \
+                    ['instance_type_loader',  # pylint: disable=no-else-continue
```
vponomaryov: If we remove the redundant pylint comment then we can make it be one line.

sdcm/mgmt/cli.py
```diff
@@ -546,7 +546,7 @@ def create_restore_task(self, restore_schema=False, restore_data=False, location
         LOGGER.debug("Created task id is: {}".format(task_id))
         return RestoreTask(task_id=task_id, cluster_id=self.id, manager_node=self.manager_node)
 
-    def create_backup_task(self, dc_list=None,  # pylint: disable=too-many-arguments,too-many-locals,too-many-branches
+    def create_backup_task(self, dc_list=None,  # pylint: disable=too-many-arguments,too-many-locals,too-many-branches  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/monitorstack/__init__.py
```diff
@@ -37,7 +37,7 @@ class ErrorUploadAnnotations(Exception):
     pass
 
 
-def restore_monitoring_stack(test_id, date_time=None):  # pylint: disable=too-many-return-statements,too-many-locals
+def restore_monitoring_stack(test_id, date_time=None):  # pylint: disable=too-many-return-statements,too-many-locals  # noqa: PLR0911
```
vponomaryov: Pylint comment is redundant.

sdcm/prometheus.py
```diff
@@ -68,7 +68,7 @@ def start_metrics_server():
 
 
 def nemesis_metrics_obj(metric_name_suffix=''):
-    global NM_OBJ  # pylint: disable=global-statement,global-variable-not-assigned
+    global NM_OBJ  # pylint: disable=global-statement,global-variable-not-assigned  # noqa: PLW0602
```
vponomaryov: Pylint comment is redundant.

sdcm/utils/remote_logger.py
```diff
@@ -575,7 +575,7 @@ def _logger_cmd(self) -> str:
         return f"echo \"I`date -u +\"%m%d %H:%M:%S\"`              {message}\" >> {self._target_log_file} 2>&1"
 
 
-def get_system_logging_thread(logs_transport, node, target_log_file):  # pylint: disable=too-many-return-statements
+def get_system_logging_thread(logs_transport, node, target_log_file):  # pylint: disable=too-many-return-statements  # noqa: PLR0911
```
vponomaryov: Pylint comment is redundant.

sdcm/utils/get_username.py
```diff
@@ -24,7 +24,7 @@ def get_email_user(email_addr: str) -> str:
     return email_addr.strip().split("@")[0]
 
 
-def get_username() -> str:  # pylint: disable=too-many-return-statements
+def get_username() -> str:  # pylint: disable=too-many-return-statements  # noqa: PLR0911
```
vponomaryov: Pylint comment is redundant.

sdcm/tester.py
```diff
@@ -2264,7 +2264,7 @@ def create_keyspace(self, keyspace_name, replication_factor):
             does_keyspace_exist = self.wait_validate_keyspace_existence(session, keyspace_name)
         return does_keyspace_exist
 
-    def create_table(self, name, key_type="varchar",  # pylint: disable=too-many-arguments,too-many-branches
+    def create_table(self, name, key_type="varchar",  # pylint: disable=too-many-arguments,too-many-branches  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/stress_thread.py
```diff
@@ -92,7 +92,7 @@ def __exit__(self, exc_type, exc_val, exc_tb):
 class CassandraStressThread(DockerBasedStressThread):  # pylint: disable=too-many-instance-attributes
     DOCKER_IMAGE_PARAM_NAME = 'stress_image.cassandra-stress'
 
-    def __init__(self, loader_set, stress_cmd, timeout, stress_num=1, keyspace_num=1, keyspace_name='', compaction_strategy='',  # pylint: disable=too-many-arguments
+    def __init__(self, loader_set, stress_cmd, timeout, stress_num=1, keyspace_num=1, keyspace_name='', compaction_strategy='',  # pylint: disable=too-many-arguments  # noqa: PLR0913
```
vponomaryov: Pylint comment is redundant.

sdcm/sct_events/events_processes.py
```diff
@@ -166,7 +166,7 @@ def __str__(self):
 
 
 def create_default_events_process_registry(log_dir: Union[str, Path]):
-    global _EVENTS_PROCESSES  # pylint: disable=global-statement
+    global _EVENTS_PROCESSES  # pylint: disable=global-statement  # noqa: PLW0603
```
vponomaryov: Pylint comment is redundant.

vponomaryov (changes requested):
Added comments for the first commit.

Can we have just one commit in the PR to avoid lots of re-reviews after catching undesired changes?
Also, probably, removing all the pylint messages as a follow-up PR (second one).

artifacts_test.py
```diff
@@ -436,7 +436,7 @@ def get_email_data(self):
         email_data = self._get_common_email_data()
         try:
             node = self.node
-        except Exception:  # pylint: disable=broad-except
+        except (ValueError, IndexError):
```
vponomaryov, line 438: It may be also `AttributeError` and some other.
I don't see any benefit narrowing down the exception coverage here.

docker/alternator-dns/dns_server.py
```diff
@@ -34,7 +34,7 @@ def livenodes_update():
             # If we're successful, replace livenodes by the new list
             livenodes = a
             print(livenodes)
-        except Exception:  # pylint: disable=broad-except
+        except Exception:  # pylint: disable=broad-except  # noqa: BLE001
```
vponomaryov: Redundant pylint message.

functional_tests/scylla_operator/conftest.py
```diff
@@ -175,7 +175,7 @@ def _bring_cluster_back_to_original_state(
             db_cluster.restart_scylla()
         db_cluster.wait_for_nodes_up_and_normal(
             nodes=db_cluster.nodes, verification_node=db_cluster.nodes[0])
-    except Exception as exc:  # pylint: disable=broad-except
+    except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
```
vponomaryov: Redundant pylint message.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -318,7 +318,7 @@ def wait_for_cleanup_logs(log_follower_name, log_follower, db_cluster):
                 time.sleep(4)
                 db_cluster.nodes[0].run_cqlsh(cmd=f"DROP KEYSPACE IF EXISTS {current_ks_name}", timeout=60)
                 time.sleep(4)
-            except Exception as exc:  # pylint: disable=broad-except
+            except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
```
vponomaryov: Redundant pylint message.

longevity_test.py
```diff
@@ -462,13 +462,9 @@ def _flush_all_nodes(self):
...
     def get_email_data(self):
         self.log.info("Prepare data for email")
-        email_data = {}
         grafana_dataset = {}
 
-        try:
-            email_data = self._get_common_email_data()
-        except Exception as error:  # pylint: disable=broad-except
-            self.log.exception("Error in gathering common email data: Error:\n%s", error, exc_info=error)
+        email_data = self._get_common_email_data()
```
vponomaryov, line 464: This change is about changing logic , not just lint.
Is it intentional?

performance_regression_test.py
```diff
@@ -176,7 +176,7 @@ def display_results(self, results, test_name=''):
             with open(os.path.join(self.logdir, 'jenkins_perf_PerfPublisher.xml'), 'w', encoding="utf-8") as pref_file:
                 content = """<report name="%s report" categ="none">%s</report>""" % (test_name, test_xml)
                 pref_file.write(content)
-        except Exception as ex:  # pylint: disable=broad-except
+        except Exception as ex:  # pylint: disable=broad-except  # noqa: BLE001
```
vponomaryov: Redundant pylint message.

vponomaryov (commented):
Second commit comments

longevity_test.py
fruch: yes, it's the same across all tests, and in majority of them it's not under try/except, the all to this function `get_email_data()` is also wrapper with try/except clause.
anyhow there no point moving one, if we didn't had this minimal set of information, we won't be able to format the email report out of it at this point.

we should be raising events on those failures, not printing it to the log only.

artifacts_test.py
vponomaryov: The `AttributeError` is totally possible here.

soyacz:
I'd leave pylint comments to minimize the change. Removing them is similar to reformat which we wanted to avoid.

vponomaryov:
> I'd leave pylint comments to minimize the change. Removing them is similar to reformat which we wanted to avoid.

I proposed to leave only first commit in this PR.
Second, second PR with pylint comments removal.
And then all other, separately.

Doing so, we could merge the first commit today.

artifacts_test.py
fruch: yeah, if you remove `db_cluster` assignment from `Tester.__init__`

fruch:
> > I'd leave pylint comments to minimize the change. Removing them is similar to reformat which we wanted to avoid.
> 
> I proposed to leave only first commit in this PR. Then, second PR with pylint comments removal. And then all other, separately.
> 
> Doing so, we could merge the first commit today. Also, doing so we can improve places where the linter does strange things such as 2 string objects instead of one on a single line...

I fail to see how splitting it to multiple PR would help, just for reviewing it easier ?

vponomaryov:
> > > I'd leave pylint comments to minimize the change. Removing them is similar to reformat which we wanted to avoid.
> > 
> > 
> > I proposed to leave only first commit in this PR. Then, second PR with pylint comments removal. And then all other, separately.
> > Doing so, we could merge the first commit today. Also, doing so we can improve places where the linter does strange things such as 2 string objects instead of one on a single line...
> 
> I fail to see how splitting it to multiple PR would help, just for reviewing it easier ?

Just? Possibility to have full review will allow to catch lots of undesired changes where it is better to update some places manually.

fruch:
> > > > I'd leave pylint comments to minimize the change. Removing them is similar to reformat which we wanted to avoid.
> > > 
> > > 
> > > I proposed to leave only first commit in this PR. Then, second PR with pylint comments removal. And then all other, separately.
> > > Doing so, we could merge the first commit today. Also, doing so we can improve places where the linter does strange things such as 2 string objects instead of one on a single line...
> > 
> > 
> > I fail to see how splitting it to multiple PR would help, just for reviewing it easier ?
> 
> Just? Possibility to have full review will allow to catch lots of undesired changes where it is better to update some places manually.

If you are gonna read it all line by line, every signal change here. I'll just merge it now and that's it.

that's not the point of such a change, it's not to review the changes, themselves but the tool / rules we are using.

vponomaryov:
> If you are gonna read it all line by line, every signal change here. I'll just merge it now and that's it.
> 
> that's not the point of such a change, it's not to review the changes, themselves but the tool / rules we are using.

I don't see arguments against creating first PR with first commit and further keeping adequate number of changes per PR.
I see only threat to merge if I review code in the place where it is the main goal - `to review code`.

We already could merge first one or two commits with full confidence in the made changes.

artifacts_test.py
vponomaryov: The absence of the `nodes` attr is also one of possible causes.
So, the statement `those are the only two options` is false because there are 3 different possible exceptions.

fruch: same as before it's in `nodes` are being assigned in `BaseCluster.__init__` 

to sum it up, we know exactly what can get broken here, i.e. the list of node can be empty, cause test failed before starting the node.

fruch:
> > If you are gonna read it all line by line, every signal change here. I'll just merge it now and that's it.
> > that's not the point of such a change, it's not to review the changes, themselves but the tool / rules we are using.
> 
> I don't see arguments against creating first PR with first commit and further keeping adequate number of changes per PR. I see only threat to merge if I review code in the place where it is the main goal - `to review code`.
> 
> We already could merge first one or two commits with full confidence in the made changes.

Separating those make no sense, and would me that we'll have almost no lint on master until the rest of those would be introduced. 

or we have confidence in the tool we are introducing, or we don't, I don't see a point review every single automated change that the tool was doing (which 99% of those change are)

vponomaryov (approved):
LGTM

vponomaryov:
@fruch need to rebase it.

soyacz (approved):
LGTM

Let's do it :)

PR #5802, opened 2023-02-10
feature(node): support new 'replace_node_first_boot' Scylla option

vponomaryov:
In Scylla 5.2 was added possibility to replace a node by it's host_id. 
It can be done by setting `replace-node-first-boot` Scylla option instead of old `replace-address-first-boot` one.

So, add support for it.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -267,8 +267,9 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         # enable bootstrap.
         self.enable_auto_bootstrap = True
 
-        # If node is a replacement for a dead node, store dead node private ip here
+        # If a node is a replacement for a dead node, store a dead node's private ip/host id here
         self.replacement_node_ip = None
+        self.replacement_host_id = None
 
         self._kernel_version = None
         self._uuid = None
@@ -408,6 +409,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             scylla_yml.auto_bootstrap = self.enable_auto_bootstrap
         if self.replacement_node_ip:
             scylla_yml.replace_address_first_boot = self.replacement_node_ip
+        if self.replacement_host_id:
+            scylla_yml.replace_node_first_boot = self.replacement_host_id
         if append_scylla_yaml := self.parent_cluster.params.get('append_scylla_yaml'):
             append_scylla_yaml = yaml.safe_load(append_scylla_yaml)
             if any(substr in append_scylla_yaml for substr in (
@@ -2752,6 +2755,17 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             )
         )
 
+    @property
+    def is_replacement_by_host_id_supported(self):
+        # NOTE: compare 'LegacyVersion' returns 'False' for the following comparison:
+        #         LegacyVersion("5.2.0") > LegacyVersion('5.2.0~dev')
+        #       and 'True' for the following:
+        #         LegacyVersion("5.2.0~rc1") > LegacyVersion('5.2.0~dev')
+        #       So, compare to '5.1.999' because any 5.2+ version provides 'True' in this case
+        if self.is_enterprise:
+            return LegacyVersion(self.scylla_version) > LegacyVersion('2022.2.999')
+        return LegacyVersion(self.scylla_version) > LegacyVersion('5.1.999')
+
     def _gen_cqlsh_cmd(self, command, keyspace, timeout, host, port, connect_timeout):
         """cqlsh [options] [host [port]]"""
         credentials = self.parent_cluster.get_db_auth()
@@ -4383,7 +4397,7 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
         nodes_status = node.get_nodes_status()
         check_nodes_status(nodes_status=nodes_status, current_node=node)
 
-        self.clean_replacement_node_ip(node)
+        self.clean_replacement_node_options(node)
 
     def install_scylla_manager(self, node):
         pkgs_url = self.params.get("scylla_mgmt_pkg")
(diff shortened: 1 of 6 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-10gb-3h-gce-test/5/consoleFull

sdcm/cluster.py
```diff
@@ -2752,6 +2755,17 @@ def is_cqlsh_support_cloud_bundle(self):
...
 
+    @property
+    def is_replacement_by_host_id_supported(self):
+        # NOTE: compare 'LegacyVersion' returns 'False' for the following comparison:
+        #         LegacyVersion("5.2.0") > LegacyVersion('5.2.0~dev')
+        #       and 'True' for the following:
+        #         LegacyVersion("5.2.0~rc1") > LegacyVersion('5.2.0~dev')
+        #       So, compare to '5.1.999' because any 5.2+ version provides 'True' in this case
+        if self.is_enterprise:
+            return LegacyVersion(self.scylla_version) > LegacyVersion('2022.2.999')
```
fgelcer, line 2766: isn't it better to be more explicit, and test with greater or equal?

```suggestion
            return LegacyVersion(self.scylla_version) >= LegacyVersion('2023.1.0')
```

sdcm/cluster.py
```diff
@@ -2752,6 +2755,17 @@ def is_cqlsh_support_cloud_bundle(self):
...
+        return LegacyVersion(self.scylla_version) > LegacyVersion('5.1.999')
```
fgelcer, line 2767: same here

```suggestion
        return LegacyVersion(self.scylla_version) >= LegacyVersion('5.2.0')
```

fgelcer (approved):
2 questions, but independent on the answer, LGTM

sdcm/cluster.py
vponomaryov: The answer is written in the comment above the code.
Using your variant any `dev` and `rc` versions will return `False` result. The current one will return correct `True` result in this case.

fruch: I think we should be using `parse_scylla_version`

roydahan: Can we check for it in help or scylla.yaml instead?
Every time someone changes the versions format or something like this we discover too late that these things stopped working.

roydahan (approved):
In general looks great to me.
The only thing I would have changed is the way we check that this feature is supported or not.
I don't mind it will be done by a followup PR.

sdcm/cluster.py
vponomaryov: I plan upload the version comparison solution today.
It will allow us to stop using `LegacyVersion` which blocks us from usage of latest `packaging` package.
And will correctly compare versions.

fruch: roy's suggestion is also quite valid in this case, and we already have the code the does handle the `scylla --help` output, and should be more robust 

and we still need to handle versions correctly for other cases where the difference is not manifested in `scylla --help`

vponomaryov: We will talk about it in a PR that will be uploaded.
For the moment I can say that I added support to the most, known to me, "parsing" specifics the way that
"scylla-binary" and "docker tag" will be easily comparable using new feature that will be uploaded.

vponomaryov: It is here: https://github.com/scylladb/scylla-cluster-tests/pull/5817

roydahan:
@fruch please decide if you want to merge.
If we merge today it can go into weekly runs of master and later this week to be backported.

vponomaryov:
> @fruch please decide if you want to merge. If we merge today it can go into weekly runs of master and later this week to be backported.

@fruch the https://github.com/scylladb/scylla-cluster-tests/pull/5817 can be merged after this one, I will just update that PR replacing introduced here `LegacyVersion` usage with the new version comparing approach.

So, I don't see any problem merging this one.

fruch:
> @fruch please decide if you want to merge. If we merge today it can go into weekly runs of master and later this week to be backported.

O.k., let's merge, to avoid dragging more PRs into it.

fruch:
> > @fruch please decide if you want to merge. If we merge today it can go into weekly runs of master and later this week to be backported.
> 
> @fruch the #5817 can be merged after this one, I will just update that PR replacing introduced here `LegacyVersion` usage with the new version comparing approach.
> 
> So, I don't see any problem merging this one.

I think @roydahan suggestion for using `scylla --help` a preferred option, but it's not urgent.

vponomaryov:
> > > @fruch please decide if you want to merge. If we merge today it can go into weekly runs of master and later this week to be backported.
> > 
> > 
> > @fruch the #5817 can be merged after this one, I will just update that PR replacing introduced here `LegacyVersion` usage with the new version comparing approach.
> > So, I don't see any problem merging this one.
> 
> I think @roydahan suggestion for using `scylla --help` a preferred option, but it's not urgent.

For the moment the scylla version used here in comparison is taken by using `scylla --version`.
How is the `scylla --help` is better? `Build-ID` is the diff , yes, but it is not needed for version comparisons at all.
So, what is the point?

fruch:
> > > > @fruch please decide if you want to merge. If we merge today it can go into weekly runs of master and later this week to be backported.
> > > 
> > > 
> > > @fruch the #5817 can be merged after this one, I will just update that PR replacing introduced here `LegacyVersion` usage with the new version comparing approach.
> > > So, I don't see any problem merging this one.
> > 
> > I think @roydahan suggestion for using `scylla --help` a preferred option, but it's not urgent.
> 
> For the moment the scylla version used here in comparison is taken by using `scylla --version`.
> How is the `scylla --help` is better? `Build-ID` is the diff , yes, but it is not needed for version comparisons at all.
> ```
> # scylla --help | head -n 1 
> Scylla version 5.1.1-0.20221208.1cfedc5b59f4 with build-id 0c3e5de166947928f2d129ae826a71c53f31302c starting ...
> 
> # scylla --version
> 5.1.1-0.20221208.1cfedc5b59f4
> ```
> The `5.1.1-0.20221208.1cfedc5b59f4` is the same in both cases and it is exactly what is needed in version comparison.
> So, what is the point?

Sorry, you missed the intention in Roy's suggestion, the idea is to use the help for checking the command line arguments available, not getting the version out of it.

I.e. it's a way to check what is supported without checking the version,  and it would work even without knowing which version support, or for example if a feature would be backported (not that this feature would)

PR #5808, opened 2023-02-14
fix(decommission_streaming_err): remove timeout/retry from rebuild

fruch:
b26acc9530f5629002984b3a1566a2ac18c98164 introduced a 5min timeout of the rebuild command, that won't work on any dataset the same and isn't part of the intended fix, also doing it 3 times doesn't really works any faster.

so timeout and retry are removed.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3348,9 +3348,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
         ParallelObject(objects=[trigger, watcher], timeout=1200).call_objects()
         if new_node := decommission_post_action():
-            new_node.run_nodetool("rebuild", timeout=300, retry=3)
+            new_node.run_nodetool("rebuild")
         else:
-            self.target_node.run_nodetool(sub_cmd="rebuild", timeout=300, retry=3)
+            self.target_node.run_nodetool(sub_cmd="rebuild")
 
     def start_and_interrupt_repair_streaming(self):
         """


sdcm/nemesis.py
```diff
@@ -3348,9 +3348,9 @@ def decommission_post_action():
 
         ParallelObject(objects=[trigger, watcher], timeout=1200).call_objects()
         if new_node := decommission_post_action():
-            new_node.run_nodetool("rebuild", timeout=300, retry=3)
+            new_node.run_nodetool("rebuild")
```
fgelcer, line 3351: shouldn't we have a timeout anyway? like 30 min? or perhaps even 1 hour?

fruch: I rather not blindly set a timeout, without having data.

I'm just reverting this into what is/was working before, we don't have timeout on any other place we did rebuild

@soyacz is working on a way we could define timeouts relative to the data-set size (or other constraints), and then we can circle back to such place and do proper work of collecting data and giving estimates.

setting 1 hour is same as setting 5min, I have on idea if which case it gonna break.

fgelcer (approved):
LGTM

sdcm/nemesis.py
vponomaryov: As it is said by the @KnifeyMoloko there, it was part of the race fix:

```
This PR's commits address this and an additional potential source of racy behavior by: a) moving the log follower to be initiated before the trigger function is called and b) by introducing timeouts to the nodetool remoter calls.
```

@fruch are you sure that the above statement is false?

fruch: the nodetool command under question was decommission, **not** the rebuild commands.
the rebuild commands are outside of the parallelObejct call

so yes I'm sure

vponomaryov: O.k.

vponomaryov (approved):
LGTM

PR #5815, opened 2023-02-15
fix(K8S): use latest stable scylla-operator v1.8.0

vponomaryov:
If we do not specify scylla operator version for any local K8S (kind) test run then we get following incompatibility error:

    no matches for kind "PodDisruptionBudget" in version "policy/v1beta1"

The root cause for it is that we use K8S v1.25 where 'policy/v1beta1' API support was dropped.

So, use latest stable scylla-operator version which won't use this old API.


vponomaryov:
@fruch @soyacz 
Look at the checks, K8S tests passed. We can merge it.

PR #5817, opened 2023-02-15
feature(versions): add classes for scylla versions comparisons

vponomaryov:
In various cases we need to know scylla or scylla-operator version to make some desicions.
The structure of versions differs for GA and dev/rc versions. Also, it differs based on the source of version - binary or docker image.
We used to use `packaging.version.LegacyVersion` class for comparisons. But it got removed from the latest `packaging` package versions.

So, stop using that library and replace it with our own even more improved classes for comparisons of Scylla and Scylla operator versions.

Examples of usage:

    ComparableScyllaVersion("5.1.0~rc1") < "5.1.0"
    ComparableScyllaVersion("5.1.rc1") > "5.1.0-0.20220922.539a55e35"
    ComparableScyllaVersion("5.2.rc1") > "5.2.0-dev-0.20230109.08b3a9c786d9"

The same is done for Scylla Operator
in the ComparableScyllaOperatorVersion class where it's own specific version workarounds are implemented.

### DIFF
--- a/sdcm/utils/version_utils.py
+++ b/sdcm/utils/version_utils.py
@@ -125,6 +125,142 @@ def parse_scylla_version(version_to_parse: str) -> version.Version:
     return parse_version(major_version[0])
 
 
+class ComparableScyllaVersion:
+    """Accepts and compares known 'non-semver' and 'semver'-like Scylla versions."""
+
+    def __init__(self, version_string: str):
+        parsed_version = self.parse(version_string)
+        self.v_major = int(parsed_version[0])
+        self.v_minor = int(parsed_version[1])
+        self.v_patch = int(parsed_version[2])
+        self.v_pre_release = parsed_version[3] or ''
+        self.v_build = parsed_version[4] or ''
+
+    @staticmethod
+    def parse(version_string: str):
+        """Parse scylla-binary and scylla-docker-tag versions into a proper semver structure."""
+        # NOTE: remove 'with build-id' part if exists and other possible non-semver parts
+        _scylla_version = (version_string or '').split(" ")[0]
+
+        # NOTE: replace '~' which gets returned by the scylla binary
+        _scylla_version = _scylla_version.replace('~', '-')
+
+        # NOTE: remove docker-specific parts if version is taken from a docker tag
+        _scylla_version = _scylla_version.replace('-aarch64', '')
+        _scylla_version = _scylla_version.replace('-x86_64', '')
+
+        # NOTE: make short scylla version like '5.2' be correct semver string
+        _scylla_version_parts = _scylla_version.split('.')
+        if len(_scylla_version_parts) == 2:
+            _scylla_version = f"{_scylla_version}.0"
+        elif len(_scylla_version_parts) > 2 and re.search(
+                r"\D+", _scylla_version_parts[2].split("-")[0]):
+            _scylla_version = f"{_scylla_version_parts[0]}.{_scylla_version_parts[1]}.0-{_scylla_version_parts[2]}"
+            for part in _scylla_version_parts[3:]:
+                _scylla_version += f".{part}"
+
+        # NOTE: replace '-0' with 'dev-0', '-1' with 'dev-1' and so on
+        #       to match docker and scylla binary version structures correctly.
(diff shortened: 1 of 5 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/229/consoleFull

sdcm/utils/version_utils.py
```diff
@@ -125,6 +125,142 @@ def parse_scylla_version(version_to_parse: str) -> version.Version:
...
+
+    def __str__(self):
+        result = f"{self.v_major}.{self.v_minor}.{self.v_patch}"
+        if self.v_pre_release:
+            result += f"-{self.v_pre_release}"
+        if self.v_build:
+            result += f"+{self.v_build}"
+        return result
+
+    def _transform_to_comparable(self, other):
```
fruch, line 186: out of curiosity, did you tested this one with `packaging.version.Version`  ?

fruch:
@vponomaryov can we replace `parse_scylla_version` usages as well ?

sdcm/utils/version_utils.py
vponomaryov:
```
>>> from packaging.version import Version
>>> from packaging.version import LegacyVersion
>>> LegacyVersion("5.2.0") > LegacyVersion('5.2.0~dev')
False
>>> Version("5.2.0") > Version('5.2.0~dev')
  File "<stdin>", line 1, in <module>
  File "/home/kiparis/.local/lib/python3.10/site-packages/packaging/version.py", line 266, in __init__
    raise InvalidVersion(f"Invalid version: '{version}'")
packaging.version.InvalidVersion: Invalid version: '5.2.0~dev'
```

The introduced here classes return proper `True` result.

vponomaryov:
> @vponomaryov can we replace `parse_scylla_version` usages as well ?

I think yes.

fruch:
> > @vponomaryov can we replace `parse_scylla_version` usages as well ?
> 
> I think yes.

so let's do that

vponomaryov:
> > > @vponomaryov can we replace `parse_scylla_version` usages as well ?
> > 
> > 
> > I think yes.
> 
> so let's do that

I would prefer to have it as a separate PR. `parse_scylla_version` uses `packaging.version.Version` which is not deprecated and allows us to use latest `packaging` package version.
This PR, removes dependency on the `packaging.version.LegacyVersion`, which is already removed in the latest versions of the `packaging` package.

fruch:
> > > > @vponomaryov can we replace `parse_scylla_version` usages as well ?
> > > 
> > > 
> > > I think yes.
> > 
> > so let's do that
> 
> I would prefer to have it as a separate PR. `parse_scylla_version` uses `packaging.version.Version` which is not deprecated and allows us to use latest `packaging` package version.
> This PR, removes dependency on the `packaging.version.LegacyVersion`, which is already removed in the latest versions of the `packaging` package.
> 

O.k. do it in separate PR, 

Also we can unpin packaging after this one

fruch (approved):
LGTM

sdcm/utils/version_utils.py
fruch: This works ?


```python
version_utils.ComparableScyllaVersion("5.2.0") < Version("5.3.0)
```

vponomaryov: > ```python
> ComparableScyllaVersion
> ```

For now it accepts only `str` and the same class type objects.
So, following will work:
```
```

soyacz (approved):

vponomaryov:
> > > > > @vponomaryov can we replace `parse_scylla_version` usages as well ?
> > > > 
> > > > 
> > > > I think yes.
> > > 
> > > 
> > > so let's do that
> > 
> > 
> > I would prefer to have it as a separate PR. `parse_scylla_version` uses `packaging.version.Version` which is not deprecated and allows us to use latest `packaging` package version.
> > This PR, removes dependency on the `packaging.version.LegacyVersion`, which is already removed in the latest versions of the `packaging` package.
> 
> O.k. do it in separate PR,
> 
> Also we can unpin packaging after this one

https://github.com/scylladb/scylla-cluster-tests/pull/5833

Issue #5819, opened 2023-02-15
Issue when creating SCT runner for Azure


soyacz:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

During create sct runner stage, runner is not created. Jenkins log shows:
```
05:01:55  Creating 'sct-runner-1.6-instance-75d9e3db' VM in resource group SCT-75d9e3db-ae7b-45aa-a35e-d50dbb0db2f6-eastus-1...
05:01:55  Instance params: InstanceDefinition(name='sct-runner-1.6-instance-75d9e3db', image_id='/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/SCT_IMAGES-eastus/providers/Microsoft.Compute/galleries/SCT_IMAGES/images/sct-runner/versions/1.6.0', type='Standard_E2s_v3', user_name='ubuntu', tags={'RunByUser': 'timer', 'TestName': 'null', 'TestId': '75d9e3db-ae7b-45aa-a35e-d50dbb0db2f6', 'version': 'scylla-master', 'CreatedBy': 'SCT', 'JenkinsJobTag': 'jenkins-scylla-master-rolling-upgrade-rolling-upgrade-azure-image-test-3', 'NodeType': 'sct-runner', 'keep': '16', 'keep_action': 'terminate', 'UserName': 'ubuntu', 'bastion': 'true', 'launch_time': 'February 11, 2023, 04:01:26'}, arch=<VmArch.X86: 'x86_64'>, root_disk_size=80, data_disks=None)
05:01:55  Error when sending create vm request for VM sct-runner-1.6-instance-75d9e3db: (InvalidReservedTagValue) The tag names 'TestName' have reserved tag values. The reserved tag values are: 'null, (null)'.
05:01:55  Code: InvalidReservedTagValue
05:01:55  Message: The tag names 'TestName' have reserved tag values. The reserved tag values are: 'null, (null)'.
05:01:55  Traceback (most recent call last):
05:01:55    File "/home/jenkins/slave/workspace/scylla-master/rolling-upgrade/rolling-upgrade-azure-image-test/scylla-cluster-tests/./sct.py", line 1651, in <module>
05:01:55      cli.main(prog_name="hydra")
05:01:55    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
05:01:55      rv = self.invoke(ctx)
[...]

PR #5821, opened 2023-02-16
fix(pipeline): set empty test_name when not set

soyacz:
When `test_name` param is empty, toJson coverter converts value to `null`. Better to use empty string instead (e.g. null value is forbidden in Azure tags)

refs: #5819


fruch (approved):
LGTM

PR #5826, opened 2023-02-16
fix(_destroy_data_and_restart_scylla): continue inspite of lack of files

ShlomiBalalis:
In some cases, _destroy_data_and_restart_scylla will fail because there aren't enough data file to delete from non-system keyspaces. I added an except clause so that the method will continue in such cases all the same.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1022,7 +1022,19 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     @decorate_with_context(ignore_ycsb_connection_refused)
     def _destroy_data_and_restart_scylla(self):
-
+        def find_five_files_to_destroy():
+            all_files_to_destroy = []
+            try:
+                for _ in range(5):
+                    for _ in range(3):
+                        file_for_destroy = self._choose_file_for_destroy(tables)
+                        if file_for_destroy not in all_files_to_destroy:
+                            all_files_to_destroy.append(file_for_destroy)
+                            break
+            except NoFilesFoundToDestroy:
+                if not all_files_to_destroy:
+                    raise
+            return all_files_to_destroy
         ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=self.target_node, filter_empty_tables=False)
         if not ks_cfs:
             raise UnsupportedNemesis(
@@ -1048,12 +1060,11 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self.target_node.stop_scylla_server(verify_up=False, verify_down=True)
 
         try:
-
             # Remove 5 data files
-            for _ in range(5):
-
-                file_for_destroy = self._choose_file_for_destroy(tables)
-
+            all_files_to_destroy = find_five_files_to_destroy()
+            if len(all_files_to_destroy) < 5:
+                self.log.warning("Expected to find 5 files to destroy, but only found %s", len(all_files_to_destroy))
+            for file_for_destroy in all_files_to_destroy:
                 result = self.target_node.remoter.sudo('rm -f %s' % file_for_destroy)
                 if result.stderr:
                     raise FilesNotCorrupted('Files were not corrupted. CorruptThenRepair nemesis can\'t be run. '


fgelcer (approved):
LGTM, but was it tested?

ShlomiBalalis:
> LGTM, but was it tested?

Running: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/Shlomo/job/longevity-5gb-1h-CorruptThenRebuildMonkey-aws-test/

sdcm/nemesis.py
```diff
@@ -1022,7 +1022,19 @@ def _choose_file_for_destroy(self, ks_cfs):
 
     @decorate_with_context(ignore_ycsb_connection_refused)
     def _destroy_data_and_restart_scylla(self):
-
+        def find_five_files_to_destroy():
+            all_files_to_destroy = []
+            try:
+                for _ in range(5):
```
fruch, line 1028: why do we need double loop here ? it's really not clear

sdcm/nemesis.py
```diff
@@ -1022,7 +1022,19 @@ def _choose_file_for_destroy(self, ks_cfs):
 
     @decorate_with_context(ignore_ycsb_connection_refused)
     def _destroy_data_and_restart_scylla(self):
-
+        def find_five_files_to_destroy():
```
fruch, line 1025: there no sleep in the loop here, so it might run in quick secession and all might fail to retrieve a file 
I would think some wait is needed between the trials, and once there's enough files, this loop can stop.

sdcm/nemesis.py
ShlomiBalalis: There's a chance that `_choose_file_for_destroy` will pick a file that is already in the list. The second loop is a retry/re-raffle.

fruch: to avoid duplication I would use something like `set`

sdcm/nemesis.py
ShlomiBalalis: `_choose_file_for_destroy` already has a retry, so there's that.
Also, I don't see why you would need to wait between the trials. It's not like more are going to be created since scylla is down at this point.

sdcm/nemesis.py
roydahan: It's a retry loop.

roydahan: It's just not very readable, since he use the break if it's not in the list.

roydahan: but I think it's good enough.

sdcm/nemesis.py
roydahan: We do this deletions when the scylla node is already stopped, so there is no reason to wait.

roydahan (approved):

sdcm/nemesis.py
fruch: I'm still not following what this code is trying to solve, we have better ways of retries.

And I don't understand which problem it's trying to solve.

roydahan: I know we have better ways, I just explained what he tries to do here.
The PR is trying to "solve" or "ease" the requirement of 5 files to be destroyed, so we won't fail the test with ERROR in case we found less than 5 files.

fgelcer:
@scylladb/qa-maintainers please backport

fruch:
> @scylladb/qa-maintainers please backport

it's already there

fgelcer:
no, it is not... there is a similar commit with name `fix(_destroy_data_and_restart_scylla): fail only when no file is found`

this one is called `fix(_destroy_data_and_restart_scylla): continue inspite of lack of files`

fruch:
> no, it is not... there is a similar commit with name `fix(_destroy_data_and_restart_scylla): fail only when no file is found`
> 
> this one is called `fix(_destroy_data_and_restart_scylla): continue inspite of lack of files`

this is the commit, and it's in the release (why the PR and the commit doesn't have the same title, that's a good question..., but still it's there)
https://github.com/scylladb/scylla-cluster-tests/pull/5826/commits/1dfea9f24cee51cb6d7bcf9bdcc2fbe30df64fb3

PR #5827, opened 2023-02-16
fix(gce_download): Skip first element by name upon download

aleksbykov:
list_container_objects method in gce_download_dir returns list of objects in gce bucket. Previously first returned element
 was folder itself. Latest run shows that now method return only objects
in bucket. Current implementation could skip first object or if only one object in bucket don't download anything and later fail the job.

Add checks element by name and filter out objects with non-relevant names

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -2119,8 +2119,9 @@ def gce_download_dir(bucket, path, target):
 
     container = driver.get_container(container_name=bucket)
     dir_listing = driver.list_container_objects(container, ex_prefix=path)
-    # first element is gce dir itself. Skip it.
-    for obj in dir_listing[1:]:
+    for obj in dir_listing:
+        if obj.name in [".", "..", path]:
+            continue
         rel_path = obj.name[len(path):]
         local_file_path = os.path.join(target, rel_path)
 


sdcm/utils/common.py
```diff
@@ -2120,7 +2120,9 @@ def gce_download_dir(bucket, path, target):
     container = driver.get_container(container_name=bucket)
     dir_listing = driver.list_container_objects(container, ex_prefix=path)
     # first element is gce dir itself. Skip it.
```
soyacz, line 2122: this comment started to lie

fruch: Sound like we should have integration test for this function

fruch:
@aleksbykov update us when you're tested this fix

sdcm/utils/common.py
aleksbykov: remove comment

aleksbykov:
> @aleksbykov update us when you're tested this fix

Already passed this point in this job: 
https://jenkins.scylladb.com/view/master/job/scylla-master/job/reproducers/job/reproduce-12249-large-partition-200k-test/10

fruch:
> > @aleksbykov update us when you're tested this fix
> 
> @fruch , Already passed this point in this job: 
> https://jenkins.scylladb.com/view/master/job/scylla-master/job/reproducers/job/reproduce-12249-large-partition-200k-test/10
> 

So why is it draft ?

aleksbykov:
> So why is it draft ?

Wait while checks passed. I moved to Ready for Review

fruch (approved):

PR #5829, opened 2023-02-16
fix(get_supported_scylla_base_versions): adding 2023.1

fgelcer:
we have a map, of which OSS version relates to an
enterprise, so we could calculate the base version when running rolling upgrade.
up until now, the base version was only 2022.2


fruch (approved):
LGTM

Even though I think this map should be reversed, and new addition should be at the end.

fgelcer:
> LGTM
> 
> Even though I think this map should be reversed, and new addition should be at the end.

you may be correct, but the whole format is newer ones are added to left... and the right ones are deprecated

PR #5834, opened 2023-02-20
Manager restore task large scenarios

ShlomiBalalis:


rayakurl:
This PR is a draft until https://github.com/scylladb/scylla-manager/issues/3316 is backported into manager 3.1

rayakurl:
@ShlomiBalalis - is this PR ready for review?

ShlomiBalalis:
> @ShlomiBalalis - is this PR ready for review?

Yes!

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
```diff
@@ -0,0 +1,19 @@
+test_duration: 2100
+
+round_robin: true
+
+instance_type_db: 'i3en.3xlarge'
+instance_type_loader: 'c5.2xlarge'
```
rayakurl, line 6: Didn't we move to i4i family?

rayakurl (dismissed):
approved, just a small question

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
```diff
@@ -0,0 +1,19 @@
+test_duration: 2100
+
+round_robin: true
+
+instance_type_db: 'i3en.3xlarge'
```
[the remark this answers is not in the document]
ShlomiBalalis: Not on all runs as far as I know. It was even Roy that recommended that instance type in the first place

mgmt_cli_test.py
```diff
@@ -199,17 +200,27 @@ def verify_backup_success(self, mgr_cluster, backup_task, keyspace_name='keyspac
...
                                                  keyspace_and_table_list=per_keyspace_tables_dict)
 
-    def restore_schema_with_task(self, mgr_cluster, backup_task, timeout):
+    def restore_schema_with_task(self, mgr_cluster, backup_task, timeout, location_list=None):
         snapshot_tag = backup_task.get_snapshot_tag()
-        restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=self.locations,
+        self.restore_schema_with_task_from_snapshot_tag(mgr_cluster=mgr_cluster, snapshot_tag=snapshot_tag,
+                                                        timeout=timeout, location_list=location_list)
+
+    def restore_schema_with_task_from_snapshot_tag(self, mgr_cluster, snapshot_tag, timeout, location_list=None):
```
roydahan, line 208: From the name it looks like it restores only the schema and not the data.

mgmt_cli_test.py
```diff
@@ -459,6 +470,55 @@ def get_email_data(self):
...
+        self.restore_data_with_task(mgr_cluster=mgr_cluster, backup_task=backup_task, timeout=110000)
+        self.run_verification_read_stress()
+
+    @staticmethod
+    def _get_persistent_snapshots():
+        with open("defaults/manager_persistent_snapshots.yaml", encoding="utf-8") as mgmt_snapshot_yaml:
+            persistent_manager_snapshots_dict = yaml.safe_load(mgmt_snapshot_yaml)
+        return persistent_manager_snapshots_dict
+
+    def test_restore_multiple_keyspaces(self):
```
roydahan, line 493: I don't understand the purpose of the test.
From its name it may be sound that the purpose is to test a restore of multiple keyspaces.
In that case, I would expect the test to (backup?) multiple keyspace (> 10, 50?, 100?) and try to restore them.
(It could also restore the same one to multiple new keyspaces)

However, from the code it looks like it just restore 2 keyspaces that were pre-backup and stored in AWS.
And from what I know, you do it mostly because you needed this as a way to restore to a clean cluster including the schema).

mgmt_cli_test.py
```diff
@@ -459,6 +470,55 @@ def get_email_data(self):
...
+        manager_tool = mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0])
+        mgr_cluster = manager_tool.get_cluster(cluster_name=self.CLUSTER_NAME) \
+            or manager_tool.add_cluster(name=self.CLUSTER_NAME, db_cluster=self.db_cluster,
+                                        auth_token=self.monitors.mgmt_auth_token)
+        if self.params.get('cluster_backend') != 'aws':
+            self.log.error("Test supports only AWS ATM")
+            return
```
roydahan, line 500: Why?
Even with the current implementation, couldn't you restore from AWS to GCP / Azure?

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
```diff
@@ -0,0 +1,28 @@
+test_duration: 4320
+
```
roydahan, line 2: General question, do we delete these backups to S3 at some point?
If so, what take care of it? (teardown? , do we have a scheduled job for it?).

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
```diff
@@ -0,0 +1,28 @@
...
+prepare_write_cmd: ["cassandra-stress write cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..1342177280",
```
roydahan, line 3: I think it would make more sense to split this test into 2 tests, one test that creates the keyspace and do the backup.
(with option to persist the backup for later usages (like you do in tests above).

And another test, that only restore from a previous backup to a clean cluster.

This will improve things by:
1. Allow you to run only one test if you need (e.g. problem with backup, so you run only this one)
2. It will cut the time to run the test by ("half") because you can run them both in parallel.
3. In most cases we won't even run the 5tb backup test, but only the restore.
4. the restore test can run on bigger cluster, while the backup is enough to run on single node cluster.

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
```diff
@@ -0,0 +1,28 @@
...
+                    "cassandra-stress write cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1342177281..2684354560",
+                    "cassandra-stress write cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=2684354561..4026531840",
```
roydahan, line 5: replace all cl=ONE with cl=ALL.
Now it "doesn't matter" but if one later decide to change the test to more than one node, it will matter and it's better one won't fall in this trap.

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
```diff
@@ -0,0 +1,28 @@
...
+                    "cassandra-stress write cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=4026531841..5368709120"]
+
+stress_read_cmd: ["cassandra-stress read cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..1342177280",
+                  "cassandra-stress read cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1342177281..2684354560",
```
roydahan, line 9: In this case it's actually good to leave it cl=ONE even if we increase the cluster size (and we should) for the splitted test.
Because reading with cl=ONE can catch data loss / data consistency issues.

roydahan, line 9: In this case it's actually good to leave it cl=ONE even if we increase the cluster size (and we should) for the splitted test.
Because reading with cl=ONE can catch data loss / data consistency issues.

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
```diff
@@ -0,0 +1,28 @@
...
+                  "cassandra-stress read cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=2684354561..4026531840",
```
roydahan, line 10: For restoring to bigger cluster, you'll need to take care of the RF which is now ONE.
What is the UX that the manager suggests for this use case?
(E.g user backed up a cluster with RF=1, then wants to restore it to a cluster with RF=3).

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
```diff
@@ -0,0 +1,28 @@
+test_duration: 4320
+
+prepare_write_cmd: ["cassandra-stress write cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..1342177280",
+                    "cassandra-stress write cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1342177281..2684354560",
```
roydahan, line 4: BTW, isn't "rate threads=50" too low?
Do you have a monitor from this test? we need to make sure we utilize it to the max throughput in order to save time.

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
```diff
@@ -0,0 +1,28 @@
...
+
+stress_read_cmd: ["cassandra-stress read cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..1342177280",
+                  "cassandra-stress read cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1342177281..2684354560",
+                  "cassandra-stress read cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=2684354561..4026531840",
+                  "cassandra-stress read cl=ONE n=1342177280 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=4026531841..5368709120"]
+
+round_robin: true
+
+instance_type_db: 'i3en.6xlarge'
+instance_type_loader: 'c5.2xlarge'
```
roydahan, line 16: probably too big.

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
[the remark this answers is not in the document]
roydahan: This instance type is needed only for the large scale dataset (5TB).
What is the dataset size you test here? 2TB?
If so, it can be reduced to significantly, especially if you use 5 db nodes.
(calculate exactly what you need).

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
```diff
@@ -0,0 +1,19 @@
...
+instance_type_loader: 'c5.2xlarge'
+nemesis_class_name: NodeTerminateAndReplace
```
roydahan, line 7: Do you use nemesis in this test? where exactly?

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
```diff
@@ -0,0 +1,19 @@
...
+
+region_name: us-east-1
```
roydahan, line 9: Add a comment here that the region is specified because the backup bucket is in the same region.

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
```diff
@@ -0,0 +1,19 @@
...
+n_db_nodes: 5
+n_loaders: 1
+n_monitor_nodes: 1
+
+post_behavior_db_nodes: "destroy"
+post_behavior_loader_nodes: "destroy"
+post_behavior_monitor_nodes: "destroy"
+
+user_prefix: manager-regression
+space_node_threshold: 6442
```
roydahan, line 19: remove

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
```diff
@@ -0,0 +1,19 @@
+test_duration: 2100
+
+round_robin: true
```
roydahan, line 3: ?

roydahan (changes requested):

mgmt_cli_test.py
ShlomiBalalis: Because it does...
the restore-schema option only restores the schema, without the data.

mgmt_cli_test.py
ShlomiBalalis: Access to our bucket is limited by the IAM role, so I'm not sure we can even access the bucket from azure/gcp instances in the first place. Moreover, I'm sure the actual restore will take longer so we'd have to reconsider the timeouts that were already set, and for that I'll need to rerun the test to see the required times. So for now I chose to leave it at that.

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
ShlomiBalalis: The life cycle rule defined in the bucket: https://s3.console.aws.amazon.com/s3/buckets/manager-backup-tests-us-east-1?region=us-east-1&tab=management

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
ShlomiBalalis: Right, this was a remnant from a older version of the test. Removing.

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
ShlomiBalalis: When you restore a schema all parameters like the replication factor will remain the same as they were before. You'd need to modify the RF manually after restoring the schema.

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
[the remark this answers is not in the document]
ShlomiBalalis: The dataset will fit on i3en,2xlarge, yes, but the run time is increased significally as a result - a restore that previously took 13.5 hours now takes over 18.5. I'm not sure it's worth it.

mgmt_cli_test.py
roydahan: But don't we have tests for backup also in GCP and Azure?
If not, why would we even need to check the backend?

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
roydahan: right.
And do you do this?

defaults/manager_persistent_snapshots.yaml
```diff
@@ -0,0 +1,18 @@
...
+      expected_timeout: 1800  # 30 minutes
+      keyspace_name: "10gb_sizetiered"
+      confirmation_stress_command: "cassandra-stress read cl=QUORUM n=10485760 -schema 'keyspace=10gb_sizetiered replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..10485760"
+    100:
+      snapshot_tag: 'sm_20230223130733UTC'
+      expected_timeout: 9000  # 150 minutes
+      keyspace_name: "100gb_sizetiered"
+      confirmation_stress_command: "cassandra-stress read cl=QUORUM n=104857600 -schema 'keyspace=100gb_sizetiered replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..104857600"
+    2048:
+      snapshot_tag: 'sm_20230226074656UTC'
```
roydahan, line 14: In a followup PR:
Add another backup of 1TB that is taken with ICS.
Taken with RF=3 and a NetworkTopology in the schema.
When writing the dataset, use only one loader (you can ask @fgelcer for commands).

Then, add a test that restores this 1TB into a 9 nodes cluster (smaller ones - e.g. i4i.xlarge).
When restoring the data, measure the time and report it to the test result of how long it took to restore.

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
roydahan, line 1: AFAIK, there is an issue to backup 5TB dataset.
If this issue, isn't planned to be fixed for 3.1, change the test to 4TB and comment that on the issue that his was changed.

Also, report from the test to the test output (as events the time it took to backup, restore and verify).

roydahan (changes requested):

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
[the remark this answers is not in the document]
ShlomiBalalis: I kept the instance type at 3xlarge, I don't believe the extra run time is worth it

roydahan:
ping

mgmt_cli_test.py
ShlomiBalalis: We have a restore test in gce and azure. This specific test requires accessing an s3 bucket, which is only accessible from amazon instances. I added this check as a reminder

ShlomiBalalis:
> ping

I believe I answered everything. Did I miss something?

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
roydahan, line 1: Do we want to change the name from 5TB to 4TB?
Also the jenkins pipeline name.

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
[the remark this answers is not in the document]
roydahan: Leave it for now.
Please share somwhere (in the test plan probably) the runtimes we know.

roydahan:
Just small question regarding the 4TB vs 5TB.
Also, if we have 4TB dataset ready, I suggest you'll upload it to the set of datasets and the restore test will run only this the next time (without populating and backup).

test-cases/manager/manager-backup-and-restore-5TB-single-node.yaml
ShlomiBalalis: Done

test-cases/manager/manager-multiple-restores-schema-and-data.yaml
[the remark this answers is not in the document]
ShlomiBalalis: Added to the test plan

test-cases/manager/manager-backup-and-restore-4TB-single-node.yaml
```diff
@@ -0,0 +1,34 @@
...
+
+# Using cl=ALL opposed to cl=ONE so that in the future, when we will use multi node setup, the stress will work all the same.
+prepare_write_cmd: ["cassandra-stress write cl=ONE n=1006632960 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..1006632960",
+                    "cassandra-stress write cl=ONE n=1006632960 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1006632961..2013265920",
+                    "cassandra-stress write cl=ONE n=1006632960 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=2013265921..3019898880",
+                    "cassandra-stress write cl=ONE n=1006632960 -schema 'replication(factor=1) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=3019898881..4026531840"]
```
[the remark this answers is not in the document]
ShlomiBalalis: Yes, sorry. Accidentally reverted it when I changed the data size to 4TB. Fixed.

roydahan (approved):

roydahan (approved):

ShlomiBalalis:
@roydahan Can you please backport?

ShlomiBalalis:
@scylladb/qa-maintainers Can anyone backport?

roydahan:
@ShlomiBalalis where are the jenkins jobs of these tests?
Neither me nor @fgelcer could find them on Thursday.

PR #5836, opened 2023-02-20
test(create release jobs): upgrade with raft enabled

fgelcer:
as of now, RAFT is disabled for all upgrade jobs,
so, to test the RAFT feature on upgrade jobs,
there were few jobs introduced to test it, but their jobs are not yet being created automatically, and
will now have its own group, and jobs, and later,
should be added to the upgrade triggers as well.

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -1369,6 +1369,7 @@ def create_test_release_jobs(branch, username, password, sct_branch, sct_repo):
         ('raft', 'SCT Raft Experimental'),
         ('disabled_raft', 'SCT ConsistentClusterManagement disabled'),
         ('nemesis', 'SCT Individual Nemesis'),
+        ('upgrade-with-raft', 'SCT Rolling Upgrades With Raft Enabled'),
     ]:
         server.create_directory(name=group_name, display_name=group_desc)
 
@@ -1389,6 +1390,9 @@ def create_test_release_jobs(branch, username, password, sct_branch, sct_repo):
                 server.create_pipeline_job(jenkins_file, group_name)
         if group_name == 'disabled_raft':
             for jenkins_file in glob.glob(f'{base_path}/consistent_cluster_management_disabled/*'):
+                server.create_pipeline_job(jenkins_file, group_name)
+        if group_name == 'upgrade-with-raft':
+            for jenkins_file in glob.glob(f'{base_path}/{group_name}/*.jenkinsfile'):
                 server.create_pipeline_job(jenkins_file, group_name)
         if group_name == 'nemesis':
             for jenkins_file in glob.glob(f'{base_path}/nemesis/*'):


sct.py
```diff
@@ -1369,6 +1369,7 @@ def create_test_release_jobs(branch, username, password, sct_branch, sct_repo):
         ('raft', 'SCT Raft Experimental'),
         ('disabled_raft', 'SCT ConsistentClusterManagement disabled'),
         ('nemesis', 'SCT Individual Nemesis'),
+        ('upgrade-with-raft', 'SCT Rolling Upgrades With Raft Enabled'),
```
vponomaryov, line 1372: We have `rolling-upgrade-raft-disabled` on the line 1361 and `rolling-upgrade` on the line 1360.
How this new one different than the one on line 1360?

Also, maybe rename it to have the same template? By adding `rolling-` prefix?

fgelcer: you are right, they will be duplicated... i will add a better filtering for it, so we can have them split into right groups

fgelcer: ok, now i'm reading it again, and we don't have `rolling-upgrade-raft-disabled`, as this is the default behavior, but we do have other tests with `disabled_raft` (some longevity jobs)...
on `rolling-upgrade` we will have all the jobs that are `rolling-upgrade`, and reside inside the `base_path` to the `jenkinsfile`, but the upgrade jobs with raft enabled are placed in a separated folder inside the `base_path`, so there is no duplication, nor issues on the jobs creation.
one could complain that these jobs could go into `raft` group, but AFAICT we want these rolling upgrade to be apart intentionally, as one day in future versions, we will reverse the logic, and raft will be enabled by default, and these tests will run with raft disabled.

also, one could argue that we want to call the group (folder) `rolling-upgrade-with-raft`, and that would be valid, but too long for not enough, but the jobs inside of it, will be called `rolling-upgrade....` as all other rolling upgrade jobs, just in a different Jenkins path.

@vponomaryov , hope i answered your comments above.

vponomaryov: And none of jobs were created due to absent parent dir.

sct.py
```diff
@@ -1390,6 +1391,9 @@ def create_test_release_jobs(branch, username, password, sct_branch, sct_repo):
         if group_name == 'disabled_raft':
             for jenkins_file in glob.glob(f'{base_path}/consistent_cluster_management_disabled/*'):
                 server.create_pipeline_job(jenkins_file, group_name)
+        if group_name == 'upgrade-with-raft':
+            for jenkins_file in glob.glob(f'{base_path}/{group_name}/*.jenkinsfile'):
```
vponomaryov, line 1395: Looks like it fails.
Need to precreate `{base_path}/{group_name}` dirs first and only then jenkins files.

sct.py
fgelcer: yes, of course, the idea was to see what the code intended to do, and not to really do

sct.py
fgelcer: it was intentional to have it failing in the link above... it was to see when the code was calling for the files i want

roydahan (approved):

PR #5837, opened 2023-02-21
improvement(k8s-eks): cleanup OIDC providers

vponomaryov:
After merge of the https://github.com/scylladb/scylla-cluster-tests/pull/5658 we create OIDC providers running on EKS backend. 
And the problem with it is that we do not cleanup created OIDC providers. 
It doesn't get automatically deleted with an associated EKS cluster too.

So, delete it explicitly to avoid limits violations.

### DIFF
--- a/sdcm/utils/aws_utils.py
+++ b/sdcm/utils/aws_utils.py
@@ -41,6 +41,10 @@ class EksClusterCleanupMixin:
     def ec2_client(self):
         return boto3.client('ec2', region_name=self.region_name)
 
+    @cached_property
+    def iam_client(self):
+        return boto3.client('iam', region_name=self.region_name)
+
     @property
     def owned_object_tag_name(self):
         return f'kubernetes.io/cluster/{self.short_cluster_name}'
@@ -94,7 +98,8 @@ class EksClusterCleanupMixin:
             # but before retrying of the destroy_nodegroups
             self.destroy_attached_security_groups()
             if not self.cluster_exists:
-                return
+                break
+        self.destroy_oidc_provider()
 
     def check_if_all_network_interfaces_detached(self, sg_id):
         for interface_description in self.ec2_client.describe_network_interfaces(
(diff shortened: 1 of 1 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-provision-eks-cluster/113/consoleFull

soyacz (approved):
In case destroy oidc provider fails or delete cluster fails after deletion, it will leave oidc provider until manual removal. I don't think it will be frequent issue, but if you know some nice&easy solution better to implement it right away.

fruch (approved):
LGTM

fruch:
> In case destroy oidc provider fails or delete cluster fails after deletion, it will leave oidc provider until manual removal. I don't think it will be frequent issue, but if you know some nice&easy solution better to implement it right away.

I think we need to add this to the task we have for cleanups for k8s resources.

vponomaryov:
> In case destroy oidc provider fails or delete cluster fails after deletion, it will leave oidc provider until manual removal. I don't think it will be frequent issue, but if you know some nice&easy solution better to implement it right away.

All the preceding cleanups are in `try`-blocks. So, deletion of `OIDC provider`s is reachable in anyway. Even if EKS cluster deletion fails.
`oidc provider` may fail, in current case only, if AWS platform API is unavailable or something like this.
So, if it is platform problem, then you won't be able to delete anyway.

soyacz:
> > In case destroy oidc provider fails or delete cluster fails after deletion, it will leave oidc provider until manual removal. I don't think it will be frequent issue, but if you know some nice&easy solution better to implement it right away.
> 
> All the preceding cleanups are in `try`-blocks. So, deletion of `OIDC provider`s is reachable in anyway. Even if EKS cluster deletion fails. `oidc provider` may fail, in current case only, if AWS platform API is unavailable or something like this. So, if it is platform problem, then you won't be able to delete it anyway.

Yes, that may be the case. So one idea could be to add creation_time and duration tags so we could filter not only by cluster-name and make decision even without it.

PR #5839, opened 2023-02-21
fix(instance_root_disk_size): lower duration requirement for extra space

ShlomiBalalis:
Previously, we increased the sct runner's disk by 40GB if a scenario was at lease 3 days long. However, it seems that even shorter runs require an additional disk space, so I altered instance_root_disk_size so that the additional disk space will be added on every scenario that is longer than 36 hours

### DIFF
--- a/sdcm/sct_runner.py
+++ b/sdcm/sct_runner.py
@@ -155,8 +155,8 @@ class SctRunner(ABC):
     @staticmethod
     def instance_root_disk_size(test_duration) -> int:
         disk_size = 80  # GB
-        if test_duration and test_duration > 3 * 24 * 60:  # 3 days
-            # add 40G more space for jobs which test_duration is longer than 3 days
+        if test_duration and test_duration > 1.5 * 24 * 60:  # 1.5 days
+            # add 40G more space for jobs which test_duration is longer than 1.5 days
             return disk_size + 40
         return disk_size
 


ShlomiBalalis:
The scenario that required this change was longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis-test, whose latest run failed because the sct runner ran out of space.

vponomaryov (approved):
LGTM

roydahan:
@ShlomiBalalis I'll accept that as a quick fix for now, but
You need to check what bloated the logs in your case, because previously that disk space was good enough.
If you see same log messages repeating themselves too much, you should report a scylla issue for it.

roydahan (approved):

yarongilor:
This might also be required for the longevity of multi-keyspace-60h.
The sct-runner got disconnected in: https://argus.scylladb.com/test/86571bb4-7ca9-45f5-9e8f-0315da77b786/runs?additionalRuns[]=1886ff81-68c5-4404-b077-a9f9d7997aa6

Happened in 5.2.
@scylladb/qa-maintainers , can you please backport?

soyacz:
@roydahan do you approve backport?

I think we could implement a thread in sct that monitors disk usage, and if reaches threshold (e.g. 5GB free left) request to cloud disk resize and resize it in OS. That dynamic disk size mgmt would help us in all the cases without tricks with timings.

fgelcer:
> @roydahan do you approve backport?
> 
> I think we could implement a thread in sct that monitors disk usage, and if reaches threshold (e.g. 5GB free left) request to cloud disk resize and resize it in OS. That dynamic disk size mgmt would help us in all the cases without tricks with timings.

but in other hand, if we suddenly have a flood of logs, we may be increasing disk size by too much, before failing, and perhaps not everyone will check that, and the issue can stay undercover for long time

fruch:
> @roydahan do you approve backport?

Yes I think it's safe to backport 

> 
> I think we could implement a thread in sct that monitors disk usage, and if reaches threshold (e.g. 5GB free left) request to cloud disk resize and resize it in OS. That dynamic disk size mgmt would help us in all the cases without tricks with timings.

roydahan:
> @roydahan do you approve backport?
> 
> I think we could implement a thread in sct that monitors disk usage, and if reaches threshold (e.g. 5GB free left) request to cloud disk resize and resize it in OS. That dynamic disk size mgmt would help us in all the cases without tricks with timings.

Sounds like an overkill, we can live with the few more GBs we need in advance.

PR #5841, opened 2023-02-22
test(nemesis): added create index nemesis

soyacz:
Currently we test indexes by having some c-s profiles with MV's. Indexes are created along with tables and are populated incrementally. We miss the scenario, where an user decides to create index after some time, when table is full of data.

This commit introduces new nemesis covering this scenario. When executed, gets a table with the most partitions and creates index on some random regular or static column. Waits for creation and verify if index can be used in a query without ALLOW FILTERING option without any issue. Drops index after nemesis to not change schema permanently.

refs: https://github.com/scylladb/qa-tasks/issues/1011

### DIFF
--- a/data_dir/nemesis.yml
+++ b/data_dir/nemesis.yml
@@ -8,6 +8,7 @@
   - networking = False
   - kubernetes = True
   - limited = True
+  - schema_changes = True
   - free_tier_set = True
 - disrupt_add_remove_dc:
   - disruptive = False
@@ -17,6 +18,10 @@
   - topology_changes = True
 - disrupt_corrupt_then_scrub:
   - disruptive = False
+- disrupt_create_index:
+  - disruptive = False
+  - schema_changes = True
+  - free_tier_set = True
 - disrupt_decommission_streaming_err:
   - disruptive = True
   - topology_changes = True
@@ -54,6 +59,7 @@
 - disrupt_grow_shrink_new_rack:
   - disruptive = True
   - kubernetes = True
+  - config_changes = True
 - disrupt_hard_reboot_node:
   - disruptive = True
   - kubernetes = True
@@ -61,6 +67,7 @@
   - free_tier_set = True
 - disrupt_hot_reloading_internode_certificate:
   - disruptive = False
+  - config_changes = True
 - disrupt_increase_shares_by_attach_another_sl_during_load:
   - disruptive = False
   - sla = True
@@ -97,6 +104,7 @@
   - disruptive = False
   - kubernetes = True
   - limited = True
+  - schema_changes = True
   - free_tier_set = True
 - disrupt_multiple_hard_reboot_node:
   - disruptive = True
@@ -183,14 +191,17 @@
   - sla = True
 - disrupt_resetlocalschema:
   - disruptive = False
+  - config_changes = True
   - free_tier_set = True
 - disrupt_restart_with_resharding:
   - disruptive = True
   - kubernetes = True
   - topology_changes = True
+  - config_changes = True
 - disrupt_rolling_config_change_internode_compression:
   - disruptive = True
   - full_cluster_restart = True
+  - config_changes = True
 - disrupt_rolling_restart_cluster:
   - disruptive = True
   - kubernetes = True
@@ -198,6 +209,7 @@
 - disrupt_rolling_restart_cluster:
   - disruptive = True
   - kubernetes = True
+  - config_changes = True
   - free_tier_set = True
 - disrupt_run_cdcstressor_tool:
   - disruptive = False
@@ -246,6 +258,7 @@
   - limited = True
 - disrupt_switch_between_password_authenticator_and_saslauthd_authenticator_and_back:
   - disruptive = True
+  - config_changes = True
 - disrupt_terminate_and_replace_node:
   - disruptive = True
   - kubernetes = False
@@ -258,13 +271,17 @@
   - kubernetes = True
 - disrupt_toggle_cdc_feature_properties_on_table:
   - disruptive = False
+  - schema_changes = True
+  - config_changes = True
   - free_tier_set = True
 - disrupt_toggle_table_gc_mode:
   - kubernetes = True
   - disruptive = False
+  - schema_changes = True
   - free_tier_set = True
 - disrupt_toggle_table_ics:
   - kubernetes = True
+  - schema_changes = True
   - free_tier_set = True
 - disrupt_truncate:
   - disruptive = False
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -35,7 +35,7 @@ from concurrent.futures import ThreadPoolExecutor
 from threading import Lock
 from types import MethodType  # pylint: disable=no-name-in-module
 
-from cassandra import ConsistencyLevel
+from cassandra import ConsistencyLevel, InvalidRequest
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 from invoke import UnexpectedExit
 from elasticsearch.exceptions import ConnectionTimeout as ElasticSearchConnectionTimeout
@@ -87,7 +87,7 @@ from sdcm.utils.common import (get_db_tables, generate_random_string,
                                update_certificates, reach_enospc_on_node, clean_enospc_on_node,
                                parse_nodetool_listsnapshots,
                                update_authenticator, ParallelObject,
-                               ParallelObjectResult)
+                               ParallelObjectResult, sleep_for_percent_of_duration)
 from sdcm.utils.compaction_ops import CompactionOps, StartStopCompactionArgs
 from sdcm.utils.decorators import retrying, latency_calculator_decorator
 from sdcm.utils.decorators import timeout as timeout_decor
@@ -99,6 +99,8 @@ from sdcm.utils.k8s import (
 from sdcm.utils.k8s.chaos_mesh import MemoryStressExperiment, IOFaultChaosExperiment, DiskError
 from sdcm.utils.ldap import SASLAUTHD_AUTHENTICATOR, LdapServerType
 from sdcm.utils.loader_utils import DEFAULT_USER, DEFAULT_USER_PASSWORD, SERVICE_LEVEL_NAME_TEMPLATE
+from sdcm.utils.nemesis_utils.indexes import get_random_column_name, create_index, \
+    wait_for_index_to_be_built, verify_query_by_index_works, drop_index
 from sdcm.utils.replication_strategy_utils import temporary_replication_strategy_setter, \
     NetworkTopologyReplicationStrategy, ReplicationStrategy, SimpleReplicationStrategy
 from sdcm.utils.sstable.load_utils import SstableLoadUtils
@@ -4098,6 +4100,31 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                                                   for error in error_events if error)
                                         )
 
+    def disrupt_create_index(self):
+        """
+        Create index on a random column (regular or static) of a table with the most number of partitions and wait until it gets build.
+        Then verify it can be used in a query. Finally, drop the index.
+        """
+        with self.cluster.cql_connection_patient(self.target_node, connect_timeout=300) as session:
+            ks_cf_list = self.cluster.get_non_system_ks_cf_list(self.target_node, filter_out_mv=True)
+            if not ks_cf_list:
+                raise UnsupportedNemesis("No table found to create index on")
+            ks, cf = random.choice(ks_cf_list).split('.')
+            column = get_random_column_name(session, ks, cf)
+            try:
+                index_name = create_index(session, ks, cf, column)
+            except InvalidRequest as exc:
+                LOGGER.warning(exc)
+                raise UnsupportedNemesis(  # pylint: disable=raise-missing-from
+                    "Tried to create already existing index. See log for details")
+            try:
+                wait_for_index_to_be_built(self.target_node, ks, index_name, timeout=7200)
+                verify_query_by_index_works(session, ks, cf, column)
+                sleep_for_percent_of_duration(self.tester.test_duration * 60, percent=1,
+                                              min_duration=300, max_duration=2400)
+            finally:
+                drop_index(session, ks, index_name)
+
 
 def disrupt_method_wrapper(method, is_exclusive=False):  # pylint: disable=too-many-statements
     """
@@ -5466,3 +5493,13 @@ class SlaNemeses(Nemesis):
 
     def disrupt(self):
         self.call_random_disrupt_method(disrupt_methods=self.disrupt_methods_list)
+
+
+class CreateIndexNemesis(Nemesis):
+
+    disruptive = False
+    schema_changes = True
+    free_tier_set = True
+
+    def disrupt(self):
+        self.disrupt_create_index()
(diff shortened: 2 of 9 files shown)


vponomaryov:
@soyacz look at unit tests failure, it has cross-imports.

soyacz:
> @soyacz look at unit tests failure, it has cross-imports.

I'm still testing it, I'll update with job when confirmed fully working.

fgelcer:
@yarongilor , it may be relevant to something you wanted to do

yarongilor:
> @yarongilor , it may be relevant to something you wanted to do

@fgelcer , @soyacz , @roydahan , this PR may have quite a similar coverage to another new initiated PR of https://github.com/scylladb/scylla-cluster-tests/pull/5852

So perhaps we can consider adding some missing coverage in this PR or any other idea you think of.
The main missing coverage in this PR is taking a node down prior to creating an index, then take it up afterwards.

soyacz:
> > @yarongilor , it may be relevant to something you wanted to do
> 
> @fgelcer , @soyacz , @roydahan , this PR may have quite a similar coverage to another new initiated PR of #5852
> 
> So perhaps we can consider adding some missing coverage in this PR or any other idea you think of. The main missing coverage in this PR is taking a node down prior to creating an index, then take it up afterwards.

Two most notable differences between our approach is
1. you create MV, I create just index on column
2. you plan to get one node down when creating MV

@roydahan @fruch @fgelcer what do you think, should I implement it this way here?

fgelcer:
> > > @yarongilor , it may be relevant to something you wanted to do
> > 
> > 
> > @fgelcer , @soyacz , @roydahan , this PR may have quite a similar coverage to another new initiated PR of #5852
> > So perhaps we can consider adding some missing coverage in this PR or any other idea you think of. The main missing coverage in this PR is taking a node down prior to creating an index, then take it up afterwards.
> 
> Two most notable differences between our approach is
> 
> 1. you create MV, I create just index on column
> 2. you plan to get one node down when creating MV
> 
> @roydahan @fruch @fgelcer what do you think, should I implement it this way here?

could you please evaluate the extra effort in here? 

@yarongilor , i think you should add your own test for your task, as we should not only add and remove MV during toggling node up/down, but also add some verification on correctness of data in some extent.

soyacz:
> > Two most notable differences between our approach is
> > 
> > 1. you create MV, I create just index on column
> > 2. you plan to get one node down when creating MV
> > 
> > @roydahan @fruch @fgelcer what do you think, should I implement it this way here?
> 
> could you please evaluate the extra effort in here?

Adding stop/start one node is pretty easy, just matter of testing how it works.
As to replace index by MV also not that hard, just different syntax (a bit more complicated) I suppose and different way of MV creation verification.
I rather wonder **if** we want to implement this way.

yarongilor:
> evaluate the extra effort in here?

i actually meant having both MV + SI in this nemesis. any reason not having both?

fgelcer:
> > evaluate the extra effort in here?
> 
> i actually meant having both MV + SI in this nemesis. any reason not having both?

for this specific nemesis, we want to test SI, so @yarongilor , please create your own nemesis with MV only (including the verification needed)

soyacz:
@yarongilor @fgelcer I spoke with @fruch and we agreed to not add node operations during this nemesis. There are other cases - not only getting node down that would be worth to test. General direction is to create simple disruptions and in future we'll be adding parallel nemesis. 
E.g. one idea was to run one nemesis in a loop along with disruptive ones (eg. topology changes).

soyacz:
I tested it, looks working fine:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/34/

I decreased verbosity of index build status verification to not flood the logs.

I'm not sure if we want to backport it - it's not strictly a feature test in 5.2/2023.1 and we might want to avoid reshuffling SisyphusNemesis in these branches jobs.

data_dir/nemesis.yml
```diff
@@ -97,6 +104,7 @@
   - disruptive = False
   - kubernetes = True
   - limited = True
+  - schema_changes = True
```
fruch, line 107: @roydahan you test doesn't cover the tags/labels of nemesis, hence changes in tags doesn't force update to those files

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -0,0 +1,98 @@
...
+    raise TimeoutError(f"Timeout error while creating index {index_name}. "
+                       f"stdout\n: {result.stdout}\n"
+                       f"stderr\n: {result.stderr}")
+
+
+def verify_query_by_index_works(session, ks, cf, column) -> None:
+    # get some value from table to use it in query by index
+    result = session.execute(f'SELECT "{column}" FROM {ks}.{cf} limit 1')
+    value = list(result)[0][0]
+    match type(value).__name__:
```
fruch, line 75: Those are all the possible types ?

doesn't user profile or s-b or Gemini might create other types ?

Please consider verifying this nemesis with those as well

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -0,0 +1,98 @@
...
+    except Exception as exc:  # pylint: disable=broad-except
+        InfoEvent(message=f"Index {ks}.{cf}({column}) does not work in query: {query}. Reason: {exc}",
+                  severity=Severity.ERROR).publish()
+    if len(list(result)) == 0:
+        InfoEvent(message=f"Index {ks}.{cf}({column}) does not work. No rows returned for query {query}",
+                  severity=Severity.ERROR).publish()
+
+
+def drop_index(session, ks, index_name) -> None:
+    LOGGER.info('starting dropping index: {}.{}'.format(ks, index_name))
```
yarongilor, line 93: lazy formatting?

soyacz: fixed

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -0,0 +1,98 @@
...
+            return
+        time.sleep(30)
+    raise TimeoutError(f"Timeout error while creating index {index_name}. "
+                       f"stdout\n: {result.stdout}\n"
+                       f"stderr\n: {result.stderr}")
+
+
+def verify_query_by_index_works(session, ks, cf, column) -> None:
+    # get some value from table to use it in query by index
+    result = session.execute(f'SELECT "{column}" FROM {ks}.{cf} limit 1')
```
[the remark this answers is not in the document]
soyacz: here we just get value to put to `where` clause - later query all rows.

sdcm/utils/nemesis_utils/indexes.py
soyacz: I tried using more types, found `datetime` needed also quoting.
Generally, number of python types is much more limited than column types and I found only these 3 to require quoting or adding some reformat.

sdcm/utils/nemesis_utils/indexes.py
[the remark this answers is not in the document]
yarongilor: ok, i just have a concern it might reproduce the issue of https://github.com/scylladb/scylladb/issues/12268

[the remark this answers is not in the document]
soyacz: adding SimpleStatement with fetch_size shall help?

sdcm/utils/nemesis_utils/indexes.py
fruch: you write those 3, and there only two types here in the match/case...

fruch (dismissed):

fruch:
@yarongilor can you take a look, about the things you raised ? if they are addressed

data_dir/nemesis.yml
roydahan: I didn't think about that...

sdcm/utils/nemesis_utils/indexes.py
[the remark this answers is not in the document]
roydahan: Yes

roydahan (approved):

roydahan:
There is a conflict + one comment from @fruch about types and one adjustment to do for fetch_size.

roydahan:
It's a task came from a feature for 5.2, so it needs backport to 5.2.

sdcm/nemesis.py
```diff
@@ -3990,6 +3992,26 @@ def format_error_for_sla_test_and_raise(error_events):
...
+                index_name = create_index(session, ks, cf, column)
+            except InvalidRequest as exc:
+                LOGGER.warning(exc)
+                raise UnsupportedNemesis(  # pylint: disable=raise-missing-from
+                    "Tried to create already existing index. See log for details")
+            try:
+                wait_for_index_to_be_built(self.target_node, ks, index_name, timeout=1200)
+                verify_query_by_index_works(session, ks, cf, column)
+            finally:
+                drop_index(session, ks, index_name)
```
roydahan, line 4013: I think that before dropping the index, as I wrote in @yarongilor 's PR for adding MV, we should let the test run with the index for a while.
In the other PR I suggested 1-2% of the test duration, but didn't give much thought.
Maybe:
5m < 1% < 40m
(minimum 5m, 1% for the general case, maximum 40m for the very long cases).

soyacz: done

sdcm/utils/nemesis_utils/indexes.py
soyacz: seems I didn't push changes... thanks for checking

soyacz:
I've found some issues in `get_table_with_biggest_partitions_count` - was not excluding system_ tables and not counting properly partitions, as they were separated by token ranges (I added sum and group by to the query). Fixed. Still one more run to verify all works well: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-MajorCompactionMonkey-aws-test/5/

soyacz:
One more thing: shall I rerun `create-nemesis-pipelines` or we do it when branching?

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -0,0 +1,101 @@
...
+from sdcm.cluster import BaseNode
+from sdcm.sct_events import Severity
+from sdcm.sct_events.database import DatabaseLogEvent
+from sdcm.sct_events.filters import DbEventsFilter
+from sdcm.sct_events.system import InfoEvent
+
+LOGGER = logging.getLogger(__name__)
+
+
+def get_table_with_biggest_partitions_count(session) -> Tuple[str, str]:
```
yarongilor, line 30: It is not clear why using this method. Why limiting the tables to be selected for creating SI?
IIUC, a test can have, for example, a s-b table with large-partitions, but it will never be tested if there's a c-s table with a higher number of partitions. Why not give a chance to all user tables to be tested?

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -0,0 +1,101 @@
...
+    """Gets table with the most number of partitions"""
+    LOGGER.debug('Getting table with the most number of partitions')
+    res = session.execute('select keyspace_name, table_name,  sum(partitions_count) as partitions_count'
+                          ' from system.size_estimates GROUP BY keyspace_name, table_name')
+    rows_sorted = sorted(res, key=lambda x: x.partitions_count, reverse=True)
+    row = next(row for row in rows_sorted if not row.keyspace_name.startswith("system_"))
```
yarongilor, line 36: Are materialized-views filtered out here?

fruch:
> One more thing: shall I rerun `create-nemesis-pipelines` or we do it when branching?

Yes, master should be uptodate

data_dir/nemesis.yml
fruch: I would add it once adding the 2nd part of running from the nemesis file

sdcm/utils/nemesis_utils/indexes.py
soyacz: You seem to be right. I was thinking that there are small ks/cf in the cluster, but apparently there are not many. I switched it to `get_non_system_ks_cf_list` with MV exclude.

sdcm/utils/nemesis_utils/indexes.py
soyacz: now yes

soyacz:
> > One more thing: shall I rerun `create-nemesis-pipelines` or we do it when branching?
> 
> Yes, master should be uptodate

done

soyacz:
Recent update:
1. now we don't filter for biggest table, just using non-system ones. 
2. replaced few logger.info with `InfoEvent` so crucial actions are in events.
3. Created jobs files.

@fruch @yarongilor please re-review

sdcm/utils/common.py
```diff
@@ -2966,3 +2966,11 @@ def __exit__(self, exit_type, value, _traceback):
 def time_period_str_to_seconds(time_str: str) -> int:
     """Transforms duration string into seconds int. e.g. 1h -> 3600, 1h22m->4920 or 10m->600"""
     return sum([int(g[0] or 0) * 3600 + int(g[1] or 0) * 60 + int(g[2] or 0) for g in duration_pattern.findall(time_str)])
+
+
+def sleep_for_percent_of_duration(duration: int, percent: int, min_duration: int, max_duration: int):
+    """Waits the percentage of a duration in seconds, with a minimum and maximum duration"""
```
roydahan, line 2972: Worth to explain it like min < duration * percentage < max, it's hard to understand it like this.

test-cases/nemesis/longevity-5gb-1h-CreateIndexNemesis.yaml
```diff
@@ -0,0 +1,31 @@
+test_duration: 90
```
roydahan, line 1: Coll that you remembered to add it!
(We will need to remember this for every nemesis we add).

roydahan (dismissed):

test-cases/nemesis/longevity-5gb-1h-CreateIndexNemesis.yaml
soyacz: I didn't remember... this is automatically generated file :sweat_smile:

fruch: we should be checking it somehow, in a unittest or a new step for the CI

sdcm/utils/common.py
soyacz: done

test-cases/nemesis/longevity-5gb-1h-CreateIndexNemesis.yaml
soyacz: oh, you meant the file :) yes, somehow I remembered

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -0,0 +1,88 @@
...
+                  severity=Severity.ERROR).publish()
+
+
+def drop_index(session, ks, index_name) -> None:
+    InfoEvent(message=f"Starting dropping index: {ks}.{index_name}").publish()
+    with DbEventsFilter(
+            db_event=DatabaseLogEvent.DATABASE_ERROR,
+            line="Error applying view update"):
+        session.execute(f'DROP INDEX {ks}.{index_name}')
+        time.sleep(30)  # errors can happen only within several seconds after index drop #12977
```
fruch, line 88: nitpick:
```suggestion
        time.sleep(30)  # errors can happen only within several seconds after index drop scylladb/scylladb#12977
```

fruch (approved):
LGTM

roydahan (approved):

Issue #5843, opened 2023-02-22
password is printed to log


yarongilor:
The ldap bind pwd is printed in debug log:
```
t:2023-02-21 09:20:19,546 f:remote_file.py  l:63   c:sdcm.remote.remote_file p:DEBUG > ldap_bind_passwd: scylla-qa2021!
```

Originated from https://github.com/scylladb/scylla-enterprise/issues/2678#issuecomment-1438670172
We should avoid printing it.

## Installation details

Cluster size: 6 nodes (i3.4xlarge)

Scylla Nodes used in this run:
- longevity-100gb-4h-2023-1-db-node-3bc25371-6 (13.53.36.99 | 10.0.0.192) (shards: 14)
- longevity-100gb-4h-2023-1-db-node-3bc25371-5 (13.48.104.37 | 10.0.1.50) (shards: 14)
- longevity-100gb-4h-2023-1-db-node-3bc25371-4 (13.49.21.145 | 10.0.3.78) (shards: 14)
- longevity-100gb-4h-2023-1-db-node-3bc25371-3 (13.53.82.247 | 10.0.0.132) (shards: 14)
- longevity-100gb-4h-2023-1-db-node-3bc25371-2 (16.16.28.143 | 10.0.2.173) (shards: 14)
- longevity-100gb-4h-2023-1-db-node-3bc25371-1 (13.48.13.188 | 10.0.3.248) (shards: 14)

OS / Image: `ami-0f0eae5efea109684` (aws: eu-north-1)

Test: `longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-test`
Test id: `3bc25371-ed46-4a18-9aaf-151df3dfff7c`
Test name: `enterprise-2023.1/longevity/longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-test`
Test config file(s):

- [longevity-100gb-4h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/d318637cf75441501644e71531a5cd1feaea0278/test-cases/longevity/longevity-100gb-4h.yaml)

<details>
<summary>
Logs and commands
</summary>

- Restore Monitor Stack command: `$ hydra investigate show-monitor 3bc25371-ed46-4a18-9aaf-151df3dfff7c`
[...]

PR #5847, opened 2023-02-22
improvement(yaml content): Mask passwords when printed to log

yarongilor:
For security reasons, password values of yaml parameters,
	better be masked with asteriks when logged.
	Fixes: #5843

Fixes: [#5843](https://github.com/scylladb/scylla-cluster-tests/issues/5843)

### DIFF
--- a/sdcm/remote/remote_file.py
+++ b/sdcm/remote/remote_file.py
@@ -23,10 +23,27 @@ from sdcm import wait
 
 
 LOGGER = logging.getLogger(__name__)
+# The below set of SECURED_PARAMETERS should include relevant password-parameter-names,
+# to be masked by asterisks when printed to log.
+SECURED_PARAMETERS = {'ldap_bind_passwd'}
 
 
 def read_to_stringio(fobj):
     return StringIO(fobj.read())
+
+
+def secured_masked_content(parameters_content: str) -> str:
+    yaml_content = {}
+    out = yaml.safe_load(parameters_content)
+    if isinstance(out, dict):
+        yaml_content.update(out)
+    if not yaml_content:
+        LOGGER.warning("Got an invalid parameters-content. cannot mask any value.")
+        return ''
+    for key, value in yaml_content.items():
+        if key in SECURED_PARAMETERS:
+            yaml_content[key] = '*' * len(value)
+    return yaml.safe_dump(yaml_content)
 
 
 # pylint: disable=too-many-locals,too-many-arguments
@@ -60,7 +77,7 @@ def remote_file(remoter, remote_path, serializer=StringIO.getvalue, deserializer
         with open(local_tempfile, "w", encoding="utf-8") as fobj:
             fobj.write(content)
 
-        LOGGER.debug("New content of `%s':\n%s", remote_path, content)
+        LOGGER.debug("New content of `%s':\n%s", remote_path, secured_masked_content(parameters_content=content))
 
         remote_tempfile = remoter.run("mktemp").stdout.strip()
         remote_tempfile_move_cmd = f"mv '{remote_tempfile}' '{remote_path}'"


fgelcer:
why is it in draft?
verify why the checks failed

yarongilor:
Tested in: https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-repro/4/

yarongilor:
> why is it in draft? verify why the checks failed

Not sure why it fails with:
```
10:56:27    File "/tmp/jenkins/workspace/scylla-staging/yarongilor/longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-repro/scylla-cluster-tests/sdcm/remote/remote_file.py", line 40, in secured_masked_content
10:56:27      for key, value in yaml_content.items():
10:56:27  AttributeError: 'str' object has no attribute 'items'
```
It works ok when tested locally:
```
masked_content = secured_masked_content(parameters_content=parameters_content)
masked_content
"experimental: true\nldap_attr_role: cn\nldap_bind_dn: cn=admin,dc=scylla-qa,dc=com\nldap_bind_passwd: '**************'\nldap_url_template: ldap://3.91.246.6:389/dc\nlisten_address: 10.0.3.248\n"
```

@enaydanov , @fruch , do you have any idea? could it be a python env issue?

fgelcer:
> > why is it in draft? verify why the checks failed
> 
> Not sure why it fails with:
> 
> ```
> 10:56:27    File "/tmp/jenkins/workspace/scylla-staging/yarongilor/longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-repro/scylla-cluster-tests/sdcm/remote/remote_file.py", line 40, in secured_masked_content
> 10:56:27      for key, value in yaml_content.items():
> 10:56:27  AttributeError: 'str' object has no attribute 'items'
> ```
> 
> It works ok when tested locally:
> 
> ```
> masked_content = secured_masked_content(parameters_content=parameters_content)
> masked_content
> "experimental: true\nldap_attr_role: cn\nldap_bind_dn: cn=admin,dc=scylla-qa,dc=com\nldap_bind_passwd: '**************'\nldap_url_template: ldap://3.91.246.6:389/dc\nlisten_address: 10.0.3.248\n"
> ```
> 
> @enaydanov , @fruch , do you have any idea? could it be a python env issue?

it is failing in your new code... perhaps because you changed the `dict()` to be a string?

yarongilor:
> > > why is it in draft? verify why the checks failed
> > 
> > 
> > Not sure why it fails with:
> > ```
> > 10:56:27    File "/tmp/jenkins/workspace/scylla-staging/yarongilor/longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-repro/scylla-cluster-tests/sdcm/remote/remote_file.py", line 40, in secured_masked_content
> > 10:56:27      for key, value in yaml_content.items():
> > 10:56:27  AttributeError: 'str' object has no attribute 'items'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It works ok when tested locally:
> > ```
> > masked_content = secured_masked_content(parameters_content=parameters_content)
> > masked_content
> > "experimental: true\nldap_attr_role: cn\nldap_bind_dn: cn=admin,dc=scylla-qa,dc=com\nldap_bind_passwd: '**************'\nldap_url_template: ldap://3.91.246.6:389/dc\nlisten_address: 10.0.3.248\n"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @enaydanov , @fruch , do you have any idea? could it be a python env issue?
> 
> it is failing in your new code... perhaps because you changed the `dict()` to be a string?

i didn't change any dict to be a string. i defined a dict variable, then the CI code claims it's a string.

yarongilor:
> > > > > why is it in draft? verify why the checks failed
> > > > 
> > > > 
> > > > Not sure why it fails with:
> > > > ```
> > > > 10:56:27    File "/tmp/jenkins/workspace/scylla-staging/yarongilor/longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-repro/scylla-cluster-tests/sdcm/remote/remote_file.py", line 40, in secured_masked_content
> > > > 10:56:27      for key, value in yaml_content.items():
> > > > 10:56:27  AttributeError: 'str' object has no attribute 'items'
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > It works ok when tested locally:
> > > > ```
> > > > masked_content = secured_masked_content(parameters_content=parameters_content)
> > > > masked_content
> > > > "experimental: true\nldap_attr_role: cn\nldap_bind_dn: cn=admin,dc=scylla-qa,dc=com\nldap_bind_passwd: '**************'\nldap_url_template: ldap://3.91.246.6:389/dc\nlisten_address: 10.0.3.248\n"
> > > > ```
> > > > 
> > > > 
> > > >     
> > > >       
> > > >     
> > > > 
> > > >       
> > > >     
> > > > 
> > > >     
> > > >   
> > > > @enaydanov , @fruch , do you have any idea? could it be a python env issue?
> > > 
> > > 
> > > it is failing in your new code... perhaps because you changed the `dict()` to be a string?
> > 
> > 
> > i didn't change any dict to be a string. i defined a dict variable, then the CI code claims it's a string.
> 
> You defined variable which may be a dict, it also may be of lots of other types -> string, integer, boolean... Just example for you:
> 
> ```
> $ python
> >>> import yaml
> >>> yaml.safe_load('')
> >>> yaml.safe_load('foo')
> 'foo'
> >>> yaml.safe_load('foo: bar')
> {'foo': 'bar'}
> >>> yaml.safe_load('true')
> True
> >>> yaml.safe_load('1')
> 1
> ```
> 
> So, the statement `i defined a dict variable` is false. It would be true if it was guaranteed to be so, but it is not. The `dict` part in the `yaml_content: dict = ...` construction is only `type hint`, it doesn't change types anyhow and doesn't raise errors in case of mismatch.

thanks @vponomaryov , it looks like a funny python weakness - under forcing "types" then over-forcing its methods.. fixed now.

fgelcer (approved):
LGTM

sdcm/remote/remote_file.py
```diff
@@ -23,12 +23,29 @@
...
+# The below set of SECURED_PARAMETERS should include relevant password-parameter-names,
+# to be masked by asterisks when printed to log.
+SECURED_PARAMETERS = {'ldap_bind_passwd'}
 
 
 def read_to_stringio(fobj):
     return StringIO(fobj.read())
 
 
+def secured_masked_content(parameters_content: str) -> str:
```
fruch, line 35: What does parameters mean in the input of this function ?

sdcm/remote/remote_file.py
```diff
@@ -23,12 +23,29 @@
 
 
 LOGGER = logging.getLogger(__name__)
+# The below set of SECURED_PARAMETERS should include relevant password-parameter-names,
+# to be masked by asterisks when printed to log.
+SECURED_PARAMETERS = {'ldap_bind_passwd'}
```
fruch, line 28: This solution would work only on first level of keys, i.e. when there a password on a nested structure this would be enough (for example SSL configuration which is nested)

yarongilor: @fruch , that is not correct (and also not the goal of this specific fix). it will hide the whole value including nested content, which might not be optimal, but does achieve the goal of secured logging..
just an example:
```
yaml_content
{'prepare_write_cmd': '   - >-\n    bin/ycsb load dynamodb -P workloads/workloadc -threads 80 -p recordcount=6990500\n    -p fieldcount=10 -p fieldlength=512 -p dataintegrity=true\n    -p insertstart=0 -p insertcount=3495250\n\n  - >-\n    bin/ycsb load dynamodb -P workloads/workloadc -threads 80 -p recordcount=6990500\n    -p fieldcount=10 -p fieldlength=512 -p dataintegrity=true\n    -p insertstart=3495250 -p insertcount=3495250\n\n  - >-\n    bin/ycsb load dynamodb -P workloads/workloadc -threads 80 -p recordcount=6990500\n    -p fieldcount=10 -p fieldlength=512 -p dataintegrity=true\n    -p insertstart=0 -p insertcount=3495250 -p table=usertable_no_lwt\n\n  - >-\n    bin/ycsb load dynamodb -P workloads/workloadc -threads 80 -p recordcount=6990500\n    -p fieldcount=10 -p fieldlength=512 -p dataintegrity=true\n    -p insertstart=3495250 -p insertcount=3495250 -p table=usertable_no_lwt '}

for key, value in yaml_content.items():
    if key in ['prepare_write_cmd']:
        yaml_content[key] = '*' * len(value)
        
yaml_content
{'prepare_write_cmd': '********************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************************'}
```

This fix was requested for the specific ldap parameter. If someone faces a different need in the future, it would be fine to adjust it.

sdcm/remote/remote_file.py
yarongilor: it aims to make it clearer to understand it's not a plain string, but a string that could be yaml-loaded to a form of a dictionary.
It will only mask by parameter names received in such a format.

sdcm/remote/remote_file.py
fruch: if I would want to hide `truststore_password`, you are saying I would need to hide all of `client_encryption_options` ?
```yaml
client_encryption_options:
  enabled: true
  keystore: <Path to keystore>/keystore.jks
  keystore_password: <keystore password> ## The password used when generating the keystore.
  truststore: <Path to keystore>/keystore.jks
  truststore_password: <keystore password>
  require_client_auth: <true or false>
```

Anyhow just wondering how hiding this from log, while we still upload the actual yaml into a public s3 bucket, is really keeping this password safe .... ?

sdcm/remote/remote_file.py
fruch: it's not SCT params, it's aiming for scylla.yaml, I can understand why it would be call `content` or `yaml_string`, but what `parameters_content` tells me ?

also I think it should be 'mask_secured_content` to be a bit more clear what it's doing

sdcm/remote/remote_file.py
yarongilor: @fruch , i think one difference is that a password printed to log can be much easier searched through a github search or a similar, since it will be found on scylla issues details. But perhaps this S3 bucket is another security issue. @mykaul , please comment any advise per this discussion ( originated from https://github.com/scylladb/scylla-enterprise/issues/2678#issuecomment-1438670172 )

sdcm/remote/remote_file.py
yarongilor: @fruch , does `yaml_content` sound better?

fruch: yes

fruch:
@yarongilor is this one still relevant ?

yarongilor:
> @yarongilor is this one still relevant ?

@fruch , i think this can be merged, yes. It addressed @mykaul request (https://github.com/scylladb/scylla-enterprise/issues/2678#issuecomment-1438670172) and it can do no harm.

fruch:
@soyacz can you give a look as well ?

Or open other PR the cover it more generically?

sdcm/remote/remote_file.py
```diff
@@ -23,12 +23,29 @@
...
 
 
 def read_to_stringio(fobj):
     return StringIO(fobj.read())
 
 
+def secured_masked_content(parameters_content: str) -> str:
+    yaml_content = {}
+    out = yaml.safe_load(parameters_content)
```
soyacz, line 37: it's placed under `remote_file` - which is not always yaml and content may cause it to raise exception and do actual harm.

soyacz (changes requested):
this solution is very limited.
I think we could find more generic one with use of regexp and parse line by line in search for passwords keywords like `password|passwd|secret` etc. and replace the word after.

This requires unit testing - so when we want to fix issues we don't introduce regressions.

sdcm/remote/remote_file.py
yarongilor: `yaml` is yet another valid type of a remote file. Please let me know if there's any better suggestion.
any line of code in python may raise an exception.. that's how python is built.

yarongilor:
> @yarongilor is this one still relevant ?

@fruch , this question is originated from a request by @mykaul.
I cannot tell better than you.
I didn't find any run of this test in last year as well.
Please advise weather it's relevant and a more general solution is needed.

fruch:
> > @yarongilor is this one still relevant ?
> 
> @fruch , this question is originated from a request by @mykaul. I cannot tell better than you. I didn't find any run of this test in last year as well. Please advise weather it's relevant and a more general solution is needed.

since scylla can have password in multiple levels of the configuration, we need better solution 
as mentioned long ago, this idea is limited, we are not gonna accept it as is.

if we really don't run that test that often, it's not such a severe problem, and we can wait for better implementation

yarongilor:
> > > @yarongilor is this one still relevant ?
> > 
> > 
> > @fruch , this question is originated from a request by @mykaul. I cannot tell better than you. I didn't find any run of this test in last year as well. Please advise weather it's relevant and a more general solution is needed.
> 
> since scylla can have password in multiple levels of the configuration, we need better solution as mentioned long ago, this idea is limited, we are not gonna accept it as is.
> 
> if we really don't run that test that often, it's not such a severe problem, and we can wait for better implementation

closing then.

fruch:
> `cloud_prom_bearer_token` is the closest thing to credentials SCT holds, and that not in scylla configuration, so that PR won't be covering it anyhow

I think the original issue was about ldap server credentials, a testing server. (and now that password is put in OSS issue, so it can't be any worse than that :\)

the fact it's needs to be pass down to scylla clear in configuration, I would consider a bug in scylla-enterprise...

PR #1473, opened 2019-11-25
Support to workaround kernel bug in scylla setup

amoskong:
Support to workaround kernel bug in scylla setup.

We don't want the kernel bug block our upgrade test, we always need to enable
it if gce backend is used. We can remove or disable this workaround when the
kernel bug is fixed in default kernel of GCE instance.
    
Bad kernel in gce instance:
- 3.10.0-1062.1.2.el7.x86_64
- 3.10.0-1062.4.3.el7.x86_64 (latest, still not fix)

## PR pre-checks (self review)

- [x] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [x] I gave variables/functions meaningful self-explanatory names
- [x] I didn't leave commented-out/debugging code
- [x] I didn't copy-paste code
- [x] I added the relevant `backport` labels
- [x] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5851, opened 2023-02-23
fix(offline-installer): remove deprecated seastar parameters

yaronkaikov:
all non-root offline installer tests fail with the following error message:
```
error: unrecognised option '--num-io-queues=8'
```

Following the latest update of seastar
https://github.com/scylladb/scylladb/commit/4f5a460db91de59018815106abb54c1246bbb014. one of the changes was https://github.com/scylladb/seastar/commit/6ad997708f20636a2e8394c83165cf0009d7b77e

Since `num-io-queues` is no longer available it should be removed

### DIFF
--- a/configurations/io.conf
+++ b/configurations/io.conf
@@ -1,1 +1,1 @@
-SEASTAR_IO="--num-io-queues=8 --io-properties-file=/etc/scylla.d/io_properties.yaml"
+SEASTAR_IO="--io-properties-file=/etc/scylla.d/io_properties.yaml"
--- a/query_limits_test.py
+++ b/query_limits_test.py
@@ -66,7 +66,7 @@ class QueryLimitsTest(ClusterTester):
         self.db_cluster.run("sudo cp /tmp/l /etc/sysconfig/scylla-server")
         self.db_cluster.run("sudo chown root.root /etc/sysconfig/scylla-server")
         # Configure seastar IO to use the smallest acceptable values (slow disk)
-        self.db_cluster.run('echo \'SEASTAR_IO="--num-io-queues 1 --max-io-requests 4"\' > /tmp/m')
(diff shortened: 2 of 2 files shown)


yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/artifacts-offline-install/job/artifacts-ubuntu1804-nonroot-test/13/

fgelcer (dismissed):
LGTM

soyacz (dismissed):

fgelcer:
@roydahan ping

fgelcer (dismissed):

roydahan:
This is the commit message for io.conf we have right now:

```
gce: add a parameter to workaround kernel bug in scylla setup  There is a known kernel bug will cause scylla_io_setup to fail in executing iotune.  The kernel bug doesn't occur all the time, so we can get some succeed gce instance, the config files are copied from a succeed gce instance (same instance type, same test). 
```

I need to better understand where we use it and if the kernel bug still there (probably not).
In that case, we should remove completly the io.conf and run scylla_io_setup.

roydahan:
I suggest we remove this workaround and go back testing with scylla_io_setup as it should.
If we still bump into the issue, it'll become a blocker for iotune and no workarounds.

roydahan:
This is the PR that introduced the workaround: https://github.com/scylladb/scylla-cluster-tests/pull/1473

@soyacz I assigned it to you since @fruch is out and your team is responsible on artifacts this week.
Could you please take care of it?

soyacz:
> This is the PR that introduced the workaround: #1473
> 
> @soyacz I assigned it to you since @fruch is out and your team is responsible on artifacts this week. Could you please take care of it?

Ok, I'll create PR with workaround removal.

yaronkaikov:
Since @soyacz will take care of it, closing this PR

fgelcer:
closing this in favor of #5857

PR #5852, opened 2023-02-23
improvement(MV): A new nemesis to add/remove MV dynamically

yarongilor:
improvement(Materialized view): A nemesis to add/remove MV dynamically
    
            Test adding/removing MV + node down and data verification.

Task: https://github.com/scylladb/qa-tasks/issues/1012

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2862,6 +2862,41 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     def disrupt_show_toppartitions(self):  # pylint: disable=function-redefined
         return self._disrupt_show_toppartitions(allow_new_api=False)
 
+    def disrupt_add_remove_mv(self):
+        node1, node2 = self.cluster.nodes[:2]
+        self.log.info("Stopping Scylla on node1")
+        node1.stop_scylla()
+        ks_name = 'new_mv_keyspace'
+        base_table_name = 'standard1'
+        view_name = f'{base_table_name}_view'
+        rows_number = 100000
+        try:
+            InfoEvent(message='Write c-s data to a new table').publish()
+            write_cmd = f"cassandra-stress write no-warmup cl=QUORUM n={rows_number} -schema 'keyspace={ks_name} replication(factor=3)' " \
+                        f"-mode cql3 native -rate threads=2 -pop seq=1..{rows_number} -log interval=5"
+            write_thread = self.tester.run_stress_thread(stress_cmd=write_cmd, round_robin=True)
+            self.tester.verify_stress_thread(cs_thread_pool=write_thread)
+            InfoEvent(message='Create a materialized-view for the new table').publish()
+            with self.cluster.cql_connection_patient(node=node2) as session:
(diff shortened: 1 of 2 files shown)


fgelcer:
@yarongilor --> https://github.com/scylladb/scylla-cluster-tests/pull/5841#issuecomment-1445951385

fgelcer:
so, i suggested to do all sort of creating MV + stop and start nodes, and the idea is to use as base table, something that you will create for the nemesis itself, and IMHO some 100K lines would suffice, so the verification in the end could be to read using `CL=QUORUM`, comparing the base table with the MV (that will be sync'ed after the node[s] that was[were] down will be restarted, repaired and reshaped.

yarongilor:
@fgelcer , please approve running a test with AWS instances

yarongilor:
Tested in: https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/longevity-100gb-4h-test-restart-then-repair/10/

yarongilor:
Opened issue: https://github.com/scylladb/scylladb/issues/13052

soyacz (commented):
This nemesis is very limited - creates MV on a small table (100k rows), always the same.  Also scenario is very specific - this looks rather like functional test than nemesis. 

Remember to add documentation, so we can get brief overview what nemesis does without reading implementation details.

roydahan (changes requested):
Exactly as @soyacz wrote.
If the nemesis doesn’t use the test data (in this case one of the base tables) or in other cases, the test configuration,it’s not a nemesis.

In this case, the test should look for a base table from non_system_ks_cf and create the view for it.
Then, after view building completed it should let it run for a while (e.g 2% of test_duration) and drop the view.
If we want to test view drop while building, we may want to build 2 views, one to wait for building and another to drop while building.

yarongilor:
> Exactly as @soyacz wrote. If the nemesis doesn’t use the test data (in this case one of the base tables) or in other cases, the test configuration,it’s not a nemesis.
> 
> In this case, the test should look for a base table from non_system_ks_cf and create the view for it. Then, after view building completed it should let it run for a while (e.g 2% of test_duration) and drop the view. If we want to test view drop while building, we may want to build 2 views, one to wait for building and another to drop while building.

@roydahan , @soyacz , @fgelcer ,

- the original plan suggested in the task was indeed using an existing table with data for creating MV.
- @fgelcer later suggested this different approach of using a newly created table and MV.
- So i moved from original implementation to this approach.
- We can possibly combine both approaches - one new limited table with validation. one existing table (without data validation).
- Or we can remove current implementation completely and implement only the original one.
- It's not clear why continue running for %2 of test duration that can take 2 - 3 hours. why not running for 5 - 10 minutes?
- I can add the part of removing MV as a second step of the nemesis.

roydahan:
I don't see a point to do a view building for newly created table because we already have this in our user profile.

The test should behave quite similarly to the new nemesis of building index that @soyacz introduced and use the same utils he introduced (and possibly add few more).

Regarding the 2% it was a suggestion, the idea is to have something that is relative to the test duration and not always 5 minutes. we can have 1% as well, we should start doing it for more nemesis we have or adding...

PR #5862, opened 2023-02-27
fix(disk-and-cache): use ssh transport for logs

aleksbykov:
Use ssh transport for logs for disk-and-cache
performance tests

### DIFF
--- a/test-cases/performance/perf-regression-latency-125gb.yaml
+++ b/test-cases/performance/perf-regression-latency-125gb.yaml
@@ -30,3 +30,4 @@ use_mgmt: true
 store_perf_results: true
 send_email: true
 email_recipients: ['scylla-perf-results@scylladb.com']
+logs_transport: "ssh"
--- a/test-cases/performance/perf-regression-throughput-125gb.yaml
+++ b/test-cases/performance/perf-regression-throughput-125gb.yaml
@@ -34,3 +34,4 @@ send_email: true
 email_recipients: ['scylla-perf-results@scylladb.com']
 
 email_subject_postfix: 'disk and cache'
+logs_transport: "ssh"


aleksbykov:
@roydahan could you merge this PR to branch-perf-v12 only

fgelcer:
@aleksbykov , it would be nice to understand why you need to change it... a bit more context helps to understand the change (even if a simple one)

fgelcer (approved):
other than the comment above, LGTM

aleksbykov:
> it would be nice to understand why you need to change it... a bit more context helps to understand the change (even if a simple one)

@fabio, i checked the job status and the latest succesful run was from 6.11.2022. after that we switch ubuntu-22.04. 
Job failed because runner can't get the logs from nodes. i think, we don' backport syslog-ng to perf-v12, and runner continue use the rsyslog which is removed by default from ubuntu-22.04. Other jobs regular throughput, latency, latency with nemesis was switched to used log_transport ssh. 
i switch old job to use my pr, and now it is running

vponomaryov:
> > it would be nice to understand why you need to change it... a bit more context helps to understand the change (even if a simple one)
> 
> @fabio, i checked the job status and the latest succesful run was from 6.11.2022. after that we switch ubuntu-22.04. Job failed because runner can't get the logs from nodes. i think, we don' backport syslog-ng to perf-v12, and runner continue use the rsyslog which is removed by default from ubuntu-22.04. Other jobs regular throughput, latency, latency with nemesis was switched to used log_transport ssh. i switch old job to use my pr, and now it is running

Please, write it in the commit message so it could appear in the git commit history and spill the light for the root cause.

roydahan (approved):

roydahan:
It's a merge to an old branch that is going to die soon.

Issue #5812, opened 2023-02-15
`get_nodes_status` generates too many requests to cloud api


soyacz:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

ClusterHealthValidator started publishing error events like:
```
2023-02-14 21:25:11.262: (ClusterHealthValidatorEvent Severity.ERROR) period_type=one-time event_id=948c165d-1824-41fa-acb5-5dda55483624: type=NodeStatus node=longevity-5gb-1h-SoftRebootNodeMonk-db-node-249f30ed-3 error=Current node Node longevity-5gb-1h-SoftRebootNodeMonk-db-node-249f30ed-3 [3.252.166.103 | 10.4.3.176] (seed: True). The node Node longevity-5gb-1h-SoftRebootNodeMonk-db-node-249f30ed-1 [34.240.130.155 | 10.4.1.211] (seed: True) exists in the gossip but doesn't exist in the nodetool.status
```

The reason for it is that `get_nodes_status` we try to get `Node` object by finding it by ip 
`node := self.parent_cluster.find_node_by_ip(node_ip)`
And finding node by ip triggers `_refresh_instance_state` which does `ec2_instance_wait_public_ip` - making queries to AWS
When we run many jobs in parallel we deplete quota on these requests and cause these errors.

I think we could try first get node by ip without refreshing

## Impact

error events in logs. Possibly can lead to fail ClusterHealthValidation

## How frequently does it reproduce?

Often when running all nemesis tests at once. 

## Installation details

Kernel Version: 5.15.0-1028-aws
Scylla version (or git commit hash): `5.2.0~rc1-20230207.8ff4717fd010` with build-id `78fbb2c25e9244a62f57988313388a0260084528`

Cluster size: 3 nodes (i4i.large)

Scylla Nodes used in this run:
- longevity-5gb-1h-SoftRebootNodeMonk-db-node-249f30ed-3 (3.252.166.103 | 10.4.3.176) (shards: 2)
- longevity-5gb-1h-SoftRebootNodeMonk-db-node-249f30ed-2 (34.245.61.44 | 10.4.1.56) (shards: 2)
- longevity-5gb-1h-SoftRebootNodeMonk-db-node-249f30ed-1 (34.240.130.155 | 10.4.1.211) (shards: 2)

OS / Image: `ami-05e1d6aa4f71f3f25` (aws: eu-west-1)

[...]

PR #5869, opened 2023-02-28
fix(nodetool_status): reduce number of api calls

soyacz:
`get_nodes_status` generates (n, n*n) requests to cloud api which leads
to errors in `find_node_by_ip`.

Reduce number of cloud api calls by creating node_ip->node map and
reuse it for all nodes in `get_nodes_status` method.

refs: #5812

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2643,10 +2643,10 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         nodes_status = {}
         try:
             statuses = self.parent_cluster.get_nodetool_status(verification_node=self)
-
+            node_ip_map = self.parent_cluster.get_ip_to_node_map()
             for dc, dc_status in statuses.items():
                 for node_ip, node_properties in dc_status.items():
-                    if node := self.parent_cluster.find_node_by_ip(node_ip):
+                    if node := node_ip_map.get(node_ip):
                         nodes_status[node] = {'status': node_properties['state'], 'dc': dc}
                     else:
                         if node_ip:
@@ -2671,6 +2671,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             result = session.execute(f"select {', '.join(columns)} from system.peers")
             cql_results = result.all()
         err = ''
+        node_ip_map = self.parent_cluster.get_ip_to_node_map()
         for row in cql_results:
             peer = row.peer
             try:
@@ -2681,7 +2682,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
                 err += current_err
                 continue
 
-            if node := self.parent_cluster.find_node_by_ip(peer):
+            if node := node_ip_map.get(peer):
                 peers_details[node] = row._asdict()
             else:
                 current_err = f"'get_peers_info' failed to find a node by IP: {peer}\n"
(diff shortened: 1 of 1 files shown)


sdcm/cluster_gce.py
```diff
@@ -104,6 +104,7 @@ def _instance_wait_safe(self, instance_method: Callable[P, R], *args: P.args, **
                 f"Timeout while running '{instance_method.__name__}' method on GCE instance '{self._instance.id}'"
             ) from None
 
+    @cache
     def _refresh_instance_state(self):
```
fruch, line 108: IP addresses can't change in GCE ?

soyacz: As I read, only when instance is stopped which I believe we don't do and if yes, shouldn't we fix seed's in that case?

fruch: We are doing stop only in EC2, but only since we weren't using local disks in GCP, and now we do use local disks, and should be able to do the same sequence for GCP.

If we aren't not doing it today, we shouldn't assume we are not gonna do it.

Anyhow what do you mean by seeds in this context ?

soyacz: I mean - if IP address changed, wouldn't we need to replace them in scylla.yaml?
So if we don't do it and everything works, can we assume these IP's are not changing?

fruch: the ip address that can change is the public address, local address doesn't change in AWS (and I'm guessing in GCP as well)

anyhow you need to make sure this cache doesn't break the sequence of `refresh_ip_address` that has it's own cache removal process, you need to drop the catch on this one as well on any call to `refresh_ip_address`

hence doing this a back-end specific cache, kind of complicates the situation, i.e. that one backend has cache for this function and one doesn't

soyacz: yes, there's some issue with `refresh_ip_address`. Situation around this is very unclear - to obtain ip's we often use `_refresh_instance_state` and it's not obvious how and when to replace it with something else. There's no other unified way to get IP's. 
I'll rethink the solution.

soyacz: Messing with getting ip's is quite dangerous and a way too complicated here. I changed to safer solution. See updated commit/PR message.

sdcm/cluster.py
```diff
@@ -2643,10 +2643,10 @@ def get_nodes_status(self):
         nodes_status = {}
         try:
             statuses = self.parent_cluster.get_nodetool_status(verification_node=self)
-
+            node_ip_map = {ip: node for node in self.parent_cluster.nodes for ip in node.get_all_ip_addresses()}
             for dc, dc_status in statuses.items():
                 for node_ip, node_properties in dc_status.items():
-                    if node := self.parent_cluster.find_node_by_ip(node_ip):
+                    if node := node_ip_map.get(node_ip):
```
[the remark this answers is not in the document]
soyacz: > There are also `get_peers_info` and `get_gossip_info`, which, moreover, are wrapped with retries decorator.

these methods fail in case ip is not found and retry - if we cache it will be failing until cache is deleted (or ttl'ed) making it unusable. 

I'll reuse current solution there too.

fruch (approved):

vponomaryov (approved):
LGTM

soyacz:
provision tests failed due various issues (not related). I run test here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-MajorCompactionMonkey-aws-test/3/

soyacz:
> provision tests failed due various issues (not related). I run test here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-MajorCompactionMonkey-aws-test/3/

passed

fruch (approved):
LGTM

PR #5871, opened 2023-03-01
fix(cluster.py): ensure iptables is installed on DB node

enaydanov:
For the Scylla Machine images built after https://github.com/scylladb/scylla-machine-image/commit/3ab18c9e6b09580611ab73626d4fb37c3f503b3b we need to install iptables for networking nemeses.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3098,9 +3098,13 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         Generates random firewall rule to drop/reject packets for inter-node communications, port 7000 and 7001
         """
         name = 'RejectInterNodeNetwork'
+
+        self._install_iptables()
+
         textual_matching_rule, matching_rule = self._iptables_randomly_get_random_matching_rule()
         textual_pkt_action, pkt_action = self._iptables_randomly_get_disrupting_target()
         wait_time = random.choice([10, 60, 120, 300, 500])
+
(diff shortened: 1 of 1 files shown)


sdcm/cluster.py
```diff
@@ -4329,6 +4329,9 @@ def node_setup(self, node: BaseNode, verbose: bool = False, timeout: int = 3600)
 
         node.update_repo_cache()
 
+        if node.distro.is_ubuntu:
```
fruch, line 4332: Maybe we should install it as part of the nemesis that needs it ?

enaydanov: OK, let's do it this way.

fruch (approved):

sdcm/nemesis.py
```diff
@@ -3161,6 +3175,10 @@ def disrupt_network_reject_thrift(self):
             wait_time=wait_time
         )
 
+    def _install_iptables(self) -> None:
+        if self.target_node.distro.is_ubuntu:  # iptables is missing in a minimized Ubuntu installation
+            self.target_node.remoter.sudo("apt install iptables")
```
roydahan, line 3180: What happens when it’s already installed?
Do we have “fake errors” in the log?

fruch: it would say something like:
```
iptables is already the newest version (1.8.7-1ubuntu5).
```

and returns without failure. (status code 0)

roydahan: Ok, that’s good enough

roydahan (approved):

Issue #5767, opened 2023-02-05
c-s read stress doesn't finish according to requested duration of 240 minutes, failing its docker with timeout (15000 seconds)


yarongilor:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

scenario run c-s read stress for 4 hours on gp3 configuration. then c-s docker is killed after 4 hours + 10 minutes for timeout.
it is reproduced twice with gp3 and once with 5.2 100gb-4h longevity (Id: 8e7ebd5a-e754-40a7-93cd-bc4372e919c9)
error log:
```
2023-02-05 10:32:55.008: (CassandraStressEvent Severity.CRITICAL) period_type=end event_id=4a1d4308-86bd-4046-bf96-a02059262923 during_nemesis=DestroyDataThenRepair duration=4h10m4s: node=Node longevity-100gb-4h-master-loader-node-446c7138-1 [44.211.229.117 | 10.12.3.18] (seed: False)
stress_cmd=cassandra-stress read  cl=QUORUM duration=240m -schema 'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=20 -pop seq=1..20971520 -col 'n=FIXED(10) size=FIXED(512)' -log interval=5
errors:
Stress command execution failed with: Command did not complete within 15000 seconds!
Command: 'sudo  docker exec 40cb9f96cf2fb718fac65edb3934a9fdf7e05367a27ddacd282da859b3ad7564 /bin/sh -c \'echo TAG: loader_idx:0-cpu_idx:0-keyspace_idx:1; STRESS_TEST_MARKER=JVF9876RSNMZ4TIJMHPC; cassandra-stress read  cl=QUORUM duration=240m -schema keyspace=keyspace1 \'"\'"\'replication(factor=3) compaction(strategy=SizeTieredCompactionStrategy)\'"\'"\' -mode cql3 native -rate threads=20 -pop seq=1..20971520 -col \'"\'"\'n=FIXED(10) size=FIXED(512)\'"\'"\' -log interval=5 -transport \'"\'"\'truststore=/etc/scylla/ssl_conf/client/cacerts.jks truststore-password=cassandra\'"\'"\' -node 10.12.3.62,10.12.1.171,10.12.2.198,10.12.3.42,10.12.1.39,10.12.2.17 -errors skip-unsupported-columns\''
Stdout:
[...]

PR #5872, opened 2023-03-02
fix(stress): fix c-s timeouts

soyacz:
Recently c-s changed behavior and for duration workflows: it added warmup iterations. For heavy load, warmup doesn't fit to 10 minutes duration margin we assumed for c-s startup and causes SCT to timeout c-s thread.

This fix removes warmup by appending ` no-warmup` to all c-s cmd's. 
refs: #5767

### DIFF
--- a/sdcm/stress_thread.py
+++ b/sdcm/stress_thread.py
@@ -102,6 +102,10 @@ class CassandraStressThread(DockerBasedStressThread):  # pylint: disable=too-man
     def create_stress_cmd(self, cmd_runner, keyspace_idx):
         stress_cmd = self.stress_cmd
 
+        if "no-warmup" not in stress_cmd:
+            # add no-warmup to stress_cmd if it's not there. See issue #5767
+            stress_cmd = re.sub(r'(cassandra-stress [\w]+)', r'\1 no-warmup', stress_cmd)
+
         # When using cassandra-stress with "user profile" the profile yaml should be provided
         if 'profile' in stress_cmd and not self.profile:
             # support of using -profile in sct test-case yaml, assumes they exists data_dir


sdcm/tester.py
```diff
@@ -1807,6 +1807,8 @@ def run_stress_cassandra_thread(
         else:
             timeout = get_timeout_from_stress_cmd(stress_cmd) or self.get_duration(duration)
 
+        if "no-warmup" not in stress_cmd:
+            stress_cmd += " no-warmup"
```
fruch, line 1811: It's plain like that ? It shouldn't be part of a specific group of parameters ?

soyacz: your'e right it's not, need to put it other place

soyacz: fixed

sdcm/tester.py
```diff
@@ -1807,6 +1807,11 @@ def run_stress_cassandra_thread(
         else:
             timeout = get_timeout_from_stress_cmd(stress_cmd) or self.get_duration(duration)
 
+        if "no-warmup" not in stress_cmd and "stress user" not in stress_cmd:
```
fruch, line 1810: I think it's better to use `user profile=`
```suggestion
        if "no-warmup" not in stress_cmd and "user profile=" not in stress_cmd:
```

fruch: Also we are sure user-profile doesn't do warm-up ?

roydahan (dismissed):

sdcm/tester.py
soyacz: damn, I tested it but didn't notice it was already set. It supports no-warmup.

soyacz: fixed, and I found one more use case where before c-s was some java settings provided - also fixed.

fruch (dismissed):
LGTM

vponomaryov (approved):
LGTM

sdcm/tester.py
```diff
@@ -1807,6 +1807,13 @@ def run_stress_cassandra_thread(
         else:
             timeout = get_timeout_from_stress_cmd(stress_cmd) or self.get_duration(duration)
 
+        if "no-warmup" not in stress_cmd:
+            # add no-warmup to stress_cmd if it's not there see issue #5872
```
fgelcer, line 1811: did you mean issue #5767 ?

fgelcer (dismissed):
LGTM

sdcm/tester.py
soyacz: yes, fixed

soyacz:
I switched the place - I noticed I did it in `run_stress_cassandra_thread` which is not best place as still some places bypass it. I moved it to `CassandraStressThread` and used `re.sub` for brevity.
Please re-review.

fgelcer (approved):
LGTM

vponomaryov (approved):
LGTM

Issue #5874, opened 2023-03-05
TRUNCATE ... USING TIMEOUT is failing on operator 


fruch:
## Issue description

```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4108, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1850, in disrupt_truncate_large_partition
    self.target_node.run_cqlsh(cmd=f'TRUNCATE {ks_name}.{table} USING TIMEOUT {truncate_timeout}',
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 2779, in run_cqlsh
    cqlsh_out = self.remoter.run(cmd, timeout=timeout + 30,  # we give 30 seconds to cqlsh timeout mechanism to work
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/kubernetes_cmd_runner.py", line 164, in run
    return super().run(cmd, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 613, in run
    result = _run()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 67, in inner
    return func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 604, in _run
    return self._run_execute(cmd, timeout, ignore_status, verbose, new_session, watchers)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/kubernetes_cmd_runner.py", line 203, in _run_execute
    return super()._run_execute(cmd, timeout=timeout, ignore_status=ignore_status, verbose=verbose,
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 530, in _run_execute
    result = connection.run(**command_kwargs)
  File "/usr/local/lib/python3.10/site-packages/invoke/runners.py", line 380, in run
    return self._run_body(command, **kwargs)
  File "/usr/local/lib/python3.10/site-packages/invoke/runners.py", line 442, in _run_body
    return self.make_promise() if self._asynchronous else self._finish()
  File "/usr/local/lib/python3.10/site-packages/invoke/runners.py", line 509, in _finish
    raise UnexpectedExit(result)
[...]

PR #5854, opened 2023-02-23
improvement(nemesis.py): truncate nemesis using timeout

fgelcer:
adding to the truncate command, in the truncate
nemesis the new functionality of adding
`using timeout`, so we can have it running longer, and avoid failing for known issue where truncate
times out.

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #5875, opened 2023-03-05
Revert "improvement(nemesis.py): truncate nemesis using timeout"

fruch:
Reverts scylladb/scylla-cluster-tests#5854

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/5874


fgelcer (approved):
LGTM

PR #5881, opened 2023-03-06
fix(cluster.py): install s-b commands went broken

fgelcer:
not sure what changed, but on job longevity-10gb-3h-gce-test we are seeing failures since last October,
related to the function `install_scylla_bench` failing because `rm -f` on a directory doesn't work, and then `mkdir` fails with file already exists, making the whole installation process to fail.
adding the `-r` flag to the `rm` command will fix it.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -585,8 +585,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             sb_version = f'heads/{sb_version}'
         self.remoter.sudo(shell_script_cmd(f"""\
             rm -rf /usr/local/go
-            rm -f /tmp/sb_install || true
-            mkdir /tmp/sb_install
+            rm -rf /tmp/sb_install || true
+            mkdir -p /tmp/sb_install
             cd /tmp/sb_install
             curl -Lo go.tar.gz https://storage.googleapis.com/golang/go1.17.4.linux-amd64.tar.gz
             tar -C /usr/local -xvzf go.tar.gz


sdcm/cluster.py
```diff
@@ -585,8 +585,8 @@ def install_scylla_bench(self):
             sb_version = f'heads/{sb_version}'
         self.remoter.sudo(shell_script_cmd(f"""\
             rm -rf /usr/local/go
-            rm -f /tmp/sb_install || true
-            mkdir /tmp/sb_install
+            rm -rf /tmp/sb_install || true
+            mkdir -p /tmp/sb_install
```
enaydanov, line 589: It's redundant to use `-p` switch because the directory should be deleted at this point. But it doesn't matter, actually.

enaydanov (approved):

vponomaryov (approved):
LGTM

sdcm/cluster.py
```diff
@@ -585,8 +585,8 @@ def install_scylla_bench(self):
...
             cd /tmp/sb_install
             curl -Lo go.tar.gz https://storage.googleapis.com/golang/go1.17.4.linux-amd64.tar.gz
```
vponomaryov, line 591: Direct installation of the `s-b` is too old approach. We use docker images.
Latest `s-b` supports `go1.18+`. `1.17` which is used here is not supported by the latest `scylla-bench`.
If not latest one is used then need to check the compatibility.

fgelcer: @vponomaryov , the target branch is 2021.1, so old branch, old code, and it doesn't support using the docker, as we have on newer branches

sdcm/cluster.py
fgelcer: yes, but what failed the test was the `mkdir`, so i prefer to "ignore" its errors, although now the `rm` will always succeed

sdcm/cluster.py
vponomaryov: I am saying that the `go` version used here may be too old. The go version is specified on the line 591 here.

fgelcer: i'm giving it a ride [here](https://jenkins.scylladb.com/job/enterprise-2021.1/job/longevity/job/longevity-10gb-3h-gce-test/88/) and soon we will know if it works

PR #5888, opened 2023-03-08
fix(scylla_yaml): handle undefined keys

fruch:
before the fix, if a specific key defined in `scylla.yaml` but not defined in ScyllaYaml object, it would be remove from the final configuration file.

after this change any define key in `scylla.yaml` would be pass along regardles if it's defined in ScyllaYaml or not.

example of such problemtic key is `force_schema_commit_log` that was defined true only on scylla.yaml, but never defined on SCT code (also it was set to default true on scylla code in config.cc)


soyacz (approved):

sdcm/provision/scylla_yaml/scylla_yaml.py
```diff
@@ -342,8 +346,8 @@ def _update_dict(self, obj: dict, fields_data: dict):
         for attr_name, attr_value in obj.items():
             attr_info = fields_data.get(attr_name, None)
             if attr_info is None:
-                raise ValueError("Provided unknown attribute `%s`" % (attr_name,))
-            if hasattr(attr_info.type_, "__attrs_attrs__"):
+                logger.warning("Provided unknown attribute `%s`", attr_name)
```
soyacz, line 349: This still is not showing warning when providing undefined config options in class constructor - this is the reason we missed the issue - should have been failed.

fruch: this code flow is only on update, not creation.

once we get to the update flow, the unknown fields are long gone by now

soyacz: yes, that's why we missed the issue - if it was in creation flow it would fail in artifact test I suppose.
The same, now we won't be seeing warnings when existing scylla.yaml has something we don't specify.

fruch: we will see those warning once updates are taking place

roydahan (approved):

PR #5890, opened 2023-03-08
fix(longevity): keyspace for data validator

juliayakovlev:
Test https://jenkins.scylladb.com/job/scylla-master/job/raft/job/longevity-lwt-parallel-schema- changes-with-disruptive-24h-raft-test/8/ failed with error:
```
File /home/ubuntu/scylla-cluster-tests/sdcm/cluster.py, line 3380, in cql_connection return self._create_session(node=node, keyspace=keyspace, user=user, password=password, 
File /home/ubuntu/scylla-cluster-tests/sdcm/cluster.py, line 3364, in _create_session session.set_keyspace(keyspace)
File cassandra/cluster.py, line 3362, in cassandra.cluster.Session.set_keyspace 
File cassandra/metadata.py, line 1599, in cassandra.metadata.protect_name 
File cassandra/metadata.py, line 1626, in cassandra.metadata.maybe_escape_name 
File cassandra/metadata.py, line 1622, in cassandra.metadata.is_valid_name 
TypeError: expected string or bytes-like object
```

Add keyspace validation before create a session and add debug print to find the problem.

### DIFF
--- a/longevity_lwt_test.py
+++ b/longevity_lwt_test.py
@@ -49,8 +49,12 @@ class LWTLongevityTest(LongevityTest):
         # Wait for MVs data will be fully inserted (running on background)
         time.sleep(300)
 
+        # Data validation can be run when no nemesis, that almost not happens in case of parallel nemesis.
+        # If we even will catch period when no nemesis are running, it may happens that the nemesis will start on the
+        # middle of data validation and fail it
         if self.db_cluster.nemesis_count > 1:
             self.data_validator = MagicMock()
+            self.data_validator.keyspace_name = None
             DataValidatorEvent.DataValidator(severity=Severity.WARNING,
                                              message="Test runs with parallel nemesis. Data validator is disabled."
                                              ).publish()
@@ -82,7 +86,13 @@ class LWTLongevityTest(LongevityTest):
 
     def validate_data(self):
         node = self.db_cluster.nodes[0]
-        with self.db_cluster.cql_connection_patient(node, keyspace=self.data_validator.keyspace_name) as session:
+        if not (keyspace := self.data_validator.keyspace_name):
+            DataValidatorEvent.DataValidator(severity=Severity.WARNING,
+                                             message="Failed fo get keyspace name. Data validator is disabled."
+                                             ).publish()
+            return
+
+        with self.db_cluster.cql_connection_patient(node, keyspace=keyspace) as session:
             self.data_validator.validate_range_not_expected_to_change(session=session)
             self.data_validator.validate_range_expected_to_change(session=session)
             self.data_validator.validate_deleted_rows(session=session)
--- a/sdcm/utils/data_validator.py
+++ b/sdcm/utils/data_validator.py
@@ -209,6 +209,7 @@ class LongevityDataValidator:
             cs_profile = profiles[0]
             _, profile = get_profile_content(cs_profile)
             self._keyspace_name = profile['keyspace']
+            LOGGER.debug("Keyspace name: %s, profile: %s", self._keyspace_name, profile)
 
         return self._keyspace_name
 


yarongilor:
Wouldn't it be usefull to add a debug print between these 2 lines as well - 
```
                LOGGER.warning('Keyspace is not recognized. '
                               'Data validation can\'t be performed')
                return self._keyspace_name
```
?

fruch:
@juliayakovlev, seems like that in parallel nemesis you replaced the data validator with a magicmock object:
 https://github.com/scylladb/scylla-cluster-tests/commit/1b2544f7f72fe5088f3084d275eb21b5ae671522

you can't expect `MagicMock().keyspace_name` to return a valid string

Anyhow if you are doing those queries when it's a mock, something is off.

I would argue that replace something with a MagicMock isn't a very good way to enable/disable it.

Anyhow I'm not sure why on parallel nemesis the validation can work, since we don't have a way to wait for all the nemesis threads to finish ?

juliayakovlev:
> > > @juliayakovlev, seems like that in parallel nemesis you replaced the data validator with a magicmock object: [1b2544f](https://github.com/scylladb/scylla-cluster-tests/commit/1b2544f7f72fe5088f3084d275eb21b5ae671522)
> > > you can't expect `MagicMock().keyspace_name` to return a valid string
> > 
> > 
> > Ah, it's parallel nemesis test. I added setting MagicMock().keyspace_name = None. So with last chages a validation will not be preformed
> > > Anyhow if you are doing those queries when it's a mock, something is off.
> > > I would argue that replace something with a MagicMock isn't a very good way to enable/disable it.
> > > Anyhow I'm not sure why on parallel nemesis the validation can work, since we don't have a way to wait for all the nemesis threads to finish ?
> 
> @juliayakovlev, again can we revisit the reason why parallel nemesis can't do validation ? (and document it inside the code)

Because of data validation can be run when no nemesis, that almost not happens in case parallel nemesis. So it was decided to disable data validation for parallel nemesis. I will add a comment

fruch:
> > > > @juliayakovlev, seems like that in parallel nemesis you replaced the data validator with a magicmock object: [1b2544f](https://github.com/scylladb/scylla-cluster-tests/commit/1b2544f7f72fe5088f3084d275eb21b5ae671522)
> > > > you can't expect `MagicMock().keyspace_name` to return a valid string
> > > 
> > > 
> > > Ah, it's parallel nemesis test. I added setting MagicMock().keyspace_name = None. So with last chages a validation will not be preformed
> > > > Anyhow if you are doing those queries when it's a mock, something is off.
> > > > I would argue that replace something with a MagicMock isn't a very good way to enable/disable it.
> > > > Anyhow I'm not sure why on parallel nemesis the validation can work, since we don't have a way to wait for all the nemesis threads to finish ?
> > 
> > 
> > @juliayakovlev, again can we revisit the reason why parallel nemesis can't do validation ? (and document it inside the code)
> 
> Because of data validation can be run when no nemesis, that almost not happens in case parallel nemesis. So it was decided to disable data validation for parallel nemesis. I will add a comment

We should have a better way to wait for all nemesis to finish. for this test case, doesn't it worth doing it without the validation ?

juliayakovlev:
> > > > > @juliayakovlev, seems like that in parallel nemesis you replaced the data validator with a magicmock object: [1b2544f](https://github.com/scylladb/scylla-cluster-tests/commit/1b2544f7f72fe5088f3084d275eb21b5ae671522)
> > > > > you can't expect `MagicMock().keyspace_name` to return a valid string
> > > > 
> > > > 
> > > > Ah, it's parallel nemesis test. I added setting MagicMock().keyspace_name = None. So with last chages a validation will not be preformed
> > > > > Anyhow if you are doing those queries when it's a mock, something is off.
> > > > > I would argue that replace something with a MagicMock isn't a very good way to enable/disable it.
> > > > > Anyhow I'm not sure why on parallel nemesis the validation can work, since we don't have a way to wait for all the nemesis threads to finish ?
> > > 
> > > 
> > > @juliayakovlev, again can we revisit the reason why parallel nemesis can't do validation ? (and document it inside the code)
> > 
> > 
> > Because of data validation can be run when no nemesis, that almost not happens in case parallel nemesis. So it was decided to disable data validation for parallel nemesis. I will add a comment
> 
> We should have a better way to wait for all nemesis to finish. for this test case, doesn't it worth doing it without the validation ?

Data validation runs (1) between nemeses and (2) when all nemeses are finished. 
In case of parallel nemeses perform (1) is very problematic because if we event will catch period when no nemesis are running, it may happens that the nemesis will start on the middle of data validation and fail it.
We can try to run DV in the end of test

yarongilor (approved):
LGTM

fruch (approved):
LGTM

PR #5892, opened 2023-03-08
improvement(logcollector): Implement log rotation

temichus:
Split big SCT files logs(more than 3GB) into multiple s3 objects

task: https://github.com/scylladb/qa-tasks/issues/928

### DIFF
--- a/sdcm/log_archive.sh
+++ b/sdcm/log_archive.sh
@@ -1,0 +1,14 @@
+# $1 - file to archive
+# $2 - size splitter in Bytes
+# $3 - archive name
+wc -lc $1 | awk '{print $1,$2}' |  while read lines filesize
+do
+  parts=$(( ( filesize + $2 - 1 ) / $2 ))
+  lines_part=$(( ( lines + parts - 1 ) / parts ))
+  echo "file will be split into $parts parts, per $lines_part lines"
+  for (( i=1; i<=lines; i+=lines_part))
+  do
+    start_time=$(sed -n "${i}p" $1 | awk '{print $2,$3}' | sed -e 's/t://g;s/ /__/g;s/-/_/g;s/:/_/g;s/,/_/g;')
+    tail -n "+$i" "$1" | head -n "$lines_part"  | gzip > "${start_time}.$3.gz"
+  done
+done
--- a/sdcm/logcollector.py
+++ b/sdcm/logcollector.py
@@ -748,6 +748,12 @@ class LogCollector:
     def update_db_info(self):
         pass
 
+    def _compress_file(self, src_path: str, src_name: str) -> str:  # pylint: disable=no-self-use
+        archive_name = f"{src_name}.tar.gz"
+        with tarfile.open(archive_name, "w:gz") as tar:
+            tar.add(src_path, arcname=src_name)
+        return archive_name
+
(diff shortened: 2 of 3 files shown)


fgelcer (dismissed):
small nitpick, but LGTM

fgelcer (approved):
LGTM

sdcm/logcollector.py
```diff
@@ -965,6 +967,25 @@ class SCTLogCollector(LogCollector):
     ]
     cluster_log_type = 'sct-runner'
     cluster_dir_prefix = 'sct-runner'
+    too_big_log_size = 3*1024*1024*1024
+
+    def _create_tar_file(self, src_path: str, src_name: str) -> list[str]:
```
soyacz, line 972: why not to name it `_compress_file`? - since it is not creating `tar` file always anymore :)

sdcm/log_archive.sh
```diff
@@ -0,0 +1,14 @@
...
+# $3 - archive name
+wc -lc $1 | awk '{print $1,$2}' |  while read lines filesize
+do
+  parts=$(( ( filesize + $2 - 1 ) / $2 ))
+  lines_part=$(( ( lines + parts - 1 ) / parts ))
+  echo "file will be split into $parts parts, per $lines_part lines"
+  for (( i=1; i<=lines; i+=lines_part))
+  do
+    start_time=$(sed -n "${i}p" $1 | awk '{print $2,$3}' | sed -e 's/t://g;s/ /__/g;s/-/_/g;s/:/_/g;s/,/_/g;')
+    tail -n "+$i" "$1" | head -n "$lines_part"  | gzip > "${start_time}.$3.gz"
```
soyacz, line 12: To further minimize disk usage, this script could return commands instead of executing them. Later in python, could compress->send to s3->remove part by part.

But this might be an overkill, so I'm not insisting. Just posting an idea.

soyacz (commented):
I'd propose different solution:
1. remove `FileLog(name='sct.log',` from `SCTLogCollector` class and don't change any logic in it (maybe just rename this class)
2. create another class for collecting only `sct.log` with new logic 

This way we have all the events and additional logs around sct in separate archive always, which is good when you need to just look into events (which I often do).

sdcm/logcollector.py
fruch: I don't have strong opinions either way regarding this one, 
I don't think there a cost issue either way (seems like)

soyacz: My opinion is also not strong - I have optic fiber at my place :)

vponomaryov: If `2 different archives on S3` mean `2 separate archives on S3` then `yes`.
`sct.log` is always much bigger than all other `event` files and doesn't really need to be in a single archive with anything else.

So, I see gain in having 2 separate archives, but don't see any real harm in continue having all of it in a single archive.

soyacz (approved):
LGTM

roydahan (approved):

roydahan (approved):

roydahan (approved):

mykaul:
Might have missed a great opportunity to move to zstd instead of gzip.

soyacz:
Still can be done as separate task. Mixing with this one wouldn't make it faster.

PR #5896, opened 2023-03-09
fix(Jenkinsfile): create diffrent directory for each provision test

fruch:
Since we are now runnig on sct-runner, and we are running in provision
tests in paralell on the same worker, we need to make sure we run each
parallel on it's own direcotry.

since we could fail on locking when checking out code, or by mistake using
the wrong sct-runner, since it's base on a file with the ip in the working
directory.


fruch:
@soyacz 
FYI seems like the `jenkins/provision_azure` is broken, and I could understand why, I'll give it anther rerun on it's own

fruch:
pending on https://github.com/scylladb/scylla-cluster-tests/pull/5901, since without it provision test doesn't work at all

vponomaryov (approved):
LGTM

Issue #4472, opened 2022-02-21
disrupt_nodetool_decommission nemesis never ends if addition of a node fails


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: master

### Logs
* test_id: 490f8141-7f93-46c8-a1d9-ceffaa3688f0
* job log: https://jenkins.scylladb.com/job/scylla-master/job/longevity/job/longevity-100gb-4h-ebs-gp3-test/35/console

### Description

If we run `disrupt_nodetool_decommission` nemesis and it fails to add a node then this nemesis never ends.
It leads to the absence of the nemesis after it. So, we do not get any errors here and CI job may finish successfully.
Which is false positive.

Running `longevity-100gb-4h-ebs-gp3-test` job and starting mentioned nemesis we see following picture:

`node-2` successfully gets decommissioned in 15 minutes:

```
2022-02-20T05:12:16+00:00 db-node-490f8141-2 !    INFO |  [shard 0] storage_service - DECOMMISSIONING: starts
2022-02-20T05:12:16+00:00 db-node-490f8141-2 !    INFO |  [shard 0] storage_service - decommission[281195e3-7c7d-458a-a68c-0d73f2a1ca33]: Started decommission operation,
                                                                        removing node=10.0.1.235, sync_nodes=[10.0.3.86, 10.0.2.192, 10.0.3.235, 10.0.2.60, 10.0.1.17, 10.0.1.235], ignore_nodes=[]
2022-02-20T05:12:16+00:00 db-node-490f8141-2 !    INFO |  [shard 0] storage_service - decommission[281195e3-7c7d-458a-a68c-0d73f2a1ca33]: Added node=10.0.1.235 as leaving node, coordinator=10.0.1.235
2022-02-20T05:12:16+00:00 db-node-490f8141-2 !    INFO |  [shard 0] storage_service - decommission[281195e3-7c7d-458a-a68c-0d73f2a1ca33]: Started heartbeat_updater
2022-02-20T05:12:16+00:00 db-node-490f8141-2 !    INFO |  [shard 0] storage_service - DECOMMISSIONING: unbootstrap starts
[...]

PR #5899, opened 2023-03-10
fix(wait_for_init): fail fast on boot failure

temichus:
Fix wait_for_init to fail fast on scylla boot failures

this PR introduced the following changes into waiting-for node Up algorithm (wait_for_init_wrap wrapper):

by default:
1) Change Max severity for all SYSTEM_ERROR_EVENTS to CRITICAL (via add_severity_limit_rules)
2) Create a New filter *Startup failed* for all SYSTEM_ERROR_EVENTS events via EventsSeverityChangerFilter
3) yield - perform node boot
4) Remove  a New filter from  SYSTEM_ERROR_EVENTS 
5) Change Max severity for all SYSTEM_ERROR_EVENTS back to the previous (via add_severity_limit_rules)

you can pass the "critical_node_setup_events" keyword argument in the method  that wrapped by  wait_for_init_wrap
to change events and patterns  to your own one : [{'event': EventType, 'regex': r'REXEX'} or pass [] to back to the previous behavior

task : https://github.com/scylladb/qa-tasks/issues/70

issue: https://github.com/scylladb/scylla-cluster-tests/issues/4472

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -37,6 +37,7 @@ from collections import defaultdict
 from collections.abc import Iterator
 from dataclasses import dataclass
 from pathlib import Path
+from contextlib import ExitStack, contextmanager
 import packaging.version
 
 import yaml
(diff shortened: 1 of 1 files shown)


sdcm/cluster.py
```diff
@@ -3669,6 +3672,33 @@ def verify_node_setup(start_time):
                 cl_inst.log.error(msg)
                 raise NodeSetupTimeout(msg)
 
+        @contextmanager
+        def critical_node_setup_events():
+            default_critical_events = [{'event': DatabaseLogEvent.SYSTEM_PAXOS_TIMEOUT, 'regex': r'.*Startup failed.*'}]
```
fruch, line 3677: SYSTEM_PAXOS_TIMEOUT is just one example you were seeing
but it can be from multiple types of events, depend on the failure message.

I think "Startup failed" from all of the DatabaseLogEvent should be cought here.

sdcm/cluster.py
```diff
@@ -3678,26 +3708,27 @@ def verify_node_setup(start_time):
             cl_inst.update_db_binary(node_list, start_service=False)
             cl_inst.update_db_packages(node_list, start_service=False)
 
-        for node in node_list:
-            if isinstance(cl_inst, BaseScyllaCluster) \
-                    and not getattr(cl_inst, 'params', {}).get('use_legacy_cluster_init'):
-                init_nodes.append(node)
-                start_time = time.perf_counter()
-                node_setup(node)
+        with critical_node_setup_events():
```
fruch, line 3711: In general I'm also missing here an option to disable this, and leaving room for negative testing

example, if I want to test a flow that would end up with node failing boot, how would I call `cluster.add_nodes()` without enabling this feature

fruch: o.k. just check what you are suggest here is working, and the parameter is pass down as expected

fruch:
@temichus you have example of runs with this code ? or such that might demonstrate it in action, i.e. stopping the provision in the middle initiated by this code ?

fruch:
> @temichus you have example of runs with this code ? or such that might demonstrate it in action, i.e. stopping the provision in the middle initiated by this code ?

we would need to decide if this safe enough for going to release branches or not.

sdcm/cluster.py
fruch: o.k.

fruch (approved):
LGTM

soyacz (commented):
This events stuff is very complicated, and I don't fully understand the impact.
E.g. I still don't understand why you change `regex`. Or if WARNING event won't become critical. Also `add_severity_limit` is executed twice for each node, making `SctEvent._sct_event_types_registry.limit_rules` to grow 2x number of critical event types (maybe that's not a big deal, but still e.g. adding `change_severity_limit` method would play better here.

Would be nice if you added commit message with brief description what is the problem and how your fix tries to tackle it.

soyacz (approved):

PR #5900, opened 2023-03-12
fix(host_id): get host id from `nodetool info`

fruch:
get host id from `nodetool info` and not from our `nodetool status` implemetiontion which is much more complex since it's mapping the node based on IP, and we don't need that for getting the host_id

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -368,11 +368,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
 
     @property
     def host_id(self):
-        full_nodetool_status = self.parent_cluster.get_nodetool_status(verification_node=self)
-        for data_center in full_nodetool_status:
-            if self.ip_address in full_nodetool_status[data_center]:
-                return full_nodetool_status[data_center][self.ip_address]['host_id']
-        raise AssertionError(f"Could not find the requested node {self.ip_address} in nodetool status")
+        return self.parent_cluster.get_nodetool_info(self, publish_event=False).get('ID')
 
     @property
     def db_node_instance_type(self) -> Optional[str]:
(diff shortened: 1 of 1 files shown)


fruch:
this affect the sni usage that uses the host_id to connect to specific nodes

soyacz (dismissed):

sdcm/cluster.py
```diff
@@ -368,11 +368,15 @@ def region(self):
 
     @property
     def host_id(self):
-        full_nodetool_status = self.parent_cluster.get_nodetool_status(verification_node=self)
-        for data_center in full_nodetool_status:
-            if self.ip_address in full_nodetool_status[data_center]:
-                return full_nodetool_status[data_center][self.ip_address]['host_id']
-        raise AssertionError(f"Could not find the requested node {self.ip_address} in nodetool status")
+        info = self.run_nodetool('info', publish_event=False)
```
vponomaryov, line 371: I like the approach that is used in the existing `def get_nodetool_info(node)` method.
It is the approach where we transform whole output to python dict and then just pick up any key out of it.

So, as a result I would expect something like following:

```
class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):
    def get_nodetool_info(self, publish_event=True):
        res = self.run_nodetool('info', publish_event=publish_event)
        proper_yaml_output = "\n".join([line for line in res.stdout.splitlines() if ":" in line])
        info_res = yaml.safe_load(proper_yaml_output)
        return info_res

    @property
    def host_id(self):
        return self.get_nodetool_info(publish_event=False).get('ID')

    ...

class BaseScyllaCluster:
    @staticmethod
    def get_nodetool_info(node):
        return node.get_nodetool_info()

    ...
```

vponomaryov (changes requested):

sdcm/cluster.py
fruch: I've actully copy pasted this code from CCM, didn't see we have implementation for it

fruch: but your direction better

soyacz (approved):

PR #5868, opened 2023-02-28
fix(nemeses): support SLA nemeses for non-SLA tests

juliayakovlev:
1. SLA feature need that cluster will be configured with authenticator. Not all our longevities run with authenticator. So we need to filter out SLA nemeses.

2. Cassandra-stress default table 'keyspace1.standard1' can have different columns definition. SLA nemeses start cassandra-stress load on this table and need to use same columns definition. This commit get columns definition from cassandra-stress command (from test yaml) and run load with correct '-col' parameter.

3. To prevent parmission error 'User roleyyy has no SELECT permission on table ... or any of its parents', create the role as SUPERUSER.

4. Added temporary solution. We do not want to run SLA nemeses during not-SLA test until the feature is stable.
Example here: https://jenkins.scylladb.com/job/scylla-staging/job/yulia/job/longevity-1tb-7days-test/21/
```
Nemesis Information
Class: SlaNemeses
Name: disrupt_replace_service_level_using_detach_during_load
Status: Skipped
Duration: 1 second
Skip reason

SLA nemesis can be run during SLA test only
```

5. Added workaround for issue https://github.com/scylladb/scylla-enterprise/issues/2572
6. Due to different failures and known issues added issues numbers to the part of errors. Examples:
```
sdcm.nemesis.NemesisSubTestFailure: Step: Attach service level 'sl800_cd1b80a2' with 800 shares to role20_cd1b80a2. Validate scheduler runtime during load. Error:
 - (TestStepEvent Severity.ERROR) period_type=end event_id=2bb46fda-524a-4af5-b421-6503437afb8c during_nemesis=IncreaseSharesByAttachAnotherSlDuringLoad duration=11m38s: step=Attach service level 'sl800_cd1b80a2' with 800 shares to role20_cd1b80a2. Validate scheduler runtime during load  errors=
(Node 10.0.0.7) - Service level with higher shares got less resources unexpectedly. CPU%: 97.11. Runtime per service level group:
  sl:sl800_cd1b80a2 (shares 800): 44.06
  sl:sl500_cd1b80a2 (shares 500): 174.3
  (Possible issue scylla-enterprise#2572 or scylla-enterprise#2717)
```

```
sdcm.nemesis.NemesisSubTestFailure: Step: Run stress command and validate scheduler runtime during load. Error:
[...]

PR #5905, opened 2023-03-13
fix(autorizer): change to CassandraAuthorizer

juliayakovlev:
According to https://github.com/scylladb/scylla-docs/issues/1918 and documentation https://docs.scylladb.com/stable/operating-scylla/security/enable-authorization correct value for autorized is 'CassandraAuthorizer', not 'ScyllaAuthorizer'. Change it for SLA tests

### DIFF
--- a/test-cases/features/sl-workloads-test.yaml
+++ b/test-cases/features/sl-workloads-test.yaml
@@ -15,4 +15,4 @@ round_robin: true
 authenticator: 'PasswordAuthenticator'
 authenticator_user: cassandra
 authenticator_password: cassandra
-authorizer: 'ScyllaAuthorizer'
+authorizer: 'CassandraAuthorizer'
--- a/test-cases/features/system-sla-test.yaml
+++ b/test-cases/features/system-sla-test.yaml
@@ -17,4 +17,4 @@ round_robin: true
 authenticator: 'PasswordAuthenticator'
 authenticator_user: cassandra
 authenticator_password: cassandra
-authorizer: 'ScyllaAuthorizer'
+authorizer: 'CassandraAuthorizer'


juliayakovlev:
I did not run the tests

fgelcer (approved):
LGTM

fgelcer:
@juliayakovlev , please set labels... and perhaps we should run a short test to verify this?

fruch:
> @juliayakovlev , please set labels... and perhaps we should run a short test to verify this?

It was verified on other test cases as working, I think it's enough.

fruch:
> @juliayakovlev , please set labels... and perhaps we should run a short test to verify this?

It was verified on other test cases as working, I think it's enough.

juliayakovlev:
> > @juliayakovlev , please set labels... and perhaps we should run a short test to verify this?
> 
> It was verified on other test cases as working, I think it's enough.

@fgelcer this param was verified in new SLA tests from https://github.com/scylladb/scylla-cluster-tests/pull/5868

juliayakovlev:
> @juliayakovlev , please set labels... 

added

> and perhaps we should run a short test to verify this?

fruch (approved):

juliayakovlev:
@fruch maybe we do not need to backport it meanwhile the #5868  is not backported

fruch:
> @fruch maybe we do not need to backport it meanwhile the #5868 is not backported

I don't think it's related, this is the correct thing todo, regardless of the version

juliayakovlev:
> > @fruch maybe we do not need to backport it meanwhile the #5868 is not backported
> 
> I don't think it's related, this is the correct thing todo, regardless of the version

@fruch It's connected because of this change:

Screenshot from 2023-03-14 09-31-52

Without it the test may fail

fruch:
> > > @fruch maybe we do not need to backport it meanwhile the #5868 is not backported
> > 
> > 
> > I don't think it's related, this is the correct thing todo, regardless of the version
> 
> @fruch It's connected because of this change:
> 
> Screenshot from 2023-03-14 09-31-52
> 
> Without it the test may fail

If so I'll wait for you to test it, before backporting. 

If it does effect, then we need to report it as a bug and clarify the correct way of creating a role, with or without superuser.

PR #5908, opened 2023-03-14
fix(autorizer): change authorizer for old SLA tests

juliayakovlev:
According to https://github.com/scylladb/scylla-docs/issues/1918 and documentation https://docs.scylladb.com/stable/operating-scylla/security/enable-authorization correct value for autorized is 'CassandraAuthorizer', not 'ScyllaAuthorizer'. Change it for SLA tests

### DIFF
--- a/sla_per_user_system_test.py
+++ b/sla_per_user_system_test.py
@@ -357,9 +357,10 @@ class SlaPerUserTest(LongevityTest):
 
         read_users = []
         for share in shares:
-            read_users.append({'user': User(session=session, name='user%d' % share, password='user%d' % share).create(),
+            read_users.append({'user': User(session=session, name='user%d' % share, password='user%d' % share,
+                                            superuser=True).create(),
                                'role': Role(session=session, name='role%d' % share, password='role%d' % share,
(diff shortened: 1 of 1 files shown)


fruch:
@juliayakovlev I'm not sure I'm following what's here ? it's the title/description doesn't really describe what in the code here...

fruch:
@juliayakovlev see:
https://github.com/scylladb/scylla-dtest/pull/3084

I've reproduced it in dtest, and seem like since we were using wrong `authorizer` it was defaulting to 
AllowAllAuthorizer (default setting) - which performs no checking and so effectively grants all permissions to all roles. This must be used if AllowAllAuthenticator is the configured [authenticator](https://docs.scylladb.com/stable/operating-scylla/security/authentication.html).

which explained why it was working for before that, and when we set it correctly we started running into trouble, which this change of using superuser helps.

I still think we should be defaulting to creating roles with superuser, but selectively use the correct command to grant them the privileges they need, we shouldn't be testing only superuser roles...

fruch (approved):
LGTM

PR #5927, opened 2023-03-20
feature(k8s): k8s client based scylla pod logger

soyacz:
Currently we use `kubectl` based pod logger which sometimes causes problems because stops updating logs. This causes various issues, as ScyllaDB logs are needed by nemesis and to detect failures.

Switch to k8s python client based logger as it was proven to be more reliable.

refs: https://github.com/scylladb/qa-tasks/issues/979

### DIFF
--- a/sdcm/utils/remote_logger.py
+++ b/sdcm/utils/remote_logger.py
@@ -18,14 +18,27 @@ from abc import abstractmethod, ABCMeta
 from datetime import datetime
 from functools import cached_property
 from threading import Thread, Event as ThreadEvent
+from typing import Generator
 from multiprocessing import Process, Event
 from textwrap import dedent
 
+import kubernetes as k8s
+from dateutil.parser import isoparse
+from urllib3.exceptions import (
+    MaxRetryError,
+    ProtocolError,
+    ReadTimeoutError,
+)
+
 from sdcm import wait
 from sdcm.remote import RemoteCmdRunnerBase
+from sdcm.sct_events import Severity
 from sdcm.sct_events.decorators import raise_event_on_failure
+from sdcm.sct_events.system import TestFrameworkEvent
 from sdcm.utils.k8s import KubernetesOps
 from sdcm.utils.decorators import retrying
+
+logging.getLogger('parso.python.diff').setLevel(logging.WARNING)
 
 
 class LoggerBase(metaclass=ABCMeta):
@@ -302,6 +315,160 @@ class KubectlGeneralLogger(CommandNodeLoggerBase):
         return f"{cmd} >> {self._target_log_file} 2>&1"  # pylint: disable=protected-access
 
 
+class K8sClientLogger(LoggerBase):  # pylint: disable=too-many-instance-attributes
+    READ_REQUEST_TIMEOUT = 1200  # 20 minutes
+    RECONNECT_DELAY = 30
+    CHUNK_SIZE = 64
+
+    def __init__(self, pod: "sdcm.cluster_k8s.BasePodContainer", target_log_file: str):
+        """
+        Reads logs from a k8s pod using k8s client API and forwards it to a file.
+
+        It is self-healing and will re-open the stream if it is closed and continue reading logs from place it stopped.
+        """
+        super().__init__(target_log_file)
+        self._pod_name = pod.name
+        self._parent_cluster = pod.parent_cluster
+        self._last_log_timestamp = ""
+        self._last_read_timestamp = None
+        self._k8s_core_v1_api = KubernetesOps.core_v1_api(self._parent_cluster.k8s_cluster.get_api_client())
+        self._termination_event = ThreadEvent()
+        self._file_object = open(self._target_log_file, "a", encoding="utf-8")  # pylint: disable=consider-using-with
+        self._log_reader = None
+        self._stream = None
+        self._thread = Thread(target=self._log_loop)
+
+    def start(self):
+        self._log.info("Starting logger for pod %s", self._pod_name)
+        self._thread.start()
+
+    def stop(self, timeout=None):
+        self._log.info("Stopping logger for pod %s", self._pod_name)
+        self._termination_event.set()
+        if self._stream:
+            self._stream.close()
+        self._thread.join(timeout)
+        self._file_object.close()
+
+    def _wait_for_pod_not_pending_or_unknown(self, namespace, pod_name, timeout=60):
+        self._log.debug("Waiting for pod %s to be in not pending/unknown state", self._pod_name)
+        start_time = time.time()
+        while time.time() - start_time < timeout:
+            try:
+                pod = self._k8s_core_v1_api.read_namespaced_pod(name=pod_name, namespace=namespace)
+                pod_status = pod.status.phase
+                if pod_status.lower() not in ("pending", "unknown"):
+                    self._log.debug("Pod %s is in '%s' state", self._pod_name, pod_status)
+                    return True
+            except k8s.client.exceptions.ApiException as exc:
+                self._log.debug("Error occurred while fetching pod %s status: %s", self._pod_name, exc)
+            time.sleep(5)
+        self._log.debug("Timeout waiting for pod: '{pod_name}' is still in '{pod_status}' state.")
+        return False
+
+    @retrying(n=20, sleep_time=3, allowed_exceptions=(ConnectionError, ))
+    def _open_stream(self) -> None:
+        if self._stream:
+            self._stream.close()
+            self._stream = None
+        if self._termination_event.is_set():
+            return
+        self._wait_for_pod_not_pending_or_unknown(self._parent_cluster.namespace, self._pod_name)
+        self._log.debug("Opening log reader stream for pod %s", self._pod_name)
+        since_seconds = round(time.time() - self._last_read_timestamp + 1) if self._last_read_timestamp else None
+        try:
+            self._stream = self._k8s_core_v1_api.read_namespaced_pod_log(
+                name=self._pod_name,
+                namespace=self._parent_cluster.namespace,
+                container=self._parent_cluster.container,
+                timestamps=True,
+                follow=True,
+                since_seconds=since_seconds,
+                # NOTE: need to set a timeout, because GKE's 'pod_log' API tends to hang
+                _request_timeout=self.READ_REQUEST_TIMEOUT,
+                _preload_content=False
+            )
+
+            self._log_reader = self._read_log_lines(self._stream)
+            self._reread_logs_till_last_logged_timestamp()
+        except (k8s.client.rest.ApiException, StopIteration) as exc:
+            self._log.warning(
+                "'_open_stream()': failed to open pod log stream:\n%s", exc)
+            # NOTE: following is workaround for the error 401 which may happen due to
+            #       some config data corruption during the forced socket connection failure
+            self._k8s_core_v1_api = KubernetesOps.core_v1_api(self._parent_cluster.k8s_cluster.get_api_client())
+            raise ConnectionError(str(exc)) from None
+
+    def _reread_logs_till_last_logged_timestamp(self):
+        """Upon reconnection, reread the log till last logged timestamp is found to avoid duplication"""
+        if not self._last_log_timestamp:
+            return
+        log_timestamp, line = next(self._log_reader)
+        if isoparse(log_timestamp) >= isoparse(self._last_log_timestamp):
+            self._log.debug("Current timestap is greater than last logged timestamp. No need to reread logs")
+            self._write_log_line(log_timestamp, line)
+            return
+        self._log.debug("Rereading logs till %s", self._last_log_timestamp)
+        while True:
+            for _ in range(1000):
+                log_timestamp = next(self._log_reader)[0]
+                if log_timestamp == self._last_log_timestamp:
+                    return
+            # each 1000 lines, check if we didn't pass the date, this should not happen and possibly we can drop this code after some time
+            if isoparse(log_timestamp) >= isoparse(self._last_log_timestamp):
+                TestFrameworkEvent(source="K8sClientLogger",
+                                   message=f"Failed to find last log timestamp in the log stream for {self._pod_name} after reconnection."
+                                           f" Possibly some log lines are missed or duplicated. Last reread timestamp: {log_timestamp}",
+                                   severity=Severity.ERROR).publish()
+                return
+
+    def _write_log_line(self, timestamp, line):
+        self._last_read_timestamp = time.time()
+        self._last_log_timestamp = timestamp
+        self._file_object.write(line)
+
+    def _read_log_lines(self, stream) -> Generator[tuple[str, str], None, None]:
+        """Reads log lines. Returns a tuple of timestamp and log line"""
+        buffer = ''
+        while not self._termination_event.is_set():
+            chunk = stream.read(self.CHUNK_SIZE).decode("utf-8")
+            if not chunk:
+                break
+            buffer += chunk
+            while '\n' in buffer:
+                line, buffer = buffer.split('\n', 1)
+                try:
+                    timestamp, line = line.split(maxsplit=1)
+                    yield timestamp, line + "\n"
+                except ValueError:
+                    # sometimes returns empty line without timestamp, ignore it
+                    pass
+                except StopIteration:
+                    return
+
+    def _log_loop(self):
+        self._open_stream()
+        while not self._termination_event.is_set():
+            try:
+                timestamp, line = next(self._log_reader)
+                self._write_log_line(timestamp, line)
+            except StopIteration:
+                # pod probably has been deleted, waiting for a while and trying to reconnect
+                self._log.debug("Stream from pod %s logs has been closed, "
+                                "waiting for %s seconds and trying to reconnect",
+                                self._pod_name, self.RECONNECT_DELAY)
+                time.sleep(self.RECONNECT_DELAY)
+                self._open_stream()
+            except (MaxRetryError, ProtocolError, ReadTimeoutError, TimeoutError, AttributeError) as exc:
+                self._log.debug(
+                    "'_read_log_line()': failed to read from pod %s log stream:%s", self._pod_name, exc)
+                self._open_stream()
+            except Exception as exc:  # pylint: disable=broad-except
+                self._log.error(
+                    "'_read_log_line()': failed to read from pod %s log stream:%s", self._pod_name, exc)
+                self._open_stream()
+
+
 class KubectlClusterEventsLogger(CommandClusterLoggerBase):
     restart_delay = 30
 
@@ -404,6 +571,8 @@ def get_system_logging_thread(logs_transport, node, target_log_file):  # pylint:
         return DockerGeneralLogger(node, target_log_file)
     if logs_transport == 'kubectl':
         return KubectlGeneralLogger(node, target_log_file)
+    if logs_transport == 'k8s_client':
+        return K8sClientLogger(node, target_log_file)
     if logs_transport == 'ssh':
         if node.init_system == 'systemd':
             if 'db-node' in node.name and node.is_nonroot_install and node.remoter.run(
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -1093,6 +1093,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             self._db_log_reader_thread.stop()
         if self._alert_manager and self._alert_manager.is_alive():
             self._alert_manager.stop()
+        if self._journal_thread:
+            self._journal_thread.stop(timeout=5)
 
     @log_run_info
     def wait_till_tasks_threads_are_stopped(self, timeout: float = 120):
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -1756,7 +1756,7 @@ class BasePodContainer(cluster.BaseNode):  # pylint: disable=too-many-public-met
         return self.parent_cluster.k8s_cluster.params.get('k8s_scylla_datacenter')
 
     def start_journal_thread(self):
-        self._journal_thread = get_system_logging_thread(logs_transport="kubectl",
+        self._journal_thread = get_system_logging_thread(logs_transport="k8s_client",
                                                          node=self,
                                                          target_log_file=self.system_log)
         if self._journal_thread:


soyacz:
run: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/multitenant-eks/10/
still I see some places to improve, but overall, ready for review

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +312,128 @@ def _logger_cmd(self) -> str:
...
+    CHUNK_SIZE = 1024
+
+    def __init__(self, node: "sdcm.cluster_k8s.BasePodContainer", target_log_file: str):
+        """
+        Reads logs from a k8s pod using k8s client API and forwards it to a file.
+
+        It is self-healing and will re-open the stream if it is closed and continue reading logs from place it stopped.
+        """
+        super().__init__(target_log_file)
+        logging.getLogger('parso.python.diff').setLevel(logging.WARNING)
```
fruch, line 327: is this cause of using `dateutil.parser` ?

soyacz: I didn't track it, but I think so.

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +312,128 @@ def _logger_cmd(self) -> str:
...
+        if isoparse(timestamp) >= isoparse(self._last_log_time):
+            self._log.debug("Current timestap is greater than last logged timestamp. No need to reread logs")
+            self._write_log_line(timestamp, line)
+            return
+        self._log.debug("Rereading logs till %s", self._last_log_time)
+        while time.time() - start_time < 60:
+            # reread with timeout in case of connection freeze
+            if next(self._log_reader)[0] == self._last_log_time:
+                return
+        TestFrameworkEvent(source="K8sClientLogger",
```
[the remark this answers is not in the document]
soyacz: Rather not, I tested it with this setting because this should not happen at all and I wanted to know it. I'll reduce it to Warning.

soyacz:
retry: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/multitenant-eks/11/

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,131 @@ def _logger_cmd(self) -> str:
...
+        self._termination_event = ThreadEvent()
+        self._file_object = open(self._target_log_file, "w", encoding="utf-8")  # pylint: disable=consider-using-with
+        self._log_reader = None
+        self._thread = Thread(target=self._log_loop)
+
+    def start(self):
+        self._thread.start()
+
+    def stop(self, timeout=None):
+        self._log.info("Stopping logger for node %s", self._node.name)
```
vponomaryov, line 342: Better to add similar log message also to the `def start` method above.

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,131 @@ def _logger_cmd(self) -> str:
         return f"{cmd} >> {self._target_log_file} 2>&1"  # pylint: disable=protected-access
 
 
+class K8sClientLogger(LoggerBase):  # pylint: disable=too-many-instance-attributes
+    READ_REQUEST_TIMEOUT = 300
+    RECONNECT_DELAY = 30
+    CHUNK_SIZE = 1024
+
+    def __init__(self, node: "sdcm.cluster_k8s.BasePodContainer", target_log_file: str):
```
vponomaryov, line 321: Why do you use word `node`? This new class is suitable for getting logs out of `any K8S pod's container`.
Refering with only `node` word in K8S has different meaning, because we work with K8S cluster here, not only Scylla cluster...
I think better to rename it to `pod`.

The same is true for all the log messages below.

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,131 @@ def _logger_cmd(self) -> str:
...
+                since_seconds=since_seconds,
+                # NOTE: need to set a timeout, because GKE's 'pod_log' API tends to hang
+                _request_timeout=self.READ_REQUEST_TIMEOUT,
+                _preload_content=False
+            )
+            self._log_reader = self._read_log_lines(stream)
+            self._reread_logs_till_last_logged_timestamp()
+        except (k8s.client.rest.ApiException, StopIteration) as exc:
+            self._log.warning(
+                "'_open_stream()': failed to open ScyllaDB node log stream:\n%s", exc)
```
vponomaryov, line 369: It can be used for any K8S pod. Need to remove `ScyllaDB` word from it.

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,131 @@ def _logger_cmd(self) -> str:
...
+        self._log.info("Stopping logger for node %s", self._node.name)
+        self._termination_event.set()
+        self._thread.join(timeout)
+        self._file_object.close()
+
+    @retrying(n=20, sleep_time=3, allowed_exceptions=(ConnectionError, ))
+    def _open_stream(self) -> None:
+        self._log.debug("Opening log reader stream for node %s", self._node.name)
+        # read log since last read timestamp + 60 seconds to avoid missing logs due to some time difference between sct-runner and k8s
+        since_seconds = int(time.time() - isoparse(self._last_log_time).timestamp() +
```
vponomaryov, line 351: This approach looks very dangerous, because it depends on the time sync of SCT runner and target pod.
The diff may be more than 60 seconds and it is not guaranteed to have the same time zones.

I would really like to see usage of a single source of time -> either SCT runner or pod.

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,131 @@ def _logger_cmd(self) -> str:
...
+
+    def _log_loop(self):
+        self._open_stream()
+        while not self._termination_event.is_set():
+            try:
+                timestamp, line = next(self._log_reader)
+                self._write_log_line(timestamp, line)
+            except StopIteration:
+                # pod probably has been deleted, waiting for a while and trying to reconnect
+                self._log.debug("Stream from ScyllaDB node %s logs has been closed, "
```
vponomaryov, line 426: Same here -> s/ScyllaDB node/'%s' pod/

vponomaryov (changes requested):

sdcm/utils/remote_logger.py
soyacz: fixed, used sct runner time

soyacz:
I changed a bit code for reread logs after reconnection: don't use timeout, just from time to time check if something isn't wrong (shouldn't be but just want to make sure, possibly we can drop it when proven working ok)

sdcm/utils/remote_logger.py
vponomaryov: It stayed the same

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,135 @@ def _logger_cmd(self) -> str:
...
+
+    def __init__(self, node: "sdcm.cluster_k8s.BasePodContainer", target_log_file: str):
+        """
+        Reads logs from a k8s pod using k8s client API and forwards it to a file.
+
+        It is self-healing and will re-open the stream if it is closed and continue reading logs from place it stopped.
+        """
+        super().__init__(target_log_file)
+        logging.getLogger('parso.python.diff').setLevel(logging.WARNING)
+        self._node = node
```
vponomaryov, line 329:
```
def __init__(self, pod: "sdcm.cluster_k8s.BasePodContainer", target_log_file: str):
    ...
    self._pod_name = pod.name
    ...
```

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,135 @@ def _logger_cmd(self) -> str:
...
+
+    def stop(self, timeout=None):
+        self._log.info("Stopping logger for pod %s", self._node.name)
+        self._termination_event.set()
+        self._thread.join(timeout)
+        self._file_object.close()
+
+    @retrying(n=20, sleep_time=3, allowed_exceptions=(ConnectionError, ))
+    def _open_stream(self) -> None:
+        self._log.debug("Opening log reader stream for node %s", self._node.name)
```
vponomaryov, line 351: s/node/pod/

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,135 @@ def _logger_cmd(self) -> str:
...
+                                           f" Possibly some log lines are missed or duplicated. Last reread timestamp: {log_timestamp}",
+                                   severity=Severity.WARNING).publish()
+                return
+
+    def _write_log_line(self, timestamp, line):
+        self._last_read_timestamp = time.time()
+        self._last_log_timestamp = timestamp
+        self._file_object.write(line)
+
+    def _read_log_lines(self, stream) -> Generator[tuple[str, str], None, None]:
```
vponomaryov, line 405: From the name I expect to get a `line` on each new iteration, but on the line 409 the `chunk size` gets applied for the read operation.
Are you sure everything is ok here? Or am I missing something here?

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +313,135 @@ def _logger_cmd(self) -> str:
...
+            # each 1000 lines, check if we didn't pass the date, this should not happen and possibly we can drop this code after some time
+            if isoparse(log_timestamp) >= isoparse(self._last_log_timestamp):
+                TestFrameworkEvent(source="K8sClientLogger",
+                                   message=f"Failed to find last log timestamp in the log stream for {self._node.name} after reconnection."
+                                           f" Possibly some log lines are missed or duplicated. Last reread timestamp: {log_timestamp}",
+                                   severity=Severity.WARNING).publish()
+                return
+
+    def _write_log_line(self, timestamp, line):
+        self._last_read_timestamp = time.time()
```
vponomaryov, line 401: This `_last_read_timestamp` var keeps the time of `processing` of a line, right?
And it is not tied to a timestamp of a processed line...
So, I am not sure that it will behave correctly when we really get some reconnections after some stream hangings?

vponomaryov (changes requested):

sdcm/utils/remote_logger.py
soyacz: yes, there's `yield` 417 - so it always returns lines

sdcm/utils/remote_logger.py
soyacz: `_last_read_timestamp` is used to get into more-or-less the log place we read the most recent log and start reread.
Then timestamp of processed line is used only to compare re-read lines - when reread log line gets this timestamp, we consider log is re-read. It is nanosecond precision so should be valid marker. I don't see the other way for it (e.g. using file marker would lead to errors when pod is recreated).

vponomaryov (changes requested):

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +315,134 @@ def _logger_cmd(self) -> str:
...
+        self._last_log_timestamp = timestamp
+        self._file_object.write(line)
+
+    def _read_log_lines(self, stream) -> Generator[tuple[str, str], None, None]:
+        """Reads log in chunks and split to lines. Returns a tuple of timestamp and log line"""
+        buffer = ''
+        while True:
+            chunk = stream.read(self.CHUNK_SIZE).decode("utf-8")
```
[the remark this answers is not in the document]
soyacz: I don't think is guaranteed, as in case of closing stream lines would be generated anyway. And in case of reconnection, we would re-read them again. The real problem is that we could wait long for these lines to get to SCT.

I refactored this to use more native solution, see added commit and let me know what you think.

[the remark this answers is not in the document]
vponomaryov: Looks ok, need to see it's work in a CI job.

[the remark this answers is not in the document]
soyacz: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/multitenant-eks/12/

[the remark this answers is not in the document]
soyacz: for me, from logs perspective this job looks fine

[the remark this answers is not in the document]
vponomaryov: Logs are not complete.

Streamed logs for `loaders-2-0` last command are following:

```
...
java.io.IOException: Operation x10 on key(s) [374e38324d334f4c3030]: Error executing: (NoHostAvailableException): All host(s) tried for query failed (no host was tried)

	at org.apache.cassandra.stress.Operation.error(Operation.java:141)com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (no host was tried)

	at org.apache.cassandra.stress.Operation.timeWithRetry(Operation.java:119)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:101)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:109)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:264)
	at org.apache.cassandra.stress.StressAction$Consumer.run(StressAction.java:473)
java.io.IOException: Operation x10 on key(s) [3835304d364c37313530]: Error executing: (NoHostAvailableException): All host(s) tried for query failed (no host was tried)

	at org.apache.cassandra.stress.Operation.error(Operation.java:141)
	at org.apache.cassandra.stress.Operation.timeWithRetry(Operation.java:119)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:101)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.ja
```

Pod `namespaces/sct-loaders-2/pods/sct-loaders-2-0-pod-2` has following logs ending:

```
...
java.io.IOException: Operation x10 on key(s) [374e38324d334f4c3030]: Error executing: (NoHostAvailableException): All host(s) tried for query failed (no host was tried)

	at org.apache.cassandra.stress.Operation.error(Operation.java:141)com.datastax.driver.core.exceptions.NoHostAvailableException: All host(s) tried for query failed (no host was tried)

	at org.apache.cassandra.stress.Operation.timeWithRetry(Operation.java:119)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:101)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:109)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:264)
	at org.apache.cassandra.stress.StressAction$Consumer.run(StressAction.java:473)
java.io.IOException: Operation x10 on key(s) [3835304d364c37313530]: Error executing: (NoHostAvailableException): All host(s) tried for query failed (no host was tried)

	at org.apache.cassandra.stress.Operation.error(Operation.java:141)
	at org.apache.cassandra.stress.Operation.timeWithRetry(Operation.java:119)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:101)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:109)
	at org.apache.cassandra.stress.operations.predefined.CqlOperation.run(CqlOperation.java:264)
	at org.apache.cassandra.stress.StressAction$Consumer.run(StressAction.java:473)
ERROR 19:10:01,325 [Control connection] Cannot connect to any host, scheduling retry in 2000 milliseconds
WARN  19:10:01,390 Not using advanced port-based shard awareness with adc37a2d24a9e407c857ef164daa059e-1115191262.eu-north-1.elb.amazonaws.com:9142:de527fd7-8303-4d49-b3bb-e916c019d1c8.cql.sct-cluster-2.sct.scylladb.com because it's disabled in configuration
WARN  19:10:01,396 Not using advanced port-based shard awareness with adc37a2d24a9e407c857ef164daa059e-1115191262.eu-north-1.elb.amazonaws.com:9142:7f7c14e0-eb1c-450a-8502-703c0d0acd3d.cql.sct-cluster-2.sct.scylladb.com because it's disabled in configuration
```

So, here is the missing part:

```
                                                                                             va:264)
	at org.apache.cassandra.stress.StressAction$Consumer.run(StressAction.java:473)
ERROR 19:10:01,325 [Control connection] Cannot connect to any host, scheduling retry in 2000 milliseconds
WARN  19:10:01,390 Not using advanced port-based shard awareness with adc37a2d24a9e407c857ef164daa059e-1115191262.eu-north-1.elb.amazonaws.com:9142:de527fd7-8303-4d49-b3bb-e916c019d1c8.cql.sct-cluster-2.sct.scylladb.com because it's disabled in configuration
WARN  19:10:01,396 Not using advanced port-based shard awareness with adc37a2d24a9e407c857ef164daa059e-1115191262.eu-north-1.elb.amazonaws.com:9142:7f7c14e0-eb1c-450a-8502-703c0d0acd3d.cql.sct-cluster-2.sct.scylladb.com because it's disabled in configuration
```

In other loaders happens the same.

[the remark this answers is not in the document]
soyacz: but new code relates only to scylla pods, loaders are done different way. We can try switching loaders to use the same resort as scylla pods, but I didn't plan it for this PR.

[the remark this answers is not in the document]
fruch: I don't think we shouldn't touch the dynamic loader code for now, we can later align them to use the same. but it's not the target of this change.

fruch:
@soyacz 

please check why the eks functional test is failing:
```
E           AssertionError: Start of resharding hasn't been detected on the 'sct-cluster-us-east1-b-us-east1-2' node.
```

maybe we need to run it on latest ?

sdcm/utils/remote_logger.py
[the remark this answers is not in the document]
vponomaryov: My bad, sorry. It is my loader solution lacks data because it uses chunk size >.<

vponomaryov:
@soyacz Please, squash commits.

soyacz:
> @soyacz
> 
> please check why the eks functional test is failing:
> 
> ```
> E           AssertionError: Start of resharding hasn't been detected on the 'sct-cluster-us-east1-b-us-east1-2' node.
> ```
> 
> maybe we need to run it on latest ?

I found some issue, logger is started multiple times and sometimes it makes logs not appearing. I tried to investigate where and why this happens but I couldn't find the reason. Found only `start_task_threads` method is called sometimes after restart of node, and maybe because previous `_journal_thread` is not stopped it leads to missing logs. I added stop() in case it exists to see if this helps.

fruch:
> I found some issue, logger is started multiple times and sometimes it makes logs not appearing. I tried to investigate where and why this happens but I couldn't find the reason. Found only `start_task_threads` method is called sometimes after restart of node, and maybe because previous `_journal_thread` is not stopped it leads to missing logs. I added stop() in case it exists to see if this helps.

cloud be that it was our issue to begin with ?

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +315,131 @@ def _logger_cmd(self) -> str:
...
+        self._termination_event.set()
+        self._thread.join(timeout)
+        self._file_object.close()
+
+    @retrying(n=20, sleep_time=3, allowed_exceptions=(ConnectionError, ))
+    def _open_stream(self) -> None:
+        if self._termination_event.is_set():
+            return
+        self._log.debug("Opening log reader stream for pod %s", self._pod_name)
+        # read log since last read timestamp + 60 seconds margin
```
vponomaryov, line 355: How did you come to the 60 sec margin?

soyacz: just wanted to be some time before last read, but after thinking I don't think we need it (just a fraction of sec should be enough).

sdcm/utils/remote_logger.py
```diff
@@ -302,6 +315,131 @@ def _logger_cmd(self) -> str:
         return f"{cmd} >> {self._target_log_file} 2>&1"  # pylint: disable=protected-access
 
 
+class K8sClientLogger(LoggerBase):  # pylint: disable=too-many-instance-attributes
+    READ_REQUEST_TIMEOUT = 300
+    RECONNECT_DELAY = 30
+    CHUNK_SIZE = 128
```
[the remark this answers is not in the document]
soyacz: Still I'm playing around with it and I forgot to remove. Basically, approach with getting next(stream) to fetch the line does not work as expected - despite read timeout it hangs. I think we need to move back to chunked version.

soyacz:
Looks it finally passed the test without the log freeze. Waiting for pod readiness before reading log helped.
The only caveat I see is that sometimes we might not get logs when pod hangs in 'pending' or 'unknown' status (but in that case we should still have logs from k8s logs collected at the end).

I rerun the test with dropped 'chunking' approach, with reading line by line as with chunking we lose some log lines from the end. Let's see how it works now.

soyacz:
> I rerun the test with dropped 'chunking' approach, with reading line by line as with chunking we lose some log lines from the end. Let's see how it works now.

Failed the test. Retry with chunking (maybe we got lucky and still it is not properly fixed).
The culprit usually (maybe always) is the `test_check_operator_operability_when_scylla_crd_is_incorrect` - somehow this test is causing this freeze.

vponomaryov:
> > I rerun the test with dropped 'chunking' approach, with reading line by line as with chunking we lose some log lines from the end. Let's see how it works now.
> 
> Failed the test. Retry with chunking (maybe we got lucky and still it is not properly fixed). The culprit usually (maybe always) is the `test_check_operator_operability_when_scylla_crd_is_incorrect` - somehow this test is causing this freeze.

What exactly you found that made you think that the referenced functional test is guilty?
The problem with logs appears in any Longevity where K8S functional tests don't run at all...

soyacz:
provision_k8s-eks step failed due missing resharding. What is strange about this is that db log for node `1-1` is 750MB while others less than 150MB. In log we can see large amount of repair lines and resharding actually is not started as test induces.
I verified if it doesn't contain duplicates with `sort system.log | uniq -d` and didn't find anything relevant. 

Possibly this repairing is Scylla feature or it got into some repair loop bug. Pipeline works with version 5.0.5. I triggered pipeline with 2022.2.3. And did run multitenant scenario here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/multitenant-eks/13/

I propose to do final review & merge if no major issues are found to test it in the real battle (anyway in case of issues we can always switch back with single line change).

fruch (dismissed):
LGTM

vponomaryov:
> And did run multitenant scenario here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/multitenant-eks/13/

The job has failed on the EKS node setup. No, need to rebase it on top of the latest `master` SCT branch making the `node-setup` fix be applied and rerun the job.

For the moment it is unclear whether it works in general or not.

soyacz:
@vponomaryov it is rebased, eks setup passed: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/multitenant-eks/16/

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #5928, opened 2023-03-20
fix(diff_gr0_token_ring): Check group0 and token_ring

aleksbykov:
Decommission could fail or aborted at moment when all
    data was streamed and node was removed from token ring
    but stay in raft group0. This could affect on raft group0
    avialability because garbage voter stay in group0.
    This mostly happened for DecommissionStreamingErrMonkey.
    New procedure of 'Handling Cluster Membership Change Failures'
    was presented to handle such cases.
    
    To avoid that, compare members of token ring and
    group0 and remove from group0 garbage member with
    removenode operation. After that node could be
    terminated
    
    Issue: https://github.com/scylladb/scylladb/issues/11723
    Issue: https://github.com/scylladb/scylladb/issues/13122
    Commit: https://github.com/scylladb/scylladb/commit/c2a2996c2b0e11176930d63c419b4fdaa16dd230

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -186,6 +186,14 @@ class NodeStayInClusterAfterDecommission(Exception):
     """ raise after decommission finished but node stay in cluster"""
 
 
+class NodeCleanedAfterDecommissionAborted(Exception):
+    """ raise after decommission aborted and node cleaned from group0(Raft)"""
+
+
+class Group0MembersNotConsistenWithTokenRingMembers(Exception):
+    """ raise if set of group0 members differs from Tokein Ring members after removing Ghost members"""
+
+
 def prepend_user_prefix(user_prefix: str, base_name: str):
     return '%s-%s' % (user_prefix, base_name)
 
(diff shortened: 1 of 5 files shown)


sdcm/nemesis.py
```diff
@@ -3387,7 +3387,18 @@ def decommission_post_action():
...
-        log_follower = self.target_node.follow_system_log(patterns=["DECOMMISSIONING: unbootstrap starts"])
+        terminate_patterns = ["DECOMMISSIONING: unbootstrap starts",
+                              "DECOMMISSIONING: unbootstrap done",
+                              "becoming a group 0 non-voter",
+                              "became a group 0 non-voter",
+                              "leaving token ring",
+                              "left token ring",
+                              "Finished token ring movement"]
+
+        terminate_pattern = random.choice(terminate_patterns)
```
fruch, line 3398: Please randomize based on the nemesis seed... (or some seed)

So it would be reproducable

sdcm/cluster.py
```diff
@@ -3034,6 +3034,44 @@ def reload_config(self):
...
+        return token_ring_members
+
+    def get_group0_members(self) -> list[dict[str, str]]:
+        self.log.debug("Get group0 members")
+        group0_members = []
+        result = self.run_cqlsh("select value from system.scylla_local where key = 'raft_group0_id'",
+                                split=True, num_retry_on_failure=3)
+        if not result or len(result) < 3:
+            return []
+        raft_group0_id = result[3].strip()
```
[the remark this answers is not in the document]
roydahan: @aleksbykov ping

sdcm/nemesis.py
roydahan: It won't be reproducible anyway, it's so racy that it's very very hard to reproduce it.
But we can try I guess...

roydahan (commented):
@aleksbykov please address comments you have, we need to merge it and test ASAP.
Also, please test yourself in Individual nemesis staging job few times.

roydahan:
@aleksbykov / @temichus what's the status of this PR?
This should get the highest priority for 5.2 / 2023.1, we need to test this nemesis in many runs and make sure that the new procedure suggested by the dev works.

sdcm/nemesis.py
```diff
@@ -3085,6 +3085,12 @@ def remove_node():
             # dead_nodes_list, so the health validator terminate the job
             if exit_status != 0:
                 self.log.error(f"nodetool removenode command exited with status {exit_status}")
+                # check and clean difference between group0 and token ring,
+                garbage_host_ids = self.cluster.diff_token_ring_group0_members(verification_node)
+                self.log.debug("Difference between token ring and group0 is %s", garbage_host_ids)
+                if garbage_host_ids:
+                    self.cluster.clean_group0_garbage(verification_node)
```
soyacz, line 3092: why not to check diff token ring members when exit_status == 0?
As I understand, if we remove node without issues, there should be no garbage_host_ids, and I think we should raise error event in that case, otherwise if exit_status != 0 we can skip raising error and just clean garbage.
Maybe this is the reason of https://github.com/scylladb/scylladb/issues/13379?

aleksbykov: I think it is not related to that issue. 
You right, if exit_code == 0, the operation finished successfull, and there is no garbage. In this we don't need to do anything, just add new new and rebuild.
But if exit_status !=0, in this case we could have some garbage and should clean, but if clean will be successfull we don't need do anything, just continue as above, for that we should n't remove node from dead node list and no need to raise error, just logged (why removenode failed). if clean garbage failed, and garbage left in group0, then we raise an issue and remove from dead_node list, so ClusterHealth validator raise critical issue

aleksbykov:
> 

@roydahan , i am running single nemesis jobs and staging job to be sure it is working

sdcm/nemesis.py
soyacz: > if exit_code == 0, the operation finished successfull, and there is no garbage

I was wondering if we can be sure of that statement - what in a case of rc==0 but something unnoticed failed in Scylla's inner workings. Is there some problem with checking also in that case?

sdcm/nemesis.py
aleksbykov: yep, it is difficult to catch exactly the moment after log detected. But various line and dataset per job could help with that.

sdcm/nemesis.py
aleksbykov: i think we have another mechanism to detect that. ClusterHealthValidator, DB log err events.

soyacz (approved):

PR #5929, opened 2023-03-20
fix(longevity-100gb-4h-ebs-gp3): throttle c-s to 10K

fruch:
Since this case is using EBS slow disk which are limited to 16K IOPS, we are hitting timeout too soon if we are no going to be throttle c-s at all.

## Testing

- [] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-100gb-4h-ebs-gp3-test/9/


fruch:
This doesn't help, since the iotune output isn't matching the real disk capabilities

Issue #5608, opened 2022-12-28
Azure (hard) reboot request takes 18 minutes to be executed


ShlomiBalalis:
### Prerequisites

* [x] Are you rebased to master ?
* [ ] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: master

### Description

At `2022-12-19 12:12:31.538` a RebuildStreamingErr nemesis has started on node 2. After the rebuild starts:
```
2022-12-19T12:14:29+00:00 longevity-10gb-3h-master-db-node-eastus-2     !INFO | scylla[48672]:  [shard 0] range_streamer - Rebuild starts, nr_ranges_remaining=4288
```
We reboot the node: https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/nemesis.py#L3320, https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/nemesis.py#L3181
At first we see that we indeed see the uptime check, and the reboot starts:
```
< t:2022-12-19 12:14:30,622 f:remote_base.py  l:520  c:RemoteLibSSH2CmdRunner p:DEBUG > Running command "uptime -s"...
< t:2022-12-19 12:14:31,202 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > 2022-12-19 08:23:02
< t:2022-12-19 12:14:31,203 f:base.py         l:142  c:RemoteLibSSH2CmdRunner p:DEBUG > Command "uptime -s" finished with status 0
< t:2022-12-19 12:14:31,209 f:cluster.py      l:1008 c:sdcm.cluster_azure   p:DEBUG > Node longevity-10gb-3h-master-db-node-763ea095-eastus-2 [20.102.50.246 | 10.0.0.6] (seed: False): Hardly rebooting node
```
The Azure reboot function print a very specific info message that help identify that the azure request specifically took a while:
https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/cluster_azure.py#L118
...
https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/provision/azure/virtual_machine_provider.py#L142

From the sct log:
```
< t:2022-12-19 12:14:31,775 f:virtual_machine_provider.py l:146  c:sdcm.provision.azure.virtual_machine_provider p:INFO  > Waiting for reboot of instance: longevity-10gb-3h-master-db-node-eastus-2...
[...]

PR #5931, opened 2023-03-20
fix(azure): add more timeout for hard reboots

fruch:
since Azure can schedule the restart up to 15-18min into the
future, we need to wait more time for it to finish.
Fix: #5608

## Testing

- [x] - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-10gb-3h-azure-test/15/

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3399,8 +3399,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             disrupt_func_kwargs={"target_node": self.target_node, "hard": True, "verify_ssh": True},
             delay=0
         )
-
-        ParallelObject(objects=[trigger, watcher], timeout=1200).call_objects()
+        timeout = 1200
+        if self.cluster.params.get('cluster_backend') == 'azure':
+            timeout += 1200  # Azure reboot can take up to 20min to initiate
+        ParallelObject(objects=[trigger, watcher], timeout=timeout).call_objects()
         if new_node := decommission_post_action():
             new_node.run_nodetool("rebuild")
         else:
@@ -3418,7 +3420,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.target_node.run_nodetool, sub_cmd="repair", warning_event_on_exception=(Exception,), retry=0, timeout=600,
         )
         log_follower = self.target_node.follow_system_log(patterns=["Repair 1 out of"])
-
+        timeout = 1200
+        if self.cluster.params.get('cluster_backend') == 'azure':
+            timeout += 1200  # Azure reboot can take up to 20min to initiate
         watcher = partial(
             self._call_disrupt_func_after_expression_logged,
             log_follower=log_follower,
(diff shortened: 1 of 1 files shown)


fgelcer (approved):
LGTM

soyacz (approved):

roydahan:
@fruch I think this PR is wrong.
It shouldn't take so long to reboot a node and something is off and need to be reported if this is the case as you mentioned above.

In a run I'm looking at, I see that a boot takes 6 minutes (which still sounds a lot to me):
```
Mar 23 18:06:45 longevity-10gb-3h-2023-1-db-node-eastus-8 systemd[1]: scylla-image-setup.service: Consumed 14.818s CPU time.
-- Boot 17f33b6a12f34833915e651c6b79a865 --
Mar 23 18:12:20 longevity-10gb-3h-2023-1-db-node-eastus-8 systemd[1]: Starting Scylla Cloud Image Setup service...
...
...
Mar 23 18:12:50 longevity-10gb-3h-2023-1-db-node-eastus-8 scylla[904]:  [shard 0] init - Scylla version 2023.1.0~rc3-0.20230321.80de75947b7a initialization completed.

```

This is a [run](https://argus.scylladb.com/test/495648b9-1123-4547-a33b-eef73457c1db/runs?additionalRuns[]=302e60f7-0150-4882-9525-5ba9253ebb18) without your fix and it should have passed even without your fix (1200 timeout was good enough).

However, it looks like the actual issue is something wrong (by parallel_object?), we rebooted the host twice although the nemesis didn't planned to do it and it's not clear if the watcher missed the first boot of the node or not.
Check out the logs of node-8.
It had boot in 17:46:56 and another one in 18:06.

fruch:
> @fruch I think this PR is wrong.
> It shouldn't take so long to reboot a node and something is off and need to be reported if this is the case as you mentioned above.
> 
> In a run I'm looking at, I see that a boot takes 6 minutes (which still sounds a lot to me):
> ```
> Mar 23 18:06:45 longevity-10gb-3h-2023-1-db-node-eastus-8 systemd[1]: scylla-image-setup.service: Consumed 14.818s CPU time.
> -- Boot 17f33b6a12f34833915e651c6b79a865 --
> Mar 23 18:12:20 longevity-10gb-3h-2023-1-db-node-eastus-8 systemd[1]: Starting Scylla Cloud Image Setup service...
> ...
> ...
> Mar 23 18:12:50 longevity-10gb-3h-2023-1-db-node-eastus-8 scylla[904]:  [shard 0] init - Scylla version 2023.1.0~rc3-0.20230321.80de75947b7a initialization completed.
> 
> ```
> 
> This is a [run](https://argus.scylladb.com/test/495648b9-1123-4547-a33b-eef73457c1db/runs?additionalRuns[]=302e60f7-0150-4882-9525-5ba9253ebb18) without your fix and it should have passed even without your fix (1200 timeout was good enough).
> 
> However, it looks like the actual issue is something wrong (by parallel_object?), we rebooted the host twice although the nemesis didn't planned to do it and it's not clear if the watcher missed the first boot of the node or not.
> Check out the logs of node-8.
> It had boot in 17:46:56 and another one in 18:06.
> 
> 
> 

It nothing todo with parallel, we've seen it multiple times, azure doesn't gurrentee reboot would be immediate, and they even write it down that the operation is scheduled to a later time.

Now that our account is "real", maybe we can open a support ticket. But until then, this change is the best we can do.

roydahan:
Do they guarantee soft reboot?
Is so, we can do that instead for Azure.
I still don't understand what caused the reboot at 17:46.

fruch:
> Do they guarantee soft reboot?
> Is so, we can do that instead for Azure.
> I still don't understand what caused the reboot at 17:46.

They don't guarantee anything, it's not written anywhere that I could find, it's just an observation of how the APIs behave.

If it's soft reboot vs. timeout of 10 more minutes, I don't think it's a big difference for us.

I didn't into the specific case, I would assume you are seeing the next nemesis in action, there's no retries that I know of in that sequence (it's not an ssh command, it's an API call, and parallel object doesn't have retries)

fruch:
I suggest we let it run for a while and see, if there are more issues with this fix or if it's enough

PR #5934, opened 2023-03-21
feature(aws): add suport to multi az

soyacz:
We need to support multi-az deployments in SCT to be able to cover new
test cases. Especially we aim for grow/shrink cluster tests when working
in multi-rack deployments.

Add support to multi az in AWS provisioning code (old provision). To use
multi-az, specify multiple coma-separated availability zones
(e.g. `a,b,c`).
When multi-az is specified, job's provision step is skipped.

`GrowShrinkClusterNemesis` is able to work with multi-az on AWS
backend. When `rack` is set to `None` in `discrupt_grow_shrink_cluster`,
it will be growing cluster evenly, otherwise grow only in specified rack.
Adapted 3 jobs to use multi-az on AWS.

refs: https://github.com/scylladb/qa-tasks/issues/1070

### DIFF
--- a/jenkins-pipelines/longevity-cdc-100gb-4h.jenkinsfile
+++ b/jenkins-pipelines/longevity-cdc-100gb-4h.jenkinsfile
@@ -6,6 +6,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 longevityPipeline(
     backend: 'aws',
     region: 'eu-west-1',
+    availability_zone: 'a,b,c',
     test_name: 'longevity_test.LongevityTest.test_custom_time',
     test_config: 'test-cases/longevity/longevity-cdc-100gb-4h.yaml'
 
--- a/jenkins-pipelines/longevity-topology-changes-3h.jenkinsfile
+++ b/jenkins-pipelines/longevity-topology-changes-3h.jenkinsfile
@@ -6,6 +6,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 longevityPipeline(
     backend: 'aws',
     region: 'eu-west-1',
+    availability_zone: 'a,b,c',
     test_name: 'longevity_test.LongevityTest.test_custom_time',
     test_config: '''["test-cases/longevity/longevity-topology-changes-3h.yaml"]''',
 
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3125,12 +3125,16 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
 
         # NOTE: following is needed in case of K8S where we init multiple DB clusters first
         #       and only then we add nodes to it calling code in parallel.
+        # round-robin racks when backend is aws and multiple az's are specified
+        azs = len(self.params.get('availability_zone').split(",")) if self.params.get('cluster_backend') == 'aws' else 1
         if add_nodes:
             if isinstance(n_nodes, list):
                 for dc_idx, num in enumerate(n_nodes):
-                    self.add_nodes(num, dc_idx=dc_idx, enable_auto_bootstrap=self.auto_bootstrap)
+                    for az_index in range(num):
+                        self.add_nodes(1, dc_idx=dc_idx, rack=az_index % azs, enable_auto_bootstrap=self.auto_bootstrap)
             elif isinstance(n_nodes, int):  # legacy type
-                self.add_nodes(n_nodes, enable_auto_bootstrap=self.auto_bootstrap)
+                for az_index in range(n_nodes):
+                    self.add_nodes(1, rack=az_index % azs, enable_auto_bootstrap=self.auto_bootstrap)
             else:
                 raise ValueError('Unsupported type: {}'.format(type(n_nodes)))
             self.run_node_benchmarks()
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3517,11 +3517,16 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     def decommission_nodes(self, add_nodes_number, rack, is_seed: Optional[Union[bool, DefaultValue]] = DefaultValue,
                            dc_idx: Optional[int] = None):
-        for _ in range(add_nodes_number):
+        for idx in range(add_nodes_number):
             if self._is_it_on_kubernetes():
+                if rack is None and self._is_it_on_kubernetes():
+                    rack = 0
                 self.set_target_node(rack=rack, is_seed=is_seed, allow_only_last_node_in_rack=True)
             else:
-                self.set_target_node(is_seed=is_seed, dc_idx=dc_idx)
+                racks_count = len(self.tester.params.get("availability_zone").split(","))
+                rack_idx = rack if rack is not None else idx % racks_count
+                # if rack is not specified, round-robin racks
+                self.set_target_node(is_seed=is_seed, dc_idx=dc_idx, rack=rack_idx)
             self.log.info("Next node will be removed %s", self.target_node)
 
             try:
@@ -3533,33 +3538,34 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                           f'{str(exc)}').publish()
 
     def disrupt_grow_shrink_cluster(self):
-        self._grow_cluster(rack=0)
-        self._shrink_cluster(rack=0)
+        self._grow_cluster(rack=None)
+        self._shrink_cluster(rack=None)
 
     def disrupt_grow_shrink_new_rack(self):
+        if not self._is_it_on_kubernetes():
+            raise UnsupportedNemesis("Adding new rack is not supported for non-k8s Scylla clusters")
         rack = max(self.cluster.racks) + 1
         self._grow_cluster(rack)
         self._shrink_cluster(rack)
 
-    def _grow_cluster(self, rack=0):
-        if rack > 0:
-            if not self._is_it_on_kubernetes():
-                raise UnsupportedNemesis("SCT rack functionality is implemented only on kubernetes")
-            self.log.info("Rack deletion is not supported on kubernetes yet. "
-                          "Please see https://github.com/scylladb/scylla-operator/issues/287")
-
+    def _grow_cluster(self, rack=None):
+        if rack is None and self._is_it_on_kubernetes():
+            rack = 0
         add_nodes_number = self.tester.params.get('nemesis_add_node_cnt')
+        racks_count = len(self.tester.params.get("availability_zone").split(","))
         self.log.info("Start grow cluster on %s nodes", add_nodes_number)
         InfoEvent(message=f"Start grow cluster on {add_nodes_number} nodes").publish()
-        for _ in range(add_nodes_number):
-            InfoEvent(message=f'GrowCluster - Add New node to {rack} rack').publish()
-            added_node = self.add_new_node(rack=rack)
+        for idx in range(add_nodes_number):
+            # if rack is not specified, round-robin racks to spread nodes evenly
+            rack_idx = rack if rack is not None else idx % racks_count
+            InfoEvent(message=f'GrowCluster - Add New node to {rack_idx} rack').publish()
+            added_node = self.add_new_node(rack=rack_idx)
             self.unset_current_running_nemesis(added_node)
             InfoEvent(message=f'GrowCluster - Done adding New node {added_node.name}').publish()
         self.log.info("Finish cluster grow")
         time.sleep(self.interval)
 
-    def _shrink_cluster(self, rack=0):
+    def _shrink_cluster(self, rack=None):
         add_nodes_number = self.tester.params.get('nemesis_add_node_cnt')
         self.log.info("Start shrink cluster by %s nodes", add_nodes_number)
         InfoEvent(message=f'Start shrink cluster by {add_nodes_number} nodes').publish()
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -1312,7 +1312,8 @@ class SCTConfiguration(dict):
              help="""Use legacy cluster initialization with autobootsrap disabled and parallel node setup"""),
         dict(name="availability_zone", env="SCT_AVAILABILITY_ZONE",
              type=str,
-             help="Availability zone to use. Same for multi-region scenario."),
+             help="""Availability zone to use. Specify multiple (comma separated) to deploy resources to multi az (works on AWS).
+                  "Same for multi-region scenario."""),
         dict(name="aws_fallback_to_next_availability_zone", env="SCT_AWS_FALLBACK_TO_NEXT_AVAILABILITY_ZONE",
              type=boolean,
              help="""Try all availability zones one by one in order to maximize the chances of getting
--- a/test-cases/longevity/longevity-cdc-100gb-4h.yaml
+++ b/test-cases/longevity/longevity-cdc-100gb-4h.yaml
@@ -1,11 +1,13 @@
 test_duration: 290
 
+pre_create_keyspace: "CREATE KEYSPACE cdc_test WITH replication = {'class': 'NetworkTopologyStrategy', 'eu-westscylla_node_west': '3'}  AND durable_writes = true;"
 stress_cmd: [ "cassandra-stress user no-warmup profile=/tmp/cdc_profile.yaml ops'(insert=2,read1=1,update_number=1,update_name=1,delete1=1)' cl=QUORUM duration=200m -mode cql3 native -rate threads=100",
               "cassandra-stress user no-warmup profile=/tmp/cdc_profile_preimage.yaml ops'(insert=2,read1=1,update_number=1,update_name=1,delete1=1)' cl=QUORUM duration=200m -mode cql3 native -rate threads=100",
               "cassandra-stress user no-warmup profile=/tmp/cdc_profile_postimage.yaml ops'(insert=2,read1=1,update_number=1,update_name=1,delete1=1)' cl=QUORUM duration=200m -mode cql3 native -rate threads=100",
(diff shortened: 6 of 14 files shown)


sdcm/cluster.py
```diff
@@ -3125,15 +3125,18 @@ def __init__(self, cluster_uuid=None, cluster_prefix='cluster', node_prefix='nod
...
+                    for az_index in range(num):
+                        self.add_nodes(1, dc_idx=dc_idx, rack=az_index % azs, enable_auto_bootstrap=self.auto_bootstrap)
             elif isinstance(n_nodes, int):  # legacy type
-                self.add_nodes(n_nodes, enable_auto_bootstrap=self.auto_bootstrap)
+                for az_index in range(n_nodes):
+                    self.add_nodes(1, rack=az_index % azs, enable_auto_bootstrap=self.auto_bootstrap)
             else:
                 raise ValueError('Unsupported type: {}'.format(type(n_nodes)))
-            self.run_node_benchmarks()
+            # self.run_node_benchmarks()
```
fruch, line 3139: Put this into it's own commit

soyacz: this is mistake, shouldn't have been commited

soyacz:
test drive: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-mini-test-1h-test/2/

fgelcer:
> test drive: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-mini-test-1h-test/2/

i'm also giving it a ride [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/disney_elasticity/job/1tb-grow-shrink/11/)

soyacz:
I selected 3 longevities to use multi-az https://github.com/scylladb/scylla-cluster-tests/pull/5934/commits/696d48c224614d074e03a33f04cc07d4c7453b9a
@roydahan please review if selection is right

soyacz:
Ready for review.
I propose doing below things in separate PR's to enable testing multi-az quicker:
1. Adapt new AWS provisioning to support multi-az
2. Make `AddRemoveRackNemesis` work with AWS backend

test-cases/longevity/longevity-sla-100gb-4h.yaml
fruch, line 1: I'm not sure this one is stable enough for being a basis for this, also it's a enterprise feature only.

@roydahan maybe one of one running weekly ?

test-cases/longevity/longevity-topology-changes-3h.yaml
```diff
@@ -22,6 +23,7 @@ nemesis_class_name: 'SisyphusMonkey'
 nemesis_selector: ['topology_changes']
 nemesis_interval: 5
 nemesis_filter_seeds: false
+nemesis_add_node_cnt: 2
```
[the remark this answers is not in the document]
soyacz: because 2 az's (4 db nodes)

test-cases/longevity/longevity-sla-100gb-4h.yaml
fgelcer: > maybe one of one running weekly ?

i agree

fgelcer (commented):
agree with the `longevity-sla` not being stable enough, so i would suggest using the longevity 4h for it... or add a short longevity to have this configuration, and add it to the weekly trigger (think the 2nd is the best option)

fruch:
> agree with the `longevity-sla` not being stable enough, so i would suggest using the longevity 4h for it... or add a short longevity to have this configuration, and add it to the weekly trigger (think the 2nd is the best option)

we specifically said we don't want it on short longevity that are automatically triggered 
until we'll know how stable the multi AZ runs, we don't want to generate too much noise for ourselves

test-cases/longevity/longevity-sla-100gb-4h.yaml
soyacz: I tried to find some job with 6 nodes.
I think we could adapt `test-cases/longevity/longevity-cdc-100gb-4h.yaml`. To make it working, modify   `cdc_profile.yaml` to use Networking Strategy ks or `pre_create_keyspace` with proper replication strategy.

soyacz: @fruch @fgelcer ^^

fruch:
@roydahan, waiting for you input on which case to add this one (and a review in general)

soyacz:
@fruch @fgelcer I replaced SLA with `test-cases/longevity/longevity-cdc-100gb-4h.yaml`. Triggered weekly, 3AZ's.
@roydahan please review if tests selection looks ok.

sdcm/sct_config.py
```diff
@@ -1312,7 +1312,8 @@ class SCTConfiguration(dict):
              help="""Use legacy cluster initialization with autobootsrap disabled and parallel node setup"""),
         dict(name="availability_zone", env="SCT_AVAILABILITY_ZONE",
              type=str,
-             help="Availability zone to use. Same for multi-region scenario."),
+             help="""Availability zone to use. Specify multiple (comma separated) to deploy resources to multi az (works on AWS).
```
vponomaryov, line 1315: From the pipeline changes I would expect it to work also in Azure.
So, is it correct that only AWS really works now?

sdcm/nemesis.py
```diff
@@ -3518,10 +3518,7 @@ def decommission_node(self, node):
     def decommission_nodes(self, add_nodes_number, rack, is_seed: Optional[Union[bool, DefaultValue]] = DefaultValue,
                            dc_idx: Optional[int] = None):
         for _ in range(add_nodes_number):
-            if self._is_it_on_kubernetes():
-                self.set_target_node(rack=rack, is_seed=is_seed, allow_only_last_node_in_rack=True)
-            else:
-                self.set_target_node(is_seed=is_seed, dc_idx=dc_idx)
+            self.set_target_node(rack=rack, is_seed=is_seed, allow_only_last_node_in_rack=True)
```
vponomaryov, line 3521: Why in non-K8S cases we don't allow non-last node in a rack to be disrupted anymore?

sdcm/nemesis.py
```diff
@@ -3533,33 +3530,32 @@ def decommission_nodes(self, add_nodes_number, rack, is_seed: Optional[Union[boo
...
-    def _grow_cluster(self, rack=0):
-        if rack > 0:
-            if not self._is_it_on_kubernetes():
-                raise UnsupportedNemesis("SCT rack functionality is implemented only on kubernetes")
-            self.log.info("Rack deletion is not supported on kubernetes yet. "
-                          "Please see https://github.com/scylladb/scylla-operator/issues/287")
-
+    def _grow_cluster(self, rack=None):
         add_nodes_number = self.tester.params.get('nemesis_add_node_cnt')
+        racks_count = len(self.tester.params.get("availability_zone").split(","))
```
vponomaryov, line 3545: Current approach for creating Scylla racks on K8S backends in SCT doesn't depend on the AZs.
So, I expect it to have some problems on EKS backend where we specify always 2 AZs just to workaround EKS specific requirements, but use, de-facto, only first one.

sdcm/cluster.py
```diff
@@ -3125,12 +3125,15 @@ def __init__(self, cluster_uuid=None, cluster_prefix='cluster', node_prefix='nod
 
         # NOTE: following is needed in case of K8S where we init multiple DB clusters first
         #       and only then we add nodes to it calling code in parallel.
+        azs = len(self.params.get('availability_zone').split(","))
```
vponomaryov, line 3128: As I wrote below, I expect it to have problems with EKS backend.

sdcm/nemesis.py
soyacz: I think I misunderstood that - reverting this part.

sdcm/nemesis.py
soyacz: I don't get what is the problem. Replaced code:
```
        for _ in range(add_nodes_number):
            InfoEvent(message=f'GrowCluster - Add New node to {rack} rack').publish()
            added_node = self.add_new_node(rack=rack)
```
So rack was used in k8s tests. 
The only difference I see is that we grow cluster in all racks when on k8s (when rack is not specified as in `disrupt_grow_shrink_cluster`)

sdcm/sct_config.py
soyacz: Only AWS works now. Pipeline just disables `provision step` when multi-az is used.
In case Azure/GCE was used I expect it to fail at next stage.

sdcm/cluster.py
soyacz: this is done to add nodes to each rack uniformly, so currently all the nodes in k8s are in the same rack?

vponomaryov: > this is done to add nodes to each rack uniformly, so currently all the nodes in k8s are in the same rack?

Yes.

sdcm/nemesis.py
vponomaryov: We use single rack in K8S even when we specify 2 AZs in case of EKS.
The only situation when we create second rack it is using `disrupt_grow_shrink_new_rack` nemesis.

soyacz: ok, I changed the code so on k8s will use rack=0 when is not specified (`None`).

vponomaryov (approved):
LGTM

jenkins-pipelines/longevity-cdc-100gb-4h.jenkinsfile
```diff
@@ -6,6 +6,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 longevityPipeline(
     backend: 'aws',
     region: 'eu-west-1',
+    availability_zone: 'a,b,c',
```
roydahan, line 9: Why do we put it as a pipeline parameter and not in the test_yaml?
It's hard to track it like this.

jenkins-pipelines/longevity-topology-changes-3h.jenkinsfile
```diff
@@ -6,6 +6,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 longevityPipeline(
     backend: 'aws',
     region: 'eu-west-1',
+    availability_zone: 'a,b',
```
roydahan, line 9: I think that we should put it in the yaml and set it to 3 az.

test-cases/longevity/longevity-topology-changes-3h.yaml
roydahan, line 26: It's a weird case to test RF=3 with 2 AZs and adding 2 nodes.
I don't even know what we should expect.
For now, let's go with the recommended and supported case: RF=3, 3 AZs, adding 3 nodes every time.

roydahan: Also, let's change:
n_db_nodes to 3
seeds_num: 3

@aleksbykov please review these suggested changes as well.

jenkins-pipelines/longevity-cdc-8h-multi-dc-topology-changes.jenkinsfile
```diff
@@ -6,6 +6,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 longevityPipeline(
     backend: 'aws',
     region: '''["eu-west-1", "us-east-1"]''',
+    availability_zone: 'a,b',
```
roydahan, line 9: Let's move to the test yaml and set the AZ and RF to be the same.

roydahan (changes requested):

jenkins-pipelines/longevity-cdc-100gb-4h.jenkinsfile
soyacz: @roydahan 
because pipelines define availability_zone defaults - so it will override yaml.
Solutions I see:
1. add yet another param to test yaml which will be confusing (setting az in jenkins pipeline will differ from actual az's)
2. add availability_zone to test yaml and leave jenkins params as now (so it's clearer when reading test yaml) - my preference.

@roydahan please let me know if you have another idea or which you prefer

jenkins-pipelines/longevity-cdc-8h-multi-dc-topology-changes.jenkinsfile
soyacz: this test has 4 nodes per DC, so it will make cluster unbalanced from the start in that case. 
I change multidc case to `longevity-counters-multidc.yaml`

jenkins-pipelines/longevity-cdc-100gb-4h.jenkinsfile
fruch: Also the current disable of the new provision step depends on this pipeline parameter.

So we need some indication on the pipeline level as well

soyacz:
@roydahan I applied changes, please see my responses. 
Also note that I changed multi-dc scenario to `longevity-counters-multidc.yaml` as previous was using 4 nodes per dc.

jenkins-pipelines/longevity-cdc-100gb-4h.jenkinsfile
roydahan: I see.
I don't like option 1 either.
The problem with option 2 is that one may want to change that and he'll change that only in the yaml but it won't take effect since the pipeline overrides it, right?
(In that case, it doesn't really help and better to leave as is).

test-cases/longevity/longevity-cdc-100gb-4h.yaml
```diff
@@ -1,11 +1,13 @@
 test_duration: 290
 
+pre_create_keyspace: "CREATE KEYSPACE cdc_test WITH replication = {'class': 'NetworkTopologyStrategy', 'eu-westscylla_node_west': '3'}  AND durable_writes = true;"
```
roydahan, line 3: @soyacz / @fruch another problem I see now is that we're going to be tightly coupled with the "region" mentioned here.
It means that if someone runs the test from a different region (can be changed from pipeline params), the test won't work.

roydahan (dismissed):
Overall it looks good to me.
There are still some usability issues making our life a bit harder.
It shouldn't block the PR, so we can start using it, but it should be addressed as part of a followup task.

sdcm/nemesis.py
```diff
@@ -3533,33 +3539,35 @@ def decommission_nodes(self, add_nodes_number, rack, is_seed: Optional[Union[boo
...
 
     def disrupt_grow_shrink_cluster(self):
-        self._grow_cluster(rack=0)
-        self._shrink_cluster(rack=0)
+        self._grow_cluster(rack=None)
+        self._shrink_cluster(rack=None)
 
     def disrupt_grow_shrink_new_rack(self):
+        if not self._is_it_on_kubernetes():
+            raise UnsupportedNemesis("Adding new rack is not supported for non-k8s Scylla clusters")
```
vponomaryov, line 3547: Remove it. It is false statement and unneeded skip of nemesis for K8S.

The `racks` feature is supported in K8S. It is not tied to the real AZs anyhow. Because we can create several "racks" having them in one AZ and idea under it may be to use different node configurations (i.e. asymmetric cluster). Only affinity rules and node selectors, when specified in a K8S `rack` configuration, imply some of AZs to be picked up.

sdcm/nemesis.py
```diff
@@ -3517,11 +3517,17 @@ def decommission_node(self, node):
 
     def decommission_nodes(self, add_nodes_number, rack, is_seed: Optional[Union[bool, DefaultValue]] = DefaultValue,
                            dc_idx: Optional[int] = None):
-        for _ in range(add_nodes_number):
+        for idx in range(add_nodes_number):
             if self._is_it_on_kubernetes():
+                if rack is None and self._is_it_on_kubernetes():
+                    # currently despite setting multi az on k8s, we still use only one rack
```
vponomaryov, line 3523: Remove this comment, as I wrote below, K8S `racks` are not tied anyhow to real platform AZs.
You can leave `rack = 0` here, but really correct approach would be to read the `scyllacluster` K8S object spec and get the num of configured `K8S racks` out of it.

sdcm/nemesis.py
```diff
@@ -3533,33 +3539,35 @@ def decommission_nodes(self, add_nodes_number, rack, is_seed: Optional[Union[boo
...
-    def _grow_cluster(self, rack=0):
-        if rack > 0:
-            if not self._is_it_on_kubernetes():
-                raise UnsupportedNemesis("SCT rack functionality is implemented only on kubernetes")
-            self.log.info("Rack deletion is not supported on kubernetes yet. "
-                          "Please see https://github.com/scylladb/scylla-operator/issues/287")
-
+    def _grow_cluster(self, rack=None):
+        if rack is None and self._is_it_on_kubernetes():
+            # currently despite setting multi az on k8s, we still use only one rack
```
vponomaryov, line 3554: Same here.

vponomaryov (changes requested):

test-cases/longevity/longevity-cdc-100gb-4h.yaml
soyacz: @roydahan I agree, this problem exists currently anyway and I already created some task regarding this: https://github.com/scylladb/scylla-cluster-tests/issues/5969

sdcm/nemesis.py
soyacz: `if not self._is_it_on_kubernetes()` - for k8s it is not skipped.

vponomaryov: You are right, I miss-looked.

vponomaryov (dismissed):
LGTM

fruch:
@soyacz 

Test case linting is failing:
```
10:31:38      raise ValueError(res)
10:31:38  ValueError: Unsupported config option/s found:
10:31:38  	 * 'availability_zones: a,b,c'
```

soyacz:
@fruch  checks fixed

fruch (approved):

PR #5937, opened 2023-03-22
fix(sct.log): sct.log tarball is missing the test_id

temichus:
since #5892 the tarball doesn't have the test_id in it's name, it's a bit annoying when you are looking at multiple jobs at the same time, and it's not match with the test_id as it was before


vponomaryov (approved):
LGTM

PR #5944, opened 2023-03-23
improvement(nemesis.py): Covering MVs and Indexes and USING TIMESTAMP

k0machi:
This change adds a random chance to use a "USING TIMESTAMP"
clause instead of normal 50% of the partition, inteded to cover the
usage of both "USING TIMESTAMP" and introduce potential variability into
partition makeup that could potentially reflect on mvs/gsi updates due
to large deletions.

Task: scylladb/qa-tasks#1009

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -21,6 +21,7 @@ import copy
 import datetime
 import inspect
 import logging
+import math
 import os
 import random
 import re
@@ -149,6 +150,14 @@ class LogContentNotFound(Exception):
 
 
 class LdapNotRunning(Exception):
+    pass
+
+
+class TimestampNotFound(Exception):
+    pass
+
+
+class PartitionNotFound(Exception):
     pass
 
 
@@ -2140,6 +2149,34 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self.log.debug(f'Partitions for delete: {partitions_for_delete}')
         return partitions_for_delete
 
+    def get_random_timestamp_from_partition(self, ks_cf, pkey, partition_percentage=0.25) -> tuple[int, int]:
+        """
+            Get a write timestamp from a "pivot" row inside a single partition.
+            partition_percentage controls where the "pivot" is, default is 25% of the partition size (total number of rows)
+            Returns timestamp and the clustering key value as tuple
+        """
+        with self.cluster.cql_connection_patient(node=self.target_node) as session:
+            number_of_rows = session.execute(
+                SimpleStatement(f"select count(ck) from {ks_cf} where pk = {pkey}")).one().system_count_ck
+            fetch_limit = max(math.ceil(number_of_rows * partition_percentage), 11)
+            self.log.debug(
+                "[%s_using_timestamp] Partition size: %s, fetching up to %s",
+                self.get_disrupt_name(),
+                number_of_rows,
+                fetch_limit
+            )
+            partition = session.execute(SimpleStatement(
+                f"select pk, ck from {ks_cf} where pk = {pkey} limit {fetch_limit}")).all()
+            delete_mark = partition[-1].ck
+            timestamp = session.execute(
+                SimpleStatement(f"select writetime(v) from {ks_cf} where pk = {pkey} and ck = {delete_mark}")).one().writetime_v
+            if not timestamp:
+                message = f"Unable to get writetime for row (pk = {pkey}, ck = {delete_mark})"
+                self.log.error(message)
+                raise TimestampNotFound(message)
+
+            return timestamp, delete_mark
+
     def run_deletions(self, queries, ks_cf):
         for cmd in queries:
             self.log.debug(f'delete query: {cmd}')
@@ -2147,6 +2184,26 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 session.execute(cmd, timeout=3600)
 
         self.target_node.run_nodetool('flush', args=ks_cf.replace('.', ' '))
+
+    def _verify_using_timestamp_deletions(self, ks_cf: str, verification_queries: list[tuple[int, int, int]]):
+        mv_not_configured = False
+        mv_table_name = ".".join([ks_cf.split(sep=".")[0], "view_test"])
+        with self.cluster.cql_connection_patient(self.target_node, connect_timeout=300) as session:
+            # pylint: disable=invalid-name
+            for pk, ck, ts in verification_queries:
+                result = session.execute(SimpleStatement(
+                    f"SELECT pk, ck, writetime(v) FROM {ks_cf} WHERE pk = {pk} AND ck = {ck}")).one()
+                assert not result or result.writetime_v != ts, \
+                    f"USING TIMESTAMP: deletion failed for ({pk}, {ck}), row still exists, timestamp used: {ts}"
+                if not mv_not_configured:
+                    try:
+                        result = session.execute(
+                            SimpleStatement(
+                                f"SELECT pk, ck, writetime(v) FROM {mv_table_name} WHERE pk = {pk} AND ck = {ck}")).one()
+                        assert not result or result.writetime_v != ts, f"USING TIMESTAMP: deletion failed for ({pk}, {ck}), " \
+                            f"row still exists in MV (!!!), timestamp used: {ts}"
+                    except InvalidRequest:
+                        mv_not_configured = True
 
     def delete_half_partition(self, ks_cf):
         self.log.debug('Delete by range - half of partition')
@@ -2160,6 +2217,29 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         for pkey, ckey in partitions_for_delete.items():
             queries.append(f"delete from {ks_cf} where pk = {pkey} and ck > {int(ckey[1] / 2)}")
         self.run_deletions(queries=queries, ks_cf=ks_cf)
+        return partitions_for_delete
+
+    def delete_by_range_using_timestamp(self, ks_cf: str):
+        self.log.debug('Delete by range - using timestamp')
+
+        partitions_for_delete = self.choose_partitions_for_delete(10, ks_cf, with_clustering_key_data=False)
+        if not partitions_for_delete:
+            message = "Unable to find partitions to delete"
+            self.log.error(message)
+            raise PartitionNotFound(message)
+
+        queries = []
+        verification_queries = []
+        partition_percentage = random.randint(25, 75) / 100
+        for pkey, _ in partitions_for_delete.items():
+            self.log.debug("Using USING TIMESTAMP clause in the deletion for this partition: %s", pkey)
+            timestamp, clustering_key = self.get_random_timestamp_from_partition(
+                ks_cf=ks_cf, pkey=pkey, partition_percentage=partition_percentage)
+            queries.append(SimpleStatement(f"delete from {ks_cf} using timestamp {timestamp} where pk = {pkey}"))
+            verification_queries.append([pkey, clustering_key, timestamp])
+
+        self.run_deletions(queries=queries, ks_cf=ks_cf)
+        self._verify_using_timestamp_deletions(ks_cf=ks_cf, verification_queries=verification_queries)
 
         return partitions_for_delete
 
@@ -2241,7 +2321,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self.verify_initial_inputs_for_delete_nemesis()
 
         ks_cf = 'scylla_bench.test'
-        partitions_for_exclude = self.delete_half_partition(ks_cf)
+        if random.random() > 0.5:
+            partitions_for_exclude = self.delete_half_partition(ks_cf)
+        else:
+            partitions_for_exclude = self.delete_by_range_using_timestamp(ks_cf)
 
         self.delete_range_in_few_partitions(ks_cf, partitions_for_exclude)
 
--- a/test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
+++ b/test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
@@ -27,7 +27,7 @@ prepare_verify_cmd: ["scylla-bench -workload=sequential -mode=read -replication-
                     ]
 
 stress_cmd: [
-"scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=300 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -partition-offset=1001 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s",
+             "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=300 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -partition-offset=1001 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s",
              "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=300 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -partition-offset=1301 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s",
              "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=400 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -partition-offset=1601 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s",
              "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=15 -clustering-row-count=100000 -clustering-row-size=uniform:3072..5120 -concurrency=100 -connection-count=100 -rows-per-request=10 -consistency-level=quorum -timeout=90s -iterations 10 -validate-data",
@@ -51,6 +51,8 @@ stress_read_cmd: [
                   "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1000 -clustering-row-count=100000 -clustering-row-size=uniform:3072..5120 -rows-per-request=10 -consistency-level=quorum -timeout=90s -concurrency=100 -connection-count=100 -iterations=0 -duration=6420m",
                   "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1000 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -rows-per-request=10 -consistency-level=quorum -timeout=90s -partition-offset=1001 -concurrency=100 -connection-count=100 -iterations=0 -duration=6420m"
                  ]
+
+post_prepare_cql_cmds: "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where pk is not null and ck is not null PRIMARY KEY(pk, ck) with comment = 'TEST VIEW'"
 
 n_db_nodes: 5
 n_loaders: 4
--- a/test-cases/longevity/longevity-large-partition-3h.yaml
+++ b/test-cases/longevity/longevity-large-partition-3h.yaml
@@ -25,6 +25,7 @@ stress_cmd: ["scylla-bench -workload=sequential -mode=write -replication-factor=
              "scylla-bench -workload=sequential -mode=read  -replication-factor=3 -partition-count=25 -clustering-row-count=5555 -partition-offset=76   -clustering-row-size=uniform:1024..2048   -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 0 -duration=170m -validate-data"
             ]
 
+post_prepare_cql_cmds: "CREATE INDEX si_text ON scylla_bench.test(pk) with comment = 'TEST INDEX'"
 
 n_db_nodes: 5
 n_loaders: 4


sdcm/nemesis.py
```diff
@@ -2141,6 +2141,26 @@ def delete_half_partition(self, ks_cf):
...
+        with self.cluster.cql_connection_patient() as session:
+            full_partition = session.execute(f"select pk, ck from {ks_cf} where pk = {pk_to_use}").all()
+            delete_mark = int(len(full_partition) * 0.25)
+            timestamp = session.execute(
+                f"select pk, ck from {ks_cf} where pk = {pk_to_use} and ck = {delete_mark}").one().writetime_y
+            if not timestamp:
+                self.log.error(f"Unable to get writetime for row {pk_to_use},{delete_mark}")
+                return
+
+            session.execute(f"delete from {ks_cf} where pk = {pk_to_use} using timestamp {timestamp}")
```
soyacz, line 2162: missing verification - e.g. count rows before and after deletion - verify how many deleted.

sdcm/nemesis.py
```diff
@@ -2141,6 +2141,26 @@ def delete_half_partition(self, ks_cf):
 
         return partitions_for_delete
 
+    def delete_range_with_timestamp(self, ks_cf):
+        self.log.debug('Delete single partition range with timestamp')
+        partition_to_delete = self.choose_partitions_for_delete(partitions_amount=1, ks_cf=ks_cf)
+        if not partition_to_delete:
+            self.log.error('Not found partitions for delete')
```
soyacz, line 2148: shouldn't we raise skip nemesis exception to mark it's skipped?

sdcm/nemesis.py
```diff
@@ -2223,6 +2243,16 @@ def disrupt_delete_by_rows_range(self):
 
         self.delete_range_in_few_partitions(ks_cf, partitions_for_exclude)
 
+    def disrupt_delete_with_timestamp(self):
+        """
+        Delete a large amount of rows from a single partition using WITH TIMESTAMP clause
+        """
+        self.verify_initial_inputs_for_delete_nemesis()
+
+        ks_cf = 'scylla_bench.test'
```
soyacz, line 2252: so, it works only when scylla-bench is used, add this info to docstring, also make nemesis skip if no such ks_cf.

soyacz (changes requested):
some comments and you need to update nemesis.yml and nemesis_classes.yml by running unit tests `test_list_all_available_nemesis`
also run `sct.py create_nemesis_pipelines` to update pipeline (and adapt config to have scylla_bench.test ks_cf)

sdcm/nemesis.py
vponomaryov: Yes, we should. For now, everyone will look at the nemesis results and see there false positive `passed` instead of proper `skipped`.

vponomaryov (changes requested):

sdcm/nemesis.py
vponomaryov: I don't see technical dependency on the scylla-bench here.

sdcm/nemesis.py
```diff
@@ -4567,6 +4597,15 @@ def disrupt(self):
         self.disrupt_delete_10_full_partitions()
 
 
+class DeleteLargePartitionWithTimestampMonkey(Nemesis):
```
vponomaryov, line 4600: The `large` part is not guaranteed here.

sdcm/nemesis.py
k0machi: Ah, I was looking at how the other scylla-bench delete nemesis were done, will raise a skip there instead

sdcm/nemesis.py
k0machi: Good point, will add a check to see that we have at least 50 rows maybe?

fruch: I think the LargePartition in the name here only indicate the test case this nemesis only relevant to specific test case
i.e. this whole Nemesis is tied up to a very specific  test configuration

we don't have a clear way to mark those kind of nemesis, but I would recommend checking the test configuration for the expect configuration, i.e. if scylla-bench is used, and it's parameters match the expectation (or at least the config file name that is used)

as for the size, I think `space_node_threshold` to enough size should be o.k. to verify enough data is written.

fruch:
@k0machi 
- please provide some reference to run with this nemesis
- run the unittest that should update the nemesis list, and generate the pipeline for this new nemesis

sdcm/nemesis.py
```diff
@@ -2141,6 +2141,26 @@ def delete_half_partition(self, ks_cf):
...
+        pk_to_use = partition_to_delete.keys()[0]
+
+        with self.cluster.cql_connection_patient() as session:
+            full_partition = session.execute(f"select pk, ck from {ks_cf} where pk = {pk_to_use}").all()
+            delete_mark = int(len(full_partition) * 0.25)
+            timestamp = session.execute(
+                f"select pk, ck from {ks_cf} where pk = {pk_to_use} and ck = {delete_mark}").one().writetime_y
+            if not timestamp:
+                self.log.error(f"Unable to get writetime for row {pk_to_use},{delete_mark}")
+                return
```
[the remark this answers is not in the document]
roydahan: I agree with @fruch better to fail and handle it if we reach such a state.

sdcm/nemesis.py
```diff
@@ -2223,6 +2243,16 @@ def disrupt_delete_by_rows_range(self):
 
         self.delete_range_in_few_partitions(ks_cf, partitions_for_exclude)
 
+    def disrupt_delete_with_timestamp(self):
```
roydahan, line 2246: I don't think this case is so unique to deserve a special nemesis of its own.
I commented that on the task as well.
We already have other "delete nemesis", you can just add the timestamp there as a random choice.

test-cases/longevity/longevity-large-partition-3h.yaml
```diff
@@ -25,6 +25,7 @@ stress_cmd: ["scylla-bench -workload=sequential -mode=write -replication-factor=
              "scylla-bench -workload=sequential -mode=read  -replication-factor=3 -partition-count=25 -clustering-row-count=5555 -partition-offset=76   -clustering-row-size=uniform:1024..2048   -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 0 -duration=170m -validate-data"
             ]
 
+post_prepare_cql_cmds: "CREATE INDEX si_text ON scylla_bench.test(pk) with comment = 'TEST INDEX'"
```
roydahan, line 28: Let's add it in a separate commit and add it also to other large partition test cases.
not all of them.
One with MV not only INDEX.

roydahan (dismissed):

test-cases/longevity/longevity-large-partition-3h.yaml
k0machi: The other large partition test has an MV already, hence why I only added an index to this one. I'll split the commit.

k0machi:
@vponomaryov @roydahan Reworked the PR.

Valerii, for you specifically - I've added a materialized view and an index to two large partitions tests, be aware of that when reworking the stress commands in those longevities.

sdcm/nemesis.py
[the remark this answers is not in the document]
k0machi: Done

sdcm/nemesis.py
k0machi: Not needed anymore.

sdcm/nemesis.py
k0machi: Done, using "delete_half_partition" as base.

sdcm/nemesis.py
```diff
@@ -2141,6 +2142,20 @@ def choose_partitions_for_delete(self, partitions_amount, ks_cf, with_clustering
         self.log.debug(f'Partitions for delete: {partitions_for_delete}')
         return partitions_for_delete
 
+    def get_random_timestamp_from_partition(self, ks_cf, pkey, partition_percentage=0.25) -> int:
+        with self.cluster.cql_connection_patient() as session:
+            full_partition = list(session.execute(f"select pk, ck from {ks_cf} where pk = {pkey}").all())
```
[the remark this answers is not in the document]
k0machi: I believe I need both since I do count them, but then also I need a specific `ck` value out of that result set to then use to get `writetime_y`

soyacz (commented):
Please provide some link to test run

sdcm/nemesis.py
```diff
@@ -2141,6 +2142,20 @@ def choose_partitions_for_delete(self, partitions_amount, ks_cf, with_clustering
...
+            idx_to_use = floor(len(full_partition) * partition_percentage)
+            delete_mark = full_partition[idx_to_use]
+            timestamp = session.execute(
+                f"select pk, ck from {ks_cf} where pk = {pkey} and ck = {delete_mark}").one().writetime_y
```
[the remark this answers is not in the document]
k0machi: pk is already known, we are working within the same partition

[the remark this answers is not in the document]
soyacz: i mean `delete_mark` is a result of a query that fetches pk and ck, so I wonder how it is you don't select ck value from it?

sdcm/nemesis.py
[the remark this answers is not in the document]
k0machi: Revised, but I believe that is still kinda bandaid - e.g. if partition percentage is 0.5, then with supposition that pk+ck is 10KB, and there are 200000 rows, we will still take 1GB of memory. The keys are bigint I think, so 8 bytes (I believe) per column, so it shouldn't be _that_ bad if we retrieve that many rows now.

k0machi:
[Job](https://argus.scylladb.com/workspace?state=WyIxODNhN2FiZi1mMDEyLTQ1YjctYTBhOC04OTNlMDVkNDE5ZWUiXQ) is here, testing changes.

sdcm/nemesis.py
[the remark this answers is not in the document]
soyacz: > [Job](https://argus.scylladb.com/workspace?state=WyIxODNhN2FiZi1mMDEyLTQ1YjctYTBhOC04OTNlMDVkNDE5ZWUiXQ) is here, testing changes.

Failed:
`f"select count(ck) from {ks_cf} where pk = {pkey}").one().system_count_start_time
AttributeError: 'Row' object has no attribute 'system_count_start_time'
`

soyacz:
> [Job](https://argus.scylladb.com/workspace?state=WyIxODNhN2FiZi1mMDEyLTQ1YjctYTBhOC04OTNlMDVkNDE5ZWUiXQ) is here, testing changes.


Failed:
```
f"select count(ck) from {ks_cf} where pk = {pkey}").one().system_count_start_time AttributeError: 'Row' object has no attribute 'system_count_start_time'
```

k0machi:
Latest run `#11` is now passing, nemesis doesn't impact itself and works.

@soyacz @temichus

test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
```diff
@@ -52,6 +52,8 @@ stress_read_cmd: [
                   "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1000 -clustering-row-count=200000 -clustering-row-size=uniform:10..10240 -rows-per-request=10 -consistency-level=quorum -timeout=90s -partition-offset=1001 -concurrency=100 -connection-count=100 -iterations=0 -duration=6420m"
                  ]
 
+post_prepare_cql_cmds: "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where pk is not null and ck is not null PRIMARY KEY(pk, ck) with comment = 'TEST VIEW'"
```
roydahan, line 55: There are other "large partitions" tests, don't the need it too?

Also, did you test that it doesn't affect the load on the cluster too much?

roydahan:
Let's not ask too many reviewers and then ignore our request.
Leaving @soyacz and @vponomaryov who already reviewed it once + @juliayakovlev who introduced the original nemesis.

sdcm/nemesis.py
```diff
@@ -2158,8 +2181,16 @@ def delete_half_partition(self, ks_cf):
...
         for pkey, ckey in partitions_for_delete.items():
-            queries.append(f"delete from {ks_cf} where pk = {pkey} and ck > {int(ckey[1] / 2)}")
+            if random.random() < 0.5:
+                queries.append(f"delete from {ks_cf} where pk = {pkey} and ck > {int(ckey[1] / 2)}")
+            else:
+                self.log.debug("Using USING TIMESTAMP clause in the deletion for this partition: %s", pkey)
+                timestamp = self.get_random_timestamp_from_partition(
+                    ks_cf=ks_cf, pkey=pkey, partition_percentage=partition_percentage)
+                queries.append(f"delete from {ks_cf} using timestamp {timestamp} where pk = {pkey}")
+
```
soyacz, line 2193: this is inside `delete_half_partition` function which by definition deletes half partition... which now is not true.
Let's move it somewhere else and do the decision which one to ouse earlier so it's clearer what was the flow of the test.

soyacz (changes requested):

test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
k0machi: The 8h large-partition longevity already has a materialized view in it, and I've added an Index to the 3h one, the load wasn't affected that much during the 8h one, but I will start a 4days one today to check. There are also 2days and 4days longevities, with 2days one I think seldom being run, I can add a materialized view to the 4days one and run it in parallel with 200k_pks one.

sdcm/nemesis.py
k0machi: I've moved the code (along with other required bits) into a separate `delete_by_range_using_timestamp` method and now the decision which path to take happens in the disruption func `disrupt_delete_by_rows_range`.

test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
vponomaryov: @k0machi have you made sure that in the test runs where `the load wasn't affected that much during the 8h one` the https://github.com/scylladb/scylla-cluster-tests/issues/6098 was not caught?
If it was caught then the load becomes very-simple having 100% cache coverage.
So, please, look at the cache dashboards and make sure it was `true positive` and not `false positive`.

sdcm/nemesis.py
```diff
@@ -2141,6 +2142,28 @@ def choose_partitions_for_delete(self, partitions_amount, ks_cf, with_clustering
...
+                fetch_limit
+            )
+            partition = session.execute(f"select pk, ck from {ks_cf} where pk = {pkey} limit {fetch_limit}").all()
+            delete_mark = partition[-1].ck
+            timestamp = session.execute(
+                f"select writetime(v) from {ks_cf} where pk = {pkey} and ck = {delete_mark}").one().writetime_v
+            if not timestamp:
+                message = f"Unable to get writetime for row (pk = {pkey}, ck = {delete_mark})"
+                self.log.error(message)
+                raise Exception(message)
```
vponomaryov, line 2163: Maybe better to create specific exception class and use here?
Not sure it is good idea to use the `Exception` directly.

k0machi: Will do

sdcm/nemesis.py
```diff
@@ -2161,6 +2188,25 @@ def delete_half_partition(self, ks_cf):
...
+            self.log.error('Not found partitions for delete')
+            return partitions_for_delete
+
+        queries = []
+        partition_percentage = min(max(random.random(), 0.25), 0.75)
+        for pkey, _ in partitions_for_delete.items():
+            self.log.debug("Using USING TIMESTAMP clause in the deletion for this partition: %s", pkey)
+            timestamp = self.get_random_timestamp_from_partition(
+                ks_cf=ks_cf, pkey=pkey, partition_percentage=partition_percentage)
+            queries.append(f"delete from {ks_cf} using timestamp {timestamp} where pk = {pkey}")
```
soyacz, line 2207: what is expected behavior of such deletion? How it is verified is behaving correctly?

juliayakovlev: yes, this part is missed here

k0machi: Done, I've added some simple checks to see that at least the starting row was affected by deletion (in both mv and base table)

sdcm/nemesis.py
```diff
@@ -2161,6 +2188,43 @@ def delete_half_partition(self, ks_cf):
...
+        for pkey, _ in partitions_for_delete.items():
+            self.log.debug("Using USING TIMESTAMP clause in the deletion for this partition: %s", pkey)
+            timestamp, clustering_key = self.get_random_timestamp_from_partition(
+                ks_cf=ks_cf, pkey=pkey, partition_percentage=partition_percentage)
+            queries.append(f"delete from {ks_cf} using timestamp {timestamp} where pk = {pkey}")
+            verification_queries.append([pkey, clustering_key, timestamp])
+
+        self.run_deletions(queries=queries, ks_cf=ks_cf)
+
+        mv_not_configured = False
```
soyacz, line 2213: move it to some private method so it's easier to follow test flow.

soyacz (changes requested):
one more fix needed and good to go.

soyacz (changes requested):
one more fix needed and good to go.

soyacz (changes requested):
one more fix needed and good to go.

sdcm/nemesis.py
k0machi: Done

soyacz (approved):
LGTM

soyacz:
@k0machi please rebase to fix unit tests errors
@vponomaryov please review

k0machi:
I'm currently waiting for 4days longevity to finish again after rebasing, to see if load increases from that materialized view addition, should be done soon.

k0machi:
@vponomaryov 

200pks-4days with the MV:

image

Without:

image

sdcm/nemesis.py
```diff
@@ -2161,6 +2209,29 @@ def delete_half_partition(self, ks_cf):
...
+        return partitions_for_delete
+
+    def delete_by_range_using_timestamp(self, ks_cf: str):
+        self.log.debug('Delete by range - using timestamp')
+
+        partitions_for_delete = self.choose_partitions_for_delete(10, ks_cf, with_clustering_key_data=False)
+        if not partitions_for_delete:
+            message = "Unable to find partitions to delete"
+            self.log.error(message)
+            raise PartitionNotFound(message)
```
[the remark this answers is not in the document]
k0machi: I've explained in the other comment why I decided to add an exception here instead - this is not a directly called disruption, it's a function called by `disrupt_delete_rows_by_range` (one of the three potential ones), the disruption already checks skip conditions there, so not finding partitions here is an error (either with the schema or test) that should be raised.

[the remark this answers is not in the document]
vponomaryov: Ok, if you think it must exactly `error` then let it be.

vponomaryov (approved):
LGTM

soyacz (commented):
@k0machi still some unit-tests failure, is branch rebased?

k0machi:
> @k0machi still some unit-tests failure, is branch rebased?

rebased it again for a good measure

roydahan:
In general LGTM, but I want @soyacz and @yarongilor who were recently dealing wit large partitions, deletions and tombstones to approve.

sdcm/nemesis.py
```diff
@@ -2141,6 +2150,28 @@ def choose_partitions_for_delete(self, partitions_amount, ks_cf, with_clustering
...
+            partition_size = session.execute(
+                f"select count(ck) from {ks_cf} where pk = {pkey}").one().system_count_ck
+            fetch_limit = max(math.ceil(partition_size * partition_percentage), 11)
+            self.log.debug(
+                "[%s_using_timestamp] Partition size: %s, fetching up to %s",
+                self.get_disrupt_name(),
+                partition_size,
+                fetch_limit
+            )
+            partition = session.execute(f"select pk, ck from {ks_cf} where pk = {pkey} limit {fetch_limit}").all()
```
yarongilor, line 2164: unpaged query

sdcm/nemesis.py
```diff
@@ -2149,6 +2180,23 @@ def run_deletions(self, queries, ks_cf):
...
 
+    def _verify_using_timestamp_deletions(self, ks_cf: str, verification_queries: list[tuple[int, int, int]]):
+        mv_not_configured = False
+        mv_table_name = ".".join([ks_cf.split(sep=".")[0], "view_test"])
+        with self.cluster.cql_connection_patient(self.target_node, connect_timeout=300) as session:
+            # pylint: disable=invalid-name
+            for pk, ck, ts in verification_queries:
+                result = session.execute(f"SELECT pk, ck FROM {ks_cf} WHERE pk = {pk} AND ck = {ck}").one()
+                assert not result, f"USING TIMESTAMP: deletion failed for ({pk}, {ck}), row still exists, timestamp used: {ts}"
+                if not mv_not_configured:
```
yarongilor, line 2191: isn't it always True?

test-cases/longevity/longevity-large-partition-3h.yaml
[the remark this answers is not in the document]
k0machi: The idea with index is to just introduce additional load on table manipulation, would there be an issue with just using partition key as index? It of course results in 1:1 mapping, but would that be a problem?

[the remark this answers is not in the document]
k0machi: As for 4days longevity - will run the nemesis with it today

sdcm/nemesis.py
```diff
@@ -2149,6 +2180,23 @@ def run_deletions(self, queries, ks_cf):
...
         self.target_node.run_nodetool('flush', args=ks_cf.replace('.', ' '))
 
+    def _verify_using_timestamp_deletions(self, ks_cf: str, verification_queries: list[tuple[int, int, int]]):
+        mv_not_configured = False
+        mv_table_name = ".".join([ks_cf.split(sep=".")[0], "view_test"])
+        with self.cluster.cql_connection_patient(self.target_node, connect_timeout=300) as session:
+            # pylint: disable=invalid-name
+            for pk, ck, ts in verification_queries:
+                result = session.execute(f"SELECT pk, ck FROM {ks_cf} WHERE pk = {pk} AND ck = {ck}").one()
+                assert not result, f"USING TIMESTAMP: deletion failed for ({pk}, {ck}), row still exists, timestamp used: {ts}"
```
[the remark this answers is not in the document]
k0machi: Added a writetime check

sdcm/nemesis.py
```diff
@@ -2149,6 +2180,23 @@ def run_deletions(self, queries, ks_cf):
...
+                if not mv_not_configured:
+                    try:
+                        result = session.execute(
+                            f"SELECT pk, ck FROM {mv_table_name} WHERE pk = {pk} AND ck = {ck}").one()
+                        assert not result, f"USING TIMESTAMP: deletion failed for ({pk}, {ck}), " \
```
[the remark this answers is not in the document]
k0machi: Ditto

sdcm/nemesis.py
```diff
@@ -2161,6 +2209,29 @@ def delete_half_partition(self, ks_cf):
...
+
+        partitions_for_delete = self.choose_partitions_for_delete(10, ks_cf, with_clustering_key_data=False)
+        if not partitions_for_delete:
+            message = "Unable to find partitions to delete"
+            self.log.error(message)
+            raise PartitionNotFound(message)
+
+        queries = []
+        verification_queries = []
+        partition_percentage = min(max(random.random(), 0.25), 0.75)
```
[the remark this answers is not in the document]
k0machi: Done

sdcm/nemesis.py
k0machi: Done

sdcm/nemesis.py
```diff
@@ -2141,6 +2150,28 @@ def choose_partitions_for_delete(self, partitions_amount, ks_cf, with_clustering
         self.log.debug(f'Partitions for delete: {partitions_for_delete}')
         return partitions_for_delete
 
+    def get_random_timestamp_from_partition(self, ks_cf, pkey, partition_percentage=0.25) -> tuple[int, int]:
+        with self.cluster.cql_connection_patient(node=self.target_node) as session:
+            partition_size = session.execute(
+                f"select count(ck) from {ks_cf} where pk = {pkey}").one().system_count_ck
```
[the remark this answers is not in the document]
k0machi: Done

sdcm/nemesis.py
```diff
@@ -2141,6 +2150,28 @@ def choose_partitions_for_delete(self, partitions_amount, ks_cf, with_clustering
         self.log.debug(f'Partitions for delete: {partitions_for_delete}')
         return partitions_for_delete
 
+    def get_random_timestamp_from_partition(self, ks_cf, pkey, partition_percentage=0.25) -> tuple[int, int]:
+        with self.cluster.cql_connection_patient(node=self.target_node) as session:
```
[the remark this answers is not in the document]
k0machi: Done

sdcm/nemesis.py
```diff
@@ -2148,6 +2185,26 @@ def run_deletions(self, queries, ks_cf):
...
 
+    def _verify_using_timestamp_deletions(self, ks_cf: str, verification_queries: list[tuple[int, int, int]]):
+        mv_not_configured = False
+        mv_table_name = ".".join([ks_cf.split(sep=".")[0], "view_test"])
+        with self.cluster.cql_connection_patient(self.target_node, connect_timeout=300) as session:
+            # pylint: disable=invalid-name
+            for pk, ck, ts in verification_queries:
+                result = session.execute(SimpleStatement(
+                    f"SELECT pk, ck, writetime(v) FROM {ks_cf} WHERE pk = {pk} AND ck = {ck}")).one()
+                assert not result or result.writetime_v != ts, \
```
soyacz, line 2196: shouldn't we use here greater comparison and verify all the rows instead of `one()`?: `result.writetime_v > ts`

soyacz (approved):

roydahan:
@k0machi please run 2 different jobs that use this nemesis only in staging and run for at least several hours.
E.g.
longevity-large-partitions-8h
longevity-assimetric-12h.

BTW, since these jobs are short time jobs, we probably need to set the GC_Grace_Period to something short.
@yarongilor / @soyacz WDYT? 

(Alexey, do not run till we have an answer for that)

soyacz:
> @k0machi please run 2 different jobs that use this nemesis only in staging and run for at least several hours. E.g. longevity-large-partitions-8h longevity-assimetric-12h.
> 
> BTW, since these jobs are short time jobs, we probably need to set the GC_Grace_Period to something short. @yarongilor / @soyacz WDYT?
> 
> (Alexey, do not run till we have an answer for that)

I don't see the reason for that - test is not verifying tombstones, and only way we currently verify it is by tombstone_gc_verification thread. This could lead to some false negatives (when table repair history shows table was repaired at t=1, and test did delete with timestamp t=0, then verification thread could tell that found 'tombstones in a post-repair sstable')

unless I'm missing something. @yarongilor ?

yarongilor:
> > @k0machi please run 2 different jobs that use this nemesis only in staging and run for at least several hours. E.g. longevity-large-partitions-8h longevity-assimetric-12h.
> > BTW, since these jobs are short time jobs, we probably need to set the GC_Grace_Period to something short. @yarongilor / @soyacz WDYT?
> > (Alexey, do not run till we have an answer for that)
> 
> I don't see the reason for that - test is not verifying tombstones, and only way we currently verify it is by tombstone_gc_verification thread. This could lead to some false negatives (when table repair history shows table was repaired at t=1, and test did delete with timestamp t=0, then verification thread could tell that found 'tombstones in a post-repair sstable')
> 
> unless I'm missing something. @yarongilor ?

gc_grace_seconds may cause an issue where it's duration expires before cluster runs (and complete) a repair.
So test may have a problem unless it makes sure a repair always run in an interval shorter than gc_grace_seconds. This might be possible to configure via scylla_manager.

roydahan:
if we set it to mode: repair, than we don't need to schedule anything in scylla-manager.

yarongilor:
> if we set it to mode: repair, than we don't need to schedule anything in scylla-manager.

right, `gc_grace_seconds` is not relevant to mode `repair`. But we have to avoid running the nemesis of `ToggleGcModeMonkey`.

roydahan:
Ok, let's take it off this task.
@yarongilor I think you have a task to align all these large partitions and validations, we can take it there.

Issue #5912, opened 2023-03-15
raft disabled jobs still use raft


soyacz:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Despite adding sct configuration that disables raft by adding config file: `configurations/raft/consistent_cluster_management_off.yaml`
Raft is not disabled. Downloading `scylla.yaml` file still shows `true` value for `consistent_cluster_management`. 

## Impact

We don't test what we wanted.

## Installation details

Kernel Version: 5.15.0-1031-aws
Scylla version (or git commit hash): `2023.1.0~rc2-20230302.726f8a090337` with build-id `efcc950cda22875aa488c5295c6cc68f35d9cc6f`

Cluster size: 4 nodes (i3en.3xlarge)

Scylla Nodes used in this run:
- longevity-large-collections-12h-202-db-node-51dd5878-5 (54.247.7.226 | 10.4.2.162) (shards: 10)
- longevity-large-collections-12h-202-db-node-51dd5878-4 (52.213.255.32 | 10.4.0.199) (shards: 10)
- longevity-large-collections-12h-202-db-node-51dd5878-3 (18.203.172.45 | 10.4.2.74) (shards: 10)
- longevity-large-collections-12h-202-db-node-51dd5878-2 (3.250.105.3 | 10.4.1.77) (shards: 10)
- longevity-large-collections-12h-202-db-node-51dd5878-1 (54.171.138.230 | 10.4.3.121) (shards: 10)

OS / Image: `ami-0336a5a646b08dc25` (aws: eu-west-1)

Test: `longevity-large-collections-12h-test`
Test id: `51dd5878-8da1-4d87-a40d-8dd6b1f2481e`
Test name: `enterprise-2023.1/disabled_raft/longevity-large-collections-12h-test`
Test config file(s):

- [longevity-large-collections-12h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/4ec1600f015410814cdbe6d94ce29af866d6ccda/test-cases/longevity/longevity-large-collections-12h.yaml)

<details>
<summary>
Logs and commands
</summary>

- Restore Monitor Stack command: `$ hydra investigate show-monitor 51dd5878-8da1-4d87-a40d-8dd6b1f2481e`
[...]

PR #5951, opened 2023-03-27
fix(ScyllaYaml): remove defaults

soyacz:
Having defaults set, it is not possible to put some value into
`scylla.yaml` because default values are skipped. This causes problems
when having default `scylla.yaml` with some value that is not default in
`config.cc` (e.g. `consistent_cluster_management`).

Remove default values from `ScyllaYaml` class, so in case we explicitly
set one, it will be always put into `scylla.yaml`.

--

When publishing `ScyllaYamlUpdateEvent` message is not shown in events
log.

Add message info to `ScyllaYamlUpdateEvent` for better visibility of
scylla.yaml changes in events log.

fixes: #5912


fruch (approved):
LGTM

fruch:
@soyacz 

you need to handle also `unit_tests/test_scylla_yaml_builders.py` seems like they are failing cause of this change as well, even I don't expected them to fail

unit_tests/test_data/test_scylla_yaml_builders/PR-provision-test.result.json
```diff
@@ -14,5 +14,9 @@
     }
   ],
   "experimental": true,
-  "prometheus_address": "0.0.0.0"
+  "prometheus_address": "0.0.0.0",
+  "alternator_enforce_authorization": false,
+  "auto_bootstrap": true,
+  "enable_ipv6_dns_lookup": false,
+  "hinted_handoff_enabled": "enabled"
```
fruch, line 21: @soyacz why do we have those 4 extra ones ? where they are coming from ? the original yaml had them and then were equal to the defaults ?

fruch (dismissed):
LGTM

roydahan:
Let's make sure we give it a quick test.

unit_tests/test_data/test_scylla_yaml_builders/PR-provision-test.result.json
soyacz: I think it comes from `ScyllaYamlClusterAttrBuilder`, previously they were defaults -> skipped from yaml.

soyacz:
I had to do some tuning to `hinted_handoff_enabled` as `nodetool scrub` was complaining about malformed scylla.yaml, apparently it does not accept `enabled` as value (I checked tests, and looks that we use `True` or `False` when we play with this setting in tests anyway).

Now all test passes, except Azure provision test - fails on syntax error for query `TRUNCATE ks_truncate_large_partition.test_table USING TIMEOUT 600s`.

fruch:
@roydahan, @enaydanov, can you review ?

roydahan (approved):

fgelcer:
> TRUNCATE ks_truncate_large_partition.test_table USING TIMEOUT 600

@soyacz , it probably fails because `TRUNCATE` with parameter `USING TIMEOUT` is support from 5.2 (#5965 )
i don't know which version you are using here in the test that failed, nor why you have `USING TIMEOUT` in the `TRUNCATE` command, but it is unrelated to your changes

fgelcer:
the test is running with Scylla `5.0.5`
so where ever this this truncate command is coming from, it needs to be versioned

roydahan:
I backported it and several other commits that it was dependent on it.
Forgot to label.
thx.

PR #5952, opened 2023-03-27
fix(argus.py): Handle TypeError exception in is_uuid as well

temichus:
Handle TypeError exception is_uuid that happened during hydra clean-resources --user <user> command

```
  File "/Users/roy/git/scylla/scylla-cluster-tests/sdcm/utils/argus.py", line 27, in is_uuid
    UUID(uuid)
  File "/Users/roy/.pyenv/versions/3.10.0/lib/python3.10/uuid.py", line 171, in __init__
    raise TypeError('one of the hex, bytes, bytes_le, fields, '
TypeError: one of the hex, bytes, bytes_le, fields, or int arguments must be given

```

### DIFF
--- a/sdcm/utils/argus.py
+++ b/sdcm/utils/argus.py
@@ -26,7 +26,7 @@ def is_uuid(uuid) -> bool:
     try:
         UUID(uuid)
         return True
-    except (ValueError, AttributeError):
+    except (ValueError, AttributeError, TypeError):
         return False
 
 


roydahan (dismissed):
Tested also locally and it works.

roydahan:
@temichus you have few types in the commit message.

fruch (approved):
LGTM

fruch:
I just faced that today, and stashed away the exact fix a minute ago...

fgelcer:
does it need to be backported?

fruch:
> does it need to be backported?

I think it's mainly affect doing some operation without test_id, like clean resources, I don't think too many people are doing that from the release branches directly.

again it probably not gonna do any harm.

fruch:
> > > does it need to be backported?
> > 
> > 
> > I think it's mainly affect doing some operation without test_id, like clean resources, I don't think too many people are doing that from the release branches directly.
> > again it probably not gonna do any harm.
> 
> AFAICT i saw that on Jenkins runs, meaning that it should have `test_id`... so perhaps we have now another issue to find, why it did not have `test_id`, if we do pass it as a parameter to the `hydra` command

Please share a link and put in in an issue, otherwise no one would be looking at it

fgelcer:
> > > > does it need to be backported?
> > > 
> > > 
> > > I think it's mainly affect doing some operation without test_id, like clean resources, I don't think too many people are doing that from the release branches directly.
> > > again it probably not gonna do any harm.
> > 
> > 
> > AFAICT i saw that on Jenkins runs, meaning that it should have `test_id`... so perhaps we have now another issue to find, why it did not have `test_id`, if we do pass it as a parameter to the `hydra` command
> 
> Please share a link and put in in an issue, otherwise no one would be looking at it

don't remember where it was, but if i find it, i will write in here

PR #5964, opened 2023-03-29
fix(gemini-oracle): update oracle to newer versions

fruch:
recently had multiple cases of oracle timing out on gemini runs on master, we should try using newer releases for the oracle.

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -114,7 +114,7 @@ authenticator_password: ''
 # gemini defaults
 n_test_oracle_db_nodes: 1
 gemini_seed: 0
-oracle_scylla_version: '5.0.10'
+oracle_scylla_version: '5.1.7'
 append_scylla_args_oracle: '--enable-cache false'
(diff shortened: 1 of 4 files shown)


aleksbykov (approved):

fgelcer (approved):
LGTM

fruch:
Looks like I've made a mistake and 5.1.7 isn't released yet.

I'll change it to 5.1.6

fruch:
> It's kind of defeating the purpose of using the test oracle while we bump it over and over.

it's problematic to begin with, if we assume the oracle doesn't have bugs...

chasing 5.0 issues seems kind of pointless, since it's about to be deprecated soon.

fruch:
@roydahan so you don't want to update those ? i.e. how do you want to move forward with gemini runs ? is there a point on keep running them ? 

do we have a list of issue that was ever found by gemini runs ?

fruch:
Those were updated in a different PR

PR #5972, opened 2023-03-31
fix(k8s): use exclusive nemesis lock for cluster health validation too

vponomaryov:
We have several exclusive nemesis which disrupt more than 1 Scylla pod from different Scylla clusters. In this case dedicated lock is used to block all other nemesis when the exclusive one gets run. 
But the problem with it is that when we enable cluster health validation it fails with high probability because it is not locked as side nemesis are.
So, fix the exclusive nemesis logic the way that cluster health validation gets lock'ed too.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4209,110 +4209,113 @@ def disrupt_method_wrapper(method, is_exclusive=False):  # pylint: disable=too-m
     def wrapper(*args, **kwargs):  # pylint: disable=too-many-statements
         # pylint: disable=too-many-locals
         # pylint: disable=too-many-branches
-        args[0].set_target_node()
         method_name = method.__name__
-        args[0].current_disruption = "".join(p.capitalize() for p in method_name.replace("disrupt_", "").split("_"))
-        args[0].cluster.check_cluster_health()
-        num_nodes_before = len(args[0].cluster.nodes)
-        start_time = time.time()
-        args[0].log.debug('Start disruption at `%s`', datetime.datetime.fromtimestamp(start_time))
-        class_name = args[0].get_class_name()
-        if class_name.find('Chaos') < 0:
-            args[0].metrics_srv.event_start(class_name)
-        result = None
-        status = True
-        # pylint: disable=protected-access
-        args[0]._set_current_disruption(f"{args[0].current_disruption} {args[0].target_node}")
-        args[0].set_current_running_nemesis(node=args[0].target_node)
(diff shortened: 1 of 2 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-multitenant-eks/55/consoleFull

sdcm/nemesis.py
```diff
@@ -4209,110 +4209,113 @@ def data_validation_prints(args):
...
-        args[0].cluster.check_cluster_health()
-        num_nodes_after = len(args[0].cluster.nodes)
-        if num_nodes_before != num_nodes_after:
-            args[0].log.error('num nodes before %s and nodes after %s does not match' %
-                              (num_nodes_before, num_nodes_after))
-        # TODO: Temporary print. Will be removed later
-        data_validation_prints(args=args)
+        nemesis_run_info_key = f"{id(args[0])}--{method_name}"
+        try:
+            NEMESIS_LOCK.acquire()  # pylint: disable=consider-using-with
```
fruch, line 4215: why always acquire it ?, and not limit the locking into the case we `using is_exclusive` ones ?

fruch: I would expect it to be a mirror image of how it's being released

fruch (approved):
LGTM

sdcm/nemesis.py
fruch: I do see it exactly as it was before... still a bit weird.

sdcm/nemesis.py
```diff
@@ -4209,110 +4209,113 @@ def data_validation_prints(args):
...
+            if num_nodes_before != num_nodes_after:
+                args[0].log.error('num nodes before %s and nodes after %s does not match' %
+                                  (num_nodes_before, num_nodes_after))
+            # TODO: Temporary print. Will be removed later
+            data_validation_prints(args=args)
+        finally:
+            if is_exclusive:
+                # NOTE: sleep the nemesis interval here because the next one is already
+                #       ready to start right after the lock gets released.
+                if args[0].tester.params.get('k8s_tenants_num') > 1:
```
soyacz, line 4310: If `args[0].tester.params.get('k8s_tenants_num')` returns none or not a number it will cause exception and fail releasing `NEMESIS_LOCK`

soyacz (changes requested):
Just one remark, besides LGTM

PR #5981, opened 2023-04-04
fix(CreateIndexNemesis): skip querying by index if null value

soyacz:
When querying base table column to have value to query by index, result may return None value. This makes query to have 'WHERE = None' which is not correct syntax. Scylla does not support putting 'is null' in where clause either.

Fix is about skipping query by index in that case.

commit 2:
When querying by index value may contain ' sign. In that case it must be
escaped to properly format the query.

Fix is by properly escaping ' character from value in where clause.


fruch (approved):
LGTM

PR #5985, opened 2023-04-05
fix(multidc tests): create loader in each region

juliayakovlev:
The default load balancing policy in scylla-tools-java is DCAwareRoundRobinPolicy which prefers nodes in the local datacenter. Unless all local nodes are tried, it will not send queries to other nodes in remote DC.

Prevent issue: https://github.com/scylladb/java-driver/issues/211


fgelcer (approved):
LGTM

fruch:
@juliayakovlev can you share a screen shot from a run with this configuration? to validate it really works as we expect ?

juliayakovlev:
> @juliayakovlev can you share a screen shot from a run with this configuration? to validate it really works as we expect ?

Looks like it does not solve the problem. Updating the issue.
Screenshot from 2023-04-09 18-10-17


But we still need merge it I think. All other multiDC tests run with such configuration

roydahan:
I commented on the issue - https://github.com/scylladb/java-driver/issues/211#issuecomment-1501185634

fruch (approved):
LGTM

PR #5986, opened 2023-04-05
fix(CreateIndexNemesis): prolong errors filter time after drop

soyacz:
After dropping an index it is common that some nodes will report errors in logs (https://github.com/scylladb/scylladb/issues/12977) for longer period than we expected it. These errors are not severe, rather expected.

Prolong event filter for this error to 3 minutes so it is not failing the test.

### DIFF
--- a/sdcm/utils/nemesis_utils/indexes.py
+++ b/sdcm/utils/nemesis_utils/indexes.py
@@ -20,7 +20,7 @@ from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 from sdcm.cluster import BaseNode
 from sdcm.sct_events import Severity
 from sdcm.sct_events.database import DatabaseLogEvent
-from sdcm.sct_events.filters import DbEventsFilter
+from sdcm.sct_events.filters import EventsFilter
 from sdcm.sct_events.system import InfoEvent
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -85,4 +85,4 @@ def drop_index(session, ks, index_name) -> None:
             db_event=DatabaseLogEvent.DATABASE_ERROR,
             line="Error applying view update"):
         session.execute(f'DROP INDEX {ks}.{index_name}')
-        time.sleep(30)  # errors can happen only within several seconds after index drop #12977
+        time.sleep(180)  # errors can happen only within few minutes after index drop scylladb/#12977
```
fruch, line 88: Doesn't the DbFilter has it's own mechanized to keep ignoring error ? Or it's just the ChangeSeverity  has it ?


I don't think we need artificial sleep

soyacz: I think ChangeSeverity  has it, do you think would be better to replace with it?

fruch: Or add the support to this one, but I think it's better option then just slepping

fruch: use `extra_time_to_expiration` it available on `EventsFilter`

fruch:
@soyacz you have a conflict to solve here

sdcm/utils/nemesis_utils/indexes.py
soyacz: thanks, fixed.

vponomaryov (approved):
LGTM

fruch (approved):

fruch:
@soyacz 

Can you backport this one to 2023.1/5.2 ?

PR #5988, opened 2023-04-05
improvement(grafana): use more correct SCT grafana dashboard structure

vponomaryov:
The grafana version that is used by the scylla-operator refuses to import our json file with the SCT dashboard returning following error:

  Dashboard title cannot be empty

So, update the json file doing following steps:
- Remove first-level '"dashboard": {...}' wrapper-parameter.
- Add first-level '"uid": "grafana"' parameter.
- Move all the small first-level parameters to the top of the file for easier/faster reading.

After it we will be able to import that file to the both used grafana versions - SCT's and scylla-operator's ones.

### DIFF
--- a/data_dir/scylla-dash-per-server-nemesis.master.json
+++ b/data_dir/scylla-dash-per-server-nemesis.master.json
@@ -1,5 +1,21 @@
 {
-  "dashboard": {
+    "uid": "grafana",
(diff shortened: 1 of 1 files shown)


vponomaryov:
It is not needed when the dashboard gets uploaded using grafana API. The PR changes work only using UI.

PR #5989, opened 2023-04-06
fix(adaptive_timeout): fix get io-properties.yaml

soyacz:
When k8s scylla node is decommissioned right after spawn, it still have no io_properties.yaml created leading to error on `NodeLoadInfoService` creation. This is critical as it fails the test flow.

Move getting io_properties.yaml to time when it's needed for calculations - which is try/catched and not causing errors.
+ some minor fixes


vponomaryov:
@soyacz is it verified in a CI job?

soyacz:
> @soyacz is it verified in a CI job?

I verified it locally, I added proper test label.
It's important as it fixes k8s functional tests which currently may fail due to this.

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -43,7 +43,7 @@ def convert_to_mb(value) -> float:
                 number *= 1024 * 1024
             case 'PB':
                 number *= 1024 * 1024 * 1024
-        return number
+        return int(number)
```
vponomaryov, line 46: This one looks like a bugfix that must be be in a separate commit.

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -82,7 +81,7 @@ def _get_node_load(self) -> tuple[float, float, float]:
         except Exception as exc:  # pylint: disable=broad-except
             LOGGER.debug("Couldn't get node load from prometheus metrics. Error: %s", exc)
             # fallback to uptime
-            load_1, load_5, load_15 = self.remoter.run('uptime', retry=3).stdout.split("load average: ")[1].split(",")
+            load_1, load_5, load_15 = self.remoter.run('uptime').stdout.split("load average: ")[1].split(",")
```
vponomaryov, line 84: Removal of retries also sounds as a separate fix which requires description.
It is not clear now why it is needed at all. How exactly it makes thing be better?

soyacz:
@vponomaryov I split commits. As to run - I did one and proved it works as expected (https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-5989/2/ job failed but different reason)

vponomaryov (approved):
LGTM

soyacz:
@aleksbykov can you review so I could merge it?

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -57,19 +57,18 @@ def __init__(self, remoter: RemoteCmdRunner, name: str, scylla_version: str):
...
         self._name = name
         self._scylla_version = scylla_version
-        self._io_properties = yaml.safe_load(self._get_io_properites_yaml())
 
-    @cache
-    def _get_io_properites_yaml(self):
-        return self.remoter.run('cat /etc/scylla.d/io_properties.yaml', retry=3).stdout
+    @cached_property
+    def _io_properties(self):
+        return yaml.safe_load(self.remoter.run('cat /etc/scylla.d/io_properties.yaml', verbose=False).stdout)
```
aleksbykov, line 63: Do you expect, that if command failed by any reason, the test will also failed? don't wont to use ignore_status=True?

aleksbykov (approved):
LGTM , small comment

sdcm/utils/adaptive_timeouts/load_info_store.py
soyacz: I catch it in some other place - this is the reason of this PR.

If run with ignore_status then it would cache wrong value and effectively deactivate feature.

PR #5991, opened 2023-04-06
improvement(grafana): properly calculate nodes number in K8S's grafana

vponomaryov:
Prometheus deployed by Scylla operator uses 2 scrapes on each Scylla pod.
So, filter out second one in the node count dashboards to avoid doubled values.

### DIFF
--- a/data_dir/scylla-dash-per-server-nemesis.master.json
+++ b/data_dir/scylla-dash-per-server-nemesis.master.json
@@ -229,7 +229,7 @@
         "pluginVersion": "8.5.2",
         "targets": [
           {
-            "expr": "count(up{job=\"scylla\", cluster=~\"$cluster|$^\"})",
+            "expr": "count(up{job=\"scylla\", cluster=~\"$cluster|$^\", endpoint!=\"node-exporter\"})",
             "intervalFactor": 1,
             "legendFormat": "Total Nodes",
             "refId": "A",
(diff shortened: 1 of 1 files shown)


vponomaryov:
Scylla-operator's prometheus:
Screenshot from 2023-04-05 17-02-33

And 3-node Scylla cluster looks like following:
Screenshot from 2023-04-05 17-04-10

With this fix it filters out 1 set out of 2 calculating total num of nodes and down ones.
Running in a common monitoring where we don't such these keys and values everything works as usual, because the new addition means "such pair must be absent". So, it influences only scylla-operator's monitoring case.

soyacz:
> Prometheus deployed by Scylla operator uses 2 scrapes on each Scylla pod.
> So, filter out second one in the node count dashboards to avoid doubled values.

Isn't it issue with scylla-operator and fixed on that side?

vponomaryov:
> > Prometheus deployed by Scylla operator uses 2 scrapes on each Scylla pod.
> > So, filter out second one in the node count dashboards to avoid doubled values.
> 
> Isn't it issue with scylla-operator and fixed on that side?

It is not issue. There are 2 types of metrics sources - `scylla` binary and `node-exporter`. Both run in each scylla pod.

soyacz (approved):
LGTM

vponomaryov:
@fruch ?

fruch (approved):
LGTM

PR #5992, opened 2023-04-06
feature(cassandra-harry): Allow harry to run longer

soyacz:
cassandra-harry creates `validation.log` and `operations.log` which
store all the queries executed. This makes disk quickly fill up so
we can't run it for long.

Create `validation.log` and `operations.log` symbolic links to
`/dev/null` so disk is not filled up and can run longer.
Additionally set logger to INFO level instead of DEBUG.
Also rebuild of docker image caused update of cassandra-harry binaries
with some new features commited in Mar 14, 2023 in harry's repo.
Due to that needed to modify `external.yaml` config file to have random
schema.

Add 24h test with cassandra-harry.

refs: https://github.com/scylladb/qa-tasks/issues/906

### DIFF
--- a/docker/cassandra-harry/Dockerfile
+++ b/docker/cassandra-harry/Dockerfile
@@ -6,9 +6,13 @@ RUN apt -y update && apt -y install make \
 
 FROM openjdk:11 as app
 COPY --from=builder /cassandra-harry /cassandra-harry
+COPY external.yaml /cassandra-harry/conf/external.yaml
+COPY logback-dtest.xml /cassandra-harry/test/conf/logback-dtest.xml
 
 RUN sed -i '2s;^;HARRY_HOME=/cassandra-harry\n;'  /cassandra-harry/scripts/cassandra-harry
 RUN sed -i 's/external-.*-SNAPSHOT.jar/external-*-SNAPSHOT.jar/' /cassandra-harry/scripts/cassandra-harry
 RUN ln -svf /cassandra-harry/scripts/cassandra-harry /usr/bin/cassandra-harry
+RUN ln -svf /dev/null /cassandra-harry/validation.log
+RUN ln -svf /dev/null /cassandra-harry/operation.log
 
 WORKDIR /cassandra-harry
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -210,7 +210,7 @@ stress_image:
   alternator-dns: 'scylladb/hydra-loaders:alternator-dns-0.1'
   cdc-stresser: 'scylladb/hydra-loaders:cdc-stresser-20210630'
   kcl: 'scylladb/hydra-loaders:kcl-jdk8-20210526-ShardSyncStrategyType-PERIODIC'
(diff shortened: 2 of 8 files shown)


soyacz:
first run just to see how it works: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-harry-24h-test/3/ (if can run all nemesis)

soyacz:
> first run just to see how it works: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-harry-24h-test/3/ (if can run all nemesis)

verbosity was too low, c-h ended without showing much clue what happened - recovered to DEBUG.
Also it stopped during nospc nemesis - so it seems we cant run with disruptive. 
Retry without disruptive and better verbosity.
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-harry-24h-test/4/

roydahan:
+1 for your work.
One more thing we need to understand in this task if it makes sense that we can't run this test with disruptive nemesis.
If we maintain quorum during the test, I don't know if there is any reason why not (but I've never drilled down to cassandra-harry code).

soyacz:
> +1 for your work. One more thing we need to understand in this task if it makes sense that we can't run this test with disruptive nemesis. If we maintain quorum during the test, I don't know if there is any reason why not (but I've never drilled down to cassandra-harry code).

I'll try to investigate, harry has plenty of options which still I don't understand.
Also there's not only problem with disruptive errors, recent job without disruptive nemesis failed due some Harry's error:
```
ERROR 09:53:00 Failed due to exception: java.lang.IllegalStateException: RTS is too old to convert to LTS: 1680853881014000
```

soyacz:
I try without nemesis at all to see if c-h can handle such long runs: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-harry-24h-test/5/

docker/cassandra-harry/Dockerfile
```diff
@@ -6,9 +6,13 @@ RUN apt -y update && apt -y install make \
 
 FROM openjdk:11 as app
 COPY --from=builder /cassandra-harry /cassandra-harry
+COPY external.yaml /cassandra-harry/conf/external.yaml
+COPY logback-dtest.xml /cassandra-harry/test/conf/logback-dtest.xml
 
 RUN sed -i '2s;^;HARRY_HOME=/cassandra-harry\n;'  /cassandra-harry/scripts/cassandra-harry
 RUN sed -i 's/external-.*-SNAPSHOT.jar/external-*-SNAPSHOT.jar/' /cassandra-harry/scripts/cassandra-harry
 RUN ln -svf /cassandra-harry/scripts/cassandra-harry /usr/bin/cassandra-harry
+RUN ln -svf /dev/null /cassandra-harry/validation.log
```
fruch, line 15: For investigations, we might want a way to save those logs

fruch:
closing until someone would need to attend to it again

Issue #5996, opened 2023-04-09
MultiDC test: provisioned instances for 2nd and 3th (in the region parameter list) region are ignored and new instances are created in "Run SCT Test"


juliayakovlev:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Test with multi DC cluster. AWS. I did not check on GCE or Azure

Regions: eu-west-1, us-west-2, us-east-1

Instances were provisioned for every region.
But just for first region in the list provisioned instances were used. Region **eu-west-1**:
```
14:09:42  < t:2023-04-09 11:09:42,713 f:cluster_aws.py  l:351  c:sdcm.cluster         p:INFO  > Cluster lwt-longevity-multi-dc-24h-create-l-db-cluster-678b7602 (AMI: ['ami-0336a5a646b08dc25', 'ami-017ff4adbd0789b7d', 'ami-0bfa3b8209fe4bb00'] Type: i3.8xlarge): Found provisioned instances = [ec2.Instance(id='i-0092510240b9f0b51'), ec2.Instance(id='i-0849cac6e917e98c0'), ec2.Instance(id='i-0a110ffa7ef3849cf'), ec2.Instance(id='i-0060e05ba902345a6')]
14:09:43  < t:2023-04-09 11:09:43,365 f:cluster.py      l:499  c:sdcm.cluster_aws     p:INFO  > Node lwt-longevity-multi-dc-24h-create-l-db-node-678b7602-1 [176.34.76.141 | 10.4.3.132] (seed: False): Detected Linux distribution: UBUNTU22
14:10:08  < t:2023-04-09 11:10:08,548 f:cluster.py      l:499  c:sdcm.cluster_aws     p:INFO  > Node lwt-longevity-multi-dc-24h-create-l-db-node-678b7602-2 [54.228.89.246 | 10.4.0.119] (seed: False): Detected Linux distribution: UBUNTU22
14:10:33  < t:2023-04-09 11:10:33,664 f:cluster.py      l:499  c:sdcm.cluster_aws     p:INFO  > Node lwt-longevity-multi-dc-24h-create-l-db-node-678b7602-3 [52.50.88.132 | 10.4.0.83] (seed: False): Detected Linux distribution: UBUNTU22
[...]

PR #6002, opened 2023-04-10
fix(create instances): consider DC when create an in instance

juliayakovlev:
In multi DC test  provisioned instances for not first (in the region parameter list) region are ignored and new instances are created in 'Run SCT Test'. As result in all regions, except first one) duplicated instances are created.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/5996


sdcm/cluster_aws.py
```diff
@@ -340,7 +340,8 @@ def update_bootstrap(ec2_user_data, enable_auto_bootstrap):
         return ec2_user_data
 
     def _create_or_find_instances(self, count, ec2_user_data, dc_idx):
-        if self.nodes:
+        region_names = self.params.get('region_name')
+        if [node for node in self.nodes if node.region == region_names[dc_idx]]:
```
dkropachev, line 344: I even remember this discussion, it was agreed that we are doing it this way, to not complicate the PR.
Your solution would no work, since when SCT is just starting there are no such nodes in the given DC, and you fall into `self._create_instances(count, ec2_user_data, dc_idx)`.

I think code need to address all the cases - i.e. partially provisioned instances ... etc..
In order to do that you need to have something like this:
```
    def _create_or_find_instances(self, count, ec2_user_data, dc_idx):
        already_in_use = [node._instance.instance_id for node in self.nodes if node.dc_idx == dc_idx]
        unused_on_cloud = [i for i in self._get_instances(dc_idx) if i.instance_id in already_in_use]
        if len(unused_on_cloud) <= count:
            self.log.info('Found provisioned instances = %s', unused_on_cloud[:count])
            return unused_on_cloud[:count]

        newly_provisioned = self._create_instances(count - len(unused_on_cloud), ec2_user_data, dc_idx)
        self.log.info('Found provisioned instances = %s, and provisioned additional instances = %s ', unused_on_cloud, newly_provisioned)
        return unused_on_cloud + newly_provisioned
```

juliayakovlev:
Tested:
1. single DC: https://jenkins.scylladb.com/job/scylla-staging/job/yulia/job/staging-10gb-3h/363/
2. multi DC: 
   https://jenkins.scylladb.com/job/scylla-staging/job/yulia/job/longevity-lwt-24h-multidc-test/11/
   https://jenkins.scylladb.com/job/scylla-staging/job/yulia/job/longevity-lwt-24h-multidc-test/12/

Both jobs ran 2 nemeses:
- disrupt_nodetool_decommission
- disrupt_terminate_and_replace_node

sdcm/cluster_aws.py
```diff
@@ -339,25 +341,33 @@ def update_bootstrap(ec2_user_data, enable_auto_bootstrap):
...
+        self.log.info('Found provisioned instances = %s, and provisioned additional instances = %s ', unused_on_cloud,
+                      newly_provisioned)
+        return unused_on_cloud + newly_provisioned
 
     # pylint: disable=too-many-arguments
     def add_nodes(self, count, ec2_user_data='', dc_idx=0, rack=0, enable_auto_bootstrap=False):
         ec2_user_data = self.prepare_user_data(enable_auto_bootstrap=enable_auto_bootstrap)
 
-        instances = self._create_or_find_instances(count=count, ec2_user_data=ec2_user_data, dc_idx=dc_idx, az_idx=rack)
+        instances = self._create_or_find_instances(count=count, ec2_user_data=ec2_user_data, dc_idx=dc_idx,
```
fruch, line 369: this is the only call to `_create_or_find_instances`, so why do we need the `running` argument ?

sdcm/cluster_aws.py
```diff
@@ -339,25 +341,33 @@ def update_bootstrap(ec2_user_data, enable_auto_bootstrap):
                 ec2_user_data += ' --bootstrap false '
         return ec2_user_data
 
-    def _create_or_find_instances(self, count, ec2_user_data, dc_idx, az_idx=0):
-        if self.nodes:
-            return self._create_instances(count, ec2_user_data, dc_idx, az_idx)
+    # pylint: disable=too-many-arguments
+    def _create_or_find_instances(self, count, ec2_user_data, dc_idx, az_idx, running=False):
```
fruch, line 345: `running` argument isn't clear in this context, why would I care about the state, cause of currently decommissioned node ?

sdcm/cluster_aws.py
juliayakovlev: By default `running=False` and all found instances will be chosen, even the instance that was decommissioned and can not be used (actually, creation of the node on this instance fails). So I filter such instances out

fruch: again since it's the only call, why do we need a parameter ? is there a use case we want to get the instances that are decommissioned ?

juliayakovlev: Not for existent case. Just to do that more flexible.

fruch: I would suggest to remove it, i.e. it's more places to reason about it and explain this parameter, and it's not something easy to assume on your own.

soyacz:
For me this commit is trying to solve more issues than referenced one, and I'm not sure these issues are relevant at all.
Also code is hard to understand: e.g.names like `already_in_use` and `unused_on_cloud` looks like synonyms, major logic change introduced issue that 'running' status started to be needed (it was needed only when REUSE_CLUSTER, but this is a fix to some other issue). Commit message is not explaining how it is fixing the issue.

soyacz:
alternative fix in https://github.com/scylladb/scylla-cluster-tests/pull/6042

sdcm/cluster_aws.py
```diff
@@ -339,25 +341,33 @@ def update_bootstrap(ec2_user_data, enable_auto_bootstrap):
...
         return ec2_user_data
 
-    def _create_or_find_instances(self, count, ec2_user_data, dc_idx, az_idx=0):
-        if self.nodes:
-            return self._create_instances(count, ec2_user_data, dc_idx, az_idx)
+    # pylint: disable=too-many-arguments
+    def _create_or_find_instances(self, count, ec2_user_data, dc_idx, az_idx):
         if self.test_config.REUSE_CLUSTER:
-            instances = self._get_instances(dc_idx)
+            instances = self._get_instances(dc_idx, running=True)
```
soyacz, line 347: this line is worth commiting, as currently we could have issue with decommissioned nodes be taken up.

PR #6012, opened 2023-04-14
fix(event_analyzer): call 'kill_test' only once

vponomaryov:
We may get flooded with the `CRITICAL` events and each of them will trigger call of `kill_test` method where we send SIGUSR2 signal to python main thread.
At first, we should call this method only once.
At second, we will fall into endless loop of failures where we catch following old python bug: https://bugs.python.org/issue24283

So, run `kill_test` method only once to avoid endless failure loop.

### DIFF
--- a/sdcm/sct_events/events_analyzer.py
+++ b/sdcm/sct_events/events_analyzer.py
@@ -51,9 +51,18 @@ class EventsAnalyzer(BaseEventsProcess[Tuple[str, Any], None], threading.Thread)
                 except TestFailure:
                     self.kill_test(sys.exc_info())
 
-    def kill_test(self, backtrace_with_reason) -> None:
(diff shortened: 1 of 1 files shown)


vponomaryov:
Fixes the https://github.com/scylladb/scylla-cluster-tests/issues/5626

Without the fix: https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/gke/job/longevity-scylla-operator-3h-gke/24/consoleFull
With the fix: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-gke/23/consoleFull

fruch (approved):
LGTM

soyacz (approved):
LGTM

PR #6027, opened 2023-04-17
improvement(managerPipeline): include the provision type in the trigger

ShlomiBalalis:
When the spot price is too high, we are resolt to use on demand intances. The thing is that every downstream job that is triggered afterwards by the first build will still use spot by default, and thus fail. To save the time of rebuilding all of said jobs I decided to also transfer the provision_type param in the trigger.


fgelcer (approved):
LGTM

soyacz:
how about fallback to on_demand?

fgelcer:
> how about fallback to on_demand?

i think this is not necessary for the manager tests, for now... as we don't run all these jobs on a very frequent fashion other than release process, we should be fine running them with `on_demand`, but we need to pass it downstream to next jobs...
and i expect to "end" our work with the manager soon, so won't bother to invest in it for now

fruch (approved):
LGTM

fruch:
> how about fallback to on_demand?

fallback isn't relevant in this case, we don't have any information to do this decision on this level of triggering
i.e. it's a feature on it's own, not related to manager.

PR #6036, opened 2023-04-18
improvement(cluster): find datacenter according to region

juliayakovlev:
Create mapping `region  - <datacenter name>`.
Datacenter name can be found from 'nodetool status' output. Receive it is possible for DB node only.
We can use this mapping, for example, when need to find in which datacenter a loader is located. It may be helpfull in multi DC tests

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3242,6 +3242,14 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
             grouped_by_region[node.region].append(node)
         return grouped_by_region
 
+    def get_datacenter_name_per_region(self):
+        datacenter_name_per_region = {}
+        for region, nodes in self.nodes_by_region().items():
+            status = nodes[0].get_nodes_status()
+            datacenter_name_per_region[region] = status[nodes[0]]['dc']
+
+        return datacenter_name_per_region
+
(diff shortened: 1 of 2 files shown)


sdcm/cluster.py
```diff
@@ -3242,6 +3242,14 @@ def nodes_by_region(self, nodes=None) -> dict:
             grouped_by_region[node.region].append(node)
         return grouped_by_region
 
+    def datacenter_name_per_region(self):
```
soyacz, line 3245: we don't change datacenters during the test. `@cache`? 
Also if it's method, add `get_` prefix to it, or use it as` @property` or `@cached_property`

soyacz (commented):
Would be nice to add code that uses it so we can see example. 
But I'm ok with only this change, as I also need this change for my task.

sdcm/cluster.py
juliayakovlev: We have nemesis that add new datacenter, for example

juliayakovlev:
> Would be nice to add code that uses it so we can see example. But I'm ok with only this change, as I also need this change for my task.

I have another commit, almost done, just needs this ability. It will use this function

sdcm/cluster.py
juliayakovlev: renamed the function

soyacz (approved):

PR #6037, opened 2023-04-18
feature(k8s-ingress): add logging thread

fruch:
Since on long runs we might have lots of restart of the haproxy ingress, and at the end of the test we only have partial log.

this change is adding a new thread that would read the logs during the test, so we can collect all the logs from a run.

## Testing
- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/functional-eks/8/

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -86,7 +86,7 @@ from sdcm.utils.decorators import log_run_info, retrying
 from sdcm.utils.decorators import timeout as timeout_wrapper
 from sdcm.utils.k8s.chaos_mesh import ChaosMesh
 from sdcm.utils.remote_logger import get_system_logging_thread, CertManagerLogger, ScyllaOperatorLogger, \
-    KubectlClusterEventsLogger, ScyllaManagerLogger, KubernetesWrongSchedulingLogger
+    KubectlClusterEventsLogger, ScyllaManagerLogger, KubernetesWrongSchedulingLogger, HaproxyIngressLogger
 from sdcm.utils.sstable.load_utils import SstableLoadUtils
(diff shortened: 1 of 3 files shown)


vponomaryov:
@fruch why is it `draft` ? It looks ok.

fruch:
> @fruch why is it `draft` ? It looks ok.

I didn't yet fully validated it's working as I expected (and EKS was broken till yesterday, to fully test)

vponomaryov:
> > @fruch why is it `draft` ? It looks ok.
> 
> I didn't yet fully validated it's working as I expected (and EKS was broken till yesterday, to fully test)

I suspect it may suffer from connection hangings because of the restarts...
Probably need to wait for the https://github.com/scylladb/scylla-cluster-tests/pull/5927 finalization and reuse for this logger.

fruch:
> > > @fruch why is it `draft` ? It looks ok.
> > 
> > 
> > I didn't yet fully validated it's working as I expected (and EKS was broken till yesterday, to fully test)
> 
> I suspect it may suffer from connection hangings because of the restarts... Probably need to wait for the #5927 finalization and reuse for this logger.

it's would be better the current situation, I don't have any logic built ontop of it, or code reading those logs.
so this shouldn't be blocked by #5927

fruch:
@vponomaryov @soyacz been able to verify on the last PR CI run that it works as expected

vponomaryov:
> @vponomaryov @soyacz been able to verify on the last PR CI run that it works as expected

From the code point of view I see that it must work, it uses the same approach as some other existing threads.
I was referring to the possible `log hanging` related to the pod restarts.
But, since it is separate thread which is not going to fail a test if logs stuck, can be merged.

vponomaryov (approved):
LGTM

Issue #6041, opened 2023-04-18
Job started with wrong number of loader nodes (7 instead of 4) 


Mark-Gurevich:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: branch-2023.1
* **scylla**: 2023.1.0-rc4

### Logs
* test_id: a90bf6ba-74a0-41e7-a283-27375d73dd9f
* job log: https://cloudius-jenkins-test.s3.amazonaws.com/a90bf6ba-74a0-41e7-a283-27375d73dd9f/20230413_212706/sct-a90bf6ba.log.tar.gz
* loaders logs: https://cloudius-jenkins-test.s3.amazonaws.com/4ef11918-2f24-459d-a8e2-83e2cfaba5f1/20230417_101752/loader-set-4ef11918.tar.gz

### Description
The job started with 7 loader nodes instead of 4 while in the config file we have 4.
We can see that c-s ran on 4 nodes but 7 were started.

https://argus.scylladb.com/test/6865de3b-2605-4449-bc27-55bd19b1c4e2/runs?additionalRuns%5B%5D=8ace7e5f-6988-4f46-895b-6ee229dd35c5

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:** [What you expected to happen]
Job expected to start 4 loader nodes.
**Actual behavior:** [What actually happened]
Job started 7 loader nodes.

soyacz:
error is in _create_or_find_instances logic: previously it was executed just once, after adding multi-az support it is executed more times and it is not detecting provisioned instances, it add new ones.
I try to find solution now.

fruch:
> error is in _create_or_find_instances logic: previously it was executed just once, after adding multi-az support it is executed more times and it is not detecting provisioned instances, it add new ones. I try to find solution now.

it's exactly what Julia is trying to solve

fruch:
> error is in _create_or_find_instances logic: previously it was executed just once, after adding multi-az support it is executed more times and it is not detecting provisioned instances, it add new ones. I try to find solution now.

it's happening all the time, or only on jobs using multiple AZ ?

fruch:
@soyacz please look at https://github.com/scylladb/scylla-cluster-tests/pull/6002, I think it might be the exact issue.
we thought it would be happening only on multiple regions...

[...]

PR #6042, opened 2023-04-18
fix(provision): fix duplicate provisioning after mult-az intro

soyacz:
When running `self.add_nodes` multiple times, the second run does not find provisioned instances because `self.nodes` is not empty anymore. It was not an issue before multi-az support because it was executed just once (except multi-dc, where probably we had this issue before).

When single az is provided, `self.add_nodes` will be called just like it was before. Issue with multidc is not fixed - this will be tackled in separate PR.

fixes: #6041

Second commit:
fix(multi-dc): fix double provision in multi-dc
In multidc scenario `self.add_nodes` is executed once per dc.
`_create_or_find_instances` when executed for second time finds
`self.nodes` is not empty, provisions new instances.

Filter nodes by dc in `_create_or_find_instances` so it tries to detect
instances for each dc before creating ones.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/5996

### DIFF
--- a/sdcm/cluster_aws.py
+++ b/sdcm/cluster_aws.py
@@ -340,7 +340,8 @@ class AWSCluster(cluster.BaseCluster):  # pylint: disable=too-many-instance-attr
         return ec2_user_data
 
     def _create_or_find_instances(self, count, ec2_user_data, dc_idx, az_idx=0):
-        if self.nodes:
+        nodes = [node for node in self.nodes if node.dc_idx == dc_idx]
+        if nodes:
             return self._create_instances(count, ec2_user_data, dc_idx, az_idx)
         if self.test_config.REUSE_CLUSTER:
             instances = self._get_instances(dc_idx)
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3153,7 +3153,7 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
     Cluster of Node objects.
     """
 
-    # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-arguments,too-many-locals,too-many-branches
     def __init__(self, cluster_uuid=None, cluster_prefix='cluster', node_prefix='node', n_nodes=3, params=None,
                  region_names=None, node_type=None, extra_network_interface=False, add_nodes=True):
         self.extra_network_interface = extra_network_interface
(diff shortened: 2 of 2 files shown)


soyacz:
test single-dc: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-mini-test-1h-test/6/

soyacz:
multidc test: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/reproducer-counters-multidc-raft/4/

juliayakovlev:
Please, see this fix:
https://github.com/scylladb/scylla-cluster-tests/pull/6002/files
`_create_or_find_instances` is changing there

soyacz:
I verified logs (and yesterday on AWS itself for multidc) and looks these commits are enough to fix.
@fruch @juliayakovlev

sdcm/cluster_aws.py
```diff
@@ -340,7 +340,8 @@ def update_bootstrap(ec2_user_data, enable_auto_bootstrap):
         return ec2_user_data
 
     def _create_or_find_instances(self, count, ec2_user_data, dc_idx, az_idx=0):
-        if self.nodes:
+        nodes = [node for node in self.nodes if node.dc_idx == dc_idx]
```
juliayakovlev, line 343: This is exactly solution that I suggested to do. @dkropachev suggested more advance solution that can cover more cases

fruch (approved):
LGTM

sdcm/cluster_aws.py
fruch: Since we assume the issue is only with multi-dc cases, I thought we have time to try a better soultion @dkropachev suggest.

but we were missing the root casue that made it start happening, and that was @soyacz AZ related changes.

fruch: now it's become more urgent, since it's effect all cases, not just multi-region ones, and we need quick fix (less complicated, at least for now)

fruch:
> I verified logs (and yesterday on AWS itself for multidc) and looks these commits are enough to fix. @fruch @juliayakovlev

did you tried also any of the multi AZ cases ?

sdcm/cluster_aws.py
soyacz: > Since we assume the issue is only with multi-dc cases, I thought we have time to try a better soultion @dkropachev suggest.
> 
> but we were missing the root casue that made it start happening, and that was @soyacz AZ related changes.

After code analysis I think this was happening all the time for multi dc, AZ support only made it worse.

soyacz:
trying here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-topology-changes-3h-test/4/

fruch:
> trying here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-topology-changes-3h-test/4/

@soyacz once you confirm, lets merge and backport

sdcm/cluster_aws.py
vponomaryov: Why `az_idx` is not part of the condition?

soyacz:
there was an issue with multi-az when having 3 az's 3 db_nodes and 1 loader - loader didn't fit to assertion of `num % len(azs) == 0`.
Fixed: nodes are spread as evenly as possible.

soyacz:
retesting:
single dc/az: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-mini-test-1h-test/7/
multidc: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/reproducer-counters-multidc-raft/5/
multiaz: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-topology-changes-3h-test/5/

sdcm/cluster_aws.py
vponomaryov: @soyacz ^ ?

soyacz: for now when having multi-az we don't follow this logic as provision step is disabled. When adapting provision step to support multi-az then it must be implemented. I already work on it. (currently we don't support filtering by az when getting ec2 instances)

vponomaryov: @soyacz thanks for the explanation, I think, it would be really useful to have it as a comment in the code.

soyacz:
@fruch tests show fix works properly.

fruch (approved):
LGTM

PR #6048, opened 2023-04-18
fix(eks): make node-setup work again

vponomaryov:
For node setup in EKS backend we use 'scylla-machine-image' docker image. It is based on the scylla 4.1
And recently it's repos were dropped completely.
So, it caused any package installation to be failed. In our case we use 'mdadm' package which is absent in the image and always gets installed when a pod with it gets started.

So, disable scylla package repos allowing packages to be installed from the main centos repos.

Signed-off-by: Maciej Zimnoch <maciej@scylladb.com>
Signed-off-by: Valerii Ponomarov <valerii.ponomarov@scylladb.com>

Fixes: #6022


sdcm/k8s_configs/eks/scylla-node-prepare.yaml
```diff
@@ -186,6 +186,14 @@ spec:
...
           - name: SCYLLAD_CONF_MOUNT
             value: /mnt/scylla.d/
+        lifecycle:
+          postStart:
+            exec:
+              command:
+                - "/bin/bash"
+                - "-c"
+                - "--"
+                - "yum-config-manager --disable scylla --disable scylla-generic --disable scylladb-scylla-3rdparty"
```
fruch, line 196: close to may suggestion building a new image that only removed the repo files.

but this one is a nice trick (should write it down in the back of my mind :))

fruch (approved):
LGTM

soyacz (dismissed):

zimnx:
It's a race, your postStart handler isn't guaranteed to be called before container entrypoint, where first action in entrypoint script is updating yum repositories which is failing. Please wait for the new image.

vponomaryov:
> It's a race, your postStart handler isn't guaranteed to be called before container entrypoint, where first action in entrypoint script is updating yum repositories which is failing. Please wait for the new image.
> 
> BTW, this commit isn't signed by me.

Technically there is a `race`, but the entrypoint is python script and I doubt we will get it run and reach the package setup point faster than the postStart hook does it's fast binary job.

About the `sign`, it is reference to your contribution. If you insist on removing it, no problem, will remove it.

soyacz (dismissed):

sdcm/k8s_configs/eks/scylla-node-prepare.yaml
```diff
@@ -186,6 +186,14 @@ spec:
             value: /mnt/hostfs/mnt/raid-disks/disk0
           - name: SCYLLAD_CONF_MOUNT
             value: /mnt/scylla.d/
+        lifecycle:
+          postStart:
```
tnozicka, line 190: isn't this racy?
`However, there is no guarantee that the hook will execute before the container ENTRYPOINT` https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/

fruch: @tnozicka would we have a replacement image anytime soon ? without a fix, we can't run EKS/GKE at all..

tnozicka: why don't use use entrypoint==bash and do it there before running the script?

vponomaryov: It is one more way. I don't mind to redefine the entrypoint.

vponomaryov:
The latest PR state with the `guaranteed` approach was verified here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-provision-eks-cluster/130/consoleFull

fruch (approved):
LGTM

PR #6034, opened 2023-04-18
fix(nonroot-artifcat-test): fix the location of cqlsh

fruch:
Since in nonroot install cqlsh isn't on the path, we need to point to the exact location of it to make it work

Fixes: #6017

## Testing

- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-ubuntu2004-test/24/

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6049, opened 2023-04-18
fix(get_group0_members): Use python-driver instead cqlsh

aleksbykov:
for nonroot installation we don't have cqlsh in path. switch to use driver for getting group0 members instead cqlsh


fruch:
FYI, I've already addressed it in #6034, but I think still having it with the driver might help debugging timeout and such, since we have much more control on how the call is being made.

fruch (approved):
LGTM

soyacz (approved):

juliayakovlev (approved):

PR #6054, opened 2023-04-20
fix(upgrade_test): shim-signed package removal workaround

fgelcer:
since scylladb/scylla-enterprise#2818 is happening on one of our base versions, we need to do this
"manual" removal, so test won't fail.
i expect that, once scylladb/scylla-enterprise#2818 will be fixed, we can revert this commit.


yaronkaikov (approved):

upgrade_test.py
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env python
-
+import sys
```
soyacz, line 2: why added `sys` here?

fgelcer: you are right, leftover from previous trials, removing

soyacz (approved):

PR #6061, opened 2023-04-23
fix(rebuild): make sure we don't ever retry rebuild

fruch:
we should have retry on rebuild, it wouldn't never work, and would fail like this:
```
Command: "/usr/bin/nodetool -u cassandra -pw 'cassandra'  rebuild -- us-east "
Exit code: 1
Stdout:
nodetool: Scylla API server HTTP POST to URL '/storage_service/rebuild'
failed: std::runtime_error (Operation rebuild is in progress, try again)
See 'nodetool help' or 'nodetool help <command>'.
```

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1827,7 +1827,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     def repair_nodetool_rebuild(self):
         with adaptive_timeout(Operations.REBUILD, self.target_node, timeout=HOUR_IN_SEC * 48):
-            self.target_node.run_nodetool('rebuild')
+            self.target_node.run_nodetool('rebuild', retry=0)
 
     def disrupt_nodetool_cleanup(self):
         # This fix important when just user profile is run in the test and "keyspace1" doesn't exist.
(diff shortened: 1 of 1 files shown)


fgelcer (approved):
perhaps the default for `nodetool` commands must be without retries, and not only rebuild?

LGTM

fruch:
> perhaps the default for `nodetool` commands must be without retries, and not only rebuild?
> 

Some can be retried, like `nodetool status`, `nodetool gossipinfo`, remove it from all might create extra noise for us.
keep in mind, we removing retrying it mean we'll have more connectivity error raising up.

soyacz:
> > perhaps the default for `nodetool` commands must be without retries, and not only rebuild?
> 
> Some can be retried, like `nodetool status`, `nodetool gossipinfo`, remove it from all might create extra noise for us. keep in mind, we removing retrying it mean we'll have more connectivity error raising up.

Maybe we could add default 1-2 retries on network errors, despite of retry setting?

fruch:
> > > perhaps the default for `nodetool` commands must be without retries, and not only rebuild?
> > 
> > Some can be retried, like `nodetool status`, `nodetool gossipinfo`, remove it from all might create extra noise for us. keep in mind, we removing retrying it mean we'll have more connectivity error raising up.
> 
> Maybe we could add default 1-2 retries on network errors, despite of retry setting?
> 

This is the default, on network error we retry 3 times.

PR #6064, opened 2023-04-23
fix(sct.log): remove warning event from being printed

fruch:
* remove the warning events from being printed in sct.log, we have them in a seperate log if needed

* remove the `abort requested` logs from sct.log, we have too many of them, and they are not helpful for us

Ref: https://github.com/scylladb/scylladb/issues/10447

### DIFF
--- a/sdcm/db_log_reader.py
+++ b/sdcm/db_log_reader.py
@@ -46,6 +46,7 @@ class DbLogReader(Process):
         '] repair - Repair',
         'repair id [id=',
         '] stream_session - [Stream ',
+        '] storage_proxy - Exception when communicating with',
     ]
     # pylint: disable=too-many-arguments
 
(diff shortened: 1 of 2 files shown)


soyacz:
> * remove the warning events from being printed in sct.log, we have them in a seperate log if needed

Why we want to do to that? All the events are in separate logs so according to this logic we should stop printing them too...

fruch:
> > * remove the warning events from being printed in sct.log, we have them in a seperate log if needed
> 
> Why we want to do to that? All the events are in separate logs so according to this logic we should stop printing them too...

Since there lots of warnings and we aren't using them, not like critical/errors that we investigate and having them in the log is helpful. Most warnings are from db logs and we don't do anything with it, I've added it to DB logs so we won't treat them as errors.

fgelcer (approved):
LGTM

Issue #5415, opened 2022-11-14
Retying of _add_and_init_new_cluster_node cause the problems during nemesis terminate_and_replace_nemesis


juliayakovlev:
Function`sdcm.nemesis.Nemesis._add_and_init_new_cluster_node` runs with retying decorator:
```
 @retrying(n=3, sleep_time=60, allowed_exceptions=(NodeSetupFailed, NodeSetupTimeout))
    def _add_and_init_new_cluster_node(self, old_node_ip=None, timeout=MAX_TIME_WAIT_FOR_NEW_NODE_UP, rack=0):
```

It may cause to the problem during nemesis `terminate_and_replace_nemesis`.
In the test below this nemesis was triggered 3 times. 
Every run new added node (_node1_) replaced the terminated node, but initialization took more then 6 hours (it's default timeout for a node up) and NodeSetup timeout is raised. When it happens, we remove this new node from our cluster object:
```
        except (NodeSetupFailed, NodeSetupTimeout):
            self.log.warning("TestConfig of the '%s' failed, removing it from list of nodes" % new_node)
            self.cluster.nodes.remove(new_node)
            self.log.warning("Node will not be terminated. Please terminate manually!!!")
            raise
```
But node initialization was completed  later (after 6 hours) and this node in appear in Scylla cluster. This is `nodetool status` output, where `10.4.1.114` is new added node:
```
< t:2022-11-12 11:13:44,612 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > Status=Up/Down
|/ State=Normal/Leaving/Joining/Moving
--  Address     Load       Tokens       Owns    Host ID                               Rack
UN  10.4.0.64   1.04 TB    256          ?       62acdcac-49fb-4268-8abe-19fc912a8fc3  1a
UN  10.4.1.114  940.83 GB  256          ?       6c3dfcc3-4824-4d59-831f-d7bba9c622d3  1a
UN  10.4.2.34   1 TB       256          ?       ac5bbd2c-3c1c-4393-a652-410ddf0b0e41  1a
UN  10.4.0.93   1021.63 GB  256          ?       43c0f7df-8051-4dde-9ed8-21db0ebc2890  1a
```

But from SCT point of view new node setup failed and we start adding one more node (_node2_). _node2_ failed to startup because try to replace terminated node that already was replaced and not in the cluster. 
Due to _node2_ failure 3th node is adding (_node3_) and also fails with same error as _node2_.
[...]

PR #6065, opened 2023-04-24
fix(nemesis.py): Remove retry of adding a new node

yarongilor:
Retry adding a new node is problematic.
	It caused a failure of grow-shrink-cluster nemesis.
	Refs: https://github.com/scylladb/scylla-cluster-tests/issues/5415

Refs: https://github.com/scylladb/scylla-cluster-tests/issues/5415
[ICS Longevity 1TB Argus](https://argus.scylladb.com/test/d0b1d90e-dd6a-4a7c-96c3-3d31a43fc892/runs?additionalRuns[]=0e842a26-6b62-4164-ad08-513974539d88)

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1188,7 +1188,6 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if not ContainerManager.is_running(self.tester.localhost, 'ldap'):
             raise LdapNotRunning("LDAP server was supposed to be running, but it is not")
 
-    @retrying(n=3, sleep_time=60, allowed_exceptions=(NodeSetupFailed, NodeSetupTimeout))
     def _add_and_init_new_cluster_node(self, old_node_ip=None, host_id=None,
                                        timeout=MAX_TIME_WAIT_FOR_NEW_NODE_UP, rack=0):
         """When old_node_ip or host_id are not None then replacement node procedure is initiated"""


fgelcer (approved):
the problem here is not exactly the retry, but what you do with the failed nodes left in the cluster, when it fails...
like, if you timed out (or setup failed) adding a node, you must at least stop it, before retrying, otherwise you will get into conflicts

as a workaround, LGTM

fruch:
> the problem here is not exactly the retry, but what you do with the failed nodes left in the cluster, when it fails...
> like, if you timed out (or setup failed) adding a node, you must at least stop it, before retrying, otherwise you will get into conflicts
> 
> as a workaround, LGTM

I think the retry code is coming from days the SCT had lots issues. In case of failure to Scylla joining, there is no real reason to retry we want to be aware of the issue.

This retry is on top of too many actions, it never worked correctly, i.e. it's not aware what broke and what needs to be undo

fgelcer:
> This retry is on top of too many actions, it never worked correctly, i.e. it's not aware what broke and what needs to be undo

that is exactly what i meant... removing this retry is a workaround, and not a proper fix, and a proper fix will need to understand few scenarios, and act accordingly

soyacz (approved):
LGTM.
@fgelcer we might need some followup task to handle failures, but must be more specific - showing the scenario of failure.

fruch:
> > This retry is on top of too many actions, it never worked correctly, i.e. it's not aware what broke and what needs to be undo
> 
> that is exactly what i meant... removing this retry is a workaround, and not a proper fix, and a proper fix will need to understand few scenarios, and act accordingly

I think we can agree that failure to bootstrap  a  node we should stop the test, there's no point of continuing the test.
we assumed at a time that if we do retries it would enable us to carry on the test, I think we were wrong.

fruch:
> > > This retry is on top of too many actions, it never worked correctly, i.e. it's not aware what broke and what needs to be undo
> > 
> > 
> > that is exactly what i meant... removing this retry is a workaround, and not a proper fix, and a proper fix will need to understand few scenarios, and act accordingly
> 
> I think we can agree that failure to bootstrap a node we should stop the test, there's no point of continuing the test. we assumed at a time that if we do retries it would enable us to carry on the test, I think we were wrong.

I can understand smaller scope reties, like retry of the provisioning part on case of specific error, not the whole sequence of adding or replacing a node.

fgelcer:
@yarongilor , why is it still a draft?

yarongilor:
@fgelcer , @fruch , @soyacz ,
Following the above discussion, i'm not adding any additional code. It may be considered for the future.
This fix is good enough to be merged at this point.

fruch (approved):
LGTM

fgelcer:
> @fgelcer , @fruch , @soyacz , Following the above discussion, i'm not adding any additional code. It may be considered for the future. This fix is good enough to be merged at this point.

@yarongilor , please open a task in QA-Tasks for a followup work on it, as @fruch mentioned... so in case we do have a provision failing with some cloud backend related errors, we could be "smart" and do some retries, but if the error is in Scylla's side, then the test must fail.

PR #6079, opened 2023-04-28
fix(FileFollower): refactor iterator

aleksbykov:
Poller instance created on each iteration and
file registered  each time. If line is None or
not end on "\n" file is not unregistered and
Poller object could conitnue to eat memory while
will not be removed by garbage collector.

Create 1 Poller object and register/unregister
 file in Poller object only per iteration

Thanks to @dkropachev for help and providing solution

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -1598,15 +1598,19 @@ class FileFollowerIterator():  # pylint: disable=too-few-public-methods
     def __iter__(self):
         with open(self.filename, encoding="utf-8") as input_file:
             line = ''
+            poller = select.poll()  # pylint: disable=no-member
+            registered = False
             while not self.thread_obj.stopped():
-                poller = select.poll()  # pylint: disable=no-member
-                poller.register(input_file, select.POLLIN)  # pylint: disable=no-member
+                if not registered:
+                    poller.register(input_file, select.POLLIN)  # pylint: disable=no-member
+                    registered = True
                 if poller.poll(100):
                     line += input_file.readline()
                 if not line or not line.endswith('\n'):
                     time.sleep(0.1)
                     continue
                 poller.unregister(input_file)
+                registered = False
                 yield line
                 line = ''
             yield line


sdcm/utils/common.py
```diff
@@ -1598,18 +1598,21 @@ def __init__(self, filename, thread_obj):
...
                 if poller.poll(100):
                     line += input_file.readline()
                 if not line or not line.endswith('\n'):
                     time.sleep(0.1)
                     continue
                 poller.unregister(input_file)
+                registered = False
                 yield line
                 line = ''
-            yield line
```
soyacz, line 1612: won't removing this cause `StopIteration` exception somewhere?

soyacz (commented):
How did you find this issue? Was there some visible memory leak or you just found it by accident?

sdcm/utils/common.py
aleksbykov: it was lost during recreating pr. Return it back. it have to on place

aleksbykov:
> How did you find this issue? Was there some visible memory leak or you just found it by accident?

During investigation the jobs for 5.2 and 2023.1 https://jenkins.scylladb.com/job/enterprise-2023.1/job/longevity/job/longevity-schema-topology-changes-12h-test/4/, where job mostrly failed because hydra eat all memory in 4-5 hours of running. With help of @dkropachev who provide solution, we found that possible place where memory could be leaked. But looks that there is also another place where memory leak could happened.

aleksbykov:
@soyacz @temichus @fruch Could you review?

fruch:
> > How did you find this issue? Was there some visible memory leak or you just found it by accident?
> 
> During investigation the jobs for 5.2 and 2023.1 https://jenkins.scylladb.com/job/enterprise-2023.1/job/longevity/job/longevity-schema-topology-changes-12h-test/4/, where job mostrly failed because hydra eat all memory in 4-5 hours of running. With help of @dkropachev who provide solution, we found that possible place where memory could be leaked. But looks that there is also another place where memory leak could happened.

we commonly have those type of things introduced to SCT,  can you share some of the tools/methods used to identify the origin of those leaks ? so other might be able to investigate such issues ?

PR #6081, opened 2023-04-28
fix(disrupt_add_remove_mv): choose node without running nemesis

aleksbykov:
In parallel running of nemesis add_remove_mv nemesis could choose node where another nemesis already running.
Add filtering for choosing node free from running nemesis

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4205,46 +4205,56 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         Verify the MV can be used in a query.
         Finally, drop the MV.
         """
-        node1, node2 = self.cluster.nodes[:2]
-        ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=node2, filter_empty_tables=True, filter_out_mv=True)
-        if not ks_cfs:
-            raise UnsupportedNemesis(
-                'Non-system keyspace and table are not found. nemesis can\'t be run')
-        ks_name, base_table_name = random.choice(ks_cfs).split('.')
-        view_name = f'{base_table_name}_view'
-        with self.cluster.cql_connection_patient(node2) as session:
-            primary_key_columns = get_column_names(session=session, ks=ks_name, cf=base_table_name, is_primary_key=True)
-            # selecting a supported column for creating a materialized-view (not a collection type).
-            column = get_random_column_name(session=session, ks=ks_name,
-                                            cf=base_table_name, filter_out_collections=True)
-            if not column:
+
+        free_nodes = [node for node in self.cluster.nodes if not node.running_nemesis]
+        if not free_nodes:
+            raise UnsupportedNemesis("Not enough free nodes for nemesis. Skipping.")
+        cql_query_executor_node = random.choice(free_nodes)
+        self.set_current_running_nemesis(cql_query_executor_node)
+        try:
+            ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=cql_query_executor_node,
+                                                            filter_empty_tables=True, filter_out_mv=True)
+            if not ks_cfs:
                 raise UnsupportedNemesis(
-                    'A supported column for creating MV is not found. nemesis can\'t run')
-            column = f'"{column}"'
-            self.log.info("Stopping Scylla on node1")
-            node1.stop_scylla()
-            InfoEvent(message=f'Create a materialized-view for table {ks_name}.{base_table_name}').publish()
-            try:
-                with EventsFilter(event_class=DatabaseLogEvent,
-                                  regex='.*view - Error applying view update.*',
-                                  extra_time_to_expiration=30):
-                    self.tester.create_materialized_view(ks_name, base_table_name, view_name, [column],
-                                                         primary_key_columns, session,
-                                                         mv_columns=[column] + primary_key_columns)
-            except Exception as error:  # pylint: disable=broad-except
-                self.log.warning('Failed creating a materialized view: %s', error)
-                node1.start_scylla()
-                raise
-            try:
-                self.log.info("Starting Scylla on node1")
-                node1.start_scylla()
-                node1.run_nodetool(sub_cmd="repair -pr")
-                wait_for_view_to_be_built(self.target_node, ks_name, view_name, timeout=7200)
-                session.execute(SimpleStatement(f'SELECT * FROM {ks_name}.{view_name} limit 1', fetch_size=10))
-                sleep_for_percent_of_duration(self.tester.test_duration * 60, percent=1,
-                                              min_duration=300, max_duration=2400)
-            finally:
-                drop_materialized_view(session, ks_name, view_name)
+                    'Non-system keyspace and table are not found. nemesis can\'t be run')
+            ks_name, base_table_name = random.choice(ks_cfs).split('.')
+            view_name = f'{base_table_name}_view'
+            with self.cluster.cql_connection_patient(cql_query_executor_node) as session:
+                primary_key_columns = get_column_names(
+                    session=session, ks=ks_name, cf=base_table_name, is_primary_key=True)
+                # selecting a supported column for creating a materialized-view (not a collection type).
+                column = get_random_column_name(session=session, ks=ks_name,
+                                                cf=base_table_name, filter_out_collections=True)
+                if not column:
+                    raise UnsupportedNemesis(
+                        'A supported column for creating MV is not found. nemesis can\'t run')
+                column = f'"{column}"'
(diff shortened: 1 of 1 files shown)


sdcm/nemesis.py
```diff
@@ -4205,7 +4205,12 @@ def disrupt_add_remove_mv(self):
         Verify the MV can be used in a query.
         Finally, drop the MV.
         """
-        node1, node2 = self.cluster.nodes[:2]
+        try:
+            node1, node2 = [node for node in self.cluster.nodes if not node.running_nemesis][:2]
```
soyacz, line 4209: first, we should use `self.target_node` as the node which gets restarted, because now nemesis may take up to 3 nodes to be set as `running_nemesis`. It also will be released later when setting new target node before next nemesis.

The second thing is that still node2 can be left in `running_nemesis` state in case of some uncaught exception along the way. Still this solution is not safe.

fgelcer, line 4209: i'm not sure i understand the problem, but looks like you want to choose a node to be a target node for this nemesis, that doesn't have another nemesis running in it... and if that is the case, why not to:
```
self.target_node = random.choice([node for node in self.cluster.nodes if not node.running_nemesis])
```

sdcm/nemesis.py
```diff
@@ -4205,7 +4205,12 @@ def disrupt_add_remove_mv(self):
...
+        except ValueError:
+            raise UnsupportedNemesis("Not enough free nodes for nemesis. Skipping.") from ValueError
+        self.set_current_running_nemesis(node1)
+        self.set_current_running_nemesis(node2)
```
fgelcer, line 4213: why are you unsetting running nemesis on `node1` and `node2` ? if there is a nemesis running on them, why are you unsetting?

sdcm/nemesis.py
```diff
@@ -4234,6 +4239,8 @@ def disrupt_add_remove_mv(self):
             except Exception as error:  # pylint: disable=broad-except
                 self.log.warning('Failed creating a materialized view: %s', error)
                 node1.start_scylla()
+                self.unset_current_running_nemesis(node1)
+                self.unset_current_running_nemesis(node2)
```
fgelcer, line 4243: same here

sdcm/nemesis.py
```diff
@@ -4245,6 +4252,8 @@ def disrupt_add_remove_mv(self):
                                               min_duration=300, max_duration=2400)
             finally:
                 drop_materialized_view(session, ks_name, view_name)
+                self.unset_current_running_nemesis(node1)
+                self.unset_current_running_nemesis(node2)
```
fgelcer, line 4256: same here

sdcm/nemesis.py
fruch: I agree with @soyacz we should be using target node, and not inventing a new method to select node, node that we only do cql command on doesn't need to be selected as currently running nemesis.

sdcm/nemesis.py
aleksbykov: i am going to set the nemesis on them, so other parallel running nemesis could be executed on them

aleksbykov:
@soyacz @fruch @fgelcer @juliayakovlev could you review. I remove 3rd node. And hope save logic of nemesis

fruch (approved):
LGTM

sdcm/nemesis.py
```diff
@@ -4205,46 +4205,56 @@ def disrupt_add_remove_mv(self):
...
+                    self.target_node.start_scylla()
+                    self.target_node.run_nodetool(sub_cmd="repair -pr")
+                    wait_for_view_to_be_built(self.target_node, ks_name, view_name, timeout=7200)
+                    session.execute(SimpleStatement(f'SELECT * FROM {ks_name}.{view_name} limit 1', fetch_size=10))
+                    sleep_for_percent_of_duration(self.tester.test_duration * 60, percent=1,
+                                                  min_duration=300, max_duration=2400)
+                finally:
+                    drop_materialized_view(session, ks_name, view_name)
+        finally:
+            self.unset_current_running_nemesis(cql_query_executor_node)
```
fgelcer, line 4257: is this unset really needed? isn't the default action of each nemesis to `unset_current_running_nemesis` when the call ends?

fgelcer (approved):
2 small questions from my side, but LGTM

sdcm/nemesis.py
fgelcer: so maybe i don't understand the initial problem here... why using `self.target_node` isn't good enough?

sdcm/nemesis.py
```diff
@@ -4205,46 +4205,56 @@ def disrupt_add_remove_mv(self):
...
+                    raise UnsupportedNemesis(
+                        'A supported column for creating MV is not found. nemesis can\'t run')
+                column = f'"{column}"'
+                self.log.info("Stopping Scylla on node %s", self.target_node.name)
+                self.target_node.stop_scylla()
+                InfoEvent(message=f'Create a materialized-view for table {ks_name}.{base_table_name}').publish()
+                try:
+                    with EventsFilter(event_class=DatabaseLogEvent,
+                                      regex='.*view - Error applying view update.*',
+                                      extra_time_to_expiration=30):
```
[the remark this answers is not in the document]
fruch: I think it's in seconds...

It's not changed as part of this PR.

Anyhow it's a question to yaron that implemented it, and addressed this issue, he should have the answers if it's enough or not, assuming the cases that errors were raised were investigated

sdcm/nemesis.py
fgelcer: if so, the 1st line here (that creates a `free_nodes` list) must be under a condition, like
`if self.target_node.running_nemesis:`

fruch: We can't use the target node, since we want to do cql operation while it's down.

Target node is marked with nemesis before the disrupt function is even called, there's no point checking it.

fgelcer: ah, i see, now i got it

sdcm/nemesis.py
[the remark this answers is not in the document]
fgelcer: @yarongilor ?

sdcm/nemesis.py
soyacz: > I agree with @soyacz we should be using target node, and not inventing a new method to select node, node that we only do cql command on doesn't need to be selected as currently running nemesis.

@fruch But in that case, won't parallel nemesis cause failure of this one in case the second node gets disrupted?

fruch: if `cql_connection_patient_exclusive` is used yes, if `cql_connection_patient` is used we connect to all of the up nodes, so it would work regardless

fruch: to be on the safe side, assuming some code might, the nemesis is marking that node now.

We really should change the API of `cql_connection_patient` to not pass down a node, to avoid this confusion. cause the node isn't really get used, and by default we pass all ip of all nodes

sdcm/nemesis.py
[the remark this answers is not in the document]
yarongilor: i don't understand why this PR is needed. @aleksbykov , was there any failure in parallel nemesis you can point to?

aleksbykov:
Single nemesis passed: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/nemesis-clone/job/longevity-5gb-1h-AddRemoveMVNemeis-aws-test/

sdcm/nemesis.py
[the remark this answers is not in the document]
fruch: we shouldn't be picking nodes like that in a nemesis
```
node1, node2 = self.cluster.nodes[:2]
```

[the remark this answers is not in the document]
yarongilor: @fruch , this is why this nemesis has `disruptive = True`.
There are, for example other nemesis that are restarting ALL cluster nodes and don't define any additional "target node".
So if we want to fully address this case, we better have nemesis target_nodes as a list and not a single node definition.
This fix is just a very partial workaround for the general problem.

sdcm/nemesis.py
yarongilor: regardless of cql-connection, i don't think all non-disruptive nemeses can ran in parallel with this nemesis. For example - `NoCorruptRepairMonkey` would probably fail (with 3 nodes configuration).

sdcm/nemesis.py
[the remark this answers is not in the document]
fruch: The common case is one target node, in the uncommon cases we need to be more careful. (Like in the add/decommission node)

Changing the definition of the target node isn't really needed

Issue #6085, opened 2023-05-01
After a node reboot, we need to wait for scylla service to be up (not for the ssh to be up as we do now) 


rayakurl:
We need to wait for scylla to be up after we reboot a node, not for ssh to be up as we do now.

Nemesis for example: disrupt_decommission_streaming_err

During DecommissionStreamingErr nemesis we interrupt the decommission process on a db node with a hard reboot. Than, if the decommission was interrupted successfully (i.e. it's still in the cluster), we run a nodetool rebuild command on it. In this case however the rebuild didn't start, since we failed to ssh to longevity-10gb-3h-5-2-db-node-98470f24-2 .

We failed to ssh since cloud-init restarted the sshd (twice):
`
2023-04-27T21:31:25+00:00 longevity-10gb-3h-5-2-db-node-98470f24-2     !INFO | cloud-init[736]: + systemctl restart sshd
2023-04-27T21:31:25+00:00 longevity-10gb-3h-5-2-db-node-98470f24-2     !INFO | cloud-init[736]: + systemctl restart sshd
2023-04-27T21:31:25+00:00 longevity-10gb-3h-5-2-db-node-98470f24-2     !INFO | cloud-init[736]: + systemctl start sshd
`

At the same time the test tried to run nodetool rebuild via ssh:
`
< t:2023-04-27 21:31:50,323 f:remote_base.py  l:520  c:RemoteLibSSH2CmdRunner p:DEBUG > Running command "/usr/bin/nodetool  rebuild "...
`
and failed:
`
[...]

PR #2663, opened 2020-09-14
fix(wait for node up): wait for the target node up after reboot

juliayakovlev:
Wait for target node is up and normal after reboot to prevent failure due the node during
the start and in DN status

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [x] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6086, opened 2023-05-01
fix(nemesis): stream error wait full restart

fgelcer:
since we are not waiting for the node to fully
restart, we are experiencing some issues when
we try to run `nodetool` commands, and hence
adding here a proper way to make sure all the
required Scylla modules have started, before
calling next commands on the node that just
restarted.

Fixes: #6085

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3331,6 +3331,13 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 self.log.debug(f"{name}: failed to execute cleanup command "
                                f"{cmd} on node {node} due to the following error: {str(exc)}")
 
+    def wait_node_fully_start(self, node, timeout=3600):
+        self.log.info('Waiting scylla services to start after node reboot')
+        node.wait_db_up(timeout)
+        self.log.info('Waiting JMX services to start after node reboot')
+        node.wait_jmx_up()
+        self.cluster.wait_for_nodes_up_and_normal(nodes=[node])
+
     @decorate_with_context([
         ignore_ycsb_connection_refused,
         ignore_view_error_gate_closed_exception
@@ -3456,9 +3463,12 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if self.cluster.params.get('cluster_backend') == 'azure':
             timeout += 1200  # Azure reboot can take up to 20min to initiate
         ParallelObject(objects=[trigger, watcher], timeout=timeout).call_objects()
+
         if new_node := decommission_post_action():
+            self.wait_node_fully_start(new_node)
             new_node.run_nodetool("rebuild", retry=0)
         else:
+            self.wait_node_fully_start(self.target_node)
             self.target_node.run_nodetool(sub_cmd="rebuild", retry=0)
 
     def start_and_interrupt_repair_streaming(self):
@@ -3484,6 +3494,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             delay=1
         )
         ParallelObject(objects=[trigger, watcher], timeout=timeout).call_objects()
+
+        self.wait_node_fully_start(self.target_node)
+
         with adaptive_timeout(Operations.REBUILD, self.target_node, timeout=HOUR_IN_SEC * 48):
             self.target_node.run_nodetool("rebuild", retry=0)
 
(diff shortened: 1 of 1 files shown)


fgelcer:
giving it a try [here](https://jenkins.scylladb.com/job/enterprise-2023.1/job/Reproducers/job/longevity-3h-gce-stream-err/2/) as we are seeing it failing everywhere

sdcm/nemesis.py
```diff
@@ -3457,9 +3457,19 @@ def decommission_post_action():
         if self.cluster.params.get('cluster_backend') == 'azure':
             timeout += 1200  # Azure reboot can take up to 20min to initiate
         ParallelObject(objects=[trigger, watcher], timeout=timeout).call_objects()
+
+        def wait_restarted_node_fully_start(node):
+            self.log.info('Waiting scylla services to start after node reboot')
+            node.wait_db_up()
+            self.log.info('Waiting JMX services to start after node reboot')
+            node.wait_jmx_up()
+            self.cluster.wait_for_nodes_up_and_normal(nodes=[node])
```
fruch, line 3466: Doing all three checks  seems like an overkill to me...

fruch (commented):
There is one more nemesis that calls reboot like that, that we should also fix

sdcm/nemesis.py
fgelcer: in one hand, if we moved to the next check, it should no take much... we could have only the 3rd one, but maybe it is not enough?

yarongilor (commented):
Why not having this new code (optionally/ by default) be included in `run_nodetool` function as well? why not prevent running a nodetool command if scylla is not up on node?

sdcm/nemesis.py
```diff
@@ -3332,6 +3332,14 @@ def _run_commands_wait_and_cleanup(  # pylint: disable=too-many-arguments
                 self.log.debug(f"{name}: failed to execute cleanup command "
                                f"{cmd} on node {node} due to the following error: {str(exc)}")
 
+    def wait_restarted_node_fully_start(self, node):
```
fruch, line 3335: looks like this is following https://github.com/scylladb/scylla-cluster-tests/pull/2663 footsteps.

and seems like a pattern across multiple places (and not just those too), we should consider moving this function into cluster or node

sdcm/nemesis.py
```diff
@@ -3515,7 +3529,7 @@ def start_and_interrupt_rebuild_streaming(self):
             delay=1
         )
         ParallelObject(objects=[trigger, watcher], timeout=timeout + 60).call_objects()
-        self.target_node.wait_db_up(timeout=300)
```
fruch, line 3518: you just lost the timeout the nemesis has set...

fruch:
@fgelcer pre-commit stage has failed...

sdcm/nemesis.py
fgelcer: add the timeout (but the problem is, there are 3 different "wait" functions, each has a different timeout default and format)

sdcm/nemesis.py
fgelcer: we can consider it in a PR where we can organize everything.. as the purpose of this PR is to fix a persistent issue we are seeing while trying to release 5.2

fgelcer:
> Why not having this new code (optionally/ by default) be included in `run_nodetool` function as well? why not prevent running a nodetool command if scylla is not up on node?

cause i can imagine scenarios where you do want to test `nodetool` commands in some sort of unhealth cluster...
also, functions should have 1 responsibility, so `run_nodetool` should run `nodetool` commands, and not run checks if either the cluster is up or not...

fgelcer:
> @fgelcer pre-commit stage has failed...

fixed

roydahan:
@fruch please review

roydahan:
@aleksbykov please review.

sdcm/nemesis.py
```diff
@@ -3332,6 +3332,13 @@ def _run_commands_wait_and_cleanup(  # pylint: disable=too-many-arguments
                 self.log.debug(f"{name}: failed to execute cleanup command "
                                f"{cmd} on node {node} due to the following error: {str(exc)}")
 
+    def wait_restarted_node_fully_start(self, node, timeout=3600):
```
aleksbykov, line 3335: better to rename to `wait_node_fully_start`

aleksbykov (approved):
LGTM, small comment to name. Also in future pr, this checks should be moved to method reboot node or method nodetool should have check that jmx is up.

sdcm/nemesis.py
fgelcer: maybe in the future, if moved into `cluster.py` it will make more sense, but since this function is here being used to wait to restarted nodes to fully start, the current name makes more sense.
fixed anyway

fruch (approved):
LGTM, 

Can you open a task from moving it to cluster and align all the places we use this sequence.

fgelcer:
> LGTM,
> 
> Can you open a task from moving it to cluster and align all the places we use this sequence.

https://github.com/scylladb/qa-tasks/issues/1146

fgelcer:
ping @aleksbykov @yarongilor @rayakurl

aleksbykov (approved):
LGTM

fruch:
> @scylladb/qa-maintainers Can this be backported to 2022.1? We see a similiar error there as well

done

PR #6039, opened 2023-04-18
fix(health_events): add retry and did not raise exception

juliayakovlev:
Nemesis thread in the https://jenkins.scylladb.com/job/enterprise-2023.1/job/longevity/job/ longevity_mv_synchronous_updates-12h/3/ test failed due to 'get_group0_members' timeout.

Add retry decorator and did not raise exception on 'get_group0_members' and 'get_token_ring_members' during node health check.

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6090, opened 2023-05-03
fix(get_group0_members): re-raise the exception to make retry work

ilya-rarov:
The retry for this method was introduced in the PR https://github.com/scylladb/scylla-cluster-tests/pull/6039
Since most of the code of `get_group0_members` is placed in `try` block any possible exception is concealed inside the `except` block and does not go outside.
So the `retrying` decorator never has a reason to start retrying.
Added `raise` for the exception to trigger the retry for the method.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3064,6 +3064,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             self.log.error(err_msg)
             InfoEvent(message=err_msg, severity=Severity.ERROR).publish()
 
+            raise
+
         self.log.debug("Group0 members: %s", group0_members)
         return group0_members
 


ShlomiBalalis (approved):

sdcm/cluster.py
```diff
@@ -3064,6 +3064,8 @@ def get_group0_members(self) -> list[dict[str, str]]:
             self.log.error(err_msg)
             InfoEvent(message=err_msg, severity=Severity.ERROR).publish()
 
+            raise exc
```
fgelcer, line 3067: AFAIK you can just:
```suggestion
            raise
```

fgelcer (commented):
little comment, and there are checks that failed, but other than that, LGTM

sdcm/cluster.py
fgelcer: > little comment, and there are checks that failed, but other than that, LGTM

i re-triggered the checks, as the precommit one seems to be failed on an environment error, and may go away when re-running it

ilya-rarov: > AFAIK you can just:

np, changed this

fgelcer (approved):
LGTM

fruch (approved):
LGTM

PR #6092, opened 2023-05-04
fix(yaml): Add config yaml to disable speculative retry

aleksbykov:
For performance tests speculative retry could affect on max throughput. Disable it on scylla server side with new additional config yaml for keyspace1.standard1 table, created by c-s performance tests

### DIFF
--- a/configurations/disable_speculative_retry.yaml
+++ b/configurations/disable_speculative_retry.yaml
@@ -1,0 +1,1 @@
+post_prepare_cql_cmds: "ALTER TABLE keyspace1.standard1 WITH speculative_retry = 'NONE'"


configurations/disable_speculative_retry.yaml
```diff
@@ -0,0 +1 @@
+post_prepare_cql_cmds: "ALTER TABLE keyspace1.standard1 WITH speculative_retry = 'NONE'"
```
fruch, line 1: where's the pipeline that uses this one ?

i.e. what's the point of merging this on it's own ?

fgelcer (commented):
agree with @fruch , let's add this one while using it somewhere

configurations/disable_speculative_retry.yaml
aleksbykov: We give a try it now for performance throughput test. If result will be good and we got approval to disable speculative retry for perf throughput test, it will be just added to current pipelines: jenkins-pipelines/perf-regression-throughput-30min.jenkinsfile

The goal to add in such way, is to easly use this in current perf tests for investigation

aleksbykov: @fruch @fgelcer

fgelcer (approved):
i think it won't hurt us to have this config file in the repo, so LGTM

fruch:
> i think it won't hurt us to have this config file in the repo, so LGTM

I won't, but I've seen too many cases things like this were added with no context, and actual usage ended up in a release branch and not in master (i.e. audit for example)

PR #6096, opened 2023-05-05
fix(c-s timeout): prolong timeout for c-s

soyacz:
Sometimes c-s fails due timeout and create critical event. We don't know the reason of it and would like to measure how much is actually needed.

Wrap c-s command around soft-timeout context manager, so we don't kill c-s and still get error event in case of timeout.
Prolong by 5 minutes hard timeout to reduce false timeout events ratio.


soyacz:
ping @fruch @juliayakovlev

fruch (approved):
LGTM

PR #6015, opened 2023-04-14
fix(cluster): Add checks for raft disabling

aleksbykov:
Some jobs for 5.2 could run with raft disabled.
In this case group0 is empty and not need to check consistency of group0 and tokenring

Add new method to get status of raft (is_enabled, is_ready) if raft is disabled, group0_members will be empty, and consistency check will be skipped

Failed jobs: 
- https://jenkins.scylladb.com/job/enterprise-2023.1/job/upgrade-with-raft/job/rolling-upgrade-ubuntu22.04-with-raft-test/4/
- https://jenkins.scylladb.com/job/enterprise-2023.1/job/disabled_raft/job/longevity-1tb-7days-test/3/

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6097, opened 2023-05-06
fix(k8s): make raft checks work on K8S

vponomaryov:
Due to the K8S specifics we do not use the `ScyllaYaml` class.
So, when common code assumes we use it we get the following error:

    'dict' object has no attribute 'consistent_cluster_management'

The reason for it is that we use completely different approach in K8S.
So, fix the newly introduced raft handling code [1] to work with K8S.

[1] https://github.com/scylladb/scylla-cluster-tests/pull/6015


fruch (approved):
LGTM

Issue #6095, opened 2023-05-04
[K8S] 14 tenant test timeout on setup of clusters 


fruch:
While working on enabling sni proxy for the 14 tenants test cases

A run base on SCT master that timeout: 
https://argus.scylladb.com/test/37ffa9a1-870c-4858-8404-4cea2c335766/runs?additionalRuns[]=e02d9abf-8e61-4ff1-a74f-c96796c785a8

Last working run:
https://argus.scylladb.com/test/37ffa9a1-870c-4858-8404-4cea2c335766/runs?additionalRuns[]=abc8aebb-9279-4400-9469-ab9a1c375ab5

* Tried taking the image of the working run, didn't help
* Tried running with the stable v1.8.0 operator version didn't help
* Tried reverting a1f63c77fd59b6438a35d2415f5eb5a24947d183 and  d95ab1b89965df77b795eb84c562815298774d4e, also didn't helped

vponomaryov:
At first, I found out that https://github.com/scylladb/scylla-cluster-tests/pull/6042 has changed the approach for creating K8S scylla pods.
Before we were creating and processing scylla pods by 1.
Now we create all the pods, wait for it and then process all of them in a narrow time range.
So, API call density got increased. Especially for multitenant case.

Then, I made it work the old way and ran a CI job: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-12h-multitenant-eks/40/consoleFull
It passed the problematic point.
BUT, I haven't been able to reproduce it anymore.

Then, playing with it, I found out that some of threads (3-10 out of 14) hang on various steps inside of the `_create_node` method calls where we do lots of different stuff including API calls.

Then, I bumped the K8S version and the `vpc_cni` plugin for it. The same results.

Then, I tried to apply the API rate limiter for the EKS and it didn't help too.
Only increased spent time in places before the problematic point.

So, to me, it sounds like something has changed in the platform itself.

P.S. single cluster creation works if we merge the fix for yet another breakage: https://github.com/scylladb/scylla-cluster-tests/pull/6097

fruch:
I've merged #6097, and we need to keep on this investigation...

fruch:
I'm trying one more guess, reverting latest change to the hydra image, maybe there were changes there we didn't notices,
since whom introduced it did add the change to the `requirements.txt` to his PR

fruch:
[...]

PR #6101, opened 2023-05-07
fix(k8s): run only one 'node.init()' method at once

vponomaryov:
Recently we found out that K8S multitenant setup fails due to multiple threads (3-10 out of 14).
And the point of hanging is the `node.init()` method which gets called 42 times in a narrow time range running 14 tenants.

So, make run only 1 at a time using the lock mechanism.

Fixes: #6095

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -97,6 +97,7 @@ ANY_KUBERNETES_RESOURCE = Union[  # pylint: disable=invalid-name
     Resource, ResourceField, ResourceInstance, ResourceList, Subresource,
 ]
 NAMESPACE_CREATION_LOCK = Lock()
+NODE_INIT_LOCK = Lock()
 
 CERT_MANAGER_TEST_CONFIG = sct_abs_path("sdcm/k8s_configs/cert-manager-test.yaml")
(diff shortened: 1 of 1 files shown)


vponomaryov:
Verified here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-12h-multitenant-eks/61/console

fruch:
@vponomaryov do we know which code inside `node.init()` is the problem ? is there the `start_task_threads` ?

fruch:
@vponomaryov 
this didn't solved the issue, I've took it into my branch, and it still timeout:
https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-scylla-operator-12h-multitenant-eks/46/

vponomaryov:
> @vponomaryov do we know which code inside `node.init()` is the problem ? is inside `start_task_threads` ? i.e. clearly this is a patch for something we don't seem to understand why it's hanging, I rather understand it first before apply a patch to hide it.

It was hanging on completely different steps, kind of `freezing`. So, there was no more specific place than the whole `init()` method.

> @vponomaryov this didn't solved the issue, I've took it into my branch, and it still timeout: https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-scylla-operator-12h-multitenant-eks/46/

In my case it passed 2 times from 2, and without it 0 out of ~15.

fruch (approved):
LGTM

PR #6107, opened 2023-05-08
improvement(k8s-eks): Update K8S node tags and SGs only once

vponomaryov:
For the moment we update EC2 nodes with our tags and security groups for each of the Scylla pods.
When we run just 1 Scylla cluster everything is ok. But when we run multitenant setup we duplicate these calls by number of tenants. And highest number of tenants we use is 14.

So, remove that redundancy just by memoizing the completed updates.

### DIFF
--- a/sdcm/cluster_k8s/eks.py
+++ b/sdcm/cluster_k8s/eks.py
@@ -16,6 +16,7 @@ import logging
 import time
 import pprint
 from textwrap import dedent
+from threading import Lock
 from typing import List, Dict, Literal, ParamSpec, TypeVar
 from functools import cached_property
 from collections.abc import Callable
@@ -50,6 +51,8 @@ R = TypeVar("R")  # pylint: disable=invalid-name
 
 # we didn't add configuration for all the rest 'io1', 'io2', 'gp2', 'sc1', 'st1'
 SUPPORTED_EBS_STORAGE_CLASSES = ['gp3', ]
+
+EC2_INSTANCE_UPDATE_LOCK = Lock()
 
 
 # pylint: disable=too-many-instance-attributes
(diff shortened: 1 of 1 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-12h-multitenant-eks/64/console

sdcm/cluster_k8s/eks.py
```diff
@@ -645,8 +667,8 @@ def add_nodes(self,
                                       rack=rack,
                                       enable_auto_bootstrap=enable_auto_bootstrap)
         for node in new_nodes:
-            self.k8s_cluster.set_security_groups(node.ec2_host)
-            self.k8s_cluster.set_tags(node.ec2_host.id)
+            self.k8s_cluster.set_security_groups(node.ec2_instance_id)
+            self.k8s_cluster.set_tags(node.ec2_instance_id)
```
soyacz, line 671: why not setting it when provisioning instances e.g. in `get_cluster_k8s_eks`?

vponomaryov: Because we terminate nodes (which causes auto-recreation) in nemesis and do upgrades.
In these cases we get new nodes which do not have tags and sgs be applied.

fruch: there no way to apply those on the nodegroup ?

vponomaryov: > there no way to apply those on the nodegroup ?

https://github.com/scylladb/scylla-cluster-tests/blob/38b7b3e04abfc33029d87b5e6b13792967ee9460/sdcm/cluster_k8s/eks.py#L457
https://github.com/scylladb/scylla-cluster-tests/blob/38b7b3e04abfc33029d87b5e6b13792967ee9460/sdcm/cluster_k8s/eks.py#L470

fruch: Seems like we might consider switch to Launch Templates, so we can pass both Tags and SGs in it.

It's shame that they don't do it out of the box...

fruch: @vponomaryov can you open a task to checkout if how complex is gonna be to use Launch Templates for the node group ?

fruch (approved):
LGTM

soyacz (approved):

sdcm/cluster_k8s/eks.py
vponomaryov: @fruch https://github.com/scylladb/qa-tasks/issues/1155

PR #6115, opened 2023-05-08
feature(gemini): update to version 1.8.0

fruch:
Update packages and golang (scylladb/gemini#296)
Fix(inflight): reduce memory footprint by using struct{} instead of bool (scylladb/gemini#291) Fix warmup hunging (scylladb/gemini#292)


docker/gemini/Dockerfile
```diff
@@ -7,11 +7,11 @@ RUN apk add curl
 ARG version
 ENV version=${version:-1.7.6}
 
-RUN curl -LO https://s3.amazonaws.com/downloads.scylladb.com/gemini/${version}/gemini_${version}_Linux_x86_64.tar.gz ;\
+RUN curl -LO https://github.com/scylladb/gemini/releases/download/v${version}/gemini_${version}_Linux_x86_64.tar.gz ;\
     tar -xvf gemini_${version}_Linux_x86_64.tar.gz ;\
     chmod a+x gemini
 
-FROM bitnami/minideb:bullseye
+FROM ubuntu:22.04
```
fruch, line 14: needed to match the glibc that is using in building gemini in github actions

soyacz: isn't minideb using glibc?

fruch:
@dkropachev 

Seems like with all the fixes it's still segfaulting 😭 
https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/gemini-1tb-10h-test/7/

docker/gemini/Dockerfile
soyacz: btw. which github actions are you referring to?

fruch: > isn't minideb using glibc?

It is but it wasn't the correct version

fruch: > btw. which github actions are you referring to?

I thought it was built by CI

But looking at callstack from it, it was built on someone's local machine.

we should have a proper CI to build it

roydahan:
@fruch is it segfaulting the same way as it was or looks like something new?

vponomaryov (approved):
Even if it segfaults the same way, don't see a reason not to switch to newer version.
Especially knowing that there will be more updates further.

roydahan (approved):

PR #6117, opened 2023-05-09
fix(load_and_stream): fix log followers

soyacz:
In case `load_and_stream` finish too quickly we get into error when we get ending stream log line when getting lines to verify if it started. In that case wait for ending will fail as `done` line has already been consumed.

Fix is about splitting log follower to 2: one for start and another for done - so waiting for start will not consume lines about stream is done.


juliayakovlev (approved):

vponomaryov (approved):
LGTM

PR #6122, opened 2023-05-10
fix(manager): bumping manager version

fgelcer:
manager 3.1 was recently released, and then
we want to start using it in our regular tests.
For it, i added the repository list of manager
3.1, and changed the `test_default.yaml` to
use it.
also, removing 2.6, as we support last 2
versions only.

### DIFF
--- a/defaults/manager_versions.yaml
+++ b/defaults/manager_versions.yaml
@@ -7,6 +7,14 @@ manager_repos_by_version:
     ubuntu18: 'http://downloads.scylladb.com/manager/deb/unstable/unified-deb/master/latest/scylla-manager.list'
     ubuntu20: 'http://downloads.scylladb.com/manager/deb/unstable/unified-deb/master/latest/scylla-manager.list'
     ubuntu22: 'http://downloads.scylladb.com/manager/deb/unstable/unified-deb/master/latest/scylla-manager.list'
+  "3.1":
+    centos7: 'http://downloads.scylladb.com.s3.amazonaws.com/rpm/centos/scylladb-manager-3.1.repo'
+    centos8: 'http://downloads.scylladb.com.s3.amazonaws.com/rpm/centos/scylladb-manager-3.1.repo'
+    debian10: 'http://downloads.scylladb.com.s3.amazonaws.com/deb/debian/scylladb-manager-3.1.list'
+    debian11: 'http://downloads.scylladb.com.s3.amazonaws.com/deb/debian/scylladb-manager-3.1.list'
+    ubuntu18: 'http://downloads.scylladb.com.s3.amazonaws.com/deb/ubuntu/scylladb-manager-3.1.list'
+    ubuntu20: 'http://downloads.scylladb.com.s3.amazonaws.com/deb/ubuntu/scylladb-manager-3.1.list'
+    ubuntu22: 'http://downloads.scylladb.com.s3.amazonaws.com/deb/ubuntu/scylladb-manager-3.1.list'
   "3.0":
     centos7: 'http://downloads.scylladb.com.s3.amazonaws.com/rpm/centos/scylladb-manager-3.0.repo'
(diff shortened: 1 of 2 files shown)


roydahan (approved):

fgelcer:
> Please remove manager 2.6 as well

done

ShlomiBalalis (approved):
> > Please remove manager 2.6 as well
> 
> done

And add it to the commit message

fgelcer:
> > > Please remove manager 2.6 as well
> > 
> > 
> > done
> 
> And add it to the commit message

done... something else?

ShlomiBalalis (approved):
> > > > Please remove manager 2.6 as well
> > > 
> > > 
> > > done
> > 
> > 
> > And add it to the commit message
> 
> done... something else?

~~The commit really shouldn't be a fix~~

# Nope!

roydahan (approved):

roydahan:
Let it run in the weekend runs, if all good we will backport to 5.2 and 2023.1

fgelcer:
@ShlomiBalalis to check master runs from last couple of weeks, and to confirm we don't have any regressions, so we can finally backport it to 2023.1 and 5.2

roydahan:
Are we backporting this?
Did we check master runs with it?

ShlomiBalalis:
> Are we backporting this? Did we check master runs with it?

I looked over several of the latest longevity runs and I can confirm there was not a single manager failure. We're GTG

roydahan:
did we ever compared the repair / backup time between 3.0 and 3.1?

ShlomiBalalis:
> did we ever compared the repair / backup time between 3.0 and 3.1?

Not officially, but I didn't see any significant difference here or there. Which makes sense since that part of the code hasn't changed in 3.1

roydahan:
Still need to be validated by @fgelcer , but in a latest performance test it took double the time.

fgelcer:
> Still need to be validated by @fgelcer , but in a latest performance test it took double the time.

from a comment by @michoecho it seems to be a problem in Scylla core, and not related to the manager...
i'm running a reproducer to compare a master where repair took 3h 45m against latest master that took 7h

ShlomiBalalis:
> > Still need to be validated by @fgelcer , but in a latest performance test it took double the time.
> 
> from a comment by @michoecho it seems to be a problem in Scylla core, and not related to the manager... i'm running a reproducer to compare a master where repair took 3h 45m against latest master that took 7h

@fgelcer Was it in scylla core, then? Can we backport?

michoecho:
> > > Still need to be validated by @fgelcer , but in a latest performance test it took double the time.
> > 
> > 
> > from a comment by @michoecho it seems to be a problem in Scylla core, and not related to the manager... i'm running a reproducer to compare a master where repair took 3h 45m against latest master that took 7h
> 
> @fgelcer Was it in scylla core, then? Can we backport?

I investigated the problem in more detail, and it's not fixable.

Simplified, the phenomenon is this:  Scylla exhibits some performance bimodality. Sometimes a core (interestingly: not a shard, but the full 2-shard core) can become ~10% slower than usual, for no particular (actionable) reason.

In the "good" runs, the interactive workload of cassandra-stress requires ~80% of the core's CPU to be sustained. But in the "bad" runs, some cores become slower by 10%, which raises the requirement to ~90%.
As a consequence, non-interactive workloads, like compaction and repair, receive ~10% instead of ~20% of CPU (and are also slower by 10% themselves), which makes repair twice as long.

As for what is the true reason of the bimodality, potential answers include:
- variations in how kernel distributes work over cores
- variations in how Scylla uses CPU caches
- variations in CPU frequency
- and more

I don't know which one it is and finding out would be probably very hard, if possible at all. And even if we knew, we probably couldn't do anything about either.

So I think the best thing to do with this phenomenon is to cope with it.

Issue #5969, opened 2023-03-31
Automatic NetworkTopologyStrategy for c-s and CREATE commands


soyacz:
In SCT we mostly use simpleStrategy for c-s workflows. While it's convenient it is not reflecting recommended production replication strategy which is network one. 
I propose to make c-s to automatically switch replication strategy from simple to network in the runtime (requires region knowledge).
benefits:
- more production like testing
- we could drop setting network replication strategies parameters in c-s commands so we can switch regions for the tests where we specify it

soyacz:
Also should fix region names for cases we already hardcode regions when using NetworkTopology.

soyacz:
Note: 
Currently we don't use Ec2MultiRegionSnitch and we apply some own way of naming datacenters in SCT.
Test how Ec2MultiRegionSnitch works and how it names dc's. Adapt SCT to use similar if needed.

fgelcer:
any progress with this task?

soyacz:
> any progress with this task?

not yet, it's planned.

roydahan:
Do we still need it? (I think not, right?)

soyacz:
Yes, it's fixed

PR #6126, opened 2023-05-11
feature(stress): automatic dc names for NetworkStrategy

soyacz:
When we use NetworkTopologyStrategy in stress commands we currently
hardcode datacenter names. This leads to failed tests when executed in
different regions.

Make dc names to be replaced dynamically so we don't face this issue
anymore. Use placeholders like `<dcX>` in c-s, precreate_schema and
profile files.

refs: https://github.com/scylladb/scylla-cluster-tests/issues/5969

### DIFF
--- a/unit_tests/dummy_remote.py
+++ b/unit_tests/dummy_remote.py
@@ -85,12 +85,20 @@ class LocalLoaderSetDummy(BaseCluster):
     def is_kubernetes():
         return False
 
+    @staticmethod
+    def replace_dc_placeholders_with_dc_names(cmd):
+        datacenter_names = ['datacenter1']
+        for idx, dc in enumerate(datacenter_names):
+            cmd = cmd.replace(f"<dc{idx}>", dc)
+        return cmd
+
 
 class LocalScyllaClusterDummy(BaseScyllaCluster):
     # pylint: disable=super-init-not-called
     def __init__(self):
         self.name = "LocalScyllaClusterDummy"
         self.params = {}
+        self.datacenter_names = ["datacenter1"]
 
     @staticmethod
     def get_db_auth():
--- a/configurations/manager/200gb_dataset.yaml
+++ b/configurations/manager/200gb_dataset.yaml
@@ -2,4 +2,4 @@ test_duration: 1200
 
 instance_type_db: 'i3.xlarge'
 
-stress_cmd: "cassandra-stress write cl=QUORUM n=200200300 -schema 'replication(strategy=NetworkTopologyStrategy,us-eastscylla_node_east=2,us-west-2scylla_node_west=1)' -mode cql3 native -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=400200300..600400600"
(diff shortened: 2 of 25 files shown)


soyacz:
another approach could be by guessing dc names during preparing sct_config... from one hand less complicated but from the other more... @fruch  WDYT?

unit_tests/dummy_remote.py
```diff
@@ -85,12 +85,20 @@ def get_db_auth():
     def is_kubernetes():
         return False
 
+    @staticmethod
+    def replace_dc_placeholders_with_dc_names(cmd):
+        datacenter_names = ['datacenter1']
+        for idx, dc in enumerate(datacenter_names):
+            cmd = cmd.replace(f"<dc{idx}>", dc)
```
fruch, line 92: Maybe we can also handle the case
Of placeholder wasn't set ?

And replace the whole setting with the correct layout ? Or at least fail the test notifying us the actual datacenter isn't aligned with the c-s command ?

soyacz: I'm not sure if I understand you.
Generally, I wanted to keep possibility to hardcode (for whatever the reason) and also possibility to still use SimpleStrategy.
Also I thought of replacing all c-s commands in the repo which use NetworkStrategy to apply there placehloders. 

In case we don't want to ever use hardcoded datacenters, we could add the check at sct_config level - fail in case NetworkTopology and hardcoded datacenter names.

soyacz:
> Another with profile in progress: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-cdc-8h-multi-dc-topology-changes-test/2/

Failed, retest with fix: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-cdc-8h-multi-dc-topology-changes-test/3/

soyacz:
test passed with profile: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-cdc-8h-multi-dc-topology-changes-test/4/ (failure was due spot termination, but schema was already created and nemesis running)

vponomaryov:
@soyacz Please, rebase it removing the merge conflicts.

vponomaryov:
Why is it labeled that `backports` are not needed when we use the `network topology strategy` in the `5.2` with the `hardcoded` region values?
I think it must be backported there.

fgelcer (commented):
is PR still necessary?
why not push everything to #6178 ?

soyacz:
> is PR still necessary? why not push everything to #6178 ?

#6178 is limited only to cases when replication factor is equal for each DC. With this fix we'll be able to cover that cases too - although currently there are not so many cases like that. Let's first merge #6178 and then see how it works. Changing to draft, for now.

fruch:
@soyacz are we still needs this one ?

soyacz:
@fruch this is for handling cases where we use different RF factors for different DC's - currently these cases have hardcoded ones. Because this is rare scenario and we don't change regions when running (I suppose), we can close it without merging.

PR #6130, opened 2023-05-11
feature(k8s): install scylla-operator's monitoring in K8S

vponomaryov:
Add the new scylla-opeator's monitoring support.
Also, add our SCT dashboard to it.
    
Following new config options were added:
- `k8s_n_monitor_nodes` - optional. Gets used for provisioning
  of the scylla-operator's monitoring pods. If not set or set to '0'
  then `auxiliary` node pool will be used.
- `k8s_instance_type_monitor` - optional. If not set then
  the `instance_type_monitor` option value will be used for EKS
  and `gce_instance_type_monitor` for GKE.
  In case of local K8S setups is just ignored.
    
The existing `k8s_deploy_monitoring` config option is still
controls the installation of the scylla-operator's monitoring
during the `provisioning resources` step.
Useful for longevity test runs.
    
In case of the `K8S functional tests` it is not really needed because
2 new functional tests were added which cover both monitoring types
supported by the latest, for now, scylla-operator.
    
Works on each of the K8S backends and expected to work also in multi-tenant setups.
Doesn't depend on the SNI setup.


vponomaryov:
The SCT dashboard needs this improvement PR: https://github.com/scylladb/scylla-cluster-tests/pull/5991

soyacz:
how can we open these dashboards when deployed?

vponomaryov:
> how can we open these dashboards when deployed?

There are 2 ways:
1) `https`: Domain name + port (`{cluster_name}-grafana.{namespace}.svc.cluster.local:3000`) making the DN be resolved into the IP address of the grafana pod/service.
2) `http`: `IP:3000` with insecure session.

Both will require grafana login/pass which can be taken from the secret object in K8S.

defaults/k8s_local_kind_config.yaml
```diff
@@ -27,6 +27,7 @@ k8s_use_chaos_mesh: true
 # NOTE: if we do not specify 'k8s_n_loader_pods_per_cluster' then value of the 'n_loaders' is used
 n_loaders: 1
 n_monitor_nodes: 1
+k8s_n_monitor_nodes: 0
```
fruch, line 30: shouldn't we have at least one test case with this enabled ? 
or we plan to enable it only for EKS based job ?

fruch (approved):
LGTM

good at a beginning, but we should have at least one job with this enabled so we can start testing it on a regular basis

soyacz (approved):
> > how can we open these dashboards when deployed?
> 
> There are 2 ways:
> 
> 1. `https`: Domain name + port (`{cluster_name}-grafana.{namespace}.svc.cluster.local:{port}`) making the DN be resolved into the IP address of the grafana pod/service. (This one is used here to add the SCT dashboard)
> 2. `http`: `{ip_addr}:{port}` with insecure session.
> 
> Both will require grafana login/pass which can be taken from the `secret` object in K8S.

ugh, not that straightforward thing to do. Worth documenting on the occasion how to do that (or add some script to get all the details). Also question, does log collector collect it? can we restore it later (and will we know credentials?).

besides this remark which can be done in followup task, LGTM.

fruch:
> > > how can we open these dashboards when deployed?
> > 
> > There are 2 ways:
> > 
> > 1. `https`: Domain name + port (`{cluster_name}-grafana.{namespace}.svc.cluster.local:{port}`) making the DN be resolved into the IP address of the grafana pod/service. (This one is used here to add the SCT dashboard)
> > 2. `http`: `{ip_addr}:{port}` with insecure session.
> > 
> > Both will require grafana login/pass which can be taken from the `secret` object in K8S.
> 
> ugh, not that straightforward thing to do. Worth documenting on the occasion how to do that (or add some script to get all the details). Also question, does log collector collect it? can we restore it later (and will we know credentials?).
>
It was agreed we don't yet collect it.
This PR sets the base for installing it via operator, next step would be setting it in EKS via sni proxy, and that should grant us public access to it (with known given credentials picked from the right secrects)
It was agreed that since this part is under test, will not switch to work soley with this monitor, and enable it only on specific job intend to test operator monitor installation

> besides this remark which can be done in followup task, LGTM.

vponomaryov:
> ugh, not that straightforward thing to do. Worth documenting on the occasion how to do that (or add some script to get all the details). 

https://github.com/scylladb/scylla-operator/blob/master/docs/source/monitoring.md

vponomaryov:
@fruch , @soyacz 
Operator has 2 separate modes for monitoring.
So, I think about adding 2 separate functional tests to cover the monitoring case.
It will require small rework of the existing logic in addition to the new tests.

WDYT?

fruch:
> @fruch , @soyacz 
> Operator has 2 separate modes for monitoring.
> So, I think about adding 2 separate functional tests to cover the monitoring case.
> It will require small rework of the existing logic in addition to the new tests.
> 
> WDYT?

Im not sure what exactly those modes are - the documention page you referred to doesn't say anything on any mode.

We should stick to what's serverlesd is using first. And have this case running regular (add it to the weekly trigger)

And other cases we can run on releases

fruch:
> > > @fruch , @soyacz
> > > Operator has 2 separate modes for monitoring.
> > > So, I think about adding 2 separate functional tests to cover the monitoring case.
> > > It will require small rework of the existing logic in addition to the new tests.
> > > WDYT?
> > 
> > 
> > Im not sure what exactly those modes are - the documention page you referred to doesn't say anything on any mode.
> > We should stick to what's serverlesd is using first. And have this case running regular (add it to the weekly trigger)
> > And other cases we can run on releases
> 
> There are `platform` and `saas` type of the monitoring deployments. Differs by the set of dashboards. This PR supports only one variant - the default one.

If so I think it's small enough difference, and we shouldn't bother too much about it

soyacz:
> @fruch , @soyacz Operator has 2 separate modes for monitoring. So, I think about adding 2 separate functional tests to cover the monitoring case. It will require small rework of the existing logic in addition to the new tests.
> 
> WDYT?

I wouldn't change this PR, better create a task for it and plan it

vponomaryov:
> > @fruch , @soyacz Operator has 2 separate modes for monitoring. So, I think about adding 2 separate functional tests to cover the monitoring case. It will require small rework of the existing logic in addition to the new tests.
> > WDYT?
> 
> I wouldn't change this PR, better create a task for it and plan it

Actually, after thinking about it more I also came to idea that the code should have been splitted apart into several commits instead of 1:
- With monitoring yaml files taken from the `scylla-operator` repo
- With the k8s utils improvement (`server-side` param). We should have it in a separate commit to keep it in case we revert the monitoring-specific parts.
- The monitoring-specific logic

Also, @fruch agreed to have a functional tests for it in a private discussion.

So, @soyacz , @fruch what do we do?

soyacz:
> > > @fruch , @soyacz Operator has 2 separate modes for monitoring. So, I think about adding 2 separate functional tests to cover the monitoring case. It will require small rework of the existing logic in addition to the new tests.
> > > WDYT?
> > 
> > 
> > I wouldn't change this PR, better create a task for it and plan it
> 
> Actually, after thinking about it more I also came to idea that the code should have been splitted apart into several commits instead of 1:
> 
> * With monitoring yaml files taken from the `scylla-operator` repo
> * With the k8s utils improvement (`server-side` param). We should have it in a separate commit to keep it in case we revert the monitoring-specific parts.
> * The monitoring-specific logic
> 
> Also, @fruch agreed to have a functional tests for it in a private discussion.
> 
> So, @soyacz , @fruch what do we do?

sure, It would be better to have it split, but I don't see big difference of reverting all of it in case we needed to. 
Besides, functional tests idea seems good.

vponomaryov:
EKS: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-eks/61/consoleFull
GKE: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-gke/27/consoleFull
Look at:
- `Deploy 'ScyllaDBMonitoring'`... which is deployment of the monitoring + SCT dashboard addon to the Grafana service.
- Regular `ClusterHealthValidatorEvent` occurrence which checks

vponomaryov:
> sure, It would be better to have it split, but I don't see big difference of reverting all of it in case we needed to.

After split we have 5 commits. And, in case we `really` will need to revert the feature, then we will need to revert, most probably, only latest 5th commit. All other may stay and be utilized by some other new code changes.

> Besides, functional tests idea seems good.

Done.

soyacz (approved):
One more thing we gained after splitting the commits is that commit messages (and git blame later) are very informative, very nice.
LGTM

fruch (approved):
LGTM

PR #740, opened 2019-01-08
Fix GCE: Timeout to create instance (120G disk)

aleksbykov:
Adding new parameter gce_job_timeout to config yaml,
which allow to control timeout gce job when create disk with
size larger than 80GB

PR #6137, opened 2023-05-15
fix(libcloud) - remove libcloud out of the code

fruch:
since we have feature we want to use that aren't supported
by libcloud for gce, we want to remove all usages of it.

### Testing

- [x] longevity - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-gce-test/38/
- [x] longevity - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-gce-test/39/
- [x] artifact - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-rocky9-test/21/
- [x] GKE - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-scylla-operator-3h-gke/12/
- [x] multi-region - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/features-google-cloud-snitch-mutli-dc-test/

### DIFF
--- a/docker/env/version
+++ b/docker/env/version
@@ -1,1 +1,1 @@
-1.42-azure-storage
+1.43-remove-libcloud
--- a/docker/env/version
+++ b/docker/env/version
@@ -1,1 +1,1 @@
-1.41-dependbot-fixes
+1.42-azure-storage
--- a/requirements.txt
+++ b/requirements.txt
@@ -16,9 +16,15 @@ apache-libcloud==3.3.1 \
     --hash=sha256:c3722c4dd58b0ee4beaf4e615e0ba82505d9a915e55319dc84383687166a3eed \
     --hash=sha256:d7450453eaf5904eb4fb4f74cf9f37dc83721a719bce34f5abb336b1a1ab974d
     # via -r ../../requirements.in
+<<<<<<< HEAD
 argus-alm==0.11.4 \
     --hash=sha256:3aa2f9a3bbdc91eef48c65180457e647a538c14b46c82d8562c4930d344e6706 \
     --hash=sha256:8d3df5ea11ca6e5857315ab3ea55f41dd5c1f2a20b50bd95c7f8d8fc15ea9bca
+=======
+argus-alm==0.10.2 \
+    --hash=sha256:6d77612672df54fefe66d35f7b6cc325a3681c3c924aa4e768ea987cbbbfb13b \
+    --hash=sha256:9dbff20d339f9e943dd1e68000a9bebc9df97bd434c3902f35ebf09a37f0eaf1
+>>>>>>> bb9e91317 (fix(manager-backup): remove libcloud usage from `mgmt_cli_test`)
     # via -r ../../requirements.in
 astroid==2.8.6 \
     --hash=sha256:5f6f75e45f15290e73b56f9dfde95b4bf96382284cde406ef4203e928335a495 \
@@ -51,6 +57,7 @@ azure-core==1.26.4 \
     # via
     #   azure-identity
     #   azure-mgmt-core
+    #   azure-storage-blob
     #   msrest
 azure-identity==1.6.1 \
     --hash=sha256:69035c81f280fac5fa9c55f87be3a359b264853727486e3568818bb43988080e \
@@ -84,6 +91,10 @@ azure-mgmt-resourcegraph==8.0.0 \
 azure-mgmt-subscription==1.0.0 \
     --hash=sha256:22f606f298419f466a8149811fc762686c93da00a7dc15d3b7cdbf22b96cf5db \
     --hash=sha256:438e1373851b8616ef4e07ae3ab50f7654e830a0cac83ff3c67d0124825f0f06
+    # via -r ../../requirements.in
+azure-storage-blob==12.16.0 \
+    --hash=sha256:43b45f19a518a5c6895632f263b3825ebc23574f25cc84b66e1630a6160e466f \
+    --hash=sha256:91bb192b2a97939c4259c72373bac0f41e30810bbc853d5184f0f45904eacafd
     # via -r ../../requirements.in
 bcrypt==4.0.1 \
     --hash=sha256:089098effa1bc35dc055366740a067a2fc76987e8ec75349eb9484061c54f535 \
@@ -123,17 +134,29 @@ botocore==1.29.49 \
     #   awscli
     #   boto3
     #   s3transfer
+<<<<<<< HEAD
 botocore-stubs==1.29.141 \
     --hash=sha256:6d62f82692b4e25456962a00028402c68ad541dd2a3619b57605959550a2a81e \
     --hash=sha256:867889a4dc3eae9a479a3720accfefa03324f5415fe199e44d996b28a1bf097b
+=======
+botocore-stubs==1.29.130 \
+    --hash=sha256:5f6f1967d23c45834858a055cbf65b66863f9f28d05f32f57bf52864a13512d9 \
+    --hash=sha256:622c4a5cd740498439008d81c5ded612146f4f0d575341c12591f978edbbe733
+>>>>>>> bb9e91317 (fix(manager-backup): remove libcloud usage from `mgmt_cli_test`)
     # via boto3-stubs
 build==0.10.0 \
     --hash=sha256:af266720050a66c893a6096a2f410989eeac74ff9a68ba194b3f6473e8e26171 \
     --hash=sha256:d5b71264afdb5951d6704482aac78de887c80691c52b88a9ad195983ca2c9269
     # via pip-tools
+<<<<<<< HEAD
 cachetools==5.3.1 \
     --hash=sha256:95ef631eeaea14ba2e36f06437f36463aac3a096799e876ee55e5cdccb102590 \
     --hash=sha256:dce83f2d9b4e1f732a8cd44af8e8fab2dbe46201467fc98b3ef8f269092bf62b
+=======
+cachetools==5.3.0 \
+    --hash=sha256:13dfddc7b8df938c21a940dfa6557ce6e94a2f1cdfa58eb90c805721d58f2c14 \
+    --hash=sha256:429e1a1e845c008ea6c85aa35d4b98b65d6a9763eeef3e37e92728a12d1de9d4
+>>>>>>> bb9e91317 (fix(manager-backup): remove libcloud usage from `mgmt_cli_test`)
     # via google-auth
 certifi==2023.5.7 \
     --hash=sha256:0f0d56dc5a6ad56fd4ba36484d6cc34451e1c6548c61daad8c320169f91eddc7 \
@@ -329,6 +352,7 @@ cryptography==40.0.2 \
     --hash=sha256:d5a1bd0e9e2031465761dfa920c16b0065ad77321d8a8c1f5ee331021fda65e9
     # via
     #   azure-identity
+    #   azure-storage-blob
     #   msal
     #   paramiko
     #   pyjwt
@@ -392,9 +416,15 @@ google-api-python-client==2.24.0 \
     --hash=sha256:1c6267c7f018d86afc62dd8ec844a3e7189022a35ee239ed704cbda010888998 \
     --hash=sha256:75edc13a5c6dd9ff6f47721162a845edc343b4987ee059842e9a7905ad62e917
     # via -r ../../requirements.in
+<<<<<<< HEAD
 google-auth==2.19.0 \
     --hash=sha256:be617bfaf77774008e9d177573f782e109188c8a64ae6e744285df5cea3e7df6 \
     --hash=sha256:f39d528077ac540793dd3c22a8706178f157642a67d874db25c640b7fead277e
+=======
+google-auth==2.18.0 \
+    --hash=sha256:c66b488a8b005b23ccb97b1198b6cece516c91869091ac5b7c267422db2733c7 \
+    --hash=sha256:ef3f3a67fa54d421a1c155864570f9a8de9179cedc937bda496b7a8ca338e936
+>>>>>>> bb9e91317 (fix(manager-backup): remove libcloud usage from `mgmt_cli_test`)
     # via
     #   google-api-core
     #   google-api-python-client
@@ -587,7 +617,9 @@ invoke==1.6.0 \
 isodate==0.6.1 \
     --hash=sha256:0751eece944162659049d35f4f549ed815792b38793f07cf73381c1c87cbed96 \
     --hash=sha256:48c5881de7e8b0a0d648cb024c8062dc84e7b840ed81e864c7614fd3c127bde9
-    # via msrest
+    # via
+    #   azure-storage-blob
+    #   msrest
 isort==5.12.0 \
     --hash=sha256:8bef7dde241278824a6d83f44a544709b065191b95b6e50894bdc722fcba0504 \
     --hash=sha256:f84c2818376e66cf843d497486ea8fed8700b340f308f076c6fb1229dff318b6
@@ -748,9 +780,15 @@ mypy-boto3-dynamodb==1.26.115 \
     --hash=sha256:280d4b71b716e221887cfbd8aeb27699efb2ed5f5ac98621878dbe99492dcd8c \
     --hash=sha256:b94d69617d118421d625120d38b81772ef3aa2ba2d98e771008f9c72388f60f1
     # via boto3-stubs
+<<<<<<< HEAD
 mypy-boto3-ec2==1.26.136 \
     --hash=sha256:51be9f2188802b0e27e9f85f7bdf773dbb93859ed4fd99e71738710e53b4d78f \
     --hash=sha256:d02842ffc7504d7a0f7728652b2321729a4b8635251c68ae4692e52a8df40b02
+=======
+mypy-boto3-ec2==1.26.129 \
+    --hash=sha256:b0ea692c77d9bf5a8a3a127d06e5fc0278e1adaad0a1bc338e9ff0a80547637c \
+    --hash=sha256:cb30dbe6bd4fed7060a03c925bac6c46c5d5880c2928fadffe3f3cb7881bcbd2
+>>>>>>> bb9e91317 (fix(manager-backup): remove libcloud usage from `mgmt_cli_test`)
     # via boto3-stubs
 mypy-boto3-pricing==1.26.82 \
     --hash=sha256:47b9f754fb3bca54f84eb3948343b706276ee2e5a599d46a9b55128f36463227 \
@@ -791,6 +829,7 @@ ordered-set==4.1.0 \
     --hash=sha256:046e1132c71fcf3330438a539928932caf51ddbc582496833e23de611de14562 \
     --hash=sha256:694a8e44c87657c59292ede72891eb91d34131f6531463aab3009191c77364a8
     # via deepdiff
+<<<<<<< HEAD
 orjson==3.8.14 \
     --hash=sha256:01640ab79111dd97515cba9fab7c66cb3b0967b0892cc74756a801ff681a01b6 \
     --hash=sha256:017de5ba22e58dfa6f41914f5edb8cd052d23f171000684c26b2d2ab219db31e \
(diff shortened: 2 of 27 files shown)


sdcm/cluster_k8s/gke.py
```diff
@@ -400,7 +407,9 @@ def nodes_dest_ip(self) -> str:
 
     @property
     def k8s_node(self):
-        return self.parent_cluster.k8s_cluster.gce_services[0].ex_get_node(name=self.node_name)
+        return self.parent_cluster.k8s_cluster.gce_services.get(project=self.parent_cluster.k8s_cluster.gce_project,
```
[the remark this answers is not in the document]
fruch: yes  it's on my todo list for this one

sdcm/logcollector.py
```diff
@@ -1417,44 +1418,48 @@ def get_aws_instances_by_testid(self):
                 global_ip=self.get_aws_ip_address(instance),
                 tags={**self.tags, "NodeType": "loader", }))
 
+    @staticmethod
```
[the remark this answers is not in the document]
fruch: it's not a property, echo instance passed to it has different ip address..

sdcm/tester.py
```diff
@@ -1038,7 +1036,7 @@ def get_cluster_gce(self, loader_info, db_info, monitor_info):
                                                gce_image_size=db_info['disk_size'],
                                                gce_n_local_ssd=db_info['n_local_ssd'],
                                                gce_instance_type=db_info['type'],
-                                               services=services,
+                                               services=service,
```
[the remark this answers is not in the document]
fruch: yes we can rename, I was first more focused on it working.

didn't tested multi-dc yet, at the end GCE works with zones, and all the APIs need project and zone, not a diffrent client like in libcloud or with boto3.
so if we get the zone correct the rest should work with same client object.

sdcm/tester.py
```diff
@@ -1449,22 +1447,21 @@ def get_cluster_k8s_gke(self):
...
 
-        services, gce_datacenter = list(services.values()), list(services.keys())
         self.k8s_cluster = gke.GkeCluster(
             gke_cluster_version=self.params.get("gke_cluster_version"),
             gke_k8s_release_channel=self.params.get("gke_k8s_release_channel"),
             gce_disk_size=self.params.get("root_disk_size_db"),
             gce_disk_type=self.params.get("gce_root_disk_type_db"),
             gce_network=self.params.get("gce_network"),
-            services=services,
+            services=service,
```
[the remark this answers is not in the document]
fruch: not yet

sdcm/cluster_gce.py
```diff
@@ -308,17 +343,25 @@ def to_short_name(value):
...
-            gce_job_default_timeout = self._gce_services[dc_idx].connection.timeout
-            self._gce_services[dc_idx].connection.timeout = gce_job_default_timeout * 3
-            self.log.info("Job complete timeout is set to %ss" %
-                          self._gce_services[dc_idx].connection.timeout)
+        #gce_job_default_timeout = None
+        # if self._gce_image_size and int(self._gce_image_size) > 80:
+        #    gce_job_default_timeout = self._gce_services .connection.timeout
+        #    self._gce_services[dc_idx].connection.timeout = gce_job_default_timeout * 3
+        #    self.log.info("Job complete timeout is set to %ss" %
+        #                  self._gce_services[dc_idx].connection.timeout)
```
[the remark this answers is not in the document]
fruch: I've removed this

this was introduced 5 years ago
https://github.com/scylladb/scylla-cluster-tests/pull/740

we don't have any configuration with 120Gb or 80Gb root disks anymore, I don't think it's needed

sdcm/tester.py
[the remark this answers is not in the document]
fruch: I've renamed it to be `gce_service` across the board

sdcm/utils/cloud_monitor/resources/static_ips.py
```diff
@@ -43,13 +43,13 @@ def __init__(self, eip, region):
...
-    def __init__(self, static_ip: GCEAddress):
-        used_by = static_ip.extra.get("users")
+    def __init__(self, static_ip: GceAddress):
+        used_by = static_ip.users
         super().__init__(
             cloud="gce",
             name=static_ip.name,
             address=static_ip.address,
-            region=static_ip.region if isinstance(static_ip.region, str) else static_ip.region.name,
+            region=static_ip.region if isinstance(static_ip.region, str) else static_ip.region,
```
vponomaryov, line 52: The condition here is useless because we have identical values on both sides.
Please either remove the condition or fix one of the values to be unique.

sdcm/utils/common.py
```diff
@@ -58,11 +58,9 @@
 from mypy_boto3_ec2 import EC2Client, EC2ServiceResource
 from mypy_boto3_ec2.service_resource import Image as EC2Image
 import docker  # pylint: disable=wrong-import-order; false warning because of docker import (local file vs. package)
-import libcloud.storage.providers
-import libcloud.storage.types
-from libcloud.compute.base import Node as GCENode, NodeImage as GCEImage
-from libcloud.compute.providers import get_driver
-from libcloud.compute.types import Provider
+from google.cloud.storage import Blob
```
vponomaryov, line 61: Since it is located in the `common` module I think it is better to rename `Blob` to something more specific.
For example:
```from google.cloud.storage import Blob as GceStorageBlob```

File: sct.py [hunk 8]
```diff
@@ -97,7 +97,7 @@
     send_perf_email
 from sdcm.parallel_timeline_report.generate_pt_report import ParallelTimelinesReportGenerator
 from sdcm.utils.aws_utils import AwsArchType
-from sdcm.utils.gce_utils import SUPPORTED_PROJECTS
+from sdcm.utils.gce_utils import SUPPORTED_PROJECTS, gce_public_addresses
```
vponomaryov, line 100: It is update of what is done in the previous commit of this PR.
So, would be really good to reduce such unneeded changes in scope of the single PR.

sdcm/utils/gce_utils.py
```diff
@@ -271,3 +262,245 @@ def _get_log_entries(self, service, query, page_token=None):
...
+    else:
+        instance.machine_type = f"zones/{zone}/machineTypes/{machine_type}"
+
+    if accelerators:
+        instance.guest_accelerators = accelerators
+
+    if preemptible:
+        # Set the preemptible setting
+        warnings.warn(
+            "Preemptible VMs are being replaced by Spot VMs.", DeprecationWarning
```
vponomaryov, line 458: As I understand the logic, the message must be different.
It must be something like following:
```'preemptible' option is deprecated, use 'spot' instead.```

The current message, not seeing code, makes me think that some is in process of replacement right now.
Which is false.

Then, since we add new code, why adding the deprecated option in general?

**Review by vponomaryov — DISMISSED** [review]

File: sdcm/utils/gce_utils.py (hunk 9 above)
Comment (fruch) [reply sdcm/utils/gce_utils.py:458]: it's actually copy paste form google own code sample, it mean the feature "Preemptible VMs" is replace with "Spot VMs"

anyhow we need this code since we now use preemptible, and we want to switch to using spots, we can ditch this warning.
it doesn't really helps us.

File: sct.py (hunk 8 above)
Comment (fruch) [reply sct.py:100]: it just a mistake, a fix that was apply into the wrong commit.
I'll move it into the right place

File: sdcm/utils/cloud_monitor/resources/static_ips.py (hunk 6 above)
Comment (fruch) [reply sdcm/utils/cloud_monitor/resources/static_ips.py:52]: removed the condition, it's really not needed anymore

File: sdcm/utils/common.py (hunk 7 above)
Comment (fruch) [reply sdcm/utils/common.py:61]: changed it (even this is was used only to get tab compilation on the blob)

File: sct.py (hunk 8 above)
Comment (fruch) [reply sct.py:100]: fixed

File: sdcm/utils/gce_utils.py (hunk 9 above)
Comment (fruch) [reply sdcm/utils/gce_utils.py]: I've removed the misleading warning

**Comment by fruch** [comment]
multi-dc was very slow on bootstrap of new nodes (~20m vs. ~1h), apparently someone forgot to pass the instance/machine type to the create_instance API... so everything was running with "n1-standard-1"... 🤦🏼

**Comment by fruch** [comment]
@vponomaryov @soyacz this one is ready to final review, now that affect jobs passed

File: docker/env/version [hunk 10]
```diff
@@ -1 +1 @@
-1.39-azure-storage
+1.39-remove-libcloud
```
vponomaryov, line 1: Please, bump also integer part. Looking at the image tags not knowing context/history it is unclear which one is newer.

Jenkinsfile
```diff
@@ -71,7 +71,7 @@ pipeline {
         }
     }
     parameters {
-        string(defaultValue: "5.0.5",
+        string(defaultValue: "5.2.0",
```
vponomaryov, line 74: This change could be done in a separate PR and be merged much faster.

sdcm/utils/common.py
```diff
@@ -58,7 +58,7 @@
 from mypy_boto3_ec2 import EC2Client, EC2ServiceResource
 from mypy_boto3_ec2.service_resource import Image as EC2Image
 import docker  # pylint: disable=wrong-import-order; false warning because of docker import (local file vs. package)
-from google.cloud.storage import Blob
+from google.cloud.storage import Blob as AzureStorageBlob
```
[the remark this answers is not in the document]
fruch: I confused it to be part of the manager changes commit
fixed and put into place

docker/env/version
fruch: changed the version number

I was neglecting it, since I have two commit creating new docker image in the same PR.

Jenkinsfile
fruch: moved to https://github.com/scylladb/scylla-cluster-tests/pull/6179

sdcm/cluster_k8s/gke.py
```diff
@@ -400,7 +407,9 @@ def nodes_dest_ip(self) -> str:
 
     @property
     def k8s_node(self):
-        return self.parent_cluster.k8s_cluster.gce_services[0].ex_get_node(name=self.node_name)
+        return self.parent_cluster.k8s_cluster.gce_service.get(project=self.parent_cluster.k8s_cluster.gce_project,
```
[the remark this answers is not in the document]
fruch: fixed

sdcm/cluster_gce.py
```diff
@@ -105,22 +130,29 @@ def _instance_wait_safe(self, instance_method: Callable[P, R], *args: P.args, **
...
         self._instance = instance
-        ip_tuple = (instance.public_ips, instance.private_ips)
+
+        ip_tuple = (list(gce_public_addresses(instance)), list(gce_private_addresses(instance)))
         return ip_tuple
 
     @property
     def region(self):
-        return self._gce_service.region.name
+        return self._instance.zone.split('/')[-1][:2]
```
soyacz, line 144: region is only first 2 characters

sdcm/cluster_gce.py
```diff
@@ -152,16 +184,18 @@ def check_spot_termination(self):
     def restart(self):
         # When using local_ssd disks in GCE, there is no option to Stop and Start an instance.
         # So, for now we will keep restart the same as hard reboot.
-        self._instance_wait_safe(self._instance.reboot)
+        self._instance_wait_safe(self._gce_service.reset, instance=self._instance.name,
```
soyacz, line 187: shouln't we drop `_instance_wait_safe` in favor of `wait_for_extended_operation` (this one is used also for `.reset` method in GCE docs)
same for `hard_reboot`

sdcm/cluster_gce.py
```diff
@@ -152,16 +184,18 @@ def check_spot_termination(self):
...
+                                 project=self.project, zone=self.zone)
 
     def _safe_destroy(self):
         try:
-            self._gce_service.ex_get_node(self.instance_name)
-            self._instance.destroy()
-        except ResourceNotFoundError:
+            self._gce_service.delete(instance=self._instance.name,
+                                     project=self.project, zone=self.zone)
```
soyacz, line 197: let's use here `wait_for_extended_operation` too

sdcm/cluster_gce.py
```diff
@@ -308,17 +343,16 @@ def to_short_name(value):
...
-        # if size of disk is larget than 80G, then
-        # change the timeout of job completion to default * 3.
+            request = compute_v1.InstancesSetLabelsRequest()
+            request.labels = _instance.labels
+            request.label_fingerprint = _instance.label_fingerprint
+            request.labels.update(normalized_tags)
+
+            self._gce_service.set_labels(project=self.project, zone=self._gce_zone_names[dc_idx],
+                                         instance=_instance.name,
+                                         instances_set_labels_request_resource=request)
```
soyacz, line 354: this code block is repeated multiple times - is it feasible to extract/reuse one method/function?

sdcm/utils/gce_utils.py
```diff
@@ -47,34 +47,58 @@
...
     availability_zones = SUPPORTED_REGIONS.get(region, None)
     if not availability_zones:
         raise Exception(f'Unsupported region: {region}')
-    return f"{region}-{random.choice(availability_zones)}"
+    return f"{random.choice(availability_zones)}"
+
+
+def get_gce_compute_instances_client() -> tuple[compute_v1.InstancesClient, dict]:
+    info = KeyStore().get_gcp_credentials()
+    credentials = service_account.Credentials.from_service_account_info(info)
```
soyacz, line 59: this function takes few sec's to execute and is used multiple times, maybe cache `get_gcp_credentials()` method?

soyacz (changes requested):
Several comments.
I wonder if we shouldn't drop `_instance_wait_safe` in favor of `wait_for_extended_operation` everywhere.

sdcm/cluster_gce.py
fruch: you are correct should be `[:-2]`

fruch: fixed

sdcm/cluster_gce.py
fruch: `_instance_wait_safe` is doing exponential backoff
`wait_for_extended_operation` is very specific and can working with only some api calls, and it won't every retry, it only for way for specific operation to finish.

so I'm not sure I'm ready to remove the those retries just yet, since I don't know the reason they were introduced for.

sdcm/cluster_gce.py
fruch: refactored it into `gce_utils.gce_set_labels`

fruch:
@soyacz @vponomaryov can you give it a final review ?

sdcm/utils/gce_utils.py
```diff
@@ -271,3 +261,275 @@ def _get_log_entries(self, service, query, page_token=None):
...
+         disk_type: the type of disk you want to create. This value uses the following format:
+            "zones/{zone}/diskTypes/(pd-standard|pd-ssd|pd-balanced|pd-extreme)".
+            For example: "zones/us-west3-b/diskTypes/pd-ssd"
+        disk_size_gb: size of the new disk in gigabytes
+        boot: boolean flag indicating whether this disk should be used as a boot disk of an instance
+        source_image: source image to use when creating this disk. You must have read access to this disk. This can be one
+            of the publicly available images or an image from one of your projects.
+            This value uses the following format: "projects/{project_name}/global/images/{image_name}"
+        auto_delete: boolean flag indicating whether this disk should be deleted with the VM that uses it
+
```
vponomaryov, line 332: The most unclear `type_` parameter is not described here.

sdcm/cluster_k8s/gke.py
```diff
@@ -191,11 +198,11 @@ def __init__(self,
         self.gce_disk_type = gce_disk_type
         self.gce_disk_size = gce_disk_size
         self.gce_network = gce_network
-        self.gce_services = services
+        self.gce_service, info = gce_service
```
vponomaryov, line 201: If the `info` is not used then shouldn't it be `_`? How does the `Pylint` not complain about it?

requirements.txt
```diff
@@ -12,19 +12,9 @@ anyconfig==0.12.0 \
     --hash=sha256:2119065ad835d0e4cbcb26a58fab0705952c5657c05fa90ef7a11cdecbbafb1f \
     --hash=sha256:3091d7675740686fade85755537f2a7c6ccefa8659c5bb56174e02ded74b96d5
     # via -r ../../requirements.in
-apache-libcloud==3.3.1 \
-    --hash=sha256:c3722c4dd58b0ee4beaf4e615e0ba82505d9a915e55319dc84383687166a3eed \
-    --hash=sha256:d7450453eaf5904eb4fb4f74cf9f37dc83721a719bce34f5abb336b1a1ab974d
-    # via -r ../../requirements.in
-<<<<<<< HEAD
```
vponomaryov: We should not have any commit with broken syntax. So, need to fix all these merge conflicts in the second commit of this PR.

sdcm/utils/gce_utils.py
```diff
@@ -60,23 +61,59 @@ def append_zone(region: str) -> str:
...
-
 def get_gce_service(region: str) -> GceDriver:
     credentials = KeyStore().get_gcp_credentials()
     return _get_gce_service(credentials, append_zone(region))
 
 
+def gce_public_addresses(instance: compute_v1.Instance):
+    for interface in instance.network_interfaces:
+        for config in interface.access_configs:
+            yield config.nat_i_p
```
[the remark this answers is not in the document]
fruch: changed it to return a list

requirements.txt
```diff
@@ -16,9 +16,15 @@ apache-libcloud==3.3.1 \
...
     # via -r ../../requirements.in
+<<<<<<< HEAD
 argus-alm==0.11.4 \
     --hash=sha256:3aa2f9a3bbdc91eef48c65180457e647a538c14b46c82d8562c4930d344e6706 \
     --hash=sha256:8d3df5ea11ca6e5857315ab3ea55f41dd5c1f2a20b50bd95c7f8d8fc15ea9bca
+=======
+argus-alm==0.10.2 \
+    --hash=sha256:6d77612672df54fefe66d35f7b6cc325a3681c3c924aa4e768ea987cbbbfb13b \
+    --hash=sha256:9dbff20d339f9e943dd1e68000a9bebc9df97bd434c3902f35ebf09a37f0eaf1
+>>>>>>> bb9e91317 (fix(manager-backup): remove libcloud usage from `mgmt_cli_test`)
```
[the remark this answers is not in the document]
fruch: It's my mistake, I didn't rebuild the intermediate docker image, only the last one.

I've fixed this, and built/push the image

sdcm/utils/gce_utils.py
fruch: added docstring for this parameter as well

sdcm/cluster_k8s/gke.py
fruch: it is used few line below

sdcm/utils/gce_utils.py
```diff
@@ -271,3 +261,275 @@ def _get_log_entries(self, service, query, page_token=None):
...
+
+
+def disk_from_image(  # pylint: disable=too-many-arguments
+    disk_type: str,
+    boot: bool,
+    disk_size_gb: int = None,
+    source_image: str = None,
+    auto_delete: bool = True,
+    device_name: str = None,
+    type_: str = None,
```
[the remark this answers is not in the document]
fruch: added docstring for this parameter as well

requirements.txt
fruch: fixed

sdcm/sct_runner.py
```diff
@@ -818,7 +818,7 @@ def list_sct_runners(cls, verbose: bool = True) -> list[SctRunnerInfo]:
                 region_az=region,
                 instance=instance,
                 instance_name=instance.name,
-                public_ips=instance.public_ips,
+                public_ips=instance.network_interfaces[0].access_configs[0].nat_i_p,
```
soyacz, line 828: maybe reuse?: `gce_public_addresses(instance)[0]`

sdcm/cluster_gce.py
soyacz: hm?

sdcm/cluster_k8s/gke.py
```diff
@@ -493,17 +507,17 @@ def _get_instances(self, dc_idx):
             tags_dict={'MonitorId': self.monitor_id, 'NodeType': self.node_type})
         instances_by_zone = self._get_instances_by_prefix(dc_idx)
         instances = []
-        attr_name = 'public_ips' if self._node_public_ips else 'private_ips'
+        ip_generator = gce_public_addresses if self._node_public_ips else gce_private_addresses
         for node_zone in instances_by_zone:
             # Filter nodes by zone and by ip addresses
-            if not getattr(node_zone, attr_name):
+            if not list(ip_generator(node_zone)):
```
soyacz, line 513: nit: no need to `list` it anymore

sdcm/utils/gce_utils.py
```diff
@@ -234,3 +266,278 @@ def _get_log_entries(self, service, query, page_token=None):
...
```
soyacz, line 323: in cases it accepts specified set of strings, better use `Literal["PERSISTENT", "SCRATCH"]`

soyacz (dismissed):
Besides small comments. LGTM.

sdcm/sct_runner.py
fruch: fixed

sdcm/cluster_k8s/gke.py
fruch: fixed

sdcm/utils/gce_utils.py
fruch: added the typing

sdcm/cluster_gce.py
fruch: add `wait_for_extended_operation` here as wall

Issue #6144, opened 2023-05-17
[unittest] TestMBM.test_execute_mbm_from_cli once in a while with no clear reason


fruch:
```
17:38:23  ______________________ TestMBM.test_execute_mbm_from_cli _______________________
17:38:23  
17:38:23  self = <unit_tests.test_microbenchmarking.TestMBM testMethod=test_execute_mbm_from_cli>
17:38:23  
17:38:23      def test_execute_mbm_from_cli(self):
17:38:23          result_path = os.path.join(os.path.dirname(__file__), 'test_data/test_microbenchmarking/PFF_with_AVGAIO')
17:38:23          html_report = tempfile.mkstemp(suffix=".html", prefix="microbenchmarking-")[1]
17:38:23          with subprocess.Popen(['./microbenchmarking.py', 'check',
17:38:23                                 '--results-path', result_path,
17:38:23                                 '--email-recipients', 'alex.bykov@scylladb.com',
17:38:23                                 '--report-path', html_report,
17:38:23                                 '--hostname', self.mbra.hostname],
17:38:23                                cwd=self.cwd,
17:38:23                                stdout=subprocess.PIPE,
17:38:23                                stderr=subprocess.PIPE) as ps:
17:38:23              stdout, _ = ps.communicate()
17:38:23              print(_)
17:38:23              print(stdout)
17:38:23  >           self.assertEqual(ps.returncode, 0)
17:38:23  E           AssertionError: 1 != 0
17:38:23  
17:38:23  unit_tests/test_microbenchmarking.py:281: AssertionError
17:38:23  ----------------------------- Captured stdout call -----------------------------
[...]

PR #6147, opened 2023-05-18
fix(microbenchmarking): pass es index as cli parameter

aleksbykov:
After changing es index for microbenchmarking test resutls, unit tests start failing. Unit test try to get response from file and if not it try to get data from ES. But after changing ES index return results are empty.

Add new cli parameter for microbenchmarking to pass ES index name to work with.

unit tests are passed

Fixes: #6144

### DIFF
--- a/unit_tests/test_microbenchmarking.py
+++ b/unit_tests/test_microbenchmarking.py
@@ -65,10 +65,12 @@ class MicroBenchmarkingResultsAnalyzerMock(MicroBenchmarkingResultsAnalyzer):
 
 class TestMBM(unittest.TestCase):  # pylint: disable=too-many-public-methods
     def setUp(self):
-        self.mbra = MicroBenchmarkingResultsAnalyzerMock(email_recipients=('alex.bykov@scylladb.com', ))
+        self.mbra = MicroBenchmarkingResultsAnalyzerMock(email_recipients=(
+            'alex.bykov@scylladb.com', ), es_index="microbenchmarking")
         self.mbra.hostname = 'godzilla.cloudius-systems.com'
         self.cwd = os.path.join(os.path.dirname(__file__), '..', 'sdcm')
         self.mbra.test_run_date = "2019-06-27_11:39:40"
+        self.es_index = "microbenchmarking"
 
     @staticmethod
     def get_result_obj(path):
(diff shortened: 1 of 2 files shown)


aleksbykov:
fix #6144

unit_tests/test_microbenchmarking.py
```diff
@@ -267,11 +269,18 @@ def test_help_parameter_for_check(self):
     def test_execute_mbm_from_cli(self):
         result_path = os.path.join(os.path.dirname(__file__), 'test_data/test_microbenchmarking/PFF_with_AVGAIO')
         html_report = tempfile.mkstemp(suffix=".html", prefix="microbenchmarking-")[1]
+        LOGGER.info(['./microbenchmarking.py', 'check',
```
fruch, line 272: just put the arguments in a list, not much point duplicating it.

unit_tests/test_microbenchmarking.py
```diff
@@ -65,10 +65,12 @@ def _get_hash(*args, **kwargs):
 
 class TestMBM(unittest.TestCase):  # pylint: disable=too-many-public-methods
     def setUp(self):
-        self.mbra = MicroBenchmarkingResultsAnalyzerMock(email_recipients=('alex.bykov@scylladb.com', ))
+        self.mbra = MicroBenchmarkingResultsAnalyzerMock(email_recipients=(
+            'alex.bykov@scylladb.com', ), es_index="microbenchmarking")
         self.mbra.hostname = 'godzilla.cloudius-systems.com'
         self.cwd = os.path.join(os.path.dirname(__file__), '..', 'sdcm')
         self.mbra.test_run_date = "2019-06-27_11:39:40"
+        self.es_index = "microbenchmarking"
```
fruch, line 73: don't we want some coverage that the new index also works ?

aleksbykov: Once we have more results in it and old microbenchmark will be removed, we switch to new index also unittests. 
Logic was not changed at all how microbenchmark get data, only index name

fruch: I would recommend open a task for it, not forgetting that unittest should be part of removal of the old index

fruch (approved):
LGTM

PR #6156, opened 2023-05-18
fix(add_remove_dc): wait for rebuild

soyacz:
There are situations where ssh connection is dropped while rebuild. This leads to start decommission while rebuild is in progress (e.g. in `add_remove_dc`) - failing decommission and leaving dangling db node.

Workaround ssh issue by explicitly waiting for rebuild to finish, so decommission can be triggered.

To achieve that, new context manager `wait_for_log_lines` was created, where provided db node and regex patterns to wait for (start regex and end pattern can be provided separately). See unit tests for examples.

refs: scylladb/scylladb#14004

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -113,7 +113,7 @@ from sdcm.utils.sstable.load_utils import SstableLoadUtils
 from sdcm.utils.toppartition_util import NewApiTopPartitionCmd, OldApiTopPartitionCmd
 from sdcm.utils.version_utils import MethodVersionNotFound, scylla_versions
 from sdcm.utils.raft import Group0MembersNotConsistentWithTokenRingMembersException
-from sdcm.wait import wait_for
+from sdcm.wait import wait_for, wait_for_log_lines
 from test_lib.compaction import CompactionStrategy, get_compaction_strategy, get_compaction_random_additional_params, \
     get_gc_mode, GcMode
 from test_lib.cql_types import CQLTypeBuilder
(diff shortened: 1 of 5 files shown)


sdcm/wait.py
```diff
@@ -124,3 +125,26 @@ def retry_logger(retry_state: tenacity.RetryCallState) -> None:
...
+def wait_for_log_lines(node, start_line_patterns, end_line_patterns, start_timeout=60, end_timeout=120):
+    """Waits for given lines patterns to appear in node logs despite exception raised"""
+    start_follower = node.follow_system_log(patterns=start_line_patterns)
+    end_follower = node.follow_system_log(patterns=end_line_patterns)
+    start_time = time.time()
+    try:
+        yield
+    finally:
+        started = any(start_follower)
+        while not started and (time.time() - start_time < start_timeout):
```
fruch, line 140: I'm not sure I under stand this logic
if it's the finally clause why we need a loop like this for the start message ?

sdcm/wait.py
```diff
@@ -124,3 +125,26 @@ def retry_logger(retry_state: tenacity.RetryCallState) -> None:
...
+
+
+@contextmanager
+def wait_for_log_lines(node, start_line_patterns, end_line_patterns, start_timeout=60, end_timeout=120):
+    """Waits for given lines patterns to appear in node logs despite exception raised"""
+    start_follower = node.follow_system_log(patterns=start_line_patterns)
+    end_follower = node.follow_system_log(patterns=end_line_patterns)
+    start_time = time.time()
+    try:
+        yield
```
fruch, line 137: we need to test how it looks when exception is being raised, i.e. the case we are timing out or failing.
do we still see the errors despite of the other exception being raise ?

sdcm/nemesis.py
```diff
@@ -4020,7 +4020,10 @@ def disrupt_add_remove_dc(self) -> None:
                     strategy.replication_factors.update({new_dc_name: 1})
                     replication_strategy_setter(**{keyspace: strategy})
                 InfoEvent(message='execute rebuild on new datacenter').publish()
-                new_node.run_nodetool(sub_cmd=f"rebuild -- {datacenters[0]}", retry=0)
+                with wait_for_log_lines(node=new_node, start_line_patterns=["rebuild.*started with keyspaces="],
```
fruch, line 4023: need to think/see how this fits with the `adaptive_timeout` or soft timeout

sdcm/wait.py
soyacz: When running `finally` quickly after command start we may still not get the starting line, so we need to wait. Besides in case this line does not show up, we may want to fail faster than waiting for full timeout.

sdcm/wait.py
soyacz: sure, I'll add unit-test for this

sdcm/nemesis.py
soyacz: still we we should be able to use both context managers I think - and `adaptive_timeout` yielded value could be used here in `end_timeout`.

fruch (approved):
LGTM

PR #6158, opened 2023-05-19
improvement(Argus): Catch failed runs earlier

temichus:
Currently, we report to Argus when the test successfully started but we need to find a way to track the status of
the run even before (for example we encountered an issue that Releng didn't provide AMI id and the run failed but in Argus, we see that the test never ran)
commands created: "create-argus-test-run", "finish-argus-test-run" and run before the sct runner created and after Clean SCT Runners.

SCT code still supports old pipelines where   Argus test run created and finished in SCT


docker/env/version
```diff
@@ -1 +1 @@
-1.38-argus-0.10.2-update
+1.38-argus-0.11.1-update
```
fruch, line 1: Update the number at the beginning, the fact argurs version number is part of this name is just cause the last update was related to argus, and we are lazy to rebuild the docker image before merges.

vars/createArgusTestRun.groovy
```diff
@@ -0,0 +1,19 @@
...
+    def test_config = groovy.json.JsonOutput.toJson(params.test_config)
+
+    sh """#!/bin/bash
+        set -xe
+
+        echo "Creating Argus test run ..."
+
+        export SCT_CLUSTER_BACKEND="${params.backend}"
+        export SCT_CONFIG_FILES=${test_config}
+        export SCT_COLLECT_LOGS=false
```
fruch, line 13: sound irrelevant to this call

vars/longevityPipeline.groovy
```diff
@@ -315,6 +330,21 @@ def call(Map pipelineParams) {
                     }
                 }
             }
+            stage('Finish Argus Test Run') {
```
fruch, line 333: we'll need this stage on all pipelines, not just this one...

artifacts / rolling upgrades / manager / operator. and on siren-test as well.

docs/configuration_options.md
```diff
@@ -329,6 +329,7 @@
...
 | **<a href="#user-content-stress_image" name="stress_image">stress_image</a>**  | Dict of the images to use for the stress tools | {'ndbench': 'scylladb/hydra-loaders:ndbench-jdk8-20210720', 'ycsb': 'scylladb/hydra-loaders:ycsb-jdk8-20220918', 'nosqlbench': 'scylladb/hydra-loaders:nosqlbench-4.15.49', 'cassandra-stress': '', 'scylla-bench': 'scylladb/hydra-loaders:scylla-bench-v0.1.17', 'gemini': 'scylladb/hydra-loaders:gemini-1.7.7', 'alternator-dns': 'scylladb/hydra-loaders:alternator-dns-0.1', 'cdc-stresser': 'scylladb/hydra-loaders:cdc-stresser-20210630', 'kcl': 'scylladb/hydra-loaders:kcl-jdk8-20210526-ShardSyncStrategyType-PERIODIC', 'harry': 'scylladb/hydra-loaders:cassandra-harry-jdk11-20220816'} | SCT_STRESS_IMAGE
 | **<a href="#user-content-enable_argus" name="enable_argus">enable_argus</a>**  | Control reporting to argus | True | SCT_ENABLE_ARGUS
+| **<a href="#user-content-argus_test_id" name="argus_test_id">argus_test_id</a>**  | Reuse existing argus test run id | N/A | SCT_ARGUS_TEST_ID
```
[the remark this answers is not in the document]
fruch: I'm lost... I didn't understand what this is ID is used for, and for what do we need it ?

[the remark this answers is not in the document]
soyacz: why we can't create SCT_TEST_ID in the first jenkins step and reuse it later in every step?

sct.py
```diff
@@ -1668,6 +1672,67 @@ def configure_jenkins_builders(cloud_provider, regions):
             raise NotImplementedError("configure_jenkins_builders doesn't support Azure yet")
 
 
+@cli.command("create-argus-test-run", help="Initialize an argus test run.")
+def create_test_run():
+    try:
+        argus_test_id_path = Path("argus_test_id")
```
vponomaryov, line 1678: We have possibility to run 2+ different tests on a single SCT runner:
1) Performance throughput CI job defined at `jenkins-pipelines/operator/performance/perf-regression-throughput-eks.jenkinsfile` (3 parallel tests).
2) CI tests which we can trigger with labels in `github`. Example: https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-6162/9/console (5 parallel tests)

As a result, if I understand the new logic correctly then it may cause mess in the above mentioned cases.

So, need to use a `test logdir` the same way as the `TestConfig` does it storing there it's `test_id` file.

sdcm/sct_config.py
```diff
@@ -1420,6 +1420,9 @@ class SCTConfiguration(dict):
         dict(name="enable_argus", env="SCT_ENABLE_ARGUS", type=boolean,
              help="Control reporting to argus"),
 
+        dict(name="argus_test_id", env="SCT_ARGUS_TEST_ID", type=str,
+             help="Reuse existing argus test run id"),
```
vponomaryov, line 1424: I think that here must be added more details because it is expected to be set by jenkins and not by user.

sdcm/tester.py
```diff
@@ -378,16 +378,20 @@ def __init__(self, *args, **kwargs):  # pylint: disable=too-many-statements,too-
...
-                job_name=get_job_name(),
-                job_url=get_job_url(),
-                started_by=get_username(),
-                commit_id=get_git_commit_id(),
-                runner_public_ip=get_sct_runner_ip(),
-                runner_private_ip=get_my_ip(),
-                sct_config=self.params,
-            )
-            self.log.info("Initialized Argus TestRun with test id %s", self.test_config.argus_client().run_id)
+            if self.params.get('argus_test_id') != self.test_config.test_id():
```
vponomaryov, line 381: Shouldn't we also update the `self.params.get('argus_test_id')` value for case we call this method second+ time?

Also, cannot we just call Argus API to get to know whether there is an existing Argus test run with this ID or not?
If we can, won't it drop the need to introduce the new config option?

docs/configuration_options.md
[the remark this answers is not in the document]
vponomaryov: The idea that gets solved by this new option is not configuration but `is created` check.
By the logic of this PR it is expected to be set automatically and only to a single value.
So, I think it is bad idea to to do so.
As I wrote below in the `sdcm/tester.py` file comment:
```
cannot we just call Argus API to get to know whether there is an existing Argus test run with this ID or not?
```

vponomaryov (changes requested):

sct.py
fruch: Regarding sharing of runners, all those examples are not sharing runners, and they do we need to fix that.

I do agree we should use the test dir, the tricky part, is that test dir is created when the test starts, so you can put anything in it before that, you'll need something equivalent, that can be found at later stages while not assuming only one test is running at a runner.

vars/createArgusTestRun.groovy
```diff
@@ -0,0 +1,18 @@
...
+def call(Map params) {
+    def test_config = groovy.json.JsonOutput.toJson(params.test_config)
+
+    sh """#!/bin/bash
+        set -xe
+
+        echo "Creating Argus test run ..."
+
+        export SCT_CLUSTER_BACKEND="${params.backend}"
+        export SCT_CONFIG_FILES=${test_config}
```
k0machi, line 12: better to quote the variable `"${test_config}"`

vars/finishArgusTestRun.groovy
```diff
@@ -0,0 +1,20 @@
...
+def call(Map params, RunWrapper currentBuild) {
+    def test_config = groovy.json.JsonOutput.toJson(params.test_config)
+    def test_status = currentBuild.currentResult
+
+    sh """#!/bin/bash
+    set -xe
+    echo "Finishing Argus test run ..."
+
+    export SCT_CLUSTER_BACKEND="${params.backend}"
+    export SCT_CONFIG_FILES=${test_config}
```
k0machi, line 14: quoting

vars/finishArgusTestRun.groovy
```diff
@@ -0,0 +1,20 @@
...
+
+    ./docker/env/hydra.sh finish-argus-test-run --jenkins-status ${test_status}
```
k0machi, line 16: quoting

k0machi (changes requested):

vponomaryov (approved):
LGTM

k0machi (approved):

fruch (approved):

PR #6165, opened 2023-05-22
fix(_terminate_node): Filter error abort request upon node termination

aleksbykov:
DBErrors `torage_proxy - .* to read from keyspace1.standard1: abort requested` could happen during terminate_and_replace_node. Such error were seen during perf latency with operations. Add filter to ignore such errors

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -75,7 +75,7 @@ from sdcm.sct_events import Severity
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.decorators import raise_event_on_failure
 from sdcm.sct_events.filters import DbEventsFilter, EventsSeverityChangerFilter, EventsFilter
-from sdcm.sct_events.group_common_events import (ignore_alternator_client_errors, ignore_no_space_errors,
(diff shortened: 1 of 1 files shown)


fruch:
You probably want to mention
https://github.com/scylladb/scylladb/issues/10447

In the commit, or even in a comment in the code.

fruch:
> You probably want to mention [scylladb/scylladb#10447](https://github.com/scylladb/scylladb/issues/10447)
> 
> In the commit, or even in a comment in the code.

@aleksbykov, also where this one should be backported

aleksbykov:
@fruch , i prepared this PR for performance tests, because we see similar issue during node terminated. But job uses old branch(perf-v12) and recently this jobs was disable, i didn't set any target. ON new jobs we didn't catch issue. I can adopt the fix also for drain if it is requiired.

fruch:
> @fruch , i prepared this PR for performance tests, because we see similar issue during node terminated. But job uses old branch(perf-v12) and recently this jobs was disable, i didn't set any target. ON new jobs we didn't catch issue. I can adopt the fix also for drain if it is requiired.

so I don't understand, do we need this change on master or not ? if not please close or point to the branch where it's needed

aleksbykov:
CLosed, No need for perf test on branch-v14

PR #6190, opened 2023-05-26
fix(cluster.py): do not print false errors not finding 'cpu_set' value

vponomaryov:
Read `/etc/scylla.d/cpuset.conf` file contents and parse it in python
instead of using 'grep' command which returns exit code '1' when no match is found.
It will allow us to avoid following false error messages:

    > Command: 'grep "^CPUSET" /etc/scylla.d/cpuset.conf'
    >
    > Exit code: 1
    >
    > Stdout:
    >
    >
    >
    > Stderr:
    >
    >
    >

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -556,13 +556,20 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             # or
             #   'CPUSET="--cpuset 1-7,9-15,17-23,25-31 "'
             # And so on...
-            grep_result = self.remoter.run('grep "^CPUSET" /etc/scylla.d/cpuset.conf')
+            cpuset_file_lines = self.remoter.run("cat /etc/scylla.d/cpuset.conf").stdout
         except Exception as exc:  # pylint: disable=broad-except
             self.log.error(f"Failed to get CPUSET. Error: {exc}")
             return ''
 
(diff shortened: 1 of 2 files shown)


sdcm/cluster.py
```diff
@@ -556,7 +556,11 @@ def cpuset(self):
             # or
             #   'CPUSET="--cpuset 1-7,9-15,17-23,25-31 "'
             # And so on...
-            grep_result = self.remoter.run('grep "^CPUSET" /etc/scylla.d/cpuset.conf')
+            # NOTE: consider as error only 2+ exit codes because the '1' exit code in case of 'grep'
+            #       means 'match not found' which happens always on K8S.
+            grep_result = self.remoter.run(
+                "grep \"^CPUSET\" /etc/scylla.d/cpuset.conf"
+                " || EXIT_CODE=$? ; if [[ $EXIT_CODE != 1 ]]; then exit $EXIT_CODE; fi")
```
fgelcer, line 563: shound't it be `[[ $EXIT_CODE -eq  1 ]]` ?

fruch, line 563: wouldn't it be clearer to do all this logic on the python level ?

fruch (changes requested):
see if you can move the logic from bash to python

fruch (approved):
LGTM

soyacz (approved):

PR #6202, opened 2023-05-30
fix(k8s-ci): use the same `cluster_name` in all K8S monitoring methods

vponomaryov:
In K8S functional tests we use `db_cluster.name` attr for the
`ScyllaDBMonitoring` object name, but in the
`check_kubernetes_monitoring_health` method we automatically pick up
the `db_cluster.scylla_cluster_name`. Both are different in CI.
    
So, fix the K8S monitoring functional tests to use the same attribute
as the health checker does.

- [c] All new and existing unit tests passed (CI)

### DIFF
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -155,7 +155,8 @@ def test_rolling_restart_cluster(db_cluster):
 
 
 def _scylla_cluster_monitoring_ckecks(db_cluster: ScyllaPodCluster, monitoring_type: str):
-    k8s_cluster, cluster_name, namespace = db_cluster.k8s_cluster, db_cluster.name, db_cluster.namespace
+    k8s_cluster = db_cluster.k8s_cluster
+    cluster_name, namespace = db_cluster.scylla_cluster_name, db_cluster.namespace
 
     # NOTE: deploy prometheus operator if absent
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

sdcm/sct_config.py
```diff
@@ -1758,6 +1758,8 @@ def __init__(self):
             self['user_prefix'] = "{}-{}".format(user_prefix, version_tag)[:35]
         else:
             self['user_prefix'] = user_prefix[:35]
+        if 'k8s' in cluster_backend:
+            self['user_prefix'] = self['user_prefix'].lower()
```
fruch, line 1762: we might do this closer to where it goes to k8s ? i.e. this is related to something new, we didn't had this before...

vponomaryov: It appeared with the addon of the K8S monitoring functional tests which use DB cluster name for K8S object creation.
There were no other such cases before.

fruch:
Check why unittest fails

fruch (approved):
LGTM

sdcm/sct_config.py
vponomaryov: Not actual anymore. Just uploaded really correct fix.

vponomaryov:
With the current state of the PR the K8S monitoring functional tests have passed.

PR #6208, opened 2023-06-01
fix(k8s): make static loaders work again

vponomaryov:
The `fix(k8s): make pod waiters use pod selectors` commit in the (https://github.com/scylladb/scylla-cluster-tests/pull/6130) PR made 
the `static` loaders be broken by using wrong selector. 
The common label between `dynamic` and `static` loaders is `loader-cluster-name`
not `loader-name` which exists only in `dynamic` case.

The labels may be checked by looking at the config files located at
the `sdcm/k8s_configs/loaders/` directory and `metadata.labels` field.

So, use proper loaders pod selector which will work in both cases.


fruch (approved):
LGTM

PR #6212, opened 2023-06-02
fix(k8s): set Scylla storage size in 14 tenants longevity

fruch:
In case of 14 tenants setup we should use at most '1/14' of the available storage. Available is 3.5Tb, so, set 249Gb for each of 14 tenants to make it possible to work with reserved storage.

9029b271c61e5c9d6e62b3c6f178516ee1be7fb3 changed this only for the perf tests, and missed the longvities


vponomaryov (approved):
LGTM

fruch:
passed clusters creation in:
https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-scylla-operator-12h-multitenant-eks/58/

@vponomaryov I still wonder if this number is low enough, and would we have space to grow cluster nemesis ? or other topology changes nemesis ? should we leave some space for at least one node per cluster ?

soyacz (approved):

PR #6213, opened 2023-06-02
fix(tox): use config syntax compatible for 3.x and 4.x versions

vponomaryov:
The current `tox.ini` config was written for the `3.23.x` versions. It is pretty old versions. Now we have `4.5.x`.
So, update the `tox.ini` config to support both due to the following reason:
- `LTS Ubuntu 22` has `tox` package, but it is of the `3.21.x` version
- We can install any `tox` package version from sources such as `4.5.1`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -16,8 +16,8 @@ sitepackages = False
 
 [testenv]
 envdir = {toxworkdir}/{envname}
-passenv = HOME PYTEST_* SCT_*
-whitelist_externals = *
+passenv = HOME,PYTEST_*,SCT_*
+allowlist_externals = *
 commands =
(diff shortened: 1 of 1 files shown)


fruch:
you are using tox to build the local environment ? how it's better the what's in our docs using pyenv ?

vponomaryov:
> you are using tox to build the local environment ? how it's better the what's in our docs using pyenv ?

Yes, I use `tox` tool. I don't need to remember where I activated and deactivated the env.
Using tox I only add `tox -e py310 -- ` prefix and that's it.

soyacz (approved):
LGTM.
I didn't use tox, but sounds interesting. Maybe small doc how to use that in our scope?

fruch (approved):
LGTM

PR #6178, opened 2023-05-23
 feature(replication): use network topology replication everywhere

soyacz:
It is planned to deprecate `SimpleStrategy` and we shouldn't use it in
SCT tests (as anyway it's not recommended for production).

We can avoid providing datacenter names in NetworkTopologyStrategy when
all DC's use the same replication factor by providing
`replication_factor` key. In that case all DC's ought to have the same
RF equal to this value.

Switch all usages of `SimpleStrategy` to `NetworkTopologyStrategy`.
Replace all explicitly set DC names wherever test uses same RF factor
for all DC's (all single DC scenarios and several multi-dc ones)

refs: https://github.com/scylladb/scylla-cluster-tests/issues/5969

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6216, opened 2023-06-04
fix(longevity-cdc-100gb-4h): stop hardcoding NetworkTopologyStrategy

fruch:
cause of change to the case to support multiple AZs the case was changed to use NetworkTopologyStrategy, but was hardcoded to specific region, changing it to use `replication_factor` so we don't need to know regions/datacenters in advence.


soyacz:
fixed in: https://github.com/scylladb/scylla-cluster-tests/pull/6178

fgelcer:
> fixed in: #6178

if it was fixed in there, we would have a conflict in here, but we don't

fruch:
done in https://github.com/scylladb/scylla-cluster-tests/pull/6178

PR #6182, opened 2023-05-23
test(latency during upgrade): based on other latency test

fgelcer:
used as base the acceptance criteria latency test,
the latency during operations, with 650GB dataset,
this test will run as the "operations" a rolling upgrade
on all the nodes.
currently it has the default of the base test (3 nodes)
on a single DC and same region. (more to be added).

QA-Task is https://github.com/scylladb/qa-tasks/issues/1191

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6221, opened 2023-06-07
refactor(performance): Perf emails

temichus:
1 appends the backend and node type to the existing subject of Perf Emails that support
differend backends

2 move Prometheus stats to the bottom of perf emails body

### DIFF
--- a/microbenchmarking_test.py
+++ b/microbenchmarking_test.py
@@ -33,4 +33,6 @@ class PerfSimpleQueryTest(ClusterTester):
                 specific_tested_stats={"perf_simple_query_result": json.loads(output)},
                 doc_id_with_timestamp=True)
             if self.create_stats:
-                PerfSimpleQueryAnalyzer(self._test_index, self._es_doc_type).check_regression(self._test_id)
+                is_gce = self.params.get('cluster_backend') == 'gce'
+                PerfSimpleQueryAnalyzer(self._test_index, self._es_doc_type).check_regression(
+                    self._test_id, is_gce=is_gce)
--- a/sdcm/results_analyze/__init__.py
+++ b/sdcm/results_analyze/__init__.py
@@ -107,6 +107,11 @@ class BaseResultsAnalyzer:  # pylint: disable=too-many-instance-attributes
                 [(setup_param.replace('gce_', ''), test_doc['_source']['setup_details'].get(setup_param))])
         return setup_details
 
+    @staticmethod
+    def _get_email_tags(test_doc, is_gce):
+        setup_details = BaseResultsAnalyzer._get_setup_details(test_doc, is_gce)
+        return f"[{setup_details['cluster_backend']}][{setup_details['instance_type_db']}]"
+
     def _test_version(self, test_doc):
         if test_doc['_source'].get('versions'):
             for value in ('scylla-server', 'scylla-enterprise-server'):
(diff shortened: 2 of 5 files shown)


temichus:
@roydahan @fgelcer @aleksbykov put your thoughts regarding these changes before I start testing them 

from @avikivity 
Some ergonomics:
 - move "Performance Regression Compare Results" to the end of the title. It's repetitive and doesn't contain information.
 - add the node type
 - drop the latency measurements, they're useless. At least don't start with latency on throughput tests.

temichus:
Screenshot from 2023-06-09 17-59-53

fgelcer (dismissed):
it will conflict with my current PR #6182 , but LGTM

aleksbykov (approved):
LGTM, I little concern about begin of email subject. It starts with `[perf-regression]`
May be it is better to `Perf Regression` or `Performance Regression`, WDYT?

temichus:
> LGTM, I little concern about begin of email subject. It starts with `[perf-regression]` May be it is better to `Perf Regression` or `Performance Regression`, WDYT?

I prefer to use something without whitespaces for easy search in emails mess

temichus:
@fruch @soyacz  Could you please merge/backport it?

sdcm/results_analyze/__init__.py
```diff
@@ -1455,7 +1462,7 @@ def check_regression(self, test_name, test_results, test_details):  # pylint: di
             test_details=test_details,
             screenshots=test_details.pop("screenshots"),
         )
-        subject = f"Performance Regression: {test_name} - {format_timestamp(test_details['start_time'])}"
+        subject = f"[perf_regression] {test_name} - {format_timestamp(test_details['start_time'])}"
```
roydahan, line 1465: Changing the subject at this point will be more painful than keeping it.
If one wants to search the inbox, one will need to search both the old subject and the new one while the short version isn't that much shorter.

I suggest not changing the subject.

sdcm/results_analyze/__init__.py
```diff
@@ -1473,7 +1480,7 @@ def __init__(self, es_index, es_doc_type, email_recipients=(), logger=None, even
                          email_template_fp="results_search_best_throughput_config.html", logger=logger, events=events)
 
     def check_regression(self, test_name, setup_details, test_results):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
-        subject = f"Performance Regression Best throughput with configuation - {test_name} - {setup_details['start_time']}"
+        subject = f"[search_best_throughput][perf_regression] - {test_name} - {setup_details['start_time']}"
```
roydahan, line 1483: same as above

roydahan (changes requested):
Please do no change any subject, see my comment for it inline.

temichus:
@roydahan,  I've updated PR now it just appends the backend and node type to the existing subject. Please check
testing in progress.

temichus:
perf_simple_query 
Screenshot from 2023-08-17 15-25-58

temichus:
perf tests reporting
Screenshot from 2023-08-17 16-24-59

roydahan (dismissed):
Looks good, just update the commit message because you didnt change the title eventually.

microbenchmarking_test.py
```diff
@@ -33,4 +33,6 @@ def test_perf_simple_query(self):
                 specific_tested_stats={"perf_simple_query_result": json.loads(output)},
                 doc_id_with_timestamp=True)
             if self.create_stats:
-                PerfSimpleQueryAnalyzer(self._test_index, self._es_doc_type).check_regression(self._test_id)
+                is_gce = bool(self.params.get('cluster_backend') == 'gce')
```
aleksbykov, line 36: Result of expression a == b , is bool, so no need to explicitly

aleksbykov (approved):

microbenchmarking_test.py
temichus: fixed

roydahan (approved):

PR #6223, opened 2023-06-07
fix(k8s-eks): use proper disk sizes for Scylla pods

vponomaryov:
Before, when we were using `static local volume provisioner`, the disk usage was not limited and a test run could use whole disk space even if PVCs for Scylla pods were having a lower limit than the real disk capacity.

When we switched to the `dynamic local volume provisioner` the PVC size limit started making the effect not allowing to use storage more than configured.

The default value of 500Gi becomes not enough in some cases. For example the `large collections` one.

So, configure the Scylla pods storage to be `k8s_scylla_disk_gi: 3490` by default which is close to the real disk available capacity (`3.5Tb`). Update the multitenant configurations separately
based on the number of tenants.

It will allow us to avoid the following error:

    Compacting of 3 sstables interrupted due to: \
        storage_io_error (Storage I/O error: 28: No space left on device)

### DIFF
--- a/configurations/operator/perf-regression-latency-and-throughput-multitenant-2-clients.yaml
+++ b/configurations/operator/perf-regression-latency-and-throughput-multitenant-2-clients.yaml
@@ -1,6 +1,9 @@
 test_duration: 540
 user_prefix: 'perf-regrn-lat-and-thpt-2-clients'
 
+# NOTE: K8S Scylla nodes have 3.5Tb disk size
+k8s_scylla_disk_gi: 1745
+
 # NOTE: deploy 8 K8S nodes of the 'loader' type and then create 4 loader pods per DB cluster
 n_loaders: 8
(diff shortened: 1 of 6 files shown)


vponomaryov:
Example of the error: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/longevity-scylla-operator-large-collections-12h/25/consoleFull
```
11:51:32  < t:2023-06-07 08:51:32,056 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:ERROR > ERROR 2023-06-07 08:50:46,449 [shard  2] compaction_manager - Compaction task 0x60700a0b3c10 for table large_collection_test.table_with_large_collection [0x607008e20460]: failed due to storage io error: Storage I/O error: 28: No space left on device: stopping
```

PR #6227, opened 2023-06-08
fix(cdc-100gb): use any-region compatible replication config

vponomaryov:
For the moment the `longevity-cdc-100gb-4h` is configured the way that it works only in `eu-west` regions:

    replication = {'class': 'NetworkTopologyStrategy', 'eu-west': '3'}

Since we support more regions than compatible with the configuration we should use region-agnostic configuration like following:

    replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': '3'}

So, update the `longevity-cdc-100gb-4h` job because it is used for testing of new patch-releases of the `branch-5.2` Scylla branch.


vponomaryov:
It is needed exactly for the `branch-5.2` where the `longevity-cdc-100gb-4h` is among CI jobs triggered to test minor releases and it get assigned with `us-east-1` region, not `eu-west-*`.

fruch:
We should back port also to 2023.1

fruch:
> We should back port also to 2023.1

@vponomaryov can you do the backport?

PR #6235, opened 2023-06-11
fix(wait_for): Support stop wait_for by event

aleksbykov:
During prepare new nemesis BootstrapStreaming error, for some case it is need to stop wait_for function ealier and don't wait whole timeout

example when you expect that bootstrap will failed and you can find appropriate error in log, but not need to raise any exception or critical event, wait for db up will continue to wait while port will be available

For such case it will be good to have flag, if it is set, than stop wait_for function

- ~[ ] New configuration option are added and documented (in `sdcm/sct_config.py`)~
- ~[ ] I have updated the Readme/doc folder accordingly (if needed)~

### DIFF
--- a/sdcm/wait.py
+++ b/sdcm/wait.py
@@ -22,17 +22,14 @@ from collections.abc import Callable
 
 import tenacity
 
+from sdcm.exceptions import WaitForTimeoutError, ExitByEventError
 
 LOGGER = logging.getLogger("sdcm.wait")
 
 R = TypeVar("R")  # pylint: disable=invalid-name
 
 
-class WaitForTimeoutError(Exception):
-    pass
-
-
-def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
+def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, stop_event=None, **kwargs):  # pylint: disable=too-many-arguments
     """
     Wrapper function to wait with timeout option.
 
@@ -42,6 +39,7 @@ def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
     :param timeout: Timeout in seconds
     :param throw_exc: Raise exception if timeout expired, but disrupt_func result is not True
     :param kwargs: Keyword arguments to disrupt_func
+    :param stop_event: instance of threading.Event class to stop retrying
     :return: Return value of disrupt_func.
     """
     if not timeout:
@@ -57,11 +55,14 @@ def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
             retry_state.attempt_number,
             retry_state.outcome._exception or retry_state.outcome._result,
         )
+    stops = [tenacity.stop_after_delay(timeout)]
+    if stop_event:
+        stops.append(tenacity.stop.stop_when_event_set(stop_event))
 
     try:
         retry = tenacity.Retrying(
             reraise=throw_exc,
-            stop=tenacity.stop_after_delay(timeout),
+            stop=tenacity.stop_any(*stops),
             wait=tenacity.wait_fixed(step),
             before_sleep=retry_logger,
             retry=(tenacity.retry_if_result(lambda value: not value) | tenacity.retry_if_exception_type())
@@ -70,6 +71,11 @@ def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
 
     except Exception as ex:  # pylint: disable=broad-except
         err = f"Wait for: {text or func.__name__}: timeout - {timeout} seconds - expired"
+        raising_exc = WaitForTimeoutError(err)
+        if stop_event and stop_event.is_set():
+            err = f": {text or func.__name__}: stopped by Event"
+            raising_exc = ExitByEventError(err)
+
         LOGGER.error(err)
         if hasattr(ex, 'last_attempt') and ex.last_attempt.exception() is not None:  # pylint: disable=no-member
             LOGGER.error("last error: %r", ex.last_attempt.exception())  # pylint: disable=no-member
@@ -77,7 +83,7 @@ def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
             LOGGER.error("last error: %r", ex)
         if throw_exc:
             if hasattr(ex, 'last_attempt') and not ex.last_attempt._result:  # pylint: disable=protected-access,no-member
-                raise WaitForTimeoutError(err) from ex
+                raise raising_exc from ex
             raise
 
     return res
--- a/unit_tests/test_wait.py
+++ b/unit_tests/test_wait.py
@@ -18,10 +18,12 @@ import threading
 import time
 import unittest
 
+from concurrent.futures import ThreadPoolExecutor
+
 import pytest
 
 from sdcm.cluster import BaseNode
-from sdcm.wait import wait_for, wait_for_log_lines
+from sdcm.wait import wait_for, wait_for_log_lines, WaitForTimeoutError, ExitByEventError
 
 logging.basicConfig(level=logging.DEBUG)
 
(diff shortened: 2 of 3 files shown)


sdcm/wait.py
```diff
@@ -53,18 +53,27 @@ def retry_logger(retry_state):
...
             retry=(tenacity.retry_if_result(lambda value: not value) | tenacity.retry_if_exception_type())
         )
         res = retry(func, **kwargs)
 
     except Exception as ex:  # pylint: disable=broad-except
+        if stop_event and stop_event.is_set():
+            err = f": {text or func.__name__}: stopped by Event"
+            if hasattr(ex, 'last_attempt'):
+                res = ex.last_attempt._result  # pylint: disable=protected-access,no-member
+            return res
```
fruch, line 75: why returning and not raising error as `throw_exc` might suggest ?

sdcm/cluster.py
```diff
@@ -1302,12 +1302,12 @@ def _report_housekeeping_uuid(self, verbose=False):
...
 
-    def wait_db_up(self, verbose=True, timeout=3600):
+    def wait_db_up(self, verbose=True, timeout=3600, stop_event: threading.Event | None = None):
         text = None
         if verbose:
             text = '%s: Waiting for DB services to be up' % self
-        wait.wait_for(func=self.db_up, step=60, text=text, timeout=timeout, throw_exc=True)
-        self.db_init_finished = True
+        res = wait.wait_for(func=self.db_up, step=60, text=text, timeout=timeout, throw_exc=True, stop_event=stop_event)
+        self.db_init_finished = bool(res)
```
fruch, line 1310: since `throw_exc=True` I would assume that in case of failure `wait_for` won't return.
so why changing this to look at the result ?

fruch:
@aleksbykov I would suggest adding a few unittests when touch such heavily used implementation.

sdcm/wait.py
```diff
@@ -53,18 +53,27 @@ def retry_logger(retry_state):
             retry_state.attempt_number,
             retry_state.outcome._exception or retry_state.outcome._result,
         )
+    stops = [tenacity.stop_after_delay(timeout)]
+    if stop_event := kwargs.pop("stop_event", None):
```
temichus, line 57: add stop_event in doc string

sdcm/wait.py
```diff
@@ -53,18 +53,27 @@ def retry_logger(retry_state):
...
+        stops.append(tenacity.stop.stop_when_event_set(stop_event))
 
     try:
         retry = tenacity.Retrying(
             reraise=throw_exc,
-            stop=tenacity.stop_after_delay(timeout),
+            stop=tenacity.stop_any(*stops),
```
temichus, line 63: 1st unit test for this, that stop now is list

sdcm/wait.py
```diff
@@ -53,18 +53,27 @@ def retry_logger(retry_state):
...
             wait=tenacity.wait_fixed(step),
             before_sleep=retry_logger,
             retry=(tenacity.retry_if_result(lambda value: not value) | tenacity.retry_if_exception_type())
         )
         res = retry(func, **kwargs)
 
     except Exception as ex:  # pylint: disable=broad-except
+        if stop_event and stop_event.is_set():
```
temichus, line 71: 2dt unit test for this, for different combinations for timeout(exceed/nor exceed) , event (set/not set), func result (result/exception)

sdcm/wait.py
aleksbykov: if wait_for is stop by stop stopped event, it could be done only explicitly via Event.set(), and there is no need to raise the Exception, just check the result from last retry if it will be required.

sdcm/cluster.py
aleksbykov: This was as example how it could be used. Removed this changes

aleksbykov:
Removed unrelated changes, Added unit tests

sdcm/wait.py
aleksbykov: done

temichus: not covered:  func result (result/exception)

I think we need cover logic when there is an exception and stop_event  set/not set

temichus (commented):
not covered: func result (result/exception)

I think we need cover logic when there is an exception and stop_event set/not set

because it is one of the core logic in this PR

sdcm/wait.py
aleksbykov: What kind of exception: by timeout  or exception in target function?

aleksbykov: - test_04_return_result_before_stop_event_and_wait_timeout - return target function result before event was set and timeout
- test_04_raise_by_timeout_before_set_event  - raise by timeout before event set

temichus: exception in target function and  event  set
exception in target function and  event is not  set

aleksbykov: Added. But due to we retry if any exception triggered, the behavior was not changed

aleksbykov:
@fruch @soyacz @vponomaryov could you check?

unit_tests/test_wait.py
```diff
@@ -79,6 +81,109 @@ def callback(arg1, arg2):
...
         self.assertEqual(len(calls), 1)
 
+    def test_04_stop_by_event(self):
+        calls = []
+
+        def callback(arg1, arg2):
+            calls.append((arg1, arg2))
+            if len(calls) == 5:
+                return "what ever"
+            return False
```
vponomaryov, line 91: This is defined separately for each of the unit tests. Please, remove the code duplication.

unit_tests/test_wait.py
```diff
@@ -79,6 +81,109 @@ def callback(arg1, arg2):
...
+        th = ThreadPoolExecutor(max_workers=1).submit(wait_for, func=callback, timeout=3,
+                                                      step=.5, arg1=1, arg2=3, throw_exc=False, stop_event=ev)
+        time.sleep(1)
+        ev.set()
+        res = th.result()
+        assert not th.exception()
+        self.assertFalse(res)
+        self.assertTrue(len(calls) < 5)
+
+    def test_04_stop_by_event_without_exception(self):
```
vponomaryov, line 124: This test and the below one could be parameterized method/function.

unit_tests/test_wait.py
```diff
@@ -79,6 +81,109 @@ def callback(arg1, arg2):
...
+        calls = []
+
+        def callback(arg1, arg2):
+            calls.append((arg1, arg2))
+            if len(calls) == 8:
+                return "what ever"
+            return False
+
+        def set_stop_in_timeout(ev: threading.Event):
+            time.sleep(4)
```
vponomaryov, line 176: This is unit test. It should not depend on the real time.
In good case it must be solved with the `mock`s.

sdcm/wait.py
```diff
@@ -38,6 +38,7 @@ def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
     :param timeout: Timeout in seconds
     :param throw_exc: Raise exception if timeout expired, but disrupt_func result is not True
     :param kwargs: Keyword arguments to disrupt_func
+    :param stop_event: instance of threading.Event class to stop retrying
```
vponomaryov, line 41: Here the original logic for parameters is violated.
Would be better to define the `stop_event` on the line 31 explicitly.

sdcm/wait.py
vponomaryov: Then, to keep the interface be sane, need to rename that param to something like `throw_exc_on_timeout`.

vponomaryov: Yes, interface will be changed. And yes, it will require updates to the existing explicit usages of it by it's consumers.
It is smaller problem than it is now - `not throwing exception when it is explicitely set to be "True"`.

vponomaryov: @fruch , @temichus WDYT?

aleksbykov: @vponomaryov ,  i preview how many changes is required for renaming, i decided to raise exception on if throw_exc true and event set. Is it ok?

unit_tests/test_wait.py
aleksbykov: fixed

unit_tests/test_wait.py
aleksbykov: done

unit_tests/test_wait.py
aleksbykov: removed the time dependency

sdcm/wait.py
aleksbykov: added

aleksbykov:
@vponomaryov could you take a look one more time. I change behavior, if throw_exc is set , the exception will be raised if wait_for will be stopped by stop event

sdcm/wait.py
vponomaryov: > @vponomaryov , i preview how many changes is required for renaming, i decided to raise exception on if throw_exc true and event set. Is it ok?

Ok. But I don't see it in the PR's code changes.

unit_tests/test_wait.py
```diff
@@ -80,6 +82,108 @@ def callback(arg1, arg2):
...
+            if len(self.calls) == set_after:
+                ev.set()
+
+    @parameterized.expand([(True, ), (False, )])
+    def test_04_stop_by_event(self, throw_exc):
+        self.callback_return_true_after = 3
+        th = threading.Thread(target=self.set_stop_in_timeout, kwargs={"ev": self.ev, "set_after": 1})
+        th.start()
+        if throw_exc:
+            self.assertRaisesRegex(Exception, "callback: stopped by Event", wait_for,
```
vponomaryov, line 117: It must be some `ValueError` exception where we should see a message about the incompatibility of the `throw_exc` parameter and the `stop_event`. Isn't it?

vponomaryov (changes requested):

sdcm/wait.py
aleksbykov: You can see, that in Exception block if event was set , i just change the error message, and if throw_exc is true, then it will be raised with another error message

unit_tests/test_wait.py
aleksbykov: if throw_exc is true and wait_for is stopped by event, then Retry exception will be raised with appropriate error message. and test expect that , if false, then exception will not be raised

sdcm/wait.py
```diff
@@ -70,11 +74,17 @@ def retry_logger(retry_state):
 
     except Exception as ex:  # pylint: disable=broad-except
```
temichus, line 75: what about this?

```
    except Exception as ex:  # pylint: disable=broad-except
        if stop_event and stop_event.is_set():
            err = f": {text or func.__name__}: stopped by Event"
            raise WaitForStopEventtError(err) from ex
        else:
            err = f"Wait for: {text or func.__name__}: timeout - {timeout} seconds - expired"
            LOGGER.error(err)
            if hasattr(ex, 'last_attempt') and ex.last_attempt.exception() is not None:  # pylint: disable=no-member
                LOGGER.error("last error: %r", ex.last_attempt.exception())  # pylint: disable=no-member
            else:
                LOGGER.error("last error: %r", ex)
                res = ex.last_attempt._result if hasattr(  # pylint: disable=protected-access,no-member
                    ex, 'last_attempt') else None

            if throw_exc:
                if hasattr(ex, 'last_attempt') and not ex.last_attempt._result:  # pylint: disable=protected-access,no-member
                    raise WaitForTimeoutError(err) from ex
```

and now the function will have several scenarios
not changed:
return result - when func returns result
WaitForTimeoutError - when func returns an error, the event is not set(or none), timeout is exceeded
new:
WaitForStopEventtError - when func returns an error but the event is set, and SCT code will expect this error  to do not "raise any exception or critical event,"
It will be like just one more condition of exit without touching existing logic 

also, as I understood, exit by the event will be some kind of positive exit. so I think LOGGER.error is not needed in this case

aleksbykov:
@vponomaryov , @temichus could you take a look. Fixed according your latest comments

temichus (approved):

sdcm/wait.py
```diff
@@ -32,7 +32,11 @@ class WaitForTimeoutError(Exception):
     pass
 
 
-def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, **kwargs):
+class ExitByEventError(Exception):
```
vponomaryov, line 35: We have dedicated module for it here: https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/exceptions.py

Please, reuse it.

vponomaryov (changes requested):
One inline comment plus following:

Do you have an example CI run where this new parameter gets used?
For now, it is addition of a unused functionality...

sdcm/wait.py
temichus: @vponomaryov 

```
class WaitForTimeoutError(Exception):
    pass
```
 from the line 31, located in this file.
Should we move WaitForTimeoutError in this case as well?

aleksbykov: i moved both exceptions to sdcm.exceptions module

aleksbykov:
@vponomaryov could you take a look again. I moved exceptions to module sdcm.exception.
Regarding to CI and using. I was going to use this functionality in new nemesis BootstrapStreamingError. But decided to introduce it in new PR.
So once this pr will be merged, i will rework DecommissionStreamingError/BootstrapStreamingError to use this functionality.
it should help to avoid not need timeout waiting

temichus:
@vponomaryov, Could you please re-review?

vponomaryov:
@aleksbykov , @temichus 

From code point of view looks ok, but the concern that the code is not used is still actual.
We should not merge code that is not tested even in one CI job.

@aleksbykov 
It is totally ok if you add one more commit with the usage in this PR.

fruch:
> @aleksbykov , @temichus
> 
> From code point of view looks ok, but the concern that the code is not used is still actual. We should not merge code that is not tested even in one CI job.
> 
> @aleksbykov It is totally ok if you add one more commit with the usage in this PR.

seems like it's test to some degree, i.e. it has unit-tests.
as long we have enough confidence and unit-tests it won't do regression (or other tests proving no regression)

I think it's o.k. to merge such an extension code, and do follow ups PRs using it.
I agree I would rather prefer it on the same PR, and not span across multiple PRs, to figure out how something is working.

but both direction are fine by me.

soyacz (commented):
Assuming it will be used because we need this to prevent from false errors I'm ok to create another PR later with use of it.
Question is: why to separate it? What if solution proves to be wrong?

fruch:
> Assuming it will be used because we need this to prevent from false errors I'm ok to create another PR later with use of it. Question is: why to separate it? What if solution proves to be wrong?

Similar to any change we do, it might be proven wrong or incomplete later on.

example that we might want such a thing, if we have multiple cases we want to use it, and each usage would be its own PR.
and there are lots of examples we are doing such thing, putting it some machinery and only later using it for something.
for example the introduction of the nemesis.yml, it was introduced first on it's own PR, assuming there would be two usages of ti, and then other PR was using it, and the 2nd part never happened.

as I've said, I'm o.k. with both options, as long as who leads it is aware and has a clear plan.

soyacz (approved):

fruch:
@aleksbykov is this change still needed for any task ?

aleksbykov:
> @aleksbykov is this change still needed for any task ?

@fruch , yes, i use to fix BootstrapStreaming error

PR #6247, opened 2023-06-13
feature(EaR-KMS): Add tests for EaR AWS KMS service

fruch:
Scylla added a support for AWS KMS to be used with
the enterprise feature of Encryption-at-Rest.

* adding new test configuration yaml for it
* support it in scylla.yaml

### Testing
- [x] - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/EaR-longevity-kms-200gb-6h-test/2/

### DIFF
--- a/configurations/kms-ear.yaml
+++ b/configurations/kms-ear.yaml
@@ -7,3 +7,5 @@ append_scylla_yaml: |
       kms_test:
           master_key: 'alias/kms_encryption_test'
           aws_region: 'us-east-1'
+
+append_scylla_args: '--logger-log-level kms=trace'
--- a/configurations/kms-ear.yaml
+++ b/configurations/kms-ear.yaml
@@ -1,0 +1,9 @@
+
+scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'kms_test'}"
+
+# config kms_hosts for AWS KMS service account
+append_scylla_yaml: |
+  kms_hosts:
+      kms_test:
+          master_key: 'alias/kms_encryption_test'
+          aws_region: 'us-east-1'
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -1595,6 +1595,14 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
                 self.parent_cluster.proposed_scylla_yaml,
                 self.proposed_scylla_yaml
             )
+            is_kms = bool(scylla_yml.kms_hosts)
+
+        if is_kms:
+            # Hack for overriding issue https://github.com/scylladb/scylla-enterprise/issues/2792
+            # TODO: should be remove once a proper fix is implemented
+            self.remoter.sudo("find /opt/scylladb/ -iname *libp11-kit.so* | sudo xargs rm",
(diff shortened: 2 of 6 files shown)


sdcm/cluster.py
```diff
@@ -1595,6 +1595,32 @@ def config_setup(self,
                 self.parent_cluster.proposed_scylla_yaml,
                 self.proposed_scylla_yaml
             )
+            is_kms = bool(scylla_yml.kms_hosts)
+
+        if is_kms:
+            # Hack for overriding issue https://github.com/scylladb/scylla-enterprise/issues/2792
+            # TODO: should be remove once a proper fix is implemented
+            self.remoter.sudo("find /opt/scylladb/ -iname *libp11-kit.so* | sudo xargs rm",
```
vponomaryov, line 1603: Shouldn't it consider distro? As I understood, it must be run only on Ubuntu/Debian, right?

configurations/kms-ear.yaml
```diff
@@ -0,0 +1,11 @@
+
+scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'kms_test'}"
+
+# config kms_hosts for AWS KMS service account
+append_scylla_yaml: |
+  kms_hosts:
+      kms_test:
+          master_key: 'alias/kms_encryption_test'
+          aws_region: 'us-east-1'
```
vponomaryov, line 9: The job ran in the `eu-west-1` region but here we have different one -> `us-east-1`.
How then it works? Is it ignored for now?

configurations/kms-ear.yaml
```diff
@@ -0,0 +1,11 @@
+
+scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'kms_test'}"
+
+# config kms_hosts for AWS KMS service account
+append_scylla_yaml: |
+  kms_hosts:
+      kms_test:
+          master_key: 'alias/kms_encryption_test'
```
vponomaryov, line 8: Where do we get it from?

configurations/kms-ear.yaml
fruch: it's actually doesn't matter the key we are using is needs to be in this region.

KMS code doesn't know what's the current region, and works only by this configuration.
we can raise an issue for it to be able to figure out the region if not specified

configurations/kms-ear.yaml
fruch: It's a key we manually created.

I guess at a point we should add code that create/delete the key for each test, once we want to do more things with it
like force a roll of that key, I'm guess we don't want to affect any other currently running test.

so yeah hard code key, is just a start to get it working.

configurations/kms-ear.yaml
```diff
@@ -0,0 +1,11 @@
+
+scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'kms_test'}"
```
vponomaryov, line 2: Shouldn't it be `KmsRealKeyProviderFactory` instead of the `KmsKeyProviderFactory`?

PR #6250, opened 2023-06-14
refactor scale test

roydahan:

### DIFF
--- a/longevity_test.py
+++ b/longevity_test.py
@@ -24,7 +24,10 @@ from cassandra import AlreadyExists, InvalidRequest
 
 from sdcm.tester import ClusterTester
 from sdcm.utils import loader_utils
+from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
 from sdcm.utils.operations_thread import ThreadParams
+from sdcm.sct_events.system import InfoEvent
+from sdcm.cluster import MAX_TIME_WAIT_FOR_NEW_NODE_UP
 
 
 class LongevityTest(ClusterTester, loader_utils.LoaderUtilsMixin):
@@ -104,6 +107,25 @@ class LongevityTest(ClusterTester, loader_utils.LoaderUtilsMixin):
             self.run_tombstone_gc_verification_thread(**tombstone_gc_verification_params)
 
         self.run_prepare_write_cmd()
+
+        # Grow cluster to target size if requested
+        if cluster_target_size := self.params.get('cluster_target_size'):
+            add_node_cnt = self.params.get('add_node_cnt')
+            node_cnt = len(self.db_cluster.nodes)
+
+            InfoEvent(message=f"Starting to grow cluster from {node_cnt} to {cluster_target_size}").publish()
+
+            while node_cnt < cluster_target_size:
+                InfoEvent(message=f"f'Adding node number {node_cnt + 1}").publish()
+                new_nodes = self.db_cluster.add_nodes(count=add_node_cnt, enable_auto_bootstrap=True)
+                self.monitors.reconfigure_scylla_monitoring()
+                up_timeout = MAX_TIME_WAIT_FOR_NEW_NODE_UP
+                with adaptive_timeout(Operations.NEW_NODE, node=self.db_cluster.nodes[0], timeout=up_timeout):
+                    self.db_cluster.wait_for_init(node_list=new_nodes, timeout=up_timeout, check_node_health=False)
+                self.db_cluster.wait_for_nodes_up_and_normal(nodes=new_nodes)
+                node_cnt = len(self.db_cluster.nodes)
+
+            InfoEvent(message=f"f'Growing cluster finished, new cluster size is {node_cnt}, ").publish()
 
         # Collect data about partitions and their rows amount
         validate_partitions = self.params.get('validate_partitions')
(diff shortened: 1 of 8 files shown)


roydahan:
First Successful run **without** nemesis:
Setup of 15 nodes took 1h10m.
Adding 10 more nodes took 2 hours.

Screenshot 2023-07-13 at 14 19 03

longevity_test.py
```diff
@@ -105,6 +108,25 @@ def test_custom_time(self):
 
         self.run_prepare_write_cmd()
 
+        # Grow cluster to target size if requested
```
roydahan, line 111: At the moment this part is taking place after the prepare (so we have some data in the cluster while growing) but without any stress.
On one hand, it's better to test the grow under stress, but OTOH it may take too long so the stress will complete before the cluster grew to the target size and before any nemesis has run or not enough nemesis.

Hence, I prefer to leave it like this for now and see how it goes.

roydahan:
Still under final testing (had to run some reproducers to check a possible issue).
I'll squash soon all commits and fix pre-commit issues.

But in general ready for first review.

roydahan:
Full run with nemesis is running here:
http://3.253.40.91:3000/d/fmH0g_CVk/scale-base-test-scylla-per-server-metrics-nemesis-master?orgId=1&from=now-30m&to=now&refresh=30s

https://argus.scylladb.com/test/73ca5f18-5a00-4812-a059-aeb1eb59c50d/runs?additionalRuns[]=78990975-c9ac-47bb-b932-391272005a47

roydahan:
This is ready for review.
Still not bullet proof in terms of issues that may happen with various nemesis but the test works better than it was and more organized in terms of our jobs.

roydahan:
re-organizing commits and resolving conflicts.

fgelcer (approved):
small comment, but if you have proof it works, LGTM

longevity_test.py
```diff
@@ -105,6 +108,25 @@ def test_custom_time(self):
...
+            node_cnt = len(self.db_cluster.nodes)
+
+            InfoEvent(message=f"Starting to grow cluster from {node_cnt} to {cluster_target_size}").publish()
+
+            while node_cnt < cluster_target_size:
+                InfoEvent(message=f"f'Adding node number {node_cnt + 1}").publish()
+                new_nodes = self.db_cluster.add_nodes(count=add_node_cnt, enable_auto_bootstrap=True)
+                self.monitors.reconfigure_scylla_monitoring()
+                up_timeout = MAX_TIME_WAIT_FOR_NEW_NODE_UP
+                with adaptive_timeout(Operations.NEW_NODE, node=self.db_cluster.nodes[0], timeout=up_timeout):
```
soyacz, line 123: we wait here for `add_node_cnt` - in case it is >1 our stats for `NEW_NODE` operations will be spoiled.
Anyway we add nodes one by one, what is the reason of adding multiple like this?

roydahan: Sorry, but I didn't understand the question

soyacz: we add multiple nodes at once `self.db_cluster.add_nodes(count=add_node_cnt, enable_auto_bootstrap=True)` what is the reason of doing like this instead of 1 node at a time?

roydahan: We don't add multiple nodes at once, since it's not supported for scylla (for now).
The only correct way is to add them one by one.

test-cases/scale/scale-cluster.yaml
```diff
@@ -1,25 +1,46 @@
...
+
+nemesis_class_name: 'SisyphusMonkey'
 nemesis_interval: 5
-nemesis_class_name: 'ChaosMonkey'
+nemesis_during_prepare: false
+
+user_prefix: 'cluster-base-scale-test'
 
-cluster_health_check: false
+use_preinstalled_scylla: true
```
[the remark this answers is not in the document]
roydahan: Saw it, but prefer to leave it here to be on the safe side.

longevity_test.py
soyacz: So I'd propose to switch count to `1` in `self.db_cluster.add_nodes(count=add_node_cnt, enable_auto_bootstrap=True)`

Otherwise `with adaptive_timeout(Operations.NEW_NODE...` will collect wrong duration for `NEW_NODE` operation

roydahan: I don't understand why it would collect wrong duration, it's the same as all other add nodes we have.
Each add_node should be recorded only once.

roydahan: Well, in this context it's always 1, and when we change how scylla supports adding more than one (tablets), it will change the statstics of adding nodes anyway.

longevity_test.py
```diff
@@ -105,6 +108,25 @@ def test_custom_time(self):
...
 
+        # Grow cluster to target size if requested
+        if cluster_target_size := self.params.get('cluster_target_size'):
+            add_node_cnt = self.params.get('add_node_cnt')
+            node_cnt = len(self.db_cluster.nodes)
+
+            InfoEvent(message=f"Starting to grow cluster from {node_cnt} to {cluster_target_size}").publish()
+
+            while node_cnt < cluster_target_size:
+                InfoEvent(message=f"f'Adding node number {node_cnt + 1}").publish()
```
juliayakovlev, line 119: Why double "f" here? Is it needed?
```
message=f"f'Adding node number {node_cnt + 1}"
```

juliayakovlev (approved):
It looks good. One small typo

longevity_test.py
roydahan: I missing that comment and that typo :)

PR #6260, opened 2023-06-18
improvement(azure): add feature for cloud_provider initial support.

fgelcer:
Since 5.2 and 2023.1 will be the 1st official versions
of Scylla to support Azure (OSS and Enterprise
respectively), we needed a mechanism to add support
for it. It right now gives us a place to add
the initial supported version for different cloud
providers, if they will be added to Scylla.
So let's suppose that, for each OSS we upgrade from
previous version + current version previous released patches.
In case of Azure cloud provider, we will no do that, as
for `5.2` as an example, `5.1` doesn't officially support
Azure images... so we will know how to correctly
filter, in our upgrade jobs, what is the "oldest"
version we should use in these tests.

Relates to https://github.com/scylladb/qa-tasks/issues/1116

### DIFF
--- a/unit_tests/test_base_version.py
+++ b/unit_tests/test_base_version.py
@@ -17,11 +17,11 @@ import unittest
 from utils.get_supported_scylla_base_versions import UpgradeBaseVersion  # pylint: disable=no-name-in-module, import-error
 
 
-def general_test(scylla_repo='', linux_distro=''):
+def general_test(scylla_repo='', linux_distro='', cloud_provider=None):
     scylla_version = None
 
     version_detector = UpgradeBaseVersion(scylla_repo, linux_distro, scylla_version)
-    version_detector.set_start_support_version()
+    version_detector.set_start_support_version(cloud_provider)
     _, version_list = version_detector.get_version_list()
     return version_list
 
@@ -35,6 +35,24 @@ class TestBaseVersion(unittest.TestCase):
         linux_distro = 'centos'
         version_list = general_test(scylla_repo, linux_distro)
         self.assertEqual(version_list, ['5.3'])
+
+    def test_ubuntu_22_azure_5_3(self):
+        scylla_repo = self.url_base + '/branch-5.3/deb/unified/2023-05-23T22:36:08Z/scylladb-5.3/scylla.list'
+        linux_distro = 'ubuntu-jammy'
+        cloud_provider = 'azure'
+        version_list = general_test(scylla_repo, linux_distro, cloud_provider)
+        self.assertEqual(version_list, ['5.2'])
+
+    def test_ubuntu_22_azure_2023_1(self):
+        scylla_repo = self.url_base +\
+            '-enterprise/enterprise-2023.1/deb/unified/2023-05-17T23:03:35Z/scylladb-2023.1/scylla.list'
+        linux_distro = 'ubuntu-jammy'
+        cloud_provider = 'azure'
+        version_list = general_test(scylla_repo, linux_distro, cloud_provider)
+        # TODO: add 2023.1 to the list below:
+        # once 2023.1 will be released, it will appear in the list below, but as it is not only on an RC,
+        # it was filtered out during `self.filter_rc_only_version(ent_base_version)`
+        self.assertEqual(version_list, ['5.2'])
 
     def test_4_5_with_centos8(self):
         scylla_repo = self.url_base + '/branch-4.5/rpm/centos/2021-08-29T00:58:58Z/scylla.repo'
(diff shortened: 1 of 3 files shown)


unit_tests/test_base_version.py
```diff
@@ -36,6 +36,24 @@ def test_master(self):
         version_list = general_test(scylla_repo, linux_distro)
         self.assertEqual(version_list, ['5.3'])
 
+    def test_ubuntu_22_azure_5_3(self):
+        scylla_repo = self.url_base + '/branch-5.3/deb/unified/2023-05-23T22:36:08Z/scylladb-5.3/scylla.list'
```
yaronkaikov, line 40: shouldn't we use the stable path ?

unit_tests/test_base_version.py
```diff
@@ -36,6 +36,24 @@ def test_master(self):
...
+        linux_distro = 'ubuntu-jammy'
+        cloud_provider = 'azure'
+        version_list = general_test(scylla_repo, linux_distro, cloud_provider)
+        self.assertEqual(version_list, ['5.2'])
+
+    def test_ubuntu_22_azure_2023_1(self):
+        scylla_repo = self.url_base +\
+                      '-enterprise/enterprise-2023.1/deb/unified/2023-05-17T23:03:35Z/scylladb-2023.1/scylla.list'
```
yaronkaikov, line 48: Same here

yaronkaikov (changes requested):

unit_tests/test_base_version.py
fgelcer: 1st, this is only a unittest, and i don't think it will change much from the unittest perspective.. but i can try to change... do we have a "stable path" for 5.3 already?

unit_tests/test_base_version.py
fgelcer: ditto?

unit_tests/test_base_version.py
yaronkaikov: @fgelcer So this is the default repo if `scylla_repo` is not provided? The official repo file is https://s3.amazonaws.com/downloads.scylladb.com/deb/ubuntu/scylla-5.3.list

fgelcer: > @fgelcer So this is the default repo if `scylla_repo` is not provided? The official repo file is https://s3.amazonaws.com/downloads.scylladb.com/deb/ubuntu/scylla-5.3.list

no, in this case, this is used only for this unittest... this is a test we run as part of the CI of SCT new code..
it is not used in any SCT jobs... just a matter to verify (in this case) if the `base version` mechanism that calculates based on the repo URL does the job correctly...
i have added a new "feature" to the base version mechanism, as Azure initial support is `5.2` and `2023.1`, so there is no point running upgrade from any older version them these...

fruch:
@fgelcer 

The title of this PR doesn't say anything...

The description also isn't clear what exactly is being added here.

Now we can get the latest releases based on images ? and not based on repo ?

fgelcer:
> @fgelcer
> 
> The title of this PR doesn't say anything...
> 

the "title" was too long, so it entered next line...
the idea here is to add the initial support version for a new cloud provider (Azure), so tests that automatically calculate the base version (rolling upgrade) will choose only versions that are supported, even if the version should be upgrading from 2 previous versions (or more, in case of Enterprise), only as far as it is bigger or equal the initial support introduced will be used...

> The description also isn't clear what exactly is being added here.

well, i thought that this is clear, i will try to rephrase it

> 
> Now we can get the latest releases based on images ? and not based on repo ?

yaronkaikov (approved):

utils/get_supported_scylla_base_versions.py
```diff
@@ -18,6 +18,7 @@
 supported_src_oss = {'2021.1': '4.3', '2022.1': '5.0', '2022.2': '5.1', '2023.1': '5.2', '2023.2': '5.3'}
 # If new support distro shared repo with others, we need to assign the start support versions. eg: centos8
 start_support_versions = {'centos-8': {'scylla': '4.1', 'enterprise': '2021.1'}}
+start_support_cloud_provider = {'azure': {'scylla': '5.2', 'enterprise': '2023.1'}}
```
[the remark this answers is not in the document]
temichus: >in that case, we will fail somewhere else, most likely when running init_and_verify_sct_config()

you just introduced the start_support_cloud_provider variable and used in only in set_start_support_version. to "Setting start support version". how it can affect init_and_verify_sct_config()?

> perhaps, we won't need it anymore, once 5.2 will be old enough (and not supported anymore)..

we just released 5.2. when it will be not supported anymore?



anyway, we can leave it as is because we are not introducing new cloud support every week.

temichus (approved):

fgelcer:
ping @scylladb/qa-maintainers

PR #6268, opened 2023-06-20
fix(mgmt_cli_test): Properly check if gsutil is installed

ShlomiBalalis:
Previously, we checked if gsutil was installed based on whether or not the $HOME/google-cloud-sdk directory existed, which turned out to be an incorrect way of verifying that. Instead, we will now try to check gsutil's version.

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -89,21 +89,11 @@ class BackupFunctionsMixIn(LoaderUtilsMixin):
         self._run_cmd_with_retry(executor=node.remoter.sudo, cmd=shell_script_cmd(cmd))
 
     def install_gsutil_dependencies(self, node):
-        def is_dir_exists(path):
-            shell = dedent(f"""
-            if [ -d {path} ]
-            then
-                exit 0
-            else
-                exit 1
-            fi
-            """)
-            result = node.remoter.sudo(shell, ignore_status=True)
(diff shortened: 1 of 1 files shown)


fgelcer:
add to the commit message the motivation of this change (what failed, where and when)

mgmt_cli_test.py
```diff
@@ -89,21 +89,11 @@ def install_awscli_dependencies(self, node):
...
-            result = node.remoter.sudo(shell, ignore_status=True)
+        def is_gsutil_installed():
+            result = node.remoter.run("gsutil -v", ignore_status=True)
             return result.exited == 0
-
-        sdk_directory_path = '$HOME/google-cloud-sdk'
-        if not is_dir_exists(path=sdk_directory_path):
-            self._run_cmd_with_retry(executor=node.remoter.run, cmd=shell_script_cmd("""\
+        if not is_gsutil_installed():
+            self._run_cmd_with_retry(executor=node.remoter.sudo, cmd=shell_script_cmd("""\
```
fgelcer, line 96: from a quick check i did on live nodes related to this fix, this script doesn't work, as `gsutil` is not installed in none of the default paths set in the `PATH` environment variable...

fruch:
@ShlomiBalalis is this still relevent ?

PR #6273, opened 2023-06-20
fix(upgrade test): not stop test on execute session timeout

juliayakovlev:
Add "try .. except" in" truncate_entries" method. It is for preventing of stop the test on request timeout.

We have seen request timeout in rolling-upgrade tests. Example:
https://jenkins.scylladb.com/job/enterprise-2023.1/job/rolling-upgrade/job/rolling-upgrade-with-sla-test/23/

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -14,6 +14,7 @@
 # Copyright (c) 2016 ScyllaDB
 
 # pylint: disable=too-many-lines
+import traceback
 from itertools import zip_longest, chain, count as itertools_count
 import json
 from pathlib import Path
@@ -28,6 +29,7 @@ from argus.client.sct.types import Package
 from sdcm import wait
 from sdcm.cluster import BaseNode
 from sdcm.fill_db_data import FillDatabaseData
+from sdcm.sct_events import Severity
 from sdcm.stress_thread import CassandraStressThread
 from sdcm.utils.version_utils import get_node_supported_sstable_versions
 from sdcm.sct_events.system import InfoEvent
@@ -46,7 +48,13 @@ def truncate_entries(func):
     def inner(self, *args, **kwargs):
         node = args[0]
         with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks') as session:
-            self.cql_truncate_simple_tables(session=session, rows=NUMBER_OF_ROWS_FOR_TRUNCATE_TEST)
+            InfoEvent(message="Start truncate simple tables").publish()
+            try:
+                self.cql_truncate_simple_tables(session=session, rows=NUMBER_OF_ROWS_FOR_TRUNCATE_TEST)
+                InfoEvent(message="Finish truncate simple tables").publish()
+            except Exception as details:  # pylint: disable=broad-except
+                InfoEvent(message=f"Failed truncate simple tables. Error: {str(details)}. Traceback: {traceback.format_exc()}",
+                          severity=Severity.ERROR).publish()
             self.validate_truncated_entries_for_table(session=session, system_truncated=True)
 
         func_result = func(self, *args, **kwargs)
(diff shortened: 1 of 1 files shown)


fruch (dismissed):
LGTM

@juliayakovlev can you add reference where an issue was spotted, since there isn't any issue reference in the description/commit, I think it was in sla upgrade test, but it would have written if we run into more issue similar to it

juliayakovlev:
> LGTM
> 
> @juliayakovlev can you add reference where an issue was spotted, since there isn't any issue reference in the description/commit, I think it was in sla upgrade test, but it would have written if we run into more issue similar to it

Added

upgrade_test.py
```diff
@@ -45,14 +47,21 @@ def truncate_entries(func):
     @wraps(func)
     def inner(self, *args, **kwargs):
         node = args[0]
-        with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks') as session:
-            self.cql_truncate_simple_tables(session=session, rows=NUMBER_OF_ROWS_FOR_TRUNCATE_TEST)
+        with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks', connect_timeout=240) as session:
```
fgelcer, line 50: shouldn't this timeout be a bit bigger than the DML command?

upgrade_test.py
```diff
@@ -45,14 +47,21 @@ def truncate_entries(func):
...
+                          severity=Severity.ERROR).publish()
+            InfoEvent(message="Finish truncate simple tables").publish()
+
             self.validate_truncated_entries_for_table(session=session, system_truncated=True)
 
         func_result = func(self, *args, **kwargs)
 
         # re-new connection
-        with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks') as session:
+        with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks', connect_timeout=240) as session:
```
fgelcer, line 64: same here

fgelcer (commented):
other than a little comment, LGTM

upgrade_test.py
juliayakovlev: increased to 300

upgrade_test.py
juliayakovlev: increased to 300

upgrade_test.py
fgelcer: i see it is still 240

upgrade_test.py
```diff
@@ -46,7 +48,14 @@ def truncate_entries(func):
...
         node = args[0]
         with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks') as session:
-            self.cql_truncate_simple_tables(session=session, rows=NUMBER_OF_ROWS_FOR_TRUNCATE_TEST)
+            InfoEvent(message="Start truncate simple tables").publish()
+            try:
+                self.cql_truncate_simple_tables(session=session, rows=NUMBER_OF_ROWS_FOR_TRUNCATE_TEST)
+            except Exception as details:  # pylint: disable=broad-except
+                InfoEvent(message=f"Failed truncate simple tables. Error: {str(details)}. Traceback: {traceback.format_exc()}",
+                          severity=Severity.ERROR).publish()
+            InfoEvent(message="Finish truncate simple tables").publish()
```
fgelcer, line 57: move this to inside the try clause, as we need it only if the exception is not raised

upgrade_test.py
```diff
@@ -114,14 +123,19 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
...
+            InfoEvent(message="Start read data from truncated tables").publish()
+            try:
+                count = self.rows_to_list(session.execute(truncate_query.format(table_name)))
+                self.assertEqual(str(count[0][0]), '0',
+                                 msg='Expected that there is no data in the table truncate_ks.{}, but found {} rows'
+                                 .format(table_name, count[0][0]))
+            except Exception as details:  # pylint: disable=broad-except
+                InfoEvent(message=f"Failed read data from truncated tables. Error: {str(details)}. Traceback: {traceback.format_exc()}",
+                          severity=Severity.ERROR).publish()
+            InfoEvent(message="Finish read data from truncated tables").publish()
```
fgelcer, line 138: same here

upgrade_test.py
```diff
@@ -46,7 +48,14 @@ def truncate_entries(func):
     def inner(self, *args, **kwargs):
         node = args[0]
         with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks') as session:
-            self.cql_truncate_simple_tables(session=session, rows=NUMBER_OF_ROWS_FOR_TRUNCATE_TEST)
+            InfoEvent(message="Start truncate simple tables").publish()
+            try:
+                self.cql_truncate_simple_tables(session=session, rows=NUMBER_OF_ROWS_FOR_TRUNCATE_TEST)
+            except Exception as details:  # pylint: disable=broad-except
```
fgelcer, line 54: out of curiosity, a timeout will raise an exception too, right?

juliayakovlev: I added all exception details to the error event.
We do not want to raise Exception as it will stop the test. In this case it's not a goal. The test will fail as got error event, but will continue.

upgrade_test.py
juliayakovlev: moved

upgrade_test.py
juliayakovlev: moved

upgrade_test.py
fgelcer: yes, i understand that we won't raise the exception, but my question was, a timeout (inside the try clause) will raise an exception, right?

fgelcer (approved):
LGTM

upgrade_test.py
juliayakovlev: > yes, i understand that we won't raise the exception, but my question was, a timeout (inside the try clause) will raise an exception, right?

Ah, I see. Yes, it will

juliayakovlev:
@fruch please, review

fgelcer:
@roydahan @enaydanov @vponomaryov @soyacz , please backport this one

roydahan:
Backported to 2023.1

PR #6278, opened 2023-06-21
fix(perf_stress): Split stress multiplier by stress workload

aleksbykov:
Depend on stress workload, number of processes could give different max throughput. Split stress multiplier on 3 w,r,m. This will allow to configure different number of stress process on loader instance for write, read, mixed independently.

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -118,6 +118,9 @@ append_scylla_args_oracle: '--enable-cache false'
 
 # cassandra-stress defaults
 stress_multiplier: 1
+stress_multiplier_w: 1
+stress_multiplier_r: 1
+stress_multiplier_m: 1
 keyspace_num: 1
 cs_user_profiles: []
 cs_duration: '50m'
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -965,7 +965,13 @@ class SCTConfiguration(dict):
 
         # LongevityTest
         dict(name="stress_multiplier", env="SCT_STRESS_MULTIPLIER", type=int,
-             help=""),
+             help="Number of cassandra-stress processes"),
+        dict(name="stress_multiplier_w", env="SCT_STRESS_MULTIPLIER_W", type=int,
+             help="Number of cassandra-stress processes for write workload"),
+        dict(name="stress_multiplier_r", env="SCT_STRESS_MULTIPLIER_R", type=int,
+             help="Number of cassandra-stress processes for read workload"),
+        dict(name="stress_multiplier_m", env="SCT_STRESS_MULTIPLIER_M", type=int,
+             help="Number of cassandra-stress processes for mixed workload"),
         dict(name="run_fullscan", env="SCT_RUN_FULLSCAN", type=list,
              help=""),
         dict(name="run_full_partition_scan", env="SCT_run_full_partition_scan", type=str,
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -224,7 +224,10 @@
 | **<a href="#user-content-cassandra_stress_population_size" name="cassandra_stress_population_size">cassandra_stress_population_size</a>**  |  | 1000000 | SCT_CASSANDRA_STRESS_POPULATION_SIZE
 | **<a href="#user-content-cassandra_stress_threads" name="cassandra_stress_threads">cassandra_stress_threads</a>**  |  | 1000 | SCT_CASSANDRA_STRESS_THREADS
 | **<a href="#user-content-add_node_cnt" name="add_node_cnt">add_node_cnt</a>**  |  | 1 | SCT_ADD_NODE_CNT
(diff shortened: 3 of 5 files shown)


defaults/test_default.yaml
```diff
@@ -118,6 +118,9 @@ append_scylla_args_oracle: '--enable-cache false'
 
 # cassandra-stress defaults
 stress_multiplier: 1
+stress_multiplier_w: 1
```
fruch, line 121: I don't really like the kind of specific test parameters.

Those parameters won't work in all of the tests, and it's not clear to figure out which they would and in which they wouldn't.

I would think this is a per stress command parameter maybe we could add it to the stress command in a way we could extract it out of the command itself ? We had define something like that for the idea of having diffrent docker image per command, but it wasn't implemented.

fgelcer: +1

in case we have `cassandra-stress write no-warmup cl=QUORUM n=30000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=100 -pop seq=1..30000000` (the one you intent to change in this PR), you could for example add `processes=2` in the end, so inside the code that takes care of running the threads, you check that value, to use as a multiplier, and don't send it to the inner call of c-s, and then you have something that will work much easier to all c-s commands.. WDYT?

fruch: I would call it `stress_multiplier` to avoid confusion, processes might be a bit vague in this regards.

Anyhow laying the foundations for having SCT parameters per command, would be helpful in more cases, like if we want one with `round_robin` and one without, defining which command is the main load and should be critical and which command shouldn't stop the test.

and I think we can have more examples like, that we were a bit limited by setting a "global' configuration option.

aleksbykov:
@fruch , @roydahan , @fgelcer could you review and backport to perf-v14?

fgelcer (approved):
for the sake of having the job passing, i'm approving it, but i think the right way to go is to revert it, and add a parser to the stress commands, and do collect parameters like this `multiplier` to do it correctly in the call to the stress thread

sdcm/sct_config.py
```diff
@@ -965,7 +965,13 @@ class SCTConfiguration(dict):
 
         # LongevityTest
         dict(name="stress_multiplier", env="SCT_STRESS_MULTIPLIER", type=int,
-             help=""),
+             help="Number of cassandra-stress processes"),
```
fruch, line 968: If we are introducing a new SCT  parameter, we should generate the docs.

Also if they are relevant only to very specific cases, it should be documented

defaults/test_default.yaml
aleksbykov: May be then to split c-s parameters and and control parameters with -- 
`cassandra-stress write no-warmup cl=QUORUM n=30000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=100 -pop seq=1..30000000 -- stress_multiplier=2 round_robin=True`

fgelcer: > May be then to split c-s parameters and and control parameters with -- `cassandra-stress write no-warmup cl=QUORUM n=30000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=100 -pop seq=1..30000000 -- stress_multiplier=2 round_robin=True`

yes, that is my suggestion

roydahan: Laying the foundations is good idea.
However, I wouldn't put it as suggested above.
We probably need to pick a better data structure, like nested yaml or list of lists or something like that...
E.g.
prepare_stress_cmd: [["c-s command", round_robin: True, stress_multiplier: 2],
                                      ["c-s command", round_robin: True, stress_multiplier: 3]]

aleksbykov:
This change is required to performance tests. The max throughput is reached on different configurations of loader instances, number of processes, and threads for write, read, mixed payload. No jobs are failed.

defaults/test_default.yaml
aleksbykov: For now may we can use this implementation: https://github.com/scylladb/scylla-cluster-tests/pull/6293 and create task to refactore c-s commands config style?

roydahan: Yes, I agree.
I don't want to hold this PR for the refactor.

roydahan (approved):

aleksbykov:
@soyacz , @roydahan could you backport this pr

vponomaryov:
@aleksbykov it cannot be backported cleanly. Looks like it depend on on some other, not backported change.

aleksbykov:
@vponomaryov , create another PR for perf-v14: https://github.com/scylladb/scylla-cluster-tests/pull/6329

aleksbykov:
Could first backported this PR: https://github.com/scylladb/scylla-cluster-tests/pull/5957
And then this one: https://github.com/scylladb/scylla-cluster-tests/pull/6278

aleksbykov:
@vponomaryov many thanks

PR #6279, opened 2023-06-21
fix(list images): avoid getting exception

fgelcer:
since moved to check new releng account for
the images that will be created in there in
the near future, we are experiencing still
exceptions when there are no tags (but for
some reason, there was meta.data), so the
idea is to also check if there are tags, so
we won't try to "list them" whem they don't
exist.

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -2056,13 +2056,13 @@ def get_ami_tags(ami_id, region_name):
     scylla_images_ec2_resource = get_scylla_images_ec2_resource(region_name=region_name)
     new_test_image = scylla_images_ec2_resource.Image(ami_id)
     new_test_image.reload()
-    if new_test_image and new_test_image.meta.data:
+    if new_test_image and new_test_image.meta.data and new_test_image.tags:
         return {i['Key']: i['Value'] for i in new_test_image.tags}
     else:
         ec2_resource: EC2ServiceResource = boto3.resource('ec2', region_name=region_name)
         test_image = ec2_resource.Image(ami_id)
         test_image.reload()
-        if test_image and test_image.tags and test_image.meta.data:
+        if test_image and test_image.meta.data and test_image.tags:
             return {i['Key']: i['Value'] for i in test_image.tags}
         else:
             return {}


ShlomiBalalis (approved):

Annamikhlin (approved):

sdcm/utils/common.py
```diff
@@ -2056,13 +2056,13 @@ def get_ami_tags(ami_id, region_name):
...
     new_test_image.reload()
-    if new_test_image and new_test_image.meta.data:
+    if new_test_image and new_test_image.meta.data and new_test_image.tags:
         return {i['Key']: i['Value'] for i in new_test_image.tags}
     else:
         ec2_resource: EC2ServiceResource = boto3.resource('ec2', region_name=region_name)
         test_image = ec2_resource.Image(ami_id)
         test_image.reload()
-        if test_image and test_image.tags and test_image.meta.data:
+        if test_image and test_image.meta.data and test_image.tags:
```
fruch, line 2065: Not sure how changing the order here changes anything

fgelcer: changes, because in Python the checks are made left-to-right, and if we don't have `image.meta.data` we surely won't have `image.tags`, but the opposite can happen

fgelcer: cause in case `image.meta.data` doesn't exist, the `image.tags` will raise an exception when we try to check it... i still don't understand why it happens, but it is related to how the image is created by the `EC2.client.image` returning an object without anything inside of it...

fgelcer: if you don't understand, you can check it yoursef by running:
```
from sdcm.utils.common import get_ami_tags

ami_id_new_account = 'ami-043cdaa9ca233a0ed'
ami_id_old_account = 'ami-06e3bb1313f8b3b58'

region = 'us-east-1'

new_res = get_ami_tags(ami_id_new_account, region)
old_res = get_ami_tags(ami_id_old_account, region)
print(f'{new_res=}')
print(f'{old_res=}')
```

fruch: o.k. so it's point less to check tags as well, if it would be empty it would be enough

```diff
diff --git a/sdcm/utils/common.py b/sdcm/utils/common.py
index 95698f441..a4465fd47 100644
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -2062,7 +2062,7 @@ def get_ami_tags(ami_id, region_name):
         ec2_resource: EC2ServiceResource = boto3.resource('ec2', region_name=region_name)
         test_image = ec2_resource.Image(ami_id)
         test_image.reload()
-        if test_image and test_image.tags and test_image.meta.data:
+        if test_image and test_image.meta.data:
             return {i['Key']: i['Value'] for i in test_image.tags}
         else:
             return {}
```

fruch: also what are those images ? deperacted images ? or images just from different accounts ?

fruch: BTW this is a bug on boto3:
https://github.com/boto/boto3/issues/2531

fruch (approved):
LGTM

sdcm/utils/common.py
fgelcer: > also what are those images ? deperacted images ? or images just from different accounts ?

i saw a case recently (on a job by @ShlomiBalalis ), where we did have `image.meta.data`, but tags weren't in there... so i rather prefer to add this check, as the line below (which iterates through the tags) will fail if `image.tags` is `None`

i prefer to take the safe side, and check both...

and thank you for the issue... i mentioned to @yaronkaikov that it should be an issue, but i didn't find one...

PR #6280, opened 2023-06-22
fix(releng account images): check owner_id

fgelcer:
since we started working on adding support
to images created on the releng account,
this `ami_built_by_scylla` function was
experiencing same failure as we saw on
PR #6279, but in a different function call.
I have made a small refactor on the function
`get_ami_tags`, so now it will include the
`owner_id`, and we don't need to repeat
code in `ami_built_by_scylla` call as well.

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6282, opened 2023-06-22
fix(elastic): store ami name

temichus:
store ami_name_db_scylla in elastic

### DIFF
--- a/sdcm/db_stats.py
+++ b/sdcm/db_stats.py
@@ -31,7 +31,7 @@ import requests
 
 from sdcm.es import ES
 from sdcm.test_config import TestConfig
-from sdcm.utils.common import normalize_ipv6_url
+from sdcm.utils.common import normalize_ipv6_url, get_ami_tags
 from sdcm.utils.git import get_git_commit_id
 from sdcm.utils.decorators import retrying
 from sdcm.sct_events.system import ElasticsearchEvent
@@ -582,6 +582,9 @@ class TestStatsMixin(Stats):
         except Exception as ex:  # pylint: disable=broad-except
             LOGGER.error('Failed getting scylla versions: %s', ex)
 
+        # append Scylla build mode in scylla_versions
+        versions['scylla-server']['build_mode'] = node.scylla_build_mode
+
         return versions
 
     def get_setup_details(self):
@@ -605,6 +608,14 @@ class TestStatsMixin(Stats):
                     setup_details['n_db_nodes'] = sum([int(i) for i in value.split()])
                 else:
                     setup_details[key] = value
+
+        if self.params.get('cluster_backend') == 'aws':
+            setup_details["ami_tags_db_scylla"] = []
+            region_names = self.params.region_names
+            ami_list = self.params.get('ami_id_db_scylla').split()
+            for ami_id, region_name in zip(ami_list, region_names):
+                tags = get_ami_tags(ami_id, region_name)
+                setup_details["ami_tags_db_scylla"].append(tags)
 
         new_scylla_packages = self.params.get('update_db_packages')
         setup_details['packages_updated'] = bool(new_scylla_packages and os.listdir(new_scylla_packages))
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -1671,7 +1671,8 @@ class SCTConfiguration(dict):
                                 break
                         else:
                             raise ValueError(f"AMIs for {scylla_version=} not found in {region}")
-                    self.log.debug("Found AMI %s for scylla_version='%s' in %s", ami.image_id, scylla_version, region)
+                    self.log.debug("Found AMI %s(%s) for scylla_version='%s' in %s",
+                                   ami.name, ami.image_id, scylla_version, region)
                     ami_list.append(ami)
                 self['ami_id_db_scylla'] = " ".join(ami.image_id for ami in ami_list)
             elif not self.get("gce_image_db") and self.get("cluster_backend") == "gce":
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2114,6 +2114,15 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         return scylla_version
 
     @optional_cached_property
+    def scylla_build_mode(self) -> Optional[str]:
+        result = self.remoter.run(f"{self.add_install_prefix('/usr/bin/scylla')} --build-mode", ignore_status=True)
+        if result.ok:
+            return result.stdout.strip()
+        self.log.debug("Unable to get ScyllaDB build-mode using `%s':\n%s\n%s",
+                       result.command, result.stdout, result.stderr)
+        return None
+
+    @optional_cached_property
     def scylla_version(self) -> Optional[str]:
         if scylla_version := self.scylla_version_detailed:
             if match := SCYLLA_VERSION_RE.match(scylla_version):
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -1866,6 +1866,7 @@ def get_branched_ami(scylla_version: str, region_name: str, arch: AwsArchType = 
     filters = [
         {"Name": "tag:branch", "Values": [branch, ], },
         {"Name": "architecture", "Values": [arch, ], },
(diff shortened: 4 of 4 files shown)


temichus:
testet on perf_v14 branch  https://github.com/scylladb/scylla-cluster-tests/compare/branch-perf-v14...temichus:scylla-cluster-tests:put_ami_name_in_elastic_v14?expand=1

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/simple_query_weekly_microbenchmark_test/15/console

Screenshot from 2023-06-21 16-29-16

fgelcer (approved):
LGTM, but why do we need it? why isn't the result of `scylla --version` not enough?

fruch:
> LGTM, but why do we need it? why isn't the result of `scylla --version` not enough?

I guess if you want to rerun, it's much easier with ami-id and not scylla version.

sdcm/sct_config.py
```diff
@@ -1578,6 +1578,7 @@ class SCTConfiguration(dict):
     ]
     ami_id_params = ['ami_id_db_scylla', 'ami_id_loader', 'ami_id_monitor', 'ami_id_db_cassandra', 'ami_id_db_oracle']
     aws_supported_regions = ['eu-west-1', 'eu-west-2', 'us-west-2', 'us-east-1', 'eu-north-1', 'eu-central-1']
+    additional_data_for_elastic = {}
```
fruch, line 1581: it's weird that the configuration object of SCT is use for storage like this...
no very intuitive, and I don't want people to start assuming they can change things on `params` during the test case....

sdcm/db_stats.py
```diff
@@ -606,6 +606,10 @@ def get_setup_details(self):
                 else:
                     setup_details[key] = value
 
+        setup_details["additional_setup_details"] = {}
+        for key, value in self.params.additional_data_for_elastic.items():
```
fruch, line 610: I think we can use `get_ami_tags` function to get any information we need out of the AMIs
I think we can do that when report it needed, and not at the beginning of the test, saving it till the end of the test.
that small information gonna be kept on all tests for no reason.
also it would better if the logic is in one place for those things and less scattred

temichus:
> > > LGTM, but why do we need it? why isn't the result of `scylla --version` not enough?
> > 
> > 
> > I guess if you want to rerun, it's much easier with ami-id and not scylla version.
> 
> now I see it's the name, and I have no idea who would need the name... and for what purposes, so be almost identical to the version

@fruch 

previously it was a bug when SST takes DEBUG Scylla version Ami. now its fixed https://github.com/scylladb/scylla-cluster-tests/pull/5502

But @roydahan asked to  store information about is it debug Scylls or not in elastic - https://github.com/scylladb/qa-tasks/issues/1101

According to  Slack discussion in #engeniring and investigation, I realised that scylla ami with "bedug" is not debug
```
build-mode returns release

scyllaadm@ip-10-99-17-9:~$ scylla --build-mode
release

for ec2 https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#InstanceDetails:instanceId=i-0a0ddcd6302d5f834

sipnned up from  [ami-06298c5b14be2dd3e (debug-scylla-5.4.0-dev-x86_64-2023-06-12T11-09-27)](https://us-east-1.console.aws.amazon.com/ec2/home?region=us-east-1#Images:visibility=private;imageId=ami-06298c5b14be2dd3e)

and for https://hub.docker.com/r/scylladb/scylla-debug
```

So, looks like there is only one way to determine DEBUG Scylla ami in AWS - its ami name itself

temichus:
rebased with https://github.com/scylladb/scylla-cluster-tests/pull/6280 and tested https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/simple_query_weekly_microbenchmark_test/17/

fgelcer:
how was this AMI ID found? manually entered, or was "found" when sending `master:latest`? if the latter, the fix will go to that function to ignore all images that are called `debug-*` (and IIRC we already do that, but maybe there is a new introduced bug about it)

fruch:
> how was this AMI ID found? manually entered, or was "found" when sending `master:latest`? if the latter, the fix will go to that function to ignore all images that are called `debug-*` (and IIRC we already do that, but maybe there is a new introduced bug about it)

Could be that those fixes weren't backported to perf branches

Anyhow the debug images should be clearly tagged, or the other ways around, not tagged with any tag.

It's something we need to sort out with the releng team, collecting data won't help avoiding it.

temichus:
@fgelcer @fruch  as i dentionet this bu fixed in https://github.com/scylladb/scylla-cluster-tests/pull/5502

@fruch yes, collecting data won't help avoiding it. But it will help easily investigate. previously I and Roy take AMI ID for all bad results perf and go to AWS Console to check the name/tags and determine that this is Debug ami. now it will be stored in Elastic.

temichus:
> And see my inline notes, as to the implementation itself.
changed to post in Ecastic AMI tags

sdcm/db_stats.py
```diff
@@ -606,6 +606,14 @@ def get_setup_details(self):
...
                     setup_details[key] = value
 
+        if self.params.get('cluster_backend') == 'aws':
+            setup_details["ami_tags_db_scylla"] = []
+            region_names = self.params.region_names
+            ami_list = self.params.get('ami_id_db_scylla').split()
+            for ami_id, region_name in zip(ami_list, region_names):
+                tags = get_ami_tags(ami_id, region_name)
+                setup_details["ami_tags_db_scylla"].append(tags)
+
```
fruch, line 616: You are aware it's all the tags ? And not just the name?

fruch (approved):
LGTM

fgelcer (dismissed):
LGTM

roydahan:
I'm ok with storing the tags, although I don't think it's what we need to based on.
I saw the slack discussion, but these are my takeaways:
We need to exclude packages that has "debug" in their names (because they are non-official) + build-mode that isn't "release" (because they are scylla-debug).
And store this information and show it in the report.

temichus:
> I'm ok with storing the tags, although I don't think it's what we need to based on. I saw the slack discussion, but these are my takeaways: 
>We need to exclude packages that has "debug" in their names (because they are non-official) 

done in in https://github.com/scylladb/scylla-cluster-tests/pull/5502 need just backport in perf-v14
> +build-mode that isn't "release" (because they are scylla-debug). 

PR was updated to filter only release build types from branch searching
> And store this information and show it in the report.

I think that better to implement it when we move to perf report in argus. since current email reporting has lack common code. So I will need to uptate+test all perf reports which will take some time. It's not worth it, I think.

fruch (dismissed):
LGTM

fruch:
> My problem is that it relies too much on the tags which has many bugs.
> I would like to get the build-mode from the build itself.

I would argue this change doesn't rely on any tag, it just collects them all. which is needed for debugging purposes.

The new safe guard from getting debug mode images, is also useful in my POV.

We rarely use debug builds, and yes we could do a followup PR to collect the build mode from the command line and send it to elastic and argus and emails (any place we send out the scylla version)

roydahan:
> > My problem is that it relies too much on the tags which has many bugs.
> > I would like to get the build-mode from the build itself.
> 
> I would argue this change doesn't rely on any tag, it just collects them all. which is needed for debugging purposes.
> 
> The new safe guard from getting debug mode images, is also useful in my POV.
> 
> We rarely use debug builds, and yes we could do a followup PR to collect the build mode from the command line and send it to elastic and argus and emails (any place we send out the scylla version)

I didn't say none of them is useful, I just don't rely on the tags to be correct and I want to have also the actual "build-mode". I don't mind it will be a different PR, but to have it and hopefully to show it in the report so I won't need to go to the database to check if the low result is a result of debug build that accidentally got in our way and wasn't tagged properly.

temichus:
> My problem is that it relies too much on the tags which has many bugs. I would like to get the build-mode from the build itself.

Added storing build_mode rom the build itself.


Screenshot from 2023-06-27 13-57-02

fgelcer (approved):
LGTM, but don't you want to add to the email report as well?

fruch (approved):
LGTM

PR #6288, opened 2023-06-23
fix(k8s-perf): set correct PVC size for K8S thrpt tests

vponomaryov:
In the K8S throughput tests we use smaller instance type then in all other EKS tests.
So, we should redefine the disk size there which must be smaller than the default value.
Without it the Scylla PVC won't be scheduled breaking Scylla pods provisioning.


fruch (approved):
LGTM

PR #6291, opened 2023-06-25
fix(longevity-counters-multidc): unping scylla-bench

fruch:
the case was pinned to older s-b cause of: scylladb/scylla-bench#114 but since then we implemented s-b retires that should fix most of the of the timeout observed in 2023.1 run.

Ref: https://github.com/scylladb/scylla-bench/issues/114

### DIFF
--- a/test-cases/longevity/longevity-counters-multidc.yaml
+++ b/test-cases/longevity/longevity-counters-multidc.yaml
@@ -23,9 +23,3 @@ nemesis_add_node_cnt: 3
 
 server_encrypt: true
 internode_encryption: 'dc'
-
-use_legacy_cluster_init: false
-
-# pin s-b version to mitigate https://github.com/scylladb/scylla-bench/issues/114
-stress_image:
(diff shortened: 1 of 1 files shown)


vponomaryov (commented):
Maybe it is better to update the stress commands to have explicit configuration of the retries and intervals?

soyacz (approved):

fruch:
> Maybe it is better to update the stress commands to have explicit configuration of the retries and intervals?

We'll run it at least once, and see if we need more than the defaults.

Setting them blindly doesn't really goona help

soyacz:
> > Maybe it is better to update the stress commands to have explicit configuration of the retries and intervals?
> 
> We'll run it at least once, and see if we need more than the defaults.

it works with defaults

fruch:
> > > Maybe it is better to update the stress commands to have explicit configuration of the retries and intervals?
> > 
> > We'll run it at least once, and see if we need more than the defaults.
> 
> it works with defaults
> 

Let's merge and backport

vponomaryov (approved):
LGTM

PR #6302, opened 2023-06-27
 feature(counter_filter): Event procces to count events

aleksbykov:
For perf tests with operations, it is required to
 collect statisitcs about events: Reactor stall and
 sort them by stall duration. Also on next step,
 it is required to decode all reactor stalls by operations
 New event process is presented and new context manager.
 Context manager allow to start/stop count events and
 collect some stats
 event process allow to filter collected events and
 save events to files in specified directory


aleksbykov:
Example of report:
Screenshot from 2023-06-27 17-42-42
Screenshot from 2023-06-27 17-42-29

aleksbykov:
Staging job is running

aleksbykov:
Staging tests passed: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/scylla-master-perf-regression-latency-650gb-with-nemesis/14, report was sent

soyacz (commented):
Why it requires special event handler and not just counting on event publish?
Can you also elaborate why EventStatHandler saves events?

sdcm/report_templates/results_latency_during_ops_short.html
```diff
@@ -163,6 +166,19 @@ <h2>{{ operation }}</h2>
                             {% endfor %}
                         {% endif %}
                     </table>
+                    {% for cycle in results['cycles'] %}
```
fgelcer, line 169: nitpick: i would add all of this new for-loop after the current line 182 `<span STYLE="font-size:12px" class="red">* All latency values are in ms. if latency has color red, check detailed HDR report</span>` so we have that little legend on the tables closer to the table

temichus (dismissed):

sdcm/sct_events/event_counter.py
```diff
@@ -0,0 +1,192 @@
...
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2020 ScyllaDB
```
fgelcer, line 12: nitpick: 
```suggestion
# Copyright (c) 2023 ScyllaDB
```

temichus:
> Why it requires special event handler and not just counting on event publish? 

It can be tons of events, and I do think that we need to count all of them + I like that @aleksbykov implemented it as a separate module and does not touch the existing Events logic, because it is already complicated, and changing it may lead us to more complicated module and bugs

>Can you also elaborate why EventStatHandler saves events?

according to the code,  for loging purpose + reporting an issue(this is case for REACTOR_STALLED particulary)

fgelcer (commented):
for now, 2 little nitpick comments

sdcm/sct_events/setup.py
```diff
@@ -84,6 +86,7 @@ def stop_events_device(_registry: Optional[EventsProcessesRegistry] = None) -> N
         EVENTS_HANDLER_ID,
         EVENTS_ANALYZER_ID,
         EVENTS_MAIN_DEVICE_ID,
+        EVENTS_COUNTER_ID,
```
soyacz, line 89: I think it should go before `EVENTS_MAIN_DEVICE_ID`, possibly before EVENTS_HANDLER_ID.

sdcm/sct_events/event_counter.py
```diff
@@ -0,0 +1,192 @@
...
+                                        save_dir=self._save_file_path,
+                                        stats={})
+        self._counter_device.add_counter(self._id, event_stat_data)
+        self._counter_device.start_counter()
+
+    def stop_event_counter(self):
+        if counter_data := self._counter_device.get_counter(self._id):
+            self._statistics = counter_data.stats
+            self._counter_device.remove_counter(self._id)
+        self._counter_device.stop_counter()
```
soyacz, line 176: What's the point of having different counters if on `__exit__` we make  `EventsCounter` not to count?
Why do we need different counters? Maybe it would be enough to get initial count value on enter and get diff on exit instead adding different counters?

aleksbykov:
> Why it requires special event handler and not just counting on event publish? Can you also elaborate why EventStatHandler saves events?

Originally we need count only one Reactor stall event and only during specified operation. Specified operation could run several times one by one. This need for performance latency test with operations. But in future list of event could be extended for example with Kernal Stack events. Once operation finished, we don't need to count any events any more. Because number of events could large, to avoid memory overloading by main process, i decided to run such counter in another process, so if anything go wrong not to kill test itself.
Reactor stalls events will be decoded in batch for investigation. That why, they need to be saved to file. But EventsStatHandler don't require that and was use in testing puposes. I will remove it

sdcm/report_templates/results_latency_during_ops_short.html
aleksbykov: Done

sdcm/sct_events/event_counter.py
aleksbykov: done

sdcm/sct_events/event_counter.py
soyacz: I think start/stop counter is not only redundant in that case, but also may lead to errors when within `EventCounterContextManager` someone will open another `EventCounterContextManager` with different counter - then on exit will stop counting for all counters.

I think we could just drop `self._start_count` Event idea and count only if register contain counters, otherwise, skip any work.

soyacz:
> Because number of events could large, to avoid memory overloading by main process, i decided to run such counter in another process, so if anything go wrong not to kill test itself.

How events number can overload memory? if we just counting we don't increase memory over time...

aleksbykov:
> > Because number of events could large, to avoid memory overloading by main process, i decided to run such counter in another process, so if anything go wrong not to kill test itself.
> 
> How events number can overload memory? if we just counting we don't increase memory over time...

If we count many events in parallel, also for some  events we want additional operations as for Reactor stall we want to parse and collect additional info

sdcm/sct_events/event_counter.py
aleksbykov: > I think start/stop counter is not only redundant in that case, but also may lead to errors when within `EventCounterContextManager` someone will open another `EventCounterContextManager` with different counter - then on exit will stop counting for all counters.
> 

stop counting will happened , only if no any registered context managers will stay in _register. So if some on open and close counterevent_cm, it will not affect on others, and only latest closed counter_cm will stop counting

> I think we could just drop `self._start_count` Event idea and count only if register contain counters, otherwise, skip any work.

It is interesting idea, will check it now

roydahan (dismissed):

sdcm/sct_events/event_counter.py
aleksbykov: @soyacz , removed the _event. Added 2 more unit tests could take a look

roydahan (approved):

soyacz (approved):
LGTM

soyacz (commented):
@aleksbykov safe to merge or you want to test it on staging before?

aleksbykov:
@soyacz , i am running 2 staging jobs, will update you after they finished

aleksbykov:
> @soyacz , i am running 2 staging jobs, will update you after they finished

Bot jobs are passed. Regular longevity-4h(where no call to counter) and latency 650 GB with nemesis are passed

aleksbykov:
@roydahan , the problem happened for unit tests. it is happened, because branch-perf-v14 doesn't have this commit:
```
commit 9697300188dd5142428574624134048222648ddf
Author: Lukasz Sojka <lukasz.sojka@scylladb.com>
Date:   Fri Mar 10 11:32:05 2023 +0100

    feature(adaptive-timeouts): calculate timeouts based on node load
```
but if try to backport it , then another conflict happened with `nemesis.py`, if try to backport commit which was not backported for resolving issue in nemesis.py, this will build long chain of commits which were not backported to perf-v14

@roydahan , WDYT, if i prepare new pr explicitly for perf-v14 where resolve the unit-test conflicts?

roydahan:
Ok, please send a PR directly to v14.

PR #6303, opened 2023-06-27
refactor(append_scylla_args): Cleanup all different usages of it

roydahan:
We had many different usages of scylla_args which may cause us to miss some issues in scylla.
Since we support thresholds for warnings and errors of reactor stalls, it's ok to have them all as low as 25 ms in test_default. This was the main usage of changeing append_scylla_args. Other usages like adding expermintal flags aren't relevant anymore. So, we left with 3 main "versions" of this parameter:
1. default - should be used in all of our tests.
2. Performance - Reducing stall detector to 5 ms and remove the "enable-sstable-key-validation" which I suspect can hurt performance.
3. Special - for example, set memory to 4GB.


roydahan:
This should be an FYI for everyone who test scylla, just to be aware that if there is a new failure, it may be due to enabling a missing flag and necessarily a new regression.

test-cases/scylla-operator/kubernetes-scylla-upgrade.yaml
```diff
@@ -22,9 +22,10 @@ user_prefix: 'kubernetes-scylla-upgrade'
 #authenticator_user: 'cassandra'
 #authenticator_password: 'cassandra'
 
-# Does not work due to the lack of support, could be anabled after - https://github.com/scylladb/scylla-cluster-tests/issues/2747
-#append_scylla_args: '--blocked-reactor-notify-ms 500 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1'
-
+# TODO: add '--abort-on-seastar-bad-alloc' arg to the 'append_scylla_args' option when
+#       following operator bug gets fixed: https://github.com/scylladb/scylla-operator/issues/991
+#       '--blocked-reactor-notify-ms 100' cannot be set, because it gets set by operator itself
+append_scylla_args: '--abort-on-lsa-bad-alloc 1 --abort-on-internal-error 1 --abort-on-ebadf 1 --enable-sstable-key-validation 1'
```
vponomaryov, line 28: It is already defined in each of the K8S backends default config files.
So, if change this file then only by removing the red lines here.

soyacz (approved):

test-cases/scylla-operator/kubernetes-scylla-upgrade.yaml
roydahan: ok

roydahan:
I'll wait for ack from @fruch @fgelcer @aleksbykov and @temichus

temichus (approved):

fruch (approved):
LGTM

fruch:
@roydahan you still want this backported to v14 ?

roydahan:
No

PR #6319, opened 2023-07-03
fix(test_config): wait for logdir symlink removed

temichus:
added wait_for_removed inline fuctuion to make sure that symlink is removed before create it again

### DIFF
--- a/sdcm/test_config.py
+++ b/sdcm/test_config.py
@@ -98,6 +98,13 @@ class TestConfig(metaclass=Singleton):  # pylint: disable=too-many-public-method
             latest_symlink = os.path.join(base, "latest")
             if os.path.islink(latest_symlink):
                 os.remove(latest_symlink)
+
+                @retrying(n=10, sleep_time=0.5, allowed_exceptions=(AssertionError,))
+                def wait_for_removed():
+                    if os.path.islink(latest_symlink):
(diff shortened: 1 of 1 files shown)


sdcm/test_config.py
```diff
@@ -98,6 +98,13 @@ def make_new_logdir(cls, update_latest_symlink: bool, postfix: str = "") -> str:
             latest_symlink = os.path.join(base, "latest")
             if os.path.islink(latest_symlink):
                 os.remove(latest_symlink)
+
+                @retrying(n=10, sleep_time=0.5, allowed_exceptions=(AssertionError,))
```
fruch, line 102: if it's cause by multiple jobs running at the same time, this won't solve it

temichus: @fruch 
with this fix it will be 100% sure that it is "multiple jobs running" if it happens again

BTW did we support "multiple jobs running" for this case? probably not and the fix should be in the Jenkins level

fruch: in the past we were running a lot multiple jobs on the same builder, but then SCT runner was introduced
but artifact test doesn't use SCT runner.

might test hydra command should be logging, os shouldn't be logging into latest.

temichus: @fruch another possible fif where we do not touch latest. 
https://github.com/scylladb/scylla-cluster-tests/pull/6320

PR #6323, opened 2023-07-03
fix(k8s): do not run nemesis during prepare in 2 tenants longevity test

vponomaryov:
The reason is that the prepare load almost overloads the Scylla pods and any disruption to any of the Scylla clusters will cause loader failures.
So, to avoid it disable nemesis during the 'prepare' phase which runs only for bunch of minutes.


vponomaryov:
Example of the failure is here: https://argus.scylladb.com/test/8043d239-4bee-4ad1-b9e3-541f261a214d/runs?additionalRuns%5B%5D=753bad24-6ad4-4e5b-a776-c2ba88becd25

Monitoring:
Screenshot from 2023-07-03 19-03-54
Screenshot from 2023-07-03 19-04-01

fruch (approved):
LGTM

fruch:
Another option would be to ease the prepare load a bit.

But I don't think it's critical for us at this stage to do nemesis on prepare

soyacz (approved):

PR #6328, opened 2023-07-04
feature(nemesis): added ToggleAuditNemesis

soyacz:
New nemesis for testing Scylla audit capabilities. It turns Audit Log on when is not enabled and verifies few scenarios. At the end leaves some basic auditing on.
In case Audit Log is enabled on nemesis start, disable it and finish nemesis.

refs: https://github.com/scylladb/qa-tasks/issues/407


soyacz:
first test drive: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/audit-nemesis/2/

soyacz:
> first test drive: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/audit-nemesis/2/

Job failed - unfortunately without logs. Marked as draft before I get to conclusion&fix.

sdcm/nemesis.py
```diff
@@ -4413,6 +4419,77 @@ def disrupt_add_remove_mv(self):
...
+        if audit.is_enabled():
+            InfoEvent("Audit was enabled -> disabling it").publish()
+            audit.disable()
+            return
+
+        keyspaces_for_audit = [ks_cf.split(".")[0] for ks_cf in test_ks_cf_list]
+        InfoEvent("Enabling full audit").publish()
+        audit_config = AuditConfiguration(
+            store="table",
+            categories=["DCL", "DDL", "AUTH", "ADMIN", "DML", "QUERY"],
```
fruch, line 4446: I'm think we should make sure this nemesis ends, we don't leave audit with QUERY, it would break most of the test cases...

roydahan: Not sure what it will cause, we tested it in the past and it reduced the throughput significantly.

sdcm/nemesis.py
```diff
@@ -4413,6 +4419,90 @@ def disrupt_add_remove_mv(self):
...
+
+        if not (test_ks_cf_list := self.cluster.get_non_system_ks_cf_list(db_node=self.target_node)):
+            raise UnsupportedNemesis("Not found non-system keyspaces")
+
+        audit = Audit(self.cluster)
+
+        if audit.is_enabled():
+            InfoEvent("Audit was enabled -> disabling it").publish()
+            audit.disable()
+            return
```
juliayakovlev, line 4440: Maybe we need to set this nemesis as "skipped" and not "succeeded". And reason "Audit was enabled -> disabling it. Nemesis were not run". Something like this

sdcm/nemesis.py
```diff
@@ -4413,6 +4419,90 @@ def disrupt_add_remove_mv(self):
...
+            categories=["DCL", "DDL", "AUTH", "ADMIN", "DML", "QUERY"],
+            keyspaces=keyspaces_for_audit,
+            tables=[],
+        )
+        try:
+            audit.configure(audit_config)
+            keyspace_name = keyspaces_for_audit[0]
+            errors = []
+            audit_start = time.time()
+            time.sleep(300)  # Wait for audit log work for some time and not breaking Scylla
```
juliayakovlev, line 4462: I think 5 min it's so big. We can have the troubles to read from the audit table

sdcm/nemesis.py
```diff
@@ -4413,6 +4419,90 @@ def disrupt_add_remove_mv(self):
...
+
+            rows = audit.get_audit_log(from_timestamp=audit_start, category="DML", limit_rows=100)
+            if not rows:
```
juliayakovlev, line 4465: Because of limit of 100 rows we can miss the audit entries. Do you see that audit rows are appear immediately with our load?

sdcm/nemesis.py
soyacz: I think this is a good idea

sdcm/nemesis.py
soyacz: Yes, we found issues when having Audit with QUERY/DML on high traffic keyspaces anyway - redesigned the test to use separate ks only for Audit verification purposes.

sdcm/nemesis.py
soyacz: yes, changed that to be higher than the number of rows that are expected in audit

sdcm/nemesis.py
```diff
@@ -4417,6 +4423,106 @@ def disrupt_add_remove_mv(self):
...
+        if audit.is_enabled():
+            audit.disable()
+            raise UnsupportedNemesis("Audit was enabled -> disabling it")
+
+        audit_keyspace = "audit_keyspace"
+        keyspaces_for_audit = [audit_keyspace]
+        InfoEvent(f"Enabling full audit for keyspaces: {keyspaces_for_audit}").publish()
+        audit_config = AuditConfiguration(
+            store=store,
+            categories=["DCL", "DDL", "AUTH", "ADMIN", "DML", "QUERY"],
```
ptrsmrn, line 4450: In case audit is extended with more categories, like `OTHER`, `ERROR` and `PREPARE`, which Cassandra offers, this test won't be testing all of the categories anymore. I wonder if there's any way to always be sure we enable all categories of statements. Perhaps we could parse scylla's config parameters (either `scylla.yaml` or `--help` output) to see what categories are available, but I believe not all of them are exposed properly, we may miss or even hide some in the future, so this is also unreliable.

sdcm/nemesis.py
```diff
@@ -4417,6 +4423,106 @@ def disrupt_add_remove_mv(self):
...
+            self.tester.verify_stress_thread(cs_thread_pool=write_thread)
+            read_cmd = f"cassandra-stress read no-warmup cl=ALL n=1000 " \
+                       f" -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)" \
+                       f" keyspace={audit_keyspace}' -mode cql3 native -rate 'threads=1 throttle=1000/s'" \
+                       f" -pop seq=1..1000 -col 'n=FIXED(1) size=FIXED(128)' -log interval=5"
+            read_thread = self.tester.run_stress_thread(
+                stress_cmd=read_cmd, round_robin=True, stop_test_on_failure=False)
+            self.tester.verify_stress_thread(cs_thread_pool=read_thread)
+            InfoEvent(message='Verifying Audit table contents').publish()
+            rows = audit.get_audit_log(from_datetime=audit_start, category="DML", limit_rows=1100)
```
ptrsmrn, line 4475: I assume that we begin with a clear `audit.audit_log`, right? And we're fetching 1.1k rows, while inserting only 1k, just to make sure we really fetched all of the rows inserted in this session. But what if, by mistake, e.g. 1.2k rows have been inserted, 200 coming from `USE KS`. We'd remove these 200 extra rows, but because we only fetched 1.1 rows, we'd be left with 1100-200=900 rows, and 900 != 1000, so the next check would fail, but perhaps there's really 1k DML rows in audit, it's just that we haven't fetched all of them. I guess that the bottom line here is that we should maybe not limit the number of rows fetched.
OTOH, if `audit.audit_log` is not empty, we'd be fetching garbage from previous runs? - That's why assume it is empty :)

sdcm/nemesis.py
```diff
@@ -4417,6 +4423,106 @@ def disrupt_add_remove_mv(self):
...
+            # filter out USE keyspace rows due to https://github.com/scylladb/scylla-enterprise/issues/3169
+            rows = [row for row in rows if not row.operation.startswith("USE")]
```
ptrsmrn, line 4477: If we expect only DML, perhaps we should filter out every other category, not only the USE KS?

sdcm/nemesis.py
ptrsmrn: Also, maybe we should log the number of rows fetched for debugging purposes.

sdcm/nemesis.py
soyacz: I think the only way is to stay tuned to audit changes and adapt accordingly - even changing nemesis for the new cases. E.g. currently here we verify only DML, QUERY and DDL. We can think of extending it also to other categories - but this is out of the scope now.

sdcm/nemesis.py
soyacz: so, the idea to overcome these issues is to first fetch only ones that are after `from_datetime` - which is right before c-s start. The second, you are right in case of 200 `USE KS` statements we might miss valid rows, possibly we can drop it. From the other hand, assuming it is working now, when we start to see 200 `USE KS` won't it be some kind of regression than suddenly we started to log so many of these?

sdcm/nemesis.py
soyacz: `USE KS` is DML category, that's why it is filtered later manually like this

sdcm/nemesis.py
ptrsmrn: IMO we could parse scylla.yaml, because I believe it's dev's responsibility to keep it up-to-date, but I guess we can also just change the nemesis if that happens.

sdcm/nemesis.py
soyacz: > Also, maybe we should log the number of rows fetched for debugging purposes.

we do it in `get_audit_log` method

soyacz:
looks like it works with syslog: https://argus.scylladb.com/test/a3398e28-90ad-4641-a860-c9c5c1ad6cb5/runs?additionalRuns[]=b255d86f-7e2b-422a-9d7d-c9e7703041f9
the only issue was with timeout during dropping table - I added `USING TIMEOUT 5m` to give it some more time.

sdcm/nemesis.py
```diff
@@ -4417,6 +4423,106 @@ def disrupt_add_remove_mv(self):
         finally:
             self.unset_current_running_nemesis(cql_query_executor_node)
 
+    def disrupt_toggle_audit(self, store: AuditStore):
```
fruch, line 4426: this is problematic, we don't support very well parameter to disrupt function, i.e. it would always gonna be empty (when used in any Nemesis class, beside ToggleAuditNemesis.

so or we set a default value here to syslog, or we refactor ro multiple disrupt functions calling other function

sdcm/nemesis.py
```diff
@@ -4417,6 +4423,106 @@ def disrupt_add_remove_mv(self):
...
+            Enable audit log with all categories and user keyspaces (if audit already enabled, disable it and finish the Nemesis),
+            verify audit log content,
+            reduce categories by excluding DML and QUERY,
+            verify DDL are logged in audit log correctly. Leaves audit log enabled this way.
+        """
+        if not self.target_node.is_enterprise:
+            raise UnsupportedNemesis("Auditing feature is only supported by Scylla Enterprise")
+
+        if store == "syslog" and self._is_it_on_kubernetes():
+            raise UnsupportedNemesis("syslog store is not supported on Kubernetes #scylla-operator/1299")
```
fruch, line 4437: niptick: the issue isn't in github format

```
scylladb/scylla-operator#1299
```

sdcm/audit.py
```diff
@@ -0,0 +1,229 @@
...
+def get_audit_log_rows(node,  # pylint: disable=too-many-locals
+                       from_datetime: Optional[datetime] = None,
+                       category: Optional[AuditCategory] = None,
+                       operation: Optional[str] = None,
+                       limit_rows: int = 1000
+                       ) -> List[AuditLogRow]:
+    with node.open_system_log(on_datetime=from_datetime) as log_file:
+        found_rows = 0
+        for line in log_file:
+            if '!NOTICE | scylla-audit' in line:
```
fruch, line 120: I would be careful to depend on the log line structure that much, it might change, for example on VMs vs on docker/k8s or artifacts test when we use ssh for logs.

I would look for the level  "NOTICE" and for the source/component: "scylla-audit"

fruch: Anyhow this type of code kind of calls for a unittest... (reading logs is always behaves a bit differently then we anticipate)

fruch (changes requested):
- parameter to disrupt function
- unittest for logs parsing

sdcm/audit.py
fruch: this code part might explian why it was seem to not work in: 
https://github.com/scylladb/scylla-operator/issues/1299
from my POV, we can set it to not work at all in k8s for now.

I would recommend testing first with docker backend, and only then with k8s. (anyhow both are nice to have, and not a priority for now)

sdcm/nemesis.py
soyacz: refactor to multiple disrupt functions is better - different nemesis was planned for table so later we would need to rename and break ES stats and possibly some other areas.

sdcm/audit.py
soyacz: improved and added unit tests

sdcm/nemesis.py
fruch: We'll adapt once those would be available, as we do for any new feature / bug fixes. i.e. once we'll fix/add the dtest for this feature, we'll need to fix/add those to SCT as well.

sdcm/nemesis.py
```diff
@@ -4414,6 +4420,110 @@ def disrupt_add_remove_mv(self):
...
+        """
+            Enable audit log with all categories and user keyspaces (if audit already enabled, disable it and finish the Nemesis),
+            verify audit log content,
+            reduce categories by excluding DML and QUERY,
+            verify DDL are logged in audit log correctly. Leaves audit log enabled this way.
+        """
+        if not self.target_node.is_enterprise:
+            raise UnsupportedNemesis("Auditing feature is only supported by Scylla Enterprise")
+
+        if store == "syslog" and self._is_it_on_kubernetes():
```
fruch, line 4436: nitpick: skip also the docker backend

fruch (dismissed):
LGTM

fruch:
let do a last run, to see everything works, and then we'll merge

soyacz:
> let do a last run, to see everything works, and then we'll merge

Small fixes were required + rebased. Works: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/audit-nemesis/20/ (job aborted when I saw it works fine)

fruch (approved):
LGTM

PR #1033, opened 2019-06-12
Check `nodetool upgradesstables` works after all nodes upgraded

fruch:
## PR pre-checks (self review)

- [x] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle)
- [x] I gave variables/functions meaningful self-explanatory names
- [x] I didn't leave commented-out/debugging code
- [x] I didn't copy-paste code
- [x] I added the relevant `backport` labels
- [ ] ~~New configuration option are added and documented (in `sdcm/sct_config.py`)~~
- [ ] ~~I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)~~
- [x] All new and existing unit tests passed (`hydra unit-tests`)
- [ ] ~~I have updated the Readme/doc folder accordingly (if needed)~~

PR #6335, opened 2023-07-06
fix(upgrade test): do not specfiy -a when "nodetools upgradesstables"

### LINKED ISSUE #6334: `nodetool upgradesstables` takes almost 6 hours to complete
see https://github.com/scylladb/scylladb/issues/14548 for more details.

tchaikov:
`-a` is the short form of `--include-all-sstables`, which instruct the tool and in turn scylla to compact *all* sstables no matter if their format version is identical to the latest format version. at the time of writing, it is "me". by default, nodetool does exclude sstables of current version from being compacted.

in https://github.com/scylladb/scylladb/issues/14548, we found that scylla took almost 6 hours to complete this command. and it compacted more than 873GiB data in total. this behavior is expected. but the use case is not very popular. as the whole point of "nodetools upgradesstables" is to upgrade the sstable format from an old one the the latest one supported by scylla.

so, in this change, we just drop the `-a` option from the `nodetool upgradesstables` command line.

Fixes #6334

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -379,7 +379,7 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
             node.remoter.sudo(
                 f"cp {main_dir / filename} {main_dir / subdir / filename}")
 
-            node.run_nodetool(sub_cmd="upgradesstables", args="-a")
+            node.run_nodetool(sub_cmd="upgradesstables")
         if queue:
             queue.put(upgradesstables_available)
             queue.task_done()


roydahan (approved):

roydahan:
@fgelcer you can give it a try in your latency test which should be very quick.

fruch:
I remember vaguely that when I did it on #1033, it what the instruction on the upgrade docs said...

but now the docs doesn't mention this at all anymore...

regardless we do have a use case when this -a is needed, and that's on disable/enable EaR

fruch (approved):

fruch:
> I remember vaguely that when I did it on #1033, it what the instruction on the upgrade docs said...
> 
> but now the docs doesn't mention this at all anymore...
> 
> regardless we do have a use case when this -a is needed, and that's on disable/enable EaR

BTW it's almost the first feature I've implemented on SCT :)

fgelcer:
> @fgelcer you can give it a try in your latency test which should be very quick.

@roydahan , let's merge the upgrade latency test, so we can have a 1st official run with it

Issue #6236, opened 2023-06-11
AddRemoveDc breaks when simulated racks are being used


fruch:
## Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Since introduction of #6111, seems like the new node in `AddRemoveDc` is not added in a new DC
```
2023-06-11 08:42:15.917: (DisruptionEvent Severity.ERROR) period_type=end event_id=88b74d12-5e8b-4a8a-a02d-dcb1c1213217 duration=30m11s: nemesis_name=AddRemoveDc target_node=Node longevity-cdc-100gb-4h-master-db-node-d1b683b2-5 [54.229.157.143 | 10.4.2.208] (seed: True) errors=new datacenter was not registered
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4425, in wrapper
result = method(*args[1:], **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4009, in disrupt_add_remove_dc
assert new_dc_list, "new datacenter was not registered"
AssertionError: new datacenter was not registered
```

new node `10.4.2.123` is expected to be in `eu-west-1 _nemesis_dc`, but it's not:
```
< t:2023-06-11 08:31:15,279 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > Datacenter: eu-west-1
< t:2023-06-11 08:31:15,281 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > =====================
< t:2023-06-11 08:31:15,281 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > Status=Up/Down
< t:2023-06-11 08:31:15,282 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > |/ State=Normal/Leaving/Joining/Moving
< t:2023-06-11 08:31:15,303 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > --  Address     Load       Tokens       Owns    Host ID                               Rack
< t:2023-06-11 08:31:15,337 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > UN  10.4.2.48   4.12 GB    256          ?       98eb8dae-c1ee-4dd1-b523-25344b6b33e9  RACK2
[...]

PR #6347, opened 2023-07-11
fix(AddRemoveDcNemesis): fix on simulated racks

soyacz:
Problem is with detection - which is based on `scylla.yaml` taken from new node which has default setting - and failing to detect `GossipingPropertyFileSnitch` is used.

Use test param `endpoint_snitch` which should be always valid (even when using simulated racks) for detection of endpoint snitch.

fixes: #6236


fruch (approved):
LGTM

vponomaryov (approved):
LGTM

PR #6349, opened 2023-07-11
feature(cassandra-stress): rack aware support

fruch:
`RackAwareRoundRobinPolicy` was introduced in scylladb/java-driver#227 and now we'll support passing it to c-s automatically every time we use multiple racks setup, the loader would be spread on multiple AZs, and we should pass each loader the corresponding rack, similar to how we are doing it for DCs

Ref: https://github.com/scylladb/scylla-tools-java/pull/335
Ref: https://github.com/scylladb/java-driver/pull/227

### Testing:
- [x] - simulated Racks - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-cdc-100gb-4h-test/7/
- [x] - base case for comparing (2023.1) - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-cdc-100gb-4h-test/8/
- [x]  - run of case with the fix - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-cdc-100gb-4h-test/14/
- [x] - Multi-DC real AZ - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-multi-dc-rack-aware-test/12/

### DIFF
--- a/docker/cassandra-stress/Dockerfile-src
+++ b/docker/cassandra-stress/Dockerfile-src
@@ -9,10 +9,6 @@ ENV GIT_FORK=${GIT_FORK:-https://github.com/scylladb/scylla-tools-java}
 RUN git clone -b ${BRANCH} --single-branch ${GIT_FORK}
 RUN apt -y update && apt-get -y install ant
 
-# revert of cassandra driver (without shard-aware)
-COPY .gitconfig /root/
-RUN cd scylla-tools-java && git revert 1deceb8a605e563f01ef1821d48316cea766da31 && git revert 87a98f0a85c9db8a9c7602e66d3d2cca56903cdf
-
 # build
 RUN cd scylla-tools-java && ant
 
--- a/docker/cassandra-stress/README.md
+++ b/docker/cassandra-stress/README.md
@@ -6,16 +6,16 @@ export DEB_LIST_URL=https://s3.amazonaws.com/downloads.scylladb.com/unstable/scy
 export CS_DOCKER_IMAGE=scylladb/hydra-loaders:cassandra-stress-$(date +'%Y%m%d')
 docker build . -t ${CS_DOCKER_IMAGE} -f Dockerfile-deb --build-arg DEB_LIST_URL=$DEB_LIST_URL
 docker push ${CS_DOCKER_IMAGE}
-echo "${CS_DOCKER_IMAGE}" > image
 ```
 
-### Building from source without shard aware driver (slower)
+### Building from fork source (slower)
+
+this is just example of such usage, you should point it to the fork/branch you need
 
 ```
-export BRANCH=branch-5.0
+export BRANCH=use_rack_aware
+export GIT_FORK=https://github.com/sylwiaszunejko/scylla-tools-java
 export CS_DOCKER_IMAGE=scylladb/hydra-loaders:cassandra-stress-${BRANCH}-$(date +'%Y%m%d')
-docker build . -t ${CS_DOCKER_IMAGE} -f Dockerfile
-docker build . -t ${CS_DOCKER_IMAGE} -f Dockerfile-src --build-arg BRANCH=${BRANCH}
+docker build . -t ${CS_DOCKER_IMAGE} -f Dockerfile-src --build-arg BRANCH=${BRANCH} --build-arg GIT_FORK=${GIT_FORK}
 docker push ${CS_DOCKER_IMAGE}
-echo "${CS_DOCKER_IMAGE}" > image
 ```
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3213,6 +3213,14 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
 
         return rack_names_mapping
 
+    def get_nodes_per_datacenter_and_rack_idx(self, db_nodes: list[BaseNode] | None = None):
+        db_nodes = db_nodes if db_nodes else self.nodes
+        nodes_mapping = {}
+        for (region, rack), nodes in self.nodes_by_racks_idx_and_regions(nodes=db_nodes).items():
+            nodes_mapping[(region, rack)] = nodes
+
+        return nodes_mapping
+
     def send_file(self, src, dst, verbose=False):
         for loader in self.nodes:
             loader.remoter.send_files(src=src, dst=dst, verbose=verbose)
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2599,7 +2599,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             for dc, dc_status in statuses.items():
                 for node_ip, node_properties in dc_status.items():
                     if node := node_ip_map.get(node_ip):
-                        nodes_status[node] = {'status': node_properties['state'], 'dc': dc}
+                        nodes_status[node] = {'status': node_properties['state'],
+                                              'dc': dc, 'rack': node_properties['rack']}
                     else:
                         if node_ip:
                             LOGGER.error("Get nodes statuses. Failed to find a node in cluster by IP: %s", node_ip)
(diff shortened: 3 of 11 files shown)


fruch:
opening it as draft, since the new c-s with the support isn't yet merged

sdcm/cluster.py
```diff
@@ -3241,6 +3250,16 @@ def get_datacenter_name_per_region(self, db_nodes=None):
 
         return datacenter_name_per_region
 
+    def get_rack_names(self, db_nodes=None):
+        rack_names_mapping = {}
+        for rack, nodes in self.nodes_by_racks(nodes=db_nodes).items():
+            if status := nodes[0].get_nodes_status():
```
soyacz, line 3256: Can't we get_nodes_status just once before for loop?

sdcm/cluster.py
```diff
@@ -3241,6 +3250,16 @@ def get_datacenter_name_per_region(self, db_nodes=None):
 
         return datacenter_name_per_region
 
+    def get_rack_names(self, db_nodes=None):
```
soyacz, line 3253: I think this is not going to work, if rack names in different dc's differ.

We should take racks per DC and select one from it

soyacz (changes requested):

sdcm/cluster.py
fruch: I guess we can, I've copied the logic from the `per_region_dc` function

sdcm/cluster.py
fruch: the idea is that we take all of the rack names, and try to match them with the loader rack name

so you are saying it would be something like

eu-west-1 
0: eu-west-1-a
1: eu-west-1-b
2: eu-west-1-c

us-east-1
0: us-east-1-a
1: us-east-1-b


so mapping base on the rack variable, which is just a number, would be enough ?

soyacz: yes, even in unit test second dc uses `2a` while first `1a`.
But now I see you fixed it.

unit_tests/test_cluster.py
```diff
@@ -507,6 +508,33 @@ def test_datacenter_name_per_region(self):  # pylint: disable=no-self-use
...
+                          "==================",
+                          "Status=Up/Down",
+                          "|/ State=Normal/Leaving/Joining/Moving",
+                          "--  Address   Load       Tokens       Owns    Host ID                               Rack",
+                          "UN  10.1.59.34    21.71 GB   256          ?       e5bcb094-e4de-43aa-8dc9-b1bf74546346  2a"
+                          ]
+                         )
+        node1 = NodetoolDummyNode(resp=resp, myregion="east-us", myname="10.0.59.34", myrack=1)
+        node2 = NodetoolDummyNode(resp=resp, myregion="east-us", myname="10.0.198.153", myrack=1)
+        node3 = NodetoolDummyNode(resp=resp, myregion="west-us", myname='10.1.59.34', myrack=2)
```
soyacz, line 529: IIUC to be precise, in such configuration, all the nodes should have `rack=0` - because in single rack scenarios we use only this index.
You could extend this test and make second DC to have 2 racks (e.g.:
```node4 = NodetoolDummyNode(resp=resp, myregion="west-us", myname='10.1.59.34', myrack=1)
[...]
assert datacenter_name_per_region == {('east-us', '1'): '1a', ('west-us', '1'): '2a', ('west-us', '2'): '2a'}
```

btw. `datacenter_name_per_region` rename to `rack_names_per_datacenter_and_rack_idx` or smthing like this

soyacz (approved):

unit_tests/test_cluster.py
fruch: changed the name, I don't think we need more unittest, the cases are the same, and it would work.

anyhow I would limit it to case of that we have more then on rack

sdcm/cluster.py
```diff
@@ -3231,6 +3232,14 @@ def nodes_by_region(self, nodes=None) -> dict:
             grouped_by_region[node.region].append(node)
         return grouped_by_region
 
+    def nodes_by_racks_idx_and_regions(self, nodes: list[BaseNode] | None = None) -> dict:
+        """:returns {availability_zone: [list of nodes]}"""
```
soyacz, line 3236: outdated comment

defaults/test_default.yaml
```diff
@@ -206,7 +206,7 @@ stress_image:
   ndbench: 'scylladb/hydra-loaders:ndbench-jdk8-20210720'
   ycsb: 'scylladb/hydra-loaders:ycsb-jdk8-20220918'
   nosqlbench: 'scylladb/hydra-loaders:nosqlbench-4.15.49'
-  cassandra-stress: '' # default would be same version as scylla under test
+  cassandra-stress: 'scylladb/scylladb-releng:5.4.0-dev-0.20230711.e1ef3b2ced5d-x86_64' # default would be same version as scylla under test
```
soyacz, line 209: temporary or we'll live with that until i'ts on scylla master?

jenkins-pipelines/longevity-counters-multidc.jenkinsfile
```diff
@@ -5,7 +5,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 
 longevityPipeline(
     backend: 'aws',
-    availability_zone: 'a',
+    availability_zone: 'a,b,c',
```
soyacz, line 8: so, we rollback this test to be true multi-az?

sdcm/utils/docker_remote.py
```diff
@@ -100,7 +100,7 @@ def check_spot_termination(self):
 
     @property
     def region(self):
-        return "docker"
+        return "eu-north-1"
```
soyacz, line 103: why?

defaults/test_default.yaml
fruch: temporary for testing, I'll take it out before merge

jenkins-pipelines/longevity-counters-multidc.jenkinsfile
fruch: no, it's for testing, I wanted to see it working in real AZs (not the simulated)

sdcm/utils/docker_remote.py
fruch: since now the logic of c-s is calling the newly added function for getting rack, this region needs to be align with the loader for it to work.
so both parts of the test needs to be in sync for that logic to work correctly

sdcm/cluster.py

defaults/test_default.yaml
fruch: removed

fruch:
integration tests would be addressed in #7043

fruch:
Results for the latest runs are showing the rack aware is working as expected
- new test case introduce multi-dc with multi-az (rack)
- current case that use multi-az are also working with this change

@roydahan @soyacz @juliayakovlev can you give a final review ?

docker/cassandra-stress/README.md
```diff
@@ -12,10 +12,10 @@ echo "${CS_DOCKER_IMAGE}" > image
 ### Building from source without shard aware driver (slower)
 
 ```
-export BRANCH=branch-5.0
+export BRANCH=use_rack_aware
+export GIT_FORK=https://github.com/sylwiaszunejko/scylla-tools-java
```
soyacz, line 16: is there some PR that we wait for? Worth to mention when we can switch back to scylla's repo

docker/cassandra-stress/Dockerfile-src
```diff
@@ -10,8 +10,8 @@ RUN git clone -b ${BRANCH} --single-branch ${GIT_FORK}
 RUN apt -y update && apt-get -y install ant
 
 # revert of cassandra driver (without shard-aware)
-COPY .gitconfig /root/
-RUN cd scylla-tools-java && git revert 1deceb8a605e563f01ef1821d48316cea766da31 && git revert 87a98f0a85c9db8a9c7602e66d3d2cca56903cdf
+# COPY .gitconfig /root/
```
soyacz, line 13: worth to mention what must happen to bring it back. If not, better remove.

soyacz (dismissed):
besides few clarifications, LGTM

Do we have tests that verify rack awareness with c-s somewhere (dtest, or one in tools/driver test)? E.g. by verifying metrics if client's get data only from specific rack, so we know this feature works at all and there are no regressions (longevity will not verify it)

docker/cassandra-stress/README.md
fruch: it's example of usage from fork.

the common case is running from scylla image, and then it's from the formal repos

those build are mostly for testing, or when specific job would need a specific version (i.e. like perf tests)

docker/cassandra-stress/Dockerfile-src
fruch: removed it

PR #6357, opened 2023-07-13
fix(nemesis): remove disrupt_remove_service_level_while_load nemesis

juliayakovlev:
This old nemesis that was added before all SLA nemeses. The problem with this nemesis that main service level is deleted while running that cause to c-s load failure.
There is new `disrupt_replace_service_level_using_drop_during_load` nemesis that perform same test.

It was found here:
https://jenkins.scylladb.com/job/enterprise-2022.2/job/Reproducers/job/longevity-sla-100gb-4h-tes/13/


juliayakovlev:
@fruch @roydahan 
There are 2 jobs. Do we need to delete or stop its triggering?

roydahan:
I don't see a reason why to remove this nemesis.
Let's discuss it before we continue here.

juliayakovlev:
After re-checking:
Service level "sl200_2" was deleted during `disrupt_remove_service_level_while_load` nemesis. New SL "sl200_3" was created instead of it and attached to the role that cassandra-stress load was run with.
The problem that new added service level "sl200_3" was not propagated to 3 nodes from 6 (looks like https://github.com/scylladb/scylla-enterprise/issues/948). But it should not cause to the load failure.
Close this PR.

PR #6358, opened 2023-07-13
fix(sshd_config): enable keep-alive configuration

fruch:
Since we are running on some cases into disconnection of ssh during some nodetool operation, it was suggested that we'll add this to `sshd_config`:

```
ClientAliveInterval 60
ClientAliveCountMax 10
```

Ref: https://github.com/scylladb/scylladb/issues/14004

### DIFF
--- a/sdcm/provision/common/utils.py
+++ b/sdcm/provision/common/utils.py
@@ -118,6 +118,8 @@ def configure_sshd_script():
     sed -i "s/#MaxSessions \(.*\)$/MaxSessions 1000/" /etc/ssh/sshd_config || true
     sed -i "s/#MaxStartups \(.*\)$/MaxStartups 60/" /etc/ssh/sshd_config || true
     sed -i "s/#LoginGraceTime \(.*\)$/LoginGraceTime 15s/" /etc/ssh/sshd_config || true
+    sed -i "s/#ClientAliveInterval \(.*\)$/ClientAliveInterval 60/" /etc/ssh/sshd_config || true
+    sed -i "s/#ClientAliveCountMax \(.*\)$/ClientAliveCountMax 10/" /etc/ssh/sshd_config || true
     """)
 
 


sdcm/provision/common/utils.py
```diff
@@ -118,6 +118,8 @@ def configure_sshd_script():
     sed -i "s/#MaxSessions \(.*\)$/MaxSessions 1000/" /etc/ssh/sshd_config || true
     sed -i "s/#MaxStartups \(.*\)$/MaxStartups 60/" /etc/ssh/sshd_config || true
     sed -i "s/#LoginGraceTime \(.*\)$/LoginGraceTime 15s/" /etc/ssh/sshd_config || true
+    sed -i "s/#ClientAliveInterval \(.*\)$/ClientAliveInterval 60/" /etc/ssh/sshd_config || true
+    sed -i "s/#ClientAliveCountMax \(.*\)$/ClientAliveCountMax 10/" /etc/ssh/sshd_config || true
```
xemul, line 122: This will make sshd live-ping _all_ ssh session, while my initial suggestion was to make _client_ turn on live-pings on those sessions that are considered long and prone to connection aborts. But I've no strong preference of my plan against this one. Since its SCT-only thing, likely this doesn't change much

xemul (approved):

sdcm/provision/common/utils.py
fruch: the client is using python-ssh(libssh), and I'm not sure it's reading the configuration as one might expect

doing it on the server side is a bit clear that it would do what's says on the cover.

if you think the client side it more relevant, I can dig further on it.

xemul: No, let's try with server-side configuration and see how it goes

vponomaryov (approved):
LGTM

Issue #6181, opened 2023-05-23
"start_and_interrupt_decommission_streaming" waits static and too small amount of time


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [ ] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: 2023.1.0~rc5-20230429.a47bcb26e42e

### Logs
* test_id: da817906-6710-4a13-b5d7-b4ee9ef7c87e
* job log: [Argus](https://argus.scylladb.com/test/a388b58f-ff99-4bdb-b813-d07a2111d1d1/runs?additionalRuns[]=da817906-6710-4a13-b5d7-b4ee9ef7c87e), [CI job](https://jenkins.scylladb.com/job/enterprise-2023.1/job/longevity/job/longevity-large-partition-4days-test/3)

### Description

The `start_and_interrupt_decommission_streaming` method in the `nemesis.py` module picks up random stage of the decommission
process as trigger for the next step:

```
        terminate_patterns = self.target_node.raft.get_log_patterns_for_decommission_abort()            
        self.use_nemesis_seed()                                                                         
        terminate_pattern = random.choice(terminate_patterns)
```

Where the values available for random pick up are following:
```
        return ["DECOMMISSIONING: unbootstrap starts",                                                 
                "DECOMMISSIONING: unbootstrap done",                                                   
                "becoming a group 0 non-voter",                                                        
                "became a group 0 non-voter",                                                          
                "leaving token ring",                                                                  
                "left token ring",                                                                     
                "Finished token ring movement"]
```

Each of them will be awaited 1200 at max.
In current case was picked up following one:
```
< t:2023-05-19 11:44:02,356 f:nemesis.py      l:3442 c:sdcm.nemesis         p:DEBUG > sdcm.nemesis.SisyphusMonkey: Reboot node after log message: 'left token ring'
```

Decommission start:
```
2023-05-19T11:44:03+00:00 longevity-large-partitions-4d-2023--db-node-da817906-8     !INFO | scylla[7273]:  [shard 0] storage_service - DECOMMISSIONING: starts
```

The expected message:
```
2023-05-19T12:22:46+00:00 longevity-large-partitions-4d-2023--db-node-da817906-8     !INFO | scylla[7273]:  [shard 0] storage_service - decommission[c3810463-df5c-471b-b42d-a81d93cad59d]: left token ring
```

[...]

PR #6360, opened 2023-07-13
fix(raft_operation): Timeout for log message depend on appearence in log

aleksbykov:
Decommission could take different time depend on instance, size of dataset
number of nodes. Decommission streaming error randomly choose log line after
which decommission process should be aborted. Default timeout to wait
log message could be not enough to wait for messages from the end of
decommission.

Split messages for start and end process and return with method
the expected timeout.

Issue #6627
Issue #6181

- [ ] ~New configuration option are added and documented (in `sdcm/sct_config.py`)~
- [ ] ~I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)~
- [ ] ~I have updated the Readme/doc folder accordingly (if needed)~

### DIFF
--- a/sdcm/utils/raft/__init__.py
+++ b/sdcm/utils/raft/__init__.py
@@ -1,7 +1,9 @@
 import contextlib
 import logging
+import random
 
-from typing import Protocol
+from enum import Enum
+from typing import Protocol, NamedTuple, Mapping, Iterable
 from collections import namedtuple
 
 from sdcm.sct_events.database import DatabaseLogEvent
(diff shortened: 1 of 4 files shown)


sdcm/utils/raft/__init__.py
```diff
@@ -18,6 +18,12 @@ class Group0MembersNotConsistentWithTokenRingMembersException(Exception):
 class RaftFeatureOperations(Protocol):
     _node: 'BaseNode'
 
+    expected_timeout_begin_message = 1200
+    expected_timeout_end_message = 7200
```
fruch, line 22: where those times are coming from ?

sdcm/utils/raft/__init__.py
```diff
@@ -40,8 +46,32 @@ def clean_group0_garbage(self, raise_exception: bool = False) -> None:
     def diff_token_ring_group0_members(self) -> list[str]:
         ...
 
+    def get_log_patterns_for_decommission_abort(self) -> list[str]:
+        return self.decommission_begin_messages + self.decommission_end_messages
+
+    def get_message_waiting_timeout(self, message):
```
fruch, line 52: I don't like this API which is based on strings which defined in too different places... 
it will go out of sync...

sdcm/utils/raft/__init__.py
vponomaryov: In the referenced bug report the decommission operation took about 40 minutes.
So, `1h` timeout for the `end` messages look ok.
But `20 minutes` for the `begin` one is too much. It must be much smaller. At least `5m`...

aleksbykov:
Job with 1 nemesis passed latest 4 times: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/nemesis-clone/job/longevity-5gb-1h-BootstrapStreamingErrorNemesis-aws-test/
job for 3 hours with single DecommissionStreamErr nemesis passed for aws, marked as pass for gce. all nemesis passed, the job failed due to c-s command stucked.
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-1tb-7days-2Disruptive-Nemesises/101
Running longer tests

aleksbykov:
@vponomaryov @soyacz @fruch could you review please

fruch (dismissed):
LGTM

But first fix the precommit issue

sdcm/utils/raft/__init__.py
```diff
@@ -16,8 +18,58 @@ class Group0MembersNotConsistentWithTokenRingMembersException(Exception):
...
+    "gce": {LogPosition.BEGIN: 300, LogPosition.END: 3600},
+    "azure": {LogPosition.BEGIN: 1200, LogPosition.END: 7200},
+}
+
+ABORT_DECOMMISSION_LOG_PATTERNS: Iterable[MessagePosition] = [
+    MessagePosition("DECOMMISSIONING: unbootstrap starts", LogPosition.BEGIN),
+    MessagePosition("DECOMMISSIONING: unbootstrap done", LogPosition.END),
+    MessagePosition("becoming a group 0 non-voter", LogPosition.END),
+    MessagePosition("became a group 0 non-voter", LogPosition.END),
+    MessagePosition("leaving token ring", LogPosition.END),
```
soyacz, line 52: Is it possible including log delays and and processing time & triggering action time to catch the moment of 'leaving token ring' before 'left token ring'
Similar question is for more places, I wonder if these precise moments are useful in this case.

soyacz (approved):
my comment is general one, not actually related to this refactor.
LGTM

fruch (approved):

fruch:
Let's give it a week or two on master before backporting

fruch:
@temichus  @aleksbykov the backport for 2023.1/5.2 isn't clean, please open PRs for those, if this is needed for those branches

Issue #6362, opened 2023-07-13
Rolling upgrade AMI tests do not search images in the releng AWS account


ilya-rarov:
## Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

In the job `rolling-upgrade-ami-arm-test` SCT failed while performing `SetUp`.
At some point, `ec2_ami_get_root_device_name` was executed - it looked for the image `ami-024b197c962502cad`, but could not find it and created an empty Image object.
Later, while trying to get the value of the property `root_device_name` of this object the code failed with the error:
```
06:04:54  ERROR: test_rolling_upgrade (upgrade_test.UpgradeTest)
06:04:54  Upgrade half of nodes in the cluster, and start special read workload
06:04:54  ----------------------------------------------------------------------
06:04:54  Traceback (most recent call last):
06:04:54    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 164, in wrapper
06:04:54      return method(*args, **kwargs)
06:04:54    File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 119, in inner
06:04:54      res = func(*args, **kwargs)
06:04:54    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 794, in setUp
06:04:54      self.init_resources()
06:04:54    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 1818, in init_resources
06:04:54      self.get_cluster_aws(loader_info=loader_info, db_info=db_info,
06:04:54    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 1195, in get_cluster_aws
06:04:54      init_db_info_from_params(db_info, params=self.params, regions=regions)
06:04:54    File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/aws_utils.py", line 250, in init_db_info_from_params
06:04:54      root_device = root_device if root_device else ec2_ami_get_root_device_name(
06:04:54    File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/aws_utils.py", line 359, in ec2_ami_get_root_device_name
06:04:54      if image.root_device_name:
06:04:54    File "/usr/local/lib/python3.10/site-packages/boto3/resources/factory.py", line 386, in property_loader
06:04:54      return self.meta.data.get(name)
[...]

PR #6366, opened 2023-07-16
fix(get_scylla_ami_versions): filter debug versions

fruch:
In PR #5502 filter was introduced to `get_branched_ami` but now seems like we case of release like versions which are debug versions (and doesn't get shared
with our account), and we need to ignroe them as well.

Fixes: #6362


fruch:
@yaronkaikov since we do we have debug image that looks like release ? and should we expect them also on releases branches ? (i.e. show we backport this to branches ?)

yaronkaikov:
> @yaronkaikov since we do we have debug image that looks like release ? and should we expect them also on releases branches ? (i.e. show we backport this to branches ?)

we may have debug in releases, but it's not that common, i suggest backporting

fgelcer (approved):
LGTM

ilya-rarov (approved):
LGTM

PR #6377, opened 2023-07-18
fix(sct_runner): fix create-runner-image on GCP

enaydanov:
`hydra create-runner-image` failed to run on GCP after #6137

The instance type is selected based on test_duration here: https://github.com/scylladb/scylla-cluster-tests/blob/f77e6b3d0bada938c74c7f495fc14c14d57be761/sdcm/sct_runner.py#L405

and doesn't need to be overwritten in the `_create_instance(...)` method

### DIFF
--- a/sdcm/sct_runner.py
+++ b/sdcm/sct_runner.py
@@ -742,7 +742,7 @@ class GceSctRunner(SctRunner):  # pylint: disable=too-many-instance-attributes
                                  auto_delete=True)]
 
         instance = create_instance(project_id=self.project_name, zone=region_az,
-                                   machine_type=self.instance_type(test_duration),
+                                   machine_type=instance_type,
                                    instance_name=instance_name,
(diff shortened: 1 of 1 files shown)


fruch:
Can you share how a failure looks ?

The image creation fails cause of a wrong type ? or duration isn't set on image creation ?

enaydanov:
You put `self.instance_type(duration)` to a wrong place, actually. This breaks not only image creation but selection of custom instance type too.

enaydanov:
The call of `create_instance()` for image creation don't pass `test_duration` and pass exact `instance_type`.

fruch:
> The call of `create_instance()` for image creation don't pass `test_duration` and pass exact `instance_type`.

O.K. this is the information I couldn't see here.

fruch (approved):
LGTM

PR #6378, opened 2023-07-18
fix(remote): allow use non-RSA private keys

enaydanov:
Change parameter passed to paramiko Connection from `pkey` to `key_filename`. This allows paramiko to guess the type of a private key.

### DIFF
--- a/sdcm/remote/remote_cmd_runner.py
+++ b/sdcm/remote/remote_cmd_runner.py
@@ -16,7 +16,7 @@ import os
 import threading
 
 from fabric import Connection, Config
-from paramiko import SSHException, RSAKey
+from paramiko import SSHException
 from paramiko.ssh_exception import NoValidConnectionsError, AuthenticationException
 from invoke.watchers import StreamWatcher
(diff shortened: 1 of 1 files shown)


fruch:
Some context would be nice.

Is there any job we still use fabric/paramiko and not python-ssh ?

enaydanov:
We use paramiko remote runner in siren-tests.

fruch:
> We use paramiko remote runner in siren-tests.

Any specific reason we do so ? or just by mistake it's not configured with libssh2 ?

enaydanov:
It not used for communications with DB nodes, just for low-traffic connection to a maintenance host.
It's not a big deal to switch to libssh2 too, but this fix is nice to have anyway.

fruch:
> It not used for communications with DB nodes, just for low-traffic connection to a maintenance host. It's not a big deal to switch to libssh2 too, but this fix is nice to have anyway.

let make sure we do switch, so SCT would be able to ditch the fabric support, if we'll decide so.

soyacz (approved):

PR #6381, opened 2023-07-18
Rework cloud manager instance discovery

enaydanov:
- Use correct IP address for SSH connection
- Don't use support user credentials (the SSH key will be provided by siren-tests)
- Small refactoring of `find_and_append_cloud_manager_instance_to_collecting_nodes()`

Need to backport https://github.com/scylladb/scylla-cluster-tests/pull/6381/commits/e1e85b4508b227a6076920152b8d066ce382d4cb to SCT release branches to stop sync support key.

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -54,7 +54,7 @@ post_behavior_loader_nodes: "destroy"
 post_behavior_monitor_nodes: "keep-on-failure"
 post_behavior_k8s_cluster: "keep-on-failure"
 
-cloud_credentials_path: '~/.ssh/support'
+cloud_credentials_path: ''
 use_cloud_manager: false
 cloud_prom_bearer_token: ''
 
--- a/get-qa-ssh-keys.sh
+++ b/get-qa-ssh-keys.sh
@@ -9,7 +9,6 @@ umask 077
 aws s3 sync s3://${SSH_KEYS_BUCKET} ${SSH_KEYS_DIR} --exact-timestamps --exclude '*' \
   --include scylla-qa-ec2 \
   --include scylla-test \
-  --include support \
   --include scylla_test_id_ed25519 \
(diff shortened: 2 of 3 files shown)


defaults/test_default.yaml
```diff
@@ -54,7 +54,6 @@ post_behavior_loader_nodes: "destroy"
 post_behavior_monitor_nodes: "keep-on-failure"
 post_behavior_k8s_cluster: "keep-on-failure"
 
-cloud_credentials_path: '~/.ssh/support'
```
fruch, line 57: If this is removed, let's remove it completely from sct_config

get-qa-ssh-keys.sh
```diff
@@ -9,7 +9,6 @@ umask 077
 aws s3 sync s3://${SSH_KEYS_BUCKET} ${SSH_KEYS_DIR} --exact-timestamps --exclude '*' \
   --include scylla-qa-ec2 \
   --include scylla-test \
-  --include support \
```
fruch, line 12: This support key isn't gonna be used at all ?

Let's check if we can remove it completely (and stop downloading it also on SCT release branches)

defaults/test_default.yaml
fruch: If not, let's leave an empty default.

enaydanov: OK, I'll set it to empty string.

get-qa-ssh-keys.sh
enaydanov: I think we need to keep it in keystore, but stop to download in release branches.

fruch: So let's put this in a separate commit so we can backport

enaydanov: It's already done, btw :)

fruch (approved):
LGTM

Issue #6353, opened 2023-07-12
Collect more data on test failure


xemul:
When scylla times-out and gets aborted it's tempting to know more info about the system  other than logs  _before_ it gets killed (i.e. -- _before_ the coredump is produced), in particular

- the output of the `lsof` command to get the file descriptors scylla is holding
- the output of the `netstat -tanp` command to get the list of sockets and their states

refs: scylladb/scylladb#13100

fruch:
it would be easy to collect those, and spit them in sct own log.

as separate files would also not going to be that hard.

@soyacz can you look at it ?


@xemul is it reproducible in one of those cases ? so we'll be able to test that it would work ?

soyacz:
> 
> @soyacz can you look at it ?

@fruch I looked and I'm not sure if this is the right approach, but I was thinking about adding some command to `ExecStop` (e.g. if PID exists after some time, run above commands (and redirect output to some files). Possibly this could be done as part of `sdcm.cluster.BaseNode.fix_scylla_server_systemd_config`. WDYT?

fruch:
> I was trying to add some code to/after 'stop_scylla_server' but looks that our timeout functionality does not work for that case - it should timeout after 300 seconds, while it returns right only after service stops (after SIGABRT, ~20 minutes).
> Need to investigate why this timeout does not work (I have some ideas), but I struggle to use `RemoteLibSSH2CmdRunner` alone for reproduction...
> I can take a look on Monday, unless @fruch you select someone else for this.

Don't worry, it will be waiting for you...

PR #6384, opened 2023-07-18
feature(node): collect lsof and netstat on scylla stop freeze

soyacz:
When Scylla freezes on stop, SIGABRT is sent by systemd automatically
creating coredump. It would be informative if `lsof` and `netstat`
output was attached to issue along with core dumps.

In case timeout occures during stopping scylla, collect `lsof` and
`netstat` output. After collection send SIGABRT to Scylla process.

To make this possible, needed to fix our sshlib2 based remoter:

    fix(ssh): Fix close session on timeout
    
    When command times out and does not print anything to stdout/stderr,
    then timeout error does not appear and command hangs until it quits.
    This causes error e.g. when we run `systemctl stop scylla` and there's a
    freeze in stopping Scylla's process.
    
    Fix is about ignoring `channel.close()` return code, which in that case
    returns `LIBSSH2_ERROR_EAGAIN`. This code in case of closing
    channel means that it would block (it is not an error per se:
    https://www.libssh2.org/libssh2_channel_close.html).
    Later we anyway wait for channel to be closed with
    `channel.wait_closed()`.

Fixes https://github.com/scylladb/scylla-cluster-tests/issues/6353

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -43,7 +43,7 @@ import yaml
 import requests
 from paramiko import SSHException
 from tenacity import RetryError
-from invoke.exceptions import UnexpectedExit, Failure, CommandTimedOut
+from invoke.exceptions import UnexpectedExit, Failure
 from cassandra import ConsistencyLevel
 from cassandra.auth import PlainTextAuthProvider
 from cassandra.cluster import Cluster as ClusterDriver  # pylint: disable=no-name-in-module
@@ -116,7 +116,7 @@ from sdcm.sct_events.filters import EventsSeverityChangerFilter
 from sdcm.utils.auto_ssh import AutoSshContainerMixin
 from sdcm.monitorstack.ui import AlternatorDashboard
 from sdcm.logcollector import GrafanaSnapshot, GrafanaScreenShot, PrometheusSnapshots, upload_archive_to_s3, \
-    save_kallsyms_map
+    save_kallsyms_map, collect_diagnostic_data
 from sdcm.utils.ldap import LDAP_SSH_TUNNEL_LOCAL_PORT, LDAP_BASE_OBJECT, LDAP_PASSWORD, LDAP_USERS, \
     LDAP_PORT, DEFAULT_PWD_SUFFIX
 from sdcm.utils.remote_logger import get_system_logging_thread
@@ -2411,12 +2411,27 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         if verify_up:
             self.wait_jmx_up(timeout=timeout)
 
-    @retrying(n=3, sleep_time=5, allowed_exceptions=NETWORK_EXCEPTIONS + (CommandTimedOut, ),
+    @retrying(n=3, sleep_time=5, allowed_exceptions=NETWORK_EXCEPTIONS,
               message="Failed to stop scylla.server, retrying...")
     def stop_scylla_server(self, verify_up=False, verify_down=True, timeout=300, ignore_status=False):
         if verify_up:
             self.wait_db_up(timeout=timeout)
-        self.stop_service(service_name='scylla-server', timeout=timeout, ignore_status=ignore_status)
+        try:
+            self.stop_service(service_name='scylla-server', timeout=timeout, ignore_status=ignore_status)
+        except Exception as details:  # pylint: disable=broad-except
+            if isinstance(details, RetryableNetworkException):
+                details = details.original
+            if details.__class__.__name__.endswith("CommandTimedOut"):
+                LOGGER.error(
+                    "timeout during stopping scylla-server, collect diagnostic data and wait for systemd to kill it")
+                collect_diagnostic_data(self)
+                wait.wait_for(func=lambda:
+                              self._service_cmd(service_name="scylla-server", cmd='is-active',
+                                                timeout=timeout, ignore_status=True).stdout.strip() == "inactive",
+                              step=60, text="still waiting for scylla-server to stop", timeout=900, throw_exc=True)
+
+            else:
+                raise
         if verify_down:
             self.wait_db_down(timeout=timeout)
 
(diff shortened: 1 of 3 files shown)


soyacz:
I just need to think how to test it...

sdcm/cluster.py
```diff
@@ -2411,12 +2411,20 @@ def start_scylla(self, verify_up=True, verify_down=False, timeout=500):
...
         if verify_up:
             self.wait_db_up(timeout=timeout)
-        self.stop_service(service_name='scylla-server', timeout=timeout, ignore_status=ignore_status)
+        try:
+            self.stop_service(service_name='scylla-server', timeout=1, ignore_status=ignore_status)
+        except CommandTimedOut:
+            LOGGER.error(
+                "timeout during stopping scylla-server, collect diagnostic data and wait for SIGABRT triggered by systemd")
+            collect_diagnostic_data(self)
+            wait.wait_for(func=lambda: "Active: inactive" in
```
yarongilor, line 2425: Isn't this line doing the same as the below `verify_down` ?

sdcm/cluster.py
```diff
@@ -4370,7 +4381,7 @@ def node_setup(self, node: BaseNode, verbose: bool = False, timeout: int = 3600)
             node.install_package('python3')
 
         node.update_repo_cache()
-
+        node.install_package('lsof')
```
fruch, line 4384: we'll need to test it across all distros (if the package is named the same)

fruch:
@soyacz can you reference the issue, where this was asked for ?

soyacz:
> @soyacz can you reference the issue, where this was asked for ?

https://github.com/scylladb/scylla-cluster-tests/issues/6353

sdcm/cluster.py
fruch: we install it, if it's not found

fruch: and yes, I think it was added to the image at some point

fruch:
> > @soyacz can you reference the issue, where this was asked for ?
> 
> #6353

add "Fixes #6353" to the PR description (or even to the commit)

also we  should call this collection when timing out of nodetool calls as well.
i.e. other place we kill scylla on timeout

sdcm/logcollector.py
```diff
@@ -844,6 +844,33 @@ def save_kallsyms_map(node):
...
+    Log Entities should have unique names, otherwise won't be collected."""
+    LOGGER.info('Collecting diagnostics data from: %s', node.name)
+    if remote_node_dir := create_remote_storage_dir(node):
+        for log_entity in log_entities:
+            if os.path.exists(os.path.join(node.logdir, log_entity.name)):
+                LOGGER.debug("Diagnostic file '%s' already exists and not changed. Skipping collection.",
+                             log_entity.name)
+                continue
+            try:
+                log_entity.collect(node, node.logdir, remote_node_dir)
```
fruch, line 868: the collect only bring it back to the test runner ? or upload to s3 ?

sdcm/remote/libssh2_client/__init__.py
```diff
@@ -635,7 +635,8 @@ def _complete_run(self, channel: Channel, exception: Exception,  # pylint: disab
         result.stderr = stderr.getvalue()
         if channel is not None:
             try:
-                self.session.eagain(channel.close, timeout=self.timings.channel_close_timeout)
+                with self.session.lock:
```
fruch, line 638: @dkropachev what do you think about this one ?

soyacz: also take a look at the commit message for explanation

sdcm/logcollector.py
soyacz: to test runner, similar like kallsyms

sdcm/cluster.py
soyacz: it checks `is_port_used` - and in case of freeze after closing port, we don't know - I wanted to simulate same behavior as systemctl - waiting for status inactive. I'm not sure about this solution thou.

sdcm/cluster.py
soyacz: from all sources I found over internet, lsof and netstat are coming from the same packages: `lsof` and `net-tools`

soyacz:
finally I made it working: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-mini-test-1h-test/19/ - see db log.

The hard part was that CommandTimedOut was wrapped around other exception and I couldn't make collection triggered properly. I needed to search specific message in exception. Another thing is that after fixing timeout functionality, I'm not sure if scylla get's stopped properly (if stop is not aborted) - that's why I added manual SIGABRT send to scylla process. So I wonder if it is ok and if we shouldn't prolong timeout currently set to 5 minutes (previously SIGABRT was sent after 20 minutes by systemd).

The run doesn't show netstat log because was not installed. It's fixed and I'll retry with it later.

xemul (dismissed):

sdcm/cluster.py
```diff
@@ -2411,12 +2411,27 @@ def start_scylla(self, verify_up=True, verify_down=False, timeout=500):
...
 
-    @retrying(n=3, sleep_time=5, allowed_exceptions=NETWORK_EXCEPTIONS + (CommandTimedOut, ),
+    @retrying(n=3, sleep_time=5, allowed_exceptions=NETWORK_EXCEPTIONS,
               message="Failed to stop scylla.server, retrying...")
     def stop_scylla_server(self, verify_up=False, verify_down=True, timeout=300, ignore_status=False):
         if verify_up:
             self.wait_db_up(timeout=timeout)
-        self.stop_service(service_name='scylla-server', timeout=timeout, ignore_status=ignore_status)
+        try:
+            self.stop_service(service_name='scylla-server', timeout=1, ignore_status=ignore_status)
```
fruch, line 2420: The one here is for testing I guess ?

Anyhow I'm not sure about the kill, it's o.k. that we'll have this info regardless if we'll have a coredump.

If something get stuck it will coredump, if just slow, we might not.

soyacz:
centos8 test: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifact-centos8-sheduler-error/6/

soyacz:
do we want to backport it?

fruch:
> do we want to backport it?

Let's let's run a bit on the master, before thinking where to backport

If the a reproducer for issue, we can run it as reproducer with the master version. or with a specific branch for it.

xemul (approved):

PR #6390, opened 2023-07-19
fix(upgrade_test): handle uuid sstable identifier on upgradesstable

fruch:
since scylladb/scylladb/pull/13932 sstable identifier are not numbers, and the regex we were using in upgradesstable procudure assumed they are.

failing like the following:
```
AssertionError: Failed to upgrade the sstable format [False, False, False, False]
```

Ref: https://github.com/scylladb/scylladb/pull/13932

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -404,7 +404,7 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
                     r"find /var/lib/scylla/data/system -type f ! -path '*snapshots*' -printf %f\\n")
                 all_sstable_files = result.stdout.splitlines()
 
-                sstable_version_regex = re.compile(r'(\w+)-\d+-(.+)\.(db|txt|sha1|crc32)')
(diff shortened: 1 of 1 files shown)


fgelcer (approved):
LGTM

upgrade_test.py
```diff
@@ -404,7 +404,7 @@ def wait_for_node_to_finish():
                     r"find /var/lib/scylla/data/system -type f ! -path '*snapshots*' -printf %f\\n")
                 all_sstable_files = result.stdout.splitlines()
 
-                sstable_version_regex = re.compile(r'(\w+)-\d+-(.+)\.(db|txt|sha1|crc32)')
+                sstable_version_regex = re.compile(r'(\w+)-.+-(.+)\.(db|txt|sha1|crc32)')
```
soyacz, line 407: isn't . too broad? [\d\w]+ wouldn't be better?

fruch: it's a uuid, an in general we don't care what's there, we care only on the sstable format version, which is the two letter at the beginning, I'm o.k. with it to be broad.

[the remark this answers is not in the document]
fruch: almost identical from my POV, but I'll change

PR #6394, opened 2023-07-19
improvement(test-cases): Add bare metal performance test

k0machi:
This commit fixes some issues regarding bare metal instances -
specifically it fixes setup for loader and monitor nodes, adds a new
keystore method to retrive bare metal cluster information instead of
storing it in the test case and fixes syslong-ng setup script to make
sure it reinstalls if it's present - to make repeated runs possible as
previously it would just do nothing and syslog-ng would transfer logs to
an old sct runner instead. Some similar fixes for prometheus and
scylla_setup scripts to make sure we can run them multiple times on the
same instance.

Task: scylladb/qa-tasks#1233

### DIFF
--- a/sdcm/cluster_baremetal.py
+++ b/sdcm/cluster_baremetal.py
@@ -1,5 +1,7 @@
 import logging
-from typing import Optional
+from typing import Optional, TypedDict
+
+from invoke import UnexpectedExit
 
 from sdcm import cluster
 
@@ -10,6 +12,22 @@ LOADER_NAME = 'loader-node'
 MONITOR_NAME = 'monitor-node'
 
 
+class NodeInfo(TypedDict):
+    public_ip: str
+    private_ip: str
+
+
+class NodeCredentialInformation(TypedDict):
+    username: str
+    node_list: list[NodeInfo]
+
+
+class BareMetalCredentials(TypedDict):
+    db_nodes: NodeCredentialInformation
+    loader_nodes: NodeCredentialInformation
+    monitor_nodes: NodeCredentialInformation
+
+
 class NodeIpsNotConfiguredError(Exception):
     pass
 
@@ -17,9 +35,11 @@ class NodeIpsNotConfiguredError(Exception):
 class PhysicalMachineNode(cluster.BaseNode):
     log = LOGGER
 
-    def __init__(self, name, parent_cluster, public_ip, private_ip, credentials, base_logdir=None, node_prefix=None):  # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-arguments
+    def __init__(self, name, parent_cluster: 'PhysicalMachineCluster',
+                 public_ip, private_ip, credentials, base_logdir=None, node_prefix=None):
         ssh_login_info = {'hostname': None,
-                          'user': credentials.name,
+                          'user': getattr(parent_cluster, "ssh_username", credentials.name),
                           'key_file': credentials.key_file}
         self._public_ip = public_ip
         self._private_ip = private_ip
@@ -37,6 +57,21 @@ class PhysicalMachineNode(cluster.BaseNode):
     def _get_public_ip_address(self) -> Optional[str]:
         return self._public_ip
 
+    @property
+    def region(self):
+        return "baremetal"
+
+    def scylla_setup(self, disks, devname: str):
+        try:
+            super().scylla_setup(disks, devname)
+        except UnexpectedExit as exc:
+            # Covering for case when scylla-setup script has already been run. If the command is scylla_setup
+            # and there's "already" in the stdout of that command, we can skip this method safely as nics and disks
+            # were already configured on this node.
+            if "scylla_setup" in exc.result.command and "already" in exc.streams_for_display()[0].lower():
+                return
+            raise exc
+
     def _get_private_ip_address(self) -> Optional[str]:
         return self._private_ip
 
@@ -45,17 +80,6 @@ class PhysicalMachineNode(cluster.BaseNode):
         # since not using hostname anywhere
         # self.remoter.run('sudo hostnamectl set-hostname {}'.format(self.name))
         pass
-
-    def detect_disks(self, nvme=True):  # pylint: disable=unused-argument
-        """
-        Detect local disks
-        """
-        min_size = 50 * 1024 * 1024 * 1024  # 50gb
-        result = self.remoter.run('lsblk -nbo KNAME,SIZE,MOUNTPOINT -s -d')
-        lines = [line.split() for line in result.stdout.splitlines()]
-        disks = ['/dev/{}'.format(l[0]) for l in lines if l and int(l[1]) > min_size and len(l) == 2]
-        assert disks, 'Failed to find disks!'
-        return disks
 
     def reboot(self, hard=True, verify_ssh=True):
         raise NotImplementedError("reboot not implemented")
@@ -74,12 +98,17 @@ class PhysicalMachineCluster(cluster.BaseCluster):  # pylint: disable=abstract-m
         self.nodes = []
         self.credentials = kwargs.pop('credentials')
         n_nodes = kwargs.get('n_nodes')
-        self._node_public_ips = kwargs.get('public_ips', None) or []
-        self._node_private_ips = kwargs.get('private_ips', None) or []
+        self._node_public_ips = kwargs.pop('public_ips', None) or []
+        self._node_private_ips = kwargs.pop('private_ips', None) or []
         node_cnt = n_nodes[0] if isinstance(n_nodes, list) else n_nodes
         if len(self._node_public_ips) < node_cnt or len(self._node_private_ips) < node_cnt:
             raise NodeIpsNotConfiguredError('Physical hosts IPs are not configured!')
         super().__init__(**kwargs)
+
+    @property
+    def ssh_username(self) -> str:
+        # pylint: disable=no-member
+        return self._ssh_username
 
     def _create_node(self, name, public_ip, private_ip):
         node = PhysicalMachineNode(name,
@@ -105,6 +134,8 @@ class ScyllaPhysicalCluster(cluster.BaseScyllaCluster, PhysicalMachineCluster):
 
     def __init__(self, **kwargs):
         user_prefix = kwargs.pop('user_prefix')
+        if username := kwargs.pop('ssh_username', None):
+            self._ssh_username = username
         kwargs.update(dict(
             cluster_prefix=cluster.prepend_user_prefix(user_prefix, 'db-cluster'),
             node_prefix=cluster.prepend_user_prefix(user_prefix, 'db-node'),
@@ -112,27 +143,22 @@ class ScyllaPhysicalCluster(cluster.BaseScyllaCluster, PhysicalMachineCluster):
         ))
         super().__init__(**kwargs)
 
-    def node_setup(self, node, verbose=False, timeout=3600):
-        """
-        Configure scylla.yaml on cluster nodes.
-        We have to modify scylla.yaml on our own because we are not on AWS,
-        where there are auto config scripts in place.
-        """
-        # self._node_setup(node, verbose)
 
-
-class LoaderSetPhysical(PhysicalMachineCluster, cluster.BaseLoaderSet):  # pylint: disable=abstract-method
+class LoaderSetPhysical(cluster.BaseLoaderSet, PhysicalMachineCluster):  # pylint: disable=abstract-method
 
     def __init__(self, **kwargs):
         user_prefix = kwargs.pop('user_prefix')
+        if username := kwargs.pop('ssh_username', None):
+            self._ssh_username = username
         kwargs.update(dict(
             cluster_prefix=cluster.prepend_user_prefix(user_prefix, 'loader-set'),
             node_prefix=cluster.prepend_user_prefix(user_prefix, 'loader-node'),
             node_type='loader'
         ))
-        super().__init__(**kwargs)
+        cluster.BaseLoaderSet.__init__(self, kwargs["params"])
+        PhysicalMachineCluster.__init__(self, **kwargs)
 
-    @classmethod
+    @ classmethod
     def _get_node_ips_param(cls, ip_type='public'):
         return cluster.BaseLoaderSet.get_node_ips_param(ip_type)
 
@@ -141,6 +167,8 @@ class MonitorSetPhysical(cluster.BaseMonitorSet, PhysicalMachineCluster):
 
     def __init__(self, **kwargs):
         user_prefix = kwargs.pop('user_prefix')
+        if username := kwargs.pop('ssh_username', None):
+            self._ssh_username = username
         kwargs.update(dict(
             cluster_prefix=cluster.prepend_user_prefix(user_prefix, 'monitor-set'),
             node_prefix=cluster.prepend_user_prefix(user_prefix, 'monitor-node'),
@@ -149,4 +177,5 @@ class MonitorSetPhysical(cluster.BaseMonitorSet, PhysicalMachineCluster):
         cluster.BaseMonitorSet.__init__(self,
                                         targets=kwargs["targets"],
                                         params=kwargs["params"])
+        kwargs.pop("targets")
         PhysicalMachineCluster.__init__(self, **kwargs)
--- a/sdcm/keystore.py
+++ b/sdcm/keystore.py
@@ -101,6 +101,9 @@ class KeyStore:  # pylint: disable=too-many-public-methods
     def get_argus_rest_credentials(self):
         return self.get_json("argus_rest_credentials.json")
 
+    def get_baremetal_config(self, config_name: str):
+        return self.get_json(f"{config_name}.json")
+
 
 def pub_key_from_private_key_file(key_file):
     try:
--- a/sdcm/provision/common/utils.py
+++ b/sdcm/provision/common/utils.py
@@ -152,6 +152,8 @@ def install_syslogng_service():
         SYSLOG_NG_INSTALLED=""
         if yum --help 2>/dev/null 1>&2 ; then
             if rpm -q syslog-ng ; then
+                rm /etc/syslog-ng/syslog-ng.conf  # Make sure we have default syslog-ng.conf
+                yum reinstall -y syslog-ng
                 SYSLOG_NG_INSTALLED=1
             else
                 for n in 1 2 3 4 5 6 7 8 9; do # cloud-init is running it with set +o braceexpand
@@ -170,6 +172,9 @@ def install_syslogng_service():
             fi
         elif apt-get --help 2>/dev/null 1>&2 ; then
             if dpkg-query --show syslog-ng ; then
+                rm /etc/syslog-ng/syslog-ng.conf  # Make sure we have default syslog-ng.conf
+                apt-get purge --autoremove -y syslog-ng
+                DPKG_FORCE=confmiss apt-get --reinstall -y install syslog-ng
                 SYSLOG_NG_INSTALLED=1
             else
                 cat /etc/apt/sources.list
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -911,7 +911,7 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
             db_cluster = self.db_cluster
         # No need to change system tables when running via scylla-cloud
         # Also, when running a Alternator via scylla-cloud, we don't have CQL access to the cluster
-        if self.params.get('db_type') == 'cloud_scylla':
+        if self.params.get('db_type') == 'cloud_scylla' or self.params.get("cluster_backend") == "baremetal":
             # TODO: move this skip to siren-tools when possible
             self.log.warning("Skipping this function due this job run from Siren cloud!")
             return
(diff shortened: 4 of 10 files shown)


k0machi:
There are currently two major problems right now:

1. SCT is unable to cope with a node that doesn't have an nvme drive in a performance test - removing RAID setup code leads to later errors when tests try to execute `fstrim` on an arbitrary block device.
2. The runs are not repeatable - SCT tries to reinstall (update) scylla and fails when scylla_setup and other setup scripts return non-zero exit codes due to actions of that script being already done, e.g. "RAID already configured"

sdcm/tester.py
```diff
@@ -1364,20 +1364,23 @@ def get_cluster_baremetal(self):
             private_ips=self.params.get('db_nodes_private_ip'),
             user_prefix=self.params.get('user_prefix'),
             credentials=self.credentials,
+            ssh_username='ubuntu',
```
aleksbykov, line 1367: I think better to store ssh_username in config files. May add new parameters `baremetal_db_ssh_username`, `baremetal_loader_ssh_username` or similar. Because it will be also need in future the ssh key file, which should be stored in aws key store

aleksbykov:
> There are currently two major problems right now:
> 
> 1. SCT is unable to cope with a node that doesn't have an nvme drive in a performance test - removing RAID setup code leads to later errors when tests try to execute `fstrim` on an arbitrary block device.

I think it should not failed without nvme drives. during setup if 1 disk or several but they are not nvme detected or passed scylla should be configure correctly.

> 2. The runs are not repeatable - SCT tries to reinstall (update) scylla and fails when scylla_setup and other setup scripts return non-zero exit codes due to actions of that script being already done, e.g. "RAID already configured"

i think we need to refactor setup for baremetal, and add checks that if hardware is already configured, don't run that scripts and only remove scylla configs and data dirs and reinstall it.

roydahan:
> 



> > There are currently two major problems right now:
> > 
> > 1. SCT is unable to cope with a node that doesn't have an nvme drive in a performance test - removing RAID setup code leads to later errors when tests try to execute `fstrim` on an arbitrary block device.
> 
> I think it should not failed without nvme drives. during setup if 1 disk or several but they are not nvme detected or passed scylla should be configure correctly.

It's ok, we don't need to handle this if this is not handled at the moment.
We can assume that Scylla must have NVMe.
> 
> > 2. The runs are not repeatable - SCT tries to reinstall (update) scylla and fails when scylla_setup and other setup scripts return non-zero exit codes due to actions of that script being already done, e.g. "RAID already configured"
> 
> i think we need to refactor setup for baremetal, and add checks that if hardware is already configured, don't run that scripts and only remove scylla configs and data dirs and reinstall it.

Another option is to use "reuse_cluster" with "update_db_packages" in case we want to install new scylla.
> 
> do you have jobs or logs?

sdcm/cluster_baremetal.py
```diff
@@ -17,9 +20,9 @@ class NodeIpsNotConfiguredError(Exception):
 class PhysicalMachineNode(cluster.BaseNode):
     log = LOGGER
 
-    def __init__(self, name, parent_cluster, public_ip, private_ip, credentials, base_logdir=None, node_prefix=None):  # pylint: disable=too-many-arguments
+    def __init__(self, name, parent_cluster: 'PhysicalMachineCluster', public_ip, private_ip, credentials, base_logdir=None, node_prefix=None):  # pylint: disable=too-many-arguments
         ssh_login_info = {'hostname': None,
-                          'user': credentials.name,
```
fruch, line 22: We need to be extra careful in changing baremetal implementation, since the siren tests are using it, so test it with siren at some point before merging this

test-cases/performance/perf-regression-throughput-baremetal-5gb.yaml
```diff
@@ -0,0 +1,63 @@
...
+run_db_node_benchmarks: false
+ssh_transport: fabric
+
+use_mgmt: false
+
+n_db_nodes: 3
+n_loaders: 1
+n_monitor_nodes: 1
+
+db_nodes_public_ip:
```
fruch, line 27: We need to see how this part of the configuration isn't coming from public source code

sdcm/tester.py
fruch: I don't understand what's wrong with the configuration option we have today for the db user, why can't we use that ?

Why have other defaults in the code, or why introduce new configuration options ?

fruch:
> > 
> 
> 
> 
> > > There are currently two major problems right now:
> > > 
> > > 1. SCT is unable to cope with a node that doesn't have an nvme drive in a performance test - removing RAID setup code leads to later errors when tests try to execute `fstrim` on an arbitrary block device.
> > 
> > I think it should not failed without nvme drives. during setup if 1 disk or several but they are not nvme detected or passed scylla should be configure correctly.
> 
> It's ok, we don't need to handle this if this is not handled at the moment.
> We can assume that Scylla must have NVMe.
> > 
> > > 2. The runs are not repeatable - SCT tries to reinstall (update) scylla and fails when scylla_setup and other setup scripts return non-zero exit codes due to actions of that script being already done, e.g. "RAID already configured"
> > 
> > i think we need to refactor setup for baremetal, and add checks that if hardware is already configured, don't run that scripts and only remove scylla configs and data dirs and reinstall it.
> 
> Another option is to use "reuse_cluster" with "update_db_packages" in case we want to install new scylla.

update_db_packages might work 

We can also implement cleanup for bare metal, that would clear scylla installation and unmount disks, and what's needed to make install work again after that.

> > 
> > do you have jobs or logs?
> 
>

k0machi:
> do you have jobs or logs?

I am currently starting the job manually via hydra, so no logs unfortunately, once I stabilize it enough to at least pass once I'll run it in jenkins

> I don't understand what's wrong with the configuration option we have today for the db user, why can't we use that ?

This is just a temporary plug to satisfy the class, as for `db_user`, is there also `loader_user` param? I haven't touched that aside that because I wasn't sure if I should be adding new params or not

> update_db_packages might work

Will try

> We need to see how this part of the configuration isn't coming from public source code

Can't we pass it as env as with other SCT Config values?

> We need to be extra careful in changing baremetal implementation, since the siren tests are using it, so test it with siren at some point before merging this

Hm, I was under impression this wasn't used much, can @enaydanov elaborate?

> I think it should not failed without nvme drives. during setup if 1 disk or several but they are not nvme detected or passed scylla should be configure correctly.

It's not a configuration issue, there are places later during stress that just run `fstrim` and if the disk doesn't support it, it errors out and causes a shutdown. It seems that EBS Virtual Block devices do not support that operation.

```
xvda    202:0    0    30G  0 disk
└─xvda1 202:1    0    30G  0 part /
nvme0n1 259:0    0 442.4G  0 disk /var/lib/scylla
scyllaadm@argusdb-node-1:~$ sudo fstrim /
fstrim: /: the discard operation is not supported
```

k0machi:
Ready for review, more or less, will add a jenkins pipeline in a moment, need to shorten the normal parallel one to only contain one test that (maybe) can be specified in the params (so you could have one job that could do read, write and mixed). Or maybe just split it into 3 jobs.

sdcm/cluster_baremetal.py
k0machi: credentials.name here is ssh keyfile comment - it doesn't seem it would work with pretty much any key we have.

sdcm/cluster_baremetal.py
```diff
@@ -37,6 +58,18 @@ def init(self):
...
 
+    @property
+    def region(self):
+        return "baremetal"
+
+    def scylla_setup(self, disks, devname: str):
+        try:
+            super().scylla_setup(disks, devname)
+        except UnexpectedExit as exc:
+            if "scylla_setup" in exc.result.command and "already" in exc.streams_for_display()[0].lower():
```
temichus, line 68: please leave a comment here that describe this if statement logic

sdcm/tester.py
```diff
@@ -1355,29 +1355,32 @@ def get_cluster_docker(self):
 
     def get_cluster_baremetal(self):
         # pylint: disable=too-many-locals,too-many-statements,too-many-branches
-
+        baremetal_info: cluster_baremetal.BareMetalCredentials = KeyStore().get_baremetal_credentials()
```
fruch, line 1358: you gonna have multiple of those, and not just one set of configuration... 
probably need to name them and have a configuration option to point to them

similar to how we do it fro the internal-qa repo, as in the 5000 table case:

```yaml
cs_user_profiles:
    - scylla-qa-internal/cust_d/templated_tables_mv.yaml
```

sdcm/provision/common/utils.py
```diff
@@ -152,6 +152,8 @@ def install_syslogng_service():
         SYSLOG_NG_INSTALLED=""
         if yum --help 2>/dev/null 1>&2 ; then
             if rpm -q syslog-ng ; then
+                rm /etc/syslog-ng/syslog-ng.conf  # Make sure we have default syslog-ng.conf
```
fruch, line 155: why do we need this reinstall ? why isn't it enough we have syslog-ng installed ?

sdcm/cluster_baremetal.py
```diff
@@ -149,4 +176,5 @@ def __init__(self, **kwargs):
         cluster.BaseMonitorSet.__init__(self,
                                         targets=kwargs["targets"],
                                         params=kwargs["params"])
+        kwargs.pop("targets")
```
fruch, line 179: what's special about targets ? and why do we need to pop it out ?

test-cases/performance/perf-regression-throughput-baremetal-5gb.yaml
```diff
@@ -0,0 +1,54 @@
...
+                    "cassandra-stress write no-warmup cl=ALL n=312500 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=10 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=937501..1250000"]
+
+
+stress_cmd_w: "cassandra-stress write no-warmup cl=ALL n=1250000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=4 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=1..1250000"
+stress_cmd_r: "cassandra-stress read no-warmup  cl=QUORUM duration=30m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=4 -col 'size=FIXED(128) n=FIXED(8)' -pop 'dist=gauss(1..1250000,625001,1250000)' "
+stress_cmd_m: "cassandra-stress mixed no-warmup cl=QUORUM duration=30m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=4 -col 'size=FIXED(128) n=FIXED(8)' -pop 'dist=gauss(1..1250000,625001,1250000)' "
+stress_multiplier: 1
+instance_type_db: 'i3.2xlarge'
```
fruch, line 13: I think this should be removed, we should have code using it beside provisioning code

sdcm/cluster_baremetal.py
```diff
@@ -74,13 +98,20 @@ def __init__(self, **kwargs):
...
-        self._node_private_ips = kwargs.get('private_ips', None) or []
+        self._node_public_ips = kwargs.pop('public_ips', None) or []
+        self._node_private_ips = kwargs.pop('private_ips', None) or []
         node_cnt = n_nodes[0] if isinstance(n_nodes, list) else n_nodes
         if len(self._node_public_ips) < node_cnt or len(self._node_private_ips) < node_cnt:
             raise NodeIpsNotConfiguredError('Physical hosts IPs are not configured!')
         super().__init__(**kwargs)
 
+    @property
+    def ssh_username(self) -> str:
```
fruch, line 109: I don't understand the need of this property, also we don't want default like `ubuntu` to be in the code.
this should have it's own new sct configuration option

sdcm/keystore.py
```diff
@@ -101,6 +101,9 @@ def get_argusdb_credentials(self):
     def get_argus_rest_credentials(self):
         return self.get_json("argus_rest_credentials.json")
 
+    def get_baremetal_credentials(self):
```
fruch, line 104: this name is misleading, since it's more then just credentials.
it's also the information of a specific set of baremetal setup

sdcm/tester.py
k0machi: Yes, I thought of that. I can implement this in this PR if needed.

test-cases/performance/perf-regression-throughput-baremetal-5gb.yaml
k0machi: Will do

sdcm/cluster_baremetal.py
k0machi: I've moved this into bare metal config already, maybe just raise an exception there if it's unset?

sdcm/keystore.py
k0machi: Will change

sdcm/cluster_baremetal.py
k0machi: PhysicalMachineCluster -> BaseCluster has no parameter called `targets` and so the constructor raises an exception.

sdcm/keystore.py
k0machi: Done

sdcm/cluster_baremetal.py
k0machi: Changed to re-raise AttributeError

sdcm/tester.py
k0machi: Done, implemented as `s3_baremetal_config`

k0machi:
Here's a [link](https://argus.scylladb.com/test/4224a5ba-77ce-4d0a-a4c3-c424a657db17/runs?additionalRuns[]=3e8eb05d-0612-4973-bedd-701a3f7a1043) to the test job in Argus.

roydahan (commented):
Overall looks good to me.
As @fruch mentioned we need to make sure @enaydanov review it and possibly test siren longevity making sure it doesn't break anything.

Config and example looks convenient to me.

test-cases/performance/perf-regression-throughput-baremetal-5gb.yaml
k0machi: There's a slight problem with changing this - in the pipeline the SCT Runner step actually runs with `AWS` configuration, to create a runner inside AWS, so I'm not sure if I can remove it (talking about `instance_type_db`)

roydahan:
@k0machi is it waiting to you or reviewers?

k0machi:
> @k0machi is it waiting to you or reviewers? 

Waiting for Evgeniy to review, he should do it tomorrow.

enaydanov:
@k0machi can you rebase this PR to the latest master?

k0machi:
> @k0machi can you rebase this PR to the latest master?

Done @enaydanov

roydahan:
@enaydanov did you review this?

sdcm/cluster_baremetal.py
enaydanov: Can you change it to something like `getattr(self.parent_cluster, "ssh_username", credentials.name)` ?

sdcm/cluster_baremetal.py
enaydanov: Why do you need to catch exception at all in this case?

enaydanov (changes requested):

k0machi:
> Why do you need to catch exception at all in this case?
It's to shut up the pylint about `no-member`, just to be explicit that we may be accessing non-existent attribute there

> Can you change it to something like getattr(self.parent_cluster, "ssh_username", credentials.name) ?
Done

enaydanov:
> > Why do you need to catch exception at all in this case?
> 
> It's to shut up the pylint about `no-member`, just to be explicit that we may be accessing non-existent attribute there
> 
In that case it's better to use explicit `# pylint: disable=...` or add a comment. It's not obvious.

Also, if you'll prefer to stay with try/except, it's better to use just `raise` to not add a duplicate to a traceback.

sdcm/cluster_baremetal.py
```diff
@@ -105,32 +134,29 @@ class ScyllaPhysicalCluster(cluster.BaseScyllaCluster, PhysicalMachineCluster):
 
     def __init__(self, **kwargs):
         user_prefix = kwargs.pop('user_prefix')
+        if username := kwargs.pop('ssh_username'):
```
enaydanov, line 137: This is also a problem. Need to change it to `kwargs.pop('ssh_username', None)`

enaydanov (changes requested):

sdcm/cluster_baremetal.py
```diff
@@ -10,16 +12,34 @@
...
 class PhysicalMachineNode(cluster.BaseNode):
     log = LOGGER
 
-    def __init__(self, name, parent_cluster, public_ip, private_ip, credentials, base_logdir=None, node_prefix=None):  # pylint: disable=too-many-arguments
+    # pylint: disable=too-many-arguments
+    def __init__(self, name, parent_cluster: 'PhysicalMachineCluster',
+                 public_ip, private_ip, credentials, base_logdir=None, node_prefix=None):
         ssh_login_info = {'hostname': None,
-                          'user': credentials.name,
+                          'user': getattr(self.parent_cluster, "ssh_username", credentials.name),
```
enaydanov, line 42: We have a problem here: `self.parent_cluster` will be assigned later in `super().__init__()`, need to change it to just `parent_cluster`. I think it's came from my example, sorry about that :(

enaydanov (changes requested):

enaydanov (dismissed):

k0machi:
@fruch @soyacz Looks like this PR is ready to be merged.

test-cases/performance/perf-regression-throughput-baremetal-5gb.yaml
fruch: So check, having unrelated configuration would cause confusion

vars/perfRegressionBareMetalSinglePipeline.groovy
```diff
@@ -0,0 +1,329 @@
+#!groovy
```
fruch, line 1: What does the Single mean here ?

vars/perfRegressionBareMetalSinglePipeline.groovy
```diff
@@ -0,0 +1,329 @@
...
+            }
+            stage("Create SCT Runner") {
+                steps {
+                    wrap([$class: 'BuildUser']) {
+                        dir('scylla-cluster-tests') {
+                            timeout(time: 10, unit: 'MINUTES') {
+                                script {
+                                    // Always use aws for sct runner
+                                    def awsParams = [:]
+                                    awsParams["backend"] = "aws"
```
fruch, line 158: All the pipeline duplicated just for this configuration of the runner ?

I would suggest introducing an optional runner_backend option to SCT for that

fruch (commented):
Mostly comment about wording and configuration 

And the copy of the pipeline, that we should try to minimize.

vars/perfRegressionBareMetalSinglePipeline.groovy
k0machi: The previous pipeline was called Matrix, so this is the opposite of that.

vars/perfRegressionBareMetalSinglePipeline.groovy
k0machi: Yes, that would be a good thing to implement, should I do it with this PR or move it to a separate task?

vars/perfRegressionBareMetalSinglePipeline.groovy
fruch: You can just remove that word

vars/perfRegressionBareMetalSinglePipeline.groovy
fruch: It depends if it's a 20-30min work, let's do it now, if you estimate it to be more then that, let's write a separate task for it

vars/perfRegressionBareMetalSinglePipeline.groovy
k0machi: Done

vars/perfRegressionBareMetalSinglePipeline.groovy
k0machi: Reverted that, changed to adjusting `getCloudProviderFromBackend` to map `baremetal` to `aws`

k0machi: Seems to be working fine now, agent is properly selected and sct runner is created using old params

vars/perfRegressionBareMetalPipeline.groovy
```diff
@@ -0,0 +1,323 @@
+#!groovy
+import groovy.json.JsonSlurper
```
fruch, line 2: @k0machi 

one last thing, can you change code to run with `perfRegressionParallelPipeline.groovy`, it's now should have everything you were missing, i.e. creating sct runner

fruch (changes requested):
I think now perfRegressionBareMetalPipeline.groovy is redundant, it doesn't have anything different then all other pipelines

vars/perfRegressionBareMetalPipeline.groovy
k0machi: The parallel pipeline will try to deploy extra instances for read, write and mixed tests, where as this pipeline only does specified subtest in parameters. Since we only have one set of baremetal instances I think that's not going to work.

fruch: it's depend on the list of subtest you supply to it, some tests have only one, and should work in you case.

I rather not duplicate pipelines, unless there's a real reason, so far I'm no convinced there is

k0machi:
> I think now perfRegressionBareMetalPipeline.groovy is redundant, it doesn't have anything different then all other pipelines

That is indeed the case, I've removed the file and reapplied the original pipeline which seems to be [working](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/alexey/job/perf-regression-throughput-baremetal-example-test/31/)

fruch (approved):
LGTM

Issue #6395, opened 2023-07-20
Use -pr for repair system_auth and make it parallel


soyacz:
There was discussion that we do repair of system_auth wrong: a very slow way. This causes problems in multi-dc scenarios with many nodes, causing test to timeout.
https://github.com/scylladb/scylla-enterprise/issues/3103#issuecomment-1635172393

Let's add -pr to repair system_auth and run it in parallel.

PR #6396, opened 2023-07-20
fix(system_auth): speed up repair system_auth during node setup

soyacz:
Repairing `system_auth` keyspace after altering replication strategy during node setup takes very long, making multi-dc tests to fail due timeout.

Lets add `-pr` parameter to repair and do it in parallel as advised by developers (and similar to Scylla Manager does).

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/6395

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -46,7 +46,7 @@ from argus.client.sct.types import Package, EventsInfo, LogLink
 from argus.backend.util.enums import TestStatus
 from sdcm import nemesis, cluster_docker, cluster_k8s, cluster_baremetal, db_stats, wait
 from sdcm.cluster import BaseCluster, NoMonitorSet, SCYLLA_DIR, TestConfig, UserRemoteCredentials, BaseLoaderSet, BaseMonitorSet, \
-    BaseScyllaCluster, BaseNode
+    BaseScyllaCluster, BaseNode, MINUTE_IN_SEC
 from sdcm.cluster_azure import ScyllaAzureCluster, LoaderSetAzure, MonitorSetAzure
 from sdcm.cluster_gce import ScyllaGCECluster
 from sdcm.cluster_gce import LoaderSetGCE
@@ -888,6 +888,10 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
                 validate_raft_on_nodes(nodes=db_cluster.nodes)
 
     def set_system_auth_rf(self, db_cluster=None):
+
+        def _nodetool_repair(node):
+            node.run_nodetool(sub_cmd="repair -pr system_auth", timeout=MINUTE_IN_SEC * 20)
+
         if not db_cluster:
             db_cluster = self.db_cluster
         # No need to change system tables when running via scylla-cloud
@@ -915,8 +919,9 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
                              num_retry_on_failure=3)
         self.log.debug("system_auth description: %s", res.stdout)
         self.log.info('repair system_auth keyspace ...')
-        for node in self.db_cluster.nodes:
-            node.run_nodetool(sub_cmd="repair system_auth")
+        parallel_objects = ParallelObject(self.db_cluster.nodes, num_workers=min(
+            32, len(self.db_cluster.nodes)), timeout=MINUTE_IN_SEC * 25)
+        parallel_objects.run(_nodetool_repair)
         self.log.info('repair system_auth keyspace done')
 
     @cache


sdcm/tester.py
```diff
@@ -915,8 +921,9 @@ def set_system_auth_rf(self, db_cluster=None):
                              num_retry_on_failure=3)
         self.log.debug("system_auth description: %s", res.stdout)
         self.log.info('repair system_auth keyspace ...')
-        for node in self.db_cluster.nodes:
-            node.run_nodetool(sub_cmd="repair system_auth")
+        parallel_objects = ParallelObject(self.db_cluster.nodes, num_workers=min(
+            32, len(self.db_cluster.nodes)), timeout=HOUR_IN_SEC * 2)
```
fgelcer, line 925: 2 hours for this repair? isn't it too much?

fruch: without the parallelism it can take more the 2h
regardless it's depends on the number of nodes you setup, at the beginning

so multi-dc usually starts from 9 nodes, scale-tests also can start from a quite high number.

so we should set it high enough

@soyacz maybe we should use the adaptive_timeout ? so we could measure it ? (and make the timeout on the number of nodes ?)

sdcm/tester.py
```diff
@@ -888,6 +888,12 @@ def setUp(self):  # pylint: disable=too-many-branches,too-many-statements
                 validate_raft_on_nodes(nodes=db_cluster.nodes)
 
     def set_system_auth_rf(self, db_cluster=None):
+
+        def _nodetool_repair(node):
+            self.log.info('starting repair system_auth keyspace on node %s ...', node.name)
+            node.run_nodetool(sub_cmd="repair -pr system_auth")
```
fruch, line 894: we have the nodetool events, not sure if we need those info print per node (maybe debug would be enough)

sdcm/tester.py
fgelcer: 1st, they are now running in parallel in all nodes... so, this timeout should be technically for each repair on `system_auth` table... how is it possible it takes hours?? if that is the case, we have a huge issue, since this table should be almost empty

soyacz: I was rather thinking to measure how long it takes and put hard timeout on it (or soft one, but without adaptive timeout as currently all the factors we measure don't impact this repair time. Worth to consider some additional metrics put to ES). I think so because this is one of the first things we do in cluster, no nemesis, no load, no data - should always be fast and deterministic. This hard timeout should take into considerations number of nodes (or anything that affects repair speed).

sdcm/tester.py
soyacz: possibly not, I'll remove it

sdcm/tester.py
soyacz: > 1st, they are now running in parallel in all nodes... so, this timeout should be technically for each repair on `system_auth` table... how is it possible it takes hours?? if that is the case, we have a huge issue, since this table should be almost empty

Actually we have issue with that in multi-dc - that's why this PR was created

fgelcer: > Actually we have issue with that in multi-dc - that's why this PR was created

before we were running repair in a serial manner, but now they are in parallel... for a tiny table... 2 hours to have all of them, regardless the amount of nodes, sounds too much... but a run or two will prove it

soyacz: Yes, I'll be running tests with this to see actual numbers. Without `-pr` for multi-dc scenario with 9 nodes (together) repair of one node took almost 1h. Currently it takes long because there's plenty of requests per range (not table rows) and in multi-dc network delay greatly impacts it.

fruch: > > Actually we have issue with that in multi-dc - that's why this PR was created
> 
> before we were running repair in a serial manner, but now they are in parallel... for a tiny table... 2 hours to have all of them, regardless the amount of nodes, sounds too much... but a run or two will prove it

Sounds weird, according to:
https://github.com/scylladb/scylla-enterprise/issues/3103

The numbers of ranges that need to be checked across all the nodes in the RF, has nothing to do with the amount of data.

soyacz:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/reproducer-counters-multidc-raft/21/
whole cluster system_auth repair process takes 9 minutes (multi-dc scenario)

soyacz:
I set timeout 20 minutes per repair (25 minutes for general parallel object timeout) - it's hard timeout and should fail the test in case of error. 
WDYT?

fruch:
> I set timeout 20 minutes per repair (25 minutes for general parallel object timeout) - it's hard timeout and should fail the test in case of error. 
> WDYT? 

Sounds o.k. but still I think it would be nice to collect that information across the different setups and tests (since we did have an issue it can take 2 hours, yes wrongly done, but still)

fruch (approved):
LGTM

roydahan (approved):

soyacz:
It's not backported yet, @fruch shall we (along with https://github.com/scylladb/scylla-cluster-tests/pull/6652)?

fruch:
> It's not backported yet, @fruch shall we (along with #6652)?

we could

PR #6434, opened 2023-07-27
feat(repair): enable single-host-parallelism for repairs

dkropachev:
https://github.com/scylladb/scylla-manager/pull/3496 brought new repair option --single-host-parallelism we need to test it

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -362,6 +362,9 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
             self.test_restore_backup_with_task()
         with self.subTest('Repair Multiple Keyspace Types'):
             self.test_repair_multiple_keyspace_types()
+        if self._repair_host_parallelism_feature_exists:
+            with self.subTest('Repair Multiple Keyspace Types With Single Host Parallelism'):
+                self.test_repair_multiple_keyspace_types(host_parallelism=1)
         with self.subTest('Mgmt Cluster CRUD'):
             self.test_mgmt_cluster_crud()
         with self.subTest('Mgmt cluster Health Check'):
@@ -864,15 +867,23 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         assert manager_from_version[0] != manager_tool.sctool.version[0], "Manager version not changed after rollback."
         self.log.info('finishing test_manager_rollback_upgrade')
 
-    def test_repair_multiple_keyspace_types(self):  # pylint: disable=invalid-name
-        self.log.info('starting test_repair_multiple_keyspace_types')
+    @property
+    def _repair_host_parallelism_feature_exists(self) -> bool:
         manager_tool = mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0])
+        return manager_tool.feature('repair', 'single-host-parallelism')
+
+    def test_repair_multiple_keyspace_types(self, host_parallelism: int = None):  # pylint: disable=invalid-name
+        manager_tool = mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0])
+        test_name = 'test_repair_multiple_keyspace_types with single-host-parallelism={}'.format(
+            host_parallelism) if host_parallelism is not None else 'test_repair_multiple_keyspace_types'
+
+        self.log.info('starting {}'.format(test_name))
         mgr_cluster = manager_tool.get_cluster(cluster_name=self.CLUSTER_NAME) \
             or manager_tool.add_cluster(name=self.CLUSTER_NAME, db_cluster=self.db_cluster,
                                         auth_token=self.monitors.mgmt_auth_token)
         self._create_keyspace_and_basic_table(self.SIMPLESTRATEGY_KEYSPACE_NAME, "SimpleStrategy", replication_factor=2)
         self._create_keyspace_and_basic_table(self.LOCALSTRATEGY_KEYSPACE_NAME, "LocalStrategy")
-        repair_task = mgr_cluster.create_repair_task()
+        repair_task = mgr_cluster.create_repair_task(host_parallelism=host_parallelism)
         task_final_status = repair_task.wait_and_get_final_status(timeout=7200)
         assert task_final_status == TaskStatus.DONE, 'Task: {} final status is: {}.'.format(repair_task.id,
                                                                                             str(repair_task.status))
@@ -897,7 +908,7 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         self.log.info("the sctool repair command was completed successfully")
 
         mgr_cluster.delete()  # remove cluster at the end of the test
-        self.log.info('finishing test_repair_multiple_keyspace_types')
+        self.log.info('finishing {}'.format(test_name))
 
     def test_enospc_during_backup(self):
         self.log.info('starting test_enospc_during_backup')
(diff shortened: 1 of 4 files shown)


sdcm/nemesis.py
```diff
@@ -2616,10 +2616,25 @@ def _mgmt_backup(self, backup_specific_tables):
...
+            raise UnsupportedNemesis('Scylla-manager configuration is not defined!')
+        if not self.cluster.scylla_manager_tool.feature("repair", "single-host-parallelism"):
+            self.log.info('single-host-parallelism is not supported running normal repair')
+            self.disrupt_mgmt_repair_cli_normal()
+            return
+
+        match random.choice(["no_no_parallelism", "normal"]):
+            case "no_no_parallelism":
+                self.log.info('single-host-parallelism is supported running repair with single-host-parallelism=1')
+                self._mgmt_repair_cli(host_parallelism=1)
```
fgelcer, line 2629: for longevity tests, 1 sounds too low... could we use something like `len(self.cluster.nodes)/2` or `len(self.cluster.nodes)/3`

sdcm/nemesis.py
```diff
@@ -2616,10 +2616,25 @@ def _mgmt_backup(self, backup_specific_tables):
...
     @latency_calculator_decorator(legend="Scylla-Manger repair")
     def disrupt_mgmt_repair_cli(self):
+        if not self.cluster.params.get('use_mgmt') and not self.cluster.params.get('use_cloud_manager'):
+            raise UnsupportedNemesis('Scylla-manager configuration is not defined!')
+        if not self.cluster.scylla_manager_tool.feature("repair", "single-host-parallelism"):
+            self.log.info('single-host-parallelism is not supported running normal repair')
+            self.disrupt_mgmt_repair_cli_normal()
+            return
+
+        match random.choice(["no_no_parallelism", "normal"]):
```
fgelcer, line 2626: instead of this return here, for consistency, i would add the `match` clause inside an `else` clause

sdcm/nemesis.py
```diff
@@ -2616,10 +2616,25 @@ def _mgmt_backup(self, backup_specific_tables):
 
     @latency_calculator_decorator(legend="Scylla-Manger repair")
     def disrupt_mgmt_repair_cli(self):
+        if not self.cluster.params.get('use_mgmt') and not self.cluster.params.get('use_cloud_manager'):
+            raise UnsupportedNemesis('Scylla-manager configuration is not defined!')
+        if not self.cluster.scylla_manager_tool.feature("repair", "single-host-parallelism"):
```
fgelcer, line 2621: on an extra thought about that, it will affect the performance test that uses this function...
@temichus @aleksbykov @roydahan , WDYT?

sdcm/nemesis.py
```diff
@@ -2616,10 +2616,25 @@ def _mgmt_backup(self, backup_specific_tables):
...
+            self.log.info('single-host-parallelism is not supported running normal repair')
+            self.disrupt_mgmt_repair_cli_normal()
```
fgelcer, line 2623: where is this function's definition?

fgelcer: shouldn't it be just the same as the below `normal` mgmt repair? AFAIU it needs still all the definition of asserting if the task is not done.... that now is missing from this call

fgelcer (changes requested):

sdcm/nemesis.py
temichus: @fgelcer 
Yes, in any case, need to implement an option to explicitly set  "normal repair" for perf tests
But also we can run perf tests with single-host-parallelism and check how it affects perf tests

fgelcer: yes, for perf tests we must have the ability to control it from the call, as right now we don't want to change it..
about running with this new feature, we can, and should... but it will lower the amount of work the repair is doing, so obviously it will improve the latency...

fgelcer:
@dkropachev , another 2 things:
1. pre-checks failed.. please fix
2. we are right now testing parallelism of 1, or the default... we should add here some randomness of the amount of parallelism, if not set by the caller

sdcm/nemesis.py
roydahan: I don't think we want this in our regular nemesis.
It will cause repairs to take very long and we don't have the issue of parallelism in our test.

sdcm/nemesis.py
dkropachev: @fgelcer , from what I understand about this flag is that major use-case for it is to be set to 1, in order to reduce stale repair probability. I guess we could do random from 1 to something. Also take into account that it is not number of nodes participating in the repair, it is maximum number of repairs to run on single node at any time.

sdcm/nemesis.py
dkropachev: removed this part based on @roydahan comment

sdcm/nemesis.py
dkropachev: removed it from the nemesis

sdcm/nemesis.py
dkropachev: This part of code is removed

dkropachev:
> @dkropachev , another 2 things:
> 
> 1. pre-checks failed.. please fix
> 2. we are right now testing parallelism of 1, or the default... we should add here some randomness of the amount of parallelism, if not set by the caller

Nemesis part is removed as requested by @roydahan

mgmt_cli_test.py
```diff
@@ -362,6 +362,8 @@ def test_manager_sanity(self):
             self.test_restore_backup_with_task()
         with self.subTest('Repair Multiple Keyspace Types'):
             self.test_repair_multiple_keyspace_types()
+        with self.subTest('Repair Multiple Keyspace Types With Single Host Parallelism'):
```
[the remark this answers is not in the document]
dkropachev: Well, I inside of the function I had a step to check if it is available, when you pointed it out and realized that even better not to run it at all and move this logic one level up.
Please take a look

fgelcer (approved):
LGTM

sdcm/mgmt/cli.py
```diff
@@ -1275,6 +1285,12 @@ def version(self):
         cmd = "version"
         return self.run(cmd=cmd, is_verify_errorless_result=True)
 
+    def feature(self, command: str, feature: str) -> bool:
+        return "--{}".format(feature) in self.run(
```
fruch, line 1289: can you supply an example of this command output

sdcm/mgmt/operator.py
```diff
@@ -420,3 +428,8 @@ def add_cluster(self, name, host=None, db_cluster=None, client_encrypt=None, dis
 
     def upgrade(self, scylla_mgmt_upgrade_to_repo):
         raise NotImplementedError()
+
+    def feature(self, command: str, feature: str) -> bool:
+        if command == "repair" and feature == "single-host-parallelism":
```
fruch, line 433: can you explain this logic please, why would operator not support it ?

fruch: I rather have here `raise NotImplementedError()` until someone actually test it, and verify this works or not on operator,

after it's already established we aren't gonna use this on nemesis for now.

sdcm/nemesis.py
dkropachev: done

sdcm/mgmt/cli.py
dkropachev: Added

fruch (approved):
LGTM

PR #6435, opened 2023-07-27
fix(mv_si_4d): improve prepare stress

fgelcer:
since we introduced custom time test, this
test is having too long prepare, hence it
times out... for this improvement, we are
trying to reduce its duration by removing
all reads from prepare, increasing parallelism
and removing nemesis during prepare.

### DIFF
--- a/test-cases/longevity/longevity-mv-si-4days.yaml
+++ b/test-cases/longevity/longevity-mv-si-4days.yaml
@@ -1,6 +1,6 @@
 test_duration: 6550
-prepare_write_cmd: ["cassandra-stress user profile=/tmp/c-s_profile_4mv_5queries.yaml ops'(insert=15,read1=1,read2=1,read3=1,read4=1,read5=1)' cl=QUORUM n=100000000 -mode cql3 native -rate threads=10",
-                    "cassandra-stress user profile=/tmp/c-s_profile_3si_5queries.yaml ops'(insert=25,si_read1=1,si_read2=1,si_read3=1,si_read4=1,si_read5=1)' cl=QUORUM n=100000000 -mode cql3 native -rate threads=10"
+prepare_write_cmd: ["cassandra-stress user profile=/tmp/c-s_profile_4mv_5queries.yaml ops'(insert=15,read1=1,read2=1,read3=1,read4=1,read5=1)' cl=QUORUM n=100000000 -mode cql3 native -rate threads=200",
+                    "cassandra-stress user profile=/tmp/c-s_profile_3si_5queries.yaml ops'(insert=25,si_read1=1,si_read2=1,si_read3=1,si_read4=1,si_read5=1)' cl=QUORUM n=100000000 -mode cql3 native -rate threads=200"
                    ]
 stress_read_cmd: ["cassandra-stress user profile=/tmp/c-s_profile_4mv_5queries.yaml ops'(insert=15,read1=1,read2=1,read3=1,read4=1,read5=1)' cl=QUORUM duration=5760m -mode cql3 native -rate threads=10",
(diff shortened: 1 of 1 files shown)


fgelcer:
giving it a try [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/custom-time/job/mv-si-4d-reduced-preparella-/2/) to confirm the improvement on the prepare part (using as custom duration 60 min)

fgelcer:
[here is a snapshot](https://snapshots.raintank.io/dashboard/snapshot/giXQct8SwsaCin1wIbGVjepPQt66oP0j) of a master run (as it was before this PR's changes) - `test_id: 9eaec735-1bef-48a3-a6c2-cea6c753c816`

[here is a snapshot](https://snapshots.raintank.io/dashboard/snapshot/C7J14qfLaFMrvPhsqfsh0Js2gOpPzprY) using this PR (the duration itself is much shorter, so here comparing only the prepare) - `test_id: f091570b-0c39-4d7a-b131-22a9feebe2b4`

roydahan (changes requested):
Unneeded change.

fruch:
> Since we have nemesis_during_prepare here and we have a lot of milage with this longevity, I think that we don't need to change it after all.
> It's better to have 12 hours run when we have nemesis during prepare than having 12h while prepare is faster but don't have nemesis for 5h.
> 
> So, if needed we can just extend the prepare timeout (if exist) or completely remove it and discard this PR.
> 

Nemesis during preparation I can see why we'll want.

But all those extra reads during preparation, they seem a bit pointless, and look a bit like a mistake no one noticed so far.

fgelcer:
> > Since we have nemesis_during_prepare here and we have a lot of milage with this longevity, I think that we don't need to change it after all.
> > It's better to have 12 hours run when we have nemesis during prepare than having 12h while prepare is faster but don't have nemesis for 5h.
> > So, if needed we can just extend the prepare timeout (if exist) or completely remove it and discard this PR.
> 
> Nemesis during preparation I can see why we'll want.
> 
> But all those extra reads during preparation, they seem a bit pointless, and look a bit like a mistake no one noticed so far.

so, @roydahan , should i keep the nemesis during prepare, and drop the reads after all? or leave as it is today (that is not great, because it takes ~10h to prepare... what is a bit too much for the custom time test... it means, we will need to increase the prepare custom time for this job as well

fgelcer:
> Another possibility is to use large instance type, leave everything as today but increase the threads number to make sure it finishes quicker.

i will do drop the reads during prepare (they are not really needed), increase instance type, and increase even more the prepare threads, but will keep nemesis during prepare...

made the changes, giving it a run [in here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/custom-time/job/mv-si-4d-reduced-preparella-/4/) to see if it will be enough

fgelcer:
> > Another possibility is to use large instance type, leave everything as today but increase the threads number to make sure it finishes quicker.
> 
> i will do drop the reads during prepare (they are not really needed), increase instance type, and increase even more the prepare threads, but will keep nemesis during prepare...
> 
> made the changes, giving it a run [in here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/custom-time/job/mv-si-4d-reduced-preparella-/4/) to see if it will be enough

[here is a monitor](https://snapshots.raintank.io/dashboard/snapshot/bdR1Tngpj7Ni5GeXTxiBaSj3JIj4BlK3) for that
the prepare took here 2 hours 20 minutes

yarongilor (commented):
Not sure i follow - is a single write command duplicated on 2 loaders? why not splitting the content?

roydahan (approved):

PR #6440, opened 2023-07-30
fix(1tb longevity): change nemesis seed

fgelcer:
seeing in latest master effort (tier1 job)
that is using the new custom time feature,
so we can stabilize tests, that the 1TB job
is running repeatedly few not very important
nemeses, hence the idea to update it.

### DIFF
--- a/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
+++ b/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
@@ -30,7 +30,7 @@ gce_instance_type_loader: 'c2-standard-16'
 
 
 nemesis_class_name: 'SisyphusMonkey'
-nemesis_seed: '031'
+nemesis_seed: '042'
 nemesis_interval: 30
 nemesis_filter_seeds: false
 


fgelcer:
@soyacz , based on this current `nemesis_seed`, can i have the list of nemesis that will be selected for this job? so we can decide if this is good enough, or if i need to pick another value?
i see that we have a test that "tests" nemesis, but i couldn't find one that gives me reliable result (most of them got stuck locally)

fruch:
> @soyacz , based on this current `nemesis_seed`, can i have the list of nemesis that will be selected for this job? so we can decide if this is good enough, or if i need to pick another value? i see that we have a test that "tests" nemesis, but i couldn't find one that gives me reliable result (most of them got stuck locally)

I think this can do the trick:

```diff
diff --git a/unit_tests/test_nemesis.py b/unit_tests/test_nemesis.py
index d4e775c74..d19063e81 100644
--- a/unit_tests/test_nemesis.py
+++ b/unit_tests/test_nemesis.py
@@ -204,6 +204,16 @@ def test_use_disabled_monkey():

     assert 'disrupt_toggle_table_gc_mode' in collected_disrupt_methods_names

+def test_check_seed():
+    tester = FakeTester()
+
+    tester.params["nemesis_exclude_disabled"] = True
+    tester.params["nemesis_selector"] = []
+    tester.params["nemesis_seed"] = "ABC"
+    sisyphus = SisyphusMonkey(tester, None)
+
+    print("\n".join(disrupt.__name__ for disrupt in sisyphus.disruptions_list))
+

 class TestSisyphusMonkeyNemesisFilter:
     # pylint: disable=no-self-use
```

fgelcer:
> This is going to help us exactly for few days till we merge the next new nemesis.
> 
> We need to think maybe we want to have a better filter for this test that will limit the nemesis to only 20-30 types of nemesis.

so, maybe reducing the `nemesis_multiply_factor` to 2 would make it work... or either, only for the custom time trigger to set the `nemesis_multiply_factor` to 1 or 2

fgelcer:
> This is going to help us exactly for few days till we merge the next new nemesis.
> 
> We need to think maybe we want to have a better filter for this test that will limit the nemesis to only 20-30 types of nemesis.

based on this comment, moved the PR to ready for review

roydahan:
Not sure why

fruch:
> This is going to help us exactly for few days till we merge the next new nemesis.
> 
> We need to think maybe we want to have a better filter for this test that will limit the nemesis to only 20-30 types of nemesis.
> 

We can label the ones you think are more  important and limit the case to those.

Not that I'm sure how we will "score" the nemesis

fruch:
> > This is going to help us exactly for few days till we merge the next new nemesis.
> > 
> > We need to think maybe we want to have a better filter for this test that will limit the nemesis to only 20-30 types of nemesis.
> 
> so, maybe reducing the `nemesis_multiply_factor` to 2 would make it work... or either, only for the custom time trigger to set the `nemesis_multiply_factor` to 1 or 2

`nemesis_multiply_factor`is multipling each test, and then randomizing, we could do the other way around, randomize and then duplicate the list 6 times.

We'll have less chances of duplicates on shorter runs. If that was the main issue.

fruch:
Would need to implement different way of selecting nemesis set for a specific job

This change wouldn't be permanent fix for such an issue

Issue #6439, opened 2023-07-30
Load was stopped before nemesis finish and as result nemesis was marked as failed (scheduler runtime is 0)


juliayakovlev:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Last nemesis fails after `TEST`END` event.
This nemesis `disrupt_sla_increase_shares_during_load` wait for load running and then measure scheduler runtime during load.
The problem here that in`sdcm.tester.ClusterTester.stop_resources` we first stop load and then waiting for nemesis finish.
```
    def stop_resources(self):  # pylint: disable=no-self-use
        self.log.debug('Stopping all resources')
        with silence(parent=self, name="Kill Stress Threads"):
            self.kill_stress_thread()

        # Stopping nemesis, using timeout of 30 minutes, since replace/decommission node can take time
        if self.db_cluster:
            self.get_nemesis_report(self.db_cluster)
            self.stop_nemesis(self.db_cluster)
            self.stop_resources_stop_tasks_threads(self.db_cluster)
            self.get_backtraces(self.db_cluster)
```

Due to load was stopped before nemesis finish, scheduler runtime is 0. And as result nemesis failed

## Impact

*Describe the impact this issue causes to the user.*

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Kernel Version: 5.15.0-1038-aws
Scylla version (or git commit hash): `2023.1.0~rc7-20230621.b8e0eed13d19` with build-id `854ffecaf8918f0fe0cde4c200486b7b84fd7cb0`

Cluster size: 3 nodes (i4i.large)

Scylla Nodes used in this run:
- longevity-5gb-1h-SlaIncreaseSharesD-db-node-b347237a-3 (16.16.63.4 | 10.0.3.253) (shards: 2)
- longevity-5gb-1h-SlaIncreaseSharesD-db-node-b347237a-2 (13.53.114.94 | 10.0.3.146) (shards: 2)
- longevity-5gb-1h-SlaIncreaseSharesD-db-node-b347237a-1 (16.170.238.113 | 10.0.1.176) (shards: 2)

OS / Image: `` (aws: undefined_region)

Test: `longevity-5gb-1h-SlaIncreaseSharesDuringLoad-aws`
Test id: `b347237a-2880-452d-9671-d76288737065`
[...]

PR #6442, opened 2023-07-30
feature(nemesis): introduce `raise_exception_in_thread`

fruch:
`raise_exception_in_thread` is introduced as a way to force kill a nemesis thread for the following reasons:

a) so we don't wait for 30min at the end of every test (can be
   more depend on the nemsis/testcase))
b) so nemesis are depended on the load (ex. SLA), and since we
   kill the load before stopping the nemesis, if we are currently
   during such a nemesis, it would spit false posisitve events
   since it won't find enough load on the cluster as it expects.

Fixes: #6439

### Testing
- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/25/
- [x] - https://argus.scylladb.com/test/4f234961-de7a-4d4a-a0bb-0795c66a7212/runs?additionalRuns[]=20d3d721-2bb5-4150-a423-6839d9603c44

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -86,6 +86,7 @@ from sdcm.utils.common import (
     download_dir_from_cloud,
     generate_random_string,
     prepare_and_start_saslauthd_service,
+    raise_exception_in_thread,
 )
 from sdcm.utils.ci_tools import get_test_name
 from sdcm.utils.distro import Distro
@@ -135,6 +136,7 @@ from sdcm.paths import (
     SCYLLA_MANAGER_TLS_KEY_FILE,
 )
 from sdcm.sct_provision.aws.user_data import ScyllaUserDataBuilder
+from sdcm.exceptions import KillNemesis
 
 
 # Test duration (min). Parameter used to keep instances produced by tests that
@@ -4387,6 +4389,7 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
         # pylint: disable=protected-access
         current_thread_frames = sys._current_frames()
         for nemesis_thread in self.nemesis_threads:
+            raise_exception_in_thread(nemesis_thread, KillNemesis)
             nemesis_thread.join(timeout)
             if nemesis_thread.is_alive():
                 stack_trace = traceback.format_stack(current_thread_frames[nemesis_thread.ident])
--- a/sdcm/exceptions.py
+++ b/sdcm/exceptions.py
@@ -55,3 +55,7 @@ class AuditLogTestFailure(Exception):
 class BootstrapStreamErrorFailure(Exception):  # pylint: disable=too-few-public-methods
     """ raised if node was not boostrapped after bootstrap
     streaming was aborted """
+
+
+class KillNemesis(BaseException):
+    """Exception that would be raised, when a nemesis thread is killed at teardown of the test"""
--- a/sdcm/exceptions.py
+++ b/sdcm/exceptions.py
@@ -1,0 +1,57 @@
+class NoFilesFoundToDestroy(Exception):
+    pass
+
+
+class NoKeyspaceFound(Exception):
+    pass
+
+
+class FilesNotCorrupted(Exception):
+    pass
+
+
+class LogContentNotFound(Exception):
+    pass
+
+
+class LdapNotRunning(Exception):
+    pass
+
+
+class TimestampNotFound(Exception):
+    pass
+
+
+class PartitionNotFound(Exception):
+    pass
+
+
+class WatcherCallableException(Exception):
+    """ raised when a watcher function in a trigger - watcher pair fails"""
+
+
+class UnsupportedNemesis(Exception):
+    """ raised from within a nemesis execution to skip this nemesis"""
+
+
+class CdcStreamsWasNotUpdated(Exception):
+    """ raised if messages:
+          - Generation {}: streams description table already updated
+          - CDC description table successfully updated with generation
+        were not found in logs
+    """
+
+
+class NemesisSubTestFailure(Exception):
+    """ raised if nemesis got error from sub test
+    """
+
+
+class AuditLogTestFailure(Exception):
+    """ raised if nemesis got error from audit log validation
+    """
+
+
+class BootstrapStreamErrorFailure(Exception):  # pylint: disable=too-few-public-methods
+    """ raised if node was not boostrapped after bootstrap
+    streaming was aborted """
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -118,6 +118,21 @@ from sdcm.utils.toppartition_util import NewApiTopPartitionCmd, OldApiTopPartiti
 from sdcm.utils.version_utils import MethodVersionNotFound, scylla_versions
 from sdcm.utils.raft import Group0MembersNotConsistentWithTokenRingMembersException
 from sdcm.wait import wait_for, wait_for_log_lines
+from sdcm.exceptions import (
+    NoFilesFoundToDestroy,
+    NoKeyspaceFound,
+    FilesNotCorrupted,
+    LogContentNotFound,
+    LdapNotRunning,
+    TimestampNotFound,
+    PartitionNotFound,
+    WatcherCallableException,
+    UnsupportedNemesis,
+    CdcStreamsWasNotUpdated,
+    NemesisSubTestFailure,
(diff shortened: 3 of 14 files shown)


sdcm/utils/common.py
```diff
@@ -3143,3 +3143,8 @@ def SoftTimeoutContext(timeout: int, operation: str):  # pylint: disable=invalid
     if duration > timeout:
         SoftTimeoutEvent(operation=operation, soft_timeout=timeout,
                          duration=duration).publish_or_dump()
+
+
+def raise_exception_in_thread(thread: threading.Thread, exception_type: Type[BaseException]):
+    res = ctypes.pythonapi.PyThreadState_SetAsyncExc(ctypes.c_ulong(thread.ident), ctypes.py_object(exception_type))
```
juliayakovlev, line 3149: Maybe we should raise if result of the api call is not as expected, like:
```
res = pythonapi.PyThreadState_SetAsyncExc(c_long(thread.ident),
                                                      py_object(exception_type))
        if res == 0:
            raise ValueError("invalid thread id")
        elif res != 1:
            # "if it returns a number greater than one, you're in trouble,
            # and you should call it again with exc=NULL to revert the effect"
            pythonapi.PyThreadState_SetAsyncExc(c_long(tid), None)
            raise SystemError("PyThreadState_SetAsyncExc failed")
```

juliayakovlev:
Why do we prefer to stop stress before nemesis?

fruch:
> Why do we prefer to stop stress before nemesis?

the assumption is that without stress, most nemesis would finish much more faster.
we could try that as well, but I would still like a way to stop nemesis with further questions.

sdcm/utils/common.py
fruch: this case might have, if calling this function multiple times without holding the GIL.
I don't think it could in our case.

But I still didn't see it happening as I would expect

soyacz:
This will not work sometimes (randomly) - because we often in our code use `except Exception` and we ignore/transform it.
So, we would need to retry on threads that didn't end several times or use `BaseException` as base class for `KillNemesis` - but I'm not sure if we should do it (it is not recommeded, but from the [other hand](https://docs.python.org/3/tutorial/errors.html#:~:text=BaseException%20is%20the%20common%20base,that%20the%20program%20should%20terminate.) it is used to mark fatal-error that usually should cause python exit)

fruch:
> This will not work sometimes (randomly) - because we often in our code use `except Exception` and we ignore/transform it. So, we would need to retry on threads that didn't end several times or use `BaseException` as base class for `KillNemesis` - but I'm not sure if we should do it (it is not recommeded, but from the [other hand](https://docs.python.org/3/tutorial/errors.html#:~:text=BaseException%20is%20the%20common%20base,that%20the%20program%20should%20terminate.) it is used to mark fatal-error that usually should cause python exit)

it would be a good exercise to flush out those silly places were we used too broad Exception catching.
we are using Base in `CriticalTestFailure` on the main stress thread, so maybe that would work o.k. as well

sdcm/cluster.py
```diff
@@ -4329,6 +4329,9 @@ def start_nemesis(self, interval=None, cycles_count: int = -1):
 
     @log_run_info("Stop nemesis threads on cluster")
     def stop_nemesis(self, timeout=10):
+        # avoid cyclic dependency between cluster.py and nemesis.py
+        from sdcm.nemesis import KillNemesis  # pylint: disable=import-outside-toplevel
```
vponomaryov, line 4333: Looks like it is time to create dedicated module for exceptions.

sdcm/nemesis.py
```diff
@@ -3970,8 +3977,8 @@ def disrupt_run_cdcstressor_tool(self):
         ks_tables_with_cdc = self.cluster.get_all_tables_with_cdc(self.target_node)
         if not ks_tables_with_cdc:
             self.log.warning("CDC is not enabled on any table. Skipping")
-            UnsupportedNemesis("CDC is not enabled on any table. Skipping")
-            return
+            raise UnsupportedNemesis("CDC is not enabled on any table. Skipping")
+
```
vponomaryov, line 3981: It looks like a totally unrelated to the PR idea.
This fix must be in a separate PR.

vponomaryov (changes requested):
If the only reason to implement this PR is fixing the https://github.com/scylladb/scylla-cluster-tests/issues/6439
Then, I think, would be better to enhance that particular nemesis to be smart enough and track the load presence itself.

For example, it can call prometheus API and check even some details of the load and not just making an assumption about it's presence as it does now.

fruch:
> If the only reason to implement this PR is fixing the https://github.com/scylladb/scylla-cluster-tests/issues/6439
> Then, I think, would be better to enhance that particular nemesis to be smart enough and track the load presence itself.
> 
> For example, it can call prometheus API and check even some details of the load and not just making an assumption about it's presence as it does now.

It's not one nemesis, it's set of them.
Also it's a long standing issue that we didn't implement a way to stop the nemesis immediately, but waiting for 30min to let it finish, which is not always the case, it's not the first time we had issues similar to that, or cases we get failure of the last nemesis being reported.

I don't want to make nemesis more complicated then necessary, and also be aware when the main load is stopped/ended.

soyacz:
> it would be a good exercise to flush out those silly places were we used too broad Exception catching.

image
Having 106+ exceptions handled like this it won't be an easy task...

> we are using Base in `CriticalTestFailure` on the main stress thread, so maybe that would work o.k. as well

If so, we could try that way.


> If the only reason to implement this PR is fixing the #6439
> Then, I think, would be better to enhance that particular nemesis to be smart enough and track the load presence itself.

Anyway, we lack a way of stopping nemesis in case we no longer care of its result (and we must finish in short time to not get timeout error). I couldn't figure out any other feasible way of doing it than proposed by @fruch.

vponomaryov (commented):
The new exceptions module could be in a separate commit.

fruch:
> The new exceptions module could be in a separate commit.

moved it to a separate commit

vponomaryov (approved):
LGTM

soyacz (dismissed):
LGTM
@fruch did you test if it works as expected? not killing too much?

fruch:
> LGTM
> @fruch did you test if it works as expected? not killing too much?

I tested it a few times, but none of them was during a working nemesis

I'm still waiting for @juliayakovlev to try it on the original job that run into that issue

fruch:
proven to be working better now, handling also stress failure that can happen during teardowns.

@juliayakovlev @soyacz @vponomaryov if you can give it a final review...

sdcm/exceptions.py
```diff
@@ -0,0 +1,60 @@
...
+
+class NemesisSubTestFailure(Exception):
+    """ raised if nemesis got error from sub test
+    """
+
+
+class AuditLogTestFailure(Exception):
+    """ raised if nemesis got error from audit log validation
+    """
+
```
vponomaryov, line 53: precommit complains that here must be 2 empty lines.

sdcm/nemesis.py
```diff
@@ -136,80 +152,12 @@
...
-
-class NoMandatoryParameter(Exception):
-    """ raised from within a nemesis execution to skip this nemesis"""
-
-
 class DefaultValue:  # pylint: disable=too-few-public-methods
     """
     This is class is intended to be used as default value for the cases when None is not applicable
     """
     ...
```
vponomaryov, line 160: Same here, one more empty line must be added.

vponomaryov (commented):
Precommit and unit tests failed. Need to fix.

sdcm/nemesis.py
fruch: fixed

sdcm/exceptions.py
fruch: fixed

fruch:
> Precommit and unit tests failed. Need to fix.

unit test was spot on, I was comparing the type of exception, and not the actual exception.

now it's works much better

fruch:
> > Precommit and unit tests failed. Need to fix.
> 
> unit test was spot on, I was comparing the type of exception, and not the actual exception.
> 
> now it's works much better

and now that's a unit test to cover the case of KillNemesis exception

vponomaryov (approved):
LGTM

Why backport is set only to `2023.1`?
What about others?

soyacz (approved):
LGTM

PR #6446, opened 2023-08-01
fix(kms_user): drop kms_hack.conf and place credential to ~scylla

syuu1228:
Modifying HOME environment breaks "dist: make p11-kit-trust.so able to work in relocatable package" PR on scylla-enterprise. It should use correct homedir of scylla user(/var/lib/scylla).

related scylladb/scylla-enterprise#3023


vponomaryov:
@syuu1228 
Thanks for the fix, but we should remove this creds workaround, not fix.
Here is the replacement PR with the removal: https://github.com/scylladb/scylla-cluster-tests/pull/6448

syuu1228:
@vponomaryov Okay, closing this PR.

Issue #6370, opened 2023-07-17
SCT does not update the argus report with updated configuration parameters from environment variables


ShlomiBalalis:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: 2022.2

### Logs
* test_id: a4dad539-7326-435a-8450-e24418143d0e
* job: https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-ami-test/1186/parameters/

### Description

Ever sinse https://github.com/scylladb/scylla-cluster-tests/pull/6158 was merged, SCT reports to argus only the base/default configuration from the yaml, and does not update the configuration based on the environment variables.

For example, the following [argus report](https://argus.scylladb.com/test/f5f76221-b454-4dd3-a91a-133678511161/runs?additionalRuns[]=a4dad539-7326-435a-8450-e24418143d0e) of an ami artifact test. The build was triggered with `i4i.xlarge` instance type, but argus shows that the instance type was `i3.large`, since it is the instance type stated in the scenario's yaml.

![Screenshot from 2023-07-17 15-07-31](https://github.com/scylladb/scylla-cluster-tests/assets/40186893/8314a419-042f-4d22-948d-d078fcf0d569)

![Screenshot from 2023-07-17 14-17-32](https://github.com/scylladb/scylla-cluster-tests/assets/40186893/c87e2566-b5ef-4f21-bd8e-94285b2ffd55)

### Steps to Reproduce

1. trigger a job with a param whose default parameter is overriden
2. check the argus page of said job

**Expected behavior:** The argus report should include the correct information on the specific run

**Actual behavior:** The argus report only includes the default config of the job

ShlomiBalalis:
@temichus It's this one

temichus:
PR: https://github.com/scylladb/scylla-cluster-tests/pull/6458

temichus:
@ShlomiBalalis please verify and close this issue.

PR #6451, opened 2023-08-02
fix(Argus reporting): fix ussue #6370

temichus:
add dynamic ENV exports at create-argus-test-run
for the following pipelines:
longevityPipeline
artifactsPipeline
jepsenPipeline
longevityPipeline
managerPipeline
rollingOperatorUpgradePipeline
rollingUpgradePipeline

https://github.com/scylladb/scylla-cluster-tests/issues/6370


temichus:
@vponomaryov , @soyacz , @fruch 

Please review. It is not a good solution but I did not find any other that will be capable to maintain those fi-fi statements for different jobs  without code duplication  in Groovy

vars/managerPipeline.groovy
```diff
@@ -178,7 +178,87 @@ def call(Map pipelineParams) {
                             wrap([$class: 'BuildUser']) {
                                 dir('scylla-cluster-tests') {
                                     timeout(time: 5, unit: 'MINUTES') {
-                                        createArgusTestRun(params)
```
fruch, line 181: Why can't we consolidate all those into one function ? And not create duplicates of the env handling logic it should be almost identical across the pipeline line.

especially those parts which are tested for existence first.

I think we should get unified, and not create more unique copies of that.

Introducing a new environment variable would be more painful then it's today.

temichus: the problem is that not identical.  it contains additional logic and different params
for example  [artifactsPipeline.groovy](https://github.com/scylladb/scylla-cluster-tests/pull/6451/files#diff-10408a52873447dcb93fa54ca89448ec42fd97d21136edea55eb798d0c4fa9a8)

dynamically cange instance type in for statement fro not param variable 
```

                                                    if [[ ! -z "${instance_type}" ]]; then
                                                        case "${params.backend}" in
                                                            "aws")
                                                                export SCT_INSTANCE_TYPE_DB="${instance_type}"
                                                                ;;
                                                            "gce")
                                                                export SCT_GCE_INSTANCE_TYPE_DB="${instance_type}"
                                                                ;;
                                                            "azure")
                                                                export SCT_AZURE_INSTANCE_TYPE_DB="${instance_type}"
                                                                ;;
                                                        esac
                                                    fi
```
Tomorrow will continue working on it a bit for now I see the solution  to create 1 function per pipeline 


```

```

temichus: @fruch, I tried to move it into function. Unsuccessfully

An export variable has gone since we exit SH block.
and moving it to a lower-level bash script will be very problematic since we are using  Groovy objects.

soyacz:
@temichus did you try to export all the params to yaml (or json) so later could be used in SCT like typical config yaml (append it to other config yamls). Create Argus run could take it as additional input then maybe?

temichus:
> @temichus did you try to export all the params to yaml (or json) so later could be used in SCT like typical config yaml (append it to other config yamls). Create Argus run could take it as additional input then maybe?

good idea, but it sounds like big code refactoring. I and @k0machi  have an idea of how to fix it from Argus's side. 

but I like moving exports to something readable and maintainable. We can create a separate task for this if  fix from Argus's side solve this issue

PR #6453, opened 2023-08-02
fix(mv si 4d): increase monitor instance type

fgelcer:
during master tier1 runs, we seen monitor having
some dificulties to show dashboards, for more
than one run, so we here increase the instance
type for the monitor, expecting it to work
better for future runs.

### DIFF
--- a/test-cases/longevity/longevity-mv-si-4days.yaml
+++ b/test-cases/longevity/longevity-mv-si-4days.yaml
@@ -18,6 +18,8 @@ instance_type_db: 'i4i.4xlarge'
 gce_instance_type_db: 'n2-highmem-32'
 gce_n_local_ssd_disk_db: 16
 
+instance_type_monitor: 't3.xlarge'
+
 nemesis_class_name: 'SisyphusMonkey'
 nemesis_seed: '023'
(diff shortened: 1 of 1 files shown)


fruch:
difficulties to show dashboards ? during the run ? when generating snapshot ? did it had any affect on the test ?

fgelcer:
> difficulties to show dashboards ? during the run ? when generating snapshot ? did it had any affect on the test ?

during the run, while trying to cut the time ranges for example... i wasn't unable to see any direct effect in the test though

yarongilor:
Change commit: `increase loader instance type` -> `increase monitor instance type`

fgelcer:
> Change commit: `increase loader instance type` -> `increase monitor instance type`

done

fruch:
> > difficulties to show dashboards ? during the run ? when generating snapshot ? did it had any affect on the test ?
> 
> during the run, while trying to cut the time ranges for example... i wasn't unable to see any direct effect in the test though

I'm guessing lots of cases might suffer from a similar issue, but as long as the data collection works well enough, I think we can live with it.

fruch (approved):

PR #6461, opened 2023-08-04
fix(test_scan_positive): fix unit tests

temichus:
fix test_scan_positive test according to new validation logic in validate_forward_service_requests_dispatched_to_other_nodes


unit_tests/test_scan_operation_thread.py
```diff
@@ -15,7 +15,17 @@
...
-from sdcm.scan_operation_thread import ScanOperationThread, ThreadParams, PrometheusDBStats
+from sdcm.utils.decorators import retrying, Retry
+
+
+def mock_retrying_decorator(*args, **kwargs):  # pylint: disable=unused-argument
+    """Decorate by doing nothing."""
+    return retrying(1, 1, allowed_exceptions=(Retry, ))
+
+
+patch('sdcm.utils.decorators.retrying', mock_retrying_decorator).start()
```
vponomaryov, line 26: Please, don't do it this way.
Patch stuff locally either by using context manager way (`with ...`) or patching test methods...

vponomaryov (changes requested):

unit_tests/test_scan_operation_thread.py
temichus: with patch "retrying" 
======================== 1 passed, 8 warnings in 11.10s ========================
without 

=================== 1 passed, 8 warnings in 71.16s (0:01:11) ===================

temichus: but I see what you mean - "it is not safe to do like this" .

temichus: @vponomaryov  Updated

it is still globally but now it should be safe and works only within one file

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

PR #6467, opened 2023-08-07
fix(nemesis.py): grow function presumes integer

fgelcer:
in master, there is a check if it is `None`, then
assign `0` to it, but in this branch, it doesn't
have such code.
It failed in this branch, as it was trying to
compare a `NoneType` with `> 0`.
This fix follows same behavior we have in master,
but the commit that introduced it in master
changed several unrelated things, making it
impossible to cherry-pick.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3485,20 +3485,17 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                           f'{str(exc)}').publish()
 
     def disrupt_grow_shrink_cluster(self):
-        self._grow_cluster(rack=0)
-        self._shrink_cluster(rack=0)
+        self._grow_cluster(rack=None)
+        self._shrink_cluster(rack=None)
 
     def disrupt_grow_shrink_new_rack(self):
         rack = max(self.cluster.racks) + 1
         self._grow_cluster(rack)
         self._shrink_cluster(rack)
 
-    def _grow_cluster(self, rack=0):
-        if rack > 0:
-            if not self._is_it_on_kubernetes():
-                raise UnsupportedNemesis("SCT rack functionality is implemented only on kubernetes")
-            self.log.info("Rack deletion is not supported on kubernetes yet. "
-                          "Please see https://github.com/scylladb/scylla-operator/issues/287")
+    def _grow_cluster(self, rack=None):
+        if rack is None and self._is_it_on_kubernetes():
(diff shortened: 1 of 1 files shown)


fgelcer:
[running](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/perf-acceptance-criteria/job/fix-rack-order/) to ensure it did fix the whole problem we just had

sdcm/nemesis.py
```diff
@@ -3485,20 +3485,17 @@ def decommission_nodes(self, add_nodes_number, rack, is_seed: Optional[Union[boo
                           f'{str(exc)}').publish()
 
     def disrupt_grow_shrink_cluster(self):
-        self._grow_cluster(rack=0)
-        self._shrink_cluster(rack=0)
+        self._grow_cluster(rack=None)
+        self._shrink_cluster(rack=None)
```
temichus, line 3489: you changed only _grow_cluster function logic tu support None and convert it to ) but not _shrink_cluster
and i see that 
_shrink_cluster calls decommission_nodes, that calls set_target_node, that calls _get_target_nodes

and putting None instead of 0 - might  change test behaviour
```
        if rack is not None:
            nodes = [node for node in nodes if node.rack == rack]
```

please test it before merging,
otherwise, it is better just to change disrupt_run_unique_sequence function for commit "improvement(perf during ops): change operations order' :
self._grow_cluster(rack=None) -> self._grow_cluster(rack=0) and
self._shrink_cluster(rack=None) -> self._grow_cluster(rack=0)

fruch (approved):
LGTM

even that this change is more then what's needed for it to be fixed.

sdcm/nemesis.py
fgelcer: https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/perf-acceptance-criteria/job/fix-rack-order/2/

fgelcer: the `_shrink_cluster` will run `decommission_nodes` in a loop, so it can't be using `self.target_node`, as the target node is only 1 (and here we will iterate 3 times in this for-loop).

changing the `rack` from `0` to `None` doesn't change anything, because in this test we don't use the concept of racks at all

fgelcer:
@temichus @aleksbykov , please review, so we can merge it, and run the job today, so we can analyse the results tomorrow

aleksbykov (approved):

Issue #6469, opened 2023-08-07
`create_index` and `wait_for_index_to_be_built` need to expect MV errors


fruch:
## Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

When `disrupt_create_index` is used during parallel nemesis, it should expect 
errors like the following, since some node can be down during MV/Index creation:
```
< t:2023-08-07 02:16:00,108 f:db_log_reader.py l:114  c:sdcm.db_log_reader   p:DEBUG > 2023-08-07T02:16:00+00:00 longevity-parallel-topology-schema--db-node-f9c7a7e5-3  !WARNING | scylla[5398]:  [shard 6] view - Error executing build step for base keyspace1.standard1: exceptions::unavailable_exception (Cannot achieve consistency level for cl ONE. Requires 1, alive 0)
```

## Installation details

Kernel Version: 5.15.0-1040-aws
Scylla version (or git commit hash): `2023.1.0~rc8-20230731.b6f7c5a6910c` with build-id `f6e718548e76ccf3564ed2387b6582ba8d37793c`

Cluster size: 5 nodes (i4i.2xlarge)

Scylla Nodes used in this run:
- longevity-parallel-topology-schema--db-node-f9c7a7e5-9 (54.75.126.66 | 10.4.1.97) (shards: 7)
- longevity-parallel-topology-schema--db-node-f9c7a7e5-8 (54.194.27.56 | 10.4.3.105) (shards: 7)
- longevity-parallel-topology-schema--db-node-f9c7a7e5-7 (63.33.208.200 | 10.4.1.200) (shards: 7)
- longevity-parallel-topology-schema--db-node-f9c7a7e5-6 (3.253.61.179 | 10.4.1.183) (shards: 7)
- longevity-parallel-topology-schema--db-node-f9c7a7e5-5 (3.248.218.96 | 10.4.3.90) (shards: 7)
- longevity-parallel-topology-schema--db-node-f9c7a7e5-4 (3.251.73.225 | 10.4.2.34) (shards: 7)
- longevity-parallel-topology-schema--db-node-f9c7a7e5-3 (34.249.188.75 | 10.4.0.186) (shards: 7)
- longevity-parallel-topology-schema--db-node-f9c7a7e5-21 (54.75.12.199 | 10.4.3.56) (shards: 7)
[...]

PR #6475, opened 2023-08-09
fix(wait_for_view_to_be_built): ignore some errors in parallel nemesis

soyacz:
In case test uses parallel nemesis, in case one node is down during create index/view, db throws errors like `Error applying view update`. We want to ignore such errors.

Added EventsFilter during wait for index/view to be built in situations we use parallel nemesis.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/6469

### DIFF
--- a/sdcm/utils/nemesis_utils/indexes.py
+++ b/sdcm/utils/nemesis_utils/indexes.py
@@ -14,6 +14,7 @@
 import logging
 import random
 import time
+from contextlib import nullcontext
 
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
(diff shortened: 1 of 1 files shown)


juliayakovlev (dismissed):

sdcm/utils/common.py
```diff
@@ -3152,3 +3152,9 @@ def SoftTimeoutContext(timeout: int, operation: str):  # pylint: disable=invalid
     if duration > timeout:
         SoftTimeoutEvent(operation=operation, soft_timeout=timeout,
                          duration=duration).publish_or_dump()
+
+
+@contextmanager
+def DummyContextManager():  # pylint: disable=invalid-name
```
vponomaryov, line 3158: I think it is redundant.
The `contextlib.nullcontext` does exactly the same and this lib is already used in the `sdcm/utils/nemesis_utils/indexes.py` module.

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -76,15 +77,20 @@ def wait_for_index_to_be_built(node: BaseNode, ks, index_name, timeout=300) -> N
...
-            InfoEvent(message=f"Index {ks}.{view_name} was built").publish()
-        if f"{ks}.{view_name} has finished building" in result.stdout:
-            InfoEvent(message=f"View/index {ks}.{view_name} was built").publish()
-            return
-        time.sleep(30)
+    with (EventsFilter(  # ignore apply view errors when multiple nemesis are running scylladb/scylla-cluster-tests#6469
+            event_class=DatabaseLogEvent.DATABASE_ERROR,
+            regex=".*view - Error applying view update to.*",
+            extra_time_to_expiration=180,
+    ) if node.parent_cluster.nemesis_count > 1 else DummyContextManager()):
```
vponomaryov, line 84: s/`DummyContextManager()`/`contextlib.nullcontext()`/

sdcm/utils/common.py
soyacz: thanks, I wasn't aware of it Fixed.

sdcm/utils/nemesis_utils/indexes.py
soyacz: fixed

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #6476, opened 2023-08-09
feature(KMS): add KMS key rotation on test level

vponomaryov:
With this change it becomes possible to enable KMS key rotation and dynamic alias creation for a test.
To do it just use `auto` as a value for the `kms_host` key in the value of the `scylla_encryption_options` config option. 
Example:

```
  scylla_encryption_options: "{ \
    'cipher_algorithm': 'AES/ECB/PKCS5Padding', \
    'secret_key_strength' : 128, \
    'key_provider': 'KmsKeyProviderFactory', \
    'kms_host': 'auto'}"
```

Also, there is no need to specify the `kms_hosts` info in the `append_scylla_yaml` config option as it was before. 
It is allowed and will work if some specific KMS key must be used, but rotation will not be done in this case.
The rotation is done only for the dynamically/automatically created KMS key alias.

### DIFF
--- a/configurations/kms-ear.yaml
+++ b/configurations/kms-ear.yaml
@@ -1,11 +1,2 @@
-
-scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'kms_test'}"
-
-# config kms_hosts for AWS KMS service account
-append_scylla_yaml: |
-  kms_hosts:
-      kms_test:
-          master_key: 'alias/kms_encryption_test'
-          aws_region: 'us-east-1'
-
+scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'auto'}"
 append_scylla_args: '--logger-log-level kms=trace'
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -71,9 +71,11 @@ from sdcm.remote.remote_file import remote_file, yaml_file_to_dict, dict_to_yaml
 from sdcm import wait, mgmt
 from sdcm.sct_config import SCTConfiguration
 from sdcm.sct_events.continuous_event import ContinuousEventsRegistry
+from sdcm.sct_events.system import AwsKmsEvent
 from sdcm.snitch_configuration import SnitchConfig
 from sdcm.utils import properties
 from sdcm.utils.adaptive_timeouts import Operations, adaptive_timeout
+from sdcm.utils.aws_kms import AwsKms
 from sdcm.utils.benchmarks import ScyllaClusterBenchmarkManager
 from sdcm.utils.common import (
     S3Storage,
@@ -418,6 +420,9 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             if any(substr in append_scylla_yaml for substr in (
                     "system_key_directory", "system_info_encryption", "kmip_hosts")):
                 install_encryption_at_rest_files(self.remoter)
+            for kms_host_name, kms_host_data in append_scylla_yaml.get("kms_hosts", {}).items():
+                if kms_host_data["aws_region"] == "auto":
+                    append_scylla_yaml["kms_hosts"][kms_host_name]["aws_region"] = self.region
             scylla_yml.update(append_scylla_yaml)
 
         return scylla_yml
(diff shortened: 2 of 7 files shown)


configurations/kms-ear.yaml
```diff
@@ -1,11 +1,3 @@
...
-scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'kms_test'}"
-
-# config kms_hosts for AWS KMS service account
-append_scylla_yaml: |
-  kms_hosts:
-      kms_test:
-          master_key: 'alias/kms_encryption_test'
-          aws_region: 'us-east-1'
-
+server_encrypt: false
```
vponomaryov, line 1: JFYI: there is some problem with the `server_encrypt: true`
So, disabled it to unblock the testing of the feature added here.

vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-EaR-longevity-kms-200gb-6h-test/13/consoleFull

configurations/kms-ear.yaml
fruch: which problem is that ? the fact it crashes after reboot ? 
if so please take the lastest enterprise image to test with.

vponomaryov: Yes, the very first reboot was causing a crash.
Trying latest with the `server_encrypt: true`

sdcm/cluster.py
```diff
@@ -418,6 +419,9 @@ def proposed_scylla_yaml(self) -> ScyllaYaml:
             if any(substr in append_scylla_yaml for substr in (
                     "system_key_directory", "system_info_encryption", "kmip_hosts")):
                 install_encryption_at_rest_files(self.remoter)
+            for kms_host_name, kms_host_data in append_scylla_yaml.get("kms_hosts", {}).items():
+                if kms_host_data["aws_region"] == "auto":
```
fruch, line 423: I don't understand this logic, where is the `aws_region` set as auto ?

fruch (dismissed):
LGTM

vponomaryov:
It is verified here (`us-east-1`): https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-EaR-longevity-kms-200gb-6h-test/15/
And here (`eu-north-1`): https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-EaR-longevity-kms-200gb-6h-test/14/
In both cases passed the `disrupt_mgmt_backup` nemesis among bunch of other ones.

sdcm/cluster.py
```diff
@@ -4385,6 +4389,34 @@ def stop_nemesis(self, timeout=10):
                 stack_trace = traceback.format_stack(current_thread_frames[nemesis_thread.ident])
                 threads_tracebacks.append("\n".join(stack_trace))
 
+    def start_kms_key_rotation_thread(self, rotation_interval_m=60) -> None:
```
fruch, line 4392: where this is coming from ? I don't remember we talk about introducing such a thread...

sdcm/cluster.py
```diff
@@ -4385,6 +4389,34 @@ def stop_nemesis(self, timeout=10):
...
+        append_scylla_yaml = yaml.safe_load(self.params.get("append_scylla_yaml")) or {}
+        for kms_host_name, kms_host_data in append_scylla_yaml.get("kms_hosts", {}).items():
+            if "auto" in kms_host_name:
+                kms_key_alias_name = kms_host_data["master_key"]
+                break
+        if not kms_key_alias_name:
+            self.log.debug("Automatic KMS key is not configured. Skip creation of it's rotation thread.")
+            return None
+
+        def _rotate_kms_key(kms_key_alias_name, rotation_interval_m):
```
fruch, line 4405: any thread we should decorate with `@raise_event_on_failure` otherwise we won't notice failures on it

sdcm/cluster.py
```diff
@@ -4385,6 +4389,34 @@ def stop_nemesis(self, timeout=10):
...
+                aws_kms.rotate_kms_key(kms_key_alias_name=kms_key_alias_name)
+
+        self.log.info("Start KMS key rotation thread for the '%s' alias", kms_key_alias_name)
+        aws_kms = AwsKms(region_names=self.params.region_names)
+        kms_key_rotation_thread = threading.Thread(
+            name='KmsKeyRotationThread',
+            target=_rotate_kms_key,
+            kwargs={"kms_key_alias_name": kms_key_alias_name, "rotation_interval_m": rotation_interval_m},
+            daemon=True)
+        kms_key_rotation_thread.start()
```
fruch, line 4417: shouldn't we implement something to stop this thread ? on teardown of the test ?

vponomaryov: I don't see practical need here. It sleeps for long time and then makes just 1 API call.

sdcm/cluster.py
fruch: I thought it would be a nemesis, like any other disruption we are doing
doing it on it's own thread meaning we have less information when it's happening (i.e. no events, like nemesis)

and if we want it in parallel to other nemesis, we cloud use parallel nemesis

sdcm/cluster.py
fruch: if an api call would fail, the whole thread would stop, and we won't notice it.

fruch: failures should go unnoticed

fruch: I would argue the while loop should have a try/except clause

sdcm/cluster.py
vponomaryov: > I thought it would be a nemesis, like any other disruption we are doing doing it on it's own thread 

We never clarified the implementation of the rotation in our discussions. `I thought ` about a thread...

> meaning we have less information when it's happening (i.e. no events, like nemesis)

I can add an event creation here if really needed. For the moment it has a log message.

> and if we want it in parallel to other nemesis, we cloud use parallel nemesis

Why bother with nemesis? What is the gain?

sdcm/cluster.py
vponomaryov: Ok, will do.

sdcm/cluster.py
fruch: also we aren't doing the flow with calling `nodetool upgradesstable` in here at all.
and we probably want that case as well

sdcm/cluster.py
vponomaryov: > failures should go unnoticed

The `not`-word-thief has returned!

> I would argue the while loop should have a try/except clause

A can add try-block, not a problem.

sdcm/cluster.py
fruch: in a nemesis we have all the machinery in place, events, error handling, reporting, healthchecks.
and some degree of control on selecting them.

you are correctly is doesn't communicate directly with any specific node, so it's not the classic nemesis.

vponomaryov: > also we aren't doing the flow with calling `nodetool upgradesstable` in here at all. and we probably want that case as well

In the scope of the whole test run that longs for hours it is redundant.
The compactions, which happen all the time, will do the job we need - update sstables.

sdcm/tester.py
```diff
@@ -874,6 +900,7 @@ def setUp(self):  # pylint: disable=too-many-branches,too-many-statements
             if db_cluster:
                 db_cluster.validate_seeds_on_all_nodes()
                 validate_raft_on_nodes(nodes=db_cluster.nodes)
+                db_cluster.start_kms_key_rotation_thread()
```
fruch, line 903: can get add SCT configuration for setting `rotation_interval_m` ? i'm not sure a default 60m would be suitable for all cases ?

fruch (dismissed):
LGTM,
I only think we should have configuration for the rotation interval

sdcm/tester.py
vponomaryov: Done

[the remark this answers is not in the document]
vponomaryov: The separate class for handling of the KMS is already existing -> `sdcm/utils/aws_kms.py`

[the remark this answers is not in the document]
soyacz: handling KMS is one thing, making it working is the second and the third is putting it to use. 
I think just `cluster` being a bucket for everything that doesn't fit to precise classes is not good. We could have another class for handling this as a thread and just use it (probably just in tester class, which is operating the test)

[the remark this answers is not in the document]
vponomaryov: > handling KMS is one thing, making it working is the second and the third is putting it to use. I think just `cluster` being a bucket for everything that doesn't fit to precise classes is not good. We could have another class for handling this as a thread and just use it (probably just in tester class, which is operating the test)

The solution sounds as out of scope for this PR.
We all know that the SCT has a lot of problems. But improvement of these problems should not block new features coverage and solving business needs...
So, I ok to have such separation, but not in this PR.

vponomaryov:
@fruch ?

sdcm/tester.py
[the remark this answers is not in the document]
fruch: @vponomaryov  Let add it as an improvement task/issue

fruch (approved):
LGTM

PR #6486, opened 2023-08-11
Update monitoring branch and Fix Overview screenshot

k0machi:
- test(test_default.yaml): set monitoring version to 4.5
- fix(monitorstack/ui.py): Fix panel name for 4.4/4.5 monitoring

Closes #6408 

### DIFF
--- a/sdcm/utils/docker_utils.py
+++ b/sdcm/utils/docker_utils.py
@@ -17,7 +17,7 @@ import logging
 import warnings
 from pprint import pformat
 from types import SimpleNamespace  # pylint: disable=no-name-in-module
-from typing import List, Optional, Union, Any, Tuple
+from typing import List, Optional, Union, Any, Tuple, Protocol
 from functools import cache
 
 import docker
@@ -76,6 +76,12 @@ class _Name(SimpleNamespace):
 
     def __bool__(self):
         return self.full is not None
+
+
+class INodeWithContainerManager(Protocol):
+    # pylint: disable=too-few-public-methods
+    _containers: dict[str, Container]
+    tags: dict[str, str]
 
 
 class ContainerManager:  # pylint: disable=too-many-public-methods)
@@ -141,18 +147,18 @@ class ContainerManager:  # pylint: disable=too-many-public-methods)
     default_docker_client = _docker
 
     @classmethod
-    def get_docker_client(cls, instance: object, name: Optional[str] = None) -> DockerClient:
+    def get_docker_client(cls, instance: INodeWithContainerManager, name: Optional[str] = None) -> DockerClient:
         container = None if name is None else cls.get_container(instance, name, raise_not_found_exc=False)
         if container is None:
             return cls._get_docker_client_for_new_container(instance, _Name(name))
         return container.client
 
     @classmethod
-    def _get_docker_client_for_new_container(cls, instance: object, name: _Name) -> DockerClient:
+    def _get_docker_client_for_new_container(cls, instance: INodeWithContainerManager, name: _Name) -> DockerClient:
         return cls._get_attr_for_name(instance, name, "docker_client", default=cls.default_docker_client)
 
     @staticmethod
-    def _get_attr_for_name(instance: object,
+    def _get_attr_for_name(instance: INodeWithContainerManager,
                            name: _Name,
                            attr: str,
                            default: Optional[Any] = None,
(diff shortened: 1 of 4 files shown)


k0machi:
There's another issue that I found is that the "Publish snapshot to raintank.io" feature seems to be missing, but I'm not sure if that's being used yet, so I haven't fixed it yet.

k0machi:
[Argus](https://argus.scylladb.com/test/b0b22734-758e-47b4-b6ec-29e64d212d52/runs?additionalRuns[]=a9f0932b-a8da-4bd5-96bc-f2788f226061)

fruch:
> [Argus](https://argus.scylladb.com/test/b0b22734-758e-47b4-b6ec-29e64d212d52/runs?additionalRuns%5B%5D=a9f0932b-a8da-4bd5-96bc-f2788f226061)

those jobs still show only one screenshot...

fruch (changes requested):
still missing one of the screenshots

fruch:
> > [Argus](https://argus.scylladb.com/test/b0b22734-758e-47b4-b6ec-29e64d212d52/runs?additionalRuns%5B%5D=a9f0932b-a8da-4bd5-96bc-f2788f226061)
> 
> those jobs still show only one screenshot...

@k0machi also you don't need such long job to get those screenshots, you can run the docker provision test (or aws/gce ones on this PR)

k0machi:
Okay, I've corrected the locators for real this time, there seems to be an issue with screenshots specifically for argus, looking into that now, as monitor-set.tar.gz contains both screenshots

k0machi:
And after further investigation it appears that screenshots are fixed as long as there are no failures during the run or if the test takes too long - You can see that `#57` in Argus has both screenshots without any changes, it appears that there's an issue with Docker Container object, because overview screenshot fails instantly with "Cannot get screenshot" when trying to call `self.browser.set_window_size`

k0machi:
Provision test seems to be passing with both screenshots, running my staging job to check if adding a retry for docker reload fixes the issue.

soyacz:
> Okay, I've corrected the locators for real this time, there seems to be an issue with screenshots specifically for argus, looking into that now, as monitor-set.tar.gz contains both screenshots.
> 
> UPD: I can see that email however is also missing a second screenshot, so problem is somewhere between email_data and collecting them

IIRC we collect screenshots 2 times - in teardown and in 'collect-logs' stage. So looks that it works in teardown and breaks the next collection. Maybe we don't close properly browsers (or other similar problem)?

k0machi:
> IIRC we collect screenshots 2 times - in teardown and in 'collect-logs' stage. So looks that it works in teardown and breaks the next collection. Maybe we don't close properly browsers (or other similar problem)?

To me it seems that docker daemon stops working for some reason, maybe something shuts it off? One thing is that it's only the overview screenshot that fails - eventually when the collector moves to the per-nemesis screenshot, the daemon suddenly responds and works. This seems to not happen if a longevity is short - I've shortened `#57` to one hour and both screenshots got collected fine. What I can see is that overview screenshot seems to be collected fairly often, not just in teardown and collect logs stage, there are at least 5-6 spots where it's collected successfuly (during nemeses I think).

sdcm/utils/docker_utils.py
```diff
@@ -176,15 +176,23 @@ def _get_attr_for_name(instance: object,
...
         container = instance._containers.get(name)
         if container is None and raise_not_found_exc:
             raise NotFound(f"There is no container `{name}' for {instance}.")
         if container:
-            container.reload()
+            try:
+                container.reload()
+            except OSError:
+                LOGGER.warning("Container failed to refresh, checking docker connection...")
+                docker_client = cls.get_docker_client(instance, name)
```
soyacz, line 191: `cls.get_docker_client` if it gets `name` param it runs `cls.get_container` - so we have recurrence here :)
I propose to not provide name for `cls.get_docker_client` in that case.

k0machi: I've added code for that, let's see how this will behave now.

k0machi:
Latest run contains both screenshots, however the exception handler didn't start in it, so may still not be fixed, but looks good so far.

soyacz:
> Latest run contains both screenshots, however the exception handler didn't start in it, so the docker issue may still not be fixed, but looks good so far.

ok, so let's merge this one and handle docker issues (if they appear) in different PR.

soyacz (approved):

sdcm/utils/docker_utils.py
```diff
@@ -78,6 +78,12 @@ def __bool__(self):
         return self.full is not None
 
 
+class INodeWithContainerManager(Protocol):
```
aleksbykov, line 81: Why this needed as Protocol?

aleksbykov (approved):
LGTM, short question

sdcm/utils/docker_utils.py
k0machi: Protocol is here for static type checking, it's a class that describes an interface

fruch (approved):
LGTM

PR #6489, opened 2023-08-11
improvement(kms): add one more variation of the KMS nemesis

vponomaryov:
Before we had only one variation - without KMS key rotation.
So, add one more - with KMS key rotation.

### DIFF
--- a/data_dir/nemesis.yml
+++ b/data_dir/nemesis.yml
@@ -56,7 +56,10 @@
 - disrupt_drain_kubernetes_node_then_replace_scylla_node:
   - disruptive = True
   - kubernetes = True
-- disrupt_enable_disable_table_encryption_aws_kms_provider:
+- disrupt_enable_disable_table_encryption_aws_kms_provider_with_rotation:
+  - disruptive = True
+  - kubernetes = False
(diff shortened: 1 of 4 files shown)


data_dir/nemesis.yml
```diff
@@ -112,7 +112,7 @@
   - limited = True
 - disrupt_mgmt_restore:
   - disruptive = True
-  - kubernetes = False
```
fruch, line 115: why are touching this one ?

fruch:
@vponomaryov rerun the unittest that create the nemesis list, the new nemesis are missing from it

data_dir/nemesis.yml
fruch: I'm guessing someone update it,and didn't run the unittest again....

vponomaryov:
> @vponomaryov rerun the unittest that create the nemesis list, the new nemesis are missing from it

The nemesis list you are referring to considers only those classes where we call disruption methods directly.
So, this change is correct according to the existing logic.

data_dir/nemesis.yml
vponomaryov: Yes, this diff is in master code.

fruch (approved):
LGTM

PR #6452, opened 2023-08-02
fix(upgrade_tests): write data for validation with CL=ALL

fruch:
since our foucs is to vaildate data loss/corruption cause by the upgrade, we should make sure we write the data with CL=ALL, to make sure we are less susceptible
to temporey issues (like disk hiccups and such), that c-s default retry (10 times) should able to overcome.

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6493, opened 2023-08-13
Revert "fix(upgrade_tests): write data for validation with CL=ALL"

fruch:
Reverts scylladb/scylla-cluster-tests#6452

Looks like the `stress_before_upgrade` is actually happing during upgrades and using cl=all breaks all rolling upgrades.


fgelcer (approved):
LGTM

fgelcer:
@fruch , once you backport to 2023.1 and 5.2, LMK so i can re-trigger the rolling upgrade jobs that failed because of it

fruch:
> @fruch , once you backport to 2023.1 and 5.2, LMK so i can re-trigger the rolling upgrade jobs that failed because of it

I have backported it, for triggering you might ask releng which job to trigger

fgelcer:
> > @fruch , once you backport to 2023.1 and 5.2, LMK so i can re-trigger the rolling upgrade jobs that failed because of it
> 
> I have backported it, for triggering you might ask releng which job to trigger

i will trigger them, thank you

mykaul:
I'm unsure that this is the right way to handle it. If I understand correctly, you can't use CL=ALL because some nodes are unavailable?

fruch:
> I'm unsure that this is the right way to handle it. If I understand correctly, you can't use CL=ALL because some nodes are unavailable?

the initial assumption of the change I've did was that `stress_before_upgrade` is done while all node are up, so we can use ALL
but it wasn't true in all of the test using it, and in some the `stress_before_upgrade` was used while the 1st/2nd node are be upgrades, we can use ALL while nodes are down.

we'll need to align the test code that `stress_before_upgrade` would be use before the upgrade, and a different command with QURUM could be used during the upgrade.

PR #6501, opened 2023-08-15
test(k8s): add new functional test to cover DB node cleanups

vponomaryov:
Operator `v1.10` runs cleanups [1] on each of the old nodes after adding a new one like it is recommended by the Scylla documentation [2]. So, cover this functionality with the functional test.

[1] scylladb/scylla-operator/pull/1294
[2] https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/add-node-to-cluster.html

### DIFF
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -36,6 +36,7 @@ from sdcm.cluster_k8s import (
     SCYLLA_OPERATOR_NAMESPACE
 )
 from sdcm.mgmt import TaskStatus
+from sdcm.utils.common import ParallelObject
 from sdcm.utils.k8s import (
     convert_cpu_units_to_k8s_value,
     convert_cpu_value_from_k8s_to_units,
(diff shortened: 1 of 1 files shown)


vponomaryov:
This test becomes needed in the upcoming `operator-v1.10` testing which is going to start this week.

vponomaryov:
The new test has passed successfully.
Search for the `::test_add_new_node_and_check_old_nodes_are_cleaned_up`: https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-6501/2/execution/node/115/log/?consoleFull

functional_tests/scylla_operator/test_functional.py
```diff
@@ -154,6 +155,44 @@ def test_rolling_restart_cluster(db_cluster):
...
+    log_followers = {}
+    for node in db_cluster.nodes:
+        for keyspace in db_cluster.nodes[0].run_cqlsh('describe keyspaces').stdout.split():
+            log_followers[f"{node.name}--{keyspace}"] = node.follow_system_log(patterns=[
+                f"api - force_keyspace_cleanup: keyspace={keyspace} "])
+
+    def wait_for_cleanup_logs(log_follower_name, log_follower, db_cluster):
+        db_rf = len(db_cluster.nodes)
+        while not list(log_follower):
+            # NOTE: log lines located as last may not be caught fast enough because of the chunk-size limitation.
```
soyacz, line 169: hmm, can we `echo "some log message" >> <path to scylla log> to achieve it?

vponomaryov: It is done here by creating and dropping a keyspace.

soyacz: yes, I see it, but was wondering if we could have done it in a way that does not change db schema...

soyacz (approved):
LGTM

vponomaryov:
@scylladb/qa-maintainers Please, review it and merge. We will need it in the upcoming scylla-operator release testing.

vponomaryov:
The functional tests become working with this PR: https://github.com/scylladb/scylla-cluster-tests/pull/6503

soyacz:
> > LGTM, only question: in PR there is mention that after remove node this procedure also should be run, doest additional test with node removal need?
> 
> If I understood you correctly that you propose to add new steps to some of the existing tests instead of adding a completely new one, then yes, it will work too. Looked at the list of the existing tests, and the `test_orphaned_services_after_shrink_cluster` looks the most suitable for it. So, it is question whether we want to mix various checks into one test or not.
> 
> @soyacz WDYT? @aleksbykov thanks for the question!

There are always pros and cons. What I see:
pros:
1. one test less, faster execution, lower costs, 
2. this cleanup is a part of add_node procedure in operator (which raises question, if we shouldn't add this wait to add_node to not start e.g. next nemesis when cleanup is still in progress?)
cons:
1. more work (need to verify also if operator is in 1.10 version before this check), but rather not much
2. in case this fails it may hide other errors which are more important (this error could then just raise error event without stopping the test flow to avoid that problem)

my verdict is to make it as a part existing test, but I'm not forcing it to be in this PR - can create another task for this refactor.

aleksbykov:
> > LGTM, only question: in PR there is mention that after remove node this procedure also should be run, doest additional test with node removal need?
> 
> If I understood you correctly that you propose to add new steps to some of the existing tests instead of adding a completely new one, then yes, it will work too. Looked at the list of the existing tests, and the `test_orphaned_services_after_shrink_cluster` looks the most suitable for it. So, it is question whether we want to mix various checks into one test or not.
> 
> @soyacz WDYT? @aleksbykov thanks for the question!

Not exactly. New test is good enough, i just want to say, that PR said that this functionality is added in 2 cases: when node added ( it is covered by PR) and when node removed (this i didn't see in PR). So question was, should be added second functional test , which will check that keyspaces cleaned after node removed

vponomaryov:
> > > LGTM, only question: in PR there is mention that after remove node this procedure also should be run, doest additional test with node removal need?
> > 
> > 
> > If I understood you correctly that you propose to add new steps to some of the existing tests instead of adding a completely new one, then yes, it will work too. Looked at the list of the existing tests, and the `test_orphaned_services_after_shrink_cluster` looks the most suitable for it. So, it is question whether we want to mix various checks into one test or not.
> > @soyacz WDYT? @aleksbykov thanks for the question!
> 
> Not exactly. New test is good enough, i just want to say, that PR said that this functionality is added in 2 cases: when node added ( it is covered by PR) and when node removed (this i didn't see in PR). So question was, should be added second functional test , which will check that keyspaces cleaned after node removed

Only addition of a new node requires `cleanup` (`cleanup` of data that moved to a new node).
Removal of a node leads only to the extending of data and doesn't require the `cleanup` operation.

aleksbykov:
The description of https://github.com/scylladb/scylla-operator/pull/1294 said:
```
Running node cleanup after scaling a cluster allows to avoid the
accumulation of unnecessary data on the node disks. When nodes are added
or removed from the cluster, they gain or lose some tokens, which can
```

vponomaryov:
> The description of [scylladb/scylla-operator#1294](https://github.com/scylladb/scylla-operator/pull/1294) said:
> 
> ```
> Running node cleanup after scaling a cluster allows to avoid the
> accumulation of unnecessary data on the node disks. 

This test covers this for 100%.

> ```
> When nodes are added or removed from the cluster, they gain or lose some tokens, which can

This just description where we know that nodes gain or lose tokens.
And the `cleanup` is needed only when `old` nodes lose token, which move to a `new` node.

fruch (approved):
LGTM

PR #6514, opened 2023-08-17
fix(nemesis) check length of nodetool getendpoints output 

lattwood:
There were no hits for getendpoints in this repository, which returns the addresses/names of replicas that own a given key, and tests just verified the replication configuration didn't change by inspecting the schema. This validates the replication setting on the KEYSPACE after a node replacement by getting the cluster to calculate the replicas for a given key before and after node replacement. This ensures the effective replication configuration has stayed the same.

I believe this would have caught the issue in scylladb/scylladb#15017 prior to 5.2.0-rc0 being released.

Given I don't have access to your Jenkins instance, I won't be able to validate CI here.

---

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1462,10 +1462,18 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             states = [val['state'] for dc in status.values() for ip, val in dc.items() if ip == node_ip]
             return states[0] if states else None
 
+        def nodetool_getendpoints(keyspace: str, table: str, key: str) -> List["str"] | None:
+            node = random.choice(self.cluster.nodes)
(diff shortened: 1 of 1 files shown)


soyacz:
thanks @lattwood !
This looks ok. 
I was looking to reproduce issue with our dtest framework here: https://github.com/scylladb/scylla-dtest/blob/next/replace_address_test.py - they're faster and easier to work with and usually preferred - especially when issue does not require high cluster load.
But actually couldn't confirm your issue... thus I'm not sure if this also help.

lattwood:
Oh, I don't have visibility into the dtest framework as that's your private repository.

It's also looking like the bug I ran into was around related to this, except I had 2 nodes in all DCs so just weird things happened instead of crashing https://github.com/scylladb/scylladb/issues/14184

Issue #6495, opened 2023-08-14
artifacts-ami-arm-test: Failing due to wrong availability zone


yaronkaikov:
`
[...]

PR #6518, opened 2023-08-18
fix(provision): fix fallback to next availability zone

soyacz:
When using instance type that is not available in all availability zones for given region, provision may fail. This is because we try provision in az's starting from first az. When first az does not support given instance type it fails with error `Spot request cannot be fulfilled due to invalid availability zone.`

Fix is by filtering az's that support provided instance type.

fixes: #6495 

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -1210,7 +1210,8 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         def _get_all_zones_common_params() -> list[dict]:
             all_zones_common_params = []
             aws_region = AwsRegion(region_name=regions[0])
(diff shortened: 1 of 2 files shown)


soyacz:
test run: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-ami-arm-test/2/

fgelcer (approved):
LGTM

roydahan (approved):

roydahan:
LGTM, we can also add a retry with a different AZ (if we don't have it already).

soyacz:
> LGTM, we can also add a retry with a different AZ (if we don't have it already).

Actually the path with this feature was broken, so hopefully it will work now. 
We should be able to see it when running with spot.

PR #6520, opened 2023-08-20
fix(upgrade): not fail on truncated entries timeouts

juliayakovlev:
Read from truncated table fails with timeout (60sec) from time to time in case of mixed cluster (one node upgraded from OSS to 2023.1).
Add 'adaptive_timeout' to insert into and  read from truncated tables to be able to check how long time the query runs.
Also add try .. catch for insert into truncate tables to prevent test stop on timeout.

### DIFF
--- a/sdcm/utils/adaptive_timeouts/__init__.py
+++ b/sdcm/utils/adaptive_timeouts/__init__.py
@@ -32,6 +32,13 @@ def _get_soft_timeout(node_info_service: NodeLoadInfoService, timeout: int | flo
         return timeout, {}
 
 
+def _get_query_timeout(node_info_service: NodeLoadInfoService, timeout: int | float = None, query: str = None) -> \
+        tuple[int | float, dict[str, Any]]:
+    timeout, stats = _get_soft_timeout(node_info_service=node_info_service, timeout=timeout)
+    stats["query"] = query
+    return timeout, stats
+
+
 class Operations(Enum):
     """Available operations for adaptive timeouts. Each operation maps to a function to calculate timeout based on node load info.
 
@@ -48,6 +55,7 @@ class Operations(Enum):
     SCRUB = ("scrub", _get_soft_timeout, ("timeout",))
     SOFT_TIMEOUT = ("soft_timeout", _get_soft_timeout, ("timeout",))
     FLUSH = ("flush", _get_soft_timeout, ("timeout",))
+    QUERY = ("query", _get_query_timeout, ("timeout", "query"))
 
 
 @contextmanager
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -21,6 +21,7 @@ import logging
 import random
 import time
 import re
+import traceback
 
 from collections import OrderedDict
 from uuid import UUID
@@ -30,7 +31,10 @@ from cassandra.util import sortedset, SortedSet  # pylint: disable=no-name-in-mo
 from cassandra import ConsistencyLevel
 from cassandra.protocol import ProtocolException  # pylint: disable=no-name-in-module
 
+from sdcm.sct_events import Severity
+from sdcm.sct_events.system import InfoEvent
 from sdcm.tester import ClusterTester
+from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
 from sdcm.utils.decorators import retrying
 from sdcm.utils.cdc.options import CDC_LOGTABLE_SUFFIX
 from sdcm.utils.version_utils import ComparableScyllaVersion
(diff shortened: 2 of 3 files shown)


juliayakovlev:
test run: https://jenkins.scylladb.com/job/scylla-staging/job/yulia/job/rolling_upgrade_with_sla/16/
Test failed but it not connected to changes, the code change passed

roydahan (approved):

soyacz (commented):
there might be some problem with reading these stats as there's no distinction between other `SOFT_TIMEOUT`s.
Two things I see could be done for that:
1. add another `Operation` 
2. extend data when storing to ES - e.g. add `config_files`, so we could filter by specific test. (this should be done in separate PR/commit)

juliayakovlev:
> there might be some problem with reading these stats as there's no distinction between other `SOFT_TIMEOUT`s. Two things I see could be done for that:
> 
> 1. add another `Operation`
> 2. extend data when storing to ES - e.g. add `config_files`, so we could filter by specific test. (this should be done in separate PR/commit)

Done

juliayakovlev:
Test passed: https://jenkins.scylladb.com/job/scylla-staging/job/yulia/job/rolling_upgrade_with_sla/17/
There is an error event but it not caused by this commit. Test was continue to run and was not stopped

juliayakovlev:
@soyacz @roydahan @fgelcer 
WDYT - adaptive_timeout.timeout = 300 is it not so big?

sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -48,6 +55,7 @@ class Operations(Enum):
     SCRUB = ("scrub", _get_soft_timeout, ("timeout",))
     SOFT_TIMEOUT = ("soft_timeout", _get_soft_timeout, ("timeout",))
     FLUSH = ("flush", _get_soft_timeout, ("timeout",))
+    QUERY = ("query", _get_query_timeout, ("timeout",))
```
soyacz, line 58: you missed "query" parameter - this causes ES to not get the value for it.
See https://70f106c98484448dbc4705050eb3f7e9.us-east-1.aws.found.io:9243/app/discover#/?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:now-15m,to:now))&_a=(columns:!(),filters:!(),index:'94fb1d70-eb66-11ed-b96a-256b7e8b2261',interval:auto,query:(language:kuery,query:'operation%20:%20QUERY'),sort:!())

soyacz:
> @soyacz @roydahan @fgelcer WDYT - adaptive_timeout.timeout = 300 is it not so big?

this is only soft timeout, I think it could be this 60s, but make the real timeout longer - this way test still fails after 60s (with SOFT TIMEOUT event) but query will continue and actual time will be measured and query-able in ES.

sdcm/utils/adaptive_timeouts/__init__.py
juliayakovlev: fixed

juliayakovlev:
> > @soyacz @roydahan @fgelcer WDYT - adaptive_timeout.timeout = 300 is it not so big?
> 
> this is only soft timeout, I think it could be this 60s, but make the real timeout longer - this way test still fails after 60s (with SOFT TIMEOUT event) but query will continue and actual time will be measured and query-able in ES.

I want to see how long time the query will take. I increased `session.execute` timeout

soyacz:
> > > @soyacz @roydahan @fgelcer WDYT - adaptive_timeout.timeout = 300 is it not so big?
> > 
> > 
> > this is only soft timeout, I think it could be this 60s, but make the real timeout longer - this way test still fails after 60s (with SOFT TIMEOUT event) but query will continue and actual time will be measured and query-able in ES.
> 
> I want to see how long time the query will take. I increased `session.execute` timeout

You can make adaptive_timeout.timeout = 60 so it will be visible in test run when query exceeded it.

juliayakovlev:
> > > > @soyacz @roydahan @fgelcer WDYT - adaptive_timeout.timeout = 300 is it not so big?
> > > 
> > > 
> > > this is only soft timeout, I think it could be this 60s, but make the real timeout longer - this way test still fails after 60s (with SOFT TIMEOUT event) but query will continue and actual time will be measured and query-able in ES.
> > 
> > 
> > I want to see how long time the query will take. I increased `session.execute` timeout
> 
> You can make adaptive_timeout.timeout = 60 so it will be visible in test run when query exceeded it.

Done

juliayakovlev:
@roydahan @soyacz @fgelcer please, review

soyacz (approved):
LGTM

roydahan (approved):

roydahan:
Let's wait for the weekend runs and if all good, backport to 2023.1 and 5.2

juliayakovlev:
The tests were not triggered during weekend yet

juliayakovlev:
@fruch It works. I think we can backport it

PR #6537, opened 2023-08-22
fix(k8s): make perf thrpt job be working

vponomaryov:
Some time ago the `append_scylla_args` option was added to the general config we use for testing throughput on K8S:

```
  test-cases/performance/perf-regression.100threads.30M-keys.yaml
```

And it's values are not compatible with the scylla-operator. 
So, redefine that config option to make Scylla work with this config.

### DIFF
--- a/configurations/operator/perf-regression-throughput.yaml
+++ b/configurations/operator/perf-regression-throughput.yaml
@@ -1,0 +1,4 @@
+# NOTE: main configuration adds '--blocked-reactor-notify-ms' arg
+#       which gets set by the scylla-operator itself. So, redefine this config option
+#       by removing the duplication of the scylla arg which will make Scylla fail.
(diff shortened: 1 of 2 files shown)


vponomaryov:
Without PR: https://jenkins.scylladb.com/job/scylla-operator/job/operator-1.10/job/performance/job/perf-regression-throughput-eks/2
With PR: https://jenkins.scylladb.com/job/scylla-operator/job/operator-1.10/job/performance/job/perf-regression-throughput-eks/4

soyacz (approved):
LGTM
This is a quick fix, but in future if any other job will set this param it will cause error again.
Maybe we should tackle it by `SCTConfiguration` where we would remove scylla args set by scylla-operator (and raise some warning)?

juliayakovlev (approved):

Issue #6543, opened 2023-08-27
[disrupt_mgmt_restore] c-s fails with: "Error while computing token map for keyspace 100gb_sizetiered_5_2 with datacenter"


yarongilor:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

The new nemesis of `disrupt_mgmt_restore` got to the last step of read verification.
Then c-s read stress failed with:
```
< t:2023-08-05 04:35:53,800 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2023-08-05 04:35:53.792: (CassandraStressEvent Severity.ERROR) period_type=end event_id=e6145591-c7e1-408e-9bd7-0cd408db8095 during_nemesis=MgmtRestore duration=57s: node=Node longevity-mv-si-4d-master-loader-node-141f4ec7-1 [52.209.11.43 | 10.4.10.226] (seed: False)
< t:2023-08-05 04:35:53,800 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > stress_cmd=cassandra-stress read cl=QUORUM n=52428800 -schema 'keyspace=100gb_sizetiered_5_2 replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..52428800
< t:2023-08-05 04:35:53,800 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > errors:
< t:2023-08-05 04:35:53,800 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 
< t:2023-08-05 04:35:53,800 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > Stress command completed with bad status 1: WARN  04:35:00,473 Error while computing token map for keyspace 100gb_sizetiered_5_2 with datacenter

```

## Impact

The c-s failure caused test to finish.

## How frequently does it reproduce?

reproducible.

## Installation details

Kernel Version: 5.15.0-1040-aws
Scylla version (or git commit hash): `5.4.0~dev-20230730.14dee7a94602` with build-id `432eaee3090f86b6465505b3fe6ac145fddc5381`

Cluster size: 5 nodes (i4i.4xlarge)

[...]

PR #6544, opened 2023-08-27
fix(stress_thread): `stop_test_on_failure` affect log parsing

fruch:
when `stop_test_on_failure=False` is used by nemesis code, it was affecting only the return value check to not raise critical events, but if any log line that was raising critical events would still be raised and stop the test in the middle.

this change propgates the `stop_test_on_failure` into the log parsing code to act inlined with the rest of the error handling logic

Ref: #6543

### Testing
- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/6

### DIFF
--- a/sdcm/stress_thread.py
+++ b/sdcm/stress_thread.py
@@ -23,7 +23,7 @@ from pathlib import Path
 from functools import cached_property
 
 from sdcm.loader import CassandraStressExporter, CassandraStressHDRExporter
-from sdcm.cluster import BaseLoaderSet  # , BaseNode
+from sdcm.cluster import BaseLoaderSet
 from sdcm.prometheus import nemesis_metrics_obj
 from sdcm.sct_events import Severity
 from sdcm.utils.common import FileFollowerThread, get_data_dir_path, time_period_str_to_seconds, SoftTimeoutContext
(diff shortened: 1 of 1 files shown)


ShlomiBalalis (approved):

sdcm/stress_thread.py
```diff
@@ -63,6 +64,8 @@ def run(self) -> None:
 
                     if pattern.search(line):
                         event.add_info(node=self.node, line=line, line_number=line_number).publish()
+                        if event.severity == Severity.CRITICAL and not self.stop_test_on_failure:
```
juliayakovlev, line 67: Do we need same for scylla-bench?

fruch: I'm not sure we have critical log events in scylla-bench.

I'll check tomorrow if we have any

juliayakovlev: AFAIR we can get critical event when stress command fails

soyacz (dismissed):
LGTM

sdcm/stress_thread.py
fruch: scylla-bench only has this log error:
```
ScyllaBenchLogEvent.add_subevent_type("ConsistencyError", severity=Severity.ERROR, regex=r"received only")
```
so no criticals, and no need for implementing there for now.

juliayakovlev: See this run: https://argus.scylladb.com/test/3190b0e9-a489-4e19-b855-32dc837b5431/runs?additionalRuns[]=e9fb4ce4-bec9-4429-995e-fc13f77a27bb

fruch: as I said it's not log event, and we might have an issue with s-b logic that we should look at on it's own issue/PR

fruch:
Last run now proved this fix is working as expected

sdcm/stress_thread.py
```diff
@@ -62,6 +63,9 @@ def run(self) -> None:
                         event.event_id = self.event_id
 
                     if pattern.search(line):
+                        if event.severity == Severity.CRITICAL and not self.stop_test_on_failure:
+                            event = event.clone()  # so we don't change the severity to other stress threads
```
fruch, line 67: Initially I've change the level only after publishing, which wasn't good enough

soyacz (approved):
LGTM

PR #6548, opened 2023-08-29
feature(sct): add support for using dns names in cluster

soyacz:
We want to test if Scylla works with dns names instead of IP addresses.

Adding proper methods and conditions in various places according to test configuration.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -251,6 +251,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
 
         self.remoter: Optional[RemoteCmdRunnerBase] = None
 
+        self._use_dns_names: bool = parent_cluster.params.get('use_dns_names') or False
         self._spot_monitoring_thread = None
         self._journal_thread = None
         self._docker_log_process = None
@@ -803,6 +804,14 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             self._public_ip_address_cached = self._get_public_ip_address()
         return self._public_ip_address_cached
 
+    @property
+    def public_dns_name(self) -> str:
+        return self.name
+
+    @property
+    def private_dns_name(self) -> str:
+        return self.name
+
     def _get_public_ip_address(self) -> Optional[str]:
         public_ips, _ = self._refresh_instance_state()
         if public_ips:
@@ -894,6 +903,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
 
         Use it for localhost connections (e.g., cqlsh)
         """
+        if self._use_dns_names:
+            return self.private_dns_name
         return self.ip_address
 
     @property
@@ -3951,7 +3962,10 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
 
     @property
     def seed_nodes_ips(self):
-        seed_nodes_ips = [node.ip_address for node in self.nodes if node.is_seed]
+        if self.params.get('use_dns_names'):
+            seed_nodes_ips = [node.private_dns_name for node in self.nodes if node.is_seed]
+        else:
+            seed_nodes_ips = [node.ip_address for node in self.nodes if node.is_seed]
         assert seed_nodes_ips, "We should have at least one selected seed by now"
         return seed_nodes_ips
 
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -251,7 +251,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
 
         self.remoter: Optional[RemoteCmdRunnerBase] = None
 
-        self._use_dns_names: bool = parent_cluster.params.get('use_dns_names') or False
+        self._use_dns_names: bool = parent_cluster.params.get('use_dns_names') if parent_cluster else False
         self._spot_monitoring_thread = None
         self._journal_thread = None
         self._docker_log_process = None
@@ -434,8 +434,9 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
     def refresh_ip_address(self):
         # Invalidate ip address cache
         self._private_ip_address_cached = self._public_ip_address_cached = self._ipv6_ip_address_cached = None
-        self.__dict__.pop('cql_ip_address', None)
-
+        self.__dict__.pop('cql_address', None)
+        self.__dict__.pop('public_dns_name', None)
+        self.__dict__.pop('private_dns_name', None)
         if self.ssh_login_info["hostname"] == self.external_address:
             return
 
@@ -861,7 +862,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         raise NotImplementedError()
 
     @cached_property
-    def cql_ip_address(self):
+    def cql_address(self):
         if self.test_config.IP_SSH_CONNECTIONS == 'public':
             return self.external_address
         with self.remote_scylla_yaml() as scylla_yaml:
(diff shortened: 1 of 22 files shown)


soyacz:
basic is working: https://argus.scylladb.com/workspace?state=WyJjMzhkMjA1MC0xZTEyLTQ3OWEtODk0Mi00MGNmYWFlNDQ4NTkiXQ
Next job I run with lower load and longer duration to try as many nemesis as possible.

sdcm/cluster.py
```diff
@@ -3939,7 +3950,10 @@ def set_seeds(self, wait_for_timeout=300, first_only=False):
 
     @property
     def seed_nodes_ips(self):
```
fruch, line 3952: maybe we need to rename this function ? since it doesn't return ips only anymore ?

sdcm/cluster.py
```diff
@@ -792,6 +793,14 @@ def public_ip_address(self) -> Optional[str]:
             self._public_ip_address_cached = self._get_public_ip_address()
         return self._public_ip_address_cached
 
+    @property
+    def public_dns_name(self) -> str:
+        return self.name
```
fruch, line 798: i'm not sure we'll want to default it like that...

not if we didn't tested it for GCP/Azure/docker

maybe raising not implemented is better for now ?

sdcm/sct_config.py
```diff
@@ -1473,6 +1473,9 @@ class SCTConfiguration(dict):
              help="""Forces GossipingPropertyFileSnitch (regardless `endpoint_snitch`) to simulate racks.
              Provide number of racks to simulate."""),
 
+        dict(name="use_dns_names", env="SCT_USE_DNS_NAMES", type=boolean,
```
fruch, line 1476: I would suggest adding some verification step, to make sure we don't enable it for backend that are not tested yet.
or that we won't support for now, like k8s one ?

soyacz:
longer duration failed due some manager nemesis issue.
It managed to execute 44 nemesis - none shown problems with dns name.
argus: https://argus.scylladb.com/test/c38d2050-1e12-479a-8942-40cfaae44859/runs?additionalRuns[]=7aac12d4-ef23-4927-9411-8a5eb232ab88

Only thing I found is to make also manager to communicate with dns name instead of ip.

sdcm/cluster.py
soyacz: yes, and I found this naming is really messed up all around.
I'll try to make it right.

soyacz:
@fgelcer which test do you think we could add  this`use_dns_names` setting to? Some tier1 one?

test-cases/longevity/longevity-mini-test-1h.yaml
```diff
@@ -1,5 +1,5 @@
-test_duration: 180
-stress_cmd: ["cassandra-stress write cl=QUORUM duration=90m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=50 -pop seq=1..10000000 -log interval=5"]
+test_duration: 1000
+stress_cmd: ["cassandra-stress write cl=QUORUM duration=900m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=2 -pop seq=1..10000000 -log interval=5"]
```
fgelcer, line 2: i guess you changed this just to debug purpose, so, once you are done, please, revert the changes in here

sdcm/cluster.py
```diff
@@ -250,6 +250,7 @@ def __init__(self, name, parent_cluster, ssh_login_info=None, base_logdir=None,
 
         self.remoter: Optional[RemoteCmdRunnerBase] = None
 
+        self._use_dns_names: bool = parent_cluster.params.get('use_dns_names') or False
```
fgelcer, line 253: we usually set it in one of the config files, either specific to backend, like `aws_config.yaml` or `test_default.yaml`, so setting it in here like this, will be much harder to understand the defaults, when someone will decide to change it

fgelcer (commented):
2 little comments

fgelcer:
> @fgelcer which test do you think we could add this`use_dns_names` setting to? Some tier1 one?

why not to run in all of them? i believe that, most of access in the field are made using DNS, instead of IP addresses... so we can keep few tests with private, and few tests with public IPs, but test mainly DNS... @fruch @roydahan , WDYT?

roydahan:
> > @fgelcer which test do you think we could add this`use_dns_names` setting to? Some tier1 one?
> 
> why not to run in all of them? i believe that, most of access in the field are made using DNS, instead of IP addresses... so we can keep few tests with private, and few tests with public IPs, but test mainly DNS... @fruch @roydahan , WDYT?

Because it's not the case.

fruch:
> > > @fgelcer which test do you think we could add this`use_dns_names` setting to? Some tier1 one?
> > 
> > why not to run in all of them? i believe that, most of access in the field are made using DNS, instead of IP addresses... so we can keep few tests with private, and few tests with public IPs, but test mainly DNS... @fruch @roydahan , WDYT?
> 
> Because it's not the case.

So randomly select one of the tier1 cases ? :)

Or we'll duplicate some test case and run it in releases only ?

fruch:
> > > > @fgelcer which test do you think we could add this`use_dns_names` setting to? Some tier1 one?
> > > 
> > > 
> > > why not to run in all of them? i believe that, most of access in the field are made using DNS, instead of IP addresses... so we can keep few tests with private, and few tests with public IPs, but test mainly DNS... @fruch @roydahan , WDYT?
> > 
> > 
> > Because it's not the case.
> 
> So randomly select one of the tier1 cases ? :)
> 
> Or we'll duplicate some test case and run it in releases only ?

@roydahan @fgelcer, we need an answer

roydahan:
Pick one of the tier1 cases.

soyacz:
I added it to 3day longevity

soyacz:
@fruch pylint passes - for it for simplicity I added default implementation of dns name - which will not work for some backends (for some will) - anyway we run it only in AWS and in case we need support for GCE/Azure we can implement proper methods.

unit tests fixed

sdcm/cluster_aws.py
```diff
@@ -513,6 +513,18 @@ def external_address(self):
         else:
             return self._instance.private_ip_address
 
+    @cached_property
```
fruch, line 516: I wonder if we need to uncache it on reboot of a node

fruch (approved):
LGTM.

One comment about caching

sdcm/cluster_aws.py
soyacz: in that case we should raise some issue - as it would break cluster network.

fruch: seems like we are using by default the IP name 
```
Hostname type
IP name: ip-10-4-0-64.eu-west-1.compute.internal
```
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-naming.html

there is one case where we stop and start instance, it's in `AWSNode.restart()` it's part of nemesis that is long disabled.
but we are calling there `refresh_ip_address`, that clears all ip related caches, since we've seen cases the public ip address is changing, I hope that local ip address never changes, cause our hostname are matching it.

I still think we should clear the cache in `refresh_ip_address` calls, and maybe generate error event if it was changing.

soyacz: regardless, dns name should never change

soyacz: in that case, it may happen. Added proper cache invalidation in `refresh_ip_address` (btw. fixing `cql_ip_address` invalidation along)

fruch (approved):
LGTM

fruch:
@soyacz do you have a short run of tier1 test it was added to ? 
just a quick proof it's not completely breaking it ?

and then I think we can merge this one

soyacz:
> @soyacz do you have a short run of tier1 test it was added to ? just a quick proof it's not completely breaking it ?
> 
> and then I think we can merge this one

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-50gb-3days-test/2/ - runs ok (got into nemesis part and run few of them). I aborted it after verification.

soyacz:
ready for merge I think

fruch (approved):
LGTM

PR #6551, opened 2023-08-29
fix(azure-vm): disable azure agents

fruch:
since those agents are enabling auditd, which in turn spams our logs with information we don't need

Fixes: https://github.com/scylladb/scylla-enterprise-machine-image/issues/63

### Testing

- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-azure-test/18/

### DIFF
--- a/sdcm/provision/azure/virtual_machine_provider.py
+++ b/sdcm/provision/azure/virtual_machine_provider.py
@@ -168,6 +168,7 @@ class VirtualMachineProvider:
     def _get_os_profile(computer_name: str, admin_username: str,
                         admin_password: str, ssh_public_key: str, custom_data: str) -> Dict[str, Any]:
         os_profile = {"os_profile": {
+            "allow_extension_operations": False,
             "computer_name": computer_name,
             "admin_username": admin_username,
             "admin_password": admin_password,
--- a/unit_tests/provisioner/fake_azure_service.py
+++ b/unit_tests/provisioner/fake_azure_service.py
@@ -505,7 +505,7 @@ class FakeVirtualMachines:
                         }
                     },
                     "secrets": [],
-                    "allowExtensionOperations": True,
+                    "allowExtensionOperations": False,
                     "requireGuestProvisionSignal": True
                 },
                 "networkProfile": {


fruch:
it's in draft until I'll get a change to test it's actually working

sdcm/provision/azure/virtual_machine_provider.py
```diff
@@ -168,6 +168,7 @@ def add_tags(self, name: str, tags: Dict[str, str]) -> VirtualMachine:
     def _get_os_profile(computer_name: str, admin_username: str,
                         admin_password: str, ssh_public_key: str, custom_data: str) -> Dict[str, Any]:
         os_profile = {"os_profile": {
+            "allowExtensionOperations": False,
```
soyacz, line 171: I'm not sure if it will work - other's are in snake_case: `allow_extension_operations`

fruch: as if Azure docs were that clear about it 😭

fruch: after digging a bit more, I've found:
https://learn.microsoft.com/en-us/python/api/azure-mgmt-compute/azure.mgmt.compute.v2021_03_01.models.osprofile?view=azure-python#parameters

fruch: fixed

fruch:
looks like this change is doing what's needed

soyacz:
I don't see it as valid solution - rather a workaround.
We don't control how users create VM's and some might need this setting set to true.
My other concern is that we break something else when disabling `allowExtensionOperations`

soyacz:
`sudo sed -i 's/^enabled.*/enabled = no/' /etc/audit/auditd.conf` during VM preparation wouldn't help?

fruch:
> `sudo sed -i 's/^enabled.*/enabled = no/' /etc/audit/auditd.conf` during VM preparation wouldn't help?

no, cause the install is happening by an agent, 
so we'll have to install auditd and do it, and assume the agent won't override our configuration.

fruch:
I don't want any such agent running uncontrolled during testing, installing things or configuring things I'm not aware of.
we have enough surprises as is.

Issue #6232, opened 2023-06-11
Events process EventsAnalyzer is still alive after timeout


fruch:
## Issue description

- [ ] This issue is a regression.
- [x] It is unknown if this issue is a regression.

during teardown `EventsAnalyzer` doesn't seem to stop correctly:
```
2023-06-08 10:01:27.662: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=f0b75a5e-07b5-409d-84c7-e991bb1591e5 during_nemesis=RemoveNodeThenAddNode, source=LongevityTest.stop_event_analyzer() message=stop_event_analyzer (silenced) failed with:
exception=Events process EventsAnalyzer is still alive after timeout
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/./sct.py", line 1729, in <module>
cli.main(prog_name="hydra")
File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
rv = self.invoke(ctx)
File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
return _process_result(sub_ctx.command.invoke(sub_ctx))
File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
return ctx.invoke(self.callback, **ctx.params)
File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
return __callback(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/./sct.py", line 1104, in run_test
unittest.main(module=None, argv=['python -m unittest', argv],
File "/usr/local/lib/python3.10/unittest/main.py", line 101, in __init__
self.runTests()
File "/usr/local/lib/python3.10/unittest/main.py", line 271, in runTests
self.result = testRunner.run(self.test)
File "/usr/local/lib/python3.10/unittest/runner.py", line 176, in run
test(result)
File "/usr/local/lib/python3.10/unittest/suite.py", line 84, in __call__
return self.run(*args, **kwds)
File "/usr/local/lib/python3.10/unittest/suite.py", line 122, in run
test(result)
File "/usr/local/lib/python3.10/unittest/suite.py", line 84, in __call__
return self.run(*args, **kwds)
File "/usr/local/lib/python3.10/unittest/suite.py", line 122, in run
test(result)
File "/usr/local/lib/python3.10/unittest/case.py", line 650, in __call__
return self.run(*args, **kwds)
[...]

PR #6559, opened 2023-08-31
fix(unit_test): set --experimental-features explicitly

tchaikov:
now that we've deprecated and later stopped handling the option "experimental". let's specify the "experimetan_features" explictly. this should address the failures in the upgrade tests which expect that the experimental features are enabled by "--experimental 1" command line option.

Fixes https://github.com/scylladb/scylladb/issues/15214

### DIFF
--- a/configurations/raft/enable_raft_experimental.yaml
+++ b/configurations/raft/enable_raft_experimental.yaml
@@ -1,3 +1,6 @@
 append_scylla_yaml: |
   experimental_features:
+    - alternator-streams
+    - alternator-ttl
+    - keyspace-storage-options
     - raft
--- a/configurations/raft/enable_raft_experimental.yaml
+++ b/configurations/raft/enable_raft_experimental.yaml
@@ -1,6 +1,3 @@
 append_scylla_yaml: |
   experimental_features:
-    - alternator-streams
-    - alternator-ttl
-    - keyspace-storage-options
     - raft
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -14,10 +14,6 @@ manager_scylla_backend_version: '2022'
 
 scylla_repo_loader: 'https://s3.amazonaws.com/downloads.scylladb.com/rpm/centos/scylla-4.6.repo'
 
-experimental_features:
-  - alternator-streams
-  - alternator-ttl
-  - keyspace-storage-options
 round_robin: false
 
 append_scylla_args: '--blocked-reactor-notify-ms 25 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1 --enable-sstable-key-validation 1'
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -14,7 +14,10 @@ manager_scylla_backend_version: '2022'
 
 scylla_repo_loader: 'https://s3.amazonaws.com/downloads.scylladb.com/rpm/centos/scylla-4.6.repo'
 
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
+  - keyspace-storage-options
 round_robin: false
 
 append_scylla_args: '--blocked-reactor-notify-ms 25 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1 --enable-sstable-key-validation 1'
--- a/internal_test_data/complex_test_case_with_version.yaml
+++ b/internal_test_data/complex_test_case_with_version.yaml
@@ -19,7 +19,10 @@ nemesis_interval: 5
 user_prefix: 'longevity-50gb-4d-not-jenkins'
 space_node_threshold: 644245094
 ip_ssh_connections: 'private'
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
+  - keyspace-storage-options
 instance_provision: 'on_demand'
 
 
--- a/internal_test_data/complex_test_case_with_version.yaml
+++ b/internal_test_data/complex_test_case_with_version.yaml
@@ -19,10 +19,6 @@ nemesis_interval: 5
 user_prefix: 'longevity-50gb-4d-not-jenkins'
 space_node_threshold: 644245094
 ip_ssh_connections: 'private'
-experimental_features:
-  - alternator-streams
-  - alternator-ttl
-  - keyspace-storage-options
 instance_provision: 'on_demand'
 
 
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -1689,8 +1689,8 @@ class KubernetesCluster(metaclass=abc.ABCMeta):  # pylint: disable=too-many-publ
         #       So, define some options here by default. It may be extended later if needed.
         with self.remote_scylla_yaml(namespace=namespace) as scylla_yml:
             # Process cluster params
-            if self.params.get("experimental"):
-                scylla_yml["experimental"] = self.params.get("experimental")
+            if self.params.get("experimental_features"):
+                scylla_yml["experimental_features"] = self.params.get("experimental_features")
             if self.params.get("hinted_handoff"):
                 scylla_yml["hinted_handoff_enabled"] = self.params.get(
                     "hinted_handoff").lower() in ("enabled", "true")
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -436,7 +436,6 @@ class FillDatabaseData(ClusterTester):
                 [[-4]]
             ],
             'min_version': '1.7',
-            'skip_condition': "'experimental_features' in self.params",
             'max_version': '',
             'skip': ''},
         {
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -436,7 +436,7 @@ class FillDatabaseData(ClusterTester):
                 [[-4]]
             ],
             'min_version': '1.7',
-            'skip_condition': "self.params.get('experimental')",
+            'skip_condition': "'experimental_features' in self.params",
             'max_version': '',
             'skip': ''},
         {
--- a/sdcm/provision/scylla_yaml/scylla_yaml.py
+++ b/sdcm/provision/scylla_yaml/scylla_yaml.py
@@ -268,7 +268,6 @@ class ScyllaYaml(BaseModel):  # pylint: disable=too-few-public-methods,too-many-
     developer_mode: bool = None  # False
     skip_wait_for_gossip_to_settle: int = None  # -1
     force_gossip_generation: int = None  # -1
-    experimental: bool = None  # False
     experimental_features: list[str] = None  # []
     lsa_reclamation_step: int = None  # 1
     prometheus_port: int = None  # 9180
--- a/test-cases/features/alternator-ttl/longevity-alternator-1h-scan-12h-ttl-no-lwt-2h-grace-4loaders-sisyphus.yaml
+++ b/test-cases/features/alternator-ttl/longevity-alternator-1h-scan-12h-ttl-no-lwt-2h-grace-4loaders-sisyphus.yaml
@@ -61,10 +61,11 @@ alternator_secret_access_key: 'password'
 # TTL mode is experimental in version 5.1.
 # Enable TTL feature in Scylla.
 # Set 'alternator_ttl_period_in_seconds' to 1 hour for the TTL scan interval.
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
+  - keyspace-storage-options
 append_scylla_yaml: |
-  experimental_features:
-    - alternator-ttl
   alternator_ttl_period_in_seconds: 3600
 
 # Set 'gc_grace_seconds' for 2 hours
--- a/test-cases/features/alternator-ttl/longevity-alternator-1h-scan-12h-ttl-no-lwt-2h-grace-4loaders-sisyphus.yaml
+++ b/test-cases/features/alternator-ttl/longevity-alternator-1h-scan-12h-ttl-no-lwt-2h-grace-4loaders-sisyphus.yaml
@@ -62,9 +62,7 @@ alternator_secret_access_key: 'password'
 # Enable TTL feature in Scylla.
 # Set 'alternator_ttl_period_in_seconds' to 1 hour for the TTL scan interval.
 experimental_features:
-  - alternator-streams
   - alternator-ttl
-  - keyspace-storage-options
 append_scylla_yaml: |
   alternator_ttl_period_in_seconds: 3600
 
--- a/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-36m-ttl-8m-grace-sisyphus-rewrite-expired-data.yaml
+++ b/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-36m-ttl-8m-grace-sisyphus-rewrite-expired-data.yaml
@@ -53,10 +53,11 @@ alternator_secret_access_key: 'password'
 # Enable TTL feature in Scylla.
 # Set 'alternator_ttl_period_in_seconds' to 1 hour for the TTL scan interval.
 # Set 'gc_grace_seconds' for 8 minutes.
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
+  - keyspace-storage-options
 append_scylla_yaml: |
-  experimental_features:
-    - alternator-ttl
   alternator_ttl_period_in_seconds: 240
 
 post_prepare_cql_cmds: "ALTER TABLE alternator_usertable_no_lwt.usertable_no_lwt with gc_grace_seconds = 480;"
--- a/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-36m-ttl-8m-grace-sisyphus-rewrite-expired-data.yaml
+++ b/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-36m-ttl-8m-grace-sisyphus-rewrite-expired-data.yaml
@@ -54,9 +54,7 @@ alternator_secret_access_key: 'password'
 # Set 'alternator_ttl_period_in_seconds' to 1 hour for the TTL scan interval.
 # Set 'gc_grace_seconds' for 8 minutes.
 experimental_features:
-  - alternator-streams
   - alternator-ttl
-  - keyspace-storage-options
 append_scylla_yaml: |
   alternator_ttl_period_in_seconds: 240
 
--- a/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-multiple-ttl-8m-grace.yaml
+++ b/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-multiple-ttl-8m-grace.yaml
@@ -102,9 +102,7 @@ alternator_secret_access_key: 'password'
 # TTL: 16 different ones. The longer is 10 days, the second longer is 2.5 hours
 # GC Grace: 8 minutes
 experimental_features:
-  - alternator-streams
   - alternator-ttl
-  - keyspace-storage-options
 append_scylla_yaml: |
   alternator_ttl_period_in_seconds: 240
 
--- a/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-multiple-ttl-8m-grace.yaml
+++ b/test-cases/features/alternator-ttl/longevity-alternator-4m-scan-multiple-ttl-8m-grace.yaml
@@ -101,10 +101,11 @@ alternator_secret_access_key: 'password'
 # Scan: 4 minutes
 # TTL: 16 different ones. The longer is 10 days, the second longer is 2.5 hours
 # GC Grace: 8 minutes
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
+  - keyspace-storage-options
 append_scylla_yaml: |
-  experimental_features:
-    - alternator-ttl
   alternator_ttl_period_in_seconds: 240
 
 post_prepare_cql_cmds: "ALTER TABLE alternator_usertable_no_lwt.usertable_no_lwt with gc_grace_seconds = 480;"
--- a/test-cases/features/alternator-ttl/longevity-alternator-disable-enable-ttl.yaml
+++ b/test-cases/features/alternator-ttl/longevity-alternator-disable-enable-ttl.yaml
@@ -53,9 +53,7 @@ alternator_secret_access_key: 'password'
 # (YCSB TTL parameter is set to 10 minutes)
 # Set 'gc_grace_seconds' for 8 minutes,
 experimental_features:
-  - alternator-streams
   - alternator-ttl
-  - keyspace-storage-options
 append_scylla_yaml: |
   alternator_ttl_period_in_seconds: 240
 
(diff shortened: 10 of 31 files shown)


fgelcer (changes requested):
LGTM, but it won't fix the issue, as in this patch you are fixing the unit test only, while the problem we are having is in the rolling upgrade test... some extra changes must be done to really fix the issue we have in the upgrade tests

tchaikov:
i see. i am on it.

tchaikov:
> There is something I don't understand. **WHY** do the upgrade tests enable any experimental features - let alone tried to enable all of them (with the now-removed "-experimental" option)? Isn't the whole idea that experimental features are **not** guaranteed to survive upgrades?
> 
> In other words I am asking - shouldn't we remove "--experimental 1" and not add **anything** to replace it?

that's good question. how about leaving it to another change? as i think this change is supposed to bring us back to the status before the "experimental" option got removed.

nyh:
> > There is something I don't understand. **WHY** do the upgrade tests enable any experimental features - let alone tried to enable all of them (with the now-removed "-experimental" option)? Isn't the whole idea that experimental features are **not** guaranteed to survive upgrades?
> > In other words I am asking - shouldn't we remove "--experimental 1" and not add **anything** to replace it?
> 
> that's good question. how about leaving it to another change? as i think this change is supposed to bring us back to the status before the "experimental" option got removed.

That's up to the developers of this test to decide. I am not sure anyone realized they were even using "--experimental" (which has been deprecated for 2 years) so never considered removing it. The fact we never had a problem with it was probably just luck - it's very possible that enabling an experimental feature on one version and then upgrading will break!

So I think "--experimental" was a mistake, and adding now a random list of experimental features to replace it makes little sense (which features do we list? when we invent new experimental features, are we supposed to update this list?). Personally I would just consider the use of "--experimental" a mistake, and remove it. But again, I want the developers of this test to make this decision.

fruch:
> > > There is something I don't understand. **WHY** do the upgrade tests enable any experimental features - let alone tried to enable all of them (with the now-removed "-experimental" option)? Isn't the whole idea that experimental features are **not** guaranteed to survive upgrades?
> > > In other words I am asking - shouldn't we remove "--experimental 1" and not add **anything** to replace it?
> > 
> > that's good question. how about leaving it to another change? as i think this change is supposed to bring us back to the status before the "experimental" option got removed.
> 
> That's up to the developers of this test to decide. I am not sure anyone realized they were even using "--experimental" (which has been deprecated for 2 years) so never considered removing it. The fact we never had a problem with it was probably just luck - it's very possible that enabling an experimental feature on one version and then upgrading will break!
> 
> So I think "--experimental" was a mistake, and adding now a random list of experimental features to replace it makes little sense (which features do we list? when we invent new experimental features, are we supposed to update this list?). Personally I would just consider the use of "--experimental" a mistake, and remove it. But again, I want the developers of this test to make this decision.

Well we never got the memo that the experimental flag is gone.

I was by default enabled, so we'll cover those all the time. yes in upgrade it could be highly problematic, we have tests specific for those for example for raft upgrades.

If experimental flag isn't working anymore, we should remove it completely first.

I'm o.k. with introducing experimental_feature to SCT configuration, but agree with @nyh, we should use it only for those the tests require, and not trying to add those back those untouched feature in unrelated test, and for sure not by default.

If there is an experimental feature that need attention and testing it's the developer of that feature responsibility to communicate that need and the testing requirements, we should try to automatically cover any experimental feature, not without intention.

nyh:
> Well we never got the memo that the experimental flag is gone.

:-)

> I was by default enabled, so we'll cover those all the time. yes in upgrade it could be highly problematic, we have tests specific for those for example for raft upgrades.

Raft was never (if I remember correctly) part of the big "--experimental" flag and always required a separate flag (that got renamed a few months ago).
It seems although we don't promise that upgrades will work with experimental features, in practice they usually do. So we never caught any failed upgrades even with the experimental flag. The question I have, though, is why do we want to test those kinds of upgrades that we specifically warn that will NOT work? I think we're in agreement that we shouldn't, so the "--experimental" flag can simply be removed, replaced by nothing.

> If experimental flag isn't working anymore, we should remove it completely first.
> 
> I'm o.k. with introducing experimental_feature to SCT configuration, but agree with @nyh, we should use it only for those the tests require, and not trying to add those back those untouched feature in unrelated test, and for sure not by default.

I agree - if we have an upgrade test that test some experimental specific feature (e.g., UDF or Alternator Streams, etc.) across upgrades, then we should enable that specific experimental feature (and watch out that this test may fail if the feature changes in a non-compatible way!). But if we don't, I think we shouldn't enable all of them just because they exist.

tchaikov:
```
[2023-08-31T11:48:42.032Z] An unexpected error has occurred: CalledProcessError: command: ('/usr/local/bin/python', '-mnodeenv', '--prebuilt', '--clean-src', '/home/jenkins/.cache/pre-commit/repoqczcwt3u/node_env-16.15.1', '-n', '16.15.1')
[2023-08-31T11:48:42.032Z] return code: -15
[2023-08-31T11:48:42.032Z] expected return code: 0
[2023-08-31T11:48:42.032Z] stdout: (none)
[2023-08-31T11:48:42.032Z] stderr:
[2023-08-31T11:48:42.032Z]     /usr/local/lib/python3.10/site-packages/nodeenv.py:48: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
[2023-08-31T11:48:42.032Z]       from pkg_resources import parse_version
[2023-08-31T11:48:42.032Z]      * Install prebuilt node (16.15.1) .Incomplete read while readingfrom https://nodejs.org/download/release/v16.15.1/node-v16.15.1-linux-x64.tar.gz - IncompleteRead(7880434 bytes read, 25153200 more expected)
[2023-08-31T11:48:42.032Z]     Incomplete read while readingfrom https://nodejs.org/download/release/v16.15.1/node-v16.15.1-linux-x64.tar.gz - IncompleteRead(19594994 bytes read, 13438640 more expected)
```

the failure is not related.

in the meanwhile. let me try to remove the "fat" from the tests.

unit_tests/conftest.py
```diff
@@ -77,8 +77,16 @@ def fixture_docker_scylla(request: pytest.FixtureRequest):
...
     ssl_dir = (Path(__file__).parent.parent / 'data_dir' / 'ssl_conf').absolute()
+    # experimental features previous enabled by '--experimental' umbrella flag, which
+    # was deprecated and removed, so let's specify them explictly. 'udf' should be
+    # enabled separately along with 'enable_user_defined_functions'
+    experimental_features = ['alternator-streams',
+                             'alternator-ttl',
+                             'keyspace-storage-options']
+    experimental_flags = ' '.join(f'--experimental-features {feature}'
+                                  for feature in experimental_features)
     scylla = RemoteDocker(LocalNode("scylla", cluster), image_name=docker_version,
```
fruch, line 88: let's not touch this part in less we have too, we don't need the experimental flags here at all

fruch: @nyh BTW we might want to remove `--experimental` from scylla docker as well...

configurations/raft/enable_raft_experimental.yaml
```diff
@@ -1,3 +1,6 @@
 append_scylla_yaml: |
   experimental_features:
+    - alternator-streams
```
fruch, line 3: remove those from here
this should only have raft (and even that might need to be changed)

defaults/test_default.yaml
```diff
@@ -14,7 +14,10 @@ manager_scylla_backend_version: '2022'
 
 scylla_repo_loader: 'https://s3.amazonaws.com/downloads.scylladb.com/rpm/centos/scylla-4.6.repo'
 
-experimental: true
+experimental_features:
```
fruch, line 17: default should be empty

internal_test_data/complex_test_case_with_version.yaml
```diff
@@ -19,7 +19,10 @@ nemesis_interval: 5
 user_prefix: 'longevity-50gb-4d-not-jenkins'
 space_node_threshold: 644245094
 ip_ssh_connections: 'private'
-experimental: true
+experimental_features:
```
fruch, line 22: unit test can be empty as well

sdcm/cluster_k8s/__init__.py
```diff
@@ -1689,8 +1689,8 @@ def init_scylla_config_map(self, namespace: str = SCYLLA_NAMESPACE, **kwargs):
         #       So, define some options here by default. It may be extended later if needed.
         with self.remote_scylla_yaml(namespace=namespace) as scylla_yml:
             # Process cluster params
-            if self.params.get("experimental"):
-                scylla_yml["experimental"] = self.params.get("experimental")
+            if self.params.get("experimental_features"):
+                scylla_yml["experimental_features"] = self.params.get("experimental_features")
```
fruch, line 1693: we might need to explain that this flag can be overridden by `append_scylla_yaml`, on `experimental_features` configuration doc

sdcm/provision/scylla_yaml/scylla_yaml.py
```diff
@@ -268,7 +268,6 @@ def set_authorizer(cls, authorizer: str):
     developer_mode: bool = None  # False
     skip_wait_for_gossip_to_settle: int = None  # -1
     force_gossip_generation: int = None  # -1
-    experimental: bool = None  # False
```
fruch, line 271: if it's removed from here, `experimental_feature` should be added instead

test-cases/features/alternator-ttl/longevity-alternator-1h-scan-12h-ttl-no-lwt-2h-grace-4loaders-sisyphus.yaml
```diff
@@ -61,10 +61,11 @@ alternator_secret_access_key: 'password'
 # TTL mode is experimental in version 5.1.
 # Enable TTL feature in Scylla.
 # Set 'alternator_ttl_period_in_seconds' to 1 hour for the TTL scan interval.
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
```
fruch, line 66: should only be `alternator-ttl` the rest are not relevent

test-cases/features/alternator-ttl/longevity-alternator-4m-scan-36m-ttl-8m-grace-sisyphus-rewrite-expired-data.yaml
```diff
@@ -53,10 +53,11 @@ alternator_secret_access_key: 'password'
 # Enable TTL feature in Scylla.
 # Set 'alternator_ttl_period_in_seconds' to 1 hour for the TTL scan interval.
 # Set 'gc_grace_seconds' for 8 minutes.
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
```
fruch, line 58: should only be alternator-ttl the rest are not relevant

unit_tests/conftest.py
tchaikov: actually, i think the foremost important thing is not to do the cleanup. but to fix the test failure -- that's why i jumped in and prioritize it.

test-cases/features/alternator-ttl/longevity-alternator-4m-scan-multiple-ttl-8m-grace.yaml
```diff
@@ -101,10 +101,11 @@ alternator_secret_access_key: 'password'
 # Scan: 4 minutes
 # TTL: 16 different ones. The longer is 10 days, the second longer is 2.5 hours
 # GC Grace: 8 minutes
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
```
fruch, line 106: should only be alternator-ttl the rest are not relevent

test-cases/features/alternator-ttl/longevity-alternator-disable-enable-ttl.yaml
```diff
@@ -52,10 +52,11 @@ alternator_secret_access_key: 'password'
 # Set 'alternator_ttl_period_in_seconds' to 4 minutes for the TTL scan interval.
 # (YCSB TTL parameter is set to 10 minutes)
 # Set 'gc_grace_seconds' for 8 minutes,
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
```
fruch, line 57: should only be alternator-ttl the rest are not relevant

test-cases/features/alternator-ttl/longevity-alternator-ttl-big-dataset.yaml
```diff
@@ -58,10 +58,11 @@ alternator_secret_access_key: 'password'
 # Enable TTL feature in Scylla.
 # Set 'alternator_ttl_period_in_seconds' to 1 hour for the TTL scan interval.
 # gc_grace_seconds: 2 hours.
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
```
fruch, line 63: should only be alternator-ttl the rest are not relevant

test-cases/features/alternator-ttl/longevity-alternator-ttl-large-writes.yaml
```diff
@@ -58,10 +58,11 @@ alternator_secret_access_key: 'password'
 # Scan interval: 10 minutes
 # TTL (YCSB parameter): 8 minutes.
 # gc_grace_seconds: 20 minutes
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
```
fruch, line 63: should only be alternator-ttl the rest are not relevant

test-cases/features/alternator-ttl/longevity-alternator-ttl-lwt.yaml
```diff
@@ -39,8 +39,9 @@ alternator_secret_access_key: 'password'
 # TTL mode is experimental in version 5.1.
 # Enable TTL feature in Scylla.
 # Set 'alternator_ttl_period_in_seconds' to 7 minutes for the TTL scan interval.
-experimental: true
+experimental_features:
+  - alternator-streams
+  - alternator-ttl
```
fruch, line 44: should only be alternator-ttl the rest are not relevant

test-cases/longevity/longevity-alternator-streams-100gb-12h.yaml
```diff
@@ -46,7 +46,10 @@ nemesis_interval: 5
 
 user_prefix: 'alternator-streams-4h'
 space_node_threshold: 64424
-experimental: true
+experimental_features:
+  - alternator-streams
```
fruch, line 50: should only be `alternator-streams` the rest are not relevant

test-cases/longevity/longevity-cdc-100gb-4h.yaml
```diff
@@ -25,4 +25,4 @@ space_node_threshold: 64424
 
 server_encrypt: true
 
-experimental: false
+experimental_features: []
```
fruch, line 28: should be remove, cause default should be empty

test-cases/scale/scale-cluster.yaml
```diff
@@ -14,7 +14,7 @@ instance_type_db: 'm5d.large'
 
 user_prefix: 'cluster-scale-test'
 ssh_transport: 'libssh2'
-experimental: false
+experimental_features: []
```
fruch, line 17: should be remove, cause default should be empty

fruch:
> ```
> [2023-08-31T11:48:42.032Z] An unexpected error has occurred: CalledProcessError: command: ('/usr/local/bin/python', '-mnodeenv', '--prebuilt', '--clean-src', '/home/jenkins/.cache/pre-commit/repoqczcwt3u/node_env-16.15.1', '-n', '16.15.1')
> [2023-08-31T11:48:42.032Z] return code: -15
> [2023-08-31T11:48:42.032Z] expected return code: 0
> [2023-08-31T11:48:42.032Z] stdout: (none)
> [2023-08-31T11:48:42.032Z] stderr:
> [2023-08-31T11:48:42.032Z]     /usr/local/lib/python3.10/site-packages/nodeenv.py:48: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
> [2023-08-31T11:48:42.032Z]       from pkg_resources import parse_version
> [2023-08-31T11:48:42.032Z]      * Install prebuilt node (16.15.1) .Incomplete read while readingfrom https://nodejs.org/download/release/v16.15.1/node-v16.15.1-linux-x64.tar.gz - IncompleteRead(7880434 bytes read, 25153200 more expected)
> [2023-08-31T11:48:42.032Z]     Incomplete read while readingfrom https://nodejs.org/download/release/v16.15.1/node-v16.15.1-linux-x64.tar.gz - IncompleteRead(19594994 bytes read, 13438640 more expected)
> ```
> 
> the failure is not related.
> 
> in the meanwhile. let me try to remove the "fat" from the tests.

I'm not sure which "fat" exactly you are referring.

regardless, the failure of CI is related to https://github.com/ekalinin/nodeenv/issues/324, is a long standing issue with nodeenv module, just rebuild the CI job once it happens.

I've also add the SCT integration and provision test to validate this change. (so every time you push, it would run them as part of CI)

fruch (changes requested):
let's just limit this to the removal of `experimental` flag first, we don't need it.
then we can consider adding `experimental_features`

tchaikov:
> let's just limit this to the removal of `experimental` flag first, we don't need it.
> then we can consider adding `experimental_features`

"removal" alone does not address the problem of https://github.com/scylladb/scylladb/issues/15214

please note, before i dropped `experimental` flag. a bunch of so called "random" experimental features were enabled by the "experimental = true" settings in sct. after it was dropped. these "random" features were disabled. that's why tests are failing.

so, i am trying to reenable these features.

but what we are arguing here is to

reviewers:

> remove this and that

author if this change

> i want to bring the test back to normal, and *then* worry about "this" and "that"

tchaikov:
> > ```
> > [2023-08-31T11:48:42.032Z] An unexpected error has occurred: CalledProcessError: command: ('/usr/local/bin/python', '-mnodeenv', '--prebuilt', '--clean-src', '/home/jenkins/.cache/pre-commit/repoqczcwt3u/node_env-16.15.1', '-n', '16.15.1')
> > [2023-08-31T11:48:42.032Z] return code: -15
> > [2023-08-31T11:48:42.032Z] expected return code: 0
> > [2023-08-31T11:48:42.032Z] stdout: (none)
> > [2023-08-31T11:48:42.032Z] stderr:
> > [2023-08-31T11:48:42.032Z]     /usr/local/lib/python3.10/site-packages/nodeenv.py:48: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
> > [2023-08-31T11:48:42.032Z]       from pkg_resources import parse_version
> > [2023-08-31T11:48:42.032Z]      * Install prebuilt node (16.15.1) .Incomplete read while readingfrom https://nodejs.org/download/release/v16.15.1/node-v16.15.1-linux-x64.tar.gz - IncompleteRead(7880434 bytes read, 25153200 more expected)
> > [2023-08-31T11:48:42.032Z]     Incomplete read while readingfrom https://nodejs.org/download/release/v16.15.1/node-v16.15.1-linux-x64.tar.gz - IncompleteRead(19594994 bytes read, 13438640 more expected)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > the failure is not related.
> > in the meanwhile. let me try to remove the "fat" from the tests.
> 
> 
> I'm not sure which "fat" exactly you are referring.
> 

all the comments above.

fruch:
> > let's just limit this to the removal of `experimental` flag first, we don't need it.
> > then we can consider adding `experimental_features`
> 
> "removal" alone does not address the problem of [scylladb/scylladb#15214](https://github.com/scylladb/scylladb/issues/15214)

yes, it should.
if don't enable `experimental` on the base versions, there won't be a problem across all upgrade tests.

> 
> please note, before i dropped `experimental` flag. a bunch of so called "random" experimental features were enabled by the "experimental = true" settings in sct. after it was dropped. these "random" features were disabled. that's why tests are failing.
> 
> so, i am trying to reenable these features.

no need to reenable anything on SCT side...

> 
> but what we are arguing here is:
> 
> reviewers:
> 
> > remove this and that
> 
> author of this change
> 
> > i really want to bring the test back to normal, and _then_ worry about "this" and "that"

you'll have to trust us... :)

sdcm/provision/scylla_yaml/scylla_yaml.py
tchaikov: see the next line. it's already there.

tchaikov:
the test failure of

```
17:43:13  FAILED unit_tests/test_sct_events_file_logger.py::TestFileLogger::test_get_events_grouped_by_category_limit - AssertionError: Events process EventsFileLogger is still alive after timeout
```

it's a known issue, see #6232

tchaikov:
v2:

- add a commit to remove unnecessary `experimental-feature`s.

sdcm/fill_db_data.py
```diff
@@ -436,7 +436,7 @@ class FillDatabaseData(ClusterTester):
                 [[-4]]
             ],
             'min_version': '1.7',
-            'skip_condition': "self.params.get('experimental')",
+            'skip_condition': "'experimental_features' in self.params",
```
fruch, line 439: I believe we can remove this logic althoughter

fruch (dismissed):
LGTM

But I would like to see it pass the provisioning test at least once

sdcm/fill_db_data.py
tchaikov: dropped.

tchaikov:
v3:

- do not set `skip_condition` based on `experimental_features`.

tchaikov:
> > the test failure of
> > ```
> > 17:43:13  FAILED unit_tests/test_sct_events_file_logger.py::TestFileLogger::test_get_events_grouped_by_category_limit - AssertionError: Events process EventsFileLogger is still alive after timeout
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > it's a known issue, see #6232
> 
> This might be fixed by my fix for #6232, but looks `timeout=1` is anyway too short as pooler always wait's for 1s - so could be flaky. You can make `timeout=2` in `self.file_logger.stop(timeout=1)` in `tearDown` method of `TestFileLogger` class.

Thank you, Lukasz. this is not in the scope of this change. if this change persist, i will try to create a dedicated change to address it. but i don't really understand that part, so i am not qualified to create a change for it. if possible, could you include the proposed change in your PR ?

fruch:
> > > the test failure of
> > > ```
> > > 17:43:13  FAILED unit_tests/test_sct_events_file_logger.py::TestFileLogger::test_get_events_grouped_by_category_limit - AssertionError: Events process EventsFileLogger is still alive after timeout
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > it's a known issue, see #6232
> > 
> > This might be fixed by my fix for #6232, but looks `timeout=1` is anyway too short as pooler always wait's for 1s - so could be flaky. You can make `timeout=2` in `self.file_logger.stop(timeout=1)` in `tearDown` method of `TestFileLogger` class.
> 
> Thank you, Lukasz. this is not in the scope of this change. if this test failure persists, i can try to create a dedicated change to address it. but i don't really understand that part, so i am not qualified to create a change for it. if possible, could you include the proposed change in your PR ?

@tchaikov that unittest is flaky for a while , leave it for us to fix.

fruch (approved):
LGTM

fgelcer:
@scylladb/qa-maintainers please backport this one to `branch-perf-v14` as it is failing part of the tests

fgelcer:
[failure example](https://jenkins.scylladb.com/job/scylla-master/job/scylla-master-perf-regression-latency-650gb-during-upgrade/6/)

fruch:
> @scylladb/qa-maintainers please backport this one to `branch-perf-v14` as it is failing part of the tests

backported

PR #6560, opened 2023-08-31
improvement(template-longevity-5gb-1h): Add MV and large partition load

ShlomiBalalis:

### DIFF
--- a/sdcm/utils/nemesis.py
+++ b/sdcm/utils/nemesis.py
@@ -70,7 +70,7 @@ class NemesisJobGenerator:
             nemesis_config_body = Template(self.nemesis_config_file_template).render(
                 {"nemesis_class": cls})
             with (self.nemesis_test_config_dir / new_config_name).open(mode="w") as file:
-                file.write(nemesis_config_body)
+                file.write(nemesis_config_body + '\n')
             LOGGER.info("Created test config file: %s", new_config_name)
 
     def create_job_files_from_template(self) -> list[Path]:
@@ -86,7 +86,7 @@ class NemesisJobGenerator:
                 })
             job_path = self.base_nemesis_job_dir / job_file_name
             with job_path.open("w") as file:
-                file.write(nemesis_job_groovy_source)
+                file.write(nemesis_job_groovy_source + "\n")
             LOGGER.info("Created pipeline file: %s", job_file_name)
 
             pipeline_files.append(job_path)
--- a/templates/test_config/template-longevity-5gb-1h.yaml.j2
+++ b/templates/test_config/template-longevity-5gb-1h.yaml.j2
@@ -1,7 +1,7 @@
 test_duration: 90
 
 prepare_write_cmd:  ["cassandra-stress write cl=QUORUM n=5048570 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=80 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5",
-                     "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=1000 -clustering-row-count=555 -clustering-row-size=uniform:10..2048 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=30s -validate-data"
+                     "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=10000 -clustering-row-count=555 -clustering-row-size=uniform:128..2048 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=30s -validate-data"
                      ]
 
 post_prepare_cql_cmds: ['CREATE MATERIALIZED VIEW keyspace1.sort_by_C1 AS SELECT key, "C1" FROM keyspace1.standard1 WHERE "C1" IS NOT NULL and key IS NOT NULL PRIMARY KEY ("C1", key)'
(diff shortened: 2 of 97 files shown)


templates/test_config/template-longevity-5gb-1h.yaml.j2
```diff
@@ -1,9 +1,18 @@
...
-             "cassandra-stress read  cl=QUORUM duration=60m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=40 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5"
+             "cassandra-stress read  cl=QUORUM duration=60m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=40 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5",
+             "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=25 -clustering-row-count=5555 -clustering-row-size=uniform:1024..2048 -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 0 -duration=60m -validate-data"
```
roydahan, line 15:
```suggestion
             "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1000 -clustering-row-count=555 -clustering-row-size=uniform:10..2048 -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 0 -duration=60m -validate-data"
```

roydahan (changes requested):
I suggested few changes in number of partitions, amount of rows and size of rows.
I'm not sure about the concurrency and connections count you have there.
Anyway, need few iterations of testing.

Run 1 with my suggested changes and see how it looks in terms of load.
Then, run it for several different nemesis (use the ones we set for the nightly runs).

The hardest part would be to apply it for all the nemesis jobs, I'm afraid it won't be possible using the job builder since we've done some changes to the yamls of some jobs.

ShlomiBalalis:
> The hardest part would be to apply it for all the nemesis jobs, I'm afraid it won't be possible using the job builder since we've done some changes to the yamls of some jobs.

Yeah, there are several manual changes. This will be fun!

ShlomiBalalis:
Inital run passed: https://jenkins.scylladb.com/job/scylla-staging/job/Shlomo/job/nemesis/job/longevity-5gb-1h-CorruptThenRepairMonkey-aws-test_wih-large-partition/6/
I'll try a few different scenarios

fgelcer (commented):
only thing we have discussed, and is not yet in here, is the latest changes only "adding" the diff, instead of always starting from an brand new file from template...
like in few cases, that changes were made to different parameters, we would not need to care about them, in this case...

from a little local test i did, here is how we would add the diff only:
1. read config file (if exists) and loads into a Python dict
2. generates a new config file into a Python dict
3. runs a diff between both dicts, as: `original.update(dict(set(new.items() - original.items())))`
4. save back the original config files, including the diff changes into the file

everything else, seem fine, specially seen multiple tests did have good workload, and finished with success

fgelcer:
@roydahan , this change will affect the nightly jobs

sdcm/utils/nemesis.py
```diff
@@ -70,7 +70,7 @@ def create_test_cases_from_template(self):
             nemesis_config_body = Template(self.nemesis_config_file_template).render(
                 {"nemesis_class": cls})
             with (self.nemesis_test_config_dir / new_config_name).open(mode="w") as file:
-                file.write(nemesis_config_body)
+                file.write(nemesis_config_body + '\n')
```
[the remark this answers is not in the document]
ShlomiBalalis: I honestly think my solution is more readable. @roydahan Can you be the tie breaker?

ShlomiBalalis:
@roydahan The row size change has worked

fgelcer:
@roydahan , if this is done, i suggest getting it merged, so we can run on the official master jobs, and if there will be follow up fixes required, they should be sent as separated PRs...
so, please, review, and merge

fgelcer (approved):
LGTM

templates/test_config/template-longevity-5gb-1h.yaml.j2
```diff
@@ -1,9 +1,15 @@
 test_duration: 90
 
-prepare_write_cmd:  "cassandra-stress write cl=QUORUM n=5048570 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=80 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5"
+prepare_write_cmd:  ["cassandra-stress write cl=QUORUM n=5048570 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=80 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5",
```
roydahan, line 3: Remove the compaction strategy.

ShlomiBalalis: Done

fgelcer (changes requested):
pre-commit checks failed on YAML files verification... multiple of them... looks like you remove closing parenthesis/brackets

please, check and fix (according to the pre-check)

ShlomiBalalis:
> pre-commit checks failed on YAML files verification... multiple of them... looks like you remove closing parenthesis/brackets
> 
> please, check and fix (according to the pre-check)

It's funny, I didn't get any error when rebasing before, but did now when amending. I think our checks are not running during interactive rebase.
Anyway, should be fixed now.

fgelcer (approved):
LGTM

roydahan (approved):

roydahan:
Merging and hoping for the best in the nightly runs...

fgelcer:
i don't htink we need to backport this change to any other branches, but @roydahan WDYT?

roydahan:
We probably should, but let's see it's stable first.
The individual nemesis are already in 2023.1.

PR #6561, opened 2023-08-31
fix(event_process): fix stop events process under stream of events

soyacz:
When there's a lot of events coming to events process (e.g. `events_analyzer`) it doesn't verify state of `stop_event`. This makes it to last longer than stop timeout and failing the test usually.

Fix is about verifying state of `stop_event` after each event in `inbound_events` generator - not only when it's silent for 1 second.

fixes: #6232

### DIFF
--- a/sdcm/sct_events/events_device.py
+++ b/sdcm/sct_events/events_device.py
@@ -137,7 +137,7 @@ class EventsDevice(multiprocessing.Process):
     def inbound_events(self, stop_event: StopEvent) -> Generator[Any, None, None]:
         with zmq.Context() as ctx, self._sub_socket(ctx) as sub:
             while not stop_event.is_set():
-                while sub.poll(timeout=self.sub_polling_timeout):
+                if sub.poll(timeout=self.sub_polling_timeout):
                     yield sub.recv_pyobj(flags=zmq.NOBLOCK)
 
     # pylint: disable=import-outside-toplevel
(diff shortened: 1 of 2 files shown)


sdcm/sct_events/events_device.py
```diff
@@ -137,7 +137,7 @@ def _sub_socket(self, ctx: zmq.Context) -> zmq.Socket:
     def inbound_events(self, stop_event: StopEvent) -> Generator[Any, None, None]:
         with zmq.Context() as ctx, self._sub_socket(ctx) as sub:
             while not stop_event.is_set():
-                while sub.poll(timeout=self.sub_polling_timeout):
+                if sub.poll(timeout=self.sub_polling_timeout):
```
fruch, line 140: Oh my...

reminds me all those times I've written `def` instead of `class`

fruch (approved):
That's a great catch...

fruch:
Can we run one manager job to confirm?

soyacz:
> Can we run one manager job to confirm?

https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/centos-sanity-ipv6-test/1/

Also I've added an unit test that reproduces the issue. Btw. need to check why unittests fail...

soyacz:
> > Can we run one manager job to confirm?
> 
> https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/centos-sanity-ipv6-test/1/
> 
> Also I've added an unit test that reproduces the issue. Btw. need to check why unittests fail...

All passing

fruch (approved):
LGTM

Issue #6562, opened 2023-09-03
scylla_shards property shouldn't return strings


fruch:
seems like in some case (i.e. asymmetric case), `scylla_shards` return a string:
```
2023-08-30 21:23:58.986: (DisruptionEvent Severity.ERROR) period_type=end event_id=c3830a08-9661-49e4-9527-764d9d4250f4 duration=16s: nemesis_name=NodetoolRefresh target_node=Node longevity-50gb-12h-master-db-node-2a3b5f30-3 [54.87.128.100 | 10.12.11.62] (seed: True) errors='>' not supported between instances of 'str' and 'int'
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5011, in wrapper
result = method(*args[1:], **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1605, in disrupt_nodetool_refresh
if shards_num > 1:
TypeError: '>' not supported between instances of 'str' and 'int'
```

this is the fallbacks in the implementation: 
```
return self.smp or self.cpuset or self.cpu_cores
```

`smp` is int, `cpu_cores` is int, cpuset is string

fruch:
this seems to go back to when asymmetric case was introduced:
https://github.com/scylladb/scylla-cluster-tests/commit/97230389e58f8202df98f6aed1f959d8a716d5b3

fruch:
the argus link to the run I've run into it:
[https://argus.scylladb.com/test/6920a9a5-5864-4760-bf1b-97db557fa556/runs?additionalRuns[]=2a3b5f30-0494-4456-b4a0-b26a3a9617a5](https://argus.scylladb.com/test/6920a9a5-5864-4760-bf1b-97db557fa556/runs?additionalRuns%5B%5D=2a3b5f30-0494-4456-b4a0-b26a3a9617a5)

juliayakovlev:
@fruch 

`cpuset` is also `int` if it found

```
        core_counter = 0
        for range_or_singular in scylla_cpu_set:
            # Output of finding is of the following structure:
            #   [('1', '7', ''), ('', '', '15')]
            # First element is caught range, second element is caught singular
            core_counter += 1
            if not range_or_singular[2]:
                core_counter += int(range_or_singular[1]) - int(range_or_singular[0])
        return core_counter or ''
```

[...]

PR #6564, opened 2023-09-03
fix(db node): return scylla_shard as int always

juliayakovlev:
`scylla_shards` may return string instead of integer value. It cause to failure with error like:
```
TypeError: '>' not supported between instances of 'str' and 'int'
```

This property may receive the shards value from 3 different properties: `cpu_cores`, `smp` or `cpuset`. Every this property is used by different functions. So to decrease the impact of this fix, the shards value will be converted inside `scylla_shards` property, not in every property.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/6562


fruch (approved):
LGTM

soyacz (approved):
LGTM

PR #6577, opened 2023-09-05
fix(nemesis): Add new nemesis for topology changes feature

aleksbykov:
New nemesis for raft feature consistent-topology-changes:
ban removed node if it try back to cluster
The new nemesis send signal to scylla process and set it PAUSED
After the node with paused scylla marked as DN, the node is removed
After that signal SIGCONT is send and scylla is backed. but any connection
from that node should be ignored.

### DIFF
--- a/data_dir/nemesis.yml
+++ b/data_dir/nemesis.yml
@@ -234,6 +234,14 @@
   - topology_changes = True
 - disrupt_rebuild_streaming_err:
   - disruptive = True
+- disrupt_refuse_connection_with_block_scylla_ports_on_banned_node:
+  - disruptive = True
+  - topology_changes = True
+  - kubernetes = False
+- disrupt_refuse_connection_with_send_sigstop_signal_to_scylla_on_banned_node:
+  - disruptive = True
+  - topology_changes = True
+  - kubernetes = False
 - disrupt_remove_node_then_add_node:
   - disruptive = True
   - kubernetes = False
--- a/sdcm/exceptions.py
+++ b/sdcm/exceptions.py
@@ -91,3 +91,7 @@ class RaftTopologyCoordinatorNotFound(Exception):
 
 class NemesisStressFailure(Exception):
     """Exception to be raised to stop Nemesis flow, if stress command failed"""
+
+
+class BannedQueryExecUnexpectedSuccess(Exception):
+    """Exception when query executed successfully on banned node"""
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -38,8 +38,9 @@ from concurrent.futures import ThreadPoolExecutor
 from threading import Lock
 from types import MethodType  # pylint: disable=no-name-in-module
 
-from cassandra import ConsistencyLevel, InvalidRequest
+from cassandra import ConsistencyLevel, InvalidRequest, Unavailable
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
+from cassandra.cluster import NoHostAvailable, OperationTimedOut  # pylint: disable=no-name-in-module
 from invoke import UnexpectedExit
 from elasticsearch.exceptions import ConnectionTimeout as ElasticSearchConnectionTimeout
 from argus.common.enums import NemesisStatus
@@ -127,6 +128,7 @@ from sdcm.utils.loader_utils import DEFAULT_USER, DEFAULT_USER_PASSWORD, SERVICE
 from sdcm.utils.nemesis_utils.indexes import get_random_column_name, create_index, \
     wait_for_index_to_be_built, verify_query_by_index_works, drop_index, get_column_names, \
     wait_for_view_to_be_built, drop_materialized_view, is_cf_a_view
+from sdcm.utils.nemesis_utils import node_operations
 from sdcm.utils.node import build_node_api_command
 from sdcm.utils.replication_strategy_utils import temporary_replication_strategy_setter, \
     NetworkTopologyReplicationStrategy, ReplicationStrategy, SimpleReplicationStrategy
@@ -157,12 +159,14 @@ from sdcm.exceptions import (
     BootstrapStreamErrorFailure,
     QuotaConfigurationFailure,
     NemesisStressFailure,
+    BannedQueryExecUnexpectedSuccess,
 )
 from test_lib.compaction import CompactionStrategy, get_compaction_strategy, get_compaction_random_additional_params, \
     get_gc_mode, GcMode, calculate_allowed_twcs_ttl, get_table_compaction_info
 from test_lib.cql_types import CQLTypeBuilder
 from test_lib.sla import ServiceLevel, MAX_ALLOWED_SERVICE_LEVELS
 from sdcm.utils.topology_ops import FailedDecommissionOperationMonitoring
+
 
 LOGGER = logging.getLogger(__name__)
 # NOTE: following lock is needed in the K8S multitenant case
@@ -3720,9 +3724,12 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
         node_to_remove = self.target_node
         up_normal_nodes = self.cluster.get_nodes_up_and_normal(verification_node=node_to_remove)
-        # node_to_remove must be different than node
-        verification_node = random.choice([n for n in self.cluster.nodes if n is not node_to_remove])
 
+        with Nemesis.run_nemesis(node_list=up_normal_nodes, nemesis_label="RemoveNodeAddNode") as verification_node:
+            self._remove_node_add_node(verification_node=verification_node, node_to_remove=node_to_remove)
+
+    def _remove_node_add_node(self, verification_node, node_to_remove, remove_node_host_id=None):
+        # node_to_remove must be different than node
         # node_to_remove is single/last seed in cluster, before
         # it will be terminated, choose new seed node
         num_of_seed_nodes = len(self.cluster.seed_nodes)
@@ -3732,14 +3739,17 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.cluster.update_seed_provider()
 
         # get node's host_id
-        removed_node_status = self.cluster.get_node_status_dictionary(
-            ip_address=node_to_remove.ip_address, verification_node=verification_node)
-        assert removed_node_status is not None, "failed to get host_id using nodetool status"
-        host_id = removed_node_status["host_id"]
+        if not remove_node_host_id:
+            removed_node_status = self.cluster.get_node_status_dictionary(
+                ip_address=node_to_remove.ip_address, verification_node=verification_node)
+            assert removed_node_status is not None, "failed to get host_id using nodetool status"
+            host_id = removed_node_status["host_id"]
+        else:
+            host_id = remove_node_host_id
 
         with ignore_ycsb_connection_refused():
             # node stop and make sure its "DN"
-            node_to_remove.stop_scylla_server(verify_up=True, verify_down=True)
+            node_to_remove.stop_scylla_server(verify_up=False, verify_down=True)
 
             # terminate node
             self._terminate_cluster_node(node_to_remove)
@@ -3747,20 +3757,19 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         @retrying(n=3, sleep_time=5, message="Removing node from cluster...")
         def remove_node():
             removenode_reject_msg = r"Rejected removenode operation.*the node being removed is alive"
-            # nodetool removenode 'host_id'
-            rnd_node = random.choice([n for n in self.cluster.nodes if n is not self.target_node])
-            self.log.info("Running removenode command on {}, Removing node with the following host_id: {}"
-                          .format(rnd_node.ip_address, host_id))
-            with adaptive_timeout(Operations.REMOVE_NODE, rnd_node, timeout=HOUR_IN_SEC * 48):
-                res = rnd_node.run_nodetool("removenode {}".format(
-                    host_id), ignore_status=True, verbose=True, long_running=True, retry=0)
-            if res.failed and re.match(removenode_reject_msg, res.stdout + res.stderr):
-                raise Exception(f"Removenode was rejected {res.stdout}\n{res.stderr}")
+            with Nemesis.run_nemesis(node_list=self.cluster.nodes, nemesis_label="RemoveNodeAddNode") as rnd_node:
+                self.log.info("Running removenode command on {}, Removing node with the following host_id: {}"
+                              .format(rnd_node.ip_address, host_id))
+                with adaptive_timeout(Operations.REMOVE_NODE, rnd_node, timeout=HOUR_IN_SEC * 48):
+                    res = rnd_node.run_nodetool("removenode {}".format(
+                        host_id), ignore_status=True, verbose=True, long_running=True, retry=0)
+                if res.failed and re.match(removenode_reject_msg, res.stdout + res.stderr):
+                    raise Exception(f"Removenode was rejected {res.stdout}\n{res.stderr}")
 
             return res.exit_status
 
         # full cluster repair
-        up_normal_nodes.remove(node_to_remove)
+        up_normal_nodes = self.cluster.get_nodes_up_and_normal(verification_node)
         # Repairing will result in a best effort repair due to the terminated node,
         # and as a result requires ignoring repair errors
         with DbEventsFilter(db_event=DatabaseLogEvent.RUNTIME_ERROR,
@@ -3772,56 +3781,52 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                     self.log.error(f"failed to execute repair command "
                                    f"on node {node} due to the following error: {str(details)}")
 
-            # WORKAROUND: adding here the continuation of the nemesis to avoid the late filter messages above failing
-            # the entire nemesis.
-
-            exit_status = remove_node()
-
-            # if remove node command failed by any reason,
-            # we will remove the terminated node from
-            # dead_nodes_list, so the health validator terminate the job
-            if exit_status != 0:
-                self.log.error(f"nodetool removenode command exited with status {exit_status}")
-                # check difference between group0 and token ring,
-                garbage_host_ids = verification_node.raft.get_diff_group0_token_ring_members()
-                self.log.debug("Difference between token ring and group0 is %s", garbage_host_ids)
-                if garbage_host_ids:
-                    # if difference found, clean garbage and continue
-                    verification_node.raft.clean_group0_garbage()
+        exit_status = remove_node()
+        # if remove node command failed by any reason,
+        # we will remove the terminated node from
+        # dead_nodes_list, so the health validator terminate the job
+        if exit_status != 0:
+            self.log.error(f"nodetool removenode command exited with status {exit_status}")
+            # check difference between group0 and token ring,
+            garbage_host_ids = verification_node.raft.get_diff_group0_token_ring_members()
+            self.log.debug("Difference between token ring and group0 is %s", garbage_host_ids)
+            if garbage_host_ids:
+                # if difference found, clean garbage and continue
+                verification_node.raft.clean_group0_garbage()
+            else:
+                # group0 and token ring are consistent. Removenode failed by meanigfull reason.
+                # remove node from dead_nodes list to raise critical issue by HealthValidator
+                self.log.debug(
+                    f"Remove failed node {node_to_remove} from dead node list {self.cluster.dead_nodes_list}")
+                node = next((n for n in self.cluster.dead_nodes_list if n.ip_address ==
+                            node_to_remove.ip_address), None)
+                if node:
+                    self.cluster.dead_nodes_list.remove(node)
                 else:
-                    # group0 and token ring are consistent. Removenode failed by meanigfull reason.
-                    # remove node from dead_nodes list to raise critical issue by HealthValidator
-                    self.log.debug(
-                        f"Remove failed node {node_to_remove} from dead node list {self.cluster.dead_nodes_list}")
-                    node = next((n for n in self.cluster.dead_nodes_list if n.ip_address ==
-                                node_to_remove.ip_address), None)
-                    if node:
-                        self.cluster.dead_nodes_list.remove(node)
-                    else:
-                        self.log.debug(f"Node {node.name} with ip {node.ip_address} was not found in dead_nodes_list")
+                    self.log.debug(f"Node {node.name} with ip {node.ip_address} was not found in dead_nodes_list")
 
-            # verify node is removed by nodetool status
-            removed_node_status = self.cluster.get_node_status_dictionary(
-                ip_address=node_to_remove.ip_address, verification_node=verification_node)
-            assert removed_node_status is None, \
-                "Node was not removed properly (Node status:{})".format(removed_node_status)
+        # verify node is removed by nodetool status
+        removed_node_status = self.cluster.get_node_status_dictionary(
+            ip_address=node_to_remove.ip_address, verification_node=verification_node)
+        assert removed_node_status is None, \
+            "Node was not removed properly (Node status:{})".format(removed_node_status)
 
-            # add new node with same type (data node / zero token node)
-            new_node_args = {"count": 1, "rack": self.target_node.rack}
-            if self.target_node._is_zero_token_node:
-                new_node_args.update({"is_zero_node": True})
-            new_node = self._add_and_init_new_cluster_nodes(**new_node_args)[0]
-            # in case the removed node was not last seed.
-            if node_to_remove.is_seed and num_of_seed_nodes > 1:
-                new_node.set_seed_flag(True)
-                self.cluster.update_seed_provider()
-            # after add_node, the left nodes have data that isn't part of their tokens anymore.
-            # In order to eliminate cases that we miss a "data loss" bug because of it, we cleanup this data.
-            # This fix important when just user profile is run in the test and "keyspace1" doesn't exist.
-            try:
-                self.nodetool_cleanup_on_all_nodes_parallel()
-            finally:
-                self.unset_current_running_nemesis(new_node)
+        # add new node with same type (data node / zero token node)
+        new_node_args = {"count": 1, "rack": self.target_node.rack}
+        if self.target_node._is_zero_token_node:
+            new_node_args.update({"is_zero_node": True})
+        new_node = self._add_and_init_new_cluster_nodes(**new_node_args)[0]
+        # in case the removed node was not last seed.
+        if node_to_remove.is_seed and num_of_seed_nodes > 1:
+            new_node.set_seed_flag(True)
+            self.cluster.update_seed_provider()
+        # after add_node, the left nodes have data that isn't part of their tokens anymore.
+        # In order to eliminate cases that we miss a "data loss" bug because of it, we cleanup this data.
+        # This fix important when just user profile is run in the test and "keyspace1" doesn't exist.
+        try:
+            self.nodetool_cleanup_on_all_nodes_parallel()
+        finally:
+            self.unset_current_running_nemesis(new_node)
 
     @target_all_nodes
     def disrupt_network_reject_inter_node_communication(self):
@@ -5435,6 +5440,88 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.target_node.start_scylla()
             assert self.target_node != new_coordinator_node, \
                 f"New coordinator node was not elected while old one {coordinator_node.name} was stopped"
+
+    @target_all_nodes
+    def disrupt_refuse_connection_with_block_scylla_ports_on_banned_node(self):
+        self._refuse_connection_from_banned_node(use_iptables=True)
+
+    @target_all_nodes
+    def disrupt_refuse_connection_with_send_sigstop_signal_to_scylla_on_banned_node(self):
+        self._refuse_connection_from_banned_node(use_iptables=False)
+
+    def _refuse_connection_from_banned_node(self, use_iptables=False):
+        """Banned node could not connect with rest nodes in cluster
+
+        If node was removed from cluster for any reason, even if removed node
+        become alive and try to communicate with rest node in cluster, all connections
+        from it should be refused by other nodes in cluster
+        1. on target node block any scylladb process
+        1.1 Pause process
+        1.2 Block with iptables port 9100/10000
+        2. Wait and remove target node from cluster
+        3. start scylla on target node
+        4. Create exclusive connection to target node
+        5. Execute cql command on target node and validate that no operation
+        from target node passed to cluster
+        """
+        if not self.target_node.raft.is_consistent_topology_changes_enabled:
+            raise UnsupportedNemesis("Raft feature: consistent-topology-changes is not enabled")
+        if self._is_it_on_kubernetes():
+            raise UnsupportedNemesis("Skip test for K8S because no supported yet")
+        keyspace_name = "banned_keyspace"
+        table_name = "table1"
+
+        def drop_keyspace(node):
+            with self.cluster.cql_connection_patient(node=node) as session:
+                LOGGER.debug("Drop keyspace %s", keyspace_name)
+                session.execute(f"DROP KEYSPACE IF EXISTS {keyspace_name}", timeout=300)
+
+        simulate_node_unavailability = node_operations.block_scylla_ports if use_iptables else node_operations.pause_scylla_with_sigstop
+        with self.run_nemesis(node_list=self.cluster.nodes,
+                              nemesis_label=f"Running {simulate_node_unavailability.__name__}") as working_node, ExitStack() as stack:
+            target_host_id = self.target_node.host_id
+            stack.callback(self._remove_node_add_node, verification_node=working_node, node_to_remove=self.target_node,
+                           remove_node_host_id=target_host_id)
+
+            self.tester.create_keyspace(keyspace_name, replication_factor=3)
+            self.tester.create_table(name=table_name, keyspace_name=keyspace_name, key_type="bigint",
+                                     columns={"name": "text"})
+            stack.callback(drop_keyspace, node=working_node)
+
+            with simulate_node_unavailability(self.target_node):
+                # target node stopped by Contextmanger. Wait while its status will be updated
+                wait_for(node_operations.is_node_seen_as_down, timeout=600, throw_exc=True,
+                         down_node=self.target_node, verification_node=working_node, text=f"Wait other nodes see {self.target_node.name} as DOWN...")
+                self.log.debug("Remove node %s : hostid: %s with blocked scylla from cluster",
+                               self.target_node.name, target_host_id)
+                working_node.run_nodetool(f"removenode {target_host_id}", retry=0, long_running=True)
+                assert node_operations.is_node_removed_from_cluster(removed_node=self.target_node, verification_node=working_node), \
+                    f"Node {self.target_node.name} with host id {target_host_id} was not removed. See log errors"
+
+                # Context manager at exit  start scylla on target node.
+                # But node already removed from cluster. So any operations from it
+                # should be banned. If query executed succesfull, raise an error
+            assert self.target_node.db_up(), f"Scylla was not up on node {self.target_node.name}"
+
+            with self.cluster.cql_connection_exclusive(node=self.target_node) as session:
+                for key in random.sample(range(1, 100001), 1000):
+                    try:
+                        stmt = SimpleStatement(f"INSERT INTO {keyspace_name}.{table_name} (key, name) VALUES ({key}, 'name{key}');",
+                                               consistency_level=ConsistencyLevel.QUORUM)
+                        session.execute(stmt)
+                        self.log.error("Banned query passed to cluster from banned node")
+                        raise BannedQueryExecUnexpectedSuccess(
+                            "Query from banned node was executed succesful with Consistency.QUORUM")
+                    except (NoHostAvailable, OperationTimedOut, Unavailable) as exc:
+                        self.log.debug("Query failed with error: %s as expected", exc)
+
+            with self.cluster.cql_connection_patient(working_node) as session:
+                LOGGER.debug("Check keyspace %s.%s is empty", keyspace_name, table_name)
+                result = list(session.execute(f"SELECT * from {keyspace_name}.{table_name}"))
+                LOGGER.debug("Query result %s", result)
+                assert not result, f"New rows were added from banned node, {result}"
+
+            drop_keyspace(working_node)
 
 
 def disrupt_method_wrapper(method, is_exclusive=False):  # pylint: disable=too-many-statements  # noqa: PLR0915
@@ -7058,3 +7145,21 @@ class SerialRestartOfElectedTopologyCoordinatorNemesis(Nemesis):
 
     def disrupt(self):
         self.disrupt_serial_restart_elected_topology_coordinator()
+
+
+class IsolateNodeWithProcessSignalNemesis(Nemesis):
+    disruptive = True
+    topology_changes = True
+    kubernetes = False
+
+    def disrupt(self):
+        self.disrupt_refuse_connection_with_send_sigstop_signal_to_scylla_on_banned_node()
+
+
+class IsolateNodeWithIptableRuleNemesis(Nemesis):
+    disruptive = True
+    topology_changes = True
+    kubernetes = False
+
+    def disrupt(self):
+        self.disrupt_refuse_connection_with_block_scylla_ports_on_banned_node()
--- a/sdcm/utils/nemesis_utils/node_operations.py
+++ b/sdcm/utils/nemesis_utils/node_operations.py
@@ -1,0 +1,51 @@
+import contextlib
+import logging
+
+from typing import Optional
+
+from sdcm.cluster import BaseNode
+
+LOGGER = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def block_scylla_ports(target_node: "BaseNode", ports: list[int] | None = None):
+    ports = ports or [7001, 7000, 9042, 9142, 19042, 19142]
+    target_node.install_package("iptables")
+    target_node.start_service("iptables", ignore_status=True)
+    target_node.log.debug("Block connections %s", target_node.name)
+    for port in ports:
+        target_node.remoter.sudo(f"iptables -A INPUT -p tcp --dport {port} -j DROP")
+        target_node.remoter.sudo(f"iptables -A OUTPUT -p tcp --dport {port} -j DROP")
+    yield
+    target_node.log.debug("Remove all iptable rules %s", target_node.name)
+    for port in ports:
+        target_node.remoter.sudo(f"iptables -D INPUT -p tcp --dport {port} -j DROP")
+        target_node.remoter.sudo(f"iptables -D OUTPUT -p tcp --dport {port} -j DROP")
+    target_node.stop_service("iptables", ignore_status=True)
+
+
+@contextlib.contextmanager
+def pause_scylla_with_sigstop(target_node: "BaseNode"):
+    target_node.log.debug("Send signal SIGSTOP to scylla process on node %s", target_node.name)
+    target_node.remoter.sudo("pkill --signal SIGSTOP -e scylla", timeout=60)
+    yield
+    target_node.log.debug("Send signal SIGCONT to scylla process on node %s", target_node.name)
+    target_node.remoter.sudo(cmd="pkill --signal SIGCONT -e scylla", timeout=60)
+
+
+def is_node_removed_from_cluster(removed_node: BaseNode, verification_node: BaseNode) -> bool:
+    LOGGER.debug("Verification node %s", verification_node.name)
+    cluster_status: Optional[dict] = removed_node.parent_cluster.get_nodetool_status(
+        verification_node=verification_node)
+    if not cluster_status:
+        return False
+    result = []
+    for dc in cluster_status:
+        result.append(removed_node.ip_address not in cluster_status[dc].keys())
+    return all(result)
+
+
+def is_node_seen_as_down(down_node: BaseNode, verification_node: BaseNode) -> bool:
+    LOGGER.debug("Verification node %s", verification_node.name)
+    return down_node not in verification_node.parent_cluster.get_nodes_up_and_normal(verification_node)
--- a/sdcm/utils/raft/__init__.py
+++ b/sdcm/utils/raft/__init__.py
@@ -13,6 +13,7 @@ from sdcm.sct_events import Severity
 from sdcm.utils.features import is_consistent_topology_changes_feature_enabled, is_consistent_cluster_management_feature_enabled
 from sdcm.utils.health_checker import HealthEventsGenerator
 from sdcm.wait import wait_for
+from sdcm.rest.raft_api import RaftApi
 
 
 LOGGER = logging.getLogger(__name__)
@@ -137,6 +138,9 @@ class RaftFeatureOperations(ABC):
         log_patterns = self.TOPOLOGY_OPERATION_LOG_PATTERNS.get(operation)
         return list(map(self.get_message_waiting_timeout, log_patterns))
 
+    def call_read_barrier(self):
+        ...
+
 
 class RaftFeature(RaftFeatureOperations):
     TOPOLOGY_OPERATION_LOG_PATTERNS: dict[TopologyOperations, Iterable[MessagePosition]] = {
@@ -172,19 +176,24 @@ class RaftFeature(RaftFeatureOperations):
         state = self.get_status()
         return state == "use_post_raft_procedures"
 
+    def get_group0_id(self, session) -> str:
+        try:
+            row = session.execute("select value from system.scylla_local where key = 'raft_group0_id'").one()
+            return row.value
+        except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
+            err_msg = f"Get group0 members failed with error: {exc}"
+            LOGGER.error(err_msg)
+            return ""
+
     def get_group0_members(self) -> list[dict[str, str]]:
         LOGGER.debug("Get group0 members")
         group0_members = []
         try:
             with self._node.parent_cluster.cql_connection_patient_exclusive(node=self._node) as session:
-                row = session.execute("select value from system.scylla_local where key = 'raft_group0_id'").one()
-                if not row:
-                    return []
-                raft_group0_id = row.value
-
+                raft_group0_id = self.get_group0_id(session)
+                assert raft_group0_id, "Group0 id was not found"
                 rows = session.execute(f"select server_id, can_vote from system.raft_state  \
                                         where group_id = {raft_group0_id} and disposition = 'CURRENT'").all()
-
                 for row in rows:
                     group0_members.append({"host_id": str(row.server_id),
                                            "voter": row.can_vote})
@@ -347,6 +356,28 @@ class RaftFeature(RaftFeatureOperations):
         LOGGER.debug("Group0 and token-ring are consistent on node %s (host_id=%s)...",
                      self._node.name, self._node.host_id)
 
+    def call_read_barrier(self):
+        """ Wait until the node applies all previously committed Raft entries
+
+        Any schema/topology changes are committed with Raft group0 on node. Before
+        change is written to node, raft group0 checks that all previous schema/
+        topology changes were applied. Raft triggers read_barrier on node, and
+        node applies all previous changes(commits) before new schema/operation
+        write will be applied. After read barrier finished, it guarantees that
+        node has all schema/topology changes, which was done in cluster before
+        read_barrier started on node.
+
+        """
+        with self._node.parent_cluster.cql_connection_patient_exclusive(node=self._node) as session:
+            raft_group0_id = self.get_group0_id(session)
+            assert raft_group0_id, "Group0 id was not found"
+        try:
+            api = RaftApi(self._node)
+            result = api.read_barrier(group_id=raft_group0_id)
+            LOGGER.debug("Api response %s", result)
+        except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
+            LOGGER.error("Trigger read-barrier via rest api failed %s", exc)
+
 
 class NoRaft(RaftFeatureOperations):
     TOPOLOGY_OPERATION_LOG_PATTERNS = {
@@ -405,6 +436,9 @@ class NoRaft(RaftFeatureOperations):
         LOGGER.debug("Raft feature is disabled on node %s (host_id=%s)", self._node.name, self._node.host_id)
 
         yield None
+
+    def call_read_barrier(self):
+        ...
 
 
 def get_raft_mode(node) -> RaftFeature | NoRaft:
--- a/unit_tests/test_nemesis_sisyphus.py
+++ b/unit_tests/test_nemesis_sisyphus.py
@@ -80,7 +80,7 @@ def test_list_all_available_nemesis(generate_file=True):
     disruption_list, disruptions_dict, disruption_classes = sisyphus.get_list_of_disrupt_methods(
         subclasses_list=subclasses, export_properties=True)
 
-    assert len(disruption_list) == 90
+    assert len(disruption_list) == 92
 
     if generate_file:
         with open(sct_abs_path('data_dir/nemesis.yml'), 'w', encoding="utf-8") as outfile1:
--- a/data_dir/nemesis_classes.yml
+++ b/data_dir/nemesis_classes.yml
@@ -29,6 +29,8 @@
 - EnospcMonkey
 - GrowShrinkClusterNemesis
 - HardRebootNodeMonkey
+- IsolateNodeWithIptableRuleNemesis
+- IsolateNodeWithProcessSignalNemesis
 - LoadAndStreamMonkey
 - MajorCompactionMonkey
 - MemoryStressMonkey
--- a/jenkins-pipelines/oss/nemesis/longevity-5gb-1h-IsolateNodeWithIptableRuleNemesis-aws.jenkinsfile
+++ b/jenkins-pipelines/oss/nemesis/longevity-5gb-1h-IsolateNodeWithIptableRuleNemesis-aws.jenkinsfile
@@ -1,0 +1,12 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    backend: 'aws',
+    region: 'eu-west-1',
+    test_name: 'longevity_test.LongevityTest.test_custom_time',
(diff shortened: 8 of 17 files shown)


aleksbykov:
Currently staging jobs are failed with https://github.com/scylladb/scylladb/issues/15686

aleksbykov:
@scylladb/qa-maintainers could you take a look?

jenkins-pipelines/nemesis/longevity-5gb-1h-IsolateNodeWithIptableRuleNemesis-aws.jenkinsfile
roydahan, line 1: please add also for Azure.

jenkins-pipelines/nemesis/longevity-5gb-1h-IsolateNodeWithProcessSignalNemesis-aws.jenkinsfile
roydahan, line 1: Same

sdcm/nemesis.py
```diff
@@ -4904,6 +4905,121 @@ def are_gate_closed_messages_raised(log_reader):
...
+        def prepare_test_keyspace(node):
+            """ Create new table for verification missing data
+
+            Create new keyspace and table which allow to
+            verify that no operations come from banned node
+            """
+            with self.cluster.cql_connection_patient(node) as session:
+                session.execute(
+                    "CREATE KEYSPACE IF NOT EXISTS banned_keyspace \
+                        WITH replication = {'class': 'SimpleStrategy', 'replication_factor' : 3};")
```
soyacz, line 4935: let's use `NetworkTopologyStrategy`

sdcm/nemesis.py
```diff
@@ -4904,6 +4905,121 @@ def are_gate_closed_messages_raised(log_reader):
...
+                session.execute("CREATE TABLE IF NOT EXISTS banned_keyspace.table1 (id BIGINT PRIMARY KEY, name text);")
+
+        def read_barrier(node):
+            """ issue read barrier on node to check that
+            node topology and schema are updated"""
+            with self.cluster.cql_connection_patient(node) as session:
+                session.execute("DROP TABLE IF EXISTS noexist_ks.noexist_cf;")
+
+        @contextlib.contextmanager
+        def use_network_firwall(target_node: BaseNode):
```
soyacz, line 4945: I propose to separate `simulate_node_unavailability` context manager to separate file so we don't extend more `nemesis.py` file. Also it will enhance SCT framework with reusable functionality.
My proposal:
```
def simulate_node_unavailability(node: BaseNode, using_iptables: bool = False)
    if using_iptables:
        return use_network_firewall(node)
    else:
        return use_stop_signal(node)
```

We even could go further by adding `wait_for_cluster_ack_node_is_down` argument to this method as it seems to be quite common and further reduce `nemesis.py` complexity.

sdcm/nemesis.py
```diff
@@ -4904,6 +4905,121 @@ def are_gate_closed_messages_raised(log_reader):
...
+            # Run read barrier on working node to validate that it update raft topology state
+            read_barrier(working_node)
+
+        # Context manager at exit  start scylla on target node.
+        # But node already removed from cluster. So any operations from it
+        # should be banned. If query executed succesfull, raise an error
+        try:
+            with self.cluster.cql_connection_exclusive(node=self.target_node) as session:
+                stmt = SimpleStatement("INSERT INTO banned_keyspace.table1 (id, name) VALUES (1, 'name1');",
+                                       consistency_level=ConsistencyLevel.ALL)
```
soyacz, line 5001: shouldn't we test also consistency level `ONE`?
so later we could validate if inserted row didn't go through some other path than query.

sdcm/nemesis.py
```diff
@@ -4904,6 +4905,121 @@ def are_gate_closed_messages_raised(log_reader):
...
+            verify that no operations come from banned node
+            """
+            with self.cluster.cql_connection_patient(node) as session:
+                session.execute(
+                    "CREATE KEYSPACE IF NOT EXISTS banned_keyspace \
+                        WITH replication = {'class': 'SimpleStrategy', 'replication_factor' : 3};")
+                session.execute("CREATE TABLE IF NOT EXISTS banned_keyspace.table1 (id BIGINT PRIMARY KEY, name text);")
+
+        def read_barrier(node):
+            """ issue read barrier on node to check that
```
soyacz, line 4939: can you improve comment explaining how this is creating `read barrier` and what it means?

unit_tests/test_nemesis_sisyphus.py
soyacz, line 1: please update also `nemesis.yaml`

sdcm/nemesis.py
aleksbykov: Level One, could be unstable, because sometime the query will succesfully write on banned node (if partition will belong to the node) and some time it will be failed, because the query will be executed on the the not available nodes

jenkins-pipelines/nemesis/longevity-5gb-1h-IsolateNodeWithIptableRuleNemesis-aws.jenkinsfile
aleksbykov: Added

jenkins-pipelines/nemesis/longevity-5gb-1h-IsolateNodeWithProcessSignalNemesis-aws.jenkinsfile
aleksbykov: Added

sdcm/nemesis.py
aleksbykov: fixed

unit_tests/test_nemesis_sisyphus.py
aleksbykov: done

sdcm/nemesis.py
aleksbykov: Done

unit_tests/test_nemesis_sisyphus.py
aleksbykov: done

sdcm/nemesis.py
aleksbykov: Fixed

roydahan:
In my queue, hopefully by tomorrow.

data_dir/nemesis.yml
```diff
@@ -332,3 +332,8 @@
   - disruptive = False
   - kubernetes = True
   - free_tier_set = True
+- disrupt_refusing_connection_from_banned_node:
+  - disruptive = True
+  - kubernetes = False
+  - free_tier_set = True
```
roydahan, line 338: I think it's not relevant.

sdcm/cluster.py
```diff
@@ -4875,6 +4875,16 @@ def get_db_nodes_cpu_mode(self):
...
 
+    def read_barrier(self, node: BaseNode):
+        """ Trigger read_barrier on the node
+
+        Read_barrier is functionality, which guarantees that after it finished,
+        node has applied all Raft commits, which was done in cluster before
+        read_barrier started on node.
+        """
+        with self.cql_connection_patient(node) as session:  # pylint: disable=no-member
+            session.execute("DROP TABLE IF EXISTS noexist_ks.noexist_cf;")
```
roydahan, line 4886: Sorry, but hard for me to understand how the above explanation related to the code.
Please explain how dropping this table if exist related.

sdcm/nemesis.py
roydahan: @aleksbykov IIUC what you're trying to do here, it will work deterministically only when cluster size is 3.
Since you're using RF=3, it doesn't force one replica to reside on the target node that you just took out of the cluster.
Maybe you should change the RF of this Banned keyspace to be as big as the cluster size.

sdcm/nemesis.py
```diff
@@ -4960,6 +4962,82 @@ def are_gate_closed_messages_raised(log_reader):
...
+                session.execute(stmt)
+                raise BannedCQLOperation("wrong operations were executed")
+        finally:
+            # because target node removed from cluster, terminate it
+            # and bootstrap new one
+            LOGGER.debug("Terminating node %s", self.target_node.name)
+            self.cluster.terminate_node(self.target_node)
+            LOGGER.debug("Restore number of cluster nodes")
+            new_node = self._add_and_init_new_cluster_node()
+            LOGGER.debug("New node %s was added", new_node.name)
```
roydahan, line 5029: Possibly not enough just to add, if RBNO isn't enabled, you need to run repair.
Otherwise, you may end up with possible data loss.

sdcm/nemesis.py
```diff
@@ -4960,6 +4962,82 @@ def are_gate_closed_messages_raised(log_reader):
...
+            try:
+                with self.cluster.cql_connection_patient(new_node) as session:
+                    LOGGER.debug("Check create keyspace is empty")
+                    result = list(session.execute("SELECT * from banned_keyspace.table1"))
+                    LOGGER.debug("Query result %s", result)
+                    assert not result, f"New rows were added from banned node, {result}"
```
roydahan, line 5035: I don't understand this assert message or possibly the logic here.
What are you trying to check?

Also, why is it relevant only after the new node addition or from the new node and not from the working node.

sdcm/nemesis.py
```diff
@@ -4960,6 +4962,82 @@ def are_gate_closed_messages_raised(log_reader):
...
+        If node was removed from cluster for any reason, even if removed node
+        become alive and try to communicate with rest node in cluster, all connections
+        from it should be refused by other nodes in cluster
+        1. on target node block any scylladb process
+        1.1 Pause process
+        1.2 Block with iptables port 9100/10000
+        2. Wait and remove target node from cluster
+        3. start scylla on target node
+        4. Create exclusive connection to target node
+        5. Execute cql command on target node and validate that no operation
```
roydahan, line 4977: How is executing cql connection from a node related to the fact it was removed from the cluster?

sdcm/nemesis.py
```diff
@@ -6458,3 +6536,22 @@ class DisableBinaryGossipExecuteMajorCompaction(Nemesis):
...
+
+    def disrupt(self):
+        self.disrupt_refusing_connection_from_banned_node(use_iptables=False)
+
+
+class IsolateNodeWithIptableRuleNemesis(Nemesis):
+    disruptive = True
+    topology_changes = True
+    kubernetes = False
+    free_tier_set = True
```
roydahan, line 6554: remove

roydahan (changes requested):
Please see my comments.
Also, please test with a regular longevity that is n_db_nodes > 3.

sdcm/nemesis.py
aleksbykov: @roydahan , these lines should be remove? and file sdcm/nemesis.py shouldn't be updated?

sdcm/nemesis.py
aleksbykov: Scylla was paused on the node, and after the node was removed from cluster, scylla was starting again. The removed node still think that it is in the cluster. But the cluster doesn't know about it and closed all connections from it. So when run any cql command on the removed node, it executes it and try to replicate to rest nodes on cluster. But cluster have to  ban any connection from this node

sdcm/nemesis.py
aleksbykov: The logic is next:
if cql query wsa executed succesfully from banned node, nemesis raises exception, but because cluster has not enough nodes, in _finaly_ block new node was added to restore number of nodes , and checking that data was written from banned node are replicatated to cluster with assert

if cql query failed to be executed, it means that node was banned, in _finaly_ block we again restore the number of cluster nodes and verify again that no data from banned node are in cluster, if somehow data from banned node reach cluster, assert will raise an issue.

sdcm/nemesis.py
aleksbykov: Added repair for all nodes

sdcm/cluster.py
aleksbykov: Added more explanations

sdcm/nemesis.py
roydahan: Only the "free_tier_set", I don't think it's relevant for this nemesis.

sdcm/nemesis.py
roydahan: Ok, but:
Do you check what is the actual failure?
It doesn't look so, in theory the cql port should be opened, but it can be down and the failure will still be considered as failed, isn't it?

I'll comment also in the actual step.

sdcm/nemesis.py
roydahan: Going back to this, with bit more understanding of what you're trying to test here.
CL=ALL is enough here as long as RF > 1, since it tries to replicate the other replicas to a cluster that it's not part of anymore.

data_dir/nemesis.yml
aleksbykov: fixed

sdcm/nemesis.py
aleksbykov: fixed

sdcm/nemesis.py
aleksbykov: @roydahan , So you suggest to add specific Exceptions or add a check that port on banned node is opened? Do i correctly understand you?

aleksbykov:
@soyacz , could you review?

sdcm/utils/raft/__init__.py
```diff
@@ -122,6 +126,14 @@ def __init__(self, node: "BaseNode") -> None:
     def is_enabled(self) -> bool:
         return True
 
+    @property
+    def consistent_topology_changes_enabled(self) -> bool:
+        with self._node.remote_scylla_yaml() as scylla_yaml:
+            if self._node.is_kubernetes():
+                return "consistent-topology-changes" in scylla_yaml.get("experimental_features", [])
```
soyacz, line 133: @vponomaryov why it is necessary to do such exceptions for k8s? can we make it seamless like on vm based nodes so we don't have to remember to have different logic for k8s?

sdcm/cluster.py
soyacz: sorry, but I still don't understand `read barrier` concept and can't find any docs about it. E.g. what 'read' we refer to?
I also don't understand concept of 'raft commits' - are they referred to schema or topology?

Maybe better to rename this method to 'ensure_...` or `wait_for_raft...` 

Can this cql timeout when under high load and we have some still some raft commits to apply?

sdcm/nemesis.py
```diff
@@ -4973,6 +4974,106 @@ def are_gate_closed_messages_raised(log_reader):
...
+        with self.cluster.cql_connection_exclusive(node=self.target_node) as session:
+            for consistency_level in consistency_levels:
+                level_name = consistency_value_to_name(consistency_level)
+                query_results[level_name] = True
+                try:
+                    stmt = SimpleStatement("INSERT INTO banned_keyspace.table1 (id, name) VALUES (1, 'name1');",
+                                           consistency_level=consistency_level)
+                    session.execute(stmt)
+                except Exception as exc:  # pylint: disable=broad-except
+                    self.log.debug("Query failed with error: %s", exc)
```
soyacz, line 5043: if a node doesn't know it was banned, how CL=ONE can fail?

soyacz (commented):
still some remarks and questions

sdcm/utils/raft/__init__.py
vponomaryov: K8S has the `scylla_yaml` as dict. VMs have it as data class instance with attributes.
Why? Because the data class instance is not suitable for K8S where Scylla configuration is partially configured only by operator.
There are several solutions. One of them is to add `__getitem__` support for the data class instance used for VMs. So, after it we could use 'get' method everywhere.

sdcm/exceptions.py
```diff
@@ -71,3 +71,7 @@ class WaitForTimeoutError(Exception):
 
 class ExitByEventError(Exception):
     """Exception that would be raised if wait.wait_for stopped by event"""
+
+
+class BannedCQLOperation(Exception):
```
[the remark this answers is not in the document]
aleksbykov: yes, i removed it

sdcm/nemesis.py
```diff
@@ -4973,6 +4974,106 @@ def are_gate_closed_messages_raised(log_reader):
...
+                    query_results[level_name] = False
+        try:
+            assert not any(list(query_results.values())), \
+                f"Query from banned node was executed succesful for CL: {query_results}"
+        finally:
+            # because target node removed from cluster, terminate it
+            # and bootstrap new one
+            LOGGER.debug("Terminating node %s", self.target_node.name)
+            self.cluster.terminate_node(self.target_node)
```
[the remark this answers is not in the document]
aleksbykov: i wrapped more code.

sdcm/cluster.py
aleksbykov: @soyacz , i updated the doc string, hope it become a bit clear. Read barrier is part of Raft feature, it will be difficult to put in doc string description for all raft feature

aleksbykov:
@soyacz , can you review?

soyacz (approved):
lgtm
can you share links with test drive with recent changes?

aleksbykov:
> lgtm can you share links with test drive with recent changes?

@soyacz can you review again. i rebased to latest master of sct and fix problem with iptables. Latest run also passed for both modes:
JOB: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h/86

aleksbykov:
@fruch @vponomaryov can you review?

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+from typing import Optional
+from random import choice
+
+from sdcm.cluster import BaseNode
+
+LOGGER = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def use_network_firewall(target_node: "BaseNode"):
```
soyacz, line 13: rename: `blocked_scylla_ports`

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+        target_node.remoter.sudo(f"iptables -A OUTPUT -p tcp --dport {port} -j DROP")
+    yield
+    target_node.log.debug("Remove all iptable rules %s", target_node.name)
+    for port in ports:
+        target_node.remoter.sudo(f"iptables -D INPUT -p tcp --dport {port} -j DROP")
+        target_node.remoter.sudo(f"iptables -D OUTPUT -p tcp --dport {port} -j DROP")
+
+
+@contextlib.contextmanager
+def use_send_signal_process(target_node: "BaseNode"):
```
soyacz, line 28: rename: `scylla_stopped_by_sigstop`

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+    yield
+    target_node.log.debug("Send signal SIGCONT to scylla process on node %s", target_node.name)
+    target_node.remoter.sudo(cmd="pkill --signal SIGCONT -e scylla", timeout=60)
+
+
+def get_random_free_verification_node(node: "BaseNode") -> "BaseNode":
+    return choice([active_node for active_node in node.parent_cluster.nodes if not active_node.running_nemesis])
+
+
+def check_node_removed_from_cluster(node: "BaseNode"):
```
soyacz, line 40: rename: `is_node_removed_from_cluster`
add return type hint

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+    cluster_status: Optional[dict] = node.parent_cluster.get_nodetool_status(verification_node=verification_node)
+    if not cluster_status:
+        return False
+    result = []
+    for dc in cluster_status:
+        result.append(node.ip_address not in list(cluster_status[dc].keys()))
+    return all(result)
+
+
+def check_node_seen_as_down_in_cluster(down_node: BaseNode):
```
soyacz, line 53: rename: `is_node_seen_as_down`
add return type hint

sdcm/nemesis.py
```diff
@@ -5006,6 +5008,113 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+            verify that no operations come from banned node
+            """
+            with self.cluster.cql_connection_patient(node) as session:
+                session.execute(
+                    f"CREATE KEYSPACE IF NOT EXISTS banned_keyspace \
+                        WITH replication = {{'class': 'NetworkTopologyStrategy', \
+                            'replication_factor' : {self.tester.params.get('n_db_nodes')} }};")
+                session.execute("CREATE TABLE IF NOT EXISTS banned_keyspace.table1 (id BIGINT PRIMARY KEY, name text);")
+
+        simulate_node_unavailability = node_operations.simulate_node_unavailability(use_iptables)
```
soyacz, line 5042: `node_unavailable` = `'blocked_scylla_ports' if use_iptables else 'scylla_stopped_by_sigstop'`

I don't see value in hiding this in function.

soyacz (commented):
Proposal for some names change to be more readable (The reason of renaming is noted in [pep](https://peps.python.org/pep-0343/#examples)

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+def check_node_removed_from_cluster(node: "BaseNode"):
+    LOGGER.debug("Get random verification node without running nemesis")
+    verification_node = get_random_free_verification_node(node)
+    LOGGER.debug("Chosen verification node %s", verification_node.name)
+    cluster_status: Optional[dict] = node.parent_cluster.get_nodetool_status(verification_node=verification_node)
+    if not cluster_status:
+        return False
+    result = []
+    for dc in cluster_status:
+        result.append(node.ip_address not in list(cluster_status[dc].keys()))
```
vponomaryov, line 49: Why wrapping into a `list`?
The `cluster_status[dc].keys()` is iterable and should work ok with `in` operator.

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+        target_node.remoter.sudo(f"iptables -D INPUT -p tcp --dport {port} -j DROP")
+        target_node.remoter.sudo(f"iptables -D OUTPUT -p tcp --dport {port} -j DROP")
+
+
+@contextlib.contextmanager
+def use_send_signal_process(target_node: "BaseNode"):
+    target_node.log.debug("Send signal SIGSTOP to scylla process on node %s", target_node.name)
+    target_node.remoter.sudo("pkill --signal SIGSTOP -e scylla", timeout=60)
+    yield
+    target_node.log.debug("Send signal SIGCONT to scylla process on node %s", target_node.name)
```
vponomaryov, line 32: Shouldn't the node log prefix be enough to understand what the target node is?

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+from random import choice
+
+from sdcm.cluster import BaseNode
+
+LOGGER = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def use_network_firewall(target_node: "BaseNode"):
+    ports = [7001, 7000, 9042, 9142, 19042, 19142]
```
vponomaryov, line 14: Can we make it be configurable?
It seems useful for some other cases too where we may want to block only one/two ports...

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,64 @@
...
+
+from typing import Optional
+from random import choice
+
+from sdcm.cluster import BaseNode
+
+LOGGER = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
```
vponomaryov, line 12: This func definitely will not work on K8S.
Can you please add check for K8S and raise explicit exception, i.e. `NotImplementedError`?

sdcm/nemesis.py
```diff
@@ -5006,6 +5008,113 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+        1. on target node block any scylladb process
+        1.1 Pause process
+        1.2 Block with iptables port 9100/10000
+        2. Wait and remove target node from cluster
+        3. start scylla on target node
+        4. Create exclusive connection to target node
+        5. Execute cql command on target node and validate that no operation
+        from target node passed to cluster
+        """
+        if not self.target_node.raft.consistent_topology_changes_enabled:
```
vponomaryov, line 5026: Add check for K8S and skip if it is on it.

sdcm/nemesis.py
```diff
@@ -5006,6 +5008,113 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                    stmt = SimpleStatement("INSERT INTO banned_keyspace.table1 (id, name) VALUES (1, 'name1');",
+                                           consistency_level=ConsistencyLevel.ALL)
+                    session.execute(stmt)
+                    raise BannedQueryExecutedSuccessful(
+                        "Query from banned node was executed succesful with Consistency.ALL")
+                except BannedQueryExecutedSuccessful:
+                    self.log.error("Banned query passed to cluster from banned node")
+                    raise
+                except Exception as exc:  # pylint: disable=broad-except
+                    self.log.debug("Query failed with error: %s as expected", exc)
```
vponomaryov, line 5078: How do we know that it failed with the expected error if we catch broad `Exception` exception?
Looks error prone as of now.

aleksbykov: Set explicit exception.

sdcm/nemesis.py
aleksbykov: added

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: Disable nemesis for K8S

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: Added optional parameter, which could take list of ports to disable

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: it will be easy to search and read during job investigation if log name will be in log message.

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: removed

sdcm/nemesis.py
aleksbykov: Removed extra function

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: done

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: done

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: done

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: done

aleksbykov:
job is running: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h/87

aleksbykov:
@vponomaryov @soyacz , fixed accroding your comments. Can you take a look?

soyacz (approved):
LGTM

aleksbykov:
@vponomaryov can you review?

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,115 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+            with self.cluster.cql_connection_exclusive(node=self.target_node) as session:
+                try:
+                    stmt = SimpleStatement("INSERT INTO banned_keyspace.table1 (id, name) VALUES (1, 'name1');",
+                                           consistency_level=ConsistencyLevel.ALL)
+                    session.execute(stmt)
+                    raise BannedQueryExecutedSuccessful(
+                        "Query from banned node was executed succesful with Consistency.ALL")
+                except BannedQueryExecutedSuccessful:
+                    self.log.error("Banned query passed to cluster from banned node")
+                    raise
```
vponomaryov, line 5081: This `except` looks redundant.

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,115 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+            # should be banned. If query executed succesfull, raise an error
+            assert self.target_node.is_port_used(
+                port=9042, service_name="scylla-service"), f"Port 9042 is closed on node {self.target_node.name}"
+
+            with self.cluster.cql_connection_exclusive(node=self.target_node) as session:
+                try:
+                    stmt = SimpleStatement("INSERT INTO banned_keyspace.table1 (id, name) VALUES (1, 'name1');",
+                                           consistency_level=ConsistencyLevel.ALL)
+                    session.execute(stmt)
+                    raise BannedQueryExecutedSuccessful(
```
vponomaryov, line 5077: Maybe better to change the name from the meaning point of view?
Example: `BannedQueryExecUnexpectedSuccess`?

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,115 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+        finally:
+            # because target node removed from cluster, terminate it
+            # and bootstrap new one
+            LOGGER.debug("Terminating node %s", self.target_node.name)
+            self.cluster.terminate_node(self.target_node)
+            # check node was removed previously, if not remove it from cluster
+            removed_node_status = self.cluster.get_node_status_dictionary(
+                ip_address=self.target_node.ip_address, verification_node=working_node)
+            if removed_node_status is not None:
+                working_node.run_nodetool(f"removenode {target_host_id}", retry=3)
```
vponomaryov, line 5097: Why do we expect 2 possible states after running `nodetool removenode` on the line 5059?
And why we are ok with both?

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,115 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                    try:
+                        self.repair_nodetool_repair(node=node, publish_event=False)
+                    except Exception as details:  # pylint: disable=broad-except
+                        self.log.error(f"failed to execute repair command "
+                                       f"on node {node} due to the following error: {str(details)}")
+
+            try:
+                with self.cluster.cql_connection_patient(new_node) as session:
+                    LOGGER.debug("Check create keyspace is empty")
+                    result = list(session.execute("SELECT * from banned_keyspace.table1"))
```
vponomaryov, line 5114: Such a thing as the keyspace and table names better to define once as vars and then use it to avoid possible typos mismatches.

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,115 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                    LOGGER.debug("Query result %s", result)
+                    assert not result, f"New rows were added from banned node, {result}"
+            finally:
+                with self.cluster.cql_connection_patient(working_node) as session:
+                    LOGGER.debug("Drop create keyspace")
```
vponomaryov, line 5119: In the `Drop create keyspace` statement I see 2 commands `drop` and then `create` a keyspace.
But according to the code below it is not correct assumption.
So, better to rephrase it to avoid misunderstandings.

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,115 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                    raise BannedQueryExecutedSuccessful(
+                        "Query from banned node was executed succesful with Consistency.ALL")
+                except BannedQueryExecutedSuccessful:
+                    self.log.error("Banned query passed to cluster from banned node")
+                    raise
+                except (NoHostAvailable, OperationTimedOut, Unavailable) as exc:
+                    self.log.debug("Query failed with error: %s as expected", exc)
+
+        except Exception as exc:  # pylint: disable=broad-except
+            self.log.error("Method failed with error %s", exc)
```
vponomaryov, line 5086: Too ambiguous statement. Better to replace `method` with concrete type of it.

sdcm/nemesis.py
aleksbykov: you right. we can just raise

sdcm/nemesis.py
aleksbykov: renamed

sdcm/nemesis.py
aleksbykov: Once we trigger removenode, the node marked as non-voter and banned from cluster. The banned node and rest nodes in the cluster could see each other with different state. So if for any reason any step is failed or passed (failed to remove node, not blocking ports, or anything else) or node was removed successfully.   in finally block, i want to restore the cluster health. for that i need to remove target node totally from cluster and add new one. So if any reason target node stay in cluster in main nemesis flow, it will be removed from cluster in finally block and doesn't matter in which state it was.

sdcm/nemesis.py
aleksbykov: updated

sdcm/nemesis.py
aleksbykov: Updated

aleksbykov:
@vponomaryov @soyacz can you review again?

aleksbykov:
@soyacz @fruch @vponomaryov can you review it?

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,116 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                                           consistency_level=ConsistencyLevel.ALL)
+                    session.execute(stmt)
+                    self.log.error("Banned query passed to cluster from banned node")
+                    raise BannedQueryExecUnexpectedSuccess(
+                        "Query from banned node was executed succesful with Consistency.ALL")
+                except (NoHostAvailable, OperationTimedOut, Unavailable) as exc:
+                    self.log.debug("Query failed with error: %s as expected", exc)
+
+        except Exception as exc:  # pylint: disable=broad-except
+            self.log.error("Isolate node with %s failed with error %s", simulate_node_unavailability.__name__, exc)
```
soyacz, line 5087: should also `session.execute(f"DROP KEYSPACE IF EXISTS {keyspace_name}")` in case of reraising exception - as rest of code won't be executed and keyspace not cleaned up.

soyacz (commented):
Where are GCE jenkins jobs?

Also rebase/push a commit so unit tests are retriggered.

sdcm/nemesis.py
aleksbykov: It will be deleted in finally block anyway.

aleksbykov:
@soyacz . done

sdcm/nemesis.py
```diff
@@ -6564,3 +6677,21 @@ class EndOfQuotaNemesis(Nemesis):
...
+class IsolateNodeWithProcessSignalNemesis(Nemesis):
+    disruptive = True
+    topology_changes = True
+    kubernetes = False
+
+    def disrupt(self):
+        self.disrupt_refusing_connection_from_banned_node(use_iptables=False)
+
+
+class IsolateNodeWithIptableRuleNemesis(Nemesis):
```
fruch, line 6691: This option isn't gonna be used by any Nemesis

Only if this class is used directly.

disrupt function shouldn't have parameters, it causes the wrong expectations.

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,57 @@
...
+def stop_scylla_with_sigstop(target_node: "BaseNode"):
+    target_node.log.debug("Send signal SIGSTOP to scylla process on node %s", target_node.name)
+    target_node.remoter.sudo("pkill --signal SIGSTOP -e scylla", timeout=60)
+    yield
+    target_node.log.debug("Send signal SIGCONT to scylla process on node %s", target_node.name)
+    target_node.remoter.sudo(cmd="pkill --signal SIGCONT -e scylla", timeout=60)
+
+
+def get_random_free_verification_node(node: "BaseNode") -> "BaseNode":
```
fruch, line 36: There is a function for this that marks the node as nemesis, no need to reinvent it

This isn't safe for parallel nemesis

sdcm/utils/raft/__init__.py
```diff
@@ -124,6 +128,15 @@ def __init__(self, node: "BaseNode") -> None:
     def is_enabled(self) -> bool:
         return True
 
+    @property
+    def consistent_topology_changes_enabled(self) -> bool:
```
fruch, line 132: This function isn't correct now that this feature is enabled by default 

I would recommend reading the features of internal tables, and not from scylla.yaml

fruch (changes requested):
Need to adapt, now that topology is enabled by default

sdcm/nemesis.py
aleksbykov: create 2 different disrupt functions without parameter and switch particular classes to use them

sdcm/utils/nemesis_utils/node_operations.py
aleksbykov: i removed this function and use exist one

aleksbykov:
Job was passed: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h/93/

aleksbykov:
@fruch @soyacz @vponomaryov , can you review, please?

configurations/nemesis/longevity-5gb-1h-nemesis.yaml
```diff
@@ -19,7 +19,7 @@ instance_type_db: 'i4i.large'
 gce_instance_type_db: 'n1-highmem-2'
 gce_instance_type_loader: 'e2-standard-2'
 azure_instance_type_db: 'Standard_L8s_v3'
-instance_type_loader: 'c5.large'
+instance_type_loader: 'c6i.large'
```
roydahan, line 22: Should be in a separate commit.

sdcm/utils/raft/__init__.py
roydahan: @aleksbykov in his comment @fruch recommended to check if the feature is enabled by using internal tables instead of scylla_yaml.
Please check this option which would be safer.

roydahan (dismissed):
LGTM.
But there are small final adjustments need to be done in order to be merged.

sdcm/utils/raft/__init__.py
fruch: @aleksbykov 

this new logic basically mean we won't be able to run with SCT master with older releases of scylla
or worse we'll be able to run, but thing would fail unclearly, same the the issue we had in dtest cause of this kind of logic

I insist we should be reading the features out of scylla internal tables

fruch (changes requested):
`consistent_topology_changes_enabled` should be limiting us to only newer scylla versions

sdcm/utils/raft/__init__.py
aleksbykov: Fixed. Add new static methods and use them to check raft features state

aleksbykov:
@roydahan  @fruch i fixed accroding your latest comments. Could you please review

sdcm/utils/raft/__init__.py
```diff
@@ -4,12 +4,11 @@
 
 from enum import Enum
 from typing import Protocol, NamedTuple, Mapping, Iterable
-from collections import namedtuple
 
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.filters import EventsSeverityChangerFilter
 from sdcm.sct_events import Severity
-from sdcm.utils.version_utils import ComparableScyllaVersion
+# from sdcm.utils.version_utils import ComparableScyllaVersion
```
fruch, line 11: Remove this line

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,123 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+        2. Wait and remove target node from cluster
+        3. start scylla on target node
+        4. Create exclusive connection to target node
+        5. Execute cql command on target node and validate that no operation
+        from target node passed to cluster
+        """
+        if not self.target_node.raft.consistent_topology_changes_enabled:
+            raise UnsupportedNemesis("Raft feature: consistent-topology-changes is not enabled")
+        if self._is_it_on_kubernetes():
+            raise UnsupportedNemesis("Skip test for K8S because no supported yet")
```
fruch, line 5038: Not supported, you mean it was tested ?

Or are there some specific things you know aren't gonna work in k8s ?

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,123 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+        keyspace_name = "banned_keyspace"
+        table_name = "table1"
+
+        def prepare_test_keyspace(node):
+            """ Create new table for verification missing data
+
+            Create new keyspace and table which allow to
+            verify that no operations come from banned node
+            """
+            with self.cluster.cql_connection_patient(node) as session:
```
fruch, line 5048: Why not using any of the functionally available in SCT for creating keyspaces ?

creating a keyspace with RF=numbers of nodes, can be problematic for tablets, if the keyspace isn't gonna be evacuated.

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,123 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                    wait_for(node_operations.is_node_removed_from_cluster, timeout=300, throw_exc=True,
+                             removed_node=self.target_node, verification_node=working_node)
+                    # Run read barrier on working node to validate that it update raft topology state
+                    self.cluster.call_read_barrier(working_node)
+
+                # Context manager at exit  start scylla on target node.
+                # But node already removed from cluster. So any operations from it
+                # should be banned. If query executed succesfull, raise an error
+                assert self.target_node.is_port_used(
+                    port=9042, service_name="scylla-service"), f"Port 9042 is closed on node {self.target_node.name}"
```
fruch, line 5080: Why not using node.is_up ?
Why hard coding the port ?

What would happen if we enable client encryption and disable the clear port ? (It's doable, but we don't yet do that in SCT)

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,123 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+
+                with self.cluster.cql_connection_exclusive(node=self.target_node) as session:
+                    try:
+                        stmt = SimpleStatement(f"INSERT INTO {keyspace_name}.{table_name} (id, name) VALUES (1, 'name1');",
+                                               consistency_level=ConsistencyLevel.ALL)
```
fruch, line 5085: Why using ALL ? This nemesis might break during parallel nemesis. 

This nemesis is marked as destructive?

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,123 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                        self.log.debug("Query failed with error: %s as expected", exc)
+
+            except Exception as exc:  # pylint: disable=broad-except
+                self.log.error("Isolate node with %s failed with error %s", simulate_node_unavailability.__name__, exc)
+                raise
+            finally:
+                # because target node removed from cluster, terminate it
+                # and bootstrap new one
+                LOGGER.debug("Terminating node %s", self.target_node.name)
+                self.cluster.terminate_node(self.target_node)
```
fruch, line 5100: Flow of terminating, and then adding node can be problematic for tablets

We need to test those nemesis with tablet enabled (soon it would be the default)

sdcm/nemesis.py
```diff
@@ -5008,6 +5011,123 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                try:
+                    with self.cluster.cql_connection_patient(new_node) as session:
+                        LOGGER.debug("Check keyspace %s.%s is empty", keyspace_name, table_name)
+                        result = list(session.execute(f"SELECT * from {keyspace_name}.{table_name}"))
+                        LOGGER.debug("Query result %s", result)
+                        assert not result, f"New rows were added from banned node, {result}"
+                finally:
+                    with self.cluster.cql_connection_patient(working_node) as session:
+                        LOGGER.debug("Drop keyspace %s", keyspace_name)
+                        session.execute(f"DROP KEYSPACE IF EXISTS {keyspace_name}", timeout=300)
```
fruch, line 5128: There are so many steps to the "cleanup" of this nemesis, so on any issues with terminate or adding node, or repair, you aren't gonna reach this stage...

I would really recommend using multiple context managers or exitstack, cause you gonna have issues like that

fruch (commented):
In general LGTM, and my previous comments were addressed

1) we need a review from someone working on raft, I don't have the details to review correctness of what implemented here 

2) since tablets is going to be enabled in the upcoming days ,we need this tested with tablets before merging

sdcm/utils/raft/__init__.py
aleksbykov: fixed

sdcm/nemesis.py
aleksbykov: it was confirmed by Valery: https://github.com/scylladb/scylla-cluster-tests/pull/6577#discussion_r1511411383

sdcm/nemesis.py
aleksbykov: switch to use sct methods for create keyspace/table

sdcm/nemesis.py
aleksbykov: yes. nemesis is disruptive, because it remove and new node

sdcm/nemesis.py
aleksbykov: Change logic. I use same steps as for remove_node_add_node nemesis. for this nemesis i set it as callback for exit stack

sdcm/nemesis.py
aleksbykov: was changed

sdcm/nemesis.py
aleksbykov: switch to use it

aleksbykov:
New in latest changes:
 - refactor method remove_node_add_node: create new method which allow to remove node and the and new node. This could be used in any places, where cluster require remove node and add new node. Now it is used in 2 nemesis RemoveNodeThenAddnode and IsolateNode*
 - create new method to get group0 id. it will help to run read barrier via rest_api
 - add new rest api call to trigger read barrier for node it replace the cql command and avoid to create new method in cluster
 - refactor main method for new nemesis: _refuse_connection_from_banned_node. It become more readable. remove unnecessary block of try..except. 
Logic of nemesis stay the same:
block scylla on a node (iptables sigstop), wait while it be marked as down,  run removenode operation for it, release scylla, try to send to send any query from blocked node and verify that it failed and no data on other nodes. terminate id and add new node according instructions (with repair and cleanup).

aleksbykov:
@fruch @soyacz can you review?

sdcm/nemesis.py
```diff
@@ -5151,6 +5163,95 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+                wait_for(node_operations.is_node_seen_as_down, timeout=600, throw_exc=True,
+                         down_node=self.target_node, verification_node=working_node, text=f"Wait other nodes see {self.target_node.name} as DOWN...")
+                self.log.debug("Remove node %s : hostid: %s with blocked scylla from cluster",
+                               self.target_node.name, target_host_id)
+                working_node.run_nodetool(f"removenode {target_host_id}", retry=3)
+                self.log.debug("Wait while all node update topology status")
+                wait_for(node_operations.is_node_removed_from_cluster, timeout=300, throw_exc=True,
+                         removed_node=self.target_node, verification_node=working_node, text=f"Waite other nodes mark node {self.target_node.name} as REMOVED...")
+                # Run read barrier on working node to validate that it update raft topology state
+                working_node.raft.call_read_barrier()
```
patjed41, line 5229: We don't have execute the read barrier here. You can remove it.

It is true that we must ensure all live nodes apply the Raft entry responsible for banning the removed node before we test that the removed node is actually banned. Read barrier is a mechanism that allows it.

However, in the current implementation of Scylla, we ban the node being removed at the beginning of the removal procedure, so that it doesn't disrupt us if it comes back to live. We commit the Raft entry responsible for banning the node early and wait until all nodes apply it by using a global read barrier.

We didn't implement the Raft-based removenode this way initially. In the old (and incorrect) implemantion, the removed node was banned at the end of the removal procedure and we didn't execute any read barrier after that. With such implementation, executing read barriers on all nodes in this test would be necessary.

patjed41:
```python
wait_for(node_operations.is_node_removed_from_cluster, timeout=300, throw_exc=True,
         removed_node=self.target_node, verification_node=working_node, text=f"Waite other nodes mark node {self.target_node.name} as REMOVED...")
```
is also unnecessary. The test should always pass without it. When `nodetool removenode` finishes, all nodes should have already banned the removed node.

patjed41 (changes requested):

sdcm/nemesis.py
```diff
@@ -5151,6 +5163,95 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
...
+            verify that no operations come from banned node
+            """
+
+            with self.cluster.cql_connection_exclusive(node=node) as session:
+                tablets_config = TabletsConfiguration(
+                    enabled=True) if is_tablets_feature_enabled(session=session) else None
+
+            self.tester.create_keyspace(keyspace_name, replication_factor=3, tablets_config=tablets_config)
+            self.tester.create_table(name=table_name, keyspace_name=keyspace_name, key_type="bigint",
+                                     columns={"name": "text"})
```
roydahan, line 5207: I don't  think this is the correct place for this method to be.

sdcm/nemesis.py
```diff
@@ -3549,34 +3557,36 @@ def disrupt_remove_node_then_add_node(self):  # pylint: disable=too-many-branche
...
+                with adaptive_timeout(Operations.REMOVE_NODE, rnd_node, timeout=HOUR_IN_SEC * 48):
+                    res = rnd_node.run_nodetool("removenode {}".format(host_id), ignore_status=True, verbose=True)
+                if res.failed and re.match(removenode_reject_msg, res.stdout + res.stderr):
+                    raise Exception(f"Removenode was rejected {res.stdout}\n{res.stderr}")
 
             return res.exit_status
 
         # full cluster repair
-        up_normal_nodes.remove(node_to_remove)
+        up_normal_nodes = self.cluster.get_nodes_up_and_normal(verification_node)
```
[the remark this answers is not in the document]
aleksbykov: This change allow to get un nodes in one place.  `stop_scylla_server` check cql port only after `systemctl stop scylla-server` command has finished,  and scylla-server is fully down. Also this code is executed only after node instance will be fully terminated. 
This is the safest way to get all node up and normal for next operations

sdcm/nemesis.py
```diff
@@ -3591,50 +3601,52 @@ def remove_node():
             # WORKAROUND: adding here the continuation of the nemesis to avoid the late filter messages above failing
             # the entire nemesis.
```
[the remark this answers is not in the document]
aleksbykov: removed

sdcm/nemesis.py
```diff
@@ -3591,50 +3601,52 @@ def remove_node():
...
-                        f"Remove failed node {node_to_remove} from dead node list {self.cluster.dead_nodes_list}")
-                    node = next((n for n in self.cluster.dead_nodes_list if n.ip_address ==
-                                node_to_remove.ip_address), None)
-                    if node:
-                        self.cluster.dead_nodes_list.remove(node)
-                    else:
-                        self.log.debug(f"Node {node.name} with ip {node.ip_address} was not found in dead_nodes_list")
-
+                    self.log.debug(f"Node {node.name} with ip {node.ip_address} was not found in dead_nodes_list")
             # verify node is removed by nodetool status
```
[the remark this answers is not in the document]
aleksbykov: fixed

sdcm/provision/scylla_yaml/scylla_yaml.py
```diff
@@ -350,6 +350,7 @@ def set_authorizer(cls, authorizer: str):
     audit_keyspaces: str = None  # None
 
     compaction_collection_items_count_warning_threshold: int = None  # None
+    force_gossip_topology_changes: bool = None  # False
```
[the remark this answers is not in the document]
aleksbykov: removed

unit_tests/test_scylla_yaml.py
```diff
@@ -255,6 +255,7 @@ def test_scylla_yaml(self):
                 'file_cache_size_in_mb': None,
                 'force_gossip_generation': None,
                 'force_schema_commit_log': True,
+                'force_gossip_topology_changes': None,
```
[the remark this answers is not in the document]
aleksbykov: removed

sdcm/utils/raft/__init__.py
```diff
@@ -311,6 +321,32 @@ def is_cluster_topology_consistent(self) -> bool:
 
         return not diff and not non_voters_ids and len(group0_ids) == len(token_ring_ids) == num_of_nodes
 
+    def call_read_barrier(self):
+        """ Verify all raft commits applied on node
```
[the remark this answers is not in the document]
aleksbykov: Fixed

sdcm/nemesis.py
aleksbykov: I removed lines you mentioned, and simple assert to be sure that remove node operation was finished successfully

aleksbykov:
Nemesis are passed themselves. Job s are failed by issue: https://github.com/scylladb/scylladb/issues/22954
 - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h-test/6/
 -  https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h-test/5/

sdcm/nemesis.py
aleksbykov: Remove it and set as steps

aleksbykov:
@roydahan @fruch @soyacz @patjed41 can you review it again. I fixed according your latest comments

sdcm/nemesis.py
```diff
@@ -5403,6 +5409,88 @@ def disrupt_serial_restart_elected_topology_coordinator(self):
...
+        keyspace_name = "banned_keyspace"
+        table_name = "table1"
+
+        def drop_keyspace(node):
+            with self.cluster.cql_connection_patient(node=node) as session:
+                LOGGER.debug("Drop keyspace %s", keyspace_name)
+                session.execute(f"DROP KEYSPACE IF EXISTS {keyspace_name}", timeout=300)
+
+        simulate_node_unavailability = node_operations.block_scylla_ports if use_iptables else node_operations.stop_scylla_with_sigstop
+        with self.run_nemesis(node_list=self.cluster.nodes,
```
soyacz, line 5446: all nodes as target or just data nodes?

soyacz (commented):
few comments/questions

sdcm/nemesis.py
```diff
@@ -5403,6 +5409,88 @@ def disrupt_serial_restart_elected_topology_coordinator(self):
...
+                              nemesis_label=f"Running {simulate_node_unavailability.__name__}") as working_node, ExitStack() as stack:
+            target_host_id = self.target_node.host_id
+            stack.callback(self._remove_node_add_node, verification_node=working_node, node_to_remove=self.target_node,
+                           remove_node_host_id=target_host_id)
+
+            tablets_config = TabletsConfiguration(
```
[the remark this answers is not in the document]
aleksbykov: remove it

sdcm/nemesis.py
aleksbykov: target and working node could be any node in cluster. nodetool status show both nodes types, Cql command also could be run from both node types

sdcm/utils/nemesis_utils/node_operations.py
```diff
@@ -0,0 +1,49 @@
...
+
+from sdcm.cluster import BaseNode
+
+LOGGER = logging.getLogger(__name__)
+
+
+@contextlib.contextmanager
+def block_scylla_ports(target_node: "BaseNode", ports: list[int] | None = None):
+    ports = ports or [7001, 7000, 9042, 9142, 19042, 19142]
+    target_node.install_package("iptables")
```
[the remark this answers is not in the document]
aleksbykov: Added enable iptable service and disable it on exit

soyacz (approved):
LGTM
@aleksbykov please run it again after recent changes (especially with iptables). Make it longer than 2h (4h should be ok) so we can see it can run multiple times.

aleksbykov:
@soyacz ,
 on gce it passed: https://argus.scylladb.com/tests/scylla-cluster-tests/00bd9d62-8be6-4481-aac5-e564e0ea4ae5
 waiting result on aws

patjed41 (approved):
Looks good, you should only remove "experimental" from the PR cover letter.

Some side notes about banning nodes in the Raft-based topology (consistent topology changes) that may be helpful:
- the replaced node is banned from the cluster just like the removed node, so we could consider adding a similar test with replace instead of removenode + add
- not only the removed/replaced node is banned, but also the ignored nodes (other dead nodes that must be specified with `--ignore-dead-nodes` or `--ignore-dead-nodes-for-replace` to successfully remove/replace one dead node), we could consider adding a new test (or extending this test) that makes two nodes unavailable, removes/replaces one of them while ignoring the other, and verifies that both are banned
- removed/replaced/ignored nodes are banned at the beginning of the removenode/replace operation, which means that they become permanently banned even if the operation fails; currently this can be used as a hack to, for example, unblock tablet migrations in the presence of a dead node, but we plan to add a nodetool API to not rely on this hack (see https://github.com/scylladb/scylladb/issues/21281; "mark as permanently dead" and "ban" are two different things, but they happen together in the current implementation); maybe it will be useful in SCT in some other nemesis

aleksbykov:
@soyacz , job is still running https://argus.scylladb.com/tests/scylla-cluster-tests/ddb4ed1d-5148-4683-ad17-9231797d3a24 but nemesis passed 3 times

temichus:
@roydahan, @fruch, This PR has 2 approvals, but it still has requested changes from you. Could you please review this PR ?

sdcm/nemesis.py
```diff
@@ -3549,34 +3557,36 @@ def disrupt_remove_node_then_add_node(self):  # pylint: disable=too-many-branche
...
+                ip_address=node_to_remove.ip_address, verification_node=verification_node)
+            assert removed_node_status is not None, "failed to get host_id using nodetool status"
+            host_id = removed_node_status["host_id"]
+        else:
+            host_id = remove_node_host_id
 
         with ignore_ycsb_connection_refused():
             # node stop and make sure its "DN"
-            node_to_remove.stop_scylla_server(verify_up=True, verify_down=True)
+            node_to_remove.stop_scylla_server(verify_up=False, verify_down=True)
```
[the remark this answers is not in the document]
fruch: I don't know why this was changed, but i's not very critical check since we check status of node between nemesis runs

fruch (approved):
LGTM

PR #6581, opened 2023-09-06
fix(nemesis): init new nodes must set them as seeds

fgelcer:
since #6136 we have been testing our clusters
like all nodes are seeds (since Scylla is now
"seedless" for quite some time), but when running
`disrupt_grow_shrink_cluster` for the 5th time,
we start failing with error:
```
AssertionError: We should have at least one selected seed by now
```

this fix will also be fixing all other nemeses
that do add new nodes to the cluster, when running for quite some time, probability of hitting it on
regular replace node operations is smaller, since
this nemesis runs 3 times add new node in each loop iteration.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1188,6 +1188,8 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         else:
             new_node.replacement_node_ip = old_node_ip
         try:
+            self.cluster.set_seeds()
+            self.cluster.update_seed_provider()
             with adaptive_timeout(Operations.NEW_NODE, node=self.cluster.nodes[0], timeout=timeout):
                 self.cluster.wait_for_init(node_list=[new_node], timeout=timeout, check_node_health=False)
             self.cluster.clean_replacement_node_options(new_node)


fgelcer:
job currently is failing on master --> https://argus.scylladb.com/test/24e3a5d3-9092-4adf-a0a6-24210e792e45/runs?additionalRuns[]=8b6879f7-9202-434e-b086-297bc4f66b25

re-running on a staging job, based on the master one (using custom time) --> https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/seed_node_failure_on_add_new_node/job/grow_shrink_cluster/2/

sdcm/nemesis.py
```diff
@@ -1188,6 +1188,7 @@ def _add_and_init_new_cluster_node(self, old_node_ip=None, host_id=None,
         else:
             new_node.replacement_node_ip = old_node_ip
         try:
+            self.cluster.set_seeds()
```
fruch, line 1191: I think you need to call also:
```                
self.cluster.update_seed_provider()
```

fgelcer: most likely you are right... added

fgelcer:
@fruch , the 1st run (without the call to `update_seed_provider()`) has been successful much longer than the actual master run, meaning, from SCT perspective, it was not necessary, but yes, we must in the end also update the `scylla.yaml`...
the 2nd run is not yet at a point where it fails on master, but IMHO it is ready to be merged, so this nightly run will pass, with this fix in...

@soyacz @vponomaryov @roydahan

fruch (approved):
LGTM

vponomaryov (approved):
LGTM

fgelcer:
seems like the latest code change do what is expected. please backport this one following by #6586

fgelcer:
@scylladb/qa-maintainers ^^

PR #6585, opened 2023-09-07
fix(rolling upgrade): disable audit

juliayakovlev:
Audit was disabled by default in 2023.1 by
https://github.com/scylladb/scylla-enterprise/pull/3094. But it is still enabled in 2022.1 and 2022.2.
It cause to error in cassandra-stress:
```
```

Disable audit in rolling upgrade tests to prevent this failure

### DIFF
--- a/test-cases/upgrades/generic-rolling-upgrade.yaml
+++ b/test-cases/upgrades/generic-rolling-upgrade.yaml
@@ -14,6 +14,9 @@ scylla_linux_distro: 'centos'
 
 user_prefix: 'rolling-upgrade'
 
+append_scylla_yaml: |
+  audit: "none"
+
 server_encrypt: true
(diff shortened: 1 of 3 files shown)


fruch:
@roydahan FYI, 

the idea is todo this only for 2023.1 branch, since the disable isn't gonna be backported into 2022.* branches,
and it generate too much noise for us.

roydahan:
Don't do it.
We need to test what exist in the field and not what makes the test green.

If we want to ignore this error, you can send a PR to do that.

roydahan (changes requested):
Change is not approved

juliayakovlev:
Closed. Opened PR with another solution: https://github.com/scylladb/scylla-cluster-tests/pull/6589

PR #6586, opened 2023-09-07
fix(nemesis): init new node when they are seeds

fgelcer:
following change proposed on #6581, we missed
another longevity, that does replace node,
as the node cannot be a seed, when it is
replacing a node (see error below):
```
[shard 0:main] init - Bad configuration:
 replace-node-first-boot is not allowed for seed nodes
```
moving the `update_seed_provider()` after
we run `clean_replacement_node_options()`,
so it won't conflict.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1188,11 +1188,11 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         else:
             new_node.replacement_node_ip = old_node_ip
         try:
-            self.cluster.set_seeds()
-            self.cluster.update_seed_provider()
             with adaptive_timeout(Operations.NEW_NODE, node=self.cluster.nodes[0], timeout=timeout):
                 self.cluster.wait_for_init(node_list=[new_node], timeout=timeout, check_node_health=False)
             self.cluster.clean_replacement_node_options(new_node)
+            self.cluster.set_seeds()
+            self.cluster.update_seed_provider()
         except (NodeSetupFailed, NodeSetupTimeout):
             self.log.warning("TestConfig of the '%s' failed, removing it from list of nodes" % new_node)
             self.cluster.nodes.remove(new_node)


fgelcer:
must be backported after #6581

fgelcer:
running both [the grow and shrink cluster](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/seed_node_failure_on_add_new_node/job/grow_shrink_cluster/4/) and [the replace node](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/seed_node_failure_on_add_new_node/job/longevity-5gb-1h-NodeTerminateAndReplace-aws-test/2/) nemeses

fruch:
Does that check in scylla core makes any sense now ?

Cause seed is just a contact address now, I think check is just some forgotten

@asias is there still a reason not todo replacement on seed nodes ?

fgelcer:
> Does that check in scylla core makes any sense now ?
> 
> Cause seed is just a contact address now, I think check is just some forgotten
> 
> @asias is there still a reason not todo replacement on seed nodes ?

well, i'm not completly sure about it, but for now, it does fail

fgelcer:
BTW, the terminate and replace issue that was introduced by #6581 is fixed by this commit

fruch (approved):
LGTM

asias:
> Does that check in scylla core makes any sense now ?
> 
> Cause seed is just a contact address now, I think check is just some forgotten
> 
> @asias is there still a reason not todo replacement on seed nodes ?

We can replace a seed node with no problem. Just do not list itself as seed when running the replace ops.

See: https://github.com/scylladb/scylladb/issues/3889

```
commit 399d79fc6f1413f68f88e617386d9c4f54da1889
Author: Amnon Heiman <amnon@scylladb.com>
Date:   Sun Jul 7 10:08:08 2019 +0300

    init: do not allow replace-address for seeds
    
    If a node is a seed node, it can not be started with
    replace-address-first-boot or the replace-address flag.
    
    The issue is that as a seed node it will generate new tokens instead of
    replacing the existing one the user expect it to replaec when supplying
    the flags.
    
    This patch will throw a bad_configuration_error exception
    in this case.
    
    Fixes #3889
    
    Signed-off-by: Amnon Heiman <amnon@scylladb.com>

```

vponomaryov (approved):
LGTM

fgelcer:
pinged to backport on the previous PR

PR #6590, opened 2023-09-11
improvement(event): add AuthenticationError c-s subevent

juliayakovlev:
Add CassandraStressLogEvent.add_subevent_type WARNING for cassandra-stress error:

This error caused by audit enabled. By default audit is disabled in 20223.1 by https://github.com/scylladb/scylla-enterprise/pull/3094. But it won't be enabled in 2022.1 and 2022.2.
This message generates too much noise for us. We do not need it will fail the test.


roydahan:
IIUC you change this error to be warning everywhere and not only during rolling upgrades.
In this case, I think it's too broad and may cause us to miss issues.

juliayakovlev:
> IIUC you change this error to be warning everywhere and not only during rolling upgrades. In this case, I think it's too broad and may cause us to miss issues.

@roydahan 
I found another run (with Scylla version 2022.2):
https://argus.scylladb.com/test/c3dd6458-2baf-4938-a03f-370b6365470c/runs?additionalRuns[]=0ceb3cea-2cc6-46f1-9d48-95e3cde45bb3

Do we want to fail the tests on this error?

sdcm/sct_events/loaders.py
```diff
@@ -215,15 +216,18 @@ def msgfmt(self):
                                           regex="Using optimized driver")
 CassandraStressLogEvent.add_subevent_type("SchemaDisagreement", severity=Severity.WARNING,
                                           regex="No schema agreement")
-
+CassandraStressLogEvent.add_subevent_type("AuthenticationError", severity=Severity.WARNING,
```
fruch, line 219: Let's do the regex a bit more specific to the actual issue i.e. that it's consistency error with cl ONE

And add a comment here, with a reference to the issue

fruch (changes requested):

fruch:
> IIUC you change this error to be warning everywhere and not only during rolling upgrades.
> In this case, I think it's too broad and may cause us to miss issues.

If we do the regex a bit more specific I think it should be good enough

fruch:
@roy if you still insist on a specific case

We might be able to use something like that:

https://github.com/scylladb/scylla-cluster-tests/blob/756feddd2d46006ac88e50e65b086cdee948dc91/performance_regression_alternator_test.py#L25

fruch:
> > IIUC you change this error to be warning everywhere and not only during rolling upgrades. In this case, I think it's too broad and may cause us to miss issues.
> 
> @roydahan 
> I found another run (with Scylla version 2022.2):
> https://argus.scylladb.com/test/c3dd6458-2baf-4938-a03f-370b6365470c/runs?additionalRuns[]=0ceb3cea-2cc6-46f1-9d48-95e3cde45bb3
> 
> Do we want to fail the tests on this error?

Yeah this fix to disable audit wasn't got to 2022.2, so no surprises there.

This PR is only for 2023.1 where is we backported to, and we have issue on on upgrades.

for 2022.2 one might consider this PR (with a bit more specific regex)

data_dir/nemesis.yml
fruch, line 1: can you revert changes to this file ? that are not related to this change

fruch (approved):
LGTM

fruch:
@roydahan 

> IIUC you change this error to be warning everywhere and not only during rolling upgrades. In this case, I think it's too broad and may cause us to miss issues.

Now it's much more specific to the exact error we are talking about with cl=ONE, and it's only goes to 2023.1 (where is generate the noise for us)

@roydahan I think that's specific enough

data_dir/nemesis.yml
juliayakovlev: reverted

roydahan:
It's specific for that error, but the question is if it can happen only due to audit.
Also, let's link this PR in the original issue for this and also in the code write add a comment referring to that issue so we will remember why we do it.

juliayakovlev:
> It's specific for that error, but the question is if it can happen only due to audit. Also, let's link this PR in the original issue for this and also in the code write add a comment referring to that issue so we will remember why we do it.

Added explanation in the code

juliayakovlev:
@roydahan @fruch please, review

fruch (approved):
LGTM

fruch:
@roydahan this is now popping again in 2024.1

https://github.com/scylladb/scylla-enterprise/issues/3707

cause we ignored it only on 2023.1, still 2022.2 is default with audit enabled.
I don't understand why we can't disable it on all releases, since it was a mistake to enable it (without proper testing)

roydahan:
Well, I was in favor of pushing it to all releases but I don't decide.
I don't mind we will backport this PR also to 2024.1

PR #6594, opened 2023-09-11
feature(aws): placement group creation

temichus:
this commit is:
1) Include provision step into the performance pipeline
2) Add the ability to create load and DB instances within one AWS placement group
and config parameter to enable this feature
3) Update i4i Throughput perf test to  use the new feature
4) Update clean_placement_groups_aws functionality
to support --post-behavior flag in sct.py clean-resources command

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -626,6 +626,10 @@ class SCTConfiguration(dict):
         dict(name="security_group_ids", env="SCT_SECURITY_GROUP_IDS", type=str_or_list,
              help="AWS security groups ids to use"),
 
+        dict(name="use_placement_group", env="SCT_USE_PLACEMENT_GROUP", type=str,
+             help="if true, create 'cluster' placement group for test case "
+                  "for low-latency network performance achievement"),
+
         dict(name="subnet_id", env="SCT_SUBNET_ID", type=str_or_list,
              help="AWS subnet ids to use"),
 
@@ -2032,6 +2036,8 @@ class SCTConfiguration(dict):
         self._check_partition_range_with_data_validation_correctness()
         self._verify_scylla_bench_mode_and_workload_parameters()
 
+        self._validate_placement_group_required_values()
+
     def _replace_docker_image_latest_tag(self):
         docker_repo = self.get('docker_image')
         scylla_version = self.get('scylla_version')
@@ -2112,6 +2118,14 @@ class SCTConfiguration(dict):
         for nodes_num in [int(i) for i in str(self.get('n_db_nodes')).split(' ')]:
             assert nodes_num % az_count == 0, \
                 f"Number of db nodes ({nodes_num}) should be divisible by number of availability zones ({az_count})"
+
+    def _validate_placement_group_required_values(self):
+        if self.get("use_placement_group"):
+            az_count = len(self.get('availability_zone').split(',')) if self.get('availability_zone') else 1
+            regions_count = len(self.region_names)
+            assert az_count == 1 and regions_count == 1, \
+                (f"Number of Regions({regions_count}) and AZ({az_count}) should be 1 "
+                 f"when param use_placement_group is used")
 
     def _check_per_backend_required_values(self, backend: str):
         if backend in self.available_backends:
--- a/sdcm/sct_provision/aws/cluster.py
+++ b/sdcm/sct_provision/aws/cluster.py
@@ -15,8 +15,7 @@ import abc
 from functools import cached_property
 from typing import List, Dict
 
-from pydantic import BaseModel  # pylint: disable=no-name-in-module
-
+from pydantic import BaseModel
 from sdcm import cluster
 from sdcm.provision.aws.instance_parameters import AWSInstanceParams
 from sdcm.provision.aws.provisioner import AWSInstanceProvisioner
@@ -29,6 +28,7 @@ from sdcm.sct_provision.aws.instance_parameters_builder import ScyllaInstancePar
 from sdcm.sct_provision.aws.user_data import ScyllaUserDataBuilder, AWSInstanceUserDataBuilder
 from sdcm.sct_provision.common.utils import INSTANCE_PROVISION_SPOT, INSTANCE_PROVISION_SPOT_FLEET
 from sdcm.test_config import TestConfig
+from sdcm.provision.aws.utils import create_cluster_placement_groups_aws
 
 
 class ClusterNode(BaseModel):
@@ -57,6 +57,7 @@ class ClusterBase(BaseModel):
     _NODE_NUM_PARAM_NAME = None
     _INSTANCE_PARAMS_BUILDER = None
     _USER_PARAM = None
+    _USE_PLACEMENT_GROUP = True
 
     @property
     def _provisioner(self):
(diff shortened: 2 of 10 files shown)


sdcm/sct_config.py
```diff
@@ -626,6 +626,9 @@ class SCTConfiguration(dict):
         dict(name="security_group_ids", env="SCT_SECURITY_GROUP_IDS", type=str_or_list,
              help="AWS security groups ids to use"),
 
+        dict(name="use_placement_group", env="SCT_USE_PLACEMENT_GROUP", type=str,
+             help="if true, create placement group for test case for low-latency network performance achievement"),
```
soyacz, line 630: there are different kinds of placement groups, mention the strategy here.

sdcm/sct_config.py
```diff
@@ -2113,6 +2118,14 @@ def _validate_number_of_db_nodes_divides_by_az_number(self):
             assert nodes_num % az_count == 0, \
                 f"Number of db nodes ({nodes_num}) should be divisible by number of availability zones ({az_count})"
 
+    def _validate_placement_group_required_values(self):
+        if self.get("use_placement_group"):
+            az_count = len(self.get('availability_zone').split(',')) if self.get('availability_zone') else 1
+            regions_count = len(self.region_names)
+            assert az_count == 1 and regions_count == 1, \
+                (f"Number of Regions({regions_count}) and AZ({az_count}) should be 1 "
```
soyacz, line 2126: I think we can still support multi-dc - provisioner provisions region by region

sdcm/sct_provision/aws/cluster.py
```diff
@@ -297,4 +309,16 @@ def _user_data(self) -> str:
...
+class PlacementGroup(ClusterBase):
+
+    @property
+    def _user_data(self) -> str:
+        return ''
+
+    def provision(self):
+        if self.placement_group_name:
+            create_cluster_placement_groups_aws(
+                name=self.placement_group_name, tags=self.common_tags, region=self._region(0))
```
soyacz, line 321: for multi-region this would need to iterate over regions

soyacz (commented):
I don't know the reason why perf tests didn't use provisioning as separate step. Maybe @fruch or @fgelcer knows. I'm also not sure if we backported all the provisioning fixes to perf-v14 branch.
Besides, I don't see why we must limit to only single dc, shouldn't be hard to extend to multi (but I'm ok if it is decided not to run ever with placement groups in this cases).
Besides, would be good to see job running (best if used v14 branch as a base, to be 100% sure).

sdcm/sct_config.py
fruch: we are gonna use placement groups for trying to get stable results, and place the whole test on the same place.
multidc case kind of contradicts that, and I think it's o.k. that we don't  support placement groups for multidc.

i.e. we don't need it in the foreseen future, no need to waste time on implementing/testing it.

sdcm/sct_provision/aws/cluster.py
fruch: same as above, we don't need this feature for multidc

sdcm/sct_provision/aws/instance_parameters_builder.py
```diff
@@ -82,7 +83,9 @@ def InstanceType(self) -> str:  # pylint: disable=invalid-name
 
     @property
     def Placement(self) -> Optional[AWSPlacementInfo]:  # pylint: disable=invalid-name
-        return AWSPlacementInfo(AvailabilityZone=self._region_name + self._availability_zones[self.availability_zone])
+        return AWSPlacementInfo(
```
fruch, line 86: what about the case we don't use placement groups ?

temichus:
> > I don't know the reason why perf tests didn't use provisioning as separate step. Maybe @fruch or @fgelcer knows. I'm also not sure if we backported all the provisioning fixes to perf-v14 branch. Besides, I don't see why we must limit to only single dc, shouldn't be hard to extend to multi (but I'm ok if it is decided not to run ever with placement groups in this cases). Besides, would be good to see job running (best if used v14 branch as a base, to be 100% sure).
> 
> there no real reason why perf tests are not using provisioning, someone had to do it and test it, while the perf branches didn't supported it , I think v14 should have what's needed, but it's an effort to achieve
> 
> @temichus is just building ontop of it to focus the work, once someone would add the provisioning step to those tests, he would need to implement this as well.
> 
> It's not the first time we don't implement something in provisioning code, at first.


@fruch @soyacz 

regarding multi-DC - this feature is only for test cases that want to achieve low-latency network performance. multi DC, multi-region, multi-AZ is not applicable(don't need)  for this case 

regarding v14 - if backport will be a problem @roydahan is ok to create v15.

Regarding jobs - in progress now

the main question for now is why I am faced with pylint error: import-error  in places that I'm not touching

fruch:
> the main question for now is why I am faced with pylint error: import-error in places that I'm not touching

if it's locally, it's an environmental issue of some kind
if in CI we need to check

also you touched other files, that might affect import order, and that can cause issues.

sdcm/sct_provision/aws/instance_parameters_builder.py
temichus: it will be 'None' as previous: 
```
class AWSPlacementInfo(BaseModel):
    AvailabilityZone: str
    GroupName: str = None
    Tenancy: Literal['default', 'dedicated', 'host'] = 'default'

```


this case is used  in MonitoringCluster:
```
# disable placement group for monitor nodes, because it doesn't need low-latency network performance
_USE_PLACEMENT_GROUP = False
```

sdcm/sct_config.py
temichus: comment was updated

temichus:
placement group and nodes creation log: 
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/scylla-master-perf-regression-throughput-non-shard-aware-i4i-test/15/execution/node/221/log/

load and DB instances have 
```
{'AvailabilityZone': 'us-east-1a', 'GroupName': 'perf-regression-i4i-ubuntu-placement_group-79d24417'}
```

monitor:
```
'Placement': {'AvailabilityZone': 'us-east-1a'}
```

temichus:
cleaning resources log: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/scylla-master-perf-regression-throughput-non-shard-aware-i4i-test/15/execution/node/472/log/

temichus:
passed job: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/scylla-master-perf-regression-throughput-non-shard-aware-i4i-test/23/

temichus:
> > the main question for now is why I am faced with pylint error: import-error in places that I'm not touching
> 
> if it's locally, it's an environmental issue of some kind if in CI we need to check
> 
> also you touched other files, that might affect import order, and that can cause issues.

fixed by add init file [sdcm/sct_provision/aws/__init__.py](https://github.com/scylladb/scylla-cluster-tests/pull/6594/files#diff-3848adbdfb8d6f641a07d13fddb8a3a1afb13045b743ea52babfe4fd41ddd3b8)

sdcm/sct_provision/aws/instance_parameters_builder.py
fruch: o.k. was confusing a bit

why don't we do the placement for monitor ? would it hurt us to do so ? or just cause it's irrelevant ?

soyacz (approved):
LGTM

fruch (approved):
LGTM

fruch:
backporting to v14, would be done on PR separately by @temichus, when all the pieces would be available

Issue #6591, opened 2023-09-11
When starting a longevity with a scylla_repo installation instead of scylla AMI, scylla is not installed on any node that is added by nemeses


ShlomiBalalis:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: branch-2023.1
* **scylla**: 2023.1

### Logs
* test_id: 86ad51f9-6a75-466b-8a07-da7553a7ac48
* job: https://jenkins.scylladb.com/view/enterprise-2023.1/job/enterprise-2023.1/job/SCT_Enterprise_Features/job/FIPS/job/longevity-100gb-4h-fips-test/3

### Description

This longevity, unlike most, doesn't use a Scylla image, but instead uses a clean Ubuntu image and installs scylla on top of it.
At first, the cluster starts just fine, and the first few nemeses runs are completely fine, but as soon as a nemesis tries to add a new node, SCT does not install scylla on the new node, and thus the nemesis fails:
```
< t:2023-09-10 22:13:25,279 f:remote_base.py  l:520  c:RemoteLibSSH2CmdRunner p:DEBUG > Running command "/usr/bin/scylla --version"...
< t:2023-09-10 22:13:25,780 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > bash: /usr/bin/scylla: No such file or directory
< t:2023-09-10 22:13:25,780 f:cluster.py      l:2122 c:sdcm.cluster_aws     p:DEBUG > Node longevity-fips-2023-1-db-node-86ad51f9-7 [18.202.219.215 | 10.4.2.30] (seed: False): Unable to get ScyllaDB version using `/usr/bin/scylla --version':
< t:2023-09-10 22:13:25,780 f:cluster.py      l:2122 c:sdcm.cluster_aws     p:DEBUG > 
< t:2023-09-10 22:13:25,780 f:cluster.py      l:2122 c:sdcm.cluster_aws     p:DEBUG > bash: /usr/bin/scylla: No such file or directory
< t:2023-09-10 22:13:25,780 f:remote_base.py  l:520  c:RemoteLibSSH2CmdRunner p:DEBUG > Running command "dpkg-query --show --showformat '${Version}' scylla"...
[...]

PR #6596, opened 2023-09-12
fix(Nemesis.add_node): support using `use_preinstalled_scylla: false`

fruch:
since is_replacement_by_host_id_supported is called  before starting a node, 
and in `use_preinstalled_scylla: false` case scylla is not yet installed, 
we should check other running nodes for version

if we don't, we'll fail with a `None` version:

```
ValueError: Cannot parse provided 'None' scylla_version for the comparison.
Transformed scylla_version:
```

Fix: #6591

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1183,7 +1183,11 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             count=1, dc_idx=self.target_node.dc_idx, enable_auto_bootstrap=True, rack=rack)[0]
         self.monitoring_set.reconfigure_scylla_monitoring()
         self.set_current_running_nemesis(node=new_node)  # prevent to run nemesis on new node when running in parallel
-        if new_node.is_replacement_by_host_id_supported:
+
+        # since we need this logic before starting a node, and in `use_preinstalled_scylla: false` case
+        # scylla is not yet installed, we should check the target node for version,
+        # it should be up and with scylla executable available
+        if self.target_node.is_replacement_by_host_id_supported:
             new_node.replacement_host_id = host_id
         else:
             new_node.replacement_node_ip = old_node_ip


sdcm/cluster.py
```diff
@@ -2803,9 +2803,16 @@ def is_cqlsh_support_cloud_bundle(self):
 
     @property
     def is_replacement_by_host_id_supported(self):
+        if self.parent_cluster.params.get("use_preinstalled_scylla"):
+            # since we need this logic before starting a node, and in use_preinstalled_scylla case
+            # scylla is not yet installed, we should check other running nodes for version
+            other_up_node = random.choice([n for n in self.parent_cluster.nodes if n != self])
+            scylla_version = other_up_node.scylla_version
```
vponomaryov, line 2810: The `up` part is not guaranteed here. It may be down caused by a parallel disruptive nemesis...

fruch: I don't need scylla up, I need the node to be up, and have scylla installed.
it's gonna be tricker to guarante all that with parallel nemesis.

maybe the target node of the nemesis would be better to test, regardless, and not the new node, or other nodes

ShlomiBalalis (dismissed):

sdcm/cluster.py
fruch: I've simplified it, and just using the target_node of the nemesis, that node should be installed and ready

sdcm/nemesis.py
```diff
@@ -1183,7 +1183,11 @@ def _add_and_init_new_cluster_node(self, old_node_ip=None, host_id=None,
             count=1, dc_idx=self.target_node.dc_idx, enable_auto_bootstrap=True, rack=rack)[0]
         self.monitoring_set.reconfigure_scylla_monitoring()
         self.set_current_running_nemesis(node=new_node)  # prevent to run nemesis on new node when running in parallel
-        if new_node.is_replacement_by_host_id_supported:
+
+        # since we need this logic before starting a node, and in use_preinstalled_scylla case
```
fgelcer, line 1187: it has a typo... `in `use_preinstalled_scylla: false` case

fgelcer (commented):
when we use scylla `use_preinstalled_scylla: false`, what piece of code is installing Scylla? it should be before we get to this replacement, since we will still need replacement configuration, in case we want to install Scylla, and then replace a terminated node, no?

sdcm/cluster.py

sdcm/nemesis.py
fruch: fixed

vponomaryov (approved):
LGTM

fgelcer (commented):
you fixed, but didn't reply to my question... does it work the code that will create a node that doesn't have pre-installed Scylla, to install Scylla, and then replace a terminated node?

fruch:
> you fixed, but didn't reply to my question... does it work the code that will create a node that doesn't have pre-installed Scylla, to install Scylla, and then replace a terminated node?

It should, and @ShlomiBalalis is the original job found the issue with this change to verify it's fixing it

fgelcer:
> > you fixed, but didn't reply to my question... does it work the code that will create a node that doesn't have pre-installed Scylla, to install Scylla, and then replace a terminated node?
> 
> It should, and @ShlomiBalalis is the original job found the issue with this change to verify it's fixing it

if @ShlomiBalalis confirms it works in his use case, then it LGTM

fruch:
https://jenkins.scylladb.com/job/scylla-staging/job/Shlomo/job/longevity-100gb-4h-fips-test/13/

Seems like there are more issues past this point

Failing to install some packages - I'm guessing a race between cloud init and node setup. we have tools to avoid it, need to apply them.

And stress is failing cause it looking for scylla.yaml of node where it doesn't yet exist, we should have a proper fallback in the there, small fix.

fgelcer (approved):
LGTM

fgelcer:
@ShlomiBalalis , once merged (and backported to 2023.1), do a new run for 2023.1.1, and fix the issues there are after the

PR #6601, opened 2023-09-13
fix(wait for node init): check native_transport

fgelcer:
following issue investigation on grow and shrink
cluster individual nemesis, we observed that
the nodes being added, are taking much longer
than what we expected (we check for `nodetool status` being `UN` before we continue), but in this case,
nodes were not yet fully serving I/O, making
eventually the cluster to lose quorum, stopping
the c-s workload with `NoHostAvailableException`.

Refs: scylladb/scylladb#15335


fgelcer:
gave it a run [here](https://argus.scylladb.com/test/d59e6009-371e-4957-bb3a-1d9e70c6ba3f/runs?additionalRuns[]=fe2eaeab-a5a4-45a3-a6fd-d6af96086f2b) and although the run is not yet over, we are not seeing the effect of nodes with much lower load (because they are part of the cluster, but are not fully serving I/O), as you can see [here](https://snapshots.raintank.io/dashboard/snapshot/UMZ4fpXKtxes72ZiQZVImFc7fsb4ZijJ)

let's wait for the run to finish, as recently on master, they are failing consistently

fgelcer:
live monitor (until the test end) --> http://3.238.2.27:3000/d/db22ac9e-239f-47a8-8cf3-97ad3ac2e3ac/grow-shrink-cluster-native-transport-scylla-per-server-metrics-nemesis-master?from=1694621090230&to=now&var-by=instance&var-cluster=&var-dc=All&var-node=All&var-shard=All&var-mount_point=%2Fvar%2Flib%2Fscylla&var-sg=statement&var-func=sum&var-dash_version=master&var-all_scyllas_versions=All&var-count_dc=All&var-scylla_version=master&var-monitoring_version=4.5.0&orgId=1

sdcm/cluster.py
```diff
@@ -4307,6 +4307,16 @@ def _wait_for_nodes_up_and_normal():
 
         _wait_for_nodes_up_and_normal()
 
+    @retrying(n=360, sleep_time=10, allowed_exceptions=ClusterNodesNotReady, message="Waiting for CQLPing to return")
+    def wait_native_transport(self, node):
```
fruch, line 4311: Not sure I know what this API means...

But why not implement it on BaseNode ?

fgelcer: in one hand, this is part of the node service itself, but in other hand, this is related to how that node is serving the cluster...
if you think this is necessary, i can move it to the BaseNode

fgelcer: about "not knowing what this API means", we could use instead some `node.run_cqlsh(<some query>)` instead

sdcm/nemesis.py
```diff
@@ -1199,6 +1199,7 @@ def _add_and_init_new_cluster_node(self, old_node_ip=None, host_id=None,
             self.log.warning("Node will not be terminated. Please terminate manually!!!")
             raise
         self.cluster.wait_for_nodes_up_and_normal(nodes=[new_node])
+        self.cluster.wait_native_transport(node=new_node)
```
soyacz, line 1202: need to verify it works also on k8s

sdcm/cluster.py
soyacz, line 4311: is there a difference from this and `wait_db_up` where we wait for cql port used?

sdcm/nemesis.py
fruch: noting unique here, it's a remote command, and assumes curl is installed (which i'm quite sure is in scylla docker images)

sdcm/cluster.py
```diff
@@ -4307,6 +4307,16 @@ def _wait_for_nodes_up_and_normal():
 
         _wait_for_nodes_up_and_normal()
 
+    @retrying(n=360, sleep_time=10, allowed_exceptions=ClusterNodesNotReady, message="Waiting for CQLPing to return")
```
fruch, line 4310: I wouldn't use the term CQLPing, since it's not an actual something anywhere.

sdcm/cluster.py
fruch: also what about using the newly introduced `node.wait_node_fully_start()`

we introduced it for restarts, but the case here is similar..

fgelcer: AFAIU, `node.wait_node_fully_start()` by itself won't be enough, as the new node while joining the cluster, will have both jmx and Scylla services running, will report in `nodetool status` it is `UN`, but it won't yet be servicing full I/O.

it could be related to lucky (either good or bad) that the code I run, on a test that is consistently failing on master, now passed with the change in this PR.

about the idea to move it elsewhere, like inside `BaseNode`, i will do the change now and push it to re-review this patch

soyacz: @vponomaryov do you think it could help in operator to wait for node being fully up?

fruch: I have a better explanation what's going on here:

we don't set the `verification_node` on the `self.cluster.wait_for_nodes_up_and_normal(nodes=[new_node])` call
hence it randomly pick the node to do nodetool status on, and it can be new_node, and why you ask it, it would say UN, while other node won't yet see it as UP, this we for sure.

so calling, `node.wait_node_fully_start()` should do the same trick as what you introduced in this PR 

I would also recommend fixing `wait_for_nodes_up_and_normal` to handle this situation better, and not pick randomly the `verification_node`,  if it's single node to be waited, it should use the same one for verification

I'm guessing since you are running this specific nemesis with 3 node growshrink at a time, this situation just happens more.

fruch: also `_add_new_node_in_new_dc` should be handle the same as we do in PR (how ever we fix it)

fgelcer: @fruch moved to the `BaseNode`

fgelcer: @soyacz @fruch , made changes, and now i'm using `wait_node_fully_start()`, and the new function `wait_native_transport()` is now part of it

sdcm/cluster.py
fgelcer: re-named

sdcm/cluster.py
fgelcer: > I have a better explanation what's going on here:
> 
> we don't set the `verification_node` on the `self.cluster.wait_for_nodes_up_and_normal(nodes=[new_node])` call hence it randomly pick the node to do nodetool status on, and it can be new_node, and why you ask it, it would say UN, while other node won't yet see it as UP, this we for sure.
> 
> so calling, `node.wait_node_fully_start()` should do the same trick as what you introduced in this PR
> 
> I would also recommend fixing `wait_for_nodes_up_and_normal` to handle this situation better, and not pick randomly the `verification_node`, if it's single node to be waited, it should use the same one for verification
> 
> I'm guessing since you are running this specific nemesis with 3 node growshrink at a time, this situation just happens more.

i think that is not exactly what happens, because as it was said multiple times, a node being `UN` doesn't mean much... there is also a RFE to add a better way to query if the node is fully serving, or not --> https://github.com/scylladb/scylladb/issues/8275

fgelcer: also, added `_add_new_node_in_new_dc()` to use the check `new_node.wait_node_fully_start()`

sdcm/cluster.py
```diff
@@ -187,6 +188,10 @@ class ScyllaRequirementError(Exception):
     pass
 
 
+class NodeNotReady(Exception):
```
vponomaryov, line 191: We have special module for exceptions here: https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/exceptions.py

sdcm/cluster.py
vponomaryov: > @vponomaryov do you think it could help in operator to wait for node being fully up?

If operator developers do the same there? If it helps in general, it may help.
But I think they would like to see official and recommended way for getting this info.
In current case it looks like workaround for the absent proper way of checking node readiness.

sdcm/utils/sstable/load_utils.py
```diff
@@ -121,9 +122,8 @@ def run_load_and_stream(cls, node,  # pylint: disable=too-many-arguments
 
             # `load_and_stream` parameter is not supported by nodetool yet. This is workaround
             # https://github.com/scylladb/scylla-tools-java/issues/253
-            load_api_cmd = 'curl -X POST --header "Content-Type: application/json" --header ' \
-                           f'"Accept: application/json" "http://127.0.0.1:10000/storage_service/sstables/{keyspace_name}?' \
-                           f'cf={table_name}&load_and_stream=true"'
+            path = f'/storage_service/sstables/{keyspace_name}?cf={table_name}&load_and_stream=true'
+            load_api_cmd = build_node_api_command(path_url=path, request_method=RequestMethods.POST)
```
vponomaryov, line 126: Before the curl command here was without `-s` flag. Now it will have.
Is it intentional?

fgelcer: `-s` is silent mode..... perhaps this function will need it, but @juliayakovlev may know that better...
in any case, i could perhaps remove the `-s` from all occurances, as we could "afford" to have more details in the logs...
WDYT?

sdcm/cluster.py
fgelcer: correct, @vponomaryov

sdcm/cluster.py
fgelcer: moved it there

sdcm/utils/sstable/load_utils.py
vponomaryov: @fgelcer I don't know how useful the output is. I would like to see opinion from @juliayakovlev

fgelcer:
ping @fruch @soyacz @vponomaryov @roydahan

fruch (commented):
LGTM

vponomaryov (approved):
LGTM

fgelcer:
> LGTM

can you please approve/merge?

fruch (approved):
LGTM

fgelcer:
BTW, do we want to backport it? @roydahan @fruch ? WDYT?

roydahan:
If works well on master, we can backport to 2023.1 next week.

PR #6083, opened 2023-04-30
feature(nemesis): Added manager restore nemesis

ShlomiBalalis:
Since manager 3.1, manager now supports restoring previously created backups. In this PR I added a new manager restore nemesis, that restores data from previously created snapshots (seen in manager_persistent_snapshots.yaml)

## PR pre-checks (self review)

- [x] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [x] I didn't leave commented-out/debugging code
- [x] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6603, opened 2023-09-15
improvement(nemesis): increase mgmt restore timeout for encrypted cases

vponomaryov:
Usage of the encryption slow downs Scylla operations including the `scylla manager restore` one.
So, add adaptive timeout context manager for the `restore` operation
to tolerate slow process but allow the operation to be finished in a reasonable time frame.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2792,13 +2792,22 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
         restore_task = mgr_cluster.create_restore_task(restore_data=True,
                                                        location_list=location_list, snapshot_tag=chosen_snapshot_tag)
-        restore_task.wait_and_get_final_status(step=30, timeout=chosen_snapshot_info["expected_timeout"])
+        if self.cluster.params.get('server_encrypt'):
+            soft_mgmt_restore_timeout = int(chosen_snapshot_info["expected_timeout"] * 1.2)
+            hard_mgmt_restore_timeout = int(chosen_snapshot_info["expected_timeout"] * 2)
+        else:
+            soft_mgmt_restore_timeout = chosen_snapshot_info["expected_timeout"]
+            hard_mgmt_restore_timeout = int(chosen_snapshot_info["expected_timeout"] * 1.1)
+        with adaptive_timeout(Operations.SOFT_TIMEOUT, self.cluster.nodes[0], timeout=soft_mgmt_restore_timeout):
+            restore_task.wait_and_get_final_status(step=30, timeout=hard_mgmt_restore_timeout)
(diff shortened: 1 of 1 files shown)


vponomaryov:
It will need to be backported to branches where the base commit for it gets backported: https://github.com/scylladb/scylla-cluster-tests/pull/6083

fruch:
@ShlomiBalalis can you take a look at it ?

roydahan:
We need a Scylla issue or at least something to be added to release note
describing the slowdown and by how much.

vponomaryov:
> We need a Scylla issue or at least something to be added to release note describing the slowdown and by how much.

Perf results: https://docs.google.com/spreadsheets/d/1SIc1I3I4EbNgCIATv7oPmXdqxQi69SwdiC08WIJ4Ya4/edit#gid=0

Or in short:
`latency` diff:
- write: ~2.7ms (KMS) vs ~2.6ms = -4%
- read:  ~2.8ms (KMS) vs ~2.5ms = -12%

`thrpt` diff:
- write: 214098 op/s / 368783 op/s = 0.58
- read:  292514 op/s / 376659 op/s = 0.77

fgelcer (approved):
LGTM

roydahan:
I didn't mean throughput and latency (although it may be interesting for something else, as long as it's measured with the correct job).

I meant the ability backup/restore time (and throughput of it) while using KMS.

vponomaryov:
> I didn't mean throughput and latency (although it may be interesting for something else, as long as it's measured with the correct job).
> 
> I meant the ability backup/restore time (and throughput of it) while using KMS.

Reference to the perf results means `we process less data in the same time using encryption`.
And `scylla manager restore` operation is directly related to the `data processing`.
So, this operation must be slower when we use encryption.
The question is how big is the time margin for the snapshots we have.

I am ok to leave the original timeouts as the `soft timeouts`, but keep `hard timeouts be bigger` just to let the operation finish in cases it is just timing question.
Keep in mind that the exceeding the `soft timeout` will cause creation of the ERROR SCT event.
So, we will catch it.

fruch:
It's the first time we tested EaR with backup,

Writing sstable is expected to slower in EaR, no surprises there.

It's just highlight that there are aspect related to back we didn't plan to test.

Now we did.

Who exactly do you want to "sign" those numbers ? scylla-manager developers, core developers, product ?

roydahan:
Product, or in other words @tzach in a organized way.

vponomaryov:
@roydahan what do we do with this?

roydahan:
We need to ask Product management, what are the expectations and check that we meet the expectations.
If backup and/or restore are slower by X%, the need to approve it before we set the tests not to fail for it.

vponomaryov:
No interest on it for long time. Closing.

PR #6607, opened 2023-09-17
feature(cluster): Add support refactored setup scripts and config files

syuu1228:
On https://github.com/scylladb/scylladb/pull/14674, we are trying to refactor setup scripts and config files.

Since it big change, it will  break compatibility with existing scripts. To avoid Jenkins test failure when we merge the PR, we need to modify following SCT code:
 - /etc/scylla.d/cpuset.conf and /etc/scylla.d/io.conf are replaced by /etc/scylla.d/seastar.conf
 - scylla_sysconfig_setup renamed to scylla_perftune_setup

Also, to support both older version of Scylla and new version of Scylla with same SCT code, we need to detect which version of Scylla is running.
To detect that, this patch adds BaseNode.has_seastar_conf property, detects Scylla version and access different setup scripts / config files.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -570,13 +570,19 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
             # or
             #   'CPUSET="--cpuset 1-7,9-15,17-23,25-31 "'
             # And so on...
-            cpuset_file_lines = self.remoter.run("cat /etc/scylla.d/cpuset.conf").stdout
+
+            if self.has_seastar_conf:
+                cpuset_file_lines = self.remoter.run("cat /etc/scylla.d/seastar.conf").stdout
+                cpuset_startswith = "cpuset="
+            else:
+                cpuset_file_lines = self.remoter.run("cat /etc/scylla.d/seastar.conf").stdout
+                cpuset_startswith = "CPUSET="
         except Exception as exc:  # pylint: disable=broad-except
             self.log.error(f"Failed to get CPUSET. Error: {exc}")
             return ''
 
         for cpuset_file_line in cpuset_file_lines.split("\n"):
-            if not cpuset_file_line.startswith("CPUSET="):
+            if not cpuset_file_line.startswith(cpuset_startswith):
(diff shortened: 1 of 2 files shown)


syuu1228:
On this patch I used newly added config file to detect "refactored version", but there are another way to do that, such as from Scylla version or adding a file to detect the mode like $SCYLLA_DIR/SCYLLA-NONROOT-FILE.

If you think current way is not best way, please comment that.

fruch:
@syuu1228 
please fix the pre-commit issues, I recommend installing pre-commit locally 

also one of the unittest is failing cause of this change, `unit_tests/test_cluster.py::test_base_node_cpuset` please fix

sdcm/cluster.py
```diff
@@ -797,6 +803,11 @@ def is_enterprise(self):
             return "scylla-enterprise" in result or "No matching items found" not in result
         return "scylla-enterprise" in self.remoter.sudo("apt-cache search scylla-enterprise", ignore_status=True).stdout
 
+    @cached_property
```
fruch, line 806: on rolling upgrade we might need to clear this cache, or just leave it without cache.

fruch:
@syuu1228 

please share jobs with artifacts tests running this (both with new scylla configuration, and the old one)
we also need to test it with rolling upgrades, to see we are not mixing it, or caching between versions

fruch:
> On this patch I used newly added config file to detect "refactored version", but there are another way to do that, such as from Scylla version or adding a file to detect the mode like $SCYLLA_DIR/SCYLLA-NONROOT-FILE.
> 
> If you think current way is not best way, please comment that.

if we have some installer version, it might be more clear way to detection, but assume some specific file exists.
but both should work.

also this change might effect on CCM or dtest ?

fruch:
@yarongilor @syuu1228 is this PR still needed ?

yarongilor:
> @yarongilor @syuu1228 is this PR still needed ?

@yaronkaikov ?

fruch:
if someone gonna go back to https://github.com/scylladb/scylladb/pull/14674

he could reopen this one

PR #6610, opened 2023-09-18
fix(node utils): health check started failing

fgelcer:
since #6601 was introduced, because a function
that checks for token rings started failing
on the `curl` command builder.
this commit aims to fix it.

### DIFF
--- a/sdcm/utils/node.py
+++ b/sdcm/utils/node.py
@@ -21,9 +21,10 @@ class RequestMethods(Enum):
     POST = 'POST'
 
 
-def build_node_api_command(path_url, request_method: RequestMethods = RequestMethods.GET, api_port=10000):
+def build_node_api_command(path_url, request_method: RequestMethods = RequestMethods.GET, api_port=10000, silent=True):
     if not path_url.startswith('/'):
         path_url = '/' + path_url
+    silent_flag = '-s ' if silent else ''
 
-    return f'curl -X {request_method} --header "Content-Type: application/json" --header ' \
+    return f'curl -X {silent_flag}{request_method.value} --header "Content-Type: application/json" --header ' \
            f'"Accept: application/json" "http://127.0.0.1:{api_port}{path_url}"'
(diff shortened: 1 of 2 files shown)


fgelcer:
https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/native-transport-issue/job/grow_shrink_cluster_native_transport/4/

fgelcer:
@roydahan @fruch @soyacz @vponomaryov , we better to merge it before the daily jobs are triggered

sdcm/cluster.py
```diff
@@ -3108,7 +3108,7 @@ def reload_config(self):
     def get_token_ring_members(self) -> list[dict[str, str]]:
         self.log.debug("Get token ring members")
         token_ring_members = []
-        token_ring_members_cmd = build_node_api_command('/storage_service/host_id')
+        token_ring_members_cmd = build_node_api_command('/storage_service/host_id', silent=False)
```
fruch, line 3111: @fgelcer this seems backwards then what you are trying to fix, this command needs to silent

Anyhow I would prefer adding a parameter of extra arguments, and start mapping all of curl command options

fruch (changes requested):

sdcm/cluster.py
fgelcer: did i disable silence on the wrong command? i will fix it, but the whole mapping of the CURL command, should be done incrementally IMHO, as i don't think we will use them all

fgelcer: @fruch , returned this one to be silent, and remove what is not supposed to be silent

fgelcer:
[here](https://argus.scylladb.com/test/d59e6009-371e-4957-bb3a-1d9e70c6ba3f/runs?additionalRuns[]=0e282481-9593-446e-85a3-c6ab7529771c) we can see that the issue was fixed

roydahan (approved):

PR #6611, opened 2023-09-18
fix(test_default): stop keeping db nodes

fgelcer:
since most of the time, we don't have
enough people investigating failures,
and in many cases, we have issues being
reproduced, and instances are being left
alive for long time.
in any case, when we want to reproduce an
issue, we manually set the test to `keep`
so changing it to `destroy` won't impact
us too much..

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -48,10 +48,10 @@ seeds_num: 1
 instance_provision: "spot"
 
 execute_post_behavior: false
-post_behavior_db_nodes: "keep-on-failure"
+post_behavior_db_nodes: "destroy"
 post_behavior_loader_nodes: "destroy"
 post_behavior_monitor_nodes: "keep-on-failure"
-post_behavior_k8s_cluster: "keep-on-failure"
+post_behavior_k8s_cluster: "destroy"
 
 cloud_credentials_path: ''
 use_cloud_manager: false
--- a/unit_tests/provisioner/test_azure_region_definition_builder.py
+++ b/unit_tests/provisioner/test_azure_region_definition_builder.py
@@ -52,7 +52,8 @@ def test_can_create_basic_scylla_instance_definition_from_sct_config():
 
     instance_definition = InstanceDefinition(name=f"{prefix}-db-node-eastus-1", image_id=env_config.SCT_AZURE_IMAGE_DB,
                                              type="Standard_L8s_v3", user_name="scyllaadm", root_disk_size=30,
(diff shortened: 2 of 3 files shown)


defaults/test_default.yaml
```diff
@@ -48,7 +48,7 @@ seeds_num: 1
 instance_provision: "spot"
 
 execute_post_behavior: false
-post_behavior_db_nodes: "keep-on-failure"
+post_behavior_db_nodes: "destroy"
```
fruch, line 51: I'm not sure this is enough, we need to change the longevity pipeline defaults

fgelcer: i think it doesn't have defaults

fgelcer: @fruch , fixed

fruch: I think both post_behavior_monitor_nodes, post_behavior_k8s_cluster, should be changed as well

fgelcer: done

defaults/test_default.yaml
```diff
@@ -48,10 +48,10 @@ seeds_num: 1
...
 post_behavior_loader_nodes: "destroy"
 post_behavior_monitor_nodes: "keep-on-failure"
```
vponomaryov, line 53: Do we want to keep monitoring node too?

fruch: The default here is a bit meaningless, since of the default on the pipelines.

It wouldn't affect local runs cause cleanup is a separate task (and people don't usually call it as in the pipe, so all the keep on failure logic doesn't kick-in)

Would be nice to align it with the pipeline, not more then that

roydahan: I don't want to change it, not here and not in the pipelines.
Only in some triggered pipelines where we clearly approve it.

fgelcer: i don't think we should keep DB nodes by default, as no one is using them, and they are costing money without any reason... if someone wants to reproduce an issue, and intentionally keep them alive, this is something else, as the person will be owner of these instances, other than `timer` or `qa`, when automatic trigger is run

roydahan (changes requested):
Not approved.

amoskong:
In the past, SCT left the instances in failed job (with critical event, etc).
And another Jenkins periodic job will try to release them according to the keep/alive lable in aws/gce.

If you change the default behavior from keep-on-failure to destroy, the instances will be destoried even the job has critical event.
I think you can't do this for jobs of stable release.

roydahan:
@amoskong great to hear from you :)

We know, but we reached to a point when we hardly need these instances.

PR #6616, opened 2023-09-19
feature(edquot): Add quota configuration and EndOfQuotaNemesis

Mark-Gurevich:
According to task: https://github.com/scylladb/qa-tasks/issues/1086

This commit adds the possibility to configure quota on a node and adds the EndOfQuotaNemesis
The EndOfQuotaNemesis configures quota with current disk usage + 1GB and waits for `Disk quota exceeded` message in logs. Then the quota is cleared and scylla-server service is restarted.

### DIFF
--- a/data_dir/nemesis.yml
+++ b/data_dir/nemesis.yml
@@ -332,3 +332,6 @@
   - disruptive = False
   - kubernetes = True
   - free_tier_set = True
+- disrupt_end_of_quota_nemesis:
+  - disruptive = True
+  - kubernetes = False
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -82,7 +82,7 @@ from sdcm.sct_events.group_common_events import (ignore_alternator_client_errors
                                                  ignore_scrub_invalid_errors, ignore_view_error_gate_closed_exception,
                                                  ignore_stream_mutation_fragments_errors,
                                                  ignore_ycsb_connection_refused, decorate_with_context,
-                                                 ignore_reactor_stall_errors,
+                                                 ignore_reactor_stall_errors, ignore_disk_quota_exceeded_errors,
                                                  ignore_error_apply_view_update)
 from sdcm.sct_events.health import DataValidatorEvent
 from sdcm.sct_events.loaders import CassandraStressLogEvent, ScyllaBenchEvent
@@ -97,6 +97,8 @@ from sdcm.utils.common import (get_db_tables, generate_random_string,
                                parse_nodetool_listsnapshots,
                                update_authenticator, ParallelObject,
                                ParallelObjectResult, sleep_for_percent_of_duration, get_views_of_base_table)
+from sdcm.utils.quota import configure_quota_on_node_for_scylla_user_context, is_quota_enabled_on_node, enable_quota_on_node, \
+    write_data_to_reach_end_of_quota
 from sdcm.utils.compaction_ops import CompactionOps, StartStopCompactionArgs
 from sdcm.utils.context_managers import nodetool_context
 from sdcm.utils.decorators import retrying, latency_calculator_decorator
@@ -137,6 +139,7 @@ from sdcm.exceptions import (
     NemesisSubTestFailure,
     AuditLogTestFailure,
     BootstrapStreamErrorFailure,
+    QuotaConfigurationFailure,
 )
 from test_lib.compaction import CompactionStrategy, get_compaction_strategy, get_compaction_random_additional_params, \
     get_gc_mode, GcMode
@@ -1682,6 +1685,42 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                         reach_enospc_on_node(target_node=node)
                     finally:
                         clean_enospc_on_node(target_node=node, sleep_time=sleep_time)
+
+    def disrupt_end_of_quota_nemesis(self, sleep_time=30):
+        """
+        Nemesis flow
+        ---------------------
+        1. Enable quota on the node.
+        2. Check disk usage.
+        3. Define quota size same as disk usage + 3GB.
+        4. Approach end of quota in a loop with fallocate file that takes 90% of quota size each iteration.
+        5. Wait for end of quota message to appear in the log - "Disk quota exceeded" / Except I/O Error if happens.
+        6. Remove the sparse files.
+        7. Remove the quota by setting it to 0.
+        8. Restart scylla server.
+        9. Verify scylla is up.
+        """
+
+        node = self.target_node
+        if self._is_it_on_kubernetes():
+            raise UnsupportedNemesis("Skipping nemesis for kubernetes")
+
+        result = node.remoter.run('cat /proc/mounts')
+        if '/var/lib/scylla' not in result.stdout:
+            raise UnsupportedNemesis("Scylla doesn't use an individual storage, skip end of quota test")
+
+        enable_quota_on_node(node)
+        quota_enabled = is_quota_enabled_on_node(node)
+        if not quota_enabled:
+            raise QuotaConfigurationFailure("Failed to configure quota on the cluster")
+
+        with ignore_disk_quota_exceeded_errors(node):
+            with configure_quota_on_node_for_scylla_user_context(node) as quota_size:
+                write_data_to_reach_end_of_quota(node, quota_size)
+            LOGGER.debug('Sleep 15 seconds before restart scylla-server')
+            time.sleep(15)
+            node.restart_scylla_server()
+            node.wait_db_up()
 
     def disrupt_remove_service_level_while_load(self):
         # Temporary solution. We do not want to run SLA nemeses during not-SLA test until the feature is stable
@@ -6485,3 +6524,11 @@ class DisableBinaryGossipExecuteMajorCompaction(Nemesis):
 
     def disrupt(self):
         self.disrupt_disable_binary_gossip_execute_major_compaction()
+
+
+class EndOfQuotaNemesis(Nemesis):
+    disruptive = True
+    config_changes = True
+
+    def disrupt(self):
+        self.disrupt_end_of_quota_nemesis()
--- a/sdcm/utils/quota.py
+++ b/sdcm/utils/quota.py
@@ -1,0 +1,128 @@
+from __future__ import absolute_import, annotations
+
+import logging
+import time
+from contextlib import contextmanager
+
+from sdcm.exceptions import QuotaConfigurationFailure
+from sdcm.wait import wait_for
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def enable_quota_on_node(node):
+    if not is_quota_enabled_on_node(node):
+        LOGGER.info('Enabling quota on node {}'.format(node.name))
+        change_default_grub_cmd = 'sed -i s\'/GRUB_CMDLINE_LINUX_DEFAULT="quiet splash"/' \
+                                  'GRUB_CMDLINE_LINUX_DEFAULT="quiet splash rootflags=uquota,pquota"/\' ' \
+                                  '/etc/default/grub'
+        mk_grub_cmd = 'grub-mkconfig -o /boot/grub/grub.cfg'
+        enable_uquota_on_mount_cmd = 'sed -i s\'/Options=noatime,discard/Options=noatime,discard,uquota/\' ' \
+                                     '/etc/systemd/system/var-lib-scylla.mount'
+
+        node.remoter.sudo(cmd=change_default_grub_cmd)
+        node.remoter.sudo(cmd=mk_grub_cmd)
+        node.remoter.sudo(cmd=enable_uquota_on_mount_cmd)
+        LOGGER.debug('Rebooting node: "{}"'.format(node.name))
+        node.reboot(hard=False)
+        node.wait_node_fully_start()
+
+
+def is_quota_enabled_on_node(node):
+    """ Verify usrquota is enabled on scylla user. """
+    LOGGER.info("Verifying quota is configured on the node {}".format(node.name))
+    verify_usrquot_in_mount_cmd = "mount | grep /var/lib/scylla | awk {'print $6'} | cut -d ',' -f 8"
+    result = node.remoter.run(cmd=verify_usrquot_in_mount_cmd).stdout.rstrip().replace(")", "")
+    if not result == "usrquota":
+        LOGGER.warning("User quota on user scylla at /var/lib/scylla is not enabled")
+        return False
+    return True
+
+
+def get_currently_used_space_by_scylla_user(node) -> int:
+    result = node.remoter.sudo("xfs_quota -x -c 'report' /var/lib/scylla | grep scylla | tail -1")
+    used_space = int(result.stdout.split()[1])
+    return used_space
+
+
+def get_quota_size_to_configure(node):
+    used_space = get_currently_used_space_by_scylla_user(node)
+    LOGGER.debug('Currently used space by scylla user is: {}K'.format(used_space))
+    return used_space + 3000000
+
+
+@contextmanager
+def configure_quota_on_node_for_scylla_user_context(node):
+    """
+    Configure quota with current disk usage size by scylla user + 3GB on /var/lib/scylla mount point
+    Verify quota was configured
+    yield the quota size
+    Remove quota configuration
+    """
+    quota_size = get_quota_size_to_configure(node)
+    LOGGER.info("Configuring quota for scylla user with quota size: {}K on node {}".format(quota_size, node))
+    conf_quota_cmd = f"xfs_quota -x -c 'limit bsoft={quota_size}K bhard={quota_size}K scylla' /var/lib/scylla"
+    node.remoter.sudo(conf_quota_cmd, verbose=True)
+    verify_quota_is_configured_for_scylla_user(node)
+    try:
+        yield quota_size
+    finally:
+        remove_quota_configuration_from_scylla_user_on_node(node)
+
+
+def verify_quota_is_configured_for_scylla_user(node):
+    """Verifies quota ia actually configured for scylla user, but doesn't verify enforcement in ON"""
+    get_scylla_user_quota_value = 'xfs_quota -x -c "report -h" /var/lib/scylla | grep scylla | tail -1 | awk {\'print $4\'}'
+    LOGGER.info("Verifying quota is configured for scylla user")
+    quota_value = node.remoter.sudo(get_scylla_user_quota_value).stdout.rstrip()
+    if quota_value == "0":
+        LOGGER.error("Failed to configure quota on user")
+        raise QuotaConfigurationFailure(
+            "Quota was not configured for scylla user. The value for hard quota is: {}".format(quota_value))
+    LOGGER.info("Quota was configured for scylla user with value: {}".format(quota_value))
+
+
+def remove_quota_configuration_from_scylla_user_on_node(node):
+    LOGGER.info('Remove quota configuration from user \'scylla\' by setting it to 0')
+    quota_remove_cmd = "xfs_quota -x -c 'limit bsoft=0 bhard=0 scylla' /var/lib/scylla"
+    node.remoter.sudo(quota_remove_cmd, verbose=True)
+
+
+def write_data_to_reach_end_of_quota(node, quota_size):
+    LOGGER.debug('Watching the log for the relevant message for quota...')
+    quota_exceeded_appearances = node.follow_system_log(
+        patterns=[r'Shutting down communications due to I\/O errors until operator intervention: Disk error:.*Disk quota exceeded'])
+    node.remoter.sudo('chsh -s /bin/bash scylla')  # Enable login for scylla user
+
+    def approach_end_of_quota():
+        if bool(list(quota_exceeded_appearances)):
+            return True
+        currently_used_space = get_currently_used_space_by_scylla_user(node)
+        occupy_space_size = int((quota_size - currently_used_space) * 90 / 100)
+        occupy_space_cmd = f"su - scylla -c 'fallocate -l {occupy_space_size}K /var/lib/scylla/occupy_90percent.{time.time()}'"
+        try:
+            LOGGER.debug('Cost 90% free space on /var/lib/scylla/ by {}'.format(occupy_space_cmd))
+            node.remoter.sudo(occupy_space_cmd, Verbose=True)
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.warning("We should have reached the expected I/O error and quota has exceeded\n"
+                           "Message: {}".format(str(exc)))
+        return bool(list(quota_exceeded_appearances))
+
+    try:
+        wait_for(func=approach_end_of_quota,
+                 timeout=300,
+                 step=10,
+                 text="Waiting for 'Disk error: std::system_error (error system:122, Disk quota exceeded)' "
+                 "string to appear in the node log",
+                 throw_exc=False
+                 )
+        LOGGER.debug('Sleep 5 minutes before releasing space to scylla')
+        time.sleep(300)
+
+    finally:
+        LOGGER.debug('Delete occupy_90percent file to release space to scylla-server')
+        node.remoter.sudo('rm -rf /var/lib/scylla/occupy_90percent.*')  # Remove the sparse files
+
+        LOGGER.debug('Disable scylla user login')
+        node.remoter.sudo('usermod -s /sbin/nologin scylla')  # Disable login for scylla user
--- a/unit_tests/test_nemesis_sisyphus.py
+++ b/unit_tests/test_nemesis_sisyphus.py
@@ -71,7 +71,7 @@ def test_list_all_available_nemesis(generate_file=True):
     disruption_list, disruptions_dict, disruption_classes = sisyphus.get_list_of_disrupt_methods(
         subclasses_list=subclasses, export_properties=True)
 
-    assert len(disruption_list) == 84
+    assert len(disruption_list) == 85
 
     if generate_file:
         with open('data_dir/nemesis.yml', 'w', encoding="utf-8") as outfile1:
--- a/jenkins-pipelines/nemesis/longevity-5gb-1h-EndOfQuotaNemesis-aws.jenkinsfile
+++ b/jenkins-pipelines/nemesis/longevity-5gb-1h-EndOfQuotaNemesis-aws.jenkinsfile
@@ -1,0 +1,12 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    backend: 'aws',
+    region: 'eu-west-1',
+    test_name: 'longevity_test.LongevityTest.test_custom_time',
+    test_config: 'test-cases/nemesis/longevity-5gb-1h-EndOfQuotaNemesis.yaml'
+
+)
--- a/sdcm/exceptions.py
+++ b/sdcm/exceptions.py
@@ -65,6 +65,10 @@ class KillNemesis(BaseException):
     """Exception that would be raised, when a nemesis thread is killed at teardown of the test"""
 
 
+class QuotaConfigurationFailure(Exception):
+    """Exception that would be raised, if quota configuration failed"""
+
+
 class WaitForTimeoutError(Exception):
     """Exception that would be raised timeout exceeded in wait.wait_for function"""
 
--- a/sdcm/sct_events/group_common_events.py
+++ b/sdcm/sct_events/group_common_events.py
@@ -134,6 +134,32 @@ def ignore_no_space_errors(node):
 
 
 @contextmanager
+def ignore_disk_quota_exceeded_errors(node):
+    with ExitStack() as stack:
+        stack.enter_context(DbEventsFilter(
+            db_event=DatabaseLogEvent.BACKTRACE,
+            line="Disk quota exceeded",
+            node=node,
+        ))
+        stack.enter_context(DbEventsFilter(
+            db_event=DatabaseLogEvent.FILESYSTEM_ERROR,
+            line="Disk quota exceeded",
+            node=node,
+        ))
+        stack.enter_context(DbEventsFilter(
+            db_event=DatabaseLogEvent.DATABASE_ERROR,
+            line="Disk quota exceeded",
+            node=node,
+        ))
+        stack.enter_context(DbEventsFilter(
+            db_event=DatabaseLogEvent.DISK_ERROR,
+            line="Disk quota exceeded",
+            node=node,
+        ))
+        yield
+
+
+@contextmanager
 def ignore_mutation_write_errors():
     with ExitStack() as stack:
         stack.enter_context(EventsSeverityChangerFilter(
--- a/test-cases/nemesis/longevity-5gb-1h-EndOfQuotaNemesis.yaml
+++ b/test-cases/nemesis/longevity-5gb-1h-EndOfQuotaNemesis.yaml
@@ -1,0 +1,31 @@
+test_duration: 90
+
+prepare_write_cmd:  "cassandra-stress write cl=QUORUM n=5048570 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=80 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5"
+
+stress_cmd: ["cassandra-stress write cl=QUORUM duration=60m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=40 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5",
+             "cassandra-stress read  cl=QUORUM duration=60m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=40 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5"
+             ]
+
+n_db_nodes: 3
+n_loaders: 1
+n_monitor_nodes: 1
+seeds_num: 3
+
+instance_type_db: 'i4i.large'
+gce_instance_type_db: 'n1-highmem-2'
+gce_instance_type_loader: 'e2-standard-2'
+azure_instance_type_db: 'Standard_L8s_v3'
+instance_type_loader: 'c5.large'
+azure_instance_type_loader: 'Standard_F2s_v2'
+
+nemesis_class_name: 'EndOfQuotaNemesis'
+nemesis_interval: 3
+nemesis_filter_seeds: false
+nemesis_during_prepare: true
+
+gce_n_local_ssd_disk_db: 1
+
+user_prefix: 'longevity-5gb-1h-EndOfQuotaNemesis'
+
+server_encrypt: true
+client_encrypt: true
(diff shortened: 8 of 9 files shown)


sdcm/nemesis.py
```diff
@@ -1661,6 +1662,52 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+                result = self.target_node.remoter.run('cat /proc/mounts')
+                if '/var/lib/scylla' not in result.stdout:
+                    LOGGER.error("Scylla doesn't use an individual storage, skip edquota test")
+                    continue
+
+                try:
+                    configure_quota_on_node(target_node=node)
+                    wait.wait_for_log_lines(node=node, start_line_patterns='Disk quota exceeded',
+                                            end_line_patterns='Will retry in 3200ms',
+                                            start_timeout=60, end_timeout=600)
```
fruch, line 1703: The time 10min and size 1gb, are highly dependent on the test case, in some it would take more then 10min to fill 1gb.

Anyhow it's not clear what the end message you are looking for, it's a bit vague, and not clear why it's needed.

sdcm/nemesis.py
```diff
@@ -6409,3 +6456,11 @@ class DisableBinaryGossipExecuteMajorCompaction(Nemesis):
 
     def disrupt(self):
         self.disrupt_disable_binary_gossip_execute_major_compaction()
+
+
+class EdquotNemesis(Nemesis):
```
fruch, line 6461: Name of the class should be similar to the name of the distrupt function
Edquot is clear or known enough, from my POV.

The function name is clearer

Mark-Gurevich: I haven't fully got you. Should I change the `disrupt_end_of_quota_nemesis` to `disrupt_edquot_nemesis` you mean ?

sdcm/nemesis.py
Mark-Gurevich: Anyway, the `wait_for_log_lines` didn't work at all. 
In the last run it jumped right away to the next method `clean_edquota_on_node` so something it wrong here.

fruch: I believe it's a context manger, so you are using it the wrong way

Mark-Gurevich: Yep, I saw that already. Thank you for the input.

Mark-Gurevich:
Do I need to set the new `enable_quota` to `false` anywhere in the `defaults` for example in `test_default.yaml` ?

fruch:
> Do I need to set the new `enable_quota` to `false` anywhere in the `defaults` for example in `test_default.yaml` ?

yes, it's best if any newly introduced parameter has a default in test_default.yaml, or in the different backend defaults files.
there should be any defaults set in the code, default should be in once place (or per backend)

sdcm/nemesis.py
fruch: I think the other way around `EndOfQoutaNemesis` has more meaning to me then `EdquotNemesis` sounds like a typo at first look.

yeah, I know we have `EnospcNemesis`, but that name was like that before I joined scylla...
`NoSpaceNemesis` would have been more readable (especially for newcomers)

Mark-Gurevich: I totally agree with you. 
I've used the `EDQUOT` because that's how it is described [here](https://github.com/scylladb/qa-tasks/issues/1086) and [here](https://github.com/scylladb/scylladb/pull/12653)
I guess its a known term in Linux.
Whatever you decide, no problem to change this tiny thing.

Mark-Gurevich:
> > Do I need to set the new `enable_quota` to `false` anywhere in the `defaults` for example in `test_default.yaml` ?
> 
> yes, it's best if any newly introduced parameter has a default in test_default.yaml, or in the different backend defaults files. there should be any defaults set in the code, default should be in once place (or per backend)

As I expected. Added to test_default.yaml `enable_quota: false`

sdcm/nemesis.py
```diff
@@ -1661,6 +1662,63 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+
+        if all_nodes:
+            nodes = self.cluster.nodes
+            InfoEvent('EDQUOT test on {}'.format([n.name for n in nodes])).publish()
+        else:
+            nodes = [self.target_node]
+
+        for node in nodes:
+            if self._is_it_on_kubernetes():
+                LOGGER.info("Skipping nemesis for kubernetes")
```
soyacz, line 1690: raise UnsupportedNemesis

sdcm/nemesis.py
```diff
@@ -1661,6 +1662,63 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+                break
+            node.verify_quota_is_enabled_on_node()
+            with ignore_disk_quota_exceeded_errors(node=node):
+                result = self.target_node.remoter.run('cat /proc/mounts')
+                if '/var/lib/scylla' not in result.stdout:
+                    LOGGER.error("Scylla doesn't use an individual storage, skip edquota test")
```
soyacz, line 1696: raise UnsupportedNemesis

sdcm/sct_config.py
```diff
@@ -1484,6 +1484,9 @@ class SCTConfiguration(dict):
         dict(name="use_dns_names", env="SCT_USE_DNS_NAMES", type=boolean,
              help="""Use dns names instead of ip addresses for nodes in cluster"""),
 
+        dict(name="enable_quota", env="SCT_ENABLE_QUOTA", type=bool,
+             help="Enable possibility of quota configuration on the cluster during setup")
```
soyacz, line 1488: why it is required to enable like this?

sdcm/nemesis.py
```diff
@@ -1661,6 +1662,63 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+                                     'Create keyspace1.standard1 if it does not exist')
+                        self._prepare_test_table(ks='keyspace1', table='standard1')
+                        assert wait_for(
+                            func=lambda: list(quota_exceeded),  # pylint: disable=cell-var-from-loop
+                            timeout=1200,
+                            text="Waiting for 'Disk quota exceeded' string to appear in the node log ",
+                            throw_exc=False,
+                        ), "Quota has not been reached"
+                    else:
+                        LOGGER.error("Failed to configure quota on node: {}".format(node))
```
soyacz, line 1715: raise UnsupportedNemesis or other error if appropriate

sdcm/utils/common.py
```diff
@@ -2756,6 +2756,54 @@ def clean_enospc_on_node(target_node, sleep_time):
...
+        conf_quota_cmd = f"xfs_quota -x -c 'limit bsoft={quota_size}K bhard={quota_size}K scylla' /var/lib/scylla"
+        try:
+            target_node.remoter.sudo(conf_quota_cmd, verbose=True, )
+            return verify_quota_is_configured_for_scylla_user(target_node=target_node)
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.error("Quota configuration command has failed: {}".format(str(exc)))
+            return False
+
+
+def clean_edquota_on_node(target_node, sleep_time):
```
soyacz, line 2793: why do we need `sleep_time`?

sdcm/nemesis.py
```diff
@@ -1661,6 +1662,63 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+            nodes = self.cluster.nodes
+            InfoEvent('EDQUOT test on {}'.format([n.name for n in nodes])).publish()
+        else:
+            nodes = [self.target_node]
+
+        for node in nodes:
+            if self._is_it_on_kubernetes():
+                LOGGER.info("Skipping nemesis for kubernetes")
+                break
+            node.verify_quota_is_enabled_on_node()
```
soyacz, line 1692: `verify_quota_is_configured_for_scylla_user` and `configure_quota_on_node` use target node as argument, why here is different?

sdcm/nemesis.py
```diff
@@ -1661,6 +1662,63 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+
+                try:
+                    LOGGER.info("Trying to configure quota on node: {}".format(node))
+                    success = configure_quota_on_node(target_node=node)
+                    if success:
+                        LOGGER.debug('Watching the log for "Disk quota exceeded"')
+                        quota_exceeded = node.follow_system_log(patterns=['Disk quota exceeded'])
+                        LOGGER.debug('Run additional load with prepare_test_table to reach end of quota. '
+                                     'Create keyspace1.standard1 if it does not exist')
+                        self._prepare_test_table(ks='keyspace1', table='standard1')
```
soyacz, line 1707: this will not fill disk anyway if table already exists

sdcm/utils/common.py
```diff
@@ -2756,6 +2756,54 @@ def clean_enospc_on_node(target_node, sleep_time):
     target_node.wait_db_up()
 
 
+def verify_quota_is_configured_for_scylla_user(target_node) -> bool:
```
soyacz, line 2759: put it to separate file rather than `common.py` - it is not that common (maybe smthing like `xfs_quota.py`?)

soyacz:
I wonder if creating 1GB file with dd and changing ownership to `scylla` would do the job instead creating kesypace, table and populating it...

fgelcer:
> I wonder if creating 1GB file with dd and changing ownership to `scylla` would do the job instead creating kesypace, table and populating it...

well, it would, but as we should not login as Scylla user, we won't use the right user quota, and won't test the behavior on Scylla's side

soyacz:
> > I wonder if creating 1GB file with dd and changing ownership to `scylla` would do the job instead creating kesypace, table and populating it...
> 
> well, it would, but as we should not login as Scylla user, we won't use the right user quota, and won't test the behavior on Scylla's side

we don't have to login as Scylla, just change file ownership.

sdcm/nemesis.py
Mark-Gurevich: ok

sdcm/nemesis.py
Mark-Gurevich: Changed. Why we don't raise in enospc nemesis in the same place then ?

sdcm/nemesis.py
Mark-Gurevich: Why we have methods in `BaseScyllaCluster` and `BaseNode` classes and other methods are in sdcm/utils/common.py? 
`verify_quota_is_enabled_on_node` is related to node setup and verify the possibility of adding quota configuration (enabled). So, I added it to `BaseNode` class.
Maybe I should change the name for `configure_quota_on_node` to `configure_quota_on_node_for_scylla_user`.
Anyway, both these methods (`verify_quota_is_configured_for_scylla_user` and `configure_quota_on_node`) are in the utils/common.py and require a parameter as target_node.
That's the logic of SCT which is not very clear to me why there is a division between `BaseNode` and methods in `common.py` and other files in `sdcm/utils`.

sdcm/utils/common.py
Mark-Gurevich: I can put it to separate file but why we don't create a separate file for enospc methods then?

sdcm/sct_config.py
Mark-Gurevich: Actually we can enable quota on node during the nemesis and restart the node.
@fgelcer asked me to make it this way, so when it is enabled on the whole cluster we can use the nemesis or `configure_quota_on_node_for_scylla_user` separately.

roydahan (changes requested):

sdcm/nemesis.py
```diff
@@ -1665,6 +1666,77 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
                         clean_enospc_on_node(target_node=node, sleep_time=sleep_time)
 
+    def write_data_with_cs_to_reach_end_of_quota(self, node):
+        keyspace_quota = 'ks_quota'
+        try:
+            stress_cmd = "cassandra-stress write no-warmup n=200000 cl=QUORUM -mode native cql3 " \
+                         f"-schema 'replication(strategy=NetworkTopologyStrategy," \
+                         f"replication_factor={self.tester.reliable_replication_factor})' -log interval=5"
+            self.tester.run_stress_thread(
+                stress_cmd=stress_cmd, keyspace_name=keyspace_quota, stop_test_on_failure=False, round_robin=True)
```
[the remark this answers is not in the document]
roydahan: .

sdcm/sct_config.py
roydahan: All this is not needed.

Mark-Gurevich:
https://jenkins.scylladb.com/job/scylla-staging/job/mark/job/longevity-5gb-1h-EndOfQuotaNemesis-aws/15/

Mark-Gurevich:
> https://jenkins.scylladb.com/job/scylla-staging/job/mark/job/longevity-5gb-1h-EndOfQuotaNemesis-aws/15/

The test is working. I checked the logs and verified it has actually configured quota, got the expected message, went down and restarted Scylla service.

sdcm/nemesis.py
```diff
@@ -1688,6 +1691,76 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
                     finally:
                         clean_enospc_on_node(target_node=node, sleep_time=sleep_time)
 
+    def write_data_with_cs_to_reach_end_of_quota(self, node):
```
roydahan, line 1694: I already commented long ago that this is not needed and what is need to be done is to write a sparse file as we do in ENOSPC nemesis.

roydahan: Not only it's not needed, it adds a load that some of the tests may not expect.

sdcm/nemesis.py
```diff
@@ -1665,6 +1666,77 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+                         f"-schema 'replication(strategy=NetworkTopologyStrategy," \
+                         f"replication_factor={self.tester.reliable_replication_factor})' -log interval=5"
+            self.tester.run_stress_thread(
+                stress_cmd=stress_cmd, keyspace_name=keyspace_quota, stop_test_on_failure=False, round_robin=True)
+        except IOError as exc:
+            LOGGER.info("We should have reached the expected I/O error and quota has exceeded\n Message: {}".format(str(exc)))
+        finally:
+            LOGGER.debug("Dropping keyspace: %s", keyspace_quota)
+            with self.cluster.cql_connection_patient(node=self.target_node) as session:
+                session.execute(f""" DROP KEYSPACE IF EXISTS {keyspace_quota} """)
```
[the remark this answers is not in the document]
roydahan: related to my previous comment.... while this is a month old comment.

sdcm/utils/quota.py
```diff
@@ -0,0 +1,70 @@
...
+                                 '/etc/systemd/system/var-lib-scylla.mount'
+
+    node.remoter.sudo(cmd=change_default_grub_cmd)
+    node.remoter.sudo(cmd=mk_grub_cmd)
+    node.remoter.sudo(cmd=enable_uquota_on_mount_cmd)
+    LOGGER.debug('Rebooting node: "{}"'.format(node.name))
+    node.reboot(hard=False)
+    LOGGER.debug("node reboot takes very long time and next methods fails. "
+                 "Adding sleep for 2 minutes here before wait_node_fully_start")
+    time.sleep(120)
```
roydahan, line 56: Shouldn't be like that, at the maximum you should do "wait_for" and increase timeout.

unit_tests/test_nemesis_sisyphus.py
```diff
@@ -71,7 +71,7 @@ def test_list_all_available_nemesis(generate_file=True):
     disruption_list, disruptions_dict, disruption_classes = sisyphus.get_list_of_disrupt_methods(
         subclasses_list=subclasses, export_properties=True)
 
-    assert len(disruption_list) == 84
+    assert len(disruption_list) == 85
```
roydahan, line 74: Missing output of nemesis.yaml with the new nemesis.

sdcm/nemesis.py
```diff
@@ -1688,6 +1691,76 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+            LOGGER.info("Trying to configure quota on node: {}".format(node))
+            configure_quota_on_node_for_scylla_user(target_node=node)
+            verify_quota_is_configured_for_scylla_user(target_node=node)
+
+            LOGGER.debug('Run additional load with write_data_with_cs_to_reach_end_of_quota')
+            self.write_data_with_cs_to_reach_end_of_quota(node)
+
+            remove_quota_configuration_from_node(target_node=node)
+            LOGGER.debug('Sleep a while before restart scylla-server')
+            time.sleep(sleep_time / 2)
```
roydahan, line 1760: Ugly and I don't understand why needed.

roydahan (changes requested):

data_dir/nemesis.yml
```diff
@@ -332,3 +332,7 @@
   - disruptive = False
   - kubernetes = True
   - free_tier_set = True
+- disrupt_end_of_quota_nemesis:
+  - disruptive = True
+  - kubernetes = False
+  - free_tier_set = True
```
roydahan, line 338: I think this one isn't relevant for this nemesis.

sdcm/utils/quota.py
```diff
@@ -0,0 +1,76 @@
+from __future__ import absolute_import, annotations
+
+import time
+import logging
+from sdcm.exceptions import QuotaConfigurationFailure
+
+LOGGER = logging.getLogger(__name__)
+
+
+def configure_quota_on_node_for_scylla_user(target_node):
```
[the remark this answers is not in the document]
roydahan: He wasn't able to make it work for directory quota, so the workaround was to use a user quota.

sdcm/nemesis.py
roydahan: Still relevant.

Mark-Gurevich:
I hope this can be finally merged.
https://argus.scylladb.com/workspace?state=WyI4NWJiY2UxZC1iYmVhLTQ3ZGItOThkNS01NzExOWM3ZjQyNWIiXQ
https://jenkins.scylladb.com/job/scylla-staging/job/mark/job/longevity-5gb-1h-EndOfQuotaNemesis-aws/23/

sdcm/nemesis.py
```diff
@@ -1688,6 +1692,42 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+            raise UnsupportedNemesis("Scylla doesn't use an individual storage, skip end of quota test")
+
+        enable_quota_on_node(node)
+        quota_enabled = verify_quota_is_enabled_on_node(node)
+        if not quota_enabled:
+            raise QuotaConfigurationFailure("Failed to configure quota on the cluster")
+
+        with ignore_disk_quota_exceeded_errors(node=node):
+            quota_size = configure_quota_on_node_for_scylla_user(node=node)
+            verify_quota_is_configured_for_scylla_user(node=node)
```
soyacz, line 1725: if this fails, then quota on node will persist.
One proposal is to use try..catch or better to make `configure_quota_on_node_for_scylla_user` to be context manager which removes quota upon `__exit__`

sdcm/utils/quota.py
```diff
@@ -0,0 +1,113 @@
...
+    LOGGER.info("Verifying quota is configured for scylla user")
+    quota_value = node.remoter.sudo(get_scylla_user_quota_value).stdout.rstrip()
+    if quota_value == "0":
+        LOGGER.error("Failed to configure quota on user")
+        raise QuotaConfigurationFailure(
+            "Quota was not configured for scylla user. The value for hard quota is: {}".format(quota_value))
+    LOGGER.info("Quota was configured for scylla user with value: {}".format(quota_value))
+
+
+def remove_quota_configuration_from_node(node):
```
soyacz, line 74: this doesn't remove quota configuration from node, rather from user. Fix confusing name.

sdcm/utils/quota.py
```diff
@@ -0,0 +1,113 @@
...
+
+        node.remoter.sudo(cmd=change_default_grub_cmd)
+        node.remoter.sudo(cmd=mk_grub_cmd)
+        node.remoter.sudo(cmd=enable_uquota_on_mount_cmd)
+        LOGGER.debug('Rebooting node: "{}"'.format(node.name))
+        node.reboot(hard=False)
+        node.wait_node_fully_start()
+
+
+def verify_quota_is_enabled_on_node(node):
```
soyacz, line 33: `verify_` in the name points to verification -> should not return anything (like it doesn't `verify_quota_is_configured_for_scylla_user`)
To keep consistency in naming, use `is_quota_enabled_on_node` for returning quota status

sdcm/utils/quota.py
```diff
@@ -0,0 +1,113 @@
...
+    node.remoter.sudo('chsh -s /bin/bash scylla')  # Enable login for scylla user
+
+    def approach_end_of_quota():
+        if bool(list(quota_exceeded_appearances)):
+            return True
+        currently_used_space = get_currently_used_space_by_scylla_user(node)
+        occupy_space_size = int((quota_size - currently_used_space) * 90 / 100)
+        occupy_space_cmd = f"su - scylla -c 'fallocate -l {occupy_space_size}K /var/lib/scylla/occupy_90percent.{time.time()}'"
+        with EventsSeverityChangerFilter(new_severity=Severity.NORMAL,  # End of quota created CRITICAL event
+                                         event_class=CassandraStressEvent,
```
soyacz, line 93: how about other load tools like s-b?
Also, shouln't be done as a part of the `ignore_disk_quota_exceeded_errors` so in case c-s fails in between we don't create critical?

sdcm/utils/quota.py
```diff
@@ -0,0 +1,113 @@
...
+    node.remoter.sudo(quota_remove_cmd, verbose=True)
+
+
+def write_data_to_reach_end_of_quota(node, quota_size):
+    LOGGER.debug('Watching the log for the relevant message for quota...')
+    quota_exceeded_appearances = node.follow_system_log(
+        patterns=[r'Shutting down communications due to I\/O errors until operator intervention: Disk error:.*Disk quota exceeded'])
+    node.remoter.sudo('chsh -s /bin/bash scylla')  # Enable login for scylla user
+
+    def approach_end_of_quota():
```
soyacz, line 86: why not just creating file that occupies 100%?

soyacz (commented):
Still some places to fix, but overall, looks good.
Generally good structure and namings, makes it easy to follow and review!

sdcm/utils/quota.py
fruch: why would c-s gonna fail ? or have critical events in this case ? i.e. its one node going down.

c-s need to continue without any issues, if not, we should open a bug, I don't think we do this during regular ENSPC nemesis

sdcm/utils/quota.py
```diff
@@ -0,0 +1,128 @@
...
+        if bool(list(quota_exceeded_appearances)):
+            return True
+        currently_used_space = get_currently_used_space_by_scylla_user(node)
+        occupy_space_size = int((quota_size - currently_used_space) * 90 / 100)
+        occupy_space_cmd = f"su - scylla -c 'fallocate -l {occupy_space_size}K /var/lib/scylla/occupy_90percent.{time.time()}'"
+        # with EventsSeverityChangerFilter(new_severity=Severity.NORMAL,  # End of quota created CRITICAL event
```
fruch, line 107: can we remove this comment ?

sdcm/utils/quota.py
```diff
@@ -0,0 +1,124 @@
+from __future__ import absolute_import, annotations
+
+import logging
+import time
+from contextlib import contextmanager
+
+from sdcm.exceptions import QuotaConfigurationFailure
+# from sdcm.sct_events import Severity
+# from sdcm.sct_events.filters import EventsSeverityChangerFilter
+# from sdcm.sct_events.loaders import CassandraStressEvent
```
soyacz, line 10: remove these comments

soyacz (dismissed):
LGTM

Mark-Gurevich:
There is still a problem. If we wait for db to go down after  'Quota has exceeded' then, the test becomes flaky. It may fail as there is no space to flush memtables.
`[shard 1:main] table - Memtable flush failed due to: storage_io_error (Storage I/O error: 122: Disk quota exceeded). Will retry in 10000ms`
I've opened an [issue](https://github.com/scylladb/scylladb/issues/15908) on this, but Yaniv says it's expected behaviour. So the only option is to clean the sparse files immediately after we reach end of quota.
@roydahan Please advice.

image

Maybe we can try to ignore this part:
```
2023-10-31T12:39:15.188+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3      !ERR | scylla[582]:  [shard 1:main] table - failed to write sstable /var/lib/scylla/data/keyspace1/standard1-86364b8077e711eebea6fe9ae48a3174/me-3gao_0z5d_4ke7k2nn0ue3opyjr7-big-Data.db: storage_io>
2023-10-31T12:39:15.188+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3  !WARNING | scylla[582]:  [shard 1:main] table - Memtable flush failed due to: storage_io_error (Storage I/O error: 122: Disk quota exceeded). Will retry in 10000ms
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3  !WARNING | systemd[1]: scylla-server.service: State 'stop-sigterm' timed out. Killing.
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3   !NOTICE | systemd[1]: scylla-server.service: Killing process 582 (scylla) with signal SIGABRT.
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]: Aborting on shard 0.
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]: Backtrace:
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5a9efd8
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5ad5212
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   /opt/scylladb/libreloc/libc.so.6+0x3dbaf
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   /opt/scylladb/libreloc/libc.so.6+0x10ab5c
2023-10-31T12:39:22.938+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5afea40
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5afa7b2
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5afaedc
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5ab1eac
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5ab0f29
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5a51ff7
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x5a511ac
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x130e16a
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x130fbc0
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x130c6fc
2023-10-31T12:39:22.939+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   /opt/scylladb/libreloc/libc.so.6+0x27b89
2023-10-31T12:39:22.940+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   /opt/scylladb/libreloc/libc.so.6+0x27c4a
2023-10-31T12:39:22.940+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | scylla[582]:   0x130a124
2023-10-31T12:39:22.940+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | systemd[1]: Created slice Slice /system/systemd-coredump.
2023-10-31T12:39:22.940+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3     !INFO | systemd[1]: Started Process Core Dump (PID 1915/UID 0).
2023-10-31T12:39:22.940+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3  !WARNING | kernel: xfs filesystem being remounted at /run/systemd/unit-root/var/lib/systemd/coredump supports timestamps until 2038 (0x7fffffff)
2023-10-31T12:39:22.940+00:00 longevity-5gb-1h-EndOfQuotaNemesis--db-node-e21f0c0a-3  !WARNING | kernel: xfs filesystem being remounted at /run/systemd/unit-root/var/lib/systemd/coredump supports timestamps until 2038 (0x7fffffff)
```

It creates 3 Errors:
1. (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=666f8c6e-5ead-4ef4-9e36-598ac1b2858d: type=ABORTING_ON_SHARD regex=Aborting on shard 
2.  (CoreDumpEvent Severity.ERROR)
3.  (ClusterHealthValidatorEvent Severity.CRITICAL)

**I suggest just to remove the sparse files immediately after end of quota reached and only then wait for db to go down.**


Even after trying to remove the sparse files the service still fails to go down, and nemesis fails. 

image

Mark-Gurevich:
Ready for merge
https://jenkins.scylladb.com/job/scylla-staging/job/mark/job/longevity-5gb-1h-EndOfQuotaNemesis-aws/30/

roydahan:
Please see my comment here: https://github.com/scylladb/scylladb/issues/15908#issuecomment-1793862000

sdcm/utils/quota.py
```diff
@@ -0,0 +1,120 @@
...
+        return bool(list(quota_exceeded_appearances))
+
+    wait_for(func=approach_end_of_quota,
+             timeout=300,
+             step=10,
+             text="Waiting for 'Disk error: std::system_error (error system:122, Disk quota exceeded)' "
+             "string to appear in the node log",
+             throw_exc=False
+             )
+    node.remoter.sudo('rm -rf /var/lib/scylla/occupy_90percent.*')  # Remove the sparse files
```
fruch, line 119: where is the code that clears this file ?

sdcm/nemesis.py
```diff
@@ -1688,6 +1691,40 @@ def disrupt_nodetool_enospc(self, sleep_time=30, all_nodes=False):
...
+
+        enable_quota_on_node(node)
+        quota_enabled = is_quota_enabled_on_node(node)
+        if not quota_enabled:
+            raise QuotaConfigurationFailure("Failed to configure quota on the cluster")
+
+        with ignore_disk_quota_exceeded_errors(node):
+            with configure_quota_on_node_for_scylla_user_context(node) as quota_size:
+                write_data_to_reach_end_of_quota(node, quota_size)
+            node.restart_scylla_server()
```
fruch, line 1725: this is happening too quickly, you should align to the flow done in `clean_enospc_on_node`
it's first give something like 30sec for the scylla to shut down on it's own,
clears the temporary file
wait more 15sec
and then restarts scylla

if you are doing stop/restart on you own before scylla shut itself down.

you are hitting the systemd timeout, and core dumps scylla.

please stick to the same exact flow of ENOSPC, scylla behavior needs to be identical (from scylla code POV)
you trying total different flow here, is kind of making all kind of assumptions which aren't exactly accurate

roydahan: Even longer than 30sec, we should make sure cluster react to the node failure.

Mark-Gurevich:
This is ready to be merged - passed in all variation with and without wait.
@scylladb/qa-maintainers

roydahan:
1. There are still some comments that aren't resolved.

2. We should merge it with enough sleep time during the problem (5 minutes), till the node is down.
IIUC, there is an issue with this - https://github.com/scylladb/scylladb/issues/15908
So if we merge this it may cause issues.

3. Let's send a followup PR to ENOSPC nemesis to also wait 5 minutes during the ENOSPC.

Mark-Gurevich:
> 1. There are still some comments that aren't resolved.

Which comments do you mean?
> 2. We should merge it with enough sleep time during the problem (5 minutes), till the node is down.
>    IIUC, there is an issue with this - [scylla server service shut down can't complete with `disrupt_end_of_quota_nemesis` running - due to lack of disk space to flush memtables scylladb#15908](https://github.com/scylladb/scylladb/issues/15908)
>    So if we merge this it may cause issues.

I've run tests with 5 and 10 minutes wait (runs 33,34) and they passed.

Israel wrote:
>if you are doing stop/restart on you own before scylla shut itself down.
>you are hitting the systemd timeout, and core dumps scylla.

The reason for the failure was probably trying to `stop_scylla_server` manually after it was already stopping when reached end of quota. 

> 3. Let's send a followup PR to ENOSPC nemesis to also wait 5 minutes during the ENOSPC.

Do you want me to change in the ENOSPC just the wait time to exact 5 minutes( why 5 ?)
Isn't it better to put some condition instead of relying on specific time?
Anyway, if we send a PR for ENOSPC with 5 minutes I should probably change to 5 minutes in this PR as well.
Right now it is 60 seconds.
@roydahan

sdcm/utils/quota.py
fruch: or in other word, we need to make sure we always execute this part, this should be in one finally clause of some context manager, to make sure we clear those

fruch (dismissed):
LGTM

Waiting for handling those two:
- https://github.com/scylladb/scylla-cluster-tests/pull/6616#discussion_r1413484132
- stable 5min to let scylla go down, as @roydahan requested

Mark-Gurevich:
> LGTM
> 
> Waiting for handling those two:
> 
> * [feature(edquot): Add quota configuration and EndOfQuotaNemesis #6616 (comment)](https://github.com/scylladb/scylla-cluster-tests/pull/6616#discussion_r1413484132)

Added try - finally block
> * stable 5min to let scylla go down, as @roydahan requested

Changed to 5 mintues wait.

@fruch  - Test passed: 
https://jenkins.scylladb.com/job/scylla-staging/job/mark/job/longevity-5gb-1h-EndOfQuotaNemesis-aws/

Please merge.

soyacz (approved):
lgtm

fruch (approved):

PR #6617, opened 2023-09-20
fix(random region): add region resolution to getJobTimeouts

fgelcer:
since we separated the function that gets the
timeout for each stage, and the random region
resolution was not adapted to it, this commit
aims to fix that.

### DIFF
--- a/vars/getJobTimeouts.groovy
+++ b/vars/getJobTimeouts.groovy
@@ -2,12 +2,13 @@
 
 List<Integer> call(Map params, String region){
     // handle params which can be a json list
+    def current_region = initAwsRegionParam(params.region, region)
     def test_config = groovy.json.JsonOutput.toJson(params.test_config)
     def cmd = """#!/bin/bash
     export SCT_CLUSTER_BACKEND="${params.backend}"
     export SCT_CONFIG_FILES=${test_config}
     if [[ -n "${params.region ? params.region : ''}" ]] ; then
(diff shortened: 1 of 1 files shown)


fgelcer:
worked [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/individual_nemesis/job/longevity-5gb-1h-GrowShrinkClusterNemesis-aws/9/)

vars/getJobTimeouts.groovy
```diff
@@ -2,12 +2,13 @@
 
 List<Integer> call(Map params, String region){
     // handle params which can be a json list
+    def current_region = initAwsRegionParam(params.region, region)
```
fruch, line 5: someday we'll need to rename `initAwsRegionParam` to something more suitable...
since it would work also for other backends

@fgelcer did you tried it for GCP/Azure ? i.e. that random works ? (you might want know that before triggering such jobs)

fgelcer: i will open a task for that, but the jobs i have are AWS only, so i didn't check other cloud providers

yaronkaikov (approved):

vars/getJobTimeouts.groovy
soyacz: to make others working it is needed to update `export SCT_GCE_DATACENTER` and `export SCT_AZURE_REGION_NAME`

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #6620, opened 2023-09-21
fix(k8s): add class providing key based locks and use it for K8S

vponomaryov:
The `test_add_new_node_and_check_old_nodes_are_cleaned_up` K8S functional test has race condition where it may fail with the following error:
```
  Command: 'kubectl cp /tmp/tmpj8_823ao.yaml scylla/sct-cluster-dc-1-kind-0:/tmp/tmpj8_823ao.yaml -c scylla'
  Exit code: 2
  Stderr:
  tar: tmpj8_823ao.yaml: Cannot open: File exists
  tar: Exiting with failure status due to previous errors
  command terminated with exit code 2
```

It is caused by the concurrent `cqlsh` cmd calls to the same Scylla pods. In this test there are about 7 concurrent such calls equal to the number of keyspaces in Scylla.

So, fix it by using lock mechanism per each pod.

### DIFF
--- a/sdcm/remote/kubernetes_cmd_runner.py
+++ b/sdcm/remote/kubernetes_cmd_runner.py
@@ -33,6 +33,7 @@ from sdcm.utils.k8s import KubernetesOps
 from sdcm.utils.common import (
     deprecation,
     generate_random_string,
+    KeyBasedLock,
 )
 from sdcm.utils.decorators import retrying
 from sdcm.wait import wait_for
@@ -41,6 +42,7 @@ from .base import RetryableNetworkException
 from .remote_base import RemoteCmdRunnerBase, StreamWatcher
 
 LOGGER = logging.getLogger(__name__)
+KEY_BASED_LOCKS = KeyBasedLock()
 
 
 def is_scylla_bench_command(command):
@@ -213,8 +215,9 @@ class KubernetesCmdRunner(RemoteCmdRunnerBase):
     # pylint: disable=too-many-arguments,unused-argument
     @retrying(n=3, sleep_time=5, allowed_exceptions=(RetryableNetworkException, ))
     def send_files(self, src, dst, delete_dst=False, preserve_symlinks=False, verbose=False):
-        KubernetesOps.copy_file(self.kluster, src, f"{self.namespace}/{self.pod_name}:{dst}",
-                                container=self.container, timeout=300)
+        with KEY_BASED_LOCKS.get_lock(f"k8s--{self.kluster.name}--{self.namespace}--{self.pod_name}"):
+            KubernetesOps.copy_file(self.kluster, src, f"{self.namespace}/{self.pod_name}:{dst}",
+                                    container=self.container, timeout=300)
         return True
 
     def _run_on_retryable_exception(self, exc: Exception, new_session: bool) -> bool:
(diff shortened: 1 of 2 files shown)


vponomaryov:
Verified here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/253/consoleFull

sdcm/utils/common.py
```diff
@@ -98,6 +98,20 @@
 SCYLLA_GCE_IMAGES_PROJECT = "scylla-images"
 
 
+class KeyBasedLock():  # pylint: disable=too-few-public-methods
+    """Class designed for creating locks based on hashable keys."""
+
+    def __init__(self):
+        self.key_lock_mapping = {}
```
fruch, line 105: No code is cleaning this mapping...

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -213,8 +215,9 @@ def receive_files(self, src, dst, delete_dst=False, preserve_perm=True, preserve
     # pylint: disable=too-many-arguments,unused-argument
     @retrying(n=3, sleep_time=5, allowed_exceptions=(RetryableNetworkException, ))
     def send_files(self, src, dst, delete_dst=False, preserve_symlinks=False, verbose=False):
-        KubernetesOps.copy_file(self.kluster, src, f"{self.namespace}/{self.pod_name}:{dst}",
-                                container=self.container, timeout=300)
+        with KEY_BASED_LOCKS.get_lock(f"k8s--{self.kluster.name}--{self.namespace}--{self.pod_name}"):
```
fruch, line 218: I think we should have a property for the lock name

vponomaryov: Why? It is going to be the only place of it's usage. At least for upcoming future.

sdcm/utils/common.py
vponomaryov: It may be added, yes. But where and why we will delete it?
We don't delete existing locks in other places.

fruch: in other places we have one lock, here you are gonna have one per pod.
we should clean it when destroy pod/node, shouldn't we ?

sdcm/remote/kubernetes_cmd_runner.py
fruch: i.e. if we gonna clean that lock from the mapping

sdcm/utils/common.py
vponomaryov: We destroy pod but then another appears.
In worst case we will have +1 lock - the one used for the temporary pod caused by `cluster extension` operation.

sdcm/remote/kubernetes_cmd_runner.py
vponomaryov: If we have more usages, yes. But key deletion is also not really needed. At least, now.

sdcm/utils/common.py
fruch: One  lock +-  per loader pod.

o.k. forgot it's just in dynamic loader, and that it might be for any node

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #6634, opened 2023-09-27
fix(k8s): make the `cleanup` test be more stable

vponomaryov:
In the `test_add_new_node_and_check_old_nodes_are_cleaned_up` K8S functional test we use utilitary CQL queries which just create noise in DB logs to make our log reader move on fast enough. And those queries may timeout from time to time.
We don't really care whether they timeout or not while some of them pass.
So, ignore those queries errors.
Also, increase the client timeout from `10` to `60` seconds to reduce number of failed queries.


fruch (approved):
LGTM

Issue #6573, opened 2023-09-04
jenkins failed by error: java.io.IOException: Cannot read more than 2Gib at a time


temichus:
jenkins failed by error: java.io.IOException: Cannot read more than 2Gib at a time
 
 
looks like log was spammed by
06:02:40 < t:2023-09-02 04:02:37,247 f:scan_operation_thread.py l:386 c:FullPartitionScanOperation p:WARNING > 67767796

need to move logs from FullPartitionScanOperation file(DEBUG lrvrl) than have log rotation functiolality

temichus:
Debug logs are not displayed in Jenkins. so I think moving to the DEBUG level should help

roydahan:
@yarongilor I think this is the diff prints for your validate partition code.
We probably need to print it to a different log or not include it in the stdout.

fgelcer:
didn't make in sprint 83, moving to sprint 84

yarongilor:
A fix PR is waiting for review: https://github.com/scylladb/scylla-cluster-tests/pull/6636

fruch:
> An update for the PR can be done now, following merging a dependant PR of [scylladb/qa-tasks#1153](https://github.com/scylladb/qa-tasks/issues/1153)

WUT ?

how those two even related ? 
we need to stop spamming the logs, regardless of whatever happening. those two issue are not depend on each other.

roydahan:
@fruch I thought we won't have the issue with the log if we stop doing data validation.

Looking at the original description from @temichus it seems that it's coming from FullPartitionScanOperation, but how does it fill the stdout?

fruch:
> @fruch I thought we won't have the issue with the log if we stop doing data validation.

we stopped it on one use case, you didn't want to disable it on all of them.
and those kind of go together,  FullPartitionScanOperation is enable on almost all large partition cases
so we might have pointed the validation with issues of the fullscan,
both validations are problematic, and cause us more harm than helping us. 

> 
> Looking at the original description from @temichus it seems that it's coming from FullPartitionScanOperation, but how does it fill the stdout?

it's print the whole data it found to the log as warning level, that also comes to the stdout

roydahan:
Yes, but in this case we disabled the verification.

Anyway, can someone quickly cook a patch to stop printing the output to log as warning?
(Printing in debug will remove the problem from Jenkins).

Better patch would be a different log.

Issue #6056, opened 2023-04-20
FullPartitionScanOperation prints 300Mb+ diff to sct.log


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: any

### Logs
* test_id: 9881a5ec-ad3c-4921-9791-a197ff0669a3
* job log: https://jenkins.scylladb.com/job/enterprise-2023.1/job/Reproducers/job/longevity-large-partition-4days-test/4

### Description

Running a job with large paritions with enabled `FullPartitionScanOperation` thread we can get huge diff output into the SCT.log which is useless and greatly decrease the debugability:

```
  66310 < t:2023-04-20 10:29:27,234 f:scan_operation_thread.py l:518  c:FullPartitionScanOperation p:DEBUG > Comparing scan queries output files by: diff -y --suppress-common-lines  /tmp/tmpmaahlfr6 /tmp/tmp07mhi534
  66311 < t:2023-04-20 10:29:27,234 f:local_cmd_runner.py l:62   c:LocalCmdRunner       p:DEBUG > Running command "diff -y --suppress-common-lines  /tmp/tmpmaahlfr6 /tmp/tmp07mhi534"...
  66312 < t:2023-04-20 10:29:27,334 f:base.py         l:222  c:LocalCmdRunner       p:DEBUG > 129249829                           <
  66313 < t:2023-04-20 10:29:27,334 f:base.py         l:222  c:LocalCmdRunner       p:DEBUG > 129249830                           <
...
 500000 < t:2023-04-20 10:29:42,212 f:base.py         l:222  c:LocalCmdRunner       p:DEBUG > 129683473                           <
...
 800000 < t:2023-04-20 10:30:03,070 f:base.py         l:222  c:LocalCmdRunner       p:DEBUG > 129911748                           <
...
 918444 < t:2023-04-20 10:30:19,445 f:base.py         l:222  c:LocalCmdRunner       p:DEBUG > 129793358                           <
...
1376741 < t:2023-04-20 10:31:29,886 f:base.py         l:222  c:LocalCmdRunner       p:DEBUG > 129335222                           <
...
1462183 < t:2023-04-20 10:31:44,732 f:base.py         l:222  c:LocalCmdRunner       p:DEBUG > 129249829                           <
...
[...]

PR #6185, opened 2023-05-24
 improvement(large-partitions): Run Deletion-nemeses + validate_partitions

yarongilor:
improvement(large-partitions): Run Deletion-nemeses+validate_partitions
    
            Changed validate_partitions to intersect with protected "partition_range_with_data_validation".
            This is instead of simply counting all existing partitions.
            Example:
                    max_partitions_in_test_table: 400
                    partition_range_with_data_validation: 0-100
                    ==> outcome:
                    1. Deletion nemeses (like DeleteByRowsRangeMonkey) will only select partition range
                    of 101 - 400.
                    2. validate_partitions will count rows for partition range 0 - 100 only.
    
            Fixes: #6183

The PR is rewritten in order to have a solution for all longevities to run deletion-nemeses along side validate_partitions, avoiding verification collisions.
Fixes: #6183

The following tests were executed in order to verify this PR:

- [upgrade test](https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/rolling-upgrade-ubuntu20.04-test-yg/2/)
- [Longevity 3h GCE](https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/longevity-10gb-3h-gce-test2/15/)
- [Longevity LWT 3h](https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/longevity-lwt-3h-test-validate-partitions/2/)

## PR pre-checks (self review)

- [x] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [x] I didn't leave commented-out/debugging code
- [x] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6636, opened 2023-09-28
fix(FullPartitionScanOperation): Do not print full diff output

yarongilor:
When full partition scan data is validated, there might
	be a large diff. It cannot be fully printed to log.
	So it will now print the diff size and few first lines only.
	Fixes: #6573 , #6065
Fixes: #6573
Fixes: #6056

This PR is a partial quick fix.
There is an additional fix planned to address the reverse-queries partition validation.
It is in the scope of task: https://github.com/scylladb/qa-tasks/issues/1199

### DIFF
--- a/sdcm/scan_operation_thread.py
+++ b/sdcm/scan_operation_thread.py
@@ -383,7 +383,9 @@ class FullPartitionScanOperation(FullscanOperationBase):
                 self.log.debug("Compared output of normal and reversed queries is identical!")
                 result = True
             else:
-                self.log.warning("Normal and reversed queries output differs: \n%s", result.stdout.strip())
+                output = result.stdout.strip()
+                self.log.warning("Normal and reversed queries output differs (diff-size is %s): \n%s",
+                                 len(output.splitlines()), output[200])
                 ls_cmd = f"ls -alh {file_names}"
                 head_cmd = f"head {file_names}"
                 for cmd in [ls_cmd, head_cmd]:


fgelcer (approved):
seems like a little workaround to have the overload failures in Jenkins fixed, but the real fix is not exactly that.. and it will be dealt as part of another task/PR

sdcm/scan_operation_thread.py
```diff
@@ -383,7 +383,9 @@ def _compare_output_files(self) -> bool:
                 self.log.debug("Compared output of normal and reversed queries is identical!")
                 result = True
             else:
-                self.log.warning("Normal and reversed queries output differs: \n%s", result.stdout.strip())
+                output = result.stdout.strip()
```
roydahan, line 386: Maybe we should print it to a new special log file?

sdcm/scan_operation_thread.py
```diff
@@ -383,7 +383,9 @@ def _compare_output_files(self) -> bool:
...
+                self.log.warning("Normal and reversed queries output differs (diff-size is %s): \n%s",
```
roydahan, line 387: Maybe should be an error?
No one looks at warning...

sdcm/scan_operation_thread.py
vponomaryov: Agree. And data writing must be done directly in the command which we define in line 377.
For the moment we load whole diff to the RAM.
Then, knowing that the diff may be huge 2Gb+, it becomes a need to pipe the data to a file directly.

vponomaryov (changes requested):

sdcm/scan_operation_thread.py
yarongilor: @roydahan , the problem is we cannot tell it's a real error for sure.
It can be confirmed only if there's no background load or any deletion nemesis running.
That's why this check is less functional and more statistic, in the overall results.
(Perhaps, optimally, we would save the failures/success statistics to ES or something)

roydahan: I disagree. any delete nemesis or additional writes aren't or at least shouldn't happen for this validated range.
This range is only being read during the test.

sdcm/scan_operation_thread.py
yarongilor: @roydahan , @vponomaryov ,
I don't think we should currently print and save an extra log file for it.
No developer ever looked at these errors or asked to have any such full output. Not sure it gives any added value, surely not at the moment. We better only consider it per demand.
@vponomaryov , the diff is actually quite small, not sure why you think a single diff is getting to 2Gb+, at least not in recent master code.
I don't expect any memory issue for an operation that apparently takes about 13ms like - 
```
< t:2022-09-28 14:36:56,703 f:scan_operation_thread.py l:326  c:FullPartitionScanThread p:INFO  > Comparing scan queries output files by: diff -y --suppress-common-lines  /tmp/tmp1q3hu1c1 /tmp/tmpuhunh1i7
< t:2022-09-28 14:36:56,716 f:scan_operation_thread.py l:332  c:FullPartitionScanThread p:WARNING > Normal and reversed queries output differs: 
< t:2022-09-28 14:36:56,716 f:scan_operation_thread.py l:332  c:FullPartitionScanThread p:WARNING > 190497180						      <
```

sdcm/scan_operation_thread.py
fruch: also no one would be looking on error prints, this should be an event with error level, and fail the test.
otherwise it's pointless verification

sdcm/scan_operation_thread.py
vponomaryov: @yarongilor the `2Gb+` value was taken from the bugreport:
```
jenkins failed by error: java.io.IOException: Cannot read more than 2Gib at a time
```
Knowing that in common cases the log is up to 10Mb we may assume that there is significant `flood`.

Personally, I met the `300Mb+` side of the diff here: https://github.com/scylladb/scylla-cluster-tests/issues/6056

And, according to the current logic, whole this diff always goes to the RAM first and only then gets processed.

sdcm/scan_operation_thread.py
yarongilor: @roydahan , @fruch , i think what was defined as the main goal for this validation is challenging scylla's reversed queries in terms of sanity to see no coredumps / memory issues / other exceptions, during load + deletions + cache + multiple sstables/ memtables updates.
The data validation is quite a bonus with this regard.
Now what you are suggesting above is a much limited and quite "static" validation. It mean validating data that was only written once during prepare.
So a solution could be (when `validate-date: true`) to validate data only if the partition is in range of `partition_range_with_data_validation` . This will have a dependency and requires https://github.com/scylladb/scylla-cluster-tests/pull/6185 to be merged first. please confirm.

sdcm/scan_operation_thread.py
temichus: it can be implemented as sending 1 line WARNING/ERROR message/event with a summary as @yarongilor  suggested but without output[200]  in and all other logs sent as DEBUG. bit SCT lig will be split automatically

sdcm/scan_operation_thread.py
amoskong:
```
self.log.warning("Normal and reversed queries output differs (diff-size is %s lines): \n%s <...clipped...>",
```

fruch:
@yarongilor, can you please finish this one, it's still happens a tier1 job

@bhalevy, FYI

fruch:
Addressed in #7079

PR #6648, opened 2023-10-03
fix(manager_cli): extend timeout on test_restore_backup_with_task

karol-kokoszka:

- [ X] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ X] I didn't leave commented-out/debugging code
- [ X] New configuration option are added and documented (in `sdcm/sct_config.py`)

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -547,7 +547,7 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         assert backup_task_status == TaskStatus.DONE, \
             f"Backup task ended in {backup_task_status} instead of {TaskStatus.DONE}"
         self.verify_backup_success(mgr_cluster=mgr_cluster, backup_task=backup_task, restore_data_with_task=True,
-                                   timeout=1000)
+                                   timeout=3000)
         self.run_verification_read_stress()
         mgr_cluster.delete()  # remove cluster at the end of the test
         self.log.info('finishing test_restore_backup_with_task')


fgelcer (approved):
@karol-kokoszka @ShlomiBalalis , was this timeout enough prior to manager `3.2.2`?
if yes, then we need to understand what is taking longer, and if no, this seems very strange, otherwise we would have already increased it for either `3.2.0` or `3.2.1`

in the meantime, LGTM

karol-kokoszka:
@fgelcer looks the timeout was enough from manager < 3.2.2.
There is an easy explanation of why it was enough.

Small_table_optimization means that for small tables (per size) on the keyspaces where RF=number of nodes, all token ranges are send in single repair job.
It increases the performance of the repair, but may kill the performance of the cluster when the table is big.
Before 3.2.2 we had a bug that was applying small_table_optimization even for big tables when RF=number of nodes.

In 3.2.2, this optimization is not applied to big tables anymore.
It decreases the performance of the repair, but increases the perf of the cluster during the repair.

fgelcer:
> @fgelcer looks the timeout was enough for manager < 3.2.2. There is an easy explanation of why it was enough.
> 
> Small_table_optimization means that for small tables (per size) on the keyspaces where RF=number of nodes, all token ranges are send in single repair job. It increases the performance of the repair, but may kill the performance of the cluster when the table is big. Before 3.2.2 we had a bug that was applying small_table_optimization even for big tables when RF=number of nodes.
> 
> In 3.2.2, this optimization is not applied to big tables anymore. It decreases the performance of the repair, but increases the perf of the cluster during the repair.

wait, we are talking about a restore job (what this PR is about)...
in this case, shouldn't we use the max possible throughput, as when restoring a keyspace/table, we expect it to not be in use yet? why was it affected by the changes to the small_table_optimization repair?

karol-kokoszka:
> wait, we are talking about a restore job (what this PR is about)...
in this case, shouldn't we use the max possible throughput, as when restoring a keyspace/table, we expect it to not be in use yet? why was it affected by the changes to the small_table_optimization repair?

It's affected by the change, because max possible throughput that we can provide with manager is still smaller than what small-tables-optimization does.
Small table optimization sends all token ranges at once.
Max throughput is token ranges per single repair job = max_repair_ranges_in_parallel. More info about the intensity is here https://manager.docs.scylladb.com/stable/repair/#repair-intensity

mgmt_cli_test.py
```diff
@@ -547,7 +547,7 @@ def test_restore_backup_with_task(self):
         assert backup_task_status == TaskStatus.DONE, \
             f"Backup task ended in {backup_task_status} instead of {TaskStatus.DONE}"
         self.verify_backup_success(mgr_cluster=mgr_cluster, backup_task=backup_task, restore_data_with_task=True,
-                                   timeout=1000)
+                                   timeout=3000)
```
roydahan, line 550: Need to check if it make sense to put such a high timeout.
Better to use a timeout that is close enough to the expected completion time, so we won't miss future regressions.

fgelcer:
@scylladb/qa-maintainers please merge and backport
test passed with this fix [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/manager/job/repair-timeout-fix-by-karol/2/)

mgmt_cli_test.py
fgelcer: so, last run (staging job) took 31 minutes, where the current timeout is 50 minutes...
let's get it running couple more times, and we can confirm it takes really around 31 minutes, and we can give another 4 - 5 just to be sure a little network disturbance won't fail the test...

soyacz (approved):
LGTM

mgmt_cli_test.py
vponomaryov: It is good case to use the `adaptive_timeout` context manager.
Example: https://github.com/scylladb/scylla-cluster-tests/blob/2aa7dc9c5d387469fd8816223ecb22bab22fd986/sdcm/nemesis.py#L4608

Reason: it will allow us to have `soft timeout` that will create effor SCT event if exceeded but not interrupt the operation. And have the `hard timeout` on the operation itself.

PR #6652, opened 2023-10-04
fix(set_auth_rf): remove parallelism

soyacz:
There are timeouts errors during repair of system_auth when setting system_auth during node setup. Looks like repair freezes (db log hang for 20m on repair roles table).

Because it was never recommended to run this repair in parallel (but should not affect either). Removing parallelism so we can exclude this as a root cause of freezing repair of system_auth table.

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -905,9 +905,6 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
 
     def set_system_auth_rf(self, db_cluster=None):
 
-        def _nodetool_repair(node):
-            node.run_nodetool(sub_cmd="repair -pr system_auth", timeout=MINUTE_IN_SEC * 20)
-
         if not db_cluster:
             db_cluster = self.db_cluster
         # No need to change system tables when running via scylla-cloud
@@ -935,9 +932,8 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
                              num_retry_on_failure=3)
         self.log.debug("system_auth description: %s", res.stdout)
         self.log.info('repair system_auth keyspace ...')
-        parallel_objects = ParallelObject(self.db_cluster.nodes, num_workers=min(
-            32, len(self.db_cluster.nodes)), timeout=MINUTE_IN_SEC * 25)
-        parallel_objects.run(_nodetool_repair)
+        for node in self.db_cluster.nodes:
+            node.run_nodetool(sub_cmd="repair -pr system_auth", timeout=MINUTE_IN_SEC * 20)
         self.log.info('repair system_auth keyspace done')
 
     @cache


sdcm/tester.py
```diff
@@ -935,9 +932,8 @@ def _nodetool_repair(node):
                              num_retry_on_failure=3)
         self.log.debug("system_auth description: %s", res.stdout)
         self.log.info('repair system_auth keyspace ...')
-        parallel_objects = ParallelObject(self.db_cluster.nodes, num_workers=min(
-            32, len(self.db_cluster.nodes)), timeout=MINUTE_IN_SEC * 25)
-        parallel_objects.run(_nodetool_repair)
+        for node in self.db_cluster.nodes:
+            node.run_nodetool(sub_cmd="repair -pr system_auth", timeout=MINUTE_IN_SEC * 20)
```
fgelcer, line 936: IIUC the flag `-pr` is only on primary replica, hence we ran them in parallel... so, shouldn't we remove the `-pr` at all if we are running sequentially?

soyacz: `-pr` is recommended according to https://github.com/scylladb/scylla-enterprise/issues/3103#issuecomment-1633394228

fgelcer: `-pr` is recommended when running all nodes in parallel, and you are now calling them sequentially, and it may also be unnecessary, as it would work doing only once, in one of the nodes without the `-pr`

IIUC @asias ' comment, the test currently does the right thing, but better to hear a confirmation

fgelcer:
> So, this change looks significant and worth creating an issue.

there is an issue, and the code is currently doing what @asias said (and you quoted above)...
we are not trying to "workaround" it to take less time, IIUC

vponomaryov:
> > So, this change looks significant and worth creating an issue.
> 
> there is an issue, and the code is currently doing what @asias said (and you quoted above)... we are not trying to "workaround" it to take less time, IIUC

This PR doesn't refer to any issue. The code here removes the `parallel` part, it violates the `to repair on all nodes in parallel` statement.

soyacz:
the clue for fixing repair time was to run with `-pr`, this PR tries to adhere more to what our docs say: https://enterprise.docs.scylladb.com/stable/operating-scylla/nodetool-commands/repair.html#nodetool-repair
`"Use the nodetool repair -pr on each node in the cluster, sequentially."`

fruch:
it's might be worthwhile to mention https://github.com/scylladb/scylla-enterprise/issues/3103 about the cases we have that it got stuck ?

fruch:
> the clue for fixing repair time was to run with `-pr`, this PR tries to adhere more to what our docs say: https://enterprise.docs.scylladb.com/stable/operating-scylla/nodetool-commands/repair.html#nodetool-repair `"Use the nodetool repair -pr on each node in the cluster, sequentially."`

I think @soyacz is correct, we should follow the docs.
and if we want to test how to speed things up, we should be doing it on it's own, not on every single test case in SCT...
and once it's the approved and tested and documented way, switch all cases to it

fgelcer (approved):
> the clue for fixing repair time was to run with -pr, this PR tries to adhere more to what our docs say: https://enterprise.docs.scylladb.com/stable/operating-scylla/nodetool-commands/repair.html#nodetool-repair
`"Use the nodetool repair -pr on each node in the cluster, sequentially."`

based on the comment above, and the content in the official document of Scylla, LGTM

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

fgelcer:
are backports required, @soyacz ?

soyacz:
hmm, PR introducing this change was marked to be backported but wasn't (https://github.com/scylladb/scylla-cluster-tests/pull/6396).
We don't run many multi-dc multi-az tests on release branches, so I'm not sure if it is worth to do so.
@fruch WDYT?

fruch:
> hmm, PR introducing this change was marked to be backported but wasn't (#6396). We don't run many multi-dc multi-az tests on release branches, so I'm not sure if it is worth to do so. @fruch WDYT?

there's no rush to backport it, as we can see t (https://github.com/scylladb/scylla-cluster-tests/pull/6396) was proven to be unstable, and good thing we didn't backported it.

let's wait for an actual reason for backporting.

PR #6655, opened 2023-10-04
improvement(manager_cli): use adaptive_timeout instead of large timeout

fgelcer:
since Manager 3.2.2 is doing lower parallelism in
repairs, to have lower impact on latency, the used timeout (since manager 3.2.0) became too short, and it was updated on #6648 , but it may be too much,
as we still want to fail with timeout at some point.

during the last 2 runs done as part of #6648 staging and release of Manager 3.2.2, the duration was
31 and 32 minutes (respectively), hence using now
`adaptive_timeout` of 36 minutes.

running [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/manager/job/repair-timeout-fix-by-karol/5/)

@roydahan , FYI


karol-kokoszka (approved):

mgmt_cli_test.py
```diff
@@ -546,8 +547,9 @@ def test_restore_backup_with_task(self):
         backup_task_status = backup_task.wait_and_get_final_status(timeout=1500)
         assert backup_task_status == TaskStatus.DONE, \
             f"Backup task ended in {backup_task_status} instead of {TaskStatus.DONE}"
-        self.verify_backup_success(mgr_cluster=mgr_cluster, backup_task=backup_task, restore_data_with_task=True,
-                                   timeout=3000)
+        timeout = 36 * 60
+        with adaptive_timeout(Operations.MGMT_REPAIR, self.db_cluster.nodes[0], timeout=timeout):
+            self.verify_backup_success(mgr_cluster=mgr_cluster, backup_task=backup_task, restore_data_with_task=True)
```
vponomaryov, line 552: Is it intentional that the `verify_backup_success` method doesn't have anymore provided `timeout` param which should have been the `hard timeout`?

sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -58,6 +58,9 @@ class Operations(Enum):
     CREATE_MV = ("create_mv", _get_soft_timeout, ("timeout",))
     MAJOR_COMPACT = ("major_compact", _get_soft_timeout, ("timeout",))
     REPAIR = ("repair", _get_soft_timeout, ("timeout",))
+    MGMT_REPAIR = ("mgmt_repair", _get_soft_timeout, ("timeout",))
+    BACKUP = ("backup", _get_soft_timeout, ("timeout",))
+    RESTORE = ("restore", _get_soft_timeout, ("timeout",))
```
vponomaryov, line 63: At first, the 3 new attributes are not consistent in naming. One has `MGMT_` prefix other don't have.
And at second, 2 out of 3 attributes are not used. So, are you sure we should add it now?

vponomaryov (changes requested):

mgmt_cli_test.py
fgelcer: will add

sdcm/utils/adaptive_timeouts/__init__.py
fgelcer: as you can see in the line above `MGMT_REPAIR`, there is already a repair, and that is the reason to have `MGMT_*` before the "new" repair... other operations today can't be done without the manager, hence it is not necessary to have the prefix to them.

the other 2 are not in use, because they will be added in following up PRs, but since i'm already changing it here, and to avoid future conflicts, i added all the necessary attributes in here at once.

fruch: anyhow, why this PR is adding options it's not using ?

when you introduce it to backup and restore you can define them...

fruch:
@fgelcer 

don't we want those numbers collected as in release branches ?

fgelcer:
> @fgelcer
> 
> don't we want those numbers collected as in release branches ?

what numbers? the timeout? we will perhaps get in there, but for now, with manager `3.2.2` 32 minutes will be enough, and it has some 10% more to the soft timeout, and 50 minutes to the hard timeout

fruch:
> > @fgelcer
> > don't we want those numbers collected as in release branches ?
> 
> what numbers? the timeout? we will perhaps get in there, but for now, with manager `3.2.2` 32 minutes will be enough, and it has some 10% more to the soft timeout, and 50 minutes to the hard timeout

I was talking about where we want this code backported

fruch (approved):
LGTM

fgelcer:
> > > @fgelcer
> > > don't we want those numbers collected as in release branches ?
> > 
> > 
> > what numbers? the timeout? we will perhaps get in there, but for now, with manager `3.2.2` 32 minutes will be enough, and it has some 10% more to the soft timeout, and 50 minutes to the hard timeout
> 
> I was talking about where we want this code backported

for now, it should only hit `manager-3.2`, and others, we will see

sdcm/utils/adaptive_timeouts/__init__.py
fgelcer: i foresee conflicts (because of backports) if each one of these will add the relevant parameter at different times... so, because we may already have it in there, once we will start adding such adaptive timeouts to other `mgmt_cli_test.py`

vponomaryov (approved):
LGTM

Issue #6330, opened 2023-07-05
Filesystem errors when dropping tables


soyacz:
When we drop table, Scylla is logging warn messages with 'filesystem_error' which is catch as an ERROR event in SCT:
```
2023-06-25 00:42:30.932: (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=d7200486-9ca5-460b-84a6-c185e935ff69: type=FILESYSTEM_ERROR regex=filesystem_error line_number=10370 node=sct-cluster-us-east1-b-us-east1-1
WARN  2023-06-25 00:40:20,202 [shard  0] sstable - Could not remove table directory "/var/lib/scylla/data/drop_table_during_repair_ks_1/standard1-66c8714012ef11eea150e788892de373/snapshots": std::filesystem::__cxx11::filesystem_error (error system:39, filesystem error: remove failed: Directory not empty [/var/lib/scylla/data/drop_table_during_repair_ks_1/standard1-66c8714012ef11eea150e788892de373/snapshots]). Ignored.
```
There's an issue related to this in Scylla: https://github.com/scylladb/scylladb/issues/13538 without conclusion.

I propose to ignore filesystem_error when it starts with `WARN` and contains `snapshots` word

fruch:
if I'll pick a string, it would be `Directory not empty`

PR #6664, opened 2023-10-06
fix(event): reduce severity of 'Directory not empty' error

soyacz:
When dropping keyspace we run into `filesystem_error` due to failure in directory deletion. This is valid behavior as Scylla should not delete snapshots (often created during drop automatically).

Fix is about reducing severity of such log lines to `NORMAL` so we don't fail tests because of that.

Fixes: #6330


fgelcer (dismissed):
IIUC the problem, it should be fine, but it would be much more precise if we "reduce" its severity when dropping a table (so we can ignore it only for that specific table)...
but other than it, LGTM

fruch:
@soyacz 

FYI it's a bug that was fixed in scylla master (but wasn't backported, since it's cosmetic)

https://github.com/scylladb/scylladb/issues/13538

So we should leave it in a comment, and it should be remove once fix is backported (or when current releases would be obsolete)

soyacz:
> 
> So we should leave it in a comment, and it should be remove once fix is backported (or when current releases would be obsolete)

we run tests SCT master also with 2023.1/5.2 and then it is also causing noise. 
I added a comment to remove this workaround when we drop support for these.

fgelcer (changes requested):
need to fix unit tests:
```
08:17:20  =================================== FAILURES ===================================
08:17:20  ______________ TestDatabaseLogEvent.test_system_error_events_list ______________
08:17:20  
08:17:20  self = <unit_tests.test_sct_events_database.TestDatabaseLogEvent testMethod=test_system_error_events_list>
08:17:20  
08:17:20      def test_system_error_events_list(self):
08:17:20  >       self.assertSetEqual(set(dir(DatabaseLogEvent)) - set(dir(LogEvent)),
08:17:20                              {ev.type for ev in SYSTEM_ERROR_EVENTS})
08:17:20  E       AssertionError: Items in the first set but not the second:
08:17:20  E       'DIRECTORY_NOT_EMPTY'
08:17:20  
08:17:20  unit_tests/test_sct_events_database.py:82: AssertionError
```

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #6670, opened 2023-10-10
fix(k8s): fix fstrim scylla disks on k8s

soyacz:
When running `fstrim_scylla_disks` on K8s it fails with error: `Unable to use a TTY - input is not a terminal or the right kind of file`. This is possibly caused by `-ti` flag.

Remove `-ti` flag from running trim command.

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -2162,7 +2162,7 @@ class BaseScyllaPodContainer(BasePodContainer):  # pylint: disable=abstract-meth
             f"Failed to find pods using '{pods_selector}' selector on '{self.node_name}' node "
             f"in '{namespace}' namespace. Didn't run the 'fstrim' command")
         self.parent_cluster.k8s_cluster.kubectl(
-            f"exec -ti {podnames[0]} -- sh -c 'fstrim -v {scylla_disk_path}'",
+            f"exec {podnames[0]} -- sh -c 'fstrim -v {scylla_disk_path}'",
             namespace=namespace)
 
 


fruch (approved):
LGTM

can you just point where it was seen ?

soyacz:
https://argus.scylladb.com/test/30d13802-7666-422c-b95e-ad816869653a/runs?additionalRuns[]=5cb29798-8987-40d0-a59e-cb4c6e9828b1

vponomaryov (changes requested):
The error in the referenced test run is following:

```
error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "8a32e2b6528f9aa37d2dc639a4ec4b9319836fe0200d9b1f24645ab94e9982a2": OCI runtime exec failed: exec failed: unable to start container process: error adding pid 62539 to cgroups: failed to write 62539: openat2 /sys/fs/cgroup/systemd/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod49f63626_ef88_4964_b40d_0fe7fb0a2c4a.slice/cri-containerd-d4da402f06f9577c8be64ba4ea01070f2646c6a9da0e3e2e838d4670f61e47c1.scope/cgroup.procs: no such file or directory: unknown
```

And the root cause of the failure doesn't get fixed here.
It is some instability in the node configuration which appear from time to time.

We would not have successful runs of the performance jobs if the command options, which get removed here, were the reason of the failure.

soyacz:
> The error in the referenced test run is following:
> 
> ```
> error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "8a32e2b6528f9aa37d2dc639a4ec4b9319836fe0200d9b1f24645ab94e9982a2": OCI runtime exec failed: exec failed: unable to start container process: error adding pid 62539 to cgroups: failed to write 62539: openat2 /sys/fs/cgroup/systemd/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod49f63626_ef88_4964_b40d_0fe7fb0a2c4a.slice/cri-containerd-d4da402f06f9577c8be64ba4ea01070f2646c6a9da0e3e2e838d4670f61e47c1.scope/cgroup.procs: no such file or directory: unknown
> ```
> 
> And the root cause of the failure doesn't get fixed here. It is some instability in the node configuration which appear from time to time.
> 
> We would not have successful runs of the performance jobs if the command options, which get removed here, were the reason of the failure.



> The error in the referenced test run is following:
> 
> ```
> error: Internal error occurred: error executing command in container: failed to exec in container: failed to start exec "8a32e2b6528f9aa37d2dc639a4ec4b9319836fe0200d9b1f24645ab94e9982a2": OCI runtime exec failed: exec failed: unable to start container process: error adding pid 62539 to cgroups: failed to write 62539: openat2 /sys/fs/cgroup/systemd/kubepods.slice/kubepods-besteffort.slice/kubepods-besteffort-pod49f63626_ef88_4964_b40d_0fe7fb0a2c4a.slice/cri-containerd-d4da402f06f9577c8be64ba4ea01070f2646c6a9da0e3e2e838d4670f61e47c1.scope/cgroup.procs: no such file or directory: unknown
> ```
> 
> And the root cause of the failure doesn't get fixed here. It is some instability in the node configuration which appear from time to time.
> 
> We would not have successful runs of the performance jobs if the command options, which get removed here, were the reason of the failure.

true, this sometimes passes.
Anyway, we always get the error `Unable to use a TTY - input is not a terminal or the right kind of file` - even for passing ones.

as to error, is it possible we get not running pod? Maybe we should filter just for running ones?

soyacz:
@vponomaryov 
How about instead running in specific pod, we executed it with something like this:?
```
kubectl run -it --rm --privileged --restart=Never --node-name=<scylla-node> fstrim --image=docker.io/scylladb/k8s-local-volume-provisioner:0.1.0-rc.0 -- bash -c 'fstrim -v {scylla_disk_path}'
```

fruch:
> @vponomaryov 
> How about instead running in specific pod, we executed it with something like this:?
> ```
> kubectl run -it --rm --privileged --restart=Never --node-name=<scylla-node> fstrim --image=docker.io/scylladb/k8s-local-volume-provisioner:0.1.0-rc.0 -- bash -c 'fstrim -v {scylla_disk_path}'
> ```

I'm not sure it would change the situation

If I understand correctly it's a bug in dockerd or something like that

Maybe just documenting it in an SCT issue is enough, so we can follow it up to see how many times is happening

vponomaryov:
> as to error, is it possible we get not running pod? Maybe we should filter just for running ones?

Sometime ago I played with such node a bit, and `exec` was not working there. 

@soyacz if you still want to remove that `warning` message by removing the `-ti` options then need to update the PR and commit description to say exactly this. Because it doesn't fix the `fstrim` problem.

soyacz:
> @soyacz if you still want to remove that `warning` message by removing the `-ti` options then need to update the PR and commit description to say exactly this. Because it doesn't fix the `fstrim` problem.

if it is not going to help, no need for that.

Closing with issue: https://github.com/scylladb/scylla-cluster-tests/issues/6681

PR #6677, opened 2023-10-11
improvement(perf latency): unify steady state

fgelcer:
since 2 new tests were added, and they had to add
"steady_state_latency" function, this PR will only unify them into 1 single function, so changes will in one single place, and the nemesis class doesn't need to be aware of it anymore.

Refs: scylladb/qa-tasks#1516

### DIFF
--- a/performance_regression_test.py
+++ b/performance_regression_test.py
@@ -61,6 +61,10 @@ class PerformanceRegressionTest(ClusterTester):  # pylint: disable=too-many-publ
         # emails for each test. When we move to use SCT Runners, it won't be necessary.
         self._clean_email_data()
         super().__init__(*args)
+        # next 3 lines, is a workaround to have it working inside `latency_calculator_decorator`
+        self.cluster = self.db_cluster
+        self.monitoring_set = self.monitors
+        self.tester = self
 
     @teardown_on_exception
     @log_run_info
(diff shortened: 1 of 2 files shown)


performance_regression_test.py
```diff
@@ -271,6 +271,18 @@ def _run_cql_commands(self, cmds, node=None):
             with self.db_cluster.cql_connection_patient(node) as session:
                 session.execute(cmd)
 
+    @latency_calculator_decorator
+    def steady_state_latency(self, sleep_time):  # pylint: disable=no-self-use
```
soyacz, line 275: I think better name would be `measurement_time` instead of `sleep_time`

soyacz (dismissed):
one small comment, otherwise
LGTM

fruch:
@fgelcer you have precommit issue to handle

fgelcer:
[with upgrade](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/move_steady_state/job/latency_during_upgrade/2/) passed

fgelcer:
> @fgelcer you have precommit issue to handle

fixed

performance_regression_test.py
fgelcer: changed to `steady_state_time`

soyacz (dismissed):
LGTM

fgelcer:
i have placed the "new" call to steady state in the wrong place... i have fixed it, and running the [unique sequence test](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/move_steady_state/job/latency_during_unique_sequence/3/) and the [grow and shrink cluster](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/move_steady_state/job/latency_during_grow_shrink/3/)

fgelcer:
changed another small thing, since the "latency during upgrades" worked, and the other ones didn't
change is to move the "workaround" that creates the `self.cluster`, `self.monitoring_set` and `self.tester` to the `PerformanceRegressionTest` `__init__()` class

re-testing

performance_regression_test.py
```diff
@@ -319,6 +323,18 @@ def run_mixed_workload(self, nemesis=False):
...
         self.check_regression()
 
+    @latency_calculator_decorator
+    def steady_state_latency(self, steady_state_time):  # pylint: disable=no-self-use
+        InfoEvent(message='Starting Steady State calculation for %ss' % steady_state_time).publish()
+        time.sleep(steady_state_time)
+        InfoEvent(message='Ended Steady State calculation. Took %ss' % steady_state_time).publish()
+
+    @latency_calculator_decorator
+    def post_upgrades_steady_state(self, steady_state_time):  # pylint: disable=no-self-use
```
temichus, line 333: I think that this should be moved to the PerformanceRegressionUpgradeTest
because it is not used in PerformanceRegressionUpgradeTest and probably will never be used

fgelcer: i'm doesn't fully agree that it will never be used, but also because of its similarity, if the `steady_state_calculation` is already in the parent class, this one has not reason to be in the child class, but IMHO should go along with that

fruch:
@soyacz, can you take over this one ?

fruch:
this is not needed anymore

PR #6678, opened 2023-10-12
improvement(perf latency): email subject to have nemesis name

fgelcer:
since we have a new test, that does run a different nemesis class, we want to have different email subjects to each one, since "operations" is already not enough.

email subject will change "operations" to "`nemesis_class_name`", but everything else should stay the same.

### DIFF
--- a/sdcm/results_analyze/__init__.py
+++ b/sdcm/results_analyze/__init__.py
@@ -222,11 +222,12 @@ class LatencyDuringOperationsPerformanceAnalyzer(BaseResultsAnalyzer):
     Get latency during operations performance analyzer
     """
 
-    def __init__(self, es_index, es_doc_type, email_recipients=(), logger=None, events=None):   # pylint: disable=too-many-arguments
+    def __init__(self, es_index, es_doc_type, email_recipients=(), logger=None, events=None,  # pylint: disable=too-many-arguments
+                 nemesis_class_name=None):
         super().__init__(es_index=es_index, es_doc_type=es_doc_type, email_recipients=email_recipients,
                          email_template_fp="results_latency_during_ops_short.html", logger=logger, events=events)
         self.percentiles = ['percentile_90', 'percentile_99']
-        self.test_name_for_email_subject = 'latency during operations'
+        self.test_name_for_email_subject = f'latency during {nemesis_class_name}'
 
     def get_debug_events(self):
         return self.get_events(event_severity=[Severity.DEBUG.name])
@@ -497,10 +498,11 @@ class LatencyDuringOperationsPerformanceAnalyzer(BaseResultsAnalyzer):
 
 class LatencyDuringUpgradesPerformanceAnalyzer(LatencyDuringOperationsPerformanceAnalyzer):
     def __init__(self, es_index, es_doc_type, email_recipients=(), logger=None,  # pylint: disable=too-many-arguments
-                 events=None):
+                 events=None, nemesis_class_name='upgrades'):
         super().__init__(es_index=es_index, es_doc_type=es_doc_type, email_recipients=email_recipients,
                          logger=logger, events=events)
         self.percentiles = ['percentile_90', 'percentile_99']
+        self.log.debug('we set %s for compatibility reason, but not used in this test', nemesis_class_name)
         self.test_name_for_email_subject = 'latency during upgrades'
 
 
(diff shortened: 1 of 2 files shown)


sdcm/results_analyze/__init__.py
```diff
@@ -222,11 +222,12 @@ class LatencyDuringOperationsPerformanceAnalyzer(BaseResultsAnalyzer):
...
     """
 
-    def __init__(self, es_index, es_doc_type, email_recipients=(), logger=None, events=None):   # pylint: disable=too-many-arguments
+    def __init__(self, es_index, es_doc_type, email_recipients=(), logger=None, events=None,
+                 nemesis_class_name=None):   # pylint: disable=too-many-arguments
         super().__init__(es_index=es_index, es_doc_type=es_doc_type, email_recipients=email_recipients,
                          email_template_fp="results_latency_during_ops_short.html", logger=logger, events=events)
         self.percentiles = ['percentile_90', 'percentile_99']
-        self.test_name_for_email_subject = 'latency during operations'
+        self.test_name_for_email_subject = f'latency during {nemesis_class_name}'
```
fruch, line 230: This thing should come from SCT configuration.

One should be able to put whatever he wants in the configuration for specific use case

fgelcer: > This thing should come from SCT configuration.
> 
> One should be able to put whatever he wants in the configuration for specific use case

it does come from SCT configuration, but by taking the used `nemesis_class_name`

fruch: I still don't understand why not using `email_subject_postfix` as the other perf test....

fruch: I rather see it more align to the rest of the tests and with better control of each case

fruch:
I think this would be better option to maintine: 
https://github.com/scylladb/scylla-cluster-tests/pull/6679

sdcm/results_analyze/__init__.py
fgelcer: here are 2 examples of email subject:
```
[aws][i3en.2xlarge] Performance Regression Compare Results (latency during upgrades 650gb) - test_latency_write_with_upgrade - 5.4.0~dev-0.20231006.498e3ec435be - 2023-10-11 19:09:57
```

and:
```
[aws][i3en.2xlarge] Performance Regression Compare Results (latency during operations 650gb) - test_latency_mixed_with_nemesis - 5.4.0~dev.20231003.d543b96d1837 - 2023-10-04 17:13:04
```

considering the extent of the subject already, to append something as a suffix (or prefix) will make it even longer, and some times, part of the important information is not shown in the list of emails

fgelcer:
> I think this would be better option to maintine: #6679

your solution is breaking other perf tests (like `latency during upgrades` test), and doesn't need to add another field for it, we could have it automatic calculated (as it is proposed in this current PR), and the suggestion of your on the PR above could be in addition, or to replace that part of the email subject, if it exists, otherwise, it would be how it is proposed in this PR

sdcm/tester.py
```diff
@@ -3115,6 +3115,7 @@ def wait_data_dir_reaching(self, size, node):
     def check_latency_during_ops(self, op_is_upgrade=False):
         start_time = self.start_time if not self.create_stats else self._stats["test_details"]["start_time"]
         end_time = time.time()
+        nemesis_class_name = self.params.get('nemesis_class_name') or 'upgrades'
```
temichus, line 3118: agree with @fruch that https://github.com/scylladb/scylla-cluster-tests/pull/6679 is probably a better solution because this line brings "magic logic" where we assume that if it is not a nemesis test it is an upgrades test

For me, it looks the same as
1) sct.py::send_email logic depends on JOB_NAME. 
2) tricky logic around ClusterTester.get_email_data for proper work of making  email in PerformanceRegressionUpgradeTest we have to force it to return {}

```
class PerformanceRegressionUpgradeTest(PerformanceRegressionTest, UpgradeTest):  # pylint: disable=too-many-ancestors
    def get_email_data(self):  # pylint: disable=no-self-use
        return PerformanceRegressionTest.get_email_data(self)
```

all these 3 solutions make our life harder

roydahan:
Are we closing this one in favor of #6679? (I'm fine with it)

fruch:
> Are we closing this one in favor of #6679? (I'm fine with it)

yes

PR #6679, opened 2023-10-12
fix(perf-latency): use `email_subject_postfix` for subject

fruch:
as other perf test cases, we want some degree of control on the email report subjects, so we can tell between the test cases we have.

now each case can define how the test would be shown in emails

### DIFF
--- a/sdcm/results_analyze/__init__.py
+++ b/sdcm/results_analyze/__init__.py
@@ -226,7 +226,6 @@ class LatencyDuringOperationsPerformanceAnalyzer(BaseResultsAnalyzer):
         super().__init__(es_index=es_index, es_doc_type=es_doc_type, email_recipients=email_recipients,
                          email_template_fp="results_latency_during_ops_short.html", logger=logger, events=events)
         self.percentiles = ['percentile_90', 'percentile_99']
-        self.test_name_for_email_subject = 'latency during operations'
 
     def get_debug_events(self):
         return self.get_events(event_severity=[Severity.DEBUG.name])
@@ -407,7 +406,7 @@ class LatencyDuringOperationsPerformanceAnalyzer(BaseResultsAnalyzer):
         except Exception as exc:  # pylint: disable=broad-except
             LOGGER.error("Compare results failed: %s", exc)
 
-    def check_regression(self, test_id, data, is_gce=False, node_benchmarks=None):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements
+    def check_regression(self, test_id, data, is_gce=False, node_benchmarks=None, email_subject_postfix=None):  # pylint: disable=too-many-locals, too-many-branches, too-many-statements, too-many-arguments
         doc = self.get_test_by_id(test_id)
         full_test_name = doc["_source"]["test_details"]["test_name"]
         test_name = full_test_name.split('.')[-1]  # Example: longevity_test.LongevityTest.test_custom_time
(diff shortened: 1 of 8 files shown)


test-cases/performance/perf-regression-latency-650gb-grow-shrink.yaml
```diff
@@ -35,3 +35,4 @@ send_email: true
 email_recipients: ["scylla-perf-results@scylladb.com"]
 use_prepared_loaders: true
 use_hdr_cs_histogram: true
+email_subject_postfix: 'latency during grow-shrink'
```
fgelcer, line 38: i can be "calculated" automatically, based on existing content, not need to add such parameter in this test

test-cases/performance/perf-regression-latency-650gb-with-nemesis.yaml
```diff
@@ -35,3 +35,4 @@ send_email: true
 email_recipients: ["scylla-perf-results@scylladb.com"]
 use_prepared_loaders: true
 use_hdr_cs_histogram: true
+email_subject_postfix: 'latency during operations'
```
fgelcer, line 38: i can be "calculated" automatically, based on existing content, not need to add such parameter in this test

sdcm/results_analyze/__init__.py
```diff
@@ -448,7 +447,7 @@ def check_regression(self, test_id, data, is_gce=False, node_benchmarks=None):
         dataset_size = re.search(r'(\d{3}gb)', config_files).group() or 'unknown size'
 
         subject = (f'{self._get_email_tags(doc, is_gce)} Performance Regression Compare Results '
-                   f'({self.test_name_for_email_subject} {dataset_size}) -'
+                   f'({email_subject_postfix} {dataset_size}) -'
```
fgelcer, line 450: with this change, `latency during upgrade` is now broken

fgelcer (changes requested):
doesn't address what #6678 suggests to fix, but it could be added, if exist, instead of `self.test_name_for_email_subject` perhaps, and not instead

test-cases/performance/perf-regression-latency-650gb-with-nemesis.yaml
fruch: I don't want anymore logic spread across multiple places

sdcm/results_analyze/__init__.py
fruch: move reasons to move it away

test-cases/performance/perf-regression-latency-650gb-with-nemesis.yaml
fgelcer: you don't want logic spread, so let's duplicate parameters in an even more spreaded places

sdcm/results_analyze/__init__.py
fruch: this line is now removed, and moved to configuration.

test-cases/performance/perf-regression-latency-650gb-with-nemesis.yaml
fruch: yes, to the configuration, as it is in other perf test

fgelcer: this specific test (and all of its forms - 3 of them right now) are very different from the other tests, and as such, they use a different report (and a different template), so i don't think it is the right approach...
anyway, this is IMHO, and you are the maintainer, and now the responsible for it.. i don't like your approach, because we are having a unncessary parameter in addition to the existing ones

soyacz: I like the new approach - we can reuse more code. Email subject is something that should be configurable from configuration file (at least in some part like prefix).
Now we can get rid of `LatencyDuringUpgradesPerformanceAnalyzer` which was created only to change email subject.

fruch: Yeah you have a point there.

I think that which analyzer is used should also be in configuration,

Since we have code selecting them based on job names during the send email stage

fruch: @soyacz removed the unneeded class

soyacz (approved):
LGTM

PR #6691, opened 2023-10-17
test(k8s-func): add func test with quasi multiDC setup

vponomaryov:
The `v1.11.0*` scylla-operator allows to configure manual multiDC DB cluster
using the new `externalSeeds` config option where we can specify IP addresses of the nodes from another DC/cloud.

So, cover it's basic functionality in the `quasi-multidc` functional K8S test by making a DB cluster out of the 2 separate `ScyllaCluster` K8S resources created by the same Scylla-operator.

### DIFF
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -12,6 +12,8 @@
 # See LICENSE for more details.
 #
 # Copyright (c) 2021 ScyllaDB
+
+#  pylint: disable=too-many-lines
 import logging
 import os
 import random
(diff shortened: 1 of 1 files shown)


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/258/consoleFull
The failure is expected - https://github.com/scylladb/scylla-operator/issues/1484
Special attention to the K8S logs.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -87,6 +93,128 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
...
+            "get svc --no-headers -o=custom-columns=':.spec.clusterIP'"
+            f" -l scylla/cluster={cluster_name} -l scylla-operator.scylladb.com/scylla-service-type=member",
+            namespace=namespace).stdout.split()
+        assert svc_ips
+        assert len(svc_ips) == 3
+        assert all(svc_ip in ('', 'None') for svc_ip in svc_ips), "SVC IPs were expected to be absent"
+
+        # NOTE: read Scylla pods IPs
+        pods_data = get_pod_names_and_ips(cluster_name=cluster_name, namespace=namespace)
+        try:
```
soyacz, line 166: IIUC one try-catch block would be enough if cluster installation would be tracked in some variable.
Another approach could be detection of namespaces - if namespace exists, delete cluster and namespace.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -87,6 +93,128 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
         f"Pods: {yaml.safe_dump(pods_with_wrong_image_tags, indent=2)}")
 
 
+@pytest.mark.required_operator("v1.11.0")
+def test_deploy_quasi_multidc_db_cluster(db_cluster: ScyllaPodCluster):  # pylint: disable=too-many-locals,too-many-statements,too-many-branches
+    """
+    Deploy 2 ScyllaCluster objects and combine them into a single DB cluster.
```
soyacz, line 99: IIUC would be worth to add info that quasi-multidc means one cluster but two namespaces (and using PodIPs for connectivity as you already mentioned)

Also add info what is tested here - verify peers for each node lists nodes from both dc's.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -87,6 +93,128 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
...
+            log.debug(db_cluster.k8s_cluster.helm_install(
+                target_chart_name=target_chart_name2, source_chart_name="scylla-operator/scylla",
+                version=operator_version, use_devel=True, values=values, namespace=namespace2))
+            db_cluster.k8s_cluster.kubectl_wait(
+                "--all --for=condition=Ready pod", namespace=namespace2, timeout=1200)
+            pods_data |= get_pod_names_and_ips(cluster_name=cluster_name2, namespace=namespace2)
+            all_pod_ips = {}
+            for current_namespace in pods_data.keys():
+                for pod_ip in pods_data[current_namespace].values():
+                    all_pod_ips[pod_ip] = dc_name + "" if current_namespace == namespace else "-2"
```
soyacz, line 183: `ip_to_dc_map`?

functional_tests/scylla_operator/test_functional.py
```diff
@@ -87,6 +93,128 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
...
+        svc_ips = kubectl(
+            "get svc --no-headers -o=custom-columns=':.spec.clusterIP'"
+            f" -l scylla/cluster={cluster_name} -l scylla-operator.scylladb.com/scylla-service-type=member",
+            namespace=namespace).stdout.split()
+        assert svc_ips
+        assert len(svc_ips) == 3
+        assert all(svc_ip in ('', 'None') for svc_ip in svc_ips), "SVC IPs were expected to be absent"
+
+        # NOTE: read Scylla pods IPs
+        pods_data = get_pod_names_and_ips(cluster_name=cluster_name, namespace=namespace)
```
soyacz, line 165: `namespace_pod_to_ip_map`?

functional_tests/scylla_operator/libs/helpers.py
```diff
@@ -41,6 +41,14 @@ class PodStatuses(Enum):
     CRASH_LOOP_BACK_OFF = 'CrashLoopBackOff'
 
 
+def create_namespace(db_cluster: ScyllaPodCluster, namespace: str) -> None:
+    log.info("Create '%s' namespace", namespace)
+    namespaces = yaml.safe_load(db_cluster.k8s_cluster.kubectl("get namespaces -o yaml").stdout)
+    if not [ns["metadata"]["name"]
+            for ns in namespaces["items"] if ns["metadata"]["name"] == namespace]:
+        db_cluster.k8s_cluster.kubectl(f"create namespace {namespace}")
```
soyacz, line 49: I think for debugging purpose, would be good to add warning log that namespace exists and skipping creation.

soyacz (approved):
some remarks, but overall LGTM

functional_tests/scylla_operator/libs/helpers.py
```diff
@@ -41,6 +41,14 @@ class PodStatuses(Enum):
     CRASH_LOOP_BACK_OFF = 'CrashLoopBackOff'
 
 
+def create_namespace(db_cluster: ScyllaPodCluster, namespace: str) -> None:
```
fruch, line 44: don't we have this help code already somewhere inside the k8s cluster implementation ?

functional_tests/scylla_operator/test_functional.py
```diff
@@ -87,6 +93,128 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
...
+            for current_namespace in pods_data.keys():
+                for pod_ip in pods_data[current_namespace].values():
+                    all_pod_ips[pod_ip] = dc_name + "" if current_namespace == namespace else "-2"
+
+            log.info("Verify DB cluster's peers info")
+            for current_namespace in pods_data.keys():
+                for pod_name, pod_ip in pods_data[current_namespace].items():
+                    cqlsh_cmd = "select peer, data_center, rack, rpc_address from system.peers;"
+                    cqlsh_output = parse_cqlsh_select_output(kubectl(
+                        f"exec {pod_name} -- /bin/cqlsh -e \"{cqlsh_cmd}\"",
```
fruch, line 190: we are doing this since we don't have a node object for that those ones ? i.e. using cqlsh like this ?

functional_tests/scylla_operator/libs/helpers.py
```diff
@@ -230,3 +238,22 @@ def verify_resharding_on_k8s(db_cluster: ScyllaPodCluster, cpus: Union[str, int,
         assert not liveness_probe_failures_list, (
             f"liveness probe has failures: {liveness_probe_failures_list}")
     log.info("Resharding has successfully ended on whole Scylla cluster.")
+
+
+# TODO: add support for the single column output which doesn't have the "|" symbol
+def parse_cqlsh_select_output(cqlsh_output: str):
```
fruch, line 244: this seems like a general purpose helper for using cqlsh, and not something specific to the functional tests, we should consider move it to it's own module under sdcm.utils

functional_tests/scylla_operator/libs/helpers.py
fruch: after reading the rest of the PR, I can guess why we need a separate implementation of it.
but we should still write it down, why we have this separated from the k8s cluster

sdcm/logcollector.py
```diff
@@ -1204,6 +1204,7 @@ class KubernetesLogCollector(BaseSCTLogCollector):
         FileLog(name='haproxy_ingress.log', search_locally=True),
         FileLog(name='*_cluster_events.log', search_locally=True),
         FileLog(name='kubectl.version', search_locally=True),
+        DirLog(name='test_deploy_quasi_multidc_db_cluster/*', search_locally=True),
```
fruch, line 1207: what logs do we have there ? and why do we need test specific collector here ?

functional_tests/scylla_operator/libs/helpers.py
soyacz: also to simplify, could use `select json` instead parsing lines like this.

sdcm/logcollector.py
vponomaryov: In the new test we create and delete resources which get cleaned up at the end of the test.
So, we gather narrow set of K8S logs with the namespaces and their resources which are going to be deleted after it.
It is needed for debug. Without gathering of the logs we need either to keep resources around of gather logs at that time. And I chose second variant, because first one makes no sense.

functional_tests/scylla_operator/test_functional.py
vponomaryov: Yes, for now the approach we have for creating DB cluster and DB node objects is not flexible.
So, I used this way of getting data.

functional_tests/scylla_operator/test_functional.py
vponomaryov: Ok, will change.

functional_tests/scylla_operator/test_functional.py
vponomaryov: Usage of variable will also need a condition and indentation.
So, I don't see any benefit.
Current approach is readable enough, IMHO.

functional_tests/scylla_operator/libs/helpers.py
vponomaryov: @soyacz Indeed! I will use the `select json`. Thanks

functional_tests/scylla_operator/libs/helpers.py
vponomaryov: Ok. Will add

functional_tests/scylla_operator/libs/helpers.py
vponomaryov: I can add it as method for the K8S cluster.

functional_tests/scylla_operator/test_functional.py
vponomaryov: `namespace_pod_to_ip_map` is not really correct if follow this logic.
It may be, using this logic, following: `namespace_pod_name_to_ip_map`.
But it is still ambiguous. Name doesn't mean we have hierarchy.
So, I tend to leave it `as is`.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -12,7 +12,10 @@
 # See LICENSE for more details.
 #
 # Copyright (c) 2021 ScyllaDB
+
+#  pylint: disable=too-many-lines
```
fruch, line 16: maybe it's good sign to start breaking it into more files...

functional_tests/scylla_operator/test_functional.py
fruch: o.k.

functional_tests/scylla_operator/test_functional.py
```diff
@@ -87,6 +91,130 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
...
+                    assert len(table_rows) == 5, "Expected 5 peers"
+                    for row in table_rows:
+                        assert row["peer"] == row["rpc_address"]
+                        assert row["peer"] != pod_ip
+                        assert row["peer"] in ip_to_dc_map
+                        assert row["rack"] == db_cluster.k8s_cluster.params.get('k8s_scylla_rack')
+                        assert row["data_center"] == ip_to_dc_map[row["peer"]]
+            need_to_collect_logs = False
+        finally:
+            if need_to_collect_logs:
```
fruch, line 201: I would have put the cleanups into a test finalizers instead of try/finally , but it's not that critical

functional_tests/scylla_operator/test_functional.py
fruch: as I've wrote blow
I would have put the cleanups into a test finalizers instead of try/finally , but it's not that critical

functional_tests/scylla_operator/test_functional.py
fruch: both are fine, it's a local function in the test, and hopefully won't be using it in any other place

fruch (changes requested):
LGTM, but I would ask for a comment on the log collection part, 
inside the test we can fuss around as we please, but on the other part we should document a bit more.

functional_tests/scylla_operator/test_functional.py
vponomaryov: Pytest has bug where finalizers don't have strict order: https://github.com/pytest-dev/pytest/issues/8489
So, it doesn't fit our case. We should, at first, gather logs, and only then delete resources. Not vice-versa.

functional_tests/scylla_operator/test_functional.py
vponomaryov: Probably. 
But it requires decision about the criteria for splitting.
First thought that comes into my mind is following:
- common (majority of tests)
- mgmt
- cloud-bundle -specific

This new test is not really different than the existing `test_deploy_helm_with_default_values` one.

fruch: No, we should split it to more files:
- sanity/basic
- multiple clusters
- reproducers

in theory we can have one file per test even... we don't have they many

fruch: and yes it's not blocking

functional_tests/scylla_operator/test_functional.py
fruch: why do you need ordering ?, from the looks of it you are collecting logs from two different scylla cluster
why do you care who will be first and who will be last ?

vponomaryov: Ordering between logs gathering and cleanup, both will need to go to finalizers.

vponomaryov:
The latest uploaded PR state has passed here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/262/consoleFull

sdcm/logcollector.py
vponomaryov: Done

vponomaryov:
Unit tests failed with the unrelated breakage which gets fixed here: https://github.com/scylladb/scylla-cluster-tests/pull/6700

vponomaryov:
Update: some flakiness is observed in the attempts to use `cqlsh`. Sometimes commands just fail.
Example: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/263/consoleFull
Will investigate more tomorrow.

vponomaryov:
Updated the PR making it have only one commit, because another, complementary one, has already been merged here: https://github.com/scylladb/scylla-cluster-tests/pull/6732
The new functional test works itself, but it causes some bad side effect on following tests which try to provision 4th pods.
So, will file a bug about it.

soyacz (approved):
LGTM

vponomaryov:
It went out to be configuration inconsistency with the setup we use.
Or in other words. This PR is totally unrelated to the above mentioned problems.
Here is the fix for the `test_add_new_node_and_check_old_nodes_are_cleaned_up` functional test that fails when we add this test, which just changes timing.
The fix for the `Scylla cleanup` pods scheduling is here: https://github.com/scylladb/scylla-cluster-tests/pull/6749

vponomaryov:
Just rebased after merge of the https://github.com/scylladb/scylla-cluster-tests/pull/6750 which had overlapping code changes.

@fruch , @soyacz do you still have some objections for it?

From my side it is ready to merged.

soyacz (approved):
LGTM

PR #6695, opened 2023-10-18
fix(microbenchmark): switch ES index

soyacz:
Recently cpu settings were changed for microbenchmark tests: https://github.com/scylladb/scylla-pkg/pull/3618
Because of that, historical perf results are not relevant.

Switch ES index so we make comparisons to results with the same CPU settings.

### DIFF
--- a/sdcm/microbenchmarking.py
+++ b/sdcm/microbenchmarking.py
@@ -43,7 +43,7 @@ setup_stdout_logger()
 LOGGER = logging.getLogger("microbenchmarking")
 LOGGER.setLevel(logging.DEBUG)
 
-MICROBENCHMARK_INDEX_NAME = 'microbenchmarkingv2'
+MICROBENCHMARK_INDEX_NAME = 'microbenchmarkingv3'
 
 
 @contextlib.contextmanager


sdcm/microbenchmarking.py
```diff
@@ -43,7 +43,7 @@
 LOGGER = logging.getLogger("microbenchmarking")
 LOGGER.setLevel(logging.DEBUG)
 
-MICROBENCHMARK_INDEX_NAME = 'microbenchmarkingv2'
```
fruch, line 46: @temichus is the new type of microbenchmarks the one running withing SCT, share this index ? or they are on a separate index ?

soyacz: @temichus ^^

roydahan: I'm not sure I fully understand the question, but from looking in elasticsearch, all the recent microbnechmarks results are in "microbenchmarkingv2" index.

fruch: I was talking about the `PerfSimpleQueryAnalyzer`

and seems like it's using a different index coming for configuration:
```
custom_es_index: 'microbenchmarkingtest'
```

but with a very similar name, I was asking to make sure we don't reset those as well at the same time.

Issue #6688, opened 2023-10-17
Collect system.log on db instance eviction


soyacz:
When we decommision scylla node we evict VM instance. 
We collect `system.log` from db nodes and because VM's of decommisioned nodes don't exist, we skip it.
But this log is valuable (have proper timestamps and is complete).

Task is about collecting it before instance eviction.

SCT PR:

- https://github.com/scylladb/scylla-cluster-tests/pull/6696

roydahan:
We need to think what to do in the case when we terminate a node and expect it to be almost immediate operation.
If it adds a significant amount of time to collect the log from that instance, it may affect the tests and probably many of the stats we collect.

fruch:
> We need to think what to do in the case when we terminate a node and expect it to be almost immediate operation. If it adds a significant amount of time to collect the log from that instance, it may affect the tests and probably many of the stats we collect.

collection of single node, shouldn't be that long.
it's just a handful of ssh commands, we are not uploading it to s3 at that stage. (as far as I understand)
we should measure it as we do this change.

anyhow we can just do the system.log part, if doing all of the collection is too much time.

fruch:
Also if we have code that assume that taking a node down is immediately, that code part should be fixed (I don't think we have such delicate code)

fruch:
@soyacz was trying it out in https://github.com/scylladb/scylla-cluster-tests/pull/6696

but for now we are dropping it, it's a bit more complicated than we estimated

fruch:
@soyacz 

this was raise again in the context of doing scylla code coverage runs with SCT, and that we'll need to make sure we dump and save the coverage information whenever we stop/kill a node.

soyacz:
where coverage information is stored?

fruch:
> where coverage information is stored?

we didn't yet have runs with it on SCT, we were discussing it with @eliransin, that recently pushed all of the support for that into scylla core.

it would be wherever we'll point it to (on the VM itself)

roydahan:
Let's try to have a quick implementation only for coverage collection.
It's a 60MB file, should be fairly quick.

(We also need to make sure we dump the metrics before any violent kill (hard reboot, kill -9, etc), but that should be a different task).

PR #6696, opened 2023-10-18
improvement(logs): collect db logs upon VM termination

soyacz:
We need `system.log` to be collected for decommissioned/removed nodes for bugs investigation - this log contain proper timestamps. Also other collected usually files might be useful.

Collect logs like we collect them after tests for terminated VM's.

fixes: #6688

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -53,6 +53,7 @@ from cassandra.policies import WhiteListRoundRobinPolicy, HostFilterPolicy, Roun
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
 from argus.backend.util.enums import ResourceState
+from argus.client.sct.types import LogLink
 from sdcm.node_exporter_setup import NodeExporterSetup
 from sdcm.db_log_reader import DbLogReader
 from sdcm.mgmt import AnyManagerCluster, ScyllaManagerError
@@ -121,7 +122,7 @@ from sdcm.sct_events.filters import EventsSeverityChangerFilter
 from sdcm.utils.auto_ssh import AutoSshContainerMixin
 from sdcm.monitorstack.ui import AlternatorDashboard
 from sdcm.logcollector import GrafanaSnapshot, GrafanaScreenShot, PrometheusSnapshots, upload_archive_to_s3, \
-    save_kallsyms_map, collect_diagnostic_data
+    save_kallsyms_map, collect_diagnostic_data, ScyllaLogCollector
 from sdcm.utils.ldap import LDAP_SSH_TUNNEL_LOCAL_PORT, LDAP_BASE_OBJECT, LDAP_PASSWORD, LDAP_USERS, \
     LDAP_PORT, DEFAULT_PWD_SUFFIX
 from sdcm.utils.remote_logger import get_system_logging_thread
@@ -3421,6 +3422,12 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
             ))
         if node in self.nodes:
             self.nodes.remove(node)
+        ScyllaLogCollector.cluster_log_type = node.name
+        log_links = ScyllaLogCollector([node],
+                                       test_id=self.test_config.test_id(),
+                                       storage_dir=os.path.join(self.logdir, "collected_logs"),
+                                       params=self.params).collect_logs()
+        self.test_config.argus_client().submit_sct_logs([LogLink(log_name=node.name, log_link=log_links[0])])
         node.destroy()
 
     def get_db_auth(self):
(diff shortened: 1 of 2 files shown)


soyacz:
only draft until I verify it works

soyacz:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-mini-test-1h-test/32/

sdcm/cluster.py
```diff
@@ -3421,6 +3421,10 @@ def terminate_node(self, node: BaseNode, scylla_shards: int = 0) -> None:
             ))
         if node in self.nodes:
             self.nodes.remove(node)
+        ScyllaLogCollector([node],
```
fruch, line 3424: this would is one BaseCluster, so this would happen for all type of nodes...
we probably would want to limit it to scylla nodes only.

fruch: on the other side of it, we never call `terminate_node` for loader and monitor nodes...
and I can't think of a case we might do that, but still I would try to limit it, and make it clearer what it's aimed for.

sdcm/cluster.py
```diff
@@ -3421,6 +3422,12 @@ def terminate_node(self, node: BaseNode, scylla_shards: int = 0) -> None:
             ))
         if node in self.nodes:
             self.nodes.remove(node)
+        ScyllaLogCollector.cluster_log_type = node.name
+        log_links = ScyllaLogCollector([node],
+                                       test_id=self.test_config.test_id(),
+                                       storage_dir=os.path.join(self.logdir, "collected_logs"),
+                                       params=self.params).collect_logs()
+        self.test_config.argus_client().submit_sct_logs([LogLink(log_name=node.name, log_link=log_links[0])])
```
fruch, line 3430: why do we need to report to argus at this point ? isn't enough that we have the files in place, for the collection phase to pick them up ?

i.e. i'm not sure we want to upload to s3 at this point...

soyacz: I wanted to use existing means of collecting logs which does it by default. I'm not sure yet if this is feasible yet.
Collection phase is not collecting them if collected in current form anyway, as collection dir does not match.

The one positive point about publishing it to s3 is that logs get removed right away from the sct-runner. For single node it's not taking long (db logs tar.gz usually is less than 10MB's).

sdcm/cluster.py
soyacz: here we specify just one node - and it's not collecting other db nodes logs because we don't provide `local_search_path` to `collect_logs` method

sdcm/cluster.py
```diff
@@ -3421,6 +3422,12 @@ def terminate_node(self, node: BaseNode, scylla_shards: int = 0) -> None:
             ))
         if node in self.nodes:
             self.nodes.remove(node)
+        ScyllaLogCollector.cluster_log_type = node.name
```
fruch, line 3425: I'm not sure what this is trying todo ... have a unique type ? I'm quite sure we have a list somewhere of the types, and this won't fit.

soyacz:
this is how it looks in this form: https://argus.scylladb.com/test/c38d2050-1e12-479a-8942-40cfaae44859/runs?additionalRuns[]=9b1cdaef-2024-4c1f-8e6a-db8aecc9697e
basically, each evicted node generates log entry in argus for future download.

I can prepare some alternative solution with completely new approach (e.g. by downloading logs to db node directory, so it will be collected hopefully by collect-logs step).

WDYT?

fruch:
> this is how it looks in this form: [https://argus.scylladb.com/test/c38d2050-1e12-479a-8942-40cfaae44859/runs?additionalRuns[]=9b1cdaef-2024-4c1f-8e6a-db8aecc9697e](https://argus.scylladb.com/test/c38d2050-1e12-479a-8942-40cfaae44859/runs?additionalRuns%5B%5D=9b1cdaef-2024-4c1f-8e6a-db8aecc9697e) basically, each evicted node generates log entry in argus for future download.

for one or two nodes, it's looks o.k., if you gonna have a dosen, i.e. in a 3-4 days run, it gonna be quite annoying 
> 
> I can prepare some alternative solution with completely new approach (e.g. by downloading logs to db node directory, so it will be collected hopefully by collect-logs step).

I like this approach a bit better, assuming we don't know to rewrite all o the collectors
I would put it into a specific folder, and upload everything there on collection 

> 
> WDYT?

soyacz:
I was trying to download all the logs to db node directory for future collection during collect-logs stage. But this is not trivial as `system.log` already exists in node directory (soft link to `messages.log`). When collecting from existing node at collect-logs stage, it is downloaded correctly due some fancy side effect.
Also our code is not adapted to collecting `CommandLog` when node does not exist.

So, needs some more work if we want doing it this way. For now, I freeze development of this feature due to higher priority tasks.

fruch:
I think we can live with this implementation for a while

We just need to time in longer runs, like the schema-topoloy case if tier1

If it's short enough, I think it's acceptable.

And have lots of extra files reported in Argus, maybe we can collapse them in Argus or something like that

soyacz:
@fruch, problem with missing logs was raised by devs recently, can we continue this one until something better will be implemented?

soyacz:
closing this in favor of https://github.com/scylladb/scylla-cluster-tests/pull/8735 (cherry pick of first commit from this PR)

PR #6706, opened 2023-10-23
feature(cluster_replacement): add cluster replacement tests

Mark-Gurevich:
feature(cluster_replacement): add cluster replacement feature tests file
    
This PR adds cluster replacement feature tests
According to task: https://github.com/scylladb/qa-tasks/issues/1164
Rolling cluster replacement with Multi-DC and Multi-AZ
Adding a new node and then decommission one node.

### DIFF
--- a/full_cluster_replacement_test.py
+++ b/full_cluster_replacement_test.py
@@ -1,0 +1,96 @@
+import logging
+from longevity_test import LongevityTest
+from sdcm import prometheus
+from sdcm.sct_events.system import InfoEvent
+from sdcm.utils import loader_utils
+
+LOGGER = logging.getLogger(__name__)
+
+
+class FullClusterReplacement(LongevityTest, loader_utils.LoaderUtilsMixin):
+    """
+    The purpose of these tests is replacement of all nodes in the ScyllaDB.
+    Cluster Replacement QA task: https://github.com/scylladb/qa-tasks/issues/1164
+    Cluster replacement test plan: https://docs.google.com/document/d/1aKCTs6SiH9ORDYvkAQ8upOpVwZqaxsW9VfbJqKG63g0/edit
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.metrics_srv = prometheus.nemesis_metrics_obj()
+        self.stress_queue = []
+
+    def start_load(self):
+        keyspace_num = self.params.get('keyspace_num')
+        stress_cmd = self.params.get('stress_cmd')
+        # Start the load
+        self.run_pre_create_schema()
+        self.run_prepare_write_cmd()
+        # self.run_post_prepare_cql_cmds() Uncomment when MV solved: Ref: https://github.com/scylladb/scylladb/issues/15717
+        self.assemble_and_run_all_stress_cmd(self.stress_queue, stress_cmd, keyspace_num)
+
+    def verify_stress_threads_after_node_replacement(self):
+        for stress in self.stress_queue:
+            self.verify_stress_thread(cs_thread_pool=stress)
+
+    def add_new_node_then_decommission_one_node(self, node_to_be_removed, dc_idx=0, new_instance_type=None):
+        """
+        This method use to replaces a node in a cluster by 'Add a new node to the cluster and then decommission the old node' procedure # pylint: disable=line-too-long
+        It adds a new node to the cluster, waits for initialization, decommissions the specified node,
+        and reconfigures the monitoring system.
+        The procedure from: https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/replace-running-node.html#add-a-new-node-to-the-cluster-and-then-decommission-the-old-node # pylint: disable=line-too-long
+        """
+        self.metrics_srv.event_start('add_node')
+        new_node = \
+            self.db_cluster.add_nodes(count=1, dc_idx=dc_idx, enable_auto_bootstrap=True, instance_type=new_instance_type)[
+                0]
+        self.db_cluster.wait_for_init(node_list=[new_node], timeout=3600, check_node_health=False)
+        self.metrics_srv.event_stop('add_node')
+        self.monitors.reconfigure_scylla_monitoring()
+
+        # Decommission node to be replaced
+        InfoEvent(message='Decommission node: {}'.format(node_to_be_removed)).publish()
+        self.db_cluster.decommission(node_to_be_removed)
+        for node in self.db_cluster.nodes:
+            node.run_nodetool(sub_cmd='cleanup')
+
+    def test_rolling_full_cluster_replacement_add_then_decommission_nodes(self):
+        InfoEvent(message="Starting cluster replacement test by add then decommission nodes...").publish()
+        self.start_load()
+        new_instance_type = self.db_cluster.params.get('new_instance_type')
+        # Run node replacement procedure for all nodes and verify data
+        for node in self.db_cluster.nodes:
+            self.add_new_node_then_decommission_one_node(node_to_be_removed=node, new_instance_type=new_instance_type)
+            self.verify_stress_threads_after_node_replacement()
+
+        InfoEvent(message="Finished cluster replacement test by add then decommission nodes...").publish()
+
+    def test_multidc_rolling_full_cluster_replacement_add_then_decommission_nodes(self):
+        new_instance_type = self.params.get('new_instance_type')
+
+        InfoEvent(message="Starting multi-dc cluster replacement test by add then decommission nodes "
+                          "with repair and major compaction in the end...").publish()
+        self.start_load()
+
+        # Run node replacement procedure for all nodes and verify data
+        number_of_nodes_in_each_dc = int(len(self.db_cluster.nodes)/2)
+        for i, node in enumerate(self.db_cluster.nodes):
+            if i >= number_of_nodes_in_each_dc:
+                dc_idx = 0
+            else:
+                dc_idx = 1
+            self.add_new_node_then_decommission_one_node(
+                node_to_be_removed=node, dc_idx=dc_idx, new_instance_type=new_instance_type)
+            self.verify_stress_threads_after_node_replacement()
+
+        # Full Repair and major compaction
+        # This is not required for the replacement procedure, but can be a part of the test to check repair and
+        # compaction on a replaced cluster.
+        for node in self.db_cluster.nodes:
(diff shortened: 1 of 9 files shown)


Mark-Gurevich:
The test passed and I've verified nodes replacement took place.
Data load was copied from individual nemesis.
https://jenkins.scylladb.com/job/scylla-staging/job/mark/job/features-full-cluster-replacement-test/10/

Mark-Gurevich:
@scylladb/qa-maintainers Please review this PR.

full_cluster_replacement_test.py
```diff
@@ -0,0 +1,64 @@
...
+from sdcm import prometheus
+from sdcm.sct_events.system import InfoEvent
+from sdcm.utils import loader_utils
+
+LOGGER = logging.getLogger(__name__)
+
+
+class FullClusterReplacement(LongevityTest, loader_utils.LoaderUtilsMixin):
+    """These tests should use multi data center (multi-DC) and multi Availability Zone (multi-AZ)
+    TODO: More info needs to be added here"""
```
fruch, line 12: Links to relevet issues

And few words on the goal of this test

test-cases/features/full-cluster-replacement-multidc.yaml
```diff
@@ -0,0 +1,22 @@
+test_duration: 360
+
+prepare_write_cmd:  ["cassandra-stress write cl=QUORUM n=5048570 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=80 -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5",
```
fruch, line 3: The task talks about a much bigger data set, 5gb is kind of trivial. and is enough to just test the code flow

fruch (dismissed):
Looks o.k. but not clear why we don't define this flow as nemesis

full_cluster_replacement_test.py
```diff
@@ -0,0 +1,64 @@
...
+from longevity_test import LongevityTest
+from sdcm import prometheus
+from sdcm.sct_events.system import InfoEvent
+from sdcm.utils import loader_utils
+
+LOGGER = logging.getLogger(__name__)
+
+
+class FullClusterReplacement(LongevityTest, loader_utils.LoaderUtilsMixin):
+    """These tests should use multi data center (multi-DC) and multi Availability Zone (multi-AZ)
```
soyacz, line 11: you didn't set multi-AZ

soyacz (commented):
from where this procedure was taken from? according to scylla docs, cleanup should be performed: https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/replace-running-node.html#add-a-new-node-to-the-cluster-and-then-decommission-the-old-node

test-cases/features/full-cluster-replacement-multidc.yaml
Mark-Gurevich: Ok so how much we want for this?
I think we need specific definitions for all the test cases in dtest/SCT/Nemesis in the [test plan](https://docs.google.com/document/d/1aKCTs6SiH9ORDYvkAQ8upOpVwZqaxsW9VfbJqKG63g0/edit) and [test cases](https://docs.google.com/spreadsheets/d/1nWNU50dKVz12use1VSClwonbcCwQPmnuk0OIPg589ec/edit#gid=0) docs.
The docs are still in draft state and need to be improved.

Mark-Gurevich:
> Looks o.k. but not clear why we don't define this flow as nemesis

We will create a nemesis for this flow with some restriction for dataset and cluster size.
This feature test file should grow with additional tests with different parameters for cluster size, raft, dataset and others foloowing guide lines described in the [test plan](https://docs.google.com/document/d/1aKCTs6SiH9ORDYvkAQ8upOpVwZqaxsW9VfbJqKG63g0/edit).

fruch:
> > Looks o.k. but not clear why we don't define this flow as nemesis
> 
> We will create a nemesis for this flow with some restriction for dataset and cluster size.
> This feature test file should grow with additional tests with different parameters for cluster size, raft, dataset and others following guide lines described in the [test plan](https://docs.google.com/document/d/1aKCTs6SiH9ORDYvkAQ8upOpVwZqaxsW9VfbJqKG63g0/edit).

I'm not sure I understand the reasoning in the plan, but isn't this flow is something we want to test across all cases ? And just create few cases that would run this specific nemesis with the data set and parameters needed ?

This look like we are developing the flow as test first, just to copy it later into a nemesis..., and that a recipe for things to get out of sync.

full_cluster_replacement_test.py
fruch: From the plan sounds like the answer is yes

test-cases/features/full-cluster-replacement-multidc.yaml
fruch: The plan didn't define it, but from my POV

There are two things to consider

1) you want a basic case like the individual nemesis to make sure flow is o.k. (again coming back to why nemesis kind of fit this better), and for that you don't need a very big data set

2) achieving the general goal of the plan, it is properly gonna be 6-9 nodes (multi DC, multi az), and a big data set, how big, I don't know it depends on the context of that task/plan, which I don't have.

We have tests with data sets from 100gb to 1tb

5gb is just a starting position we set for validation of nemesis

Maybe @roydahan has more context about the goal of this goal

full_cluster_replacement_test.py
Mark-Gurevich: added

full_cluster_replacement_test.py
```diff
@@ -0,0 +1,67 @@
...
+        number_of_nodes_in_each_dc = int(len(self.db_cluster.nodes)/2)
+        for i, node in enumerate(self.db_cluster.nodes):
+            if i >= number_of_nodes_in_each_dc:
+                dc_idx = 0
+            else:
+                dc_idx = 1
+            self.replace_node(node_to_be_removed=node, dc_idx=dc_idx)
+            self.verify_stress_threads_after_node_replacement()
+
+        # Full Repair and major compaction
```
soyacz, line 61: where this need comes from? Why there's no cleanup?

Mark-Gurevich: Added cleanup to the `replace_node` method.
This came from Fabio's request in the test plan.

roydahan: Worth to mention in a comment that this is not a must in terms of the replacement procedure.
It's another part of the test to test a full repair and major compaction on a replaced cluster.

full_cluster_replacement_test.py
```diff
@@ -0,0 +1,74 @@
...
+    Cluster Replacement QA task: https://github.com/scylladb/qa-tasks/issues/1164
+    Cluster replacement test plan: https://docs.google.com/document/d/1aKCTs6SiH9ORDYvkAQ8upOpVwZqaxsW9VfbJqKG63g0/edit
+    """
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.metrics_srv = prometheus.nemesis_metrics_obj()
+        self.stress_queue = []
+
+    def replace_node(self, node_to_be_removed, dc_idx=1):
```
roydahan, line 22: Please change the name of it to be "replace_node_by_add_and_decommission"

jenkins-pipelines/features-full-cluster-replacement.jenkinsfile
```diff
@@ -0,0 +1,14 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    availability_zone: 'a,b',
+    backend: 'aws',
+    region: '''["eu-west-1", "eu-west-2"]''',
+    test_name: 'full_cluster_replacement_test.FullClusterReplacement.test_rolling_full_cluster_replacement_add_then_remove_nodes',
```
roydahan, line 10: In all relevant places, it needs to be clear that replacement here is by adding and decommissioning.
(Which isn't really a node replacement...)

roydahan (changes requested):

full_cluster_replacement_test.py
Mark-Gurevich: ok

Mark-Gurevich:
> I'm not sure I understand the reasoning in the plan, but isn't this flow is something we want to test across all cases ? And just create few cases that would run this specific nemesis with the data set and parameters needed ?
> 
> This look like we are developing the flow as test first, just to copy it later into a nemesis..., and that a recipe for things to get out of sync.

Partially you are right.
There are many tests to be added here with specific configurations, not related to all the test cases which we want to test with a nemesis.

jenkins-pipelines/features-full-cluster-replacement.jenkinsfile
Mark-Gurevich: Done

full_cluster_replacement_test.py
Mark-Gurevich: Done

full_cluster_replacement_test.py
Mark-Gurevich: Added a comment

roydahan:
@Mark-Gurevich I suggest stop extending this PR so it won't be too big so we can merge it.

Mark-Gurevich:
> @Mark-Gurevich I suggest stop extending this PR so it won't be too big so we can merge it.

I agree. 
The nemeses require more changes so I'll remove them from this PR.

Mark-Gurevich:
I'm running the tests and adjusting dataset and instance types.
Israel asked to make the load much bigger than simple 5 GB.
Currently testing the "Basic test" with 100 GB `i4i.4xlarge` as it failed with end of space with `i4i.xlarge` instances.
The upscale test I will use smaller dataset 10 GB with instance type `i4i.2xlarge` upscale to `i4i.4xlarge`.

soyacz:
> I'm running the tests and adjusting dataset and instance types. Israel asked to make the load much bigger than simple 5 GB. Currently testing the "Basic test" with 100 GB `i4i.4xlarge` as it failed with end of space with `i4i.xlarge` instances. The upscale test I will use smaller dataset 10 GB with instance type `i4i.2xlarge` upscale to `i4i.4xlarge`.

i4i.xlarge should not fail with end of space with 100GB of data - something is misconfigured and requires some investigation.

fruch:
will reopen if we need it again

PR #6711, opened 2023-10-23
improvement(k8s): use actual value for the `fs.aio-max-nr` option

vponomaryov:
Scylla `5.0+` define the `fs.aio-max-nr` option to be `30000000`.
So, update our K8S code to align with it.

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -899,10 +899,10 @@ class KubernetesCluster(metaclass=abc.ABCMeta):  # pylint: disable=too-many-publ
                 "disabled": False,
                 "ingressClassName": "haproxy",
             }}}
-        # NOTE: '5578536' value is defined in the scylla repo here:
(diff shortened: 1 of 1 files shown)


vponomaryov:
I faced the need to update it running multiDC tests on EKS.
May fail in some other cases.
But anyway, we should update it because we use only 5.0+ Scylla versions, not older.

soyacz:
> I faced the need to update it running multiDC tests on EKS. May fail in some other cases. But anyway, we should update it because we use only 5.0+ Scylla versions, not older.

Do we have it in docs that this should be updated?

vponomaryov:
> > I faced the need to update it running multiDC tests on EKS. May fail in some other cases. But anyway, we should update it because we use only 5.0+ Scylla versions, not older.
> 
> Do we have it in docs that this should be updated?

It is not `required`.
But we set it to cover this functionality.

Docs: https://github.com/scylladb/scylla-operator/blob/master/docs/source/generic.md#configure-container-kernel-parameters

soyacz (approved):
LGTM

PR #6712, opened 2023-10-23
feature(argus): Gemini integration

k0machi:
This commit adds a way for SCT to submit gemini results to argus.
The results are collected during email data phase, with information
mirroring that of the email and additionally submitting the gemini seed
used.

Task: scylladb/qa-tasks#1481

### DIFF
--- a/docker/env/version
+++ b/docker/env/version
@@ -1,1 +1,1 @@
-1.49-update-urllib3
+1.50-gemini-argus
--- a/requirements.in
+++ b/requirements.in
@@ -29,7 +29,7 @@ mysql-connector-python==8.0.26
 docker==4.4.4  # can't use >=5 because of tcconfig==0.26.0
 python-jenkins==1.7.0
 ssh2-python==1.0.0
-argus-alm==0.11.4
+argus-alm==0.11.6
 parameterized==0.8.1
 pylint==2.11.1  # Needed for pre-commit hooks
 autopep8==1.5.7  # Needed for pre-commit hooks
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -548,6 +548,50 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         except Exception:  # pylint: disable=broad-except
             self.log.error("Error saving logs to Argus", exc_info=True)
 
+    def argus_collect_gemini_results(self):
+        try:
+            # pylint: disable=no-member
+            if not getattr(self, "gemini_results"):
+                return
+
+            if self.loaders:
+                gemini_version = self.loaders.gemini_version
+            else:
+                self.log.warning("Failed to get gemini version as loader instance is not created")
+                gemini_version = ""
+
+            seed = self.params.get("gemini_seed")
+            gemini_command, *_ = self.gemini_results["cmd"]
+            if not seed:
+                seed_match = re.search(r"--seed (?P<seed>\d+) ", gemini_command)
+                if seed_match:
+                    seed = seed_match.groupdict().get("seed", -1)
+                else:
+                    seed = -1
+
+            results = self.gemini_results["results"]
+            results = results[0] if len(results) > 0 else None
+            if not results or not isinstance(results, dict):
+                self.log.warning("Results object is not a dictionary: %s\nReplacing with empty dict.", results)
+                results = {}
+
+            self.test_config.argus_client().submit_gemini_results({
+                "gemini_command": "\n".join(self.gemini_results["cmd"]),
+                "gemini_read_errors": results.get("read_errors", -1),
+                "gemini_read_ops": results.get("read_ops", -1),
+                "gemini_seed": seed,
+                "gemini_status": self.gemini_results["status"],
+                "gemini_version": gemini_version,
+                "gemini_write_errors": results.get("write_errors", -1),
+                "gemini_write_ops": results.get("write_ops", -1),
+                "oracle_node_ami_id": self.params.get("ami_id_db_oracle"),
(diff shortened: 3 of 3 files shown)


k0machi:
[Staging job](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/alexey/job/gemini-3h-with-nemesis-test/2/)

gemini_test.py
```diff
@@ -138,6 +139,33 @@ def get_email_data(self):
...
+
+            # pylint: disable=no-member
+            self.test_config.argus_client().submit_gemini_results({
+                "gemini_command": self.gemini_results["cmd"],
+                "gemini_read_errors": self.gemini_results["results"].get("read_errors", -1),
+                "gemini_read_ops": self.gemini_results["results"].get("read_ops", -1),
+                "gemini_seed": seed,
+                "gemini_status": self.gemini_results["status"],
+                "gemini_version": gemini_version,
+                "gemini_write_errors": self.gemini_results["results"].get("write_errors", -1),
```
soyacz, line 159: I was thinking results are a list, does it work this way?

k0machi: I've looked into `results_gemini.html` and it appears to be a `dict` as template calls `.items()` on it.

fruch: @k0machi 

```
{% for res in results %}
```
there a loop on the template, I think @soyacz is correct.

gemini_test.py
```diff
@@ -138,6 +139,33 @@ def get_email_data(self):
...
+                "gemini_write_ops": self.gemini_results["results"].get("write_ops", -1),
+                "oracle_node_ami_id": self.params.get("ami_id_db_oracle"),
+                "oracle_node_instance_type": self.params.get("instance_type_db_oracle"),
+                "oracle_node_scylla_version": self.cs_db_cluster.nodes[0].scylla_version if self.cs_db_cluster else "N/A",
+                "oracle_nodes_count": self.params.get("n_test_oracle_db_nodes"),
+            })
+        except Exception:  # pylint: disable=broad-except
+            self.log.warning("Error submitting gemini results to argus")
```
fruch, line 167: please make sure we print the callstack here, otherwise it would be painful to debug.

BTW you last run ends here

gemini_test.py
```diff
@@ -138,6 +139,33 @@ def get_email_data(self):
             self.log.warning("Failed to get gemini version as loader instance is not created")
             gemini_version = ""
 
+        try:
```
fruch, line 142: I'm wondering why we are doing this on the gemini_test.py ? and no in a more central place ?
i.e. we have case that we might be running gemini of other cases, wouldn't we ?

gemini_test.py
k0machi: Yep, missed that one, added exception info.

gemini_test.py
k0machi: I'm not sure, I'll check it in this place first, but yeah, might want to put in a more appropriate place, should probably go into normal `ClusterTester` into teardown stage.

gemini_test.py
k0machi: Yep, didn't quite notice it, it's a list of dicts

gemini_test.py
k0machi: Moved to `ClusterTester` now, done in teardown stage after the email data call, so should be equivalent.

k0machi:
Results of the staging job [here](https://argus.scylladb.com/test/2e5ca785-f2de-4cc0-aa2f-7e21a9cdd809/runs?additionalRuns[]=59c98dd5-5790-4408-ae19-754ed9e241c9). Seems to be working.

soyacz:
How about renaming 'Gemini' section in Argus to 'Load' (inside load, having Gemini specified) so we could in future support also c-s, s-b and others.
Would be nice to have these info for other stress tools (including versions of the tools used)

fruch:
> How about renaming 'Gemini' section in Argus to 'Load' (inside load, having Gemini specified) so we could in future support also c-s, s-b and others. Would be nice to have these info for other stress tools (including versions of the tools used)

yes, and the commands, we should have it from the SCT configuration
c-s version is a canworms of it's own (i.e. both c-s and driver version used)

fruch (approved):
from SCT side, of providing the information this looks o.k.
on Argus end we might have small followups

soyacz (approved):

mykaul:
From https://jenkins.scylladb.com/job/scylla-5.4/job/upgrade-with-raft/job/rolling-upgrade-ubuntu22.04-with-raft-test/3/consoleFull#-1852483892fcc21424-66d2-4bd8-8e0d-9746405e5b16 : 

```
 < t:2023-10-30 11:43:43,577 f:tester.py       l:593  c:UpgradeTest          p:WARNING > Error submitting gemini results to argus < t:2023-10-30 11:43:43,577 f:tester.py       l:593  c:UpgradeTest          p:WARNING > Error submitting gemini results to argus
 < t:2023-10-30 11:43:43,577 f:tester.py       l:593  c:UpgradeTest          p:WARNING > Traceback (most recent call last):
 < t:2023-10-30 11:43:43,577 f:tester.py       l:593  c:UpgradeTest          p:WARNING >   File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 554, in argus_collect_gemini_results
 < t:2023-10-30 11:43:43,577 f:tester.py       l:593  c:UpgradeTest          p:WARNING >     if not getattr(self, "gemini_results"):
 < t:2023-10-30 11:43:43,577 f:tester.py       l:593  c:UpgradeTest          p:WARNING > AttributeError: 'UpgradeTest' object has no attribute 'gemini_results'. Did you mean: 'get_gemini_results'?
```

PR #6717, opened 2023-10-24
fix(rolling-upgrades): switch to using docker based loaders

fruch:
we are still using the c-s inside the loader image, and that is an older version of c-s and java-drivers, so we try again to update it to work with the docker based ones.

after this, we might fix the c-s to a spcific docker image of some release, to make sure it's the same version across the whole upgrade


fruch:
was created on wrong fork

PR #6718, opened 2023-10-24
fix(rolling-upgrades): switch to using docker based loaders

fruch:
we are still using the c-s inside the loader image, and that is an older version of c-s and java-drivers, so we try again to update it to work with the docker based ones.

after this, we might fix the c-s to a specific docker image of some release, to make sure it's the same version across the whole upgrade

### Testing
- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu20.04-test/8/ took 5h
- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu20.04-test/9/ took 5.5h

### DIFF
--- a/test-cases/upgrades/rolling-upgrade.yaml
+++ b/test-cases/upgrades/rolling-upgrade.yaml
@@ -39,6 +39,4 @@ gemini_seed: 66
 
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
 
-use_prepared_loaders: true
-
 use_preinstalled_scylla: false


roydahan:
Do we have a place where we can easily see which version of c-s we used?

fruch:
> Do we have a place where we can easily see which version of c-s we used and the driver?

1) no we don't, c-s doesn't print out it's version nor the driver version.
2) once docker is used, we check which docker image was used, checking sct.log for `docker.io` would tell you which image was pulled (the confusing part is, that that image would be pulled even when using the VM c-s)

fruch:
One job pass with this, I need to compare the duration, last time this check was causing those tests timeout

fruch:
> One job pass with this, I need to compare the duration, last time this check was causing those tests timeout, and we didn't found the actual root cause, and reverted the docker based loaders

timing look similar to this week runs
I'll give it one more run, to see how it goes

roydahan:
> > Do we have a place where we can easily see which version of c-s we used and the driver?
> 
> 1. no we don't, c-s doesn't print out it's version nor the driver version.
> 2. once docker is used, we check which docker image was used, checking sct.log for `docker.io` would tell you which image was pulled (the confusing part is, that that image would be pulled even when using the VM c-s)

1. I think there is or there was a way to take the version of c-s.
2. The driver we probably can take from clients table in scylla.

Let's have a task for that, we will need it soon.

fruch:
> > > Do we have a place where we can easily see which version of c-s we used and the driver?
> > 
> > 
> > 
> > 1. no we don't, c-s doesn't print out it's version nor the driver version.
> > 2. once docker is used, we check which docker image was used, checking sct.log for `docker.io` would tell you which image was pulled (the confusing part is, that that image would be pulled even when using the VM c-s)
> 
> 1. I think there is or there was a way to take the version of c-s.
> 2. The driver we probably can take from clients table in scylla.
> 
> Let's have a task for that, we will need it soon.


here's the task for it https://github.com/scylladb/qa-tasks/issues/1543.

@roydahan regardless, I still think we should move forward with this change, for the two sample runs I've didn't
looks like it's doesn't drastically change the rolling upgrades cases. 

next we'll need to check again the perf tests, and see the affect. last time we tried i'm not sure what were the results.

roydahan:
For rolling upgrades, if you prove it stable, I'm fine with it.
For performance, I don't want it ever.

fruch:
> For rolling upgrades, if you prove it stable, I'm fine with it.
> For performance, I don't want it ever.

But I do, I want to get rid of those images
I don't mind hardcoded any version of c-s you'll want.

fruch:
> Why do you care about these images? 

cause it's complicated how setup our loader, and for GCP/Azure it's waste us a lot of time on each setup.
I want to drop all the code handling installation of loaders (i.e. installation java and scylla-tools, and such)
we are not using it anymore, it's there only so we can recreate images if needed.

example from this week:
https://github.com/scylladb/scylla-cluster-tests/issues/6738

seems like a simple things to change, and when we did, it broke GCP/Azure
cause we had to keep all the information about where we have images, where we don't have images
in our head, and assumed leave that option empty should be enough, and it wasn't.



> In order to move to the docker based, we need to prove it doesn't change the results in all performance tests. and AFAIR, the last time we tried it, the results were bad.

maybe it was bad back then, be didn't had any information about why it was bad, and since that trial, we fixes a lot of issue with the docker based loader, I think they are in much better shape to try it again.

roydahan:
I see it's merged and part of 5.4 already.
Possibly the cause of the gemini failure with exit code 137.
Let's try my PR to fix it and if not, you'll find the way to fix it....

https://github.com/scylladb/scylla-cluster-tests/pull/6851

fruch:
> I see it's merged and part of 5.4 already.
> Possibly the cause of the gemini failure with exit code 137.
> Let's try my PR to fix it and if not, you'll find the way to fix it....
> 
> https://github.com/scylladb/scylla-cluster-tests/pull/6851

Gemini is always running with docker,

And we don't run multi-dc case regularly, so cloud be related

roydahan:
The failure wasn't in the multi-dc. The Gemini was with docker but all the rest weren't, now probably the other dockers consume more memory.

fruch:
> The failure wasn't in the multi-dc. The Gemini was with docker but all the rest weren't, now probably the other dockers consume more memory.

It was passing o.k. on master and in a few other jobs I've run.

I'm guessing it might be related to the different db instance type which are used, in each (and maybe number of disks)

Will need a wider picture of where exactly is reproduced, to understand the actual cause

PR #6720, opened 2023-10-25
Revert "improvement(gce): change `instance_type_db_oracle:` for gemini tests"

k0machi:
This reverts commit 6e3d34cd1b9fc6533a4e5a3874fd36ad401b9a81.

The reverted commit in question breaks gemini jobs, as the currently used AMI 
for oracle nodes does not support the instance types in question, failing at 
node_setup stage with

```
2023-10-23T11:47:59.434+00:00 gemini-with-nemesis-3h-normal-argus-oracle-db-node-4709e913-1      !ERR | scylla[12015]:  [shard 0] init - Startup failed: std::runtime_error (Bad I/O Scheduler configuration)
```

### DIFF
--- a/test-cases/cdc/cdc-15m-replication-gemini.yaml
+++ b/test-cases/cdc/cdc-15m-replication-gemini.yaml
@@ -7,7 +7,7 @@ n_db_nodes: 3
 instance_type_db: 'i4i.large'
 
 n_test_oracle_db_nodes: 1
-instance_type_db_oracle: 'i4i.large'
+instance_type_db_oracle: 'i3.large'
 
 n_loaders: 1
 instance_type_loader: 'c5.large'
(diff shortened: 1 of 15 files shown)


fruch (approved):
LGTM

fruch:
@Mark-Gurevich FYI, it was part of a change you did, but the gemini cases wasn't test
and the oracle version wasn't supporting i4i instances

soyacz:
hmm, I don't know where my comment go, but I was proposing to upgrade machines from i3.large - as they are less powerful than test cluster instances and we may get to some troubles on more demanding load.
I propose to use i3en.xlarge at least.

@fruch WDYT?

fruch:
> hmm, I don't know where my comment go, but I was proposing to upgrade machines from i3.large - as they are less powerful than test cluster instances and we may get to some troubles on more demanding load. I propose to use i3en.xlarge at least.
> 
> @fruch WDYT?

it's out of the scope of this PR, this is just fixing it to be able to run those tests
it's not about stabilizing the gemini tests, which we are not running anymore on a regular cadence on master  (i.e. they are not in tier1)

and yes in some cases (one that was running regularly), the oracle is using a bigger instance (i3.16xlarge)

PR #6733, opened 2023-10-26
fix(nemesis.py): restore only OSS snapshots on OSS

fgelcer:
since enterprise has an in-memory feature that
is not avaiable on OSS, we got a restore failing
on `malformed_sstable_exception`:
```
Column in_memory missing in current schema
```

Fixes: scylladb/scylladb#15846

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2818,7 +2818,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.log.info("Restoring the schema of the keyspace '%s'", chosen_snapshot_info["keyspace_name"])
             restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=location_list,
                                                            snapshot_tag=chosen_snapshot_tag)
-            restore_task.wait_and_get_final_status(step=10, timeout=120)
+            step = 10
+            extra_time = 300
+            restore_task.wait_and_get_final_status(step=step,
+                                                   timeout=chosen_snapshot_info['expected_timeout'] / step + extra_time)
             assert restore_task.status == TaskStatus.DONE, f'Schema restoration of {chosen_snapshot_tag} has failed!'
             self.cluster.restart_scylla()  # After schema restoration, you should restart the nodes
 
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2774,7 +2774,14 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
             self.use_nemesis_seed()
             chosen_snapshot_size = random.choice(fitting_snapshot_sizes)
-            snapshot_tag = random.choice(list(snapshot_groups_by_size[chosen_snapshot_size]["snapshots"].keys()))
+            if self.cluster.nodes[0].is_enterprise:
+                snapshot_tag = random.choice(list(snapshot_groups_by_size[chosen_snapshot_size]["snapshots"].keys()))
+            else:
+                all_snapshots = snapshot_groups_by_size[chosen_snapshot_size]["snapshots"]
+                oss_snapshots = [snapshot_key for snapshot_key, snapshot_value in all_snapshots.items() if
+                                 snapshot_value['scylla_product'] == "oss"]
+
+                snapshot_tag = random.choice(oss_snapshots)
             snapshot_info = snapshot_groups_by_size[chosen_snapshot_size]["snapshots"][snapshot_tag]
             snapshot_info.update({"expected_timeout": snapshot_groups_by_size[chosen_snapshot_size]["expected_timeout"],
                                   "number_of_rows": snapshot_groups_by_size[chosen_snapshot_size]["number_of_rows"]})
--- a/defaults/manager_persistent_snapshots.yaml
+++ b/defaults/manager_persistent_snapshots.yaml
@@ -9,14 +9,17 @@ aws:
         'sm_20230702185929UTC':
           keyspace_name: "5gb_sizetiered_2022_2"
           scylla_version: "2022.2.9"
+          scylla_product: "enterprise"
           number_of_nodes: 5
         'sm_20230702201949UTC':
           keyspace_name: "5gb_sizetiered_2022_1"
           scylla_version: "2022.1.7"
+          scylla_product: "enterprise"
           number_of_nodes: 5
         'sm_20230702190638UTC':
           keyspace_name: "5gb_sizetiered_5_2"
           scylla_version: "5.2.3"
+          scylla_product: "oss"
           number_of_nodes: 5
     10:
       number_of_rows: 10485760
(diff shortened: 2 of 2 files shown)


fgelcer:
[testing here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/manager/job/restore-nemesis/2/)

sdcm/nemesis.py
```diff
@@ -2774,7 +2774,15 @@ def choose_snapshot(snapshots_dict):
 
             self.use_nemesis_seed()
             chosen_snapshot_size = random.choice(fitting_snapshot_sizes)
-            snapshot_tag = random.choice(list(snapshot_groups_by_size[chosen_snapshot_size]["snapshots"].keys()))
+            if self.cluster.nodes[0].is_enterprise:
+                snapshot_tag = random.choice(list(snapshot_groups_by_size[chosen_snapshot_size]["snapshots"].keys()))
+            else:
+                oss_regex = r'\d\.\d\.\d'
```
bhalevy, line 2780: This seems fragile as a way to detect OSS versions

In any case, can't we have a double digit patch version?

bhalevy: Also the version in the master branch would have `.dev` in it.
I'm not sure if this regex will match it.

fgelcer: first of all, this is related to the already backed up snapshot versions, it has nothing to do with system under tests..
what does test if this is enterprise or OSS is `self.cluster.nodes[0].is_enterprise`

anyway, i thought of a better way to represent it, and will do the changes, and push it soon

fruch: We have better functions to handle this kind of logic, and also we have our own implementation of versions comparing exact for matching those

Anyhow in memory is remove, and 2023.1 doesn't have it's so it should work if we have snapshot from 2023.1

fgelcer: > We have better functions to handle this kind of logic, and also we have our own implementation of versions comparing exact for matching those
> 
> Anyhow in memory is remove, and 2023.1 doesn't have it's so it should work if we have snapshot from 2023.1

[these are the snapshots](https://github.com/scylladb/scylla-cluster-tests/blob/master/defaults/manager_persistent_snapshots.yaml) we have... it means that, snapshots from `2022.1.X` and `2022.2.Y` are not usable, not even on enterprise master branch?
if that is the case, a much bigger refactor must be done in here

fgelcer: @bhalevy ^^

bhalevy: > Anyhow in memory is remove, and 2023.1 doesn't have it's so it should work if we have snapshot from 2023.1

@fruch it wasn't be removed from the schema tables, otherwise upgrade from 2022.X to 2023.1 would not work.

fruch: > > Anyhow in memory is remove, and 2023.1 doesn't have it's so it should work if we have snapshot from 2023.1
> 
> @fruch it wasn't be removed from the schema tables, otherwise upgrade from 2022.X to 2023.1 would not work.

So it's gonna stay there for good ?

bhalevy: We are not actively working on removing it.
I think the best way would be to have the manager restore the schema using some high-level logic.
There are other schema-related incompatibilities between enterprise and open source like IncrementalCompactionStrategy, encryption at rest, and more, that need to be dealt with.

fgelcer: changed the logic, running it with latest code [in here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/manager/job/restore-nemesis/4/)

sdcm/nemesis.py
```diff
@@ -2811,7 +2818,8 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
             self.log.info("Restoring the schema of the keyspace '%s'", chosen_snapshot_info["keyspace_name"])
             restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=location_list,
                                                            snapshot_tag=chosen_snapshot_tag)
-            restore_task.wait_and_get_final_status(step=10, timeout=120)
+            step = 10
+            restore_task.wait_and_get_final_status(step=step, timeout=chosen_snapshot_info['expected_timeout'] / step)
```
bhalevy, line 2822: This hunk looks unrelated

fgelcer: well, it is related to the fact that the test is still not passing, and now it is failing on this call, timing out while waiting for this task to end... but now i have moved it to its own commit

bhalevy (approved):
lgtm

fruch (approved):
LGTM

beside the fact this one closes:
https://github.com/scylladb/scylladb/issues/15846

it's a workaround, and this is still an issue between scylla-core and scylla-manager

fgelcer:
@fruch , the part of the choosing the right snapshot to restore seems now fine, but it is failing with timeout.

maybe i don't fully understand how the implementation works, but `step` is the amount of cycles we will wait for `timeout`, right? if that is the case, there is something wrong with the call, since it fails with timing out after 90 seconds (while the timeout is 900)

i'm still trying to debug it, but i would be happy to get some directions, if you know

fgelcer:
> LGTM
> 
> beside the fact this one closes: [scylladb/scylladb#15846](https://github.com/scylladb/scylladb/issues/15846)
> 
> it's a workaround, and this is still an issue between scylla-core and scylla-manager

if you think that htis is still an issue, then i can remove the `Fixes` and use `Refs` instead

fruch:
> > LGTM
> > beside the fact this one closes: [scylladb/scylladb#15846](https://github.com/scylladb/scylladb/issues/15846)
> > it's a workaround, and this is still an issue between scylla-core and scylla-manager
> 
> if you think that htis is still an issue, then i can remove the `Fixes` and use `Refs` instead

nevermind, it's merged, and I've reopen the issue

fruch:
> @fruch , the part of the choosing the right snapshot to restore seems now fine, but it is failing with timeout.
> 
> maybe i don't fully understand how the implementation works, but `step` is the amount of cycles we will wait for `timeout`, right? if that is the case, there is something wrong with the call, since it fails with timing out after 90 seconds (while the timeout is 900)
> 
> i'm still trying to debug it, but i would be happy to get some directions, if you know

docs say this:

```
def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, stop_event=None, **kwargs):  # pylint: disable=too-many-arguments
    """
    Wrapper function to wait with timeout option.

    :param func: Function to evaluate.
    :param step: Time to sleep between attempts in seconds
    :param text: Text to print while waiting, for debug purposes
    :param timeout: Timeout in seconds
    :param throw_exc: Raise exception if timeout expired, but disrupt_func result is not True
    :param kwargs: Keyword arguments to disrupt_func
    :param stop_event: instance of threading.Event class to stop retrying
    :return: Return value of disrupt_func.
    """
```

so step it time we sleep between attempts, and timeout is the whole timeout.
regardless you are changing the time it waits for schema restore, it shouldn't be extended
if it's timing out, we have a bug needs to be handled

are you testing with master or with 5.4 ?

fgelcer:
> > @fruch , the part of the choosing the right snapshot to restore seems now fine, but it is failing with timeout.
> > maybe i don't fully understand how the implementation works, but `step` is the amount of cycles we will wait for `timeout`, right? if that is the case, there is something wrong with the call, since it fails with timing out after 90 seconds (while the timeout is 900)
> > i'm still trying to debug it, but i would be happy to get some directions, if you know
> 
> docs say this:
> 
> ```python
> def wait_for(func, step=1, text=None, timeout=None, throw_exc=True, stop_event=None, **kwargs):  # pylint: disable=too-many-arguments
>     """
>     Wrapper function to wait with timeout option.
> 
>     :param func: Function to evaluate.
>     :param step: Time to sleep between attempts in seconds
>     :param text: Text to print while waiting, for debug purposes
>     :param timeout: Timeout in seconds
>     :param throw_exc: Raise exception if timeout expired, but disrupt_func result is not True
>     :param kwargs: Keyword arguments to disrupt_func
>     :param stop_event: instance of threading.Event class to stop retrying
>     :return: Return value of disrupt_func.
>     """
> ```
> 
> so step it time we sleep between attempts, and timeout is the whole timeout. regardless you are changing the time it waits for schema restore, it shouldn't be extended if it's timing out, we have a bug needs to be handled
> 
> are you testing with master or with 5.4 ?

yes, aligning it in #6737

tests failing (that led me to this fix) was in master, but same applies to 5.4

PR #6737, opened 2023-10-29
fix(nemesis): mgmt restore timeout

fgelcer:
fix(nemesis): mgmt restore timeout

since #6733 we made changes to this nemesis
(because of OSS vs enterprise system under test
and snapshot versions), but the timeout is still
broken, and from the staging jobs, they did
not fully pass.
this last fix is to align the use of the
`wait_for` calls, where step and timeout
should work now.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2815,10 +2815,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.log.info("Restoring the schema of the keyspace '%s'", chosen_snapshot_info["keyspace_name"])
             restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=location_list,
                                                            snapshot_tag=chosen_snapshot_tag)
-            step = 10
-            extra_time = 300
-            restore_task.wait_and_get_final_status(step=step,
-                                                   timeout=chosen_snapshot_info['expected_timeout'] / step + extra_time)
+
+            restore_task.wait_and_get_final_status(step=10,
+                                                   timeout=6*60)  # giving 6 minutes to restore the schema
             assert restore_task.status == TaskStatus.DONE, f'Schema restoration of {chosen_snapshot_tag} has failed!'
             self.cluster.restart_scylla()  # After schema restoration, you should restart the nodes
 
--- a/defaults/manager_persistent_snapshots.yaml
+++ b/defaults/manager_persistent_snapshots.yaml
@@ -4,7 +4,7 @@ aws:
   snapshots_sizes:
     5:
       number_of_rows: 5242880
-      expected_timeout: 900  # 15 minutes
+      expected_timeout: 1800  # 30 minutes
       snapshots:
         'sm_20230702185929UTC':
           keyspace_name: "5gb_sizetiered_2022_2"
@@ -23,7 +23,7 @@ aws:
           number_of_nodes: 5
     10:
       number_of_rows: 10485760
-      expected_timeout: 1800  # 30 minutes
+      expected_timeout: 3600  # 60 minutes
       snapshots:
         'sm_20230223105105UTC':
           keyspace_name: "10gb_sizetiered"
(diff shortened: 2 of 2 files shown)


sdcm/nemesis.py
```diff
@@ -2818,16 +2818,15 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
             self.log.info("Restoring the schema of the keyspace '%s'", chosen_snapshot_info["keyspace_name"])
             restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=location_list,
                                                            snapshot_tag=chosen_snapshot_tag)
-            step = 10
-            extra_time = 300
-            restore_task.wait_and_get_final_status(step=step,
-                                                   timeout=chosen_snapshot_info['expected_timeout'] / step + extra_time)
+
+            restore_task.wait_and_get_final_status(step=10, timeout=chosen_snapshot_info['expected_timeout'])
```
fruch, line 2822: why can't we just revert this change ? to the static timeout it had ? 

this shouldn't take 30min and even not 15min, it's just the system tables 
if it's stuck we have a bug somewhere, and should be handled, not waiting more.

sdcm/nemesis.py
```diff
@@ -2818,16 +2818,15 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
             assert restore_task.status == TaskStatus.DONE, f'Schema restoration of {chosen_snapshot_tag} has failed!'
             self.cluster.restart_scylla()  # After schema restoration, you should restart the nodes
 
         restore_task = mgr_cluster.create_restore_task(restore_data=True,
                                                        location_list=location_list, snapshot_tag=chosen_snapshot_tag)
-        restore_task.wait_and_get_final_status(step=30, timeout=chosen_snapshot_info["expected_timeout"])
+        timeout_plus_extra_time = 1.3 * chosen_snapshot_info['expected_timeout']
```
fruch, line 2828: where is 1.3 coming from ?

sdcm/nemesis.py
fgelcer: because it is not being enough... and the case i'm currently testing (the most challenging one) is this nemesis during prepare (in the individual nemesis), so it won't be enough 120 seconds

sdcm/nemesis.py
fgelcer: i'm giving it a tolerance to timeout to another 30%

fgelcer: 1 out of 4 nemesis runs passed [here](https://argus.scylladb.com/test/37bfd989-07ba-4778-8899-36d02a013aa9/runs?additionalRuns[]=fb4c19af-675c-4717-b0a2-346fac44c5b6)

still need some work, or to cancel in this case nemesis during prepare (will test that too)

fruch: > i'm giving it a tolerance to timeout to another 30%

If we need extra time, why not update the source yaml with this extra time ?

sdcm/nemesis.py
fruch: The expected timeouts are those measured for the actual restore aren't they ?

It's a bit weird to expect the restore of the scheme to take as long the actual data restore, maybe 2min isn't enough and we need to enlarge it by to 10min, but it should be based on the number that were defined for the data

fgelcer: again, my suspicious is about these nemesis runs during prepare taking much much longer than initially planned

fgelcer:
[here it worked](https://argus.scylladb.com/test/37bfd989-07ba-4778-8899-36d02a013aa9/runs?additionalRuns[]=4e2d690c-a765-4697-b040-703cc66207cd) (and adjusted the timeouts to be closer to the reality of these results), although it had a failure, that i will investigate

fgelcer:
@rayakurl @karol-kokoszka @fruch , please review so we can merge this one, that is affecting all restore nemesis scenarios...
i discussed that with Shlomo, and the problem seems to be that in the manager 3.2, we run a repair operation as part of the restore, making the "restore" operation take much longer

fruch (approved):
Lgtm

Issue #6743, opened 2023-10-31
Failure to run monitor container - can't find image 'mikefarah/yq:3' locally


mykaul:
From https://jenkins.scylladb.com/job/scylla-5.4/job/longevity/job/longevity-twcs-3h-test/3/ : 
```
Command: "sudo docker run --rm -v /home/ubuntu/sct-monitoring/scylla-monitoring-src/config:/workdir mikefarah/yq:3 yq w -i node_exporter_servers.yml '[0].targets[+]' ''[10.12.1.132]:9100''"
Exit code: 125
Stdout:
Stderr:
Unable to find image 'mikefarah/yq:3' locally
```

Additionally, we use v3 - but apparently v4 was released quite some time ago. Different than v3, so might require some changes. Upgrade guide - https://mikefarah.gitbook.io/yq/v/v4.x/upgrading-from-v3

mykaul:
I could not find the use of 'yq' in this repo, only the change to explicitly use v3 @ https://github.com/scylladb/scylla-cluster-tests/commit/a7b6a7b01fbedd7271b4362218ea9f560fae698e
@fruch  ?

mykaul:
BTW, the issue may be a temporary hiccup in Dockerhub:
```
docker: Error response from daemon: received unexpected HTTP status: 503 Service Unavailable.
```

(which is again why I suggest moving to Quay.IO!)

mykaul:
https://www.dockerstatus.com/pages/incident/533c6539221ae15e3f000031/65400e537b554505b4c8c0df

mykaul:
To complete the picture, this is how siren deploys yq - still unsure for which purpose: https://github.com/scylladb/siren/blob/8f648e513969f82648d338456dea054afb166c8a/cluster/setup/scripts/yq.sh

fruch:
> w -i node_exporter_servers.yml '[0].targets[+]' ''[10.12.1.132]:9100''"

the usage is right here on the command:
```
yq w -i node_exporter_servers.yml '[0].targets[+]' ''[10.12.1.132]:9100''"
```

it's used to update the targets in the monitor yaml, since we didn't had yq available, and in some cases we just get monitor image as is, we used this, now we probably have better means to edit that file. (for example using the `remote_file` context manager)

fruch:
I'll try to get rid of it in:
https://github.com/scylladb/scylla-cluster-tests/pull/6746

PR #6746, opened 2023-10-31
fix(monitor-setup): stop using `mikefarah/yq` docker

fruch:
we need to edit a remote file, an we used a docker image of yq for that.

changing it to use SCT `remote_file` context manager

Fixes: #6743

### Testing
- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-5gb-1h-GrowShrinkClusterNemesis-aws-test/2/

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5559,7 +5559,6 @@ class BaseMonitorSet:  # pylint: disable=too-many-public-methods,too-many-instan
               message="Waiting for reconfiguring scylla monitoring")
     def reconfigure_scylla_monitoring(self):
         for node in self.nodes:
-            cluster_backend = self.params.get("cluster_backend")
             monitoring_targets = []
             for db_node in self.targets["db_cluster"].nodes:
                 monitoring_targets.append(f"[{getattr(db_node, self.DB_NODES_IP_ADDRESS)}]:9180")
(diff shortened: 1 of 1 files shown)


sdcm/cluster.py
```diff
@@ -5573,10 +5572,9 @@ def reconfigure_scylla_monitoring(self):
                 python3 genconfig.py -s -n -d {self.monitoring_conf_dir} {monitoring_targets}
             """), verbose=True)
 
-            if cluster_backend != "docker":
-                node.remoter.sudo(f"docker run --rm -v {self.monitoring_conf_dir}:/workdir"
-                                  f" mikefarah/yq:3 yq w -i node_exporter_servers.yml"
-                                  f" '[0].targets[+]' ''[{node.private_ip_address}]:9100''", verbose=True)
+            with node._remote_yaml(f'{self.monitoring_conf_dir}/node_exporter_servers.yml') as exporter_yaml:  # pylint: disable=protected-access
```
soyacz, line 5575: did you test it on docker backend?

soyacz (commented):
now, when we don't need to make exception for docker (or any other backend), why not just to updtate `monitoring_targets` and remove whole idea of updating `node_exproter_servers.yml` file?

sdcm/cluster.py
fruch: doing the docker provision test now

fruch: but it's only for case of adding new node, and so it should be tested directly

soyacz: `reconfigure_scylla_monitoring` doesn't take node param, it reconfigures everything.

fruch: There is a loop on the nodes

soyacz: monitoring nodes

soyacz: monitoring nodes

fruch: It's adding the node exporter of the monitor

fruch:
> now, when we don't need to make exception for docker (or any other backend), why not just to updtate `monitoring_targets` and remove whole idea of updating `node_exproter_servers.yml` file?

`monitoring_targets` is a list we pass on for the db nodes.

adding to that might not exactly get us the result we expect

sdcm/cluster.py
fruch: no, but we are using that code multiple times on scylla yaml, should work as well.

vponomaryov (approved):
LGTM

soyacz (approved):

PR #6749, opened 2023-11-01
fix(k8s): use `preferred podAntiAffinity rules` in `ScyllaCluster` K8S objects

vponomaryov:
For the moment we use 'required' podAntiAffinity rules in the ScyllaCluster objects to make sure Scylla pods of a same DB cluster get scheduled on separate nodes.
The problem is that `Scylla cleanup` pods get scheduled using the same affinity rules making it require at least one K8S Scylla node be without Scylla pods.

Also, our functional test for checking that `cleanup` functionality is not stable due to it, because it utilizes all 4 K8S Scylla nodes.

So, allow `Scylla cleanup` pods be scheduled anytime by using the `preferred podAntiAffinity rules` in single-tenant test runs. Keep old `required` way only for the multi-tenant test runs because, in this case, Scylla pods use less than 50% of K8S Scylla nodes resources.


soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #6759, opened 2023-11-05
fix(monitor): call `disable_daily_triggered_services` on all node types

fruch:
Since we can have those unattanded services doing things in the background while trying to use apt-get, we also want those disabled on monitor and loader nodes.

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4944,6 +4944,7 @@ class BaseLoaderSet():
 
         # update repo cache and system after system is up
         node.update_repo_cache()
+        node.disable_daily_triggered_services()
 
         if TestConfig().REUSE_CLUSTER:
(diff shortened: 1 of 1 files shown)


soyacz (approved):
LGTM
Btw. Shouldn't we update from time to time Monitor/loader used AMI's, so we work with more up-to-date packages?

fruch:
> LGTM
> Btw. Shouldn't we update from time to time Monitor/loader used AMI's, so we work with more up-to-date packages?

We should,

There is a task to move the monitor to official images.

The loaders are a bit problematic, cause we need them static more or less for performance tests, but we can update them for the rest of the tests, it's harder to build them automatically (it's currently mostly manual process)

juliayakovlev (approved):

PR #6761, opened 2023-11-06
feature(commitlog): Checking commitlog in sct

temichus:
Checking "free segments" and "commitlog directory" metrics in SCT task: https://github.com/scylladb/qa-tasks/issues/1523

introdus new SCT background thread CommitLogCheckThread

if commitlog-use-hard-size-limit is enabled,
this thread will check the following metrics during the test:

1) commit log directory does not exceed the limit at any stage of the test. if it exceeded - send an error event that fails the test.

2) free segments don't drop to zero at any stage of the test, otherwise, send an error event that fails the test.

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -238,3 +238,5 @@ use_placement_group: false
 validate_large_collections: false
 
 scylla_d_overrides_files: []
+
+run_commit_log_check_thread: true
--- a/sdcm/commit_log_check_thread.py
+++ b/sdcm/commit_log_check_thread.py
@@ -1,0 +1,187 @@
+import traceback
+import logging
+import time
+import re
+import threading
+from dataclasses import dataclass
+from distutils.util import strtobool
+from sdcm.sct_events.database import CommitLogCheckErrorEvent
+from sdcm.rest.remote_curl_client import RemoteCurlClient
+
+
+@dataclass
+class CommitlogConfigParams:
+    def __init__(self, db_cluster, ):
+        logger = logging.getLogger(self.__class__.__name__)
+        with db_cluster.cql_connection_patient(
+                node=db_cluster.nodes[0],
+                connect_timeout=300,) as session:
+            self.use_hard_size_limit = bool(strtobool(session.execute(
+                "SELECT value FROM system.config WHERE name='commitlog_use_hard_size_limit'").one().value))
+            self.segment_size_in_mb = int(session.execute(
+                "SELECT value FROM system.config WHERE name='commitlog_segment_size_in_mb'").one().value)
+            self.max_disk_size = int(RemoteCurlClient(
+                host="localhost:10000", endpoint="commitlog", node=db_cluster.nodes[0]).run_remoter_curl(
+                method="GET", path="metrics/max_disk_size", params=None).stdout)
+            self.smp = len(re.findall(
+                "shard",
+                db_cluster.nodes[0].remoter.run('seastar-cpu-map.sh -n scylla').stdout))
+            self.total_space = int(self.max_disk_size / self.smp)
+
+            logger.debug("CommitlogConfigParams")
+            logger.debug("smp: %s", self.smp)
+            logger.debug("max_disk_size: %s", self.max_disk_size)
+            logger.debug("total_space: %s", self.total_space)
+            logger.debug("use_hard_size_limit: %s", self.use_hard_size_limit)
+            logger.debug("segment_size_in_mb: %s", self.segment_size_in_mb)
+
+    smp: int
+    total_space: int
+    max_disk_size: int
+    use_hard_size_limit: bool
+    segment_size_in_mb: int
+
+
+@dataclass
+class PrometheusQueries:
+    def __init__(self, commitlog_params: CommitlogConfigParams):
+        self.overflow_commit_log_directory = self.overflow_commit_log_directory.format(
+            commitlog_total_space=commitlog_params.total_space,
+            commitlog_segment_size_in_mb=commitlog_params.segment_size_in_mb
+        )
+
+        self.zero_free_segments = self.zero_free_segments.format(
+            commitlog_total_space=commitlog_params.total_space,
+            commitlog_segment_size_in_mb=commitlog_params.segment_size_in_mb
+        )
+    # Prometheus queries with detailed description below
+    overflow_commit_log_directory = ("scylla_commitlog_disk_total_bytes>=({commitlog_total_space}"
+                                     "%2B({commitlog_segment_size_in_mb}%2A1048576))")
+    """
+        returns commitlog total size on disk if commit log directory exceed the limit
+
+        commit_log_directory_query description:
+        %2A is *
+        %2B is +
+
+        {commitlog_total_space} :
+        set to max_disk_size(value from API) divided  to number for shards
+        see CommitLogCheckThread.init_commitlog_params_from_db()
+
+        + {commitlog_segment_size_in_mb}*1048576) :
+        add size of 1 segment,  Scylla actually stops creating new segments not when the next segment
+        would cause us to exceed the value,
+        but instead it checks if we've already reached or exceeded the limit.
+
+        scylla_commitlog_disk_total_bytes > :
+        filter all good values when scylla_commitlog_disk_total_bytes<= commitlog_total_space + 1 segment
+        and return from Prometheus only bad
+    """
+
+    zero_free_segments = \
+        ("scylla_commitlog_disk_active_bytes>((scylla_commitlog_disk_total_bytes>={commitlog_total_space})"
+         "%2D({commitlog_segment_size_in_mb}%2A1048576))")
+    """
+        returns commitlog active size on disk if number of free segments drop to zero
+
+        free_segments_query description:
+        %2A is *
+        %2D is -
+
+        {commitlog_total_space} :
+        set to max_disk_size(value from API) divided  to number for shards
+        see CommitLogCheckThread.init_commitlog_params_from_db()
+
+        scylla_commitlog_disk_total_bytes >= {commitlog_total_space}:
+        Scylla has 0 free segments and allocates new segments until exceed limit {commitlog_total_space}
+        this condition will ignore result when Scylla does not exceed  segments limit
+
+        - {commitlog_segment_size_in_mb}*1048576) :
+        minus size of 1 segment,  to ensure that Scylla has at least 1 free segment number of active_bytes should be
+        less then "scylla_commitlog_disk_total_bytes"-"1 segment"
+
+        scylla_commitlog_disk_active_bytes > :
+        filter all good values when scylla_commitlog_disk_active_bytes<= scylla_commitlog_disk_total_bytes - 1 segment
+        and return from Prometheus only bad
+    """
+
+
+# pylint: disable=too-many-instance-attributes
+class CommitLogCheckThread:
+    """
+        if commitlog-use-hard-size-limit is enabled,
+        this thread will check the following metrics during the test:
+
+        1) commit log directory does not exceed the limit at any stage of the test.
+           if it exceeded - send an error event that fails the test.
+
+        2) free segments don't drop to zero at any stage of the test,
+        otherwise, send an error event that fails the test.
+    """
+
+    # Prometheus API limit: check_interval/discreteness must be less then 11000
+    check_interval = 10 * 60
+    discreteness = 1
+
+    def __init__(self, custer_tester, test_duration, termination_event=None, thread_name: str = ""):
+        self.log = logging.getLogger(self.__class__.__name__)
+        self.prometheus = custer_tester.prometheus_db
+        self.test_duration = test_duration
+
+        self.commitlog_params = CommitlogConfigParams(custer_tester.db_cluster)
+        self.prometheus_queries = PrometheusQueries(self.commitlog_params)
+        self.start_time = None
+
+        self._thread = threading.Thread(daemon=True, name=f"{self.__class__.__name__}_{thread_name}", target=self.run)
+        self.termination_event = termination_event
+        if not self.termination_event:
+            self.termination_event = custer_tester.db_cluster.nemesis_termination_event
+
+    def start(self):
+        if self.commitlog_params.use_hard_size_limit:
+            self.log.debug("starting CommitLogCheckThread")
+            self._thread.start()
+        else:
+            self.log.debug(
+                "CommitLogCheckThread was not started due to commitlog_use_hard_size_limit is %s",
+                self.commitlog_params.use_hard_size_limit)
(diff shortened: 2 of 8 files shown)


sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,192 @@
...
+            trace = traceback.format_exc()
+            CommitLogCheckErrorEvent(
+                message=f"CommitLogCheckThread failed: {exc.__repr__()} with traceback {trace}").publish()
+            raise
+        self.log.debug("CommitLogCheckThread finished")
+
+    def commit_log_directory_checker(self, start_time, end_time):
+        response = self.prometheus.query(
+            self.prometheus_queries.commit_log_directory_query, start_time, end_time, self.discreteness)
+        self.log.warning("scylla_commitlog_directory_diff_res")
```
fruch, line 178: please remove the warnings if not needed

defaults/test_default.yaml
```diff
@@ -233,3 +233,5 @@ use_dns_names: false
 use_placement_group: false
 
 scylla_d_overrides_files: []
+
+run_commit_log_check_thread: true
```
fruch, line 237: by default we enable the thread, but the feature is enabled only on one test, wouldn't it gonna fail on any other test ?

test-cases/longevity/longevity-10gb-3h.yaml
```diff
@@ -21,3 +21,5 @@ space_node_threshold: 64424
 
 gce_n_local_ssd_disk_db: 2
 run_db_node_benchmarks: true
+
+append_scylla_args: '--commitlog-use-hard-size-limit=true'
```
fruch, line 25: is this feature gonna be default in scylla ? do we want it default in SCT ?

temichus: yes, it will be default in the future https://github.com/scylladb/scylladb/pull/10974.

But I changed here just for testing. will remove when PR will be ready

defaults/test_default.yaml
temichus: feature will be enabled by default Scylla https://github.com/scylladb/scylladb/pull/10974

iIdealy it should work in all tests. but this line is made to remove it by default, instead of reverting all code. or disable in some tests if needed 

scilla req: The commitlog didn't go over the size limit. It is a little more involved, we should make sure that the size of the commitlog directory never exceeds the limit, we can't rely on metrics since if there is a bug in the metrics we will never catch an actual bug where the directory is larger than the stats and the size limit was actually breached.

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,192 @@
...
+            commitlog_total_space_in_mb=commitlog_params.total_space_in_mb,
+            commitlog_segment_size_in_mb=commitlog_params.segment_size_in_mb
+        )
+
+        self.free_segments_query = self.free_segments_query.format(
+            commitlog_total_space_in_mb=commitlog_params.total_space_in_mb,
+            commitlog_segment_size_in_mb=commitlog_params.segment_size_in_mb
+        )
+    # Prometheus queries with detailed description below
+    commit_log_directory_query = ("scylla_commitlog_disk_total_bytes>({commitlog_total_space_in_mb}"
```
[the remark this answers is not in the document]
temichus: because returning  any value means an error - data will be returned per shard:

```
{'metric': {'__name__': 'scylla_commitlog_disk_total_bytes', 'instance': '10.4.3.68', 'job': 'manager_agent', 'shard': '0'}, 'values': [[1698765005.833, '8757706752'], [1698765006.833, '8757706752']..............
```

temichus:
testing runs : 
1) special cluster where the thread shows errors   https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/longevity-10gb-3h-test-pr-5797/109/

2) longevity-10gb-3h with commitlog-use-hard-size-limit=true (failed but thread works correctly) https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/longevity-10gb-3h-test-pr-5797/110/

temichus:
@scylladb/qa-maintainers,  PR is tested and ready for review. Can be merged now  because the thread started only if  commitlog-use-hard-size-limit is enabled (now  there are 0 such tests ) or postponed and merged after https://github.com/scylladb/scylladb/pull/10974 

But for sure we should present this functionality to all QAs before merge because it will affect all tests in SCT

soyacz (commented):
quite late comment, but I wonder if we couldn't achieve this with `PrometheusAlertManagerListener` and setup proper alerts in prometheus.
Maybe worth to consider it for future (unless we can't use it).

Besides, small comments to fix and overall LGTM

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,177 @@
...
+        add size of 1 segment,  Scylla actually stops creating new segments not when the next segment
+        would cause us to exceed the value,
+        but instead it checks if we've already reached or exceeded the limit.
+
+        scylla_commitlog_disk_total_bytes > :
+        filter all good values when scylla_commitlog_disk_total_bytes<= commitlog_total_space_in_mb + 1 segment
+        and return from Prometheus only bad
+    """
+
+    free_segments_query = \
```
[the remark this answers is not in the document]
temichus: done

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,177 @@
...
+            raise
+        self.log.debug("CommitLogCheckThread finished")
+
+    def commit_log_directory_checker(self, start_time, end_time):
+        response = self.prometheus.query(
+            self.prometheus_queries.commit_log_directory_query, start_time, end_time, self.discreteness)
+        self.log.debug("scylla_commitlog_directory_diff_res: {}".format(response))
+        if response:
+            CommitLogCheckErrorEvent(
+                message=f"commit log directory exceed the limit. Prometeus responce: {response}").publish()
```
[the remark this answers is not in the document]
temichus: done

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,180 @@
...
+        would cause us to exceed the value,
+        but instead it checks if we've already reached or exceeded the limit.
+
+        scylla_commitlog_disk_total_bytes > :
+        filter all good values when scylla_commitlog_disk_total_bytes<= commitlog_total_space_in_mb + 1 segment
+        and return from Prometheus only bad
+    """
+
+    zero_free_segments = \
+        ("scylla_commitlog_disk_active_bytes>((scylla_commitlog_disk_total_bytes>{commitlog_total_space_in_mb})"
```
wmitros, line 71: I'm not that familiar with prometheus queries format but for the `scylla_commitlog_disk_total_bytes>{commitlog_total_space_in_mb}` part we should have a `>=` (it doesn't matter now because in most default configurations `commitlog_total_space_in_mb` is not actually a multiple of segment size)

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,180 @@
...
+        %2D is -
+
+        {commitlog_total_space_in_mb} :
+        if commitlog_total_space_in_mb is not -1 use  commitlog_total_space_in_mb*1048576
+        else commitlog_total_space_in_mb='scylla_memory_total_memory' metric
+        see CommitLogCheckThread.init_commitlog_params_from_db()
+
+        scylla_commitlog_disk_total_bytes > {commitlog_total_space_in_mb}:
+        Scylla has 0 free segments and allocates new segments until exceed limit {commitlog_total_space_in_mb}
+        this condition will ignore result when Scylla doe not exceed  segments limit
```
wmitros, line 87:
```suggestion
        this condition will ignore result when Scylla does not exceed  segments limit
```

wmitros (commented):
I checked the invariants of the commitlog that we're checking and only found minor issues. However, recently I found that the method for getting the commitlog limit (based on whether the config option is -1) is not actually precise, so instead of hard-coding all the adjustments here I just posted https://github.com/scylladb/scylladb/pull/16020 which should allow getting the exact limit directly from the metrics (new metric will be named `scylla_commitlog_max_disk_size`). We may have to postpone merging this until that one's merged as well, but maybe it's not a problem as we're also waiting for https://github.com/scylladb/scylladb/pull/10974

sdcm/commit_log_check_thread.py
temichus: fixed

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,180 @@
...
+                "SELECT value FROM system.config WHERE name='commitlog_total_space_in_mb'").one().value)
+            self.use_hard_size_limit = bool(strtobool(session.execute(
+                "SELECT value FROM system.config WHERE name='commitlog_use_hard_size_limit'").one().value))
+            self.segment_size_in_mb = int(session.execute(
+                "SELECT value FROM system.config WHERE name='commitlog_segment_size_in_mb'").one().value)
+
+            if self.total_space_in_mb == -1:
+                self.total_space_in_mb = "scylla_memory_total_memory"
+            else:
+                self.total_space_in_mb = self.total_space_in_mb * 1048576
```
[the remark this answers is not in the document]
temichus: fixed

temichus:
Moved to draft 

waiting for https://github.com/scylladb/scylladb/pull/16020
blocked by: https://github.com/scylladb/scylladb/pull/10974

temichus:
@wmitros I've updated PR to use /commitlog/metrics/max_disk_size API
Could you please check that the calculation is right?

test run with new code version https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/longevity-10gb-3h-test-pr-5797/117/

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,192 @@
...
+            commitlog_total_space=commitlog_params.total_space,
+            commitlog_segment_size_in_mb=commitlog_params.segment_size_in_mb
+        )
+
+        self.zero_free_segments = self.zero_free_segments.format(
+            commitlog_total_space=commitlog_params.total_space,
+            commitlog_segment_size_in_mb=commitlog_params.segment_size_in_mb
+        )
+    # Prometheus queries with detailed description below
+    overflow_commit_log_directory = ("scylla_commitlog_disk_total_bytes>({commitlog_total_space}"
```
wmitros, line 60: If we want to be absolutely precise, we should have actually have a `>=` here (I might have missed it in the review of previous version)

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,192 @@
...
+
+@dataclass
+class CommitlogConfigParams:
+    def __init__(self, db_cluster, ):
+        logger = logging.getLogger(self.__class__.__name__)
+        with db_cluster.cql_connection_patient(
+                node=db_cluster.nodes[0],
+                connect_timeout=300,) as session:
+            self.total_space = int(session.execute(
+                "SELECT value FROM system.config WHERE name='commitlog_total_space_in_mb'").one().value)
```
wmitros, line 20: If we continue with the approach with the API, this is probably not necessary anymore

wmitros (approved):
LGTM, I found only 1 place where we weren't exactly precise. As we discussed, it seems like the approach without the API was good too, so if you prefer we can revert to that one. Otherwise we can just go ahead with this one

sdcm/commit_log_check_thread.py
temichus: fixed

sdcm/commit_log_check_thread.py
temichus: fixed

temichus:
@soyacz, @fruch - could you please re review?  this PR ready to be merged.

sdcm/commit_log_check_thread.py
```diff
@@ -0,0 +1,190 @@
...
+        try:
+            thread_end_time = time.time() + self.test_duration
+
+            self.start_time = time.time()
+
+            while time.time() < thread_end_time and not self.termination_event.is_set():
+                interval_end_time = self.start_time + self.check_interval
+
+                while time.time() < interval_end_time and not self.termination_event.is_set():
+                    self.termination_event.wait(60)
```
fruch, line 163: can you explain this, it's not clear what this loop is doing...

fruch: if I read it correctly it works every minute ? isn't that a bit too excessive ?

fruch: the whole play with saving clocktime, and playing with it's clear

why not just wait the interval ?

temichus: this was put to be able exit from thread every 60s(when test finished and termination_event is set) not every 600s 

but according to [doc](https://docs.python.org/3/library/threading.html#event-objects) it should work also as 
 self.termination_event.wait(self.check_interval) will test it

fruch: you can see all the other places we have a temnication_event and a loop, like in Nemesis thread, that we do exactly that

temichus: updated

fruch (approved):
LGTM

soyacz (approved):
LGTM

soyacz:
@temichus please verify for next few weeks if works as expected and ping me for backport

temichus:
@soyacz @fruch FYI
For backport it requires 2 Scylla PRs 
1)https://github.com/scylladb/scylladb/pull/10974
2) https://github.com/scylladb/scylladb/pull/16020

yaronkaikov:
@temichus it seems that this commit is breaking the artifact tests. see here for example https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-ubuntu2204-test/374/consoleFull

fruch:
I'm reverting this until fixed
```
02:02:11  CommitLogCheckThread_:
02:02:11  Traceback (most recent call last):
02:02:11    File "/usr/local/lib/python3.10/threading.py", line 1016, in _bootstrap_inner
02:02:11      self.run()
02:02:11    File "/usr/local/lib/python3.10/threading.py", line 953, in run
02:02:11      self._target(*self._args, **self._kwargs)
02:02:11    File "/tmp/workspace/scylla-master/artifacts/artifacts-ubuntu2204-test/scylla-cluster-tests/sdcm/commit_log_check_thread.py", line 162, in run
02:02:11      self.overflow_commit_log_directory_checker(self.start_time, interval_end_time)
02:02:11    File "/tmp/workspace/scylla-master/artifacts/artifacts-ubuntu2204-test/scylla-cluster-tests/sdcm/commit_log_check_thread.py", line 174, in overflow_commit_log_directory_checker
02:02:11      response = self.prometheus.query(
```

PR #6769, opened 2023-11-07
fix(remote_file): write inplace into the file

fruch:
since in some cases there's a process listing on changes
to this file, doing so with `mv` fails to continue the watch
on those files

### Testing
- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-5gb-1h-GrowShrinkClusterNemesis-aws-test/5/

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5562,7 +5562,6 @@ class BaseMonitorSet:  # pylint: disable=too-many-public-methods,too-many-instan
               message="Waiting for reconfiguring scylla monitoring")
     def reconfigure_scylla_monitoring(self):
         for node in self.nodes:
-            cluster_backend = self.params.get("cluster_backend")
             monitoring_targets = []
             for db_node in self.targets["db_cluster"].nodes:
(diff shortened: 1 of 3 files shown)


fruch:
@soyacz @vponomaryov found the root cause of why there was no updates, the `mv` that `remote_file` was doing
caused prometheus to stop getting update on that file, and changing it in-place solve the issue 

here a dashboard showing the monitor node data is there in all of the test, while before it was gone after the first update
image

soyacz:
> @soyacz @vponomaryov found the root cause of why there was no updates, the `mv` that `remote_file` was doing caused prometheus to stop getting update on that file, and changing it in-place solve the issue

nice catch,
the only left thing is to fix unit tests:
```
FAILED unit_tests/test_remoter.py::TestRemoteFile::test_remote_file - AssertionError: 'bash -cxe "cat \'temporary\' > \'/some/path/some.file\'\nrm \'temporary\'\n"' != "mv 'temporary' '/some/path/some.file'"
```

vponomaryov (changes requested):
Good catch.
Left to fix unit tests.

soyacz (approved):
LGTM

PR #6775, opened 2023-11-08
improvement(kallsyms decoder): add script that decodes kallsyms

fgelcer:
since we collect kallsyms, and they are sometimes
needed to add valuable informatino to existing issues, i have created this tool to decode kernel callstacks from tests. it is not yet ready to be integrated
with some automatic decoded, but the content is ready to be used, as a standalone script for the now.

### DIFF
--- a/utils/decode_kallsyms.py
+++ b/utils/decode_kallsyms.py
@@ -1,0 +1,53 @@
+import bisect
+import re
+
+
+def kallsyms_search(kallsyms_lines: list[str], address: int):
+    # kallsyms lines looks like this
+    #
+    # ffffffffaa88e2f0 T kfree
+    # ffffffffaa88e400 T __ksize
+    # ffffffffaa88e530 T ksize
+    key = lambda line: int(line.split()[0], 16) - 1
+    return bisect.bisect_right(kallsyms_lines, address, key=key) - 1
+
+
+def cut_addresses_from_log_line_kernel_callstack(line):
+    my_regex = re.compile('0x([0-9a-fA-F]{16})')
+    return my_regex.findall(line)
+
+
+def get_lines_from_log(file_path):
+    file_content = {}
+    with open(file_path, 'r') as file:  # pylint: disable=unspecified-encoding
+        for line in file:
+            if ('kernel callstack' in line and '0x' in line) or 'Reactor stalled for' in line:
+                file_content[line] = cut_addresses_from_log_line_kernel_callstack(line)
+    return file_content
+
+
+def get_kallsyms(path_to_kallsyms):
+    with open(path_to_kallsyms, 'r') as file:  # pylint: disable=unspecified-encoding
+        file_content = file.readlines()
+    return file_content
+
+
+def append_content_to_file(file_stream, content):
+    file_stream.write(content + '\n')
+
+
+def decode_kernel_callstacks(results_file='results.log', input_file='system.log', kallsyms_file='kallsyms',
+                             clear_output_file=True):
+    file_content = get_kallsyms(kallsyms_file)
+    if clear_output_file:
+        with open(results_file, 'w'):  # pylint: disable=unspecified-encoding
+            print(f'clearing file {results_file}')
+
+    with open(results_file, 'a') as file:  # pylint: disable=unspecified-encoding
+        for full_line, stall in get_lines_from_log(input_file).items():
+            append_content_to_file(file, '#' * 76)
+            append_content_to_file(file, full_line)
+            if 'Reactor stalled for' not in full_line:
+                for frame in stall:
+                    append_content_to_file(file, file_content[kallsyms_search(file_content, int(frame, 16))])
+            append_content_to_file(file, '#' * 76)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -34,7 +34,7 @@ repos:
     hooks:
     -   id: autopep8
         name: autopep8
-        entry: autopep8 -i -j 2 --max-line-length=120 --ignore=E226,E24,W50,W690,E402
+        entry: autopep8 -i -j 2 --max-line-length=120 --ignore=E226,E24,W50,W690,E402,E731
         language: system
         types: [python]
         exclude: '\.sh$'


vponomaryov (changes requested):
As of now, this file, if imported, will try to process files. 
Which is unexpected for the modules kept under the `sdcm` directory.

I think it must be placed in here: `utils/decode_kallsyms.py`

sdcm/utils/decode_kallsyms.py
```diff
@@ -0,0 +1,48 @@
...
+        file.write(content + '\n')
+
+
+# TODO: dynamically get the 3 files below (from command line arguments for example where results could have a default)
+results_file = 'results.log'
+file_content = get_kallsyms('kallsyms')
+open(results_file, 'w').close()
+
+for full_line, stall in get_lines_from_log('system.log').items():
+    append_content_to_file(results_file, '############################################################################')
```
vponomaryov, line 42: The line separator could look like `'#' * 76` and defined before the loop

sdcm/utils/decode_kallsyms.py
```diff
@@ -0,0 +1,48 @@
+import bisect
+import re
+import subprocess
+
+
+def binary_search_recursive(array, element):
+    return bisect.bisect_right(array, element, key=lambda y: int(y.split()[0], 16)) - 1
+
+
+
```
vponomaryov, line 10: Redundant line here and at the end of the file.

sdcm/utils/decode_kallsyms.py
```diff
@@ -0,0 +1,47 @@
...
+    with open(file_path, 'a') as file:
+        file.write(content + '\n')
+
+
+# TODO: dynamically get the 3 files below (from command line arguments for example where results could have a default)
+results_file = 'results.log'
+file_content = get_kallsyms('kallsyms')
+open(results_file, 'w').close()
+
+for full_line, stall in get_lines_from_log('system.log').items():
```
vponomaryov, line 40: The `system.log` should have been in a var defined before the loop.
It is always desirable to keep logic separate from input data definition.

sdcm/utils/decode_kallsyms.py
fgelcer: done

sdcm/utils/decode_kallsyms.py
```diff
@@ -0,0 +1,48 @@
...
+        file_content = file.readlines()
+    return file_content
+
+
+def append_content_to_file(file_path, content):
+    with open(file_path, 'a') as file:
+        file.write(content + '\n')
+
+
+# TODO: dynamically get the 3 files below (from command line arguments for example where results could have a default)
```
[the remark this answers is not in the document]
fgelcer: done

sdcm/utils/decode_kallsyms.py
fgelcer: removed

sdcm/utils/decode_kallsyms.py
fgelcer: done

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,45 @@
...
+def decode_kernel_callstacks(results_file='results.log', input_file='system.log', kallsyms_file='kallsyms',
+                             clear_output_file=True):
+    file_content = get_kallsyms(kallsyms_file)
+    if clear_output_file:
+        open(results_file, 'w').close()
+
+    for full_line, stall in get_lines_from_log(input_file).items():
+        append_content_to_file(results_file, '#' * 76)
+        append_content_to_file(results_file, full_line)
+        for line in stall:
```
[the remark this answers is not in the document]
fgelcer: changed

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,45 @@
+import bisect
+import re
+
+
+def binary_search_recursive(array, element):
+    return bisect.bisect_right(array, element, key=lambda y: int(y.split()[0], 16)) - 1
+
+
+def cut_addresses_from_log_line_kernel_callstack(line):
+    my_regex = re.compile('0x([0-9a-f]{16})')
```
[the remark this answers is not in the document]
fgelcer: added

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,45 @@
+import bisect
+import re
+
+
+def binary_search_recursive(array, element):
+    return bisect.bisect_right(array, element, key=lambda y: int(y.split()[0], 16)) - 1
```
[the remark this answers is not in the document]
fgelcer: i used your suggestion in another review you did, in another file... using your suggestion here (@vponomaryov , if you could confirm it works the same with the kernel callstacks you have, would be appreciated)

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,45 @@
...
+    return file_content
+
+
+def get_kallsyms(path_to_kallsyms):
+    with open(path_to_kallsyms, 'r') as file:
+        file_content = file.readlines()
+    return file_content
+
+
+def append_content_to_file(file_path, content):
```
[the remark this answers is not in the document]
fgelcer: the open with mode `w` is used only once, to clear the file from previous content, in the beginning of the function run, but we don't have a new to choose the mode like you suggest here, although now, we perhaps can only use open, since we are only passing the file stream to the writer function

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,45 @@
...
+    return bisect.bisect_right(array, element, key=lambda y: int(y.split()[0], 16)) - 1
+
+
+def cut_addresses_from_log_line_kernel_callstack(line):
+    my_regex = re.compile('0x([0-9a-f]{16})')
+    return my_regex.findall(line)
+
+
+def get_lines_from_log(file_path):
+    file_content = dict()
```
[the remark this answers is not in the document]
fgelcer: dict is the best format for Python to use, and in this case, it doesn't really change much the logic on the purpose of this dictionary usage, since we want to print both the line as it is in the log, and the decoded kernel callstack

utils/decode_kallsyms.py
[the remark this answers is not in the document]
michoecho: I know, I remember that issue. But that was a quick suggestion. Now that we are trying to commit it, we should add some meaningful names and comments.

Note that the new suggestion does the same thing as the old one, only the names are different.

utils/decode_kallsyms.py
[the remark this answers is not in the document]
michoecho: This was just a nitpick. Dict is okay, if you like it more.

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,57 @@
+import bisect
+import re
+
+
+def binary_search_recursive(array, element):
+    return bisect.bisect_right(array, element, key=lambda y: int(y.split()[0], 16)) - 1
```
michoecho, line 6: Since you added the `kallsyms_search` variant, you can now remove the `binary_search_recursive` function.

michoecho: Aside from that, LGTM.

vponomaryov: Agree, need to remove the `binary_search_recursive` function as unused.

fgelcer: yes, it was a mistake, i forgot to remove the function.. now it is removed

michoecho (approved):

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,53 @@
...
+import re
+
+
+def kallsyms_search(kallsyms_lines: list[str], address: int):
+    # kallsyms lines looks like this
+    #
+    # ffffffffaa88e2f0 T kfree
+    # ffffffffaa88e400 T __ksize
+    # ffffffffaa88e530 T ksize
+    def key(line): return int(line.split()[0], 16) - 1  # pylint: disable=multiple-statements
```
vponomaryov, line 11: Actually, there is the `lambda` built-in that allows to do one-liner functions:

```
key = lambda line: int(line.split()[0], 16) - 1
```

Also, it won't require suppressing of pylint rules.

vponomaryov (dismissed):
LGTM

utils/decode_kallsyms.py
michoecho: It started as a lambda, then @fgelcer changed it to a function, and now back to a lambda. I wonder what's that about.

vponomaryov (approved):
LGTM

utils/decode_kallsyms.py
fgelcer: yes, from PEP8 PoV you were right that a lambda like it is would be an error, but we decided to ignore it...
funtionality wise, it does what is expected, so if you are not strongly against it, please approve

fgelcer:
@fruch @soyacz @michoecho

michoecho:
LGTM

utils/decode_kallsyms.py
```diff
@@ -0,0 +1,53 @@
...
+    with open(path_to_kallsyms, 'r') as file:  # pylint: disable=unspecified-encoding
+        file_content = file.readlines()
+    return file_content
+
+
+def append_content_to_file(file_stream, content):
+    file_stream.write(content + '\n')
+
+
+def decode_kernel_callstacks(results_file='results.log', input_file='system.log', kallsyms_file='kallsyms',
```
fruch, line 39: I would be nice if there would be an example showing how to use it.

i.e. now you can run this from commandline, so you need to write more code to use it.

fgelcer: it can be called from a simple Python code, and give the relevant files, and collect the output...
the idea of this is to futurely allow us to automatically decode kernel callstacks from within SCT, and not a command line script (that can be easily adapted from this existing code BTW)

fruch: event something like, probably would do the trick :

```python
if __name__ == "__main__":
      decode_kernel_callstacks()
```

fgelcer: i can be added later, because it don't think it should be part of me adding this tool to SCT repo, since it is not yet used anywhere

fruch: I was more concern about someone that needs it manually now, and will start to figure out himself what exactly he need todo.

later automating it, is a different story

fruch (approved):
LGTM

Issue #6689, opened 2023-10-17
increase buffer size for syslog-ng


soyacz:
During network issues or high traffic syslog-ng buffers fill up and drop messages.

Let's try increasing syslogng buffers to tackle this issue.

fruch:
We try to guesstimate how much logs scylla can generate within 30sec, and based on that set the size

question is how we should do, taking one 12h job, and analyze the logs timestamps ? to find the size ?

roydahan:
First, I would check with @kbr-scylla what is the scenario that reproduces for him consistently the drop of syslog-ng.
Then, with that I would start by doubling the buffer and see if it solve the issue.
Then, I would test it on one of the MV or LWT longevities that we have and spam the logs and see if it causes any issue.

kbr-scylla:
> First, I would check with @kbr-scylla what is the scenario that reproduces for him consistently the drop of syslog-ng.

https://jenkins.scylladb.com/job/scylla-staging/job/KamilBraun/job/longevity-schema-changes-3h-test-artsiom_mishuta-clone/

I checked some of the recent runs, and they had the problem.
(I didn't check early runs, but I would guess they also had this problem)


> We try to guesstimate how much logs scylla can generate within 30sec, and based on that set the size

30s is pretty conservative. Why so low? We're already keeping the entire Scylla log in system.log. Having a separate buffer that is just a fraction of that should not be a problem. Even if we picked e.g. 10 minutes it would still be a small fraction of system.log during longevity test.

kbr-scylla:
https://jenkins.scylladb.com/job/scylla-staging/job/KamilBraun/job/longevity-schema-changes-3h-test-artsiom_mishuta-clone/16/
https://cloudius-jenkins-test.s3.amazonaws.com/9f0e827b-8a9a-4767-b4b4-0582d3f9415f/20231007_113814/db-cluster-9f0e827b.tar.gz   

Example of event from longevity-ParallelAddDropColumnMonk-db-node-9f0e827b-1/messages.log:
```
2023-10-07T10:46:10.432+00:00 longevity-ParallelAddDropColumnMonk-db-node-9f0e827b-1      !ERR | syslog-ng[4611]: EOF occurred while idle; fd='39'
[...]

PR #6776, opened 2023-11-08
fix(syslogng): increase buffer sizes

soyacz:
We have problems with missing log lines when there's a flood from scylla log.

Increase buffer sizes so we syslogng survives log bursts.

fixes: #6689

### DIFF
--- a/sdcm/provision/common/utils.py
+++ b/sdcm/provision/common/utils.py
@@ -45,8 +45,8 @@ def configure_syslogng_target_script(host: str, port: int, throttle_per_second: 
         disk_buffer_option=""
         if syslog-ng -V | grep disk; then
             disk_buffer_option="disk-buffer(
-                    mem-buf-size(100000)
-                    disk-buf-size(2000000)
+                    mem-buf-size(1048576)
+                    disk-buf-size(104857600)
                     reliable(yes)
                     dir(\\\"/tmp\\\")
                 )"


fgelcer (approved):
LGTM

soyacz:
I'm giving it a try here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/38/
just in case it broke something badly

fruch (approved):
Lgtm

fgelcer:
> I'm giving it a try here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/38/ just in case it broke something badly

other than increasing memory usage in the Runner, i don't see any other thing that would be affected by this change

soyacz:
> > I'm giving it a try here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/38/ just in case it broke something badly
> 
> other than increasing memory usage in the Runner, i don't see any other thing that would be affected by this change

actually it happens on db/loader nodes. 
Anyway I learned to always test if affect all the jobs. SCT have still corners that I can miss.

fruch:
> > I'm giving it a try here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-many-clients-4h-test/38/ just in case it broke something badly
> 
> other than increasing memory usage in the Runner, i don't see any other thing that would be affected by this change

It's not on the runner side, it might affect it, but this buffer is on the nodes

soyacz:
generally this change works, but still we have eof/Broken pipe (32) errors when lines are long (e.g. 60kb - we have such when running disrupt_add_drop_column nemesis). I don't think it is important now, as no one looks for these lines.

PR #6783, opened 2023-11-09
fix(email_report): pick up proper region values for all backends

vponomaryov:
Current logic of region name pick up is following:
- `use "region_name" (aws) if defined else "gce_datacenter" (gce)`

And it gets done so for each of the backends.
So, fix it by using the 'cluster_backend' type for detecting which config option should be used.


vponomaryov:
Without this fix we have following mess:

Screenshot from 2023-11-09 19-13-32
Screenshot from 2023-11-09 19-15-47

soyacz (approved):
LGTM

PR #6785, opened 2023-11-10
fix(k8s): wait for pods readiness in `add_nodes` method

vponomaryov:
Creating Scylla pods in the `add_nodes` method we wait only for the `Running` state of pods.
In this state Scylla pods are running and Scylla is only starting.
It is ok to do so in case of the single region,
because while a last Scylla pod is being bootstrapped we do other stuff
such as package installation and monitoring configuration.

But running the multiDC setup, it is not acceptable, 
because we create next DB members right after the last one is created in the first region.
As of now, we get following error in the multiDC setup:

```
  ERROR 2023-11-02 09:56:21,418 [shard 0] init - Startup failed: \
    std::runtime_error (Node 10.0.9.215 has gossip status=UNKNOWN. \
    Try fixing it before adding new node to the cluster.)
```

So, fix it by waiting for the pod `readiness` in the `add_nodes` method.

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -2347,7 +2347,11 @@ class PodCluster(cluster.BaseCluster):
 
         # Wait while whole cluster (on all racks) including new nodes are up and running
         current_dc_nodes = [node for node in self.nodes if node.dc_idx == dc_idx]
-        self.wait_for_pods_running(pods_to_wait=count, total_pods=len(current_dc_nodes) + count, dc_idx=dc_idx)
+        expected_dc_nodes_count = len(current_dc_nodes) + count
+        self.wait_for_pods_running(
+            pods_to_wait=count, total_pods=expected_dc_nodes_count, dc_idx=dc_idx)
+        self.wait_for_pods_readiness(
+            pods_to_wait=count, total_pods=expected_dc_nodes_count, dc_idx=dc_idx)
 
         # Register new nodes and return whatever was registered
         k8s_pods = KubernetesOps.list_pods(self.k8s_clusters[dc_idx], namespace=self.namespace)


sdcm/cluster_k8s/__init__.py
```diff
@@ -2347,7 +2347,11 @@ def add_nodes(self,
 
         # Wait while whole cluster (on all racks) including new nodes are up and running
         current_dc_nodes = [node for node in self.nodes if node.dc_idx == dc_idx]
-        self.wait_for_pods_running(pods_to_wait=count, total_pods=len(current_dc_nodes) + count, dc_idx=dc_idx)
+        expected_dc_nodes_count = len(current_dc_nodes) + count
+        self.wait_for_pods_running(
+            pods_to_wait=count, total_pods=expected_dc_nodes_count, dc_idx=dc_idx)
+        self.wait_for_pods_readiness(
```
fruch, line 2353: doesn't that mean there are some unneed calls later one, that would be waiting for readiness ?, or since is a general wait for all of the pods (including loaders and others), there's not point of doing something different for db nodes ?

vponomaryov: Yes, it is general waiter. And it's specifics affects exactly Scylla, because exactly Scylla takes minutes betweeb `Running` and `Ready` pod states.

fruch:
@vponomaryov 

seems like this change is breaking the manger cluster 

https://argus.scylladb.com/test/cc4d518c-2e17-4251-83c2-0bbf7f70ca1c/runs?additionalRuns[]=46e56b9d-d97a-407a-a15a-5c3f5a97f677

vponomaryov:
> @vponomaryov
> 
> seems like this change is breaking the manger cluster
> 
> [https://argus.scylladb.com/test/cc4d518c-2e17-4251-83c2-0bbf7f70ca1c/runs?additionalRuns[]=46e56b9d-d97a-407a-a15a-5c3f5a97f677](https://argus.scylladb.com/test/cc4d518c-2e17-4251-83c2-0bbf7f70ca1c/runs?additionalRuns%5B%5D=46e56b9d-d97a-407a-a15a-5c3f5a97f677)

Here is the fix: https://github.com/scylladb/scylla-cluster-tests/pull/6838

PR #6786, opened 2023-11-10
feature(k8s-eks): support multi K8S clusters creation (multiDC) in EKS

vponomaryov:
List of changes:
- Add new function to the `eks.py` module which should be used for the creation of any new EKS cluster.
- Utilize that new function in the `tester.py` module creating K8S/EKS clusters.
- Use the `params.region_names` attribute of the Tester class instance to define the number of K8S clusters we want to create. It will be equal to the number of configured AWS regions in the `region_name` SCT config option.
- Update the `clean_clusters_eks` function in the `sdcm/utils/common.py` module with the region info.
  Do it for ease of debugging.

What this change doesn't:
- CI job configuration. It is planeed to be added separately.

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -65,7 +65,7 @@ from sdcm.tombstone_gc_verification_thread import TombstoneGcVerificationThread
 from sdcm.utils.alternator.consts import NO_LWT_TABLE_NAME
 from sdcm.utils.aws_kms import AwsKms
 from sdcm.utils.aws_region import AwsRegion
-from sdcm.utils.aws_utils import init_monitoring_info_from_params, get_ec2_network_configuration, get_ec2_services, \
+from sdcm.utils.aws_utils import init_monitoring_info_from_params, get_ec2_services, \
     get_common_params, init_db_info_from_params, ec2_ami_get_root_device_name
 from sdcm.utils.ci_tools import get_job_name, get_job_url
 from sdcm.utils.common import format_timestamp, wait_ami_available, update_certificates, \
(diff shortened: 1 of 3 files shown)


sdcm/tester.py
```diff
@@ -1690,110 +1691,48 @@ def get_cluster_k8s_gke(self):  # pylint: disable=too-many-branches
...
+        for i in range(self.k8s_clusters[0].tenants_number):
             self.db_clusters_multitenant.append(eks.EksScyllaPodCluster(
-                k8s_clusters=[k8s_cluster],
+                k8s_clusters=self.k8s_clusters,
                 scylla_cluster_name=self.params.get("k8s_scylla_cluster_name") + (f"-{i + 1}" if i else ""),
                 user_prefix=(f"{i + 1}-" if i else "") + self.params.get("user_prefix"),
                 n_nodes=self.params.get("k8s_n_scylla_pods_per_cluster") or self.params.get("n_db_nodes"),
                 params=deepcopy(self.params),
-                node_pool_name=scylla_pool.name,
+                node_pool_name=self.k8s_clusters[0].SCYLLA_POOL_NAME,
```
fruch, line 1722: we assume pool names would be identical ?
or this PR doesn't yet have the scylla pods spread in both k8s cluster ?

vponomaryov: Only `auxiliary/default` node pool name differs for the same thing.
The `scylla-pool` name is defined only by us and there is no any need to make it be different for some backends.

fruch (approved):
LGTM

Issue #6790, opened 2023-11-12
left over from manager restore nemesis may not be a fit for other nemeses


fgelcer:
## Issue description

we had a `disrupt_mgmt_restore` nemesis just before `disrupt_add_remove_dc`, and it ended up having a failure :
```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5074, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4390, in disrupt_add_remove_dc
    self._switch_to_network_replication_strategy(self.cluster.get_test_keyspaces() + system_keyspaces)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4369, in _switch_to_network_replication_strategy
    replication_strategy = ReplicationStrategy.get(node, keyspace)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/replication_strategy_utils.py", line 24, in get
    create_ks_statement = node.run_cqlsh(f"describe {keyspace}").stdout.splitlines()[1]
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 2856, in run_cqlsh
    cqlsh_out = self.remoter.run(cmd, timeout=timeout + 120,  # we give 30 seconds to cqlsh timeout mechanism to work
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 613, in run
    result = _run()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 70, in inner
    return func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 604, in _run
    return self._run_execute(cmd, timeout, ignore_status, verbose, new_session, watchers)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 537, in _run_execute
    result = connection.run(**command_kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/libssh2_client/__init__.py", line 620, in run
    return self._complete_run(channel, exception, timeout_reached, timeout, result, warn, stdout, stderr)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/libssh2_client/__init__.py", line 655, in _complete_run
[...]

PR #6791, opened 2023-11-12
improvement(restore nemesis): remove created keyspaces

fgelcer:
once the restore nemesis is done, we don't have
any need or interest in keeping the keyspaces
created by this nemesis.
We have seen also in some specific runs, a different nemesis failing, while trying to access a specific keyspace created by this nemesis, and hence,
it should be completely remove.
Refs: #6790

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2800,43 +2800,57 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             raise UnsupportedNemesis('Scylla-manager configuration is not defined!')
         if self.cluster.params.get('cluster_backend') != 'aws':
             raise UnsupportedNemesis("The restore test only supports AWS at the moment")
-        mgr_cluster = self.cluster.get_cluster_manager()
-        cluster_backend = self.cluster.params.get('cluster_backend')
-        persistent_manager_snapshots_dict = get_persistent_snapshots()
-        target_bucket = persistent_manager_snapshots_dict[cluster_backend]["bucket"]
-        chosen_snapshot_tag, chosen_snapshot_info = choose_snapshot(persistent_manager_snapshots_dict[cluster_backend])
 
-        self.log.info("Restoring the keyspace %s", chosen_snapshot_info["keyspace_name"])
-        location_list = [f"{self.cluster.params.get('backup_bucket_backend')}:{target_bucket}"]
-        test_keyspaces = self.cluster.get_test_keyspaces()
-        if chosen_snapshot_info["keyspace_name"] not in test_keyspaces:
-            self.log.info("Restoring the schema of the keyspace '%s'", chosen_snapshot_info["keyspace_name"])
-            restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=location_list,
+        try:
+            mgr_cluster = self.cluster.get_cluster_manager()
+            cluster_backend = self.cluster.params.get('cluster_backend')
+            persistent_manager_snapshots_dict = get_persistent_snapshots()
+            target_bucket = persistent_manager_snapshots_dict[cluster_backend]["bucket"]
+            chosen_snapshot_tag, chosen_snapshot_info = (
+                choose_snapshot(persistent_manager_snapshots_dict[cluster_backend]))
+
+            self.log.info("Restoring the keyspace %s", chosen_snapshot_info["keyspace_name"])
+            location_list = [f"{self.cluster.params.get('backup_bucket_backend')}:{target_bucket}"]
+            test_keyspaces = self.cluster.get_test_keyspaces()
+            if chosen_snapshot_info["keyspace_name"] not in test_keyspaces:
+                self.log.info("Restoring the schema of the keyspace '%s'",
+                              chosen_snapshot_info["keyspace_name"])
+                restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=location_list,
+                                                               snapshot_tag=chosen_snapshot_tag)
+
+                restore_task.wait_and_get_final_status(step=10,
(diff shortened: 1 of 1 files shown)


fgelcer:
giving it a ride [here](https://jenkins.scylladb.com/job/scylla-staging/job/fabio/job/sct_issue_6790/job/mgmt-restore-individual-nemesis/2/)

sdcm/nemesis.py
```diff
@@ -2799,43 +2799,56 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
+
+            confirmation_stress_template = (
+                persistent_manager_snapshots_dict)[cluster_backend]["confirmation_stress_template"]
+            stress_queue = execute_data_validation_thread(command_template=confirmation_stress_template,
+                                                          keyspace_name=chosen_snapshot_info["keyspace_name"],
+                                                          number_of_rows=chosen_snapshot_info["number_of_rows"])
+            for stress in stress_queue:
+                self.tester.verify_stress_thread(cs_thread_pool=stress)
+        except Exception as ex:
+            raise ex
```
fruch, line 2847: I think those are not needed

soyacz (approved):
LGTM,
That would be nice case for golang style 'defer': https://stackoverflow.com/a/43028386

@fruch maybe we could add such helper to our toolset in sct?

soyacz:
@fgelcer the problem with this keyspace is that yaml containing it's name uses double quotes - while we should not use any or single ones (like in all of the yamls). See defaults/manager_persistent_snapshots.yaml

soyacz:
@fgelcer please rebase to fix precommit

fgelcer:
> @fgelcer the problem with this keyspace is that yaml containing it's name uses double quotes - while we should not use any or single ones (like in all of the yamls). See defaults/manager_persistent_snapshots.yaml

the yaml file holds it as a string, that's why it has double quotes, but from this log entry, you see it is doing what is desired, and it doesn't have the double quotes inside the value of it:
```
sdcm.nemesis.MgmtRestore: Restoring the keyspace 10gb_sizetiered_5_2
```

since the running `cqlsh` commands on the target now were not doing exactly what we needed, because of wrong quotes escaping, i moved to use the Python driver, but it is still not giving us what we needed:
```
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5093, in wrapper
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2852, in disrupt_mgmt_restore
    session.execute(f'DROP KEYSPACE IF EXISTS "{keyspace}"')
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 1799, in execute_verbose
    return execute_orig(*args, **kwargs)
  File "cassandra/cluster.py", line 2699, in cassandra.cluster.Session.execute
  File "cassandra/cluster.py", line 5018, in cassandra.cluster.ResponseFuture.result
cassandra.OperationTimedOut: errors={'10.4.2.108:9042': 'Client request timeout. See Session.execute[_async](timeout)'}, last_host=10.4.2.108:9042
```

what i have done before, was adding a `\` in addition to escaping the double quotes, what ended up in Python being:
```
Command: '/usr/bin/cqlsh --no-color   --request-timeout=180 --connect-timeout=60 --ssl -e "DROP KEYSPACE IF EXISTS \\"10gb_sizetiered_5_2\\"" 10.4.0.124 9042'
```

but it still did not work properly:
```
Exit code: 2

Stdout:



Stderr:

<stdin>:1:OperationTimedOut: errors={'Connection defunct by heartbeat': 'Client request timeout. See Session.execute[_async](timeout)'}, last_host=10.4.0.124:9042
<stdin>:1:Warning: schema version mismatch detected; check the schema versions of your nodes in system.local and system.peers.
```

@soyacz , any lights on this?

fruch:
@fgelcer 
this sound like https://github.com/scylladb/scylla-cqlsh/issues/34

which is known, and currently isn't handled correctly, 
timeout of syncing the schema when query is changing schema

fgelcer:
> @fgelcer this sound like [scylladb/scylla-cqlsh#34](https://github.com/scylladb/scylla-cqlsh/issues/34)
> 
> which is known, and currently isn't handled correctly, timeout of syncing the schema when query is changing schema

if that is the case, we could merge this one

sdcm/nemesis.py
```diff
@@ -2800,43 +2800,57 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
+                                                          number_of_rows=chosen_snapshot_info["number_of_rows"])
+            for stress in stress_queue:
+                self.tester.verify_stress_thread(cs_thread_pool=stress)
+        finally:
+            if chosen_snapshot_info:
+                if "keyspace_name" in chosen_snapshot_info.keys():
+                    keyspace = chosen_snapshot_info["keyspace_name"]
+                    InfoEvent(message=f'Removing test {keyspace=}', severity=Severity.WARNING)
+                    with self.cluster.cql_connection_patient(self.target_node) as session:
+                        session.execute(f'DROP KEYSPACE IF EXISTS "{keyspace}"')
```
soyacz, line 2853: keyspace will be deleted from scylla, but data files will remain in snapshots dir, taking space.
Do we care about it?

fruch: Yes, nemesis shouldn't be leaving snapshots

fruch: Let's open an issue for that and solve it on different PR.

fruch (approved):
Lgtm

sdcm/nemesis.py
soyacz: https://github.com/scylladb/scylla-cluster-tests/issues/6830

soyacz (approved):
LGTM

Issue #6796, opened 2023-11-13
Add ability to add a node during a test which is a different instance type than the one defined in cluster.BaseNode


roydahan:
In some new test cases, e.g. (a [full cluster replacement](https://github.com/scylladb/qa-tasks/issues/1164) with a new instance type), there is a need to support add_node that has a different instance_type than the original nodes of the cluster.

E.g. we have a cluster of i4i.large and we would like to add a new node that is of a type i4i.2xlarge.

fruch:
This PR https://github.com/scylladb/scylla-cluster-tests/pull/6797 is a draft for adding this support.

adding ontop of a configuration, or a hardcode test with the instances to use.

PR #6797, opened 2023-11-13
feature(cluster): add `instance_type` to `add_nodes()` call

fruch:
So now whom ever is calling the `.add_nodes()` can define a diffrent instance type

note: this change doesn't add any configuration option for holding those types, and doesn't do any check that they would fit the backend being used, that on the caller to select the correct types

Closes: #6796

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -2339,7 +2339,9 @@ class PodCluster(cluster.BaseCluster):
                   ec2_user_data: str = "",
                   dc_idx: int = 0,
                   rack: int = 0,
-                  enable_auto_bootstrap: bool = False) -> List[BasePodContainer]:
+                  enable_auto_bootstrap: bool = False,
+                  instance_type=None,
+                  ) -> List[BasePodContainer]:
 
         # TODO: make it work when we have decommissioned (by nodetool) nodes.
         #       Now it will fail because pod which hosts decommissioned Scylla member is reported
@@ -2732,7 +2734,8 @@ class ScyllaPodCluster(cluster.BaseScyllaCluster, PodCluster):  # pylint: disabl
                   # NOTE: 'dc_idx=None' means 'create %count% nodes on each K8S cluster'
                   dc_idx: int = None,
                   rack: int = 0,
-                  enable_auto_bootstrap: bool = False) -> List[BasePodContainer]:
+                  enable_auto_bootstrap: bool = False,
+                  instance_type=None) -> List[BasePodContainer]:
         if dc_idx is None:
             dc_idx = list(range(len(self.k8s_clusters)))
         elif isinstance(dc_idx, int):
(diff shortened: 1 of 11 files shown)


roydahan (dismissed):

sdcm/cluster_k8s/gke.py
```diff
@@ -505,7 +505,8 @@ def add_nodes(self,
                   ec2_user_data: str = "",
                   dc_idx: int = None,
                   rack: int = 0,
-                  enable_auto_bootstrap: bool = False) -> List[GkeScyllaPodContainer]:
+                  enable_auto_bootstrap: bool = False,
+                  instance_type=None) -> List[GkeScyllaPodContainer]:
```
vponomaryov, line 509: In the EKS module the new param gets provided to the following call of the `super()` method.
So, need either to remove there or add it here the same way.

sdcm/cluster_k8s/__init__.py
```diff
@@ -2339,7 +2339,9 @@ def add_nodes(self,
                   ec2_user_data: str = "",
                   dc_idx: int = 0,
                   rack: int = 0,
-                  enable_auto_bootstrap: bool = False) -> List[BasePodContainer]:
+                  enable_auto_bootstrap: bool = False,
```
vponomaryov, line 2342: I think we should add check for K8S methods and raise some exception if it is not `None`.
Why? Because this new parameter is not compatible with theScylla nodes/pods provisioning on K8S platform.

vponomaryov (changes requested):

soyacz (approved):
LGTM

sdcm/cluster_k8s/__init__.py
fruch: added an assert here

sdcm/cluster_k8s/gke.py
fruch: forward the call to the super call, and there would be an assert there

sdcm/cluster_k8s/__init__.py
```diff
@@ -2742,6 +2745,9 @@ def add_nodes(self,
         else:
             count = [count]
         assert len(count) in (1, len(dc_idx))
+
+        assert instance_type is None, "eks can't provision different instance types"
```
vponomaryov, line 2749: This is common code, not EKS-specific. So, message is wrong.

sdcm/cluster_k8s/__init__.py
```diff
@@ -3001,7 +3007,9 @@ def add_nodes(self,
                   # NOTE: 'dc_idx=None' means 'create %count% nodes on each K8S cluster'
                   dc_idx: int = None,
                   rack: int = 0,
-                  enable_auto_bootstrap: bool = False) -> List[BasePodContainer]:
+                  enable_auto_bootstrap: bool = False,
+                  instance_type=None
```
vponomaryov, line 3011: The loaders method doesn't have the assertion for `None` value.

sdcm/cluster_k8s/__init__.py
fruch: Which is fine, we don't support this for and k8s setup

vponomaryov: > Which is fine, we don't support this for and k8s setup

I am not sure I understand your comment.
So, please, just remove word 'eks' from the assertion message.

vponomaryov (approved):
LGTM

PR #6802, opened 2023-11-13
improvement(k8s): handle better the target node change in K8S nemesis

vponomaryov:
Make the `_disrupt_kubernetes_then_decommission_and_add_scylla_node` nemesis method
pick the latest Scylla pod in a rack only in the same region as the current target node.
It will allow us to avoid possible confusion in case of the debugging a failure in the multiDC setup.

Also, remove redefinition of the target node
in the `_disrupt_kubernetes_then_replace_scylla_node` nemesis method.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1392,7 +1392,6 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     def _disrupt_kubernetes_then_replace_scylla_node(self, disruption_method):
         if not self._is_it_on_kubernetes():
             raise UnsupportedNemesis('Supported only on kubernetes')
-        self.set_target_node()
         node = self.target_node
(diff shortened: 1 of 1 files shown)


sdcm/nemesis.py
```diff
@@ -1392,7 +1392,6 @@ def disrupt_terminate_kubernetes_host_then_replace_scylla_node(self):
     def _disrupt_kubernetes_then_replace_scylla_node(self, disruption_method):
         if not self._is_it_on_kubernetes():
             raise UnsupportedNemesis('Supported only on kubernetes')
-        self.set_target_node()
```
fruch, line 1395: Why removing it here ?

fruch: I.e it's not clear where is the duplicate

vponomaryov: The main nemesis logic sets it before giving control to a nemesis.

fruch (approved):
Lgtm

PR #6803, opened 2023-11-14
test(artifact): disable writeback cache setting verification

soyacz:
It was decided to not to set writeback cache setting in perftune, while we specifically verify this setting. This causes artifact test fail.

Idea is to disable this verification temporarily so it does not break artifact tests and enable it back when [the
change](https://github.com/scylladb/scylla-machine-image/pull/394) is merged.

### DIFF
--- a/artifacts_test.py
+++ b/artifacts_test.py
@@ -226,13 +226,6 @@ class ArtifactsTest(ClusterTester):  # pylint: disable=too-many-public-methods
                                 f"matches: {snitch_matches_scylla_yaml}"
                             )
 
-    def verify_write_back_cache_param(self) -> None:
-        if self.params.get("cluster_backend") in ("gce", "azure") and self.params.get("use_preinstalled_scylla"):
-            expected_write_back_cache_param = 0
-        else:
-            expected_write_back_cache_param = None
-        self.assertEqual(self.write_back_cache, expected_write_back_cache_param)
-
     def verify_docker_latest_match_release(self) -> None:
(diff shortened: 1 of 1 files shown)


yaronkaikov (dismissed):

fruch (dismissed):
LGTM

fruch:
@yaronkaikov 

is the change in SMI, is going to 5.4 as well ? 

@soyacz please don't yet merge this (until we must)
maybe we need to remove this check totally

yaronkaikov:
> @yaronkaikov
> 
> is the change in SMI, is going to 5.4 as well ?
> 
> @soyacz please don't yet merge this (until we must) maybe we need to remove this check totally

It shouldn't unless someone will insist

yaronkaikov:
> > @yaronkaikov
> > is the change in SMI, is going to 5.4 as well ?
> > @soyacz please don't yet merge this (until we must) maybe we need to remove this check totally

We need to merge it before i am merging my PR in SMI
> 
> It shouldn't unless someone will insist

fruch:
@soyacz now that i'm thinking on it, lets remove this check, if we are not setting it anymore,
not point of checking we aren't setting it.

(and it would be something we can merge upfront, and also backport/FF without concern)

soyacz:
> @soyacz now that i'm thinking on it, lets remove this check, if we are not setting it anymore, not point of checking we aren't setting it.
> 
> (and it would be something we can merge upfront, and also backport/FF without concern)

done

fruch (approved):
LGTM

PR #6815, opened 2023-11-16
feature(EaR-kms): enable user encryption by default

fruch:
using the feature from scylladb/scylla-enterprise#3389 when we have `scylla_encryption_options` configured with KMS we would enable the `user_info_encryption` and the `system_info_encryption` similar to who it would be working on scylla-cloud

Ref: scylladb/scylla-enterprise#3389

#### Testing
- [x] - 2023.1: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/EaR-longevity-kms-200gb-6h-test/21/
- [x] - enterprise: https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/EaR-longevity-kms-200gb-6h-test/31/
- [x] regular case -  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/27/

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -796,6 +796,8 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         )
 
     def prepare_kms_host(self) -> None:
+        if self.params.is_enterprise and self.params.get('cluster_backend') == 'aws' and not self.params.get('scylla_encryption_options'):
+            self.params['scylla_encryption_options'] = "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'auto'}"  # pylint: disable=line-too-long
         if not (scylla_encryption_options := self.params.get("scylla_encryption_options") or ''):
             return None
         kms_host = (yaml.safe_load(scylla_encryption_options) or {}).get("kms_host") or ''
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -816,6 +816,17 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
             # NOTE: the 'aws_region' will be changed per each node separately depending on it's region
             'aws_region': 'auto'
         }
+        append_scylla_yaml['user_info_encryption'] = {
+            'enabled': True,
+            'key_provider': 'KmsKeyProviderFactory',
+            'kms_host': kms_host,
+        }
+        append_scylla_yaml['system_info_encryption'] = {
+            'enabled': True,
+            'key_provider': 'KmsKeyProviderFactory',
+            'kms_host': kms_host,
+        }
+        self.log.warning("`user_info_encryption` and `system_info_encryption` are configured to use KMS by default")
         self.params["append_scylla_yaml"] = yaml.safe_dump(append_scylla_yaml)
         return None
 
--- a/sdcm/utils/sstable/sstable_utils.py
+++ b/sdcm/utils/sstable/sstable_utils.py
@@ -3,6 +3,10 @@ import json
 import logging
 import random
 
+from sdcm.paths import SCYLLA_YAML_PATH
+from sdcm.utils.version_utils import ComparableScyllaVersion
+from sdcm.exceptions import SstablesNotFound
+
 
 class NonDeletedTombstonesFound(Exception):
     pass
@@ -51,24 +55,41 @@ class SstableUtils:
         self.log.debug('Got %s sstables %s', len(selected_sstables), message)
         return selected_sstables
 
-    def is_sstable_encrypted(self, sstables=None) -> list:
+    def is_sstable_encrypted(self, sstables=None) -> list:  # pylint: disable=too-many-branches
         if not sstables:
             sstables = self.get_sstables()
         if isinstance(sstables, str):
             sstables = [sstables]
         sstables_encrypted_mapping = {}
+
+        if not sstables:
+            raise SstablesNotFound(f"sstables for '{self.keyspace}.{self.table}' wasn't found")
+
+        if ComparableScyllaVersion(self.db_node.scylla_version) >= '2023.2.0~rc0':
+            dump_cmd = (
+                f"{self.db_node.add_install_prefix('/usr/bin/scylla')} sstable dump-scylla-metadata"
+                f" --scylla-yaml-file {self.db_node.add_install_prefix(SCYLLA_YAML_PATH)}"
+                "  --logger-log-level scylla-sstable=debug"
+                f" --keyspace {self.keyspace} --table {self.table} --sstables"
+            )
+        else:
+            dump_cmd = 'sstabledump'
         for sstable in sstables:
             sstables_res = self.db_node.remoter.sudo(
-                f"{self.db_node.add_install_prefix('/usr/bin/scylla')} sstable dump-scylla-metadata"
-                "  --logger-log-level scylla-sstable=debug"
-                f" --sstables {sstable} --keyspace {self.keyspace} --table {self.table}",
+                f"{dump_cmd} {sstable}",
                 ignore_status=True, verbose=True)
+
             self.log.debug("sstables_res.stdout: %s", sstables_res.stdout)
             self.log.debug("sstables_res.stderr: %s", sstables_res.stderr)
             # NOTE: if we have 'stdout' then the data was successfully read and it means there was no encryption
             if sstables_res.stdout:
                 self.log.debug("Successfully read the sstable located at '%s'.", sstable)
-                sstables_encrypted_mapping[sstable] = False
+                if dump_cmd == 'sstabledump':
+                    sstables_encrypted_mapping[sstable] = False
+                else:
+                    scylla_metadata = json.loads(sstables_res.stdout)['sstables']
+                    sstables_encrypted_mapping[sstable] = all('scylla_encryption_options' in metadata.get('extension_attributes', {})
+                                                              for table, metadata in scylla_metadata.items())
             # NOTE: case when sstable exists and it is encrypted:
             #       [shard 0:main] seastar - Exiting on unhandled exception: \
             #       sstables::malformed_sstable_exception (Buffer improperly sized to hold requested data. \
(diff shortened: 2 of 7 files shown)


vponomaryov:
@fruch is is tested?
Also, need to fix unit tests.

roydahan:
@fruch I suggest we also print a log message or something to our exported configuration in bold, mentioning that KMS feature is enabled.
It will be easy to forget about it.

fruch:
> @fruch I suggest we also print a log message or something to our exported configuration in bold, mentioning that KMS feature is enabled.
> It will be easy to forget about it.
> 

What do you mean exported configuration ? To argus ?

Logs I can add, but bold, that's only for Ridge Forester

roydahan:
> What do you mean exported configuration ? To argus ?
Everywhere you can think of :)

roydahan:
@mykaul FYI

sdcm/utils/sstable/sstable_utils.py
```diff
@@ -64,6 +65,7 @@ def is_sstable_encrypted(self, sstables=None) -> list:
 
                 sstables_res = self.db_node.remoter.sudo(
                     f"{self.db_node.add_install_prefix('/usr/bin/scylla')} sstable dump-scylla-metadata"
+                    f" --scylla-yaml-file {self.db_node.add_install_prefix(SCYLLA_YAML_PATH)}"
```
vponomaryov, line 68: Add comment referring to the existing bug: https://github.com/scylladb/scylla-enterprise/issues/3461 ?

sdcm/utils/sstable/sstable_utils.py
```diff
@@ -58,11 +60,18 @@ def is_sstable_encrypted(self, sstables=None) -> list:
             sstables = [sstables]
         sstables_encrypted_mapping = {}
         for sstable in sstables:
-            sstables_res = self.db_node.remoter.sudo(
-                f"{self.db_node.add_install_prefix('/usr/bin/scylla')} sstable dump-scylla-metadata"
-                "  --logger-log-level scylla-sstable=debug"
-                f" --sstables {sstable} --keyspace {self.keyspace} --table {self.table}",
-                ignore_status=True, verbose=True)
+            if ComparableScyllaVersion(self.db_node.scylla_version) >= '2023.2':
```
vponomaryov, line 64: This must be done out of the sstables loop.

fruch: yes, I need to refactor this anyhow, since the logic is gonna be different, since we'll now gonna be ready the output of `scylla sstable`

sdcm/utils/sstable/sstable_utils.py
vponomaryov: The issue is not aligned for 100%, ok.

Is there another issue which describes the error we get not using the `--scylla-yaml-file`?
Why do we need to add the `--scylla-yaml-file` flag?

fruch: looks at the scylla code there is automatic detection of the schema, the scylla yaml is the last option, not the first.
so it might find the schema, but then wouldn't have loaded the configuration for the encryption to work.
@denesb, we might want to consider scylla.yml to be the default ?

so for encryption to work, we have to specific the configuration. as we now do in dtest as well.

sdcm/utils/sstable/sstable_utils.py
fruch: I've refactored it, to be on the top of the function

sdcm/utils/sstable/sstable_utils.py
denesb: Yes, I can make `scylla.yaml` the default. Let me cook a patch.

denesb: Opened an issue to track this effort: https://github.com/scylladb/scylladb/issues/16132

The change is not trivial, but it is not difficult either.

sdcm/utils/sstable/sstable_utils.py
```diff
@@ -51,24 +54,38 @@ def get_sstables(self, from_minutes_ago: int = 0):
...
-    def is_sstable_encrypted(self, sstables=None) -> list:
+    def is_sstable_encrypted(self, sstables=None) -> list:  # pylint: disable=too-many-branches
         if not sstables:
             sstables = self.get_sstables()
         if isinstance(sstables, str):
             sstables = [sstables]
         sstables_encrypted_mapping = {}
-        for sstable in sstables:
-            sstables_res = self.db_node.remoter.sudo(
+        if ComparableScyllaVersion(self.db_node.scylla_version) >= '2023.2':
```
vponomaryov, line 68: Shouldn't the version have the `rc` part?

sdcm/utils/sstable/sstable_utils.py
```diff
@@ -143,7 +147,7 @@ def get_table_repair_date(self) -> str | None:
                                                     user=self.user, password=self.password) as session:
             try:
                 query = f"SELECT repair_time from system.repair_history WHERE keyspace_name = '{self.keyspace}' " \
-                        f"AND table_name = '{self.table}' ALLOW FILTERING;"
+                    f"AND table_name = '{self.table}' ALLOW FILTERING;"
```
vponomaryov, line 150: Is it intentional change?

sdcm/tester.py
```diff
@@ -796,6 +796,11 @@ def is_encrypt_keys_needed(self):
         )
 
     def prepare_kms_host(self) -> None:
+        if self.params.is_enterprise and self.params.get('cluster_backend') == 'aws' and not self.params.get('scylla_encryption_options'):
```
vponomaryov, line 799: Why using that implicit way?
Why not set it in the `default` configuration?

sdcm/tester.py
```diff
@@ -796,6 +796,11 @@ def is_encrypt_keys_needed(self):
...
+            self.params['scylla_encryption_options'] = \
```
vponomaryov, line 800: Please, do not use backslashes.
Please, keep key and value on the same lines.

sdcm/utils/sstable/sstable_utils.py
```diff
@@ -83,9 +88,8 @@ def is_sstable_encrypted(self, sstables=None) -> list:  # pylint: disable=too-ma
                     sstables_encrypted_mapping[sstable] = False
                 else:
                     scylla_metadata = json.loads(sstables_res.stdout)['sstables']
-                    sstables_encrypted_mapping[sstable] = \
-                        all('scylla_encryption_options' in metadata.get('extension_attributes', {})
-                            for table, metadata in scylla_metadata.items())
+                    sstables_encrypted_mapping[sstable] = all('scylla_encryption_options' in metadata.get('extension_attributes', {})
```
vponomaryov, line 91: Please, use this line state in the original commit of implementation in this PR.

vponomaryov (changes requested):
Please, see inline comments and also need to address the precommit failures.

sdcm/utils/sstable/sstable_utils.py
fruch: It should...

It's quite an annoying caveat, and I'm guessing it's not the last time we're gonna miss it

sdcm/utils/sstable/sstable_utils.py
fruch: Probably not

sdcm/utils/sstable/sstable_utils.py
fruch: Not sure what you meant

vponomaryov: This line gets added in one commit and updated in another without real reason.
I propose to use the latest state in the first commit so second one would not change this line at all.

sdcm/utils/sstable/sstable_utils.py
vponomaryov: GA versions are always newer than the `RC` ones.
So, the behavior is correct.

sdcm/utils/sstable/sstable_utils.py
fruch: thanks @denesb

sdcm/utils/sstable/sstable_utils.py
fruch: changed

sdcm/utils/sstable/sstable_utils.py
fruch: reverted

sdcm/tester.py
fruch: you'll have to explain it further, since I'm not sure what do you mean by that.

sdcm/utils/sstable/sstable_utils.py
fruch: fixed

sdcm/tester.py
fruch: keeping it all in the same line, it becomes a bit long...
no sure why to insist on it. but I don't care.

sdcm/tester.py
vponomaryov: New config option may be added to define the need to configure encryption.
In this case it will become `explicit` way.
For now, everyone will be confused to see encryption in place not having it in configuration.

vponomaryov: @roy WDYT? ^

fruch: again the whole idea is to enable it by default for all of enterprise runs on AWS, I don't see how putting something in to configuration would change the confusion here.

adding one more configuration value, just for this default won't be a big change, and still gonna be confusing.

sdcm/tester.py
```diff
@@ -796,6 +796,8 @@ def is_encrypt_keys_needed(self):
         )
 
     def prepare_kms_host(self) -> None:
+        if self.params.is_enterprise and self.params.get('cluster_backend') == 'aws' and not self.params.get('scylla_encryption_options'):
+            self.params['scylla_encryption_options'] = "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmsKeyProviderFactory', 'kms_host': 'auto'}"  # pylint: disable=line-too-long
```
roydahan, line 800: Probably would be better to read these from a configuration file.
But can be done in a followup PR if we agree on that.

roydahan (approved):

PR #6817, opened 2023-11-16
fix(sct_config): validate images/ami aren't empty

fruch:
`_check_version_supplied` validation was changing
if we have image/ami or `scylla_repo`

and cases of not having an iamge, but having `scylla_repo` we get till instance creation to find out we don't have image provided,

even when `scylla_repo` provied we need a base image to install ontop of (artifacts/upgrade has those in thier pipeline, longevity and other doesn't)


fruch:
the integration test is fail cause of a know unstable test with ycsb
beside that everything was working o.k.

soyacz (approved):
lgtm

PR #6822, opened 2023-11-17
improvement(k8s): use bigger K8S pod readiness timeout

vponomaryov:
Mainly we run K8S CI job with really small load.
Small load allows us to get Scylla pods up and running much faster than when we have big load.

So, increase the K8S pod readiness timeout from '5' to '10' to cover upcoming new CI jobs with bigger load.

### DIFF
--- a/sdcm/cluster_k8s/__init__.py
+++ b/sdcm/cluster_k8s/__init__.py
@@ -1762,7 +1762,7 @@ class BasePodContainer(cluster.BaseNode):  # pylint: disable=too-many-public-met
     parent_cluster: PodCluster
 
     pod_readiness_delay = 30  # seconds
-    pod_readiness_timeout = 5  # minutes
+    pod_readiness_timeout = 10  # minutes
     pod_terminate_timeout = 5  # minutes
 
     def __init__(self, name: str, parent_cluster: PodCluster, node_prefix: str = "node", node_index: int = 1,


vponomaryov:
It will allow us to minimize following timeout errors:

```
Command: 'kubectl --kubeconfig=%path%/.kube/config --cache-dir=%path%/.kube/http-cache--long-eks-multidc-3h-dev-acfde8-2 --namespace=scylla \
            rollout status statefulset/sct-cluster-eu-west-1-rack-1 --watch=true --timeout=60m'

Exit code: 1

Stdout:

Waiting for 1 pods to be ready...

Stderr:

error: timed out waiting for the condition
```

fruch:
We should consider how to apply AdaptiveTimeout here, and then we can make it relative to the data size or load.

vponomaryov:
> We should consider how to apply AdaptiveTimeout here, and then we can make it relative to the data size or load.

Sure we can, but I think we can do it after merge of this one.
The adaptive timeout is not really needed if we need to add just several minutes.

If some job with low load fails due to some other reason (rare case) then we will wait on for 5 minutes more.
So, I would set `nice to have` priority to the `adaptive timeout` case in here.

soyacz (approved):
LGTM

fruch:
> > We should consider how to apply AdaptiveTimeout here, and then we can make it relative to the data size or load.
> 
> Sure we can, but I think we can do it after merge of this one.
> The adaptive timeout is not really needed if we need to add just several minutes.
> 
> If some job with low load fails due to some other reason (rare case) then we will wait only for 5 minutes more.
> So, I would set `nice to have` priority to the `adaptive timeout` case in here.

I agree it's not urgent, but on the other hand we are not running that big load cases that often (maybe we should, replace one of weekly runs with something from tier1 we run on VMs)

fruch (approved):
Lgtm

PR #6828, opened 2023-11-19
fix(upgrade-test): enable back the usage of sstabledump

fruch:
since in older versions i.e. 2023.1 and lower, we
can use `scylla sstable` tools yet, we should revert to use the old `sstabledump` tool

#### Testing
- [x] https://jenkins.scylladb.com/job/enterprise-2023.1/job/branch_2023.1_future/job/rolling-upgrade-ami-arm-test/3/

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -33,7 +33,10 @@ from sdcm.fill_db_data import FillDatabaseData
 from sdcm.sct_events import Severity
 from sdcm.stress_thread import CassandraStressThread
 from sdcm.utils.user_profile import get_profile_content
-from sdcm.utils.version_utils import get_node_supported_sstable_versions
+from sdcm.utils.version_utils import (
+    get_node_supported_sstable_versions,
+    ComparableScyllaVersion,
+)
 from sdcm.sct_events.system import InfoEvent
 from sdcm.sct_events.database import (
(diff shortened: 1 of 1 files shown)


soyacz (approved):
if works, LGTM

juliayakovlev (approved):

fruch:
> if works, LGTM

trying it out, we'll knew in few hours

upgrade_test.py
```diff
@@ -765,11 +768,15 @@ def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-stat
 
         # Verify sstabledump
         InfoEvent(message='Starting sstabledump to verify correctness of sstables').publish()
+        if ComparableScyllaVersion(self.db_cluster.nodes[0].scylla_version) < "2023.2" or \
+                ComparableScyllaVersion(self.db_cluster.nodes[0].scylla_version) < "5.4":
```
vponomaryov, line 772: This condition will always provide `True` for any OSS version.
It is bug.
Need to reorder condition parts or add a check for enterprise-type in addition to existing conditions.

fruch: good catch, as some point we might want to add something to ComparableScyllaVersion, to help doing those in less verbose manner

PR #6844, opened 2023-11-22
fix(k8s): make mgmt work in K8S multiDC setup

vponomaryov:
To make Scylla-Manager work in MultiDC K8S setups we should do following [1]:
- Install it only in one region
- Make Scylla pods use auth token from that the only scylla-manager.

So, do above changes and update the multiDC config files. Make 3h config write twice less data so the 'mgmt backup' could fit the 3h time range in case one wants to test it using that job.

[1] https://github.com/scylladb/scylla-operator/blob/master/docs/source/multidc/multidc.md#scylla-manager


vponomaryov:
Tested here:  https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-multidc-3h-eks/56/consoleFull

Screenshot from 2023-11-22 17-04-25

The `mgmt repair cli` nemesis was running `ok`, just didn't fit the test run length.

fruch (approved):
LGTM

PR #6846, opened 2023-11-23
fix(upgrade-test): clear the cached scylla version upon upgrade

tchaikov:
the scylla_version is a cached property of BaseNode. once the property is set, the property stays with the owner instance unless it is cleared with `forget_scylla_version()` method. this method is called in the test of `test_kubernetes_scylla_upgrade()` after all nodes are upgraded. but we also do rolling uprade in `test_rolling_upgrade`, which uses sstabledump or
`scylla sstable dump-data` to verify the correctness of sstabels after upgrading some nodes in the cluster on first node which has upgrade to the version being tested. but if we upgrade from an older version which do not support uuid-identifier sstable and to a version which support this feature. we are still using the cached version, which implies that it does not support this feature, and we can use the java-based sstabledump for reading the sstables. but unfortunately, the sstables have been upgraded to the newer supported version and their identifiers (filenames) are also the uuid-based ones. this is why the upgrade test from pre-5.4 to 5.4 fails with this very test.

in this change, we clear the cached version number when a certain node is upgraded.

Fixes https://github.com/scylladb/scylladb/issues/16147

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -329,6 +329,7 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
         new_ver = result.stdout.strip()
         assert self.orig_ver != new_ver, "scylla-server version isn't changed"
         self.new_ver = new_ver
+        node.forget_scylla_version()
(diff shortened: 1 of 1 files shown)


vponomaryov (approved):
LGTM

Issue #6861, opened 2023-11-28
Start of node exporter (via docker run) fails


mykaul:
```
docker: Error response from daemon: Get "https://quay.io/v2/": context deadline exceeded (Client.Timeout exceeded while awaiting headers).
See 'docker run --help'.
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 3661, in node_setup
cl_inst.node_setup(_node, **setup_kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 5123, in node_setup
self.install_scylla_monitoring(node)
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 119, in inner
res = func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 5466, in install_scylla_monitoring
self.start_node_exporter(node)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 5254, in start_node_exporter
node.remoter.run("bash -ce '%s'" % start_node_exporter_script)
File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 614, in run
result = _run()
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 70, in inner
return func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 605, in _run
return self._run_execute(cmd, timeout, ignore_status, verbose, new_session, watchers)
File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 538, in _run_execute
result = connection.run(**command_kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/libssh2_client/__init__.py", line 620, in run
return self._complete_run(channel, exception, timeout_reached, timeout, result, warn, stdout, stderr)
File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/libssh2_client/__init__.py", line 655, in _complete_run
raise UnexpectedExit(result)
sdcm.remote.libssh2_client.exceptions.UnexpectedExit: Encountered a bad command exit code!
[...]

PR #6862, opened 2023-11-28
fix(monitor): install node_exporter instead of using docker

fruch:
at some point we started using docker for running the node_exporter
on the monitoring machine, but since then we have implemented
installation of node_exporter for loader nodes.

we are switching to that, since the docker based options
seems to be unstable (since 2020 it was failing, and wasn't tracked)

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/6861

#### Testing
- [x] https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/31/

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5223,6 +5223,9 @@ class BaseMonitorSet:  # pylint: disable=too-many-public-methods,too-many-instan
         else:
             raise ValueError('Unsupported Distribution type: {}'.format(str(node.distro)))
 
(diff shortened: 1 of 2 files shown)


fruch:
> @fruch shouldn't the test CI job use `docker` backend?

cause of that if that was there ?

maybe I'll do a swipe of provision tests, once I'll confirm it works as expected for AWS run

vponomaryov (approved):
LGTM

PR #6864, opened 2023-11-28
fix(perf-regression): use `stop_test_on_stress_failure: false`

fruch:
we don't want to stop those test on case of failures like this we still want to finish the test and report the results even if some queries were failing


roydahan:
It's even worse in performance test because it will publish irrelevant results.

PR #6866, opened 2023-11-28
feature(create_test_release_jobs): create whole tree structure

fruch:
Since we want to re-organize our pipelines, we are introducing a function that can create all pipeline based on direcotry structure

* supports freestyle jobs (i.e. for triggers to be in the code)
* operator job creation is using it, it was doing something similar before


vponomaryov:
@fruch I am not sure I understand the problem you are solving here.
Existing code creates tree of jobs already. So, what exactly gets changed?

fruch:
> @fruch I am not sure I understand the problem you are solving here.
> Existing code creates tree of jobs already. So, what exactly gets changed?

Operator jobs were, the rest wasn't

vponomaryov:
> > @fruch I am not sure I understand the problem you are solving here.
> > Existing code creates tree of jobs already. So, what exactly gets changed?
> 
> Operator jobs were, the rest wasn't

`The rest` doesn't use the new `create_job_tree` function. What am I missing here?

fruch:
> > > @fruch I am not sure I understand the problem you are solving here.
> > > Existing code creates tree of jobs already. So, what exactly gets changed?
> > 
> > Operator jobs were, the rest wasn't
> 
> `The rest` doesn't use the new `create_job_tree` function. What am I missing here?

The next commit that I didn't pushed :)

vponomaryov (approved):
LGTM

fruch:
@roydahan 

I'm looking at the perf tests, and I was trying to setup them up in a way we won't be moving them( with the exception of the simple query microbenchmark, which for some reason is on top level folder)

image

I'm using symbolic links between enterprise and master

what should we do with the rest of the perf pipelines ? put them in a folder that isn't generating anything on jenkins ?
or
should we generate them into a folder reproducers/staging ?

I'm guessing we won't have perf jobs for releases

fruch:
@roydahan 

As we discussed, that we are not going to be working on moving jobs in jenkins/Argus
so we can commence with this one, and getting it working on master - aligning any trigger need on scylla-pkg

- then we can do 2nd phase of arranging into tiers folder
- we can use symlink for case we might want same job in two folders (we are doing so in performance folders)

fruch:
> When did we discuss that? I thought we want to use the opportunity to move things in Jenkins as well.

we talked about moving jobs to preserve their history, and we agreed we don't need todo that.
if you have any suggestion of wishes to refactor this statuture, it can be in follow ups PR, or state exactly what are those changes, this whole PR propuse it to enable those changes.

fruch:
@roydahan 

here's is the new tree structure (including the names as in Argus in 2024.1)
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/create_repo/

Should we keep the RBNO folder, and the raft experimental folder ? to we can drop them completely ?

other then that commence with this change next week.
for an actual release, we need to consider running it without creating triggers, since most of them are aimed at master...

maybe the flow of triggers creation should be separated completely from the jobs ?

roydahan:
> @roydahan
> 
> here's is the new tree structure (including the names as in Argus in 2024.1) https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/create_repo/
> 
> Should we keep the RBNO folder, and the raft experimental folder ? to we can drop them completely ?

RBNO can be removed.
Raft Experimental can also be removed, assuming that @temichus is taking care of adding it back as part of changes experimental flag.
If we're talking about next release, so also "Raft Disabled" can be removed since from next release it won't be possible.

> 
> other then that commence with this change next week. for an actual release, we need to consider running it without creating triggers, since most of them are aimed at master...

I didn't understand this part.
> 
> maybe the flow of triggers creation should be separated completely from the jobs ?

#

fruch:
> > @roydahan
> > 
> > here's is the new tree structure (including the names as in Argus in 2024.1) https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/create_repo/
> > 
> > Should we keep the RBNO folder, and the raft experimental folder ? to we can drop them completely ?
> 
> RBNO can be removed.
> Raft Experimental can also be removed, assuming that @temichus is taking care of adding it back as part of changes experimental flag.
> If we're talking about next release, so also "Raft Disabled" can be removed since from next release it won't be possible.

We aren't going to backport this change into any current release, so I'll remove those 3 folders.

And we are set to go

> 
> > 
> > other then that commence with this change next week. for an actual release, we need to consider running it without creating triggers, since most of them are aimed at master...
> 
> I didn't understand this part.

Now that I've introduced the triggers into the code, we have triggers we want for the release, and triggers we want only for master, like weekly tier1, or individual nemesis nighty, which we don't want want to start triggering on release.

I've addressed that by splitting the trigger for master only into it's own folder, that we update only for master.



> > 
> > maybe the flow of triggers creation should be separated completely from the jobs ?
> 
> #

fruch:
One side effect on all teams, would be that jobs in staging would need to be updated (and it won't be the same as current releases), to the correct paths in code.
so once this get merged, most of the jobs people have in staging wouldn't run.

roydahan:
But most of the jobs in staging are done with forks, branches.

fruch:
> But most of the jobs in staging are done with forks, branches.

any work start would start ontop of master, would have this PR, once it's merged.
people are working/developing on master, mostly, at least I do.

PR #6871, opened 2023-11-29
fix(cleanup-api): enlarge timeout to 10min

fruch:
the default timeout for `run_remoter_curl` is 2min, and seems like in some cases that's not enough and we might
get into the retry and get the following errors:

```
cleanup request failed: there is an ongoing cleanup on keyspace1.standard1
```

we enlarge it to 10min, and if it would be needed we'll apply AdaptiveTimeout for it.

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -747,7 +747,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         of the command in (4).
         """
         compaction_args = self._prepare_start_stop_compaction()
-        trigger_func = partial(compaction_args.compaction_ops.trigger_cleanup_compaction)
+        trigger_func = partial(compaction_args.compaction_ops.trigger_cleanup_compaction, timeout=500)
         watch_func = partial(compaction_args.compaction_ops.stop_on_user_compaction_logged,
                              node=compaction_args.target_node,
                              mark=compaction_args.target_node.mark_log(),
--- a/sdcm/rest/storage_service_client.py
+++ b/sdcm/rest/storage_service_client.py
@@ -29,11 +29,11 @@ class StorageServiceClient(RemoteCurlClient):
 
         return self.run_remoter_curl(method="POST", path=path, params=params, timeout=360)
 
-    def cleanup_ks_cf(self, keyspace: str, cf: Optional[str] = None) -> Result:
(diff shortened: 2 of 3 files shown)


sdcm/nemesis.py
```diff
@@ -747,7 +747,7 @@ def disrupt_start_stop_cleanup_compaction(self):
         of the command in (4).
         """
         compaction_args = self._prepare_start_stop_compaction()
-        trigger_func = partial(compaction_args.compaction_ops.trigger_cleanup_compaction)
+        trigger_func = partial(compaction_args.compaction_ops.trigger_cleanup_compaction, timeout=500)
```
vponomaryov, line 750: `500s` is not `10m`.
`600s` is `10m`.

So, need either update the commit description or change the timeout value to be `600` seconds.

sdcm/rest/storage_service_client.py
```diff
@@ -29,11 +29,11 @@ def compact_ks_cf(self, keyspace: str, cf: Optional[str] = None) -> Result:
 
         return self.run_remoter_curl(method="POST", path=path, params=params, timeout=360)
 
-    def cleanup_ks_cf(self, keyspace: str, cf: Optional[str] = None) -> Result:
+    def cleanup_ks_cf(self, keyspace: str, cf: Optional[str] = None, timeout: int = 500) -> Result:
```
vponomaryov, line 32: same here.

sdcm/utils/compaction_ops.py
```diff
@@ -52,8 +52,8 @@ def trigger_scrub_compaction(self,
 
         return self.storage_service_client.scrub_ks_cf(**params)
 
-    def trigger_cleanup_compaction(self, keyspace: str = "keyspace1", cf: str = "standard1") -> Result:
-        return self.storage_service_client.cleanup_ks_cf(keyspace=keyspace, cf=cf)
+    def trigger_cleanup_compaction(self, keyspace: str = "keyspace1", cf: str = "standard1", timeout: int = 500) -> Result:
```
vponomaryov, line 55: And here.

sdcm/rest/storage_service_client.py
fruch: change to 600

sdcm/utils/compaction_ops.py
fruch: change to 600

PR #6879, opened 2023-11-30
fix(cassandra-stress): wait for first stress to create schema

fruch:
up until now, we had a 30sec sleep after the first stress run in
each stress thread, when round_robin was used with multiple commands
we still get to two stress command racing at the begining
and might fail on schema creation.

there is an ongoing work, as maybe fix the c-s code, but for the
sake of supporting older c-s version (ones we use in perf branches)
we introduce a new way to wait for the first load to start (via
checking metrics), and only then release other loads

#### Testing
- [ ] 200gb round robin: https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/33/

### DIFF
--- a/sdcm/stress_thread.py
+++ b/sdcm/stress_thread.py
@@ -17,6 +17,7 @@ import time
 import uuid
 import logging
 import contextlib
+import threading
 from typing import Any
 from itertools import chain
 from pathlib import Path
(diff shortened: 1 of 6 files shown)


sdcm/stress_thread.py
```diff
@@ -343,11 +350,17 @@ def run(self):
...
+                        CassandraStressThread.should_wait_for_schema = True
+
                     self.results_futures += [self.executor.submit(self._run_cs_stress,
                                                                   *(loader, loader_idx, cpu_idx, ks_idx))]
-                    if loader_idx == 0 and cpu_idx == 0 and self.max_workers > 1:
-                        # Wait for first stress thread to create the schema, before spawning new stress threads
-                        time.sleep(30)
+                    if _wait_for_schema and self.prometheus_db:
+                        self.prometheus_db.metric_has_data(ANY_CS_METEIC_QUERY, n=10)
+                        self.schema_created.set()
```
roydahan, line 363: Now that we have consistent_schema_management, I don't think we should do it.
The main reasons are:
1. It's a good place where we actually test this part of the code.
2. It can create us a problem, especially in performance tests where we want all stress commands to run at the same time.
With this fix, I'm afraid that the first stress will run few minutes before the others and our stats in the beginning and in the end of the tests will be less accurate.

fruch: 1. we seen issues even with `consistent_schema_management: false`, once we have a fix we can remove.

2. regardless, this option is better then the 30sec sleep we have now.
    removing the sleep would create issues when using multiplier 

this implementation has one assumption, that there is only one schema we care about, i.e. if there would be multiple keyspace/tables, it would work only on the first one.

fruch:
we can drop this, it was a work around for an actual bug...
and hopefully it already been fixed

Issue #6119, opened 2023-05-09
SCT generates false negative RUNTIME_ERROR events running `decommission` operation in K8S


vponomaryov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: any?

### Logs
* test_id: 9c3e6d95-056c-4bae-be3a-9140da5e4d52
* job log: [Argus](https://argus.scylladb.com/test/028aa2df-5ce5-4883-a4a8-16bfbbe8dcc8/runs?additionalRuns%5B%5D=9c3e6d95-056c-4bae-be3a-9140da5e4d52) , [CI job](https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/eks/job/longevity-scylla-operator-3h-eks/76/consoleFull)

### Description

The `RUNTIME_ERROR` DatabaseLogEvent has following definition:

```
DatabaseLogEvent.add_subevent_type("RUNTIME_ERROR", severity=Severity.ERROR,                        
                                   regex="std::runtime_error")
```

And it catches false negative error messages described here: https://github.com/scylladb/scylla-operator/issues/1237

As a result it creates `false ERROR SCT events` for the following duplicated operator `decommission` API calls:
```
INFO  2023-04-30 02:07:35,170 [shard  3] api - decommission
{"L":"INFO","T":"2023-04-30T02:07:35.173Z","N":"scylla_client","M":"HTTP","host":"localhost:10000","method":"POST","uri":"/storage_service/decommission","duration":"5ms","status":500,"bytes":97,"dump":"HTTP/1.1 500 Internal Server Error\r\nContent-Length: 97\r\nContent-Type: application/json\r\nDate: Sun, 30 Apr 2023 02:07:35 GMT\r\nServer: Seastar httpd\r\n\r\n{\"message\": \"std::runtime_error (Operation decommission is in progress, try again)\", \"code\": 500}"}
E0430 02:07:35.177395       1 sidecar/controller.go:141] syncing key 'scylla/sct-cluster-us-east1-b-us-east1-1-0' failed: can't decommision a node: can't decommission the node: agent [HTTP 500] std::runtime_error (Operation decommission is in progress, try again)
```

So, need to make SCT be smart enough to workaround the operator's bug.

### Steps to Reproduce

[...]

PR #6880, opened 2023-11-30
fix(k8s): ignore decomission errors from operator agent logs

fruch:
Since we have issues with doing this correctly on the specific place it might be happening, we'll ignore this one globaly

Ref: scylladb/scylla-cluster-tests#6119

### DIFF
--- a/sdcm/sct_events/setup.py
+++ b/sdcm/sct_events/setup.py
@@ -73,6 +73,13 @@ def start_events_device(log_dir: Optional[Union[str, Path]] = None,
     EventsSeverityChangerFilter(new_severity=Severity.WARNING,
                                 event_class=DatabaseLogEvent.RUNTIME_ERROR,
                                 regex='.*view update generator not plugged to push updates').publish()
+
+    # cause of issue https://github.com/scylladb/scylla-cluster-tests/issues/6119
+    EventsSeverityChangerFilter(new_severity=Severity.WARNING,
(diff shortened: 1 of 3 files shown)


sdcm/sct_events/setup.py
```diff
@@ -73,6 +73,13 @@ def start_events_device(log_dir: Optional[Union[str, Path]] = None,
     EventsSeverityChangerFilter(new_severity=Severity.WARNING,
                                 event_class=DatabaseLogEvent.RUNTIME_ERROR,
                                 regex='.*view update generator not plugged to push updates').publish()
+
+    # cause of issue https://github.com/scylladb/scylla-cluster-tests/issues/6119
+    EventsSeverityChangerFilter(new_severity=Severity.WARNING,
+                                event_class=DatabaseLogEvent.RUNTIME_ERROR,
+                                regex=r'.*sidecar/controller.go.*std::runtime_error '
+                                      r'\(Operation decommission is in progress, try again\)').publish()
```
vponomaryov, line 81: Shouldn't the regex end with `.*`? Or add one more symbol `]`?

fruch: unittest show this line was caught, good enough for me

fruch: I learnt the hard way we should check any of those.

cause we tend todo them multiple times

vponomaryov (approved):
LGTM

Issue #6565, opened 2023-09-03
post-prepare-cql fails after disrupt_create_index: "Index si_text is a duplicate of existing index test_v_nemesis"


yarongilor:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

The create-index nemesis ran before post-prepare-cql:
```
< t:2023-08-31 09:09:04,530 f:nemesis.py      l:1736 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.SisyphusMonkey: >>>>>>>>>>>>>Started random_disrupt_method create_index
< t:2023-08-31 09:09:45,955 f:common.py       l:1750 c:utils                p:DEBUG > Executing CQL 'CREATE INDEX test_v_nemesis ON scylla_bench.test("v")' ...
< t:2023-08-31 09:09:45,960 f:common.py       l:1758 c:utils                p:DEBUG > Executing CQL 'CREATE INDEX test_v_nemesis ON scylla_bench.test("v")' ...
```
then post-prepare-cql failed with: 
```
2023-08-31 09:15:37.584: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=53b45899-6b93-4012-9d33-e42316b7818d, source=LongevityTest.test_custom_time (longevity_test.LongevityTest)() message=Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/longevity_test.py", line 106, in test_custom_time
self.run_prepare_write_cmd()
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/loader_utils.py", line 209, in run_prepare_write_cmd
self.run_post_prepare_cql_cmds()
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/loader_utils.py", line 152, in run_post_prepare_cql_cmds
self._run_cql_commands(post_prepare_cql_cmds)
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/loader_utils.py", line 143, in _run_cql_commands
session.execute(cmd)
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 1751, in execute_verbose
return execute_orig(*args, **kwargs)
File "cassandra/cluster.py", line 2699, in cassandra.cluster.Session.execute
File "cassandra/cluster.py", line 5018, in cassandra.cluster.ResponseFuture.result
[...]

PR #6886, opened 2023-12-03
fix(longevity): remove MV creation in large partition tests

juliayakovlev:
Remove MV and SI creation during 'post-prepare-cql' as it may cause to index nemesis failure.

The problem that scylla_bench.test table has few columns for few indexes creation.
So post-prepare-cql or index nemesis (depend how was run first) may fail with error like:
```
cassandra.InvalidRequest: Error from server: code=2200 [Invalid query] message="Index si_text is a duplicate of existing index test_v_nemesis"
```

Meanwhile it was decided do not perform full fix of the issue and just remove MV/SI creation from 'post-prepare-cql' section.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/6565

### DIFF
--- a/test-cases/longevity/longevity-large-partition-3h.yaml
+++ b/test-cases/longevity/longevity-large-partition-3h.yaml
@@ -28,9 +28,7 @@ stress_cmd: ["scylla-bench -workload=sequential -mode=write -replication-factor=
              "scylla-bench -workload=sequential -mode=read  -replication-factor=3 -partition-count=25 -clustering-row-count=5555 -partition-offset=76   -clustering-row-size=uniform:1024..2048   -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 0 -duration=170m -validate-data"
             ]
 
-post_prepare_cql_cmds: ["CREATE INDEX si_text ON scylla_bench.test(v)",
-                        "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where ck IS NOT NULL AND pk is not null PRIMARY KEY (pk, ck) with comment = 'TEST VIEW'"
-                       ]
+post_prepare_cql_cmds: "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where ck IS NOT NULL AND pk is not null PRIMARY KEY (pk, ck) with comment = 'TEST VIEW'"
 
 n_db_nodes: 5
 n_loaders: 4


nyh:
> Remove MV and SI creation during 'post-prepare-cql' as it may cause to index nemesis failure. The problem that scylla_bench.test table has so few columns for few indexes creation.
> 
> Fixes: #6565

Obviously, not adding views or indexes will "fix" any view-related failures. But I didn't understand what your commit message claims. If this is a bug in the test, and you don't want to spend time fixing this bug because this test shouldn't be about views, then fine - but then mention the test bug and say there's no need to fix it because we don't want to test views here. The commit message doesn't say, and to be honest, I didn't understand what it said.

Anyway, to the actual matter: The error "duplicate of existing index" you saw means your test tried to index **the same column that was already indexed** (specifically, column `v`) with a different index name, but this is not allowed. So the problem is not one piece of code that creates the index, but the fact that your test apparently has two pieces of test code which enable an index on the same column but with the same name. Why does your test do that? It's a bug in the test. Can't it be fixed? But again, if you don't care about views in this test at all, it's fine to remove them. Just explain why you don't care about them.

juliayakovlev:
> > Remove MV and SI creation during 'post-prepare-cql' as it may cause to index nemesis failure. The problem that scylla_bench.test table has so few columns for few indexes creation.
> > Fixes: #6565
> 
> Obviously, not adding views or indexes will "fix" any view-related failures. But I didn't understand what your commit message claims. If this is a bug in the test, and you don't want to spend time fixing this bug because this test shouldn't be about views, then fine - but then mention the test bug and say there's no need to fix it because we don't want to test views here. The commit message doesn't say, and to be honest, I didn't understand what it said.
> 
> Anyway, to the actual matter: The error "duplicate of existing index" you saw means your test tried to index **the same column that was already indexed** (specifically, column `v`) with a different index name, but this is not allowed. So the problem is not one piece of code that creates the index, but the fact that your test apparently has two pieces of test code which enable an index on the same column but with the same name. Why does your test do that? It's a bug in the test. Can't it be fixed? But again, if you don't care about views in this test at all, it's fine to remove them. Just explain why you don't care about them.

Updated commit message

nyh:
I still don't understand the text of the commit message. You said "The problem that scylla_bench.test table has few columns for few indexes creation." - what does this mean? Did you mean (but didn't write...) the test has a bug where it tries to index the same column with multiple index names, and it's hard to fix this bug so it's easier to just remove the index? (and also the non-index view, by the way... why?)

And can you include a statement about what we lose from the testing power by not doing MV in this test. Do we not lose anything because there are other 17 tests which already test MV? Or conversely, is it a big loss because this is the only test of MV with large base partitions? (?)

If you and @yarongilor and @fruch agreed say that it's not important that this test has an MV, I believe you, and the patch is fine. But it will be nice to see an explanation that doesn't involve reading all the discussion in https://github.com/scylladb/scylla-cluster-tests/issues/6565 (which I didn't).

roydahan (dismissed):
This is not the direction I want.
We added these with intention to catch some of the issues we had in the field.
We didn’t want to rely on the nemesis to catch it.

the fix should be the other way around, you can check if an index already exist and if it does, don’t create it.

fruch:
> This is not the direction I want. We added these with intention to catch some of the issues we had in the field. We didn’t want to rely on the nemesis to catch it.
> 
> the fix should be the other way around, you can check if an index already exist and if it does, don’t create it.

the way those index/view are is introduced in `post_prepare_cql_cmds` that for some of the case that do nemesis during prepare, it would happen before any logic can be apply, those can apply any logic.

so or we don't do nemesis during prepare, and we can apply needed logic on the nemesis (i.e. skip if there no enough columns)

roydahan:
Need to discuss it.

fruch:
> Need to discuss it.

I talked yesterday with @juliayakovlev 

and we think only removing the Index from the configuration would be enough change to avoid the issue.

There's no real reason to insist on it just a MV with specific limitations, all the other MVs can stay.

juliayakovlev:
> > Need to discuss it.
> 
> I talked yesterday with @juliayakovlev
> 
> and we think only removing the Index from the configuration would be enough change to avoid the issue.
> 
> There's no real reason to insist on it just a MV with specific limitations, all the other MVs can stay.

@fruch @roydahan please, review

fruch (approved):
LGTM

mykaul:
@fruch  / @juliayakovlev - who's backporting this to 2024? (and I can only assume earlier releases, such as 2023 as well?)

fruch:
> @fruch  / @juliayakovlev - who's backporting this to 2024? (and I can only assume earlier releases, such as 2023 as well?)

It's already got to 2024.1 (we are using labels to show it did)

As for 2023.1, after we'll be done with 2023.1.3, we get all of the fixes/new features in

Issue #6887, opened 2023-12-03
`artifacts-docker-test` failing in timeout when 'wait for: Node artifacts-docker-jenkins-db-node..'


Annamikhlin:
`artifacts-docker-test` starting to fail with error: 
```
06:55:26  < t:2023-12-02 04:55:25,708 f:docker_utils.py l:383  c:sdcm.utils.docker_utils p:INFO  > Build Docker image `scylla-sct:scylla-db-731418cc'
07:04:44  < t:2023-12-02 05:04:41,267 f:wait.py l:79   c:sdcm.wait  p:ERROR > Wait for: Node artifacts-docker-jenkins-db-node-731418cc-0 [172.17.0.2 | 172.17.0.2] (seed: False): Waiting for SSH to be up: timeout - 500 seconds - expired
```
https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-docker-test/761/

yaronkaikov:
@fruch We need this to backport to all releases. Since the removal of the ssh from the docker image was also backported

fruch:
@yaronkaikov please use the PR for backports labels, it's easier for me to track it (and the tooling we have for backporting, labels the PRs)

yaronkaikov:
> @yaronkaikov please use the PR for backports labels, it's easier for me to track it (and the tooling we have for backporting, labels the PRs)

Sure. Thanks

PR #6888, opened 2023-12-03
fix(docker): enable the ssh service back

fruch:
since scylladb/scylladb@7ce6962141dd4def9ce56837db691e2a8a983289 has removed the ssh service from it's code, we need to introduce it back on SCT end for docker backend to be workable

Fixes: #6887


yaronkaikov:
@fruch in 2022.2 it seems there is still an issue, https://jenkins.scylladb.com/job/enterprise-2022.2/job/artifacts/job/artifacts-docker-test/41/consoleFull Can you please have a look?

fruch:
> @fruch in 2022.2 it seems there is still an issue, https://jenkins.scylladb.com/job/enterprise-2022.2/job/artifacts/job/artifacts-docker-test/41/consoleFull Can you please have a look?

seems like in this change I've broke the version check, fixing it in:
https://github.com/scylladb/scylla-cluster-tests/pull/6958

mykaul:
@fruch - we use SSH since we do not have a 'docker exec' kind of way to get into the containers and run commands, I reckon?

fruch:
> @fruch - we use SSH since we do not have a 'docker exec' kind of way to get into the containers and run commands, I reckon?

it use ssh, since it's much more simple, and match how we work with VMs

for k8s we are not using ssh, but for that we have to have a full implementation or the Remoter, that includes running command, and copying files, we might have such for docker as well, but it doesn't worth the effort

fruch:
> > @fruch - we use SSH since we do not have a 'docker exec' kind of way to get into the containers and run commands, I reckon?
> 
> it use ssh, since it's much more simple, and match how we work with VMs
> 
> for k8s we are not using ssh, but for that we have to have a full implementation or the Remoter, that includes running command, and copying files, we might have such for docker as well, but it doesn't worth the effort

this check is relevant only to docker backend.
which we use for docker artifact test, and for local development only.

PR #6853, opened 2023-11-24
feature(k8s): add MultiDC support to GKE backend

vponomaryov:
Changes:
- Create `deploy_k8s_gke_cluster` function in the `gke.py` module that will be able to provision GKE clusters.
- Update the `gcloud` attr of the `GkeCluster` class to use `gce_zone` in the gcloud docker container name to be able to use multiple K8S clusters.
- Create `gce_datacenters` attr for the `params` object which will solve the same needs as `region_names` for AWS/EKS backends.
- Start respecting the `availability_zone` SCT config option. If not set then the `b` will be used as before. If set then apply.

## PR pre-checks (self review)

- [x] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [x] I didn't leave commented-out/debugging code
- [x] I added the relevant `backport` labels
- [ ] ~~New configuration option are added and documented (in `sdcm/sct_config.py`)~~
- [ ] ~~I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)~~
- [x] All new and existing unit tests passed (CI)
- [ ] ~~I have updated the Readme/doc folder accordingly (if needed)~~

PR #6894, opened 2023-12-04
fix(k8s-gke): set AZ explicitly as non-`a` for the `us-east1` region

vponomaryov:
After merge of the `MultiDC support for GKE backend` PR (https://github.com/scylladb/scylla-cluster-tests/pull/6853) the SCT started respecting the `availability_zone` config option.
It led to the situation that we started getting following error:

```
  ERROR: (gcloud.container.clusters.create) ResponseError: \
    code=400, \
    message=Specified location "us-east1-a" is not a valid zone in the cluster's region "us-east1".
```

It is caused by the fact that the default AZ was `a` and it is absent in the `us-east1` region.

So, fix this situation by explicitly setting the AZ for the GKE jobs.

### DIFF
--- a/defaults/k8s_gke_config.yaml
+++ b/defaults/k8s_gke_config.yaml
@@ -1,5 +1,6 @@
 instance_provision: 'on_demand'
 gce_datacenter: 'us-east1'
+availability_zone: 'c'
 gce_network: 'qa-vpc'
 gce_image_username: 'scylla-test'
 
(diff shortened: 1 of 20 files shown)


vponomaryov:
Example of the failure: https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/gke/job/longevity-scylla-operator-3h-gke/59/consoleFull

fruch (approved):
LGTM

Issue #6878, opened 2023-11-30
Improper describe command error during add_remove_dc nemesis


aleksbykov:
### Prerequisites

* [x] Are you rebased to master ?
* [x] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: MASTER
* **scylla**: 5.4.0~rc1-20231029.24efacf90d0b

### Logs
* test_id:  e9db935f-ffe4-459b-8569-1fb494924933
* job log: https://cloudius-jenkins-test.s3.amazonaws.com/e9db935f-ffe4-459b-8569-1fb494924933/20231109_051158/sct-e9db935f.log.tar.gz
* Argus: https://argus.scylladb.com/test/feba54c6-b2a9-4889-9991-6cc027f6fb9d/runs?additionalRuns[]=e9db935f-ffe4-459b-8569-1fb494924933

### Description

cqlsh command failed during getting description of keyspace:
```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5074, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4390, in disrupt_add_remove_dc
    self._switch_to_network_replication_strategy(self.cluster.get_test_keyspaces() + system_keyspaces)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4369, in _switch_to_network_replication_strategy
    replication_strategy = ReplicationStrategy.get(node, keyspace)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/replication_strategy_utils.py", line 24, in get
    create_ks_statement = node.run_cqlsh(f"describe {keyspace}").stdout.splitlines()[1]
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 2856, in run_cqlsh
    cqlsh_out = self.remoter.run(cmd, timeout=timeout + 120,  # we give 30 seconds to cqlsh timeout mechanism to work
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 613, in run
    result = _run()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 70, in inner
    return func(*args, **kwargs)
[...]

PR #6897, opened 2023-12-04
fix(run_cqlsh): make sure we escape double quotes

fruch:
since some keyspace or table names might be with quotes we should escape them, before passing it to the cqlsh command line

Ref: scylladb/scylla-cluster-tests#6790
Fixes: #6878

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2709,14 +2709,17 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
                       auth_params=auth_params, use_keyspace=use_keyspace, timeout=timeout,
                       connect_timeout=connect_timeout, ssl_params=ssl_params)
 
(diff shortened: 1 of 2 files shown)


soyacz:
why not just remove quotes from manager restore keyspaces names?

fruch:
> why not just remove quotes from manager restore keyspaces names?

Cause using quotes is part of the CQL spec

fruch:
> > why not just remove quotes from manager restore keyspaces names?
> 
> Cause using quotes is part of the CQL spec

See the note regarding the upper/lower case in:
https://docs.datastax.com/en/cql-oss/3.x/cql/cql_reference/cqlshDescribe.html

vponomaryov (dismissed):
LGTM

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

PR #6902, opened 2023-12-05
improvement(k8s): stabilize `test_deploy_quasi_multidc_db_cluster` test

vponomaryov:
It happens that the deletion of a namespace
in the 'test_deploy_quasi_multidc_db_cluster' K8S functional test may hang and finally return non-zero code as follows:

```
  Error executing command: "kubectl delete namespace t-podip-quasi-multidc"; Exit status: -9
  STDOUT: namespace "t-podip-quasi-multidc" deleted
```

Logs show that namespaces in such cases really get deleted.
So, skip this kind of errors because it is not an issue for us and only noise.

Also, the namespace deletion in this test, if not hangs, takes about `10-12` seconds.
So, reduce the timeout from the `300s/5m` to `30s` to minimize wasted time.


vponomaryov:
Example of the failure is here: https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/functional/job/functional-k8s-local-kind-aws/143/consoleFull

PR #6909, opened 2023-12-06
Fix cql_test sample test

xemul:
The thing is broken and doesn't succeed when run
Also, while at it, make it log its result in a nicer way

### DIFF
--- a/cql_test.py
+++ b/cql_test.py
@@ -33,7 +33,7 @@ class CQLExampleTest(ClusterTester):
             self.create_keyspace(keyspace_name='ks', replication_factor=1)
             # pylint: disable=no-member
             session.execute("""
-                CREATE TABLE test1 (
+                CREATE TABLE ks.test1 (
                     k int,
                     c1 int,
                     c2 int,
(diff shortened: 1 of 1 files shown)


vponomaryov:
> > [2023-12-06T12:03:35.775Z] ✖   type must be one of [ci, docs, feature, fix, improvement, perf, refactor, revert, style, test, unit-test] [type-enum]
> > [2023-12-06T12:03:35.775Z] ✖   scope may not be empty [scope-empty]
> 
> What is `type` and `scope` here?

```
type(scope): topic

details
```

In current case it may be following:
```
fix(cql_test): specify keyspace explicitly

The `cql_test` test fails to run CQL commands because keyspace is not specified.
So, set it explicitly in the CQL command.

Also, print it's output in a more nicer way.
```

Also, I don't see a real reason to have 2 commits here.
Both can be combined into one commit.

xemul:
upd:
- added `fix(cql_tst):` prefix
- squashed two changes into one

fruch (approved):
LGTM

first time I've noticed this test... no one really using it.

a nice minimal case get into trying out docker backend

PR #6912, opened 2023-12-06
feature(cluster_docker): support using specific docker network

fruch:
now all nodes are created on the default bridge network
but for some cases we have components using docker-compose
which by default creates it's own network, we can communicate
to our nodes from those docker-compose based setups.

so this new configuration option, would enable to tell upfront
that we need to create the nodes in a specific network.


sdcm/cluster.py
```diff
@@ -5124,12 +5124,13 @@ def node_setup(self, node, **kwargs):  # pylint: disable=unused-argument
...
 
         if self.test_config.REUSE_CLUSTER:
-            node.disable_daily_triggered_services()
+
             self.configure_scylla_monitoring(node)
             self.restart_scylla_monitoring(sct_metrics=True)
             set_grafana_url(f"http://{normalize_ipv6_url(node.external_address)}:{self.grafana_port}")
             return
 
+        node.disable_daily_triggered_services()
```
vponomaryov, line 5133: This change doesn't fit the `mocking` part. It is change of logic for other backends.
If it is intentional then we have bug now? Or why is it changed?

soyacz:
doesn't it break monitoring?

sdcm/utils/docker_utils.py
```diff
@@ -309,9 +309,11 @@ def wait_for_status(cls, instance: INodeWithContainerManager, name: str, status:
     @retrying(n=10, sleep_time=1, allowed_exceptions=(Retry, ))
     def get_ip_address(cls, instance: INodeWithContainerManager, name: str) -> str:
         ip_address = cls.get_container(instance, name).attrs["NetworkSettings"]["IPAddress"]
-        if not ip_address:
+        network_ip_address = cls.get_container(
+            instance, name).attrs["NetworkSettings"]["Networks"][instance.parent_cluster.params.get('docker_network')]["IPAddress"]
```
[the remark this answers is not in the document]
fruch: Yep in the following commit doing the same for integration tests, I did exactly that

fruch:
> doesn't it break monitoring?

You mean it might break monitoring.
Cause of the call that was moved ?

Or the docker networking setup?

sdcm/cluster.py
fruch: now on it's own commit

fruch:
> > > doesn't it break monitoring?
> > 
> > 
> > You mean it might break monitoring. Cause of the call that was moved ?
> > Or the docker networking setup?
> 
> docker networking setup - prometheus will try to get to node exporter/scylla metrics and I think we don't forward these ports

it might, if it doesn't setup with host networking

vponomaryov (approved):
LGTM

PR #6914, opened 2023-12-07
fix(quota): skip quota nemesis on not aws backends

Mark-Gurevich:
The nemesis currently fails on non aws backends
Added a skip check
SCT issue ref: https://github.com/scylladb/scylla-cluster-tests/issues/6915

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1704,6 +1704,11 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if self._is_it_on_kubernetes():
             raise UnsupportedNemesis("Skipping nemesis for kubernetes")
 
+        # Need to check on a live cluster with GCP and Azure how the quota configuration works / fails
+        # SCT Issue REF: https://github.com/scylladb/scylla-cluster-tests/issues/6915
+        if self.cluster.params.get('cluster_backend') != 'aws':
+            raise UnsupportedNemesis("The end of quota test only supports AWS at the moment")
+
         result = node.remoter.run('cat /proc/mounts')
         if '/var/lib/scylla' not in result.stdout:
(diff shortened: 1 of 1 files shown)


roydahan:
Why are they failing?
I'm not aware of anything there that should be a backend specific.

vponomaryov:
@Mark-Gurevich is there a bug that can be referenced here?
What is the problem with other backends?

fruch:
I've noticed a few jobs in 5.4, and in some testing in 2023.1

I've asked @Mark-Gurevich to handle it

@Mark-Gurevich as suggested here, opening an issue linking to the failures log
Could help giving people context to approve such a PR

And for follow-up on the actual issue

sdcm/nemesis.py
```diff
@@ -1704,6 +1704,10 @@ def disrupt_end_of_quota_nemesis(self, sleep_time=30):
         if self._is_it_on_kubernetes():
             raise UnsupportedNemesis("Skipping nemesis for kubernetes")
 
+        # Check on a live cluster with GCP and Azure how the quota configuration works / fails
+        if self.cluster.params.get('cluster_backend') != 'aws':
+            raise UnsupportedNemesis("The end of quota test only supports AWS at the moment")
```
vponomaryov, line 1709: Add bug ref here please.

vponomaryov (changes requested):
Need to add reference to the bug (https://github.com/scylladb/scylla-cluster-tests/issues/6915) in the commit message.
So, anyone looking at the git commit history will be able to find it easily.

Mark-Gurevich:
> > Need to add reference to the bug (#6915) in the commit message. So, anyone looking at the git commit history will be able to find it easily.
> 
> This is still actual - exactly the commit message is also useful to contain the ref to the bug.

Sorry I didn't get it about the commit message first time. My bad.

fruch (approved):
LGTM

Issue #6916, opened 2023-12-07
[K8S] data prefill step for the functional tests fails with the wrong CQL syntax


vponomaryov:
## Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

K8S functional tests fail on the setup step with the following error:

```
> E           Command: '/usr/bin/cqlsh --no-color   --request-timeout=120 --connect-timeout=60  -e "\\n                CREATE KEYSPACE keyspace1 WITH replication = {\'class\': \'NetworkTopologyStrategy\',\\n                \'replication_factor\': 3}\\n                " 10.16.59.131 9042'
> E                                                                                                    
> E           Exit code: 2                                                                             
> E                                                                                                    
> E           Stdout:                                                                                  
> E                                                                                                    
> E                                                                                                    
> E                                                                                                    
> E           Stderr:                                                                                  
> E                                                                                                    
> E           <stdin>:1:Invalid syntax at char 1                                                       
> E           <stdin>:1:  \n                CREATE KEYSPACE keyspace1 WITH replication = {'class': 'NetworkTopologyStrategy',\n                'replication_factor': 3}\n                ;   
> E           <stdin>:1:  ^                                                                            
>                                                                                                      
> /usr/local/lib/python3.10/site-packages/invoke/runners.py:509: UnexpectedExit 
```

## Impact

K8S functional tests cannot run due to the setup error.

## How frequently does it reproduce?

100%

## Installation details

Kernel Version: 5.15.0-1026-aws
Scylla version (or git commit hash): `2023.1.2-20231001.646df23cc4b3` with build-id `367fcf1672d44f5cbddc88f946cf272e2551b85a`

Operator Image: scylladb/scylla-operator:latest
Operator Helm Version: v1.12.0-alpha.0-144-g60f7824
Operator Helm Repository: https://storage.googleapis.com/scylla-operator-charts/latest
Cluster size: 3 pods

Scylla Nodes used in this run:
**No resources left at the end of the run**

Test: `vp-functional-k8s-local-kind-aws`
Test id: `da3d4370-068e-454c-a42d-09cfb5526209`
Test name: `scylla-staging/valerii/vp-functional-k8s-local-kind-aws`
Test config file(s):

- [functional.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/c67c444f3761ec05061423a5ad59bd1115637485/test-cases/scylla-operator/functional.yaml)

<details>
<summary>
[...]

PR #6918, opened 2023-12-07
fix(run_cqlsh): make it work having new line and space symbols in cmd

vponomaryov:
Recently we merged PR (https://github.com/scylladb/scylla-cluster-tests/pull/6897) which fixed the escape quotes approach.
But it added another problem that if we have new lines of additional spaces
in the beginning or ending of the command then it doesn't filter it out making CQLSH fails with the syntax errors.

So, fix it by removing the redundant symbols calling the `strip()` string method.
Also, cover this change in unit tests.

Fixes: #6916

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2710,7 +2710,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
                       connect_timeout=connect_timeout, ssl_params=ssl_params)
 
         # escape double quotes, that might be on keyspace/tables names
-        command = json.dumps(command)
+        command = json.dumps(command.strip())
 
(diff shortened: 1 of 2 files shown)


vponomaryov:
It should be backported to any branch where https://github.com/scylladb/scylla-cluster-tests/pull/6897 goes

vponomaryov:
Tested here: https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/316/consoleFull

fruch (approved):
LGTM

vponomaryov:
@fruch , @soyacz 
It went out to be only partial fix.
Here is second part of the fix for the functional tests: https://github.com/scylladb/scylla-cluster-tests/pull/6919

PR #6919, opened 2023-12-07
fix(sstables_utils): do not use multiline string defining cqlsh command

vponomaryov:
This is follow-up fix for the recently merged PRs (https://github.com/scylladb/scylla-cluster-tests/pull/6897) and (https://github.com/scylladb/scylla-cluster-tests/pull/6918).

We should not use multi-line strings defining cql commands, because any new lines in the middle of the string won't be removed and will cause following syntax errors:

```
  Command: '/usr/bin/cqlsh --no-color   --request-timeout=120 --connect-timeout=60 \
    -e "CREATE KEYSPACE keyspace1 WITH replication = \
    {\'class\': \'NetworkTopologyStrategy\',\\n             \'replication_factor\': 3}"  10.16.254.197 9042'
  Exit code: 2

  Stdout:
  Stderr:

  <stdin>:1:Invalid syntax at char 82
  <stdin>:1:  CREATE KEYSPACE keyspace1 WITH replication = \
    {'class': 'NetworkTopologyStrategy',\n                'replication_factor': 3};
  <stdin>:1:
```

### DIFF
--- a/sdcm/utils/sstable/load_utils.py
+++ b/sdcm/utils/sstable/load_utils.py
@@ -205,10 +205,9 @@ class SstableLoadUtils:
     @classmethod
     def create_keyspace(cls, node, keyspace_name: str = "keyspace1", strategy: str = 'NetworkTopologyStrategy',
                         replication_factor: int = 1):
-        node.run_cqlsh("""
-                CREATE KEYSPACE %s WITH replication = {'class': '%s',
-                'replication_factor': %s}
-                """ % (keyspace_name, strategy, replication_factor))
+        node.run_cqlsh(
+            "CREATE KEYSPACE %s WITH replication = {'class': '%s', 'replication_factor': %s}" % (
+                keyspace_name, strategy, replication_factor))
 
     @classmethod
     def create_table_for_load(cls, node, schema_file_and_path: str, session):


soyacz:
maybe better to sanitize the command with replace eol's?

vponomaryov:
> maybe better to sanitize the command with replace eol's?

If you men exactly the fixed one, then I don't see any need in sanitizing because multi-line is really redundant here.

vponomaryov:
Testing it here: https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-functional-k8s-local-kind-aws/318/consoleFull

soyacz:
> > maybe better to sanitize the command with replace eol's?
> 
> If you men exactly the fixed one, then I don't see any need in sanitizing because multi-line is really redundant here.

I mean to make cqlsh runner `command.strip().replace('/n', " ")` so anyone who create multiline command will not fail the test due to that issue

vponomaryov:
> > > maybe better to sanitize the command with replace eol's?
> > 
> > 
> > If you men exactly the fixed one, then I don't see any need in sanitizing because multi-line is really redundant here.
> 
> I mean to make cqlsh runner `command.strip().replace('/n', " ")` so anyone who create multiline command will not fail the test due to that issue

What if the `\n` going to be part of a command?
Also, what is the possible reason to use exactly multi-line? I don't see such a limitation.

fruch:
One thing we could add is failing straight aware with clear message is the command is multiple lines, cause it for sure gonna fail

Anyhow the idea of limiting it to a single line, is not. that clear, and we probably have more places it might fail in a similar way.

Maybe just escaping the double quotes would be a better option, and would support multil line as before

vponomaryov:
> > All the proposals do not block this change and can be done in addition anytime.
> 
> But we will need a followup, can you raise an issue / reopen, and make sure it's on our board. There are a lot of calls to this function, I don't know now which is broken.

- https://github.com/scylladb/scylla-cluster-tests/issues/6921

PR #6922, opened 2023-12-07
feature(loaders): add cql-stress C-S stress tool

muzarski:
Ref: https://github.com/scylladb/cql-stress

cql-stress is a stress tool implemented in Rust. It supports two CLI frontends: scylla-bench and cassandra-stress.

In this commit we introduce a cql-stress-cassandra-stress stress tool to SCT. Currently, the C-S frontend supports 4 basic commands:
- read
- write
- counter_read
- counter_write

## Changes
- docker
  - created docker image definition containing `cql-stress-cassandra-stress` executable
  - pushed docker image to registry
  - updated `defaults/test_default.yaml` accordingly
- loader
  - introduced `CqlStressCassandraStressThread` class deriving from `CassandraStressThread`
    - overridden `create_stress_cmd` and `_run_cs_stress` methods
  - introduced `CqlStressCassandraStressEvent` with critical default severity
  - introduced `CqlStressCassandraStressExporter` which collects the stats on the fly and record them to prometheus
  - added unit tests
- data_dir
  - updated `data_dir/scylla-dash-per-server-nemesis.master.json`. The generated diff looks strange, but the idea was to just copy-paste the existing `cassandra-stress` config for `write` and `read` workloads.

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -216,6 +216,7 @@ stress_image:
   kcl: 'scylladb/hydra-loaders:kcl-jdk8-20210526-ShardSyncStrategyType-PERIODIC'
   harry: 'scylladb/hydra-loaders:cassandra-harry-jdk11-20220816'
   latte: 'scylladb/hydra-loaders:latte-rust1_73-20231025'
+  cql-stress-cassandra-stress: 'scylladb/hydra-loaders:cql-stress-cassandra-stress-20240119'
 
 service_level_shares: [1000]
 
--- a/sdcm/cql_stress_cassandra_stress_thread.py
+++ b/sdcm/cql_stress_cassandra_stress_thread.py
@@ -1,0 +1,164 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2023 ScyllaDB
+
+import logging
+import re
+import os
+import time
+import contextlib
+from typing import Any
+from sdcm.loader import CqlStressCassandraStressExporter
+from sdcm.prometheus import nemesis_metrics_obj
+from sdcm.sct_events.loaders import CQL_STRESS_CS_ERROR_EVENTS_PATTERNS, CqlStressCassandraStressEvent
+from sdcm.stress_thread import CassandraStressThread
+from sdcm.utils.common import FileFollowerThread, SoftTimeoutContext
+from sdcm.utils.docker_remote import RemoteDocker
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class CqlStressCassandraStressEventsPublisher(FileFollowerThread):
+    def __init__(self, node: Any, log_filename: str, event_id: str = None):
+        super().__init__()
+
+        self.node = str(node)
+        self.log_filename = log_filename
+        self.event_id = event_id
+
+    def run(self) -> None:
+        while not self.stopped():
+            if not os.path.isfile(self.log_filename):
+                time.sleep(0.5)
+                continue
+
+            for line_number, line in enumerate(self.follow_file(self.log_filename)):
+                if self.stopped():
+                    break
+
+                for pattern, event in CQL_STRESS_CS_ERROR_EVENTS_PATTERNS:
+                    if self.event_id:
+                        event.event_id = self.event_id
+
+                    if pattern.search(line):
+                        event.add_info(node=self.node, line=line, line_number=line_number).publish()
+
+
+class CqlStressCassandraStressThread(CassandraStressThread):
+    DOCKER_IMAGE_PARAM_NAME = 'stress_image.cql-stress-cassandra-stress'
+
+    def __init__(self, loader_set, stress_cmd, timeout, stress_num=1, keyspace_num=1, keyspace_name='', compaction_strategy='',  # pylint: disable=too-many-arguments
+                 profile=None, node_list=None, round_robin=False, client_encrypt=False, stop_test_on_failure=True,
+                 params=None):
+        super().__init__(loader_set=loader_set, stress_cmd=stress_cmd, timeout=timeout,
+                         stress_num=stress_num, node_list=node_list,  # pylint: disable=too-many-arguments
+                         round_robin=round_robin, stop_test_on_failure=stop_test_on_failure, params=params,
+                         keyspace_num=keyspace_num, keyspace_name=keyspace_name, profile=profile,
+                         client_encrypt=client_encrypt, compaction_strategy=compaction_strategy)
+
+    def create_stress_cmd(self, cmd_runner, keyspace_idx, loader):  # pylint: disable=too-many-branches
+        stress_cmd = self.stress_cmd
+
+        stress_cmd = self.append_no_warmup_to_cmd(stress_cmd)
+        stress_cmd = self.adjust_cmd_keyspace_name(stress_cmd, keyspace_idx)
+        stress_cmd = self.adjust_cmd_compaction_strategy(stress_cmd)
+
+        if '-col' in stress_cmd:
+            # In cassandra-stress, '-col n=' parameter defines number of columns called (C0, C1, ..., Cn) respectively.
+            # It accepts a distribution as value, however, only 'FIXED' distribution is accepted for the cql mode.
+            # In cql-stress, we decided to accept a number instead.
+            # To support the c-s syntax in test-cases files, we replace occurrences
+            # of n=FIXED(k) to n=k.
+            stress_cmd = re.sub(r' (\'?)n=[\s]*fixed\(([0-9]+)\)(\'?)', r' \1n=\2\3',
+                                stress_cmd, flags=re.IGNORECASE)
+
+        if '-rate' in stress_cmd:
+            # In cassandra-stress either '-rate throttle=' or '-rate fixed=' parameter is accepted.
+            # In cql-stress we decided to change it so 'fixed' is a boolean flag
+            # specifying whether the tool should display coordination-omission fixed latencies.
+            stress_cmd = re.sub(r' fixed=[\s]*([0-9]+/s)',
+                                r' throttle=\1 fixed', stress_cmd)
+
+        if '-pop' in stress_cmd and "seq=" in stress_cmd:
+            # '-pop seq=x..y' syntax is not YET supported by cql-stress.
+            # TODO remove when seq=x..y syntax is supported.
+            LOGGER.warning(
+                "-pop seq=x..y syntax is not YET supported by cql-stress-cassandra-stress. Replacing seq=x..y with dist=SEQ(x..y).")
+            stress_cmd = re.sub(r' seq=[\s]*([\d]+\.\.[\d]+)',
+                                r" 'dist=SEQ(\1)'", stress_cmd)
+
+        stress_cmd = self.adjust_cmd_node_option(stress_cmd, loader, cmd_runner)
+        return stress_cmd
+
+    def _run_cs_stress(self, loader, loader_idx, cpu_idx, keyspace_idx):  # pylint: disable=too-many-locals,too-many-branches,too-many-statements
+        # TODO:
+        # - Add support for profile yaml once cql-stress supports 'user' command.
+        # - Adjust metrics collection once cql-stress displays the metrics grouped by operation (mixed workload).
+
+        cleanup_context = contextlib.nullcontext()
+        os.makedirs(loader.logdir, exist_ok=True)
+
+        stress_cmd_opt = self.stress_cmd.split(
+            "cql-stress-cassandra-stress", 1)[1].split(None, 1)[0]
+
+        log_id = self._build_log_file_id(loader_idx, cpu_idx, keyspace_idx)
+        log_file_name = os.path.join(
+            loader.logdir, f'cql-stress-cassandra-stress-{stress_cmd_opt}-{log_id}.log')
+
+        LOGGER.debug('cassandra-stress local log: %s', log_file_name)
+
+        cmd_runner_name = loader.ip_address
+
+        cpu_options = ""
+        if self.stress_num > 1:
+            cpu_options = f'--cpuset-cpus="{cpu_idx}"'
+
+        cmd_runner = cleanup_context = RemoteDocker(loader, self.docker_image_name,
+                                                    command_line="-c 'tail -f /dev/null'",
+                                                    extra_docker_opts=f'{cpu_options} '
+                                                    '--network=host '
+                                                    f'--label shell_marker={self.shell_marker}'
+                                                    f' --entrypoint /bin/bash')
+        stress_cmd = self.create_stress_cmd(cmd_runner, keyspace_idx, loader)
+        LOGGER.info('Stress command:\n%s', stress_cmd)
+
+        tag = f'TAG: loader_idx:{loader_idx}-cpu_idx:{cpu_idx}-keyspace_idx:{keyspace_idx}'
+
+        if self.stress_num > 1:
+            node_cmd = f'STRESS_TEST_MARKER={self.shell_marker}; taskset -c {cpu_idx} {stress_cmd}'
+        else:
+            node_cmd = f'STRESS_TEST_MARKER={self.shell_marker}; {stress_cmd}'
+        node_cmd = f'echo {tag}; {node_cmd}'
+
+        result = None
+        with cleanup_context, \
+                CqlStressCassandraStressExporter(instance_name=cmd_runner_name,
+                                                 metrics=nemesis_metrics_obj(),
+                                                 stress_operation=stress_cmd_opt,
+                                                 stress_log_filename=log_file_name,
+                                                 loader_idx=loader_idx, cpu_idx=cpu_idx), \
+                CqlStressCassandraStressEventsPublisher(node=loader, log_filename=log_file_name) as publisher, \
+                CqlStressCassandraStressEvent(node=loader, stress_cmd=self.stress_cmd,
+                                              log_file_name=log_file_name) as cs_stress_event:
+            publisher.event_id = cs_stress_event.event_id
+            try:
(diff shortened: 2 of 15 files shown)


defaults/test_default.yaml
```diff
@@ -216,6 +216,7 @@ stress_image:
   kcl: 'scylladb/hydra-loaders:kcl-jdk8-20210526-ShardSyncStrategyType-PERIODIC'
   harry: 'scylladb/hydra-loaders:cassandra-harry-jdk11-20220816'
   latte: 'scylladb/hydra-loaders:latte-rust1_73-20231025'
+  cql-stress-cassandra-stress: 'scylladb/hydra-loaders:cql-stress-cassandra-stress-20231130'
```
fruch, line 219: It tool itself has some release cadence and versions we can use them and not dates

Also it would be nice if the docker images would be built on the tool repo

Gemini now does it.

sdcm/tester.py
```diff
@@ -1866,19 +1867,20 @@ def run_stress_cassandra_thread(
...
-                                          keyspace_num=keyspace_num,
-                                          compaction_strategy=compaction_strategy,
-                                          profile=profile,
-                                          node_list=self.db_cluster.nodes,
-                                          round_robin=round_robin,
-                                          client_encrypt=self.params.get('client_encrypt'),
-                                          keyspace_name=keyspace_name,
-                                          stop_test_on_failure=stop_test_on_failure,
-                                          params=params or self.params).run()
+        cls = CqlStressCassandraStressThread if "cql-stress" in stress_cmd else CassandraStressThread
```
fruch, line 1870: I would rather have this explicit in its own function

This is a bit confusing to follow

muzarski:
> Do you run this one with a longevity test for a couple of hours, with nemesis ?

No, unfortunately I didn't have an opportunity to do so. I've only tested the solution with two provision test cases:
- https://github.com/scylladb/scylla-cluster-tests#run-test-locally-with-aws-backend
- https://github.com/scylladb/scylla-cluster-tests#run-test-locally-with-docker-backend

I obviously adjusted the `stress_cmd` config parameter to make use of `cql-stress-cassandra-stress`.

I am new to SCT, and I am still not sure about a lot of available config parameters.

Should I create a seperate test-case as in https://github.com/scylladb/scylla-cluster-tests/pull/6735/commits/cec116422c0f3fb373ca2204c8fd31d41f17aa7f? Should I create a jenkins pipeline for that? Which nemesis should be used? What SCT backend should I use?

I'd appreciate any tips regarding preparation (and running) the test case that would allow us to compare both (cql-stress vs c-s) of the solutions.

fruch:
Since it's a drop-in, I would suggest taking a current case that can support it, and duplicate it to use this tool

And then run both on Jenkins under scylla-staging, so results are visible to others

@soyacz can you help him with showing how to reuse existing tests with override configuration ?

soyacz:
> 
> @soyacz can you help him with showing how to reuse existing tests with override configuration ?

Sure, @muzarski ping me on the Slack we can have call so I can answer also some other questions

sdcm/tester.py
vponomaryov: Agree.
Moreover, `cls` is reserved name for the current class reference.

sdcm/loader.py
```diff
@@ -202,6 +202,57 @@ def split_line(self, line: str) -> list:
...
+
+        if "Results:" in line:
+            # Stress test finished.
+            self.do_skip = True
+
+        return self.do_skip
+
+    @staticmethod
+    def split_line(line: str) -> list:
+        return [element.strip() for element in line.split(',')]
```
vponomaryov, line 253: At first, it is unused.
At second, it is not specific to this class.
So, if you want such utility function/method then need to place in some of the `utils` modules.

sdcm/cql_stress_cassandra_stress_thread.py
```diff
@@ -0,0 +1,158 @@
...
+from sdcm.sct_events.loaders import CqlStressCassandraStressEvent
+from sdcm.stress_thread import CassandraStressThread
+from sdcm.utils.common import SoftTimeoutContext
+from sdcm.utils.docker_remote import RemoteDocker
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class CqlStressCassandraStressThread(CassandraStressThread):
```
vponomaryov, line 29: Why not `CqlStressThread` ?

sdcm/cql_stress_cassandra_stress_thread.py
```diff
@@ -0,0 +1,158 @@
...
+
+        if '-pop' in stress_cmd:
+            # '-pop seq=x..y' syntax is not YET supported by cql-stress.
+            # TODO remove when seq=x..y syntax is supported.
+            stress_cmd = re.sub(r' seq=[\s]*([\d]+\.\.[\d]+)',
+                                r" 'dist=SEQ(\1)'", stress_cmd)
+
+        if self.node_list and '-node' not in stress_cmd:
+            stress_cmd += " -node "
+            if self.loader_set.test_config.MULTI_REGION:
```
vponomaryov, line 84: It should consider the SCT config option called `region_aware_loader`.

sdcm/cql_stress_cassandra_stress_thread.py
```diff
@@ -0,0 +1,158 @@
...
+                datacenter_name_per_region = self.loader_set.get_datacenter_name_per_region(
+                    db_nodes=self.node_list)
+                if loader_dc := datacenter_name_per_region.get(loader.region):
+                    stress_cmd += f"datacenter={loader_dc} "
+                else:
+                    LOGGER.error("Not found datacenter for loader region '%s'. Datacenter per loader dict: %s",
+                                 loader.region, datacenter_name_per_region)
+
+            node_ip_list = [n.cql_address for n in self.node_list]
+            stress_cmd += ",".join(node_ip_list)
```
vponomaryov, line 96: It could be simplified to the following:

```
stress_cmd += ",".join((n.cql_address for n in self.node_list))
```

sdcm/cql_stress_cassandra_stress_thread.py
```diff
@@ -0,0 +1,158 @@
...
+            # In cql-stress we decided to change it so 'fixed' is a boolean flag
+            # specifying whether the tool should display coordination-omission fixed latencies.
+            stress_cmd = re.sub(r' fixed=[\s]*([0-9]+/s)',
+                                r' throttle=\1 fixed', stress_cmd)
+
+        if '-pop' in stress_cmd:
+            # '-pop seq=x..y' syntax is not YET supported by cql-stress.
+            # TODO remove when seq=x..y syntax is supported.
+            stress_cmd = re.sub(r' seq=[\s]*([\d]+\.\.[\d]+)',
+                                r" 'dist=SEQ(\1)'", stress_cmd)
```
vponomaryov, line 80: This part worth adding a `warning` log message.

sdcm/cql_stress_cassandra_stress_thread.py
```diff
@@ -0,0 +1,158 @@
...
+                         stress_num=stress_num, node_list=node_list,  # pylint: disable=too-many-arguments
+                         round_robin=round_robin, stop_test_on_failure=stop_test_on_failure, params=params,
+                         keyspace_num=keyspace_num, keyspace_name=keyspace_name, profile=profile,
+                         client_encrypt=client_encrypt, compaction_strategy=compaction_strategy)
+
+    def create_stress_cmd(self, cmd_runner, keyspace_idx, loader):  # pylint: disable=too-many-branches
+        stress_cmd = self.stress_cmd
+
+        if "no-warmup" not in stress_cmd:
+            # add no-warmup to stress_cmd if it's not there. See issue #5767
```
vponomaryov, line 45: Which project's issue?

sdcm/cql_stress_cassandra_stress_thread.py
```diff
@@ -0,0 +1,158 @@
...
+        cleanup_context = contextlib.nullcontext()
+        os.makedirs(loader.logdir, exist_ok=True)
+
+        stress_cmd_opt = self.stress_cmd.split(
+            "cql-stress-cassandra-stress", 1)[1].split(None, 1)[0]
+
+        log_id = self._build_log_file_id(loader_idx, cpu_idx, keyspace_idx)
+        log_file_name = \
+            os.path.join(
+                loader.logdir, f'cql-stress-cassandra-stress-{stress_cmd_opt}-{log_id}.log')
```
vponomaryov, line 113: Why not like following:

```
log_file_name = os.path.join(
    loader.logdir, f'cql-stress-cassandra-stress-{stress_cmd_opt}-{log_id}.log')
```

?

sdcm/cql_stress_cassandra_stress_thread.py
fruch: this looks like  a code that should be shared, is looks this was copied from cassandra-stress implementation.
and shouldn't (also there an open PR to extend this logic for rack aware configuration)

sdcm/cql_stress_cassandra_stress_thread.py
fruch: also big portion of `CassandraStressThread` implementation is copied here, 
so maybe there no reason to inherit from it at all ?

if we do inherit, we should move the relevant part which needs to be different to their own functions/properties that can be overridden

vponomaryov:
@muzarski I see that you uploaded new changes but haven't posted any inline comment.
So, either you haven't written any or haven't published the `draft` comments. Which one is true?

muzarski:
> @muzarski I see that you uploaded new changes but haven't posted any inline comment. So, either you haven't written any or haven't published the `draft` comments. Which one is true?

I haven't pushed any commits regarding the review comments yet. I didn't have time to go through the comments yet.

The commits I've pushed are just draft commits. I'll re-request the review once the code is ready for it.

muzarski:
Sorry for the delay, I had to adjust some things to make the longevity test run correctly. Here is the jenkins pipeline: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/muzarki/job/longevity-100gb-4h-test-cql-stress/.

Note that the last run failed, but as I spoke with @soyacz, the issue is not related with cql-stress.

I'll start fixing the review comments now.

sdcm/cql_stress_cassandra_stress_thread.py
muzarski: > Why not `CqlStressThread` ?

As `cql-stress` supports two frontends (`cassandra-stress` and `scylla-bench`), I preferred to be more specific.

sdcm/loader.py
muzarski: It's used by the parent class `StressExporter`. The default implementation is empty.

muzarski:
v2:
- rebased on top of master
- fixed review comments, including:
  - created separate `run_cql_stress_cassamdra_thread` function
  - extracted common logic of `CassandraStressThread::create_stress_cmd` to functions so they can be reused by `CqlStressCassandraStressThread`

muzarski:
v3:
- initiate `result` variable before `with` block in `_run_cs_stress`

sdcm/loader.py
fruch: this is the StressExporter api, that's o.k. 

if the output of this tool is identical to c-s, it might be o.k. to inherit from `CassandraStressExporter`, and just replace `create_metrix_gauge`

anyhow it's a nitpick from my POV.

fruch:
@muzarski 

I'm looking at the dashboard and results from:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/muzarki/job/longevity-100gb-4h-test-cql-stress/11

image

1) this dashboard looks duplicated
2) seems like metrics dropped at ~13:00
3) seem to corrlate to logs in 
```
[2m2024-01-05T13:09:31.237763Z[0m [31mERROR[0m [2mcql_stress_cassandra_stress::operation::write[0m[2m:[0m write error [3merror[0m[2m=[0mInvalid message: Frame error: early eof [3mpartition_key[0m[2m=[0mBlob([76, 79, 55, 77, 53, 57, 56, 76, 77, 48])
```

can we drop the ansi colors ? 

I'm guess `early eof` is cause of a node going down, we'll need to be able to tell a bit better which line is critical to stop the test (for example validation error, or failure after x retries)
see `CassandraStressEventsPublisher` for how it's used there

need to check why those lines are breaking the reporting code

without those stats working, it's a bit hard to compare it to regular c-s results/runs

muzarski:
> @muzarski
> 
>     1. this dashboard looks duplicated
I'll deduplicate it.

>     2. seems like metrics dropped at ~13:00
> 
>     3. seem to corrlate to logs in
> 
> 
> ```
> �[2m2024-01-05T13:09:31.237763Z�[0m �[31mERROR�[0m �[2mcql_stress_cassandra_stress::operation::write�[0m�[2m:�[0m write error �[3merror�[0m�[2m=�[0mInvalid message: Frame error: early eof �[3mpartition_key�[0m�[2m=�[0mBlob([76, 79, 55, 77, 53, 57, 56, 76, 77, 48])
> ```

It might be because `StressExporter` api failed to parse the `early eof` line. I'll adjust the `skip_line` logic so we skip the lines emitted by [tracing](https://docs.rs/tracing/latest/tracing/) crate.

> can we drop the ansi colors ?

Opened a PR: https://github.com/scylladb/cql-stress/pull/66.

> I'm guess `early eof` is cause of a node going down, we'll need to be able to tell a bit better which line is critical to stop the test (for example validation error, or failure after x retries) see `CassandraStressEventsPublisher` for how it's used there

Yes, I believe the `early eof` error is caused by `disrupt_stop_wait_start_scylla_server` nemesis. Some operations need to be retried to be successful in this case.

I believe that the only `CRITICAL`  event which should stop the test is the read validation error. Are there any other critical events we should consider?

fruch:
> > @muzarski
> > 
> >     1. this dashboard looks duplicated
> I'll deduplicate it.
> 
> >     2. seems like metrics dropped at ~13:00
> > 
> >     3. seem to corrlate to logs in
> > 
> > 
> > ```
> > �[2m2024-01-05T13:09:31.237763Z�[0m �[31mERROR�[0m �[2mcql_stress_cassandra_stress::operation::write�[0m�[2m:�[0m write error �[3merror�[0m�[2m=�[0mInvalid message: Frame error: early eof �[3mpartition_key�[0m�[2m=�[0mBlob([76, 79, 55, 77, 53, 57, 56, 76, 77, 48])
> > ```
> 
> It might be because `StressExporter` api failed to parse the `early eof` line. I'll adjust the `skip_line` logic so we skip the lines emitted by [tracing](https://docs.rs/tracing/latest/tracing/) crate.
> 
> > can we drop the ansi colors ?
> 
> Opened a PR: https://github.com/scylladb/cql-stress/pull/66.
> 
> > I'm guess `early eof` is cause of a node going down, we'll need to be able to tell a bit better which line is critical to stop the test (for example validation error, or failure after x retries) see `CassandraStressEventsPublisher` for how it's used there
> 
> Yes, I believe the `early eof` error is caused by `disrupt_stop_wait_start_scylla_server` nemesis. Some operations need to be retried to be successful in this case.
> 
> I believe that the only `CRITICAL`  event which should stop the test is the read validation error. Are there any other critical events we should consider?

I don't have nothing in mind right now, but once we'll have more mileage with this tool, we will probably have some more things we'll want to raise out of the log.

muzarski:
v3:
- deduplicated `cql-stress-cassandra-stress ops` dashboard
- added `CqlStressCassandraStressEventsPublisher` which checks for `read validation error` events. See exemplary log from `cql-stress` when row validation fails: `2024-01-18T15:57:52.108866Z ERROR cql_stress_cassandra_stress::operation::read: read validation error error=Query result doesn't contain any rows. partition_key=Blob([75, 75, 79, 51, 76, 48, 57, 48, 49, 49])`
- Adjusted `CqlStressCassandraStressExporter` so it ignores lines emitted by `tracing` crate. It means - to ignore lines containing any of the `TRACE`, `DEBUG`, `INFO`, `WARN`, `ERROR` keywords.

muzarski:
Scheduled a SCT run with new changes: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/muzarki/job/longevity-100gb-4h-test-cql-stress/12/.

roydahan:
Any progress with this one?

muzarski:
@fruch I addressed review comments and scheduled a run (which was a success)

sdcm/cql_stress_cassandra_stress_thread.py
fruch: this comment seem to be copied from somewhere else...
I don't know what the issue, but we shouldn't be using warmup, it's doesn't have any meaning for scylla, its something todo with cassandra java implementation ontop of the JVM, that need time to "stabilize"

muzarski:
> key observation:
> 
>     1. cql-stress is doing more then x2 requests !
> 
>     2. latency of p99 is below ~2ms vs. ~10ms in c-s

In addition, notice that `cql-stress` workload uses `rate threads=30`, when c-s workload uses `rate threads=40`. I didn't want to overload the cluster so decided to use a smaller number of threads here (actually, this is not the number of threads, but number of async tokio tasks in the pool).

> seems like we still have some duplication in the p99 and p95 graphs

Looking at the photos, I see that we have dashboards for both c-s and cql-stress (C-S frontend). Note that some nemesis could run cassandra-stress in the background and so c-s metrics were collected as well.

fruch:
> > key observation:
> > 
> >     1. cql-stress is doing more then x2 requests !
> > 
> >     2. latency of p99 is below ~2ms vs. ~10ms in c-s
> 
> In addition, notice that `cql-stress` workload uses `rate threads=30`, when c-s workload uses `rate threads=40`. I didn't want to overload the cluster so decided to use a smaller number of threads here (actually, this is not the number of threads, but number of async tokio tasks in the pool).

That's something we'll need todo per use case and it's requirements, for this one hitting 50% CPU look o.k.
> 
> > seems like we still have some duplication in the p99 and p95 graphs
> 
> Looking at the photos, I see that we have dashboards for both c-s and cql-stress (C-S frontend). Note that some nemesis could run cassandra-stress in the background and so c-s metrics were collected as well.

Yeah, I've noticed it, but still we have cql-stress read panels duplicated here

Anyhow that's good enough to start using in some of other cases

> 
>

muzarski:
@fruch resolved the conflict

fruch (approved):
LGTM

roydahan:
Great work @muzarski!
I was happy to see that you took care also the grafana dashboards and was able to run and demonstrate everything.

roydahan:
@fruch I see that the new pipeline was added under "rust" folder.
This is probably not handled in our tool that creates the jobs in Jenkins automatically.
Is your PR for "pipelines" re-org going to take care of that or we need to manually fix it now?

PR #6923, opened 2023-12-07
fix(run_cqlsh): stop specifying host and port on call to cqlsh

fruch:
Seems like by doing so SCT code was masking an issue the scylla docker was wrongly configured the setup needed for cqlsh to work with specifying host and port

Ref: https://github.com/scylladb/scylladb/issues/16329


fruch:
and in the meanwhile give a test with all backends, to make sure we don't have surprises in other places.

fruch:
And on VMs it doesn't work, we need to specify the address

fruch:
passed the docker with latest as expected

roydahan (approved):

PR #6931, opened 2023-12-12
fix(disrupt_add_remove_dc): put back c-s thread into context

fruch:
as part of #6873, one verification thread, that was supposed to be after node decommission, we done after keyspace was dropped putting back the decommission part into the sequence, and that validation thread into the context as well.

Ref: #6873

#### Tests
- [x] - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-AddRemoveDcNemesis-aws-test/10/

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4453,9 +4453,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self.tester.create_keyspace("keyspace_new_dc", replication_factor={
                                     datacenters[0]: min(3, len(self.cluster.nodes))})
         node_added = False
-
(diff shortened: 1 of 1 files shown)


juliayakovlev (dismissed):

soyacz (commented):
in the original test, `temporary_replication_strategy_setter` was recovering replication strategy to original one (removing `new_dc_name`) and then doing verifications.

I think we cannot mix finalizer with replication setter or we need to intentionally drop `new_dc_name` from `strategy` and apply it.

fruch:
@soyacz @juliayakovlev 

Now it's more similar to how it was setup before, and passing the job

soyacz (approved):
LGTM

PR #6228, opened 2023-06-09
fix(Rollback): disable recover_system_tables

temichus:
set recover_system_tables option to False
due to https://github.com/scylladb/scylladb/issues/11907

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I didn't leave commented-out/debugging code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #6932, opened 2023-12-12
fix(upgrade_tests): enable `recover_system_tables` as needed

fruch:
when rolling back from enterpise back to OSS, we are gonna hit this error:

```
init - Startup failed: sstables::malformed_sstable_exception
(Failed to read partition from SSTable /var/lib/scylla/data/system_schema/.../me-182-big-Data.db
due to Column in_memory missing in current schema)
```

in order to avoid it, we need to recover the schema from snapshot

in PR #6228 it was disabled complately, but was missing this case of rollback into OSS, which still needs it.

Ref: scylladb/scylladb#11907

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -36,6 +36,7 @@ from sdcm.utils.user_profile import get_profile_content
 from sdcm.utils.version_utils import (
     get_node_supported_sstable_versions,
     ComparableScyllaVersion,
+    is_enterprise,
 )
 from sdcm.sct_events.system import InfoEvent
 from sdcm.sct_events.database import (
@@ -395,6 +396,9 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
             recover_conf(node)
             node.remoter.run('sudo systemctl daemon-reload')
 
+        result = node.remoter.run('scylla --version')
+        new_ver = result.stdout.strip()
+
         node.remoter.run('sudo cp /etc/scylla/scylla.yaml-backup /etc/scylla/scylla.yaml')
         result = node.remoter.run('sudo find /var/lib/scylla/data/system')
         snapshot_name = re.findall(r"system/peers-[a-z0-9]+/snapshots/(\d+)\n", result.stdout)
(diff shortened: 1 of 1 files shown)


temichus:
@bhalevy, @tgrabiec 
Could you please also take a look?
Looks like it is missing point.

according to @bhalevy :
```
I think that we can safely remove the "Restore system tables" section from 4.6 and on and correspondingly, from 2022.1 and on in the enterprise.
In any case, QA should test that first to make sure we didn't miss anything.
```

so probably it should be a Scylla issue, not a test case change. But, I might be wrong

fruch:
> @bhalevy, @tgrabiec Could you please also take a look? Looks like it is missing point.
> 
> according to @bhalevy :
> 
> ```
> I think that we can safely remove the "Restore system tables" section from 4.6 and on and correspondingly, from 2022.1 and on in the enterprise.
> In any case, QA should test that first to make sure we didn't miss anything.
> ```
> 
> so probably it should be a Scylla issue, not a test case change. But, I might be wrong

those issues suggest we should be doing that:
- https://github.com/scylladb/scylla-enterprise/issues/1950

the case you and @bhalevy was looking at was just OSS -> OSS, and Enterprise - > Enterprise, 
and not OSS -> Enterprise

bhalevy:
@fruch https://github.com/scylladb/scylla-enterprise/issues/1950 is about enterprise -> OSS, isn't it?
I think that OSS -> enterprise upgrade of the schema should work.

fruch:
> @fruch https://github.com/scylladb/scylla-enterprise/issues/1950 is about enterprise -> OSS, isn't it?
> I think that OSS -> enterprise upgrade of the schema should work.

We are talking on rollback during upgrade from OSS -> enterprise 

Which is a switch from enterprise to OSS

Which still breaks without this recovery of the system tables...

fruch:
> @fruch https://github.com/scylladb/scylla-enterprise/issues/1950 is about enterprise -> OSS, isn't it?
> I think that OSS -> enterprise upgrade of the schema should work.

We are talking on rollback during upgrade from OSS -> enterprise 

Which is a switch from enterprise to OSS

Which still breaks without this recovery of the system tables....

bhalevy:
> > @fruch [scylladb/scylla-enterprise#1950](https://github.com/scylladb/scylla-enterprise/issues/1950) is about enterprise -> OSS, isn't it?
> > I think that OSS -> enterprise upgrade of the schema should work.
> 
> We are talking on rollback during upgrade from OSS -> enterprise
> 
> Which is a switch from enterprise to OSS
> 
> Which still breaks without this recovery of the system tables...

For rolling back an upgrade from OSS to enterprise, restoring the schema tables should restore the original SSTables that were backed up from OSS, so restoring is OSS -> OSS.

fruch:
> > > @fruch [scylladb/scylla-enterprise#1950](https://github.com/scylladb/scylla-enterprise/issues/1950) is about enterprise -> OSS, isn't it?
> > > I think that OSS -> enterprise upgrade of the schema should work.
> > 
> > We are talking on rollback during upgrade from OSS -> enterprise
> > 
> > Which is a switch from enterprise to OSS
> > 
> > Which still breaks without this recovery of the system tables...
> 
> For rolling back an upgrade from OSS to enterprise, restoring the schema tables should restore the original SSTables that were backed up from OSS, so restoring is OSS -> OSS.

And this PR is enabling exactly that, that was disabled by @temichus, and was about to be removed from docs.

bhalevy (approved):

PR #6937, opened 2023-12-13
fix(arm): detect the ARM instance types on AWS more correctly

vponomaryov:
Following problems exist in the current code of the ARM detection on AWS:
- Some of the existing prefixes are ambiguous such as `m6` which match `m6g` (ARM) and `m6i` (Intel) processors.
- Some of the instance types are just absent which can be used in K8S. Example: `t4g.large`

So, fix the `ARM_ARCH_PREFIXES` var in the `sdcm/utils/aws_utils.py` module be less ambiguous and contain more possible ARM prefixes.

### DIFF
--- a/sdcm/utils/aws_utils.py
+++ b/sdcm/utils/aws_utils.py
@@ -27,7 +27,21 @@ from sdcm.test_config import TestConfig
 from sdcm.keystore import KeyStore
 
 LOGGER = logging.getLogger(__name__)
-ARM_ARCH_PREFIXES = ('im4', 'is4', 'a1.', 'inf', 'm6', 'c6g', 'r6', 'm7', 'c7g', 'r7')
+# NOTE: non storage-optimized instance types may be used for K8S auxiliary/loader nodes
+ARM_ARCH_PREFIXES = (
+    # General purpose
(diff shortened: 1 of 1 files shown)


soyacz (approved):
LGTM
but maybe we could just verify `g` in arch prefix? So we are future proof also.

fruch (approved):
LGTM

vponomaryov:
> but maybe we could just verify `g` in arch prefix? So we are future proof also.

It is incorrect assumption.
There following non-arm instance types with `g` letter: `g5`, `g4dn`, `g4ad` and `g3`. May be other.

fruch:
> LGTM but maybe we could just verify `g` in arch prefix? So we are future proof also.

it doesn't sound like something that is gonna be guaranteed

the best way would be to check it via the API, for that specific instance type
or get the whole list like in:
```
aws ec2 describe-instance-types --region eu-west-1  --filter Name=processor-info.supported-architecture,Values=arm64 --query 'InstanceTypes[*].InstanceType'
```

vponomaryov:
Closing in favor this:
- https://github.com/scylladb/scylla-cluster-tests/pull/6942

PR #6939, opened 2023-12-13
fix(k8s): use latest `scylla-manager` version `3.2.5`

vponomaryov:
Do it because it is latest, recommended and supports ARM instance types.


vponomaryov:
Tested here: https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-functional-eks-arm/5/consoleFull

defaults/k8s_eks_config.yaml
```diff
@@ -17,7 +17,9 @@ eks_nodegroup_role_arn: 'arn:aws:iam::797456418907:role/helm-test-worker-nodes-N
 eks_vpc_cni_version: 'v1.15.3-eksbuild.1'
 
 scylla_version: '5.2.11'
-scylla_mgmt_agent_version: '3.1.0'
+scylla_mgmt_agent_version: '3.2.5'
```
fruch, line 20: why we have those default set for both EKS/GKE/functional ? why can't those be part of test default ? 
are they used in VMs ?

fruch (changes requested):
I think we can consilade the manager related configuration into one default.
there not really case for having them in 4 different places

defaults/k8s_eks_config.yaml
vponomaryov: Done

fruch (approved):
Lgtm

PR #6944, opened 2023-12-14
fix(audit-nemesis): fix number of rows error

soyacz:
When c-s fails with timeout or failure it retries the query/update. This causes more audit rows than we expect and fail the nemesis. And because CL=ALL it is more likely to fail due cluster load.

Switch c-s CL to ONE - audit should log such updates/queries without any issue and nemesis should pass.

refs: https://github.com/scylladb/scylla-enterprise/issues/3682

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4818,14 +4818,14 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             errors = []
             audit_start = datetime.datetime.now() - datetime.timedelta(seconds=5)
             InfoEvent(message='Writing/Reading data from audited keyspace').publish()
-            write_cmd = f"cassandra-stress write no-warmup cl=ALL n=1000 -schema" \
+            write_cmd = f"cassandra-stress write no-warmup cl=ONE n=1000 -schema" \
                         f" 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)" \
                         f" keyspace={audit_keyspace}' -mode cql3 native -rate 'threads=1 throttle=1000/s'" \
                         f" -pop seq=1..1000 -col 'n=FIXED(1) size=FIXED(128)' -log interval=5"
             write_thread = self.tester.run_stress_thread(
                 stress_cmd=write_cmd, round_robin=True, stop_test_on_failure=False)
             self.tester.verify_stress_thread(cs_thread_pool=write_thread)
-            read_cmd = f"cassandra-stress read no-warmup cl=ALL n=1000 " \
+            read_cmd = f"cassandra-stress read no-warmup cl=ONE n=1000 " \
                        f" -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)" \
                        f" keyspace={audit_keyspace}' -mode cql3 native -rate 'threads=1 throttle=1000/s'" \
                        f" -pop seq=1..1000 -col 'n=FIXED(1) size=FIXED(128)' -log interval=5"


soyacz:
test drive here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/audit-nemesis/22/

ptrsmrn (approved):

fruch:
@soyacz 

I think we should be handling the case of retries, it might happen also with ONE,
if those retries are eventually passing, we probably need not count the failures (assuming it would be fixed)

anyone, I think for starter we can assume there would be more items then we except, since retries.

also if some follow is overload to this degree, we need to to understand why.
(and if it's parallel nemesis, we need to consider if this nemesis can run in parallel to others)

soyacz:
> @soyacz
> 
> I think we should be handling the case of retries, it might happen also with ONE, if those retries are eventually passing, we probably need not count the failures (assuming it would be fixed)
> 
> anyone, I think for starter we can assume there would be more items then we except, since retries.
> 
> also if some follow is overload to this degree, we need to to understand why. (and if it's parallel nemesis, we need to consider if this nemesis can run in parallel to others)

In case cl ONE fails, there should be no audit log. This nemesis also works on separate keyspace so should not interfere with other nemesis - unless there are topology changes - so we could have 2 nodes down (as toggling audit requires rolling restart currently).

There's another possible track we spoke with @ptrsmrn - statements that fail should be marked in audit log with `error=true` (but I'm not sure when fail happens - e.g. is timeout enough, and in case of temporary replication failure).
IMHO, this should fix this case, let's wait for test results.

soyacz:
run failed due existing bug (unrelated): https://github.com/scylladb/scylla-enterprise/issues/3661

retrying on GCE: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/audit-nemesis/23/

fruch:
> > @soyacz any update on this one ? are we still sure it's the best approach here ?
> 
> I think so:
> 
> 1. when failed to write with CL=ONE - there should be no audit log, and c-s should repeat insert
> 2. if inserted with CL=ONE - there should be one audit log - this is what we assert

I don't understand how quorum is changing the logic of audit, it should be logged once on the coordinator
if it fails and if it pass, and cause of retries we might have more audit but marked as failed.

cl=ONE might lower the chances of retries needed, but  its still there.

if we are counting failures in, that something need fixing (in SCT or core)
also as a temporary measure (assuming its a core issue), we can assert on bigger or equal then the number of requests sent.

soyacz:
> > > @soyacz any update on this one ? are we still sure it's the best approach here ?
> > 
> > 
> > I think so:
> > 
> > 1. when failed to write with CL=ONE - there should be no audit log, and c-s should repeat insert
> > 2. if inserted with CL=ONE - there should be one audit log - this is what we assert
> 
> I don't understand how quorum is changing the logic of audit, it should be logged once on the coordinator if it fails and if it pass, and cause of retries we might have more audit but marked as failed.
> 
> cl=ONE might lower the chances of retries needed, but its still there.
> 
> if we are counting failures in, that something need fixing (in SCT or core) also as a temporary measure (assuming its a core issue), we can assert on bigger or equal then the number of requests sent.

IIUC when coordinator writes, and row is inserted but replicas didn't respond, audit should happen. IMHO, from user perspective, such insert is valid and I'm not sure if marking it with error should happen. 
We would need clear answer when 'error' is used in auditing? (when trying to do something without permissions? - that would be valid use of this field IMHO)

ptrsmrn:
> > > > @soyacz any update on this one ? are we still sure it's the best approach here ?
> > > 
> > > 
> > > I think so:
> > > 
> > > 1. when failed to write with CL=ONE - there should be no audit log, and c-s should repeat insert
> > > 2. if inserted with CL=ONE - there should be one audit log - this is what we assert
> > 
> > 
> > I don't understand how quorum is changing the logic of audit, it should be logged once on the coordinator if it fails and if it pass, and cause of retries we might have more audit but marked as failed.
> > cl=ONE might lower the chances of retries needed, but its still there.
> > if we are counting failures in, that something need fixing (in SCT or core) also as a temporary measure (assuming its a core issue), we can assert on bigger or equal then the number of requests sent.
> 
> IIUC when coordinator writes, and row is inserted but replicas didn't respond, audit should happen. IMHO, from user perspective, such insert is valid and I'm not sure if marking it with error should happen. We would need clear answer when 'error' is used in auditing? (when trying to do something without permissions? - that would be valid use of this field IMHO)

@KrzaQ wrote a test for failed auditing, and, to our surprise, when a query is failed, sometimes audit logs are not logged at all and sometimes they are logged with `error = true` - https://github.com/scylladb/scylla-dtest/pull/3783/files#diff-3961593aca64edf2d0076176ec34bcd99ca90fb02b9dfe324674155ac2bf1eb6R784. We were unable to identify variable(s) responsible for either not saving audit logs or saving it with `error = true`.

soyacz:
> > > > > @soyacz any update on this one ? are we still sure it's the best approach here ?
> > > > 
> > > > 
> > > > I think so:
> > > > 
> > > > 1. when failed to write with CL=ONE - there should be no audit log, and c-s should repeat insert
> > > > 2. if inserted with CL=ONE - there should be one audit log - this is what we assert
> > > 
> > > 
> > > I don't understand how quorum is changing the logic of audit, it should be logged once on the coordinator if it fails and if it pass, and cause of retries we might have more audit but marked as failed.
> > > cl=ONE might lower the chances of retries needed, but its still there.
> > > if we are counting failures in, that something need fixing (in SCT or core) also as a temporary measure (assuming its a core issue), we can assert on bigger or equal then the number of requests sent.
> > 
> > 
> > IIUC when coordinator writes, and row is inserted but replicas didn't respond, audit should happen. IMHO, from user perspective, such insert is valid and I'm not sure if marking it with error should happen. We would need clear answer when 'error' is used in auditing? (when trying to do something without permissions? - that would be valid use of this field IMHO)
> 
> @KrzaQ wrote a test for failed auditing, and, to our surprise, when a query is failed, sometimes audit logs are not logged at all and sometimes they are logged with `error = true` - https://github.com/scylladb/scylla-dtest/pull/3783/files#diff-3961593aca64edf2d0076176ec34bcd99ca90fb02b9dfe324674155ac2bf1eb6R784. We were unable to identify variable(s) responsible for either not saving audit logs or saving it with `error = true`.

But is this correct behavior? I was thinking that in case of writing with let's say CL=ALL when some but not all replicas respond, data is saved but may not be consistent across all the nodes temporarily. Audit serves to verify who and what changed in database, in that case I consider it to be a valid data change and should be audited although query from client perspective failed with consistency error.
We need to have clear statement what error field serves.

fruch:
@soyacz we still think this is the needed fix from the test end ?

soyacz:
> @soyacz we still think this is the needed fix from the test end ?

yes, I think it is valid solution.
Still need to uncover the purpose of 'error' field in the audit logs - but this is different case

fruch (approved):
LGTM

fruch:
@soyacz backport labels ?

soyacz:
> @soyacz backport labels ?

I think all active branches.

btw. It's strange that after 2 reviews I still cannot merge due missing 2 approvals.

fruch:
> > @soyacz backport labels ?
> 
> I think all active branches.
> 
> btw. It's strange that after 2 reviews I still cannot merge due missing 2 approvals.

It's configured to need 2 reviews from specific group

fruch:
@soyacz please tag all those active branches

soyacz:
> @soyacz please tag all those active branches

I think I did - is there any more missing?

fruch:
> > @soyacz please tag all those active branches
> 
> I think I did - is there any more missing?

branch-5.2 isn't relevant, this nemesis code didn't ever reached it (and it's OSS, so no need to port it there)

PR #6946, opened 2023-12-14
feature(kafka-localstack): introducing docker-compose base kafka setup

fruch:
Since we want to be able to run scylla kafka connectors with scylla clusters create by SCT, we are introducing here the first of kafka backend that would be used for local development (with SCT docker backend)

* include a way to configure the connector as needed (also multi ones)
* get it install from hub or by url

### Testing
- [x] https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-kafka-cdc-docker-test/8/
- [x] https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-kafka-cdc-aws-test/7/


sdcm/kafka/kafaka_cluster.py
```diff
@@ -0,0 +1,158 @@
+# This program is free software; you can redistribute it and/or modify
```
vponomaryov, line 1: Nice file name ^_^

sdcm/sct_config.py
```diff
@@ -1516,6 +1516,13 @@ class SCTConfiguration(dict):
 
         dict(name="validate_large_collections", env="SCT_VALIDATE_LARGE_COLLECTIONS", type=boolean,
              help="Enable validation for large cells in system table and logs"),
+
+        dict(name="kafka_backend", env="SCT_KAFKA_BACKEND", type=str,
+             help="Enable validation for large cells in system table and logs",
+             choices=("none", "localstack", "vm", "msk")),
+
+        dict(name="kafka_connectors", env="SCT_KAFKA_CONNECTORS", type=str_or_list_or_eval,
+             help="configuration for setup up kafka connectors"),
```
vponomaryov, line 1525: Would be good to know where to get the config structure

sdcm/kafka/kafaka_cluster.py
```diff
@@ -0,0 +1,158 @@
...
+        )
+        self._journal_thread: KafkaLogger | None = None
+        self.init_repository()
+
+    def init_repository(self):
+        # TODO: make the url configurable
+        # TODO: create a fork/branch and remove default connector installers
+        # TODO: switch to specific branch/sha
+        # TODO: get the version after install, and send out to Argus
+        clone_repo(
```
vponomaryov, line 54: What about pinning version?

sdcm/kafka/kafaka_cluster.py
```diff
@@ -0,0 +1,158 @@
...
+from sdcm.utils.git import clone_repo
+from sdcm.utils.common import get_sct_root_path
+from sdcm.utils.remote_logger import CommandClusterLoggerBase
+
+# TODO: write/think more about the consumers
+
+logger = logging.getLogger(__name__)
+
+
+class KafkaLogger(CommandClusterLoggerBase):
```
vponomaryov, line 32: Other such loggers are placed in the `sdcm.utils.remote_logger`.
So, why this one is here?

sdcm/sct_config.py
```diff
@@ -1516,6 +1516,13 @@ class SCTConfiguration(dict):
 
         dict(name="validate_large_collections", env="SCT_VALIDATE_LARGE_COLLECTIONS", type=boolean,
```
vponomaryov, line 1517: Probably also should be added check for backend. No?

sdcm/kafka/kafaka_cluster.py
fruch: no particular reason, it was easier to develop it like that.
Also it's a bit different that the other ones, since it's not being initiate in the same fashion.

I can move, probably when we'll have more kind of this setup, like on VM, or as SaaS this would have more variants

sdcm/sct_config.py
fruch: what do you mean ? validation for `kafka_backend` and needed parameters ?

sdcm/kafka/kafaka_cluster.py
vponomaryov: No, I mean to pull `specific` version instead of the `latest/current`

sdcm/sct_config.py
vponomaryov: 2 parts:
- on which cluster backend it can run
- kafka backend configuration correctness/validation

sdcm/kafka/kafaka_cluster.py
fruch: I've read it printing.... 

yeah we should pin it, and I want to fork it first, to remove a few things to make it start faster (it install 2 connector at the beginning that we don't need)

sdcm/sct_config.py
fruch: I think we'll point to one or two examples
also I introduce a pydantic dataclass hat would be able to verify this part of configuration

sdcm/kafka/kafaka_cluster.py
fruch: yep, I keep mistyping it, renamed this file.

sdcm/sct_config.py
fruch: pydantic dataclass was introduced, it should be a bit clearer now

sdcm/kafka/kafaka_cluster.py
fruch: I've now pinned it, to a fork.

later we'll move it to scylladb org, and start doing versions

sdcm/kafka/kafka_cluster.py
```diff
@@ -0,0 +1,184 @@
...
+                    [node.cql_address for node in db_cluster.nodes]
+                )
+                connector_data["config"]["scylladb.contact.points"] = scylla_addresses
+            case "com.scylladb.cdc.debezium.connector.ScyllaConnector":
+                scylla_addresses = ",".join(
+                    [f"{node.cql_address}:{node.CQL_PORT}" for node in db_cluster.nodes]
+                )
+                connector_data["config"][
+                    "scylla.cluster.ip.addresses"
+                ] = scylla_addresses
```
vponomaryov, line 143: Looks as unreasonable line splitting.
Maybe like following:
```
scylla_addresses = [f"{node.cql_address}:{node.CQL_PORT}" for node in db_cluster.nodes]
connector_data["config"]["scylla.cluster.ip.addresses"] = ",".join(scylla_addresses)
```

?

sdcm/kafka/kafka_cluster.py
```diff
@@ -0,0 +1,184 @@
...
+from sdcm.wait import wait_for
+from sdcm.remote import LOCALRUNNER
+from sdcm.utils.git import clone_repo
+from sdcm.utils.common import get_sct_root_path
+from sdcm.utils.remote_logger import CommandClusterLoggerBase
+from sdcm.kafka.kafka_config import SctKafkaConfiguration
+
+# TODO: write/think more about the consumers
+
+logger = logging.getLogger(__name__)
```
vponomaryov, line 29: Use uppercase for the global objects.
I already tried to find `logger` in method scope.

sdcm/tester.py
```diff
@@ -1554,6 +1562,12 @@ def get_cluster_k8s_local_kind_cluster(self):
         else:
             self.monitors = NoMonitorSet()
 
+    def get_cluster_kafka(self):
+        kafka_backend = self.params.get('kafka_backend')
+        if kafka_backend == 'localstack':
```
vponomaryov, line 1567: New config option supports more than 1 value for the kafka backend.
So, need to have some logic for handling others. For example - `raise NotImplementedError()`.

sdcm/kafka/kafka_cluster.py
```diff
@@ -0,0 +1,184 @@
...
+        self.docker_compose_path = (
+            Path(get_sct_root_path()) / "kafka-stack-docker-compose"
+        )
+        self._journal_thread: KafkaLogger | None = None
+        self.init_repository()
+
+    def init_repository(self):
+        # TODO: make the url configurable
+        # TODO: get the version after install, and send out to Argus
+        repo_url = "git@github.com:fruch/kafka-stack-docker-compose.git"
```
soyacz, line 52: I'd propose to copy it to SCT so we can adapt & review it here

test-cases/kafka/longevity-kafka-cdc-docker.yaml
```diff
@@ -0,0 +1,35 @@
...
+]
+
+stress_cmd: ["cassandra-stress write cl=ONE duration=1m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=1) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=10 -pop seq=1..10000000 -log interval=5",
+             ]
+
+n_loaders: 1
+n_monitor_nodes: 1
+n_db_nodes: 1
+
+instance_type_runner: c5.2xlarge
```
soyacz, line 14: let's use c7

test-cases/kafka/longevity-kafka-cdc-docker.yaml
soyacz, line 1: need to remember to update keyspace, number of nodes and stress_cmd before merging

soyacz:
are there kafka metrics worth to add to monitoring? If yes, can be done in followup task.

fruch:
> are there kafka metrics worth to add to monitoring? If yes, can be done in followup task.

this one is a local setup of kafka, I don't think monitoring is needed, as least not yet. (we have monitoring data of the sct-runner)

soyacz:
> > are there kafka metrics worth to add to monitoring? If yes, can be done in followup task.
> 
> this one is a local setup of kafka, I don't think monitoring is needed, as least not yet. (we have monitoring data of the sct-runner)

they look nice: https://grafana.com/docs/grafana-cloud/monitor-infrastructure/integrations/integration-reference/integration-kafka/
It could help us when there are issues with kafka-connector

sdcm/kafka/kafka_cluster.py
```diff
@@ -0,0 +1,184 @@
...
+# TODO: write/think more about the consumers
+
+logger = logging.getLogger(__name__)
+
+
+class KafkaLogger(CommandClusterLoggerBase):  # pylint: disable=too-few-public-methods
+    # pylint: disable=invalid-overridden-method
+    @cached_property
+    def _logger_cmd(self) -> str:
+        return f"{self._cluster.compose_context} logs --no-color --tail=1000 >>{self._target_log_file}"
```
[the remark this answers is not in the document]
soyacz: when running logs with --tail=1000 in a loop won't it get last 1000 lines each time (which can be the same)?

in DockerScyllaLogger we use `-f` - but I don't know if it is supported by compose.

test-cases/kafka/longevity-kafka-cdc-docker.yaml
soyacz: what is the purpose?

fruch: it's a preview, that we can discuss with the team owns the connectors, after that, they would define the cases they need and gonna work on.

test-cases/kafka/longevity-kafka-cdc-docker.yaml
fruch: updated it to c7

sdcm/kafka/kafka_cluster.py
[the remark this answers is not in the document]
fruch: we are running this command once, i.e. we are not restarting the docker-compose setup, the docker logger need to handle restarts of nodes, and probably isn't doing that very good, anyhow.

fruch:
> > > are there kafka metrics worth to add to monitoring? If yes, can be done in followup task.
> > 
> > 
> > this one is a local setup of kafka, I don't think monitoring is needed, as least not yet. (we have monitoring data of the sct-runner)
> 
> they look nice: https://grafana.com/docs/grafana-cloud/monitor-infrastructure/integrations/integration-reference/integration-kafka/ It could help us when there are issues with kafka-connector

JMX never looks nice... 

it's too early for this, once we'll have VMs and a full cluster, we might consider installation of those.

now I care more on the functional side of things, and how this setup integrates with a longevity test.
and the real missing part is reading/writing to kafka for the actual test/verification

sdcm/kafka/kafka_cluster.py
fruch: fixed

sdcm/kafka/kafka_cluster.py
fruch: fixed

sdcm/kafka/kafka_cluster.py
```diff
@@ -0,0 +1,184 @@
...
+        # TODO: handle user/password
+        # TODO: handle client encryption SSL
+
+        connector_data = connector_config.dict(by_alias=True, exclude_none=True)
+        match connector_config.config.connector_class:
+            case "io.connect.scylladb.ScyllaDbSinkConnector":
+                scylla_addresses = ",".join(
+                    [node.cql_address for node in db_cluster.nodes]
+                )
+                connector_data["config"]["scylladb.contact.points"] = scylla_addresses
```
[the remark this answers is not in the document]
fruch: Not sure exactly which structure you are referring to, of the connector_data ? 

it's documented inside each connector repo, and there are links to it on `kafka_config.py` which show the structure.

what else you are looking for ?

sdcm/sct_config.py
fruch: for the first there's no current limitation on backend, in theory it can run with any backend

as for kafka_backend, only one is implemented, validation for it is pointless at this point.

for the 2nd there is validation

sdcm/kafka/kafaka_cluster.py
fruch: move to remote_logger, even that it's not exactly remote :)

jenkins-pipelines/oss/kafka_connectors/longevity-kafka-cdc-aws.jenkinsfile
```diff
@@ -0,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    backend: 'docker',
```
vponomaryov, line 7: aws?

jenkins-pipelines/oss/kafka_connectors/longevity-kafka-cdc-aws.jenkinsfile
```diff
@@ -0,0 +1,13 @@
...
+    region: 'eu-west-1',
+    availability_zone: 'c',
+
+    test_name: 'longevity_test.LongevityTest.test_custom_time',
+    test_config: 'test-cases/kafka/longevity-kafka-cdc-docker.yaml',
```
vponomaryov, line 12: If the config file is suitable also to AWS, why then the file has the "docker" word in it?

sdcm/kafka/kafka_consumer.py
```diff
@@ -0,0 +1,104 @@
...
+from sdcm.sct_config import SCTConfiguration
+from sdcm.kafka.kafka_config import SctKafkaConfiguration
+from sdcm.utils.common import generate_random_string
+from sdcm.wait import wait_for
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class KafakaCDCReaderThread(Thread):  # pylint: disable=too-many-instance-attributes
```
vponomaryov, line 31: Typo in class name

unit_tests/test_kafka.py
```diff
@@ -0,0 +1,82 @@
...
+    files="unit_tests/test_data/kafka_connectors/kafka-connect-scylladb.yaml"
+)
+def test_02_kafka_scylla_sink_connector(docker_scylla, kafka_cluster, params):
+    """
+    setup kafka with kafka-connect-scylladb with docker based scylla node
+    - from confluent-hub
+    - from GitHub release url
+    """
+
+    # docker_scylla.run_cqlsh(
```
vponomaryov, line 105: Do you still need this debug code?

sdcm/kafka/kafka_config.py
```diff
@@ -0,0 +1,37 @@
...
+    scylla_table_names: Optional[str] = Field(alias="scylla.table.names")
+    scylla_user: Optional[str] = Field(alias="scylla.user")
+    scylla_password: Optional[str] = Field(alias="scylla.password")
+
+    # kafka-connect-scylladb
+    # see https://github.com/scylladb/kafka-connect-scylladb/blob/master/documentation/CONFIG.md
+    scylladb_contact_points: Optional[str] = Field(alias="scylladb.contact.points")
+    scylladb_keyspace: Optional[str] = Field(alias="scylladb.keyspace")
+    scylladb_user: Optional[str] = Field(alias="scylladb.user")
+    scylladb_password: Optional[str] = Field(alias="scylladb.password")
```
vponomaryov, line 26: The diff between `scylladb_password` and `scylla_password` may be understood only looking at the source code and not by the attr name.
Better to rename it to reduce the ambiguity.

fruch: Where do you see scylla_password here ?

Otherwise I don't understand what the question

jenkins-pipelines/oss/kafka_connectors/longevity-kafka-cdc-aws.jenkinsfile
fruch: I'll rename the file

It was before I thought of reusing it to multiple backends

fruch:
So the longevity code we have basically works

But it hangs cause we don't have code to stop the Kafka reading thread, might use the idea of teardown validator to validate and stop the reading thread

sdcm/kafka/kafka_config.py
vponomaryov: I see it on line 19. The same about the `scylla_user` and `scylladb_user`

fruch: This configuration is an overlap of two different connectors

The discrepancy you see here is between them.

You can see in the link each one points to a different connector

jenkins-pipelines/oss/kafka_connectors/longevity-kafka-cdc-aws.jenkinsfile
fruch: fixed, it's now aws

sdcm/kafka/kafka_consumer.py
fruch: Fixed the name of this class

unit_tests/test_kafka.py
fruch: this test isn't complete, 
it's just install this connector, but doesn't really utilize it.

this connector write data from kafka to scylla, so we'll need to write a producer code (similar to the KafkaConsumer one)
that would be able to write data into kafka, and somehow read it from one of our stress tool, or via the python driver.
would still to put some more thought on how todo it.

also more research is need to figure out how to operator this connector 
following: https://github.com/scylladb/kafka-connect-scylladb/blob/master/documentation/QUICKSTART.md

fruch:
> > So the longevity code we have basically works
> > 
> > But it hangs cause we don't have code to stop the Kafka reading thread, might use the idea of teardown validator to validate and stop the reading thread
> 
> I don't understand why we cannot add this verification to teardown itself? Why teardown validator is required?

It was an idea, validators seemed like a natural place for it 

I'm now trying a different approach of adding this logic to the reader thread itself.

sdcm/kafka/kafka_cluster.py
```diff
@@ -0,0 +1,171 @@
...
+    def __init__(self, remoter=LOCALRUNNER):
+        super().__init__(cluster_prefix="kafka", add_nodes=False)
+        self.remoter = remoter
+        self.docker_compose_path = (
+            Path(get_sct_root_path()) / "kafka-stack-docker-compose"
+        )
+        self._journal_thread: DockerComposeLogger | None = None
+        self.init_repository()
+
+    def init_repository(self):
```
soyacz, line 41: why we need to clone whole repository? don't we need only `full-stack.yml` and `conduktor.yml` file?
(btw. how conduktor is used in our context? maybe we could skip it?)

sdcm/tester.py
```diff
@@ -811,10 +816,10 @@ def is_encrypt_keys_needed(self):
...
     def prepare_kms_host(self) -> None:
         if (self.params.is_enterprise and ComparableScyllaVersion(self.params.scylla_version) >= '2023.1.3'
-            and self.params.get('cluster_backend') == 'aws'
-            and not self.params.get('scylla_encryption_options')
-            and self.params.get("db_type") != "mixed_scylla"  # oracle probably doesn't support KMS
-            ):
+                and self.params.get('cluster_backend') == 'aws'
+                and not self.params.get('scylla_encryption_options')
+                and self.params.get("db_type") != "mixed_scylla"  # oracle probably doesn't support KMS
+                ):
```
soyacz, line 822: same unrelated change I did once... why pycharm does that?

test-cases/kafka/longevity-kafka-cdc.yaml
```diff
@@ -0,0 +1,41 @@
...
+use_mgmt: false
+
+monitor_swap_size: 0
+
+docker_network: 'kafka-stack-docker-compose_default'
+
+kafka_backend: 'localstack'
+
+kafka_connectors:
+  - version: 'hub:scylladb/scylla-cdc-source-connector:1.1.2'
```
soyacz, line 31: `version` is confusing when trying follow the code - used to install connector: how `version` can be installed - why not to split it to `source` and `version`?

sdcm/kafka/kafka_cluster.py
```diff
@@ -0,0 +1,171 @@
...
+from sdcm.utils.remote_logger import DockerComposeLogger
+from sdcm.kafka.kafka_config import SctKafkaConfiguration
+
+# TODO: write/think more about the consumers
+
+LOGGER = logging.getLogger(__name__)
+
+
+class LocalKafkaCluster(cluster.BaseCluster):
+    def __init__(self, remoter=LOCALRUNNER):
```
soyacz, line 32: will sct_runner survive high load on kafka?

sdcm/kafka/kafka_cluster.py
fruch: The idea is that we can followup on updates

We can later optimize it if we find it an issue (i.e. remove the unneeded files from our fork)

sdcm/tester.py
fruch: I can revert it

test-cases/kafka/longevity-kafka-cdc.yaml
fruch: This field is used for multiple things yes,

similar to scylla_version in dtest

splitting it to multiple fields that are inter depends, it makes the config more confusing and more fields to expose in pipelines.

sdcm/kafka/kafka_cluster.py
fruch: It might, we can scale the runner as needed.

The idea is to have a setup that can work completely locally with docker backed for development

The next stage is building a cluster of kafka instead of the docker compose setup, and then some Kafka SaaS.

So for this initial step we can about building functionally, not yet about scale. scaling would be tested on VMs or on SaaS.

sdcm/kafka/kafka_cluster.py
soyacz: it's simple repo, 2 files needed, why we can't follow up manually? This will be similar experience. Just need some docs for that (anyway needed)

sdcm/kafka/kafka_cluster.py
soyacz: worth adding a note that it's on sct runner and its size should be increased.
Generally small docstring would be great.

sdcm/kafka/kafka_consumer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def stop(self):
+        self.termination_event.set()
+        self.consumer.close()
+
+    def kill(self):
+        self.termination_event.set()
+        self.consumer.close()
+
+    def verify_results(self) -> (list[dict | None], list[str | None]):
```
soyacz, line 105: not used, not implemented

fruch:
Two Jobs introduced are passing now

One small pre-commit issue, and it's good to go

sdcm/kafka/kafka_cluster.py
fruch: the repo is small enough, not going to change it.

sdcm/kafka/kafka_consumer.py
```diff
@@ -0,0 +1,107 @@
...
+        return topics
+
+    def wait_for_topic(self, topic, timeout):
+        def check_topic_exists():
+            topics = self.get_topics()
+            return topic in topics
+
+        wait_for(check_topic_exists, text=f"waiting for topic={topic}", timeout=timeout)
+
+    def run(self):
```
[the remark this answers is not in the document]
fruch: I don't see the point of repeat the same comment multiple times

sdcm/kafka/kafka_consumer.py
fruch: it is called, since we use this thread as a stress command.

sdcm/tester.py
fruch: reveted

sdcm/kafka/kafka_consumer.py
```diff
@@ -0,0 +1,107 @@
...
+            if len(self.keys) >= self.read_number_of_key:
+                self.stop()
+
+    def stop(self):
+        self.termination_event.set()
+        self.consumer.close()
+
+    def kill(self):
+        self.termination_event.set()
+        self.consumer.close()
```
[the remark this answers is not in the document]
fruch: changed

test-cases/kafka/longevity-kafka-cdc.yaml
fruch: changed it

unit_tests/test_kafka.py
fruch: so yes it's still needed for whom would complete the work for this connecter

test-cases/kafka/longevity-kafka-cdc.yaml
```diff
@@ -0,0 +1,41 @@
+test_duration: 60
+pre_create_keyspace: [
+  "CREATE KEYSPACE IF NOT EXISTS keyspace1 WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', 'replication_factor': 1 };",
+  "CREATE TABLE IF NOT EXISTS keyspace1.standard1 (key blob PRIMARY KEY, \"C0\" blob, \"C1\" blob, \"C2\" blob, \"C3\" blob, \"C4\" blob) WITH  cdc = {'enabled': true, 'preimage': false, 'postimage': true, 'ttl': 600}"
+]
+
+stress_cmd: ["cassandra-stress write cl=ONE n=1000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=1) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=10 -pop seq=1..10000000 -log interval=5",
+             "python_thread thread=KafakaCDCReaderThread ; read_number_of_key=1000"]
```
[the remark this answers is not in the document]
fruch: fixed

sdcm/kafka/kafka_consumer.py
```diff
@@ -0,0 +1,107 @@
...
+
+    def run(self):
+        while not self.termination_event.is_set():
+            records = self.consumer.poll(timeout_ms=1000)
+            for _, consumer_records in records.items():
+                for msg in consumer_records:
+                    data = json.loads(msg.value).get('payload', {}).get('after', {})
+                    key = base64.b64decode(data.get('key')).decode()
+                    self.keys.add(key)
+                    LOGGER.debug(key)
```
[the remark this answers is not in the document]
fruch: removed

fruch:
@Bouncheck 

I would recommend you try it again, to get familiar with it.

PR #6950, opened 2023-12-14
sla_tests: use SL REST API instead of binary protocol on/off workaround

Jadw1:
Current workaround with restarting binary protocol to update connections' scheduling group couldn't be done on all nodes because of running nemeses, which resulted in problems like https://github.com/scylladb/scylla-enterprise/issues/2572

This patch changed the method to use recently created REST API endpoints https://github.com/scylladb/scylla-enterprise/pull/3242 (`/service_levels/switch_tenants` and `/service_levels/count_connections`) to trigger scheduling groups update and to validate them.

### **Note** 
**This PR to run correctly requires https://github.com/scylladb/scylla-enterprise/pull/3242 to be in scylla and none of the 2023.x 2024.x have that PR.**

### DIFF
--- a/sdcm/sla/sla_tests.py
+++ b/sdcm/sla/sla_tests.py
@@ -106,10 +106,7 @@ class Steps(SlaUtils):
 
     # pylint: disable=too-many-arguments
     def attach_sl_and_validate_scheduler_runtime(self, tester, new_service_level, role_for_attach,
-                                                 read_roles, prometheus_stats, sleep=600,
-                                                 # restart_scylla parameter is temporary - wWorkaround for issue
-                                                 # https://github.com/scylladb/scylla-enterprise/issues/2572
-                                                 restart_scylla=False):
+                                                 read_roles, prometheus_stats, sleep=600):
         @retrying(n=15, sleep_time=1, message="Wait for service level has been attached to the role",
                   allowed_exceptions=(Exception, ValueError,))
         def validate_role_service_level_attributes_against_db():
@@ -123,12 +120,31 @@ class Steps(SlaUtils):
                 role_for_attach.attach_service_level(new_service_level)
                 role_for_attach.validate_role_service_level_attributes_against_db()
 
-                # Workaround for issue https://github.com/scylladb/scylla-enterprise/issues/2572
-                if restart_scylla:
-                    nodes_for_restart = [node for node in tester.db_cluster.nodes if node.jmx_up() and node.db_up()
-                                         and not node.running_nemesis]
-                    tester.db_cluster.restart_binary_protocol(nodes=nodes_for_restart)
-                # End workaround
+                # Fix for issue https://github.com/scylladb/scylla-enterprise/issues/2572
+                for node in tester.db_cluster.nodes:
+                    if node.db_up():
+                        node.remoter.run(
+                            "curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json'"
+                            " http://127.0.0.1:10000/service_levels/switch_tenants"
+                        )
+
+                # Wait for tenant switch
+                # The above call to '/service_levels/switch_tenants' only marks
+                # all of the CQL connections to switch tenant (it's scheduling group).
+                # The actual switch on a connection will happen when the connection finshes processing current request.
+                time.sleep(5*60)
+
+                # Print connections map on each node to validate connections' scheduling groups
+                for node in tester.db_cluster.nodes:
+                    if node.db_up():
+                        result = node.remoter.run(
+                            "curl -X GET --header 'Content-Type: application/json' --header 'Accept: application/json'"
+                            " http://127.0.0.1:10000/service_levels/count_connections"
+                        )
+
+                        connections_map = result.stdout.strip()
+                        LOGGER.debug("[NODE {}] {}".format(node.private_ip_address, connections_map))
+                # End fix
 
                 start_time = time.time() + 60
                 time.sleep(sleep)
@@ -255,8 +271,7 @@ class SlaTests(Steps):
                                                                   role_for_attach=role_low,
                                                                   read_roles=read_roles,
                                                                   prometheus_stats=prometheus_stats,
-                                                                  sleep=600,
-                                                                  restart_scylla=True))
+                                                                  sleep=600))
 
             finally:
                 self.verify_stress_threads(tester=tester, stress_queue=stress_queue)
@@ -438,8 +453,7 @@ class SlaTests(Steps):
                                                                   role_for_attach=role_high,
                                                                   read_roles=read_roles,
                                                                   prometheus_stats=prometheus_stats,
-                                                                  sleep=600,
-                                                                  restart_scylla=True))
+                                                                  sleep=600))
 
             finally:
                 self.verify_stress_threads(tester=tester, stress_queue=stress_queue)
@@ -515,8 +529,7 @@ class SlaTests(Steps):
                                                                   role_for_attach=role_low,
                                                                   read_roles=read_roles,
                                                                   prometheus_stats=prometheus_stats,
-                                                                  sleep=600,
-                                                                  restart_scylla=True))
+                                                                  sleep=600))
 
             finally:
                 self.verify_stress_threads(tester=tester, stress_queue=stress_queue)


sdcm/sla/sla_tests.py
```diff
@@ -123,12 +123,25 @@ def validate_role_service_level_attributes_against_db():
                 role_for_attach.attach_service_level(new_service_level)
                 role_for_attach.validate_role_service_level_attributes_against_db()
 
-                # Workaround for issue https://github.com/scylladb/scylla-enterprise/issues/2572
-                if restart_scylla:
```
temichus, line 127: restart_scylla  after this PR will be unused. so it needs to be deleted from "attach_sl_and_validate_scheduler_runtime" method and it usage

sdcm/sla/sla_tests.py
```diff
@@ -123,12 +123,25 @@ def validate_role_service_level_attributes_against_db():
...
-                    tester.db_cluster.restart_binary_protocol(nodes=nodes_for_restart)
-                # End workaround
+                for node in tester.db_cluster.nodes:
+                    if node.db_up():
+                        node.remoter.run(
+                            "curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/json'"
+                            " http://127.0.0.1:10000/service_levels/switch_tenants"
+                        )
+
+                time.sleep(5*60)
```
temichus, line 133: what we are waiting for here? Could you please leave a comment?  Can we make some smart wait? for example by periodically checking service_levels/count_connections or it is not the case here?

fruch:
@Jadw1 
see the linting errors in:
https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-6950/2/execution/node/23/log

fruch:
@Jadw1 

can you share a run demonstrating this works ?, also can you clarify in which release its gonna be available ?

sdcm/sla/sla_tests.py
```diff
@@ -123,12 +120,29 @@ def validate_role_service_level_attributes_against_db():
                 role_for_attach.attach_service_level(new_service_level)
                 role_for_attach.validate_role_service_level_attributes_against_db()
 
-                # Workaround for issue https://github.com/scylladb/scylla-enterprise/issues/2572
-                if restart_scylla:
-                    nodes_for_restart = [node for node in tester.db_cluster.nodes if node.jmx_up() and node.db_up()
-                                         and not node.running_nemesis]
-                    tester.db_cluster.restart_binary_protocol(nodes=nodes_for_restart)
-                # End workaround
+                for node in tester.db_cluster.nodes:
```
juliayakovlev, line 123: Do you plan to handle this issue (https://github.com/scylladb/scylla-enterprise/issues/2572)? If yes - we still need the comment that it is workaround. So when the issue will be fixed - we will remove it from the code

juliayakovlev (commented):
Generally it looks ok. Small comment

juliayakovlev:
I ran the test with this commit (uses API) and received the result worse then when binary disable/enable is used as workaround.

Test ran with **SL REST API**:
https://argus.scylladb.com/test/d10575b9-86ef-485a-8bd7-089f423436d3/runs?additionalRuns[]=5426c441-6d6f-4cfb-8fe9-f10a91585e0f

test ran with **binary disable/enable**:
https://argus.scylladb.com/test/7bdbd517-fe3d-441b-aa66-a581404f305a/runs?additionalRuns[]=44214055-422a-4038-bd2e-bb9efcdc789d

juliayakovlev:
The issue was reproduced with using this commit: https://github.com/scylladb/scylla-enterprise/issues/2572#issuecomment-1873796736

Jadw1:
> @Jadw1
> 
> can you share a run demonstrating this works ?, also can you clarify in which release its gonna be available ?

Run with the PR: https://argus.scylladb.com/test/4b2b31e3-1d72-4af3-a281-cb436db14cc2/runs?additionalRuns[]=7a69fcb1-3f77-4982-8309-a95d370437fc

Run without the PR: https://argus.scylladb.com/test/5e62ce40-3da5-4002-9260-dbc5e545e1d4/runs?additionalRuns[]=af3a15a5-7630-41f4-a501-4f2dd27a0d78

Both those run was on the same scylla version (quite latest enterprise). Two of the nemesis failed on run with this PR but this might be due to https://github.com/scylladb/scylla-enterprise/issues/3737#issuecomment-1911024842. But all of the connections are working under correct scheduling group (the information is in sct runner's logs) and this is what this PR is fixing.
Without the PR 4 nemsis failed.

Note that for this PR to run correctly, https://github.com/scylladb/scylla-enterprise/pull/3242 has to be in scylla and **none of the 2023.x 2024.x have that PR**. That's why I've run the test against current enterprise version.
Maybe we should backport service levels REST API PR before merging this PR? Cc @eliransin @mykaul

Jadw1:
precommit failed but I don't see any clear error in jenkins logs. And precommit was successful locally on my machine

sdcm/sla/sla_tests.py
Jadw1: removed

sdcm/sla/sla_tests.py
Jadw1: added a comment

vponomaryov:
> precommit failed but I don't see any clear error in jenkins logs. And precommit was successful locally on my machine

Jenkins logs are following:
```
13:43:04  ⧗   input: sla_tests: use SL REST API instead of binary protocol on/off workaround
13:43:04  ✖   type must be one of [ci, docs, feature, fix, improvement, perf, refactor, revert, style, test, unit-test] [type-enum]
13:43:04  ✖   scope may not be empty [scope-empty]
```

So, it means that the commit topic violates expected structure.
It must be like following:
```
fix(sla_tests): use SL REST API instead of binary protocol on/off workaround
```

Jadw1:
@temichus @juliayakovlev @fruch Can we merge this?

juliayakovlev:
@Jadw1 which versions https://github.com/scylladb/scylla-enterprise/pull/3242 commit is / will be backported to?

Jadw1:
None of the versions have it. I guess we should backport it to 2024.1, idk how about 2023.
@mykaul @kostja Can you confirm/answer?

juliayakovlev:
> None of the versions have it. I guess we should backport it to 2024.1, idk how about 2023. @mykaul @kostja Can you confirm/answer?

@Jadw1 
If so - we can not merge this commit. It will fail the test. Or I miss something?

kostja:
@juliayakovlev I think we should not backport it unless there is an impact on the customer. Do you have a reason to backport it? Could you move this sct to run with the master instead?

mykaul:
> @juliayakovlev I think we should not backport it unless there is an impact on the customer. Do you have a reason to backport it? Could you move this sct to run with the master instead?

We are not going to backport it - it should indeed run on master.

fruch:
> > @juliayakovlev I think we should not backport it unless there is an impact on the customer. Do you have a reason to backport it? Could you move this sct to run with the master instead?
> 
> We are not going to backport it - it should indeed run on master.

this test isn't running regularly on master, it running as part of a release.
the question was mostly to understand where we should backport this one.

fruch:
> > > @juliayakovlev I think we should not backport it unless there is an impact on the customer. Do you have a reason to backport it? Could you move this sct to run with the master instead?
> > 
> > 
> > We are not going to backport it - it should indeed run on master.
> 
> this test isn't running regularly on master, it running as part of a release. the question was mostly to understand where we should backport this one.

FYI, whom would run SLA with master SCT, might run into it.
whom tests SLA, should be aware of it, otherwise he's in for a supprise.

fruch (approved):
LGTM

juliayakovlev:
If th

> @juliayakovlev I think we should not backport it unless there is an impact on the customer. Do you have a reason to backport it? Could you move this sct to run with the master instead?

@kostja @fruch 
if this rest api won't be part of version, how we can run the test and use it?

kostja:
It is part of master, so will be part of a future release. The question was do you need it for an older release? I'd presume no, unless there is a good reason to run this test with this old release (please share it).

PR #6951, opened 2023-12-15
improvement(k8s): reuse operator's `NodeConfig` for static vol provisioner

vponomaryov:
The Custom Resource Definition (CRD) called `NodeConfig` developed by the scylla-operator dev team is used to configure K8S nodes which are going to host Scylla pods.
It is able to configure SSD disks into RAID arrays. We were using it only running the dynamic local volume provisioner, but not the static one.

So, start using it for each of the volume provisioners we have. It will allow us to:
- Simplify existing code
- Stop using redundant pods - `scylla-machine-image` in EKS and `gcr.io/google-containers/startup-script` in GKE.

Also:
- Update the `bitnami/kubectl` image to be `1.26.11` that is the most recent one which supports `x86` and `ARM` architectures.
- Use much newer docker image for the `static local volume provisioner`. It also supports `x86` and `ARM` architectures.
- Update a bit the `static local volume provisioner` config to be aligned with the new image we switched to.

### DIFF
--- a/sdcm/k8s_configs/eks/scylla-node-prepare.yaml
+++ b/sdcm/k8s_configs/eks/scylla-node-prepare.yaml
@@ -79,7 +79,7 @@ spec:
       serviceAccountName: node-setup-daemonset
       containers:
       - name: cpu-policy
-        image: bitnami/kubectl:1.21.4
+        image: bitnami/kubectl:1.26.11
         imagePullPolicy: Always
         env:
           - name: NODE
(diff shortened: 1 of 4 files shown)


vponomaryov:
Tested here:
GKE-x86: https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-gke/153/
EKS-x86: https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-functional-eks/46/
EKS-arm: https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-functional-eks-arm/12/

sdcm/k8s_configs/eks/scylla-node-prepare.yaml
```diff
@@ -174,8 +174,9 @@ spec:
         app: node-pkg-installer
     spec:
       containers:
-      # NOTE: image is built on Feb 1, 2023 -> scylla-operator:1.9.0-alpha.1
-      - image: "scylladb/scylla-operator@sha256:10e3e7dddc2de1bbd473710223a4be5e8c5ef5dc9bb1a61d81fdc398d9dcd74c"
+      # TODO: set the following image to some static tag that supports also ARM arch (when appears)
+      #       As of Dec 15, 2023 only the 'latest' tag does support ARM.
+      - image: "scylladb/scylla-operator:latest"
```
fruch, line 179: Maybe we should set some specific sha of the manifest ?

fruch (approved):

sdcm/k8s_configs/eks/scylla-node-prepare.yaml
vponomaryov: Any `sha` is single-arch.

fruch (approved):
LGTM

PR #6952, opened 2023-12-15
feature(tablets_config): Add config file for tablets experimental

xemul:
The tablets feature also needs consistent-topology-changes to be enabled

### DIFF
--- a/configurations/tablets.yaml
+++ b/configurations/tablets.yaml
@@ -1,0 +1,4 @@
+append_scylla_yaml: |
+  experimental_features:
+    - tablets
+    - consistent-topology-changes


fruch:
We can merge it, 
but don't you want to introduce some pipelines that would point to this configuration and try them out ?

mykaul:
> We can merge it, but don't you want to introduce some pipelines that would point to this configuration and try them out ?

Let's merge and add pipelines in a different flow/PR.

xemul:
> We can merge it, but don't you want to introduce some pipelines that would point to this configuration and try them out ?

I though that we merge it, then wait for AMI with scylladb/scylladb#16364, then take _existing_ pipelines and run them with and without this config (AFAICS config files are editable when one creates a job) thus comparing tables vs non-tablets

fruch:
> > We can merge it, but don't you want to introduce some pipelines that would point to this configuration and try them out ?
> 
> I though that we merge it, then wait for AMI with [scylladb/scylladb#16364](https://github.com/scylladb/scylladb/pull/16364), then take _existing_ pipelines and run them with and without this config (AFAICS config files are editable when one creates a job) thus comparing tables vs non-tablets

yes, one can change the parameters in the job, but we usually prefer to specific jobs one can refer to, and that can be triggered without knowing in advance which configuration you need to add 

I would also assume you'll need more than those two experimental configuration to enable tablets

fruch (approved):
LGTM

xemul:
> I would also assume you'll need more than those two experimental configuration to enable tablets

That's the point -- with https://github.com/scylladb/scylladb/pull/16364 no other changes are needed

xemul:
> yes, one can change the parameters in the job, but we usually prefer to specific jobs one can refer to

I don't disagree, but I think that's the next step. Right now the goal is to unblock the development and let them run comparison tests -- tablets off vs tablets on. After a while we'll know better which exact tests we need on the regular basis and add necessary pipelines

fruch:
> > yes, one can change the parameters in the job, but we usually prefer to specific jobs one can refer to
> 
> I don't disagree, but I think that's the next step. Right now the goal is to unblock the development and let them run comparison tests -- tablets off vs tablets on. After a while we'll know better which exact tests we need on the regular basis and add necessary pipelines

O.k. 
Here's a referal of the work @soyacz did few monthes back (on the reorg we move all the information to @temichus team)
https://docs.google.com/document/d/1cFDx3IeSWysnWfG5G-Qj8nzlSVQvcP9AUZ0fKhh_dPU/edit?usp=sharing

I would recommend using the pipelines under `nemesis/` so you can test each nemesis on it's own
there a small subset of it, we run every day, so you can have reference

mykaul:
How can I easily set the initial no. of tablets?

fruch:
> How can I easily set the initial no. of tablets?

depends on you definition of easy, but one need to track for each test case how the table is generated, and if it's generated by external tool  c-s, s-b, gemini and such, one need to add a call to build the keyspace/table before the tool

ShlomiBalalis:
@rayakurl @scylladb/qa-maintainers Please backport to the manager brnaches

Issue #6943, opened 2023-12-13
[K8S] Scylla-operator upgrade scenario doesn't catch failure upgrading `ScyllaCluster` object and continues to work with old one


vponomaryov:
The upgrade of the scylla-operator from `v1.10.0` version to the `v1.11.0-rc.0` failed to trigger the `ScyllaCluster` object upgrade, then our test didn't catch it and continued working with old state of the `ScyllaCluster` object.
So, it comes out that the upgrade was not completed as it should.

Error that should have been caught:
```
status:
  conditions:
...
  - lastTransitionTime: "2023-10-05T22:33:32Z"
    message: 'can''t apply statefulset update: can''t update apps/v1, Kind=StatefulSet
      "scylla/sct-cluster-us-east1-b-us-east1": StatefulSet.apps "sct-cluster-us-east1-b-us-east1"
      is invalid: spec: Forbidden: updates to statefulset spec for fields other than
      ''replicas'', ''ordinals'', ''template'', ''updateStrategy'', ''persistentVolumeClaimRetentionPolicy''
      and ''minReadySeconds'' are forbidden'
    observedGeneration: 4
    reason: Error
    status: "True"
    type: StatefulSetControllerDegraded
```

Scylla pods are unexpectedly old:
```
scylla-operator    scylla-operator-585b8b677c-wv4gc     1/1    Running    1 (8m25s ago)   8m26s   10.0.2.67
scylla-operator    scylla-operator-585b8b677c-xxzqr     1/1    Running    1 (8m26s ago)   8m33s   10.0.1.194
scylla-operator    webhook-server-cffd5fd54-cbmdx       1/1    Running    0               8m13s   10.0.2.11
scylla-operator    webhook-server-cffd5fd54-njmrh       1/1    Running    0               8m33s   10.0.1.61
scylla             sct-cluster-us-east1-b-us-east1-0    2/2    Running    1 (113m ago)    115m    10.0.1.199
scylla             sct-cluster-us-east1-b-us-east1-1    2/2    Running    1 (110m ago)    112m    10.0.0.50
scylla             sct-cluster-us-east1-b-us-east1-2    2/2    Running    1 (106m ago)    108m    10.0.2.86
scylla             sct-cluster-us-east1-b-us-east1-3    2/2    Running    1 (4m59s ago)   7m23s   10.0.3.117
```

These old Scylla pods have following old init container image version:
```
  hostname: sct-cluster-us-east1-b-us-east1-0
[...]

PR #6963, opened 2023-12-18
fix(k8s): fix scylla-operator upgrade test

vponomaryov:
Current `scylla-operator upgrade` logic doesn't check that
- The `ScyllaCluster` object gets upgraded right after the `scylla-operator`.
- The `status.conditions` doesn't have errors
- The `init` container of the Scylla pods are of the new version.

It was sleeping for 1 minute and waiting for the rollout to finish.
But if, for any reason, rollout doesn't start, the code picks up the old `rollout` state and moves on with the old `ScyllaCluster` K8S object state.

So, add proper checks to cover problems with the `ScyllaCluster` upgrades:
- Check that scylla pods get new `UID` values meaning they were recreated
- Read ScyllaCluster's `status.conditions` and look for Error's
- Check all the `scylla-operator` images to be of the new version.

Closes: #6943

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -1076,7 +1076,7 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
             namespace=self.k8s_cluster._scylla_operator_namespace  # pylint: disable=protected-access
         ).stdout.strip().split(":")[-1]
 
-    def test_kubernetes_operator_upgrade(self):
+    def test_kubernetes_operator_upgrade(self):  # pylint: disable=too-many-locals,too-many-statements
         self.k8s_cluster.check_scylla_cluster_sa_annotations()
 
         InfoEvent(message='Step1 - Populate DB with data').publish()
@@ -1092,6 +1092,13 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
             stress_cmd=self._cs_add_node_flag(self.params.get('stress_cmd_r'))))
 
         InfoEvent(message='Step4 - Upgrade scylla-operator').publish()
+        old_scylla_pods_uids = {}
+        for scylla_pod in self.db_cluster.nodes:
+            old_scylla_pods_uids[scylla_pod.name] = scylla_pod.k8s_pod_uid
+        progressing_last_transition_time_before_upgrade = self.k8s_cluster.kubectl(
+            f"get --no-headers scyllacluster {self.params.get('k8s_scylla_cluster_name')}"
+            " -o jsonpath='{.status.conditions[?(@.type==\"Progressing\")].lastTransitionTime}'",
+            namespace=self.db_cluster.namespace).stdout.strip()
         base_docker_image_tag = self._get_current_operator_image_tag()
         upgrade_docker_image = self.params.get('k8s_scylla_operator_upgrade_docker_image') or ''
         self.k8s_cluster.upgrade_scylla_operator(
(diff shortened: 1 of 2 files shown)


upgrade_test.py
```diff
@@ -1111,16 +1114,38 @@ def test_kubernetes_operator_upgrade(self):
...
+            f"get scyllacluster {self.params.get('k8s_scylla_cluster_name')} -o json",
+            namespace=self.db_cluster.namespace).stdout.strip())["status"]["conditions"]
+        for condition in scylla_cluster_conditions:
+            assert condition["reason"] != "Error", "'ScyllaCluster' failed to be updated: %s" % condition
+
+        pods_with_wrong_image_tags = []
+        pods = json.loads(self.k8s_cluster.kubectl(
+            "get pods -o json", namespace=self.db_cluster.namespace).stdout.strip())["items"]
+        for pod in pods:
+            for container_type in ("c", "initC"):
```
fruch, line 1133: can you add a comment what are those types ?

fruch (dismissed):

upgrade_test.py
vponomaryov: Each pod has `containers` and `initContainers`.
So, here we have the loop to cover both container types.

vponomaryov:
Upgrade job PASSes when there is no scylla-operator bug: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-upgrade-operator-k8s-eks/17/consoleFull
Upgrade job FAILs when there is scylla-operator bug: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-upgrade-operator-k8s-eks/15/consoleFull

upgrade_test.py
```diff
@@ -1111,16 +1114,38 @@ def test_kubernetes_operator_upgrade(self):
...
+            scylla_pod.wait_till_k8s_pod_get_uid(
+                ignore_uid=old_scylla_pods_uids[scylla_pod.name], throw_exc=True, timeout=900)
+        for scylla_pod in self.db_cluster.nodes[::-1]:
+            scylla_pod.wait_for_pod_readiness()
+
+        scylla_cluster_conditions = json.loads(self.k8s_cluster.kubectl(
+            f"get scyllacluster {self.params.get('k8s_scylla_cluster_name')} -o json",
+            namespace=self.db_cluster.namespace).stdout.strip())["status"]["conditions"]
+        for condition in scylla_cluster_conditions:
+            assert condition["reason"] != "Error", "'ScyllaCluster' failed to be updated: %s" % condition
```
tnozicka, line 1127: you should check `Available=True,Progressing=False,Degraded=False`
https://github.com/scylladb/scylla-operator/blob/eb144ac50175c040e486f47eb9468908181ce678/pkg/api/scylla/v1/types_cluster.go#L543

tnozicka: nit: I assume this could also check that Progressing condition has a last update timestamp > last check before upgrade, to be sure it changed

vponomaryov: Done

upgrade_test.py
vponomaryov: Added comment

vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-upgrade-operator-k8s-eks/18/consoleFull

fruch (approved):
LGTM

PR #6965, opened 2023-12-18
fix(MV-errors): skip all view update error globaly

fruch:
we had multiple places where we tried to apply a filtering/demoating of view update errors, and they keep popping up in all kind of cases

* cases of parallel nemesis
* cases our log reading slow down, and those pop out of context, since filter is gone

so cause of those issue, and the fact those aren't gonna be fixed any time soon, we'll apply this filter globaly until all of the view update issues would be addressed

Ref: https://github.com/scylladb/scylladb/issues/16206
Ref: https://github.com/scylladb/scylladb/issues/16259
Ref: https://github.com/scylladb/scylladb/issues/15598

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -76,12 +76,16 @@ from sdcm.sct_events import Severity
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.decorators import raise_event_on_failure
 from sdcm.sct_events.filters import DbEventsFilter, EventsSeverityChangerFilter, EventsFilter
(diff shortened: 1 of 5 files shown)


roydahan:
I'm fine with it, but we need to mention in one of the issues that we applied this global filter (with link to this PR).
And in a Todo comment here referencing to the MV issue.

fruch:
> I'm fine with it, but we need to mention in one of the issues that we applied this global filter (with link to this PR). And in a Todo comment here referencing to the MV issue.

Done

roydahan (approved):

PR #6967, opened 2023-12-19
fix(version_utils): support more version options

fruch:
now that we get version from more sources (from images tags) we have cases the version looks like:

`2024.2.0.dev.0.20231219.c7cdb16538f2.1`

i.e. dash was replace by a dot, by some part of the SMI code (cause of GCP tag limitations)

so `ComparableScyllaOperatorVersion` was extended to support it

### DIFF
--- a/sdcm/utils/version_utils.py
+++ b/sdcm/utils/version_utils.py
@@ -55,7 +55,7 @@ SEMVER_REGEX = re.compile(
         (?P<minor>0|[1-9]\d*)
         \.
         (?P<patch>0|[1-9]\d*)
-        (?:-(?P<prerelease>
+        (?:[-.](?P<prerelease>
             (?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)
             (?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*
         ))?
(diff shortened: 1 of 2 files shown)


sdcm/utils/version_utils.py
```diff
@@ -55,7 +55,7 @@
         (?P<minor>0|[1-9]\d*)
         \.
         (?P<patch>0|[1-9]\d*)
-        (?:-(?P<prerelease>
+        (?:[-.](?P<prerelease>
```
vponomaryov, line 58: This will fix the problem from the SCT side.
But it is not correct thing to do.
This is not correct because it won't be the `SEMVER_REGEX` anymore.

unit_tests/test_version_utils.py
```diff
@@ -608,6 +608,7 @@ def test_comparable_scylla_operator_versions_compare(version_string_left, versio
     ("1.9.0-alpha.1-nightly", "1.9.0-alpha.1"),
     ("1.9.0-alpha.1-2-g3321624", "1.9.0-alpha.1-2-g3321624"),
     ("1.9.0-alpha.1-2-g3321624-nightly", "1.9.0-alpha.1-2-g3321624"),
+    ("2024.2.0.dev.0.20231219.c7cdb16538f2.1", "2024.2.0-dev.0.20231219.c7cdb16538f2.1"),
```
vponomaryov, line 611: This is the `ComparableScyllaOperatorVersion` unit test, not the `ComparableScyllaVersion.

vponomaryov:
@fruch I will upload the PR with the extension of the `parse` method.

vponomaryov:
Here is the proper fix/improvement: https://github.com/scylladb/scylla-cluster-tests/pull/6970

PR #6975, opened 2023-12-19
fix(monitoring): don't specify scylla target ports

fruch:
seems like since scylladb/scylla-monitoring#2000, cause we are specifying the ports, monitoring is using those also for the agents, and cause all scylla metric to be duplicated

Ref: scylladb/scylla-enterprise#3708

### Testing

- [x] - ipv6 - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-ipv6-test/4/

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5378,7 +5378,7 @@ class BaseMonitorSet:  # pylint: disable=too-many-public-methods,too-many-instan
         for node in self.nodes:
             monitoring_targets = []
             for db_node in self.targets["db_cluster"].nodes:
-                monitoring_targets.append(f"{normalize_ipv6_url(getattr(db_node, self.DB_NODES_IP_ADDRESS))}:9180")
+                monitoring_targets.append(f"{normalize_ipv6_url(getattr(db_node, self.DB_NODES_IP_ADDRESS))}")
             monitoring_targets = " ".join(monitoring_targets)
             node.remoter.sudo(shell_script_cmd(f"""\
                 cd {self.monitor_install_path}


amnonh:
It looks fine, but needs to be tested just because you are using ipv6

fruch:
> It looks fine, but needs to be tested just because you are using ipv6

if we pass `[ipv6]` without specifying  a port, it won't work ?

amnonh:
I'm not 100% sure, better check

soyacz:
restoring monitor from https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-ipv6-test/2/ shows no data

fruch:
> restoring monitor from https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-ipv6-test/2/ shows no data

yeah, just saw it.

I would need to ditch the brackets

amnonh:
There used to be a limitation on IPv6 that forced you to add port numbers, I've just created [scylla-monitoring#2136](https://github.com/scylladb/scylla-monitoring/issues/2136) to remove that limitation.

amnonh:
you have

> > restoring monitor from https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-ipv6-test/2/ shows no data
> 
> yeah, just saw it.
> 
> I would need to ditch the brackets

ipv6 must use brackets

fruch:
> you have
> 
> > > restoring monitor from https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-ipv6-test/2/ shows no data
> > 
> > 
> > yeah, just saw it.
> > I would need to ditch the brackets
> 
> ipv6 must use brackets

by spec, it should used them if port is needed to be specified.

but yes last run showed, it doesn't work either way, with or without brackets

amnonh:
I've already fixed and backported, so now you need to use brackets with no ports and it should work

fruch:
> I've already fixed and backported, so now you need to use brackets with no ports and it should work

Validation is really needed for all those configuration options, you can easily get into a broken setup, cause of all those things.

PR #6983, opened 2023-12-20
fix(aws_builder): change the builders to configure gp3 root disks

fruch:
update the launch templates to use gp3 disk,

and introduce code that can create new launch template version and update them to be the default, before this change the only check we did if the launch template existed or not, and never update anything with those templates
#### Testing
- [x] apply it on one region and test
- [x] apply it to all regions

### DIFF
--- a/sdcm/utils/aws_builder.py
+++ b/sdcm/utils/aws_builder.py
@@ -213,6 +213,60 @@ class AwsBuilder:
         self.associate_elastic_ip()
         self.add_to_jenkins()
 
+    @property
+    def launch_template_name(self):
+        return "aws-sct-builders"
+
+    def get_root_ebs_info_from_ami(self, ami_id: str) -> str:
+        res = self.region.resource.Image(ami_id)
+        return res.block_device_mappings[0].get('Ebs', {})
+
+    def get_launch_template_data(self, runner: AwsSctRunner) -> dict:
+        return dict(
+            LaunchTemplateData={
+                'BlockDeviceMappings': [
+                    {
+                        "DeviceName": "/dev/sda1",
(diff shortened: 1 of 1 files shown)


fruch:
@soyacz @vponomaryov @yaronkaikov   for awareness that i'm gonna update all of the ASGs SCT is using, so it might have some effect I didn't know about.

I did the change for eu-west-1 (and for the ASG used by SCT CI)
After feedback on the changes here, I'll make it on all regions (and merged it)

yaronkaikov (approved):

sdcm/utils/aws_builder.py
```diff
@@ -114,104 +114,59 @@ def name(self):
...
-        instances = self.region.client.describe_instances(Filters=[{"Name": "tag:Name",
-                                                                    "Values": [self.name]},
-                                                                   {"Name": "instance-state-name",
-                                                                    "Values": ["running"]}])
-
-        if instances["Reservations"] and instances["Reservations"][0]["Instances"]:
-            return self.region.resource.Instance(instances["Reservations"][0]["Instances"][0]['InstanceId'])
-
-        tags = {
-            "bastion": "true",
```
soyacz, line 135: so we no longer use builders as bastions?

vponomaryov (approved):
LGTM

sdcm/utils/aws_builder.py
fruch: we use the sct-runner available on that region, if we'll see it's not enough, we'll thing if we want few machines like that, but they won't be connect to jenkins.

fruch: in GCP so far we don't have any, no one complain about it. (so far)

PR #6985, opened 2023-12-21
fix(dns): Change monitor instance type

Mark-Gurevich:
The Scylla server for scylla-manager failed to start on t2.small instance type on the monitoring node in features-dns test. Removed the paramter from thsi test yaml config file.

The test passed in my staging folder with bigger instance for the monitoring node:

https://jenkins.scylladb.com/job/scylla-staging/job/mark/job/features-dns-cluster-5min-test/

https://argus.scylladb.com/workspace?state=WyIxMzBhODk4YS0wMWZlLTQyODQtODg1My05ODE2NmQ5ZGYwOGUiXQ

### DIFF
--- a/test-cases/features/dns-cluster-5min.yaml
+++ b/test-cases/features/dns-cluster-5min.yaml
@@ -10,6 +10,5 @@ n_monitor_nodes: 1
 
 instance_type_db: 'i4i.2xlarge'
 instance_type_loader: 'c5.4xlarge'
-instance_type_monitor: 't2.small'
 
 user_prefix: 'cases-dns'


mykaul:
I don't understand this change at all:
1. If the default is now different, why not remove the parameter altogether?
2. If t2.small is wrong, there are many other tests using it!

Mark-Gurevich:
> I don't understand this change at all:
> 
> 1. If the default is now different, why not remove the parameter altogether?

You are right. If we stick with default its better to remove the param in this particular test config file. Otherwise, we may try with smaller instance like `t3.medium` for example.

> 2. If t2.small is wrong, there are many other tests using it! (https://github.com/search?q=repo%3Ascylladb%2Fscylla-cluster-tests%20%22t2.small%22&type=code )

There are 41 YAML config files with `t2.small`, `t3.small`, `t3.large`, `t3.xlarge`, `m5.2xlarge` values.
Maybe we should remove some of them and/or change the default.
Need expert advice here.
@fruch

roydahan:
Why does the monitor fail to start on the t2.small?
Did you verify it was able to start with the t3 default.

It's a very short test, small monitor instance should have suffice here.

soyacz:
@Mark-Gurevich probably you need to start scylla server for Manager with developer mode.

fruch (approved):
LGTM

fruch:
> Why does the monitor fail to start on the t2.small? 
> Did you verify it was able to start with the t3 default.

since we use manager by default, and scylla can't run on `t2.small` anymore.
the default should be o.k. for this case. 

> 
> It's a very short test, small monitor instance should have suffice here.

fruch:
> > I don't understand this change at all:
> > 
> > 1. If the default is now different, why not remove the parameter altogether?
> 
> You are right. If we stick with default its better to remove the param in this particular test config file. Otherwise, we may try with smaller instance like `t3.medium` for example.
> 
> > 2. If t2.small is wrong, there are many other tests using it! (https://github.com/search?q=repo%3Ascylladb%2Fscylla-cluster-tests%20%22t2.small%22&type=code )
> 
> There are 41 YAML config files with `t2.small`, `t3.small`, `t3.large`, `t3.xlarge`, `m5.2xlarge` values. Maybe we should remove some of them and/or change the default. Need expert advice here. @fruch

no, we should remove `t2.small` and `t3.small`

the default was change in bbdc4ab657e246b13c17c81f83b3086bbadf0b84

but someone didn't clean it up.

fruch:
also all the cases using t2.small, are quite old one or one not in use (or used by unittests)
all of the t3.small, are mostly in perf tests that seems to be disabling manager (and most of those are not being run regularly)

PR #7000, opened 2023-12-25
fix(modify_table_speculative_retry): Set to reasonable numbers

roydahan:
The modify_table nemesis which picks one of the table properties and alter them can choose a very low percentile or value for the speculative_retry which can cause high load of retries that may fail the stress eventually.


fruch:
@roydahan 
that mean we have lots of places we count on the `speculative_retry`, and not only on c-s retries ?

roydahan:
I think that's in addition to it, not instead of.

fruch (approved):
LGTM

PR #7001, opened 2023-12-25
improvement(parallel-schema-changes-12h): Increase nemesis_multiply

roydahan:
The stack of nemesis is running out after several hours into the test. Hence, we would like to increase the nemesis_multiply_factor to higher than the default (6). I chose 17 as a special random number.

### DIFF
--- a/test-cases/longevity/longevity-parallel-schema-changes-12h.yaml
+++ b/test-cases/longevity/longevity-parallel-schema-changes-12h.yaml
@@ -24,6 +24,7 @@ nemesis_selector: [["topology_changes"],["!disruptive","schema_changes"]]
 nemesis_interval: 10
 nemesis_filter_seeds: false
 nemesis_during_prepare: false
+nemesis_multiply_factor: 17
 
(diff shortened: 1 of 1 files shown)


fruch:
This might create a clustering of the same nemesis for the shorter runs of this job

roydahan:
Do the shorter runs use the same yaml?
If so, need to specify also this parameter in their config file.

fruch:
> Do the shorter runs use the same yaml? If so, need to specify also this parameter in their config file.

yes they do use the same yaml, this is part of the tier1 cases.

the better solution is to cycle in that list

fruch:
> > Do the shorter runs use the same yaml? If so, need to specify also this parameter in their config file.
> 
> yes they do use the same yaml, this is part of the tier1 cases.
> 
> the better solution is to cycle in that list

see https://github.com/scylladb/scylla-cluster-tests/pull/7005

roydahan:
Fine by me.
(BTW, we have the same multiply by 18 in the similar tests with multi-dc).

PR #7005, opened 2023-12-25
fix(nemesis): cycle the nememsis_list

fruch:
instend of pop nemesis of of the list, and assuming we'll have enough of them for the run, we are now cycling ontop of that list for ever.

all the other parameters, like `nemesis_seed` and `nemesis_multiply_factor` still apply for making the list, but now we don't need to fine tune cases base on the length they are running

### Testing

- [x] provision aws/gcp

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -28,6 +28,7 @@ import re
 import time
 import traceback
 import json
+import itertools
 from distutils.version import LooseVersion
 from contextlib import ExitStack
 from typing import Any, List, Optional, Type, Tuple, Callable, Dict, Set, Union, Iterable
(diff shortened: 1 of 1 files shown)


roydahan:
I know we discussed it and I said it's an acceptable behaviour.
Now when looking at the code I think that my original intention was to support cases that we want to run till we cover the entire list and then to stop the test.

I know we didn't do it (it's my "mini-test" that I never got back to it) and eventually we chose the individual nemesis approach.
Do you think the use case still relevant so we shall support both?

roydahan (approved):
LGTM

fruch:
> I know we discussed it and I said it's an acceptable behaviour.
> Now when looking at the code I think that my original intention was to support cases that we want to run till we cover the entire list and then to stop the test.
> 
> I know we didn't do it (it's my "mini-test" that I never got back to it) and eventually we chose the individual nemesis approach.
> Do you think the use case still relevant so we shall support both?

We could support both, just populate the cycle with iter, and catch StopIteration

But let's do that when it's needed, meanwhile the cycle case is more common

PR #7006, opened 2023-12-25
feature(start_scylla_server): add adaptive timeout

fruch:
since we have seen a few time scylla might take more time to start then we might anticipate:

```
sdcm.remote.libssh2_client.exceptions.CommandTimedOut: Command did not complete within 500 seconds!
Command: 'sudo systemctl start scylla-server.service'
Stdout:
Stderr:
```

we now default to soft timeout of 500s same the timeout before, and hard timeout of 2000s (~30min)

### Testing

- [x] - provision test
- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-5gb-1h-StopWaitStartMonkey-aws-test/2

### Reminders

### DIFF
--- a/sdcm/utils/adaptive_timeouts/__init__.py
+++ b/sdcm/utils/adaptive_timeouts/__init__.py
@@ -55,6 +55,7 @@ class Operations(Enum):
     DECOMMISSION = ("decommission", _get_decommission_timeout, ())
     NEW_NODE = ("new_node", _get_soft_timeout, ("timeout",))
     CREATE_INDEX = ("create_index", _get_soft_timeout, ("timeout",))
+    START_SCYLLA = ("start_scylla", _get_soft_timeout, ("timeout",))
     CREATE_MV = ("create_mv", _get_soft_timeout, ("timeout",))
     MAJOR_COMPACT = ("major_compact", _get_soft_timeout, ("timeout",))
     REPAIR = ("repair", _get_soft_timeout, ("timeout",))
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2317,7 +2317,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         verify_up_timeout = verify_up_timeout or self.verify_up_timeout
         if verify_down:
             self.wait_db_down(timeout=timeout)
-        self.start_service(service_name='scylla-server', timeout=timeout)
+        with adaptive_timeout(operation=Operations.START_SCYLLA, node=self, timeout=timeout):
+            self.start_service(service_name='scylla-server', timeout=timeout * 4)
         if verify_up:
             self.wait_db_up(timeout=verify_up_timeout)
 


sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -55,6 +55,7 @@ class Operations(Enum):
     DECOMMISSION = ("decommission", _get_decommission_timeout, ())
     NEW_NODE = ("new_node", _get_soft_timeout, ("timeout",))
     CREATE_INDEX = ("create_index", _get_soft_timeout, ("timeout",))
+    START_SCYLLA = ("create_index", _get_soft_timeout, ("timeout",))
```
vponomaryov, line 58: Typo in the first argument

vponomaryov (changes requested):

sdcm/utils/adaptive_timeouts/__init__.py
fruch: right, not total clear why we aren't using the enum name for this purpose...

vponomaryov (approved):
LGTM

juliayakovlev:
This is Avi's comment:

```
Yes startup can be delayed for commitlog replay, reshape, or reshard. It's good to warn when startup takes too long, but we shouldn't fail when one of these events causes startup to be slow.
```

So we should not create error event in this case

fruch:
> This is Avi's comment:
> 
> ```
> Yes startup can be delayed for commitlog replay, reshape, or reshard. It's good to warn when startup takes too long, but we shouldn't fail when one of these events causes startup to be slow.
> ```
> 
> [scylladb/scylla-enterprise#3723 (comment)](https://github.com/scylladb/scylla-enterprise/issues/3723#issuecomment-1869468148)
> 
> So we should not create error event in this case

we increased the hard timeout by 4, so we should be able to continue to the next phase.
and there would be an error event to warn us if it take more then 500s

later when we have more information (cause it is collected), we can see better numbers, 
we can't have infinity and not errors about it

juliayakovlev:
> > This is Avi's comment:
> > ```
> > Yes startup can be delayed for commitlog replay, reshape, or reshard. It's good to warn when startup takes too long, but we shouldn't fail when one of these events causes startup to be slow.
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > [scylladb/scylla-enterprise#3723 (comment)](https://github.com/scylladb/scylla-enterprise/issues/3723#issuecomment-1869468148)
> > So we should not create error event in this case
> 
> we increased the hard timeout by 4, so we should be able to continue to the next phase. and there would be an error event to warn us if it take more then 500s
> 
> later when we have more information (cause it is collected), we can see better numbers, we can't have infinity and not errors about it

[here](https://github.com/scylladb/scylla-enterprise/issues/3723#issuecomment-1869468148) start up took about 31min

fruch:
> > > This is Avi's comment:
> > > ```
> > > Yes startup can be delayed for commitlog replay, reshape, or reshard. It's good to warn when startup takes too long, but we shouldn't fail when one of these events causes startup to be slow.
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > [scylladb/scylla-enterprise#3723 (comment)](https://github.com/scylladb/scylla-enterprise/issues/3723#issuecomment-1869468148)
> > > So we should not create error event in this case
> > 
> > 
> > we increased the hard timeout by 4, so we should be able to continue to the next phase. and there would be an error event to warn us if it take more then 500s
> > later when we have more information (cause it is collected), we can see better numbers, we can't have infinity and not errors about it
> 
> [here](https://github.com/scylladb/scylla-enterprise/issues/3723#issuecomment-1869468148) start up took about 31min

31 until systemctl returned ? or until scylla opened CQL ports ?, I'm not sure it's the same, we are waiting for scylla to be listening for much more 500s

juliayakovlev:
@fruch 
ports became opened after resharding finish, right befor "Scylla version init completed"

fruch:
> ports became opened after resharding finish, right befor "Scylla version init completed"

and `systemctl start` doesn't wait for ports to be opened

it's actually based on this message 
https://github.com/scylladb/scylladb/commit/3053d63c7f7a837b898fafda63c27f55b0afc943

PR #7008, opened 2023-12-25
fix(upgrade_test): remove raft from upgrades

fruch:
since `consistent_cluster_management` is deprecated, and enabled by default,
we no longer need by default to use it in upgrade test and we can remove the test doesn't enable/disabled

### Testing

- [ ] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/rolling-upgrade-ami-test/7/

### Reminders


fruch:
@temichus 

there is still other usage of `consistent_cluster_management` in the code, especially in `get_raft_mode` which I have no clue how it should change

also there are lot of other cases using `consistent_cluster_management`, that need to be removed or changed.

fruch:
this PR on it's own doesn't help the tooling issue, we'll need to remove `consistent_cluster_management` from configuration before dump the sstable, as a workaround

roydahan:
I don't want to throw  all this code and jobs.
Raft previously was "consistent_cluster_management" now, it should be changed to "Consistent_topology_changes".

@temichus I'm closing this PR, and what's needed from your team is just to replace that parameter with the new parameter and have all the jobs and behaviour the same (or review and tell if it should be the same).

Issue #6553, opened 2023-08-30
2 nemesis: NonDisruptive and Disruptive running on same node.


aleksbykov:
### Prerequisites

* [x] Are you rebased to master ?
* [ ] Is it reproducible ?
* [x] Did you perform a cursory search if this issue isn't opened ?

### Versions
* **SCT**: master
* **scylla**: master

### Logs
* test_id: 96a7d2d0-91a1-45a6-a853-2e383a82cb1d
* job log: https://jenkins.scylladb.com/job/scylla-master/job/longevity/job/longevity-schema-topology-changes-12h-test/9/
* argus: https://argus.scylladb.com/test/024e78ba-3671-4fd0-bafc-b48b2a978f59/runs?additionalRuns[]=96a7d2d0-91a1-45a6-a853-2e383a82cb1d

### Description

2 nemesises running on same node, was cause that 1 nemesis was failed.

1: disrupt_toggle_table_ics | longevity-parallel-topology-schema--db-node-96a7d2d0-5 | Failed | 2023-08-26 05:33:44 | 2023-08-26 05:34:08
2: disrupt_restart_with_resharding | longevity-parallel-topology-schema--db-node-96a7d2d0-5 | Succeeded | 2023-08-26 05:23:06 | 2023-08-26 05:34:53

_disrupt_toggle_table_ics_ start on node where still running disruptive nemesis _disrupt_restart_with_resharding_.

This cause the _disrupt_toggle_table_ics_ to failed with reset connection

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:** [What you expected to happen]

**Actual behavior:** [What actually happened]

fruch:
Assuming this time isn't part of the health checks,

How exactly was it causing it to fail ?
Please put here a bit more information that some would be able to work with.

fruch:
seen again on 2024.1 runs

this nemesis failed to communicate with the node, since it was currently being drained
```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5062, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2783, in disrupt_toggle_table_ics
    self.toggle_table_ics()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2500, in toggle_table_ics
[...]

PR #7016, opened 2023-12-26
feature(nemesis): introduce lock for target selection

fruch:
since we run into multiple cases on parallel nemesis that there were multiple nemesis using the same node

we are introducing a lock over the selection of target nodes so we won't be able to pick it multiple times

Fixes: #6553

### Testing

- [x] parallel nemesis case with enterprise build - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis-test/4/ https://argus.scylladb.com/test/eedcde34-5d24-4c94-84d7-07608f49b9da/runs?additionalRuns[]=af0022c1-586b-4556-989f-f6aec800f2b2
- [x] provision tests

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -28,7 +28,7 @@ import traceback
 import itertools
 import json
 import ipaddress
-
+from importlib import import_module
 from typing import List, Optional, Dict, Union, Set, Iterable, ContextManager, Any, IO, AnyStr
 from datetime import datetime
 from textwrap import dedent
@@ -52,7 +52,6 @@ from cassandra.cluster import NoHostAvailable  # pylint: disable=no-name-in-modu
 from cassandra.policies import RetryPolicy
 from cassandra.policies import WhiteListRoundRobinPolicy, HostFilterPolicy, RoundRobinPolicy
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
-
 from argus.backend.util.enums import ResourceState
 from sdcm.node_exporter_setup import NodeExporterSetup, SyslogNgExporterSetup
 from sdcm.db_log_reader import DbLogReader
@@ -150,7 +149,6 @@ from sdcm.exceptions import (
     NodeNotReady,
     SstablesNotFound,
 )
-from sdcm.utils.context_managers import run_nemesis
 
 # Test duration (min). Parameter used to keep instances produced by tests that
 # are supposed to run longer than 24 hours from being killed
@@ -287,7 +285,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         self.stop_wait_db_up_event = threading.Event()
         self.lock = threading.Lock()
 
-        self._running_nemesis = None
+        self.running_nemesis = None
 
         # We should disable bootstrap when we create nodes to establish the cluster,
         # if we want to add more nodes when the cluster already exists, then we should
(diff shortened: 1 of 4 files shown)


sdcm/cluster.py
```diff
@@ -4338,9 +4316,10 @@ def _rotate_kms_key(kms_key_alias_name, kms_key_rotation_interval, db_cluster):
                         message=f"Failed to rotate AWS KMS key for the '{kms_key_alias_name}' alias",
                         traceback=traceback.format_exc()).publish()
                 try:
-                    target_node = [node for node in db_cluster.nodes if not node.running_nemesis][0]
-                    self.log.debug("Target node for 'rotate_kms_key' is %s", target_node.name)
-                    with run_nemesis(node=target_node, nemesis_name="KMS encryption check"):
+                    nemesis_class = self.nemesis[0] if self.nemesis else getattr(
+                        import_module('sdcm.nemesis'), "Nemesis")
+                    with nemesis_class.run_nemesis(node_list=db_cluster.nodes, nemesis_label="KMS encryption check") as target_node:
```
vponomaryov, line 4321: In parameters you provide list of nodes but in the context manager you save the result as `one` node.
Is it expected? If yes, why so? Looks not intuitive and confusing...

sdcm/nemesis.py
```diff
@@ -292,6 +294,23 @@ def wrapper(self, *args, **kwargs):
...
+    @staticmethod
+    @contextmanager
+    def run_nemesis(node_list: list['BaseNode'], nemesis_label: str):
+        """
+        pick a node out of a `node_list`, and mark is as running_nemesis
+        for the duration of this context
+        """
+        with NEMESIS_TARGET_SELECTION_LOCK:
+            free_nodes = [node for node in node_list if not node.running_nemesis]
+            node = random.choice(free_nodes)
```
vponomaryov, line 306: Shouldn't it have check for emptiness?

fruch: we can, check and raise error, anyhow error would be raise if list would be empty :)

vponomaryov: If we can raise expected error then why not?

sdcm/cluster.py
fruch: it's expected, this context manager is selecting a node, so the logic of how to select a node that isn't during nemesis
is encapsulated in it.

not sure what the confusion here

sdcm/nemesis.py
fruch: fixed it, and now we do

sdcm/cluster.py
vponomaryov: Strange logic, but o.k.

sdcm/cluster.py
```diff
@@ -4394,9 +4371,11 @@ def _rotate_kms_key(kms_key_alias_name, kms_key_rotation_interval, db_cluster):
                         message=f"Failed to rotate AWS KMS key for the '{kms_key_alias_name}' alias",
                         traceback=traceback.format_exc()).publish()
                 try:
-                    target_node = [node for node in db_cluster.nodes if not node.running_nemesis][0]
                     self.log.debug("Target node for 'rotate_kms_key' is %s", target_node.name)
```
vponomaryov, line 4374: The `target_node` is not defined at this step.
The `precommit` also complains about it.

vponomaryov (changes requested):
Need to fix the breakage with undefined `target_node`.

sdcm/cluster.py
fruch: moved the line into the context manager

vponomaryov (approved):
LGTM

PR #7017, opened 2023-12-27
fix(longevity-fips): Set the longevity to run non-disruptive nemesis

roydahan:
Since this longevity uses a special clean AMI and some nemesis do not handle that well and need to install scylla, this commit suggests a workaround to run only non-disruptive nemesis. There is nothing special with the longevity for fips and in term of coverage it would be better to cover some administrative commands than topology changes or things like that.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

PR #7019, opened 2023-12-27
fix(large-partitions): disable validation part

fruch:
the validation code isn't yet fully working
and can stop/fail the nemesis thread in the middle of a test case, for now it's going to be disabled

Ref: https://github.com/scylladb/qa-tasks/issues/1578

### Testing

- [x] - no testing is needed

### Reminders


roydahan (changes requested):
First, I don't want to disable.

Second, we can just set it to False instead of commenting out.

bhalevy:
I'm good with the intent, but I don't know much about the implementation.
FWIW, it seems fine to me.

fruch:
> First, I don't want to disable.
> 
> Second, we can just set it to False instead of commenting out.
> 

O.k. you can do what you want with it

PR #7024, opened 2023-12-28
fix(longevity-2TB-*-1dis-2nondis): limit manager nemesis to on thread

fruch:
since most manager task are on the whole cluster, we should limit them to be selected only on one `NonDisruptive` thread. otherwise it would generate lots of nice of failed manager tasks

### Testing

- [x] parallel nemesis case with enterprise build - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis-test/4/ https://argus.scylladb.com/test/eedcde34-5d24-4c94-84d7-07608f49b9da/runs?additionalRuns[]=af0022c1-586b-4556-989f-f6aec800f2b2

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis.yaml
+++ b/test-cases/longevity/longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis.yaml
@@ -30,6 +30,7 @@ instance_type_runner: 'r5.2xlarge'
 
 cluster_health_check: false
 nemesis_class_name: 'DisruptiveMonkey:1 NonDisruptiveMonkey:2'
+nemesis_selector: [[],[],["!manager_operation"]]
 nemesis_interval: 30
 nemesis_during_prepare: false
 nemesis_filter_seeds: false
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4275,12 +4275,11 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
 
     def add_nemesis(self, nemesis, tester_obj):
         for nem in nemesis:
-            for _ in range(nem['num_threads']):
-                nemesis_obj = nem['nemesis'](tester_obj=tester_obj,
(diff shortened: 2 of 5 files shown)


test-cases/longevity/longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis.yaml
```diff
@@ -29,6 +29,7 @@ instance_type_loader: 'c5.2xlarge'
 
 cluster_health_check: false
 nemesis_class_name: 'DisruptiveMonkey:1 NonDisruptiveMonkey:2'
+nemesis_selector: [[],[],["!manager_operation"]]
```
roydahan, line 32: Did you test it?
I'm surprised it works in this format.
Anyway, if you test it, It's approved.

roydahan (dismissed):

test-cases/longevity/longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis.yaml
fruch: I know it works in this format

And we have several tests using it like that

And even have a unit test for those

vponomaryov: @fruch Here you define 3 nemesis selectors for 2 nemesis threads. Why?
Why not 2 nemesis selectors?

fruch: It expands to 3 here:

https://github.com/scylladb/scylla-cluster-tests/blob/53cd2d2662565307007e2de2ce4765a9bf30c831/sdcm/cluster.py#L4276

But yeah this won't work as I expected, small refactoring in needed, so each one would be able to have it's own selector

vponomaryov (approved):
LGTM

Issue #6842, opened 2023-11-22
sctool restore failed: get resource: create restore target, units and views: init units: get tombstone_gc of 5gb_sizetiered_5_2.standard1: not found


bhalevy:
I'm not sure if this is the right place to file this issue, please move it if it doesn't belong here.

In https://argus.scylladb.com/workspace?state=WyI5NjIwNzVjYi00MzQxLTQxZTUtODQ3OC0wZjNjMzBiZGNmZDYiXQ
we hit a `sctool restore` error:

```
2023-11-22 01:54:57.285: (DisruptionEvent Severity.ERROR) period_type=end event_id=76426d5b-d8c6-4441-99f6-556cce93a7e3 duration=13m17s: nemesis_name=MgmtRestore target_node=Node longevity-100gb-4h-5-4-db-node-dbdba817-10 [34.226.192.27 | 10.12.10.156] (seed: True) errors=Encounter
ed an error on sctool command: restore -c d6df5d09-c109-4062-815f-9ab10639ec82 --restore-tables --location s3:manager-backup-tests-permanent-snapshots-us-east-1  --snapshot-tag sm_20230702190638UTC: Encountered a bad command exit code!

Command: 'sudo sctool restore -c d6df5d09-c109-4062-815f-9ab10639ec82 --restore-tables --location s3:manager-backup-tests-permanent-snapshots-us-east-1  --snapshot-tag sm_20230702190638UTC'

Exit code: 1

Stdout:



Stderr:

Error: get resource: create restore target, units and views: init units: get tombstone_gc of 5gb_sizetiered_5_2.standard1: not found
Trace ID: g9weRzKuQfC-EmdX6H9V_Q (grep in scylla-manager logs)



Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/mgmt/cli.py", line 1125, in run
    res = self.manager_node.remoter.sudo(f"sctool {cmd}")
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/base.py", line 123, in sudo
    return self.run(cmd=cmd,
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 614, in run
    result = _run()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 70, in inner
    return func(*args, **kwargs)
[...]

PR #7026, opened 2023-12-28
fix(nemesis): skip the `mgmt_restore` nemesis as unstable

vponomaryov:
### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2799,7 +2799,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self._mgmt_backup(backup_specific_tables=False)
 
     def disrupt_mgmt_restore(self):
-        def get_total_scylla_partition_size():
+        raise UnsupportedNemesis("'mgmt_restore' nemesis is too unstable.")
+
+        def get_total_scylla_partition_size():  # pylint: disable=unreachable
             result = self.cluster.nodes[0].remoter.run("df -k | grep /var/lib/scylla")  # Size in KB
             free_space_size = int(result.stdout.split()[1]) / 1024 ** 2  # Converting to GB
             return free_space_size


sdcm/nemesis.py
```diff
@@ -2799,7 +2799,9 @@ def disrupt_mgmt_backup(self):
         self._mgmt_backup(backup_specific_tables=False)
 
     def disrupt_mgmt_restore(self):
-        def get_total_scylla_partition_size():
+        raise UnsupportedNemesis("'mgmt_restore' nemesis is too unstable.")
```
fruch, line 2802: can you add a reference to a SCT issue (open one if doesn't exist)

vponomaryov: @roydahan said in the Argus discussion following:
```
I'm not bothered by the mgmt_restore that fails and is a known issue
```
So, I would like to hear the answer for this question too. What is that `known issue`...

roydahan:
@mykaul & @tzach please approve this change.
It's going to drop the current coverage of mgmt_restore nemesis that creates instability in regression testing.

Someone from manager team probably needs to map all the cases this nemesis can/may fail and decide how to proceed.
For example, restoring OSS to Enterprise - should work or not and how far back? (only matching releases e.g. 5.4 to 2024.1 or also 5.2 to 2024.1?)
Restoring enterprise to enterprise and how far back.
Restoring non-encrypted to fully encrypted cluster
Restoring backup that was done in one region to cluster that runs in another region (restore works, but the reads will fail until one will understand to alter the keyspace to the correct region)
Etc.

tzach:
The problem is not just the test - we have a real regression in restore we need to fix in 5.4
The test should be fixed to capture this issue.

roydahan:
The test is what found the problem and it's one of the edge cases I mentioned above.

However, there are too many problems that the test finds with regard to all the questions above and there is one general issue about it for the test failure
(Currently assigned to @rayakurl).

rayakurl:
@roydahan - @dkropachev is already working on fixing the problem with the nemesis. You can monitor the progress in https://github.com/scylladb/scylla-cluster-tests/pull/7029.
This PR should be closed.

sdcm/nemesis.py
rayakurl: AFAIK the "known issue" is https://github.com/scylladb/scylla-cluster-tests/issues/6842. 
But @dkropachev is working on resolving it. So the PR is not needed.

dkropachev:
> @roydahan, @vponomaryov - @dkropachev is already working on fixing the problem with the nemesis. You can monitor the progress in #7029. This PR should be closed.

Let's merge https://github.com/scylladb/scylla-cluster-tests/pull/7029 instead of disabling it, it will make it work for scylla up to `5.2`.

For scylla `5.4` restore procedure is not working due to https://github.com/scylladb/scylladb/issues/16349, and probably you would want to disable it for `5.4`.

fruch:
> > @roydahan, @vponomaryov - @dkropachev is already working on fixing the problem with the nemesis. You can monitor the progress in #7029. This PR should be closed.
> 
> Let's merge https://github.com/scylladb/scylla-cluster-tests/pull/7029 instead of disabling it, it will make it work for scylla up to `5.2`.
> 
> For scylla `5.4` restore procedure is not working due to https://github.com/scylladb/scylladb/issues/16349, and probably you would want to disable it for `5.4`.

If so we do need to disable it on master
FYI it never reached the 5.2 branch, it wasn't ready when it started.

If it's broken for 5.4 and 2024.1, we should disable it until proven working

vponomaryov:
Need also to consider following bug:
- https://github.com/scylladb/scylla-cluster-tests/issues/7122

It may really affect the results based on the used mgmt snapshot for the `restore` operation.

PR #7033, opened 2023-12-28
fix(data_validation longevity-large-partition-8h): set correct value

roydahan:
It seems like a copy-paste mistake done when adding data_validation to this test. the number of rows the test writes during the prepare is 5555, then the validation is comparing the number of rows actually written to the "limit_rows_number" (very bad naming for parameter). If the value is less than this parameter, it generates an error saying: "Found missing rows for partitions".

### Testing

- [ ]

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-large-partition-8h.yaml
+++ b/test-cases/longevity/longevity-large-partition-8h.yaml
@@ -68,4 +68,4 @@ data_validation: |
   primary_key_column: "pk"
   max_partitions_in_test_table: 400
   partition_range_with_data_validation: 0-100
-  limit_rows_number: 10000
+  limit_rows_number: 5555


fruch (approved):

test-cases/longevity/longevity-large-partition-8h.yaml
```diff
@@ -68,4 +68,4 @@ data_validation: |
   primary_key_column: "pk"
   max_partitions_in_test_table: 400
   partition_range_with_data_validation: 0-100
-  limit_rows_number: 10000
+  limit_rows_number: 5555
```
yarongilor, line 71: no need for this value at all (copy-paste issue). i'll remove it in next PR.

fruch: Care to explain, why it's not needed?

yarongilor: @fruch , so it'll simply use the default (no limit). please see in documentation: https://github.com/scylladb/scylla-cluster-tests/blob/7bb11a2a3eae4b94c7b96928aaec3e6982793d10/sdcm/utils/database_query_utils.py#L46

fruch: I read it 3 times, but I still don't understand why one case or then would want to set this number.
it's name doesn't tell me anything about it's function/reason.

if if something is wrong with any of the test case in this regards, please open PRs to fix those, no need to wait for some other unrelated fixes for that.

anyhow the main issue I have with those validation that upon failure they kill the nemesis thread.

roydahan:
@yarongilor I don't know if it helped or not, but in this run we don't get this error anymore.
(maybe it wasn't able to query all partitions, but please check):
https://argus.scylladb.com/test/e65a1c4d-0dee-467c-9fff-c9ee282638cc/runs?additionalRuns[]=9556a473-fb93-4f62-a22b-7dd9886f1208

test-cases/longevity/longevity-large-partition-8h.yaml
fruch: > @fruch , don't you think this validation error is severe enough to kill nemesis thread?
> it means some partition rows are unexpectedly missing.

I don't know enough what is the logic of this check to say.
But one should prove it on every cases it was introduced.

And if it was decided to stop the test, raise a critical error, not just silently kill the nemesis thread.

> 
> As for `limit_rows_number` - the main reason Roy requested it is to avoid choking the system by running a query of, for example, 10M rows, every nemesis.

yarongilor:
> @yarongilor I don't know if it helped or not, but in this run we don't get this error anymore. (maybe it wasn't able to query all partitions, but please check): [https://argus.scylladb.com/test/e65a1c4d-0dee-467c-9fff-c9ee282638cc/runs?additionalRuns[]=9556a473-fb93-4f62-a22b-7dd9886f1208](https://argus.scylladb.com/test/e65a1c4d-0dee-467c-9fff-c9ee282638cc/runs?additionalRuns%5B%5D=9556a473-fb93-4f62-a22b-7dd9886f1208)

@roydahan , correct it ran ok. it had 2 failures and 30 passes:
```
yarongilor@yarongilor:~/Downloads/logs/last$ grep 'Validate partitions info' sct.log -c
32
yarongilor@yarongilor:~/Downloads/logs/last$ grep -c 'Failed to collect partition info' sct.log 
2
```

PR #7034, opened 2023-12-31
fix(featrues-testcases): disable manager on too small monitor nodes

fruch:
those case doesn't need manager installed, and they are fine with having smaller monitor node, that manager (scylla) can't be installed on.

### Testing

- [x] no testing is needed

### Reminders

### DIFF
--- a/test-cases/features/per-partition-limit.yaml
+++ b/test-cases/features/per-partition-limit.yaml
@@ -18,7 +18,7 @@ n_monitor_nodes: 1
 
 instance_type_db: 'i4i.large'
 instance_type_loader: 'c5.large'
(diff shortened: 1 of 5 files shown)


test-cases/features/per-partition-limit.yaml
```diff
@@ -21,3 +21,5 @@ instance_type_loader: 'c5.large'
 instance_type_monitor: 't2.small'
 
 user_prefix: 'per-part-limit'
+
```
roydahan, line 24: While at it, can you please change t2 to t3?

fruch: added a new to commit, to change all t2 to t3

Issue #6825, opened 2023-11-19
KMS encryption check is failing


fruch:
## Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

all the places we have the check are failing like that:
```
2023-11-17 04:07:00.179: (AwsKmsEvent Severity.ERROR) period_type=one-time event_id=65ccf23b-1715-4a35-97f7-6d3fba4b188a: message=Failed to check the fact of encryption (KMS) for sstables
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 4478, in _rotate_kms_key
assert all(encryption_result is True for encryption_result in encryption_results)
AssertionError
```

I some of the code making `scylla sstable` read encrypted was introduced and breaks the assumptions of this call

## Impact

validation of encryption isn't working, and we can't be  sure encryption was enabled

## How frequently does it reproduce?

it reproduce on enterprise builds and on 2024.1 builds, and 2023.1.2 and up

## Installation details

Kernel Version: 5.15.0-1045-aws
Scylla version (or git commit hash): `2023.1.2-20231001.646df23cc4b3` with build-id `367fcf1672d44f5cbddc88f946cf272e2551b85a`

Cluster size: 4 nodes (i4i.4xlarge)

Scylla Nodes used in this run:
- longevity-encrypt-at-rest-200gb-6h--db-node-ed99c513-4 (54.242.142.57 | 10.12.10.61) (shards: 14)
- longevity-encrypt-at-rest-200gb-6h--db-node-ed99c513-3 (3.89.29.181 | 10.12.9.108) (shards: 14)
- longevity-encrypt-at-rest-200gb-6h--db-node-ed99c513-2 (3.80.183.203 | 10.12.10.196) (shards: 14)
- longevity-encrypt-at-rest-200gb-6h--db-node-ed99c513-1 (54.210.80.244 | 10.12.11.111) (shards: 14)

OS / Image: `ami-08d83ace0c0d35567` (aws: undefined_region)

Test: `EaR-longevity-kms-200gb-6h-test`
Test id: `ed99c513-0cd9-4e42-b890-ac7d53735965`
Test name: `scylla-staging/fruch/EaR-longevity-kms-200gb-6h-test`
Test config file(s):

[...]

PR #7036, opened 2023-12-31
fix(nemesis): flush system_schema before doing encryption check

fruch:
To avoid those errors, when checking a newly created table:
```
error processing arguments: could not load schema via schema-tables:
std::runtime_error (Failed to find keyspace1.tmp_encrypted_table in schema tables)
```

Fixes: #6825

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/38/

### Reminders


roydahan (dismissed):

fruch:
Flush helped, but still there are issues to address here, don't yet merge

fruch:
Now all issues were addressed.

flush of the schema did helped, and we also need to take into account the fact tables are encrypted by default, so the check for clear, was wrong

fruch:
@vponomaryov I've fix those according to you comments

vponomaryov (approved):
LGTM

PR #7043, opened 2024-01-01
fix(get_any_ks_cf_list): speed it up

fruch:
recent change in c7a08adaf756d94050cc179da52060f3d5db17f0,
started using `nodetool cfstats`, and we've seen
that it being called multiple for each table, and take quite
some time.

now we first scan the names into set, and then iterate over
that set, to remove those table which doesn't have data in
them.

### Testing

- [ ] integration tests

### Reminders


soyacz (approved):
LGTM

Issue #7032, opened 2023-12-28
Avoid known (minor) "marshaling error" issue in CREATE_INDEX nemesis


roydahan:
As part of this run: https://argus.scylladb.com/test/171cec11-a4d7-4637-baf9-4258d5069eaf/runs?additionalRuns[]=2f516c16-82c9-4cc6-87e2-dc6cb9c309d3

During disrupt_create_index nemesis, we have a verification query to the index.
In this case, it chose a coulmn that is a timestamp column and query it
(https://github.com/scylladb/scylla-cluster-tests/blob/039a041cf5d02c35917c573d06ac0439eca0177d/sdcm/utils/nemesis_utils/indexes.py#L109)
It failed with: 
message="marshaling error: unable to parse date '2017-06-05 08:26:10.475000': marshaling error: Milliseconds length exceeds expected (6)"

Apparently, there is a very old issue with this marshaling_error on timestamp issue.
a good summary in this commit: https://github.com/scylladb/scylladb/commit/f5cda3cfc3850ca54e79f2621cf3d4440744e8b0

We need to find a way to discard this error in such case and fix this nemesis.

roydahan:
@fruch I think that @soyacz wrote this nemesis, let's had a way to workaround this issue.

nyh:
@roydahan I think you put the wrong link. The relevant issue is https://github.com/scylladb/scylladb/issues/16575 (it's a new issue I opened yesterday, on a problem already fixed in pull request https://github.com/scylladb/scylladb/pull/14726). I also wrote a test for it yesterday: https://github.com/nyh/scylla/commit/f7ca8dfc0932358f10c437e004ad9ea0923daf0a. 

As the issue explains, "2017-06-05 08:26:10.475000" is not allowed by Scylla because Scylla only allows **three digits** of sub-second precision for a timestamp value. This makes sense, because the timestamp officially has only millisecond resolution - although Cassandra DOES allow any precision (which is just truncated) so this timestamp would have worked in Cassandra.

The question is where you got that "2017-06-05 08:26:10.475000" in your test. If it's a constant, you should fix it to ""2017-06-05 08:26:10.475". If it is generated by an earlier output from Scylla, you must be running a version which doesn't have https://github.com/scylladb/scylladb/pull/14726 backported - perhaps 5.2?

nyh:
> @roydahan I think you put the wrong link.

[...]

PR #7057, opened 2024-01-02
fix(verify_query_by_index_works): pass values as is

fruch:
change the code to using execute parameters, and stop converting the values given from the driver to strings since in the process we are generating wrong values like for example in timestamp, extra precision digits which scylla fails on.

we should be using named placeholder as much as we can (same as we tell our users to do)

Fixes: #7032
Ref: https://python-driver.docs.scylladb.com/stable/getting-started.html#passing-parameters-to-cql-queries

### Testing

- [x] tested this small change ontop dtest code copied from this one

### Reminders


roydahan (approved):
LGTM and I saw your comment on the issue as well.
I just wonder why one did this conversion originally...

fruch:
> LGTM and I saw your comment on the issue as well. I just wonder why one did this conversion originally...

because we have very few examples in our code, using parameters as it should

soyacz (approved):
LGTM

PR #7059, opened 2024-01-02
feature(azure): block all external communication

fruch:
from now on, all resources created by SCT wouldn't be available by public network, only sct runners would be accessable, and would be used as jump server if nodes needs to be reached

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/provision/azure/network_security_group_provider.py
+++ b/sdcm/provision/azure/network_security_group_provider.py
@@ -24,15 +24,15 @@ from sdcm.utils.azure_utils import AzureService
 LOGGER = logging.getLogger(__name__)
 
 
-def rules_to_payload(rules: Iterable) -> List[Dict[str, Union[str, int]]]:
+def rules_to_payload(rules: Iterable, public_access: bool) -> List[Dict[str, Union[str, int]]]:
     """convert iterable rules to format accepted by provisioner"""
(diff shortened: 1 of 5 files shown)


soyacz (commented):
We could also try with not providing public ip at all in Azure 

also I'd rather see a field in `InstanceDefinition` instead changing `get_or_create_instances` args

fruch:
> We could also try with not providing public ip at all in Azure 

That specific per instance

> 
> also I'd rather see a field in `InstanceDefinition` instead changing `get_or_create_instances` args 

But the NSG isn't per instance, that doesn't make any sense

Also the fact it's the same NSG for all of the test, made this change do nothing, cause the SCT runner create it, and others only get it.

soyacz:
so could apply additional rule to block all with higher priority

fruch:
> so could apply additional rule to block all with higher priority

that's not a new rule, that's a new NSG, that shouldn't be applied to runner

soyacz:
> > so could apply additional rule to block all with higher priority
> 
> that's not a new rule, that's a new NSG, that shouldn't be applied to runner

I'd be opting for not giving public IP's to machines that are not meant to public

fruch:
@soyacz already implemented this one 

closing

PR #7075, opened 2024-01-08
fix(nemesis): enlarge timeout of creating view/indexes

fruch:
In practice we have lots of cases in multiple cases that creation takes more then 2h we setup initial, but still finising in less then 4h and extending the hard timeout to 8h

from the data we have now [1], of one node, there no corrlation to any metric to the time this operation might take.

[1]: https://70f106c98484448dbc4705050eb3f7e9.us-east-1.aws.found.io:9243/goto/8b9466d0-ae41-11ee-81c7-3986d18dafd5

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4732,8 +4732,8 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 raise UnsupportedNemesis(  # pylint: disable=raise-missing-from
                     "Tried to create already existing index. See log for details")
             try:
-                with adaptive_timeout(operation=Operations.CREATE_INDEX, node=self.target_node, timeout=7200):
-                    wait_for_index_to_be_built(self.target_node, ks, index_name, timeout=18000)
+                with adaptive_timeout(operation=Operations.CREATE_INDEX, node=self.target_node, timeout=14400) as timeout:
+                    wait_for_index_to_be_built(self.target_node, ks, index_name, timeout=timeout * 2)
                 verify_query_by_index_works(session, ks, cf, column)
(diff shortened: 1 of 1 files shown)


soyacz:
I briefly checked e.g. b308ed6b-2f44-467b-8f68-6f5e5f835a4a and there are 6 runs with this nemesis and only one was failing with timeout. Don't we have an Scylla issue for that? There's not much data in some tests that should raise timeout error for this.

Besides I think AdaptiveTimeout could take load more properly - e.g. instead of cpu load 5 we could take reactor load, moreover we could switch to prometheus to get average from measured period.

soyacz (approved):
Assuming we'll take care of notes above, LGTM

fruch:
> I briefly checked e.g. b308ed6b-2f44-467b-8f68-6f5e5f835a4a and there are 6 runs with this nemesis and only one was failing with timeout. Don't we have an Scylla issue for that? There's not much data in some tests that should raise timeout error for this.
> 
> Besides I think AdaptiveTimeout could take load more properly - e.g. instead of cpu load 5 we could take reactor load, moreover we could switch to prometheus to get average from measured period.

the problem with this specific operation is that a whole cluster op, not just one node.
and in parallel nemesis for example, if one node is down while view is built it can take much more time.
so we don't have all of that information in the adaptive timeout, but we can just collect it.
we do have issues for cases it did reach the hard-limit, and if it's really stuck it would reach it again.

in general building of the view is not something we can easily quantify in all situations, we do want to see it doesn't get stuck, but we still don't want to fail test for it passing after 3 hours for example, at least not at first

vponomaryov (approved):
LGTM

Issue #6992, opened 2023-12-22
The `ics-longevity-toggle-strategy-large-partitions-24h-test` CI job has loader nodes utilization imbalance leading to stress cmds failures


vponomaryov:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Running the `ics-longevity-toggle-strategy-large-partitions-24h-test` CI job which has 4 loader nodes the 2nd one has 3 crashes of stress commands:

1) 
```
Hdr memory consumption:	 45902400 bytes

 time   ops/s  rows/s errors max    99.9th 99th   95th   90th   median mean   
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x407042]

runtime stack:
time.sendTime({0x6f3900?, 0xc0008d6c00}, 0xc0001c4000?)
	/usr/local/go/src/time/sleep.go:145 +0x52

goroutine 1 [chan receive]:
github.com/scylladb/scylla-bench/pkg/results.(*TestResults).GetResultsFromThreadsAndMerge(0xc000103080)
	/go/scylla-bench-0.1.18/pkg/results/thread_results.go:57 +0x22c
github.com/scylladb/scylla-bench/pkg/results.(*TestResults).GetTotalResults(0xc000103080)
	/go/scylla-bench-0.1.18/pkg/results/thread_results.go:82 +0xcc
main.main()
	/go/scylla-bench-0.1.18/main.go:640 +0x3b1e
```

2) 
```
Hdr memory consumption:	 45902400 bytes

2023/12/19 14:24:38 [query statement="SELECT pk, ck, v FROM scylla_bench.test WHERE pk = ? AND ck >= ?  LIMIT 10 " values=[10 778] consistency=QUORUM] || retry: attempt №0, sleep for 48.372823ms
2023/12/19 14:24:50 [query statement="SELECT pk, ck, v FROM scylla_bench.test WHERE pk = ? AND ck >= ?  LIMIT 10 " values=[10 1190] consistency=QUORUM] || retry: attempt №0, sleep for 75.240727ms
fatal error: unexpected signal during runtime execution
[signal SIGSEGV: segmentation violation code=0x1 addr=0xb6 pc=0x41cfe0]

runtime stack:
runtime.throw({0x7627a0?, 0x7f215bffede0?})
	/usr/local/go/src/runtime/panic.go:992 +0x71
[...]

PR #7077, opened 2024-01-09
ci(longevity): add 1 loader to longevity-large-partition-200k_pks-4days

vponomaryov:
`scylla-bench` may use much more memory than it is expected, i.e. leak for some period of time.
And if we run too many `scylla-bench` commands on a single loader node then we may get crashes of stress commands.
So, avoid it, add one more loader node to reduce utilization of loader nodes.

Ref: #6992

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

Issue #7080, opened 2024-01-09
`hydra investigate show-monitor` doesn't work (Fedora 38, podman)


michoecho:
On a fresh Fedora 38 system, with `podman` (and `podman-docker`) installed, using SCT master, I run `hydra investigate show-monitor 0d40b784-fe2e-484b-9b2d-7e78c3cbe186` and I receive the following output:
```
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
There is scylladb/hydra:v1.56-bump_paramiko_3.4.0 in local cache, using it.
Obtaining QA SSH keys...
QA SSH keys obtained.
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
Going to run './sct.py  investigate show-monitor 0d40b784-fe2e-484b-9b2d-7e78c3cbe186'...
Emulate Docker CLI using podman. Create /etc/containers/nodocker to quiet msg.
WARN[0000] Using the Podman API service with TCP sockets is not recommended, please see `podman system service` manpage for details
Obtaining QA SSH keys...
QA SSH keys obtained.
/usr/local/lib/python3.10/site-packages/elasticsearch/connection/http_urllib3.py:209: UserWarning: Connecting to https://746f0ad652a3447d83b1572f657c67cb.us-east-1.aws.found.io:9243 using SSL with verify_certs=False is insecure.
  warnings.warn(
install-bash-completion current path: /home/fedora/scylla-cluster-tests
When Podman daemon is used we don't login
New directory created: /home/fedora/sct-results/20240109-203016-495036-investigate-show-monitor
Search monitoring stack archive files for test id 0d40b784-fe2e-484b-9b2d-7e78c3cbe186 and restoring...
Checking that docker is available...
Docker is available
Restoring monitoring stack from archive 0d40b784-fe2e-484b-9b2d-7e78c3cbe186/20240101_165351/monitor-set-0d40b784.tar.gz
Download file https://cloudius-jenkins-test.s3.amazonaws.com/0d40b784-fe2e-484b-9b2d-7e78c3cbe186/20240101_165351/monitor-set-0d40b784.tar.gz to directory /tmp/tmpashh5vo3
Downloading 0d40b784-fe2e-484b-9b2d-7e78c3cbe186/20240101_165351/monitor-set-0d40b784.tar.gz from cloudius-jenkins-test
Downloaded finished
[...]

PR #7085, opened 2024-01-10
fix(monitorstack): make it work when running hydra via podman

fruch:
* podman now defaults to short-name-mode="enforcing", and the monitor stack doesn't able to pull images when working via socket, it doesn't have tty to offer which repository to alias it, setting short-name-mode="permissive" on ~/.config/containers/registries.conf alleviate this issue

* scylla-monitoring was using DOCKER_HOST inside it's start scripts, hence breaking our ability to use it to point to podman socket, a fix was subbmitted but until then we patch those out of the scripts so older collected monitoring data can be used

Ref: https://github.com/scylladb/scylla-monitoring/pull/2149
Fixes: #7080

### Testing

- [ ]

### Reminders

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,11 @@ export AWS_PROFILE=DeveloperAccessRole
 
 # Install hydra (docker holding all requirements for running SCT)
 sudo ./install-hydra.sh
+
+# if using podman, we need to disable enforcing of short name usage, without it monitoring stack won't run from withing hydra
+echo 'unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.redhat.com", "docker.io", "quay.io"]
+short-name-mode="permissive"
+' > ~/.config/containers/registries.conf
 ```
 
 ### Run a test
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -30,6 +30,11 @@ export AWS_PROFILE=DeveloperAccessRole
 
 # Install hydra (docker holding all requirements for running SCT)
 sudo ./install-hydra.sh
+
+# if using podman, we need to disable enforcing of short name usage, without it monitoring stack won't run from withing hydra
+echo 'unqualified-search-registries = ["registry.fedoraproject.org", "registry.access.redhat.com", "docker.io", "quay.io"]
+short-name-mode="permissive"
+' > /home/fruch/.config/containers/registries.conf
```
soyacz, line 37: replace `fruch` with `~` or `$HOME`

fruch: replaced

soyacz (approved):
LGTM

PR #7087, opened 2024-01-11
fix(ComparableScyllaVersion): support old non-semver versions

fruch:
there are case we might need to use very old verison like `2022.1-rc8` which are not semver complient, and we need to align it before using it

### Testing

- [ ]

### Reminders


unit_tests/test_version_utils.py
```diff
@@ -450,6 +450,8 @@ def raise_request_error(**kwargs):
     ("5.1.1", (5, 1, 1, '', '')),
     ("5.1.0-rc1", (5, 1, 0, 'rc1', '')),
     ("5.1.0~rc1", (5, 1, 0, 'rc1', '')),
+    ("2022.1-rc8", (2022, 1, 0, 'rc8', '')),
+    ("2022.1~rc8", (2022, 1, 0, 'rc8', '')),
```
vponomaryov, line 454: The same line is added twice.

unit_tests/test_version_utils.py
```diff
@@ -461,6 +463,7 @@ def raise_request_error(**kwargs):
     ("5.2.0-dev-0.20230109.08b3a9c786d9-aarch64", (5, 2, 0, "dev-0.20230109", "08b3a9c786d9")),
     ("2024.2.0.dev.0.20231219.c7cdb16538f2.1", (2024, 2, 0, "dev-0.20231219", "c7cdb16538f2.1")),
     ("2024.1.0.rc2.0.20231218.a063c2c16185.1", (2024, 1, 0, "rc2-0.20231218", "a063c2c16185.1")),
+
```
vponomaryov, line 466: Redundant change

sdcm/utils/version_utils.py
```diff
@@ -146,7 +146,11 @@ def parse(version_string: str):
 
         # NOTE: make short scylla version like '5.2' be correct semver string
         _scylla_version_parts = _scylla_version.split('.')
-        if len(_scylla_version_parts) == 2:
+        if '-' in _scylla_version_parts[1]:
```
vponomaryov, line 149: Unit tests fail on the single-element input values such as `2023`:
```
15:01:28  >       if '-' in _scylla_version_parts[1]:
15:01:28  E       IndexError: list index out of range
```

So, need to check that second element does exist.

vponomaryov (changes requested):

unit_tests/test_version_utils.py
fruch: removed

sdcm/utils/version_utils.py
fruch: fixed

PR #7041, opened 2023-12-31
feature(syslogng): move all default filters to client side

fruch:
since we have more and more filters on the log we apply by default trying to move them into client side, and not send them out at all, and not waste time apply those filter while reading the logs

### Testing

- [x] provision test AWS/GCP/Azure
- [x] one longer longevity - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/41/

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code
- [x] k8s provision tests - at least the eks one passed
- [x]  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/15/

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevent to this change (if needed)

PR #7100, opened 2024-01-14
feature(issues): introduce SkipPerIssues

fruch:
SkipPerIssues would enable us to skip logic per and open item

```
if SkipPerIssues('scylladb/scylladb#1234', params):
   Do skip logic ...

```

* we can also keep the skip with labeling the issue with `sct-2023.1-skip` and have a fine grind contol for which version this should be skipped

* it can support a list of issues, and it won't skip if any of them are open

### Testing

- [x] https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-large-partition-200k-pks-4days-gce-test/8

### Reminders


roydahan:
I thought you don't want the code that automatically runs if the issue is closed like the require marker.
What's the solution for older releases where the issue won't be fixed for?

fruch:
> I thought you don't want the code that automatically runs if the issue is closed like the require marker.
> What's the solution for older releases where the issue won't be fixed for?

As written in the description, labels on the issue or issues

sdcm/utils/issues.py
```diff
@@ -0,0 +1,87 @@
...
+    issue_id: int | None
+
+
+class SkipPerIssues:
+    def __init__(self, issues: list[str] | str, params: SCTConfiguration | dict):
+        self.params = params
+
+        github_access = KeyStore().get_json("github_access.json")
+        auth = github.Auth.Token(token=github_access["token"])
+        self.github = github.Github(auth=auth)
```
soyacz, line 36: need clear error message if this fails due e.g. token expiration. Also, should not fail the test in such case I believe.

Btw. hopefully this token has only read permissions.

sdcm/utils/issues.py
```diff
@@ -0,0 +1,87 @@
...
+        except UnknownObjectException:
+            logging.warning("couldn't find issue: %s", issue)
+            TestFrameworkEvent(source=self.__class__.__name__,
+                               message=f"couldn't find issue: {issue}",
+                               severity=Severity.WARNING,
+                               trace=sys._getframe().f_back).publish()  # pylint: disable=protected-access
+            return None
+
+    def issues_opened(self):
+        return any(issue.state != 'closed' for issue in self.issues)
```
soyacz, line 78: when testing Scylla for release branches and issue is not backported won't this allow to Nemesis run and fail?

sdcm/utils/issues.py
```diff
@@ -0,0 +1,87 @@
...
+            # TODO raise event about it
+            logging.warning("couldn't parse issue: %s", issue)
+            TestFrameworkEvent(source=self.__class__.__name__,
+                               message=f"couldn't parse issue: {issue}",
+                               severity=Severity.WARNING,
+                               trace=sys._getframe().f_back).publish()  # pylint: disable=protected-access
+            return None
+        try:
+            return self.github.get_user(issue_parsed.user_id).get_repo(issue_parsed.repo_id).get_issue(issue_parsed.issue_id)
+        except UnknownObjectException:
```
soyacz, line 69: in case of any GH issue (e.g. unavailable) it will raise an unhandled exception failing the nemesis.
One option would be to raise "UnsupportedNemesis" to mark nemesis as skipped in that case.

soyacz (commented):
Brief review, with major concerns.

Another approach we could take is to create static code analysis that would find all the lines with issues mentioned and verify if is closed and send some report / output report which "skips" could be removed. Based on this report, manually remove issue from code (and backport when issue fix is backported)

With this approach we can properly handle issue fix backports and reduce overhead in SCT during the test. Also problems with GH won't affect us.

sdcm/utils/issues.py
fruch: This won't be used only for nemesis, so we can't assume that the exception we'll want

fruch:
> Brief review, with major concerns.
> 
> Another approach we could take is to create static code analysis that would find all the lines with issues mentioned and verify if is closed and send some report / output report which "skips" could be removed. Based on this report, manually remove issue from code (and backport when issue fix is backported)
> 
> With this approach we can properly handle issue fix backports and reduce overhead in SCT during the test. Also problems with GH won't affect us.

@roydahan ask was to minimize the manual work we need todo around those things, once they were agreed to be skip/ignored

The concern is that those would be forgotten and would keep disabled (we have plenty of examples as such)

As for run time issues with GitHub connectivity, we probably want to default to skip on those cases.

sdcm/utils/issues.py
fruch: It uses the same token as dtest is using
 
And yes should be handled the same as issue not found

sdcm/utils/issues.py
fruch: That what the labels are for, to control on which branch we disable enable, it would be ontop of the issue, and not depended on SCT code being backported or not.

Most of those are Scylla issues, and enabled/disabling something in SCT won't be dependent on backport in SCT

sdcm/utils/issues.py
```diff
@@ -0,0 +1,87 @@
...
+                    repo_id=obj.get("repo_id") or "scylladb",
+                    issue_id=int(obj.get("issue_id"))
+                )
+        return None
+
+    @lru_cache
+    def get_issue_details(self, issue):
+        issue_parsed = self.parse_issue(issue)
+        if not issue_parsed:
+            # TODO raise event about it
```
fruch, line 60: remove the TODO:, code is raising event

sct.py
```diff
@@ -959,7 +959,7 @@ def show_monitor(test_id, date_time, kill, cluster_name):
 
     if not containers:
         click.echo('Errors were found when restoring Scylla monitoring stack')
-        kill_running_monitoring_stack_services()
+        # kill_running_monitoring_stack_services()
```
fruch, line 962: need too undo

sdcm/nemesis.py
```diff
@@ -567,10 +568,10 @@ def __str__(self):
             return str(self.__class__)
 
     def _kill_scylla_daemon(self):
-        with EventsSeverityChangerFilter(new_severity=Severity.WARNING,
-                                         event_class=CassandraStressLogEvent,
-                                         regex=".*Connection reset by peer.*",
-                                         extra_time_to_expiration=30):
+        with (EventsSeverityChangerFilter(new_severity=Severity.WARNING,
```
fruch, line 571: need to undo

fruch:
@roydahan @soyacz @vponomaryov @juliayakovlev 

in this PRs there are examples of:
- skipping nemesis based an issues - issue closed nemesis would run
- skipping event filtering - filter are added as long as the issue is open (or label on the issue `sct-2023.1-skip`)

sdcm/utils/issues.py
```diff
@@ -0,0 +1,93 @@
...
+    def issues_opened(self):
+        return any(issue.state != 'closed' for issue in self.issues)
+
+    def issues_labeled(self):
+        branch_version = '.'.join(self.params.get('scylla_version').split('.')[0:2])
+        issues_labels = sum([issue.labels for issue in self.issues], [])
+
+        return any(f'sct-{branch_version}-skip' in label.name for label in issues_labels)
+
+    def __bool__(self):
```
soyacz, line 92: Conluding, if any issue is opened or it contains `sct-{branch_version}-skip` in label, then it resolves to `True`
Worth to add it to docstring so users will know how to add skip label to issue.

sdcm/utils/issues.py
```diff
@@ -0,0 +1,93 @@
...
+                               message=f"failed to get issue {issue}",
+                               severity=Severity.ERROR,
+                               exception=exc).publish()
+            return None
+
+    def issues_opened(self):
+        return any(issue.state != 'closed' for issue in self.issues)
+
+    def issues_labeled(self):
+        branch_version = '.'.join(self.params.get('scylla_version').split('.')[0:2])
```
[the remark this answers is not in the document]
fruch: yes, it's a bug, we should be using `self.params.scylla_version`, which should always be populated

but we should handle the cases it not

sct.py
fruch: removed

sdcm/nemesis.py
fruch: removed

sdcm/utils/issues.py
fruch: I'll add it

unit_tests/test_utils_issues.py
```diff
@@ -0,0 +1,37 @@
...
+    with events.get_raw_events_log().open() as events_file:
+        for line in events_file.readlines():
+            events_list.append(json.loads(line))
+
+    assert events_list[-1]['message'] == "couldn't parse issue: ABC"
+    assert events_list[-1]['severity'] == 'WARNING'
+
+
+def test_open_sources_open_issues():
+    assert SkipPerIssues(["http://github.com/scylladb/scylladb/issues/16206",
```
vponomaryov, line 24: It is fragile to depend on the status of the real issues of the alive projects.
Need either to use special project where we are not expecting the status changes or mock it.

sdcm/utils/issues.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@dataclass
+class Issue:
+    user_id: str | None
+    repo_id: str | None
+    issue_id: int | None
+
+
+class SkipPerIssues:
```
vponomaryov, line 30: This class doesn't handle the `skip`.
So, I think, it is incorrect to use that word for it.
What about `GithubIssue`?

sdcm/utils/issues.py
```diff
@@ -0,0 +1,93 @@
...
+import github
+import github.Auth
+from github.GithubException import UnknownObjectException
+
+from sdcm.keystore import KeyStore
+from sdcm.sct_config import SCTConfiguration
+from sdcm.sct_events.base import Severity
+from sdcm.sct_events.system import TestFrameworkEvent
+
+scylla_issue_pattern = re.compile(
```
vponomaryov, line 16: Why word `scylla` is used? The pattern supports any project...

`github_issue_short_pattern` ?

sdcm/utils/issues.py
```diff
@@ -0,0 +1,93 @@
...
+    r"^\s*((((?P<user_id>[\w:-]+)/)?(?P<repo_id>[\w:-]+))?#)?(?P<issue_id>\d+)\s*$", re.IGNORECASE)
+
+scylla_issue_https_pattern = re.compile(
```
vponomaryov, line 19: Same here.

`github_issue_https_pattern` ?

sdcm/utils/issues.py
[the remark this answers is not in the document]
fruch: fixed the bug, and handled the case it's not defined (it's None)

sdcm/utils/issues.py
fruch: added a docstring to the class

unit_tests/lib/events_utils.py
```diff
@@ -44,7 +45,8 @@ def setup_events_processes(cls, events_device: bool, events_main_device: bool, r
             cls.events_processes_registry_patcher.start()
         if events_device:
             start_events_device(_registry=cls.events_processes_registry)
-            enable_default_filters(SCTConfiguration())
+            with environment(SCT_CLUSTER_BACKEND='docker'):
```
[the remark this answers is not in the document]
fruch: cause now `enable_default_filters` started using that configuration, and it was not fully usable, cause backend wasn't defined

unit_tests/test_utils_issues.py
fruch: I've create special issues for it under qa-tasks, so it won't be based on real issues

sdcm/utils/issues.py
fruch: it's does hold the logic about skipping, and that is it's sole purpose.
it's not a general purpose issue with it's details.

sdcm/utils/issues.py
fruch: mostly copy-paste from dtest.

fruch: changed the name

sdcm/utils/issues.py
fruch: changed the name

unit_tests/test_events.py
```diff
@@ -421,19 +421,6 @@ def test_default_filters(self):
                                "10.12.0.219: std::runtime_error (view update generator not plugged to push updates)") \
                 .publish()
 
-            DatabaseLogEvent.RUNTIME_ERROR() \
```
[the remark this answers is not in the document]
fruch: cause this logic was introduced, ontop of the filter:
```
if sct_config.get('cluster_backend').startswith("k8s"):
```

since this log line is k8s setup only, we don't need to filter on any other setup.
 
yes it's not directly related, and also would be introduced by #7041

sdcm/utils/issues.py
```diff
@@ -0,0 +1,93 @@
...
+        try:
+            return self.github.get_user(issue_parsed.user_id).get_repo(issue_parsed.repo_id).get_issue(issue_parsed.issue_id)
+        except UnknownObjectException:
+            logging.warning("couldn't find issue: %s", issue)
+            TestFrameworkEvent(source=self.__class__.__name__,
+                               message=f"couldn't find issue: {issue}",
+                               severity=Severity.WARNING,
+                               trace=sys._getframe().f_back).publish()  # pylint: disable=protected-access
+            return None
+        except Exception as exc:  # pylint: disable=
```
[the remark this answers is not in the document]
fruch: fixed it

soyacz (approved):
LGTM

fruch:
in the change here, have one issue that can be remove, since it was removed
- https://github.com/scylladb/scylladb/issues/7909
 
@roydahan, we still have those two issues, which are closed, but not fixed on scylla ends, so this code logic doesn't cover those case, or we should add new issue for those ? reopen ? tag them in special way, to skip regardless of the version ? 

- https://github.com/scylladb/scylla-enterprise/issues/2710
- https://github.com/scylladb/scylla-enterprise/issues/3148

fruch:
@soyacz 
I've introduced a new hydra command to scan the usages, as you suggested

soyacz:
> @soyacz I've introduced a new hydra command to scan the usages, as you suggested

Nice!

soyacz (dismissed):
LGTM

unit_tests/test_utils_issues.py
```diff
@@ -0,0 +1,48 @@
...
+        ["https://github.com/scylladb/qa-tasks/issues/1614",
+         "scylladb/qa-tasks#1614"], params)
+
+
+def test_closed_issues_with_skip_tag(params):
+    params.scylla_version = '2023.1.3'
+    assert SkipPerIssues(
+        ["https://github.com/scylladb/qa-tasks/issues/1615",
+         "scylladb/qa-tasks#1615"], params)
+
```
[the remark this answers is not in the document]
fruch: I don't think the risk is that high, in the sense it would skip or not skip a filter (or a nemesis), the price of miss doing it, is not that high.

we have similar mechanism in dtest (without the tags part), and I don't think we ever run into issue with its implementation.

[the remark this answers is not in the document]
vponomaryov: We are talking about adding bunch of unit tests that cover more cases.
It's implementation must be simple.
So, argument `it will not be painful` if we have a bug in the place that can be covered by the requested unit tests doesn't sounds reasonable.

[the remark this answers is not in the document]
fruch: Added those tests,  they are part of the  integration tests

soyacz (approved):
LGTM

sdcm/nemesis.py
```diff
@@ -4281,21 +4300,24 @@ def disrupt_memory_stress(self):
...
+
+        if SkipPerIssues(['scylladb/scylla-cluster-tests#7098',
+                          'scylladb/scylla-cluster-tests#6928'], params=self.tester.params):
+            # since we might get into uncontrolled situations with this nemesis
+            # see https://github.com/scylladb/scylla-cluster-tests/issues/6928
+            raise UnsupportedNemesis("Disabled cause of https://github.com/scylladb/scylla-cluster-tests/issues/6928")
+
+        if self.target_node.distro.is_rhel_like:
+            self.target_node.install_epel()
+            self.target_node.remoter.sudo('yum install -y stress-ng')
```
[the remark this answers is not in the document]
fruch: this is not part of this PR. (i.e. just unconnected this one part)

Issue #6936, opened 2023-12-13
`end_of_quota_nemesis` nemesis fails incorrectly parsing the `mount` command


vponomaryov:
In the `longevity-mv-si-4days-test` CI job we use `instance_type_db: i4i.8xlarge` where RAID disks from several SSD get created.
And when the `end_of_quota_nemesis` nemesis gets run there we get following error:
```
2023-12-09 20:37:33,968 f:quota.py        l:34   c:sdcm.utils.quota     p:INFO  > Verifying quota is configured on the node longevity-mv-si-4d-master-db-node-1df78dff-6
2023-12-09 20:37:33,968 f:remote_base.py  l:521  c:RemoteLibSSH2CmdRunner p:DEBUG > Running command "mount | grep /var/lib/scylla | awk {'print $6'} | cut -d ',' -f 8"...
...
2023-12-09 20:37:34,056 f:base.py         l:228  c:RemoteLibSSH2CmdRunner p:DEBUG > sunit=2048  
2023-12-09 20:37:34,057 f:base.py         l:142  c:RemoteLibSSH2CmdRunner p:DEBUG > Command "mount | grep /var/lib/scylla | awk {'print $6'} | cut -d ',' -f 8" finished with status 0
2023-12-09 20:37:34,057 f:quota.py        l:38   c:sdcm.utils.quota     p:WARNING > User quota on user scylla at /var/lib/scylla is not enabled
2023-12-09 20:37:34,062 f:nemesis.py      l:5077 c:sdcm.nemesis         p:ERROR > sdcm.nemesis.SisyphusMonkey: Unhandled exception in method <bound method Nemesis.disrupt_end_of_quota_nemesis of <sdcm.nemesis.SisyphusMonkey         object at 0x7f668fdbe530>> < t:2023-12-09 20:37:34,062 f:nemesis.py      l:5077 c:sdcm.nemesis         p:ERROR > sdcm.nemesis.SisyphusMonkey: Unhandled exception in method <bound method Nemesis.disrupt_end_of_quota_nemesis of <sdcm.nemesis.SisyphusMonkey object at 0x7f668fdbe530>>
2023-12-09 20:37:34,062 f:nemesis.py      l:5077 c:sdcm.nemesis         p:ERROR > Traceback (most recent call last):
[...]

PR #7105, opened 2024-01-15
fix(end_of_quota_nemesis): parse mounts correctly

fruch:
Read all the mount options, and look in them for the expected value

Fixes: #6936

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-5gb-1h-EndOfQuotaNemesis-aws-test/3
- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-5gb-1h-EndOfQuotaNemesis-gce-test/3
- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-5gb-1h-EndOfQuotaNemesis-azure-test/4

### Reminders

### DIFF
--- a/sdcm/sct_events/group_common_events.py
+++ b/sdcm/sct_events/group_common_events.py
@@ -140,21 +140,25 @@ def ignore_disk_quota_exceeded_errors(node):
             db_event=DatabaseLogEvent.BACKTRACE,
             line="Disk quota exceeded",
             node=node,
+            extra_time_to_expiration=3600,  # one hour
         ))
         stack.enter_context(DbEventsFilter(
             db_event=DatabaseLogEvent.FILESYSTEM_ERROR,
             line="Disk quota exceeded",
             node=node,
+            extra_time_to_expiration=3600,  # one hour
         ))
         stack.enter_context(DbEventsFilter(
             db_event=DatabaseLogEvent.DATABASE_ERROR,
             line="Disk quota exceeded",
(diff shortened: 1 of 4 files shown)


fruch:
with this small change, it pass with all clouds, and with the instance types it had issues with.

sdcm/sct_events/group_common_events.py
```diff
@@ -140,21 +140,25 @@ def ignore_disk_quota_exceeded_errors(node):
             db_event=DatabaseLogEvent.BACKTRACE,
             line="Disk quota exceeded",
             node=node,
+            extra_time_to_expiration=3600,  # one hour
```
soyacz, line 143: why so long?

soyacz (commented):
one question only
LGTM

sdcm/sct_events/group_common_events.py
vponomaryov: Because we have log reading lags up to 1 hour?

vponomaryov (approved):
LGTM

sdcm/sct_events/group_common_events.py
fruch: Yes, but even that wouldn't be enough

There are lags bigger than, it's just a temp patch, that might help on short longevity

PR #7110, opened 2024-01-16
feature(sct-runner): collect runner metrics using node_exporter

fruch:
* create a new version of sct-runner with node exporter baked in
* set it up on the monitoring stack

this would help use collect data about what's going on with the sct-runner, and help us pin point issues with the test code or the need to pick bigger instances/disks from some tests.

### Testing

- [x] provision test

### Reminders


fruch:
provision test shows it's working, and we do have the sct-runner info in the monitoring.
not easy to identify it, since it doesn't show anything specific, and we don't have a reference of it's private address anywhere.

but this is a start, and now we'll be able to cross check with when we start to get the lag in log handling
next thing to have metrics of syslogng itself on each node, and on the runner, so we can eliminate and log transport issue as well.

soyacz (approved):
LGTM

PR #7111, opened 2024-01-16
fix(c-s): fix audit related consistency error filter

fruch:
b33d585488b6f0209538327bd11c631828e9211f introduce a filter that wasn't totally working, this commit fixes the regex and introduce a unittest coverage for it

### Testing

- [x] unittests only

### Reminders

### DIFF
--- a/sdcm/sct_events/setup.py
+++ b/sdcm/sct_events/setup.py
@@ -129,7 +129,7 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
     # issue that should be track https://github.com/scylladb/scylla-enterprise/issues/3148
     EventsSeverityChangerFilter(new_severity=Severity.WARNING,
                                 event_class=CassandraStressLogEvent.ConsistencyError,
-                                regex="Authentication error on host.*Cannot achieve consistency level for cl ONE").publish()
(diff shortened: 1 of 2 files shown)


sdcm/sct_events/setup.py
```diff
@@ -129,7 +129,7 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
     # issue that should be track https://github.com/scylladb/scylla-enterprise/issues/3148
     EventsSeverityChangerFilter(new_severity=Severity.WARNING,
                                 event_class=CassandraStressLogEvent.ConsistencyError,
-                                regex="Authentication error on host.*Cannot achieve consistency level for cl ONE").publish()
+                                regex=r".*Authentication error on host.*Cannot achieve consistency level for cl ONE").publish()
```
soyacz, line 132: possibly just adding "r" is enough. If we dropped first .* then it would be faster

soyacz (approved):
LGTM

juliayakovlev (approved):

sdcm/sct_events/setup.py
fruch: it wasn't enough, I've checked

PR #7120, opened 2024-01-18
fix(nemesis): disrupt_modify_table failure

juliayakovlev:
It is allowed that compaction strategy has only 'class' defined, and 'compaction_window_size' and 'compaction_window_unit' are not defined. Like:
```
compaction = {'class': 'TimeWindowCompactionStrategy'}
```

In this case the nemesis fails with error:
```
KeyError: 'compaction_window_unit'
```

Example of failed test:
https://argus.scylladb.com/test/079cddb4-4eac-4594-a2f0-8452dab5eb40/runs?additionalRuns[]=73ec8004-3881-4cbb-b37b-11c75c1c1126

It first time from 2022 that this nemesis was run

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2687,8 +2687,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                      "dttl": 86400}
 
             """
-            current_unit = settings["compaction"]["compaction_window_unit"]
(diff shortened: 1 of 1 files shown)


fruch:
I think there now validation to the parameters, that wasn't there before, probably it was backported into 2023.1

fruch (approved):
LGTM

PR #7121, opened 2024-01-18
feature(syslog-ng): collect syslog-ng stats

fruch:
so we can track rate/drop of logging, we introduce an exporter for syslog-ng.

example of the metrics we will have:
```
syslog_ng_destination_messages_dropped_total{destination="...",id="remote_sct#0",type="dst.syslog"} 0
syslog_ng_destination_messages_processed_total{destination="...",id="remote_sct#0",type="dst.syslog"} 18137
syslog_ng_destination_messages_stored_total{destination="...",id="remote_sct#0",type="dst.syslog"} 25617
syslog_ng_destination_messages_written_total{destination="...",id="remote_sct#0",type="dst.syslog"} 0
```

### Testing

- [x] aws provision test

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -53,7 +53,7 @@ from cassandra.policies import WhiteListRoundRobinPolicy, HostFilterPolicy, Roun
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
 from argus.backend.util.enums import ResourceState
-from sdcm.node_exporter_setup import NodeExporterSetup
+from sdcm.node_exporter_setup import NodeExporterSetup, SyslogNgExporterSetup
 from sdcm.db_log_reader import DbLogReader
 from sdcm.mgmt import AnyManagerCluster, ScyllaManagerError
 from sdcm.mgmt.common import get_manager_repo_from_defaults, get_manager_scylla_backend
@@ -4419,6 +4419,9 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
 
         if not self.test_config.REUSE_CLUSTER:
             node.disable_daily_triggered_services()
+            if self.params.get('logs_transport') == 'syslog-ng':
+                SyslogNgExporterSetup().install(node)
+
             nic_devname = node.get_nic_devices()[0]
             if install_scylla:
                 self._scylla_install(node)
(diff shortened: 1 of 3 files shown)


fruch:
need to handle permission to the unix socket:
```
2024-01-21T08:38:04.658+00:00 PR-provision-test-PR-7121-db-node-082b50ea-1     !INFO | syslog_ng_exporter[5898]: time="2024-01-21T08:38:04Z" level=error msg="Error scraping syslog-ng: Error connecting to syslog-ng: dial unix /var/lib/syslog-ng/syslog-ng.ctl: connect: permission denied" source="syslog_ng_exporter.go:117"
```

fruch:
this is how a new panel for processed/dropped log lines:
image

it would help us identify spikes of logging per node

sdcm/node_exporter_setup.py
```diff
@@ -42,3 +42,42 @@ def install(node: "BaseNode | None" = None, remoter: "Remoter | None" = None):
...
+
+class SyslogNgExporterSetup:  # pylint: disable=too-few-public-methods
+    @staticmethod
+    def install(node: "BaseNode | None" = None, remoter: "Remoter | None" = None):
+        assert node or remoter, "node or remoter much be pass to this function"
+        if node:
+            node.install_package('wget')
+            remoter = node.remoter
+        remoter.sudo(shell_script_cmd("""
+            wget https://github.com/brandond/syslog_ng_exporter/releases/download/0.1.0/syslog_ng_exporter
```
fruch, line 55: the only concern is. this code is not very maintained, and there's only one release of it.

soyacz: In case of problems we can write own exporter

sdcm/cluster.py
```diff
@@ -4419,6 +4419,8 @@ def node_setup(self, node: BaseNode, verbose: bool = False, timeout: int = 3600)
 
         if not self.test_config.REUSE_CLUSTER:
             node.disable_daily_triggered_services()
+            SyslogNgExporterSetup().install(node)
```
soyacz, line 4422: shouldn't we fence it with `if self.params.get('logs_transport') == 'syslog-ng':` so we don't install it for e.g. artifact tests where we have `ssh` backend?

fruch: yeah, we probably should

sdcm/node_exporter_setup.py
fruch: yeah it's not really that hard to rewrite it.
anyhow it's not a critical part, and we'll be able to live without it, if it would break for any reason.

sdcm/cluster.py

soyacz (approved):
LGTM

roydahan (approved):
Very cool!
LGTM

PR #7123, opened 2024-01-19
fix(mgmt-restore): report proper `mgmt-restore` nemesis results

vponomaryov:
When we run `mgmt_restore` nemesis we may get failed stress commands which get triggered by this nemesis.
And when it fails, nemesis doesn't catch errors and reports 'Success'.

So, fix it by checking the stress commands results.

Related: #7122

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-eks-mgmt-restore/10

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevent to this change (if needed)

PR #7124, opened 2024-01-19
fix(k8s): Always define `AutomaticTLSCertificates` with bool value in K8S

vponomaryov:
Starting with the `v1.11.0` version of the scylla-operator the TLS feature gets enabled by default.
At first, we do not need it.
At second, it makes the scylla-manager operations fail [1].

So, disable it explicitly if we do not expect it to overwrite the default setting.

[1] https://github.com/scylladb/scylla-manager/issues/3679

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-eks-mgmt-restore/10

### Reminders


vponomaryov:
https://github.com/scylladb/scylla-cluster-tests/pull/7123#issuecomment-1900746073

fruch (approved):
LGTM

PR #7125, opened 2024-01-19
feature(cleanup): added azure cleanup according to keep tag

soyacz:
We are moving cloud cleanup process to SCT codebase. At first adding Azure support that was not implemented so far.

Introduce external cleaning azure cloud script that based on `creation_time` and `keep` tag deletes VM's or whole ResourceGroup in case all machines from it shall be deleted.

refs: https://github.com/scylladb/qa-tasks/issues/1535

### Testing

- [ ]

### Reminders

### DIFF
--- a/utils/cloud_cleanup/azure/clean_azure.py
+++ b/utils/cloud_cleanup/azure/clean_azure.py
@@ -1,0 +1,178 @@
+#!/usr/bin/env python3
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2024 ScyllaDB
+
+import argparse
+import os
+import sys
+from datetime import datetime, timedelta
+from typing import Callable
+
+sys.path.append(os.path.join(os.path.dirname(__file__), '..', '..', '..'))
+
+from sdcm.utils.azure_utils import AzureService  # pylint: disable=wrong-import-position
+from sdcm.utils.log import setup_stdout_logger  # pylint: disable=wrong-import-position
+
+LOGGER = setup_stdout_logger()
+azure_service = AzureService()
+resource_client = azure_service.resource
+compute_client = azure_service.compute
+DEFAULT_KEEP_HOURS = 14
+
+
+def get_keep_hours(v_m, default=DEFAULT_KEEP_HOURS):
+    keep = v_m.tags.get('keep', "").lower() if v_m.tags else None
+    if keep == "alive":
+        return -1
+    try:
+        return int(keep)
+    except (ValueError, TypeError):
+        return default
+
+
+def get_vm_creation_time(v_m, resource_group_name):
+    try:
+        creation_time_str = v_m.tags.get('creation_time', "") if v_m.tags else ""
+        creation_time = datetime.fromisoformat(creation_time_str)
+    except ValueError:
+        LOGGER.info("Error parsing creation time tag of VM: %s", creation_time_str)
+        # set creation time to now in instance.tags so next time will be processed properly
+        creation_time = datetime.utcnow()
+        tags = v_m.tags or {}
+        tags.update({'creation_time': creation_time.isoformat(sep=" ", timespec="seconds")})
+        try:
+            compute_client.virtual_machines.begin_update(resource_group_name, v_m.name, parameters={
+                "tags": tags,
+            })
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.info(
+                "Failed to update VM tags: %s in resource group: %s with exception: %s",
+                v_m.name, resource_group_name, exc)
+    return creation_time
+
+
+def get_rg_creation_time(resource_group):
+    try:
+        creation_time_str = resource_group.tags.get('creation_time', "") if resource_group.tags else ""
+        creation_time = datetime.fromisoformat(creation_time_str)
+    except ValueError:
+        LOGGER.info("Error parsing creation time tag of RG: %s", creation_time_str)
+        creation_time = datetime.utcnow()
+        tags = resource_group.tags or {}
+        tags.update({'creation_time': creation_time.isoformat(sep=" ", timespec="seconds")})
+        resource_group.tags = tags
+        try:
+            resource_client.resource_groups.create_or_update(resource_group.name, resource_group)
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.info("Failed to update RG tags: %s with exception: %s", resource_group.name, exc)
+    return creation_time
+
+
+def get_keep_action(v_m) -> Callable:
+    return v_m.tags.get('keep_action', "stop").lower() if v_m.tags else "stop"
+
+
+def should_keep(creation_time, keep_hours):
+    if keep_hours <= 0:
+        return True
+    try:
+        keep_date = creation_time + timedelta(hours=keep_hours)
+        return datetime.utcnow() < keep_date
+    except (TypeError, ValueError) as exc:
+        LOGGER.info("error while defining if should keep: %s. Keeping.", exc)
+        return True
+
+
+def delete_virtual_machine(resource_group_name, vm_name, dry_run=False):
+    if dry_run:
+        LOGGER.info("[DRY RUN] Would delete VM: %s in resource group: %s", vm_name, resource_group_name)
+    else:
+        LOGGER.info("Deleting VM: %s in resource group: %s", vm_name, resource_group_name)
+        try:
+            compute_client.virtual_machines.begin_delete(resource_group_name, vm_name)
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.info(
+                "Failed to delete VM: %s in resource group: %s with exception: %s", vm_name, resource_group_name, exc)
+
+
+def stop_virtual_machine(resource_group_name, vm_name, dry_run=False):
+    if dry_run:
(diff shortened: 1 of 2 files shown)


fruch:
Don't we want to clean all other resources as well ?

In Azure we create a whole resource group for each test, don't we ?

fruch:
let create a pipeline for this one

soyacz:
> Don't we want to clean all other resources as well ?
> 
> In Azure we create a whole resource group for each test, don't we ?

Script cleans VM by VM if there's at least one that needs to be kept, otherwise cleans whole RG.
After keep time passes, should trigger whole RG cleanup in that case.

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,111 @@
...
+def delete_resource_group(resource_group_name, dry_run=False):
+    if dry_run:
+        print(f"[DRY RUN] Would delete resource group: {resource_group_name}")
+    else:
+        print(f"Deleting resource group: {resource_group_name}")
+        resource_client.resource_groups.begin_delete(resource_group_name)
+
+
+def clean_azure_instances(dry_run=False):
+    groups = [group for group in resource_client.resource_groups.list() if group.name.startswith('SCT-')]
```
fruch, line 76: we need to handle resources, not just SCT's

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,111 @@
...
+
+from sdcm.utils.azure_utils import AzureService
+
+azure_service = AzureService()
+resource_client = azure_service.resource
+compute_client = azure_service.compute
+
+
+def get_keep_hours(v_m, default=14):
+    keep = v_m.tags.get('keep') if v_m.tags else None
```
fruch, line 28: we need `keep_action`

vponomaryov: Do we really have something else than the `terminate` keep action?

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,111 @@
...
+                print(f"Keeping VM: {v_m.name} in resource group: {group.name}")
+                clean_group = False
+            else:
+                vms_to_delete.append(v_m.name)
+
+        if clean_group:
+            delete_resource_group(group.name, dry_run=dry_run)
+        else:
+            for vm_name in vms_to_delete:
+                delete_virtual_machine(group.name, vm_name, dry_run=dry_run)
```
vponomaryov, line 92: I think it will be better to use `try`-block here to try to delete as much as possible resources.

soyacz: currently it just triggers delete (doesn't wait for the result), but anyway could fail for some reason.
I'll add it

utils/cloud_cleanup/azure/clean_azure.py
fruch: we have definition for putting there `stop`, SCT doesn't use it
and that's the default if not defined, i.e. if someone isn't aware of the rules, it would stop it.

utils/cloud_cleanup/azure/clean_azure.py
fruch: I think we should wait for result, maybe at the end of the loop ?

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,169 @@
...
+        creation_time = datetime.fromisoformat(creation_time_str)
+    except ValueError:
+        print(f"Error parsing creation time tag of VM: {str(creation_time_str)}")
+        # set creation time to now in instance.tags so next time will be processed properly
+        creation_time = datetime.utcnow()
+        tags = v_m.tags if v_m.tags else {}
+        tags.update({'creation_time': creation_time.isoformat(sep=" ", timespec="seconds")})
+        try:
+            compute_client.virtual_machines.begin_update(resource_group_name, v_m.name, parameters={
+                "tags": tags
```
vponomaryov, line 51: If you use multi-line for such resources as dict then it is more useful to use the `comma` at the end of each element line.

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,169 @@
...
+
+def get_action(v_m) -> Callable:
+    return v_m.tags.get('keep_action', "stop") if v_m.tags else "stop"
+
+
+def should_keep(creation_time, keep_hours):
+    if keep_hours <= 0:
+        return True
+    try:
+        keep_date = creation_time + timedelta(hours=keep_hours)
```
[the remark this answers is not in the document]
soyacz: yes, that was the logic.
I changed it to `True` so will behave in the opposite way.
Typo's should not happen, as usually these are set by SCT, but if happens then might mean someone manually changed and to be on the safe side better not to terminate instance.

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def delete_virtual_machine(resource_group_name, vm_name, dry_run=False):
+    if dry_run:
+        print(f"[DRY RUN] Would delete VM: {vm_name} in resource group: {resource_group_name}")
+    else:
+        print(f"Deleting VM: {vm_name} in resource group: {resource_group_name}")
+        try:
+            compute_client.virtual_machines.begin_delete(resource_group_name, vm_name)
+        except Exception as exc:  # pylint: disable=broad-except
```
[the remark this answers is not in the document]
soyacz: next script run will do retry. Let it run for some time and we'll see if we need retries.

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+def delete_resource_group(resource_group_name, dry_run=False):
+    if dry_run:
+        print(f"[DRY RUN] Would delete resource group: {resource_group_name}")
+    else:
+        print(f"Deleting resource group: {resource_group_name}")
+        try:
+            resource_client.resource_groups.begin_delete(resource_group_name)
+        except Exception as exc:  # pylint: disable=broad-except
```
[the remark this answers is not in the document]
soyacz: next script run will do retry. Let it run for some time and we'll see if we need retries.

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,169 @@
...
+        else:
+            for action, vm_name in vms_to_process:
+                action(group.name, vm_name, dry_run=dry_run)
+
+
+if __name__ == "__main__":
+    arg_parser = argparse.ArgumentParser('azure_clean')
+    arg_parser.add_argument("--duration", type=int,
+                            help="duration to keep non-tagged instances running in hours",
+                            default=os.environ.get('DURATION', 14))
```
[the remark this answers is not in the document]
soyacz: mistake. fixed

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,169 @@
...
+        print(f"Deleting resource group: {resource_group_name}")
+        try:
+            resource_client.resource_groups.begin_delete(resource_group_name)
+        except Exception as exc:  # pylint: disable=broad-except
+            print(f"Failed to delete resource group: {resource_group_name} with exception: {exc}")
+
+
+def clean_azure_instances(dry_run=False):
+    groups = [group for group in resource_client.resource_groups.list() if (
+        group.tags.get('keep', "") != "alive" if group.tags else True)]
```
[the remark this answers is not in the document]
soyacz: good point, fixing

soyacz:
thanks @vponomaryov for review, I adapted the code to reflect it. Please check out if there's something more you see could be improved.

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,187 @@
...
+    arg_parser.add_argument("--dry-run", action=argparse.BooleanOptionalAction,
+                            help="do not stop or terminate anything",
+                            default=os.environ.get('DRY_RUN', False))
+
+    args = arg_parser.parse_args()
+
+    is_dry_run = bool(args.dry_run)
+
+    if is_dry_run:
+        LOGGER.error("dry mode on")
```
vponomaryov, line 185: Please, change the message to something like `'Dry run' mode on`.

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,187 @@
...
+def delete_resource_group(resource_group_name, dry_run=False):
+    if dry_run:
+        LOGGER.info("[DRY RUN] Would delete resource group: {resource_group_name}".format(
+            resource_group_name=resource_group_name))
+    else:
+        LOGGER.info("Deleting resource group: {resource_group_name}".format(resource_group_name=resource_group_name))
+        try:
+            resource_client.resource_groups.begin_delete(resource_group_name)
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.info("Failed to delete resource group: {resource_group_name} with exception: {exc}".format(
```
vponomaryov, line 135: I am not sure how linter doesn't complain about such lines.
But in common case a `LOGGER` line must use lazy formatting:
```
LOGGER.info("foo: %s, then %s", a, b)
```

utils/cloud_cleanup/azure/clean_azure.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def should_keep(creation_time, keep_hours):
+    if keep_hours <= 0:
+        return True
+    try:
+        keep_date = creation_time + timedelta(hours=keep_hours)
+        return datetime.utcnow() < keep_date
+    except (TypeError, ValueError):
+        return True
```
vponomaryov, line 94: Probably log a `debug` message here?

soyacz:
ok, comments adapted, job test here: https://jenkins.scylladb.com/view/QA/job/QA-tools/job/cleanup-cloud/ 
(issue with 2 failed runs with `AttributeError: 'NoneType' object has no attribute 'lower'` is fixed)

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #7128, opened 2024-01-21
fix(tier1): longevity-schema-topology-changes-12h bigger runner

fruch:
this case is running multiple nemesis threads and
it keeps showing the ssh `Bad packet length` issue

now that we are collecting metrics of runner, it shows we are getting into 100% cpu.

trying this change to see if it help with this issue

Ref: https://github.com/scylladb/scylladb/issues/14004

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/7/
- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/8/
- [ ] ~~https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/9/~~
- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/10/
- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/11/

### Reminders

### DIFF
--- a/sdcm/db_log_reader.py
+++ b/sdcm/db_log_reader.py
@@ -29,6 +29,11 @@ from sdcm.utils.common import make_threads_be_daemonic_by_default
 
 LOGGER = logging.getLogger(__name__)
 
+# we aren't going to process log line which are bigger than
+# this value, i.e. not event would be generated base on those,
+# but they would still be in the logs
+LOG_LINE_MAX_PROCESSING_SIZE = 1024 * 5
+
 
 class DbLogReader(Process):
     # pylint: disable=too-many-instance-attributes
@@ -91,6 +96,10 @@ class DbLogReader(Process):
             if self._last_log_position:
                 db_file.seek(self._last_log_position)
             for index, line in enumerate(db_file, start=self._last_line_no + 1):
+                if len(line) > LOG_LINE_MAX_PROCESSING_SIZE:
+                    # trim to avoid filling the memory when lot of long line is writen
+                    line = line[:LOG_LINE_MAX_PROCESSING_SIZE]
+
                 # Postpone processing line with no ending in case if half of line is written to the disc
                 if line[-1] == '\n' or self._skipped_end_line > 20:
                     self._skipped_end_line = 0
--- a/sdcm/sct_runner.py
+++ b/sdcm/sct_runner.py
@@ -446,8 +446,8 @@ class AwsSctRunner(SctRunner):
     BASE_IMAGE = "ami-07c2ae35d31367b3e"  # Canonical, Ubuntu, 22.04 LTS, amd64 jammy image build on 2022-12-01
     SOURCE_IMAGE_REGION = "eu-west-2"  # where the source Runner image will be created and copied to other regions
     IMAGE_BUILDER_INSTANCE_TYPE = "t3.small"
-    REGULAR_TEST_INSTANCE_TYPE = "t3.large"  # 2 vcpus, 8G, 36 CPU credits/hour
-    LONGTERM_TEST_INSTANCE_TYPE = "r5.large"  # 2 vcpus, 16G
+    REGULAR_TEST_INSTANCE_TYPE = "m7i-flex.large"  # 2 vcpus, 8G
+    LONGTERM_TEST_INSTANCE_TYPE = "m7i-flex.xlarge"  # 4 vcpus, 16G
 
     def __init__(self, region_name: str, availability_zone: str):
         super().__init__(region_name=region_name, availability_zone=availability_zone)


fruch:
the graph of memory/CPU:
image

the memory is also concerning, and we'll need to look into it.

test-cases/longevity/longevity-parallel-schema-changes-12h.yaml
```diff
@@ -18,6 +18,7 @@ n_loaders: 2
 n_monitor_nodes: 1
 
 instance_type_db: 'i4i.2xlarge'
+instance_type_runner: 'r5.2xlarge'
```
roydahan, line 21: As we discussed, I suggest trying with m7i-flex.xlarge or m6i.xlarge if the former isn't available on all regions.

sdcm/sct_runner.py
```diff
@@ -447,7 +447,7 @@ class AwsSctRunner(SctRunner):
     SOURCE_IMAGE_REGION = "eu-west-2"  # where the source Runner image will be created and copied to other regions
     IMAGE_BUILDER_INSTANCE_TYPE = "t3.small"
     REGULAR_TEST_INSTANCE_TYPE = "t3.large"  # 2 vcpus, 8G, 36 CPU credits/hour
```
vponomaryov, line 449: Shouldn't we also change the regular instance type?
Reason: We get log lags probably caused by the SCT runner overload even on short longevities...

sdcm/sct_runner.py
```diff
@@ -447,7 +447,7 @@ class AwsSctRunner(SctRunner):
...
-    LONGTERM_TEST_INSTANCE_TYPE = "r5.large"  # 2 vcpus, 16G
+    LONGTERM_TEST_INSTANCE_TYPE = "m7i-flex.xlarge"  # 4 vcpus, 16G
```
vponomaryov, line 450: Is it present on all the supported, by us, regions?

fruch: yes, I've checked it, it's available on all region we are using

sdcm/sct_runner.py
fruch: me and @roydahan discussed it.

we'll put this for couple of week see the effect on tier1, and then we'll probably update this one as well

test-cases/longevity/longevity-parallel-schema-changes-12h.yaml
fruch: changed the default to m7i-flex

fruch:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/7/

showed only 1min lag, which is something we can live with for now

but on the other hand we still have peaks of cpu usage (on all 4 cpus)
image

soyacz:
> https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/7/
> 
> showed only 1min lag, which is something we can live with for now
> 
> but on the other hand we still have peaks of cpu usage (on all 4 cpus) image

can we map these peaks to some nemesis/log burst/other event?

fruch:
> > https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-mv-si-4days-test/7/
> > showed only 1min lag, which is something we can live with for now
> > but on the other hand we still have peaks of cpu usage (on all 4 cpus) image
> 
> can we map these peaks to some nemesis/log burst/other event?

those correlate to:
- terminate_and_replace_node
- seed_decommision
- destroy_data_then_repair

and seems like in all of those point we are having reshape, that print 30K-56K long lines

```
grep -inr 'Reshape sec' ~/Documents/sct-e460579d.log  | awk '{print length}' | sort -n
```

this on it's own seems to be slowing us considerably

I think we shouldn't try matching regexes on lines bigger then 16K

sdcm/db_log_reader.py
```diff
@@ -98,6 +98,8 @@ def _read_and_publish_events(self) -> None:
                     self._skipped_end_line += 1
                     continue
                 try:
+                    if len(line) > 16384:
```
vponomaryov, line 101: Maybe it is better to have line size limit for each of the regexes?
Set 16k by default and allow to use much smaller sizes for short messages.
It will minimize the number of regex match operations a lot and replace it with fast condition checks.

fruch: it's a bit more problematic, I don't want to searches, and I don't want to print those in SCT logs
and there are the backtraces regexes, that it would need to skip as well

basically I want to ignore those big one, and not start thinking in each regexes 

now I would experiment to see which size is affecting the worst on the CPU. so I'll be able to pick the smallest number possible here.

soyacz: best would be to peek into the machine during high load and see which process takes so much. I'm not sure of the mechanism Python would take more than 1cpu due to GIL and how it works, so unless we don't match regexps in separate processes, possibly some other process takes the rest 3 cpus.

roydahan: You can completely remove backtrace detection of old format (larger than one row).
If there is such, we will open an issue to change it to one liner.

fruch: we already has such an issue, years ago, but reactor stalls weren't changed.

roydahan: Can you please point me to this issue? (I recall one, but I thought we changed everything).

You can proceed with that, I'll add my comment there that it's going to be deprecated.

fruch: I've raise few years back,
https://github.com/scylladb/scylladb/issues/5464

but not everything was changed, just a small part

roydahan: This one is closed.
Whatever wasn't fixed require new issue.
If we know that reactor stalls are still like that, please open a new issue for it.
I'll alert there that soon SCT will ignore reactor stalls backtraces and won't decode them if it's not fixed.

fruch: > best would be to peek into the machine during high load and see which process takes so much. I'm not sure of the mechanism Python would take more than 1cpu due to GIL and how it works, so unless we don't match regexps in separate processes, possibly some other process takes the rest 3 cpus.

we have a separate subprocess for each node we are tracking that doesn't the reading and regexes, so if multiple nodes generate those big lines it would hit multiple CPUs

the limit was proven to help, I'm gonna take it even to a lower number, like 5K, we don't want try match any regexes on a big prints, they won't be of any interest to us (compactio/resharding/bootstrap), we are mostly looking for errors level print, aborts, segfaults, and that should be our focus when generating events.

the rest of the code that follow a file for specific regexs, doesn't have this logs, and would chock a bit on those big line, but it usually has one or two regexes to match

fruch:
situation with skip > 16K long lines, seems to be much better:

image

there still are few small peak examined, and I would be lower the limit to 5K

fruch:
last run was much better in terms of CPU, but memory took a hit, and crashes the test:

image


I'm refactoring it a bit, moving skip to the beginning of the loop, and clear the memory immediately in case of skipping.
let's see how it would work

soyacz:
> now it looks much better:
> 
> image
> 
> so I wonder if we really need the 4 cpu machines....

looks like 2 would suffice...

roydahan:
We can reduce back to 2 and revisit if we an issue arises... 
But looks great.
Now let me check what you removed in terms of functionality :)

sdcm/db_log_reader.py
```diff
@@ -29,6 +29,11 @@
 
 LOGGER = logging.getLogger(__name__)
 
+# we aren't going to process log line which are bigger than
+# this value, i.e. not event would be generated base on those,
+# but they would still be in the logs
+LOG_LINE_MAX_PROCESSING_SIZE = 1024 * 5
```
roydahan, line 35: Do you have examples of the big ones that cross this threshold?
Will we skip them even if they are errors?

What about the big stack trace we have for coredumps, will we miss them?

sdcm/sct_runner.py
roydahan: I think you can change it from t3 anyway.
It can remain large, but change it to something more predictable than t3.

sdcm/db_log_reader.py
fruch: Yes, if it's too big, we won't look for errors in it.

We were getting them from resharding print, that list all the sstables it processed in one line.

Coredumps info isn't coming via the log, we are calling coredumpctl and getting it directly, the parts in the log are multi lines, and don't get to those sizes.

One line backtraces can be long, but not 5K long

sdcm/sct_runner.py
fruch: r5.large for all ? I don't think it really matters for the test, it's just about price

we shouldn't be maxing it out (and now we could track it happing)

sdcm/db_log_reader.py
roydahan: I'm worried that the 5K is a random size you picked.
Is there a way to grep an existing log based on line size and see which messages falls into this criteria?

sdcm/sct_runner.py
roydahan: m7i-flex.large instead of the t3 and m7i-flex.xlarge for the LONGTERM can stay.

sdcm/db_log_reader.py
fruch: I did this analysis on a 12h run, and the logs are compaction (reshape/reshard) summeries, and bootstrap that shows all the ranges.

none of those are something we match, and they don't print any errors in them.

regardless they are not gone, they would still be in the log.

the logs line we rely on, are not even close to that size.

what the worst thing that would happen ? we'll miss some silly error we gonna ignore, we'll miss a backtrace of a coredump we are catching, and someone would need to manually decode when he open a bug. (and then we tweak this number, I'm very doubtful it would ever happen)

roydahan: No, I'm afraid that the worst case we will miss critical log line that detect a corrupted sstable for example.

sdcm/db_log_reader.py
```diff
@@ -91,6 +96,10 @@ def _read_and_publish_events(self) -> None:
             if self._last_log_position:
                 db_file.seek(self._last_log_position)
             for index, line in enumerate(db_file, start=self._last_line_no + 1):
+                if len(line) > LOG_LINE_MAX_PROCESSING_SIZE:
+                    del line  # to avoid filling the memory when lot of logs is writen at the same time
+                    continue
```
soyacz, line 101: instead of `del line` and `continue`, I'd do `line = line[:LOG_LINE_MAX_PROCESSING_SIZE]`
did you try that?
So in that case we would still parse beginning of log line where most important info is stored (e.g. Scylla's log level, which can be ERROR)

sdcm/db_log_reader.py
fruch: when is the last time you've seen one of those critical logs line that is that big ?

I don't have any data to suggest there are long critical lines, is it worth doubling the number of cpus of the test machines  for ?

sdcm/db_log_reader.py
fruch: I didn't tried that

sdcm/sct_runner.py
fruch: change it

sdcm/db_log_reader.py
fruch: looks good so far:
image

@roydahan seems like @soyacz idea is working nicely as well

roydahan: Cool, so I'm in favor of that.

fruch: o.k. merging it, so we can have in the weekly tier1 runs

PR #7132, opened 2024-01-22
fix(db_log_reader): remove parsing of multi line backtrace

fruch:
parsing of those is slowing us down complicated how read log files.

it was decided to stop supporting it, and support only one-liner backtraces

Ref: https://github.com/scylladb/scylladb/issues/16922

### Testing

- [ ]

### Reminders


soyacz (dismissed):
IIRC multiline backtraces are prepended by "backtrace: " line, so we could detect that and parse next lines with this complex BACKTREACE_RE until it doesn't match.

But this requires testing and effort which I'm not sure we want to put, or do it in followup task.

So, LGTM

roydahan:
This should wait for a scylla issue if we know backtraces that are still multi-line.
Also, all backports cannot happen until (probably never) the change will reach the old branches.

fruch:
> IIRC multiline backtraces are prepended by "backtrace: " line, so we could detect that and parse next lines with this complex BACKTREACE_RE until it doesn't match.
> 
> But this requires testing and effort which I'm not sure we want to put, or do it in followup task.
> 
> So, LGTM

you are describing what this code that is going to be remove, is doing.
and it's not always working correctly, or how one might expect.
also there are case backtrace can be interleved (coming from diffrent shards)

fruch:
> This should wait for a scylla issue if we know backtraces that are still multi-line. Also, all backports cannot happen until (probably never) the change will reach the old branches.

I think it can only help in release branches, less callstack we'll identify the better, isn't it ? :)

fruch:
> > @roydahan
> > I've raised [scylladb/scylladb#16922](https://github.com/scylladb/scylladb/issues/16922)
> > and even a PR to suggest addressing it, in the same way reactor stalls were addressed
> 
> @roydahan
> 
> [scylladb/scylladb#16922](https://github.com/scylladb/scylladb/issues/16922) now is assigned to me 😕
> 
> and [scylladb/seastar#2052](https://github.com/scylladb/seastar/pull/2052) doesn't seems to go anywhere, i.e. where was that card we can pull on issues which are important to us ?

Nothing change since then...

roydahan:
You can send the PR to SCT to remove the mulit-line support.
The only problem is that we won't be able to use newer version of SCT to test an older scylla and detecting multi-line backtrace.
Unless, you adda backward compatibility.

fruch:
the change on seastar is moving, for now closing this.

for quite some time we don't have too many issue with the multiline, and we found the root cause of the slowdowns we have (long lines)

roydahan:
Moving or not moving?

PR #7134, opened 2024-01-22
fix(ignore_no_space_errors): add 30min to the ignore time

fruch:
we are seeing case that even smallest lag in the logging system of less then 5min, can cause this nemesis to generate error on those log lines, adding 30min to the ignore, should be sufficent.

bigger lag then that in the logging systeam would be addresssed on it's own

### Testing

- [x] - no test are needed

### Reminders


soyacz (approved):
LGTM

juliayakovlev (approved):

Issue #6819, opened 2023-11-17
`disrupt_disable_binary_gossip_execute_major_compaction` nemesis always fails on it's `gate closed` assertion not checking `CQL` and `UN` state as it should


vponomaryov:
Running the `disrupt_disable_binary_gossip_execute_major_compaction` nemesis against the DB node using `2023.1.2` Scylla version on K8S the following error appears all the time:

```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5080, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4973, in disrupt_disable_binary_gossip_execute_major_compaction
    assert not gate_closed_appearing, \
AssertionError: After re-enabling binary and gossip, 'gate closed' messages continue to appear
```

DB node logs indeed have the `gate closed` kind of messages for long time.
But it is the only problem I observe about it - flood of those messages:
```
INFO  2023-11-15 14:42:17,680 [shard  6] compaction - [Compact keyspace1.standard1 22988180-83c5-11ee-ad04-52268e54852b] Compacted 2 sstables to [/var/lib/scylla/data/keyspace1/standard1-ed6d804083c011ee892052258e54852b/me-48-big-Data.db:level=0]. 1GB to 1GB (~100% of original) in 16927ms = 86MB/s. ~1349632 total partitions merged to 1349466.
INFO  2023-11-15 14:49:17,084 [shard  0] cql_server_controller - CQL server stopped
I1115 14:49:25.269323       1 sidecar/probes.go:122] "readyz probe: node is not ready" Service="scylla/sct-cluster-eu-west-1-rack-1-0"
I1115 14:49:25.272549       1 sidecar/probes.go:122] "readyz probe: node is not ready" Service="scylla/sct-cluster-eu-west-1-rack-1-0"
INFO  2023-11-15 14:49:27,440 [shard  3] api - stop_gossiping
WARN  2023-11-15 14:49:27,440 [shard  0] storage_service - Stopping gossip by operator request
INFO  2023-11-15 14:49:27,440 [shard  0] gossip - My status = NORMAL
INFO  2023-11-15 14:49:27,440 [shard  0] gossip - Announcing shutdown
INFO  2023-11-15 14:49:27,446 [shard  0] storage_service - Node 10.4.9.183 state jump to normal
...
[...]

PR #7137, opened 2024-01-23
fix(nemesis): fix `disable_binary_gossip_execute_major_compaction`

vponomaryov:
Check the gossip status and CQL workability in the end of the `disrupt_disable_binary_gossip_execute_major_compaction` nemesis instead of looking for the `gate closed` message in DB logs.

Fixes: #6819

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-eks-nodetool-flush-and-reshard/26

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2747,7 +2747,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         return f'{cqlsh_cmd} {options} -e {command} {host}'
 
     def run_cqlsh(self, cmd, keyspace=None, timeout=120, verbose=True, split=False, connect_timeout=60,
-                  num_retry_on_failure=1):
+                  num_retry_on_failure=1, retry_interval=3):
         """Runs CQL command using cqlsh utility"""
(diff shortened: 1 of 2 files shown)


sdcm/cluster.py
```diff
@@ -2747,7 +2747,7 @@ def _gen_cqlsh_cmd(self, command, keyspace, timeout, connect_timeout):
         return f'{cqlsh_cmd} {options} -e {command} {host}'
 
     def run_cqlsh(self, cmd, keyspace=None, timeout=120, verbose=True, split=False, connect_timeout=60,
-                  num_retry_on_failure=1):
+                  num_retry_on_failure=1, retry_interval=3):
```
fruch, line 2750: Not sure I really like the internal retry in this function

And that this default 3s fits all the usages that have retires

vponomaryov: The only current usage is in the `set_system_auth_rf` with 3 retries. Don't see any problem there.
Plus new addition in this fixed nemesis.

fruch: o.k. doesn't warrants a change,

just bug me in the eye that it's not align to other apis we are using `@retrying` or it's friends

fruch (approved):

PR #7138, opened 2024-01-24
fix(rolling-upgrade): set raft upgrade test only from 2022.2

soyacz:
5.4 and 2023.1 already have Raft enabled by default when starting fresh as base versions. So there's no need to test it in separate job.

Fix by specifying base_version explicitly instead of automatic resolver.

### Testing

- [ ]

### Reminders


roydahan (approved):

fruch (approved):
LGTM

PR #7141, opened 2024-01-24
test(tablets): added two tablets enabled longevities

ShlomiBalalis:
Since scylla bench is required for tablets, I added two scenarios that use scylla bench: longevity-large-partition-200k-pks-4days and longevity-schema-topology-changes-12h. Also, created a folder for tablet-enabled jobs in jenkins

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -189,6 +189,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     has_steady_run: bool = False    # flag that signal that nemesis should be run with perf tests with steady run
     schema_changes: bool = False
     config_changes: bool = False
+    tablets_phase_1_supported = True  # flag that signal that the nemesis can run on a tablet-enabled cluster
     free_tier_set: bool = False     # nemesis should be run in FreeTierNemesisSet
     manager_operation: bool = False  # flag that signals that the nemesis uses scylla manager
     delete_rows: bool = False  # A flag denotes a nemesis deletes partitions/rows, generating tombstones.
@@ -455,6 +456,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             config_changes: Optional[bool] = None,
             free_tier_set: Optional[bool] = None,
             manager_operation: Optional[bool] = None,
+            tablets_phase_1_supported: Optional[bool] = None,
     ) -> List[str]:
         return self.get_list_of_methods_by_flags(
             disruptive=disruptive,
@@ -467,6 +469,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             config_changes=config_changes,
             free_tier_set=free_tier_set,
             manager_operation=manager_operation,
+            tablets_phase_1_supported=tablets_phase_1_supported,
         )
 
     def _is_it_on_kubernetes(self) -> bool:
@@ -486,6 +489,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             free_tier_set: Optional[bool] = None,
             sla: Optional[bool] = None,
             manager_operation: Optional[bool] = None,
+            tablets_phase_1_supported: Optional[bool] = None,
     ) -> List[str]:
         subclasses_list = self._get_subclasses(
             disruptive=disruptive,
@@ -499,6 +503,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             free_tier_set=free_tier_set,
             sla=sla,
             manager_operation=manager_operation,
+            tablets_phase_1_supported=tablets_phase_1_supported,
         )
         disrupt_methods_list = []
         for subclass in subclasses_list:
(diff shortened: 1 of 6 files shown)


fruch (approved):
LGTM

ShlomiBalalis:
After discussing it with @roydahan, we've concluded that since only bootstrap and decommission are currently supported when using tablets, we will use only few specific nemeses for these runs. Switching it to draft

bhalevy:
removenode support is coming up in https://github.com/scylladb/scylladb/pull/16894

ShlomiBalalis:
Updated the list.

> removenode support is coming up in [scylladb/scylladb#16894](https://github.com/scylladb/scylladb/pull/16894)

I can update the list when it is merged, but if addnode is not supported atm (which is not afaik) I don't think there will be a relevant nemesis to add.

sct.py
```diff
@@ -1434,6 +1435,9 @@ def create_test_release_jobs(branch, username, password, sct_branch, sct_repo):
         if group_name == 'raft':
             for jenkins_file in glob.glob(f'{base_path}/raft/*'):
                 server.create_pipeline_job(jenkins_file, group_name)
+        if group_name == 'tablets':
+            for jenkins_file in glob.glob(f'{base_path}/tablets/*'):
+                server.create_pipeline_job(jenkins_file, group_name)
```
roydahan, line 1440: I think that this is going to change soon with @fruch's PR.

sdcm/nemesis.py
```diff
@@ -186,6 +186,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     has_steady_run: bool = False    # flag that signal that nemesis should be run with perf tests with steady run
     schema_changes: bool = False
     config_changes: bool = False
+    tablet_phase_1_supported = False  # flag that signal that the nemesis can run on a tablet-enabled cluster
```
roydahan, line 189: I think that a better approach would be to mark those that aren't supported.
Then, you can combine filters with ! and do things like "topology_changes" and "!tablets_supported" and more.

sct.py
ShlomiBalalis: Yes, I've discussed this with him. If his will be merged first, I'll rebase (Though, I can't seem to see any existing PR about it)

fruch: this is the PR
https://github.com/scylladb/scylla-cluster-tests/pull/6866

ShlomiBalalis: Clearly, I didn't go back far enough

sdcm/nemesis.py
```diff
@@ -186,6 +186,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     has_steady_run: bool = False    # flag that signal that nemesis should be run with perf tests with steady run
     schema_changes: bool = False
     config_changes: bool = False
+    tablet_phase_1_supported = True  # flag that signal that the nemesis can run on a tablet-enabled cluster
```
fruch, line 189: there are 74 occurrences of `tablet_phase_1_supported = False` in nemesis.py,
can we change the default to False, and enable the relevant ones ?

configurations/tablets.yaml
```diff
@@ -1,3 +1,5 @@
+nemesis_class_name: 'SisyphusMonkey'
```
fruch, line 1: I'm not sure that you want to introduce those here.

this configuration file is used in performance tests, and overwriting those would break them

sdcm/nemesis.py
ShlomiBalalis: Done

sdcm/nemesis.py
ShlomiBalalis: Well, that's what roy asked to do in the comment above.
But I updated the list, now there are much less occurrences

configurations/tablets.yaml
ShlomiBalalis: ...if it's only meant to be used in performance, then it really shouldn't have been named just tablets.yaml. That's confusing.
But shouldn't we use the same nemesis class for tablets performance tests as well? Since those are the only ones that are supported by tablets in the first place? @bhalevy @roydahan @xemul

ShlomiBalalis:
> Please keep it in draft till you test it.

Added 3 scenarios:
1. 4h 100gb - to test with Cassandra Stress: Failed with https://github.com/scylladb/scylladb/issues/16846#issuecomment-1918977478

2. large partition 8h - to test with scylla bench: currently running (seems good so far)
3. The cassandra harry scenario: passed

ShlomiBalalis:
@bhalevy Please review, specifically the nemesis list. There are a few actions that I wasn't sure whether they are supported or not: specific nodetool commands (like drain, scrub) sla, cdc, auth etc.

configurations/tablets.yaml
fruch: @xemul is using it in a case we it's selecting a very specific nemesis, and this change would override it.

ShlomiBalalis: Moved my config to `configurations/tablets_supported_nemeses.yaml`, but I still use `configurations/tablets.yaml, would that be alright?

fruch:
@ShlomiBalalis please rebase and move tablets folder to `jenkins-pipelines/oss`

ShlomiBalalis:
> @ShlomiBalalis please rebase and move tablets folder to `jenkins-pipelines/oss`

Done

fruch (approved):
LGTM

PR #7146, opened 2024-01-25
fix(kms): make KMS logic be stable having multiple test keyspaces

vponomaryov:
The KMS logic we have works fine if we have just one test keyspace.
If we have multiple additional keyspaces then we may get following errors:

```
  error processing arguments: could not load schema via schema-tables: \
    std::runtime_error (Failed to find keyspace1.standard1 in schema tables)
```

Such additional keyspaces may be added by multiple nemesis such as:
- `disrupt_load_and_stream`
- `disrupt_nodetool_refresh`
- `disrupt_mgmt_restore`

So, fix it by reading 'cfstats' and picking up the keyspace with the biggest number of sstables. It is going to be the keyspace+table generated by the main load/stress commands.

### Testing

Example of the failure without this fix:
- https://argus.scylladb.com/test/bb28c47f-3f97-4357-b520-8151420427e9/runs?additionalRuns%5B%5D=1eb61c51-3a04-4f1c-9174-00b18402160c

Example of the fix:
- https://argus.scylladb.com/test/116f89d3-7b3a-4cae-b8bc-5a5cb8306acf/runs?additionalRuns[]=26922fba-846d-4145-90ad-ebb1b5141a3a
- https://argus.scylladb.com/test/105e12f0-26ba-4d7d-b945-53e472772f3f/runs?additionalRuns[]=bdda8670-739f-4687-bd21-e8fb14afac34

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4389,8 +4389,18 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
                     target_node = [node for node in db_cluster.nodes if not node.running_nemesis][0]
                     self.log.debug("Target node for 'rotate_kms_key' is %s", target_node.name)
                     with run_nemesis(node=target_node, nemesis_name="KMS encryption check"):
-                        ks_cf = db_cluster.get_non_system_ks_cf_list(db_node=target_node, filter_out_mv=True)[0]
-                        SstableUtils(db_node=target_node, ks_cf=ks_cf).check_that_sstables_are_encrypted()
+                        ks_cf_list = db_cluster.get_non_system_ks_cf_list(
+                            db_node=target_node, filter_out_table_with_counter=True, filter_out_mv=True,
+                            filter_empty_tables=True)
+                        chosen_ks_cf, chosen_ks_cf_sstables_num = 'keyspace1.standard1', 0
+                        for ks_cf in ks_cf_list:
+                            res = target_node.run_nodetool(
+                                sub_cmd='cfstats', args=ks_cf, timeout=300, retry=3, publish_event=False,
+                                warning_event_on_exception=(Failure, UnexpectedExit, Libssh2_UnexpectedExit))
+                            cf_stats = target_node._parse_cfstats(res.stdout)  # pylint: disable=protected-access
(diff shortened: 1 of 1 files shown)


fruch:
@vponomaryov but all those should be encrypted and working, once we are doing enable encrypt by default 

I don't understand how the size of the table has anything to do with our ability to use the sstable tool ?

I think we need to flush the system_schema keyspace before we start, and disable compaction on the target table and the `system_schema.tables`, the tool can't really handle file being delete from under it's feet, and if we don't flush, we don't have gurrente the schema would be on disk.

vponomaryov:
> @vponomaryov but all those should be encrypted and working, once we are doing enable encrypt by default
> 
> I don't understand how the size of the table has anything to do with our ability to use the sstable tool ?
> 
> I think we need to flush the system_schema keyspace before we start, and disable compaction on the target table and the `system_schema.tables`, the tool can't really handle file being delete from under it's feet, and if we don't flush, we don't have gurrente the schema would be on disk.

We were checking only one table in one keyspace.
And we were getting errors when some additional test keyspace was being added/deleted.
So, in this PR, I add filtering for the keyspaces assuming we pick up always the correct one - the main one which is populated by the main stress commands.

fruch:
> > @vponomaryov but all those should be encrypted and working, once we are doing enable encrypt by default
> > I don't understand how the size of the table has anything to do with our ability to use the sstable tool ?
> > I think we need to flush the system_schema keyspace before we start, and disable compaction on the target table and the `system_schema.tables`, the tool can't really handle file being delete from under it's feet, and if we don't flush, we don't have gurrente the schema would be on disk.
> 
> We were checking only one table in one keyspace. And we were getting errors when some additional test keyspace was being added/deleted. So, in this PR, I add filtering for the keyspaces assuming we pick up always the correct one - the main one which is populated by the main stress commands.

picking only the main load, doesn't guarantee you won't suffer similar issues with the main load, for example trying to reach which there compaction on the schema keyspace, and the tool would fail cause the files of the schema were deleted.

so this PR might minimize some of the failure, but you'll still have failures happening

vponomaryov:
> so this PR might minimize some of the failure, but you'll still have failures happening

This PR is about fixing the concrete problem described above - choosing keyspace+table for the encryption check.

fruch (approved):
LGTM

PR #7150, opened 2024-01-27
fix(raft): Raft always enabled

aleksbykov:
Raft feature is enabled by default starting from
    scylla 5.5 (master) and options consistent-cluster-management is
    depricated by default.
    Added to new checks for scylla version on node. If scylla version
    is >= 5.5, always enable raft, otherwise keep old logic based on
    consistent-cluster-management flag

### Testing
scylla version 5.5:
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h-no-raft/6
scylla version 5.4:
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h-no-raft/7
scylla version 5.4 with disabled raft
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h-no-raft/8

All jobs are passed

### Reminders

### DIFF
--- a/sdcm/utils/raft/__init__.py
+++ b/sdcm/utils/raft/__init__.py
@@ -9,9 +9,11 @@ from collections import namedtuple
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.filters import EventsSeverityChangerFilter
 from sdcm.sct_events import Severity
+from sdcm.utils.version_utils import ComparableScyllaVersion
 
 
 LOGGER = logging.getLogger(__name__)
+RAFT_DEFAULT_SCYLLA_VERSION = "5.5.0-dev"
 
 
 class Group0MembersNotConsistentWithTokenRingMembersException(Exception):
@@ -108,7 +110,7 @@ class RaftFeatureOperations(Protocol):
         return list(map(self.get_message_waiting_timeout, log_patterns))
 
 
-class Raft(RaftFeatureOperations):
+class RaftFeature(RaftFeatureOperations):
     TOPOLOGY_OPERATION_LOG_PATTERNS: dict[TopologyOperations, Iterable[MessagePosition]] = {
         TopologyOperations.DECOMMISSION: ABORT_DECOMMISSION_LOG_PATTERNS,
         TopologyOperations.BOOTSTRAP: ABORT_BOOTSTRAP_LOG_PATTERNS,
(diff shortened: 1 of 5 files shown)


soyacz:
how about the cases we run test for previous Scylla version (e.g reproducers) or when doing upgrade tests?

aleksbykov:
@soyacz @fruch @vponomaryov  could you review?

soyacz (commented):
still one concern about upgrade path: when consistent cluster management is not set for e.g. 5.4 and we upgrade to 5.5 where it is enabled by default, can we upgrade and rollback such a node?
also see that in `BaseNode` class we have:
```
    @cached_property
    def raft(self):
        return get_raft_mode(self)
```
so, need to invalidate cache in the case of upgrade

aleksbykov:
> still one concern about upgrade path: when consi. stent cluster management is not set for e.g. 5.4 and we upgrade to 5.5 where it is enabled by default, can we upgrade and rollback such a node? also see that in `BaseNode` class we have:
> 
> ```
>     @cached_property
>     def raft(self):
>         return get_raft_mode(self)
> ```
> 
> so, need to invalidate cache in the case of upgrade

by default the consistent cluster management is enabled on 5.4. So if this parameter is not set explicitly , it will be added automatically. About cache_property, thanks for catch it, i will fix it

soyacz:
> 
> by default the consistent cluster management is enabled on 5.4. So if this parameter is not set explicitly , it will be added automatically. About cache_property, thanks for catch it, i will fix it

so, why `RAFT_DEFAULT_SCYLLA_VERSION = "5.5"` and not `5.4`?

sdcm/utils/raft/__init__.py
```diff
@@ -9,9 +9,11 @@
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.filters import EventsSeverityChangerFilter
 from sdcm.sct_events import Severity
+from sdcm.utils.version_utils import ComparableScyllaVersion
 
 
 LOGGER = logging.getLogger(__name__)
+RAFT_DEFAULT_SCYLLA_VERSION = "5.5"
```
vponomaryov, line 16: Probably it should be `5.4`.
If not, and `5.5` is expected to be here.
Then, need to use some `dev` version of it, because when you will run a `5.5.0-dev*` version it will be smaller than `5.5` on the line 321 you add below.

aleksbykov:
> > by default the consistent cluster management is enabled on 5.4. So if this parameter is not set explicitly , it will be added automatically. About cache_property, thanks for catch it, i will fix it
> 
> so, why `RAFT_DEFAULT_SCYLLA_VERSION = "5.5"` and not `5.4`?

because 5.4 has flag consistent-cluster-management and raft still could be disabled, while 5.5-dev doesn't have the flag at all

sdcm/utils/raft/__init__.py
aleksbykov: thanks will fix

aleksbykov: fixed

soyacz (approved):
LGTM

aleksbykov:
@fruch @vponomaryov @temichus could you review?

temichus (approved):

aleksbykov:
@soyacz @fruch could you merged it?

PR #7169, opened 2024-01-30
fix(rolling-upgrades): ignore audit related log errors

fruch:
Cause audit is enabled by default in 2022.1/2, we are getting those errors after/during upgrades, and we should ignore them until scylladb/scylla-enterprise#3814 is fixed

Ref: scylladb/scylla-enterprise#3092
Ref: scylladb/scylla-enterprise#3814

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/sct_events/setup.py
+++ b/sdcm/sct_events/setup.py
@@ -131,6 +131,16 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
                                 event_class=CassandraStressLogEvent.ConsistencyError,
                                 regex=r".*Authentication error on host.*Cannot achieve consistency level for cl ONE").publish()
 
+    if sct_config.get('new_scylla_repo') or sct_config.get('new_version'):
+        # scylladb/scylla-enterprise#3814
+        # scylladb/scylla-enterprise#3092
+        # skip audit related issue when upgrading from older versions it's default in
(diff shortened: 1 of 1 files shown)


sdcm/sct_events/setup.py
```diff
@@ -131,6 +131,16 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
                                 event_class=CassandraStressLogEvent.ConsistencyError,
                                 regex=r".*Authentication error on host.*Cannot achieve consistency level for cl ONE").publish()
 
+    if sct_config.get('new_scylla_repo') or sct_config.get('new_version'):
+        # scylladb/scylla-enterprise#3814
+        # scylladb/scylla-enterprise#3092
+        # skip audit related issue when upgrading from older versions it's default in
+        # TODO: remove when branch 2022.2 is deprecated
+        EventsSeverityChangerFilter(new_severity=Severity.WARNING,
```
soyacz, line 139: why warning instead of 'debug'/'info'?

fruch: same as all of the above, that move it to warning.

I guess the initial idea was that those can be found easily if needed, but it doesn't really make any difference

juliayakovlev (approved):

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

PR #7170, opened 2024-01-30
improvement(k8s-eks): use EBS gp3 disks

vponomaryov:
Make 2 following main changes to switch all the EKS volume usages from the `gp2` to the `gp3` type:
- Explicitly specify the disk type for all the K8S nodes created as part of the K8S cluster
- Configure the `gp3` storage class for K8S PVC.

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-functional-eks/64

### Reminders


sdcm/cluster_k8s/eks.py
```diff
@@ -214,29 +214,25 @@ def launch_template_name(self) -> str:
...
-            block_devices.append(root_disk_def)
+            root_ebs_block_device_params["VolumeSize"] = self.disk_size
+        root_disk_def = LaunchTemplateBlockDeviceMappingRequestTypeDef(
+            DeviceName="/dev/xvda",
+            Ebs=LaunchTemplateEbsBlockDeviceRequestTypeDef(**root_ebs_block_device_params))
         launch_template = RequestLaunchTemplateDataTypeDef(
             KeyName=self.ssh_key_pair_name,
             EbsOptimized=False,
-            BlockDeviceMappings=block_devices,
-            NetworkInterfaces=[{
```
fruch, line 236: This isn't needed ?

vponomaryov: It is prohibited here, because security groups are specified out of the launch-template already.

fruch: So before this change we weren't using the template at all ?

Do we clean those templates ?

soyacz (approved):
LGTM

sdcm/cluster_k8s/eks.py
vponomaryov: Yes, we were not using `launch templates` before.
No, we do not clean those templates.

fruch: we should add the needed code for cleaning them up, in a followup PR

PR #7176, opened 2024-02-01
fix(commit log): add exceeding time gate

temichus:
according to issue https://github.com/scylladb/scylladb/issues/17069 cylla_commitlog_disk_total_bytes can exceed the limit during/after replay. For a short bit. added gate for 1 min per shard, per request for 10 min range  with 1 sec discreteness

### Testing

- pass run https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/longevity-10gb-3h-test/58/execution/node/163/log/

failed_ run (to test calculation logic ) https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/longevity-10gb-3h-test/59/execution/node/163/log/

### Reminders

### DIFF
--- a/sdcm/commit_log_check_thread.py
+++ b/sdcm/commit_log_check_thread.py
@@ -128,6 +128,8 @@ class CommitLogCheckThread:
     # Prometheus API limit: check_interval/discreteness must be less then 11000
     check_interval = 10 * 60
     discreteness = 1
+    # scylla_commitlog_disk_total_bytes can exceed the limit during/after replay. For a short bit.
+    exceed_time_interval = 60
 
     def __init__(self, custer_tester, test_duration, termination_event=None, thread_name: str = ""):
         self.log = logging.getLogger(self.__class__.__name__)
@@ -176,13 +178,33 @@ class CommitLogCheckThread:
                 message=f"CommitLogCheckThread failed: {exc.__repr__()} with traceback {trace}").publish()
         self.log.debug("CommitLogCheckThread finished")
 
+    @staticmethod
+    def get_commit_log_directory_max_exceed_time(prometheus_response):
+        # get max time diff from response per shard(line in response)
+        # response example:
+        # [{'metric': {'__name__': 'scylla_commitlog_disk_total_bytes',
+        #              'instance': '10.0.0.6', 'job': 'scylla', 'shard': '0'},
+        #              'values': [[1706415723.153, '54324629504'],  [1706415743.153, '54324629504']]},
+        #  {'metric': {'__name__': 'scylla_commitlog_disk_total_bytes',
+        #             'instance': '10.0.0.7', 'job': 'scylla', 'shard': '1'},
+        #             'values': [[1706415724.153, '54324629504'], [1706415725.153, '54324629504']]}
+        # ]
+        exceed_time_list = []
+        for line in prometheus_response:
+            time_list = [item[0] for item in line["values"]]
+            time_list.sort()
+            exceed_time_list.append(time_list[-1] - time_list[0])
+        exceed_time_list.sort()
+        return exceed_time_list[-1]
+
     def overflow_commit_log_directory_checker(self, start_time, end_time):
         response = self.prometheus.query(
             self.prometheus_queries.overflow_commit_log_directory, start_time, end_time, self.discreteness)
         self.log.debug("overflow_commit_log_directory: %s", response)
-        if response:
+        if response and self.get_commit_log_directory_max_exceed_time(response) > self.exceed_time_interval:
             CommitLogCheckErrorEvent(
-                message=f"commit log directory exceed the limit. Prometheus response: {response}").publish()
+                message=f"commit log directory exceed the limit longer that expected."
+                        f" Prometheus response: {response}").publish()
 
     def zero_free_segments_checker(self, start_time, end_time):
         response = self.prometheus.query(


fruch:
@temichus can you change the typo in the headline ? is should be `exceeding`

temichus:
> @temichus can you change the typo in the headline ? is should be `exceeding`

fixed

sdcm/commit_log_check_thread.py
```diff
@@ -176,13 +178,26 @@ def run_thread(self):
...
         self.log.debug("CommitLogCheckThread finished")
 
+    @staticmethod
+    def get_commit_log_directory_max_exceed_time(prometheus_response):
+        # get max time diff from response per shard(line in response)
+        exceed_time_list = []
+        for line in prometheus_response:
+            time_list = [item[0] for item in line["values"]]
+            time_list.sort()
+            print(time_list[-1] - time_list[0])
```
fruch, line 188: don't use print

sdcm/commit_log_check_thread.py
```diff
@@ -176,13 +178,26 @@ def run_thread(self):
                 message=f"CommitLogCheckThread failed: {exc.__repr__()} with traceback {trace}").publish()
         self.log.debug("CommitLogCheckThread finished")
 
+    @staticmethod
+    def get_commit_log_directory_max_exceed_time(prometheus_response):
+        # get max time diff from response per shard(line in response)
```
fruch, line 183: I would recommend putting an example of the response in the docstring, so one can understand the logic below

sdcm/commit_log_check_thread.py
```diff
@@ -176,13 +178,26 @@ def run_thread(self):
...
+        exceed_time_list = []
+        for line in prometheus_response:
+            time_list = [item[0] for item in line["values"]]
+            time_list.sort()
+            print(time_list[-1] - time_list[0])
+            exceed_time_list.append(time_list[-1] - time_list[0])
+            exceed_time_list.sort()
```
fruch, line 190: not sure why sorting it on every round ?, and not at the end when returning ?

sdcm/commit_log_check_thread.py
```diff
@@ -176,13 +178,26 @@ def run_thread(self):
                 message=f"CommitLogCheckThread failed: {exc.__repr__()} with traceback {trace}").publish()
         self.log.debug("CommitLogCheckThread finished")
 
+    @staticmethod
+    def get_commit_log_directory_max_exceed_time(prometheus_response):
+        # get max time diff from response per shard(line in response)
+        exceed_time_list = []
+        for line in prometheus_response:
+            time_list = [item[0] for item in line["values"]]
```
fruch, line 186: isn't that what you are trying to achieve ?
```python
max_time = max(item[0] -  line["values"][0][0] for item in line["values"])
```

temichus: line["values"][0][0] - it is firs item 

I'm looking for the diff between min d max value 
ideally, it should be just ```line["values"][0][-1]-line["values"][0][0]`` but I have not found info that prometheus guarantees time ordering in response, so in finding max and min value by sorting

sdcm/commit_log_check_thread.py
temichus: removed

sdcm/commit_log_check_thread.py
temichus: you are right

sdcm/commit_log_check_thread.py
temichus: added

sdcm/commit_log_check_thread.py
fruch: I'm still not sure what's the calculation done in this function, and why the assumption the time would ever be bigger  than 60sec


those time are exactly the times prometheus is querying them, and if you aren't looking at the value of the metric, it don't mean anything about what was scylla doing or not doing, you should be looking at the value as well.

elcallio:
This seems like a bit of overkill/overgenerosity v.s. the counters. The "problem" you are chasing is a.) only when doing non-clean start with CL replay happening and b.) Only on shard 0 in this case. Is there no way to make this time allowance dependent on relation to a replay sequence (check log)? Also worth noting is that unless you change sharding/limits between crash and restart, the total limit across all shards should still be held up more or less (maybe an extra segment per shard).

sdcm/commit_log_check_thread.py
temichus: i'm filtering values in request https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/commit_log_check_thread.py#L64

so Prometheus' response has only bad values.

initially, it was an assumption that if we got any bad value from Prometheus' response - produce Error event
now, according to a discussion with @elcallio , it can be bad values for a short period. 
so I added such a gate for 1 minute per shard, per request for 10 minimum range with 1 sec discreteness

temichus: according to @elcallio comment. it also needs to be limited to shart 0
```b.) Only on shard 0 in this case.```

temichus:
> This seems like a bit of overkill/overgenerosity v.s. the counters. The "problem" you are chasing is a.) only when doing non-clean start with CL replay happening and b.) Only on shard 0 in this case. Is there no way to make this time allowance dependent on relation to a replay sequence (check log)? Also worth noting is that unless you change sharding/limits between crash and restart, the total limit across all shards should still be held up more or less (maybe an extra segment per shard).

this is SCT. In this framework, a lot of stuff happens in parallel(workload, nemesis, scan_operations). I can not rely on logs 

but I can limit this check only to shart 0 

or change the logic itself to check limits not per shard but by node( total limit across all shards)

CC devs who provided initial requirements for this check
@eliransin , @wmitros

elcallio (approved):
I guess I am ok with this approach, if it solves the immediate problem. Moving forward I think it would be better to make the measurements more situation aware, i.e. restricting this to replay-case.

Issue #7122, opened 2024-01-19
Data validation stress commands in the `mgmt_restore` nemesis fail on all 5gb persistent snapshots and 3 out 4 10gb ones


vponomaryov:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

When we run `mgmt_restore` nemesis, it picks up some of the existing persistent snapshots specified in the `defaults/manager_persistent_snapshots.yaml` file:

```
aws:                                                                                                   
  bucket: "manager-backup-tests-permanent-snapshots-us-east-1"                                         
  confirmation_stress_template: "cassandra-stress read cl=QUORUM n={num_of_rows} -schema 'keyspace={keyspace_name} replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq={sequence_start}..{sequence_end}"
  snapshots_sizes:                                                                                     
    5:                                                                                                 
      number_of_rows: 5242880                                                                          
      expected_timeout: 1800  # 30 minutes                                                             
      snapshots:                                                                                       
        'sm_20230702185929UTC':                                                                        
          keyspace_name: "5gb_sizetiered_2022_2"                                                       
          scylla_version: "2022.2.9"                                                                   
          scylla_product: "enterprise"                                                                 
          number_of_nodes: 5                                                                           
        'sm_20230702201949UTC':                                                                        
          keyspace_name: "5gb_sizetiered_2022_1"                                                       
          scylla_version: "2022.1.7"                                                                   
          scylla_product: "enterprise"                                                                 
          number_of_nodes: 5                                                                           
        'sm_20230702190638UTC':                                                                        
          keyspace_name: "5gb_sizetiered_5_2"                                                          
          scylla_version: "5.2.3"                                                                      
          scylla_product: "oss"                                                                        
          number_of_nodes: 5                                                                           
    10:                                                                                                
      number_of_rows: 10485760                                                                         
      expected_timeout: 3600  # 60 minutes                                                             
      snapshots:                                                                                       
        'sm_20230223105105UTC':                                                                        
          keyspace_name: "10gb_sizetiered"                                                             
          scylla_version: "5.1.6"                                                                      
          scylla_product: "oss"                                                                        
          number_of_nodes: 3                                                                           
[...]

PR #7177, opened 2024-02-01
Fixing manager restore nemesis region issue

ShlomiBalalis:
### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -15,6 +15,7 @@ from collections import defaultdict
 from copy import deepcopy
 from dataclasses import asdict
 
+import random
 import logging
 import os
 import re
@@ -959,9 +960,7 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
             self.run_commit_log_check_thread(self.get_duration(None))
 
     def set_system_auth_rf(self, db_cluster=None):
-
-        if not db_cluster:
-            db_cluster = self.db_cluster
+        db_cluster = db_cluster or self.db_cluster
         # No need to change system tables when running via scylla-cloud
         # Also, when running a Alternator via scylla-cloud, we don't have CQL access to the cluster
         if self.params.get('db_type') == 'cloud_scylla' or self.params.get("cluster_backend") == "baremetal":
@@ -975,21 +974,25 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         if self.test_config.REUSE_CLUSTER:
             self.log.debug("Skipping change rf of system_auth for reusing cluster")
             return
-        node = self.db_cluster.nodes[0]
+        self.set_ks_strategy_to_network_and_rf_according_to_cluster(keyspace="system_auth", db_cluster=db_cluster)
+
+    def set_ks_strategy_to_network_and_rf_according_to_cluster(self, keyspace, db_cluster=None, repair_after_alter=True):
+        db_cluster = db_cluster or self.db_cluster
+        node = random.choice([node for node in self.db_cluster.nodes if not node.running_nemesis])
         nodes_by_region = self.db_cluster.nodes_by_region()
         dc_by_region = self.db_cluster.get_datacenter_name_per_region(db_cluster.nodes)
         datacenters = {}
         for region in nodes_by_region:
             datacenters.update({dc_by_region[region]: len(nodes_by_region[region])})
         self.log.debug("Number of nodes by datacenter %s", datacenters)
-        NetworkTopologyReplicationStrategy(**datacenters).apply(node, "system_auth")
-        res = node.run_cqlsh('DESC KEYSPACE system_auth',
-                             num_retry_on_failure=3)
-        self.log.debug("system_auth description: %s", res.stdout)
-        self.log.info('repair system_auth keyspace ...')
-        for node in self.db_cluster.nodes:
-            node.run_nodetool(sub_cmd="repair -pr system_auth", timeout=MINUTE_IN_SEC * 20)
-        self.log.info('repair system_auth keyspace done')
+        NetworkTopologyReplicationStrategy(**datacenters).apply(node, keyspace)
+        res = node.run_cqlsh(f'DESC KEYSPACE {keyspace}', num_retry_on_failure=3)
+        self.log.debug("%s description: %s", keyspace, res.stdout)
+        if repair_after_alter:
+            self.log.info('repair %s keyspace ...', keyspace)
+            for node in self.db_cluster.nodes:
+                node.run_nodetool(sub_cmd=f"repair -pr {keyspace}", timeout=MINUTE_IN_SEC * 20)
+            self.log.info('repair %s keyspace done', keyspace)
 
     @cache
     def pre_create_alternator_tables(self):
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -74,6 +74,7 @@ from sdcm.utils.ci_tools import get_job_name, get_job_url
 from sdcm.utils.common import format_timestamp, wait_ami_available, update_certificates, \
     download_dir_from_cloud, get_post_behavior_actions, get_testrun_status, download_encrypt_keys, rows_to_list, \
     make_threads_be_daemonic_by_default, ParallelObject, clear_out_all_exit_hooks, change_default_password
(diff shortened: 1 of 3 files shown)


ShlomiBalalis:
In run # 11 the nemesis passed several times: https://argus.scylladb.com/test/fed0b2b9-66f8-4cd7-bef0-c2603e5b4a75/runs?additionalRuns[]=b38e7583-97cb-4aac-b34e-3b337512478f
(it only ultimately failed because of spot termiation)

sdcm/tester.py
```diff
@@ -958,11 +959,7 @@ def setUp(self):  # pylint: disable=too-many-branches,too-many-statements
             self.run_commit_log_check_thread(self.get_duration(None))
 
     def set_system_auth_rf(self, db_cluster=None):
-
-        if not db_cluster:
-            db_cluster = self.db_cluster
-        # No need to change system tables when running via scylla-cloud
-        # Also, when running a Alternator via scylla-cloud, we don't have CQL access to the cluster
```
vponomaryov, line 965: Comment removal looks redundant.

sdcm/tester.py
```diff
@@ -958,11 +959,7 @@ def setUp(self):  # pylint: disable=too-many-branches,too-many-statements
...
+        db_cluster = db_cluster or self.db_cluster
```
vponomaryov, line 962: It looks redundant, because the `db_cluster` discovery gets handled by the `set_ks_rf` method.

vponomaryov (changes requested):
Please, squash 2nd and 3rd commits into a single one.

sdcm/tester.py
ShlomiBalalis: Yes, but you need it for the `if len(db_cluster.nodes) == 1` check

vponomaryov: Ok.

ShlomiBalalis:
> Please, squash 2nd and 3rd commits into a single one. Also, need to fix unit tests.

> Comment removal looks redundant.

Done and done

vponomaryov:
> > Please, squash 2nd and 3rd commits into a single one. Also, need to fix unit tests.
> 
> > Comment removal looks redundant.
> 
> Done and done

I don't see changes to unit tests.

Also, would be good to mention the https://github.com/scylladb/scylla-cluster-tests/issues/7122 in the main commit.

ShlomiBalalis:
> > > Please, squash 2nd and 3rd commits into a single one. Also, need to fix unit tests.
> > 
> > 
> > > Comment removal looks redundant.
> > 
> > 
> > Done and done
> 
> I don't see changes to unit tests.

Yeah, test_temporary_replication_strategy_setter_rolls_back_on_exit needs fixing. On it.

> Also, would be good to mention, in the main commit, the bug that gets fixed here (#7122).

I'm not sure it will completely solve it, so I didn't think it should be mentioned atm

vponomaryov:
> > Also, would be good to mention, in the main commit, the bug that gets fixed here (#7122).
> 
> I'm not sure it will completely solve it, so I didn't think it should be mentioned ATM

According to the description of the fix, it cover the root cause of the described bug.
So, I expect it to have 100% fix coverage.
Or am I missing something in this picture?

ShlomiBalalis:
> > > Also, would be good to mention, in the main commit, the bug that gets fixed here (#7122).
> > 
> > 
> > I'm not sure it will completely solve it, so I didn't think it should be mentioned ATM
> 
> According to the description of the fix, it covers the root cause of the described bug. So, I expect it to have 100% fix coverage. Or am I missing something in this picture?

I'm not sure it's the only root cause. If it was I don't think only `3 out 4 10gb ones` was a thing. That's why I planned on fixing the region issue first, and then see if it completely solves the issue.

sdcm/tester.py
```diff
@@ -974,21 +973,24 @@ def set_system_auth_rf(self, db_cluster=None):
         if self.test_config.REUSE_CLUSTER:
             self.log.debug("Skipping change rf of system_auth for reusing cluster")
             return
-        node = self.db_cluster.nodes[0]
+        self.set_ks_rf(keyspace="system_auth", db_cluster=db_cluster)
+
+    def set_ks_rf(self, keyspace, db_cluster=None):
```
fruch, line 978: Name is a bit vague.

It doesn't state to which replication factor it's gonna change and it's not in the parameters.

Also seems like this function belongs to Cluster and not Tester

sdcm/tester.py
```diff
@@ -974,21 +973,24 @@ def set_system_auth_rf(self, db_cluster=None):
...
-        self.log.debug("system_auth description: %s", res.stdout)
-        self.log.info('repair system_auth keyspace ...')
+        NetworkTopologyReplicationStrategy(**datacenters).apply(node, keyspace)
+        res = node.run_cqlsh(f'DESC KEYSPACE "{keyspace}"', num_retry_on_failure=3)
+        self.log.debug("%s description: %s", keyspace, res.stdout)
+        self.log.info('repair %s keyspace ...', keyspace)
         for node in self.db_cluster.nodes:
-            node.run_nodetool(sub_cmd="repair -pr system_auth", timeout=MINUTE_IN_SEC * 20)
-        self.log.info('repair system_auth keyspace done')
+            node.run_nodetool(sub_cmd=f"repair -pr {keyspace}", timeout=MINUTE_IN_SEC * 20)
```
fruch, line 992: I'm not sure  repair is the side effect you want in restore nemesis, isn't that part of the restore itself ?

vponomaryov:
Please, combine commit with unit tests updates and the logic change that caused the need to update it.

> > > > Also, would be good to mention, in the main commit, the bug that gets fixed here (#7122).
> > > 
> > > 
> > > I'm not sure it will completely solve it, so I didn't think it should be mentioned ATM
> > 
> > 
> > According to the description of the fix, it covers the root cause of the described bug. So, I expect it to have 100% fix coverage. Or am I missing something in this picture?
> 
> I'm not sure it's the only root cause. If it was I don't think only `3 out 4 10gb ones` was a thing. That's why I planned on fixing the region issue first, and then see if it completely solves the issue.

You may use `Ref: #foo`

sdcm/utils/replication_strategy_utils.py
```diff
@@ -25,7 +25,7 @@ def get(cls, node: BaseNode, keyspace: str):
         return ReplicationStrategy.from_string(create_ks_statement)
 
     def apply(self, node: BaseNode, keyspace: str):
-        cql = f"ALTER KEYSPACE {keyspace} WITH replication = {self}"
+        cql = f'ALTER KEYSPACE "{keyspace}" WITH replication = {self}'
```
fruch, line 28: I suggest running it via cql_qoute function, which is intended exactly for those cases

fruch: Then you won't need to change all the unit tests

sdcm/tester.py
ShlomiBalalis: > Name is a bit vague.

> It doesn't state to which replication factor it's gonna change and it's not in the parameters.

I don't disagree, I was just compiling with the previously written method

sdcm/tester.py
ShlomiBalalis: It happens automatically after a restore, but you do need to repair a keyspace after altering it, do you not?
Also, I don't see why that would be an issue. It's not like we're repairing after restoring the data, only the schema.

fruch: for system_auth we are doing tha to make sure the auth data is on all nodes.

in this case isn't restore procedure puts date into each node ? i.e. if doing repair is part of the instructions for restore ? 
what the end user is going to do, and again where is the documentation for this flow, of data restored into different DC layout

fruch:
@ShlomiBalalis 2 last standing issues:
- https://github.com/scylladb/scylla-cluster-tests/pull/7177#issuecomment-1921894603
- https://github.com/scylladb/scylla-cluster-tests/pull/7177#discussion_r1477312997

sdcm/tester.py
ShlomiBalalis: Added an option to not repair after the alteration, and used it in my function call.
Also opened an issue for the docs to be amended: https://github.com/scylladb/scylla-manager/issues/3719

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #7179, opened 2024-02-02
test(perf): add perf throughput bisection script draft

soyacz:
Bisecting manually with multiple jenkins jobs is not effective due to difference in instances performance.

This commit adds bisect script that updates binaries and reruns the test on the same machines, avoiding flakiness due hardware change. Still it's a draft for further improvement when needed.

### Testing

- [ ]

### Reminders

### DIFF
--- a/utils/bisect_throughput.py
+++ b/utils/bisect_throughput.py
@@ -1,0 +1,219 @@
+#!/usr/bin/env python3
+
+# pylint: disable=all
+# pip3 install paramiko requests
+# TODO: fix upgrade if fails to start scylla - for some builds it freezes, in that case abort upgrade and go to next version,
+# also it is prohibited to do downgrade sometimes, so best would be to bootstrap cluster from scratch
+# TODO: clean up the code, split to smaller functions (with one purpose), add comments, refactor it
+
+import re
+import threading
+import time
+
+import paramiko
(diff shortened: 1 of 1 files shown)


utils/bisect_throughput.py
```diff
@@ -0,0 +1,219 @@
+#!/usr/bin/env python3
+
+# pylint: disable=all
+# pip3 install paramiko requests
```
fruch, line 4: it it's depends on SCT environment, it would have both.

utils/bisect_throughput.py
fruch, line 1: it it proven useful  ? have you been able to find a version start the regression ?

soyacz: When I tried this, I started a job with last-known-good version and got poor results and changing version didn't further decrease throughput - I was having similar results. So it is actually confirming that version does not affect throughput... 
at least for OSS version.


usefulness currently is limited because replace binaries doesn't work always, 2 things I noticed so far:
1. some binaries are just not working, need to handle that
2. bisection jump back and forth - while downgrade is not supported and Scylla does not start if detects it

soyacz:
closing in favor of: https://github.com/scylladb/scylla-cluster-tests/pull/7196

PR #7180, opened 2024-02-05
fix(k8s-eks): make S3 access really work on multi-tenant setups

vponomaryov:
Create separate unique AWS IAM role for each of the K8S service accounts used by Scylla clusters we create.
It will enable the possibility to use scylla manager in each of the Scylla tenants.

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster_k8s/eks.py
+++ b/sdcm/cluster_k8s/eks.py
@@ -556,14 +556,14 @@ class EksCluster(KubernetesCluster, EksClusterCleanupMixin):  # pylint: disable=
     def get_ec2_instance_by_id(self, instance_id):
         return boto3.resource('ec2', region_name=self.region_name).Instance(id=instance_id)
 
-    def create_iamserviceaccount_for_s3_access(self):
+    def create_iamserviceaccount_for_s3_access(self, db_cluster_name: str, namespace: str = SCYLLA_NAMESPACE):
         tags = ",".join([f"{key}={value}" for key, value in self.tags.items()])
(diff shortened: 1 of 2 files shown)


vponomaryov:
Being tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-3h-multitenant-eks/67/consoleFull
Already see that location check work.
Just backuping is very slow having small load on tenants.

@fruch from above observations I can say that probably we may want to provide scylla-manager service more resources.

vponomaryov:
Mgmt backups passed on both tenants after reducing the data size: https://argus.scylladb.com/test/626c2bab-1e46-4353-b92a-c7d8714e2495/runs?additionalRuns[]=40c5eac8-97ab-4615-a3e6-4753bd6b6657

So, this change really makes the S3 access working for our EKS deployments.

vponomaryov:
@fruch , @soyacz 

Any objections?

soyacz:
> @fruch , @soyacz
> 
> Any objections?

Is this only for multitenant setup?

vponomaryov:
> > @fruch , @soyacz
> > Any objections?
> 
> Is this only for multitenant setup?

It is for all EKS setups.
Just original issue was related only to multitenant setups.
With further `fixes` I broke even single-tenant setups.
This change is what should have been done originally with enablement of S3 service for EKS.

fruch (approved):
LGTM

soyacz (approved):
LGTM

PR #7185, opened 2024-02-08
improvement(upgrade): retry on count from truncated table

juliayakovlev:
Issue: https://github.com/scylladb/scylla-enterprise/issues/3804 
Workaround for this issue - try to retry query count of rows after upgrade/rollback if it failed and check if it solves the issue.

Another change - session timeout.
Now it is defined:
`session.default_timeout = 60.0 * 5`

But actually the query failed after 60 sec.
I added timeout in the query and want to check if it will have an impact. May be we have a problem with `session.default_timeout `

### Testing

- [ ]

### Reminders

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -27,12 +27,14 @@ import contextlib
 
 from argus.client.sct.types import Package
 from cassandra import ConsistencyLevel
+from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
 from sdcm import wait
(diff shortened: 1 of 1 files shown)


fruch (dismissed):
LGTM

fruch (dismissed):
LGTM

roydahan:
Looks good and the reproducer has passed.
If we have more queries like that we should add there retries.

roydahan (approved):

PR #7188, opened 2024-02-09
feature(sdcm/reporting): Initial tooling version reporting

k0machi:
This commit adds a helper class and an interface to report versions of
various tools used in SCT test runs. Currently, the example
implementation includes support for reporting scylla's python-driver
version to both log and argus. Each class receives a node and
(optionally) an argus client and then user calls .report() method to
collect the data (tool implementation specific, for other tools the plan
is to use node.remoter to get specific versions and report them as
'node.name/tool_version' to both log and argus.

Task: scylladb/qa-tasks#1543

### Testing

- [x] [Jenkins](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/alexey/job/alexey-argus-testing/64/)
- [x] [Argus](https://argus.scylladb.com/test/b0b22734-758e-47b4-b6ec-29e64d212d52/runs?additionalRuns[]=a71a7de0-5848-4c05-8fa8-99231d5ce817)

### Reminders

### DIFF
--- a/sdcm/reporting/tooling_reporter.py
+++ b/sdcm/reporting/tooling_reporter.py
@@ -1,0 +1,87 @@
+import logging
+
+from cassandra import __version__ as PYTHON_DRIVER_VERSION
+from argus.client.sct.client import ArgusSCTClient
+from argus.client.sct.types import Package
+
+from sdcm.remote.base import CommandRunner
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ToolReporterBase():
+
+    TOOL_NAME = None
+
+    def __init__(self, runner: CommandRunner, command_prefix: str = None, argus_client: ArgusSCTClient = None) -> None:
+        self.runner = runner
+        self.command_prefix = command_prefix
+        self.argus_client = argus_client
+        self.version: str | None = None
+
+    def __str__(self) -> str:
+        return f"{self.__class__.__name__}()"
+
+    def _report_to_log(self) -> None:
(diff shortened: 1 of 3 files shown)


k0machi:
Seems to be working:

image

k0machi:
Added support for cassandra-stress, also reworked what the class receives - seems to work. One question is do we always rely on `cassandra-stress` being in PATH? or do we sometimes specify absolute path to it?

image

fruch (approved):
LGTM

sdcm/reporting/tooling_reporter.py
```diff
@@ -0,0 +1,87 @@
...
+class CassandraStressVersionReporter(ToolReporterBase):
+    # pylint: disable=too-few-public-methods
+    TOOL_NAME = "cassandra-stress"
+
+    def _collect_version_info(self) -> None:
+        output = self.runner.run(f"{self.command_prefix} {self.TOOL_NAME} version")
+        LOGGER.info("%s: Collected cassandra-stress output:\n%s", self, output.stdout)
+        field_map = {
+            "Version": "cassandra-stress",
+            "scylla-java-driver": "scylla-java-driver",
```
soyacz, line 73: @k0machi did you test c-s with c* driver? This possibly won't work with that.

fruch: The last time we used Cassandra one was years ago....


The actual question we missed here, how it would work with versions that don't have the version command changes.

PR #7189, opened 2024-02-11
improvement(upgrade): retry query execute

juliayakovlev:
Issue: https://github.com/scylladb/scylla-enterprise/issues/3820 Workaround for this issue - try to retry queries if it failed

### Testing

- [ ] https://jenkins.scylladb.com/job/enterprise-2024.1/job/reproducers/job/rolling-upgrade-azure-image-test/2/

### Reminders

### DIFF
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -26,6 +26,7 @@ from collections import OrderedDict
 from uuid import UUID
 
 from cassandra import InvalidRequest
+from cassandra.query import SimpleStatement, FETCH_SIZE_UNSET  # pylint: disable=no-name-in-module
 from cassandra.util import sortedset, SortedSet  # pylint: disable=no-name-in-module
 from cassandra import ConsistencyLevel
 from cassandra.protocol import ProtocolException  # pylint: disable=no-name-in-module
@@ -40,6 +41,7 @@ from sdcm.utils.version_utils import ComparableScyllaVersion
 LOGGER = logging.getLogger(__name__)
 
 
+# pylint: disable=too-many-public-methods
 class FillDatabaseData(ClusterTester):
     """
     Fill scylla with many types of records, tables and data types (taken from dtest) originally by Andrei.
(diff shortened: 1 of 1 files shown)


sdcm/fill_db_data.py
```diff
@@ -3309,8 +3306,13 @@ def run_db_queries(self, session, default_fetch_size):
                             item["cdc_tables"][cdc_table] = self.get_cdc_log_rows(session, cdc_table)
                             LOGGER.debug(item["cdc_tables"][cdc_table])
 
+    @staticmethod
+    @retrying()
+    def run_query_with_retry(session, query):
```
fruch, line 3311: I would add `**kwargs`, so we can add things like timeout and such to the `execute`

sdcm/fill_db_data.py
```diff
@@ -3309,8 +3306,13 @@ def run_db_queries(self, session, default_fetch_size):
                             item["cdc_tables"][cdc_table] = self.get_cdc_log_rows(session, cdc_table)
                             LOGGER.debug(item["cdc_tables"][cdc_table])
 
+    @staticmethod
+    @retrying()
```
fruch, line 3310: I would add the exact number of retries here, so it would clearer to read.

sdcm/fill_db_data.py
```diff
@@ -3237,29 +3235,28 @@ def _run_db_queries(self, item, session):
         for i in range(len(item['queries'])):
             try:
                 if item['queries'][i].startswith("#SORTED"):
-                    res = session.execute(item['queries'][i].replace('#SORTED', ''))
+                    res = self.run_query_with_retry(session=session, query=item['queries'][i].replace('#SORTED', ''))
```
fruch, line 3238: I'm not sure we want retries for all of the `_run_db_queries`, we don't usually have issues with those, and if they won't work at first shot, we do want to know about them

sdcm/fill_db_data.py
juliayakovlev: done

juliayakovlev: done

sdcm/fill_db_data.py
juliayakovlev: done

sdcm/fill_db_data.py
juliayakovlev: done

roydahan: @fruch its a good question if we want to know about them or not.
Originally we wanted and reported some of them, but looks like the approach is to ask first for retries.
We can ask in one of the existing issues or in a new issue or in Slack engineering what should be the default behavior.

fruch: We properly want to fix the driver and have support of exponential backoffs, and set it per session 

https://github.com/scylladb/python-driver/issues/91

kbr-scylla:
But this may mask a real issue, I think we need to investigate https://github.com/scylladb/scylla-enterprise/issues/3820 to find root cause first (sorry it's taking so long, we're mostly focusing on delivering raft-topology now)

PR #7195, opened 2024-02-13
fix(create_test_release_jobs): reconfig_job shouldn't drop data

fruch:
the introduction of reconfiguring jenkins job, had a side effect of dropping the pipeline arguments from the job configuration. since the template we were using is mostly empty, and assume the the first run of the job would update the pipeline parameters (and thier defaults).

so now the code that the configuration, and merge it with the suggest template from the code, so the needed part that should be updated would be updated as expected.

### Testing

- [ ]

### Reminders

### DIFF
--- a/utils/build_system/create_test_release_jobs.py
+++ b/utils/build_system/create_test_release_jobs.py
@@ -14,6 +14,8 @@
 import os
 import logging
 from pathlib import Path
+from xml.etree import ElementTree as etree
+from copy import deepcopy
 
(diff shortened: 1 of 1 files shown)


soyacz (approved):
LGTM

fruch:
> I'm merging this one, otherwise all of the trigger jobs would start failing on maste/enterprise.

look like just saying didn't got it to happen.... 😭

PR #7196, opened 2024-02-13
improvement(bisection): add bisect decorator

soyacz:
Added bisect decorator that adds bisection functionality to given test. Used by wrapping a test case (e.g. test_write) and by provide `bisect_start_date` and optionally `bisect_end_date` (if not provided, most recent build is used as last).

Idea of the decorator is to first run a test case which should set value for `self.bisect_ref_value` which will be used for evaluation for other Scylla versions to determine regression. Then using bisection algorithm the decorator will be updating Scylla binaries, cleaning up all data and reruning test which should update `self.bisect_result_value`. When test finishes, evaluates if regression happened by comparing `self.bisect_ref_value` >= `self.bisect_result_value`.

Binaries are taken from repo files - currently works only with ubuntu/debian only.
Bisection resolution depends on builds availabilty at downloads.scylladb.com.

Potentially `bisect_test` decorator can be easily reused for any other test.

### Testing

- [x] - tested in https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/bisect-scylla-master-perf-regression-throughput-non-shard-aware-i4i-test/16/ which didn't find regression in perf test

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -389,6 +389,8 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
 
         time.sleep(0.5)
         InfoEvent(message=f"TEST_START test_id={self.test_config.test_id()}").publish()
+        self.bisect_ref_value = None
+        self.bisect_result_value = None
 
     def _init_test_duration(self):
         self._stress_duration: int = self.params.get('stress_duration')
--- a/docs/bisecting-with-sct.md
+++ b/docs/bisecting-with-sct.md
@@ -1,0 +1,50 @@
+# Bisecting with SCT
+There was an effort to make it easier to bisect with SCT (performance tests). The following is a guide on how to do it.
+currently works with ubuntu/debian only.
+
+# steps
+## modify test
+1. decorate the test you want to bisect with the `@bisect_test` decorator (from `from sdcm.utils.bisect_test import bisect_test`)
+2. Decorator will add two class variables to the test class:
+    * `bisect_result_value` - defines current test result value as comparable number
+    * `bisect_ref_value` - define reference test result value as comparable number
+3. Make the test to set `bisect_ref_value` only for the first test run
+4. Each run should also update `bisect_result_value` with the result value for further comparison
+3. Upon test end, decorator will compare `bisect_result_value` and `bisect_ref_value` - if result value is greater than reference value, will mark current
+Scylla version as 'good', otherwise as 'bad'
+4. bisect decorator will erase data, reinstall Scylla to next bisected version (based on result) and rerun the test
+
+## run test
+Run test with providing `bisect_start_date` and `bisect_end_date` (optionally) sct config variables.
+Also adapt `test_duration` to larger value (e.g. 8x current duration).
+Based on these dates, decorator will
+find available packages and start bisecting process.
+Verify logs to see last good and bad Scylla versions.
+
+# example
+```python
+from sdcm.utils.bisect_test import bisect_test
+
+@bisect_test
+def test_write(self):
+    base_cmd_w = self.params.get('stress_cmd_w')
+    stress_multiplier = self.params.get('stress_multiplier')
+    if stress_multiplier_w := self.params.get("stress_multiplier_w"):
+        stress_multiplier = stress_multiplier_w
+
+    self.create_test_stats(doc_id_with_timestamp=True)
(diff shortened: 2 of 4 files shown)


sdcm/cluster.py
```diff
@@ -4598,6 +4599,9 @@ def _scylla_install(self, node):
             if ":" in scylla_repo.rsplit("/", 1)[-1]:
                 scylla_repo, version = scylla_repo.rsplit(":", 1)
             node.install_scylla(scylla_repo=scylla_repo, scylla_version=version)
+        # remove dependency when using scylla ami but installing from repo
```
fruch, line 4602: I would rather put this kind of thing under some flag, for this case, and not to be done after every installation

sdcm/tester.py
```diff
@@ -678,6 +680,7 @@ def configure_ldap(self, node, use_ssl=False):
                                           user=ldap_username, password=LDAP_PASSWORD, ldap_entry=user_entry)
 
     def _init_test_timeout_thread(self) -> threading.Timer:
+        self.start_time = time.time()
```
fruch, line 683: why moving this one ? it's used in more places then just this function....

unit_tests/test_regression_finder.py
fruch, line 1: you want to unittest something ?, i.e. don't add an empty file...

sdcm/cluster.py
```diff
@@ -1734,7 +1734,7 @@ def install_package(self,
             pkg_cmd = 'zypper'
             package_name = f"{package_name}-{package_version}" if package_version else package_name
         else:
-            pkg_cmd = 'apt-get'
+            pkg_cmd = 'DEBIAN_FRONTEND=noninteractive apt-get -o Dpkg::Options::="--force-confdef" -o Dpkg::Options::="--force-confold"'
```
fruch, line 1737: I would need more context why we are doing this here, cause we need to reinstall things by force ?

soyacz: when installing previous/later versions of scylla, scylla.yaml changes - this makes it to stay not touched. 
Need to do it automatically, installation hangs.

sdcm/tester.py
soyacz: yes, but _init_tets_timeout_thread is used during test initialization, so it's updated anyway.
Resetting start_time is important as test is restarted, otherwise timeout would occur.
From the other hand, anyway need to adjust test_duration in config file as jenkins would kill test anyway.

So I'd need to fix also 'get_test_duration' to make it fully working.

sdcm/cluster.py
soyacz: I didn't consider it harmful, so for sake of simplicity put it this way.
Do you find it important?

fruch: it changes something in the installation, if we are going to do it alway, why it's there from the first place.
if it has some purpose we shouldn't be removing it by default.

so I don't know if it's important or not, and rather not find out it is, and our upgrade test might be missing things that might have to users, cause of we remove it.

I rather not take that risk.

sdcm/tester.py
fruch: I would skip that change, and change the test_duration for 12h as a starter.

and if we really would need such a case that can constantly run and extend it's timeout, we'll need to design it and consider how to implement, now we don't need such a thing.

sdcm/cluster.py
fruch: the noninteractive part, I'm o.k., the rest of the options, I'm not sure it would fit all of the installation we are doing with function....

I would extend it, and pass the options only from that one or two places it's needed.

so we won't need to test all of the other usages with this command.

performance_regression_test.py
```diff
@@ -500,6 +501,7 @@ def _scylla_bench_prepare_table(self):
             """)
 
     # Base Tests
+    @bisect_test
```
fruch, line 504: remove this, and move it to docs

soyacz: removed

fruch (dismissed):
LGTM

fruch:
@soyacz 

Now it has conflicts

soyacz:
> @soyacz
> 
> Now it has conflicts

fixed

fruch (approved):
LGTM

soyacz:
Conflicts resolved

fruch (approved):
LGTM

Issue #7191, opened 2024-02-12
Replace i3.* with i4i.* instances (for example - i3.large -> i4i.large)


mykaul:
The price is almost the same, the performance is substantially better. See https://instances.vantage.sh/?filter=i4i.large|i3.large&compare_on=true&selected=i4i.large,i3.large for comparison.

mykaul:
Example - https://github.com/scylladb/scylla-cluster-tests/blob/ff98c60bfbcfda893971557bf1890432090dac15/unit_tests/test_config.py#L86

mykaul:
Same for c5.* and c6i (or c6a) - see https://instances.vantage.sh/?filter=c5.large|c6i.large&compare_on=true&selected=c5.large,c6i.large and https://github.com/scylladb/scylla-cluster-tests/blob/ff98c60bfbcfda893971557bf1890432090dac15/test-cases/cdc/cdc-replication-longevity.yaml#L13

fruch:
> Example - https://github.com/scylladb/scylla-cluster-tests/blob/ff98c60bfbcfda893971557bf1890432090dac15/unit_tests/test_config.py#L86

This a unittest, it not code that really provision a VM in AWS

We already did a sweep and replaced most of the i3, with exception of few tests that would need to be recalibrated for example SLA tests.

And we have very few performance tests that need to be calibrated with i4

mykaul:
> > Example - https://github.com/scylladb/scylla-cluster-tests/blob/ff98c60bfbcfda893971557bf1890432090dac15/unit_tests/test_config.py#L86
> 
> This a unittest, it not code that really provision a VM in AWS

Yep, saw that - but want to ensure folks don't copy-paste from it.

> 
> We already did a sweep and replaced most of the i3, with exception of few tests that would need to be recalibrated for example SLA tests.
> 
> And we have very few performance tests that need to be calibrated with i4

Yes, I don't think we should touch them without calibration.

fruch:
> > > Example - https://github.com/scylladb/scylla-cluster-tests/blob/ff98c60bfbcfda893971557bf1890432090dac15/unit_tests/test_config.py#L86
> > 
> > 
> > This a unittest, it not code that really provision a VM in AWS
> 
> Yep, saw that - but want to ensure folks don't copy-paste from it.
> 
> > We already did a sweep and replaced most of the i3, with exception of few tests that would need to be recalibrated for example SLA tests.
[...]

PR #7198, opened 2024-02-14
Change instance types to use newer, mostly same cost, EC2 instances

mykaul:
Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/7191

### Testing
UNTESTED - need to run most of the test to see they still work - and well.

I did not change SLA/WLP tests, as I assume they need further tweaking for this change.

### Reminders


soyacz:
why not switching c5 to c7i?

test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
```diff
@@ -30,4 +30,4 @@ gemini_table_options:
 stress_cdclog_reader_cmd: "cdc-stressor -duration 215m -stream-query-round-duration 30s"
 
 db_type: mixed_scylla
-instance_type_db_oracle: 'i3.8xlarge'
+instance_type_db_oracle: 'i4i.8xlarge'
```
fruch, line 33: this was a problem previously, since we had oracle versions which don't support i4i (2021.1, I think)
this needs testing

mykaul:
> why not switching c5 to c7i?

1. Spot and on-demand cost a bit more than c6i - see https://instances.vantage.sh/?filter=c5.large|c6i.large|c7i.large&compare_on=true&selected=c5.large,c6i.large,c7i.large 
2. Since it's newer, I kinda assumed there are less spot instances of it.

test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
mykaul: Of course. Everything needs testing.
Partially related - we should upgrade oracle at some point...

fruch: @roydahan update it last in #6988

we probably need a better definition when those need to change, and not only when things breaks (or AMIs get deleted)

roydahan:
There are 2 main considerations need to be taken here:
1. Some of these need to be tested, changing the loader can in theory push more load to the cluster and it may cause tests to fail - assuming this is somehow tested before merge is WRONG.
2. There were some reservations that were made by Karin based on types and generations, some of them for us, some from customers who churned - we need to check the current reservations and do the comparisons.

test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
mykaul: I think we should update it on every major Scylla release. We continue to fix bugs, even in data consistency. I don't know if the Oracle has cache disabled, for example (I hope!), but we've had some rare bugs there - with RF=1 exposing them, etc.

mykaul:
> There are 2 main considerations need to be taken here:
> 
> 1. Some of these need to be tested, changing the loader can in theory push more load to the cluster and it may cause tests to fail - assuming this is somehow tested before merge is WRONG.

So how can we easily test those? Perhaps split this big PR to multiple, per topic?

> 2. There were some reservations that were made by Karin based on types and generations, some of them for us, some from customers who churned - we need to check the current reservations and do the comparisons.

Who can follow up on this?
The main motivation is cost reduction and improved efficiency.

test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
roydahan: It is using the following in defaults to disable cache:
append_scylla_args_oracle: '--enable-cache false'

The assumption is that old releases are good and tested and working for customers and new releases may regress (which is mostly what happens).

Anyway, i4i should work for every release we have here.

roydahan:
> > 2. There were some reservations that were made by Karin based on types and generations, some of them for us, some from customers who churned - we need to check the current reservations and do the comparisons.
> 
> Who can follow up on this? The main motivation is cost reduction and improved efficiency.

I can followup on this.
I understand the motivation, but if we are using reserved instances, it will achieve the opposite.

mykaul:
Checked with Karin - no worries

roydahan:
> Checked with Karin - no worries

Yes, I also checked with her on Thursday, she said that all already expired.

roydahan:
I think that the only place that this change may affect anything is longevities, but I doubt it will affect too much on stress.
@fruch your call if to merge as is and see how it goes in master weekly or to require some testing.

fruch:
> I think that the only place that this change may affect anything is longevities, but I doubt it will affect too much on stress. @fruch your call if to merge as is and see how it goes in master weekly or to require some testing.

I'm o.k. with all but the gemini oracles, they would need testing on it's own, I don't thing the version we are using supports i4i, we already reverted this kind of change.

roydahan:
> > I think that the only place that this change may affect anything is longevities, but I doubt it will affect too much on stress. @fruch your call if to merge as is and see how it goes in master weekly or to require some testing.
> 
> I'm o.k. with all but the gemini oracles, they would need testing on it's own, I don't thing the version we are using supports i4i, we already reverted this kind of change.

You're right, 2021.1 doesn't support i4i.
I don't want to change that until we clear the status if Gemini found data issue or not.

mykaul:
I'll remove oracle infra changes then.

roydahan (approved):

roydahan:
@fruch any idea what fails the checks?
I couldn't understand it from the Jenkins output.

fruch:
> > @fruch any idea what fails the checks? I couldn't understand it from the Jenkins output.
> 
> Commit name violates expected commit structure.
> 
> It is now: `Change instance types to use newer, mostly same cost, EC2 instances`.
> 
> Should be like: `improvement(aws): Use newer instance types of mostly the same cost`

I've fixed the commit title

fruch (approved):
LGTM

PR #7200, opened 2024-02-14
feature(tablets): add trigger for one tablets case in weekly runs

fruch:
this introduce one pipeline across both OSS/enterprise that would enable tablets on the latency during grow shirnk case

### Testing

- [ ]

### Reminders

### DIFF
--- a/jenkins-pipelines/performance/scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink-tablets.jenkinsfile
+++ b/jenkins-pipelines/performance/scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink-tablets.jenkinsfile
@@ -1,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+perfRegressionParallelPipeline(
+    backend: "aws",
+    availability_zone: 'a,b,c',
+    test_name: "performance_regression_test.PerformanceRegressionTest",
+    test_config: """["test-cases/performance/perf-regression-latency-650gb-grow-shrink.yaml", "configurations/tablets.yaml"]""",
+    sub_tests: ["test_latency_write_with_nemesis", "test_latency_read_with_nemesis", "test_latency_mixed_with_nemesis"],
+    test_email_title: "latency during grow-shrink (tablets)",
+)
--- a/jenkins-pipelines/performance/scylla-master/weekly-enterprise-performance-trigger.xml
+++ b/jenkins-pipelines/performance/scylla-master/weekly-enterprise-performance-trigger.xml
@@ -30,7 +30,7 @@ provision_type=on_demand</properties>
               <textParamValueOnNewLine>false</textParamValueOnNewLine>
             </hudson.plugins.parameterizedtrigger.PredefinedBuildParameters>
           </configs>
-          <projects>../scylla-enterprise/scylla-enterprise-perf-regression-latency-shard-aware-1TB-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-with-nemesis,../scylla-enterprise/scylla-enterprise-perf-regression-throughput-non-shard-aware-i4i-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink</projects>
+          <projects>../scylla-enterprise/scylla-enterprise-perf-regression-latency-shard-aware-1TB-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-with-nemesis,../scylla-enterprise/scylla-enterprise-perf-regression-throughput-non-shard-aware-i4i-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink-tablets</projects>
           <condition>SUCCESS</condition>
           <triggerWithNoParameters>false</triggerWithNoParameters>
           <triggerFromChildProjects>false</triggerFromChildProjects>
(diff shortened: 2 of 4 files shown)


jenkins-pipelines/performance/scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink-tablets.jenkinsfile
```diff
@@ -0,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+perfRegressionParallelPipeline(
+    backend: "aws",
+    availability_zone: 'a,b,c',
```
soyacz, line 8: I just spotted we set this test to be multi-az, while jenkins pipeline resets it anyway to 1 zone.
So if this test meant to be multi az, it is not triggered this way.

fruch: it was meant to be multi-az, jenkins pipeline should be using what's written here as default
or that our trigger is limiting it...

roydahan: Do they want it to run in multi-AZ?
Is our job that runs in branch-perf-v14 runs in multi-AZ as well?

roydahan: Why do we have this job duplicated for master and for enterprise?
Perf jobs shouldn't be duplicated so we can compare releases in one email.

jenkins-pipelines/performance/scylla-master/weekly-enterprise-performance-trigger.xml
```diff
@@ -30,7 +30,7 @@ provision_type=on_demand</properties>
               <textParamValueOnNewLine>false</textParamValueOnNewLine>
             </hudson.plugins.parameterizedtrigger.PredefinedBuildParameters>
           </configs>
-          <projects>../scylla-enterprise/scylla-enterprise-perf-regression-latency-shard-aware-1TB-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-with-nemesis,../scylla-enterprise/scylla-enterprise-perf-regression-throughput-non-shard-aware-i4i-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink</projects>
+          <projects>../scylla-enterprise/scylla-enterprise-perf-regression-latency-shard-aware-1TB-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-with-nemesis,../scylla-enterprise/scylla-enterprise-perf-regression-throughput-non-shard-aware-i4i-test,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink,../scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink-tablets</projects>
```
roydahan, line 33: I don't understand why we have these jobs

jenkins-pipelines/performance/scylla-enterprise/scylla-enterprise-perf-regression-latency-650gb-grow-shrink-tablets.jenkinsfile
fruch: again part of the refactoring, I want each directory to match the directory we are using 
at the beginning I've used symlinks, but jenkins is stupid and can't work with symlink correctly

fruch: regardless I've disable the automatic creation for the perf jobs,
there is an inherent problem with building from master for the v14 branch (location of the jobs in the source is different)

so we will be generating from master, only once we'll be running from master
BTW @avikivity was asking about it last week, in regards of tablet related changes that would be done by the team working on tablet would be done on master, and we won't be backporting it to v14. 

I think we should start doing perf runs on master.

roydahan: We won’t do perf runs using master because it will make the tests even more flakey.

what about the multi AZ?
Is it needed here?
Is this how we run it in the original job?

fruch: > Do they want it to run in multi-AZ? Is our job that runs in branch-perf-v14 runs in multi-AZ as well?

They already do, I don't know if they cared about it or not.
and yes it's multi-AZ also on v14

fruch: > We won’t do perf runs using master because it will make the tests even more flakey.

this whole process of creating a new branch once a year, and vetting it, isn't any better.
and the end result we have flakiness we can't detect for months regardless.
working on a branch doesn't make this process any easier.

> 
> what about the multi AZ? Is it needed here? Is this how we run it in the original job?

roydahan: It is.
You won't convince me otherwise.
We don't have to use the same branch, each new test that is not backporting easily can run from a new branch.

fruch: that mean keep manually setup of the job, and having not real visibility or control on what actually runs, as those expend to more and more cases.

it's not not gonna be maintainable like that, and call you took 6 years ago, when SCT was in a condition master could be run at all. it's not the same situation today.

I won't be maintaining a branch per test, and you won't convince otherwise...

PR #7211, opened 2024-02-15
fix(azure): find formal images in community gallary

vponomaryov:
since this is the way to lookup the images our use would use
we are switching to this method, without it we can pickup unpromoated
images, which can lead to rolling upgrades to fail, since base and target
would be identical
    
(cherry picked from commit a50205dc16bdda3a7d157b95beae485af4a9a255)
   
Backport notes:
- Was rebuilt new hydra docker image with `1.60.1-add-libcloud-for-branch-5.2` tag.
  It was needed because of the code base difference where we needed to
  add `apache-libcloud` package to dependencies.
- All of it is needed to make newly added azure-rolling-update CI job work.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

PR #7224, opened 2024-02-22
feature(aws-okta): add support for streamlining OKTA usage

fruch:
this would check first if with default/used profile we
can reach AWS, and if we don't try to use `gimmie-aws-creds`
tools to create a new profile.

### Testing

- [ ]

### Reminders


sdcm/utils/aws_okta.py
```diff
@@ -0,0 +1,80 @@
...
+import configparser
+import datetime
+import logging
+
+import boto3
+from botocore.exceptions import NoCredentialsError
+from gimme_aws_creds.main import GimmeAWSCreds
+from gimme_aws_creds.ui import CLIUserInterface
+
+account_id = os.environ.get('SCT_AWS_ACCOUNT_ID', '797456418907')
```
roydahan, line 26: Is it safe to put the account id here?
Also, which account is it?

sdcm/utils/aws_okta.py
```diff
@@ -0,0 +1,80 @@
...
+            if expire_date > datetime.datetime.now(tz=datetime.timezone.utc):
+                return section
+    return None
+
+
+def try_auth_with_okta():
+    if not can_get_to_aws_account():
+        if not pathlib.Path("~/.okta_aws_login_config").expanduser().exists():
+            raise ValueError("OKTA isn't configured, use this guide to configure it:\n"
+                             "https://www.notion.so/How-to-login-on-AWS-CLI-and-assume-a-role-bcc4e36042ea4ae9a76ea65e7aafe283?pvs=4")
```
roydahan, line 64: Can’t we just trigger the configuration part using the gimme_aws_cred?

sdcm/utils/log.py
```diff
@@ -199,3 +199,20 @@ def configure_logging(exception_handler=None,  # pylint: disable=too-many-argume
             'loggers': loggers
         }
     logging.config.dictConfig(replace_vars(config, variables))
+
+
+def disable_loggers_during_startup():
```
roydahan, line 204: Why is it needed?
Your commits doesn’t explain this part

sdcm/utils/aws_okta.py
fruch: It's not the best, it's already listed in the code in a few places (and maybe others)

We don't have any other option currently, we need it so we can select the correct one, otherwise user would need to pick it every single time here uses a command

sdcm/utils/log.py
fruch: Cause we are doing multiple logins into AWS, it keeps printing again and again in info level, that it found the AWS configuration file

Before it happened once, now it happens multiple times, and this is for suppressing it, which we'll do later as well when we get to reconfigure the loggers.

sdcm/utils/aws_okta.py
fruch: Cause who get to it, need to know what to fill in... 

we'll need to point somehow people to the instructions

sdcm/utils/aws_okta.py
fruch: https://github.com/search?q=repo%3Ascylladb%2Fscylla-cluster-tests%20797456418907&type=code

fruch: https://docs.aws.amazon.com/accounts/latest/reference/manage-acct-identifiers.html

> "... While account IDs, like any identifying information, should be used and shared carefully, they are not considered secret, sensitive, or confidential information."

sdcm/utils/aws_okta.py
roydahan: I meant to trigger the OKTA configuration directly when we catch that it's not configured.

roydahan: It can be a later improvement if we think it's possible and useful.

roydahan (approved):

Issue #4782, opened 2022-05-23
Hydra downloads ssh keys twice per run - thats one or two times too many


nyh:
When I run `hydra bash` (for example), I see SSH keys being downloaded twice:
```
$ hydra bash
Pull version v1.13 from Docker Hub...
v1.13: Pulling from scylladb/hydra
Digest: sha256:0629f8a22c712fbfde7ff889db6d75cee7e5d55b147b84ab89f6d6a1b1092fd6
Status: Image is up to date for scylladb/hydra:v1.13
docker.io/scylladb/hydra:v1.13
Obtaining QA SSH keys...
QA SSH keys obtained.
Going to run 'bash'...
Obtaining QA SSH keys...
QA SSH keys obtained.
```
Indeed, docker/env/hydra.sh has two calls:
```
# Check for SSH keys
if [[ -z "$HYDRA_HELP" ]]; then
    if [ -z "$HYDRA_DRY_RUN" ]; then
        ${SCT_DIR}/get-qa-ssh-keys.sh
    else
        echo ${SCT_DIR}/get-qa-ssh-keys.sh
    fi
fi

...
if [[ -z "${HYDRA_HELP}" ]]; then
    PREPARE_CMD+="; ${SCT_DIR}/get-qa-ssh-keys.sh"
fi
```

As a first step, we shouldn't do this twice per run - once should be enough.

As a second step, I think it's no necessary to do the actual download even once per run. If the user already has these files, why do we need to download them again on every run? Are we worried that one day we'll want to change these keys and nobody will remember to delete the old keys? If this is the case, how about only checking for a new file once per hour or once per day (say), not every time?

On my machine (and in the fast office network), one run of `get-qa-ssh-keys.sh` takes 1.3 seconds, and the two of them take around 2.6 seconds. This isn't enough time to get coffee, but is certainly noticeable and wasteful.

nyh:
> I agree, we could also potentially store known keys' fingerprints (since those are intended to be public anyway) inside the script, so that if we ever change the keys, the new ones will be downloaded anyway.

Right, this would be similar to what the script already does regarding docker versions, where it doesn't try to reload an image with the same version it expects (although it turns out this code was broken too - see https://github.com/scylladb/scylla-cluster-tests/pull/4784).

[...]

PR #7225, opened 2024-02-22
fix(hydra): remove call to awscli to download keys

fruch:
this would complicate the sisuation once we need
want to use diffent aws profiles, so this move into `sct.py` itself, where we'll be able to even do that selectivly  as needed, or push is down further to places it's gonna be used (i.e. no every command need
validating/downloading those)

### Testing

- [ ]

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -40,6 +40,7 @@ from argus.backend.util.enums import TestStatus
 
 import sct_ssh
 import sct_scan_issues
+from sdcm.keystore import KeyStore
 from sdcm.localhost import LocalHost
 from sdcm.provision import AzureProvisioner
 from sdcm.provision.provisioner import VmInstance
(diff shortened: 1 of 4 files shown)


sct.py
```diff
@@ -185,6 +186,10 @@ def getTestCaseNames(self, testCaseClass):
               help="Install paths for extra python packages to install, scylla-cluster-plugins for example")
 def cli():
     LOGGER.info("install-bash-completion current path: %s", os.getcwd())
+    if not is_using_aws_mock():
+        key_store = KeyStore()
+        key_store.sync(keys=['scylla-qa-ec2', 'scylla-test', 'scylla_test_id_ed25519'],
+                       local_path=Path('~/.ssh/').expanduser(), permissions=0o0600)
```
roydahan, line 192: What is this key "scylla_test_id_ed25519"?

roydahan (dismissed):

sct.py
fruch: In rocky9 we started using this key, we should slowly move everything to this type of key, and remove all the tricks we do in ssh servers to enable old types

nyh:
See also https://github.com/scylladb/scylla-cluster-tests/issues/4782

sct.py
nyh: It's an interesting question. I don't think there's a clear consensus of why the newer elliptic-key algorithm is better than the older RSA. Neither are "quantum safe", or offer any other huge advantage over the other. There are some small advantages like key length which doesn't matter to us at all. I think all the hoopla against RSA comes from this - https://www.openssh.com/txt/release-8.2 - it isn't RSA which sucks but SHA-1, and the original openssh RSA implementation used SHA-1 (the current one uses SHA-2 by default, and doesn't suffer from this problem).

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

sct.py
fruch: I care more about things that would work best out of the box, then which key is really superior.

And the rsa ones don't work out of the box, and were causing use issues cause we need to do all kinds of work arounds to enable it

PR #7229, opened 2024-02-26
test_unreachable_subnets

juliayakovlev:
Will be added

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/aws_region.py
+++ b/sdcm/utils/aws_region.py
@@ -31,13 +31,14 @@ class AwsRegion:
     SCT_VPC_CIDR_TMPL = "10.{}.0.0/16"
     SCT_SECURITY_GROUP_NAME = "SCT-2-sg"
     SCT_TEST_SECURITY_GROUP_NAME_TMPL = "SCT-2-sg-{}"
-    SCT_SUBNET_NAME = "SCT-2-subnet-{availability_zone}"
+    SCT_SUBNET_NAME = "SCT-2-subnet-{availability_zone}{subnet_index}"
     SCT_INTERNET_GATEWAY_NAME = "SCT-2-igw"
-    SCT_ROUTE_TABLE_NAME = "SCT-2-rt"
+    SCT_ROUTE_TABLE_NAME = "SCT-2-rt{index}"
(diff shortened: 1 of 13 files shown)


sdcm/utils/aws_region.py
```diff
@@ -551,3 +551,184 @@ def configure(self):
         self.create_sct_ssh_security_group()
         self.create_sct_key_pair()
         LOGGER.info("Region configured successfully.")
+
+
+class AwsRegion(AwsRegionBase):
```
fruch, line 556: not sure I really understand the hierarchy here

it's hard to understand what are the changes like that, cause the code it copied

fruch:
@juliayakovlev see this: https://github.com/scylladb/scylla-cluster-tests/commit/06326a630429e94214f26c3112d1ceca16611f77

I've tested, and the network configuration out of the box seems o.k., didn't run a whole test
but here:

```bash
ubuntu@ip-10-0-8-220:~/sct-results$ cat /etc/netplan/50-cloud-init.yaml
# This file is generated from information provided by the datasource.  Changes
# to it will not persist across an instance reboot.  To disable cloud-init's
# network configuration capabilities, write a file
# /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg with the following:
# network: {config: disabled}
network:
    ethernets:
        enp39s0:
            dhcp4: true
            dhcp4-overrides: &id001
                route-metric: 100
            dhcp6: true
            dhcp6-overrides: *id001
            match:
                macaddress: 0e:a2:74:38:23:69
            set-name: enp39s0
        enp40s0:
            dhcp4: true
            dhcp4-overrides: &id002
                route-metric: 200
            dhcp6: true
            dhcp6-overrides: *id002
            match:
                macaddress: 0e:f9:b6:8b:7d:55
            set-name: enp40s0
    version: 2
``` 

if we'll find we need our hack in user-data we can try this:
https://github.com/scylladb/scylla-cluster-tests/commit/484420b1cc60e61c42edf6b7b0f87efaf11033b4

this I didn't tested

juliayakovlev:
Close with https://github.com/scylladb/scylla-cluster-tests/pull/7284

PR #7231, opened 2024-02-27
fix(manager_cli): set --force-non-ssl-session-port if encryption enabled but port not defined

karol-kokoszka:
There is long lasting issue in Scylla https://github.com/scylladb/scylladb/issues/7206#issuecomment-689598343
which causes Scylla to fallback the SSL port to non-SSL one when the encryption on session is enabled, but SSL port is not defined in config.
Manager sanity tests checks if the healthcheck works with encryption enabled, but at the same time, the scylla configuration used for this test, doesn't specify SSL port (so, it fallsback to non-SSL).
See the `scylla.yaml` from test
```
alternator_enforce_authorization: false
api_address: 127.0.0.1
api_doc_dir: /opt/scylladb/api/api-doc/
api_port: 10000
api_ui_dir: /opt/scylladb/swagger-ui/dist/
audit: table
audit_categories: DCL,DDL,AUTH,ADMIN
auto_bootstrap: true
batch_size_fail_threshold_in_kb: 1024
batch_size_warn_threshold_in_kb: 128
broadcast_rpc_address: 10.15.3.97
cas_contention_timeout_in_ms: 1000
client_encryption_options:
  certificate: /etc/scylla/ssl_conf/client/test.crt
  enabled: true               <---------------------------- ENABLED but native_transport_port_ssl is missing
  keyfile: /etc/scylla/ssl_conf/client/test.key
  truststore: /etc/scylla/ssl_conf/client/catest.pem
cluster_name: manager-regression-manager--db-cluster-53ae9be8
commitlog_segment_size_in_mb: 32
commitlog_sync: periodic
commitlog_sync_period_in_ms: 10000
commitlog_total_space_in_mb: -1
enable_ipv6_dns_lookup: false
endpoint_snitch: org.apache.cassandra.locator.Ec2Snitch
experimental: true
force_schema_commit_log: true
hinted_handoff_enabled: false
listen_address: 10.15.3.97
murmur3_partitioner_ignore_msb_bits: 12
native_shard_aware_transport_port: 19042
native_transport_port: 9042
num_tokens: 256
partitioner: org.apache.cassandra.dht.Murmur3Partitioner
prometheus_address: 0.0.0.0
read_request_timeout_in_ms: 5000
rpc_address: 10.15.3.97
rpc_port: 9160
seed_provider:
- class_name: org.apache.cassandra.locator.SimpleSeedProvider
  parameters:
  - seeds: 10.12.3.185,10.12.1.112,10.15.3.97
write_request_timeout_in_ms: 2000
```

According to the scylla's behavior, manager is expected to fallback to non-SSL port. The point is that it a bad habit on SM side to always try to connect to this SSL port first and then fallback to non-SSL if connection cannot be established. This check is the only way SM could realize that SSL port is not specified.

Starting with manager-3.2.6, CLI supports additional flag/parameter on cluster. It's `--force-non-ssl-session-port` which defines to always use non-SSL port, even if the encryption is enabled.
The fallback and the check if SSL port is opened is removed from SM 3.2.6.

This PR updates manager test to update cluster with `--force-non-ssl-session-port` when the encryption is enabled.

SCT sanity fails without it https://jenkins.scylladb.com/view/scylla-manager/job/manager-3.2/job/centos-sanity-test/25

I'm unable to prove directly that this PR solves the issue without starting the SCT test locally, but I failed in doing that. More info in one of the comments below (that's why it's DRAFT).

### Testing

- [ ]

### Reminders

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -700,7 +700,7 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         # Scylla-manager should pick up client encryption setting automatically
         healthcheck_task.wait_for_status(list_status=[TaskStatus.DONE], step=5, timeout=240)
 
-        mgr_cluster.update(client_encrypt=True)
+        mgr_cluster.update(client_encrypt=True, force_non_ssl_session_port=mgr_cluster.sctool.is_minimum_3_2_6_version())
         time.sleep(30)  # Make sure healthcheck task is triggered
         healthcheck_task.wait_for_status(list_status=[TaskStatus.DONE], step=5, timeout=240)
         sleep = 40
--- a/sdcm/mgmt/cli.py
+++ b/sdcm/mgmt/cli.py
@@ -43,6 +43,8 @@ REPAIR_TIMEOUT_SEC = 7200  # 2 hours
 
 
 new_command_structure_minimum_version = LooseVersion("3.0")
+forcing_tls_minimum_version = LooseVersion("3.2.6")
+
 # TODO: remove these checks once manager 2.6 is no longer supported
 
 
@@ -692,7 +694,7 @@ class ManagerCluster(ScyllaManagerBase):
         cmd = "cluster delete -c {}".format(self.id)
         self.sctool.run(cmd=cmd, is_verify_errorless_result=True)
 
-    def update(self, name=None, host=None, client_encrypt=None):  # pylint: disable=too-many-arguments
+    def update(self, name=None, host=None, client_encrypt=None, force_non_ssl_session_port=False):  # pylint: disable=too-many-arguments
         """
         $ sctool cluster update --help
         Modify a cluster
@@ -712,6 +714,8 @@ class ManagerCluster(ScyllaManagerBase):
             cmd += " --host={}".format(host)
         if client_encrypt:
             cmd += " --ssl-user-cert-file {} --ssl-user-key-file {}".format(SSL_USER_CERT_FILE, SSL_USER_KEY_FILE)
+        if force_non_ssl_session_port:
+            cmd += "  --force-non-ssl-session-port"
         self.sctool.run(cmd=cmd, is_verify_errorless_result=True)
 
     def delete_task(self, task: ManagerTask):
@@ -1298,6 +1302,10 @@ class SCTool:
     @property
     def is_v3_cli(self):
         return self.parsed_client_version >= new_command_structure_minimum_version
+
+    @property
+    def is_minimum_3_2_6_version(self):
+        return self.parsed_client_version >= forcing_tls_minimum_version
 
 
 class ScyllaMgmt:


karol-kokoszka:
@rayakurl @ilya-rarov I need some help on that.

I'm not able to execute the SCT locally.
Starting it with `aws` as a backend ends up with `sdcm.exceptions.WaitForTimeoutError: Wait for: Node manager-regression-karkok-db-node-7f7c359c-1 [3.232.95.196 | 10.12.3.181] (seed: False): Waiting for SSH to be up: timeout - 500 seconds - expired`
Starting it with `docker` fails with 
```
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/tester.py", line 173, in wrapper
    return method(*args, **kwargs)
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/utils/decorators.py", line 119, in inner
    res = func(*args, **kwargs)
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/tester.py", line 873, in setUp
    self.init_resources()
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/tester.py", line 1791, in init_resources
    self.get_cluster_docker()
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/tester.py", line 1417, in get_cluster_docker
    self.db_cluster = cluster_docker.ScyllaDockerCluster(n_nodes=[self.params.get("n_db_nodes"), ],
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/cluster_docker.py", line 292, in __init__
    super().__init__(docker_image=docker_image,
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/cluster.py", line 3822, in __init__
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/cluster_docker.py", line 221, in __init__
    super().__init__(cluster_prefix=cluster_prefix,
  File "/home/karkok/dev/scylla-cluster-tests/sdcm/cluster.py", line 3133, in __init__
    for idx in range(num):
TypeError: 'str' object cannot be interpreted as an integer
```

... do you have any clues what do I have to do to test it locally ?

fruch:
Please an issue with the commands you were trying 

For runing with AWS from you local development setup, one needs to enable public newtorking

For for the docker backend, need to check, but sounds like it's missing some configuration. We'll need to see the full command being used

karol-kokoszka:
Hey, @fruch 
This is what I've done:
- install hydra `sudo ./install-hydra.sh`
- copied SSH keys from the bucket (get-qa-ssh-keys.sh doesn't work for me Access Denied)
- `export SCT_INSTANCE_PROVISION=on_demand`
- `export SCT_ENABLE_ARGUS=false`
- `export SCT_SCYLLA_VERSION=5.2.1` (although I should do it against enterprise)
- for **docker** `➜  scylla-cluster-tests git:(force_non_ssl_session_port_manager) ✗ hydra run-test mgmt_cli_test.MgmtCliTest.test_manager_sanity --backend docker --config test-cases/manager/manager-regression-multiDC-set-distro.yaml`
- for **aws** `➜  scylla-cluster-tests git:(force_non_ssl_session_port_manager) ✗ hydra run-test mgmt_cli_test.MgmtCliTest.test_manager_sanity --backend aws --config test-cases/manager/manager-regression-multiDC-set-distro.yaml`

fruch:
> Hey, @fruch This is what I've done:
> 
> * install hydra `sudo ./install-hydra.sh`
> * copied SSH keys from the bucket (get-qa-ssh-keys.sh doesn't work for me Access Denied)
> * `export SCT_INSTANCE_PROVISION=on_demand`
> * `export SCT_ENABLE_ARGUS=false`
> * `export SCT_SCYLLA_VERSION=5.2.1` (although I should do it against enterprise)
> * for **docker** `➜  scylla-cluster-tests git:(force_non_ssl_session_port_manager) ✗ hydra run-test mgmt_cli_test.MgmtCliTest.test_manager_sanity --backend docker --config test-cases/manager/manager-regression-multiDC-set-distro.yaml`

looking at it again, the docker backend for the manager tests is a bit problematic, since there no actual monitor node for docker backend, it's uses the monitoring stack directly, hence there no place that the manager can be installed

> * for **aws** `➜  scylla-cluster-tests git:(force_non_ssl_session_port_manager) ✗ hydra run-test mgmt_cli_test.MgmtCliTest.test_manager_sanity --backend aws --config test-cases/manager/manager-regression-multiDC-set-distro.yaml`

for running aws from local machine, see https://github.com/scylladb/scylla-cluster-tests?tab=readme-ov-file#run-test-locally-with-aws-backend
you are missing are missing `--config configurations/network_config/all_addresses_ipv6_public.yaml`


also probably the quickest way to try that out, is to clone one of those jobs into staging folder, point it to your fork/branch, and run it the parameters you need (same as used in release/master runs)

karol-kokoszka:
@fruch I tried to execute the test using jenkins.
This job https://jenkins.scylladb.com/view/scylla-manager/job/manager-3.2/job/debian11-sanity-test/configure configuration points to my forked repo github.com/karol-kokoszka/scylla-cluster-tests and it should be executed against `force_non_ssl_session_port_manager` branch.

Unfortunately test fails on setting up the node (? I guess)
Do you have any clues what am I missing ? https://jenkins.scylladb.com/view/scylla-manager/job/manager-3.2/job/debian11-sanity-test/22/console
I expect the test to fail still, becuase it checks if the version is > 3.2.6 but we havent tag any new version yet, but I cannot even reach the point where it fails on particular test yet.

fruch:
> @fruch I tried to execute the test using jenkins. This job https://jenkins.scylladb.com/view/scylla-manager/job/manager-3.2/job/debian11-sanity-test/configure configuration points to my forked repo github.com/karol-kokoszka/scylla-cluster-tests and it should be executed against `force_non_ssl_session_port_manager` branch.
> 
> Unfortunately test fails on setting up the node (? I guess) Do you have any clues what am I missing ? https://jenkins.scylladb.com/view/scylla-manager/job/manager-3.2/job/debian11-sanity-test/22/console I expect the test to fail still, becuase it checks if the version is > 3.2.6 but we havent tag any new version yet, but I cannot even reach the point where it fails on particular test yet.

the failure is 
```
18:39:57  Warning: apt-key is deprecated. Manage keyring files in trusted.gpg.d instead (see apt-key(8)).
18:39:57  E: This command can only be used by root.
```

2 observations, your branch is based on master SCT, and not branch-3.2. might be the reason, but it might be a something new, and since those test are not run regular with SCT master, no one noticed it.

Also I would recommend cloning jobs into staging folder, and changing the official jobs of branch-3.2 for your testing.
https://jenkins.scylladb.com/view/staging/job/scylla-staging/

Issue #7232, opened 2024-02-27
rolling upgrades failed on step "Will check paged query before upgrading nodes" 


temichus:
## Packages

Scylla version: `5.4.3-20240211.cf42ca0c2a65` with build-id `3b28ea7f1f552247f942244e28fe84917c997dcb`

Kernel Version: `5.10.0-28-cloud-amd64`

## Issue description

rolling upgrade tests failed with an error before the upgrade itself:
```
2024-02-24 05:35:12.049: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=23026c49-5e74-49eb-b04d-354d02dd01ab, source=UpgradeTest.test_rolling_upgrade (upgrade_test.UpgradeTest)() message=Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/upgrade_test.py", line 660, in test_rolling_upgrade
self.paged_query()
File "/home/ubuntu/scylla-cluster-tests/sdcm/fill_db_data.py", line 3389, in paged_query
full_res = self.rows_to_list(
File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 2419, in rows_to_list
return [list(row) for row in rows]
TypeError: 'NoneType' object is not iterable
```

 on step ```Will check paged query before upgrading nodes```

## Installation details

Cluster size: 4 nodes (n2-highmem-48)

Scylla Nodes used in this run:
- rolling-upgrade--debian-bullseye-db-node-cc3be31d-0-4 (35.237.87.235 | 10.142.0.217) (shards: 44)
- rolling-upgrade--debian-bullseye-db-node-cc3be31d-0-3 (35.237.104.100 | 10.142.0.214) (shards: 44)
- rolling-upgrade--debian-bullseye-db-node-cc3be31d-0-2 (34.139.253.94 | 10.142.0.203) (shards: 44)
- rolling-upgrade--debian-bullseye-db-node-cc3be31d-0-1 (34.23.120.123 | 10.142.0.94) (shards: 44)

OS / Image: `https://www.googleapis.com/compute/v1/projects/debian-cloud/global/images/family/debian-11` (gce: undefined_region)

Test: `rolling-upgrade-debian11-test`
Test id: `cc3be31d-e886-423a-a9fc-3f8d337648e8`
Test name: `scylla-master/rolling-upgrade/rolling-upgrade-debian11-test`
Test config file(s):

[...]

PR #7233, opened 2024-02-27
fix(fetch_all_rows): handle default `timeout=None` correctly

fruch:
25d8252 introduced getting the timeout of the future been returned from `execute_async`, it was tested and working o.k. except one path that `fetch_all_rows` was given `timeout=None` (which is the default)

then `fetch_all_rows` blindly put the `None` into
`session.default_timeout`, we never want unlimited timeout set.

fix is to not set `default_timeout` with `None`

Fixes: #7232

### Testing

- [x] introduced a new integration test

### Reminders


vponomaryov (approved):
LGTM

juliayakovlev (approved):

PR #7236, opened 2024-02-29
improvement(kms): improve KMS enc-on checks when nodetool cfstats fails

vponomaryov:
In the `start_kms_key_rotation_thread` it is possible to face the situation when the `nodetool cfstats` command fails (i.e. during `upgrade` test).
And we do not handle the case when it fails leading to the following error:

```
  File "scylla-cluster-tests/sdcm/cluster.py", line 4371, in _rotate_kms_key
  cf_stats = target_node._parse_cfstats(res.stdout)
  AttributeError: 'NoneType' object has no attribute 'stdout'
```

So, handle the case of the 'None' value as a return value for the call of the `target_node._parse_cfstats` function.

Note that, in good case, we should fix SCT the way that a node that is being upgraded should not be picked up by the KMS encryption checks and node upgrade should wait while the KMS encryption check finishes (several minutes).

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4389,9 +4389,12 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
                             res = target_node.run_nodetool(
                                 sub_cmd='cfstats', args=ks_cf, timeout=300, retry=3, publish_event=False,
                                 warning_event_on_exception=(Failure, UnexpectedExit, Libssh2_UnexpectedExit))
-                            cf_stats = target_node._parse_cfstats(res.stdout)  # pylint: disable=protected-access
-                            if int(cf_stats["SSTable count"]) > chosen_ks_cf_sstables_num:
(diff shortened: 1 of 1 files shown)


vponomaryov:
Example of the error: [enterprise-2024.1/rolling-upgrade/rolling-upgrade-ami-test#12](https://argus.scylladb.com/test/ec795529-c9e8-419b-9160-ad106aba85a6/runs?additionalRuns[]=dd9f9a65-b229-4a06-9b3f-8eafdd08c240)

Being tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-rolling-upgrade-ami-test/3/

fruch:
@vponomaryov 

I think we can use the same way nemesis work, and marked the upgraded one as being under nemesis, it would help abit

roydahan (approved):

fruch (approved):
LGTM

PR #7247, opened 2024-03-06
improvement(get_username.py): User impersonation for automated builds

k0machi:
This commit adds a new job parameter to job pipelines intended to be
filled by job automation - requested_by_user. This allows SCT to
properly track resources if a job was started by a third party API
call, e.g. through Argus.

Task: scylladb/qa-tasks#1612

### Testing

- [x] [Argus](https://argus.scylladb.com/test/b0b22734-758e-47b4-b6ec-29e64d212d52/runs?additionalRuns[]=aa930251-34da-480c-bd90-ef87e3092a34)
- [x] [Jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/alexey-argus-testing/69/)

### Reminders

### DIFF
--- a/sdcm/utils/get_username.py
+++ b/sdcm/utils/get_username.py
@@ -25,7 +25,12 @@ def get_email_user(email_addr: str) -> str:
 
 
 def get_username() -> str:  # pylint: disable=too-many-return-statements
-    # First check that we running on Jenkins try to get user email
+    # First we check if user is being impersonated by an api call
+    actual_user_from_request = os.environ.get('BUILD_USER_REQUESTED_BY')
+    if actual_user_from_request:
+        return actual_user_from_request
+
+    # Then check that we running on Jenkins try to get user email
     email = os.environ.get('BUILD_USER_EMAIL')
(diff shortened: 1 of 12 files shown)


k0machi:
Regarding editing job description to contain impersonated username - I feel like that would be good to delegate to argus - when starting a job it can automatically adjust the description to append requested user. Would like some input on that.

fruch (approved):

soyacz (approved):
LGTM

soyacz:
One thing I don't yet understand - if clicking rebuild redirects to Jenkins and requires to use jenkins to rebuild, then why do we need this change?

fruch:
> One thing I don't yet understand - if clicking rebuild redirects to Jenkins and requires to use jenkins to rebuild, then why do we need this change?

it's now doing that, we plan on changing it, and building a better form on Argus end, and not go via jenkins forms.
also we want few more other capabilities like clone job and run it, and bulk run (and probably more).

so for that we need this small change, since Argus would be using one specific token (i.e. user)

PR #7248, opened 2024-03-06
Move cloud devops cleanup to sct

soyacz:
Moving cleanup from cloud-devops repo to SCT.
GCE script was recreated from scratch as was requiring unused by SCT dependencies
and anyway was not working properly.

refs: https://github.com/scylladb/qa-tasks/issues/1535

### Testing

- [ ]

### Reminders


soyacz:
need to adjust jenkins job and fix imports when running with hydra.

soyacz:
Ok, all cleanup scripts moved and pipeline adjusted. Ready for review.

utils/cloud_cleanup/aws/clean_aws.py
```diff
@@ -0,0 +1,346 @@
...
+    ec2 = boto3.resource('ec2', region_name=region_name)
+
+    volumes = ec2.volumes.all()
+
+    count_kept_volume = 0
+    count_deleted_volume = 0
+    for volume in volumes:
+        debug("checking volume %s %s %s %s" % (volume.id, volume.volume_id, volume.state, volume.tags))
+        keep_alive = keep_alive_volume(volume)
+        if keep_alive or volume.state == "in-use":
```
fruch, line 249: https://github.com/scylladb/cloud-devops/issues/27

still need to be fixed, exactly here...

utils/cloud_cleanup/gce/clean_gce.py
```diff
@@ -0,0 +1,98 @@
...
+                            default=os.environ.get('DRY_RUN', False))
+
+    args = arg_parser.parse_args()
+
+    is_dry_run = bool(args.dry_run)
+
+    if is_dry_run:
+        LOGGER.error("'Dry run' mode on")
+
+    projects = ['sct-project-1', 'local-ssd-latency']
```
fruch, line 94: we have one project missing here, we should use `SUPPORTED_PROJECTS`

```
SUPPORTED_PROJECTS = {'gcp', 'gcp-sct-project-1',
                      'gcp-local-ssd-latency'} | {os.environ.get('SCT_GCE_PROJECT', 'gcp-sct-project-1')}
```

jenkins-pipelines/qa/hydra-cleanup-cloud.jenkinsfile
```diff
@@ -46,11 +46,29 @@ pipeline {
                 }
             }
         }
-        stage('Run hydra clean_azure.py') {
+        stage('Clean AWS') {
```
fruch, line 49: we made sure that if there a failure with one of the stages, it moves to the next one ?

mykaul:
Are we sure it's easier to develop all this code rather than using Ansible (for example - https://docs.ansible.com/ansible/latest/collections/azure/azcollection/azure_rm_virtualmachine_module.html ) ?

utils/cloud_cleanup/aws/clean_aws.py
soyacz: fixed (I suppose) - added to `keep_alve_volume` def.

fruch:
> Are we sure it's easier to develop all this code rather than using Ansible (for example - https://docs.ansible.com/ansible/latest/collections/azure/azcollection/azure_rm_virtualmachine_module.html ) ?

yes

utils/cloud_cleanup/gce/clean_gce.py
soyacz: I don't understand it - we don't have project that are prefixed with `gcp-` - I believe it is added just for downloading credentials and we take project from credentials info (please correct me if I'm wrong). So we actually have 2 projects 'sct-project-1` and `local-ssd-latency`. I don't know then what is `gcp` project in that case.

Somehow, default credentials work for both of them.

fruch: We have 3 projects 

skilled-adapter-452 is what we call gcp (it was the first project.

We still want to scan it 

You can take a look at the emails report, people are running machines there, some from out older branches from before we used sct-project-1

utils/cloud_cleanup/aws/clean_aws.py
fruch: o.k. didn't saw it

soyacz: you couldn't as it wasn't that time in place :)

utils/cloud_cleanup/gce/clean_gce.py
soyacz: again your'e right, and I got into problem of insufficient privileges - so I adapted code to work similarly as in other places (creating a new client for each credentials and reusing credential's project_id).

soyacz:
@fruch I noticed dangling instances in `Scylla` project would be deleted - e.g. instances even from year 2020. 
This is not due the changed behavior from stopping to terminating as there are running instances since 2022 (e.g. `sl-test-2`).
I'll let know on the slack engineering channel that these gonna be deleted.

fruch:
> @fruch I noticed dangling instances in `Scylla` project would be deleted - e.g. instances even from year 2020. 
> This is not due the changed behavior from stopping to terminating as there are running instances since 2022 (e.g. `sl-test-2`).
> I'll let know on the slack engineering channel that these gonna be deleted.

O.k. could be we are not looking at network tags ? In GCP we might have an old instance that has a network label 'keep:alive'

soyacz:
@fruch some are using `labels` instead of metadata. I'll adjust the code to take also labels so in case someone used that doesn't find machine terminated.
I can extend also to `tags` next to labels - for the sake of safety.

fruch:
@soyacz needs your confirmation, this code doesn't gonna erase labeled GCP instances, and then we can merge it

soyacz:
> @soyacz needs your confirmation, this code doesn't gonna erase labeled GCP instances, and then we can merge it
> 
> and disable the old pipelines

yes, it's not going to erase labeled instances, also if there's `keep-alive` network tag also machine will be skipped.

fruch (approved):
LGTM

vponomaryov (approved):
LGTM
Hope it doesn't cause problems to K8S resources.

fruch:
> LGTM
> Hope it doesn't cause problems to K8S resources.

It actually  addresses one of the k8s related issues with volumes

vponomaryov:
> > LGTM
> > Hope it doesn't cause problems to K8S resources.
> 
> It actually addresses one of the k8s related issues with volumes

It actually doesn't touch K8S clusters and will delete it's nodes which will get auto-created right-away by node auto-scaler.
It will last endlessly while human doesn't intervene.

Also, there are some other resource kinds which are not covered here.

fruch:
> > > LGTM
> > > Hope it doesn't cause problems to K8S resources.
> > 
> > 
> > It actually addresses one of the k8s related issues with volumes
> 
> It actually doesn't touch K8S clusters and will delete it's nodes which will get auto-created right-away by node auto-scaler. It will last endlessly while human doesn't intervene.
> 
> Also, there are some other resource kinds which are not covered here.

so nothing new it is,

if you know of such resources, open an issue that then need to be added, if it's k8s related or other.

now we'll be able to do the cleanup, at the same time we introduce a new resource type usage.

vponomaryov:
> > > > LGTM
> > > > Hope it doesn't cause problems to K8S resources.
> > > 
> > > 
> > > It actually addresses one of the k8s related issues with volumes
> > 
> > 
> > It actually doesn't touch K8S clusters and will delete it's nodes which will get auto-created right-away by node auto-scaler. It will last endlessly while human doesn't intervene.
> > Also, there are some other resource kinds which are not covered here.
> 
> so nothing new it is,
> 
> if you know of such resources, open an issue that then need to be added, if it's k8s related or other.
> 
> now we'll be able to do the cleanup, at the same time we introduce a new resource type usage.

Launch templates, load balancers.

PR #7249, opened 2024-03-06
fix(azure): make upgrade work on Azure

vponomaryov:
Ref: https://github.com/scylladb/scylladb/issues/17659
Ref: https://github.com/scylladb/scylla-tools-java/issues/338

### Testing

- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-rolling-upgrade-azure-image-test/7

### Reminders

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -692,11 +692,12 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
             assert all(tables_upgraded), "Failed to upgrade the sstable format {}".format(tables_upgraded)
 
         # Verify sstabledump
-        InfoEvent(message='Starting sstabledump to verify correctness of sstables').publish()
-        self.db_cluster.nodes[0].remoter.run(
-            'for i in `sudo find /var/lib/scylla/data/keyspace_complex/ -type f |grep -v manifest.json |'
-            'grep -v snapshots |head -n 1`; do echo $i; sudo sstabledump $i 1>/tmp/sstabledump.output || '
-            'exit 1; done', verbose=True)
+        if self.params.get("cluster_backend") != "azure":
+            InfoEvent(message='Starting sstabledump to verify correctness of sstables').publish()
+            self.db_cluster.nodes[0].remoter.run(
+                'for i in `sudo find /var/lib/scylla/data/keyspace_complex/ -type f |grep -v manifest.json |'
+                'grep -v snapshots |head -n 1`; do echo $i; sudo sstabledump $i 1>/tmp/sstabledump.output || '
+                'exit 1; done', verbose=True)
 
         InfoEvent(message='Step8 - Run stress and verify after upgrading entire cluster').publish()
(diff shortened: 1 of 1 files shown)


upgrade_test.py
```diff
@@ -692,11 +693,13 @@ def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-stat
             assert all(tables_upgraded), "Failed to upgrade the sstable format {}".format(tables_upgraded)
 
         # Verify sstabledump
-        InfoEvent(message='Starting sstabledump to verify correctness of sstables').publish()
-        self.db_cluster.nodes[0].remoter.run(
-            'for i in `sudo find /var/lib/scylla/data/keyspace_complex/ -type f |grep -v manifest.json |'
-            'grep -v snapshots |head -n 1`; do echo $i; sudo sstabledump $i 1>/tmp/sstabledump.output || '
-            'exit 1; done', verbose=True)
+        current_scylla_version = ComparableScyllaVersion(self.db_cluster.nodes[0].scylla_version)
+        if current_scylla_version >= ("2024.1.0" if self.db_cluster.nodes[0].is_enterprise else "5.4.0"):
```
fruch, line 697: I think you want to change this logic to include 5.2, no ?

vponomaryov: I made it be applicable for enterprise branches too.
Is it a problem?

fruch: I don't understand how it's gonna work for 5.2..

fruch: I'm still lost and I don't understand what you are doing here

vponomaryov: @fruch if it is not clear what am I doing here then need to open ref issues and read a bit about it.

fruch: yeah I did read them
https://github.com/scylladb/scylla-tools-java/issues/338#issuecomment-1980948132

but cause you open a new issue, the discussion got splitted, and @mykaul say one things, while me and @denesb discussion something else

I'm not o.k. with removing completely this check from all of the rolling upgrades in 5.2
two options:
1) disable it only for Azure
2) test if @denesb assumption that `scylla sstable dump` is backported and functional in 5.2, and backport the needed code in SCT to use that (as in SCT master)

denesb: I don't think I will backport any more scylla-sstable stuff to 5.2, it has like a month before EOD and 2023.1 (the enterprise version that is based on it) already has all the backports.
scylla-sstable as-is in 5.2 *should* work for most use-cases, but there are minor fixes missing and a major feature as well: dumping MV/SI sstables doesn't work at all.

fruch: I see someone is optimistic about when 6.0 gonna be out :)

@vponomaryov, if so just replace the logic with checking `self.params.get('cluster_backend') == 'azure'` instead of versions, we don't need to support other newer versions on branch-5.2, not much point in that

fruch: and we don't need this in any other branch as far as I know.
in 2023.1 we already using the `scylla sstable`

Issue #7252, opened 2024-03-07
Operator tests should advertise xunit/junit results


mykaul:
So we can see them in Jenkins (and Argus?) as they are running regular pytest or whatnot test suite that does provide that.
Currently, it looks like everything case failed, which is far from true.
@vponomaryov and I were looking at this earlier this week.

PR #7256, opened 2024-03-07
feature(k8s-functional): report test results with junit xml

fruch:
we should be sending those results to Argus and to Jenkins so it would be easier to debug those runs

Ref: #7252

### Testing

- [ ]

### Reminders


sct.py
```diff
@@ -1755,6 +1760,24 @@ def finish_argus_test_run(jenkins_status):
         LOGGER.error("Failed to submit data to Argus", exc_info=True)
 
 
+@cli.command("fetch-junit-form-runner", help="copy the junit.xml back")
```
vponomaryov, line 1763: Typo - `form`.

sdcm/logcollector.py
```diff
@@ -1014,6 +1014,8 @@ class BaseSCTLogCollector(LogCollector):
         FileLog(name='result_gradual_increase.log'),
         FileLog(name='partition_range_scan_diff_*.log',
                 search_locally=True),
+        FileLog(name='junit.xml',
```
vponomaryov, line 1017: Why is it in the base class?
Should be in the K8S log collector.

vars/runSctTest.groovy
```diff
@@ -171,4 +171,14 @@ def call(Map params, String region, functional_test = false, Map pipelineParams
     fi
     echo "end test ....."
     """
+
+    if (functional_test != null && functional_test) {
+        sh """
+            RUNNER_IP=\$(cat sct_runner_ip||echo "")
+            if [[ -n "\${RUNNER_IP}" ]] ; then
+                ./docker/env/hydra.sh fetch-junit-form-runner --runner-ip \${RUNNER_IP}
```
vponomaryov, line 179: Same here - typo.

sct.py
```diff
@@ -1755,6 +1760,24 @@ def finish_argus_test_run(jenkins_status):
...
+@click.argument("runner-ip", type=str)
+def fetch_junit(runner_ip):
```
vponomaryov, line 1765: Why so specific func?
Why not "fetch_file"? So it could be reused elsewhere...

sdcm/logcollector.py
fruch: it should come with the SCT logs, it has nothing to do with k8s

sct.py
fruch: one can refactor it, when there would be a need, there isn't a need

sct.py
fruch: fixed

sct.py
```diff
@@ -1150,14 +1150,19 @@ def run_pytest(target, backend, config, logdir):
...
     sys.stdout = OutputLogger(logfile, sys.stdout)
     sys.stderr = OutputLogger(logfile, sys.stderr)
     if not target:
         print("argv is referring to the directory or file that contain tests, it can't be empty")
         sys.exit(1)
-    sys.exit(pytest.main(['-s', '-v', '-p', 'no:warnings', target]))
+    return_code = pytest.main(['-s', '-v', f'--junit-xml={junit_file}', '-p', 'no:warnings', target])
+    test_config = get_test_config()
+    # temp, until we'll have API for sending it
+    test_config.argus_client().submit_sct_logs([LogLink(log_name=junit_file.name, log_link=str(junit_file))])
```
vponomaryov, line 1164: It must respect the `SCT_ENABLE_ARGUS` config option.

vponomaryov: Ok, at moment of writing comment I forgot about mocking it.
Double-checked and having disabled argus it really doesn't break anything.

PR #7262, opened 2024-03-11
improvement(metrics): allow to disable node metrics

soyacz:
For testing purposes we tried to disable metrics to see performance impact of collecting them. This is the code to disable metrics with api

### Testing

- [X] -  tested locally with docker backend

### Reminders

### DIFF
--- a/sdcm/utils/scylla_metrics_ctrl.py
+++ b/sdcm/utils/scylla_metrics_ctrl.py
@@ -1,0 +1,22 @@
+from typing import Literal
+
+from sdcm.cluster import BaseNode
+
+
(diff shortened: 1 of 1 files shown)


soyacz:
@fruch this feature was added for testing performance without metrics, but currently not used anywhere. Do we merge it or close it?

fruch (approved):
I don't like merging unused code, but it might be a good reference.

It would be nice if there would be a configuration that can call this code

soyacz:
> I don't like merging unused code, but it might be a good reference.
> 
> It would be nice if there would be a configuration that can call this code

I moved code to external file to avoid bloating `Cluster` class. Should now be less intrusive.

fruch (approved):
LGTM

PR #7264, opened 2024-03-12
improvement(event): add AuthenticationError c-s subevent

juliayakovlev:
Add CassandraStressLogEvent.add_subevent_type WARNING for cassandra-stress error:
```
```

This error caused by audit enabled. And during rollback test may cause to error from c-s. By default audit is disabled in 20223.1 by scylladb/scylla-enterprise#3094. But it won't be disabled in 2022.1 and 2022.2.
This message generates too much noise for us. We do not need it will fail the test.

This change was added to branch-2023.1 only by https://github.com/scylladb/scylla-cluster-tests/pull/6590. But was missed when the branch was recreated

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/sct_events/loaders.py
+++ b/sdcm/sct_events/loaders.py
@@ -174,6 +174,7 @@ class CassandraStressLogEvent(LogEvent, abstract=True):
     TooManyHintsInFlight: Type[LogEventProtocol]
     ShardAwareDriver: Type[LogEventProtocol]
     SchemaDisagreement: Type[LogEventProtocol]
+    AuthenticationError: Type[LogEventProtocol]
 
 
 class SchemaDisagreementErrorEvent(SctEvent):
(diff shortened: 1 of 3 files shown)


sdcm/sct_events/loaders.py
```diff
@@ -223,15 +224,22 @@ def msgfmt(self):
                                           regex="Using optimized driver")
 CassandraStressLogEvent.add_subevent_type("SchemaDisagreement", severity=Severity.WARNING,
                                           regex="No schema agreement")
-
+# By default audit is disabled in 20223.1 by https://github.com/scylladb/scylla-enterprise/pull/3094.
+# But it won't be disabled in 2022.1 and 2022.2.
+# This message generates too much noise for us. We do not need it will fail the test. Create WARNING message, not ERROR.
+# This event will be cretaed in branch 2023.1, not in 2022.x
+CassandraStressLogEvent.add_subevent_type("AuthenticationError", severity=Severity.WARNING,
```
fruch, line 231: Now I might complicate things a bit...

maybe we can introduce it as a filter ? and put it in `enable_default_filters` like the rest ?
and also use `SkipPerIssue` ?

juliayakovlev: There is no opened issue for that.

juliayakovlev: @fruch 
I can add filter in `enable_default_filters` as :
```
DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line='Rate-limit: supressed').publish()
```

fruch: > There is no opened issue for that. 
> 

There is but it's closed, we should get it tagged with the needed tag so it would be skip in 2023.1 and 2024.1

juliayakovlev: > > There is no opened issue for that.
> 
> There is but it's closed, we should get it tagged with the needed tag so it would be skip in 2023.1 and 2024.1

Which tag do you mean? Label?

juliayakovlev: > > There is no opened issue for that.
> 
> There is but it's closed, we should get it tagged with the needed tag so it would be skip in 2023.1 and 2024.1

@fruch  who can put the labels?

juliayakovlev: As I see from the code, it should be label like 'sct-2023.1-skip'

juliayakovlev:
Close this PR as we already use SkipPerIssue for this case.
Two labels `sct-2023.1-skip` and `sct-2024.1-skip` are added to https://github.com/scylladb/scylla-enterprise/issues/3148. So this SkipPerIssue should to start to work

PR #7269, opened 2024-03-13
fix(cleanup): make cleanup terminate by default

soyacz:
In case there's no keep_action tag set we stop by default - I suppose we should be terminating instances to avoid growing number of stopped instances.

### Testing

- [ ]

### Reminders


vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #7270, opened 2024-03-13
fix(RollingConfigChangeInternodeCompression): properly identify None

ShlomiBalalis:
When the current internode_compression value in the scylla.yaml is "none", it is represented as None when being read by python. So, when get_internode_compression_new_value_randomly tries to find the current value of internode_compression in the values list ['dc', 'all', 'none'] it cannot find it, since the none in the list is a string. As a result, the value of the new internode_compression chosen by the nemesis could be 'none' again, and thus it has not changed at all, and the nemesis did not execute its primary function.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

PR #7281, opened 2024-03-19
fix(issues): handle call to github api a bit better

fruch:
since we are hitting the rate-limit of the API calls too often we are doing the following things:

* stop all retries - since on rate limit it would just gonna hang for us
* import the call sequence - so it would be one api call per issue, and not 3
* cache the client at the class level, so we won't be doing login multiple times

it don't handle all the issue we might have with it, but it can elivate some of the hanging it cause for our tests

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/issues.py
+++ b/sdcm/utils/issues.py
@@ -36,13 +36,21 @@ class SkipPerIssues:
     and would return false if all issues closed.
     """
 
+    _github = None
+
+    @classmethod
+    @property
+    def github(cls):
+        if not cls._github:
+            github_access = KeyStore().get_json("github_access.json")
+            auth = github.Auth.Token(token=github_access["token"])
+            cls._github = github.Github(auth=auth, retry=None)
+            rate = cls._github.get_rate_limit()
+            logging.debug(rate.raw_data)
+        return cls._github
+
     def __init__(self, issues: list[str] | str, params: SCTConfiguration | dict):
         self.params = params
-
-        github_access = KeyStore().get_json("github_access.json")
-        auth = github.Auth.Token(token=github_access["token"])
-        self.github = github.Github(auth=auth)
-
         issues = [issues] if isinstance(issues, str) else issues
 
         self.issues = [self.get_issue_details(issue) for issue in issues]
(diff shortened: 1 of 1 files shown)


fruch:
followup should be: 
- apply similar fix also in dtest (we are using the same key/user there)
- switch to github app, where we should have X3 of the limit (@k0machi maybe we can give it a try with the app we have for argus)
- consider "centralized" cache of the data, maybe in Argus ?

sdcm/utils/issues.py
```diff
@@ -36,13 +36,21 @@ class SkipPerIssues:
...
+
+    @classmethod
+    @property
+    def github(cls):
+        if not cls._github:
+            github_access = KeyStore().get_json("github_access.json")
+            auth = github.Auth.Token(token=github_access["token"])
+            cls._github = github.Github(auth=auth, retry=None)
+            rate = cls._github.get_rate_limit()
+            logging.debug(rate.raw_data)
```
soyacz, line 49: add a log message to be able to find/understand it in logs

soyacz (approved):
we'll get `TestFrameworkEvent` Error events in case of GH being down/network temporary issue. 
I don't know if this is a problem, but if it turns to be one, we'll need to add some kind of simple&fast retry or reduce severity of the event.

besides minor issue LGTM.

juliayakovlev (approved):

fruch:
> we'll get `TestFrameworkEvent` Error events in case of GH being down/network temporary issue. 
> I don't know if this is a problem, but if it turns to be one, we'll need to add some kind of simple&fast retry or reduce severity of the event.
> 
> besides minor issue LGTM.

I think for the current situation, we'll take down the severity.

It's gonna fail, and quite a lot of times

I'm still puzzled by what surfed hitting this rate limit this week, I suspect recent dtest change, but couldn't prove it yet

vponomaryov (approved):
LGTM

fruch:
It's still problematic, cause the unittests are failing, cause some of their expectations is depended this logic

I think of moving those test into the integration suite

PR #7282, opened 2024-03-19
fix(perf): fix steady state latency measurement in perf upgrade test

soyacz:
Latency measured during c-s startup can be high and should not be counted as 'steady'.

Postpone latency measurement in perf upgrade test by 60 secs to skip c-s startup period.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

PR #7283, opened 2024-03-19
fix(issues): don't raise errors from rate-limit failure

fruch:
since we are still running out of our limits alot, we shouldn't raise error events on those

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/issues.py
+++ b/sdcm/utils/issues.py
@@ -6,7 +6,7 @@ from dataclasses import dataclass
 
 import github
 import github.Auth
-from github.GithubException import UnknownObjectException
+from github.GithubException import UnknownObjectException, RateLimitExceededException
 
 from sdcm.keystore import KeyStore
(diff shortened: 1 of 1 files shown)


sdcm/utils/issues.py
```diff
@@ -88,6 +88,9 @@ def get_issue_details(self, issue):
                                severity=Severity.WARNING,
                                trace=sys._getframe().f_back).publish()  # pylint: disable=protected-access
             return None
+        except RateLimitExceededException as exc:
+            logging.debug('RateLimitExceededException raise: %s', str(exc))
```
vponomaryov, line 92: Why `debug`? It should be visible in jenkins for fast debugging.
Please, make it be `warning`.

fruch: no, I don't see the need for it.

we shouldn't spam the output with things are not really important, this isn't really important

PR #7289, opened 2024-03-21
feature(provision): enable capacity reservation for aws

soyacz:
Sometimes we get failures due no capacity (also for on_demand instances) and it's hard to trigger a test at all. This is often a problem for performance tests where we use older instance types.

Introducing capacity reservation feature for AWS cloud. It works by first reserving capacity for all instance types used by test for test duration (reserves predicted max number of nodes that test may use based on test config parameters) and during instance creation uses it to spin it. In case given AZ cannot serve requested instances, it fallback to
another one until successful.

### Testing

- [x] - https://argus.scylladb.com/workspace?state=WyJhMmFmMDkxZS0zY2JkLTQwNjgtYjgyNS0zNWQ1MTFmMTVkMTciXQ (grow shrink cluster test)
- [x] perf test - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/scylla-enterprise-perf-regression-latency-shard-aware-1TB-test/9 (acutally it did fallback to another AZ in this run)

### Reminders

### DIFF
--- a/sdcm/provision/aws/capacity_reservation.py
+++ b/sdcm/provision/aws/capacity_reservation.py
@@ -1,0 +1,236 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2024 ScyllaDB
+import logging
+import time
+from datetime import datetime, timedelta
+from typing import List, Dict, Tuple
+
+from botocore.exceptions import ClientError
+import boto3
+
+from sdcm.exceptions import CapacityReservationError
+
+LOGGER = logging.getLogger(__name__)
+
+
+class SCTCapacityReservation:
+    """Class for managing capacity reservations for AWS instances.
+
+    Serves namespacing for capacity reservations and provides methods for creating and cancelling reservations."""
+    reservations: Dict[str, Dict[str, str]] = {}
+
+    @staticmethod
+    def _get_cr_request_based_on_sct_config(params) -> Tuple[dict[str, int], int]:
+        instance_counts = {}
+        nemesis_node_count = params.get("nemesis_add_node_cnt") or 0
+        cluster_max_size = (params.get("cluster_target_size") or params.get("n_db_nodes")) + nemesis_node_count
+        instance_counts[params.get("instance_type_db")] = instance_counts.get(
+            params.get("instance_type_db"), 0) + cluster_max_size
+        instance_counts[params.get("instance_type_loader")] = instance_counts.get(params.get("instance_type_loader"),
+                                                                                  0) + params.get("n_loaders")
+        # don't reserve capacity for monitor - as usually it's not a problem to spin it
+        duration = params.get("test_duration") + 60  # 60 to have margin for test setup
+        instance_counts = {k: v for k, v in instance_counts.items() if v > 0}  # remove 0 values
+        return instance_counts, duration
+
+    @classmethod
+    def get_cr_from_aws(cls, params) -> None:
+        """Retrieves capacity reservations for given test_id from AWS."""
+        if not cls.is_capacity_reservation_enabled(params):
+            LOGGER.info("Capacity reservation is not enabled. Skipping reservation.")
+            return
+        test_id = params.get("reuse_cluster") or params.get("test_id")
+        ec2 = boto3.client('ec2', region_name=params.region_names[0])
+        reservations = ec2.describe_capacity_reservations(
+            Filters=[
+                {
+                    'Name': 'tag:test_id',
+                    'Values': [test_id]
+                },
+                {
+                    'Name': 'state',
+                    'Values': ['active']
+                }
+            ]
+        )
+        result = {}
+        availability_zone = params.get("availability_zone")
+        for reservation in reservations['CapacityReservations']:
+            availability_zone = reservation['AvailabilityZone'][-1]
+            instance_type = reservation['InstanceType']
+            if availability_zone not in result:
+                result[availability_zone] = {}
+            result[availability_zone][instance_type] = reservation['CapacityReservationId']
+        if result:
+            LOGGER.info("Found capacity reservations: %s", result)
+            params["availability_zone"] = availability_zone
+        else:
+            LOGGER.info("No capacity reservations found.")
+        cls.reservations = result
+
+    @staticmethod
+    def _get_supported_availability_zones(ec2, instance_types: List[str], initial_az: str) -> List[str]:
+        response = ec2.describe_instance_type_offerings(
+            LocationType='availability-zone',
+            Filters=[
+                {
+                    'Name': 'instance-type',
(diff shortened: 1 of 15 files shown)


soyacz:
See this is ignoring AZ setting and it creates CR after spinning sct-runner so possibly runner is in different AZ.
Do we accept this caveat or I should change it?

fruch:
> See this is ignoring AZ setting and it creates CR after spinning sct-runner so possibly runner is in different AZ.
> Do we accept this caveat or I should change it? 

It would work, but network cost of logging is a bit concerning 

We can start experimenting like that, but eventually I think we do want to include the runner in our reservation

soyacz:
grow shrink cluster test passed - It created CR (and used initial AZ as preferred one) and later used it twice during the test. It was created for duration of 2.5h hour but was cancelled in teardown (and later tried to be cleaned up in cleanup jenkins step).
Now I'll try a problematic perf test to see if it helps us with our problems.

soyacz:
perf test uncovered missing feature - it should work with placement group.

sdcm/provision/aws/capacity_reservation.py
```diff
@@ -0,0 +1,236 @@
...
+from datetime import datetime, timedelta
+from typing import List, Dict, Tuple
+
+from botocore.exceptions import ClientError
+import boto3
+
+LOGGER = logging.getLogger(__name__)
+
+
+# Shlould runner be created in the same az always? - in that case CR needs to be created before sct-runner $0.01/GB data transfer price
```
fruch, line 23: I think we should include SCT

soyacz: in that case, we'll need to split provision step to 2: spin up placement_group and reserve capacity and then create runner then provision step may happen (as need runner_ip to set up syslogng).
Quite complcated - i propose doing it in followup task

fruch:
@vponomaryov @juliayakovlev 

can you give @soyacz a hand in review this one ?

fruch (commented):
LGTM

sdcm/provision/aws/capacity_reservation.py
```diff
@@ -0,0 +1,241 @@
...
+
+
+class CapacityReservationError(Exception):
```
vponomaryov, line 27: We have `sdcm/exceptions.py` module for such things.

sdcm/provision/aws/capacity_reservation.py
```diff
@@ -0,0 +1,241 @@
...
+        instance_counts = {k: v for k, v in instance_counts.items() if v > 0}  # remove 0 values
+        return instance_counts, duration
+
+    @classmethod
+    def get_cr_from_aws(cls, params) -> None:
+        """Retrieves capacity reservations for given test_id from AWS."""
+        if not cls.is_capacity_reservation_enabled(params):
+            LOGGER.info("Capacity reservation is not enabled. Skipping reservation.")
+            return
+        test_id = params.get("reuse_cluster") or params.get("test_id")
```
vponomaryov, line 57: Are you sure about the `reuse_cluster` value?
Have you really verified this case?

sdcm/provision/aws/capacity_reservation.py
```diff
@@ -0,0 +1,241 @@
...
+        """
+        if not cls.is_capacity_reservation_enabled(params):
+            LOGGER.info("Capacity reservation is not enabled. Skipping reservation.")
+            return
+        cls.get_cr_from_aws(params)
+        if cls.reservations:
+            LOGGER.info("Capacity reservation already created. Skipping reservation.")
+            return
+        region = params.region_names[0]
+        test_id = params.get("reuse_cluster") or params.get("test_id")
```
vponomaryov, line 131: Same here.

sdcm/provision/aws/capacity_reservation.py
soyacz: `reuse_cluster` is taken into account at later stages of a test - not in `sct_config` where `get_cr_from_aws` is used.
This could lead to miss used capacity in case of reusing cluster.

I don't expect to use capacity reservations and reusing cluster case anyway, this is just a form of protection

soyacz: reuse_cluster holds test_id when is used

there's no point to reuse cluster and set capacity reservation to true - if cluster was created (so it can be reused), there's no point in setting it.

vponomaryov: No one is going to keep in mind to change some config options.
People will just reuse config files.
So, argument `no point` will be violated in practice.

fruch: Reuse cluster in this case should use the reservation that should exist , and continue to the step

We need just to validate that exactly what's gonna happen

And that reuse works, we don't test the reuse cases regularly, we should take the opportunity 😉

An idea, add reuse case to on or all of the provision tests

soyacz: this looks like separate feature to add it to SCT test pipeline. Also would need to add capacity_reservation setting to be able to test it with it.
I'll do tests manually.

soyacz: I verified it manually - works.
Need to remember that CR is for specified time duration (max test duration + 1h), so it becomes unusable after that time (but test will work, as default behavior is to skip CR params if CR's are not found for given test). It will cause problems if CR caused fallback to another AZ - in that case need to set in test config AZ manually when reusing cluster.

sdcm/provision/aws/capacity_reservation.py
fruch: o.k. please create the task for it.
we'll need to see if we can track the cost

soyacz: https://github.com/scylladb/scylla-cluster-tests/issues/7398

vponomaryov:
@soyacz Need to rebase it on top of the latest master.

sdcm/sct_config.py
```diff
@@ -1927,6 +1931,8 @@ def __init__(self):
         if self.get("k8s_enable_sni") and not self.get("k8s_enable_tls"):
             raise ValueError("'k8s_enable_sni=true' requires 'k8s_enable_tls' also to be 'true'.")
 
+        SCTCapacityReservation.get_cr_from_aws(self)
```
vponomaryov, line 1934: Shouldn't we raise an error if this new option is set to `true` and the cluster backend is not AWS?

fruch: Like any other backend specific configuration, we should make sure we ignore it

For example we don't fail if some k8s settings are set, or when having a case that specific aws instance types while running on GCP.

We should add validation only if it's common gotcha, not something that would force us to split to multiple configuration or multiple pipelines

vponomaryov: K8S settings have the "k8s" prefix or "gke" or "eks".
This one doesn't. If it is expected to be applied only to aws then need to add appropriate prefix.

soyacz: still we have aws/gce/etc specific options that don't fail, e.g. `data_volume_disk_type` or `use_placement_groups`

sdcm/provision/aws/capacity_reservation.py
soyacz: fixed

sdcm/sct_config.py
fruch: the a cosmetic issue, regarding configuration name prefixes.
one can later search and replace those.

the logic is that we don't introduce validation just for the sake of it.

I don't think this case needs one.

PR #7291, opened 2024-03-21
fix(test_default_filters): take it out of unittest

fruch:
since this test is depened on github API calls, and we are hitting the rate-limitting for them a bit too often now.

moving this test to integration test, and removeing the event default filter code from the test code,
so non of the test would be using it.

### Testing

- [ ]

### Reminders


vponomaryov (approved):
LGTM

PR #7293, opened 2024-03-24
refactor(tablets/longevity-100gb-4h): Disable hinted handoff

ShlomiBalalis:
Since it is disabled in pretty much every customer cluster, there's no point in leaving it enabled in all but one scenario. Hence, I'm disabling it in the tablets 4h 100gb scenario

### Testing

- [ ]

### Reminders

### DIFF
--- a/configurations/disable_hinted_handoff.yaml
+++ b/configurations/disable_hinted_handoff.yaml
@@ -1,0 +1,1 @@
+hinted_handoff: 'disabled'
--- a/jenkins-pipelines/oss/tablets/longevity-100gb-4h.jenkinsfile
+++ b/jenkins-pipelines/oss/tablets/longevity-100gb-4h.jenkinsfile
@@ -8,6 +8,6 @@ longevityPipeline(
     region: 'eu-west-1',
     availability_zone: 'c',
     test_name: 'longevity_test.LongevityTest.test_custom_time',
-    test_config: '''["test-cases/longevity/longevity-100gb-4h.yaml", "configurations/tablets.yaml", "configurations/tablets_supported_nemeses.yaml"]''',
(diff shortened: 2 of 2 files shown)


roydahan:
AFAIR we disabled it by default to all test cases.
Did we leave it specifically set for this one?

if so, did you see a specific issue with HH that made you disabling it?

ShlomiBalalis:
> AFAIR we disabled it by default to all test cases. Did we leave it specifically set for this one?

Yes, you disabled it in all tests except 4h 100gb, which I had the bad luck of using for tablets. I thought it'd be better to edit it in that test case rather than switching to a new one.

> if so, did you see a specific issue with HH that made you disabling it?

Yes, the usual message: `was overloaded: Too many in flight hints`

roydahan:
> > AFAIR we disabled it by default to all test cases. Did we leave it specifically set for this one?
> 
> Yes, you disabled it in all tests except 4h 100gb, which I had the bad luck of using for tablets. I thought it'd be better to edit it in that test case rather than switching to a new one.
> 
> > if so, did you see a specific issue with HH that made you disabling it?
> 
> Yes, the usual message: `was overloaded: Too many in flight hints`

It’s unusual for this test, so I wouldn’t just dismiss it and have someone to look at the overload problem itself - the hints are probably just the symptom.

We can disable it here, but once tablets gets into master as default we will bump into the same issue and we wouldn’t just dismiss it by disabling HH.

fruch:
@ShlomiBalalis 

Has something changed with this? Do we still need this PR?

yarongilor:
i think it's not relevant after tablets became enabled by default.

fruch:
> i think it's not relevant after tablets became enabled by default.

if this case is now stable with tablets, then yes
information would be helpful

fruch:
doesn't seem it's need anymore, i.e. @ShlomiBalalis didn't response, and @yarongilor suggested it's not

PR #7294, opened 2024-03-24
feature(monitoring): set monitoring branch to 4.7

amnonh:
- [x ] I didn't leave commented-out/debugging code

### DIFF
--- a/defaults/aws_config.yaml
+++ b/defaults/aws_config.yaml
@@ -9,7 +9,7 @@ instance_type_monitor: 't3.large'
 # manual on creating loader AMI's: see docs/new_loader_ami.md
 ami_id_loader: 'scylla-qa-loader-ami-v20-ubuntu22'
 # manual on updating monitor images see: docs/monitoring-images.md
-ami_id_monitor: 'scylladb-monitor-4-6-2-2024-02-13t08-06-04z'
+ami_id_monitor: 'scylladb-monitor-4-7-2-2024-05-13t08-38-47z'
 
 availability_zone: 'a'
 root_disk_size_monitor: 50  # GB, remove this field if default disk size should be used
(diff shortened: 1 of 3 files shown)


fruch:
I've mark 2 provision test to run

we'll check their monitoring output, and if o.k., we can merged this one

fruch:
@amnonh 

why we don't have images for this branch ? (i.e. in other PR i'm trying out the images, and I only see 4.6.2)

fruch:
our dashboard becomes a bit more broken with this version:
image

for some reason the default view show only the metric name, and not the values (we'll need to fix that)

fruch (dismissed):
LGTM

roydahan:
I don’t want to merge it if it breaks our view

soyacz:
os monitoring also doesn't work properly. I could see only one instance metrics. Tested on branch-perf-v14:
image

roydahan:
@fruch please correct me if I'm wrong, but now that you moved to monitor images, it's not enough to just change the monitor branch.

fruch:
> @fruch please correct me if I'm wrong, but now that you moved to monitor images, it's not enough to just change the monitor branch.
> 

We haven't yet merged it, but yes it changes a bit this flow of updating, hence why I've asked here where are the images

soyacz:
@amnonh possibly problem I'm seeing is due missing backport of https://github.com/scylladb/scylla-cluster-tests/pull/6975 to branch-perf-v14

fruch:
> our dashboard becomes a bit more broken with this version: image
> 
> for some reason the default view show only the metric name, and not the values (we'll need to fix that)

@amnonh can you help us figure this one ? its what holding us from merging this (and the fact we now need to specify also the images for 4.7)

fruch:
@yaronkaikov 

can you clarify where the monitoring images should be found ? on which account ? 
and why aren't they public ?

fruch:
I've check the latest runs of this PR, and the issue with the nemesis panel isn't there anymore:

image


once I'll find the right image for GCE, this is good to go

fruch (approved):
LGTM

PR #7297, opened 2024-03-25
test(longevity): add longevity with rf=1

juliayakovlev:
This PR is trying to cover customer issue https://github.com/scylladb/scylladb/issues/16759.
This is from the issue description:
```
After Upgrading Scylla from 5.2.11 via 5.4.0 to 5.4.1, we started observing missing data in Scylla. 
Every once in a while, an INSERT is committed successfully but the inserted values are not visible 
until Scylla is restarted. 
As far as we can tell, version 5.2.11 was not affected while 5.4.1 is. 5.4.0 was not long enough 
in operation to make a reliable statement.

The suspected bug occurs extremely rarely, making it hard for us to reproduce. 
Out of ~800M inserts per day, only a dozen are affected.
```

The load is run on keyspace with RF 1 SCT test. 
We use scylla-bench because its validate data better then in cassandra-stress

### Testing

- [x] [longevity-1-rf-test-12h](https://argus.scylladb.com/test/e6ef2a1c-7267-4566-9954-f49caa755c50/runs?additionalRuns[]=e45799d2-84a2-4126-b279-720a37f5c17d) - Scylla version `5.4.1`
- [x] [longevity-1-rf-test-12h](https://argus.scylladb.com/test/e6ef2a1c-7267-4566-9954-f49caa755c50/runs?additionalRuns[]=9c2c48ba-045b-4ef3-90c4-268beca319f3) - Scylla version `5.4.6`

### Reminders

### DIFF
--- a/jenkins-pipelines/oss/longevity/longevity-1-rf-test-12h.jenkinsfile
+++ b/jenkins-pipelines/oss/longevity/longevity-1-rf-test-12h.jenkinsfile
@@ -1,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    backend: 'aws',
+    region: 'eu-west-1',
+    availability_zone: 'c',
+    test_name: 'longevity_test.LongevityTest.test_custom_time',
+    test_config: 'test-cases/longevity/longevity-1-rf-test-12h.yaml',
+    instance_provision_fallback_on_demand: true
+)
--- a/test-cases/longevity/longevity-1-rf-test-12h.yaml
+++ b/test-cases/longevity/longevity-1-rf-test-12h.yaml
@@ -1,0 +1,33 @@
+test_duration: 800
+
+prepare_write_cmd:  ["scylla-bench -workload=sequential -mode=write -replication-factor=1 -partition-count=100 -clustering-row-count=55555                       -clustering-row-size=uniform:1024..2048 -concurrency=10 -connection-count=10 -consistency-level=one -rows-per-request=10 -timeout=30s -validate-data" ,
+                     "scylla-bench -workload=sequential -mode=write -replication-factor=1 -partition-count=100 -clustering-row-count=55555 -partition-offset=101 -clustering-row-size=uniform:1024..2048 -concurrency=10 -connection-count=10 -consistency-level=one -rows-per-request=10 -timeout=30s -validate-data",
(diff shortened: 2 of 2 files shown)


sdcm/cluster.py
```diff
@@ -3586,19 +3586,29 @@ def cql_connection_patient_exclusive(self, node, keyspace=None,
...
+                           filter_by_keyspace: list = None, filter_out_by_replication_factor: int | None = None) -> List[str]:
         regular_column_names = ["keyspace_name", "table_name"]
         materialized_view_column_names = ["keyspace_name", "view_name"]
         regular_table_names, materialized_view_table_names = set(), set()
         where_clause = ""
+        # This filter is usefull when the some
+        if filter_out_by_replication_factor:
+            with self.cql_connection_patient(db_node, connect_timeout=600) as session:
+                result = list(session.execute("select keyspace_name, replication from system_schema.keyspaces"))
+                filter_by_keyspace = [row.keyspace_name for row in result
```
fruch, line 3607: logic here is a bit weird:
for example `filter_out_by_replication_factor=2`, it would pick KS with rf=1 and KS with rf=3

I would think it's maybe bigger than 1, or bigger then 2

also I would put the logic of retrieving the RF, on it's own utility function, and run this at the end, after all the other filters

fruch: overwriting `filter_by_keyspace` meaning you can use those options at the same time..

juliayakovlev: I cannot think about another example when this filter may be needed. 
If it may be case when we want to filter out RF = 2, but RF 1 or 3 is OK. What is the reason to filter out RF 1 together with RF 2?

fruch: we don't have such a case now, but if we introduce a feature into our testing, we should strive for it to be concise and clear logic.

we don't have cases that used RF=2 now, but we didn't had one that use RF=1 till this one.
one possible use case, if for when we need to change RF, before decommission. we should be able to look for the keyspaces which have `RF=number of nodes` so we can ALTER it. (see https://github.com/scylladb/scylla-cluster-tests/pull/7457 as an example)

juliayakovlev: Here it is another case. 
It is opposite: we need to find any keyspace that RF is not equal (actually, pretty clear logic). So logic "select that more then" or "equal" is not suitable for this case.
If you do not like the solution from this commit, I can create additional function that go over received function and filter them again (that will need additional query of DB)

juliayakovlev:
@fruch As we discussed, I removed the filter out keyspace by replication factor from this PR.
So in the this new longevity expected `disrupt_toggle_table_gc_mode` failure with error:
```
ConfigurationException: tombstone_gc option with mode = repair not supported for table with RF one or local replication strategy
```

It will be solved in the separate PR

juliayakovlev:
@fruch this PR is ready for merge. Please, review

fruch (approved):
LGTM

PR #7298, opened 2024-03-25
improvement(tooling_reporter.py): Report cassandra-stress version once

### LINKED ISSUE #7276: cassandra-stress version should be reported to Argus only once
image

we should report each version only once,
also we should split it, and report the java-driver on it's own row
the `(with java-driver 3.11.5.2)` can go into its own field (new or current field)

k0machi:
This commit adds caching for cassandra-stress versions (drivers
included)

Fixes #7276

### Testing

- [ ] [Argus](#)
- [ ] [Jenkins](#)

### Reminders

### DIFF
--- a/sdcm/reporting/tooling_reporter.py
+++ b/sdcm/reporting/tooling_reporter.py
@@ -1,4 +1,5 @@
 import logging
+from functools import lru_cache
 
 from cassandra import __version__ as PYTHON_DRIVER_VERSION
 from argus.client.sct.client import ArgusSCTClient
@@ -9,6 +10,13 @@ from sdcm.remote.base import CommandRunner
 LOGGER = logging.getLogger(__name__)
 
 
+@lru_cache
+def report_package_to_argus(client: ArgusSCTClient, tool_name: str, package_version: str, additional_data: str = None):
+    package = Package(name=f"{tool_name}", version=package_version,
+                      date=None, revision_id=None, build_id=additional_data)
+    client.submit_packages([package])
+
+
 class ToolReporterBase():
 
     TOOL_NAME = None
@@ -17,6 +25,7 @@ class ToolReporterBase():
         self.runner = runner
         self.command_prefix = command_prefix
         self.argus_client = argus_client
+        self.additional_data = None
         self.version: str | None = None
 
     def __str__(self) -> str:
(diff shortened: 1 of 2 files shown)


sdcm/reporting/tooling_reporter.py
```diff
@@ -85,3 +96,20 @@ def _collect_version_info(self) -> None:
         self.version = f"{result.get('cassandra-stress', '#FAILED_CHECK_LOGS')}"
         if driver_version := result.get("scylla-java-driver"):
             self.version += f" (with java-driver {driver_version})"
+            self.driver_version = driver_version
+
+        cached_driver_versions: list[str] = self.VERSION_CACHE.get(self.version, [])
+        if self.driver_version:
+            cached_driver_versions.append(self.driver_version)
+        self.VERSION_CACHE[self.version] = cached_driver_versions
```
soyacz, line 104: IIUC  we populate cache and then we verify if it's inside the cache - won't `_is_duplicate_version` always return True?

soyacz (approved):
LGTM if proven it works

sdcm/reporting/tooling_reporter.py
fruch: I've discussed it with @k0machi offline, and we should be trying to use `@cache` instead to make this work as expected

k0machi: > IIUC we populate cache and then we verify if it's inside the cache - won't `_is_duplicate_version` always return True?

the idea is to always return True when there's a version in the cache but no driver - meaning we are using cassandra stress that doesn't report its driver version, but I'm planning to rework this a bit into using lru_cache instead on report method, so it's going to be work a bit differently anyway

fruch:
@k0machi 

what's up with this one ?

k0machi:
> @k0machi
> 
> what's up with this one ?

Running a [test job](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/alexey-argus-testing/80/) now, replaced the code with just `lru_cache` for argus reporting itself.

k0machi:
@fruch Added separate java-driver version (and now reusing build_id for additional information for cassandra stress, running tests now)

fruch (approved):
LGTM

PR #7300, opened 2024-03-25
fix(_kill_scylla_daemon): ignore hints_manager's segment trunction error

ShlomiBalalis:
When Scylla is killed ungracefully, like in _kill_scylla_daemon, there's a chance that the hints manager will fail to truncate one of its segments. As such, I added an ignore around the ungraceful kill at _kill_scylla_daemon

### Testing

- [ ]

### Reminders


ShlomiBalalis:
> Can you point to the issue this is happening in ? In which version and such ?
> 
> How many times it was observed?

This is the first time I've seen it, so I can only assume it's new. I did not open an issue on the subject since @bhalevy confirmed that the error is to be expected

fruch:
> > Can you point to the issue this is happening in ? In which version and such ?
> > How many times it was observed?
> 
> This is the first time I've seen it, so I can only assume it's new. I did not open an issue on the subject since @bhalevy confirmed that the error is to be expected

we do need some breadcrumbs, to understand why we did something and whom took the decision.
and reference to when it's started happening (i.e. at least the job it was first seen in)

roydahan:
Or a Scylla issue that claims this error is expected.
And if it does, why is it an error?

pehala:
Closing this, if we need this in the future we can reopen

Issue #7307, opened 2024-03-26
'smp' scylla command line argument has no effect when cluster is created on docker backend


dimakr:
### Issue description

Even though a specific value can be set for `--smp` option when starting scylla from command line, it has no effect if cluster is created on docker backend. The created scylla containers use only one host CPU core as this value is currently hardcoded: https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/cluster_docker.py#L60

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

#### Steps to Reproduce

1. Start a test on docker backend with a scenario that puts substantial load onto the cluster
```
export SCT_APPEND_SCYLLA_ARGS=' --smp 2 --memory 4G'
export SCT_CONFIG_FILES='["configurations/nemesis/longevity-5gb-1h-nemesis.yaml","configurations/nemesis/AbortRepairMonkey.yaml"]'
./sct.py run-test longevity_test.LongevityTest.test_custom_time --backend docker`
```
2. Wait for the stress commands to start
3. Run `docker stats --no-stream --format "table {{.Name}}\t{{.CPUPerc}}"` command to see CPUs utilization by containers

**Expected behavior:** CPU utilization by db-node containers is (up to) 200%. 

**Actual behavior:** CPU utilization by db-node containers is ~100% indicating that the containers utilize only one core.
```
longevity-5gb-1h-AbortRepairMonkey--db-node-e9201320-2                  100.76%
longevity-5gb-1h-AbortRepairMonkey--db-node-e9201320-1                  103.39%
longevity-5gb-1h-AbortRepairMonkey--db-node-e9201320-0                  105.97%
```

### Impact

The issue prevents from utilizing available CPU resources of a local machine, when running tests on docker backend locally.
It could be also required to throttle the stress commands load, otherwise the test can fail due to docker container performance degradation.

### How frequently does it reproduce?

When tests are started on docker backend.

### Installation details

SCT Version: master
Scylla version: 5.4.4
Environment: local machine with 20 cores.
Test config: [longevity-5gb-1h-nemesis.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/master/configurations/nemesis/longevity-5gb-1h-nemesis.yaml)
Additional test config parameters:
```use_mgmt: false
monitor_swap_size: 0

send_email: false
enable_argus: false

server_encrypt: false
client_encrypt: false
```

fruch:
[...]

PR #7308, opened 2024-03-26
improvement(cluster-docker): start container with custom number of cpus

dimakr:
So far containers of Docker backend cluster were started with hardcoded number of cpus, which is 1. This resulted in containers of scylla cluster nodes being overloaded during c-s and/or s-b stress commands execution.

The commit allows to start the containers with the specified number of cpus. The number is based on "smp" option value from a test config.

The commit addresses issue scylladb/scylla-cluster-tests#7307.

### Testing
- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster_docker.py
+++ b/sdcm/cluster_docker.py
@@ -52,12 +52,18 @@ class NodeContainerMixin:
         volumes = {
             '/var/run/docker.sock': {"bind": '/var/run/docker.sock', "mode": "rw"},
         }
+
+        # TODO: add parsing of cpuset option from append_scylla_args parameter
+        scylla_args = self.parent_cluster.params.get('append_scylla_args')
+        smp_match = re.search(r'--smp\s(\d+)', scylla_args)
+        smp = int(smp_match.group(1)) if smp_match else 1
+
         return dict(name=self.name,
                     image=self.node_container_image_tag,
                     command=f'--seeds="{seed_ip}"' if seed_ip else None,
                     volumes=volumes,
                     network=self.parent_cluster.params.get('docker_network'),
-                    nano_cpus=10**9)  # Same as `docker run --cpus=1 ...' CLI command.
+                    nano_cpus=smp*10**9)  # Same as `docker run --cpus=N ...' CLI command.
 
 
 class DockerNode(cluster.BaseNode, NodeContainerMixin):  # pylint: disable=abstract-method


sdcm/cluster_docker.py
```diff
@@ -52,12 +52,18 @@ def node_container_run_args(self, seed_ip):
         volumes = {
             '/var/run/docker.sock': {"bind": '/var/run/docker.sock', "mode": "rw"},
         }
+
+        # TODO: add parsing of cpuset option from append_scylla_args parameter
```
fruch, line 56: Why the TODO, looks like you are doing just that

sdcm/cluster_docker.py
```diff
@@ -52,12 +52,18 @@ def node_container_run_args(self, seed_ip):
...
+        scylla_args = self.parent_cluster.params.get('append_scylla_args')
```
fruch, line 57: We are now doing it for all nodes, including loader nodes, so we might want to limit it per node type, and do it just for the db one for now, and think how it would be better to control each one on its own

fruch (commented):
Remove the TODO

sdcm/cluster_docker.py
```diff
@@ -52,12 +52,18 @@ def node_container_run_args(self, seed_ip):
...
+        smp_match = re.search(r'--smp\s(\d+)', scylla_args)
```
vponomaryov, line 58: This solution looks not optimal.

Maybe do following:
- Read number of cores (CORE_NUM) on the node where docker runs
- Read "n_db_nodes" value (i.e. '3') add "1" (CONT_NUM) to allow to run "extend cluster" kind of nemesis.
- Decide what is the CPU usage per container by diving CORE_NUM by CONT_NUM
  * if it is less than 2, apply 1.
  * if it is not round like "2.0" then use integer part.

soyacz:
I'd propose to limit cpu based on instance types of test config. e.g. i4i.large with 2 cpus should set this to 2cpu cores.
the same for loaders.
This way we could mimic what is run in the wild (of course it won't be the same, but at least closer).
In case local machine is too small, should raise warning (or even ERROR event) to clearly show too little resources for given test.

sdcm/cluster_docker.py
dimakr: > Why the TODO, looks like you are doing just that

This is about cpuset parameter that can be passed to scylla. In this case the variability of values for the parameter require smore complex logic to parse it. Like here https://github.com/scylladb/scylla-cluster-tests/blob/4121b79f03e891dc1503f536fdb2866c4f1e37e8/sdcm/cluster.py#L549

sdcm/cluster_docker.py
vponomaryov: My proposal is based exactly on the idea of all-in one solution.

fruch: I don't think we should do calculation for this case, it would be really problematic, and we won't know the exact setup that was used for a passing test.

there's not real need for it to be "production" like, just that we'll have reference of cases that works for our needs
and that we can configure what needed

fruch:
> I'd propose to limit cpu based on instance types of test config. e.g. i4i.large with 2 cpus should set limits to 2cpu cores. the same for loaders. This way we could mimic what is run in the wild (of course it won't be the same, but at least closer). In case local machine is too small, should raise warning (or even ERROR event) to clearly show too little resources for given test.

I agree in general that we should do something equivalent to the case we have working. I don't think we should automated it, like we don't automated picking instance types for GCP base on AWS instances. picking the parameters has lot of variables to consider, automating for this case is not needed

sdcm/cluster_docker.py
fruch: I don't think we need to support it, it means we'll want to do CPU pinning, and that type of tuning it not in the scope of what we need from the docker backend

sdcm/cluster_docker.py
fruch: I think we are o.k. with limiting this solution to db nodes only for now.

We need to adapt cases to docker backend anyhow regardless, cause of multiple reasons.

If we'll found we need a way to define number of CPU for loader, we should add it to SCT configuration and align it

dimakr: Changed the current logic to set smp value from scylla_args for db-nodes only. Other nodes get smp=1. 

This is an example of running this locally:
```
❯ grep smp configurations/nemesis/additional_configs/docker_backend_local.yaml
append_scylla_args: '--smp 4 --memory 2G'
❯ docker ps -q | xargs -I {} docker inspect --format='{{.Name}}: {{.HostConfig.NanoCpus}}' {} | awk '{print $1 ": " $2 / 1000000000 " CPUs"}'

/cool_allen:: 0 CPUs
/agraf:: 0 CPUs
/agrafrender:: 0 CPUs
/aprom:: 0 CPUs
/promtail:: 0 CPUs
/loki:: 0 CPUs
/aalert:: 0 CPUs
/longevity-1gb-1h-nemesis-dmitriy-loader-node-949e8078-0:: 1 CPUs
/longevity-1gb-1h-nemesis-dmitriy-db-node-949e8078-2:: 4 CPUs
/longevity-1gb-1h-nemesis-dmitriy-db-node-949e8078-1:: 4 CPUs
/longevity-1gb-1h-nemesis-dmitriy-db-node-949e8078-0:: 4 CPUs
```

As an observation, even with the rate limited stress commands for docker backend CPU utilization by db nodes is higher than 1. So the change can be helpful if we want to avoid unintentional throttling of db nodes performance on docker backend.
```
docker stats --no-stream --format "table {{.Name}}\t{{.CPUPerc}}\t{{.MemUsage}}\t{{.MemPerc}}"
NAME                                                      CPU %     MEM USAGE / LIMIT   MEM %
wizardly_montalcini                                       78.31%    119.8MiB / 31GiB    0.38%
friendly_taussig                                          44.65%    587.8MiB / 31GiB    1.85%
modest_varahamihira                                       43.13%    564.6MiB / 31GiB    1.78%
agraf                                                     0.05%     229.2MiB / 31GiB    0.72%
agrafrender                                               0.00%     107.1MiB / 31GiB    0.34%
aprom                                                     0.00%     198.1MiB / 31GiB    0.62%
promtail                                                  0.24%     78.18MiB / 31GiB    0.25%
loki                                                      0.23%     97.41MiB / 31GiB    0.31%
aalert                                                    0.48%     39.58MiB / 31GiB    0.12%
longevity-1gb-1h-nemesis-dmitriy-loader-node-949e8078-0   0.72%     264.3MiB / 31GiB    0.83%
longevity-1gb-1h-nemesis-dmitriy-db-node-949e8078-2       142.32%   1.864GiB / 31GiB    6.01%
longevity-1gb-1h-nemesis-dmitriy-db-node-949e8078-1       167.39%   1.933GiB / 31GiB    6.24%
longevity-1gb-1h-nemesis-dmitriy-db-node-949e8078-0       266.26%   2.059GiB / 31GiB    6.64%
```

vponomaryov (approved):
LGTM

PR #7321, opened 2024-04-05
feature(teardown-validators): nodetool scrub validate

soyacz:
We want to validate scylla cluster after longevity test to verify if it's in correct state. Aiming for checks that require no load and/or affect system performance and/or can be done once after the test.

Implemented first validator that verifies if sstables are not corrupted using `nodetool scrub`. In case corruption is found, quarantined by scrub sstables are uploaded to s3 and proper error event is raised with link.

refs: https://github.com/scylladb/qa-tasks/issues/1577

### Testing

- [X] -  longevity (kill scylla nemesis) - https://argus.scylladb.com/test/38256c5f-88a2-4356-b777-c655de35e9cb/runs?additionalRuns[]=6dd23b42-74d7-436e-aa2a-80b50c959a40

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -242,3 +242,11 @@ validate_large_collections: false
 scylla_d_overrides_files: []
 
 run_commit_log_check_thread: true
+
+# teardown validators
+teardown_validators:
+  scrub:
+    enabled: false
+    timeout: 1200
+    keyspace: ''
+    table: ''
--- a/sdcm/teardown_validators/sstables.py
+++ b/sdcm/teardown_validators/sstables.py
@@ -1,0 +1,68 @@
+import logging
+from functools import partial
+
+from sdcm.cluster import BaseNode
+from sdcm.sct_events import Severity
+from sdcm.sct_events.teardown_validators import ValidatorEvent, ScrubValidationErrorEvent
+from sdcm.teardown_validators.base import TeardownValidator
+from sdcm.utils.common import S3Storage, ParallelObject
+from sdcm.utils.s3_remote_uploader import upload_remote_files_directly_to_s3
+
+LOGGER = logging.getLogger(__name__)
+
+
+class SstablesValidator(TeardownValidator):  # pylint: disable=too-few-public-methods
+    validator_name = 'scrub'
+
+    def _upload_corrupted_files(self, node: BaseNode, quarantine_log_lines):
+        # get quarantine dir from lines:
+        # INFO  2024-04-02 12:40:24,787 [shard 0:stre] sstable - Moving sstable /var/lib/scylla/data/system_schema/columns-24101c25a2ae3af787c1b40ee1aca33f/me-3gey_0z3c_2h5vl2ogebmg26ku9t-big-Data.db to "/var/lib/scylla/data/system_schema/columns-24101c25a2ae3af787c1b40ee1aca33f/quarantine"  # pylint: disable=line-too-long
+        # print(log_lines)
+        quarantine_dirs = {line.split(' to "')[1][:-2]
+                           for line in quarantine_log_lines
+                           if 'sstable - Moving sstable' in line and 'quarantine' in line}
+        if not quarantine_dirs:
+            LOGGER.error('No quarantine directories found in db logs on node: %s', node.name)
+            return "<No quarantine directories found>"
+        s3_link = upload_remote_files_directly_to_s3(
+            node.ssh_login_info, list(quarantine_dirs), s3_bucket=S3Storage.bucket_name,
+            s3_key=f"{self.cluster.uuid}/{node.name}-corrupted-sstables.tar.gz",
+            max_size_gb=100, public_read_acl=True)
+        return s3_link
+
+    def _run_nodetool_scrub(self, node: BaseNode, keyspace: str, table: str, timeout=1200):
+        node.wait_db_up(timeout=300)
+        finish_scrub_follower = node.follow_system_log(patterns=['Finished scrubbing in validate mode'])
+        quarantine_lines = node.follow_system_log(patterns=['sstable - Moving sstable'], start_from_beginning=True)
+        result = node.run_nodetool(sub_cmd='scrub', args=f"--mode VALIDATE --no-snapshot {keyspace} {table}".strip(),
+                                   timeout=timeout, coredump_on_timeout=False, ignore_status=True)
+        if not result.ok:
+            ValidatorEvent(
+                message=f'Error running nodetool scrub on node {node.name}: {result.stdout}\n{result.stderr}',
+                severity=Severity.ERROR).publish()
+        scrub_finish_lines = list(finish_scrub_follower)
+        if not scrub_finish_lines:
+            ValidatorEvent(
+                message=f'No scrubbing validation message found in db logs on node: {node.name}', severity=Severity.ERROR).publish()
+            return
+        invalid_sstables = [line for line in scrub_finish_lines if 'sstable is invalid' in line]
+        if invalid_sstables:
+            LOGGER.error("Invalid sstables found: %s.", invalid_sstables)
+            s3_link = self._upload_corrupted_files(node, list(quarantine_lines))
+            ScrubValidationErrorEvent(node.name, s3_link).publish()
+
+    def validate(self):
+        keyspace = self.configuration.get("keyspace")
+        table = self.configuration.get("table")
+        timeout = self.configuration.get("timeout", 1200)
+        run_scrub = partial(self._run_nodetool_scrub, keyspace=keyspace, table=table, timeout=timeout)
+        run_scrub.__name__ = run_scrub.func.__name__
+        try:
+            LOGGER.info("Running nodetool scrub on all nodes in validation mode")
+            parallel_obj = ParallelObject(objects=self.cluster.nodes, timeout=timeout)
+            parallel_obj.run(run_scrub, ignore_exceptions=False, unpack_objects=True)
+            LOGGER.info("Nodetool scrub validation finished")
+        except Exception as exc:  # pylint: disable=broad-except
+            LOGGER.error("Error during nodetool scrub validation: %s", exc)
+            ValidatorEvent(
+                message=f'Error during nodetool scrub validation: {exc}', severity=Severity.ERROR).publish()
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -65,6 +65,7 @@ from sdcm.scan_operation_thread import ScanOperationThread
 from sdcm.nosql_thread import NoSQLBenchStressThread
 from sdcm.scylla_bench_thread import ScyllaBenchThread
 from sdcm.cassandra_harry_thread import CassandraHarryThread
+from sdcm.teardown_validators import teardown_validators_list
 from sdcm.tombstone_gc_verification_thread import TombstoneGcVerificationThread
 from sdcm.utils.alternator.consts import NO_LWT_TABLE_NAME
 from sdcm.utils.aws_kms import AwsKms
@@ -2798,6 +2799,10 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         self.teardown_started = True
         with silence(parent=self, name='Sending test end event'):
             InfoEvent(message="TEST_END").publish()
+        self.log.info("Post test validators are starting...")
+        for validator_class in teardown_validators_list:
+            for db_cluster in self.db_clusters_multitenant:
+                validator_class(self.params, db_cluster).validate()
         self.log.info('TearDown is starting...')
         self.stop_timeout_thread()
         self.stop_event_analyzer()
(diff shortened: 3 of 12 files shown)


soyacz:
checkout [Argus](https://argus.scylladb.com/test/38256c5f-88a2-4356-b777-c655de35e9cb/runs?additionalRuns[]=6dd23b42-74d7-436e-aa2a-80b50c959a40) for error events (duplicate [#issue](https://github.com/scylladb/scylladb/issues/18152) - but I'm not sure if not found different error now, next time I'll rerun it on master:latest).
See the event on the top contains the link to malformed sstables (5.7GB)  - I added this feature as I believe investigating sstable issues require sstables itself. Several questions:

1. do we want to limit upload size e.g. by taking only subset?
2. maybe use better compression?
3. there are many more DB events caused by this scrub - shall we lower their severity (or filter it out completely) during that validation?

* disclaimer - result shows only one event due limit that is already taken out

soyacz:
> checkout [Argus](https://argus.scylladb.com/test/38256c5f-88a2-4356-b777-c655de35e9cb/runs?additionalRuns%5B%5D=6dd23b42-74d7-436e-aa2a-80b50c959a40) for error events (duplicate [#issue](https://github.com/scylladb/scylladb/issues/18152) - but I'm not sure if not found different error now, next time I'll rerun it on master:latest). See the event on the top contains the link to malformed sstables (5.7GB) - I added this feature as I believe investigating sstable issues require sstables itself. Several questions:
> 
> 1. do we want to limit upload size e.g. by taking only subset?
> 2. maybe use better compression?
> 3. there are many more DB events caused by this scrub - shall we lower their severity (or filter it out completely) during that validation?
> 
> * disclaimer - result shows only one event due limit that is already taken out

@roydahan can you help answering these questions?

roydahan:
> checkout [Argus](https://argus.scylladb.com/test/38256c5f-88a2-4356-b777-c655de35e9cb/runs?additionalRuns%5B%5D=6dd23b42-74d7-436e-aa2a-80b50c959a40) for error events (duplicate [#issue](https://github.com/scylladb/scylladb/issues/18152) - but I'm not sure if not found different error now, next time I'll rerun it on master:latest). See the event on the top contains the link to malformed sstables (5.7GB) - I added this feature as I believe investigating sstable issues require sstables itself. Several questions:
> 
> 1. do we want to limit upload size e.g. by taking only subset?

If the tool identifies the corrupted sstables, no need to limit and we should upload only the corrupted.

> 2. maybe use better compression?

Use the best compression you can find.

> 3. there are many more DB events caused by this scrub - shall we lower their severity (or filter it out completely) during 
that validation?

I see them and I don't know what they mean.
I wouldn't lower their severity or filter them out for now.
Please open issues for them (at least some) so we can get clarifications about them and later decide what to do with them.
> 
> * disclaimer - result shows only one event due limit that is already taken out

soyacz:
xz compression doesn't bring improvements, 
I think random nature of inserted data makes it hardly compressible.

fruch:
@soyacz 

Did you had a run with master ? To see the issue you have are really fixed one ?

I think we do want to target at least one tier1 case with this check

defaults/test_default.yaml
```diff
@@ -241,3 +241,6 @@ validate_large_collections: false
 scylla_d_overrides_files: []
 
 run_commit_log_check_thread: true
+
+# teardown validators
+teardown_validate_sstables: true
```
fruch, line 246: Having a boolean configuration might a bit naive

Maybe we should consider things like percentage of stable to scrub ?

Also maybe a configuration which is a list of possible validators to use, and not introduce multiple top level SCT configurations for each validator ? (I know it's the first one we introduce, but there would be more that we should be thinking of)

soyacz: > Having a boolean configuration might a bit naive

what else do we need here? Timeout?

> Maybe we should consider things like percentage of stable to scrub ?

I don't know the way to take percentage

> Also maybe a configuration which is a list of possible validators to use, and not introduce multiple top level SCT configurations for each validator ? (I know it's the first one we introduce, but there would be more that we should be thinking of)

I was thinking about that and was not sure, I took these factors when making decision:
1. when one test decides on disabling this one, then overriding one configuration would mean overriding all validators configuration if we used one config for all
2. different validators might need different configuration, so we are not stuck to naive config like here
3. putting in the same area in `test_default.yaml` will make easy to find possible validators (we could include also ones that are disabled by default)

But I'm open to different arugments :)

fruch: > > Having a boolean configuration might a bit naive
> 
> what else do we need here? Timeout?
> 
> > Maybe we should consider things like percentage of stable to scrub ?
> 
> I don't know the way to take percentage

don't you feed it sstables one by one ? I guess you could random select and feed the tool only part of the sstables ?

> 
> > Also maybe a configuration which is a list of possible validators to use, and not introduce multiple top level SCT configurations for each validator ? (I know it's the first one we introduce, but there would be more that we should be thinking of)
> 
> I was thinking about that and was not sure, I took these factors when making decision:
> 
> 1. when one test decides on disabling this one, then overriding one configuration would mean overriding all validators configuration if we used one config for all

that's not always try, for example see the what we use `stress_image`, we can replace specific one, without touch others, i.e. we can extend/overwrite inner structures. 

> 2. different validators might need different configuration, so we are not stuck to naive config like here

that's for sure is correct, so something similar to how `run_fullscan` is being used, i.e. with converting it to a dataclass
we know we gonna have parameters for each one, we should dine it as such from start

> 3. putting in the same area in `test_default.yaml` will make easy to find possible validators (we could include also ones that are disabled by default)

I don't think we gonna have any of them in test_default.yaml, not for a very long time, until we prove each one of them, we won't have them by default for all tests.

> 
> But I'm open to different arugments :)

soyacz: Ok, I agree I'll adapt accordingly. I'll check how `stress_image` works as I don't know yet mechanism behind merging dicts by sct_config. Anyway assuming these won't be defaults, we can specify test by test what validators are used.

As to scrub validation, we don't provide sstables, we can provide keyspace - this will be added to config param.

> Agree with @fruch Example of the new config option I would like to see:
> 
> ```
> teardown_validators:
>   nodetool_scrub: '{"mode": "VALIDATE", "timeout_per_node": 1200}'
> ```

I disagree to provide "mode" in config - we always want to validate. In case we'll need different modes, we'll add this.

fruch: We can have a percentage for the upload of broken stables, if we find uploading all of them is too big

soyacz: > We can have a percentage for the upload of broken stables, if we find uploading all of them is too big

Sounds like an good idea, can we do it in followup task?
Also would require some more thinking - e.g. what are the thresholds for selecting percentage and what percentage.

sdcm/teardown_validators/sstables.py
```diff
@@ -0,0 +1,61 @@
...
+        if not result.ok:
+            ValidatorEvent(
+                message=f'Error running nodetool scrub on node {node.name}: {result.stdout}\n{result.stderr}',
+                severity=Severity.ERROR).publish()
+        scrub_finish_lines = list(finish_scrub_follower)
+        if not scrub_finish_lines:
+            ValidatorEvent(
+                message=f'No scrubbing validation message found in db logs on node: {node.name}', severity=Severity.ERROR).publish()
+            return
+        invalid_sstables = [line for line in scrub_finish_lines if 'sstable is invalid' in line]
```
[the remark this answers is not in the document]
soyacz: it's not a fancy regexp that needs explanation - just verification if line contains `sstable is invalid` which later is not even parsed.

sdcm/tester.py
```diff
@@ -2788,6 +2789,10 @@ def tearDown(self):
         self.teardown_started = True
         with silence(parent=self, name='Sending test end event'):
             InfoEvent(message="TEST_END").publish()
+        self.log.info("Post test validators are starting...")
+        for validator_class in teardown_validators_list:
+            if validator_class.is_enabled(self.params):
+                validator_class(self.db_cluster).validate()
```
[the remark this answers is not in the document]
soyacz: fixed. Iterating over list of clusters

soyacz:
I adapted code according to comments. Configuration now looks like this:
```
# teardown validators
teardown_validators:
  scrub:
    enabled: false
    timeout: 1200
    keyspace: ''
    table: ''
```

unit_tests/test_tester.py
```diff
@@ -34,7 +34,8 @@ def _load_environment_variables(self):
         return {
             'config_files': ['test-cases/PR-provision-test-docker.yaml'],
             'cluster_backend': 'docker',
-            'run_commit_log_check_thread': False
+            'run_commit_log_check_thread': False,
+            'teardown_validate_sstables': False
```
vponomaryov, line 38: It becomes left-over

sdcm/teardown_validators/sstables.py
```diff
@@ -0,0 +1,68 @@
...
+            s3_link = self._upload_corrupted_files(node, list(quarantine_lines))
+            ScrubValidationErrorEvent(node.name, s3_link).publish()
+
+    def validate(self):
+        keyspace = self.configuration.get("keyspace")
+        table = self.configuration.get("table")
+        timeout = self.configuration.get("timeout", 1200)
+        run_scrub = partial(self._run_nodetool_scrub, keyspace=keyspace, table=table, timeout=timeout)
+        run_scrub.__name__ = run_scrub.func.__name__
+        print(run_scrub.func.__name__)
```
vponomaryov, line 59: Debug line that should be removed?

sdcm/teardown_validators/sstables.py
```diff
@@ -0,0 +1,68 @@
...
+
+
+class SstablesValidator(TeardownValidator):  # pylint: disable=too-few-public-methods
+    validator_name = 'scrub'
+
+    def _upload_corrupted_files(self, node: BaseNode, quarantine_log_lines):
+        # get quarantine dir from lines:
+        # INFO  2024-04-02 12:40:24,787 [shard 0:stre] sstable - Moving sstable /var/lib/scylla/data/system_schema/columns-24101c25a2ae3af787c1b40ee1aca33f/me-3gey_0z3c_2h5vl2ogebmg26ku9t-big-Data.db to "/var/lib/scylla/data/system_schema/columns-24101c25a2ae3af787c1b40ee1aca33f/quarantine"  # pylint: disable=line-too-long
+        # print(log_lines)
+        quarantine_dirs = {line.split(
```
vponomaryov, line 21: Can you please split that long line to the following more readable smaller lines?
```
quarantine_dirs = {
    line.split(' to "')[1][:-2]
    for line in quarantine_log_lines
    if 'sstable - Moving sstable' in line and 'quarantine' in line
}
```

sdcm/teardown_validators/sstables.py
soyacz: fixed

unit_tests/test_tester.py
soyacz: fixed

sdcm/teardown_validators/sstables.py
soyacz: fixed

defaults/test_default.yaml
fruch: Yes it can be a followup, it was just used as example of possible configuration we might want to add

defaults/test_default.yaml
```diff
@@ -242,3 +242,11 @@ validate_large_collections: false
 scylla_d_overrides_files: []
 
 run_commit_log_check_thread: true
+
+# teardown validators
+teardown_validators:
+  scrub:
```
fruch, line 248: I do agree documenting for those should be in it's own markdown, explaining the why and giving examples of usages.

fruch (approved):
LGTM

soyacz:
one more thing, this validation was enabled for "individual nemesis" only - is that enough?

fruch:
> one more thing, this validation was enabled for "individual nemesis" only - is that enough?

I think we should add a commit to enable it for one of the tier1 tests

soyacz:
> > one more thing, this validation was enabled for "individual nemesis" only - is that enough?
> 
> I think we should add a commit to enable it for one of the tier1 tests

Ok, added commit enabling it for parallel nemesis with topology changes tier1 test.

defaults/test_default.yaml
```diff
@@ -242,3 +242,11 @@ validate_large_collections: false
...
+    enabled: false
```
roydahan, line 249: Maybe we should set it to "true" for a while to see what it catches and how stable the feature is.

roydahan (approved):
LGTM, just one comment about the feature default.
I think we should start with enabling by default for all tests and see how it goes...

fruch:
> LGTM, just one comment about the feature default.
> I think we should start with enabling by default for all tests and see how it goes...

fruch (approved):
LGTM

fruch:
After a week seeing it on the teir1 case, we can consider adding it to more tests, or all

PR #7322, opened 2024-04-05
improvement(docker-backend): run Nemesis on docker backend in Jenkins

dimakr:
The change adds a capability to execute Nemesis cases on Docker backend in Jenkins.

Typically Docker backend is used for executing tests on an engineer's local environment. However, a standard test setup for a Scylla cluster - including the Scylla cluster itself, test runner, loader, monitoring instances - requires significant HW resources. This can sometimes prevent an engineer from being able to normally execute other tasks on the local machine, especially considering that a regular longevity+Nemesis test configuration runs for approximately 90 minutes.
To mitigate this problem, we can use the existing Jenkins pipelines to run a test runner instance for Docker backend in AWS cloud provider.

The following changes are introduced:
- longevity pipeline is updated to create only test runner instance when Docker is selected as a backend. This way the test runner instance is also serving the role of Docker backend node, the same way how tests are executed locally.
- configuration of storage disk is performed on creating an AWS test runner instance, when Docker is selected as a backend.
- sdcm.utils.nemesis.NemesisJobGenerator class is updated to be able to generate Jenkins job files for individual Nemesis cases;
- Individual Nemesis jenkins job files are generated for Docker backend.

The related task: https://github.com/scylladb/qa-tasks/issues/1636

### Testing
Test run in Jenkins on Docker backend: [CorruptThenScrubMonkey nemesis run on docker backend in Jenkins](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-docker-test/5/)
Test run in Jenkins on AWS (regression check): [CorruptThenScrubMonkey nemesis run on AWS in Jenkins](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-docker-test/4/)

### Reminders

### DIFF
--- a/sdcm/sct_runner.py
+++ b/sdcm/sct_runner.py
@@ -19,6 +19,7 @@ import string
 import random
 import tempfile
 import datetime
+import glob
 from contextlib import suppress
 from enum import Enum
 from functools import cached_property
@@ -505,7 +506,7 @@ class AwsSctRunner(SctRunner):
 
         return aws_region.resource.Image(existing_amis[0]["ImageId"])
 
-    # pylint: disable=too-many-arguments,too-many-locals
+    # pylint: disable=too-many-arguments,too-many-locals,too-many-statements
     def _create_instance(self,
                          instance_type: InstanceTypeType,
                          base_image: Any,
@@ -600,7 +601,66 @@ class AwsSctRunner(SctRunner):
         LOGGER.info("Got public IP: %s", instance.public_ip_address)
         self.instance = instance
 
+        # If the SCT runner instance is used as a Docker backend for tests
+        if self.params.get("cluster_backend") == "docker":
+            LOGGER.info("Configure unused disks to use as a persistent storage for Docker.")
+            LOGGER.debug("Connecting to instance...")
+            remoter = self.get_remoter(host=instance.public_ip_address, connect_timeout=120)
+
+            LOGGER.debug("Getting unused disks...")
+            disks = self._get_unused_disks(remoter)
+            LOGGER.debug("The following unused disks are detected: %s", disks)
+            if disks:
+                mount_point = "/mnt/docker-persistent-storage"
+                self._setup_disks(remoter, disks, mount_point)
+                LOGGER.debug("Disks are configured.")
+                self._set_docker_data_root(remoter, mount_point)
+                LOGGER.debug("Persistent storage for Docker is configured.")
+            else:
+                LOGGER.debug("No unused disks found.")
+
+            remoter.stop()
+
         return instance
+
+    @staticmethod
+    def _get_unused_disks(remoter: Any) -> list[str]:
+        result = remoter.run("lsblk -dnpo NAME,TYPE | grep disk | cut -d' ' -f1")
+        devices = result.stdout.strip().splitlines()
+
+        unused = []
+        for device in devices:
+            dev = device.replace("/dev/", '')
+            if len(glob.glob("/sys/class/block/{dev}/{dev}*".format(dev=dev))) == 0:
+                unused.append(device)
+        return unused
+
+    @staticmethod
+    def _setup_disks(remoter: Any, disks: list, mount_point: str) -> None:
+        raid_device = "/dev/md0"
+        raid_level = 0
+        # Create RAID if more than one unused disk is available
+        if len(disks) != 1:
+            remoter.sudo(f"mdadm --create --verbose {raid_device} --level={raid_level} "
+                         f"--raid-devices={len(disks)} " + " ".join(disks))
+            remoter.sudo(f"mkfs.xfs {raid_device}")
+        else:
+            remoter.sudo(f"mkfs.xfs {disks[0]}")
+
+        mount_dev = disks[0] if len(disks) == 1 else raid_device
+        remoter.sudo(f"mkdir -p {mount_point}")
+        remoter.sudo(f"mount {mount_dev} {mount_point}")
+
+        # Ensure that the configured device mounts on boot
+        remoter.sudo(f"sh -c \"echo '{mount_dev} {mount_point} xfs defaults 0 0' >> /etc/fstab\"")
+
+    @staticmethod
+    def _set_docker_data_root(remoter: Any, docker_data_dir: str) -> None:
+        remoter.sudo("systemctl stop docker")
+        remoter.sudo(f"rsync -a /var/lib/docker/ {docker_data_dir}")
+        remoter.sudo("rm -rf /var/lib/docker")
+        remoter.sudo(f"ln -s {docker_data_dir} /var/lib/docker")
+        remoter.sudo("systemctl start docker")
 
     def _stop_image_builder_instance(self, instance: Any) -> None:
         instance.stop()
--- a/sdcm/utils/nemesis.py
+++ b/sdcm/utils/nemesis.py
@@ -21,6 +21,7 @@ class NemesisJobGenerator:
         "aws": "eu-west-1",
         "gce": "us-east1",
         "azure": "eastus",
+        "docker": "eu-west-1"
     }
 
     def __init__(self, base_job: str = None, base_dir: str | Path = Path("."), backend: str = None) -> 'NemesisJobGenerator':
@@ -109,12 +110,14 @@ class NemesisJobGenerator:
 
     def create_job_files_from_template(self) -> list[Path]:
         pipeline_files = []
+        backend_config = config.NEMESIS_REQUIRED_ADDITIONAL_CONFIGS.get(self.backend, [])
         for cls in self.nemesis_class_list:
             additional_configs = config.NEMESIS_REQUIRED_ADDITIONAL_CONFIGS.get(cls, [])
             additional_params = config.NEMESIS_ADDITIONAL_PIPELINE_PARAMS.get(cls, {})
             config_name = [
                 str(self.nemesis_test_config_dir / f"{self.base_job}-nemesis.yaml"),
                 str(self.nemesis_test_config_dir / f"{cls}.yaml"),
+                *backend_config,
                 *additional_configs,
             ]
             job_file_name = f"{self.base_job}-{cls}-{self.backend}.jenkinsfile"
--- a/configurations/nemesis/RejectInterNodeNetworkMonkey.yaml
+++ b/configurations/nemesis/RejectInterNodeNetworkMonkey.yaml
@@ -1,0 +1,2 @@
+nemesis_class_name: 'RejectInterNodeNetworkMonkey'
+user_prefix: 'longevity-5gb-1h-RejectInterNodeNetworkMonkey'
--- a/configurations/nemesis/RestartThenRepairNodeMonkey.yaml
+++ b/configurations/nemesis/RestartThenRepairNodeMonkey.yaml
@@ -1,0 +1,2 @@
+nemesis_class_name: 'RestartThenRepairNodeMonkey'
+user_prefix: 'longevity-5gb-1h-RestartThenRepairNodeMonkey'
--- a/configurations/nemesis/ValidateHintedHandoffShortDowntime.yaml
+++ b/configurations/nemesis/ValidateHintedHandoffShortDowntime.yaml
@@ -1,0 +1,2 @@
+nemesis_class_name: 'ValidateHintedHandoffShortDowntime'
+user_prefix: 'longevity-5gb-1h-ValidateHintedHandoffShortDowntime'
--- a/configurations/nemesis/additional_configs/docker_backend_local.yaml
+++ b/configurations/nemesis/additional_configs/docker_backend_local.yaml
@@ -20,6 +20,9 @@ nemesis_interval: 3
 nemesis_filter_seeds: false
 nemesis_during_prepare: true
 
+# Instance type for the case when tests are executed on Docker backend in AWS
+instance_type_runner: 'i4i.xlarge'
+
 # NOTE: the parameters reduce footprint of scylla cluster for docker backend, comparing to default values
 append_scylla_args: '--smp 1 --memory 2G'
 use_mgmt: false
--- a/jenkins-pipelines/oss/nemesis/longevity-5gb-1h-AbortRepairMonkey-docker.jenkinsfile
+++ b/jenkins-pipelines/oss/nemesis/longevity-5gb-1h-AbortRepairMonkey-docker.jenkinsfile
@@ -1,0 +1,12 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    backend: 'docker',
+    region: 'eu-west-1',
+    test_name: 'longevity_test.LongevityTest.test_custom_time',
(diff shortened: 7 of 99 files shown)


vars/createArgusTestRun.groovy
```diff
@@ -2,14 +2,15 @@
...
+    def cloud_provider = getCloudProviderFromBackend(params.backend)
     sh """#!/bin/bash
         set -xe
 
         echo "Creating Argus test run ..."
         if [[ -n "${params.requested_by_user}" ]] ; then
             export BUILD_USER_REQUESTED_BY=${params.requested_by_user}
         fi
-        export SCT_CLUSTER_BACKEND="${params.backend}"
+        export SCT_CLUSTER_BACKEND="${cloud_provider}"
```
fruch, line 13: Why are we changing this ? It doesn't seem to be related to this change set

sdcm/utils/nemesis.py
```diff
@@ -113,8 +114,8 @@ def create_job_files_from_template(self) -> list[Path]:
             additional_configs = config.NEMESIS_REQUIRED_ADDITIONAL_CONFIGS.get(cls, [])
             additional_params = config.NEMESIS_ADDITIONAL_PIPELINE_PARAMS.get(cls, {})
             config_name = [
-                str(self.nemesis_test_config_dir / f"{self.base_job}-nemesis.yaml"),
```
fruch, line 116: Why are you switching the order ?

sdcm/sct_runner.py
```diff
@@ -600,8 +602,56 @@ def _create_instance(self,
...
+        if os.getenv("SCT_CLUSTER_BACKEND") == "docker":
+            LOGGER.info("Configure unused disks to use as a Docker container volume for Scylla data.")
+            LOGGER.info("Connecting to instance...")
+            remoter = self.get_remoter(host=instance.public_ip_address, connect_timeout=120)
+
+            LOGGER.info("Getting unused disks...")
+            disks = self._get_unused_disks(remoter)
+            LOGGER.info("The following unused disks are detected: %s", disks)
+            if disks:
+                self._setup_disks(remoter, disks)
```
fruch, line 615: Missing the part we pointing docker to those mount points

configurations/nemesis/longevity-1gb-1h-nemesis.yaml
```diff
@@ -0,0 +1,40 @@
+test_duration: 90
```
fruch, line 1: This file seems like a duplication,

And now we have 3 identical files with different names, we should align it and leave with only one.

fruch (commented):
Looks in a good direction, few bits we need to address

vars/createArgusTestRun.groovy
dimakr: This is my mistake.
Initial implementation for this PR used another word in Jenkins parameters, to indicate docker backend. And that word was not known in sdcm.SCTConfiguration.defaults_config_files map, so some logic was added to work around that.
But at the end simple 'docker' wording for indicating backend in Jenkins is working well for existing sct_config logic.
So this change is not needed, reverting it back.

configurations/nemesis/longevity-1gb-1h-nemesis.yaml
dimakr: Adding this file was done to reduce as much as possible the changes in NemesisJobGenerator logic.
The Job generator takes 'base-job' parameter and uses its name (e.g. 'longevity-1gb-1h') to find the base job, the base test config, etc. So we just add a new config and in the generator code we change nothing (or little), and then we are able to generate dedicated nemeses for a new backend.

Another option that was considered of how to adjust NemesisJobGenerator is to re-use 'configurations/nemesis/additional_configs/docker_backend_local.yaml' config. But it would require some more changes in JobGenerator.
Also it is less flexible in terms of parameters  - for instance at some point we can decide to use not i4i.xlarge instance type for SCTRunner on AWS, but 2xlarge, or 4xlarge, etc. This means we can also increase memory limits in 'append_scylla_args parameter', maybe increase test data volume from 1 GB to 5 GB, etc. Then it will not affect local docker backend runs experience if a separate config exists for Jenkins docker backend runs.

What do you think?

fruch: I don't think we're gonna support multiple bases for the generator, if we'll need a certain setup we could build it, not to generate.


Regardless of the name wet should pick, we should have one copy of it

sdcm/utils/nemesis.py
dimakr: This is just to keep consistency in test resources naming.

Now all dedicated nemesis configs (the configurations/nemesis/*Monkey.yaml ones) have user prefix defined as 
e.g. 'user_prefix: 'longevity-5gb-1h-CorruptThenRepairMonkey''.
And if a user runs a docker backend on Jenkins, and sees that job uses configurations/nemesis/longevity-1gb-1h-nemesis.yaml base config, and in log sees that prepare stress command writes only 1G of data, but then sees that test resources are named as 'longevity-5gb-*-db-node-*' - it can cause confusion.

*Monkey.yaml configs are changed rarely (if ever). So when the 'longevity-1gb-1h-' (or similar) config is placed last in the list of configs for Job, we overwrite the user_prefix parameter value to exactly what the test does.

dimakr: Reverting this as well. The custom user_prefix will come from docker_backend_local config, as we decided not to multiply base configs for Job generator.

sdcm/sct_runner.py
dimakr: Fixed this.
The docker now is on this additional disk:
```
ubuntu@ip-10-4-1-70:~/sct-results$ lsblk | grep mnt
nvme1n1      259:1    0 872.6G  0 disk /mnt/docker-persistent-storage

ubuntu@ip-10-4-1-70:~/sct-results$ docker inspect -f '{{ .Mounts }}' 15d4d598f324
[{bind  /var/run/docker.sock /var/run/docker.sock  rw true rprivate} {volume 93dd165165820f86c4ed24990106de312be3249acdaaaf27844ab17198c0db44 /mnt/docker-persistent-storage/volumes/93dd165165820f86c4ed24990106de312be3249acdaaaf27844ab17198c0db44/_data /var/lib/scylla local  true }]
```

Re-checked jobs:
Docker backend: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-docker-test/34/
AWS backend: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-docker-test/35/

vars/createArgusTestRun.groovy
fruch: so please revert it

fruch (changes requested):
LGTM, but need to revert this change https://github.com/scylladb/scylla-cluster-tests/pull/7322/commits/a508cdeab92450eed7f19ddeade7d79b8355481e#r1564750379

soyacz:
> LGTM, but need to revert this change [a508cde#r1564750379](https://github.com/scylladb/scylla-cluster-tests/commit/a508cdeab92450eed7f19ddeade7d79b8355481e#r1564750379)

I think the `vars/createArgusTestRun.groovy` change was reverted, all the rest should be kept - unless you have something else in mind.

soyacz (dismissed):
LGTM

sdcm/sct_runner.py
```diff
@@ -600,8 +602,67 @@ def _create_instance(self,
         LOGGER.info("Got public IP: %s", instance.public_ip_address)
         self.instance = instance
 
+        # If the SCT runner instance is used as a Docker backend for tests
+        if os.getenv("SCT_CLUSTER_BACKEND") == "docker":
```
vponomaryov, line 606: `self.params.get("cluster_backend")` must be used for it.
There is big enough probability to not have exactly env var be set, but yaml config option.

sdcm/sct_runner.py
```diff
@@ -600,8 +602,67 @@ def _create_instance(self,
...
+            LOGGER.info("Configure unused disks to use as a persistent storage for Docker.")
+            LOGGER.info("Connecting to instance...")
+            remoter = self.get_remoter(host=instance.public_ip_address, connect_timeout=120)
+
+            LOGGER.info("Getting unused disks...")
```
vponomaryov, line 611: Such kind of a message looks like a `LOGGER.debug(...)`, but not `info`.

sdcm/sct_runner.py
```diff
@@ -600,8 +602,67 @@ def _create_instance(self,
         LOGGER.info("Got public IP: %s", instance.public_ip_address)
         self.instance = instance
 
+        # If the SCT runner instance is used as a Docker backend for tests
+        if os.getenv("SCT_CLUSTER_BACKEND") == "docker":
+            LOGGER.info("Configure unused disks to use as a persistent storage for Docker.")
+            LOGGER.info("Connecting to instance...")
```
vponomaryov, line 608: Such kind of a message looks like a `LOGGER.debug(...)`, but not `info`.

sdcm/sct_runner.py
```diff
@@ -600,8 +602,67 @@ def _create_instance(self,
...
+            remoter = self.get_remoter(host=instance.public_ip_address, connect_timeout=120)
+
+            LOGGER.info("Getting unused disks...")
+            disks = self._get_unused_disks(remoter)
+            LOGGER.info("The following unused disks are detected: %s", disks)
```
vponomaryov, line 613: Same here

sdcm/sct_runner.py
```diff
@@ -600,8 +602,67 @@ def _create_instance(self,
...
+
+        unused = []
+        for device in devices:
+            dev = device.replace("/dev/", '')
+            if len(glob.glob("/sys/class/block/{dev}/{dev}*".format(dev=dev))) == 0:
+                unused.append(device)
+        return unused
+
+    @staticmethod
+    def _setup_disks(remoter: Any, disks: list, mount_point: str) -> None:
```
vponomaryov, line 640: Disk setup doesn't look as a requirement for being able to run a test on docker backend.
We do not expect perf to be great in this case.

Moreover, commit topic and description doesn't tell anything about the disk setup.
About such a significant change as changing disks formatting...

vars/createArgusTestRun.groovy
```diff
@@ -2,15 +2,14 @@
...
-    def cloud_provider = getCloudProviderFromBackend(params.backend)
     sh """#!/bin/bash
         set -xe
 
         echo "Creating Argus test run ..."
         if [[ -n "${params.requested_by_user}" ]] ; then
             export BUILD_USER_REQUESTED_BY=${params.requested_by_user}
         fi
-        export SCT_CLUSTER_BACKEND="${cloud_provider}"
+        export SCT_CLUSTER_BACKEND="${params.backend}"
```
vponomaryov, line 12: This file change reverting must have been done in the proper commit.

sdcm/sct_runner.py
dimakr: The implemented logic is intended to be used only when Docker backend test run is started in Jenkins. In this case the env parameter is always set here https://github.com/scylladb/scylla-cluster-tests/blob/1f38a636ca1695c7bb0287dc9fd176dec7e192bf/vars/runSctTest.groovy#L28

And `self.params.get("cluster_backend")` won't help here. When running tests from Jenkins there is this differentiation between 'cluster backend for tests' and cloud provider for the 'cluster', like it is defined in https://github.com/scylladb/scylla-cluster-tests/blob/1f38a636ca1695c7bb0287dc9fd176dec7e192bf/vars/getCloudProviderFromBackend.groovy#L3

So, when the test is started from Jenkins, in the condition above we need to check what the selected backend for tests is (SCT_CLUSTER_BACKEND has this value), not the cloud provider (the value returned by `self.params.get("cluster_backend")`)

sdcm/sct_runner.py
dimakr: Without a dedicated physical disk Nemesis cases are always failing, when executed on Docker backend in Jenkins.
Only after configuring a free disk (which comes with e.g. an i4i instance type) to be used as a permanent storage for Docker - the Nemesis cases started to pass.

dimakr:
> LGTM, but need to revert this change [a508cde#r1564750379](https://github.com/scylladb/scylla-cluster-tests/commit/a508cdeab92450eed7f19ddeade7d79b8355481e#r1564750379)

The change is reverted in the dependent commit.
Is it OK to have it this way, and we will have all the changes combined in the final squash?

sdcm/sct_runner.py
vponomaryov: @dimakr 
The `self.params.get("cluster_backend")` will always be equal to the `os.getenv("SCT_CLUSTER_BACKEND")` if the latter is set. And I proposed you to use the former approach with `self.params` to cover cases when env var is not set.
The `vars/getCloudProviderFromBackend.groovy` file is not related to the raised concern at all.

sdcm/sct_runner.py
vponomaryov: If nemesis that depends on the disk perf fails all the time without local NVMes then there should be detection of it and proper skipping.

vponomaryov:
> > LGTM, but need to revert this change [a508cde#r1564750379](https://github.com/scylladb/scylla-cluster-tests/commit/a508cdeab92450eed7f19ddeade7d79b8355481e#r1564750379)
> 
> @fruch The change is reverted in the dependent commit. Is it OK to have it this way, and we will have all the changes combined in the final squash?

We use `rebase and merge` and not `squash and merge` actions.
We don't want to lose git commit history details.

Then, in your case it is pretty simple action - need to revert whole file changes.

fruch:
> > > LGTM, but need to revert this change [a508cde#r1564750379](https://github.com/scylladb/scylla-cluster-tests/commit/a508cdeab92450eed7f19ddeade7d79b8355481e#r1564750379)
> > 
> > 
> > @fruch The change is reverted in the dependent commit. Is it OK to have it this way, and we will have all the changes combined in the final squash?
> 
> We use `rebase and merge` and not `squash and merge` actions. We don't want to lose git commit history details.
> 
> Then, in your case it is pretty simple action - need to revert whole file changes.

@dimakr 

It might sound a bit pedantic, but yes we are striving to have a clear commits flow, especially since we might need to backport fixes/features into release branches 

Also there are some more comments to address about a usage of environmental variable, and some logs level

In less you have a very strong argument, complying to what maintener asks is usually the quickest way (they even guard me, from cutting some corners)

sdcm/sct_runner.py
dimakr: Changes to use params for getting cluster_backend value.

sdcm/sct_runner.py
dimakr: Changed to debug instead of the very first message where we notify about 'configuring'.
It helps to see immediately in that this step was executed, e.g. in Jenkins, without the need to download and look through test logs

sdcm/sct_runner.py
dimakr: Updated the commit message.
With regard to the configuring the dedicated disk - no Nemesis case is able to run on EBS, So the dedicated disk is necessary to be able to run docker backend on AWS.

vars/createArgusTestRun.groovy
dimakr: Reverted in the proper commit.

dimakr:
The comments are addressed.
This is the sanity checks that the latest changes do not brake the functionality: [longevity-5gb-1h-docker-test_build_48](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-docker-test/48/)

soyacz (approved):
LGTM

fruch (approved):
LGTM

Issue #7277, opened 2024-03-18
All loaders have the same index `0` when `round_robin` SCT config option is `true`


vponomaryov:
## Packages

Scylla version: `2024.1.2-20240228.2c85a811d0be` with build-id `147ca5a1ee4dc1a8c97215a2a5ad7b015940a42f`

Operator Image: `scylladb/scylla-operator:latest`
Operator Helm Version: `v1.13.0-alpha.0-49-gf356138-latest`
Operator Helm Repository: `https://storage.googleapis.com/scylla-operator-charts/latest`
Kernel Version: `5.10.210-201.852.amzn2.x86_64`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

If we configure more than 1 loader and set the `round_robin` SCT config option to be `true` then
`loader_idx` is always `0`.
It affects following stuff:
- Log file names (low priority problem)
- Stats update in ES (high priority problem)

In the test run used for reporting this issue we have following log results for loaders:
![Screenshot from 2024-03-18 14-41-27](https://github.com/scylladb/scylla-cluster-tests/assets/3796058/91324da6-86c9-4e65-979b-c9cd095dd177)

But we have 2 loaders here.

## Impact

Wrong information that introduces confusion for nothing.

## How frequently does it reproduce?

100%

## Installation details

Cluster size: 4 nodes (i4i.4xlarge)

Scylla Nodes used in this run:
**No resources left at the end of the run**

OS / Image: `` (k8s-eks: `eu-north-1`)

Test: `vp-longevity-scylla-operator-alternator-https-3h-eks`
Test id: `c52bd59b-69ff-4d82-9096-a23a66b08d41`
Test name: `scylla-staging/valerii/vp-longevity-scylla-operator-alternator-https-3h-eks`
Test config file(s):

- [longevity-scylla-operator-alternator-https-3h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/8a849fceb7a4aeb562799eaf1d80a15f1fbd1493/test-cases/scylla-operator/longevity-scylla-operator-alternator-https-3h.yaml)

<details>
<summary>
Logs and commands
</summary>

- Restore Monitor Stack command: `$ hydra investigate show-monitor c52bd59b-69ff-4d82-9096-a23a66b08d41`
[...]

PR #7326, opened 2024-04-05
fix(loaders): use proper loader indexes when `round_robin` is enabled

vponomaryov:
When we use `round_robin=true` we provide list of one loader node to the stress commands handling code then use `enumerate` func incorrectly providing the index as `0`.

Fix it by switching usage of the `enumerate` func to the `node_index` attribute of loader node objects.

Side effect is that in all `round_robin=false` cases the first index will be `1`, not `0` as has always been.

Fixes: #7277

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

PR #7350, opened 2024-04-17
fix(perf): prolong wait for no compactions

soyacz:
When using tablets, splitting them is treated as compaction and may take very long. This causes latency during grow-shrink perf test with tablets to fail.

Fix is by prolonging this timeout so there's enough room for that.

### Testing

- [ ]

### Reminders


vponomaryov (approved):
LGTM

PR #7354, opened 2024-04-19
test(perf): set initial tablets to 32 in perf tests

soyacz:
Grow-shrink cluster scenario perf test with tablets inserts 650GB of data. This triggers split tablets in steps from 8->16->32->64 or even to
128. Splitting takes time and prolongs the test significantly.

Commit is about increasing initial tablets for ks created by c-s to 32 (8x4) so test is faster and we still can observe splitting to 64 and see effects of this dynamic split during the test.

### Testing

- [ ]

### Reminders

### DIFF
--- a/configurations/tablets-initial-32.yaml
+++ b/configurations/tablets-initial-32.yaml
@@ -1,0 +1,5 @@
+append_scylla_yaml: |
+  experimental_features:
+    - tablets
+    - consistent-topology-changes
+  tablets_initial_scale_factor: 4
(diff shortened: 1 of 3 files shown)


fruch (approved):
LGTM

fruch:
@soyacz we have a run passing with this ?

roydahan:
So IIUC, now we won't test the default of tablets for performance?

fruch:
> So IIUC, now we won't test the default of tablets for performance?

For a specific test

fruch:
@xemul can you verity this is the best configuration to set for this case ?

xemul:
Looks sane to me

fruch:
> Looks sane to me

@soyacz let's make sure you backported it it..

PR #7363, opened 2024-04-24
improvement(logging): add hostname info to RemoteCmdRunner

soyacz:
When investigating SCT run, logs often I miss information about node command is executed on.

Adding `<self.hostname>: ` prefix to most log usages in `RemoteCmdRunner` class so it's easier to debug runs.

Example result line:
`<172.17.0.2>: Running command "dpkg -l |grep scylla|awk '{print $2 "-" $3}'"...`

### Testing

- [X] - tested locally with docker
- [x] - added provision test

### Reminders

### DIFF
--- a/sdcm/remote/base.py
+++ b/sdcm/remote/base.py
@@ -135,19 +135,21 @@ class CommandRunner(metaclass=ABCMeta):
 
     def _print_command_results(self, result: Result, verbose: bool, ignore_status: bool):
         """When verbose=True and ignore_status=True that means nothing will be printed in any case"""
+        hostname = self.hostname
         if verbose and not result.failed:
             if result.stderr:
-                self.log.debug('STDERR: %s', result.stderr)
+                self.log.debug('<%s>: STDERR: %s', hostname, result.stderr)
 
-            self.log.debug('Command "%s" finished with status %s', result.command, result.exited)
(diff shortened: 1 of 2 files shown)


sdcm/remote/base.py
```diff
@@ -137,17 +137,18 @@ def _print_command_results(self, result: Result, verbose: bool, ignore_status: b
         """When verbose=True and ignore_status=True that means nothing will be printed in any case"""
         if verbose and not result.failed:
             if result.stderr:
```
vponomaryov, line 139: Please, define some local var for the following reasons:
- Better readability
- Reduce CPU actions by getting attr again and again (not critical, but why not?)

Example:
```
hostname = self.hostname
...
self.log.debug('<%s>: STDERR: %s', hostname, result.stderr)
...
self.log.debug('<%s>: Command "%s" finished with status %s', hostname, result.command, result.exited)
```

dimakr (approved):
LGTM

sdcm/remote/base.py
soyacz: ok, done

dimakr (approved):

fruch (approved):
LGTM

Even it would be nice if we can also add it to every single output string, since there are mixed outputs at the same time, which are hard to understand

Issue #7357, opened 2024-04-22
Don't put the output of `coredumpctl info` in SCT-based issue reports.


michoecho:
As discussed in https://github.com/scylladb/scylladb/issues/18312#issuecomment-2068806266, the output of `coredumpctl info` is almost always useless for people working with the issue — it only adds noise, a lot of it.

Here's another complaint about the same thing from @kbr-scylla: https://github.com/scylladb/scylladb/issues/17786#issuecomment-2059213151.

If the output of `coredumpctl info` is a part of some SCT issue template, please remove it from the template.

---

While we are at it, please change this thing
```
download_instructions=gsutil cp gs://upload.scylladb.com/core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000/core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000.gz .
gunzip /var/lib/systemd/coredump/core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000.gz
```
to
```
gsutil cp gs://upload.scylladb.com/core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000/core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000.gz .
gunzip core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000.gz
```
The `download_instructions=` only breaks easy copy-paste, while the `/var/` path doesn't make any sense.

---

Edit: in case it isn't obvious, the output of `coredumpctl info` I mean is this thing:
```
2024-04-13 23:51:23.387 <2024-04-13 23:29:14.000>: (CoreDumpEvent Severity.ERROR) period_type=one-time event_id=36a35814-8e4f-4547-bc2d-39060113f17e node=Node longevity-tls-50gb-3d-master-db-node-1e2fb19f-4 [54.74.6.242 | 10.4.9.129] (seed: True)
corefile_url=https://storage.cloud.google.com/upload.scylladb.com/core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000/core.scylla.112.ceb3df3d7fe14734b36c216669f9067f.4539.1713050954000000.gz
backtrace=           PID: 4539 (scylla)
           UID: 112 (scylla)
           GID: 118 (scylla)
        Signal: 11 (SEGV)
[...]

PR #7368, opened 2024-04-25
fix(coredump): remove unnecessary info

soyacz:
We don't need details about system modules in coredump details we post in issues. Removing it from the coredump event so it's never reposted there. Fixes also download instructions.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/7357

### Testing

- [ ]

### Reminders


sdcm/coredump.py
```diff
@@ -211,7 +211,7 @@ def _upload_coredump(self, core_info: CoreDumpInfo):
                               "'%s' 'https://%s'" % (coredump, upload_url))
         download_url = 'https://storage.cloud.google.com/%s' % upload_url
         self.log.info("You can download it by %s (available for ScyllaDB employee)", download_url)
-        download_instructions = 'gsutil cp gs://%s .\ngunzip %s' % (upload_url, coredump)
+        download_instructions = 'gsutil cp gs://%s .\ngunzip %s' % (upload_url, coredump.rsplit('/', 1)[-1])
```
fruch, line 214: Why not use pathlib ? i.e. it's a path we are handling here..

fruch (approved):
LGTM

michoecho (approved):
LGTM.

(Though I think that none of `coredumpctl info` is useful, neither the parts you filtered out nor the ones you left. I mean, it's not a bad idea to have it in the logs, but I wouldn't paste them into github issues. But that's just my opinion.)

vponomaryov (changes requested):
Need to fix the unit tests

fruch:
> LGTM.
> 
> (Though I think that none of `coredumpctl info` is useful, neither the parts you filtered out nor the ones you left. I mean, it's not a bad idea to have it in the logs, but I wouldn't paste them into github issues. But that's just my opinion.)

When you open a bug you can pick which information you think is important, I think the backtrace can be helpful, for example if you don't have the decoded backtraces from scylla.

Anyhow there's a link to Argus, and there you can find all of the data you need, and eventually people would stop cheery-picking the wrong data

soyacz:
> Need to fix the unit tests

unit tests fixed

fruch (approved):
LGTM

Issue #7376, opened 2024-04-26
[Azure] `InvalidRequestContent` error in the collect logs stage when SCT tries to update SCT runner node tags


vponomaryov:
## Packages

Scylla version: `2024.1.3-20240401.64115ae91a55` with build-id `73323c880c7f35e108792dd047cb313d926d3ea9`
Kernel Version: `5.15.0-1059-azure`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

In test runs SCT tries to update SCT runner node tags after successful logs gathering:

```
15:03:54  +---------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
15:03:54  |                                                      Collected logs by test-id: 9e6c9e42-f001-4907-b568-1061188093f3                                                      |
15:03:54  +---------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
15:03:54  | Cluster set               | Link                                                                                                                                          |
15:03:54  +---------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
15:03:54  | db-cluster                | https://cloudius-jenkins-test.s3.amazonaws.com/9e6c9e42-f001-4907-b568-1061188093f3/20240426_115509/db-cluster-9e6c9e42.tar.gz                |
15:03:54  | sct-runner-events         | https://cloudius-jenkins-test.s3.amazonaws.com/9e6c9e42-f001-4907-b568-1061188093f3/20240426_115509/sct-runner-events-9e6c9e42.tar.gz         |
15:03:54  | sct-runner-python-log     | https://cloudius-jenkins-test.s3.amazonaws.com/9e6c9e42-f001-4907-b568-1061188093f3/20240426_115509/sct-9e6c9e42.log.tar.gz                   |
15:03:54  | loader-set                | https://cloudius-jenkins-test.s3.amazonaws.com/9e6c9e42-f001-4907-b568-1061188093f3/20240426_115509/loader-set-9e6c9e42.tar.gz                |
15:03:54  | monitor-set               | https://cloudius-jenkins-test.s3.amazonaws.com/9e6c9e42-f001-4907-b568-1061188093f3/20240426_115509/monitor-set-9e6c9e42.tar.gz               |
[...]

Issue #7375, opened 2024-04-26
[GCE] `bad argument type for built-in operation` error in the `collect logs` stage when SCT tries to update SCT runner node tags


vponomaryov:
## Packages

Scylla version: `2024.1.3-20240401.64115ae91a55` with build-id `73323c880c7f35e108792dd047cb313d926d3ea9`
Kernel Version: `5.15.0-1054-gcp`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

In test runs SCT tries to update SCT runner node tags after successful logs gathering:

```
15:28:31  +---------------------------------------------------------------------------------------------------------------------------------------------------------------------------+
15:28:31  |                                                      Collected logs by test-id: 756aa4c9-80f8-4a77-acd2-c521dc189bdb                                                      |
15:28:31  +---------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
15:28:31  | Cluster set               | Link                                                                                                                                          |
15:28:31  +---------------------------+-----------------------------------------------------------------------------------------------------------------------------------------------+
15:28:31  | db-cluster                | https://cloudius-jenkins-test.s3.amazonaws.com/756aa4c9-80f8-4a77-acd2-c521dc189bdb/20240426_114508/db-cluster-756aa4c9.tar.gz                |
15:28:31  | sct-runner-events         | https://cloudius-jenkins-test.s3.amazonaws.com/756aa4c9-80f8-4a77-acd2-c521dc189bdb/20240426_114508/sct-runner-events-756aa4c9.tar.gz         |
15:28:31  | sct-runner-python-log     | https://cloudius-jenkins-test.s3.amazonaws.com/756aa4c9-80f8-4a77-acd2-c521dc189bdb/20240426_114508/sct-756aa4c9.log.tar.gz                   |
15:28:31  | loader-set                | https://cloudius-jenkins-test.s3.amazonaws.com/756aa4c9-80f8-4a77-acd2-c521dc189bdb/20240426_114508/loader-set-756aa4c9.tar.gz                |
15:28:31  | monitor-set               | https://cloudius-jenkins-test.s3.amazonaws.com/756aa4c9-80f8-4a77-acd2-c521dc189bdb/20240426_114508/monitor-set-756aa4c9.tar.gz               |
[...]

PR #7377, opened 2024-04-26
fix(set-tags): fix setting of tags onto SCT runners

vponomaryov:
Fixes:
- GCE, translate boolean tag value to str with lowercase.
- Azure, use proper API method for setting tags - `update_at_scope`. It will allow us to fix following error:
```
    (InvalidRequestContent) The request content was invalid and could not \
      be deserialized: 'Could not find member 'operation' on object of type \
        'TagsDefinition'. Path 'operation', line 1, position 13.'.
```
- All, make `logs_collected` bool attr have correct value instead of the `false positive` all the time where any non-empty str value was considered to be `True`.

Fixes: #7375
Fixes: #7376

### Testing

- [scylla-staging/valerii/vp-longevity-10gb-3h-test#79](https://argus.scylladb.com/test/631c0991-5548-434e-b7a3-f47eb2db8777/runs?additionalRuns[]=96d5419b-a234-4fcd-9072-11132090f56a)
- [scylla-staging/valerii/vp-longevity-10gb-3h-gce-test#22](https://argus.scylladb.com/test/f20862ff-cf95-44cf-8d5a-78da3d9b1f0e/runs?additionalRuns[]=76c4526b-3c91-4279-afdf-d177929dd46f)
- [scylla-staging/valerii/vp-longevity-10gb-3h-azure-test#25](https://argus.scylladb.com/test/bc3f0da1-8e01-4885-9d0c-687b1b64ac7e/runs?additionalRuns[]=3cde7ee0-a86b-4eb5-a9b0-39644af6adf5)

### Reminders

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -182,6 +182,16 @@ def generate_random_string(length):
         random.choice(string.ascii_uppercase + string.digits) for x in range(length - 1))
 
 
+def str_to_bool(bool_as_str: str) -> bool:
+    if isinstance(bool_as_str, bool):
+        return bool_as_str
+    elif isinstance(bool_as_str, str):
+        return bool_as_str.lower() in ("true", "yes", "y", "1")
+    elif bool_as_str is None:
+        return False
+    raise ValueError("'bool_as_str' is of the unexpected type: %s" % type(bool_as_str))
+
+
 def get_data_dir_path(*args):
     sct_root_path = get_sct_root_path()
     data_dir = os.path.join(sct_root_path, "data_dir", *args)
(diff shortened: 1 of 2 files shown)


sdcm/utils/common.py
```diff
@@ -182,6 +182,16 @@ def generate_random_string(length):
         random.choice(string.ascii_uppercase + string.digits) for x in range(length - 1))
 
 
+def str_to_bool(bool_as_str: str) -> bool:
```
soyacz, line 185: could reuse: `from distutils.util import strtobool`

vponomaryov: It is deprecated: https://peps.python.org/pep-0632/#migration-advice

soyacz (approved):
LGTM

sdcm/utils/common.py
fruch: Following this change, we should replace other usages as well, we have one in sct_config.py

fruch (approved):
LGTM

Issue #7364, opened 2024-04-24
Internode compression test


soyacz:
Because of new internode_compression algorithms that are more advanced (e.g. rebuilding dictionary for compression each hour and redistributing it) we need to turn it on for some longevity tests and verify how it works under load at scale for longer periods (at least 1 day).

soyacz:
@mykaul I couldn't find details about this new algorithm, please share a link to proper PR/issue

cc @ShlomiBalalis

fruch:
We don't have any longevity using it (only scale and rolling upgrades)

So we should enable it and have a run without those new features, so we can compare both

dimakr:
As a sanity check for the feature we selected the `[longevity-100gb-4h-test](https://jenkins.scylladb.com/job/scylla-enterprise/job/longevity/job/longevity-100gb-4h-test/)` case, to be executed with the following set of dict training parameters values (+ `internode_compression` enabled):
```
rpc_dict_training_min_time_seconds: 900
rpc_dict_update_period_seconds: 150
rpc_dict_training_min_bytes: 100000000
internode_compression_zstd_max_cpu_fraction: 0.05
rpc_dict_training_when: "when_leader"
```
The [longevity-100gb-4h-test + dict training, with cpu_fraction 0.05](https://jenkins.scylladb.com/job/scylla-enterprise/job/longevity/job/longevity-100gb-4h-test/) build failed on disrupt_multiple_hard_reboot_node disruption, but results analysis shows that the dict training feature itself didn't cause the fail.
The analysis summary:

> 1. The failure is indeed unrelated to the feature. The cause of that failure was that, after a reboot of one of the nodes, the restarted node attempts to serve client queries before it sees that other nodes are UP — so it fails due to unavailability. Since this is a single-node problem, the system should survive it because the clients should retry with other coordinators and the other coordinators should succeed. But apparently they didn't. My main suspicion in this case is this is actually a misconfiguration of the clients, who don't retry properly. (Because this isn't the first time I see such a problem, and the last time I checked the clients weren't properly configured for high availability). But theoretically it could also be caused by a driver problem, bad metadata sent from servers to client, or maybe the restarted node started serving earlier than expected.
> 2. The feature was behaving as expected. Just, as far as performance effects go, the workload isn't very interesting, because it sends mostly random data, which cannot be compressed

[...]

PR #7401, opened 2024-05-06
feature(advanced RPC compression): enable dictionary training and ZSTD compression

michoecho:
This patch:
- enables dictionary training (by changing `rpc_dict_training_when` from the default value of 'never' to 'when_leader),
- increases (by 4x) the frequency of dictionary training and updates, to stress the feature more,
- enables ZSTD internode compression (provided that `internode_compression` is also enabled) by giving it a nonzero CPU usage limit.

However, this patch doesn't enable `internode_compression` itself, because it's a performance-affecting option. To actually put the feature to use, a test must enable `internode_compression` on its own.

### Testing

- Syntax check.
- Untested.

Refs #7364


michoecho:
Marking as draft, because I haven't tested it.

This PR enables (I hope) dictionary training by default. It should be non-invasive, so there should be no reason to enable it only for a subset of tests.

The next step would be to also enable `internode_compression: all` for some longevity tests. It can't be enabled by default for all tests because it could disturb performance tests.

michoecho:
I have no idea how to test this.

fruch:
> I have no idea how to test this.

I've enabled the labels, so we can have a basic sanity for it (even that it default to running 5.2, we'll need to run it again with `latest:master`)

but you should define a specific cases you want to try, i.e. a longevity that you can have a reference of the run to compare with
for example teir1 (which are run on a weekly basis with master):

https://argus.scylladb.com/workspace?state=WyJlMTdmZjM5YS1mM2Y2LTRhYzYtYTE2MS04NzEyNjdmZDM5NTUiLCI3ZjYwNDIxMy01Y2FkLTQ0MTgtOTIxZS0xY2E5ODkxODA3MWMiLCI5OTY1MDFhZC1jMTM3LTQwZWYtOGRjYy1hN2QxMmYwYmEzYjYiLCI0ODEyMmRmZS0xNDhiLTRhOTktYTQzYy0xYTcxYzY4Njk5NzMiLCI0MjhiMmZiMy1lOWFjLTQyNzYtYjcxYS05YmY0YjQ1YTQzMWEiLCI5ODA1MDczMi1kZmUzLTQ2NGMtYTY2YS1mMjM1YmFkMzA4MjkiLCJlY2Q0OTdjMC04MmQ2LTQyNjktYjA1My1mNWMyMTU3ZTA0YWUiLCIxZTMzM2RmOC1hN2U4LTQxNzEtOGFiNy0xZDdiZGVhOTA3ZDUiXQ

you can start with a short variations, of some of those, like `longevity-100gb-4h` which also runs quite regularly

sdcm/provision/scylla_yaml/scylla_yaml.py
```diff
@@ -147,6 +147,11 @@ def set_endpoint_snitch(cls, endpoint_snitch: str):
     internode_send_buff_size_in_bytes: int = None  # 0
     internode_recv_buff_size_in_bytes: int = None  # 0
     internode_compression: Literal['none', 'all', 'dc'] = None  # "none"
+    rpc_dict_training_when: Literal['always', 'never', 'when_leader'] = 'when_leader'
```
fruch, line 150: I would recommend putting the actual defaults here (i.e. None)

and create an SCT configuration for specific test that would enabled it, for the sake of testing.
enabling it for all by default, should preferably done by scylla-core itself if decided as such.

see the `configurations` folder, and as example `configurations/tablets-initial-32.yaml` how a specific test can set it's own scylla.yaml parameters

michoecho:
> but you should define a specific cases you want to try, i.e. a longevity that you can have a reference of the run to compare with
for example teir1 (which are run on a weekly basis with master):

@fruch Do you mean that I should manually clone the relevant jenkins job and run it, or is there some better way?

sdcm/provision/scylla_yaml/scylla_yaml.py
michoecho: > I would recommend putting the actual defaults here (i.e. None)
> 
> and create an SCT configuration for specific test that would enabled it, for the sake of testing. enabling it for all by default, should preferably done by scylla-core itself if decided as such.

Why? The good thing about enabling it by default is that it gives more coverage with little effort. What's the drawback?

fruch:
> > but you should define a specific cases you want to try, i.e. a longevity that you can have a reference of the run to compare with
> > for example teir1 (which are run on a weekly basis with master):
> 
> @fruch Do you mean that I should manually clone the relevant jenkins job and run it, or is there some better way?

no there no better way, you can try out the new Argus clone command:
https://argus.scylladb.com/test/f05fea04-eb74-4961-94fb-f71c67df52cb/runs?additionalRuns[]=c221d97d-1b34-4c27-b75a-283abf8bd790

sdcm/provision/scylla_yaml/scylla_yaml.py
fruch: the drawback is that we won't be testing the defaults from that point onwards.

if it's zero risk, change the default in scylla, if not the are reasons not to do so, they apply here in SCT as well.

michoecho: > I would recommend putting the actual defaults here (i.e. None)
> and create an SCT configuration for specific test that would enabled it, for the sake of testing.

@fruch 

In that case, would it be acceptable to enable dict training and internode compression for all *longevity* tests?

We can't enable this feature by default in Scylla because it has a performance cost. For some users, RPC compression is useless (e.g. because they have an on-premise deployment and they don't pay for the amount of data sent over network), and we don't want to regress them.

But we do want it to be stable and relatively cheap "by default". We might want to enable it by default in the cloud, because every cluster pays the network costs there.

> the drawback is that we won't be testing the defaults from that point onwards.

Internode compression only adds code paths, it doesn't remove them. So when you enable it, you are testing strictly more code. And I think that's a good thing?

soyacz:
In upgrade tests we start from 2024.1 or 5.4 - how this change affect these versions? Ignoring or failing?

michoecho:
> In upgrade tests we start from 2024.1 or 5.4 - how this change affect these versions? Ignoring or failing?

I don't understand the question. None of these options are present in 2024.1/5.4, so it shouldn't affect their behaviour at all.

fruch:
> > In upgrade tests we start from 2024.1 or 5.4 - how this change affect these versions? Ignoring or failing?
> 
> @[#7401 (comment)](https://github.com/scylladb/scylla-cluster-tests/pull/7401#issuecomment-2101998012) I don't understand the question. None of these options are present in 2024.1/5.4, so it shouldn't affect their behaviour at all.

I think @soyacz meant that we enable internode compression in rolling upgrades from 5.4/2024.1
and we can run those cases to see the effect, and also the effect of those configuration before and after quite clearly.

sdcm/provision/scylla_yaml/scylla_yaml.py
michoecho: @fruch ^Ping.

fruch: so you are saying it's not gonna be enabled by default for core, cause of fear of regression.
so it might cause regression on some of the tests

and for the "cheap" argument, every time we said something like that, we regretted it (see the enabling audit by default, we after a while got reverted)

I would still argue you do want specific case that can show different compression option vs. other options

and in this PR, you are not enabling the internode_compression at all, so it would only affect some cases using it.
which are not that many cases

fruch: Anyhow, we should be trying it on a few cases before merging it
- one short longevity
- one case of rolling upgrades

as followup work after merge:
- one performance test - during operation(nemesis) one

kbr-scylla: >  so it might cause regression on some of the tests

which is good, because it means we detected a bug, no?

fruch: @kbr-scylla it means who owns the feature so do that leg work thinking/testing, not some other poor guy that happen to be trying do something else.

if we want to cover things, it should be systematic, not completely random.
people should be building test cases for their features. 

"lets enable all features all the time", kind of like let's default to RBNO,  and who cares about streaming from that point onwards....

fruch:
this change was covered in
https://github.com/scylladb/scylla-cluster-tests/pull/7923

it was enabled in two tests, and can be extended to others as needed

PR #7393, opened 2024-05-02
fix(getJobTimeouts.groovy): Increase testStartupTimeout to 80 minutes

k0machi:
This fixes an issue with certain jobs that spend a lot of time on
prepare stress stage / setup stage, as current calculation results in a
way closer timeout to the timeout thread, leading to interruptions

Fixes #7385

### Testing

- [ ]

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #7403, opened 2024-05-06
Revert "fix(getJobTimeouts.groovy): Increase testStartupTimeout to 80 minutes"

fruch:
Reverts scylladb/scylla-cluster-tests#7393


fruch:
we found that there's a different issue with the specific test, and duration of s-b commands

PR #7409, opened 2024-05-08
CentOS7 deprecation adjustments for manager tests

mikliapko:
The PR addresses the CentOS7 deprecation changes:
- sanity tests switched to use Ubuntu image instead of centos;
- ipv6 tests switched to Ubuntu as well.

Addresses https://github.com/scylladb/scylla-manager/issues/3840

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/sct/job/manager-ubuntu-22-sanity/7/
- [x] https://argus.scylladb.com/workspace?state=WyIzMTI3MTYyMS00ZGVkLTRmNDItYmYzYS1hMWE4OTFiNDUwODkiXQ

### Reminders

### DIFF
--- a/configurations/manager/ubuntu22.yaml
+++ b/configurations/manager/ubuntu22.yaml
@@ -1,2 +1,0 @@
-ami_id_monitor: 'ami-09db26f1ef0a9f406'  # Canonical, Ubuntu, 22.04 LTS, amd64 jammy image build on 2022-05-06
-ami_monitor_user: 'ubuntu'
--- a/jenkins-pipelines/manager/ubuntu22-manager-sanity-ipv6.jenkinsfile
+++ b/jenkins-pipelines/manager/ubuntu22-manager-sanity-ipv6.jenkinsfile
@@ -1,0 +1,15 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+managerPipeline(
+    backend: 'aws',
+    region: 'us-west-1',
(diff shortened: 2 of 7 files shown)


configurations/manager/ubuntu22.yaml
```diff
@@ -1,2 +1,2 @@
-ami_id_monitor: 'ami-09db26f1ef0a9f406'  # Canonical, Ubuntu, 22.04 LTS, amd64 jammy image build on 2022-05-06
+ami_id_monitor: 'scylladb-monitor-4-6-2-2024-02-13t08-06-04z'
```
fruch, line 1: It's a bit problematic

Jobs using this config would stick with it, 

And won't ever get an updated monitor image, and at some point you'll be missing some monitoring features

mikliapko: Why it was set as default in `defaults/aws_config.yaml` in such case?

mikliapko: And should I then just leave `ami-09db26f1ef0a9f406` image without changes?

fruch: The idea is that there are specific places where we set the default image per backend

If we clone it to more places no one would update them when there's a new version of it

So I would recommend removing it from here, but keep in mind it might get updated with a different version at some point with you knowing about it

mikliapko: ah, okay, I got your point. I misunderstood what you meant above. My idea was to specify it explicitly to make visible without going deeper into additional config files. But I agree it can lead to sticking with this image for a long time. Will update the PR.

mikliapko: Fixed

fruch (approved):

PR #7415, opened 2024-05-09
test(perf): i4i.4xlarge perf with split ranges

soyacz:
We want each c-s to operate on own range so processes don't write the same rows. Also increase datasize and instance type (to minimize noisy-neighbor problem). Splitting manually to avoid too much code changes and risk regressions (added script for doing the split). Read&mixed test are aimed to close to 100% cache hit ratio. Used as a separate test class to not affect current tests. Also this test won't use node benchmarks for now, as there's no proof for relation between benchmark result and test result.

refs: https://github.com/scylladb/qa-tasks/issues/1508

### Testing

- [ ] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/perf-regression-throughput-non-shard-aware-i4i-4xlarge-test-650gb/52/

### Reminders

### DIFF
--- a/configurations/disable-ear.yaml
+++ b/configurations/disable-ear.yaml
@@ -1,1 +1,1 @@
-scylla_encryption_options: "{'key_provider': 'none'}"
+scylla_encryption_options: '{"key_provider": "none"}'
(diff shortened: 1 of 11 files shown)


soyacz:
closing as won't be needed for now.

PR #7420, opened 2024-05-12
feature(teardown-validators): validator to work on test context

fruch:
since we have more types of validators which are not depended on db cluster alone, we change the parameter to validators to be a the tester instance, and not a db cluster as before.

### Testing

- [ ]

### Reminders


fruch:
@soyacz @dimakr 

I need this for kafka related, so I can validate the results of the kafka reader thread

And I think this would fit also the parallel nemesis validator that @dimakr is working on

fruch:
Not sure it needed, closing for now

Issue #7430, opened 2024-05-14
cannot run cassandra-stress because it is not found


soyacz:
nightly images stopped to ship scylla-tools package. This made our approach of using latest nightly docker image for c-s obsolete.

We need to find another way of shipping c-s to our loader nodes.

## Packages

Scylla version: `5.5.0~dev-20240513.2ce643d06bf0` with build-id `2e2c89cbb469c1231861753c4af823040a31579e`

Kernel Version: `5.15.0-1060-aws`

OS / Image: `ami-0de7228fdbd40d174` (aws: undefined_region)

Test: `longevity-100gb-4h-test`
Test id: `b2d93c9e-ebec-4409-8a4f-047a38c7daae`
Test name: `scylla-master/longevity/longevity-100gb-4h-test`
Test config file(s):

- [longevity-100gb-4h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/29885d9894813da0cd9688cc95e09fb6cd0f1569/test-cases/longevity/longevity-100gb-4h.yaml)

<details>
<summary>
Logs and commands
</summary>

- Restore Monitor Stack command: `$ hydra investigate show-monitor b2d93c9e-ebec-4409-8a4f-047a38c7daae`
- Restore monitor on AWS instance using [Jenkins job](https://jenkins.scylladb.com/view/QA/job/QA-tools/job/hydra-show-monitor/parambuild/?test_id=b2d93c9e-ebec-4409-8a4f-047a38c7daae)
- Show all stored logs command: `$ hydra investigate show-logs b2d93c9e-ebec-4409-8a4f-047a38c7daae`

## Logs:
- **db-cluster-b2d93c9e.tar.gz** - [https://cloudius-jenkins-test.s3.amazonaws.com/b2d93c9e-ebec-4409-8a4f-047a38c7daae/20240514_002840/db-cluster-b2d93c9e.tar.gz](https://cloudius-jenkins-test.s3.amazonaws.com/b2d93c9e-ebec-4409-8a4f-047a38c7daae/20240514_002840/db-cluster-b2d93c9e.tar.gz)
[...]

PR #7431, opened 2024-05-14
fix(c-s): use 5.4.6 image for c-s

soyacz:
Recently scylla-tools package is not installed in scylladb docker images. This makes c-s not being installed in docker images we use for loaders.

Workaround by pinning version with c-s installed.

workarounds: https://github.com/scylladb/scylla-cluster-tests/issues/7430

### Testing

- [ ] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-ScyllaKillMonkey-aws-test/11/

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -209,7 +209,7 @@ stress_image:
   ndbench: 'scylladb/hydra-loaders:ndbench-jdk8-20210720'
   ycsb: 'scylladb/hydra-loaders:ycsb-jdk8-20220918'
   nosqlbench: 'scylladb/hydra-loaders:nosqlbench-4.15.49'
-  cassandra-stress: '' # default would be same version as scylla under test
+  cassandra-stress: 'scylladb/scylla:5.4.6' # empty default would be same version as scylla under test
   scylla-bench: 'scylladb/hydra-loaders:scylla-bench-v0.1.20'
   gemini: 'scylladb/hydra-loaders:gemini-v1.8.6'
   alternator-dns: 'scylladb/hydra-loaders:alternator-dns-0.1'


dimakr (approved):
LGTM

roydahan:
According to our readme, this shouldn't have happened and the fallback should have work without pinning this version:

```
### Finding correct cassandra-stress image

Since we are going to be using scylla docker images for loaders, we need a way to find the correct docker image, based on a
given version or sha of scylla (both for OSS and Enterprise versions), if we'll find that match base on sha would be problematic
we'll fall back to last available release (smaller than the currently under test version, ex. when testing 5.2-dev, we'll take 5.1.x version)
```
We need to understand where is the issue with searching for the correct docker image to be used.

Meanwhile, we got the same issue for 5.4 and we probably will continue getting it in many places.
Not sure why it was decided as "backport/none".

fruch:
> @roydahan We should start working on a release schedule for c-s on its own.

that's why.

it was intended as a temporary measure, cause we didn't planned anything for this removal of c-s.
and we won't plan until we'll have the first c-s packaged and released

roydahan:
It's not related.
First, it's an issue in SCT that should find the correct the docker per release.
Second, even if we didn't fix the issue and we have a workaround, we should backport it to the relevant branches.
Third, the release of c-s on its own will not happen very soon, there are still some decisions that haven't been made there.

fruch:
> It's not related.
> First, it's an issue in SCT that should find the correct the docker per release.

Raise an issue

> Second, even if we didn't fix the issue and we have a workaround, we should backport it to the relevant branches.

The workaround was known to be need on master, no anywhere else, if you backported, releases would use some random version out of master... but go right ahead and backport it

> Third, the release of c-s on its own will not happen very soon, there are still some decisions that haven't been made there.
> 

removing c-s from images was a mistake 
FYI so many examples gonna break (and our home exercise as well)

I think it should be revert until the repackage is done. 

I'm tired of attenting to all those fallouts

PR #7365, opened 2024-04-24
fix(manager): upgrade Scylla version used in Manager to latest

mikliapko:
Upgraded the version of Scylla used in Manager from 2022.1 (2021 for debian10) to the latest 2024.1 (2023 for debian10).

Should be merged together with https://github.com/scylladb/scylla-cluster-tests/pull/7435

### Testing

- [x] The [CI run](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/sct/job/manager-centos-sanity/12/) from personal fork;
- [x] Argus [link](https://argus.scylladb.com/workspace?state=WyIxMzc2ODAyZC0xN2YxLTRiM2MtODhiMy0zNzcwOWI5NDMzZGYiXQ).

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #7435, opened 2024-05-14
Switch manager tests to run on singleDC environment

mikliapko:
Closes https://github.com/scylladb/scylla-manager/issues/3850

Should be merged together with https://github.com/scylladb/scylla-cluster-tests/pull/7365

Since there is an issue with multiDC cluster restore when the EaR is turned on (https://github.com/scylladb/scylla-manager/issues/3829), it was decided to:
- switch them to use single-DC cluster;
- leave one test with multiDC cluster for enterprise version 2022 (no encryption feature implemented).

After https://github.com/scylladb/scylla-manager/issues/3829 resolution, multiDC cluster would be returned back.

### Testing

- [x] Jenkins jobs:
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/sct/job/manager-master-versions-update/job/sct-feature-test-backup/
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/sct/job/manager-master-versions-update/job/ubuntu22-sanity-test/
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/sct/job/manager-master-versions-update/job/older-enterprise-ami-sanity-test/

### Reminders

### DIFF
--- a/jenkins-pipelines/manager/centos-manager-backup-200gb.jenkinsfile
+++ b/jenkins-pipelines/manager/centos-manager-backup-200gb.jenkinsfile
@@ -5,9 +5,9 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 
 managerPipeline(
     backend: 'aws',
-    region: '''["us-east-1", "us-west-2"]''',
+    region: 'us-west-2',
     test_name: 'mgmt_cli_test.MgmtCliTest.test_backup_feature',
-    test_config: '''["test-cases/manager/manager-regression-multiDC-set-distro.yaml", "configurations/manager/200gb_dataset.yaml"]''',
+    test_config: '''["test-cases/manager/manager-regression-singleDC-set-distro.yaml", "configurations/manager/200gb_dataset.yaml"]''',
 
     post_behavior_db_nodes: 'destroy',
     post_behavior_loader_nodes: 'destroy',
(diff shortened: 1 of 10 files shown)


jenkins-pipelines/manager/centos-manager-backup-200gb.jenkinsfile
```diff
@@ -5,9 +5,9 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 
 managerPipeline(
     backend: 'aws',
-    region: '''["us-east-1", "us-west-2"]''',
+    region: 'us-west-2',
```
fruch, line 8: it's on purpose you are putting each case on a different region ?

(if you can, you can even use `random`, I've fixed to be working a while back)

fruch (approved):
LGTM

mgmt_cli_test.py
```diff
@@ -349,9 +349,8 @@ def test_manager_sanity(self):
             self.test_mgmt_cluster_crud()
         with self.subTest('Mgmt cluster Health Check'):
             self.test_mgmt_cluster_healthcheck()
-        # test_healthcheck_change_max_timeout requires a multi dc run. And since ipv6 cannot run in multi dc, this test
```
fruch, line 352: good that this comment is going away, cause the ipv6 part of it, is alway wrong. we can use ipv6 on AWS with multi-dc cases.

jenkins-pipelines/manager/centos-manager-backup-200gb.jenkinsfile
mikliapko: I randomly chose one of the regions supported before in multiDC runs: "us-east-1" or "us-west-2"

fruch: Just to explain, setting the region as random, can help a bit with the spot issues, since tests would spread to all of the supported regions, and won't all run on the exact same ones

fruch (approved):

PR #7436, opened 2024-05-14
feature(tls-certs): create client and server certs/keys for each test run

dimakr:
All TLS/SSL keys and certificates used in SCT were created manually around 2016, both the client and server ones.
This prevents from using newer TLS versions with certification checks (which should be default), as IPs/DNS node identifiers are not configured in certificates.

The change reworks the logic of working with SSL/TLS, so that for each test run:
- a new root CA certificate/key pair is created
- individual certificate/key pairs are created for each server and client nodes, certificates are signed by the CA
- SAN extension is added to certificates, to include IP and DNS identifiers of a node
- CA cert is also converted to Java truststore, which is required for cassandra-stress command with enabled encryption

Additionally, the change adds a possibility to enable mutual TLS in SCT configuration. TLS parameters of c-s, s-b and latte stress commands are updated appropriately, when mtls is enabled.

Old certificates are deleted from SCT as no longer needed.

### Testing
- [x] [PR-provision-test config + server_encrypt + client_encrypt + mtls](https://argus.scylladb.com/test/b3fab77b-7384-4801-950f-31c2983c3ae7/runs?additionalRuns[]=31524803-dfe6-4736-b28d-3f54e973a2d8)
- [x] [PR-provision-test config + server_encrypt + client_encrypt](https://argus.scylladb.com/test/b3fab77b-7384-4801-950f-31c2983c3ae7/runs?additionalRuns[]=c030390c-c5c5-4f86-8800-6ba9a8d3eeb5)
- [x] [NodeTerminateAndReplace nemesis +  server_encrypt + client_encrypt](https://argus.scylladb.com/test/b3fab77b-7384-4801-950f-31c2983c3ae7/runs?additionalRuns[]=cbce241b-bee1-45af-b060-2ee61210d8ce)
- [x] [DecommissionMonkey nemesis +  server_encrypt + client_encrypt](https://argus.scylladb.com/test/b3fab77b-7384-4801-950f-31c2983c3ae7/runs?additionalRuns[]=2f25faee-9a9f-4d75-8f83-ac31dabb8588)
- [x] [SslHotReloadingNemesis nemesis +  server_encrypt + client_encrypt](https://argus.scylladb.com/test/b3fab77b-7384-4801-950f-31c2983c3ae7/runs?additionalRuns[]=d44c5d2c-e047-4f1a-8523-312c958bd547)
- [x] [artifacts-ami-test +  server_encrypt + client_encrypt](https://argus.scylladb.com/test/dce06acb-820c-4660-af39-241e50175c80/runs?additionalRuns[]=9b4df039-e130-4687-a45c-f196f2f58b10)

### Reminders


fruch:
@dimakr direction looks good, but we need to rebuild the hydra docker image for this one

maybe that what fails the unittests

fruch:
- [ ] remove the old certificates related code

sdcm/provision/helpers/certificate.py
```diff
@@ -46,3 +67,187 @@ def install_encryption_at_rest_files(remoter):
...
+        f'-destkeystore {jks_file} -deststoretype JKS -deststorepass {jks_password} -destkeypass {jks_password}')
+    subprocess.run(cmd, shell=True, check=True)
+
+
+def cleanup_ssl_config():
+    """Remove SCT ssl_config files that are dynamically created during the test"""
+    for file in [CA_CERT_FILE, CA_KEY_FILE, SERVER_CSR_FILE, CLIENT_CERTFILE, CLIENT_KEYFILE, CLIENT_TRUSTSTORE,
+                 CLIENT_KEYSTORE_FILE, SERVER_CERT_FILE, SERVER_KEY_FILE, SERVER_KEYSTORE_FILE, DER_CERT_FILE,
+                 JKS_TRUSTSTORE_FILE, PKCS12_KEYSTORE_FILE]:
+        if os.path.exists(file):
```
fruch, line 218: use `Path.unlink(missing_ok=False)`

also maybe save those as Path object to begin with ?

sdcm/tester.py
```diff
@@ -2854,6 +2860,10 @@ def tearDown(self):
         self.stop_event_device()
         if self.params.get('collect_logs'):
             self.collect_sct_logs()
+        cleanup_ssl_config()
+        # Revert cqlshrc to original state
+        update_cqlshrc.cache_clear()
```
fruch, line 2865: why clear this, and why update the cqlsh file at the end of the test ? not sure what do we gain by doing so ?

dimakr: Cleaning up is to not commit SSL artifacts that were created at the beginning of the test - CA certificates, keys, Java keystore - so that a person won't need to remove them manually, if works with SCT locally.

The same for cqlshrc - if a test configuration enables hostname validation, we rollback the attribute value after the test, to not commit the change.

fruch: we might want to limit this to when it running locally, in CI it's just a waste of time 

we should just add those to .gitignore from this change onwards

sdcm/cluster.py
```diff
@@ -4991,8 +5012,12 @@ def node_setup(self, node, verbose=False, **kwargs):  # pylint: disable=unused-a
             self.log.info("Don't install anything because bare loaders requested")
             return
 
-        if self.params.get('client_encrypt'):
-            node.config_client_encrypt()
+        with NODE_CONFIG_SETUP_LOCK:
```
soyacz, line 5015: why lock is needed?

dimakr: This is to correctly distribute individual certificates per cluster node.
The individual certificates are named the same across the nodes (the unique info such as IPs, dns names added to certificates via its SAN extension). And as the cluster nodes are created in parallel their own threads, without the lock there are situations when thread-1 creates a cert for node-1, and thread-2 creates its cert for node-2, then its time for thread-1 to copy cert-1 to node-1, but there is already cert-2.

soyacz: can we use different names or directories for each node?

soyacz: > Using different names and/or folders won't be optimal, as we would need to set different names and locations of keys/certs in server|client_encryption_details parameters in scylla.yaml, for each server.

From my understanding, we create these certs on sct-runner and then copy to server, why it cannot be created  in a 
subdir and copied to common location on server, so scylla_yaml is the same for all the nodes?

soyacz: and question, maybe to not overengineer, how much time it takes to copy&install these certs?

dimakr: > From my understanding, we create these certs on sct-runner and then copy to server, why it cannot be created in a subdir and copied to common location on server, so scylla_yaml is the same for all the nodes?

This is exactly what is performed now:
1. Before the locked section the `/etc/scylla/ssl-config` dir has only CA cert
2. Lock is acquired
4. db-to-db cert/key and db-to-client cert/key, unique for a db node, are created in the dir.
3. `/etc/scylla/ssl-config` dir is copied as a whole to a server
4. Lock is released
5. ... Repeat for other node

If I understand correctly, the concern for current approach was raised due to ```3. `ssl-config` dir is copied as a whole to a server``` step was a part of node.node_setup logic, which I also added to locked section and that is not a right approach (we would not want to lock node_setup flow).
This can be reworked so that only creating and copying operations are locked. For that, copying part should be moved out of node.node_setup, and this way node.nod_setup remains paralleled as it was.

soyacz: I think you're mixing dirs :)
`CA_CERT_FILE = get_data_dir_path('ssl_conf', 'ca.pem')` has CA cert
and we create unique per node in data dir path - that's why we cannot run in parallel.
If per node cert files were in `get_data_dir_path('ssl_conf, self.name, '<cert>') then we could run everything in parallel

configurations/temp_encrypt_config.yaml
fruch, line 1: remember to remove

sdcm/tester.py
```diff
@@ -2020,7 +2023,7 @@ def run_stress_thread_bench(self, stress_cmd, duration=None, round_robin=False,
         stop_test_on_failure = False if not self.params.get("stop_test_on_stress_failure") else stop_test_on_failure
 
         if self.params.get("client_encrypt") and ' -tls' not in stress_cmd:
-            stress_cmd += ' -tls '
+            stress_cmd += f' -tls -tls-ca-cert-file {SCYLLA_SSL_CONF_DIR}/ca.pem '
```
fruch, line 2026: let move it into scylla-bench own code

sdcm/tester.py
```diff
@@ -2909,6 +2912,7 @@ def tearDown(self):
         self.stop_event_device()
         if self.params.get('collect_logs'):
             self.collect_sct_logs()
+        cleanup_ssl_config()
```
fruch, line 2915: you probably want to add @silence ontop of it.

so it never gonna fail the test

unit_tests/test_cassandra_stress_thread.py
```diff
@@ -84,9 +84,10 @@ def test_03_cassandra_stress_client_encrypt(request, docker_scylla, params):
     loader_set = LocalLoaderSetDummy()
 
     cmd = (
-        """cassandra-stress write cl=ONE duration=1m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=1) """
-        """compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native """
-        """-rate threads=10 -pop seq=1..10000000 -log interval=5"""
+        "cassandra-stress write cl=ONE duration=1m -schema 'compaction(strategy=SizeTieredCompactionStrategy) "
+        "replication(strategy=NetworkTopologyStrategy,replication_factor=1)' -mode cql3 native -rate threads=10 "
+        "-pop seq=1..10000000 -log interval=5 "
+        "-transport 'truststore=/etc/scylla/ssl_conf/truststore.jks truststore-password=cassandra'"
```
dimakr, line 90: Check if CassandraStressThread adds `-transport` parameter properly, when is instantiated in unit test (it should) and remove this hardcoded parameter.

sdcm/cluster.py
```diff
@@ -5006,7 +5031,12 @@ def node_setup(self, node, verbose=False, **kwargs):  # pylint: disable=unused-a
             self.log.info("Don't install anything because bare loaders requested")
             return
 
+        node.create_node_certificate(node.ssl_conf_dir / 'test.crt', node.ssl_conf_dir / 'test.key')
```
soyacz, line 5034: so, when using docker based loaders we send certs, why creating them for vm based loaders? Shouldn't they also be sent later when created?
Please add a test with param `use_prepared_loaders: true` so we are sure it works in this case as well.

dimakr: Added check for `use_prepared_loaders` in LoaderSet.node_setup, to put certificate to loader VM when this flag is enabled. The test for this is below (I added a temporary test config with this flag enabled to run with PR-provision test, as existing config with the flag are 5h+ long).
[PR-provision-test config + use_prepared_loaders + server_encrypt + client_encrypt + mtls](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/certs-rework-tests/24/)

Repeated a test where docker based loader is used:
[DecommissionMonkey nemesis + server_encrypt + client_encrypt](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/certs-rework-tests/25/)

sdcm/scylla_bench_thread.py
```diff
@@ -145,14 +146,31 @@ def verify_results(self):
...
+    def create_stress_cmd(self, stress_cmd, cmd_runner):
         if self.connection_bundle_file:
             stress_cmd = f'{stress_cmd.strip()} -cloud-config-path={self.target_connection_bundle_file}'
         else:
             # Select first seed node to send the scylla-bench cmds
             ips = ",".join([n.cql_address for n in self.node_list])
             stress_cmd = f'{stress_cmd.strip()} -nodes {ips}'
 
+            if self.params.get("client_encrypt"):
+                loader = self.loader_set.get_loader()
```
fruch, line 158: I rather not use `get_loader()` function, it's would interfere with the order of selections of loader nodes.

just pick the first one, if it's not important to have this specific one

fruch: you can pass in the loader from where this function is called

sdcm/scylla_bench_thread.py
```diff
@@ -145,14 +146,31 @@ def verify_results(self):
...
+                    if ssl_file.is_file():
+                        cmd_runner.send_files(str(ssl_file),
+                                              str(SCYLLA_SSL_CONF_DIR / ssl_file.name),
+                                              verbose=True)
+                stress_cmd = f'{stress_cmd.strip()} -tls -tls-ca-cert-file {SCYLLA_SSL_CONF_DIR}/ca.pem'
+
+                if self.params.get("client_encrypt_mtls"):
+                    stress_cmd = (f'{stress_cmd.strip()} -tls-client-key-file {SCYLLA_SSL_CONF_DIR}/test.key '
+                                  f'-tls-client-cert-file {SCYLLA_SSL_CONF_DIR}/test.crt')
+                if self.params.get("validate_hostname"):
```
fruch, line 169: there's not such SCT configuration option, and if we do introduce it, we might want a better name for it.

fruch: we can leave the whole code under comment, until someone would be addressing it.

configurations/temp_cert_test_config.yaml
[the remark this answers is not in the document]
dimakr: done

sdcm/scylla_bench_thread.py
dimakr: done

sdcm/scylla_bench_thread.py
dimakr: addressed - hostname validation control is to be implemented after it's fixed in s-b

fruch (approved):
LGTM

soyacz:
congrats for finishing this PR @dimakr !

PR #7444, opened 2024-05-16
fix(stress): define duration in *-large-partition-200k_pks-4days.yaml

vponomaryov:
The configuration located at
`longevity-large-partition-200k_pks-4days.yaml` file contains dozens of stress commands and only small part of it contains the `duration` parameter.
Commands which don't have that param run without time limit causing CI job to reach the JOB timeout.

So, set the duration param to make stress commands be finite.

### Testing

- [ ]

### Reminders


vponomaryov:
Example of the unlimited stress cmd run is here: [scylla-master/tier1/longevity-large-partition-200k-pks-4days-gce-test#14](https://argus.scylladb.com/test/996501ad-c137-40ef-8dcc-a7d12f0ba3b6/runs?additionalRuns[]=b3cc439d-30f3-4663-98fc-2c469620603a)
The job was configured to run `24h`, but stress cmds ran more before the job timeout:
```
27h48m15s    1001   10010      0 58ms   29ms   21ms   17ms   15ms   9.4ms  9.9ms  
27h48m16s    1035   10350      0 24ms   24ms   21ms   17ms   15ms   8.8ms  9.6ms  
```

soyacz (commented):
according to [docs](https://github.com/scylladb/scylla-bench?tab=readme-ov-file#other-notable-options), when having `-sequential` workflow, then it's not infinite (unless `-iterations` are set to `0`). Maybe performance of scylla is very low?

vponomaryov:
> according to [docs](https://github.com/scylladb/scylla-bench?tab=readme-ov-file#other-notable-options), when having `-sequential` workflow, then it's not infinite (unless `-iterations` are set to `0`). Maybe performance of scylla is very low?

Actually, it doesn't matter whether we have poor perf in Scylla or not from the test setup point of view.
Not having the "duration=" option in stress commands we do not set the defined stress duration in configuration.

Then, majority of commands do have "iterations" param, but it is not obvious at all to understand it's time frame.
But, we have facts that commands do run more than 27 hours when configuration was set to have "24h".

fruch:
I think @roydahan was playing around with this case as well, and I think the outcome was that the duration alone doesn't help.

we and run into https://github.com/scylladb/scylla-bench/issues/138

let's sync with him, to see what is the outcome of him work to refactor this case a bit

roydahan:
You can close this PR.
We will use my PR that simplifies the entire test case.

PR #7446, opened 2024-05-17
fix(azure): fix loader disk size

soyacz:
Recent reduction of loader disk size caused issue with provision as 20GB is lower than image requires.

Fix by not specifying it at all so image decides on the size.

### Testing

- [ ] - provision azure

### Reminders


dimakr (approved):
LGTM

PR #7456, opened 2024-05-20
test(manager): introduce Manager installation test on different OS

mikliapko:
Closes https://github.com/scylladb/scylla-manager/issues/3852

The test is designed for execution on non-main OS distributions (for example, debian10) where the main goal is to execute installation test and verify manager + agent are up and running.

The motivation of having such a test:
- We have never faced OS-specific failures in backup/restore/repair functionality. Thus, there is no need to run the sanity test (which includes these functionality verification) against non-main operation systems;
- Installation test takes only 20 minutes in average against ~2 hours for Manager sanity test;
- The Manager sanity is still here and will be executed on the main OS - Ubuntu 22.

### Testing

- [x] The folder with Manager installation jobs (https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/sct/job/manager-master/job/installation-tests/)

### Reminders

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -1147,3 +1147,33 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
                 f'Task {current_task_status} did not remain in "{TaskStatus.STOPPED}" status, but instead ' \
                 f'reached "{current_task_status}" status'
         self.log.info('finishing test_suspend_and_resume_without_starting_tasks')
+
+    def test_manager_installed_and_functional(self):
+        """Verify that the Manager is installed and functional.
+
+        The test is intended for execution on non-main OS distributions (for example, debian10)
+        where the main goal is to execute installation test.
+        The rest of the checks are executed on Ubuntu distribution in test_manager_sanity.
+        """
+        self.log.info('starting test_manager_installed_and_functional')
+
+        manager_node = self.monitors.nodes[0]
+        scylla_node = self.db_cluster.nodes[0]
+        manager_tool = mgmt.get_scylla_manager_tool(manager_node=manager_node)
+
+        # Check scylla-manager and scylla-manager-agent are up and running (/ping endpoint)
+        manager_node.is_manager_server_up()
+        scylla_node.is_manager_agent_up()
+
+        # Check the sctool version method is callable
+        self.log.info("Got Manager's version: %s", manager_tool.sctool.version)
+
+        # Add cluster and verify hosts health
+        mgr_cluster = manager_tool.add_cluster(name=self.CLUSTER_NAME, db_cluster=self.db_cluster,
+                                               auth_token=self.monitors.mgmt_auth_token)
+        dict_host_health = mgr_cluster.get_hosts_health()
+        for host_health in dict_host_health.values():
+            assert host_health.status == HostStatus.UP, "Host status is not 'UP'"
+            assert host_health.rest_status == HostRestStatus.UP, "Host REST status is not 'UP'"
+
+        self.log.info('finishing test_manager_installed_and_functional')
--- a/jenkins-pipelines/manager/debian10-manager-install.jenkinsfile
+++ b/jenkins-pipelines/manager/debian10-manager-install.jenkinsfile
@@ -1,0 +1,15 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+managerPipeline(
+    backend: 'aws',
+    region: 'us-east-1',
+    test_name: 'mgmt_cli_test.MgmtCliTest.test_manager_installed_and_functional',
+    test_config: '''["test-cases/manager/manager-installation-set-distro.yaml", "configurations/manager/debian10.yaml"]''',
+
+    post_behavior_db_nodes: 'destroy',
+    post_behavior_loader_nodes: 'destroy',
+    post_behavior_monitor_nodes: 'destroy',
+)
--- a/jenkins-pipelines/manager/debian11-manager-install.jenkinsfile
+++ b/jenkins-pipelines/manager/debian11-manager-install.jenkinsfile
@@ -1,0 +1,15 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+managerPipeline(
+    backend: 'aws',
+    region: 'us-east-1',
+    test_name: 'mgmt_cli_test.MgmtCliTest.test_manager_installed_and_functional',
+    test_config: '''["test-cases/manager/manager-installation-set-distro.yaml", "configurations/manager/debian11.yaml"]''',
+
+    post_behavior_db_nodes: 'destroy',
+    post_behavior_loader_nodes: 'destroy',
+    post_behavior_monitor_nodes: 'destroy',
+)
--- a/jenkins-pipelines/manager/ubuntu20-manager-install.jenkinsfile
+++ b/jenkins-pipelines/manager/ubuntu20-manager-install.jenkinsfile
@@ -1,0 +1,15 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+managerPipeline(
+    backend: 'aws',
+    region: 'us-east-1',
+    test_name: 'mgmt_cli_test.MgmtCliTest.test_manager_installed_and_functional',
+    test_config: '''["test-cases/manager/manager-installation-set-distro.yaml", "configurations/manager/ubuntu20.yaml"]''',
+
+    post_behavior_db_nodes: 'destroy',
+    post_behavior_loader_nodes: 'destroy',
+    post_behavior_monitor_nodes: 'destroy',
+)
--- a/jenkins-pipelines/manager/ubuntu22-manager-install.jenkinsfile
+++ b/jenkins-pipelines/manager/ubuntu22-manager-install.jenkinsfile
@@ -1,0 +1,15 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+managerPipeline(
+    backend: 'aws',
+    region: 'us-east-1',
+    test_name: 'mgmt_cli_test.MgmtCliTest.test_manager_installed_and_functional',
+    test_config: 'test-cases/manager/manager-installation-set-distro.yaml',
+
+    post_behavior_db_nodes: 'destroy',
+    post_behavior_loader_nodes: 'destroy',
+    post_behavior_monitor_nodes: 'destroy',
+)
--- a/test-cases/manager/manager-installation-set-distro.yaml
+++ b/test-cases/manager/manager-installation-set-distro.yaml
@@ -1,0 +1,9 @@
+test_duration: 60
+
+instance_type_db: 'i4i.large'
+
+n_db_nodes: 1
+n_loaders: 0
+n_monitor_nodes: 1
+
+user_prefix: manager-installation


mgmt_cli_test.py
```diff
@@ -1147,3 +1147,24 @@ def test_suspend_and_resume_without_starting_tasks(self):
...
+        The rest of the checks are executed on Ubuntu distribution in test_manager_sanity.
+        """
+        self.log.info('starting test_manager_installed_and_functional')
+
+        manager_node = self.monitors.nodes[0]
+        scylla_node = self.db_cluster.nodes[0]
+
+        # Check scylla-manager and scylla-manager-agent are up and running (/ping endpoint)
+        manager_node.is_manager_server_up()
+        scylla_node.is_manager_agent_up()
```
soyacz, line 1165: shouldn't the test also verify communication between agent and server? 
Or at least cluster is registered in server?
I believe several short functional checks could find some regressions without sacrificing a lot of time.

mikliapko: yes, makes sense, added several checks on cluster addition and health check.

soyacz (approved):
LGTM

karol-kokoszka (approved):

mgmt_cli_test.py
soyacz: please retry test builds if everything passes

mikliapko: Yes, it works, verified here:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/sct/job/manager-master/job/installation-tests/job/ubuntu22-installation/4/

mgmt_cli_test.py
```diff
@@ -1147,3 +1147,33 @@ def test_suspend_and_resume_without_starting_tasks(self):
...
+        manager_node = self.monitors.nodes[0]
+        scylla_node = self.db_cluster.nodes[0]
+        manager_tool = mgmt.get_scylla_manager_tool(manager_node=manager_node)
+
+        # Check scylla-manager and scylla-manager-agent are up and running (/ping endpoint)
+        manager_node.is_manager_server_up()
+        scylla_node.is_manager_agent_up()
+
+        # Check the sctool version method is callable
+        _ = manager_tool.sctool.version
```
ilya-rarov, line 1169: I wonder if it is better to put sctool version into logger (something like that)
```python
self.log.info("Got sctool version: %s", manager_tool.sctool.version)
```
or into `assert`
```python
assert manager_tool.sctool.version
```
instead of this assignment.

roydahan:
I don’t understand why we need this PR.
Isn’t the sanity per OS do the exact same thing?

@ShlomiBalalis ?

roydahan:
How much longer?
I don’t think we never faced such an issue, rarly we did.
But anyway, if sanity Is not that long and anyway one of the comments above suggests that you actually verify that manager works, this PR doesn’t worth it.

if sanity became too long over the years, we can consider shortening it.

fruch:
> How much longer?
> I don’t think we never faced such an issue, rarly we did.
> But anyway, if sanity Is not that long and anyway one of the comments above suggests that you actually verify that manager works, this PR doesn’t worth it.
> 
> if sanity became too long over the years, we can consider shortening it.
> 

It's ~2h for each sanity, and those are using spots

From my POV, this is more then fine to minimize to multiple disto cases to mostly installation

roydahan:
From my POV it’s an unneeded duplication.
As I wrote above, we can/should shorten the sanity to what it has been originally, a quick sanity test.

fruch:
> From my POV it’s an unneeded duplication. As I wrote above, we can/should shorten the sanity to what it has been originally, a quick sanity test.

yet again, I'm all about let whom owns it to take the calls of how to structure it,
I don't see any duplication in this, it's the other way around, it replacing some of the jobs with shorter versions.

mikliapko:
> From my POV it’s an unneeded duplication. As I wrote above, we can/should shorten the sanity to what it has been originally, a quick sanity test.

Yes, it's not a duplication because the goal is to replace sanity jobs for debian10, debian11, ubuntu20 distributions with these installation jobs and use the sanity job only for ubuntu22.

One more thing, these short installation jobs can be very good when we need to verify any changes in SCT setup steps or images, etc. We already had this case here (https://github.com/scylladb/scylla-cluster-tests/pull/7459) and instead of running ~1.5-2h jobs for every distro we've run these installation jobs that gave the result in 20 minutes.

mikliapko:
> But anyway, if sanity Is not that long and anyway one of the comments above suggests that you actually verify that manager works, this PR doesn’t worth it.

Some extra checks were added in test to address this comment.

roydahan:
> > From my POV it’s an unneeded duplication. As I wrote above, we can/should shorten the sanity to what it has been originally, a quick sanity test.
> 
> Yes, it's not a duplication because the goal is to replace sanity jobs for debian10, debian11, ubuntu20 distributions with these installation jobs and use the sanity job only for ubuntu22.
> 
> One more thing, these short installation jobs can be very good when we need to verify any changes in SCT setup steps or images, etc. We already had this case here (#7459) and instead of running ~1.5-2h jobs for every distro we've run these installation jobs that gave the result in 20 minutes.

In that case why do we need it also for ununtu22 in this PR?
And what are the triggers changes?

roydahan:
> > But anyway, if sanity Is not that long and anyway one of the comments above suggests that you actually verify that manager works, this PR doesn’t worth it.
> 
> Some extra checks were added in test to address this comment.

And now it looks like the Sanity when it was started.
In the past we had an issue on one platform that manager repair failed with too “many open files”.
With this change we will never hit it, because we stop testing repair when manager is Debian for example.
And if we continue using the sanity, half of it is duplicated with this PR.

mikliapko:
> In that case why do we need it also for ununtu22 in this PR?

From my point of view, we can also benefit from having such a job if we need to quickly verify ubuntu22 installation process for some particular cases (changes in setup procedure).

> And what are the triggers changes?

It's not implemented yet, but we are planning to implement the triggers for them.

mikliapko:
> And now it looks like the Sanity when it was started. 

But it takes 15 minutes instead of 1.5-2 hours.

> In the past we had an issue on one platform that manager repair failed with too “many open files”. With this change we will never hit it, because we stop testing repair when manager is Debian for example. And if we continue using the sanity, half of it is duplicated with this PR.

I haven't known about such issue. In such case we may consider to leave, for example, one per release run of all sanity tests on all platforms (manager-3.2 release branch) and run installation tests only in master.

When I joined to the project, all the sanity tests except Ubuntu22 and CentOS7 (which is deprecated already) have been disabled in Jenkins master branch for more than 1 year. So, nobody really cared much about these runs in master all that time.

roydahan:
> > And now it looks like the Sanity when it was started.
> 
> But it takes 15 minutes instead of 1.5-2 hours.
> 
> > In the past we had an issue on one platform that manager repair failed with too “many open files”. With this change we will never hit it, because we stop testing repair when manager is Debian for example. And if we continue using the sanity, half of it is duplicated with this PR.
> 
> I haven't known about such issue. In such case we may consider to leave, for example, one per release run of all sanity tests on all platforms (manager-3.2 release branch) and run installation tests only in master.
> 
> When I joined to the project, all the sanity tests except Ubuntu22 and CentOS7 (which is deprecated already) have been disabled in Jenkins master branch for more than 1 year. So, nobody really cared much about these runs in master all that time.

Because the developers probably never looked at master so it was worthless to run on master.
Once you run it only for releases, it means that if there is an issue you fix it and it’s a release blocker.
Hence you don’t really care if it takes 15mins or 2 hours because it runs only when you have releases.

IMO, if you want to have it in master, it should be tracked by developers, and if you want it to be tracked by developers, it should be part of their CI.
And if it’s part of their CI, the place for it is probably GitHub actions like the current CI they have.

mikliapko:
> Hence you don’t really care if it takes 15mins or 2 hours because it runs only when you have releases.

From my perspective, I find it more efficient to not run 2 hours job if the same result can be got in 15 minutes job. It's not only the question of time but the cloud resources as well.

> IMO, if you want to have it in master, it should be tracked by developers, and if you want it to be tracked by developers, it should be part of their CI. And if it’s part of their CI, the place for it is probably GitHub actions like the current CI they have.

I don't really think they run something from SCT in Actions.

mikliapko:
@rayakurl probably you may also add something to this discussion?

roydahan:
> > Hence you don’t really care if it takes 15mins or 2 hours because it runs only when you have releases.
> 
> From my perspective, I find it more efficient to not run 2 hours job if the same result can be got in 15 minutes job. It's not only the question of time but the cloud resources as well.
> 
> > IMO, if you want to have it in master, it should be tracked by developers, and if you want it to be tracked by developers, it should be part of their CI. And if it’s part of their CI, the place for it is probably GitHub actions like the current CI they have.
> 
> I don't really think they run something from SCT in Actions.

If it's only installation they don't need to use SCT.
Let's go back to the basics, who requested to test installations on master, how frequently and why?

mikliapko:
> If it's only installation they don't need to use SCT. Let's go back to the basics, who requested to test installations on master, how frequently and why?

So, the idea comes from one of the meeting with the Manager team where we discussed CI approach.
The team confirmed (@karol-kokoszka @Michal-Leszczynski @tzach) that nobody faced any OS specific issues in Manager before and it makes sense to replace the sanity tests for secondary distributions (debian10, debian11, ubuntu20) with just installations checks to safe time and resources.

roydahan:
> So, the idea comes from one of the meeting with the Manager team where we discussed CI approach.
> The team confirmed (@karol-kokoszka @Michal-Leszczynski @tzach) that nobody faced any OS specific issues in Manager before and it makes sense to replace the sanity tests for secondary distributions (debian10, debian11, ubuntu20) with just installations checks to safe time and resources.

So the best thing to save time and resources is to not run anything in case no is is looking at failures on master.
And the easiest and cheapest way to detect installations issues would be GitHub actions that runs on CI and people will actually look at them.

rayakurl:
Sorry for the late response,  was on sick leave for a few days...

@roydahan The testing strategy for manager is  determined by the owners of Manager, which are @mikliapko @karol-kokoszka, @Michal-Leszczynski and mysled. We have bi weekly meetings with ALL the Stakeholders (Yaniv, Gabi, Tzach, myself) and additional grooming meetings for the new features and tests.
There is no point to re-discuss in this PR all the testing strategy, what is triggered, why and when. We already had a couple of meetings for that with everyine that are owning this project and it's not productive. 

I see that for this PR we have 2 approvals from the manager Stakeholders + 1 from @soyacz who is one of the SCT maintainers
@fruch and myself recently discussed a procedure of reviewing SCT manager PRs and in this PR we followed the exact procedure. 

So if you have any suggestions for the code itself as the other reviewers did- you are more than welcome to write.
Other comments are not for this PR.

roydahan (changes requested):
So, I'm not going to duplicate the tests in SCT for this.
There are already tests for all distributions which do install.
Since I'm also maintainer I'm not approving that.

My suggestion here instead of duplicating tests is to support a new mode for the sanity that will skip some sub-tests.
Then, if you decides to enable back the tests in master you can enable them with the skip mode in the trigger and for releases without the skip mode.

P.S Better to link to the trigger PRs in scylla-pkg if you want a complete review.

roydahan:
And as I wrote above, another good approach would be to use GH action that installs manager and will be part of the team CI.

rayakurl:
@gmizrahi please read @roydahan comments above. I don't know how to proceed.
We are the owners of Manager and happen to use SCT as an infra for some of the manager tests due to historical reasons. I personally talked to Israel and we agreed on a procedure of 2 approvals from within the manager task force + 1 from a SCT maintainer.
We followed this procedure but still can't merge this PR.
@mikliapko created the PR after it was discussed in a meeting and agreed with all the people that are involved in the project, spent time working on it, testing it, addressing comments. This is a waste of our time. 

If the owners of Manager don't have authority to decide which pipelines we run in SCT - SCT is completely useless for us and we might as well create a different infra owned by us. Another solution will be to skip the maintainers review for manager PRs, otherwise we cannot make progress.
@gmizrahi please advise.

gmizrahi:
@roydahan as we now own manager testing and these are manager specific changes:
a. please add @mikliapko @rayakurl and myself as maintainers for this part of the code
b. approve the PR (as two of your team even approved the PR as well + manager is our domain).

cc @mykaul

mykaul:
> * We have never faced OS-specific failures in backup/restore/repair functionality. Thus, there is no need to run the sanity test (which includes these functionality verification) against non-main operation systems;

Isn't this an argument why it's not very important?

> * Installation test takes only 20 minutes in average against ~2 hours for Manager sanity test;

Ouch! 20 minutes? What is taking so loooong? I'm very interested to know. Installation logs might help.

mikliapko:
> Ouch! 20 minutes? What is taking so loooong? I'm very interested to know. Installation logs might help.

It's not the test itself that takes 20 minutes but infra preparation steps.
We hope it should be improved by added yesterday provisioning step to manager pipeline (https://github.com/scylladb/scylla-cluster-tests/pull/7469). Haven't tested these installation tests with this step yet.

mikliapko:
> So, I'm not going to duplicate the tests in SCT for this. There are already tests for all distributions which do install. Since I'm also maintainer I'm not approving that.
> 
> My suggestion here instead of duplicating tests is to support a new mode for the sanity that will skip some sub-tests. Then, if you decides to enable back the tests in master you can enable them with the skip mode in the trigger and for releases without the skip mode.
> 
> P.S Better to link to the trigger PRs in scylla-pkg if you want a complete review.

In current implementation I don't see really many intersections between these two tests:

**Sanity test:**
1. Generate load
2. Do basic backup test
3. Do restore test
4. Do repair
5. Test cluster CRUD operations
6. Health check test
7. Suspend and resume test
8. Client encryption test

**Installation test:**
1. Check scylla-manager and scylla-manager-agent are up and running (/ping endpoint)
2. Check the sctool version method is callable and returns valid info
3. Add cluster and verify hosts health

So, of course, they can be joined but from my point of view it will make sanity test more complex.
Moreover, I don't see a big issue in introducing new test. 
To be honest, the Manager tests are quite far from being laconic and non-containing duplicated code. And I agree that it's not a reason to introduce new duplicates but I personally don't see much duplication here.

soyacz:
as to test speed - collect logs phase takes long because of getting monitoring screenshots mainly.
I assume this is not relevant for these tests and we could provide means to skip it for this cases.

roydahan:
> @roydahan as we now own manager testing and these are manager specific changes: a. please add @mikliapko @rayakurl and myself as maintainers for this part of the code b. approve the PR (as two of your team even approved the PR as well + manager is our domain).

Sure, once each will reach few hundreds significant commits we will consider adding new maintainers.
> 
> cc @mykaul

roydahan:
> > So, I'm not going to duplicate the tests in SCT for this. There are already tests for all distributions which do install. Since I'm also maintainer I'm not approving that.
> > My suggestion here instead of duplicating tests is to support a new mode for the sanity that will skip some sub-tests. Then, if you decides to enable back the tests in master you can enable them with the skip mode in the trigger and for releases without the skip mode.
> > P.S Better to link to the trigger PRs in scylla-pkg if you want a complete review.
> 
> In current implementation I don't see really many intersections between these two tests:
> 
> **Sanity test:**
> 
> 1. Generate load
> 2. Do basic backup test
> 3. Do restore test
> 4. Do repair
> 5. Test cluster CRUD operations
> 6. Health check test
> 7. Suspend and resume test
> 8. Client encryption test
> 
> **Installation test:**
> 
> 1. Check scylla-manager and scylla-manager-agent are up and running (/ping endpoint)
> 2. Check the sctool version method is callable and returns valid info
> 3. Add cluster and verify hosts health
> 
> So, of course, they can be joined but from my point of view it will make sanity test more complex. Moreover, I don't see a big issue in introducing new test. To be honest, the Manager tests are quite far from being laconic and non-containing duplicated code. And I agree that it's not a reason to introduce new duplicates but I personally don't see much duplication here.

@mikliapko again, if it’s only installation of manager, have you considered using simple GH actions that will be part of CI?

mykaul:
> > @roydahan as we now own manager testing and these are manager specific changes: a. please add @mikliapko @rayakurl and myself as maintainers for this part of the code b. approve the PR (as two of your team even approved the PR as well + manager is our domain).
> 
> Sure, once each will reach few hundreds significant commits we will consider adding new maintainers.

Hundreds? 

> 
> > cc @mykaul

mykaul:
> as to test speed - collect logs phase takes long because of getting monitoring screenshots mainly. I assume this is not relevant for these tests and we could provide means to skip it for this cases.

I hope there's an issue on this - why would taking a screenshot take long? (I assume there's a reason, perhaps we can improve that).

mikliapko:
> @mikliapko again, if it’s only installation of manager, have you considered using simple GH actions that will be part of CI?

No, we haven't tried this option. But what for? 
The QA team doesn't look into GH Actions jobs that devs are taking care of. We want to have all the tests interesting and important for us in one place, together with the rest Manager's SCT jobs.

roydahan:
> > as to test speed - collect logs phase takes long because of getting monitoring screenshots mainly. I assume this is not relevant for these tests and we could provide means to skip it for this cases.
> 
> I hope there's an issue on this - why would taking a screenshot take long? (I assume there's a reason, perhaps we can improve that).

Probably since the monitor is very very slow.
It’s a small node that runs manager (with its Scylla) and monitor.

roydahan:
> > @mikliapko again, if it’s only installation of manager, have you considered using simple GH actions that will be part of CI?
> 
> No, we haven't tried this option. But what for? The QA team doesn't look into GH Actions jobs that devs are taking care of. We want to have all the tests interesting and important for us in one place, together with the rest Manager's SCT jobs.

That’s exactly my whole point.
Since we don’t want only QA to watch these jobs and finding AFTER the fact that there was a breaking commit.
That’s why these tests were disabled at first place. Developers didn’t watch them, hence no one watch them so it was a waste to run them.
So, instead the same thing will happen with your new tests, having a simple GH action that only installs it will allow developers to detect these as part of their CI. It can be also set as an extended CI.

rayakurl:
> > > @mikliapko again, if it’s only installation of manager, have you considered using simple GH actions that will be part of CI?
> > 
> > 
> > No, we haven't tried this option. But what for? The QA team doesn't look into GH Actions jobs that devs are taking care of. We want to have all the tests interesting and important for us in one place, together with the rest Manager's SCT jobs.
> 
> That’s exactly my whole point. Since we don’t want only QA to watch these jobs and finding AFTER the fact that there was a breaking commit. That’s why these tests were disabled at first place. Developers didn’t watch them, hence no one watch them so it was a waste to run them. So, instead the same thing will happen with your new tests, having a simple GH action that only installs it will allow developers to detect these as part of their CI. It can be also set as an extended CI.

@roydahan we have a defined testing strategy,  we have tests in GH actions maintained and watched by @karol-kokoszka and @Michal-Leszczynski .
So, We have a few levels of tests and we have a clear plan on how to test manager.
We are defining a procedure on who looks at the runs and have high collaboration between the qa and the developers as we are part of the same group now.

Having said that, as far as I know this is not a concern of a SCT maintainer.
As far as I understand, a maintainer of a repository should check the correctness of the code, verify that the new code is not introducing new regressions etc. 
The manager tests are owned by us and we decide how it's tested,  when and by whom. It's discussed and pre approved by everyone who are involved in the project.
And IMO this PR is not the place to discuss it.

For the scope of this PR, Mikita already showed that the new test doesn't duplicate the old test, and Israel who is a maintainer didn't think so as well, as well as Lukasz, who initially approved the PR. We have approval from 2 maintainers and 2 members of the manager task force, which is more than we need.
So nothing should hold this PR from being merged.

@mykaul, @gmizrahi - if I am wrong, and the SCT maintainers job is to define the test strategy (never witnessed this before),  all the maintainers should attend all the manager meetings that we have (bi-weekly, weekly grooming etc) , be part of the manager channel and to keep up to date with all the changes that we make in real time.
I don't think that this is productive but your call.

roydahan:
It does introduce duplication.
It doesn’t remove the older sanities nor linking to a Scylla-pkg PR that changes the triggers accordingly.

rayakurl:
This PR is pending a few days now.
@roydahan it's waiting for you to reply to Mikita's comment

roydahan:
Waiting to see the triggers PR which is anyway blocking these tests.

mikliapko:
> Waiting to see the triggers PR which is anyway blocking these tests.

Is it a must-have for any tests to be merged?

We are planning to introduce the triggers for these installation tests as a part of this task (https://github.com/scylladb/scylla-manager/issues/3856) where we plan to rework the triggering process itself and for manual builds allow to choose the group of tests to execute depending on their priority. I just don't wanna do the double work adding the triggers for installation tests only if in the near time this triggers would be reworked again.

Moreover, not having such triggers yet doesn't prevent us from manually executing these tests for now.

@roydahan

gmizrahi:
@roydahan please approve the PR or grant me permissions to approve it.

Thanks!

mikliapko:
@roydahan 

While waiting for this PR to be merged, we've implemented the task (https://github.com/scylladb/scylla-manager/issues/3856) where the tests triggering approach has been reworked a bit. Installation tests were also added.

See the PR https://github.com/scylladb/scylla-pkg/pull/4109

roydahan (approved):

gmizrahi:
thank you @roydahan

vponomaryov:
@mergify backport manager-3.2

Issue #7447, opened 2024-05-17
Scylla test with all addresses ipv6 public fails on nodetool status


soyacz:
## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Tried provision test with `configurations/network_config/all_addresses_ipv6_public.yaml` config and failed on `Waiting for nodes to join the cluster`. 
While nodetool status shows all nodes up:
```
< t:2024-05-17 12:43:13,872 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > Datacenter: eu-west
< t:2024-05-17 12:43:13,873 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > ===================
< t:2024-05-17 12:43:13,873 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > Status=Up/Down
< t:2024-05-17 12:43:13,873 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > |/ State=Normal/Leaving/Joining/Moving
< t:2024-05-17 12:43:13,873 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > -- Address                                 Load    Tokens Owns Host ID                              Rack
< t:2024-05-17 12:43:13,873 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > UN 2a05:d018:12e3:f000:2415:944:1500:6022  1.61 MB 256    ?    a35d9f4b-220a-43d7-8317-bcf2a1e999b3 1a  
< t:2024-05-17 12:43:13,873 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > UN 2a05:d018:12e3:f000:4a6e:f30a:cbf0:f73  1.85 MB 256    ?    dd8fd382-83c8-4e31-a9a7-aaaa3bb98d11 1a  
< t:2024-05-17 12:43:13,873 f:base.py         l:230  c:RemoteLibSSH2CmdRunner p:DEBUG > UN 2a05:d018:12e3:f000:4eca:abac:68a4:7d4d 1.67 MB 256    ?    ae2d7ff7-85a6-4b90-9bb4-d39ef1a0d260 1a  
[...]

PR #7047, opened 2024-01-01
fix(get_nodetool_status): convert ipv6 to full format

fruch:
recent change to some java libraries used by nodetool is causing ipv6 to be presented as rfc5952 section-4 without leading zeros, and SCT was comapring those as string with other places we get addresses from.

now we convert it to the full form, after getting
it from nodetool, and still keep it as string

Ref: https://www.rfc-editor.org/rfc/rfc5952#section-4

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-10gb-3h-ipv6-test/5/

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevent to this change (if needed)

PR #7460, opened 2024-05-21
fix(cluster): ipv6 address in network_interfaces object

juliayakovlev:
The problem was found during IPv6 test.
IPv6 was converted to full format when parse nodetool status result. This change was represented by  https://github.com/scylladb/scylla-cluster-tests/pull/7047

But IPv6 address was not converted when collected info about nodes network interfaces (AWS).

During checking nodes status, we compare between 'nodetool status' output and node address that kept in the network_interfaces object. It fails because the IPv6 address miss leading zeros. As result the test fails with errors 'Failed to find a node in cluster by IP'

Also fix the same problem for peer and gossip output.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/7447

### Testing

- [x] [longevity-10gb-3h-ipv6-test](https://argus.scylladb.com/test/f4a292e9-e796-4467-91c2-2c2f9a77c96b/runs?additionalRuns[]=b05b2283-2dd3-4542-b631-e682064a1ef5) - IPv6 test
- [x] [longevity-100gb-4h](https://argus.scylladb.com/test/a2b4b1af-a5c3-43a6-965f-c0ae9500370c/runs?additionalRuns[]=73db56bd-e3a4-49e7-86d1-a11da908b917) - IPv4 test

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2744,7 +2744,8 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         for row in cql_results:
             peer = row.peer
             try:
-                ipaddress.ip_address(row.peer)
+                # make sure we use ipv6 long format (some tools remove leading zeros)
+                peer = ipaddress.ip_address(row.peer).exploded
             except ValueError as exc:
                 current_err = f"Peer '{peer}' is not an IP address, err: {exc}\n"
                 LOGGER.warning(current_err)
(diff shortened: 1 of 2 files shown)


sdcm/cluster.py
```diff
@@ -2776,7 +2776,7 @@ def get_gossip_info(self) -> dict[BaseNode, dict]:
             if line.startswith('SCHEMA:'):
                 schema = line.replace('SCHEMA:', '')
             elif line.startswith('RPC_ADDRESS:'):
-                ip = line.replace('RPC_ADDRESS:', '')
+                ip = ipaddress.ip_address(line.replace('RPC_ADDRESS:', '')).exploded
```
fruch, line 2779: I wasn't considering the outout of gossipinfo... 👍🏼

fruch (approved):
LGTM

sdcm/cluster.py
juliayakovlev: I am validating that it won't broke IPv4 test

fruch (approved):
LGTM

Issue #7242, opened 2024-03-04
Optimize test setup time


roydahan:
Test setup include several steps like provisioning instances, setting up db nodes, loaders, monitor.
Some steps are doing installations that may not be necessary or could be replaced with an image (e.g. monitor image), some may be done in parallel and currently being done serially because we need an input from a previous step, etc.

This task should start by mapping run times per step for a typical longevity setup, per step per node.
Mentioning which step is parallel to another and which is serial and which happens in parallel to all nodes and which is serial.

Then, we should start creating the sub-tasks to optimize each step.

mykaul:
https://github.com/scylladb/scylla-cluster-tests/issues/7025 is one such example.

soyacz:
referencing prepare monitor image PR: https://github.com/scylladb/scylla-cluster-tests/pull/7305
It has great breakdown on the timings

soyacz:
Please find my analysis of test setup timings ( numbers in nested trees are parts of root duration):
- SCT starts (first log line) - out of scope (possibly we can get some time out of previous steps in jenkins)
- TEST_START event (`message=TEST_START`) - 15s
    - `SCTConfiguration()` - 5s
    - events devices setup - 6s
- update certificates - <1s
- init_resources - 297s (hardware)
    - getting Security groups and subnets - 1s
    - init 6 db nodes (`Init nodes`) - 200s, sequentially:
        - detect distro, set hostname etc. 2s
        - 29s for syslogng/ssh setup (mostly syslogng) per node
    - init 2 loader nodes  (`Init nodes`)  - 47 sec
        - similar, but a bit faster than db nodes
    - 1 monitor node  (`Init nodes`)  - 45 sec
- Init db cluster (`Class instance: Cluster`) - 1207s
    - init wrap: 1167 (`TestConfig duration`) (parallel node setup)
    - update scylla.yaml - set seeds and seed provider. Sequentially. - 23s
    - check node health - 6s
    - set auth rf: 12s
- init loaders - 333s
    - init wrap: 332s  (`TestConfig duration`)
- init monitor nodes: 892s
    - init wrap: 891s  (`TestConfig duration`)
- Argus collect packages - 22s
- argus get scylla version: 1s
- validate seeds, raft: 20s

Setup alltogether ~46m30sec (2790s)

soyacz:
My proposal:
[...]

PR #7464, opened 2024-05-22
improvement(apt): use builtin feature for lock release

soyacz:
Since February 2020 apt gives an builtin way to wait for apt lock release.

Use it to speed up packages installation (as waiting for lsof result takes 0.5s)

refs: https://github.com/scylladb/scylla-cluster-tests/issues/7242

### Testing

- [ ]

### Reminders


fruch (dismissed):
LGTM,

but we still need to clear the 2 tests failures (the longevity and upgrade test), form quick look doesn't look related, but still need to validate

soyacz:
> LGTM,
> 
> but we still need to clear the 2 tests failures (the longevity and upgrade test), form quick look doesn't look related, but still need to validate

yes, unrelated: 
- https://github.com/scylladb/scylla-cluster-tests/issues/7480
- upgrade failed due jenkins restart 

I rerun job to verify this PR again.

sdcm/sct_config.py
```diff
@@ -2428,7 +2428,7 @@ def get_version_based_on_conf(self):  # pylint: disable=too-many-locals
             amis = self.get('ami_id_db_scylla').split()
             region_name = self.region_names[0]
             tags = get_ami_tags(ami_id=amis[0], region_name=region_name)
-            scylla_version = tags.get('scylla_version')
+            scylla_version = tags.get('scylla_version') or tags.get('ScyllaVersion')
```
vponomaryov, line 2431: How this change is related to the goal of the commit?

soyacz: I just fixed it while testing on some older Scylla version where we used different tags.

fruch: it would be nice if it was in it's own separate commit, but it's cosmetic from my POV

fruch:
@soyacz what about apply this in cloud-init parts as well ? we don't have there logic for waiting for apt to be available ?

there an equivalent in yum ?

sdcm/sct_config.py
soyacz: fixed

soyacz:
> @soyacz what about apply this in cloud-init parts as well ? we don't have there logic for waiting for apt to be available ?

Added to cloud-init script

> 
> there an equivalent in yum ?

From what I saw, yum has it by default, showing warnings when waiting for lock.

soyacz:
trying once again longevity test, just in case of error after introducing it to cloud-init: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-ScyllaKillMonkey-aws-test/14/

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #7466, opened 2024-05-22
fix(offline ubuntu24.04) skip python2 installation

yaronkaikov:
failed in https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/artifacts-offline-install/job/artifacts-ubuntu2404-test/3 with the following error:
```
11:07:40  Command: 'sudo DEBIAN_FRONTEND=noninteractive apt-get -o Dpkg::Options::="--force-confold" -o Dpkg::Options::="--force-confdef" install -y python2'
```

Since in ubuntu24.04 we don't have python2 package we should skip it

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2005,7 +2005,7 @@ class BaseNode(AutoSshContainerMixin, WebDriverContainerMixin):  # pylint: disab
         if not nonroot:
             self.install_package(package_name='xfsprogs mdadm')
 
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

fruch:
it's a bit of a left over, for releases branches we keep in sync.
but this would work

yaronkaikov:
Verified with https://jenkins.scylladb.com/job/scylla-master/job/releng-testing/job/artifacts-offline-install/job/artifacts-ubuntu2404-test/4/

PR #7479, opened 2024-05-23
fix(test_base_version): stop hardcoding the current latest branch

fruch:
since we run into this again and again, on every new release branch created, we shouldn't pin this test to specific one checking it's greater then current known version it enough

### Testing

- [ ]

### Reminders

### DIFF
--- a/unit_tests/test_base_version.py
+++ b/unit_tests/test_base_version.py
@@ -14,6 +14,8 @@
 
 import unittest
 
+from sdcm.utils.version_utils import ComparableScyllaVersion
+
 from utils.get_supported_scylla_base_versions import UpgradeBaseVersion  # pylint: disable=no-name-in-module, import-error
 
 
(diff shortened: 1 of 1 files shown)


vponomaryov (dismissed):
LGTM

soyacz:
ugh, I didn't notice this PR before adding mine... now this conflicted.

unit_tests/test_base_version.py
```diff
@@ -34,7 +36,8 @@ def test_master(self):
         scylla_repo = self.url_base + '/master/rpm/centos/2021-08-29T00:58:58Z/scylla.repo'
         linux_distro = 'centos'
         version_list = general_test(scylla_repo, linux_distro)
-        self.assertEqual(version_list, ['6.0'])
+        assert len(version_list) == 1
+        assert ComparableScyllaVersion(version_list[0]) > '5.4'
```
vponomaryov, line 40: Need to limit it to be below the possible enterprise versions:

```
'5.4' < ComparableScyllaVersion(version_list[0]) < '2018.1'
```

vponomaryov (changes requested):

unit_tests/test_base_version.py
fruch: Not that critical from POV

The whole idea is to break it away from changes in branches

soyacz (approved):
LGTM

unit_tests/test_base_version.py
vponomaryov: It is very simple and useful change where code example is already provided.
It covers possible bug where a scylla version may be returned from different scylla type.

fruch:
@mergify backport manager-3.2

PR #7484, opened 2024-05-26
feature(jenkins): New jobs with gossip-topology-changes

aleksbykov:
Raft topology changes feature is enabled by default and always run for new and upgraded cluster starting from version 6.0

to run cluster with gossip topology operations, parameter force-gossip-topology-changes: true should be added to scylla.yaml

Created 3 simple jobs as example and for test purposes

### Testing

- [ ]

### Reminders

### DIFF
--- a/configurations/raft/consistent_topology_changes_off.yaml
+++ b/configurations/raft/consistent_topology_changes_off.yaml
@@ -1,0 +1,2 @@
+append_scylla_yaml: |
+  force-gossip-topology-changes: true
--- a/jenkins-pipelines/oss/gossip-topology-changes/longevity-100gb-4h-gossip-topology.jenkinsfile
+++ b/jenkins-pipelines/oss/gossip-topology-changes/longevity-100gb-4h-gossip-topology.jenkinsfile
@@ -1,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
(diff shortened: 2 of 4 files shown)


fruch:
@aleksbykov 

is there some place there was a discussion / test plan, which jobs should be duplicated ? 
(otherwise, what exactly we should be reviewing here ?)

fruch:
@aleksbykov @temichus 

I never got an answer regarding this PR, is that needed ? it's expected to be part of teir1 ? part of a release testing ?
is that needed for 2024.2 ?

aleksbykov:
> @aleksbykov @temichus
> 
> I never got an answer regarding this PR, is that needed ? it's expected to be part of teir1 ? part of a release testing ? is that needed for 2024.2 ?

@fruch , Raft enabled by default for enterprise 2024.2 and 6.x versions. This job could be useful, if we need to recheck scylla behavior with old functionality (gossip) while the gossip will not be totally removed from scylla code base.
the flag `force-gossip-topology-changes` was introduced for testing purposes only.
I don't think this jobs should be part of tier1 or release testing, but they could be added as example of functionality without raft topology and tablets

roydahan:
@aleksbykov , @temichus if we need to test this variation of configuration for long term, then this configuration should be part of plans and tests of future releases.

fruch:
@temichus @aleksbykov, if those are not needed for upcoming releases, please close this PR.

PR #7508, opened 2024-05-29
fix(gce-logcollector): use private addresses as default

fruch:
since now we are working as default with private addresses in
gcp and the public addresses can't be used, we now need to use
private addresses for the log collecting

similar fix as in  930c31d926c9d45b444687e1db3e1378890ee6a7 for aws

### Testing

- [x] provision tests showed great improvement in speed for gce (from ~30m to ~8m)

### Reminders


fruch:
@soyacz seems like a fallout from closing GCE networking, that we didn't noticed for quite some time

vponomaryov (approved):
LGTM

juliayakovlev (approved):

PR #7295, opened 2024-03-25
fix(upgrade): upgrade with raft topology procedure

aleksbykov:
After upgrade to latest master(6.0) raft topology feature or
tablets + raft topology features will be enabled by default
To switch cluster from gossiper to raft topology, raft topology
procedure should be executed. It is described here:
https://github.com/scylladb/scylladb/blob/c5601a749e21fc710958a7c84316ecdf5943022c/docs/dev/topology-over-raft.md
section: Upgrade from legacy topology to raft-based topology

parameter 'enable_tablets_on_upgrade` is used to control upgrade with/without tablets. By default tablets are enabled 
with scylla parameter `enable_tablets` and it is added to scylla.yaml starting from 6.0. Upon upgrade this parameter should be set explicitly to true to enable tablets or set to false or not be added to scylla.yaml so tablets stay disabled after upgrade

raft topology is enabled by default, and could be disabled only with parameter `enforce_gossip_topology_changes`. if this parameter is added to scylla.yaml before upgrade, then raft feature: consistent topology changes will not be enabled

### Testing
run upgrade from 5.4 -> 6.0.
- [with enable tablets on upgrade](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/rolling-upgrade-staging-abykov/job/rolling-upgrade-ami-test/17)
- [with enable tablets on upgrade](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/rolling-upgrade-staging-abykov/job/rolling-upgrade-ami-test/16)
- [with disable tablets on upgrade](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/rolling-upgrade-staging-abykov/job/rolling-upgrade-ami-test/15)

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevent to this change (if needed)

PR #7512, opened 2024-05-30
fix(tablets): remove tablets from experimental features

yarongilor:
fix(tablets): remove tablets and consistent-topology-changes from experimental features
    
            tablets feature is not experimental anymore and is default
            to be enabled. tablets is dependent on consistent-topology-changes.
            refs: https://github.com/scylladb/scylladb/pull/18898
refs: https://github.com/scylladb/scylladb/pull/18898

i'm not sure this PR is needed and about its dependencies or anything else.. 
The PR also assumes there's no need of updating scylla yaml with: `"enable_tablets": "false"`
### Testing

- [ ]

### Reminders


defaults/test_default.yaml
```diff
@@ -136,7 +136,6 @@ scylla_version: ''
 test_upgrade_from_installed_3_1_0: false
 target_upgrade_version: ''
 disable_raft: true
```
bhalevy, line 138: Is this still true?  Do we disable raft by default?

upgrade_test.py
```diff
@@ -371,15 +368,6 @@ def _rollback_node(self, node, upgrade_sstables=True):
         node.run_nodetool("snapshot")
         node.stop_scylla_server(verify_down=False)
 
-        if self.params.get("enable_tablets_on_upgrade"):
-            with node.remote_scylla_yaml() as scylla_yml:
-                current_experimental_features = scylla_yml.experimental_features
-                current_experimental_features.remove("tablets")
```
bhalevy, line 377: Should we instead edit scylla.yaml to remove the newly introduced `enable_tablets` config option?
(See scylladb/scylladb#18898)

defaults/test_default.yaml
yarongilor: IIUC this flag is not relevant anymore, unless still relevant to only 1 out of 2 raft features. @aleksbykov , @roydahan , can you please confirm?

roydahan: Probably not.
@aleksbykov / @temichus ?

roydahan:
@yarongilor your PR shouldn't touch raft experimntal - @aleksbykov / @temichus please organise it so whatever should be tested with "raft disabled" (consistent-topology) is correctly configured and remove all the unnecessary flags / configurations.

@yarongilor your PR should flip the flag, now to support "disabling" tablets and have the duplicated jobs that are testing tablets with "disabled raft" - @aleksbykov & @temichus should be synced because some of these should be for 6.0 (so maybe they already working on it).

yarongilor:
> @yarongilor your PR shouldn't touch raft experimntal - @aleksbykov / @temichus please organise it so whatever should be tested with "raft disabled" (consistent-topology) is correctly configured and remove all the unnecessary flags / configurations.
> 
> @yarongilor your PR should flip the flag, now to support "disabling" tablets and have the duplicated jobs that are testing tablets with "disabled raft" - @aleksbykov & @temichus should be synced because some of these should be for 6.0 (so maybe they already working on it).

Conclusions:
This PR will only change tablets flag and will modify a directory to have tablets disabled. 
The upgrade will also be adjusted to add the tablets flag.

defaults/test_default.yaml
aleksbykov: This flag is only was used for rolling upgrade tests. Job with disabled raft uses additional yaml with relevant

yarongilor:
Addressed in https://github.com/scylladb/scylla-cluster-tests/pull/7295 and https://github.com/scylladb/scylla-cluster-tests/pull/7655

PR #7515, opened 2024-05-30
fix(configure_remote_logging): wait for cloud-init

fruch:
since one cases we run without provision phase, we might get into `configure_remote_logging`, before cloud-init has finished to installations of packages, and file we use to indicate no need to configure syslog-ng again isn't yet create, and we can run into case we might try to run `apt-get update` at the same time, and fail with what would look like premission issues

this change also remove the hack of disabling ssh at the start and at the end of cloud-init script, since we wait for cloud-init before doing any extra configuration.

### Testing

- [x] tested local the aws provision test
- [ ] all of the provision tests

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -146,6 +146,7 @@ from sdcm.paths import (
     SCYLLA_MANAGER_TLS_KEY_FILE,
 )
 from sdcm.sct_provision.aws.user_data import ScyllaUserDataBuilder
+
 from sdcm.exceptions import (
     KillNemesis,
     NodeNotReady,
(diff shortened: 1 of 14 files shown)


fruch:
Seems like docker/k8s needs to skip this check, since the docker image doesn't have it ..

sdcm/cluster.py
```diff
@@ -363,6 +365,7 @@ def init(self) -> None:
         # Start task threads after ssh is up, otherwise the dense ssh attempts from task
         # threads will make SCT builder to be blocked by sshguard of gce instance.
         self.wait_ssh_up(verbose=True)
+        self.wait_for_cloud_init()
```
soyacz, line 368: I wouldn't add it to base class `init` - Azure does it differently (so this will be duplication) and as you mentioned, docker/baremetal and possibly others might have problem with that.
I propose to add it to `AWSNode` and `GCENode` `init` methods only

vponomaryov: Agree. It is not applicable to docker/k8s.

So, need to make a `raise NotImplementedError()` implementation by default and then redefine it per backend.

vponomaryov (changes requested):
Implementation must be backend-specific.

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

PR #7525, opened 2024-06-02
fix(rolling-upgrade): Set of commits to fix rolling upgrade from 6.0-6.1

aleksbykov:
Set of commits which are fix several issues in rolling upgrades from 6.0 to 6.1

[fix(restore-backup): Restore only existent files](https://github.com/scylladb/scylla-cluster-tests/pull/7525/commits/0fec37f69d339ace09f30355a8917c3029b9c7aa) - Fixing script error when try to restore backup config files which could be missing

[fix(fill_db-UP): Temprorary disable cdc for rolling upgrade](https://github.com/scylladb/scylla-cluster-tests/pull/7525/commits/7cae54dc8d86b12abaf16b6a063a13c9bb8f9e51) - Disable creating table with cdc options because it is not supported with tablets

[fix(upgrade_test): Wait while all nodes up after rollback](https://github.com/scylladb/scylla-cluster-tests/pull/7525/commits/4a923f8705b2e9a2e66a1695dc3982c8b2a0d569) - need to wait after node upgrade/rollback that all nodes are up and normal. it is required by raft topology feature

[fix(get_highest_sstable_version): get enabled sstable version from db](https://github.com/scylladb/scylla-cluster-tests/pull/7525/commits/99d7eebd37bbf467ca86bbc5b56b00590418521a) - Log message with enabled format sstable is dissappeared from log file. Added new methods to get supported sstable format from system_local

[fix(upgrade): no assert on upggrade sstable](https://github.com/scylladb/scylla-cluster-tests/pull/7525/commits/b07f1b9219426e0cf159d87fdde8c1f5ecacba16) - change assert to log error for current sstable format.

Latest to commits could be https://github.com/scylladb/scylladb/issues/18995

### Testing
- [Job with fixes](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/rolling-upgrade-staging-abykov/job/rolling-upgrade-ubuntu22.04-test/8)

### Reminders

### DIFF
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -23,6 +23,7 @@ import time
 import re
 
 from collections import OrderedDict
+from functools import cached_property
 from uuid import UUID
 
 from cassandra import InvalidRequest
@@ -32,9 +33,11 @@ from cassandra.protocol import ProtocolException  # pylint: disable=no-name-in-m
 
 from sdcm.tester import ClusterTester
 from sdcm.utils.database_query_utils import fetch_all_rows
-from sdcm.utils.decorators import retrying
+from sdcm.utils.decorators import retrying, optional_cached_property
 from sdcm.utils.cdc.options import CDC_LOGTABLE_SUFFIX
 from sdcm.utils.version_utils import ComparableScyllaVersion
+from sdcm.utils.features import is_tablets_feature_enabled
+from sdcm.utils.issues import SkipPerIssues
 
 
 LOGGER = logging.getLogger(__name__)
@@ -51,8 +54,6 @@ class FillDatabaseData(ClusterTester):
     NULL_VALUES_SUPPORT_ENTERPRISE_MIN_VERSION = "2021.1.3"
     NEW_SORTING_ORDER_WITH_SECONDARY_INDEXES_OS_MIN_VERSION = "4.4.rc0"
     NEW_SORTING_ORDER_WITH_SECONDARY_INDEXES_ENTERPRISE_MIN_VERSION = "2021.1.dev"
-    CDC_SUPPORT_MIN_VERSION = "4.3"
-    CDC_SUPPORT_MIN_ENTERPRISE_VERSION = "2021.1.dev"
 
     base_ks = "keyspace_fill_db_data"
     # List of dictionaries for all items tables and their data
@@ -3030,11 +3031,13 @@ class FillDatabaseData(ClusterTester):
             return match.groupdict()["table_name"]
         return None
 
-    @staticmethod
-    def cql_create_simple_tables(session, rows):
+    def cql_create_simple_tables(self, session, rows):
         """ Create tables for truncate test """
-        create_query = "CREATE TABLE IF NOT EXISTS truncate_table%d (my_id int PRIMARY KEY, col1 int, value int) " \
-            "with cdc = {'enabled': true, 'ttl': 0}"
+        create_query = "CREATE TABLE IF NOT EXISTS truncate_table%d (my_id int PRIMARY KEY, col1 int, value int) "
+        # Cannot create CDC log for a table,
+        # because keyspace uses tablets.See issue  scylladb/scylladb#16317.
+        if self.enable_cdc_for_tables:
+            create_query += "with cdc = {'enabled': true, 'ttl': 0}"
         for i in range(rows):
             session.execute(create_query % i)
             # Added sleep after each created table
@@ -3117,7 +3120,9 @@ class FillDatabaseData(ClusterTester):
                 self.all_verification_items[test_num]['skip_condition'] = True
                 with self._execute_and_log(f'Created tables for test "{test_name}" in {{}} seconds'):
                     for create_table in item['create_tables']:
-                        if self.version_cdc_support():
+                        # Cannot create CDC log for a table keyspace_fill_db_data.order_by_with_in_test,
+                        # because keyspace uses tablets. See issue scylladb/scylladb#16317.
+                        if self.enable_cdc_for_tables:
                             create_table = self._enable_cdc(item, create_table)  # noqa: PLW2901
                         # wait a while before creating index, there is a delay of create table for
                         # waiting the schema agreement
@@ -3172,12 +3177,17 @@ class FillDatabaseData(ClusterTester):
             version_with_support = self.NON_FROZEN_SUPPORT_OS_MIN_VERSION
         return self.parsed_scylla_version >= version_with_support
 
-    def version_cdc_support(self):
-        if self.is_enterprise:
-            version_with_support = self.CDC_SUPPORT_MIN_ENTERPRISE_VERSION
-        else:
-            version_with_support = self.CDC_SUPPORT_MIN_VERSION
-        return self.parsed_scylla_version >= version_with_support
+    @cached_property
+    def enable_cdc_for_tables(self) -> bool:
+        if self.tablets_enabled and SkipPerIssues(issues="https://github.com/scylladb/scylladb/issues/16317", params=self.params):
+            return False
+        return True
+
+    @cached_property
+    def tablets_enabled(self) -> bool:
+        """Check is tablets enabled on cluster"""
+        with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
+            return is_tablets_feature_enabled(session)
 
     @retrying(n=3, sleep_time=20, allowed_exceptions=ProtocolException)
     def truncate_table(self, session, truncate):  # pylint: disable=no-self-use
--- a/test-cases/upgrades/rolling-upgrade.yaml
+++ b/test-cases/upgrades/rolling-upgrade.yaml
@@ -35,7 +35,8 @@ gemini_cmd: "gemini -d --duration 2h \
 --max-mutation-retries 5 --max-mutation-retries-backoff 500ms \
 --async-objects-stabilization-attempts 5 --async-objects-stabilization-backoff 500ms \
 --replication-strategy \"{'class': 'NetworkTopologyStrategy', 'replication_factor': '3'}\" \
---table-options \"cdc={'enabled': true}\" --test-username cassandra --test-password cassandra"
+--test-username cassandra --test-password cassandra"
+
 gemini_seed: 66
 
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -356,6 +356,8 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
             self.upgradesstables_if_command_available(node)
             InfoEvent(message='upgrade_node - ended to "upgradesstables_if_command_available"').publish()
 
+        self.db_cluster.wait_all_nodes_un()
+
     @truncate_entries
     # https://github.com/scylladb/scylla/issues/10447#issuecomment-1194155163
     def rollback_node(self, node, upgrade_sstables=True):
@@ -456,6 +458,8 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
 
         if upgrade_sstables:
             self.upgradesstables_if_command_available(node)
+
+        self.db_cluster.wait_all_nodes_un()
 
     @staticmethod
     def upgradesstables_if_command_available(node, queue=None):  # pylint: disable=invalid-name
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -41,6 +41,7 @@ from sdcm.utils.version_utils import (
     get_node_supported_sstable_versions,
     ComparableScyllaVersion,
     is_enterprise,
+    get_node_enabled_sstable_version
 )
 from sdcm.sct_events.system import InfoEvent
 from sdcm.sct_events.database import (
@@ -480,10 +481,16 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
 
         :return:
         """
-        output = []
+        enabled_sstable_format_features = []
         for node in self.db_cluster.nodes:
-            output.extend(get_node_supported_sstable_versions(node.system_log))
-        return max(set(output))
+            enabled_sstable_format_features.extend(get_node_supported_sstable_versions(node.system_log))
+
+        if not enabled_sstable_format_features:
+            for node in self.db_cluster.nodes:
+                with self.db_cluster.cql_connection_patient_exclusive(node) as session:
+                    enabled_sstable_format_features.extend(get_node_enabled_sstable_version(session))
+
+        return max(set(enabled_sstable_format_features))
 
     def wait_for_sstable_upgrade(self, node, queue=None):
         all_tables_upgraded = True
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -664,13 +664,13 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
         self.metric_has_data(
             metric_query='sct_cassandra_stress_write_gauge{type="ops", keyspace="keyspace1"}', n=5)
 
-        # start gemini write workload
-        if self.version_cdc_support():
-            InfoEvent(message="Start gemini during upgrade").publish()
-            gemini_thread = self.run_gemini(self.params.get("gemini_cmd"))
-            # Let to write_stress_during_entire_test complete the schema changes
-            self.metric_has_data(
-                metric_query='gemini_cql_requests', n=10)
+        InfoEvent(message="Start gemini during upgrade").publish()
+        gemini_cmd = self.params.get("gemini_cmd")
+        if self.enable_cdc_for_tables:
+            gemini_cmd += " --table-options \"cdc={'enabled': true}\""
+        gemini_thread = self.run_gemini(gemini_cmd)
+        self.metric_has_data(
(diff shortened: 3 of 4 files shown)


aleksbykov:
Not sure, do we need backport to 6.0

fruch:
> Not sure, do we need backport to 6.0

Take a look at the 6.0 upgrade tests, and you'll know if it's needed or not.

It couldn't hurt does it ?

fruch (changes requested):
@aleksbykov 

the job you mentioned with the fix, didn't reached that part at all

soyacz:
@aleksbykov please mark it as a draft - there are many commented out parts which look like testing only.

fruch:
@aleksbykov it would be helpful if you would communicate a bit of the situation, just marking people as reviewer, doesn't mean it would get attention.

I wasn't even aware this PR is fixing issues that I was assuming tablets team should be fixing(and they didn't as well)

test-cases/upgrades/rolling-upgrade.yaml
```diff
@@ -35,7 +35,8 @@ gemini_cmd: "gemini -d --duration 2h \
 --max-mutation-retries 5 --max-mutation-retries-backoff 500ms \
 --async-objects-stabilization-attempts 5 --async-objects-stabilization-backoff 500ms \
 --replication-strategy \"{'class': 'NetworkTopologyStrategy', 'replication_factor': '3'}\" \
---table-options \"cdc={'enabled': true}\" --test-username cassandra --test-password cassandra"
+--test-username cassandra --test-password cassandra"
```
yarongilor, line 38: Is cdc not covered at all? or covered on another upgrade scenario/keyspace?
Do we plan testing both upgrade modes of cdc+vnodes vs. tablets?

aleksbykov: As i know, now it is impossible to enable cdc on table with tablets due to issue #16317. So i disable it temporary for while it not be fixed

fruch: I would expect someone from the team handling tablets to have a list of things needed to be tested with tablets disabled

CDC, and rolling upgrades, or a mix of them, probably need to be on this list.

yarongilor: @fruch , i think the [following doc](https://docs.google.com/document/d/1WNW0Wo27AK9RoPhU38UHmX9TAzeILuoKv8OX4b7kn4Y/edit?usp=sharing) can help with that.

LWT, CDC, Resharding, and few more.
We can either adjust the current longevities for that or cover it under the new folder of ["tablets_disabled" PR](https://github.com/scylladb/scylla-cluster-tests/pull/7655).

aleksbykov:
> @aleksbykov it would be helpful if you would communicate a bit of the situation, just marking people as reviewer, doesn't mean it would get attention.
> 
> I wasn't even aware this PR is fixing issues that I was assuming tablets team should be fixing(and they didn't as well)

I agree my fault, that i didn't noticed. i just wait the fix of https://github.com/scylladb/scylladb/issues/18995 which should be fast, to remove the not need commit, but looks like it is delayed. 
The set of commits  fixes regular rolling upgrades with disabling the cdc.
Could @scylladb/qa-maintainers review the pr?

sdcm/fill_db_data.py
```diff
@@ -3033,8 +3034,11 @@ def _get_table_name_from_query(query):
     @staticmethod
     def cql_create_simple_tables(session, rows):
         """ Create tables for truncate test """
-        create_query = "CREATE TABLE IF NOT EXISTS truncate_table%d (my_id int PRIMARY KEY, col1 int, value int) " \
-            "with cdc = {'enabled': true, 'ttl': 0}"
+        create_query = "CREATE TABLE IF NOT EXISTS truncate_table%d (my_id int PRIMARY KEY, col1 int, value int) "
+        # Cannot create CDC log for a table,
+        # because keyspace uses tablets.See issue  # 16317.
+        if not is_tablets_feature_enabled(session):
```
yarongilor, line 3040: shouldn't it be backported to branch 6.0?

yarongilor (approved):
LGTM

sdcm/fill_db_data.py
aleksbykov: I didn't found this error upon upgrade from 5.4 to 6.0, looks this issue is only relevant only when create table with cdc only for new tables.

fruch: Still I think we should backport to 6.0

Next minor release should test 6.0.0 -> 6.0.1, and those changes would be needed

test-cases/upgrades/rolling-upgrade.yaml
fruch: @aleksbykov something doesn't add up here, where are you removing cdc from gemini command, and not skipping the gemini command

seems from the code that the gemini command was added for the sake of testing CDC.

now if someone would disable raft and tablets, gemini would still run without cdc.
not sure that the expected outcome.

i.e. one should consider the support matrix of both options, when looking at those changes

upgrade_test.py
```diff
@@ -356,6 +357,8 @@ def _upgrade_node(self, node, upgrade_sstables=True, new_scylla_repo=None, new_v
             self.upgradesstables_if_command_available(node)
             InfoEvent(message='upgrade_node - ended to "upgradesstables_if_command_available"').publish()
 
+        self.db_cluster.wait_all_nodes_un()
```
fruch, line 360: can you elaborate on this change ? 

before we didn't wait for the node to be up normal, before moving to the next one ?

why do we need to check them all ? is the docs for upgrades were update with this addition ?

fruch: I think we need better references for changing the procedure like that

sdcm/fill_db_data.py
fruch: a suggestion 

we already have `version_cdc_support` which has bunch of relatively old versions

but if one would put the logic to disable cdc related testing into that function.
it would cleanly disable cdc related checks, without doing it in multiple places.

beside this one, that seems to have not be guarded by version_cdc_support

fruch (commented):
few suggestions/comments

mainly one should consider the mix of disabling raft topology and tablets, and that those test would still be operational

sdcm/fill_db_data.py
roydahan: @fruch is right, also no one will remember to enable it back.

upgrade_test.py
roydahan: It's safer, we added the same thing in rolling restart.
Before raft, it was a bit faster so we harfly noticed it, but in general of course all should wait for UN.

aleksbykov:
@fruch , can you take a look?
I added new property to check status of tablets. and change a bit gemini part in rolling_upgrade tests. Now if tablets is enabled gemini will be started without cdc and if tablets will be disabled, cdc options will be added to gemini command.
jobs are passed:
 - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/rolling-upgrade-staging-abykov/job/rolling-upgrade-ami-test/20 - Passed
 - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/rolling-upgrade-staging-abykov/job/rolling-upgrade-ubuntu22.04-test/11

I decided not add new sct parameter which could control behavior of enabling/disabling cdc with tablets, and left version_cdc_support method without changes.
i think in next PR this method could be safely removed at all, because we don't have version without cdc support

upgrade_test.py
```diff
@@ -664,10 +675,16 @@ def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-stat
...
 
-        # start gemini write workload
-        if self.version_cdc_support():
+        gemini_cmd = self.params.get("gemini_cmd")
+        gemini_thread = None
+
+        if gemini_cmd:
             InfoEvent(message="Start gemini during upgrade").publish()
-            gemini_thread = self.run_gemini(self.params.get("gemini_cmd"))
+            # TODO: workaround for issue #16317
```
fruch, line 683: #16317  ? in which repo ?

upgrade_test.py
```diff
@@ -777,7 +794,9 @@ def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-stat
         if all(upgradesstables):
             InfoEvent(message='Upgrading sstables if new version is available').publish()
             tables_upgraded = self.db_cluster.run_func_parallel(func=self.wait_for_sstable_upgrade)
-            assert all(tables_upgraded), "Failed to upgrade the sstable format {}".format(tables_upgraded)
+            # assert all(tables_upgraded), "Failed to upgrade the sstable format {}".format(tables_upgraded)
```
fruch, line 797: what is this comment ?

sdcm/fill_db_data.py
```diff
@@ -3179,6 +3185,12 @@ def version_cdc_support(self):
             version_with_support = self.CDC_SUPPORT_MIN_VERSION
         return self.parsed_scylla_version >= version_with_support
 
+    @optional_cached_property
```
fruch, line 3188: why `optional_cached_property` ? it can return None ?

upgrade_test.py
```diff
@@ -479,6 +484,12 @@ def get_highest_supported_sstable_version(self):  # pylint: disable=invalid-name
         output = []
         for node in self.db_cluster.nodes:
             output.extend(get_node_supported_sstable_versions(node.system_log))
+
+        if not output:
+            for node in self.db_cluster.nodes:
+                with self.db_cluster.cql_connection_patient_exclusive(node) as session:
+                    output.extend(get_node_enabled_sstable_version(session))
```
fruch, line 491: I don't understand this code ?

why not replacing this one with the previous ?

upgrade_test.py
```diff
@@ -664,10 +675,16 @@ def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-stat
         self.metric_has_data(
             metric_query='sct_cassandra_stress_write_gauge{type="ops", keyspace="keyspace1"}', n=5)
 
-        # start gemini write workload
-        if self.version_cdc_support():
```
fruch, line 668: you wrote you gonna remove it in a follow PR, but started removing usage of it here...

upgrade_test.py
```diff
@@ -777,7 +794,9 @@ def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-stat
         if all(upgradesstables):
             InfoEvent(message='Upgrading sstables if new version is available').publish()
             tables_upgraded = self.db_cluster.run_func_parallel(func=self.wait_for_sstable_upgrade)
-            assert all(tables_upgraded), "Failed to upgrade the sstable format {}".format(tables_upgraded)
+            # assert all(tables_upgraded), "Failed to upgrade the sstable format {}".format(tables_upgraded)
+            if not all(tables_upgraded):
```
fruch, line 798: why switching from assert to error print ?

if there's an issue in scylla, please point to it.

and use SkipPerIssue, so it would be stated back when issue is resolved

upgrade_test.py
```diff
@@ -664,10 +675,16 @@ def test_rolling_upgrade(self):  # pylint: disable=too-many-locals,too-many-stat
...
-        if self.version_cdc_support():
+        gemini_cmd = self.params.get("gemini_cmd")
+        gemini_thread = None
+
+        if gemini_cmd:
             InfoEvent(message="Start gemini during upgrade").publish()
-            gemini_thread = self.run_gemini(self.params.get("gemini_cmd"))
+            # TODO: workaround for issue #16317
+            if self.version_cdc_support() and not self.tablets_enabled:
+                gemini_cmd += " --table-options \"cdc={'enabled': true}\""
```
fruch, line 685: I don't understand what's the point of running gemini without cdc ?

if the whole purpose it was introduce was for testing cdc...

roydahan: It's not the whole point of running Gemini.
Let's not remove it.

upgrade_test.py
aleksbykov: It was rewritten. repo added

upgrade_test.py
aleksbykov: This commit was removed. Issue with sstable format was fixed.

upgrade_test.py
aleksbykov: I renamed the variable to be more meaningful. and also i don't remove old code, which use logs, because  master branch is also used with different version, where old functionality is used.
So if sstable_format was not parsed from logs (as in old versions) then it will be get from scylla table for new versions

upgrade_test.py
aleksbykov: removed this method at all. all latest supported scylla versions supports the cdc.

upgrade_test.py
aleksbykov: Issue was fixed. i remove the commit at all, because it is not needed anymore

upgrade_test.py
aleksbykov: a bit change the code but save the main idea, to run gemini during rolling upgrade with and without cdc support

aleksbykov:
new :
 - removed not related commit after issue was fixed: scylladb/scylladb#18995
 - removed old method and vars for version_cdc_support
 - set properities as cached_property
 - add new property to control enabling cdc options for tables based on feature tablets state and SkipPerIssue
 - rename var 'output' -> ''enabled_sstable_format_features' in method 'get_highest_supported_sstable_version' and leave support 2 behaviors get formats from logs and if not found from scylla table

aleksbykov:
@fruch can you take a look?

upgrade_test.py
```diff
@@ -476,10 +481,16 @@ def get_highest_supported_sstable_version(self):  # pylint: disable=invalid-name
...
         :return:
         """
-        output = []
+        enabled_sstable_format_features = []
         for node in self.db_cluster.nodes:
-            output.extend(get_node_supported_sstable_versions(node.system_log))
-        return max(set(output))
+            enabled_sstable_format_features.extend(get_node_supported_sstable_versions(node.system_log))
+
+        if not enabled_sstable_format_features:
```
fruch, line 488: I still don't understand this fallback.. and why it's here...

upgrade_test.py
fruch: reading it from logs doesn't works for new versions  ?
reading it from system tables doesn't work for old versions ?

I'm a bit a lot, what exactly happened that we need to touch any of those parts

if we do, I would want to know why, and make sure we don't keep duplication for no real reason.

fruch (approved):
LGTM

fruch:
@aleksbykov I would expect a follow up on sstable upgrade part, to clean it clarify it up.

upgrade_test.py
roydahan, line 1: @fruch / @syuu1228 IIUC, we need to revert this commit but only from branch-6.1.
In master the tests are from version without jmx to version without jmx.
But in 6.1 we do it from 6.0 that is with jmx (AFAIU) to 6.1 that is without jmx.

Am i right here?

fruch: what about 2024.2 ? 

I think we just need to fix this change, `s/autobackup/backup/`, and test it to be working

roydahan: Does 2024.2 have jmx in it or not?

And do we have any documentation or change in procedure for upgrades with jmx and without?

roydahan: Now @yaronkaikov told me that 6.0 also doesn't have jmx installed.

roydahan: ah but we didn't have the PR from @aleksbykov to remove it from SCT.

fruch: Yes, cause the issue is when upgrading from a version that doesn't have jmx

syuu1228: > @fruch / @syuu1228 IIUC, we need to revert this commit but only from branch-6.1. In master the tests are from version without jmx to version without jmx. But in 6.1 we do it from 6.0 that is with jmx (AFAIU) to 6.1 that is without jmx.

If we won't revert this commit on master, we should apply fix like this:
```
diff --git a/upgrade_test.py b/upgrade_test.py
index 7b8835318..4c9de0425 100644
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -117,7 +117,7 @@ def recover_conf(node):
         node.remoter.run(
             r'for conf in $( rpm -qc $(rpm -qa | grep scylla) | grep -v contains ) '
             r'/etc/systemd/system/{var-lib-scylla,var-lib-systemd-coredump}.mount; '
-            r'do if test -e $conf.backup; then sudo cp -v $conf.backup $conf; fi; done')
+            r'do if test -e $conf.autobackup; then sudo cp -v $conf.autobackup $conf; fi; done')
     else:
         node.remoter.run(
             r'for conf in $(cat /var/lib/dpkg/info/scylla-*server.conffiles '
```
Otherwise upgrade test will keep failing.
Since backup_conf() is backuping io.conf to io.conf.autobackup, but recover_conf() is trying to restore the file from io.conf.backup which is not available, and fails.
It is not about the patch purpose (restore only existant files since we dropping java things), it's about backup filename mismatch (maybe mistake on the patch?).

fruch: Yes it was a mistake in that patch

Since centos8 was deprecated quite before that change was merged, it wasn't noticed until we had centos9 

You have the diff needed please open a PR, and then we could backport it to where it's needed.

syuu1228: > Yes it was a mistake in that patch
> 
> Since centos8 was deprecated quite before that change was merged, it wasn't noticed until we had centos9
> 
> You have the diff needed please open a PR, and then we could backport it to where it's needed.

Okay I will

Issue #7334, opened 2024-04-11
`hydra run-test` stuck at "Building Docker image", but no CPU is used


kbr-scylla:
I ran
```
export SCT_STRESS_DURATION=5
export SCT_SCYLLA_VERSION="2024.1.2"
export SCT_DOCKER_IMAGE="scylladb/scylla-enterprise"

hydra run-test longevity_test.LongevityTest.test_custom_time --backend docker \
  --config configurations/nemesis/additional_configs/docker_backend_local.yaml \
  --config configurations/nemesis/ModifyTableMonkey.yaml

```

and it's stuck at
```
< t:2024-04-11 11:52:03,720 f:tester.py       l:393  c:LongevityTest        p:INFO  > SCT Test duration: 365; Stress duration: 5; Prepare duration: 300
< t:2024-04-11 11:52:03,728 f:base.py         l:267  c:sdcm.sct_events.base p:WARNING > Unable to get events main device: You should create default EventsProcessRegistry first
< t:2024-04-11 11:52:03,769 f:tester.py       l:411  c:LongevityTest        p:INFO  > Submitted Argus TestRun with test id <MagicMock name='mock.run_id' id='139767728255184'>
< t:2024-04-11 11:52:03,771 f:tester.py       l:417  c:LongevityTest        p:INFO  > sct_runner info in Argus TestRun is updated
< t:2024-04-11 11:52:03,772 f:tooling_reporter.py l:26   c:sdcm.reporting.tooling_reporter p:INFO  > PythonDriverReporter(): scylla-cluster-tests/python-driver version is 3.25.11
< t:2024-04-11 11:52:13,632 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-04-11 11:52:13.628: (InfoEvent Severity.NORMAL) period_type=not-set event_id=20c1b0ee-83f5-4da1-a613-7d751b11bb73: message=TEST_START test_id=6cefca84-9701-4004-b749-7f5dc02d29bc
< t:2024-04-11 11:52:13,701 f:cluster.py      l:3103 c:sdcm.cluster         p:INFO  > <sdcm.cluster_docker.ScyllaDockerCluster object at 0x7f1e37fe0e50>: Init nodes
[...]

PR #7526, opened 2024-06-02
feature(docker-backend): log progress of building/pulling

fruch:
in cases there a problematic internet connection, or other issues that might slow down the build, it's not clear to a user what's going on during that time.

with this change we'll have progress on the sct.log

Closes: #7334

### Testing

- [x] docker provision test

### Reminders

### DIFF
--- a/sdcm/utils/docker_utils.py
+++ b/sdcm/utils/docker_utils.py
@@ -19,11 +19,13 @@ from pprint import pformat
 from types import SimpleNamespace  # pylint: disable=no-name-in-module
 from typing import List, Optional, Union, Any, Tuple, Protocol
 from functools import cache
+import itertools
 
 import docker
 from docker.errors import DockerException, NotFound, ImageNotFound, NullResource, BuildError
 from docker.models.images import Image
 from docker.models.containers import Container
+from docker.utils.json_stream import json_stream
 
 from sdcm.remote import LOCALRUNNER
 from sdcm.remote.base import CommandRunner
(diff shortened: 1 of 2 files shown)


sdcm/utils/docker_utils.py
```diff
@@ -399,18 +401,29 @@ def build_container_image(cls, instance: INodeWithContainerManager, name: str, *
...
-                            "\n".join(item['stream'] for item in docker_build_error.build_log if "stream" in item))
-            raise
-
-        LOGGER.debug(">>> Build log for Docker image `%s': >>>", image_tag)
-        for entry in logs:
-            if "stream" in entry:
-                LOGGER.debug(entry["stream"].rstrip())
-        LOGGER.debug("<<<")
+
+        resp = docker_client.api.build(tag=image_tag, **dockerfile_args, **build_args)
```
soyacz, line 405: please add a note that this code was taken from docker module, so in case it fails (e.g. regexp/api changes) we know where to copy&adapt from again.

sdcm/utils/docker_utils.py
```diff
@@ -399,18 +401,29 @@ def build_container_image(cls, instance: INodeWithContainerManager, name: str, *
...
+            if 'error' in chunk:
+                raise BuildError(chunk['error'], result_stream)
+            if 'stream' in chunk:
+                LOGGER.info(chunk["stream"].rstrip())
+                match = re.search(
+                    r'(^Successfully built |sha256:)([0-9a-f]+)$',
+                    chunk['stream']
+                )
+                if match:
+                    image_id = match.group(2)
```
vponomaryov, line 420: Parsing out the image ID from the logs looks fragile.
We don't make any influence on the docker source code and dependency on the output structure may make us fix it periodically for each of the supported SCT branches.

Cannot we make some addition API call after the `build` one finishes?

fruch: no, this is exactly the code from the `docker.images.build`

sdcm/utils/docker_utils.py
vponomaryov: @fruch please, move the ported code to a separate function and call it here for ease of future updates.

sdcm/utils/docker_utils.py
vponomaryov: Since you want to keep it as-is then added comment above.

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

PR #7541, opened 2024-06-03
ci(pipeline): add possibility to byo scylladb for longevity and upgrades

vponomaryov:
It becomes more and more useful to build custom scylladb images
with unmerged changes to the scylladb repo.

So, add possibility to do it by utilizing the "byo" pipeline available
in the "scylla-pkg" project. This pipeline gets utilized as an existing
Jenkins CI job which must be refered to in the "byo_job_path" parameter
like the following:
 
```   
./my-custom-scylla-byo-job/
```

### Testing

- [ ]

### Reminders

### DIFF
--- a/vars/perfRegressionParallelPipeline.groovy
+++ b/vars/perfRegressionParallelPipeline.groovy
@@ -211,7 +211,7 @@ def call(Map pipelineParams) {
                                         }
 
                                         stage("Provision Resources for ${sub_test}") {
-                                                catchError() {
+                                                catchError(stageResult: 'FAILURE') {
                                                     script {
                                                         wrap([$class: 'BuildUser']) {
                                                             dir('scylla-cluster-tests') {
--- a/vars/provisionResources.groovy
+++ b/vars/provisionResources.groovy
@@ -67,19 +67,21 @@ def call(Map params, String region){
 
     if [[ -n "${params.scylla_ami_id ? params.scylla_ami_id : ''}" ]] ; then
         export SCT_AMI_ID_DB_SCYLLA="${params.scylla_ami_id}"
-    elif [[ -n "${params.gce_image_db ? params.gce_image_db : ''}" ]] ; then
+    fi
+    if [[ -n "${params.gce_image_db ? params.gce_image_db : ''}" ]] ; then
         export SCT_GCE_IMAGE_DB="${params.gce_image_db}"
-    elif [[ -n "${params.azure_image_db ? params.azure_image_db : ''}" ]] ; then
+    fi
+    if [[ -n "${params.azure_image_db ? params.azure_image_db : ''}" ]] ; then
         export SCT_AZURE_IMAGE_DB="${params.azure_image_db}"
-    elif [[ -n "${params.scylla_version ? params.scylla_version : ''}" ]] ; then
+    fi
+    if [[ -n "${params.scylla_version ? params.scylla_version : ''}" ]] ; then
         export SCT_SCYLLA_VERSION="${params.scylla_version}"
-    elif [[ -n "${params.scylla_repo ? params.scylla_repo : ''}" ]] ; then
+    fi
+    if [[ -n "${params.scylla_repo ? params.scylla_repo : ''}" ]] ; then
         export SCT_SCYLLA_REPO="${params.scylla_repo}"
-    elif [[ "${params.backend ? params.backend : ''}" == *"k8s"* ]] ; then
-        echo "Kubernetes backend can have empty scylla version. It will be taken from defaults of the scylla helm chart"
-    else
-        echo "need to choose one of SCT_AMI_ID_DB_SCYLLA | SCT_SCYLLA_VERSION | SCT_SCYLLA_REPO | SCT_GCE_IMAGE_DB"
-        exit 1
+    fi
+    if [[ -n "${params.new_scylla_repo ? params.new_scylla_repo : ''}" ]] ; then
+        export SCT_NEW_SCYLLA_REPO="${params.new_scylla_repo}"
     fi
 
     if [[ -n "${params.oracle_scylla_version ? params.oracle_scylla_version : ''}" ]] ; then
--- a/vars/rollingUpgradePipeline.groovy
+++ b/vars/rollingUpgradePipeline.groovy
@@ -73,6 +73,23 @@ def call(Map pipelineParams) {
             string(defaultValue: '',
                    description: 'Actual user requesting job start, for automated job builds (e.g. through Argus)',
                    name: 'requested_by_user')
+
+            // NOTE: Optional parameters for BYO ScyllaDB stage
+            string(defaultValue: '',
+                   description: 'Custom "scylladb" repo to use. Leave empty if byo is not needed. If specified then need to define "base_versions" param explicitly.',
+                   name: 'byo_scylla_repo')
+            string(defaultValue: '',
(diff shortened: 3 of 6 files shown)


vponomaryov:
Examples:
- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-10gb-3h-test/103
- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-10gb-3h-gce-test/26
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-rolling-upgrade-ubuntu22.04-test/4

Note that it depends on the update of the "scylla-pkg" project (to be uploaded)

vars/rollingUpgradePipeline.groovy
```diff
@@ -170,6 +187,103 @@ def call(Map pipelineParams) {
                                                 }
                                             }
                                         }
+                                        stage('BYO Scylladb [optional]') {
+                                            catchError(stageResult: 'FAILURE') {
+                                                script {
+                                                    wrap([$class: 'BuildUser']) {
+                                                        dir('scylla-cluster-tests/scylla-pkg') {
```
fruch, line 194: what's in `scylla-pkg` directory ? why not move to SCT top level dir as the rest of the places ?

vars/rollingUpgradePipeline.groovy
```diff
@@ -170,6 +187,103 @@ def call(Map pipelineParams) {
...
+                                                            if ( params.byo_scylla_repo && params.byo_scylla_branch ) {
```
fruch, line 195: can't we package this whole logic into it's own module, like the rest of the stage are implemented ?

fruch: so we can reuse it in more pipelines, as needed ?

vars/perfRegressionParallelPipeline.groovy
```diff
@@ -211,7 +211,7 @@ def call(Map pipelineParams) {
                                         }
 
                                         stage("Provision Resources for ${sub_test}") {
-                                                catchError() {
+                                                catchError(stageResult: 'FAILURE') {
```
fruch, line 214: how this change is related to this PR ?

fruch:
@vponomaryov 

overwall looks good, it would be helpful if there a section on docs/ with a few pointer on how it should be used
and then as followup we should sit down with @k0machi to define the rebuild/clone forms for it, which validation are needed and such.

vars/rollingUpgradePipeline.groovy
vponomaryov: If you look at the both approach for longevity and upgrades, you will see that they differ.
We can move it out, but it will be 2 modules, not one.

vars/rollingUpgradePipeline.groovy
vponomaryov: Because here we call remote project's pipeline which we do not control.
It may have it's own output files and so on.

Do you some problems having a sub-dir?

fruch: It's just not exactly clear why, it looks like we might be cloning something, while we don't.

vars/perfRegressionParallelPipeline.groovy
fruch: o.k.

vars/rollingUpgradePipeline.groovy
fruch: yes having it as modules would be helpful, so it can be applied to more pipelines.

I don't want to see people start copying the whole thing from pipeline to pipeline.

vars/rollingUpgradePipeline.groovy
fruch: also what do you mean with output files ? we are just calling a job, it's running on a different job, and gonna generate any files, and output files we need, we'll need to pull on our own from the job, don't we ?

vars/provisionResources.groovy
```diff
@@ -67,19 +67,21 @@ def call(Map params, String region){
...
+    if [[ -n "${params.scylla_version ? params.scylla_version : ''}" ]] ; then
         export SCT_SCYLLA_VERSION="${params.scylla_version}"
-    elif [[ -n "${params.scylla_repo ? params.scylla_repo : ''}" ]] ; then
+    fi
+    if [[ -n "${params.scylla_repo ? params.scylla_repo : ''}" ]] ; then
         export SCT_SCYLLA_REPO="${params.scylla_repo}"
-    elif [[ "${params.backend ? params.backend : ''}" == *"k8s"* ]] ; then
-        echo "Kubernetes backend can have empty scylla version. It will be taken from defaults of the scylla helm chart"
-    else
-        echo "need to choose one of SCT_AMI_ID_DB_SCYLLA | SCT_SCYLLA_VERSION | SCT_SCYLLA_REPO | SCT_GCE_IMAGE_DB"
```
fruch, line 81: it's bit problematic we take those out, now if we have conflicting parameters like that, we might not notice it...

this kind of issue with the artifact tests, waste me lots of hours, until I've figure I've test the wrong thing (I've passing both scylla_repo and unified_packgae, and we don't have any code that warns you about it's conflicting)

vars/rollingUpgradePipeline.groovy
fruch: not critical

vars/rollingUpgradePipeline.groovy
vponomaryov: done

vars/provisionResources.groovy
vponomaryov: If you find a way to check env vars at this point then let me know.

Without ability to check env vars setting explicitly in pipeline we should either hack with writing to files (ugly, IMHO) or remove these exclusive checks (current variant).

vponomaryov:
Latest code is tested here:
- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-10gb-3h-test/104
- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-rolling-upgrade-ubuntu22.04-test/5

vars/provisionResources.groovy
fruch: I'm not following exactly

I understand this current logic, doesn't apply for the case you are doing.

I don't understand why we can't change the logic to fit, and how reading environment variable is related to that (and what is the problem preventing reading environment variables)

I'm arguing that this check is important, and taking it out, would could lead to confusion if one didn't supplied information about which scylla he wants to run (i.e. the default when you trigger a job)

vponomaryov: > I'm not following exactly
> 
> I understand this current logic, doesn't apply for the case you are doing.

We get failures not having any of the required parameters. These changes are a must.

> I don't understand why we can't change the logic to fit, and how reading environment variable is related to that (and what is the problem preventing reading environment variables)

This is the jenkins bug that complicates things: https://issues.jenkins.io/browse/JENKINS-28921

> I'm arguing that this check is important, and taking it out, would could lead to confusion if one didn't supplied information about which scylla he wants to run (i.e. the default when you trigger a job from jenkins)

We have checks in SCT itself.

fruch: well, we'll need to count on Argus to have those check as well.

and use argus as to avoid those silly mistakes

fruch (approved):
LGTM

PR #7550, opened 2024-06-04
fix(wait_for_cloud_init): cases cloud-init isn't installed

fruch:
in some artifact test we do have cases the distro isn't installing cloud-init by default, and we should be calling it at all.

### Testing

- [ ]

### Reminders


vponomaryov (approved):
LGTM

soyacz (approved):

PR #7554, opened 2024-06-04
improvement(sct-config): convert append_scylla_yaml to dict

dimakr:
Currently the 'append_scylla_yaml' attribute is defined in test configurations as a multiline string representation of a dict. And is processed correspondingly in SCT.
The downside of it is if there are several test configurations for a test run that have 'append_scylla_yaml' attribute in them, SCTConfiguration overwrites the attribute value for each loaded config. As a result the last test configuration 'wins', in terms of that only its value is added to scylla.yaml.

The change converts value of all 'append_scylla_yaml' attribute occurrences in configs to a dict, so that it is natively merged by anyconfig. It also adjusts places where 'append_scylla_yaml' is used in the code, to process it
as a dict, not a string.

Related task: https://github.com/scylladb/scylla-cluster-tests/issues/7364

### Testing
- [["configurations/audit.yaml","test-cases/features/stop-compaction.yaml"] configs](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/sct-debug/51/) this is to ensure that 'append_scylla_yaml' values from different configs are merged
- [["test-cases/longevity/longevity-topology-changes-3h.yaml","configurations/raft/enable_raft_experimental.yaml"] configs](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-AbortRepairMonkey-docker-test/37/) - experimental_features attribute is supported as a standalone attribute in test config, but we had a number of instances where it was a part of append_scylla_yaml. This build is to ensure that experimental_features attribute is correctly merged to scylla.yaml, after moving out of append_scylla_yaml.
- [["test-cases/artifacts/ubuntu2004-fips.yaml", "configurations/local-ear.yaml"] configs](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/artifacts-ubuntu2004-fips-test/) this is to ensure that when there were nested subattributes in string version of append_scylla_yaml, after conversion to dict they are correctly merged and added to the final scylla.yaml.
- [["test-cases/longevity/longevity-encryption-at-rest-200GB-6h.yaml", "configurations/kms-ear.yaml"] configs](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/append_scylla_yaml_test/2/) - this is to ensure that KMS logic is not affected, as its attributes was changed in this commit.

### Reminders

### DIFF
--- a/configurations/audit.yaml
+++ b/configurations/audit.yaml
@@ -1,5 +1,5 @@
-append_scylla_yaml: |
-  audit: "syslog"
-  audit_categories: "DCL,DDL,AUTH,ADMIN"
-  audit_tables: "keyspace1.standard1,scylla_bench.test"
-  audit_keyspaces: "keyspace1,scylla_bench"
+append_scylla_yaml:
+  audit: 'syslog'
+  audit_categories: 'DCL,DDL,AUTH,ADMIN'
+  audit_tables: 'keyspace1.standard1,scylla_bench.test'
+  audit_keyspaces: 'keyspace1,scylla_bench'
--- a/configurations/tablets-initial-32.yaml
+++ b/configurations/tablets-initial-32.yaml
@@ -1,5 +1,5 @@
-append_scylla_yaml: |
-  experimental_features:
-    - tablets
-    - consistent-topology-changes
+experimental_features:
+  - tablets
+  - consistent-topology-changes
+append_scylla_yaml:
   tablets_initial_scale_factor: 4
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -536,7 +536,7 @@ class SCTConfiguration(dict):
         dict(name="append_scylla_args_oracle", env="SCT_APPEND_SCYLLA_ARGS_ORACLE", type=str,
              help="More arguments to append to oracle command line"),
 
-        dict(name="append_scylla_yaml", env="SCT_APPEND_SCYLLA_YAML", type=str,
+        dict(name="append_scylla_yaml", env="SCT_APPEND_SCYLLA_YAML", type=dict_or_str,
              help="More configuration to append to /etc/scylla/scylla.yaml"),
 
         # Nemesis config options
--- a/test-cases/jepsen/jepsen_with_raft.yaml
+++ b/test-cases/jepsen/jepsen_with_raft.yaml
@@ -20,6 +20,5 @@ jepsen_test_count: 5
 jepsen_test_run_policy: any
 bare_loaders: true
 
-append_scylla_yaml: |
-  experimental_features:
-    - raft
+experimental_features:
+  - raft
--- a/test-cases/performance/perf-regression-throughput-cdc-mixed-poll-batching.yaml
+++ b/test-cases/performance/perf-regression-throughput-cdc-mixed-poll-batching.yaml
@@ -24,9 +24,8 @@ store_perf_results: true
 append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
 
 
-append_scylla_yaml: |
-  experimental_features:
-    - cdc
+experimental_features:
+  - cdc
 
 email_recipients: ['scylla-perf-results@scylladb.com', 'cdc@scylladb.com']
 backtrace_decoding: false
--- a/test-cases/performance/perf-regression-write-throughput-cdc.yaml
+++ b/test-cases/performance/perf-regression-write-throughput-cdc.yaml
@@ -17,9 +17,8 @@ store_perf_results: true
 append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
 
 
-append_scylla_yaml: |
-  experimental_features:
-    - cdc
+experimental_features:
+  - cdc
 
 email_recipients: ['scylla-perf-results@scylladb.com', 'cdc@scylladb.com ']
 backtrace_decoding: false
--- a/configurations/hytrust-kmip-ear.yaml
+++ b/configurations/hytrust-kmip-ear.yaml
@@ -2,17 +2,16 @@
 scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmipKeyProviderFactory', 'kmip_host': 'kmip_test'}"
 
 # enable system_info_encryption, config kmip_hosts
-append_scylla_yaml: |
-  system_key_directory: /etc/encrypt_conf/
+append_scylla_yaml:
+  system_key_directory: '/etc/encrypt_conf/'
   system_info_encryption:
-      enabled: True  # system_info_encryption
-      key_provider: LocalFileSystemKeyProviderFactory  # system_info_encryption
-      secret_key_file: '/etc/scylla/encrypt_conf/system_info_encryption_keyfile'
-
+    enabled: true  # system_info_encryption
+    key_provider: 'LocalFileSystemKeyProviderFactory'  # system_info_encryption
+    secret_key_file: '/etc/scylla/encrypt_conf/system_info_encryption_keyfile'
   kmip_hosts:
-       kmip_test:
-           hosts: 52.21.171.245
-           certificate: /etc/encrypt_conf/hytrust-kmip-scylla.pem
-           keyfile: /etc/encrypt_conf/hytrust-kmip-scylla.pem
-           truststore: /etc/encrypt_conf/hytrust-kmip-cacert.pem
-           priority_string: 'SECURE128:+RSA:-VERS-TLS1.0:-ECDHE-ECDSA'
+    kmip_test:
+      hosts: '52.21.171.245'
+      certificate: '/etc/encrypt_conf/hytrust-kmip-scylla.pem'
+      keyfile: '/etc/encrypt_conf/hytrust-kmip-scylla.pem'
+      truststore: '/etc/encrypt_conf/hytrust-kmip-cacert.pem'
+      priority_string: 'SECURE128:+RSA:-VERS-TLS1.0:-ECDHE-ECDSA'
--- a/configurations/kmip-ear.yaml
+++ b/configurations/kmip-ear.yaml
@@ -2,17 +2,14 @@
 scylla_encryption_options: "{ 'cipher_algorithm' : 'AES/ECB/PKCS5Padding', 'secret_key_strength' : 128, 'key_provider': 'KmipKeyProviderFactory', 'kmip_host': 'kmip_test'}"
 
 # enable system_info_encryption, config kmip_hosts
-append_scylla_yaml: |
-  system_key_directory: /etc/encrypt_conf/
+append_scylla_yaml:
+  system_key_directory: '/etc/encrypt_conf/'
   system_info_encryption:
-      enabled: True  # system_info_encryption
-      key_provider: LocalFileSystemKeyProviderFactory  # system_info_encryption
-      secret_key_file: '/etc/scylla/encrypt_conf/system_info_encryption_keyfile'
-
+    enabled: true  # system_info_encryption
+    key_provider: 'LocalFileSystemKeyProviderFactory'  # system_info_encryption
+    secret_key_file: '/etc/scylla/encrypt_conf/system_info_encryption_keyfile'
   kmip_hosts:
-       kmip_test:
-           hosts: kmip-interop1.cryptsoft.com
-           certificate: /etc/encrypt_conf/SCYLLADB.pem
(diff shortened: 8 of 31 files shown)


dimakr:
There is a failed unit tests - unit_tests/test_base_version.py::TestBaseVersion::test_2021_1_with_centos8.
But I am not sure that this is something related to the change in question.

sdcm/sct_config.py
```diff
@@ -1678,13 +1678,36 @@ def __init__(self):
...
+            Custom merge strategy for anyconfig to merge 'append_scylla_yaml' attribute values from multiple configs.
+
+            :param self: mapping object to update with `other`
+            :param other: mapping object to update `self`
+            :param key: key of mapping object to update
+            :param val: value to update self alternatively
+            """
+            if key == 'append_scylla_yaml':
+                if key in self and key in other:
+                    self[key] += other[key]
```
fruch, line 1693: I'm sure pure string append would be enough in this case....

maybe we need to parse it from yaml, and then merge and turn it back to string ?

what happens if we'll have multiple values ?

i.e. keep in mind `append_scylla_yaml` is currently defined as string, that holds a yaml, and not part of the yaml

fruch: regardless I think this specific change might need a unit test or a few of them
proving how it would work

fruch:
> There is a failed unit test - unit_tests/test_base_version.py::TestBaseVersion::test_2021_1_with_centos8. But I am not sure that this is something related to the change in question.

if that test keep failing, we can drop it.
it's point to a very old release folder, that might have been deleted.

even that the failure seem like some networking issue

sdcm/sct_config.py
dimakr: >I'm sure pure string append would be enough in this case....

Not sure, maybe I'm missing something from the comment.
The `self[key] += other[key]` exactly appends value of `append_scylla_yaml` from one config (config that is being merged) to the value of the other (config to which merge is performed)

dimakr: >regardless I think this specific change might need a unit test or a few of them
proving how it would work

I'll unit test. Test run on real test configurations show that merging the value works as expected.

fruch: build a test that both config file has the same key in `append_scylla_yaml`, and you'll get a broken yaml inside of it

config1
```
append_scylla_yaml: |
  enable_sstables_mc_format: true
  enable_sstables_md_format: true
```

config2
```
append_scylla_yaml: |
  enable_sstables_mc_format: false
  enable_sstables_md_format: false
```

the outcome would be:

```
append_scylla_yaml: |
  enable_sstables_mc_format: true
  enable_sstables_md_format: true
  enable_sstables_mc_format: false
  enable_sstables_md_format: false
```

wouldn't it ?

fruch: we need the ability to overwrite things inside the yaml **IN** `append_scylla_yaml` value

soyacz:
@dimakr did you try replacing type of `append_scylla_yaml` sct config from `str` to `dict_or_str` so it behaves the same as `stress_image` which support override out of the box?

sdcm/sct_config.py
fruch: we can change the definition of `append_scylla_yaml` to stop being a string, and make it a first class yaml structure.
but we'll need to carefully attend to the 68 usages of it while doing so. (it's not a very big change, but need to be careful)

fruch: @soyacz saying the same https://github.com/scylladb/scylla-cluster-tests/pull/7554#issuecomment-2146955643

with with the exact details and example of other config option handling the same situation.

dimakr: Ok, so we don't want to be able to merge values of `append_scylla_yaml` attribute as they are used now in the STC.
But to rework how this attribute is used and get rid of defining its value as a multiline string everywhere in SCT?

dimakr: got it, will be back soon

dimakr:
I converted `append_scylla_yaml` values in test configs from multiline string to dict values.
Also adjusted few places in code that previously were loading the string value of the attribute.

I executed few builds that use test configurations with `append_scylla_yaml` - there are passed and failed runs, which is expected as such test configurations fail in other jobs.
But in the scope of the current change - to ensure that merging of `append_scylla_yaml` values works - they do the job:
- [["configurations/audit.yaml","test-cases/features/stop-compaction.yaml"] configs](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/sct-debug/51/) this one is to ensure that 'append_scylla_yaml' values from different configs are merged
- [["test-cases/longevity/longevity-topology-changes-3h.yaml","configurations/raft/enable_raft_experimental.yaml"] configs](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-AbortRepairMonkey-docker-test/37/). I see that `experimental_features` attribute is supported as a standalone attribute in test config, but we had a number of instances where it was a part of  `append_scylla_yaml`. This build is to ensure that `experimental_features` attribute is correctly merged to scylla.yaml, after moving out of  `append_scylla_yaml`.
- [["test-cases/artifacts/ubuntu2004-fips.yaml", "configurations/local-ear.yaml"] configs](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/artifacts-ubuntu2004-fips-test/) just another run to ensure that when there were nested subattributes in string version of `append_scylla_yaml`, after conversion to dict they are correctly merged and added to the final scylla.yaml

@soyacz @fruch

configurations/audit.yaml
```diff
@@ -1,5 +1,5 @@
-append_scylla_yaml: |
-  audit: "syslog"
-  audit_categories: "DCL,DDL,AUTH,ADMIN"
-  audit_tables: "keyspace1.standard1,scylla_bench.test"
-  audit_keyspaces: "keyspace1,scylla_bench"
+append_scylla_yaml:
+  audit: 'syslog'
```
fruch, line 2: our of curiosity, when move from double quote to single ?

test-cases/jepsen/jepsen_with_raft.yaml
```diff
@@ -20,6 +20,5 @@ jepsen_test_count: 5
 jepsen_test_run_policy: any
 bare_loaders: true
 
-append_scylla_yaml: |
-  experimental_features:
-    - raft
+experimental_features:
```
fruch, line 23: this job need to be delete... there's no such flag anymore

test-cases/performance/perf-regression-throughput-cdc-mixed-poll-batching.yaml
```diff
@@ -24,9 +24,8 @@ store_perf_results: true
 append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
 
 
-append_scylla_yaml: |
-  experimental_features:
-    - cdc
+experimental_features:
```
fruch, line 27: those needs to be remove

test-cases/performance/perf-regression-write-throughput-cdc.yaml
```diff
@@ -17,9 +17,8 @@ store_perf_results: true
 append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
 
 
-append_scylla_yaml: |
-  experimental_features:
-    - cdc
+experimental_features:
```
fruch, line 20: those needs to be remove 

cdc isn't experimental for quite some time

test-cases/performance/perf-regression-throughput-cdc-mixed-poll-batching.yaml
fruch: cdc isn't experimental for quite some time

fruch (approved):
beside some cleanups, which are critical 

this looks version, and would improve the usability and readability of sct configuration

test-cases/jepsen/jepsen_with_raft.yaml
dimakr: Only `experimental_features` attribute is to be deleted, or the whole test config + its dedicated jenkins job?

configurations/audit.yaml
dimakr: Single quotes are always safer in yaml, so that there is no need to escape spec. characters (like `\` or `"`, etc.) if they are introduced to string values in future.

soyacz: I agree, e.g. double quotes in `defaults/manager_persistent_snapshots.yaml` caused us quite a lot of troubles - as are included in parsed dict.

configurations/tablets-initial-32.yaml
```diff
@@ -1,5 +1,5 @@
-append_scylla_yaml: |
-  experimental_features:
-    - tablets
-    - consistent-topology-changes
+experimental_features:
+  - tablets
+  - consistent-topology-changes
```
soyacz, line 3: btw. recently these got removed from experimental.
To avoid hassle with merge conflicts later I'd propose to switch it to `enable_tablets: true` right away

soyacz:
I propose to run a test with KMS - @dimakr please run one test with Scylla enterprise so it's enabled and we see it's working.

configurations/tablets-initial-32.yaml
dimakr: Maybe it's better to do both changes - cleanup of usages of these 2 features as experimental and add `enable_tablets` flag to scylla.yaml - in a separate PR?
For `consistent-topology-changes` there is a test config, that sets only this option - enables as experimental feature - `consistent-topology-changes`. That test config is used in 31 places in the SCT.
For `tablets` - it is also used in 2 test configs that do only that - enable it as experimental feature. And these 2 configs are used in 11 places in SCT.

soyacz: yes, thinking again we should split it to separate PR as it will also require change in `ScyllaYaml` - too much for this change.

vponomaryov (approved):
LGTM.

Agree that would be good to see an enterprise Scylla with KMS be tested.

fruch:
@dimakr did you test it with enterprise ?

dimakr:
> @dimakr did you test it with enterprise ?

Still in progress. Jon was killed due to spot instances.

fruch:
> > @dimakr did you test it with enterprise ?
> 
> @fruch Still in progress. Job was killed due to spot instances.

@dimakr 

I'm guessing this is the job running 
https://argus.scylladb.com/test/d86e92f6-f70b-4680-9882-67352cf5b0de/runs?additionalRuns[]=f8a7dbe1-b28b-4fb2-a98d-df2e392bf633

I would recommend putting a link to it, on the description (even when it's ongoing)

also please update the PR description, i.e. remove the part under testing that refer to examples of string concatenation, it would be confusing to who gonna read it later on

dimakr:
> I'm guessing this is the job running [https://argus.scylladb.com/test/d86e92f6-f70b-4680-9882-67352cf5b0de/runs?
Yes, this is the job. At the end I had to use on_demand instances as with spot ones there were always capacity related interruptions.
This job failed. But it failed with the same error as it had in the past, e.g. https://jenkins.scylladb.com/view/enterprise-2024.1/job/enterprise-2024.1/job/SCT_Enterprise_Features/job/EncryptionAtRest/job/EaR-longevity-kms-200gb-6h-test/2/.
Can it be considered that changes in attributes merging around KMS attributes haven't affected that logic?
@fruch @soyacz

soyacz:
> > I'm guessing this is the job running [https://argus.scylladb.com/test/d86e92f6-f70b-4680-9882-67352cf5b0de/runs?
> > Yes, this is the job. At the end I had to use on_demand instances as with spot ones there were always capacity related interruptions.
> > This job failed. But it failed with the same error as it had in the past, e.g. https://jenkins.scylladb.com/view/enterprise-2024.1/job/enterprise-2024.1/job/SCT_Enterprise_Features/job/EncryptionAtRest/job/EaR-longevity-kms-200gb-6h-test/2/.
> > Can it be considered that changes in attributes merging around KMS attributes haven't affected that logic?
> > @fruch @soyacz

You can  download db logs and see if related stuff is there in scylla.yaml files. Also can look if keyspaces were altered to be encrypted (possibly by searching for `ALTER ...` in sct log.

fruch:
> > I'm guessing this is the job running [https://argus.scylladb.com/test/d86e92f6-f70b-4680-9882-67352cf5b0de/runs?
> > Yes, this is the job. At the end I had to use on_demand instances as with spot ones there were always capacity related interruptions.

any job that run more than 3-4 hours, spot would be problematic for it, we are using instance which has quite high demand (some of it by us...) 

> > This job failed. But it failed with the same error as it had in the past, e.g. https://jenkins.scylladb.com/view/enterprise-2024.1/job/enterprise-2024.1/job/SCT_Enterprise_Features/job/EncryptionAtRest/job/EaR-longevity-kms-200gb-6h-test/2/.
> > Can it be considered that changes in attributes merging around KMS attributes haven't affected that logic?
> > @fruch @soyacz

PR #7561, opened 2024-06-04
ci(github-actions): add action to force backport labels

fruch:
we have it on scylla core repo, and we should have it here in SCT as well, so people open PRs would need to specify backport before it can be merged

### Testing

- [ ]

### Reminders

### DIFF
--- a/.github/workflows/pr-require-backport-label.yaml
+++ b/.github/workflows/pr-require-backport-label.yaml
@@ -1,0 +1,22 @@
+name: PR require backport label
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+    branches:
+      - master
+
+jobs:
+  label:
+    if: github.event.pull_request.draft == false
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: minimum
+          count: 1
+          labels: "backport/none\nbackport/[\\w-]*[\\d.]*$"
+          use_regex: true
+          add_comment: false


.github/workflows/pr-require-backport-label.yaml
```diff
@@ -0,0 +1,22 @@
+name: PR require backport label
+on:
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+    branches:
+      - master
+      - next
```
vponomaryov, line 7: We don't have "next"

.github/workflows/pr-require-backport-label.yaml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: minimum
+          count: 1
+          labels: "backport/none\nbackport/\\d.\\d"
```
vponomaryov, line 20: This regex won't work on some of our labels such as `backport/manager` and `backport/perf-v14`.

.github/workflows/pr-require-backport-label.yaml
fruch: removed

.github/workflows/pr-require-backport-label.yaml
fruch: replaced it with a more fitting one.

.github/workflows/pr-require-backport-label.yaml
```diff
@@ -0,0 +1,22 @@
...
+    runs-on: ubuntu-latest
+    permissions:
+      issues: write
+      pull-requests: write
+    steps:
+      - uses: mheap/github-action-required-labels@v5
+        with:
+          mode: minimum
+          count: 1
+          labels: "backport/none\nbackport/.*-v?[\\d.]*$"
```
vponomaryov, line 20: It won't catch the `backport/2024.1` label

vponomaryov (changes requested):

.github/workflows/pr-require-backport-label.yaml
fruch: I think now it's a better version of the regex, that does cover all of the options

.github/workflows/pr-require-backport-label.yaml
vponomaryov: Not removed.

fruch: now it's removed

vponomaryov (approved):
LGTM

Issue #7240, opened 2024-03-04
`disrupt_add_remove_dc` nemesis breaks `disrupt_add_drop_column` one running in parallel


vponomaryov:
### Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

If we run `disrupt_add_remove_dc` nemesis in parallel to the `disrupt_add_drop_column` one then we can get following error:

```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5116, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2449, in disrupt_add_drop_column
    self._add_drop_column_run_in_cycle()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2178, in _add_drop_column_run_in_cycle
    self._add_drop_column()
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2144, in _add_drop_column
    self._add_drop_column_target_table = self._add_drop_column_get_target_table(
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2069, in _add_drop_column_get_target_table
    current_tables = self._get_all_tables_with_no_compact_storage(self._add_drop_column_tables_to_ignore)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2060, in _get_all_tables_with_no_compact_storage
    tables = get_db_tables(session, ks, with_compact_storage=False)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 2101, in get_db_tables
    for table in list(session.cluster.metadata.keyspaces[ks].tables.keys()):
KeyError: 'keyspace_new_dc'
```

It is caused by the concurrency of a new keyspace addition (`disrupt_add_remove_dc nemesis`)
and driver session update in addition to the unsafe coding assuming driver's session (`disrupt_add_drop_column` nemesis) knows about that newly added keyspace.

#### Steps to Reproduce

1. Run any longevity where above 2 mentioned nemesis run in parallel
2. See error

**Expected behavior:** The `disrupt_add_drop_column` should not fail with the `KeyError: 'keyspace_new_dc'` error.

[...]

PR #7565, opened 2024-06-05
fix(longevity): get table for 'add_drop_column'

juliayakovlev:
If run disrupt_add_remove_dc nemesis in parallel to the disrupt_add_drop_column one then we can get following error:
```
  KeyError: 'keyspace_new_dc'
```

It is caused by the concurrency of a new keyspace addition (disrupt_add_remove_dc nemesis) and driver session update in addition to the unsafe coding assuming driver's session (disrupt_add_drop_column nemesis) knows about that newly added keyspace. Another possible problem: new added keyspace was dropped in exacly time when we run describe of the table.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/7240

### Testing

- [x] [longevity-schema-topology-changes-12h](https://argus.scylladb.com/test/ed048b23-03ef-4486-a005-3cbac0163c87/runs?additionalRuns[]=e64dc80b-8b56-4652-83f4-06764c5e6a3a)

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2040,7 +2040,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if tables_to_skip is None:
             tables_to_skip = {}
         to_be_skipped_default = tables_to_skip.get('*', '').split(',')
-        with self.cluster.cql_connection_patient(self.tester.db_cluster.nodes[0]) as session:
+        with self.cluster.cql_connection_patient(self.target_node) as session:
             query_result = session.execute('SELECT keyspace_name FROM system_schema.keyspaces;')
             for result_rows in query_result:
                 keyspaces.extend([row.lower()
@@ -2055,7 +2055,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                     to_be_skipped = []
                 else:
                     to_be_skipped = to_be_skipped.split(',') + to_be_skipped_default
-                tables = get_db_tables(session, ks, with_compact_storage=False)
+                tables = get_db_tables(session=session,
+                                       keyspace_name=ks,
+                                       node=self.target_node,
+                                       with_compact_storage=False)
                 if to_be_skipped:
                     tables = [table for table in tables if table not in to_be_skipped]
                 if not tables:
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -54,6 +54,7 @@ from pathlib import Path
 
 import requests
 import boto3
+from invoke import UnexpectedExit
 from mypy_boto3_s3 import S3Client, S3ServiceResource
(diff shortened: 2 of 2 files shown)


juliayakovlev:
Test is running now and it looks ok meanwhile

sdcm/utils/common.py
```diff
@@ -2116,12 +2118,20 @@ def get_db_tables(session, ks, with_compact_storage=True):
...
+            LOGGER.error("Failed to describe '%s.%s' table. Maybe the table has been deleted. Error: %s",
+                         keyspace_name, row.table_name, err.result.stderr)
+            continue
+
         if with_compact_storage is None:
-            output.append(table)
-        elif ("with compact storage" in table_code.lower()) == with_compact_storage:
-            output.append(table)
+            output.append(row.table_name)
+        elif ("WITH COMPACT STORAGE" in create_table_statement) == with_compact_storage:
```
vponomaryov, line 2133: Are you sure about the text case?
Maybe better to add ".upper()" to the "create_table_statement" var to avoid possible case mismatches in some of the cql versions?

sdcm/utils/common.py
```diff
@@ -2116,12 +2118,20 @@ def get_db_tables(session, ks, with_compact_storage=True):
...
+            output.append(row.table_name)
```
vponomaryov, line 2134: Looks like you have the same actions for 2 conditions.
So, it can be simplified to the following:

```
if ("WITH COMPACT STORAGE" in (create_table_statement.upper() or "")) == with_compact_storage:
    output.append(row.table_name)
```

sdcm/nemesis.py
```diff
@@ -2055,7 +2055,10 @@ def _get_all_tables_with_no_compact_storage(self, tables_to_skip=None):
                     to_be_skipped = []
                 else:
                     to_be_skipped = to_be_skipped.split(',') + to_be_skipped_default
-                tables = get_db_tables(session, ks, with_compact_storage=False)
+                tables = get_db_tables(session=session,
+                                       keyspace_name=ks,
+                                       node=self.tester.db_cluster.nodes[0],
```
vponomaryov, line 2060: Are we sure that this first node is always alive considering that it may be disrupted by concurrent nemesis?

vponomaryov (changes requested):

sdcm/utils/common.py
juliayakovlev: it added above, in the line 2123

sdcm/nemesis.py
juliayakovlev: Ok, this node is used for session create. But I think you are right. Changing it to target_node

sdcm/utils/common.py
juliayakovlev: changed the condition

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #7578, opened 2024-06-05
fix(alternator): improve the lookup logic of `salted_hash`

fruch:
since we are looking on multiple keyspaces, it wasn't enough to handle the case of non-existing keyspaces, but also on existing keyspace and table, and empty responses

### Testing

- [x] test with unit tests (the PR to fix them, would come separately)

### Reminders


vponomaryov (approved):
LGTM if it works.

fruch:
> LGTM if it works.

Tested with integration tests, and applied and tested this logic in dtest as well

PR #7591, opened 2024-06-06
improvement(speed): enable parallel db nodes startup

soyacz:
Now we can start db nodes in parallel (since 6.0).

Add config parameter `parallel_node_operations` which adds control over consistent topology changes goodies gave to Scylla.

### Testing

- [x] - provision tests (with false settings)
- [x] - [1h-individual-nemesis-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-MajorCompactionMonkey-aws-test/12/)
- [x] - [non-root-artifact-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-amazon2023-nonroot-test/5/)

### Reminders


soyacz:
@roydahan @fruch I was thinking of one parameter that will control parallel nodes operations we now support - so in future grow/shrink operations will use it also. Please review if you agree with this approach or we want to fine-grained control for each operation.

defaults/test_default.yaml
```diff
@@ -71,6 +71,7 @@ email_subject_postfix: ''
 collect_logs: false
 
 hinted_handoff: 'disabled'
+parallel_node_operations: true  # supported from Scylla 6.0
```
vponomaryov, line 74: Making it to be default looks fragile approach as of now.
"master" SCT gets used not only for latest Scylla versions.
Personally I use lots of different Scylla versions with master SCT.

Can we make it be implicit by checking scylla version?

roydahan: I agree, checking scylla version and/or feature enabled should be the better/safer approach.
But, we may still want to have parameter to control it (even if supported by the version/feature).

sdcm/cluster.py
```diff
@@ -3857,9 +3859,20 @@ def critical_node_setup_events():
...
             while len(setup_results) != len(node_list):
                 verify_node_setup_or_startup(start_time, setup_queue, setup_results)
-            # startup serially
+            # startup
             for node in node_list:
-                node_startup(node, startup_queue)
+                if cl_inst.parallel_startup:
+                    threading.Thread(target=node_startup, args=(node, startup_queue), daemon=True).start()
+                else:
+                    node_startup(node, startup_queue)
```
fruch, line 3867: lets try once if we could bootstrap all of the node at the same time

soyacz: yes it also works. Code adapted to be fully parallel and actually also simpler.

soyacz:
@fruch I set flag default to false and made it `true` for typical longevity-100gb-4h. Let me know if it's fine.

roydahan:
> @fruch I set flag default to false and made it `true` for typical longevity-100gb-4h. Let me know if it's fine.

I think that the default should be true.
6.0 is already out with this.

sdcm/sct_config.py
```diff
@@ -386,6 +386,9 @@ class SCTConfiguration(dict):
         dict(name="use_mgmt", env="SCT_USE_MGMT", type=boolean,
              help="When define true, will install scylla management"),
 
+        dict(name="parallel_node_operations", env="SCT_PARALLEL_NODE_OPERATIONS", type=boolean,
+             help="When defined true, will run node operations in parallel. Supported operations: startup"),
```
roydahan, line 390: You already have the code for add node and decommission, aren't you?
I think you can add it to this PR.

soyacz: decommission / grow is more complicated and requires some work (e.g. in my code it doesn't work yet with multi-az).
I wanted to make this one in and later add more complicated ones (and easier to revert in case of issues).

fruch:
> > @fruch I set flag default to false and made it `true` for typical longevity-100gb-4h. Let me know if it's fine.
> 
> I think that the default should be true. 6.0 is already out with this.

I disagree, it's on the  team that owns that feature, to properly test across different types of tests, before enabling it for all of the tests.
@soyacz did it for a specific reason, enable something on all case should be done with caution and with more testing

soyacz:
> > > @fruch I set flag default to false and made it `true` for typical longevity-100gb-4h. Let me know if it's fine.
> > 
> > 
> > I think that the default should be true. 6.0 is already out with this.
> 
> I disagree, it's on the team that owns that feature, to properly test across different types of tests, before enabling it for all of the tests. @soyacz did it for a specific reason, enable something on all case should be done with caution and with more testing

I'm not having a strong opinion for that. Both have own pros&cons, so I chosen safer one.

soyacz:
@fruch @roydahan I have no strong opinion, please agree with each other what approach we take here.

vponomaryov:
I agree with @fruch , the default should not be changed.
We don't even have the enterprise version yet supporting tablets.

I see addition of this new option to pipeline parameters as a really proper way now, so we could choose before triggering each test run and configure triggers for correct versions.

soyacz:
> I agree with @fruch , the default should not be changed. We don't even have the enterprise version yet supporting tablets.
> 

Enterprise dev version supports that.

> I see addition of this new option to pipeline parameters as a really proper way now, so we could choose before triggering each test run and configure triggers for correct versions.

I'm not a fan of adding new pipeline parameters, as there is too many of them and people whether don't know or can't decide.
Version is verified, so if it is not supported, then it is disabled. The only exception is when someone forces gossip instead raft for node operations (but can disable parallel_node_operations along).
Other operations like grow/shrink can be decided based on enabled features instead of versions (when cluster is up).

sdcm/cluster.py
```diff
@@ -3857,9 +3858,23 @@ def critical_node_setup_events():
...
                 verify_node_setup_or_startup(start_time, setup_queue, setup_results)
-            # startup serially
+            # startup
+            parallel_startup = False
+            node = node_list[0]
+            if isinstance(cl_inst, BaseScyllaCluster) and cl_inst.parallel_node_operations:
+                if node.is_enterprise:
+                    parallel_startup = ComparableScyllaVersion(node.scylla_version) >= "2024.2.0~dev"
+                else:
+                    parallel_startup = ComparableScyllaVersion(node.scylla_version) >= "6.0.0~rc2"
```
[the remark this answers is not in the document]
soyacz: fixed

sdcm/cluster.py
aleksbykov: additionally we need to check that scylla.yaml doesn't include the `force_gossip_topology_changes=true` .  Because in this case raft topology is disabled, and nodes could n't start in parallel. 
Safer way for cluster startup is start first node, check that CONSISTENT_TOPOLOGY_CHANGES is enabled, and then start in parallel all other nodes. 
all next nodes adding to cluster could be added to cluster based on sct parameter or based CONSISTENT_TOPOLOGY_CHANGES got from first node

soyacz: this complicates it a bit. I saw you are implementing this force in different PR - how about giving some more logic in node property (`self.parallel_node_operations = self.params.get("parallel_node_operations") or False` to include this condition when you implement it?

fruch: More reason why this should be default, it should be carefully set/unset for some cases, i.e. gossip mode ones, upgrade tests can enable it either.

We could start with the tier1 to have it battle proof, and then consider the next steps and what needed to be tested

soyacz: ok, I'll make it false by default and later we can create PR with setting it as true where we can discuss pros and cons of this change.
In the meantime I think I'll refactor this a bit by moving decision part to `node` class instead of `wait_for_init_wrap` so we can add more logic in a clearer way.

roydahan:
Default should be "true".
If there will be issues with that we will need to find them out.

@temichus your team should give it a proper testing once we enable that, since both enabling this feature and 6.0 testing didn't cover that.

fruch:
> Default should be "true".

At the end, yes, after proper testing

> If there will be issues with that we will need to find them out.

What happened to testing things prior to getting them merged ?

> 
> @temichus your team should give it a proper testing once we enable that, since both enabling this feature and 6.0 testing didn't cover that.
>

roydahan:
> What happened to testing things prior to getting them merged ?
> 
> 

We passed that when the feature was merged almost 2 months ago.
Let's set it as the default and live with the consequences.
Otherwise also 2024.2 won't have enough mileage with that.

soyacz:
@roydahan @fruch this is ready to be merged. Let's have it running for a week and enable it by default then.

roydahan:
I don't think we need it but it just wasting time now, so I'm merging and let's submit the next one to be readt.

roydahan (approved):

PR #7605, opened 2024-06-09
improvement(event): find Scylla reactor_stall issue by known keyword

juliayakovlev:
This commit presents new helpful mechanism that allow to find the known Scylla issue by backtrace while decoding and add this issue to the event.
New mapping file keeps "keyword-issue" for reactor stalls only for now. But it can be expanded for other backtraces (for example, coredumps) . 
Found issue is added to the new event attribute 'known_issue'. 
Argus may add this issue to the event caption using the new attribute.

Example of event:
```
2024-06-09 09:34:34.233 <2024-06-09 09:34:34.186>: (DatabaseLogEvent Severity.DEBUG) period_type=one-time event_id=5e3d5d75-154b-44db-9a48-af9dbfe8753b: 
type=REACTOR_STALLED regex=Reactor stalled line_number=9249 node=rolling-upgrade--ubuntu-jammy-db-node-7d4f6b49-0-2 
known_issue=https://github.com/scylladb/scylladb/issues/88282
```

### Testing

- [x] [rolling_upgrade_with_sla](https://argus.scylladb.com/test/872fbe83-8dcc-4069-b010-f88efb64fc96/runs?additionalRuns[]=7d4f6b49-0be1-4eb4-9879-f3cc111f55a3) - got and recognized reactor stalled with known issue keyword
- [x] [longevity-100gb-4h](https://argus.scylladb.com/test/657bf1f3-69a5-4c36-bdbf-bf17718f7087/runs?additionalRuns[]=ea72ebe6-905c-4bf0-8f3a-41c32c7972e5) - valudate that other events are not broken

### Reminders


test-cases/longevity/longevity-100gb-4h.yaml
```diff
@@ -6,12 +6,12 @@ stress_cmd: ["cassandra-stress write cl=QUORUM duration=240m -schema 'replicatio
              "cassandra-stress read  cl=QUORUM duration=240m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=40 -pop seq=1..20971520 -col 'n=FIXED(10) size=FIXED(512)' -log interval=5"
              ]
 
-n_db_nodes: 6
-n_loaders: 2
+n_db_nodes: 3
```
fruch, line 9: try doing those in it's own commits

now it's hard to take it out, let's assume I will want to merge...

sdcm/cluster.py
```diff
@@ -1499,12 +1500,19 @@ def decode_backtrace(self):
                     scylla_debug_file = self.copy_scylla_debug_info(obj["node"], obj["debug_file"])
                 output = self.decode_raw_backtrace(scylla_debug_file, " ".join(event.raw_backtrace.split('\n')))
                 event.backtrace = output.stdout
+                self.log.debug("Search for issue for %s event", event.type)
```
fruch, line 1503: once ready get rid of all of the debug prints, they are a bit excessive

sdcm/sct_events/base.py
```diff
@@ -497,6 +498,8 @@ def msgfmt(self):
             fmt += " line_number={0.line_number}"
         if self.node is not None:
             fmt += " node={0.node}"
+        if hasattr(self, "known_issue") and self.known_issue:
```
fruch, line 501: I won't if we could make it a bit more visible...

since people won't notice it, if they are not looking for it....

sdcm/utils/issues_by_keyword/find_known_issue.py
```diff
@@ -0,0 +1,38 @@
...
+        self.map = defaultdict(list)
+        for backtrace_type, values in yaml_details.items():
+            for one_pair in values:
+                self.map[backtrace_type].append(MapKeywords(keyword=one_pair["keyword"], issue=one_pair["issue"]))
+        LOGGER.debug("MapKeywords: %s", self.map)
+
+    def find_issue(self, backtrace_type: str, decoded_backtrace: str) -> str | None:
+        LOGGER.debug("Search for an issue for %s backtrace", backtrace_type)
+        for all_keywords in self.map.get(backtrace_type, []):
+            if all_keywords.keyword in decoded_backtrace:
```
fruch, line 36: we might consider finding more then one issue per backtraces ? and make it known_issues ?

juliayakovlev: Not that I am aware about such case

juliayakovlev: It maybe few keywords for one issue.
If we open the issue by backtrace and there is the already opened an issue for the same/similar backtrace, the new opened issue is closed as duplicate. Even it is OSS and Enterprise
So I think it is not expected few issues for same backtrace

fruch (approved):
LGTM

sdcm/utils/issues_by_keyword/issue_by_keyword.yaml
```diff
@@ -0,0 +1,12 @@
...
+    issue: https://github.com/scylladb/scylladb/issues/2577
+
+  - keyword: partition_entry::upgrade
+    issue: https://github.com/scylladb/scylladb/issues/2577
+
+  - keyword: utils::uleb64_express_encode
+    issue: https://github.com/scylladb/scylladb/issues/8828
+
+  - keyword: logalloc::region_impl::object_descriptor::encode stalls
+    issue: https://github.com/scylladb/scylladb/issues/8828
```
soyacz, line 12: Aren't these keywords too generic and we start to mislabel events with wrong issues?
Did you discuss these keywords with developers?

When investigating I'm afraid we'll start to ignore all events labelled as known issue. Idea to prevent it would be to change severity based on issue status - if issue is closed severity should be increased to error (to possibly reopen it or create new one).

juliayakovlev: I received those keywords from Avi

juliayakovlev: > Aren't these keywords too generic and we start to mislabel events with wrong issues? Did you discuss these keywords with developers?
> 
> When investigating I'm afraid we'll start to ignore all events labelled as known issue. Idea to prevent it would be to change severity based on issue status - if issue is closed severity should be increased to error (to possibly reopen it or create new one).

Hmm, this is the good idea

fruch: SkipPerIssue can be used here

sdcm/utils/issues_by_keyword/find_known_issue.py
fruch: I thought more of cases we have multiple issue within the same backtrace...
Don't know if it gonna happen a lot or not, but not that critical for now, if we'll need it, it would be easily expended

soyacz (approved):
LGTM

sdcm/utils/issues_by_keyword/issue_by_keyword.yaml
fruch: @juliayakovlev let's add it as a follow up

juliayakovlev: @fruch do I remember correct that we need to add "skip..<version>" label to the issue?

juliayakovlev: 'sct-{branch_version}-skip'

juliayakovlev:
@fruch when do you plan to backport it?

PR #7608, opened 2024-06-09
fix(syslogng-install): fix retries in installation script

fruch:
Now the the ssh "lock" we have if out, there are few reentrancy issue we need to solve in the syslog-ng installation
script
 
### Testing

- [x] provision tests

### Reminders

### DIFF
--- a/sdcm/provision/common/utils.py
+++ b/sdcm/provision/common/utils.py
@@ -141,10 +141,10 @@ def install_syslogng_service():
             else
                 cat /etc/apt/sources.list
                 for n in 1 2 3 4 5 6 7 8 9; do # cloud-init is running it with set +o braceexpand
-                    if apt-get -y update 2>&1 | tee /tmp/syslog_ng_install.output || grep NO_PUBKEY \
-        /tmp/syslog_ng_install.output; then
+                    if apt-get -y update ; then
                         break
                     fi
+                    sleep 0.5
                 done
 
                 for n in 1 2 3; do # cloud-init is running it with set +o braceexpand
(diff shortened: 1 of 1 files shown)


sdcm/provision/common/utils.py
```diff
@@ -135,16 +135,16 @@ def install_syslogng_service():
...
-                apt-get purge -y syslog-ng*
-                DPKG_FORCE=confmiss apt-get --reinstall -y install syslog-ng
+                apt-get purge -o DPkg::Lock::Timeout=300 -y syslog-ng*
+                DPKG_FORCE=confmiss apt-get --reinstall -o DPkg::Lock::Timeout=300 -y install syslog-ng
                 SYSLOG_NG_INSTALLED=1
             else
                 cat /etc/apt/sources.list
                 for n in 1 2 3 4 5 6 7 8 9; do # cloud-init is running it with set +o braceexpand
-                    if apt-get -y update 2>&1 | tee /tmp/syslog_ng_install.output || grep NO_PUBKEY \
-        /tmp/syslog_ng_install.output; then
```
soyacz, line 145: @fruch do you have idea why it was done this way?

soyacz (approved):
LGTM

sdcm/provision/common/utils.py
fruch: I can only guess, someone was trying to ignore specific errors

No have the key configured correctly, properly was quite common at some point

But regardless it was wrong, and broken

It does point we might still want some kind of lock on this logging setup, and the done file isn't completely enough

vponomaryov (approved):
LGTM

fruch:
> @fruch I got problem with syslog-ng installation again in branch-2023.1. The branch should include this fix:
> 
> https://argus.scylladb.com/test/599ae8f5-b6e0-463c-af6f-74b8a2702438/runs?additionalRuns[]=1d189b65-9f7e-4fa7-9b72-724098439251

Please raise an SCT issue, looks like it's something else, cloud init was done, but syslog-ng wasn't installed.

Seems like it reproduces, one would need a closer look into logs to see what going on

soyacz:
In db logs you can find that `E: Unable to locate package syslog-ng`, I don't know yet why.

PR #7613, opened 2024-06-10
fix(hydra): use python3.11 `Don’t prepend a potentially unsafe path to sys.path`

fruch:
cause if we don't, we have some modules in utils which are identical in their names to some builtins, and it would fail like that:
```
  File "..././sdcm/utils/get_username.py", line 15, in <module>
    import getpass
  File ".../3.11.3/lib/python3.11/getpass.py", line 17, in <module>
    import contextlib
  File ".../3.11.3/lib/python3.11/contextlib.py", line 6, in <module>
    from collections import deque
  File ".../3.11.3/lib/python3.11/collections/__init__.py", line 36, in <module>
    from operator import eq as _eq
ImportError: cannot import name 'eq' from 'operator'
(.../sdcm/utils/operator/__init__.py)
```

this snippet of python code is run before we get into the hydra image hence we don't control the python version being used...

### Testing

- [ ]

### Reminders

### DIFF
--- a/docker/env/hydra.sh
+++ b/docker/env/hydra.sh
@@ -2,6 +2,9 @@
 
 set -eo pipefail
 
+# make sure the call to get_username.py doesn't import any local modules
+export PYTHONSAFEPATH=true
+
 CMD=$@
 DOCKER_ENV_DIR=$(readlink -f "$0")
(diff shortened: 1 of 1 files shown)


yaronkaikov (approved):

soyacz (approved):
LGTM

PR #7632, opened 2024-06-11
test(manager): Introduce the sanity test on mixed cluster (vnodes+tablets)

mikliapko:
The PR partially addresses https://github.com/scylladb/scylla-manager/issues/3853.

The PR consists of:
- New test that performs sanity checks on cluster with mixed keyspaces - 1 vnodes keyspace + 1 tablets keyspace;
- .jenkinsfile that handles test execution in CI;
- A few fixes/extensions for some load/stress methods to handle non-default or several keyspace names;
- Fix for system_auth keyspace repair test.

### Testing

- [x] New test: 
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-sanity-test/16/
- [x] Jenkins job:
https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-vnodes-tablets-cluster-test/
- [x] Regression sanity test: 
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/debian11-sanity-test/2/
- [x] Regression upgrade test: 
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-upgrade-test/1/
- [x] Ubuntu sanity (Scylla 6.0) to verify repair test fixes:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-sanity-test/17/

### Reminders

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -200,30 +200,31 @@ class BackupFunctionsMixIn(LoaderUtilsMixin):
         if restore_schema:
             self.db_cluster.restart_scylla()  # After schema restoration, you should restart the nodes
 
-    def run_verification_read_stress(self):
+    def run_verification_read_stress(self, ks_names=None):
         stress_queue = []
         stress_cmd = self.params.get('stress_read_cmd')
         keyspace_num = self.params.get('keyspace_num')
         InfoEvent(message='Starting read stress for data verification').publish()
         stress_start_time = datetime.now()
-        self.assemble_and_run_all_stress_cmd(stress_queue, stress_cmd, keyspace_num)
+        if ks_names:
+            self.assemble_and_run_all_stress_cmd_by_ks_names(stress_queue, stress_cmd, ks_names)
+        else:
+            self.assemble_and_run_all_stress_cmd(stress_queue, stress_cmd, keyspace_num)
         for stress in stress_queue:
             self.verify_stress_thread(cs_thread_pool=stress)
         stress_run_time = datetime.now() - stress_start_time
         InfoEvent(message=f'The read stress run was completed. Total run time: {stress_run_time}').publish()
 
-    def _generate_load(self, keyspace_name_to_replace=None):
+    def _generate_load(self, keyspace_name: str = None):
         self.log.info('Starting c-s write workload')
         stress_cmd = self.params.get('stress_cmd')
-        if keyspace_name_to_replace:
-            stress_cmd = stress_cmd.replace("keyspace1", keyspace_name_to_replace)
-        stress_thread = self.run_stress_thread(stress_cmd=stress_cmd)
+        stress_thread = self.run_stress_thread(stress_cmd=stress_cmd, keyspace_name=keyspace_name)
         self.log.info('Sleeping for 15s to let cassandra-stress run...')
         time.sleep(15)
         return stress_thread
 
-    def generate_load_and_wait_for_results(self, keyspace_name_to_replace=None):
-        load_thread = self._generate_load(keyspace_name_to_replace=keyspace_name_to_replace)
+    def generate_load_and_wait_for_results(self, keyspace_name: str = None):
+        load_thread = self._generate_load(keyspace_name=keyspace_name)
         load_results = load_thread.get_results()
         self.log.info(f'load={load_results}')
 
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -841,8 +841,10 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         self.log.info('Task: {} is done.'.format(repair_task.id))
         self.log.debug("sctool version is : {}".format(manager_tool.sctool.version))
 
-        expected_keyspaces_to_be_repaired = ["system_auth", "system_distributed",  # pylint: disable=invalid-name
-                                             self.NETWORKSTRATEGY_KEYSPACE_NAME]
+        expected_keyspaces_to_be_repaired = ["system_distributed", self.NETWORKSTRATEGY_KEYSPACE_NAME]
+        if not self.db_cluster.nodes[0].raft.is_consistent_topology_changes_enabled:
+            expected_keyspaces_to_be_repaired.append("system_auth")
+        self.log.debug("Keyspaces expected to be repaired: {}".format(expected_keyspaces_to_be_repaired))
         per_keyspace_progress = repair_task.per_keyspace_progress
         self.log.info("Looking in the repair output for all of the required keyspaces")
         for keyspace_name in expected_keyspaces_to_be_repaired:
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -42,6 +42,7 @@ from sdcm.sct_events.system import InfoEvent
 from sdcm.sct_events.group_common_events import ignore_no_space_errors, ignore_stream_mutation_fragments_errors
 from sdcm.utils.gce_utils import get_gce_storage_client
 from sdcm.utils.azure_utils import AzureService
+from sdcm.utils.tablets.common import TabletsConfiguration
 from sdcm.exceptions import FilesNotCorrupted
 
 
@@ -172,21 +173,24 @@ class BackupFunctionsMixIn(LoaderUtilsMixin):
                             keyspace_and_table_list=keyspace_and_table_list)
 
     # pylint: disable=too-many-arguments
-    def verify_backup_success(self, mgr_cluster, backup_task, keyspace_name='keyspace1', tables_names=None,
+    def verify_backup_success(self, mgr_cluster, backup_task, ks_names: list = None, tables_names: list = None,
                               truncate=True, restore_data_with_task=False, timeout=None):
+        if ks_names is None:
+            ks_names = ['keyspace1']
         if tables_names is None:
             tables_names = ['standard1']
-        per_keyspace_tables_dict = {keyspace_name: tables_names}
+        ks_tables_map = {keyspace: tables_names for keyspace in ks_names}
         if truncate:
-            for table_name in tables_names:
-                self.log.info(f'running truncate on {keyspace_name}.{table_name}')
-                self.db_cluster.nodes[0].run_cqlsh(f'TRUNCATE {keyspace_name}.{table_name}')
+            for ks, tables in ks_tables_map.items():
+                for table_name in tables:
+                    self.log.info(f'running truncate on {ks}.{table_name}')
+                    self.db_cluster.nodes[0].run_cqlsh(f'TRUNCATE {ks}.{table_name}')
         if restore_data_with_task:
             self.restore_backup_with_task(mgr_cluster=mgr_cluster, snapshot_tag=backup_task.get_snapshot_tag(),
                                           timeout=timeout, restore_data=True)
         else:
             self.restore_backup_from_backup_task(mgr_cluster=mgr_cluster, backup_task=backup_task,
-                                                 keyspace_and_table_list=per_keyspace_tables_dict)
+                                                 keyspace_and_table_list=ks_tables_map)
 
     def restore_backup_with_task(self, mgr_cluster, snapshot_tag, timeout, restore_schema=False, restore_data=False,
                                  location_list=None):
@@ -335,7 +339,7 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         self.log.info("creating the table {} in the keyspace {}".format(table_name, keyspace_name))
         self.create_table(table_name, keyspace_name=keyspace_name)
 
-    def test_manager_sanity(self):
+    def test_manager_sanity(self, prepared_ks: bool = False, ks_names: list = None):
         """
         Test steps:
         1) Run the repair test.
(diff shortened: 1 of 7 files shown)


mgmt_cli_test.py
```diff
@@ -370,6 +372,30 @@ def test_manager_sanity(self):
...
+        """
+        Test steps:
+        1) Create tablets keyspace and propagate some data.
+        2) Create vnodes keyspace and propagate some data.
+        3) Run sanity test (test_manager_sanity).
+        """
+        self.log.info('starting test_manager_sanity_vnodes_tablets_cluster')
+
+        tablets_keyspace = "tablets_keyspace"
+        self.create_keyspace(tablets_keyspace, replication_factor=3)
```
soyacz, line 385: I'd be more explicit here and also added `TabletsConfiguration` so in case tablets are not turned on it should fail.

mikliapko: Fixed

mgmt_cli_test.py
```diff
@@ -173,21 +173,24 @@ def restore_backup_from_backup_task(self, mgr_cluster, backup_task, keyspace_and
...
         if tables_names is None:
             tables_names = ['standard1']
-        per_keyspace_tables_dict = {keyspace_name: tables_names}
+        ks_tables_map = {keyspace: tables_names for keyspace in ks_names}
         if truncate:
-            for table_name in tables_names:
-                self.log.info(f'running truncate on {keyspace_name}.{table_name}')
-                self.db_cluster.nodes[0].run_cqlsh(f'TRUNCATE {keyspace_name}.{table_name}')
+            for ks, tables in ks_tables_map.items():
+                for table_name in tables:
```
soyacz, line 185: nit (in general): I see a tendency to introduce small complications to save couple lines of code/reuse parts but decreasing readability. E.g. here instead of iterating over map, could just iterate over `ks_names` and `tables_names`.

That is my opinion only, just consider it.

sdcm/utils/loader_utils.py
```diff
@@ -63,21 +63,26 @@ def _create_counter_table(self):
                     AND speculative_retry = '99.0PERCENTILE';
             """)
 
-    def assemble_and_run_all_stress_cmd(self, stress_queue, stress_cmd, keyspace_num):
+    def assemble_and_run_all_stress_cmd(self, stress_queue, stress_cmd, keyspace_num, ks_names=None):
```
soyacz, line 66: This logic gets complicated and there's no clear doc what takes precedence over what.
Consider splitting this function.

sdcm/tester.py
```diff
@@ -291,6 +291,20 @@ class ClusterInformation(NamedTuple):
     schema_versions: List[SchemaVersion]
 
 
+@dataclass
+class TabletsConfiguration:
```
soyacz, line 295: nice one,
consider moving it to `replication_strategy_utils.py`

sdcm/utils/loader_utils.py
mikliapko: Agree about getting more complex. I will consider splitting.

sdcm/tester.py
mikliapko: Will move, thanks

mgmt_cli_test.py
mikliapko: For this specific case, I'd prefer to keep it without changes.
It doesn't add much complexity and prevents us from further changes of these lines of code if we decide to rework the tables per keyspace assembling approach (for example, create different tables for different keyspaces).

mikliapko:
Comments addressed, retesting the changes:
- New test: https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-vnodes-tablets-cluster-test/6/
- Regression for sanity test: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/debian11-sanity-test/3/

roydahan:
The code is ok, but I think the approach is wrong.

First, a basic functional test (if needed) could/should be done in dtest.
Second, AFAIK @ShlomiBalalis has/had a task to include the mix in a longevity.
I would claim that All longevities should include the mix mode as we're probably going to have this mix for quite long time in the field till tablets will be fully production ready.
Third, the same can/should be done for all manager current tests, have the mix mode built in for all tests and no need for additional test.

The only question left is how to do it, one way could be the way you suggested here, but it'll require one to go and refactor too many tests.

@ShlomiBalalis what is the way you planned to do for longevity?

mikliapko:
> Third, the same can/should be done for all manager current tests, have the mix mode built in for all tests and no need for additional test.

Actually, I'd prefer to have both these options because of two reasons:
- we will still have old Enterprise releases (w/o tablets) in Prod we need to test;
- if the customer decides to disable tablets in new version, we will have the same configuration without tablets.

> The only question left is how to do it, one way could be the way you suggested here, but it'll require one to go and refactor too many tests.

Yes, I agree with you that in terms of applying this approach across the tests, it's not the optimal one. 
Ideally, we should have an opportunity to define type (mixed/not mixed) in configuration and go with that. 
But that's more significant intervention into SCT code with possible side-effects which I've tried to avoid here because of non-knowing the codebase and tests well enough. I suppose introducing of such approach should come from framework maintainers team at first.

Here we just wanted to have some tests to verify this configuration as soon as possible.

FYI @rayakurl

sdcm/cluster.py
```diff
@@ -2773,6 +2773,12 @@ def is_replacement_by_host_id_supported(self):
             return ComparableScyllaVersion(self.scylla_version) > '2022.3.0~dev'
         return ComparableScyllaVersion(self.scylla_version) > '5.2.0~dev'
 
+    @property
+    def is_cluster_have_system_auth_ks(self):
+        if self.is_enterprise:
+            return ComparableScyllaVersion(self.scylla_version) <= '2024.1'
```
fruch, line 2779: I think this can check raft topology, since it's coming with it

sdcm/utils/replication_strategy_utils.py
```diff
@@ -95,3 +96,17 @@ def __call__(self, **keyspaces: ReplicationStrategy) -> None:
         for keyspace, strategy in keyspaces.items():
             self._preserve_replication_strategy(keyspace)
             strategy.apply(self.node, keyspace)
+
+
+@dataclass
+class TabletsConfiguration:
```
fruch, line 102: Not sure this one directly related to replication

probably related to keyspaces, and we don't have a specific module for those yet

mgmt_cli_test.py
```diff
@@ -33,7 +33,7 @@
 from sdcm.mgmt.cli import ScyllaManagerTool
 from sdcm.mgmt.common import reconfigure_scylla_manager, get_persistent_snapshots
 from sdcm.remote import shell_script_cmd
-from sdcm.tester import ClusterTester
+from sdcm.tester import ClusterTester, TabletsConfiguration
```
fruch, line 36: Import from the right location

sdcm/utils/replication_strategy_utils.py
mikliapko: May we leave it without change until maintainers team implements it?
Or I can do it by myself, just give me an idea where to place this new module (to prevent further questions) and I'll put this class there.

soyacz: it is kind-of-related, I don't know better place for now.

mgmt_cli_test.py
mikliapko: Fixed

roydahan:
> > Third, the same can/should be done for all manager current tests, have the mix mode built in for all tests and no need for additional test.
> 
> Actually, I'd prefer to have both these options because of two reasons:
> 
> * we will still have old Enterprise releases (w/o tablets) in Prod we need to test;
> * if the customer decides to disable tablets in new version, we will have the same configuration without tablets.

We don't need to test both options, like you do in other features, if the feature is enabled you create this keyspace.
Otherwise, you don't.
Having another duplication of another full cluster sanity to check an option every cluster can enable is a waste.

> 
> > The only question left is how to do it, one way could be the way you suggested here, but it'll require one to go and refactor too many tests.
> 
> Yes, I agree with you that in terms of applying this approach across the tests, it's not the optimal one. Ideally, we should have an opportunity to define type (mixed/not mixed) in configuration and go with that. But that's more significant intervention into SCT code with possible side-effects which I've tried to avoid here because of non-knowing the codebase and tests well enough. I suppose introducing of such approach should come from framework maintainers team at first.
> 
> Here we just wanted to have some tests to verify this configuration as soon as possible.

It's not a configuration, it's a keyspace that is created in a certain way.

> 
> FYI @rayakurl

sdcm/cluster.py
fruch: Its raft topology changes, I'm not sure what exactly is_raft checks

mikliapko: Okay, I'll double-check it, perhaps, I used this check mistakenly

mikliapko:
> We don't need to test both options, like you do in other features, if the feature is enabled you create this keyspace. Otherwise, you don't. Having another duplication of another full cluster sanity to check an option every cluster can enable is a waste.

@roydahan Seems I got your point.

The problem with having the mix mode built in for all tests is that almost all Manager tests and helpers are designed to work with only one keyspace (in some cases not only technically, but logically too). Adding mixed mode as the default requires significant refactoring that affects not only Manager tests and helpers but SCT core (to not create mixed keyspace in test itself how it's done in this PR but during cluster configuration steps).

But I need to agree that this approach looks more logical and closer to real-life scenarios.

FYI @rayakurl

sdcm/cluster.py
Michal-Leszczynski: > I think this can check raft topology, since it's coming with it

Is that the case? Even when raft topology/schema was enabled in Scylla 5.4, system_auth was still there.
I thought that it's a change introduced in 6.0 that requires raft schema - but in 6.0 it's impossible to disable raft schema so it's more about the version than configuration.

Michal-Leszczynski (approved):

sdcm/cluster.py
mikliapko: There is one more observation in addition to Michal's reply.
If to query Scylla enabled_features for:
- Scylla 5.4 with enabled `consistent-topology-changes` - no `SUPPORTS_CONSISTENT_TOPOLOGY_CHANGES` among the features;
- Scylla 6.0 - `SUPPORTS_CONSISTENT_TOPOLOGY_CHANGES` presents there.

I believe we can use it as a definition point.

rayakurl:
> The code is ok, but I think the approach is wrong.
> 
> First, a basic functional test (if needed) could/should be done in dtest.

we are not sure about maintaining dtests. We talked about replacing those with the developers tests.
@Michal-Leszczynski , @karol-kokoszka - do you have such test? do you think we need to dtest to cover it?  

 Second, AFAIK @ShlomiBalalis has/had a task to include the mix in a longevity. 

@ShlomiBalalis not responding in the PRs or chat / mail for a few weeks now. We heard about those plans but didn't see the implementation. 

I would claim that All longevities should include the mix mode as we're probably going to have this mix for quite long time in the field till tablets will be fully production ready. Third, the same can/should be done for all manager current tests, have the mix mode built in for all tests and no need for additional test.

Since this was not even added to longevities yet,  not to talk about the longevites should have run like this for a while, we will not change all the manager jobs to run like this. We will have only this one as a POC.
When we will get an approval from core qa that this mixed mode was tested and is stable enough, we can change all manager jobs to work like this.
Until then, looks like the manager qa is the first team to test this configuration and we are doing it although it's not the best approach. 

> 
> The only question left is how to do it, one way could be the way you suggested here, but it'll require one to go and refactor too many tests.
> 
We don't intend to refactor all the tests. Once we will have a ack from core qa about the stability of this mixed mode, we can discuss a global solution.

> @ShlomiBalalis what is the way you planned to do for longevity?

CC @gmizrahi

sdcm/cluster.py
fruch: Yes auth_v2 goes hand by hand with raft topology changes.

And it can be disabled with force_gossip (not the exact name, I don't remember it)

So I still recommend checking scylla features when possible, it's a better option then counting on versions (since feature can be enabled/disabled in all bunch of ways)

mikliapko: Yes, I fixed it and retested here:
- Scylla 2024.1 - https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-sanity-test/21/
- Scylla 6.0 - https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-vnodes-tablets-cluster-test/11/

mikliapko:
@yarongilor I found out that you are a person we can contact regarding vnodes + tablets testing.
Could you please provide some details about your plans to include such kind of tests into SCT, what would be the approach, etc?

sdcm/utils/replication_strategy_utils.py
yarongilor: we probably should have something similar to `sdcm/utils/sstable/sstable_utils.py` like `sdcm/utils/keyspace/keyspace_utils.py`

mikliapko: One more approach, I can add a new module called `tablets` into `sdcm/utils/` and move this code there. Later, anyone who continue to work with tablets may extend or rework it depending on his needs.
@soyacz @fruch What do you think about it?

fruch: I'm o.k. with that

mikliapko: okay, will rework then

mikliapko: Fixed and retested here:
https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-vnodes-tablets-cluster-test/13/

mikliapko:
@scylladb/qa-maintainers Could you please take a look again after fixes?

fruch (approved):
LGTM

PR #7633, opened 2024-06-11
improvement(speedup): skip java and scylla-tools install

soyacz:
When using docker based loaders we don't need to install java and scylla-tools.

Skipping it to save some preparation time.

refs: https://github.com/scylladb/qa-tasks/issues/1650

### Testing

- [x] - provision tests
- [x] - test [with `use_prepared_loaders: false`](https://argus.scylladb.com/test/5bd8d2cf-5760-405e-8c70-fa0fd2bad5f0/runs?additionalRuns%5B%5D=1ce4dc25-3523-4565-9200-ccd42be0d5bb)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5006,36 +5006,8 @@ class BaseLoaderSet():
             self.log.info("Don't install anything because bare loaders requested")
             return
 
-        if self.params.get('client_encrypt'):
-            node.config_client_encrypt()
-
-        # Install java-8 for workaround hdrhistogram
-        if node.distro.is_rhel_like:
-            node.install_package(package_name='java-1.8.0-openjdk-devel')
-            node.remoter.sudo("ln -sf /usr/lib/jvm/java-1.8.0-openjdk-1.8.0.*/jre/bin/java* /etc/alternatives/java",
-                              verbose=True, ignore_status=True)
-        else:
-            node.install_package(package_name='openjdk-8-jre')
-            node.remoter.sudo("ln -sf /usr/lib/jvm/java-1.8.0-openjdk-*/bin/java* /etc/alternatives/java",
-                              verbose=True, ignore_status=True)
-
-        result = node.remoter.run('test -e ~/PREPARED-LOADER', ignore_status=True)
         node.remoter.sudo("bash -cxe \"echo '*\t\thard\tcore\t\tunlimited\n*\t\tsoft\tcore\t\tunlimited' "
                           ">> /etc/security/limits.d/20-coredump.conf\"")
-        if result.exit_status == 0:
-            self.log.debug('Skip loader setup for using a prepared AMI')
-            return
-
-        elif node.distro.is_debian10 or node.distro.is_debian11:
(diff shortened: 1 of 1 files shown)


fruch (changes requested):
There is a misunderstanding here, my guess you are running with scylla AMI as loader, and not one of our prepared loaders, hence why you get all of the installation of things you don't expect to.


If it's Scylla AMI one should use the bare loaders to not install anything on them, not even docker.

soyacz:
with this change loader setup time is ~75 seconds - less than db nodes.
[test](https://argus.scylladb.com/test/5bd8d2cf-5760-405e-8c70-fa0fd2bad5f0/runs?additionalRuns[]=1ce4dc25-3523-4565-9200-ccd42be0d5bb)

soyacz:
but true - I'm confused as we have 2 settings: `bare_loaders` (`false` everywhere except jepsen tests) and `use_prepared_loaders` and in the code we verify if loader is prepared by `test -e ~/PREPARED-LOADER`
In my opinion we should remove one or even both by making it automatic based on `PREPARED-LOADER` file

fruch:
> but true - I'm confused as we have 2 settings: `bare_loaders` (`false` everywhere except jepsen tests) and `use_prepared_loaders` and in the code we verify if loader is prepared by `test -e ~/PREPARED-LOADER`
> In my opinion we should remove one or even both by making it automatic based on `PREPARED-LOADER` file

Each one is for a different reason, one is when we identify it's our own created image, and one when it's based on images we didn't create.

Yes it might consolidate a bit, 
The idea in general was to have just docker based loaders, and ditch the rest of the installations and hacks.

It needs a bit more cleanup work, if we want to optimize it further for all of the use cases

soyacz:
@fruch let's have a talk about it - something is not working correctly anyway and we need to define what it is.

fruch:
> @fruch let's have a talk about it - something is not working correctly anyway and we need to define what it is.

We still need a talk ? Or I persuaded you it's not needed for now ?

soyacz:
> > @fruch let's have a talk about it - something is not working correctly anyway and we need to define what it is.
> 
> We still need a talk ? Or I persuaded you it's not needed for now ?

This PR appeared for a reason - to speed up loader setup as it's still not optimal. Despite we use our AMI's for loaders it somehow still reaches e.g. java installation. Maybe we should use `bare_loaders` true as default?

fruch:
> > > @fruch let's have a talk about it - something is not working correctly anyway and we need to define what it is.
> > 
> > We still need a talk ? Or I persuaded you it's not needed for now ?
> 
> This PR appeared for a reason - to speed up loader setup as it's still not optimal. Despite we use our AMI's for loaders it somehow still reaches e.g. java installation. Maybe we should use `bare_loaders` true as default?

The java installation was a work around to get HDR working, maybe it's not needed anymore. one would need to confirm that first.

why moving of docker installation is needed ?

fruch:
Scylla tools installation is needed for GCP and Azure

Since they don't have preloaded images, we shouldn't remove it without handling that

PR #7635, opened 2024-06-11
improvement(k8s): allow to configure K8S version via jenkins

vponomaryov:
### Testing

- [ ]

### Reminders


vponomaryov:
Tested here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-scylla-operator-basic-12h-eks/7

vars/perfRegressionParallelPipeline.groovy
```diff
@@ -81,6 +81,10 @@ def call(Map pipelineParams) {
                    description: 'Name of the test to run',
                    name: 'test_name')
 
+            string(defaultValue: "${pipelineParams.get('k8s_version', '')}",
```
soyacz, line 84: it's not enough. Need to also modify bash script in `Run ${sub_test}` section

vars/rollingOperatorUpgradePipeline.groovy
```diff
@@ -22,6 +22,10 @@ def call(Map pipelineParams) {
             string(defaultValue: "${pipelineParams.get('availability_zone', 'b')}",
                description: 'Availability zone',
                name: 'availability_zone')
+            string(defaultValue: "${pipelineParams.get('k8s_version', '')}",
```
zimnx, line 25: Both versions (base and target) should be configurable, as we may want to test upgrades between patch releases as well. Target can be optional with same behavior it's now.

vponomaryov: Target versioning will require more changes. It is not configurable as of now. Only implicit increment.

vars/perfRegressionParallelPipeline.groovy
vponomaryov: This pipeline calls `runSctTest` function which is updated as needed.
So, I don't see a problem. Update to the `runSctTest` function is the only one needed.

zimnx (approved):
thanks for updates

fruch (approved):
LGTM

PR #7637, opened 2024-06-12
Move disable auto upgrades to cloud init

soyacz:
Apt daily triggers disabling takes time and is quite late in the
proces. This prolongs setup due chance that triggers are fired before
installation of syslog-ng and other packages and makes apt lock to be
awaited. Also doing it with `remoter` class in multiple commands is not
most efficient way of doing it.

Moved disabling apt daily triggers to cloud-init step as the first in
the process.

closes: https://github.com/scylladb/qa-tasks/issues/1695

### Testing

- [x] - all provision tests
- [x] - [rocky9 based artifact test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-rocky9-test/3/)
- [x] - [debian11 based artifact test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-debian11-test/2/)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -63,6 +63,7 @@ from sdcm.mgmt.common import get_manager_repo_from_defaults, get_manager_scylla_
 from sdcm.prometheus import start_metrics_server, PrometheusAlertManagerListener, AlertSilencer
 from sdcm.log import SDCMAdapter
 from sdcm.provision.common.configuration_script import ConfigurationScriptBuilder
+from sdcm.provision.common.utils import disable_daily_apt_triggers
 from sdcm.provision.scylla_yaml import ScyllaYamlNodeAttrBuilder
 from sdcm.provision.scylla_yaml.certificate_builder import ScyllaYamlCertificateAttrBuilder
 from sdcm.provision.scylla_yaml.cluster_builder import ScyllaYamlClusterAttrBuilder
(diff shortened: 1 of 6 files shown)


sdcm/sct_provision/user_data_objects/sshd.py
```diff
@@ -26,6 +26,6 @@ def is_applicable(self) -> bool:
     @property
     def script_to_run(self) -> str:
         script = configure_sshd_script()
-        script = configure_ssh_accept_rsa()
+        script += configure_ssh_accept_rsa()
```
vponomaryov, line 29: How did Azure tests work without sshd configuration?

sdcm/provision/common/configuration_script.py
```diff
@@ -74,6 +75,7 @@ def _script_body(self) -> str:
         # 4. There is race condition between sct and boot script, disable ssh to mitigate it
         # 5. Make sure that whenever you use "cat <<EOF >>/file", make sure that EOF has no spaces in front of it
         script = ''
+        script += disable_daily_apt_triggers()
```
vponomaryov, line 78: Check for ubuntu or debian?

sdcm/sct_provision/user_data_objects/sshd.py
fruch: We are not using it that much, so we didn't notice any issues related to that, those are most configuration for handling better cases of reconnection and such

sdcm/provision/common/configuration_script.py
fruch: There is an internal check for apt-get existence, that would be enough

fruch: Keep in mind we create this script before a node is created (i.e. cloud-init)

sdcm/cluster.py
```diff
@@ -2945,24 +2945,6 @@ def set_seed_flag(self, is_seed):
         if hasattr(self.log, 'extra'):
             self.log.extra['prefix'] = str(self)
 
-    def disable_daily_triggered_services(self):
```
fruch, line 2948: We need to consider some cases where we don't have cloud-init working

docker backend as an example, we might count on the docker build process to disable all of those.

I did run into the cases in artifact tests when we don't have cloud-init installed or it's disabled on some distros. and for them we repeat the startup script again, so it's covered as well

sdcm/provision/common/utils.py
```diff
@@ -192,3 +192,15 @@ def install_syslogng_exporter():
     systemctl enable syslog_ng_exporter.service
     systemctl start syslog_ng_exporter.service
 """)
+
+
+def disable_daily_apt_triggers():
+    return dedent("""\
+    if apt-get --help >/dev/null 2>&1 ; then
```
fruch, line 199: We don't have the equivalent for CentOS/rhel/rocky ?

Other disablement of things that can move into this part ?

sdcm/provision/common/configuration_script.py
soyacz: yes, we don't know underlying OS at this stage.

sdcm/cluster.py
soyacz: but are docker images run these processes? I doubt as I think there's no even systemd there runnig.
I think this problem happens only in the cloud instances.

sdcm/provision/common/utils.py
soyacz: No, rhel based ones don't have it - for them we disable firewall only. I think moving it won't bring much improvement.

sdcm/cluster.py
fruch: I'm don't know if they do or don't, just suggesting we should validate it's not.

as for other distros, it might affect only artifact test, and we'll rarely see issues.
but it might also affect manager tests, that uses other distro for the monitoring for example.

it would be a bit of shame if we shove this part away, and it would start popping up across the board again.
that's the concern, and I wondering if we need to keep some fallback for those cases.

soyacz: I can apply similar trick like for syslog-ng - in case there's no marker file in dir it could run the old path. WDYT?

soyacz: actually, this is already applied if we use syslog-ng as stuff from `ConfigurationScriptBuilder` will be executed then anyway and skipped if already done.

sdcm/provision/common/utils.py
fruch: o.k. lets leave it as is

sdcm/cluster.py
fruch: well this trick of redoing the cloud-init, kind of masquerade which platform/distro we are using supports cloud-init and which are not.
so it's hard to tell which one we need to test when changing things around cloud-init...

so I don't know even which one exactly to point that we should test...

fruch (dismissed):
LGTM

sdcm/cluster.py
soyacz: cloud-init is always available for AWS, GCE and Azure I suppose. So only cases we don't have it is docker and bare metal cases.

fruch: it really depend on the image we are using, some images might not have cloud-init in them, or it's disabled by propose.
and we have quite vast of possibilities of images/distros we are using

fruch: one more related think, we are not using syslog-ng for artifact tests, but ssh, partly cause of those issues (at least that what I think)

soyacz: ok, I see the point. I fixed the code to retry this disabling in case if was not disabled. Please review again.

fruch (approved):
LGTM

PR #7651, opened 2024-06-13
feature(teardown-validators): parallel nemeses error events validator

dimakr:
In some cases (e.g. parallel nemeses) we don't want to fail the test based on an error event, but rather look for coredumps or critical level events.

The change introduces a teardown validator, which takes into account only pre-defined error events to make a decision if the test is failed.

The related task: https://github.com/scylladb/qa-tasks/issues/1671

### Testing
- [x] :green_circle: [longevity-5gb-1h-CorruptThenScrubMonkey + scrub + error events validators](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/sct-debug/53/)

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -250,6 +250,13 @@ teardown_validators:
     timeout: 1200
     keyspace: ''
     table: ''
+  test_error_events:
+    enabled: false
+    failing_events:
+      - event_class: DatabaseLogEvent
+        event_type: RUNTIME_ERROR
+        regex: '.*runtime_error.*'
+      - event_class: CoreDumpEvent
 
 kafka_backend: null
 kafka_connectors: []
--- a/sdcm/teardown_validators/base.py
+++ b/sdcm/teardown_validators/base.py
@@ -1,7 +1,11 @@
+from __future__ import annotations
 import logging
+from typing import TYPE_CHECKING
 
-from sdcm.cluster import BaseCluster
 from sdcm.sct_config import SCTConfiguration
+
+if TYPE_CHECKING:
+    from sdcm.tester import ClusterTester
 
 LOGGER = logging.getLogger(__name__)
 
@@ -9,9 +13,9 @@ LOGGER = logging.getLogger(__name__)
 class TeardownValidator:  # pylint: disable=too-few-public-methods
     validator_name = None  # must be defined by child classes
 
-    def __init__(self, params: SCTConfiguration, cluster: BaseCluster):
+    def __init__(self, params: SCTConfiguration, tester: ClusterTester):
         self.params = params
-        self.cluster = cluster
+        self.tester = tester
         self.configuration = params.get(f"teardown_validators.{self.validator_name}")
         self.is_enabled = self.configuration.get("enabled", False)
         if not self.is_enabled:
--- a/sdcm/teardown_validators/events.py
+++ b/sdcm/teardown_validators/events.py
@@ -1,0 +1,66 @@
+import json
+import logging
+import re
+from functools import reduce
+from typing import Optional, Union
+
+from sdcm.sct_events.events_device import get_events_main_device
+from sdcm.sct_events import Severity
+from sdcm.teardown_validators.base import TeardownValidator
+
+LOGGER = logging.getLogger(__name__)
+
+
+class FailingEventsFilter:
+    def __init__(
+            self,
+            event_class: str,
+            event_type: Optional[str] = None,
+            regex: Optional[Union[str, re.Pattern]] = None):
+        """Filter to match only events that qualify as valid reason to fail a test."""
+        self.event_class = event_class
+        self.event_type = event_type
+        self.regex = re.compile(regex) if isinstance(regex, str) else regex
+
+    def filter_events(self, events: list[dict]) -> list[dict]:
+        filtered_events = []
+        for event in events:
+            if self._matches(event):
+                filtered_events.append(event)
+        return filtered_events
+
+    def _matches(self, event: dict) -> bool:
+        match = (
+            (event['severity'] == Severity.ERROR.name) and
+            (event['base'] == self.event_class) and
+            (self.event_type is None or event['type'] == self.event_type) and
+            (self.regex is None or self.regex.search(event['line']))
+        )
+        return match
+
+
+class ErrorEventsValidator(TeardownValidator):  # pylint: disable=too-few-public-methods
+    """
+    A teardown validator that checks test events log for presence of specific error events and critical-level events.
+    If the test doesn't have any of the defined failing error events or critical events, its status
+    defaults to 'SUCCESS'.
+
+    Usage:
+        An instance of ErrorEventsValidator can be used during teardown of a test, where specific error events
+        should not be considered as test failures.
+        This is particularly useful in complex testing environments where multiple parallel operations may lead to
+        non-critical errors that should not directly influence the overall test outcome.
+    """
+    validator_name = 'test_error_events'
+
+    def validate(self):
+        LOGGER.info("Running validation of failing events for parallel nemeses")
+        filters = [FailingEventsFilter(**event) for event in self.configuration.get("failing_events")]
+        raw_events_log = get_events_main_device(_registry=self.tester.events_processes_registry).raw_events_log
+
+        with open(raw_events_log, encoding='utf-8') as events_file:
+            initial_events = (json.loads(line) for line in events_file)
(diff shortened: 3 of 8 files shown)


sdcm/teardown_validators/events.py
```diff
@@ -0,0 +1,27 @@
...
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ErrorEventsValidator(TeardownValidator):  # pylint: disable=too-few-public-methods
+    validator_name = 'error_events'
+
+    def validate(self):
+        # Do not run validator if there were no parallel nemeses
+        if len(self.tester.get_nemesis_class()) == 1:
```
fruch, line 16: Let's leave for SCT configuration to decide which case to enable it

No need to enforce it here

sdcm/teardown_validators/events.py
```diff
@@ -0,0 +1,27 @@
...
+import re
+
+from sdcm.sct_events import Severity
+from sdcm.sct_events.file_logger import get_events_grouped_by_category
+from sdcm.teardown_validators.base import TeardownValidator
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ErrorEventsValidator(TeardownValidator):  # pylint: disable=too-few-public-methods
```
fruch, line 11: We should document the validator somewhere

A docstring on the class should be a good start

Later some doc under docs/ would be also helpful, to point to those

fruch (commented):
The direction looks good 👍 

Few polishing comments

sdcm/teardown_validators/events.py
```diff
@@ -0,0 +1,27 @@
...
+        # Do not run validator if there were no parallel nemeses
+        if len(self.tester.get_nemesis_class()) == 1:
+            return
+
+        LOGGER.info("Running validation of ERROR events for parallel nemeses")
+        error_events = get_events_grouped_by_category(
+            _registry=self.tester.events_processes_registry)[Severity.ERROR.name]
+        failing_event_types = self.configuration.get("failing_events")
+        pattern = re.compile('|'.join(failing_event_types), re.IGNORECASE)
+        failing_events = [event for event in error_events if pattern.search(event)]
```
soyacz, line 24: I'd propose to iterate over `get_events_main_device(_registry=self.tester.events_processes_registry).raw_events_log` (I'm not sure of this line) and parse each line (the're jsons) and match by e.g. event type/subtype. It will give us more possibilities as pattern search is limited and error prone.

sdcm/teardown_validators/sstables.py
```diff
@@ -26,7 +27,7 @@ def _upload_corrupted_files(self, node: BaseNode, quarantine_log_lines):
             return "<No quarantine directories found>"
         s3_link = upload_remote_files_directly_to_s3(
             node.ssh_login_info, list(quarantine_dirs), s3_bucket=S3Storage.bucket_name,
-            s3_key=f"{self.cluster.uuid}/{node.name}-corrupted-sstables.tar.gz",
+            s3_key=f"{node.parent_cluster.cluster.uuid}/{node.name}-corrupted-sstables.tar.gz",
```
soyacz, line 30: shouldn't be `node.parent_cluster.uuid`?

defaults/test_default.yaml
```diff
@@ -247,6 +247,10 @@ teardown_validators:
     timeout: 1200
     keyspace: ''
     table: ''
+  error_events:
```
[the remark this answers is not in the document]
soyacz: I'd propose `test_status` - it is validating test status.

sdcm/teardown_validators/events.py
dimakr: added

defaults/test_default.yaml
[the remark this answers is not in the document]
dimakr: I stopped with `test_error_events`.
It describes the exact function of the new validator, i.e. validate errors from a test.

sdcm/tester.py
```diff
@@ -2873,14 +2873,17 @@ def clean_resources(self):
 
         self.destroy_credentials()
 
+    def iter_all_db_cluster(self):
```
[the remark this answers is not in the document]
dimakr: It was a part of another PR.
Changed back to list, we probably won't have large list of db clusters during a test)

sdcm/teardown_validators/events.py
dimakr: Good suggestion, thank you
After changing to the suggested approach it is much better to work with native python data structure

sdcm/teardown_validators/events.py
```diff
@@ -0,0 +1,37 @@
...
+        An instance of ErrorEventsValidator can be used during teardown of a test, where specific error events
+        should not be considered as test failures.
+        This is particularly useful in complex testing environments where multiple parallel operations may lead to
+        non-critical errors that should not directly influence the overall test outcome.
+    """
+    validator_name = 'test_error_events'
+
+    def validate(self):
+        LOGGER.info("Running validation of ERROR events for parallel nemeses")
+        failing_event_types = [et.lower() for et in self.configuration.get("failing_events")]
```
soyacz, line 27: are you sure it works when lowering event name? 

Would be good to have it unit tested.

dimakr: Right. Regex search approach was performed with ignorecase modifier, but after switching to parsing events to json we need to take care of this ourselves.
Fixed this and added `unit_tests/test_teardown_validator.py` unit tests. Their results are in https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-7651/4/

sdcm/teardown_validators/base.py
```diff
@@ -9,9 +8,9 @@
 class TeardownValidator:  # pylint: disable=too-few-public-methods
     validator_name = None  # must be defined by child classes
 
-    def __init__(self, params: SCTConfiguration, cluster: BaseCluster):
+    def __init__(self, params: SCTConfiguration, tester: 'sdcm.tester.ClusterTester'):
```
fruch, line 11:
```
from typing import TYPE_CHECKING
if TYPE_CHECKING:
   from sdcm....
```

sdcm/teardown_validators/events.py
```diff
@@ -0,0 +1,37 @@
...
+        raw_events_log = get_events_main_device(_registry=self.tester.events_processes_registry).raw_events_log
+
+        with open(raw_events_log, encoding='utf-8') as events_file:
+            events = (json.loads(line) for line in events_file)
+            failing_error_events = [
+                event for event in events
+                if event['severity'] == Severity.ERROR.name and event['base'].lower() in failing_event_types]
```
fruch, line 34: we probably would want it to work similar to how filtering it working:
https://github.com/scylladb/scylla-cluster-tests/blob/60cddb41a438512ce1abab6006966b33bbaf2bd8/sdcm/sct_events/setup.py#L129

dimakr: Added a similar filter functionality into sdcm/teardown_validators/events.py

fruch (approved):
LGTM

fruch:
@dimakr 

ruff isn't happy 
```
12:31:38  sdcm/teardown_validators/sstables.py:77:20: BLE001 Do not catch blind exception: `Exception`
```

and also lets apply it on `longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis.yaml` where it's targeted

dimakr:
> lets apply it on `longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis.yaml` where it's targeted

@fruch
do we want to let only critical and coredump events to fail that scenario?
Or there are some error events which we should never filter out? And have them fail the test, if occur

fruch:
> > lets apply it on `longevity-2TB-48h-authorization-and-tls-ssl-1dis-2nondis-nemesis.yaml` where it's targeted
> 
> @fruch do we want to let only critical and coredump events to fail that scenario? Or there are some error events which we should never filter out? And have them fail the test, if occur

for now yes.

you can see the last time it had run in 2024.1:

https://argus.scylladb.com/test/215eaa3d-464c-4935-9e89-009361fbac68/runs?additionalRuns[]=66ea2c77-343c-461f-b722-3802090fc8ec

it had lots of error, and even on coredump, but we still decided it's o.k. (after cross checking the issues)
with this change, we might have just crossed checked the 1 coredump, and waiver the rest of the error events , right away.

it's a 2 days case, which back then took 8h to setup and fill 2Tb, hopefully it would be used in the next enterprise release.

fruch (approved):
LGTM

PR #7655, opened 2024-06-16
fix(tablets): switch tablets-enabled tests folder to disabled

yarongilor:
tablets feature is not experimental anymore and is defaulted
        to be enabled. So tests of previous tablets-enabled folder are switched to be
        "disabled", in case we would like to test a disabled mode.
        the label of "tablets_phase_1_supported" is removed as well.
        refs: https://github.com/scylladb/scylladb/pull/18898
refs: https://github.com/scylladb/scylladb/pull/18898

### Testing

- [ ]

### Reminders


bhalevy:
But tablets aren't enabled by default for dtest.
It still requires the --tablets pytest option.

data_dir/nemesis.yml
```diff
@@ -16,7 +16,6 @@
   - run_with_gemini = False
   - limited = True
   - topology_changes = True
-  - tablets_phase_1_supported = False
```
roydahan, line 19: What does it mean for nemesis that aren't supported with "Tablets"?
Will they fail?
In this case, it can't go like that and either each nemesis will have a special handling for that, or you get and approval from whoever is to approve it that we won't run all these nemesis anymore and we keep this filter but set it to be the default for all longevities.

fruch:
> But tablets aren't enabled by default for dtest. It still requires the --tablets pytest option.

this is SCT, what dtest has todo with it ?

bhalevy:
> > But tablets aren't enabled by default for dtest. It still requires the --tablets pytest option.
> 
> this is SCT, what dtest has todo with it ?

Sorry, I was thinking about the dtests in the tablets folder in jenkins

data_dir/nemesis.yml
fruch: @roydahan he's removing this label that @ShlomiBalalis introduced.

@yarongilor don't really know which one works and which one doesn't 

I'm o.k. with the assumption they are all working, we didn't limited it in any case, just in few shlomo was testing prior to making tablets default.

it's water under the bridge now

yarongilor: @roydahan , as @fruch mentioned, it is already "too late" (after tablets are enabled by default).
See, for example - https://github.com/scylladb/scylla-cluster-tests/issues/7671

We do roughly know what isn't supported with tablets: CDC, LWT, resharding and few more.
So there are few fixes to add. and there are few "feature specific" longevities to adjust to temporarily run without tablets.

roydahan (approved):

fruch (approved):
LGTM

PR #7656, opened 2024-06-16
fix(db_log_reader): take build-id from logs

fruch:
so far we picked the build-id from the current running scylla executable in some cases this fails, since the backtrace we are reading might be from diffrent version of scylla.

instead we start reading it from the same log file we are reading the backtrace from, so it would be in sync and remove the room for getting the wrong one

Ref: scylladb/qa-tasks#951

### Testing

- [x] provision test confirm it's working

### Reminders

### DIFF
--- a/sdcm/db_log_reader.py
+++ b/sdcm/db_log_reader.py
@@ -54,6 +54,7 @@ class DbLogReader(Process):
         '] storage_proxy - Exception when communicating with',
     ]
     # pylint: disable=too-many-arguments
+    BUILD_ID_REGEX = re.compile(r'build-id\s(.*?)\sstarting\s\.\.\.')
 
     def __init__(self,
                  system_log: str,
@@ -75,6 +76,7 @@ class DbLogReader(Process):
         self._last_log_position = 0
         self._remoter = remoter
         self._skipped_end_line = 0
+        self._build_id = None
         super().__init__(name=self.__class__.__name__, daemon=True)
 
     @cached_property
@@ -124,6 +126,10 @@ class DbLogReader(Process):
 
                     if json_log:
                         continue
+
+                    if match := self.BUILD_ID_REGEX.search(line):
+                        self._build_id = match.groups()[0]
+                        LOGGER.debug("Found build-id: %s", self._build_id)
 
                     match = BACKTRACE_RE.search(line)
                     one_line_backtrace = []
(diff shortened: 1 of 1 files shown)


sdcm/db_log_reader.py
```diff
@@ -125,6 +127,9 @@ def _read_and_publish_events(self) -> None:
                     if json_log:
                         continue
 
+                    if match := self.BUILD_ID_REGEX.search(line):
```
soyacz, line 130: we could do some perf optimizations as this will be executed for each line:
1. limit line length (not to parse long lines)
2. Optimize regex to be faster, e.g. `re.compile(r'build-id ([a-f0-9]{40}) starting')` (is ignorecase required?)

soyacz:
btw. I wonder if such real-time log parsing couldn't be used to track cluster state (decommisions, repairs, raft states) and pack it into simple object that could be used to track states - instead what we do currently in many places on it's own (e.g. in abort decommission nemesis).

sdcm/db_log_reader.py
fruch: > we could do some perf optimizations as this will be executed for each line:
> 
> 1. limit line length (not to parse long lines)

we already throw out 5K lines

> 2. Optimize regex to be faster, e.g. `re.compile(r'build-id ([a-f0-9]{40}) starting')` (is ignorecase required?)

ignore case isn't required


```python
import re
import timeit

BUILD_ID_REGEX = re.compile(r'build-id\s(.*?)\sstarting\s\.\.\.', flags=re.IGNORECASE)

BUILD_ID_REGEX_1 = re.compile(r'build-id\s(.*?)\sstarting\s\.\.\.')

BUILD_ID_REGEX_2 = re.compile(r'build-id\s([a-f0-9]{40})\sstarting')

l = "abcdefghij1234567890" * 256
res = timeit.timeit('BUILD_ID_REGEX.search(l)', 'from __main__ import BUILD_ID_REGEX, l')
print(res)


res = timeit.timeit('BUILD_ID_REGEX_1.search(l)', 'from __main__ import BUILD_ID_REGEX_1, l')
print(res)

res = timeit.timeit('BUILD_ID_REGEX_2.search(l)', 'from __main__ import BUILD_ID_REGEX_2, l')
print(res)
```


```
🟢 ❯ python test_123.py
26.814458078995813
1.6338670099939918
1.613332103996072
```

fruch:
> btw. I wonder if such real-time log parsing couldn't be used to track cluster state (decommisions, repairs, raft states) and pack it into simple object that could be used to track states - instead what we do currently in many places on it's own (e.g. in abort decommission nemesis).

it would suffer from the same problems as the mechanism we have today, if there a delay, it won't help.

and that's the most problematic issue we have.

we should replace things with API calls as much as possible, I rather do million https calls then trust the logging system for test logic.

sdcm/db_log_reader.py
fruch: remove the `re.IGNORECASE`

the rest is a bit less meaningful

fruch:
provision test shows it's working:
```
< t:2024-06-17 13:59:39,148 f:db_log_reader.py l:132  c:sdcm.db_log_reader   p:DEBUG > Found build-id: 19149958503a961b9fc3b558e9e6cc22d3994169
```

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

PR #7670, opened 2024-06-17
test(resharding): skip when tablets are enabled

yarongilor:
resharding is currently not supported with tablets,
	so nemesis or validations should be skipped.
	refs: https://github.com/scylladb/scylladb/issues/16739
refs: https://github.com/scylladb/scylladb/issues/16739
### Testing

- [ ]

### Reminders


bhalevy:
tablets does not support resharding per-se the same way its done with vnodes, but it still supports a scenario where number of shards of a node is increased.
Unlike vnodes, where data is resharded synchronously when the node is reatarted, the node starts normally, resharding only the vnode keyspaces, but tablets-enabled tables start as is, and after the node starts, it is expected that the load balancer will detect imbalance across shards  since the newly added shards own no tablets, and then the load balancer should migrate tablet replicas to the new shards to restore balance across shards.

This should still be tested...
Can we incorporate that instead of just skipping the nemesis when increasing the shard count?
(It should be benign with tablets)

fruch:
> tablets does not support resharding per-se the same way its done with vnodes, but it still supports a scenario where number of shards of a node is increased.
> Unlike vnodes, where data is resharded synchronously when the node is reatarted, the node starts normally, resharding only the vnode keyspaces, but tablets-enabled tables start as is, and after the node starts, it is expected that the load balancer will detect imbalance across shards  since the newly added shards own no tablets, and then the load balancer should migrate tablet replicas to the new shards to restore balance across shards.
> 
> This should still be tested...
> Can we incorporate that instead of just skipping the nemesis when increasing the shard count?
> (It should be benign with tablets)

There are 3 different cases of resharding mentioned in this PR

Only the k8s case is changing the number of CPUs/shards

soyacz:
> Only the k8s case is changing the number of CPUs/shards

And for that case we reduce the number of shards, so I believe it's also not supported.

bhalevy:
> > Only the k8s case is changing the number of CPUs/shards
> 
> And for that case we reduce the number of shards, so I believe it's also not supported.

Yes, reducing the number of shards isn't supported with tablets

bhalevy:
> > > > Only the k8s case is changing the number of CPUs/shards
> > > 
> > > 
> > > And for that case we reduce the number of shards, so I believe it's also not supported.
> > 
> > 
> > Yes, reducing the number of shards isn't supported with tablets
> 
> @bhalevy , @fruch , @soyacz , It should be clear what's the impact of this non supported. If the approach is now only skipping resharding-validation, will the nemesis scenario complete ok? or will scylla generate an error for that?
> 
> I can run a job to test what really happens of course, or see how dtest works with such a scenario.

Let's do that.  I'm not exactly sure what is the expected behavior when reducing the number of shards with tablets (when some tablets are allocated to shards that do no longer exist).
Cc @tgrabiec

tgrabiec:
> Let's do that. I'm not exactly sure what is the expected behavior when reducing the number of shards with tablets (when some tablets are allocated to shards that do no longer exist).
> Cc @tgrabiec

Currently we refuse to boot with an error.

yarongilor:
> > Let's do that. I'm not exactly sure what is the expected behavior when reducing the number of shards with tablets (when some tablets are allocated to shards that do no longer exist).
> > Cc @tgrabiec
> 
> Currently we refuse to boot with an error.

@bhalevy  the [test](https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/eks/job/longevity-scylla-operator-3h-eks-nodetool-flush-and-reshard-test/6/) failed with:
```
10:35:00,256 [shard  0:main] init - Startup failed: std::runtime_error (Detected a tablet with invalid replica shard, reducing shard count with tablet-enabled tables is not yet supported. Replace the node instead.)
```
so we have to skip this part of nemesis.

yarongilor:
@soyacz , @fruch , i think this PR could be merged.

sdcm/cluster.py
```diff
@@ -2555,21 +2556,24 @@ def restart_node_with_resharding(self, murmur3_partitioner_ignore_msb_bits: int
...
-        if resharding_started:
-            self.log.debug(f'Resharding has been started successfully '
-                           f'(murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits})')
-        else:
-            raise Exception(f'Resharding has not been started '
-                            f'(murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits}) '
-                            'Check the log for the details')
+        # If tablets in use, skipping resharding validation since it doesn't work the same as vnodes
+        with self.parent_cluster.cql_connection_patient(self) as session:
+            if not is_tablets_feature_enabled(session=session):
```
fruch, line 2561: please add the SkipPerIssue, so it would point to the relevant issue.

also for docs, and also so it would start working when resolved

yarongilor: @fruch fixed.

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #7688, opened 2024-06-18
feature(coredumps): support zstd compressed coredumps

fruch:
if we have coredumps compressed with zstd, we shouldn't compress them again also we should write the proper tool in the download instructions

Ref: scylladb/scylla-machine-image#462
Ref: scylladb/scylladb#18854

### Testing

- [ ]

### Reminders


soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

p.s. 

> if we have coredumps compressed with zstd, we should compress them again also we should write the proper tool in the download instructions

we should NOT lose `not` word in statements like this.

fruch:
> LGTM
> 
> p.s. 
> 
> > if we have coredumps compressed with zstd, we should compress them again also we should write the proper tool in the download instructions
> 
> we should NOT lose `not` word in statements like this.

Maybe we'll leave as an exercise for the reader ? 😉

fruch:
> > LGTM
> > p.s.
> > > if we have coredumps compressed with zstd, we should compress them again also we should write the proper tool in the download instructions
> > 
> > 
> > we should NOT lose `not` word in statements like this.
> 
> Maybe we'll leave as an exercise for the reader ? 😉

fixed the commit message and the PR description

Issue #7255, opened 2024-03-07
coredump files get corrupted


bhalevy:
As seen in https://github.com/scylladb/scylla-enterprise/issues/3968
Coredump files that were collected could not be opened by gdb.

The symptom looks like:
```
Failed to read a valid object file image from memory.
```
And additional printouts that are probably unrelated:
```
Trying host libthread_db library: search-path /opt/scylladb/libreloc/libthread_db.so.1.
open failed: No such file or directory.
thread_db_load_search returning 0
```

@fruch theorized that since scylla was observed to crash in quick succession maybe new coredumps caused the previous coredumps that get truncated by coredumpctl.

We need to make sure in testing that if scylla crashed, then the scylla service wouldn't start until the previous coredump was collected (and archived on s3/gc? can we do that?).

@avikivity is that something we can/should do on production machine images?

fruch:
here's one example of such occurrence: 
```
# SCT starts compressing 
2024-02-25T22:13:05.713+00:00 multi-dc-rackaware-tablets--db-node-7a99b942-5   !NOTICE | sudo[17453]: scyllaadm : PWD=/home/scyllaadm ; USER=root ; COMMAND=/usr/bin/pigz --fast --keep /var/lib/systemd/coredump/core.scylla.112.7ca8b11c592f4c37b5e4e54b56772b37.17368.1708899083000000
...
# systemd-coredump decides this file should be remove (probably cause of no space, or any other calculation it might do)
2024-02-25T22:13:15.213+00:00 multi-dc-rackaware-tablets--db-node-7a99b942-5     !INFO | systemd-coredump[17550]: Removed old coredump core.scylla.112.7ca8b11c592f4c37b5e4e54b56772b37.17368.1708899083000000.
...
# SCT finishes the compression
2024-02-25T22:14:23.213+00:00 multi-dc-rackaware-tablets--db-node-7a99b942-5     !INFO | sudo[17453]: pam_unix(sudo:session): session closed for user root
```


the default values for `coredump.conf` are probably not good enough for us in some cases
https://manpages.debian.org/testing/systemd-coredump/coredump.conf.5.en.html

I would recommend running it with `ExternalSizeMax=infinity` as a start

roydahan:
Maybe a possible workaround for now to be able to upload the coredump is to change Coredump event to be CRITICAL event, which will stop the test and hopefully prevent the next crash that probably corrupts the coredump upload.
(It will also leave the machine up and running).

[...]

PR #7689, opened 2024-06-18
feature(coredump): use hardlinks to prevert from clearing coredump

fruch:
this change is for creating hardlink on any found core and compressing uploading using the hardlink, so it won't be cleared by systemd-coredump in the midst of SCT processing of the coredump.

hardlinks are created in `/var/lib/scylla/hardlinks` so it would share the same physical device as the coredumps directory, but still won't be scan for deletion by systemd-coredump.

Fixes: #7255

### Testing

- [x] 🟢  https://jenkins.scylladb.com/job/scylla-enterprise/job/reproducers/job/scylla-master-perf-regression-latency-650gb-grow-shrink/9/ [argus](https://argus.scylladb.com/test/a9ff9a8c-6b1d-43f4-afa2-3e31cdee2d9a/runs?additionalRuns[]=1039e962-4edd-482e-9a20-f937ddd865be) coredump were intact 

### Reminders


soyacz:
for testing purposes, I propose to create temporary nemesis `CreateCoredumpNemesis` that will execute:
```
sudo apt install -y gdb
cd /var/lib/scylla/coredump/
sudo gcore $(pidof scylla)
```

fruch:
> for testing purposes, I propose to create temporary nemesis `CreateCoredumpNemesis` that will execute:
> ```
> sudo apt install -y gdb
> cd /var/lib/scylla/coredump/
> sudo gcore $(pidof scylla)
> ```

For this one it's not enough, we need a case of scylla crashing again and again during boot, to validate this really solves the issue

Meanwhile I'm using the exact same KMS related failure you run into to try validating 

But we could do the nemesis anyhow, but will need to add a test way to have nemesis which are only for testing, and excluded from other nemesis groups and selections

fruch:
hard link across mount point aren't exactly working (even if it's attached to the same device)
```
< t:2024-06-21 07:09:09,112 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.2.71>: ln: failed to create hard link '/var/lib/scylla/hardlinks/core.scylla.112.d711354b6d3842ac9ae19caaa8d437e1.6747.1718953646000000' => '/var/lib/systemd/coredump/core.scylla.112.d711354b6d3842ac9ae19caaa8d437e1.6747.1718953646000000': Invalid cross-device link
```

I'll try it on the same folder, hopefully systemd-coredump doesn't do it recursively

michoecho:
> I'll try it on the same folder, hopefully systemd-coredump doesn't do it recursively

If I'm reading the source code correctly, it shouldn't be doing it recursively.

fruch:
@michoecho 

seems like kind of worked

```
#0  0x00007f8ddc25f884 in __pthread_kill_implementation () from /opt/scylladb/libreloc/libc.so.6
#1  0x00007f8ddc20eafe in raise () from /opt/scylladb/libreloc/libc.so.6
#2  0x00007f8ddc1f787f in abort () from /opt/scylladb/libreloc/libc.so.6
#3  0x0000000001b69aa8 in seastar::memory::on_allocation_failure (size=131072) at ./seastar/src/core/memory.cc:2167
#4  0x00000000017bcf75 in seastar::memory::finish_allocation (ptr=<optimized out>, size=131072) at ./seastar/src/core/memory.cc:1628
#5  seastar::memory::allocate_slowpath (size=<optimized out>) at ./seastar/src/core/memory.cc:1690
#6  0x0000000002329918 in seastar::memory::allocate (size=131072) at ./seastar/src/core/memory.cc:1707
#7  malloc (n=<optimized out>) at ./seastar/src/core/memory.cc:2216
#8  utils::chunked_vector<std::pair<sstables::metadata_type, unsigned int>, 131072ul>::new_chunk (this=0x604004c9f650, n=16384) at ./utils/chunked_vector.hh:406
#9  utils::chunked_vector<std::pair<sstables::metadata_type, unsigned int>, 131072ul>::make_room (this=0x604004c9f650,
    n=<error reading variable: Cannot access memory at address 0x89649277>, stop_after_one=false) at ./utils/chunked_vector.hh:449
#10 utils::chunked_vector<std::pair<sstables::metadata_type, unsigned int>, 131072ul>::reserve (this=0x604004c9f650,
    n=<error reading variable: Cannot access memory at address 0x89649277>) at ./utils/chunked_vector.hh:178
#11 sstables::parse<unsigned int, std::pair<sstables::metadata_type, unsigned int> > (s=..., v=<optimized out>, in=..., arr=...) at sstables/sstables.cc:355
#12 0x0000000002db9a47 in std::__n4861::coroutine_handle<seastar::internal::coroutine_traits_base<void>::promise_type>::resume (this=<optimized out>)
    at /usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../include/c++/13/coroutine:240
#13 seastar::internal::coroutine_traits_base<void>::promise_type::run_and_dispose (this=0x0) at ./seastar/include/seastar/core/coroutine.hh:122
#14 seastar::reactor::run_tasks (this=0x6040000b6000, tq=...) at ./seastar/src/core/reactor.cc:2694
#15 seastar::reactor::run_some_tasks (this=0x6040000b6000) at ./seastar/src/core/reactor.cc:3157
#16 0x00000000028b2598 in seastar::reactor::do_run (this=0x6040000b6000) at ./seastar/src/core/reactor.cc:3325
#17 0x000000000640a5b6 in seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_0::operator()() const (this=0x60000004bb80)
    at ./seastar/src/core/reactor.cc:4568
#18 std::__invoke_impl<void, seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_0&>(std::__invoke_other, seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_0&) (__f=...) at /usr/bin/../lib/gcc/x86_64-redhat-linux/13/../../../../include/c++/13/bits/invoke.h:61
#19 std::__invoke_r<void, seastar::smp::configure(seastar::smp_options const&, seastar::reactor_options const&)::$_0&>(seastar::smp::configure(seastar::smp_options const&, seastar::re--Type <RET> for more, q to quit, c to continue without paging--
```

can you download some, and confirm too ?

michoecho:
> can you download some, and confirm too ?

I downloaded some cores and looked at the logs. The patch seems to be working as expected.

By the way, there is more room for improvement — if Scylla were to restart fast enough, SCT might not even be fast enough to hardlink the core file before it is removed. It could be prevented by running the coredump scraping thread directly on the server, and watching the coredump directory with inotify. But this race probably isn't a problem in practice; at least according to this test, there seems to be quite a lot of time (~30s) between the creation of the coredump and the restart of Scylla. So we can return to this particular problem only if it ever turns out to be a problem in practice.

fruch:
> > can you download some, and confirm too ?
> 
> I downloaded some cores and looked at the logs. The patch seems to be working as expected.
> 
> By the way, there is more room for improvement — if Scylla were to restart fast enough, SCT might not even be fast enough to hardlink the core file before it is removed. It could be prevented by running the coredump scraping thread directly on the server, and watching the coredump directory with inotify. But this race probably isn't a problem in practice; at least according to this test, there seems to be quite a lot of time (~30s) between the creation of the coredump and the restart of Scylla. So we can return to this particular problem only if it ever turns out to be a problem in practice.

I had a dream of sending it straight to s3 from the node, but it involves lots of changes.

Anyhow as you said we'll follow it up, if this workaround isn't won't cut it, (or it if would break)

fruch:
> 
> > 
> > I had a dream of sending it straight to s3 from the node, but it involves lots of changes.
> >
> Once I implemented way to do that with `sdcm.utils.s3_remote_uploader.upload_remote_files_directly_to_s3`
> stil, uses sct-runner as a proxy, but skips saving anything on it.
> Would require some refactor to handle compression better, but it's not hard.

I was thinking about straight out of the node, with zero SCT intervention 

Now that I think of it should be to s3, but to GCP, and we can use the exact same urls we use today, and the uploads are public, so you don't need to handle complex credentials (we let users upload cores there)

fruch:
@scylladb/qa-maintainers if someone can give one final look ?

sdcm/coredump.py
```diff
@@ -308,6 +316,16 @@ def systemd_version(self):
...
 
+    @contextmanager
+    def hard_link_corefile(self, corefile):
+        hard_links_path = Path(corefile).parent / 'hardlinks'
+        link_path = hard_links_path / Path(corefile).name
+        self.node.remoter.sudo(f'mkdir -p {hard_links_path}', ignore_status=True)
+        self.log.debug(f'doing: ln {corefile} {link_path}')
+        self.node.remoter.sudo(f'ln {corefile} {link_path}', ignore_status=True)
+        yield link_path
+        self.node.remoter.sudo(f'rm -f {link_path}', ignore_status=True)
```
soyacz, line 327: I'm not sure if we should ignore statuses her. What is the reason?

soyacz (approved):
One question, otherwise LGTM

sdcm/coredump.py
vponomaryov: Best effort for a secondary goal.

vponomaryov (approved):
LGTM

sdcm/coredump.py
fruch: worse case we'll be filling the disk, and we get to that, we are in a shity situattion already :)

PR #7691, opened 2024-06-19
fix(tls-certs): fix usage of certs in mgmt_cli test and improve naming of tls artifacts 

dimakr:
After TLS certificates preparation was reworked in SCT, individual TLS certificates are created and distributed for loader and cluster nodes.

The change adds the same logic for creating and distributing certificate/key to a manager node (which is usually a monitor node in the mgmt_cli tests).
Also certificate/key naming in SCT is changed to avoid redefining them in different modules - the names are defined as constants in one place now and re-used where necessary.

### Testing
- [x] :green_circle: [ubuntu22-sanity-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/ubuntu22-sanity-test/1/)
- [x] :green_circle: [regression longevity test with server+client encryptions enabled](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/certs-rework-tests/26/)

### DIFF
--- a/sdcm/mgmt/cli.py
+++ b/sdcm/mgmt/cli.py
@@ -16,6 +16,7 @@ import json
 import time
 import logging
 import datetime
+from pathlib import Path
 from re import findall
 from textwrap import dedent
 from statistics import mean
@@ -24,11 +25,12 @@ from distutils.version import LooseVersion
 
 import requests
 from invoke.exceptions import Failure as InvokeFailure
-from sdcm.remote.libssh2_client.exceptions import Failure as Libssh2Failure
 
+from sdcm.remote.libssh2_client.exceptions import Failure as Libssh2Failure
 from sdcm import wait
 from sdcm.mgmt.common import \
     TaskStatus, ScyllaManagerError, HostStatus, HostSsl, HostRestStatus, duration_to_timedelta, DEFAULT_TASK_TIMEOUT
+from sdcm.provision.helpers.certificate import TLSAssets
 from sdcm.utils.distro import Distro
 from sdcm.wait import WaitForTimeoutError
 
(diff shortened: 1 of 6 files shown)


mikliapko:
@dimakr 
Could you please check if the initial implementation got into manager-3.3 branch? 
If so, we need to backport this change there.

dimakr:
> @dimakr Could you please check if the initial implementation got into manager-3.3 branch? If so, we need to backport this change there.

@mikliapko 
The initial implementation of certs handling is not backported to any branches and seems like there is no plans to do it.
So we should be safe here.

sdcm/mgmt/cli.py
```diff
@@ -37,8 +37,8 @@
 STATUS_DONE = 'done'
 STATUS_ERROR = 'error'
 SSL_CONF_DIR = '/tmp/ssl_conf'
-SSL_USER_CERT_FILE = SSL_CONF_DIR + '/db.crt'
-SSL_USER_KEY_FILE = SSL_CONF_DIR + '/db.key'
+SSL_USER_CERT_FILE = SSL_CONF_DIR + '/test.crt'
```
fruch, line 40: should be imported from the right place ? to avoid defining multiple times ?

fruch: i.e. form the `sdcm.provision.helpers.certificates` ?

dimakr: Performed tidying up for the cert names here and in a few other places.

fruch:
nitpick: maybe rettitle it, since now it's mostly of the tidying up, while fixing an issue.

fruch (approved):
🟢 LGTM (that circle looks familiar 🤣)

mikliapko:
I suppose the commit message may be retitled as well

mikliapko:
Since it fixes the issue with manager sanity and test is green now, LGTM

mikliapko (approved):

mikliapko:
@dimakr 
Strange but ubuntu20-sanity-test for Manager fails with the issue that has been fixed here for ubuntu 22.
https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/ubuntu20-sanity-test/599/

Is it possible to have any dependency on specific distro in the code you've implemented here?

Issue #7162, opened 2024-01-27
SCYLLA_VERSION_RE.match(scylla_version) should have a fallback if it doesn't find the scylla version and not to be run when using prepared loaders


avikivity:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/avi/job/avi-regression-latency-650gb-grow-shrink/8/console

It takes about 25 minutes to do some setup, then decide it's not working.

```
22:08:56  Started by user [Avi Kivity](https://jenkins.scylladb.com/user/avikivity)
```

```
22:43:54  < t:2024-01-27 20:43:53,792 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-01-27 20:43:53.791: (InfoEvent Severity.NORMAL) period_type=not-set event_id=565d49e0-7939-490f-88ae-d56e99e7ef13: message=TEST_END
```

roydahan:
First, I created a separate "epic" task to optimize test setup time (#7242 ).

In this case, the issue is that SCT couldn't detect the provided scylla version, probably it was a private build with non-matching pattern of scylla version.

The expectation should be a fallback to choose a loader from master.

File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/version_utils.py", line 342, in get_docker_image_by_version
22:52:46  short_scylla_version = SCYLLA_VERSION_RE.match(scylla_version)

Another optimization we should do is not doing this docker loading at all when using "use_prepared_loaders: true"

fruch:
we don't have the function of selecting a docker image, so we can close this one

Issue #7160, opened 2024-01-27
update_db_packages - Installation failure should be raised as a critical event (which will stop the test)


avikivity:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/avi/job/avi-regression-latency-650gb-grow-shrink/8/console

```
22:37:24  < t:2024-01-27 20:37:18,252 f:cluster.py      l:2118 c:sdcm.cluster_aws     p:WARNING > Node perf-latency-grow-shrink-ubuntu-db-node-42855fcd-1 [54.247.210.237 | 10.4.3.94] (seed: True): All attempts to get ScyllaDB version failed. Looks like there is no ScyllaDB installed.
22:39:21  < t:2024-01-27 20:39:07,508 f:cluster.py      l:2118 c:sdcm.cluster_aws     p:WARNING > Node perf-latency-grow-shrink-ubuntu-db-node-42855fcd-1 [54.247.210.237 | 10.4.3.94] (seed: True): All attempts to get ScyllaDB version failed. Looks like there is no ScyllaDB installed.
22:39:21  < t:2024-01-27 20:39:19,800 f:cluster.py      l:2118 c:sdcm.cluster_aws     p:WARNING > Node perf-latency-grow-shrink-ubuntu-db-node-42855fcd-1 [54.247.210.237 | 10.4.3.94] (seed: True): All attempts to get ScyllaDB version failed. Looks like there is no ScyllaDB installed.
```

but before it says

```
[...]

PR #7692, opened 2024-06-19
fix(update_db_packages): fix issues with hanging threads during update

dimakr:
update_db_packages procedure executes updating of scylla packages in threads for each node and uses sdcm.utils.decorators.retrying mechanism for retries. Through this decorator it is allowed for UnexpectedExit exception to occur up to the specified number of times, when installing packages, then the UnexpectedExit exception should be reraised.

For some reason reraise doesn't return control back to the caller if update function is running in threading.Thread threads. So if for some reason a thread with packages update is continuously failing and all retries are exhausted, the thread hangs.

The change switches from threading.Thread to concurrent.futures.ThreadPoolExecutor for paralleling packages update on nodes. This approach doesn't have issues with reraising exceptions in threads.

Related tasks: https://github.com/scylladb/scylla-cluster-tests/issues/7162 and https://github.com/scylladb/scylla-cluster-tests/issues/7160

### Testing
- [x] :green_circle: [db packages installation scenario intended to fail on update_db_packages step](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/regression-latency-650gb-grow-shrink/10/)
- [x] :green_circle: [positive test run of updating db packages](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-test/45/)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -40,6 +40,7 @@ from collections.abc import Iterator
 from dataclasses import dataclass
 from pathlib import Path
 from contextlib import ExitStack, contextmanager
+from concurrent.futures import ThreadPoolExecutor, as_completed
 import packaging.version
 
 import yaml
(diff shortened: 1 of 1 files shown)


dimakr:
@fruch 
as was discussed, executed a positive scenario for the update_db_packages flow, added reference to the description

fruch (approved):
LGTM

PR #7701, opened 2024-06-20
feature(github): cache github issues status in S3

fruch:
Ref: https://github.com/scylladb/qa-tasks/issues/1678

### Testing

- [ ]

### Reminders

### DIFF
--- a/.github/workflows/cache-issues.yaml
+++ b/.github/workflows/cache-issues.yaml
@@ -1,0 +1,27 @@
+name: Cache issues status
+on:
+  schedule:
+    - cron: '0 */6 * * *'
+
+jobs:
+  collect_n_upload:
+    runs-on: ubuntu-latest
+    steps:
+      - run: |
+          mkdir -p issues
+          for repo in scylladb scylla-enterprise scylla-manager scylla-operator scylla-cluster-tests scylla-dtest qa-tasks scylla-tools-java ; do
+            gh issue list --state all --json number,state,labels --limit 30000 --template '{{range .}}{{.number}},{{.state}},{{range .labels}}{{.name}}|{{end}}{{println ""}}{{end}}' --repo scylladb/$repo > issues/scylladb_$repo.csv
+          done
+        env:
+          GH_TOKEN: ${{ secrets.ISSUE_ASSIGNMENT_TO_PROJECT_TOKEN }}
+      - name: Upload folder to bucket
+        uses: a-sync/s3-uploader@2.0.1
+        with:
+          args: --recursive
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.S3_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.S3_AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: 'us-east-1'
+          S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
+          S3_KEY: 'issues'
+          FILE: ./issues
--- a/sdcm/utils/issues.py
+++ b/sdcm/utils/issues.py
@@ -1,5 +1,6 @@
 import re
 import sys
+import csv
 import logging
 from functools import lru_cache
 from dataclasses import dataclass
@@ -7,8 +8,9 @@ from dataclasses import dataclass
 import github
 import github.Auth
 import github.Issue
+import github.Label
 from github.GithubException import UnknownObjectException, RateLimitExceededException
-
+from botocore.exceptions import ClientError
 from sdcm.keystore import KeyStore
 from sdcm.sct_config import SCTConfiguration
 from sdcm.sct_events.base import Severity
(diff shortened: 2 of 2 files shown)


.github/workflows/cache-issues.yaml
```diff
@@ -0,0 +1,33 @@
+name: Cache issues status
+on:
+  schedule:
+    # Daily at 8:20 UTC
+    - cron: '20 8 * * *'
+  pull_request:
```
fruch, line 6: remove pull_request and push, it was just for testing.

.github/workflows/cache-issues.yaml
```diff
@@ -0,0 +1,33 @@
+name: Cache issues status
+on:
+  schedule:
+    # Daily at 8:20 UTC
+    - cron: '20 8 * * *'
```
fruch, line 5: pick other time, also consider running it more then once a day

sdcm/utils/issues.py
```diff
@@ -81,6 +105,18 @@ def get_issue_details(self, issue):
                                severity=Severity.WARNING,
                                trace=sys._getframe().f_back).publish()  # pylint: disable=protected-access
             return None
+        try:
+            if issue_details := self.cache.get_issue(owner=issue_parsed.user_id, repo_id=issue_parsed.repo_id, issue_id=issue_parsed.issue_id):
+                labels = [dict(name=label) for label in issue_details['labels'].strip().rstrip('|').split('|')]
```
fruch, line 110: this can move to CachedGitHubIssues (that would be copied into dtest)

sdcm/utils/issues.py
```diff
@@ -28,6 +30,26 @@ class Issue:
...
 
 
+class CachedGitHubIssues:
+    def __init__(self):
+        self.storage = KeyStore()
+
+    @lru_cache()
+    def get_repo_data(self, owner, repo):
+        scsv = self.storage.get_file_contents(f'issues/{owner}_{repo}.csv')
+        print('got repo')
```
soyacz, line 40: debug leftover

sdcm/utils/issues.py
soyacz: +1

sdcm/utils/issues.py
```diff
@@ -28,6 +30,26 @@ class Issue:
...
+        return {issue['id']: issue for issue in csv.DictReader(scsv.decode().splitlines(), fieldnames=("id", "state", "labels"))}
+
+    def get_issue(self, owner: str, repo_id: str, issue_id: str | int):
+        repo_issues_mapping = self.get_repo_data(owner, repo_id)
+        return repo_issues_mapping.get(str(issue_id))
+
+
+@lru_cache(maxsize=1)
+def get_issues_cache():
+    return CachedGitHubIssues()
```
soyacz, line 50: can also make it singleton class

sdcm/utils/issues.py
```diff
@@ -28,6 +30,26 @@ class Issue:
     issue_id: int | None
 
 
+class CachedGitHubIssues:
```
soyacz, line 33: I'd add docstring how cache is populated - reference to gh action.

.github/workflows/cache-issues.yaml
```diff
@@ -0,0 +1,33 @@
...
+          GH_TOKEN: ${{ secrets.ISSUE_ASSIGNMENT_TO_PROJECT_TOKEN }}
+      - name: Upload folder to bucket
+        uses: a-sync/s3-uploader@2.0.1
+        with:
+          args: --recursive
+        env:
+          AWS_ACCESS_KEY_ID: ${{ secrets.S3_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.S3_AWS_SECRET_ACCESS_KEY }}
+          AWS_REGION: 'us-east-1'
+          S3_BUCKET: ${{ secrets.AWS_S3_BUCKET }}
```
soyacz, line 31: make sure it's not readable by everyone

.github/workflows/cache-issues.yaml
soyacz: once per hour? IIRC limits are hour-based

.github/workflows/cache-issues.yaml
```diff
@@ -0,0 +1,33 @@
+name: Cache issues status
```
soyacz, line 1: is there some retry mechanism when it fails? 
Will we be informed when it fails and how?

soyacz:
I wonder also about if this cache could be reused by Argus. In that case we might want to also cache also issue titles.

.github/workflows/cache-issues.yaml
fruch: I think one a day should be enough

we don't want thing to start running too soon, 
consider a test currently running with version that doesn't yet have a fix.

we don't need this accurate to the hour level, one day is enough.

sdcm/utils/issues.py
fruch: just used @cache on the class, it's exactly the same

.github/workflows/cache-issues.yaml
fruch: it's exactly where the rest of important info that SCT save, it's not public

.github/workflows/cache-issues.yaml
fruch: no retry, and it's visible on the actions page,

I think admin would get notification, but not sure.

sdcm/utils/issues.py
fruch: added one

sdcm/utils/issues.py
fruch: removed

sdcm/utils/issues.py
fruch: moved it

.github/workflows/cache-issues.yaml
fruch: we'll see, maybe as a follow-up we'll add a step to send mail notification,
but it's a bit more work to configure such a thing. (need to pass gmail cred in)

soyacz: ok, let's see how it looks in practice and adapt as needed

soyacz (approved):
LGTM

.github/workflows/cache-issues.yaml
fruch: done

PR #7702, opened 2024-06-20
improvement(parallel-operations): grow-shrink cluster

soyacz:
Adding and decommissioning nodes should be run in parallel as this is currently Scylla's default behavior.

Refactored code to support it. Can be disabled if
`parallel_node_operations` param is set to false.

### Testing

- [x] - provision tests
- [x] - [add 3 nodes in parallel, single AZ](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-GrowShrinkClusterNemesis-aws-test/23/)
- [x] - [add 3 nodes sequentially, single az](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-GrowShrinkClusterNemesis-aws-test/22/)
- [x] - [add 3 nodes in parallel, 3 az's](https://argus.scylladb.com/test/a2af091e-3cbd-4068-b825-35d511f15d17/runs?additionalRuns[]=3c8be89c-3ca1-4065-958c-c4ef0cf3ae05) - works
- [x] - [grow-shrink on gce](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/nemesis/job/longevity-5gb-1h-GrowShrinkClusterNemesis-gce-test/2/)
- [x] - [terminate and replace node](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/nemesis/job/longevity-5gb-1h-NodeTerminateAndReplace-aws-test/2/)
- [x] - [grow-shrink perf test - parallel](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/scylla-master-perf-regression-latency-650gb-grow-shrink/38/)
- [x] - [grow-shrink perf test - sequential](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/scylla-master-perf-regression-latency-650gb-grow-shrink/40/) - failed due reasons fixed by https://github.com/scylladb/scylla-cluster-tests/pull/7761 I consider it working

### Reminders

### DIFF
--- a/sdcm/cluster_aws.py
+++ b/sdcm/cluster_aws.py
@@ -389,9 +389,21 @@ class AWSCluster(cluster.BaseCluster):  # pylint: disable=too-many-instance-attr
         instance_dc = 0 if self.params.get("simulated_regions") else dc_idx
         # if simulated_racks, create all instances in the same az
         instance_az = 0 if self.params.get("simulated_racks") else rack
-        instances = self._create_or_find_instances(
-            count=count, ec2_user_data=ec2_user_data, dc_idx=instance_dc, az_idx=instance_az,
-            instance_type=instance_type)
+        instances = []
+        if rack is None:
+            # define how many nodes should be created on each rack, e.g. for 3 nodes and 2 racks it will be [2, 1]
+            base = count // self.racks_count
+            extra = count % self.racks_count
+            rack_distribution = [base + 1 if i < extra else base for i in range(self.racks_count)]
+        else:
+            # otherwise create all nodes on the specified rack
+            rack_distribution = [count if i == rack else 0 for i in range(self.racks_count)]
+        self.log.info('rack distribution: %s', rack_distribution)
+        for rack_idx, rack_count in enumerate(rack_distribution):
+            if rack_count == 0:
+                continue  # when provisioning fewer nodes than racks
+            instances.extend(self._create_or_find_instances(
+                count=rack_count, ec2_user_data=ec2_user_data, dc_idx=instance_dc, az_idx=rack_idx, instance_type=instance_type))
         for node_index, instance in enumerate(instances):
             self._node_index += 1
             # in case rack is not specified, spread nodes to different racks
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3976,6 +3976,7 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
             LOGGER.info("Starting nodes in parallel")
         else:
             LOGGER.info("Starting nodes sequentially")
+            self.parallel_node_operations = False  # disable parallel operations regardless of test config
 
         return parallel_startup
 
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -73,6 +73,7 @@ from sdcm.prometheus import nemesis_metrics_obj
 from sdcm.provision.scylla_yaml import SeedProvider
 from sdcm.provision.helpers.certificate import update_certificate, TLSAssets
 from sdcm.remote.libssh2_client.exceptions import UnexpectedExit as Libssh2UnexpectedExit
+from sdcm.rest.remote_curl_client import RemoteCurlClient
 from sdcm.sct_events import Severity
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.decorators import raise_event_on_failure
@@ -1224,8 +1225,8 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if not ContainerManager.is_running(self.tester.localhost, 'ldap'):
             raise LdapNotRunning("LDAP server was supposed to be running, but it is not")
 
-    def _add_and_init_new_cluster_node(self, old_node_ip=None, host_id=None,
-                                       timeout=MAX_TIME_WAIT_FOR_NEW_NODE_UP, rack=0):
+    def _replace_cluster_node(self, old_node_ip=None, host_id=None,
+                              timeout=MAX_TIME_WAIT_FOR_NEW_NODE_UP, rack=0):
         """When old_node_ip or host_id are not None then replacement node procedure is initiated"""
         # TODO: make it work on K8S when we have decommissioned (by nodetool) nodes.
         #       Now it will fail because pod which hosts decommissioned Scylla member is reported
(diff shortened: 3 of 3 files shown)


soyacz:
There's one remark about this feature: `latency_calculator_decorator` which showed 3 cycles when adding 3 nodes, now will only show one cycle. I don't know yet how to fix it - before that, I need to know if we really need it. @roydahan cc.

fruch:
Since we were having issues that latency was bad during specific node addition 

It might be useful to keep, 

But it complicates the code leaving a whole chain of functions duplicated.

Maybe having a dedicated specific nemesis is the better option, especially when we need to support force-gossip-topology at the same time ?

soyacz:
> Since we were having issues that latency was bad during specific node addition
> 
> It might be useful to keep,
> 
> But it complicates the code leaving a whole chain of functions duplicated.
> 
> Maybe having a dedicated specific nemesis is the better option, especially when we need to support force-gossip-topology at the same time ?

This affects only perf test - which can stay on branch perf-v14 for force-gossip-topology.
I just have got an idea how to modify code to keep support one-by-one addition/decommission. I'll update the code (or write a comment if my idea fails).

soyacz:
> This affects only perf test - which can stay on branch perf-v14 for force-gossip-topology. I just have got an idea how to modify code to keep support one-by-one addition/decommission. I'll update the code (or write a comment if my idea fails).

ok, I think I managed to update grow-shrink nemesis to add nodes one by one, so latency wrapper should work as expected when parallel operations are off.

sdcm/utils/raft/__init__.py
```diff
@@ -331,7 +331,11 @@ def is_cluster_topology_consistent(self) -> bool:
 
 def get_raft_mode(node) -> RaftFeature | NoRaft:
     with node.parent_cluster.cql_connection_patient(node) as session:
-        return RaftFeature(node) if is_consistent_cluster_management_feature_enabled(session) else NoRaft(node)
+        raft = RaftFeature(node) if is_consistent_cluster_management_feature_enabled(session) else NoRaft(node)
+        if not raft.is_enabled:
+            LOGGER.debug("Disabling parallel nodes operations because raft is disabled")
+            node.parent_cluster.parallel_node_operations = False
+        return raft
```
fruch, line 338: Not sure I really like the side effect in such a function

We should find a better place to use this function, and apply those changes to configure

soyacz: yes, I also wasn't happy about it. I'll try to find a better place.

fruch: Also one day we'll need to lock down the configuration and make it immutable during test run time.

We can apply the logic just before we gonna use the flag configuration

fruch: Now I see it's not configuration...  but still confusing to some degree, cause there is such configuration

soyacz: I removed it. Actually we need to disable `parallel_node_operations` when we force gossip - when this config param will be introduced it should be taken into account. Otherwise, version&config base should be enough for now.

sdcm/cluster_aws.py
```diff
@@ -389,9 +389,21 @@ def add_nodes(self, count, ec2_user_data='', dc_idx=0, rack=0, enable_auto_boots
...
+            # define how many nodes should be created on each rack, e.g. for 3 nodes and 2 racks it will be [2, 1]
+            base = count // self.racks_count
+            extra = count % self.racks_count
+            rack_distribution = [base + 1 if i < extra else base for i in range(self.racks_count)]
+        else:
+            # otherwise create all nodes on the specified rack
+            rack_distribution = [count if i == rack else 0 for i in range(self.racks_count)]
+        self.log.info('rack distribution: %s', rack_distribution)
+        for rack_idx, rack_count in enumerate(rack_distribution):
+            if rack_count == 0:
```
vponomaryov, line 403: This condition always will return `False`, because the `rack_count` will always be a `list`.
So, it is bug that should be fixed.

vponomaryov (changes requested):

sdcm/cluster_aws.py
soyacz: `rack_distribution` is a list which we iterate through with enumerate

configurations/nemesis/GrowShrinkClusterNemesis.yaml
```diff
@@ -1,2 +1,8 @@
 nemesis_class_name: 'GrowShrinkClusterNemesis'
 user_prefix: 'longevity-5gb-1h-GrowShrinkClusterNemesis'
+stress_cmd:
+    - "cassandra-stress write cl=QUORUM duration=60m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate 'threads=40 throttle=5000/s' -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5"
+    - "cassandra-stress read  cl=QUORUM duration=60m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate 'threads=40 throttle=5000/s' -pop seq=1..5048570 -col 'n=FIXED(8) size=FIXED(128)' -log interval=5"
+#    - "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=10000 -clustering-row-count=555 -clustering-row-size=uniform:128..2048 -concurrency=100 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=30s -retry-number=30 -retry-interval=80ms,1s -iterations 0 -duration=60m -validate-data"
+# disabled temporarily due https://github.com/scylladb/scylla-cluster-tests/issues/7705
+use_mgmt: false
```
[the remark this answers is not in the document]
soyacz: yes, here are some debug leftovers, will be removed before merge.

sdcm/cluster_aws.py
vponomaryov: Ah, right. Nevermind.

vponomaryov:
LGTM, if it works.

@soyacz are the test results from the current PR state?

vponomaryov (commented):
LGTM, if it works and debug part gets removed.

@soyacz are the test results from the current PR state?

soyacz:
> LGTM, if it works and debug part gets removed.

debug part removed.

> 
> @soyacz are the test results from the current PR state?

Not much added/removed, so can say yes.

vponomaryov (approved):
LGTM

fruch (approved):

PR #7600, opened 2024-06-07
fix(benchmarks): disable node benchmarks

soyacz:
After examination of test runs, node benchmarks don't bring any valuable information: seems there's no correlation between its results and performance of nodes.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/7468

### Testing

- [ ]

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #7704, opened 2024-06-20
fix(screenshot): revert to specific resolution

fruch:
we were using height=-1 for the renderer to automticlly capture all of the dashboard.

seems like with monitoring 4.7, this got broken cause of newer grafana version, and dashboard are coming out cut, and some ever completly broken and not readable.

couldn't pinpoint which exact version is causing it.

so meanwhile reverting to use exact resolution, that seems to be working as expected

Fixes: #7600

### Testing

- [ ] provision test - should check the snapshot

### Reminders

### DIFF
--- a/sdcm/logcollector.py
+++ b/sdcm/logcollector.py
@@ -467,7 +467,6 @@ class GrafanaScreenShot(GrafanaEntity):
     Extends:
         GrafanaEntity
     """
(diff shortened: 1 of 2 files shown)


fruch:
Seems like it's working better with this fix

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

PR #7743, opened 2024-06-25
[branch-2024.2] fix(byo): use proper default values for BYO scylla params

vponomaryov:
After branching to the `branch-2024.2` we should update the BYO params default values to make it be proper ones.

### Testing

- [ ]

### Reminders

### DIFF
--- a/vars/byoScylladb.groovy
+++ b/vars/byoScylladb.groovy
@@ -23,7 +23,7 @@ def call(Map params, boolean build_image){
     if (params.byo_job_path) {
         jobToTrigger = params.byo_job_path
     } else {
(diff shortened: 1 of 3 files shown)


fruch:
I was aware this branch was created...

fruch:
We need to think of a better way of handling those, we would forget about in new branches

Let's discuss with @yaronkaikov, maybe it can be done as part of the branch creation.

fruch (approved):

yaronkaikov:
> We need to think of a better way of handling those, we would forget about in new branches
> 
> Let's discuss with @yaronkaikov, maybe it can be done as part of the branch creation.

We can add it from our side if needed, let's open an issue so we wont loss track of it

PR #7756, opened 2024-06-26
fix(nemesis): `hot_reloading_internode_certificate` to update correct certs

fruch:
since the refactor of reloading certs we started
seeing those errors to start new nodes:
```
[shard  0:main] init - Startup failed: std::system_error
   (error GnuTLS:-60, The certificate and the given key do not match.)
```

those were happening after `hot_reloading_internode_certificate` was trying to reload the certificates, problem it had an assumption updating certificate once, and sending it to all of the nodes. but now we generate certs per node, and it wasn't copying it from the correct place, hence generated cert that doesn't match the key.

ontop of that seems like the validation in that nemesis was wrongly using `follow_system_log`, cause it to missed looking for the actual line that shows the certs are reloaded (it seems to be broken like that since d2a7910b98e823086c69b01bec6d6735444e8a53)

### Testing

- [x] 🔴  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-SslHotReloadingNemesis-aws-test/3/
- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-SslHotReloadingNemesis-aws-test/4/

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4302,7 +4302,6 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.target_node.get_scylla_config_param("server_encryption_options"))["certificate"]
         in_place_crt = self.target_node.remoter.run(f"cat {ssl_files_location}",
                                                     ignore_status=True).stdout
-        update_certificate()
         node_system_logs = {}
 
         if SkipPerIssues('https://github.com/scylladb/scylladb/issues/7909', params=self.tester.params):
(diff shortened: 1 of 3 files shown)


sdcm/nemesis.py
```diff
@@ -4267,10 +4266,11 @@ def check_ssl_reload_log(node_system_log):
         with context_manager:
             for node in self.cluster.nodes:
                 node_system_logs[node] = node.follow_system_log(
-                    patterns=f'messaging_service - Reloaded {{{ssl_files_location}}}')
-                node.remoter.send_files(src=f'data_dir/ssl_conf/{TLSAssets.DB_CERT}', dst='/tmp')
-                node.remoter.run(f"sudo cp -f /tmp/{TLSAssets.DB_CERT} {ssl_files_location}")
-                new_crt = node.remoter.run(f"cat {ssl_files_location}").stdout
+                    patterns=[f'messaging_service - Reloaded {ssl_cert_location}'])
```
fruch, line 4269: put this oneliner change into it's own commit, so we can backport it

soyacz:
test failed with `sdcm.exceptions.LogContentNotFound: Reload SSL message not found in node log`

Btw. I propose to add restarting one node after the test temporarily to make sure it works. Just for the sake of testing this PR and to be removed before merge.

fruch:
> test failed with `sdcm.exceptions.LogContentNotFound: Reload SSL message not found in node log`
kind of expected, after it wasn't in use for almost 4 years... 

> 
> Btw. I propose to add restarting one node after the test temporarily to make sure it works. Just for the sake of testing this PR and to be removed before merge.

first I need to fix the validation, or I won't get into any restart code at the end. it's a nice touch, we'll see about it.

fruch:
the fix to the validation should go on it's own in: https://github.com/scylladb/scylla-cluster-tests/pull/7787
so it can be backported to all releases

sdcm/nemesis.py
fruch: it's not on it's own PR
https://github.com/scylladb/scylla-cluster-tests/pull/7787

PR #7757, opened 2024-06-26
fix(jepsen): revert of old RSA keys

fruch:
since jepsen is using very old ssh library (jsch), we can only use RSA keys and our new keys are ECDSA, so we need to revert using the old key we have

### Testing

- [ ] 🕦 https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/jepsen-test-all/14/

### Reminders

### DIFF
--- a/test-cases/jepsen/jepsen.yaml
+++ b/test-cases/jepsen/jepsen.yaml
@@ -20,3 +20,8 @@ jepsen_scylla_repo: 'https://github.com/scylladb/jepsen.git'
 jepsen_test_count: 5
 jepsen_test_run_policy: any
 bare_loaders: true
+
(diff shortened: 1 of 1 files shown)


fruch:
> @fruch test failed. Looks like there is no pair-key on monitor node...

It's worse, we don't have the hack in cloud-init to enable rsa keys.

So reverting is not that straight forward

Back to drawing board

soyacz:
can we make jepsen to use different keys than SCT?
So we could enable accepting these rsa keys as part of jepsen test itself.

fruch:
update jepsen, and found a way to use the newer keys, closing this one

PR #7763, opened 2024-06-27
fix(hydra-ssh): tunnel shouldn't pick bastions with other keys

fruch:
since we switch diffrent keys, the tunneling can fail picking the a correct bastion, we now filter the the list of bastions to make sure we don't mix, and use diffrent keys from the one configured for an instance.

(those could be coming for older releases,
or people using older branches)

### Testing

- [x] tested locally

### Reminders

### DIFF
--- a/sct_ssh.py
+++ b/sct_ssh.py
@@ -76,6 +76,7 @@ def aws_find_bastion_for_instance(instance: dict) -> dict:
     region = get_region(instance)
     tags = {'bastion': 'true'}
     bastions = list_instances_aws(tags, running=True, region_name=region)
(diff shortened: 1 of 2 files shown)


soyacz (commented):
To clarify my understanding: It happens when someone runs a test from older branch, creating new bastion which could be used by recent branch?

fruch:
> To clarify my understanding: It happens when someone runs a test from older branch, creating new bastion which could be used by recent branch?

Yes,

So far we have picked them based on bastion=true, (and in AWS we were checking it's the VPC)

But now we might be mixing the keys as well, so now we are checking the keys as well.

technically we could mix the keys, but we need to make sure we are using the correct key for each instance

soyacz (approved):
LGTM

Issue #7780, opened 2024-06-30
`_add_and_init_new_cluster_nodes` return value is treated as node


fruch:
seems since 798228e278ad721eb580331697f13fa82280620 was merged 

decommission nemesis is failing in all kind of ways:

```python
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5201, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1329, in disrupt_nodetool_decommission
    return self._nodetool_decommission(add_node=add_node)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1315, in _nodetool_decommission
    if new_node.is_seed != target_is_seed:
AttributeError: 'list' object has no attribute 'is_seed'
```

`_add_and_init_new_cluster_nodes` is called, and it return a list of nodes, and not just one

PR #7783, opened 2024-06-30
fix(nemesis): handle fallout from parallel-operations changes

fruch:
PR #7702 introduced changes to nemesis code, that start retruning mutiple nodes back from the refactored `_add_and_init_new_cluster_node` function, some of the places using the function, was expecting exactly one node in return, and not a list.

Fixes: #7780

### Testing

- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-NodeTerminateAndReplace-aws-test/2
- [x] 🟢  https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/longevity-5gb-1h-DecommissionMonkey-aws-test/2

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1265,7 +1265,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         InfoEvent(message="FinishEvent - New Node is up and normal").publish()
         return new_node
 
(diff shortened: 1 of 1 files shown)


soyacz:
I wonder if it's worth to introduce type checking like `mypy` to pre-commit hook - how much work it would cost us.
With it we could have prevented this error.

fruch:
> I wonder if it's worth to introduce type checking like `mypy` to pre-commit hook - how much work it would cost us. With it we could have prevented this error.

it might worth it, but it would be tedious, i.e. doing it file by file.

PR #7761, opened 2024-06-27
Fix disable tablets and perf stats upload to es

soyacz:
2 issues I found when testing perf on master after recent changes:
1. `enable_tablets` was not added to `ScyllaYaml` making `disable_tablets.yaml` configuration not working
2. `append_scylla_yaml` since is a dict, when has value, ES refuses to insert it due text expectation.

Both issues fixed with these 2 commits.

### Testing

- [ ] - [elasticity test, different branch with the same fix applied](https://jenkins.scylladb.com/job/scylla-staging/job/lukasz/job/scylla-master-perf-regression-latency-650gb-grow-shrink/43/)

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #7799, opened 2024-07-01
feature(perf-tests): disable tablets by default

fruch:
all of the current test should be running without tablets enabled specific case for tablets would be pick and run with it enabled

### Testing

- [ ]

### Reminders


soyacz:
in that case, https://github.com/scylladb/scylla-cluster-tests/pull/7761 needs to be backported too. Also need to create `tablets_disabled.yaml` (or backport https://github.com/scylladb/scylla-cluster-tests/pull/7655 which I'm not sure if feasible).

fruch:
> in that case, #7761 needs to be backported too. Also need to create `tablets_disabled.yaml` (or backport #7655 which I'm not sure if feasible).

backported the relevant things

PR #7822, opened 2024-07-01
fix(artifacts-test): make creation of CA and Java truststore idempotent

dimakr:
There are a couple of error-prone situations when CA cert/key pair is created:
- the CA is created for every test run, even if the configuration under test
doesn't have encryption enabled (hence don't need CA to sign nodes certs)
- in case of artifacts tests there are situations when a Jenkins builder node
hosts several artifacts tests in a row, i.e. the test environment is not cleaned up
between tests. In this case SCT will try to import CA pem certificate to existing
Java truststore and will fail

The change addresses both problems:
- creation of CA is made optional, depending on whether client and/or server encryption
are enabled in the test configuration
- creation of CA is made idempotent between different executions of a test on the
same node - a check is added to CA creation procedure to ensure that CA is not already
present on the node

### Testing
- [x] :green_circle: [artifacts-ubuntu2004-arm-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts-ubuntu2004-arm-test/4/)
- [x] :green_circle: [manager node test (which enables client encryption on the fly)](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/ubuntu22-sanity-test/15/)
- [x] :green_circle: [ssl_hot_reloading nemesis test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/certs-rework-tests/28/)
- [x] :green_circle: [pr-provision test with client+server encryption](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/sct-pr-provision-test/9/)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -69,7 +69,7 @@ from sdcm.provision.scylla_yaml.certificate_builder import ScyllaYamlCertificate
 from sdcm.provision.scylla_yaml.cluster_builder import ScyllaYamlClusterAttrBuilder
 from sdcm.provision.scylla_yaml.scylla_yaml import ScyllaYaml
 from sdcm.provision.helpers.certificate import (
-    install_client_certificate, install_encryption_at_rest_files, create_certificate,
+    create_ca, install_client_certificate, install_encryption_at_rest_files, create_certificate,
     export_pem_cert_to_pkcs12_keystore, CA_CERT_FILE, CA_KEY_FILE, JKS_TRUSTSTORE_FILE, TLSAssets)
 from sdcm.remote import RemoteCmdRunnerBase, LOCALRUNNER, NETWORK_EXCEPTIONS, shell_script_cmd, RetryableNetworkException
 from sdcm.remote.libssh2_client import UnexpectedExit as Libssh2_UnexpectedExit
@@ -2283,12 +2283,6 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
             self.download_scylla_manager_repo(manager_repo_url)
         self.install_package(package_names)
 
-        self.log.debug("Create and send client TLS certificate/key to the node")
-        self.create_node_certificate(cert_file=self.ssl_conf_dir / TLSAssets.CLIENT_CERT,
-                                     cert_key=self.ssl_conf_dir / TLSAssets.CLIENT_KEY)
-        self.remoter.run(f'mkdir -p {mgmt.cli.SSL_CONF_DIR}')
-        self.remoter.send_files(src=str(self.ssl_conf_dir) + '/', dst=str(mgmt.cli.SSL_CONF_DIR))
-
         if self.is_docker():
             try:
                 self.remoter.run("echo no | sudo scyllamgr_setup")
@@ -4085,6 +4079,16 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
                 node.restart_scylla()
 
     def enable_client_encrypt(self):
+        create_ca(self.test_config.tester_obj().localhost)
+        for node in self.nodes:
+            node.create_node_certificate(cert_file=node.ssl_conf_dir / TLSAssets.DB_CERT,
+                                         cert_key=node.ssl_conf_dir / TLSAssets.DB_KEY,
+                                         csr_file=node.ssl_conf_dir / TLSAssets.DB_CSR)
+            # Create client facing node certificate, for client-to-node communication
+            node.create_node_certificate(
+                node.ssl_conf_dir / TLSAssets.DB_CLIENT_FACING_CERT, node.ssl_conf_dir / TLSAssets.DB_CLIENT_FACING_KEY)
+            for src in (CA_CERT_FILE, JKS_TRUSTSTORE_FILE):
+                shutil.copy(src, node.ssl_conf_dir)
         self.log.debug("Enabling client encryption on nodes")
         with self.patch_params() as params:
             params['client_encrypt'] = True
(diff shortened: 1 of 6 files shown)


sdcm/provision/helpers/certificate.py
```diff
@@ -93,6 +98,18 @@ def install_encryption_at_rest_files(remoter):
...
+    def wrapper(*args, **kwargs):
+        if CA_CERT_FILE.exists() and CA_KEY_FILE.exists():
+            LOGGER.info(
+                "CA %s certificate and %s key already exist. Skipping creation.", CA_CERT_FILE, CA_KEY_FILE)
+            return
+        return func(*args, **kwargs)
+    return wrapper
+
+
+@cached_ca
```
fruch, line 112: Why the built in @cache decorator isn't enough ?

fruch (commented):
Why not move the call to this out of test setup, while you are at it ?

sdcm/provision/helpers/certificate.py
soyacz: caching-like between tests I suppose - IIUC previous one from the same job are left.
I wonder why we need that wrappers and can't this code be moved to functions itself

dimakr:
> Why not move the call to this out of test setup, while you are at it ?

What if to put the CA creation under a condition that checks if client/server encrypt is enabled in config?
Then the call to create a CA will remain in a single place in SCT, but will be done only when needed.

fruch:
> > Why not move the call to this out of test setup, while you are at it ?
> 
> What if to put the CA creation under a condition that checks if client/server encrypt is enabled in config? Then the call to create a CA will remain in a single place in SCT, but will be done only when needed. @fruch

manager test enable encryption during the test, and we don't have any condition we can check

dimakr:
> manager test enable encryption during the test, and we don't have any condition we can check

Right, for that scenario we need to call for CA creation separately in any case - whether the create_ca is moved out of tester.setUp or put under a condition..

fruch:
> > manager test enable encryption during the test, and we don't have any condition we can check
> 
> Right, for that scenario we need to call for CA creation separately in any case - whether the create_ca is moved out of tester.setUp or put under a condition..

so if we do that where it's being used, it might be good enough ?

dimakr:
> so if we do that where it's being used, it might be good enough ?

It should, just need to test that there won't be a problem/race when loader/cluster node sets are created in their threads, as both need CA to sign their certs/keys - https://github.com/scylladb/scylla-cluster-tests/blob/1cbc0c8fe7811ac067fa165ff50533c51c23bca5/sdcm/tester.py#L978

sdcm/cluster.py
```diff
@@ -3957,6 +3957,9 @@ def __init__(self, *args, **kwargs):
         self._node_cycle = None
         self.params = kwargs.get('params', {})
         self.parallel_node_operations = self.params.get("parallel_node_operations") or False
+        if any([self.params.get('client_encrypt'), self.params.get('server_encrypt'),
+                self.params.get('internode_encryption')]):
+            create_ca(self.test_config.tester_obj().localhost)
```
fruch, line 3962: I think this can be move to BaseCluster, or even back to tester, if it's guarded like this 

in general we should avoid doing too much actions inside `__init__`, when doing unit tests it tend to complex things

soyacz: +1
it uses `tester_obj` so possibly `Tester` is better place

sdcm/provision/helpers/certificate.py
```diff
@@ -321,3 +337,8 @@ def c_s_transport_str(client_mtls: bool) -> str:
         transport_str = (
             f'{transport_str} keystore={SCYLLA_SSL_CONF_DIR}/{TLSAssets.PKCS12_KEYSTORE} keystore-password=cassandra')
     return transport_str
+
+
+def create_ca(localhost):
```
soyacz, line 342: maybe rename to `create_and_import_ca`?

sdcm/provision/helpers/certificate.py
dimakr: Correct, checking the file is to ensure the we do not try to recreate it in another test process

sdcm/cluster.py
dimakr: Reverted this back

sdcm/provision/helpers/certificate.py
dimakr: I think 'create_ca' is better to show the idea that we prepare here a common CA, which is to be used to sign/validate other certificates.
The fact that cassandra-stress doesn't support PEM formatted CA and requires extra step to convert it to java keystore format are technical specifics, which we probably can omit when defining API name for the operation (after all c-s is 1 of 3 stress tools that are used in tests, the others does use PEM formatted CAs).

fruch (approved):

soyacz (approved):
LGTM

sdcm/tester.py
```diff
@@ -903,7 +902,8 @@ def setUp(self):  # pylint: disable=too-many-branches,too-many-statements
         self.connections = []
         make_threads_be_daemonic_by_default()
 
-        self.create_ca()
+        if any([self.params.get('client_encrypt'), self.params.get('server_encrypt')]):
+            create_ca(self.localhost)
```
vponomaryov, line 906: It should not run on K8S backends. K8S does this stuff on it's own.

vponomaryov (changes requested):

sdcm/cluster.py
```diff
@@ -4650,15 +4654,16 @@ def node_setup(self, node: BaseNode, verbose: bool = False, timeout: int = 3600)
...
-        # Create node certificate for internode communication
-        node.create_node_certificate(cert_file=node.ssl_conf_dir / TLSAssets.DB_CERT,
-                                     cert_key=node.ssl_conf_dir / TLSAssets.DB_KEY,
-                                     csr_file=node.ssl_conf_dir / TLSAssets.DB_CSR)
-        # Create client facing node certificate, for client-to-node communication
-        node.create_node_certificate(
-            node.ssl_conf_dir / TLSAssets.DB_CLIENT_FACING_CERT, node.ssl_conf_dir / TLSAssets.DB_CLIENT_FACING_KEY)
-        for src in (CA_CERT_FILE, JKS_TRUSTSTORE_FILE):
-            shutil.copy(src, node.ssl_conf_dir)
+        if self.params.get('server_encrypt') and not self.params.get("cluster_backend").startswith("k8s"):
```
vponomaryov, line 4657: K8S redefines the `node_setup` method: https://cloud.google.com/compute/docs/images/os-details#debian
So, here this check was not needed.
It was needed in the `tester.py`'s `setUp` part.

sdcm/cluster.py
```diff
@@ -5098,14 +5103,15 @@ def node_setup(self, node, verbose=False, **kwargs):  # pylint: disable=unused-a
             self.log.info("Don't install anything because bare loaders requested")
             return
 
-        node.create_node_certificate(node.ssl_conf_dir / TLSAssets.CLIENT_CERT,
-                                     node.ssl_conf_dir / TLSAssets.CLIENT_KEY)
-        for src in (CA_CERT_FILE, JKS_TRUSTSTORE_FILE):
-            shutil.copy(src, node.ssl_conf_dir)
-        if self.params.get('client_encrypt'):
+        if self.params.get('client_encrypt') and not self.params.get("cluster_backend").startswith("k8s"):
```
vponomaryov, line 5106: Same here. The `k8s` check was needed only in the `tester.py`.
Redundant, not harmful.

dimakr: Got it. Reverted for nodes

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

soyacz:
don't we need backports for this?

fruch:
> don't we need backports for this? 

https://github.com/scylladb/scylla-cluster-tests/pull/7436

Wasn't backported anywhere, so this shouldn't

fruch (approved):

dimakr:
@Mergifyio refresh

dimakr:
@Mergifyio backport branch-2024.2

PR #7827, opened 2024-07-02
fix(parallel-operations): disable when force gossip

soyacz:
When scylla yaml has `force_gossip_topology_changes` set to `True` then parallel node operations should be disabled.

Added proper handling, configuration and updated `ScyllaYaml` model.

### Testing

- [x] - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/nemesis/job/longevity-5gb-1h-GrowShrinkClusterNemesis-gce-test/4/

### Reminders

### DIFF
--- a/configurations/force-gossip-topology-changes.yaml
+++ b/configurations/force-gossip-topology-changes.yaml
@@ -1,0 +1,3 @@
+append_scylla_yaml:
+  force_gossip_topology_changes: true
+  enable_tablets: false
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3956,7 +3956,8 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
         self.test_config = TestConfig()
         self._node_cycle = None
         self.params = kwargs.get('params', {})
-        self.parallel_node_operations = self.params.get("parallel_node_operations") or False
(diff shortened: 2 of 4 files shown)


fruch (approved):
LGTM

vponomaryov (changes requested):
Need to update unit tests.
After it LGTM.

soyacz:
> Need to update/fix unit tests. After it LGTM.

yeah, I always forget about it. 
Anyway I did one more update to disable tablets along in the configuration as test failed when creating schema (possibly due tablets enabled).

fruch:
> > Need to update/fix unit tests. After it LGTM.
> 
> yeah, I always forget about it. Anyway I did one more update to disable tablets along in the configuration as test failed when creating schema (possibly due tablets enabled).

I think we should fix those test to not be so picky about the content, we gain nothing from it.

soyacz:
> > > Need to update/fix unit tests. After it LGTM.
> > 
> > 
> > yeah, I always forget about it. Anyway I did one more update to disable tablets along in the configuration as test failed when creating schema (possibly due tablets enabled).
> 
> I think we should fix those test to not be so picky about the content, we gain nothing from it.

Anyway, fixed.
Jenkins test also looks like passing now and `Starting nodes sequentially` is in logs.

aleksbykov (approved):
LGTM

Issue #7496, opened 2024-05-28
perf_simple_query results table is broken


soyacz:
`logallocs_per_op` was added recently, breaking results table sent over email:
![image](https://github.com/scylladb/scylla-cluster-tests/assets/18242288/30d504bd-83bf-42e7-a22f-29fa9ab9db09)

PR #7832, opened 2024-07-02
fix(perf-simple-query): fix results table

soyacz:
When new column to perf simple query test is added it breaks the result table.

Fix by getting available columns from current run and create table with all of them. In case there's no result for that column in the past, replace with `N/A`.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/7496

### Testing

tested in staging:
image

- [X] I added the relevant `backport` labels

### Reminders


temichus (approved):
LGTM

fruch (approved):

Issue #7781, opened 2024-06-30
Run long background ssh commands on host


fruch:
We have multiple reports of this issue  https://github.com/scylladb/scylladb/issues/14004

which basically boils down to the node being stressed out during the a long operation, that we don't automatically retry (because logically we can't retry those on disconnect, like decommission/cleanup and such) 

we observed this issue is happening all the time with other commands as well, but those commands get retried, and we don't notice it.

the suggest is to add implementation in remoter code that can send off the command to run on the host, while polling it from time to time
and then getting back the log of the operation

each command in this flow (beside the initial call), can be retried, and this lower the chances of being hit by this issue to practicality zero (i.e. just one quick command of starting the command would be with a retries, and the rest would be)

other options would be to try to protect the sshd from those spikes, and preserve resources for it. but that's not completely guaranteed as well, and might affect scylla (since those resource would not be available for it)

Ref: https://github.com/scylladb/scylladb/issues/14004

fruch:
@roydahan @soyacz @vponomaryov, what do you think about this direction ?

roydahan:
So IIUC, you suggest that the remoter will have an "async" mode and polling is your direction of implementing it?

fruch:
> So IIUC, you suggest that the remoter will have an "async" mode and polling is your direction of implementing it?

yes, something like the following

```python
pid = remoter.run('f{cmd} 2>&1 > cmd_output.log ; echo "$!"').stdout
while curr_time > timeout:
    # check pid still alive
    sleep(10)
# read the cmd_output.log back and return it
```

roydahan:
In general I'm fine with it, I'm just trying to think if we can change something in the tools/long running commands themselves so somehow they can send us back the result / write it into a file / socket / whatever that will make it more integrated between the tools.

fruch:
> In general I'm fine with it, I'm just trying to think if we can change something in the tools/long running commands themselves so somehow they can send us back the result / write it into a file / socket / whatever that will make it more integrated between the tools.

well yes, it would be best if long running command could be list inside scylla and tracked via APIs, and not just a commandline tool.

I think there was some working related to it done by @Deexie with task manager, you can see https://github.com/scylladb/scylla-dtest/pull/2992 for how it's being use

soyacz:
> > So IIUC, you suggest that the remoter will have an "async" mode and polling is your direction of implementing it?
> 
> yes, something like the following
> 
> ```python
[...]

PR #7834, opened 2024-07-02
feature(remoter): execuation of long running command

fruch:
cause we have issues with long running ssh commands that we can retry, if node is a bit load at the time of the operation we might get disconnected, and fail the command complectly from SCT end.

in this change we introduce a way to run the command in the background while logging stdout, stderr, and exit code.

and poll for the finish of the process by its pid

in the end we'll collect all the logs, and retry results as if it with a regular remoter execution

Fixes: #7781

### Testing

- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-DecommissionMonkey-aws-test/11/ [Argus](https://argus.scylladb.com/test/ad50c9cd-7808-49cc-94dc-71d8a39f4ea0/runs)
- [x] 🟢   tier1 case where it happend last week: https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-schema-topology-changes-12h-test/3 [Argus](https://argus.scylladb.com/test/9d982f7d-6c51-4359-9aff-1a1ce6263c0c/runs?additionalRuns[]=d26dc22f-fda9-4b6e-b677-1d9fc1b5a263)

### Reminders

### DIFF
--- a/sdcm/remote/remote_long_running.py
+++ b/sdcm/remote/remote_long_running.py
@@ -1,0 +1,69 @@
+import time
+import logging
+from uuid import uuid4
+from pathlib import Path
+from contextlib import ExitStack
+
+from sdcm.remote.remote_cmd_runner import RemoteCmdRunnerBase
+from sdcm.remote.libssh2_client.result import Result
+from sdcm.remote.libssh2_client.exceptions import UnexpectedExit
+
+logger = logging.getLogger(__name__)
+
+
+def run_long_running_cmd(remoter: RemoteCmdRunnerBase,
+                         cmd: str,
+                         timeout: float | None = None,
+                         ignore_status: bool = False,
+                         verbose: bool = True,
+                         retry: int = 0):
+    """
+    Run a long-running command on the remote host. The command is executed in the background and the function waits
+    for the process of the command to finish.
+
+    this function is useful for running commands that take a long time to finish, and we might get ssh disconnects during
+    the command, this is designed to be able to keep on check on the remote process after the ssh reconnects.
+    hence this function doesn't support retrying of the commands, and aiming at command that can be executed only once.
+    i.e. nodetool decommission or nodetool darin
+
+    The function returns the result of the command, same as remote.run() or remote.sudo() functions.
+    """
+    assert retry == 0, "retry is not supported for long running commands, always use it with retry=0"
+
+    cmd_uuid = uuid4()
+    cmd_stdout_log = Path('/tmp') / f'remoter_stdout_{cmd_uuid}.log'
+    cmd_stderr_log = Path('/tmp') / f'remoter_stderr_{cmd_uuid}.log'
+    cmd_exit_log = Path('/tmp') / f'remoter_exit_{cmd_uuid}.log'
+    cmd_bash = Path('/tmp') / f'remoter_cmd_{cmd_uuid}.sh'
+
+    with ExitStack() as stack:
+        def clear_tmp_files():
+            for log_file in (cmd_stdout_log, cmd_stderr_log, cmd_exit_log, cmd_bash):
+                remoter.run(f'rm {log_file}', verbose=False, ignore_status=True)
+
+        stack.callback(clear_tmp_files)
+        remoter.run(f'echo "{cmd}" > {cmd_bash}')
+
+        start_time = time.perf_counter()
+
+        remote_command = f'(bash -o pipefail {cmd_bash} ; echo $? > {cmd_exit_log}) >& {cmd_stdout_log} 2> {cmd_stderr_log} </dev/null  &  echo $!'
+        if verbose:
+            logger.debug("<%s> execute run_long_running_cmd: %s", remoter.hostname, remote_command)
+        pid = remoter.run(remote_command, retry=0, verbose=verbose).stdout
+
+        remoter.run(f'while [ -e /proc/{pid.strip()} ]; do sleep 0.1; done', timeout=timeout, retry=10, verbose=False)
+
+        stdout = str(remoter.run(f'cat {cmd_stdout_log}', verbose=False).stdout)
+        stderr = str(remoter.run(f'cat {cmd_stderr_log}', verbose=False).stdout)
+        exit_code = int(remoter.run(f'cat {cmd_exit_log}', verbose=False).stdout)
+        result = Result(command=cmd, stdout=stdout,
+                        stderr=stderr, exited=exit_code,
+                        hide=('stderr', 'stdout'), pty=False)
+
+        result.duration = time.perf_counter() - start_time
+        result.exit_status = exit_code  # for compatibility with subprocess.run
+
+        if not ignore_status and exit_code != 0:
+            raise UnexpectedExit(result=result)
+
+        return result
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4994,7 +4994,7 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
 
     def decommission(self, node: BaseNode, timeout: int | float = None):
         with adaptive_timeout(operation=Operations.DECOMMISSION, node=node):
-            node.run_nodetool("decommission", timeout=timeout, retry=0)
+            node.run_nodetool("decommission", timeout=timeout, long_running=True, retry=0)
         self.verify_decommission(node)
 
     @property
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -34,7 +34,7 @@ from importlib import import_module
 from typing import List, Optional, Dict, Union, Set, Iterable, ContextManager, Any, IO, AnyStr, Callable
 from datetime import datetime
 from textwrap import dedent
-from functools import cached_property, wraps, lru_cache
+from functools import cached_property, wraps, lru_cache, partial
 from collections import defaultdict
 from collections.abc import Iterator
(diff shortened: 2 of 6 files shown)


sdcm/remote/remote_long_running.py
```diff
@@ -0,0 +1,44 @@
...
+    cmd_exit_log = Path('/tmp') / f'remoter_exit_{cmd_uuid}.log'
+    cmd_bash = Path('/tmp') / f'remoter_cmd_{cmd_uuid}.sh'
+
+    remoter.run(f'echo "{cmd}" > {cmd_bash}')
+    remote_command = f'(bash +e +x -o pipefail {cmd_bash} ; echo $? > {cmd_exit_log}) >& {cmd_stdout_log} 2> {cmd_stderr_log} </dev/null  &  jobs -p'
+    if verbose:
+        logger.debug("<%s> execute run_long_running_cmd: %s", remoter.hostname, remote_command)
+    pid = remoter.run(remote_command, retry=0, verbose=verbose).stdout
+
+    remoter.run(f'while [ -e /proc/{pid.strip()} ]; do sleep 0.1; done', timeout=timeout, retry=retry, verbose=False)
```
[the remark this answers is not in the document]
fruch: yes the idea that this command could be retried multiple times, and would be limited mostly by timeout

while the actual nodetool command we can't retry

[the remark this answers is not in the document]
soyacz: but this retry mechanism with this implementation doesn't work correctly - in case of network error (we want to mitigate) while loop will be retried only as per invocation (let's say no retry) - in that case we're not protected.

And from the other hand, if we want to retry x times, then we should watch `exit_code` and make decision based on that

sdcm/remote/remote_long_running.py
```diff
@@ -0,0 +1,53 @@
...
+                    stderr=stderr, exited=exit_code,
+                    hide=('stderr', 'stdout'), pty=False)
+
+    result.duration = time.perf_counter() - start_time
+    result.exit_status = exit_code  # for compatibility with subprocess.run
+
+    if not ignore_status and exit_code != 0:
+        raise UnexpectedExit(result=result)
+
+    for log_file in (cmd_stdout_log, cmd_stderr_log, cmd_exit_log, cmd_bash):
```
fruch, line 50: this should go into a finally clause, cause in case of error being raise
those file will accumulate on the host /tmp

fruch: or use ExitStack

fruch: used ExistStack

fruch:
V2:
- used ExitStack for cleanup of tmp files
- added unit tests that works with LOCALRUNNER

sdcm/remote/remote_long_running.py
```diff
@@ -0,0 +1,57 @@
...
+        exit_code = int(remoter.run(f'cat {cmd_exit_log}', verbose=False).stdout)
+        result = Result(command=cmd, stdout=stdout,
+                        stderr=stderr, exited=exit_code,
+                        hide=('stderr', 'stdout'), pty=False)
+
+        result.duration = time.perf_counter() - start_time
+        result.exit_status = exit_code  # for compatibility with subprocess.run
+
+        if not ignore_status and exit_code != 0:
+            raise UnexpectedExit(result=result)
```
soyacz, line 55: I think this part should be wrapped by retry mechanism and retry (using retry from `ron_long_running_cmd` param, which currently is not used).

fruch: well this target commands that should be retried, i.e. if they could, we won't have any issue...

I don't think it's needed, we can remove the retry parameter as far as I'm concerned

i.e. I don't want all of remoter command to start using this, just the one which are long and we can retry the command itself.

fruch: nodetool decomission
nodetool cleanup
nodetool drain
nodetool rebuild
nodetool repair

and that's it...

fruch: and in 2nd pass, might consider the stress command to be like that as well.
just for the sake of leaving the command behind so one can experiment with commands easier

soyacz: ok, but leave info that retry mechanism does not work with long running cmd (or even raise error event so users are aware of that)

as to list of commands, in that case worth to adapt `_wait_for_tablets_balanced` - this also is long (retries when using `run_remoter_curl` can be removed then).

fruch: that can be a followup

sdcm/remote/remote_long_running.py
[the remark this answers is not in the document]
fruch: now hardcode for 10 retries, if we'll find occasion it might happen more than that we could make it higher

sdcm/remote/remote_long_running.py
```diff
@@ -0,0 +1,57 @@
...
+        def clear_tmp_files():
+            for log_file in (cmd_stdout_log, cmd_stderr_log, cmd_exit_log, cmd_bash):
+                remoter.run(f'rm {log_file}', verbose=False, ignore_status=True)
+
+        stack.callback(clear_tmp_files)
+        remoter.run(f'echo "{cmd}" > {cmd_bash}')
+
+        start_time = time.perf_counter()
+
+        remote_command = f'(bash +e +x -o pipefail {cmd_bash} ; echo $? > {cmd_exit_log}) >& {cmd_stdout_log} 2> {cmd_stderr_log} </dev/null  &  jobs -p'
```
soyacz, line 37: `+e +x` - shouldn't it be `-e -x`?

soyacz, line 37: `jobs -p` might bring wrong result if another background job is currently running.
Why not `& echo $!`?

sdcm/remote/remote_long_running.py
```diff
@@ -0,0 +1,57 @@
...
+        if verbose:
+            logger.debug("<%s> execute run_long_running_cmd: %s", remoter.hostname, remote_command)
+        pid = remoter.run(remote_command, retry=0, verbose=verbose).stdout
+
+        remoter.run(f'while [ -e /proc/{pid.strip()} ]; do sleep 0.1; done', timeout=timeout, retry=10, verbose=False)
```
soyacz, line 42: if we reach timeout, command is not stopped - this is different than typical way

fruch: it's not, if we start execution of decommission, it continue even if we timeout or disconnect.
same goes to all of nodetool commands, there no way to stop them once they reached scylla side

sdcm/remote/remote_long_running.py
fruch: we can try it out, for some reason when working on it, it didn't worked as I expected.

fruch: we can enable it on verbose maybe, but we don't want base debug
and we don't want to fail on the first failure, i.e. depends what's the command being passed in, i.e. the user of this code, can enable/disable those flag  as needed.

fruch: removing those, since those are bash defaults anyhow

fruch: changes it, and it seem to be working good enough

sdcm/remote/remote_long_running.py
soyacz: I think `nodetool` command was stopping (I know decommission is not stopped due `nodetool` interruption)

fruch: no sure I understood what do you mean here ? 

today if ssh disconnect nodetool itself gonna die, and the command would continue 

but with this code, since we are running it in the background, ssh disconnect won't kill the nodetool command, and this while loop can be killed multiple times, it won't affect the nodetool command

soyacz: I'm just saying that in case of timeout we should kill `pid` - like normal ssh disconnect would do.
Otherwise, if a command hangs indefinitely it will stay running forever.

fruch: and ? even if we kill it might still be running on scylla end...

it don't think it make much difference at that point, if we really want to cancel such operation, we should be looking into can scylla can provide, it's killing the client isn't enough for most of the cases we are talking about.

soyacz: for these cases not (nodetool) but for others might be problematic - anyway I'm not having now good example, so I'm not insisting. 
Just worth to remember this flaw.

soyacz (approved):
LGTM

sdcm/remote/remote_long_running.py
```diff
@@ -0,0 +1,69 @@
...
+                         verbose: bool = True,
+                         retry: int = 0):
+    """
+    Run a long-running command on the remote host. The command is executed in the background and the function waits
+    for the process of the command to finish.
+
+    this function is useful for running commands that take a long time to finish, and we might get ssh disconnects during
+    the command, this is designed to be able to keep on check on the remote process after the ssh reconnects.
+    hence this function doesn't support retrying of the commands, and aiming at command that can be executed only once.
+    i.e. nodetool decommission or nodetool darin
```
vponomaryov, line 27: `s/darin/drain/` - typo to be changed if you happen to upload new patch-set.

vponomaryov (approved):
LGTM

PR #7838, opened 2024-07-03
perf(tablets): create new jobs for tablet enabled runs

fruch:
### Testing

- [x] https://argus.scylladb.com/view/performance-regression-tests

### Reminders


jenkins-pipelines/performance/scylla-enterprise/scylla-enterprise-perf-regression-throughput-non-shard-aware-i4i-tablets.jenkinsfile
```diff
@@ -0,0 +1,15 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+perfRegressionParallelPipeline(
+    backend: "aws",
+    region: "us-east-1",
+    test_name: "performance_regression_test.PerformanceRegressionTest",
+    test_config: '''["test-cases/performance/perf-regression.100threads.30M-keys-i4i-enterprise.yaml","configurations/perf-loaders-non-shard-aware-config.yaml", "configurations/disable_speculative_retry.yaml"]''',
```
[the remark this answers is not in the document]
fruch: so technically this config should be deleted from master/v15:
configurations/perf-loaders-non-shard-aware-config.yaml

since it's not complete, and need to point to different users / keys 

if we want to disable shared_aware it should be done vi scylla configuration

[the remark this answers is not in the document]
fruch: in the end I'm alter this configuration to disable shard aware with configuration of scylla
and `use_prepared_loaders: true`

fruch:
The situation seem much better now:
https://argus.scylladb.com/view/performance-regression-tests

and results seem to be similar to previous runs on v14 (minus it's without tablets)

I'm ready to get this merged.

fruch:
with on exception, the grow-shrink case now with multi AZ, is much longer, i.e. 3h to add node and 2 hours to decommission a node

maybe we should enable multi-az just on elasticity case, when we parallelize those things ?

jenkins-pipelines/performance/branch-perf-v15/scylla-master/perf-regression/scylla-master-perf-regression-throughput-non-shard-aware-i4i-tablets.jenkinsfile
```diff
@@ -0,0 +1,14 @@
...
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+perfRegressionParallelPipeline(
+    backend: "aws",
+    region: "us-east-1",
+    test_name: "performance_regression_test.PerformanceRegressionTest",
+    test_config: '''["test-cases/performance/perf-regression.100threads.30M-keys-i4i.yaml", "configurations/disable_shared_aware_drivers.yaml", "configurations/disable_speculative_retry.yaml"]''',
+    sub_tests: ["test_write", "test_read", "test_mixed"],
+    test_email_title: "throughput - non-shard-aware / tablets",
```
soyacz, line 12: non-shard-aware / tablets - we need to think if we need such a case - IMHO this case should be tested with tablets aware driver.

sdcm/results_analyze/__init__.py
```diff
@@ -729,18 +729,18 @@ def check_regression(self, test_id, is_gce=False, email_subject_postfix=None,
...
         if not doc:
-            self.log.error('Cannot find test by id: {}!'.format(test_id))
-            return False
+            raise ValueError(f'Cannot find test by id: {test_id}')
         self.log.debug(PP.pformat(doc))
 
         test_stats = self._test_stats(doc)
         if not test_stats:
-            return False
+            raise ValueError(f'Cannot find test by id: {doc.get("_id", "unknown test")}')
```
soyacz, line 737: what is the impact when raising instead of returning False?

utils/build_system/create_test_release_jobs.py
```diff
@@ -75,6 +75,8 @@ def create_directory(self, name: Path | str, display_name: str):
             new_path = str(Path(self.base_job_dir) / name)
 
             if self.jenkins.job_exists(new_path):
+                if name in ['scylla-master', 'scylla-enterprise']:
```
soyacz, line 78: can you add a comment why we do it for future reference?

soyacz (dismissed):
Several minor comments only.
This is huge step forward in organizing these tests, thanks!!
LGTM

jenkins-pipelines/performance/branch-perf-v15/scylla-master/perf-regression/scylla-master-perf-regression-throughput-non-shard-aware-i4i-tablets.jenkinsfile
fruch: This actually a very interesting question 

If we disable shared aware like here, is tablet aware still kicks in ?

jenkins-pipelines/performance/branch-perf-v14/Perf-Regression/scylla-release-perf-regression-latency-shard-aware-1TB.jenkinsfile
```diff
@@ -7,7 +7,7 @@ perfRegressionParallelPipeline(
     backend: "aws",
     region: "us-east-1",
     test_name: "performance_regression_test.PerformanceRegressionTest",
-    test_config: '''["test-cases/performance/perf-regression-latency-1TB.yaml", "configurations/perf-loaders-shard-aware-config.yaml", "configurations/tablets_disabled.yaml"]''',
+    test_config: '''["test-cases/performance/perf-regression-latency-1TB.yaml", "configurations/perf-loaders-shard-aware-config.yaml", "configurations/tablets_disabled.yaml", "configurations/disable_kms.yaml"]''',
```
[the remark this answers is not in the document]
fruch: It's not backported there, and yes KMS isn't available on that branch

sdcm/results_analyze/__init__.py
fruch: I want every failure to be raise as event 

With return false it's losing some part of the context

utils/build_system/create_test_release_jobs.py
fruch: We can probably remove, it's just slow the process a bit, the update of such folder with so many jobs under it, seems to be taking a while.

And I don't want to override anything in a folder that wasn't created with SCT.

jenkins-pipelines/performance/branch-perf-v15/scylla-master/perf-regression/scylla-master-perf-regression-throughput-non-shard-aware-i4i-tablets.jenkinsfile
fruch: after discussing with @roydahan 
we removed the non-shard-aware from the weekly runs

we could run it if need, but from now on, only shard-aware driver would be used

utils/build_system/create_test_release_jobs.py
fruch: added a comment saying exactly this

PR #7847, opened 2024-07-04
fix(nemesis): parallel cleanups in all nemesis using cleanups

fruch:
quite some time ago the `disrupt_nodetool_cleanup` change to work in parallel, but other nemesis that do the same sequence of running cleanup on all nodes (after topology changes) we still doing that in a sequence

this change align it for all of the places cleanup is used.

Note: for tablet cleanup should be no-op, and it's not really matter, but this change is for backporting into older branches

Closes: scylladb/qa-tasks#1690

### Testing

- [x] 🟢   run decommission - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-DecommissionMonkey-aws-test/10
- [x] 🟢  run add remove node - https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-TerminateAndRemoveNodeMonkey-aws-test/2

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1316,11 +1316,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 new_node.set_seed_flag(target_is_seed)
                 self.cluster.update_seed_provider()
             try:
-                test_keyspaces = self.cluster.get_test_keyspaces()
-                for node in self.cluster.nodes:
-                    with adaptive_timeout(Operations.CLEANUP, node=node, timeout=HOUR_IN_SEC * 48):
-                        for keyspace in test_keyspaces:
-                            node.run_nodetool(sub_cmd='cleanup', args=keyspace, retry=0)
+                self.nodetool_cleanup_on_all_nodes_parallel()
             finally:
                 self.unset_current_running_nemesis(new_node)
         return new_node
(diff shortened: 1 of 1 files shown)


soyacz (commented):
but shouldn't it depend on `parallel_node_operations`? 
What if someone uses Scylla 5.4/2024.1 without raft?

fruch:
> but shouldn't it depend on `parallel_node_operations`? What if someone uses Scylla 5.4/2024.1 without raft?

no, it's not related at all

sdcm/nemesis.py
```diff
@@ -3585,11 +3584,7 @@ def remove_node():
             # In order to eliminate cases that we miss a "data loss" bug because of it, we cleanup this data.
             # This fix important when just user profile is run in the test and "keyspace1" doesn't exist.
             try:
-                test_keyspaces = self.cluster.get_test_keyspaces()
-                for node in self.cluster.nodes:
-                    with adaptive_timeout(Operations.CLEANUP, node, timeout=HOUR_IN_SEC * 48):
-                        for keyspace in test_keyspaces:
-                            node.run_nodetool(sub_cmd='cleanup', args=keyspace, retry=0)
+                self.nodetool_cleanup_on_all_nodes_parallel()
```
vponomaryov, line 3587: It stopped using keyspace argument, was it done intentionally?

fruch: yes,
specifying specific keyspaces complicates the work, and might fail on mismatch in schema (i.e. parallel nemesis)

anyhow to instruction for for those procedure is doing full cleanup after topology change (also for system keyspaces)

this change happen more than a year ago, and one might find the whole discussion there.
this PR just port this change into the other relevant places that were forgotten

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

roydahan (approved):

PR #7864, opened 2024-07-07
feature(perf-test): add option for `perf_extra_jobs_to_compare`

fruch:
before we have all kind of logic that prevented us from getting results from staging jobs, and from nested folders

as part of refactoring the perf jobs, we do want to be able to have nested folder, and also the ability to run staging job compare to offical jobs, without naming them exactly the same.

so in this change we introduce a new sct configuration `perf_extra_jobs_to_compare`, offical jobs would set it with thier actual location, so every time it would run in staging it would be comapre with the offical run.

this was introduced also to the `hydra perf-regression-report` so we can manually create reports and needed.

### Testing

- [ ]

### Reminders

### DIFF
--- a/vars/perfRegressionParallelPipeline.groovy
+++ b/vars/perfRegressionParallelPipeline.groovy
@@ -106,6 +106,10 @@ def call(Map pipelineParams) {
             string(defaultValue: '',
                    description: 'Actual user requesting job start, for automated job builds (e.g. through Argus)',
                    name: 'requested_by_user')
+            string(defaultValue: "${pipelineParams.get('perf_extra_jobs_to_compare', '')}",
+                   description: 'jobs to compare performance results with, for example if running in staging, '
+                                + 'we still can compare with official jobs',
+                   name: 'perf_extra_jobs_to_compare')
         }
         options {
             timestamps()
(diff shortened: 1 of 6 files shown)


vars/perfRegressionParallelPipeline.groovy
```diff
@@ -106,6 +106,10 @@ def call(Map pipelineParams) {
             string(defaultValue: '',
                    description: 'Actual user requesting job start, for automated job builds (e.g. through Argus)',
                    name: 'requested_by_user')
+            string(defaultValue: "${pipelineParams.get('perf_extra_jobs_to_compare', '')}",
```
soyacz, line 109: how about setting default to e.g. regression tests (better would be to official runs, but there's too many versions now and would made the report too long)?
This way, running from staging for the first time would produce the report (currently fails due missing reference) and would give also comparison to weekly results.

fruch: that's exactly what I was planning todo
but first need to define their "official" locations

soyacz (commented):
LGTM
@fruch did you test it somewhere? or just cli command?

soyacz (approved):

fruch:
> LGTM
> @fruch did you test it somewhere? or just cli command?

Both cli, and a few runs

PR #1099, opened 2019-07-01
Add longevity-counters-4days-multidc.yaml to `/test-cases`

fruch:
## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I gave variables/functions meaningful self-explanatory names
- [ ] I didn't leave commented-out/debugging code
- [ ] I didn't copy-paste code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (`hydra unit-tests`)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #7911, opened 2024-07-09
fix(byo): make BYO ScyllaDB work with multidc setups

vponomaryov:
`scylla-pkg` project expects the `COPY_AMI_TO_REGIONS` param
to accept a string where region names are separated with comma.

So, translate our region(s) value into the expected format using BYO ScyllaDB.

Also, use `release` build mode
- to reduce redundant actions
- to avoid possible bugs related to 'all' case (less tested case).

### Testing

- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-alternator-3h-multidc-test/6

### Reminders

### DIFF
--- a/vars/byoScylladb.groovy
+++ b/vars/byoScylladb.groovy
@@ -1,4 +1,5 @@
 #!groovy
+import groovy.json.JsonSlurper
 
 def call(Map params, boolean build_image){
     if (! (params.byo_scylla_repo && params.byo_scylla_branch) ) {
@@ -30,6 +31,12 @@ def call(Map params, boolean build_image){
         jobToTrigger = currentJobDirectoryPath + '/' + jobToTrigger[2..-1].trim()
     }
 
+    try {
+        copyAmiToRegions = new JsonSlurper().parseText(params.region)
+        copyAmiToRegions = copyAmiToRegions.join(",").replace(" ", "")
+    } catch(Exception) {
+        copyAmiToRegions = params.region
+    }
     byoParameterList = [
         string(name: 'DEFAULT_PRODUCT', value: params.byo_default_product),
         string(name: 'DEFAULT_BRANCH', value: params.byo_default_branch),
@@ -49,8 +56,7 @@ def call(Map params, boolean build_image){
         string(name: 'RELENG_BRANCH', value: params.byo_default_branch),
         //
         booleanParam(name: 'BUILD_WITH_CMAKE', value: false),
-        // NOTE: 'BUILD_MODE' must be empty string for '5.4'/'2024.1' and older branches
-        string(name: 'BUILD_MODE', value: 'ALL'),
+        string(name: 'BUILD_MODE', value: 'release'),
         booleanParam(name: 'CODE_COVERAGE', value: false),
         booleanParam(name: 'TEST_DEBUG_INFO', value: false),
         // SPECIAL_CONFIGURE_PY_PARAMS = '' // TODO: add it's support?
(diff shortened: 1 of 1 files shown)


vponomaryov:
Example of the error without this fix: https://jenkins.scylladb.com/job/enterprise-2024.1/job/byo/job/byo_build_tests_dtest/15/consoleFull

vars/byoScylladb.groovy
```diff
@@ -1,5 +1,22 @@
 #!groovy
 
+def list_as_comma_separated_str(input) {
+    // NOTE: transform all possible single-/multi region values into the scylla-pkg compatible format:
```
fruch, line 4: isn't easier to read it as json, and then format it with comma separated string ?

vponomaryov: It supports 4 structures:
- list
- string with pickled list
- comma separated regions
- space separated regions

fruch: but why we need to support all those variants that we aren't actually using

fruch: i.e. SCT doesn't support space/comma separated for regions

vponomaryov: > but why we need to support all those variants that we aren't actually using

```
$ git grep "region_name" jenkins-pipelines/
...
test-cases/features/add-new-dc.yaml:region_name: 'eu-west-1 eu-west-2'
test-cases/longevity/longevity-ndbench-12-nodes-multidc-4h.yaml:region_name: 'us-east-1 eu-west-1'
test-cases/longevity/longevity-ndbench-192-nodes-multidc-4h.yaml:region_name: 'us-east-1 eu-west-1'
...
test-cases/manager/manager-regression-multiDC-set-distro.yaml:region_name: 'us-east-1 us-west-2'
...
test-cases/scylla-operator/longevity-scylla-operator-multidc-12h.yaml:region_name: 'eu-north-1 eu-west-1'
test-cases/scylla-operator/longevity-scylla-operator-multidc-3h.yaml:region_name: 'eu-north-1 eu-west-1'

```

fruch: as you can see all those examples are in the test cases yamls, 
none of them in the pipelines, i.e. it won't work in the pipelines

since we need to pass it to SCT via environment variables, and we don't support all those variation for reading from env variables.

we are reading back from test-cases into the pipeline, but most for calculation of the pipeline timeouts, I don't think we read back the region, it's working only base on pipeline definitions

fruch: anyhow you test it, and pass space separated region into one of our pipe, and see if it works. I'm betting it's not gonna work.

vponomaryov: `vars/getJenkinsLabels.groovy` fails trying to do the same thing you proposed.
One of us added this limitation here: https://github.com/scylladb/scylla-cluster-tests/pull/1099/files
From my point of view it is unneeded restriction.
I really would like to use `foo, bar, quuz` approach like we do with AMI ids and other parameters.

fruch: As you can see now it can't be used, and pipeline currently assumes one specific option regarding the region parameter

SCT configuration is using multiple conflicting options to pass in lists, we can raise an issue to attend to that.

Anyhow I would want to support only one option if possible per a specific configuration option, I don't groovy/python variation for all of those.

vponomaryov: I added it as you requested but needed to add try-except block to support single-dc case with just name as string.

vponomaryov:
Tested here:
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-alternator-3h-multidc-test/15
- https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/valerii/job/vp-longevity-10gb-3h-test/113

`scylla-pkg` side fix is here:
- https://github.com/scylladb/scylla-pkg/pull/4218

fruch (approved):
LGTM

vars/byoScylladb.groovy
fruch: We can have an issue for supporting all the options for lists, but we'll need to do it across all of the relevant values, in both ends, the pipelines and SCT.

For example ditch the env variable (or keep them for development only), and generate an extra configuration yaml file in pipelines ?

Anyhow let's move the discussion into a separate issue.

PR #7919, opened 2024-07-10
feature(ruff): enable PLR0916 and preview mode

fruch:
PLR016 is needed, since we had similar check in pylint, and we hit it in backport to older branches

also the preview enable this feature for this visual feature for the CLI:

```
sdcm/tester.py:838:13: PLR0916 Too many Boolean expressions (6 > 5)
    |
837 |       def prepare_kms_host(self) -> None:
838 |           if (self.params.is_enterprise and ComparableScyllaVersion(self.params.scylla_version) >= '2023.1.3'
    |  _____________^
839 | |             and self.params.get('cluster_backend') == 'aws'
840 | |             and not self.params.get('scylla_encryption_options')
841 | |             and not self.params.get('enterprise_disable_kms')
842 | |             and self.params.get("db_type") != "mixed_scylla"  # oracle probably doesn't support KMS
    | |____________________________________________________________^ PLR0916
```

[skip-lint]

### Testing

- [ ]

### Reminders

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -5,7 +5,7 @@ module.exports = {
         // Subject
         'subject-empty': [2, 'never'],
         'subject-full-stop': [2, 'never', '.'],
-        'subject-max-length': [2, 'always', 100],
+        'subject-max-length': [2, 'always', 120],
         'subject-min-length': [2, 'always', 10],
         // Type
         'type-enum': [2,'always',['ci','docs','feature','fix','improvement','perf','refactor','revert','style','test', 'unit-test', 'build']],
@@ -15,7 +15,7 @@ module.exports = {
         'scope-min-length': [2, 'always', 3],
         // Body
         'body-min-length': [2, 'always', 30],
-        'body-max-line-length': [2, 'always', 100],
+        'body-max-line-length': [2, 'always', 120],
         'body-leading-blank': [2, 'always'],
     },
     parserPreset: {
--- a/.github/search_commits.py
+++ b/.github/search_commits.py
@@ -27,7 +27,7 @@ def get_parser():
     return parser.parse_args()
 
 
-def main():  # pylint: disable=too-many-locals
+def main():  # pylint: disable=too-many-locals  # noqa: PLR0914
     args = get_parser()
     github = Github(github_token)
     repo = github.get_repo(args.repository, lazy=False)
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,7 +41,7 @@ repos:
 
     -   id: ruff
         name: ruff
-        entry: ruff check --fix
+        entry: ruff check --fix --preview
         language: system
         types: [python]
 
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -95,7 +95,7 @@ def test_single_operator_image_tag_is_everywhere(db_cluster):
 
 
 @pytest.mark.required_operator("v1.11.0")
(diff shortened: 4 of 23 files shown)


commitlint.config.js
```diff
@@ -15,9 +15,10 @@ module.exports = {
         'scope-min-length': [2, 'always', 3],
         // Body
         'body-min-length': [2, 'always', 30],
-        'body-max-line-length': [2, 'always', 100],
+        'body-max-line-length': [2, 'always', 130],
         'body-leading-blank': [2, 'always'],
     },
+    ignores: [ (message) => message.includes('[skip-lint]') ],
```
soyacz, line 21: how does it work?

soyacz (commented):
Did you try running `ruff .` after enabling this boolean check?

fruch:
> Did you try running `ruff .` after enabling this boolean check?

wanted to see it in action in CI

now I'll fix it, on it's own commit (that needs to be backported)

commitlint.config.js
fruch: this one enable us to skip the commit lint, maybe it's need to be renamed.

the idea is that some commit message, like ones having backtraces, enforcing 100 max body line length is too much
this give a way to skip it, in certain cases.

yet again, we might just remove the line length limit  for body

commitlint.config.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
         // Subject
         'subject-empty': [2, 'never'],
         'subject-full-stop': [2, 'never', '.'],
-        'subject-max-length': [2, 'always', 100],
+        'subject-max-length': [2, 'always', 130],
```
vponomaryov, line 8: Why is it changed?

commitlint.config.js
```diff
@@ -15,9 +15,10 @@ module.exports = {
         'scope-min-length': [2, 'always', 3],
         // Body
         'body-min-length': [2, 'always', 30],
-        'body-max-line-length': [2, 'always', 100],
+        'body-max-line-length': [2, 'always', 130],
```
vponomaryov, line 18: Same here, why is it changed?

commitlint.config.js
vponomaryov: I disagree to increase the message body limit.
I used line continuations in such cases.

The problem is that lots of people will start abusing this limit absence in common cases.

All lengths must fit half of the 1920p screen, IMHO.

fruch: I'm tired of manually edit backtrace to fit, and anyhow I'm putting it with triple backtick.

which make it readable just fine in github 

half screen ? where did you came up with that ?

other repo we don't have those limitation, and we are just find with it., people are no abusing them in that sense.

my requirement is, I want to be able to paste backtrace right into the commit.
and in some case it's important to see it all (like the example in this PR)

commitlint.config.js
fruch: I want it bigger, checked with my 13" screen, to see when github gonna move to the next line.

commitlint.config.js
fruch: I want it bigger, checked with my 13" screen, to see when github gonna move to the next line.

commitlint.config.js
vponomaryov: > half screen ? where did you came up with that ?

Split screen approach.

commitlint.config.js
vponomaryov: Some time ago we agree to 120 symbols. Can we stick to this value?

fruch: we can stick to it,

next time, if this keep wasting time for me, I'll gonna completely remove the commitlint 
it doesn't really contribute to people writing good commit messages...

PR #7920, opened 2024-07-10
improvement(ruff): add f-string logs check

soyacz:
We miss f-string log checks in ruff and cause issues with backporting (and anyway we want to keep it).

Fixed along issues with it in most of cases.

### Testing

- [ ]

### Reminders

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -191,7 +191,7 @@ class BackupFunctionsMixIn(LoaderUtilsMixin):
         if truncate:
             for ks, tables in ks_tables_map.items():
                 for table_name in tables:
-                    self.log.info(f'running truncate on {ks}.{table_name}')
+                    self.log.info('running truncate on %s.%s', ks, table_name)
                     self.db_cluster.nodes[0].run_cqlsh(f'TRUNCATE {ks}.{table_name}')
         if restore_data_with_task:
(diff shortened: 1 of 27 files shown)


fruch:
I'm, not completely sure we want to keep (for example for dtest we don't enforce it)
the speed up from it isn't that great (or relevant to SCT), and the usability vs. f-string , kind of sucks..

we could disable it on the branches... 

I'm not sure

soyacz:
> I'm, not completely sure we want to keep (for example for dtest we don't enforce it) the speed up from it isn't that great (or relevant to SCT), and the usability vs. f-string , kind of sucks..
> 
> we could disable it on the branches...
> 
> I'm not sure

I wasn't taking decision to enable it, so I conformed.
I agree this is not relevant to SCT as anyway we always run with debug logs (and we don't use trace logs).

Let's other maintainers take the side

roydahan:
I like f-strings and hate this stupid lint check to replace it.

fruch:
> I like f-strings and hate this stupid lint check to replace it.

ruff is even **nicer**, he know to replace it automatically, not just to warn you about it.

vponomaryov (approved):
LGTM

fruch:
@soyacz remember to close this, and disable this check on releases pylint

soyacz:
closing in favor of https://github.com/scylladb/scylla-cluster-tests/pull/7957

Issue #7884, opened 2024-07-08
If BYO step fails, the pipeline should fail right away.


roydahan:
Seems like failures of the BYO step, are not failing the whole pipeline, and we end up starting the test
and failing like the following:
```
assert all(self.get(o) for o in options_must_exist), \
14:54:40  < t:2024-07-08 11:54:40,199 f:log.py          l:23   c:sdcm.utils.log       p:ERROR > AssertionError: scylla version/repos wasn't configured correctly
14:54:40  < t:2024-07-08 11:54:40,199 f:log.py          l:23   c:sdcm.utils.log       p:ERROR > configure those options: ['ami_id_db_scylla']
14:54:40  < t:2024-07-08 11:54:40,199 f:log.py          l:23   c:sdcm.utils.log       p:ERROR > and those environment variables: ['SCT_AMI_ID_DB_SCYLLA']
```

Seen in: https://jenkins.scylladb.com/job/enterprise-2024.1/job/reproducers/job/longevity-lwt-multi-dc-Tesla/4/

vponomaryov:
The bug topic must be `If BYO step fails fail the pipeline right away`.

The error you got is caused by the fact of the BYO failure:

```
14:51:43  Also:   org.jenkinsci.plugins.workflow.actions.ErrorAction$ErrorId: 662af161-7966-456c-b915-84eac0c1054d
14:51:43  groovy.lang.MissingPropertyException: No such property: jobResults for class: byoScylladb
14:51:43  	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.unwrap(ScriptBytecodeAdapter.java:66)
14:51:43  	at org.codehaus.groovy.runtime.ScriptBytecodeAdapter.getProperty(ScriptBytecodeAdapter.java:471)
14:51:43  	at PluginClassLoader for script-security//org.kohsuke.groovy.sandbox.impl.Checker$7.call(Checker.java:377)
14:51:43  	at PluginClassLoader for script-security//org.kohsuke.groovy.sandbox.GroovyInterceptor.onGetProperty(GroovyInterceptor.java:68)
14:51:43  	at PluginClassLoader for script-security//org.jenkinsci.plugins.scriptsecurity.sandbox.groovy.SandboxInterceptor.onGetProperty(SandboxInterceptor.java:347)
14:51:43  	at PluginClassLoader for script-security//org.kohsuke.groovy.sandbox.impl.Checker$7.call(Checker.java:375)
[...]

PR #7922, opened 2024-07-10
improvement(byo): make CI jobs fail fast if BYO fails

vponomaryov:
Fixes: #7884

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-10gb-3h-test/110

### Reminders

### DIFF
--- a/vars/longevityPipeline.groovy
+++ b/vars/longevityPipeline.groovy
@@ -249,6 +249,18 @@ def call(Map pipelineParams) {
                         }
                     }
                 }
+                post{
+                    failure {
+                        script{
+                            sh "exit 1"
+                        }
+                    }
+                    unstable {
+                        script{
+                            sh "exit 1"
+                        }
+                    }
+                }
             }
             stage('Create SCT Runner') {
                 steps {


fruch:
@vponomaryov what about the rolling upgrade ?

vponomaryov:
> @vponomaryov what about the rolling upgrade ?

Current pipeline structure for rolling upgrade is different and the same logic cannot be used:

```
java.lang.NoSuchMethodError: No such DSL method 'steps' found among steps [...]
```

If moving BYO above, then Argus test run start should be moved, but we should have Argus test run per unique upgrade combination.

fruch:
> > @vponomaryov what about the rolling upgrade ?
> 
> Current pipeline structure for rolling upgrade is different and the same logic cannot be used:
> 
> ```
> java.lang.NoSuchMethodError: No such DSL method 'steps' found among steps [...]
> ```
> 
> If moving BYO above, then Argus test run start should be moved, but we should have Argus test run per unique upgrade combination.

so a fix for rolling-update would be on a separate PR ?

fruch:
@vponomaryov  wait a minute, on the rolling upgrade case, we build each split it's own BYO ?

that's something we missed, we should build it before we split in parallel run, i.e. the whole idea there is to upgrade into same version in of the splits...

vponomaryov:
> @vponomaryov wait a minute, on the rolling upgrade case, we build each split it's own BYO ?

As of now - yes.

> that's something we missed, we should build it before we split in parallel run, i.e. the whole idea there is to upgrade into same version in of the splits...

Then I will need to move it out of the Argus test run precedence.

fruch (approved):

fruch:
> > so a fix for rolling-update would be on a separate PR ?
> 
> Probably, will add it here.
> 
> > @vponomaryov wait a minute, on the rolling upgrade case, we build each split it's own BYO ?
> 
> As of now - yes.
> 
> > that's something we missed, we should build it before we split in parallel run, i.e. the whole idea there is to upgrade into same version in of the splits...
> 
> Then I will need to move it out of the Argus test run precedence.

Let's raise as an issue so we don't forget it.

But building 3 times in one job is going to be quite excessive (i.e we only implemented the target version, not the base)

vponomaryov:
@fruch 

As I wrote above I planned to update this PR.
Since, this one is merged already, created another PR for the upgrade part:
- https://github.com/scylladb/scylla-cluster-tests/pull/7939

PR #7956, opened 2024-07-11
feature(python-uv): replacement for pip and pip-compile

fruch:
using the new pip written in rust

building out requirement.txt take less then a second (with warn cache) installing the packaging inside the docker build, took ~16s

Ref: https://github.com/astral-sh/uv

### Testing

- [ ]

### Reminders


docker/env/build_n_push.sh
```diff
@@ -25,7 +25,7 @@ else
     echo "Hydra image with version $VERSION not found locally. Building..."
     cd "${DOCKER_ENV_DIR}"
     REQUIREMENTS_IN=$(realpath --relative-to=${DOCKER_ENV_DIR} ${SCT_DIR}/requirements.in)
-    pip-compile $REQUIREMENTS_IN --allow-unsafe --generate-hashes --resolver=backtracking > ${SCT_DIR}/${PY_PREREQS_FILE}
+    uv pip compile $REQUIREMENTS_IN --generate-hashes > ${SCT_DIR}/${PY_PREREQS_FILE}
```
soyacz, line 28: need to also update requirements.txt in this way to be in line. Also neest to update dev setup docs

vponomaryov: The `requirements.txt` gets updated exactly here, it is referenced by the `PY_PREREQS_FILE` env var.

requirements.txt
```diff
@@ -218,34 +214,28 @@ boto3==1.28.4 \
     # via
     #   -r requirements.in
     #   gimme-aws-creds
-boto3-stubs[dynamodb,ec2,pricing,s3]==1.28.4 \
+boto3-stubs==1.28.4 \
```
fruch, line 217: one concern that we loose the install groups.

not sure if it's critical

fruch:
running provision test for one more round
1st round, AWS/Azure ones passed, GCE got spot termination, and docker failed (not sure why)

fruch:
> running provision test for one more round 1st round, AWS/Azure ones passed, GCE got spot termination, and docker failed (not sure why)

2nd round of tests, worked better, and unittest showed we pick very old version of `prettytable`, and we need to update to newer version that doesn't generate syntax errors in runtime

fruch:
3rd round of test was working fine (k8s-local-kind-aws failing cause of unrelated issues)

soyacz (commented):
how about updating `docs/install-local-env.md`?

soyacz (approved):
LGTM

soyacz:
@fruch I think we need to backport it to all releases branches. Otherwise all hydra images updates PR's won't be consistent.
And we need to update e.g. Argus client there.
WDYT?

fruch:
> @fruch I think we need to backport it to all releases branches. Otherwise all hydra images updates PR's won't be consistent.
> And we need to update e.g. Argus client there.
> WDYT?

That's fine, but still on some releases we will need to build their own images

soyacz:
> > @fruch I think we need to backport it to all releases branches. Otherwise all hydra images updates PR's won't be consistent.
> > And we need to update e.g. Argus client there.
> > WDYT?
> 
> That's fine, but still on some releases we will need to build their own images

Which releases? In that case I'd rather not doing that - too many changes/updates.

fruch:
> > > @fruch I think we need to backport it to all releases branches. Otherwise all hydra images updates PR's won't be consistent.
> > > And we need to update e.g. Argus client there.
> > > WDYT?
> > 
> > That's fine, but still on some releases we will need to build their own images
> 
> Which releases? In that case I'd rather not doing that - too many changes/updates.

If you want to align the Argus client across all releases, there's a price to pay.

It's not that bad, the critical packages we need are pinned, so it's not that risky, it's just manual labor.

soyacz:
> 
> If you want to align the Argus client across all releases, there's a price to pay.
> 

This is the question - shall we align Argus client in all releases? I don't know what's the approach here...

PR #7957, opened 2024-07-12
improvement(pylint): disable f-string-interpolation error

soyacz:
In SCT we always run with debug log thus we don't care about having f-strings in log messages. There's no improvement from having this check turned on and f-string based messages are more readable and more natural to python programmers.
This check is disabled in our `ruff` linter but still we often backport commits to different branches where we still use it.

Disabling this check in .pylintrc.

### Testing

- [ ]

### Reminders

### DIFF
--- a/pylintrc
+++ b/pylintrc
@@ -147,6 +147,7 @@ disable=print-statement,
         bad-continuation,
         cyclic-import,
         duplicate-code,
+        logging-fstring-interpolation,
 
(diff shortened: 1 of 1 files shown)


fruch:
Maybe after the merged and backported, we should delete this file from the master branch.

I've forgot about when introducing ruff

fruch (approved):
LGTM

PR #8021, opened 2024-07-15
feature(sct_config): refactor to use pydantic

fruch:
### Testing

- [ ]

### Reminders


sdcm/sct_config.py
```diff
@@ -2115,34 +2477,36 @@ def _dotted_get(self, key: str):
...
+
+        def no_op(x):
+            return x
+
+        for annotation in field.metadata:
+            if isinstance(annotation, BeforeValidator):
+                from_env_func = annotation.func
+                break
+        else:
+            from_env_func = no_op
```
soyacz, line 2491: this repeats, could be moved to function `get_field_cast_function` or smthing like this.

sdcm/sct_config.py
```diff
@@ -2054,51 +2400,67 @@ def get_config_option(cls, name):
...
+
+            def no_op(x):
+                return x
+
+            for annotation in field.metadata:
+                if isinstance(annotation, BeforeValidator):
+                    from_env_func = annotation.func
+                    break
+            else:
+                from_env_func = no_op
```
soyacz, line 2424: are you sure no_op is right here? Env vars will always return strings - shouldn't be casted to annotated one?

fruch: this is when there is no annotation we can work with

fruch: so every annotation we'll have can be assumed to convert string to anything.

for example if I put there other pydantic structure ? or `BaseNode` (just example)

defaults/test_default.yaml
```diff
@@ -34,7 +34,7 @@ cluster_health_check: true
 
 add_node_cnt: 1
 
-reuse_cluster: false
```
fruch, line 37: this is a bug, regardless of this PR

sdcm/sct_config.py
```diff
@@ -135,7 +143,10 @@ def int_or_list(value):
     raise ValueError("{} isn't int or list".format(value))
 
 
-def dict_or_str(value):
+IntOrList = Annotated[int | list[int], BeforeValidator(int_or_list)]
```
fruch, line 146: we might find better name to those types

sdcm/sct_config.py
```diff
@@ -169,1450 +186,1774 @@ def boolean(value):
...
+    return Field(*args, **kwargs)
 
-        dict(name="stress_cmd_lwt_mixed", env="SCT_STRESS_CMD_LWT_MIXED",
-             type=str,
-             help="Stress command for LWT performance test for mixed lwt load"),
 
-        dict(name="stress_cmd_lwt_mixed_baseline", env="SCT_STRESS_CMD_LWT_MIXED_BASELINE",
-             type=str,
-             help="Stress command for LWT performance test for mixed lwt load baseline"),
+available_backends: list[str] = [
```
fruch, line 198: this move away from the class, keeping any properties which are not part of pydantic model, create lots of boilerplate to find a way to ignore them during translation to yaml or other formats

sdcm/sct_config.py
```diff
@@ -169,1450 +186,1774 @@ def boolean(value):
...
 
-        dict(name="sct_public_ip", env="SCT_SCT_PUBLIC_IP", type=str,
-             help="""
-                Override the default hostname address of the sct test runner,
-                for the monitoring of the Nemesis.
-                can only work out of the box in AWS
-             """),
-        dict(name="sct_ngrok_name", env="SCT_NGROK_NAME", type=str,
-             help="""
+    multi_region_params: Annotated[list[str], IgnoredType] = Field(default=[], exclude=True)
```
fruch, line 216: here is example of two possible ways of ignoring `IgnoredType` and `Field(exclude=True)`

both are a bit long, and not completely sure behave similarly

sdcm/sct_config.py
```diff
@@ -169,1450 +186,1774 @@ def boolean(value):
...
+    regions_data: Annotated[dict[str, dict[str, str]], IgnoredType] = Field(default={}, exclude=True)
+
+    artifact_scylla_version: str | None = None
+    is_enterprise: bool = False
+
+    config_files: StringOrList = sct_field(
```
fruch, line 222: lots of changes with the help of copilot, it was a bit faster.

I'm not sure if we want to ditch the `env` definitions, and just assume names are identical always, and drop all of them

sdcm/sct_config.py
```diff
@@ -2533,27 +2896,27 @@ def update_argus_with_version(self, scylla_version: str, package_name: str):
...
-        # TODO: automate the process if we gonna keep using them more, or replace the whole configuration with pydantic/dataclasses
-        if kafka_connectors := self.get('kafka_connectors'):
-            out['kafka_connectors'] = [connector.dict(by_alias=True, exclude_none=True)
-                                       for connector in kafka_connectors]
-        if mgmt_restore_params := self.get("mgmt_restore_params"):
-            out["mgmt_restore_params"] = mgmt_restore_params.dict(by_alias=True, exclude_none=True)
-        if mgmt_agent_backup_config := self.get("mgmt_agent_backup_config"):
-            out["mgmt_agent_backup_config"] = mgmt_agent_backup_config.dict(by_alias=True, exclude_none=True)
-        return out
+    # def dict(self):
```
fruch, line 2899: this is still something trick left to handle, the configuration which need to be translated to pydantic

the code as it's own' probably doesn't support it correctly now, i.e. it won't translate from string to pedantic model, and not from dict/yaml

fruch:
@copilot can you take over merge from master and solve the conflicts here ?, and apply the same treatment to all new configurations in sct_config.py

PR #8041, opened 2024-07-17
fix(kms): disable enterprise file stream

fruch:
since we have a problem for now to enable both features we disable file based stream, when ever we enable KMS

all other EaR cases, would need to handle it on their own assuming they would want to enable tablets

Ref: scylladb/scylla-enterprise#4246

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -88,6 +88,7 @@ from sdcm.utils.git import get_git_commit_id, get_git_status_info
 from sdcm.utils.ldap import LDAP_USERS, LDAP_PASSWORD, LDAP_ROLE, LDAP_BASE_OBJECT, \
     LdapConfigurationError, LdapServerType
 from sdcm.utils.log import configure_logging, handle_exception
+from sdcm.utils.issues import SkipPerIssues
 from sdcm.db_stats import PrometheusDBStats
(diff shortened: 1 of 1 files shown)


sdcm/tester.py
```diff
@@ -906,6 +907,12 @@ def prepare_kms_host(self) -> None:
             'key_provider': 'KmsKeyProviderFactory',
             'kms_host': kms_host,
         }
+
+        if SkipPerIssues("scylladb/scylla-enterprise#4246", self.params):
```
fruch, line 911: @mykaul 

as long as this issue would be open, we'll be disabling file streaming, on all places we by default enabled KMS

mykaul (approved):

sdcm/tester.py
```diff
@@ -906,6 +907,12 @@ def prepare_kms_host(self) -> None:
...
+            # until https://github.com/scylladb/scylla-enterprise/issues/4246 would be solved
+            # we should disable the enterprise file streaming feature
+            append_scylla_yaml['enable_file_stream'] = False
```
vponomaryov, line 914: Will this option work for any supported Scylla enterprise version?
2023.1 ?

fruch: No need, older versions doesn't have this option, so they won't do anything.

PR #8085, opened 2024-07-22
fix(sdcm/reporting/tooling_reporting): Add exception handler for Argus

k0machi:
This fixes an issue where a test might stop if we can't report a tooling
version to argus.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

soyacz:
Do we ever want argus to fail SCT test?
If not, maybe worth to introduce https://stackoverflow.com/questions/6598053/python-global-exception-handling for `ArgusClientError`?

fruch:
> Do we ever want argus to fail SCT test? If not, maybe worth to introduce https://stackoverflow.com/questions/6598053/python-global-exception-handling for `ArgusClientError`?

is that the only type of exception it can throw ?

PR #8109, opened 2024-07-23
feature(grow-shrink): option to set the instance type for grow-shrink

fruch:
the commit introduces `nemesis_grow_shrink_instance_type` so one can select the instance type that would be use during the grow-shrink nemesis

when this option is used, the shrink phase is going to decomission exactly the node that was added, so we will leave the cluster in the same state (number of nodes, and their sizes)

Closes: scylladb/qa-tasks#1738

### Testing

- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-5gb-1h-GrowShrinkClusterNemesis-aws-test/7/ [Argus](https://argus.scylladb.com/test/bc471848-79b2-4321-b3b9-e78a5cb67563/runs?additionalRuns[]=b46f43bf-f007-489c-b4bf-7071b3257e4f)
  test with i4i.2xlarge grow with i4i.large, with 1 node, and with 3 nodes

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1265,13 +1265,14 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         InfoEvent(message="FinishEvent - New Node is up and normal").publish()
         return new_node
 
-    def _add_and_init_new_cluster_nodes(self, count, timeout=MAX_TIME_WAIT_FOR_NEW_NODE_UP, rack=None) -> list[BaseNode]:
+    def _add_and_init_new_cluster_nodes(self, count, timeout=MAX_TIME_WAIT_FOR_NEW_NODE_UP, rack=None, instance_type: str = None) -> list[BaseNode]:
         if rack is None and self._is_it_on_kubernetes():
             rack = 0
         self.log.info("Adding %s new nodes to cluster...", count)
         InfoEvent(message=f'StartEvent - Adding {count} new nodes to cluster').publish()
         new_nodes = self.cluster.add_nodes(
-            count=count, dc_idx=self.target_node.dc_idx, enable_auto_bootstrap=True, rack=rack)
+            count=count, dc_idx=self.target_node.dc_idx, enable_auto_bootstrap=True, rack=rack,
+            instance_type=instance_type)
         self.monitoring_set.reconfigure_scylla_monitoring()
         for new_node in new_nodes:
             self.set_current_running_nemesis(node=new_node)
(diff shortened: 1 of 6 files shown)


fruch:
tested with:

```yaml
nemesis_grow_shrink_instance_type:  'i4i.large'
nemesis_add_node_cnt: 3
```

fruch:
@temichus please validate this fit the use cases you'll need in context of "serverless v2" work

soyacz (approved):
LGTM

sdcm/nemesis.py
```diff
@@ -4080,10 +4086,13 @@ def disrupt_grow_shrink_cluster(self):
         if not self.has_steady_run and sleep_time_between_ops:
             self.steady_state_latency()
             self.has_steady_run = True
-        self._grow_cluster(rack=None)
+        new_nodes = self._grow_cluster(rack=None)
+
+        # pass on the exact nodes only if we have specific types for them
+        new_nodes = new_nodes if self.tester.params.get('nemesis_grow_shrink_instance_type') else None
```
temichus, line 4092: maybe we should support 2 parameters 

'nemesis_grow_instance_type'
'nemesis_shrink_instance_type'

because the current grow_shrink logic works like

1) setup cluster, nodes: a,b,c
2) grow, nodes a,b,c,d,e,f
3) shrink, nodes d,e,f

In the end, we have only new nodes. but with hawing nemesis grow shrink instance_type we will be able to grow_shrink only exactly the same nodded without the ability to touch existing

sdcm/nemesis.py
```diff
@@ -4043,19 +4044,24 @@ def _decommission(node):
         InfoEvent(f'FinishEvent - ShrinkCluster has done decommissioning {len(nodes)} nodes').publish()
 
     def _decommission_nodes(self, nodes_number, rack, is_seed: Optional[Union[bool, DefaultValue]] = DefaultValue,
-                            dc_idx: Optional[int] = None):
+                            dc_idx: Optional[int] = None, exact_nodes: list[BaseNode] | None = None):
```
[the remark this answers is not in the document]
fruch: we do gonna put both parameter at the same time, one is overriding the other
I don't think copying it would be more helpful in this case

so far we didn't had any requirement for this nemesis to no decommission the same amount of nodes

sdcm/nemesis.py
fruch: if one want a full specific sequence, he can build a specific nemesis, or a specific test case.

I don't want to overload much more on this poor nemesis.

roydahan: I agree that the config is fine as is, we cannot make it to granular just to cover one test scenario that is probably not that interesting for the long run.

roydahan (approved):

sdcm/nemesis.py
temichus: but you will change the current logic of the test/menesis 

instead of having only new nodes at the end of the test, it will be old(initial) ones

soyacz: and that seems to be correct, assuming we design the test for specific node size, leaving cluster after this nemesis with smaller nodes will cause failures due to overload.

PR #8110, opened 2024-07-23
fix(upgrade-tests): skip counter test if tablets enabled

fruch:
part of the rolling upgrade we create and test some tables after each step, one of those is `counters_test: Validate counter support` since counter isn't supported by tablets, we need to skip it once tablets is enabled.

### Testing

- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-debian11-test/11/ [Argus](https://argus.scylladb.com/test/94307779-73d9-4e4b-8c69-de3431a72ccd/runs?additionalRuns[]=8766ab65-1074-48cc-bb24-ed9cb1fddf39)
- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-debian11-test/12/ [Argus](https://argus.scylladb.com/test/94307779-73d9-4e4b-8c69-de3431a72ccd/runs?additionalRuns[]=8766ab65-1074-48cc-bb24-ed9cb1fddf39)

### Reminders

### DIFF
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -435,6 +435,7 @@ class FillDatabaseData(ClusterTester):
             ],
             'min_version': '1.7',
             'max_version': '',
+            'skip_condition': 'not self.is_counter_supported',
             'skip': ''},
         {
             'name': 'indexed_with_eq_test: Check that you can query for an indexed column even with a key EQ clause',
@@ -3180,6 +3181,12 @@ class FillDatabaseData(ClusterTester):
     @cached_property
     def enable_cdc_for_tables(self) -> bool:
         if self.tablets_enabled and SkipPerIssues(issues="https://github.com/scylladb/scylladb/issues/16317", params=self.params):
+            return False
+        return True
+
+    @cached_property
+    def is_counter_supported(self) -> bool:
+        if self.tablets_enabled and SkipPerIssues(issues="scylladb/scylladb#18180", params=self.params):
             return False
         return True
 


soyacz (commented):
after some time we'll need to enable it back.
Can we apply here `SkipPerIssue` maybe?

yarongilor:
why not creating counters before tablets is enabled?

roydahan:
> why not creating counters before tablets is enabled?

The whole idea is that tablets don't support counters.

yarongilor:
> > why not creating counters before tablets is enabled?
> 
> The whole idea is that tablets don't support counters.

right, the idea is having a keyspace created as non-tablets, so it supports counters even after an upgrade to a tablets-enabled version. it should not have any impact of such a keyspace.

fruch:
> > > why not creating counters before tablets is enabled?
> > 
> > 
> > The whole idea is that tablets don't support counters.
> 
> right, the idea is having a keyspace created as non-tablets, so it supports counters even after an upgrade to a tablets-enabled version. it should not have any impact of such a keyspace.

you are free to implement any case you think for tablets, I'm just trying to get rolling upgrade working, cause no one went over all the fallout tablets generates and attend to it.

fruch:
> after some time we'll need to enable it back. Can we apply here `SkipPerIssue` maybe?

I guess we could use: https://github.com/scylladb/scylladb/issues/18180

vponomaryov (approved):
LGTM

Issue #7470, opened 2024-05-22
Log collection stage can fail on collection of jepsen data, on test which aren't running jepsen


fruch:
```
13:59:30  Traceback (most recent call last):
13:59:30    File "/home/ubuntu/scylla-cluster-tests/./sct.py", line 1797, in <module>
13:59:30      cli.main(prog_name="hydra")
13:59:30    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
13:59:30      rv = self.invoke(ctx)
13:59:30    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
13:59:30      return _process_result(sub_ctx.command.invoke(sub_ctx))
13:59:30    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
13:59:30      return ctx.invoke(self.callback, **ctx.params)
13:59:30    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
13:59:30      return __callback(*args, **kwargs)
13:59:30    File "/home/ubuntu/scylla-cluster-tests/./sct.py", line 1211, in collect_logs
13:59:30      collected_logs = collector.run()
13:59:30    File "/home/ubuntu/scylla-cluster-tests/sdcm/logcollector.py", line 1678, in run
13:59:30      if result := log_collector.collect_logs(local_search_path=local_dir_with_logs):
13:59:30    File "/home/ubuntu/scylla-cluster-tests/sdcm/logcollector.py", line 1294, in collect_logs
13:59:30      if jepsen_archive := self.archive_log_remotely(jepsen_node, "./jepsen-scylla", "jepsen-data"):
13:59:30    File "/home/ubuntu/scylla-cluster-tests/sdcm/logcollector.py", line 686, in archive_log_remotely
13:59:30      if not node.remoter.run(f"tar czf '{archive_name}' -C '{archive_dir}' '{log_filename}'", ignore_status=True).ok:
13:59:30    File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 614, in run
13:59:30      result = _run()
13:59:30    File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 70, in inner
13:59:30      return func(*args, **kwargs)
13:59:30    File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_base.py", line 603, in _run
[...]

PR #8111, opened 2024-07-23
fix(logcollector): do not fail the whole log collection phase on a failure of one collector 

dimakr:
Log collection phase of a test run can fail if a failure occurs when collecting logs of a specific type.

Handling of unexpected exceptions is added, when iterating over log collectors, so that it continues with the next collector, instead of failing the whole 
log collection sequence.
Additionally the change adds skipping of jepsen logs collection sequence if no jepsen test was executed during SCT run.

Fixes https://github.com/scylladb/scylla-cluster-tests/issues/7470.

### Testing

- [x] :green_circle: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/sct-pr-provision-test/23/

### Reminders

### DIFF
--- a/sdcm/logcollector.py
+++ b/sdcm/logcollector.py
@@ -1584,11 +1584,16 @@ class Collector:  # pylint: disable=too-many-instance-attributes,
                                                   storage_dir=self.storage_dir,
                                                   params=self.params)
             LOGGER.info("Start collect logs for cluster %s", log_collector.cluster_log_type)
-            if result := log_collector.collect_logs(local_search_path=local_dir_with_logs):
-                results[log_collector.cluster_log_type] = result
-                LOGGER.info("collected data for %s\n%s\n", log_collector.cluster_log_type, result)
-            else:
-                LOGGER.warning("There are no logs collected for %s", log_collector.cluster_log_type)
+            try:
+                if result := log_collector.collect_logs(local_search_path=local_dir_with_logs):
+                    results[log_collector.cluster_log_type] = result
+                    LOGGER.info("collected data for %s\n%s\n", log_collector.cluster_log_type, result)
+                else:
+                    LOGGER.warning("There are no logs collected for %s", log_collector.cluster_log_type)
+            except Exception:  # pylint: disable=broad-except  # noqa: BLE001
+                LOGGER.warning(
+                    "%s is not able to collect logs. Moving to the next log collector",
+                    log_collector.__class__.__name__, exc_info=True)
         return results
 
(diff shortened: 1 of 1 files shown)


sdcm/logcollector.py
```diff
@@ -1202,6 +1203,9 @@ class JepsenLogCollector(LogCollector):
     cluster_dir_prefix = "jepsen-data"
 
     def collect_logs(self, local_search_path: Optional[str] = None) -> list[str]:
+        if 'jepsen' not in get_test_name():
```
fruch, line 1206: hmmm, I've just fixed some issue on a different place assuming think about job names.... 

it's hacky, and can come back biting

fruch: see
https://github.com/scylladb/scylla-cluster-tests/pull/8114/commits/2eaaaa3f8f2ab64bf1500c5253fd23bc5d227646

dimakr: I was struggling to find a way to clearly determine if the executed test was a jepsen test. For now found only this option - to check the job name

fruch: I would say the best way would be to have access to the test configuration.

And validating specific configuration is set.

But that should suffice for now

fruch: let's see if @soyacz and @vponomaryov might have other ideas regarding this

vponomaryov: The main fix makes the `jepsen` commit be not really needed.
I would prefer the latter one to be removed.

dimakr: I was thinking that maybe skip would save some time, for the case when we try to archive remote jepsen logs that don't exist. But `if jepsen_archive := self.archive_log_remotely(..)` call ends immediately in this case, as in:
```
17:28:05  Start collect logs for cluster jepsen-data
17:28:05  Unable to archive log `jepsen-scylla' to `./jepsen-data.tar.gz'
17:28:05  There are no logs collected for jepsen-data
```
So we gain nothing here and probably should indeed skip this? @fruch

fruch: Let's drop that part

The main issue is that the collection failed, 

There are two ways to avoid it, for now one is enough.

If it should still cause an issue, we'll get back to the whiteboard

dimakr: Removed the 2nd commit.

vponomaryov (approved):
LGTM

fruch (approved):

PR #8112, opened 2024-07-24
feature(elasticsearch): move to using api key

fruch:
switching form user/password to using api keys (tokens) also switching from using urls to using the elastic cloud_id which is what recommended by elasticsearch cloud, and enable compression.

Ref: https://www.elastic.co/guide/en/elasticsearch/client/python-api/current/connecting.html#connect-ec
Ref: https://github.com/scylladb/qa-tasks/issues/1747

### Testing

- [x] unittest (microbenchmarking.py tests was broken and fix)
- [x] with local command that uses elasticsearch 

### Reminders

### DIFF
--- a/sdcm/es.py
+++ b/sdcm/es.py
@@ -1,4 +1,5 @@
 import logging
+from functools import cached_property
 
 import elasticsearch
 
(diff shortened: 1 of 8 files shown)


soyacz (approved):
Can you create simple readme how to roll the token in case it will be needed in future?

Besides, LGTM

fruch:
> Can you create simple readme how to roll the token in case it will be needed in future?
> 
> Besides, LGTM

as all other cases, replacing it in the qa-keystore 
as for generating the api key, that's written in elastic cloud docs
for now only @roydahan and me can generate it

fruch:
V2:

- fix the microbenchmarking analyzer, it was using the configuration for the kibana url.

vponomaryov:
Change LGTM, but was it verified?

fruch:
> Change LGTM, but was it verified?

yes, with unittests, and with local command that uses elastic.

vponomaryov (approved):
LGTM

soyacz (approved):

Issue #7336, opened 2024-04-11
`hydra tunnel` to monitor node: `connect to: http://127.0.0.1:57581`: page returns "connection was reset"


kbr-scylla:
```
$ hydra tunnel --test-id 495a9f7c-4a9e-4fc7-be1f-77f28883a1da
...
? Select machine:  aws - longevity-5gb-1h-CorruptThenRepairM-monitor-node-495a9f7c-
1 - 34.248.3.99 10.4.1.128 - eu-west-1                                             
                                                                                   
tunnel into: longevity-5gb-1h-CorruptThenRepairM-monitor-node-495a9f7c-1           
ssh -i ~/.ssh/scylla-qa-ec2 -N -L 57581:10.4.1.128:3000 -o "UserKnownHostsFile=/dev/null" -o "StrictHostKeyChecking=no" -o ServerAliveInterval=10 jenkins@54.220.169.25
connect to: http://127.0.0.1:57581                                                 
Warning: Permanently added '54.220.169.25' (ECDSA) to the list of known hosts. 
```

But when I open http://127.0.0.1:57581 I see this:
![Screenshot from 2024-04-11 14-24-21](https://github.com/scylladb/scylla-cluster-tests/assets/120044486/39011347-8dc9-4095-a2d5-ae7c4411b21e)

And the terminal prints a lot of:
```
channel 2: open failed: connect failed: Connection refused                         
channel 2: open failed: connect failed: Connection refused                         
channel 2: open failed: connect failed: Connection refused                         
channel 2: open failed: connect failed: Connection refused                         
```

kbr-scylla:
Hm, perhaps the monitor node was not yet setup, that's why it's returning the error.

(but it could return a more useful error message)
(and it's weird that it did manage to identify some IP address for the monitor node etc.)

kbr-scylla:
Yup, after waiting for a little bit and seeing in the logs that the monitor node is up, I managed to connect

PR #8134, opened 2024-07-25
fix(hydra-tunnel): wait for connectivity

fruch:
when using this command before monitoring stack is ready it suggest the user too soon a url, that isn't ready.

this change introduced by default a check for connectivity and the user would see a progress indication until the port is open.

this change also fixes the usage of the correct ssh keys matching the current ones used (it was wrongly hard-code before)

Fixes: scylladb/scylla-cluster-tests#7336

### Testing

- [x] test manually 

### Reminders

### DIFF
--- a/sct_ssh.py
+++ b/sct_ssh.py
@@ -13,6 +13,7 @@
 import os
 import pty
 import shlex
+import socket
 import subprocess
 from functools import singledispatch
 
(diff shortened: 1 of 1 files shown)


soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

PR #8148, opened 2024-07-25
feature(argus): sent latency decorator results to argus

soyacz:
We want to store perf results in Argus. As first step sending latency decorator results: latencies and stalls events counts.

It works by sending results as soon as are created (upon finishing decorator work).
Updated Argus client to one supporting this feature.

refs: https://github.com/scylladb/qa-tasks/issues/1244

### Testing

- [x] - [grow-shrink perf like test scenario](https://argus.scylladb.com/workspace?state=WyI0NmZiODVlOS1hNTNjLTQ3OTAtOWFjZi1lYzU1YTc2N2ZhYmYiXQ)

### Reminders


sdcm/utils/decorators.py
```diff
@@ -155,6 +158,96 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
+class LatencyCalculatorMixedResult(GenericResultTable):
```
fruch, line 161: I think we should put those in it's own module

sdcm/utils/decorators.py
```diff
@@ -155,6 +158,96 @@ def wrapped(*args, **kwargs):
...
+
+
+workload_to_table = {
+    "mixed": LatencyCalculatorMixedResult,
+    "write": LatencyCalculatorWriteResult,
+    "read": LatencyCalculatorReadResult
+}
+
+
+def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, description: str, cycle: int, result: dict):
```
fruch, line 220: this is not a decorator, so me might put it in a diffrent module ?

unit_tests/test_argus_results.py
```diff
@@ -0,0 +1,8 @@
+from sdcm.sct_events.event_counter import STALL_INTERVALS
+from sdcm.utils.decorators import ReactorStallStatsResult
+
+
+def test_reactor_stalls_intervals():
+    """protecting against changes in the intervals so result table columns are always correct"""
```
fruch, line 6: why not code it, that it add the columns base on the same definition ?

sdcm/utils/decorators.py
```diff
@@ -155,6 +158,96 @@ def wrapped(*args, **kwargs):
...
+    for workload in ["write", "read"]:
+        summary = result["hdr_summary"]
+        if workload.upper() not in summary:
+            continue
+        for percentile in ["90", "99"]:
+            value = summary[workload.upper()][f"percentile_{percentile}"]
+            result_table.add_result(column=f"P{percentile} {workload}",
+                                    row=f"Cycle #{cycle}",
+                                    value=value,
+                                    status=Status.PASS if value < 10 else Status.ERROR)
```
fruch, line 233: the value 10, should come from the same definition we had so far

sdcm/utils/decorators.py
```diff
@@ -155,6 +158,96 @@ def wrapped(*args, **kwargs):
...
+    "write": LatencyCalculatorWriteResult,
+    "read": LatencyCalculatorReadResult
+}
+
+
+def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, description: str, cycle: int, result: dict):
+    result_table = workload_to_table[workload]()
+    result_table.name = f"{name} - {workload} - latencies"
+    result_table.description = description
+    for workload in ["write", "read"]:
```
fruch, line 224: what about mixed ?

sdcm/utils/decorators.py
soyacz: like for all results gathered in argus or just for performance?

sdcm/utils/decorators.py
```diff
@@ -155,6 +158,96 @@ def wrapped(*args, **kwargs):
     return wrapped
 
 
+class LatencyCalculatorMixedResult(GenericResultTable):
+    class Meta:
+        name = ""  # to be set by the decorator
```
fruch, line 163: if so, why we need to define those here ?

I thought that we would define the exact tables, and then use it where need.

now it's seem most of the things are dynamic but the columns (which can be dynamic as well)

sdcm/utils/decorators.py
soyacz: it's closely related to this one, I'm not sure. 
maybe along with tables above would make sense.

sdcm/utils/decorators.py
soyacz: latency results json contains write and read (both for mixed workload).
But this is a bug, as it overrides workload param passed to function.

sdcm/utils/decorators.py
fruch: I don't know yet, but those are not decorators, as the module says.

I think it makes sense to gather all of those in one place, so it would be one place to checkout what are we collecting.

even that no I see lots of the things names/descriptions are created dynamically, and it would be hard to see, with opening up the DB (or have argus to show all of them)

soyacz: Moved to different module `argus_results.py`

unit_tests/test_argus_results.py
soyacz: coded, test replaced with one that tests whole sending process.

sdcm/utils/decorators.py
soyacz: extracted definition and used it in existing place and here.

sdcm/utils/decorators.py
soyacz: fixed to not cause confusion

sdcm/utils/decorators.py
soyacz: yes, it's dynamic as latency_decorator is decorating different methods - and each method is a separate measurement (thus requires different table name).
Columns are static - (as per definition which can change in future)

soyacz:
@fruch I adapted the code to yours findings (thanks for that).
I reran test job to make sure all is still fine. I removed test jenkinsfile/config (which was fast and cheap).

soyacz:
looks it works, if we merge this one I'll proceed to next test results like latency/throughput.

soyacz:
> looks it works, if we merge this one I'll proceed to next test results like latency/throughput.

@fruch where we should backport it? is 6.1/2024.2/v15 enough?

fruch (approved):
LGTM

PR #8158, opened 2024-07-28
fix(nemesis-tests-trigger): disable trigger

fruch:
we disable the trigger long time ago,
but recently fix the job that update all sct jenkins jobs and that enabled it back, we don't need those runs for now.

### Testing

- [ ]

### Reminders


soyacz:
@fruch question, when we change such a setting, is there more actions required on e.g. jenkins side to make it effective?

fruch:
> @fruch question, when we change such a setting, is there more actions required on e.g. jenkins side to make it effective?

we need to run the create_jobs pipeline, to do the change (for now it's happens only once a week)

in this case I've also manually disabled it.

PR #8159, opened 2024-07-28
fix(fill_db_data): remove cache form properties

fruch:
since the status can change after the upgrade itself we should be caching those, and we should calculate them every single time.

also remove udt specific logic form the loop doing the skip condition, it's not needed anythmore, now that we have function calls for doing the logic for each case.

### Testing

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu20.04-test/19/

### Reminders

### DIFF
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -23,7 +23,6 @@ import time
 import re
 
 from collections import OrderedDict
-from functools import cached_property
 from uuid import UUID
 
 from cassandra import InvalidRequest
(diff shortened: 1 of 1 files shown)


temichus:
@fruch looks like issue still reproducing in https://argus.scylladb.com/test/cc80460f-2278-4fe5-8781-1776502bc7e1/runs?additionalRuns[]=c1d82b8a-d876-44a2-9bef-81809a8c7b56

PR #8172, opened 2024-07-29
feat(gpg) add the 2024 pub key (RSA)

yaronkaikov:
Adding a new GPG key in RSA format which supports both deb and rpm signing

Related to https://github.com/scylladb/scylla-enterprise/issues/4222


fruch:
let's do it with #8173, with clearing a bit some of the duplications

PR #8187, opened 2024-07-29
fix(hydra): don't running any python code outside hydra image

fruch:
when doing so we can easily break users depends on which python version they are using, we shouldn't assume anything about user env.

### Testing

- [ ]

### Reminders

### DIFF
--- a/docker/env/hydra.sh
+++ b/docker/env/hydra.sh
@@ -14,7 +14,6 @@ SCT_DIR=$(dirname "${DOCKER_ENV_DIR}")
 SCT_DIR=$(dirname "${SCT_DIR}")
 VERSION=v$(cat "${DOCKER_ENV_DIR}/version")
 HOST_NAME=SCT-CONTAINER
-RUN_BY_USER=$(python3 "${SCT_DIR}/sdcm/utils/get_username.py")
 USER_ID=$(id -u "${USER}"):$(id -g "${USER}")
 HOME_DIR=${HOME}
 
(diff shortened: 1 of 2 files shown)


docker/env/hydra.sh
```diff
@@ -245,8 +244,6 @@ function run_in_docker () {
     $([[ -n "$HYDRA_DRY_RUN" ]] && echo echo) \
     $tool ${REMOTE_DOCKER_HOST} run --rm ${TTY_STDIN} --privileged \
         -h ${HOST_NAME} \
-        -l "TestId=${SCT_TEST_ID}" \
```
vponomaryov, line 248: Why removing SCT_TEST_ID?

docker/env/hydra.sh
```diff
@@ -14,7 +14,6 @@ SCT_DIR=$(dirname "${DOCKER_ENV_DIR}")
 SCT_DIR=$(dirname "${SCT_DIR}")
 VERSION=v$(cat "${DOCKER_ENV_DIR}/version")
 HOST_NAME=SCT-CONTAINER
-RUN_BY_USER=$(python3 "${SCT_DIR}/sdcm/utils/get_username.py")
```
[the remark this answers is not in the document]
fruch: it's for labeling the hydra command itself, it's pointless

it from back when we have builder used from multiple tests at the same time, and we need to differentiated them from each other

now that we have  sct-runner it's doesn't help us much.

docker/env/hydra.sh
fruch: same as below, it's meant for when we were running lots of test at the same time on the builders

we aren't doing that anymore, hence those labels are not really needed, or helpful.

fruch: Unittest fails cause both of them, and would need to amended, either way

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

roydahan (approved):

Issue #8183, opened 2024-07-29
wait for jmx up does not work


soyacz:
https://github.com/scylladb/scylla-cluster-tests/pull/8046 introduced regression for waiting JMX port up: verification is fixed to `self.cql_address` which is not valid for JMX.

Test: `artifacts-ubuntu2204-nonroot-test`
Test id: `47773f63-db39-43f3-ae65-34d1b865aeed`
Test name: `enterprise-2024.2/artifacts-offline-install/artifacts-ubuntu2204-nonroot-test`
Test config file(s):

- [ubuntu2204.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/3cc2bafe6c9aadcc18cb8cd66e3ffab08e89f7ec/test-cases/artifacts/ubuntu2204.yaml)

<details>
<summary>
Logs and commands
</summary>

- Restore Monitor Stack command: `$ hydra investigate show-monitor 47773f63-db39-43f3-ae65-34d1b865aeed`
- Restore monitor on AWS instance using [Jenkins job](https://jenkins.scylladb.com/view/QA/job/QA-tools/job/hydra-show-monitor/parambuild/?test_id=47773f63-db39-43f3-ae65-34d1b865aeed)
- Show all stored logs command: `$ hydra investigate show-logs 47773f63-db39-43f3-ae65-34d1b865aeed`

## Logs:
- **db-cluster-47773f63.tar.gz** - [https://cloudius-jenkins-test.s3.amazonaws.com/47773f63-db39-43f3-ae65-34d1b865aeed/20240724_221833/db-cluster-47773f63.tar.gz](https://cloudius-jenkins-test.s3.amazonaws.com/47773f63-db39-43f3-ae65-34d1b865aeed/20240724_221833/db-cluster-47773f63.tar.gz)
- **sct-runner-events-47773f63.tar.gz** - [https://cloudius-jenkins-test.s3.amazonaws.com/47773f63-db39-43f3-ae65-34d1b865aeed/20240724_221833/sct-runner-events-47773f63.tar.gz](https://cloudius-jenkins-test.s3.amazonaws.com/47773f63-db39-43f3-ae65-34d1b865aeed/20240724_221833/sct-runner-events-47773f63.tar.gz)
[...]

PR #8190, opened 2024-07-29
fix(jmx): fix wait for jmx up

soyacz:
As we now verify ports being open from sct runner, we cannot use it for jmx as it listens on localhost.

Fix by verifying JMX being up by verification of log message in scylla-jmx service status.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/8183

### Testing

- [x] - [non root artifact test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-ubuntu2204-nonroot-test/9/)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -1231,7 +1231,9 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
     def jmx_up(self):
         if not self.is_service_exists(service_name='scylla-jmx'):
             return True
-        return self.is_port_used(port=7199, service_name="scylla-jmx")
+        return self.remoter.run(f"{self.systemctl} is-active scylla-jmx.service && "
+                                f"{self.systemctl} status scylla-jmx.service | grep 'JMX is enabled to receive remote connections on port'",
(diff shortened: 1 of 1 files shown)


yaronkaikov (approved):

soyacz:
> @soyacz
> 
> I wonder maybe we could use something like: https://github.com/ParallelSSH/parallel-ssh/blob/master/pssh%2Fclients%2Fnative%2Ftunnel.py
> 
> And have a way to tunnel any host port to a local port, and fix that across all the place we need direct communication with the node, and we'll be left with one assumption that we have ssh access to the node (or direct, or via bastion/jumpserver)

I wonder if it's stable - if network disruptions/vm restarts won't break it (or complicate the way we use it).
Instead of relying on constant connection, I'd rather run small/tiny agent on db nodes that would serve different purposes - available through http/rpc or even the other way round: agent would send events to SCT - but this is bigger arch change we could discuss in more detail if you see sense in it.

This PR is a quick fix for this rare/small check I'd recommend to merge it to unblock tests and think of better solution.

fruch (approved):
LGTM

PR #8196, opened 2024-07-30
test(tablets): disable tablets for unsupported features on 6.1

yarongilor:
Configure LWT/CDC pipelines with tablets-disabled,
	so it can be tested in 6.1, where tablets are enabled by default.

### Testing

- [ ]

### Reminders


yarongilor:
@roydahan , not sure why raised any need for a scylla-pkg PR as well. why not simply enabling existing jobs for 6.1 this way?

yarongilor:
Are there additional applicable pipeline for this configuration? counters test?

roydahan (changes requested):
I don't want to disable it in the original cases, but want to duplicate it to a "tablets disabled" or something similar folder.

Then, when you have the new tests, to have them triggered as part of master and 6.1.

yarongilor:
addressed in https://github.com/scylladb/scylla-cluster-tests/pull/8207

PR #8204, opened 2024-07-31
improvement(investigation): add test_method to config

soyacz:
When looking at the test run in Argus (and in general) it's hard to get the idea what test method was used - which is sometimes hard to get for jobs that run tests in parallel (e.g. perf tests with different test methods for each workload: write, read, mixed).

Added new config param `test_method` so it will be visible in sct.log and can be consumed and properly displayed in Argus.

refs: https://github.com/scylladb/argus/issues/375

### Testing

- [x] - provision test

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -1156,6 +1156,7 @@ def run_test(argv, backend, config, logdir):
     if logdir:
         os.environ['_SCT_LOGDIR'] = logdir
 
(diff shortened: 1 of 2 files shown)


fruch (approved):
LGTM

Weird that we didn't do that so far...

PR #8232, opened 2024-08-04
fix(convert_name_to_ami_if_needed): lookup on both accounts

fruch:
so far this function was look just into the main account, and not into the scylla image account, now that we want to use this feature, to lookup monitor images, it need to be checking both to find an AMI by its name.

### Testing

- [ ]

### Reminders


yaronkaikov (approved):

soyacz (approved):
LGTM

PR #8268, opened 2024-08-06
feature(pipelines): read extra environment variables from text parameter

fruch:
introducing a way to inject specific `SCT_*` environment variables, so we can configure a job with any sct configuration that is usable via environment variables, without creating pipeline parameter for each one.

we'll need this to pass in docker images to jobs, in the flow of drivers CI

### Testing

- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/58/

### Reminders


fruch:
TODO: add the parameter to all pipelines

roydahan (dismissed):
Cool idea.

vars/loadEnvFromString.groovy
```diff
@@ -0,0 +1,11 @@
+#!groovy
+
+def call(String parameter) {
+    if (parameter == null || parameter.isEmpty()) {
+        return
+    }
+    def props = readProperties text: parameter
+    props.each { key, value ->
+        env[key] = value
```
vponomaryov, line 9: Shouldn't we have some restricted env vars such as "cluster_backend", region name and some other for which we already have pipeline parameters?

fruch: that won't be needed, those would be set again by the pipeline.

i.e. we set those first, and any pipeline setting something else, would win.

so it might be confusing to who gonna be trying to set those from this.

those are for things that doesn't have pipeline parameters.

to sum it up, I don't think there's a need for safegurads, time would tell,
at most, `use at you own cation !!!` on the top of it :)

soyacz (approved):
LGTM

Why only longevity pipeline?

fruch:
> LGTM
> 
> Why only longevity pipeline?

tested the approach first, and a comment up there ⬆️  says need to copy it to the rest of pipeline.
I just need to get to it.

soyacz (approved):
LGTM

vponomaryov:
@fruch are you really sure that `vars/perfSearchBestConfigParallelPipeline.groovy` should really be removed?
@aleksbykov ^

aleksbykov:
> @fruch are you really sure that `vars/perfSearchBestConfigParallelPipeline.groovy` should be removed? @aleksbykov ^

@vponomaryov , @fruch , if we stop using this job: https://jenkins.scylladb.com/view/QA/job/QA-tools/job/perf-search-best-throughput-config-test/, then this file dont need

fruch:
> > @fruch are you really sure that `vars/perfSearchBestConfigParallelPipeline.groovy` should be removed? @aleksbykov ^
> 
> @vponomaryov , @fruch , if we stop using this job: https://jenkins.scylladb.com/view/QA/job/QA-tools/job/perf-search-best-throughput-config-test/, then this file dont need

kind of shame to have a full duplication of the perfParallel pipeline, just for as set of parameters
that now can be set via the newly introduced options, if anyone want to use it.

people are really using those jobs, and the fine tuning to select load for perf test is still done manually

I can leave this pipeline for now, but it would be missing any other feature we might introduce to pipeline (it's already is missing few).

mikliapko:
@fruch 
Could you please backport it into manager-3.3 as well?
Not sure I'll need it there somewhere but want to have it there for consistence

Issue #7427, opened 2024-05-13
Delay for writing log messages to log could cause reshard nemesis failing


aleksbykov:
### Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Node3 was decommissioned with `Decommission_seed_node` nemesis and new node was added node11. But during cleanup
on node11 issue: https://github.com/scylladb/scylladb/issues/14004 happened with network exception. In several nemesis, node11 was choosen for new nemesis `disrupt_restart_with_resharding`. First steps of nemesis are stop scylla update scylla.yaml and start scylla. And according to sct.log, command was executed succesfully on node11. 

Sct reported that command was executed on node11:
```
< t:2024-05-12 05:23:06,708 f:cluster.py      l:635  c:sdcm.cluster_aws     p:DEBUG > Node longevity-parallel-topology-schema--db-node-2a3bb187-11 [34.244.68.23 | 10.4.10.153]: CPUSET on node longevity-parallel-topology-schema--db-node-2a3bb187-11: CPUSET="--cpuset 1-7 "
< t:2024-05-12 05:23:06,974 f:nemesis.py      l:1001 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.SisyphusMonkey: Restart node with resharding. New murmur3_partitioner_ignore_msb_bits value: 15
< t:2024-05-12 05:23:06,974 f:decorators.py   l:118  c:sdcm.utils.decorators p:DEBUG > BEGIN: stop_scylla <AWSNode>
< t:2024-05-12 05:23:06,974 f:remote_base.py  l:521  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.4.10.153>: Running command "sudo systemctl stop scylla-server.service"...

< t:2024-05-12 05:23:55,490 f:base.py         l:143  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.4.10.153>: Command "sudo systemctl stop scylla-server.service" finished with status 0


< t:2024-05-12 05:24:00,261 f:remote_base.py  l:521  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.4.10.153>: Running command "sudo systemctl start scylla-server.service"...

< t:2024-05-12 05:24:44,815 f:base.py         l:143  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.4.10.153>: Command "sudo systemctl start scylla-server.service" finished with status 0

```
[...]

PR #8288, opened 2024-08-07
fix(restart_with_reshard): wait log messages for 10 minutes

aleksbykov:
Some times log messages are written to log file with delay, this could cause that scylla already started, but log_follower couldn't find appropriate log messages, because they will be written later. This happened with RestartWIthReshardNemesis.
Scylla was stopped, scylla yaml was updated with new `murmur3_partitioner_ignore_msb_bits` value. Then scylla started and resharding is expected. Resharding is checked by log messages, but if for some reason log messages will be written with delay to log (log_follower will fail the nemesis)

Fixes: scylladb/scylla-cluster-tests#7427

### Testing
- [Job with master passed](https://argus.scylladb.com/test/dc8131ca-2823-443a-9816-9eae668d6b6d/runs?additionalRuns[]=62cca91b-dc27-47aa-ae6b-ffdc8aa0cd2d)
- [Job with branch-2024.1](https://argus.scylladb.com/test/dc8131ca-2823-443a-9816-9eae668d6b6d/runs?additionalRuns[]=ae6cdd01-884d-4ff2-ac65-ffbfda7d9051)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -121,6 +121,7 @@ from sdcm.utils.version_utils import (
 )
 from sdcm.utils.net import get_my_ip
 from sdcm.utils.node import build_node_api_command
+from sdcm.wait import wait_for_log_lines
 from sdcm.sct_events import Severity
 from sdcm.sct_events.base import LogEvent, add_severity_limit_rules, max_severity
 from sdcm.sct_events.health import ClusterHealthValidatorEvent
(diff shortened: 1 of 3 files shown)


sdcm/cluster.py
```diff
@@ -2572,24 +2573,30 @@ def restart_node_with_resharding(self, murmur3_partitioner_ignore_msb_bits: int
...
+            if resharding_finished:
+                self.log.debug(f'Resharding has been finished successfully '
+                               f'(murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits})')
+            else:
+                raise Exception(f'Resharding has not been finished within {start_scylla_timeout}'
+                                f'(murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits}) '
+                                'Check the log for the details')
+            return True
+
+        wait_for(wait_resharding_log, step=5, timeout=900, text="Wait resharding log messages...")
```
fruch, line 2599: maybe the `wait_for_log_lines` context manager is a better fit for this case ? 

what do you think @soyacz

soyacz: yes, I think this is exactly the case for `wait_for_log_lines`
@aleksbykov you can enhance this method with another info about context that could be appended to error log messages in it and also add debug messages in some places.

sdcm/cluster.py
```diff
@@ -2567,29 +2568,20 @@ def restart_node_with_resharding(self, murmur3_partitioner_ignore_msb_bits: int
...
-
-        if resharding_finished:
-            self.log.debug(f'Resharding has been finished successfully '
-                           f'(murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits})')
-        else:
-            raise Exception(f'Resharding has not been finished within {start_scylla_timeout}'
-                            f'(murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits}) '
-                            'Check the log for the details')
+        resharding_not_started_msg = f'Resharding has not been started \
+        (murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits}) Check the log for the details'
```
fruch, line 2573: not sure how those lines are better than a massage having the regular expression that didn't matched the logs

you can catch the exception, and append more context info if needed, changing the exception message in this case, make it a bit harder to debug.

soyacz: yes, I was thinking more about appending the context.
See this msg misses info about exact pattern (and sometimes just Scylla changes log structure) and missing node affected.

sdcm/cluster.py
```diff
@@ -2568,28 +2569,20 @@ def restart_node_with_resharding(self, murmur3_partitioner_ignore_msb_bits: int
         with self.remote_scylla_yaml() as scylla_yml:
             scylla_yml.murmur3_partitioner_ignore_msb_bits = murmur3_partitioner_ignore_msb_bits
 
-        search_reshard_start = self.follow_system_log(patterns=[DB_LOG_PATTERN_RESHARDING_START])
-        search_reshard_finish = self.follow_system_log(patterns=[DB_LOG_PATTERN_RESHARDING_FINISH])
-        start_scylla_timeout = 7200
-        self.start_scylla(timeout=start_scylla_timeout)
-        resharding_started = list(search_reshard_start)
-        resharding_finished = list(search_reshard_finish)
+        resharding_start_ctx = f'Resharding started on node {self.name} with \
```
soyacz, line 2572: no need to mention 'resharding started...' just giving `resharding_details=f"Resharding: (murmur3_partitioner_ignore_msb_bits={murmur3_partitioner_ignore_msb_bits})"`
which could be the same for both start and stop.
And later in error msg just append (to the end with some prefix like `<original exc mst>. Context: {err_msg_context}`

soyacz (approved):
LGTM

fruch (approved):

PR #8296, opened 2024-08-08
fix(provision-tests): add `requested_by_user` pipeline parameter

fruch:
since `requested_by_user` was missing, we ended up with provision tests resources tagged by `null` user

azure test shows it like:
```
Tag 'RunByUser' has value 'null' which is not allowed by Azure API. Replacing with empty string.
```

### Testing

- [x] provision test

### Reminders

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -73,6 +73,9 @@ pipeline {
         string(defaultValue: "5.2.11",
                description: 'the scylla version to use for the provision tests',
                name: 'scylla_version')
+        string(defaultValue: '',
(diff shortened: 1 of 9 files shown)


soyacz (commented):
another approach that would be easier/(safer?) would be replacing `null` in `actual_user_from_request` from `get_username` function with empty string.

vponomaryov (approved):
LGTM

fruch:
> another approach that would be easier/(safer?) would be replacing `null` in `actual_user_from_request` from `get_username` function with empty string.

I would rather stop the issue at the start of the flow, then at the end, in multiple unrelated places

PR #8297, opened 2024-08-08
fix(keystore): stop reusing boto3 clients in multiple threads

fruch:
It seems like it's not safe to share clients between threads and we started seeing cases of SCT segfaulting
right at the start, during the `Keystore.sync()`
which run multiple threads to make the process faster but it doesn't help if it also make things unstable

looking at botocore issues, seem like it's documented that botocore isn't threadsafe

Ref: https://github.com/boto/botocore/issues/3164

### Testing

- [ ] provision test

### Reminders

### DIFF
--- a/sdcm/keystore.py
+++ b/sdcm/keystore.py
@@ -29,9 +29,13 @@ SSHKey = namedtuple("SSHKey", ["name", "public_key", "private_key"])
 
 
 class KeyStore:  # pylint: disable=too-many-public-methods
-    def __init__(self):
-        self.s3: S3ServiceResource = boto3.resource("s3")
-        self.s3_client: S3Client = boto3.client("s3")
+    @property
+    def s3(self) -> S3ServiceResource:
+        return boto3.resource("s3")
+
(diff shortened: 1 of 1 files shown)


sdcm/keystore.py
```diff
@@ -29,9 +29,13 @@
...
-    def __init__(self):
-        self.s3: S3ServiceResource = boto3.resource("s3")
-        self.s3_client: S3Client = boto3.client("s3")
+    @property
+    def s3(self) -> S3ServiceResource:
+        return boto3.resource("s3")
+
+    @property
+    def s3_client(self) -> S3Client:
+        return boto3.client("s3")
```
soyacz, line 38: how this change helps?

vponomaryov: It returns new object each time the `s3_client` attribute is called.
It allows to avoid object sharing in different threads.

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

PR #8300, opened 2024-08-09
test(manager): implement test for no-delta backup

mikliapko:
Closes https://github.com/scylladb/scylla-manager/issues/3936

Changes:
- The test that runs two backups one by one, no inserting any data to DB between them, and checks the 2nd backup takes time -> 0 has been added;
- Corresponding config file for 100GB dataset, test-case yaml and .jenkinsfile have been added as well.

### Testing

- [x] https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/no-delta-backup-test/8/

### Reminders

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -21,7 +21,7 @@ from functools import cached_property
 import re
 import time
 from textwrap import dedent
-from datetime import datetime
+from datetime import datetime, timedelta
 from dataclasses import dataclass
 
 import boto3
@@ -42,6 +42,7 @@ from sdcm.utils.common import reach_enospc_on_node, clean_enospc_on_node
 from sdcm.utils.loader_utils import LoaderUtilsMixin
 from sdcm.sct_events.system import InfoEvent
 from sdcm.sct_events.group_common_events import ignore_no_space_errors, ignore_stream_mutation_fragments_errors
+from sdcm.utils.compaction_ops import CompactionOps
 from sdcm.utils.gce_utils import get_gce_storage_client
 from sdcm.utils.azure_utils import AzureService
 from sdcm.utils.tablets.common import TabletsConfiguration
@@ -601,6 +602,53 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         self.run_verification_read_stress(ks_names)
         mgr_cluster.delete()  # remove cluster at the end of the test
         self.log.info('finishing test_basic_backup')
+
+    def test_no_delta_backup_at_disabled_compaction(self):
+        """The purpose of test is to check that delta backup (no changes to DB between backups) takes time -> 0.
+
+        Important test precondition is to disable compaction on all nodes in the cluster.
+        Otherwise, new set of SSTables is created what ends up in the situation that almost no deduplication is applied.
+
+        For more details https://github.com/scylladb/scylla-manager/issues/3936#issuecomment-2277611709
+        """
+        self.log.info('starting test_consecutive_backups')
+
+        self.log.info('Run write stress')
+        self.run_prepare_write_cmd()
+
+        self.log.info('Disable compaction for every node in the cluster')
+        compaction_ops = CompactionOps(cluster=self.db_cluster)
+        for node in self.db_cluster.nodes:
+            compaction_ops.disable_autocompaction_on_ks_cf(node=node)
+
+        self.log.info('Prepare Manager')
+        manager_tool = mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0])
+        mgr_cluster = self._ensure_and_get_cluster(manager_tool, force_add=True)
+
+        self.log.info('Run backup #1')
+        backup_task_1 = mgr_cluster.create_backup_task(location_list=self.locations)
+        backup_task_1_status = backup_task_1.wait_and_get_final_status(timeout=3600)
+        assert backup_task_1_status == TaskStatus.DONE, \
+            f"Backup task ended in {backup_task_1_status} instead of {TaskStatus.DONE}"
+        self.log.info(f'Backup task #1 duration - {backup_task_1.duration}')
+
+        self.log.info('Run backup #2')
+        backup_task_2 = mgr_cluster.create_backup_task(location_list=self.locations)
+        backup_task_2_status = backup_task_2.wait_and_get_final_status(timeout=60)
+        assert backup_task_2_status == TaskStatus.DONE, \
+            f"Backup task ended in {backup_task_2_status} instead of {TaskStatus.DONE}"
+        self.log.info(f'Backup task #2 duration - {backup_task_2.duration}')
+
+        assert backup_task_2.duration < timedelta(seconds=15), "No-delta backup took more than 15 seconds"
+
+        self.log.info('Verify restore from backup #2')
+        self.verify_backup_success(mgr_cluster=mgr_cluster, backup_task=backup_task_2,
+                                   restore_data_with_task=True, timeout=3600)
+
+        self.log.info('Run verification read stress')
+        self.run_verification_read_stress()
+
+        self.log.info('finishing test_consecutive_backups')
 
     def test_restore_backup_with_task(self, ks_names: list = None):
         self.log.info('starting test_restore_backup_with_task')
--- a/jenkins-pipelines/manager/ubuntu22-manager-backup.jenkinsfile
+++ b/jenkins-pipelines/manager/ubuntu22-manager-backup.jenkinsfile
@@ -1,0 +1,15 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+managerPipeline(
+    backend: 'aws',
+    region: 'us-east-1',
+    test_name: 'mgmt_cli_test.MgmtCliTest.test_no_delta_backup_at_disabled_compaction',
+    test_config: """['test-cases/manager/manager-no-delta-backup-set-dataset.yaml']""",
+
+    post_behavior_db_nodes: 'destroy',
+    post_behavior_loader_nodes: 'destroy',
+    post_behavior_monitor_nodes: 'destroy',
+)
--- a/test-cases/manager/manager-no-delta-backup-set-dataset.yaml
+++ b/test-cases/manager/manager-no-delta-backup-set-dataset.yaml
@@ -1,0 +1,14 @@
+test_duration: 200
+
+prepare_write_cmd: "cassandra-stress write cl=ONE n=10485760 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..10485760"
+stress_read_cmd: "cassandra-stress read cl=ONE n=10485760 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..10485760"
+
+instance_type_db: 'i4i.xlarge'
+instance_type_loader: 'c6i.xlarge'
+
+region_name: 'us-east-1'
+n_db_nodes: 3
+n_monitor_nodes: 1
+n_loaders: 1
+
+user_prefix: manager-backup


mgmt_cli_test.py
```diff
@@ -602,6 +603,53 @@ def test_basic_backup(self, ks_names: list = None):
...
+        """
+        self.log.info('starting test_consecutive_backups')
+
+        self.log.info('Run write stress')
+        self.run_prepare_write_cmd()
+
+        self.log.info('Disable compaction for every node in the cluster')
+        compaction_ops = CompactionOps(cluster=self.db_cluster)
+        for node in self.db_cluster.nodes:
+            compaction_ops.disable_autocompaction_on_ks_cf(node=node)
```
soyacz, line 622: I think you might also `flush` memtables to disk before doing backup.

soyacz, line 622: I wonder if disable_autocompaction disables also tablets load balancer and tablet split/merge mechanism.

mgmt_cli_test.py
```diff
@@ -602,6 +603,53 @@ def test_basic_backup(self, ks_names: list = None):
...
+        backup_task_2 = mgr_cluster.create_backup_task(location_list=self.locations)
+        backup_task_2_status = backup_task_2.wait_and_get_final_status(timeout=60)
+        assert backup_task_2_status == TaskStatus.DONE, \
+            f"Backup task ended in {backup_task_2_status} instead of {TaskStatus.DONE}"
+        self.log.info(f'Backup task #2 duration - {backup_task_2.duration}')
+
+        assert backup_task_2.duration < timedelta(seconds=15), "No-delta backup took more than 15 seconds"
+
+        self.log.info('Verify restore from backup #2')
+        self.verify_backup_success(mgr_cluster=mgr_cluster, backup_task=backup_task_2,
```
soyacz, line 645: maybe worth to delete backup 1 to see if it doesn't break backup 2?

configurations/manager/10GB_dataset.yaml
```diff
@@ -0,0 +1,8 @@
+round_robin: true
+
+prepare_write_cmd: ["cassandra-stress write cl=ONE n=10485760 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..10485760"]
+stress_read_cmd: ["cassandra-stress read cl=ONE n=10485760 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(64) n=FIXED(16)' -pop seq=1..10485760"]
+
+instance_type_db: 'i4i.2xlarge'
```
soyacz, line 6: this overrides db type from `test-cases/manager/manager-no-delta-backup-set-dataset.yaml` which one is valid?

mgmt_cli_test.py
Michal-Leszczynski, line 622: I'm not sure if disabling autocompaction stops ongoing compactions, or does it just prevent new ones from starting. If it does not stop them, then it's possible that there would be a compaction running during the backup. It would result in creating new files that needs to be uploaded as a part of the second backup.

You could create table, disable autocompaction and only after that run the c-s command to fill the table.

Michal-Leszczynski: I believe that taking a snapshot flushes table on its own.

Michal-Leszczynski: That's an interesting question. @denesb do you know the answer?

mgmt_cli_test.py
mikliapko: IMO It might be a good improvement for the test.
@Michal-Leszczynski What do you think about it?

mgmt_cli_test.py
mikliapko: I suppose we are pretty save in this regard. 
Here is the output of `sctool progress ...` for the two consecutive backup operations:
```
Running command "sudo sctool  -c 75f34691-49ca-4fdb-bfa3-d56229b4e809 progress backup/2a9c1120-6074-474b-80e5-001393f359f7" finished with status 0
Run:		96414835-5676-11ef-ac67-02a275dbf199
Status:		DONE
Start time:	09 Aug 24 17:41:19 UTC
End time:	09 Aug 24 17:43:39 UTC
Duration:	2m19s
Progress:	100%
Snapshot Tag:	sm_20240809174120UTC
Datacenters:	
- us-east

╭─────────────┬──────────┬─────────┬─────────┬──────────────┬────────╮
│ Host        │ Progress │    Size │ Success │ Deduplicated │ Failed │
├─────────────┼──────────┼─────────┼─────────┼──────────────┼────────┤
│ 10.12.0.64  │     100% │ 11.406G │ 11.406G │            0 │      0 │
│ 10.12.0.69  │     100% │ 11.406G │ 11.406G │            0 │      0 │
│ 10.12.2.186 │     100% │ 11.406G │ 11.406G │            0 │      0 │
╰─────────────┴──────────┴─────────┴─────────┴──────────────┴────────╯
```
No-delta backup operation (the test was executed from the PR branch):
```
Running command "sudo sctool  -c 75f34691-49ca-4fdb-bfa3-d56229b4e809 progress backup/783ae129-6762-46cb-98bb-4321aaf1d1e6"...

Run:		1045364a-5677-11ef-ac68-02a275dbf199
Status:		DONE
Start time:	09 Aug 24 17:44:44 UTC
End time:	09 Aug 24 17:44:47 UTC
Duration:	2s
Progress:	100%
Snapshot Tag:	sm_20240809174444UTC
Datacenters:	
- us-east

╭─────────────┬──────────┬─────────┬─────────┬──────────────┬────────╮
│ Host        │ Progress │    Size │ Success │ Deduplicated │ Failed │
├─────────────┼──────────┼─────────┼─────────┼──────────────┼────────┤
│ 10.12.0.64  │     100% │ 11.406G │ 11.406G │      11.406G │      0 │
│ 10.12.0.69  │     100% │ 11.406G │ 11.406G │      11.406G │      0 │
│ 10.12.2.186 │     100% │ 11.406G │ 11.406G │      11.406G │      0 │
╰─────────────┴──────────┴─────────┴─────────┴──────────────┴────────╯
```
As you can see, all the data is deduplicated in case of second backup and no size changes happened.

configurations/manager/10GB_dataset.yaml
mikliapko: Fixed, removed `instance_type_db` from test-cases yaml

mgmt_cli_test.py
soyacz: If this is a problem, we have something implemented in perf tests `_wait_no_compactions_running` - could be moved and reused if necessary.

configurations/manager/10GB_dataset.yaml
mikliapko: Ah, okay, test-cases yaml should contain instance_type_db parameter, swapped

mgmt_cli_test.py
denesb: No, disable_autocompaction only disables compactions (and waits for any running one to finish).

denesb: Note that this only refers to automatic compaction -- user can still start compaction via API (e.g. major, scrub, etc).

mgmt_cli_test.py
Michal-Leszczynski: I don't fully understand what "delete backup 1" means.
Is it about deleting the task? Or is it about deleting the snapshot?
We could do the first one, although this wouldn't be relevant to the test.
We can't do the second one, as then the backup would need to be fully re-uploaded.

mikliapko: If we can't do the second one, then it actually doesn't make sense.

soyacz: ok, so it means delta-backup strongly relies on previous snapshots and these can't be deleted. I hope it's documented somewhere.

mikliapko: Hm, just curious, in `backup delete` [doc](https://manager.docs.scylladb.com/stable/sctool/backup.html#backup-delete) there is such info:
```
Deduplicated files are persisted unless their reference count drops to zero.
```
I suppose the flow with running backup deletion with `sctool backup delete --snapshot-tag BACKUP_1_TAG` should keep deduplicated files required for backup 2 and we are kinda safe to restore from it.
@Michal-Leszczynski What do you think?

mikliapko: Whatever, guys @soyacz @Michal-Leszczynski, if you don't mind I'd suggest to move this investigation and possible test improvement to a separate ticket to not block this PR now since we don't have much time at the moment to work on it?

Michal-Leszczynski (approved):

configurations/manager/10GB_dataset.yaml
fruch: just to clarify, anything under `test-cases` folder is checked, and expected to be full functioning cases as it, hence all the mandatory field are checked

anything under `configurations` is known to be partial configuration, and not being tested by CI for correctness and missing options.

mgmt_cli_test.py
fruch: in dtest when disabling compactions, we wait for compactions to finish before moving on.

even that it's not completely clear if backup can work as expected during compactions, why manager is stop the compactions on it's own on the tables is currently working on ?, why does the user need to care about it ?

configurations/manager/10GB_dataset.yaml
```diff
@@ -0,0 +1,7 @@
+round_robin: true
```
soyacz, line 1: If there are no plans to reuse this 10GB dataset in any other test with these settings, better move it to `test-cases/manager/manager-no-delta-backup-set-dataset.yaml`

soyacz (approved):
LGTM

configurations/manager/10GB_dataset.yaml
soyacz: I mean whole file, not only `round_robin` setting.

mikliapko: I'd prefer to keep it here in separate file.
It allows us to be more flexible and reuse this dataset in another tests or run this test with different dataset.
It's sometimes a case for the runs which we trigger manually from time to time.

mgmt_cli_test.py
mikliapko: > in dtest when disabling compactions, we wait for compactions to finish before moving on.

I'll add this check in the test to wait compaction to finish.

> even that it's not completely clear if backup can work as expected during compactions, why manager is stop the compactions on it's own on the tables is currently working on ?, why does the user need to care about it ?

The test checks the specific case of data deduplication for consecutive no-delta backup. This scenario might be achieved only with compaction disabled. See more details in the [comment](https://github.com/scylladb/scylla-manager/issues/3936#issuecomment-2277611709).

cc @karol-kokoszka

configurations/manager/10GB_dataset.yaml
mikliapko: okay, I got the point, the test should be fully functioning with only test-cases configuration provided, will move

mikliapko: Done

mikliapko:
Comments have been addresses, retested here:
https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/no-delta-backup-test/8/

karol-kokoszka (approved):

roydahan:
> @soyacz @fruch I suppose every comment is addressed. Could you please take a look once again?

Both are on PTO, coming back next week if it can wait.

mikliapko:
> Both are on PTO, coming back next week if it can wait.

yes, it can wait, no problem

soyacz (approved):
LGTM

soyacz:
need one more review from e.g. @grzywin to be able to merge this one.

mgmt_cli_test.py
```diff
@@ -602,6 +603,53 @@ def test_basic_backup(self, ks_names: list = None):
...
+        mgr_cluster = self._ensure_and_get_cluster(manager_tool, force_add=True)
+
+        self.log.info('Run backup #1')
+        backup_task_1 = mgr_cluster.create_backup_task(location_list=self.locations)
+        backup_task_1_status = backup_task_1.wait_and_get_final_status(timeout=3600)
+        assert backup_task_1_status == TaskStatus.DONE, \
+            f"Backup task ended in {backup_task_1_status} instead of {TaskStatus.DONE}"
+        self.log.info(f'Backup task #1 duration - {backup_task_1.duration}')
+
+        self.log.info('Run backup #2')
```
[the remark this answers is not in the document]
grzywin: @mikliapko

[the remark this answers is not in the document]
mikliapko: Yes, I agree that it makes code less redundant, but I don't think it really adds to readability.
Moreover, from my point of view, it adds to complexity making it harder to understand the test whose main idea is to run two backups one by one. I want to keep them as two separate steps to emphasize that.
One more thing, in the further [steps](https://github.com/mikliapko/scylla-cluster-tests/blob/incremental-backup/mgmt_cli_test.py#L642:L646) the variable `backup_task_2` is used several times. Again, for easier test understanding it's better to have it as `backup_task_2` not simply `backup_task`.

@grzywin What do you think?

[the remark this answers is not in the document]
grzywin: @mikliapko Personally, I prefer to loop such redundant code fragments. It is easier for me to read and understand, but it is not a 'must'. I missed the fact that you are using `backup_test_2` several times later on, in such case I agree to leave it as it is.

grzywin (approved):
lgtm

mikliapko:
> need one more review from e.g. @grzywin to be able to merge this one.

@soyacz I think we are good to merge now, thanks!

soyacz:
hmm, still can't - @fruch can you help out merging this one? Need 2 approves from reviewers with write access

fruch (approved):

PR #8314, opened 2024-08-09
feature(reuse-cluster): add a test for reuse-cluster option of SCT

dimakr:
The change makes the reuse-cluster feature testable/guarded, so we can ensure that it works when it is offered to a wider audience for use in their dev work.

Closes: https://github.com/scylladb/qa-tasks/issues/1749

### Testing
- [ ]

### Reminders

### DIFF
--- a/reuse_cluster_test.py
+++ b/reuse_cluster_test.py
@@ -1,0 +1,75 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2024 ScyllaDB
+import logging
+import unittest
+
+from sdcm.cluster import TestConfig
+from sdcm.sct_events.events_processes import create_default_events_process_registry
+from sdcm.tester import ClusterTester
+from sdcm.utils.loader_utils import LoaderUtilsMixin
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+class ReuseClusterTest(unittest.TestCase):
+    """Test ability to reuse existing cluster in tests"""
+
+    # Define reusable equivalent to ClusterTester that can be instantiated multiple times during the test
+    ReusableCluster = type('ReusableCluster', (ClusterTester, LoaderUtilsMixin), {})
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.cluster = None
+        self.ReusableCluster.events_processes_registry = (
+            create_default_events_process_registry(log_dir=TestConfig().logdir()))
+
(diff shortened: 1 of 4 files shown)


soyacz (commented):
wouldn't it be more robust if added it to aws-provision test in jenkins pipeline as another step?

dimakr:
> wouldn't it be more robust if added it to aws-provision test in jenkins pipeline as another step?

Not sure if it is exactly how the reuse_cluster is expected to be used. I.e. we need to ensure that the whole cluster_init sequence is repeated for a cluster and it is operational after that.
If add it as an additional step to existing test, then it would be reusing the same cluster_tester object and part of init/teardown sequence would be skipped.

But I experience another issue now - from local machine (the target case for the reuse option) the test is running successfully all the time. But in Jenkins I'm getting the error:
```
22:56:48  ======================================================================
22:56:48  ERROR: test_reuse_cluster (reuse_cluster_test.ReuseClusterTest)
22:56:48  ----------------------------------------------------------------------
22:56:48  Traceback (most recent call last):
22:56:48    File "/home/ubuntu/scylla-cluster-tests/reuse_cluster_test.py", line 63, in test_reuse_cluster
22:56:48      self.provision_cluster(reuse_cluster_id=test_id)
22:56:48    File "/home/ubuntu/scylla-cluster-tests/reuse_cluster_test.py", line 42, in provision_cluster
22:56:48      self.cluster = self.ReusableCluster()
22:56:48    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 383, in __init__
22:56:48      self.test_config.configure_syslogng(self.localhost)
22:56:48    File "/home/ubuntu/scylla-cluster-tests/sdcm/test_config.py", line 220, in configure_syslogng
22:56:48      cls._link_running_syslog_logdir(node.syslogng_log_dir)
22:56:48    File "/home/ubuntu/scylla-cluster-tests/sdcm/test_config.py", line 214, in _link_running_syslog_logdir
22:56:48      current_logdir.rmdir()
22:56:48    File "/usr/local/lib/python3.10/pathlib.py", line 1215, in rmdir
22:56:48      self._accessor.rmdir(self)
22:56:48  NotADirectoryError: [Errno 20] Not a directory: '/home/ubuntu/sct-results/20240811-204702-400924/hosts'
```
@soyacz @fruch maybe you know what I'm missing, saw similar behaviour earlier?

dimakr:
Seems like the issue is that when executing in Jenkins the provision-resources stage creates its own logdir (opposite to local run, when this stage is not executed) in one container.
Then it creates another logdir when the Run SCT stage is executed, in another container.

In case of local run we start immediately from 'run SCT' step and don't see the issue..

vponomaryov:
> > wouldn't it be more robust if added it to aws-provision test in jenkins pipeline as another step?
> 
> Not sure if it is exactly how the reuse_cluster is expected to be used. I.e. we need to ensure that the whole cluster_init sequence is repeated for a cluster and it is operational after that. If add it as an additional step to existing test, then it would be reusing the same cluster_tester object and part of init/teardown sequence would be skipped.

Calling hydra command for running tests twice (where before 2nd we set SCT_REUSE_CLUSTER env var) from a pipeline must be completely sufficient.

dimakr:
After discussion it was decided that it's better to re-use the existing provision test (and just add a new / extend existing pipeline with a reuse_cluster option), than introducing a new test.
Closing this PR as not needed.

PR #8351, opened 2024-08-13
improvement(argus): migrate latency results to Argus script

soyacz:
Added simple script for migrating latency results data to Argus. Still semi automatic (need to specify job paths manually) but does the job.

### Testing

- [x] - migrated few jobs results

### Reminders

### DIFF
--- a/utils/migrate_latency_results_to_argus.py
+++ b/utils/migrate_latency_results_to_argus.py
@@ -1,0 +1,128 @@
+#!/usr/bin/env python
+import sys
+import os.path
+import datetime
+import logging
+import logging.config
+from collections import OrderedDict
+
+import argus
+from argus.client.sct.client import ArgusSCTClient
+from elasticsearch import Elasticsearch
+
+from sdcm.argus_results import send_result_to_argus
+from sdcm.keystore import KeyStore
+
+sys.path.append(os.path.join(os.path.dirname(__file__), ".."))
+
+
+LOGGER = logging.getLogger(__name__)
+
+
+def migrate(creds, job_name, days=7, index_name="performancestatsv2", dry_run=True):  # pylint: disable=too-many-locals
+    ks = KeyStore()
+    es_conf = ks.get_elasticsearch_token()
+    elastic_search = Elasticsearch(**es_conf)
+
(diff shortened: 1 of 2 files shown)


utils/migrate_latency_results_to_argus.py
```diff
@@ -0,0 +1,114 @@
...
+                except argus.client.base.ArgusClientError:
+                    print(
+                        f"Failed to send {operation} - {workload} - latencies to Argus: {hit['_source']['test_details']['job_url']}")
+                continue
+            for idx, cycle in enumerate(latency_during_ops[operation]["cycles"], start=1):
+                if dry_run:
+                    print(f"Would send {operation} - {workload} - latencies - cycle {idx} to Argus")
+                    continue
+                try:
+                    send_result_to_argus(
```
fruch, line 82: do we handle duplicates report ? i.e. we'll want to consider running with production argus data ?

not critical to handle it, but just to know if it would be an option or not

fruch (approved):
LGTM

utils/migrate_latency_results_to_argus.py
soyacz: In case of duplication it will override the data.
Still one more thing to fix, as I want to upload it in proper order (to fix issue with wrong order of tables in Argus).

soyacz:
I updated code with proper ordering. But still there's one more thing before merging it:
I think I made mistake when designing table names - workload should be at the beginning of a result (table) name instead after decorated function name - it's more natural and will be better for filtering graphs feature (I also think of filtering results like in graphs).
I'll add separate commit for that.

soyacz:
@fruch We didn't make it before our holidays, let's merge it then when I'm back.

fruch:
> @fruch We didn't make it before our holidays, let's merge it then when I'm back.

If it's working as you expect we can merge it, but no one is going to run it while you are away :)

soyacz:
@fruch can you merge it?

soyacz:
tables names fixed in Argus prod. Master/Enterprise perf tests data migrated (during ops/elasticity/upgrade tests).

PR #8369, opened 2024-08-14
improvement(monitoring): remove dashboard json after upload

cezarmoise:
Dashboard json remains in the git repo after a test run. This removes it after it is no longer needed.

### Testing

- tested locally

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -181,11 +181,6 @@ MAX_TIME_WAIT_FOR_DECOMMISSION: int = HOUR_IN_SEC * 6
 LOGGER = logging.getLogger(__name__)
 
 
-def remove_if_exists(file_path):
-    if os.path.exists(file_path):
-        os.remove(file_path)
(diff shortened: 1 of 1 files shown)


sdcm/cluster.py
```diff
@@ -5809,6 +5809,7 @@ def start_scylla_monitoring(self, node):
         self.add_sct_dashboards_to_grafana(node)
         self.save_sct_dashboards_config(node)
         self.save_monitoring_version(node)
+        remove_if_exists(self.sct_dashboard_json_file)
```
fruch, line 5812: I would recommend using: 
https://docs.python.org/3/library/pathlib.html#pathlib.Path.unlink

and remove completely `remove_if_exists`

fruch: in general we should shy away from `os.path`, at some day it would be deprecated

fruch (approved):
LGTM

Issue #8198, opened 2024-07-30
CreateIndex nemesis was started (and failed) on a node that was previously terminated by NodeTerminateAndReplace parallel nemesis 


dimakr:
## Packages

Scylla version: `2024.1.8-20240724.fc3e399a25f3` with build-id `646cf933d8926947ade5b2a7cbc5bacb145df4fb`
Kernel Version: `5.15.0-1066-aws`

## Issue description

During [enterprise-2024.1/longevity/longevity-multidc-schema-topology-changes-12h-test#26](https://argus.scylladb.com/test/8764d79e-d785-4b99-9c70-99e2f24d0a18/runs?additionalRuns[]=97c11d18-65ec-4dfa-9b9d-70ba669c3f11) test `disrupt_terminate_and_replace_node` and `disrupt_create_index` nemeses were started in parallel and were targeted onto the same node-7.
NodeTerminateAndReplace nemesis started node termination at 02:43:52 and finished at 02:49:38:
```
2024-07-25 02:43:52,385 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-07-25 02:43:52.383: (InfoEvent Severity.NORMAL) period_type=not-set event_id=b225fcb9-0252-4021-b079-81bdd9c5508c: message=StartEvent - Terminate node and wait 5 minutes
...
2024-07-25 02:49:38,553 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-07-25 02:49:38.552: (InfoEvent Severity.NORMAL) period_type=not-set event_id=7f4d8643-394f-4b82-b161-8b1661ea938f: message=FinishEvent - target_node was terminated
```
CreateIndex tried to start index creation on node-7 at 02:54:47:
```
2024-07-25 02:54:47,644 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-07-25 02:54:47.641: (InfoEvent Severity.NORMAL) period_type=not-set event_id=3f17c704-6a15-44d2-8442-7f30371e414f: message=Starting creating index: keyspace1.standard2(c1)
```
and eventually failed as the node was no longer available, with the error:
```
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5094, in wrapper
result = method(*args[1:], **kwargs)
[...]

PR #8381, opened 2024-08-15
fix(nemesis): make sure `set_target_node` always set `current_disruption`

fruch:
before there was two seprate calls to `set_target_node` and to `set_current_disruption` that could end up with `set_target_node` setting None to the `target_node.running_nemesis`

this fix move `set_current_disruption` into `set_target_node` to avoid this problem, and also introduce a new paramter to `set_target_node` so any user of it can set what's the data that would be saved on the target node (it's only for debugging)

Fixes: #8198

### Testing

- [x] :green_circle: https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-schema-topology-changes-12h-test/6/

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -380,7 +380,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     def set_target_node(self, dc_idx: Optional[int] = None, rack: Optional[int] = None,
                         is_seed: Union[bool, DefaultValue, None] = DefaultValue,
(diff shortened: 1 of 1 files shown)


fruch:
Need to cross check the logs that it now behaves as expected

fruch:
> Need to cross check the logs that it now behaves as expected

validated in the logs, we don't have case of `Set current_disruption -> None` anymore

vponomaryov (approved):
LGTM

PR #1624, opened 2019-12-26
Fix #1622: send_email command to handle empty email_recipients

aleksbykov:
Add to sct.py send_email checks:
if email-recipients is empty, don't send email

## PR pre-checks (self review)

- [ ] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [ ] I gave variables/functions meaningful self-explanatory names
- [ ] I didn't leave commented-out/debugging code
- [ ] I didn't copy-paste code
- [ ] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #8388, opened 2024-08-15
fix(scylla-bench): refator results mapping

fruch:
add information to each value in the mapping of it's type so we can correctly convert it, without generating needless warnings.

also adding a few missing value to the mapping

Fixes: #1624

### Testing

- [x] integration tests

### Reminders

### DIFF
--- a/sdcm/scylla_bench_thread.py
+++ b/sdcm/scylla_bench_thread.py
@@ -15,6 +15,7 @@ import os
 import re
 import uuid
 import time
+import builtins
 import logging
 import contextlib
 from enum import Enum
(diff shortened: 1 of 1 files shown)


vponomaryov:
> Fixes: #1624

The referenced thing is a PR, which looks unrelated.

fruch:
> > Fixes: #1624
> 
> The referenced thing is a PR, which looks unrelated.

My bad, it is the right number from the wrong repo

https://github.com/scylladb/qa-tasks/issues/1624

juliayakovlev (approved):

sdcm/scylla_bench_thread.py
```diff
@@ -78,30 +79,40 @@ class ScyllaBenchThread(DockerBasedStressThread):  # pylint: disable=too-many-in
...
-        '95th': 'latency 95th percentile',
-        '90th': '90th',
-        'median': 'latency median',
-        'Operations/s': 'op rate',
-        'Rows/s': 'row rate',
-        'Total ops': 'Total partitions',
-        'Time (avg)': 'Total operation time',
+        'Mode': (str, 'Mode'),
+        'Workload': (str, 'Workload'),
+        'Timeout': (int, 'Timeout'),
```
vponomaryov, line 84: Have you tested it?

Timeout may have letters like in the `1m` value. How it should be translated to an `int`eger?

fruch: convert_metric_to_ms seems to be supporting it.

Worst case it would be translated to string, currently no code is reading it

vponomaryov (approved):
LGTM

Issue #7261, opened 2024-03-10
From time to time monitor creation fails despite the instance was created


juliayakovlev:
From time to time monitor creation fails despite the instance was created. Error:
```
< t:2024-03-06 20:54:36,659 f:gce_utils.py    l:504  c:sdcm.utils.gce_utils p:DEBUG > Creating the rolling-upgrade--centos-8-monitor-node-aff52e4d-0-1 instance in us-east1-c...
< t:2024-03-06 20:54:38,001 f:retry.py        l:351  c:urllib3.util.retry   p:DEBUG > Converted retries value: 3 -> Retry(total=3, connect=None, read=None, redirect=None, status=None)
< t:2024-03-06 20:54:46,895 f:gce_utils.py    l:314  c:sdcm.utils.gce_utils p:DEBUG > Warnings during instance creation:
< t:2024-03-06 20:54:46,895 f:gce_utils.py    l:316  c:sdcm.utils.gce_utils p:DEBUG >  - DISK_SIZE_LARGER_THAN_IMAGE_SIZE: Disk size: '50 GB' is larger than image size: '20 GB'. You might need to resize the root
 repartition manually if the operating system does not support automatic resizing. See https://cloud.google.com/compute/docs/disks/add-persistent-disk#resize_pd for details.
< t:2024-03-06 20:54:46,895 f:gce_utils.py    l:510  c:sdcm.utils.gce_utils p:DEBUG > Instance rolling-upgrade--centos-8-monitor-node-aff52e4d-0-1 created.
< t:2024-03-06 20:54:46,927 f:decorators.py   l:72   c:sdcm.utils.decorators p:DEBUG > '_create_node_with_retries': failed with 'Forbidden("GET https://compute.googleapis.com/compute/v1/projects/sct-project-1/zones/us-east1-c/instances/rolling-upgrade--centos-8-monitor-node-aff52e4d-0-1: Quota exceeded for quota metric 'Read requests' and limit 'Read requests per minute per region' of service 'compute.googleapis.com' for consumer 'project_number:1070746702980'.")', retrying [#0]
```

After that we retry to create. And it fails again because the instance exists:
```
< t:2024-03-06 21:24:48,468 f:gce_utils.py    l:504  c:sdcm.utils.gce_utils p:DEBUG > Creating the rolling-upgrade--centos-8-monitor-node-aff52e4d-0-1 instance in us-east1-c...
[...]

PR #8389, opened 2024-08-15
fix(gce): delete instance in cases of retries

fruch:
since we are creating the instances with retries, we sometimes
get into case the first attempt failed cause of api quota issues
and on the retry we fail to create since instance already exists

this change is remove the limit to retry delete instance only when it's spot
and it delete any type of instance on case of failures related to spot or
to quota issues.

Fixes: #7261

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster_gce.py
+++ b/sdcm/cluster_gce.py
@@ -404,8 +404,7 @@ class GCECluster(cluster.BaseCluster):  # pylint: disable=too-many-instance-attr
         )
         instance = self._create_node_with_retries(name=name,
                                                   dc_idx=dc_idx,
-                                                  create_node_params=create_node_params,
-                                                  spot=spot)
+                                                  create_node_params=create_node_params)
 
         self.log.debug('Created %s instance %s', 'spot' if spot else 'on-demand', instance)
         try:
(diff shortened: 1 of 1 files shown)


fruch:
seems like in the case of spot instances, we destroy the instance after each failure, maybe that would be a safer option ?

sdcm/cluster_gce.py
```diff
@@ -428,6 +428,9 @@ def _create_node_with_retries(self,
         try:
             return create_instance(**create_node_params)
         except google.api_core.exceptions.GoogleAPIError as gbe:
+            if 'already exists' in str(gbe):
+                return self._get_instances_by_name(name=name, dc_idx=dc_idx)
```
vponomaryov, line 432: As I understand, it is useful on the 2nd+ attempt.
If so, then shouldn't we call that `_get_instances_by_name` method first and only after nothing found call the `create_instance` function?

soyacz: I think this happens when retrying instance creation and found that instance was already created

soyacz (dismissed):
LGTM

soyacz (approved):
LGTM

vponomaryov (changes requested):
Need to update commit message because it is misleading after the latest PR update.

The change itself LGTM

fruch:
> Need to update commit message because it is misleading after the latest PR update.
> 
> The change itself LGTM

update the commit message and the PR description

PR #8393, opened 2024-08-17
fix(restore-backup): Fix incorrect suffix on recover_conf()

### LINKED ISSUE #8392: Rollback test fails with "I/O Scheduler is not properly configured"
Rollback test fails with "I/O Scheduler is not properly configured".
This is because backup_conf() is backuping io.conf to io.conf.autobackup, but recover_conf() is trying to restore the file from io.conf.backup which is not available, so scylla-server unable to find correct io.conf settings.
It seems like the bug introduced at the commit e3bdc706c055d6164b5bc716016694446bdd3d17.

syuu1228:
The commit e3bdc70 mistakenly renamed suffix of backup config file, it become inconsistent between backup_conf() and recover_conf(). It should use .autobackup since backup_conf() uses it.

Fixes #8392
Fixes scylladb/scylladb#19777

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -117,7 +117,7 @@ def recover_conf(node):
         node.remoter.run(
             r'for conf in $( rpm -qc $(rpm -qa | grep scylla) | grep -v contains ) '
             r'/etc/systemd/system/{var-lib-scylla,var-lib-systemd-coredump}.mount; '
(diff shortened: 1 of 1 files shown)


vponomaryov:
It gets fixed in another existing PR here: https://github.com/scylladb/scylla-cluster-tests/pull/8393

syuu1228:
Closing since the PR was duplicated

Issue #8289, opened 2024-08-07
Test node operations with streaming (NOT with RBNO)


mykaul:
We apparently do not test node operations with streaming - only with RBNO, which is the default. That's fine for the majority of the tests, but we need some sanity around streaming. Please make sure we have simple add/decommission/replace nemsis tests with streaming (RBNO disabled).

CC @kbr-scylla

pehala:
@kbr-scylla @fruch 
Are you aware of any problems that `GrowShrinkClusterNemesis`, `TerminateAndRemoveNodeMonkey`, `NodeTerminateAndReplace`, `DecommissionMonkey` might have with streaming instead of RBNO? If not, then this task would be just adding the respective jobs/testcases into our regular CI?

mykaul:
> @kbr-scylla @fruch Are you aware of any problems that `GrowShrinkClusterNemesis`, `TerminateAndRemoveNodeMonkey`, `NodeTerminateAndReplace`, `DecommissionMonkey` might have with streaming instead of RBNO? If not, then this task would be just adding the respective jobs/testcases into our regular CI?

The only question is if we wish to clone more jobs and change the method from RBNO to streaming. I prefer converting SOME of the existing jobs to use streaming instead of RBNO. I understand the concern with randomizing - it is less predicatable - but it also has its value as well, as it'll ensure we continue to test more features, they will be tested with streaming. I do suggest we just change some existing longevities to use streaming for the time being.

fruch:
> > @kbr-scylla @fruch Are you aware of any problems that `GrowShrinkClusterNemesis`, `TerminateAndRemoveNodeMonkey`, `NodeTerminateAndReplace`, `DecommissionMonkey` might have with streaming instead of RBNO? If not, then this task would be just adding the respective jobs/testcases into our regular CI?
> 
> The only question is if we wish to clone more jobs and change the method from RBNO to streaming. I prefer converting SOME of the existing jobs to use streaming instead of RBNO. I understand the concern with randomizing - it is less predicatable - but it also has its value as well, as it'll ensure we continue to test more features, they will be tested with streaming. I do suggest we just change some existing longevities to use streaming for the time being.

Adding new, or converting some, both are o.k.

Someone that owns the feature and its testing can take those calls.

Since for quite some time it wasn't tested with streaming, we don't have any information regarding any part that is ready for that or not, at some point they did work with streaming.


Picking the cases, should be random, but it might.

[...]

PR #8402, opened 2024-08-19
Use streaming instead of RBNO for a few longevity tests

pehala:
### Overview
* Use streaming for one tier1 (`longevity-mv-si-4days`) and one tier2 (`longevity-cdc-8h-multi-dc-topology-changes`) longevity tests
    
### Testing

- [x] Run `longevity-mv-si-4days` in jenkins
- [ ] Run `longevity-cdc-8h-multi-dc-topology-changes` in jenkins

### Questions
* Are those two tests enough? Should I add more tests from different categories?
    * Are those even right tests? 
* How can I verify that it uses streaming?
* For how long do I need to run tests to verify correctness of this PR?

Fixes https://github.com/scylladb/scylla-cluster-tests/issues/8289

### DIFF
--- a/configurations/disable_rbno.yaml
+++ b/configurations/disable_rbno.yaml
@@ -1,0 +1,4 @@
+# configuration to disable RBNO (https://opensource.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/repair-based-node-operation.html)
+# and use good old streaming method
+append_scylla_yaml:
+  enable_repair_based_node_ops: false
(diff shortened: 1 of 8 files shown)


roydahan:
I'm not in favor of the renaming, it helps to have the history in one place.
If we do that renaming to make it clear that this one uses streaming, maybe it would be better to duplicate these and just replace the ones we have in in tier1 with them.
(i.e. move the originals to "longevity" and put the new ones in tier1, tier2).

fruch:
> I'm not in favor of the renaming, it helps to have the history in one place. If we do that renaming to make it clear that this one uses streaming, maybe it would be better to duplicate these and just replace the ones we have in in tier1 with them. (i.e. move the originals to "longevity" and put the new ones in tier1, tier2).

also we should add the new jobs into the tier1/teir2 triggers while we add them

pehala:
* Move old tests to `longevity`
* Change triggers to use new tests

roydahan (approved):
LGTM

fruch (approved):
LGTM

fruch:
@pehala 

A general comment, if you run any Jenkins jobs, please supply link to them or to Argus, so whom need would be able to cross check what was running

fruch:
@roydahan @pehala 

Do we want to change the triggers for release 2024.1 as well ?

roydahan:
Preferably yes, assuming we tested them.

PR #8403, opened 2024-08-20
improvement(stress): Report two lines of the error, instead of only 100 characters

### LINKED ISSUE #8291: Truncated log message in the event
SCT truncates the log messages, so the full message is not visible in the events.
Job example: https://argus.scylladb.com/test/4051b1c9-c265-4cdb-bf19-e24748ed3302/runs?additionalRuns[]=fdaae0bd-f82c-4726-a498-7edbd36aa7ef

Actual event message:
```
Stress command completed with bad status 1: com.datastax.driver.core.exceptions.InvalidQueryException: Cannot create CDC log for a table cdc_tes
```
But the full message from the cassandra-stress log is actualy reports the root cause of the problem: #16317 :
```
Stress command completed with bad status 1: com.datastax.driver.core.exceptions.InvalidQueryException: Cannot create CDC log for a table cdc_test.test_table, because keyspace uses tablets. See issue #16317.
```

so, we should print the full message instead of truncating it to speed up Agrus investigations

pehala:
Fixes #8291

### Testing

- [x] Locally, using wrong stress command

### DIFF
--- a/sdcm/stress/base.py
+++ b/sdcm/stress/base.py
@@ -151,5 +151,7 @@ def format_stress_cmd_error(exc: Exception) -> str:
     """Format nicely the exception from a stress command failure."""
 
     if hasattr(exc, "result") and exc.result.failed:
-        return f"Stress command completed with bad status {exc.result.exited}: {exc.result.stderr:.100}"
+        # Report only first two lines
+        message = "\n".join(exc.result.stderr.splitlines()[:2])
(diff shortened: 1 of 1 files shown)


vponomaryov (approved):
LGTM

fruch (approved):

soyacz (approved):

PR #8447, opened 2024-08-27
improvement(perf-simple-query): send results to argus

soyacz:
Sending perf-simple-query results to Argus.

refs: https://github.com/scylladb/qa-tasks/issues/1244

### Testing

- [x] - [perf-simple-query test - argus](https://argus.scylladb.com/test/20c7e63e-f5ba-4ca2-946a-efec89b57949/runs?additionalRuns[]=2ba20be4-3f86-49e1-9f47-f3431eea2a82)

### Reminders

### DIFF
--- a/microbenchmarking_test.py
+++ b/microbenchmarking_test.py
@@ -11,6 +11,8 @@
 #
 # Copyright (c) 2023 ScyllaDB
 import json
+
+from sdcm.argus_results import send_perf_simple_query_result_to_argus
 from sdcm.tester import ClusterTester, teardown_on_exception, log_run_info
 from sdcm.utils.microbenchmarking.perf_simple_query_reporter import PerfSimpleQueryAnalyzer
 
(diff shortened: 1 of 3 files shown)


soyacz:
See this implementation sends results as is without assessing if there's a regression (all pass).
To do that, I'd need to change the logic behind comparing to previous results - currently it is done by assessing **previous** results instead of current one. I think this needs to be redesigned to validate **current** result against previous one.

soyacz:
found issue in Argus preventing from displaying perf-simple-query results. Fix: https://github.com/scylladb/argus/pull/436

soyacz:
this is how it would look if argus was working:
image

fruch:
> See this implementation sends results as is without assessing if there's a regression (all pass). To do that, I'd need to change the logic behind comparing to previous results - currently it is done by assessing **previous** results instead of current one. I think this needs to be redesigned to validate **current** result against previous one.

is this compare put the currently results into ES before it does the checks ?

anyhow, why does it make any difference to how the logic is done ? can we report to argus after the fact ?

soyacz:
> > See this implementation sends results as is without assessing if there's a regression (all pass). To do that, I'd need to change the logic behind comparing to previous results - currently it is done by assessing **previous** results instead of current one. I think this needs to be redesigned to validate **current** result against previous one.
> 
> is this compare put the currently results into ES before it does the checks ?
> 
> anyhow, why does it make any difference to how the logic is done ? can we report to argus after the fact ?

after rethinking I can make it working as expected (or similar to the email report): if any of previous result in given column is not within limits then it's going to be reported as a failure.

soyacz:
ok, fixed. Now it looks like this:
image

soyacz:
@fruch one more thing I can change, but I'm not sure if it brings benefits:
when sending result rows, instead of #1 label I can add scylla major version like `6.2.0~dev`. So in case new scylla release is created, next results will come as `6.3.0~dev` creating a new 'series' in Argus's graphs showing distinction between them. 
WDYT?

fruch:
> @fruch one more thing I can change, but I'm not sure if it brings benefits: when sending result rows, instead of `#1` label I can add scylla major version like `6.2.0~dev`. So in case new scylla release is created, next results will come as `6.3.0~dev` creating a new 'series' in Argus's graphs showing distinction between them. WDYT?

if we have graph of those results, it's doesn't really needed. the group would be show the history as needed.

you can ask for feedback on the weekly meeting next week (and maybe show the argus area, to get some feedback of it)

fruch (approved):
LGTM

soyacz:
@juliayakovlev can you take a look if this LGTY?

juliayakovlev:
> ok, fixed. Now it looks like this: image

Commonly it looks OK. Just what are the rectangles in the table headers?

soyacz:
> 
> Commonly it looks OK. Just what are the rectangles in the table headers?

these are missing units - need to fix it in Argus to not show them if unit is not present.

juliayakovlev (approved):

PR #8462, opened 2024-08-27
ci(provisionResources): set SCT_N_DB_NODES variable on provision stage

mikliapko:
Having this parameter SCT_N_DB_NODES set, allows to vary the number of nodes in Manager jobs where the input param (n_db_nodes) exists and is used from time to time.

### Testing

- [x] [Test](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-installation-test/6/) with set n_db_nodes value (3) in Jenkins job. Default value defined by test-case.yaml is 1. The test was started with 3 DB nodes.

### Reminders

### DIFF
--- a/vars/provisionResources.groovy
+++ b/vars/provisionResources.groovy
@@ -130,6 +130,10 @@ def call(Map params, String region){
         export SCT_MANAGER_VERSION="${params.manager_version}"
     fi
 
+    if [[ -n "${params.n_db_nodes ? params.n_db_nodes : ''}" ]] ; then
+        export SCT_N_DB_NODES=${params.n_db_nodes}
+    fi
+
     if [[ -n "${params.pytest_addopts ? params.pytest_addopts : ''}" ]] ; then
         export PYTEST_ADDOPTS="${params.pytest_addopts}"
     fi


vars/provisionResources.groovy
```diff
@@ -130,6 +130,10 @@ def call(Map params, String region){
         export SCT_MANAGER_VERSION="${params.manager_version}"
     fi
 
+    if [[ -n "${params.n_db_nodes ? params.n_db_nodes : ''}" ]] ; then
```
fruch, line 133: I don't sure I like this idea of have this as pipeline parameters

seem that I've missed that you introduced it manager pipeline.

is that used in any "formal" triggering ? or just for experimentations ?

fruch: if it's just for experimentation, https://github.com/scylladb/scylla-cluster-tests/pull/8268 might be a better fit

mikliapko: not formal triggers, only manual use

mikliapko: I see your solution, will try to add this param in Manager pipeline

vponomaryov: @fruch 

This change utilizes already existing pipeline parameter, it doesn't add one new to the user interface.
It should be totally ok to do so.

The https://github.com/scylladb/scylla-cluster-tests/pull/8268 looks unrelaed to the goal of this PR.

vponomaryov (approved):
LGTM

vars/provisionResources.groovy
mikliapko: no, I haven't verified longevity pipeline.
So, if there is such probability we will just go with solution suggested by Israel (running the test to verify it right now) and close this one without merging.

mikliapko: Yes, it works, closing this PR:
https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-installation-test/7/

PR #8484, opened 2024-08-28
test(integration): add compression tests for c-s

fruch:
since we run into issues c-s around snappy compression we should have a set of tests that can help us identify those kind of issues

### Testing

- [ ]

### Reminders

### DIFF
--- a/unit_tests/test_cassandra_stress_thread.py
+++ b/unit_tests/test_cassandra_stress_thread.py
@@ -141,3 +141,33 @@ def test_04_cassandra_stress_multi_region(request, docker_scylla, params):
 
     assert "latency 99th percentile" in output[0]
     assert float(output[0]["latency 99th percentile"]) > 0
+
+
+@pytest.mark.parametrize("compressor, cql_compression", [
+    ("Deflate", "none"), ("LZ4", "lz4"), ("Snappy", "snappy"), ("Zstd", "none")
+])
+def test_05_cassandra_stress_compression(request, docker_scylla, params, compressor, cql_compression):
+    loader_set = LocalLoaderSetDummy(params=params)
+
+    cmd = (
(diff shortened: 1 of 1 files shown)


mykaul:
That was a bug/issue in cassandra-stress - shouldn't cassandra-stress CI have such a test? It's not a scylla issue?

unit_tests/test_cassandra_stress_thread.py
```diff
@@ -141,3 +141,33 @@ def cleanup_thread():
 
     assert "latency 99th percentile" in output[0]
     assert float(output[0]["latency 99th percentile"]) > 0
+
+
+@pytest.mark.parametrize("compressor, cql_compression", [
+    ("Deflate", "none"), ("LZ4", "lz4"), ("Snappy", "snappy"), ("Zstd", "none")
```
mykaul, line 147: I didn't know we supported deflate for CQL compression? I think we are mixing sstable compression and CQL protocol level compression?

CodeLieutenant: Here CQL compression for Deflate and ZStd is none, so there will be no compression in those cases, and the first compression is for Scylla (schema level) and the later one is compression for the protocol, which tipped me off guard when first checking.

There is a `'` after Compressor making it a part of -schema argument

CodeLieutenant (approved):

vponomaryov (approved):
LGTM

fruch:
> That was a bug/issue in cassandra-stress - shouldn't cassandra-stress CI have such a test? It's not a scylla issue?

yes, right now we don't have integration test on c-s at all, while we do have them for SCT.

unit_tests/test_cassandra_stress_thread.py
fruch: yes it's mixing it, the issue we had was with cql compression, that only support snappy / lz4

Issue #7531, opened 2024-06-02
disrupt_grow_shrink_cluster should be marked as 'skipped' if it failed to grow as it could not find instances to be allocated for it


mykaul:
Example:
```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5111, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4005, in disrupt_grow_shrink_cluster
    self._grow_cluster(rack=None)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4029, in _grow_cluster
    added_node = self.add_new_node(rack=rack_idx)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 178, in wrapped
    res = func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3973, in add_new_node
    return self._add_and_init_new_cluster_node(rack=rack)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1231, in _add_and_init_new_cluster_node
    new_node = self.cluster.add_nodes(
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster_aws.py", line 892, in add_nodes
    added_nodes = super().add_nodes(
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster_aws.py", line 387, in add_nodes
    instances = self._create_or_find_instances(
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster_aws.py", line 366, in _create_or_find_instances
    return self._create_instances(count, ec2_user_data, dc_idx, az_idx, instance_type=instance_type)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster_aws.py", line 229, in _create_instances
    instances = self._create_on_demand_instances(count, interfaces, ec2_user_data, dc_idx,
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster_aws.py", line 158, in _create_on_demand_instances
    instances = self._ec2_services[dc_idx].create_instances(**params)
  File "/usr/local/lib/python3.10/site-packages/boto3/resources/factory.py", line 580, in do_action
    response = action(self, *args, **kwargs)
[...]

PR #8513, opened 2024-08-28
fix(nemesis): skip nemesis if there is capacity errors

fruch:
if there's capacity error during a nemesis, we don't have much we can do about it, so we are gonna skip the nemesis insted just failing the nemesis

this change introduces a new decorator to help doing it on multiple calls in nemesis.py, later it can be extended to handle also other backend capacity issues (GCE/Azure), for now it handles only AWS capacity errors

Fixes: #7531

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/decorators.py
+++ b/sdcm/utils/decorators.py
@@ -17,15 +17,15 @@ import logging
 import datetime
 import json
 import os
-
 from functools import wraps, partial, cached_property
 from typing import Optional, Callable
 
+from botocore.exceptions import ClientError
+
 from sdcm.argus_results import send_result_to_argus
 from sdcm.sct_events.database import DatabaseLogEvent
-
 from sdcm.sct_events.event_counter import EventCounterContextManager
-
+from sdcm.exceptions import UnsupportedNemesis
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


sdcm/utils/decorators.py
```diff
@@ -285,3 +285,18 @@ def static_init(cls):
...
+    """
+    Decorator to skip nemesis that fail due to capacity issues
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except ClientError as ex:
+            if "InsufficientInstanceCapacity" in str(ex):
+                raise UnsupportedNemesis("Capacity Issue") from ex
```
fruch, line 300: I wonder if we want this behavior in performance tests, skipping the nemesis would change the behavior, and we can get empty report that can skew the results, maybe we better off with failing the nemesis ? 

so maybe introducing a kill switch for this feature ?, so we can disable it for perf tests ?

soyacz: in perf tests we `use_capacity_reservation: true` where it won't start the test without capacity, possibly need to add to more tests (we use it only when placement groups are true too).

fruch: It can still happen if we miss calculating, or if one wood alters the test, for example different instance types on grow ?

soyacz: different instance types on grow & capacity reservation: https://github.com/scylladb/scylla-cluster-tests/issues/8514

And if one alters the test on the runtime... I don't think we're going to do that in perf tests.

fruch: so this change is good to go from you POV ?

Issue #8527, opened 2024-09-02
artifact tests failed with 'NoneType' object is not subscriptable


Annamikhlin:
### Issue description

artifact tests failed with error:

```
01:39:09  ======================================================================
01:39:09  ERROR: test_scylla_service (artifacts_test.ArtifactsTest)
01:39:09  ----------------------------------------------------------------------
01:39:09  Traceback (most recent call last):
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/artifacts_test.py", line 50, in setUp
01:39:09      super().setUp()
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/sdcm/tester.py", line 185, in wrapper
01:39:09      return method(*args, **kwargs)
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/sdcm/utils/decorators.py", line 121, in inner
01:39:09      res = func(*args, **kwargs)
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/sdcm/tester.py", line 958, in setUp
01:39:09      self.init_resources()
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/sdcm/tester.py", line 1913, in init_resources
01:39:09      self.get_cluster_gce(loader_info=loader_info, db_info=db_info,
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/sdcm/tester.py", line 1256, in get_cluster_gce
01:39:09      self.db_cluster = ScyllaGCECluster(gce_image=self.params.get('gce_image_db').strip(),
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/sdcm/cluster_gce.py", line 548, in __init__
01:39:09      super().__init__(
01:39:09    File "/tmp/workspace/scylla-master/artifacts/artifacts-centos9-test/scylla-cluster-tests/sdcm/cluster.py", line 3968, in __init__
01:39:09      super().__init__(*args, **kwargs)
[...]

PR #8531, opened 2024-09-02
fix(artifacts): skip configuring syslog-ng for artifacts test

dimakr:
ConfigurationScriptBuilder should skip syslog-ng setup steps for artifacts tests, which do not use syslog-ng logs transport.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/8527

### Testing
- [x] :green_circle: [artifacts-ubuntu2204-test](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/artifacts-ubuntu2204-test/2/)
- [x] :green_circle: regular PR provision test + cluster reuse flow were executed (and pass) as part PR CI

### Reminders

### DIFF
--- a/sdcm/provision/common/configuration_script.py
+++ b/sdcm/provision/common/configuration_script.py
@@ -86,10 +86,11 @@ class ConfigurationScriptBuilder(AttrBuilder, metaclass=abc.ABCMeta):
         # 5. Make sure that whenever you use "cat <<EOF >>/file", make sure that EOF has no spaces in front of it
         script = ''
 
-        script += configure_syslogng_destination_conf(
-            host=self.syslog_host_port[0],
-            port=self.syslog_host_port[1],
-            throttle_per_second=SYSLOGNG_LOG_THROTTLE_PER_SECOND)
+        if self.logs_transport == 'syslog-ng':
+            script += configure_syslogng_destination_conf(
+                host=self.syslog_host_port[0],
+                port=self.syslog_host_port[1],
+                throttle_per_second=SYSLOGNG_LOG_THROTTLE_PER_SECOND)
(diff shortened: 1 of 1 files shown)


sdcm/provision/common/configuration_script.py
```diff
@@ -86,10 +86,11 @@ def _script_body(self) -> str:
         # 5. Make sure that whenever you use "cat <<EOF >>/file", make sure that EOF has no spaces in front of it
         script = ''
 
-        script += configure_syslogng_destination_conf(
-            host=self.syslog_host_port[0],
-            port=self.syslog_host_port[1],
-            throttle_per_second=SYSLOGNG_LOG_THROTTLE_PER_SECOND)
+        if self.logs_transport == 'syslog-ng':
+            script += configure_syslogng_destination_conf(
```
Annamikhlin, line 90: Why not to add it under if statement on line 96?
https://github.com/scylladb/scylla-cluster-tests/blob/2223ba3cd8b5373e09d976b0b7984fbbd36208b5/sdcm/provision/common/configuration_script.py#L96

dimakr: We need to run `configure_syslogng_destination_conf` before this statement `script += self._skip_if_already_run()`
This `configure_syslogng_destination_conf` function prepares config for where logs will be sent. And if it is a new node then that skip check will have no effect and the remaining part of syslon-ng setup follows. If it is an existing node (when reusing the cluster), we need to just update logs destination and skip the syslog-ng setup.

This PR is just a quick fix. And the plan is to switch to syslog-ng for artifacts as well. So this and other `logs_transport` related ifs will disappear from this class.

fruch (approved):
LGTM

PR #8563, opened 2024-09-05
ci(custom-d1): add missing `row_count` option in the `prepare write` step

vponomaryov:
Without this option the default value `1000000` was used.
And this default value is much smaller than the desired value which was set only with the `-d` key.
    
So, explicitly defined `row_count` option in addition to the `-d` one
with the same values just to write expected amount of data once.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

soyacz (approved):
LGTM

PR #8564, opened 2024-09-05
fix(hdr file): latency from HDR file for untrottled load

juliayakovlev:
When c-s load runs without fixed/throttle, the histogram data with WRITE-rt/READ-rt (latency) tags is not created in the HDR file.
In this case all statistics will be reported with WRITE-st/READ-st (throughput) tags. It is according to the cassandra-stress code.
So if rows with the hdr_tag is not found - try to find data with another tag

### Testing

- [x] [perf-regression-latency-650gb-during-rolling-upgrade](https://argus.scylladb.com/test/8d048290-f7f4-4af5-9e09-882202a35656/runs?additionalRuns[]=d1ade4c1-255f-4195-bac7-15e535562b2f)

### Reminders

### DIFF
--- a/sdcm/utils/csrangehistogram.py
+++ b/sdcm/utils/csrangehistogram.py
@@ -1,3 +1,4 @@
+import copy
 import glob
 import os.path
 import sys
@@ -82,6 +83,10 @@ def make_cs_range_histogram_summary_from_log_line(
 class _CSHistogramThroughputTags(Enum):
     WRITE = "WRITE-st"
     READ = "READ-st"
+
+
+TAG_PAIRS = {"WRITE": [_CSHistogramThroughputTags.WRITE.value, CSHistogramTags.WRITE.value],
+             "READ": [_CSHistogramThroughputTags.READ.value, CSHistogramTags.READ.value]}
 
 
 _CSHISTOGRAM_TAGS_MAPPING = {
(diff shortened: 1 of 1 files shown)


sdcm/utils/csrangehistogram.py
```diff
@@ -229,25 +229,62 @@ def build_from_log_line(self, log_line: str, hst_log_start_time: float) -> dict[
...
+    def switch_tag(hdr_tag):
+        tag_pairs = [[_CSHistogramThroughputTags.WRITE.value, CSHistogramTags.WRITE.value],
+                     [_CSHistogramThroughputTags.READ.value, CSHistogramTags.READ.value]]
+        for pair in tag_pairs:
+            if hdr_tag in pair:
+                return [tag for tag in pair if tag != hdr_tag][0]
+        return None
+
+    def _build_histogram_from_file(self, hdr_file: str, hdr_tag: str) -> _CSRangeHistogram | None:
+        def analyze_hdr_file():
```
fruch, line 242: would  be nice to have return value documented for this internal one

sdcm/utils/csrangehistogram.py
```diff
@@ -229,25 +229,62 @@ def build_from_log_line(self, log_line: str, hst_log_start_time: float) -> dict[
...
 
-    def _build_histogram_from_file(self, hdr_file: str, hdr_tag: str) -> _CSRangeHistogram:
+    @staticmethod
+    def switch_tag(hdr_tag):
+        tag_pairs = [[_CSHistogramThroughputTags.WRITE.value, CSHistogramTags.WRITE.value],
+                     [_CSHistogramThroughputTags.READ.value, CSHistogramTags.READ.value]]
+        for pair in tag_pairs:
+            if hdr_tag in pair:
+                return [tag for tag in pair if tag != hdr_tag][0]
+        return None
```
fruch, line 239: why not just a mapping like that:
```python
def switch_tag(hdr_tag):
    mapping = {_CSHistogramThroughputTags.WRITE.value: CSHistogramTags.WRITE.value,
                       _CSHistogramThroughputTags.READ.value: CSHistogramTags.READ.value}
    # extend it with the reverse mapping
    mapping.update(dict(map(reversed, mapping.items())))
    return mapping.get(hdr_tag)
```

fruch: also it's might be called `fallback_tag` (if at some point we'll have more then 2 available)

juliayakovlev: do you mean it may be something except of throughput and latency?

sdcm/utils/csrangehistogram.py
juliayakovlev: added

sdcm/utils/csrangehistogram.py
juliayakovlev: > why not just a mapping like that:
> 
> ```python
> def switch_tag(hdr_tag):
>     mapping = {_CSHistogramThroughputTags.WRITE.value: CSHistogramTags.WRITE.value,
>                        _CSHistogramThroughputTags.READ.value: CSHistogramTags.READ.value}
>     # extend it with the reverse mapping
>     mapping.update(dict(map(reversed, mapping.items())))
>     return mapping.get(hdr_tag)
> ```

done

fruch: yes, right now there are 3 types of tags
there one for the wait time, but we don't care about it on it's own.

tomorrow we'll send add some other metric for some reason into the hdr (again maybe not)

juliayakovlev: @fruch 
I changed the function. Please, review

fruch (approved):
LGTM

juliayakovlev:
> LGTM

@fruch I think it has a problem. Checking

juliayakovlev:
> > LGTM
> 
> @fruch I think it has a problem. Checking

@fruch HDR data collected and reported. Test passed. Please, review

fruch (approved):
LGTM

PR #8565, opened 2024-09-05
[Manager] Test to run restore with L&S outside of Manager

mikliapko:
Closes https://github.com/scylladb/scylla-manager/issues/3993

The PR introduces the test that validate restore with L&S outside of Scylla Manager.
In addition to the test, contextmanager class that allows to measure execution time of wrapped function was added.

### Testing

- [x] no extra params [run](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/backup-restore-outside-manager/37/);
- [x] with extra params [run](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/backup-restore-outside-manager/38/) 
- [x] [regression](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-sanity-test/27/) 

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -41,6 +41,7 @@ from sdcm.nemesis import MgmtRepair
 from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
 from sdcm.utils.common import reach_enospc_on_node, clean_enospc_on_node
 from sdcm.utils.loader_utils import LoaderUtilsMixin
+from sdcm.utils.time_utils import ExecutionTimer
 from sdcm.sct_events.system import InfoEvent
 from sdcm.sct_events.group_common_events import ignore_no_space_errors, ignore_stream_mutation_fragments_errors
 from sdcm.utils.compaction_ops import CompactionOps
@@ -68,14 +69,17 @@ class SnapshotData:
     - prohibit_verification_read: if True, the verification read will be prohibited. Most likely, such a backup was
       created via c-s user profile.
     - number_of_rows: number of data rows written in the DB
+    - node_ids: list of node ids where backup was created
     """
     bucket: str
     tag: str
     exp_timeout: int
     keyspaces: list[str]
+    ks_tables_map: dict[str, list[str]]
     cs_read_cmd_template: str
     prohibit_verification_read: bool
     number_of_rows: int
+    node_ids: list[str]
 
 
 class BackupFunctionsMixIn(LoaderUtilsMixin):
@@ -174,7 +178,20 @@ class BackupFunctionsMixIn(LoaderUtilsMixin):
             return base_id.replace('-', '')
         return base_id
 
-    def restore_backup(self, mgr_cluster, snapshot_tag, keyspace_and_table_list):  # pylint: disable=too-many-locals
+    def restore_backup_without_manager(self, mgr_cluster, snapshot_tag, ks_tables_list, location=None,
+                                       precreated_backup=False):
+        """Restore backup without Scylla Manager but using the `nodetool refresh` operation
+        (https://opensource.docs.scylladb.com/stable/operating-scylla/nodetool-commands/refresh.html).
+
+        The method downloads backup files from the backup bucket (aws s3 cp ...) and then loads them (nodetool refresh)
+        into the cluster node by node and table by table.
+
+        Two flows are supported:
+          - restore to a new cluster from pre-created backup.
+          - restore from backup created at the same cluster.
+        The difference between these two flows is in the way the node_id is retrieved - if the backup was created with
+        the cluster under test use node_ids of cluster under test, otherwise, get node_id from `sctool backup files...`
+        """
         backup_bucket_backend = self.params.get("backup_bucket_backend")
         if backup_bucket_backend == "s3":
             install_dependencies = self.install_awscli_dependencies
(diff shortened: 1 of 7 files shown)


sdcm/utils/time_utils.py
```diff
@@ -0,0 +1,27 @@
...
+    def __enter__(self):
+        self.start_time = datetime.now()
+        return self
+
+    def __exit__(self, exc_type, exc_value, traceback):
+        self.end_time = datetime.now()
+
+    @property
+    def duration(self) -> timedelta:
+        return self.end_time - self.start_time
```
roydahan, line 27: IIRC we have something else already that you can use here.

test-cases/manager/manager-restore-outside-with-las-and-primary-replica-only.yaml
```diff
@@ -0,0 +1,14 @@
+test_duration: 600
```
roydahan, line 1: Seems like a dup of the second one with just 1-2 additional yamls, so they can share the same yaml and this one will be additional configuration on top of it.
It makes it easier when you want to do the same change in both tests in order to do comparisons.

mgmt_cli_test.py
```diff
@@ -188,25 +193,42 @@ def restore_backup(self, mgr_cluster, snapshot_tag, keyspace_and_table_list):  #
         else:
             raise ValueError(f'{backup_bucket_backend=} is not supported')
 
-        per_node_backup_file_paths = mgr_cluster.get_backup_files_dict(snapshot_tag)
-        for node in self.db_cluster.nodes:
+        nodetool_refresh_extra_flags = self.params.get('mgmt_nodetool_refresh_flags')
+        if not nodetool_refresh_extra_flags:
+            nodetool_refresh_extra_flags = ""
```
soyacz, line 198: can be reduced to `nodetool_refresh_extra_flags = self.params.get('mgmt_nodetool_refresh_flags') or ""`

mgmt_cli_test.py
```diff
@@ -174,7 +178,8 @@ def get_table_id(self, node, table_name, keyspace_name=None, remove_hyphen=True)
             return base_id.replace('-', '')
         return base_id
 
-    def restore_backup(self, mgr_cluster, snapshot_tag, keyspace_and_table_list):  # pylint: disable=too-many-locals
+    def restore_backup_outside_of_manager(self, mgr_cluster, snapshot_tag, ks_tables_list, location=None,
```
soyacz, line 181: I think this method requires docstring - telling what it does, limitations etc. 
It's not obvious for 'outside' users to understand what 'outside_of_manager' means (for me it would be better `restore_backup_without_manager`)
Please also add link to official scylla docs on this procedure.

soyacz (commented):
A few remarks.
One  more thing. It was not yet spread to wider audience (soon I hope I'll finish the feature completely and do it), but you can store durations in Argus and easily view them in "Results" tab, and even graph across different runs of the same test to spot regressions.
You can play by adding:
```
def send_backup_duration_to_argus(argus_client: ArgusClient, duration: int, series: str = "SM", status: Status = Status.PASS):
    class BackupDurationResult(GenericResultTable):
        class Meta:
            name = "Backup Duration"
            description = "A Description of backup - shared between all runs in jenkins job"
            Columns = [
                ColumnMetadata(name="Duration", unit="HH:MM:SS", type=ResultType.DURATION)
            ]

    result_table = BackupDurationResult()
    result_table.add_result(column="Duration", row=series, value=duration, status=status)
    argus_client.submit_results(result_table)
```
To `sdcm/argus_results.py` and use this function accordingly (see how others are used).
In case you have questions, let me know.

sdcm/utils/time_utils.py
mikliapko: I tried to find something like this but didn't

test-cases/manager/manager-restore-outside-with-las-and-primary-replica-only.yaml
mikliapko: make sense, will fix it

mgmt_cli_test.py
mikliapko: Agree, will fix it

mikliapko:
> A few remarks. One more thing. It was not yet spread to wider audience (soon I hope I'll finish the feature completely and do it), but you can store durations in Argus and easily view them in "Results" tab, and even graph across different runs of the same test to spot regressions. You can play by adding:
> 
> ```
> def send_backup_duration_to_argus(argus_client: ArgusClient, duration: int, series: str = "SM", status: Status = Status.PASS):
>     class BackupDurationResult(GenericResultTable):
>         class Meta:
>             name = "Backup Duration"
>             description = "A Description of backup - shared between all runs in jenkins job"
>             Columns = [
>                 ColumnMetadata(name="Duration", unit="HH:MM:SS", type=ResultType.DURATION)
>             ]
> 
>     result_table = BackupDurationResult()
>     result_table.add_result(column="Duration", row=series, value=duration, status=status)
>     argus_client.submit_results(result_table)
> ```
> 
> To `sdcm/argus_results.py` and use this function accordingly (see how others are used). In case you have questions, let me know.

Thanks for describing it here.
It'd be very useful improvement not only for this PR but for restore benchmark tests we've introduced recently too.
I decided to move implementation into separated [ticket](https://github.com/scylladb/scylla-manager/issues/4027) since it's not directly related to this PR.

mikliapko:
All the comments have been addressed, retesting done:
https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/backup-restore-outside-manager/40/

fruch (approved):
LGTM

sdcm/utils/time_utils.py
roydahan: I think something like "duration".
If you want to expand it and measure also latency, you should have latency_calculator which measures both latency and duration.

Issue #8514, opened 2024-08-29
Capacity reservation reserves wrong instances when nemesis_grow_shrink_instance_type is used


soyacz:
When using `nemesis_grow_shrink_instance_type` setting, capacity reservation code does not take it into account and reserves wrong instances for the test.

We need to adjust `sdcm.provision.aws.capacity_reservation.SCTCapacityReservation._get_cr_request_based_on_sct_config` method.

PR #8586, opened 2024-09-08
fix(capacity_reservation): take `nemesis_grow_shrink_instance_type` into account

fruch:
`nemesis_grow_shrink_instance_type` was introduced, but on the capacity_reservation logic we didn't took it into account

Fixes: #8514

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/provision/aws/capacity_reservation.py
+++ b/sdcm/provision/aws/capacity_reservation.py
@@ -14,6 +14,7 @@ import logging
 import time
 from datetime import datetime, timedelta
 from typing import List, Dict, Tuple
+from collections import defaultdict
 
 from botocore.exceptions import ClientError
 import boto3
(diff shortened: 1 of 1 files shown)


sdcm/provision/aws/capacity_reservation.py
```diff
@@ -33,7 +33,14 @@ class SCTCapacityReservation:
     def _get_cr_request_based_on_sct_config(params) -> Tuple[dict[str, int], int]:
         instance_counts = {}
         nemesis_node_count = params.get("nemesis_add_node_cnt") or 0
-        cluster_max_size = (params.get("cluster_target_size") or params.get("n_db_nodes")) + nemesis_node_count
+
+        cluster_max_size = (params.get("cluster_target_size") or params.get("n_db_nodes"))
+
+        if nemesis_grow_shrink_instance_type := params.get("nemesis_grow_shrink_instance_type"):
+            instance_counts[nemesis_grow_shrink_instance_type] = nemesis_node_count
```
soyacz, line 40: in unrealistic scenario when `nemesis_grow_shrink_instance_type` is the same like `instance_type_db` or like monitoring/loader node type then this cause problems. 
```
instance_counts[nemesis_grow_shrink_instance_type] = instance_counts.get(nemesis_grow_shrink_instance_type, 0) + nemesis_node_count
```

soyacz (approved):
Besides of error for unrealistic scenario.
LGTM

sdcm/provision/aws/capacity_reservation.py
fruch: we can guard that in the configuration itself, i.e. forbid using the same instance type as db as part of validation ?

soyacz: I think above proposal is easier and I think we don't have to restrict it then.

fruch: why not use defaultdict with 0 ? and add to it as needed ?

soyacz: also can be done this way.

fruch: it's looks much better now

I've test it with the unittests, but didn't want to complex them with add this logic into them

PR #8588, opened 2024-09-09
fix(fips): pick ubuntu pro fips straight from marketplace 

fruch:
- **fix(fips): pick ubuntu pro fips straight from marketplace**

  take latest images from marketplace so we won't need to manually
  update those configuration with newer AMIs

- **fix(fips): switch fips check to be based on sct configuration**

  since we have cases we'll want to run ubuntu pro (EOL versions as
  example), that won't have fips enable, the assumption that ubuntu pro
  mean only the fips images, isn't valid anymore.
  
  in this change we introduce a new configuration option, that
  would help us list the distro features we expect, so the test
  would be able to verify those on installation.

### Testing

- [x] 🟢  artifact test - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-ubuntu2004-fips-test/13/
- [ ] longevity test
 
### Reminders


configurations/longevity-fips-and-encryptions.yaml
```diff
@@ -1,4 +1,5 @@
-ami_id_db_scylla: 'ubuntu-pro-fips-server/images/hvm-ssd/ubuntu-focal-20.04-amd64-pro-fips-server-20221121-7bc828d1-c072-4d33-a989-fbad50380cfb'
+ami_id_db_scylla: 'resolve:ssm:/aws/service/marketplace/prod-k6fgbnayirmrc/latest' # Ubuntu Pro FIPS 20.04 LTS amd64
+linux_distro_features: ['fips']
```
soyacz, line 2: I'd call it `assert_linux_distro_features`

sdcm/cluster.py
```diff
@@ -4616,13 +4616,15 @@ def node_setup(self, node: BaseNode, verbose: bool = False, timeout: int = 3600)
         if self.params.get('logs_transport') == 'ssh':
             node.install_package('python3')
 
-        if node.distro.is_ubuntu:
-            result = node.remoter.sudo("pro status", ignore_status=True)
+        # check fips enabled only if we specify it in sct config (so we can use ubuntu pro in other cases)
+        if node.distro.is_ubuntu and 'fips' in self.params.get('linux_distro_features'):
```
vponomaryov, line 4620: The `linux_distro_features` may be `None` and in this case we will get failure in the condition itself.
So, need to make it be safe coded.

fruch: it's has a default in `test_default.yaml`, which should suffice.

but I've added a check anyhow

configurations/longevity-fips-and-encryptions.yaml
fruch: change the name of it

soyacz (approved):
LGTM

PR #8589, opened 2024-09-09
fix(artifacts): add rocky9-selinux

fruch:
since there are bugs related centos9 stream that we might not gonna attend to.
adding the test case to work also with rocky9

Ref: https://github.com/scylladb/scylladb/issues/19325

### Testing

- [x] 🔴 rocky9-selinux - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-rocky9-selinux-test/2/
- [x] run with a fix from @syuu1228 -

### Reminders

### DIFF
--- a/jenkins-pipelines/oss/artifacts/artifacts-rocky9-selinux.jenkinsfile
+++ b/jenkins-pipelines/oss/artifacts/artifacts-rocky9-selinux.jenkinsfile
@@ -1,0 +1,13 @@
+#! groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+artifactsPipeline(
+    test_config: 'test-cases/artifacts/rocky9-selinux.yaml',
+    backend: 'gce',
+    provision_type: 'spot',
+
+    timeout: [time: 30, unit: 'MINUTES'],
+    post_behavior_db_nodes: 'destroy'
+)
--- a/test-cases/artifacts/rocky9-selinux.yaml
+++ b/test-cases/artifacts/rocky9-selinux.yaml
@@ -1,0 +1,21 @@
+backtrace_decoding: false
+cluster_backend: 'gce'
+gce_image_db: 'https://www.googleapis.com/compute/v1/projects/rocky-linux-cloud/global/images/family/rocky-linux-9'
+gce_instance_type_db: 'n1-standard-2'
+gce_root_disk_type_db: 'pd-ssd'
+root_disk_size_db: 50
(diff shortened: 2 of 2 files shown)


fruch:
@roydahan @syuu1228 

https://github.com/scylladb/scylladb/issues/19325 happens also on rocky9, we back in square one.

fruch:
@yaronkaikov @syuu1228 

should we push this replacement, of wait for the fix to be confirmed ? (since centos9-selinux are failing on master and some releases)

syuu1228:
> > @yaronkaikov @syuu1228
> > should we push this replacement, of wait for the fix to be confirmed ? (since centos9-selinux are failing on master and some releases)
> 
> @yaronkaikov @syuu1228, we still need this replacement ? or cento9 is now fixed too ?

I think we always should test on rocky9 (or alma9) instead of centos9 stream, since centos9 stream is like "beta" distribution of RHEL9, may contain bugs which does not exist on RHEL9.
In fact we saw such bug on https://github.com/scylladb/scylladb/issues/19325, at this time the bug also released on RHEL9, though.

fruch:
> > > @yaronkaikov @syuu1228
> > > should we push this replacement, of wait for the fix to be confirmed ? (since centos9-selinux are failing on master and some releases)
> > 
> > @yaronkaikov @syuu1228, we still need this replacement ? or cento9 is now fixed too ?
> 
> I think we always should test on rocky9 (or alma9) instead of centos9 stream, since centos9 stream is like "beta" distribution of RHEL9, may contain bugs which does not exist on RHEL9.
> In fact we saw such bug on https://github.com/scylladb/scylladb/issues/19325, at this time the bug also released on RHEL9, though.

Sounds like testing on centos9 stream, did give us a heads-up on an issue that ended being on all of the downstreams.

The current situation is that both works ? if so maybe we should leave them both ?

roydahan:
If both works, I'm fine with leave them both for now.
We could replace centos9 to RHEL9 to be on the safe side of our support matrix, but it doesn't worth the investment at this point.

fruch:
> If both works, I'm fine with leave them both for now.
> We could replace centos9 to RHEL9 to be on the safe side of our support matrix, but it doesn't worth the investment at this point.
> 

centos9 started passing

https://argus.scylladb.com/test/070b0d31-e1ec-4173-abf5-39122f8c4011/runs?additionalRuns

And rocky9 passed on staging


As for support of RHEL9, it shouldn't be that hard to add it

fruch:
@roydahan @yaronkaikov 

changed this PR to introduce rocky9-selinux, and not remove centos9-selinux

yaronkaikov (approved):

roydahan (approved):

Issue #7907, opened 2024-07-09
Apply the default event filters again after upgrade to prevent ` Authentication error on host ` error


juliayakovlev:
## Packages

Base Scylla version: `2022.2.19-20240328.7cba40ad90d8` with build-id `ba3d72e54a63aea9441129edfd889bdc79e422a4`
Target Scylla version (or git commit hash): `2023.1.10-20240706.21cffccc1ccd` with build-id `87fcdaf894b5dfb25d26f12d7d4aed71b84bfd44`

Kernel Version: `6.5.0-1023-gcp`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Audit was disabled by https://github.com/scylladb/scylla-enterprise/issues/3148.
But it was not disabled in `2022.2`
So every time after upgrade from this version the test suffers from ` Authentication error on host /10.142.0.200:9042: Cannot achieve consistency level for cl ONE. Requires 1, alive 0` error.
We need to apply the default event filters again after upgrade

## Impact

*Describe the impact this issue causes to the user.*

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 3 nodes (n1-highmem-8)

Scylla Nodes used in this run:
- rolling-upgrade-sla--ubuntu-jammy-db-node-d318d953-0-3 (34.148.166.180 | 10.142.0.200) (shards: 7)
- rolling-upgrade-sla--ubuntu-jammy-db-node-d318d953-0-2 (34.23.79.253 | 10.142.0.196) (shards: 7)
- rolling-upgrade-sla--ubuntu-jammy-db-node-d318d953-0-1 (34.138.142.209 | 10.142.0.192) (shards: 7)

OS / Image: `https://www.googleapis.com/compute/v1/projects/ubuntu-os-cloud/global/images/family/ubuntu-2204-lts` (gce: undefined_region)

Test: `rolling-upgrade-with-sla-test`
Test id: `d318d953-d742-4d92-a8e6-4d29c59c5605`
Test name: `enterprise-2023.1/rolling-upgrade/rolling-upgrade-with-sla-test`
Test config file(s):

- [generic-rolling-upgrade.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/2d1cef1f9da0f3a751daa419b2d6c078676d446e/test-cases/upgrades/generic-rolling-upgrade.yaml)

<details>
<summary>
Logs and commands
[...]

PR #8607, opened 2024-09-09
fix(upgrade): enable default_filters after node upgrade

juliayakovlev:
Audit was disabled by scylladb/scylla-enterprise#3148. But it was not disabled in 2022.2
So every time after upgrade from this version the test suffers from 'Authentication error on host /10.142.0.200:9042: Cannot achieve consistency level for cl ONE. Requires 1, alive 0 error.'
This commit enables the default event filters again after node upgrade/rollback

Task: https://github.com/scylladb/scylla-cluster-tests/issues/7907

### Testing

- [ ] [rolling-upgrade-with-sla](https://argus.scylladb.com/test/c715037b-caed-479f-b550-944f46ea7576/runs?additionalRuns[]=b6aab9aa-c5cd-4ecd-89f8-b0f791587c95)

### Reminders

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -34,6 +34,7 @@ from sdcm import wait
 from sdcm.cluster import BaseNode
 from sdcm.fill_db_data import FillDatabaseData
 from sdcm.sct_events import Severity
+from sdcm.sct_events.setup import enable_default_filters
 from sdcm.stress_thread import CassandraStressThread
 from sdcm.utils.decorators import retrying
 from sdcm.utils.user_profile import get_profile_content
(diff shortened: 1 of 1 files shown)


upgrade_test.py
```diff
@@ -362,6 +363,7 @@ def _upgrade_node(self, node, upgrade_sstables=True, new_scylla_repo=None, new_v
         InfoEvent(message='upgrade_node - starting to "_update_argus_upgraded_version"').publish()
         self._update_argus_upgraded_version(node, new_ver)
         InfoEvent(message='upgrade_node - ended to "_update_argus_upgraded_version"').publish()
+        enable_default_filters(sct_config=self.params)
```
fruch, line 366: just calling it won't cut it,

you need to be able to call it with multiple version ? i.e. once for base, and once for target version

those are not update/changed in SctConfiguration.

it's need to go down to the SkipPerIssue implementation to be aware also what is target version

juliayakovlev:
skip tag is added to the issue. It should resolve the problem

PR #8608, opened 2024-09-09
fix(is_enterprise): don't cache if no scylla is installed

fruch:
cause of recent improvement to speeding up test setup, now we can reach to situation monitor stack might spin before first db node is installed (i.e. when not using scylla images).

that in turn might cache the outcome of `is_enterprise` with a None value since no version is yet installed which later would cause the installation to try install scylla oss version.

this change is moving to a logic that check for the existence of both OSS and enterprise version, and if both doesn't exists it don't cache any value, so when eventually we do have a version available it would return the correct answer.

### Testing

- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-fips-test/5
- [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-fips-test/8/

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -782,21 +782,29 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
             startup_interface_command = "ip link set {} up"
         self.remoter.sudo(startup_interface_command.format(interface_name))
 
-    @cached_property
+    @optional_cached_property
     def is_enterprise(self) -> bool | None:
         _is_enterprise = None
+
         if self.distro.is_rhel_like:
-            result = self.remoter.sudo("yum search scylla-enterprise 2>&1", ignore_status=True)
-            if result.ok:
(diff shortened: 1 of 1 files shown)


soyacz:
I tried yesterday to investigate this issue too, but couldn't find the culprit.
Still I don't understand where MonitorNode in SCT code verifies `is_enterprise` on db node.
Maybe it should use `params.is_enterprise`?

fruch:
> I tried yesterday to investigate this issue too, but couldn't find the culprit. Still I don't understand where MonitorNode in SCT code verifies `is_enterprise` on db node. Maybe it should use `params.is_enterprise`?

is long long process that comes from using `node.scylla_version` that calls `node.scylla_pkg` that calls `is_enterprise`

later today I'll spin the offending job with this fix to confirm.

caching can easily backfire, if we aren't careful enough

fruch:
> LGTM Does it fix the issue with scylla installation?

yes, the case now runs correctly

vponomaryov (approved):
LGTM

PR #8631, opened 2024-09-11
fix(aws-provision): have a default for `user_data_format_version`

fruch:
since we don't set a default for `user_data_format_version` in case we are using images which are not scylla for the db nodes

we'll be hitting the following error in provision step:

```
pydantic.error_wrappers.ValidationError: 1 validation error for ScyllaUserDataBuilder
user_data_format_version
  none is not an allowed value (type=type_error.none.not_allowed)
```

in all other backend we have default for this value, and in AWS we were missing it, it's only affecting once case we have that is running longevity without scylla images.

### Testing

- [x]  🔴 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-fips-test/6/

### Reminders

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -2471,6 +2471,7 @@ class SCTConfiguration(dict):
                                             ('oracle_user_data_format_version', ami_id_db_oracle)]:
                 if ami_list:
                     user_data_format_versions = set()
+                    self[key_to_update] = '2'
                     for ami_id, region_name in zip(ami_list, region_names):
                         if not ami_built_by_scylla(ami_id, region_name):
                             continue


sdcm/sct_config.py
```diff
@@ -2471,6 +2471,7 @@ def verify_configuration_urls_validity(self):  # pylint: disable=too-many-branch
                                             ('oracle_user_data_format_version', ami_id_db_oracle)]:
                 if ami_list:
                     user_data_format_versions = set()
+                    self[key_to_update] = '2'
```
soyacz, line 2474: why 2?
We set it in AWS here to 3: https://github.com/scylladb/scylla-cluster-tests/blob/12fd46fa03dd3cfc786779ca5957d68a31e476c7/sdcm/cluster.py#L3390

fruch: It makes no difference if it's not SMI image

fruch:
provision stage worked with this change (test in link failed later stage, we'll need to run again, when other fixes for that case would be merged)

soyacz (approved):
LGTM

PR #8632, opened 2024-09-11
improvement(decorator): adapt latency_calculator_decorator

juliayakovlev:
If the decorator is applied dynamically, 'self' argument is not transferred  via 'args' and may be found in bounded function.
This commit allow to use the 'self' parameter of different instance types: ClusterTester, Nemesis. ClusterTester and Nemesis classes have a different naming convention for cluster, monitor and tester params. As result using of latency_calculator_decorator requires adaptation of those names to name them same.

### Testing

- [x] [perf-regression-latency-650gb-during-rolling-upgrade](https://argus.scylladb.com/test/8d048290-f7f4-4af5-9e09-882202a35656/runs?additionalRuns[]=fc6cabb3-a93a-4d83-9a8e-20471e8e4583)
- [x] [longevity-100gb-4h](https://argus.scylladb.com/test/657bf1f3-69a5-4c36-bdbf-bf17718f7087/runs?additionalRuns[]=636dbf58-fc51-426a-a81c-85cfbc89c42a)

### Reminders

### DIFF
--- a/sdcm/utils/decorators.py
+++ b/sdcm/utils/decorators.py
@@ -170,23 +170,40 @@ def latency_calculator_decorator(original_function: Optional[Callable] = None, *
     def wrapper(func):
 
         @wraps(func)
-        def wrapped(*args, **kwargs):  # pylint: disable=too-many-branches, too-many-locals
+        def wrapped(*args, **kwargs):  # noqa: PLR0914
+            from sdcm.tester import ClusterTester
+            from sdcm.nemesis import Nemesis
             start = time.time()
-            start_node_list = args[0].cluster.nodes[:]
-            reactor_stall_stats = {}
+            # If the decorator is applied dynamically, "self" argument is not transferred  via "args" and may be found in bounded function
+            _self = getattr(func, "__self__", None) or args[0]
+            if isinstance(_self, ClusterTester):
+                cluster = _self.db_cluster
+                tester = _self
+                monitoring_set = _self.monitors
+            elif isinstance(_self, Nemesis):
+                cluster = _self.cluster
+                tester = _self.tester
+                monitoring_set = _self.monitoring_set
+            else:
+                raise ValueError(
+                    f"Not expected instance type '{type(_self)}'. Supported types: 'ClusterTester', 'Nemesis'")
+
+            # Keep for debug purposes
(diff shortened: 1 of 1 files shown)


fruch (dismissed):
LGTM

sdcm/utils/decorators.py
```diff
@@ -170,23 +170,40 @@ def latency_calculator_decorator(original_function: Optional[Callable] = None, *
...
+            elif isinstance(_self, Nemesis):
+                cluster = _self.cluster
+                tester = _self.tester
+                monitoring_set = _self.monitoring_set
+            else:
+                raise ValueError(
+                    f"Not expected instance type '{type(_self)}'. Supported types: 'ClusterTester', 'Nemesis'")
+
+            # Keep for debug purposes
+            LOGGER.info("latency_calculator_decorator cluster: %s", cluster)
```
soyacz, line 192: I'm not sure if we need that, at least setting level to DEBUG.

soyacz (approved):
besides one small thing related debug log line, LGTM

sdcm/utils/decorators.py
```diff
@@ -170,23 +170,40 @@ def latency_calculator_decorator(original_function: Optional[Callable] = None, *
...
-        def wrapped(*args, **kwargs):  # pylint: disable=too-many-branches, too-many-locals
+        def wrapped(*args, **kwargs):  # noqa: PLR0914
+            from sdcm.tester import ClusterTester
+            from sdcm.nemesis import Nemesis
             start = time.time()
-            start_node_list = args[0].cluster.nodes[:]
-            reactor_stall_stats = {}
+            # If the decorator is applied dynamically, "self" argument is not transferred  via "args" and may be found in bounded function
+            _self = getattr(func, "__self__", None) or args[0]
+            if isinstance(_self, ClusterTester):
```
vponomaryov, line 179: We can check the `_self.__class__` value instead of making unnecessary imports:

```
_self = getattr(func, "__self__", None) or args[0]

if "ClusterTester" in _self.__class__:
    ...
elif "Nemesis" in _self.__class__:
    ...
else:
 raise ValueError(...)
```

fruch: I rather not, that won't work for inheritage

PR #8640, opened 2024-09-11
fix(sct.py): Ignore TEST_ERROR inside finish_argus_test_run

### LINKED ISSUE #8614: Argus is not setting `TEST-ERROR` for some tests
For some reason [this run](https://argus.scylladb.com/test/aa05fe67-89bd-4e30-88db-5e2b2e3d986d/runs?additionalRuns[]=b1611080-4440-472a-9c5a-9e61b1728d80) and [this run](https://argus.scylladb.com/test/f3e70652-634d-43ae-ba44-1ee974306fff/runs?additionalRuns%5B%5D=dd3d1632-6bc3-4140-9c09-4b86a29554dd)
Did not handle this event as `TEST-ERROR`
```
2024-09-06 00:01:56.534: (SpotTerminationEvent Severity.CRITICAL) period_type=one-time event_id=289e2787-9d81-4d8b-b098-f8258e954662: node=Node longevity-10gb-3h-master-db-node-b1611080-eastus-2 [None | 10.0.0.6] message=Got spot termination event for node: ['longevity-10gb-3h-master-db-node-eastus-2']. VM eviction time is Fri, 06 Sep 2024 00:02:05 GMT.
```

My primary suspicion is that SCT runner closed before it got to event collection and the new bit we added that uploads the events during log collection should probably trigger the same logic.

Another thing is that they're both run on Azure, and maybe we handle something differently here, althout the regex for plain Spot Termination Event is very simple and I don't see yet what could be stopping it from being matched.

k0machi:
This commit corrects an issue where a test that would previously finish
with test error inside sct stage would get overwritten by jenkins
setting status to failed due to pipeline health

Fixes #8614

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -1783,7 +1783,7 @@ def finish_argus_test_run(jenkins_status):
         test_config.set_test_id_only(params.get('test_id'))
         test_config.init_argus_client(params)
         status = test_config.argus_client().get_status()
-        if status in [TestStatus.PASSED, TestStatus.FAILED]:
+        if status in [TestStatus.PASSED, TestStatus.FAILED, TestStatus.TEST_ERROR]:
             LOGGER.info("Argus TestRun already finished with status %s", status.value)
             return
         new_status = TestStatus.FAILED


vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

Issue #8639, opened 2024-09-11
hydra failed with ModuleNotFoundError: No module named '_pytest._py'


asias:
### Issue description

- [ ] This issue is a regression.
- [x ] It is unknown if this issue is a regression.

#### Steps to Reproduce

```
On fedora 40.

[asias@hjpc2 scylla-cluster-tests]$ git log -1
commit 9950447888bb4d1c09864e5e08f591a7459b6a27 (HEAD -> m20240911, origin/master, origin/HEAD)
Author: Israel Fruchter <[fruch@scylladb.com](mailto:fruch@scylladb.com)>
Date:   Mon Sep 9 09:44:59 2024 +0300

    fix(fips): switch fips check to be based on sct configuration
   
    since we have cases we'll want to run ubuntu pro (EOL versions as
    example), that won't have fips enable, the assumption that ubuntu pro
    mean only the fips images, isn't valid anymore.
   
    in this change we introduce a new configuration option, that
    would help us list the distro features we expect, so the test
    would be able to verify those on installation.

[asias@hjpc2 scylla-cluster-tests]$ sudo sh ./install-hydra.sh
Docker version 26.1.1, build 4cf5afa
Docker is installed.
'/usr/local/bin/hydra' -> '/home/asias/src/cloudius-systems/scylla-cluster-tests/docker/env/hydra.sh'
Hydra installed.
aws-cli/2.17.6 Python/3.12.5 Linux/6.10.3-200.fc40.x86_64 source/x86_64.fedora.40
AWS CLI installed.
==================================     NOTES      =================================================
To check that Hydra is installed, run 'hydra bash' anywhere in bash.
It will create a SCT Docker container and connect to its bash.
When running Hydra for the first time it will build the SCT Docker image. Please be patient!
===================================================================================================
Please run 'aws configure' to configure AWS CLI
[asias@hjpc2 scylla-cluster-tests]$ hydra investigate show-logs c44328c3-8dfd-4e00-a8c9-e8bc6c57c52a
There is scylladb/hydra:v1.75-event-collect-argus-0.12.7 in local cache, using it.
Going to run './sct.py  investigate show-logs c44328c3-8dfd-4e00-a8c9-e8bc6c57c52a'...
Traceback (most recent call last):
  File "/home/asias/src/cloudius-systems/scylla-cluster-tests/./sct.py", line 34, in <module>
    import pytest
[...]

PR #8675, opened 2024-09-12
fix(hydra): overmount "%home%/.local" dir

fruch:
It will allow us to stop using locally installed python packages
which easily may become incompatible with the SCT code.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/8639

### Testing

- [x] provision tests

### Reminders

### DIFF
--- a/docker/env/hydra.sh
+++ b/docker/env/hydra.sh
@@ -248,6 +248,7 @@ function run_in_docker () {
         -v /tmp:/tmp \
         -v /var/tmp:/var/tmp \
         -v "${HOME_DIR}:${HOME_DIR}" \
+        --tmpfs "${HOME_DIR}/.local:exec,uid=$(id -u ${USER}),gid=$(id -g ${USER})" \
         -w "${SCT_DIR}" \
         -e JOB_NAME="${JOB_NAME}" \
         -e BUILD_URL="${BUILD_URL}" \


docker/env/hydra.sh
```diff
@@ -247,7 +247,9 @@ function run_in_docker () {
         -v "${SCT_DIR}:${SCT_DIR}" \
         -v /tmp:/tmp \
         -v /var/tmp:/var/tmp \
-        -v "${HOME_DIR}:${HOME_DIR}" \
+        -v "${HOME_DIR}/.aws:${HOME_DIR}/.aws" \
+        -v "${HOME_DIR}/.ssh:${HOME_DIR}/.ssh" \
+        -v "${HOME_DIR}/sct-results:${HOME_DIR}/sct-results" \
```
vponomaryov, line 252: The precommit failed with the following error:

```
Error saving credentials: mkdir /home/jenkins/.docker: permission denied
```

So, looks like the list of sub-directories mounts is not complete.
At least `${HOME_DIR}/.docker` is missing

vponomaryov (changes requested):

docker/env/hydra.sh
vponomaryov: For K8S may be needed `.helm` and `.kube`...

fruch: yeah, it would take some time to scope them all out...

vponomaryov (approved):
LGTM

docker/env/hydra.sh
vponomaryov: Oh, it also wants `.cache` dir

vponomaryov (changes requested):
Need to add the `.cache` dir too

vponomaryov:
@fruch do you plan to update it?

fruch:
> @fruch do you plan to update it?

I didn't get to it this week, want to take over ?

sdcm/utils/aws_okta.py
```diff
@@ -39,7 +39,7 @@ def can_get_to_aws_account():
         response = sts.get_caller_identity()
         assert response['Account'] == account_id
         LOGGER.info("logged in as %s", response['Arn'])
-    except (NoCredentialsError, AssertionError):
+    except (NoCredentialsError, AssertionError, EndpointConnectionError):
```
fruch, line 42: This is a mistake, it is not intended to be part of this change

vponomaryov:
@fruch left to squash commits into one and rename the only commit to something like following:

```
fix(hydra): overmount "%home%/.local" dir

It will allow us to stop using locally installed python packages
which easily may become incompatible with the SCT code.
```

vponomaryov:
@fruch all failures are unrelated.
`precommit` failed due to not satisfied commit topic structure of the temporary commits. Will be ok after squashing of commits.
`integration` stage failed on 2 tests without any relation to this change. Those 2 tests fail constantly on master and require fixing.

docker/env/hydra.sh
```diff
@@ -248,6 +248,7 @@ function run_in_docker () {
         -v /tmp:/tmp \
         -v /var/tmp:/var/tmp \
         -v "${HOME_DIR}:${HOME_DIR}" \
+        -v "/tmp/.local-${SCT_TEST_ID}:${HOME_DIR}/.local" \
```
fruch, line 251: why no use tmpfs ?

vponomaryov: You can pick up any other dir if you want.

docker/env/hydra.sh
```diff
@@ -248,6 +248,7 @@ function run_in_docker () {
         -v /tmp:/tmp \
         -v /var/tmp:/var/tmp \
         -v "${HOME_DIR}:${HOME_DIR}" \
+        --tmpfs "${HOME_DIR}/.local:exec,uid=$(id -u ${USER}),gid=$(id -g ${USER})" \
```
vponomaryov, line 251: This code also gets run on builders.
This is the reason I added TEST_ID to the source dir name.

vponomaryov: Also, running locally 2 different hydra commands for different test runs now will overlap.

fruch: what is gonna overlap with this, --tmpfs mean it would be generate on every docker run command inside the docker.

it won't be depended on anything external to that specific run.

vponomaryov: It can be simplified by using the existing var:

```
USER_ID=$(id -u "${USER}"):$(id -g "${USER}")
```

fruch: I thought of that, but it's not the exact same syntax
one is `1000:1000` and one `uid=1000,gid=1000`

so it's better left as is.

vponomaryov (approved):
LGTM

PR #8680, opened 2024-09-12
improvement(hdr stats): collect throughput and report throughput in Result tab

juliayakovlev:
Collect throughput statistics from HDR file and report in Elasticsearch.
Now only latency is reported in the Report tab for performance tests.
This commit add throughput as well

### Testing

- [x] [-perf-regression-latency-650gb-during-rolling-upgrade](https://argus.scylladb.com/test/8d048290-f7f4-4af5-9e09-882202a35656/runs?additionalRuns[]=c1930454-806a-412c-abb3-7a3f9b9d0e3c)

### Reminders

### DIFF
--- a/sdcm/argus_results.py
+++ b/sdcm/argus_results.py
@@ -39,9 +39,11 @@ class LatencyCalculatorMixedResult(GenericResultTable):
             ColumnMetadata(name="P90 read", unit="ms", type=ResultType.FLOAT),
             ColumnMetadata(name="P99 write", unit="ms", type=ResultType.FLOAT),
             ColumnMetadata(name="P99 read", unit="ms", type=ResultType.FLOAT),
+            ColumnMetadata(name="Throughput write", unit="op/s", type=ResultType.INTEGER),
+            ColumnMetadata(name="Throughput read", unit="op/s", type=ResultType.INTEGER),
             ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION),
             ColumnMetadata(name="Overview", unit="", type=ResultType.TEXT),
-            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT),
+            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT)
         ]
 
 
@@ -52,9 +54,10 @@ class LatencyCalculatorWriteResult(GenericResultTable):
         Columns = [
             ColumnMetadata(name="P90 write", unit="ms", type=ResultType.FLOAT),
             ColumnMetadata(name="P99 write", unit="ms", type=ResultType.FLOAT),
+            ColumnMetadata(name="Throughput write", unit="op/s", type=ResultType.INTEGER),
             ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION),
             ColumnMetadata(name="Overview", unit="", type=ResultType.TEXT),
-            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT),
+            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT)
         ]
 
 
@@ -65,9 +68,10 @@ class LatencyCalculatorReadResult(GenericResultTable):
         Columns = [
             ColumnMetadata(name="P90 read", unit="ms", type=ResultType.FLOAT),
             ColumnMetadata(name="P99 read", unit="ms", type=ResultType.FLOAT),
+            ColumnMetadata(name="Throughput read", unit="op/s", type=ResultType.INTEGER),
             ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION),
             ColumnMetadata(name="Overview", unit="", type=ResultType.TEXT),
-            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT),
+            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT)
         ]
 
 
@@ -106,6 +110,13 @@ def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, de
                                     row=f"Cycle #{cycle}",
                                     value=value,
                                     status=Status.PASS if value < operation_error_thresholds[f"percentile_{percentile}"] else Status.ERROR)
+        if value := summary[operation.upper()].get("throughput", None):
+            # TODO: This column will be validated in the gradual test. `PASS` is temporary status. Should be handled later
+            result_table.add_result(column=f"Throughput {operation.lower()}",
+                                    row=f"Cycle #{cycle}",
+                                    value=value,
+                                    status=Status.UNSET)
+
     result_table.add_result(column="duration", row=f"Cycle #{cycle}",
                             value=result["duration_in_sec"], status=Status.PASS)
     try:
--- a/sdcm/utils/decorators.py
+++ b/sdcm/utils/decorators.py
@@ -237,6 +237,10 @@ def latency_calculator_decorator(original_function: Optional[Callable] = None, *
             result["hdr_summary"] = tester.get_cs_range_histogram(stress_operation=workload,
                                                                   start_time=start,
                                                                   end_time=end)
+            hdr_throughput = 0
+            for summary, values in result["hdr_summary"].items():
+                hdr_throughput += values["throughput"]
+            result["cycle_hdr_throughput"] = round(hdr_throughput)
             result["reactor_stalls_stats"] = reactor_stall_stats
 
             if "steady" in func.__name__.lower():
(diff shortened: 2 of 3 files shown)


sdcm/argus_results.py
```diff
@@ -106,6 +113,12 @@ def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, de
                                     row=f"Cycle #{cycle}",
                                     value=value,
                                     status=Status.PASS if value < operation_error_thresholds[f"percentile_{percentile}"] else Status.ERROR)
+        if value := summary[operation.upper()].get("throughput", None):
+            result_table.add_result(column=f"Throughput {operation}",
```
soyacz, line 117: safer would be `{operation.lower()}`, just in case (would raise error and fail sending result).

sdcm/utils/decorators.py
```diff
@@ -237,6 +237,10 @@ def wrapped(*args, **kwargs):  # noqa: PLR0914
             result["hdr_summary"] = tester.get_cs_range_histogram(stress_operation=workload,
                                                                   start_time=start,
                                                                   end_time=end)
+            hdr_throughput = 0
+            for summary, values in result["hdr_summary"].items():
+                hdr_throughput += values["throughput"]
+            result["cycle_hdr_throughput"] = round(hdr_throughput)
```
soyacz, line 243: where is it used?

sdcm/argus_results.py
```diff
@@ -106,6 +113,12 @@ def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, de
...
+                                    row=f"Cycle #{cycle}",
+                                    value=value,
+                                    status=Status.PASS)
```
soyacz, line 120: better set `UNSET` if we don't validate it.
(might need to rebase this PR as this status was added just recently)

soyacz (commented):
just a few remarks, otherwise LGTM.

Worth to unit test it in `unit_tests/test_argus_results.py` (by adding proper prop to `unit_tests/test_data/latency_decorator_result.json`)

sdcm/argus_results.py
```diff
@@ -42,6 +42,9 @@ class Meta:
             ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION),
             ColumnMetadata(name="Overview", unit="", type=ResultType.TEXT),
             ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT),
+            ColumnMetadata(name="Throughput write", unit="ops", type=ResultType.INTEGER),
+            ColumnMetadata(name="Throughput read", unit="ops", type=ResultType.INTEGER),
+            ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION)
```
soyacz, line 47: Better put it before screenshot columns - will look better in Argus

soyacz:
Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)

fruch:
> Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)

we can call it `TPS` ? which is quite common shortcut for it,

sdcm/argus_results.py
```diff
@@ -42,6 +42,9 @@ class Meta:
             ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION),
             ColumnMetadata(name="Overview", unit="", type=ResultType.TEXT),
             ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT),
+            ColumnMetadata(name="Throughput write", unit="ops", type=ResultType.INTEGER),
```
fruch, line 45: maybe we should make the unit="op/s" ?

sdcm/utils/decorators.py
fruch: `hdr_summary` isn't sent to argus at all, so this might be used in email template.
but yes in this PR it's not used, but would be saved into ES

sdcm/argus_results.py
juliayakovlev: done

sdcm/utils/decorators.py
juliayakovlev: > `hdr_summary` isn't sent to argus at all, so this might be used in email template. but yes in this PR it's not used, but would be saved into ES

this is exactly the reason. I'll use it in the gradual test email

sdcm/argus_results.py
juliayakovlev: I plan to validate it in gradual test. `PASS` is temporary status. Added `TODO`

fruch: so UNSET should be the temporary.

@soyacz is working on a better way to express the logic for validation we'll want to do for each metric, and we'll align to that when it would be ready

sdcm/argus_results.py
juliayakovlev: done

juliayakovlev:
> > Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)
> 
> we can call it `TPS` ? which is quite common shortcut for it,

Is `Thrpt` or `TPS` clear enough?

sdcm/argus_results.py
juliayakovlev: I think it is better to report throughput with latency, not separate it.

juliayakovlev: I think the throughput stats should be placed with latency

sdcm/argus_results.py
juliayakovlev: changed

juliayakovlev:
@fruch @soyacz if the fixes are ok and you do not have a new comments, lets merge it. I need this PR in gradual test. Thanks

soyacz:
> > Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)
> 
> we can call it `TPS` ? which is quite common shortcut for it,

TPS refers to Transactions Per Second, I'm not sure if we can call it. Maybe just 'writes` and `reads` with `[op/s]` unit would be better then?

juliayakovlev:
> > > Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)
> > 
> > 
> > we can call it `TPS` ? which is quite common shortcut for it,
> 
> TPS refers to Transactions Per Second, I'm not sure if we can call it. Maybe just 'writes`and`reads`with`[op/s]` unit would be better then?

@fruch @soyacz What is the problem with "Throughput"? I think it is clear for everybody

fruch:
> > > > Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)
> > > 
> > > 
> > > we can call it `TPS` ? which is quite common shortcut for it,
> > 
> > TPS refers to Transactions Per Second, I'm not sure if we can call it. Maybe just 'writes`and`reads`with`[op/s]` unit would be better then?
> 
> @fruch @soyacz What is the problem with "Throughput"? I think it is clear for everybody

@soyacz concern was that `read Throughput` is lengthy and makes the table less readable, as we add more data into it.

I think ops are identical to transactions in our context, and for choosing arbitration, it's better than one that is less known.

juliayakovlev:
> > > > > Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)
> > > > 
> > > > 
> > > > we can call it `TPS` ? which is quite common shortcut for it,
> > > 
> > > 
> > > TPS refers to Transactions Per Second, I'm not sure if we can call it. Maybe just 'writes`and`reads`with`[op/s]` unit would be better then?
> > 
> > 
> > @fruch @soyacz What is the problem with "Throughput"? I think it is clear for everybody
> 
> @soyacz concern was that `read Throughput` is lengthy and makes the table less readable, as we add more data into it.
> 
> I think ops are identical to transactions in our context, and for choosing arbitration, it's better than one that is less known.

Screenshot from 2024-09-17 09-13-52

It looks so.
Please, let me know how to change it. I need this PR merged

fruch:
> > > > > > Maybe also we should shorten the word `Throughput` to e.g. `Thrpt` for more concise table (currently takes full width)
> > > > > 
> > > > > 
> > > > > we can call it `TPS` ? which is quite common shortcut for it,
> > > > 
> > > > 
> > > > TPS refers to Transactions Per Second, I'm not sure if we can call it. Maybe just 'writes`and`reads`with`[op/s]` unit would be better then?
> > > 
> > > 
> > > @fruch @soyacz What is the problem with "Throughput"? I think it is clear for everybody
> > 
> > 
> > @soyacz concern was that `read Throughput` is lengthy and makes the table less readable, as we add more data into it.
> > I think ops are identical to transactions in our context, and for choosing arbitration, it's better than one that is less known.
> 
> Screenshot from 2024-09-17 09-13-52
> 
> It looks so. Please, let me know how to change it. I need this PR merged

to me it's readable enough like that, Argus can be fine-tune it's css to minimize it a bit further if needed to add more data

fruch (approved):
LGTM

soyacz (approved):

Issue #8721, opened 2024-09-16
Why do we relay on `sh` when run docker 


dkropachev:
We want to introduced custom `scylla-bench`, which is based on `scratch`, so there is no shell.
As result [run](https://jenkins.scylladb.com/job/scylla-staging/job/sylwia.szunejko/job/longevity-large-partition-asymmetric-cluster-3h-test/16) is failing with:
```
[2024-09-16T16:02:20.635Z] < t:2024-09-16 16:02:19,262 f:base.py         l:147  c:RemoteLibSSH2CmdRunner p:ERROR > <10.4.1.125>: Error executing command: "sudo  docker exec 4d7237fa463dccbb62a9942ba0e98ad68b25e897cb33a00181cdc1a32210a73f /bin/sh -c 'scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=100 -clustering-row-count=5555                       -clustering-row-size=uniform:1024..2048 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=30s -validate-data -error-at-row-limit 1000 -nodes 10.4.2.113,10.4.3.237,10.4.3.250,10.4.3.53,10.4.3.32'"; Exit status: 1

[2024-09-16T16:02:20.635Z] < t:2024-09-16 16:02:19,264 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-09-16 16:02:19.262: (ScyllaBenchEvent Severity.CRITICAL) period_type=end event_id=70b93f59-f12e-4a6d-a0ef-8314abb8ba81 duration=0s: node=Node longevity-large-partitions-3h-maste-loader-node-a7dd38ed-1 [3.252.137.240 | 10.4.1.125]

[2024-09-16T16:02:20.635Z] < t:2024-09-16 16:02:19,264 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > stress_cmd=scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=100 -clustering-row-count=5555                       -clustering-row-size=uniform:1024..2048 -concurrency=10 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=30s -validate-data -error-at-row-limit 1000 -nodes 10.4.2.113,10.4.3.237,10.4.3.250,10.4.3.53,10.4.3.32

[...]

PR #8723, opened 2024-09-17
scylla_bench_thread: make it support docker image with no shell

dkropachev:
Closes https://github.com/scylladb/scylla-cluster-tests/issues/8721

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/scylla_bench_thread.py
+++ b/sdcm/scylla_bench_thread.py
@@ -20,6 +20,7 @@ import logging
 import contextlib
 from enum import Enum
 
+
 from sdcm.loader import ScyllaBenchStressExporter
 from sdcm.prometheus import nemesis_metrics_obj
 from sdcm.provision.helpers.certificate import SCYLLA_SSL_CONF_DIR, TLSAssets
@@ -158,7 +159,7 @@ class ScyllaBenchThread(DockerBasedStressThread):  # pylint: disable=too-many-in
 
         return sb_summary, errors
 
-    def create_stress_cmd(self, stress_cmd, loader, cmd_runner):
+    def create_stress_cmd(self, stress_cmd, loader, cmd_runner, is_shell_supported):
         if self.connection_bundle_file:
             stress_cmd = f'{stress_cmd.strip()} -cloud-config-path={self.target_connection_bundle_file}'
         else:
@@ -183,7 +184,9 @@ class ScyllaBenchThread(DockerBasedStressThread):  # pylint: disable=too-many-in
                     # server_names = ' '.join(f'-tls-server-name {ip}' for ip in ips.split(","))
                     # stress_cmd = f'{stress_cmd.strip()} -tls-host-verification {server_names}'
 
-        return stress_cmd
+        if is_shell_supported:
+            return stress_cmd
+        return stress_cmd.removeprefix('scylla-bench')
 
     def _run_stress(self, loader, loader_idx, cpu_idx):  # pylint: disable=too-many-locals
         cmd_runner = None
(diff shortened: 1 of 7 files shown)


dkropachev:
@vponomaryov, can you please take a look if it is good enought

sdcm/scylla_bench_thread.py
```diff
@@ -203,6 +207,8 @@ def _run_stress(self, loader, loader_idx, cpu_idx):  # pylint: disable=too-many-
             )
             cmd_runner_name = loader.ip_address
 
+        is_shell_supported = cmd_runner.run(cmd="sh -c true", timeout=self.timeout, retry=0).returncode == 0
```
vponomaryov, line 210: Not being 0 it will raise exception.
Need to use `ignore_status=True` in such a case.

sdcm/scylla_bench_thread.py
```diff
@@ -183,7 +185,9 @@ def create_stress_cmd(self, stress_cmd, loader, cmd_runner):
                     # server_names = ' '.join(f'-tls-server-name {ip}' for ip in ips.split(","))
                     # stress_cmd = f'{stress_cmd.strip()} -tls-host-verification {server_names}'
 
-        return stress_cmd
+        if is_shell_supported:
+            return stress_cmd
+        return stress_cmd.removeprefix('scylla-bench')
```
vponomaryov, line 190: There is logic that depends on the tool/binary name that is going to be used.
For example, dynamic loader in K8S uses it.
Maybe somewhere else.

sdcm/utils/docker_remote.py
```diff
@@ -102,7 +106,9 @@ def get_log(self):
         return self.node.remoter.run(f"{self.sudo_needed} docker logs {self.docker_id}").stdout.strip()
 
     def run(self, cmd, *args, **kwargs):
-        return self.node.remoter.run(f'{self.sudo_needed} docker exec {self.docker_id} /bin/sh -c {shlex.quote(cmd)}', *args, **kwargs)
+        if self.is_shell_supported:
+            return self.node.remoter.run(f'{self.sudo_needed} docker exec {self.docker_id} /bin/sh -c {shlex.quote(cmd)}', *args, **kwargs)
+        return self.node.remoter.run(f'{self.sudo_needed} docker exec {self.docker_id} {cmd}', *args, **kwargs)
```
vponomaryov, line 111: I think that something in SCT may fail without `shlex.quote` usage in this case.

vponomaryov (commented):
Comments inline. But in general I am not sure it is the way we should follow.

This implicit approach looks fragile and we have a lot of different places where we use either shell env vars or pipes...
So, personally, I would prefer to have shell in each loader image.

@fruch and @soyacz WDYT?

sdcm/scylla_bench_thread.py
```diff
@@ -183,7 +184,9 @@ def create_stress_cmd(self, stress_cmd, loader, cmd_runner):
...
```
fruch, line 189: this change is not related to have sh or not.
it's depend on having an entrypoint set to what.

I rather align this to set explicitly the entry point SCT expect, then modifying the command, based on unrelated logic

sdcm/scylla_bench_thread.py
dkropachev: > Not being 0 it will raise exception. Need to use `ignore_status=True` in such a case.

fixed.

dkropachev: > this is not an assert statement...

?

sdcm/scylla_bench_thread.py
dkropachev: Fixed that part of logic to use env instead

sdcm/utils/docker_remote.py
dkropachev: fixed

fruch:
@dkropachev do we still need this one ?

dkropachev:
this PR is getting too intrusive, closing it, will make a special dockerfile with a shell in it.

PR #8731, opened 2024-09-17
improvement(test-stages): add ability to skip test stages

dimakr:
The change introduces a feature that allows specific stages of a test to be skipped. This would allow to optionally:
- start from a specific test stage (e.g. start from the main load while skipping the prepare load)
- skip intermediary test stages (e.g. bypass nemesis disruption stage)

The new functionality is intended to complement the reuse_cluster capability, allowing re-running subset of test stages of initial test scenario.
However, it can also be used standalone when certain stages are non-essential for a particular scenario.

Closes: https://github.com/scylladb/qa-tasks/issues/1761

### Testing
- [ ] :clock1: [PR-provision-test with skipped prepare cmd](url)

### Reminders


dimakr:
@fruch draft for the https://github.com/scylladb/qa-tasks/issues/1761 task implementation.
Do you think it is something we can continue with? I tried some other options, but they are messy. This one is readable, makes the least amount of changes in the test itself.

sdcm/utils/decorators.py
```diff
@@ -321,3 +321,15 @@ def wrapper(*args, **kwargs):
                 raise UnsupportedNemesis("Capacity Issue") from ex
             raise
     return wrapper
+
+
+def optional_stage(stage_name):
```
fruch, line 326: A docstring explaining the usage and  pointing to docs explain this one, could be useful

fruch:
> @fruch draft for the https://github.com/scylladb/qa-tasks/issues/1761 task implementation.
> Do you think it is something we can continue with? I tried some other options, but they are messy. This one is readable, makes the least amount of changes in the test itself.

The idea looks quite solid and not too intrusive, we need a bit of examples on how to use it.

We should check how we can override with environment variables, so we would be able to use this also in Jenkins runs

One idea that I've seen is Julia and Roy using a lot in the gradual perf test, running a specific round of a loop, if we can use a similar approach idea for running a specific iteration from a loop. but that can be an extension to this one.

soyacz:
I'd propose making it automatic:
create a decorator that will:
1.  write to a file info about given function that is completed (e.g line in e.g.  /tmp/completed-steps-<test_id> or somewhere in results dir)
2. If decorator sees that line and is reusing cluster, then skip decorated method.

Only decorate methods that are intended for skipping when reusing (remember about Perf tests methods too).

dimakr:
> I'd propose making it automatic: create a decorator that will:
> 
> 1. write to a file info about given function that is completed (e.g line in e.g.  /tmp/completed-steps-<test_id> or somewhere in results dir)
> 2. If decorator sees that line and is reusing cluster, then skip decorated method.
> 
> Only decorate methods that are intended for skipping when reusing (remember about Perf tests methods too).

I think implementing the functionality this way would restrict us from some use cases:
- using the feature standalone, not only with reuse_cluster. Probably it won't be a common a scenario, but if a user would need to skip some test stages of a test when running it for th 1st time, the feature would help with it.
- when reusing a cluster that was created by someone (something) else. E.g. when someone creates a reproducer, provisions and keeps a cluster and passes it to a dev who needs it for their work (update packages, rerun only subset of test stages, etc.)

What do you think?

soyacz:
> > I'd propose making it automatic: create a decorator that will:
> > 
> > 1. write to a file info about given function that is completed (e.g line in e.g.  /tmp/completed-steps-<test_id> or somewhere in results dir)
> > 2. If decorator sees that line and is reusing cluster, then skip decorated method.
> > 
> > Only decorate methods that are intended for skipping when reusing (remember about Perf tests methods too).
> 
> I think implementing the functionality this way would restrict us from some use cases:
> 
> * using the feature standalone, not only with reuse_cluster. Probably it won't be a common a scenario, but if a user would need to skip some test stages of a test when running it for th 1st time, the feature would help with it.

Even if such a feature would be needed, then we could add it to this decorator (besides checking a file, check param supplied in params).

> * when reusing a cluster that was created by someone (something) else. E.g. when someone prepares a reproducer, provisions and keeps a cluster and passes it to a dev who needs it for their work (update packages, rerun only subset of test stages, etc.)
> 

This could be overcomed by keeping info about completed stages on loader (we never destroy loaders).

> What do you think?

Generally, what I think is that adding feature doesn't mean people would use it because:
1. they don't know about it
2. they don't know how to use 
3. it is not obvious for them which stages might be skipped and which not

fruch:
> > > I'd propose making it automatic: create a decorator that will:
> > > 
> > > 1. write to a file info about given function that is completed (e.g line in e.g.  /tmp/completed-steps-<test_id> or somewhere in results dir)
> > > 2. If decorator sees that line and is reusing cluster, then skip decorated method.
> > > 
> > > Only decorate methods that are intended for skipping when reusing (remember about Perf tests methods too).
> > 
> > 
> > I think implementing the functionality this way would restrict us from some use cases:
> > 
> > * using the feature standalone, not only with reuse_cluster. Probably it won't be a common a scenario, but if a user would need to skip some test stages of a test when running it for th 1st time, the feature would help with it.
> 
> Even if such a feature would be needed, then we could add it to this decorator (besides checking a file, check param supplied in params).

regardless of needed or not, I don't want to tie this option with reuse_cluster, and even reuse cluster itself should be tied with local files, consider case I would run it from jenkins, each time from different builder/runner.

> 
> > * when reusing a cluster that was created by someone (something) else. E.g. when someone prepares a reproducer, provisions and keeps a cluster and passes it to a dev who needs it for their work (update packages, rerun only subset of test stages, etc.)
> 
> This could be overcomed by keeping info about completed stages on loader (we never destroy loaders).
> 
> > What do you think?
> 
> Generally, what I think is that adding feature doesn't mean people would use it because:
> 
> 1. they don't know about it
> 2. they don't know how to use
> 3. it is not obvious for them which stages might be skipped and which not

those statements are valid for every single feature we develop, if we'll fail to document and advertise it properly.

As for how common is this need, for working on perf test is quite common, and for any development we need to experiment a lot, and for reproductions of issues as well.

this was asking in multiple variations, as the work you had for using jupyter notebooks to exercise some specific SCT code, which people don't know either and don't use it.

configurations/optional_test_stages.yaml
```diff
@@ -0,0 +1,5 @@
+optional_test_stages:
+  prepare_write: false
+  stress: true
```
fruch, line 3: main_stress / main_load ?

dimakr: Changed

dimakr:
@fruch This change adds the underlying mechanism for skipping test stages, and uses it in longevity tests to mark the appropriate test stage methods/calls.
I suggest to have a separate PR(s) for enabling the mechanism in tests which structure differ from longevities (performance, manager, tests).

sdcm/utils/decorators.py
```diff
@@ -333,3 +333,30 @@ def wrapper(*args, **kwargs):
...
+            # making import here, to work around circular import issue
+            from sdcm.cluster import TestConfig
+            optional_test_stages = TestConfig().tester_obj().optional_test_stages
+            skipped_stages = [stage for stage in stage_names if not optional_test_stages[stage]]
+
+            if not skipped_stages:
+                return func(*args, **kwargs)
+            else:
+                skipped_stages_str = ', '.join(skipped_stages)
+                LOGGER.info(f"'{func.__name__}' is skipped as '{skipped_stages_str}' test stage(s) is disabled.")
```
fruch, line 360: Maybe warning, so it would be visible in stdout as well ? i.e. on Jenkins output ?

fruch (approved):
LGTM

sdcm/utils/decorators.py
dimakr: Agree. Changed this, now it is:
```
< t:2024-09-30 18:31:20,143 f:decorators.py   l:360  c:sdcm.utils.decorators p:WARNING > 'run_pre_create_keyspace' is skipped as 'prepare_write' test stage(s) is disabled.
< t:2024-09-30 18:31:20,143 f:decorators.py   l:360  c:sdcm.utils.decorators p:WARNING > 'run_pre_create_schema' is skipped as 'prepare_write' test stage(s) is disabled.
```

docs/skip-test-stages.md
```diff
@@ -0,0 +1,77 @@
...
+  data_validation: true
+EOF
+
+export SCT_SCYLLA_VERSION=6.1
+
+# run longevity test, passing the newly created configuration file
+hydra run-test longevity_test.LongevityTest.test_custom_time --backend aws --config configurations/nemesis/longevity-5gb-1h-nemesis.yaml --config configurations/network_config/test_communication_public.yaml --config configurations/optional_test_stages.yaml
+
+# or use environment variable to pass the 'optional_test_stages' parameter to the test
+export SCT_OPTIONAL_TEST_STAGES=$(cat ./configurations/optional_test_stages.yaml)
```
soyacz, line 61: this doesn't look like working... shouldn't be single quoted? Also, in examples json is provided while here it's going to be yaml (but maybe it works, I didn't try).

sdcm/utils/loader_utils.py
```diff
@@ -221,6 +224,7 @@ def run_post_prepare_cql_cmds(self):
             self.log.debug("Execute post prepare queries: %s", post_prepare_cql_cmds)
             self._run_cql_commands(post_prepare_cql_cmds)
 
+    @optional_stage('prepare_write')
     def run_prepare_write_cmd(self):
```
soyacz, line 228: these `run_prepare_write_cmd`s are starting nemesis when flag is set `if self.params.get('nemesis_during_prepare'):`, 
so skipping it will make nemesis thread not to be started.

sdcm/utils/decorators.py
```diff
@@ -333,3 +333,30 @@ def wrapper(*args, **kwargs):
...
+    """
+    stage_names = stage_names if isinstance(stage_names, list) else [stage_names]
+
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            # making import here, to work around circular import issue
+            from sdcm.cluster import TestConfig
+            optional_test_stages = TestConfig().tester_obj().optional_test_stages
+            skipped_stages = [stage for stage in stage_names if not optional_test_stages[stage]]
```
soyacz, line 354: this logic slightly differs from `def optional_stage_enabled` and seem to raise exception if a stage is not in `optional_test_stages`. Shouldn't reuse this function here?

soyacz (commented):
Few questions and remarks

sdcm/utils/loader_utils.py
dimakr: Agree. If I understand it correctly, the nemesis is 'included' into the test stage in this case - i.e. if stage is not executed the nemesis shouldn't as well.

soyacz: if someone wants a nemesis to start during prepare stage it doesn't mean he/she wants to skip it in case of skipping prepare stage.
It just specifies the moment of starting it.

dimakr: I see. The name of the parameter can be misleading then. Something like `start_nemesis_during_prepare` would be more clear.
I'll update from decorating `run_prepare_write_cmds` to using `optional_test_stages` function inside it.

sdcm/utils/decorators.py
dimakr: in tester.py the defaultdict is used to define `optional_test_stages` attr, where it returns default value for keys not present in the dict.
So both version will work, but i prefer this one, will update the function to use `optional_test_stages[stage]` as well.

docs/skip-test-stages.md
dimakr: > in examples json is provided while here it's going to be yaml

sct_config.py::dict_or_str function was updated to be more robust, so it can consume json and yaml formatted strings (just using ast.literal_eval, as it was previously, was failing even on some dumb case, when e.g. bollean is provded as 'true', but not 'True').

>this doesn't look like working... shouldn't be single quoted?

it does work, e.g.:
```
❯ export SCT_OPTIONAL_TEST_STAGES=$(cat ./configurations/optional_test_stages.yaml)
❯ echo $SCT_OPTIONAL_TEST_STAGES
optional_test_stages:
  prepare_write: true
  wait_total_space_used: true
```
But you helped to find the issue in this guide - the env. var shouldn't include `optional_test_stages` key itself, only it's value. I'll update this.

Then, for setting yaml-formatted string into env. variable it would require a command like:
```
❯ export SCT_OPTIONAL_TEST_STAGES=$( { read; cat; } < ./configurations/optional_test_stages.yaml )
❯ echo $SCT_OPTIONAL_TEST_STAGES
  prepare_write: true
  wait_total_space_used: true
  main_load: true
  post_test_load: true
  nemesis: true
  data_validation: true
```

Do you think we need to provide this more complicated example? Or providing only example of setting the env variable from json-formatted string is enough?

sdcm/utils/decorators.py
soyacz: hmm, I see, but this defaultdict returns `True` as default, shouldn't be False?

soyacz: maybe we should rename "SCT_OPTIONAL_TEST_STAGES" to "SCT_ENABLED_TEST_STAGES"?

docs/skip-test-stages.md
dimakr: Updated doc page

sdcm/utils/loader_utils.py
dimakr: Updated the prepare_write stage handling

sdcm/utils/decorators.py
dimakr: For me the `optional` version seems to be more descriptive - the name suggests that there could be optional stages in an SCT, and value defines which of them are (in the particular run).
WDYT?

fruch: the name optional fit the description of the feature

but @soyacz is correct that it not completely clear as to what setting something to true in that mapping.

true might be that it's optional, while the the true is that is enabled (mandatory to run)

so @soyacz suggestion doesn't clarify the situation a bit

dimakr: My approach to naming the parameter was:
- the parameter name implies that there is a feature of optionality of test stages. And when there is a value set for the parameter (in contrast to empty default value) then the feature is enabled
- the parameter value is mapping of such optional test stages names to an indication whether they will be executed during test run or not

Maybe we can combine suggestions and use something like `enabled_optional_test_stages`?

soyacz: from my side, word 'optional' means 'not run' by default - like when buying a car, you don't start with all the optional features on by default. And for typical run (not reuse cluster) these are not optional usually.
This is why I'm bit confused with it. Perhaps I'm the only one.

soyacz: another approach would be by inverting logic, and make "skip_test_stages"

docs/skip-test-stages.md
fruch: did you tried this ?
```
export SCT_OPTIONAL_TEST_STAGES.prepare_write='true'
```

sdcm/utils/decorators.py
dimakr: Added doc.string

docs/skip-test-stages.md
dimakr: This way of setting hierarchical env. var is failing with `export: not valid in this context: SCT_SKIP_TEST_STAGES.nemesis`.
Seems like the only way to set structured data in an env. var is to serialize it into json/yaml strings when setting.

cluster_configuration_tests.py
```diff
@@ -91,6 +92,7 @@ def test_change_seed_address_to_hostname(self):
                 node.start_scylla(verify_down=False, verify_up=True)
         base_cmd_w = self.params.get('stress_cmd')
 
-        # run a workload
-        cs_thread_pool = self.run_stress_thread(stress_cmd=base_cmd_w)
-        self.verify_stress_thread(cs_thread_pool=cs_thread_pool)
+        if not skip_optional_stage('main_load'):
```
soyacz, line 95: `skip_optional_stage` -> `skip_test_stage`

add_new_dc_test.py
```diff
@@ -69,6 +73,7 @@ def reconfigure_keyspaces_to_use_network_topology_strategy(self, keyspaces: List
             node.run_cqlsh(cql)
         self.log.info("Replication Strategies for {} reconfigured".format(keyspaces))
 
+    @optional_stage('prepare_write')
```
soyacz, line 76: `@optional_stage` -> `test_stage`

soyacz (commented):
I think one of the main use cases for this are perf regression tests where we play with load a lot and want to skip preloading data. Please check `performance_regression_gradual_grow_throughput.PerformanceRegressionPredefinedStepsTest` and `performance_regression_test.PerformanceRegressionTest`

dimakr:
> I think one of the main use cases for this are perf regression tests where we play with load a lot and want to skip preloading data. Please check `performance_regression_gradual_grow_throughput.PerformanceRegressionPredefinedStepsTest` and `performance_regression_test.PerformanceRegressionTest`

@soyacz I wanted to address non-longevity kind of tests - performance, manager -  in the next separatePR, after this one is approved and merged. This one is already hard to maintain.

add_new_dc_test.py
dimakr: I was having doubts if to change this one. `optional_stage` name is assuming some alternative - i.e. just from decorator name it can be read that the decorated function can be skipped (as it is optional).
`test_stage` is not clear from this perspective. Maybe something as `skip_if_test_stage_disabled` would be approrpiate?

soyacz: ok

soyacz (approved):
LGTM

fruch (approved):
LGTM

Issue #8720, opened 2024-09-16
`node.follow_system_log` may not find a really existing log line


vponomaryov:
### Issue description

- [ ] This issue is a regression.
- [x] It is unknown if this issue is a regression.

Running `disrupt_load_and_stream` nemesis we expect appearance of `start` and `end` log messages for the `load and stream` operation.

In current case in first 2 used nodes logs the expected log lines were correctly found:

node-5:
```
2024-09-14 03:34:08,395 f:load_utils.py   l:116  c:utils                p:INFO  > Running load and stream on the node longevity-large-partitions-200k-pks-db-node-f25dd9b8-0-5 for keyspace1.standard1'
...
2024-09-14T03:34:19.865+00:00 longevity-large-partitions-200k-pks-db-node-f25dd9b8-0-5     !INFO | scylla[5477]:  [shard  2:strm] sstables_loader - Done loading new SSTables for keyspace=keyspace1, table=standard1, load_and_stream=true, primary_replica_only=false, status=succeeded
...
2024-09-14 03:34:19,942 f:wait.py         l:161  c:sdcm.wait            p:DEBUG > Start line patterns ['(?:storage_service|sstables_loader) - load_and_stream:'] were found.
...
2024-09-14 03:34:19,954 f:wait.py         l:167  c:sdcm.wait            p:DEBUG > End line patterns ['(?:storage_service|sstables_loader) - Done loading new SSTables for keyspace=keyspace1, table=standard1, load_and_stream=true.*status=(.*)'] were found.
...
```

node-4:
```
2024-09-14 03:34:25,024 f:load_utils.py   l:116  c:utils                p:INFO  > Running load and stream on the node longevity-large-partitions-200k-pks-db-node-f25dd9b8-0-4 for keyspace1.standard1'
...
Sep 14 03:34:34.436147 longevity-large-partitions-200k-pks-db-node-f25dd9b8-0-4 scylla[5486]:  [shard  2:strm] sstables_loader - Done loading new SSTables for keyspace=keyspace1, table=standard1, load_and_stream=true, primary_replica_only=false, status=succeeded
...
2024-09-14 03:34:34,570 f:wait.py         l:161  c:sdcm.wait            p:DEBUG > Start line patterns ['(?:storage_service|sstables_loader) - load_and_stream:'] were found.
...
[...]

PR #8743, opened 2024-09-18
fix(wait_for_log_lines): fix saturating cpu when waiting for log lines

soyacz:
We missed waiting time before recreating iterator. This causes unnecessary high cpu utilization.

refs: https://github.com/scylladb/scylla-cluster-tests/issues/8720

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/wait.py
+++ b/sdcm/wait.py
@@ -156,12 +156,14 @@ def wait_for_log_lines(node, start_line_patterns, end_line_patterns, start_timeo
         started = any(start_follower)
         while not started and (time.time() - start_time < start_timeout):
             started = any(start_follower)
+            time.sleep(0.1)
         if not started:
             raise TimeoutError(start_ctx)
         LOGGER.debug("Start line patterns %s were found.%s", start_line_patterns, error_msg_ctx)
         ended = any(end_follower)
         while not ended and (time.time() - start_time < end_timeout):
             ended = any(end_follower)
+            time.sleep(0.1)
         if not ended:
             raise TimeoutError(end_ctx)
         LOGGER.debug("End line patterns %s were found.%s", end_line_patterns, error_msg_ctx)
--- a/unit_tests/test_wait.py
+++ b/unit_tests/test_wait.py
@@ -223,7 +223,7 @@ class TestWaitForLogLines:
         t.daemon = True
         file_path.touch()
         with pytest.raises(TimeoutError, match="Timeout occurred while waiting for start log line"), \
-                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.3, end_timeout=1):
+                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.4, end_timeout=1.2):
             t.start()
 
     def test_wait_for_log_timeout_when_no_end_line(self, tmp_path):
@@ -234,7 +234,7 @@ class TestWaitForLogLines:
         t.daemon = True
         file_path.touch()
         with pytest.raises(TimeoutError, match="Timeout occurred while waiting for end log line"), \
-                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.1, end_timeout=0.3):
+                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.5, end_timeout=0.7):
             t.start()
 
     def test_wait_for_log_reraises_exception(self, tmp_path):
@@ -245,7 +245,7 @@ class TestWaitForLogLines:
         t.daemon = True
         file_path.touch()
         with pytest.raises(ValueError, match="dummy error"), \
-                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.1, end_timeout=0.3):
+                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.5, end_timeout=0.7):
             t.start()
             raise ValueError('dummy error')
 
@@ -254,7 +254,7 @@ class TestWaitForLogLines:
         node = DummyNode(log_path=file_path)
         file_path.touch()
         with pytest.raises(TimeoutError, match="Timeout occurred while waiting for start log line") as exc_info, \
-                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.1, end_timeout=0.3):
+                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7):
             raise ValueError('dummy error')
         assert "ValueError" in str(exc_info.getrepr())
 
@@ -267,5 +267,5 @@ class TestWaitForLogLines:
         file_path.touch()
         expected_match = "Timeout occurred while waiting for end log line \['end'\] on node: node_1. Context: Wait end line"
         with pytest.raises(TimeoutError, match=expected_match), \
-                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.1, end_timeout=0.3, error_msg_ctx="Wait end line"):
+                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7, error_msg_ctx="Wait end line"):
             t.start()


fruch:
@soyacz 

this change is breaking the unittest, that set smaller timeouts then those sleeps

soyacz:
> @soyacz
> 
> this change is breaking the unittest, that set smaller timeouts then those sleeps

should be fixed (added more margins and reduced the sleep to 0.05)

unit_tests/test_wait.py
```diff
@@ -223,7 +223,7 @@ def test_wait_for_log_timeout_when_no_start_line(self, tmp_path):
         t.daemon = True
         file_path.touch()
         with pytest.raises(TimeoutError, match="Timeout occurred while waiting for start log line"), \
-                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.3, end_timeout=1):
+                wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.5, end_timeout=1.5):
```
vponomaryov, line 226: Need just to mock the `time.sleep` per a unit test.
It is not needed at all to wait for real time intervals.

Moreover, we should not change time values just to make unit tests work.
If it was needed to have `0.1s` sleep, then usage of a mock should completely solve the problem.

vponomaryov (changes requested):

unit_tests/test_wait.py
soyacz: fixed

sdcm/wait.py
```diff
@@ -156,12 +156,14 @@ def wait_for_log_lines(node, start_line_patterns, end_line_patterns, start_timeo
         started = any(start_follower)
         while not started and (time.time() - start_time < start_timeout):
             started = any(start_follower)
+            time.sleep(0.05)
```
vponomaryov, line 159: I think that `0.05` is too small for this case.
Our time windows are measured mostly in minutes.

unit_tests/test_wait.py
```diff
@@ -210,10 +211,11 @@ def test_can_wait_for_log_start_line_and_end_line(self, tmp_path):
...
         file_path.touch()
-        with wait_for_log_lines(node=node, start_line_patterns=["rebuild.*started with keyspaces=", "Rebuild starts"],
-                                end_line_patterns=["rebuild.*finished with keyspaces=", "Rebuild succeeded"],
-                                start_timeout=3, end_timeout=5):
-            write_thread.start()
+        with patch('time.sleep', return_value=None):
+            with wait_for_log_lines(node=node, start_line_patterns=["rebuild.*started with keyspaces=", "Rebuild starts"],
+                                    end_line_patterns=["rebuild.*finished with keyspaces=", "Rebuild succeeded"],
+                                    start_timeout=3, end_timeout=5):
+                write_thread.start()
```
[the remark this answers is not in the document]
soyacz: still, there's a thread in background that pushes log lines in 0.1 interval, I don't think we should modify these tests more.

sdcm/wait.py
soyacz: doesn't make much difference - waking up 200 times a second is not a big deal.

unit_tests/test_wait.py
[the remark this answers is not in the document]
vponomaryov: You can use `assert_called_once()`. It is useful to know that it was called when we have it as a `fix`.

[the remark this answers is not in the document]
soyacz: I don't see the point of asserting constants. Only will cause test failure if we decide to change it.

sdcm/wait.py
fruch: it's way better than the zero time it was sleeping before :)

Let's start with that, and we'll see if we need to adapt it further

unit_tests/test_wait.py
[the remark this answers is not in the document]
fruch: Anyhow if something would be change in the implementation, test would need to change.

it's make sense to test argument to such call, if we had any logic involved, without one, it's doesn't add much.

fruch (dismissed):
LGTM

soyacz:
Even if patching `time.sleep` tests were not stable.
I adjusted a bit timeouts to make it more stable.
Also prolonged sleep to 0.1s (so actually, waking up 10 times a sec instead 20).

sdcm/wait.py
vponomaryov, line 159: Is it intentional that we have 2 different values for 2 sleeps?

soyacz: no, forgotten about this one.

soyacz: fixed

fruch:
@soyacz unittest now fails again..

vponomaryov:
> @soyacz unittest now fails again..

It passes for me locally.
Interesting, may it reproduced the bug?

soyacz:
> > @soyacz unittest now fails again..
> 
> It passes for me locally and the failed test takes about `0.04s`. Interesting, maybe it reproduced the bug?



> > @soyacz unittest now fails again..
> 
> It passes for me locally and the failed test takes about `0.04s`. Interesting, maybe it reproduced the bug?

could be, but now with patching, my fix is disabled for unit tests :stuck_out_tongue:

vponomaryov:
> > > @soyacz unittest now fails again..
> > 
> > 
> > It passes for me locally and the failed test takes about `0.04s`. Interesting, maybe it reproduced the bug?
> 
> > > @soyacz unittest now fails again..
> > 
> > 
> > It passes for me locally and the failed test takes about `0.04s`. Interesting, maybe it reproduced the bug?
> 
> could be, but now with patching, my fix is disabled for unit tests 😛

It got reproduced with current PR state - with patching.

vponomaryov:
I managed to repro bug making run the test thousand times:

```
$ git diff
diff --git a/unit_tests/test_wait.py b/unit_tests/test_wait.py
index 27ef74c86..1f56b113c 100644
--- a/unit_tests/test_wait.py
+++ b/unit_tests/test_wait.py
@@ -242,17 +242,20 @@ class TestWaitForLogLines:
                 t.start()
 
     def test_wait_for_log_reraises_exception(self, tmp_path):
-        file_path = tmp_path / "wait_for_log_lines_1.log"
-        node = DummyNode(log_path=file_path)
-        lines = ["start", "cde", "end"]  # no start line provided
-        t = threading.Thread(target=write_to_file, args=(file_path, lines))
-        t.daemon = True
-        file_path.touch()
-        with patch('time.sleep', return_value=None):
-            with pytest.raises(ValueError, match="dummy error"), \
-                    wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7):
-                t.start()
-                raise ValueError('dummy error')
+        for i in range(1, 1111):
+            file_path = tmp_path / "wait_for_log_lines_{}.log".format(i)
+            node = DummyNode(log_path=file_path)
+            lines = ["fake", "start", "cde", "end", "foo"]  # no start line provided
+            t = threading.Thread(target=write_to_file, args=(file_path, lines))
+            t.daemon = True
+            file_path.touch()
+            with patch('time.sleep', return_value=None):
+                with pytest.raises(ValueError, match="dummy error"), \
+                        wait_for_log_lines(
+                            node=node, start_line_patterns=["start"],
+                            end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7):
+                    t.start()
+                    raise ValueError('dummy error')
 
     def test_wait_for_log_reraises_exception_and_timeout_error(self, tmp_path):
         file_path = tmp_path / "wait_for_log_lines_1.log"

```

Exactly second follower gets timeout when it happens.

soyacz:
> I managed to repro bug making run the test thousand times:
> 
> ```
> $ git diff
> diff --git a/unit_tests/test_wait.py b/unit_tests/test_wait.py
> index 27ef74c86..1f56b113c 100644
> --- a/unit_tests/test_wait.py
> +++ b/unit_tests/test_wait.py
> @@ -242,17 +242,20 @@ class TestWaitForLogLines:
>                  t.start()
>  
>      def test_wait_for_log_reraises_exception(self, tmp_path):
> -        file_path = tmp_path / "wait_for_log_lines_1.log"
> -        node = DummyNode(log_path=file_path)
> -        lines = ["start", "cde", "end"]  # no start line provided
> -        t = threading.Thread(target=write_to_file, args=(file_path, lines))
> -        t.daemon = True
> -        file_path.touch()
> -        with patch('time.sleep', return_value=None):
> -            with pytest.raises(ValueError, match="dummy error"), \
> -                    wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7):
> -                t.start()
> -                raise ValueError('dummy error')
> +        for i in range(1, 1111):
> +            file_path = tmp_path / "wait_for_log_lines_{}.log".format(i)
> +            node = DummyNode(log_path=file_path)
> +            lines = ["fake", "start", "cde", "end", "foo"]  # no start line provided
> +            t = threading.Thread(target=write_to_file, args=(file_path, lines))
> +            t.daemon = True
> +            file_path.touch()
> +            with patch('time.sleep', return_value=None):
> +                with pytest.raises(ValueError, match="dummy error"), \
> +                        wait_for_log_lines(
> +                            node=node, start_line_patterns=["start"],
> +                            end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7):
> +                    t.start()
> +                    raise ValueError('dummy error')
>  
>      def test_wait_for_log_reraises_exception_and_timeout_error(self, tmp_path):
>          file_path = tmp_path / "wait_for_log_lines_1.log"
> ```
> 
> Exactly second follower gets timeout when it happens.

and if you remove 'with patch('time.sleep', return_value=None):`? still reproduces?

vponomaryov:
> > I managed to repro bug making run the test thousand times:
> > ```
> > $ git diff
> > diff --git a/unit_tests/test_wait.py b/unit_tests/test_wait.py
> > index 27ef74c86..1f56b113c 100644
> > --- a/unit_tests/test_wait.py
> > +++ b/unit_tests/test_wait.py
> > @@ -242,17 +242,20 @@ class TestWaitForLogLines:
> >                  t.start()
> >  
> >      def test_wait_for_log_reraises_exception(self, tmp_path):
> > -        file_path = tmp_path / "wait_for_log_lines_1.log"
> > -        node = DummyNode(log_path=file_path)
> > -        lines = ["start", "cde", "end"]  # no start line provided
> > -        t = threading.Thread(target=write_to_file, args=(file_path, lines))
> > -        t.daemon = True
> > -        file_path.touch()
> > -        with patch('time.sleep', return_value=None):
> > -            with pytest.raises(ValueError, match="dummy error"), \
> > -                    wait_for_log_lines(node=node, start_line_patterns=["start"], end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7):
> > -                t.start()
> > -                raise ValueError('dummy error')
> > +        for i in range(1, 1111):
> > +            file_path = tmp_path / "wait_for_log_lines_{}.log".format(i)
> > +            node = DummyNode(log_path=file_path)
> > +            lines = ["fake", "start", "cde", "end", "foo"]  # no start line provided
> > +            t = threading.Thread(target=write_to_file, args=(file_path, lines))
> > +            t.daemon = True
> > +            file_path.touch()
> > +            with patch('time.sleep', return_value=None):
> > +                with pytest.raises(ValueError, match="dummy error"), \
> > +                        wait_for_log_lines(
> > +                            node=node, start_line_patterns=["start"],
> > +                            end_line_patterns=["end"], start_timeout=0.4, end_timeout=0.7):
> > +                    t.start()
> > +                    raise ValueError('dummy error')
> >  
> >      def test_wait_for_log_reraises_exception_and_timeout_error(self, tmp_path):
> >          file_path = tmp_path / "wait_for_log_lines_1.log"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Exactly second follower gets timeout when it happens.
> 
> and if you remove 'with patch('time.sleep', return_value=None):`? still reproduces?

No and if I use a separate `start_time` value for the second follower, it also works.

I think, the proper fix here must be combined from the following changes:
- Stop using 2 followers and use just 1 to avoid double work. In real test runs it may be thousands of huge lines processing that could be avoided.
- Stop counting log line processing into the timeout. We should stick to the time marks of the logs

soyacz:
> > 
> > and if you remove 'with patch('time.sleep', return_value=None):`? still reproduces?
> 
> No and if I use a separate `start_time` value for the second follower, it also works.
> 
> I think, the proper fix here must be combined from the following changes:
> 
> * Stop using 2 followers and use just 1 to avoid double work. In real test runs it may be thousands of huge lines processing that could be avoided.

Removing one follower and appending regexps of the other won't make much difference - still each line will have to go through the same regexps. Improvement wont' be big.

> * Stop counting log line processing into the timeout. We should stick to the time marks of the logs

I also trying without patch (using `pytest-repeat` package) and it works.
It's not trivial to remove 1 follower and keep current behavior, where we intentionally wait for start message and end message.
Let's see if merging this patch erases the issue and if not, let's proceed with removing 1 follower.

soyacz:
tests pass, I think we should merge it.

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #8759, opened 2024-09-19
fix(get_my_public_ip): update the list of services

fruch:
some of the services we are using are a bit buggy
and was returning us ipv6 address, even when we were asking for ipv4 one.

so revising the list of services used, and changing the validation to be more specific and validate it's ipv4.

this was failing docker-artifact housekeeping check, cause addresses wasn't matching, and started out of the blue, without any code changes in that portions of the code (and no change relevant from scylla end as well).

### Testing

- [x]  🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-docker-test/28/

### Reminders

### DIFF
--- a/sdcm/utils/net.py
+++ b/sdcm/utils/net.py
@@ -5,7 +5,11 @@ import ipaddress
 
 import os
 import socket
+import logging
+
 import requests
+
+LOGGER = logging.getLogger(__name__)
 
 
 def get_sct_runner_ip() -> str:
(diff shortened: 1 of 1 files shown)


sdcm/utils/net.py
```diff
@@ -31,8 +31,8 @@ def get_my_public_ip() -> str:
             result.raise_for_status()
 
             ip_address = result.text.strip()
-            ipaddress.ip_address(ip_address)  # validating that we got IP address
-        except (ValueError, requests.exceptions.RequestException):
+            ipaddress.IPv4Address(ip_address)  # validating that we got IPv4 address
```
vponomaryov, line 34: Cannot we return from this place?

sdcm/utils/net.py
```diff
@@ -31,8 +31,8 @@ def get_my_public_ip() -> str:
...
+        except (ipaddress.AddressValueError, ipaddress.NetmaskValueError, requests.exceptions.RequestException):
             if idx == len(hostnames) - 1:
```
vponomaryov, line 36: Since we want to go over all services why then we do not consider the presence of an already detected IP address in the `ip_address` var?

For the moment the last `https://myip.dnsomatic.com` service is unresponsive and we will get unneeded exception on the line 37.

sdcm/utils/net.py
fruch: changed the code to return as soon it find a valid ip

soyacz (approved):
LGTM

sdcm/utils/net.py
fruch: anyhow that code was trying to make sure we propagate the actual failure back.
but was doing excess calls to all of those services.

since we use that in one single case, we didn't noticed too many issue with it.

vponomaryov (approved):
LGTM

PR #8774, opened 2024-09-19
fix(database_query_utils): make PartitionRowsValidationEvent contains diff between before and after state

dkropachev:
When there is difference test ends with following message:
```
sdcm.sct_events.events_analyzer.TestFailure: Got critical event: (PartitionRowsValidationEvent Severity.CRITICAL) period_type=one-time event_id=318ca98d-c349-4cad-bddc-889918d9f326: message=Row amount in partitions is not same before and after running of nemesis: {0: 5555, 2: 5555, 5: 5555, 7: 5555, 8: 5555, 11: 5555, 12: 5555, 13: 5555, 14: 5555, ....}
```

Which does not tell you what is the difference.

### Testing
Not tested

### DIFF
--- a/sdcm/utils/database_query_utils.py
+++ b/sdcm/utils/database_query_utils.py
@@ -81,7 +81,7 @@ class PartitionsValidationAttributes:  # pylint: disable=too-few-public-methods,
                             ' using timeout 5m'
         return count_pk_rows_cmd
 
-    def collect_partitions_info(self, ignore_limit_rows_number: bool = False) -> dict | None:
+    def collect_partitions_info(self, ignore_limit_rows_number: bool = False) -> dict[int, int] | None:
         # Get and save how many rows in each partition.
         # It may be used for validation data in the end of test.
         # By default, the count is limited to partitions_attributes.limit_rows_number (if exist),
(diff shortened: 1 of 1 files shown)


yarongilor (dismissed):
LGTM

yarongilor:
Should fix commit message.

fruch:
@dkropachev @yarongilor 

can we have a run of it, to validate it's working as expected ?

also, we need to fix the commit message commitlint doesn't agree with it

yarongilor:
> @dkropachev @yarongilor
> 
> can we have a run of it, to validate it's working as expected ?
> 
> also, we need to fix the commit message commitlint doesn't agree with it

tested in https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/byo-longevity-test-yg/118/

yarongilor:
> > @dkropachev @yarongilor
> > can we have a run of it, to validate it's working as expected ?
> > also, we need to fix the commit message commitlint doesn't agree with it
> 
> @dkropachev , @fruch , The [test](https://argus.scylladb.com/tests/scylla-cluster-tests/ff5696df-ab11-43b8-a5be-f43957e69c60) ran fine. One small issue is it reports a diff, so it is less informative IMO. meaning, it would be better and more useful to see: {0: '0' - '5555', 1: '123' - '4321'} so we can tell for sure, the partition 0 is totally gone, for example, while partition 1 not.
> 
> ```
> 2025-02-25 12:14:53.085: (PartitionRowsValidationEvent Severity.CRITICAL) period_type=one-time event_id=312a2d4c-7ad6-4bae-901d-33b49c050d68: message=Row amount in partitions is not same before and after running of nemesis, difference(after-before): {0: '-5555', 1: '-5555', 2: '-5555', 3: '-5555'}
> ```

ping @fruch , @dkropachev . It would also be fine to merge as is.

yarongilor (approved):

PR #6192, opened 2023-05-28
feature(artifact_test): testing the output of perftune.py

ShlomiBalalis:
perftune.py is a script within seastar (which is a part of scylla_setup)
that configures several system parameters, like the amount of CPU cores that
are designated to handle IRQ, and which are not.

I have recorded the expected results as of master 20230717.567b4536892f
in defaults/perftune_results.json, and the test will compare them to
the output of the script.

https://github.com/scylladb/scylla-enterprise/issues/2909

## PR pre-checks (self review)

- [x] I followed [KISS principle](https://en.wikipedia.org/wiki/KISS_principle) and [best practices](https://docs.google.com/document/d/1jihgOKb5iGRlD8_HQ92O0JbLk1kASUoZT23i_MXFSKI)
- [x] I didn't leave commented-out/debugging code
- [x] I added the relevant `backport` labels
- [ ] New configuration option are added and documented (in `sdcm/sct_config.py`)
- [ ] I have added tests to cover my changes (Infrastructure only - under `unit-test/` folder)
- [ ] All new and existing unit tests passed (CI)
- [ ] I have updated the Readme/doc folder accordingly (if needed)

PR #8788, opened 2024-09-19
feature(artifact_test): testing the output of perftune.py

fruch:
perftune.py is a script within seastar (which is a part of scylla_setup) that configures several system parameters, like the amount of CPU cores that are designated to handle IRQ, and which are not.

Ref: https://github.com/scylladb/scylla-enterprise/issues/2909

Replacement of #6192

### Testing

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-docker-test/34/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ami-test/16/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-azure-image-test/52/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-gce-image-test/4/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ami-arm-test/5/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-rocky8-test/14/

### Reminders

### DIFF
--- a/sdcm/utils/perftune_validator.py
+++ b/sdcm/utils/perftune_validator.py
@@ -1,0 +1,238 @@
+import logging
+import random
+import traceback
+from functools import cached_property
+
+
+import yaml
+from deepdiff import DeepDiff
+
+from sdcm.sct_events import Severity
+from sdcm.sct_events.system import PerftuneResultEvent
+from sdcm.utils.version_utils import get_systemd_version
+
+
+# https://docs.scylladb.com/stable/operating-scylla/admin-tools/perftune.html
+
+
+PERFTUNE_LOCATION = "/opt/scylladb/scripts/perftune.py"
+TEMP_PERFTUNE_YAML_PATH = "/tmp/perftune.yaml"
+PERFTUNE_EXPECTED_RESULTS_PATH = "defaults/perftune_results.json"
+
+
+def get_machine_architecture_type(node):
+    result = node.remoter.run("uname -m")
+    return result.stdout.strip()
+
+
+def get_number_of_cpu_cores(node) -> int:
+    result = node.remoter.run("grep -c ^processor /proc/cpuinfo")
+    cores_num = int(result.stdout)
+    return cores_num
+
+
+class PerftuneExpectedResult:
+    def __init__(self, number_of_cpu_cores, nic_name, architecture):
+        self.nic_name = nic_name
+        self.total_cpu_bits = number_of_cpu_cores
+        self.architecture = architecture
+        self.expected_compute_bits: int = 0
+
+        match number_of_cpu_cores:
+            case 1 | 2 | 4:
+                self.expected_irq_bits: int = number_of_cpu_cores
+                self.expected_compute_bits: int = number_of_cpu_cores
+            case 8:
+                self.expected_irq_bits: int = 1
+            case 12 | 24 | 32:
+                self.expected_irq_bits: int = 2
+            case 16:
+                if self.architecture == "aarch64":
+                    self.expected_irq_bits: int = 1
+                else:
+                    self.expected_irq_bits: int = 2
+            case 48 | 64:
+                self.expected_irq_bits: int = 4
+            case 80 | 72 | 96 | 128:
+                self.expected_irq_bits: int = 8
+
+        if self.expected_compute_bits == 0:
+            self.expected_compute_bits: int = number_of_cpu_cores - self.expected_irq_bits
+
+    @staticmethod
+    def count_bits(mask_string) -> int:
+        return sum([int(m, base=16).bit_count() for m in mask_string.split(',')])
+
+    def get_expected_options_file_contents(self) -> dict:
+        return {
+            "nic": [self.nic_name],
+            "irq_core_auto_detection_ratio": 16,
+            "tune": ["net"]
+        }
+
+
+class PerftuneExecutor:
+    def __init__(self, node, nic_name):
+        self.node = node
+        self.nic_name = nic_name
+
+    def get_cpu_mask(self) -> str:
+        result = self.node.remoter.run(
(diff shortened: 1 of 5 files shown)


defaults/perftune_results.json
```diff
@@ -0,0 +1,179 @@
...
+      },
+      "32": {
+        "get-cpu-mask": "0xfffefffe",
+        "get-irq-cpu-mask": "0x00010001",
+        "dump-options-file": {"cpu_mask": "0xffffffff", "irq_cpu_mask": "0x00010001"}
+      },
+      "48": {
+        "get-cpu-mask": "0x0000ffff,0xfcfffffc",
+        "get-irq-cpu-mask": "0x03000003",
+        "dump-options-file": {"cpu_mask": "0x0000ffff,0xffffffff", "irq_cpu_mask": "0x03000003"}
```
soyacz, line 43: won't these results depend on instances generation? e.g. when AWS introduces i5i, we start to fail these tests .
also, how do you know these are correct :sweat_smile: ?

fruch: I don't know if it's correct or not

this is a PR started by shlomo, and I've trying to see what we can get out of it.

I'm playing with throwing this file away, and just validate there are enough bits on, on each mask.

roydahan: I don't remember what was the outcome of our call with Shlomo, but I recall I wasn't happy with the methodology here.
Need to refresh my memory with the original PR.

fruch: This file is going away, and now I'll do the calculation of what is expected inside the code

There won't be exact masks, but the number of CPUs for scylla and number of CPUs for IRQ

sdcm/utils/perftune_validator.py
```diff
@@ -0,0 +1,238 @@
...
+
+
+class PerftuneExpectedResult:
+    def __init__(self, number_of_cpu_cores, nic_name, architecture):
+        self.nic_name = nic_name
+        self.total_cpu_bits = number_of_cpu_cores
+        self.architecture = architecture
+        self.expected_compute_bits: int = 0
+
+        match number_of_cpu_cores:
```
fruch, line 41: @roydahan 

here's the new logic, it's just maps to how many irq cpus we expect.

sdcm/utils/perftune_validator.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def compare_perftune_results(self) -> None:
+        PerftuneResultEvent(
+            message="Checking the output of perftune.py",
+            severity=Severity.NORMAL).publish()
+        try:
+            self.compare_cpu_mask()
+            self.compare_irq_cpu_mask()
+            if not self.node.is_nonroot_install:
+                # The following checks are not applicable for non-root installations,
```
fruch, line 228: I don't know if it's intended or not, but non-root installs doesn't call perftune at all.

fruch:
@roydahan have you seen this one ?

sdcm/utils/perftune_validator.py
```diff
@@ -0,0 +1,238 @@
...
+            case 1 | 2 | 4:
+                self.expected_irq_bits: int = number_of_cpu_cores
+                self.expected_compute_bits: int = number_of_cpu_cores
+            case 8:
+                self.expected_irq_bits: int = 1
+            case 12 | 24 | 32:
+                self.expected_irq_bits: int = 2
+            case 16:
+                if self.architecture == "aarch64":
+                    self.expected_irq_bits: int = 1
```
roydahan, line 51: @vladzcloudius isn't this a bug?
Why would we set only 1 core for irq in arm?

sdcm/utils/perftune_validator.py
```diff
@@ -0,0 +1,238 @@
...
+                severity=Severity.ERROR).publish()
+
+    def compare_perftune_results(self) -> None:
+        PerftuneResultEvent(
+            message="Checking the output of perftune.py",
+            severity=Severity.NORMAL).publish()
+        try:
+            self.compare_cpu_mask()
+            self.compare_irq_cpu_mask()
+            if not self.node.is_nonroot_install:
```
soyacz, line 227: could return here to drop nesting.

soyacz (commented):
Maybe it's a good opportunity to verify defaults shipped with AMI? Could be in followup too if we want it.

fruch:
> Maybe it's a good opportunity to verify defaults shipped with AMI? Could be in followup too if we want it.

defaults you mean the content of `io_properites.yaml` ?

sdcm/utils/perftune_validator.py
```diff
@@ -0,0 +1,238 @@
...
+
+class PerftuneExpectedResult:
+    def __init__(self, number_of_cpu_cores, nic_name, architecture):
+        self.nic_name = nic_name
+        self.total_cpu_bits = number_of_cpu_cores
+        self.architecture = architecture
+        self.expected_compute_bits: int = 0
+
+        match number_of_cpu_cores:
+            case 1 | 2 | 4:
```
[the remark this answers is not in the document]
fruch: as it written like that it's much clearer which options we really tested out.

soyacz:
> > Maybe it's a good opportunity to verify defaults shipped with AMI? Could be in followup too if we want it.
> 
> defaults you mean the content of `io_properites.yaml` ?

yes, how much we differ from that - shouldn't be much (what is much I don't know yet)

fruch:
> > > Maybe it's a good opportunity to verify defaults shipped with AMI? Could be in followup too if we want it.
> > 
> > 
> > defaults you mean the content of `io_properites.yaml` ?
> 
> yes, how much we differ from that - shouldn't be much (what is much I don't know yet)

this would be a followup, and a bit more complex to define clearly 
we are struggling with it on some perf issues.

roydahan (approved):

PR #8802, opened 2024-09-22
fix(nemesis): skip `disrupt_no_corrupt_repair` if needed

fruch:
since there currently no for fix for scylladb/scylladb#18059 we'll be skipping this nemesis until a fix for that issue would be available

Ref: scylladb/scylladb#18059

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1628,6 +1628,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self._kill_scylla_daemon()
 
     def disrupt_no_corrupt_repair(self):
+
+        if SkipPerIssues("https://github.com/scylladb/scylladb/issues/18059", self.tester.params):
+            raise UnsupportedNemesis('Disabled due to https://github.com/scylladb/scylladb/issues/18059 not fixed yet')
+
         # prepare test tables and fill test data
         for i in range(10):
             self.log.debug('Prepare test tables if they do not exist')


roydahan:
On one hand, I don't like it, but OTOH we can't continue wasting our time in investigations of the same issues...

soyacz (approved):
LGTM

soyacz:
> On one hand, I don't like it, but OTOH we can't continue wasting our time in investigations of the same issues...
> 
> @fruch, if we merge it, add to the issue a lable that reminds us it's being skipped, something like sct-skip.

If issue is closed then `SkipPerIssue` should result in `False` and contiune with nemesis.

elcallio (approved):

roydahan:
> > On one hand, I don't like it, but OTOH we can't continue wasting our time in investigations of the same issues...
> > @fruch, if we merge it, add to the issue a lable that reminds us it's being skipped, something like sct-skip.
> 
> If issue is closed then `SkipPerIssue` should result in `False` and contiune with nemesis.

I know, I just want a reference on the issue itself that it's being skipped from testing.

fruch:
> > > On one hand, I don't like it, but OTOH we can't continue wasting our time in investigations of the same issues...
> > > @fruch, if we merge it, add to the issue a lable that reminds us it's being skipped, something like sct-skip.
> > 
> > If issue is closed then `SkipPerIssue` should result in `False` and contiune with nemesis.
> 
> I know, I just want a reference on the issue itself that it's being skipped from testing.

Maybe we should do the same idea we had for dtest ? automatically tag the issues ?

roydahan:
Yes, we should.

PR #8818, opened 2024-09-23
Report Manager restore results to Argus

mikliapko:
Closes https://github.com/scylladb/scylla-manager/issues/4027
Closes https://github.com/scylladb/scylla-manager/issues/4016

The goal is to keep and visualize Manager restore benchmark results in Argus.

Changes:
- Implemented the separate function to send Manager restore benchmark results to Argus. Resulting table consist of three fields: overall restore, download and load&stream and post-restore repair durations;
- Such Argus handler is added into restore benchmark tests;
- New properties were added into Manager cli to return info about version timestamp and post restore repair duration.
- Added jenkinsfile for manager restore benchmark test which will be included into release pipelines. Thus, we will
have the history of restore speed improvements or degradation release after release.

### Testing

- [x] https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/restore-benchmark-test-test/3/
- [x] https://jenkins.scylladb.com/view/scylla-manager/job/manager-master/job/restore-benchmark-test-test/4/

Example of graphs in [Argus](https://argus.scylladb.com/workspace?state=WyJlYjkzYjA2MS1mNjZiLTQ5NDQtYWIxMi0wZjhjMjI2NmYzODgiXQ) (the dataset is very small (1Gb) exclusively for testing purposes):
image

### Reminders

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -30,8 +30,9 @@ import yaml
 from invoke import exceptions
 
 from sdcm import mgmt
+from sdcm.argus_results import send_manager_benchmark_results_to_argus
 from sdcm.mgmt import ScyllaManagerError, TaskStatus, HostStatus, HostSsl, HostRestStatus
-from sdcm.mgmt.cli import ScyllaManagerTool
+from sdcm.mgmt.cli import ScyllaManagerTool, RestoreTask
 from sdcm.mgmt.common import reconfigure_scylla_manager, get_persistent_snapshots
 from sdcm.provision.helpers.certificate import TLSAssets
 from sdcm.remote import shell_script_cmd
@@ -1359,6 +1360,22 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         extra_params = self.params.get('mgmt_restore_extra_params')
         return extra_params if extra_params else None
 
+    def _send_restore_results_to_argus(self, task: RestoreTask, manager_version_timestamp: int,
+                                       dataset_label: str = None):
+        total_restore_time = int(task.duration.total_seconds())
+        repair_time = int(task.post_restore_repair_duration.total_seconds())
+        results = {
+            "restore time": (total_restore_time - repair_time),
+            "repair time": repair_time,
+            "total": total_restore_time,
+        }
+        send_manager_benchmark_results_to_argus(
+            argus_client=self.test_config.argus_client(),
+            result=results,
+            sut_timestamp=manager_version_timestamp,
+            row_name=dataset_label,
+        )
+
     def test_backup_and_restore_only_data(self):
         """The test is extensively used for restore benchmarking purposes and consists of the following steps:
         1. Populate the cluster with data (currently operates with datasets of 500GB, 1TB, 2TB, 5TB);
@@ -1388,6 +1405,9 @@ class MgmtCliTest(BackupFunctionsMixIn, ClusterTester):
         task = self.restore_backup_with_task(mgr_cluster=mgr_cluster, snapshot_tag=backup_task.get_snapshot_tag(),
                                              timeout=110000, restore_data=True, extra_params=extra_params)
         self.manager_test_metrics.restore_time = task.duration
+
+        manager_version_timestamp = manager_tool.sctool.client_version_timestamp
+        self._send_restore_results_to_argus(task, manager_version_timestamp)
 
         self.run_verification_read_stress()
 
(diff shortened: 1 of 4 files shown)


sdcm/argus_results.py
```diff
@@ -170,3 +170,16 @@ def _get_status_based_on_previous_results(metric: str):
...
+def send_manager_benchmark_results_to_argus(argus_client: ArgusClient, result: dict, sut_timestamp: int):
+    class RestoreResult(GenericResultTable):
+        class Meta:
+            name = f"Restore duration"
+            description = "Restore duration"
+            Columns = [ColumnMetadata(name=param, unit="", type=ResultType.INTEGER) for param in result.keys()]
+
+    result_table = RestoreResult(sut_timestamp=sut_timestamp)
+    for key, value in result.items():
+        result_table.add_result(column=key, row="#1", value=value, status=Status.PASS)
```
soyacz, line 184: worth to add dataset size, e.g. instead of labeling row with `#1` could use `1TB` or snapshot name `1tb_1t_ics`

mgmt_cli_test.py
```diff
@@ -1359,6 +1360,16 @@ def get_restore_extra_parameters(self) -> str:
...
         return extra_params if extra_params else None
 
+    def _send_restore_results_to_argus(self, task: RestoreTask, manager_version_timestamp: int):
+        restore_time = task.duration.total_seconds()
+        repair_time = task.post_restore_repair_duration.total_seconds()
+        results = {
+            "overall_restore": restore_time,
+            "download_and_load&stream": (restore_time - repair_time),
+            "repair": repair_time,
+        }
```
soyacz, line 1370: please make it ordered dict - as it might update Argus table metadata randomly due change in columns ordering.
also, consider better naming and order (like: restore time, repair time and total)

mikliapko: Didn't get your point about ordered dict as since Python3.7 basic Python dict is ordered by default.

sdcm/argus_results.py
```diff
@@ -170,3 +170,20 @@ def _get_status_based_on_previous_results(metric: str):
...
+def send_manager_benchmark_results_to_argus(argus_client: ArgusClient, result: dict, sut_timestamp: int,
+                                            row_name: str = None):
+    if not row_name:
+        row_name = "#1"
+
+    class RestoreResult(GenericResultTable):
+        class Meta:
+            name = f"Restore benchmark"
+            description = "Restore benchmark"
+            Columns = [ColumnMetadata(name=param, unit="", type=ResultType.FLOAT) for param in result.keys()]
```
fruch, line 184: I would recommend no creating this class in runtime 
and clearly define the columns and their units 

the purpose of this class is that it would be visible and clear what exactly get reported

mikliapko:
> @mikliapko That's nice! Could we add a units to those results? I'm not sure what 120 stands for.

It's seconds. We can add units I believe, forgot about it.
It'd be even better to use minutes.

sdcm/argus_results.py
mikliapko: Fixed

mikliapko:
> > @mikliapko That's nice! Could we add a units to those results? I'm not sure what 120 stands for.
> 
> It's seconds. We can add units I believe, forgot about it. It'd be even better to use minutes.

How it looks in minutes
image

sdcm/argus_results.py
mikliapko: Fixed

soyacz:
> > > @mikliapko That's nice! Could we add a units to those results? I'm not sure what 120 stands for.
> > 
> > 
> > It's seconds. We can add units I believe, forgot about it. It'd be even better to use minutes.
> 

Let's keep in seconds (use DURATION type for column). In future we can think of improving formatting on Argus side to match needs.

mikliapko:
> > > > @mikliapko That's nice! Could we add a units to those results? I'm not sure what 120 stands for.
> > > 
> > > 
> > > It's seconds. We can add units I believe, forgot about it. It'd be even better to use minutes.
> 
> Let's keep in seconds (use DURATION type for column). In future we can think of improving formatting on Argus side to match needs.

Yeah, it'd be great to make it more human-readable, especially for large values in seconds like 5000, for example.
For now, keeping in seconds as you suggested.
Fixed.

sdcm/mgmt/cli.py
```diff
@@ -1302,6 +1323,17 @@ def client_version(self):
...
 
+    @property
+    def client_version_timestamp(self) -> int:
+        """Gets the timestamp of the client version
+
+        Example, for client version `3.3.3-0.20240912.924034e0d` the timestamp is `1726099200` (2024-09-12 00:00:00)
+        """
+        date_str = self.client_version.split('.')[-2]
+        date_obj = datetime.datetime.strptime(date_str, "%Y%m%d")
+        timestamp = int(time.mktime(date_obj.timetuple()))
```
soyacz, line 1334: same can be done with `date_obj.timestamp()`

mikliapko: Fixed

mikliapko:
@soyacz Should we have one more maintainer's review to merge it?

soyacz:
> @soyacz Should we have one more maintainer's review to merge it?

or just someone from QA group

PR #8837, opened 2024-09-25
improvement(results): update Argus Client

soyacz:
Updated Argus Client to support best results tracking and Argus data validation.

Refs: https://github.com/scylladb/qa-tasks/issues/1244

### Testing

- [x] - [perf-simple-querry](https://argus.scylladb.com/workspace?state=WyIyMGM3ZTYzZS1mNWJhLTRjYTItOTQ2YS1lZmVjODliNTc5NDkiXQ)

### Reminders


soyacz:
Actually, there's an error in Argus client, it makes `ValidationRules` mandatory, which wasn't intended.
I think I'm going to fix it first and then come back when done.

soyacz:
waiting for https://github.com/scylladb/argus/pull/463

fruch:
@soyacz can you clarify the status of this one ? if it's for review, what exactly we should be looking at ?

soyacz:
> @soyacz can you clarify the status of this one ? if it's for review, what exactly we should be looking at ?

Yes, it's for review. Upgrading client will add possibility for Argus to automatically mark results as passed/failed if "ValidationRules" are provided.
Because latency calculator/perf simple query regression checker is doing it already, we don't need to add them.
Before merge, I'll modify this PR to include some validation rules for SM testing @mikliapko recently introduced.

sdcm/argus_results.py
```diff
@@ -93,10 +92,15 @@ class Meta:
...
         description = "Restore benchmark"
         Columns = [
-            ColumnMetadata(name="restore time", unit="s", type=ResultType.DURATION),
-            ColumnMetadata(name="repair time", unit="s", type=ResultType.DURATION),
-            ColumnMetadata(name="total", unit="s", type=ResultType.DURATION),
+            ColumnMetadata(name="restore time", unit="s", type=ResultType.DURATION, higher_is_better=False),
+            ColumnMetadata(name="repair time", unit="s", type=ResultType.DURATION, higher_is_better=False),
+            ColumnMetadata(name="total", unit="s", type=ResultType.DURATION, higher_is_better=False),
         ]
+        ValidationRules = {
```
soyacz, line 99: @mikliapko please review if these numbers seem reasonable - 10% longer duration than 'best result' will mark test as failed.

mikliapko: Looks reasonable to me

mikliapko:
Doesn't directly related to this PR, just a general question about this approach - for SM it would be a good improvement to show not only SUT timestamp but SM version as well. Is it feasible to implement?

mikliapko (approved):
What is related to SM part - LGTM

soyacz:
> Doesn't directly related to this PR, just a general question about this approach - for SM it would be a good improvement to show not only SUT timestamp but SM version as well. Is it feasible to implement?

I'd need to understand the use case first.
Varying both Scylla and SM versions makes hard to track regressions - it will never be clear if it comes from SM or from Scylla itself. Think if it wouldn't it be better to pin one.
The other thing, I don't know how to show it on graph?

mikliapko:
> > Doesn't directly related to this PR, just a general question about this approach - for SM it would be a good improvement to show not only SUT timestamp but SM version as well. Is it feasible to implement?
> 
> I'd need to understand the use case first. Varying both Scylla and SM versions makes hard to track regressions - it will never be clear if it comes from SM or from Scylla itself. Think if it wouldn't it be better to pin one. The other thing, I don't know how to show it on graph?

In the graphs that we have for Manager, we use SUT timestamp of Manager. So, there is nothing here related to Scylla SUT.

From my point of view, it'd be nice to have SUT + version to improve readability since in Manager restore benchmark we want to compare with previous releases and it's hard to keep in mind to which release specific SUT date corresponds.

But it's not a "show-stopper", for sure. Just a small improvement that could have made these Manager graphs more readable.

soyacz:
> > > Doesn't directly related to this PR, just a general question about this approach - for SM it would be a good improvement to show not only SUT timestamp but SM version as well. Is it feasible to implement?
> > 
> > 
> > I'd need to understand the use case first. Varying both Scylla and SM versions makes hard to track regressions - it will never be clear if it comes from SM or from Scylla itself. Think if it wouldn't it be better to pin one. The other thing, I don't know how to show it on graph?
> 
> In the graphs that we have for Manager, we use SUT timestamp of Manager. So, there is nothing here related to Scylla SUT.
> 
> From my point of view, it'd be nice to have SUT + version to improve readability since in Manager restore benchmark we want to compare with previous releases and it's hard to keep in mind to which release specific SUT date corresponds.
> 
> But it's not a "show-stopper", for sure. Just a small improvement that could have made these Manager graphs more readable.

There's a plan to introduce comparison between tests. In that case, if each release had own restore benchmark jobs it should be possible to compare across them.
Also, you could use 'rows' for Scylla version - but this shouldn't change often as each row will be represented as separate series on one graph, so we don't want to have 10 lines, after 10 tests (rather points in that case). E.g. if each row would be marked as Scylla version e.g. 2024.1.3 then there would be 2024.1.3 series on graph, after switch to 2024.1.4 another series would appear (breaking connection line between 2024.1.3 series, which is good IMHO clearly showing impact of Scylla version change). Then SUT can be still bound to SM build date, showing impact of SM changes on restore timings.

fruch (approved):
LGTM

juliayakovlev:
@soyacz @fruch 

this PR was not fully merged in `branch-perf-v15`. Here https://github.com/scylladb/scylla-cluster-tests/compare/4d181748d1add3bdec64cba9b19c3c8c9263344b..6ccf64966d3bf263d3452d525df30cda6072c20d import of `ValidationRule` was removed. This caused to the problem after backport of https://github.com/scylladb/scylla-cluster-tests/pull/9295: performance tests failed with error:
```
2024-12-29 07:51:02.610: (DisruptionEvent Severity.ERROR) period_type=end event_id=62ab9ec3-05c9-4a24-8b6a-5dec90fbfd8a duration=31m2s: nemesis_name=RunUniqueSequence target_node=Node perf-latency-nemesis-ubuntu-db-node-6aa2a1b1-3 [3.239.110.219 | 10.12.1.252] errors=name 'ValidationRule' is not defined
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5222, in wrapper
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4366, in disrupt_run_unique_sequence
self.steady_state_latency()
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 263, in wrapped
send_result_to_argus(
File "/home/ubuntu/scylla-cluster-tests/sdcm/argus_results.py", line 125, in send_result_to_argus
result_table.validation_rules = {metric: ValidationRule(**rules) for metric, rules in error_thresholds.items()}
File "/home/ubuntu/scylla-cluster-tests/sdcm/argus_results.py", line 125, in <dictcomp>
result_table.validation_rules = {metric: ValidationRule(**rules) for metric, rules in error_thresholds.items()}
NameError: name 'ValidationRule' is not defined
```

Example of failed test:


https://argus.scylladb.com/tests/scylla-cluster-tests/6aa2a1b1-5653-42ab-8922-3563513236ef

soyacz:
fix here: https://github.com/scylladb/scylla-cluster-tests/pull/9624

Issue #8144, opened 2024-07-25
`decommission_streaming_err` nemesis times out too early when some `END-around` log message is awaited before rebooting target DB node


vponomaryov:
### Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Running the `disrupt_decommission_streaming_err` nemesis SCT picks up one of the DB log messages to be awaited before rebooting the target node.

And when some `END`-around command gets picked up like the following:

```
2024-07-24 19:52:55,353 f:nemesis.py      l:3873 c:sdcm.nemesis         p:DEBUG > sdcm.nemesis.DecommissionStreamingErrMonkey: Reboot node after log message: 'Finished token ring movement'
```

Then it gets assigned to short time out (`80m`) and fails because of it.

The problem is that the decommission was going on ok, just required more time to reach required step.
The target node is yellow on the screenshot below:

![Screenshot from 2024-07-24 17-36-38](https://github.com/user-attachments/assets/0dafdbf2-0879-40d6-8d25-f0be6eb5866d)

#### Steps to Reproduce

1. Run `longevity-multi-keyspaces-with-tablets` Ci job with the `disrupt_decommission_streaming_err` nemesis
2. See error

**Expected behavior:** timeout value must be more closer to the real life.

**Actual behavior:** timeout is too small

### How frequently does it reproduce?

100%

### Installation details

SCT Version: master
Scylla version (or git commit hash): master

### Logs
* test_id: 7d0040da-769a-4e8c-bc93-5566e70b51e3
* job log: [scylla-staging/valerii/vp-longevity-multi-keyspaces-with-tablets#4](https://argus.scylladb.com/test/7fd0596d-6ee0-4ecb-ad9f-43b168205f4e/runs?additionalRuns[]=7d0040da-769a-4e8c-bc93-5566e70b51e3) , [CI](https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-multi-keyspaces-with-tablets/4)

soyacz:
Is this test passing without tablets?
For me it looks like a very slow decommission.
And the question, is `Finished token ring movement` log message valid for tablets?

vponomaryov:
> Is this test passing without tablets? 

I didn't run it without tablets

> For me it looks like a very slow decommission. 

Cluster state directly influences the speed of decommission.
Data size, CPU load, disk load and so on...

[...]

PR #8843, opened 2024-09-25
Feature(TopologyOpsMonitor): Monitor decommission topology operation

aleksbykov:
Decommission operation could be terminated for different reason, but decommission process will be run on node and could successfully be decommissioned. But because of nemesis is terminated by exception cluster health validator could abort whole test run because node status is Decommissioning. This start happened with nemesis DecommissionStreamingErr, when log message could not be found and process is aborted by timeout, while decommission continue to run.

To Catch such case FailedDecommissionOperationMonitoring is presented It is ContextManager which could be used to safely run decommission operation and check node status and wait decommission will be finished if command aborted or terminated.

Fix: #8144

### Testing

- [Job passed](https://argus.scylladb.com/test/b3050635-9416-40a0-86ea-7f4b2dc3ad33/runs?additionalRuns[]=5513d3d1-769b-48f1-a5ba-70cd84fbb89c)

### Reminders


fruch (approved):
LGTM

aleksbykov:
@soyacz , can you take a look?

sdcm/utils/topology_ops.py
```diff
@@ -0,0 +1,99 @@
...
+    process still running, wait while it be finished and check operation status
+    """
+
+    def __init__(self, target_node: "BaseNode", verification_node: "BaseNode",
+                 timeout: int | float = 7200, expected_exception: tuple[Exception | BaseException, ...] | None = None):
+        self.timeout = timeout
+        self.target_node = target_node
+        self.db_cluster: "BaseScyllaCluster" = target_node.parent_cluster
+        self.target_node_ip = target_node.ip_address
+        expected_exception = expected_exception or set()
```
soyacz, line 43: do we need that? From my understanding it is used to skip waiting for finish node decommission if test ended (KillNemesis exception raised).
At least rename it to something that reflects that (`FailedDecommissionOperationMonitoring` that expects `KillNemesis` is weird)

sdcm/utils/topology_ops.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def __enter__(self):
+        return self
+
+    def __exit__(self, exc_type, exc_val, exc_tb):
+        if exc_type:
+            LOGGER.warning("Decommission failed with error: %s", traceback.format_exception(exc_type, exc_val, exc_tb))
+            decommission_in_progress = self.is_node_decommissioning()
+            if not decommission_in_progress:
+                self.db_cluster.verify_decommission(self.target_node)
```
soyacz, line 55: `verify_decommission` terminates the node and does bunch of sttuff - how long it takes? Maybe we could also  skip that in case of KillNemesis exception?

aleksbykov: I set skip any checks if KillNemesis was raised

sdcm/utils/topology_ops.py
aleksbykov: i just want to avoid annoying error event with kill nemesis, Remove that part of code

aleksbykov:
Fixed according comments. Running staging job:
https://jenkins.scylladb.com/job/scylla-staging/job/abykov/job/longevity-100gb-4h-decommission_streamin_err_8144/25

sdcm/utils/topology_ops.py
soyacz: If there's annoying error then it's a good reason to add it - my intention was to change the naming, as we don't expect any errors, rather ignoring them (or workaround possible errors due to fact of end of the test)

soyacz (approved):
LGTM

aleksbykov:
@fruch Can you take a look and approve one more time

fruch (approved):

Issue #7860, opened 2024-07-04
Add more linting rules to ruff


fruch:
A tab of the rules/set we want to activate 

### Tasks
- [ ] https://docs.astral.sh/ruff/rules/#isort-i - worth doing, it would align things, and can avoid some merge conflicts
- [ ] https://docs.astral.sh/ruff/rules/#flake8-datetimez-dtz - this can help avoid lot's of time offset issues, and also future proof, like some api being depreciated - we have 147 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-use-pathlib-pth - we have 830 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-logging-log - make sure we have proper name for out loggers, we have 4 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-unused-arguments-arg - we have 321 issue with that
- [x] https://docs.astral.sh/ruff/rules/undefined-name/ - F821
- [x] https://docs.astral.sh/ruff/rules/undefined-local/ - F823
- [x] https://docs.astral.sh/ruff/rules/f-string-missing-placeholders/#f-string-missing-placeholders-f541
- [ ] https://docs.astral.sh/ruff/rules/private-member-access/#private-member-access-slf001 - we have lots of those, not sure how critical it would be to fix or ignore those
- [ ] https://docs.astral.sh/ruff/rules/missing-f-string-syntax/ - we have 2 of those
- [ ] https://docs.astral.sh/ruff/rules/explicit-f-string-type-conversion/ - we have 45 of those
- [ ] https://docs.astral.sh/ruff/rules/implicit-optional/ - we have 284 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-builtins-a -  we have 7 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-logging-log - we have 4 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-logging-format-g - we have 430 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-pie-pie - we have 62 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-print-t20 - we have 167 of those
- [ ] https://docs.astral.sh/ruff/rules/#flake8-pytest-style-pt - we have ~1500 of those

PR #8852, opened 2024-09-26
improvement(ruff): enable `F821` and `F823` rules

vponomaryov:
Enable following rules:
- F821, `undefined-name` -> `Undefined name {name}. {tip}`
- F823, `undefined-local` -> `Local variable {name} referenced before assignment`

### Testing

- [ ]

### Reminders

### DIFF
--- a/multiple_dc_test.py
+++ b/multiple_dc_test.py
@@ -15,6 +15,7 @@
 
 # TODO: this test seem to be broken, hence disabling all pylint checks
 # pylint: disable=all
+# ruff: noqa: F821
 
 from sdcm.tester import ClusterTester
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,5 +1,10 @@
 [tool.ruff]
-lint.select = ["PL", "PLR0916", "PLR0913","PLR0914", "YTT", "BLE", "F841", "F401"]
+lint.select = [
+    "BLE",
+    "F401", "F821", "F823", "F841",
+    "PL", "PLR0913","PLR0914", "PLR0916",
+    "YTT",
+]
 
 lint.ignore = ["E501", "PLR2004"]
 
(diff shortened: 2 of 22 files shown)


vponomaryov:
I wasted some time catching missing imports in runtime instead of the pre-commit checks.

Probably we want to add more, but let's add these first.

fruch:
> I wasted some time catching missing imports in runtime instead of the pre-commit checks.
> 
> Probably we want to add more, but let's add these first.

I had a list of one it might worth enabling, let's list this there:
https://github.com/scylladb/scylla-cluster-tests/issues/7860

vponomaryov:
> > I wasted some time catching missing imports in runtime instead of the pre-commit checks.
> > Probably we want to add more, but let's add these first.
> 
> I had a list of one it might worth enabling, let's list this there: #7860

Added.

soyacz (approved):
LGTM

soyacz:
I wonder if we shouldn't use git feature that can ignore specific commits in git blame: https://graphite.dev/guides/git-blame-ignore-revs
WDYT?

fruch (approved):
LGTM

vponomaryov:
> I wonder if we shouldn't use git feature that can ignore specific commits in git blame: https://graphite.dev/guides/git-blame-ignore-revs WDYT?

How is this related to this change?

soyacz:
generally, introducing changes just to make linting happy is hiding more important history from git blame. I think we could benefit from it with these kind of PR's.

fruch:
> > I wonder if we shouldn't use git feature that can ignore specific commits in git blame: https://graphite.dev/guides/git-blame-ignore-revs WDYT?
> 
> How is this related to this change?

I think @soyacz suggests that we'll mark such a change that touches multiple places cause of linting as one that won't be shown in git blame,

I think we could experiment with it, it can ease a bit of investigation work around those touched places in the code.

It's nice to have, not more than that.

fruch:
we got bit by not having this on v15, trying to backport

PR #7485, opened 2024-05-26
fix(nodebootstrapabortmanager): always clean scylla data before rebootstrap

aleksbykov:
With raft topology, node which failed to bootstrap , removed from cluster
and banned. It could be added to cluster but it need to clean all scylla data before rebootstrap.
Depending on log message after which bootstrap is going to be aborted,
Bootstrap process could be finished ealier than command for abort will be executed
and node could be added to cluster.
New:
 - Added additinal checks whether bootstrap finished successfull  before it would be aborted
 - Added additional changeseverity filters for error messages expected to appear during
bootstrap aborting process
 - Added decommission operation monitor, which allow to check whether decommission is running
after command failed for any reason and wait process to be finished.

### Testing
- [Test passed with single nemesis](https://jenkins.scylladb.com/job/scylla-staging/job/abykov/job/longevity-100gb-4h-bootstrap-test/103/)
- [Regular job](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h-bootstrap-test/110)
- [Regular jobs with single nemesis](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/abykov/job/longevity-100gb-4h/116)
- [Another test](https://jenkins.scylladb.com/job/scylla-staging/job/abykov/job/longevity-100gb-4h-bootstrap-test/109/)

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #8867, opened 2024-09-29
fix(RAFT) fix wrong raft.is_enabled usage

temichus:
the whole idea of putting RAFT into separate module is to remove hundreds of spaghetti lines like:
    if not current_node.raft.is_enabled

and put Raft helper functions under
node.raft.<HELPER> to avoid hundreds of helpers imports

it can be exceptions in case of using raft.is_enabled as part of a complex 'if' when code logic depends not only on the Raft feature

### Testing

- [x] 5.4.9 (no tablets) https://argus.scylladb.com/tests/scylla-cluster-tests/73916f48-5483-4718-baa0-70388edf66b1

```
< t:2024-10-23 10:26:53,705 f:__init__.py     l:293  c:sdcm.utils.raft      p:DEBUG > Check group0 and token ring consistency on node longevity-10gb-3h-wait-for-db-node-73916f48-1 (host_id=01d35667-0e45-465c-81f1-eaefd772a168)...
```
```
< t:2024-10-23 10:36:55,202 f:common.py       l:36   c:sdcm.utils.tablets.common p:INFO  > Tablets are disabled, skipping wait for balance
```
- [x] master:latest https://argus.scylladb.com/tests/scylla-cluster-tests/8fb46f18-2711-4d92-b5be-c432ddaa30e8
```
< t:2024-10-23 10:49:06,796 f:__init__.py     l:310  c:sdcm.utils.raft      p:DEBUG > Group0 and token-ring are consistent on node longevity-10gb-3h-wait-for-db-node-8fb46f18-1 (host_id=f6b1fdca-29f8-403c-8d0d-f57b6a12432f)...
```
```
< t:2024-10-23 10:57:25,589 f:common.py       l:45   c:sdcm.utils.tablets.common p:INFO  > Tablets are balanced
```

### Reminders

### DIFF
--- a/sdcm/utils/health_checker.py
+++ b/sdcm/utils/health_checker.py
@@ -268,25 +268,4 @@ def check_schema_agreement_in_gossip_and_peers(node, retries: int = CHECK_NODE_H
 def check_group0_tokenring_consistency(group0_members: list[dict[str, str]],
                                        tokenring_members: list[dict[str, str]],
                                        current_node) -> HealthEventsGenerator:
-    if not current_node.raft.is_enabled:
-        LOGGER.debug("Raft feature is disabled on node %s (host_id=%s)", current_node.name, current_node.host_id)
-        return
-    LOGGER.debug("Check group0 and token ring consistency on node %s (host_id=%s)...",
-                 current_node.name, current_node.host_id)
-    token_ring_node_ids = [member["host_id"] for member in tokenring_members]
-    for member in group0_members:
-        if member["voter"] and member["host_id"] in token_ring_node_ids:
-            continue
-        error_message = f"Node {current_node.name} has group0 member with host_id {member['host_id']} with " \
-            f"can_vote {member['voter']} and " \
-            f"presents in token ring {member['host_id'] in token_ring_node_ids}. " \
-            f"Inconsistency between group0: {group0_members} " \
-            f"and tokenring: {tokenring_members}"
-        LOGGER.error(error_message)
-        yield ClusterHealthValidatorEvent.Group0TokenRingInconsistency(
-            severity=Severity.ERROR,
-            node=current_node.name,
-            error=error_message,
-        )
-    LOGGER.debug("Group0 and token-ring are consistent on node %s (host_id=%s)...",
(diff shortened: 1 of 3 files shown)


temichus:
waiting for https://github.com/scylladb/scylla-cluster-tests/pull/7485 to rebase 
and want to get agreement for this change from @fruch and @soyacz before testing this change

soyacz (commented):
I understand the need for `check_group0_tokenring_consistency` which is raft related, but moving `wait_for_tablets_balanced` under raft is not intuitive. Just because it requires raft I don't see the reason to move it out of 'tablets' module. Can you elaborate what are benefits here?

temichus:
> I understand the need for `check_group0_tokenring_consistency` which is raft related, but moving `wait_for_tablets_balanced` under raft is not intuitive. Just because it requires raft I don't see the reason to move it out of 'tablets' module. Can you elaborate what are benefits here?

same as in description 
1) avoid hundreds of spaghetti lines like:
       if not current_node.raft.is_enabled
             return
      ............
      
2)   put Raft helper functions under
node.raft. to avoid hundreds of helpers imports


but agree that it is not intuitive in some cases but if we are talking that all different behaviours/algorithms between RAFT and Gossip are located under node.raft. . In this case, tablets algorithms that depends on raft.is_enabled should be under the raft

fruch:
`wait_for_tablets_balanced` doesn't have anything to do with raft, it checks if tablets are enabled and it can't be enabled without raft.

It was a mistake, we should remove the raft check from that function.

soyacz:
> `wait_for_tablets_balanced` doesn't have anything to do with raft, it checks if tablets are enabled and it can't be enabled without raft.
> 
> It was a mistake, we should remove the raft check from that function.

it was added as first verification before querying db for 'tablets enabled' feature. Possibly can be removed to avoid 'sphagetti lines' and keep intuitive.

temichus:
@aleksbykov @soyacz 
I've updates pr according to comments at set it ready for review

soyacz:
@temichus please add backport labels

aleksbykov (approved):
LGTM

Issue #8693, opened 2024-09-13
"Parent connection is aborting on shard" should not be raised as error event


soyacz:
As per discussion in https://github.com/scylladb/scylla-enterprise/issues/4691#issuecomment-2348867638, errors like:
```
Sep 05 19:19:50.107534 longevity-mv-si-4d-2024-2-db-node-608c20c6-5 scylla[7237]:  [shard 24:strm] rpc - client 10.12.11.11:50184: server stream connection dropped: Unknown parent connection 39d40018 on shard 24
Sep 05 19:19:50.115196 longevity-mv-si-4d-2024-2-db-node-608c20c6-5 scylla[7237]:  [shard 23:strm] rpc - client 10.12.11.11:49613: server stream connection dropped: Parent connection 38350017 is aborting on shard 23
Sep 05 19:19:50.115232 longevity-mv-si-4d-2024-2-db-node-608c20c6-5 scylla[7237]:  [shard 23:strm] rpc - client 10.12.11.11:52973: server stream connection dropped: Parent connection 38350017 is aborting on shard 23
```
should not be raised as ErrorEvent.

kbr-scylla:
@gleb-cloudius please give your opinion -- doesn't this signify internal problem with Seastar RPC implementation? Especially the "Unknown parent connection" thing looks like it should never happen

(If yes, we could still treat it as error event, just a different one from the original "aborting on shard")

gleb-cloudius:
Looks like both of those errors are results of the same race. "Parent connection is aborting" happen when a connection we are opening a stream on is closing and "Unknown parent connection" happens when the parent connection was already closed. Both can happen and both are benign. Stream opening should fail.

kbr-scylla:
Ok, so this is no reason for error event.

fruch:
@dimakr please attend to this one, i.e. filtering those in `enable_default_filters`

PR #8884, opened 2024-09-30
fix(sct_events): ignore 'aborting on shard' events on rpc errors

dimakr:
Closes: https://github.com/scylladb/scylla-cluster-tests/issues/8693

### Testing
- [ ]

### Reminders

### DIFF
--- a/sdcm/sct_events/setup.py
+++ b/sdcm/sct_events/setup.py
@@ -144,5 +144,12 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
     DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line='Rate-limit: suppressed').publish()
     DbEventsFilter(db_event=DatabaseLogEvent.WARNING, line='abort_requested_exception').publish()
 
+    # As per discussion in https://github.com/scylladb/scylla-enterprise/issues/4691#issuecomment-2348867638 and
+    # https://github.com/scylladb/scylla-cluster-tests/issues/8693#issuecomment-2358147285
+    # the 'aborting on shard' error is acceptable when RPC connections are dropped
+    EventsSeverityChangerFilter(new_severity=Severity.WARNING,
+                                event_class=DatabaseLogEvent.ABORTING_ON_SHARD,
+                                regex=r'.*Parent connection [\d]+ is aborting on shard').publish()
+
 
 __all__ = ("start_events_device", "stop_events_device", "enable_default_filters")


dimakr:
@fruch added backports to 2024.2 and 2024.1 branches, as the `longevity-mv-si-4days` test from the original issue is regularly executed as a part of Tier1 longevities there.
But was not adding, for now, backports to oss branches, as the test is not executed there.

sdcm/sct_events/setup.py
```diff
@@ -144,5 +144,10 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
     DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line='Rate-limit: suppressed').publish()
     DbEventsFilter(db_event=DatabaseLogEvent.WARNING, line='abort_requested_exception').publish()
 
+    if SkipPerIssues('scylladb/scylla-enterprise#4691', params=sct_config):
```
fruch, line 147: I'm not sure it should be based on this issue, if it agreed there it's acceptable, the issue would be closed.

And it doesn't seem like there's an agreement there to ever fix this error print.

So for now let's leave a reference to the discussion in a comment in the code.

dimakr: Updated

sdcm/sct_events/setup.py
```diff
@@ -144,5 +144,12 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
     DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line='Rate-limit: suppressed').publish()
     DbEventsFilter(db_event=DatabaseLogEvent.WARNING, line='abort_requested_exception').publish()
 
+    # As per discussion in https://github.com/scylladb/scylla-enterprise/issues/4691#issuecomment-2348867638 and
+    # https://github.com/scylladb/scylla-cluster-tests/issues/8693#issuecomment-2358147285
+    # the 'aborting on shard' error is acceptable when RPC connections are dropped
+    EventsSeverityChangerFilter(new_severity=Severity.WARNING,
+                                event_class=DatabaseLogEvent.ABORTING_ON_SHARD,
+                                regex=r'.*rpc - client.*Parent connection.*is aborting on shard.*').publish()
```
soyacz, line 152: I think it's a bit too expensive for CPU,
Maybe just `Parent connection [\d]* is aborting on shard` would be enough?

sdcm/sct_events/setup.py
```diff
@@ -144,5 +144,12 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
     DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line='Rate-limit: suppressed').publish()
     DbEventsFilter(db_event=DatabaseLogEvent.WARNING, line='abort_requested_exception').publish()
 
+    # As per discussion in https://github.com/scylladb/scylla-enterprise/issues/4691#issuecomment-2348867638 and
+    # https://github.com/scylladb/scylla-cluster-tests/issues/8693#issuecomment-2358147285
+    # the 'aborting on shard' error is acceptable when RPC connections are dropped
+    EventsSeverityChangerFilter(new_severity=Severity.WARNING,
```
soyacz, line 150: instead of `EventsSeverityChangerFilter` maybe we could just add another event type that is verified before `aborting on shard` one?
@fruch WDYT?

fruch: It's quite equivalent, in both cases it's running one more regex per line

The configuration like here is a bit more clear to my taste, adding a new type in the right place is a bit more tricky to do correctly.

soyacz: still, I think we could improve regex, as current one seems expensive

sdcm/sct_events/setup.py
dimakr: Updated the regex

sdcm/sct_events/setup.py
```diff
@@ -144,5 +144,12 @@ def enable_default_filters(sct_config: SCTConfiguration):  # pylint: disable=unu
...
+                                event_class=DatabaseLogEvent.ABORTING_ON_SHARD,
+                                regex=r'.*Parent connection [\d]+ is aborting on shard').publish()
```
soyacz, line 152: I'm not sure if initial .* is required, 
won't it work just with e.g. `server stream connection dropped: Parent connection [\d]+ is aborting on shard`?

dimakr: EventsFilter.eval_filter uses re.match to eval the event.

Still, I double checked - only with initial `.*` the needed error will be matched.

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #8886, opened 2024-09-30
improvement(gemini): run gemini from scylladb/gemini

CodeLieutenant:
Dockerfile has been moved to gemini project, and now SCT should use that image.

Changes:
- Use image from scylladb/gemini
- Remove dockerfile for gemini and point in readme location for the images
- Add CQL Statement Logging to gemini output
- Forward outputs from docker to $HOME/*.log
- Run default gemini flags from gemini_thread.py

### Testing
- :red_circle: https://jenkins.scylladb.com/job/scylla-staging/job/dusan/job/gemini/job/gemini-1tb-10h-test/


sdcm/gemini_thread.py
```diff
@@ -148,6 +183,13 @@ def _run_stress(self, loader, loader_idx, cpu_idx):
             results_copied = docker.receive_files(src=self.gemini_result_file, dst=local_gemini_result_file)
             assert results_copied, "gemini results aren't available, did gemini even run ?"
 
+            local_gemini_test_statements_file = os.path.join(
```
fruch, line 186: @CodeLieutenant 

I've added this into your PR, copying those files back

CodeLieutenant: I'm not sure why we need this copy, as these files are mounted in docker container, we should have them inside $HOME

fruch: it's on the loader machine $HOME, not where the SCT test is running.

we are collecting the logs from where the test is running in this case, we can optimize it later.
but now for simplicity it's the fastest way to collect it as as the gemini_results fail is being collected.

Issue #8710, opened 2024-09-16
Add retrying to download repo files


soyacz:
Sometimes SCT fails to download SM repo file and fails the test. Idea is to add retries to curl command:
```
Command: 'sudo curl -o /etc/apt/sources.list.d/scylla-manager.list -L https://downloads.scylladb.com/deb/debian/scylladb-manager-3.3.list'
Exit code: 28
Stdout:
Stderr:
0     0    0     0    0     0      0      0 --:--:--  0:04:52 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:04:53 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:04:54 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:04:55 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:04:56 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:04:57 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:04:58 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:04:59 --:--:--     0
0     0    0     0    0     0      0      0 --:--:--  0:05:00 --:--:--     0
curl: (28) SSL connection timeout
```

## Impact

Failed test

## Installation details

Cluster size: 12 nodes (i3en.2xlarge)

Scylla Nodes used in this run:
- parallel-topology-schema-changes-mu-db-node-98698798-9 (18.170.50.54 | 2a05:d01c:0964:7d02:e164:c482:366b:7f1f) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-98698798-8 (18.175.245.228 | 2a05:d01c:0964:7d02:5af2:4b11:ed54:5e80) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-98698798-7 (18.170.57.14 | 2a05:d01c:0964:7d02:9636:2468:0ec9:acdf) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-98698798-6 (3.252.51.114 | 2a05:d018:12e3:f002:2457:c43a:7b1d:2540) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-98698798-5 (34.244.67.233 | 2a05:d018:12e3:f002:453a:6e4c:6afd:2a71) (shards: 7)
[...]

PR #8890, opened 2024-10-01
fix(setup): add retrying to download sm repo file

soyacz:
Sometimes SM repo file download fails due various reasons.

Adding retrying as other steps that interact with network.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/8710

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -1804,7 +1804,7 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
             repo_path = '/etc/zypp/repos.d/scylla-manager.repo'
         else:
             repo_path = '/etc/apt/sources.list.d/scylla-manager.list'
-        self.remoter.sudo(f"curl -o {repo_path} -L {scylla_repo}")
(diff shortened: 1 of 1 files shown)


mikliapko (dismissed):
LGTM

fruch (approved):
LGTM

sdcm/cluster.py
```diff
@@ -1804,7 +1804,7 @@ def download_scylla_manager_repo(self, scylla_repo: str) -> None:
             repo_path = '/etc/zypp/repos.d/scylla-manager.repo'
         else:
             repo_path = '/etc/apt/sources.list.d/scylla-manager.list'
-        self.remoter.sudo(f"curl -o {repo_path} -L {scylla_repo}")
+        self.remoter.sudo(f"curl -o {repo_path} -L {scylla_repo}", retry=3)
```
fruch, line 1807: one might also use curl argument for retrying, they might work better then our own retry :\

fruch: we have few places in the code we are doing:
```
--retry 5 --retry-max-time 300
```

soyacz: true, let's use that

PR #8912, opened 2024-10-03
fix(cleanup): cleanup sct-runner when logs collected

soyacz:
When test fail, often sct-runners are kept for long - especially in perf tests. This is because of broken/unclear logic behind setting sct-runner keep flags.

Fixed that logic to:
- when test timeouted and logs collected, keep sct-runner for another 6 h
- if logs were collected, regardless of test status, terminate sct-runner immediately
- otherwise keep for additional 48h

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/sct_runner.py
+++ b/sdcm/sct_runner.py
@@ -1261,16 +1261,8 @@ def _manage_runner_keep_tag_value(utc_now: datetime,
                                   dry_run: bool = False) -> SctRunnerInfo:
     LOGGER.info("Managing runner's tags. Timeout flag: %s, logs_collected: %s, dry_run: %s",
                 timeout_flag, sct_runner_info.logs_collected, dry_run)
-
-    if test_status == "SUCCESS" and sct_runner_info.logs_collected:
-        if not dry_run:
-            sct_runner_info.sct_runner_class.set_tags(sct_runner_info, {"keep": "0", "keep-action": "terminate"})
-        sct_runner_info.keep = 0
-        sct_runner_info.keep_action = "terminate"
-        return sct_runner_info
-
-    if not timeout_flag and sct_runner_info.logs_collected:
-        current_run_time_hrs = int((utc_now - sct_runner_info.launch_time).total_seconds() // 3600)
+    current_run_time_hrs = int((utc_now - sct_runner_info.launch_time).total_seconds() // 3600)
+    if timeout_flag and sct_runner_info.logs_collected:
         new_keep_value = int(sct_runner_info.keep) - current_run_time_hrs + 6
 
         if new_keep_value > 0:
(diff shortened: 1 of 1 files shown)


soyacz:
this is example test, where sct-runner was set to be kept for 64h, while logs were collected and no reason to keep it: https://jenkins.scylladb.com/job/scylla-staging/job/lukasz/job/manager-restore-benchmark/3/

soyacz:
I'm not sure if this logic is right, and where should be backported.

sdcm/sct_runner.py
```diff
@@ -1279,7 +1271,18 @@ def _manage_runner_keep_tag_value(utc_now: datetime,
             sct_runner_info.keep = new_keep_value
         return sct_runner_info
 
-    LOGGER.info("No changes to make to runner tags.")
+    if sct_runner_info.logs_collected:
+        if not dry_run:
+            sct_runner_info.sct_runner_class.set_tags(sct_runner_info, {"keep": "0", "keep-action": "terminate"})
```
fruch, line 1276: this code is running on this runner ? isn't there a possibility it would get cleared right away ?
maybe 1 is a safer option ?

why in perf tests are not getting into the first if ?

soyacz: that's the intention - clear right away (it's the last step in the pipeline, part of `clean_sct_runners` method) if logs are collected.

Perf tests often fail with an error event, and not getting into the first if. The second seemed to have broken logic: `if not timeout_flag` - meaning only if not timed out (basically always - our tests rarely timeout recently) we prolong by another 6 hours which makes no sense.

Two changes:
1. invert logic behind timeout_flag - when test timed out, add another 6h just in case some more info are required to be obtained from runner (e.g. see which process frozen).
2. Logic was changed to simpler one: if logs are collected, terminate runner, otherwise prolong duration by 48h.

soyacz:
basically this PR is about cleaning sct-runners if logs are collected right after the test. Do we want that behavior?

fruch:
> basically this PR is about cleaning sct-runners if logs are collected right after the test. Do we want that behavior?

we we have a way during the tests to prevent this cleanup ? 

if the machine would be manually marked as keep, would it still gonna be deleted ?

i.e. we need to have some escape hatch if during the test we decided we want to keep the resources.

now we have a few hours after the end of the test to keep it.

soyacz:
> > basically this PR is about cleaning sct-runners if logs are collected right after the test. Do we want that behavior?
> 
> we we have a way during the tests to prevent this cleanup ?
> 
> if the machine would be manually marked as keep, would it still gonna be deleted ?

Yes, but from my understanding it's going to be removed also with current code if test didn't timeout (usual case).
Possibly we should change this logic and keep the 'keep: alive' tag if present (and add it if db nodes are marked as keep? E.g. in reuse cluster case)

> 
> i.e. we need to have some escape hatch if during the test we decided we want to keep the resources.
> 
> now we have a few hours after the end of the test to keep it.

yes. That's the question about this PR, whether we need (and why) to keep SCT runner after the test.

fruch (approved):
LGTM

as a followup we'll document the process of keep resources  during the run (and validate it's working as expected)

soyacz:
@fruch I retriggered checks and they passed.
Also created issue about preventing sct-runner cleanup: https://github.com/scylladb/scylla-cluster-tests/issues/9071

PR #8920, opened 2024-10-06
improvement(results): store cycle's 'start time'

soyacz:
### Testing

- [x] - [example result](https://argus.scylladb.com/test/47ee4c1d-bff9-42fc-8911-29bc79f2ef40/runs?additionalRuns[]=9684194c-1815-4dd9-a717-f82300d001cc) - see final solution moved 'start time' column before screenshot links.

### Reminders

### DIFF
--- a/sdcm/argus_results.py
+++ b/sdcm/argus_results.py
@@ -11,6 +11,8 @@
 #
 # Copyright (c) 2024 ScyllaDB
 import json
+import time
+from datetime import timezone, datetime
 
 from argus.client import ArgusClient
 from argus.client.generic_result import GenericResultTable, ColumnMetadata, ResultType, Status, ValidationRule
@@ -41,8 +43,10 @@ class LatencyCalculatorMixedResult(GenericResultTable):
             ColumnMetadata(name="Throughput write", unit="op/s", type=ResultType.INTEGER, higher_is_better=True),
             ColumnMetadata(name="Throughput read", unit="op/s", type=ResultType.INTEGER, higher_is_better=True),
             ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION, higher_is_better=False),
+            # help jump into proper place in logs/monitoring
+            ColumnMetadata(name="start time", unit="", type=ResultType.TEXT),
             ColumnMetadata(name="Overview", unit="", type=ResultType.TEXT),
-            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT)
+            ColumnMetadata(name="QA dashboard", unit="", type=ResultType.TEXT),
         ]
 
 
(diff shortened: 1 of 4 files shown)


juliayakovlev (dismissed):
LGTM
What about end time - do we want it?

soyacz:
> LGTM What about end time - do we want it?

there's duration shown, would be bit redundant I think

sdcm/argus_results.py
```diff
@@ -41,8 +43,9 @@ class Meta:
             ColumnMetadata(name="Throughput write", unit="op/s", type=ResultType.INTEGER, higher_is_better=True),
             ColumnMetadata(name="Throughput read", unit="op/s", type=ResultType.INTEGER, higher_is_better=True),
             ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION, higher_is_better=False),
+            ColumnMetadata(name="start time", unit="", type=ResultType.TEXT),
```
fruch, line 46: Why don't we want this as a number ?

Do we assume we'll never plot or sort by this one ?

utils/migrate_latency_results_to_argus.py
```diff
@@ -81,13 +81,17 @@ def sort_func(item):
...
                     send_result_to_argus(argus_client=client, workload=workload, name=operation,
-                                         description=description, cycle=0, result=result)
+                                         description=description, cycle=0, result=result, start_time=start_time)
                 except argus.client.base.ArgusClientError:
                     print(
                         f"Failed to send {operation} - {workload} - latencies to Argus: {hit['_source']['test_details']['job_url']}")
                 continue
             for idx, cycle in enumerate(latency_during_ops[operation]["cycles"], start=1):
+                start_time = cycle["hdr_summary"].get("READ", cycle["hdr_summary"].get("WRITE"))["start_time"]
+                start_time = start_time if start_time < 1000000000000 else start_time / 1000
```
fruch, line 94: What is this magic number?

soyacz: for unknown reasons, sometimes we get timestamp in milliseconds and sometimes in seconds

sdcm/argus_results.py
soyacz: no, it just serves investigation purposes to quickly get to proper log place (or looking at monitoring).

fruch: maybe we should document this at some place in the code ?

utils/migrate_latency_results_to_argus.py
soyacz: one saved in ES - usually Steady State (but i think I spotted also in some other cases - hard to tell).

sdcm/argus_results.py
soyacz: I added a comment line. Also found that rebasing have 'eaten' changes in other tables - fixed.

fruch (approved):

Issue #5045, opened 2022-07-21
show-monitor default time range


soyacz:
when recovering monitoring with `hydra show-monitor`, grafana dashboard is set to recent 30 mins and obviously shows nothing for that range. It's cumbersome to set proper time range and to navigate with it.
There should be possibility to make default time range that fits test time range. First google search showed: https://community.grafana.com/t/default-time-range-for-dashboards/13937
There should be some setting that we can automatically set.

fruch:
question is when to put those default into the configuration file of the dashboard ? when we collect the, or during the show-monitor command itself.

few more issues related:
* change default dashboard to the nemesis one (our own)
* change the default event filter, to show ERROR and CRITICAL, not WARNING and NORMAL

soyacz:
> question is when to put those default into the configuration file of the dashboard ? when we collect the, or during the show-monitor command itself.
> 
> few more issues related:
> 
> * change default dashboard to the nemesis one (our own)
> * change the default event filter, to show ERROR and CRITICAL, not WARNING and NORMAL

I propose doing it in separate task.

yaronkaikov:
@soyacz This is breaking docker artifact tests for `master`, `enterprise`, and `6.2` 
@lsfreitas @Annamikhlin FYI

fruch:
@soyacz `update_default_time_range` should be noop, for cases we don't have monitor setup...

soyacz:
> [@soyacz](https://github.com/soyacz) `update_default_time_range` should be noop, for cases we don't have monitor setup...

I tested it without monitor, but docker backend requires nodes to get monitor install path - I missed that case.

PR #8921, opened 2024-10-06
improvement(monitoring): restore monitor with test time range

soyacz:
Restoring monitor data with `hydra investigate show-monitor` sets default time range to recent 30m. This is unconvinient as requires to find test related time range manually.

This change hardcodes time range in dashboard defaults to range of test duration.

closes: https://github.com/scylladb/scylla-cluster-tests/issues/5045
closes: https://github.com/scylladb/argus/issues/454

### Testing

- [x] - provision test - try restoring `hydra investigate show-monitor 81aa1353-e2c5-4ac3-b864-ca4b0f81dede`

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -33,7 +33,7 @@ import json
 import ipaddress
 from importlib import import_module
 from typing import List, Optional, Dict, Union, Set, Iterable, ContextManager, Any, IO, AnyStr, Callable
-from datetime import datetime
+from datetime import datetime, timezone
 from textwrap import dedent
 from functools import cached_property, wraps, lru_cache, partial
 from collections import defaultdict
@@ -5854,6 +5854,8 @@ class BaseMonitorSet:  # pylint: disable=too-many-public-methods,too-many-instan
         self.save_sct_dashboards_config(node)
         self.save_monitoring_version(node)
         Path(self.sct_dashboard_json_file).unlink(missing_ok=True)
+        end_time = self.params["test_duration"] * 60 + time.time() + 120 * 60  # 2h margin
+        self.update_default_time_range(time.time(), end_time)
 
     def save_monitoring_version(self, node):
         node.remoter.run(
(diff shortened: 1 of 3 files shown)


soyacz:
second pipeline run was with gce (also works): `hydra investigate show-monitor 38785640-83ec-47b9-9a44-4da97bdcebe7`

sdcm/cluster.py
```diff
@@ -5879,6 +5879,24 @@ def save_sct_dashboards_config(self, node):
         node.remoter.run('mkdir -p {}'.format(sct_monitoring_addons_dir), ignore_status=True)
         node.remoter.send_files(src=self.sct_dashboard_json_file, dst=sct_monitoring_addons_dir)
 
+    def update_default_time_range(self, start_timestamp: float, end_timestamp: float) -> None:
```
fruch, line 5882: let add a call to specific it on start of the monitoring stack

k0machi: Isn't it already there? `tester.py:2975`

sdcm/cluster.py
```diff
@@ -5879,6 +5879,24 @@ def save_sct_dashboards_config(self, node):
...
+        """
+        Specify the Grafana time range for all dashboards by updating the JSON files.
+
+        This method will find all JSON files in the `grafana/build/<subdir>` directories and replace
+        the "from" and "to" time range values with the provided start and end timestamps.
+        """
+        start_iso = datetime.fromtimestamp(start_timestamp, tz=timezone.utc).strftime('%Y-%m-%dT%H:%M:%SZ')
+        end_iso = datetime.fromtimestamp(end_timestamp, tz=timezone.utc).strftime('%Y-%m-%dT%H:%M:%SZ')
+        cmd = (f"find {os.path.join(self.monitor_install_path, 'grafana/build/')} -name '*.json' -exec "
```
k0machi, line 5891: Two questions:

Wouldn't it be better to parse the dashboard json and change the fields? I believe the `from` field is in the top level of the dashboard object.

Also, does this need a reload or does it persist after reload? That is, how does grafana handle this change?

k0machi (approved):
LGTM

sdcm/cluster.py
soyacz: in `grafana/build/` dir there are multiple dashboard files, different for each scylladb release. This command changes `from` and `to` fields in each of it (there's only one of them in each file).
This becomes the default.

sdcm/cluster.py
soyacz: yes, but in case of CI test timeout we might not get to 'teardown'. So idea is to update also at the beginning of a test and later update to proper values.

soyacz:
@fruch setting ranges at the beginning of a test works instantly for current run. So, after opening Grafana, user needs to narrow down (which is easy, so for me it's ok). WDYT?

fruch:
> @fruch setting ranges at the beginning of a test works instantly for current run. So, after opening Grafana, user needs to narrow down (which is easy, so for me it's ok). WDYT?

anyhow the default before was 6h from current time. 
it's not a big change for most tests, I consider it acceptable.

fruch (approved):
LGTM

PR #9002, opened 2024-10-14
feature(kafka-conn-sink): add tests for kafka sink connector

dimakr:
The change implements a test scenario for ScyllaDB Sink Connector (https://github.com/scylladb/kafka-connect-scylladb):
- it writes given number of messages (in AVRO or JSON data format) to a Kafka topic 
- then reads from Scylla rows from the corresponding table and validates (compares) the data

Closes: https://github.com/scylladb/qa-tasks/issues/1718

### Testing

- [ ]


sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,476 @@
...
+                self.collected_messages.append({'key': key_dict, 'value': value_dict})
+                LOGGER.debug("Consumed message: key=%s, value=%s", key_dict, value_dict)
+                last_message_time = current_time
+        finally:
+            self.consumer.close()
+
+    def read_rows_from_scylla(self) -> None:
+        LOGGER.info("Reading rows from ScyllaDB table '%s.%s'.", self.keyspace, self.table)
+        with self.tester.db_cluster.cql_connection_patient(self.tester.db_cluster.nodes[0]) as session:
+            query = f"SELECT * FROM {self.keyspace}.{self.table}"
```
fruch, line 412: for scaling, we'll need to read page by page

dimakr: Added reading of batches of rows in scylla.
Also as was discussed, added seeded generation of data to be written to Kafka; and later when we validate data in scylla we do not read it from kafka, but regenerate using the known seeds.

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+        elif normalized_type == 'boolean':
+            value = rand_gen.choice([True, False])
+        elif normalized_type == 'float':
+            value = rand_gen.uniform(0, 1000000)
+        elif normalized_type == 'double':
+            value = rand_gen.uniform(0, 10000000000)
+        elif normalized_type == 'bytes':
+            value = ''.join(rand_gen.choices(ascii_letters, k=10)).encode('utf-8')
+        else:
+            LOGGER.warning("Unsupported field type: %s. Using None.", field_type)
```
soyacz, line 179: should raise error event or fail the test immediately

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+                if field_name == 'id':
+                    data[field_name] = seed
+                else:
+                    data[field_name] = self.generate_field_value(field['type'], rand_gen)
+            return data
+        else:
+            return self.generate_field_value(schema['type'], rand_gen)
+
+    def generate_field_value(self, field_type: str, rand_gen: random.Random) -> int | float | str | bool | bytes | None:
+        type_mapping = {
```
soyacz, line 147: Maybe instead of mapping str to str, better map field_type to generator function? should be more performant and clearer.

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+
+    def wait_for_topic(self, topic: str, timeout: int) -> None:
+        wait_for(lambda: topic in self.get_topics(), text=f"waiting for topic={topic}", timeout=timeout)
+
+    def delivery_report(self, err: str, record: Message) -> None:
+        if err:
+            self.failed_count += 1
+            LOGGER.error("Kafka record delivery failed: %s", err)
+        else:
+            self.produced_count += 1
```
soyacz, line 242: `self.produced_count` -> `self.passed_count`

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+            LOGGER.debug(
+                "Kafka record delivered to %s [%s] at offset %s", record.topic(), record.partition(), record.offset())
+
+    def run(self) -> None:
+        start_time = time.time()
+        while not self.termination_event.is_set():
+            if self.num_records and self.produced_count >= self.num_records:
```
soyacz, line 249: will it ever end if something is so badly configured that nothing is deliverred? Shouldn't we base on `self.produced_count + self.failed_count`?

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+                        value, SerializationContext(self.topic, MessageField.VALUE)
+                    ) if self.value_serializer else None)
+            elif self.record_format == 'json-with-schema':
+                if self.key_schema:
+                    key = self.generate_json_with_schema_key(self.current_key)
+                    key_serialized = self.key_serializer(key, SerializationContext(self.topic, MessageField.KEY))
+                else:
+                    key_serialized = None
+                value = self.generate_json_with_schema_value(self.current_key)
+                value_serialized = self.value_serializer(value, SerializationContext(self.topic, MessageField.VALUE))
```
soyacz, line 272: this is very inefficient. Because `record_format` does not change, we could upfront create key and value generators once and use it in this loop.
Note, that each time we pass this loop, also `generate_avro_key` and `generate_avro_value` loads the same json.

configurations/kafka/kafka-sink-avro.yaml
```diff
@@ -0,0 +1,24 @@
+stress_cmd: >
+  python_thread -thread=KafkaProducerThread -record-format=avro -topic=table1 -num-records=1000
+  -key-schema='{"type":"record","name":"key_schema","fields":[{"name":"id","type":"int"}]}'
+  -value-schema='{"type":"record","name":"value_schema","fields":[{"name":"firstName","type":"string"},{"name":"lastName","type":"string"}]}'
+stress_read_cmd: >
+  python_thread -thread=KafkaValidatorThread -keyspace=keyspace1 -table=table1 -topic=table1 -record-format=avro -batch-size=500
```
soyacz, line 6: why batch-size here is 500 and not 1000 like in producer?

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+        self.keyspace = kwargs.get('keyspace', None)
+        self.table = kwargs.get('table', None)
+        self.current_key = 1
+        self.batch_wait_time = 10
+        self.batch_poll_interval = 2
+        self.validated_rows = 0
+
+    def run(self) -> None:
+        while not self.termination_event.is_set():
+            batch_end = min(self.current_key + self.batch_size - 1, self.num_records)
```
soyacz, line 353: but `num_records` is not set by params in `python_thread -thread=KafkaValidatorThread` - how does it work?

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+
+class KafkaValidatorThread(KafkaThreadBase):
+    """
+    Thread to validate that records added to Kafka are correctly written to ScyllaDB.
+    """
+
+    def __init__(self, tester, params: SCTConfiguration, timeout: int, **kwargs):
+        super().__init__(tester, params, timeout, **kwargs)
+        self.mismatches = []
+        self.batch_size = int(kwargs.get('batch-size', 1000))
```
soyacz, line 343: Please add documentation on params for these threads. E.g. what is batch_size.

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+            query = f"SELECT * FROM {self.keyspace}.{self.table} WHERE id >= %s AND id <= %s ALLOW FILTERING"
+            rows = list(session.execute(query, parameters=(range_start, range_end)))
+            LOGGER.debug("Retrieved %d rows from ScyllaDB for the current batch.", len(rows))
+            return rows
+
+    def compare_rows(self, rows: list) -> None:
+        LOGGER.info("Comparing batch of ScyllaDB rows with expected data.")
+        for row in rows:
+            id_value = row.id
+            expected_data = self.generate_data_from_schema(json.loads(self.value_schema), id_value)
```
soyacz, line 389: please use explicit arugments names to make it clear that `id_value` is seed

configurations/kafka/kafka-sink-avro.yaml
```diff
@@ -0,0 +1,24 @@
+stress_cmd: >
+  python_thread -thread=KafkaProducerThread -record-format=avro -topic=table1 -num-records=1000
```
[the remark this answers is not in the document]
dimakr: As it was discussed, now this test and the one for CDC source connector are a kind of POC - for now we introduce only the baseline, which checks functionality. Then once we decide what to use for Kafka (VMs or SaaS) we can start scaling.

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,410 @@
...
+                    LOGGER.error("Invalid format or type for %s: %s", option, value)
+            else:
+                filtered_options[option] = value
+        else:
+            LOGGER.warning("Unknown Kafka option %s encountered.", option)
+
+    return filtered_options
+
+
+class KafkaThreadBase(Thread):
```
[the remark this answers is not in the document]
dimakr: The test is initial setup and functionality check for Sink connector. And it continues on top of the 'python_thread' stress cmd approach that was selected for Source connector.
Once we move from initial setup/functionality check, we can start thinking about scaling and optimization.

configurations/kafka/kafka-sink-avro.yaml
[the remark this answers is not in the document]
soyacz: still, I think it can handle more than 1k. Before we scale it would be great to know this solution limitations.

sdcm/utils/common.py
```diff
@@ -2672,3 +2673,47 @@ def skip_optional_stage(stage_names: str | list[str]) -> bool:
...
+                    # Next token is the value for the current option
+                    options[option_name] = next_token
+                else:
+                    # Option is a flag
+                    options[option_name] = True
+                    if next_token:
+                        # Next token is another option; re-insert it into the iterator
+                        tokens_iter = iter([next_token] + list(tokens_iter))
+        else:
+            LOGGER.warning(f"Unknown argument or option '{token}' encountered.")
```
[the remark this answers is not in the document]
dimakr: This filter_kafka_options function is for parsing only kafka specific options, and as additional functionality the not-kafka-related options are filtered out. This warning is just to inform that such non-kafka options were provided to python_thread command, they should not fail the execution.
But I'll remove the warning, so it won't cause confusion.

sdcm/kafka/kafka_producer.py
dimakr: Added raising error event.

sdcm/kafka/kafka_producer.py
dimakr: Changed to field_type:gen_function mapping

sdcm/kafka/kafka_producer.py
dimakr: Right, changed to checking `self.produced_count + self.failed_count` total (and removed checking for self.num_records presence, as it doesn't bring value - we always have num_records, either from stress command params, or default one).
In case something wrong is happening with producing records we have the thread timeout, and if it is expired the thread will be stopped (termination_event will be set).

sdcm/kafka/kafka_producer.py
dimakr: Missed this. Initial implementation of the test was without batching, but as it was added the num-records should be passed explicitly (otherwise it will take the 1000 default one).
Fixed in configs.

sdcm/kafka/kafka_producer.py
dimakr: Fixed this

sdcm/kafka/kafka_producer.py
dimakr: Added doc. strings.

configurations/kafka/kafka-sink-avro.yaml
dimakr: Changed the code to align with the test config.

sdcm/kafka/kafka_producer.py
dimakr: Fixed. Deserialized the schemas once in constructor.

sdcm/kafka/kafka_producer.py
dimakr: For me `passed` in a name is immediately something about passing the tests. I doubt it fits here.
Producer/produced is a term from Kafka itself, for that purpose it was selected here. Another alternative could be delivered_count ('delivery' term is also used in kafka extensively).

soyacz: yes, then 'delivered_count' and 'undelivered_count' should be used in that case (I got confused with `failed_count`)

dimakr: Right, `failed` can also be considered as some test status related thing at the first place )
Changed the names.

test-cases/kafka/longevity-kafka-sink.yaml
```diff
@@ -0,0 +1,19 @@
+test_duration: 60
+
+n_loaders: 1
+n_monitor_nodes: 1
+n_db_nodes: 1
+
+instance_type_runner: c7i.2xlarge
+instance_type_db: i4i.large
```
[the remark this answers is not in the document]
dimakr: Executed test on AWS - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/kafka-sink-connector-test/11/ (also added the run it to PR description)

configurations/kafka/kafka-sink-avro.yaml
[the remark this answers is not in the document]
dimakr: I tried he 1M records on my machine on docker backend, with batches of 10000 rows when validating.
It looks like this:
```
...
< t:2024-10-18 14:44:25,509 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 1 to 10000.
< t:2024-10-18 14:44:26,093 f:nemesis.py      l:426  c:sdcm.nemesis         p:INFO  > sdcm.nemesis.NoOpMonkey: Interval: 60 s
< t:2024-10-18 14:44:27,853 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:44:28,179 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 10001 to 20000.
< t:2024-10-18 14:44:31,356 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:44:31,647 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 20001 to 30000.
< t:2024-10-18 14:44:35,246 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:44:35,533 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 30001 to 40000.
< t:2024-10-18 14:44:39,949 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:44:40,220 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 40001 to 50000.
< t:2024-10-18 14:44:45,236 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:44:45,540 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 50001 to 60000.
< t:2024-10-18 14:44:51,854 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:44:52,122 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 60001 to 70000.
< t:2024-10-18 14:44:59,550 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:44:59,828 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 70001 to 80000.
< t:2024-10-18 14:45:08,845 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:45:09,123 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 80001 to 90000.
< t:2024-10-18 14:45:19,242 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:45:19,523 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 90001 to 100000.
< t:2024-10-18 14:45:32,938 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:45:33,205 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 100001 to 110000.
< t:2024-10-18 14:45:47,756 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:45:48,040 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 110001 to 120000.
< t:2024-10-18 14:46:03,969 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:46:04,250 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 120001 to 130000.
< t:2024-10-18 14:46:06,485 f:kafka_producer.py l:278  c:sdcm.kafka.kafka_producer p:INFO  > Requested number of records is produced, stopping Kafka producer thread.
< t:2024-10-18 14:46:22,238 f:kafka_producer.py l:440  c:sdcm.kafka.kafka_producer p:INFO  > Comparing batch of ScyllaDB rows with expected data.
< t:2024-10-18 14:46:22,517 f:kafka_producer.py l:432  c:sdcm.kafka.kafka_producer p:INFO  > Reading rows from ScyllaDB for keys 130001 to 140000.
...
< t:2024-10-18 15:13:38,209 f:kafka_producer.py l:428  c:sdcm.kafka.kafka_producer p:INFO  > All records processed, stopping Kafka validator thread.

.
----------------------------------------------------------------------
Ran 1 test in 2119.390s
```

So the Producer and Validator threads start at the same time. It takes ~1m40s to produce 1M records in Kafka. Then Validator spends ~30m to validate all rows in Scylla (I used here batches of 10000).
At the beginning it spends ~2s seconds per batch. But this time is growing and at the end of validation it spends ~20s per batch. This is probably expected as the range query that selects rows from Scylla is executed with ALLOW FILTERING. There is no probably other straightforward way, as the Sink Connector has settings/capabilities to transform a selected field from Kafka to only primary key in Scylla, there is no mechanism to define a clustering key.
But I think we should be concerned about performance only when we start thinking about scaling - for now it is only a docker backend on a local machine. Moreover, this is the current approach of validating all rows that were sent from Kafka to Scylla. At the end we can decide to verify data consistency only by verifying some few randomly selected rows in Scylla.

sdcm/utils/common.py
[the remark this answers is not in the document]
soyacz: still not removed

sdcm/kafka/kafka_producer.py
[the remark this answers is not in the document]
soyacz: if the intent is to just do setup/functionality check that's fine. From my understanding we wanted to also verify how much throughput we can get from this setup before we decide to jump to better solutions - which anyway will require more performant producers/validators. 
We can do it in followup if needed.

sdcm/kafka/kafka_producer.py
soyacz: Good. See you can go even further by crating `key_generator` and `value_generator` before while loop starts to further optimize. But again, if this is only for functional test then this implementation is fine.

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,464 @@
...
+            LOGGER.error("Kafka record delivery failed: %s", err)
+        else:
+            self.delivered_count += 1
+            LOGGER.debug(
+                "Kafka record delivered to %s [%s] at offset %s", record.topic(), record.partition(), record.offset())
+
+    def run(self) -> None:
+        start_time = time.time()
+        while not self.termination_event.is_set():
+            if self.delivered_count + self.undelivered_count >= self.num_records:
```
soyacz, line 277: btw. why we cannot rely on `current_key` number?

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,464 @@
...
+            python_thread -thread=KafkaValidatorThread
+            -keyspace=keyspace1 -table=table1 -record-format=json-with-schema -num-records=1000 -batch-size=500
+            -value-schema='{"type":"struct","fields":[{"type":"string","optional":false,"field":"name"},{"type":"string","optional":true,"field":"department"}]}'
+        ```
+        """
+        super().__init__(tester, params, timeout, **kwargs)
+        self.mismatches = []
+        self.batch_size = int(kwargs.get('batch-size', 500))
+        self.keyspace = kwargs.get('keyspace', None)
+        self.table = kwargs.get('table', None)
```
soyacz, line 399: after stripping off kafka arguments it's good to verify all the arguments if are matching required ones. This way we would prevent silently passing if someone makes typo in them. E.g. if someone passes `-num-rcords` then we would use default 1000 instead of selected one and no one would know.

sdcm/kafka/kafka_producer.py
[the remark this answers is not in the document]
fruch: it's intended for functionally first, once we'll have a clear structure of what we need.

we might choose to implement in others languages

regardless the kafka itself for now would share the resources of the SCT runner.
only after we'll be able such how much we can scale that setup, we'll look into what can be improved

sdcm/kafka/kafka_producer.py
dimakr: We can. Initial implementation was not having pagination/batching, so the thread instance didn't have current_key attribute at that time.

sdcm/kafka/kafka_producer.py
```diff
@@ -0,0 +1,476 @@
...
+            else:
+                filtered_options[option] = value
+
+    return filtered_options
+
+
+class KafkaThreadBase(Thread):
+    def __init__(self, tester, params: SCTConfiguration, timeout: int, **kwargs):
+        super().__init__(daemon=True)
+        self._check_thread_options(**kwargs)
```
soyacz, line 82: another approach would be to use `pop` instead `get` from `kwargs` and verify at the end that `kwargs` is empty. So we wouldn't have to force users using it and use defaults.

dimakr:
As was discussed, switched from using kafka-python to confluent-kafka-python client in kafka_consumer module. confluent-kafka-python has some features out of the box (e.g. schemas serialization/deserialization functionality), which kafka-python don't have. Also confluent client is actively maintained (last release is from 2024), when kafka-python last stable release is from 2020.

sdcm/kafka/kafka_producer.py
dimakr: After a second thought I think it's better to keep this from of a condition. It keeps semantics consistent - the same objects are checked (requested num of records in kafka against currently processed nu of records).
current_key is a field of a Kafka message, and reading the condition when using it would mean - compare value of a field of kafka message with the requested num of records.

fruch (approved):
LGTM

soyacz (dismissed):
LGTM
@fruch It's creating new hydra image so maybe we should backport it to other branches to avoid future problems with backports?

sdcm/utils/common.py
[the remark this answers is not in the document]
dimakr: Right. Similar warning was in 2 places, in both confusing. Now removed from both places.

fruch:
> LGTM @fruch It's creating new hydra image so maybe we should backport it to other branches to avoid future problems with backports?

yes we should

dimakr:
@fruch resolved conflict in docker/env/version

soyacz (approved):
LGTM

fruch (approved):

PR #9005, opened 2024-10-14
refactor(SCHEMA_FAILURE): Change default severity of SCHEMA_FAILURE

roydahan:
SCT detects the regex "Failed to load schema version" and creates an ERROR event out of it even if the message in scylla is with INFO severity. this behviour raises issues like
"https://github.com/scylladb/scylladb/issues/19093" which gets P3 priority but continue failing SCT tests over and over. Hence, this commit suggests to reduce this message severity.

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/sct_events/database.py
+++ b/sdcm/sct_events/database.py
@@ -115,8 +115,9 @@ DatabaseLogEvent.add_subevent_type("COMPACTION_STOPPED", severity=Severity.NORMA
                                    regex="compaction_stopped_exception")
 DatabaseLogEvent.add_subevent_type("BAD_ALLOC", severity=Severity.ERROR,
                                    regex="std::bad_alloc")
+# Due to scylla issue https://github.com/scylladb/scylladb/issues/19093, we need to suppress the below error
 DatabaseLogEvent.add_subevent_type("SCHEMA_FAILURE", severity=Severity.ERROR,
-                                   regex="Failed to load schema version")
+                                   regex=r'(.*ERROR|!ERR).*Failed to load schema version')
 DatabaseLogEvent.add_subevent_type("RUNTIME_ERROR", severity=Severity.ERROR,
                                    regex="std::runtime_error")
 # remove below workaround after dropping support for Scylla 2023.1 and 5.2 (see scylladb/scylla#13538)
--- a/unit_tests/test_cluster.py
+++ b/unit_tests/test_cluster.py
@@ -117,7 +117,7 @@ class TestBaseNode(unittest.TestCase, EventsUtilsMixin):
 
     def test_search_system_log(self):
         critical_errors = list(self.node.follow_system_log(start_from_beginning=True))
-        self.assertEqual(34, len(critical_errors))
+        self.assertEqual(33, len(critical_errors))
 
     def test_search_system_log_specific_log(self):
         errors = list(self.node.follow_system_log(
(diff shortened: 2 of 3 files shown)


defaults/severities.yaml
```diff
@@ -30,7 +30,7 @@ DatabaseLogEvent.GATE_CLOSED: WARNING
 DatabaseLogEvent.RESTARTED_DUE_TO_TIME_OUT: WARNING
 DatabaseLogEvent.EMPTY_NESTED_EXCEPTION: WARNING
 DatabaseLogEvent.BAD_ALLOC: ERROR
-DatabaseLogEvent.SCHEMA_FAILURE: ERROR
+DatabaseLogEvent.SCHEMA_FAILURE: WARNING
```
soyacz, line 33: this is rather not needed as severities.yaml define max severities only.

soyacz (dismissed):
LGTM

sdcm/sct_events/database.py
```diff
@@ -115,7 +115,8 @@ def add_info(self: T_log_event, node, line: str, line_number: int) -> T_log_even
                                    regex="compaction_stopped_exception")
 DatabaseLogEvent.add_subevent_type("BAD_ALLOC", severity=Severity.ERROR,
                                    regex="std::bad_alloc")
-DatabaseLogEvent.add_subevent_type("SCHEMA_FAILURE", severity=Severity.ERROR,
+# Due to scylla issue https://github.com/scylladb/scylladb/issues/19093, we need to suppress the below error
+DatabaseLogEvent.add_subevent_type("SCHEMA_FAILURE", severity=Severity.WARNING,
```
vponomaryov, line 119: Should the `SkipPerIssues` be used here?

fruch:
As per https://github.com/scylladb/scylladb/issues/19093#issuecomment-2413360802

This should be skipped for a specific case only, and not by default

roydahan:
My assumption here is that if Scylla raises this message as ERROR, then it will become an error anyway.
Is that what will happen?
Maybe indeed I need to remove it from the max severity of WARNING.

And in that case, I'm not going to handle it in SCT, Scylla should know when to throw error and when INFO.

sdcm/sct_events/database.py
temichus: agree.
```
severity=Severity.ERROR
if SkipPerIssues("scylladb/scylladb#19093", params):
    severity=Severity.WARNING
    
DatabaseLogEvent.add_subevent_type("SCHEMA_FAILURE", severity=severity ........
```

but probably for this, we need to bring the config 
```
from sdcm.sct_config import init_and_verify_sct_config
params=init_and_verify_sct_config()
```

roydahan:
Change in direction, I decided to change the regex we look for and raise ERROR event only when the log message from scylla is ERR.

sdcm/sct_events/database.py
```diff
@@ -115,8 +115,9 @@ def add_info(self: T_log_event, node, line: str, line_number: int) -> T_log_even
                                    regex="compaction_stopped_exception")
 DatabaseLogEvent.add_subevent_type("BAD_ALLOC", severity=Severity.ERROR,
                                    regex="std::bad_alloc")
+# Due to scylla issue https://github.com/scylladb/scylladb/issues/19093, we need to suppress the below error
 DatabaseLogEvent.add_subevent_type("SCHEMA_FAILURE", severity=Severity.ERROR,
-                                   regex="Failed to load schema version")
+                                   regex=r'(.*ERROR|!ERR).*Failed to load schema version')
```
soyacz, line 120: in `DATABASE_ERROR` we use `(^ERROR|!\s*?ERR).*` for matching errors, possibly we should use the same.
I'm not sure of the log format we get here -  I see in unit tests multiple formats.

fruch (approved):
LGTM

sdcm/sct_events/database.py
roydahan: The unit test failed when I used "^", I wasn't sure how updated our system.log is.

PR #9013, opened 2024-10-17
fix(SLA) fix 1to5 ratio test for 2024.3

temichus:
fix 1to5 ratio test for 2024.3
where tablets are enabled by default

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/edbfccd0-c328-4dfb-aff2-67d0165573a4 - test failed looks like due to https://github.com/scylladb/scylla-enterprise/issues/3737, but there is no message that cluster is not loaded enought

### Reminders

### DIFF
--- a/sla_per_user_system_test.py
+++ b/sla_per_user_system_test.py
@@ -22,6 +22,7 @@ from sdcm.es import ES
 from sdcm.sct_events import Severity
 from sdcm.sct_events.workload_prioritisation import WorkloadPrioritisationEvent
 from test_lib.sla import ServiceLevel, Role, User
+from sdcm.utils.features import is_tablets_feature_enabled
 
 
 # pylint: disable=too-many-public-methods
@@ -37,7 +38,7 @@ class SlaPerUserTest(LongevityTest):
                                 ' -schema \'replication(strategy=NetworkTopologyStrategy,replication_factor=3)\' ' \
         '-mode cql3 native user={user} password={password} -rate threads={threads} ' \
         'throttle=10000/s -pop seq={pop}'
-    STRESS_READ_CMD = 'cassandra-stress read cl=ALL duration={duration} -mode cql3 native user={user} ' \
+    STRESS_READ_CMD = 'cassandra-stress read cl=ALL duration={duration} -mode connectionsPerHost=16 cql3 native user={user} ' \
                       'password={password} -rate threads={threads} -pop {pop}'
     STRESS_MIXED_CMD = r"cassandra-stress mixed ratio\(write={write_ratio},read={write_ratio}\) cl=QUORUM " \
                        "duration={duration} " \
@@ -371,6 +372,14 @@ class SlaPerUserTest(LongevityTest):
 
     def _two_users_load_througput_workload(self, shares, load, expected_shares_ratio=None):
         session = self.prepare_schema()
+
+        if is_tablets_feature_enabled(session=session):
+            self.run_pre_create_keyspace()
+            # after several test runs with Tablets decided to decrease by half of the percent(usually tests show about 96.8 - 97.5)
+            # due to unbalanced shards utilization with tablets(particular tablet belong to particular shard)
+            # during gauss distribution read
+            self.MIN_CPU_UTILIZATION = 96.5
+
         self.create_test_data_and_wait_no_compaction()
 
         # Define Service Levels/Roles/Users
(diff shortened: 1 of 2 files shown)


aleksbykov (dismissed):
LGTM
Short question, why we limit connections per host?

temichus:
> LGTM Short question, why we limit connections per host?

2 stress commands from different users run from different nodes (, 'round_robin': True)

```
 read_queue = self.run_stress_and_verify_threads(params={'stress_cmd': read_cmds, 'round_robin': True})
```

sla_per_user_system_test.py
```diff
@@ -37,7 +38,7 @@ class SlaPerUserTest(LongevityTest):
                                 ' -schema \'replication(strategy=NetworkTopologyStrategy,replication_factor=3)\' ' \
         '-mode cql3 native user={user} password={password} -rate threads={threads} ' \
         'throttle=10000/s -pop seq={pop}'
-    STRESS_READ_CMD = 'cassandra-stress read cl=ALL duration={duration} -mode cql3 native user={user} ' \
+    STRESS_READ_CMD = 'cassandra-stress read cl=ALL duration={duration} -mode connectionsPerHost=16 cql3 native user={user} ' \
```
juliayakovlev, line 41: It may have an impact on results of non-tablets test. Did you run such test?

sla_per_user_system_test.py
```diff
@@ -371,6 +372,13 @@ def test_read_throughput_1to5_ratio(self):
 
     def _two_users_load_througput_workload(self, shares, load, expected_shares_ratio=None):
         session = self.prepare_schema()
+
+        if is_tablets_feature_enabled(session=session):
+            session.execute("""
```
juliayakovlev, line 377: There is `pre_create_keyspace` test parameter. 
Longevity test has a function `longevity_test.LongevityTest.run_pre_create_keyspace` (that inherits in SLA test)
This command can be added to the test yaml.
Or under `scylla-cluster-tests/configurations` and the used in the jenkins pipeline.
See example here:
https://github.com/scylladb/scylla-cluster-tests/blob/99d539bbdc3a4ec912511e761275f0c202fd9a30/jenkins-pipelines/manager/debian11-manager-install.jenkinsfile#L10.

This way prevents hardcoded cql command inside the code

juliayakovlev (changes requested):

sla_per_user_system_test.py
temichus: but with ```pre_create_keyspace``` and ```scylla-cluster-tests/configurations``` I cant use if is_tablets_feature_enabled

and adaptively create keyspace for Scylla versions with and without tablets

sla_per_user_system_test.py
temichus: good point, will check

sla_per_user_system_test.py
juliayakovlev: yes, you still can handle this command when tablets are enabled.
for you decision. I think it is better to put it in test yaml (if you do not like configurations )

sla_per_user_system_test.py
temichus: in 2024.2 without tablets test pass https://argus.scylladb.com/tests/scylla-cluster-tests/bb6dd4c4-f720-4486-8455-4ac8128a6e8e

juliayakovlev (dismissed):

temichus:
after several test runs decided to decrease MIN_CPU_UTILIZATION to 96.5(half of per cent)
due to still unbalanced shards utilization with tablets
image

temichus:
retested with lattest changes https://argus.scylladb.com/tests/scylla-cluster-tests/78679646-f241-4e48-984e-6dc5cfa73c91

@aleksbykov @juliayakovlev please rereview

sla_per_user_system_test.py
```diff
@@ -371,6 +372,11 @@ def test_read_throughput_1to5_ratio(self):
 
     def _two_users_load_througput_workload(self, shares, load, expected_shares_ratio=None):
         session = self.prepare_schema()
+
+        if is_tablets_feature_enabled(session=session):
+            self.run_pre_create_keyspace()
+            self.MIN_CPU_UTILIZATION = 96.5
```
soyacz, line 378: can you add a commentline why it was added?

soyacz (commented):
One small comment, otherwise LGTM

soyacz (approved):
LGTM

aleksbykov (approved):
LGTM

PR #9018, opened 2024-10-21
full_storage_utilization_test: Storage utilization at 90% cluster size

Lakshmipathi:
Populate data until it reaches over 90% disk storage then perform db and topology change cluster options.

### Description
This PR covers basic part of 90% storage utilization, populate data when the cluster reaches over 90% disk usage
performs scaleout operation.Goal of this task is to run the cluster at 90% disk utilization. This will help the users to utilize their instances fully.  To help achieve this goal, we need to utilize features like concurrent topology changes, tablets and migrations.

**Keeping this as draft version so that it acts as common base between dev and qa team so it helps to improve this PR further.** 

### Testing

- [ ] 

### Reminders


full_storage_utilization_test.py
```diff
@@ -0,0 +1,149 @@
...
+        Sleep for 90 minutes.
+        Add 4th node.
+        """
+        self.log.info("Running df command on all nodes:")
+        self.get_df_output()
+        self.run_stress(85)
+        self.run_stress(93)
+
+        self.log.info("Adding a new node")
+        self.add_new_node()
```
paszkow, line 33: One major comment here. Adding a node, will always succeed because you no longer write data. Your implementation of `run_stress_until_target` to prepare a cluster is fine. Once this is achieved, you want to rather react based on the storage utilization.

I think, this part of adding a node shall be written differently. Namely, you want to monitor the disk usage while writing bunch of data and if a certain threshold is reached, you extend a cluster by adding a new node.

Lakshmipathi: Updated the PR, now after reaching 90% storage limit, script starts writing 10GB data over 30mins and add node is started. So there will be on-going write operation during topology change.

Lakshmipathi:
Recent log run can be found here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/LakshmipathiGanapathi/job/byo-longevity-test/185/consoleText (ignore error from send_email step from the job)

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+        return f'cassandra-stress write cl=ONE n={n} -mode cql3 native -rate threads=10 -pop dist="uniform(1..{seq_end})" ' \
+               f'-col "size=FIXED({row_size}) n=FIXED(1)" -schema "replication(strategy=NetworkTopologyStrategy,replication_factor=3)"'
+
+    def setUp(self):
+        super().setUp()
+        self.start_time = time.time()
+
+    def start_slow_write(self):
+        self.run_stress_thread(stress_cmd=self.stress_cmd_30mins, keyspace_name=f"{self.action_type}")
```
yarongilor, line 26: what's `action_type`? why it's the keyspace name?

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
...
+        self.log.info("Wait for 2 mins for the stress command to start")
+        time.sleep(120)
```
yarongilor, line 28: consider adding some constant definition variables for the sleep values

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
...
+        time.sleep(1800)
+
+    def drop_data(self, keyspace_name):
+        '''
+        Drop keyspace and clear snapshots.
+        '''
+        node = self.db_cluster.nodes[0]
+        self.log.info("Dropping some data")
+        cql = f"DROP KEYSPACE {keyspace_name}"
+        node.run_cqlsh(cql)
```
yarongilor, line 49: why not using the driver for CQL?

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
...
+        self.remove_node()
+        time.sleep(1800)
+
+    def drop_data(self, keyspace_name):
+        '''
+        Drop keyspace and clear snapshots.
+        '''
+        node = self.db_cluster.nodes[0]
+        self.log.info("Dropping some data")
+        cql = f"DROP KEYSPACE {keyspace_name}"
```
yarongilor, line 48: if decreasing dataset by dropping a keyspace, there could be no background read for it, if it is relevant somehow.

sdcm/sct_config.py
```diff
@@ -1231,6 +1231,9 @@ class SCTConfiguration(dict):
         dict(name="perf_gradual_throttle_steps", env="SCT_PERF_GRADUAL_THROTTLE_STEPS", type=dict,
              help="Used for gradual performance test. Define throttle for load step in ops. Example: {'read': ['100000', '150000'], 'mixed': ['300']}"),
 
+        # StorageUtilizationTest
+        dict(name="action_type", env="SCT_ACTION_TYPE", type=str,
```
yarongilor, line 1235: consider a less general name. like "auto_scaling_action_type"

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
...
+    def remove_node(self):  
+        self.log.info('Removing a first node from the cluster')
+        node_to_remove = self.db_cluster.nodes[0]
+        self.log.info(f"Node to be removed: {node_to_remove.name}")
+        self.db_cluster.decommission(node_to_remove)
+        self.db_cluster.wait_for_nodes_up_and_normal(nodes=[node_to_remove])
+        self.log.info(f"Node {node_to_remove.name} has been removed from the cluster")
+        self.monitors.reconfigure_scylla_monitoring()
+    
+    def flush_all_nodes(self):
```
yarongilor, line 137: the below util methods possibly already exist under sstabe-util or other utils directory.

full_storage_utilization_test.py
Lakshmipathi: ok, will update. waited 2 mins, so that throttled write operation will be running while we perform either scale-out or scale-in. Basic idea to have on-going read/write during topology change, its one of the suggestion by @lukepio

full_storage_utilization_test.py
Lakshmipathi: renamed it (auto_scaling_action_type) as suggested below. Just wanted a unique keyspace for each action. I think this may be not that important, will drop the keyspace_name, let it use default value.

full_storage_utilization_test.py
Lakshmipathi: I did a grep for cql on the codebase and saw above references for `node.run_cqlsh`, Are you referring to something like `with self.db_cluster.cql_connection_patient(node) as session` and use that session.

full_storage_utilization_test.py
Lakshmipathi: thanks, I didn't know that about drop keyspace. I think in this case, we don't read/write into that keyspace (keyspace_large1) - its mostly idle after disk usage across first 20%

full_storage_utilization_test.py
Lakshmipathi: grep -r 'node.run_nodetool("flush")' .` or  `gep -r 'flush_all'`  shows no util methods. longevity_test.py has similar function but not from util should we create one and add to util and update this file along with longevity_test.py?

full_storage_utilization_test.py
```diff
@@ -0,0 +1,197 @@
...
+        super().setUp()
+        self.start_time = time.time()
+
+    def start_throttle_write(self):
+        self.run_stress_thread(stress_cmd=self.stress_cmd_throttled)
+        self.log.info("Wait for 2 mins for the stress command to start")
+        time.sleep(self.sleep_time_before_autoscale)        
+        
+    def scale_out(self):
+        self.start_slow_write()
```
pehala, line 33: do we care about the performance of this? If so, will the output of this be visible in results?

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
+import time
+from sdcm.tester import ClusterTester
+
+
+class FullStorageUtilizationTest(ClusterTester):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.soft_limit = 75
```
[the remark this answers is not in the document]
Lakshmipathi: ok, let me add storage_utlitization_softlimit,storage_utlitization_hardlimit entries to yaml and ensure we have default value

sdcm/sct_config.py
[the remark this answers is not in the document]
Lakshmipathi: This is to control test-script via yaml entries. We can have something like `storage_utitlization_scaleout.yaml` with `action_type` as scale_out and another yaml like `storage_utitlization_scalein.yaml` with `action_type` as scale_in and so on.  This can used to cover our existing testcases in testplan with single script and multiple yaml files.

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
...
+        self.hard_limit = 90
+        self.stress_cmd_30mins = self.params.get('stress_cmd')
+        self.add_node_cnt = self.params.get('add_node_cnt')
+        self.action_type = self.params.get('action_type')
```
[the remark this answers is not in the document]
Lakshmipathi: ok, will add documentation for this config entries.

sdcm/sct_config.py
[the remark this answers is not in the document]
pehala: so everytime we add custom yaml parameter to some SCT test, we need to add it here?

Lakshmipathi:
grafana chart for scale-in. From the chart, around 10:15 disk usage reduced from 90% since a new node was added to 3-node cluster. Around 11:00 disk usage increased to 90% since node-1 was decommissioned. More details for this run can found from `hydra investigate show-monitor 1ffa6d64-004a-4443-a3c9-d52a18ea08e1` 

scalein

sdcm/sct_config.py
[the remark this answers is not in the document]
Lakshmipathi: Yes, I think so. I ran into unknown config entry error without these changes on this file.

full_storage_utilization_test.py
Lakshmipathi: we don't want to performance numbers from this. But we want to ensure this is a throttled write. So that it won't fill-up data during topology change operation. @paszkow  mentioned we need to run some disk-io during add-node operation.  With his experiments while performing 2k write/second he ran into enospc error.

full_storage_utilization_test.py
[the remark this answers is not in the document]
Lakshmipathi: documented them under ./docs/configuration_options.md

full_storage_utilization_test.py
```diff
@@ -0,0 +1,195 @@
...
+
+    def get_max_disk_usage(self):
+        max_usage = 0
+        for node in self.db_cluster.nodes:
+            result = node.remoter.run("df -h --output=pcent /var/lib/scylla | sed 1d | sed 's/%//'")
+            usage = int(result.stdout.strip())
+            max_usage = max(max_usage, usage)
+        return max_usage
+
+    def get_max_disk_used(self):
```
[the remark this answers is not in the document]
Lakshmipathi: now merged these two to return max_usage(in percentage), max_used.

docs/configuration_options.md
```diff
@@ -368,3 +368,6 @@
 | **<a href="#user-content-mgmt_agent_backup_config" name="mgmt_agent_backup_config">mgmt_agent_backup_config</a>**  | Manager agent backup general configuration: checkers, transfers, low_level_retries | N/A | SCT_MGMT_AGENT_BACKUP_CONFIG
 | **<a href="#user-content-mgmt_reuse_backup_snapshot_name" name="mgmt_reuse_backup_snapshot_name">mgmt_reuse_backup_snapshot_name</a>**  | Name of backup snapshot to use in Manager restore benchmark test | N/A | SCT_MGMT_REUSE_BACKUP_SNAPSHOT_NAME
 | **<a href="#user-content-mgmt_skip_post_restore_stress_read" name="mgmt_skip_post_restore_stress_read">mgmt_skip_post_restore_stress_read</a>**  | Skip post-restore c-s verification read in the Manager restore benchmark tests | N/A | SCT_MGMT_SKIP_POST_RESTORE_STRESS_READ
+| **<a href="#user-content-auto_scaling_action_type" name="auto_scaling_action_type">auto_scaling_action_type</a>**  | Refers to autoscale task like scaleout, scalein etc that needs to be performed at 90% storage| N/A | SCT_AUTO_SCALING_ACTION_TYPE
```
swasik, line 371: I would not call it auto-scaling, just scaling. Auto-scaling is a different feature that we would like to implement in the future - automatic change of provisioned capacity units when we are close to the utilization limit.

sdcm/sct_config.py
```diff
@@ -1231,6 +1231,14 @@ class SCTConfiguration(dict):
         dict(name="perf_gradual_throttle_steps", env="SCT_PERF_GRADUAL_THROTTLE_STEPS", type=dict,
              help="Used for gradual performance test. Define throttle for load step in ops. Example: {'read': ['100000', '150000'], 'mixed': ['300']}"),
 
+        # StorageUtilizationTest
+        dict(name="auto_scaling_action_type", env="SCT_AUTO_SCALING_ACTION_TYPE", type=str,
+             help="Perform specific tasks (scale_in,scale_out etc)"),
```
fruch, line 1236: The description doesn't explain anything 

Also if there are multiple values supported, it should be listed as options=

fruch (commented):
I would recommend moving most of the utility functions into its own module

So it can be reused, like for the case of doing the same sequences inside a nemesis.

Also for having unit tests for those functions, it would be easy if it's on a separated module.

full_storage_utilization_test.py
```diff
@@ -0,0 +1,200 @@
...
+from sdcm.tester import ClusterTester
+from sdcm.utils.tablets.common import wait_for_tablets_balanced
+
+
+class FullStorageUtilizationTest(ClusterTester):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.sleep_time_before_autoscale = 120
+        self.sleep_time_fill_disk = 1800
+        self.soft_limit = self.params.get('diskusage_softlimit') if self.params.get('diskusage_softlimit') is not None else 70
```
pehala, line 11: doesnt `get` allow default value?

full_storage_utilization_test.py
```diff
@@ -0,0 +1,200 @@
...
+        node = self.db_cluster.nodes[0]
+        self.log.info("Dropping some data")
+        cql = f"DROP KEYSPACE {keyspace_name}"
+        node.run_cqlsh(cql)
+        #node.run_nodetool(f"clearsnapshot")
+
+    def perform_action(self):        
+        self.log_disk_usage()
+        # Trigger specific action
+        if self.auto_scaling_action_type == "scale_out":
```
pehala, line 65: this screams Enum :)

test-cases/scale/full-storage-utilization.yaml
```diff
@@ -0,0 +1,19 @@
+test_duration: 3600
+n_db_nodes: 4
+n_loaders: 1
+n_monitor_nodes: 1
+user_prefix: 'df-test'
+instance_type_db: 'i4i.large'
+instance_provision: 'spot'
+
+post_behavior_db_nodes: "destroy"
```
pehala, line 9: this configuration option seems like it should be set from Jenkins instead

full_storage_utilization_test.py
fruch: `SctConfiguration.get` doesn't, on purpose, to make sure all default are listed in specific designated places under `defaults/test_defaults.yaml` or per backend.

and not scattered across the code, in multiple places the same config options might be use.
at some point it was like that, and it was mass changing the default, or even knowing what they are.

I would recommend default to be kept in configuration files, and not in the code like the above.

full_storage_utilization_test.py
fruch: for one single CQL statement, both works.

but cqlsh is a bit limited and harder to configure things, like timeouts, paging when needed, and so.

cqlsh mean opening ssh connection to a node, and opening process to execute the CQL statement
with the driver, it's just opening a session, all done in the same context of the test (communicating to the DB), but it's would be faster then using cqlsh.

full_storage_utilization_test.py
fruch: general recommendation

if you put sleep in test code, even if it's configurable, it should have a clear reasoning in a comment before it.

regardless we should always look for a way to way for the actual "thing" we are really waiting for

in this case you probably wanted to wait for writes to start, and we have such example in the rolling upgrade tests:

```
self.metric_has_data(
            metric_query='sct_cassandra_stress_write_gauge{type="ops", keyspace="keyspace1"}', n=5)
```

it's a way to wait for a specific metric to appear

it might not fit perfectly, but you might be able to extend to fit your needs.

full_storage_utilization_test.py
fruch: there's `_flush_all_nodes()` function which sits on longevity_test.py, and not used at all.
there are more places in the code, in manager test doing flush on all nodes.

this kind of a utility function, can be put on `BaseScyllaCluster` for example, so it can be used from multiple places

sdcm/sct_config.py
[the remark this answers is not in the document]
fruch: yes, we are strict about the SCT configuration, every single one should be listed here and with explanations what's the usage of it.

if it's specific to some tests, or type of tests.

we need this to be able to convert correctly from environment variable to python types.
and for validation that the configuration (we cover it during SCT CI)

as @ Lakshmipathi  point multiple config files can be used, for example to have a set of configuration that can be shared between multiple tests. or for enabling/disabling features.

sdcm/sct_config.py
Lakshmipathi: Updated it now.

full_storage_utilization_test.py
Lakshmipathi: modified now :)

full_storage_utilization_test.py
Lakshmipathi: ok, update the code and added these two limits into `defaults/test_defaults.yaml`

full_storage_utilization_test.py
Lakshmipathi: ok, updated the code to use session.

full_storage_utilization_test.py
Lakshmipathi: ok, added into BaseScyllaCluster and using it from there.

full_storage_utilization_test.py
```diff
@@ -0,0 +1,233 @@
...
+        self.run_stress_until_target(target_used_size, target_usage)
+
+        self.log_disk_usage()
+        self.log.info(f"Wait for {sleep_time} seconds")
+        time.sleep(sleep_time)  
+        self.log_disk_usage()
+
+    def run_stress_until_target(self, target_used_size, target_usage):
+        current_usage, current_used = self.get_max_disk_usage()
+        num = 0
```
cezarmoise, line 140: I don't think this affects anything, but because the function is called twice in sequence and num resets inside, on the second run (hard limit), it will write data in existing keyspaces, not new ones.

Also, large keyspaces start at number 1, but small keyspaces wil start at a later number.

Lakshmipathi: ah, right. Let me fix it and ensure num doesn't reset during second run. Will also update the code to handle small keyspace properly.

full_storage_utilization_test.py
```diff
@@ -0,0 +1,237 @@
...
+        wait_for_tablets_balanced(self.db_cluster.nodes[0])
+    
+    def get_max_disk_usage(self):
+        max_usage = 0
+        max_used = 0
+        for node in self.db_cluster.nodes:
+            result = node.remoter.run("df -h -BG --output=used,pcent /var/lib/scylla | sed 1d | sed 's/G//' | sed 's/%//'")
+            used, usage = result.stdout.strip().split()
+            max_usage = max(max_usage, int(usage))
+            max_used = max(max_used, int(used))
```
cezarmoise, line 197:
```suggestion
            info = self.get_disk_info(node)
            max_usage = max(max_usage, info["used_percent"])
            max_used = max(max_used, info["used"])
```
Since we already have the function that calls `df`

Lakshmipathi: updated the code.

pehala:
@Lakshmipathi Could we finish this PR and add other cases in followups?

Lakshmipathi:
> @Lakshmipathi Could we finish this PR and add other cases in followups? Is it ready for review?

Yes, I was moving some code into utils and testing them. Now pushed and its ready for review.

Lakshmipathi:
Any thoughts on how to view pre-commit failure messages? https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9018/workflow-stage/ gives me permission denied

pehala:
> Any thoughts on how to view pre-commit failure messages? https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9018/workflow-stage/ gives me permission denied

I can see the logs from the job https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9018/49/console but all phases seemed to be green? @fruch could you please advise on whats wrong?

fruch:
> > Any thoughts on how to view pre-commit failure messages? https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9018/workflow-stage/ gives me permission denied
> 
> I can see the logs from the job https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9018/49/console but all phases seemed to be green? @fruch could you please advise on whats wrong?

as listed in the github actions, the pre-commit phase is faling:
https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9018/49/execution/node/23/log/

the permission denied is a bug in jenkins `workflow-stage` page
got back to the root of theat job:
https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9018/
and follow the specific runs from there

regardless I would recommend installing correctly the pre-commit hook.
and run it again locally, after rebase/amend, that it something might miss the changed files

docs/configuration_options.md
```diff
@@ -371,3 +371,6 @@
...
 | **<a href="#user-content-zero_token_instance_type_db" name="zero_token_instance_type_db">zero_token_instance_type_db</a>**  | Instance type for zero token node | i4i.large | SCT_ZERO_TOKEN_INSTANCE_TYPE_DB
 | **<a href="#user-content-use_zero_nodes" name="use_zero_nodes">use_zero_nodes</a>**  | If True, enable support in sct of zero nodes(configuration, nemesis) | false | SCT_USE_ZERO_NODES
+| **<a href="#user-content-scaling_action_type" name="scaling_action_type">scaling_action_type</a>**  | Refers to scaling task like scaleout, scalein etc that needs to be performed when storage reaches specific threshold| N/A | SCT_AUTO_SCALING_ACTION_TYPE
+| **<a href="#user-content-diskusage_softlimit" name="diskusage_softlimit">diskusage_softlimit</a>**  | Soft disk usage limit until this limit data will be written as 10% of available diskspace, after this chunksize will be 1%| 70 | SCT_DISKUSAGE_SOFTLIMIT
```
swasik, line 375: 1 GB not 1%

sdcm/utils/full_storage_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        smaller_dataset = False
+        
+        space_needed = target_used_size - current_used
+        # Calculate chunk size as 10% of space needed
+        chunk_size = int(space_needed * 0.1)
+        
+        while current_used < target_used_size and current_usage < target_usage:
+            # Write smaller dataset near the threshold (15% or 30GB of the target)
+            smaller_dataset = (((target_used_size - current_used) < 30) or 
+                             ((target_usage - current_usage) <= 15))
```
swasik, line 86: I think it should be <= hardlimit - softlimit if we want to be inline with documentation (not 15)

sdcm/utils/full_storage_utils.py
```diff
@@ -0,0 +1,117 @@
...
+            
+            if not smaller_dataset:
+                self.total_large_ks += 1
```
swasik, line 89: nit: I would prefer large_ks_cnt - in the end it will be total but we use it as a counter

docs/configuration_options.md
```diff
@@ -371,3 +371,6 @@
 | **<a href="#user-content-n_db_zero_token_nodes" name="n_db_zero_token_nodes">n_db_zero_token_nodes</a>**  | Number of zero token nodes in cluster. Value should be set as "0 1 1"<br>for multidc configuration in same manner as 'n_db_nodes' and should be equal<br>number of regions | N/A | SCT_N_DB_ZERO_TOKEN_NODES
 | **<a href="#user-content-zero_token_instance_type_db" name="zero_token_instance_type_db">zero_token_instance_type_db</a>**  | Instance type for zero token node | i4i.large | SCT_ZERO_TOKEN_INSTANCE_TYPE_DB
 | **<a href="#user-content-use_zero_nodes" name="use_zero_nodes">use_zero_nodes</a>**  | If True, enable support in sct of zero nodes(configuration, nemesis) | false | SCT_USE_ZERO_NODES
+| **<a href="#user-content-scaling_action_type" name="scaling_action_type">scaling_action_type</a>**  | Refers to scaling task like scaleout, scalein etc that needs to be performed when storage reaches specific threshold| N/A | SCT_AUTO_SCALING_ACTION_TYPE
```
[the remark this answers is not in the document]
Lakshmipathi: ah, missed this. Now updated

full_storage_utilization_test.py
```diff
@@ -0,0 +1,146 @@
...
+
+class ScalingActionType(Enum):
+    SCALE_OUT = "scale_out"
+    SCALE_IN = "scale_in"
+
+class FullStorageUtilizationTest(ClusterTester):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.num_stress_threads = 10
+        self.sleep_time_before_scale = 120
```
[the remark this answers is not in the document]
Lakshmipathi: Now it uses sleep_seconds_before_scale

sdcm/utils/full_storage_utils.py
Lakshmipathi: ok, renamed total_small_ks as well.

sdcm/utils/full_storage_utils.py
Lakshmipathi: Yes, updated. Let me run few tests for verification.

Lakshmipathi:
ok, now pre-commit check passed.

full_storage_utilization_test.py
```diff
@@ -0,0 +1,146 @@
...
+           self.log.info(f"Invalid ActionType {self.scaling_action_type}")
+        DiskUtils.log_disk_usage(self.db_cluster.nodes)
+
+    def test_storage_utilization(self):
+        """
+        Write data until 90% disk usage is reached.
+        Sleep for 60 minutes.
+        Perform specific action.
+        """
+        self.run_stress(self.softlimit, sleep_time=self.sleep_time_fill_disk)
```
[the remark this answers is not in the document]
Lakshmipathi: From the test runs,  having `sleep` after softlimit before hardlimit helps to reach 90% more smoothly. Otherwise test ran into issues due to compaction.

full_storage_utilization_test.py
```diff
@@ -0,0 +1,146 @@
...
+        self.log.info("Dropping some data")
+        query = f"DROP KEYSPACE {keyspace_name}"
+        with self.db_cluster.cql_connection_patient(node) as session:
+           session.execute(query)
+           #node.run_nodetool(f"clearsnapshot")
+        DiskUtils.log_disk_usage(self.db_cluster.nodes)     
+
+    def perform_scale_in(self):
+        '''
+        If we are already at 90% disk utilization, first scale-out then
```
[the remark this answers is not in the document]
Lakshmipathi: will update the comment.

sdcm/utils/full_storage_utils.py
```diff
@@ -0,0 +1,117 @@
...
+        max_usage = 0
+        max_used = 0
+        for node in nodes:
+            info = DiskUtils.get_disk_info(node)
+            max_usage = max(max_usage, info["used_percent"])
+            max_used = max(max_used, info["used"])
+        return max_usage, max_used
+
+    @staticmethod
+    def calculate_target_used_size(nodes, target_percent):
```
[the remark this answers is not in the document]
Lakshmipathi: ok, added comments and renamed the function. Regarding mixed cluster, @cezarmoise is working on those cases, his PR should be handling them. (It will be ready for review once this gets merged).

full_storage_utilization_test.py
roydahan, line 1: Most of this file just duplicates functions we already have in performance_regression.py and could be integrated directly there.

Not only it saves duplications but also benefits from all automations like decorators we have for them.

swasik: Thanks @roydahan for pointing to this! @Lakshmipathi could you reuse the existing code where possible? And do not worry about the lost code you have to delete - I am sure it was great learning experience to write it and you will long benefit from it.

Lakshmipathi: Sure, will check performance_regression.py and re-use them where-ever required. Less code is better :)

Lakshmipathi:
Closing this PR in favor of https://github.com/scylladb/scylla-cluster-tests/pull/9787

Issue #9025, opened 2024-10-22
SCT defines the same aws_region for all nodes of multiDC cluster in scylla.yaml (kms_hosts:auto section)


mikliapko:
During nodes configuration (config_setup -> proposed_scylla_yaml) SCT defines the same `aws_region` for all nodes of multiDC cluster in scylla.yaml:
```
kms_hosts:
  auto:
    aws_region: us-east-1
    aws_use_ec2_credentials: true
    master_key: alias/testid-10224f75-e3b0-4879-a336-e1293bcb2d77
```
For my case, I have multiDC cluster with nodes in two regions - `us-east-1` and `us-west-2`. You can take a look into at scylla.yaml files for [run](https://argus.scylladb.com/tests/scylla-cluster-tests/6a48fe73-69f0-463e-986a-cc7c4ade6cdc) - `aws_region: us-east-1` is defined for all nodes.

I suppose this issue was introduced [here](https://github.com/scylladb/scylla-cluster-tests/commit/44ce384eaf6c7366bebe9210c2ba2b3fb2c1eca6#diff-d2939618bee56b9b221c3d9a1e8403478691c8196009d6dbfccc6c5f13ae1951R63) since changing the [line](https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/cluster.py#L518) from:
```
append_scylla_yaml["kms_hosts"][kms_host_name]["aws_region"] = self.vm_region
```
to:
```
append_scylla_yaml["kms_hosts"][kms_host_name]["aws_region"] = self.region
```
resolved configuration issue.

vponomaryov:
The `self.vm_region` is expected to be exactly "real" region of a VM, and each appropriate DB node python class must has proper reference.
When simulated multi-dc is disabled it must be equal to the `self.region` attribute.
So, the usage of attribute is correct, the bug is in the improper value stored there.

mikliapko:
> The `self.vm_region` is expected to be exactly "real" region of a VM, and each appropriate DB node python class must have proper reference. When simulated multi-dc is disabled it must be equal to the `self.region` attribute. So, the usage of attribute is correct, the bug is `improper value stored there`.

I see.
@vponomaryov Would you be able to prepare the fix for that in the coming days?

vponomaryov:
[...]

PR #7383, opened 2024-04-30
improvement(setup): Parallelize setup

soyacz:
In order to speed up test setup, made setup of db cluster, loaders,
monitors and oracle cluster to configure in parallel.
Verification and Argus related work was put into one-off background
thread.

Removed dependency to db_cluster for loaders as was not used in code (if needed we can add it in different place).

refs: https://github.com/scylladb/scylla-cluster-tests/issues/7242

### Testing

- [x] - [gemini test](https://argus.scylladb.com/test/5b50c349-c40c-440c-b727-f69c2e00acb4/runs?additionalRuns[]=1e020da2-3a83-4812-920d-3b51207c2693) - provision passed, errors due lost connectivity to loader after some time. More details in argus.
- [x] - provision test (jenkins pipeline, all backends passed including k8s functional tests)
- [x] - [k8s multitenant](https://argus.scylladb.com/test/80b4b4e8-be74-4439-98a8-889102638f31/runs?additionalRuns[]=e374e4cd-39a6-4822-bdee-08a7824c2e7c) - mutation write error during short period of time only. 
- [X] - [artifact debian](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-debian11-test/3/)
- [X] - [artifact ami](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-ami-test/3/)
- [x] - [ms ldap](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-100gb-4h-ldap-authorization-authenticator-ms-ad-test/3/)

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #9026, opened 2024-10-22
fix(kms): set proper node regions in multi-dc setups

vponomaryov:
The AWS-KMS code in DB node python class uses shared dictionary
from the DB cluster class for updating the KMS endpoint region.

It was not a problem when DB nodes setup was serial.
In this case shared object was changed by each node but had proper value in needed time frame.

After implementation of the parallel DB nodes setup (https://github.com/scylladb/scylla-cluster-tests/pull/7383) 
we started getting problems that only one state of that shared object was being applied for all nodes.

In single-dc setups everything was correct just because there was no diff among DB node's region names.
But in multi-dc setups values from DB nodes started being applied to each other.

So, fix it by just deep-copying that shared dictionary to avoid updates of a shared object.

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/9025

### Testing

- https://argus.scylladb.com/tests/scylla-cluster-tests/42b2d981-559b-4b64-b931-289e9f4ab928

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -15,6 +15,7 @@
 from __future__ import annotations
 
 import contextlib
+import copy
 import queue
 import logging
 import os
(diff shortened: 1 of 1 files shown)


vponomaryov:
Example of the failure: [enterprise-2024.2/tier1/longevity-multidc-schema-topology-changes-12h-test#7](https://argus.scylladb.com/tests/scylla-cluster-tests/d96e7c16-1105-4f5d-92bd-2b149705343f)
Fix tested here: https://argus.scylladb.com/tests/scylla-cluster-tests/42b2d981-559b-4b64-b931-289e9f4ab928

mikliapko (approved):
LGTM

soyacz (approved):
LGTM

vponomaryov:
Just for history:

Description of the PR is not complete.
The bug appeared not only after parallelization of DB nodes setup [1], but also after converting the `append_scylla_yaml` SCT config option from string type to dict [2].

[1] https://github.com/scylladb/scylla-cluster-tests/pull/7383
[2] https://github.com/scylladb/scylla-cluster-tests/pull/7554

PR #9073, opened 2024-10-28
fix(nodetool rebuild): use repair instead of rebuild if no tablets support

yarongilor:
if no tables support for nodetool rebuild, test should use an alternative action of repair. it should then disable load-balancing and repair all nodes in this datacenter. 
refs: https://github.com/scylladb/scylladb/issues/17575
refs: https://github.com/scylladb/scylladb/issues/20084#issuecomment-2323211550

### Testing

- [ ]

### Reminders


yarongilor:
@bhalevy , can you please advise, following https://github.com/scylladb/scylladb/issues/20084#issuecomment-2323211550 -

IIUC, in case https://github.com/scylladb/scylladb/issues/17852 is open all DC nodes should be manually repaired (not implemented here yet).
but then, otherwise, there is no need for a repair at all? or only repair target node?

and secondly, i'm not sure, is it right to backport this fix to 2024/6.x ?  (it may have an extensive impact on longevities and testing for this PR)

sdcm/nemesis.py
```diff
@@ -3976,8 +3976,15 @@ def start_and_interrupt_repair_streaming(self):
...
         self.target_node.wait_node_fully_start()
 
-        with adaptive_timeout(Operations.REBUILD, self.target_node, timeout=HOUR_IN_SEC * 48):
-            self.target_node.run_nodetool("rebuild", long_running=True, retry=0)
+        is_rebuild_supported = SkipPerIssues('scylladb/scylladb#17575', params=self.tester.params)
+        # If tablets in use and rebuild is not supported, running a DC repair instead.
+        with self.cluster.cql_connection_patient(self.target_node) as session:
+            if is_tablets_feature_enabled(session=session) and not is_rebuild_supported:
+                for node in [n for n in self.cluster.nodes if n.dc_idx == self.target_node.dc_idx]:
+                    node.run_nodetool(sub_cmd="repair")
```
fruch, line 3984: I would recommend doing long_running=True, retry=0

also maybe to consider hard timeout

fruch: also I'm not sure you have guarantee all the nodes in this DC are up and running...

yarongilor:
@bhalevy , @pehala , please advise - 
i see that nodetool rebuild continues to be tested in sct with no complains or reported issues.
on the other hand it is indeed skipped in dtest like:
https://github.com/scylladb/scylla-dtest/blob/1f70dde42ceeb2f2c45bbd773bb00bc7d8bb56ad/rebuild_test.py#L47
so if this workaround still relevant then i'll address comments and run tests for it.

sdcm/nemesis.py
yarongilor: fixed like:
```
                for node in [n for n in self.cluster.nodes if n.dc_idx == self.target_node.dc_idx and n.db_up()]:
                    node.run_nodetool(sub_cmd="repair", long_running=True, retry=0)
```
@fruch , since `long_running` support already mentioned, how about using scylla task manager for monitoring such commands progress and results?
There's already a dtest covering task manager in https://github.com/scylladb/scylla-dtest/pull/4957

pehala:
> @bhalevy , @pehala , please advise -
> i see that nodetool rebuild continues to be tested in sct with no complains or reported issues.
> on the other hand it is indeed skipped in dtest like:

I believe new tablet repair is being worked on, not sure about timelines

pehala:
@yarongilor 
new tablet repair (i.e. nodetool cluster repair) is currently being worked on, please inquire further and update the PR

yarongilor:
> @yarongilor new tablet repair (i.e. nodetool cluster repair) is currently being worked on, please inquire further and update the PR

`nodetool cluster repair` is now merged so we can consider switching to use it.

PR #9093, opened 2024-10-31
fix(logger_log_level): use map for logger_log_level in scylla.yaml

dimakr:
logger_log_level Scylla configuration parameter is a map of logger class to logger level (as described in https://opensource.docs.scylladb.com/stable/reference/configuration-parameters.html#confval-logger_log_level, for instance), so we should use dict instead of string for setting this in scylla_yaml.py.

### Testing

- [ ]


vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

PR #9094, opened 2024-10-31
fix(libssh2-runner): fix slow stderr read when no stdout

soyacz:
In case there's only stderr stream and no data on stdout, stderr stream processing waits for stdout processing timeout (0.5s by default). This causes long processing of many stderr lines as this delay happens for each line.

Fix by verifying first if given stream has data and proceed to next if not.

refs: https://github.com/scylladb/java-driver/issues/258

### Testing

- [X] - tested locally against remote machine

### Reminders


sdcm/remote/libssh2_client/__init__.py
```diff
@@ -440,25 +440,31 @@ def _process_output(  # pylint: disable=too-many-arguments, too-many-branches
...
                     stdout = reader.stdout.get(timeout=timeout_read_data_chunk)
                     data = stdout.decode(encoding, errors='ignore') + '\n'
                     stdout_stream.write(data)
                     for watcher in watchers:
                         watcher.submit_line(data)
+                    data_processed = True
                 except Exception:  # pylint: disable=broad-except  # noqa: BLE001
                     pass
-            if stderr_stream is not None:
+            if stderr_stream is not None and reader.stderr.qsize():
```
vponomaryov, line 455: This is redundant change, the next line does the same check.

sdcm/remote/libssh2_client/__init__.py
```diff
@@ -440,25 +440,31 @@ def _process_output(  # pylint: disable=too-many-arguments, too-many-branches
...
                         stderr = reader.stderr.get(timeout=timeout_read_data_chunk)
                         data = stderr.decode(encoding) + '\n'
                         stderr_stream.write(data)
                         for watcher in watchers:
                             watcher.submit_line(data)
+                        data_processed = True
                     except Exception:  # pylint: disable=broad-except  # noqa: BLE001
                         pass
+            if not data_processed:  # prevent unbounded loop in case no data on the streams
+                sleep(timeout_read_data_chunk)
```
vponomaryov, line 467: Are we sure that the `timeout_read_data_chunk` is integer/float here?

vponomaryov (commented):
Have you tested it?

sdcm/remote/libssh2_client/__init__.py
soyacz: true, I didn't notice it's a bit different in stderr handler.
Removed duplication from the inside, making one indent less.

sdcm/remote/libssh2_client/__init__.py
soyacz: no, it can be `None`. I added proper condition.

soyacz:
> Have you tested it?

I reproduced and tested it with this snippet:
```
import logging
from sdcm.remote import RemoteLibSSH2CmdRunner

logging.basicConfig(level=logging.DEBUG, format='%(asctime)s - %(levelname)s - %(message)s')

ip = "34.246.191.181"
RemoteLibSSH2CmdRunner(hostname=ip, user="<username>", key_file="<key>").run("for i in {1..10}; do >&2 echo 'This is stderr line $i'; done")
```

and indeed it helped.

vponomaryov (approved):
LGTM

fruch:
@soyacz 

Let's also test it with the jobs it was happening, i.e. the 3h gce test, a couple of times

soyacz:
> @soyacz
> 
> Let's also test it with the jobs it was happening, i.e. the 3h gce test, a couple of times

Tested here: https://argus.scylladb.com/tests/scylla-cluster-tests/f56211a0-6946-42e6-ac74-03810637b85e and passes. Log messages thrown to stderr by c-s after completing workload is read fast.
As it was always reproduced I consider running it once to be enough.

fruch (approved):
LGTM

sdcm/remote/libssh2_client/__init__.py
```diff
@@ -440,25 +440,30 @@ def _process_output(  # pylint: disable=too-many-arguments, too-many-branches
             if perf_counter() > end_time:
                 reader.stop()
                 return False
-            if stdout_stream is not None:
+
+            data_processed = False
+            if stdout_stream is not None and reader.stdout.qsize():
                 try:
                     stdout = reader.stdout.get(timeout=timeout_read_data_chunk)
```
dkropachev, line 447: @soyacz , does ithis PR realy do what it says ?
I would expect that you would want to make `stdout.get` and `stderr,get` non-blocking, which this PR did not do.

soyacz: I believe `not realiable!` statement is for the size - it's not precise, just approximate - which we don't care, info that there's something is enough.
This PR indeed fixes the issue.

dkropachev: Good to hear that, but I still think it would be better to do `.get(timeout=-1)` instead checking on `.qsize()`.

soyacz: under the hood, `.get` also uses similar ways like `qsize` to get the content.

PR #9102, opened 2024-10-31
feature(raft): Rolling restart raft topology coordinator node

aleksbykov:
When current topology coordinator is not available, new round
of election of new coordinator should be started and new
raft topology coordinator node will be elected.    

Added new function to search current coordinator node
Added new nemesis to Rolling restart of elected coordinator node

### Testing

- [ ]

### Reminders


sdcm/nemesis.py
```diff
@@ -5151,6 +5151,35 @@ def disrupt_disable_binary_gossip_execute_major_compaction(self):
             self.target_node.restart_scylla_server()
             raise
 
+    def disrupt_rolling_restart_topology_coordinator_node(self):
```
fruch, line 5154: the name rolling restart doesn't related to what going on with this nemesis

it's multiple restarts of the same node.

rolling means, all of the cluster is being restarted (one by one, hence where the rolling comes from)

sdcm/utils/raft/common.py
```diff
@@ -44,6 +50,27 @@ def validate_raft_on_nodes(nodes: list[BaseNode]) -> None:
     LOGGER.debug("Raft is ready!")
 
 
+def get_topology_coordinator_node(cluster: BaseScyllaCluster) -> BaseNode:
+    active_nodes = cluster.get_nodes_up_and_normal()
```
fruch, line 54: maybe this function should be guarded with ` self.target_node.raft.is_consistent_topology_changes_enabled` ?

to make sure it's not getting called when raft topology is disabled ?

fruch (changes requested):
Name of this nemesis should have `rolling` in it.

sdcm/utils/raft/common.py
aleksbykov: added

aleksbykov:
@soyacz , @fruch can you take a look?

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,35 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+    def disrupt_serial_restart_elected_topology_coordinator(self):
+        """ Rolling restart of topology coordinator node,
+        should trigger new coordinator node election
+        """
+        if not self.target_node.raft.is_consistent_topology_changes_enabled:
+            raise UnsupportedNemesis("Consistent topology changes feature is disabled")
+
+        self.use_nemesis_seed()
+        num_of_restarts = random.randint(1, len(self.cluster.nodes))
+        election_wait_timeout = random.choice([1, 5, 10, 15])
```
soyacz, line 5258: let's log debug these params, so it's clear how many restarts and timeout was selected.
In later message about random wait may be shortened a bit then.

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,35 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+
+        for _ in range(num_of_restarts):
+            coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
+            self.log.debug("Coordinator node: %s, %s", coordinator_node, coordinator_node.name)
+            if coordinator_node != self.target_node:
+                self.set_current_running_nemesis(coordinator_node)
```
soyacz, line 5264: I think we shouldn't care if it's target_node or not. Target node should be unset (with `unset_current_running_nemesis`) before loop starts - as it's not being target node anymore.

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,35 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+            self.log.debug("Wait random timeout %s to new coordinator will be elected", election_wait_timeout)
+            time.sleep(election_wait_timeout)
+            new_coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
+            self.log.debug("Coordinator node: %s, %s", new_coordinator_node, new_coordinator_node.name)
+            coordinator_node.start_scylla()
+            assert coordinator_node != new_coordinator_node, \
+                f"New coordinator node was not elected while old one {coordinator_node.name} was stopped"
+            self.unset_current_running_nemesis(coordinator_node)
+            if not self.target_node.running_nemesis:
+                self.set_current_running_nemesis(self.target_node)
```
soyacz, line 5276: I think it's redundant. target nodes change for each loop iteration

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,35 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+        for _ in range(num_of_restarts):
+            coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
+            self.log.debug("Coordinator node: %s, %s", coordinator_node, coordinator_node.name)
+            if coordinator_node != self.target_node:
+                self.set_current_running_nemesis(coordinator_node)
+
+            coordinator_node.stop_scylla()
+            self.log.debug("Wait random timeout %s to new coordinator will be elected", election_wait_timeout)
+            time.sleep(election_wait_timeout)
+            new_coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
```
soyacz, line 5269: in the case of running parallel nemesis this one may take the parallel nemesis target node and break it.
I think we should verify if new coordinator node is not being a target node already and wait until it does not (or just stop this nemesis and go to another one with proper log message).

sdcm/utils/raft/common.py
```diff
@@ -44,6 +50,29 @@ def validate_raft_on_nodes(nodes: list[BaseNode]) -> None:
     LOGGER.debug("Raft is ready!")
 
 
+def get_topology_coordinator_node(cluster: BaseScyllaCluster) -> BaseNode | None:
+    active_nodes: list[BaseNode] = cluster.get_nodes_up_and_normal()
+    if not active_nodes[0].raft.is_cluster_topology_consistent():
+        return
```
soyacz, line 56: wasn't @temichus intention to avoid such if statements? Shouldn't it go under raft object?

soyacz:
> [ Passed for PR ] (https://argus.scylladb.com/tests/scylla-cluster-tests/aa2a2300-e6bc-4d36-8081-b35ff7517523) - job failed due to cassandra, Nemesis functionality passed

Ins't it a Scylla issue?

sdcm/nemesis.py
aleksbykov: done

sdcm/nemesis.py
aleksbykov: remove it

sdcm/nemesis.py
aleksbykov: removed

sdcm/nemesis.py
aleksbykov: Set to skip nemesis if topology coordinator is busy with another nemesis

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,37 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+            raise UnsupportedNemesis("Consistent topology changes feature is disabled")
+
+        self.use_nemesis_seed()
+        num_of_restarts = random.randint(1, len(self.cluster.nodes))
+        self.log.debug("Number of serial restart of topology coordinator: %s", num_of_restarts)
+        election_wait_timeout = random.choice([1, 5, 10, 15])
+        self.log.debug("Wait new topology coordinator election timeout: %s", election_wait_timeout)
+        self.unset_current_running_nemesis(self.target_node)
+        for _ in range(num_of_restarts):
+            coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
```
soyacz, line 5263: could reuse `new_coordinator_node` with:
`coordinator_node = new_coordinator_node or get_topology_coordinator_node(cluster=self.cluster) `

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,37 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+
+            coordinator_node.stop_scylla()
+            self.log.debug("Wait random timeout %s to new coordinator will be elected", election_wait_timeout)
+            time.sleep(election_wait_timeout)
+            new_coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
+            self.log.debug("New coordinator node: %s, %s", new_coordinator_node, new_coordinator_node.name)
+            coordinator_node.start_scylla()
+            assert coordinator_node != new_coordinator_node, \
+                f"New coordinator node was not elected while old one {coordinator_node.name} was stopped"
+            self.unset_current_running_nemesis(coordinator_node)
```
soyacz, line 5278: let's unset before assert

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,37 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+        self.log.debug("Number of serial restart of topology coordinator: %s", num_of_restarts)
+        election_wait_timeout = random.choice([1, 5, 10, 15])
+        self.log.debug("Wait new topology coordinator election timeout: %s", election_wait_timeout)
+        self.unset_current_running_nemesis(self.target_node)
+        for _ in range(num_of_restarts):
+            coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
+            if coordinator_node.running_nemesis:
+                self.log.debug("The coordinator node is busy with nemesis %s. Skip nemesis", coordinator_node.name)
+                raise UnsupportedNemesis(f"Coordinator node is busy with nemesis {coordinator_node.name}")
+            self.set_current_running_nemesis(coordinator_node)
```
soyacz, line 5267: while doing it, also do `self.target_node = coordinator_node`, so when nemesis fails at any point next target_node selection will do automatically unset_current_running_nemesis

soyacz (commented):
Few more small issues, otherwise LGTM.

if we backport it, it will shuffle all nemesis lists on backported branches. Do we want that backported? cc @roydahan

sdcm/nemesis.py
aleksbykov: i don't want to that way, because something could happened , and coordinator node will be reelected between iterations. Better to search again before restart

sdcm/nemesis.py
aleksbykov: agree. fixed

sdcm/nemesis.py
aleksbykov: done

aleksbykov:
> Few more small issues, otherwise LGTM.
> 
> if we backport it, it will shuffle all nemesis lists on backported branches. Do we want that backported? cc @roydahan

i think we can live with it on master only

aleksbykov:
Staging jobs were passed: 
 - https://argus.scylladb.com/tests/scylla-cluster-tests/652ce6e8-acee-443a-b98a-a00a815d29b2

aleksbykov:
@fruch @soyacz can you review?

soyacz (approved):
LGTM

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,37 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+            raise UnsupportedNemesis("Consistent topology changes feature is disabled")
+
+        self.use_nemesis_seed()
+        num_of_restarts = random.randint(1, len(self.cluster.nodes))
+        self.log.debug("Number of serial restart of topology coordinator: %s", num_of_restarts)
+        election_wait_timeout = random.choice([1, 5, 10, 15])
+        self.log.debug("Wait new topology coordinator election timeout: %s", election_wait_timeout)
+        self.unset_current_running_nemesis(self.target_node)
+        for _ in range(num_of_restarts):
+            self.target_node = coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
```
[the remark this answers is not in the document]
fruch: we have machinery for selecting target_node, all selection of target nodes should work in similar fashion

see the logic of `run_nemesis`

doing so in multiple setup, is prone to errors and can lead to selection of the same target nodes from multiple nemesis threads, every selection should go via the same lock we introduced for that purpose

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,38 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+            self.set_current_running_nemesis(coordinator_node)
+            self.log.debug("Coordinator node: %s, %s", coordinator_node, coordinator_node.name)
+
+            coordinator_node.stop_scylla()
+            self.log.debug("Wait random timeout %s to new coordinator will be elected", election_wait_timeout)
+            time.sleep(election_wait_timeout)
+            new_coordinator_node = get_topology_coordinator_node(cluster=self.cluster)
+            self.log.debug("New coordinator node: %s, %s", new_coordinator_node, new_coordinator_node.name)
+            coordinator_node.start_scylla()
+            self.unset_current_running_nemesis(coordinator_node)
```
fruch, line 5277: if there was a failure you won't reach this, and would let go of that node, no more nemesis would run on it

we should use context manager (similar to run_nemesis), or try/finally clause here.

sdcm/nemesis.py
fruch: @aleksbykov please update the PR headline and the commit headline, to match the name change

sdcm/nemesis.py
soyacz: but won't nemesis mashinery unset running nemesis upon finishing? (that's why need to re-set target node each loop)

sdcm/nemesis.py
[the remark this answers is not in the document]
soyacz: with `run nemesis` it's going to fail the test when coordinator node is one that currently runs nemesis due to `assert free_nodes, f"couldn't find nodes for running:`{nemesis_label}`, are all nodes running nemesis ?"`

aleksbykov:
i set to use run_nemesis method

aleksbykov:
@soyacz @fruch , Can you take a look?

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,38 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+        """ Serial restart of elected topology coordinator node,
+        should trigger new coordinator node election
+        """
+        if not self.target_node.raft.is_consistent_topology_changes_enabled:
+            raise UnsupportedNemesis("Consistent topology changes feature is disabled")
+
+        self.use_nemesis_seed()
+        num_of_restarts = random.randint(1, len(self.cluster.nodes))
+        self.log.debug("Number of serial restart of topology coordinator: %s", num_of_restarts)
+        election_wait_timeout = random.choice([1, 5, 10, 15])
```
roydahan, line 5260: There are several randoms here without a "seed".
If one case detects an issue, it would be very hard to reproduce it without controlling the seed.

sdcm/nemesis.py
```diff
@@ -5246,6 +5246,38 @@ def disrupt_grow_shrink_zero_nodes(self):
...
+        self.log.debug("Wait new topology coordinator election timeout: %s", election_wait_timeout)
+        for _ in range(num_of_restarts):
+            coordinator_node = get_topology_coordinator_node(self.target_node)
+            self.log.debug("Coordinator node: %s, %s", coordinator_node, coordinator_node.name)
+            self.unset_current_running_nemesis(self.target_node)
```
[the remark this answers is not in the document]
aleksbykov: @soyacz , i do that because target_node could be the topology coordinator, and nemesis will be aborted at first run. From other side, The target_node is already locked for nemesis and it could be safely used to find coordinator node.
in other hand there will be too many nested with .. statements

aleksbykov:
Job passed: https://argus.scylladb.com/tests/scylla-cluster-tests/8fa92dde-8660-4b51-9c71-483691bf29f6

sdcm/nemesis.py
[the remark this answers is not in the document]
soyacz: as for verification nodes we usually don't lock it, so no need to do it here too I suppose.
Anyway if you unset running nemesis from self.target_node for a while, it may be taken in that time by parallel nemesis, also self.target_node may happen to become coordinator, so it will fail in that case I suppose.
Better just select verification node randomly out of all nodes that are not currently running nemesis.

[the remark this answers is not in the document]
fruch: I think that's a mistake we don't lock them.

other nemesis can take them down, and we'll fail while doing any action with them,

every time we work with a node in nemesis code, the node should be marked, and when we mark, we should hold the lock.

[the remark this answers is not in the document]
soyacz: in this particular case - disruptive nemesis - no other nemesis should take it down (as we don't run 2 disruptive in parallel) and verification is non disruptive (and should be short). We may lock it for a short a while just to give a good example.
Anyway currently, self.target_node may be down due to this nemesis itself - so this must be changed.

aleksbykov:
Job passed: https://argus.scylladb.com/tests/scylla-cluster-tests/c9b31857-e67c-4122-9794-cec2be11bb7c

@soyacz @fruch Can you take a look again

sdcm/nemesis.py
fruch: we probably should refactor `use_nemesis_seed` to return a random object, and not assume it's setting the seed globally 

@soyacz one more item for refactoring list :)

sdcm/nemesis.py
```diff
@@ -380,6 +380,12 @@ def publish_event(self, disrupt, status=True, data=None):
         severity = Severity.NORMAL if status else Severity.ERROR
         DisruptionEvent(nemesis_name=disrupt, severity=severity, **data).publish()
 
+    def switch_target_node(self, node: BaseNode):
+        self.unset_current_running_nemesis(self.target_node)
+        self.set_current_running_nemesis(node)
+        with NEMESIS_TARGET_SELECTION_LOCK:
+            self.target_node = node
```
fruch, line 387: I think holding the lock once is a bit better, i.e. leave no chance to others to pick something else
```suggestion
    def switch_target_node(self, node: BaseNode):
         with NEMESIS_TARGET_SELECTION_LOCK:
              self.target_node = None
              node.running_nemesis = self.current_disruption
              self.target_node = node
```

fruch (approved):
LGTM

on nitpick on the switch target node locking

the unittest failure, is waivered from my POV (it's not related to those change, and started happening yesterday on multiple runs

sdcm/nemesis.py
soyacz: i think that it is not needed at all - `self.target_node = node` should not be race'y as is only changed by current nemesis (parallel nemesis does not change it as is different object).
and node. and better keep use `set_current_running_nemesis` as this is a subject of refactor and we might want to set things differently.

soyacz (approved):

aleksbykov:
Failed unittest: FAILED unit_tests/test_sct_events_grafana.py::TestGrafana::test_grafana - AssertionError: 1 != 0 which is not related to the PR

aleksbykov:
and locally this test is passed

soyacz:
> Failed unittest: FAILED unit_tests/test_sct_events_grafana.py::TestGrafana::test_grafana - AssertionError: 1 != 0 which is not related to the PR

I see this test recently got flaky, I rerun the pipeline and it looks this time will pass. @aleksbykov please raise sct issue for this flakiness.

fruch:
> > > > Failed unittest: FAILED unit_tests/test_sct_events_grafana.py::TestGrafana::test_grafana - AssertionError: 1 != 0 which is not related to the PR
> > > 
> > > 
> > > I see this test recently got flaky, I rerun the pipeline and it looks this time will pass. @aleksbykov please raise sct issue for this flakiness.
> > 
> > 
> > I have a guess, it's the pylint in older branches, now we are sharing the builders with them... we probably can lower it back to 1 builder per CI run, as we did before.
> 
> Do we need issue? 

yes

>  @soyacz @fruch can you approve and merge?

fruch (approved):
LGTM

soyacz (approved):

PR #9152, opened 2024-11-06
Manager test configuration updates

mikliapko:
Changes:
- Since Manager 3.4 has been released, added the links to corresponding repo/list files in manager_versions.yaml;
- Removed manager-related mergify backporting rules.
Starting from release 3.4 it was decided to stop supporting the separate release branch for Manager and keep only master.
For the last 9 months I'm working with Manager, there were no cases we were needed to execute different tests for different Manager versions. So, keeping two branches and constant backporting (+ possible conflicts resolution) seems redundant.
- Raised Scylla versions to more relevant for several jobs.


.github/mergify.yml
```diff
@@ -177,41 +177,3 @@ pull_request_rules:
           - branch-perf-v16
         assignees:
           - "{{ author }}"
-  - name: Automate backport pull request manager-3.2
```
fruch, line 180: Why just removing ? you want want to add 3.4 ?

mikliapko: We agreed with the team that starting from release 3.4 we stop supporting the separate release branch for Manager and keep only master. For the last 9 months I'm working with Manager, there were no cases we were needed to execute different tests for different Manager versions. So, keeping two branches and constant backporting (+ possible conflicts resolution) seems redundant.

fruch: @roydahan FYI

it's similar to what we did for scylla-operator in SCT, we stopped branching at some point.

what would you do for minor releases ? still run tests from master ? 
that would require much more attention to backward compatibility in the tests, which isn't that great now.

mikliapko: Yes, run everything from master.

Actually, I prefer the approach with separate branches because sometimes it might come in handy (it wasn't a case last time as I mentioned above but it may appear).
The only concern and reason we decided to switch, conflicts resolution during backporting. I got into such cases with previous 3.3 release where the `manager-3.3` branch was missing some commits from master on which my solution was based. I spent some time figuring out what I should backport to make it work but failed and left the branch outdated (if I'm not mistaken it was related to Argus charts).

Since I can't keep an eye on every PR which is being merged into master on the subject of missing backport manager label, we have to switch to this approach with single branch. If there's ever a need for a separate branch, we can always introduce it again.

vars/managerPipeline.groovy
```diff
@@ -123,7 +123,7 @@ def call(Map pipelineParams) {
                    name: 'scylla_mgmt_pkg')
 
             string(defaultValue: "${pipelineParams.get('target_manager_version', '')}",
-                   description: 'master_latest|3.2|3.1. Only for upgrade test',
+                   description: 'master_latest|3.3|3.2. Only for upgrade test',
```
vponomaryov, line 126: Probably small typo? Must be `3.4`?

vponomaryov (approved):
LGTM

vars/managerPipeline.groovy
mikliapko: Yes, makes sense to improve it since the upgrade can be run from the released 3.4 to latest, thanks

mikliapko: Fixed

vponomaryov (approved):
LGTM

.github/mergify.yml
fruch: when you introduce it, it hard to pick the right commit to cut from
cause you might be missing things related to manager, or to SCT stability on the other hand.

I guess with manager is relatively ease, cause the volume of related commits to manager is quite small.

fruch (approved):
LGTM

roydahan:
@mikliapko this one should have been backported to the same branches as this PR https://github.com/scylladb/scylla-cluster-tests/pull/9236/files.

Now all branches are broken.

Issue #9134, opened 2024-11-05
_rotate_kms_key: `SkipPerIssues` raises `AttributeError` as `self.params` object has no attribute 'scylla_version'


mikliapko:
Here is a traceback of failure:
```
2024-11-04 17:38:44.815: (ThreadFailedEvent Severity.ERROR) period_type=one-time event_id=91b83f10-bb92-45bc-a7db-6ff7f054ea0a: message='dict' object has no attribute 'scylla_version'
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/decorators.py", line 26, in wrapper
return func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 4589, in _rotate_kms_key
if SkipPerIssues("https://github.com/scylladb/scylla-enterprise/issues/3896", self.params):
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/issues.py", line 171, in __bool__
return self.issues_opened() or self.issues_labeled()
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/issues.py", line 162, in issues_labeled
if self.params.scylla_version:
AttributeError: 'dict' object has no attribute 'scylla_version'
```

Argus: https://argus.scylladb.com/tests/scylla-cluster-tests/439869f1-5f7a-4840-b6ef-ad1f83ec465e

The issue is in this particular [code](https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/cluster.py#L4589):
```
                if SkipPerIssues("https://github.com/scylladb/scylla-enterprise/issues/3896", self.params):
                    self.log.warning("KMS encryption check is skipped due to the 'scylla-enterprise/issues/3896'")
                    continue
```
I suppose during the time the test waits to proceed with rotation the object `self.params` references to might have been changed somewhere else.

In general, this `if SkipPerIssues` clause can be completely removed since the issue it linked to is closed and doesn't have skip label.

@vponomaryov Please, let me know what do you think?
If you don't mind removing this `if` clause, I can remove it. Just decided to raise this question, since this `self.params` issue may exist somewhere else.

vponomaryov:
> Here is a traceback of failure:
> 
> ```
[...]

PR #9153, opened 2024-11-06
fix(_rotate_kms_key): remove SkipPerIssues since the ticket is resolved

mikliapko:
As per discussion in https://github.com/scylladb/scylla-cluster-tests/issues/9134, this clause can be removed as the ticket is resolved and doesn't have sct skip label (what could have made it being skipped). At the same, it led to unexpected failures when self.params appeared to be dict type instance instead of SCTConfiguration as expected.

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/older-enterprise-ami-sanity-test/5/

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4586,9 +4586,6 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
                     AwsKmsEvent(
                         message=f"Failed to rotate AWS KMS key for the '{kms_key_alias_name}' alias",
                         traceback=traceback.format_exc()).publish()
-                if SkipPerIssues("https://github.com/scylladb/scylla-enterprise/issues/3896", self.params):
-                    self.log.warning("KMS encryption check is skipped due to the 'scylla-enterprise/issues/3896'")
-                    continue
                 try:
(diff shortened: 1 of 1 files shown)


fruch:
@mikliapko 

yes we can remove this code, but I would suggest going to the root cause of why params argument wasn't set creating the cluster, we shouldn't let it happen.

fruch:
I think the actual fix was suppose to be:
```
s/self.params/db_cluster.params/
```

mikliapko:
> @mikliapko
> 
> yes we can remove this code, but I would suggest going to the root cause of why params argument wasn't set creating the cluster, we shouldn't let it happen.

Yes, totally agree with you and I raised a bug about it (https://github.com/scylladb/scylla-cluster-tests/issues/9134) that remains open.

mikliapko:
> I think the actual fix was suppose to be:
> 
> ```
> s/self.params/db_cluster.params/
> ```

okay, but still it looks like this peace of code I'm removing is not relevant anymore since the original issue is closed and not labeled as `sct-...-skip`.

vponomaryov (approved):
LGTM

roydahan (approved):

Issue #8946, opened 2024-10-08
Retire artifacts tests for amazon linux 2


fruch:
@roydahan asked for the removal of those jobs in:
https://github.com/scylladb/scylladb/issues/18631#issuecomment-2383749588

- [ ] remove from master
- [x] remove the trigger  - https://github.com/scylladb/scylla-pkg/pull/4566
- [ ] delete in jenkins / argus 
- [ ] Update Scylla documentation

roydahan:
@yaronkaikov Good opportunity to start practicing towards owning our support matrix.

yaronkaikov:
> [@yaronkaikov](https://github.com/yaronkaikov) Good opportunity to start practicing towards owning our support matrix.

Let start with passing the knowledge before you start assigning us the QA related tasks :-)

PR #9164, opened 2024-11-10
fix(artifacts): retire amazon2

yaronkaikov:
Based on scylladb/scylladb#18631 (comment)

Fixes: scylladb/scylla-cluster-tests#8946

Need https://github.com/scylladb/scylla-pkg/pull/4566 to get merged first


fruch:
@yaronkaikov maybe we should backport to release where it's failing ? i.e. 2024.2 ?

fruch (changes requested):
- `configurations/arm/amazon2.yaml` can be removed
- `sdcm/utils/distro.py` KNOWN_OS can be updated to remove `2` from `'("AMAZON", "amzn", ["2", "2023"], DistroBase.RHEL),`

yaronkaikov:
> * `configurations/arm/amazon2.yaml` can be removed
> * `sdcm/utils/distro.py` KNOWN_OS can be updated to remove `2` from `'("AMAZON", "amzn", ["2", "2023"], DistroBase.RHEL),`

Done

sdcm/cluster_aws.py
```diff
@@ -873,10 +872,6 @@ def traffic_control(self, tcconfig_params=None):
             self.remoter.run('sudo bash -cxe "%s"' % tc_command)
 
     def install_traffic_control(self):
```
Annamikhlin, line 874: seems `install_traffic_control` function is not needed any more

yaronkaikov: it is been used by `nemesis.py`

fruch: as yaron was saying, it is being using in nemesis.py.

and isn't related to this PR

fruch (approved):
LGTM

sdcm/cluster_aws.py
Annamikhlin: But now after this change the function is not installing anything... so the the is not exactly correct
and looking as the following:
```
    def install_traffic_control(self):
        return self.remoter.run("/sbin/tc -h", ignore_status=True).ok
```

Just checking and returns ok... 
and it was added only for amazon2...

yaronkaikov:
@roydahan if you decide to add `backport/2024.2` it must also be aligned with pkg, otherwise we will continue triggering this job

roydahan:
So please align, because I deleted these jobs from 2024.2

PR #9180, opened 2024-11-12
refactor(instance_provision_fallback): Change default to true in AWS

roydahan:
Recently we're facing too many provision failures of spot instances, hence we would like to change the default in AWS of instance_provision_fallback_on_demand from false to true which should retry the proviosn with on_demand.

### Testing

- [ ]

### Reminders

### DIFF
--- a/defaults/aws_config.yaml
+++ b/defaults/aws_config.yaml
@@ -1,6 +1,6 @@
 instance_provision: 'spot'
 spot_max_price: 0.80
-instance_provision_fallback_on_demand: false
+instance_provision_fallback_on_demand: true
 region_name:
   - eu-west-1
 user_credentials_path: '~/.ssh/scylla_test_id_ed25519'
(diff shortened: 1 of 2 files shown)


fruch (dismissed):
LGTM


don't know when last we tried in longevity and not artifact, but seems like it should be working

soyacz (commented):
It won't fix the problem you try to solve - `LongevityPipeline` overrides this setting.
Better would be to adjust the pipeline by overriding this setting like we do e.g. in `jenkins-pipelines/oss/longevity/longevity-100gb-4h.jenkinsfile`

roydahan:
removed the default value from LongevityPipeline, though didn't test yet if it works like that for boolean parameter.

fruch:
@roydahan tested that pipeline works as expected ?

roydahan:
> @roydahan tested that pipeline works as expected ?

Not yet.

PR #9188, opened 2024-11-13
fix(capacity_reservation): use it when adding nodes in runtime

fruch:
seems like by mistake the code was trying to read SCT configuration from the wrong place, cause capacity_reservation to be ignored when nodes are added during the test.

### Testing

- [x] no need of testing

### Reminders


soyacz (approved):
LGTM

Issue #9024, opened 2024-10-21
Hydra fails with KeyError: 'endpoint_resolver' from BOTO S3 Client


CodeLieutenant:
## Explanation:

There is an issue when `boto` client tries to connect from multiple threads to S3, to get the SSH keys needed for the machine. When `max_workers` is set to `1` the issue dissapears.

`sdcm/utils/keystore.py`
```diff
def sync(self, keys, local_path, permissions=0o777):
    """Syncs the local and remote S3 copies"""
-    with ThreadPoolExecutor(max_workers=len(keys)) as executor:
+    with ThreadPoolExecutor(max_workers=1) as executor:
        args = [(key, local_path, permissions) for key in keys]
        list(executor.map(lambda p: self.get_obj_if_needed(*p), args))
```

```
> hydra run-test longevity_test.LongevityTest.test_custom_time --backend aws --config test-cases/gemini/gemini-basic-3h.yaml

There is scylladb/hydra:v1.77-argus-0.12.10 in local cache, using it.
Going to run './sct.py  run-test longevity_test.LongevityTest.test_custom_time --backend aws --config test-cases/gemini/gemini-basic-3h.yaml'...
/usr/local/lib/python3.10/site-packages/paramiko/pkey.py:100: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "cipher": algorithms.TripleDES,
/usr/local/lib/python3.10/site-packages/paramiko/transport.py:259: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "class": algorithms.TripleDES,
logged in as ******
Traceback (most recent call last):
  File "/home/dmalusev/Code/ScyllaDB/scylla-cluster-tests/./sct.py", line 1866, in <module>
    cli.main(prog_name="hydra")
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1654, in invoke
    super().invoke(ctx)
  File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
[...]

PR #9193, opened 2024-11-13
fix(keystore): add a lock on boto3 client creation

fruch:
since the `KeyStore.sync()` call is start multiple threads that each start it's own boto3 client, we are running into cases we fail like:
```
self = <botocore.session.ComponentLocator object at 0x719afedcfb20>
name = 'credential_provider'
    def get_component(self, name):
        if name in self._deferred:
            factory = self._deferred[name]
            self._components[name] = factory()
>           del self._deferred[name]
E           KeyError: 'credential_provider'
```

so we are putting a lock ontop of client creation, to make sure we don't thread unsafe code of boto3 session creation in parallel

Fix: scylladb/scylla-cluster-tests#9024

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/keystore.py
+++ b/sdcm/keystore.py
@@ -14,6 +14,7 @@
 import os
 import json
 import hashlib
+import threading
 from typing import BinaryIO
 from concurrent.futures.thread import ThreadPoolExecutor
 from collections import namedtuple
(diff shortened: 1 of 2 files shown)


CodeLieutenant (approved):

vponomaryov (approved):
LGTM

Issue #8370, opened 2024-08-15
deprecate `stress_multiplier` options completely 


fruch:
we found out this options is causing more harm than good
1) it's replicate the exact name commands, which for write or mixed, generate override we don't always intend
2) it was proven (in several) that in c-s increasing load with more thread works better
3) this option doesn't work correctly with docker loaders 

given all of those, I would suggest removing this option from SCT
and as follow up adjusting cases that was using with other options we have (number of threads) 

there is usage of it, only in 3 config files, 
2 in cases plan to deprecate, 
and on in alternator perf case, that isn't used for quite some time

roydahan:
Once we deprecate the tests, It's fine to deprecate it.

fruch:
I decide not to deprecate this option, it is used directly in more tests.
and might be useful in some situation which aren't performance tests

PR #9218, opened 2024-11-14
fix(stress_threads): clearing `stress_multiplier` option, and `stress_num` arguments

fruch:
those option are not really optimal was to double load, since they are doing the same exact commands, we stopped using them in performance tests and we should stop using it across the board, hence it is removed.

Fixes: #8370

### Testing

- [ ]

### Reminders


fruch:
it's mid way, still need to clearout all of the `stress_num` arguments 
and there's a lot of it out there

fruch:
gave up on this one, leave it, since this option still needed

Issue #9133, opened 2024-11-05
Node failed to run because of  X.509 certificate error


amnonh:
Test: https://jenkins.scylladb.com/job/scylla-staging/job/amnon/job/longevity-100gb-4h-test/9/ failed

The reason, a node failed with the following error:
11:26:16  2024-11-05T09:20:31.616+00:00 longevity-100gb-4h-allow-pr-db-node-9fdeb40d-5      !ERR | scylla[6924]:  [shard  0:main] init - Startup failed: std::system_error (error GnuTLS:-62, Unknown Subject Alternative name in X.509 certificate.)

fruch:
@dimakr 

do we have enough logs to investigate this one ? 

i.e. printout of each certificate ? to see what we have missing here, and why ?

if not we should introduce those kind of logs, for catching this next time.

dimakr:
The reason for the issue is known and the fix https://github.com/scylladb/scylla-cluster-tests/pull/8066 addresses it. But the test was executed from the branch `allow_protobu` of the https://github.com/amnonh/scylla-cluster-tests SC fork, which misses the fix - the fix was merged on Jul 22 and the branch was synced with upstream on Jul 10.

But to facilitate certificates related issues analysis the PR https://github.com/scylladb/scylla-cluster-tests/pull/9219 introduces a change, so that TLS/SSL artifacts created during a test are collected (and published to Argus).

PR #9219, opened 2024-11-14
improvement(tls-certs): collect and keep TLS/SSL artifacts

dimakr:
Collect TLS/SSL artifacts (server, client certificates; CA certificate; etc.) and keep them after a test is finished, similarly to how logs are collected and published.
This will facilitate root cause analysis of SCT failures caused by certificate related issues (e.g. "Unknown Subject Alternative name in X.509 certificate" kind of errors, etc.).

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/9133

### Testing
- [x] :green_circle: [pr-provision-test with enabled encryption, certs are collected and published to argus](https://argus.scylladb.com/tests/scylla-cluster-tests/a429edb6-f820-49d0-a68e-bb7383daa891)
- [x] :green_circle: [pr-provision-test with no encryption enabled, certs are not collected/published](https://argus.scylladb.com/tests/scylla-cluster-tests/0b5adecf-1ccc-4147-bb19-c54de7641ef1)

### Reminders


sdcm/logcollector.py
```diff
@@ -1297,6 +1297,17 @@ def collect_logs(self, local_search_path: Optional[str] = None) -> list[str]:  #
...
 
+class CertsCollector(BaseSCTLogCollector):
+    log_entities = [
+        DirLog(name='ssl_conf/*', search_locally=True),
+    ]
+    cluster_log_type = 'ssl-conf'
+    cluster_dir_prefix = 'ssl-conf'
+
+    def collect_logs(self, local_search_path: Optional[str] = None) -> list[str]:
+        return super().collect_logs(local_search_path=local_search_path)
```
vponomaryov, line 1308: Why is it needed?

sdcm/tester.py
```diff
@@ -2986,6 +2991,7 @@ def tearDown(self):
         time.sleep(1)  # Sleep is needed to let final event being saved into files
         self.save_email_data()
         self.argus_collect_gemini_results()
+        self.collect_certs()
```
vponomaryov, line 2994: Shouldn't it be called before the line `2988` (`self.clean_resources()`)?

sdcm/logcollector.py
```diff
@@ -1297,6 +1297,17 @@ def collect_logs(self, local_search_path: Optional[str] = None) -> list[str]:  #
         return [s3_link]
 
 
+class CertsCollector(BaseSCTLogCollector):
+    log_entities = [
+        DirLog(name='ssl_conf/*', search_locally=True),
+    ]
+    cluster_log_type = 'ssl-conf'
+    cluster_dir_prefix = 'ssl-conf'
```
vponomaryov, line 1305: Typo? `ssl_conf`?

vponomaryov (changes requested):

sdcm/logcollector.py
dimakr: Removed. 
I was looking at/repeated another k8s related collector, but It is not needed indeed.

sdcm/tester.py
dimakr: Certs are collected from SCT runner, so cleaning cluster resources doesn't affect collecting certs.

fruch: we might also consider collecting from nodes, to validate we didn't mess-up transferring them to the correct nodes ?

vponomaryov (dismissed):
LGTM

sdcm/tester.py
vponomaryov: Gathering certs from nodes is also interesting, IMHO.
But we can do it later.

dimakr: Added gathering certificates and cqlshrc from nodes (we had an a failure once with the error `configparser.DuplicateOptionError: While reading from '/home/ubuntu/.cassandra/cqlshrc' [line  8]: option 'usercert' in section 'ssl' already exists`, so collecting cqlshrc would help in investigating, if it occurs again)

dimakr:
Failed unittest - `tests/test_base_version.py::TestBaseVersion::test_2024_2` - is not related to the change (it is probably a post-effect of promoting 2024.2 yesterday at the evening)

sdcm/tester.py
vponomaryov: So, probably now it should be moved before the step where DB nodes get cleaned up?

dimakr: But maybe it is better to move the 'collect_certs` method before `clean_resources` one. If it confused once, will confuse again. And rename it to smth like collect_ssl_conf

vponomaryov (changes requested):
Screenshot from 2024-11-20 19-27-53

Please, use DB node names for directories names like it is done in DB logs.

fruch (approved):
LGTM

even with the ip address, we don't need this that often

PR #9233, opened 2024-11-17
fix(test_parallelobject): replace `self.assertAlmostEqual` with  `pytest.approx`

fruch:
test was casting time into int, and then diffing with expected timeout that was causing cases it might fail, if test was slowed down even by a little bit.

also defined the num_workers, so it won't be automatically set by number of CPU the test machine has, and won't be more deterministic

### Testing

- [ ]

### Reminders

### DIFF
--- a/unit_tests/test_parallelobject.py
+++ b/unit_tests/test_parallelobject.py
@@ -4,6 +4,7 @@ import logging
 import random
 import concurrent.futures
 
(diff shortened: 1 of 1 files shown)


vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

Issue #9032, opened 2024-10-23
[modify_table_memtable_flush_period_in_ms] nemesis failed due to configuration issue


timtimb0t:
## Packages

Scylla version: `6.3.0~dev-20241018.b11d50f59191` with build-id `d5fe38f8fd12d9b834688320151f36fb8c1e050d`

Kernel Version: `6.8.0-1017-aws`

## Issue description

**modify_table_memtable_flush_period_in_ms** nemesis failed with error:
```
cassandra.protocol.ConfigurationException: <Error from server: code=2300 [Query invalid because of configuration issue] message="memtable_flush_period_in_ms must be 0 or greater than 60000">
```

No visible impact 

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 5 nodes (i4i.2xlarge)

Scylla Nodes used in this run:
- longevity-parallel-topology-schema--db-node-ebc54a43-9 (18.200.171.196 | 10.4.9.202) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-8 (54.194.129.37 | 10.4.11.252) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-7 (34.240.58.153 | 10.4.8.220) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-62 (54.155.22.121 | 10.4.9.164) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-61 (54.228.240.234 | 10.4.8.90) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-60 (54.78.237.111 | 10.4.8.36) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-6 (52.30.126.73 | 10.4.11.167) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-59 (34.250.116.214 | 10.4.9.39) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-58 (54.73.14.40 | 10.4.8.39) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-57 (99.81.108.213 | 10.4.10.175) (shards: 7)
- longevity-parallel-topology-schema--db-node-ebc54a43-56 (54.220.49.204 | 10.4.11.250) (shards: 7)
[...]

PR #9234, opened 2024-11-17
[modify table nemesis] memtable flush period

temichus:
update modify_table_memtable_flush_period_in_ms rand value to be be 0 or greater than 60000

according to commit https://github.com/scylladb/scylladb/commit/b965729f0a9784f43e673d50c4db0c20ca0c2b48

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/9032

### Testing

- [ ]

### Reminders


sdcm/nemesis.py
```diff
@@ -2797,7 +2797,7 @@ def modify_table_memtable_flush_period_in_ms(self):  # pylint: disable=invalid-n
             The number of milliseconds before Cassandra flushes memtables associated with this table.
             default: memtable_flush_period_in_ms = 0
         """
-        self._modify_table_property(name="memtable_flush_period_in_ms", val=random.randint(1, 5000))
+        self._modify_table_property(name="memtable_flush_period_in_ms", val=random.choice([0, random.randint(60000, 200000)]))
```
fruch, line 2800: what's the behavior in 0, it's like not setting it ? why we want to cover it ?

temichus: 0 means table with very low write rate can have memtables avoid flush for days due to low commit_log

denesb: `N == 0` -> no periodic memtable flush
`0 < N < 6000` -> value rejected
`6000 < N` -> memtables are flushed every N milliseconds

Issue #9027, opened 2024-10-23
sct docker backend deletes all docker instances on the system in teardown


asias:
It should only delete instances started by sct test, not all instances on the system.

```
< t:2024-10-23 02:04:54,597 f:remote_base.py  l:560  c:RemoteLibSSH2CmdRunner p:DEBUG > <172.18.0.5>: Running command "docker ps -a -q | xargs docker rm -f"...
```

asias:
Got another example:
```
< t:2024-10-23 07:00:20,328 f:cluster.py      l:1998 c:sdcm.cluster_docker  p:INFO  > Node perf-latency-nemesis-asias-monitor-node-da4ee3ad-0 [127.0.0.1 | 127.0.0.1]: Installing Scylla...                                                                                                                                                                                                                                                                                   We trust you have received the usual lecture from the local System                                                                                                                                                                     Administrator. It usually boils down to these three things:                                                                                                                                                                                                                                                                                                                                                                                                                       #1) Respect the privacy of others.                                                                                                                                                                                                     #2) Think before you type.                                                                                     
    #3) With great power comes great responsibility.        
```

fruch:
this one goes way back when YCSB was introduced for alternator 

https://github.com/scylladb/scylla-cluster-tests/pull/1790/commits/f971bb9eea2941b7d7c9c60a139c54aa3d8d5f56

that code was written assuming it runs on a loader node, so docker backend running locally wasn't considered.

we should overwrite `kill_docker_loaders` for docker backend, and kill more selectively based on the test_id label we should have to the docker run (and and we are missing them, we should introduce those labels for the stress docker instances)

dimakr:
@fruch
Not sure that the command `docker ps -a -q | xargs docker rm -f` deletes containers on the system as the only place where it is executed is
https://github.com/scylladb/scylla-cluster-tests/blob/1434c81d781993ee476ca3d7c382544b5f2fc750/sdcm/cluster.py#L5274
I.e. in case of docker backend it is requested to be executed in a loader container, not in the on the SCT runner (local host).

fruch:
> @fruch
> Not sure that the command `docker ps -a -q | xargs docker rm -f` deletes containers on the system as the only place where it is executed is
> https://github.com/scylladb/scylla-cluster-tests/blob/1434c81d781993ee476ca3d7c382544b5f2fc750/sdcm/cluster.py#L5274
[...]

PR #9235, opened 2024-11-18
fix(stress-threads): kill only labeled docker containers on loaders

dimakr:
There is no any filtering of docker instances, when killing containerized stress threads on loader nodes during teardown of a test. In the case of docker backend this can result in deleting all containers in the system.

The change fixes this by labeling stress thread related docker containers when they are created; and deleting only the labeled containers during the teardown.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/9027

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5269,9 +5269,12 @@ class BaseLoaderSet():
                                  str(loader), str(ex))
 
     def kill_docker_loaders(self):
+        test_id = self.tags.get('TestId')
         for loader in self.nodes:
             try:
-                loader.remoter.run(cmd='docker ps -a -q | xargs docker rm -f', verbose=True, ignore_status=True)
+                loader.remoter.run(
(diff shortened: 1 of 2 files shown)


fruch (approved):
LGTM

vponomaryov (approved):
LGTM

soyacz (approved):

soyacz:
@dimakr how did you test it?
backports?

dimakr:
> @dimakr how did you test it? backports?

As it is against docker backend I tested/executed it locally (pausing at the corresponding places to ensure that stress containers are labeled as expected and can be filtered by the labels). And ran `test-provision-docker` as a regression check.
Wrt to backports, from the original issue it is not clear which version it was failing for, but we can probably backport it 2024.x and 6.x branches, what do you think? (although it is docker backend specific, so not many people are probably using the backend, but still, it would nice to have the patch in some recent versions)

soyacz (approved):

Issue #9274, opened 2024-11-19
TypeError: can only concatenate str (not "int") to str:  cluster_max_size += nemesis_node_count


asias:
With the following yam config, sct fails to run

n_db_nodes: '2 2'
nemesis_add_node_cnt: 3
use_capacity_reservation: true

```

08:28:40    File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_config.py", line 2626, in update_argus_with_version
08:28:40      test_config.argus_client().submit_packages([package])
08:28:40    File "/usr/local/lib/python3.10/site-packages/argus/client/sct/client.py", line 110, in submit_packages
08:28:40      self.check_response(response)
08:28:40    File "/usr/local/lib/python3.10/site-packages/argus/client/base.py", line 68, in check_response
08:28:40      raise ArgusClientError(
08:28:40  argus.client.base.ArgusClientError: ('API Error encountered using endpoint: POST /api/v1/client/sct/ac947416-42a9-4922-8dc4-c59b6ad5cc0d/packages/submit', 'Run not found')
08:28:40  Provision syslog-ng logging service
08:28:40  New directory created: /home/ubuntu/sct-results/20241119-062840-212561
08:28:40  Symlink `/home/ubuntu/sct-results/latest' updated to `/home/ubuntu/sct-results/20241119-062840-212561'
08:28:44  syslog-ng listen on port 32768 (config: /tmp/syslog-ngejs6ug9i.conf)
08:28:44  Provision aws cloud resources
08:28:44  No capacity reservations found.
08:28:44  Traceback (most recent call last):
08:28:44    File "/home/ubuntu/scylla-cluster-tests/./sct.py", line 1881, in <module>
08:28:44      cli.main(prog_name="hydra")
08:28:44    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
08:28:44      rv = self.invoke(ctx)
08:28:44    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
08:28:44      return _process_result(sub_ctx.command.invoke(sub_ctx))
08:28:44    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
08:28:44      return ctx.invoke(self.callback, **ctx.params)
08:28:44    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
08:28:44      return __callback(*args, **kwargs)
[...]

PR #9285, opened 2024-11-19
fix(capacity-reservation): disable for multi-dc

soyacz:
multi-dc scenarios are not supported in capacity reservations. Disable it when multi-dc is used to not raise errors.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/9274

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/provision/aws/capacity_reservation.py
+++ b/sdcm/provision/aws/capacity_reservation.py
@@ -118,10 +118,12 @@ class SCTCapacityReservation:
     @staticmethod
     def is_capacity_reservation_enabled(params: dict) -> bool:
         """Returns True if capacity reservation is enabled."""
+        is_single_dc = str(params.get("n_db_nodes")).isdigit() or params.get('simulated_regions') > 0
         return (params.get("cluster_backend") == "aws"
                 and (params.get("test_id") or params.get("reuse_cluster"))
                 and params.get('use_capacity_reservation') is True
-                and params.get('instance_provision') == 'on_demand')
+                and params.get('instance_provision') == 'on_demand'
+                and is_single_dc)
 
     @classmethod
     def reserve(cls, params) -> None:


sdcm/provision/aws/capacity_reservation.py
```diff
@@ -118,10 +118,12 @@ def _get_supported_availability_zones(ec2, instance_types: List[str], initial_az
     @staticmethod
     def is_capacity_reservation_enabled(params: dict) -> bool:
         """Returns True if capacity reservation is enabled."""
+        is_single_dc = str(params.get("n_db_nodes")).isdigit()
```
vponomaryov, line 121: At first, the `simulated` multi-dc should work with the reservation.
At second, I believe it must be done on the config initialization step by checking that `use_capacity_reservation` is `False` when real multi-dc is enabled.

vponomaryov (changes requested):

sdcm/provision/aws/capacity_reservation.py
soyacz: the best would be if multi-dc scenarios were supported. 
This PR was aiming rather failure that happens when someone tries to use cr with multi-dc - to avoid it and show clearly that is not supported.
I think we could raise separate issue for that support and plan it if needed.

soyacz: I added verification if is simulated - in that case it's treated as single dc.

vponomaryov (approved):
LGTM

sdcm/provision/aws/capacity_reservation.py
soyacz: https://github.com/scylladb/scylla-cluster-tests/issues/9317
fallback to regions is already done, CR's are tried over all regions in case of capacity issues.

dimakr (approved):
LGTM

PR #9291, opened 2024-11-19
feature(dependabot): scan docker based loaders for updates

fruch:
since dependabot is expecting very specific format of yaml files it can find and update:
* name starts with `values_`
* docker image need to be under `image` key

since we use same dockerhub repo for most of the loader images and dependabot only handle update of one of them, we had to split it to multiple files that would be scan seperatly. `sct_config` would collect all of those back into the configuration in the same structure it was until this change

### Testing

- [ ]

### Reminders


defaults/test_default.yaml
```diff
@@ -205,19 +205,7 @@ aws_fallback_to_next_availability_zone: false
...
-  cassandra-stress: 'scylladb/cassandra-stress:3.13.0'
-  scylla-bench: 'scylladb/hydra-loaders:scylla-bench-v0.1.24'
-  gemini: 'scylladb/hydra-loaders:gemini-v1.8.6'
-  alternator-dns: 'scylladb/hydra-loaders:alternator-dns-0.1'
-  cdc-stresser: 'scylladb/hydra-loaders:cdc-stresser-20210630'
-  kcl: 'scylladb/hydra-loaders:kcl-jdk8-20210526-ShardSyncStrategyType-PERIODIC'
-  harry: 'scylladb/hydra-loaders:cassandra-harry-jdk11-20220816'
-  latte: 'scylladb/hydra-loaders:latte-0.28.1-scylladb'
-  cql-stress-cassandra-stress: 'docker.io/scylladb/hydra-loaders:cql-stress-cassandra-stress-20240718'
+stress_image: {}
```
soyacz, line 208: let's add comment where are defaults set

sdcm/sct_config.py
```diff
@@ -1797,6 +1797,9 @@ def __init__(self):  # noqa: PLR0912, PLR0914, PLR0915
             backend_config_files += self.defaults_config_files[str(backend)]
         self.multi_region_params = self.per_provider_multi_region_params.get(str(backend), [])
 
+        # load docker images defaults
+        self.load_docker_images_defaults()
```
soyacz, line 1801: won't this override image provided by `SCT_STRESS_IMAGE.cassandra-stress=...`?

fruch: it's before env variables are evaluated

defaults/test_default.yaml
fruch: all around 😭

defaults/docker_images/latte/values_latte.yaml
```diff
@@ -0,0 +1,2 @@
+latte:
+  image: 'scylladb/hydra-loaders:latte-0.26.1-scylladb'
```
vponomaryov, line 2: Wrong version.

defaults/test_default.yaml
vponomaryov: Agree that comment will be useful.

.github/dependabot.yml
```diff
@@ -14,3 +14,27 @@ updates:
...
+      - "/defaults/docker_images/alternator-dns"
+      - "/defaults/docker_images/cassandra-stress"
+      - "/defaults/docker_images/cdc-stresser"
+      - "/defaults/docker_images/cql-stress-cassandra-stress"
+      - "/defaults/docker_images/gemini"
+      - "/defaults/docker_images/harry"
+      - "/defaults/docker_images/kcl"
+      - "/defaults/docker_images/latte"
+      - "/defaults/docker_images/nosqlbench"
+      - "/defaults/docker_images/ndbench"
```
vponomaryov, line 29: If it was meant to be in alphabetical order, then `ndbench` must be before `nosqlbench`.

.github/dependabot.yml
```diff
@@ -14,3 +14,27 @@ updates:
       - "scylladb/qa-maintainers"
     assignees:
       - "scylladb/qa-maintainers"
+
+  - package-ecosystem: "docker"
```
vponomaryov, line 18: s/docker/docker-loaders/ ?

soyacz (dismissed):
LGTM

vponomaryov (changes requested):
Latte version is wrong

.github/dependabot.yml
CodeLieutenant: it's a package ecosystem, just to indicate which `package manager` should be used.

[Dependabot Docs](https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#package-ecosystem)

CodeLieutenant (dismissed):

.github/dependabot.yml
fruch: argue with copilot :)

defaults/docker_images/latte/values_latte.yaml
fruch: update to the correct image

fruch:
backporting it, even that would update all of the images. but I think that should be safe enough for most of them.

CodeLieutenant (approved):

soyacz (approved):
LGTM

soyacz:
@vponomaryov please approve if latte version is now fine and LGTY

vponomaryov (approved):
LGTM

PR #9298, opened 2024-11-20
Make `test_restore_benchmark` results comparable

kreuzerkrieg:
While we currently have a test for `restore` benchmarking, the configurations are set up in such a way that the results from different runs are not comparable. Ideally, restoring 1TB of data should take twice as long as restoring 0.5TB. However, due to the use of different instance types in each manager configuration, this expectation is not met. Additionally, ScyllaDB's autocompaction can interfere with the restore process, so the test should explicitly disable it to ensure more consistent and deterministic results.

Argus results
[500GB run](https://argus.scylladb.com/tests/scylla-cluster-tests/33b6f7ab-1221-44f2-9b98-e9d2327b10e2)
|  | restore time [s] | download bandwidth [MiB/s/shard] | l&s bandwidth [MiB/s/shard] | repair time [s] | total [s] |
| --- | --- | --- | --- | --- | --- |
| 500gb_1t_ics | $${\color{red}00:57:09}$$ | $${\color{green}17.95}$$ | $${\color{green}7.46}$$ | $${\color{red}00:18:06}$$ | $${\color{red}01:15:15}$$ |

[1TB run](https://argus.scylladb.com/tests/scylla-cluster-tests/54c5741d-0469-4e13-b26f-849c6ff72f29)
|  | restore time [s] | download bandwidth [MiB/s/shard] | l&s bandwidth [MiB/s/shard] | repair time [s] | total [s] |
| --- | --- | --- | --- | --- | --- |
| 1tb_1t_ics | $${\color{red}02:07:50}$$ | $${\color{green}17.46}$$ | $${\color{green}6.48}$$ | $${\color{red}00:43:28}$$ | $${\color{red}02:51:18}$$ |

Also, [additional results](https://docs.google.com/spreadsheets/d/1ujb7ojG6iUEmrP4ICAjY-Hqd1bPs0W0SejfCoC1PJ8k/edit?gid=0#gid=0) with 3 x i4i.4xlarge cluster

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -572,7 +572,7 @@ class ManagerTestFunctionsMixIn(
         restore_task.wait_and_get_final_status(step=30, timeout=timeout)
         assert restore_task.status == TaskStatus.DONE, f"Restoration of {snapshot_tag} has failed!"
         InfoEvent(message=f'The restore task has ended successfully. '
-                          f'Restore run time: {restore_task.duration}.').publish()
+                  f'Restore run time: {restore_task.duration}.').publish()
         if restore_schema:
             self.db_cluster.restart_scylla()  # After schema restoration, you should restart the nodes
         return restore_task
@@ -1441,6 +1441,12 @@ class ManagerRestoreBenchmarkTests(ManagerTestFunctionsMixIn):
         5. Run the verification read stress to ensure the data is restored correctly.
         """
         self.run_prepare_write_cmd()
+
+        compaction_ops = CompactionOps(cluster=self.db_cluster)
+        #  Disable keyspace autocompaction cluster-wide since we dont want it to interfere with our restore timing
+        for node in self.db_cluster.nodes:
+            compaction_ops.disable_autocompaction_on_ks_cf(node=node)
+
         manager_tool = mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0])
         mgr_cluster = self.ensure_and_get_cluster(manager_tool)
 
@@ -1489,6 +1495,11 @@ class ManagerRestoreBenchmarkTests(ManagerTestFunctionsMixIn):
                                       restore_schema=True, location_list=location)
         for ks_name in snapshot_data.keyspaces:
             self.set_ks_strategy_to_network_and_rf_according_to_cluster(keyspace=ks_name, repair_after_alter=False)
+
+        compaction_ops = CompactionOps(cluster=self.db_cluster)
+        # Disable keyspace autocompaction cluster-wide since we dont want it to interfere with our restore timing
+        for node in self.db_cluster.nodes:
+            compaction_ops.disable_autocompaction_on_ks_cf(node=node)
 
         if restore_outside_manager:
             self.log.info("Restoring the data outside the Manager")


mikliapko:
@kreuzerkrieg 
Could you please attach the run where your changes were tested?

kreuzerkrieg:
> @kreuzerkrieg Could you please attach the run where your changes were tested?

sure, will create new clean runs a add to the description shortly

mgmt_cli_test.py
```diff
@@ -1395,6 +1395,12 @@ def test_backup_and_restore_only_data(self):
         5. Run the verification read stress to ensure the data is restored correctly.
         """
         self.run_prepare_write_cmd()
+
+        compaction_ops = CompactionOps(cluster=self.db_cluster)
+        #  Disable keyspace autocompaction cluster-wide since we dont want it to interfere with our restore timing
+        for node in self.db_cluster.nodes:
+            compaction_ops.disable_autocompaction_on_ks_cf(node=node)
+
```
mikliapko, line 1403: That's something I missed during initial implementation.
Can be a good improvement for restore benchmark.
@Michal-Leszczynski What do you think?

configurations/manager/1TB_dataset.yaml
```diff
@@ -10,5 +10,5 @@ stress_read_cmd: ["cassandra-stress read cl=ONE n=268435456 -schema 'replication
                   "cassandra-stress read cl=ONE n=268435456 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=536870913..805306368",
                   "cassandra-stress read cl=ONE n=268435456 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=805306369..1073741824"]
 
-instance_type_db: 'i3en.2xlarge'
+instance_type_db: 'i3en.3xlarge'
```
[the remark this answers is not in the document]
kreuzerkrieg: > Why did you choose this specific instance type?

It is the instance type used for 2TB run, so if you want to compare three configs; 0.5, 1 and 2TB you have to use the same instance type

> cost optimization

from 0.9 to 1.3 USD per instance, I doubt we run it so frequently we will notice the difference in our AWS bill, moreover, the beefier instance should acomplish the same task in shorter time negating the cost difference

> Also, you can consider an option to specify instance type on jenkins pipeline layer (not affecting existing configuration files)

This would be even more convenient change, do we have an example PR how to do that?

mgmt_cli_test.py
kreuzerkrieg: But we have to remember that after restore the SM will enable the autocompaction, so if we want to run and measure read stress then we have to disable it once again

mikliapko: we don't collect any post-restore read stress metrics in these tests AFAIR

configurations/manager/1TB_dataset.yaml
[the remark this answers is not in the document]
mikliapko: Ah, I see that there is no separate argument for instance_type in Manager job.
But we can achieve the same result by setting required type `SCT_INSTANCE_TYPE_DB=<type>` in `extra_environment_variables` parameter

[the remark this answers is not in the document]
kreuzerkrieg: ok, will take a look at it

mgmt_cli_test.py
regevran: I think it worth explicitly disabling as a precondition to the test. It reminds everyone about our assumptions, and makes no room for future mistakes.

regevran (approved):

mgmt_cli_test.py
kreuzerkrieg: There is no such a thing. every ks created during the test will be not affected by whatever one has disabled before

configurations/manager/1TB_dataset.yaml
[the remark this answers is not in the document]
kreuzerkrieg: looks like `SCT_INSTANCE_TYPE_DB=<type>` works as expected
https://argus.scylladb.com/tests/scylla-cluster-tests/c4371f87-816c-4dce-9ca9-db289aeba30d
reverting changes on yaml files

mgmt_cli_test.py
regevran: I argued in favor leaving explicit disable as you wrote them. Am I missing something?

Michal-Leszczynski (approved):

kreuzerkrieg:
@scylladb/qa-maintainers could you push it forward please?

mikliapko (approved):
LGTM

soyacz (approved):
LGTM

Issue #8752, opened 2024-09-18
SCT for backup


regevran:
Measure backup process performance - how long does it take to backup.
Measure user-queries during backup - what is the impact of backup on user experience.
Measures inside Scylla.
SM tests may or may not be combined with the tests in this issue.

regevran:
> I think this is on @mikliapko plate as part of SM tests.

I updated the description, does it make more sense now?

soyacz:
> > I think this is on @mikliapko plate as part of SM tests.
> 
> I updated the description, does it make more sense now?

@mikliapko ^^

mikliapko:
@regevran Could you please provide some context on this issue?
- Where it comes from;
- How soon should we have such a test;
- Is there any preferred Scylla + dataset configuration to test.

Currently in Manager test scope we don't have backup performance tests, only for restore.

cc: @rayakurl

regevran:
> Where it comes from;

2024 Q4 is planned to finish with better backup and restore capabilities.
This issue is to make sure we get there with measurements.

> How soon should we have such a test;

@kreuzerkrieg is assigned to this and will start shortly working on the details.

> Is there any preferred Scylla + dataset configuration to test.

Yes. I think the focus should be on a large, full cluster backups. And we should measure heavy user loads during the backup.
The numbers (how large, what does it mean heavy load, etc.) to be discussed.

rayakurl:
@regevran - can you please add more details regarding how @mikliapko should assist?

regevran:
> @regevran - can you please add more details regarding how @mikliapko should assist?

@kreuzerkrieg is collecting information including the requested support 😸.

kreuzerkrieg:
@mikliapko I have a couple of questions.
Took a look at [this file](https://github.com/scylladb/scylla-cluster-tests/blob/master/mgmt_cli_test.py) and looks like we already have backup/restore test that measures performance of it. At least this is what the comment [states](https://github.com/scylladb/scylla-cluster-tests/blob/master/mgmt_cli_test.py#L1381C12-L1381C110)

> The test is extensively used for restore benchmarking purposes and consists of the following steps

1. Do we track results of this tests? is it reported and can be reviewed somehow?
2. Second, do we have any benchmarking of read performance during the backup? something comparable, like, we have the bench results with and without backup running?
3. Third, do we have bechmarking performance of reading from  a table residing on the same cluster where another table being restored?

[...]

PR #9222, opened 2024-11-15
refactor(manager): reorganize tests in mgmt_cli_test.py

mikliapko:
The PR aims to bring some logical structure into mgmt_cli_test.py (the majority of Manager tests).

Some details on the changes:
- tests grouped in separated classes depending on the functionality they test (ManagerBackupTests, ManagerRestoreTests, etc);
- common helper methods (load generators, database handlers) moved to separate classes (DatabaseOperations, ClusterOperations, etc);
- all Tests classes derive from ManagerTestFunctionsMixIn (kind of coordinator class) which contains some general test methods and inherits helper Operations classes;
- test_name parameter updated for all affected jenkinsfiles;
- some classes and methods renamed to correlate with changes above.

No logical changes have been done to methods, classes, tests, only reorganization across the test file.

### Testing

- [ ] [Restore benchmark](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/backup-restore-custom-dataset-test/89/)
- [ ] [Installation test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/rocky8-installation-test/7/)
- [ ] [Backup test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/sct-feature-test-backup/24/)
- [ ] [Sanity test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu22-sanity-test/43/)
- [ ] [Upgrade test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master/job/ubuntu24-upgrade-test/18/)

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

PR #9307, opened 2024-11-20
Add backup benchmarking under read stress

kreuzerkrieg:
Introduce the `test_backup_benchmark` test, which measures backup time under read stress conditions. This test performs multiple actions to consolidate all necessary data into a single table. Initially, it runs and measures the backup process, followed by the read stress test. Finally, it executes both processes asynchronously to observe how the performance of reading and backing up degrades.

More metrics could be added after https://github.com/scylladb/scylla-manager/issues/4123 is completed
Test should be re-ran after https://github.com/scylladb/scylla-manager/issues/4125 is completed

Argus results:
For [100GB run](https://argus.scylladb.com/tests/scylla-cluster-tests/83b4a96b-28bb-4501-8116-01f909250c03)
|  | backup time [s] |
| --- | --- |
| Backup during read stress | 00:18:48 |
| Backup | 00:10:18 |

|  | read time [s] |
| --- | --- |
| Read stress | 00:11:29 |
| Read stress during backup | 00:11:31 |

fixes: #8752

### DIFF
--- a/configurations/manager/100GB_dataset.yaml
+++ b/configurations/manager/100GB_dataset.yaml
@@ -1,0 +1,14 @@
+test_duration: 120
+
+prepare_write_cmd: [ "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..26214400",
+                     "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
+                     "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
+                     "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
+
+stress_read_cmd: [ "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..26214400",
+                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
+                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
+                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
+
+instance_type_db: 'i3en.3xlarge'
+instance_type_loader: 'c6i.xlarge'
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -16,6 +16,8 @@
 
 # pylint: disable=too-many-lines
 import random
+import threading
+from enum import Enum
 from pathlib import Path
 from functools import cached_property
 import re
@@ -26,11 +28,14 @@ from dataclasses import dataclass
 
 import boto3
 import yaml
+from docker.errors import InvalidArgument
 
 from invoke import exceptions
 
+from argus.client.generic_result import Status
 from sdcm import mgmt
-from sdcm.argus_results import send_manager_benchmark_results_to_argus
+from sdcm.argus_results import send_manager_benchmark_results_to_argus, submit_results_to_argus, \
+    ManagerBackupReadResult, ManagerBackupBenchmarkResult
 from sdcm.mgmt import ScyllaManagerError, TaskStatus, HostStatus, HostSsl, HostRestStatus
 from sdcm.mgmt.cli import ScyllaManagerTool, RestoreTask
 from sdcm.mgmt.common import reconfigure_scylla_manager, get_persistent_snapshots
@@ -572,7 +577,7 @@ class ManagerTestFunctionsMixIn(
         restore_task.wait_and_get_final_status(step=30, timeout=timeout)
         assert restore_task.status == TaskStatus.DONE, f"Restoration of {snapshot_tag} has failed!"
         InfoEvent(message=f'The restore task has ended successfully. '
-                          f'Restore run time: {restore_task.duration}.').publish()
+                  f'Restore run time: {restore_task.duration}.').publish()
         if restore_schema:
             self.db_cluster.restart_scylla()  # After schema restoration, you should restart the nodes
         return restore_task
@@ -1550,3 +1555,93 @@ class ManagerRestoreBenchmarkTests(ManagerTestFunctionsMixIn):
                                "Please provide the 'mgmt_reuse_backup_snapshot_name' parameter.")
 
         self.test_restore_from_precreated_backup(snapshot_name, restore_outside_manager=True)
+
+
+class ManagerReportType(Enum):
+    READ = 1
+    BACKUP = 2
+
+
+class ManagerBackupRestoreConcurrentTests(ManagerTestFunctionsMixIn):
+    def report_to_argus(self, report_type: ManagerReportType, data: dict, label: str):
+        if report_type == ManagerReportType.READ:
+            table = ManagerBackupReadResult(sut_timestamp=mgmt.get_scylla_manager_tool(
+                manager_node=self.monitors.nodes[0]).sctool.client_version_timestamp)
+        elif report_type == ManagerReportType.BACKUP:
+            table = ManagerBackupBenchmarkResult(sut_timestamp=mgmt.get_scylla_manager_tool(
+                manager_node=self.monitors.nodes[0]).sctool.client_version_timestamp)
+        else:
+            raise InvalidArgument("Unknown report type")
+
+        for key, value in data.items():
+            table.add_result(column=key, value=value, row=label, status=Status.UNSET)
+        submit_results_to_argus(self.test_config.argus_client(), table)
+
+    def create_backup_and_report(self, mgr_cluster, label: str):
+        # After the issue https://github.com/scylladb/scylla-manager/issues/4125 is resolved try to rerun it with
+        # different `transfers` settings to apply more IO pressure on the scylla cluster
+        task = mgr_cluster.create_backup_task(location_list=self.locations, rate_limit_list=["0"])
+
+        backup_status = task.wait_and_get_final_status(timeout=7200)
+        assert backup_status == TaskStatus.DONE, "Backup upload has failed!"
+
+        InfoEvent(
+            message=f'Backup total time is: {task.duration}.').publish()
+        backup_report = {
+            "backup time": int(task.duration.total_seconds()),
+        }
+        self.report_to_argus(ManagerReportType.BACKUP, backup_report, label)
+        return task
+
+    def run_read_stress_and_report(self, label):
+        stress_queue = []
+
+        for command in self.params.get('stress_read_cmd'):
+            stress_queue.append(self.run_stress_thread(command, round_robin=True, stop_test_on_failure=False))
+
+        with ExecutionTimer() as stress_timer:
+            for stress in stress_queue:
+                assert self.verify_stress_thread(cs_thread_pool=stress), "Read stress command"
+        InfoEvent(message=f'Read stress duration: {stress_timer.duration}s.').publish()
+
+        read_stress_report = {
+            "read time": int(stress_timer.duration.total_seconds()),
+        }
+        self.report_to_argus(ManagerReportType.READ, read_stress_report, label)
+
+    def test_backup_benchmark(self):
+        self.log.info("Executing test_backup_restore_benchmark...")
+
+        self.log.info("Write data to table")
+        self.run_prepare_write_cmd()
+
+        self.log.info("Disable clusterwide compaction")
+        compaction_ops = CompactionOps(cluster=self.db_cluster)
+        #  Disable keyspace autocompaction cluster-wide since we dont want it to interfere with our restore timing
+        for node in self.db_cluster.nodes:
+            compaction_ops.disable_autocompaction_on_ks_cf(node=node)
+
+        manager_tool = mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0])
+        mgr_cluster = self.ensure_and_get_cluster(manager_tool)
+
+        self.log.info("Create and report backup time")
+        backup_task = self.create_backup_and_report(mgr_cluster, "Backup")
+
+        self.log.info("Remove backup")
+        backup_task.delete_backup_snapshot()
+
+        self.log.info("Run read test")
+        self.run_read_stress_and_report("Read stress")
+
+        self.log.info("Create and report backup time during read stress")
+
+        backup_thread = threading.Thread(target=self.create_backup_and_report,
+                                         kwargs={"mgr_cluster": mgr_cluster, "label": "Backup during read stress"})
+
+        read_stress_thread = threading.Thread(target=self.run_read_stress_and_report,
+                                              kwargs={"label": "Read stress during backup"})
+        backup_thread.start()
+        read_stress_thread.start()
+
+        backup_thread.join()
+        read_stress_thread.join()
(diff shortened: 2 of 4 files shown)


kreuzerkrieg:
> Introduce the `test_backup_benchmark` test, which measures backup time under read stress conditions. This test performs multiple actions to consolidate all necessary data into a single table. Initially, it runs and measures the backup process, followed by the read stress test. Finally, it executes both processes asynchronously to observe how the performance of reading and backing up degrades.
> 
> Argus results: For [100GB run](https://argus.scylladb.com/tests/scylla-cluster-tests/6a18ccd0-1d6a-4f00-a473-

I have to admit it looks super suspicious that the read did not degrade even slightly

mikliapko:
@kreuzerkrieg There was some [refactoring](https://github.com/scylladb/scylla-cluster-tests/pull/9222) done for Manager tests.
Please, consider rebasing to master and putting your new test to the right place or probably to a separate class for Backup benchmark.

kreuzerkrieg:
> @kreuzerkrieg There was some [refactoring](https://github.com/scylladb/scylla-cluster-tests/pull/9222) done for Manager tests. Please, consider rebasing to master and putting your new test to the right place or probably to a separate class for Backup benchmark.

rebased and moved the test to another class

mgmt_cli_test.py
```diff
@@ -1550,3 +1551,88 @@ def test_restore_data_without_manager(self):
...
+                row_name=label,
+            )
+
+        def create_backup(label):
+            mgr_cluster = self.ensure_and_get_cluster(
+                mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0]))
+
+            backup_start_time = time.time()
+            task = mgr_cluster.create_backup_task(location_list=self.locations, rate_limit_list=["0"])
+            backup_task_status = task.wait_for_uploading_stage(timeout=200000)
```
mikliapko, line 1573: I think `200000` seconds timeout can be significantly decreased

mgmt_cli_test.py
```diff
@@ -1550,3 +1551,88 @@ def test_restore_data_without_manager(self):
...
+            assert backup_status == TaskStatus.DONE, "Backup upload has failed!"
+            InfoEvent(
+                message=f'Backup upload upload took: {upload_end_time - upload_start_time}.').publish()
+            InfoEvent(
+                message=f'Backup total time is: {(backup_end_time - backup_start_time) + (upload_end_time - upload_start_time)}.').publish()
+            backup_report = {
+                "backup time": int(backup_end_time - backup_start_time),
+                "upload time": int(upload_end_time - upload_start_time),
+                "total": int((backup_end_time - backup_start_time) + (upload_end_time - upload_start_time)),
+            }
```
mikliapko, line 1589: You may consider an option to get the whole backup process time as `task.duration` property.
It will minimize the number of calculations here

mgmt_cli_test.py
```diff
@@ -1550,3 +1551,88 @@ def test_restore_data_without_manager(self):
...
+            report_to_argus(backup_report, label)
+            return task
+
+        def run_read_stress(label):
+            stress_queue = []
+
+            for command in self.params.get('stress_read_cmd'):
+                stress_queue.append(self.run_stress_thread(command, round_robin=True, stop_test_on_failure=False))
+            read_start_time = time.time()
```
mikliapko, line 1598: Btw, you may consider to use `ExecutionTimer` for these duration measurements:
https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/utils/time_utils.py

mgmt_cli_test.py
```diff
@@ -1550,3 +1551,88 @@ def test_restore_data_without_manager(self):
...
+            for stress in stress_queue:
+                assert self.verify_stress_thread(cs_thread_pool=stress), "Read stress command"
+
+            read_end_time = time.time()
+            InfoEvent(message=f'Read stress duration: {read_end_time - read_start_time:.2f}s.').publish()
+            read_stress_report = {
+                "read stress time": int(read_end_time - read_start_time),
+            }
+            report_to_argus(read_stress_report, label)
```
mikliapko, line 1607: I'm not sure `report_to_argus` should be a part of these `run_read_stress` and `create_backup` methods.
At least, method names are missing this part

mgmt_cli_test.py
```diff
@@ -1550,3 +1551,88 @@ def test_restore_data_without_manager(self):
...
+            )
+
+        def create_backup(label):
+            mgr_cluster = self.ensure_and_get_cluster(
+                mgmt.get_scylla_manager_tool(manager_node=self.monitors.nodes[0]))
+
+            backup_start_time = time.time()
+            task = mgr_cluster.create_backup_task(location_list=self.locations, rate_limit_list=["0"])
+            backup_task_status = task.wait_for_uploading_stage(timeout=200000)
+            assert backup_task_status, "Backup has failed!"
```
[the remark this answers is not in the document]
kreuzerkrieg: what do you mean? the `wait_for_uploading_stage` calls `task.wait_for_status` and it returns `bool` then I'll have to assert on `is_status_reached` or something

mgmt_cli_test.py
```diff
@@ -1550,3 +1551,88 @@ def test_restore_data_without_manager(self):
...
+            backup_end_time = time.time()
```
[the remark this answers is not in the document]
Michal-Leszczynski: Not really:/ The `sctool progress` allows you to check task status, which also includes task stage. You can check the current stage and see if SM if before, during or after snapshot creation, but it won't give you direct timestamp of snapshot creation.

```go
const (
	BackupStageInit         string = "INIT"
	BackupStageSnapshot     string = "SNAPSHOT"
	BackupStageAwaitSchema  string = "AWAIT_SCHEMA"
	BackupStageIndex        string = "INDEX"
	BackupStageManifest     string = "MANIFEST"
	BackupStageSchema       string = "SCHEMA"
	BackupStageDeduplicate  string = "DEDUPLICATE"
	BackupStageUpload       string = "UPLOAD"
	BackupStageMoveManifest string = "MOVE_MANIFEST"
	BackupStageMigrate      string = "MIGRATE"
	BackupStagePurge        string = "PURGE"
	BackupStageDone         string = "DONE"
)
```

regevran:
The units in the issue should be ~[s]~ --> [h:m:s]

sdcm/argus_results.py
```diff
@@ -106,6 +106,9 @@ class Meta:
             ColumnMetadata(name="download bandwidth", unit="MiB/s/shard", type=ResultType.FLOAT, higher_is_better=True),
             ColumnMetadata(name="l&s bandwidth", unit="MiB/s/shard", type=ResultType.FLOAT, higher_is_better=True),
             ColumnMetadata(name="repair time", unit="s", type=ResultType.DURATION, higher_is_better=False),
+            ColumnMetadata(name="backup time", unit="s", type=ResultType.DURATION, higher_is_better=False),
+            ColumnMetadata(name="upload time", unit="s", type=ResultType.DURATION, higher_is_better=False),
+            ColumnMetadata(name="read stress time", unit="s", type=ResultType.DURATION, higher_is_better=False),
```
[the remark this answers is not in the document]
kreuzerkrieg: Added two new result tables

mgmt_cli_test.py
```diff
@@ -1550,3 +1551,88 @@ def test_restore_data_without_manager(self):
...
+            read_start_time = time.time()
+            for stress in stress_queue:
+                assert self.verify_stress_thread(cs_thread_pool=stress), "Read stress command"
+
+            read_end_time = time.time()
+            InfoEvent(message=f'Read stress duration: {read_end_time - read_start_time:.2f}s.').publish()
+            read_stress_report = {
+                "read stress time": int(read_end_time - read_start_time),
+            }
+            report_to_argus(read_stress_report, label)
```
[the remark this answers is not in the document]
kreuzerkrieg: Done

mgmt_cli_test.py
kreuzerkrieg: reduced

mgmt_cli_test.py
kreuzerkrieg: Done

mgmt_cli_test.py
kreuzerkrieg: Done

mgmt_cli_test.py
kreuzerkrieg: renamed

kreuzerkrieg:
> The units in the issue should be ~[s]~ --> [h:m:s]

It is the way it is presented, I think, I do send seconds to it and it presents it as a time

kreuzerkrieg:
> backup time [s]	upload time [s]	total [s]
> Backup times	00:00:12	00:09:45	00:09:11
> Backup during read stress	00:00:12	00:15:07	00:14:43
> From the code I see that `backup time` is total backup time without upload, right? It would be nice to indicate it with a more descriptive column name.

Should I name it "snapshot time"?

> Also, why `total` is smaller than `upload time`?

Good question, the total time is taken from backup `task.duration` the rest is measured in place and it doesnt add up. Ideas?

Michal-Leszczynski:
> Should I name it "snapshot time"?

It's not only about snapshot (e.g. we also fetch schema, create manifests, etc...).
In general I'm not sure why do we need to report the time up to the upload stage? Is it useful?
But I see that the upload time is also not only about upload (e.g. purging backup location from unnecessary files).
I will leave some comments in changed files in a bit.

I created and [issue](https://github.com/scylladb/scylla-manager/issues/4123) about displaying backup upload bandwidth/duration - in the future it could make things easier for such benchmarks.

mgmt_cli_test.py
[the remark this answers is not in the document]
mikliapko: I'm expecting `task.wait_for_status` raise `WaitForTimeoutError` exception if timeout reaches.
But whatever, the comment can be ignored, not significant at all.

Btw, your code might be a bit shortened:
```
assert task.wait_for_uploading_stage(timeout=200000), "Backup has failed!"
```
From my point of view, it improves readability but it's up to you.

mgmt_cli_test.py
```diff
@@ -1550,3 +1555,101 @@ def test_restore_data_without_manager(self):
...
+                manager_node=self.monitors.nodes[0]).sctool.client_version_timestamp)
+        else:
+            raise InvalidArgument("Unknown report type")
+
+        for key, value in data.items():
+            table.add_result(column=key, value=value, row=label, status=Status.UNSET)
+        submit_results_to_argus(self.test_config.argus_client(), table)
+
+    def create_backup_and_report(self, mgr_cluster, label: str):
+        task = mgr_cluster.create_backup_task(location_list=self.locations, rate_limit_list=["0"])
```
Michal-Leszczynski, line 1581: In terms of running the backup as fast as possible (and with as much cluster pressure as possible), SM 3.4 introduced `--transfers` flag which controls the amount of parallel file transfers (2 by default).

When downloading files during restore, we got the best results when setting it to `2*node_shard_cnt`.
Not sure if the same value would be the best for upload, but increasing it from 2 could definitely help.

configurations/manager/100GB_dataset.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+stress_read_cmd: [ "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..26214400",
+                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
+                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
+                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
```
Michal-Leszczynski, line 11: Maybe we would notice read degradation if we increase the read pressure?
I'm also thinking about changing it to mixed reads-writes, but then I'm not sure about how disabled auto compaction impacts this part.

mgmt_cli_test.py
regevran: We need a way to accurately and _repeatedly_ measure the same thing.
If "cluster pressure" means lots of reads from disk during backup - then yes, this is what we need.

configurations/manager/100GB_dataset.yaml
regevran: We might want to add writes later on, but for the base measures reads would be enough.
We should see degradation in times for both the upload and the queries.
I.e - backing up 100GB and reading 100GB at the same time should both take longer compared to just uploading and to just reading respectively.

kreuzerkrieg:
> > Should I name it "snapshot time"?
> 
> It's not only about snapshot (e.g. we also fetch schema, create manifests, etc...). In general I'm not sure why do we need to report the time up to the upload stage? Is it useful? But I see that the upload time is also not only about upload (e.g. purging backup location from unnecessary files). I will leave some comments in changed files in a bit.
> 
> I created and [issue](https://github.com/scylladb/scylla-manager/issues/4123) about displaying backup upload bandwidth/duration - in the future it could make things easier for such benchmarks.

Now when I see that it is negligible slice of the whole process I guess it is worth just dropping it

mgmt_cli_test.py
```diff
@@ -1550,3 +1555,101 @@ def test_restore_data_without_manager(self):
...
+
+        with ExecutionTimer() as backup_timer:
+            backup_task_status = task.wait_for_uploading_stage(timeout=600)
+            assert backup_task_status, "Backup has failed!"
+        InfoEvent(message=f'Backup without upload took: {backup_timer.duration}s.').publish()
+
+        with ExecutionTimer() as backup_upload_timer:
+            backup_status = task.wait_and_get_final_status(timeout=72000)
```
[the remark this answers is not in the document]
kreuzerkrieg: same as https://github.com/scylladb/scylla-cluster-tests/pull/9307#issuecomment-2498280918

mgmt_cli_test.py
kreuzerkrieg: tried to increase it to 16 and then to 32 and I dont see any difference. Moreover, the second run of backup timeouts since for some unknown reason the backup proves become significantly slower even when the concurrent read is over long time ago

configurations/manager/100GB_dataset.yaml
kreuzerkrieg: > Maybe we would notice read degradation if we increase the read pressure? I'm also thinking about changing it to mixed reads-writes, but then I'm not sure about how disabled auto compaction impacts this part.

quite the opposite, I'm running now with 50 threads just to get to reasonable latencies, 500 thread is such an overkill that the latency is skyrocketing and the CPU is 100% utilized.
with 50 threads I'm getting ~1.5ms latency and I expect it rise during the concurrent backup is running

mgmt_cli_test.py
kreuzerkrieg: I will keep track of it, once [the issue](https://github.com/scylladb/scylla-manager/issues/4125) is resolved I will try to rerun it with different `--transfers` settings

Michal-Leszczynski (approved):

sdcm/argus_results.py
```diff
@@ -117,6 +117,24 @@ class Meta:
...
+        ]
+
+
+class ReadResult(GenericResultTable):
+    class Meta:
+        name = "Read timing"
+        description = "Read timing"
+        Columns = [
+            ColumnMetadata(name="read time", unit="s", type=ResultType.DURATION, higher_is_better=False),
+        ]
```
mikliapko, line 135: I'd suggest to show connection to Manager in class names.
What do you think about renaming to `ManagerBackupBenchmarkResult` and `ManagerBackupBenchmarkReadResult`?
Yes, it's more complex but pretty straightforward for everyone.

mgmt_cli_test.py
```diff
@@ -1550,3 +1555,93 @@ def test_restore_data_without_manager(self):
                                "Please provide the 'mgmt_reuse_backup_snapshot_name' parameter.")
 
         self.test_restore_from_precreated_backup(snapshot_name, restore_outside_manager=True)
+
+
+class ReportType(Enum):
```
mikliapko, line 1560: And the same here about class name, please

sdcm/argus_results.py
kreuzerkrieg: I dont mind having whatever name for the class, but why it should be connected to the Manages? Is it possible that someone would use this table to report read timing to argus not in a context of manager?

mikliapko: Good question btw.
@soyacz What approach do you prefer?

soyacz: Let's add `ManagerBackup` prefix as currently it sounds too generic. 
Unless you decide to refactor and put Manager related stuff out of this file to separate one somewhere around manager code - then prefixes are redundant.

Btw. from my uninvolved perspective I don't know even what is `read` - does it refer to backup or reading from db? - descriptions could be more descriptive so anyone could understand it.

mgmt_cli_test.py
kreuzerkrieg: Done

sdcm/argus_results.py
kreuzerkrieg: Done

kreuzerkrieg:
@scylladb/qa-maintainers could you push it forward please?

kreuzerkrieg:
> I briefly looked at 100gb dataset - for me ingestion of this size should take bit more than 7m - did you verify correctness of this element?

This test uses existing ingestion machinery. I assume that ingestion works.
In any case, looks like it takes mpore or less 7 or 8 minutes
image


> also there's no jenkinsfile related this test, how it is going to be used?

As any other SM test, just drop the right test name in the existing jenkins pipeline

image

PR #9315, opened 2024-11-21
fix(dependabot): fix the python version to use in dependabot

fruch:
dependabot now default to python 3.12, hence with some of our requirements it can use pip/pip-tools, so we should be pinning the version down using .python-version file

Ref: https://github.com/dependabot/dependabot-core/issues/1455

### Testing

- [ ] would be able to test it only after merge

### Reminders


vponomaryov (approved):
LGTM

PR #9329, opened 2024-11-21
improvement(logcollector): collect schema info

juliayakovlev:
Collect `schema` and `system_schema.tables` info and save it on the runner. Add the new logs to the node logs archive.

Task: https://github.com/scylladb/qa-tasks/issues/1813

### Testing

- [x] [one network interface, 2 user keyspaces](https://argus.scylladb.com/tests/scylla-cluster-tests/ca364d5a-8eb5-4b51-a17a-4d49672a000a)
- [x] [two network interfaces](https://argus.scylladb.com/tests/scylla-cluster-tests/ea63a965-db70-4687-9264-711439bd2a37)
- [x] [gce](https://argus.scylladb.com/tests/scylla-cluster-tests/9ace342a-50af-48c2-8987-17e8a7fda145)
- [x] [artifacts-azure-image-test](https://argus.scylladb.com/tests/scylla-cluster-tests/0672b5aa-cbd9-45c8-a81e-b5160cb4a8df)

### Reminders

### DIFF
--- a/sdcm/logcollector.py
+++ b/sdcm/logcollector.py
@@ -749,6 +749,10 @@ class ScyllaLogCollector(LogCollector):
                                command='cat /var/log/cloud-init-output.log'),
                     CommandLog(name='cloud-init.log',
                                command='cat /var/log/cloud-init.log'),
+                    CommandLog(name='schema.log',
+                               command='cat schema.log'),
+                    CommandLog(name='system_schema_tables.log',
+                               command='cat system_schema_tables.log'),
                     ]
     cluster_log_type = "db-cluster"
     cluster_dir_prefix = "db-cluster"
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -2953,10 +2953,21 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
 
         self.destroy_credentials()
 
+    def save_nodes_schema(self):
+        if self.db_cluster is None:
+            self.log.info("No nodes found in the Scylla cluster")
+
+        self.log.info("Save nodes user schema in the files")
+        for node in self.db_cluster.nodes:
+            node.save_cqlsh_output_in_file(cmd="desc schema", log_file="schema.log")
+            node.save_cqlsh_output_in_file(cmd="select JSON * from system_schema.tables",
+                                           log_file="system_schema_tables.log")
+
     def tearDown(self):
         self.teardown_started = True
         with silence(parent=self, name='Sending test end event'):
             InfoEvent(message="TEST_END").publish()
+        self.save_nodes_schema()
         self.log.info("Post test validators are starting...")
         for validator_class in teardown_validators_list:
             validator_class(self.params, self).validate()
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -309,6 +309,12 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
         self._node_rack = None
         self._is_zero_token_node = False
 
+    def save_cqlsh_output_in_file(self, cmd: str, log_file: str):
+        self.log.info("Save command '%s' output in the file. Node %s", cmd, self.name)
+        result_tables = self.run_cqlsh(cmd, split=True)
+        for line in result_tables:
+            self.remoter.run(f"echo '{line}' >> {log_file}")
+
     def _is_node_ready_run_scylla_commands(self) -> bool:
         """
         When node is just created and started to configure, during first node initializing, it is impossible to connect to the node yet and
--- a/unit_tests/test_tester.py
+++ b/unit_tests/test_tester.py
@@ -178,6 +178,9 @@ class ClusterTesterForTests(ClusterTester):
         time.sleep(0.5)
         super().stop_event_device()
 
+    def save_nodes_schema(self):
+        pass
+
 
 class SubtestAndTeardownFailsTest(ClusterTesterForTests):
     def test(self):


sdcm/logcollector.py
```diff
@@ -102,9 +102,10 @@ class BaseLogEntity:  # pylint: disable=too-few-public-methods
     collect_timeout = 300
     _params = {}
 
-    def __init__(self, name, command="", search_locally=False):
+    def __init__(self, name, command="", search_locally=False, node_attr=None):
         self.name = name
         self.cmd = command
+        self.node_attr = node_attr
```
fruch, line 108: I would suggest doing it a bit more general so multiple attributes can be formatted into the command
Maybe we can format the node itself ?
```python
>>> cmd = "cqlsh {node.ip}"
>>>
>>> class Node: pass
...
>>> n = Node()
>>> n.ip = "1234"
>>>
>>> cmd.format(node=n)
'cqlsh 1234'
```

juliayakovlev:
> > It does not work for 2 interfaces
> 
> Why it doesn't work? And why do we need external address ? and not cql_ address ? We run it locally on the node it should always be the correct address for cql

This is not cluster node, but is `CollectingNode` that does not get cql_address. I exactly searching how to pass it to the `CollectingNode` class. Maybe we should pass the Scylla node and not parts of it

fruch:
> > > It does not work for 2 interfaces
> > 
> > 
> > Why it doesn't work? And why do we need external address ? and not cql_ address ? We run it locally on the node it should always be the correct address for cql
> 
> This is not a cluster node, but is `CollectingNode` that does not get cql_address. I am exactly searching how to pass it to the `CollectingNode` class. Maybe we should pass the Scylla node and not parts of it

maybe grep it from scylla yaml ? insted ? passing the actual node implementation might be a bit cumbersome solution

juliayakovlev:
> > > > It does not work for 2 interfaces
> > > 
> > > 
> > > Why it doesn't work? And why do we need external address ? and not cql_ address ? We run it locally on the node it should always be the correct address for cql
> > 
> > 
> > This is not a cluster node, but is `CollectingNode` that does not get cql_address. I am exactly searching how to pass it to the `CollectingNode` class. Maybe we should pass the Scylla node and not parts of it
> 
> maybe grep it from scylla yaml ? insted ? passing the actual node implementation might be a bit cumbersome solution

@fruch 
The problem here that we need duplicate the code, find on which interface and which address type is configured to connect for cqlsh.
Maybe is better to save it before log collector? During teardown for example. WDYT?

sdcm/logcollector.py
juliayakovlev: Changed it - prepare the files during tearDown

sdcm/tester.py
```diff
@@ -2953,10 +2953,21 @@ def clean_resources(self):
 
         self.destroy_credentials()
 
+    def save_nodes_schema_before_test_finish(self):
```
vponomaryov, line 2956: We may want to call it at any point of a test run. For example - before decommission nemesis.
So, do not include current final goal in the naming.
Include only actual action in the naming.

vponomaryov:
> > > > It does not work for 2 interfaces
> > > 
> > > 
> > > Why it doesn't work? And why do we need external address ? and not cql_ address ? We run it locally on the node it should always be the correct address for cql
> > 
> > 
> > This is not a cluster node, but is `CollectingNode` that does not get cql_address. I am exactly searching how to pass it to the `CollectingNode` class. Maybe we should pass the Scylla node and not parts of it
> 
> maybe grep it from scylla yaml ? insted ? passing the actual node implementation might be a bit cumbersome solution

On K8S you will get internal K8S IP and fail to connect to it from the SCT runner.

> @fruch The problem here that we need duplicate the code, find on which interface and which address type is configured to connect for cqlsh. Maybe is better to save it into file before log collector? During teardown for example. WDYT?

I think it must be `best effort` approach. Try out all node IPs while you do not get the data that is required.
You may remember IP type from the first node and start exactly from that type on all other nodes first.

Also, need to use `try-except` block for each of the steps to tolerate problems for one/several intermediate cmd commands.

vponomaryov (changes requested):

juliayakovlev:
> > > > > It does not work for 2 interfaces
> > > > 
> > > > 
> > > > Why it doesn't work? And why do we need external address ? and not cql_ address ? We run it locally on the node it should always be the correct address for cql
> > > 
> > > 
> > > This is not a cluster node, but is `CollectingNode` that does not get cql_address. I am exactly searching how to pass it to the `CollectingNode` class. Maybe we should pass the Scylla node and not parts of it
> > 
> > 
> > maybe grep it from scylla yaml ? insted ? passing the actual node implementation might be a bit cumbersome solution
> 
> On K8S you will get internal K8S IP and fail to connect to it from the SCT runner.
> 
> > @fruch The problem here that we need duplicate the code, find on which interface and which address type is configured to connect for cqlsh. Maybe is better to save it into file before log collector? During teardown for example. WDYT?
> 
> I think it must be `best effort` approach. Try out all node IPs while you do not get the data that is required. You may remember IP type from the first node and start exactly from that type on all other nodes first.

If to do that before test end, when the cluster still exist, there is no problem to cqlsh. No need to try all IP's one by one, because we keep information aboutl IP's configuration

> 
> Also, need to use `try-except` block for each of the steps to tolerate problems for one/several intermediate cmd commands.

fruch (approved):
LGTM

sdcm/tester.py
```diff
@@ -2953,10 +2953,21 @@ def clean_resources(self):
...
+        for node in self.db_cluster.nodes:
+            node.save_cqlsh_output_in_file(cmd="desc schema", log_file="schema.log")
+            node.save_cqlsh_output_in_file(cmd="select JSON * from system_schema.tables",
+                                           log_file="system_schema_tables.log")
+
     def tearDown(self):
         self.teardown_started = True
         with silence(parent=self, name='Sending test end event'):
             InfoEvent(message="TEST_END").publish()
+        self.save_nodes_schema()
```
vponomaryov, line 2970: @fruch , @juliayakovlev 

The ignored request to make the schema gathering be `best-effort` approach instead of the `blocking` one as it is here paid off already in the K8S test runs where K8S logs gathering step was not reached because tearDown function failed to execute CQL commands.

See: https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/upgrade/job/upgrade-scylla-k8s-eks-test/11/consoleFull

juliayakovlev: @vponomaryov 
I did not ignored your request. I explained why I did it in the tearDown step.
What is a different in K8S, that cql can't be executed on this step?
It means that for K8S we should another solution. It's not needed to try out all node IPs for other backends (aws, gce..)

fruch: for now  put the `@silence()` decorator ontop of it

vponomaryov: > @vponomaryov I did not ignored your request. I explained why I did it in the tearDown step. What is a different in K8S, that cql can't be executed on this step? It means that for K8S we should another solution. It's not needed to try out all node IPs for other backends (aws, gce..)

Some temporary bug with the CQL connectivity breakage.
Gathering logs always must be `best-effort`.
It makes no sense to stop gathering other logs if some failed.

PR #9345, opened 2024-11-24
fix(test_user_batch_custom_time): handle stress command correctly

fruch:
seems like we were running into situations we try to add all nodes ips into stress commands, and it's the test code
```
File ".../longevity_test.py", line 300, in test_user_batch_custom_time
self._run_user_stress_in_batches(batch_size=batch_size,
File ".../longevity_test.py", line 336, in _run_user_stress_in_batches
batch_params['stress_cmd'] += [params['stress_cmd'] + nodes_ips]
TypeError: list indices must be integers or slices, not str
```

in this change remove a parentheses that was causing function to return tuple and not string as expected

### Testing

- [x] 🟢 https://argus.scylladb.com/tests/scylla-cluster-tests/8d280858-295a-482a-970e-e62bfcfed597

### Reminders

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -109,6 +109,8 @@ pipeline {
             steps {
                 script {
                     try {
+                        checkoutQaInternal(params)
+
                         sh './docker/env/hydra.sh unit-tests'
                         pullRequestSetResult('success', 'jenkins/unittests', 'All unit tests are passed')
                     } catch(Exception ex) {
--- a/longevity_test.py
+++ b/longevity_test.py
@@ -19,11 +19,13 @@ import string
 import tempfile
 import itertools
 import contextlib
+from typing import List, Dict
 
 import yaml
 from cassandra import AlreadyExists, InvalidRequest
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
+from sdcm import sct_abs_path
 from sdcm.sct_events.group_common_events import \
     ignore_large_collection_warning, \
     ignore_max_memory_for_unlimited_query_soft_limit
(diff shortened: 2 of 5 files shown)


roydahan (approved):

soyacz (dismissed):
LGTM

vponomaryov (dismissed):
Need to fix unit tests

longevity_test.py
```diff
@@ -332,12 +336,13 @@ def chunks(_list, chunk_size):
             # add few stress threads with tables that weren't pre-created
             customer_profiles = self.params.get('cs_user_profiles')
             for cs_profile in customer_profiles:
+                cs_profile = sct_abs_path(cs_profile)  # noqa: PLW2901
                 # for now we'll leave to just one fresh table, to kick schema update
                 num_of_newly_created_tables = 1
                 self._pre_create_templated_user_schema(batch_start=extra_tables_idx,
                                                        batch_end=extra_tables_idx+num_of_newly_created_tables)
                 for i in range(num_of_newly_created_tables):
-                    batch.append(self.create_templated_user_stress_params(extra_tables_idx + i, cs_profile=cs_profile))
```
fruch, line 340: this is the core of the fix

fruch:
the test in 2024.2 reproducers, did proved this is fixing the issue.
and there also a unittest introduce that did demonstrate the bug, and now running cleanly.

soyacz (approved):

vponomaryov (approved):
LGTM

Jenkinsfile
```diff
@@ -109,6 +109,8 @@ pipeline {
             steps {
                 script {
                     try {
+                        checkoutQaInternal(params)
```
vponomaryov, line 112: Why is it needed here?

fruch: cause the unittest is going via code path that depends on configuration files existing in the internal repo
i.e. the 5000 tables case has the user profile yaml there.

PR #9353, opened 2024-11-25
fix(docker-artifacts-test): do not install manager by default

dimakr:
After https://github.com/scylladb/scylla-cluster-tests/pull/9331 change was introduced the artifacts-docker test started to fail as it has `use_mgmt` config setting enabled by default.
This hot fix sets `SCT_USE_MGMT=false` env variable for the artifacts-docker jenkins job, so that the test can pass the SCT config validation.

### Testing

- [ ]

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #9367, opened 2024-11-26
fix(docker-artifacts-test): add missing `,` between params

yaronkaikov:
In https://github.com/scylladb/scylla-cluster-tests/pull/9353 a new parameter was added, but a missing `,` is failing the pipeline

fixing it


Annamikhlin (approved):

fruch (approved):
LGTM

PR #9368, opened 2024-11-26
feature(create_test_release_jobs): Add job metadata files

k0machi:
This commit adds a simple per-folder definition file (and format
documentation) format for storing additional metadata on the jenkins
jobs. The metadata is appended to the end of job description and is
planned to be consumed by Argus to determine both the job type and
subtype.

Fixes scylla-db/argus#428

### Testing

- [ ]

### Reminders


docs/job_defines_format.md
```diff
@@ -0,0 +1,43 @@
+# Jenkins Job Definitions
+
+Due to a growing need of having extra metadata available to jenkins consumers
+(both users and automation) a job define system was added to SCT jenkins generator.
+
+It replaces previous `_display_name` file (which used to contain just the display
+ name of the folder) with a special .yaml file called `_jobDefines.yaml`.
```
fruch, line 7: let's stick to snack case for filenames, and replace the `defines` with `definitions`
also it's folder definitions
 
`_folder_definitions.yaml`

docs/job_defines_format.md
```diff
@@ -0,0 +1,43 @@
...
+
+```yaml
+# used by argus to determine argus plugin
+job-type: scylla-cluster-tests
+# used by argus to determine which parameter wizard to use on execution
+job-sub-type: longevity
+# replacement for _display_name file, sets the jenkins folder display name
+folder-name: Cluster - Longevity Tests
+# Per job (regex supported) job overrides, defines as a mapping
+overrides:
```
fruch, line 19: I think `overrides`, should be limited to `job-sub-type` 

I don't want people to be lazy and use it for `job-description`

utils/build_system/create_test_release_jobs.py
```diff
@@ -144,18 +152,37 @@ def _log_jenkins_exception(exc):
...
                         template_context: dict | None = None,
                         job_name_suffix: str = '-test'):
         for root, _, job_files in os.walk(local_path):
             jenkins_path = Path(root).relative_to(local_path)
+            defines = self.load_defines(Path(root))
+            job_overrides = defines.pop("overrides", {})
 
             # get display names, if available
-            display_name = jenkins_path.name
-            if '_display_name' in job_files:
```
fruch, line 157: we should leave this as fallback, until all of those are replace in the tree..

docs/job_defines_format.md
k0machi: Hmm, what's wrong with using it for job description as well? How else they could add a text description to a specific job?

utils/build_system/create_test_release_jobs.py
k0machi: Okay

docs/job_defines_format.md
k0machi: Okay

docs/job_defines_format.md
fruch: job description should come from the jenkinsfile themselves, we shouldn't put all the description in the same file, too much housekeep, for operation like moving a job from one folder to another

roydahan: I agree with @fruch here.
BTW, somehow related , I had a feature request somewhere to add like metadata tags to the test yamls, so Argus could present them and also search test cases according to them.

k0machi: Done

utils/build_system/create_test_release_jobs.py
k0machi: Reverted

k0machi:
Added job description annotations to the functionality and documentation for them.

docs/job_defines_format.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+Once template is generated the defines are applied to the job description
+along with job description from the pipeline file, like so:
+
+```
+This is a simple job description.
+Can be multi line and indented.
+Will return the description without indent.
+jenkins-pipelines/oss/longevity/longevity-cdc-100gb-4h.jenkinsfile
```
fruch, line 46: lets remove the path from the description

fruch (approved):
LGTM

Issue #9284, opened 2024-11-19
Fix Fullscanoperation thread to choose only alive node


aleksbykov:
## Packages

Scylla version: `2024.2.0-20241118.614d56348f46` with build-id `e67376d9ddfea081a3bab398f4581ecdde59911d`

Kernel Version: `5.15.0-1072-aws`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Full scan operation chose node which was then used by rolling restart nemesis and cause the error message:
```
2024-11-18 22:32:44.845: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=e79676a5-c38c-4b33-b7cf-b3f9d96610a8 during_nemesis=RollingRestartCluster duration=13s node=longevity-tls-50gb-3d-2024-2-db-node-c5d16022-6 select_from=keyspace1.standard1 message=FullScanAggregatesOperation operation failed, ReadTimeout error: ReadTimeout('Error from server: code=1200 [Coordinator node timed out waiting for replica nodes\' responses] message="Operation failed for keyspace1.standard1 - received 0 responses and 1 failures from 1 CL=ONE." info={\'consistency\': \'ONE\', \'required_responses\': 1, \'received_responses\': 0}')
```

Need to fix FullScan thread to choose only alive nodes 
-or-
Fix rolling restart nemesis to mark restarting node as busy for other operations

## Impact

Reported Error event mark job as failed.

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 6 nodes (i4i.4xlarge)

Scylla Nodes used in this run:
- longevity-tls-50gb-3d-2024-2-db-node-c5d16022-9 (52.4.92.28 | 10.12.35.142) (shards: 14)
- longevity-tls-50gb-3d-2024-2-db-node-c5d16022-8 (98.85.39.206 | 10.12.35.198) (shards: 14)
- longevity-tls-50gb-3d-2024-2-db-node-c5d16022-7 (52.72.119.13 | 10.12.34.73) (shards: 14)
- longevity-tls-50gb-3d-2024-2-db-node-c5d16022-6 (44.214.249.197 | 10.12.34.72) (shards: 14)
- longevity-tls-50gb-3d-2024-2-db-node-c5d16022-5 (35.172.65.4 | 10.12.32.10) (shards: 14)
[...]

PR #8705, opened 2024-09-15
fix(full scan thread): Added missing fetch-next-page to fetch_result_pages

yarongilor:
fetch_result_pages had a "while" loop where a call to result.fetch_next_page() is missing. Adding it back.
Fixes: #8287

### Testing

- [ ]

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #9370, opened 2024-11-26
fix(FullScan): choose non-running_nemesis node

temichus:
this commit has the following changes

1 introduce common targed_node_lock mechanism
that can be used in nemesis and Scan operations

2 FullScan operation now run only on free of nemeses node

3 change all node.running_nemesis settings to use common methods
set/unset_running_nemesis from common targed_node_lock file (except unit tests)

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/9284

### Testing

- https://argus.scylladb.com/tests/scylla-cluster-tests/d1a01caa-4f7e-40c1-8c17-69c9a6882de4

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -31,7 +31,7 @@ import json
 import itertools
 import enum
 from distutils.version import LooseVersion
-from contextlib import ExitStack, contextmanager
+from contextlib import ExitStack
 from typing import Any, List, Optional, Type, Tuple, Callable, Dict, Set, Union, Iterable
 from functools import wraps, partial
 from collections import defaultdict, Counter, namedtuple
@@ -66,6 +66,8 @@ from sdcm.cluster_k8s import (
 )
 from sdcm.db_stats import PrometheusDBStats
 from sdcm.log import SDCMAdapter
+from sdcm.target_node_lock import run_nemesis, set_running_nemesis, unset_running_nemesis, \
+    NEMESIS_TARGET_SELECTION_LOCK, CantAcquireLockException
 from sdcm.logcollector import save_kallsyms_map
 from sdcm.mgmt.common import TaskStatus, ScyllaManagerError, get_persistent_snapshots
 from sdcm.nemesis_publisher import NemesisElasticSearchPublisher
@@ -106,6 +108,7 @@ from sdcm.utils.common import (get_db_tables, generate_random_string,
                                update_authenticator, ParallelObject,
                                ParallelObjectResult, sleep_for_percent_of_duration, get_views_of_base_table)
 from sdcm.utils.features import is_tablets_feature_enabled
+from sdcm.utils.nemesis_thread_safe_operations import safe_cluster_start_stop, safe_node_restart
 from sdcm.utils.quota import configure_quota_on_node_for_scylla_user_context, is_quota_enabled_on_node, enable_quota_on_node, \
     write_data_to_reach_end_of_quota
 from sdcm.utils.compaction_ops import CompactionOps, StartStopCompactionArgs
@@ -172,7 +175,6 @@ EXCLUSIVE_NEMESIS_NAMES = (
     "disrupt_terminate_kubernetes_host_then_decommission_and_add_scylla_node",
 )
 
-NEMESIS_TARGET_SELECTION_LOCK = Lock()
 DISRUPT_POOL_PROPERTY_NAME = "target_pool"
 
 
@@ -328,24 +330,6 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         setattr(cls, func.__name__, wrapper)  # bind it to Nemesis class
         return func  # returning func means func can still be used normally
 
-    @staticmethod
-    @contextmanager
-    def run_nemesis(node_list: list['BaseNode'], nemesis_label: str):
-        """
-        pick a node out of a `node_list`, and mark is as running_nemesis
-        for the duration of this context
-        """
-        with NEMESIS_TARGET_SELECTION_LOCK:
-            free_nodes = [node for node in node_list if not node.running_nemesis]
-            assert free_nodes, f"couldn't find nodes for running:`{nemesis_label}`, are all nodes running nemesis ?"
-            node = random.choice(free_nodes)
-            node.running_nemesis = nemesis_label
-        try:
-            yield node
-        finally:
-            with NEMESIS_TARGET_SELECTION_LOCK:
-                node.running_nemesis = None
-
     def use_nemesis_seed(self):
         if nemesis_seed := self.tester.params.get("nemesis_seed"):
             random.seed(nemesis_seed)
@@ -379,14 +363,11 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.target_node = node
 
     def set_current_running_nemesis(self, node):
-        with NEMESIS_TARGET_SELECTION_LOCK:
-            node.running_nemesis = self.current_disruption
+        set_running_nemesis(node, self.current_disruption)
 
     @staticmethod
     def unset_current_running_nemesis(node):
-        if node is not None:
-            with NEMESIS_TARGET_SELECTION_LOCK:
-                node.running_nemesis = None
+        unset_running_nemesis(node)
 
     def set_target_node_pool_type(self, pool_type: NEMESIS_TARGET_POOLS = NEMESIS_TARGET_POOLS.data_nodes):
         """Set pool type to choose nodes for target node """
@@ -449,10 +430,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 self.target_node = random.choice(nodes)
 
             if current_disruption:
-                self.target_node.running_nemesis = current_disruption
+                set_running_nemesis(self.target_node, current_disruption)
                 self.set_current_disruption(current_disruption)
             elif self.current_disruption:
-                self.target_node.running_nemesis = self.current_disruption
+                set_running_nemesis(self.target_node, self.current_disruption)
             else:
                 raise ValueError("current_disruption is not set")
             self.log.info('Current Target: %s with running nemesis: %s',
@@ -980,7 +961,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     @decorate_with_context(ignore_ycsb_connection_refused)
     def disrupt_rolling_restart_cluster(self, random_order=False):
-        self.cluster.restart_scylla(random_order=random_order)
+        try:
+            safe_cluster_start_stop(self.target_node.running_nemesis, self.cluster)
+        except CantAcquireLockException as e:
+            UnsupportedNemesis(e)
 
     def disrupt_switch_between_password_authenticator_and_saslauthd_authenticator_and_back(self):
         """
@@ -3076,7 +3060,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 f'Schema restoration of {chosen_snapshot_tag} has failed!'
 
             with ignore_ycsb_connection_refused():
-                self.cluster.restart_scylla()  # After schema restoration, you should restart the nodes
+                safe_cluster_start_stop(self.target_node.running_nemesis, self.cluster)
             self.tester.set_ks_strategy_to_network_and_rf_according_to_cluster(
                 keyspace=chosen_snapshot_info["keyspace_name"], repair_after_alter=False)
 
@@ -4055,7 +4039,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                     terminate_pattern.timeout):
                 stack.enter_context(expected_start_failed_context)
             with ignore_stream_mutation_fragments_errors(), ignore_raft_topology_cmd_failing(), \
-                self.run_nemesis(node_list=self.cluster.data_nodes, nemesis_label="DecommissionStreamingErr") as verification_node, \
+                run_nemesis(node_list=self.cluster.data_nodes, nemesis_label="DecommissionStreamingErr") as verification_node, \
                 FailedDecommissionOperationMonitoring(target_node=self.target_node,
                                                       verification_node=verification_node,
                                                       timeout=full_operations_timeout):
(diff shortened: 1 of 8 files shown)


temichus:
@fruch, @soyacz, @aleksbykov Coul you please take a look at this idea before I start testing it

sdcm/nemesis.py
```diff
@@ -453,10 +431,10 @@ def set_target_node(self, dc_idx: Optional[int] = None, rack: Optional[int] = No
                 self.target_node = random.choice(nodes)
 
             if current_disruption:
-                self.target_node.running_nemesis = current_disruption
+                set_running_nemesis(self.target_node, current_disruption, self.disruptive)
```
soyacz, line 434: `_get_target_nodes` anyway works the old way - never selecting any node that runs any nemesis, regardless of descriptiveness.  So selecting node during nemesis startup will work the old way. 
I'm not saying it's wrong, just noting it.

soyacz (commented):
Generally LGTM, still needs testing as setting target node is messy (e.g. we set target node in `disrupt_method_wrapper` and unset it in `run` method.)

sdcm/scan_operation_thread.py
```diff
@@ -28,7 +29,7 @@
 from sdcm.utils.decorators import retrying, Retry
 
 if TYPE_CHECKING:
-    from sdcm.cluster import BaseNode
+    pass
```
fruch, line 32: just remove the whole if

sdcm/nemesis.py
```diff
@@ -381,14 +362,11 @@ def publish_event(self, disrupt, status=True, data=None):
         DisruptionEvent(nemesis_name=disrupt, severity=severity, **data).publish()
 
     def set_current_running_nemesis(self, node):
-        with NEMESIS_TARGET_SELECTION_LOCK:
-            node.running_nemesis = self.current_disruption
+        set_running_nemesis(node, self.current_disruption, self.disruptive)
```
fruch, line 365: how does self.disruptive would work ?

all those are set on only on nemesis classes, not the Nemesis class.

aleksbykov (dismissed):
Short remarks.
Also, because running_nemesis is list now, need to be carefully with passing to (un)set_running_nemesis function name of nemesis.

temichus:
@soyacz @fruch @aleksbykov, Thank you for your comments. I've changed the logic to choose a non running_nemesis node for run_nemesis, and it will help to do FullScan safe when several nemesis are running. But I don't think it will help for RollingRestartCluster because it 'disrupts' all nodes in the cluster. What do you think regarding updating disrupt_rolling_restart_cluster to restart only non running_nemesis nodes?

temichus:
so locking nodes 1 by one during RollingRestartCluster does not work. I'm still faced with issue https://github.com/scylladb/scylla-cluster-tests/issues/9284

lock all nodes before  RollingRestartCluster. looks promising 

FullScan got the assertion "couldn't find nodes for running:`{nemesis_label}`, are all nodes running nemesis ?"
RollingRestartCluster raise UnsupportedNemesis 


in case of collisions

fruch:
> so locking nodes 1 by one during RollingRestartCluster does not work. I'm still faced with issue #9284

what didn't work ?

cause holding all of the nodes of the cluster at the same time, sounds like a recipe for trouble, like example below  

> 
> lock all nodes before RollingRestartCluster. looks promising
> 
> FullScan got the assertion "couldn't find nodes for running:`{nemesis_label}`, are all nodes running nemesis ?" RollingRestartCluster raise UnsupportedNemesis
> 
> in case of collisions

temichus:
> > so locking nodes 1 by one during RollingRestartCluster does not work. I'm still faced with issue #9284
> 
> what didn't work ?
> 
> cause holding all of the nodes of the cluster at the same time, sounds like a recipe for trouble, like example below
> 
> > lock all nodes before RollingRestartCluster. looks promising
> > FullScan got the assertion "couldn't find nodes for running:`{nemesis_label}`, are all nodes running nemesis ?" RollingRestartCluster raise UnsupportedNemesis
> > in case of collisions

FullScan  failed with errors:
```
2024-12-02 14:50:30.865: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=9d9d7d88-9bb7-48a0-a0a0-9edd480359cd during_nemesis=RollingRestartCluster duration=1m3s node=longevity-10gb-3h-node-ope-db-node-41772392-3 select_from=keyspace1.standard1 message=FullScanAggregatesOperation operation failed: Fullscan failed - 'mapreduce_service_requests_dispatched_to_other_nodes' was not triggered
```

```
2024-12-02 15:03:31.212: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=d7d1d35c-4916-4b82-9672-16d81cdaecdc during_nemesis=RollingRestartCluster duration=1s node=longevity-10gb-3h-node-ope-db-node-41772392-1 select_from=keyspace1.standard1 message=NoHostAvailable('Unable to complete the operation against any hosts', {<Host: 10.4.3.111:9042 eu-west>: ConnectionException('Host has been marked down or removed'), <Host: 10.4.3.89:9042 eu-west>: <Error from server: code=0000 [Server error] message="mapreduce_service is shutting down">, <Host: 10.4.0.134:9042 eu-west>: ConnectionException('Host has been marked down or removed')})
```

```
2024-12-02 15:03:37.982: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=5ec56194-b552-4fea-a841-6b4677a2993d during_nemesis=RollingRestartCluster duration=2s node=longevity-10gb-3h-node-ope-db-node-41772392-1 select_from=keyspace1.standard1 message=NoHostAvailable('Unable to complete the operation against any hosts', {<Host: 10.4.3.89:9042 eu-west>: <Error from server: code=0000 [Server error] message="mapreduce_service is shutting down">, <Host: 10.4.0.134:9042 eu-west>: ConnectionException('Host has been marked down or removed'), <Host: 10.4.3.111:9042 eu-west>: <Error from server: code=0000 [Server error] message="mapreduce_service is shutting down">})
```

fruch:
@temichus 

have you tried just raising the `timout` and `connection_timeout` for the client used for the fullscans ?
so it can tolerate for nodes being down ? 

in cassandra-stress and all other tools we have retries for this prouposes, do fullscan have them ?

(FYI, you took this yourself, I didn't seen this one before you raised this PR)

temichus:
> @temichus
> 
> have you tried just raising the `timout` and `connection_timeout` for the client used for the fullscans ? so it can tolerate for nodes being down ?
> 
> in cassandra-stress and all other tools we have retries for this prouposes, do fullscan have them ?
> 
> (FYI, you took this yourself, I didn't seen this one before you raised this PR)

tried.

```
2024-12-09 14:14:30.612: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=fb92eb18-41e3-4bf0-8113-38bb4165e433 during_nemesis=RollingRestartCluster duration=40m0s node=longevity-10gb-3h-node-ope-db-node-e76e436f-1 select_from=keyspace1.standard1 message=FullScanAggregatesOperation operation failed: OperationTimedOut('errors={<Host: 10.4.2.76:9042 eu-west>: <Error from server: code=0000 [Server error] message="mapreduce_service is shutting down">}, last_host=10.4.2.234:9042'),session execution timeout 2400 is exceeded
```

```
2024-12-09 12:54:30.158: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=b1c5ce17-c805-43cd-9dda-7b0cd1e3952a during_nemesis=RollingRestartCluster duration=5m2s node=longevity-10gb-3h-node-ope-db-node-e76e436f-3 select_from=keyspace1.standard1 message=NoHostAvailable('Unable to complete the operation against any hosts', {<Host: 10.4.2.76:9042 eu-west>: ConnectionShutdown('Connection to 10.4.2.76:9042 was closed'), <Host: 10.4.2.234:9042 eu-west>: ConnectionShutdown('Connection to 10.4.2.234:19042 was closed'), <Host: 10.4.1.48:9042 eu-west>: ConnectionShutdown('Connection to 10.4.1.48:19042 was closed')})
```


```
2024-12-09 15:43:42.270 <2024-12-09 15:43:42.246>: (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=b001cd25-cf7c-4069-a1eb-56bb6a741610: type=RUNTIME_ERROR regex=std::runtime_error line_number=12014 node=longevity-10gb-3h-node-ope-db-node-e76e436f-3
2024-12-09T15:43:42.246+00:00 longevity-10gb-3h-node-ope-db-node-e76e436f-3      !ERR | scylla[9204]:  [shard 0: gms] raft_topology - drain rpc failed, proceed to fence old writes: std::runtime_error (raft topology: exec_global_command(barrier_and_drain) failed with seastar::rpc::closed_error (connection is closed))
```

But I agree with you that the statement that "fullscan broken by rolling restart" sounds strange(maybe eleven like a bug), so will continue to investigate

temichus:
@fruch @soyacz @aleksbykov 

according to @fruch,  I've back to solution with locking nodes 1 by one during RollingRestartCluster and increasing timouts

The last test run sows only 2 error messages but all nemesis pass 

```
2024-12-10 13:54:40.470: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=5ff46034-9a64-4978-a370-a35effc5c1f4 during_nemesis=RollingRestartCluster duration=40m0s node=longevity-10gb-3h-node-ope-db-node-d1a01caa-3 select_from=keyspace1.standard1 message=FullScanAggregatesOperation operation failed: OperationTimedOut("errors={'Connection defunct by heartbeat': 'Client request timeout. See Session.execute[_async](timeout)'}, last_host=10.4.0.234:9042"),session execution timeout 2400 is exceeded
```
can be https://github.com/scylladb/scylla-cluster-tests/pull/8705


and 
```
2024-12-10 11:22:38.945: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=eadc36fd-dd71-4b8c-adc5-46e1f6b664ff duration=1m56s node=longevity-10gb-3h-node-ope-db-node-d1a01caa-2 select_from=keyspace1.standard1 message=FullScanAggregatesOperation operation failed: Fullscan failed - 'mapreduce_service_requests_dispatched_to_other_nodes' was not triggered
```

is https://github.com/scylladb/scylla-enterprise/issues/4740 - will change it to warning in the next PR

temichus:
> @temichus [scylladb/scylla-enterprise#4740](https://github.com/scylladb/scylla-enterprise/issues/4740) is something happens in upgrade, I don't understand how it's related here...

wrolg link https://github.com/scylladb/scylladb/issues/21578

sdcm/nemesis.py
```diff
@@ -984,7 +963,12 @@ def disrupt_soft_reboot_node(self):
 
     @decorate_with_context(ignore_ycsb_connection_refused)
     def disrupt_rolling_restart_cluster(self, random_order=False):
-        self.cluster.restart_scylla(random_order=random_order)
+        try:
+            for node in self.cluster.nodes:
```
fruch, line 967: it's not the only place in SCT we have a rolling restart

fruch: - _enable_disable_table_encryption
- disrupt_mgmt_restore

fruch: maybe we should consider moving this logic into `restart_scylla`

temichus: node.restart_scylla and cluster.restart_scylla is used in many places(test_functional.py, mng_cli_test.py, audit.py,  artifacts_test.py, stop_compaction_test.py). But the lock mechanism is needed only when we have nemeses(in longevity tests). So let's keep it for nemeses only(at least for now)

will update _enable_disable_table_encryption and disrupt_mgmt_restore

roydahan:
First, does it solve the issue?

temichus:
> First, does it solve the issue?

yes, there is no conflict anymore between FullScan and RollingRestart. Each chooses only the  "free" node and locks it while doing an operation on ti

temichus:
after a discussion with @fruch lock mechanism is not the best because Nemeses just will not run and wait for the lock. we heed a solution to let full-scan know (somehow) that the node was rebooted during full-scan

cc @aleksbykov @roydahan

roydahan:
I'm suggesting again, to have a set of errors we can ignore approved by relevant developer and apply it in all relevant nemesis.

fruch:
> I'm suggesting again, to have a set of errors we can ignore approved by relevant developer and apply it in all relevant nemesis.

1. This is about a failure in fullscan, not nemesis
2. We already have code there that ignore fullscan results nemesis targeted the node that was the coordinator.

3. The problem we are facing is, that in rolling restart you target all nodes, without letting other parts of the code know about it, so the full scan thread doesn't know it should ignore the failure.

Lock suggested in this PR might cause high changes the nemesis doing rolling restart to be skipped in parallel nemesis cases.

I suggested that we need some way to identify that nodes restarted during the fullscan operation, but this not trivial with current event system.

One other option, since fullscan is just a stress tool, we implemented in python.
Maybe we should take the same approach, and apply retries with exponential backkoff

roydahan:
> We already have code there that ignore fullscan results nemesis targeted the node that was the coordinator.

So what's the problem to ignore also the "map reduce" specific error we are getting now?

fruch:
> > We already have code there that ignore fullscan results nemesis targeted the node that was the coordinator.
> 
> So what's the problem to ignore also the "map reduce" specific error we are getting now?

The problem is that ignore logic kicked in only if fullscan was using the same node as the current nemesis target node. which isn't the case in rolling restarts, cause we target all nodes.

If we're gonna just ignore the validation of the map reduce metric, we might as well remove it completely.
But it's just one of the symptoms we are seeing, we have there also failures of the fullscan queries themselves.

fruch:
> > One other option, since fullscan is just a stress tool, we implemented in python.
> > Maybe we should take the same approach, and apply retries with exponential backkoff
> 
> This is an option but it may hide problems.
> 
> 

How it's different from the problems it might hide in c-s when we are retrying ?

Our from current implemention that make anything into a warning if it's during a nemesis targeting the same node ?

We use retry when there's an inerit cause of failure which is expected, and we can specifically cross match.
Consider if we had implement fullscan as stress tools in multiple languages, we would have implement retries, wouldn't we ?

roydahan:
> The problem is that ignore logic kicked in only if fullscan was using the same node as the current nemesis target node. which isn't the case in rolling restarts, cause we target all nodes.
> 
> If we're gonna just ignore the validation of the map reduce metric, we might as well remove it completely. But it's just one of the symptoms we are seeing, we have there also failures of the fullscan queries themselves.

We don't need to ignore the validation, we just need to ignore that specific error we're getting now.
If there are more errors, we need to report issue about them and ignore them only if this is what's decided.
In that case, what's the problem to run full scans inside a context manager that ignores this error?

yarongilor:
if it is expected for scan to fail following server down then it shouldn't raise an error event. Instead, it could become "disruptive nemesis aware" somehow or alternatively "log aware" so it will check for node down during scan.
IIRC c-s also had such similar failures (timeout/quorum) during rolling restart nemeses.

PR #9393, opened 2024-11-27
Fix(run_unique_sequence) with multidc

aleksbykov:
When this nemesis runs with multidc configuration, disrupt_mgmt_repair and disrupt_terminate_and_replace methods could choose target node from different dc, and _shrink_cluster method periodically failed with error:
```
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4277, in _shrink_cluster
    raise Exception(error)
Exception: Not enough nodes for decommission
```

because target_node was chosen from dc where new node was not added by _grow_cluster method.

Additionaly set to run nemesis only for data nodes

### Testing

- [Job passed](https://jenkins.scylladb.com/job/scylla-staging/job/abykov/job/longevity-multi-dc-rack-aware-zero-token-dc/47)

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1638,7 +1638,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
     @target_all_nodes
     def disrupt_terminate_and_replace_node(self):  # pylint: disable=invalid-name
+        self._terminate_and_replace_node()
 
+    def _terminate_and_replace_node(self):
         def get_node_state(node_ip: str) -> List["str"] | None:
             """Gets node state by IP address from nodetool status response"""
             status = self.cluster.get_nodetool_status()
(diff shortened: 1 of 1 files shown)


sdcm/nemesis.py
```diff
@@ -4528,15 +4532,18 @@ def disrupt_run_unique_sequence(self):
             self.steady_state_latency()
             self.has_steady_run = True
         InfoEvent(message='StartEvent - start a repair by ScyllaManager').publish()
-        self.disrupt_mgmt_repair_cli()
-        InfoEvent(message='FinishEvent - Manager repair has finished').publish()
+        if self.cluster.params.get('use_mgmt') and self.cluster.params.get('use_cloud_manager'):
```
soyacz, line 4535: should be `and not`

soyacz (commented):
Can you explain how this change fixes the problem of choosing node from different dc?

fruch:
> Can you explain how this change fixes the problem of choosing node from different dc?

I guess not calling disrupt_*, won't pick new target node

aleksbykov:
> Can you explain how this change fixes the problem of choosing node from different dc?

@soyacz , @fruch already answered each disrupt_* methad patched to choose new target node. and when we have single dc it doesn't matter, but in multidc disrupt_terminate_and_replace is between _grow_cluster and _shrink_cluster, and thus target node could be switched to another DC where number of nodes was not changed and stay as initial

sdcm/nemesis.py
aleksbykov: the original condition: `if not self.cluster.params.get('use_mgmt') and not self.cluster.params.get('use_cloud_manager'):`
So here should be `or` i think

soyacz: yes

aleksbykov: fixed

fruch (approved):
LGTM

Issue #8401, opened 2024-08-19
The node that was created by `AddRemoveDc` nemesis and is been terminated, is set as target node by another nemesis


juliayakovlev:
This issue may happens when 2 run nemeses: `AddRemoveDc` and any other run in parallel. And `AddRemoveDc` should start first.
```
< t:2024-08-17 22:18:01,085 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-08-17 22:18:01.072: (DisruptionEvent Severity.NORMAL) period_type=begin event_id=b6d96f8b-66a4-44ec-89f3-89b0c340
3294: nemesis_name=AddRemoveDc target_node=Node longevity-parallel-topology-schema--db-node-38b8ee85-50 [18.202.172.255 | 10.4.9.84]
< t:2024-08-17 22:24:13,983 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-08-17 22:24:13.974: (DisruptionEvent Severity.NORMAL) period_type=begin event_id=2e58b84a-0e84-4034-bb46-5c117c17
2a94: nemesis_name=AddDropColumn target_node=Node longevity-parallel-topology-schema--db-node-38b8ee85-54 [54.228.128.54 | 10.4.8.54]
< t:2024-08-17 22:27:07,188 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-08-17 22:27:07.186: (DisruptionEvent Severity.NORMAL) period_type=end event_id=b6d96f8b-66a4-44ec-89f3-89b0c34032
94 duration=9m6s: nemesis_name=AddRemoveDc target_node=Node longevity-parallel-topology-schema--db-node-38b8ee85-50 [18.202.172.255 | 10.4.9.84]
< t:2024-08-17 22:31:58,026 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2024-08-17 22:31:58.024: (DisruptionEvent Severity.ERROR) period_type=end event_id=2e58b84a-0e84-4034-bb46-5c117c172a94 duration=7m44s: nemesis_name=AddDropColumn target_node=Node longevity-parallel-topology-schema--db-node-38b8ee85-54 [54.228.128.54 | 10.4.8.54] errors=Failed to run a command due to exception!
```

[...]

PR #9425, opened 2024-11-28
fix(nemesis): new node in add remove dc

juliayakovlev:
The issue is relevant for parallel nemesis tests.
The node that is created by AddRemoveDc nemesis, may be set as target node by another nemesis. But this node is terminated in the end of nemesis. In this case the second (running in parallel) nemesis fails.

Use 'set_current_running_nemesis' on the new node in AddRemoveDC.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/8401

### Testing

- [x] [longevity-100gb-4h](https://argus.scylladb.com/tests/scylla-cluster-tests/4457917a-976a-4e78-834e-cda07f33e699)

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4790,6 +4790,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 status = self.tester.db_cluster.get_nodetool_status()
                 new_dc_list = [dc for dc in list(status.keys()) if dc.endswith("_nemesis_dc")]
                 assert new_dc_list, "new datacenter was not registered"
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

fruch:
@juliayakovlev  you have references to jobs where you tested this ?

juliayakovlev:
> @juliayakovlev you have references to jobs where you tested this ?

Updated with job reference.
Actually, when I tested this fix, I found that there is another issue that caused by recently added variable `self._target_node_pool`. The error that is thrown by this (recently added) issue is similar as error is fixed by this commit. But root cause is different. I will open an issue for that

fruch (approved):
LGTM

PR #9433, opened 2024-12-01
improvement(syslog-ng): filter audit prints

fruch:
since in recent release of scylla audit is enabled by default we are filtering it by default for the time being.

in some nemesis code we are doing lots of separate commands which generate too many audit events, that make our logging system out of sync with the test flow, and generate false positives.

### Testing

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/86/ - [argus](https://argus.scylladb.com/tests/scylla-cluster-tests/4a3083b1-b518-4e45-bd31-e4083f8bba8b)
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/94/

### Reminders

### DIFF
--- a/sdcm/provision/common/utils.py
+++ b/sdcm/provision/common/utils.py
@@ -30,7 +30,13 @@ def configure_syslogng_target_script(hostname: str = "") -> str:
         write_syslog_ng_destination
 
         if ! grep -P "log {{.*destination\\\\(remote_sct\\\\)" /etc/syslog-ng/syslog-ng.conf; then
-            echo "log {{ source($source_name); destination(remote_sct); }};" >> /etc/syslog-ng/syslog-ng.conf
+            echo "
+        filter filter_sct {{
+            # filter audit out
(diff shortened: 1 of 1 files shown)


roydahan (approved):

soyacz (approved):
IIUC it's about os audit, not Scylla audit feature.
In that case LGTM (otherwise it will break audit nemesis).

roydahan:
Yes, OS "audit" service.

fruch:
> IIUC it's about os audit, not Scylla audit feature. In that case LGTM (otherwise it will break audit nemesis).

I might break, we'll need to test it with those nemesis as well

fruch:
As I suspected, it filters also scylla audit logs

Would to find a way to identify them from each other

I still think we can't disable auditd

fruch:
now it's filtering only the system audit, and not scylla-audit prints

PR #9332, opened 2024-11-21
fix(gce-backend): replace GCP instance types n1 with n2

dimakr:
Closes: https://github.com/scylladb/scylla-cluster-tests/issues/8971

### Testing

couple of tests where n2 instances are used:
- [x] :green_circle: [artifacts-centos9-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts-centos9-test/)
- [x] :green_circle: [longevity-5gb-1h-EndOfQuotaNemesis.yaml](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/sct-pr-provision-test/130/)

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #9439, opened 2024-12-01
fix(gce): change default number of disks to 4

fruch:
since 831116f450065f2d6b8f9d695cd4feadd6821348 changes the default instance type for gce, we need to chaneg the default disk number so we won't run into following error:

```
Number of local SSDs for an instance of type n2-highmem-8 should be one of
[0, 1, 2, 4, 8, 16, 24], while [3] is requested.
```

Ref: https://github.com/scylladb/scylla-cluster-tests/pull/9332

### Testing

- [x] 🟢 https://argus.scylladb.com/tests/scylla-cluster-tests/a17e5ddf-c3c2-4f89-96e0-9acaf5b3e5d8

### Reminders

### DIFF
--- a/defaults/gce_config.yaml
+++ b/defaults/gce_config.yaml
@@ -21,7 +21,7 @@ gce_n_local_ssd_disk_monitor: 0
 gce_instance_type_db: 'n2-highmem-8'
 gce_root_disk_type_db: 'pd-ssd'
 root_disk_size_db: 50
-gce_n_local_ssd_disk_db: 3
+gce_n_local_ssd_disk_db: 4
 
 gce_pd_standard_disk_size_db: 0
 gce_pd_ssd_disk_size_db: 0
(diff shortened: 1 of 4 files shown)


defaults/gce_config.yaml
```diff
@@ -21,7 +21,7 @@ gce_n_local_ssd_disk_monitor: 0
 gce_instance_type_db: 'n2-highmem-8'
 gce_root_disk_type_db: 'pd-ssd'
 root_disk_size_db: 50
-gce_n_local_ssd_disk_db: 3
+gce_n_local_ssd_disk_db: 4
```
dimakr, line 24: Search shows couple more k8s related places, where the number of disks is 3. Should we update there as well? 
```
❯ grep -rn 'gce_n_local_ssd_disk_db: 3' ./
./test-cases/scylla-operator/longevity-scylla-operator-multidc-12h.yaml:43:gce_n_local_ssd_disk_db: 3
./test-cases/scylla-operator/longevity-scylla-operator-multidc-3h.yaml:43:gce_n_local_ssd_disk_db: 3
./defaults/k8s_gke_config.yaml:12:gce_n_local_ssd_disk_db: 3
./defaults/gce_config.yaml:24:gce_n_local_ssd_disk_db: 3
```

fruch: if we change their instance type too, then yes

fruch: I've updated them as well to 4

roydahan (approved):

PR #9491, opened 2024-12-06
Add a helper test to prepare Manager backup snapshots

mikliapko:
The test prepares backup snapshot for its future use in restore cases (restore benchmark or nemesis). The test populates the cluster with data, runs backup and logs created snapshot details into console. Corresponding jenkinsfile and testcase yamls are added.

Together with test, the new helper method was introduced which extracts compaction strategy string from c-s cmd.

To handle the backup size, the new environmental variable was added (SCT_MGMT_PREPARE_SNAPSHOT_SIZE). It should be specified in Jenkins job.

### Testing

- [x] 🟢 2TB snapshot [job](https://argus.scylladb.com/tests/scylla-cluster-tests/0df88308-4d00-4370-abe1-016484f74b41)


fruch (approved):
LGTM

vponomaryov (approved):
LGTM

mgmt_cli_test.py
```diff
@@ -258,19 +258,45 @@ def run_verification_read_stress(self, ks_names=None):
...
+        """
+        match = re.search(r"compaction\(strategy=([^)]+)\)", cs_cmd)
+        if match:
+            strategy = match.group(1)
+            if remove_postfix:
+                strategy = re.sub(r"CompactionStrategy$", "", strategy)
+            if lower:
+                strategy = strategy.lower()
+        else:
+            raise ValueError("Compaction strategy not found in cs_cmd.")
```
soyacz, line 297: if strategy is not provided in c-s then default one is used (size tiered for oss and incremental for enterprise)

sdcm/sct_config.py
```diff
@@ -1158,6 +1158,10 @@ class SCTConfiguration(dict):
              env="SCT_MGMT_NODETOOL_REFRESH_FLAGS", type=str,
              help="Nodetool refresh extra options like --load-and-stream or --primary-replica-only"),
 
+        dict(name="mgmt_prepare_snapshot_size",
+             env="SCT_MGMT_PREPARE_SNAPSHOT_SIZE", type=int,
+             help="Size of backup snapshot in Gb to be prepared to be prepared for backup"),
```
soyacz, line 1163: double 'prepared'

mgmt_cli_test.py
```diff
@@ -1252,6 +1278,56 @@ def test_suspend_and_resume(self):
...
+        mgr_cluster = self.ensure_and_get_cluster(manager_tool)
+
+        self.log.info("Run backup and wait for it to finish")
+        backup_task = mgr_cluster.create_backup_task(location_list=self.locations, rate_limit_list=["0"])
+        backup_task_status = backup_task.wait_and_get_final_status(timeout=200000)
+        assert backup_task_status == TaskStatus.DONE, \
+            f"Backup task ended in {backup_task_status} instead of {TaskStatus.DONE}"
+
+        self.log.info("Log snapshot details")
+        self.log.info(
```
soyacz, line 1323: also if this is important, all these things could be sent to Argus as a result (text column) so these details would be visible there without peeking into logs.

soyacz (commented):
just few coments - post merge

mikliapko:
> just few coments - post merge

yeah, it makes sense to address these things. I will prepare fixes.

mgmt_cli_test.py
mikliapko: I managed to post snapshot details to Argus Results, thanks for idea!
image
@soyacz Is there any way for every consecutive run to extend the existing table with a new row, thus, keeping all the snapshots into one table without a need to browse every run in a search of snapshots?

mikliapko: Or the data on Results tab is run scoped only?

soyacz: yes, it's run scoped only

mikliapko: got it, thanks!

PR #9492, opened 2024-12-06
Support different regions in disrupt_mgmt_restore nemesis

mikliapko:
All backup snapshots used in disrupt_mgmt_restore nemesis has been duplicated into eu-west-1 region to make test functional for this region as well.

This fix adjusts yaml file with snapshots configuration and the flow nemesis uses to choose the snapshot for the test.

### Testing

- [ ]


test-cases/longevity/longevity-mv-si-4days.yaml
```diff
@@ -22,7 +22,7 @@ instance_type_monitor: 't3.xlarge'
 
 root_disk_size_runner: 120
 
-nemesis_class_name: 'SisyphusMonkey'
+nemesis_class_name: 'MgmtRestore'
```
fruch, line 25: reminder to put this back

jenkins-pipelines/manager/ubuntu22-manager-backup.jenkinsfile
```diff
@@ -5,7 +5,7 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 
 managerPipeline(
     backend: 'aws',
-    region: 'us-west-2',
```
fruch, line 8: this is for testing purposes ? or by mistake ?

mikliapko:
@fruch I'm not done with this PR yet.
Faced one more issue https://github.com/scylladb/scylla-manager/issues/4041#issuecomment-2523207414 that should be addressed.

jenkins-pipelines/manager/ubuntu22-manager-backup.jenkinsfile
mikliapko: It's by purpose, otherwise the test would fail because of region and backup bucket misconfiguration.
Before, the test was passing as `backup_bucket_region` parameter was set to `us-east-1` (means that manager agent had this region in configuration) and we used backups from the bucket at the same region.

sdcm/nemesis.py
```diff
@@ -3056,10 +3056,13 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
         cluster_backend = self.cluster.params.get('cluster_backend')
         if cluster_backend == 'k8s-eks':
             cluster_backend = 'aws'
+
         persistent_manager_snapshots_dict = get_persistent_snapshots()
-        target_bucket = persistent_manager_snapshots_dict[cluster_backend]["bucket"]
+        region = self.cluster.params.get('region_name')
```
fruch, line 3061: should we do here `.split()[0]` as done in other part of the code ?

also was did nemesis test under multi-dc case ?

fruch (changes requested):
LGTM, but we should test this with a multidc case

sdcm/nemesis.py
mikliapko: > also was did nemesis test under multi-dc case ?

No, it wasn't tested.
MultiDC scenario won't work for AWS cluster because we still have issues in Manager:
https://github.com/scylladb/scylla-manager/issues/3829
https://github.com/scylladb/scylla-manager/issues/4049

Do you have any active nemesis tests that have multiDC configuration and run `disrupt_mgmt_restore`?

fruch: we have one in tier1, and it doesn't filter this nemesis

can you add the code to skip  this nemesis using `SkipPerIssue` on those manager issues ?

mikliapko: sure, will add this check

mikliapko: > should we do here `.split()[0]` as done in other part of the code ?

Fixed to be consistent across the framework.
In current implementation with multiDC setup skip active, this change doesn't have any impact.

mikliapko: Skip per issues for multiDC setup is added as requested.

fruch (approved):
LGTM

mikliapko:
@dimakr Could you please take a look?

vponomaryov:
It's backport to the `2024.2` caused the following bug: https://github.com/scylladb/scylla-cluster-tests/issues/9775

Haven't checked the `6.2` and `6.1` branches, but they probably are affected too.

PR #9500, opened 2024-12-08
fix(multi-networks): configure scylla correctly with the right interface

fruch:
* configure `listen_interface` in scylla.yaml as needed
* return correct selected interface on `get_nic_devices()`
* run `scylla_sysconfig_setup` again on right network interface

### Testing

- [x] 🟢  https://argus.scylladb.com/tests/scylla-cluster-tests/13d394e8-b293-449c-85ee-876322dcfcbd

### Reminders

### DIFF
--- a/unit_tests/test_data/test_scylla_yaml_builders/PR-provision-test.result.json
+++ b/unit_tests/test_data/test_scylla_yaml_builders/PR-provision-test.result.json
@@ -20,5 +20,6 @@
   "alternator_enforce_authorization": false,
   "auto_bootstrap": true,
   "enable_ipv6_dns_lookup": false,
+  "listen_interface": "eth0",
   "hinted_handoff_enabled": false
 }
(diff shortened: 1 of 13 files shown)


juliayakovlev:
What is it wrong, why this change is needed?

fruch:
> What is it wrong, why this change is needed?

while investigating the job using it, I've seen that we configure all of the CPU pinning in regards of eth0
I thought it might the cause of some other issues as well I've seen in that case, but so far I showed that most of them aren't related to this change.

but it's the correct thing todo, to make sure scylla is configured correctly.

unit_tests/test_data/test_scylla_yaml_builders/PR-provision-test.result.json
```diff
@@ -20,5 +20,6 @@
   "alternator_enforce_authorization": false,
   "auto_bootstrap": true,
   "enable_ipv6_dns_lookup": false,
+  "listen_interface": "eth0",
```
soyacz, line 23: i think on GCE network interfaces have different names

fruch: we don't support multiple networks for GCP, so I don't expect it to be relevant.

I'll trigger those provision test to be on the safe side

fruch: GCP passed nicely, 
Azure failed in some weird way and missing some part of the SCT logs, I'll trigger it again

fruch: Also azure passed nicely, merging this one

fruch:
@juliayakovlev can you give one more look ?

soyacz (approved):
LGTM

juliayakovlev (approved):
looks good

PR #9502, opened 2024-12-08
fix(nemesis): move selection out of pool into `_get_target_nodes`

fruch:
since decorators can happen multiple times, and much before a nemesis might select a target node, we don't want to hold the any stale information about the db nodes, for example when multiple nemesis are called inside other nemesis.
(in NemesisSequence it can lead to selection of already remove node)

so in the commit we are change to hold just the type of the pool we expect to use, and picking from that pool only on target selection

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -103,7 +103,8 @@ from sdcm.utils.common import (get_db_tables, generate_random_string,
                                reach_enospc_on_node, clean_enospc_on_node,
                                parse_nodetool_listsnapshots,
                                update_authenticator, ParallelObject,
-                               ParallelObjectResult, sleep_for_percent_of_duration, get_views_of_base_table)
+                               ParallelObjectResult, sleep_for_percent_of_duration, get_views_of_base_table,
+                               SctDbNodesTypes)
 from sdcm.utils.features import is_tablets_feature_enabled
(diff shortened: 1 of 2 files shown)


sdcm/nemesis.py
```diff
@@ -424,7 +422,18 @@ def _get_target_nodes(
         """
         if is_seed is DefaultValue:
             is_seed = False if self.filter_seed else None
-        nodes = [node for node in self._target_node_pool if not node.running_nemesis]
+
+        match self._target_node_pool:
```
aleksbykov, line 426: target_node_pool will  always be data_type (in most cases), because set target nodes is executed in disrupt_wrapper, which is first in decorator changes, and decorator which will change the node pool type, will change it after target node was chosen

sdcm/nemesis.py
```diff
@@ -184,32 +185,32 @@ def target_data_nodes(func):
     @wraps(func)
     def wrapper(*args, **kwargs):
         try:
-            args[0].set_target_node_pool(args[0].cluster.data_nodes)
+            args[0].set_target_node_pool('data_nodes')
```
aleksbykov, line 188: may we need to rename to set_target_node_pool_type?

aleksbykov (commented):
I think we need to combine both PRs: https://github.com/scylladb/scylla-cluster-tests/pull/9517, 
@fruch  WDYT?

fruch:
close in favor of https://github.com/scylladb/scylla-cluster-tests/pull/9517

Issue #9494, opened 2024-12-08
`nodetool removenode` should call with `long_running=True`


fruch:
## Packages

Scylla version: `6.3.0~dev-20241206.7e2875d6489d` with build-id `5227dd2a3fce4d2beb83ec6c17d47ad2e8ba6f5c`

Kernel Version: `6.8.0-1019-aws`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

node-5 was remove with `nodetool remove` command running on node-6
the command run disconnected from node-6 cause of knonw ssh overload issue:
```
2024-12-07T06:42:24.802+00:00 longevity-parallel-topology-schema--db-node-f1480f18-6     !INFO | sshd[966]: Bad packet length 64330216.
```

and the nemesis failed with:
```
2024-12-07 06:42:53.463: (DisruptionEvent Severity.ERROR) period_type=end event_id=2cb3e6a3-c061-4467-b75d-ed80856b22e5 duration=54m56s: nemesis_name=RemoveNodeThenAddNode target_node=Node longevity-parallel-topology-schema--db-node-f1480f18-5 [108.128.207.201 | 10.4.8.89] errors=Node was not removed properly (Node status:{'state': 'DL', 'load': '24.85GB', 'tokens': '256', 'owns': '?', 'host_id': 'ad4da1e5-ec36-4546-bcd0-8fcf3cf468fb', 'rack': '1c'})
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5441, in wrapper
result = method(*args[1:], **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 188, in wrapper
return func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3740, in disrupt_remove_node_then_add_node
assert removed_node_status is None, \
AssertionError: Node was not removed properly (Node status:{'state': 'DL', 'load': '24.85GB', 'tokens': '256', 'owns': '?', 'host_id': 'ad4da1e5-ec36-4546-bcd0-8fcf3cf468fb', 'rack': '1c'})
```

without adding new node as replacement, so the rest fo the test is failing cause of that missing the  extra node:
```
[...]

PR #9518, opened 2024-12-09
fix(disrupt_remove_node_then_add_node): call `nodetool removenode` as long_running

fruch:
since this command can take long, and ssh connection might get interrupted we should use the `long_running=True` option, so the command would be executed on the node, and retrieve the output when it finish.

Fixes: #9494

### Testing

- [ ]

### Reminders


vponomaryov (approved):
LGTM

PR #9545, opened 2024-12-12
test(hdr logs): use thread instead process

juliayakovlev:
Use ThreadPoolExecutor instead of Process in CSHDRFileLogger class.
This change is added due to issue that c-s hdr file is not always
copied from loader to runner. As the root of the problem was not found,
it was decided to use thread.

Add HDRFileMissed warning event when found not copied HDR file.

The issue with not-copied HDR file was not reproduced while testing this commit, but I still keep a few debug log messages for a future.

### Testing

- [x] [mixed test](https://argus.scylladb.com/tests/scylla-cluster-tests/0db3cac1-4f95-455a-8a2a-c05f45898d43)
- [ ] [read test](https://argus.scylladb.com/tests/scylla-cluster-tests/1a7e60a7-fcaa-4490-b2d4-d05ae36adec0)

### Reminders

### DIFF
--- a/sdcm/stress_thread.py
+++ b/sdcm/stress_thread.py
@@ -17,6 +17,8 @@ import time
 import uuid
 import logging
 import contextlib
+from concurrent.futures import ThreadPoolExecutor
+from datetime import datetime
 from typing import Any
 from itertools import chain
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


sdcm/stress_thread.py
```diff
@@ -105,6 +128,30 @@ def validate_and_collect_hdr_file(self):
             HDRFileMissed(message=f"'{self._remote_log_file}' HDR file couldn't copied from loader {self._node.name}",
                           severity=Severity.ERROR).publish()
 
+    # @raise_event_on_failure
```
soyacz, line 131: probably not needed. anyway `self._retrieve` is catching all exceptions...

soyacz (approved):
besides one small comment, LGTM

sdcm/stress_thread.py
fruch: we should remove this comment anyhow

fruch (approved):
LGTM

soyacz (approved):
LGTM

PR #9553, opened 2024-12-15
fix(sct_config): define SCT_AWS_ACCOUNT_ID in SCTConfiguration

mikliapko:
Environmental variable SCT_AWS_ACCOUNT_ID is used in aws_okta.py:

`account_id = os.environ.get('SCT_AWS_ACCOUNT_ID', '797456418907')`

The fact it's not defined in sct_config.py fails SCT test execution for non-default AWS account, ID of which is passed as env variable:

`ValueError: Unsupported environment variables were used: -
 SCT_AWS_ACCOUNT_ID=400694452550`

The described above case is relevant for SCT tests with the cluster created in Scylla Cloud.

### Testing

- [x] None


fruch (approved):
LGTM

Issue #7653, opened 2024-06-14
`append_scylla_args` in Jenkins jobs for SCT / longevity tests


kbr-scylla:
Sometimes it's useful to run SCT tests with additional parameters passed to Scylla -- for example, recently I asked @soyacz to reproduce certain failures while enabling higher-level logging for Raft by passing `--logger-log-level raft=trace:raft_topology=trace:group0_raft_sm=trace` to Scylla: https://github.com/scylladb/scylla-enterprise/issues/3820#issuecomment-2167699606, https://github.com/scylladb/scylladb/issues/18821#issuecomment-2165880053, IIUC @soyacz did it by modifying SCT on his fork and running the job with it.

Perhaps a good investment would be to extend the regular Jenkins jobs we use with something like APPEND_SCYLLA_ARGS parameter, which would then be passed to Scylla.

This would allow e.g. choosing a different set of loggers for which we enable higher logging levels for debugging some tough issues, not necessarily related to Raft. Or generally change the Scylla configuration however we want for certain runs.

fruch:
the only problem with this one, is if the test was using this configuration, it wouldn't append to it as one might expect.
we'll need to get it figured out, but regardless of that it would be useful option for jenkins pipelines

soyacz:
maybe we could introduce `scylla_logger_settings` param? 
This should be appendable to `append_scylla_args`

fruch:
> maybe we could introduce `scylla_logger_settings` param? This should be appendable to `append_scylla_args`

logging is just an example, there could be more things one might want to enable from the command line, I don't think redefining all of them in SCT would be of much help.

we should make sure `append_scylla_args` is mergeable, and can be define multiple times, like we recently fixed `append_scylla_yaml` 

it's a bit more trick, since we'll need to flush out duplications (we have such code in ccm)

soyacz:
'mergable' will work on .yaml configs level (we should do it), but I don't see how it could work with jenkins where we pass params with env variables that override config. 
`pipeline_append_scylla_args` could be used to indicate we append to ones generated by params.

roydahan:
Now that we have the ability to change any parameter, is this still relevant?

dimakr:
> maybe we could introduce `scylla_logger_settings` param? This should be appendable to `append_scylla_args`

Appendability of additional parameters, which this change adds, is a needed thing indeed.
[...]

PR #9554, opened 2024-12-15
feature(sct_config): appendable configuration values

fruch:
introduce new option for appending to string or list configuration values

1) strings: can be appended with adding `++` at the begining of the string:
   `export SCT_APPEND_SCYLLA_ARGS="++ --overprovisioned 1"`

2) list: can be appended by adding `++` as the first item of the list
   `export SCT_SCYLLA_D_OVERRIDES_FILES='["++", "extra_file/scylla.d/io.conf"]'`

Note list would work on every config defiend with `str_or_list_or_eval` with the expection of `config_files`, `region_name`, `gce_datacenter` and `gce_datacenter` which can't be appended, and would need to be overriden completly

Ref: https://github.com/scylladb/scylla-cluster-tests/issues/7653

### Testing

- [ ]

### Reminders

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -1,10 +1,18 @@
 # scylla-cluster-tests configuration options
 
+#### Appending with environment variables or with config files
+* **strings:** can be appended with adding `++` at the beginning of the string:
+`export SCT_APPEND_SCYLLA_ARGS="++ --overprovisioned 1"`
+* **list:** can be appended by adding `++` as the first item of the list
+`export SCT_SCYLLA_D_OVERRIDES_FILES='["++", "extra_file/scylla.d/io.conf"]'`
+
 ## **config_files** / SCT_CONFIG_FILES
 
 a list of config files that would be used
 
 **default:** N/A
+
+**type:** str_or_list_or_eval
 
 
 ## **cluster_backend** / SCT_CLUSTER_BACKEND
@@ -13,12 +21,16 @@ backend that will be used, aws/gce/docker
 
 **default:** N/A
 
+**type:** str
+
 
 ## **test_method** / SCT_TEST_METHOD
 
 class.method used to run the test. Filled automatically with run-test sct command.
 
 **default:** N/A
+
+**type:** str
 
 
 ## **test_duration** / SCT_TEST_DURATION
@@ -27,12 +39,16 @@ Test duration (min). Parameter used to keep instances produced by tests<br>and f
 
 **default:** 60
 
+**type:** int
+
 
 ## **prepare_stress_duration** / SCT_PREPARE_STRESS_DURATION
 
 Time in minutes, which is required to run prepare stress commands<br>defined in prepare_*_cmd for dataset generation, and is used in<br>test duration calculation
 
 **default:** 300
+
+**type:** int
 
 
 ## **stress_duration** / SCT_STRESS_DURATION
@@ -41,12 +57,16 @@ Time in minutes, Time of execution for stress commands from stress_cmd parameter
 
 **default:** N/A
 
+**type:** int
+
 
 ## **n_db_nodes** / SCT_N_DB_NODES
 
 Number list of database data nodes in multiple data centers. To use with<br>multi data centers and zero nodes, dc with zero-nodes only should be set as 0,<br>ex. "3 3 0".
 
 **default:** N/A
+
+**type:** int_or_space_separated_ints
 
 
 ## **n_test_oracle_db_nodes** / SCT_N_TEST_ORACLE_DB_NODES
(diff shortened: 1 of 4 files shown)


dimakr (dismissed):
LGTM

fruch:
> we should somehow document it. Maybe `configuration_options.md` could have few remarks before options table stating that e.g dicts are merged, or str/lists can be appended like that?

all strings configuration can be used like that,  and everything with `type=str_or_list_or_eval`

the `configuration_options.md` need rework anyhow, it's not really usable nowdays (after github redo their UI)

fruch:
> > we should somehow document it. Maybe `configuration_options.md` could have few remarks before options table stating that e.g dicts are merged, or str/lists can be appended like that?
> 
> all strings configuration can be used like that, and everything with `type=str_or_list_or_eval`
> 
> the `configuration_options.md` need rework anyhow, it's not really usable nowdays (after github redo their UI)

https://github.com/scylladb/scylla-cluster-tests/pull/9558 should make that doc a bit more usable
and then we can introduce the information like with one is appendable on the docs

soyacz (dismissed):
Assuming we're going to document it in followup - LGTM

fruch:
> Assuming we're going to document it in followup - LGTM

now with the docs update, that would specific which one can be appended and which one can not.
not sure it's clear enough

soyacz (commented):
I don't see the code that verifies 'appendable' - shouldn't it be added and fail fast?

sdcm/sct_config.py
```diff
@@ -1821,15 +1859,15 @@ def __init__(self):  # noqa: PLR0912, PLR0914, PLR0915
...
+        merge_dicts_append_strings(self, files)
 
         # 2) load the config files
         try:
             for conf_file in list(config_files):
                 if not os.path.exists(conf_file):
                     raise FileNotFoundError(f"Couldn't find config file: {conf_file}")
             files = anyconfig.load(list(config_files))
-            anyconfig.merge(self, files)
+            merge_dicts_append_strings(self, files)
```
vponomaryov, line 1870: Commit messages says that the `config_files` is exception. So, is it?

sdcm/sct_config.py
```diff
@@ -204,13 +227,14 @@ class SCTConfiguration(dict):
 
     config_options = [
         dict(name="config_files", env="SCT_CONFIG_FILES", type=str_or_list_or_eval,
-             help="a list of config files that would be used"),
+             help="a list of config files that would be used", appendable=False),
```
vponomaryov, line 230: Do I understand it correctly that the `appendable` parameter is used only for docs?
Not limiting params possibilities?

fruch: as @soyacz suggested, I'm adding a check to see some are not wrongly used

sdcm/sct_config.py
fruch: the commit message says that about using `SCT_CONFIG_FILES='["++", "new_config.yaml"]'`
not that you can't use this feature inside config files, for example you can have two config files like that:

```yaml
# config1.yaml
append_scylla_args: '-a -b -c'
```

```yaml
# config2.yaml
append_scylla_args: '++ -d -e -f' 
```

and use them ontop of each other

some environment values are used before the configuration files, like SCT_CONFIG_FILE, SCT_REGION_NAME, SCT_CLUSTER_BACKEND, variable that holds versionss,
and using append on them is much more complicated, and doesn't make any sense (at least to me)

docs/configuration_options.md
```diff
@@ -1,2737 +1,3525 @@
...
+**type:** int
+
 
 ## **n_db_nodes** / SCT_N_DB_NODES
 
 Number list of database data nodes in multiple data centers. To use with<br>multi data centers and zero nodes, dc with zero-nodes only should be set as 0,<br>ex. "3 3 0".
 
 **default:** N/A
 
+**type:** int_or_list (appendable)
```
vponomaryov, line 69: In the code it is called `int_or_list`, yes.
But practically, we can set `int` or string like `3 3 3`.

FOr now I find it to be confusing, because if I wouldn't know, I would try to use list object for multi-dc cases.

fruch: you have a point, according to this docs

we should be able to do:
```
export SCT_N_DB_NODES='[3, 2, 3]'
```

but the code doesn't really expect a list from those, code can only handle `int` or `str`

this function should be renamed to `int_or_space_seperated_ints` :), and remove the ast.eval from it.

but that is confusing regardless of the PR that the situation.

we probably should have those values appendable, cause it's not completely defined (i.e. how do you append to int ?), those should be overridable only

fruch: I introduce that `ast.eval`, back in https://github.com/scylladb/scylla-cluster-tests/pull/1099

which is one of the first multi-dc we stabilized after the move from avocado configuration  to sct_config
I don't see any real reason for it, but I'm guessing I wanted to back out from the space separated values.

backing from it now, is a bit more tricky and would need to change lots of places that those it.

fruch: anyhow handling this in https://github.com/scylladb/scylla-cluster-tests/pull/9602, it's not for this PR

vponomaryov (approved):
LGTM

soyacz (approved):
LGTM

PR #9573, opened 2024-12-17
fix(tester): fix `reliable_replication_factor` method by setting proper RF

vponomaryov:
Various nemesis use `_prepare_test_table` method for creation of
keyspaces and tables if it is absent.
This method uses `reliable_replication_factor` property from the tester module.
    
The problem with it is the improper replication factor calculation.
It takes maximum nodes amount in a region, reduces by 1, and then uses it as RF.
In test setups which don't have those keyspaces
and have many nodes per each region,
the resulted keyspaces get replicated to much more nodes then it is needed.
    
Example: `5dcs` with `12` DB nodes each will have `RF=11`.
In summary, it will be `55` DB nodes out of `60`.
    
So, fix it by using python `min` function which will choose
either "3" or smaller value if DB nodes count requires it.

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -804,7 +804,9 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         min_nodes_dc = min([int(nodes_num) for nodes_num in n_db_nodes.split() if int(nodes_num) > 0])
         with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
             # In case tablets are enabled, it's better to set RF smaller than dc-nodes-number, so decommission is allowed.
-            return max([min_nodes_dc - 1, 1]) if is_tablets_feature_enabled(session) else min_nodes_dc
+            rf_candidate = max([min_nodes_dc - 1, 1]) if is_tablets_feature_enabled(session) else min_nodes_dc
+            # NOTE: use RF=3 at max to avoid problems on big setups
+            return min(rf_candidate, 3)
 
     @property
     def test_id(self):


vponomaryov:
With fix: [scylla-staging/valerii/vp-longevity-aws-custom-d2-workload1-multidc-big#19](https://argus.scylladb.com/tests/scylla-cluster-tests/47ea569a-6d74-471c-8b9c-2c0ae1c1a25c)
Without fix: [scylla-staging/valerii/vp-longevity-aws-custom-d2-workload1-multidc-big#18](https://argus.scylladb.com/tests/scylla-cluster-tests/9fad6d28-a270-4b1e-a750-8e09434a1624)

sdcm/nemesis.py
```diff
@@ -2069,9 +2069,10 @@ def _prepare_test_table(self, ks='keyspace1', table=None):
         test_keyspaces = self.cluster.get_test_keyspaces()
         # if keyspace or table doesn't exist, create it by cassandra-stress
         if ks not in test_keyspaces or not table_exist:
+            ks_rf = min(self.tester.reliable_replication_factor, 3)
```
fruch, line 2072: why not apply this fix in `reliable_replication_factor` ?

fruch (approved):
LGTM

I agree the reason was to make one node cluster to be able to run those nemesis
and the number should be as low as possible.

and this was a bit wasteful, and wasn't noticed cause most of our cases doesn't have that many nodes.

yarongilor (commented):
Looks like it works ok, but why not fixing `reliable_replication_factor` with:

```
        min_nodes_dc = min([int(nodes_num) for nodes_num in n_db_nodes.split() if int(nodes_num) > 0])
        with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
            # In case tablets are enabled, it's better to set RF smaller than dc-nodes-number, so decommission is allowed.
            max_rf = max([min_nodes_dc - 1, 1]) if is_tablets_feature_enabled(session) else min_nodes_dc
            return min(max_rf, 3)
```

sdcm/nemesis.py
vponomaryov: Done.

fruch (approved):
LGTM

Issue #9031, opened 2024-10-23
[disrupt_destroy_data_then_rebuild] nemises caused lots of raft_topology errors that never lead to failure


timtimb0t:
## Packages

Scylla version: `6.3.0~dev-20241018.b11d50f59191` with build-id `d5fe38f8fd12d9b834688320151f36fb8c1e050d`

Kernel Version: `6.8.0-1016-gcp`

## Issue description

New issue

During disrupt_destroy_data_then_rebuild nemises the test initiates rebuild process (as last step after data been destroyed) that caused bunch of scylla errors:

```
2024-10-19T09:16:26.885+00:00 longevity-large-partitions-200k-pks-db-node-26155658-0-1     !INFO | scylla[4507]:  [shard  0: gms] raft_topology - executing global topology command barrier_and_drain, excluded nodes: {}
2024-10-19T09:16:29.383+00:00 longevity-large-partitions-200k-pks-db-node-26155658-0-1      !ERR | scylla[4507]:  [shard  0: gms] raft_topology - drain rpc failed, proceed to fence old writes: std::runtime_error (raft topology: exec_global_command(barrier_and_drain) failed with seastar::rpc::closed_error (connection is closed))
2024-10-19T09:16:29.383+00:00 longevity-large-partitions-200k-pks-db-node-26155658-0-1     !INFO | scylla[4507]:  [shard  0: gms] raft_topology - updating topology state: advance fence version to 2108
2024-10-19T09:16:29.384+00:00 longevity-large-partitions-200k-pks-db-node-26155658-0-1     !INFO | scylla[4507]:  [shard  0: gms] raft_topology - executing global topology command barrier, excluded nodes: {}
2024-10-19T09:16:29.384+00:00 longevity-large-partitions-200k-pks-db-node-26155658-0-1  !WARNING | scylla[4507]:  [shard  0: gms] raft_topology - barrier for tablet 27470250-8dca-11ef-ae5d-04fec4937667:79 failed: seastar::broken_promise (broken promise)
2024-10-19T09:16:29.384+00:00 longevity-large-partitions-200k-pks-db-node-26155658-0-1      !ERR | scylla[4507]:  [shard  0: gms] raft_topology - topology change coordinator fiber got error std::runtime_error (raft topology: exec_global_command(barrier) failed with seastar::rpc::closed_error (connection is closed))
```
The scylla itself is alive and these errors never caused a failure. Seems like they occurred due to deleted data

[...]

PR #9580, opened 2024-12-18
fix(nemesis): filter raft-topology errors when starting/stopping nodes

timtimb0t:
raft is generating the following errors:
raft_topology - topology change coordinator fiber got error std::runtime_error
when one of the nodes is stopped, it was decided we can safely ignore those errors

fixes: #9031

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -673,6 +673,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             self.target_node.wait_jmx_up()
         self.cluster.wait_for_schema_agreement()
 
+    @decorate_with_context(ignore_raft_topology_cmd_failing)
     @target_all_nodes
     def disrupt_stop_wait_start_scylla_server(self, sleep_time=300):  # pylint: disable=invalid-name
         self.target_node.stop_scylla_server(verify_up=False, verify_down=True)
@@ -1141,7 +1142,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
 
         return sstables
 
-    @decorate_with_context(ignore_ycsb_connection_refused)
+    @decorate_with_context([ignore_ycsb_connection_refused, ignore_raft_topology_cmd_failing])
     def _destroy_data_and_restart_scylla(self, keyspaces_for_destroy: list = None, sstables_to_destroy_perc: int = 50):  # pylint: disable=too-many-statements
         tables = self.cluster.get_non_system_ks_cf_list(db_node=self.target_node, filter_empty_tables=False,
                                                         filter_by_keyspace=keyspaces_for_destroy)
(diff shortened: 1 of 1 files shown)


sdcm/nemesis.py
```diff
@@ -1152,35 +1153,36 @@ def _destroy_data_and_restart_scylla(self, keyspaces_for_destroy: list = None, s
         self.log.debug("Chosen tables: %s", tables)
 
         # Stop scylla service before deleting sstables to avoid partial deletion of files that are under compaction
-        self.target_node.stop_scylla_server(verify_up=False, verify_down=True)
+        with ignore_raft_topology_cmd_failing():
+            self.target_node.stop_scylla_server(verify_up=False, verify_down=True)
```
aleksbykov, line 1157: Here i think better to use Context manager only for operations stop/start

aleksbykov (dismissed):
LGTM

aleksbykov:
@timtimb0t , fix only precommits

sdcm/nemesis.py
```diff
@@ -1152,7 +1153,8 @@ def _destroy_data_and_restart_scylla(self, keyspaces_for_destroy: list = None, s
...
 
         try:
```
soyacz, line 1159: shouldn't this block be inside context?

soyacz (commented):
Please add description to commit why this change is needed, so when reviewing in git blane one does not have to peek into gh.

Besides, is it required for every Scylla restart or only in places when there's a wait between stop and start? If so, I suppose there could be more places needed to adjust.

sdcm/nemesis.py
timtimb0t: > shouldn't this block be inside context?

I thought the same, but after a short conversation with @aleksbykov, I decided to do everything as is, because once the node is stopped, the cluster will recognize this and won’t send any more requests to it.

timtimb0t:
> Please add description to commit why this change is needed, so when reviewing in git blane one does not have to peek into gh.
> 
> Besides, is it required for every Scylla restart or only in places when there's a wait between stop and start? If so, I suppose there could be more places needed to adjust.

You are right; there may be other places that need a similar change, but this fix is a response to the current issue. If such an error occurs again, we will likely need a new PR.

sdcm/nemesis.py
```diff
@@ -675,14 +675,15 @@ def _kill_scylla_daemon(self):
 
     @target_all_nodes
     def disrupt_stop_wait_start_scylla_server(self, sleep_time=300):  # pylint: disable=invalid-name
```
fruch, line 677: I would recommend using @decorate_with_context, it would be less of interruption to the code

sdcm/nemesis.py
fruch: why ? 

`@decorate_with_context` that already exists on this function is a natural fit for this, and was design for those cases exactly

fruch:
> Please add description to commit why this change is needed, so when reviewing in git blane one does not have to peek into gh.
> 
> Besides, is it required for every Scylla restart or only in places when there's a wait between stop and start? If so, I suppose there could be more places needed to adjust.

Yes commit should always have description, also as for the title
`sct lib` doesn't mean much, in this case `fix(nemesis): ...`

as for the description should be something more like

```
fix(nemesis): filter raft-topology errors when starting/stopping nodes

raft is generating the following errors:

[ one example of such print ]

when one of the nodes is stopped, it was decided we can safely ignore those errors

Ref: [scylla issue about how core would address that]
Fixes: #9031
```

in such manner people can understand the reasoning, and are don't need to go for walk across multiple other issue to understand it.

soyacz:
@timtimb0t please add proper backport labels

fruch (approved):
LGTM

Issue #8722, opened 2024-09-16
c-s fails to prepare test table during `disrupt_truncate` nemesis, but the test continuous and starts the disruption


dimakr:
At the beginning of `disrupt_truncate` nemesis the test ks/table are prepared with the c-s command:
```
< t:2024-09-14 12:46:50,787 f:stress_thread.py l:325  c:sdcm.stress_thread   p:INFO  > cassandra-stress write no-warmup n=400000 cl=QUORUM -mode native cql3  user=cassandra password=cassandra -schema keyspace=ks_truncate 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -log interval=5 -transport 'truststore=/etc/scylla/ssl_conf/truststore.jks truststore-password=cassandra' -node 10.0.0.5,10.0.0.6,10.0.0.7,10.0.0.8,10.0.0.14 -errors skip-unsupported-columns
```

The command fails with the error:
```
WARN  [cluster1-nio-worker-5] 2024-09-14 12:46:55,827 RequestHandler.java:303 - Query '[0 bound values] CREATE KEYSPACE IF NOT EXISTS "ks_truncate" WITH replication = {'class': 'org.apache.cassandra.locator.NetworkTopologyStrategy', 'replication_factor' : '3'} AND durable_writes = true;' generated server side warning(s): Tables in this keyspace will be replicated using Tablets and will not support CDC, LWT and counters features. To use CDC, LWT or counters, drop this keyspace and re-create it without tablets by adding AND TABLETS = {'enabled': false} to the CREATE KEYSPACE statement.
WARN  [cluster1-worker-1] 2024-09-14 12:46:56,840 ReplicationStategy.java:204 - Error while computing token map for keyspace ks_truncate with datacenter eastus_nemesis_dc: could not achieve replication factor 3 (found 1 replicas only), check your keyspace replication settings.
WARN  [cluster1-worker-2] 2024-09-14 12:46:57,282 ReplicationStategy.java:204 - Error while computing token map for keyspace ks_truncate with datacenter eastus_nemesis_dc: could not achieve replication factor 3 (found 1 replicas only), check your keyspace replication settings.
java.lang.RuntimeException: Encountered exception creating schema
	at org.apache.cassandra.stress.settings.SettingsSchema.createKeySpacesNative(SettingsSchema.java:105)
	at org.apache.cassandra.stress.settings.SettingsSchema.createKeySpaces(SettingsSchema.java:74)
[...]

PR #9585, opened 2024-12-18
fix(nemesis.py): abort nemesis on stress command failure

dimakr:
Abort the nemesis flow early when a stress command fails, as continuing it is often invalid due to the cluster being in an unexpected state.
For example, if the _prepare_test_table routine fails, subsequent steps that depend on the test table (or attempt disruptions on top of it) will also fail.

This change adds a check for results of stress command triggered within the nemesis code, ensuring the nemesis halts early if the stress command is unsuccessful.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/8722

### Testing

- [x] [GrowShrinkClusterNemesis, both c-s and s-b loads are executed during the nemesis](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-aws-test/31/)
- [x] [TruncateMonkey, both c-s and s-b loads are executed during the nemesis](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-aws-test/33/)
- [x] Tested locally to ensure/simulate a failed stress command - added an invalid parameter to cassandra-stress command in Nemesis._prepare_test_table (here https://github.com/scylladb/scylla-cluster-tests/blob/532ed3f5708429402eb83d10096b348cdd211442/sdcm/nemesis.py#L2072), which makes the stress command fail.
Executed PR-provision-test configuration for TruncateMonkey nemesis class. As a result the disrupt_truncate disruption was started few times, but interrupted early as the stress command in  _prepare_test_table
was failing:
```
< t:2024-12-20 22:04:24,417 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > sdcm.exceptions.NemesisStressFailure: Aborting 'TruncateMonkey' nemesis as 'cassandra-stress write whoa n=400000 cl=QUORUM -mode native cql3 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -log interval=5' stress command failed with the following errors:
< t:2024-12-20 22:04:24,417 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  >   on node 'PR-provision-test-dmitriy-loader-node-a2eaf144-2': ['Stress command completed with bad status 1: ']
< t:2024-12-20 22:04:24,417 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  >   on node 'PR-provision-test-dmitriy-loader-node-a2eaf144-1': ['Stress command completed with bad status 1: ']
```

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -98,6 +98,7 @@ from sdcm.sct_events.loaders import CassandraStressLogEvent, ScyllaBenchEvent
 from sdcm.sct_events.nemesis import DisruptionEvent
 from sdcm.sct_events.system import InfoEvent, CoreDumpEvent
 from sdcm.sla.sla_tests import SlaTests
+from sdcm.stress_thread import DockerBasedStressThread
 from sdcm.utils.aws_kms import AwsKms
 from sdcm.utils import cdc
 from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
@@ -155,6 +156,7 @@ from sdcm.exceptions import (
     AuditLogTestFailure,
     BootstrapStreamErrorFailure,
     QuotaConfigurationFailure,
+    NemesisStressFailure,
 )
 from test_lib.compaction import CompactionStrategy, get_compaction_strategy, get_compaction_random_additional_params, \
     get_gc_mode, GcMode, calculate_allowed_twcs_ttl_borders, get_table_compaction_info
@@ -2075,6 +2077,23 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             cs_thread = self.tester.run_stress_thread(
                 stress_cmd=stress_cmd, keyspace_name=ks, stop_test_on_failure=False, round_robin=True)
             cs_thread.verify_results()
+            self.stop_nemesis_on_stress_errors(cs_thread)
+
+    def stop_nemesis_on_stress_errors(self, stress_thread: DockerBasedStressThread) -> None:
+        # Some implementations of stress threads override logic of the base class method
+        # DockerBasedStressThread.get_results() and filter out 'events' portion of a result (e.g. c-s stress thread).
+        # To retrieve all results of a thread we need to call the base class method directly
+        stress_results = super(stress_thread.__class__, stress_thread).get_results()
+        node_errors = {}
+        for node, result, event in stress_results:
+            if event.get('errors'):
+                node_errors.setdefault(node.name, []).extend(event.errors)
+
+        if len(node_errors) == len(stress_results):  # stop only if stress command failed on all loaders
+            errors_str = ''.join(f"  on node '{node_name}': {errors}\n" for node_name, errors in node_errors.items())
+            raise NemesisStressFailure(
+                f"Aborting '{self.__class__.__name__}' nemesis as '{stress_thread.stress_cmd}' stress command failed "
+                f"with the following errors:\n{errors_str}")
 
     @scylla_versions(("5.2.rc0", None), ("2023.1.rc0", None))
     def _truncate_cmd_timeout_suffix(self, truncate_timeout):  # pylint: disable=no-self-use
@@ -2121,6 +2140,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         bench_thread = self.tester.run_stress_thread(
             stress_cmd=stress_cmd, stop_test_on_failure=False)
         self.tester.verify_stress_thread(bench_thread)
+        self.stop_nemesis_on_stress_errors(bench_thread)
 
         # In order to workaround issue #4924 when truncate timeouts, we try to flush before truncate.
         with adaptive_timeout(Operations.FLUSH, self.target_node, timeout=HOUR_IN_SEC * 2):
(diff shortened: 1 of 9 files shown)


fruch:
@dimakr 

can you run individual nemesis on at least two of the changed nemesis, to show case it's working (it can be AWS or docker backend if they are relevant)

sdcm/nemesis.py
```diff
@@ -4422,6 +4438,7 @@ def run_write_scylla_bench_load(write_cmd):
                     regex=".*sstable - Error while linking SSTable.*filesystem error: stat failed: No such file or directory.*"):
                 write_thread = self.tester.run_stress_thread(stress_cmd=write_cmd, stop_test_on_failure=False)
                 self.tester.verify_stress_thread(write_thread)
+                self.stop_nemesis_on_stress_errors(write_thread)
```
vponomaryov, line 4441: Here we run the same command on each of the existing loaders.
In this case we are totally ok to get `n_loaders - 1` failures.
We need result at least from some of them.

Probably it is the case for some other nemesis.

It runs a single command on multiple loaders just because of the default parameter `round_robin=False` defined in the `self.tester.run_stress_thread`.

dimakr:
> can you run individual nemesis on at least two of the changed nemesis, to show case it's working (it can be AWS or docker backend if they are relevant)

Executed `longevity-5gb-1h-nemesis` config for GrowShrinkClusterNemesis and TruncateMonkey nemeses. Linked the jobs in the PR descriptions. In both cases the c-s and s-b commands are executed (and stop_nemesis_on_error check is performed) during the nemesis.

sdcm/nemesis.py
dimakr: Changed the check in the new method so that a nemesis would be aborted if all 'n_loaders` stress commands failed. Checked this for both cases - `round_robin` enabled and disabled.

sdcm/scylla_bench_thread.py
```diff
@@ -251,7 +251,7 @@ def _run_stress(self, loader, loader_idx, cpu_idx):  # pylint: disable=too-many-
             except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
                 self.configure_event_on_failure(stress_event=scylla_bench_event, exc=exc)
 
-        return loader, result
+        return loader, result, scylla_bench_event
```
fruch, line 254: Maybe we need to align this on all stress commands ?

sdcm/nemesis.py
```diff
@@ -2075,6 +2077,20 @@ def _prepare_test_table(self, ks='keyspace1', table=None):
             cs_thread = self.tester.run_stress_thread(
                 stress_cmd=stress_cmd, keyspace_name=ks, stop_test_on_failure=False, round_robin=True)
             cs_thread.verify_results()
+            self.stop_nemesis_on_stress_errors(cs_thread)
+
+    def stop_nemesis_on_stress_errors(self, stress_thread: DockerBasedStressThread) -> None:
+        stress_results = super(stress_thread.__class__, stress_thread).get_results()
```
fruch, line 2083: Why is the super call needed here ? What are we trying to skip ? And why ?

sdcm/scylla_bench_thread.py
dimakr: This new `stop_nemesis_on_stress_errors` method is now applied to results of stress commands that are triggered from nemesis.py. There are 3 commands that are executed this way in nemesis code - c-s, s-b, and cql-stress-cassandra-stress.
All of them (with this^ change included) are returning tuple of (loader, result, event).

I'm not sure, if/which stress commands are also triggered from nemesis.py.

sdcm/nemesis.py
dimakr: This is to directly call the `get_results` method to get all results from threads. Some command specific `get_results` implementation would filter out failed commands. E.g. in c-s the `get_results` performs:
```
    def get_results(self) -> list[dict | None]:
        ret = []
        results = super().get_results()

        for _, result, event in results:
            if not result:
                # Silently skip if stress command threw error, since it was already reported in _run_stress
                continue
...
        return ret
```

sdcm/scylla_bench_thread.py
fruch: I think regardless of what is being used right now, it would be good to align them  to return identically from this function.

sdcm/nemesis.py
fruch: It's a bit specific c-s, we should at least put a comment explaining it, it's quite unusual to skip up the MRO like that, I barely understand the reason it's c-s is doing that, let's at least point to the reason we jump it.

sdcm/scylla_bench_thread.py
vponomaryov: Agree, would be less confusion if we had single return type for all cases.

sdcm/nemesis.py
dimakr: added the comment

sdcm/scylla_bench_thread.py
dimakr: Aligned the return value for other stress command's `run_tress'`API.

I had to add a simple local implementation of get_results() for most of stress commands, so that we don't have problems in places like here https://github.com/scylladb/scylla-cluster-tests/blob/2bfc7917dc3997925f8805797395dfd17185858a/sdcm/tester.py#L2335
But this is a temporary mechanism. After the followup task https://github.com/scylladb/qa-tasks/issues/1826 is implemented, the `(get_/verify_)results()` API will be streamlined across stress commands.

vponomaryov (approved):
LGTM

fruch (approved):

PR #9590, opened 2024-12-19
fix(manager): update expected error message for enospc restore test

mikliapko:
Error message that Manager returns for enospc scenario has been changed to more generic one. So, this change updates expected error message and removes skip per issue since it was closed without resolution.

refs:
https://github.com/scylladb/scylla-manager/issues/4087

### Testing

- [x] https://jenkins.scylladb.com/view/scylla-manager/job/manager-3.4/job/sct-feature-test-backup/6/

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -46,7 +46,6 @@ from sdcm.cluster import TestConfig
 from sdcm.nemesis import MgmtRepair
 from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
 from sdcm.utils.common import reach_enospc_on_node, clean_enospc_on_node
-from sdcm.utils.issues import SkipPerIssues
 from sdcm.utils.loader_utils import LoaderUtilsMixin
 from sdcm.utils.time_utils import ExecutionTimer
 from sdcm.sct_events.system import InfoEvent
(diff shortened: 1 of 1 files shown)


soyacz (dismissed):
LGTM

mgmt_cli_test.py
```diff
@@ -831,12 +830,11 @@ def test_enospc_before_restore(self):
...
 
-                if not SkipPerIssues("scylladb/scylla-manager#4087", self.params):
-                    full_progress_string = restore_task.progress_string(parse_table_res=False,
-                                                                        is_verify_errorless_result=True).stdout
-                    assert "not enough disk space" in full_progress_string.lower(), \
-                        f"The restore failed as expected when one of the nodes was out of disk space, " \
-                        f"but with an ill fitting error message: {full_progress_string}"
+                full_progress_string = restore_task.progress_string(parse_table_res=False,
+                                                                    is_verify_errorless_result=True).stdout
+                assert "failed to restore sstables from location" in full_progress_string.lower(), \
```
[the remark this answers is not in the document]
mikliapko: For now, I'd probably drop this assertion. 
In case of current, very common error message, it actually doesn't make much sense.
In future when we implement this, let's say, "smart" free space checker you mentioned, we can think about this test rework.

mgmt_cli_test.py
```diff
@@ -830,13 +829,6 @@ def test_enospc_before_restore(self):
                 assert final_status == TaskStatus.ERROR, \
                     f"The restore task is supposed to fail, since node {target_node} lacks the disk space to download" \
                     f"the snapshot files"
-
```
soyacz, line 833: could be useful to at least print error mesage with log.info level for clarity in case someone once needed to verify the flow.

soyacz (approved):
Besides one comment,
LGTM

mgmt_cli_test.py
mikliapko: What kind of message you'd like to see in log.info for such case?
I supposed the assertion message is quite enough to understand the testing scenario if someone wants to verify it.

soyacz: I meant to log `full_progress_string = restore_task.progress_string(parse_table_res=False,
                                                                        is_verify_errorless_result=True).stdout` instead of asserting it (as its prone to change)

mikliapko: I got your point, but it seems to me it makes little sense.. When we analyze Manager failures, we usually look into sct python log. It contains all sctool outputs logged. `full_progress_string` can be easily found there.

soyacz: ok, looks it's good to go then

mikliapko:
It was mistakenly closed without merging, reopening it

Michal-Leszczynski (approved):

mikliapko:
@fruch Could you please merge it?

vponomaryov (approved):
LGTM

PR #9602, opened 2024-12-22
fix(sct_config): rename `int_or_list` to `int_or_space_separated_ints`

fruch:
the code handling those config values is only handling int, or string and the function name should mention that.

also remove the fallback to `ast.eval`, since it can produce anything and not just str or int values, which the code would fail on.

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -122,7 +122,7 @@ def str_or_list_or_eval(value: Union[str, List[str]]) -> List[str]:
     raise ValueError(f"{value} isn't a string or a list")
 
 
-def int_or_list(value):
+def int_or_space_separated_ints(value):
     try:
         value = int(value)
         return value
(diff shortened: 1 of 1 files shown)


sdcm/sct_config.py
```diff
@@ -228,18 +224,18 @@ class SCTConfiguration(dict):
                   Time in minutes, Time of execution for stress commands from stress_cmd parameters
                   and is used in test duration calculation
              """),
-        dict(name="n_db_nodes", env="SCT_N_DB_NODES", type=int_or_list,
+        dict(name="n_db_nodes", env="SCT_N_DB_NODES", type=int_or_comma_separated_int,
```
soyacz, line 227: but we use format like: `n_db_nodes: '5 5'` - it's not int nor comma separated int... shouldn't be `int_or_space_separated_int`?

fruch: yeah, it should be space separated

fruch: changed the name (in code, docs, commit/PR)

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

Issue #9403, opened 2024-11-28
Create a backup performance test for scylla backup implementation


regevran:
When this issue is done we have a measurable, repeatable and consistent test for backing up a cluster using `scylla`.
These should be a group of tests, i.e. one configurable test with varies: nodes, data size.

PR #9605, opened 2024-12-23
test(backup): Measure read/write latency during back up

kreuzerkrieg:
RW latency during backup using `rclone`

Test requirements
* 50% reads and 50% writes.
* Compaction enabled.
* 75%-85% CPU utilization (see discussion)
* Expect up to 6-7ms P99 read latencies.
* Expect not more than 10ms read latency during backup

Argus run
https://argus.scylladb.com/tests/scylla-cluster-tests/8d8f99b0-4e9f-4ba0-90ed-8ca970416208
Results:
|  | read time [s] | op rate [op/s] | partition rate [pk/s] | row rate [row/s] | latency 99th percentile [ms] |
| --- | --- | --- | --- | --- | --- |
| Read stress: w/o concurrent backup | 00:10:24 | 47344.75 | 47344.75 | 47344.75 | 7.60 |
| Write stress: w/o concurrent backup | 00:10:24 | 48609.50 | 48609.50 | 48609.50 | 7.75 |
| Read stress: with concurrent backup | 00:10:22 | 47168.00 | 47168.00 | 47168.00 | 7.80 |
| Write stress: with concurrent backup | 00:10:22 | 47856.00 | 47856.00 | 47856.00 | 7.85 |

|  | backup time [s] |
| --- | --- |
| Backup | 00:09:28 |
| Backup during read stress | 00:18:13 |

Another run: https://argus.scylladb.com/tests/scylla-cluster-tests/cf8d61a1-1b10-45e3-a8b5-312fda3c3842

|  | read time [s] | op rate [op/s] | partition rate [pk/s] | row rate [row/s] | latency 99th percentile [ms] |
| --- | --- | --- | --- | --- | --- |
| Read stress: w/o concurrent backup | 00:10:21 | 45589.50 | 45589.50 | 45589.50 | 7.22 |
| Write stress: w/o concurrent backup | 00:10:21 | 47901.25 | 47901.25 | 47901.25 | 6.95 |
| Read stress: with concurrent backup | 00:10:22 | 48568.25 | 48568.25 | 48568.25 | 7.10 |
| Write stress: with concurrent backup | 00:10:22 | 49467.75 | 49467.75 | 49467.75 | 7.18 |

|  | backup time [s] |
| --- | --- |
| Backup | 00:09:44 |
| Backup during read stress | 00:18:07 |

We can safely say that nor read neither write are affected by backup

@bhalevy / @regevran FYI


mikliapko:
Added qa-mainteiners to Reviewers

mgmt_cli_test.py
```diff
@@ -1679,52 +1679,77 @@ def create_backup_and_report(self, mgr_cluster, label: str):
...
         for command in self.params.get('stress_read_cmd'):
-            stress_queue.append(self.run_stress_thread(command, round_robin=True, stop_test_on_failure=False))
+            if " write " in command:
+                stress_queue["write"].append(self.run_stress_thread(command))
+            elif " read " in command:
+                stress_queue["read"].append(self.run_stress_thread(command))
+            else:
+                raise InvalidArgument("Unknown stress command")
+
+        def get_stress_averages(queue):
```
mikliapko, line 1692: I'm just wondering if we can benefit from this method somewhere else?
Probably, it makes sense to move it outside the method (`StressLoadOperations`) making it available across the file?
Btw, it can be a case for the whole `run_read_stress_and_report` method.
WDYT?

mikliapko (changes requested):

soyacz:
If you have expectations for the metrics (like p99's below 8ms) consider using `ValidationRule(fixed_limit=8)` similarly for other metrics. See `sdcm.argus_results.ManagerRestoreBanchmarkResult` for example.

regevran:
Fixes: #9403

configurations/manager/100GB_dataset.yaml
```diff
@@ -5,10 +5,8 @@ prepare_write_cmd: [ "cassandra-stress write cl=ALL n=26214400 -schema 'replicat
...
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
+stress_read_cmd: [ "cassandra-stress read  cl=ALL duration=10m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=100 -col 'size=FIXED(1024) n=FIXED(1)'",
+                   "cassandra-stress write cl=ALL duration=10m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=100 -col 'size=FIXED(1024) n=FIXED(1)'", ]
```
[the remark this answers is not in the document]
kreuzerkrieg: ok, found in the code, we usually use `stress_cmd` when we want to mix different stress patterns. Looks like it is good enough

mgmt_cli_test.py
```diff
@@ -1679,52 +1679,77 @@ def create_backup_and_report(self, mgr_cluster, label: str):
...
+        stress_queue = {"read": [], "write": []}
 
         for command in self.params.get('stress_read_cmd'):
-            stress_queue.append(self.run_stress_thread(command, round_robin=True, stop_test_on_failure=False))
+            if " write " in command:
+                stress_queue["write"].append(self.run_stress_thread(command))
+            elif " read " in command:
+                stress_queue["read"].append(self.run_stress_thread(command))
+            else:
+                raise InvalidArgument("Unknown stress command")
```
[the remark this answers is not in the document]
kreuzerkrieg: Done

configurations/manager/100GB_dataset.yaml
```diff
@@ -5,10 +5,8 @@ prepare_write_cmd: [ "cassandra-stress write cl=ALL n=26214400 -schema 'replicat
...
 
-stress_read_cmd: [ "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..26214400",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
+stress_read_cmd: [ "cassandra-stress read  cl=ALL duration=10m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=100 -col 'size=FIXED(1024) n=FIXED(1)'",
```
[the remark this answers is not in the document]
kreuzerkrieg: after [there is no more](https://github.com/scylladb/scylla-cluster-tests/pull/9605#discussion_r1899141778) `stress_read_cmd` the only issue is the `100GB_dataset.yaml` file name and its location

mgmt_cli_test.py
kreuzerkrieg: Dunno. It feels like, no one needed it for X years then why to bother and do anything no one will use in another Y years :)

kreuzerkrieg:
> If you have expectations for the metrics (like p99's below 8ms) consider using `ValidationRule(fixed_limit=8)` similarly for other metrics. See `sdcm.argus_results.ManagerRestoreBanchmarkResult` for example.

I strongly oppose this type of validation as it may cause more harm than good. Occasionally, factors like "noisy neighbors" can cause failures, leading people to spend days troubleshooting. Moreover, this validation is specific to a particular type of AWS instance. Using any other instance type will result in failure.

configurations/manager/100GB_dataset.yaml
```diff
@@ -5,10 +5,8 @@ prepare_write_cmd: [ "cassandra-stress write cl=ALL n=26214400 -schema 'replicat
...
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
+stress_cmd: [ "cassandra-stress read  cl=ALL duration=10m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=100 -col 'size=FIXED(1024) n=FIXED(1)'",
+              "cassandra-stress write cl=ALL duration=10m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=100 -col 'size=FIXED(1024) n=FIXED(1)'", ]
```
roydahan, line 9: If you want to measure latency of these operations, you need to throttle the throughput by using:
" -rate threads= x **fixed=YY/s**" 

The value of "fixed" is the "throughput / num of loaders" the test is targeting.
We usually set it to 50% of max throughput.
The target of "cpu utilization" is irrelevant and shouldn't be part of the test consideration.

roydahan: Why are you using cl=ALL?
Why are you using 2 separate commands and not one stress commands that is doing "mixed" workload.

roydahan: The command doesn't define the distribution, better to state it even if it uses the default "seq".

configurations/manager/100GB_dataset.yaml
```diff
@@ -5,10 +5,8 @@ prepare_write_cmd: [ "cassandra-stress write cl=ALL n=26214400 -schema 'replicat
...
 
-instance_type_db: 'i3en.3xlarge'
+instance_type_db: 'i4i.4xlarge'
 instance_type_loader: 'c6i.xlarge'
```
roydahan, line 12: why is it part of configuration and not a specific test yaml?
I think this entire file need to be a test yaml.

roydahan: An important configuration need to be taken care of is the num of loaders, and it should be all part of one yaml.

roydahan (changes requested):
See my comments about the stress commands.
+
It needs to have a jenkins pipeline.

configurations/manager/100GB_dataset.yaml
bhalevy: > The target of "cpu utilization" is irrelevant and shouldn't be part of the test consideration.

Irrelevant to what/who? and why?
When sizung and tuning clusters in the field do you think tgat cystomers care much about sone benchmark?
That can run their app or simulated workload tgat represent it and tune the system to certain utilizantion level (cpu, network, disk throughput, erc.) to leave enough headroom for maintenance load (compaction, repair, backup, etc.(
But if in QA, the bar is currently set at 50% of the measured throughput, that's fine too for testing consistency purposes.

roydahan: The CPU utilization is irrelevant exactly because you don't know what this CPU means to the customer.
10% of it could be serving statements and the rest is doing compactions, streaming handling the backup or whatever.
You don't set your bar because this is your comfort zone, you set the bar at something meaningful to the customer, e.g. 50% of throughput.

mgmt_cli_test.py
```diff
@@ -1679,52 +1678,77 @@ def create_backup_and_report(self, mgr_cluster, label: str):
...
+        for command in self.params.get('stress_cmd'):
+            if " write " in command:
+                stress_queue["write"].append(self.run_stress_thread(command))
+            elif " read " in command:
+                stress_queue["read"].append(self.run_stress_thread(command))
+            else:
+                raise ValueError("Unknown stress command")
+
+        def get_stress_averages(queue):
+            averages = {'op rate': 0.0, 'partition rate': 0.0, 'row rate': 0.0, 'latency 99th percentile': 0.0}
```
soyacz, line 1692: Generally it's wrong to average latencies.
Consider crating a test named "mixed_latency_during_backup" and using `@latency_calculator_decorator` - we have all these implemented there in a proper way and is sending to Argus all needed stuff out of the box.

mgmt_cli_test.py
```diff
@@ -1677,52 +1678,10 @@ def create_backup_and_report(self, mgr_cluster, label: str):
...
-        write_stress_report = {
-            "read time": int(stress_timer.duration.total_seconds()),
-            "op rate": write_stats['op rate'],
-            "partition rate": write_stats['partition rate'],
-            "row rate": write_stats['row rate'],
-            "latency 99th percentile": write_stats['latency 99th percentile'],
-        }
-        self.report_to_argus(ManagerReportType.READ, write_stress_report, "Write stress: " + label)
+    @latency_calculator_decorator
+    def mixed_latency_during_backup(self):
```
soyacz, line 1682: It's not the method that should have workload type in name, rather test - need to rename `test_backup_benchmark` to `test_backup_benchmark_mixed`
or wait for: https://github.com/scylladb/scylla-cluster-tests/issues/9627

configurations/manager/100GB_dataset.yaml
```diff
@@ -5,10 +5,7 @@ prepare_write_cmd: [ "cassandra-stress write cl=ALL n=26214400 -schema 'replicat
...
 
-stress_read_cmd: [ "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..26214400",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
-                   "cassandra-stress read cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=50 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
+stress_cmd: [ "cassandra-stress mixed cl=QUORUM duration=10m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate 'threads=100 fixed=5000/s' -col 'size=FIXED(1024) n=FIXED(1)'" ]
```
soyacz, line 8: one more thing, latency_calculator_decorator requires hdr reports, to enable them add `use_hdr_cs_histogram: true` to test config

kreuzerkrieg: done

mgmt_cli_test.py
kreuzerkrieg: refactored

configurations/manager/100GB_dataset.yaml
kreuzerkrieg: done

configurations/manager/100GB_dataset.yaml
kreuzerkrieg: done

kreuzerkrieg:
> See my comments about the stress commands. + It needs to have a jenkins pipeline.

What we are going to achieve by adding (clonning) a new pipeline?

test-cases/manager/manager-backup-restore-baseline.yaml
```diff
@@ -0,0 +1,27 @@
...
+prepare_write_cmd: [ "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..26214400",
+                     "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=26214401..52428800",
+                     "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=52428801..78643200",
+                     "cassandra-stress write cl=ALL n=26214400 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=78643201..104857600" ]
+
+stress_cmd: "cassandra-stress mixed cl=QUORUM duration=10m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native  -rate 'threads=100 fixed=17000/s' -col 'size=FIXED(1024) n=FIXED(1)'"
```
roydahan, line 11: What did you base the "17000/s" on?
A typical max throughput test for mixed workload using i4i.4xlarge achieves around 500K op/s. 
4 loaders running 17K op/s is equal to 68K op/s which is about 13% of the max throughput.
(BTW, to achieve that we use 620 threads and bigger loaders - you can probably start with c7i.2xlarge).

lastly, I think that 10m may be too short to get consistent results.

test-cases/manager/manager-backup-restore-baseline.yaml
roydahan, line 1: The location of this should be probably under performance and not under manager.

roydahan: Which bring me to the question of, why didn't you just replicate the same test we have for latency_with_operations and replaced the operation with a manager backup nemesis.
If needed, you could just define new manager backup nemesis.

test-cases/manager/manager-backup-restore-baseline.yaml
kreuzerkrieg: > What did you base the "17000/s" on? A typical max throughput test for mixed workload using i4i.4xlarge achieves around 500K op/s. 4 loaders running 17K op/s is equal to 68K op/s which is about 13% of the max throughput. (BTW, to achieve that we use 620 threads and bigger loaders - you can probably start with c7i.2xlarge).

ok, then how do I calculate number of threads and op/s to get to 6-7ms latency?

> lastly, I think that 10m may be too short to get consistent results.
how long it will be considered OK-ish?

roydahan: > ok, then how do I calculate number of threads and op/s to get to 6-7ms latency?
> 
> >

Is your target to measure latency during the backup operation or to measure duration of the operation under a specific load?
If you want to measure latency, you don't know the latency :)
We usually measure the latency under a workload which is 50% of max throughput.
in your case it would be something like "-rate 'threads=620 fixed=62500/s' "

roydahan: > how long it will be considered OK-ish?

How long does the backup operation take?

regevran:
Please note that this test is to create a baseline, something to compare to the new

test-cases/manager/manager-backup-restore-baseline.yaml
kreuzerkrieg: > > how long it will be considered OK-ish?
> 
> How long does the backup operation take?

10min

soyacz:
@regevran see you can apply validation rules that will fail the test if value exceeds error threshold (configurable, see validation rules examples in other tests). In case of setting custom rules for latency_decorator_calculator, [see example ](https://github.com/scylladb/scylla-cluster-tests/commit/bc3552ae58212493d4af3cf3fe01e21671095b59). 

btw. after rebase you can remove `mixed` word from the test name and add `workload_name: mixed` to config.

kreuzerkrieg:
superceeded by https://github.com/scylladb/scylla-cluster-tests/pull/10112

PR #9608, opened 2024-12-23
fix(BaseNode): change directory for startup script upload from /tmp to $HOME.

mikliapko:
Fix changes directory for startup script upload from /tmp to $HOME. 
The change is required for DB nodes deployed in Cloud where /tmp dir is mounted with noexec option what makes script 
execution impossible there.

### Testing

- [x] siren-sct-integration [test](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/siren-tests/job/siren-sct-integration/123)

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2911,7 +2911,8 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
         return cqlsh_out if not split else list(map(str.strip, cqlsh_out.stdout.splitlines()))
 
     def run_startup_script(self):
-        startup_script_remote_path = '/tmp/sct-startup.sh'
+        remote_home_dir = self.remoter.run("echo $HOME").stdout.strip()
+        startup_script_remote_path = f"{remote_home_dir}/startup_script.sh"
 
         with tempfile.NamedTemporaryFile(mode='w+', delete=False, encoding='utf-8') as tmp_file:
             tmp_file.write(self.test_config.get_startup_script())


fruch:
@mikliapko what's the implementation in cloud ? using the `$HOME` ?

mikliapko:
> @mikliapko what's the implementation in cloud ? using the `$HOME` ?
> 
> maybe we can just switch to it ?, there's nothing holy in /tmp for SCT, we just need to run the script.

Yes, the plan is to use $HOME in cloud nodes.
If the directory doesn't matter for SCT as you mentioned above, I will just switch to $HOME in SCT without introducing new attributes.

fruch:
> > @mikliapko what's the implementation in cloud ? using the `$HOME` ?
> > maybe we can just switch to it ?, there's nothing holy in /tmp for SCT, we just need to run the script.
> 
> Yes, the plan is to use $HOME in cloud nodes. If the directory doesn't matter for SCT as you mentioned above, I will just switch to $HOME in SCT without introducing new attributes.

I think it would be better, the defining yet another "API" like that between the two parts

mikliapko:
> I think it would be better, the defining yet another "API" like that between the two parts

Changed to $HOME, testing [here](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/siren-tests/job/siren-sct-integration/123/).

@fruch 
Which SCT may I run to validate changes don't lead to any regression?

fruch:
> > I think it would be better, the defining yet another "API" like that between the two parts
> 
> Changed to $HOME, testing [here](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/siren-tests/job/siren-sct-integration/123/).
> 
> @fruch 
> Which SCT may I run to validate changes don't lead to any regression?

Running CI with AWS provision test, I think that should suffice

vponomaryov (approved):
LGTM

fruch (approved):
LGTM 

The provision test is failing on nemesis, unrelated to this change

Issue #9627, opened 2024-12-30
Make latency_calculator_decorator be usable in longevity tests


soyacz:
`latency_calculator_decorator` requires to be run from test with specific name that includes workload type (write, read, mixed). This way it is not usable within typical longevity tests (or manager tests) that have different test names.

Make this decorator be usable in any scope.
Idea is to add config param that will specify workload type explicitly rather than taking it from the test name.

PR #9628, opened 2024-12-30
improvement(latency_calculator): allow use it in longevity tests

soyacz:
`latency_calculator_decorator` requires workload to be set. Up to now it was taken from the test name. This change allows to set it by `workload_name` test parameter, so it can be used in all kind of tests.

Added latency verification for all 'individual nemesis' cases (proposal, anyway need to test with something).

closes: https://github.com/scylladb/scylla-cluster-tests/issues/9627

### Testing

- [x] - [argus](https://argus.scylladb.com/tests/scylla-cluster-tests/2280ee26-a67f-473a-b3ad-f636f08d3229)

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -268,7 +268,7 @@ zero_token_instance_type_db: 'i4i.large'
 use_zero_nodes: false
 
 latte_schema_parameters: {}
-
+workload_name: ''
 latency_decorator_error_thresholds:
   write:
     default:
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -1738,6 +1738,10 @@ class SCTConfiguration(dict):
              help="Error thresholds for latency decorator."
                   " Defined by dict: {<write, read, mixed>: {<default|nemesis_name>:{<metric_name>: {<rule>: <value>}}}"),
 
+        dict(name="workload_name", env="SCT_WORKLOAD_NAME", type=str,
+             help="Workload name, can be: write|read|mixed|unset."
+                  "Used for e.g. latency_calculator_decorator (use with 'use_hdr_cs_histogram' set to true)."
+                  "If unset, workload is taken from test name."),
     ]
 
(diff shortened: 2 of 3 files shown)


juliayakovlev (commented):
@soyacz @fruch 
Should we calculate latency for prepare stage in longevity?
https://github.com/scylladb/scylla-cluster-tests/blob/f8670793e5a5e0b8dff7252b3d2147a4ecc7e1f7/sdcm/utils/loader_utils.py#L228

Maybe we need go over tests and check where it is important to have

fruch:
> @soyacz @fruch Should we calculate latency for prepare stage in longevity?
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/f8670793e5a5e0b8dff7252b3d2147a4ecc7e1f7/sdcm/utils/loader_utils.py#L228
> 
> Maybe we need go over tests and check where it is important to have

@soyacz was just testing, to see we can collect metric into argus also in longevity, as needed.

we are not gonna collect all metric from all longevity (at least not now)

soyacz:
> > @soyacz @fruch Should we calculate latency for prepare stage in longevity?
> > https://github.com/scylladb/scylla-cluster-tests/blob/f8670793e5a5e0b8dff7252b3d2147a4ecc7e1f7/sdcm/utils/loader_utils.py#L228
> > 
> > Maybe we need go over tests and check where it is important to have
> 
> @soyacz was just testing, to see we can collect metric into argus also in longevity, as needed.
> 
> we are not gonna collect all metric from all longevity (at least not now)

Yes, I tried it only, decision to add to all longevities is separate one.

I think we could consider it individual nemesis cases - we could see the trends and possibly find regressions with it. But requires more effort as latencies are above 10ms causing test to fail, so we would need to make more decisions (whether increase error thresholds or reduce load/throttle it).

juliayakovlev (approved):

fruch (approved):
LGTM

Issue #8352, opened 2024-08-13
hydra clean-resources give error when (only) SCT_SCYLLA_VERSION is set


cezarmoise:
Running
```bash
export SCT_SCYLLA_VERSION=5.2.1
hydra run-test ...
# then later
hydra clean-resources --user `whoami` -b aws
```
Gives this error
```
There is scylladb/hydra:v1.74-tcconfig-0.29.1 in local cache, using it.
Going to run './sct.py  clean-resources --user cezar.moise -b aws'...
/usr/local/lib/python3.10/site-packages/paramiko/pkey.py:100: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "cipher": algorithms.TripleDES,
/usr/local/lib/python3.10/site-packages/paramiko/transport.py:259: CryptographyDeprecationWarning: TripleDES has been moved to cryptography.hazmat.decrepit.ciphers.algorithms.TripleDES and will be removed from this module in 48.0.0.
  "class": algorithms.TripleDES,
logged in as arn:aws:sts::797456418907:assumed-role/DeveloperAccessRole/cezar.moise@scylladb.com
New directory created: /home/cezar/sct-results/20240813-102128-166284-clean-resources
Clean all resources belong to user `cezar.moise'
Failed to load configuration files: []
The run can be interrupted by following critical events:
  * ClusterHealthValidatorEvent.NodeStatus
  * ClusterHealthValidatorEvent.ScyllaCloudClusterServerDiagnostic
  * DataValidatorEvent.UpdatedRowsValidator
  * DatabaseLogEvent.CORRUPTED_SSTABLE
  * CassandraHarryEvent.failure
  * YcsbStressEvent.failure
  * NdBenchStressEvent.failure
  * NdBenchErrorEvent.BuildFailed
  * NdBenchErrorEvent.Failure
  * CDCReaderStressEvent.failure
  * NoSQLBenchStressEvent
  * NoSQLBenchStressLogEvents.ProgressIndicatorStoppedEvent
  * CassandraStressLogEvent.OperationOnKey
  * CqlStressCassandraStressLogEvent.ReadValidationError
  * ScyllaBenchLogEvent.DataValidationError
  * ScyllaBenchLogEvent.ParseDistributionError
  * GeminiStressLogEvent.GeminiEvent
  * PrometheusAlertManagerEvent
[...]

PR #9633, opened 2024-12-31
fix(sct_config): don't fail on empty instance type

fruch:
fix the code not to do API calls to ec2 when instance type isn't defines, and fallback to assume x86 image is requested.

it was failing during cleanup command where we might not have a full configuration available, but we should still carry on and execute the cleanup (we don't really need instance types for cleanups)

Fixes: #8352

### Testing

- [x] tested it locally

### Reminders

### DIFF
--- a/sdcm/utils/aws_utils.py
+++ b/sdcm/utils/aws_utils.py
@@ -441,13 +441,15 @@ def ec2_ami_get_root_device_name(image_id, region_name):
 
 @functools.cache
 def get_arch_from_instance_type(instance_type: str, region_name: str) -> AwsArchType:
-    client: EC2Client = boto3.client('ec2', region_name=region_name)
-    instance_type_info = client.describe_instance_types(InstanceTypes=[instance_type])
     arch = 'x86_64'
-    try:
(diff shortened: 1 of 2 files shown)


sdcm/utils/aws_utils.py
```diff
@@ -441,13 +441,15 @@ def ec2_ami_get_root_device_name(image_id, region_name):
...
 @functools.cache
 def get_arch_from_instance_type(instance_type: str, region_name: str) -> AwsArchType:
-    client: EC2Client = boto3.client('ec2', region_name=region_name)
-    instance_type_info = client.describe_instance_types(InstanceTypes=[instance_type])
     arch = 'x86_64'
-    try:
-        arch = instance_type_info['InstanceTypes'][0].get('ProcessorInfo', {}).get('SupportedArchitectures')[0]
-    except (IndexError, KeyError):
-        pass
+    if instance_type:
```
soyacz, line 445: nit: could just return if no instance_type

soyacz (approved):
I generally dislike the change as `get_arch_from_instance_type` should not be called at all if there's no instance type and it's a pity we need that for cleanup process.

But for sake of issue fix and moving forward LGTM

dimakr (approved):
LGTM

vponomaryov (approved):
LGTM

PR #9638, opened 2025-01-02
fix(sct_config): `scylla-server-upgrade-target` wasn't reported to Argus

fruch:
`ARGUS_VERSION_RE` wasn't covering plain version like `6.3.0~dev` hence not sending it out to Argus.

the change is fixing the regex

this change also fixes `get_branch_version` to be able to return a full version string (including git sha and a date), cause we know Argus would need this data for metrics graphs and perf summaries.

also not having this information, is causing confusion on patch releases, when test are not get to the phase a node is upgrades

Ref: https://github.com/scylladb/argus/issues/533

### Testing

- [x] integration tests
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu20.04-test/29/

### Reminders

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -2459,10 +2459,11 @@ class SCTConfiguration(dict):
 
     def _get_target_upgrade_version(self):
         # 10) update target_upgrade_version automatically
-        new_scylla_repo = self.get('new_scylla_repo')
-        if new_scylla_repo and not self.get('target_upgrade_version'):
-            self['target_upgrade_version'] = get_branch_version(new_scylla_repo)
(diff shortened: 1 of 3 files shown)


soyacz (approved):
LGTM

fruch:
> LGTM

I've found some issues with it, when testing it.
it was picking up scylla-machine-image versions

had to refine the commands a bit more, and tested it also with manger repos

soyacz (approved):
ok, LGTM if works properly :)

fruch:
> ok, LGTM if works properly :)

Now it works better and last upgrade test proven it

PR #9631, opened 2024-12-31
Replace out-of-support CentOS distro with RHEL9 for Manager tests on Azure

mikliapko:
Closes https://github.com/scylladb/scylla-cluster-tests/issues/9630 https://github.com/scylladb/scylla-cluster-tests/issues/8226

The PR switches Manager tests on Azure from using centos-based to rhel-based.

Together with that, it implements a couple of different fixes to make such node setup possible (see list of commits for details).

### Testing

- [x] Azure backup [test](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/sct-feature-test-backup-azure/42/) - not related to PR failure, should be fixed [here](https://github.com/scylladb/scylla-cluster-tests/pull/9590)
- [x] Artifacts CentOS9 [test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/artifacts-centos9-test/2)

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

PR #9649, opened 2025-01-03
fix(manager): update jenkinsfile for Manager Azure tests

mikliapko:
Changes:
- Removed scylla version hardcode for Azure job. The image defined in azure_image_db contains Scylla 5.2.1 is outdated. Moreover, there are no restrictions with passing scylla_version directly. With this change, the Azure job will be executed with Scylla 2024.2 which is default for Manager jobs (if any specific doesn't redefine it).
- Renamed jenkinsfile to address recent move from Centos-based image for Azure jobs to RHEL-based (https://github.com/scylladb/scylla-cluster-tests/pull/9631).

### Testing

- [x] Manager Azure [job](https://jenkins.scylladb.com/job/manager-3.4/job/sct-feature-test-backup-azure/5/) - not related to change failure (fix is [here](https://github.com/scylladb/scylla-cluster-tests/pull/9590))


soyacz (commented):
looks like you forgot rename change

mikliapko:
> looks like you forgot rename change

It's part of the same commit. 
It was not obvious judging by commit header. So, reworded it to make transparent.

fruch (approved):
LGTM

PR #9658, opened 2025-01-06
fix(integration-tests): update versions used

fruch:
some images of older branches aren't available anymore we shouldn't be using that old release in those tests

we should find a better way to stop those test from breaking but for now let just move them to newer versions

### Testing

- [x] integration tests 

### Reminders


vponomaryov (approved):
LGTM

Issue #9578, opened 2024-12-18
Use Argus results for perf_simple_qurey regression validation


soyacz:
We use elasticsearch, last 10 results to verify regressions in perf microbenchmark tests. 
Argus has already enough data and is able to do comparisons based on best results - let's switch to that.

PR #9687, opened 2025-01-07
improvement(perf_simple_query): add validation rules for Argus

soyacz:
When sending perf_simple_query benchmark results, Argus will validate numbers based on all history (submitted to Argus). Current approach validates only 10 last results.

closes: https://github.com/scylladb/scylla-cluster-tests/issues/9578

### Testing

- [x] - [Argus, see graphs for limits too](https://argus.scylladb.com/tests/scylla-cluster-tests/d4bbaf2b-4e3e-4a53-a378-15d79c5036c2)

### Reminders

### DIFF
--- a/microbenchmarking_test.py
+++ b/microbenchmarking_test.py
@@ -30,7 +30,6 @@ class PerfSimpleQueryTest(ClusterTester):
         result = self.db_cluster.nodes[0].remoter.run(
             "scylla perf-simple-query --json-result=perf-simple-query-result.txt --smp 1 -m 1G")
         if result.ok:
-            regression_report = {}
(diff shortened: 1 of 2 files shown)


soyacz:
something is wrong as cpu cycles should be marked as passed, and was marked as failed (the last result - it's below the error threshold).
image

soyacz:
> something is wrong as cpu cycles should be marked as passed, and was marked as failed (the last result - it's below the error threshold).

ok, I see, new result (from run 23) was marked as new best one with much better result (more than 5% threshold) - it' snot visible on graph properly as it's from the same date.

soyacz:
I removed also logallocs _per_op validation as it always shows N/A results (and these fail)

fruch (approved):
LGTM

PR #9697, opened 2025-01-08
feature(auto-backport.py): replace Mergify

yaronkaikov:
Adding an auto-backport.py script to handle backport automation instead of Mergify.

The rules of backport are as follows:
* Merged PR with any backport/x.y label (one or more) and promoted-to-master label
* Backport PR will be automatically assigned to the original PR author
* All PRs will be open under the `scylladbbot` account - which will allow write access in case of conflicts
* PR with conflicts will open in draft mode.
* Fixing cherry-pick the wrong commit SHA. With the new script, we always take the SHA from the stable branch

**Since the SCT repo is public and GitHub doesn't clone permissions to the public repo (for forks),**
**For the first time only, each PR owner will receive an email**
**with an invite to the `scylladbbot/scylla-cluster-tests` fork**

### DIFF
--- a/.github/scripts/auto-backport.py
+++ b/.github/scripts/auto-backport.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python3
+
+import argparse
+import os
+import re
+import sys
+import tempfile
+import logging
+
+from github import Github, GithubException
+from git import Repo, GitCommandError
+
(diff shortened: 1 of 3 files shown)


fruch (approved):
LGTM

PR #9719, opened 2025-01-08
feature(sdcm) wait less for schema agreement

mykaul:
I hope by now it's stable and we can wait less (but retry more to compensate) on the 2 functions that are related to schema agrement. Cut the wait time from 10s to 5s and increased the retry count to 10.

Probably need a similar change in dtest.
BTW, the Python driver sleeps for 0.2s before retrying.

### Testing

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/101/

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2781,7 +2781,7 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
                 cql_results)
         return peers_details
 
-    @retrying(n=5, sleep_time=10, raise_on_exceeded=False)
+    @retrying(n=10, sleep_time=5, raise_on_exceeded=False)
     def get_gossip_info(self) -> dict[BaseNode, dict]:
         gossip_info = self.run_nodetool('gossipinfo', verbose=False, warning_event_on_exception=(Exception,),
                                         publish_event=False)
@@ -4394,7 +4394,7 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
             message=f"There are more then expected nodes running nemesis: {message}",
         ).publish()
 
-    @retrying(n=6, sleep_time=10, allowed_exceptions=(AssertionError,))
+    @retrying(n=10, sleep_time=5, allowed_exceptions=(AssertionError,))
     def wait_for_schema_agreement(self):
 
         for node in self.nodes:
--- a/sdcm/utils/health_checker.py
+++ b/sdcm/utils/health_checker.py
@@ -19,8 +19,8 @@ from sdcm.sct_events import Severity
 from sdcm.sct_events.health import ClusterHealthValidatorEvent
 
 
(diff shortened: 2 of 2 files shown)


mykaul:
Always a pleasure deciphering what failed in CI - I can't find it. https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-9719/ - @fruch , can someone from your team help out here?

fruch:
@vponomaryov you were looking into the healthcheck in the 5000 tables case, you might have more insights about this one

we should also consider which case is fitting to test this change.

mykaul:
@temichus - you may want to review this as well.

vponomaryov (approved):
LGTM

vponomaryov:
> @vponomaryov you were looking into the healthcheck in the 5000 tables case, you might have more insights about this one
> 
> we should also consider which case is fitting to test this change.

I don't expect any drawbacks from this change.
If you want me to run a 5000 tables test, then let me know.

For now it looks ok to me.

fruch:
> > @vponomaryov you were looking into the healthcheck in the 5000 tables case, you might have more insights about this one
> > we should also consider which case is fitting to test this change.
> 
> I don't expect any drawbacks from this change. If you want me to run a 5000 tables test, then let me know.
> 
> For now it looks ok to me.

I don't know if the 5000 tables test, but I would like to see it running in at least on longevity

temichus:
@mykaul, do you want to "wait less" to test stability or increase speed?
If Second, I am not sure that "wait less" will increase speed in SCT where we are working with a big amount of data and 4h+ tests.
More likely the speed will be the same.
More retries and less wait can lead to the opposite effect in SCT, because it may do additional load.

I agree with @fruch and would like to check 5000 tables or longevity test results before and after(speed, num of performed nemeses)

temichus (commented):
@mykaul, do you want to "wait less" to test stability or increase speed?
If Second, I am not sure that "wait less" will increase speed in SCT where we are working with a big amount of data and 4h+ tests.
More likely the speed will be the same.
More retries and less wait can lead to the opposite effect in SCT, because it may do additional load.

I agree with @fruch and would like to check 5000 tables or longevity test results before and after(speed, num of performed nemeses)

mykaul:
> @mykaul, do you want to "wait less" to test stability or increase speed? If Second, I am not sure that "wait less" will increase speed in SCT where we are working with a big amount of data and 4h+ tests. More likely the speed will be the same. More retries and less wait can lead to the opposite effect in SCT, because it may do additional load.
> 
> I agree with @fruch and would like to check 5000 tables or longevity test results before and after(speed, num of performed nemeses)

I want to be closer to real life - where customers do not really wait for long before proceeding on schema agreement. We do it to ensure schema agreement completed, but it should have already.
I understand its' not 0 seconds, but it's also not 30.

Issue #8279, opened 2024-08-07
SCT container should not include kubectl, eksctl, help by default - to cut its size by half


mykaul:
See https://github.com/scylladb/scylla-cluster-tests/blob/b83b3ad911ddf9625cdba8170662ca3d72274ff0/docker/env/Dockerfile#L56 - it downloads 3 large binaries that are hardly being used (only in operator related tests), that somewhat bloat our container.
We are downloading this container gozillion of times (650K in the last 7 days I believe). Those 3 binaries take approx half of the size of the whole container.

Instead, the operator related tests should deploy them when needed, or have their own container based off the main one.

The container is also public and those binaries have known security vulnerabilities (we could and should update them, but it's a separate topic)

fruch:
if we'll split to other containers that we build, they would be public as well.

it's would need change in a lot of of k8s implemetions, and would be big testing effort to keep operator case working after such a change.

soyacz:
@fruch while you're working on new AMI's for sct-runners, maybe would be worth to `docker pull ...` hydra image, so later tests would only download changed layers.
We could from time to time rebuild runner AMI with updated layers (we rarely change k8s stuff/base packages etc. - mostly python dependencies which is the last layer)

mykaul:
> if we'll split to other containers that we build, they would be public as well.
> 
> it's would need change in a lot of of k8s implemetions, and would be big testing effort to keep operator case working after such a change.

If we base the operator hydra on top of the regular one ('hydra-k8s') I doubt it's a big deal. And we can refresh them together, etc.

fruch:
> @fruch while you're working on new AMI's for sct-runners, maybe would be worth to `docker pull ...` hydra image, so later tests would only download changed layers.
> We could from time to time rebuild runner AMI with updated layers (we rarely change k8s stuff/base packages etc. - mostly python dependencies which is the last layer)

We might consider it, it isn't gonna change anything drastically, since we are changing those every couple of k8s releases.

fruch:
@vponomaryov 

do you have estimation on moving all this CLI usages to something like `HelmContainerMixin` ?
so we can yank them out of the hydra image ?

vponomaryov:
> We are downloading this container gozillion of times (650K in the last 7 days I believe). Those 3 binaries take approx half of the size of the whole container.

Removal of these binaries doesn't release `approx half of the size of the whole container.`.
[...]

PR #9720, opened 2025-01-08
feature(Hydra): move K8S components downloads to its own layer

mykaul:
It reduces 1 layer overall, but the hope is that I'll make this layer optional and we'll be able to build 2 different targets:; One with K8S stuff, one without.

### Testing

- [x] NOT tested

### Reminders

### DIFF
--- a/docker/env/Dockerfile
+++ b/docker/env/Dockerfile
@@ -20,10 +20,19 @@ ADD requirements.txt .
 RUN pip install uv==0.2.24
 RUN uv pip install --system -r requirements.txt
 
-FROM python:$PYTHON_IMAGE_TAG
+# Download, extract and install Kubernetes packages.
+FROM apt_base AS k8_packages
 ARG KUBECTL_VERSION=1.27.3
 ARG EKSCTL_VERSION=0.165.0
 ARG HELM_VERSION=3.12.2
+RUN DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends curl tar
+RUN curl -fsSLo /usr/local/bin/kubectl https://dl.k8s.io/release/v$KUBECTL_VERSION/bin/linux/amd64/kubectl
+RUN curl --silent --location "https://github.com/eksctl-io/eksctl/releases/download/v$EKSCTL_VERSION/eksctl_Linux_amd64.tar.gz" | tar xz -C /tmp && \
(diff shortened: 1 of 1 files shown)


fruch:
doing updates to package, would attend to this after them

mykaul:
> Is decision made at (#8279) is not actual anymore?
> 
> Then, we don't have automation for building hydra images. Some of us do it manually. So, having a need to create 2 images each time (often) means that k8s-specific one will not have enough attention and will tend to be obsolete and not match overall-SCT code at some point. The 12-14% size diff doesn't sound to me as a good trade-off for maintaining 2 images which get rebuilt really often.

I thought the change is now bigger percentage-wise, as we got rid of Java. I'm also concerned that I keep saying every run multiple pulls of this hydra images - 99.99% of the cases do not need those utilities.

Yes, that will be applicable when we build the image automatically - which we should do promptly, and not 'by hand'. 

Lastly, it keeps things more tidy, so easier to read (and update the versions).

fruch:
> > Is decision made at (#8279) is not actual anymore?
> > 
> > Then, we don't have automation for building hydra images. Some of us do it manually. So, having a need to create 2 images each time (often) means that k8s-specific one will not have enough attention and will tend to be obsolete and not match overall-SCT code at some point. The 12-14% size diff doesn't sound to me as a good trade-off for maintaining 2 images which get rebuilt really often.
> 
> I thought the change is now bigger percentage-wise, as we got rid of Java. I'm also concerned that I keep saying every run multiple pulls of this hydra images - 99.99% of the cases do not need those utilities.
> 
> Yes, that will be applicable when we build the image automatically - which we should do promptly, and not 'by hand'. 
> 
> Lastly, it keeps things more tidy, so easier to read (and update the versions).
> 

Automation of hydra images complicates the whole CI process of SCT, it's far from being simple.

Have 2 hydra images add other complications as how we select across lots of places.

The agreed solution would be to move some of the tools into their own runtime docker image, that would be pulled and run on demand, as we are doing already for a bunch of things, like java.

mykaul:
> The agreed solution would be to move some of the tools into their own runtime docker image, that would be pulled and run on demand, as we are doing already for a bunch of things, like java.

That's a good solution as well, and this PR makes it easier to go in this path.

fruch:
it's agree this is not the approach to follow with the k8s related tooling, closing for now

PR #9750, opened 2025-01-09
fix(auto-backport) wrong target branch

yaronkaikov:
Auto backport failed while searching for the wrong release branch name (it was looking for `master2024.2` instead of `branch-2024.2`

Fixing it


fruch (approved):
LGTM

PR #9786, opened 2025-01-12
feature(pipelines): collect builders logs

fruch:
we have multiple issue related to builder, and we don't have any logs from their end saved

this commit add code that collect the journalctl logs from builder and upload it to s3 and Argus.

### Testing

- [x] provision test
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ubuntu2404-test/30/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/96/

### Reminders

### DIFF
--- a/vars/collectBuilderLogs.groovy
+++ b/vars/collectBuilderLogs.groovy
@@ -1,0 +1,14 @@
+#!groovy
+
+def call(){
+    sh """#!/bin/bash
+
+	set -xe
+
+	SHORT_SCT_TEST_ID=\$(echo \$SCT_TEST_ID | cut -c1-8)
+    sudo journalctl --no-tail --no-pager -o short-precise > builder-\$SHORT_SCT_TEST_ID.log
+    tar -zcvf builder-\$SHORT_SCT_TEST_ID.log.tar.gz builder-\$SHORT_SCT_TEST_ID.log
+
+    ./docker/env/hydra.sh upload --test-id \$SCT_TEST_ID builder-\$SHORT_SCT_TEST_ID.log.tar.gz
+    """
+}
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -329,7 +329,7 @@ def list_logs_by_test_id(test_id):
     log_types = ['db-cluster', 'monitor-set', 'loader-set', 'event', 'sct', 'jepsen-data', 'siren-manager-set',
                  'prometheus', 'grafana', 'kubernetes', 'job', 'monitoring_data_stack', 'output', 'error',
                  'summary', 'warning', 'critical', 'normal', 'debug', 'left_processes', 'email_data', 'corrupted-sstables',
-                 'sstables']
+                 'sstables', 'builder']
 
     results = []
 
--- a/vars/runCollectLogs.groovy
+++ b/vars/runCollectLogs.groovy
@@ -34,4 +34,6 @@ def call(Map params, String region){
     fi
     echo "end collect logs"
     """
+
+    collectBuilderLogs()
 }


fruch:
AWS provision failed cause of access to S3 (unrelated to the PR)

```
23:22:43  Command: 'sudo sctool backup -c a05d9ea2-312c-4c8d-99b0-9f9b57e8cbde --keyspace scylla_bench,keyspace1  --location s3:manager-backup-tests-us-east-1 '
23:22:43  Exit code: 1
23:22:43  Stdout:
23:22:43  Stderr:
23:22:43  Error: create backup target: location is not accessible
23:22:43  10.4.2.70: agent [HTTP 400] operation put: s3 upload: 301 Moved Permanently: The bucket you are attempting to access must be addressed using the specified endpoint. Please send all future requests to this endpoint. (code:PermanentRedirect) - make sure the location is correct and credentials are set, to debug SSH to 10.4.2.70 and run "scylla-manager-agent check-location -L s3:manager-backup-tests-us-east-1 --debug"
23:22:43  Trace ID: DXvyKGPjQMKUCkJpyxlKCQ (grep in scylla-manager logs)
```

@mikliapko this is quite wired, how did we end doing backup to us-east-1, when the test is running in eu-west-1 ?

dimakr:
@fruch All builder logs are collected and these can be from some old days. For instance in the test run https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ubuntu2404-test/30/, there are logs for the run date itself - 12.01.2025, but also logs from 18.12.2024. Do we want to collect all logs from builder or maybe filter them ('since' option or similar)?

vars/collectBuilderLogs.groovy
```diff
@@ -0,0 +1,14 @@
+#!groovy
+
+def call(){
+    sh """#!/bin/bash
+
+	set -xe
+
+	SHORT_SCT_TEST_ID=\$(echo \$SCT_TEST_ID | cut -c1-8)
+    sudo journalctl --no-tail --no-pager -o short-precise > builder-\$SHORT_SCT_TEST_ID.log
```
soyacz, line 9: maybe adding  --since "x hour ago"

mikliapko:
> @mikliapko this is quite wired, how did we end doing backup to us-east-1, when the test is running in eu-west-1 ?

oh, I believe it's a result of getting rid of `backup_bucket_region` I haven't foreseen.
Before, the test was passing as backup_bucket_region parameter was set to us-east-1 (means that manager agent had this region in configuration). Now, manager-agent configuration file has the `region_name` there (`eu-west-1`) which leads to regions mismatch and eventually this failure.

For the sake of verifying this PR, I’d suggest rerunning the test in the us-east-1 region, if possible.

As a long term solution, I suppose we should have buckets in all regions used in testing (pre-created in advance or created in a test runtime if missing). So, I will try to handle it shortly.

fruch:
> @fruch All builder logs are collected and these can be from some old days. For instance in the test run https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/artifacts-ubuntu2404-test/30/, there are logs for the run date itself - 12.01.2025, but also logs from 18.12.2024. Do we want to collect all logs from builder or maybe filter them ('since' option or similar)?

some of them are from when the AMI was created, which is o.k.

vars/collectBuilderLogs.groovy
fruch: why not all ?

soyacz: if builder will be running for couple days, it will be big and redundant

fruch: it would be very small compared to SCT and DB logs.

we don't yet know exactly what we need of it, I would rather first collect it all.

soyacz (approved):
LGTM

dimakr (approved):
LGTM

vars/collectBuilderLogs.groovy
fruch: FY, test run for 2h, the uncompressed log was 485K

Issue #9156, opened 2024-11-07
Add testcase for scaling-out while having 90% storage utilization


pehala:

Lakshmipathi:
3-node cluster scaleout at 90%.

reached 91% disk usage and started waiting for 30mins, no write or read.

```
< t:2024-11-03 07:10:58,323 f:full_storage_utilization_test.py l:93   c:FullStorageUtilizationTest p:INFO  > Current max disk usage after writing to keyspace10: 91% (396 GB / 392.40000000000003 GB)
< t:2024-11-03 07:10:59,353 f:full_storage_utilization_test.py l:58   c:FullStorageUtilizationTest p:INFO  > Wait for 1800 seconds
```

After 30min idle time, started throttled write:

```
< t:2024-11-03 07:42:10,941 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > stress_cmd=cassandra-stress write duration=30m -rate threads=10 "throttle=1400/s" -mode cql3 native -pop seq=1..5000000 -col "size=FIXED(10240) n=FIXED(1)" -schema "replication(strategy=NetworkTopologyStrategy,replication_factor=3)"
```

Scaleout by adding a new node at 90% 
```
< t:2024-11-03 07:44:05,075 f:full_storage_utilization_test.py l:41   c:FullStorageUtilizationTest p:INFO  > Adding a new node
```

After 30mins, scaleout (3->4) cluster has disk usage at 75%, 74%, 75% and 70%
![Image](https://github.com/user-attachments/assets/812d348b-6676-4b3c-8211-9f60b050e250)


https://argus.scylladb.com/tests/scylla-cluster-tests/c5de2f39-770c-4cf3-8d8c-66fef9d91d87

swasik:
> After 30mins, scaleout (3->4) cluster has disk usage at 75%, 74%, 75% and 70%

But I see that the chart presents average disk usage. This should change quickly as we are adding more disk space even if the new space is not used. Could you also add picture for maximal disk usage across all nodes?

swasik:
The interesting fact is that after migration we have the same number of tablets everywhere but on the new node the disk utilization is ca. 5% lower. Maybe something is not cleaned yet. Could we wait a bit more time to see if the utilization will be equal in the end?

Lakshmipathi:
Started new job, with 1hr wait time just before the test ends. Will check and update whether 5% lower disk usage still exists or not.

[...]

PR #9787, opened 2025-01-13
test(performance): Add 90% storage utilization scaleout/scalein test

Lakshmipathi:
Perform scale-out around 90% disk utilization and cluster scale-in by decommissioning the newly added nodes while measuring latency under mixed-load stress.

Testcase: https://github.com/scylladb/scylla-cluster-tests/issues/9156

### Testing

- [ ]  [waiting for completion](https://jenkins.scylladb.com/job/scylla-staging/job/LakshmipathiGanapathi/job/scylla-enterprise-perf-regression-latency-650gb-elasticity/95/)

### Reminders


pehala:
Please fix all checks

sdcm/nemesis.py
```diff
@@ -5625,6 +5687,13 @@ class GrowShrinkClusterNemesis(Nemesis):
     def disrupt(self):
         self.disrupt_grow_shrink_cluster()
 
+class FullStorageUtilizationNemesis(Nemesis):
```
pehala, line 5690: I would appreciate more documentation for this class and all methods you added

sdcm/nemesis.py
```diff
@@ -225,6 +225,7 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     manager_operation: bool = False  # flag that signals that the nemesis uses scylla manager
     delete_rows: bool = False  # A flag denotes a nemesis deletes partitions/rows, generating tombstones.
     zero_node_changes: bool = False
+    full_storage_utilization: bool = False
```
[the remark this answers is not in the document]
Lakshmipathi: I added a new label to so that 90% SCT job can be grouped under it. Initially kept it as 'disrupt' but existing SCT runs may use this when `SisyphusMonkey` invoked or any job that uses `disrupt` label - which will result in job's failure. Since this job expect specific instance type and disk usage state to begin with, added it under new label.

sdcm/nemesis.py
```diff
@@ -2104,11 +2105,11 @@ def _truncate_cmd_timeout_suffix(self, truncate_timeout):  # pylint: disable=no-
         # NOTE: 'self' is used by the 'scylla_versions' decorator
         return ''
 
-    def disrupt_truncate(self):
-        keyspace_truncate = 'ks_truncate'
+    def disrupt_truncate(self, full_storage_utilization=False):
+        keyspace_truncate = 'ks_truncate' if not full_storage_utilization else 'refill_keyspace'
```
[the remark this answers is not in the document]
Lakshmipathi: ok, update the PR.

sdcm/nemesis.py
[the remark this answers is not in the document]
pehala: I really dont like the solution, we effectively want to exclude Nemesis from general use and to do that we create another group, which will always contain one nemesis. Isnt there a better way how to do this?

performance_regression_test.py
```diff
@@ -218,6 +218,12 @@ def _stop_load_when_nemesis_threads_end(self):
                                          extra_time_to_expiration=60):
             self.loaders.kill_stress_thread()
 
+    def stop_load_during_nemesis(self):
+        with EventsSeverityChangerFilter(new_severity=Severity.NORMAL,  # killing stress creates Critical error
```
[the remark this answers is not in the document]
pehala: I would downgrade it to DEBUG, I dont see any value in seeing this ever, as it is expected

pehala:
Please use common folder structure as seen in https://github.com/scylladb/scylla-cluster-tests/pull/9796

pehala:
Changing to draft, as this PR needs to be adjusted to use 3 racks

mykaul:
If it's not too late for this, or perhaps it could be done in a follow-up, I've already discussed with @pehala a more interesting and efficient way to test this:
1. Scale out - say, to 6, or 9 nodes.
2. Fill the cluster to 40+% (in case of 6 nodes) or 25+% (in case of 9 nodes)
3. Scale *in* - you will now have >80% (in case of 6 nodes) or >75% (in case of 9 nodes) filled in already.
4. Now go all the way to 90, and then you can continue the test - additional scale-out, or remove some data and scale-in, etc.

The idea is that you'll have more 'horsepower' for ingest to fill the cluster faster.

pehala:
> If it's not too late for this, or perhaps it could be done in a follow-up, I've already discussed with @pehala a more interesting and efficient way to test this:
> 
>     1. Scale out - say, to 6, or 9 nodes.
> 
>     2. Fill the cluster to 40+% (in case of 6 nodes) or 25+% (in case of 9 nodes)
> 
>     3. Scale _in_ - you will now have >80% (in case of 6 nodes) or >75% (in case of 9 nodes) filled in already.
> 
>     4. Now go all the way to 90, and then you can continue the test - additional scale-out, or remove some data and scale-in, etc.
> 
> 
> The idea is that you'll have more 'horsepower' for ingest to fill the cluster faster.

Thanks for the suggestion, it is being tested in a separate effort

pehala:
@Lakshmipathi What is missing for this PR to be ready for merging?

Lakshmipathi:
> @Lakshmipathi What is missing for this PR to be ready for merging?

@pehala , I updated PR yesterday, I think its ready for final review and merge.

pehala:
> @pehala , I updated PR yesterday, I think its ready for final review and merge.

I see merge conflict and the actual PR is in draft

Lakshmipathi:
> > @pehala , I updated PR yesterday, I think its ready for final review and merge.
> 
> I see merge conflict and the actual PR is in draft

rebased the PR now.

sdcm/nemesis.py
```diff
@@ -2124,11 +2124,11 @@ def _truncate_cmd_timeout_suffix(self, truncate_timeout):  # pylint: disable=no-
         # NOTE: 'self' is used by the 'scylla_versions' decorator
         return ''
 
-    def disrupt_truncate(self):
-        keyspace_truncate = 'ks_truncate'
+    def disrupt_truncate(self, refill_keyspace=False):
```
cezarmoise, line 2127: Why not name the keyspace in the config `ks_truncate` to avoid having to modify this?

sdcm/nemesis.py
```diff
@@ -4411,6 +4411,86 @@ def _shrink_cluster(self, rack=None, new_nodes: list[BaseNode] | None = None):
...
+        Performs cluster scale out when storage utilization is around 90%.
+
+        It adds first new node, then refills data to 90% utilization level after stopping current load.
+        Start mixed workload and add remaining new node.
+        """
+        add_nodes_number = self.tester.params.get('nemesis_add_node_cnt')
+        InfoEvent(message=f"Start grow cluster by {add_nodes_number} data nodes").publish()
+        new_nodes = []
+        stress_queue = None
+        for idx in range(add_nodes_number):
```
cezarmoise, line 4439: You can use `self._grow_cluster()` instead, it will also distribute nodes to racks, and it works with `parallel_node_opperations=true`

sdcm/nemesis.py
```diff
@@ -4411,6 +4411,86 @@ def _shrink_cluster(self, rack=None, new_nodes: list[BaseNode] | None = None):
...
+                # wait for c-s to start
+                time.sleep(120)
+            else:
+                new_nodes += self.add_new_nodes(count=1, rack=rack_idx,
+                                               instance_type=self.tester.params.get('nemesis_grow_shrink_instance_type'))
+        self.log.info("Finish cluster grow")
+        time.sleep(self.interval)
+        return new_nodes
+
+    def scalein_to_reach_full_storage(self, rack=None, new_nodes: list[BaseNode] | None = None):
```
cezarmoise, line 4469: You can use `self._shrink_cluster(rack=None, new_nodes=new_nodes)` instead of a new function.

sdcm/nemesis.py
```diff
@@ -4411,6 +4411,86 @@ def _shrink_cluster(self, rack=None, new_nodes: list[BaseNode] | None = None):
...
+        new_nodes = []
+        stress_queue = None
+        for idx in range(add_nodes_number):
+            # if rack is not specified, round-robin racks to spread nodes evenly
+            rack_idx = rack if rack is not None else idx % self.cluster.racks_count
+            if idx == 2:
+                new_nodes += self.add_new_nodes(count=1, rack=rack_idx,
+                                                instance_type=self.tester.params.get('nemesis_grow_shrink_instance_type'))
+                # Before starting refilling data current c-s need to be stopped
+                self.tester.stop_load_during_nemesis()
```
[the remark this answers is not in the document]
Lakshmipathi: With my runs, keeping original mixed-workload running and starting a new write workload to re-fill data back to 90% and then adding a node lead to latency issue due to higher CPU load.  That's the reason for stopping it before starting another c-s to re-fill data.

sdcm/nemesis.py
Lakshmipathi: With the simulated_rack:3 , seems like I can just re-use existing GrowShrinkNemesis (because when new node added data is distributed among that rack nodes only while other rack node are still at 90%. This means we dont need run c-s too to re-fill). Will update the PR.

cezarmoise:
With latest version, does the test no longer fill the cluster between the scale-out and scale-in?

Lakshmipathi:
> With latest version, does the test no longer fill the cluster between the scale-out and scale-in?

yes, refill was used to add nodes at 90% earlier. The flow used to be: 3-node cluster -> Added node-4 -> Before adding node5, since overall usage is reduced, we did refill to reach 90% again -> Added node-5 ->  decommission node4 to reach 90% again -> Drop/truncate refill keyspace -> Reaches 70% mark -> Decommission node-5 to reach 90% 

With simulated_rack:3 option, the flow now: 3-node cluster with rack1,rack2,rack3 -> added node-4 to rack1 -> rack1 data usage reduced but rack2 and rack3 disk usage still at 90% -> added node-5 to rack2 -> rack1 and rack2 data usage reduced but rack3 still at 90% -> added node-6 to rack3 -> Now overall disk usage around 50% -> If we refill data to reach 90%, then decommission will hit no space error. 

We can refill to reach 90% after scaleout but before decommission we need to drop this same data. It will be like scale-out -> refill data -> truncate refilled data -> decommission nodes.

test-cases/features/elasticity/perf-regression-latency-i4i_2xlarge-elasticity-90-percent.yaml
```diff
@@ -0,0 +1,43 @@
+test_duration: 3000
```
pehala, line 1: The name of this testcase does not reflect the title or purpose of it

pehala (changes requested):

pehala (approved):

pehala:
@scylladb/qa-maintainers Please review

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
```diff
@@ -0,0 +1,42 @@
+test_duration: 3000
+prepare_write_cmd: [
+  "cassandra-stress write no-warmup cl=ALL n=300000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=1..300000000",
+  "cassandra-stress write no-warmup cl=ALL n=300000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=300000001..600000000",
+  "cassandra-stress write no-warmup cl=ALL n=300000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=600000001..900000000",
+  "cassandra-stress write no-warmup cl=ALL n=250000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=900000001..1200000000",
```
soyacz, line 6: why 250000000?

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
```diff
@@ -0,0 +1,42 @@
...
+  "cassandra-stress write no-warmup cl=ALL n=300000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=1200000001..1500000000",
+]
+
+stress_cmd_w: "cassandra-stress write no-warmup cl=ALL n=335544320 -schema 'keyspace=refill_keyspace replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=300 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=1500000001..1835544321"
+stress_cmd_m: "cassandra-stress mixed no-warmup cl=QUORUM duration=800m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate 'threads=300 fixed=16875/s' -col 'size=FIXED(128) n=FIXED(8)' -pop 'dist=gauss(1..600000000,300000000,6000000)'"
+
+simulated_racks: 3
+n_db_nodes: 3
+n_loaders: 4
```
soyacz, line 15: 4 loaders, but there are 5 prepare_write_cmd - for better perf, use 4 commands for prepare load

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
```diff
@@ -0,0 +1,42 @@
...
+n_monitor_nodes: 1
+
+instance_type_loader: 'c6i.2xlarge'
```
soyacz, line 18: c7i should be better.
@Lakshmipathi did you verify how much cpu is utilized on loaders? maybe we could reduce the size?

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
```diff
@@ -0,0 +1,42 @@
...
+  "cassandra-stress write no-warmup cl=ALL n=300000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=600000001..900000000",
+  "cassandra-stress write no-warmup cl=ALL n=250000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=900000001..1200000000",
+  "cassandra-stress write no-warmup cl=ALL n=300000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=200 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=1200000001..1500000000",
+]
+
+stress_cmd_w: "cassandra-stress write no-warmup cl=ALL n=335544320 -schema 'keyspace=refill_keyspace replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=300 -col 'size=FIXED(128) n=FIXED(8)' -pop seq=1500000001..1835544321"
```
soyacz, line 10: is it intended to do cl=ALL? usually we base on QUORUM as this is what most users do.
btw. what is the purpose of this command? - it is not bound by duration

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
```diff
@@ -0,0 +1,42 @@
...
+round_robin: true
+append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
+backtrace_decoding: false
+print_kernel_callstack: true
+
+store_perf_results: true
+use_hdrhistogram: true
+email_recipients: ["scylla-perf-results@scylladb.com"]
+email_subject_postfix: 'fullstorage utilization test'
+parallel_node_operations: false
```
soyacz, line 37: are you sure this should be false? What is the purpose of this test?

vponomaryov: +1

Explicit enabling of the tablets below assume that `parallel_node_operations` should be supported.

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
vponomaryov: > is it intended to do cl=ALL? usually we base on QUORUM as this is what most users do. 

+1
Since we have RF=3, the disk utilization must be about the same, but CPU utilization and latency will be smaller if we use CL=Quorum.

> btw. what is the purpose of this command? - it is not bound by duration

It uses different keyspace - `refill_keyspace`, so, it adds more data/"disk utilization".

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
```diff
@@ -0,0 +1,42 @@
+test_duration: 3000
```
vponomaryov, line 1: 50 hours?

Looks too big.
Jenkins test runs may hang, so, better to reduce it

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
vponomaryov: Loader nodes utilization:
Screenshot from 2025-05-09 17-15-30

Source: [scylla-staging/LakshmipathiGanapathi/scylla-enterprise-perf-regression-latency-650gb-elasticity#98](https://argus.scylladb.com/tests/scylla-cluster-tests/8a90107a-8c81-4332-a264-767f47351a26)

RAM-wise loader nodes have big margin, but CPU-wise it exceeds 50% utilization even without 5->4 disbalance happening in the pre-load step.

So, not changing summary load values, it may stay as-is, I think.

vponomaryov (changes requested):

test-cases/features/elasticity/scalein-scaleout-elasticity-90-percent.yaml
soyacz: ok, in that case we cannot reduce this size. Unless this strange load we pose to Scylla is bogus... - @Lakshmipathi what is the reason of this higher load?

Lakshmipathi:
@pehala  Will update/fix this PR, next week.

fruch:
@pehala @Lakshmipathi is this one still needed?

Issue #9789, opened 2025-01-13
AttributeError: 'list' object has no attribute 'split'


timtimb0t:
## Packages

Scylla version: `6.3.0~dev-20250108.e51b2075dacc` with build-id `1ffc83e51d7f78126ce77667ff1140f5f4913518`

Kernel Version: `6.8.0-1020-azure`

## Issue description

During the disrupt_mgmt_restore nemesis execution the following error occurred:

```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5486, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3074, in disrupt_mgmt_restore
    is_multi_dc = len(self.cluster.params.get('region_name').split()) > 1
AttributeError: 'list' object has no attribute 'split'
```

## Impact

Only sct framework failure

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 4 nodes (Standard_L16s_v3)

Scylla Nodes used in this run:
- longevity-tls-1tb-7d-master-db-node-6cafd6b8-eastus-7 (null | 10.0.0.7) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-6cafd6b8-eastus-6 (null | 10.0.0.7) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-6cafd6b8-eastus-5 (null | 10.0.0.14) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-6cafd6b8-eastus-4 (null | 10.0.0.8) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-6cafd6b8-eastus-3 (null | 10.0.0.7) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-6cafd6b8-eastus-2 (null | 10.0.0.6) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-6cafd6b8-eastus-1 (null | 10.0.0.5) (shards: 14)

OS / Image: `/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/SCYLLA-IMAGES/providers/Microsoft.Compute/images/scylla-6.3.0-dev-x86_64-2025-01-09T11-18-00` (NO RUNNER: NO RUNNER)

[...]

PR #9790, opened 2025-01-13
fix(disrupt_mgmt_restore): adjust getting regions

temichus:
adjust getting region names
due to we have different types of values in 'region_name' for AWS(string) and AZURE(list)

### Testing

- [ ] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/artsiom_mishuta/job/longevity-1tb-5days-azure-test/2/

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3071,7 +3071,12 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             "https://github.com/scylladb/scylla-manager/issues/3829",
             "https://github.com/scylladb/scylla-manager/issues/4049"
         ]
-        is_multi_dc = len(self.cluster.params.get('region_name').split()) > 1
+
+        regions = self.cluster.params.get('region_name')
+        if isinstance(regions, str):
+            regions = regions.split()
+
+        is_multi_dc = len(regions) > 1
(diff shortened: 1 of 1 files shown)


soyacz (commented):
they all should behave the same way - Azure does not support multi-az and could not be taken care of properly.
For Azure (or GCE) we should verify `azure_region_name` (`gce_datacenter`)

temichus:
> they all should behave the same way - Azure does not support multi-az and could not be taken care of properly. For Azure (or GCE) we should verify `azure_region_name` (`gce_datacenter`)

Azure does not support multi-az  - giot


but in code  I see region_name' is used for azure 
https://github.com/scylladb/scylla-cluster-tests/issues/9789#issuecomment-2587076117

temichus:
as i see this nemesis alo not supporting milti DC

```
 if SkipPerIssues(skip_issues, params=self.tester.params) and is_multi_dc:
            raise UnsupportedNemesis("MultiDC cluster configuration is not supported by this nemesis")
```

so @soyacz could you please suggesta  fix?

can I use "self.cluster.params.region_names" for this? 

or it should be "If" statements to get region_name/azure_region_name/gce_datacenter depending on backend ?

fruch:
I think https://github.com/scylladb/scylla-cluster-tests/pull/9791, took care of this already

PR #9821, opened 2025-01-14
fix(common): change wait_for_tablets_balanced to represent what it actualy does

dkropachev:
This function checks if there are no ongoing tablet topology operations. It does not guarantee that tablets become balanced, this language confuses when you read code and events log.

### Testing
No code changes, testing is not needed.

### Reminders

### DIFF
--- a/sdcm/utils/tablets/common.py
+++ b/sdcm/utils/tablets/common.py
@@ -23,9 +23,12 @@ class TabletsConfiguration:
         return '{' + ', '.join(items) + '}'
 
 
-def wait_for_tablets_balanced(node):
+def wait_no_tablets_migration_running(node):
     """
-    Waiting for tablets to be balanced using REST API.
+    Waiting for having no ongoing tablets topology operations using REST API.
+    !!! It does not guarantee that tablets are balanced !!!
+    Keep in mind that it is good only to find when ongoing tablets topology operations is done.
+    Very next second another topology operation can be started.
 
     doing it several times as there's a risk of:
     "currently a small time window after adding nodes and before load balancing starts during which
@@ -37,9 +40,9 @@ def wait_for_tablets_balanced(node):
             return
     time.sleep(60)  # one minute gap before checking, just to give some time to the state machine
     client = RemoteCurlClient(host="127.0.0.1:10000", endpoint="", node=node)
-    LOGGER.info("Waiting for tablets to be balanced")
+    LOGGER.info("Waiting for having no ongoing tablets topology operations")
     for _ in range(3):
         client.run_remoter_curl(method="POST", path="storage_service/quiesce_topology",
                                 params={}, timeout=3600, retry=3)
         time.sleep(5)
-    LOGGER.info("Tablets are balanced")
+    LOGGER.info("All ongoing tablets topology operations are done")
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -131,7 +131,7 @@ from sdcm.utils.replication_strategy_utils import temporary_replication_strategy
     NetworkTopologyReplicationStrategy, ReplicationStrategy, SimpleReplicationStrategy
 from sdcm.utils.sstable.load_utils import SstableLoadUtils
 from sdcm.utils.sstable.sstable_utils import SstableUtils
-from sdcm.utils.tablets.common import wait_for_tablets_balanced
+from sdcm.utils.tablets.common import wait_no_tablets_migration_running
 from sdcm.utils.toppartition_util import NewApiTopPartitionCmd, OldApiTopPartitionCmd
 from sdcm.utils.version_utils import (
     MethodVersionNotFound, scylla_versions, ComparableScyllaVersion, get_systemd_version)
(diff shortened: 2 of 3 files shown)


fruch:
@dkropachev 

what did you assume from reading it ? it's all a matter of POV, from tablets they are balanced when this operation is done.

it doesn't mean that the client communication is balanced

I don't see a big difference in the wording

soyacz:
Reason it was named like this is that we wait for tablet balancer to complete, maybe switching to `wait_for_tablet_balancer_complete` this name would be better. 
also I'm not sure if this reflects what `quiscence_topology` means - tablet topology operation can be also merge/split I believe so your function also may not be precise. (btw. you have typo in the function name)

dkropachev:
> @dkropachev
> 
> what did you assume from reading it ? it's all a matter of POV, from tablets they are balanced when this operation is done.
> 
> it doesn't mean that the client communication is balanced
> 
> I don't see a big difference in the wording

Here is the difference, today I have looked at the case when load wasn't spread evenly among nodes.
It was https://jenkins.scylladb.com/job/enterprise-2024.2/job/longevity_tablets/job/longevity-gce-custom-d1-worklod1-hybrid-raid-test/6/
testid: 6b604e24-f282-457a-b546-4fdda76d96da

I pulled evens and I saw this event:
```
2025-01-10 15:06:57.053: (DisruptionEvent Severity.ERROR) period_type=end event_id=0242322b-bae3-4e17-8f8e-d6b99794805c duration=34m22s: nemesis_name=GrowShrinkCluster target_node=Node long-custom-d1-wrkld1-2024-2---db-node-6b604e24-0-3 [35.229.31.108 | 10.142.0.22] errors=Killed by tearDown - test fail
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5294, in wrapper
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4165, in disrupt_grow_shrink_cluster
    self._grow_cluster(rack=None)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4192, in _grow_cluster
    self.add_new_nodes(count=1, rack=rack_idx)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 197, in wrapped
    res = func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4103, in add_new_nodes
    wait_for_tablets_balanced(nodes[0])
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/tablets/common.py", line 38, in wait_for_tablets_balanced
    time.sleep(60)  # one minute gap before checking, just to give some time to the state machine
sdcm.exceptions.KillNemesis
```

Oh, got dam, I was thinking, that is why it was unbalanced, it is because tablets where not balanced.
And indeed tablets where unbalanced, BUT, this function does not check they are balanced or not, it just waits migration to complete or balancer to stop working, it doesn't validate it's results.
So, I was right by being wrong.
Next it got me thinking of why this nemesis failed only this time, not other times, it should fail all the time. 
That how I endup looking at sct code, looking at the scylla code to understand what is going on.
And that is only because naming was wrong, confused me and made me spend more time than I should on getting to the bottom of it.

If it would be `wait_for_tablet_balancer_complete` as @soyacz proposed, I would void doing that and did spend time on something else.

dkropachev:
> Reason it was named like this is that we wait for tablet balancer to complete, maybe switching to `wait_for_tablet_balancer_complete` this name would be better. also I'm not sure if this reflects what `quiscence_topology` means - tablet topology operation can be also merge/split I believe so your function also may not be precise. (btw. you have typo in the function name)

Can you please take a look again, is it good enough ?

yarongilor (commented):
perhaps `wait_for_tablet_balancer_complete` is not fully accurate since it implies `balancer` is an operation, but it is rather a "service" that runs all the time.
There is a notion like in `wait_no_compactions_running` that may be considered here as well so it can be rephrased like: `wait_no_tablets_migration_running` or similar.

dkropachev:
According to `/storage_service/quiesce_topology` description, I have picked best name possible `wait_no_ongoing_topology_operations`
```
Waits until there are no ongoing topology operations. Guarantees that topology operations which started before the call are finished after the call
```

sdcm/utils/tablets/common.py
```diff
@@ -23,9 +23,10 @@ def __str__(self):
         return '{' + ', '.join(items) + '}'
 
 
-def wait_for_tablets_balanced(node):
+def wait_no_ongoing_topology_operations(node):
```
soyacz, line 26: need to adjust here too

sdcm/utils/tablets/common.py
```diff
@@ -23,9 +23,10 @@ def __str__(self):
...
     """
-    Waiting for tablets to be balanced using REST API.
+    Waiting for having no ongoing tablets topology operations using REST API.
```
soyacz, line 28: worth to add this note you mentioned, that waits for current migration to complete (still another one may start in a second)

sdcm/utils/tablets/common.py
```diff
@@ -37,9 +38,9 @@ def wait_for_tablets_balanced(node):
             return
     time.sleep(60)  # one minute gap before checking, just to give some time to the state machine
     client = RemoteCurlClient(host="127.0.0.1:10000", endpoint="", node=node)
-    LOGGER.info("Waiting for tablets to be balanced")
+    LOGGER.info("Waiting for tablets topology operations to complete")
```
soyacz, line 41: these log messages could also reflect new func name

soyacz (commented):
Generally I'm ok with this change now - looks less confusing

sdcm/utils/tablets/common.py
dkropachev: done

sdcm/utils/tablets/common.py
dkropachev: done

sdcm/utils/tablets/common.py
dkropachev: done

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #9822, opened 2025-01-14
feature(jenkins-pipelines): merge all enterprise features into oss

fruch:
- first we merge all of the job into one directory

- since no enterprise branch anymore, and more enterprise releases 
  we don't need this specific command anymore

### Testing

- [ ]

```[tasklist]
### TODOs
- [ ] https://github.com/scylladb/scylla-pkg/pull/4803
``` 

### Reminders


roydahan (approved):

roydahan:
Regarding the ICS tests, we probably need to go over these, I guess that ICS is going to be the default CS, so it means the regular jobs that don't define CS will use ICS.
So ICS folder is probably redundant.

Anyway, I can do it after this move.

fruch:
> Regarding the ICS tests, we probably need to go over these, I guess that ICS is going to be the default CS, so it means the regular jobs that don't define CS will use ICS.
> So ICS folder is probably redundant.
> 
> Anyway, I can do it after this move.
> 

If there's some order you want to do, we have until branching starts

BTW I've kept the `oss` to minimize the disturbance to our staging jobs

fruch:
merging this, scylla-pkg has one trigger affected, PR is raised to update it

PR #9837, opened 2025-01-15
Improve Manager snapshot preparer helper

mikliapko:
refs: 
https://github.com/scylladb/scylla-manager/issues/4190
https://github.com/scylladb/scylla-manager/issues/4208

The PR addresses two main topics:

1. Manager snapshot preparer helper extension
- Dataset parameters (which defined in c-s cmd) aren't hardcoded anymore what gives an opportunity to widely vary its properties (replication, compaction, cl, rf, etc). 
manager_restore_benchmark_snapshots.yaml file with snapshots config for current benchmark tests is extended to include such a parameters. These params are later used on C-S read command generation stage. 
Thus, snapshot preparation test becomes widely customizable with more flexibility to adjust dataset.
- Generated backup snapshot details are sent to Argus Results instead of logging what improves usability.

2. Cloud clusters support
- Extension for test_prepare_backup_snapshot that allows to run the test either for standard cluster created via SCT or for clusters created in via siren-tests (Cloud cluster).

### Testing

- [x] [Cloud cluster snapshot](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/siren-tests/job/siren-sct-integration/160/), [Argus](https://argus.scylladb.com/tests/scylla-cluster-tests/f162625b-815f-4644-be20-6fab77496ce8) (see Results tab)
- [x] [Regular SCT cluster snapshot creation](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/helper-prepare-backup-snapshot/13/), [Argus](https://argus.scylladb.com/tests/scylla-cluster-tests/52b0856f-4f06-426a-99bb-f5a2bb89a498) (see Results tab)
- [x] [Restore from snapshot generated with a new approach](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master/job/restore-benchmark/18/)

### DIFF
--- a/defaults/manager_restore_benchmark_snapshots.yaml
+++ b/defaults/manager_restore_benchmark_snapshots.yaml
@@ -1,110 +1,164 @@
 bucket: "manager-backup-tests-permanent-snapshots-us-east-1"
-cs_read_cmd_template: "cassandra-stress read cl=ONE n={num_of_rows} -schema 'keyspace={keyspace_name} replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq={sequence_start}..{sequence_end}"
+cs_read_cmd_template: "cassandra-stress read cl={cl} n={num_of_rows} -schema 'keyspace={keyspace_name} replication(strategy={replication},replication_factor={rf}) compaction(strategy={compaction})' -mode cql3 native -rate threads=500 -col 'size=FIXED({col_size}) n=FIXED({col_n})' -pop seq={sequence_start}..{sequence_end}"
 sizes:  # size of backed up dataset in GB
   1gb_1t_ics:
     tag: "sm_20240812100424UTC"
-    schema:
-      keyspace1:
-        - standard1: 1
-    number_of_rows: 1073760
     exp_timeout: 1200  # 20 minutes (timeout for restore data operation)
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "IncrementalCompactionStrategy"
     prohibit_verification_read: false
+    dataset:
+      schema:
+        keyspace1:
+          - standard1: 1
+      num_of_rows: 1073760
+      compaction: "IncrementalCompactionStrategy"
+      cl: "ONE"
+      col_size: 1024
+      col_n: 1
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   500gb_1t_ics:
     tag: "sm_20240813112034UTC"
-    schema:
-      keyspace1:
-        - standard1: 500
-    number_of_rows: 524288000
     exp_timeout: 14400  # 4 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "IncrementalCompactionStrategy"
     prohibit_verification_read: false
+    dataset:
+      schema:
+        keyspace1:
+          - standard1: 500
+      num_of_rows: 524288000
+      compaction: "IncrementalCompactionStrategy"
+      cl: "ONE"
+      col_size: 1024
+      col_n: 1
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   500gb_1t_ics_tablets:
     tag: "sm_20240813114617UTC"
-    schema:
-      keyspace1:
-        - standard1: 500
-    number_of_rows: 524288000
     exp_timeout: 14400  # 4 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "IncrementalCompactionStrategy"
     prohibit_verification_read: false
+    dataset:
+      schema:
+        keyspace1:
+          - standard1: 500
+      num_of_rows: 524288000
+      compaction: "IncrementalCompactionStrategy"
+      cl: "ONE"
+      col_size: 1024
+      col_n: 1
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   500gb_2t_ics:
     tag: "sm_20240819203428UTC"
-    schema:
-      keyspace1:
-        - standard1: 250
-      keyspace2:
-        - standard1: 250
-    number_of_rows: 524288000
     exp_timeout: 14400  # 4 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "IncrementalCompactionStrategy"
     prohibit_verification_read: true
+    dataset:
+      schema:
+        keyspace1:
+          - standard1: 250
+        keyspace2:
+          - standard1: 250
+      num_of_rows: 524288000
+      compaction: "IncrementalCompactionStrategy"
+      cl: "ONE"
+      col_size: 1024
+      col_n: 1
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   1tb_1t_ics:
     tag: "sm_20240814180009UTC"
-    schema:
-      keyspace1:
-        - standard1: 1024
-    number_of_rows: 1073741824
     exp_timeout: 28800  # 8 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "IncrementalCompactionStrategy"
     prohibit_verification_read: false
+    dataset:
+      schema:
+        keyspace1:
+          - standard1: 1024
+      num_of_rows: 1073741824
+      compaction: "IncrementalCompactionStrategy"
+      cl: "ONE"
+      col_size: 1024
+      col_n: 1
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   1tb_4t_twcs:
     tag: "sm_20240821145503UTC"
-    schema:
-      keyspace1:
-        - t_10gb: 10
-        - t_90gb: 90
-        - t_300gb: 300
-        - t_600gb: 600
-    number_of_rows: 428571429
     exp_timeout: 28800  # 8 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "TimeWindowCompactionStrategy"
     prohibit_verification_read: true
+    dataset:
+      schema:
+        keyspace1:
+          - t_10gb: 10
+          - t_90gb: 90
+          - t_300gb: 300
+          - t_600gb: 600
+      num_of_rows: 428571429
+      compaction: "TimeWindowCompactionStrategy"
+      cl:
+      col_size:
+      col_n:
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   1tb_2t_twcs:
     tag: "sm_20240827191125UTC"
-    schema:
-      keyspace1:
-        - t_300gb: 300
-        - t_700gb: 700
-    number_of_rows: 428571429
     exp_timeout: 28800  # 8 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 9
-    compaction_strategy: "TimeWindowCompactionStrategy"
     prohibit_verification_read: true
+    dataset:
+      schema:
+        keyspace1:
+          - t_300gb: 300
+          - t_700gb: 700
+      num_of_rows: 428571429
+      compaction: "TimeWindowCompactionStrategy"
+      cl:
+      col_size:
+      col_n:
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   1.5tb_2t_ics:
     tag: "sm_20240820180152UTC"
-    schema:
-      keyspace1:
-        - standard1: 500
-      keyspace2:
-        - standard1: 1024
-    number_of_rows: 1598029824
     exp_timeout: 43200  # 12 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "IncrementalCompactionStrategy"
     prohibit_verification_read: true
+    dataset:
+      schema:
+        keyspace1:
+          - standard1: 500
+        keyspace2:
+          - standard1: 1024
+      num_of_rows: 1598029824
+      compaction: "IncrementalCompactionStrategy"
+      cl: "ONE"
+      col_size: 1024
+      col_n: 1
+      replication: "NetworkTopologyStrategy"
+      rf: 3
   2tb_1t_ics:
     tag: "sm_20240816185129UTC"
-    schema:
-      keyspace1:
-        - standard1: 2048
-    number_of_rows: 2147483648
     exp_timeout: 57600  # 16 hours
     scylla_version: "2024.2.0-rc1"
     number_of_nodes: 3
-    compaction_strategy: "IncrementalCompactionStrategy"
     prohibit_verification_read: false
+    dataset:
+      schema:
+        keyspace1:
+          - standard1: 2048
+      num_of_rows: 2147483648
+      compaction: "IncrementalCompactionStrategy"
+      cl: "ONE"
+      col_size: 1024
+      col_n: 1
+      replication: "NetworkTopologyStrategy"
+      rf: 3
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -70,21 +70,21 @@ class SnapshotData:
     - bucket: S3 bucket name
     - tag: snapshot tag, for example 'sm_20240816185129UTC'
     - exp_timeout: expected timeout for the restore operation
+    - dataset: dict with snapshot dataset details such as cl, replication, schema, etc.
     - keyspaces: list of keyspaces presented in backup
     - cs_read_cmd_template: cassandra-stress read command template
     - prohibit_verification_read: if True, the verification read will be prohibited. Most likely, such a backup was
       created via c-s user profile.
-    - number_of_rows: number of data rows written in the DB
     - node_ids: list of node ids where backup was created
     """
     bucket: str
     tag: str
     exp_timeout: int
+    dataset: dict[str, str | int | dict]
     keyspaces: list[str]
     ks_tables_map: dict[str, list[str]]
     cs_read_cmd_template: str
     prohibit_verification_read: bool
-    number_of_rows: int
     node_ids: list[str]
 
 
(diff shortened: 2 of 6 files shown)


mikliapko:
@scylladb/qa-maintainers 
Could somebody please take a look?

mgmt_cli_test.py
```diff
@@ -471,12 +471,92 @@ def get_all_snapshot_files(self, cluster_id):
...
+        The name should include all the parameters important for c-s read verification and can be used to
+        recreate such a command based on ks_name.
+        """
+        ks_name = self.ks_name_template.format(
+            size=backup_size,
+            compaction=self._abbreviate_compaction_strategy_name(cs_cmd_params.get("compaction")),
+            cl=cs_cmd_params.get("cl").lower(),
+            col_size=cs_cmd_params.get("col_size"),
+            col_n=cs_cmd_params.get("col_n"),
+            scylla_version=re.sub(r"[-.]", "_", self.params.get("scylla_version")),
```
vponomaryov, line 513: This is wrong way for getting scylla version.

Using AMI this config option will have `None` value.
Or it may have `latest` value.

The `get_version_based_on_conf` existing method should be used instead:
https://github.com/scylladb/scylla-cluster-tests/blame/8f2162aa/sdcm/sct_config.py#L2708

defaults/manager_restore_benchmark_snapshots.yaml
```diff
@@ -1,110 +1,164 @@
 bucket: "manager-backup-tests-permanent-snapshots-us-east-1"
-cs_read_cmd_template: "cassandra-stress read cl=ONE n={num_of_rows} -schema 'keyspace={keyspace_name} replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq={sequence_start}..{sequence_end}"
+cs_read_cmd_template: "cassandra-stress read cl={cl} n={num_of_rows} -schema 'keyspace={keyspace_name} replication(strategy={replication},replication_factor={rf}) compaction(strategy={compaction})' -mode cql3 native -rate threads=500 -col 'size=FIXED({col_size}) n=FIXED({col_n})' -pop seq={sequence_start}..{sequence_end}"
```
vponomaryov, line 2: Why threads=500? 
The number of threads must be dependent on the loader CPU cores.
Assuming common loader has 8 CPU cores and using 500 threads I expect it to be not efficient enough due to context switches.

mgmt_cli_test.py
```diff
@@ -364,6 +364,43 @@ def download_from_azure(self, node, source, destination):
         source = f"{source.replace('azure://', self.backup_azure_blob_service)}{self.backup_azure_blob_sas}"
         node.remoter.sudo(f"azcopy copy '{source}' '{destination}'")
 
+    @staticmethod
+    def create_s3_bucket_aws(name: str, region: str) -> None:
+        LOCALRUNNER.run(f"aws s3 mb s3://{name} --region {region}")
+
+    @staticmethod
+    def create_s3_bucket_gce(name: str, region: str) -> None:
```
vponomaryov, line 372: `s3` is amazon-specific service name.
In case of Google it is called `google storage`.
So, I think it is better to remove the terminology confusion here.

mgmt_cli_test.py
```diff
@@ -364,6 +364,43 @@ def download_from_azure(self, node, source, destination):
...
+        LOCALRUNNER.run(f"gsutil mb -l {region} gs://{name}")
+
+    @staticmethod
+    def sync_s3_buckets_aws(source: str, destination: str, acl: str = 'bucket-owner-full-control') -> None:
+        LOCALRUNNER.run(f"aws s3 sync s3://{source} s3://{destination} --acl {acl}")
+
+    @staticmethod
+    def sync_s3_buckets_gce(source: str, destination: str) -> None:
```
vponomaryov, line 380: Same here.

defaults/manager_restore_benchmark_snapshots.yaml
mikliapko: What approximations/rules can be used to define the optimal number of threads depending on loader CPU cores?

mgmt_cli_test.py
mikliapko: I'm not sure I got it.

`get_version_based_on_conf` is called as part on `init_and_verify_sct_config` where it sets `self.scylla_version` which I'm trying to retrieve in this code.

Or you mean that `get_version_based_on_conf` should be called once again:
```
scylla_version = self.params.get_version_based_on_conf()[0]
```
because during the very first call in `init_and_verify_sct_config` it might be missing some information?

fruch: `param.get('scylla_version')` vs `param.scylla_version`

first is read from what set in configuration 

The second is the computed on you mentioned 

I agree it's a bit confusing, in PR I was trying to refactor sct_config.oy into using pydantic I was changing the name (but this PR is more complicated then I assumed)

defaults/manager_restore_benchmark_snapshots.yaml
vponomaryov: In common case (not related to C-S) it should be like we do it for scylla -> `n_cpus - 1` per each 8 CPU cores.
Having 1-to-1 relationship between CPU core and thread allows to avoid context switches (heavy ops).
Having 1-to-many relationship between CPU core and thread guarantees CPU context switches.
It is pretty possible that some app may perform faster using single thread than 2 multi-threads.

So, it is matter of implementation and measurement to answer your question.

Probably, @roydahan or @soyacz may advice something here.

mgmt_cli_test.py
vponomaryov: > I'm not sure I got it.
> 
> `get_version_based_on_conf` is called as part on `init_and_verify_sct_config` where it sets `self.scylla_version` which I'm trying to retrieve in this code.
> 
> Or you mean that `get_version_based_on_conf` should be called once again:
> 
> ```
> scylla_version = self.params.get_version_based_on_conf()[0]
> ```
> 
> because during the very first call in `init_and_verify_sct_config` it might be missing some information?

Calling `param.get('scylla_version')` you read the config option value as-is. Not the calculated one.

The `scylla_version = self.params.get_version_based_on_conf()[0]` is the calculated one, considers multiple cases.
So, I recommend using this one ^.

defaults/manager_restore_benchmark_snapshots.yaml
mikliapko: @roydahan @soyacz do you have any advice?

soyacz: It depends on machines you use (loaders/db nodes). 
but if 500 works (no timeouts, latency is below 100ms) then it's good to go. If it's not a reggression perf test then there's no point to fine tune it too much - differences are not that big.
If 500 threads is too much (I saw this is used in small tests) then try with 100 or 200 and see.

mikliapko: @soyacz @vponomaryov 
Thanks for your inputs here.
To not block this PR and as this number of threads was there for some time, I suggest to move forward with what we have.
To investigate and define the optimal number of threads, I created the separate [ticket](https://github.com/scylladb/scylla-manager/issues/4237).

mgmt_cli_test.py
mikliapko: Fixed

mgmt_cli_test.py
mikliapko: Fixed

mgmt_cli_test.py
mikliapko: Fixed, retesting https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/siren-tests/job/siren-sct-integration/179/

defaults/manager_restore_benchmark_snapshots.yaml
vponomaryov: Ok, you are aware about this specifics. Let's keep it as is for now.

vponomaryov (approved):
LGTM

mikliapko:
C-S write fails to create keyspace after [changes](https://github.com/scylladb/scylla-cluster-tests/pull/9837#discussion_r1927318767):

```
10:02:18  Traceback (most recent call last):
10:02:18    File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/events_analyzer.py", line 49, in run
10:02:18      raise TestFailure(f"Stress command failed: {event}")
10:02:18  sdcm.sct_events.events_analyzer.TestFailure: Stress command failed: (CassandraStressEvent Severity.CRITICAL) period_type=end event_id=740f2e0a-8a1a-49b8-935f-59416e1a5558 duration=2s: node=Node manager-helper-improve--loader-node-355c5cca-1 [44.202.223.168 | 10.12.2.121]
10:02:18  stress_cmd=cassandra-stress write cl=QUORUM n=1048576 -schema 'keyspace=1gb_ics_quorum_1024_1_2024_2_4_0_20250124_2bc4ec25a8db_1 replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=IncrementalCompactionStrategy)' -mode cql3 native -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..1048576
10:02:18  errors:
10:02:18  
10:02:18  Stress command completed with bad status 1: java.lang.RuntimeException: Encountered exception creating schema
10:02:18  	at org.apache.cassandra.stress.settings.SettingsSchema.createKeySpacesNative(SettingsSchema.java:105)
``` 

Perhaps, keyspace name is too long, need to double-check.

mgmt_cli_test.py
```diff
@@ -391,6 +364,43 @@ def download_from_azure(self, node, source, destination):
         source = f"{source.replace('azure://', self.backup_azure_blob_service)}{self.backup_azure_blob_sas}"
         node.remoter.sudo(f"azcopy copy '{source}' '{destination}'")
 
+    @staticmethod
+    def create_s3_bucket(name: str, region: str) -> None:
```
soyacz, line 368: What are ACL's for these buckets? If open to all, let's make a note here about that.
I think we should protect against creating buckets in root, just to avoid spamming it with test stuff.
Same is for GCE.

mgmt_cli_test.py
```diff
@@ -471,12 +481,126 @@ def get_all_snapshot_files(self, cluster_id):
...
+        with open(path_, encoding="utf-8") as snapshots_yaml:
+            return yaml.safe_load(snapshots_yaml)
+
+    @staticmethod
+    def _abbreviate_compaction_strategy_name(compaction_strategy: str) -> str:
+        """Abbreviate and lower compaction strategy name which comes from c-s cmd to make it more readable in ks name.
+
+        For example, LeveledCompactionStrategy -> lcs or SizeTieredCompactionStrategy -> stcs.
+        """
+        return ''.join(char for char in compaction_strategy if char.isupper()).lower()
```
soyacz, line 500: I propose to raise an error if abbreviated cs len is <3 so If c-s command does not include capitalized cs name it will be flagged.
Also watch for the issue when compaction is not set in c-s - then default is used and this one should whether raise flag to supply it or return default abbreviation.

mgmt_cli_test.py
```diff
@@ -471,12 +481,126 @@ def get_all_snapshot_files(self, cluster_id):
...
+            col_size=cs_cmd_params.get("col_size"),
+            col_n=cs_cmd_params.get("col_n"),
+            scylla_version=re.sub(r"[-.]", "_", self.params.get_version_based_on_conf()[0]),
+        )
+        return ks_name
+
+    def calculate_rows_per_loader(self, overall_rows_num: int) -> int:
+        """Calculate number of rows per loader thread based on the overall number of rows and the number of loaders."""
+        num_of_loaders = int(self.params.get("n_loaders"))
+        return int(overall_rows_num / num_of_loaders)
```
soyacz, line 520: I don't know if precision is important, but e.g. int(10/3 loaders ) = 3, so there's going be 9 rows instead of 10.

sdcm/sct_config.py
```diff
@@ -1202,7 +1202,12 @@ class SCTConfiguration(dict):
 
         dict(name="mgmt_prepare_snapshot_size",
              env="SCT_MGMT_PREPARE_SNAPSHOT_SIZE", type=int,
-             help="Size of backup snapshot in Gb to be prepared to be prepared for backup"),
+             help="Size of backup snapshot in Gb to be prepared for backup"),
+
+        dict(name="mgmt_snapshots_preparer_params",
+             env="SCT_MGMT_SNAPSHOTS_PREPARER_PARAMS", type=dict_or_str_or_pydantic,
+             help="Custom parameter of c-s write operation used in snapshots preparer."
+                  "The whole list of parameters can be found in defaults/manager_snapshots_preparer_config.yaml"),
```
soyacz, line 1210: there's no `mgmt_snapshots_preparer_params` in this config

defaults/manager_snapshots_preparer_config.yaml
```diff
@@ -0,0 +1,23 @@
+# Description: This file contains C-S command template and the default values of parameters used in this command.
```
soyacz, line 1: this file is really confusing, it does not follow our convention of config files.
Why not add these params under `mgmt_snapshots_preparer_params` like we do for other things?

This way you also had to implement specific parser for it.

mikliapko: > Why not add these params under mgmt_snapshots_preparer_params like we do for other things?

I didn't get what you meant here

soyacz: You implemented `get_snapshot_preparer_config` to read this file, but if you just create yaml object like this:
```
mgmt_snapshots_preparer_params:
  - operation: "write"
  - cl: "QUORUM"
  - ...
```

then you could just put it to test defaults. See similar examples like `latency_decorator_error_thresholds` or `teardown_validators`.
This way avoiding custom logic and putting incompatible yaml to defaults directory

mikliapko: ah, got it now, make sense

mikliapko: But wait, the idea of `mgmt_snapshots_preparer_params` is not to provide the whole configuration but overwrite some of default parameters.
If to rework it as we discussed above, any `mgmt_snapshots_preparer_params` input from Jenkins pipeline will rewrite all defaults. But I don't like to provide all of them in Jenkins pipeline if I'd like to adjust only one.

mikliapko: Do we have any mechanism in SCT to not overwrite the dict type parameter completely but only some of the keys that provided in Jenkins pipeline?

soyacz: yes, this param must be `dict_or_str` type and then you can just override specific values in yaml (just use params that need to be overridden) or with env variable `SCT_MGMT_SNAPSHOT...PARAMS.operation=write`

mgmt_cli_test.py
mikliapko: In general, it's a good point.
I haven't considered the number of loaders which is multiple of two but, in overall, it's a valid point and I'll add smth to verify whether the `overall_rows_num` is divisible by `num_of_loaders` without remainder. If not, probably, it's worth throwing an exception since it will affect the consecutive C-S read.

sdcm/sct_config.py
mikliapko: I suppose it's connected to this particular implementation we are discussing [here](https://github.com/scylladb/scylla-cluster-tests/pull/9837#discussion_r1939003487)

mgmt_cli_test.py
mikliapko: > I propose to raise an error if abbreviated cs len is <3 so If c-s command does not include capitalized cs name it will be flagged.

AFAIK, we don't have compaction strategies for which this method will generate < 3 symbols abbreviation (https://opensource.docs.scylladb.com/stable/architecture/compaction/compaction-strategies.html).

> Also watch for the issue when compaction is not set in c-s - then default is used and this one should whether raise flag to supply it or return default abbreviation.

This helper utilization considers defined compaction strategy in C-S write command and that's defined in cmd template.
`cs_cmd_template: "cassandra-stress {operation} ... compaction(strategy={compaction})' ..."`

mikliapko: The purpose of these methods is to be used by backup snapshots preparator only which is supposed to use pre-defined  `cs_cmd_template`, thus, eliminating the need for some extra verifications.

defaults/manager_snapshots_preparer_config.yaml
mikliapko: Fixed and retested here:
https://argus.scylladb.com/tests/scylla-cluster-tests/938e55be-4527-4b11-bb41-9faa157bf417

mgmt_cli_test.py
mikliapko: Fixed and retested here:
https://argus.scylladb.com/tests/scylla-cluster-tests/938e55be-4527-4b11-bb41-9faa157bf417

mikliapko:
> C-S write fails to create keyspace after [changes](https://github.com/scylladb/scylla-cluster-tests/pull/9837#discussion_r1927318767):
> 
> ```
> 10:02:18  Traceback (most recent call last):
> 10:02:18    File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/events_analyzer.py", line 49, in run
> 10:02:18      raise TestFailure(f"Stress command failed: {event}")
> 10:02:18  sdcm.sct_events.events_analyzer.TestFailure: Stress command failed: (CassandraStressEvent Severity.CRITICAL) period_type=end event_id=740f2e0a-8a1a-49b8-935f-59416e1a5558 duration=2s: node=Node manager-helper-improve--loader-node-355c5cca-1 [44.202.223.168 | 10.12.2.121]
> 10:02:18  stress_cmd=cassandra-stress write cl=QUORUM n=1048576 -schema 'keyspace=1gb_ics_quorum_1024_1_2024_2_4_0_20250124_2bc4ec25a8db_1 replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=IncrementalCompactionStrategy)' -mode cql3 native -rate threads=500 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..1048576
> 10:02:18  errors:
> 10:02:18  
> 10:02:18  Stress command completed with bad status 1: java.lang.RuntimeException: Encountered exception creating schema
> 10:02:18  	at org.apache.cassandra.stress.settings.SettingsSchema.createKeySpacesNative(SettingsSchema.java:105)
> ```
> 
> Perhaps, keyspace name is too long, need to double-check.

Fixed and retested here:
https://argus.scylladb.com/tests/scylla-cluster-tests/938e55be-4527-4b11-bb41-9faa157bf417

soyacz (approved):
LGTM

soyacz:
need one more qa approval to merge it. @vponomaryov can you help?

mikliapko:
@fruch @vponomaryov We are still missing one maintainer's approval. 
Could you please take a look?

fruch (approved):

PR #9841, opened 2025-01-16
fix(hydra): enable python faulthandler by default

fruch:
we are running into case SCT is core dumping during shutdown of the tests we don't have enough information to understand what's going on

hopefully this can add some information that can help us figure it out

Ref: https://docs.python.org/3/library/faulthandler.html

### Testing

- [x] provision tests

### Reminders


dimakr (approved):
LGTM

nyh:
Let me know how this worked out :-) I wonder if the cause of SCT's Python shutdown crashes are related to boto3 like I suspected (without proof) in the Alternator test - or something else.

PR #9843, opened 2025-01-16
fix(test-case): update 5000 tables test case configuration

vponomaryov:
List of changes:
- Disable per-table metrics due to significant perf impact.
- Enable cluster health checks which work with this case just fine.
- Decrease the nemesis interval from 60 minutes to just 3 keeping in mind that health checks will take some time too.
- Reduce stress time for each of the 5000 commands. Having 20 minutes per cmd we will get about 1.5 days long test runs instead of the 2.5 days.
- Reduce number of loaders from 5 to 3 to use resources more efficiently. In current case the bottleneck is the RAM.

Note that this scenario hits following bug:
- https://github.com/scylladb/scylla-enterprise/issues/5093

If `destroy_data_then_repair` nemesis gets triggered aganst the setup of this scenario.

### Testing

- [scylla-staging/valerii/vp-scale-5000-tables-test#7](https://argus.scylladb.com/tests/scylla-cluster-tests/742f937f-a79f-45fd-bfcb-4d4b1d267e12)

### Reminders

### DIFF
--- a/test-cases/scale/longevity-5000-tables.yaml
+++ b/test-cases/scale/longevity-5000-tables.yaml
@@ -1,6 +1,6 @@
-test_duration: 2160
+test_duration: 2520
 
-cs_duration: '55m'
+cs_duration: '20m'
 cs_user_profiles:
     - scylla-qa-internal/cust_d/templated_tables_mv.yaml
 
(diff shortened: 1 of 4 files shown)


test-cases/scale/longevity-5000-tables.yaml
```diff
@@ -9,24 +9,25 @@ pre_create_schema: true
...
 nemesis_seed: '404'
-nemesis_interval: 60
+# NOTE: health checks will take some time, so, don't add significant waiting in this part
+nemesis_interval: 3
 
-# TODO: remove when https://github.com/scylladb/scylla-tools-java/issues/175 resolved
-stop_test_on_stress_failure: false
+append_scylla_yaml:
+  # NOTE: https://github.com/scylladb/scylla-monitoring/issues/2429
+  enable_node_aggregated_table_metrics: false
```
roydahan, line 33: I'm fine with it, please comment in that issue with a link to this PR that you're disabling the table metrics collection for this test.

roydahan (approved):

test-cases/scale/longevity-5000-tables.yaml
vponomaryov: Done
https://github.com/scylladb/scylla-monitoring/issues/2429#issuecomment-2597925630

mykaul:
Is this with tablets or with vnodes? (we probably need for both and compare between them)

fruch (approved):
LGTM

fruch:
> Is this with tablets or with vnodes? (we probably need for both and compare between them)

test can run with both, and the issue mentioned above happened with both

you can see the summerie here:
https://github.com/scylladb/qa-tasks/issues/1820#issuecomment-2596369089

but it doesn't seem like anyone is attending to the issue raised by @vponomaryov

PR #9844, opened 2025-01-19
feature(createArgusTestRun.groovy): Add argus link

### LINKED ISSUE #9634: Put links to Argus in jenkins run descriptions
This kind of 'button' or link will provide clear mapping of Jenkins job to corresponding argus run. 
Such a feature will enhance user experience with Argus's capabilities for test verification, log collection, and error event analysis and centralizes test management within Argus for improved efficiency.

Such a link may be added to job description, for example (as been discussed with Alexey)
Example:
Image

k0machi:
This commit makes it so that createArgusTestRun stage of the SCT
pipeline now appends a link to the argus page of the run.

Fixes #9634

### Testing

- [x] [Argus](https://argus.scylladb.com/tests/scylla-cluster-tests/da7f287a-7c42-465e-b9bc-21bdf58ae0dd)
- [x] [Jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/artifacts-ubuntu2204-test/10/)

### Reminders

### DIFF
--- a/vars/createArgusTestRun.groovy
+++ b/vars/createArgusTestRun.groovy
@@ -16,4 +16,17 @@ def call(Map params) {
 
         echo " Argus test run created."
     """
+    if (!currentBuild.description) {
+        currentBuild.description = ''
+    }
+    String runButton = """
+        <div style="margin: 12px 4px;">
+            <a
+                href='https://argus.scylladb.com/tests/scylla-cluster-tests/${SCT_TEST_ID}'
+            >
+                Argus: <span style='font-weight: 500'>${SCT_TEST_ID}</span>
+            </a>
+        </div>
+    """
+    currentBuild.description += "${runButton}"
 }


soyacz (dismissed):
LGTM

fruch (approved):
LGTM

vars/createArgusTestRun.groovy
```diff
@@ -16,4 +16,8 @@ def call(Map params) {
 
         echo " Argus test run created."
     """
+    if (!currentBuild.description) {
+        currentBuild.description = ''
+    }
+    currentBuild.description += "\n<a href='https://argus.scylladb.com/tests/scylla-cluster-tests/${SCT_TEST_ID}'>Argus</a>\n"
```
fruch, line 22: we should add the test id, or part of it into the test
in case there are multiple test ids in one job (and we do have those)

fruch (changes requested):
we need to support multiple test ids in one job

vars/createArgusTestRun.groovy
k0machi: Done.. with a small improvement to the visibility.

image

k0machi: However there is a small issue in the build sidebar:
image

k0machi: Okay, fixed with text-overflow, unfortunately it cuts all text in the sidebar:
image

k0machi: image
image

Final-ish look

fruch (approved):
LGTM

soyacz:
@fruch  wouldn't it be better if visible name shown test name instead of test id - so we could jump straight to desired test (for parallel runs). I don't think it's hard.

fruch:
> @fruch wouldn't it be better if visible name shown test name instead of test id - so we could jump straight to desired test (for parallel runs). I don't think it's hard.

each parallel one had a bit different logic, in some it's the name, in some it's instance type.
as is it cover 95% of what what we need

PR #9868, opened 2025-01-20
feature(ruff): scan for mutable-argument-default (B006)

fruch:
we shouldn't let those get into the the code without a very good reason (like memoize/cache)

Ref: https://docs.astral.sh/ruff/rules/mutable-argument-default/

### Testing

- [x] no testing needed

### Reminders


soyacz (approved):
LGTM

Issue #9322, opened 2024-11-21
Create/update docs for docker backend capabilities


dimakr:
It would be beneficial to have a dedicated docs page for docker backend where we can describe specifics of the docker backend, list the capabilities and limitations, etc.

For instance, now the backend installs the monitoring stack directly on the host system, not on a dedicated docker instance. But due to SCT installs scylla-manager server on monitoring node, we need to skip manager use in case of docker backend.
This is an example of limitation that is not mentioned explicitly in docs and occasionally causes devs to run test configurations that are not fit for the docker backend, resulting in new SCT issue reports.

PR #9869, opened 2025-01-21
docs(docker-backend): summarize specifics and limitations of backend

dimakr:
Add a dedicated documentation page for the Docker backend, which describes specifics and limitations of the backend.

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/9322

### DIFF
--- a/docs/docker-backend-overview.md
+++ b/docs/docker-backend-overview.md
@@ -1,0 +1,34 @@
+# Docker backend specifics
+
+### Overview
+SCT supports multiple backends for deploying and testing Scylla clusters, with the Docker backend being a practical choice for local development and testing.<br>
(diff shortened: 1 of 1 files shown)


soyacz (approved):
When running desired test with docker backend usually it will fail due to overload.
Maybe worth to add tips for successful work with docker backend - typical/example stress_cmd that will not be too heavy and how not preload cluster with 650GB data...

In general, this is an useful doc!

fruch:
> When running desired test with docker backend usually it will fail due to overload. Maybe worth to add tips for successful work with docker backend - typical/example stress_cmd that will not be too heavy and how not preload cluster with 650GB data...

this can be a followup on this one

> 
> In general, this is an useful doc!

we need more of those per backend, and per test types, upgrades, longevity, performance

fruch (approved):
LGTM

PR #9884, opened 2025-01-21
fix(_run_all_stress_cmds): skip and warn about using counters

fruch:
since counters are not supported with tablets we should skip it, and error out if such test is using counters

Ref: https://github.com/scylladb/scylladb/issues/18180

### Testing

- [x] provision tests

### Reminders

### DIFF
--- a/sdcm/utils/loader_utils.py
+++ b/sdcm/utils/loader_utils.py
@@ -15,11 +15,14 @@
 import os
 import re
 import time
+from functools import cached_property
 
 from sdcm.sct_events import Severity
 from sdcm.sct_events.system import TestFrameworkEvent
 from sdcm.utils.common import skip_optional_stage
 from sdcm.utils.decorators import optional_stage
+from sdcm.utils.issues import SkipPerIssues
+from sdcm.utils.features import is_tablets_feature_enabled
 
 DEFAULT_USER = "cassandra"
 DEFAULT_USER_PASSWORD = "cassandra"
@@ -92,6 +95,12 @@ class LoaderUtilsMixin:
             }
             self._run_all_stress_cmds(stress_queue, params)
 
+    @cached_property
+    def tablets_enabled(self):
+        # is tablets feature enabled in Scylla configuration.
+        with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
+            return is_tablets_feature_enabled(session)
+
     def _run_all_stress_cmds(self, stress_queue, params):
         stress_cmds = params['stress_cmd']
         if not isinstance(stress_cmds, list):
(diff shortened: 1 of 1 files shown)


sdcm/utils/loader_utils.py
```diff
@@ -106,8 +115,15 @@ def _run_all_stress_cmds(self, stress_queue, params):
             stress_params.update({'stress_cmd': stress_cmd})
 
             # Due to an issue with scylla & cassandra-stress - we need to create the counter table manually
+            # also tables doesn't yet support counters, so we skip the command and error about it
             if 'counter_' in stress_cmd:
-                self._create_counter_table()
+                if self.tablets_enabled and SkipPerIssues('scylladb/scylladb#18180', params=self.params):
+                    TestFrameworkEvent(severity=Severity.ERROR,
```
fruch, line 121: no sure if we should error like that, or just warning about it ? 

anyhow it's better then before that it was failing the test completely, like that it would run the whole test, and fail it in the end cause of the error event

sdcm/utils/loader_utils.py
```diff
@@ -106,8 +115,15 @@ def _run_all_stress_cmds(self, stress_queue, params):
             stress_params.update({'stress_cmd': stress_cmd})
 
             # Due to an issue with scylla & cassandra-stress - we need to create the counter table manually
+            # also tables doesn't yet support counters, so we skip the command and error about it
```
yarongilor, line 118: tables -> tablets

sdcm/utils/loader_utils.py
yarongilor: it make sense to emit such and error and fail the test.
No sure anyone will bother looking at warnings.

sdcm/utils/loader_utils.py
fruch: fixed

sdcm/utils/loader_utils.py
roydahan: I don't know if it's better.
Do we have tests that mixes both counters and other stress types?

I know we have some dedicated for counters, will they run without stress or just finish quickly?
In any case, they should be "Red" so one will know they didn't run as planned.

fruch: checking it again, we don't have such test cases

it appear only on the provision test, and in some other unit test configurations...

which is quite weird...

fruch: I've found we only have 2 tests covering counters and they are using scylla-bench only for some reason.

so this fix isn't that important, it's just hack for when we use c-s (maybe it didn't create the table correctly at some point...)

yarongilor: it looks like `test-cases/longevity/longevity-counters-multidc.yaml` and `test-cases/longevity/longevity-counters-3h.yaml` are not tested.

yarongilor (approved):
LGTM

sdcm/utils/loader_utils.py
roydahan: Why does it look so?
They are tested during releases.

roydahan: So if one runs them now with tablets they will fail the same, right?

Which is probably good, because it can be used as a negative test in case someone wants it.

For 2025.1 they will need to be tested only for "tablets-disabled" variation.
@pehala FYI

fruch: Anyhow i'm merging it, it doesn't really affect any real test case, just the one used in provisioning test

Issue #9645, opened 2025-01-02
artifact test fail on first hydra command [error mounting "tmpfs" to rootfs at ".../.local": create mountpoint for .../.local mount: mkdirat .../.local: file exists: unknown.]


fruch:
recently we run multiple times (2 so far), in a failure as such:
```
[2025-01-02T08:53:30.461Z] + echo 'Creating Argus test run ...'
[2025-01-02T08:53:30.461Z] Creating Argus test run ...
[2025-01-02T08:53:30.461Z] + [[ -n '' ]]
[2025-01-02T08:53:30.462Z] + export SCT_CLUSTER_BACKEND=gce
[2025-01-02T08:53:30.462Z] + SCT_CLUSTER_BACKEND=gce
[2025-01-02T08:53:30.462Z] + export SCT_CONFIG_FILES=test-cases/artifacts/debian11.yaml
[2025-01-02T08:53:30.462Z] + SCT_CONFIG_FILES=test-cases/artifacts/debian11.yaml
[2025-01-02T08:53:30.462Z] + ./docker/env/hydra.sh create-argus-test-run
[2025-01-02T08:53:30.462Z] Running on Build Server...
[2025-01-02T08:53:30.462Z] Pull version v1.83-pycodestyle-2.10.0 from Docker Hub...
[2025-01-02T08:53:30.462Z] v1.83-pycodestyle-2.10.0: Pulling from scylladb/hydra
[2025-01-02T08:53:30.462Z] 14726c8f7834: Pulling fs layer
[2025-01-02T08:53:30.462Z] 7d676dc8a994: Pulling fs layer
[2025-01-02T08:53:30.462Z] 03bdd165d0d2: Pulling fs layer
[2025-01-02T08:53:30.462Z] 428bad6fa242: Pulling fs layer
[2025-01-02T08:53:30.462Z] 63cd35141f3a: Pulling fs layer
[2025-01-02T08:53:30.462Z] 4e009f26260c: Pulling fs layer
[2025-01-02T08:53:30.462Z] df1671404e21: Pulling fs layer
[2025-01-02T08:53:30.462Z] b56fb9b9b894: Pulling fs layer
[2025-01-02T08:53:30.462Z] 9f5e78de2e2e: Pulling fs layer
[2025-01-02T08:53:30.462Z] 4b82f5a588e0: Pulling fs layer
[2025-01-02T08:53:30.462Z] b6a5cdc11fcc: Pulling fs layer
[...]

PR #9886, opened 2025-01-21
fix(createArgusTestRun): retry 3 times

fruch:
since we are running into cases docker is failing
on this first stage, we would retry it 3 times
with hope it would be enough to avoid those type
of error

Ref: scylladb/scylla-cluster-tests#9645

### Testing

- [ ]

### Reminders

### DIFF
--- a/vars/createArgusTestRun.groovy
+++ b/vars/createArgusTestRun.groovy
@@ -2,20 +2,22 @@
 
 def call(Map params) {
     def test_config = groovy.json.JsonOutput.toJson(params.test_config)
-    sh """#!/bin/bash
-        set -xe
+    retry(3) {
+		sh """#!/bin/bash
+			set -xe
 
-        echo "Creating Argus test run ..."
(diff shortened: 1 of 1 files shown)


vars/createArgusTestRun.groovy
```diff
@@ -2,20 +2,22 @@
 
 def call(Map params) {
     def test_config = groovy.json.JsonOutput.toJson(params.test_config)
-    sh """#!/bin/bash
-        set -xe
+    retry(3) {
```
vponomaryov, line 5: Where does this func come from?

fruch: from jenkins
https://www.jenkins.io/doc/pipeline/steps/workflow-basic-steps/#retry-retry-the-body-up-to-n-times

vponomaryov (approved):
LGTM

fruch:
Looks like it's worked o.k. we'll merge and see if it helped out

PR #9887, opened 2025-01-21
fix(performance): increase latency limits

soyacz:
We would not chase issues with slight cross of low latency limits e.g. for p90 or in some low throughput tests. Increased these values to not raise red flag for whole test.

### Testing

- [ ]

### Reminders

### DIFF
--- a/configurations/performance/latency-decorator-error-thresholds-steps-ent-tablets.yaml
+++ b/configurations/performance/latency-decorator-error-thresholds-steps-ent-tablets.yaml
@@ -13,7 +13,7 @@ latency_decorator_error_thresholds:
       P90 read:
         fixed_limit: 1
       P99 read:
-        fixed_limit: 1
+        fixed_limit: 3
     "300000":
       P90 read:
         fixed_limit: 1
@@ -23,7 +23,17 @@ latency_decorator_error_thresholds:
       P90 read:
         fixed_limit: 1
       P99 read:
+        fixed_limit: 5
+    "600000":
+      P90 read:
         fixed_limit: 3
+      P99 read:
+        fixed_limit: 50
+    "700000":
+      P90 read:
+        fixed_limit: 3
+      P99 read:
+        fixed_limit: 60
     unthrottled:
       P90 read:
         fixed_limit: null
@@ -60,6 +70,15 @@ latency_decorator_error_thresholds:
         fixed_limit: 5
       P99 read:
         fixed_limit: 5
+    "450000":
+      P90 write:
+        fixed_limit: 5
+      P90 read:
+        fixed_limit: 5
+      P99 write:
+        fixed_limit: 15
+      P99 read:
+        fixed_limit: 15
     unthrottled:
       P90 write:
         fixed_limit: null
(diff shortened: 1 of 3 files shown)


fruch (approved):
LGTM

roydahan:
Doesn't LGTM.

I don't want to just increase everything.
Based on what @juliayakovlev  shared, we have a step (I think it's the 150K) which is 0.5ms with vnodes and 2ms with tablets.

We need our tests to detect such cases and it's good we set the bar upfront with our best release.

There are customers who looks for sub-millisecond latency and we need to make sure we don't regress that either.

soyacz:
> Doesn't LGTM.
> 
> I don't want to just increase everything. Based on what @juliayakovlev shared, we have a step (I think it's the 150K) which is 0.5ms with vnodes and 2ms with tablets.
> 
> We need our tests to detect such cases and it's good we set the bar upfront with our best release.
> 
> There are customers who looks for sub-millisecond latency and we need to make sure we don't regress that either.

Sure, I can change only ones that we need: currently failing and there's decision not to investigate - like p90 5ms requirement (still I don't know if 7ms is good threshold). 
@juliayakovlev can you share which ones exactly should be tuned?

juliayakovlev:
> > Doesn't LGTM.
> > I don't want to just increase everything. Based on what @juliayakovlev shared, we have a step (I think it's the 150K) which is 0.5ms with vnodes and 2ms with tablets.
> > We need our tests to detect such cases and it's good we set the bar upfront with our best release.
> > There are customers who looks for sub-millisecond latency and we need to make sure we don't regress that either.
> 
> Sure, I can change only ones that we need: currently failing and there's decision not to investigate - like p90 5ms requirement (still I don't know if 7ms is good threshold). @juliayakovlev can you share which ones exactly should be tuned?

There is difference in ` P99 read` for read step. For tablets it is 3, for vnodes is 5

soyacz:
My conclusions from today's meeting:
1. there's no point in raising red flag for latency >10ms for throughputs that are higher than 50% of max
2. From one side we don't have boundries for latencies above 50% max throughput, but based on historical results we can set some limits. Still might be flaky and red flag might be raised prematurely, but at least it will point our attention to it (and possibly ignore if it's one time issue).

Concluding, we should again check latencies and set new limits for the above according to historical results + margin. For tablets tests we should adjust the numbers after fixing test with initial tablet count - values should be same as for vnodes (??).

@roydahan did I conclude it correctly?

roydahan:
Yes, almost :)
There is a point raising red flag for 1 and the reason is 2.

And yes, we should based it on the history, but not the "max" but the typical latency we're getting as long as it consistent.

soyacz:
Ok, I adjusted values. Also see base p90 latency was changed to 7ms - I'm not sure if it's right.

configurations/performance/latency-decorator-error-thresholds-steps-ent-tablets.yaml
```diff
@@ -13,7 +13,7 @@ latency_decorator_error_thresholds:
       P90 read:
         fixed_limit: 1
       P99 read:
-        fixed_limit: 1
+        fixed_limit: 3
```
soyacz, line 16: see this number is greater than for 300k - due it's first run and Scylla's cache was not fully populated

fruch (approved):
LGTM

soyacz:
rebased, conflicts fixed

fruch:
@juliayakovlev you took this one, right ?

PR #9888, opened 2025-01-21
fix(unit-tests): stop triggering health checks in longevity unit test

vponomaryov:
The `test_longevity.py::test_test_user_batch_custom_time` unit test uses
the `test-cases/scale/longevity-5000-tables.yaml` config file for running
a short longevity test which triggers a nemesis.

If health checks are enabled then the `nemesis call` runs much longer while health checks
are completed 2 times - before and after the nemesis.
And while it is ongoing the nemesis lock gets held.
And the problem with it is that it runs even after finish of this test.
So, any another unit test which tries to get a nemesis lock will stumble upon a held lock for 10+ minutes.

It started happening after the merge of the PR (https://github.com/scylladb/scylla-cluster-tests/pull/9843) which enabled health checks in the mentioned config file.
The affected test is following:
- `test_nemesis.py::test_list_nemesis_of_added_disrupt_methods`

Alphabetically it runs after the `test_longevity.py::test_test_user_batch_custom_time` one.

So, disable health checks to avoid side-effects and doing redundant stuff which were not planned when the test was written.

### Testing

- [ ]

### Reminders

### DIFF
--- a/unit_tests/test_longevity.py
+++ b/unit_tests/test_longevity.py
@@ -16,6 +16,17 @@ def test_test_user_batch_custom_time(params):
     class DummyLongevityTest(LongevityTest):
         def _init_params(self):
             self.params = params
+            # NOTE: running this test we get a nemesis trigerred,
+            # and if health checks are enabled then the nemesis lock gets held
+            # while it's checks are running. It may run for more than 10 minutes.
+            # Additional problem is that it is not controlled by this test,
+            # thread just runs as a side-car even after finish of this test.
+            # In this case any further unit test which runs a nemesis
+            # will stumble upon a held lock.
+            # One of such tests is following:
+            # - test_nemesis.py::test_list_nemesis_of_added_disrupt_methods
+            # So, disable health checks here.
+            self.params["cluster_health_check"] = False
 
         def start_argus_heartbeat_thread(self):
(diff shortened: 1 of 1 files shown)


vponomaryov:
Example of the backport PR where this is reproduced in 100% unit tests runs:
- https://github.com/scylladb/scylla-cluster-tests/pull/9846

vponomaryov:
@soyacz , @fruch 
The `master` branch, `2024.1`, `6.1` and `6.2` are also affected, but not fail because health checks take there only 10 seconds.
It looks like following:
```
12:37:10  unit_tests/test_longevity.py::LongevityTest::test_user_batch_custom_time SKIPPED [ 25%]
12:37:40  unit_tests/test_longevity.py::test_test_user_batch_custom_time PASSED    [ 25%]
...
12:38:14  unit_tests/test_microbenchmarking.py::TestMBM::test_object_exists PASSED [ 28%]
12:38:24  unit_tests/test_nemesis.py::test_list_nemesis_of_added_disrupt_methods PASSED [ 28%]
```

fruch (commented):
One other option would be to mock out the nemesis code completely

fruch (approved):
LGTM

PR #9895, opened 2025-01-21
fix(sct_config): `param.get()` doesn't accept default argument

fruch:
the newly introduced `update_config_based_on_version()` had faulty code that assumed there a default vaule argument
default should be configured in `test_default.yaml`

a new unit test introduce to cover this part of the sct_config.py

### Testing

- [ ]

### Reminders


dimakr (approved):
LGTM

wmitros (approved):
LGTM, after setting the default to `[]` we should be able to overwrite it only with another list, so we won't get `None` from `self.get('experimental_features')` (at least when getting the config from yaml or `SCT_EXPERIMENTAL_FEATURES`, can we set the config another way?)

In the unit test we could also check that the correct experimental feature gets set, but the issue was just about wrong number of parameters in the function call, so perhaps that's not necessary

PR #9901, opened 2025-01-22
fix(get_version_based_on_conf): master is now "enterprise"

fruch:
test was assuming any version from master wouldn't be with enterprise version scheme, that has change and now master is with 2025.1 and that's expected, the logic of the test has change to expect enterprise like version

### Testing

- [x] unittests

### Reminders


soyacz (approved):
LGTM

PR #9910, opened 2025-01-23
feature(aws-unittests): new unit test module for test aws services 

fruch:
in this commit we would be using moot in a server mode, running
inside a thread

in this way we don't have external services dependencies or
need to spin up docker instance with a mock implementation.

for now it's just demonstrate what we can achieve with it

* cover Keystore implementation
* cover AWS Provisioning (on case)
* cover listing of AMIs

### Testing

- [ ]

### Reminders


soyacz (approved):
This looks like simplification and more stable tests - LGTM.
I saw some failure related to docker pull rate limit thou - possibly need to fix something here/in pipeline

fruch:
> This looks like simplification and more stable tests - LGTM.
> I saw some failure related to docker pull rate limit thou - possibly need to fix something here/in pipeline

Seems like it's happening across the board, I'm guessing we are hitting our account limit, and it's not related to those changes, on the contrary, we are removing some docker usage from our CI

PR #9917, opened 2025-01-27
improvement(artifacts): switch to syslog-ng logs transport

dimakr:
Switch GCE and Azure based artifacts tests from ssh to syslog-ng logs transport.

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/8532

### Testing

- [x] :green_circle: [artifacts-gce-image](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts-gce-image-test/3/)
- [x] :green_circle: [artifacts-azure-image](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts-azure-image-test/)

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #9926, opened 2025-01-28
Revert "improvement(artifacts): switch to syslog-ng logs transport"

fruch:
Seems like we didn't fully tested this on on the jobs not based scylla images

Revering for now

Reverts scylladb/scylla-cluster-tests#9917


yaronkaikov:
it seems that this commit is not part of 2025.1

yaronkaikov:
This commit is not part of 2025.1 no need to backport

PR #9948, opened 2025-01-29
improvement(cleanup): add capacity reservation cleanup

soyacz:
In case of pipeline failure where CR's are used (or any other failure in CR cleanup within the test) we need to cancel them as soon as possible.

Rule:
There never should be active capacity reservations without allocated machines that are older than 15 minutes
(it's enough time to create machines with it).

Added CR cancellation to our periodic cleanup scripts.

refs: https://github.com/scylladb/qa-tasks/issues/1846

### Testing

- [X] - tested manually

### Reminders


soyacz:
btw. I saw we run cleanup scripts every 3h. I propose to change it to 1h.

fruch (approved):

fruch:
> btw. I saw we run cleanup scripts every 3h. I propose to change it to 1h.

Let's change it

PR #9985, opened 2025-02-03
fix(ci-pipeline): fix integration and provision test

soyacz:
Integration tests may fail with docker rate limiting - added docker login command to avoid that.
Provision tests fail with running counter tests against latest scylla version (tablets enabled by default). Disabled tablets in provision test config

### Testing

- [ ] - CI tests (integration + provision aws)

### Reminders

### DIFF
--- a/test-cases/PR-provision-test.yaml
+++ b/test-cases/PR-provision-test.yaml
@@ -32,3 +32,6 @@ use_preinstalled_scylla: true
 post_behavior_db_nodes: "destroy"
 post_behavior_loader_nodes: "destroy"
 post_behavior_monitor_nodes: "destroy"
+
+append_scylla_yaml:
+  enable_tablets: false  # counters are not supported with tablets
--- a/unit_tests/test_cassandra_stress_thread.py
+++ b/unit_tests/test_cassandra_stress_thread.py
@@ -79,6 +79,7 @@ def test_02_cassandra_stress_user_profile(request, docker_scylla, params):
     assert float(output[0]["latency 99th percentile"]) > 0
 
 
+@pytest.mark.skip("Disabled due to the https://github.com/scylladb/scylla-cluster-tests/issues/9988")
 @pytest.mark.docker_scylla_args(ssl=True)
 def test_03_cassandra_stress_client_encrypt(request, docker_scylla, params):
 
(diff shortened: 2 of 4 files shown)


vponomaryov (dismissed):
LGTM

test-cases/PR-provision-test.yaml
```diff
@@ -32,3 +32,6 @@ use_preinstalled_scylla: true
 post_behavior_db_nodes: "destroy"
 post_behavior_loader_nodes: "destroy"
 post_behavior_monitor_nodes: "destroy"
+
+append_scylla_yaml:
+  enable_tablets: false  # counters are not supported with tablets
```
pehala, line 37: How about adding `no_tablets.yaml` into the pipeline instead?

unit_tests/test_cassandra_stress_thread.py
```diff
@@ -79,6 +79,7 @@ def cleanup_thread():
     assert float(output[0]["latency 99th percentile"]) > 0
 
 
+@pytest.mark.skip("Disabled due to the https://github.com/scylladb/scylla-cluster-tests/issues/9988")
```
pehala, line 82: Could we use skipif instead?

soyacz: I'd require to implement fetching state of the issue or use SkipPerIssue, but I don't know how to use it in context of a unit test decorator.
I think reenabling it upon fixing the issue is good enough.

test-cases/PR-provision-test.yaml
soyacz: what difference does it make in this context?

pehala: If I look at the pipeline (which is first point of contact with the test for me) I know immediately it has tablets turned off, this way it is somewhere down in the configs.

Also I do not see any counters mention in this config, so I do not think it is important to this config, it is important to the pipeline

soyacz: actually, this test config use counters in `stress_cmd` - so it is related specifically to this one

pehala: Ah ok, sorry I totally missed that one. In that case, it is alright

soyacz:
@pehala @vponomaryov I tried to fix c-s test as disabling it just moved issue to another test with ssl. I nailed the problem (certs generation took long time to be created for the first test with ssl) but still there's an issue with https://github.com/scylladb/scylla-cluster-tests/issues/9988
Let's merge this one so I can push forward https://github.com/scylladb/scylla-cluster-tests/pull/9979

pehala (approved):

vponomaryov (approved):
LGTM

PR #9015, opened 2024-10-18
improvement(argus-client): ship argus client within sct

soyacz:
In order to ease update argus client, it was removed from docker image and included in SCT repository. This way we may update argus client without rebuilding hydra image and easily backport it to other sct branches.

Added simple script to streamline this process.

fixes: https://github.com/scylladb/qa-tasks/issues/1766

### Testing

- [X] - [example test](https://argus.scylladb.com/tests/scylla-cluster-tests/6d1c3244-a8c2-4d8a-adc1-ae022ba3100c)

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #9989, opened 2025-02-04
fix(argus): Fix Ruff errors

pehala:
### Testing

- [x] None

### Reminders

### DIFF
--- a/argus/client/__init__.py
+++ b/argus/client/__init__.py
@@ -1,1 +1,1 @@
-from argus.client.base import ArgusClient
+from argus.client.base import ArgusClient # noqa: F401
--- a/argus/client/generic_result.py
+++ b/argus/client/generic_result.py
@@ -68,7 +68,7 @@ class ResultTableMeta(type):
                 if col_name not in cls_instance.column_types:
                     raise ValueError(f"ValidationRule column {col_name} not found in the table")
                 if cls_instance.column_types[col_name] == ResultType.TEXT:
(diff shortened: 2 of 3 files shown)


vponomaryov (changes requested):
The `argus/*` are ported as-is from the https://github.com/scylladb/argus/ repo.
So, need to fix these stuff there first.

pehala:
> The `argus/*` files are ported as-is from the https://github.com/scylladb/argus/ repo.
> So, need to fix these stuff there first.

Why are we copying files from there? Shouldnt we use dependency to install it, instead of copying?

vponomaryov:
> > The `argus/*` files are ported as-is from the https://github.com/scylladb/argus/ repo.
> > So, need to fix these stuff there first.
> 
> Why are we copying files from there? Shouldnt we use dependency to install it, instead of copying?

See following:
- https://github.com/scylladb/scylla-cluster-tests/pull/9015

fruch:
@pehala we are ignore them on purpose, why are you trying to fix those ?

pehala:
> @pehala we are ignore them on purpose, why are you trying to fix those ?

I see that there are ignored on purpose. I created this patch because when I ran ruff locally (not as a part of pre check), they were automatically changed. I will close this and instead add a exclude path for ruff in general

pehala:
Instead of this I opened #10124 so the cehcking is consistent between precommit and general run

PR #9993, opened 2025-02-04
fix(LoaderUtilsMixin): table_enabled was allways True as this feature is allways enabled

jsmolar:
Enabling/disabling tablets was changed. Now we have to check scylla yaml or keystore schema.
For more info see
- https://github.com/scylladb/scylladb/pull/21451
- https://github.com/scylladb/scylladb/pull/21614

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/jsmolar/job/longevity-counters-3h-test/9/

### Reminders

### DIFF
--- a/sdcm/utils/features.py
+++ b/sdcm/utils/features.py
@@ -16,7 +16,6 @@ from cassandra.cluster import Session  # pylint: disable=no-name-in-module
 
 CONSISTENT_TOPOLOGY_CHANGES_FEATURE = "SUPPORTS_CONSISTENT_TOPOLOGY_CHANGES"
 CONSISTENT_CLUSTER_MANAGEMENT_FEATURE = "SUPPORTS_RAFT_CLUSTER_MANAGEMENT"
-TABLETS_FEATURE = "TABLETS"
 
 LOGGER = logging.getLogger(__name__)
 
@@ -75,8 +74,14 @@ def is_consistent_topology_changes_feature_enabled(session: Session) -> bool:
     return CONSISTENT_TOPOLOGY_CHANGES_FEATURE in get_enabled_features(session)
 
 
-def is_tablets_feature_enabled(session: Session) -> bool:
+def is_tablets_feature_enabled(node) -> bool:
     """ Check whether tablets enabled
-    if you need from a specific node use `patient_exclusive_cql_connection` session
     """
-    return TABLETS_FEATURE in get_enabled_features(session)
+    with node.remote_scylla_yaml() as scylla_yaml:
+        # for backward compatibility of 2024.1 and earlier
+        if "tablets" in scylla_yaml.experimental_features:
+            return True
+        if scylla_yaml.dict().get("enable_tablets"):
+            return True
+
+    return False
--- a/sdcm/utils/loader_utils.py
+++ b/sdcm/utils/loader_utils.py
@@ -98,8 +98,7 @@ class LoaderUtilsMixin:
     @cached_property
     def tablets_enabled(self):
         # is tablets feature enabled in Scylla configuration.
-        with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
-            return is_tablets_feature_enabled(session)
+        return is_tablets_feature_enabled(self.db_cluster.nodes[0])
 
     def _run_all_stress_cmds(self, stress_queue, params):
         stress_cmds = params['stress_cmd']
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5070,10 +5070,9 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
 
     def decommission(self, node: BaseNode, timeout: int | float = None) -> DataCenterTopologyRfControl | None:
         if not node._is_zero_token_node:
-            with node.parent_cluster.cql_connection_patient(node) as session:
-                if tablets_enabled := is_tablets_feature_enabled(session):
-                    dc_topology_rf_change = DataCenterTopologyRfControl(target_node=node)
-                    dc_topology_rf_change.decrease_keyspaces_rf()
+            if tablets_enabled := is_tablets_feature_enabled(node):
+                dc_topology_rf_change = DataCenterTopologyRfControl(target_node=node)
+                dc_topology_rf_change.decrease_keyspaces_rf()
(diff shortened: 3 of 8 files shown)


sdcm/utils/loader_utils.py
```diff
@@ -95,12 +93,6 @@ def assemble_and_run_all_stress_cmd_by_ks_names(self, stress_queue, stress_cmd,
             }
             self._run_all_stress_cmds(stress_queue, params)
 
-    @cached_property
-    def tablets_enabled(self):
```
pehala, line 99: I would change the logic here instead, I think we want to reuse it

pehala: In that case, rewrite that method and use it in here as well

sdcm/utils/features.py
```diff
@@ -75,8 +76,9 @@ def is_consistent_topology_changes_feature_enabled(session: Session) -> bool:
     return CONSISTENT_TOPOLOGY_CHANGES_FEATURE in get_enabled_features(session)
 
 
-def is_tablets_feature_enabled(session: Session) -> bool:
+def is_tablets_feature_enabled(node: BaseNode) -> bool:
```
pehala, line 79: Would it be possible to get any node from session instead?

sdcm/utils/features.py
```diff
@@ -75,8 +76,9 @@ def is_consistent_topology_changes_feature_enabled(session: Session) -> bool:
...
     """ Check whether tablets enabled
-    if you need from a specific node use `patient_exclusive_cql_connection` session
     """
-    return TABLETS_FEATURE in get_enabled_features(session)
+    with node.remote_scylla_yaml() as scylla_yaml:
+        enable_tablets = scylla_yaml.enable_tablets
```
pehala, line 83:
```
    with node.remote_scylla_yaml() as scylla_yaml:
        return scylla_yaml.enable_tablets
```

sdcm/utils/features.py
jsmolar: In current use cases, it is easy to get a node. Using session here would mean that I need to create a session first, then return node (which I already have).

sdcm/utils/loader_utils.py
```diff
@@ -117,7 +109,10 @@ def _run_all_stress_cmds(self, stress_queue, params):
             # Due to an issue with scylla & cassandra-stress - we need to create the counter table manually
             # also tablets doesn't yet support counters, so we skip the command and error about it
             if 'counter_' in stress_cmd:
-                if self.tablets_enabled and SkipPerIssues('scylladb/scylladb#18180', params=self.params):
+                with self.db_cluster.nodes[0].remote_scylla_yaml() as scylla_yaml:
```
pehala, line 112: Please use a `tablets_enabled` here as well

pehala (approved):

Lakshmipathi (dismissed):

cezarmoise:
Is the case where tablets are enabled globally but disabled on the test keyspace something to be considered?

soyacz:
@jsmolar there are plenty of unrelated formatting changes which blur the review process. Please try not doing that if not necessary.

sdcm/utils/features.py
```diff
@@ -75,8 +75,8 @@ def is_consistent_topology_changes_feature_enabled(session: Session) -> bool:
...
 
 
-def is_tablets_feature_enabled(session: Session) -> bool:
+def is_tablets_feature_enabled(node) -> bool:
     """ Check whether tablets enabled
-    if you need from a specific node use `patient_exclusive_cql_connection` session
     """
-    return TABLETS_FEATURE in get_enabled_features(session)
+    with node.remote_scylla_yaml() as scylla_yaml:
+        return scylla_yaml.enable_tablets
```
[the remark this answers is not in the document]
jsmolar: I added backward compatibility with 2024.1

soyacz (changes requested):
after reading more details in referenced PR's my understanding is:
1. we still have to query if tablets are enabled for 6.1/2 and 2024.1/2 - we didn't backport config param meaning there. I don't see the point of changing current implementation regards that.
2. `enable_tablets` config meaning changed only for dev/2025.1 - for this we should have a method/function to determine if tablets are enabled by default for new keyspaces (if we need it anywhere). So we should have new function: `is_tablet_default`.
3. for versions >= 2025.1 `is_tablets_feature_enabled` can just return `True` and skip verifying features.

soyacz (commented):
> @soyacz I agree with all those suggestions, but we need a hotfix first for testing 2025.1, as currently some nemesis are skipped even on Vnode runs because this check is now outdated. So I propose merging this (If it works ofc) and then immediately start working on a much better and comprehensive solution.

Ok.

@pehala please create issue that will take above considerations.
Actually, problem is deeper if we want to test both vnodes and tablets in one test. Basically, we should have 3 methods: `is_tablets_feature_enabled` (soon to be obsolete), `is_tablets_default` and `is_tablets_used` (still test may use tablets despite of tablets default set to `false`).

@jsmolar please let us know if the most recent push was tested

Lakshmipathi:
> but we need a hotfix first for testing 2025.1, as currently some nemesis are skipped even on Vnode runs because this check is now outdated. 

Ok, now I understand why this run fails https://jenkins.scylladb.com/job/scylla-staging/job/LakshmipathiGanapathi/job/longevity-counters-3h-test/14/console  with message `Tablets feature is enabled, skipping counter stress` even though tablets are _not_ enabled.

soyacz (approved):

soyacz:
> @pehala please create issue that will take above considerations.
> Actually, problem is deeper if we want to test both vnodes and tablets in one test. Basically, we should have 3 methods: `is_tablets_feature_enabled` (soon to be obsolete), `is_tablets_default` and `is_tablets_used` (still test may use tablets despite of tablets default set to `false`).

@pehala @jsmolar ^^ so we don't forget

jsmolar:
> > @pehala please create issue that will take above considerations.
> > Actually, problem is deeper if we want to test both vnodes and tablets in one test. Basically, we should have 3 methods: `is_tablets_feature_enabled` (soon to be obsolete), `is_tablets_default` and `is_tablets_used` (still test may use tablets despite of tablets default set to `false`).
> 
> @pehala @jsmolar ^^ so we don't forget

I created issue for this: https://github.com/scylladb/scylla-cluster-tests/issues/10168

PR #9996, opened 2025-02-05
fix(ToggleAuditNemesisSyslog): Remove enterprise-only restriction for audit feature

Lakshmipathi:
This patch allows audit feature tested across all Scylla versions.

### Testing

- [ ] https://argus.scylladb.com/tests/scylla-cluster-tests/c6bdd560-030a-4438-9c0e-9514db09a2be

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -5166,8 +5166,9 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             reduce categories by excluding DML and QUERY,
             verify DDL are logged in audit log correctly. Leaves audit log enabled this way.
         """
-        if not self.target_node.is_enterprise:
-            raise UnsupportedNemesis("Auditing feature is only supported by Scylla Enterprise")
+        if not self.target_node.is_enterprise and ComparableScyllaVersion(self.target_node.scylla_version) < "2025.1.0~dev":
+            raise UnsupportedNemesis(
+                "Auditing feature is only supported by Scylla Enterprise or non-enterprise versions greater than 6")
 
         if store == "syslog" and self._is_it_on_kubernetes():
             # generally syslog is not supported on K8S because of different log line format


pehala:
Interesting that the 2025.1 is not registered as enterprise, it should be (and according to other problems we have with it, it is)

sdcm/nemesis.py
```diff
@@ -5166,8 +5166,6 @@ def _disrupt_toggle_audit(self, store: AuditStore):
             reduce categories by excluding DML and QUERY,
             verify DDL are logged in audit log correctly. Leaves audit log enabled this way.
         """
-        if not self.target_node.is_enterprise:
```
pehala, line 5169: I would replace this with version check, if the major version is lower than 6 this is unsupported nemesis

soyacz:
> Interesting that the 2025.1 is not registered as enterprise, it should be (and according to other problems we have with it, it is)

Exactly, @Lakshmipathi can you point place where audit nemesis were skipped but should not?

Lakshmipathi:
> > Interesting that the 2025.1 is not registered as enterprise, it should be (and according to other problems we have with it, it is)
> 
> Exactly, @Lakshmipathi can you point place where audit nemesis were skipped but should not?

@soyacz https://github.com/scylladb/scylla-cluster-tests/blob/branch-2025.1/sdcm/nemesis.py#L5169 caused this jenkins job https://argus.scylladb.com/tests/scylla-cluster-tests/781c9243-92c0-4fe1-ac8a-291b86b04373 to skip audit nemesis

soyacz:
> > > Interesting that the 2025.1 is not registered as enterprise, it should be (and according to other problems we have with it, it is)
> > 
> > 
> > Exactly, @Lakshmipathi can you point place where audit nemesis were skipped but should not?
> 
> @soyacz https://github.com/scylladb/scylla-cluster-tests/blob/branch-2025.1/sdcm/nemesis.py#L5169 caused this jenkins job https://argus.scylladb.com/tests/scylla-cluster-tests/781c9243-92c0-4fe1-ac8a-291b86b04373 to skip audit nemesis

Ok, now I see. The problem is more generic than just Audit Nemesis.
I prepared generic fix for it: https://github.com/scylladb/scylla-cluster-tests/pull/10008

Lakshmipathi:
Closing this PR in-favour of generic fix: https://github.com/scylladb/scylla-cluster-tests/pull/10008

PR #10008, opened 2025-02-06
fix(BaseNode): fix is_enterprise property for 2025.1

soyacz:
We used to verify installed package if is enterprise to determine product.
With 2025.1 being not enterprise package it no longer works as expected.

Fix by comparing scylla version to 6.3 - all versions above are counted as enterprise: this includes all 202x.x versions.

### Testing

- [x] - tested locally truthiness of comparison
- [x] - provision test 
- [x] - [rocky9 artifact test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/artifacts-rocky9-test/11/) 

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -123,7 +123,7 @@ from sdcm.utils.version_utils import (
     get_gemini_version,
     get_systemd_version,
     ComparableScyllaVersion,
-    SCYLLA_VERSION_RE,
+    SCYLLA_VERSION_RE, is_enterprise,
 )
 from sdcm.utils.net import get_my_ip
 from sdcm.utils.node import build_node_api_command
@@ -787,7 +787,7 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
         return False
 
     def scylla_pkg(self):
-        return 'scylla-enterprise' if self.is_enterprise else 'scylla'
+        return 'scylla-enterprise' if self.is_product_enterprise else 'scylla'
 
     def file_exists(self, file_path: str) -> Optional[bool]:
         try:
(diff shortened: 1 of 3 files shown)


pehala (approved):

Lakshmipathi (dismissed):

roydahan:
> Fix by comparing scylla version to 6.3 - all versions above are counted as enterprise: this includes all 202x.x versions.

How "> 6.3" will include all 202x.x versions?

soyacz:
> > Fix by comparing scylla version to 6.3 - all versions above are counted as enterprise: this includes all 202x.x versions.
> 
> 
> 
> How "> 6.3" will include all 202x.x versions?

2000 is greater than 6 - first major versions are compared

roydahan (approved):

pehala (approved):

Lakshmipathi (dismissed):

sdcm/cluster.py
```diff
@@ -813,8 +813,12 @@ def start_network_interface(self, interface_name="eth1"):
             startup_interface_command = "ip link set {} up"
         self.remoter.sudo(startup_interface_command.format(interface_name))
 
+    @cached_property
```
vponomaryov, line 816: Will it serve as expected upgrading a node from some enterprise version like `2024.2.x` to OSS `2025.1.x` having the cache enabled?

fruch: I think @vponomaryov has a point here, we should cache here.

`scylla_version` is cached and handled already

soyacz: I left it as it was before. We cached version so this one caching makes no difference

fruch: it's being handled in:
```
    def forget_scylla_version(self) -> None:
        self.__dict__.pop("scylla_version_detailed", None)
        self.__dict__.pop("scylla_version", None)
        self.__dict__.pop('is_enterprise', None)
```

but I think we can remove the cache now that it's directly depends on scylla_version

soyacz: valid point, removed the cache from is_enterprise

vponomaryov (approved):
LGTM

pehala (approved):

PR #10013, opened 2025-02-07
fix(manager): update persistent snapshot used in Nemesis

mikliapko:
Closes https://github.com/scylladb/scylla-manager/issues/4233

Updated Manager snapshots that were based on rc versions of Scylla.

Old snapshots (based on 2024.2.0-rc1) contain system table which was introduced exclusively for this rc version and then rolled back in next version (rc2). As a result, restore to cluster with Scylla different from 2024.2.0-rc1 fails.

### Testing

- [x] [disrupt-mgr-restore](https://argus.scylladb.com/tests/scylla-cluster-tests/87a55c08-1a58-48cf-aa2c-dcc71ecd8141) with one of new snapshots 

### DIFF
--- a/defaults/manager_persistent_snapshots.yaml
+++ b/defaults/manager_persistent_snapshots.yaml
@@ -7,13 +7,13 @@ aws:
       expected_timeout: 1800  # 30 minutes
       snapshots:
         us-east-1:
-          sm_20240812150136UTC:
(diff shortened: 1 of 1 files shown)


mikliapko:
@scylladb/qa-maintainers 
Can we move forward with this PR?

soyacz (approved):

PR #10027, opened 2025-02-10
fix(cluster helper func): double quoting in keyspace names

timtimb0t:
When keyspace names start with a digit, they are returned with surrounding double quotes. This can lead to double quoting in CQL statements that can lead to SCT failure. This commit updates the get_test_keyspaces func to strip these extra quotes. If keyspace isnt quoted the change is harmless

refer: https://argus.scylladb.com/tests/scylla-cluster-tests/553b0e31-6b23-4f9b-9953-05fde9c09791

### Testing
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/eugene_test_folder/job/qutes/

### Reminders


temichus (dismissed):

sdcm/cluster.py
```diff
@@ -4491,7 +4491,7 @@ def get_test_keyspaces(self, db_node=None):
         """Function returning a list of non-system keyspaces (created by test)"""
         db_node = db_node or self.nodes[0]
         keyspaces = db_node.run_cqlsh("describe keyspaces").stdout.split()
-        return [ks for ks in keyspaces if not is_system_keyspace(ks)]
+        return [ks.strip('"') for ks in keyspaces if not is_system_keyspace(ks)]
```
fruch, line 4494: you still need to make sure all the usages are using `cql_quote_if_needed` as needed when using it in CQL statements

maybe it's time to have  `cql_unquote_if_needed` ? so it would be a bit more clear ?

soyacz, line 4494: we do it also in `nemesis.py` in line 2922, could be removed from there

soyacz: probably will cause failure for case in line `nemesis.py:4153`: need to add `cql_quote_if_needed`

soyacz: I'm not sure if we shouldn't add `cql_quote_if_needed` for `sdcm.cluster.BaseNode.get_cfstats` - requires verification which form fits it

soyacz (commented):
I verified places where it is used - and I think some places need to be fixed - see comments.

soyacz:
If always dropping quotes, please go again through all usages of `get_test_keyspaces` method and add 'quote_if_needed` (if required) or remove code that was dropping quotes (which now will be redundant).

timtimb0t:
@soyacz , could you please check again? there are few places in nemesis.py without usage  cql_unquote_if_needed

temichus (approved):

soyacz (approved):
LGTM

Issue #9861, opened 2025-01-20
AddRemoveDC nemesis should create keyspace with tablets disabled or change rf for new dc before decommission


aleksbykov:
With th latest schlla master AddRemoveDC nemesis start failing, because:
 * tablets enabled by default for every keyspace, and could be disabled during create/alter keyspace 
 * node could be decommissioned if rf of keyspace == number of nodes

AddRemoveDC nemesis first create new keyspace `keyspace_new_dc` without any options for tablets. So tablets are enabled for it.
and then create new dc with single node and alter the keyspace with rf=1 for new dc. This cause that decommission of new node failed because tablets enabled and rf = num of nodes in DC.

 More info https://github.com/scylladb/scylladb/issues/22261

AddRemoveDC should be fixed with:
  - create kyespace with disable tablets explicitly.
  - before decomission the new added node change rf=0 for new dc.

timtimb0t:
## Packages

Scylla version: `2025.2.0~dev-20250131.e1b1a2068a4d` with build-id `33145e2edb15d3b9de68da4b5144fe5f9774e65e`


Kernel Version: `6.8.0-1021-azure`


## Installation details

Cluster size: 4 nodes (Standard_L16s_v3)

Scylla Nodes used in this run:
- longevity-tls-1tb-7d-master-db-node-0b41f758-eastus-7 (null | 10.0.0.14) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-0b41f758-eastus-6 (null | 10.0.0.8) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-0b41f758-eastus-5 (null | 10.0.0.14) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-0b41f758-eastus-4 (null | 10.0.0.8) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-0b41f758-eastus-3 (null | 10.0.0.7) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-0b41f758-eastus-2 (null | 10.0.0.6) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-0b41f758-eastus-1 (null | 10.0.0.5) (shards: 14)


[...]

PR #10028, opened 2025-02-10
fix(nemesis): replication factor set to 0 before decommission

timtimb0t:
this commit changes keyspace replication factor to 0 before removing the DC to avoid scylla failures while decommissioning

refer: 9861

fix: #9861 

### Testing
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/eugene_test_folder/job/add_remove_dc/

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4849,8 +4849,13 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                         "Should have been already switched to NetworkStrategy"
                     strategy.replication_factors_per_dc.update({new_dc_name: 1})  # pylint: disable=protected-access
                     replication_strategy_setter(**{keyspace: strategy})
+
+                for key, preserved_strategy in replication_strategy_setter.preserved.items():
+                    preserved_strategy.replication_factors_per_dc[new_dc_name] = 0
+
                 InfoEvent(message='execute rebuild on new datacenter').publish()
-                with wait_for_log_lines(node=new_node, start_line_patterns=["rebuild.*started with keyspaces=", "Rebuild starts"],
+                with wait_for_log_lines(node=new_node,
+                                        start_line_patterns=["rebuild.*started with keyspaces=", "Rebuild starts"],
                                         end_line_patterns=["rebuild.*finished with keyspaces=", "Rebuild succeeded"],
                                         start_timeout=60, end_timeout=600):
                     new_node.run_nodetool(sub_cmd=f"rebuild -- {datacenters[0]}", long_running=True, retry=0)


defaults/docker_images/latte/values_latte.yaml
```diff
@@ -1,2 +1,2 @@
 latte:
-  image: scylladb/hydra-loaders:latte-0.28.1-scylladb
+  image: scylladb/hydra-loaders:latte-0.28.2-scylladb
```
aleksbykov, line 2: This should be in another PR

sdcm/nemesis.py
```diff
@@ -4860,6 +4865,13 @@ def finalizer(exc_type, *_):
                 datacenters = list(self.tester.db_cluster.get_nodetool_status().keys())
                 self._write_read_data_to_multi_dc_keyspace(datacenters)
 
+                update_replication_for_keyspaces(
+                    node=node,
+                    keyspaces=keyspaces_to_update,
+                    dc_name=new_dc_name,
+                    replication_factor=0
```
aleksbykov, line 4872: why not use setter here?

aleksbykov (dismissed):
Only update commit message, that it required with tablets enabled

soyacz:
Problem lies in different place - we discussed it before I think.
Recall:
See that `temporary_replication_strategy_setter` context manager brings back RF to value before adding node and changing RF:
```
< t:2025-01-09 15:03:06,248 f:common.py       l:1325 c:utils                p:DEBUG > Executing CQL 'ALTER KEYSPACE keyspace_new_dc WITH replication = {'class': 'NetworkTopologyStrategy', 'eu-west': 3}' ...
[...]
< t:2025-01-09 15:03:08,570 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 2025-01-09 15:03:08.568: (NodetoolEvent Severity.NORMAL) period_type=begin event_id=aae22fcf-4e23-4745-b1ae-fb73dfdeaffd: nodetool_command=decommission node=longevity-parallel-topology-schema--db-node-9ed004ce-6
[...] happened after alter keyspace
< t:2025-01-09 15:03:15,875 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > std::runtime_error (Decommission failed. See earlier errors (Rolled back: Failed to drain tablets: std::runtime_error (There are nodes with tablets to drain but no candidate nodes in DC eu-west_nemesis_dc. Consider adding new nodes or reducing replication factor.)). Request ID: d696cc5c-ce9a-11ef-cbbc-800cf5acb655)
```

So problem is that we don't wait for schema agreement after changing RF. And this is what recently must have changed in Scylla (nemesis worked so far)
So when developing a fix with that enhancement need also to fix [scylla docs about decommission dc](https://enterprise.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/decommissioning-data-center.html) to include a wait for schema agreement (this is my guess).
@pehala @bhalevy @aleksbykov @timtimb0t

sdcm/nemesis.py
```diff
@@ -4860,6 +4863,11 @@ def finalizer(exc_type, *_):
                 datacenters = list(self.tester.db_cluster.get_nodetool_status().keys())
                 self._write_read_data_to_multi_dc_keyspace(datacenters)
 
+                for keyspace in system_keyspaces + ["keyspace_new_dc"]:
+                    strategy = ReplicationStrategy.get(node, keyspace)
+                    strategy.replication_factors_per_dc.update({new_dc_name: 0})
```
temichus, line 4868: @aleksbykov @timtimb0t according to @soyacz comment and provided [link](https://enterprise.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/decommissioning-data-center.html) to doc
can we just pop the key 

```
strategy.replication_factors_per_dc.pop(new_dc_name)
```

soyacz: according to my comment, and logs provided, we already pop this by `temporary_replication_strategy_setter` upon exit. Just something changed that we need to wait some time (I don't know why, one idea is schema agreement, but it's just my guess).

pehala:
Given this fails on a few tests for 2025.1, how long will this take to complete? Could we expedite a quick fix or workaround?

timtimb0t:
> Given this fails on a few tests for 2025.1, how long will this take to complete? Could we expedite a quick fix or workaround?

I'll try to run the test today with "cheap" solution and update the review

soyacz:
@timtimb0t try with individual nemesis by cloning e.g. https://jenkins.scylladb.com/job/scylla-master/job/nemesis/job/longevity-5gb-1h-AddRemoveDcNemesis-azure-test/

fruch:
look at the issue leading to this PR

I think it's a scylla issue, that gonna be fixed in:
https://github.com/scylladb/scylladb/pull/22324

I don't we don't need to change anything in this nemesis, it's doing the procedure to the letter.
no need to disable tablets, and not need to manually set RF to zero.

if still we have issue with this specific nemesis and tablets, 
https://github.com/scylladb/scylladb/issues/22261 need to be reopen, I don't agree it's a test issue...

jsmolar:
Reproduced again on master after https://github.com/scylladb/scylladb/pull/22324 was promoted to master.
It seems that run https://argus.scylladb.com/tests/scylla-cluster-tests/8fd829a0-d43d-4ec6-8fa7-3a72ed8b138a is reproducing the same issue as this PR was meant to fix: `There are nodes with tablets to drain but no candidate nodes in DC us-east1_nemesis_dc. Consider adding new nodes or reducing replication factor.`

jsmolar:
Now I think this is correct fix to the issue, as it is stated that you need to set RF for DC to 0 in our docs:
`cqlsh> ALTER KEYSPACE <KEYSPACE_NAME> WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', '<DC_NAME1>' : 3, '<DC_NAME2>' : 3, '<DC_NAME3>' : 0};`
Which are not doing right now.

soyacz:
> Now I think this is correct fix to the issue, as it is stated that you need to set RF for DC to 0 in our docs: `cqlsh> ALTER KEYSPACE <KEYSPACE_NAME> WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', '<DC_NAME1>' : 3, '<DC_NAME2>' : 3, '<DC_NAME3>' : 0};` Which are not doing right now.

According to [docs](https://enterprise.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/decommissioning-data-center.html)
and my [comment](https://github.com/scylladb/scylla-cluster-tests/pull/10028#issuecomment-2659216366) 

All is done correctly.
Issue is somewhere else: whether in documentation (when changed, we need to adjust this nemesis to docs) or in scylla

jsmolar:
> > Now I think this is correct fix to the issue, as it is stated that you need to set RF for DC to 0 in our docs: `cqlsh> ALTER KEYSPACE <KEYSPACE_NAME> WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', '<DC_NAME1>' : 3, '<DC_NAME2>' : 3, '<DC_NAME3>' : 0};` Which are not doing right now.
> 
> According to [docs](https://enterprise.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/decommissioning-data-center.html) and my [comment](https://github.com/scylladb/scylla-cluster-tests/pull/10028#issuecomment-2659216366)
> 
> All is done correctly. Issue is somewhere else: whether in documentation (when changed, we need to adjust this nemesis to docs) or in scylla

According to [`master` version of doc](https://opensource.docs.scylladb.com/master/operating-scylla/procedures/cluster-management/decommissioning-data-center.html), simple pop is not enough. RF for DC needs to be set to 0 explicitly.

soyacz:
> > > Now I think this is correct fix to the issue, as it is stated that you need to set RF for DC to 0 in our docs: `cqlsh> ALTER KEYSPACE <KEYSPACE_NAME> WITH REPLICATION = {'class' : 'NetworkTopologyStrategy', '<DC_NAME1>' : 3, '<DC_NAME2>' : 3, '<DC_NAME3>' : 0};` Which are not doing right now.
> > 
> > 
> > According to [docs](https://enterprise.docs.scylladb.com/stable/operating-scylla/procedures/cluster-management/decommissioning-data-center.html) and my [comment](https://github.com/scylladb/scylla-cluster-tests/pull/10028#issuecomment-2659216366)
> > All is done correctly. Issue is somewhere else: whether in documentation (when changed, we need to adjust this nemesis to docs) or in scylla
> 
> According to [`master` version of doc](https://opensource.docs.scylladb.com/master/operating-scylla/procedures/cluster-management/decommissioning-data-center.html), simple pop is not enough. RF for DC needs to be set to 0 explicitly.

In that case, more appropriate fix would be to adjust `replication_strategy_setter` to handle case of added dc

jsmolar:
@soyacz Right now we are using `decommission` function for something that was not designed to do so. Yaron had a good idea, to implement a new function `remove_dc` that would handle such use case.

pehala:
> @soyacz Right now we are using `decommission` function for something that was not designed to do so. Yaron had a good idea, to implement a new function `remove_dc` that would handle such use case.

I think we can enhance the usage in a followup, but I would like to get this merge ASAP as it is showing a lot in our testing

@timtimb0t What is missing to finish this?

soyacz:
> @soyacz Right now we are using `decommission` function for something that was not designed to do so. Yaron had a good idea, to implement a new function `remove_dc` that would handle such use case.

I don't understand why we are using `decommission` in a wrong way. Can you elaborate?
AddRemoveDc nemesis was designed to follow closely documentation - this changed and nemesis needs adjustments - what @timtimb0t did seems doing exactly what was changed in the procedure.

soyacz:
> i discovered some new behavior in 2025.2.0~dev-0.20250304.e0f24c03e78c

> 
> So when we alter keyspace with original rf, it applied only for affected DC, second dc will stay as is

In that case we can drop idea of 'temporary replication setter'. Is this new or maybe old feature we missed?

jsmolar:
> > @soyacz Right now we are using `decommission` function for something that was not designed to do so. Yaron had a good idea, to implement a new function `remove_dc` that would handle such use case.
> 
> I don't understand why we are using `decommission` in a wrong way. Can you elaborate? AddRemoveDc nemesis was designed to follow closely documentation - this changed and nemesis needs adjustments - what @timtimb0t did seems doing exactly what was changed in the procedure.

Node decommission is different than dc decommission (not much, but logically they are). Our decommission functions are designed as such. I think, this is the main reason why we have this problem and all fixes so far are just workarounds.

soyacz:
> > > @soyacz Right now we are using `decommission` function for something that was not designed to do so. Yaron had a good idea, to implement a new function `remove_dc` that would handle such use case.
> > 
> > 
> > I don't understand why we are using `decommission` in a wrong way. Can you elaborate? AddRemoveDc nemesis was designed to follow closely documentation - this changed and nemesis needs adjustments - what @timtimb0t did seems doing exactly what was changed in the procedure.
> 
> Node decommission is different than dc decommission (not much, but logically they are). Our decommission functions are designed as such. I think, this is the main reason why we have this problem and all fixes so far are just workarounds. Example `decrease_keyspaces_rf` which is called from `decommission`:
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/9fa5f825ab8cde2d0fa24d0208b3291a528ee376/sdcm/utils/replication_strategy_utils.py#L229
> 
> 
> Skips reducing RF to 0 and we have to do it some other way.

Requirement for `decrease_keyspaces_rf` was introduced later (6.0) and is skipped for removing dc procedure as this procedure differs from down-scaling.
To be more precise we should have more procedures that are separated from 'decommission' like: 'remove_running_node', 'remove_unavailable_node' and 'remove_datacenter' 
@jsmolar you are welcome to address that in a proper way, but this is not part of this PR and referenced issue. This nemesis may be transformed to 'remove_datacenter' method (if it makes any difference - anyway it's the only place we do such operation).

PR #10040, opened 2025-02-10
fix(nemesis): decorator for ignoring mutation errors added

timtimb0t:
this commit adds decorator for ignoring mutation writes during remove_node_then_add_node nemesis

Ref: https://github.com/scylladb/scylladb/issues/21815

### Testing
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/eugene_test_folder/job/ignore_mutations/

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3702,7 +3702,13 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         assert removed_node_status is not None, "failed to get host_id using nodetool status"
         host_id = removed_node_status["host_id"]
 
-        with ignore_ycsb_connection_refused():
+        if SkipPerIssues('https://github.com/scylladb/scylladb/issues/21815', params=self.tester.params):
+            # TBD: To be removed after https://github.com/scylladb/scylladb/issues/21815 is resolved
+            ignore_stream_mutation_errors_due_to_issue = ignore_stream_mutation_fragments_errors
+        else:
+            ignore_stream_mutation_errors_due_to_issue = contextlib.nullcontext
+
+        with ignore_ycsb_connection_refused(), ignore_stream_mutation_errors_due_to_issue():
             # node stop and make sure its "DN"
             node_to_remove.stop_scylla_server(verify_up=True, verify_down=True)
 


temichus (dismissed):
according to issue https://github.com/scylladb/scylladb/issues/21815
not sure that ```this specific case (gate_closed_exception leading to mutation write)``` may happen only during ```disrupt_remove_node_then_add_node```. but I'm ok with merging this PR to reduce noise.

soyacz:
`ignore_stream_mutation_fragments_errors` has broader scope than linked issue. I don't think my opinion on that matter counts.

fruch:
> `ignore_stream_mutation_fragments_errors` has broader scope than linked issue. I don't think my opinion on that matter counts.

I would rather have skip, using SkipPerIssue with
https://github.com/scylladb/scylladb/issues/21815

so we'll have reference and automatically turn it back on, when scylla issue get resolves

timtimb0t:
@fruch , could you please take a look again?

sdcm/nemesis.py
```diff
@@ -3683,6 +3683,12 @@ def disrupt_remove_node_then_add_node(self):  # pylint: disable=too-many-branche
         if self._is_it_on_kubernetes():
             raise UnsupportedNemesis("On K8S nodes get removed differently. Skipping.")
 
+        if SkipPerIssues('https://github.com/scylladb/scylladb/issues/21815', params=self.tester.params):
+            # TBD: To be removed after https://github.com/scylladb/scylladb/issues/21815 is resolved
+            context_manager = ignore_stream_mutation_fragments_errors
+        else:
+            context_manager = contextlib.nullcontext()
```
fruch, line 3690: name might be a bit better, also you initialize this one twice, it won't work
```suggestion
            ignore_issues_cm = contextlib.nullcontext
```

fruch (changes requested):
nullcontext was initialize twice

sdcm/nemesis.py
```diff
@@ -3683,6 +3683,12 @@ def disrupt_remove_node_then_add_node(self):  # pylint: disable=too-many-branche
         if self._is_it_on_kubernetes():
             raise UnsupportedNemesis("On K8S nodes get removed differently. Skipping.")
 
+        if SkipPerIssues('https://github.com/scylladb/scylladb/issues/21815', params=self.tester.params):
+            # TBD: To be removed after https://github.com/scylladb/scylladb/issues/21815 is resolved
+            ignore_issues_cm = ignore_stream_mutation_fragments_errors
+        else:
+            ignore_issues_cm = contextlib.nullcontext
+
```
aleksbykov, line 3691: i suggest to use it before the context manager line number 711. it will be closer and clear to understand why this code is used.
Also `ignore_issues_cm` is not clear. better to rename `ignore(skip)_stream_mutation_errors_due_to_issue` or another meaningful name

aleksbykov: and fix pre-commits

temichus (approved):

temichus:
@fruch please rereview

PR #10015, opened 2025-02-08
ci(latte): bump latte version to 0.28.2-scylladb

vponomaryov:
It includes following main changes:
- Latencies are now calculated with regards to the coordinated omission.
- Added support for streaming the HDR histogram data in real time.
  Usage: `--hdrfile /path/to/file`

### Testing

- [ ]

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #10064, opened 2025-02-13
fix(latte): run latte docker container setting entrypoint explicitly

vponomaryov:
Recently we switched to the new latte docker image (https://github.com/scylladb/scylla-cluster-tests/pull/10015).
It is built from another Dockerfile which is stored in the latte repo.
And this new image has the latte binary as an entrypoint. Before it was bash.

So, set entrypoint and parameters for it explicitly in the latte code
to make latte container start and run for whole test as it is expected and was before.

### Testing

- [ ]

### Reminders


vponomaryov:
It should be backported everywhere where the https://github.com/scylladb/scylla-cluster-tests/pull/10015 went in.

fruch (approved):
LGTM

PR #10073, opened 2025-02-14
refactor(hdrh): support HDR histograms for multiple benchmarking tools

vponomaryov:
Current solution for HDR histograms assumes we use only `cassandra-stress`.
We are going to start using one more benchmarking tool that supports HDR histograms - `latte`.
So, refactor it's logic making the hdr tags be provided by the stress thread code which can detect it automatically.

With this refactor, we will be able to add HDR histograms support in SCT not only for `latte`,
but also for other benchmarking tools such as `scylla-bench` or `cql-stress-cassandra-stress`.

### Testing

- cassandra-stress, regression test: [scylla-staging/valerii/vp-perf-regression-latency-650gb-during-rolling-upgrade#22](https://argus.scylladb.com/tests/scylla-cluster-tests/f3841cb3-dd97-4d0d-b8b4-0cd65dd74a15)
- cassandra-stress, regression test: [scylla-staging/valerii/vp-scylla-master-perf-regression-latency-650gb-with-nemesis#3](https://argus.scylladb.com/tests/scylla-cluster-tests/26d8078d-592f-4802-b93b-464af87b6f8f)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4534,11 +4534,13 @@ class BaseScyllaCluster:  # pylint: disable=too-many-public-methods, too-many-in
                           key=key, threshold=size, keyspaces=keyspace, throw_exc=False)
 
     @optional_stage('nemesis')
-    def add_nemesis(self, nemesis, tester_obj):
+    def add_nemesis(self, nemesis, tester_obj, hdr_tags: list[str] = None):
         for nem in nemesis:
             nemesis_obj = nem['nemesis'](tester_obj=tester_obj,
                                          termination_event=self.nemesis_termination_event,
                                          nemesis_selector=nem['nemesis_selector'])
+            if hdr_tags:
+                nemesis_obj.hdr_tags = hdr_tags
             self.nemesis.append(nemesis_obj)
         self.nemesis_count = len(nemesis)
 
--- a/sdcm/scylla_bench_thread.py
+++ b/sdcm/scylla_bench_thread.py
@@ -137,6 +137,10 @@ class ScyllaBenchThread(DockerBasedStressThread):  # pylint: disable=too-many-in
         self.sb_mode: ScyllaBenchModes = ScyllaBenchModes(re.search(r"-mode=(.+?) ", stress_cmd).group(1))
         self.sb_workload: ScyllaBenchWorkloads = ScyllaBenchWorkloads(
             re.search(r"-workload=(.+?) ", stress_cmd).group(1))
+        # NOTE: scylla-bench doesn't have 'mixed' workload. Its hdr tag names are the same in all cases.
+        #       Another "raw" tag/metric is ignored because it is coordinated omission affected
+        #       and not really needed having 'coordinated omission fixed' latency one.
+        self.hdr_tags = ["co-fixed"]
 
     def verify_results(self):
         sb_summary = []
--- a/performance_regression_gradual_grow_throughput.py
+++ b/performance_regression_gradual_grow_throughput.py
@@ -185,7 +185,8 @@ class PerformanceRegressionPredefinedStepsTest(PerformanceRegressionTest):  # py
         for stress in stress_queue:
             results.extend(self.get_stress_results(queue=stress, store_results=False))
             self.log.debug("One c-s command results: %s", results[-1])
-        return results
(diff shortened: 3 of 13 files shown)


vponomaryov:
Tested here:

1) cassandra-stress, regression test: [scylla-staging/valerii/vp-perf-regression-latency-650gb-during-rolling-upgrade#22](https://argus.scylladb.com/tests/scylla-cluster-tests/f3841cb3-dd97-4d0d-b8b4-0cd65dd74a15)
Alive yet monitor node: http://13.61.33.96:3000/d/eed2gf29pswe8b/f9b42467-520e-50ec-ae71-5f27ae530fe4?orgId=1&from=1739560691052&to=1739569652379
2) cassandra-stress, regression test: [scylla-staging/valerii/vp-scylla-master-perf-regression-latency-650gb-with-nemesis#3](https://argus.scylladb.com/tests/scylla-cluster-tests/26d8078d-592f-4802-b93b-464af87b6f8f)
Alive yet monitor node: http://16.170.219.27:3000/d/ded2g95ttyhvkd/924175a0-ab2d-59db-ba11-05275fcea1bb?orgId=1&from=1739560708002&to=1739595752250

Reviewers, please look at the nemesis latencies. Whether it is ok or not.

sdcm/scylla_bench_thread.py
```diff
@@ -137,6 +137,10 @@ def __init__(self, stress_cmd, loader_set, timeout, node_list=None, round_robin=
         self.sb_mode: ScyllaBenchModes = ScyllaBenchModes(re.search(r"-mode=(.+?) ", stress_cmd).group(1))
         self.sb_workload: ScyllaBenchWorkloads = ScyllaBenchWorkloads(
             re.search(r"-workload=(.+?) ", stress_cmd).group(1))
+        # NOTE: scylla-bench doesn't have 'mixed' workload. Its hdr tag names are the same in all cases.
```
fruch, line 140: do you have jobs that enable HDR it for s-b ?

vponomaryov: No, we don't. Also, scylla-bench code doesn't use HDR file logger as of now.
I wrote it here just because addition of a support for scylla-bench becomes relatively easy - the same as upcoming latte PR.

fruch (approved):
LGTM

sdcm/cluster.py
```diff
@@ -4534,11 +4534,13 @@ def wait_total_space_used_per_node(self, size=None, keyspace='keyspace1'):
...
 
     @optional_stage('nemesis')
-    def add_nemesis(self, nemesis, tester_obj):
+    def add_nemesis(self, nemesis, tester_obj, hdr_tags: list[str] = None):
         for nem in nemesis:
             nemesis_obj = nem['nemesis'](tester_obj=tester_obj,
                                          termination_event=self.nemesis_termination_event,
                                          nemesis_selector=nem['nemesis_selector'])
+            if hdr_tags:
+                nemesis_obj.hdr_tags = hdr_tags
```
soyacz, line 4543: I know it's merged, but a bit quick one without leaving space for discussion. Anyway,

I'm not convinced nemesis object should have reference to hdr_tags.
When refactoring this, can we make it be more part of stress thread or group of stress threads (would need to introduce such object)?

vponomaryov: No, no one needs to add it.
Everything is done automatically.

vponomaryov: > still one need to add `, hdr_tags: list[str]` to arguments in every decorated function - why we cannot put hdr tags to `tester` object?

There are different sources of hdr_tags. Tester cannot know it always.
Stress thread knows it, and who runs the stress thread (there are cases when nemesis does it) must provide it further up to the HDR histograms handling code.
In different cases different stress commands get used - some use main stress, some use additional/nemesis one.

I chose this automatic approach instead of another technically suitable solution - adding one more SCT config option and requiring it to be set correctly all the time. 
The config option solution becomes really confusing because not only user defined stress commands may be measured but also ones which are part of a nemesis.

So, having current design of the SCT, it is the only stable solution.

The fact that I needed to rework half of the SCT to just add one more benchmarking tool support for HDR histograms tells us something about current SCT design.

PR #10099, opened 2025-02-17
test(performance regression): allow pre-creating the test keyspace

cezarmoise:
This would allow to set configurations options that cassandara-stress does not support like setting the initial number of tablets.
Similar to what the longevity test does.

- [X] I added the relevant `backport` labels

### Reminders

### DIFF
--- a/performance_regression_test.py
+++ b/performance_regression_test.py
@@ -22,6 +22,7 @@ from enum import Enum
 import yaml
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
+from sdcm.utils import loader_utils
 from upgrade_test import UpgradeTest
 from sdcm.tester import ClusterTester, teardown_on_exception
 from sdcm.sct_events import Severity
@@ -45,7 +46,7 @@ class PerformanceTestType(Enum):
     LATENCY = "latency"
 
 
-class PerformanceRegressionTest(ClusterTester):  # pylint: disable=too-many-public-methods
+class PerformanceRegressionTest(ClusterTester, loader_utils.LoaderUtilsMixin):  # pylint: disable=too-many-public-methods
 
     """
     Test Scylla performance regression with cassandra-stress.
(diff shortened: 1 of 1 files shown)


pehala (approved):

fruch:
@cezarmoise which test case is going to be using this ? 
i.e. I would rather see such a change coming it with the configuration using it.

also I would want to see it tested and working before merging it.

fruch:
and no point of changing this, if it won't get to the official branches for performance

pehala:
> and no point of changing this, if it won't get to the official branches for performance

We would like to use it for elasticity tests, which are derived from performance tests but are not in a specific branch, and in the future for testing performance of per-table hints, which we cannot do without this.

fruch:
> > and no point of changing this, if it won't get to the official branches for performance
> 
> We would like to use it for elasticity tests, which are derived from performance tests but are not in a specific branch, and in the future for testing performance of per-table hints, which we cannot do without this.

o.k., that's not something written in the description of this one.
anyhow I would to see it working before merging it

cezarmoise:
> > > and no point of changing this, if it won't get to the official branches for performance
> > 
> > 
> > We would like to use it for elasticity tests, which are derived from performance tests but are not in a specific branch, and in the future for testing performance of per-table hints, which we cannot do without this.
> 
> o.k., that's not something written in the description of this one. anyhow I would to see it working before merging it

Some runs with this
https://argus.scylladb.com/tests/scylla-cluster-tests/aca7b153-fa21-456e-a11e-0e41acfc4473
https://argus.scylladb.com/tests/scylla-cluster-tests/442ac766-d273-48bf-923d-a0c002824300
https://argus.scylladb.com/tests/scylla-cluster-tests/e0cd5744-57da-4f37-ba40-394550919fad


Initial tablets set to 1024
Screenshot 2025-03-04 151034

pehala (approved):

roydahan (approved):

soyacz (approved):
LGTM

PR #10118, opened 2025-02-18
fix(gemini): make gemini work without oracle passed

CodeLieutenant:
### Testing

Gemini needs to work without an oracle in upgrade tests, there was a bug which failed to run gemini since `Oracle Cluster` is `None`

- [x] :red_square: https://argus.scylladb.com/tests/scylla-cluster-tests/bc4f5bf5-f4d3-41db-bd42-22e1ddf2ef86 (Upgrade fails, but gemini works correctly without an oracle -> with this fix)

### DIFF
--- a/sdcm/gemini_thread.py
+++ b/sdcm/gemini_thread.py
@@ -18,6 +18,8 @@ import random
 import json
 import time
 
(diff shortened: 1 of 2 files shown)


fruch (approved):
LGTM

PR #10121, opened 2025-02-18
feature(ArtifactsTest): IOTune parameters validation

k0machi:
This commits adds a new subtest inside Artifacts test, that does the
check of machine image io_properties.yaml by comparing them to the
actual machine values and showing the deviation. The results are both
printed to console with deviation value and submitted to argus as a 
GenericResultTable.

Fixes scylladb/qa-tasks#1787

### Testing

- [x] [Jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/artifacts-ami-test/)
- [x] [Argus](https://argus.scylladb.com/workspace?state=WyI5MmMwYWZiZS03N2NiLTQ4ZjEtOWMzZi03YThlMzU3NDgzMWUiXQ)

### Reminders

### DIFF
--- a/artifacts_test.py
+++ b/artifacts_test.py
@@ -20,9 +20,11 @@ import json
 import yaml
 import requests
 
+from sdcm.argus_results import send_iotune_results_to_argus
 from sdcm.sct_events import Severity
 from sdcm.sct_events.database import ScyllaHousekeepingServiceEvent
 from sdcm.provision.helpers.certificate import c_s_transport_str
+from sdcm.sct_events.system import TestFrameworkEvent
 from sdcm.tester import ClusterTester
 from sdcm.utils.adaptive_timeouts import NodeLoadInfoServices
 from sdcm.utils.housekeeping import HousekeepingDB
@@ -30,6 +32,7 @@ from sdcm.utils.common import get_latest_scylla_release, ScyllaProduct
 from sdcm.utils.decorators import retrying
 from sdcm.utils.issues import SkipPerIssues
 from sdcm.utils.perftune_validator import PerftuneOutputChecker
+from sdcm.utils.validators.iotune import IOTuneValidator
 from utils.scylla_doctor import ScyllaDoctor
 
 STRESS_CMD: str = "/usr/bin/cassandra-stress"
@@ -324,6 +327,23 @@ class ArtifactsTest(ClusterTester):  # pylint: disable=too-many-public-methods
         if backend == "aws":
             with self.subTest("check ENA support"):
                 assert self.node.ena_support, "ENA support is not enabled"
+
+        if backend in ["gce", "aws", "azure"] and self.params.get("use_preinstalled_scylla"):
+            with self.subTest("check Scylla IO Params"):
+                try:
+                    validator = IOTuneValidator(self.node)
+                    validator.validate()
+                    send_iotune_results_to_argus(
+                        self.test_config.argus_client(),
+                        validator.results,
+                        self.node,
+                        self.params
+                    )
+                except Exception:  # pylint: disable=broad-except # noqa: BLE001
+                    self.log.error("IOTuneValidator failed", exc_info=True)
+                    TestFrameworkEvent(source={self.__class__.__name__},
+                                       message="Error during IOTune params validation.",
+                                       severity=Severity.ERROR).publish()
 
         with self.subTest("verify write cache for NVMe devices"):
             self.verify_nvme_write_cache()
--- a/sdcm/argus_results.py
+++ b/sdcm/argus_results.py
@@ -11,6 +11,7 @@
 #
 # Copyright (c) 2024 ScyllaDB
 import json
+import logging
 import time
 from datetime import timezone, datetime
 
@@ -20,6 +21,9 @@ from argus.client.generic_result import GenericResultTable, ColumnMetadata, Resu
 
 from sdcm.sct_events.event_counter import STALL_INTERVALS
 from sdcm.sct_events.system import FailedResultEvent
+
+
+LOGGER = logging.getLogger(__name__)
 
 LATENCY_ERROR_THRESHOLDS = {
     "replace_node": {
@@ -276,3 +280,55 @@ def send_manager_snapshot_details_to_argus(argus_client: ArgusClient, snapshot_d
     for key, value in snapshot_details.items():
         result_table.add_result(column=key, row="#1", value=value, status=Status.UNSET)
     submit_results_to_argus(argus_client, result_table)
+
+
+def send_iotune_results_to_argus(argus_client: ArgusClient, results: dict, node, params):
+    if not argus_client:
+        LOGGER.warning("Will not submit to argus - no client initialized")
+        return
+
+    class IOPropertiesResultsTable(GenericResultTable):
+        class Meta:
+            name = f"{params.get('cluster_backend')} - {node.db_node_instance_type} - Disk Performance"
+            description = "io_properties.yaml generated from live measured disk"
+            Columns = [
+                ColumnMetadata(name="read_iops", unit="iops", type=ResultType.INTEGER, higher_is_better=True),
+                ColumnMetadata(name="read_bandwidth", unit="bps", type=ResultType.INTEGER, higher_is_better=True),
+                ColumnMetadata(name="write_iops", unit="iops", type=ResultType.INTEGER, higher_is_better=True),
+                ColumnMetadata(name="write_bandwidth", unit="bps", type=ResultType.INTEGER, higher_is_better=True),
+            ]
+
+    class IOPropertiesDeviationResultsTable(GenericResultTable):
+        class Meta:
+            name = f"{params.get('cluster_backend')} - {node.db_node_instance_type} - Disk Performance Percent deviation"
+            description = "io_properties.yaml percent deviation from pre-configured disk"
+            Columns = [
+                ColumnMetadata(name="read_iops_pct_deviation", unit="%",
+                               type=ResultType.INTEGER, higher_is_better=False),
+                ColumnMetadata(name="read_bandwidth_pct_deviation", unit="%",
+                               type=ResultType.INTEGER, higher_is_better=False),
+                ColumnMetadata(name="write_iops_pct_deviation", unit="%",
+                               type=ResultType.INTEGER, higher_is_better=False),
+                ColumnMetadata(name="write_bandwidth_pct_deviation", unit="%",
+                               type=ResultType.INTEGER, higher_is_better=False),
+            ]
+
+            ValidationRules = {
+                "read_iops_pct_deviation": ValidationRule(fixed_limit=15),
+                "read_bandwidth_pct_deviation": ValidationRule(fixed_limit=15),
+                "write_iops_pct_deviation": ValidationRule(fixed_limit=15),
+                "write_bandwidth_pct_deviation": ValidationRule(fixed_limit=15),
+            }
+
+    table = IOPropertiesResultsTable()
+    for key, value in results["active"].items():
+        table.add_result(column=key, row="measured", value=value, status=Status.UNSET)
+        table.add_result(column=key, row="pre-configured", value=results["preset"][key], status=Status.UNSET)
+    submit_results_to_argus(argus_client, table)
+
+    table = IOPropertiesDeviationResultsTable()
+    for key, value in results["deviation_pct"].items():
+        table.add_result(column=f"{key}_pct_deviation", row="deviation_percent",
+                         value=value, status=Status.PASS if value < 15 else Status.WARNING)
+
+    submit_results_to_argus(argus_client, table)
--- a/sdcm/utils/validators/iotune.py
+++ b/sdcm/utils/validators/iotune.py
@@ -1,0 +1,101 @@
+from copy import deepcopy
+import logging
+from typing import TypedDict
+import yaml
+from sdcm.cluster import BaseNode
+from sdcm.remote.remote_file import remote_file
+from sdcm.sct_events import Severity
+from sdcm.sct_events.system import TestFrameworkEvent
+
+LOGGER = logging.getLogger(__name__)
+
+
+class IOTuneDiskParams(TypedDict):
+    mountpoint: str
+    read_iops: int
+    read_bandwidth: int
+    write_iops: int
+    write_bandwidth: int
+
+
+class IOProperties(TypedDict):
+    disks: list[IOTuneDiskParams]
+
+
+class IOTuneException(Exception):
+    pass
+
+
+class IOTuneValidator:
+    def __init__(self, node: BaseNode):
+        self.node = node
+        self.node_io_properties: IOProperties = {}
+        self.preset_io_properties: IOProperties = {}
+        self.results = {
+            "mountpoint": None,
+            "deviation": {},
+            "deviation_pct": {},
+            "active": {},
+            "preset": {},
+            "limits":  {},
+        }
+
+    def validate(self):
+        self._run_io_tune()
+        self._prepare_results_table()
+        self._format_results_to_console()
+
+    def _read_io_properties(self, io_props_path="/etc/scylla.d/io_properties.yaml") -> IOProperties:
+        with remote_file(self.node.remoter, io_props_path) as f:
+            return yaml.safe_load(f)
+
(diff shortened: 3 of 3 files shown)


k0machi:
Demo: 
image

```log
17:42:50  < t:2025-02-18 16:42:49,047 f:iotune.py       l:55   c:sdcm.utils.validators.iotune p:INFO  > Disk performance values validation - testing /var/lib/scylla
17:42:50  < t:2025-02-18 16:42:49,047 f:iotune.py       l:59   c:sdcm.utils.validators.iotune p:INFO  > [/var/lib/scylla] read_iops: 109519 (-0)
17:42:50  < t:2025-02-18 16:42:49,047 f:iotune.py       l:59   c:sdcm.utils.validators.iotune p:INFO  > [/var/lib/scylla] read_bandwidth: 806913408 (6)
17:42:50  < t:2025-02-18 16:42:49,047 f:iotune.py       l:59   c:sdcm.utils.validators.iotune p:INFO  > [/var/lib/scylla] write_iops: 59110 (-3)
17:42:50  < t:2025-02-18 16:42:49,047 f:iotune.py       l:59   c:sdcm.utils.validators.iotune p:INFO  > [/var/lib/scylla] write_bandwidth: 559601600 (-0)
17:42:50  < t:2025-02-18 16:42:50,119 f:artifacts_test.py l:280  c:ArtifactsTest        p:INFO  > Verify XFS mount options for /var/lib/scylla contain `discard'
```

sdcm/utils/validators/iotune.py
```diff
@@ -0,0 +1,102 @@
...
+        for disk in self.node_io_properties.get("disks"):
+            disk_copy = {**disk}
+            mountpoint = disk_copy.pop("mountpoint")
+            LOGGER.info("Disk performance values validation - testing %s", mountpoint)
+            for key, val in disk_copy.items():
+                preset_val = preset_disk.get(key)
+                diff = (val / preset_val - 1) * 100
+                LOGGER.info("[%s] %s: %s (%.0f)", mountpoint, key, val, diff)
+
+    def _submit_results_to_argus(self):
```
soyacz, line 61: this should be done outside of `IOTuneValidator` which should do only one thing: validating correctness of predefined io_properties against actual values: all things like submitting or printing in console could be done separately to not blur purpose of this class.

sdcm/utils/validators/iotune.py
```diff
@@ -0,0 +1,102 @@
...
+                    ColumnMetadata(name="write_iops", unit="iops", type=ResultType.INTEGER, higher_is_better=True),
+                    ColumnMetadata(name="write_bandwidth", unit="bps", type=ResultType.INTEGER, higher_is_better=True),
+                ]
+
+                ValidationRules = {
+                    "read_iops": ValidationRule(best_pct=10, fixed_limit=bottom_limit(preset_disk.get("read_iops"))),
+                    "read_bandwidth": ValidationRule(best_pct=10, fixed_limit=bottom_limit(preset_disk.get("read_iops"))),
+                    "write_iops": ValidationRule(best_pct=10, fixed_limit=bottom_limit(preset_disk.get("read_iops"))),
+                    "write_bandwidth": ValidationRule(best_pct=10, fixed_limit=bottom_limit(preset_disk.get("read_iops"))),
+                }
```
soyacz, line 90: I thing `best_pct` rule is not right.
I'd propose different approach: instead of sending raw metrics, send differences in absolute value. This way `higher_is_better` could be set to `False` and fixed limit (calculated by predefined io_properties.yaml metric * x%). This should work for any instance size.
In that case table name should be renamed to   f"{self.params.get('cluster_backend')} - {self.node.db_node_instance_type} IO Tune absolute deviation"
Drawback is only when we need to know exact value we would need to look into logs (could be tackled by publishing error event with full details in that case)

sdcm/utils/validators/iotune.py
```diff
@@ -0,0 +1,102 @@
...
+
+        tested_disk: IOTuneDiskParams = next(iter(self.node_io_properties.get("disks", [])))
+        tested_mountpoint = tested_disk.pop("mountpoint")
+        if tested_mountpoint != preset_disk["mountpoint"]:
+            LOGGER.warning("Disks differ - probably a mistake: %s vs %s, will not submit iotune results",
+                           tested_mountpoint, preset_disk["mountpoint"])
+            return
```
soyacz, line 97: I think we shouldn't fail silently here. Better raise error event

artifacts_test.py
```diff
@@ -325,6 +326,14 @@ def test_scylla_service(self):
             with self.subTest("check ENA support"):
                 assert self.node.ena_support, "ENA support is not enabled"
 
+        if ("gce" in backend or "aws" in backend or "azure" in backend):
```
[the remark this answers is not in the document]
k0machi: Good catch, I was testing and removed `use_preinstalled_scylla`, will put it back

sdcm/utils/validators/iotune.py
k0machi: okay, I'll move submission to the tester body.

sdcm/utils/validators/iotune.py
k0machi: We can send both I think, and that's what I was thinking as well - I'll look into doing it that way.

sdcm/utils/validators/iotune.py
k0machi: Will do

k0machi: Done

artifacts_test.py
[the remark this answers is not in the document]
k0machi: Done

sdcm/utils/validators/iotune.py
k0machi: Moved to argus_results, kept the console output in the validator

sdcm/utils/validators/iotune.py
k0machi: Done, I think. We might want to sent the baseline as well.

sdcm/argus_results.py
```diff
@@ -276,3 +281,77 @@ def send_manager_snapshot_details_to_argus(argus_client: ArgusClient, snapshot_d
...
+        LOGGER.warning("Unable to continue - node should have io_properties.yaml, but it doesn't.")
+        TestFrameworkEvent(source="send_iotune_results_to_argus",
+                           message="Unable to continue - node should have io_properties.yaml, but it doesn't.",
+                           severity=Severity.ERROR).publish()
+        return
+
+    class IOPropertiesResultsTable(GenericResultTable):
+        class Meta:
+            name = f"{params.get('cluster_backend')} - {node.db_node_instance_type} Disk Performance"
+            description = "io_properties.yaml comparison with live data"
```
soyacz, line 307: it's no longer comparison rather pure data from io_properties (row name should define 'image preset' or 'measured')

soyacz (changes requested):
`send_iotune_results_to_argus` function does validation and other unrelated to sending results to argus operations.
These should be part of validator and validator should prepare data for this function (this one should just send with minimum logic).

sdcm/argus_results.py
k0machi: Fixed

sdcm/argus_results.py
```diff
@@ -276,3 +281,77 @@ def send_manager_snapshot_details_to_argus(argus_client: ArgusClient, snapshot_d
...
+            Columns = [
+                ColumnMetadata(name="read_iops", unit="iops", type=ResultType.INTEGER, higher_is_better=True),
+                ColumnMetadata(name="read_bandwidth", unit="bps", type=ResultType.INTEGER, higher_is_better=True),
+                ColumnMetadata(name="write_iops", unit="iops", type=ResultType.INTEGER, higher_is_better=True),
+                ColumnMetadata(name="write_bandwidth", unit="bps", type=ResultType.INTEGER, higher_is_better=True),
+            ]
+
+            ValidationRules = {
```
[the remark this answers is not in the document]
k0machi: Fixed

sdcm/argus_results.py
```diff
@@ -276,3 +281,77 @@ def send_manager_snapshot_details_to_argus(argus_client: ArgusClient, snapshot_d
...
+    if not argus_client:
+        LOGGER.warning("Will not submit to argus - no client initialized")
+        return
+    preset_disk = next(iter(preset_io_props.get("disks", [])), None)
+    if not preset_disk:
+        LOGGER.warning("Unable to continue - node should have io_properties.yaml, but it doesn't.")
+        TestFrameworkEvent(source="send_iotune_results_to_argus",
+                           message="Unable to continue - node should have io_properties.yaml, but it doesn't.",
+                           severity=Severity.ERROR).publish()
+        return
```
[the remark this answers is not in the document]
k0machi: Fixed

k0machi:
> `send_iotune_results_to_argus` function does validation and other unrelated to sending results to argus operations. These should be part of validator and validator should prepare data for this function (this one should just send with minimum logic).

Done

sdcm/argus_results.py
```diff
@@ -276,3 +280,53 @@ def send_manager_snapshot_details_to_argus(argus_client: ArgusClient, snapshot_d
...
+            ValidationRules = {
+                "read_iops_abs_deviation": ValidationRule(fixed_limit=results["limits"].get("read_iops")),
+                "read_bandwidth_abs_deviation": ValidationRule(fixed_limit=results["limits"].get("read_bandwidth")),
+                "write_iops_abs_deviation": ValidationRule(fixed_limit=results["limits"].get("write_iops")),
+                "write_bandwidth_abs_deviation": ValidationRule(fixed_limit=results["limits"].get("write_bandwidth")),
+            }
+
+    table = IOPropertiesResultsTable()
+    for key, value in results["live"].items():
+        table.add_result(column=key, row="#1", value=value, status=Status.PASS)
```
soyacz, line 325: instead `#1` let's mark it as `actual` (btw. live also could be renamed to actual)

artifacts_test.py
```diff
@@ -325,6 +327,20 @@ def test_scylla_service(self):
...
+                    validator = IOTuneValidator(self.node)
+                    validator.validate()
+                    send_iotune_results_to_argus(
+                        self.test_config.argus_client(),
+                        validator.results,
+                        self.node,
+                        self.params
+                    )
+                except Exception:  # pylint: disable=broad-except # noqa: BLE001
+                    self.log.warning("IOTuneValidator failed", exc_info=True)
```
soyacz, line 342: let's raise Error event, otherwise we won't know about it

sdcm/argus_results.py
soyacz: status may be left as UNSET (so it's not green which we don't know)

sdcm/utils/validators/iotune.py
```diff
@@ -0,0 +1,102 @@
...
+            raise IOTuneException(
+                f"Disks differ - probably a mistake: {tested_mountpoint} vs {preset_disk['mountpoint']}")
+
+        for key, value in tested_disk.items():
+            deviation_val = abs(preset_disk.get(key) - value)
+            self.results["deviation"][key] = deviation_val
+            self.results["live"][key] = value
+            self.results["limits"][key] = self._bottom_limit(preset_disk[key])
+
+    def _format_results_to_console(self):
```
soyacz, line 91: why not basing on self.results?

soyacz (commented):
still few things to do, but looks overall quite good

sdcm/argus_results.py
k0machi: Done

artifacts_test.py
k0machi: Done

sdcm/utils/validators/iotune.py
k0machi: Could you elaborate? I'm not sure what's wrong here.

k0machi: Ah, I see now, yes, that could be fixed, before it was because it needed copies of the structures, now the structure it needs is done for it before, will fix.

k0machi: Fixed

soyacz: good,
that got me thinking, maybe instead sending deviation and showing limit in absolute value, how about send abs deviation in % and set fixed limit to 15%?
This way graph/results would be even more readable.

k0machi: Could be done, I could send it as a third table to have both abs and % values.

k0machi: Done

sdcm/argus_results.py
```diff
@@ -276,3 +280,53 @@ def send_manager_snapshot_details_to_argus(argus_client: ArgusClient, snapshot_d
...
+            description = "io_properties.yaml percent deviation from preset disk"
+            Columns = [
+                ColumnMetadata(name="read_iops_pct_deviation", unit="iops",
+                               type=ResultType.INTEGER, higher_is_better=False),
+                ColumnMetadata(name="read_bandwidth_pct_deviation", unit="bps",
+                               type=ResultType.INTEGER, higher_is_better=False),
+                ColumnMetadata(name="write_iops_pct_deviation", unit="iops",
+                               type=ResultType.INTEGER, higher_is_better=False),
+                ColumnMetadata(name="write_bandwidth_pct_deviation", unit="bps",
+                               type=ResultType.INTEGER, higher_is_better=False),
```
soyacz, line 313: change units here to `%`

soyacz (commented):
Small thing about units.
Besides looks good.

One more thing to decide is if 15% is the right error threshold - in original issue I see there's 10% mentioned but this number will cause artifact tests to fail and break pipelines (which we probably don't want to do)

sdcm/argus_results.py
k0machi: Done

k0machi:
@soyacz Did some adjustments  and renamed rows as well - looks better now in Argus.

k0machi:
> Small thing about units. Besides looks good.
> 
> One more thing to decide is if 15% is the right error threshold - in original issue I see there's 10% mentioned but this number will cause artifact tests to fail and break pipelines (which we probably don't want to do)

If you look at my latest runs there are results with well... 500% differences

soyacz:
> > Small thing about units. Besides looks good.
> > One more thing to decide is if 15% is the right error threshold - in original issue I see there's 10% mentioned but this number will cause artifact tests to fail and break pipelines (which we probably don't want to do)
> 
> If you look at my latest runs there are results with well... 500% difference - we might want to maybe disable limits for now or not fail the test somehow.

Yes, let's send results with status pass or warning (if crosses) instead of UNSET  - this way we won't fail. Let's keep it running for some time and then make decision if it should fail (after fixing issue with these numbers)

soyacz:
btw. deviation in % is calculated wrong: it should be  48.51% instead of 94% for write iops in https://argus.scylladb.com/tests/scylla-cluster-tests/6d7aee1b-b5fc-4ca5-9431-5be5a4ee3573

sdcm/utils/validators/iotune.py
```diff
@@ -0,0 +1,100 @@
...
+                               message=f"Disks differ - probably a mistake: {tested_mountpoint} vs {preset_disk['mountpoint']}",
+                               severity=Severity.ERROR).publish()
+            raise IOTuneException(
+                f"Disks differ - probably a mistake: {tested_mountpoint} vs {preset_disk['mountpoint']}")
+
+        self.results["mountpoint"] = tested_mountpoint
+        for key, value in tested_disk.items():
+            deviation_val = abs(preset_disk.get(key) - value)
+            self.results["deviation"][key] = deviation_val
+            self.results["deviation_pct"][key] = (deviation_val / value) * 100
```
soyacz, line 92: deviation should be calculated from `preset_disk.get(key)`

k0machi:
> btw. deviation in % is calculated wrong: it should be 48.51% instead of 94% for write iops in https://argus.scylladb.com/tests/scylla-cluster-tests/6d7aee1b-b5fc-4ca5-9431-5be5a4ee3573

It's correct, just from the other side, will change it to be relative to the preset value.

sdcm/utils/validators/iotune.py
k0machi: Fixed

PR #10124, opened 2025-02-19
improvement(pyproject.toml): Add Argus client to ignores for Ruff

pehala:
Currently, it is ignored in precommit only, so if running manually it will still complain.

### Testing

- [x] locally


fruch (approved):
LGTM

Issue #10119, opened 2025-02-18
wait_no_tablets_migration_running should not raise exception


soyacz:
In `add_new_nodes` method, in case of `wait_no_tablets_migration_running` fails to complete within 3600, method raises does not return added nodes. 
This is wrong as nodes were actually added and causes issue of not destroying them in grow-shrink scenario (and possibly other scenarios too).

PR #10125, opened 2025-02-19
fix(wait_no_tablets_migration_running): add soft timeout

soyacz:
`wait_no_tablets_migration_running` function was raising exception in case of timeout. This caused to miss info about added node and failing its decommission in tests.

Fix by applying soft timeout to this function (with hard timeout set much higer) and making it not raising exception, but just error event.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/10119

### Testing

- [x] - [grow-shrink nemesis test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/lukasz/job/longevity-5gb-1h-GrowShrinkClusterNemesis-aws-test/25/)

### Reminders

### DIFF
--- a/sdcm/utils/tablets/common.py
+++ b/sdcm/utils/tablets/common.py
@@ -4,6 +4,9 @@ from dataclasses import dataclass
 from typing import Optional
 
 from sdcm.rest.remote_curl_client import RemoteCurlClient
+from sdcm.sct_events import Severity
+from sdcm.sct_events.system import InfoEvent
+from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
 from sdcm.utils.features import is_tablets_feature_enabled
 
 LOGGER = logging.getLogger(__name__)
@@ -23,7 +26,7 @@ class TabletsConfiguration:
         return '{' + ', '.join(items) + '}'
 
 
-def wait_no_tablets_migration_running(node):
+def wait_no_tablets_migration_running(node, timeout: int = 3600):
     """
     Waiting for having no ongoing tablets topology operations using REST API.
     !!! It does not guarantee that tablets are balanced !!!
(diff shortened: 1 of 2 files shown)


pehala (approved):

sdcm/utils/tablets/common.py
```diff
@@ -40,8 +43,13 @@ def wait_no_tablets_migration_running(node):
...
-    LOGGER.info("All ongoing tablets topology operations are done")
+    try:
+        with adaptive_timeout(Operations.TABLET_MIGRATION, node, timeout=timeout):
+            for _ in range(3):
+                client.run_remoter_curl(method="POST", path="storage_service/quiesce_topology",
+                                        params={}, timeout=4*3600, retry=3)
+                time.sleep(5)
+        LOGGER.info("All ongoing tablets topology operations are done")
+    except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
+        InfoEvent(f"Failed to wait for having no ongoing tablets topology operations. Exception: {exc.__repr__()}",
```
fruch, line 54: isn't `adaptive_timeout` gonna raise an event if timing out or failing ?
if it doesn't raise events on failures, maybe it can extend to do so ?

sdcm/utils/tablets/common.py
```diff
@@ -40,8 +43,13 @@ def wait_no_tablets_migration_running(node):
...
-        client.run_remoter_curl(method="POST", path="storage_service/quiesce_topology",
-                                params={}, timeout=3600, retry=3)
-        time.sleep(5)
-    LOGGER.info("All ongoing tablets topology operations are done")
+    try:
+        with adaptive_timeout(Operations.TABLET_MIGRATION, node, timeout=timeout):
+            for _ in range(3):
+                client.run_remoter_curl(method="POST", path="storage_service/quiesce_topology",
+                                        params={}, timeout=4*3600, retry=3)
+                time.sleep(5)
```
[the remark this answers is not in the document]
soyacz: possibly loop is not needed, I'll remove it and retry testing

sdcm/utils/tablets/common.py
soyacz: adaptive timeout raises soft timeout event, but in case of hard timeout is reached (4*3600 in this case) it is going to be raised as exception - still we want to pass this case for future node cleanup but for other cases I wouldn't change that behavior.

fruch (approved):
LGTM

Issue #9998, opened 2025-02-05
Nemesis on Counter Tables cause test failures because unsupported actions


jsmolar:
Nemesis operations like `AddRemoveMv` and `ModifyTable` cause failures in the longevity-counters-3h-test. The failures occur because these Nemesis attempt to modify counter tables, which is not supported. Errors include messages such as `"Cannot set default_time_to_live on a table with counters"` and `"Materialized views are not supported on counter tables"`.

```
2025-02-05 08:17:12.830: (DisruptionEvent Severity.ERROR) period_type=end event_id=a6b63f4d-dc62-4619-ba8e-3069b33c7b9f duration=6s: nemesis_name=ModifyTable target_node=Node longevity-counters-3h-jsmolar2-db-node-db97e412-2 [44.205.22.166 | 10.12.0.231] errors=Error from server: code=2200 [Invalid query] message="Cannot set default_time_to_live on a table with counters"
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5501, in wrapper
result = method(*args[1:], **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3009, in disrupt_modify_table
disrupt_func()
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2748, in modify_table_compaction
self._modify_table_property(name="default_time_to_live", val=str(4300000))
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2173, in _modify_table_property
session.execute(cmd)
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 1318, in execute_verbose
return execute_orig(*args, **kwargs)
File "cassandra/cluster.py", line 2762, in cassandra.cluster.Session.execute
File "cassandra/cluster.py", line 5166, in cassandra.cluster.ResponseFuture.result
cassandra.InvalidRequest: Error from server: code=2200 [Invalid query] message="Cannot set default_time_to_live on a table with counters"
``` 

Occured in:
https://argus.scylladb.com/tests/scylla-cluster-tests/db97e412-b787-4ec6-9718-1f24a7cd2dad
https://argus.scylladb.com/tests/scylla-cluster-tests/fa9e65af-4920-48d3-a198-9c9614214abe

PR #9913, opened 2025-01-24
refactor(nemesis): modified compaction strategies related nemeses

timtimb0t:
modify_table_twcs_window_size nemesis modified to test twcs strategy in wider borders
modify_table_compaction nemesis modified for the same purposes.

Both collected in single commit due to modify_table_twcs_window_size nemesis depends on modify_table_compaction (ie modify_table_compaction nemesis adds randomness that should affect modify_table_twcs_window_size)

refs: 9361, 9362

fixes: #9361, #9362

### Testing
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/eugene_test_folder/job/twcs_cluster_without_changes/59/

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #10130, opened 2025-02-19
fix(modify_table_compaction): filter_out_table_with_counter is not filtering counter tables

jsmolar:
Implementation of `get_any_ks_cf_list` returned incorrect table names when
`filter_out_table_with_counter` was True. If such a table contained one
column whose type is not counter, this table would be added to the final
list and returned.

Fixes https://github.com/scylladb/scylla-cluster-tests/issues/9998

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/jsmolar/job/longevity-counters-3h-test/24/

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3709,6 +3709,7 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
 
         def execute_cmd(cql_session, entity_type):
             result = set()
+            counter_tables = set()
             is_column_type = entity_type == "column"
             column_names = regular_column_names
             if is_column_type:
@@ -3733,6 +3734,7 @@ class BaseCluster:  # pylint: disable=too-many-instance-attributes,too-many-publ
                     continue
 
                 if is_column_type and (filter_out_table_with_counter and "counter" in row.type):
+                    counter_tables.add(table_name)
                     continue
 
                 if filter_out_cdc_log_tables and getattr(row, column_names[1]).endswith(cdc.options.CDC_LOGTABLE_SUFFIX):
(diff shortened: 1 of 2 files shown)


timtimb0t:
> #9913 plays in the same area, @timtimb0t can you take a look and agree what should be done (maybe move this change to your's one?)

Sure, we discuss it in slack

sdcm/cluster.py
```diff
@@ -3733,6 +3734,7 @@ def execute_cmd(cql_session, entity_type):
                     continue
 
                 if is_column_type and (filter_out_table_with_counter and "counter" in row.type):
+                    counter_tables.add(table_name)
```
pehala, line 3737: why does this work for system and cdc tables but not counters?

jsmolar: Don't know why it does work for system or CDC.
One counter table in this case contains two columns:
Row(keyspace_name='scylla_bench', table_name='test_counters', type='bigint'), type: bigint
Row(keyspace_name='scylla_bench', table_name='test_counters', type='counter'), type: counter

The first one was the reason why `test_counters` table was not filtered.

jsmolar: With this schema:
```
CREATE TABLE scylla_bench.test_counters (
    pk bigint,
    ck bigint,
    c1 counter,
    c2 counter,
    c3 counter,
    c4 counter,
    c5 counter,
    PRIMARY KEY (pk, ck)
    ```

pehala (approved):

soyacz:
@timtimb0t did you agree with this change?

timtimb0t:
> @timtimb0t did you agree with this change?

We discussed this solution with @jsmolar in slack. Such workaround is necessary so that the nemeses do not affect each other

timtimb0t (approved):
LGTM

PR #10143, opened 2025-02-20
improvement(setup): enable parallel node operations by default

soyacz:
Feature is enabled for one longevity and perf elasticity case and we didn't face issues around that.

Enabling this feature by default will speed up cluster provisioning and make test scenarios where we add multiple nodes (`nemesis_add_node_cnt`>1) more appropriate for recent Scylla where we test doubling cluster as fast as possible.

### Testing

- [ ]

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -73,7 +73,7 @@ email_subject_postfix: ''
 collect_logs: false
 
 hinted_handoff: 'disabled'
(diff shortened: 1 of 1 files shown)


soyacz:
if approved we may consider pushing it to 2025.1 too

fruch (approved):
LGTM

vponomaryov (approved):
LGTM

Issue #10005, opened 2025-02-06
Nemesis `nodetool_decommission`.SCT picks up old cloud instance from a decommissined DB node for creation of a new one before it gets terminated


vponomaryov:
## Packages

Scylla version: `2025.1.0~rc1-20250202.28b889668011` with build-id `e46f69dbaac1a5eefd731a8d83206c0af4cbd7fa`
Kernel Version: `6.8.0-1021-aws`

## Issue description

Running the [scylla-2025.1/vnodes/longevity/longevity-user-profile-d-test#5](https://argus.scylladb.com/tests/scylla-cluster-tests/d43ebc50-217a-4394-84ad-2da313711413) CI job the `disrupt_nodetool_decommission` nemesis failed.

The decommission operation in this nemesis completed ok:

```
Feb 05 19:37:51.318923 longevity-user-profile-d-2025-1-db-node-d43ebc50-2 scylla[7919]:  [shard  0:strm] \
    storage_service - DECOMMISSIONING: done
```
```
2025-02-05 19:37:51,718 f:cluster.py      l:2666 c:sdcm.cluster_aws     p:DEBUG &gt; Node longevity-user-profile-d-2025-1-db-node-d43ebc50-2 [34.252.179.234 | 10.4.7.253] (rack: 1b): \
    Command '/usr/bin/nodetool -u cassandra -pw 'cassandra'  decommission ' duration -&gt; 194.15601649000018 s
```

But then the SCT found the existing yet cloud instance of the decommissioned DB node and tried to use it for the new node:
```
2025-02-05 19:38:25,277 f:cluster_aws.py  l:382  c:sdcm.cluster         p:INFO  &gt; Cluster longevity-user-profile-d-2025-1-db-cluster-d43ebc50 (AMI: ['ami-0cf9c4cdcacdd63c2'] Type: i3en.6xlarge): \
    Found provisioned instances = [ec2.Instance(id='i-0f5f0a59c54193023')]
...
2025-02-05 19:38:25,277 f:cluster_aws.py  l:526  c:sdcm.cluster_aws     p:DEBUG &gt; \
    Waiting until instance ec2.Instance(id='i-0f5f0a59c54193023') starts running...
```

Finally, it failed waiting for the `Running` state of the instance that was terminated:
```
[...]

PR #10144, opened 2025-02-20
fix(cluster_aws): make sure we are not picking terminated nodes

fruch:
Seems like in some situatation we might pick terminated nodes during test time, while we should have created a new node.

this commit introduce two things to help prevent it:
* wait for termination to full execute
* filter only running instances in the logic looking for instances

fixes: #10005

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster_aws.py
+++ b/sdcm/cluster_aws.py
@@ -336,7 +336,7 @@ class AWSCluster(cluster.BaseCluster):  # pylint: disable=too-many-instance-attr
         availability_zone = self.params.get('availability_zone').split(",")[az_idx] if az_idx is not None else None
         ec2 = ec2_client.EC2ClientWrapper(region_name=self.region_names[dc_idx],
                                           spot_max_price_percentage=self.params.get('spot_max_price'))
-        results = list_instances_aws(tags_dict={'TestId': test_id, 'NodeType': self.node_type},
+        results = list_instances_aws(tags_dict={'TestId': test_id, 'NodeType': self.node_type}, running=True,
                                      region_name=self.region_names[dc_idx], group_as_region=True, availability_zone=availability_zone)
         instances = results[self.region_names[dc_idx]]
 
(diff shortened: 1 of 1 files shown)


sdcm/cluster_aws.py
```diff
@@ -336,7 +336,7 @@ def _get_instances(self, dc_idx, az_idx=None):
         availability_zone = self.params.get('availability_zone').split(",")[az_idx] if az_idx is not None else None
         ec2 = ec2_client.EC2ClientWrapper(region_name=self.region_names[dc_idx],
                                           spot_max_price_percentage=self.params.get('spot_max_price'))
-        results = list_instances_aws(tags_dict={'TestId': test_id, 'NodeType': self.node_type},
+        results = list_instances_aws(tags_dict={'TestId': test_id, 'NodeType': self.node_type}, running=True,
```
soyacz, line 339: how much time it takes from provision to machine is running? If long enough we may have broken logic during test startup.

sdcm/cluster_aws.py
```diff
@@ -824,6 +824,7 @@ def destroy(self):
         self.stop_task_threads()
         self.wait_till_tasks_threads_are_stopped()
         self._instance.terminate()
+        self._instance.wait_until_terminated()
```
soyacz, line 827: this looks like a good enough solution

sdcm/cluster_aws.py
fruch: it's 1-2 minutes, in provision step we are waiting for it, and communicating with the nodes, AFAIK

sdcm/cluster_aws.py
fruch: it is not enough, since listing can get you stopped nodes (or all kind of other different states)

soyacz (approved):
LGTM
This should also fix reuse cluster scenarios when decommission nodes could be available in aws

sdcm/cluster_aws.py
fruch: or maybe just in azure, we are waiting for cloud-init (or that's in the test, not provision)

fruch: seems like we are not waiting on provision side at all...
I'll need to run a few test to see it in action

sdcm/cluster_aws.py
fruch: this one also have potential of changing nemesis duration, which might show up in performance test.

but it would be more the few minutes...

vponomaryov (approved):
LGTM

Issue #10131, opened 2025-02-19
`hydra-clean-test-resources` jenkins job fails and does not delete instances


cezarmoise:
Last good run was at be270716e9d8d372158ecc9860a0d0cc7bb141e5 

Traceback
https://jenkins.scylladb.com/view/QA/job/QA-tools/job/hydra-clean-test-resources/145/console
```
18:21:10  Traceback (most recent call last):
18:21:10    File "/tmp/jenkins/workspace/QA-tools/hydra-clean-test-resources/scylla-cluster-tests/./sct.py", line 1887, in <module>
18:21:10      cli.main(prog_name="hydra")
18:21:10    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
18:21:10      rv = self.invoke(ctx)
18:21:10    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
18:21:10      return _process_result(sub_ctx.command.invoke(sub_ctx))
18:21:10    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
18:21:10      return ctx.invoke(self.callback, **ctx.params)
18:21:10    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
18:21:10      return __callback(*args, **kwargs)
18:21:10    File "/usr/local/lib/python3.10/site-packages/click/decorators.py", line 26, in new_func
18:21:10      return f(get_current_context(), *args, **kwargs)
18:21:10    File "/tmp/jenkins/workspace/QA-tools/hydra-clean-test-resources/scylla-cluster-tests/./sct.py", line 336, in clean_resources
18:21:10      clean_func(param, dry_run=dry_run)
18:21:10    File "/tmp/jenkins/workspace/QA-tools/hydra-clean-test-resources/scylla-cluster-tests/sdcm/utils/resources_cleanup.py", line 91, in clean_cloud_resources
18:21:10      SCTCapacityReservation.get_cr_from_aws(config, force_fetch=True)
18:21:10    File "/tmp/jenkins/workspace/QA-tools/hydra-clean-test-resources/scylla-cluster-tests/sdcm/provision/aws/capacity_reservation.py", line 67, in get_cr_from_aws
18:21:10      ec2 = boto3.client('ec2', region_name=params.region_names[0])
18:21:10  IndexError: list index out of range
```

PR #10145, opened 2025-02-20
fix(cleanup): fix cleaning CRs when no region provided

soyacz:
In case region is not provided we should search all regions for active CR's for given test id instead of failing.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/10131

### Testing

- [X] - tested running locally (also without region)

### Reminders


cezarmoise (approved):

fruch (approved):
LGTM

Issue #10136, opened 2025-02-19
DataCenterTopologyRfControl doesn't work as expect in case of multi-dc


fruch:
## Packages

Scylla version: `2025.1.0~rc2-20250216.6ee17795783f` with build-id `8fc682bcfdf0a8cd9bc106a5ecaa68dce1c63ef6`

Kernel Version: `6.8.0-1021-aws`

## Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

`decrease_keyspaces_rf` logic isn't aware of multiple regions, and drop a whole region out of the replication:

```
< t:2025-02-18 08:59:19,352 f:replication_strategy_utils.py l:203  c:sdcm.utils.replication_strategy_utils p:DEBUG > Altering keyspace1 RF with: ALTER KEYSPACE keyspace1 WITH REPLICATION = { 'class' : 'NetworkTopologyStrategy', 'eu-westscylla_node_west':2 }                                                                                                               < t:2025-02-18 08:59:19,352 f:common.py       l:1317 c:utils                p:DEBUG > Executing CQL 'ALTER KEYSPACE keyspace1 WITH REPLICATION = { 'class' : 'NetworkTopologyStrategy', 'eu-westscylla_node_west':2 }' ...
```

this command can be stuck and timeout, regardless it's the wrong thing todo, and it would drop all the data from one region.

## Impact

all case using multiple regions, might fail on any nemesis that remove node first.

## How frequently does it reproduce?

100%

## Installation details

Cluster size: 6 nodes (i4i.xlarge)

Scylla Nodes used in this run:
- long-encrypt-at-rest-20gb-6h-multid-db-node-9f46998d-8 (35.176.16.52 | 10.3.2.177) (shards: -1)
- long-encrypt-at-rest-20gb-6h-multid-db-node-9f46998d-7 (34.253.210.248 | 10.4.3.232) (shards: 4)
- long-encrypt-at-rest-20gb-6h-multid-db-node-9f46998d-6 (18.171.61.165 | 10.3.0.203) (shards: 4)
- long-encrypt-at-rest-20gb-6h-multid-db-node-9f46998d-5 (18.171.210.169 | 10.3.2.90) (shards: 4)
- long-encrypt-at-rest-20gb-6h-multid-db-node-9f46998d-4 (3.8.115.132 | 10.3.1.92) (shards: 4)
[...]

PR #10153, opened 2025-02-20
fix(DataCenterTopologyRfControl): stop taking region to RF=0

fruch:
code in `_alter_keyspace_rf` was reffering only to one DC in the ALTER command, that lead to RF=0 on other DCs

this complication the situation for scylla, and surface multiple bug in operation that are talking place during this operation also is kind of change is not the intended behavier of this code

so now it would update the DC RF as needed, while leaving the other DCs RF the same as they were

Fix: #10136

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/replication_strategy_utils.py
+++ b/sdcm/utils/replication_strategy_utils.py
@@ -195,14 +195,14 @@ class DataCenterTopologyRfControl:
 
         return matching_keyspaces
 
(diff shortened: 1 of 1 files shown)


yarongilor (approved):
LGTM.
did it pass ok?
why backporting only to 2025.1?

soyacz (approved):
LGTM

Issue #10167, opened 2025-02-21
SCT fails to upload SCT runner log archives if those are splitted into several based on the size


vponomaryov:
At first CI job failed to gather logs: https://github.com/scylladb/scylla-cluster-tests/issues/9444
Then I ran the hydra log collaction command manually:
```
$ ./docker/env/hydra.sh --execute-on-runner 3.250.100.65 collect-logs --backend aws
There is scylladb/hydra:v1.88-python-driver-3.28.2 in local cache, using it.
SCT Runner IP: 3.250.100.65
...
Start collect logs for cluster db-cluster
Collecting logs on host: long-5000-tables-2025-1-db-node-acc227de-1
Collecting logs on host: long-5000-tables-2025-1-db-node-acc227de-2
Collecting logs on host: long-5000-tables-2025-1-db-node-acc227de-4
Collecting logs on host: long-5000-tables-2025-1-db-node-acc227de-3
Collecting logs on host: long-5000-tables-2025-1-db-node-acc227de-6
Collecting logs on host: long-5000-tables-2025-1-db-node-acc227de-5
<10.4.5.132>: Command rsync not available -- disabled
<10.4.4.215>: Command rsync not available -- disabled
<10.4.4.214>: Command rsync not available -- disabled
<10.4.4.27>: Command rsync not available -- disabled
...
Uploading 'db-cluster-acc227de.tar.zst' to https://cloudius-jenkins-test.s3.amazonaws.com/acc227de-a55c-4670-b803-ee9792cdfb04/20250221_120222/db-cluster-acc227de.tar.zst
Uploaded to https://cloudius-jenkins-test.s3.amazonaws.com/acc227de-a55c-4670-b803-ee9792cdfb04/20250221_120222/db-cluster-acc227de.tar.zst
Set public read access
collected data for db-cluster
['https://cloudius-jenkins-test.s3.amazonaws.com/acc227de-a55c-4670-b803-ee9792cdfb04/20250221_120222/db-cluster-acc227de.tar.zst']

Start collect logs for cluster sct-runner-events
Uploading 'sct-runner-events-acc227de.tar.zst' to https://cloudius-jenkins-test.s3.amazonaws.com/acc227de-a55c-4670-b803-ee9792cdfb04/20250221_120222/sct-runner-events-acc227de.tar.zst
[...]

PR #10172, opened 2025-02-23
fix(logcollector): introduce correct check to zstd only file

fruch:
check existing for tarball compressed with zstd, but not for zstd compressed file (as the case when splitting big sct log)

Fixes: #10167

### Testing

- [ ]

### Reminders


fruch:
@vponomaryov can you take it to 5000 tables case, if there a need to rerun it ?

PR #10188, opened 2025-02-24
test(default): set `tablets_initial_scale_factor` scylla config option

vponomaryov:
Make it be bigger than the current scylla default '1' to avoid load balance issues.

Ref: https://github.com/scylladb/scylladb/pull/22522

### Testing

- [ ]

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -1,3 +1,9 @@
+# NOTE: set the 'tablets_initial_scale_factor' explicitly while the following
+#       Scylla change is not merged and backported to older scylla branches.
+#       https://github.com/scylladb/scylladb/pull/22522
+append_scylla_yaml:
+  tablets_initial_scale_factor: 10
+
 db_type: "scylla"
 
 test_duration: 60
--- a/unit_tests/test_scylla_yaml_builders.py
+++ b/unit_tests/test_scylla_yaml_builders.py
@@ -538,7 +538,10 @@ class DummyNode(BaseNode):  # pylint: disable=abstract-method,too-many-instance-
             expected_node_yaml = expected_node_yaml.replace(
                 '__SEED_NODE_IPS__', seed_node_ips)
             expected_node_yaml = expected_node_yaml.replace('__NODE_IPV6_ADDRESS__', self.ipv6_ip_address)
-            assert json.loads(expected_node_yaml) == node_yaml.dict(
+            expected_node_yaml_dict = json.loads(expected_node_yaml)
+            if append_scylla_yaml := self.parent_cluster.params.get("append_scylla_yaml"):
+                expected_node_yaml_dict.update(append_scylla_yaml)
+            assert expected_node_yaml_dict == node_yaml.dict(
                 exclude_unset=True, exclude_defaults=True, exclude_none=True)
 
 


defaults/test_default.yaml
```diff
@@ -1,3 +1,6 @@
+append_scylla_yaml:
```
soyacz, line 1: add link to issue why we need that

soyacz (commented):
unit tests fail

tgrabiec:
> @bhalevy - is that a reasonable default?

This is the value which we're about to use in scylla by default, so yes.

defaults/test_default.yaml
vponomaryov: Done

pehala (changes requested):
This is incorrect way of achieving this, `tablets_initial_scale_factor` should not be used by users, it will also undergo a change with https://github.com/scylladb/scylladb/pull/22522. Initial option on keyspace should be used instead:
```
CREATE KEYSPACE excalibur
WITH replication = {
    'class': 'NetworkTopologyStrategy',
    'replication_factor': 3,
} AND tablets = {
    'initial': 2048
};
```

vponomaryov:
> > Users? It is used for setting up DB clusters.
> 
> ScyllaDB users should not use this option (according to developers), so we should not use it as well.

I think we understand different things under the "users" word.
We update scylla config in lots of places, because we need to test various things.
If you disagree to update scylla config and no one going to update all (really all) the keyspace definition places then what else we have?

> > Then, updating any table definition we create in tests is not applicable, In this case one should update enormous number of places and not only in SCT.
> 
> Other than SCT nothing is performance bound, missing this setting causes performance problems only. It is true that updating this properly in SCT would be a big change as we cannot do it as a default though.

The https://github.com/scylladb/scylla-qa-internal is `other` and it stores lots of schema and query configs used in our CI testing.
I filed several different bugs using longevities, not performance jobs.
Those longevities were using schema defined in that repo.

> > So, the solution must be only DB cluster config update, not per-ks/table.
> 
> It is per-ks/table now and it will be in the future (Although the initial will get replaced by other options and hopefully will work reasonably well out of the box).
> 
> If this is meant as a hotfix sure, we can do this like it temporarily, but this cannot stay as a permanent solution for this problem.

This is not hotfix, it is workaround for the Scylla bug.
And yes, it is needed at first, for testing `2025.1` rc, at second - any tablets testing.

defaults/test_default.yaml
soyacz: btw. I think this means initial tablets will be 2*2^10 isn't it? (see `configurations/tablets-initial-32.yaml` where 4 means 32 tablets).
This approach also is not taking into account number of shards, but for the sake of quick workaround might be good.

pehala:
> > > Users? It is used for setting up DB clusters.
> > 
> > 
> > ScyllaDB users should not use this option (according to developers), so we should not use it as well.
> 
> I think we understand different things under the "users" word. We update scylla config in lots of places, because we need to test various things. If you disagree to update scylla config and no one going to update all (really all) the keyspace definition places then what else we have?

I agree to update scylla config temporarily but I do not think it is right solution permanently. SCT is the closest thing we have to a real production clusters, we should use options that we recommend customers to use. Recommended way is NOT to set this config but use initial tablets option, thus we should use that. I understand it is not currently feasible, so we can use this method for now, but we need to change it, once the per-table hints + tablet scale factor patches drop and do this properly.


> > > So, the solution must be only DB cluster config update, not per-ks/table.
> > 
> > 
> > It is per-ks/table now and it will be in the future (Although the initial will get replaced by other options and hopefully will work reasonably well out of the box).
> > If this is meant as a hotfix sure, we can do this like it temporarily, but this cannot stay as a permanent solution for this problem.
> 
> This is not hotfix, it is workaround for the Scylla bug. And yes, it is needed at first, for testing `2025.1` rc, at second - any tablets testing.

This is required for 2025.1-rc as workaround, but it cannot be used for any tablets testing. If we find any performance issues with this option enabled they will be throwned out, due to invalid configuration. Only results with initial tablets are valid. 

TLDR: This is a bad way of configuring this, but we dont have a choice with current SCT setup. We need to treat it like a temporary workaround that will be removed as soon as we can and we definitely cannot use this for any performance measurements.

defaults/test_default.yaml
```diff
@@ -1,3 +1,9 @@
+# NOTE: set the 'tablets_initial_scale_factor' explicitly while the following
+#       Scylla change is not merged and backported to older scylla branches.
+#       https://github.com/scylladb/scylladb/pull/22522
```
pehala, line 3: Note: This will never not get backported to older branches, but once this drops, we should remove this either way

pehala (approved):

defaults/test_default.yaml
roydahan: @pehala we were explicitly told not to use this parameter, and use the initial_tablets inside the create keyspace command.
What has changed?

roydahan:
@pehala I was under the impression that we should not change this parameter and use only the CQL option (after your discussion with @bhalevy).
Now we transitioned back to use this default after running almost all tabelts tests?

pehala:
> @pehala I was under the impression that we should not change this parameter and use only the CQL option (after your discussion with @bhalevy). Now we transitioned back to use this default after running almost all tabelts tests?

That is true, this is meant as workaround for failing tests (https://github.com/scylladb/scylladb/pull/22522), since we cannot do globally CQL option, this is another thing. I would still prefer if we did CQL change only for tests that fail and not do this globally, but if this can unblock 2025.1 testing, I can accept it temporarily, but like I said in the previous comment, this global change has consequences on validity of our performance tests

defaults/test_default.yaml
soyacz: I just realized, this param changed its purpose... https://github.com/scylladb/scylladb/commit/f043c83ba539379411a7b7e48b964da649abcef9

fruch: and this change didn't yet reached 2025.1 branch.

fruch:
> Also good to mention, the new solution is targeted pretty soon, for 2025.1.1. So it might make more sense to do this hotfix only for 2025.1.0

we can't easily do like that in configuration, that would need to be in code.

I would suggest doing this one only to 2025.1 branch (not master), and revert it once we start 2025.1.1 testing.

roydahan:
I agree.
Let's do that ASAP, because any test without it is a waste of money

vponomaryov:
> > Also good to mention, the new solution is targeted pretty soon, for 2025.1.1. So it might make more sense to do this hotfix only for 2025.1.0
> 
> we can't easily do like that in configuration, that would need to be in code.
> 
> I would suggest doing this one only to 2025.1 branch (not master), and revert it once we start 2025.1.1 testing.

> I agree. Let's do that ASAP, because any test without it is a waste of money

- https://github.com/scylladb/scylla-cluster-tests/pull/10190

roydahan:
This one can be closed for now.

Issue #10169, opened 2025-02-23
[upgrades] image based upgrades are failing with tablets enabled, cause they are using Secondary indexes


fruch:
## Packages

Base Scylla version: `6.2.3-20250119.bff9ddde1283` with build-id `635cc6df4f69f507c809f50f3cce43de9bc205f0`
Target Scylla version (or git commit hash): `2025.2.0~dev-20250221.a6c882e4e37d` with build-id `9f913218f0d8d063b438a2506cce61776594fb64`

Kernel Version: `6.8.0-1021-aws`

## Issue description

- [x] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

upgrades from 6.2 to 2025.1 and above (with default tablets), are failing upgrades tests since Secondary indexes are not supported with tablets:
```
2025-02-22 07:27:19.088: (CassandraStressEvent Severity.CRITICAL) period_type=end event_id=a4ee812b-315c-4d6d-a95e-05c4a6b130b9 duration=6s: node=Node rolling-upgrade--ubuntu-focal-loader-node-4e99a2ca-1 [3.254.164.22 | 10.4.3.99]
stress_cmd=cassandra-stress user no-warmup profile=/tmp/complex_schema.yaml  ops'(read1=1,read3=1)' cl=ONE n=5000000 -mode cql3 native -rate threads=1000 -pop seq=1..5000000
errors:
Stress command completed with bad status 1: com.datastax.driver.core.exceptions.InvalidQueryException: Secondary indexes are not supported on base tables with tablets (keyspace 'keyspace_complex')
at com.datastax.driver.core.exceptions.InvalidQueryException.copy(InvalidQueryException.java:50)
```
## Impact

all image based upgrades are failing with this, so we can't yet tell if 2025.1 upgrades are all o.k.

## How frequently does it reproduce?

happens always

## Installation details
Cluster size: 4 nodes (im4gn.2xlarge)

Scylla Nodes used in this run:
- rolling-upgrade--ubuntu-focal-db-node-4e99a2ca-4 (18.203.234.238 | 10.4.2.100) (shards: 7)
- rolling-upgrade--ubuntu-focal-db-node-4e99a2ca-3 (34.251.239.24 | 10.4.0.81) (shards: 7)
- rolling-upgrade--ubuntu-focal-db-node-4e99a2ca-2 (34.254.100.226 | 10.4.3.92) (shards: 7)
[...]

PR #10193, opened 2025-02-25
fix(upgrade_tests): skip user profiles with index if tablets enabled

fruch:
recently that was new addition that was introduced that blocks creation of MV and SI if tablets enabled, so we must drop the usage of indexes in that case.

Fixes: #10169

### Testing

- [x] 🕥 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ami-test/19/

### Reminders

### DIFF
--- a/sdcm/fill_db_data.py
+++ b/sdcm/fill_db_data.py
@@ -3383,7 +3383,7 @@ class FillDatabaseData(ClusterTester):
             session.default_timeout = 60 * 5
             self.run_db_queries(session, session.default_fetch_size)
 
-    def paged_query(self, keyspace='keyspace_complex'):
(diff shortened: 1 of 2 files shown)


fruch:
replacing the complex profile wasn't working as expect, taking it out if tablets enabled

pehala (dismissed):

fruch:
missed some login hidden inside `paged_query()` 
fixing and trying again

fruch:
and now I've go the wrong keyspace, i.e. keyspace1 doesn't exist in the beginning of the test

pehala (approved):
The test passed, so LGTM

PR #10197, opened 2025-02-26
fix(auto backport): trigger with labeled event

yaronkaikov:
auto backport doesn't trigger when a PR is closed and the labels are added after. this is because `pull_request_target` was under a push event instead of its own

Fixing it


soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #10210, opened 2025-02-27
fix(manager): skip test_enospc_before_restore if tablets enabled

mikliapko:
Closes https://github.com/scylladb/scylla-manager/issues/4276

Temporarily skipping this tests since there is no clear understanding on how Manager should behave in such situation.
See details here https://github.com/scylladb/scylla-manager/issues/4276

Should be revisited after [issue](https://github.com/scylladb/scylla-manager/issues/4275) resolution.

### Testing

- [x] [sct-feature-test-backup](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/manager-master-clone/job/sct-feature-test-backup/2/)

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -46,6 +46,7 @@ from sdcm.cluster import TestConfig
 from sdcm.nemesis import MgmtRepair
 from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
 from sdcm.utils.common import reach_enospc_on_node, clean_enospc_on_node
+from sdcm.utils.features import is_tablets_feature_enabled
 from sdcm.utils.loader_utils import LoaderUtilsMixin
 from sdcm.utils.time_utils import ExecutionTimer
 from sdcm.sct_events.system import InfoEvent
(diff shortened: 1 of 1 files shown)


mgmt_cli_test.py
```diff
@@ -940,6 +941,12 @@ def test_enospc_during_backup(self):
         self.log.info('finishing test_enospc_during_backup')
 
     def test_enospc_before_restore(self):
+        if is_tablets_feature_enabled(self.db_cluster.nodes[0]):
+            # TODO: Get back to this restriction after https://github.com/scylladb/scylla-manager/issues/4275 resolution
+            self.log.info('Skipping test_enospc_before_restore due to enabled tablets. '
```
soyacz, line 946: why not using `SkipPerIssues`?

mikliapko: There is no clear understanding what would be the resolution of the issue I've attached.
Probably, we would need to rework this test for tablets totally, probably no.
There is a high level of uncertainty here. That's why I decided to not use `SkipPerIssues`.

soyacz (approved):
LGTM

mikliapko:
@scylladb/qa-maintainers Could someone else please take a look?

PR #10250, opened 2025-03-03
fix(wait_for_view_to_be_built) function did not return if only an ind…

mykaul:
…ex completed building

It seems that if we are checking that an only index has finished building (which will happen when we call the function from wait_for_index_to_be_built() function), then the code does add an INFO level log - but does not return.

Unrelated, added a more detailed log - of the timeout value.

### Testing

- [x] None

(Unsure if we need to backport this, perhaps)

### Reminders

### DIFF
--- a/sdcm/utils/nemesis_utils/indexes.py
+++ b/sdcm/utils/nemesis_utils/indexes.py
@@ -74,18 +74,19 @@ def wait_for_index_to_be_built(node: BaseNode, ks, index_name, timeout=300) -> N
 
 
 def wait_for_view_to_be_built(node: BaseNode, ks, view_name, timeout=300) -> None:
-    LOGGER.info('waiting for view/index %s to be built', view_name)
+    LOGGER.info(f"waiting {timeout} seconds for view/index {view_name} to be built")
     start_time = time.time()
     while time.time() - start_time < timeout:
         result = node.run_nodetool(f"viewbuildstatus {ks}.{view_name}",
                                    ignore_status=True, verbose=False, publish_event=False)
         if f"{ks}.{view_name}_index has finished building" in result.stdout:
             InfoEvent(message=f"Index {ks}.{view_name} was built").publish()
+            return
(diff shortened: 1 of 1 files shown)


mykaul:
No idea why CI failed (it never leaves a message), but CoPi is happy.

yarongilor (commented):
The fix looks good, i'm just not sure- does that mean the nemesis of disrupt_create_index always fail for this timeout? if not , why?

btw, i see this nemesis did not execute and pass for the last year:
image

mykaul:
> The fix looks good, i'm just not sure- does that mean the nemesis of disrupt_create_index always fail for this timeout? if not , why?
> 
> btw, i see this nemesis did not execute and pass for the last year: image

That was actually my question. Can we understand this better?

soyacz:
> The fix looks good, i'm just not sure- does that mean the nemesis of disrupt_create_index always fail for this timeout? if not , why?
> 

It didn't fail because it was processed by the other `if` (without `_index` suffix) - and this is because `wait_for_index_to_be_built` adds this suffix: basically, we never used this path. 
So another fix would be just dropping it as index suffix is handled by `sdcm.utils.nemesis_utils.indexes.wait_for_index_to_be_built`

mykaul:
@soyacz - apologies for dropping the ball on this - when you mentioned 'another fix would be' - did you mean an additional fix, or in this PR?

soyacz:
> @soyacz - apologies for dropping the ball on this - when you mentioned 'another fix would be' - did you mean an additional fix, or in this PR?

I just meant `if f"{ks}.{view_name}_index has finished building" in result.stdout:` is not used at all in SCT (that's why it never failed) - so instead of returning you can drop whole part. Could be done in this PR or in future or never (maybe some day we will use this path).

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

PR #10258, opened 2025-03-04
fix(DataCenterTopologyRfControl): decrease RF to 0 before decommission

jsmolar:
Alternative PR to https://github.com/scylladb/scylla-cluster-tests/pull/10028#issuecomment-2659216366
Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/9861

When `self.original_nodes_number` was 1 (which is in the case of decommission of single node in DC), the whole process is skipped. Which means no RF=0 for decommissioned data-center: https://opensource.docs.scylladb.com/master/operating-scylla/procedures/cluster-management/decommissioning-data-center.html

In addition `_get_original_nodes_number` was returning keyspaces where node was not present as two DCs had same `dc_idx` 

### Testing

- [ ] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/jsmolar/job/ongevity-5gb-1h-AddRemoveDcNemesis-aws-test/69/

### Reminders

### DIFF
--- a/sdcm/utils/replication_strategy_utils.py
+++ b/sdcm/utils/replication_strategy_utils.py
@@ -156,7 +156,7 @@ class DataCenterTopologyRfControl:
 
     def _get_original_nodes_number(self, node: 'BaseNode') -> int:
         # Get the original number of nodes in the data center
-        return len([n for n in self.cluster.data_nodes if n.dc_idx == node.dc_idx])
+        return len([n for n in self.cluster.data_nodes if n.datacenter == node.datacenter])
 
     def _get_keyspaces_to_decrease_rf(self, session) -> list:
         """
@@ -225,8 +225,8 @@ class DataCenterTopologyRfControl:
             replication-factor value.
         """
         node = self.target_node
-        # Ensure that nodes_num is 2 or greater
-        if self.original_nodes_number > 1:
+
+        if self.original_nodes_number > 0:
             with self.cluster.cql_connection_patient(node) as session:
                 decreased_rf_keyspaces = self._get_keyspaces_to_decrease_rf(session=session)
             if decreased_rf_keyspaces:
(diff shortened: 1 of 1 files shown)


fruch:
@yarongilor please review

yarongilor:
why 2 DCs had the same dc_idx? sounds like a different issue.

sdcm/utils/replication_strategy_utils.py
```diff
@@ -225,8 +225,8 @@ def decrease_keyspaces_rf(self):
             replication-factor value.
         """
         node = self.target_node
-        # Ensure that nodes_num is 2 or greater
-        if self.original_nodes_number > 1:
+
+        if self.original_nodes_number > 0:
```
yarongilor, line 229: 1. This won't work in case of a single DC with one node.
2. It would be a problem to remove a one-node DC if its keyspaces are not replicated in other DCs. It would probably fail altering keyspace.
3. That's the reason the for original condition request.
So it could be possible to have an additional check that other DCs exist and they replicate all keyspaces that are about to be removed from the dc-to-remove.
But it think it must be differentiated and much clearer as a different scenario and not just a simple node decommission.
IMO SCT better add a function like `remove_dc` which will take care and report for this special process of removing a whole DC, not just a node. of course it can then use this code as well.

yarongilor (changes requested):
this will fail in 2 scenarios. see in comment.

sdcm/utils/replication_strategy_utils.py
jsmolar: `remove_dc` is a great idea as we are currently trying to do that with tools (methods) that are not designed to do that.

soyacz:
It won't work as it does not touch AddRemoveDc nemesis parts. See update in https://github.com/scylladb/scylla-cluster-tests/pull/10028 - this should be working now

PR #10259, opened 2025-03-04
improvement(longevityPipeline): Remove default availability_zone setting

pehala:
There is no need to set default in Jenkins, test configuration files configure `availability_zone` as well and given that it is directly tied to the `db_nodes` setting, it makes sense to configure it only in configuration file and not have jenkins override the default in UI.

### Testing

- [ ] Test on Jenkins with various pipelines

### Reminders

### DIFF
--- a/vars/longevityPipeline.groovy
+++ b/vars/longevityPipeline.groovy
@@ -34,7 +34,7 @@ def call(Map pipelineParams) {
             string(defaultValue: "${pipelineParams.get('azure_region_name', 'eastus')}",
                    description: 'Azure location',
                    name: 'azure_region_name')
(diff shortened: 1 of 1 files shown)


pehala:
> Regardless it's now can still be override by environment variables, so people will need to get used to using that when needed

The option stays, you can still override it in Jenkins, it just wont be set to any default value there

fruch (dismissed):
LGTM

vponomaryov (commented):
What about other pipelines?
We have more than just "longevity"

pehala:
> What about other pipelines? We have more than just "longevity"

I will take a look at others, once I finish testing this change for longevitiy ones

fruch:
already mereged to master

PR #10112, opened 2025-02-18
test(backup): add native/rclone backup benchmarking under read/write stress

kreuzerkrieg:
RW latency during backup using rclone

Test requirements
[See here](https://docs.google.com/document/d/1mJ7n5hOP_ONNRaBFaIFgRGv2iX0JxyEiAHDf5fKSKK0/edit?tab=t.q2vqdtm13ij1)

Argus run
https://argus.scylladb.com/tests/scylla-cluster-tests/27b9e340-c851-4e74-ba37-d55e28efd8e0

|  | Size | Time [s] |
| --- | --- | --- |
| `rclone` based backup | 2.962TiB | 04:28:52 |
| `rclone` based restore | 17.766TiB | 06:27:55 |
| `rclone` based backup during R/W stress | 2.942TiB | 04:02:09 |

|  | read time [s] | op rate [op/s] | partition rate [pk/s] | row rate [row/s] | latency 99th percentile [ms] |
| --- | --- | --- | --- | --- | --- |
| Read stress: with concurrent `rclone` based backup | 01:00:38 | 86108.50 | 86108.50 | 86108.50 | 3.41 |
| Write stress: with concurrent `rclone` based backup | 01:00:38 | 99487.50 | 99487.50 | 99487.50 | 3.17 |
| Read stress: w/o concurrent backup | 01:00:37 | 84699.25 | 84699.25 | 84699.25 | 3.61 |
| Write stress: w/o concurrent backup | 01:00:37 | 99051.00 | 99051.00 | 99051.00 | 3.44 |

cc: @regevran

PR #10272, opened 2025-03-05
Native backup baseline

regevran:
Replaces: https://github.com/scylladb/scylla-cluster-tests/pull/10112


pehala:
Why create a new PR? I believe there are still many and many critical comments that were left unaddressed and now we lost the discussions.

Issue #10277, opened 2025-03-05
disrupt_enable_disable_table_encryption_aws_kms_provider_without_rotation fails pulling s-b docker image: "toomanyrequests: You have reached your unauthenticated pull rate limit"


yarongilor:
## Packages

Scylla version: `2024.2.5-20250221.cb9e2a54ae6d` with build-id `e65fa86462788ed829550cd131dd22e8c709b2bc`

Kernel Version: `5.15.0-1076-aws`

## Issue description

```
Nemesis Information
Class: Sisyphus
Name: disrupt_enable_disable_table_encryption_aws_kms_provider_without_rotation
Status: Failed
Failure reason
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5554, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4414, in disrupt_enable_disable_table_encryption_aws_kms_provider_without_rotation
    self._enable_disable_table_encryption(
  File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/group_common_events.py", line 519, in inner_func
    return func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/version_utils.py", line 724, in inner
    return mapped_func(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4510, in _enable_disable_table_encryption
    run_write_scylla_bench_load(write_cmd)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4499, in run_write_scylla_bench_load
    write_thread = self.tester.run_stress_thread(stress_cmd=write_cmd, stop_test_on_failure=False)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 2024, in run_stress_thread
    return self.run_stress_thread_bench(**params)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 2133, in run_stress_thread_bench
    bench_thread = ScyllaBenchThread(
  File "/home/ubuntu/scylla-cluster-tests/sdcm/scylla_bench_thread.py", line 121, in __init__
    super().__init__(loader_set=loader_set, stress_cmd=stress_cmd, timeout=timeout, stress_num=stress_num,
[...]

PR #10281, opened 2025-03-05
fix(docker_remote.py): Login docker-hub before pull, in case node authentication is expired

yarongilor:
In case running a docker stress like scylla-bench during a nemesis, The loader node initial docker-hub authentication may already expire. 
In order not to fail for docker-hub rate-limit, the loader node will be re-authenticated if needed.
Fixes: #10277

### Testing

- [ ]

### Reminders


yarongilor:
We can also consider maintaining a timestamp on all nodes for this authentication, but it might be a bit overkill.
It could be good enough if the function of `docker_hub_login` will check it before a docker pull and will log a message like:
```
    if match := re.search(r"^\s+Username: (.+)$", docker_info.stdout, re.MULTILINE):
        remoter.log.debug("Docker daemon is already logged in as `%s'.", match.group(1))
        return
```

pehala (approved):

PR #10289, opened 2025-03-06
fix(latte): detect latte keyspace correctly

vponomaryov:
In latte logic we use special class for handling of the detected keyspaces which is needed to share the data among several data exporters - common stress cmd results and HDR histograms data.

The problem was that having `self._value = ''` the condition for emptiness
was always giving us `False` because we were checking the class instance itself for emptiness.
So fix it by adding additional methods to make it behave like string more closely.

### Testing

- [scylla-staging/valerii/vp-longevity-gce-custom-d1-workload1-hybrid-raid#34](https://argus.scylladb.com/tests/scylla-cluster-tests/41569477-eb39-4a12-adcd-fdceafad6e4a)

### Reminders

### DIFF
--- a/sdcm/loader.py
+++ b/sdcm/loader.py
@@ -176,6 +176,18 @@ class LatteKeyspaceHolder:
     def __str__(self):
         return self._value
 
+    def __repr__(self):
+        return f"LatteKeyspaceHolder({self._value!r})"
+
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

Issue #10241, opened 2025-03-02
DC/Rack awareness in SCT


mykaul:
SCT tests should (by default) run with multiple racks - 3 in almost all cases (exceptions are 1 and 1 + virtual racks) and either a single DC, 2 DC + arbiter (AKA zero token node) or 3 DCs.

We need to enforce this standard setups as much as possible, across all tests.

mykaul:
Virtual racks - https://github.com/scylladb/scylla-cluster-tests/pull/10233

mykaul:
feature(multidc): scale cluster per dc in multidc config #10126 - https://github.com/scylladb/scylla-cluster-tests/pull/10126

pehala:
Milestones:
* Use Virtual racks in Sanity & Tier 1 tests
  * Monitor the tests in master for week or two
  * Sanity tests - https://github.com/scylladb/scylla-cluster-tests/pull/10233
* Set it as default in entire testsuite

pehala:
Encountered issue with SCT #10402, SCT tries to assign individual loaders to target one rack only. Solution is to add switch to disable this behaviour.

Tier 1 PR failed most likely on a similar isssue (#10307)

pehala:
Tier 1 tests are switched to using racks and it showcased a fundamental issue: with 3 racks we need to have 6 nodes total ( 2 in each rack) to allow headroom for disruptive nemesis, which could take nodes offline.

This will mean doubling the nodes for most tests, are you ok with this increases cost? @mykaul  @roydahan

soyacz:
> Tier 1 tests are switched to using racks and it showcased a fundamental issue: with 3 racks we need to have 6 nodes total ( 2 in each rack) to allow headroom for disruptive nemesis, which could take nodes offline.
> 
> This will mean doubling the nodes for most tests, are you ok with this increases cost? [@mykaul](https://github.com/mykaul) [@roydahan](https://github.com/roydahan)

maybe sizing down nodes when doubling?

pehala:
> maybe sizing down nodes when doubling?

We definitely can do that, good idea, it will be still be a cost increase, but much smaller

mykaul:
> Tier 1 tests are switched to using racks and it showcased a fundamental issue: with 3 racks we need to have 6 nodes total ( 2 in each rack) to allow headroom for disruptive nemesis, which could take nodes offline.
> 
[...]

Issue #9986, opened 2025-02-04
LoadAndStream Nemesis: Timeout occurred while waiting for start log line ['(?:storage_service|sstables_loader) - load_and_stream:']


pehala:
## Packages

Scylla version: `2025.1.0~rc0-20250128.f407799f252d` with build-id `dad9d694fca0e7a7b8403e6e6d1b47b9cf196a01`
Kernel Version: `6.8.0-1021-gcp`
Run: https://argus.scylladb.com/tests/scylla-cluster-tests/ef09ad89-dd8a-47f7-918a-f8d943b8870d

## Problem
`TimeoutError: Timeout occurred while waiting for start log line ['(?:storage_service|sstables_loader) - load_and_stream:'] on node: longevity-lwt-24h-2025-1-db-node-ef09ad89-0-10.`
Given streaming has been changed in this release with source available migration, I would suspect the logs changed as well

soyacz:
issue is with logs missing on sct side. Probably were dropped as just before the error there's a spike in logs (~25k/s, would need to recreate monitoring to see it actually happens and what was SCT cpu utilization)

fruch:
@soyacz would you able to take a close look, and see what is the cause of spike ?

soyacz:
I analyzed it, and cause of the spike was previous nemesis
during it there were plenty of messages related timeouts:
```
Feb 01 03:28:30.279128 longevity-lwt-24h-2025-1-db-node-ef09ad89-0-10 scylla[4116]:  [shard 3:sl:d] storage_proxy - Failed to apply mutation from 2255b3df-9c32-4512-80b3-1e852aeae1d2#3: exceptions::mutation_write_timeout_exception (Operation timed out for system.paxos - received only 0 responses from 1 CL=ONE.)
[... and like this:]
Feb 01 03:28:31.053179 longevity-lwt-24h-2025-1-db-node-ef09ad89-0-10 scylla[4116]:  [shard 0:sl:d] rpc - client 10.142.0.86:50267 msg_id 279072:  exception "Operation timed out for system.paxos - received only 0 responses from 1 CL=ONE." in no_wait handler ignored
```

and SCT couldn't process it quickly enough so processing spilled to next nemesis and eventually some messages were dropped ( despite were present in db log, were not present in sct.log).

Previous nemesis was `disrupt_remove_node_then_add_node` and finished within 10 minutes (quite fast for this... but maybe right if there's so little data in this test ~1GB per db node only)

[...]

PR #10307, opened 2025-03-06
improvement(tier1): Use virtual racks

pehala:
Needs to be retuned to increased number of nodes, which will be done after we see it in action.

* Increased number of nodes to multiples of 3
* Left out jobs which already have AZs in the configs
   * `longevity-multidc-schema-topology-changes-12h` and `longevity-50gb-3days` already have real AZs in their configuration files, followup PR will be made to run them with AZs weekly (I think they are currently reduced to a single AZ via trigger)

Part of #10241 
### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/ef3ff586-9498-4a55-a653-55e7fc3f3c01
    * Failed due to https://github.com/scylladb/scylla-cluster-tests/issues/9986, not caused by this PR, ran for 13hours

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
+++ b/test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
@@ -4,11 +4,12 @@ prepare_write_cmd: "cassandra-stress write cl=ALL n=50050075  -schema 'replicati
 stress_cmd: ["cassandra-stress write cl=QUORUM duration=11h  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=250  -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=400200300..600200300 -log interval=15"]
 stress_read_cmd: ["cassandra-stress read cl=ONE duration=11h -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=250  -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..50050075 -log interval=5"]
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "keyspace1.standard1", "interval": 5}']
-n_db_nodes: 4
+n_db_nodes: 6
 n_loaders: 1
 n_monitor_nodes: 1
+simulated_racks: 3
 
-instance_type_db: 'i4i.4xlarge'
+instance_type_db: 'i4i.2xlarge'
 gce_instance_type_db: 'n2-highmem-16'
 
 nemesis_class_name: 'SisyphusMonkey'
--- a/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
+++ b/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
@@ -12,10 +12,11 @@ stress_cmd: ["cassandra-stress mixed cl=QUORUM duration=6800m -schema 'replicati
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "random", "interval": 120}']
 round_robin: true
 
-n_db_nodes: 4
+n_db_nodes: 3
 n_loaders: 4
 n_monitor_nodes: 1
 seeds_num: 2
+simulated_racks: 3
 
 instance_type_db: 'i4i.4xlarge'
 gce_instance_type_db: 'n2-highmem-16'
--- a/test-cases/longevity/longevity-twcs-48h.yaml
+++ b/test-cases/longevity/longevity-twcs-48h.yaml
@@ -10,9 +10,10 @@ stress_read_cmd: [
     ]
 
 
-n_db_nodes: 4
+n_db_nodes: 3
 n_loaders: 3
 n_monitor_nodes: 1
+simulated_racks: 3
 
 instance_type_db: 'i3en.2xlarge'
 
--- a/test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
+++ b/test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
@@ -39,9 +39,10 @@ post_prepare_cql_cmds: ["CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELE
                         "ALTER TABLE scylla_bench.test with tombstone_gc = {'mode': 'repair', 'propagation_delay_in_seconds':'300'};"
                        ]
 
-n_db_nodes: 5
+n_db_nodes: 6
 n_loaders: 5 # Each loader will have 1 scylla-bench process at every step of the test (prepare, verify, stress & stress_read)
 n_monitor_nodes: 1
+simulated_racks: 3
 
 round_robin: true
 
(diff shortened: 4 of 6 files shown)


test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
```diff
@@ -4,9 +4,10 @@ prepare_write_cmd: "cassandra-stress write cl=ALL n=50050075  -schema 'replicati
 stress_cmd: ["cassandra-stress write cl=QUORUM duration=11h  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=250  -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=400200300..600200300 -log interval=15"]
 stress_read_cmd: ["cassandra-stress read cl=ONE duration=11h -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=250  -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..50050075 -log interval=5"]
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "keyspace1.standard1", "interval": 5}']
-n_db_nodes: 4
```
temichus, line 7: looks like 4 here for a reason. this test ```asymmetric cluster```
cc @juliayakovlev @fruch 


@pehala  does this change means that we are not supporting asymmetric with tablets

temichus (commented):
only one comment regarding asymmetric cluster testing 

Rather than that, I agree that the number of nodes and racks should always be multiples of RF during the test with tablets

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
fruch: the reason is same as all case we have 4 nodes, for make sure we can carry test is we lost one node (cause of SCT issue)

I don't see any other reason this specific case needed 4,

also the word asymmetric is a bit misleading, we have cause of asymmetric shards in each node, and this is not one of them

fruch: and yes seem like this one is use only on the asymmetric shards cases.
but they are all random per node, 3 or 4, it's not much of a difference

fruch (approved):
LGTM

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
roydahan: I think that by increasing it to 6, I would change the instance type to i4i.2xlarge.
Many/most of our customers use small instances because of this limitations and we should test with smaller instances as well.

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
```diff
@@ -16,6 +17,7 @@ nemesis_selector: ['limited']
 nemesis_seed: '026'
 nemesis_interval: 30
 nemesis_during_prepare: false
+nemesis_add_node_cnt: 3
```
roydahan, line 20: I think it's safe to already make it the default.

test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
```diff
@@ -12,10 +12,11 @@ stress_cmd: ["cassandra-stress mixed cl=QUORUM duration=6800m -schema 'replicati
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "random", "interval": 120}']
 round_robin: true
 
-n_db_nodes: 4
+n_db_nodes: 6
```
roydahan, line 15: Here we used 4 nodes because we wanted to make sure that 1TB dataset can fit into 4 x 4xlarge and whenever we reached out of space we knew something unexpected happened.

You can call it the old "90%" verification.
Increasing it to 6 without increasing the dataset will take some/most of the sting out of this test.
I don't have a good recommendation without testing it, but maybe now with the higher bar we should try it with 3 nodes.

test-cases/longevity/longevity-twcs-48h.yaml
```diff
@@ -10,9 +10,10 @@ stress_read_cmd: [
     ]
 
 
-n_db_nodes: 4
+n_db_nodes: 6
```
roydahan, line 13: I would try it also with 3 nodes instead of 4.

roydahan (changes requested):

test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
pehala: There were many improvements to disk utilization, so we can see if 3 nodes, gonna change this, thanks

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
pehala: https://github.com/scylladb/scylla-cluster-tests/pull/10345

pehala:
* Updated node counts

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
pehala: Changed the instances as well

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
roydahan: thx, this can be removed from here now.

roydahan (approved):
Just need to remove the new addition of "nemesis_add_node_cnt: 3" since it's a default now.

I would also run the 1tb case, for 12h to make sure it didn't break anything.

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey.yaml
pehala: Removed, thanks for the reminder

pehala:
v2:
  * Remove nemesis_add_node_cnt parameter, since https://github.com/scylladb/scylla-cluster-tests/pull/10345 is merged

temichus (approved):
LGTM

fruch:
> Just need to remove the new addition of "nemesis_add_node_cnt: 3" since it's a default now.
> 
> I would also run the 1tb case, for 12h to make sure it didn't break anything.

@pehala 

Did you run that ?

pehala:
> > @pehala
> > Did you run that ?
> 
> I did, but it failed due to stress failing, investigating https://argus.scylladb.com/tests/scylla-cluster-tests/11a0905f-fc52-428f-a48a-34d274f65a9b

@fruch Rerun on Azure https://argus.scylladb.com/tests/scylla-cluster-tests/11a0905f-fc52-428f-a48a-34d274f65a9b, it failed on
`Failed on java.io.IOException: Operation x0 on key(s) [324e4b31323639393230]: Data returned was not validated`. Not sure if that could be caused by this patch

roydahan:
It could be caused.
Need to check if it's due to timeout caused by overload or something else.

pehala:
Rebased and rerun test https://argus.scylladb.com/tests/scylla-cluster-tests/ef3ff586-9498-4a55-a653-55e7fc3f3c01, this one fail due to https://github.com/scylladb/scylla-cluster-tests/issues/9986, so I am pretty sure that is not caused by this

pehala:
@scylladb/qa-maintainers Please consider merging, it works now

roydahan (approved):

PR #10312, opened 2025-03-07
feature(events): introduce Severity.SUPRESS

soyacz:
A new severity level, that would help us stop generating some events we need those events for skipping some log line, so we never generate error or critical events from those.

until now we generated an event for each one, and now we'll be able to skip them since for some cases there are too many of those warnings.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

PR #10315, opened 2025-03-07
fix(tester): add traceback logging when checking regressions

cezarmoise:
If checking regression fails, this makes it easier to debug, especially errors related to the email templates.

For example, if a test does not have `use_hdrhistogram: true`, then the error would be 
> 2025-03-02 13:55:20.759: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=f0a49a5c-6891-43c3-8f99-85b5798e090f, source=PerformanceRegressionTest.check_regression() message=Failed to check regression
exception='NoneType' object is not iterable

which is not useful at all

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/1ca06f64-8d81-4da6-af33-73e14cafd2f7

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -21,6 +21,7 @@ import random
 import logging
 import os
 import re
+import sys
 import time
 import traceback
 import unittest
@@ -3345,6 +3346,7 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
             except Exception as exc:  # noqa: BLE001
                 TestFrameworkEvent(
                     message='Failed to check regression',
+                    trace=sys._getframe().f_back,
                     source=self.__class__.__name__,
                     source_method='check_regression',
                     exception=exc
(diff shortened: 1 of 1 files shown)


sdcm/tester.py
```diff
@@ -3390,6 +3392,7 @@ def check_regression_with_baseline(self, subtest_baseline):
                                                                     is_gce=is_gce,
                                                                     extra_jobs_to_compare=self.params.get('perf_extra_jobs_to_compare'))
         except Exception as exc:  # noqa: BLE001
+            self.log.exception("Error while checking regression")
             TestFrameworkEvent(
                 message='Failed to check regression',
                 source=self.__class__.__name__,
```
fruch, line 3398: you can add tracback to the event, I think it would be more useful, isn't it ? 
```suggestion
            TestFrameworkEvent(
                message='Failed to check regression',
                trace=sys._getframe().f_back,
                source=self.__class__.__name__,
```

cezarmoise: Yes, that is better.
Amended in https://github.com/scylladb/scylla-cluster-tests/commit/bbf0bcd8e8666824fe54f46eb31172c304fcd267

fruch (approved):
LGTM

PR #10345, opened 2025-03-10
improvement(test_default): Set nemesis_add_node_cnt to 3

pehala:
Related to #10241 

* Remove default values from testcases
   * No need to specify option that is the same as in default

### Testing

- [x] None


soyacz (commented):
I don't think setting 3 as default for `nemesis_add_node_cnt` while not all the tests use simulated racks or multi-az at all.

soyacz:
@temichus please review test cases where we use not 3/6 nodes but 4 or we don't use multiple az's/simulated racks. Also cases where we use RF=2 or 1 - I bet you find discrepancies that later will surface with test issues.

PR #10356, opened 2025-03-10
feature(hydra): action to build docker image automatically

fruch:
this new action can build a docker image when requirements are changed and push this new docker image information back into the PR

### Testing

- [x] add the simulated change to requirements with a version 1.0-something, to see it in action
- [x] test labels is needed

### Reminders

### DIFF
--- a/.github/workflows/build-docker-image.yaml
+++ b/.github/workflows/build-docker-image.yaml
@@ -1,0 +1,95 @@
+name: Build hydra image
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'docker/env/Dockerfile'
+      - 'docker/env/build_n_push.sh'
+      - 'requirements.in'
+      - 'requirements.txt'
+
+permissions:
+  actions: write
+  contents: write
+
+jobs:
+  check_org_membership:
+    runs-on: ubuntu-latest
+    outputs:
+      isTeamMember: ${{ steps.teamAffiliation.outputs.isTeamMember }}
+    steps:
+      - name: Check user for team affiliation
+        uses: tspascoal/get-user-teams-membership@v3
+        id: teamAffiliation
+        with:
+          GITHUB_TOKEN: ${{ secrets.AUTO_BACKPORT_TOKEN }}
+          username: ${{ github.actor }}
+          team: 'dev'
+
+  build_image:
+    needs: check_org_membership
+    if: needs.check_org_membership.outputs.isTeamMember == 'true' && contains(github.event.pull_request.labels.*.name, 'New Hydra Version')
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal access token.
+          fetch-depth: 0 # otherwise, there would be errors pushing refs to the destination repository.
+
+      - name: Check if docker image is already built
+        run: |
+          git fetch origin master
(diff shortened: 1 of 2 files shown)


.github/workflows/build-docker-image.yaml
```diff
@@ -0,0 +1,60 @@
+name: Build hydra image
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
```
fruch, line 6: TODO: make sure it would work base on label as well

.github/workflows/build-docker-image.yaml
```diff
@@ -0,0 +1,60 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
```
fruch, line 23: switch to python 3.10, and what we actually run

pehala:
Not sure if I like this approach, I do not like automatic edits of my own PR/branch, I would much rather if there was a simple way or a check that would warn me (maybe precommit?) that I need a new hydra image and easy way how to build & update it locally

fruch:
> Not sure if I like this approach, I do not like automatic edits of my own PR/branch, I would much rather if there was a simple way or a check that would warn me (maybe precommit?) that I need a new hydra image and easy way how to build & update it locally

* It means we need manual intervention for every automated PR, before we see CI run with it
* It limits it only to admins which have access to push images, people are not doing those
* It's gonna be guarded by a label, i.e. you can choose not to do it.

pehala:
> > Not sure if I like this approach, I do not like automatic edits of my own PR/branch, I would much rather if there was a simple way or a check that would warn me (maybe precommit?) that I need a new hydra image and easy way how to build & update it locally
> 
>     * It means we need manual intervention for every automated PR, before we see CI run with it
> 
>     * It limits it only to admins which have access to push images, people are not doing those
> 
>     * It's gonna be guarded by a label, i.e. you can choose not to do it.
Makes sense, I like the option to opt out

.github/workflows/build-docker-image.yaml
fruch: fixed

.github/workflows/build-docker-image.yaml
fruch: changed

.github/workflows/build-docker-image.yaml
```diff
@@ -0,0 +1,60 @@
...
+      - name: Generate a random name based on PR
+        run: |
+          docker_version=$(cut -d'-' -f1 ./docker/env/version | awk -F. '{print $1"."$2+1}')
+          sha=$(echo "${{ github.sha }}" | cut -c1-7)
+          echo "${docker_version}-PR${{ github.event.number }}-${sha}" > ./docker/env/version
+
+      - name: Log in to Docker Hub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
```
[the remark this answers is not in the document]
fruch: fixed

soyacz:
@fruch if someone works on a big PR with deps changes, each commit/amendment will cause to generate new image?

fruch:
> @fruch if someone works on a big PR with deps changes, each commit/amendment will cause to generate new image

if there a commit it generated, it should do it again
also if the "New Hydra Version" is removed, it won't try building

soyacz:
> > @fruch if someone works on a big PR with deps changes, each commit/amendment will cause to generate new image
> 
> if there a commit it generated, it should do it again also if the "New Hydra Version" is removed, it won't try building

so worth to split deps changing commit from other commits - we should mind that

PR #10358, opened 2025-03-11
improvement(treewide): Remove values from configurations, if same as default

pehala:
There is no need to redefine values which are defined in `test_defaults.yaml`.
This is a big general sweep of such cases, it is possible some values need to present regardless, let me know if you think some values should be kept.

### Testing

- [ ] None, this change should be safe as everything that was removed is in test_defaults.yaml

### DIFF
--- a/configurations/alternator/streams-with-nemesis.yaml
+++ b/configurations/alternator/streams-with-nemesis.yaml
@@ -1,4 +1,3 @@
 user_prefix: 'alternator-streams-nemesis'
 nemesis_class_name: 'DecommissionMonkey'
 nemesis_interval: 30
-nemesis_during_prepare: true
--- a/configurations/nemesis/additional_configs/docker_backend.yaml
+++ b/configurations/nemesis/additional_configs/docker_backend.yaml
@@ -17,8 +17,6 @@ n_monitor_nodes: 1
 seeds_num: 3
 
 nemesis_interval: 3
-nemesis_filter_seeds: false
-nemesis_during_prepare: true
 
 # Instance type for the case when tests are executed on Docker backend in AWS
 instance_type_runner: 'i4i.xlarge'
--- a/configurations/nemesis/longevity-5gb-1h-nemesis.yaml
+++ b/configurations/nemesis/longevity-5gb-1h-nemesis.yaml
@@ -23,8 +23,6 @@ instance_type_loader: 'c5.large'
 azure_instance_type_loader: 'Standard_F2s_v2'
 
 nemesis_interval: 3
-nemesis_filter_seeds: false
-nemesis_during_prepare: true
 
 gce_n_local_ssd_disk_db: 1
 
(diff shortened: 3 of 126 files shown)


configurations/operator/1tb-7days-no-tls.yaml
```diff
@@ -3,8 +3,8 @@ nemesis_selector: ['kubernetes']
 k8s_minio_storage_size: '2048Gi'
 
 # NOTE: operator doesn't support encryption as of January 2022
```
mikliapko, line 5: Perhaps this comment doesn't make sense anymore if you are removing `server_encrypt` and `client_encrypt`?

configurations/operator/200GB-no-tls.yaml
```diff
@@ -4,8 +4,8 @@ nemesis_selector: ['kubernetes']
 k8s_minio_storage_size: '4608Gi'
 
 # NOTE: operator doesn't support encryption as of January 2022
```
mikliapko, line 6: The same here

configurations/operator/50GB-no-tls.yaml
```diff
@@ -5,8 +5,8 @@ nemesis_add_node_cnt: 1
 k8s_minio_storage_size: '1024Gi'
 
 # NOTE: operator doesn't support encryption as of January 2022
```
mikliapko, line 7: The same here

configurations/rust/cql-stress-100gb-4h.yaml
```diff
@@ -5,5 +5,3 @@ stress_cmd: ["cql-stress-cassandra-stress write cl=QUORUM duration=240m -schema
              ]
 
 # Encryption is not yet supported by cql-stress.
```
mikliapko: The same here

configurations/operator/1tb-7days-no-tls.yaml
```diff
@@ -3,8 +3,8 @@ nemesis_selector: ['kubernetes']
...
-server_encrypt: false
-client_encrypt: false
+
+
```
mikliapko: In every config yaml parameters
```
server_encrypt: false
client_encrypt: false
```
were replaced with two empty lines.

Do you need this empty files?

mikliapko (commented):
nit: in many files the parameters were replaced with empty line(s), like, 
```
- server_encrypt: false
- client_encrypt: false
+
+
```
or 
```
- print_kernel_callstack: true
+
```

Is there any idea behind this decision or it was not by purpose?

configurations/operator/1tb-7days-no-tls.yaml
pehala: I think this was added by precommit, will check

configurations/operator/1tb-7days-no-tls.yaml
pehala: I am not sure, does manager support encryption now?

mikliapko: @grzywin may know something about it

pehala:
v2:
  * Remove unnecessary empty lines

Issue #10308, opened 2025-03-06
SCT doesn't save nemesis info into ES at all: Elasticsearch.index() got an unexpected keyword argument 'doc_type'


vponomaryov:
Argus:
- [scylla-staging/valerii/vp-longevity-gce-custom-d1-workload1-hybrid-raid#34](https://argus.scylladb.com/tests/scylla-cluster-tests/41569477-eb39-4a12-adcd-fdceafad6e4a)
- [scylla-master/longevity/longevity-100gb-4h-test#1068](https://argus.scylladb.com/tests/scylla-cluster-tests/74d0d08a-8c10-4e4e-9fa7-bc5d48c6a943)
- [scylla-master/longevity/longevity-100gb-4h-test#1067](https://argus.scylladb.com/tests/scylla-cluster-tests/b8abdd0d-75d6-4339-b3c9-72c56587da50)
- [scylla-master/longevity/longevity-100gb-4h-test#1066](https://argus.scylladb.com/tests/scylla-cluster-tests/97a736ad-0beb-4c94-9150-4c9024776725)

In the end of a nemesis SCT fails to update Elastic search service:
```
21:11:33  < t:2025-03-05 19:11:33,009 f:nemesis.py      l:5604 c:sdcm.nemesis         p:WARNING > \
    sdcm.nemesis.SisyphusMonkey: Unexpected error when attempting to update elasticsearch statistics:
21:11:33  < t:2025-03-05 19:11:33,009 f:nemesis.py      l:5604 c:sdcm.nemesis         p:WARNING > \
    Elasticsearch.index() got an unexpected keyword argument 'doc_type'
21:11:33  < t:2025-03-05 19:11:33,010 f:nemesis.py      l:5606 c:sdcm.nemesis         p:INFO  > \
    sdcm.nemesis.SisyphusMonkey: log_info: {'start': 1741201354, 'end': 1741201892, 'duration': 538, 'node': 'Node long-custom-d1-wrkld1-dev-db-node-41569477-0-7 [35.196.131.27 | 10.142.0.53]', 'subtype': 'end'}
```

It happens for each of the nemesis.
So, looks like we do not update ES with the nemesis data anymore.

The problem started appearing on the 4th of march early in the morning.
So, something broken got merged during the March 3 because early morning March 3 the problem was absent.

I suspect the root cause is the update of the elastic search library version in this PR:
- https://github.com/scylladb/scylla-cluster-tests/pull/9718

[...]

PR #10360, opened 2025-03-11
fix(elastic): remove usage of doc_type

fruch:
this argument was removed in v7 of the elasticsearch client we shouldn't use it anymore

fixes: #10308

### Testing

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/scylla-enterprise-perf-regression-latency-650gb-during-rolling-upgrade/6/

### Reminders


soyacz:
@fruch how was it used in ES? don't we rely filtering based on it somewhere?

fruch:
> @fruch how was it used in ES? don't we rely filtering based on it somewhere?

actually not, we weren't using different variant on the same document id.
so seems like it's work just fine without it.

I'm testing it now with one of the performance tests

PR #10361, opened 2025-03-11
fix(nemesis.yml, nemesis_classes.yml): update the list

cezarmoise:
These lists were not updated with recent changes.
Updated them by running unit tests.

### Testing

- [ ]

### Reminders

### DIFF
--- a/data_dir/nemesis.yml
+++ b/data_dir/nemesis.yml
@@ -18,20 +18,24 @@
   - run_with_gemini = False
   - limited = True
   - topology_changes = True
+- disrupt_add_remove_dc: []
 - disrupt_add_remove_mv:
   - disruptive = True
   - schema_changes = True
   - free_tier_set = True
+  - supports_high_disk_utilization = False
 - disrupt_bootstrap_streaming_error:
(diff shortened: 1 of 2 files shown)


pehala:
@fruch This seems like an extensive list, could we add some checks to precommit (or prepush) to catch this?

fruch:
> @fruch This seems like an extensive list, could we add some checks to precommit (or prepush) to catch this?

we have this issue also for SCT configuration docs, which people forget to update

it far from being straight forward as per-commit

i might be better doing in CI, running the test / tools, and notifying on the PR

fruch:
> > @fruch This seems like an extensive list, could we add some checks to precommit (or prepush) to catch this?
> 
> we have this issue also for SCT configuration docs, which people forget to update
> 
> it far from being straight forward as per-commit
> 
> i might be better doing in CI, running the test / tools, and notifying on the PR

and here's an example of how it can be achieved: 
https://github.com/scylladb/scylla-cluster-tests/pull/10367

fruch (approved):
LGTM

PR #10367, opened 2025-03-11
feature(actions): check generated code is up-to-date

fruch:
since we have multiple part of configuration and documentation which need to be updated via code generation tools this new workflow would be executing the tools, and updating the user on top of the PR, if they are not up-to-date

### Testing

- [x] tested by running this github action

### Reminders

### DIFF
--- a/.github/workflows/check-generated-code-updates.yaml
+++ b/.github/workflows/check-generated-code-updates.yaml
@@ -1,0 +1,53 @@
+name: Check generated code is up-to-date
+
+on:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
+    branches:
+      - master
+
+jobs:
+  check_org_membership:
+    runs-on: ubuntu-latest
+    outputs:
+      isTeamMember: ${{ steps.teamAffiliation.outputs.isTeamMember }}
+    steps:
+      - name: Check user for team affiliation
+        uses: tspascoal/get-user-teams-membership@v3
+        id: teamAffiliation
+        with:
+          GITHUB_TOKEN: ${{ secrets.AUTO_BACKPORT_TOKEN }}
+          username: ${{ github.actor }}
+          team: ${{ secrets.SCT_ACTION_GITHUB_TEAM }}
+
+  run_validations:
+    runs-on: ubuntu-latest
+    needs: check_org_membership
+    if: needs.check_org_membership.outputs.isTeamMember == 'true'
+    env:
+      AWS_ACCESS_KEY_ID: ${{ secrets.QA_USER_AWS_ACCESS_KEY_ID }}
+      AWS_SECRET_ACCESS_KEY: ${{ secrets.QA_USER_AWS_SECRET_ACCESS_KEY }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.head_ref }}
+
+      - name: Run nemesis list validation
+        uses: ./.github/actions/run-validation
+        with:
+          script: python -m pytest unit_tests/test_nemesis_sisyphus.py::test_list_all_available_nemesis
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run configuration docs validation
+        uses: ./.github/actions/run-validation
+        with:
+          script: update-conf-docs
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Run nemesis pipelines validation
+        uses: ./.github/actions/run-validation
+        with:
+          script: create-nemesis-pipelines
+          github_token: ${{ secrets.GITHUB_TOKEN }}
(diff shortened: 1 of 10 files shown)


.github/workflows/check-generated-code-updates.yaml
```diff
@@ -0,0 +1,110 @@
+name: Check generated code is up-to-date
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
```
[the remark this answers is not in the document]
fruch: done

fruch:
> Just FYI: I created an alternative solution that utilized precommit in #10385

@scylladb/qa-maintainers waiting for you feedback, on which approach would be more useful (we can mix and match, for example nemesis with pre-commit hook, and docs with comments on PR like here)

vponomaryov (approved):
LGTM*

* if works

vponomaryov:
> > Just FYI: I created an alternative solution that utilized precommit in #10385
> 
> @scylladb/qa-maintainers waiting for you feedback, on which approach would be more useful (we can mix and match, for example nemesis with pre-commit hook, and docs with comments on PR like here)

`pre-commit` easily may be disabled at some point in time by an individual with idea that it is either `slow` or `requires some system packages`...
So, this approach is a must, and second approach is good to have, IMHO.

pehala:
> @scylladb/qa-maintainers waiting for you feedback, on which approach would be more useful (we can mix and match, for example nemesis with pre-commit hook, and docs with comments on PR like here)

I agree mixing both of those approaches is the best, tbh: it wouldnt hurt having both of them, they are independent of each other.

fruch:
> > @scylladb/qa-maintainers waiting for you feedback, on which approach would be more useful (we can mix and match, for example nemesis with pre-commit hook, and docs with comments on PR like here)
> 
> I agree mixing both of those approaches is the best, tbh: it wouldnt hurt having both of them, they are independent of each other.

O.k. but we should keep them aligned

I.e. if there is a command instead of a unit test we need to keep it updated

I'll rebase and try to push this one forward first

vponomaryov (approved):
LGTM

.github/workflows/check-generated-code-updates.yaml
```diff
@@ -0,0 +1,112 @@
...
+    outputs:
+      isTeamMember: ${{ steps.teamAffiliation.outputs.isTeamMember }}
+    steps:
+      - name: Check user for team affiliation
+        uses: tspascoal/get-user-teams-membership@v3
+        id: teamAffiliation
+        with:
+          GITHUB_TOKEN: ${{ secrets.AUTO_BACKPORT_TOKEN }}
+          username: ${{ github.actor }}
+          team: 'dev'
```
pehala, line 21: Why is the team hardcoded?

.github/workflows/check-generated-code-updates.yaml
```diff
@@ -0,0 +1,112 @@
...
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ github.head_ref }}
+
+      - name: generate nemesis list
+        id: generate_nemesis_list
+        run: |
+          bash ./docker/env/hydra.sh python -m pytest unit_tests/test_nemesis_sisyphus.py::test_list_all_available_nemesis
```
pehala, line 40: All of these are basically the same, except for the command, I would unify them, e.g. by extracting this functionality to a shell script living in the repository

pehala (approved):

.github/workflows/check-generated-code-updates.yaml
fruch: it would be hard code (but maybe should be a secret)

it's one of few safe guards, so we can control who would be able to trigger this one

.github/workflows/check-generated-code-updates.yaml
fruch: could be we can template it a bit, I didn't had time to dig into how it can cleanly done in github actions

it can be done as followup

.github/workflows/check-generated-code-updates.yaml
pehala: Yeah, I was thinking about having it as a secret

.github/workflows/check-generated-code-updates.yaml
pehala: I was thinking maybe we can do it as regular bash script, which would write you the exact error message, which would live in the code and have 3 invocations of it in the GH action. It can be done as a followup though

fruch: We'll need to parameterize the heck out of it, or maybe just the command and some name for it would suffice.

.github/workflows/check-generated-code-updates.yaml
fruch: fixed it.

.github/workflows/check-generated-code-updates.yaml
fruch: I'll take a look, but it would be a followup

pehala (dismissed):

.github/workflows/check-generated-code-updates.yaml
fruch: copilot was quick enough doing most of it
I'll give it one quite shot, otherwise I'll drop it

PR #10381, opened 2025-03-12
improvement(nemesis_jobs): Improve nemesis job generation

pehala:
* Simplify templates
* Move job configuration into Nemesis classes
   * Makes it easier to see what configs are required for each nemesis
   * Add `additional_configs` and `additional_params`
* Regenerate according to new settings

Questions:
* Nemesis yaml now contains also `additional_configs` and `additional_param` fields. Is it an issue?  

### Testing

- [x] Locally, with `hydra create-nemesis-pipelines --backend docker --backend aws --backend azure --backend gce`

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

PR #10385, opened 2025-03-12
improve(nemesis_jobs): Rework nemesis jobs generation

pehala:
### Problem
Current workflow for regenerating nemesis.yaml/nemesis_classes.yaml relies on running unit tests (I also think this is bad way of doing this, unit tests should NEVER generate anything)  and for unknown reason is not executed in CI (or at least it is not failing there). Generating jobs from these files is also manual process and requires running  `hydra create-nemesis-pipelines --backend docker --backend aws --backend azure --backend gce`, which requires AWS login, even though it is not needed for this operation. Overall the process is entirely manual and done in two separate places.

### Solution
Unite the two steps into one tool (`tools/nemesis`) which can be run locally without AWS credentials and is included in precommit hooks, thus being checked both locally and in CI automatically. The tool allows running in 3 modes: regenerate everything, verify that files are up to date and verify + fix files (similar to how ruff fails, but still formats the files).

The actual logic is not changed. This PR follows up on https://github.com/scylladb/scylla-cluster-tests/pull/10381 and majority of the changes are from there.

### Drawbacks
* Generating yamls from current Nemesis takes around ~15 sec, this is caused mostly by our tight integration of Nemesis discovery and the actual code. This could be improved by separating these two in followup PR. Nonetheless this 15sec penalty is only applied if templates or actual nemesis code is changed so it doesnt apply on every commit. Solution to this could be moving this check to `prepush` phase, where it triggers less, but still manages to check the changes.
* Current code of getting all nemesis relies on mocking some classes due to existing Nemesis dicovery code, this is extremely ugly and would be nice to refactor in followup, but this was the case with existing code as well.

### TBD
- [ ] Better documentation (mostly docstrings)
   * Will be done after the initial idea is discussed and approved 

### Testing

- [x] Locally


pehala:
@roydahan @fruch Let me know if you have questions about the general idea and if you think this is a good direction to take

configurations/nemesis/longevity-5gb-1h-nemesis.yaml
```diff
@@ -19,17 +19,12 @@ instance_type_db: 'i4i.large'
...
 
 nemesis_interval: 3
 
 gce_n_local_ssd_disk_db: 1
 
-
 server_encrypt: true
 client_encrypt: true
-
-teardown_validators:
```
fruch, line 33: this should be added to the template

data_dir/nemesis.yml
```diff
@@ -68,6 +67,7 @@
 - disrupt_disable_enable_ldap_authorization:
   - disruptive = True
   - limited = True
+  - additional_configs = ['configurations/ldap-authorization.yaml']
```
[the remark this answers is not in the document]
pehala: I moved additional_configs and additional_params to the classes themselves in #10381, to make it more clear what the nemesis need for their run. Since it is a class property it gets propagated here. I can filter these two properties our in the discovery mechanism if needed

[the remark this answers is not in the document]
fruch: yes, it irrelevant to this list

fruch:
one more drawback of it, that it can't be disabled for release branches.
i.e. the original idea is that we can work by the nemesis list, and by the code, so we'll be able to run a specific set of nemesis, and even if we get some backport in with new nemesis.

this precommit hook would be in the way in those situations.
and just notification on PRs might be a bit more tolerable

pehala:
> one more drawback of it, that it can't be disabled for release branches. i.e. the original idea is that we can work by the nemesis list, and by the code, so we'll be able to run a specific set of nemesis, and even if we get some backport in with new nemesis.
> 
> this precommit hook would be in the way in those situations. and just notification on PRs might be a bit more tolerable

There is always a possibility of ignoring it, you are right that it is harder to do with precommit than with GH actions, but given that we never implemented the idea of using the generate .yml and it seems far away, I would rather have a good developer experience now (for which precommit is superior) with local solution that doesnt need more pipeline work, than choose a different solution just because of "what if" in the future.

fruch:
> > There is always a possibility of ignoring it, you are right that it is harder to do with precommit than with GH actions, but given that we never implemented the idea of using the generate .yml and it seems far away, I would rather have a good developer experience now (for which precommit is superior) with local solution that doesnt need more pipeline work, than choose a different solution just because of "what if" in the future.
> 
> One solution would be to not run this specific check on CI in release branches. Or just remove the check for release branches all together when we branch. It is one additional commit we have to do, but given the frequency I do not see it as a problem

again the problem isn't about one extra commit, but one us having a ability to pin down the list of nemesis used in a release, which is the whole purpose of generating this data (even if we aren't using it for it yet)
the pipeline generation was just joining it, it wasn't the main purpose out it.

roydahan:
I'll try to reach it later this week.

data_dir/nemesis.yml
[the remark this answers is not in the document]
pehala: Removed

fruch:
@pehala 

one of the commits fails cause `s/improveme/improvement/`

it can be added to the jscommitlint configuration, if you want

but that one more thing about the pre-commit hooks, it's not visible which one is failing (and all of them block the PR, and only admin can wavier, which is o.k. for commit lint, but other check fail the rest of the PRs)

pehala:
> @pehala
> 
> one of the commits fails cause `s/improveme/improvement/`
> 
> it can be added to the jscommitlint configuration, if you want
> 
Nah this is error on my part
> but that one more thing about the pre-commit hooks, it's not visible which one is failing (and all of them block the PR, and only admin can wavier, which is o.k. for commit lint, but other check fail the rest of the PRs)
That is true, I think we can do a better job in CI to improve the status (like for example running them through GH actions instead of jenkins).

fruch:
> > @pehala
> > one of the commits fails cause `s/improveme/improvement/`
> > it can be added to the jscommitlint configuration, if you want
> 
> Nah this is error on my part
> 
> > but that one more thing about the pre-commit hooks, it's not visible which one is failing (and all of them block the PR, and only admin can wavier, which is o.k. for commit lint, but other check fail the rest of the PRs)
> 
> That is true, I think we can do a better job in CI to improve the status (like for example running them through GH actions instead of jenkins).

even then, you won't have report for every single check in pre-commit, you can run it one by one and report maybe.

switching some part to github action can be nice, but there are some complications to it, i.e. the duration, running tests and integration tests, might not fit with the default builders

vponomaryov:
Transformation of those `15s` into `5.8s`:
- https://github.com/scylladb/scylla-cluster-tests/pull/10551

.pre-commit-config.yaml
```diff
@@ -51,6 +51,13 @@ repos:
         types: [python]
         exclude: ^argus/.*$
 
+    -   id: nemesis_jobs
+        name: Regenerate nemesis jobs
+        entry: python -m sdcm.tools.nemesis generate-all --verify --fix
```
vponomaryov, line 56: Need to use "python3" instead of the "python".
The latter one is absent in some cases.

pehala:
> Generating jobs from these files is also manual process and requires running `hydra create-nemesis-pipelines --backend docker --backend aws --backend azure --backend gce`, which requires AWS login, even though it is not needed for this operation. Overall the process is entirely manual and done in two separate places.

With #10534 is resolved, I think I can simplify this PR and integrate it into hydra as well

.pre-commit-config.yaml
pehala: Removed from this PR, I will do a followup PR which will include both `create-nemesis-pipelins`, `create-nemesis-yaml` and `update-conf-docs` in the precommit together, as they are quite similar

pehala:
v2: Rework the PR
  * Drop the new tool, integrate everything into sct.py
  * Simplify changes
  * Remove precommit changes

I also updated the cover letter. I would still like to merge https://github.com/scylladb/scylla-cluster-tests/pull/10381 first, as it contains majority of the changes

pehala:
The checks are run from master (due to `pull_request_target`) and thus the change to the workflow file is not propagated.

sdcm/utils/verify.py
```diff
@@ -0,0 +1,30 @@
...
+        old_path = Path(self.path)
+        if old_path.exists():
+            old = old_path.read_text()
+        else:
+            old = ""
+        diff = unified_diff(old.splitlines(keepends=True), new.splitlines(keepends=True), fromfile=str(self.path))
+        output = "".join(diff)
+        if len(output) > 0:
+            click.secho(output, fg="red")
+            self.__class__.ERROR = True
```
fruch, line 28: communicating via the class back for the results ?

this pattern is confusing as heck...

sdcm/utils/verify.py
```diff
@@ -0,0 +1,30 @@
+from difflib import unified_diff
+from io import StringIO
+from pathlib import Path
+
+import click
+
+
+class Fix(StringIO):
```
fruch, line 8: the name `Fix` is meaningless and not close to the actual thing this class is doing....

sdcm/utils/verify.py
pehala: Changed, because the open() does not have mechanism to return back results, I had to create a "carrier" class which would be used to set error and translate them outside of function. It should be much cleaner (though still not perfect) now.

pehala:
v3:
  * Redesigned `Fix` class and rename it to `OpenWithDiff`
     * Much cleaner error return mechanism

fruch (approved):
LGTM
now it's least it's readable when used.

pehala:
@vponomaryov Could you please re-review? You commented on this before.

The check failure is due to `pull_request_target` using old workflow file (i.e. from master)

vponomaryov (approved):
LGTM

Issue #10314, opened 2025-03-07
SCT thread failed with "AttributeError: 'NoneType' object has no attribute 'get'"


timtimb0t:
## Packages

Scylla version: `2025.2.0~dev-20250228.439463dbbf7e` with build-id `5bb9c3cb0cecc46819d360468f183eb30dedd501`

Kernel Version: `6.8.0-1023-aws`

STC framework need little adjustment to avoid error like:
```
2025-03-01 10:17:50.723: (ThreadFailedEvent Severity.ERROR) period_type=one-time event_id=40f464ed-7d00-4e21-9937-23907984779a: message='NoneType' object has no attribute 'get'
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/decorators.py", line 26, in wrapper
return func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 480, in run
self.disrupt()
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 6733, in disrupt
self.call_next_nemesis()
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2041, in call_next_nemesis
self.execute_disrupt_method(disrupt_method=next(self.disruptions_cycle))
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1973, in execute_disrupt_method
disrupt_method()
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5591, in wrapper
args[0].cluster.check_cluster_health()
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 4421, in check_cluster_health
node.check_node_health()
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 2738, in check_node_health
event = next(events, None)
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/raft/__init__.py", line 331, in check_group0_tokenring_consistency
self._node.name, self._node.host_id)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 488, in host_id
return self.parent_cluster.get_nodetool_info(self, ignore_status=True, publish_event=False).get("ID")
AttributeError: 'NoneType' object has no attribute 'get'
```

## Impact
Just SCT issue

## How frequently does it reproduce?

[...]

PR #10422, opened 2025-03-17
fix(cluster.py): None type added as returned value in host_id prop

timtimb0t:
to avoid AttributeError: 'NoneType' object has no attribute 'get'
minor fix 
ref: https://github.com/scylladb/scylla-cluster-tests/issues/10314

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -487,7 +487,8 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
 
     @property
     def host_id(self) -> str | None:
-        return self.parent_cluster.get_nodetool_info(self, ignore_status=True, publish_event=False).get("ID")
+        info = self.parent_cluster.get_nodetool_info(self, ignore_status=True, publish_event=False) or {}
+        return info.get("ID")
 
     @property
     def db_node_instance_type(self) -> Optional[str]:


sdcm/cluster.py
```diff
@@ -487,7 +487,7 @@ def region(self):
 
     @property
     def host_id(self) -> str | None:
-        return self.parent_cluster.get_nodetool_info(self, ignore_status=True, publish_event=False).get("ID")
+        return self.parent_cluster.get_nodetool_info(self, ignore_status=True, publish_event=False).get("ID", None)
```
vponomaryov, line 490: It does not fix anything.
The referenced bug description tells us that the output of the nodetool command may return `None`.
So, nothing will change here.
Moreover, `foo.get("ID", None)` and `foo.get("ID")` technically behave the same.

vponomaryov (changes requested):

sdcm/cluster.py
timtimb0t: @vponomaryov , yes, definitely, my bad.
the main idea was to make something like:
```
def host_id(self) -> str | None:
    info = self.parent_cluster.get_nodetool_info(self, ignore_status=True, publish_event=False)
    if info is not None:
        return info.get("ID")
    return None
```

fruch (approved):
LGTM

vponomaryov (approved):
LGTM

PR #10444, opened 2025-03-19
refactor(elastic): remove all elasticsearch related code

fruch:
since we are now sending performence stats into Argus, we don't need anything to be sending data to ElasticSearch anymore

### Testing

- [ ]

### Reminders


fruch:
seems like I've forgot about `adaptive_timeouts` usage of ES.

@soyacz, you think we can switch it to save those in argus ?

soyacz:
> seems like I've forgot about `adaptive_timeouts` usage of ES.
> 
> @soyacz, you think we can switch it to save those in argus ?

I propose to send as 'results': 
1. Each operation will have own table (name could be e.g. `Adaptive Timeout - decommission`)
2. Each operation execution will get new row in that table

For some tests we could even introduce validation rules based on 'best' - so it would really serve it's initial purpose. (This should play well in individual nemesis tests)

For now, I think you can just change default store to `AdaptiveTimeoutStore` class here:
```
def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: F821
                     stats_storage: AdaptiveTimeoutStore = ESAdaptiveTimeoutStore(), **kwargs):
```
It's noop, and ArgusStore can be developed separately.

fruch:
> > seems like I've forgot about `adaptive_timeouts` usage of ES.
> > @soyacz, you think we can switch it to save those in argus ?
> 
> I propose to send as 'results':
> 
> 1. Each operation will have own table (name could be e.g. `Adaptive Timeout - decommission`)
> 2. Each operation execution will get new row in that table
> 
> For some tests we could even introduce validation rules based on 'best' - so it would really serve it's initial purpose. (This should play well in individual nemesis tests)
> 
> For now, I think you can just change default store to `AdaptiveTimeoutStore` class here:
> 
> ```
> def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: F821
>                      stats_storage: AdaptiveTimeoutStore = ESAdaptiveTimeoutStore(), **kwargs):
> ```
> 
> It's noop, and ArgusStore can be developed separately.

we also would need ability to read the results back at some point

fruch:
closing in favor of https://github.com/scylladb/scylla-cluster-tests/pull/13279

PR #10448, opened 2025-03-20
fix(gce_cloud_report): report and handle stopped instances

fruch:
* fix `list_instances_gce` to return stopped instances
* fix gce cleanup to default to `keep_action==terminate`, even if action label is empty

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -1053,7 +1053,7 @@ def list_instances_gce(tags_dict: Optional[dict] = None,
     if running:
         instances = [i for i in instances if i.status == 'RUNNING']
     else:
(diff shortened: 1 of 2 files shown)


fruch:
> LGTM Shouldn't we warn again people about this? Someone might be badly surprised to lose instance parked for some reason.

too late for that, I've deleted all of the stopped one manually
people aren't aware about those "features", and not really using them.

soyacz:
> > LGTM Shouldn't we warn again people about this? Someone might be badly surprised to lose instance parked for some reason.
> 
> too late for that, I've deleted all of the stopped one manually people aren't aware about those "features", and not really using them.

:)
Anyway, I was warning people at time when I was moving it to SCT.

PR #10470, opened 2025-03-23
feature(argus): call client with extra headers

fruch:
since argus serivce will need more headers for it's to be operational, we'll need to change the calls to it, to pass the needed headers.

### Testing

- [ ]

### Reminders


fruch:
first need to release a new argus client version, and update it as part of this PR

k0machi (dismissed):

argus/version
```diff
@@ -1 +1 @@
-0.13.1
+0.14.1
```
soyacz, line 1: 0.14.2?

fruch: yes it would, didn't got yet to update that one, had issue in other fronts.

fruch: updated it

soyacz (approved):
LGTM

soyacz:
@fruch do we know the headers to test it e.g. against staging?

fruch:
> @fruch do we know the headers to test it e.g. against staging?

yes we do, I tested it locally.

we'll create a separated configuration file for staging, if for once we'll need it

Issue #10426, opened 2025-03-18
[branch-2024.1] During parallel nemeses execution, the NodetoolDecommission nemesis removed the node from the cluster before the parallel CreateIndex one was able to finish


dimakr:
## Packages

Scylla version: `2024.1.16-20250314.5c90d8c3aac9` with build-id `1280109d6e076642019b7165f94b0ad2d50fb1d4`
Kernel Version: `5.15.0-1079-aws`

## Issue description

During [enterprise-2024.1/longevity/longevity-schema-topology-changes-12h-test#33](https://argus.scylladb.com/tests/scylla-cluster-tests/64c43292-2f39-4a10-a298-556194dcada1) test run 2 nemeses were started to run in parallel:
- `disrupt_nodetool_decommission` within 17:55:48 - 18:07:51 
- `disrupt_create_index` within 17:56:41 - 18:05:23

node-5 gets decommission API request at `17:55:55` and the decommission finished at `17:58:16`:
```
< t:2025-03-14 17:55:55,063 f:db_log_reader.py l:123  c:sdcm.db_log_reader   p:DEBUG > 2025-03-14T17:55:55.061+00:00 longevity-parallel-topology-schema--db-node-64c43292-5     !INFO | scylla[6889]:  [shard 0:stre] api - decommission
...
< t:2025-03-14 17:58:16,609 f:cluster.py      l:4841 c:sdcm.cluster         p:INFO  > Decommission Node longevity-parallel-topology-schema--db-node-64c43292-5 [3.85.53.11 | 10.12.9.226] PASS
```

At the same time create_index routine is started on the node at `17:56:52`, but failed at some point as the node is removed from the cluster in the meantime: 
```
< t:2025-03-14 17:56:52,558 f:common.py       l:1777 c:utils                p:DEBUG > Executing CQL 'CREATE INDEX standard1_c3_nemesis ON keyspace1.standard1("C3")'
...
< t:2025-03-14 17:57:03,049 f:cluster.py      l:2571 c:sdcm.cluster_aws     p:DEBUG > Node longevity-parallel-topology-schema--db-node-64c43292-5 [3.85.53.11 | 10.12.9.226]: Command '/usr/bin/nodetool  viewbuildstatus keyspace1.standard1_c3_nemesis_index ' duration -> 2.9720386199987843 s
...
[...]

PR #10477, opened 2025-03-23
fix(nemesis.py): make node unselection and selection atomic in set_target_node

dimakr:
Makes unselection and then selection an atomic operation in `set_target_node`.
Previously unselection and then selection were performed not in the same lock scope, leading to a potential situation when 2 threads could pick up the same node for parallel nemeses.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/10426

### Testing

- [x] :green_circle: [longevity-schema-topology-changes-12h-test - the run executes disrupt_nodetool_decommission and disrupt_create_index 10 times in  parallel](https://argus.scylladb.com/tests/scylla-cluster-tests/e74cf286-edb4-4007-8d0e-31059fcafe28)
The test run was executed on branch-2024.1 as the `disrupt_create_index` nemesis (one of 2 parallel nemeses, which were causing the failure in the original issue) is skipped in SCT master due SkipPerIssues condition.
But the fix is applicable to master as well.

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -447,10 +447,11 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         if is_seed is DefaultValue - if self.filter_seed is True it act as if is_seed=False,
           otherwise it will act as if is_seed is None
         """
-        # first give up the current target node
-        self.unset_current_running_nemesis(self.target_node)
-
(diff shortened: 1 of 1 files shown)


vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #10496, opened 2025-03-25
fix(nemesis): make nemesis more safely by adding repair on all nodes

timtimb0t:
this change add repair at the beggining of disrupt_repair_streaming_err to make this nemesis more safe and avoid c-s data validation errors ref: https://github.com/scylladb/scylladb/issues/21428

fixes: https://github.com/scylladb/scylladb/issues/21428

### Testing
https://jenkins.scylladb.com/job/scylla-staging/job/eugene_test_folder/job/repair_repair/6/

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1898,6 +1898,21 @@ class Nemesis:
         with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
             node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
 
+    def run_repair_on_nodes(self, nodes: list, publish_event=True):
+        """
+        Execute a nodetool repair on the specified nodes, disregarding errors that may
+        arise from failed or unavailable nodes during the process.
+        """
+        if not self.cluster.params.get('use_mgmt') and not self.cluster.params.get('use_cloud_manager'):
+            for node in nodes:
+                try:
+                    with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 3):
+                        node.run_nodetool(sub_cmd="repair -pr", publish_event=publish_event)
+                except Exception as err:  # pylint: disable=broad-except  # noqa: BLE001
+                    self.log.warning(f"Repair failed to complete on node: {node}, with error: {str(err)}")
+        else:
+            self._mgmt_repair_cli()
+
     def repair_nodetool_rebuild(self):
         with adaptive_timeout(Operations.REBUILD, self.target_node, timeout=HOUR_IN_SEC * 48):
             self.target_node.run_nodetool('rebuild', long_running=True, retry=0)
@@ -3618,12 +3633,7 @@ class Nemesis:
         # and as a result requires ignoring repair errors
         with DbEventsFilter(db_event=DatabaseLogEvent.RUNTIME_ERROR,
                             line="failed to repair"):
-            for node in up_normal_nodes:
-                try:
-                    self.repair_nodetool_repair(node=node, publish_event=False)
-                except Exception as details:  # noqa: BLE001
-                    self.log.error(f"failed to execute repair command "
-                                   f"on node {node} due to the following error: {str(details)}")
+            self.run_repair_on_nodes(nodes=up_normal_nodes)
 
         exit_status = remove_node()
         # if remove node command failed by any reason,
@@ -4076,7 +4086,11 @@ class Nemesis:
     def disrupt_repair_streaming_err(self):
         """
         Stop repair in middle to trigger some streaming fails, then rebuild the data on the node.
+        Repair call before streaming is needed to avoid c-s data validation error.
+        Ref: https://github.com/scylladb/scylladb/issues/21428
         """
+        self.log.debug('Cluster repair starts')
+        self.run_repair_on_nodes(nodes=self.cluster.data_nodes)
         with ignore_raft_topology_cmd_failing():
             self.start_and_interrupt_repair_streaming()
 


sdcm/nemesis.py
```diff
@@ -2090,10 +2090,15 @@ def call_next_nemesis(self):
         self.execute_disrupt_method(disrupt_method=next(self.disruptions_cycle))
 
     @latency_calculator_decorator(legend="Run repair process with nodetool repair")
-    def repair_nodetool_repair(self, node=None, publish_event=True):
-        node = node if node else self.target_node
-        with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
-            node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+    def repair_nodetool_repair(self, node=None, publish_event=True, all_nodes=False):
+        if all_nodes:
+            for node in self.cluster.nodes:
```
aleksbykov, line 2095: you can use only data nodes `self.clsuter.data_nodes`. Repair operation could be skipped on zero-token nodes

sdcm/nemesis.py
```diff
@@ -2090,10 +2090,15 @@ def call_next_nemesis(self):
         self.execute_disrupt_method(disrupt_method=next(self.disruptions_cycle))
 
     @latency_calculator_decorator(legend="Run repair process with nodetool repair")
-    def repair_nodetool_repair(self, node=None, publish_event=True):
-        node = node if node else self.target_node
-        with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
-            node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+    def repair_nodetool_repair(self, node=None, publish_event=True, all_nodes=False):
```
aleksbykov, line 2093: i think, you need go over the method usage, and find out whether this method used in Non-disruptive nemesis(methods), so operation is not failed if node is under Disruptive nemesis. Otherwise, you need to make this changes a bit safely by adding ignore status of nodetool command or add try except

sdcm/nemesis.py
```diff
@@ -4273,6 +4278,7 @@ def disrupt_repair_streaming_err(self):
         """
         Stop repair in middle to trigger some streaming fails, then rebuild the data on the node.
         """
+        self.repair_nodetool_repair(all_nodes=True)
```
aleksbykov, line 4281: add a short comment, why we need it here

aleksbykov (commented):
Need make the repair method a bit safely

sdcm/nemesis.py
```diff
@@ -2090,10 +2090,15 @@ def call_next_nemesis(self):
...
-            node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+    def repair_nodetool_repair(self, node=None, publish_event=True, all_nodes=False):
+        if all_nodes:
+            for node in self.cluster.data_nodes:
+                with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
+                    node.run_nodetool(sub_cmd="repair", publish_event=publish_event, ignore_status=True)
+        else:
+            node = node if node else self.target_node
+            with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
+                node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
```
aleksbykov, line 2101: lets change this logic. lets try to do without if statement and additional flag
let use something like this:
```python
def repair_nodetool_repair(self, nodes: list | none, timeout: int = HOUR_IN_SEC * 48):
   nodes_to_repair = nodes or [self.target_node]
   for node in nodes_to_repair:
      try:
          with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
                  node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
      except Exception as exc:
               log.warning("repair failed with %s", exc)
```
and need to check each call of repair_nodetool_repair method to update accordingly

sdcm/nemesis.py
```diff
@@ -2090,10 +2090,14 @@ def call_next_nemesis(self):
...
-        with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
-            node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+    def repair_nodetool_repair(self, nodes: list | None, publish_event=True):
+        nodes_to_repair = nodes or [self.target_node]
+        for node in nodes_to_repair:
+            try:
+                with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
+                    node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+            except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
+                self.log.warning("Repair failed with %s", exc)
```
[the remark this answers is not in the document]
temichus: agree. You can change to hide any errors/bugs that might have happened under the warning message. If you want to do "safe repair" you need to create a separate method for this

temichus (changes requested):

aleksbykov:
Here https://github.com/timtimb0t/scylla-cluster-tests/blob/619cfb05b0c5bc8126401ef95945f7b7a65cddac/sdcm/nemesis.py#L3629 same code is used, you can replace it with your method. And add please docstring with description why we skip repair error

sdcm/nemesis.py
```diff
@@ -1900,6 +1900,14 @@ def repair_nodetool_repair(self, node=None, publish_event=True):
         with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
             node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
 
+    def run_repair_on_nodes(self, nodes: list, publish_event=True):
+        for node in nodes:
+            try:
+                with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
```
aleksbykov, line 1906: Please add here docstring, why we ignore repair errors

aleksbykov (dismissed):
Add documentation

temichus (dismissed):

sdcm/nemesis.py
```diff
@@ -1898,6 +1898,17 @@ def repair_nodetool_repair(self, node=None, publish_event=True):
         with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
             node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
 
+    def run_repair_on_nodes(self, nodes: list, publish_event=True):
+        """
+        Ignore errors so that a single failure doesn’t cause the entire nemesis failure.
+        """
```
aleksbykov, line 1904: Need another docstring.
`Execute a nodetool repair on the specified nodes, disregarding errors that may arise from failed or unavailable nodes during the process.`

aleksbykov (dismissed):
LGTM, only update the docstring

temichus (approved):

aleksbykov (approved):

timtimb0t:
@scylladb/qa-maintainers , could you please merge?

aleksbykov:
@timtimb0t , please add link to job with fix where the code was executed

sdcm/nemesis.py
```diff
@@ -1898,6 +1898,18 @@ def repair_nodetool_repair(self, node=None, publish_event=True):
...
             node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
 
+    def run_repair_on_nodes(self, nodes: list, publish_event=True):
+        """
+        Execute a nodetool repair on the specified nodes, disregarding errors that may
+        arise from failed or unavailable nodes during the process.
+        """
+        for node in nodes:
+            try:
+                with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48):
```
vponomaryov, line 1908: 48 hours timeout for a single node? Why that big value?

We have a test scenario with 60 DB nodes.
How long will it take on a setup with 60 nodes?

fruch: I wonder who's the one that suggest to do those repairs... (surprise surprise it was @fruch)

I would recommend using manager repair, or in case of tablets `nodetool cluster repair` 

doing those serially on all nodes, would be very slow and efficient

also please cross check this with the core team owning repair, what they suggestion and if they approve this

on note, on tablets this won't be 48, it would fail after 1h, and raise critical after 3h
regardless 48h doesn't make no sense in this situation, now days we don't run test for more then 12, just a handful longer ones during release

sdcm/nemesis.py
```diff
@@ -1898,6 +1898,18 @@ def repair_nodetool_repair(self, node=None, publish_event=True):
...
+                    node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
```
roydahan, line 1909: Do not use nodetool repair, use a manager repair.

timtimb0t:
@fruch , @soyacz could you please take a look?

sdcm/nemesis.py
```diff
@@ -3613,17 +3613,11 @@ def remove_node():
...
         # and as a result requires ignoring repair errors
         with DbEventsFilter(db_event=DatabaseLogEvent.RUNTIME_ERROR,
                             line="failed to repair"):
-            for node in up_normal_nodes:
-                try:
-                    self.repair_nodetool_repair(node=node, publish_event=False)
-                except Exception as details:  # noqa: BLE001
-                    self.log.error(f"failed to execute repair command "
-                                   f"on node {node} due to the following error: {str(details)}")
+            self.target_node.run_nodetool(sub_cmd="cluster repair", publish_event=True)
```
soyacz, line 3620: this works only with tablets, for vnodes based need to use manager repair

timtimb0t:
@fruch , @roydahan

sdcm/nemesis.py
```diff
@@ -1898,6 +1898,21 @@ def repair_nodetool_repair(self, node=None, publish_event=True):
...
+        Execute a nodetool repair on the specified nodes, disregarding errors that may
+        arise from failed or unavailable nodes during the process.
+        """
+        if not self.cluster.params.get('use_mgmt') and not self.cluster.params.get('use_cloud_manager'):
+            for node in nodes:
+                try:
+                    with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 3):
+                        node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+                except Exception as err:  # pylint: disable=broad-except  # noqa: BLE001
+                    self.log.warning(f"Repair failed to complete on node: {node}, with error: {str(err)}")
```
roydahan, line 1912: Would be better to run it in parallel and each with the flag --pr.
Otherwise it can take lot of time.
OTOH, I'm not sure the load it adds to the cluster, need to test it on a scenario of longevity with large cluster (not necessarily huge) and high load (with MVs / SI / etc).

fruch: I don't think we have any case we don't install manager in

I don't know if we need to push that much effort into optimizing the code path that doesn't have manager

fruch:
@timtimb0t  

 the original referenced bug talks about `disrupt_destroy_data_then_rebuild`
https://github.com/scylladb/scylladb/issues/21428#issuecomment-2490405610

but this PR doesn't do anything with that nemesis, so I'm a bit confused

timtimb0t:
> @timtimb0t
> 
> the original referenced bug talks about `disrupt_destroy_data_then_rebuild` [scylladb/scylladb#21428 (comment)](https://github.com/scylladb/scylladb/issues/21428#issuecomment-2490405610)
> 
> but this PR doesn't do anything with that nemesis, so I'm a bit confused

it was decided to place fix before the nemesis that failed (ie disrupt_repair_streaming_err) because it failed few times with the same error

sdcm/nemesis.py
timtimb0t: So what is the solution? Do we need if else condition or just run manager repair? Or we still need usual nodetool repair but with -pr flag and run it in parallel?

fruch: as @roydahan said, let's just add `-pr` flag, and that's it, optimizing it further won't be needed for now

temichus (approved):

enaydanov (approved):

sdcm/nemesis.py
```diff
@@ -3613,17 +3613,10 @@ def remove_node():
...
-        with DbEventsFilter(db_event=DatabaseLogEvent.RUNTIME_ERROR,
-                            line="failed to repair"):
-            for node in up_normal_nodes:
-                try:
-                    self.repair_nodetool_repair(node=node, publish_event=False)
-                except Exception as details:  # noqa: BLE001
-                    self.log.error(f"failed to execute repair command "
-                                   f"on node {node} due to the following error: {str(details)}")
+        self.log.debug('Cluster repair starts')
+        self._mgmt_repair_cli()
```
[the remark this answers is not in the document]
enaydanov: looks like outdated comment, mark it as resolved

sdcm/nemesis.py
enaydanov: looks like outdated comment, mark it as resolved

PR #10502, opened 2025-03-25
improvement(nemesis): Rework nemesis discovery

pehala:
# Problem statement

Currently, the Nemesis class has too many responsibilities, which causes problems, one of the problems is that it requires parameters to initialize and does a lot of logic in the `init` function and you need to provide those even if you need only part of the class. Problem manifests can be seen in `test_nemesis_sisyphus` which needs to mock Tester, Cluster and Node to filter nemesis, but none of methods used required any of this, we need to provide it because of the aforementioned problem.

# Solution

Extract Nemesis discovery (i.e. Gathering all disrupt method and matching it with subclasses) to a separate class. While doing so, also reduce the nemesis discovery methods needed. Only one method for filtering is now present (`NemesisRegistry.get_disrupt_methods`) and input is a logical phrase. To allow also extracting properties to `nemesis.yaml`/`nemesis_classes.yaml` add `gather_properties`.

`nemesis.yaml` was also changed from list containing strings, to a full on dict. Dict is sorted so the desired output is essentially the same, but it requires less processing to write/read.

All Monkey which only filtered by flags (`LimitedChaosMonkey`, `GeminiNonDisruptiveChaosMonkey`, `GeminiChaosMonkey`, `NetworkMonkey`, `NonDisruptiveMonkey`, `DisruptiveMonkey`, `FreeTierSetMonkey`)
were removed and replaced by `nemesis_selector` usage

# Changelog

* Add `NemesisRegistry` class
  * Responsible for gathering and filtering available disrupt methods
  * Semi-generic, theoretically it could be used on any class that has disrupt method, not only Nemesis
     * Currently only requirement is that is has `disrupt` method
  * Requires no runtime information to initialize
* Limit source code searching to disrupt method
  *  This improvement changes time of  `test_nemesis_sisyphus` from 13 sec to 3 sec, including python env initialization
* Rewrite  `test_nemesis_sisyphus` to demonstrate improvements
* Remove Monkeys that use filtering by flags
* Change `nemesis.yaml` structure into a dict

### Testing

- [x] https://jenkins.scylladb.com/job/scylla-staging/job/phala/job/longevity-schema-changes-3h-test/40/
   - Based on `test-cases/longevity/longevity-sla-100gb-4h` test, as that is one of the test affected by the changes
   - Failed due to a known issue https://github.com/scylladb/scylla-enterprise/issues/2572, but the nemesis was executed
- [x] https://jenkins.scylladb.com/job/scylla-staging/job/phala/job/longevity-100gb-4h-test/10/
   - General longevity to verify Sisyphus correctness

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -30,14 +30,12 @@ import traceback
 import json
 import itertools
 import enum
-import ast
 from contextlib import ExitStack, contextmanager
-from typing import Any, List, Optional, Type, Tuple, Callable, Dict, Set, Union, Iterable
-from functools import wraps, partial, lru_cache
+from typing import Any, List, Optional, Tuple, Callable, Dict, Set, Union, Iterable
+from functools import wraps, partial
 from collections import defaultdict, Counter, namedtuple
 from concurrent.futures import ThreadPoolExecutor
 from threading import Lock
-from types import MethodType  # pylint: disable=no-name-in-module
 
 from cassandra import ConsistencyLevel, InvalidRequest, Unavailable
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
@@ -45,6 +43,7 @@ from cassandra.cluster import NoHostAvailable, OperationTimedOut  # pylint: disa
 from invoke import UnexpectedExit
 from elasticsearch.exceptions import ConnectionTimeout as ElasticSearchConnectionTimeout
 from argus.common.enums import NemesisStatus
+from sdcm.nemesis_registry import NemesisRegistry
 
 from sdcm.utils.cql_utils import cql_unquote_if_needed
 from sdcm import wait
@@ -165,7 +164,6 @@ from test_lib.compaction import CompactionStrategy, get_compaction_strategy, get
 from test_lib.cql_types import CQLTypeBuilder
 from test_lib.sla import ServiceLevel, MAX_ALLOWED_SERVICE_LEVELS
 from sdcm.utils.topology_ops import FailedDecommissionOperationMonitoring
-from sdcm.utils.ast_utils import BooleanEvaluator
 
 
 LOGGER = logging.getLogger(__name__)
@@ -214,7 +212,6 @@ def target_all_nodes(func: Callable) -> Callable:
 
 
 class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-methods
-    DISRUPT_NAME_PREF: str = "disrupt_"
 
     # nemesis flags:
     topology_changes: bool = False  # flag that signal that nemesis is changing cluster topology,
@@ -236,16 +233,10 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
     zero_node_changes: bool = False
 
     def __init__(self, tester_obj, termination_event, *args, nemesis_selector=None, nemesis_seed=None, **kwargs):  # pylint: disable=unused-argument
-        for name, member in inspect.getmembers(self, lambda x: inspect.isfunction(x) or inspect.ismethod(x)):
-            if not name.startswith(self.DISRUPT_NAME_PREF):
-                continue
-            is_exclusive = name in EXCLUSIVE_NEMESIS_NAMES
-            # add "disrupt_method_wrapper" decorator to all methods are started with "disrupt_"
-            setattr(self, name,
-                    MethodType(disrupt_method_wrapper(member, is_exclusive=is_exclusive), self))
-
         # *args -  compatible with CategoricalMonkey
         self.tester = tester_obj  # ClusterTester object
+        self.nemesis_registry = NemesisRegistry(base_class=Nemesis,
+                                                excluded_list=COMPLEX_NEMESIS)
         self.cluster: Union[BaseCluster, BaseScyllaCluster] = tester_obj.db_cluster
         self.loaders = tester_obj.loaders
         self.monitoring_set = tester_obj.monitors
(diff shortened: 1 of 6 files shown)


sdcm/nemesis_registry.py
```diff
@@ -0,0 +1,161 @@
...
+    method_name = DISRUPT_PATTERN.search(inspect.getsource(nemesis_cls.disrupt))
+    if method_name:
+        return method_name.group("method_name")
+
+
+class NemesisRegistry:
+    def __init__(self, base_class: SourceType, kubernetes=False, excluded_list=None):
+        super().__init__()
+        self.base_class = base_class
+        self.kubernetes = kubernetes
```
fruch, line 25: I would take this out, if we want to filter this by default when kubernetes is used, let's add the where we read the filter

it's not a first class variable of this registry, it just one of the flags from it's concern.

sdcm/nemesis_registry.py
```diff
@@ -0,0 +1,161 @@
...
+        tmp = self.base_class.__subclasses__()
+        subclasses = []
+        while tmp:
+            for nemesis in tmp.copy():
+                subclasses.append(nemesis)
+                tmp.remove(nemesis)
+                tmp.extend(nemesis.__subclasses__())
+        return self.filter_subclasses(subclasses, logical_phrase)
+
+    def get_list_of_methods_by_flags(  # noqa: PLR0913
```
fruch, line 65: I hate those functions quite a lot, cause they force us to define the flags multiple times

flag should be defined in one place (mostly by conversion, for now), and documented there.

this repetition is really pointless

fruch (commented):
In general the direction LGTM

I think we can throw away a few more thing while we are at it. (with too much work)

sdcm/nemesis.py
```diff
@@ -529,19 +531,8 @@ def get_list_of_methods_compatible_with_backend(
...
-            kubernetes=self._is_it_on_kubernetes() or None,
-            limited=limited,
-            topology_changes=topology_changes,
-            schema_changes=schema_changes,
-            config_changes=config_changes,
-            free_tier_set=free_tier_set,
-            manager_operation=manager_operation,
-            zero_node_changes=zero_node_changes
+        return self.nemesis_registry.get_list_of_methods_compatible_with_backend(
+            disruptive, supports_high_disk_utilization, run_with_gemini, networking, limited, topology_changes, schema_changes, config_changes, free_tier_set, manager_operation, zero_node_changes
```
vponomaryov, line 535: This approach is fragile, it is easy to mistakenly swap some of the values and we will get unexpected results.
Better to use key=value approach when we have such a big list of params.

sdcm/nemesis.py
```diff
@@ -564,34 +555,12 @@ def get_list_of_methods_by_flags(  # pylint: disable=too-many-locals  # noqa: PL
...
-            limited=limited,
-            topology_changes=topology_changes,
-            schema_changes=schema_changes,
-            config_changes=config_changes,
-            free_tier_set=free_tier_set,
-            sla=sla,
-            manager_operation=manager_operation,
-            zero_node_changes=zero_node_changes
+        return self.nemesis_registry.get_list_of_methods_by_flags(
+            disruptive, supports_high_disk_utilization, run_with_gemini, networking, kubernetes, limited, topology_changes, schema_changes, config_changes, free_tier_set, sla, manager_operation, zero_node_changes
```
vponomaryov, line 559: same here.

sdcm/nemesis.py
pehala: Fair point

sdcm/nemesis_registry.py
```diff
@@ -0,0 +1,161 @@
...
+            zero_node_changes=zero_node_changes,
+        )
+        logical_phrase = " and ".join([key for key, val in args.items() if val])
+        subclasses_list = self.get_subclasses(logical_phrase=logical_phrase)
+
+        disrupt_methods_list = []
+        for subclass in subclasses_list:
+            if method_name := get_disrupt_method_from_class(subclass):
+                disrupt_methods_list.append(method_name)
+        # self.log.debug("Gathered subclass methods: {}".format(disrupt_methods_list))
```
vponomaryov, line 103: left-over?

pehala:
v2 - Drop the backwards compatibility of this patch and simplified the logic:
  * Unify filtering of nemesis
     * Should simplify a lot of unnecessary logic, only filtering method is now `get_disrupt_methods(logical_phrase)`
  * Remove DEPRECATED_NEMESIS
  * Change structure of `nemesis.yml`
     * It is now a dict, instead of list of strings. I think this is easier to parse
  * Make ComplexNemesis, which were only filtering based on flag, inherit from SisyphusMonkey (Needs testing)
     * Reduces code and unifies logic with the `SisyphusMonkey`
  * Remove kubernetes flag from NemesisRegistry, move the logic into `build_list_of_disruptions`

sdcm/nemesis.py
```diff
@@ -7020,7 +6715,7 @@ class FreeTierSetMonkey(SisyphusMonkey):
     def __init__(self, *args, **kwargs):
         # skip SisyphusMonkey __init__ to not repeat build disruption logic, but still we want to run Nemesis class __init__
         super(SisyphusMonkey, self).__init__(*args, **kwargs)  # pylint: disable=bad-super-call
-        self.build_list_of_disruptions_to_execute(nemesis_selector=['free_tier_set'])
+        self.build_list_of_disruptions_to_execute(nemesis_selector='free_tier_set')
```
soyacz, line 6718: why not utilizing `NEMESIS_SELECTOR` class attr (applies also to SlaNemeses, ZeroTokenSetMonkey)?

sdcm/nemesis.py
```diff
@@ -5783,6 +5624,18 @@ def wrapper(*args, **kwargs):  # pylint: disable=too-many-statements  # noqa: PL
     return wrapper
 
 
+class SisyphusMonkey(Nemesis):
+    NEMESIS_SELECTOR = None
```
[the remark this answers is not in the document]
pehala: It is a bug, it should be used in `self.build_list_of_disruptions_to_execute(nemesis_selector=self.NEMESIS_SELECTOR)`.

sdcm/nemesis.py
pehala: It could be used here as well, will fix

sdcm/nemesis.py
```diff
@@ -6213,38 +6066,33 @@ def _random_disrupt(self):
         self.execute_disrupt_method(bound_method)
 
 
-class LimitedChaosMonkey(Nemesis):
-
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-        self.disrupt_methods_list = self.get_list_of_methods_compatible_with_backend(limited=True)
+class LimitedChaosMonkey(SisyphusMonkey):
```
fruch, line 6069: why not throwing those  completely, i.e. no need of this class anymore.

we can put in the configure where it's needed  

```yaml
nemesis_class: 'SisyphusMonkey'
nemesis_filter: 'limited'
```

those classes only hold the doc that explain  it, it can be next to the definition of the flag  

when it's 1:1 match class to flag, I think we can get rid of the class

keep in mind those classes were create before we have any filter capabilities from the configuration.

i.e. they are just filters

sdcm/nemesis.py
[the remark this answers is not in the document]
soyacz: Will it work with parallel nemesis with different selectors when this is a class attr?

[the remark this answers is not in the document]
pehala: It was thought of as a way for subclasses to override this, so I do not think it should meddle with parallel nemesis at all. However, given that another suggestion is to remove all those "filter" classes with a configuration option, I do not think this class variable will be needed.

pehala:
> The `improvement(nemesis): Remove deprecated Nemesis` commit can and should be moved out to a separate PR and be merged.
> The same about the `fix(nemesis.yml): Update nemesis.yml and nemesis_classes.yml`.

I will extract those

> The same about the `fix(nemesis.yml): Update nemesis.yml and nemesis_classes.yml`.
> Can we consider this one (#10502) as a first in the chain and get it merged sooner than later?
> Idea is great, let's not waste time...

We can make this one first, I can edit the rest to rely on this one

pehala:
v3:
 * Removed Nemesis, which filter only by flags
   * LimitedChaosMonkey, GeminiNonDisruptiveChaosMonkey, GeminiChaosMonkey, NetworkMonkey,
NonDisruptiveMonkey, DisruptiveMonkey, FreeTierSetMonkey

next version should be the final after the extracted PR are incorporated and final comments are addressed

pehala:
v4:
 * Rebased on top of master and squashed commits
 * Add docstrings
 
I also updated cover letter and started a test run to verify correctness, as this patch now affects actual testcases. I will take this out of draft, once the run is finished and passing

sdcm/nemesis.py
```diff
@@ -6825,12 +6554,8 @@ def disrupt(self):
         self.disrupt_repair_streaming_err()
 
 
-COMPLEX_NEMESIS = [NoOpMonkey, ChaosMonkey,
-                   LimitedChaosMonkey,
-                   ScyllaCloudLimitedChaosMonkey,
-                   AllMonkey, MdcChaosMonkey,
-                   DisruptiveMonkey, NonDisruptiveMonkey, GeminiNonDisruptiveChaosMonkey,
-                   GeminiChaosMonkey, NetworkMonkey, SisyphusMonkey,
+COMPLEX_NEMESIS = [NoOpMonkey, ChaosMonkey, ScyllaCloudLimitedChaosMonkey,
```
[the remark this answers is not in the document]
pehala: It does not, it seems like `disrupt_run_unique_sequence` was always collected, see https://github.com/scylladb/scylla-cluster-tests/blob/93e7488b578f72478911c0124669a19111f71e95/data_dir/nemesis.yml#L309

sdcm/nemesis.py
```diff
@@ -6970,12 +6680,12 @@ class SlaNemeses(Nemesis):
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.build_list_of_disruptions_to_execute()
-        self.disrupt_methods_list = self.get_list_of_methods_by_flags(sla=True)
+        self.build_list_of_disruptions_to_execute(nemesis_selector="sla")
```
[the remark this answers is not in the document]
pehala: Removed, fun fact: It was never part of `COMPLEX_NEMESIS` and was not collected only because it did not find the disrupt method

sdcm/nemesis.py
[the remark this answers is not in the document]
soyacz: ineed, I see it was run 41 times in 2025.1 testing. But I'm not sure if we should - this is mainly used in perf tests.

pehala:
v5 - Unify disrupt method execution mechanisms:
Currently, two separate mechanisms exist to call disrupt methods:
`build_list_of_disruptions` used by SisyphusMonkey
`call_random_disrupt_method` used by all complex monkeys
These two mechanism differ in yaml properties it supports, in usage of random_seed
and how they discover disrupt methods.
Former using NemesisRegistry, later filtering it directly.

* Change how `build_list_of_disruptions` works
  * Rename it to `build_disruptions_by_selector`
  * Change it to return a list instead of working on `self.disruptions_list`
* Add `build_disruptions_by_name` as a replacement to `call_random_disrupt_method`
* Move `nemesis_multiply_factor` option to `shuffle_list_of_disruptions` instead
   * So it can be used in `build_disruptions_by_name`
* Remove factor scaling based on test duration
   * It seemed inconsistent and in conflict with `nemesis_multiply_factor`
* Complex Nemesis now use  `build_disruptions_by_name` and `call_next_nemesis()` similar to SisyphusMonkey
* Remove `AllMonkey` and `ChaosMonkey`, they were the same as SisyphusMonkey after the changes

This change could be theoretically extracted out of this PR, but it would mean only one of the execution patch would call the new code. I consider ComplexNemesis as mostly outdated.

vponomaryov:
@pehala 
The second commit called `improvement(treewide): Remove Nemesis, which filter by flags` is applicable on the current master as-is and doesn't really depend on the first commit, right?

If so, can you please move this second commit out of this PR to a separate one so we could merge it separately?

pehala:
> The second commit called `improvement(treewide): Remove Nemesis, which filter by flags` is applicable on the current master as-is and doesn't really depend on the first commit, right?
> 
> If so, can you please move this second commit out of this PR to a separate one so we could merge it separately (earlier)?
> 

Will do

> Also, what is expected to be done in scope of this PR to make it be "ready for review" and stop being "work in progress"?

I need to finish testing it to verify I did not break anything. Latest test seemed to be working but I need to dive deeper

pehala:
Tests passed, let me know if you would like more cases tested, so I am making this ready for review

vponomaryov (approved):
LGTM

sdcm/nemesis.py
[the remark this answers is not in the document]
fruch: it's shouldn't, but we never had proper machinery to exclude it from regular nemesis.

maybe it should be put on the complex list, so it can be used only if named directly.

but that's not for this PR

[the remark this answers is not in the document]
fruch: I would recommend raising issue just for that nemesis, and handle it sparately

fruch (approved):
LGTM

sdcm/nemesis.py
```diff
@@ -2073,22 +1866,34 @@ def nemesis_selector(self, value: str):
...
+        """
+        self.log.debug(f'nemesis_seed to be used is {self.nemesis_seed}')
         self.log.debug(f"nemesis stack BEFORE SHUFFLE is {self._disruption_list_names}")
-        random.Random(self.nemesis_seed).shuffle(self.disruptions_list)
+        nemesis_multiply_factor = nemesis_multiply_factor or self.cluster.params.get('nemesis_multiply_factor') or 1
+        random.Random(self.nemesis_seed).shuffle(disruption_list * nemesis_multiply_factor)
         self.log.info(f"List of Nemesis to execute: {self._disruption_list_names}")
 
+    @cached_property
+    def infinite_cycle(self):
```
fruch, line 1887: it's not clear why we didn't apply this long long ago.

every time we hit it, we just enlarge the number on `nemesis_multiply_factor` which is quite silly, since it was meant for the shuffle  and not for  making sure we have nemesis running along the whole test.

fruch (approved):
LGTM

sdcm/nemesis.py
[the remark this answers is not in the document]
soyacz: https://github.com/scylladb/scylla-cluster-tests/issues/10673

soyacz:
> @soyacz and @roydahan Any objections? If not, then let's just merge it

No objections from my side.

PR #10504, opened 2025-03-26
improvement(cql-stress): collect cql-stress and driver version

soyacz:
New versions of cql-stress support `version` arg.
Started collecting cql-stress-cassandra-stress version to Argus.

refs: https://github.com/scylladb/qa-tasks/issues/1829

### Testing

- [x] - [provision test](https://argus.scylladb.com/tests/scylla-cluster-tests/65cf7859-1653-4e8c-a00f-7a3d821c91a9)

### Reminders

### DIFF
--- a/sdcm/reporting/tooling_reporter.py
+++ b/sdcm/reporting/tooling_reporter.py
@@ -1,3 +1,4 @@
+import json
 import logging
 from functools import lru_cache
 
@@ -11,14 +12,14 @@ LOGGER = logging.getLogger(__name__)
 
 
 @lru_cache
-def report_package_to_argus(client: ArgusSCTClient, tool_name: str, package_version: str, additional_data: str = None):
+def report_package_to_argus(client: ArgusSCTClient, tool_name: str, package_version: str, additional_data: str = None,
+                            date: str = None, revision_id: str = None) -> None:
     package = Package(name=f"{tool_name}", version=package_version,
-                      date=None, revision_id=None, build_id=additional_data)
+                      date=date, revision_id=revision_id, build_id=additional_data)
     client.submit_packages([package])
 
 
 class ToolReporterBase():
-
     TOOL_NAME = None
 
     def __init__(self, runner: CommandRunner, command_prefix: str = None, argus_client: ArgusSCTClient = None) -> None:
@@ -27,6 +28,8 @@ class ToolReporterBase():
         self.argus_client = argus_client
         self.additional_data = None
         self.version: str | None = None
+        self.date: str | None = None
+        self.revision_id: str | None = None
 
     def __str__(self) -> str:
         return f"{self.__class__.__name__}()"
@@ -39,7 +42,8 @@ class ToolReporterBase():
             LOGGER.warning("%s: Skipping reporting to argus, client not initialized.", self)
             return
         try:
-            report_package_to_argus(self.argus_client, self.TOOL_NAME, self.version, self.additional_data)
+            report_package_to_argus(client=self.argus_client, tool_name=self.TOOL_NAME, package_version=self.version,
+                                    additional_data=self.additional_data, date=self.date, revision_id=self.revision_id)
         except Exception:  # pylint: disable=broad-except # noqa: BLE001
             LOGGER.warning("Failed reporting tool version to Argus", exc_info=True)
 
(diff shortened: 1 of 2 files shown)


fruch:
@soyacz how come the date and revision are empty?

fruch:
> @soyacz how come the date and revision are empty?

Now I see that the base code is passing both with None

We should fix that

soyacz:
> > @soyacz how come the date and revision are empty?
> 
> Now I see that the base code is passing both with None
> 
> We should fix that

fixed
@dimakr you can base on this PR with s-b version

sdcm/reporting/tooling_reporter.py
```diff
@@ -98,6 +102,31 @@ def _collect_version_info(self) -> None:
...
+        result = json.loads(output.stdout)
+
+        LOGGER.info("Result:\n%s", result)
+        cql_stress_details = result.get('cql-stress', {})
+        self.version = f"{cql_stress_details.get('version', '#FAILED_CHECK_LOGS')}"
+        self.date = cql_stress_details.get('commit_date')
+        self.revision_id = cql_stress_details.get('commit_sha')
+
+        if driver_details := result.get("scylla-driver", {}):
+            self.additional_data = f"rust-driver: {driver_details['version']}"
```
vponomaryov, line 121: Need to make the record be specific.
We may have more than 1 record with rust driver - one from `cql-stress` and another one from `latte`.
And those may have different versions at some point in time.

sdcm/reporting/tooling_reporter.py
```diff
@@ -98,6 +102,31 @@ def _collect_version_info(self) -> None:
                 driver_version=driver_version, command_prefix=None, runner=None, argus_client=self.argus_client).report()
 
 
+class CqlStressCassandraStressVersionReporter(ToolReporterBase):
+    # pylint: disable=too-few-public-methods
+    TOOL_NAME = "cql-stress-cassandra-stress"
+
+    def _collect_version_info(self) -> None:
+        output = self.runner.run(f"{self.command_prefix} {self.TOOL_NAME} version_json")
+        LOGGER.info("%s: Collected cql-stress-cassandra-stress output:\n%s", self, output.stdout)
```
vponomaryov, line 111: Maybe change it to the `debug` level?
We may have dozens of stress commands...

sdcm/reporting/tooling_reporter.py
```diff
@@ -98,6 +102,31 @@ def _collect_version_info(self) -> None:
...
+        result = json.loads(output.stdout)
+
+        LOGGER.info("Result:\n%s", result)
```
vponomaryov, line 114: Same here.

sdcm/reporting/tooling_reporter.py
```diff
@@ -108,3 +137,18 @@ def __init__(self, driver_version: str, runner: CommandRunner, command_prefix: s
 
     def _collect_version_info(self) -> None:
         pass
+
+
+class CqlStressRustDriverVersionReporter(ToolReporterBase):
+    # pylint: disable=too-few-public-methods
+    TOOL_NAME = "rust-driver"
```
vponomaryov, line 144: It must support some suffix or prefix tweaking for distinguishing rust driver versions from `cql-stress` and `latte`.

soyacz: I didn't think about Latte here - see the name of the class `CqlStressRustDriverVersionReporter`
When introducing it for Latte, one may use this class as a base

sdcm/reporting/tooling_reporter.py
soyacz: I didn't think about it - just copied over from other implementations.
Sounds it should be debug - as anyway we use Argus for getting this info.

sdcm/reporting/tooling_reporter.py
soyacz: Let's refactor this when Latte is implemented

fruch: `additional_data` in this case is only for cql-stress, @vponomaryov  not sure what's the concern here...

sdcm/reporting/tooling_reporter.py
fruch: those should be call once per command, so yes debug should be good enough

sdcm/reporting/tooling_reporter.py
fruch: for c-s it was `TOOL_NAME = "cassandra-stress"`

so this should follow the same pattern

yes each one of them might report about multiple packages, but it's per tool

sdcm/reporting/tooling_reporter.py
soyacz: fixed

sdcm/reporting/tooling_reporter.py
soyacz: fixed

sdcm/reporting/tooling_reporter.py
soyacz: I renamed `CqlStressRustDriverVersionReporter` -> `RustDriverVersionReporter`

vponomaryov:
I was referring to the following situation:

Screenshot from 2025-03-31 13-54-32

Since rust-driver is not standalone package, but used as part of a stress tool, we should not place it in the same hierchy level as stress tools.

soyacz:
> Since rust-driver is not standalone package, but used as part of a stress tool, we should not place it in the same hierchy level as stress tools.
> 
> Having it be in the "build ID' section for a stress tool - ok. But not as it is in the red wrapping.

We do it since it was introduced. What do you propose?

soyacz:
> > Since rust-driver is not standalone package, but used as part of a stress tool, we should not place it in the same hierchy level as stress tools.
> > Having it be in the "build ID' section for a stress tool - ok. But not as it is in the red wrapping.
> 
> We do it since it was introduced. What do you propose?

I mean it is going to be tracked for rust driver tests - we use similar approach for Java driver. It then is shown also on graphs as part of changed package to highlight this important change.

vponomaryov:
> > Since rust-driver is not standalone package, but used as part of a stress tool, we should not place it in the same hierchy level as stress tools.
> > Having it be in the "build ID' section for a stress tool - ok. But not as it is in the red wrapping.
> 
> We do it since it was introduced. What do you propose?

Add stress tool ref to the "build id" column for the "rust driver" record?
Print different records for "rust driver" in case we use 2 tools and they use different driver versions?

soyacz:
> > > Since rust-driver is not standalone package, but used as part of a stress tool, we should not place it in the same hierchy level as stress tools.
> > > Having it be in the "build ID' section for a stress tool - ok. But not as it is in the red wrapping.
> > 
> > 
> > We do it since it was introduced. What do you propose?
> 
> Add stress tool ref to the "build id" column for the "rust driver" record? Print different records for "rust driver" in case we use 2 tools and they use different driver versions?
> 
> My main point is to be sure which tool uses which driver version - explicit guaranteed mapping.

Ok, now I get it.
Although we can have 2 same packages like rust driver in packages list I think we should differentiate them. My proposal is then to have `cql-stress-rust-driver` and drop any reference from Build ID (as this is not intended for that info anyway)

vponomaryov:
> ...
> My proposal is then to have `cql-stress-rust-driver` ...

Works for me.
In latte case, using this approach, it will be `latte-rust-driver`.

JFYI: Latte already supports versions command like CS and cql-stress do. So, I just wait for this one to be finished.

vponomaryov (approved):
LGTM

fruch (approved):

PR #10506, opened 2025-03-27
fix(event_counter): switch to thread base class

fruch:
using it as a process is forcing us to use multiprocess sync manager which in turn can leave running processes until the test end

there no real reason for using a process with this one so switching it to the thread base implementation

### Testing

- [x] integration tests
- [x] provision tests

### Reminders


aleksbykov (approved):
LGTM

Issue #9784, opened 2025-01-12
artifacts-tests are cordumping during SCT shutdown


yaronkaikov:
Although the test passed, 
```
07:56:40  < t:2025-01-12 05:56:40,267 f:setup.py        l:89   c:sdcm.sct_events.setup p:INFO  > }
07:56:40  < t:2025-01-12 05:56:40,268 f:tester.py       l:3088 c:ArtifactsTest        p:INFO  > ================================= TEST RESULTS =================================
07:56:40  < t:2025-01-12 05:56:40,268 f:tester.py       l:3088 c:ArtifactsTest        p:INFO  > ================================================================================
07:56:40  < t:2025-01-12 05:56:40,268 f:tester.py       l:3088 c:ArtifactsTest        p:INFO  > SUCCESS :)
07:56:41  < t:2025-01-12 05:56:41,597 f:base.py         l:274  c:elasticsearch        p:INFO  > GET https://746f0ad652a3447d83b1572f657c67cb.us-east-1.aws.found.io/ [status:200 request:0.077s]
07:56:41  < t:2025-01-12 05:56:41,620 f:base.py         l:274  c:elasticsearch        p:INFO  > HEAD https://746f0ad652a3447d83b1572f657c67cb.us-east-1.aws.found.io/sct_test_runs [status:200 request:0.022s]
07:56:41  < t:2025-01-12 05:56:41,717 f:base.py         l:274  c:elasticsearch        p:INFO  > PUT https://746f0ad652a3447d83b1572f657c67cb.us-east-1.aws.found.io/sct_test_runs/sct_test_run_short_v1/f95ee4d6-1636-4033-9b2c-d6493cc90f59/_create 
```

and then the test coredump during shut down
```
07:56:46  /bin/bash: line 1:     7 Segmentation fault      (core dumped) ./sct.py run-test artifacts_test --backend gce --logdir /tmp/workspace/enterprise-2024.2/artifacts-offline-install/artifacts-debian11-nonroot-test/scylla-cluster-tests
```

#### Logs

* **db-cluster** - https://cloudius-jenkins-test.s3.amazonaws.com/f95ee4d6-1636-4033-9b2c-d6493cc90f59/20250112_055740/db-cluster-f95ee4d6.tar.gz
[...]

PR #10518, opened 2025-03-28
fix(shutdown): stop argus heartbeat thread quicker

soyacz:
We face issues with python coredumping upon SCT shutdown. One of the reason might be running argus heartbeat thread which is not awaited to complete (and it may take up to 30 secs to end).

Fix by quittin it as soon as shutdown is initiated without waiting for sleep to end.

refs: https://github.com/scylladb/scylla-cluster-tests/issues/9784

### Testing

- [x] - provision test

### Reminders


fruch (approved):
LGTM

PR #10335, opened 2025-03-10
fix(stress/base.py): refactor stress thread verify_results API

dimakr:
This is a follow up for the changes introduced in https://github.com/scylladb/scylla-cluster-tests/pull/9585.

`DockerBasedStressThread.verify_results` is not actually verifying anything, but is a duplicate of `DockerBasedStressThread.get_results` method. This change refactors the `verify_results` method and its usage in SCT:
- renames the method to `parse_results` to properly reflect its purpose
- modifies the implementation to allow reuse by other child stress thread classes - CassandraStressThread and ScyllaBenchThread no longer require their own implementations of this method

Additionally, the change improves the logic of aborting a nemesis on stress thread failures:
- removes the dedicated `Nemesis.stop_nemesis_on_stress_errors` failures handler
- introduces an optional `error_handler` parameter in `tester.py:ClusterTester.verify_stress_thread`
- updates nemesis.py to reuse the `verify_stress_thread` with the dedicated error handler, for aborting nemesis on stress failures

### Testing

Some nemeses that cover the updates in s-b, c-s threads, in tester.py ()
- [x] [TruncateMonkey, both c-s and s-b stress threads are executed during the nemesis](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-aws-test/50/parameters/)
- [x] [GrowShrinkClusterNemesis, both c-s and s-b stress threads are executed during the nemesis](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-5gb-1h-aws-test/52/parameters/)
- [x] [Kafka consumer test with python stress thread ](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/longevity-kafka-cdc-aws-test/4/)

PR-provision-test configuration with faulty c-s command (manually added non-existing parameter to the command) in _prepare_test_table, that causes abort of the nemesis:
was failing:
```
< t:2025-03-15 17:14:41,117 f:nemesis.py      l:5661 c:sdcm.nemesis         p:ERROR > sdcm.exceptions.NemesisStressFailure: Aborting 'NonDisruptiveMonkey' nemesis as stress command failed with the following errors:
< t:2025-03-15 17:14:41,117 f:nemesis.py      l:5661 c:sdcm.nemesis         p:ERROR >  on node 'PR-provision-test-dmitriy-loader-node-adf836a4-1': ['Stress command completed with bad status 1: ']
```

[...]

PR #10526, opened 2025-03-30
fix(double_cluster): stress command was waited incorrectly

fruch:
nemesis case was calling `_nemesis_stress_failure_handler` with the wrong arguments, hence failing like:

```
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5688, in wrapper
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4384, in disrupt_grow_shrink_cluster
self._double_cluster_load(duration)
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 224, in wrapped
res = func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4370, in _double_cluster_load
self._nemesis_stress_failure_handler(results, errors)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2139, in _nemesis_stress_failure_handler
if len(errors) == len(stress_pool.get_results()):
AttributeError: 'list' object has no attribute 'get_results'
```

this commit change it to be used as a callback to `verify_stress_thread` call. as `_nemesis_stress_failure_handler` was designed.

Ref: #10335

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -4366,8 +4366,8 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         self.log.info("Doubling the load on the cluster for %s minutes", duration)
         stress_queue = self.tester.run_stress_thread(
             stress_cmd=self.tester.stress_cmd, stress_num=1, stats_aggregate_cmds=False, duration=duration)
-        results, errors = stress_queue.parse_results()
-        self._nemesis_stress_failure_handler(results, errors)
+        results = self.tester.verify_stress_thread(thread_pool=stress_queue,
+                                                   error_handler=self._nemesis_stress_failure_handler)
         self.log.info(f"Double load results: {results}")
 
     @target_data_nodes


fruch:
@soyacz, just validating, the idea in that nemesis is to fail nemesis right away if stress command fails ?

dimakr (approved):
LGTM

soyacz:
> @soyacz, just validating, the idea in that nemesis is to fail nemesis right away if stress command fails ?

Hmm, that would depend on the context: In elasticity perf test - I'd say yes, but in other cases we might just raise error message and continue with shrink cluster to progress to next nemesis.
I'd say we need to catch `NemesisStressFailure` or implement a flag for `_nemesis_stress_failure_handler` to raise error event instead raising exception in some situations.

fruch:
> > > > @soyacz, just validating, the idea in that nemesis is to fail nemesis right away if stress command fails ?
> > > 
> > > 
> > > Hmm, that would depend on the context: In elasticity perf test - I'd say yes, but in other cases we might just raise error message and continue with shrink cluster to progress to next nemesis. I'd say we need to catch `NemesisStressFailure` or implement a flag for `_nemesis_stress_failure_handler` to raise error event instead raising exception in some situations.
> > 
> > 
> > that work for followup, the current situation it would fail
> > anyhow not sure if we have a clear way to distinguish perf tests from others
> 
> we don't have to distinguish perf test - we could assume we always want to proceed

anyhow the default is not to double the cluster load... I would leave it like that for now

PR #10535, opened 2025-03-31
fix(tearDown): Close DECODING_QUEUE

fruch:
if we don't close it, there a left thread `QueueFeederThread` at the end of the test, we want to eliminate of the threads/proccess left at the end of the test, to avoid any deadlock of other issue that seem to be coredump the end of the end from time to time

Ref: https://github.com/scylladb/scylla-cluster-tests/issues/9784

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -3050,6 +3050,10 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         self.stop_resources()
         self.get_test_failures()
 
+        with silence(parent=self, name='closing decoding queue as needed'):
+            if self.test_config.BACKTRACE_DECODING:
+                self.test_config.DECODING_QUEUE.close()
+
         # NOTE: running on K8S we need to gather logs otherwise a lot of
         # debugging info is lost
         if self.k8s_clusters:


sdcm/tester.py
```diff
@@ -3050,6 +3050,11 @@ def tearDown(self):
         self.stop_resources()
         self.get_test_failures()
 
+        with silence(parent=self, name='closing decoding queue as needed'):
+            self.test_config.BACKTRACE_DECODING = self.params.get('backtrace_decoding')
```
soyacz, line 3054: this line looks redundant

fruch: right, I wasn't reading it carefully enough...

sdcm/tester.py
```diff
@@ -3050,6 +3050,10 @@ def tearDown(self):
         self.stop_resources()
         self.get_test_failures()
 
+        with silence(parent=self, name='closing decoding queue as needed'):
+            if self.test_config.BACKTRACE_DECODING:
+                self.test_config.DECODING_QUEUE.close()
```
vponomaryov, line 3055: We probably want to wrap up this line into try-except block.
We should not fail whole tearDown only if this line fails.

fruch: > We probably want to wrap up this line into try-except block.
> We should not fail whole tearDown only if this line fails.

That's what silence is doing, plus it would generate an event if it would fail

Issue #9443, opened 2024-12-02
Add nemesis for sending SIGQUIT to Scylla


pehala:
With scylladb#21692, scylla should now dump diagnostics upon receiving SIGQUIT, we should add a nemesis which verifies this behaviour

soyacz:
@pehala why nemesis and not as dtes/test.py?

pehala:
It was suggested by Roy https://github.com/scylladb/scylladb/pull/21692#issuecomment-2502341086

soyacz:
> It was suggested by Roy [scylladb/scylladb#21692 (comment)](https://github.com/scylladb/scylladb/pull/21692#issuecomment-2502341086)

I see it was suggested more by Botond to test this path. I'm not sure if we want to add new nemesis - maybe just one nemesis that already do restart could do it? e..g `_destroy_data_and_restart_scylla` instead of simple `stop_scylla_server` could do this one (and maybe just once per test - as it will generate coredump IIRC)?

PR #10536, opened 2025-03-31
test(collect_diagnosis_data_test): Add ScyllaDiagnosisReport nemesis

Lakshmipathi:
Adds a new ScyllaDiagnosisReport nemesis to send signal to scylladb to dump diagnostics and verify it.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/9443

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/81aeba13-f6d3-4801-8440-0ba292463abe

### Reminders

### DIFF
--- a/collect_diagnosis_data_test.py
+++ b/collect_diagnosis_data_test.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2025 ScyllaDB
+
+from sdcm.tester import ClusterTester
+from sdcm import nemesis
+
+
+class ScyllaDiagnosisReport(ClusterTester):
+
+    def test_diagnosis_data(self):  # pylint: disable=invalid-name
+        current_nemesis = nemesis.ScyllaDiagnosisReport(
+            tester_obj=self, termination_event=self.db_cluster.nemesis_termination_event)
+        current_nemesis.disrupt()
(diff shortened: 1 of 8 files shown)


Lakshmipathi:
_Argus reports success but jenkins job reports failure. Will fix and remove draft tag from this_

sdcm/nemesis.py
```diff
@@ -1744,6 +1756,14 @@ def wait_for_old_node_to_removed():
     def disrupt_kill_scylla(self):
         self._kill_scylla_daemon()
 
+    def disrupt_sigquit_scylla(self):
+        stress_cmd = self.tester.params.get('stress_cmd')
+        self.tester.run_stress_thread(
+            stress_cmd=stress_cmd, stress_num=1, stats_aggregate_cmds=False)
```
soyacz, line 1762: nemesis rather don't run stress_cmd - unless it is by specific design (like doubling load)

collect_diagnosis_data_test.py
```diff
@@ -0,0 +1,25 @@
...
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2025 ScyllaDB
+
+from sdcm.tester import ClusterTester
+from sdcm import nemesis
+
+
+class ScyllaDiagnosisReport(ClusterTester):
```
soyacz, line 20: usually we don't add new tests - only in specific cases. 
For this should utilize 'individual nemesis' case

soyacz (commented):
maybe instead of nemesis it should be teradown validator?

sdcm/nemesis.py
```diff
@@ -703,6 +703,18 @@ def _kill_scylla_daemon(self):
             self.target_node.wait_jmx_up()
         self.cluster.wait_for_schema_agreement()
 
+    def _sigquit_scylla_daemon(self):
```
soyacz, line 706: btw. we have already `generate_coredump_file` method in `BaseNode` class - this could reuse your new method and print diagnosis dump. 
Also add node log message with `DbNodeLogger` about this signal

Lakshmipathi: Updated this and other comments in new PR https://github.com/scylladb/scylla-cluster-tests/pull/10536

Lakshmipathi:
> maybe instead of nemesis it should be teradown validator?

Ok, converted this nemesis as a teardown validator https://github.com/scylladb/scylla-cluster-tests/pull/10536  and closing this PR in favor of validator

Issue #10534, opened 2025-03-31
Commands, which do not interact with cloud, require OKTA


pehala:
Hydra/sct.py commands like create-nemesis-pipelines, create-test-release-jobs, nemesis-list require OKTA login, even though for their working OKTA is not required.

This is caused by https://github.com/scylladb/scylla-cluster-tests/blob/70f3e92c58192dba8fcdffa8b7a7300cd478f094/sct.py#L195, which indiscriminately checks okta for every CLI command.

fruch:
it's only 3 commands,  (create-test-release-jobs needs aws access to work)

this should do the trick:
```diff
diff --git a/sct.py b/sct.py
index 126d108ec..547111874 100755
--- a/sct.py
+++ b/sct.py
@@ -190,16 +190,19 @@ class SctLoader(unittest.TestLoader):
               type=click.Path(),
               expose_value=False,
               help="Install paths for extra python packages to install, scylla-cluster-plugins for example")
-def cli():
+@click.pass_context
+def cli(ctx):
     disable_loggers_during_startup()
-    try_auth_with_okta()

-    key_store = KeyStore()
-    # TODO: still leaving old keys, until we'll rebuild runner images - and reconfigure jenkins
-    key_store.sync(keys=['scylla-qa-ec2', 'scylla-test', 'scylla_test_id_ed25519'],
-                   local_path=Path('~/.ssh/').expanduser(), permissions=0o0600)
+    if ctx.invoked_subcommand not in ("update-conf-docs", "nemesis-list", "create-nemesis-pipelines"):
+        try_auth_with_okta()
+
+        key_store = KeyStore()
+        # TODO: still leaving old keys, until we'll rebuild runner images - and reconfigure jenkins
+        key_store.sync(keys=['scylla-qa-ec2', 'scylla-test', 'scylla_test_id_ed25519'],
+                       local_path=Path('~/.ssh/').expanduser(), permissions=0o0600)

-    docker_hub_login(remoter=LOCALRUNNER)
+        docker_hub_login(remoter=LOCALRUNNER)


 @cli.command('provision-resources', help="Provision resources for the test")
```

PR #10542, opened 2025-04-01
fix(sct): Dont require Okta for commands which do not need it

pehala:
* update-conf-docs, nemesis-list, create-nemesis-pipelines
* This is a hotfix and it is ugly, but for 3 commands no bigger effort is needed
* All the code was actually suggested and written by @fruch in https://github.com/scylladb/scylla-cluster-tests/issues/10534#issuecomment-2767510477

### Testing

- [x] locally

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -190,16 +190,19 @@ class SctLoader(unittest.TestLoader):
               type=click.Path(),
               expose_value=False,
               help="Install paths for extra python packages to install, scylla-cluster-plugins for example")
-def cli():
+@click.pass_context
+def cli(ctx):
     disable_loggers_during_startup()
-    try_auth_with_okta()
+    # Ugly way of filtering the few command that do not require OKTA verification
+    if ctx.invoked_subcommand not in ("update-conf-docs", "nemesis-list", "create-nemesis-pipelines"):
+        try_auth_with_okta()
 
-    key_store = KeyStore()
-    # TODO: still leaving old keys, until we'll rebuild runner images - and reconfigure jenkins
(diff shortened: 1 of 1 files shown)


fruch:
ugly is the new beautiful

doing better, is to completely regroup them differently, and that can't be done without change the command line, which I we don't want to do for this reason.

 maybe we would be able to split and decorate each command with something, but that would also gonna be tricky and error prone (everyone adding a command would need to know all of the dependencies)

fruch:
one more option, would be to move those commands out of sct.py, to their specific cli

which would be called like:

`hydra python nemesi_cli.py ...`

yet again, changing how the cli is gonna be called.

pehala:
Yeah I agree, I also thought of some solutions, but none of them seemed to be worth it for this small change.

fruch:
from new addition of command, we could consider separate modules

and that's looks interesting way add skip-able parts
https://stackoverflow.com/questions/50883972/click-how-do-i-apply-an-action-to-all-commands-and-subcommands-but-allow-a-comm

pehala:
> from new addition of command, we could consider separate modules
> 
> and that's looks interesting way add skip-able parts https://stackoverflow.com/questions/50883972/click-how-do-i-apply-an-action-to-all-commands-and-subcommands-but-allow-a-comm

Looks interesting, but I still feel this is too big of a solution for 3 commands, for a part of code which does not see many changes. I would keep the current solution and if we need to touch commands again in a bigger way we can include it there.

vponomaryov (approved):

PR #10574, opened 2025-04-02
improvement(nemesis): Remove deprecated Nemesis

pehala:
### Testing

- [x] None, as this is just removal of unused nemesis

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -71,7 +71,6 @@ from sdcm.log import SDCMAdapter
 from sdcm.logcollector import save_kallsyms_map
 from sdcm.mgmt.common import TaskStatus, ScyllaManagerError, get_persistent_snapshots
 from sdcm.nemesis_publisher import NemesisElasticSearchPublisher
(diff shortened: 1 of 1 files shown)


sdcm/nemesis.py
```diff
@@ -6941,8 +6825,6 @@ def disrupt(self):
         self.disrupt_repair_streaming_err()
```
vponomaryov, line 6827: The `UpgradeNemesisOneNode` is used in the `upgrade_schema_test.py` module.

But, that module is not used, so, since py code checkers don't complain we should be ok.

vponomaryov (approved):
LGTM

vponomaryov:
Context for reviewers:
- https://github.com/scylladb/scylla-cluster-tests/pull/10502

sdcm/nemesis.py
pehala: Given that it has not been touched since basically 2017, can I remove it as well? I also do not see any usages of it

fruch: Yes

fruch (approved):

PR #10525, opened 2025-03-30
feature(cql-stress): add support for hdr histogram

soyacz:
Since cql-stress now supports hdr histogram, it can be enabled. cqlstress thread will respect `use_hdrhistogram` config param.

refs: https://github.com/scylladb/qa-tasks/issues/1861

### Testing

- [x] - [perf upgrade test](https://argus.scylladb.com/tests/scylla-cluster-tests/5eea6056-c024-497c-9cc8-5b44a0406a7b)
- [x] - [small elasticity test](https://argus.scylladb.com/tests/scylla-cluster-tests/db02a215-add1-4c1b-a2a0-5b9037a08e14)

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #10582, opened 2025-04-03
fix(perf-regression-test): parse stress results before pushing to ES

dimakr:
Recent refactor of the stress threads API renamed public methods to better align with their function.
This change fixes the API usage to ensure stress results are correctly parsed before being pushed to ElasticSearch.

### Testing

Tested by local run, but also the following job is running
- [x] :green_circle: [mini-elasticity-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/mini-elasticity-test/6/)

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -2368,7 +2368,7 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):  # pylint: disa
         return results and not errors
 
     def get_stress_results(self, queue, store_results=True) -> list[dict | None]:
-        results = queue.get_results()
+        results = queue.parse_results()[0]
         if store_results and self.create_stats:
             self.update_stress_results(results)
         return results


sdcm/tester.py
```diff
@@ -2368,7 +2368,7 @@ def verify_stress_thread(self, thread_pool: list, error_handler: callable = None
         return results and not errors
 
     def get_stress_results(self, queue, store_results=True) -> list[dict | None]:
-        results = queue.get_results()
+        results = queue.parse_results()[0]
```
soyacz, line 2371: results shouldn't be a list? what we truncate here?

dimakr: Results is a list. But in the latest refactor with aligned the stress threads API. Now `get_results` returns raw results (previously it was parsing results for some stress types, but not for others).
We added new `parse_results` API, which returns 2 items - parsed results of successfully finished stress threads and list of threads with errors (previously get_results was just dropping errors).

soyacz:
you'll need to rerun this test after merging https://github.com/scylladb/scylla-cluster-tests/pull/10525 - we didn't get to proper step due error I think

dimakr:
> you'll need to rerun this test after merging #10525 - we didn't get to proper step due error I think

@soyacz after rebase the test run that tests this PR's change is successful

soyacz (approved):
LGTM

soyacz:
@juliayakovlev we  need one more approval to fix running perf tests from master branch.

sdcm/tester.py
juliayakovlev, line 2371: Maybe it worth to check if the result is not empty list. It may fail the test

juliayakovlev (changes requested):
please, add list validation

soyacz:
> please, add list validation

@juliayakovlev parse_results returns tuple of `results, errors` so `[0]` refers to results list

sdcm/tester.py
dimakr: parse_results() returns tuple[list, dict], so there will always 2 items in the method result.

juliayakovlev (approved):

Issue #10581, opened 2025-04-02
using "--table-options" results with "invalid table option","option":"gc_grace_seconds=60--level=info","error":"invalid table option, exactly two parts separated by '=' is needed


yarongilor:
## Packages

Scylla version: `2023.1.11-20240729.5a79e79a0320` with build-id `4daf2e1487b1ab784ff564a6c8fd75f9ddd8a9ac`

Kernel Version: `5.15.0-1066-aws`

## Issue description
The SCT yaml had:
```
gemini_table_options:
  - "gc_grace_seconds=60"
```

Then it got the following Gemini error:
```
2025-04-01 11:15:41.043 <2025-04-01 11:15:40.523>: (GeminiStressLogEvent Severity.WARNING) period_type=one-time event_id=6432c768-460c-4acf-9791-5f8f625212fd: type=GeminiEvent line_number=1 node=Node gemini-tombstones-gemini-t-loader-node-3fce66f4-1 [34.244.167.74 | 10.4.0.115]
line=invalid table option (option="gc_grace_seconds=60--level=info" error="invalid table option, exactly two parts separated by '=' is needed, input=gc_grace_seconds=60--level=info" errorVerbose="invalid table option, exactly two parts separated by '=' is needed, input=gc_grace_seconds=60--level=info
github.com/scylladb/gemini/pkg/tableopts.FromCQL
/home/runner/work/gemini/gemini/pkg/tableopts/options.go:50
github.com/scylladb/gemini/pkg/tableopts.CreateTableOptions
/home/runner/work/gemini/gemini/pkg/tableopts/options.go:74
main.createDefaultSchemaConfig
/home/runner/work/gemini/gemini/cmd/gemini/schema.go:72
main.createSchemaConfig
/home/runner/work/gemini/gemini/cmd/gemini/schema.go:28
main.run
/home/runner/work/gemini/gemini/cmd/gemini/root.go:191
github.com/spf13/cobra.(*Command).execute
/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:985
github.com/spf13/cobra.(*Command).ExecuteC
/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1117
github.com/spf13/cobra.(*Command).Execute
/home/runner/go/pkg/mod/github.com/spf13/cobra@v1.8.1/command.go:1041
main.main
/home/runner/work/gemini/gemini/cmd/gemini/main.go:33
[...]

PR #10583, opened 2025-04-03
fix(gemini_thread.py): fix appending of gemini default flags

yarongilor:
The default flags were wrongly concatenated without a leading space to the gemini command string
Fixes: #10581

### Testing

- [test table options](https://argus.scylladb.com/tests/scylla-cluster-tests/0029ed86-9610-465a-807c-f860830b25c0)

### Reminders

### DIFF
--- a/sdcm/gemini_thread.py
+++ b/sdcm/gemini_thread.py
@@ -143,11 +143,9 @@ class GeminiStressThread(DockerBasedStressThread):  # pylint: disable=too-many-i
 
         stress_cmd = self.stress_cmd.replace("\n", " ").strip()
 
-        for key, value in self.gemini_default_flags.items():
-            if not key in stress_cmd:
-                cmd += f"--{key}={value} "
+        cmd += " " + " ".join(f"--{key}={value}" for key, value in self.gemini_default_flags.items() if
+                              key not in stress_cmd) + " " + stress_cmd
 
-        cmd += stress_cmd
         self.gemini_commands.append(cmd)
         return cmd
 
--- a/unit_tests/test_gemini_thread.py
+++ b/unit_tests/test_gemini_thread.py
@@ -31,6 +31,7 @@ class DBCluster:  # pylint: disable=too-few-public-methods
 
 
 def test_01_gemini_thread(request, docker_scylla, params):
+    params.update({'gemini_table_options': ['gc_grace_seconds=60']})
     loader_set = LocalLoaderSetDummy(params=params)
     test_cluster = DBCluster([docker_scylla])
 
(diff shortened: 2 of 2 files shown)


pehala:
Please add backport label and also include a test run so we know it works

fruch:
> A side note is it's not clear why does SCT have to set all these default parameters of:
> 
> ```
>             "level": "info",
>             "request-timeout": "60s",
>             "connect-timeout": "60s",
>             "consistency": "QUORUM",
>             "async-objects-stabilization-backoff": "1s",
>             "async-objects-stabilization-attempts": 10,
>             "max-mutation-retries-backoff": "1s",
>             "max-mutation-retries": 10,
>             "dataset-size": "large",
>             "oracle-host-selection-policy": "token-aware",
>             "test-host-selection-policy": "token-aware",
>             "drop-schema": "true",
>             "cql-features": "normal",
>             "fail-fast": "true",
>             "materialized-views": "false",
>             "use-server-timestamps": "true",
>             "use-lwt": "false",
>             "use-counters": "false",
>             "max-tables": 1,
>             "max-columns": 16,
>             "min-columns": 8,
>             "max-partition-keys": 6,
>             "min-partition-keys": 2,
>             "max-clustering-keys": 4,
>             "min-clustering-keys": 2,
>             "partition-key-distribution": "normal",  # Distribution for hitting the partition
>             # These two are used to control the memory usage of Gemini
>             "token-range-slices": 512,  # Number of partitions
>             "partition-key-buffer-reuse-size": 100,  # Internal Channel Size per parittion value generation
>             "statement-log-file-compression": "zstd",
> ```
> 
> for Gemini. Are there no defaults for Gemini? or are they different? why setting such a small value of:`"token-range-slices": 512, # Number of partitions`?

cause too much of it, gemini can run out of memory and crash.

as for the default, it was cause too many times the default were changed without syncing with what SCT does or needs.
maybe excessive, but they are explicit.

sdcm/gemini_thread.py
```diff
@@ -143,11 +143,11 @@ def _generate_gemini_command(self):
 
         stress_cmd = self.stress_cmd.replace("\n", " ").strip()
 
-        for key, value in self.gemini_default_flags.items():
-            if not key in stress_cmd:
-                cmd += f"--{key}={value} "
+        cmd = " ".join(
```
pehala, line 146: `cmd = " ".join(f"--{key}={value}" for key, value in self.gemini_default_flags.items() if key not in stress_cmd) + stress_cmd`

I would use something like this

yarongilor: already fixed

yarongilor: ok

yarongilor:
> Add unit/integration tests which cover this fix, which looks to be broken too.

@vponomaryov , do you mean something like this:
```
def test_01_gemini_thread(request, docker_scylla, params):

+    params.update({'gemini_table_options': ['gc_grace_seconds=60']})

+    expected_gemini_arguments = " --table-options=\"gc_grace_seconds=60\" --level=info --request-timeout=60s --connect-timeout=60s --consistency=QUORUM --async-objects-stabilization-backoff=1s --async-objects-stabilization-attempts=10 --dataset-size=large --oracle-host-selection-policy=token-aware --test-host-selection-policy=token-aware --drop-schema=true --fail-fast=true --materialized-views=false --use-lwt=false --use-counters=false --max-tables=1 --max-columns=16 --min-columns=8 --max-partition-keys=6 --min-partition-keys=2 --max-clustering-keys=4 --min-clustering-keys=2 --partition-key-distribution=normal --token-range-slices=512 --partition-key-buffer-reuse-size=100 --statement-log-file-compression=zstd --duration=1m --warmup=0 --concurrency=5 --mode=write --cql-features=basic --max-mutation-retries=100 --max-mutation-retries-backoff=100ms --replication-strategy=\"{'class': 'NetworkTopologyStrategy', 'replication_factor': '1'}\" --table-options=\"cdc = {'enabled': true, 'ttl': 0}\" --use-server-timestamps=true"
+    gemini_cmd = gemini_thread._generate_gemini_command()
+    assert expected_gemini_arguments in gemini_cmd
     gemini_thread.run()
```

CodeLieutenant (approved):

vponomaryov (approved):
LGTM

PR #10627, opened 2025-04-10
improvement(images): drop supervisor from scylla SCT image

dimakr:
The change drops rebuilding scylladb image as `scylla-sct` image. Instead, scylladb docker image is used as is for the docker backend.
For that purpose the new DockerCmdRunner abstraction was introduced, to allow operating docker based nodes directly using docker API (instead of sshing
into container).

Additionally, artifacts_test was updated to skip subtests/operations for the services, which are no longer supported by scylladb docker image (or are about
to be dropped) - node-exporter and scylla-hosekeeping .

### Testing

- [x] :green_circle: [artifacts-docker-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts-docker-test/12/)
The test was executed against the custom built image (built in the context of work in https://github.com/scylladb/scylladb/pull/22883), that has supervisor dropped
- [x] :green_circle: `PR-provision-test-docker` test, which was executed locally to use the custom built image:
```
❯ SCT_CONFIG_FILES='["test-cases/PR-provision-test-docker.yaml"]' SCT_ENABLE_ARGUS=false SCT_SCYLLA_VERSION=supervisor-less-test1 SCT_N_MONITORS_NODES=0 SCT_DOCKER_IMAGE=syuu1228/scylla ./sct.py run-test longevity_test.LongevityTest.test_custom_time --backend docker
logged in as arn:aws:sts::797456418907:assumed-role/DeveloperAccessRole/dmytro.kruglov@scylladb.com
New directory created: /home/dmitriy/sct-results/20250612-212338-491746
Symlink `/home/dmitriy/sct-results/latest' updated to `/home/dmitriy/sct-results/20250612-212338-491746'
< t:2025-06-12 21:23:38,662 f:base.py         l:376  c:sdcm.sct_events.base p:INFO  > The run can be interrupted by following critical events:
< t:2025-06-12 21:23:38,662 f:base.py         l:376  c:sdcm.sct_events.base p:INFO  >   * ClusterHealthValidatorEvent.NodeStatus
...
< t:2025-06-12 21:27:43,350 f:tester.py       l:3132 c:LongevityTest        p:INFO  > ================================= TEST RESULTS =================================
< t:2025-06-12 21:27:43,350 f:tester.py       l:3132 c:LongevityTest        p:INFO  > ================================================================================
< t:2025-06-12 21:27:43,350 f:tester.py       l:3132 c:LongevityTest        p:INFO  > SUCCESS :)
...
.
----------------------------------------------------------------------
Ran 1 test in 245.231s

OK
```

### Reminders

### DIFF
--- a/sdcm/cluster_docker.py
+++ b/sdcm/cluster_docker.py
@@ -19,6 +19,7 @@ from functools import cached_property
 
 from sdcm import cluster
 from sdcm.remote import LOCALRUNNER
+from sdcm.remote.docker_cmd_runner import DockerCmdRunner
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.filters import DbEventsFilter
 from sdcm.utils.docker_utils import get_docker_bridge_gateway, Container, ContainerManager, DockerException
@@ -39,7 +40,7 @@ class ScyllaDockerRequirementError(cluster.ScyllaRequirementError, DockerExcepti
 class NodeContainerMixin:
     @cached_property
     def node_container_image_tag(self):
-        return self.parent_cluster.node_container_image_tag
+        return self.parent_cluster.source_image
 
     def node_container_image_dockerfile_args(self):
         return dict(path=self.parent_cluster.node_container_context_path)
@@ -86,10 +87,19 @@ class DockerNode(cluster.BaseNode, NodeContainerMixin):
             assert int(container.labels["NodeIndex"]) == node_index, "Container labeled with wrong index."
             self._containers["node"] = container
 
+    def _init_remoter(self, ssh_login_info):
+        self.remoter = DockerCmdRunner(self)
+
+    def _init_port_mapping(self):
+        pass
+
     def _set_keep_duration(self, duration_in_hours: int) -> None:
         pass
 
     def wait_for_cloud_init(self):
+        pass
+
+    def wait_ssh_up(self, verbose=True, timeout=500):
         pass
 
     @property
@@ -105,7 +115,8 @@ class DockerNode(cluster.BaseNode, NodeContainerMixin):
         #                          device_index=0
         #                          )]
 
-    def is_docker(self):
+    @staticmethod
+    def is_docker():
         return True
 
     @cached_property
@@ -139,21 +150,27 @@ class DockerNode(cluster.BaseNode, NodeContainerMixin):
         ContainerManager.get_container(self, "node").restart()
 
     def get_service_status(self, service_name: str, timeout: int = 500, ignore_status=False):
-        return self.remoter.sudo('sh -c "{0} || {0}.service"'.format(f"supervisorctl status {service_name}"),
-                                 timeout=timeout, ignore_status=ignore_status)
+        container_status = ContainerManager.get_container(self, "node").attrs['State']['Status']
+        # return similar interface to what supervisorctl would return
+        return type('ServiceStatus', (), {
+            'stdout': f"{service_name} {container_status}",
+            'stderr': "",
+            'ok': container_status in ['running', 'active', 'RUNNING']
+        })
 
     def start_scylla_server(self, verify_up=True, verify_down=False, timeout=300, verify_up_timeout=None):
         verify_up_timeout = verify_up_timeout or self.verify_up_timeout
         if verify_down:
             self.wait_db_down(timeout=timeout)
-        self.remoter.sudo('sh -c "{0} || {0}-server"'.format("supervisorctl start scylla"),
-                          timeout=timeout)
+
+        container = ContainerManager.get_container(self, "node")
+        if container.status != 'running':
+            self.log.info("Starting Docker container %s", container.name)
+            container.start()
+            ContainerManager.wait_for_status(self, "node", status="running")
+
         if verify_up:
             self.wait_db_up(timeout=verify_up_timeout)
-
-        # Need to start the scylla-housekeeping service manually because of autostart of this service is disabled
-        # for the docker backend. See, for example, docker/scylla-sct/ubuntu/Dockerfile
-        self.start_scylla_housekeeping_service(timeout=timeout)
 
     @cluster.log_run_info
     def start_scylla(self, verify_up=True, verify_down=False, timeout=300):
@@ -162,24 +179,15 @@ class DockerNode(cluster.BaseNode, NodeContainerMixin):
     def stop_scylla_server(self, verify_up=False, verify_down=True, timeout=300, ignore_status=False):
         if verify_up:
             self.wait_db_up(timeout=timeout)
+
+        container = ContainerManager.get_container(self, "node")
+        self.log.info(f"Stopping Docker container {container.name}")
         # ignoring WARN messages upon stopping - https://github.com/scylladb/scylla-cluster-tests/issues/10633
         with DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line="WARN "):
-            self.remoter.sudo('sh -c "{0} || {0}-server"'.format("supervisorctl stop scylla"),
-                              timeout=timeout)
+            container.stop(timeout=timeout)
+
         if verify_down:
             self.wait_db_down(timeout=timeout)
-
-        # Need to start the scylla-housekeeping service manually because of autostart of this service is disabled
-        # for the docker backend. See, for example, docker/scylla-sct/ubuntu/Dockerfile
-        self.stop_scylla_housekeeping_service(timeout=timeout)
-
-    def stop_scylla_housekeeping_service(self, timeout=300):
-        self.remoter.sudo('sh -c "{0} || {0}-server"'.format("supervisorctl stop scylla-housekeeping"),
-                          timeout=timeout)
-
-    def start_scylla_housekeeping_service(self, timeout=300):
-        self.remoter.sudo('sh -c "{0} || {0}-server"'.format("supervisorctl start scylla-housekeeping"),
-                          timeout=timeout)
 
     @cluster.log_run_info
     def stop_scylla(self, verify_up=False, verify_down=True, timeout=300):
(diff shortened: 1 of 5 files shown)


dimakr:
@fruch This is a draft. Can you take a look and suggest if the direction is OK and we can continue with it?

For now this change manages the processes directly, but nothing is dropped from the scylla-sct image (supervisor is just ignored).
Couple points to think about/discuss are:
- as there is no scylladb image without supervisor, I made a local copy of [docker-entrypoint.py](https://github.com/scylladb/scylladb/blob/master/dist/docker/docker-entrypoint.py). In this local copy the supervisor daemon is not started, but instead we start standalone processes that we need (e.g. sshd, node-exporter, etc.). Then in SCT when the `scylla-sct` image is built, e.g. during artifacts-dockerf-test, we copy the local version of `docker-entrypoint.py` into image.
Once the supervisor is not used/started in the scylladb docker image itself, this workaround can be dropped.
- I added abstractions/wrappers for docker container operations and service management operations. Although the docker-py lib itself provides clear APIs, I wrapped few APIs we are interested in with additional logic.
Should we keep it, or maybe it  is complicating the thing unnecessarily?

sdcm/utils/docker_service_manager.py
```diff
@@ -0,0 +1,152 @@
+import logging
+import time
+
+from sdcm.utils.docker_remote import RemoteDocker
+
+LOGGER = logging.getLogger(__name__)
+
+
+class DockerServiceManager:
```
fruch, line 9: I don't understand why we need to keep mocking the services behavior
we shouldn't execute any command inside the docker instance for scylla start/stop

it should be only:

```
docker start
docker stop
docker ps/inspect
```

or the equivalent of them in API call to docker 

no user is gonna do `pkill -TERM scylla` within the docker, docker itself is our service manager.

sdcm/cluster_docker.py
```diff
@@ -191,15 +223,19 @@ def restart_scylla_server(self, verify_up_before=False, verify_up_after=True, ti
...
+        if self.docker_service_manager:
+            self.docker_service_manager.restart_service('scylla', timeout=timeout)
+        else:
+            # ignoring WARN messages upon stopping - https://github.com/scylladb/scylla-cluster-tests/issues/10633
+            with DbEventsFilter(db_event=DatabaseLogEvent.BACKTRACE, line="WARN "):
+                self.remoter.sudo('sh -c "{0} || {0}-server"'.format("supervisorctl restart scylla"), timeout=timeout)
         if verify_up_after:
             self.wait_db_up(timeout=verify_up_timeout)
-        self.start_scylla_housekeeping_service(timeout=timeout)
+        # self.start_scylla_housekeeping_service(timeout=timeout)
```
fruch, line 238: we can remove anything related to housekeep, I'm o.k. for not test housekeep at all with docker backend

sdcm/cluster_docker.py
```diff
@@ -81,11 +83,19 @@ def __init__(self,
...
+        self.docker_service_manager = None
+        self._docker_cmd_runner = None
 
         if container is not None:
             assert int(container.labels["NodeIndex"]) == node_index, "Container labeled with wrong index."
             self._containers["node"] = container
 
+    def setup_docker_management(self, container: Container):
+        """Set up Docker command runner and service manager for direct container management"""
+        self._docker_cmd_runner = RemoteDocker(
```
fruch, line 95:
```suggestion
        self.remoter = RemoteDocker(
```

and put that in `_init_remoter`  so there won't be any ssh, and then it would surface those places needs attention

fruch: also I don't see where you remove the part it's building the docker, we should ditch the whole thing even throwing out the whole `NodeContainerMixin`

sdcm/utils/docker_service_manager.py
dimakr: It's still adraft.
The underlying task in scylladb up until today was implementing drop of supervisor in a way that docker-entrypoint.sh was starting 2 processes - scylla and housekeeping, so I thought that some APIs would be useful for fine-grained management of those services.

But seems like they are going to drop housekeeping altogether (or add to the main binary). So eventually the scylla itself will probably run as pid 1 and we indeed do not need to do any kind of `docker exec` into container for service management

fruch: even it they still run housekeeping on it's own process, I want to ignore it.
it's just a waste of our time trying to build all kind of layer just for it.

We don't need to test it in docker, period.

if it would be part of scylla binary, we'll need to refactor all of the related code, let's have less of that code.

sdcm/cluster_docker.py
dimakr: Removed building the scylla-sct docker image

sdcm/utils/docker_service_manager.py
dimakr: n/a, in the latest implementation this additional abstraction is not used

sdcm/cluster_docker.py
```diff
@@ -387,15 +410,51 @@ def __init__(self,
...
         node.remoter.sudo("apt install -y openjdk-8-jre", verbose=True, ignore_status=True)
         node.remoter.sudo("ln -sf /usr/lib/jvm/java-1.8.0-openjdk-amd64/jre/bin/java* /etc/alternatives/java",
                           verbose=True, ignore_status=True)
 
+        self.install_sudo(node, verbose=verbose)
+        self._install_docker_cli(node, verbose=verbose)
         if self.params.get('client_encrypt'):
             node.config_client_encrypt()
 
+    def _install_docker_cli(self, node, verbose=False):
```
fruch, line 423: this is for loader only, we shouldn't use that for scylla instances

sdcm/utils/docker_service_manager.py
fruch, line 9: I think we can remove this completely

soyacz:
I've tried `hydra investigate show-monitor 23df17b2-7171-4b7c-af7f-919e03bf63ad` (provision test) and it failed:
```
Restoring monitoring stack from archive 23df17b2-7171-4b7c-af7f-919e03bf63ad/20250602_101740/monitor-set-23df17b2.tar.zst
Download file https://cloudius-jenkins-test.s3.amazonaws.com/23df17b2-7171-4b7c-af7f-919e03bf63ad/20250602_101740/monitor-set-23df17b2.tar.zst to directory /tmp/tmpnt8h6dv9
Downloading 23df17b2-7171-4b7c-af7f-919e03bf63ad/20250602_101740/monitor-set-23df17b2.tar.zst from cloudius-jenkins-test
Downloaded finished
Creating monitoring stack directories for monitor-node-23df17b2-7171-4b7c-af7f-919e03bf63ad-0 cluster
Docker containers for monitoring stack are started
Nemesis dashboard file for the cluster monitor-node-23df17b2-7171-4b7c-af7f-919e03bf63ad-0 is '/tmp/tmpnt8h6dv9/scylla-monitoring-src/sct_monitoring_addons/sct-monitor-J1N4A-scylla-dash-per-server-nemesis.master.json'
Dashboard /tmp/tmpnt8h6dv9/scylla-monitoring-src/sct_monitoring_addons/sct-monitor-J1N4A-scylla-dash-per-server-nemesis.master.json loaded successfully
Error during annotations data upload Expecting value: line 1 column 1 (char 0)
Uploading annotations data [try #1]
Error during annotations data upload Expecting value: line 1 column 1 (char 0)
Uploading annotations data [try #2]
Error during annotations data upload Expecting value: line 1 column 1 (char 0)
'restore_annotations_data': Number of retries exceeded!
Error during uploading sct monitoring data Expecting value: line 1 column 1 (char 0)
Errors were found when restoring Scylla monitoring stack
Killing Grafana service
Killing Prometheus service
Killing Alert service
```

did you try with some other test?
I wonder how it looks when there's no node_exporter anymore.

dimakr:
> I've tried `hydra investigate show-monitor 23df17b2-7171-4b7c-af7f-919e03bf63ad` (provision test) and it failed:
> 
> did you try with some other test? I wonder how it looks when there's no node_exporter anymore.

yes, @fruch raised this concern that without node-exporter in the image, we won't have metrics, etc.
In the original core task of dropping supervisor, in https://github.com/scylladb/scylladb/pull/22883#issuecomment-2662749149, it was mentioned that a dedicated container should be used for node-exporter in the context of operator. Probably we can follow their approach, when they implement this (or reuse the image if possible)


> how about logging (e.g. syslog-ng)? also we no longer ship `iproute` or `rsync` pkg which might be needed by some nemesis. Did you try running with podman?
good question
This is what a basic test (pr-provision-docker) has for db node logs when executed locally:
```
❯ ll latest/PR-provision-docker-dmitriy-db-cluster-2b51a250
total 12K
drwxrwxr-x 3 dmitriy 4,0K cze 13 14:27 .
drwxrwxr-x 7 dmitriy 4,0K cze 13 14:33 ..
drwxrwxr-x 2 dmitriy 4,0K cze 13 14:27 PR-provision-docker-dmitriy-db-node-2b51a250-0
❯ ll latest/PR-provision-docker-dmitriy-db-cluster-2b51a250/PR-provision-docker-dmitriy-db-node-2b51a250-0
```

And this is for comparison what we have for pr-provision test:
```
❯ ll ./20250403-104249-259616/PR-provision-test-dmitriy-db-cluster-9820234b/PR-provision-test-dmitriy-db-node-9820234b-1
total 9,4M
drwxrwxr-x 2 dmitriy 4,0K kwi  3 10:56 .
drwxrwxr-x 7 dmitriy 4,0K kwi  3 11:04 ..
-rw-rw-r-- 1 dmitriy  900 kwi  3 10:56 autossh.log
-rw-rw-r-- 1 dmitriy 9,4M kwi  3 10:47 kallsyms_20250403_084342
lrwxrwxrwx 1 dmitriy   69 kwi  3 10:50 system.log -> ../../hosts/PR-provision-test-dmitriy-db-node-9820234b-1/messages.log
```
So without the dedicated scylla-sct image and its tooling we lack some logs.
We need to think what and how can be used instead.

fruch:
For logging in docker backend we never used syslong, so nothing new here

As for metrics, that an issue for scylla core, of how gonna be building the docker image, we'll need a confirmation from that exporter is gonna be removed, and we'll see how to approach it.

Even it mean that in the meanwhile docker backend won't gonna have monitoring data

fruch:
> @fruch After recent fixes and few adjustments, have this change passing against scylla-nightly:latest image on both, pr-rpovision-docker and artifacts-docker tests, as we discussed: 🟢 [pr-provision-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/66/) 🟢 [artifactsdocker](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts-docker-test/14/) ('adjustment' - had to return to installing sudo on db node, as during scylla config setup flow in
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/5dd5e9a2979e23c682c667010205acfc1a560081/sdcm/cluster.py#L1776
> 
> , it operates on remote config under sudo)

raise issue for that, we should be able to change `scylla_yaml` to be able to work without sudo for docker backend

dimakr:
> raise issue for that, we should be able to change `scylla_yaml` to be able to work without sudo for docker backend

Created https://github.com/scylladb/scylla-cluster-tests/issues/11230 task for this.

sdcm/cluster_docker.py
fruch: @dimakr what about this ? I can accept sudo is needed for something, but we need to remove this part

we shouldn't be installing docker cli into any docker instances

dimakr:
> @dimakr please raise an issue for the docker backend loader

Created a follow-up task https://github.com/scylladb/scylla-cluster-tests/issues/11251.
@fruch

Issue #10633, opened 2025-04-14
scylla-master - Docker failing


lsfreitas:
Hey guys,

Docker is failing for some days already and I'm clueless on what's really happening. I'm opening this issue so we can see what's the issue:

Docker build failing: https://jenkins.scylladb.com/job/scylla-master/job/docker/964/

Artifact-docker-test failing: https://jenkins.scylladb.com/job/scylla-master/job/artifacts/job/artifacts-docker-test/1328/consoleFull

Error:

```
23:13:00  ----- LAST ERROR EVENT -------------------------------------------------------
23:13:00  2025-04-11 22:12:19.559: (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=c05270ec-76d7-46ba-b73d-6d0cf5643723: type=BACKTRACE regex=^(?!.*audit:).*backtrace line_number=1330 node=artifacts-docker-jenkins-db-node-564d5f8f-0
23:13:00  WARN  2025-04-11 22:12:19,485 [shard 0:sl:d] seastar - Exceptional future ignored: seastar::gate_closed_exception (gate closed), backtrace: 0x1e0a2ee 0x162e631 0x152c425 0x2f2317d 0x1b4e71b 0x30328a8 0x3032f64 0x30329c2 0x3030b85 0x2ba8872 /opt/scylladb/libreloc/libc.so.6+0x3247 /opt/scylladb/libreloc/libc.so.6+0x330a 0x32fc674
23:13:00  ----- LAST WARNING EVENT -----------------------------------------------------
23:13:00  2025-04-11 22:11:53.585: (CommitLogCheckErrorEvent Severity.WARNING) period_type=one-time event_id=6d2fe46d-6c65-4022-be68-af1cc354fff5: message=CommitLogCheckThread will not start due to no monitors in the cluster
23:13:00  ----- LAST NORMAL EVENT ------------------------------------------------------
23:13:00  2025-04-11 22:12:46.070: (InfoEvent Severity.NORMAL) period_type=not-set event_id=4bf6af8e-6876-43fe-bda2-ba4cacba871a: message=TEST_END
23:13:00  ================================================================================
23:13:00 
 FAILED :(
```

lsfreitas:
@pehala @soyacz @vponomaryov asking for your attention here.

soyacz:
[...]

PR #10635, opened 2025-04-14
fix(docker): fix set kernel perf event paranoid set

soyacz:
When using docker and setting `print_kernel_callstack` param we need to set `kernel.perf_event_paranoid=0` - which didn't work due missing params and proper privileges for docker.

refs: https://github.com/scylladb/scylla-cluster-tests/issues/10633

### Testing

- [x] - tested locally with docker backend

### Reminders

### DIFF
--- a/docker/scylla-sct/centos/Dockerfile
+++ b/docker/scylla-sct/centos/Dockerfile
@@ -24,6 +24,7 @@ RUN microdnf update && \
         sudo \
     	openssh-server \
  		docker-ce-cli \
+        procps-ng \
         rsync && \
     microdnf clean all && \
     echo "autostart=false" >> /etc/supervisord.conf.d/scylla-server.conf && \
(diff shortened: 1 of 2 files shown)


yaronkaikov (approved):
Thanks @soyacz

soyacz:
> @lsfreitas FYI

this is not solving the original issue, just helps with mentioned error when setting this system setting.

vponomaryov:
@soyacz I don't think we should update that config option on docker backend.
With this change everyone running a docker SCT test will get this option changed on his/her machine.

I think we should update the related config option instead:
```
print_kernel_callstack: false
```

soyacz:
> @soyacz I don't think we should update the `kernel.perf_event_paranoid` host config option on docker backend. With this change everyone running a docker SCT test will get this option changed on his/her machine.

hmm, yes, that seems to be true. Better not change host settings.

> 
> I think we should update the related SCT config option instead:
> 
> ```
> print_kernel_callstack: false
> ```

so, I'd propose to disable it at code by adding `and not self.is_docker()` level in condition, as anyway it is not working and won't generate unrelevant errors. We could consider adding a warning that we skip it.:
```
if self.node_type == "db" and not self.is_kubernetes() \
                and self.parent_cluster.params.get("print_kernel_callstack"):
```

soyacz:
wrong approach. closing

PR #10638, opened 2025-04-14
fix(docker): disable `print_kernel_callstack` for docker backend

vponomaryov:
Ref: https://github.com/scylladb/scylla-cluster-tests/issues/10633

### Testing

- [ ]

### Reminders

### DIFF
--- a/defaults/docker_config.yaml
+++ b/defaults/docker_config.yaml
@@ -3,3 +3,4 @@ logs_transport: docker
 use_preinstalled_scylla: true
 docker_image: ''
 scylla_linux_distro: 'centos'
+print_kernel_callstack: false
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -397,14 +397,18 @@ class BaseNode(AutoSshContainerMixin):  # pylint: disable=too-many-instance-attr
         self.set_keep_alive()
         if self.node_type == "db" and not self.is_kubernetes() \
                 and self.parent_cluster.params.get("print_kernel_callstack"):
-            try:
-                self.remoter.sudo(shell_script_cmd("""\
-                echo 'kernel.perf_event_paranoid = 0' >> /etc/sysctl.conf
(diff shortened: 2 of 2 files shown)


soyacz:
I'd propose to disable it at code by adding and not self.is_docker() level in condition, as anyway it is not working and won't generate unrelevant errors. We could consider adding a warning that we skip it.:
```
if self.node_type == "db" and not self.is_kubernetes() \
                and self.parent_cluster.params.get("print_kernel_callstack"):
```

soyacz (approved):
LGTM

vponomaryov:
@roydahan if you are ok with it, please merge.

fruch:
@vponomaryov can you add a bit more explanation why we can't use it ?, the information on the issue isn't very clear about why this is the root cause.

and it was skipped with a different method by now

soyacz:
> @vponomaryov can you add a bit more explanation why we can't use it ?, the information on the issue isn't very clear about why this is the root cause.
> 
> and it was skipped with a different method by now

we discussed it here: https://github.com/scylladb/scylla-cluster-tests/pull/10635

vponomaryov:
> @vponomaryov can you add a bit more explanation why we can't use it ?, the information on the issue isn't very clear about why this is the root cause.

This is `not` a root cause, this is fix for reducing human confusion and making logs be clearer.
That's why the commit message only "ref"ers the https://github.com/scylladb/scylla-cluster-tests/issues/10633 and not "fixes".

fruch:
> > @vponomaryov can you add a bit more explanation why we can't use it ?, the information on the issue isn't very clear about why this is the root cause.
> 
> This is `not` the root cause fix, this is fix for reducing human confusion and making logs be clearer. That's why the commit message only "ref"ers the #10633 and not "fixes".

couldn't understand it from the whole discussion on that issue..

putting it on the commit/PR description, would have helped...

PR #10648, opened 2025-04-16
fix(docker): ignore warnings during stop

soyacz:
In docker backend we recently started to get WARN messages upon stopping node. Because this warn message is catched by db error regex, it causes docker artifact test to fail and mess Scylla CI. This warn is not considered as error and should be ignored.

This commit makes all warnings during stop/restart to be ignored (for docker backend only).

closes: https://github.com/scylladb/scylla-cluster-tests/issues/10633

### Testing

- [X] - tested locally

### Reminders


roydahan (approved):

vponomaryov (approved):
LGTM

PR #10666, opened 2025-04-21
fix(performance): not validate P90 metrix

juliayakovlev:
As discussed with @fruch , we do not want to fail the test on P90 metrix. If P90 read/write is higher then 1 or 3 (as it is defined now), the test still should be passed

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

Issue #7327, opened 2024-04-05
Prepare write cmd timeout is set incorrectly (mostly using `latte` stress tool)


vponomaryov:
## Packages

Scylla version: `2023.1.7-20240402.1f84904a7100` with build-id `d8d3f80469204be472563634963c19f86b295af0`

Kernel Version: `5.15.0-1054-gcp`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

If we set the `stress_duration` SCT config option then it gets applied to the `prepare` cmds too if commands have `duration=` substring.
Both parts get redefined - `duration` and `timeout` of cmds.
At first, we get wrong duration and timeout.
At second, it can cause false timeouts of stress commands.
In case of `latte` stress runner, the duration in prepare commands is not `time`, but `num of rows`.
So, in this failed run the `duration` stayed as-is because it didn't have the timing label such as `m`, `h` or `s`.
But the timeout was set to small value based on the duration of the main stress commands (30m).

The problem was not observed before, because the 'duration=' part gets set only in latte prepare commands and timeout smaller than the real execution of prepare commands is mostly bigger.

The false stress command timeout looks like following:
```
16:17:37  < t:2024-04-05 13:17:33,072 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > stress_cmd=latte run --tag latte --duration 82286400 --request-timeout 60 --retry-interval '2s,10s' --threads 30 --connections 3 --concurrency 180 --rate 15000 --sampling 5s -P offset=0 --function load scylla-qa-internal/custom_d1/workload1/latte/custom_d1_workload1.rn  -q -- 10.142.0.30 10.142.0.32 10.142.0.47 10.142.0.49 10.142.0.50 10.142.0.52 
16:17:37  < t:2024-04-05 13:17:33,072 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > errors:
16:17:37  < t:2024-04-05 13:17:33,072 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > 
16:17:37  < t:2024-04-05 13:17:33,072 f:file_logger.py  l:101  c:sdcm.sct_events.file_logger p:INFO  > \
    Stress command execution failed with: Command did not complete within 2580 seconds!
```

`2580 seconds` (43 minutes) consists of the following parts: 
- 30 minutes of the main stress cmd
[...]

PR #10674, opened 2025-04-22
fix(preload): use proper timeout for preload stress commands

vponomaryov:
For the moment, if we define custom duration for the main stress load
then we get a timeout based on it for every stress command, even preload one.

When it is bigger than duration of the preload command it is ok,
but when it is smaller than we get false negative result.

So, fix it by specifying preload duration in functions designed
to call exactly this type of stress commands.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/7327

### Testing

- [ ]

### Reminders


vponomaryov:
One failed integration test is unrelated to this change.
Ref: https://github.com/scylladb/scylla-cluster-tests/issues/10676

fruch (approved):
LGTM

PR #10696, opened 2025-04-24
test(cql-stress): use cql-stress in some tests

soyacz:
We want to increase coverage of rust driver in SCT tests. Replacing c-s with cql-stress in some of them.

refs: https://github.com/scylladb/qa-tasks/issues/1876

testing details: https://scylladb.atlassian.net/wiki/spaces/RND/pages/64782385/c-s+-+cql-stress+migration

### Testing

- [ ] - [one more run](https://jenkins.scylladb.com/job/scylla-staging/job/lukasz/job/longevity-100gb-4h-test/12/), for final PR

### Reminders


soyacz:
> Is it really the desired way to use this tool?
> 
> I thought that command compatibility was part of the tool design for reusing same commands. And on SCT side we would control which tool to use with some additional one config option like following:
> 
> * "cs_type": "java"
> * "cs_type": "rust"
> 
> Or something else looking as a single option which will allow to reuse existing configs.

In some cases this could be done, but in some we need to play with threads as cql-stress behaves differently in terms of performance.
Also we can reduce loader sizes/count when using cql-stress and get the same throughput's.

fruch:
> > > Is it really the desired way to use this tool?
> > > I thought that command compatibility was part of the tool design for reusing same commands. And on SCT side we would control which tool to use with some additional one config option like following:
> > > 
> > > * "cs_type": "java"
> > > * "cs_type": "rust"
> > > 
> > > Or something else looking as a single option which will allow to reuse existing configs.
> > 
> > In some cases this could be done, but in some we need to play with threads as cql-stress behaves differently in terms of performance. Also we can reduce loader sizes/count when using cql-stress and get the same throughput's.
> 
> In this case we can replace `cql-stress` with `latte` and it will be correct statement as well.
> So, what is the benefit of `cql-stress` if we are going to use everything separate for it?
> 
> If we really need different thread count, we can implement some additional "thread ratio" config option.
> But I am not sure it is really needed running longevity tests. It must matter only in perf tests.
> 
> **So, my main take is following:**
> The reusage of the same config files is the greatest value we have from "cql-stress" benchmarking tool.
> Because it's rust+performance part is covered by latte, which, moreover, doesn't require new tool releases when we need to update something that is covered in a rune script which we store in SCT...

Yes latte could be more flexible,

But the fact we want to tune some parts for when using cql-stress doesn't make it less useful for us.
it's still a drop in replacement for c-s, even if it's not yet 100% compatible, but it's close.

For cases we'll have complex requirements, latte would be preferable. for simple cases cql-stress is still what we would prefer.

Why are suggesting replacing s-b, technically speaking latte can cover it as well ?

BTW, both cql-stress and s-b do validation, which latte doesn't yet have

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,34 @@
+test_duration: 750
+prepare_write_cmd: "cql-stress-cassandra-stress write cl=ALL n=50050075  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=1000 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..50050075 -log interval=15"
+
+#stress_cmd: ["cql-stress-cassandra-stress write cl=QUORUM duration=11h  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate 'threads=250 throttle=45000/s'  -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=400200300..600200300 -log interval=15"]
```
fruch, line 4: why this commented out?

soyacz: still in progress - I was testing if this does not cause us troubles, but I found root cause https://github.com/scylladb/cql-stress/issues/138

soyacz:
I've released new cql-stress with support for client encryption and validation bug fix. Reran tests (updated description).

soyacz:
ready for review - encryption and bug with querying added columns fixed (and retested in referenced runs)

soyacz:
ping - do we proceed with it or switch to `latte`?

fruch:
> ping - do we proceed with it or switch to `latte`?

for longevity let continue with this effort, for pref test I didn't decided yet.

roydahan:
I think that also for performance it would be good to proceed, it will help us as the users of our own driver to know how to use and optimize its usage correctly.

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,32 @@
+test_duration: 750
+prepare_write_cmd: "cql-stress-cassandra-stress write cl=ALL n=50050075  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=1000 -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=1..50050075 -log interval=15"
+
+stress_cmd: ["cql-stress-cassandra-stress write cl=QUORUM duration=11h  -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate 'threads=250 throttle=45000/s'  -col 'size=FIXED(1024) n=FIXED(1)' -pop seq=400200300..600200300 -log interval=15"]
```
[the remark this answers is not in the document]
fruch: O.k.

fruch (approved):
LGTM

Issue #10697, opened 2025-04-24
Disks are not cleaned up correctly after running SCT longevity tests for Operator


grzywin:
@zbierak noticed that:

> in sct-project-1 in GCP there are hundreds of disks not attached to any instance (all with "pvc" prefix). Based on labels, they seem to be left-overs from some k8s nodes.
 
These disks comes from GCP longevity tests for Operator.

Proper fix/method needs to be implemented in SCT to make sure that disks are properly removed in the cleanup phase of the test run.

![Image](https://github.com/user-attachments/assets/2d1de852-0478-498e-83c8-0b3ecc4de651)

vponomaryov:
We have "clean_orphaned_gke_disks" function which gets called as part of the "clean_cloud_resources" one.
Both run if the "clean-resources" stage of a pipeline runs and doesn't fail.

So, if you see such a huge number of left disks then it means something got broken and need to investigate what exactly.

fruch:
> Can we set those volumes to be auto-deleted when the instances are deleted? So we don't have to clean them up later.

They are not tied to instances, but to pods, and handled by k8s.

so control is not straight forward like in VM use case.

We do have code that clean them up, when a test end, but that can fail, hence we still need a periodic cleanup

We have that period cleanup for most of the resources we are using.

PR #10699, opened 2025-04-24
fix(gcp_cleanup): Fix gcp filter to collect correct disks for clean up

grzywin:
Fix GCP filter in compute disks list command to collect correct disks for clean up

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/10697

### Testing

- [x] https://jenkins.scylladb.com/job/scylla-operator/job/operator-master/job/gke/job/longevity-scylla-operator-3h-gke-test/127/execution/node/270/log/

### DIFF
--- a/sdcm/utils/gce_utils.py
+++ b/sdcm/utils/gce_utils.py
@@ -292,7 +292,8 @@ class GkeCleaner(GcloudContainerMixin):
         disks_per_zone = {}
         try:
             disks = json.loads(self.gcloud.run(
-                'compute disks list --format="json(name,zone)" --filter="name~^gke-.*-pvc-.* AND -users:*"'))
+                'compute disks list --format="json(name,zone)" --filter="(name~^gke-.*-pvc-.* OR name~^pvc-.*) '
+                'AND -users:*"'))
         except Exception as exc:  # pylint: disable=broad-except  # noqa: BLE001
             LOGGER.error("`gcloud compute disks list' failed to run: %s", exc)
         else:


sdcm/utils/gce_utils.py
```diff
@@ -292,7 +292,8 @@ def list_orphaned_gke_disks(self) -> dict:
         disks_per_zone = {}
         try:
             disks = json.loads(self.gcloud.run(
-                'compute disks list --format="json(name,zone)" --filter="name~^gke-.*-pvc-.* AND -users:*"'))
+                'compute disks list --format="json(name,zone)" --filter="(name~^gke-.*-pvc-.* OR name~^pvc-.*) '
```
vponomaryov, line 295: As I understand, the GCP cloud platform removed "gke-" prefix and that's why it stopped matching, right?
If not why names got changed?

Then, if that filter works the standard "logical" way, then the `-users:*` part is paired only with the `name~^pvc-.`, but not with the `name~^gke-.*-pvc-.*`.
Is it intentional? To me it doesn't look as such.

So, even if it worked it must be adjusted.
At first, decide whether we expect 2 different name prefixes or not.
If not, remove unexpected variant.
If yes, then need to fix the filter where we apply "third" logical element equally to each of the first two.

vponomaryov (changes requested):
Good catch, but need to adjust the filter

sdcm/utils/gce_utils.py
grzywin: > As I understand, the GCP cloud platform removed "gke-" prefix and that's why it stopped matching, right?
> If not why names got changed?

Yes, it seems that in newer GKE clusters, the "gke-" prefix is no longer consistently used in created PVC disk names. I observed that disks now start directly with "pvc-", which is why I added the second condition.

> Then, if that filter works the standard "logical" way, then the -users:* part is paired only with the name~^pvc-., but not with the name~^gke-.*-pvc-.*.
> Is it intentional? To me it doesn't look as such.

Hm. I think that my filter will work correctly. There is a bracket before AND statement. So first we are checking if gke-pvc prefix is correct OR only pvc, and only after that we are using AND to check condition with users. So AND should be valid/applied for both cases.
`(name~^gke-.*-pvc-.* OR name~^pvc-.*) AND -users:*`

> So, even if it worked it must be adjusted.
> At first, decide whether we expect 2 different name prefixes or not.
> If not, remove unexpected variant.
> If yes, then need to fix the filter where we apply "third" logical element equally to each of the first two.

I am not 100% sure, but I think the reason the prefix changed is due to the change in the provisioner in newer versions of GKE. I would keep both versions just in case, because why not? That way, we can be sure that no matter which GKE version is used, the disk will be removed.

vponomaryov: Misslooked the brackets positions. Nevermind.

vponomaryov (approved):
LGTM

fruch (approved):

PR #10708, opened 2025-04-27
test(restore_procedure): Add jupyter notebook for procedure

aleksbykov:
Add jupyter notebook for raft recovery procedure for raft + tablets

### Testing

- [ ]

### Reminders


jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,643 @@
...
+   "source": [
+    "### Clean the cluster"
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "70774d1b",
+   "metadata": {},
+   "outputs": [
```
soyacz, line 591: at the end, remember to remove outputs from commit

aleksbykov:
To view : https://github.com/scylladb/scylla-cluster-tests/blob/fbb59002a12181de20f615d88d5ef07d31a097f4/jupyter/restore_raft_quorum_with_tablets.ipynb

jupyter/restore_raft_quorum_with_tablets.ipynb
aleksbykov: Done

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "1ac783b9",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "os.environ[\"BUILD_USER_EMAIL\"] = \"alex.bykov@scylladb.com\"\n",
```
soyacz, line 50: It’s in your best interest to switch to `sct.tester@scylladb.com` :smile:

soyacz (commented):
Please add preamble to this jupyter notebook about use case: why it's needed and what is a base for this procedure (link to scylla docs?)

Would be good to have another opinion from someone who knows the procedure.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
+{
```
patjed41, line 1: The PR title looks wrong. It's not a restore procedure but a recovery procedure.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "print(verification_node.run_nodetool(\"status\"))"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "de494749",
+   "metadata": {},
+   "source": [
+    "### Validate that raft quorum lost\n",
+    "Try to remove node dead node\n",
```
patjed41, line 298: typo: "node dead node"

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "571d6f73",
+   "metadata": {},
+   "source": [
+    "# Recovery procedure\n",
+    "Execute each step one by one\n",
+    "Use 'GET CLUSTER STATUS' cell to check cluster status between steps if need"
```
patjed41, line 330: `\n` doesn't cause the new line to appear in the notebook view, I see:
```
Execute each step one by one Use 'GET CLUSTER STATUS' cell to check cluster status between steps if need
```

There are more places like this.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "f632964f",
+   "metadata": {},
+   "source": [
+    "## Step 2.\n",
+    "Stop one of alive node in alive dc.\n",
```
patjed41, line 339: I don't understand the point of this step. Perhaps you misunderstood "2. Kill any live node that shouldn't be a member of the new group 0." from https://github.com/scylladb/scylladb/pull/22286#issue-2783582569. This step is about stopping live nodes that cannot be recovered, for example nodes that were joining the cluster during quorum loss. See the new docs: https://docs.scylladb.com/manual/master/troubleshooting/handling-node-failures.html#manual-recovery-procedure

> Ensure all live nodes are in the normal state using [nodetool status](https://docs.scylladb.com/manual/master/operating-scylla/nodetool-commands/status.html). If there is a node that is joining or leaving, it cannot be recovered. You must permanently stop it. After performing the recovery procedure, use [nodetool status](https://docs.scylladb.com/manual/master/operating-scylla/nodetool-commands/status.html) ony any other node. If the stopped node appears in the output, it means that other nodes still consider it a member of the cluster, and you should remove it with the [node removal procedure](https://docs.scylladb.com/manual/master/operating-scylla/procedures/cluster-management/remove-node.html).

This step is for corner cases, I think you can skip it here.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+   ]
+  },
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "42acf42c",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "base_alive_nodes = nodes_by_region[alive_dc_region][:-1]\n",
```
patjed41, line 373: We should take all nodes from `alive_dc_region` here, see my long comment above.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "cluster.restart_scylla(nodes=base_alive_nodes)\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "b4fc3063",
+   "metadata": {},
+   "source": [
+    "## Step 6. \n",
+    "Find recovery coordinator"
```
patjed41, line 383: We call it "recovery leader" everywhere. It would be good replace all "coordinator" occurrences with "leader" in this PR.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "    "
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "2ed22247",
+   "metadata": {},
+   "source": [
+    "## Step 8.\n",
+    "Set the new scylla.yaml parameter, recovery_leader, to Host ID of the recovery leader on each live node"
```
patjed41, line 434: We must also send the `SIGHUP` signal to all ScyllaDB processes to ensure that the changed parameter is applied (see https://docs.scylladb.com/manual/master/troubleshooting/handling-node-failures.html#manual-recovery-procedure, step 5).

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "Remove/replace all dead nodes\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "617dc05e",
+   "metadata": {},
+   "source": [
+    "### step 10.1\n",
+    "Remove stopped node in alive dc. it is possible because each rack had 2 nodes"
```
patjed41, line 486: If you revert stopping one node from alive DC, you will have to remove this step. See my comments above.

The same goes for step 10.4.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "replace_nodes_by_host_id(cluster, dead_node_host_ids_maps, verification_node)"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "c5b04413",
+   "metadata": {},
+   "source": [
+    "#### Step 10.3\n",
+    "Add new nodes to racks\n"
```
patjed41, line 530: I don't understand why you added this step. In step 10.2 we already replace all dead nodes. Why do we need more nodes?

Doing this as a test (for example to verify that topology operations work) is fine, but then you should specify that it is an additional test and not a part of the procedure.

jupyter/restore_raft_quorum_with_tablets.ipynb
```diff
@@ -0,0 +1,680 @@
...
+    "\n"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "id": "edba70d7",
+   "metadata": {},
+   "source": [
+    "## Step 11.\n",
+    "Unset recovery_leader on all nodes."
```
patjed41, line 573: We need to send the `SIGHUP` signal too (see https://docs.scylladb.com/manual/master/troubleshooting/handling-node-failures.html#manual-recovery-procedure, step 8).

patjed41 (changes requested):
I get that using jupyter notebook is helpful because it allows us to perform the steps of the recovery procedure one by one and see what is going on. It also allows us to easily experiment, for example, we can change the cluster configuration and see what happens then. But I don't understand how we will use this notebook in SCT. Can you explain the motivation for this PR and the plan for this notebook?

Another question. IIUC, we don't have any workload with client requests. Serving client requests during the recovery procedure is supported (according to CL/RF rules), so it should be tested. Maybe we should add a workload? Alive DC should normally serve requests for the whole time, and dead DC should normally serve requests after replacing the nodes.

fruch:
@aleksbykov @temichus what with this one ? if it's needed let finalize the review, if not please close it

PR #10712, opened 2025-04-28
fix(is_scylla_installed): stop using package manager

juliayakovlev:
Cherry-picked from https://github.com/scylladb/scylla-cluster-tests/pull/10690/commits/557467d5e7ed313fa5d37d6d3143477a916f76a4

Today we are using package manager to search if Scylla is installed Since we have in the repo now both `scylla` and
`scylla-enterprise`(which is a virtual package) the current implemntaion will identify as enterprise and will fail with the following error
```
There is no pre-installed ScyllaDB on ...
```

Modifying the logic to check for scylla version instead

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

fruch:
@juliayakovlev we'll need it also on v15, don't we ?

juliayakovlev:
> @juliayakovlev we'll need it also on v15, don't we ?

@fruch I opened for v15 as well - https://github.com/scylladb/scylla-cluster-tests/pull/10711

PR #10749, opened 2025-04-29
improvement(send_email): don't fail having no recipients

vponomaryov:
For the moment we always send emails after test runs.
Not all of them are needed and they just generate additional noise.

So, allow to avoid 'sending email' action by providing empty value for the recipient email parameter.

### Testing

- [scylla-staging/valerii/vp-perf-latency-regression-latte#31](https://argus.scylladb.com/tests/scylla-cluster-tests/2d87a0a7-bd91-4f11-8b46-206f98325d5c)

### Reminders

### DIFF
--- a/vars/runSendEmail.groovy
+++ b/vars/runSendEmail.groovy
@@ -20,12 +20,17 @@ def call(Map params, RunWrapper currentBuild){
     env
     echo "Start send email ..."
     RUNNER_IP=\$(cat sct_runner_ip||echo "")
-    if [[ -n "\${RUNNER_IP}" ]] ; then
-        ./docker/env/hydra.sh --execute-on-runner \${RUNNER_IP} send-email ${test_status} ${start_time} \
-        --runner-ip \${RUNNER_IP} --email-recipients "${email_recipients}"
+
+    if [[ -z "${email_recipients}" ]]; then
+        echo "Email was not sent because no recipient addresses were provided"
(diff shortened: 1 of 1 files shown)


vars/runSendEmail.groovy
```diff
@@ -20,11 +20,12 @@ def call(Map params, RunWrapper currentBuild){
     env
     echo "Start send email ..."
     RUNNER_IP=\$(cat sct_runner_ip||echo "")
+    EMAIL_RECIPIENTS_ARG=\$(if [[ -z "${email_recipients}" ]]; then echo ""; else echo "--email-recipients ${email_recipients}"; fi)
```
soyacz, line 23: why not just skipping send-email if no email recipients are provided?

vponomaryov: Do you mean skipping of the `send email` stage?
If so, then it will have side-effect on the `stage view` of each CI job.
If all stages are unchanged then we see 10 CI jobs by default.
When we change stages then we get displayed only the latest test run.

pehala: You could just not execute the send email command if no email recipients are present, without changing the stages

vponomaryov: Done

soyacz (approved):
LGTM

pehala (approved):

PR #10757, opened 2025-04-29
test(performance): add lz4 and zstd dict compression test

jsmolar:
Create a new job based on PerformanceRegressionPredefinedStepsTest to verify that dict compression won't bring unwanted latency degradation.

### Testing

- [x] [LZ4 with Dicts](https://argus.scylladb.com/tests/scylla-cluster-tests/0ac6090e-d199-4aa0-9648-6e14b110410e)
- [x] [ZSTD with Dicts](https://argus.scylladb.com/tests/scylla-cluster-tests/e36e365c-4e78-43b1-98c5-74b7094a0eac)

### Reminders

### DIFF
--- a/configurations/compression/lz4_dict_compression.yaml
+++ b/configurations/compression/lz4_dict_compression.yaml
@@ -1,0 +1,9 @@
+# Disabled speculative_retry to match the configuration of
+# scylla-enterprise-perf-regression-predefined-throughput-steps-tablets,
+# allowing this test to be used as a comparable benchmark.
+post_prepare_cql_cmds: >-
+  ALTER TABLE keyspace1.standard1
+  WITH compression = {
+    'sstable_compression': 'LZ4WithDictsCompressor'
+  }
+  AND speculative_retry = 'NONE';
--- a/configurations/compression/zstd_dict_compression.yaml
+++ b/configurations/compression/zstd_dict_compression.yaml
@@ -1,0 +1,9 @@
+# Disabled speculative_retry to match the configuration of
+# scylla-enterprise-perf-regression-predefined-throughput-steps-tablets,
+# allowing this test to be used as a comparable benchmark.
+post_prepare_cql_cmds: >-
+  ALTER TABLE keyspace1.standard1
+  WITH compression = {
+    'sstable_compression': 'ZstdWithDictsCompressor'
+  }
+  AND speculative_retry = 'NONE';
--- a/jenkins-pipelines/performance/branch-perf-v16/scylla-master/perf-features/compression/_display_name
+++ b/jenkins-pipelines/performance/branch-perf-v16/scylla-master/perf-features/compression/_display_name
@@ -1,0 +1,1 @@
+Scylla Dictionary Compression Feature Tests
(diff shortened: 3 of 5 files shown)


pehala (approved):

fruch:
@jsmolar why it's needed in pref tests branch ? i.e. those are not touch any of the test used for performance

jsmolar:
@fruch  After discussing with @pehala , we concluded there's no need to backport these changes to the perf branch.
@scylladb/qa-maintainers , do you see any benefit in having them there? The current plan is to make compression with dictionaries the default in the future, at which point this test will be removed or adjusted accordingly.

If there are no objections, I will go ahead and remove the `backport/perf-v16` label, and we can proceed with merging.

pehala (approved):

soyacz (approved):
No need to backport to perf branches - in case we test it, will require probably a different change
LGTM

soyacz:
see config file `configurations/disable_speculative_retry.yaml` contains `post_prepare_cql_cmds: "ALTER TABLE keyspace1.standard1 WITH speculative_retry = 'NONE'"` - if you add your config, this one will be overriden. Then we cannot compare both runs as will differ also on that aspect.
you should have combined one with commands in a list

fruch:
Just a note about the fact it's a performance test

Have it under the oss tree, i.e. part of each release folder, make the tracking of it across releases a manual process.

For the performance tests we recommend having them in one place (like master), and trigger new releases or dev releases in the same job, so it can be tracked across releases.

fruch:
> > Have it under the oss tree, i.e. part of each release folder, make the tracking of it across releases a manual process.
> > For the performance tests we recommend having them in one place (like master), and trigger new releases or dev releases in the same job, so it can be tracked across releases.
> 
> @fruch, this makes a lot of sense to me. So, what module is ideal for this?

so it should be:
1) backport it to v16 (or branch a new pref branch)
2) move the pipelines under `jenkins-pipelines/performance/branch-perf-v16/scylla-master/perf-regression` or a new directory under master (maybe `perf-features`) 
3) trigger manually as needed, and collect results of run in one job
4) optional: add triggers in scylla-pkg, so release automatically trigger this one  

that's how we treat the cases we are following on a weekly cadence (and on releases)

@soyacz we should also consider the use case of have those jobs in release, but still be able to track results across releases of same job

soyacz:
> 
> @soyacz we should also consider the use case of have those jobs in release, but still be able to track results across releases of same job

Having them in branches will make different sct versions across releases (different tools and who knows what else). 
Still I can add a task for it if needed.

fruch:
> Having them in branches will make different sct versions across releases (different tools and who knows what else).
> Still I can add a task for it if needed.

that's correct, but still it's flow that makes sense, no all performance test, we gonna run on a weekly cadence, and not every not on every release, but having ones that would be part of he release sequence as needed, does make sense to me.

configurations/compression/lz4_dict_compression.yaml
```diff
@@ -0,0 +1,6 @@
+post_prepare_cql_cmds: >-
```
soyacz, line 1: still I'd add some comment about speculative_retry just in case someone wants to use it in typical test and knowing why this is here.

soyacz (dismissed):
besides small comment,
LGTM

pehala (approved):

fruch (approved):
LGTM

PR #10758, opened 2025-04-29
fix(nemesis): enhance disrupt_mgmt_restore to restore schema out of Manager

mikliapko:
Sometimes the DC name of a cluster under test and DC name of a backup cluster might be different. In such a case, the regular schema restore using the Manager fails since DC names mismatch.

To work around this issue, the schema might be restored "manually" (or out of Manager) applying cql statements taken from the schema.json file stored in a backup snapshot.

### Testing

- [x] [Manager OPS Nemesis with simulated racks](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/longevity-100gb-manager-ops-4h/11/)
- [x] [Nemesis Manager restore](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/longevity-5gb-1h-MgmtRestore-aws-test/4/)
- [x] [Operator](https://argus.scylladb.com/tests/scylla-cluster-tests/64574458-c35f-4458-b8bb-8607df2d9540) - required `disrupt_mgmt_restore` has PASSED

### DIFF
--- a/sdcm/mgmt/helpers.py
+++ b/sdcm/mgmt/helpers.py
@@ -1,0 +1,84 @@
+import gzip
+import json
+import os
+import re
+from pathlib import Path
+
+from sdcm.utils.common import download_dir_from_cloud
+
+
+CREATE_KS_REPLICATION_BLOCK_PATTERN = re.compile(r"replication\s*=\s*({.*?})")
+REPLICATION_DICT_KEYS_PATTERN = re.compile(r"'([^']+)':")
+
+
+def get_schema_create_statements_from_snapshot(
+        bucket: str,
+        mgr_cluster_id: str,
+        snapshot_tag: str,
+        exclude_system_ks: bool = True,
+        exclude_roles: bool = True,
+) -> tuple[list[str], list[str]]:
+    """Parses schema information from a snapshot file and extracts CQL statements for keyspaces, tables, views and other
+    entities. Excludes entries related to system keyspaces and roles if the corresponding flags are set to True.
+
+    Important note: only AWS is supported for now.
+
+    Returns:
+        tuple[list[str], list[str]]: A tuple containing two lists:
+            - List of CQL statements for creating keyspaces.
+            - List of CQL statements for other entities (tables, views and others except excluded ones).
+    """
+    s3_dir_with_schema_json = f"s3://{bucket}/backup/schema/cluster/{mgr_cluster_id}"
+    temp_dir_with_schema_files = download_dir_from_cloud(url=s3_dir_with_schema_json)
+
+    # If several backups have been created for the same cluster, the downloaded dir will contain several files.
+    # Thus, the matching schema file should be found by the snapshot tag.
+    for filename in os.listdir(temp_dir_with_schema_files):
+        if snapshot_tag in filename:
+            schema_filename = Path(temp_dir_with_schema_files, filename)
+            break
+    else:
+        raise FileNotFoundError(f"Schema file for snapshot {snapshot_tag} not found in {s3_dir_with_schema_json}")
+
+    with gzip.open(schema_filename, 'rt', encoding='utf-8') as gz_file:
+        schema = json.load(gz_file)
+
+    keyspace_statements = []
+    other_entities_statements = []
+
+    for entry in schema:
+        if exclude_system_ks:
+            if entry["keyspace"].startswith("system"):
+                continue
+
+        if exclude_roles:
+            if entry["type"] == "role":
+                continue
+
+        if entry["type"] == "keyspace":
+            keyspace_statements.append(entry["cql_stmt"])
+        else:
+            other_entities_statements.append(entry["cql_stmt"])
+
+    return keyspace_statements, other_entities_statements
+
+
+def get_dc_name_from_ks_statement(ks_statement: str) -> list[str]:
+    """Extracts the name of a datacenter from a CREATE KEYSPACE statement.
+
+    For example, for the statement:
+
+        "CREATE KEYSPACE \"5gb_stcs_quorum_64_16_2024_2_4\"
+            WITH replication = {'class': 'org.apache.cassandra.locator.NetworkTopologyStrategy', 'us-east': '3'}
+            AND durable_writes = true
+            AND tablets = {'enabled': false};"
+
+    the function will return ["us-east"].
+    """
+    # Firstly, the code extracts the whole replication block from the statement.
+    # Then, it finds all keys in the replication dictionary and filters out the 'class' key.
+    # It doesn't search for dc_name in the replication block directly because the order of keys might be different.
+    replication_block = CREATE_KS_REPLICATION_BLOCK_PATTERN.search(ks_statement).group(1)
+    _keys = REPLICATION_DICT_KEYS_PATTERN.findall(replication_block)
+
+    return [key for key in _keys if key != 'class']
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -69,6 +69,7 @@ from sdcm.db_stats import PrometheusDBStats
 from sdcm.log import SDCMAdapter
 from sdcm.logcollector import save_kallsyms_map
 from sdcm.mgmt.common import TaskStatus, ScyllaManagerError, get_persistent_snapshots
+from sdcm.mgmt.helpers import get_dc_name_from_ks_statement, get_schema_create_statements_from_snapshot
 from sdcm.nemesis_publisher import NemesisElasticSearchPublisher
 from sdcm.prometheus import nemesis_metrics_obj
 from sdcm.provision.scylla_yaml import SeedProvider
@@ -2920,6 +2921,42 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                 stress_queue.append(read_thread)
             return stress_queue
 
+        def _restore_schema(locations: list, cluster_id: str, tag: str) -> None:
+            """Introduced to cover two flows:
+
+            - When a backup snapshot has different DC name than the current cluster's DC name -
+            restore schema out of the Manager applying CQL statements saved in schema.json file
+
+            - When backup snapshot has the same DC name as the current cluster's DC name -
+            restore schema using the Manager (sctool restore --restore-schema ...)
+            """
+            ks_statements, other_statements = get_schema_create_statements_from_snapshot(
+                bucket=locations[0].split(':')[-1],
+                mgr_cluster_id=cluster_id,
+                snapshot_tag=tag,
+            )
+
+            dc_under_test_name = next(iter(self.cluster.get_nodetool_status()))
+            # Test is supposed to work with single DC setups only, so we can take the first DC name
+            dc_from_backup_name = get_dc_name_from_ks_statement(ks_statements[0])[0]
+            self.log.debug("DC name from backup: %s, DC name under test: %s", dc_from_backup_name, dc_under_test_name)
+
+            if dc_under_test_name != dc_from_backup_name:
+                self.log.info("DC names mismatch - restoring the schema manually altering cql statements and "
+                              "applying them one by one")
+                # Alter the dc_name in keyspace cql statements to match the current cluster's dc_name
+                old_dc_block = f"'{dc_from_backup_name}':"  # Include quotes and colon to avoid unintended replacements
+                new_dc_block = f"'{dc_under_test_name}':"
+                ks_statements = [stmt.replace(old_dc_block, new_dc_block) for stmt in ks_statements]
+                # Apply cql statements one by one to restore schema
+                for cql_stmt in ks_statements + other_statements:
+                    self.target_node.run_cqlsh(cql_stmt)
+            else:
+                self.log.info("Restoring the schema using the Scylla Manager")
+                task = mgr_cluster.create_restore_task(restore_schema=True, location_list=locations, snapshot_tag=tag)
+                task.wait_and_get_final_status(step=10, timeout=6 * 60)  # 6 giving minutes to restore the schema
+                assert task.status == TaskStatus.DONE, f'Schema restoration failed: snapshot tag - {tag}'
+
         skip_issues = [
             "https://github.com/scylladb/scylla-manager/issues/3829",
             "https://github.com/scylladb/scylla-manager/issues/4049"
@@ -2951,16 +2988,12 @@ class Nemesis:  # pylint: disable=too-many-instance-attributes,too-many-public-m
         test_keyspaces = [cql_unquote_if_needed(keyspace) for keyspace in self.cluster.get_test_keyspaces()]
         # Keyspace names that start with a digit are surrounded by quotation marks in the output of a describe query
         if chosen_snapshot_info["keyspace_name"] not in test_keyspaces:
-            self.log.info("Restoring the schema of the keyspace '%s'",
-                          chosen_snapshot_info["keyspace_name"])
-            restore_task = mgr_cluster.create_restore_task(restore_schema=True, location_list=location_list,
-                                                           snapshot_tag=chosen_snapshot_tag)
-
-            restore_task.wait_and_get_final_status(step=10,
-                                                   timeout=6*60)  # giving 6 minutes to restore the schema
-            assert restore_task.status == TaskStatus.DONE, \
-                f'Schema restoration of {chosen_snapshot_tag} has failed!'
-
+            self.log.info("Restoring the schema of the keyspace '%s'", chosen_snapshot_info["keyspace_name"])
+            _restore_schema(
+                locations=location_list,
+                cluster_id=chosen_snapshot_info["cluster_id"],
+                tag=chosen_snapshot_tag,
+            )
             with ignore_ycsb_connection_refused():
                 self.cluster.restart_scylla()  # After schema restoration, you should restart the nodes
 


sdcm/mgmt/helpers.py
```diff
@@ -0,0 +1,55 @@
...
+    For example, for the statement:
+
+        "CREATE KEYSPACE \"5gb_stcs_quorum_64_16_2024_2_4\"
+            WITH replication = {'class': 'org.apache.cassandra.locator.NetworkTopologyStrategy', 'us-east': '3'}
+            AND durable_writes = true
+            AND tablets = {'enabled': false};"
+
+    the function will return "us-east".
+    """
+    replication_block = re.search(r"replication\s*=\s*({.*?})", ks_statement).group(1)
```
pehala, line 53: I would simply matched the key directly, no need to evaluate it. You can actually reuse pattern you use later.

sidenote: when using regex, please compile the statement only once

sdcm/nemesis.py
```diff
@@ -2920,6 +2921,39 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
+
+            - When backup snapshot has the same DC name as the current cluster's DC name -
+            restore schema using the Manager (sctool restore --restore-schema ...)
+            """
+            ks_statements, table_statements = get_schema_create_statements_from_snapshot(
+                bucket=locations[0].split(':')[-1],
+                mgr_cluster_id=cluster_id,
+            )
+
+            dc_under_test_name = list(self.cluster.get_nodetool_status().keys())[0]
```
pehala, line 2938: why do you need to convert to list? wouldnt `self.cluster.get_nodetool_status().keys()[0]` work as well?

sdcm/nemesis.py
```diff
@@ -2920,6 +2921,39 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
+        def _restore_schema(locations: list, cluster_id: str, tag: str) -> None:
+            """Introduced to cover two flows:
+
+            - When a backup snapshot has different DC name than the current cluster's DC name -
+            restore schema out of the Manager applying CQL statements saved in schema.json file
+
+            - When backup snapshot has the same DC name as the current cluster's DC name -
+            restore schema using the Manager (sctool restore --restore-schema ...)
+            """
+            ks_statements, table_statements = get_schema_create_statements_from_snapshot(
```
pehala, line 2933: given we have to download the snapshot, how long does that take and what are the common sizes? cant we run outofspace just because we downloaded it?

sdcm/nemesis.py
```diff
@@ -2920,6 +2921,39 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
+                bucket=locations[0].split(':')[-1],
+                mgr_cluster_id=cluster_id,
+            )
+
+            dc_under_test_name = list(self.cluster.get_nodetool_status().keys())[0]
+            dc_from_backup_name = get_dc_name_from_ks_statement(ks_statements[0])
+            self.log.debug("DC name from backup: %s, DC name under test: %s", dc_from_backup_name, dc_under_test_name)
+
+            if dc_under_test_name != dc_from_backup_name:
```
pehala, line 2942: What about doing it the other way, trying the standard restore first and if it fails we can do the manual process. We will save a bunch of work to determining the backup dc.

sdcm/nemesis.py
mikliapko: nope, we don't download the whole backup snapshot file but only the .json file with schema description

sdcm/nemesis.py
mikliapko: I don't think it makes sense here.
The whole validation process is very fast and took me 1 second to execute it.
```
< t:2025-04-30 11:44:03,171 f:nemesis.py      l:2989 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.MgmtRestore: Restoring the schema of the keyspace '10gb_stcs_quorum_64_16_2024_2_4'
< t:2025-04-30 11:44:03,172 f:common.py       l:1771 c:utils                p:INFO  > Downloading [s3://manager-backup-tests-permanent-snapshots-us-east-1/backup/schema/cluster/064c2b1c-99ad-4e7d-b3ad-0fd9591edd39] to [/tmp/download_from_cloud/3e0cb16437e2dc1fdc1ed141f57658bc]
< t:2025-04-30 11:44:03,477 f:common.py       l:1725 c:utils                p:INFO  > Downloading backup/schema/cluster/064c2b1c-99ad-4e7d-b3ad-0fd9591edd39/task_2dffc5ab-286b-4d08-b713-225a07656a9a_tag_sm_20250206093952UTC_schema_with_internals.json.gz from s3 to /tmp/download_from_cloud/3e0cb16437e2dc1fdc1ed141f57658bc/task_2dffc5ab-286b-4d08-b713-225a07656a9a_tag_sm_20250206093952UTC_schema_with_internals.json.gz
< t:2025-04-30 11:44:03,529 f:common.py       l:1784 c:utils                p:INFO  > Finished downloading [s3://manager-backup-tests-permanent-snapshots-us-east-1/backup/schema/cluster/064c2b1c-99ad-4e7d-b3ad-0fd9591edd39]
< t:2025-04-30 11:44:03,538 f:remote_base.py  l:560  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: Running command "/usr/bin/nodetool  status "...
< t:2025-04-30 11:44:03,852 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.0.86>:   41s    3933   38980      0 35ms   24ms   12ms   7ms    4.8ms  1.5ms  2.3ms  
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: Datacenter: us-east
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: ===================
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: Status=Up/Down
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: |/ State=Normal/Leaving/Joining/Moving
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: -- Address     Load      Tokens Owns Host ID                              Rack
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: UN 10.12.1.225 1.43 GB   256    ?    0a22f657-cb43-47a7-8978-9e634c5a2dab 1a  
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: UN 10.12.2.77  1.09 GB   256    ?    8fa9ac9b-f53e-4cad-834e-a210380fcf29 1a  
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: UN 10.12.3.121 511.04 KB 256    ?    e3b544b6-d9e3-4c9c-8476-7b53a4dde5f4 1a  
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: 
< t:2025-04-30 11:44:04,085 f:base.py         l:231  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: Note: Non-system keyspaces don't have the same replication settings, effective ownership information is meaningless
< t:2025-04-30 11:44:04,085 f:base.py         l:143  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.12.3.121>: Command "/usr/bin/nodetool  status " finished with status 0
< t:2025-04-30 11:44:04,108 f:db_log_reader.py l:126  c:sdcm.db_log_reader   p:DEBUG > 2025-04-30T11:44:04.037+00:00 longevity-5gb-1h-MgmtRestore-fix-sc-db-node-6f7f40b6-2     !INFO | scylla-cluster-tests[6219]: executing nodetool /usr/bin/nodetool  status  on longevity-5gb-1h-MgmtRestore-fix-sc-db-node-6f7f40b6-2 [10.12.3.121]
< t:2025-04-30 11:44:04,135 f:cluster.py      l:2684 c:sdcm.cluster_aws     p:DEBUG > Node longevity-5gb-1h-MgmtRestore-fix-sc-db-node-6f7f40b6-2 [98.80.142.135 | 10.12.3.121]: Command '/usr/bin/nodetool  status ' duration -> 0.5464648160000252 s
< t:2025-04-30 11:44:04,136 f:nemesis.py      l:2940 c:sdcm.nemesis         p:DEBUG > sdcm.nemesis.MgmtRestore: DC name from backup: us-east, DC name under test: us-east
< t:2025-04-30 11:44:04,136 f:nemesis.py      l:2952 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.MgmtRestore: Restoring the schema using the Scylla Manager
```

We definitely know that if DC names are not the same - the restore fails. 
I'd prefer to validate it at first and then act depending on validation result. I wouldn't like to run wittingly doomed to fail if it can easily avoided.

sdcm/nemesis.py
mikliapko: In python 3.x `dict.keys()` returns a `dict_keys` view object that should be converted to a list before indexing:
```
PyDev console: starting.
Python 3.12.2 (main, Sep 12 2024, 18:17:04) [GCC 11.4.0] on linux

>>> some_dict = {"a": 1, "b": 2}
>>> some_dict.keys()[0]
  File "/snap/pycharm-professional/475/plugins/python-ce/helpers/pydev/pydevconsole.py", line 364, in runcode
    coro = func()
           ^^^^^^
  File "<input>", line 1, in <module>
TypeError: 'dict_keys' object is not subscriptable
>>> type(some_dict.keys())
<class 'dict_keys'>
```

pehala: Ok, and could you access the key by name instead, that would make much more sense for a dict

sdcm/nemesis.py
pehala, line 2942: You should add condition for cloud as well, you mention that it works only on AWS for now

mikliapko: This is relevant for the whole `disrupt_mgmt_restore` test - we support only AWS for now.
So, don't think it make sense to introduce some extra conditions here if the test itself supports only one provider.

sdcm/nemesis.py
mikliapko: What do you mean by `access the key by name`?
I need to retrieve dict's key here, not value.

The dict returning by `get_nodetool_status()` has the following structure:
```
{
    "datacenter1": {
        "ip1": {
            'state': state,
            'load': load,
            'tokens': tokens,
            'owns': owns,
            'host_id': host_id,
            'rack': rack
        }
    }
...
```
This test doesn't support multiDC setup, so, I consider it's quite safe to use the approach above until we have one DC. For multiDC we'd need to refactor the whole test, so, most probably we won't skip this particular peace of code.

sdcm/mgmt/helpers.py
mikliapko: The thing is you don't need in advance what value will represent the DC name in `replication` dict to match the key directly. It can be `us-east`, `us-east-1` or any other value.

Could you please elaborate on this if I missed something?

mikliapko: So, actually, I did a few changes:
- replaced regexp with `str.replace()` for the `_restore_schema` function. It makes the code more readable and prevents from regexp utilization twice.
- regexp you mentioned is compiled before utilization

sdcm/nemesis.py
```diff
@@ -2920,6 +2921,39 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
+                self.log.info("DC names mismatch - restoring the schema manually altering cql statements and "
+                              "applying them one by one")
+                # Alter the dc_name in keyspace cql statements to match the current cluster's dc_name
+                pattern = r"(replication\s*=\s*\{.*?'class':\s*'[\w\.]+',\s*)'([^']+)'(\s*:\s*'[\d]+'.*?\})"
```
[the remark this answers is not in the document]
mikliapko: Not relevant anymore, replaced by str.replace()

sdcm/mgmt/helpers.py
```diff
@@ -0,0 +1,55 @@
...
+    replication_dict = ast.literal_eval(replication_block)
+    return list(replication_dict.keys())[1]
```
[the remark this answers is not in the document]
mikliapko: Applied

mikliapko:
Retesting after changes [here](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/longevity-100gb-manager-ops-4h/8/).

mikliapko:
@Michal-Leszczynski I'd really appreciate your review about manual schema restoration part.
The test confirms that it works but it'd be nice to get your approve as well.

mikliapko:
@grzywin This change should resolve an issue with restore in Operator.
May I please ask you to rerun your operator test using this branch and report the result here?

sdcm/nemesis.py
pehala: Dont we use it in nemesis as well?

mikliapko: We do, this `disrupt_mgmt_restore` is part of many nemesis tests we run. 
But I'm not aware of any that executes `disrupt_mgmt_restore` on GCP or Azure.

mikliapko: Moreover, we have backup snapshots in S3 only

sdcm/nemesis.py
pehala: Ah now it makes sense, thanks for explaining.

I would replace it then by next(iter(self.cluster.get_nodetool_status())), it is a minor improvement but it eliminates creating temporary list

sdcm/nemesis.py
```diff
@@ -2920,6 +2921,40 @@ def execute_data_validation_thread(command_template, keyspace_name, number_of_ro
...
+
+            dc_under_test_name = list(self.cluster.get_nodetool_status().keys())[0]
+            dc_from_backup_name = get_dc_name_from_ks_statement(ks_statements[0])
+            self.log.debug("DC name from backup: %s, DC name under test: %s", dc_from_backup_name, dc_under_test_name)
+
+            if dc_under_test_name != dc_from_backup_name:
+                self.log.info("DC names mismatch - restoring the schema manually altering cql statements and "
+                              "applying them one by one")
+                # Alter the dc_name in keyspace cql statements to match the current cluster's dc_name
+                old_dc_block = f"'{dc_from_backup_name}':"  # Include quotes and colon to avoid unintended replacements
```
pehala, line 2946: Can the `dc_name` be mentioned there multiple times?

mikliapko: It can be for multiDC setup, but we don't support such a configuration for restore nemesis for now.

pehala: If we dont expect multiple mentions, we dont need to add quotes and colons right and simplfiy the whole thing?

sdcm/mgmt/helpers.py
```diff
@@ -0,0 +1,58 @@
...
+    For example, for the statement:
+
+        "CREATE KEYSPACE \"5gb_stcs_quorum_64_16_2024_2_4\"
+            WITH replication = {'class': 'org.apache.cassandra.locator.NetworkTopologyStrategy', 'us-east': '3'}
+            AND durable_writes = true
+            AND tablets = {'enabled': false};"
+
+    the function will return "us-east".
+    """
+    pattern = re.compile(r"replication\s*=\s*({.*?})")
```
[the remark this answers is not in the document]
mikliapko: Fixed

sdcm/nemesis.py
mikliapko: I consider an approach with quotes and colons as a safer one since `dc_name` might be a part of keyspace name, for example.

sdcm/nemesis.py
mikliapko: Make sense, improved

sdcm/mgmt/helpers.py
```diff
@@ -0,0 +1,58 @@
...
+    replication_block = pattern.search(ks_statement).group(1)
+    replication_dict = ast.literal_eval(replication_block)
```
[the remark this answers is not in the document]
pehala: For that regex should suffice as well, you can parse every key/value from that dict and use the same logic as now to filter correct keys (which is also not safe given you rely on having two keys there as well), making it simpler and safer.

Even in official documentation https://docs.python.org/3/library/ast.html#ast.literal_eval literal_eval has a big red box saying it is not safe and can crash interpreter and throw various weird errors. In general, it is better to not rely on it at all.

sdcm/nemesis.py
pehala: makes sense

sdcm/mgmt/helpers.py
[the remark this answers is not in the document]
mikliapko: fixed

pehala (dismissed):

sdcm/mgmt/helpers.py
```diff
@@ -0,0 +1,62 @@
...
+    Important note: only AWS is supported for now.
+
+    Returns:
+        tuple[list[str], list[str]]: A tuple containing two lists:
+            - List of CQL statements for creating keyspaces.
+            - List of CQL statements for creating tables.
+    """
+    s3_dir_with_schema_json = f"s3://{bucket}/backup/schema/cluster/{mgr_cluster_id}"
+    temp_dir_with_schema_file = download_dir_from_cloud(url=s3_dir_with_schema_json)
+    schema_file = Path(temp_dir_with_schema_file, os.listdir(temp_dir_with_schema_file)[0])
```
Michal-Leszczynski, line 27: There is a hidden assumption that there is only a single schema file in the backup, which means that there was only a single backup made for this cluster. I would suggest to either highlight it in the function name/comment, or accept additional arg `snapshot_tag` which could be used to identify the desired schema file. The simplest way would be to just check if the schema file contains the `snapshot_tag` in its name.

sdcm/mgmt/helpers.py
```diff
@@ -0,0 +1,62 @@
...
+
+    the function will return "us-east".
+    """
+    # Firstly, the code extracts the whole replication block from the statement.
+    # Then, it finds all keys in the replication dictionary and filters out the 'class' key.
+    # It doesn't search for dc_name in the replication block directly because it might not be the second key.
+    replication_block = CREATE_KS_REPLICATION_BLOCK_PATTERN.search(ks_statement).group(1)
+    _keys = REPLICATION_DICT_KEYS_PATTERN.findall(replication_block)
+
+    return next(iter(key for key in _keys if key != 'class'))
```
Michal-Leszczynski, line 62: There is a hidden assumption that ks is replicated only in a signle DC. I would suggest to highlight it in the function name/comment or make this function return a list of all DCs.

sdcm/mgmt/helpers.py
```diff
@@ -0,0 +1,62 @@
...
+        schema = json.load(gz_file)
+
+    keyspace_statements = []
+    table_statements = []
+
+    for entry in schema:
+        if entry["type"] == "keyspace" and not entry["keyspace"].startswith("system"):
+            keyspace_statements.append(entry["cql_stmt"])
+        elif entry["type"] == "table" and not entry["keyspace"].startswith("system"):
+            table_statements.append(entry["cql_stmt"])
```
Michal-Leszczynski, line 39: Why do we want to restore just the keyspaces and tables? What about views, UDTs, and other entities?
In the context of altering schema to match new DCs, it makes sense to separate create keyspace statements from all the other statements, but we don't need to limit ourselves to tables only.

mikliapko: Good point. Thanks.
The current datasets under tests don't include views or other entities (only keyspaces and tables) but implementing it in a way you suggested makes it more universal and sustainable to dataset changes.
Will improve.

sdcm/mgmt/helpers.py
mikliapko: Will fix to return the list of dc names in helper function.
The test itself since support only single DC setups will take the first list element.

sdcm/mgmt/helpers.py
mikliapko: Good point.
Will add an extra step to look for the file with snapshot_tag in filename

mikliapko:
Retesting with the latest changes [here](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/longevity-100gb-manager-ops-4h/11/).

mikliapko:
> Retesting with the latest changes [here](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/longevity-100gb-manager-ops-4h/11/).

PASSED

Michal-Leszczynski (approved):

pehala (approved):

mikliapko:
@scylladb/qa-maintainers Can we proceed with that?

fruch (approved):
LGTM

Issue #10562, opened 2025-04-02
disrupt_truncate_large_partition nemesis failure with no attribute error


grzywin:
Some time ago following PR was merged to SCT https://github.com/scylladb/scylla-cluster-tests/pull/10335 where among others `verify_stress_thread` method was introduced. Since then, during execution of longevity tests for Operator, one of the nemesis called `disrupt_truncate_large_partition` is failing with following error:

```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5693, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2192, in disrupt_truncate_large_partition
    self.tester.verify_stress_thread(bench_thread, error_handler=self._nemesis_stress_failure_handler)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 2354, in verify_stress_thread
    error_handler(thread_pool, errors)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 2144, in _nemesis_stress_failure_handler
    raise NemesisStressFailure(
sdcm.exceptions.NemesisStressFailure: Aborting 'SisyphusMonkey' nemesis as stress command failed with the following errors:
 on node 'sct-loaders-eu-central-1-0': ["Stress command execution failed with: 'NoneType' object has no attribute 'closed'"]
 on node 'sct-loaders-eu-central-1-1': ["Stress command execution failed with: 'NoneType' object has no attribute 'closed'"]
```

Example test run:
https://argus.scylladb.com/tests/scylla-cluster-tests/44e7b05a-d33c-4623-8bfa-b3b190fa3f06

dimakr:
The check added in the [PR#10335](https://github.com/scylladb/scylla-cluster-tests/pull/10335) is checking if all stress threads (that are being executed on a Nemesis) failed. This to avoid a situation when a set of multihour stress threads are running even though the stress commands already failed inside all threads.
Here the check catches such a situation - all threads during the disruption failed and returned errors. The problem is that this error says little about what happened, but the check cannot do much about it - it just prints out the response from stress thread (it could be a response from the stress tool itself or from SCT *StressThread abstraction around it).

Quick grep shows that scylla-bench code has no `closed` attributes, vars, but SCT k8s related code has. So this is probably the direction to investigate in.

[...]

PR #10759, opened 2025-04-29
fix(k8s-pod-watcher): check for uninitialized watcher in conditionals

dimakr:
The change adds check to ensure that the watcher process object exists, before accessing its 'closed' attribute. This prevents potential "NoneType object has no attribute 'closed'" exceptions when the process is not yet initialized.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/10562

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/remote/kubernetes_cmd_runner.py
+++ b/sdcm/remote/kubernetes_cmd_runner.py
@@ -304,7 +304,7 @@ class KubernetesPodWatcher(KubernetesRunner):
     def _read_from_stream(self, num_bytes: int) -> str:
         pod_name = self.context.config.k8s_pod_name
         with self._ws_lock:
-            if self.process.closed:
+            if not self.process or self.process.closed:
                 self.pod_counter_to_live -= 1
                 if self.pod_counter_to_live < 1:
                     LOGGER.warning(
@@ -482,7 +482,9 @@ class KubernetesPodWatcher(KubernetesRunner):
 
     @property
     def process_is_finished(self) -> bool:
-        if not self.process.closed:
+        if self.process is None:
+            return True
+        elif not self.process.closed:
             return False
         return self._is_pod_failed_or_completed()
 
@@ -512,10 +514,10 @@ class KubernetesPodWatcher(KubernetesRunner):
                 LOGGER.warning(
                     "'stop()' method is called for the '%s' pod as part of the 'tearDown'. "
                     "'log reader' socket is '%s'.",
-                    pod_name, "closed" if self.process.closed else "open")
+                    pod_name, "closed" if not self.process or self.process.closed else "open")
                 self._stop_pod()
                 return
-            if self.process.closed:
+            if not self.process or self.process.closed:
                 LOGGER.warning(
                     "'stop()' method is called for the '%s' pod, which is running "
                     "having closed 'log reader' socket. Which is probably the cause for "


dimakr:
Now this change cannot be checked, as the scenario from the original issue is failing with the error `botocore.errorfactory.InvalidParameterException: An error occurred (InvalidParameterException) when calling the CreateAddon operation: Addon version specified is not supported` (which should be resolved once https://github.com/grzywin/scylla-cluster-tests/tree/fix-tests-for-operator branch changes are merged to master)

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -304,7 +304,7 @@ def _open_stream(self) -> None:
     def _read_from_stream(self, num_bytes: int) -> str:
         pod_name = self.context.config.k8s_pod_name
         with self._ws_lock:
-            if self.process.closed:
+            if self.process and self.process.closed:
```
vponomaryov, line 307: This changes the logic.
In this line change the `self.process = None` behaves like `self.process.closed = False`.
And it is completely incorrect.

The `self.process = None` must behave like `self.process.closed = True`.

Then, if we have `self.process = None` we pass through down to the line 330 in this file.
And on line 330 we will get identical error - `AttributeError` where `None` type object doesn't have `read` attribute.

So, technically, it doesn't fix the problem.

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -512,10 +512,10 @@ def stop(self) -> None:
                 LOGGER.warning(
                     "'stop()' method is called for the '%s' pod as part of the 'tearDown'. "
                     "'log reader' socket is '%s'.",
-                    pod_name, "closed" if self.process.closed else "open")
+                    pod_name, "closed" if self.process and self.process.closed else "open")
```
vponomaryov, line 515: Same here.
If the `self.process` is `None`, we should not say in logs that it is in `open` state.

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -512,10 +512,10 @@ def stop(self) -> None:
...
                 self._stop_pod()
                 return
-            if self.process.closed:
+            if self.process and self.process.closed:
```
vponomaryov, line 518: Same here.

vponomaryov (changes requested):
See inline comments

sdcm/remote/kubernetes_cmd_runner.py
dimakr: @vponomaryov I see.
so the proper condition should be `if not self.process or self.process.closed` then. In all 3 cases

sdcm/remote/kubernetes_cmd_runner.py
```diff
@@ -482,7 +482,7 @@ def _write_proc_stdin(self, data: str) -> None:
 
     @property
     def process_is_finished(self) -> bool:
-        if not self.process.closed:
+        if not self.process or not self.process.closed:
```
vponomaryov, line 485: This one must have a bit different logic.
`None` value for the `self.process` means we have no running process,
so `process_is_finished` should be `True` in this case:

```
if self.process is None:
    return True
elif not self.process.closed:
    return False
return self._is_pod_failed_or_completed()
```

sdcm/remote/kubernetes_cmd_runner.py
dimakr: Changed the conditiion

sdcm/remote/kubernetes_cmd_runner.py
dimakr: Changed the conditiion

sdcm/remote/kubernetes_cmd_runner.py
dimakr: Changed the conditiion

sdcm/remote/kubernetes_cmd_runner.py
dimakr: Changed the conditiion

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #10767, opened 2025-04-30
fix(coredump): handle cases where dump doesn't have exceutable name

fruch:
seems like in some situation coredump might not have executable, we should collec that coredmp regardless

### Testing

- [ ]

### Reminders


dimakr (approved):
LGTM

vponomaryov (approved):
LGTM

soyacz:
I don't think it's going to work... core is None type so `.get()` will fail anyway

PR #10814, opened 2025-05-05
fix(builders): limit executors to one

fruch:
since we are running into multiple issue with builders running out of memory during artifact tests, we are taking this number down from 4 to 1, until we can figure out a better solution for artifact tests (it's own builders ?)

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/aws_builder.py
+++ b/sdcm/utils/aws_builder.py
@@ -96,7 +96,7 @@ jenkins.save()
 
 class AwsBuilder:
     NUM_CPUS = 2
-    NUM_EXECUTORS = 4
+    NUM_EXECUTORS = 1
     VERSION = 'v3'
 
     def __init__(self, region: AwsRegion, params=None, number=1):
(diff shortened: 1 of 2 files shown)


soyacz:
which machines are cheaper - builders or sct-runners? maybe we should use sct-runner for these?

fruch:
> which machines are cheaper - builders or sct-runners? maybe we should use sct-runner for these?

builders are cheaper (2 cpu 4Gb memory), but we'll need builders anyhow for anything we with jenkins.

vponomaryov (approved):
LGTM

soyacz:
> > which machines are cheaper - builders or sct-runners? maybe we should use sct-runner for these?
> 
> builders are cheaper (2 cpu 4Gb memory), but we'll need builders anyhow for anything we with jenkins.

I was just thinking if keeping builders workers as is and spin sct-runner for artifact test, but in that case it is not feasible.

fruch:
> > > which machines are cheaper - builders or sct-runners? maybe we should use sct-runner for these?
> > 
> > 
> > builders are cheaper (2 cpu 4Gb memory), but we'll need builders anyhow for anything we with jenkins.
> 
> I was just thinking if keeping builders workers as is and spin sct-runner for artifact test, but in that case it is not feasible.

it's feasible, but then we'll have a builder that can serve multiple tests, and runner per test, which ~double the cost per test.

before this change we could share the builder, and not we can't. so it's also increase costs

it's hard to compare the two options, and this one is faster to apply

PR #10817, opened 2025-05-05
test(perf): add latency steady state test

vponomaryov:
We plan to start running complex stress commands as part of performance
tests for such scenarios as `custom-d1/w1` and `custom-d1/w2`.
Those use dozens of latte/rune functions which use different HDR tags.
    
So, to be able to test performance, add latency steady state test.
Also, make this test support multiple workload types with multiple HDR tags.

### Testing

- [scylla-staging/valerii/vp-perf-latency-regression-latte#36](https://argus.scylladb.com/tests/scylla-cluster-tests/b4fd90bf-5825-4b4e-b59e-db5d2b1f6aff)
- [scylla-staging/valerii/vp-scylla-master-perf-regression-latency-650gb-with-nemesis#17](https://argus.scylladb.com/tests/scylla-cluster-tests/baa1e3f7-0704-41a1-b81a-25949cf0690f)

### Reminders

### DIFF
--- a/performance_regression_test.py
+++ b/performance_regression_test.py
@@ -27,6 +27,7 @@ from sdcm.sct_events import Severity
 from sdcm.sct_events.filters import EventsSeverityChangerFilter
 from sdcm.sct_events.loaders import CassandraStressEvent
 from sdcm.sct_events.system import HWPerforanceEvent, InfoEvent
+from sdcm.utils.common import ParallelObject
 from sdcm.utils.decorators import log_run_info, latency_calculator_decorator, optional_stage
 from sdcm.utils.nemesis_utils.indexes import wait_for_view_to_be_built
 
@@ -610,6 +611,56 @@ class PerformanceRegressionTest(ClusterTester, loader_utils.LoaderUtilsMixin):  
         self.wait_no_compactions_running()
         self.run_fstrim_on_all_db_nodes()
         self.run_mixed_workload()
+
+    def test_latency_steady_state(self):
+        """Test designed to run multiple stress commands, possibly, using different stress operation types.
+
+        For example, 'latte' uses rune function names in HDR histogram tags,
+        so we should gather all the unique tags into separate lists - one per stress operation type
+        """
+        self.run_fstrim_on_all_db_nodes()
+        self.preload_data()
+        self.wait_no_compactions_running()
+        self.run_fstrim_on_all_db_nodes()
+
+        stress_operation_mapping = {}
+        for stress_cmd in self.params.get("stress_cmd"):
+            stress_thread = self.run_stress_thread(stress_cmd=stress_cmd, stress_num=1, round_robin=True)
+            stress_op = stress_thread.stress_operation  # test depends on the 'stress_operation' attr
+            assert stress_op, "stress operation type should not be empty: %s" % stress_op
+            if stress_op not in stress_operation_mapping:
+                stress_operation_mapping[stress_op] = []
+            stress_operation_mapping[stress_op].append(stress_thread)
+
+        def _test_latency_steady_state_template(workload: str, *args):
+            tester, stress_queue = args  # blindly assume 2 args
+
+            # NOTE: "steady" word from the func name will be used by the latency calculator decorator
+            @latency_calculator_decorator(workload_type=workload)
+            def _test_latency_steady_state(tester, stress_queue: list):
+                for stress_thread in stress_queue:
+                    self.verify_stress_thread(stress_thread)
+
+                # NOTE: 'hdr_tags' will be used by the latency calculator decorator
+                hdr_tags = []
+                for stress_thread in stress_queue:
+                    hdr_tags.extend(stress_thread.hdr_tags)
+                return {"hdr_tags": hdr_tags}  # must be dict with 'hdr_tags' key
+
+            # NOTE: 'tester' arg must be first and positional due to latency calculator decorator expectations
+            return _test_latency_steady_state(tester, stress_queue=stress_queue)
+
+        object_set = ParallelObject(
+            timeout=None,
+            objects=[
+                [_stress_op, self, _stress_queue]
+                for _stress_op, _stress_queue in stress_operation_mapping.items()
+            ],
+            num_workers=len(stress_operation_mapping))
+        object_set.run(
+            func=_test_latency_steady_state_template,
+            unpack_objects=True,
+            ignore_exceptions=False)
 
     def test_latency_read_with_nemesis(self):
         self.run_fstrim_on_all_db_nodes()
--- a/performance_regression_test.py
+++ b/performance_regression_test.py
@@ -17,8 +17,6 @@
 import os
 import time
 
-from enum import Enum
-
 import yaml
 from cassandra.query import SimpleStatement  # pylint: disable=no-name-in-module
 
@@ -33,12 +31,6 @@ from sdcm.utils.decorators import log_run_info, latency_calculator_decorator, op
 from sdcm.utils.nemesis_utils.indexes import wait_for_view_to_be_built
 
 KB = 1024
-
-
-class PerformanceTestWorkload(Enum):
-    WRITE = "write"
-    READ = "read"
-    MIXED = "mixed"
 
 
 class PerformanceRegressionTest(ClusterTester, loader_utils.LoaderUtilsMixin):  # pylint: disable=too-many-public-methods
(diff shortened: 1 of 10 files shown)


vponomaryov:
Regression testing:
Screenshot from 2025-05-05 15-41-37

New approach testing:
Screenshot from 2025-05-05 15-41-12

sdcm/argus_results.py
```diff
@@ -171,51 +171,86 @@ def submit_results_to_argus(argus_client: ArgusClient, result_table: GenericResu
             raise
 
 
-def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, description: str, cycle: int, result: dict,
-                         start_time: float = 0, error_thresholds: dict = None):
-    result_table = workload_to_table[workload]()
+def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, description: str,  # noqa: PLR0914
```
soyacz, line 174: we do too much in this function...
How about keeping `send_result_to_argus` as close as possible to original (make `cycle` a `str`) and introduce new function for sending summary (or function to prepare summary as result dict and reuse this func)?
at least, add some docs how it works.

sdcm/stress/base.py
```diff
@@ -53,11 +53,15 @@ def __init__(self, loader_set, stress_cmd, timeout, stress_num=1, node_list=None
         self.shutdown_timeout = 180  # extra 3 minutes
         self.stop_test_on_failure = stop_test_on_failure
         self.hdr_tags = []
+        self.stress_operation = self.set_stress_operation(stress_cmd)
 
         if "k8s" not in self.params.get("cluster_backend") and self.docker_image_name:
             for loader in self.loader_set.nodes:
                 RemoteDocker.pull_image(loader, self.docker_image_name)
 
+    def set_stress_operation(self, stress_cmd):
```
soyacz, line 62: maybe `NotImplemented`? so when used it raises straight away instead of asserting is not empty when used?

soyacz (commented):
My main question is, why we need separate test and not to reuse `steady_state_latency` method like we do in upgrade/perf with nemesis tests?

sdcm/argus_results.py
fruch: I'm not sure splitting this would make it any easier, and I don't want to read the HDR data multiple times
as long we have enough unittest to cover this complexity, I think it's reasonable.

vponomaryov: > we do too much in this function... How about keeping `send_result_to_argus` as close as possible to original (make `cycle` a `str`) 

Parameters were not changed. The `cycle` was `int` and stayed `int`.
It is kept as close as possible to the original with changes needed for new scenario - multiple stress commands.

> and introduce new function for sending summary (or function to prepare summary as result dict and reuse this func)? 

Any split will lead to a redundancy in the data processing.

> at least, add some docs how it works.

Ok, will add docstring.

sdcm/stress/base.py
vponomaryov: The design is made the way that this function is called on the class instantiation step.
Then, this new attribute gets used only in performance tests.

So, If I do as suggested then I will need to implement this function for every benchmarking tool, supported by SCT, even if those are not used in performance.
I don't see any sense in it.

performance_regression_test.py
```diff
@@ -29,18 +27,13 @@
...
 from sdcm.sct_events.loaders import CassandraStressEvent
 from sdcm.sct_events.system import HWPerforanceEvent, InfoEvent
+from sdcm.utils.common import ParallelObject
 from sdcm.utils.decorators import log_run_info, latency_calculator_decorator, optional_stage
 from sdcm.utils.nemesis_utils.indexes import wait_for_view_to_be_built
 
 KB = 1024
 
 
-class PerformanceTestWorkload(Enum):
```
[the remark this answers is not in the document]
vponomaryov: It is not correct to have such a single enum for all of the benchmarking tools when they support different sets of values.
So, having enum like this is not `it's easier to know what operations are supported`, it is `illusion of it's support, because we need to check each tool separately`

performance_regression_test.py
```diff
@@ -631,6 +612,56 @@ def test_latency(self):
...
+        for stress_cmd in self.params.get("stress_cmd"):
+            stress = self.run_stress_thread(stress_cmd=stress_cmd, stress_num=1, round_robin=True)
+            stress_op = stress.stress_operation  # test depends on the 'stress_operation' attr
+            assert stress_op, "stress operation type should not be empty: %s" % stress_op
+            if stress_op not in stress_operation_mapping:
+                stress_operation_mapping[stress_op] = []
+            stress_operation_mapping[stress_op].append(stress)
+
+        def _test_latency_steady_state_template(workload: str, *args):
+            self, stress_queue = args  # blindly assume 2 args
```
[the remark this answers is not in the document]
vponomaryov: Ok, will change.

performance_regression_test.py
```diff
@@ -631,6 +612,56 @@ def test_latency(self):
...
+        so we should gather all the unique tags into separate lists - one per stress operation type
+        """
+        self.run_fstrim_on_all_db_nodes()
+        self.preload_data()
+        self.wait_no_compactions_running()
+        self.run_fstrim_on_all_db_nodes()
+
+        stress_operation_mapping = {}
+        for stress_cmd in self.params.get("stress_cmd"):
+            stress = self.run_stress_thread(stress_cmd=stress_cmd, stress_num=1, round_robin=True)
```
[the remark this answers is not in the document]
vponomaryov: Ok, will change.

vponomaryov (commented):
> My main question is, why we need separate test and not to reuse steady_state_latency method like we do in upgrade/perf with nemesis tests?

- SCT must run some test, this is test. The referenced methods are just `sleepers` in scope of `some test`. So, it is not applicable `as is` from this point of view.
- The `nemesis_interval` is not applicable here, the `duration` is based on stress commands.
- In this case the `workload type` gets specified explicitly for the latency decorator

soyacz:
these screenshots for each hdr tag looks redundant - probably repeating what is in Summary table. I propose to remove them.

vponomaryov:
> these screenshots for each hdr tag looks redundant - probably repeating what is in Summary table. I propose to remove them.

All of the commands are different query types, different payload.
In custom-d1/w2 scenario part of workloads are large partitions...

So, we do not gain anything removing per-command results
and only lose when we want to get to know which exactly command(s) was/were problematic.

There is filter which allows to show only summary tables.

soyacz:
> > these screenshots for each hdr tag looks redundant - probably repeating what is in Summary table. I propose to remove them.
> 
> All of the commands are different query types, different payload. In custom-d1/w2 scenario part of workloads are large partitions...
> 
> So, we do not gain anything removing per-command results and only lose when we want to get to know which exactly command(s) was/were problematic.
> 
> There is filter which allows to show only summary tables.

I meant sending Overview and QA dashboard columns - these are the same for all the rows, and probably the same as in Summary.

vponomaryov:
> > > these screenshots for each hdr tag looks redundant - probably repeating what is in Summary table. I propose to remove them.
> > 
> > 
> > All of the commands are different query types, different payload. In custom-d1/w2 scenario part of workloads are large partitions...
> > So, we do not gain anything removing per-command results and only lose when we want to get to know which exactly command(s) was/were problematic.
> > There is filter which allows to show only summary tables.
> 
> I meant sending Overview and QA dashboard columns - these are the same for all the rows, and probably the same as in Summary.

Actually, `start time`, `duration`, `qa dashboard` and `overview` values must be printed per table and not per row.
But it is existing approach, I didn't change the table structure.

If I remove these cell values then filtering out the "summary" table we will not have these values in detailed table view.
So, I undersand that it is duplication, but prefer to keep it while the detailed table view is not changed the way to show these values only once.

vponomaryov:
Current view for latte:

Screenshot from 2025-05-13 20-12-32

Regression testing for existing CI job:

Screenshot from 2025-05-13 20-13-56

fruch:
@vponomaryov you gonna introduce the job for in a followup?

vponomaryov:
> @vponomaryov you gonna introduce the job for in a followup?

Yes, in a separate PR

soyacz (approved):
LGTM

vponomaryov:
> > @vponomaryov you gonna introduce the job for in a followup?
> 
> Yes, in a separate PR

Done:
- https://github.com/scylladb/scylla-cluster-tests/pull/10904

sdcm/utils/decorators.py
```diff
@@ -231,7 +231,9 @@ def wrapped(*args, **kwargs):  # noqa: PLR0914
                 return res
             monitor = monitoring_set.nodes[0]
             screenshots = monitoring_set.get_grafana_screenshots(node=monitor, test_start_time=start)
-            if 'read' in test_name:
+            if workload_type:
```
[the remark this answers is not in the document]
vponomaryov: `workload_name` is not defined/used in any of configuration files.
So, this one may be cut out of the SCT code.

The local var called `workload` can be renamed to something like `current_workload_type`, yes.

Anyway, this change is already merged.

Feel free to file SCT bug about it to have discussion in a specific place.

Issue #10766, opened 2025-04-30
hydra.sh integration-tests fail with bad command exit code


radoslawcybulski:
Running this command:
```
./docker/env/hydra.sh integration-tests --test test_ycsb_thread.py::test_01_dynamodb_api
```
fails on fedora 41 with some bad command exit code error.

Full log of running command:
```
y@fedora:~/work/scylla-cluster-tests$ ./docker/env/hydra.sh integration-tests --test test_ycsb_thread.py::test_01_dynamodb_api
There is scylladb/hydra:v1.94-PR10407-af08406 in local cache, using it.
Going to run './sct.py  integration-tests --test test_ycsb_thread.py::test_01_dynamodb_api'...
WARN[0000] Using the Podman API service with TCP sockets is not recommended, please see `podman system service` manpage for details 
logged in as arn:aws:sts::797456418907:assumed-role/DeveloperAccessRole/radoslaw.cybulski@scylladb.com
When Podman daemon is used we don't login
New directory created: /home/y/sct-results/20250430-105713-004717
Symlink `/home/y/sct-results/latest' updated to `/home/y/sct-results/20250430-105713-004717'
New directory created: /home/y/sct-results/20250430-105713-005866-integration-tests
Failed to load configuration files: []
The run can be interrupted by following critical events:
  * ClusterHealthValidatorEvent.NodeStatus
  * ClusterHealthValidatorEvent.ScyllaCloudClusterServerDiagnostic
  * DataValidatorEvent.UpdatedRowsValidator
  * DatabaseLogEvent.CORRUPTED_SSTABLE
  * CassandraHarryEvent.failure
  * YcsbStressEvent.failure
  * NdBenchStressEvent.failure
  * NdBenchErrorEvent.BuildFailed
  * NdBenchErrorEvent.Failure
  * CDCReaderStressEvent.failure
  * NoSQLBenchStressEvent
  * NoSQLBenchStressLogEvents.ProgressIndicatorStoppedEvent
  * CassandraStressLogEvent.OperationOnKey
  * CqlStressCassandraStressLogEvent.ReadValidationError
  * ScyllaBenchLogEvent.DataValidationError
  * ScyllaBenchLogEvent.ParseDistributionError
  * GeminiStressLogEvent.GeminiEvent
  * PrometheusAlertManagerEvent
  * TestTimeoutEvent
  * TestFrameworkEvent
  * SpotTerminationEvent
  * ScyllaBenchEvent
[...]

PR #10824, opened 2025-05-05
fix(integration-test): stop using sudo under podman

fruch:
docker base loader were using sudo for some operations, and we can
run sudo currently under podman, so we need to identify it's podman
we are using, and stop calling sudo

Tested with:
```
HYDRA_TOOL=podman bash -x ./docker/env/hydra.sh python -m pytest unit_tests/test_ycsb_thread.py::test_01_dynamodb_api
HYDRA_TOOL=podman bash -x ./docker/env/hydra.sh integration-tests --test test_ycsb_thread.py::test_01_dynamodb_api
```

Fix: https://github.com/scylladb/scylla-cluster-tests/issues/10766

### Testing

- [x] locally

### Reminders

### DIFF
--- a/sdcm/utils/docker_remote.py
+++ b/sdcm/utils/docker_remote.py
@@ -74,15 +74,21 @@ class RemoteDocker(BaseNode):
         except (ValueError, socket.herror):
             return self.external_address
 
-    @cached_property
-    def running_in_docker(self):
-        ok = self.node.remoter.run("test /.dockerenv", ignore_status=True).ok
-        ok |= 'docker' in self.node.remoter.run('ls /proc/self/cgroup', ignore_status=True).stdout
+    @staticmethod
+    @cache
+    def running_in_docker(node):
+        ok = node.remoter.run("test /.dockerenv", ignore_status=True).ok
+        ok |= 'docker' in node.remoter.run('ls /proc/self/cgroup', ignore_status=True).stdout
         return ok
+
+    @staticmethod
+    @cache
+    def running_in_podman(node) -> bool:
+        return 'podman' in node.remoter.run('echo $container', ignore_status=True).stdout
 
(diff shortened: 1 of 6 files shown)


fruch:
**just to note**

this fixes mostly skipping the sudo usages.

it doesn't not fixes the communication issues with podman

we can't yet communicate from the host/test directly to the addresses of the podman containers (scylla or any other) 
and change it to work across the board with host ports and addresses is not simple as it might appear.
found one way to expose the networking to the host, but it's require at least one container running prior to starting the hydra container... :(

sdcm/utils/docker_remote.py
```diff
@@ -61,7 +61,7 @@ def private_ip_address(self):
 
     @property
     def cql_address(self):
-        return self.internal_ip_address
+        return self.internal_ip_address  # '127.0.0.1' if self.running_in_podman(self.node) else
```
vponomaryov, line 64: This one looks like a debug-leftover.

vponomaryov (commented):
Small nit, everything else LGTM

fruch:
Gemini failure in integration tests are not related to those changes (they are not stable for some time now)

sdcm/utils/docker_remote.py
fruch: it's a left over from when I was tried to overcome the issue of network with podman, which I gave up on solving in this PR

fruch: I'll remove it

fruch: removed

fruch:
> Gemini failure in integration tests are not related to those changes (they are not stable for some time now)

add one more commit to try addressing the gemini issue, or at least go around it.

Issue #10610, opened 2025-04-07
Update test to have more instance per rack longevity-1tb-5days-azure-test


aleksbykov:
longevity-1TB-5days-authorization-and-tls-ssl.yaml

## Packages

Scylla version: `2025.2.0~dev-20250403.3760a1c85e83` with build-id `c9f443363fcd8263d96dee717aa9ff2907c6ad8d`
Kernel Version: `6.11.0-1012-azure`

## Issue description

- [ ] This issue is a regression.
- [x] It is unknown if this issue is a regression.

Test configured with tablets and racks, but there is not enough nodes per ack. 
Config: longevity-1TB-5days-authorization-and-tls-ssl.yaml
only 1 node per rack. These configuration cause topology operation decommission/removenode to fail, because not enough nodes in rack to migrate tablets.

Test should be updated with more nodes per rack

## Impact

*Describe the impact this issue causes to the user.*

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 3 nodes (Standard_L16s_v3)

Scylla Nodes used in this run:
- longevity-tls-1tb-7d-master-db-node-5d1aa1d0-eastus-1 (null | 10.0.0.5) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-5d1aa1d0-eastus-2 (null | 10.0.0.6) (shards: 14)
- longevity-tls-1tb-7d-master-db-node-5d1aa1d0-eastus-3 (null | 10.0.0.7) (shards: 14)

OS / Image: `/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/SCYLLA-IMAGES/providers/Microsoft.Compute/images/scylla-2025.2.0-dev-x86_64-2025-04-05T01-47-23` (azure: undefined_region)

Test: `longevity-1tb-5days-azure-test`
Test id: `5d1aa1d0-c076-4aa7-a985-ee3400b1b6b5`
Test name: `scylla-master/tier1/longevity-1tb-5days-azure-test`
Test method: `longevity_test.LongevityTest.test_custom_time`
Test config file(s):

- [longevity-1TB-5days-authorization-and-tls-ssl.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/8859a386e88870d3a35aac63267cd1d93bcdccb7/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml)

<details>
[...]

PR #10825, opened 2025-05-06
fix(configuration): cluster nodes amount changed in configuration files

timtimb0t:
this commit changes cluster nodes amount from 3 to 6 to have more instances per rack becasue previous configuration caused topology operations decommission/removenode failures, because not enough nodes in rack
 to migrate tablets.
fixes: https://github.com/scylladb/scylla-cluster-tests/issues/10610

### Testing
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/eugene_test_folder/job/nodes_per_rack_configuration/

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
+++ b/test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
@@ -12,7 +12,7 @@ stress_cmd: ["cassandra-stress mixed cl=QUORUM duration=6800m -schema 'replicati
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "random", "interval": 120}']
 round_robin: true
 
-n_db_nodes: 3
+n_db_nodes: 6
 n_loaders: 4
 n_monitor_nodes: 1
 seeds_num: 2
--- a/test-cases/longevity/longevity-twcs-48h.yaml
+++ b/test-cases/longevity/longevity-twcs-48h.yaml
@@ -10,7 +10,7 @@ stress_read_cmd: [
     ]
 
 
-n_db_nodes: 3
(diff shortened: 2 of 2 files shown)


test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml
```diff
@@ -12,7 +12,7 @@ stress_cmd: ["cassandra-stress mixed cl=QUORUM duration=6800m -schema 'replicati
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "random", "interval": 120}']
 round_robin: true
 
-n_db_nodes: 3
+n_db_nodes: 6
```
soyacz, line 15: doubling cluster performance without adjusting the load changes test significantly - we should adjust load accordingly (unless it's proven that currently is too much)

fruch: In previous runs of TWCS test the CPU usage is ~80%, an in the run currently running (attached in description), the usage is ~50%

it's make the test much more easy on resources, as @soyacz point out, I would recommend do a run for this one, as well, so it can be compare and agree it's acceptable

soyacz: also we could reduce db node sizes from 4xlarge to 2xlarge

timtimb0t: Already run the test with increased load, will share the results

fruch: o.k. but where are the change ? i.e. they are not in this PR...

Issue #10716, opened 2025-04-28
Nodes logs never been collected on AWS backend run with message "file changed as we read it"


timtimb0t:
Nodes logs never been collected for this run:
https://argus.scylladb.com/tests/scylla-cluster-tests/f3c74eb0-a2dc-48a8-98ae-87e8693a1e1d
Jenkins URL:
https://jenkins.scylladb.com/job/scylla-master/job/tier1/job/longevity-mv-si-4days-streaming-test/34/

With the following messages:
```
 Found sct result directory with logs: /home/ubuntu/sct-results/20250426-030753-254815
07:01:22  Created directory to storing collected logs: /home/ubuntu/sct-results/20250426-030753-254815/collected_logs
07:01:22  Start collect logs for cluster db-cluster
07:04:50  Collecting logs on host: longevity-mv-si-4d-master-db-node-f3c74eb0-1
07:04:50  Collecting logs on host: longevity-mv-si-4d-master-db-node-f3c74eb0-2
07:04:50  Collecting logs on host: longevity-mv-si-4d-master-db-node-f3c74eb0-3
07:04:50  Collecting logs on host: longevity-mv-si-4d-master-db-node-f3c74eb0-4
07:04:50  Collecting logs on host: longevity-mv-si-4d-master-db-node-f3c74eb0-6
07:04:50  Collecting logs on host: longevity-mv-si-4d-master-db-node-f3c74eb0-5
07:12:20  <10.4.10.192>: Command rsync not available -- disabled
07:12:20  Error occured during collecting of db.crt on host: longevity-mv-si-4d-master-db-node-f3c74eb0-3
07:12:20  'NoneType' object is not iterable
07:12:20  Error occured during collecting of client-facing.crt on host: longevity-mv-si-4d-master-db-node-f3c74eb0-3
07:12:20  'NoneType' object is not iterable
07:12:20  Error occured during collecting of cqlshrc on host: longevity-mv-si-4d-master-db-node-f3c74eb0-3
07:12:20  'NoneType' object is not iterable
07:13:32  <10.4.8.12>: Command rsync not available -- disabled
07:13:40  Error occured during collecting of db.crt on host: longevity-mv-si-4d-master-db-node-f3c74eb0-6
07:13:40  'NoneType' object is not iterable
07:13:40  Error occured during collecting of client-facing.crt on host: longevity-mv-si-4d-master-db-node-f3c74eb0-6
07:13:40  'NoneType' object is not iterable
[...]

PR #10829, opened 2025-05-06
fix(logcollector): omit copying a non-exisintg remote log file

dimakr:
The CommandLog.collect method in sdcm/logcollector.py always attempts to copy a log file from a remote node, even when no log was prepared by the command. The change adds a check to ensure that the file is prepared on a remote host, before attempting to copy it.

Refs: https://github.com/scylladb/scylla-cluster-tests/issues/10716

### Testing

- [x] :yellow_circle: [pr-provision-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/52/consoleFull)
The test itself failed, but due to some other issue. The indication that this fix helps is that there is no such entries in the SCT log (also visible in jenkins build console output):
```
11:19:15  Error occured during collecting of db.crt on host: PR-provision-test-master-db-node-c19e5b97-1
11:19:15  'NoneType' object is not iterable
```
For instance the previous build [#51](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/51/consoleFull) of the same job (executed from SCT master), has these lines.

### Reminders

### DIFF
--- a/sdcm/logcollector.py
+++ b/sdcm/logcollector.py
@@ -219,6 +219,10 @@ class CommandLog(BaseLogEntity):  # pylint: disable=too-few-public-methods
         remote_logfile = LogCollector.collect_log_remotely(node=node,
                                                            cmd=self.cmd,
                                                            log_filename=os.path.join(remote_dst, self.name))
+        if not remote_logfile:
+            LOGGER.warning(
+                "Nothing to collect. Command '%s' did not prepare log file on remote host '%s'", self.cmd, node.name)
+            return None
         LogCollector.receive_log(node=node,
                                  remote_log_path=remote_logfile,
                                  local_dir=local_dst,


fruch (dismissed):
LGTM

sdcm/logcollector.py
```diff
@@ -219,6 +219,8 @@ def collect(self, node, local_dst, remote_dst=None, local_search_path=None) -> O
         remote_logfile = LogCollector.collect_log_remotely(node=node,
                                                            cmd=self.cmd,
                                                            log_filename=os.path.join(remote_dst, self.name))
+        if not remote_logfile:
```
soyacz, line 222: I'd prefer to have warn message before returning - so at least we are aware that some log was not collected because wasn't there.

dimakr: Warning might be misleading, if this was intended to leave the function if no items were prepared remotely in the `LogCollector.collect_log_remotely` call.

I mean - CommandLog takes arbitrary command, and the command itself might be as that we collect an item if it is there (on remote node), otherwise do nothing. This is the case for ssl artifacts related command - if they are on a node, collect them, otherwise skip.

soyacz: I understand we don't want to print anything for ssl files as some tests don't have it.
what if we misss some other file e.g. io_properties.yaml (just an example) due to some bug?

fruch: then we'll have that file missing, adding a warning won't help figuring the root cause of what ever issue happened

putting the logic into place which file is expected, and which isn't. is gonna complicate this process, and the ability to run it without the full configuration of the test

dimakr: I initially also wanted to include an info/warn message. But then refused thinking that it can mislead.

But if question is raised from audience, then it probably can be helpful. In the new patchset I added a message, which indicate which command was used, which didn't produce anything to collect

dimakr: It would look like:
```
11:38:50  Nothing to collect. Command 'test -f /etc/scylla/ssl_conf/db.crt && cat /etc/scylla/ssl_conf/db.crt' did not prepare log file on remote host 'PR-provision-test-fix-ssl--db-node-131c7157-1'
```
@soyacz

soyacz: > then we'll have that file missing, adding a warning won't help figuring the root cause of what ever issue happened
> 

Sometimes people look into jenkins logs and may find out of something weird/uncommon before given log was really needed. At least showing we tried to collect it.

Solution proposed by @dimakr LGTM

soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

PR #10830, opened 2025-05-06
fix(hydra.sh): correctly parse SCT_* env var values containing `=`

dimakr:
SCT_* env variables were not parsed correctly for the `docker run --env` option, when value of the variable contained `=` characters. The issue was caused by the "greedy" regex that consumed all `=` characters up to the last one, misinterpreting the actual key-value separator.

This change adjusts the regex to correctly parse the values that include `=` characters.

### Testing

  - [x] :yellow_circle: [pr-provision-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/55/)
The test failed due to some another existing issue, but parsing and setting custom value for the SCT_PREPARE_WRITE_CMD env var is working as expected now.

Previously setting custom s-b stress commands (or any env. vars with `=` characters in the value) would result in the error:
```
12:00:17  [Pipeline] { (Create Argus Test Run)
...
12:00:19  Going to run './sct.py  create-argus-test-run'...
12:00:19  invalid argument "ode=write" for "-m, --memory" flag: invalid size: 'ode=write'
```
like here: [failed pr-provision-test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/49/console)

### Reminders

### DIFF
--- a/docker/env/hydra.sh
+++ b/docker/env/hydra.sh
@@ -185,19 +185,19 @@ fi
 DOCKER_ADD_HOST_ARGS=()
 
 # export all SCT_* env vars into the docker run
-SCT_OPTIONS=$(env | sed -n 's/^\(SCT_.*\)=.*/--env \1/p')
+SCT_OPTIONS=$(env | sed -n 's/^\(SCT_[^=]*\)=.*/--env \1/p')
 
 # export all PYTEST_* env vars into the docker run
-PYTEST_OPTIONS=$(env | sed -n 's/^\(PYTEST_.*\)=.*/--env \1/p')
+PYTEST_OPTIONS=$(env | sed -n 's/^\(PYTEST_[^=]*\)=.*/--env \1/p')
 
 # export all BUILD_* env vars into the docker run
-BUILD_OPTIONS=$(env | sed -n 's/^\(BUILD_.*\)=.*/--env \1/p')
+BUILD_OPTIONS=$(env | sed -n 's/^\(BUILD_[^=]*\)=.*/--env \1/p')
 
 # export all AWS_* env vars into the docker run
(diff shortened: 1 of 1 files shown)


fruch:
@roydahan care to check it, to see it doesn't break macos ?

(I'll check why the action for it didn't triggered)

roydahan:
> @roydahan care to check it, to see it doesn't break macos ?
> 
> (I'll check why the action for it didn't triggered)

To try with the SCT_ param that has "=" in it?
Or simply to run the test provision?

fruch:
> > @roydahan care to check it, to see it doesn't break macos ?
> > 
> > (I'll check why the action for it didn't triggered)
> 
> To try with the SCT_ param that has "=" in it?
> Or simply to run the test provision?

Run any command and uses SCT_* environment variables, with and without = in them

fruch:
a fix to the action https://github.com/scylladb/scylla-cluster-tests/pull/10832, so it would trigger on this PR.

we can merge it, and then rebase this one

fruch:
rebese this one, let's see if Mac loves this change

docker/env/hydra.sh
```diff
@@ -185,7 +185,7 @@ fi
 DOCKER_ADD_HOST_ARGS=()
 
 # export all SCT_* env vars into the docker run
-SCT_OPTIONS=$(env | sed -n 's/^\(SCT_.*\)=.*/--env \1/p')
+SCT_OPTIONS=$(env | sed -n 's/^\(SCT_[^=]*\)=.*/--env \1/p')
```
fruch, line 188: maybe we should consider doing so in the other options as well ?

dimakr: I was about to push this change, but you was first)

fruch (approved):

PR #10842, opened 2025-05-09
Introduce basic Manager ops Nemesis test

mikliapko:
The PR introduces the basic Manager Nemesis test.

It runs only manager specific operation (backup, restore, etc) against 100GB dataset during 4 hours.

The PR also contains a tiny update of Manager versions in MANAGER_CONFIG parameter description of longevityPipeline.groovy.

### Testing

- [x] [longevity-100gb-manager-ops-4h](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/longevity-100gb-manager-ops-4h/16/)

### DIFF
--- a/vars/longevityPipeline.groovy
+++ b/vars/longevityPipeline.groovy
@@ -100,7 +100,7 @@ def call(Map pipelineParams) {
             // Manager Configuration
             separator(name: 'MANAGER_CONFIG', sectionHeader: 'Manager Configuration')
             string(defaultValue: "${pipelineParams.get('manager_version', '')}",
-                   description: 'master_latest|3.2|3.1',
+                   description: 'master_latest|3.7|3.6',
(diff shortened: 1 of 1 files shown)


mikliapko:
> I'm not sure what exactly is being tested here. This test runs backup/restore/repair for 4h, but does it also has some background load and validation going on? Is restore performed on some pre-made backup, or is the backup re-executed as well?

You can find some details in [Argus](https://argus.scylladb.com/tests/scylla-cluster-tests/f6af2646-e1d7-4ce1-9847-a1491e4b1c31/nemesis).

The test runs write load during the whole test duration. 
And different nemesis tests (such as backup, restore, repair, etc (see the full list in Argus)) take place at the same time.
Nemesis disrupt event is node termination. So, all these checks are run with one of the nodes in terminated state.

PR #10852, opened 2025-05-12
feature(adaptive_timeouts): report results to Argus

fruch:
since we are ditching EleasticSearch, this commit introduces a way to report the duration and timeouts used in operations

this only implements the option to send out the data, we don't yet have a client API to retrive it from Argus.

### Testing

- [x] 🟡 https://argus.scylladb.com/tests/scylla-cluster-tests/68f900fd-a653-4faa-90b0-bf32b15b1fbd/results
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/122/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/123/
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/124/
- [x] 🟢 https://argus.scylladb.com/tests/scylla-cluster-tests/05c86a0e-1789-46c8-85aa-7dd9f36b2afc/results
 - [x] 🟢  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/131/
 
### TODOs
- [x] - drop the ESStore
- [x] - decide about metrics that we want to save
- [x] - hide the extra metrics when https://github.com/scylladb/argus/issues/690 is ready
### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -312,3 +312,5 @@ mgmt_snapshots_preparer_params:
   num_of_rows: ''
   sequence_start: ''
   sequence_end: ''
+
+adaptive_timeout_store_metrics: true
--- a/sdcm/utils/adaptive_timeouts/__init__.py
+++ b/sdcm/utils/adaptive_timeouts/__init__.py
@@ -170,9 +170,9 @@ def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: PLR0914, 
     args = {arg: kwargs[arg] for arg in required_arg_names}
     store_metrics = node.parent_cluster.params.get("adaptive_timeout_store_metrics")
     if store_metrics:
-        metrics = {}
-    else:
         metrics = NodeLoadInfoServices().get(node)
+    else:
+        metrics = {}
     if tablet_sensitive_op:
         args['tablets_enabled'] = tablets_enabled
     result = timeout_func(node_info_service=metrics, **args)
--- a/sdcm/utils/adaptive_timeouts/__init__.py
+++ b/sdcm/utils/adaptive_timeouts/__init__.py
@@ -168,16 +168,21 @@ def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: PLR0914, 
 
     _, timeout_func, required_arg_names = operation.value
     args = {arg: kwargs[arg] for arg in required_arg_names}
-
+    store_metrics = node.parent_cluster.params.get("adaptive_timeout_store_metrics")
+    if store_metrics:
+        metrics = {}
+    else:
+        metrics = NodeLoadInfoServices().get(node)
     if tablet_sensitive_op:
         args['tablets_enabled'] = tablets_enabled
-    result = timeout_func(node_info_service=NodeLoadInfoServices().get(node), **args)
+    result = timeout_func(node_info_service=metrics, **args)
     if tablet_sensitive_op:
         (soft_timeout, hard_timeout), load_metrics = result
     else:
         soft_timeout, load_metrics = result
         hard_timeout = None
-    load_metrics.update(TestInfoServices.get(node))
+    if store_metrics:
+        load_metrics.update(TestInfoServices.get(node))
 
     start_time = time.monotonic()
     timeout_occurred = False
@@ -202,7 +207,7 @@ def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: PLR0914, 
             SoftTimeoutEvent(operation=operation.name, soft_timeout=soft_timeout, duration=duration).publish_or_dump()
 
         try:
-            if load_metrics:
+            if load_metrics and store_metrics:
                 stats_storage.store(metrics=load_metrics, operation=operation.name, duration=duration,
                                     timeout=soft_timeout, timeout_occurred=timeout_occurred)
         except Exception as exc:  # noqa: BLE001
--- a/sdcm/utils/adaptive_timeouts/__init__.py
+++ b/sdcm/utils/adaptive_timeouts/__init__.py
@@ -6,7 +6,7 @@ from enum import Enum
 from typing import Any
 
 from sdcm.sct_events.system import SoftTimeoutEvent, HardTimeoutEvent
-from sdcm.utils.adaptive_timeouts.load_info_store import NodeLoadInfoService, AdaptiveTimeoutStore, ESAdaptiveTimeoutStore, \
+from sdcm.utils.adaptive_timeouts.load_info_store import NodeLoadInfoService, AdaptiveTimeoutStore, ArgusAdaptiveTimeoutStore, \
     NodeLoadInfoServices
 from sdcm.utils.features import is_tablets_feature_enabled
 
@@ -156,7 +156,7 @@ class TimeoutMonitor:
 
 @contextmanager
 def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: PLR0914, F821
-                     stats_storage: AdaptiveTimeoutStore = ESAdaptiveTimeoutStore(), **kwargs):
+                     stats_storage: AdaptiveTimeoutStore = ArgusAdaptiveTimeoutStore(), **kwargs):
     """
     Calculate timeout in seconds for given operation based on node load info and return its value.
     Upon exit, verify if timeout occurred and publish SoftTimeoutEvent if happened.
@@ -206,4 +206,4 @@ def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: PLR0914, 
                 stats_storage.store(metrics=load_metrics, operation=operation.name, duration=duration,
                                     timeout=soft_timeout, timeout_occurred=timeout_occurred)
         except Exception as exc:  # noqa: BLE001
-            LOGGER.warning("Failed to store adaptive timeout stats: \n%s", exc)
+            LOGGER.warning("Failed to store adaptive timeout stats: \n%s", exc, exc_info=True)
--- a/sdcm/utils/adaptive_timeouts/load_info_store.py
+++ b/sdcm/utils/adaptive_timeouts/load_info_store.py
@@ -10,23 +10,33 @@
 # See LICENSE for more details.
 #
 # Copyright (c) 2023 ScyllaDB
+from __future__ import annotations
+
 import logging
 import time
 import uuid
 from collections import defaultdict
-from datetime import datetime
+from datetime import datetime, timezone
 from functools import cached_property
 import re
-from typing import Any
+from typing import Any, TYPE_CHECKING
+from itertools import count
+from dataclasses import dataclass
 
 import yaml
 from cachetools import cached, TTLCache
+from argus.client.generic_result import StaticGenericResultTable, ColumnMetadata, ResultType, Status
 
 from sdcm.es import ES
+
+if TYPE_CHECKING:
+    from sdcm.cluster import BaseNode
+
 from sdcm.remote import RemoteCmdRunner
 from sdcm.test_config import TestConfig
 from sdcm.utils.decorators import retrying
 from sdcm.utils.metaclasses import Singleton
+from sdcm.argus_results import submit_results_to_argus
 
 LOGGER = logging.getLogger(__name__)
 
@@ -56,10 +66,11 @@ class NodeLoadInfoService:
     nodetool status, uptime (load), vmstat (disk utilization). Command responses are cached for some time to avoid too much requests.
     """
 
-    def __init__(self, remoter: RemoteCmdRunner, name: str, scylla_version: str):
+    def __init__(self, remoter: RemoteCmdRunner, name: str, scylla_version: str, node_idx: str):
         self.remoter = remoter
         self._name = name
         self._scylla_version = scylla_version
+        self.node_idx = node_idx
 
     @cached_property
     def _io_properties(self):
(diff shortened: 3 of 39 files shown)


sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -156,7 +156,7 @@ def _monitor_timeouts(self):
 
 @contextmanager
 def adaptive_timeout(operation: Operations, node: "BaseNode",  # noqa: PLR0914, F821
-                     stats_storage: AdaptiveTimeoutStore = ESAdaptiveTimeoutStore(), **kwargs):
+                     stats_storage: AdaptiveTimeoutStore = ArgusAdaptiveTimeoutStore(), **kwargs):
```
pehala, line 159: is the mutable default here intentionally and if so, shouldnt we make it something like a class variable? I do not like relying on mutable defaults in any case

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -251,6 +253,54 @@ def get(self, operation: str | None, timeout_occurred: bool | None = None):
...
+    def __init__(self):
+        self.test_config = TestConfig()
+
+    def send_adaptive_timeout_results_to_argus(self, results: dict):
+        argus_client = self.test_config.argus_client()
+        if not argus_client:
+            LOGGER.warning("Will not submit to argus - no client initialized")
+            return
+
+        class AdaptiveTimeoutResultsTable(GenericResultTable):
```
pehala, line 270: Could we make this a normal class instead so we could reuse it? I dont think we should be using inner classes.

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -251,6 +253,54 @@ def get(self, operation: str | None, timeout_occurred: bool | None = None):
...
 
+class ArgusAdaptiveTimeoutStore(AdaptiveTimeoutStore):
+    """
+    Report adaptive timeout results to Argus.
+    """
+
+    def __init__(self):
+        self.test_config = TestConfig()
+
+    def send_adaptive_timeout_results_to_argus(self, results: dict):
```
pehala, line 264: How about a dataclass for results, so it in clear what you actually need in the dict?

fruch: that's the current API, I don't want to mess around with it in this PR

sdcm/utils/adaptive_timeouts/__init__.py
fruch: it's not really mutable in that sense.
it's just one more way of creating a singleton

the downside of it that it's get created on loading the module (it bite us once)

Anyhow, not gonna change anything drastically in this PR

sdcm/utils/adaptive_timeouts/load_info_store.py
fruch: we are not gonna reuse it, and if we do, we can move it.
we are doing this in multiple places, since the dynamic nature those tables

sdcm/utils/adaptive_timeouts/load_info_store.py
pehala: Is this method used outside of `store`?

sdcm/utils/adaptive_timeouts/load_info_store.py
pehala: We want dynamic tables but we are using static classes to implement that, is there a reason why we implement it like this in the first place?

This can be addressed in the followup PR, but but I would like to understand the design here

sdcm/utils/adaptive_timeouts/load_info_store.py
fruch: no

sdcm/utils/adaptive_timeouts/load_info_store.py
fruch: it's following the same pattern of usage of `GenericResultTable`, those tables are not share across multiple places, hence it's doesn't matter if they are defined on their own, or inline.

if for whatever reason it would be needed on it's own, it can be moved.

sdcm/utils/adaptive_timeouts/load_info_store.py
pehala: In that case, we can change it and not break the API right?

sdcm/utils/adaptive_timeouts/load_info_store.py
pehala: `GenericResultTable.as_dict()` is only caring about values in `self`, so we can easily make this class dynamic (without changing anything else) by using init function and not having it done through Meta

pehala: I tested it locally and it cannot be done easily so closing this comments. There is actual logic tied to the Meta class, which in opinion is wrong and prevents better usage patterns.

fruch:
we are waiting for https://github.com/scylladb/argus/issues/662

fruch:
image

@soyacz duration should be int ? right now it's passing down float, and the output is a bit too lengthy about sub second information

soyacz:
> @soyacz duration should be int ? right now it's passing down float, and the output is a bit too lengthy about sub second information

Yes, should be int. We may also fix it on argus side.

fruch:
> we are waiting for [scylladb/argus#662](https://github.com/scylladb/argus/issues/662)

meanwhile did a workaround it in the code itself, to have counters per operation

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -249,6 +252,60 @@ def get(self, operation: str | None, timeout_occurred: bool | None = None):
...
+                name = f"{results['operation']} - Timeout Statistics"
+                description = "measurement of specific operation timeouts (ex. decommission, adding nodes etc.)"
+                Columns = [
+                    ColumnMetadata(name="duration", unit="HH:MM:SS", type=ResultType.DURATION, higher_is_better=False),
+                    ColumnMetadata(name="timeout", unit="HH:MM:SS", type=ResultType.DURATION),
+                    ColumnMetadata(name="end_time", unit="", type=ResultType.TEXT),
+                ]
+        cycle = next(self.cycle_counters[results['operation']]) + 1
+        table = AdaptiveTimeoutResultsTable()
+        table.add_result(column="duration", row=f"#{cycle}", value=results['duration'], status=Status.UNSET)
```
soyacz, line 281: I think this one should have status error when we cross timeout, otherwise should be PASS. 'timeout' column should be always UNSET.

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -249,6 +252,60 @@ def get(self, operation: str | None, timeout_occurred: bool | None = None):
...
+        table.add_result(column="timeout", row=f"#{cycle}", value=results["timeout"], status=Status.ERROR if results.get(
+            "timeout_occurred") else Status.UNSET)
+        table.add_result(column="end_time", row=f"#{cycle}", value=results["end_time"], status=Status.UNSET)
+        logging.debug("Submitting adaptive timeout results to Argus: %s", table.as_dict())
+        submit_results_to_argus(argus_client, table)
```
soyacz, line 286: Just mentioning for awareness: I was thinking we should catch here failures and rollback operation's cycle counter, but actually, for comparisons across runs it will be better to just miss cycle (so graph draws more correct lines). Maybe worth to add comment somewhere around cycles counter

soyacz: btw. I created issue around one possible error: https://github.com/scylladb/argus/issues/680

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -249,6 +252,60 @@ def get(self, operation: str | None, timeout_occurred: bool | None = None):
...
+        results["duration"] = int(duration)
+        results["timeout"] = timeout
+        results["timeout_occurred"] = timeout_occurred
+
+        self.send_adaptive_timeout_results_to_argus(results)
+
+    def get(self, operation: str | None, timeout_occurred: bool | None = None):
+
+        # TODO: we don't have yet API to get measurements from Argus
+        # TODO: also adaptive_timeout decorator isn't reading measurements yet...
```
soyacz, line 305: In future we might consider using `validation rules` for this - but we need more data to see if it would be valid approach.

soyacz (commented):
generally, looks good - just few small remarks

sdcm/utils/adaptive_timeouts/load_info_store.py
fruch: Graphs from those might not be really helpful, since nemesis can be shuffled from time to time.

We'll need overview dashboard for those

sdcm/utils/adaptive_timeouts/load_info_store.py
fruch: We did do that since we introduced the adaptive timeouts

And since we have hard timeouts on those, it doesn't mean much.

sdcm/utils/adaptive_timeouts/load_info_store.py
vponomaryov: Agree, red/errored duration cell is intuitively clear that it is bigger than expected.

Red/errored timeout cell is less clear.

sdcm/utils/adaptive_timeouts/load_info_store.py
fruch: I've added a dataclass, more or less exactly the same, since metrics are still define in other class (which isn't dataclass)

sdcm/utils/adaptive_timeouts/load_info_store.py
```diff
@@ -249,6 +253,77 @@ def get(self, operation: str | None, timeout_occurred: bool | None = None):
...
+
+    def store(self, metrics: dict[str, Any], operation: str, duration: float, timeout: float,
+              timeout_occurred: bool):
+        result = ArgusAdaptiveTimeoutResult(
+            operation=operation,
+            duration=int(duration),
+            timeout=int(timeout),
+            timeout_occurred=timeout_occurred,
+            end_time="N/A",
+            metrics=metrics.copy()
```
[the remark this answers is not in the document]
soyacz: My proposal is to send them.
I think we can observe for some time if these bring value - having results and graphs.
If we decide they're not useful in UI or anywhere - we ditch it or we'll introduce a flag to hide them in Argus UI (only available by query)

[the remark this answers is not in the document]
fruch: I think we need this flag, it's a long list of metrics

I'll give it a shot, and we'll how it looks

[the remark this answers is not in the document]
fruch: it's a bit too much for the design of results tab
https://argus.scylladb.com/tests/scylla-cluster-tests/a99bdc3d-dfa8-408a-8966-5e65df41909d/results

image

and it doesn't fit completely (even with empty data)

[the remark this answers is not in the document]
soyacz: ugh, doesn't look good... My current proposal is to skip it (even disable collection).
Let's first see if durations are useful at all.
Later we could bring it back (when we can disable showing certain columns), in the meantime we still have metrics where we could get data from (in case we see degradation in durations)

fruch:
I've added new way to disable adaptive_timeout reporting for performance tests
and remove the timeout for START_SCYLLA

argus/client/generic_result.py
```diff
@@ -85,21 +89,39 @@ class Cell:
...
-            "row": self.row,
-            "status": str(self.status)
-        })
+        cell.update({"column": self.column, "row": self.row, "status": str(self.status)})
         return cell
 
 
 @dataclass
-class GenericResultTable(metaclass=ResultTableMeta):
+class GenericResultTable:
```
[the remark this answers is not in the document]
pehala: Addressed in https://github.com/scylladb/scylla-cluster-tests/pull/11151

defaults/test_default.yaml
```diff
@@ -312,3 +312,5 @@ mgmt_snapshots_preparer_params:
   num_of_rows: ''
   sequence_start: ''
   sequence_end: ''
+
+adaptive_timeout_store_metrics: true
```
soyacz, line 316: let's make it `false` for docker backend (assuming we drop node_exporter on it after merging https://github.com/scylladb/scylla-cluster-tests/pull/10627)

soyacz:
> @soyacz why `visible=False` doesn't work ? https://argus.scylladb.com/tests/scylla-cluster-tests/21ff5868-4a41-4e1d-81a9-0f3e44ef7b97/results

because we didn't sync db model on prod to include new field... :cry:

fruch:
> > @soyacz why `visible=False` doesn't work ? https://argus.scylladb.com/tests/scylla-cluster-tests/21ff5868-4a41-4e1d-81a9-0f3e44ef7b97/results
> 
> because we didn't sync db model on prod to include new field... 😢

@soyacz can you raise issue for Arugs to automate as much as possible the step for deployment, so such things won't be forgotten ?

fruch:
@scylladb/qa-maintainers this now proven as working, please give it one last review

soyacz:
> > > @soyacz why `visible=False` doesn't work ? https://argus.scylladb.com/tests/scylla-cluster-tests/21ff5868-4a41-4e1d-81a9-0f3e44ef7b97/results
> > 
> > 
> > because we didn't sync db model on prod to include new field... 😢
> 
> @soyacz can you raise issue for Arugs to automate as much as possible the step for deployment, so such things won't be forgotten ?

https://github.com/scylladb/argus/issues/712

PR #10890, opened 2025-05-14
Fixes for Scylla Operator tests

grzywin:
List of changes:

- The `is_tablets_feature_enabled` method was throwing an error during Operator tests because it attempted to run the `dict()` method on an object that was already a dictionary.

- Bumped versions of Scylla, VPC CNI, Kubernetes (k8s), and Manager in the YAML files. The Kubernetes version bump is especially important, as the previously used versions are no longer supported on EKS and GKE. Note: This version bump is not aligned with the latest Scylla releases, but these versions have been tested and we know that everything works correctly. In the next step, we can proceed with further updates.

- Added the `force-conflicts` parameter to the `apply_file` method. This is required for the Scylla Operator upgrade test, which was intermittently causing issues.

- Due to recent changes in SCT and/or the EKS `ebs-csi-driver`, creation of the addon started failing because it was being created before the node pools were up and running. I moved the addon creation step to occur after the `wait_all_node_pools_to_be_ready` method is triggered.

- Longevity tests for the Operator were often failing due to this issue: https://github.com/scylladb/java-driver/issues/213#issuecomment-2719704747. The Operator team advised adding the `expose-pod-ips` configuration to resolve the problem — and indeed, it worked. I added this option to all longevity pipelines currently run for the Operator.

- While investigating functional test failures for the Operator, two tests were failing randomly when run alongside others, but passed when triggered individually. Since tests should be independent, I configured them to run first in the sequence. Thanks to this change, they now consistently pass.

- After this commit: https://github.com/scylladb/scylla-cluster-tests/commit/d49868860d1ccedade2980f6efbbb9ea48dce688, all Operator tests started failing because an extra dynamic loader was incorrectly initialized. The tests were waiting for it to be up and running — which never happened. I resolved the issue by removing `args[0].loaders` from the `nodes_set.log_message` method call.

- Increased the timeout for resharding and set the `nemesis_add_node_cnt` value for Scylla Operator longevity tests, as the global default values were removed or changed in a recent SCT commit.

- Updated Prometheus Operator monitoring manifests to be compatible with the Scylla Operator repository. Without this update, Scylla monitoring functional tests would fail.

- Increased storage capacity for the "Helm with default values" Operator test case. Without this, Scylla fails to start with the latest versions of Scylla and/or Scylla Operator due to insufficient storage. Also, mandatory labels were added to DB nodes, as they are now required in the Scylla Operator manifests.

### TESTING
- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/3d783a44-64dd-40cc-9027-ff46c8716ab1
- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/e367f28c-d494-4e63-9e6c-7ab40ad61fc7
- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/a5c584bb-30a6-4210-ae49-7afb000939d7 (disrupt_truncate_large_partition nemesis failed due to this bug https://github.com/scylladb/scylla-bench/issues/184)
- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/f1cc365b-da2e-492e-a918-95958d5f1406  (disrupt_nodetool_refresh nemesis need further investigation)

### DIFF
--- a/sdcm/cluster_k8s/eks.py
+++ b/sdcm/cluster_k8s/eks.py
@@ -118,6 +118,9 @@ def deploy_k8s_eks_cluster(k8s_cluster) -> None:
         instance_type=params.get('instance_type_db'),
         role_arn=params.get('eks_nodegroup_role_arn'),
         disk_size=params.get('root_disk_size_db'),
+        labels={
+            "scylla.scylladb.com/node-type": "scylla"
+        },
         k8s_cluster=k8s_cluster)
     k8s_cluster.deploy_node_pool(scylla_pool, wait_till_ready=False)
 
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -767,7 +767,7 @@ def test_deploy_helm_with_default_values(db_cluster: ScyllaPodCluster):
     """
 
     target_chart_name, namespace = ("t-default-values",) * 2
-    expected_capacity = '10Gi'
+    expected_capacity = '120Gi'
     need_to_collect_logs, k8s_cluster = True, db_cluster.k8s_cluster
     logdir = f"{os.path.join(k8s_cluster.logdir, 'test_deploy_helm_with_default_values')}"
 
@@ -792,6 +792,10 @@ def test_deploy_helm_with_default_values(db_cluster: ScyllaPodCluster):
 
         pods_name_and_status = get_pods_and_statuses(
             db_cluster, namespace=namespace, label=db_cluster.pod_selector)
+
+        # Sometimes the get_pods_and_statuses method incorrectly includes cleanup pods.
+        # The line below filters out these pods from our pod list.
+        pods_name_and_status = [pod for pod in pods_name_and_status if 'cleanup' not in pod['name'].lower()]
 
         assert len(pods_name_and_status) == 3, (
             f"Expected 3 pods to be created in {namespace} namespace "
--- a/sdcm/cluster_k8s/gke.py
+++ b/sdcm/cluster_k8s/gke.py
@@ -130,6 +130,9 @@ def deploy_k8s_gke_cluster(k8s_cluster) -> None:
         instance_type=params.get("gce_instance_type_db"),
         num_nodes=params.get("n_db_nodes"),
         taints=["role=scylla-clusters:NoSchedule"],
+        labels={
+            "scylla.scylladb.com/node-type": "scylla"
+        },
         k8s_cluster=k8s_cluster)
     k8s_cluster.deploy_node_pool(scylla_pool, wait_till_ready=False)
 


sdcm/utils/features.py
```diff
@@ -80,12 +80,15 @@ def is_tablets_feature_enabled(node) -> bool:
     """
     with node.remote_scylla_yaml() as scylla_yaml:
         # for backward compatibility of 2024.1 and earlier
-        scylla_conf = scylla_yaml.dict()
-        if "tablets" in (scylla_conf.get("experimental_features") or []):
+        if isinstance(scylla_yaml, dict):
```
vponomaryov, line 83: The change is ok.
The commit message for it is missleading - `fix(operator): Fix tablets feature for Operator tests`

Need to write something like following:

```
fix(utils/features.py): support 'scylla_yaml' as dict for K8S backends
```

jenkins-pipelines/operator/upgrade/upgrade-major-scylla-podips-k8s-gke.jenkinsfile
```diff
@@ -6,10 +6,8 @@ rollingOperatorUpgradePipeline(
     backend: 'k8s-gke',
     region: 'us-east1',
     availability_zone: 'c',
-    // NOTE: the lowest Scylla version which can be used for exposing Pod IPs is '2023.1.x'
-    base_versions: '["2023.1.2"]',
-    // TODO: set the new version as '2024.1.x' when gets released
-    new_version: '2024.1.0-rc2',
+    base_versions: '["2023.1.11"]',
```
vponomaryov, line 9: The `2023.1` is not supported anymore.
Use newer version.

test-cases/scylla-operator/functional.yaml
```diff
@@ -10,7 +10,7 @@ k8s_loader_cluster_name: 'sct-loaders'
 
 n_monitor_nodes: 0
 
-scylla_version: '4.6.3'
+scylla_version: '6.2.3'
```
vponomaryov, line 13: Need to use newer version - `2025.1`

jenkins-pipelines/operator/upgrade/upgrade-major-scylla-podips-k8s-eks.jenkinsfile
```diff
@@ -6,10 +6,8 @@ rollingOperatorUpgradePipeline(
     backend: 'k8s-eks',
     region: 'eu-north-1',
     availability_zone: 'c',
-    // NOTE: the lowest Scylla version which can be used for exposing Pod IPs is '2023.1.x'
-    base_versions: '["2023.1.2"]',
-    // TODO: set the new version as '2024.1.x' when gets released
-    new_version: '2024.1.0-rc2',
+    base_versions: '["2023.1.11"]',
```
vponomaryov, line 9: Same here. Too old version.

jenkins-pipelines/operator/upgrade/upgrade-major-scylla-k8s-gke.jenkinsfile
```diff
@@ -6,8 +6,8 @@ rollingOperatorUpgradePipeline(
     backend: 'k8s-gke',
     region: 'us-east1',
     availability_zone: 'c',
-    base_versions: '["2022.1.6"]',
-    new_version: '2022.2.7',
+    base_versions: '["2023.1.11"]',
```
vponomaryov, line 9: Same here. Too old version.

jenkins-pipelines/operator/upgrade/upgrade-major-scylla-k8s-eks.jenkinsfile
```diff
@@ -5,8 +5,8 @@ def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 rollingOperatorUpgradePipeline(
     backend: 'k8s-eks',
     region: 'eu-north-1',
-    base_versions: '["2022.1.6"]',
-    new_version: '2022.2.7',
+    base_versions: '["2023.1.11"]',
```
vponomaryov, line 8: Same here. Too old version.

defaults/k8s_eks_config.yaml
```diff
@@ -9,14 +9,14 @@ availability_zone: 'a,b'
...
+eks_cluster_version: '1.32'
 eks_role_arn: 'arn:aws:iam::797456418907:role/eksServicePolicy'
 eks_service_ipv4_cidr: '172.20.0.0/16'
 eks_nodegroup_role_arn: 'arn:aws:iam::797456418907:role/helm-test-worker-nodes-NodeInstanceRole-6ACHDYEKNN3I'
 # NOTE: https://docs.aws.amazon.com/eks/latest/userguide/managing-vpc-cni.html
-eks_vpc_cni_version: 'v1.15.3-eksbuild.1'
+eks_vpc_cni_version: 'v1.19.2-eksbuild.5'
 
-scylla_version: '5.2.11'
+scylla_version: '6.2.3'
```
vponomaryov, line 19: Need to use newer version - `2025.1`

defaults/k8s_gke_config.yaml
```diff
@@ -18,7 +18,7 @@ k8s_n_auxiliary_nodes: 2
 
 k8s_instance_type_auxiliary: 'n2-standard-2'
 
-scylla_version: '5.2.11'
+scylla_version: '6.2.3'
```
vponomaryov, line 21: Need to use newer version - `2025.1`

defaults/k8s_local_kind_config.yaml
```diff
@@ -4,7 +4,7 @@ mini_k8s_version: '0.20.0'
 n_db_nodes: 4
 k8s_n_scylla_pods_per_cluster: 3
 
-scylla_version: '5.2.11'
+scylla_version: '6.2.3'
```
vponomaryov, line 7: Need to use newer version - `2025.1`

sdcm/utils/k8s/__init__.py
```diff
@@ -368,7 +370,7 @@ def apply_file(cls, kluster, config_path, namespace=None,  # pylint: disable=too
                 @retrying(n=0, sleep_time=5, timeout=timeout, allowed_exceptions=RuntimeError)
                 def run_kubectl(file_name):
                     try:
-                        cls.kubectl(kluster, "apply", server_side_arg, "-f", file_name,
+                        cls.kubectl(kluster, "apply", server_side_arg, force_conflicts_arg, "-f", file_name,
```
vponomaryov, line 373: Need to add to commit description some details/info about it.

sdcm/cluster_k8s/eks.py
```diff
@@ -423,19 +429,16 @@ def create_eks_cluster(self, wait_till_functional=True):
             },
             tags=self.tags,
         )
+
```
vponomaryov, line 435: Need to add some info about this change to the commit description.

Why this update is needed?

jenkins-pipelines/operator/eks/longevity-scylla-operator-3h-eks-backup.jenkinsfile
```diff
@@ -6,7 +6,7 @@ longevityPipeline(
     backend: 'k8s-eks',
     region: 'eu-north-1',
     test_name: 'longevity_test.LongevityTest.test_custom_time',
-    test_config: 'test-cases/scylla-operator/longevity-scylla-operator-3h-backup.yaml',
+    test_config: '''["test-cases/scylla-operator/longevity-scylla-operator-3h-backup.yaml", "configurations/operator/expose-pod-ips.yaml"]''',
```
vponomaryov, line 9: Why is it needed?

Bug reference?

Need to add details to the commit message.

functional_tests/scylla_operator/conftest.py
```diff
@@ -237,3 +237,22 @@ def skip_if_no_tls(request, tester: ScyllaOperatorFunctionalClusterTester) -> No
             and not tester.params.get("k8s_enable_tls")
             and not tester.params.get("k8s_enable_sni")):
         pytest.skip('test requires k8s_enable_tls to be enabled')
+
+
+def pytest_collection_modifyitems(items):
```
vponomaryov, line 242: Docs say it has 3 parameters: https://docs.pytest.org/en/latest/reference/reference.html#collection-hooks

Then, it is unclear what failures are refered to by the commit topic - `fix(test_order): Force test order to reduce failures`

Without this change the test order is also static.
So, it is not "forcing", it is "changing" the order of tests.
Why?

vponomaryov (changes requested):
See inline comments.

Common concerns for all of the commits:
- Lack of info in commit descriptions. 
  - Reasons, bug refs, other context are absent
- Commit topics are too long. Need to satisfy precommit rules.
- All of it should not be single huge PR.

sdcm/utils/features.py
pehala: Why does K8s backend have a different type for the same `scylla_yaml`? Shouldnt be unify the types instead?

functional_tests/scylla_operator/conftest.py
pehala: Also if your tests are dependant on the test order, you have a fundamental flaw in your tests :)

sdcm/utils/features.py
vponomaryov: @pehala 
Because scylla operator on k8s handles part of the Scylla yaml exclusively and allows to define only part of options.
The ScyllaYaml python class tries to handle all of the options.

One can create a similar ScyllaYamlForOperator data class, but it will have different set of options anyway.
And there should be some reason for it.

Some compatibility class to mock ScyllaYaml interfaces may be a solution too...

Summary:
No one cares about K8S specifics developing stuff having in mind VM-based test scenarios.

pehala: > @pehala Because scylla operator on k8s handles part of the Scylla yaml exclusively and allows to define only part of options. The ScyllaYaml python class tries to handle all of the options.
> 
> One can create a similar ScyllaYamlForOperator data class, but it will have different set of options anyway. And there should be some reason for it.
> 
> Some compatibility class to mock ScyllaYaml interfaces may be a solution too...
> 
> Summary: No one cares about K8S specifics developing stuff having in mind VM-based test scenarios.

We can simply wrap it in ScyllaYaml compliant object, i.e. add a dict() method to it which will return self (and extend dict probably) or we can make ScyllaYaml dict-like both work. I also think we should care about that, or better make the code compliant enough so that we dont have to care about that.

mikliapko:
Since Kamil left the company, I'm taking over this PR

PR #10912, opened 2025-05-16
improvement(nemesis_tests): Make tests independant from production Nemesis

### LINKED ISSUE #10592: Make test_nemesis independent from Nemesis disruption methods
Currently, the `test_nemesis.py` searches through all disruptions in Nemesis class, meaning even production ones.
This makes it hard to understand tests and does mean that the actual tests are dependant on the production disruption methods.

We should rethink how we test nemesis execution/discovery process and make it independent so have higher confidence in our tests and make them simpler.

One of the solution to this problerm would be to rewrite tests after #10502, which enables us to separate or even search methods in different class tree (i.e. create our own Nemesis class tree).

pehala:
* Add NemesisRegistry tests
  * Replaces and extends existing tests for selectors and such
* Split Nemesis tests into 3 modules
   * test_registry - Tests NemesisRegistry directly
   * test_sisyphus - Tests Sisyphus and its respective functionality
   * test_specific_nemesis - Tests implementations of specific nemesis
      * Only tests that are still dependant on the implementation
* Introduce new class hierarchy for tests only
* Fixes #10592

### Testing

- [x] locally


fruch (approved):
LGTM

soyacz:
> > no backports needed?
> 
> We did not backport nemesis registry changes so probably not. Only release where it could make sense is 2025.2, as that was branched after the registry was merged

I see https://github.com/scylladb/scylla-cluster-tests/pull/10502 was backported to 2025.1 and as you mention it is in 2025.2

PR #10913, opened 2025-05-16
improvement(scylla_yaml): Remove simple parameters

pehala:
Currently, for every Scylla parameter there has to a field in `ScyllaYaml` even if the parameter is simple and we do not need any logic. This resulted in `ScyllaYaml` having countless of parameters which are out of sync with actual Scylla parameters, as we change this file only to add things, not to remove.
At the same this make testing new parameters more complicated as you have to add it here to have any effect.

This PR removes all simple parameters from ScyllaYaml and allows adding dynamic parameters to it.
It also updates validators to pydantic 2.0 style.

So far I tested only with unit tests, need to do a few testing runs to verify correctness.

### Testing

- [ ] TBD

### Reminders


pehala:
> I don't understand the motivation behind this PR and what you want to achieve here, we don't use this ScyllaYaml for defaults and we don't change them.

1) Get rid of unneccessary code
  * No need to specifiy every parameter we ever used
2) Make testing of new config parameters easier
  * No need to add to parameter to the class
3) Remove need to maintain this class
   * Currently, it is out of sync with the scylla config

We do not use it for defaults, we use it ONLY for partial name validation, which is abitrary as well, since it is out of sync with the scylla config.

pehala:
Upon discussion with @fruch, I am only gonna fix the need to add new parameter into `ScyllaYaml` everytime we want to add it into `append_scylla_yaml`, which is a bug in current implementation. The `ScyllaYaml` and how we handle the config in general (e.g. in k8s) is flawed regardless and should be redesigned independently.

Issue #10503, opened 2025-03-26
disrupt_refuse_connection_with_send_sigstop_signal_to_scylla_on_banned_node failed due to there is no raft quorum


timtimb0t:
## Packages

Scylla version: `2025.1.0~rc4-20250323.bc983017832c` with build-id `088ceb686f4b2d57120be368a4b86d6ac0a04cd5`

Kernel Version: `6.8.0-1024-aws`

disrupt_refuse_connection_with_send_sigstop_signal_to_scylla_on_banned_node nemesis failed with the following error:
```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5641, in wrapper
    result = method(*args[1:], **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5445, in disrupt_refuse_connection_with_send_sigstop_signal_to_scylla_on_banned_node
    self._refuse_connection_from_banned_node(use_iptables=False)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5475, in _refuse_connection_from_banned_node
    with self.run_nemesis(node_list=self.cluster.nodes,
  File "/usr/local/lib/python3.10/contextlib.py", line 576, in __exit__
    raise exc_details[1]
  File "/usr/local/lib/python3.10/contextlib.py", line 561, in __exit__
    if cb(*exc_details):
  File "/usr/local/lib/python3.10/contextlib.py", line 449, in _exit_wrapper
    callback(*args, **kwds)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5472, in drop_keyspace
    session.execute(f"DROP KEYSPACE IF EXISTS {keyspace_name}", timeout=300)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 1319, in execute_verbose
    return execute_orig(*args, **kwargs)
  File "cassandra/cluster.py", line 2762, in cassandra.cluster.Session.execute
  File "cassandra/cluster.py", line 5166, in cassandra.cluster.ResponseFuture.result
[...]

PR #10926, opened 2025-05-20
fix(_refuse_connection_from_banned_node): remove banned node from list before cql session

aleksbykov:
Banned node should be remove from node list before cql session to cluster will be initialized. in this case the ip address of the banned will be added to ip list for `get_node_cql_ips`.
More details in [scylladb/scylla-cluster-tests#10503](https://github.com/scylladb/scylla-cluster-tests/issues/10503#issuecomment-2894125344)

Fixes issue: scylladb/scylla-cluster-tests#10503

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -5357,6 +5357,7 @@ class Nemesis:
         simulate_node_unavailability = node_operations.block_scylla_ports if use_iptables else node_operations.pause_scylla_with_sigstop
         with self.run_nemesis(node_list=self.cluster.nodes,
                               nemesis_label=f"Running {simulate_node_unavailability.__name__}") as working_node, ExitStack() as stack:
+            stack.callback(drop_keyspace, node=working_node)
             target_host_id = self.target_node.host_id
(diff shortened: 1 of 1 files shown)


soyacz:
This looks like workaround to scylla issue - we cannot expect clients won't use banned nodes ip's - I thinks typically users just set list of ip's in clients and update it from time to time, not every time cluster topology changes or nodes get banned.
Please mark this as a workaround with comments and link issue to scylla/drivers.
besides, please comment what is `wlrr` as Copilot suggests.

soyacz:
> @soyacz can you take a look one more time

Why link to sct comment and not to scylladb issue (or driver issue - in that case python one)?

aleksbykov:
> > @soyacz can you take a look one more time
> 
> Why link to sct comment and not to scylladb issue (or driver issue - in that case python one)?

@soyacz , added python driver issue

temichus (commented):
same. LGTM, just rename nodes_for_wlrr

PR #10927, opened 2025-05-20
Add performance regression tests to Alternator

radoslawcybulski:
Add a support for hdr histograms to YCSB code.
Add a performance regression tests for alternator code using YCSB.
Add a jenkins pipeline, that runs those tests.
Add a configuration, that will run a performance regression test on master:latest weekly.

This patch adds a performance regression test suite for alternator. The suite contains 5 tests (only reads, only writes, half reads half writes for latency check and measure throughput for only reads and only writes). Those tests are run against cql (for base line), alternator with write isolation ONLY_RMW_USES_LWT and write isolation ALWAYS_USE_LWT).
Measured results are sent to Argus, where they can be viewed as time passes.

Following tests are runnable from jenkins interface:

test_full - all tests below
test_latency - all latency tests (read write mixed)
test_latency_read(self) - read only latency tests
test_latency_write(self) - write only latency tests
test_latency_mixed - read 50% write 50% latency tests
test_throughput - all throughput tests
test_throughput_read - read only throughput tests
test_throughput_write - write only throughput tests

Fixes #[1757](https://github.com/scylladb/qa-tasks/issues/1757)


test-cases/performance/perf-regression-alternator-basic.yaml
```diff
@@ -0,0 +1,82 @@
...
+stress_multiplier: 1
+alternator_write_isolation: 'forbid'
+alternator_port: 8080
+dynamodb_primarykey_type: HASH_AND_RANGE
+
+user_prefix: 'perf-alternator-latency'
+space_node_threshold: 644245094
+
+use_hdrhistogram: true
+scylla_version: '2025.1.1'
```
fruch, line 50: why hard-coding version here ?

test-cases/performance/perf-regression-alternator-latency-500gb-upgrade.yaml
```diff
@@ -17,38 +17,38 @@ prepare_write_cmd:
...
+ami_id_db_scylla_desc: 'VERSION_DESC'
 
 round_robin: true
 append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
 backtrace_decoding: false
+print_kernel_callstack: true
 
 store_perf_results: true
-email_recipients: ['scylla-perf-results@scylladb.com', 'alternator@scylladb.com']
+email_recipients: ["evgeniy.naydanov@scylladb.com"]
```
fruch, line 54: why hardcoding specific people ?

performance_regression_alternator_test.py
```diff
@@ -217,99 +381,373 @@ def test_mixed(self):
...
+        self.create_cql_ks_and_table(field_number=10)
+        self.preload_data()
+        self.wait_no_compactions_running()
+        self.run_fstrim_on_all_db_nodes()
 
-        self.check_regression_with_baseline('cql')
+    def test_latency_write_with_upgrade(self) -> None:
+        self._prepare_latency_with_upgrade()
+        self.run_workload_and_upgrade(
+            stress_cmd=self.params.get("stress_cmd_w") + " -target 3000",
```
fruch, line 737: why the -target 3000 ? isn't in the command itself in the test case yaml ?

performance_regression_alternator_test.py
```diff
@@ -28,8 +49,117 @@ def setUp(self):
         self.stack.enter_context(ignore_alternator_client_errors())
         self.stack.enter_context(ignore_operation_errors())
 
-    def _workload(self, stress_cmd, stress_num, test_name=None, sub_type=None, keyspace_num=1, prefix='', debug_message='',  # pylint: disable=too-many-arguments,arguments-differ
-                  save_stats=True, is_alternator=True):
+    def make_uuid_dir_for_hdr_files(self):
```
fruch, line 52: I'll would remove all of hdr handling logic out of the test, into it's own class in `ycsb_therad,` or under `sdcm.utils` 

it might be needed in other tests, not just this one (at some point).

performance_regression_alternator_test.py
```diff
@@ -39,15 +169,41 @@ def _workload(self, stress_cmd, stress_num, test_name=None, sub_type=None, keysp
...
+            self.build_histogram(hdr_workload, hdr_tags=self.hdr_tags)
+
         if save_stats:
             self.update_test_details(scylla_conf=True, alternator=is_alternator)
 
+
     def create_alternator_table(self, schema, alternator_write_isolation):
         node = self.db_cluster.nodes[0]
 
+        if self.params.get("alternator_enforce_authorization"):
```
fruch, line 196: can't this code call `pre_create_alternator_tables` ? or at least refactor that we'll have one call like this and both this and `pre_create_alternator_tables` would call it ?

sdcm/ycsb_thread.py
```diff
@@ -264,9 +264,14 @@ def _run_stress(self, loader, loader_idx, cpu_idx):  # pylint: disable=too-many-
                                    extra_docker_opts=f'--label shell_marker={self.shell_marker}',
                                    docker_network=self.params.get('docker_network'))
                 dns_options += f'--dns {dns.internal_ip_address} --dns-option use-vc'
+            os.makedirs(f'/tmp/hdr-output-directory/{loader_idx}/{cpu_idx}', exist_ok=True)
```
fruch, line 267: something doesn't add up here
why are we creating this directory like that ? the docker gonna run on the loader

it should be something like:
```python
tmp_hdr_dir  = loader.sudo(f"mktemp -d -t {loader_idx}_{cpu_idx} ").stdout
```

test-cases/performance/perf-regression-alternator-basic.yaml
radoslawcybulski: Testing leftover. Removed.

test-cases/performance/perf-regression-alternator-latency-500gb-upgrade.yaml
radoslawcybulski: Evgeniy put himself in when he was developing it. This file is also (i think) obsolete.

vars/perfRegressionParallelPipeline.groovy
```diff
@@ -32,8 +32,8 @@ def call(Map pipelineParams) {
                    description: 'GCE datacenter',
                    name: 'gce_datacenter')
 
-            string(defaultValue: '', description: '', name: 'scylla_ami_id')
-            string(defaultValue: '', description: '', name: 'scylla_version')
+            string(defaultValue: "${pipelineParams.get('scylla_ami_id', '')}", description: '', name: 'scylla_ami_id')
+            string(defaultValue: "${pipelineParams.get('scylla_version', '')}", description: '', name: 'scylla_version')
```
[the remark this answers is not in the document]
radoslawcybulski: I thought my code does that - pulls required values from arguments passed from jenkins?

jenkins-pipelines/performance_staging/perf-regression-alternator-latency-30min.jenkinsfile
```diff
@@ -7,6 +7,9 @@ perfRegressionParallelPipeline(
     backend: "aws",
     region: "us-east-1",
     test_name: "performance_regression_alternator_test.PerformanceRegressionAlternatorTest",
-    test_config: "test-cases/performance/perf-regression-alternator-latency-500gb-30min.yaml",
-    sub_tests: ["test_latency"],
+    test_config: "test-cases/performance/perf-regression-alternator-basic.yaml",
+    sub_tests: ["test_basic"],
+    scylla_version: 'master:latest',
```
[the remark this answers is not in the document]
fruch: I'm talking about scylla_version

The subtests are fine to hardcode

But versions shouldn't be

fruch:
> I don't know the rules inside SCT, but common sense suggests at least linking [the related issue](https://github.com/scylladb/qa-tasks/issues/1757) in the cover letter.

It is the same rules as having a clear description and commit messages 

But since I was following it technically before this PR, I didn't look into any description yet (if any).

Yes it would help to get reviews form others, and it won't get merged with those in place

jenkins-pipelines/performance_staging/perf-regression-alternator-latency-30min.jenkinsfile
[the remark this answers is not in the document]
radoslawcybulski: You're right, removed.

performance_regression_alternator_test.py
radoslawcybulski: Done.

performance_regression_alternator_test.py
radoslawcybulski: Done.

performance_regression_alternator_test.py
```diff
@@ -28,8 +49,117 @@ def setUp(self):
...
+                    )
+                    hdrh_logger.start()
+                    self.hdrh_logger_contextes.append(hdrh_logger)
+
+    def terminate_hdr_loggers(self):
+        for hdrh_logger in self.hdrh_logger_contextes:
+            hdrh_logger.stop()
+
+    def update_hdr_files(self, stress_num):
+        allowed_chars = frozenset('+,./0123456789=ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz')
```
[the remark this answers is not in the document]
radoslawcybulski: Done.

performance_regression_alternator_test.py
```diff
@@ -28,8 +49,117 @@ def setUp(self):
...
+                            pass
+        del self.directory_for_hdr_files
+
+    def initialize_hdr_loggers(self, stress_num):
+        self.hdrh_logger_contextes = []
+        for loader in self.loaders.nodes:
+            loader_index = loader.node_index
+            for work_type in work_types:
+                for stress_idx in range(0, stress_num):
+                    hdrh_logger = HDRHistogramFileLogger(
```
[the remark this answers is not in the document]
radoslawcybulski: Done.

sdcm/sct_config.py
```diff
@@ -1761,6 +1761,10 @@ class SCTConfiguration(dict):
              help="Workload name, can be: write|read|mixed|unset."
                   "Used for e.g. latency_calculator_decorator (use with 'use_hdrhistogram' set to true)."
                   "If unset, workload is taken from test name."),
+
+        dict(name="alternator_running_time_multiplier", env="ALTERNATOR_RUNNING_TIME_MULTIPLIER", type=float,
```
fruch, line 1765: SCT environment variables, should start with `SCT_`

Also I don't understand what this option is exactly for, it should be documented a bit clearer.
how running time is related to load here, again, really not clear.

fruch: it's sounds like it might be similar to `stress_duration`, but expressed in multiplier

`stress_duration` has the extra feature of setting the pipeline timeout correctly to accumulate the change in test duration, which your new flag probably lacks (hence doubling my get you to a timeout of the pipeline)

test-cases/performance/perf-regression-alternator-basic.yaml
```diff
@@ -0,0 +1,82 @@
...
+
+  - >-
+    bin/ycsb -jvm-args='-Dorg.slf4j.simpleLogger.defaultLogLevel=OFF' load dynamodb -P workloads/workloada -threads 100 -p recordcount=15000000 -p fieldcount=10 -p fieldlength=512
+    -p insertorder=hashed -p insertstart=10000000 -p insertcount=5000000 -p operationcount=5000000
+
+
+stress_cmd_w: >-2
+  bin/ycsb -jvm-args='-Dorg.slf4j.simpleLogger.defaultLogLevel=OFF' run dynamodb -P workloads/workloada -threads 100 -p recordcount=15000000
+  -p readproportion=0 -p updateproportion=1 -p scanproportion=0 -p insertproportion=0 -p requestdistribution=uniform
+  -p fieldcount=10 -p fieldlength=512 -p measurement.interval=intended -p measurementtype=hdrhistogram -p hdrhistogram.fileoutput=true -p status.interval=1 -p hdrhistogram.output.path=/tmp/hdr-output-directory/hdrh-
```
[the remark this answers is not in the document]
fruch: we should have hardcode paths like `/tmp/hdr-output-directory` in the configuration, it should be generated in the code

sdcm/sct_config.py
```diff
@@ -261,6 +261,10 @@ class SCTConfiguration(dict):
                   Time in minutes, Time of execution for stress commands from stress_cmd parameters
                   and is used in test duration calculation
              """),
+        dict(name="stress_target", env="SCT_STRESS_TARGET", type=int,
+             help="""
+                  Number of operations per second to achieve in stress commands.
```
fruch, line 266: in you case it effects all of the ycsb calls  ? can you elaborate  a bit more ? 

if it's currently only relevant to perf test, and for ycsb/alternator based on 
I would name it as such `alternator_perf_stress_target` 

otherwise people can assume they can put a number here, and it would rate limit their cassandra-stress command, while it's not gonna do that...

test-cases/performance/perf-regression-alternator-basic.yaml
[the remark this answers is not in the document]
fruch: code should be align that `use_hdrhistogram: true` would enable all the needed extensions to the command out of the code, including directories / filenames and such.

test-cases/performance/perf-regression-alternator-latency-500gb-with-nemesis.yaml
```diff
@@ -0,0 +1,59 @@
...
+alternator_port: 8080
+dynamodb_primarykey_type: HASH_AND_RANGE
+
+nemesis_class_name: 'NemesisSequence'
+nemesis_interval: 30
+nemesis_sequence_sleep_between_ops: 10
+
+user_prefix: 'perf-alternator-latency-nemesis'
+space_node_threshold: 644245094
+ami_id_db_scylla_desc: 'VERSION_DESC'
```
fruch, line 51: same here, you can drop this

fruch (changes requested):
- need to rebase and solve conflict
- we should have CI run (it's failing or didn't yet run)
- lots of small leftover that need to be remove
- from my POV 2 major issue, align the cycle title, and the hdr handling should be base on configuration (not hardcode pathes in test configureation), and even moved completly into ycsb_thread.py including the rewriteing of hdr tags.

performance_regression_alternator_test.py
```diff
@@ -28,8 +48,41 @@ def setUp(self):
...
+            self.log.error(f'unknown test_name {test_name} - some things might not work as expected')
+            
+        @latency_calculator_decorator(cycle_name=cycle_name)
+        def execute_workload_with_latency_calculator_decorator(self, *args, **kwargs):
+            return self._workload(*args, **kwargs)
+
+        ret = execute_workload_with_latency_calculator_decorator(self, hdr_workload=hdr_workload, test_name=test_name, stress_num=stress_num, **kwargs)
+        return ret
+    
+    def _workload(self, stress_cmd, stress_num=1, test_name=None, sub_type=None, keyspace_num=1, prefix='', debug_message='',  # pylint: disable=too-many-arguments,arguments-differ
```
soyacz, line 84: pylint is no longer supported

sdcm/utils/decorators.py
```diff
@@ -237,6 +238,8 @@ def wrapped(*args, **kwargs):  # noqa: PLR0914
                 workload = 'write'
             elif 'mixed' in test_name:
                 workload = 'mixed'
+            elif '_throughput' in test_name:
```
soyacz, line 241: your'e already overriding `workload_name` in params, so it could be handled by the next `elif` clause. Why it's needed?

test-cases/performance/perf-regression-alternator-latency-500gb-upgrade.yaml
```diff
@@ -17,38 +17,38 @@ prepare_write_cmd:
...
 
 n_db_nodes: 3
-n_loaders: 3
+n_loaders: 1
 n_monitor_nodes: 1
 
-instance_type_loader: 'c5.2xlarge'
-instance_type_monitor: 't3.small'
-instance_type_db: 'i3.4xlarge'
+instance_type_loader: 'c5.9xlarge'
```
soyacz, line 37: c7i should be better and more future-proof

performance_regression_alternator_test.py
```diff
@@ -28,8 +48,41 @@ def setUp(self):
         self.stack.enter_context(ignore_alternator_client_errors())
         self.stack.enter_context(ignore_operation_errors())
 
-    def _workload(self, stress_cmd, stress_num, test_name=None, sub_type=None, keyspace_num=1, prefix='', debug_message='',  # pylint: disable=too-many-arguments,arguments-differ
-                  save_stats=True, is_alternator=True):
+    def _prepare_and_execute_workload_with_latency_calculator_decorator(self, *, test_name, row_name, stress_num=1, **kwargs):
+        self.hdr_tags = ['read', 'write']
+        self.row_name_override = row_name
```
soyacz, line 53: we should enhance latency claculator decorator to accept row_name

soyacz:
@radoslawcybulski please run `bash ./docker/env/hydra.sh update-conf-docs` and commit changes made by it - so we won't see annoying comments by github-actions

docs/configuration_options.md
```diff
@@ -60,6 +60,15 @@ Time in minutes, Time of execution for stress commands from stress_cmd parameter
 **type:** int
 
 
+## **alternator_perf_stress_target** / SCT_ALTERNATOR_PERF_STRESS_TARGET
```
vponomaryov, line 63: This naming is missleading.
I would never guess that it is "rate" for "ycsb".
Also, if we set rate limit, why we say it is only for "perf"?

I would expect following:
- Name like "alternator_stress_rate" or "alternator_stress_rate_ops" or "ycsb_stress_rate_ops"...
- Be applied not only to perf tests, but in general using stress.

Also, what about supporting multiple values?

performance_regression_alternator_test.py
```diff
@@ -59,6 +125,14 @@ def create_alternator_table(self, schema, alternator_write_isolation):
...
         self.wait_no_compactions_running()
 
+    def drop_cql_table(self):
+        node = self.db_cluster.nodes[0]
+        with self.db_cluster.cql_connection_patient(node) as session:
+            try:
+                session.execute("DROP TABLE IF EXISTS ycsb.usertable;")
+                session.execute("DROP KEYSPACE IF EXISTS ycsb;")
+            except Exception as e:
+                self.log.error(f"Failed to drop CQL table: {e}")
```
vponomaryov, line 135: Need to add an empty line between these functions.

performance_regression_alternator_test.py
```diff
@@ -217,99 +293,350 @@ def test_mixed(self):
...
         base_cmd_w = self.params.get('stress_cmd_w')
         base_cmd_r = self.params.get('stress_cmd_r')
         base_cmd_m = self.params.get('stress_cmd_m')
 
-        stress_multiplier = 2
-        self.wait_no_compactions_running(n=120)
+        stress_multiplier = 1
 
+        is_basic = mode.startswith('basic')
+        run_read = mode == 'full' or mode == 'basic' or mode == 'basic-read'
```
vponomaryov, line 348: Such a logic may be simplified to the following:
```python
run_read = mode in ("full", "basic", "basic-read")
```

The same below.

performance_regression_alternator_test.py
```diff
@@ -217,99 +293,350 @@ def test_mixed(self):
...
+        run_write = mode == 'full' or mode == 'basic' or mode == 'basic-write'
+        run_mixed = mode == 'full' or mode == 'basic' or mode == 'basic-mixed'
+        run_read_throughput = mode == 'full' or mode == 'basic' or mode == 'basic-throoughput' or mode == 'basic-throughput-read'
+        run_write_throughput = mode == 'full' or mode == 'basic' or mode == 'basic-throoughput' or mode == 'basic-throughput-write'
+        self.hdr_tags = [ 'read', 'write' ]
+
```
vponomaryov, line 354: Redundant one line

performance_regression_alternator_test.py
```diff
@@ -217,99 +293,350 @@ def test_mixed(self):
...
+
+        def run_read_cql():
+            if is_basic: return
+            if not run_read: return
+            self._prepare_and_execute_workload_with_latency_calculator_decorator(
+                test_name=self.id() + '_read', sub_type='cql', stress_cmd=base_cmd_r + cmd_add_params, stress_num=stress_multiplier,
+                keyspace_num=1, is_alternator=False, row_name = 'cql')
+        def run_read_alternator_no_lwt():
```
vponomaryov, line 362: Need to add 1 empty line after each function

performance_regression_alternator_test.py
```diff
@@ -217,99 +293,350 @@ def test_mixed(self):
...
+        run_write = mode == 'full' or mode == 'basic' or mode == 'basic-write'
+        run_mixed = mode == 'full' or mode == 'basic' or mode == 'basic-mixed'
+        run_read_throughput = mode == 'full' or mode == 'basic' or mode == 'basic-throoughput' or mode == 'basic-throughput-read'
+        run_write_throughput = mode == 'full' or mode == 'basic' or mode == 'basic-throoughput' or mode == 'basic-throughput-write'
+        self.hdr_tags = [ 'read', 'write' ]
+
+
+        def run_read_cql():
+            if is_basic: return
+            if not run_read: return
```
vponomaryov, line 358:
```python
if is_basic or not run_read:
    return
```

The same is valid for other similar lines below.

performance_regression_alternator_test.py
```diff
@@ -217,99 +293,350 @@ def test_mixed(self):
...
+        )
+        self.check_latency_during_ops()
+
+
+class PerformanceRegressionAlternatorUpgradeTest(PerformanceRegressionAlternatorTest, UpgradeTest):  # pylint: disable=too-many-ancestors
+    def get_email_data(self) -> dict:
+        return PerformanceRegressionAlternatorTest.get_email_data(self)
+
+    @latency_calculator_decorator(legend="Upgrade Node")
+    def upgrade_node(self, node) -> None:  # pylint: disable=arguments-differ
```
vponomaryov, line 538: Pylint tool is not used in SCT anymore.

sdcm/send_email.py
```diff
@@ -581,7 +585,8 @@ class PerfSimpleQueryReporter(BaseEmailReporter):
 def build_reporter(name: str,  # noqa: PLR0911
                    email_recipients: Sequence[str] = (),
                    logdir: Optional[str] = None) -> Optional[BaseEmailReporter]:
-
+    # pylint: disable=too-many-return-statements,too-many-branches
```
vponomaryov, line 588: Same here, redundant pylint line

sdcm/utils/hdrhistogram.py
```diff
@@ -317,7 +333,7 @@ def _get_workload_type_by_hdr_tag(self, hdr_tag):
         hdr_tag = hdr_tag.lower().strip()
         if any(w_word in hdr_tag for w_word in ("write", "insert", "update", "delete")):
             return "WRITE"
-        elif any(r_word in hdr_tag for r_word in ("read", "select", "get", "count")):
+        elif any(r_word in hdr_tag for r_word in ("read", "select", "get", "count", "scan", "mixed")):
```
vponomaryov, line 336: "scan" is "read" - yes, but "mixed" is not.

This piece of code is used for each benchmarking tool we use with HDR tags.

So, it definitely cannot be left this way.

Also, it looks strange that HDR histogram is not really of the single real workload type - one of 2 - "read" or "write".

Running a "mixed" workload I would expect 2 HDR histogram tags - one for "reads" and one for "writes".

Anyway, if you are sure it must be still the same HDRH tag for reads and writes then need to use additional condition branch which will match only such specific cases.

test-cases/performance/perf-regression-alternator-basic.yaml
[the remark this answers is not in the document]
vponomaryov: +1
Example for the latte tool: https://github.com/scylladb/scylla-cluster-tests/blob/b14d7501332a63c44aa4b85041e93e0f72b557d2/sdcm/stress/latte_thread.py#L239-L253

performance_regression_alternator_test.py
radoslawcybulski: Reworked - now i get the value from additional parameters (SCT_***)

performance_regression_alternator_test.py
radoslawcybulski: It's because my code calls pre_create_alternator_tables instead and ignores this function.

performance_regression_alternator_test.py
radoslawcybulski: Done.

performance_regression_alternator_test.py
radoslawcybulski: Done.

performance_regression_alternator_test.py
radoslawcybulski: Not sure what do you mean?

performance_regression_alternator_test.py
radoslawcybulski: Done.

test-cases/performance/perf-regression-alternator-latency-500gb-upgrade.yaml
radoslawcybulski: Can you write full name i should use?

performance_regression_alternator_test.py
vponomaryov: Combine conditions leading to the same result and use "return" on a separate line.

sdcm/utils/hdrhistogram.py
radoslawcybulski: Removed mixed, seems to be unncecesary, as ycsb code reports here either 'read' or 'write'.

sdcm/ycsb_thread.py
```diff
@@ -242,6 +262,98 @@ def parse_final_output(result):
...
+                                data2 = []
+                            if d.startswith('#') or d.startswith("'"):
+                                data2.append(d)
+                            elif d.startswith('tail: '):
+                                ignored_tail_lines += 1
+                            else:
+                                if d[0].isdigit() or d[0] == '.':
+                                    data2.append(f'{tag_text},{d}')
+                                else:
+                                    data2.append(d)
```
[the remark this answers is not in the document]
vponomaryov: +1

It also can be simplified by removing the redundant middle `else`:
```
                            if d.startswith('#') or d.startswith("'"):
                                data2.append(d)
                            elif d.startswith('tail: '):
                                ignored_tail_lines += 1
                            elif d[0].isdigit() or d[0] == '.':
                                data2.append(f'{tag_text},{d}')
                            else:
                                data2.append(d)
```

But, first and last conditions have the same logic under it.
So, it may be simplified even more:
```
                            if d.startswith('tail: '):
                                ignored_tail_lines += 1
                            elif d[0].isdigit() or d[0] == '.':
                                data2.append(f'{tag_text},{d}')
                            else:
                                data2.append(d)
```

Then, IIUC, the first digit case is the most happening one.
And in this case we'd better update it the following way:
```
                            if d[0].isdigit() or d[0] == '.':
                                data2.append(f'{tag_text},{d}')
                            elif d.startswith('tail: '):
                                ignored_tail_lines += 1
                            else:
                                data2.append(d)
```

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+        return PerformanceRegressionAlternatorTest.get_email_data(self)
+
+    @latency_calculator_decorator(legend="Upgrade Node")
+    def upgrade_node(self, node) -> None:
+        InfoEvent(message=f"Upgrade Node {node.name} begin").publish()
+        self._upgrade_node(node)
+        InfoEvent(message=f"Upgrade Node {node.name} ended").publish()
+
+    def _stop_stress_when_finished(self) -> None:
+        with EventsSeverityChangerFilter(new_severity=Severity.NORMAL,  # killing stress creates Critical error
```
ScyllaPiotr, line 555: The comment made me laugh :wink:

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+                                         event_class=YcsbStressEvent,
+                                         extra_time_to_expiration=60):
+            self.loaders.kill_stress_thread()
+
+    @latency_calculator_decorator
+    def steady_state_latency(self) -> None:
```
ScyllaPiotr, line 561: I don't understand what it does, could the name be clearer or could a comment be added?

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+        sleep_time = self.db_cluster.params.get("nemesis_interval") * 60
+        InfoEvent(message=f"Starting Steady State calculation for {sleep_time}s").publish()
+        time.sleep(sleep_time)
+        InfoEvent(message=f"Ended Steady State calculation. Took {sleep_time}s").publish()
+
+    @latency_calculator_decorator
+    def post_upgrades_steady_state(self) -> None:
```
ScyllaPiotr, line 568: ditto

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+        self.create_test_stats(
+            sub_type=sub_type,
+            append_sub_test_to_name=False,
+            test_index=test_index,
+        )
+        stress_queue = self.run_stress_thread(
+            stress_cmd=stress_cmd,
+            stats_aggregate_cmds=False,
+        )
+        time.sleep(60)  # postpone measure steady state latency to skip start period when latency is high
```
ScyllaPiotr, line 590: I don't understand what happens, when and why.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+            self.upgrade_node(node)
+            target_version, target_build_id = _get_version_and_build_id_from_node(node)
+            versions_list.append({
+                "base_version": base_version,
+                "base_build_id": base_build_id,
+                "target_version": target_version,
+                "target_build_id": target_build_id,
+                "node_name": node.name,
+            })
+            time.sleep(120)  # sleeping 2 min to give time for cache to re-heat
```
ScyllaPiotr, line 610: So I assume the node is faced with a traffic during those 2mins which populates the cache?

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+        self._stop_stress_when_finished()
+        results = self.get_stress_results(queue=stress_queue)
+        self.update_test_details(scrap_metrics_step=60, alternator=True)
+        self.display_results(results, test_name="test_alternator_latency_with_upgrade")
+        self.update_test_details(scrap_metrics_step=60, alternator=True)
+        self.display_results(results, test_name="test_alternator_latency_during_upgrade")
+        self.check_latency_during_ops()
+
+    def _prepare_latency_with_upgrade(self) -> None:
+        self.run_fstrim_on_all_db_nodes()
```
ScyllaPiotr, line 624: The placement of `run_fstrim_on_all_db_nodes` seems random. Here it's called 2 times, at the beginning and at the end, while in other places it's called only once in the middle. What's the rule?

performance_regression_alternator_test.py
ScyllaPiotr: It's not reworked yet, right? Here and 2x below.

sdcm/db_stats.py
```diff
@@ -764,6 +764,7 @@ def get_prometheus_stats(self, alternator=False, scrap_metrics_step=None):
     def update_hdrhistograms(self, histogram_name, histogram_data):
         if "histograms" not in self._stats['results']:
             self._stats['results']['histograms'] = {}
+        LOGGER.info(f'QWERTY histogram_name: {histogram_name}, histogram_data: {histogram_data}')
```
ScyllaPiotr, line 767: All these `QWERTY` instances need to be changed or purged.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+            run_write_alternator_with_lwt,
+            run_mixed_cql,
+            run_mixed_alternator_no_lwt,
+            run_mixed_alternator_with_lwt,
+            run_read_throughput_cql,
+            run_read_throughput_alternator_no_lwt,
+            run_write_throughput_cql,
+            run_write_throughput_alternator_no_lwt,
+        )
+        duration = self.params.get('stress_duration') / len(tests_to_run)
```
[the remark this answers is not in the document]
ScyllaPiotr: I see `stress_duration` is not defined in your code.

sdcm/send_email.py
```diff
@@ -561,6 +561,10 @@ class SnitchEmailReporter(BaseEmailReporter):
     pass
 
 
+class PerformanceRegressionAlternatorTestEmailReporter(BaseEmailReporter):
```
ScyllaPiotr, line 564: Please comment on this empty class.

sdcm/send_email.py
```diff
@@ -581,7 +585,7 @@ class PerfSimpleQueryReporter(BaseEmailReporter):
 def build_reporter(name: str,  # noqa: PLR0911
                    email_recipients: Sequence[str] = (),
                    logdir: Optional[str] = None) -> Optional[BaseEmailReporter]:
-
+    LOGGER.info("Building email reporter for class: %s", name)
```
ScyllaPiotr, line 588: Just so you know, you omitted the `QWERTY` here.

test-cases/performance/perf-regression-alternator-basic.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+use_hdrhistogram: true
+alternator_stress_rate: 15000
+
+round_robin: true
+append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
+backtrace_decoding: false
+
+store_perf_results: true
+email_recipients: ['alternator@scylladb.com']
```
ScyllaPiotr, line 59: This destination is incorrect for the automated test reports. What email address do the other (non-alternator) tests use?

test-cases/performance/perf-regression-alternator-full.yaml
```diff
@@ -0,0 +1,94 @@
...
+
+use_hdrhistogram: true
+alternator_stress_rate: 15000
+
+round_robin: true
+append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
+backtrace_decoding: false
+
+store_perf_results: true
+email_recipients: ['alternator@scylladb.com']
```
ScyllaPiotr, line 59: This destination is incorrect for the automated test reports. What email address do the other (non-alternator) tests use?

ScyllaPiotr (changes requested):
We're far from finished.

sdcm/argus_results.py
```diff
@@ -171,8 +171,10 @@ def submit_results_to_argus(argus_client: ArgusClient, result_table: GenericResu
             raise
 
 
-def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, description: str, cycle: int, result: dict,
+def send_result_to_argus(argus_client: ArgusClient, workload: str, name: str, description: str, cycle: int | str, result: dict,
                          start_time: float = 0, error_thresholds: dict = None):
+    if type(cycle) is int:
```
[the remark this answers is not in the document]
radoslawcybulski: @fruch @soyacz Not sure what do you mean.

The `send_result_to_argus` function takes `cycle` parameter (an integer) and creates `row_name` from it. There's no way to override it currently (i just added one) and everyone (except me) is actually relying on it passing the int and making `row_name` from it. I'm not a stakeholder in this code, i can make some changes if @ScyllaPiotr agrees on time expenditure, but i need pointers. From my perspective i don't understand, why you want me to do that (it's working fine, everyone is using it), but i've no access to your plans for this code, so i'm missing a lot of info.

performance_regression_alternator_test.py
```diff
@@ -39,15 +92,38 @@ def _workload(self, stress_cmd, stress_num, test_name=None, sub_type=None, keysp
...
+            self.db_cluster.add_nemesis(nemesis=self.get_nemesis_class(), tester_obj=self)
+            self.db_cluster.start_nemesis(interval=interval, cycles_count=1)
+            self._stop_load_when_nemesis_threads_end()
+
         stress_queue = self.run_stress_thread(stress_cmd=stress_cmd, stress_num=stress_num, keyspace_num=keyspace_num,
                                               prefix=prefix, stats_aggregate_cmds=False)
         self.get_stress_results(queue=stress_queue, store_results=True)
+        if self.params['use_hdrhistogram']:
+            assert hdr_workload is not None, "hdr_workload must be provided when use_hdrhistogram is True"
+            self.build_histogram(hdr_workload, hdr_tags=self.hdr_tags)
```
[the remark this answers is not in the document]
radoslawcybulski: I moved hdr handling code to ycsb_thread.py module.

test-cases/performance/perf-regression-alternator-latency-500gb-with-nemesis.yaml
radoslawcybulski: Removed.

performance_regression_alternator_test.py
radoslawcybulski: Removed all pylint markers from the code.

test-cases/performance/perf-regression-alternator-latency-500gb-upgrade.yaml
radoslawcybulski: @soyacz I went with c7i.2xlarge - not sure why we needed c5.9xlarge for loader node. I will run some tests and make sure it works fine.

sdcm/argus_results.py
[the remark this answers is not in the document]
soyacz: I'm not insisting, this could be moved to followup as this effort is already is quite big.

sdcm/ycsb_thread.py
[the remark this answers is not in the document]
radoslawcybulski: Done.

test-cases/performance/perf-regression-alternator-latency-500gb-with-nemesis.yaml
```diff
@@ -17,37 +17,39 @@ prepare_write_cmd:
...
+nemesis_sequence_sleep_between_ops: 10
 
 round_robin: true
 append_scylla_args: '--blocked-reactor-notify-ms 5 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1'
 backtrace_decoding: false
+print_kernel_callstack: true
 
 store_perf_results: true
-email_recipients: ['scylla-perf-results@scylladb.com', 'alternator@scylladb.com']
+email_recipients: ["evgeniy.naydanov@scylladb.com"]
```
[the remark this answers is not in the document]
radoslawcybulski: Removed.

performance_regression_alternator_test.py
radoslawcybulski: Updated, good catch, i guess i put too much faith in cql error reporting.

performance_regression_alternator_test.py
```diff
@@ -39,15 +86,34 @@ def _workload(self, stress_cmd, stress_num, test_name=None, sub_type=None, keysp
         if save_stats:
             self.create_test_stats(test_name=test_name, sub_type=sub_type,
                                    doc_id_with_timestamp=True, append_sub_test_to_name=False)
+        self.log.info(f'QWERTY starting stress cmd: {stress_cmd}')
         stress_queue = self.run_stress_thread(stress_cmd=stress_cmd, stress_num=stress_num, keyspace_num=keyspace_num,
                                               prefix=prefix, stats_aggregate_cmds=False)
-        self.get_stress_results(queue=stress_queue, store_results=True)
+        self.log.info(f'QWERTY started stress cmd: {stress_cmd}')
```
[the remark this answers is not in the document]
radoslawcybulski: Done.

performance_regression_alternator_test.py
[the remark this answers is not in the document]
radoslawcybulski: Updated.
stress_duration comes from sct_config magic.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+        )
+        duration = self.params.get('stress_duration') / len(tests_to_run)
+        try:
+            target = self.params.get('alternator_stress_rate')
+            self.log.info(f"Using target {target} for stress tests.")
+        except KeyError:
+            target = 15000
+            self.log.info(f"Using default target {target} for stress tests.")
+        if duration < 1:
+            self.log.warning(f"Stress duration {self.params.get('stress_duration')} after dividing by {len(tests_to_run)} is less than 1 minute, setting it to 1 minute.")
```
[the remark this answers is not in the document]
radoslawcybulski: Updated.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+            duration = 1
+        cmd_add_params = f" -target {target} -p maxexecutiontime={int(60 * duration)}"
+        cmd_add_throughput_params = f" -target 999999 -p maxexecutiontime={int(60 * duration)}"
+
+        self.pre_create_alternator_tables()
+        self.alternator.set_write_isolation(node=node, isolation=alternator.enums.WriteIsolation.ONLY_RMW_USES_LWT)
```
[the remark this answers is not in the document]
radoslawcybulski: Possibly, but i'd rather put it on hold, so we could get some usage data and see what do we need. Every such option is a code that needs to handle it and makes the whole thing less and less understandable (as you keep hunting for configuration values across multiple files).

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+        self.alternator.set_write_isolation(
+            node=self.db_cluster.nodes[0],
+            isolation=alternator.enums.WriteIsolation.ALWAYS_USE_LWT,
+        )
         self._workload(
-            test_name=self.id() + '_write', sub_type='cql', stress_cmd=base_cmd_w + " -target 10000",
-            stress_num=stress_multiplier, keyspace_num=1, is_alternator=False)
+            test_name=self.id() + "_write",
+            sub_type="with-lwt",
+            stress_cmd=self.params.get("stress_cmd_w") + " -target 3000",
```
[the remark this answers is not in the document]
radoslawcybulski: That's not mine. :)
I'm not yet confident enough to know it's not used and remove it - i've a suspicion it's not (possibly yet) used and @evgeny-kirichuk was preparing a nemesis test (a test, where something is forced to go wrong, aka node is killed or something). We're not at that stage yet, so i leave it be for now.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+    def get_email_data(self) -> dict:
+        return PerformanceRegressionAlternatorTest.get_email_data(self)
+
+    @latency_calculator_decorator(legend="Upgrade Node")
+    def upgrade_node(self, node) -> None:
+        InfoEvent(message=f"Upgrade Node {node.name} begin").publish()
+        self._upgrade_node(node)
+        InfoEvent(message=f"Upgrade Node {node.name} ended").publish()
+
+    def _stop_stress_when_finished(self) -> None:
```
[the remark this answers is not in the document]
radoslawcybulski: You'd be suprised. :)
Also Not my code. :)

performance_regression_alternator_test.py
radoslawcybulski: Not my code. :)

performance_regression_alternator_test.py
radoslawcybulski: Not my code. :)

performance_regression_alternator_test.py
radoslawcybulski: Not my code. :) As in previous comments - i'm not yet wersed enough to be sure i can remove those (or comment out).

performance_regression_alternator_test.py
radoslawcybulski: I've not touched this part at all. As in previous comments, i don't understand the whole code base good enough to be sure i can remove it / comment out it safely.

sdcm/db_stats.py
radoslawcybulski: Removed.

sdcm/send_email.py
radoslawcybulski: I will leave this comment for now - i know guys are fiddling with email reporting (read - removing it) and we will probably follow suit and move to Argus somehow, but for now this log can help understand why email reporting fails.

test-cases/performance/perf-regression-alternator-basic.yaml
radoslawcybulski: qa@scylladb.com
scylla-operator@scylladb.com
scylla-perf-results@scylladb.com
operator@scylladb.com

wojciech.mitros@scylladb.com
artsiom.mishuta@scylladb.com
piodul@scylladb.com

test-cases/performance/perf-regression-alternator-full.yaml
radoslawcybulski: See the other comment please.

performance_regression_alternator_test.py
```diff
@@ -28,6 +48,33 @@ def setUp(self):
...
+            self.params['workload_name'] = 'read'
+            cycle_name = '100% read'
+        elif test_name.endswith('_write'):
+            self.params['workload_name'] = 'write'
+            cycle_name = '100% write'
+        elif test_name.endswith('_mixed'):
+            self.params['workload_name'] = 'mixed'
+            cycle_name = '50% read 50% write'
+        else:
+            self.log.error(f'unknown test_name {test_name} - some things might not work as expected')
```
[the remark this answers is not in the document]
ScyllaPiotr: I vote to comment in the code above the fun what values of `test_name` are expected and what would be the results of each value.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
-        with round_robin and list of stress_cmd - the data will load several times faster.
-        2. Run READ workload with cql.
-        3. Run READ workload without lwt.
-        4. Run READ workload with lwt.
-        5. Run WRITE workload with cql.
-        6. Run WRITE workload without lwt.
-        7. Run WRITE workload with lwt.
-        8. Run MIXED workload with cql.
-        9. Run MIXED workload without lwt.
+         1. Prepare cluster with data (reach steady_state of compactions and ~x10 capacity than RAM.)
```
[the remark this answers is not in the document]
ScyllaPiotr: Regarding both these messages, how to verify we're there or not there? Is anything specfic checked, or it is only hoped that after some specific time the state has been achieved?

sdcm/tester.py
```diff
@@ -3747,12 +3747,11 @@ def save_email_data(self):
 
         json_file_path = os.path.join(self.logdir, "email_data.json")
 
-        if email_data:
```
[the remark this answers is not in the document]
fruch: it can fail

performance_regression_alternator_test.py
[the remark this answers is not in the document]
ScyllaPiotr: > That's not mine. :)

GH shows this line as part of your modified code. I understand you might have seen it somewhere in this form first, but I propose to improve anyway.

> I'm not yet confident enough to know it's not used and remove it - i've a suspicion it's not (possibly yet) used and @evgeny-kirichuk was preparing a nemesis test (a test, where something is forced to go wrong, aka node is killed or something). We're not at that stage yet, so i leave it be for now.

We will leave the unclear value and will forget to go back to fix it. I propose otherwise.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
-            test_name=self.id() + '_write', sub_type='without-lwt', stress_cmd=base_cmd_w + " -target 10000",
-            stress_num=stress_multiplier, keyspace_num=1)
+            test_name=self.id() + "_read",
+            sub_type="with-lwt",
+            stress_cmd=self.params.get("stress_cmd_r") + " -target 3000",
+            nemesis=True,
+        )
+        self.check_latency_during_ops()
+
+    def test_latency_mixed_with_nemesis(self):
```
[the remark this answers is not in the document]
ScyllaPiotr: We're already improving now. Laying it off for later could mean we will forget this, especially that you resolve many comments that are not handled.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
-        self.run_fstrim_on_all_db_nodes()
+    def test_latency_write_with_nemesis(self):
+        self.create_alternator_table(
+            schema=self.params.get("dynamodb_primarykey_type"),
+            alternator_write_isolation=alternator.enums.WriteIsolation.FORBID_RMW,
+        )
+        self.create_cql_ks_and_table(field_number=10)
 
         self.wait_no_compactions_running()
+        self.run_fstrim_on_all_db_nodes()
```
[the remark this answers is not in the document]
fruch: years ago it scylla was doing it (i.e. there was a process doing so)

since then scylla moved to using other method that doesn't needs it (online discard), and today in production cluster no on is using fstrim regularly 

I believe we can remove it.

performance_regression_alternator_test.py
fruch: so can remove those unneeded changes  ?

performance_regression_alternator_test.py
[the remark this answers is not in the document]
ScyllaPiotr: This went to Slack where @pdbossman wrote the following:
> Fstrim is used typically when discard mount option is off, and is scheduled weekly.
It is used in several heavy write workload clusters in production [...]
Discard mount option was associated with high latencies during mass file delete processes that can happen when repairs/compactions finish up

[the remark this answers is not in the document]
fruch: none of what he mentioned is relevant to performance test we are running today, where online discard is now default.

and there no real reason to do this fstrim, before and after each workload we are running.
it was left there for years (those were introduce into those test more then 7 years ago) cause it wasn't cause no harm.

I don't think there a real reason to introduce those into newer tests

maybe @roydahan has more information we can pull his memory about those.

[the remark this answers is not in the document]
roydahan: There is no harm in doing so.
It can help reducing flakiness sometimes. I think we still do so in all our perf tests.

test-cases/performance/perf-regression-alternator-basic.yaml
ScyllaPiotr: @radoslawcybulski please use the address `alternator-perf-results@scylladb.com` for all the Alternator performance reports, the address has been created successfully.

vars/perfRegressionParallelPipeline.groovy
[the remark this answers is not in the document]
radoslawcybulski: Removed.

vars/perfRegressionParallelPipeline.groovy
```diff
@@ -138,6 +138,9 @@ def call(Map pipelineParams) {
             string(defaultValue: 'next',
                    description: 'Default branch to be used for scylla and other repositories. Default is "next".',
                    name: 'byo_default_branch')
+            string(defaultValue: "${pipelineParams.get('running_time_multiplier', '1')}",
```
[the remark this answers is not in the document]
radoslawcybulski: Removed.

sdcm/ycsb_thread.py
```diff
@@ -242,6 +262,98 @@ def parse_final_output(result):
...
+        os.makedirs(hdr_files_directory, exist_ok=True)
+        files = glob.glob(f'{hdr_files_directory}/*.hdr')
+        for f in files:
+            LOGGER.debug(f'removing old hdr file: {f}')
+            os.remove(f)
+        try:
+            os.chmod('/tmp/hdr-output-directory', 0o777)
+        except Exception:
+            LOGGER.exception(f'chmod failed for /tmp/hdr-output-directory')
+        return hdr_files_directory
```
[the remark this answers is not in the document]
radoslawcybulski: Done.

performance_regression_alternator_test.py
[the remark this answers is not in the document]
radoslawcybulski: I stand corrected, it seems in current code back test_name is used to distinct test results between each other (hence self.id is used, when creating one). Otherwise it can be any value as long as it's unique. I seem to recall it was used for something else, but this code was removed.
Added some comment.

performance_regression_alternator_test.py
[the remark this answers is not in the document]
radoslawcybulski: Removed.

performance_regression_alternator_test.py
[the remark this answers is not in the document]
radoslawcybulski: As per talk we had some time ago i removed the what i believe is dead code. It seems to be working fine so far.

performance_regression_alternator_test.py
```diff
@@ -217,99 +294,360 @@ def test_mixed(self):
...
+        self.alternator.set_write_isolation(
+            node=self.db_cluster.nodes[0],
+            isolation=alternator.enums.WriteIsolation.ALWAYS_USE_LWT,
```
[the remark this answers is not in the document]
radoslawcybulski: As per talk we had some time ago i removed the what i believe is dead code.

performance_regression_alternator_test.py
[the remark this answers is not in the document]
radoslawcybulski: As per talk we had some time ago i removed the what i believe is dead code.

performance_regression_alternator_test.py
radoslawcybulski: As per talk we had some time ago i removed the what i believe is dead code.

test-cases/performance/perf-regression-alternator-basic.yaml
radoslawcybulski: Done.

performance_regression_alternator_test.py
ScyllaPiotr: Agreed, now it's not part of the patch anymore.

performance_regression_alternator_test.py
ScyllaPiotr: Agreed, now it's not part of the patch anymore.

sdcm/ycsb_thread.py
```diff
@@ -114,7 +119,36 @@ def run(self):
...
     DOCKER_IMAGE_PARAM_NAME = "stress_image.ycsb"
-
+    WORK_TYPES = {
+        'READ': 'read', 
+        'SCAN': 'read',
+        'UPDATE': 'write',
+        'INSERT': 'write',
+        'DELETE': 'write',
+        'WRITE': 'write',
+    }
```
ScyllaPiotr, line 129: `performance_regression_alternator_test.py` has identical definition. I wonder if we can use a single definition so the usage in 2 places cannot diverge.

test-cases/performance/perf-regression-alternator-basic.yaml
ScyllaPiotr, line 1: The whole file should be analysed in order to create a human-readable paragraph into [the design doc](https://docs.google.com/document/d/1iuRCeuLjUi2Nxp0sPUZmHKZpRChIfROuUWIsdZnree0/) about the test config.

test-cases/performance/perf-regression-alternator-full.yaml
ScyllaPiotr, line 1: The whole file should be analysed in order to create a human-readable paragraph into [the design doc](https://docs.google.com/document/d/1iuRCeuLjUi2Nxp0sPUZmHKZpRChIfROuUWIsdZnree0/) about the test config.

performance_regression_alternator_test.py
```diff
@@ -71,6 +146,7 @@ def create_cql_ks_and_table(self, field_number):
 
     @optional_stage('perf_preload_data')
     def preload_data(self, compaction_strategy=None):
```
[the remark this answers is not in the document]
radoslawcybulski: At this point (ie version 1) row count is hardcoded in configuration (test-cases/performance/perf-regression-alternator-basic.yaml - see prepare_write_cmd or stress_cmd_X). As i mentioned in some talk before, i think there is a case for moving row_count / item size into configuration. I'd rather do it later on, once we get solve faster initialization of scylla.

The function returns after it's completed, at the end of the function you've a call to `self.get_stress_results`, which is like a `join` for threads - it waits for all asynchronous stuff to be completed.

docs/configuration_options.md
```diff
@@ -60,6 +60,15 @@ Time in minutes, Time of execution for stress commands from stress_cmd parameter
 **type:** int
 
 
+## **alternator_stress_rate** / SCT_ALTERNATOR_STRESS_RATE
+
+Number of operations per second to achieve in stress commands for alternator testing.
```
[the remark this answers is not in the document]
radoslawcybulski: Updated.

sdcm/send_email.py
radoslawcybulski: Updated.

performance_regression_alternator_test.py
```diff
@@ -108,208 +151,185 @@ def preload_data(self, compaction_strategy=None):
...
+            run_write_alternator_no_lwt,
+            run_write_alternator_with_lwt,
+            run_mixed_cql,
+            run_mixed_alternator_no_lwt,
+            run_mixed_alternator_with_lwt,
+            run_read_throughput_cql,
+            run_read_throughput_alternator_no_lwt,
+            run_write_throughput_cql,
+            run_write_throughput_alternator_no_lwt,
+        )
```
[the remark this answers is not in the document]
radoslawcybulski: Technically you probably want this:

```
    def test_full(self):
        self.run_test_suite_by_configuration_name('full')

    def test_latency(self):
        self.run_test_suite_by_configuration_name('basic-read')

    def test_latency_read(self):
        self.run_test_suite_by_configuration_name('basic-read')

    def test_latency_write(self):
        self.run_test_suite_by_configuration_name('basic-write')

    def test_latency_mixed(self):
        self.run_test_suite_by_configuration_name('basic-mixed')

    def test_throughput(self):
        self.run_test_suite_by_configuration_name('basic-throughput')

    def test_throughput_read(self):
        self.run_test_suite_by_configuration_name('basic-throughput-read')

    def test_throughput_write(self):
        self.run_test_suite_by_configuration_name('basic-throughput-write')
```

as what you cited is only callable indirectly. Updated.

performance_regression_alternator_test.py
```diff
@@ -108,208 +151,185 @@ def preload_data(self, compaction_strategy=None):
...
-        self._workload(sub_type='with-lwt', stress_cmd=base_cmd_w, stress_num=stress_multiplier, keyspace_num=1)
+    def test_throughput_read(self):
+        self.run_test_suite_by_configuration_name('basic-throughput-read')
 
-        self.check_regression_with_baseline('cql')
+    def test_throughput_write(self):
+        self.run_test_suite_by_configuration_name('basic-throughput-write')
 
-    def test_read(self):
+    def run_test_suite_by_configuration_name(self, mode):
```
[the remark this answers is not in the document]
radoslawcybulski: 1. not following, i think i'm missing something?
2. that's why those "executor" functions are defined inside the run_test_suite_by_configuration_name instead of being just members of the class. The logic is encoded in one whole function body, divided in such a way, that was sort-of manageable. To do what you want i'd have to unroll the loop increasing code lines. With my solution you have a separation - if you care about `run_write_cql`, you check only `def run_write_cql ...` and ignore the rest.

test-cases/performance/perf-regression-alternator-basic.yaml
radoslawcybulski: Don't think the design doc is a good place for implementation details. The file itself is messy (in general you need to go to the source code to at least get a glimpse of what is doing what) and i don't think it's a good place for comments either. Those variables are documented here (docs/configuration_options.md), not sure if we can specify all of them if we need, but i think so.

sdcm/utils/remote_logger.py
```diff
@@ -77,6 +77,7 @@ def __init__(self, node: BaseNode, target_log_file: str):
         super().__init__(target_log_file=target_log_file)
         self._node = node
         self._termination_event = ThreadEvent()
+        self._terminated_event = ThreadEvent()
```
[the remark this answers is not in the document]
fruch: I'm just saying you are trying to fix something, which isn't clear to me what the problem is.

you are referring to some log spam, and I don't know what it is and if it's important, is the the SCT log ? or the HDR log ?

in those case it helps if it's on it own commit with the explanation.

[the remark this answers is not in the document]
radoslawcybulski: The `_journal_thread` function logs some stuff, when it runs, for example

```
LOGGER.debug("Start check if remoter ready. %s", self._remote_log_file)
```

This happens in a loop. If for some reason command run by `retrieve` fails, you will get spam, because `_journal_thread` will try again immediately to call `_retrieve`  (and fail and log again and so on). I've seen 10+ gb of logs generated in a blink of an eye. Hence my change forcing `.wait` in a loop, which will prevent this behaviour.

sdcm/utils/remote_logger.py
```diff
@@ -142,20 +144,35 @@ def start(self) -> None:
...
-            self._remoter.run(f"pkill -f '{self._remote_log_file}'", ignore_status=True)
-            if self._thread.running():
-                self._thread.cancel()
-            self._started = False
+            if self._started:
+                LOGGER.debug(f"Stopping HDRHistogramFileLogger for {self._remote_log_file} -> {self._target_log_file}")
+                self._remoter.run(f"pkill -f '{self._remote_log_file}'", ignore_status=True)
+                self._termination_event.set()
+                self._terminated_event.wait(timeout=5)
+                # ensure the tail command to be stopped
```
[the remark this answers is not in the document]
radoslawcybulski: Moved.

performance_regression_alternator_test.py
```diff
@@ -108,208 +151,185 @@ def preload_data(self, compaction_strategy=None):
...
+        single_test_duration_in_seconds = int(60 * self.params.get('stress_duration') / len(tests_to_run))
+        target_ops_per_sec_for_unlimited_scenario = 999999
+        try:
+            target_ops_per_sec = self.params.get('alternator_stress_rate')
+            self.log.info(f"Using target {target_ops_per_sec} ops/s for stress tests.")
+        except KeyError:
+            target_ops_per_sec = 15000
+            self.log.info(f"Parameter alternator_stress_rate not found, using default target {target_ops_per_sec} ops/s for stress tests.")
+        if single_test_duration_in_seconds < 60:
+            self.log.warning(f"Single stress duration {self.params.get('stress_duration')}s after dividing by {len(tests_to_run)} is less than 1 minute, setting it to 1 minute.")
```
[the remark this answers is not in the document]
radoslawcybulski: Updated.

sdcm/utils/hdrhistogram.py
```diff
@@ -312,12 +327,13 @@ def _get_workload_type_by_hdr_tag(self, hdr_tag):
         #    Examples: 'fn--write', 'fn--write-batch', 'fn--get', 'fn--get-many', 'fn--read'.
         # 3) 'scylla-bench' has identical tag names for reads and writes - 'co-fixed' and 'raw'.
         #    It doesn't have 'mixed' workload type, so it's mode should be used for detecting the tag data type.
-        # 4) NOT_SUPPORTED: 'ycsb', it supports HDR histograms, but doesn't use tags in it.
-        #    So, the 'ycsb' case should be handled separately.
+        # 4) 'ycsb', it supports HDR histograms, but doesn't use tags in it.
```
vponomaryov, line 330: Please, add info about tag usage like it is done for the first 3 points above.
Does it follow one of above ways? Does it have 4th unique approach?
It will be useful for anyone later trying to understand/debug ycsb-based test runs.

sdcm/utils/hdrhistogram.py
```diff
@@ -312,12 +327,13 @@ def _get_workload_type_by_hdr_tag(self, hdr_tag):
...
+        #    We add tag to files ourselves.
         hdr_tag = hdr_tag.lower().strip()
-        if any(w_word in hdr_tag for w_word in ("write", "insert", "update", "delete")):
+        LOGGER.debug(f'Checking hdr_tag {hdr_tag} for workload type detection')
+        if any(w_word in hdr_tag.lower() for w_word in ("write", "insert", "update", "delete")):
             return "WRITE"
-        elif any(r_word in hdr_tag for r_word in ("read", "select", "get", "count")):
+        elif any(r_word in hdr_tag.lower() for r_word in ("read", "select", "get", "count", "scan")):
```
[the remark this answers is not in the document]
radoslawcybulski: Removed.

mgmt_cli_test.py
```diff
@@ -1443,7 +1443,7 @@ def _unlock_cloud_key(self) -> str | None:
         self.db_cluster.unlock_ear_key()
         return ear_key.get("keyid")
 
-    def test_prepare_backup_snapshot(self):  # pylint: disable=too-many-locals  # noqa: PLR0914
+    def test_prepare_backup_snapshot(self):  # noqa: PLR0914
```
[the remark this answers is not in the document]
soyacz: you can leave it

soyacz:
I've just looked at results tab of these tests and I wonder, if there's specific reason to not follow our naming pattern where we use `<workload> - <throughput> - latencies`  E.g. `read - 90000 - latencies` or `read - unthrottled - latencies` (latency decorator is basically used for measuring latencies, even for max aka unthrottled throughput)

If you followed this path it would be more natural for everyone who watch performance results. Current naming is very confusing (at least for me).

radoslawcybulski:
> I've just looked at results tab of these tests
This has been updated to what you requested.

> Otherwise, from our talks I understood we're close to have this PR merged, so why the PR is still a draft?
I'm waiting for other two YCSB patches to be merged.

fruch:
@radoslawcybulski please attend to the pre commit issues, it's failing CI

The unittests are failing cause of known issue

fruch:
@juliayakovlev 

Please review the trigger par, so those would run on a weekly credence, same as other performance tests

radoslawcybulski:
@fruch updated.

sdcm/ycsb_thread.py
```diff
@@ -114,6 +117,34 @@ def run(self):
...
+        self._uuid_val = uuid.uuid4()
+        self.directory_for_hdr_files = os.path.join(self.loader_set.logdir, f'hdrh-{self._uuid_val}')
+        LOGGER.debug('HDR files directory: %s', self.directory_for_hdr_files)
+        os.makedirs(self.directory_for_hdr_files, exist_ok=True)
+        self.cluster_tester = cluster_tester
+
+    def _hdr_files_directory_inside_ycsb_container(self, loader_idx, cpu_idx):
+        return f'/tmp/hdr-output-directory/{self._uuid_val}/{loader_idx}/{cpu_idx}'
+
+    def _hdr_files_directory_on_master_node(self):
```
vponomaryov, line 140: This one looks redundant

sdcm/stress/base.py
```diff
@@ -87,7 +87,7 @@ def run(self):
         for loader in self.loaders:
             for cpu_idx in range(self.stress_num):
                 self.results_futures += [self.executor.submit(self._run_stress, *(loader, loader.node_index, cpu_idx))]
-
+        LOGGER.debug("Started %d stress threads", len(self.results_futures))
```
vponomaryov, line 90: This is common code and it's statement is not correct.
Thread pool executor may have smaller `max_workers` value than the stress commands amount.
It is easy to achieve using `round_robin=true` and amount of stress commands bigger than amount of loader nodes.

So, it should be rephrased to something like following:
```
Spawned %d stress threads with execution pool workers = %d.
```

We have such test cases which fall under the situation I described:
- https://github.com/scylladb/scylla-cluster-tests/blob/master/test-cases/longevity/longevity-gce-custom-d1-workload1-hybrid-raid.yaml
- https://github.com/scylladb/scylla-cluster-tests/blob/master/test-cases/longevity/longevity-gce-custom-d1-workload2-hybrid-raid.yaml
- https://github.com/scylladb/scylla-cluster-tests/blob/master/test-cases/longevity/longevity-aws-custom-d2-workload1-3dcs.yaml
- https://github.com/scylladb/scylla-cluster-tests/blob/master/test-cases/longevity/longevity-aws-custom-d2-workload1-5dcs.yaml

sdcm/stress/base.py
```diff
@@ -96,7 +96,7 @@ def _run_stress(self, loader, loader_idx, cpu_idx):
     def get_results(self):
         results = []
         timeout = self.hard_timeout + 120
-        LOGGER.debug('Wait for %s stress threads results', self.max_workers)
+        LOGGER.debug('Wait for %s stress threads results', len(self.results_futures))
```
vponomaryov, line 99: If change it then by adding info, not replacing.

sdcm/utils/hdrhistogram.py
```diff
@@ -132,54 +133,79 @@ def build_histogram_summary(self, path: str) -> list[dict[str, dict[str, int]]]:
...
+            interval_num = 0
+            start_intervals = range(start_ts, end_ts, window_step)
+            LOGGER.debug(
+                f'Building histogram summary for {path} with tags {self.hdr_tags} and intervals {len(start_intervals)}')
+
+            max_workers = len(start_intervals)*len(self.hdr_tags)
+            max_workers = PROCESS_LIMIT if max_workers > PROCESS_LIMIT else max_workers
+
+            with ProcessPoolExecutor(max_workers=max_workers) as executor:
+                for e, start_interval in enumerate(start_intervals):
```
vponomaryov, line 180: This enumeration is not used.

sdcm/stress/base.py
```diff
@@ -87,7 +87,7 @@ def run(self):
         for loader in self.loaders:
             for cpu_idx in range(self.stress_num):
                 self.results_futures += [self.executor.submit(self._run_stress, *(loader, loader.node_index, cpu_idx))]
-
+        LOGGER.debug("Started %d stress tasks on %d threads", len(self.results_futures), self.max_workers)
```
vponomaryov, line 90: It still has incorrect statement.
We have pool of workers. So, when amount of tasks is bigger than the amount of workers, we `do not start` all of them. We start only amount of tasks equal to the `worker amount`, other are `scheduled` and will run only when some of previous finish.

So, I find this log message confusing and would understand it incorrectly debugging some failure not knowing about the pool of workers.

test-cases/performance/perf-regression-alternator-basic.yaml
```diff
@@ -0,0 +1,104 @@
...
+dynamodb_primarykey_type: HASH_AND_RANGE
+
+user_prefix: 'perf-alternator-latency'
+space_node_threshold: 644245094
+
+use_hdrhistogram: true
+alternator_stress_rate: 60000
+alternator_write_always_lwt_stress_rate: 10000
+
+round_robin: true
```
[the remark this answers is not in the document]
vponomaryov: Now each of stress commands will run on each of the loaders.
It multiplies the load by number of loaders.

Does it match the cluster size?

test-cases/performance/perf-regression-alternator-full.yaml
```diff
@@ -0,0 +1,104 @@
...
+  -p fieldcount=10 -p fieldlength=512
+
+stress_cmd_m: >-2
+  bin/ycsb run dynamodb -P workloads/workloada -threads 100 -p recordcount=15000000
+  -p readproportion=0.5 -p updateproportion=0.5 -p scanproportion=0 -p insertproportion=0 -p requestdistribution=uniform -p operationcount=0
+  -p fieldcount=10 -p fieldlength=512
+
+
+n_db_nodes: 3
+n_loaders: 9
```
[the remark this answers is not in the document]
radoslawcybulski: I got confused. I want `round-robin`, i just need more commands. I will update commands to be less redundant in some next patch.

It matches cluster size now.

sdcm/stress/base.py
```diff
@@ -96,7 +96,7 @@ def _run_stress(self, loader, loader_idx, cpu_idx):
     def get_results(self):
         results = []
         timeout = self.hard_timeout + 120
-        LOGGER.debug('Wait for %s stress threads results', self.max_workers)
+        LOGGER.debug('Wait for %d stress tasks results on %d threads', len(self.results_futures), self.max_workers)
```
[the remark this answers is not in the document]
radoslawcybulski: Updated to mention a worker pool size instead.

sdcm/stress/base.py
```diff
@@ -96,7 +96,7 @@ def _run_stress(self, loader, loader_idx, cpu_idx):
     def get_results(self):
         results = []
         timeout = self.hard_timeout + 120
-        LOGGER.debug('Wait for %s stress threads results', self.max_workers)
+        LOGGER.debug('Wait for %d stress tasks results on worker pool of size %d', len(self.results_futures), self.max_workers)
```
vponomaryov, line 99: Precommit check fails here:
```
19:53:26  diff --git a/sdcm/stress/base.py b/sdcm/stress/base.py
19:53:26  index fa86e388..b68a03ac 100644
19:53:26  --- a/sdcm/stress/base.py
19:53:26  +++ b/sdcm/stress/base.py
19:53:26  @@ -96,7 +96,8 @@ class DockerBasedStressThread:
19:53:26       def get_results(self):
19:53:26           results = []
19:53:26           timeout = self.hard_timeout + 120
19:53:26  -        LOGGER.debug('Wait for %d stress tasks results on worker pool of size %d', len(self.results_futures), self.max_workers)
19:53:26  +        LOGGER.debug('Wait for %d stress tasks results on worker pool of size %d',
19:53:26  +                     len(self.results_futures), self.max_workers)
19:53:26           for future in concurrent.futures.as_completed(self.results_futures, timeout=timeout):
19:53:26               results.append(future.result())
19:53:26   
script returned exit code 1
```

sdcm/stress/base.py
```diff
@@ -87,7 +87,7 @@ def run(self):
         for loader in self.loaders:
             for cpu_idx in range(self.stress_num):
                 self.results_futures += [self.executor.submit(self._run_stress, *(loader, loader.node_index, cpu_idx))]
-
+        LOGGER.debug("Started %d stress tasks on worker pool of size %d", len(self.results_futures), self.max_workers)
```
vponomaryov, line 90: I am sorry, but technically, it is incorrect to say `started` for all of them.
If workers pull is smaller than amount of tasks then only part of tasks is `started` another part is `scheduled`.

Example:
worker_pull = 16
stress commands = 20
started_stress_cmds = 16
scheduled_stress_cmds = 4

Those 4 in the example literally will start running only after some 4 from first 16 finish completely.

sdcm/stress/base.py
radoslawcybulski: Updated to `Started %d stress tasks on worker pool of size %d`

sdcm/stress/base.py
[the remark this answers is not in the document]
radoslawcybulski: Updated to `Wait for %d stress tasks results on worker pool of size %d`

sdcm/stress/base.py
radoslawcybulski: You're correct, changed to scheduled.

sdcm/stress/base.py
radoslawcybulski: Updated, hopefully it will pass this time. Do you have something automatic like `clang-format` that would apply those fixes?

vponomaryov:
@fruch , @juliayakovlev 
You may want to run some of existing perf tests to check that it still works like you expect.

fruch (approved):
LGTM

PR #10928, opened 2025-05-22
Add custom config option for pipelines

cezarmoise:
Add a new field to pipelines named `custom_config`.
A file named `custom_config.yaml` will be created with the field's contents and appended to the list of test configs.
This allows fast prototyping of new tests without having to go through the step of commiting a change to an existing config. 

Example usage:
Screenshot From 2025-05-22 09-26-30

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/ee0d1e57-0f1b-40ea-adb3-0fc19953a102/details

### Reminders

### DIFF
--- a/jenkins-pipelines/hackathon/hackathon-custom-config.jenkinsfile
+++ b/jenkins-pipelines/hackathon/hackathon-custom-config.jenkinsfile
@@ -1,0 +1,13 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+longevityPipeline(
+    backend: "aws",
+    region: 'eu-west-1',
+    availability_zone: 'a',
+    test_name: 'longevity_test.LongevityTest.test_custom_time',
+    test_config: 'test-cases/hackathon/hackathon-custom-config.yaml',
+    provision_type: 'on_demand',
+)
--- a/test-cases/hackathon/hackathon-custom-config.yaml
+++ b/test-cases/hackathon/hackathon-custom-config.yaml
@@ -1,0 +1,26 @@
+test_duration: 30
+stress_cmd: [
+  "cassandra-stress write cl=QUORUM duration=15m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native -rate threads=100 -pop seq=1..10000000 -log interval=5"
+]
+
+n_db_nodes: 3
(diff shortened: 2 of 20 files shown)


jenkins-pipelines/hackathon/hackathon-custom-config.jenkinsfile
cezarmoise, line 1: for testing purposes only, to be deleted before merge

test-cases/hackathon/hackathon-custom-config.yaml
cezarmoise, line 1: for testing purposes only, to be deleted before merge

fruch:
@cezarmoise 

how does it differs from using `extra_environment_variables` for testing a few extra variables ?
also how it's easier then doing git changes  and updates, that can be tracked ? i.e. if one is build a whole case.

also we it's on purpose we don't expose all of the SCT options in pipelines, to make sure we have traceability on official jobs, and avoid people from running job with untested or not according to what's in the code.

I don't think it would make anyone life easier...

pehala:
> how does it differs from using `extra_environment_variables` for testing a few extra variables ? also how it's easier then doing git changes and updates, that can be tracked ? i.e. if one is build a whole case.
> 
> also we it's on purpose we don't expose all of the SCT options in pipelines, to make sure we have traceability on official jobs, and avoid people from running job with untested or not according to what's in the code.
> 
> I don't think it would make anyone life easier...

I disagree, this is meant for prototyping and it for sure is easier than having to do a git change for every minor adjustment, `extra_environment_variables` is comparable, but I think this is more versatile. For example, I do not know how I could change `scylla_yaml` via the env vars.

fruch:
If one need to write a whole config yaml with all the option, we want it in code for tracking
if it's one of two configuration options someone need to override quickly, `extra_environment_variables` is enough.

we don't want to people to override things in official release jobs, so we loose track of what was running.

people developing should be able to use git well enough for this kind of exercise, same as in any other repository we have in scylla 

I don't see how this kind of thing is holding / slowing any development of anyone.

pehala:
> 
> The description of that pipeline parameter provides a nested example, similar to what you are refering to:
> 
> ```
> Extra environment variables to be set in the test environment, uses the java Properties File Format. \
>     Example: SCT_STRESS_IMAGE.cassandra-stress=scylladb/cassandra-stress:3.13.0 SCT_USE_MGMT=false
> ```
> 
> So, following config change:
> 
> ```
> append_scylla_yaml:                                                                                 
>   enable_tablets: false
> ```
> 
> Can be specified like following:
> 
> ```
> SCT_APPEND_SCYLLA_YAML.enable_tablets=false

If this can be already done, then we can close this as this is exactly what we need.

PR #10930, opened 2025-05-22
Default region change to us-east-1

jsmolar:
AMIs are created by default in `us-east-1`. The default region we use in Jenkins jobs is `eu-west-1`.
This PR changes the default region to match AMIs default location.

### Testing

- [ ]

### Reminders

### DIFF
--- a/vars/longevityPipeline.groovy
+++ b/vars/longevityPipeline.groovy
@@ -26,7 +26,7 @@ def call(Map pipelineParams) {
                description: 'aws|gce|azure|docker',
                name: 'backend')
 
-            string(defaultValue: "${pipelineParams.get('region', 'eu-west-1')}",
+            string(defaultValue: "${pipelineParams.get('region', 'us-east-1')}",
                description: 'Supported: us-east-1 | eu-west-1 | eu-west-2 | eu-north-1 | eu-central-1 | us-west-2 | random (randomly select region)',
                name: 'region')
             string(defaultValue: "${pipelineParams.get('gce_datacenter', 'us-east1')}",
--- a/vars/managerPipeline.groovy
+++ b/vars/managerPipeline.groovy
@@ -61,7 +61,7 @@ def call(Map pipelineParams) {
             string(defaultValue: "${pipelineParams.get('backend', 'aws')}",
                description: 'aws|gce',
                name: 'backend')
-            string(defaultValue: "${pipelineParams.get('region', 'eu-west-1')}",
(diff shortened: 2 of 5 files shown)


fruch:
@jsmolar why should we change it ? what's the problem with current situation ?

i.e. it can backfire, and leave us with not enough resources on `us-east-1`, if everyone by default would work on it.

vponomaryov:
I agree with @fruch 

At first, the problem is unclear. It doesn't matter in which region images get built at a first place, they, anyway, get copied to other supported regions.

Second, I faced problems with allocations exactly in `us-east-1` much more than in other regions.

Third, where are the majority of people who work with test VMs? - Biggest part is in Europe and Israel which are much closer to `eu` VMs and, hence, have better ping connecting to such VMs.

jsmolar:
@fruch @vponomaryov 
During the summit, I heard some developers express frustration that when using BYO AMI, the image is built by default in the us-east-1 region (as indicated by the COPY_AMI_TO_REGIONS parameter in [this Jenkins job](https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/)). With the current default region set to eu-west-1, one developer ran into an issue where their AMI was not available in the default region, forcing them to rebuild the job.

This PR was intended as a proposal to address that minor inconvenience by aligning the default region with where the AMI is initially built.

That said, we could also consider removing the default region entirely or simply accept the inconvenience and leave things as they are.

vponomaryov:
> @fruch @vponomaryov During the summit, I heard some developers express frustration that when using BYO AMI, the image is built by default in the us-east-1 region (as indicated by the COPY_AMI_TO_REGIONS parameter in [this Jenkins job](https://jenkins.scylladb.com/job/scylla-master/job/byo/job/byo_build_tests_dtest/)). With the current default region set to eu-west-1, one developer ran into an issue where their AMI was not available in the default region, forcing them to rebuild the job.
> 
> This PR was intended as a proposal to address that minor inconvenience by aligning the default region with where the AMI is initially built.
> 
> That said, we could also consider removing the default region entirely or simply accept the inconvenience and leave things as they are.

It doesn't sounds as conceptual problem.

It is known limitation that AMIs are region-based.
BYO CI jobs allow to configure copying of images to other regions as you already referred to.
Moreover, SCT has BYO feature which utilizes mentioned CI job types and resolves this problem automatically allowing you to run SCT BYO CI jobs in any region.

The switch will lead to a single region resources overcommit failing more CI jobs than it is now and I consider it a bigger problem.

PR #10938, opened 2025-05-26
feature(config): switch allocation memmory event severity

aleksbykov:
Allocation memory warning mark jobs as failed because processes as errors. Adding new config parameter which could be passed as ENV variable to change severity for such error events to warning

### Testing

- [ ]

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -308,3 +308,5 @@ mgmt_snapshots_preparer_params:
   num_of_rows: ''
   sequence_start: ''
   sequence_end: ''
+
+allocation_as_errors: true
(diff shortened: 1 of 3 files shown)


pehala:
This by itself does not solve anything as the default is still the same. For which jobs you would like to see this disabled?

fruch:
@aleksbykov 
let's just change this event type to WARNING in 2025.2

for now we can skip it, from no one master should be cleared from those until next release.

aleksbykov:
> This by itself does not solve anything as the default is still the same. For which jobs you would like to see this disabled?

The idea was to switch to warning severity by request via `extra_environment_variables` parameter in the job. Which could be easily configured in trigger.

PR #10939, opened 2025-05-26
fix(database_events): Set ALLOCATION_MEMERY AS WARNING

aleksbykov:
To remove error events noise from 2025.2 sct jobs, set allocation memory error events as noise

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/sct_events/database.py
+++ b/sdcm/sct_events/database.py
@@ -83,7 +83,7 @@ class ReactorStalledMixin(Generic[T_log_event]):
 
 
 # cause this is warning level, it's need to be before WARNING being suppressed
-DatabaseLogEvent.add_subevent_type("OVERSIZED_ALLOCATION", severity=Severity.ERROR,
+DatabaseLogEvent.add_subevent_type("OVERSIZED_ALLOCATION", severity=Severity.WARNING,
                                    regex="seastar_memory - oversized allocation:")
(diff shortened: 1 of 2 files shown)


sdcm/sct_events/database.py
```diff
@@ -83,7 +83,7 @@ def add_info(self: T_log_event, node, line: str, line_number: int) -> T_log_even
 
 
 # cause this is warning level, it's need to be before WARNING being suppressed
-DatabaseLogEvent.add_subevent_type("OVERSIZED_ALLOCATION", severity=Severity.ERROR,
+DatabaseLogEvent.add_subevent_type("OVERSIZED_ALLOCATION", severity=Severity.WARNING,
```
soyacz, line 86: I'd propose to reduce severity in `enable_default_filters` with `SkipPerIssues` referencing to issue

fruch: If it was one specific issue, yeah.

But it's an umbrella for multiple of those, it doesn't worth the back and forth for each.

pehala (approved):

sdcm/sct_events/database.py
soyacz: ok, just need to remember to take it off when they're fixed

fruch: it would take a while to chase them all in master (hopefully until 2025.3 is branched, or we'll need to repeat this one)

PR #10958, opened 2025-05-27
improvement(sct.py): Allow multiple regions in list-images

pehala:
* Changed the parameter from `region` to `regions`
   * To make the change easier
* Removed default setting of `eu-west-1` in aws branches.
   * Region is by default `eu-west-1`
* The diff is way to much complicated for what I actually did, which is just to add one for loop on top of existing code :D

### Testing

- [x] locally with `./sct.py list-images -r eu-west-1 -r eu-west-2 -v 2025.2
`


sct.py
```diff
@@ -669,73 +670,73 @@ def list_images(cloud_provider: str, branch: str, version: str, region: str, arc
...
-            branch += ":all"
-
-        match cloud_provider:
-            case "aws":
-                region = region or "eu-west-1"
-                ami_images = get_ami_images(branch=branch, region=region, arch=arch)
-                click.echo(
-                    create_pretty_table(rows=ami_images, field_names=branch_fields_with_tag_name).get_string(
-                        title=f"AMI Machine Images for {branch} in region {region}"
+    for region in regions:
```
vponomaryov, line 673: It could keep indentation as-is and be smaller change from git point of view by doing following:

```
if version is None:
    return
for region in regions:
    <all the existing code stays as-is>
```

vponomaryov (dismissed):
LGTM

timtimb0t (dismissed):

docs/configuration_options.md
```diff
@@ -147,7 +147,7 @@ scylla cloud cluster name
 
 **default:** N/A
 
-**type:** str
+**type:** str (appendable)
```
fruch, line 150: how `cloud_cluster_name` is related to this change ?

pehala: It is not, i included it because not up to date check, the issue was since fixed so I will try rebase

sct.py
```diff
@@ -646,15 +646,16 @@ def list_resources_on_azure():
               type=str,
               help="List images by version. Use '-v all' for all versions. "
                    "OSS format: <4.3> Enterprise format: <enterprise-2021.1>. Mutually exclusive with --branch.")
-@click.option('-r', '--region',
+@click.option('-r', '--region', "regions",
               type=CloudRegion(),
               help="Cloud region to query images in. Defaults to eu-west-1",
-              default="eu-west-1")
+              default=["eu-west-1"],
```
fruch, line 652: nitpick:
this default is a bit problematic, since it would fail any other backend but AWS.
we probably need a default per backend in code

and that probably why we had default in the code to begin with

pehala: Changed, to use default by the region, reuse the list from NemesisJobGenerator

PR #10959, opened 2025-05-27
Fix quoting when creating materialized views

cezarmoise:
Fixed regex in `cql_quote_if_needed`.
| | Before| After|
|--------|--------|--------|
| `cql_quote_if_needed('v')` | `'"v"'`| `'v'`|
| `cql_quote_if_needed('C0')` | `'C0'`| `'"C0"'`|
| `cql_quote_if_needed('1x')` | `'"1x"'`| `'"1x"'`|
| `cql_quote_if_needed('x1')` | `'x1'`| `'x1'`|
| `cql_quote_if_needed('x1T')` | `'"x1T"'`| `'"x1T"'`|

Refactored and fixed `create_materialized_view` to no longer rely on manually quoting column names.
Also, in nemesis, the code only worked on the assumption that the column names required quoting but the primary keys did not.

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/0e0d06c9-4d35-41f1-ab08-df5a9c414776

### Reminders

### DIFF
--- a/sdcm/utils/cql_utils.py
+++ b/sdcm/utils/cql_utils.py
@@ -23,10 +23,10 @@ def cql_quote_if_needed(identifier: str) -> str:
 
     https://cassandra.apache.org/doc/stable/cassandra/cql/definitions.html#identifiers
     """
-    identifier_regex = re.compile(r'^[^0-9][a-z0-9_]+$')
-    if not identifier_regex.match(identifier):
-        return f'"{identifier}"'
-    return identifier
+    identifier_regex = re.compile(r'^[a-z][a-z0-9_]*$')
+    if identifier_regex.match(identifier):
+        return identifier
+    return f'"{identifier}"'
 
 
 def cql_unquote_if_needed(identifier: str):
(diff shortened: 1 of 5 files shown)


sdcm/utils/cql_utils.py
```diff
@@ -23,10 +23,10 @@ def cql_quote_if_needed(identifier: str) -> str:
 
     https://cassandra.apache.org/doc/stable/cassandra/cql/definitions.html#identifiers
     """
-    identifier_regex = re.compile(r'^[^0-9][a-z0-9_]+$')
```
cezarmoise, line 26: This used to fail on column names like 'C0' (from cassandra-stress) because 'C' passed the is not a number check, then '0' passed the is lowercase or number or underscore test

fruch: that table you had in the description calls in for a unittest for this one..

cezarmoise:
> worth checking a test that runs scylla bench as main stress tool

In the test I ran, the nemesis ran 2 times, once on a scylla-bench table, once on a cassandra-stress table.

soyacz (dismissed):
LGTM

fruch (approved):

PR #10975, opened 2025-05-28
fix(scylla_yaml): Allow usage of any option in append_scylla_yaml

pehala:
Previously, the class did not allow it and it would fail on non existing field. Update testcase to verify this
* Use pydantic v2 validation
   * This could be extracted to another PR or commit if necessary

### Testing

- [x] unit testing

### DIFF
--- a/sdcm/provision/scylla_yaml/scylla_yaml.py
+++ b/sdcm/provision/scylla_yaml/scylla_yaml.py
@@ -10,23 +10,26 @@
 # See LICENSE for more details.
 #
 # Copyright (c) 2021 ScyllaDB
-from difflib import unified_diff
-from typing import List, Literal, Union, Any
-
 import logging
+from difflib import unified_diff
+from typing import Literal, List, Dict, Any, Union
+
 import yaml
-from pydantic import validator, BaseModel, ConfigDict
+from pydantic import field_validator, BaseModel, ConfigDict
 
-from sdcm.provision.scylla_yaml.auxiliaries import RequestSchedulerOptions, EndPointSnitchType, SeedProvider, \
-    ServerEncryptionOptions, ClientEncryptionOptions
-
+from sdcm.provision.scylla_yaml.auxiliaries import EndPointSnitchType, ServerEncryptionOptions, ClientEncryptionOptions, \
+    SeedProvider, RequestSchedulerOptions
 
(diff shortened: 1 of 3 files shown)


fruch (approved):
LGTM

vponomaryov:
What is the use case to do it?
What gets fixed with it?

Now we will not see whether we have typos or not.

vponomaryov:
I meant exactly typos in config option names. It is not about type hints.
Lots of Scylla config options are long and hard to remember correctly and easy to have a typo in the name.

pehala:
> I meant exactly typos in config option names. It is not about type hints. Lots of Scylla config options are long and hard to remember correctly and easy to have a typo in the name.

Well, first then there is a major disagreement on purpose of this class among maintainers and second, this class never did that particular validation well, given it has not been maintained and is only being amended. It will reject a perfectly valid option and allow outdated option which doesnt work. If we want that kind of validaiton, we need a different way how to do this

vponomaryov:
We could have a tumbler (by SCT config) to enable or disable possibility to define unsupported (for now) options.
So, both approaches would be supported from the Jenkins parameters point of view.

sdcm/provision/scylla_yaml/scylla_yaml.py
```diff
@@ -10,23 +10,26 @@
...
 
 
 class ScyllaYaml(BaseModel):
+    """
+    Model for scylla.yaml configuration.
+    Allows for any parameters, parameters here are just type hint helpers
+    """
 
-    model_config = ConfigDict(arbitrary_types_allowed=True)
+    model_config = ConfigDict(arbitrary_types_allowed=True, extra="allow")
```
vponomaryov, line 32: This extra can be condition by some new simple SCT config option.
It could be "disallow" by default.
And when one needs to test some new Scylla config option not covered by SCT yet, he would be able to enable extras and use his target Scylla config option.

fruch: It's not worth our time

if you want exact validation of the options, SCT can't be the source of truth.
It can be use as a helper for writing code in SCT, not more then that, that was the intention.

you can if you wish as a followup generate warning when there a miss-match. (which no on would be looking at)

SCT should pass-throw what user is asking, and in some cases what scylla is passing as default, i.e. new values.

PR #10977, opened 2025-05-28
improvement(precommit): Add update-conf-docs validation

pehala:
Only triggers on editing the SCTConfiguration

### Testing

- [x] Locally

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -51,6 +51,13 @@ repos:
         types: [python]
         exclude: ^argus/.*$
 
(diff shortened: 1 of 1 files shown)


pehala:
> This is not the only case when the docs must be updated. Change of the `defaults/test_default.yaml` file also makes an effect.
> 
> And probably other files in the `defaults` dir. Need to double check.

You are right, the code also loads 'defaults/test_default.yaml', so I will add match for entire `defaults` dir which should suffice

pehala:
@scylladb/qa-maintainers please review

PR #10991, opened 2025-05-29
improvements(pyproject.toml): Move dependencies to pyproject.toml

pehala:
* Simplifies setup and unifies the experience with our other tools
* Minor changes to `build_n_push`.
   * It now generates `requirements.txt` directly from `uv.lock` and deletes them afterwards
* No dependencies changed, so hydra rebuild is not necessary.

### Testing

- [x] locally

### Reminders


pyproject.toml
```diff
@@ -1,3 +1,73 @@
...
+[dependency-groups]
+dev = [
+    "autopep8==2.3.1",
+    "pre-commit==4.0.1",
+    "pycodestyle==2.12.0",
+    "ruff==0.4.7",
+]
+[tool.uv]
+package = false
+# gimme-aws-creds does not work on Windows for Python 3.9+
```
fruch, line 68: Why do we need to have any windows related support ?

pyproject.toml
```diff
@@ -1,3 +1,73 @@
...
+    "pyzmq==26.2.0",
+    "questionary==2.1.0",
+    "requests==2.32.2",
+    "scylla-driver==3.29.3",
+    "sortedcontainers==2.4.0",
+    "ssh2-python3==1.11.1",
+    "tcconfig==0.29.1",
+    "tenacity==9.0.0",
+]
+[dependency-groups]
```
fruch, line 59: Maybe we should use extras ? so it can be used with plain pip as well ?

pehala:
> I think dockerfile needs to change and use `uv sync`
> 
> Also we should update docs with the needed commands to use now

`uv sync` is designed to work with venv, so we could either trick it to use system python or we would switch to venv in container

Regarding docs, I will add it

pyproject.toml
pehala: I believe pip supports this as well: https://ichard26.github.io/blog/2025/04/whats-new-in-pip-25.1/#key-features-

fruch:
> > I think dockerfile needs to change and use `uv sync`
> > 
> > Also we should update docs with the needed commands to use now
> 
> `uv sync` is designed to work with venv, so we could either trick it to use system python or we would switch to venv in container
> 
> Regarding docs, I will add it

See how I did it in dtest, you can force it to work with system.

venv inside docker, might complicate some of the usages

pehala:
v2:
  * Use `uv sync` in Dockerfile
  * Move hydra `Dockerfile` into root
     * Simplifies `build_n_push` and also make it easier to find
  * Add install readme footnote
  * Rebase

pyproject.toml
```diff
@@ -1,3 +1,73 @@
...
+    "invoke==2.2.0",
+    "jinja2==3.1.6",
+    "kubernetes==31.0.0",
+    "ldap3==2.9.1",
+    "moto[server]==5.1.1",
+    "mysql-connector-python==9.1.0",
+    "packaging>=24",
+    "paramiko==3.5.0",
+    "parameterized==0.8.1",
+    "pbr==6.1.0",
```
vponomaryov, line 39: When a python project really uses the `pbr` package then this package should be the first one in the list of a python project dependencies.

But, as I can see, its usage is not enabled because we would need something like following in this file:

```
[build-system]
requires = [
    "pbr>=6.1.0",
    "setuptools>=76.1.0"
]
build-backend = "pbr.build"
```

So, we should either remove the `pbr` dependency as unused or start really using it.

vponomaryov (changes requested):
Added inline comment.

Also, with this change the `tox.ini` file becomes broken.

Since it is not really used by anyone, need just to delete it.

pyproject.toml
pehala: Nice catch, I have no idea about why we wanted pbr here, so I will remove it

pehala:
v3:
  * Remove pbr dependency
  * Update docker build action
     * It seems that it is running the version that is in master and not in this commit, it still install old uv

pehala:
> @pehala docker image build fails even after job rerun.

It does, because it is not using the job from this PR but from master. The job is using `pull_request_target`, which uses base of the job for the run:
`This event runs in the context of the base of the pull request, rather than in the context of the merge commit`(https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request_target)

I will create my own commit with the updated hydra

fruch:
> @scylladb/qa-maintainers I sadly do not have permissions for pushing the image and due to the usage of "pull_request_target", it is using old pipelines and thus the automatic image build wont work here, could you please assist with this?

I've add a docker image, and also changed renovate config to match

soyacz (approved):
one small remark, otherwise LGTM

fruch (dismissed):
LGTM

And pass the provision tests

I still need to rebuild

pehala:
More and more I am convinced that this will cause backport pain especially for dependency changes, should I backport this to all active branches?

@scylladb/qa-maintainers WDYT?

pehala:
> where did you faced a problem ?
> 
> this is not a trivial change to backport, and you will need to rebuild the image in all of the branches you backport to
> 
> I would recommend doing so, only if there a very strong need, no if there on dependency update.

We are changing a way how we do dependencies. If we ever want to change dependencies in the future, which might happen on 2025.1 as it is an LTS release, its gonna be painful. I think major changes should be merged into LTS versions to keep them up to date as we will have to support them for a long time. The changes might not be needed now, but then in a years time we might have a change that we want to backport but it will rely on many previous patches/reworks.

I am facing this problem currently with the ruff changes and all the patches that were missing on 2024.1, this particular one is not that important as it is only formatting, but it showcases the problem.

fruch:
> > where did you faced a problem ?
> > this is not a trivial change to backport, and you will need to rebuild the image in all of the branches you backport to
> > I would recommend doing so, only if there a very strong need, no if there on dependency update.
> 
> We are changing a way how we do dependencies. If we ever want to change dependencies in the future, which might happen on 2025.1 as it is an LTS release, its gonna be painful. I think major changes should be merged into LTS versions to keep them up to date as we will have to support them for a long time. The changes might not be needed now, but then in a years time we might have a change that we want to backport but it will rely on many previous patches/reworks.
> 
> I am facing this problem currently with the ruff changes and all the patches that were missing on 2024.1, this particular one is not that important as it is only formatting, but it showcases the problem.

we can do it for some change, but not for every big change.
the default is still like scylla core, new features are not going into release branches,
this one is a new feature

support for changing release can still be done in the old way, as it was done years before.

the exception is emergency changes which are a must across all releases.

you are pushing this for a case of backporting a cosmetic change... i.e. formatting.


backporting this specific one to all releases is problematic, and gonna be hard, since it can't be backported as is, each release has it's own specific list of packages with their specific version, jump forward without the needed fixes won't work.

so for backporting this you'll still need to craft a specific list of package for each branch, and build it's specific docker image.

so it doesn't make your life easier, backporting this would just complicate things

PR #11010, opened 2025-06-03
test(collect_diagnosis_data_test): Add DiagnosisReport teardown validator

Lakshmipathi:
Adds a new DiagnosisReport teardown validator to send signal to scylla and dump diagnostics then verify it.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/9443

### Testing

- [ ]

### Reminders


sdcm/cluster.py
```diff
@@ -2606,17 +2606,28 @@ def prepare_files_for_archive(self, fileslist):
...
-        self.remoter.sudo("pkill -f --signal 3 /usr/bin/scylla")
-        self.wait_db_down(timeout=600)
-        if restart_scylla:
-            self.log.debug('Restart scylla server')
-            self.stop_scylla(timeout=600)
-            self.start_scylla(timeout=600)
+        if generate_diagnosis:
+            InfoEvent(severity=Severity.DEBUG, message=message)
+            with DbNodeLogger([self], "send SIGQUIT to scylla", target_node=self, additional_info="diagnosis dump" if generate_diagnosis else "coredump"):
+                self.remoter.sudo("pkill -3 scylla")
```
soyacz, line 2616: `pkill -f --signal 3 /usr/bin/scylla` used below (formerly) doesn't work? Maybe we should use the same for both (can be new one if it works better)

Besides I don't think we should differentiate both cases - we should always return cluster to base state (all nodes runnig) - there could be other validators that require scylla to be up.

sdcm/cluster.py
```diff
@@ -3215,6 +3226,10 @@ def query_metadata(self, url: str, headers: dict = None, token_url: str = None,
     def systemd_version(self) -> int:
         return get_systemd_version(self.remoter.run("systemctl --version", ignore_status=True).stdout)
 
+    @property
+    def is_diagnostics_logged(self):
```
soyacz, line 3230: I believe this diagnostics dump could be useful - shouldn't we collect it with coredump thread (can be part of separate PR)

sdcm/teardown_validators/diagnosis_report.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+    validator_name = 'diagnosis_report'
+
+    def validate(self):
+        LOGGER.info("Starting diagnosis report validation")
+        test_node = self.tester.db_cluster.nodes[0]
+        try:
+            test_node.generate_coredump_file(restart_scylla=True, generate_diagnosis=True)
+            LOGGER.info("Diagnosis dump report validation passed")
+            self.tester.get_test_status = lambda: 'SUCCESS'
```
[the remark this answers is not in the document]
soyacz: I see we generate error event to mark errors (this anyway will cause test to fail): `ValidatorEvent`

sdcm/cluster.py
```diff
@@ -2606,17 +2606,28 @@ def prepare_files_for_archive(self, fileslist):
...
-            self.log.debug('Restart scylla server')
-            self.stop_scylla(timeout=600)
-            self.start_scylla(timeout=600)
+        if generate_diagnosis:
+            InfoEvent(severity=Severity.DEBUG, message=message)
+            with DbNodeLogger([self], "send SIGQUIT to scylla", target_node=self, additional_info="diagnosis dump" if generate_diagnosis else "coredump"):
+                self.remoter.sudo("pkill -3 scylla")
+                time.sleep(30)
+                if not self.is_diagnostics_logged:
+                    raise Exception("Diagnosis dump report not found")
```
soyacz, line 2619: instead of raising it should return e.g. path to diagnosis dump and validate if it exists (or even collect it for further analysis) and raise error event when it's none in validator.

PR #11026, opened 2025-06-04
fix(rack_aware): activate only if more then one rack in a region

fruch:
by mistake the logic would assume that every multi-dc case has multiple racks, it was counting racks across all regions/datacenters.

this fix introduce a new helper function reshuffle the data into mapping per datacenter, and check if we have more then one rack in any of the datacenters

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3429,6 +3429,13 @@ class BaseCluster:
 
         return rack_names_mapping
 
+    @staticmethod
+    def get_rack_names_per_datacenter_from_rack_mapping(self, rack_mapping) -> dict[str, list[str]]:
+        by_region_rack_names = defaultdict(list)
+        for (region, _), v in rack_mapping.items():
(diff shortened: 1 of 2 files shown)


fruch:
I've seen this in one multi-dc use case, and couldn't understand why rackaware was turned on in c-s.

sdcm/cluster.py
```diff
@@ -3429,6 +3429,13 @@ def get_rack_names_per_datacenter_and_rack_idx(self, db_nodes: list[BaseNode] |
 
         return rack_names_mapping
 
+    @staticmethod
+    def get_rack_names_per_datacenter_from_rack_mapping(self, rack_mapping) -> dict[str, list[str]]:
```
pehala, line 3433: Why do you have `self` here? It should either be static method without self or a normal method with self, right?

pehala: Thats exactly how I encountered it, I updated Ruff (PR incoming) and this was one of the checks added by default

Issue #10932, opened 2025-05-22
Add support parallel decommission


cezarmoise:
With parallel decommission, `verify_decommission` uses a random `verification_node` to check a node's decommision status, but it may itself be decommissioning.

Suggestion:
Modify `BaseScyllaCluster.decommission` and `BaseScyllaCluster.verify_decommission` to take either a node or a list of nodes.

soyacz:
Note: currently we have it implemented as part of `Nemesis.decommission_nodes` method.

cezarmoise:
> Note: currently we have it implemented as part of `Nemesis.decommission_nodes` method.

Yes. That still works.
The main issue is this code in `BaseScyllaCluster.verify_decommission`
```python
undecommission_nodes = [n for n in self.nodes if n != node]
verification_node = random.choice(undecommission_nodes)
```
It can happen that the verification node is one of the other decommissioned nodes and an error like this happens
```

```

soyacz:
> > Note: currently we have it implemented as part of `Nemesis.decommission_nodes` method.
> 
> Yes. That still works. The main issue is this code in `BaseScyllaCluster.verify_decommission`
> 
> undecommission_nodes = [n for n in self.nodes if n != node]
> verification_node = random.choice(undecommission_nodes)
> It can happen that the verification node is one of the other decommissioned nodes and an error like this happens because that function only knows of 1 node being decommissioned
> 
in that case we should take node that is not being a target node (all decomissioned nodes must be marked as target) - we should have such code somewhere

PR #11031, opened 2025-06-04
improvement(cluster): mark nodes that are decommissioning 

cezarmoise:
With upcomming concurent topology operations feature it is possible that during `verify_decommission` a node is chosen that is itself almost done decommissioning.
This change simply marks nodes that are in a decommissioning state so that they can be ignored where needed.

closes: #10932
refs: https://github.com/scylladb/scylladb/pull/24129

Fixes errors like
https://argus.scylladb.com/tests/scylla-cluster-tests/174be212-b9ac-4fc7-ae20-8b659bfb53ec/events
```
...
File "/home/ubuntu/scylla-cluster-tests/longevity_scaling_test.py", line 125, in _decommission
self.db_cluster.decommission(node)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 5136, in decommission
self.verify_decommission(node)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 5090, in verify_decommission
missing_host_ids = verification_node.raft.search_inconsistent_host_ids()
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/raft/__init__.py", line 390, in search_inconsistent_host_ids
with self._node.parent_cluster.cql_connection_patient_exclusive(node=self._node) as session:
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 72, in inner
return func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 3706, in cql_connection_patient_exclusive
return self.cql_connection_exclusive(**kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 3671, in cql_connection_exclusive
return self._create_session(node=node, keyspace=keyspace, user=user, password=password, compression=compression,
File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 3624, in _create_session
session = cluster_driver.connect()
File "cassandra/cluster.py", line 1823, in cassandra.cluster.Cluster.connect
File "cassandra/cluster.py", line 1849, in cassandra.cluster.Cluster.connect
File "cassandra/cluster.py", line 1836, in cassandra.cluster.Cluster.connect
File "cassandra/cluster.py", line 3673, in cassandra.cluster.ControlConnection.connect
File "cassandra/cluster.py", line 3734, in cassandra.cluster.ControlConnection._reconnect_internal
cassandra.cluster.NoHostAvailable: ('Unable to connect to any servers', {'10.4.1.102:9042': ConnectionRefusedError(111, "Tried connecting to [('10.4.1.102', 9042)]. Last error: Connection refused")})
```

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/d84e444a-e64f-4071-b63f-24168f6eaa3d/resources

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -294,6 +294,7 @@ class BaseNode(AutoSshContainerMixin):
         self.lock = threading.Lock()
 
         self.running_nemesis = None
+        self.decommissioning = False
 
         # We should disable bootstrap when we create nodes to establish the cluster,
         # if we want to add more nodes when the cluster already exists, then we should
@@ -5093,7 +5094,7 @@ class BaseScyllaCluster:
                 return None
 
         target_node_ip = node.ip_address
-        undecommission_nodes = [n for n in self.nodes if n != node]
+        undecommission_nodes = [n for n in self.nodes if n != node and not n.decommissioning]
 
         verification_node = random.choice(undecommission_nodes)
         node_ip_list = get_node_ip_list(verification_node)
@@ -5141,13 +5142,22 @@ class BaseScyllaCluster:
         self.test_config.tester_obj().monitors.reconfigure_scylla_monitoring()
 
     def decommission(self, node: BaseNode, timeout: int | float = None) -> DataCenterTopologyRfControl | None:
+        @contextlib.contextmanager
+        def mark_decommissioning(node: BaseNode):
+            node.decommissioning = True
+            try:
+                yield
+            finally:
+                node.decommissioning = False
+
         if not node._is_zero_token_node:
             if tablets_enabled := is_tablets_feature_enabled(node):
                 dc_topology_rf_change = DataCenterTopologyRfControl(target_node=node)
                 dc_topology_rf_change.decrease_keyspaces_rf()
-        with adaptive_timeout(operation=Operations.DECOMMISSION, node=node):
-            node.run_nodetool("decommission", timeout=timeout, long_running=True, retry=0)
-        self.verify_decommission(node)
+        with mark_decommissioning(node):
+            with adaptive_timeout(operation=Operations.DECOMMISSION, node=node):
+                node.run_nodetool("decommission", timeout=timeout, long_running=True, retry=0)
+            self.verify_decommission(node)
         if node._is_zero_token_node:
             return None
         return dc_topology_rf_change if tablets_enabled else None


sdcm/cluster.py
```diff
@@ -5093,7 +5094,7 @@ def get_node_ip_list(verification_node):
                 return None
 
         target_node_ip = node.ip_address
-        undecommission_nodes = [n for n in self.nodes if n != node]
+        undecommission_nodes = [n for n in self.nodes if n != node and not n.decommissioning]
```
soyacz, line 5097: how about using `self.running_nemesis` attr for that - it's quite common pattern

sdcm/cluster.py
```diff
@@ -5146,8 +5146,10 @@ def decommission(self, node: BaseNode, timeout: int | float = None) -> DataCente
                 dc_topology_rf_change = DataCenterTopologyRfControl(target_node=node)
                 dc_topology_rf_change.decrease_keyspaces_rf()
         with adaptive_timeout(operation=Operations.DECOMMISSION, node=node):
+            node.running_nemesis = "decommissioning"
```
fruch, line 5149: We should do that like this there are utility functions for that, to make sure we hold the lock for this operation.

sdcm/cluster.py
```diff
@@ -5146,8 +5146,10 @@ def decommission(self, node: BaseNode, timeout: int | float = None) -> DataCente
...
             node.run_nodetool("decommission", timeout=timeout, long_running=True, retry=0)
         self.verify_decommission(node)
+        node.running_nemesis = None
```
fruch, line 5152: Same here.

@dimakr FYI, things to consider while revisiting the whole nemesis target selection logic

sdcm/cluster.py
cezarmoise: Fixed https://github.com/scylladb/scylla-cluster-tests/commit/29641f40af456485fc70c64727e3cccc2db8ab98

sdcm/cluster.py
cezarmoise: Fixed https://github.com/scylladb/scylla-cluster-tests/commit/29641f40af456485fc70c64727e3cccc2db8ab98

cezarmoise:
If I try to import `NEMESIS_TARGET_SELECTION_LOCK` it creates a circular import error
Trying to fix it by importing in the function where it is used https://github.com/scylladb/scylla-cluster-tests/commit/aa6713e72926b1bc261a2aed047a904a0db4c628

I will also try to run a test with GrowShrinkClusterNemesis on a version of scylla that has parallel decommission

cezarmoise:
cosmetic changes: https://github.com/scylladb/scylla-cluster-tests/commit/6048b45253d55a94f70abb73f89ded558a6f9945
```diff
index 1ccb69fcf..ec4806428 100644
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5148,11 +5148,13 @@ class BaseScyllaCluster:
                 dc_topology_rf_change.decrease_keyspaces_rf()
         with adaptive_timeout(operation=Operations.DECOMMISSION, node=node):
             with NEMESIS_TARGET_SELECTION_LOCK:
-                node.running_nemesis = "decommissioning"
+                if not node.running_nemesis:
+                    node.running_nemesis = "GracefulShutdown"
             node.run_nodetool("decommission", timeout=timeout, long_running=True, retry=0)
         self.verify_decommission(node)
         with NEMESIS_TARGET_SELECTION_LOCK:
-            node.running_nemesis = None
+            if node.running_nemesis == "GracefulShutdown":
+                node.running_nemesis = None
         if node._is_zero_token_node:
             return None
         return dc_topology_rf_change if tablets_enabled else None
```
- don't overwrite the existing nemesis if it exists
- change to `GracefulShutdown` so it remains reported the same in Argus

fruch:
I've lost the fact this code is not in nemesis.py at all

1) if some nemesis code is giving instruction to decommission, its responsible for marking this node under nemesis
2) The selection of the validation node can use the context manager for that, if it needs to pick a node that isn't currently targeted, we shouldn't clone this logic again and again

cezarmoise:
Going back to just having a `decommissiong` flag on the node.
```diff
diff --git a/sdcm/cluster.py b/sdcm/cluster.py
index ad7b69b07..91be0465f 100644
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -294,6 +294,7 @@ class BaseNode(AutoSshContainerMixin):
         self.lock = threading.Lock()
 
         self.running_nemesis = None
+        self.decommissioning = False
 
         # We should disable bootstrap when we create nodes to establish the cluster,
         # if we want to add more nodes when the cluster already exists, then we should
@@ -5093,7 +5094,7 @@ class BaseScyllaCluster:
                 return None
 
         target_node_ip = node.ip_address
-        undecommission_nodes = [n for n in self.nodes if n != node]
+        undecommission_nodes = [n for n in self.nodes if n != node and not n.decommissioning]
 
         verification_node = random.choice(undecommission_nodes)
         node_ip_list = get_node_ip_list(verification_node)
@@ -5141,13 +5142,22 @@ class BaseScyllaCluster:
         self.test_config.tester_obj().monitors.reconfigure_scylla_monitoring()
 
     def decommission(self, node: BaseNode, timeout: int | float = None) -> DataCenterTopologyRfControl | None:
+        @contextlib.contextmanager
+        def mark_decommissioning(node: BaseNode):
+            node.decommissioning = True
+            try:
+                yield
+            finally:
+                node.decommissioning = False
+
         if not node._is_zero_token_node:
             if tablets_enabled := is_tablets_feature_enabled(node):
                 dc_topology_rf_change = DataCenterTopologyRfControl(target_node=node)
                 dc_topology_rf_change.decrease_keyspaces_rf()
-        with adaptive_timeout(operation=Operations.DECOMMISSION, node=node):
-            node.run_nodetool("decommission", timeout=timeout, long_running=True, retry=0)
-        self.verify_decommission(node)
+        with mark_decommissioning(node):
+            with adaptive_timeout(operation=Operations.DECOMMISSION, node=node):
+                node.run_nodetool("decommission", timeout=timeout, long_running=True, retry=0)
+            self.verify_decommission(node)
         if node._is_zero_token_node:
             return None
         return dc_topology_rf_change if tablets_enabled else None    
```

soyacz:
> I've lost the fact this code is not in nemesis.py at all
> 
> 1. if some nemesis code is giving instruction to decommission, its responsible for marking this node under nemesis

This is why there's no need to import anything from nemesis - just using the fact that node is running nemesis.

fruch:
@cezarmoise 

1) selection of verification nodes should be like:
```
            with self.run_nemesis(node_list=self.cluster.nodes, nemesis_label="search coordinator") as verification_node:
                coordinator_node = get_topology_coordinator_node(verification_node)
```

2) `cluster.decommission()` and `cluster.verify_decommission()` shouldn't be picking the verification node on their own, the nemesis code calling them should pass it down to them (i.e. solve the import issue if there are such)

during longevity picking a random node and working with, is a recipe to problems.
we shouldn't have any code randomly selecting nodes, with going via nemesis selection logic

pehala:
@cezarmoise Moving to draft due to inactivity and active change requests

Issue #5198, opened 2022-08-29
Azure 3h longevity - NIC NotFound


fruch:
## Installation details

Cluster size: 6 nodes (Standard_L8s_v3)

Scylla Nodes used in this run:
**No resources left at the end of the run**

OS / Image: `/subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/scylla-images/providers/Microsoft.Compute/images/scylla-5.2.0-dev-x86_64-2022-08-29T11-23-01Z` (azure: eastus)

Test: `longevity-10gb-3h-azure-test`
Test id: `e4aa60b7-3636-4cb0-a9db-17e76e8497fc`
Test name: `scylla-master/longevity/longevity-10gb-3h-azure-test`
Test config file(s):

- [longevity-10gb-3h.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/339459cdb463eb4f430807139d2bc9cd11c32be5/test-cases/longevity/longevity-10gb-3h.yaml)

## Issue description

during test setup couldn't found the NIC
```
2022-08-29 12:28:10.907: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=300eafb8-afb4-460c-a0ed-8644e453bcb3, source=LongevityTest.SetUp()
exception=(NotFound) Resource /subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/SCT-e4aa60b7-3636-4cb0-a9db-17e76e8497fc-eastus/providers/Microsoft.Network/networkInterfaces/longevity-10gb-3h-master-db-node-eastus-5-nic not found.
Code: NotFound
Message: Resource /subscriptions/6c268694-47ab-43ab-b306-3c5514bc4112/resourceGroups/SCT-e4aa60b7-3636-4cb0-a9db-17e76e8497fc-eastus/providers/Microsoft.Network/networkInterfaces/longevity-10gb-3h-master-db-node-eastus-5-nic not found.
```

- Restore Monitor Stack command: `$ hydra investigate show-monitor e4aa60b7-3636-4cb0-a9db-17e76e8497fc`
- Restore monitor on AWS instance using [Jenkins job](https://jenkins.scylladb.com/view/QA/job/QA-tools/job/hydra-show-monitor/parambuild/?test_id=e4aa60b7-3636-4cb0-a9db-17e76e8497fc)
[...]

PR #11034, opened 2025-06-04
fix(azure): improved handling of spot instances provision

soyacz:
When provisioning spot instances it may happen that instance is preempted during creation.
In that case it destroys all related resources like ip's, nic's etc. causing cached values to be invalid. 
Probably it also breaks fallback to `on_demand` due the same.

fix is by detecting preemption errors and recreate cache. 
Used retrying from SCT's utils to allow retry on provision error and skip when there's problem with preemption straight to fallback.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/5198

### Testing

- [x] - labelled with azure provision test 

### Reminders

### DIFF
--- a/sdcm/sct_provision/instances_provider.py
+++ b/sdcm/sct_provision/instances_provider.py
@@ -14,20 +14,20 @@
 import logging
 from typing import List, Any
 
-from tenacity import retry, stop_after_attempt
 
 from sdcm.provision import provisioner_factory
 from sdcm.provision.helpers.cloud_init import wait_cloud_init_completes
-from sdcm.provision.provisioner import PricingModel, VmInstance, ProvisionError, Provisioner, InstanceDefinition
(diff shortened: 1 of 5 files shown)


sdcm/sct_provision/instances_provider.py
```diff
@@ -38,7 +38,7 @@ def provision_instances_with_fallback(provisioner: Provisioner, definitions: Lis
                                       ) -> List[VmInstance]:
     try:
         provision_with_retry(provisioner, definitions=definitions, pricing_model=pricing_model)
-    except ProvisionError:
+    except OperationPreemptedError:
```
dimakr, line 41: Similarly - is it ok that we will fallback to on_demand only for OperationPreemptedError now?

fruch (approved):
LGTM

Issue #11038, opened 2025-06-05
scylla debug info files are not found on DB nodes on a cluster with fips enabled


dimakr:
## Packages

Scylla version: `2025.3.0~dev-20250604.8b387109fcb2` with build-id `19a79614c2a5ca02d126259b9305ac8c8afe7bfa`
Kernel Version: `5.4.0-1021-aws-fips`

## Issue description

Backtraces are not decoded during `longevity-100gb-4h-fips-test` test runs, e.g.:
```
2025-06-05 01:34:52.079 <2025-06-05 01:34:52.001>: (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=e5618b9a-f9bb-48b8-a883-1aeed49e0910: type=OVERSIZED_ALLOCATION regex=seastar_memory - oversized allocation: line_number=63677 node=longevity-fips-master-db-node-28954425-1
2025-06-05T01:34:52.001+00:00 longevity-fips-master-db-node-28954425-1  !WARNING | scylla[23914]:  [shard  2:strm] seastar_memory - oversized allocation: 212992 bytes. This is non-fatal, but could lead to latency and/or fragmentation issues. Please report: at 0x23cab1e 0x23ca5b1 0x23ca588 0x23c99cc 0x175e11b 0x175df13 0x43ebca7 0x43ee403 0x43eb58a 0x1736ff7 0x3008457 0x4eddbc5 0x4e82b9a /opt/scylladb/libreloc/libc.so.6+0x70fa7 /opt/scylladb/libreloc/libc.so.6+0xf4fcb
0x23cab1e
0x23ca5b1
0x23ca588
0x23c99cc
0x175e11b
0x175df13
0x43ebca7
0x43ee403
0x43eb58a
0x1736ff7
0x3008457
0x4eddbc5
0x4e82b9a
/opt/scylladb/libreloc/libc.so.6+0x70fa7
/opt/scylladb/libreloc/libc.so.6+0xf4fcb
```

As per sct logs the scylla.debug files are not found on nodes:
```
< t:2025-06-05 01:05:24,010 f:remote_base.py  l:559  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.4.3.56>: Running command "[[ -f /usr/lib/debug/bin/scylla.debug ]]"...
[...]

PR #11039, opened 2025-06-05
fix(backtrace_decodeing): install correct debug symbols package

fruch:
cause of recent fix to packaging, we now have dummy package prefixed as `scylla-enterprise` and the code of `node.scylla_pkg()` return scylla-enterprise, using this name for the debug symbol package doesn't work.

this change is hardcoding the name for source available versions

Fixes: #11038

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2170,10 +2170,16 @@ class BaseNode(AutoSshContainerMixin):
             f"curl -sSf get.scylladb.com/server | sudo bash -s -- --scylla-version {version}")
 
     def install_scylla_debuginfo(self) -> None:
-        if self.distro.is_rhel_like or self.distro.is_sles:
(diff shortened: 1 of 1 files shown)


dimakr (approved):
LGTM

sdcm/cluster.py
```diff
@@ -2170,10 +2170,16 @@ def web_install_scylla(self, scylla_version: Optional[str] = None) -> None:
             f"curl -sSf get.scylladb.com/server | sudo bash -s -- --scylla-version {version}")
 
     def install_scylla_debuginfo(self) -> None:
+        if ComparableScyllaVersion(self.scylla_version) > '2025.1.0~dev':
```
vponomaryov, line 2173: If to be really correct, then it should be `bigger or equal` (`>=`).
But practically, who is going to run dev version of the 2025.1?

vponomaryov (approved):
LGTM

PR #11067, opened 2025-06-09
fix(nemesis): fix shuffle list of disruptions

soyacz:
`shuffle_list_of_disruptions` were based on `shuffle` side effect, but was getting wrong list (after multiplication). Effectively making this method not working.

Fix by not relying on side effect, rather returning reshuffled list.

### Testing

- [X] - tested locally

### Reminders


soyacz:
see this fix will reshuffle nemesis lists on 2025.2 branch

pehala (approved):
Good catch, I went though this code many times and never noticed that :) Also returning the set is much cleaner

fruch (approved):
LGTM

Issue #11073, opened 2025-06-09
ValueError: Unexpected value of n_monitor_nodes parameter during Provision Resources step


timtimb0t:
During Provision Resources step the following error occurred:

```
05:07:51  Created instances: [ec2.Instance(id='i-0c10e672a7d5eca43'), ec2.Instance(id='i-09a299469cbcffefe'), ec2.Instance(id='i-0ef1dcbca97b53d31'), ec2.Instance(id='i-0fa1e573f9f2ed293'), ec2.Instance(id='i-0958d92070aceb56a'), ec2.Instance(id='i-03203b41839fe789d')].
05:07:53  ProvisionPlan: Instances has been provisioned using "On Demand":
05:07:53  [ec2.Instance(id='i-0c10e672a7d5eca43'), ec2.Instance(id='i-09a299469cbcffefe'), ec2.Instance(id='i-0ef1dcbca97b53d31'), ec2.Instance(id='i-0fa1e573f9f2ed293'), ec2.Instance(id='i-0958d92070aceb56a'), ec2.Instance(id='i-03203b41839fe789d')]
05:07:53  Traceback (most recent call last):
05:07:53    File "/home/ubuntu/scylla-cluster-tests/./sct.py", line 1841, in <module>
05:07:53      cli.main(prog_name="hydra")
05:07:53    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1055, in main
05:07:53      rv = self.invoke(ctx)
05:07:53    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
05:07:53      return _process_result(sub_ctx.command.invoke(sub_ctx))
05:07:53    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
05:07:53      return ctx.invoke(self.callback, **ctx.params)
05:07:53    File "/usr/local/lib/python3.10/site-packages/click/core.py", line 760, in invoke
05:07:53      return __callback(*args, **kwargs)
05:07:53    File "/home/ubuntu/scylla-cluster-tests/./sct.py", line 235, in provision_resources
05:07:53      layout.provision()
05:07:53    File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_provision/aws/layout.py", line 36, in provision
05:07:53      self.monitoring_cluster.provision()
[...]

PR #11078, opened 2025-06-09
fix(test): fix missing monitor nodes

soyacz:
When updating configs recently we dropped `n_monitor_nodes` config from some multidc scenarios.

fix by adding it back.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11073

### Testing

- [ ]

### Reminders


fruch:
Isn't that the default ? we have or 1 or 0 1 should be default

fruch (approved):
LGTM

soyacz:
> Isn't that the default ? we have or 1 or 0 1 should be default

apparently it's not default. If added, need to adjust these without monitor (artifact tests I suppose)

PR #11094, opened 2025-06-10
improvement(config): use 1 monitor node by default

soyacz:
Usually we want monitor node to be present in all tests (except artifact ones).

Made `n_monitor_nodes: 1` as default and removed from configs where it was set to this.

follow up from: https://github.com/scylladb/scylla-cluster-tests/pull/11078
### Testing

- [x] - I verified if artifact tests have it set to 0 and checked all files that set `n_db_nodes: and not n_monitor_nodes:` to find if we didn't expect it in some test.

### Reminders


pehala (approved):

roydahan (approved):
Good one :)

Issue #11062, opened 2025-06-09
AddDropColumn fetches tables info inefficiently


soyacz:
AddDropColumn nemesis fetches info for each table one by one using cqlsh.
This is inefficient approach and in tests with many tables (e.g. 5k tables test) this nemesis occupies whole test duration before getting to adding/dropping columns.
This nemesis could be improved by get all tables info in one shot like we do for `schema_with_internals.log` which takes ~6 seconds (uses query: `desc schema with internals`)

PR #11112, opened 2025-06-10
improvement(add_drop_column): fetch tables data more efficiently

soyacz:
When there's a test with many tables (e.g. 5k tables test), add drop column nemesis stuck because fetching table info is done one by one.

Improvement by querying tables info with `DESC SCHEMA WITH INTERNALS` statement, which responds with all keyspaces and tables.

closes: https://github.com/scylladb/scylla-cluster-tests/issues/11062

### Testing

- [X] - https://argus.scylladb.com/tests/scylla-cluster-tests/17fa3d33-c611-4a30-a1c6-86607c335659 - AddDropColumn nemesis passed step of getting tables in ~6 seconds

### Reminders

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -99,6 +99,8 @@ DEFAULT_AWS_REGION = "eu-west-1"
 DOCKER_CGROUP_RE = re.compile("/docker/([0-9a-f]+)")
 SCYLLA_AMI_OWNER_ID_LIST = ["797456418907", "158855661827"]
 SCYLLA_GCE_IMAGES_PROJECT = "scylla-images"
+CREATE_TABLE_REGEX = re.compile(
+    r'CREATE\s+TABLE\s+(?P<keyspace>[^\s.]+)\.(?P<table>[^\s(]+)\s*\([^)]+\)(?P<options>[^;]*)')
 
 
 class KeyBasedLock():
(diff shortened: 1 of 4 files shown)


fruch:
we should have asked for checking such a thing in https://github.com/scylladb/scylla-cluster-tests/pull/7565
we need to think how to catch the next time we "fix" some flow that can blow up on lots of tables cases... 

- maybe microbenchmarks forsome of those utility functions 
- other way to raise it to reviewer/developers attention ?

fruch:
@soyacz 
test had failed creating the argus run... 
you probably want to rerun

sdcm/utils/common.py
```diff
@@ -1599,28 +1599,44 @@ def get_ami_tags(ami_id, region_name):
...
+    except (UnexpectedExit, Libssh2_UnexpectedExit) as err:
+        LOGGER.error("Failed to describe schema: %s", err.result.stderr)
+        return output
+
+    # Split schema output into elements by double newlines
+    elements = [e for e in re.split(r'\n\n+', schema_output) if e.strip()]
+
+    for element in elements:
+        # Skip non-table elements
+        if not element.strip().startswith('CREATE TABLE'):
```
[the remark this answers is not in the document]
pehala: so we are stripping it just to get boolean of out it, i.e. if it is not only empty space and then we are running the strip again in the loop to actually strip it? It just seems like the same operation done twice, why cannot we use the stripped version in the list comprehension directly and only once?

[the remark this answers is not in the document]
soyacz: anyway, removed by new approach with regex

unit_tests/test_common_get_db_tables.py
```diff
@@ -0,0 +1,119 @@
...
+
+def test_get_db_tables_without_storage_filter(mock_node):
+    """Test get_db_tables returns all tables when no storage filter is applied."""
+    tables = get_db_tables(
+        keyspace_name="keyspace1",
+        node=mock_node,
+        with_compact_storage=None
+    )
+
+    expected_tables = ["standard1"]
```
pehala, line 76: this can be directly a set

soyacz: fixed

unit_tests/test_common_get_db_tables.py
```diff
@@ -0,0 +1,124 @@
...
+def test_get_db_tables_filters_for_non_compact_storage_tables(mock_node_with_schema):
+    """Test get_db_tables correctly filters for non-compact storage tables."""
+    tables = get_db_tables(
+        keyspace_name="feeds",
+        node=mock_node_with_schema,
+        with_compact_storage=False
+    )
+
+    expected_tables = ["table1"]
+    assert sorted(tables) == sorted(expected_tables), f"Expected tables {expected_tables}, got {tables}"
```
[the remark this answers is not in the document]
soyacz: fixed

soyacz:
Fixed all comments, I think this one could get in

sdcm/utils/common.py
```diff
@@ -1599,28 +1601,29 @@ def get_ami_tags(ami_id, region_name):
...
-    for row in list(session.execute(f"select table_name from system_schema.tables where keyspace_name='{keyspace_name}'")):
-        try:
-            create_table_statement = node.run_cqlsh(f"describe {keyspace_name}.{row.table_name}").stdout.upper()
-        except (UnexpectedExit, Libssh2_UnexpectedExit) as err:
-            # SCT issue https://github.com/scylladb/scylla-cluster-tests/issues/7240
-            # May happen when disrupt_add_remove_dc nemesis run in parallel to the disrupt_add_drop_column
-            LOGGER.error("Failed to describe '%s.%s' table. Maybe the table has been deleted. Error: %s",
-                         keyspace_name, row.table_name, err.result.stderr)
+    try:
+        schema_output = node.run_cqlsh("DESC SCHEMA WITH INTERNALS").stdout
```
fruch, line 1614: wondering if  same idea can improve `get_any_ks_cf_list` ?

fruch (approved):
LGTM

sdcm/utils/common.py
soyacz: do we face perf issues with it?

PR #11141, opened 2025-06-11
fix(nemesis): fail test is capacity issue would break topology

cezarmoise:
If a skip occurs here, it will leave the cluster in an unbalanced topology, which will cause errors later in other disruptions.

Example failure: https://argus.scylladb.com/tests/scylla-cluster-tests/9e935144-b1de-415b-8d63-b05446ed9c52/nemesis 
```
disrupt_terminate_and_replace_node	longevity-cdc-100gb-4h-2024-2-db-node-9e935144-5	Skipped	2025-06-09 13:05:42	2025-06-09 13:13:38
Nemesis Information
  Class: Sisyphus
  Name: disrupt_terminate_and_replace_node
  Status: Skipped
  Skip reason: Capacity Issue
Target Information
  Name: longevity-cdc-100gb-4h-2024-2-db-node-9e935144-5
  Public IP: 54.160.183.183
  Private IP: 10.12.11.25
  State: terminated
  Shards: 14
```

Then later repairs fail.

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/62740f79-9f14-4266-be14-51c0126c76bb/events

### Reminders

### DIFF
--- a/sdcm/utils/decorators.py
+++ b/sdcm/utils/decorators.py
@@ -23,9 +23,11 @@ from typing import Optional, Callable
 from botocore.exceptions import ClientError
 
 from sdcm.argus_results import send_result_to_argus
+from sdcm.sct_events import Severity
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.event_counter import EventCounterContextManager
 from sdcm.exceptions import UnsupportedNemesis
+from sdcm.sct_events.system import TestFrameworkEvent
 
 LOGGER = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


fruch:
@cezarmoise 

I don't understand how it's solve anything, the end situation would be that same, capacity mean no node was added, and this nemesis would be marked as error and not skipped, but the next nemesis in both option would be unbalanced.

the only solution I can thing about is to turn those errors into critical event, and stop the test immediately

sdcm/utils/decorators.py
```diff
@@ -373,6 +375,24 @@ def wrapper(*args, **kwargs):
...
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except ClientError as ex:
+            if "InsufficientInstanceCapacity" in str(ex):
+                TestFrameworkEvent(source=callable.__name__,
+                                   message=f"Test failed due to capacity issues: {ex}",
+                                   severity=Severity.CRITICAL).publish()
+            raise
```
pehala, line 392: will this mean that it will raise even if critical event is published? If so, wont it generate two errors?

pehala (commented):
I would like to see testing of the job to make sure it passes. Also I think we can include unit test for the decorator

sdcm/utils/decorators.py
fruch: it would, one would be nemesis failure, and on this critical event., and on the failure of where the test stopped.

critical event are in most cases not coming alone.

we could minimize it by raising a specific exception, that the nemesis loop can identify and change the nemesis event from error to critical

but I wouldn't go into that extent, not until we have more similar use case that dims it.

sdcm/utils/decorators.py
```diff
@@ -373,6 +375,24 @@ def wrapper(*args, **kwargs):
...
+    This should be used when a failure would leave the cluster in an inconsistent topology state
+    """
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        try:
+            return func(*args, **kwargs)
+        except ClientError as ex:
+            if "InsufficientInstanceCapacity" in str(ex):
+                TestFrameworkEvent(source=callable.__name__,
+                                   message=f"Test failed due to capacity issues: {ex}",
```
fruch, line 390: I would expand this message like:
```
cluster is probably unbalanced, continuing with  test would yield unknown results 
```

cezarmoise:
Tested https://argus.scylladb.com/tests/scylla-cluster-tests/62740f79-9f14-4266-be14-51c0126c76bb/events

image

pehala (approved):

dimakr (approved):

PR #11150, opened 2025-06-11
chore(deps): update to python 3.13

fruch:
it's been a while since we update python version
now that we are using uv, it's a bit more straight forward

* change again the ssh2-python library back to the original one, that got an update this year
* updated the kafka related packages

### Testing

- [ ]

### Reminders


sdcm/provision/scylla_yaml/auxiliaries.py
```diff
@@ -56,7 +56,7 @@ class SeedProvider(BaseModel):
     ]
     parameters: List[dict] = None
 
-    @validator("class_name", pre=True, always=True)
+    @field_validator("class_name", mode='before')
```
pehala, line 59: This seems to be independent of python 3.13 change. I would extract it to a separate PR

sdcm/sct_provision/aws/cluster.py
```diff
@@ -439,4 +439,4 @@ def provision(self):
                 name=self.placement_group_name, tags=self.common_tags, region=self._region(0))
 
 
-ClusterNode.update_forward_refs()
+ClusterNode.model_rebuild()
```
pehala, line 442: why? is that a Pydantic change?

pyproject.toml
```diff
@@ -44,20 +41,23 @@ dependencies = [
...
     "PyYAML==6.0.1",
     "pyzmq==26.2.0",
     "questionary==2.1.0",
     "requests==2.32.4",
     "scylla-driver==3.29.3",
     "sortedcontainers==2.4.0",
-    "ssh2-python3==1.11.1",
     "tcconfig==0.29.1",
     "tenacity==9.0.0",
+    "fastavro==1.11.1",
```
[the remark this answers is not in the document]
pehala: How about adding this to pre-commit? https://github.com/ninoseki/uv-sort

sdcm/provision/scylla_yaml/auxiliaries.py
fruch: yes, but all those warning is in my way to the warning related to python 3.13

so they would stay here, until I'm done, and then it I'll put them in separate PRs

sdcm/sct_provision/aws/cluster.py
pehala: I would extract those into separate PR, there are too many changes here already and it is unintuitive to have this bundled with python 3.13 update

pyproject.toml
```diff
@@ -16,15 +16,15 @@ dependencies = [
     "azure-mgmt-resourcegraph==8.0.0",
     "azure-mgmt-subscription==3.1.1",
     "azure-storage-blob==12.16.0",
+    "boto3==1.38.35",
```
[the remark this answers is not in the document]
fruch: this is not important enough to change

soyacz (approved):
LGTM
But would be better if formatting change was in separate PR tht was backported to other branches - so backports have less conflicts in future

soyacz:
> > LGTM But would be better if formatting change was in separate PR tht was backported to other branches - so backports have less conflicts in future
> 
> this isn't going into any release branch

I know, formatting change (only) probably could to avoid problems with future backports

PR #11151, opened 2025-06-11
chore(deps): Update Argus client to 0.15.3

pehala:
* Includes all of the files we ship in the package (including tests and version files)
* Migrate to using `StaticGenericResultTable`
### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/4ab0945c-5045-4309-a5bd-d0da6c568aed/results

### Reminders


argus/client/tests/conftest.py
fruch, line 1: damn it, those were suppose to be excluded from the package...

soyacz: let's remove it from here manually and fix in Argus for future releases

soyacz:
@pehala easy and quick test is for perf-simple-query

pehala:
Ran into issue https://github.com/scylladb/argus/pull/700, I can workaround for it here, but I think we can do a minor Argus patch with both this and also the Manifest

fruch:
@pehala  `v0.15.3` is on pypi
https://github.com/scylladb/argus/actions/runs/15605935578

argus/client/tests/conftest.py
fruch: it's fixed in Argus side

fruch (approved):
LGTM

PR #11164, opened 2025-06-15
fix(cluster_aws): stop allocating public addresses if not needed

fruch:
historically when working with 2 network interfaces we enable public address for the 2nd interface (since it's not automatic)

but since we rework the 2 network interface configuration for AWS we are not running the test with public addresses anymore.

so this change is limiting the creating of public addresses only to cases the configuration suggests using public addresses

### Testing

- [x] tested manually locally - provision + artifact test 
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-200gb-48h-network-monkey-test/4/ (run for 14 hours, just with networking nemesis)
- [ ] 🕙  https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-200gb-48h-network-monkey-test/5/ (run only with topology changes for couple of hours)
 
### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -921,7 +921,7 @@ class BaseNode(AutoSshContainerMixin):
         private_ipv4_addresses = [to_inet_ntop_format(address) for address in private_ipv4_addresses]
         return list(set(public_ipv4_addresses + private_ipv4_addresses + [to_inet_ntop_format(self._get_ipv6_ip_address())]))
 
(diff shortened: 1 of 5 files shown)


fruch:
this is doesn't work as I anticipated, and not having public ips mean no outgoing traffic to the internet...
it would me to configure NAT gateway, which doesn't proably gonna cost more then those few eip addresses that would be saved by this change.

closing this for now.

PR #11179, opened 2025-06-16
fix(nemesis): nemesis_selector missing labels define only on Nemesis class

CodeLieutenant:
When `logical_phrase` was passed into the `filter_subclasses` it will never fill `nemesis_subclasses` array, as it never appends to it cause of the `if not logical_phrase` which is always false for the if condition and `evaluator` is never run.

When `logical_phrase` is not passed (it's None), filtered nemesis is always empty as it is not appended to the new array.

Affected Runs:
- `run_with_gemini`
- `supports_high_disk_utilization`

This line is also problematic

```py
dict(**nemesis.__dict__, **{nemesis.__name__: True})
```

`nemesis` is a class name for the nemesis that will be executed. It tries to get properties on that class, but since every nemesis inherits from `Nemesis`, and in `python` you cannot see the properties from the base class, unless you redeclare them, this code never saw `run_with_gemini` flag, causing every nemesis to be filtered out, and following error:

```
2025-06-13 16:06:30.366: (ThreadFailedEvent Severity.ERROR) period_type=one-time event_id=554566bd-9bb5-4053-8069-a44f08ac4b88: message=no nemesis were selected
File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/decorators.py", line 26, in wrapper
return func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 481, in run
self.disrupt()
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5674, in disrupt
self.call_next_nemesis()
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 1899, in call_next_nemesis
assert self.disruptions_list, "no nemesis were selected"
AssertionError: no nemesis were selected
```

### Testing

- [ ]

### DIFF
--- a/sdcm/nemesis_registry.py
+++ b/sdcm/nemesis_registry.py
@@ -47,15 +47,19 @@ class NemesisRegistry:
         for nemesis in list_of_nemesis:
             if nemesis in self.excluded_list:
                 continue
-            evaluator.context = dict(**nemesis.__dict__, **{nemesis.__name__: True})
-            if logical_phrase and "disrupt_" in logical_phrase and (method_name := get_disrupt_method_from_class(nemesis)):
-                # if the `logical_phrase` has a method name of any disrupt method
-                # we look it up for the specific class and add it to the context
-                # so we can match on those as well
-                # example: 'disrupt_create_index or disrupt_drop_index'
-                evaluator.context[method_name] = True
-            if not logical_phrase or evaluator.visit(expression_ast):
+
+            instance = nemesis.__new__(nemesis)
+            evaluator.context = instance.properties()
+            evaluator.context[nemesis.__name__] = True
+            if logical_phrase:
+                if "disrupt_" in logical_phrase and (method_name := get_disrupt_method_from_class(nemesis)):
+                    evaluator.context[method_name] = True
+
+                if evaluator.visit(expression_ast):
+                    nemesis_subclasses.append(nemesis)
+            else:
                 nemesis_subclasses.append(nemesis)
+
         return nemesis_subclasses
 
     def get_disrupt_methods(self, logical_phrase: str | None = None) -> List[DisruptMethod]:
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -165,7 +165,6 @@ from test_lib.cql_types import CQLTypeBuilder
 from test_lib.sla import ServiceLevel, MAX_ALLOWED_SERVICE_LEVELS
 from sdcm.utils.topology_ops import FailedDecommissionOperationMonitoring
 
-
 LOGGER = logging.getLogger(__name__)
 # NOTE: following lock is needed in the K8S multitenant case
 NEMESIS_LOCK = Lock()
@@ -212,7 +211,6 @@ def target_all_nodes(func: Callable) -> Callable:
 
 
 class Nemesis:
-
     # nemesis flags:
     topology_changes: bool = False  # flag that signal that nemesis is changing cluster topology,
     # i.e. adding/removing nodes/data centers
@@ -284,6 +282,21 @@ class Nemesis:
         self._target_node_pool_type = NEMESIS_TARGET_POOLS.data_nodes
         self.hdr_tags = []
         self.log.debug('Instantiated %s nemesis with %d seed', self.__class__.__name__, self.nemesis_seed)
+
+    def properties(self) -> dict:
+        instance_dict = object.__getattribute__(self, '__dict__')
+        class_attrs = {}
+        for attr_name in dir(self.__class__):
+            if not attr_name.startswith('_') and attr_name not in ('additional_params', 'additional_configs'):
+                try:
+                    attr_value = getattr(self.__class__, attr_name)
+                    if not callable(attr_value):
+                        class_attrs[attr_name] = attr_value
+                except (AttributeError, TypeError):
+                    continue
+
+        combined_dict = {**class_attrs, **instance_dict}
+        return combined_dict
 
     def _init_num_deletions_factor(self):
         # num_deletions_factor is a numeric divisor. It's a factor by which the available-partitions-for-deletion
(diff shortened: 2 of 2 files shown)


pehala:
> I tried to remove those changes, but the precommit hook just adds them (the only way to push it, is to disable it, which is not the way to go)

In that case, you probably have different versions of ruff and autopep from pyproject.toml

fruch:
> 
> 
> > > `nemesis` is a class name for the nemesis that will be executed. It tries to get properties on that class, but since every nemesis inherits from `Nemesis`, and in `python` you cannot see the properties from the base class, unless you redeclare them, this code never saw `run_with_gemini` flag, causing every nemesis to be filtered out, and following error:
> > 
> > I am aware of this limitation, but since I was not the one who introduced it and it was not being an issue, I ignored it (It will be unintentionally fixed by my nemesis rework).
> 
> I made it so that there is a property() method inside Nemesis class which will provide the default values needed by classes that rely on that behavior (aka gemini, maybe some other but didn't look into the code that much). Right now gemini fails everytime (sometimes due to it's own problems, but other times, fails cause of the nemesis not being selected)
> 
> > 
> > Just to be clear the limitation should be that you cannot rely on default value in Nemesis class for filtering as those are not directly accessible on subclasses, meaning if the nemesis subclass does not specifiy the parameter it wont be considered.
> > 
> > For the second problem with logical_phrase, I am not sure if I understand, is the bug that if you dont provide logical_phrase it will return no results, instead of all methods?
> 
> ```python
> 
> if logical_phrase:
>           expression_ast = ast.parse(logical_phrase, mode="eval")
> 
> # A bit later
> if logical_phrase and "disrupt_" in logical_phrase and (method_name := get_disrupt_method_from_class(nemesis)):
>     evaluator.context[method_name] = True
> if not logical_phrase or evaluator.visit(expression_ast):
> ...
> ```
> 
> This code is the problem `expression_ast` is only available when `logical_phrase` exists, and inside this if statement it starts with `if not logical_phrase ...` which means that the evaluator will never run, as `logical_phrase` in that condition must be false (as it has been provided)

I don't understand how we get to empty logical_phrase in this case

Are any of the nemesis labeled with this one ? 

Is that was label True on nemesis, and False on specific one ? If yes, just reverse that as those flags are intended to be use.

That might be a bit more straight forward fix.

> 
> > Sidenote: Please do not include unrelated formatting changes, it makes it hard to understand whats actually changed
> 
> I tried to remove those changes, but the precommit hook just adds them (the only way to push it, is to disable it, which is not the way to go)
> 

That means you are using wrong versions of tooling, please use `uv sync` to make sure your environment is exactly as specified in pyproject.toml

sdcm/nemesis_registry.py
```diff
@@ -47,15 +47,19 @@ def filter_subclasses(self, list_of_nemesis: List[SourceType], logical_phrase: s
...
-                # so we can match on those as well
-                # example: 'disrupt_create_index or disrupt_drop_index'
-                evaluator.context[method_name] = True
-            if not logical_phrase or evaluator.visit(expression_ast):
+
+            instance = nemesis.__new__(nemesis)
+            evaluator.context = instance.properties()
+            evaluator.context[nemesis.__name__] = True
+            if logical_phrase:
+                if "disrupt_" in logical_phrase and (method_name := get_disrupt_method_from_class(nemesis)):
```
fruch, line 55: Where this is is coming from ? I pretty sure @pehala took out in purpose

sdcm/nemesis_registry.py
```diff
@@ -47,15 +47,19 @@ def filter_subclasses(self, list_of_nemesis: List[SourceType], logical_phrase: s
...
-            evaluator.context = dict(**nemesis.__dict__, **{nemesis.__name__: True})
-            if logical_phrase and "disrupt_" in logical_phrase and (method_name := get_disrupt_method_from_class(nemesis)):
-                # if the `logical_phrase` has a method name of any disrupt method
-                # we look it up for the specific class and add it to the context
-                # so we can match on those as well
-                # example: 'disrupt_create_index or disrupt_drop_index'
-                evaluator.context[method_name] = True
-            if not logical_phrase or evaluator.visit(expression_ast):
+
+            instance = nemesis.__new__(nemesis)
```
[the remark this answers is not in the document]
CodeLieutenant: Different Nemesis require different parameters, using new avoids the call for `__init__`

sdcm/nemesis_registry.py
```diff
@@ -47,15 +47,19 @@ def filter_subclasses(self, list_of_nemesis: List[SourceType], logical_phrase: s
...
+            evaluator.context = instance.properties()
```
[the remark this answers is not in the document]
CodeLieutenant: didn't know about that one

sdcm/nemesis_registry.py
CodeLieutenant: this was already in the code

fruch: Now I see it's from above, why did you drop the comment that was explaining it ?

CodeLieutenant: > Now I see it's from above, why did you drop the comment that was explaining it ?

bringing it back

sdcm/nemesis_registry.py
```diff
@@ -47,15 +47,19 @@ def filter_subclasses(self, list_of_nemesis: List[SourceType], logical_phrase: s
         for nemesis in list_of_nemesis:
             if nemesis in self.excluded_list:
                 continue
-            evaluator.context = dict(**nemesis.__dict__, **{nemesis.__name__: True})
```
fruch, line 50: Basically this is the line needs fixing with getting all the needed properties 

The rest of logic doesn't seems to need any changes

If something is messed up with the properties of selection is perfectly fine to end up with empty nemesis list

CodeLieutenant: I tried just fixing that, but again, it will always be empty (at least for gemini) even thou it should not

pehala:
I tested the empty `logical_phrase` and it returns all nemesis, which I think is correct.

Regarding the params, I think we are gonna run into a bit of problem with current proposed solution as we do not want to include every parameter (e.g. like all the functions) but only flags. I created an alternative proposal in #11186 to filter out only flags by extracting flags to separate class

pehala:
No, it is a alternative it doesnt mean it is better, but please comment on the idea. Only once we decide which is better we will close one of them.

PR #11186, opened 2025-06-17
improvement(nemesis): Filter also according to default values

pehala:
Alternative to #11179.
This PR fixes one problem, the fact that when you filter nemesis, only the concrete class parameters are used, any parameter from the subclass (e.g. the default) wont be utilized. 

My proposed solution is to extract flags into separate class that will be taken into account when accessing the parameters as well. This has consequences of adding a lot of lines into nemesis.yml and some minor adjustments in nemesis and tests. I also added filtering to the context parameters to filter out private properties (like `__dict__` or others)

I also added tests for this and the empty `logical_phrase` case

### Testing

- [x] Unit tests

### Reminders


sdcm/nemesis_registry.py
```diff
@@ -90,7 +93,9 @@ def gather_properties(self) -> Tuple[Dict[SourceType, Dict[str, bool]], Dict[str
         method_properties = {}
         for subclass in self.get_subclasses():
             properties_list = {}
-            for attribute in subclass.__dict__.keys():
+            keys = dict(**self.flag_class.__dict__)
```
vponomaryov, line 96: Please, rename the `keys` dictionary to more correct name.
For example - `nemesis_params`, `nemesis_flags` or `nemesis_filters`.

unit_tests/nemesis/test_registry.py
```diff
@@ -74,7 +79,8 @@ def registry():
         ("flag_d", {CustomNemesisD}),
         ("flag_common", {CustomNemesisA, CustomNemesisB, CustomNemesisD}),
         ("", {CustomNemesisA, CustomNemesisB, CustomNemesisC, CustomNemesisD}),
-        (None, {CustomNemesisA, CustomNemesisB, CustomNemesisC, CustomNemesisD})
+        (None, {CustomNemesisA, CustomNemesisB, CustomNemesisC, CustomNemesisD}),
+        ("flag_true", {CustomNemesisB, CustomNemesisC, CustomNemesisD}),  # flag_true is set to True in CustomNemesisA
```
vponomaryov, line 83: Typo, `flag_true` is set to `False` in the `CustomNemesisA` class.

unit_tests/nemesis/test_registry.py
```diff
@@ -7,13 +7,17 @@
...
+class FlagClass:
     flag_a = False
     flag_b = False
     flag_c = False
     flag_d = False
     flag_common = False
+    flag_true = True
+
+
+class TestNemesis(FlagClass):
```
vponomaryov, line 19: Intuitively I would expect the `CustomNemesisD` to be renamed to the `CustomNemesisAD`

vponomaryov (commented):
See inline comments, everything else LGTM

Issue #11022, opened 2025-06-04
Cannot use Duration column in PRIMARY KEY of materialized view


timtimb0t:
## Packages

Scylla version: `2025.2.0~rc2-20250530.19513fa47e62` with build-id `c61a4213002fc8e9dd534a2e2c675a12134c479e`

Kernel Version: `6.11.0-1015-gcp`

Following error occurred during **disrupt_add_remove_mv** nemesis:

```
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5525, in wrapper
    result = method(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5080, in disrupt_add_remove_mv
    self.tester.create_materialized_view(ks_name, base_table_name, view_name, [column],
  File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 2591, in create_materialized_view
    session.execute(query, timeout=600)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 1316, in execute_verbose
    return execute_orig(*args, **kwargs)
  File "cassandra/cluster.py", line 2802, in cassandra.cluster.Session.execute
  File "cassandra/cluster.py", line 5206, in cassandra.cluster.ResponseFuture.result
cassandra.InvalidRequest: Error from server: code=2200 [Invalid query] message="Cannot use Duration column 'w741r57vrg' in PRIMARY KEY of materialized view"

```

need to check if column that piked as primary key  is not **Duration** type from SCT side

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

*Describe your issue in detail and steps it took to produce it.*

## Impact

*Describe the impact this issue causes to the user.*

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 6 nodes (n2-highmem-16)

Scylla Nodes used in this run:
- longevity-large-partitions-200k-pks-db-node-8f06f152-0-8 (34.74.36.4 | 10.142.0.2) (shards: -1)
- longevity-large-partitions-200k-pks-db-node-8f06f152-0-7 (35.243.176.255 | 10.142.0.161) (shards: 14)
- longevity-large-partitions-200k-pks-db-node-8f06f152-0-6 (35.229.55.196 | 10.142.0.153) (shards: 14)
[...]

PR #11189, opened 2025-06-17
fix(nemesis): exclude 'Duration' type column as primary key for MV

timtimb0t:
this commit excludes 'Duration' column types from primary keys for MV creation
fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11022

### Reminders

### DIFF
--- a/sdcm/utils/nemesis_utils/indexes.py
+++ b/sdcm/utils/nemesis_utils/indexes.py
@@ -26,6 +26,21 @@ from sdcm.sct_events.system import InfoEvent
 LOGGER = logging.getLogger(__name__)
 
 
+UNSUPPORTED_PK_NATIVE_TYPES: set[str] = {
+    "counter",
+    "duration",
+}
+
+UNSUPPORTED_PK_COLLECTION_TYPES: tuple[str, ...] = (
+    "list",
+    "set",
+    "map",
+)
+
+# If a type string starts with this, it is frozen and therefore allowed
+FROZEN_PREFIX: str = "frozen<"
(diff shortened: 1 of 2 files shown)


sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -41,24 +41,41 @@ def is_cf_a_view(node: BaseNode, ks, cf) -> bool:
...
                           f" AND table_name = '{cf}'"
                           f" AND {filter_kind}"
                           f" ALLOW FILTERING")
     res_list = list(res)
-    column_names = [row.column_name for row in res_list]
-    if filter_out_collections:
-        collection_types = ('list', 'set', 'map')
-        column_names = [row.column_name for row in res_list if not str(row.type).startswith(collection_types)]
+
+    collection_types = ('list', 'set', 'map')
```
pehala, line 58: I would prefer for constants to be defined as such, outside the method

pehala: Or maybe better, if there is exactly one usage of said constant I would not put in inside get_column_names function but I would have it as a parameter in the nemesis itself.
`get_column_names(session=session, ks=ks_name, cf=base_table_name, is_primary_key=True, filter_out=["duration"])`

timtimb0t: Im not sure regarding the parameter idea because there are may be more types that we need to filter (i dont know what types it could be), so the constants seems to be more extendible. What do you think?

Issue #11161, opened 2025-06-12
`pr-provision-test-docker` is failing on starting monitor stack with simulated racks enabled


dimakr:
When simulated racks are enabled (the default settingg in test_defaults now), the `pr-provision-docker-test` test is failing (including in CI), with the error:
```
21:51:02  ======================================================================
21:51:02  ERROR: test_custom_time (longevity_test.LongevityTest)
21:51:02  Run cassandra-stress with params defined in data_dir/scylla.yaml
21:51:02  ----------------------------------------------------------------------
21:51:02  Traceback (most recent call last):
21:51:02    File "/home/ubuntu/scylla-cluster-tests/longevity_test.py", line 50, in setUp
21:51:02      super().setUp()
21:51:02    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 195, in wrapper
21:51:02      return method(*args, **kwargs)
21:51:02    File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/decorators.py", line 120, in inner
21:51:02      res = func(*args, **kwargs)
21:51:02    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 1066, in setUp
21:51:02      future.result()
21:51:02    File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 451, in result
21:51:02      return self.__get_result()
21:51:02    File "/usr/local/lib/python3.10/concurrent/futures/_base.py", line 403, in __get_result
21:51:02      raise self._exception
21:51:02    File "/usr/local/lib/python3.10/concurrent/futures/thread.py", line 58, in run
21:51:02      result = self.fn(*self.args, **self.kwargs)
21:51:02    File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 1047, in _create_monitors
21:51:02      monitors_init_in_parallel.run(
21:51:02    File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 502, in run
21:51:02      raise ParallelObjectException(results=results)
21:51:02  sdcm.utils.common.ParallelObjectException: [<sdcm.cluster_docker.MonitorSetDocker object at 0x72170b9707c0>]:
21:51:02   Traceback (most recent call last):
21:51:02    File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/common.py", line 486, in run
21:51:02      result = future.result(time_out)
[...]

PR #11169, opened 2025-06-15
test(runner): run tests with pytest

fruch:
switch to run the test only with pytest, and not with `unittest.main()` runner anymore.

* for the sake of testing it a better, some part of the setUp/tearDown were moved into fixtures
* hooks were added to capture test call results, and subtests results
* pytest-subtests introduced, so we can get the subtests results in fixture

Fixes: #11165

### Testing

- [ ]

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #11194, opened 2025-06-17
fix(pr-provision-docker): fix the docker backend pr-provision test

dimakr:
Fixes for PR-provision-docker test to have it passing:
- RHEL UBI9 image (which is the default one for docker backend) doesn't have pgrep command installed, but it is required in `seastar-cpu-map.sh` script.
Because of this, the setUp sequence of tester object is failing on starting CommitLogCheck thread, as the thread uses `seastar-cpu-map.sh` to read number of shards.
To fix this, installation of pgrep was added to centos based scylla-sct image.
- Starting from some point (likely after simulated racks were enabled by default in SCT) the setUp sequence of tester object started to fail on executing `create_monitors` routine, when trying to read `vm_region` info of monitor node object. Monitor node abstraction in docker backend has no implementation of `vm_region`, so it accesses the parent BaseNode class is accessed instead, where the properties are just abstract methods and have no implementation.
To fix this problem, implementation of absent `vm_region` and `region` properties was added in `cluster_docker.DockerMonitoringNode` class.
- After we switched to a different distro as the default one in SCT, using the official latest docker images doesn't work out-of-the-box.
To fix this issue, we are pinning SCT CI to use 'latest' version of docker image from nightly dockerhub repo.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11161

### Testing

- [ ]

### Reminders

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -23,10 +23,14 @@ def createRunConfiguration(String backend) {
         configuration.azure_region_name = 'eastus'
         configuration.availability_zone = ''
     } else if (backend == 'docker') {
+        // use latest version of nightly image for docker backend, until we get rid off rebuilding docker image for SCT
+        configuration.scylla_version = 'latest'
+        configuration.docker_image = 'scylladb/scylla-nightly'
         configuration.test_config = "test-cases/PR-provision-test-docker.yaml"
     } else if (backend in ['k8s-local-kind-aws', 'k8s-eks']) {
         if (scylla_version.endsWith('latest')) {
             configuration.scylla_version = 'latest'
+            configuration.docker_image = 'scylladb/scylla-nightly'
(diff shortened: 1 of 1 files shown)


fruch:
@dimakr 

you are missing this 
https://github.com/scylladb/scylla-cluster-tests/pull/11169/commits/ba2febf3ba5ccdfbdbe4fe7dd8b3ad5f0187bb9e

so far it was picking latest official docker, which is no UBI9 yet...

fruch:
@dimakr 

my fix doesn't work without putting `latest` in scylla_version

if scylla_version is blank, it would still go and take older release...

dimakr:
> @dimakr
> my fix doesn't work without putting `latest` in scylla_version
> if scylla_version is blank, it would still go and take older release...

yes, this helps
will make docker backend to use hardcoded 'latest' version, until we get rid off rebuilding scylla images

PR #11195, opened 2025-06-17
fix(pydantic): fix serializer warning

fruch:
pydantic excpect enum to be return:
```
Pydantic serializer warnings:
  Expected `enum` but got `str` with value `'instance_store'` - serialized value may not be as expected
  return self.__pydantic_serializer__.to_python(
```

change the enum to be retrun it's value with converted to string, that works as expected, and keep pydantic happy

while there aligned `raid_level` to return the enum expected regardless of configureation set or not.

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/provision/common/user_data.py
+++ b/sdcm/provision/common/user_data.py
@@ -13,13 +13,12 @@
 
 import abc
 from enum import Enum
+from typing import Literal
 
 from sdcm.provision.common.builders import AttrBuilder
 
 
-class DataDeviceType(str, Enum):
-    ATTACHED = 'attached'
-    INSTANCE_STORE = 'instance_store'
+DataDeviceType = Literal['attached', 'instance_store']
 
 
 class RaidLevelType(int, Enum):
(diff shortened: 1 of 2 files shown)


fruch:
> > Could we migrate `DataDeviceType` to be StrEnum instead? Would that fix the issue?
> 
> No, not introducing a package for this, this should be a Literal and that's it
> 
> Enums doesn't make any sense here what so ever...

pydantic can validate also a Literal

fruch:
> > > Could we migrate `DataDeviceType` to be StrEnum instead? Would that fix the issue?
> > 
> > 
> > No, not introducing a package for this, this should be a Literal and that's it
> > Enums doesn't make any sense here what so ever...
> 
> So this just a hotfix, but if someone wants to spend more time on this, the correct way is to convert this to Literal?

well I can do it now, we don't really use this Enum anywhere by when we read from configuration
and even there we are using if statement for the two options

Enums of two string options,  and we are not using them, we pass the string down to scylla
ontop of it we limit is with choices on `sct_config.py`

PR #11201, opened 2025-06-18
fix(hydra): clean-resources/list-resources default to scan all regions

fruch:
cause when people are using it locally, it doesn't scan all regions but just the default regions in each cloud

people were leaving lot of AWS resources behind in some resources with them knowning about them.

### Testing

- [x] tested locally

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -296,6 +296,10 @@ def clean_resources(ctx, post_behavior, user, test_id, logdir, dry_run, backend)
     add_file_logger()
 
     user_param = {"RunByUser": user} if user else {}
+    if user:
+        os.environ["SCT_REGION_NAME"] = os.environ.get("SCT_REGION_NAME", "")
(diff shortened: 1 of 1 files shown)


vponomaryov:
> I've tried `clean-resources --user lukasz.sojka` and it cleaned only on AWS. 

It is caused by the following code:
```
...
def clean_resources(ctx, post_behavior, user, test_id, logdir, dry_run, backend):
...
    if backend is None:
        if os.environ.get('SCT_CLUSTER_BACKEND', None) is None:
            os.environ['SCT_CLUSTER_BACKEND'] = 'aws'
    else:
        os.environ['SCT_CLUSTER_BACKEND'] = backend
...
```

As far as I remember, it is designed to cover exactly one backend because we don't have "mixed backends" test configurations.
And running this command in each CI test run we cover "current" backend only.

soyacz (approved):

fruch:
Setting SCT_CLUSTER_BACKEND empty string, make clean to work on all

we can make it as default if we wish, in CI we always set the backend

PR #11202, opened 2025-06-18
fix(nemesis): exclude 'Duration' type column as primary key for MV

timtimb0t:
this commit excludes 'Duration' and 'counter' column types from
 primary keys for MV and secondary index  creation
fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11022

Testing:
https://jenkins.scylladb.com/job/scylla-staging/job/eugene_test_folder/job/duration_columns_test/

### Reminders

### DIFF
--- a/sdcm/utils/nemesis_utils/indexes.py
+++ b/sdcm/utils/nemesis_utils/indexes.py
@@ -41,7 +41,8 @@ def is_cf_a_view(node: BaseNode, ks, cf) -> bool:
             return False
 
 
-def get_column_names(session, ks, cf, is_primary_key: bool = False, filter_out_collections: bool = False, filter_out_static_columns: bool = False) -> list:
+def get_column_names(session, ks, cf, is_primary_key: bool = False, filter_out_collections: bool = False,
+                     filter_out_static_columns: bool = False, filter_out_column_types: list[str] | None = None) -> list:
     column_types = "'regular'" if filter_out_static_columns else "'static', 'regular'"
     filter_kind = f" kind in ({column_types})" if not is_primary_key else "kind in ('partition_key', 'clustering')"
     res = session.execute(f"SELECT column_name, type FROM system_schema.columns"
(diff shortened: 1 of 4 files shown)


sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -41,7 +42,8 @@ def is_cf_a_view(node: BaseNode, ks, cf) -> bool:
             return False
 
 
-def get_column_names(session, ks, cf, is_primary_key: bool = False, filter_out_collections: bool = False, filter_out_static_columns: bool = False) -> list:
+def get_column_names(session, ks, cf, is_primary_key: bool = False, filter_out_collections: bool = False,
```
fruch, line 45: as this one gets more logic into it,

maybe it's worth having a unit/integration test for it

see `test_get_any_ks_cf_list` for a quick why to write such test

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -50,15 +52,25 @@ def get_column_names(session, ks, cf, is_primary_key: bool = False, filter_out_c
                           f" AND {filter_kind}"
                           f" ALLOW FILTERING")
     res_list = list(res)
-    column_names = [row.column_name for row in res_list]
-    if filter_out_collections:
-        collection_types = ('list', 'set', 'map')
-        column_names = [row.column_name for row in res_list if not str(row.type).startswith(collection_types)]
+    collection_types = ('list', 'set', 'map')
```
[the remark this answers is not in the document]
timtimb0t: there are may be cases when using prefix may affect the test behavior i guess. For example if we will test user define types, dont know if we need to take it into account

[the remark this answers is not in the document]
timtimb0t: Nah, my bad, we still need to exclude them so we can rely on naming

pehala (commented):
LGTM, but I would still like to see some validation it is working, e.g. by running the two affected nemesis on their own

timtimb0t:
> LGTM, but I would still like to see some validation it is working, e.g. by running the two affected nemesis on their own

To be honest i dont know how to reproduce it, added unit test for the affected function

pehala:
> To be honest i dont know how to reproduce it, added unit test for the affected function

Dont need reproducer, I just want to know if you did not break the two nemesis

unit_tests/test_get_columns_for_primary_cluster_keys.py
```diff
@@ -0,0 +1,65 @@
+import unittest
+from unittest.mock import MagicMock
+from sdcm.utils.nemesis_utils.indexes import get_column_names
+
+
+class TestGetColumnNamesWithCQLTypes(unittest.TestCase):
```
pehala, line 6: I would suggest using pytest instead, as we want to migrate to using it in the near future

pehala (approved):

sdcm/utils/nemesis_utils/indexes.py
```diff
@@ -41,7 +41,8 @@ def is_cf_a_view(node: BaseNode, ks, cf) -> bool:
             return False
 
 
-def get_column_names(session, ks, cf, is_primary_key: bool = False, filter_out_collections: bool = False, filter_out_static_columns: bool = False) -> list:
+def get_column_names(session, ks, cf, is_primary_key: bool = False, filter_out_collections: bool = False,
+                     filter_out_static_columns: bool = False, unsupported_primary_key_columns: list[str] | None = None) -> list:
```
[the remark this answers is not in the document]
vponomaryov: +1

unit_tests/test_get_columns_for_primary_cluster_keys.py
fruch: we already using pytest as test runner for the unit tests and integration tests,

it can run both types of tests, but we can build the same without test classes (and utilize more of pytest features)

fruch: @timtimb0t 

regardless I would recommend have integration test for this function, that works with actual scylla 
like the one that I pointed to

soyacz (approved):
LGTM

unit_tests/test_get_columns_for_primary_cluster_keys.py
pehala: I would like to not introduce anymore unittests cases unless it is needed, I think it is strictly worse solution than using pytest directly

pehala (approved):

soyacz (approved):
LGTM

fruch (approved):
LGTM

PR #11212, opened 2025-06-18
fix(SLs): remove scylla_scheduler_runtime_ms metric check

temichus:
Remove the scylla_scheduler_runtime_ms metric verification from test_read_throughput_1to5_ratio due to looking into resources/CPU \ is wrong because 1 operation from low shares user can consume more resources than several operations from high shares the user

There is only 1 check in the test left: check CS read ratio

also test now has lower instances->latency about 1s on read and As a result, better stability in terms of expected CS read ratio

Now test also shows stable results with lower cpu utilization

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/6c68cfa0-94a5-4599-ae7f-ee28a3c44751/events runs 90-77

### Reminders

### DIFF
--- a/sdcm/db_stats.py
+++ b/sdcm/db_stats.py
@@ -315,24 +315,6 @@ class PrometheusDBStats:
         else:
             return res
 
-    def get_scylla_scheduler_runtime_ms(self, start_time, end_time, node_ip, irate_sample_sec='30s'):
-        """
-        Get Scylla CPU scheduler runtime from PrometheusDB
-
-        :return: list of tuples (unix time, op/s)
-        """
-        if not self._check_start_end_time(start_time, end_time):
-            return {}
-        # the query is taken from the Grafana Dashborad definition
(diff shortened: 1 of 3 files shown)


piodul (commented):
The removal of the check looks ok, but what about the change to `MIN_CPU_UTILIZATION` and the change of the instance type? Should those be included in this PR, and if yes - why? I don't see it being mentioned in the description.

temichus:
> The removal of the check looks ok, but what about the change to `MIN_CPU_UTILIZATION` and the change of the instance type? Should those be included in this PR, and if yes - why? I don't see it being mentioned in the description.

also test now has lower instances->latency about 1s on read and As a result, better stability in terms of expected CS read ratio

Also need to add. "NowTest also shows stable results with lower cpu utilization"

piodul (approved):
Looks good after updating the description.

Issue #10201, opened 2025-02-26
disrupt_snapshot_operations nemesis failed with different from actual expected result


timtimb0t:
## Packages

Scylla version: `2025.1.0~rc2-20250216.6ee17795783f` with build-id `8fc682bcfdf0a8cd9bc106a5ecaa68dce1c63ef6`

Kernel Version: `6.8.0-1021-aws`

disrupt_snapshot_operations nemesis fails with error:
```
2025-02-20 03:35:16.943: (DisruptionEvent Severity.ERROR) period_type=end event_id=c5304ddb-3624-4bd6-9110-6c0e0199c9ad duration=48s: nemesis_name=SnapshotOperations target_node=Node longevity-200gb-48h-verify-limited--db-node-5f6d2ff2-5 [34.244.93.136 | 10.4.0.172] errors=Snapshot content not as expected.
Expected content: [['keyspace1', 'standard1'], ['lowload1', 'standard1']]
Actual snapshot content: [['100gb_sizetiered_6_1', 'standard1'], ['keyspace1', 'standard1'], ['lowload1', 'standard1']]
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5509, in wrapper
result = method(*args[1:], **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3468, in disrupt_snapshot_operations
self._validate_snapshot(nodetool_cmd=nodetool_cmd, snapshot_content=snapshot_content)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 3365, in _validate_snapshot
raise AssertionError(f"Snapshot content not as expected. \n"
AssertionError: Snapshot content not as expected.
Expected content: [['keyspace1', 'standard1'], ['lowload1', 'standard1']]
Actual snapshot content: [['100gb_sizetiered_6_1', 'standard1'], ['keyspace1', 'standard1'], ['lowload1', 'standard1']]
```
Seems like its parsing error because **listsnapshots** cmd shows that **100gb_sizetiered_6_1** keyspace exists

## Impact

Just SCT error

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 4 nodes (i4i.4xlarge)

Scylla Nodes used in this run:
[...]

PR #11215, opened 2025-06-19
fix(nemesis): clearing keyspaces after mngmt restore operations

timtimb0t:
this commit changes mngmt restore nemesis to  clear its keyspaces
 when it's done

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/10201

### Testing
https://jenkins.scylladb.com/job/scylla-staging/job/eugene_test_folder/job/snapshots/

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -2896,7 +2896,7 @@ class Nemesis(NemesisFlags):
         self._mgmt_backup(backup_specific_tables=False)
 
     @target_data_nodes
-    def disrupt_mgmt_restore(self):
+    def disrupt_mgmt_restore(self):  # noqa: PLR0914
         def get_total_scylla_partition_size():
             result = self.cluster.data_nodes[0].remoter.run("df -k | grep /var/lib/scylla")  # Size in KB
             free_space_size = int(result.stdout.split()[1]) / 1024 ** 2  # Converting to GB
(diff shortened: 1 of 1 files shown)


timtimb0t:
Hi @fruch , could you please help with this change? I tried to reproduce the failure with current PR state but both nemeses are passed. Could you please clarify what changes are needed more? Seems like I do something wrong

fruch:
> Hi @fruch , could you please help with this change? I tried to reproduce the failure with current PR state but both nemeses are passed. Could you please clarify what changes are needed more? Seems like I do something wrong

I'm not exactly sure what you are trying to test in this PR, but if your fix here is to clear the key space after a nemesis, that what you should test that is working (I would advice to make sure it's happening, even if nemesis fails, i.e. try/expect or some context manager)

now why the snapshot operation nemesis got wrong expectation, that's a differnet angle, and you'll first need to investigate the concurrences it was happening in.

there a complex query to get the list of expected keyspaces, and then it compare it with the actual snapshot.
I would guess there a reason why the complex query didn't got that keyspace, that is probably the root cause, but has nothing todo with PR.

sdcm/nemesis.py
```diff
@@ -3039,10 +3039,19 @@ def _restore_schema(locations: list, cluster_id: str, tag: str) -> None:
...
-        for stress in stress_queue:
-            is_passed = self.tester.verify_stress_thread(stress)
-            assert is_passed, (
-                "Data verification stress command, triggered by the 'mgmt_restore' nemesis, has failed")
+        try:
+            for stress in stress_queue:
+                is_passed = self.tester.verify_stress_thread(stress)
+                assert is_passed, (
+                    "Data verification stress command, triggered by the 'mgmt_restore' nemesis, has failed")
+        finally:
```
pehala, line 3047: Why do we have it in finally for stress? The drop is not tied to the stress, it can be simply appended after the stress section, right?

temichus (approved):

enaydanov (approved):

PR #11229, opened 2025-06-20
feature(LogCollector): Use argus proxy URLs instead of S3 links

k0machi:
This change changes how links to the logs and screenshots are displayed
to the user in SCT reporting - instead of using raw S3 urls with public
bucket, the provided link now proxies the result through argus, allowing
bucket to be private.

Closes scyllab/qa-tasks#1874

### Testing

- [ ] [Jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/alexey-argus-testing/128/)
- 

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -54,7 +54,7 @@ from sdcm.sct_runner import AwsSctRunner, GceSctRunner, AzureSctRunner, get_sct_
     update_sct_runner_tags, list_sct_runners
 from sdcm.utils.ci_tools import get_job_name, get_job_url
 from sdcm.utils.git import get_git_commit_id, get_git_status_info
-from sdcm.utils.argus import argus_offline_collect_events, get_argus_client
+from sdcm.utils.argus import argus_offline_collect_events, create_proxy_argus_s3_url, get_argus_client
 from sdcm.utils.aws_kms import AwsKms
 from sdcm.utils.azure_region import AzureRegion
 from sdcm.utils.cloud_monitor import cloud_report, cloud_qa_report
@@ -1267,7 +1267,8 @@ def collect_logs(test_id=None, logdir=None, backend=None, config_file=None):
             #  current_cluster_type will be "warning"
             if cluster_type == 'sct-runner' and cluster_type not in link:
                 current_cluster_type = link.split("/")[-1].split("-")[0]
-            table.add_row([current_cluster_type, link])
+            table.add_row([current_cluster_type, create_proxy_argus_s3_url(
+                link).format(collector.test_id, link.split("/")[-1])])
 
     click.echo(table.get_string(title="Collected logs by test-id: {}".format(collector.test_id)))
     update_sct_runner_tags(backend=backend, test_id=collector.test_id, tags={"logs_collected": True})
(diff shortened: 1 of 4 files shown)


k0machi:
As I suspected, the screenshots won't be embeddable - for that we can rework them to instead be attached to the emails themselves (compressed with webp for example):

image

fruch:
> > As I suspected, the screenshots won't be embeddable - for that we can rework them to instead be attached to the emails themselves (compressed with webp for example):
> > image
> 
> Monitoring screenshots must be safe to keep public, IMHO. It doesn't store any sensitive information. Masking links for sensitive logs makes total sense even having screenshots be public.

that what we said about the logs for years...

let's see what we can do with that first

k0machi:
@soyacz Clickable links are already there, they are above the screenshots and are tagged as "Download X Screenshot"

@vponomaryov @fruch 
So should I keep screenshots public for now and do a follow up with attachment method or just keep them public period?

fruch:
> @soyacz Clickable links are already there, they are above the screenshots and are tagged as "Download X Screenshot"
> 
> @vponomaryov @fruch 
> So should I keep screenshots public for now and do a follow up with attachment method or just keep them public period?

What stops us serving them from Argus ?

Embedding them, means beefing everyone inboxes...

Please raise a new issue for it

k0machi:
> > @soyacz Clickable links are already there, they are above the screenshots and are tagged as "Download X Screenshot"
> > @vponomaryov @fruch
> > So should I keep screenshots public for now and do a follow up with attachment method or just keep them public period?
> 
> What stops us serving them from Argus ?
> 
> Embedding them, means beefing everyone inboxes...
> 
> Please raise a new issue for it

It would be impossible to embed them while also requiring authorization as every single email provider would never make a 3rd party request with cookies attached - all links are sanitized and proxied by google themselves.

Instead I propose keeping the screenshots public - the entropy provided by both the UUID of the test and the timestamp of the folder inside makes them unindexable, the only way to get them is to have the direct link. Since we are already disallowing traversal of the bucket, the screenshots are effectively private (unless you leak the link of course)

k0machi:
> It would be impossible to embed them while also requiring authorization as every single email provider would never make a 3rd party request with cookies attached - all links are sanitized and proxied by google themselves.

It would be possible to create a weeklong pre-signed link that would work as authorization data is embedded into the URL - but that means older email embeds will eventually expire.

k0machi:
> > It would be impossible to embed them while also requiring authorization as every single email provider would never make a 3rd party request with cookies attached - all links are sanitized and proxied by google themselves.
> 
> It would be possible to create a weeklong pre-signed link that would work as authorization data is embedded into the URL - but that means older email embeds will eventually expire.

Per AWS Docs:
```
The minimum value you can set is 1, and the maximum is 604800 (seven days).
```

vponomaryov:
> > > @soyacz Clickable links are already there, they are above the screenshots and are tagged as "Download X Screenshot"
> > > @vponomaryov @fruch
> > > So should I keep screenshots public for now and do a follow up with attachment method or just keep them public period?
> > 
> > 
> > What stops us serving them from Argus ?
> > Embedding them, means beefing everyone inboxes...
> > Please raise a new issue for it
> 
> It would be impossible to embed them while also requiring authorization as every single email provider would never make a 3rd party request with cookies attached - all links are sanitized and proxied by google themselves.
> 
> Instead I propose keeping the screenshots public - the entropy provided by both the UUID of the test and the timestamp of the folder inside makes them unindexable, the only way to get them is to have the direct link. Since we are already disallowing traversal of the bucket, the screenshots are effectively private (unless you leak the link of course)

For now, our test run screenshots don't expose any sensitive data - it is just fact.
So, I don't see any problem in creation of proxy links only for logs and leave screenshots for later.

vponomaryov (approved):
LGTM

fruch (approved):
LGTM

PR #11233, opened 2025-06-22
feature(get_supported_scylla_base_versions): stop picking OSS version

fruch:
from now on, we should be picking the OSS version as bases versions all of the code path and unittest for that are removed.

one can still manually select a specific branh to upgrade from if needed, but we won't be test that by default anymore.

### Testing

- [x] 🟡 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ami-test/21/ (selection of base as expected, even that test was failing)
- [x] 🟡 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-centos9-test/16/ (selection of base as expected, even that test was failing)
- [x] update unittest for this part
 
### Reminders

### DIFF
--- a/unit_tests/test_base_version.py
+++ b/unit_tests/test_base_version.py
@@ -12,14 +12,20 @@
 # Copyright (c) 2021 ScyllaDB
 
 
-import unittest
-
 from sdcm.utils.version_utils import ComparableScyllaVersion
 
 from utils.get_supported_scylla_base_versions import UpgradeBaseVersion
 
+"""
+This module contains tests for the UpgradeBaseVersion class, which is used to determine
+the supported Scylla versions for upgrades based on the provided repository, backend, and Linux distribution.
+"""
+
 
 def general_test(scylla_repo='', linux_distro='', cloud_provider=None):
+    """
+    General test function to retrieve the list of supported Scylla versions for upgrade.
+    """
     scylla_version = None
 
     version_detector = UpgradeBaseVersion(scylla_repo, linux_distro, scylla_version)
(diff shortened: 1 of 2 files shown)


soyacz:
shouldn't we skip backport to 2025.2 too?

fruch:
> shouldn't we skip backport to 2025.2 too?

after talking with @roydahan, we agreed that from 2025.2 onward we can drop it.

unit_tests/test_base_version.py
```diff
@@ -28,99 +34,119 @@ def general_test(scylla_repo='', linux_distro='', cloud_provider=None):
...
+    version_list = general_test(scylla_repo, linux_distro)
+    assert set(version_list) == {'4.3', '2020.1', '2021.1'}
+
+
+def test_2021_1_with_centos8_rpm():
+    """ Test that development branch of 2021.1 with centos8 is returned correct versions """
+    scylla_repo = url_base + '-enterprise/branch-2021.1/rpm/centos/2021-08-29T00:58:58Z/scylla.repo'
+    linux_distro = 'centos-8'
+    version_list = general_test(scylla_repo, linux_distro)
+    assert set(version_list) == {'4.3', '2021.1'}
```
vponomaryov, line 78: Do we really need tests for such old versions?

fruch: they help testing the logic is correct, even if it's old release.
I don't thin we don't really need to add tests for each release....

utils/get_supported_scylla_base_versions.py
```diff
@@ -82,69 +83,82 @@ def set_start_support_version(self, backend: str = None) -> None:
...
+        """
+        Filter out the base versions that only have release candidates (rc) versions.
 
-    def filter_rc_only_version(self, base_version_list: list) -> list:
+        if it's master branch, we also limit the base version to the latest one,
+        which isn't release candidates (rc) only release.
+
+        :param base_version_list: list of base versions to filter
+        """
         LOGGER.info("Filtering rc versions from base version list...")
```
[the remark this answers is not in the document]
fruch: we should ever return empty result... it better of that we'll fail

unit_tests/test_base_version.py
fruch: I've removed older unsupported version checks

PR #11261, opened 2025-06-25
fix(docker_hub_login): use proper permissions when

dimakr:
A small regression was introduced in [4541ac1](https://github.com/scylladb/scylla-cluster-tests/commit/4541ac1bef87cf5c7cc9a0d302e44480e9fb532d) - `docker_utils.docker_hub_login` could try
password file create/rm operations under one user, but try to write to that file under another one.

This change fixes the `docker_hub_login` function, to perform all manipulations with password file with the same permissions.

### Testing

- [ ]

### Reminders


fruch:
Seems like our send/receive commands have trouble with files that only sudo has access to (I'm running into it in a different issue)

We'll need to address that first

fruch:
@dimakr 

https://github.com/scylladb/scylla-cluster-tests/pull/11260 is more or less ready 

want to take it with reapplying you docker refactoring changes, and try it out ?

dimakr:
> @dimakr
> #11260 is more or less ready
> want to take it with reapplying you docker refactoring changes, and try it out ?

will test today

PR #11263, opened 2025-06-25
feature(S3Storage): support downloading via Argus links

fruch:
since #11229 was merged, the reported urls into Argus, are not direct links to s3 objects, and this code part now need to figure out how to download them, when a test needs that (for example resotre of a monitor stack).

this commit introduce a way for the download to identify it's not s3 url, and follow the redirect to get the correct url to use for the actual s3 download

### Testing

- [x] tested manually, and was able to restore up a monitor stack
- [x] aws provision test - verified the show-monitor download part was working

### Reminders

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -39,7 +39,7 @@ import traceback
 import ctypes
 import shlex
 from typing import Iterable, List, Callable, Optional, Dict, Union, Literal, Any, Type
-from urllib.parse import urlparse
+from urllib.parse import urlparse, urljoin
 from unittest.mock import Mock
 from textwrap import dedent
 from contextlib import closing, contextmanager
@@ -252,6 +252,7 @@ class S3Storage():
     enable_multipart_threshold_size = 1024 * 1024 * 1024  # 1GB
     multipart_chunksize = 50 * 1024 * 1024  # 50 MB
     num_download_attempts = 5
+    s3_host_name_regex = re.compile(r"https?://([a-zA-Z0-9-]+\.s3\.amazonaws\.com)")
 
     def __init__(self, bucket=None):
         if bucket:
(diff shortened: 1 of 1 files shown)


k0machi (approved):

sdcm/utils/common.py
```diff
@@ -312,6 +313,18 @@ def set_public_access(self, key):
...
     def download_file(self, link, dst_dir):
+        """Download file from S3 bucket or Argus proxy link."""
+
+        if not self.s3_host_name_regex.match(urlparse(link).hostname):
+            # get the actual s3 link from Argus first
+            creds = KeyStore().get_argus_rest_credentials()
+            headers = {"Authorization": f"token {creds['token']}", **creds["extra_headers"]}
+
+            response = requests.head(link, allow_redirects=True, headers=headers)
+            link = response.history[-1].headers.get('location', link)
```
k0machi, line 324: One thing here - `Location` would be capitalized in this case I think, unless requests handles this.

fruch: requests gets it either way (it print them lower-case)

sdcm/utils/common.py
```diff
@@ -312,6 +313,18 @@ def set_public_access(self, key):
         acl_obj.put(ACL='', AccessControlPolicy={'Grants': grants, 'Owner': acl_obj.owner})
 
     def download_file(self, link, dst_dir):
+        """Download file from S3 bucket or Argus proxy link."""
+
+        if not self.s3_host_name_regex.match(urlparse(link).hostname):
```
soyacz, line 318: we should rather match for 'argus' domain - otherwise we could leak credentials if someone passed malicious host.

fruch: also I don't want to limit it to specific URL of argus, it could limit us in testing with staging env and such.

fruch: anyhow it's blocking people from restoring monitoring (and fail CI provisioning test), I'm pushing it in.

PR #11291, opened 2025-06-27
improvement(treewide): Use ruff format instead of autopep8

pehala:
Ruff is a superior formatting tool, in my opinion, and we already use it. We also use it in dtest.
If we want to cut down on number of changes we can ignore quotes (or extract them into followup PR), but I still think unifying them make sense. Given how many lines were changed, we can also delay this PR for later, as it will probably make conflicts in existing PRs.

* All changes were generated automatically
* Remove autopep8

Backporting to limit the future problems with backports

### Testing

- [x] None

### Reminders


pehala:
> I suggested it when we introduced ruff
> 
> Since there were too many changes to the code we didn't roll it out, cause it's gonna break every single backport to branches that don't have it.
> 
> 2024.1 and v15, we don't have ruff yet...
> 
> I think we'll need to wait a bit more

I will leave the call up to you, but 2024.1 should be not supported anymore, since 2025.1 and
regarding V15, I do not know the deprecation date but I am willing to introduce ruff there as well, via custom backport, if we wanted to give it a try

vponomaryov (changes requested):
I like how fast is the ruff formatter.

But this is not acceptable:

from
```python
        self.reconfigure_keyspaces_to_use_network_topology_strategy(
            keyspaces=system_keyspaces + ["keyspace1"],
            replication_factors={dc: len(status[dc].keys()) for dc in status}
        )
```

to
```python
        self.reconfigure_keyspaces_to_use_network_topology_strategy(keyspaces=system_keyspaces + ["keyspace1"], replication_factors={dc: len(status[dc].keys()) for dc in status})
```

178 symbols in one line ^

Need to update the `pyproject.toml` config to match the value that is set for autopep8 config - 120 symbols per line limit.
Currently it is set to be `240` >.<

Reason:
I belong to group of people who use split screen in IDE and GitHub.
Using 15.6" display on laptop and keeping readable font size in (my) best case one line fits 112 symbols.
Small count of lines to be <119 is kind of ok.

pehala:
> Need to update the `pyproject.toml` config to match the value that is set for autopep8 config - 120 symbols per line limit.
> Currently it is set to be `240` >.<

I am all for changing the line limit :)

pehala:
> Need to update the `pyproject.toml` config to match the value that is set for autopep8 config - 120 symbols per line limit.
> Currently it is set to be `240` >.<

changed

fruch:
> > I suggested it when we introduced ruff
> > Since there were too many changes to the code we didn't roll it out, cause it's gonna break every single backport to branches that don't have it.
> > 2024.1 and v15, we don't have ruff yet...
> > I think we'll need to wait a bit more
> 
> I will leave the call up to you, but 2024.1 should be not supported anymore, since 2025.1 and regarding V15, I do not know the deprecation date but I am willing to introduce ruff there as well, via custom backport, if we wanted to give it a try

there still 6 months for 2024.1 to be supported, and we don't have a due date deprecating V15.
if you want to create PRs for each with introducing ruff in, then I'm o.k. with it.

otherwise it would break every single back port we'll need to do.

pehala:
> there still 6 months for 2024.1 to be supported, and we don't have a due date deprecating V15. if you want to create PRs for each with introducing ruff in, then I'm o.k. with it.
> 
> otherwise it would break every single back port we'll need to do.

* perf-v15 - #11293
  * Catch up on all ruff related PRs, it already had ruff but was outdated. After this patch backport should be fine
* 2024.1 - #11545
  * Replacement for this PR. I added ruff for formatting only, pylint still stays

fruch:
@pehala what about https://github.com/scylladb/scylla-cluster-tests/pull/11545 ? 

if we merge this and it's not finalize, we'll have problem with any backport into 2024.1

soyacz:
IMHO we should create separate PR's to each supported branch - backports of this PR will generate many conflicts and won't be feasible to resolve them. Just rerun ruff on top of current state on each branch.
This way we can also align all the branches at the same time to avoid backport issues (just merge them in one batch).

For this we could first backport all the configs as usual, and then create reformatting PR's to each branch.
Also we could add `.git-blame-ignore-revs` for each branch with reformat commit sha so we have better 'git blame'.

soyacz:
> this is more or less what was agreed with @pehala. but the fact we need to rebuild docker image what holding it off a bit

Rebuilding only for 2024.1 I presume. This one time work I could do.

nyh:
> Since there were too many changes to the code we didn't roll it out, cause it's gonna break every single backport to branches that don't have it.

I wonder why a reasonable conclusion from this can't just be... Don't do it. It is anti-productive and has zero real benefits.

fruch (approved):
LGTM

Issue #11258, opened 2025-06-24
Gradual throughput test hangs at the end


soyacz:
perf gradual throughput test hangs upon completion.

See e.g. https://jenkins.scylladb.com/job/scylla-staging/job/yulia/job/performance/job/perf-regression-predefined-throughput-steps-sanity-vnodes/

I tried to investigate and found:
```
Collecting samples from 'python3 ./sct.py run-test performance_regression_gradual_grow_throughput.PerformanceRegressionPredefinedStepsTest.test_write_gradual_increase_load --backend gce' (python v3.10.12)
Total Samples 1700
GIL: 0.00%, Active: 2600.00%, Threads: 28

  %Own   %Total  OwnTime  TotalTime  Function (filename)                                                                                                                                                                                                                                
1200.00% 1200.00%   204.0s    204.0s   _process_output (sdcm/remote/libssh2_client/__init__.py)
1200.00% 1200.00%   204.0s    204.0s   simple_select (sdcm/remote/libssh2_client/session.py)
100.00% 100.00%   17.00s    17.00s   _run_loop (cassandra/io/libevreactor.py)
100.00% 100.00%   17.00s    17.00s   _wait_for_tstate_lock (threading.py)
  0.00% 1200.00%   0.010s    204.0s   _read_output (sdcm/remote/libssh2_client/__init__.py)
  0.00% 2500.00%   0.000s    425.0s   _bootstrap (threading.py)
  0.00% 1200.00%   0.000s    204.0s   _journal_thread (sdcm/utils/remote_logger.py)
  0.00% 2500.00%   0.000s    425.0s   _bootstrap_inner (threading.py)
  0.00% 1200.00%   0.000s    204.0s   run (sdcm/remote/remote_base.py)
  0.00% 1200.00%   0.000s    204.0s   _retrieve (sdcm/utils/remote_logger.py)
  0.00% 1200.00%   0.000s    204.0s   _run (sdcm/remote/remote_base.py)
  0.00% 2400.00%   0.000s    408.0s   run (sdcm/remote/libssh2_client/__init__.py)
  0.00% 1200.00%   0.000s    204.0s   _worker (concurrent/futures/thread.py)
  0.00% 100.00%   0.000s    17.00s   _shutdown (threading.py)
  0.00% 100.00%   0.000s    17.00s   join (threading.py)
  0.00% 100.00%   0.000s    17.00s   _python_exit (concurrent/futures/thread.py)
[...]

PR #11309, opened 2025-06-30
fix(libssh2): fix hanging ssh connection after the test

soyacz:
When test ends it may hang due awaiting ssh command completion, while
connection was closed (when finishing a test).
This happened due `tail -f` running forever against HDR Histogram log
file.

Fix by ending this `tail -f` command with `pkill`, so reader thread ends
normally and closes all ssh connections.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11258

### Testing

- [x] - [see jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/lukasz/job/perf-regression-predefined-throughput-steps-sanity-vnodes/10) - ends without issues 

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -3088,6 +3088,9 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):
         self.log.info('Test ID: {}'.format(self.test_config.test_id()))
         self._check_alive_routines_and_report_them()
         self._check_if_db_log_time_consistency_looks_good()
+        self.log.debug("Threads and processes at the end of the test:")
+        for t in threading.enumerate():
+            self.log.debug(f"Active thread: {t.name} (id={t.ident}, daemon={t.daemon}, repr={repr(t)})")
 
     @silence()
     def _check_if_db_log_time_consistency_looks_good(self):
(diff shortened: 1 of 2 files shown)


fruch:
@juliayakovlev I think this is a fix to other situations you reported before about HDR getting stuck...

vponomaryov (approved):
LGTM

fruch:
> @fruch , @soyacz Maybe we want to backport it to more branches than just `perf-v16`?

other branch doesn't use HDR, maybe v15, and even there the problem isn't just that, it's cause we remove some safety cleanup from test teardown on master. even on v16 I don't think we encountered the issue, only seen on master.

vponomaryov:
> > @fruch , @soyacz Maybe we want to backport it to more branches than just `perf-v16`?
> 
> other branch doesn't use HDR, maybe v15, and even there the problem isn't just that, it's cause we remove some safety cleanup from test teardown on master. even on v16 I don't think we encountered the issue, only seen on master.

My question was based on the general understanding that the root cause not exactly in the HDRH, but fact of getting some additional thread.
And we may get some new thread any time later...

But ok, if you think that probability of such a case doesn't worth backporting, then ok.

PR #11316, opened 2025-07-01
[Backport perf-v16] fix(libssh2): fix hanging ssh connection after the test

scylladbbot:
When test ends it may hang due awaiting ssh command completion, while
connection was closed (when finishing a test).
This happened due `tail -f` running forever against HDR Histogram log
file.

Fix by ending this `tail -f` command with `pkill`, so reader thread ends
normally and closes all ssh connections.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11258

### Testing

- [x] - [see jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/lukasz/job/perf-regression-predefined-throughput-steps-sanity-vnodes/10) - ends without issues 

### Reminders

- (cherry picked from commit 9ba505eacc7497b6988d83ee17f6a28653144500)

Parent PR: #11309


sdcm/tester.py
```diff
@@ -3031,11 +3031,18 @@ def tearDown(self):
         self.log.info('Test ID: {}'.format(self.test_config.test_id()))
         self._check_alive_routines_and_report_them()
         self._check_if_db_log_time_consistency_looks_good()
+<<<<<<< HEAD
         self.remove_python_exit_hooks()
```
fruch, line 3035: @soyacz 

I think this call is exactly why we didn't see this issue in v16

we should still backport the fix, but withut taking this out

Issue #11037, opened 2025-06-05
Cassandra-stress write load sometimes fails with WriteTimeoutException during preparing stage in performance throughput test


juliayakovlev:
## Packages

Scylla version: `2025.3.0~dev-20250603.6cbcabd10047` with build-id `0489eb8f37f00126beb5183a062e806fdc667272`
Kernel Version: `6.8.0-1029-aws`

## Issue description

- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

Cassandra-stress write load fails from time to time during preparing stage in performance throughput test.
```
2025-06-04 00:25:04.172: (CassandraStressLogEvent Severity.CRITICAL) period_type=one-time event_id=875382a0-0596-43a4-811d-1fd3fb371a0c: type=OperationOnKey regex=Operation x10 on key\(s\) \[ line_number=54190 node=Node perf-regression-predefined-steps-ub-loader-node-f1c8a617-3 [44.200.42.192 | 10.12.22.163]
java.io.IOException: Operation x10 on key(s) [303933504f4d364f4c30]: Error executing: (WriteTimeoutException): Cassandra timeout during SIMPLE write query at consistency ALL (3 replica were required but only 2 acknowledged the write)
```

The write load runs with CL = ALL and threads = 150 for each loader (4 loaders in test):
```
prepare_write_cmd: [
    "cassandra-stress write  cl=ALL n=162500001 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode connectionsPerHost=8 cql3 native -rate threads=150 -col 'size=FIXED(1024) n=FIXED(1)'  -pop seq=1..162500001",
    "cassandra-stress write  cl=ALL n=162500001 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode connectionsPerHost=8 cql3 native -rate threads=150 -col 'size=FIXED(1024) n=FIXED(1)'  -pop seq=162500002..325000002",
    "cassandra-stress write  cl=ALL n=162500001 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode connectionsPerHost=8 cql3 native -rate threads=150 -col 'size=FIXED(1024) n=FIXED(1)'  -pop seq=325000003..487500003",
[...]

PR #11354, opened 2025-07-10
fix(predefined-throughput-steps): throttle the prepare stage

fruch:
we are getting too many times into case we probably overloading the cluster this change is lowering a bit the number of threads, and throttle the load to 200K/s

so we expect to finish prepeare in one hour (more or less)

Fixes: scylladb/scylla-cluster-tests#11037

### Testing

- [x] https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/scylla-enterprise-perf-regression-predefined-throughput-steps-vnodes/58/

### Reminders

### DIFF
--- a/microbenchmarking_test.py
+++ b/microbenchmarking_test.py
@@ -13,18 +13,11 @@
 import json
 
 from sdcm.argus_results import send_perf_simple_query_result_to_argus
-from sdcm.tester import ClusterTester, teardown_on_exception, log_run_info
+from sdcm.tester import ClusterTester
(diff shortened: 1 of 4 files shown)


roydahan:
What worked for me was to change it to: "-rate threads=50".
The cluster total throughput was around 125-150K op/s and it took around 1h15m.
(Worked consistently for 6 runs)

soyacz:
maybe now with 1:1 restore, we could utilize it in perf tests?

soyacz:
> > maybe now with 1:1 restore, we could utilize it in perf tests?
> 
> you think it put the data into the 3 nodes faster ?

I'd need to see how much time it takes now and see benchmarks of 1:1 restore

PR #11371, opened 2025-07-10
feature(gemini): use new gemini version 2.0

CodeLieutenant:
- Enable Statement Logger on all workloads statements are logged now in Oracle Cluster, this removes the problem, performance bottlenck of logging to file and gives us better tracability

- Removing --fail-fast flag, this flag still exists in gemini but is not used for anything, every workload should migrate to --max-errors-to-store flags, which gives better control if and how gemini will fail

### Testing

- :green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/e1ea4694-6754-4e1f-9bde-48e987e03e2d (3 Hour Run with Nemesis)
- :green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/27e79388-87d6-409c-9f07-588eda98b19f (10 Hour Run with Nemesis)
- :green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/1d07f19a-495c-49e0-8b69-bb2ed0801490 (10 Hour Run with Nemesis different Seed)
- :green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/9563facb-5d36-48bf-9958-a8ca2f845155 (3 Hour Run without Nemesis)
- :green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/9c533fc0-4e1f-4ecd-9adf-46aa49a196e9 (Another 3 Hour Run Without nemesis

Currently there will be no backports of gemini the the release branches, as we need it to run for a month to iron out bugs that might (will) surface, after that point we will backport relevant release branches.

### Changes to the gemini flags

- Remove `--fail-fast` as explained above why
- Changed `partition-key-buffer-reuse-size` from `128` to `256` -> increases the throughput, and since we removed most (that we know of) memory leaks, there is not point in not using higher value here
- Added `io-worker-pool` -> number of IO Worker threads that will execute Scylla read/write requests (new flags from v2)
- `max-errors-to-store=N` is set by default to `1000` -> after which point (if 1000 errors happen -> read or write) gemini will fail, this does not mean gemini will be considered a success if there is <N error, even if one error is there, gemini will exit with status 1 indicating something is wrong, this just makes gemini work for longer periods of time, since not every error is gemini's fault

### Reminders

### DIFF
--- a/sdcm/gemini_thread.py
+++ b/sdcm/gemini_thread.py
@@ -83,7 +83,6 @@ class GeminiStressThread(DockerBasedStressThread):
             "test-host-selection-policy": "token-aware",
             "drop-schema": "true",
             "cql-features": "normal",
-            "fail-fast": "true",
             "materialized-views": "false",
             "use-server-timestamps": "true",
             "use-lwt": "false",
@@ -96,10 +95,11 @@ class GeminiStressThread(DockerBasedStressThread):
             "max-clustering-keys": 4,
             "min-clustering-keys": 2,
             "partition-key-distribution": "uniform",  # Distribution for hitting the partition
-            # These two are used to control the memory usage of Gemini
-            "token-range-slices": 10000,  # Number of partitions
-            "partition-key-buffer-reuse-size": 128,  # Internal Channel Size per partition value generation
+            "token-range-slices": 10000,
+            "partition-key-buffer-reuse-size": 128,
             "statement-log-file-compression": "zstd",
+            "io-worker-pool": 2048,  # Number of threads to perform IO operations (mainly Scylla reads/writes)
+            "max-errors-to-store": 1000,  # Number of error to make gemini fail, after N error, gemini will stop immediately with error
         }
 
         self.gemini_oracle_statements_file = f"gemini_oracle_statements_{self.unique_id}.log"
@@ -107,9 +107,15 @@ class GeminiStressThread(DockerBasedStressThread):
         self.gemini_result_file = f"gemini_result_{self.unique_id}.log"
 
     def _generate_gemini_command(self):
-        seed = self.params.get("gemini_seed") or random.randint(1, 100)
+        seed = self.params.get("gemini_seed")
+        if seed is None:
+            seed = random.randint(1, 100)
+
         table_options = self.params.get("gemini_table_options")
-        log_statements = self.params.get("gemini_log_cql_statements") or True
+        log_statements = self.params.get("gemini_log_cql_statements")
+        if log_statements is None:
+            log_statements = True
+
         test_nodes = ",".join(self.test_cluster.get_node_cql_ips())
 
         cmd = f"gemini \
--- a/test-cases/gemini/gemini-1tb-10h.yaml
+++ b/test-cases/gemini/gemini-1tb-10h.yaml
@@ -18,7 +18,7 @@ gemini_cmd: |
   --mode mixed
 
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
-gemini_log_cql_statements: false
+gemini_log_cql_statements: true
 
 db_type: mixed_scylla
 instance_type_db_oracle: 'i4i.16xlarge'
--- a/test-cases/gemini/gemini-basic-3h-ics.yaml
+++ b/test-cases/gemini/gemini-basic-3h-ics.yaml
@@ -16,7 +16,7 @@ gemini_cmd: |
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
 gemini_table_options:
   - "compaction={'class': 'IncrementalCompactionStrategy'}"
-gemini_log_cql_statements: false
+gemini_log_cql_statements: true
 
 db_type: mixed_scylla
 instance_type_db_oracle: 'i4i.8xlarge'
--- a/defaults/docker_images/gemini/values_gemini.yaml
+++ b/defaults/docker_images/gemini/values_gemini.yaml
@@ -1,2 +1,2 @@
 gemini:
-  image: scylladb/gemini:1.9.3
+  image: scylladb/gemini:2.0.0
--- a/test-cases/gemini/gemini-3h-cdc-postimage-write.yaml
+++ b/test-cases/gemini/gemini-3h-cdc-postimage-write.yaml
@@ -13,6 +13,6 @@ gemini_cmd: |
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
 gemini_table_options:
   - "cdc={'enabled': true, 'postimage': true}"
-gemini_log_cql_statements: false
+gemini_log_cql_statements: true
 
 db_type: scylla
--- a/test-cases/gemini/gemini-3h-cdc-preimage-write.yaml
+++ b/test-cases/gemini/gemini-3h-cdc-preimage-write.yaml
@@ -13,6 +13,6 @@ gemini_cmd: |
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
 gemini_table_options:
   - "cdc={'enabled': true, 'preimage': true}"
-gemini_log_cql_statements: false
+gemini_log_cql_statements: true
 
 db_type: scylla
--- a/test-cases/gemini/gemini-3h-cdc-write.yaml
+++ b/test-cases/gemini/gemini-3h-cdc-write.yaml
@@ -13,6 +13,6 @@ gemini_cmd: |
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
 gemini_table_options:
   - "cdc={'enabled': true}"
-gemini_log_cql_statements: false
+gemini_log_cql_statements: true
 
 db_type: scylla
--- a/test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
+++ b/test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
@@ -20,7 +20,7 @@ gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json'
 gemini_table_options:
   - "cdc={'enabled': true, 'preimage': true, 'postimage': true}"
   - "compaction={'class': 'IncrementalCompactionStrategy'}"
-gemini_log_cql_statements: false
+gemini_log_cql_statements: true
 
 stress_cdclog_reader_cmd: "cdc-stressor -duration 215m -stream-query-round-duration 30s"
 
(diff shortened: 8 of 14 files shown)


sdcm/gemini_thread.py
```diff
@@ -96,20 +95,27 @@ def __init__(self, test_cluster: BaseCluster | BaseScyllaCluster, oracle_cluster
...
 
         self.gemini_oracle_statements_file = f"gemini_oracle_statements_{self.unique_id}.log"
         self.gemini_test_statements_file = f"gemini_test_statements_{self.unique_id}.log"
         self.gemini_result_file = f"gemini_result_{self.unique_id}.log"
 
     def _generate_gemini_command(self):
-        seed = self.params.get("gemini_seed") or random.randint(1, 100)
+        seed = self.params.get("gemini_seed")
+        if seed is None:
+            seed = random.randint(0, 100)
```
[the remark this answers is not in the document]
fruch: This change adds the possibility of getting seed=0, I'm not sure why...
I think picking zero would be confusing all around ...

Originally it is limited to a specific set of seed cause there were too many issues with Gemini, that it was used to limit the scope. 

Not that it really helped.

sdcm/gemini_thread.py
```diff
@@ -96,20 +95,27 @@ def __init__(self, test_cluster: BaseCluster | BaseScyllaCluster, oracle_cluster
             "max-clustering-keys": 4,
             "min-clustering-keys": 2,
             "partition-key-distribution": "uniform",  # Distribution for hitting the partition
-            # These two are used to control the memory usage of Gemini
-            "token-range-slices": 10000,  # Number of partitions
-            "partition-key-buffer-reuse-size": 128,  # Internal Channel Size per partition value generation
+            "token-range-slices": 10000,
```
yarongilor, line 98: why not renaming to something like "partitions-number"?

sdcm/gemini_thread.py
```diff
@@ -96,20 +95,27 @@ def __init__(self, test_cluster: BaseCluster | BaseScyllaCluster, oracle_cluster
...
+            "partition-key-buffer-reuse-size": 256,
```
yarongilor, line 99: why not renaming to something human readable?

sdcm/gemini_thread.py
```diff
@@ -96,20 +95,27 @@ def __init__(self, test_cluster: BaseCluster | BaseScyllaCluster, oracle_cluster
...
             "statement-log-file-compression": "zstd",
+            "io-worker-pool": 2048,  # Number of threads to perform IO operations (mainly Scylla reads/writes)
+            "max-errors-to-store": 1000,  # Number of error to make gemini fail, after N error, gemini will stop immediately with error
```
yarongilor, line 102: if this value mainly for debugging, then it doesn't have to be the default. If the tool is stable, it's better to set the default as 1. isn't it like c-s?

yarongilor:
what's the point to keep `--fail-fast` flag if it does nothing?

yarongilor:
what about the flag of "run_with_gemini"? i think it should be an important goal to remove it.

test-cases/gemini/gemini-1tb-10h.yaml
```diff
@@ -18,7 +18,7 @@ gemini_cmd: |
   --mode mixed
```
yarongilor, line 18: If the tool is improved, then why leaving the same low throughput? or why throughput isn't improved?
The tests show results of < 2k (or 3k), why not setting a higher rate, like, for comparison, c-s 3h longevity which gets 30k?

yarongilor:
Can you please add a reference to Gemini repository corresponding PR and updated docs?

sdcm/gemini_thread.py
fruch: No it's not.

as you stated  it's not yet stable

and c-s is failing only after 10 failures by default. on the same operation , yet again it's not the same as gemini

sdcm/gemini_thread.py
fruch: not related to this PR

sdcm/gemini_thread.py
fruch: not related to this PR

fruch (changes requested):
Taking it back,

we shouldn't use seed=0

test-cases/gemini/gemini-1tb-10h.yaml
yarongilor: `Changed partition-key-buffer-reuse-size from 128 to 256 -> increases the throughput` sounds related, for example. The idea is we want to test a higher throughput. why staying with a poor throughput?

sdcm/gemini_thread.py
CodeLieutenant: Just storing errors, makes gemini work longer nothing else, e.g.
There is a failure in some nemesis. something goes wrong, couple queries fail, tool continues to work, reports those things as errors, so they get counted, at the end you will see when and why those queries failed, and then you can check other logs (nemesis, regular gemini log) to determine what when wrong and if it's a gemini issue or not.
This is not a RETRY (we have those, both mutation and validation retries -> also driver retries) and this is cumulative value for error happening across the run. And in many cases if there is a bug in gemini, this count will be filled in 1/10 of a second (10k req/s -> 1000 errors). If we dont have this kind of error, it will be really hard to see what happened. This count gives us a lot of perspective in how gemini operates, what bugs it could flush and if any of those are scylla bugs

sdcm/gemini_thread.py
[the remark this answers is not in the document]
CodeLieutenant: All those previous restriction are lifted, if gemini hangs for what-ever reason it will be consifered a bug in gemini and know on the seed (previously there were seeds where there were not enough values for primary keys and that could hang), all those restrictions are lifted.

And yes zero makes it go crazy for what every reason, something in Go's random number generator does not like 0 as seed

test-cases/gemini/gemini-basic-3h-ics.yaml
```diff
@@ -16,7 +16,7 @@ gemini_cmd: |
 gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json' # currently is not used
 gemini_table_options:
   - "compaction={'class': 'IncrementalCompactionStrategy'}"
-gemini_log_cql_statements: false
+gemini_log_cql_statements: true
```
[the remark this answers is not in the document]
fruch: for that we'll need to implement oracle support for both GCP and Azure, which quite a lot of work.
until then  that need to be tested carefully  before we can merge and backport

fruch:
@dimakr 
can you raise issue for the integration tests failing, for some reason I could reproduce it locally (or at all)

test-cases/gemini/gemini-basic-3h-ics.yaml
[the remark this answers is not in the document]
CodeLieutenant: It's already doing that, when there is no oracle hosts, it will disable the logging (basically never creates the logger)

soyacz:
@CodeLieutenant please rebase - there should be fix for c-s integration test

fruch:
@CodeLieutenant the integration test failed...

maybe it's about time to throw away that assert about the arguments ? it fails every single time you are doing gemini updates, and it's pointless

```
16:54:46  FAILED unit_tests/test_gemini_thread.py::test_01_gemini_thread - assert '--fail-fast=true' in 'gemini                 --test-cluster="172.17.0.2"                 --seed=0                 --schema-seed=0                 --profiling-port=6060                 --bind=0.0.0.0:2112                 --outfile=/gemini_result_82ab5f31-7404-482c-b5e5-a76ea7f0d22d.log                 --replication-strategy="{\'class\': \'NetworkTopologyStrategy\', \'replication_factor\': \'3\'}"                 --oracle-replication-strategy="{\'class\': \'NetworkTopologyStrategy\', \'replication_factor\': \'1\'}" --oracle-cluster="172.17.0.2" --test-statement-log-file=/gemini_test_statements_82ab5f31-7404-482c-b5e5-a76ea7f0d22d.log                     --oracle-statement-log-file=/gemini_oracle_statements_82ab5f31-7404-482c-b5e5-a76ea7f0d22d.log --test-username=cassandra                 --test-password=cassandra                 --oracle-username=cassandra                 --oracle-password=cassandra --table-options="gc_grace_seconds=60" --level=info --request-timeout=3s --connect-timeout=60s --consistency=QUORUM --async-objects-stabilization-backoff=10ms --async-objects-stabilization-attempts=10 --dataset-size=large --oracle-host-selection-policy=token-aware --test-host-selection-policy=token-aware --drop-schema=true --materialized-views=false --use-lwt=false --use-counters=false --max-tables=1 --max-columns=16 --min-columns=8 --max-partition-keys=6 --min-partition-keys=2 --max-clustering-keys=4 --min-clustering-keys=2 --partition-key-distribution=uniform --token-range-slices=10000 --partition-key-buffer-reuse-size=128 --statement-log-file-compression=zstd --io-worker-pool=2048 --max-errors-to-store=1000 --duration=1m --warmup=0 --concurrency=5 --mode=write --cql-features=basic --max-mutation-retries=100 --max-mutation-retries-backoff=100ms --replication-strategy="{\'class\': \'NetworkTopologyStrategy\', \'replication_factor\': \'1\'}" --table-options="cdc = {\'enabled\': true, \'ttl\': 0}" --use-server-timestamps=true'
```

CodeLieutenant:
> > > @CodeLieutenant the integration test failed...
> > > maybe it's about time to throw away that assert about the arguments ? it fails every single time you are doing gemini updates, and it's pointless
> > > ```
> > > 16:54:46  FAILED unit_tests/test_gemini_thread.py::test_01_gemini_thread - assert '--fail-fast=true' in 'gemini                 --test-cluster="172.17.0.2"                 --seed=0                 --schema-seed=0                 --profiling-port=6060                 --bind=0.0.0.0:2112                 --outfile=/gemini_result_82ab5f31-7404-482c-b5e5-a76ea7f0d22d.log                 --replication-strategy="{\'class\': \'NetworkTopologyStrategy\', \'replication_factor\': \'3\'}"                 --oracle-replication-strategy="{\'class\': \'NetworkTopologyStrategy\', \'replication_factor\': \'1\'}" --oracle-cluster="172.17.0.2" --test-statement-log-file=/gemini_test_statements_82ab5f31-7404-482c-b5e5-a76ea7f0d22d.log                     --oracle-statement-log-file=/gemini_oracle_statements_82ab5f31-7404-482c-b5e5-a76ea7f0d22d.log --test-username=cassandra                 --test-password=cassandra                 --oracle-username=cassandra                 --oracle-password=cassandra --table-options="gc_grace_seconds=60" --level=info --request-timeout=3s --connect-timeout=60s --consistency=QUORUM --async-objects-stabilization-backoff=10ms --async-objects-stabilization-attempts=10 --dataset-size=large --oracle-host-selection-policy=token-aware --test-host-selection-policy=token-aware --drop-schema=true --materialized-views=false --use-lwt=false --use-counters=false --max-tables=1 --max-columns=16 --min-columns=8 --max-partition-keys=6 --min-partition-keys=2 --max-clustering-keys=4 --min-clustering-keys=2 --partition-key-distribution=uniform --token-range-slices=10000 --partition-key-buffer-reuse-size=128 --statement-log-file-compression=zstd --io-worker-pool=2048 --max-errors-to-store=1000 --duration=1m --warmup=0 --concurrency=5 --mode=write --cql-features=basic --max-mutation-retries=100 --max-mutation-retries-backoff=100ms --replication-strategy="{\'class\': \'NetworkTopologyStrategy\', \'replication_factor\': \'1\'}" --table-options="cdc = {\'enabled\': true, \'ttl\': 0}" --use-server-timestamps=true'
> > > ```
> > 
> > 
> > everthing done
> 
> Hmm, integration tests still fail in CI...

fixed as it seems, everything is passing now

soyacz (approved):
LGTM

PR #11373, opened 2025-07-10
Z3 support - Artifact test

roydahan:
### Testing

- [ ]

### Reminders

### DIFF
--- a/test-cases/artifacts/gce-image.yaml
+++ b/test-cases/artifacts/gce-image.yaml
@@ -3,7 +3,7 @@ backtrace_decoding: false
 gce_instance_type_db: 'n2-standard-2'
 gce_root_disk_type_db: 'pd-ssd'
 root_disk_size_db: 50
-gce_n_local_ssd_disk_db: 16
+gce_n_local_ssd_disk_db: 0
 instance_provision: "spot"
 instance_provision_fallback_on_demand: true
 n_db_nodes: 1
--- a/jenkins-pipelines/oss/artifacts/artifacts-gce-image.jenkinsfile
+++ b/jenkins-pipelines/oss/artifacts/artifacts-gce-image.jenkinsfile
@@ -9,6 +9,7 @@ artifactsPipeline(
     test_config: 'test-cases/artifacts/gce-image.yaml',
     backend: 'gce',
     instance_provision: 'spot',
+    region: 'us-central1',
 
     timeout: [time: 45, unit: 'MINUTES'],
     post_behavior_db_nodes: 'destroy',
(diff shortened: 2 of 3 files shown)


test-cases/artifacts/gce-image.yaml
```diff
@@ -3,7 +3,7 @@ backtrace_decoding: false
 gce_instance_type_db: 'n2-standard-2'
 gce_root_disk_type_db: 'pd-ssd'
 root_disk_size_db: 50
-gce_n_local_ssd_disk_db: 16
+gce_n_local_ssd_disk_db: 0
```
roydahan, line 6: Since I want to add the z3 instance types to the same job with other instance types (n2), I need to think of a way to change this one.

roydahan: @soyacz maybe I can add it to the "if" in gce_utils?

fruch: So would suggest dropping the default for it, and putting it in the code like that

```python
import re

# mapping between GCE instance types and their number of local SSDs disk we can use for data storage
gce_instances = {
    "n2-highmem-8": 2,
    "n2-highmem-16": 4,
    "n2-highmem-32": 8,
    "n2-highmem-48": 16,
    "n2-highmem-64": 24,
    "n2-highmem-80": 24,
    "n2-highmem-96": 24,
    "n2-highmem-128": 24,
    "z3-.*": 0,
    "c3-.*": 0,
    "c4-.*": 0,
}


def test_01():
    list_gcp_instances = [
        "n2-highmem-8",
        "n2-highmem-16",
        "n2-highmem-32",
        "n2-highmem-48",
        "n2-highmem-64",
        "n2-highmem-80",
        "n2-highmem-96",
        "n2-highmem-128",
        "z3-highssd-16",
        "c3-highmem-16",
        "c4-highmem-16",
    ]
    for i in list_gcp_instances:
        for instance_type_regex, local_ssd_count in gce_instances.items():
            regex = re.compile(instance_type_regex)
            if regex.match(i):
                print(f"Instance Type: {i}, Local SSD Count: {local_ssd_count}")
```

test should still be able to set it base on their need of disks, but if you didn't each type would have default

fruch: you don't have to list them all, you can still have default, but since this default isn't going to be correct for some of the type, I just suggest to organize it in a way it would be clearer, and not build some specific code for each new supported type that should have different value  

either way you'll need to adapt the code on each new type that doesn't fit with the defaults.

roydahan:
Ready for review and passed testing.

fruch (approved):
LGTM

soyacz (approved):
LGTM

test-cases/artifacts/gce-image.yaml
vponomaryov: It is time to implement the per instance-type disk mapping, because of the following incompatibilities:
- https://github.com/scylladb/scylla-cluster-tests/issues/11684

PR #11379, opened 2025-07-12
build(deps): add dependency for azure kms keyvault to pyproject.toml

Lakshmipathi:
Add few packages for azure kms keyvault integration

### Testing

- [x]

### Reminders


pyproject.toml
```diff
@@ -7,7 +7,10 @@ dependencies = [
     "anyconfig==0.12.0",
     "awscli==1.29.4",
     "azure-identity==1.16.1",
+    "azure-keyvault-keys==4.11.0",
```
fruch, line 10: @Lakshmipathi 

2 things

1) always used the `uv sync` command so the uv.lock would be updated as well
2) seems like the label for hydra needs to be there when you open the PR (otherwise it would work on on the next push to the PR)

Issue #11374, opened 2025-07-11
NodeSetupFailed: Remove scylla-tools package


Lakshmipathi:
While trying to use `Canonical:ubuntu-24_04-lts:server-gen1:24.04.202507080` node setup failed because scylla-tools-core package requires python2.  similar issue reported while user tried to upgrade from Ubuntu-22 to 24. https://forum.scylladb.com/t/issues-in-scylla-installation/977 

```
09:23:49  E           E                   Command: 'sudo DEBIAN_FRONTEND=noninteractive apt-get -o DPkg::Lock::Timeout=120 -o Dpkg::Options::="--force-confold" -o Dpkg::Options::="--force-confdef" install -y scylla-tools'
09:23:49  E           E
09:23:49  E           E                   Exit code: 100
09:23:49  E           E
09:23:49  E           E                   Stdout:
09:23:49  E           E
09:23:49  E           E                   Reading state information...
09:23:49  E           E                   Some packages could not be installed. This may mean that you have
09:23:49  E           E                   requested an impossible situation or if you are using the unstable
09:23:49  E           E                   distribution that some required packages have not yet been created
09:23:49  E           E                   or been moved out of Incoming.
09:23:49  E           E                   The following information may help to resolve the situation:
09:23:49  E           E
09:23:49  E           E                   The following packages have unmet dependencies:
09:23:49  E           E                    scylla-tools-core : Depends: python (>= 2.7) but it is not installable or
09:23:49  E           E                                                 python2 but it is not installable
09:23:49  E           E
09:23:49  E           E                   Stderr:
09:23:49  E           E
09:23:49  E           E                   E: Unable to correct problems, you have held broken packages.
```

Job URL: https://jenkins.scylladb.com/job/scylla-staging/job/LakshmipathiGanapathi/job/per_table_hints/job/azure_longeviity/49/consoleFull

fruch:
@Lakshmipathi 

Is this happening in releases ? on master ? (The original issue with failing to install docker)

PR #11380, opened 2025-07-13
fix(loaders): remove scylla_repo_loader option

fruch:
* **fix(loaders): remove scylla_repo_loader option**
since we don't use `use_prepared_loaders` and about to deprecate it,
we also deprecating `scylla_repo_loader` which is used for installation
of scylla-tools which we don't use anymore

* **feature(loaders): update all loader to ubuntu 24.04 latest**
since the version being used are not supported anymore, and failing
to install offical docker packages.
we are updating all of the loader in all cloud and aligning them
to take latest release of ubunutu 24.04

* **feature(docker): move docker installations as need to cloud-init**
to speed things up for loaders, we are moving the installation of docker
service into the cloud-init part for the the loader nodes.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11374

### Testing

- [ ]

### Reminders


pehala:
Adding backport for 2024.2 as well: https://argus.scylladb.com/tests/scylla-cluster-tests/d24b3c66-0f63-439c-8215-cefeee146561

PR #11390, opened 2025-07-14
feature(azure): Add Azure Key Management Service (KMS) integration

Lakshmipathi:
- Implement KmsProvider class for Azure Key Vault management
  - Add Azure KMS configuration support to provisioner and VM provider
  - Add SCT configuration option 'enable_azure_kms' to enable EaR with managed identity

### Testing

- [ ]

### Reminders


sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class KmsProvider:
+    _resource_group_name: str
+    _region: str
+    _az: str
+    _azure_service: AzureService = AzureService()
+    MANAGED_IDENTITY_CONFIG = {
```
Lakshmipathi, line 28: I had issue with dynamically creating managed identity and assign it to SCT temp resource group. So created static managed identity for the VM. Should I move this to test-config  yaml or any other thoughts?

fruch: what is exactly the problem of generating everything within the SCT temp resource group ? what is the limitation that forces you to have static ?

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+        'resource_group': 'qa_azure_key_vault',
+        'identity_name': 'myManagedIdentity',
+        'principal_id': '0b5c8c2a-faa1-41d8-ab2d-e9605c2b8654'
+    }
+    SCT_SERVICE_PRINCIPAL_ID = "6c580092-c65c-4527-93d5-25177e10f54e"
```
fruch, line 33: what's that exactly and how it's different from the configuration above ?

sdcm/provision/azure/kms_provider.py
Lakshmipathi: Currently SCT (service principal) uses only minimal permissions, we need to upgrade it to higher permission (Microsoft.Authorization/roleAssignments/write, Microsoft.ManagedIdentity/userAssignedIdentities/write) if we need to do this dynamically. Also we need to add some sleep to handle the delay in dynamically added resource propagation. propagation is fine but main concern is increased expose to existing SCT principal.

sdcm/provision/azure/kms_provider.py
Lakshmipathi: keyvaults are dynamically created in the SCT temporary resource group where as `qa_azure_key_vault` keeps the static managed identity. SCT principal grants the managed identity access to keyvault

docs/configuration_options.md
```diff
@@ -3613,3 +3613,4 @@ Store adaptive timeout metrics in Argus. Disabled for performance tests only.
 **default:** True
 
 **type:** boolean
+
```
pehala, line 3616: Why?

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+import logging
+from dataclasses import dataclass
+
+from sdcm.utils.azure_utils import AzureService
+
+LOGGER = logging.getLogger(__name__)
+
+
+@dataclass
+class KmsProvider:
```
pehala, line 23: This is highly specific to Azure, I would name it as such to distinguish between providers.

sdcm/provision/azure/kms_provider.py
pehala: > see KeyStore, it should go to the keystore as json and retrieved from there

Agreed, this should definitely be in KeyStore, static resource definition should not be in code at all

sdcm/provision/azure/kms_provider.py
pehala: Again, please keep this in KeyStore

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+            ).result()
+            LOGGER.info(f"Created vault with existing identity access: {vault_name}")
+
+            # Create key
+            key_id = self._azure_service.create_vault_key(vault_uri=vault.properties.vault_uri, key_name="scylla-key")
+            vault_info = {'identity_id': self._get_managed_identity_id(), 'vault_uri': vault.properties.vault_uri,
+                          'key_name': 'scylla-key', 'key_uri': key_id}
+            LOGGER.info(f"vault_uri item : {vault_info}")
+            return vault_info
+        except Exception as e:  # noqa: BLE001
```
pehala, line 82: Could we catch more specific Exception, i.e. only Azure errors?

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+                                "certificates": ["get"]
+                            }
+                        }],
+                    }
+                }
+            ).result()
+            LOGGER.info(f"Created vault with existing identity access: {vault_name}")
+
+            # Create key
+            key_id = self._azure_service.create_vault_key(vault_uri=vault.properties.vault_uri, key_name="scylla-key")
```
pehala, line 77: Can key_name be static? What about multiple parallel runs?

sdcm/tester.py
```diff
@@ -862,6 +862,42 @@ def prepare_kms_host(self) -> None:
         self.params["append_scylla_yaml"] = append_scylla_yaml
         return None
 
+    def prepare_azure_kms(self) -> None:
+        cluster_backend = self.params.get('cluster_backend')
+        if cluster_backend == 'azure':
+            if self.params.get('enable_azure_kms'):
```
pehala, line 868: Could this be instead present as a default config for azure?

You set way too many things here, do we set the same values for AWS KMS as well?

sdcm/provision/azure/kms_provider.py
Lakshmipathi: Right, that makes sense, will change it appropriately. :+1:

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+        'principal_id': '0b5c8c2a-faa1-41d8-ab2d-e9605c2b8654'
+    }
+    SCT_SERVICE_PRINCIPAL_ID = "6c580092-c65c-4527-93d5-25177e10f54e"
+
+    def _get_managed_identity_id(self) -> str:
+        return f"/subscriptions/{self._azure_service.subscription_id}/resourcegroups/{self.MANAGED_IDENTITY_CONFIG['resource_group']}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{self.MANAGED_IDENTITY_CONFIG['identity_name']}"
+
+    def get_or_create_keyvault_and_identity(self):
+        """Create dynamic vault with existing identity"""
+        vault_name = f"scylla{self._resource_group_name.split('-')[1][:8]}kv"[:24]
```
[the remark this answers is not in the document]
Lakshmipathi: I thought we can have vault as a temporary resource and delete them at the end of tests. soft-deletion of resource
is an issue. Ok, let me create fixed vault per region. I think aws-kms uses something like, will check and avoid
dynamic vault creation per test.

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+            vault = self._azure_service.keyvault.vaults.begin_create_or_update(
+                resource_group_name=self._resource_group_name, vault_name=vault_name,
+                parameters={
+                    "location": self._region,
+                    "properties": {
+                        "tenant_id": self._azure_service.azure_credentials["tenant_id"],
+                        "sku": {"name": "standard", "family": "A"},
+                        "enabled_for_disk_encryption": True,
+                        "enable_rbac_authorization": False,
+                        "access_policies": [{
```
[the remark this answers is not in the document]
Lakshmipathi: Yes, access policy seems simpler. :+1:

sdcm/tester.py
```diff
@@ -862,6 +862,42 @@ def prepare_kms_host(self) -> None:
...
+                key_name = "scylla-key"
+                self.log.info(f"Preparing Azure KMS configuration for vault: {vault_url}")
+                # Add Azure KMS configuration to append_scylla_yaml
+                append_scylla_yaml = self.params.get("append_scylla_yaml") or {}
+                if "azure_hosts" not in append_scylla_yaml:
+                    append_scylla_yaml["azure_hosts"] = {}
+                append_scylla_yaml["azure_hosts"]["scylla-dynamic-vault"] = {
+                    'master_key': f"{vault_url}{key_name}",
+                    'key_cache_expiry': '900s',
+                    'key_cache_refresh': '1260s',
```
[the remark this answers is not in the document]
Lakshmipathi: Just want to test different values, yes, mostly user will be using default one, will update :+1:

sdcm/tester.py
fruch: see `prepare_kms_host` for the full logic

in general we enable KMS in AWS as long as it's possible (base on versions, and there a SCT config option to disable it

```python
        version_supports_kms = (self.params.is_enterprise and
                                ComparableScyllaVersion(self.params.scylla_version) >= '2023.1.3')
        backend_support_kms = self.params.get('cluster_backend') in ('aws',)
        kms_configured_in_sct = self.params.get('scylla_encryption_options')
        test_uses_oracle = self.params.get("db_type") == "mixed_scylla"
        should_enable_kms = (version_supports_kms and
                             backend_support_kms and
                             not kms_configured_in_sct and
                             not test_uses_oracle and
                             not self.params.get('enterprise_disable_kms'))
```

sdcm/utils/azure_utils.py
```diff
@@ -19,16 +19,19 @@
 from itertools import chain
 
 from azure.identity import ClientSecretCredential
+from azure.keyvault.keys import KeyClient
```
ndragazis, line 22: This is not currently installed. You need to add `azure-keyvault-keys` in the list of dependencies.

Lakshmipathi: These dependencies added via https://github.com/scylladb/scylla-cluster-tests/pull/11379/files

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+    _azure_service: AzureService = AzureService()
+    MANAGED_IDENTITY_CONFIG = {
+        'resource_group': 'qa_azure_key_vault',
+        'identity_name': 'myManagedIdentity',
+        'principal_id': '0b5c8c2a-faa1-41d8-ab2d-e9605c2b8654'
+    }
+    SCT_SERVICE_PRINCIPAL_ID = "6c580092-c65c-4527-93d5-25177e10f54e"
+
+    def _get_managed_identity_id(self) -> str:
+        return f"/subscriptions/{self._azure_service.subscription_id}/resourcegroups/{self.MANAGED_IDENTITY_CONFIG['resource_group']}/providers/Microsoft.ManagedIdentity/userAssignedIdentities/{self.MANAGED_IDENTITY_CONFIG['identity_name']}"
```
vponomaryov, line 36: Need to make it be more readable:
```
return (
    f"/subscriptions/{self._azure_service.subscription_id}"
    f"/resourcegroups/{self.MANAGED_IDENTITY_CONFIG['resource_group']}"
    f"/providers/Microsoft.ManagedIdentity"
    f"/userAssignedIdentities/{self.MANAGED_IDENTITY_CONFIG['identity_name']}"
)
```

sdcm/provision/azure/kms_provider.py
vponomaryov: +1

sdcm/provision/azure/kms_provider.py
vponomaryov: In case of AWS, we use aliases which are shared among regions.
And, for each region, each particular alias is assigned to a regional KMS key itself.
So, this `alias` feature usage allows us to use AWS KMS feature in multi-dc test setups.

in Azure case should be something similar.
Otherwise it would not be competitive with AWS.

sdcm/provision/azure/kms_provider.py
vponomaryov: +1 for KeyStore and, also, it looks as `region`-specific thing. Is it?

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+                            }
+                        }],
+                    }
+                }
+            ).result()
+            LOGGER.info(f"Created vault with existing identity access: {vault_name}")
+
+            # Create key
+            key_id = self._azure_service.create_vault_key(vault_uri=vault.properties.vault_uri, key_name="scylla-key")
+            vault_info = {'identity_id': self._get_managed_identity_id(), 'vault_uri': vault.properties.vault_uri,
```
vponomaryov, line 78: Same here, need to make it be more readable:
```
vault_info = {
    'identity_id': self._get_managed_identity_id(),
    'vault_uri': vault.properties.vault_uri,
    'key_name': 'scylla-key', 'key_uri': key_id,
}
```

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,84 @@
...
+# See LICENSE for more details.
+#
+# Copyright (c) 2025 ScyllaDB
+import logging
+from dataclasses import dataclass
+
+from sdcm.utils.azure_utils import AzureService
+
+LOGGER = logging.getLogger(__name__)
+
```
vponomaryov, line 20: What about key rotation like we have it for AWS KMS?

vponomaryov (changes requested):

sdcm/provision/azure/kms_provider.py
ndragazis: AFAIK aliases are an AWS-only feature. In fact, it's the only way to manually rotate a symmetric key in AWS.

In Azure (and GCP) there are no aliases, so we need to create new key versions and pay the price, literally ($1 per rotation).

> Otherwise it would not be competitive with AWS.

What do you mean? Competitive from a cost perspective? In the real world, key rotations on the KMS are expected to be infrequent (e.g., once a year).

Lakshmipathi: yes right, alias cant be used with azure. Azure KMS is costly compared to AWS KMS because it charges $1 per rotation where as AWS max $2 for initial two rotations then 10000 free rotations per key. Azure can't make older key version as primary so that we can transparently switch between key versions.

May be use a new SCT flag like `perform_azure_key_rotation` then run this rotation job weekly once rather than performing rotation for all azure jobs by default?  For daily runs, use pre-created keys (3 keys per region) without key rotation?

sdcm/provision/azure/kms_provider.py
Lakshmipathi: No, that is a main limitation with azure/gcp, they don't have alias for key like  aws.

sdcm/tester.py
Lakshmipathi: @vponomaryov ok, One thing: we don't update scylla yaml config after each rotation, is that right? aws alias to keys helps to avoid config after each rotation?

sdcm/provision/azure/kms_provider.py
vponomaryov: Is there a deletion time limitation for Azure KMS keys?
Does it backoff for 7+ days like AWS does?
If not (I doubt) then dynamic creation sounds reasonable where each test will use it's own key.

sdcm/sct_config.py
```diff
@@ -1439,6 +1439,9 @@ class SCTConfiguration(dict):
              help="The time interval in minutes which gets waited before the KMS key rotation happens."
                   " Applied when the AWS KMS service is configured to be used."),
 
+        dict(name="enable_azure_key_rotation", env="SCT_ENABLE_AZURE_KEY_ROTATION", type=boolean,
```
pehala, line 1442: what is the option we have for AWS for this?

test-cases/longevity/longevity-azure-kms-10gb-3h.yaml
```diff
@@ -0,0 +1,17 @@
+test_duration: 255
```
pehala, line 1: you can update existing testcase

sdcm/sct_config.py
Lakshmipathi: AWS does rotation for all jobs by default. Here we need to restrict Azure key rotation to specific jobs.

test-cases/longevity/longevity-azure-kms-10gb-3h.yaml
pehala: Also you should update default for azure to enable it in all tests

sdcm/sct_config.py
Lakshmipathi: Updated PR.

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,116 @@
...
+    def get_key_uri_for_test(cls, region: str, test_id: str) -> str:
+        kms_config = KeyStore().get_azure_kms_config()
+        vault_name = f"{kms_config['shared_vault_name']}-{region}"
+        vault_uri = f"https://{vault_name}.vault.azure.net/"
+        key_number = (hash(test_id) % 3) + 1
+        return f"{vault_uri}scylla-key-{key_number}"
+
+    def get_or_create_keyvault_and_identity(self, test_id: str):
+        """Use fixed vault with 3 pre-created keys"""
+        vault_name = f"{self._kms_config['shared_vault_name']}-{self._region}"
```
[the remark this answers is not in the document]
Lakshmipathi: Will move to a function.

sdcm/cluster.py
```diff
@@ -4790,6 +4792,31 @@ def _rotate_kms_key(kms_key_alias_name, kms_key_rotation_interval, db_cluster):
...
+        if self.params.get("cluster_backend") != 'azure':
+            return None
+
+        if not self.params.get("enable_kms_key_rotation"):
+            self.log.debug("Azure KMS key rotation is disabled")
+            return None
+
+        test_id = str(self.test_config.test_id())
+        azure_region_names = self.params.get('azure_region_name')
+        region = azure_region_names[0]
```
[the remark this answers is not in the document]
pehala: We want to make this a default and as such you need to handle all the cases. You dont need to support multi-dc with azure KMS yet but it needs to be clear we do not support it and it should fail fast if we detect it.
The jobs with multidc should also be changed so they do not use azure KMS

sdcm/sct_provision/instances_provider.py
```diff
@@ -66,6 +66,8 @@ def provision_sct_resources(params: SCTConfiguration, test_config: TestConfig, *
                                                              test_id=request.test_id,
                                                              region=request.region,
                                                              availability_zone=request.availability_zone,
+                                                             enable_azure_kms=not (params.get(
+                                                                 'enterprise_disable_kms') or False),
```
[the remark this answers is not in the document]
vponomaryov: Please, do not resolve comments while the appropriate code change is not uploaded into PR.
Local changes on your laptop don't count for `resolved`.

sdcm/provision/azure/virtual_machine_provider.py
```diff
@@ -80,6 +82,19 @@ def get_or_create(self, definitions: List[InstanceDefinition], nics_ids: List[st
                     }],
                 },
             }
+
+            if self._enable_azure_kms:
+                self._kms_provider = AzureKmsProvider(
+                    self._resource_group_name, self._region, self._az, self._azure_service)
+                # Extract test_id from resource group name
+                test_id = self._resource_group_name.split("SCT-")[-1][:36]
```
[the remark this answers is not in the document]
vponomaryov: Same here:
```
Please, do not resolve comments while the appropriate code change is not uploaded into PR.
Local changes on your laptop don't count for resolved.
```

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,116 @@
...
+
+    @property
+    def sct_service_principal_id(self):
+        return self._kms_config['sct_service_principal_id']
+
+    def _get_managed_identity_id(self) -> str:
+        return (
+            f"/subscriptions/{self._azure_service.subscription_id}"
+            f"/resourcegroups/{self.managed_identity_config['resource_group']}"
+            f"/providers/Microsoft.ManagedIdentity"
```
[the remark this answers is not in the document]
vponomaryov: Same here:
```
Please, do not resolve comments while the appropriate code change is not uploaded into PR.
Local changes on your laptop don't count for resolved.
```

sdcm/tester.py
```diff
@@ -885,6 +886,52 @@ def prepare_kms_host(self) -> None:
         self.params["append_scylla_yaml"] = append_scylla_yaml
         return None
 
+    def prepare_azure_kms(self) -> None:
+        version_supports_kms = (self.params.is_enterprise and
+                                ComparableScyllaVersion(self.params.scylla_version) >= '2023.1.3')
```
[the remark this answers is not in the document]
pehala: You are right, it was planned for 2025.3, but was postponed. So the correct version here is `2025.4.0`

sdcm/sct_provision/instances_provider.py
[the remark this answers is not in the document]
Lakshmipathi: I fixed/testing with another branch (due to [cloud-init](https://github.com/scylladb/scylla-cluster-tests/issues/11682) and pkg installs) and marking here to keep track of it. Now removed those changes and this PR updated.

sdcm/tester.py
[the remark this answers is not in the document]
Lakshmipathi: Okay, added it as `2025.4.0`

sdcm/provision/azure/kms_provider.py
Lakshmipathi: Yes, vault per region. Moved all these static values into keystore and use those values.

sdcm/cluster.py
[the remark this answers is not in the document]
Lakshmipathi: Fixed, no it detected in prepare_azure_kms() and quits.

sdcm/provision/azure/kms_provider.py
```diff
@@ -0,0 +1,116 @@
...
+    @classmethod
+    def get_key_uri_for_test(cls, region: str, test_id: str) -> str:
+        kms_config = KeyStore().get_azure_kms_config()
+        vault_name = f"{kms_config['shared_vault_name']}-{region}"
+        vault_uri = f"https://{vault_name}.vault.azure.net/"
+        key_number = (hash(test_id) % 3) + 1
+        return f"{vault_uri}scylla-key-{key_number}"
+
+    def get_or_create_keyvault_and_identity(self, test_id: str):
+        """Use fixed vault with 3 pre-created keys"""
```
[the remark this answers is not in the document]
Lakshmipathi: I followed `NUM_OF_KMS_KEYS` from aws kms to set number as 3. Now uses KeyStore's `num_of_keys` to create keys.

sdcm/cluster.py
```diff
@@ -4790,6 +4793,30 @@ def _rotate_kms_key(kms_key_alias_name, kms_key_rotation_interval, db_cluster):
...
+
+            time.sleep(self.params.get("kms_key_rotation_interval") * 60)
+            try:
+                key_uri = AzureKmsProvider.get_key_uri_for_test(region, test_id)
+                rotated_key_id = azure_service.rotate_vault_key(key_uri)
+                self.log.info(f"Azure KMS key rotated for test {test_id}: {rotated_key_id}")
+            except ResourceNotFoundError as e:
+                self.log.error(f"Azure KMS key not found for rotation: {e}")
+
+        threading.Thread(target=_rotate, daemon=True, name='AzureKmsRotationThread').start()
```
pehala, line 4816: when is it closed?

vponomaryov (changes requested):

sdcm/tester.py
```diff
@@ -885,6 +886,54 @@ def prepare_kms_host(self) -> None:
         self.params["append_scylla_yaml"] = append_scylla_yaml
         return None
 
+    def prepare_azure_kms(self) -> None:
+        version_supports_kms = ComparableScyllaVersion(self.params.scylla_version) >= '2025.4.0'
```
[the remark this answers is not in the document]
Lakshmipathi: Added.

sdcm/provision/azure/virtual_machine_provider.py
```diff
@@ -80,6 +84,19 @@ def get_or_create(self, definitions: List[InstanceDefinition], nics_ids: List[st
                     }],
                 },
             }
+
+            if self._enable_azure_kms:
+                self._kms_provider = AzureKmsProvider(
+                    self._resource_group_name, self._region, self._az, self._azure_service)
+                # Extract test_id from resource group name
+                test_id = self._resource_group_name.split(SCT_RESOURCE_GROUP_PREFIX)[-1][:36]
```
vponomaryov, line 92: This approach looks as unneeded implicit and unsafe way for getting what we have already in other attributes.
Both, `tester` and `test_config` modules have the `test_id` attr.

sdcm/sct_config.py
```diff
@@ -1440,7 +1440,12 @@ class SCTConfiguration(dict):
 
         dict(name="kms_key_rotation_interval", env="SCT_KMS_KEY_ROTATION_INTERVAL", type=int,
              help="The time interval in minutes which gets waited before the KMS key rotation happens."
-                  " Applied when the AWS KMS service is configured to be used."),
+                  " Applied when AWS KMS or Azure KMS service is configured to be used."
+                  " NOTE: Be aware that Azure Key rotations cost $1/rotation."),
+
+        dict(name="enable_kms_key_rotation", env="SCT_ENABLE_KMS_KEY_ROTATION", type=boolean,
+             help="Enable automatic KMS key rotation (supports Azure for now)"),
+        # TODO: AWS KMS needs to support the `enable_kms_key_rotation` config option
```
vponomaryov, line 1448: Please, move this comment above the `enable_kms_key_rotation` config option definition.

sdcm/utils/azure_utils.py
```diff
@@ -126,6 +129,33 @@ def all_regions(self) -> list[str]:
...
+            key_client = KeyClient(vault_url=vault_uri, credential=self.credential)
+            key = key_client.get_key(name=key_name)
+            return key
+        except ResourceNotFoundError:
+            return None
+
+    def rotate_vault_key(self, key_uri: str) -> str:
+        # Extract vault URI and key name from full key URI
+        # Format: https://vault-name.vault.azure.net/scylla-key-N
+        vault_uri = key_uri.split('scylla-key-')[0]
```
vponomaryov, line 152: These 2 can be combined into 1 line:
```
vault_uri, key_name = key_uri.split('scylla-key-')
```

sdcm/tester.py
```diff
@@ -885,6 +886,57 @@ def prepare_kms_host(self) -> None:
         self.params["append_scylla_yaml"] = append_scylla_yaml
         return None
 
+    def prepare_azure_kms(self) -> None:
+        scylla_version = self.params.scylla_version
+        if not scylla_version:
+            return None
+        version_supports_kms = ComparableScyllaVersion(self.params.scylla_version) >= '2025.4.0~dev'
```
vponomaryov, line 893: Why not using the short variant `scylla_version` which was defined on the line 890?
Or why defining the short variant then?

sdcm/tester.py
```diff
@@ -885,6 +886,57 @@ def prepare_kms_host(self) -> None:
...
+
+        test_id = str(self.test_config.test_id())
+        append_scylla_yaml = self.params.get("append_scylla_yaml") or {}
+        if "azure_hosts" not in append_scylla_yaml:
+            append_scylla_yaml["azure_hosts"] = {}
+
+        regions = self.params.get('azure_region_name')
+        if len(regions) > 1:
+            raise NotImplementedError("Azure KMS multi-dc support is not yet implemented")
+        region = regions[0]
```
vponomaryov, line 919: This line is redundant because you use it just once on the following line.
The `regions[0]` variant can be used in place.

vponomaryov, line 919: This line is redundant because you use it just once on the following line.
The `regions[0]` variant can be used in place.

sdcm/provision/azure/virtual_machine_provider.py
Lakshmipathi: > Ok, updated PR to use TestConfig for get test_id.

This caused circular dependency, reverted the change. Will check whether `tester` can be used.

sdcm/provision/azure/provisioner.py
```diff
@@ -37,10 +37,11 @@
 class AzureProvisioner(Provisioner):
     """Provides api for VM provisioning in Azure cloud, tuned for Scylla QA. """
 
-    def __init__(self, test_id: str, region: str, availability_zone: str,
+    def __init__(self, test_id: str, region: str, availability_zone: str, enable_azure_kms: bool = False,
```
vponomaryov, line 40: Remove the `enable_azure_kms` and start catching the `params` object.
Then use it to get the value for the `self._enable_azure_kms` attr.

unit_tests/provisioner/test_provision_sct_resources.py
```diff
@@ -60,9 +61,10 @@ def test_fallback_on_demand_when_spot_fails(fallback_on_demand, params, test_con
     fake_remoter.result_map = {r"sudo cloud-init status --wait": Result(stdout="..... \n status: done", stderr="nic", exited=0),
                                r"sudo cloud-init --version": Result(stdout="/bin/ 19.2-amzn", stderr="", exited=0),
                                r"ls /var/lib/sct/cloud-init": Result(stdout="done", exited=0)}
-    provision_sct_resources(params=params, test_config=test_config, azure_service=azure_service)
+    provision_sct_resources(params=params, test_config=test_config,
+                            azure_service=azure_service, enterprise_disable_kms=True)
     provisioner_eastus = provisioner_factory.create_provisioner(
-        backend="azure", test_id=params.get("test_id"), region="eastus", availability_zone="1", azure_service=azure_service)
+        backend="azure", test_id=params.get("test_id"), region="eastus", availability_zone="1", azure_service=azure_service, params=params, enterprise_disable_kms=True)
```
vponomaryov, line 67: Why the `enterprise_disable_kms=True` is provided explicitly?
Why not taken from params further as it is in the main functionality?

unit_tests/provisioner/test_provisioner.py
```diff
@@ -79,9 +79,9 @@ def definition(image_id, image_type):
     )
 
 
-@pytest.fixture(scope='module')
-def provisioner_params(test_id, region, azure_service):
-    return {"test_id": test_id, "region": region, "availability_zone": "a", "azure_service": azure_service}
+@pytest.fixture(scope='function')
```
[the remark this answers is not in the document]
pehala: You probably do not need params. Param is indeed function scope fixture but I do not see how it make sense here. You need True/False value for a test you dont need SCT configuration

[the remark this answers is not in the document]
Lakshmipathi: Updated PR.

sdcm/provision/azure/provisioner.py
```diff
@@ -38,9 +38,10 @@ class AzureProvisioner(Provisioner):
     """Provides api for VM provisioning in Azure cloud, tuned for Scylla QA. """
 
     def __init__(self, test_id: str, region: str, availability_zone: str,
-                 azure_service: AzureService = AzureService(), **_):
+                 azure_service: AzureService = AzureService(), **config):
         availability_zone = self._convert_az_to_zone(availability_zone)
         super().__init__(test_id, region, availability_zone)
+        self._enable_azure_kms = not config.get('enterprise_disable_kms') in (True, None)
```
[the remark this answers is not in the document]
Lakshmipathi: Updated PR.

vponomaryov:
@Lakshmipathi can you please re-test it using the latest PR state?

pehala (approved):

vponomaryov (approved):
LGTM

Issue #11415, opened 2025-07-16
scylla-2025.1/tier1/jepsen-all-test is triggered without a version


cezarmoise:
Example: https://jenkins.scylladb.com/job/scylla-2025.1/job/tier1/job/jepsen-all-test/15/ was triggered by https://jenkins.scylladb.com/job/scylla-2025.1/job/sct_triggers/job/tier1-aws-custom-time-trigger/12/
The triggering job has these parameters

<img width="433" height="640" alt="Image" src="https://github.com/user-attachments/assets/14c08283-cb59-4d61-a514-719c2c5c0bb6" />

Since that job wants either `scylla_version` or `scylla_repo`, it fails to start.

Compared to 2025.2
https://jenkins.scylladb.com/view/scylla-2025.2/job/scylla-2025.2/job/tier1/job/jepsen-all-test/15/ is triggered by https://jenkins.scylladb.com/job/scylla-2025.2/job/sct_triggers/job/tier1-aws-custom-time-trigger/10/

which has these parameters

<img width="263" height="651" alt="Image" src="https://github.com/user-attachments/assets/80e2bd3d-3cae-48c7-b1a5-940a83468d37" />

cezarmoise:
cc: @yaronkaikov @Annamikhlin

cezarmoise:
> same for `longevity-multidc-schema-topology-changes-12h-test` https://argus.scylladb.com/tests/scylla-cluster-tests/fff6ec46-af05-4e6a-be88-32f126e60311

Different issue. There the issue is that the regions and AZs are not configured correctly (should be 2 regions, 3 AZ, is a 1 and 1 respectively)

fruch:
> > same for `longevity-multidc-schema-topology-changes-12h-test` https://argus.scylladb.com/tests/scylla-cluster-tests/fff6ec46-af05-4e6a-be88-32f126e60311
> 
> Different issue. There the issue is that the regions and AZs are not configured correctly (should be 2 regions, 3 AZ, is a 1 and 1 respectively)

it's the same issue the the trigger job passes along AMI. while it should do `scylla_version: branch-2025.1:latest`

PR #11417, opened 2025-07-16
fix(ci-triggers): Fix tier1 oss CI triggers

vponomaryov:
Changes:
- Move the `jepsen-all-test` CI job from the AWS to GCE trigger file because it is GCE-only CI job.
- Make the `jepsen-all-test` trigger respect the `gce_image_db` trigger parameter.
- Make the `longevity-multidc-schema-topology-changes-12h-test` CI job use hardcoded `latest` value for release branch.
  It is needed because reusing the `scylla_version` or `ami_id` params
  we will get only single-dc values, which are not applicable here.
- Remove duplication of the `longevity-multidc-schema-topology-changes-12h-test` config from the
  single-dc config block.
-  Add `gce_image_db` pipeline parameter support to the `jepsen` pipeline.

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/11415

### Testing

- [ ]

### Reminders

### DIFF
--- a/jenkins-pipelines/oss/sct_triggers/tier1-aws-custom-time-trigger.xml
+++ b/jenkins-pipelines/oss/sct_triggers/tier1-aws-custom-time-trigger.xml
@@ -59,7 +59,7 @@ stress_duration=$stress_duration</properties>
               <textParamValueOnNewLine>false</textParamValueOnNewLine>
             </hudson.plugins.parameterizedtrigger.PredefinedBuildParameters>
           </configs>
-          <projects>../tier1/longevity-mv-si-4days-streaming-test,../tier1/longevity-schema-topology-changes-12h-test,../tier1/longevity-twcs-48h-test,../tier1/longevity-multidc-schema-topology-changes-12h-test,../tier1/longevity-50gb-3days-test,../tier1/longevity-150gb-asymmetric-cluster-12h-test</projects>
+          <projects>../tier1/longevity-mv-si-4days-streaming-test,../tier1/longevity-schema-topology-changes-12h-test,../tier1/longevity-twcs-48h-test,../tier1/longevity-50gb-3days-test,../tier1/longevity-150gb-asymmetric-cluster-12h-test</projects>
           <condition>SUCCESS</condition>
           <triggerWithNoParameters>false</triggerWithNoParameters>
           <triggerFromChildProjects>false</triggerFromChildProjects>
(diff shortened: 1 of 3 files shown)


jenkins-pipelines/oss/sct_triggers/tier1-aws-custom-time-trigger.xml
```diff
@@ -59,38 +59,23 @@ stress_duration=$stress_duration</properties>
...
-          <projects>../tier1/jepsen-all-test</projects>
-          <condition>SUCCESS</condition>
-          <triggerWithNoParameters>false</triggerWithNoParameters>
-          <triggerFromChildProjects>false</triggerFromChildProjects>
-        </hudson.plugins.parameterizedtrigger.BuildTriggerConfig>
-        <hudson.plugins.parameterizedtrigger.BuildTriggerConfig>
-          <configs>
-            <hudson.plugins.parameterizedtrigger.PredefinedBuildParameters>
-              <properties>scylla_version=$scylla_version
+              <properties>scylla_version=master:latest
```
vponomaryov, line 70: - Need to replace the `master:latest` with `branch-2025.3:latest` in the `2025.3` SCT branch.
- Need to replace the `master:latest` with `branch-2025.2:latest` in the `2025.2` SCT branch.
- Need to replace the `master:latest` with `branch-2025.1:latest` in the `2025.1` SCT branch.

fruch: Maybe we can find a better non manual for next releases ?

fruch: Anyhow those backport won't help without generating the job manually 

I would recommend doing this change manually in releases, regardless of backporting

fruch: We do have a templating option for those freestyle jobs that we can introduce the current SCT branch in, and use that ...

fruch: See

https://github.com/scylladb/scylla-cluster-tests/blob/master/utils/build_system/create_test_release_jobs.py#L80

vponomaryov: > Maybe we can find a better non manual for next releases ?

It will be inherited by the next releases.
What is the concern?

> Anyhow those backport won't help without generating the job manually
> 
> I would recommend doing this change manually in releases, regardless of backporting

The jepsen pipeline update is still required.

> We do have a templating option for those freestyle jobs that we can introduce the current SCT branch in, and use that ...
> See
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/master/utils/build_system/create_test_release_jobs.py#L80

So, you mean that following must work:
```
<properties>scylla_version=$sct_branch:latest
```
?

fruch: > > Maybe we can find a better non manual for next releases ?
> 
> It will be inherited by the next releases.
> What is the concern?
> 
> > Anyhow those backport won't help without generating the job manually
> > 
> > I would recommend doing this change manually in releases, regardless of backporting
> 
> The jepsen pipeline update is still required.
> 
> > We do have a templating option for those freestyle jobs that we can introduce the current SCT branch in, and use that ...
> > See
> > 
> > https://github.com/scylladb/scylla-cluster-tests/blob/master/utils/build_system/create_test_release_jobs.py#L80
> 
> So, you mean that following must work:
> ```
> <properties>scylla_version=$sct_branch:latest
> ```
> ?
> 
> 

More like `scylla_version=%(sct_branch)s:latest` since it's using old string formatting

But yes I think it can do the trick for next releases

I've used this trick in operator triggers to pass in the latest enterprise version as parameter.

vponomaryov: Done

Issue #11409, opened 2025-07-16
Jenkins Run SCT test step failed with " KeyError: 'connection'"


timtimb0t:
Jenkins run: https://jenkins.scylladb.com/job/scylla-master/job/tier1/job/longevity-1tb-5days-azure-test/80/

error itself:
```
05:28:39  sdcm/tester.py:3157: TestFailedByEvents
05:28:39  =================================== FAILURES ===================================
05:28:39  ________________________ LongevityTest.test_custom_time ________________________
05:28:39  
05:28:39  self = <longevity_test.LongevityTest testMethod=test_custom_time>
05:28:39  
05:28:39      def setUp(self):
05:28:39  >       super().setUp()
05:28:39  
05:28:39  longevity_test.py:50: 
05:28:39  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
05:28:39  sdcm/tester.py:198: in wrapper
05:28:39      return method(*args, **kwargs)
05:28:39  sdcm/utils/decorators.py:122: in inner
05:28:39      res = func(*args, **kwargs)
05:28:39  sdcm/tester.py:1067: in setUp
05:28:39      future.result()
05:28:39  /usr/local/lib/python3.10/concurrent/futures/_base.py:451: in result
05:28:39      return self.__get_result()
05:28:39  /usr/local/lib/python3.10/concurrent/futures/_base.py:403: in __get_result
05:28:39      raise self._exception
05:28:39  /usr/local/lib/python3.10/concurrent/futures/thread.py:58: in run
05:28:39      result = self.fn(*self.args, **self.kwargs)
05:28:39  sdcm/tester.py:1011: in _create_db_clusters
05:28:39      self.init_nodes(db_cluster=db_cluster)
05:28:39  sdcm/tester.py:767: in init_nodes
05:28:39      db_cluster.wait_for_init(node_list=db_cluster.nodes)
05:28:39  sdcm/cluster.py:4115: in wrapper
05:28:39      verify_node_setup_or_startup(start_time, setup_queue, setup_results)
05:28:39  _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
05:28:39  
[...]

PR #11418, opened 2025-07-17
fix(cqlshrc): update cqlshrc only once

dimakr:
This change moves updating cqshrc config to ClusterTester.setUp, to have it executed only once, before creation of cluster nodes.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11409

### Testing

- [x] :green_circle: [pr-provision tests with enabled encryption](https://argus.scylladb.com/tests/scylla-cluster-tests/99e17ecb-9fae-4fb5-bedc-0d6ee7ab0036)

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -11,6 +11,7 @@
 #
 # Copyright (c) 2016 ScyllaDB
 import shutil
+import configparser
 from collections import defaultdict
 from copy import deepcopy
 from concurrent.futures import ThreadPoolExecutor, as_completed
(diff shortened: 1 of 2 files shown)


sdcm/tester.py
```diff
@@ -982,6 +1003,9 @@ def setUp(self):  # noqa: PLR0912, PLR0915
         if (not self.params.get("cluster_backend").startswith("k8s") and
                 any([self.params.get('client_encrypt'), self.params.get('server_encrypt')])):
             create_ca(self.localhost)
+            if self.params.get('client_encrypt'):
+                cqlshrc_file = get_data_dir_path('ssl_conf', 'client', 'cqlshrc')
+                update_cqlshrc(cqlshrc_file, client_encrypt=self.params.get('client_encrypt'))
```
fruch, line 1008: if we do it only under this if, it's kind of weird to pass in `client_encrypt` as parameter 

i.e. there no case we'll call it with `false` now, so we can rename the function and remove the argument

PR #11427, opened 2025-07-17
test(perf-c4): initial setup for GCP c4 small perf tests

fruch:
this one introduce a setup for testing specficly `c4-highmem-16-lssd` which in now in preview.

### Testing

- [ ] read only - https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/c4-perf-regression-predefined-throughput-steps-vnodes-test/6/

### Reminders

### DIFF
--- a/configurations/performance/c4/c4-highmem-16-lssd.yaml
+++ b/configurations/performance/c4/c4-highmem-16-lssd.yaml
@@ -1,0 +1,9 @@
+gce_instance_type_db: 'c4-highmem-16-lssd'
+gce_root_disk_type_db: 'hyperdisk-balanced' # those doesn't persistent disks
+gce_n_local_ssd_disk_db: 2 # fpr some reason it forces to specify this value (should try 0)
+
+gce_instance_type_loader: 'c3-highcpu-22'
+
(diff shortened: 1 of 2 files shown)


configurations/performance/c4/c4-highmem-16-lssd.yaml
```diff
@@ -0,0 +1,9 @@
+gce_instance_type_db: 'c4-highmem-16-lssd'
+gce_root_disk_type_db: 'hyperdisk-balanced' # those doesn't persistent disks
+gce_n_local_ssd_disk_db: 2 # fpr some reason it forces to specify this value (should try 0)
```
soyacz, line 3: strangely, I didn't have to specify it (I used 0)

fruch: it works with both

when I use it with the default the error message said it should be 2, and with 2 it also worked

fruch:
@soyacz we can close this one, right ?

soyacz:
> @soyacz we can close this one, right ?

I think so - if we ever need that, we can reopen

PR #11459, opened 2025-07-21
Set cloud image to 4.11.0 scylladb-monitor-4-11-0-2025-07-21t09-39-20z

amnonh:
Grafana announced a couple of CVEs, new image was created with the only change is the grafana and prometheus versions


fruch (approved):
LGTM

PR #11501, opened 2025-07-23
fix(teardown): move the thread checks into fixtures

fruch:
as part of the refactor to use pytest as the test runner
the code checking the left processes/threads isn't running
really at the end of the test, and the info it's gathering is
misleading, since the event system isn't yet stopped.

### Testing

- [ ]

### Reminders


sdcm/tester.py
```diff
@@ -3172,18 +3171,29 @@ def fixture_event_system(self, setup_logging):
...
+        yield
+
+        self._check_alive_routines_and_report_them()
+        self._check_if_db_log_time_consistency_looks_good()
+        logging.debug("Threads and processes at the end of the test:")
+        for t in threading.enumerate():
+            logging.debug(f"Active thread: {t.name} (id={t.ident}, daemon={t.daemon}, repr={repr(t)})")
+
     @silence()
     def _check_if_db_log_time_consistency_looks_good(self):
```
soyacz, line 3186: It's been some time since we had issues with log time consistency. maybe not needed anymore?

sdcm/tester.py
```diff
@@ -3004,6 +3003,11 @@ def destroy_credentials(self):
                 credential.destroy()
             self.credentials = []
 
+    @classmethod
+    @property
+    def left_processes_log(cls):
```
soyacz, line 3008: why do we need that change?

sdcm/tester.py
```diff
@@ -3172,18 +3171,29 @@ def fixture_event_system(self, setup_logging):
         yield
         self.stop_event_device()
 
+    @pytest.fixture(autouse=True, name='threads_checks', scope='session')
+    def fixture_threads_checks(self):
```
soyacz, line 3175: shouldn't be `self fixture_threads_checks(self, event_system):` to make sure it runs last?

fruch: the scope make it run last, event_system is on function scope, you can't mix fixture of different scopes

fruch:
@soyacz 
the 2nd commit 269b052136a83bb5b6513d66f757f9b29bc7efa0

is the actual fix for the `log_transport: ssh` (that is used in only one test config now, won't drop cause it could be very useful for debugging)

soyacz (approved):
LGTM

PR #11514, opened 2025-07-24
improvement(provision_resources): Handle provision errors for Argus

### LINKED ISSUE #11204: In case capacity issue, change test status to "TEST-ERROR"
If a test fails to start due to a capacity issue, its current "FAILED" status is misleading. We should change this to "TEST-ERROR" to accurately reflect that the test could not be executed, rather than that it produced an incorrect result.

Furthermore, if a capacity error occurs during a nemesis test, the test status should be determined by whether any other errors, beyond the capacity issue, were encountered.

k0machi:
Since we don't handle errors that happen during provisioning via the
Event system Argus doesn't really know what happened in case of capacity
issues during provisioning stage. This commit addresses this by setting
Argus status as "TEST ERROR" in case of any error during the
provisioning.

Fixes #11204

### Testing

- [ ] [Jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/alexey-argus-testing/133/)

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -235,13 +235,19 @@ def provision_resources(backend, test_name: str, config: str):
         click.echo("No need provision logging service")
 
     click.echo(f"Provision {backend} cloud resources")
-    if backend == "aws":
-        layout = SCTProvisionLayout(params=params)
-        layout.provision()
-    elif backend == "azure":
-        provision_sct_resources(params=params, test_config=test_config)
-    else:
-        raise ValueError(f"backend {backend} is not supported")
+    try:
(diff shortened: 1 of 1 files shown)


soyacz (approved):
LGTM

vponomaryov (approved):
LGTM

soyacz:
IIRC we have code in SCT that handle capacity issues and proper reporting to Argus. This was missing part when provision step failed (as we don't get to SCT run nowdays). What are other cases?

PR #11525, opened 2025-07-25
improvement(coredump): add logging if coredumpctl fails

cezarmoise:
If coredumpctl fails for whatever reasson, `_get_core_by_pid` will return `None` and updating the core info will fail.
Add extra logging.

Fixes https://argus.scylladb.com/tests/scylla-cluster-tests/dca84a90-29fb-4c72-8414-b8f5eff01a66
```
2025-07-24 23:58:31.514: (ThreadFailedEvent Severity.ERROR) period_type=one-time event_id=9c1c2fbe-22a6-45b6-b66d-552452d8b500: message='NoneType' object has no attribute 'get'
File "/home/ubuntu/scylla-cluster-tests/sdcm/sct_events/decorators.py", line 26, in wrapper
return func(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/coredump.py", line 120, in run
self.main_cycle_body()
~~~~~~~~~~~~~~~~~~~~^^
File "/home/ubuntu/scylla-cluster-tests/sdcm/coredump.py", line 133, in main_cycle_body
new_cores = self.extract_info_from_core_pids(self.get_list_of_cores(), exclude_cores=self.found)
File "/home/ubuntu/scylla-cluster-tests/sdcm/coredump.py", line 204, in extract_info_from_core_pids
self.update_new_coredump_with_exec_information(new_core_info)
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
File "/home/ubuntu/scylla-cluster-tests/sdcm/coredump.py", line 334, in update_new_coredump_with_exec_information
core_info.update(executable=core.get('exe', r'N\A'),
^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'get'
```

### Reminders

### DIFF
--- a/sdcm/coredump.py
+++ b/sdcm/coredump.py
@@ -306,6 +306,12 @@ class CoredumpThreadBase(Thread):
     def _get_core_by_pid(self, pid: str) -> Optional[dict]:
         result = self.node.remoter.sudo(f"coredumpctl list {pid} -q --json=short", verbose=False, ignore_status=True)
         if not result.ok:
+            self.log.warning(
+                f"""Failed to get coredump info for pid {pid}:
+                Stdout:
+                {result.stdout}
+                Stderr:
+                {result.stderr}""")
             return None
 
         try:
@@ -330,9 +336,9 @@ class CoredumpThreadBase(Thread):
         return self._extract_version(release_version)
 
     def update_new_coredump_with_exec_information(self, core_info: CoreDumpInfo) -> None:
-        core = self._get_core_by_pid(core_info.pid)
-        core_info.update(executable=core.get('exe', r'N\A'),
-                         executable_version=self._get_executable_version(core.get('exe', r'N\A')))
+        if core := self._get_core_by_pid(core_info.pid):
+            core_info.update(executable=core.get('exe', r'N\A'),
+                             executable_version=self._get_executable_version(core.get('exe', r'N\A')))
 
     @staticmethod
     def _extract_version(release_version: str) -> Optional[str]:


sdcm/coredump.py
```diff
@@ -330,9 +330,9 @@ def _get_executable_version(self, executable: str) -> Optional[str]:
         return self._extract_version(release_version)
 
     def update_new_coredump_with_exec_information(self, core_info: CoreDumpInfo) -> None:
-        core = self._get_core_by_pid(core_info.pid)
```
fruch, line 333: this is now fixing anything, this is hiding away that there a problem
the main question here is why `/usr/bin/coredumpctl list 25261 -q --json=short` failed

we should add debug information for those cases, and not hide it like that.

hiding it means we could missing a coredump happening in the system.

fruch (changes requested):
we should add debug information first, before we can decide this can be ignored like that

sdcm/coredump.py
cezarmoise: Added some debug logging.
`verbose=True` would not work because when `ignore_status=True`, it will only print debug logs on success but not on failure

```diff
diff --git a/sdcm/coredump.py b/sdcm/coredump.py
index 5dabb489f..fd36c3104 100644
--- a/sdcm/coredump.py
+++ b/sdcm/coredump.py
@@ -304,8 +304,14 @@ class CoredumpThreadBase(Thread):
         pass
 
     def _get_core_by_pid(self, pid: str) -> Optional[dict]:
        result = self.node.remoter.sudo(f"coredumpctl list {pid} -q --json=short", verbose=False, ignore_status=True)
         if not result.ok:
+            self.log.warning(
+                f"""Failed to get coredump info for pid {pid}:
+                Stdout:
+                {result.stdout}
+                Stderr:
+                {result.stderr}""")
             return None
```

Issue #9909, opened 2025-01-23
Add tests for SstableUtils.corrupt_sstable


soyacz:
`SstableUtils` got new feature of corrupting sstables - used for testing actions that need to be taken by SCT when this event occurs (or sstable validator finds an issue). This feature could be also used for testing Scylla, but it's not planned yet. As it's not currently used, we should test it in integration test (possibly testing also sending corrupted sstables to s3 to protect from s3 client breaking changes).

refs: https://github.com/scylladb/scylla-cluster-tests/pull/9900

PR #11533, opened 2025-07-27
test(upload_remote_files_directly_to_s3): integration test

fruch:
Introduce a new test that upload 100Mb via `upload_remote_files_directly_to_s3` implemetion

Closes: #9909

### Testing

- [x] tested localy
- [ ] tested in CI

### Reminders


unit_tests/test_s3_remote_uploader.py
```diff
@@ -0,0 +1,141 @@
...
+# Constants for the test
+TEST_S3_BUCKET = S3Storage.bucket_name
+TEST_S3_KEY_PREFIX = os.environ.get("TEST_S3_KEY_PREFIX", "integration-test/")
+SSH_IMAGE = "linuxserver/openssh-server:latest"
+SSH_USER = "sct-user"  # linuxserver/openssh-server default user
+SSH_PORT = 2222
+
+
+@pytest.fixture(scope="module")
+def ssh_docker_server(tmp_path_factory):
```
fruch, line 22: this should be move to shared location, since more test that uses SSH might need this for testing

unit_tests/test_s3_remote_uploader.py
```diff
@@ -0,0 +1,143 @@
...
+    inspect_cmd = [
+        "docker", "inspect", "-f",
+        "{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}",
+        container_name
+    ]
+    docker_address = subprocess.run(inspect_cmd, check=True, stdout=subprocess.PIPE).stdout.decode().strip()
+    # Wait for SSH to be up
+    for _ in range(20):
+        try:
+            os.environ.pop("SSH_AUTH_SOCK", None)  # Ensure no SSH agent interference
```
pehala, line 65: how about using monkeypatch instead of accessing os.environ directly?

pehala: Sorry, I do not follow,

I meant using https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-environment-variables, which should be isolated to not affect the rest of the tests. I know we are currently using os.environ directly in tests (potential improvement for later), but I would not introduce new usages if the monkeypatch will work here

PR #11537, opened 2025-07-28
fix(trigger): not pass unnecessary params

juliayakovlev:
'new_scylla_repo' and 'base_versions' should be passed to rolling upgrade job only.
Those params have to be ampty for all other jobs. Otherwise scylla_version will be replace with wrong value

### Testing

- [ ]

### Reminders


vars/perfRegressionParallelPipelinebyRegion.groovy
```diff
@@ -200,9 +200,9 @@ def call(Map pipelineParams) {
                                     println("Building job: $job_name with sub_test: ${sub_tests}, region: ${region}")
                                         build job: job_name, wait: false, parameters: [
                                             string(name: 'scylla_version', value: params.scylla_version),
-                                            string(name: 'base_versions', value: params.base_versions),
+                                            string(name: 'base_versions', value: job_name.contains("rolling-upgrade") ? params.base_versions : null),
```
fruch, line 203: I think it would be better if we can have a flag on top of the configuration of this job, saying it's upgrade, and not count on the name of the job

those kind of hacks comes back and bite

juliayakovlev: changed

Issue #11534, opened 2025-07-27
Gemini does not expose `gemini_cql_requests` metrics


soyacz:
After update of gemini to 2.1.0 (or 2.0.0) upgrade tests stopped working due missing `gemini_cql_requests` metric.
example run: https://argus.scylladb.com/tests/scylla-cluster-tests/863c2460-b936-4e9d-b95f-c707ec2417f8
(see error.log from events log package)

fruch:
we should adapt the ingratiation test to check for this, it's the 2nd time in we are breaking this part...

soyacz:
possibly need to adjust 'metric_has_data` to take into account all new variations

fruch:
> @fruch @soyacz Do we really need to check for metrics if there are, this looks like really fragile way of seeing if gemini's is running or not, a subtle change in metrics will distrupt this flow and kill it

Renaming metrics isn't a subtle change

We must check it's working before we can move to doing the upgrade, otherwise Gemini can fail on some of its first steps, when a node is stopped.

We had similar issues with c-s, and the best approach was to use metrics to validate stress have started writing data.

If you want to suggest a different approach, that would be great, after we fix the current one.

Again changing names of matrics, is breaking change, and it's not the first time changes to gemini breaking metrics for this specific test

PR #11551, opened 2025-07-29
fix(gemini): metrics not beining read properly after 2.0

CodeLieutenant:
In 2.0 gemini, more thing were added to `gemini_cql_requests` metrics, mainly the `method` which contained what statement type is executed. This broke the upgrade tests, as in upgrade test the first value returned from prometheus is used to check `> 0`, but the prometheus ordered the metric labels in alphabetical order, making the metric currently checked to be *almost* always zero, thus failing the check and killing gemini thinking it's not running (but it actually is).

Now fetching the `InsertStatement` label makes sure that this check will succeed (maybe not on the first try as it is not yet inserting, but after a couple of try will succeed).

Closing #11534 

### Testing

- :green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/bb8ff3e6-cc34-4d4a-ad2c-94e31fb6896b (Previous check of the metrics -> not backwards compatible)
- :green_circle: https://argus.scylladb.com/tests/scylla-cluster-tests/da0ee7be-90c6-49af-84fd-4fffc034e0d2 (backwards compatible way of checking)

### Reminders

### DIFF
--- a/test-cases/upgrades/rolling-upgrade.yaml
+++ b/test-cases/upgrades/rolling-upgrade.yaml
@@ -44,6 +44,4 @@ gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json'
 
 use_preinstalled_scylla: false
 
-stress_image:
-  gemini: 'scylladb/gemini:1.9.3'
 gemini_log_cql_statements: false
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -657,7 +657,7 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
             gemini_cmd += " --table-options \"cdc={'enabled': true}\""
         gemini_thread = self.run_gemini(gemini_cmd)
         self.metric_has_data(
-            metric_query='gemini_cql_requests', n=10)
+            metric_query='sum(increase(gemini_cql_requests[1m]))', n=10)
 
         with ignore_upgrade_schema_errors():
 


upgrade_test.py
```diff
@@ -657,7 +657,7 @@ def test_rolling_upgrade(self):  # noqa: PLR0914, PLR0915
             gemini_cmd += " --table-options \"cdc={'enabled': true}\""
         gemini_thread = self.run_gemini(gemini_cmd)
         self.metric_has_data(
-            metric_query='gemini_cql_requests', n=10)
+            metric_query='gemini_cql_requests{method="InsertStatement"}', n=10)
```
fruch, line 660: Is that backward compatible? I.e would work with older Gemini versions?

test-cases/upgrades/rolling-upgrade.yaml
```diff
@@ -45,5 +45,6 @@ gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json'
 use_preinstalled_scylla: false
 
 stress_image:
-  gemini: 'scylladb/gemini:1.9.3'
+  gemini: 'scylladb/gemini:2.1.0'
```
fruch, line 48: Let's remove it from here if it solves the issue

CodeLieutenant: > maybe change the function to count all of the returned values would work in a backward compatible way ?

I was thinking exactly the same, but im not sure if `sum gemini_cql_metrics` will work in promql cause of the amount of data generated in buckets for the counter, this one is getting only the latest, and its really hard to test this part cause everything is needed here, prometheus, large amounts of data, gemini and scylla.

Im trying something like sum the last minute and and not pull everything from prometheus

upgrade_test.py
CodeLieutenant: So what should we do, backwards compatible change or leave it like this?

CodeLieutenant:
@fruch @soyacz I'm setting upgrade test as passing since it is passing from gemini's side, it's failing cause of the known issue with `reactor stalled` which (I think) already exists and issue and PR fix

CodeLieutenant:
@fruch @soyacz Should I remove the upgrade to gemini to 2.1 in this PR, or we will revert the previous PR that downgraded?

test-cases/upgrades/rolling-upgrade.yaml
fruch: I've meant sum it in python code, no in promql, in `is_metric_has_data`

upgrade_test.py
CodeLieutenant: What last, the `gemini_cql_requests{method="InsertStatement"}` or `sum(increase(gemini_cql_requests[1m]))` ?

CodeLieutenant: https://argus.scylladb.com/tests/scylla-cluster-tests/da0ee7be-90c6-49af-84fd-4fffc034e0d2 this is the new run

test-cases/upgrades/rolling-upgrade.yaml
```diff
@@ -45,5 +45,6 @@ gemini_schema_url: 'https://s3.amazonaws.com/scylla-gemini/Binaries/schema.json'
 use_preinstalled_scylla: false
 
 stress_image:
```
soyacz, line 47: this is no longer needed

fruch (approved):

PR #11626, opened 2025-08-05
fix(test_longevity): Rewrite the test to not use pytester

### LINKED ISSUE #11625: Do not depend on internal data in test_longevity.py
`test_longevity.py` uses `test-cases/scale/longevity-5000-tables.yaml`, which depends on private CS profile.
We should remove this dependency, all unit tests should be runnable with main repository only.

pehala:
* Change `params` to use `monkeypatch`
   * Removes the need of a cleanup, `monkeypatch` should handle it automatically (https://docs.pytest.org/en/stable/how-to/monkeypatch.html#monkeypatching-environment-variables)
* Rewrite `test_longevity.py` to not use pytester
* Change config to `test-cases/features/elasticity/longevity-elasticity-many-small-tables.yaml`
   * Fixes #11625

### Testing

- [x] Locally
- [x] CI

### Reminders


unit_tests/longevity/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+    with unittest.mock.patch("sdcm.tester.validate_raft_on_nodes"):
+        yield
+
+    # clear the events processes registry after each test, so next test would be ableto start it fresh
+    events_processes._EVENTS_PROCESSES = None
+
+
+@pytest.fixture(scope="function", autouse=True)
+def params(request: pytest.FixtureRequest, monkeypatch):
+    monkeypatch.delenv("HOME", raising=False)
```
fruch, line 20: Those can be removed, it was introduced cause of using pytester

unit_tests/longevity/conftest.py
```diff
@@ -0,0 +1,30 @@
...
+from sdcm.sct_events import events_processes
+
+
+@pytest.fixture(scope="function", autouse=True)
+def fixture_mock_calls(monkeypatch):
+    with unittest.mock.patch("sdcm.tester.validate_raft_on_nodes"):
+        yield
+
+    # clear the events processes registry after each test, so next test would be ableto start it fresh
+    events_processes._EVENTS_PROCESSES = None
```
fruch, line 15: Not sure this would be needed as well

unit_tests/longevity/test_longevity.py
```diff
@@ -0,0 +1,74 @@
...
+
+from longevity_test import LongevityTest
+from sdcm.cluster import NoMonitorSet
+from unit_tests.test_utils_common import DummyDbCluster, DummyNode
+from unit_tests.test_cluster import DummyDbCluster
+
+LongevityTest.__test__ = False
+
+
+@pytest.mark.sct_config(files='test-cases/features/elasticity/longevity-elasticity-many-small-tables.yaml')
```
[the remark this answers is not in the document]
pehala: > It's not exactly equivalent and covers the exact set of things the 5000 case is.

What is missing? For the purposes of the test I think they are equivalent, but maybe I am missing purpose of this test



> Also I see no problem with some tests depending on internal repos, it's something that should be tested as well
> 
> And soon we'll have more of those in regards of scylla-cloud

Then those tests should make it clear they are testing internal resources, this test is called "test_longevity", nothing indicates that purpose of this test is to test the exact scale config.

If we want to test internal resources they should be skipped correctly as well.

unit_tests/longevity/conftest.py
pehala: Removed the fixture and instead rely on the same fixture in root conftest.py

unit_tests/longevity/test_longevity.py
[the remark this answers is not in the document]
fruch: there one difference:
```yaml
add_cs_user_profiles_extra_tables: true
```

I don't think we gonna introduce tests for the internal repo, the part that is test is not the configuration, but the implementation of SCT using that configuration

other option is to create a test only configuration which won't point to the internal repo.

regardless test was add to help catch issues in `test_user_batch_custom_time` as the test name suggests, where exactly is the configuration isn't really big of a difference

Issue #11546, opened 2025-07-29
Log collection fails with: cat: /home/ubuntu/sct-results/latest/test_id: No such file or directory


yarongilor:
## Packages

Scylla version: `2024.1.19-20250707.29ea1127f616` with build-id `3aaf8e6620e01a36c4c4c421271515563cd4efa5`

Kernel Version: `5.15.0-1087-aws`

## Issue description
A test using private branch passes ok, then fails in the pipeline steps of `Collect log data` and `Send email with result` with an error of:
```
10:33:55  < t:2025-07-28 07:33:55,328 f:dedicated_host.py l:80   c:sdcm.provision.aws.dedicated_host p:INFO  > Dedicated hosts is not enabled. Skipping reservation phase.
10:33:55  Dedicated hosts is not enabled. Skipping reservation phase.
10:33:55  < t:2025-07-28 07:33:55,341 f:common.py       l:1990 c:utils                p:ERROR > test_id not found. Exit code: 1; Error details cat: /home/ubuntu/sct-results/latest/test_id: No such file or directory
10:33:55  
10:33:55  test_id not found. Exit code: 1; Error details cat: /home/ubuntu/sct-results/latest/test_id: No such file or directory
10:33:55  
10:33:55  < t:2025-07-28 07:33:55,342 f:logcollector.py l:1676 c:sdcm.logcollector    p:WARNING > No test_id provided or found
10:33:55  No test_id provided or found
10:33:55  +---------------------------------+
10:33:55  | Collected logs by test-id: None |
10:33:55  +----------------------+----------+
10:33:55  | Cluster set          | Link     |
10:33:55  +----------------------+----------+
10:33:55  +----------------------+----------+
10:33:55  < t:2025-07-28 07:33:55,342 f:sct_runner.py   l:1231 c:sdcm.sct_runner      p:INFO  > Test runner ip in update_sct_runner_tags: None; test_id: None
10:33:55  Test runner ip in update_sct_runner_tags: None; test_id: None
10:33:55  < t:2025-07-28 07:33:55,342 f:log.py          l:25   c:sdcm.utils.log       p:ERROR > Uncaught exception < t:2025-07-28 07:33:55,342 f:log.py          l:25   c:sdcm.utils.log       p:ERROR > Uncaught exception
10:33:55  < t:2025-07-28 07:33:55,342 f:log.py          l:25   c:sdcm.utils.log       p:ERROR > Traceback (most recent call last):
[...]

PR #11632, opened 2025-08-05
fix(logdir import): Move create-logging and log-directory to a fixture

yarongilor:
Since nemesis.py importing of sdcm.mgmt.operations also imports tester.py, it causes re-linking sct logdir to a wrong directory of other pipeline step. Avoiding that by using a conftest fixture that is executed only once per test in a session level. 
Fixes: #11546

### Testing

Tested in:
[ubuntu24-installation-test](https://argus.scylladb.com/tests/scylla-cluster-tests/86d55f48-f74a-4c23-ad4c-de371d97adf5
)
[ubuntu-24-upgrade](https://argus.scylladb.com/tests/scylla-cluster-tests/d4b37728-fb67-4ec4-8dbb-5c409d679ddb)
[test_manager_backup](https://argus.scylladb.com/tests/scylla-cluster-tests/76a60824-9ce6-41a3-ac04-4c1746b0cbd7)
[10gb-3h-longevity - test failure](https://argus.scylladb.com/tests/scylla-cluster-tests/393d32db-f830-4382-b0e8-ebcafbc6fc27)
==> the longevity can be manually interrupted (since no need for nemesis run of a 3 hours). Just verify the logs are collected and email is send appropriately after test failure.

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -170,8 +170,6 @@ except ImportError as import_exc:
     cluster_cloud = None
     CLUSTER_CLOUD_IMPORT_ERROR = str(import_exc)
 
-configure_logging(exception_handler=handle_exception, variables={'log_dir': TestConfig().logdir()})
-
 try:
     from botocore.vendored.requests.packages.urllib3.contrib.pyopenssl import extract_from_urllib3
 
@@ -3156,8 +3154,12 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):
 
         self.log.info('Test ID: {}'.format(self.test_config.test_id()))
 
+    @pytest.fixture(scope="session", autouse=True)
+    def configure_logging_fixture(self):
+        configure_logging(exception_handler=handle_exception, variables={'log_dir': TestConfig().logdir()})
+
     @pytest.fixture(autouse=True, name='setup_logging')
-    def fixture_setup_logging(self):
+    def fixture_setup_logging(self, configure_logging_fixture):
         self._init_logging()
 
     @pytest.fixture(autouse=True, name='event_system')


conftest.py
```diff
@@ -23,3 +25,8 @@ def pytest_runtest_logreport(report: pytest.TestReport):
     if report.when == "call" and getattr(report, "context", None):
         if report.failed:
             SUBTESTS_FAILURES[report.nodeid].append(report)
+
+
+@pytest.fixture(scope="session", autouse=True)
```
fruch, line 30: I think it would be preferable to move this fixture into tester.py

pehala: Doesnt it make sense to have fixture grouped? Tester used the logging but in the essence there is not much binding it to it and fixtures can be used from conftest everywhere

yarongilor: There could also be impact and dependency to fixtures order in file.

yarongilor: well, since there are already a bunch of fixtures in tester.py it might be better.
especially if there are relatively similar fixtures there, like:
```
    @pytest.fixture(autouse=True, name='setup_logging')
    def fixture_setup_logging(self):
        self._init_logging()
```
i'll move it there as well then.

pehala (dismissed):

yarongilor:
i noticed something weird now - the sct log file is missing in log directory.. have to recheck this.

conftest.py
yarongilor: @fruch , seems like this idea [doesn't work well](https://argus.scylladb.com/tests/scylla-cluster-tests/3295da96-8524-4116-896b-b83f11360563). It might somehow has a different order of setup that ends up with sct log missing.
[rerunning](https://argus.scylladb.com/tests/scylla-cluster-tests/2d0f7603-996c-4e38-a50f-025533b9291f) the same where the fixture moved back to conftest did get the sct log collected.

fruch:
> i noticed something weird now in the recent passed [run](https://argus.scylladb.com/tests/scylla-cluster-tests/3295da96-8524-4116-896b-b83f11360563) - the sct log file is missing in log directory.. have to recheck this.

I think the fixture was suppose to be attached to the Tester class, pytest wouldn't have pick that fixture at all, it doesn't collect from any python file, just from test files themselves or from conftest

yarongilor:
> > i noticed something weird now in the recent passed [run](https://argus.scylladb.com/tests/scylla-cluster-tests/3295da96-8524-4116-896b-b83f11360563) - the sct log file is missing in log directory.. have to recheck this.
> 
> I think the fixture was suppose to be attached to the Tester class, pytest wouldn't have pick that fixture at all, it doesn't collect from any python file, just from test files themselves or from conftest

well, the AI disagreed with that, but i can try it nonetheless..
```
Making it a method (regular or static) of ClusterTester does not help—pytest does not auto-discover fixtures defined as class methods unless the class itself is a test class and the fixture is used within its tests. Also, session-scoped, autouse fixtures are meant to be global, not tied to a test class.

Best practice:
Keep global fixtures like configure_logging_fixture in conftest.py. This ensures pytest always collects and applies them, regardless of which test modules are imported.
```

yarongilor:
> > > i noticed something weird now in the recent passed [run](https://argus.scylladb.com/tests/scylla-cluster-tests/3295da96-8524-4116-896b-b83f11360563) - the sct log file is missing in log directory.. have to recheck this.
> > 
> > 
> > I think the fixture was suppose to be attached to the Tester class, pytest wouldn't have pick that fixture at all, it doesn't collect from any python file, just from test files themselves or from conftest
> 
> well, the AI disagreed with that, but i can try it nonetheless..
> 
> ```
> Making it a method (regular or static) of ClusterTester does not help—pytest does not auto-discover fixtures defined as class methods unless the class itself is a test class and the fixture is used within its tests. Also, session-scoped, autouse fixtures are meant to be global, not tied to a test class.
> 
> Best practice:
> Keep global fixtures like configure_logging_fixture in conftest.py. This ensures pytest always collects and applies them, regardless of which test modules are imported.
> ```

@fruch , it was now [retested](https://argus.scylladb.com/tests/scylla-cluster-tests/1027b372-1c20-43ea-a5c3-814153157cf4) and passed ok (although it's not pytest best practice).

fruch:
> 
> @fruch , it was now [retested](https://argus.scylladb.com/tests/scylla-cluster-tests/1027b372-1c20-43ea-a5c3-814153157cf4) and passed ok (although it's not pytest best practice).

good thing I we decide what are the best practices in SCT, and not AI.

pehala (approved):

pehala:
> well, the AI disagreed with that, but i can try it nonetheless..

I would not count on AI for explanation :)

yarongilor:
> > well, the AI disagreed with that, but i can try it nonetheless..
> 
> I would not count on AI for explanation :)

well AI is just a tool, but this is pytest documentation - https://docs.pytest.org/en/stable/how-to/fixtures.html#scope-sharing-across-classes-modules-packages-or-session

yarongilor:
> provision failed on "sdcm.exceptions.PartitionNotFound: Unable to find partitions to delete" which is a know issue with that short test

so i guess this could now be merged then

Issue #11474, opened 2025-07-22
Docker pull was run unauthenticated but it was logged it


cezarmoise:
https://argus.scylladb.com/tests/scylla-cluster-tests/d3e2ac3c-5342-47cf-8b75-65f24054a621

```
2025-07-16 18:49:16.465: (DisruptionEvent Severity.ERROR) period_type=end event_id=a83843fa-bfec-42b5-a5a0-b519047264a3 duration=4m20s: nemesis_name=ToggleAuditSyslog target_node=Node longevity-twcs-3h-2025-2-db-node-d3e2ac3c-1 [100.24.114.18 | 10.12.2.140] errors=Encountered a bad command exit code!
Command: 'sudo docker pull scylladb/cassandra-stress:3.17.3'
Exit code: 1
Stdout:
Stderr:
Error response from daemon: toomanyrequests: You have reached your unauthenticated pull rate limit. https://www.docker.com/increase-rate-limit
Traceback (most recent call last):
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5591, in wrapper
result = method(*args, **kwargs)
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5160, in disrupt_toggle_audit_syslog
self._disrupt_toggle_audit(store="syslog")
File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5204, in _disrupt_toggle_audit
write_thread = self.tester.run_stress_thread(
File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 2026, in run_stress_thread
return self.run_stress_cassandra_thread(**params)
File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 2071, in run_stress_cassandra_thread
cs_thread = CassandraStressThread(loader_set=self.loaders,
File "/home/ubuntu/scylla-cluster-tests/sdcm/stress_thread.py", line 78, in __init__
super().__init__(loader_set=loader_set, stress_cmd=stress_cmd, timeout=timeout,
File "/home/ubuntu/scylla-cluster-tests/sdcm/stress/base.py", line 60, in __init__
RemoteDocker.pull_image(loader, self.docker_image_name)
File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/docker_remote.py", line 161, in pull_image
remote_cmd(f"docker pull {image}", verbose=True, retry=3)
[...]

PR #11638, opened 2025-08-05
fix(loader): align usage of sudo when pulling images

fruch:
for all bunch of reasons the logic between login into dockerhub and the logic when to use sudo for pulling the docker image wasn't exactly the same, and could lead to situation where we login as root but pull as user, or vice verse.
which all could lead to hiting limits of unauthenticated pulls

this change is aligning to use the same logic for both operations

Fixes: #11474

### Testing

- [ ] provision tests

### Reminders


dimakr:
We can probably skip backporting to 2024.1 as docker_hub_login functionality was not brought there

Issue #11555, opened 2025-07-30
During test startup, no provisioned instances were identified


juliayakovlev:
https://argus.scylladb.com/tests/scylla-cluster-tests/9bd99c0d-be30-49b8-afef-0033354b32bd

3 DB nodes were provisioned.
```
21:34:33  Created instances: [ec2.Instance(id='i-03653ff6e849dfb94'), ec2.Instance(id='i-0cbe5f0a7aa87452b'), ec2.Instance(id='i-024cf13ee3d80d171')].
```

But Scylla configuration on the nodes 2 and 3 were not performed and Scylla was not started on the nodes. 
It happened because the provisioned instances were not found when test was started.
```
< t:2025-07-28 18:35:17,073 f:cluster_aws.py  l:384  c:sdcm.cluster         p:INFO  > Cluster perf-latency-nemesis-ubuntu-db-cluster-9bd99c0d (AMI: ['ami-0f27af70c0500c9f4'] Type: i3en.2xlarge): 
Found provisioned instances = [ec2.Instance(id='i-024cf13ee3d80d171')]
```

This issue happened when the tests were run in `eu-west2` and `eu-west-3`.

Other runs were the issues received:
https://argus.scylladb.com/tests/scylla-cluster-tests/1a7d88b6-b2b4-4e94-bae5-81c9033b11d1
https://argus.scylladb.com/tests/scylla-cluster-tests/17813a1c-5a5d-498d-8618-e3f1feff4e34

juliayakovlev:
We need to validate that the number of provisioned instances matches the test configuration. If the required instances are not found, the test should not run.

PR #11640, opened 2025-08-06
fix(provision): wait for pending instances

soyacz:
Sometimes when test starts to fast, some instances might not be in running state before sct lists them. This causes that only part of instances are detected and test fails at later stages.

fix by waiting for pending instances to become running and returning them from `list_instances_aws` function.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11555

### Testing

- [x] - [issue reproduced with fix in action](https://jenkins.scylladb.com/job/scylla-staging/job/lukasz/job/2025/job/scylla-enterprise-perf-regression-latency-650gb-with-nemesis/16/) (search logs for `Waiting for 2 pending instances in eu-west-3 to become running`)

### Reminders


juliayakovlev (approved):

soyacz:
> LGTM
> 
> shouldn't that happen at the provision stage ? as in Azure that we wait for cloud_init (or I'm mistaken )?

I was thinking about it, but decided this way should be faster for testing (we already start SCT with all the machinery while instances are spinning up)

PR #11656, opened 2025-08-06
feature(vectore-store): add support of VS for docker backend

dimakr:
This change introduces support of vector-store service deployment on docker backend.

A few notes to take into account:
- the feature is not yet merged into scylladb master, so the custom dev build of scylla with support of VS was used for implementation (and its testing).
Once VS is merged we would need to revise this implementation.
- vectore-store service image is not yet published, so the image is built by SCT when VectorStoreSetDocker is initialized.
This is a temporary solution. Once the the image is published we will switch to consuming the released image.

Closes: https://github.com/scylladb/scylla-cluster-tests/issues/11483

### Testing

- [x] integration tests
- [x] [pr-provision-test-docker with enabled VS](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/pr-provision-test/154/)

### Reminders

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -2058,6 +2058,24 @@ local docker network to use, if there's need to have db cluster connect to other
 **type:** str (appendable)
 
 
+## **vs_docker_image** / SCT_VS_DOCKER_IMAGE
+
+Vector Store docker image repo
+
+**default:** scylladb/vector-store
+
+**type:** str (appendable)
+
+
+## **vs_version** / SCT_VS_VERSION
+
+Vector Store version / docker image tag
+
+**default:** latest
+
+**type:** str (appendable)
+
+
 ## **s3_baremetal_config** / SCT_S3_BAREMETAL_CONFIG
 
 
@@ -3647,5 +3665,41 @@ Cloud provider for Scylla Cloud deployment (aws, gce)
 Replication factor for Scylla Cloud cluster (default: 3)
 
 **default:** N/A
+
+**type:** int
+
+
+## **n_vs_nodes** / SCT_N_VS_NODES
+
+Number of vector store nodes (0 = VS is disabled)
+
+**default:** N/A
+
+**type:** int
+
+
+## **vs_port** / SCT_VS_PORT
+
+Vector Store API port
+
+**default:** 6080
+
+**type:** int
+
+
+## **vs_scylla_port** / SCT_VS_SCYLLA_PORT
+
+ScyllaDB connection port for Vector Store
+
+**default:** 9042
+
+**type:** int
+
+
+## **vs_threads** / SCT_VS_THREADS
+
+Vector indexing threads (default: number of CPU cores)
+
+**default:** 2
 
 **type:** int
--- a/sdcm/cluster_docker.py
+++ b/sdcm/cluster_docker.py
@@ -17,6 +17,7 @@ import logging
 from typing import Optional, Union, Dict
 from functools import cached_property
 
+
 from sdcm import cluster
 from sdcm.remote import LOCALRUNNER
 from sdcm.remote.docker_cmd_runner import DockerCmdRunner
@@ -26,6 +27,7 @@ from sdcm.utils.docker_utils import get_docker_bridge_gateway, Container, Contai
 from sdcm.utils.health_checker import check_nodes_status
 from sdcm.utils.nemesis_utils.node_allocator import mark_new_nodes_as_running_nemesis
 from sdcm.utils.net import get_my_public_ip
+from sdcm.utils.vector_store_client import VectorStoreClient
 
 DEFAULT_SCYLLA_DB_IMAGE = "scylladb/scylla-nightly"
 DEFAULT_SCYLLA_DB_IMAGE_TAG = "latest"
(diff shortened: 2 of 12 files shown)


docs/configuration_options.md
```diff
@@ -3649,3 +3649,57 @@ Replication factor for Scylla Cloud cluster (default: 3)
...
+## **vector_store_cpu_limit** / SCT_VECTOR_STORE_CPU_LIMIT
+
+CPU limit for vector store containers (cores)
+
+**default:** 2.0
+
+**type:** float
+
+
+## **vector_store_memory_limit** / SCT_VECTOR_STORE_MEMORY_LIMIT
```
fruch, line 3699: sound to be like docker backend specific, and should be marked as such (even in it's name)

unit_tests/conftest.py
```diff
@@ -124,7 +169,67 @@ def db_alternator_up():
     wait.wait_for(func=db_up, step=1, text='Waiting for DB services to be up', timeout=120, throw_exc=True)
     wait.wait_for(func=db_alternator_up, step=1, text='Waiting for DB services to be up alternator)',
                   timeout=120, throw_exc=True)
-    yield scylla
+
+    vector_store_cluster = None
```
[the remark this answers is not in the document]
dimakr: This was implemented due to current limitations - that scylladb does not yet support live update for `vector_store_uri` parameter. For now I made creating vector-store node close to scylladb node, so that  scylla.yaml can be immediately updated once `vector_store_uri` value is available and the scylla node restarted.
But the PR that implements live update of this parameter in scylla is already in progress. I will split this into different fixtures

sdcm/cluster_docker.py
```diff
@@ -324,6 +390,62 @@ def add_nodes(self, count, ec2_user_data="", dc_idx=0, rack=0, enable_auto_boots
         return self._get_nodes() if self.test_config.REUSE_CLUSTER else self._create_nodes(count, enable_auto_bootstrap)
 
 
+class VectorStoreSetDocker(DockerCluster):
+    """Set of Vector Store nodes"""
+
+    def __init__(self, params, scylla_cluster=None, **kwargs):
+        self.vector_store_image_name = build_vector_store_image()
+        self.scylla_cluster = scylla_cluster
```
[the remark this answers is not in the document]
dimakr: tried to decouple these guys - added additional helpers to reconfigure both side when they are ready

sdcm/cluster_docker.py
```diff
@@ -387,6 +454,114 @@ def get_scylla_args(self):
...
+    def _configure_scylla_nodes_with_vs(self):
+        """Configure Scylla nodes with Vector Store URI"""
+        if not (self.scylla_cluster and (vector_uris := self.get_vector_store_uris())):
+            return
+
+        vector_store_uri = vector_uris[0]
+        self.log.debug("Configuring Scylla nodes with vector_store_uri: %s", vector_store_uri)
+        for node in self.scylla_cluster.nodes:
+            with node.remote_scylla_yaml() as scylla_yml:
+                scylla_yml.vector_store_uri = vector_store_uri
```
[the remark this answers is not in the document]
dimakr: didn't know about this method. Added docker specific implementation of it - needed as sudo is likely not installed in images. And at some point scylla docker image will get rid of supervisor anyway.

soyacz:
I tried to run it - worked, but I realized I used released scylla and not custom one with VS support.
@dimakr how to select this custom scylla that works with VS?

dimakr:
> I tried to run it - worked, but I realized I used released scylla and not custom one with VS support. @dimakr how to select this custom scylla that works with VS?

Above I mentioned only integration tests, that image version that supports VS is hardcoded (similarly to how we hardcode a regular image version for other integration tests).
But forgot to mention that for longevity test (I used simple pr-provision-docker test), I set the version via env vars in the job:
```
SCT_DOCKER_IMAGE=scylladb/scylladb-releng
SCT_SCYLLA_VERSION=2025.4.0-dev-0.20250718.d36c9b3119b7-x86_64
SCT_N_VS_NODES=1
```

dimakr:
> @dimakr were you able to run cql queries targeting VS?

`unit_tests.test_vector_store.TestVectorStoreIntegration.test_vector_search` serves like end-to-end test where it starts real instances of docker based scylladb and VS nodes, and then performs a simple scenario of:
- creating a table with vector data, and index
- inserting some vectors into the table
- executing VS API to perform ANN searches

sdcm/utils/vector_store_client.py
```diff
@@ -0,0 +1,90 @@
...
+
+    def get_info(self) -> dict:
+        """Get Vector Store service information"""
+        return self.request('GET', '/api/v1/info').json()
+
+    def get_indexes(self) -> list[dict]:
+        """Get list of all vector indexes"""
+        return self.request('GET', '/api/v1/indexes').json()
+
+    def ann_search(self, keyspace: str, index: str, embedding: list[float], limit: int = 10) -> dict:
```
fruch, line 53: is that how user would consume VS ? not via CQL ?

docs/configuration_options.md
```diff
@@ -2058,6 +2058,24 @@ local docker network to use, if there's need to have db cluster connect to other
...
+## **vs_docker_image** / SCT_VS_DOCKER_IMAGE
+
+Vector Store docker image repo
+
+**default:** scylladb/vector-store
+
+**type:** str (appendable)
+
+
+## **vs_version** / SCT_VS_VERSION
```
pehala, line 2070: I would unify these into the entire container string, i.e. `scylladb/vector-store:latest`. It is one less configuration option and is consistent with how are containers usually specified

sdcm/utils/vector_store_client.py
fruch: I think doing CQL is the only way when know both scylla and Vector Search are configured correctly, isn't it ?

unit_tests/conftest.py
```diff
@@ -71,8 +107,13 @@ def fixture_docker_scylla(request: pytest.FixtureRequest, params):
     entryfile_path = entryfile_path / 'docker' / 'scylla-sct' / ('entry_ssl.sh' if ssl else 'entry.sh')
 
     alternator_flags = f"--alternator-port {ALTERNATOR_PORT} --alternator-write-isolation=always"
-    docker_version = docker_scylla_args.get(
-        'image', "docker.io/scylladb/scylla-nightly:2025.2.0-dev-0.20250302.0343235aa269")
+
+    # TODO: use vector search capable dev image for vector store tests, until support of vectore store is added to scylla
+    if docker_scylla_args.get('vector_store'):
```
fruch, line 112: I think we can set the image from the test decorator, it would be better then replacing it for of the tests

unit_tests/test_vector_store.py
```diff
@@ -0,0 +1,161 @@
...
+                vector_id = uuid.uuid4()
+                metadata = f"test_item_{i}"
+
+                session.execute(insert_stmt, (vector_id, vector, metadata))
+                test_vectors.append({'id': vector_id, 'vector': vector, 'metadata': metadata})
+
+        return test_vectors
+
+    @staticmethod
+    def _wait_for_vector_indexing(vector_client: 'VectorStoreClient', timeout: int = 300) -> None:  # noqa: F821
```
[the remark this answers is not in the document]
fruch: I think it was meant out of class

And yes we should strive to not use class for tests, it's complicated thinks a bit.

Utility functions, can be just function, saving any state on the test class is a recipe for trouble.

And you do want to cluster a few untill in a class, and have it available via fixture with the right scope.

dimakr:
> I'm trying to use CQL in scylla to use vector store. I operate on running cluster left after SCT test. My steps:
> 1. Restart scylla container (otherwise `InvalidRequest: Error from server: code=2200 [Invalid query] message="Vector Store is disabled"`) - @dimakr please make restarting container after setting scylla yaml with vector store

Added restart of container. Config reload is to be used once live update is implemented in scylla.yaml for vector_store_uri, in https://github.com/scylladb/scylladb/pull/25208

> 2. follow https://cassandra.apache.org/doc/latest/cassandra/getting-started/vector-search-quickstart.html but create table `WITH cdc = {'enabled':true};` (temporary thing to set - in future CDC will be enabled automatically when vector index is created). Also index must be created like this:
> 
> ```
> CREATE CUSTOM INDEX IF NOT EXISTS ann_index  ON cycling.comments_vs(comment_vector) USING 'vector_index'  WITH OPTIONS = { 'similarity_function': 'DOT_PRODUCT' };
> ```
Cassandra examples seems like not completely appropriate for our current implementation, e.g. reqs docs https://scylladb.atlassian.net/wiki/spaces/RND/pages/87261291/Requirements+-+Cloud+Vector+Search say that support of similarity functions in SELECT statement is in the backlog for Milestone3.

fruch:
> Integration tests for vector-store pass:
> ```
> 20:12:26  unit_tests/test_vector_store.py::test_vector_store_deployment PASSED     [ 87%]
> 20:13:01  unit_tests/test_vector_store.py::test_vector_search PASSED               [ 87%]
> 20:13:40  unit_tests/test_vector_store.py::test_vector_search_error_handling PASSED [ 88%]
> ``` 
> But the whole set of integration tests is failing towards the end with:
> ```
> ...
> 20:14:26  unit_tests/test_ycsb_thread.py::test_01_dynamodb_api PASSED              [ 98%]
> 20:14:53  unit_tests/test_ycsb_thread.py::test_02_dynamodb_api_dataintegrity Sending interrupt signal to process
> 20:14:56  time="2025-08-12T18:14:53Z" level=error msg="error waiting for container: command [ssh -l ubuntu -o ConnectTimeout=30 -T -- 34.242.122.227 docker system dial-stdio] has exited with exit status 255, make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr="
> 20:14:56  Cleaning SSH agent
> 20:14:56  Agent pid 14205 killed
> 20:14:56  script returned exit code 125
> ```


I think you just pushed the integration tests over the edge, time wise.

We can temporary extend the timeout for them, but later we'll need to work on their effectiveness (and maybe disable a bunch of them)

sdcm/utils/vector_store_client.py
soyacz: awesome!

dimakr:
@soyacz @fruch can you have another look on this change before your break

soyacz:
@dimakr please add few minutes to integration tests timeout or disable the longest ones for now.
Otherwise LGTM

PR #11663, opened 2025-08-07
fix(doctor): define default timeouts in cqlshrc

fruch:
when we use cqlsh from SCT we use specific command line timeout options, when scylla-doctor is using cqlsh it wasn't and getting into some cases that it times out on connection to the cluster.

this change is adding the same default we have within SCT, into cqlshrc file, so also the scylla-doctor would be using the same values

### Testing

- [ ]

### Reminders


dimakr (approved):
LGTM

PR #11681, opened 2025-08-07
improvement(test_tester): Rewrite the test to use pytester

pehala:
Current test_tester is testing results of test function, which results in:
* `Test` method showing in the output of the unit tests even though they are the ones being tested
* When something fails, it will manifest as a error in teardown
* We need to have `override_pass` mark/hack to make sure the actual method always pass

My solution to this problem is to use pytester which will run the test function in separate test process, from which we can then extract the data and compare with the expected outcome.

#### Pros
* Separates what is being tested from the test
* Clear failure logs
* No more need for `override_pass`
* More compact
   * Less bloat in the test classes
   * Less line overall  

#### Cons
* Longer running time
   * Each of the cases will setup pytest again

### Testing

- [x] Locally

### Reminders

### DIFF
--- a/unit_tests/test_tester.py
+++ b/unit_tests/test_tester.py
@@ -10,14 +10,14 @@
 # See LICENSE for more details.
 #
 # Copyright (c) 2020 ScyllaDB
-
+import re
 import time
 import threading
+import unittest
 from unittest.mock import MagicMock
 import logging
 import pytest
 
-from sdcm.utils.decorators import retrying
 from sdcm.sct_events import Severity
 from sdcm.sct_events.health import ClusterHealthValidatorEvent
 from sdcm.tester import ClusterTester, silence, TestResultEvent
@@ -41,8 +41,6 @@ ClusterTester.__test__ = False
 
 
 class ClusterTesterForTests(ClusterTester, EventsUtilsMixin):
-    __test__ = True
-
     k8s_clusters = None
     argus_heartbeat_stop_signal = threading.Event()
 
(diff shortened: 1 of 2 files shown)


pehala:
> I was like, did you just convert the test out of using pytester, and now you changed your mind ? :)

I converted a different test, which did not need pytester in my opinion and then changed this test, which is the exact usecase for pytester :D

unit_tests/test_tester.py
```diff
@@ -55,6 +53,17 @@ def _init_params(self):
     @pytest.fixture(autouse=True, name='setup_logging')
     def fixture_setup_logging(self, tmp_path):
         self._init_logging(tmp_path / self.__class__.__name__)
+        self.kafka_cluster = None
+
+    @pytest.fixture(scope="function", autouse=True)
+    def fixture_mock_issues(self):
+        """
+        In Pytester the boto credentials are not set, so we mock the issue details, so we dont need them.
```
fruch, line 61: if need we can pass in the credentials (i.e. in CI it would pass it in)

unit_tests/test_tester.py
```diff
@@ -298,9 +237,58 @@ def test(self):
...
+        "EVENT_SUMMARY: {'NORMAL': 2, 'ERROR': 1}",
+        "TEST_STATUS: FAILED",
+        "ERROR 0: save_email_data (silenced)",
+    ], id="SubtestAssertAndTeardownFailsTest"),
+])
+def test_tester_subclass(pytester, test_class, results, outcomes):
+    pytester.makepyfile(f"""
+        from unit_tests.conftest import *
+        from unit_tests.test_tester import {test_class.__name__}
+        {test_class.__name__}.__test__ = True
```
fruch, line 285: maybe it worth adding a comment explaining why doing this, and not calling pytest commnadline with the class

unit_tests/test_tester.py
pehala: I do not think it is needed. I did not delve deeper why the credentiasl are not passed, but given what they are for, I think skiping more calls to AWS is better for a unit test

unit_tests/test_tester.py
```diff
@@ -169,26 +177,14 @@ def test(self):
...
-
 
-@pytest.mark.xfail(reason="this test fails if run after `test_test_user_batch_custom_time`", strict=False)
 class SubtestAssertAndTeardownFailsTest(ClusterTesterForTests):
 
     def test(self):
         with self.subTest('SUBTEST1'):
+            # This wont create error in events
             assert False, 'Subtest1 failed'
+        # this wont get executed as the first one failed
```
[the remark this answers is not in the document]
pehala: You are right, the rest is executed as well. Updated the tests to verify that all errors were thrown

pehala:
v2:
   * Add additional checks for all errors thrown
   * Add explanation for defining special testfile for pytester

fruch (approved):
LGTM

PR #11683, opened 2025-08-08
improvement(nemesis): Run also cluster repair

pehala:
Simplistic improvement to run both types of repairs when we have tablets. It is not perfect, it does not remove normal repair, just runs `cluster repair` with tablets as well.

During investigation a few questions poped up, which I would like to answer before I fully continue and finish this PR.

1. We have two repair functions `run_repair_on_nodes` and `repair_nodetool_repair`, which works almost identically, but `run_repair_on_nodes` can run with manager as well.
   * I consider `run_repair_on_nodes` to be superior as it used manager, if it can, but it is used in fewer nemesis. Can we use `run_repair_on_nodes` by default and only select one or two nemesis which will run it always via nodetool?
2. `repair_nodetool_repair` has a huge 48H timeout, which seems extremely high considering what is happening, while `run_repair_on_nodes` uses only 2 for the same purpose
   * Do we really need 48h timeout? When does it happen? 
3. Theoretically, in pure tablet scenario we do not need old repair, is it worthwhile to add logic for this case? I.e checking how many vnode/tablet keyspaces we have before  

### Testing

- [ ] Locally
- [ ] CI
- [ ] Longevity

   - Will be done after deciding on the actual scope of the PR 

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -1851,6 +1851,10 @@ class Nemesis(NemesisFlags):
         with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48), \
                 self.action_log_scope("Start nodetool repair", target=node.name):
             node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+        if is_tablets_feature_enabled(self.target_node):
+            with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 3), \
+                    self.action_log_scope("Start nodetool cluster repair", target=node.name):
+                node.run_nodetool(sub_cmd="cluster repair", publish_event=publish_event)
 
     def run_repair_on_nodes(self, nodes: list,  ignore_down_hosts=False, publish_event=True):
         """
@@ -1858,6 +1862,15 @@ class Nemesis(NemesisFlags):
         arise from failed or unavailable nodes during the process.
         """
         if not self.cluster.params.get('use_mgmt') and not self.cluster.params.get('use_cloud_manager'):
+            # For cluster repair, we need any node
+            first_node = nodes[0]
+            if is_tablets_feature_enabled(first_node):
+                try:
+                    with adaptive_timeout(Operations.REPAIR, first_node, timeout=HOUR_IN_SEC * 3), \
+                            self.action_log_scope("Start nodetool cluster repair", target=first_node.name):
+                        first_node.run_nodetool(sub_cmd="cluster repair", publish_event=publish_event)
+                except Exception as err:  # pylint: disable=broad-except  # noqa: BLE001
+                    self.log.warning(f"Cluster repair failed to complete on node: {first_node}, with error: {str(err)}")
             for node in nodes:
                 try:
                     with adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 3):


pehala:
> * We have two repair functions `run_repair_on_nodes` and `repair_nodetool_repair`, which works almost identically, but `run_repair_on_nodes` can run with manager as well.
>   
>   * I consider `run_repair_on_nodes` to be superior as it used manager, if it can, but it is used in fewer nemesis. Can we use `run_repair_on_nodes` by default and only select one or two nemesis which will run it always via nodetool?
> 
> * `repair_nodetool_repair` has a huge 48H timeout, which seems extremely high considering what is happening, while `run_repair_on_nodes` uses only 2 for the same purpose
>   
>   * Do we really need 48h timeout? When does it happen?

@scylladb/qa-maintainers Could you please help with answering these questions?

fruch:
> > * We have two repair functions `run_repair_on_nodes` and `repair_nodetool_repair`, which works almost identically, but `run_repair_on_nodes` can run with manager as well.
> >   
> >   * I consider `run_repair_on_nodes` to be superior as it used manager, if it can, but it is used in fewer nemesis. Can we use `run_repair_on_nodes` by default and only select one or two nemesis which will run it always via nodetool?
> > * `repair_nodetool_repair` has a huge 48H timeout, which seems extremely high considering what is happening, while `run_repair_on_nodes` uses only 2 for the same purpose
> >   
> >   * Do we really need 48h timeout? When does it happen?
> 
> @scylladb/qa-maintainers Could you please help with answering these questions?
> 
> Edit: Due to no response, I answered the questions myself and went ahead with full implementation

the general answer about timeouts, is that since we didn't have too much of a machinery around it, people just enlarge things as a "fix" again and again until we reached such numbers.

you might see those things get to high number on some bigger data use cases or on longer longevity (i.e. 7 days) that we don't run anymore that often.

recently there was a request to limit all of the topology related timeouts to max 1h, when tablets is enabled.
no one asked the same for actions which aren't topology, but once can treat them the same

one more thing, if it's under adaptive timeout, we collect the information in elasticsearch (and now in Argus), so one can extract some information to see if we were getting to those high timeouts

sdcm/nemesis.py
```diff
@@ -3217,32 +3231,14 @@ def _mgmt_backup(self, backup_specific_tables):
...
+        self.run_repair_manager()
 
     @target_data_nodes
     def disrupt_mgmt_corrupt_then_repair(self):
         if not self.cluster.params.get('use_mgmt') and not self.cluster.params.get('use_cloud_manager'):
             raise UnsupportedNemesis('Scylla-manager configuration is not defined!')
         self._destroy_data_and_restart_scylla()
-        self._mgmt_repair_cli()
-
-    @latency_calculator_decorator(legend="Scylla-Manger repair")
```
fruch: Remove this has a side effect of changing the name for performance tests, and might interfere with tracking this action

We should preserve the name, in the new function, and pass it via parameter in the latency decorator 

@juliayakovlev to confirm which name used in the performance test for tracking

fruch: changing the name has a risk of slipping into performance v17, and we don't want it change on the official runs, it would make our life harder to look at the history of this operation

you can change the function name, and pass it `latency_calculator_decorator(cycle_name="_mgmt_repair_cli")`

maybe @soyacz can come up with a way argus can fetch both names, but for now those name should be changing

fruch: I fine with metric name won't be changed, until there a clear path of doing those thing across the board in a non breaking way.

sdcm/nemesis.py
```diff
@@ -1688,7 +1688,7 @@ def get_node_state(node_ip: str) -> List["str"] | None:
         try:
             if new_node.get_scylla_config_param("enable_repair_based_node_ops") == "false":
                 InfoEvent(message="StartEvent - Run repair on new node").publish()
-                self.repair_nodetool_repair(new_node)
+                self.run_repair()
```
jsmolar: You are running a repair on all nodes, instead of the original implementation of one node `new_node`.
Is it intended? If not, consider adding an option to specify a list of nodes.

sdcm/nemesis.py
```diff
@@ -2064,31 +2064,64 @@ def call_next_nemesis(self):
...
+            progress_full_string = mgr_task.progress_string(
+                parse_table_res=False, is_verify_errorless_result=True
+            ).stdout
+            if task_final_status != TaskStatus.ERROR_FINAL:
+                mgr_task.stop()
+            raise ScyllaManagerError(
+                f"Task: {mgr_task.id} final status is: {str(task_final_status)}.\nTask progress string: "
+                f"{progress_full_string}"
+            )
+        self.log.info("Task: {} is done.".format(mgr_task.id))
```
jsmolar: Again here, finish this action with one log action with task id and status

pehala: You mean unify the actions_log before the if with this one?

yarongilor:
> We have been running without cluster repair for a quite a while (at least half a year) with tablet enabled, I am not aware of any significant failures, does this mean we might be overdoing repairs or scylla does not need as much repairs for tablets?
Answer: Manager did cluster repair

IMO, from the product POV, this is the most critical fix here. It should be mentioned in title/commit as well i guess.
Not using a cluster repair with tablets caused a data consistency issue in https://github.com/scylladb/scylladb/issues/28024

sdcm/nemesis.py
```diff
@@ -2063,32 +2063,65 @@ def call_next_nemesis(self):
...
-            adaptive_timeout(Operations.REPAIR, node, timeout=HOUR_IN_SEC * 48),
-            self.action_log_scope(f"Start nodetool repair on {node.name} node"),
-        ):
-            node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+    @latency_calculator_decorator(legend="Run repair process with nodetool repair", cycle_name="repair_nodetool_repair")
+    def run_repair_nodetool(self, nodes: list, publish_event=True, timeout=HOUR_IN_SEC * 3):
+        """
+        Execute a repair using Nodetool, which runs both vnode repair (repair) and tablet repairs (cluster repair)
+        """
+        for node in nodes:
```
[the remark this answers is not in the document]
pehala: If there is nothing to repair, it will be quick so no big deal. Also system tables are still Vnodes so we have to do this regardless of user keyspace being vnode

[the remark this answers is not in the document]
yarongilor: i think it's not the instructions in documentation but it may be a documentation issue or a design issue. If there's always a need for a nodetool repair, then, for example,  it could be executed internally for tablets.

yarongilor (approved):
LGTM

yarongilor:
@pehala , shouldn't this be backported? to 2025.4?

pehala:
> @pehala if you intended to have something in Argus results tab, then it didn't pan out - unfortunatly there's no logs to verify why.
> Latency decorator works only for specific test names IIRC (doesn't work with normal 'test_custom_time'.
> Also I didn't see tables for adaptive timeout around this new repair.

The only thing I wanted was backwards compatibility, so nothing new should appear, but old names should be kept.

Issue #11589, opened 2025-08-04
Argus Heartbeat is not stopped properly


soyacz:
Recent change https://github.com/scylladb/scylla-cluster-tests/pull/11557 (and introduced issue that was workarounded with https://github.com/scylladb/scylla-cluster-tests/pull/11588)
caused Argus heartbeat thread not being stopped. In sct.log we can see at the end:
```
< t:2025-08-02 22:28:57,457 f:tester.py       l:3178 c:root                 p:DEBUG > Threads and processes at the end of the test:
[...]
< t:2025-08-02 22:28:57,457 f:tester.py       l:3180 c:root                 p:DEBUG > Active thread: argus-heartbeat (id=135489860376320, daemon=True, repr=<Thread(argus-heartbeat, started daemon 135489860376320)>)
[...]
```

PR #11727, opened 2025-08-12
fix(argus): remove session scope from finalize argus

soyacz:
Because argus_finalize fixture requires tester object to be instantiated, removed 'session' scope from it. Otherwise it causes problems with clearing out argus heartbeat thread (and caused artifact tests issues in the past).

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11589

### Testing

- [x] - [failure example](https://argus.scylladb.com/tests/scylla-cluster-tests/abc18165-bc9d-42ec-9b8c-bff6b3445f27/events)
- [x] - [passing example](https://argus.scylladb.com/tests/scylla-cluster-tests/e1ff7422-4179-42bd-9585-51b3affe1628)

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -338,7 +338,6 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):
     monitors: BaseMonitorSet = None
     loaders: Union[BaseLoaderSet, LoaderSetAWS, LoaderSetGCE] = None
     db_cluster: Union[BaseCluster, BaseScyllaCluster] = None
-    argus_heartbeat_stop_signal = threading.Event()
 
     @property
     def k8s_cluster(self):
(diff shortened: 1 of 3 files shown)


pehala (dismissed):

sdcm/tester.py
```diff
@@ -338,7 +338,6 @@ class ClusterTester(db_stats.TestStatsMixin, unittest.TestCase):
     monitors: BaseMonitorSet = None
     loaders: Union[BaseLoaderSet, LoaderSetAWS, LoaderSetGCE] = None
     db_cluster: Union[BaseCluster, BaseScyllaCluster] = None
-    argus_heartbeat_stop_signal = threading.Event()
```
fruch, line 341: I think this need to be clear from some unittests as well..

fruch (dismissed):
LGTM

sdcm/tester.py
soyacz: removed

PR #11775, opened 2025-08-17
Fix skipping part of remote log output

radoslawcybulski:
SSHLooger instances use `tail -f` to fetch content of the file over ssh. This might occasionally skip part of the log - if the remote command producing the log starts quickly and outputs a lot of log instanstly running `tail -f` might skip beggining of the log file (as `tail -f` will emit only N last lines and then follow file changes). We add `-c +0` parameter, which tells `tail -f` to start following file from byte 0 counting from beginning.

### Testing
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/evgeniy/job/perf-regression-alternator/job/scylla-master-perf-regression-alternator-latency-test/336/ (and plenty before) runs with this patch.

### DIFF
--- a/sdcm/utils/remote_logger.py
+++ b/sdcm/utils/remote_logger.py
@@ -177,7 +177,7 @@ class HDRHistogramFileLogger(SSHLoggerBase):
 
     @cached_property
     def _logger_cmd_template(self) -> str:
-        return f"tail -f {self._remote_log_file}"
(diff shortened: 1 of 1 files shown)


soyacz:
won't it bite us when there's interruption and whole log will be read out again?
We used to prevent this by opening logger before any writes to it (e.g. precreating hdrh files and establishing `tail -f` before populating them)

Issue #11724, opened 2025-08-12
disrupt_load_and_stream cant get sstable file for hashing with azure backend


timtimb0t:
[Argus](https://argus.scylladb.com/test/7c36a727-b3fc-4bc4-aa5a-0de4d5afd9c7/runs?additionalRuns[]=a6a41c83-2d59-49b2-9e81-0882fbdf7121)

Scylla version: `2025.2.2-20250808.d845de84aa84` with build-id `4ca292ad5eb2c2e0abf007fba73b77245b700edd`

Kernel Version: `6.11.0-1018-azure`

Failure:
```
< t:2025-08-10 21:42:00,840 f:base.py         l:229  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.0.0.10>: md5sum: /tmp/node2.part2.keyspace1.standard1.tar.gz: No such file or directory
< t:2025-08-10 21:42:00,841 f:base.py         l:145  c:RemoteLibSSH2CmdRunner p:ERROR > <10.0.0.10>: Error executing command: "md5sum /tmp/node2.part2.keyspace1.standard1.tar.gz"; Exit status: 1
< t:2025-08-10 21:42:00,841 f:base.py         l:150  c:RemoteLibSSH2CmdRunner p:DEBUG > <10.0.0.10>: STDERR: md5sum: /tmp/node2.part2.keyspace1.standard1.tar.gz: No such file or directory

< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > Encountered a bad command exit code!
< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > 
< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > Command: 'md5sum /tmp/node2.part2.keyspace1.standard1.tar.gz'
< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > 
< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > Exit code: 1
< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > 
< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > Stdout:
< t:2025-08-10 21:42:00,841 f:common.py       l:130  c:utils                p:ERROR > 
[...]

PR #11795, opened 2025-08-20
fix(azure): fix getting private DNS name for Azure VMs

vponomaryov:
After merge of the SCT#11735 PR we started getting following kind of errors:
```
    File "<path-to>/sdcm/utils/net.py", line 47, in resolve_ip_to_dns
      return socket.gethostbyaddr(ip_address)[0]
  socket.herror: [Errno 1] Unknown host
  During handling of the above exception, another exception occurred:
    File "<path-to>/sdcm/cluster.py", line 4030, in node_setup
      cl_inst.node_setup(_node, **setup_kwargs)
    File "<path-to>/sdcm/cluster.py", line 4871, in node_setup
      node.create_node_certificate(cert_file=node.ssl_conf_dir / TLSAssets.DB_CERT,
    File "<path-to>/sdcm/cluster.py", line 1844, in create_node_certificate
      dns_names=[self.public_dns_name, self.private_dns_name])
    File "/usr/local/lib/python3.10/functools.py", line 981, in __get__
      val = self.func(instance)
    File "<path-to>/sdcm/cluster_azure.py", line 189, in private_dns_name
      return resolve_ip_to_dns(self.private_ip_address)
    File "<path-to>/sdcm/utils/net.py", line 49, in resolve_ip_to_dns
      raise ValueError(f"Unable to resolve IP: {e}")
  ValueError: Unable to resolve IP: [Errno 1] Unknown host
```

So, do following changes to mitigate it:
- Add Azure-based DNS server to the list of DNS servers.
- Try to generate the private DNS name in case private IP address failed to be resolved into a private DNS name.
- Print IP address that failed to be resolved for case it is empty value or some other value error.

### Testing

- [ ]

### Reminders


sdcm/cluster_azure.py
```diff
@@ -187,7 +187,14 @@ def query_azure_metadata(self, path: str, api_version: str = "2024-07-17") -> st
 
     @cached_property
     def private_dns_name(self) -> str:
-        return resolve_ip_to_dns(self.private_ip_address)
+        try:
+            return resolve_ip_to_dns(self.private_ip_address)
+        except Exception as exc:  # noqa: BLE001
```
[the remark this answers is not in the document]
vponomaryov: It may be wider set of exception than the just 2 proposed.

sdcm/provision/azure/virtual_network_provider.py
```diff
@@ -56,7 +56,11 @@ def get_or_create(self, name: str = "default") -> VirtualNetwork:
                     "address_prefixes": ["10.0.0.0/16"],
                 },
                 "dhcp_options": {
-                    "dns_servers": ["8.8.8.8", "1.1.1.1"]
+                    "dns_servers": [
+                        "8.8.8.8",  # Google
+                        "1.1.1.1",  # Cloudflare
+                        "168.63.129.16",  # Azure
```
dimakr, line 62: If I understood correctly, adding this special endpoint won't help us to have both internal + external DNS resolution at the same time. I found this summary of setting/not-setting some DNS servers in `dhcp_options:dns_servers`:
```
The Azure-provided DNS server is always available at the first IP of a subnet + 1 (usually .1).
Purpose: VM-to-VM name resolution inside the VNet + forwarding to external DNS.
Example: In 10.0.0.0/24, DNS = 10.0.0.1.

By default: VM > 10.0.0.1 > forwards to 168.63.129.16. (best for internal + external DNS)
If dns_servers is set the internal DNS is ignored: VM bypasses 10.0.0.1.
```

dimakr: So there are probably 2 options - remove this 168.63.129.16 endpoint and keep only external DNS servers. And for internal name use the workaround of building name ourselves.
Or rollback the change where public DNS servers were added and add some retries instead, to be able to resolve the problem that the change was addressing (the https://github.com/scylladb/scylla-cluster-tests/issues/11724 one)

fruch: @dimakr reverting it in https://github.com/scylladb/scylla-cluster-tests/pull/11843

once someone can figure it out, we can reapply

PR #11801, opened 2025-08-20
refactor(parallel object): move to it's own new file

yarongilor:
The classes of ParallelObject, ParallelObjectResult and ParallelObjectException, are moved to it's own new file of sdcm/utils/parallel_object.py. 
This will prevent some circular import issues.
Refs: https://github.com/scylladb/scylla-cluster-tests/pull/11353

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/cluster_tools.py
+++ b/sdcm/utils/cluster_tools.py
@@ -16,6 +16,7 @@ import logging
 from typing import TYPE_CHECKING
 from collections import defaultdict
 
+
 if TYPE_CHECKING:
     from sdcm.cluster import BaseCluster, BaseNode
 
--- a/functional_tests/scylla_operator/test_functional.py
+++ b/functional_tests/scylla_operator/test_functional.py
@@ -39,7 +39,7 @@ from sdcm.cluster_k8s import (
     SCYLLA_OPERATOR_NAMESPACE
 )
 from sdcm.mgmt import TaskStatus
-from sdcm.utils.common import ParallelObject
+from sdcm.utils.parallel_object import ParallelObject
 from sdcm.utils.k8s import (
     convert_cpu_units_to_k8s_value,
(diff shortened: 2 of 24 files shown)


sdcm/utils/cluster_tools.py
```diff
@@ -15,6 +15,9 @@
 import logging
 from typing import TYPE_CHECKING
 from collections import defaultdict
+from functools import partial
+
+from sdcm.utils.parallel_object import ParallelObject
```
dimakr, line 20: pre-commit failed on this lines, the imported names are not used in the file

pehala: Each PR needs to be independent, if this is from the next PR, you separated them into commits wrong. Please fix this. You will still have to do rebase on the next PR so this will be insignificant change for it

yarongilor: it cannot be independent if it is dependent and solved a problem of the previous PR.
code freeze makes much more sense to me than tedious endless rebase but fixed it anyway.

dimakr (approved):
LGTM

sdcm/utils/cluster_tools.py
dimakr: yesterday it wouldn't be possible to merge anyway, as those maintainers who were available are not having rights to bypass failed checks
And from today seems like only me is available from test infra team, until the end of Aug

pehala: > it cannot be independent if it is dependent and solved a problem of the previous PR.

Extracting one class into another file is not dependent on anything. 

> it cannot be independent if it is dependent and solved a problem of the previous PR. code freeze makes much more sense to me than tedious endless rebase but fixed it anyway.

Code freeze in this scenario is short for: we cannot structure our PRs correctly :)

pehala (approved):

PR #11809, opened 2025-08-21
improvement(treewide): Remove unused tests

pehala:
### Why?
Tests which havent been touched or run in years do not bring any value, as they most likely cannot be run easily, only serve to confuse people and add work when any rework is happening.

### What?
* Removed tests which were not referenced by any Jenkins Pipeline, and as such couldn't be easily run today
   * Remove CMakeLists and queries-limits, which were used for `QueryLimitsTest` 
* Remove tox.ini
   * Tox was removed already

### Testing

- [x] None


pehala:
FYI: There might be more PRs in similar vain, but these files are in my opinion clear candidates for removal

jsmolar (approved):
With this we are removing tests for `admission`, if we don't need to test this in SCT I have no objections. I appreciate all other code deletion.

vponomaryov (approved):
LGTM

PR #11810, opened 2025-08-21
fix(performance): disable simulated racks for vnodes operation tests in branch-v17

juliayakovlev:
In branch `v17`, the `simulated_racks` setting was applied globally across all tests. 
However, the `v15` performance branch lacks many of the recent developments related to multi-rack 
support (e.g., simulated racks), and has historically been running with a single rack 
configuration.

By enabling multiple racks now, we risk invalidating historical results without a clear benefit. 
In Scylla Cloud, multi-rack has always been the default, but for consistency and comparability 
in performance testing, we’ve decided to:

- Keep vnodes performance under operations tests running with a single rack
- Run tablets tests with 3 racks

This change ensures continuity of results while still aligning with current best practices for 
tablets.

See related Scylla issue: https://github.com/scylladb/scylladb/issues/25924#issuecomment-3316012192

### Testing

- [x]  [perf-regression-latency-650gb-with-nemesis](https://argus.scylladb.com/tests/scylla-cluster-tests/dfeb52d4-ee57-423a-a91e-30852ad1a40f) - the one rack cluster is created

### Reminders

### DIFF
--- a/configurations/disable_simulated_racks.yaml
+++ b/configurations/disable_simulated_racks.yaml
@@ -1,0 +1,3 @@
+# Disable simulated racks for the vnodes operations performance tests
+# https://github.com/scylladb/scylladb/issues/25924#issuecomment-3308731799
+simulated_racks: 0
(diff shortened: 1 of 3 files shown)


pehala:
> We should use racks in all tests **except** for performance tests.
> 
> All performance tests are run **without** simulated or physical racks. Your change (`simulated-rack=3` set globally) was **intentionally** not backported to the performance branches.

The requirement from management that I got was all tests, but maybe that is old information or never should have included performance.

However, if every customer should use racks, then I would suspect we want performance numbers with racks as well.

@roydahan Could you confirm that performance indeed do not need racks?

fruch:
The title if the PR is a bit misleading

But the content seems o.k.

juliayakovlev:
> The title if the PR is a bit misleading
> 
> But the content seems o.k.

changed commit and PR subject

PR #11821, opened 2025-08-23
feature(scylla-cloud): VPC peering for Scylla cloud clusters

dimakr:
Introduce ability to create, configure and clean up VPC peering connections between Scylla cloud cluster and SCT infrastructure.

Closes: https://github.com/scylladb/qa-tasks/issues/1941

### Testing

- [x] :green_circle: [cloud cluster on AWS cloud provider with VPC peering](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/173/)
- [x] :green_circle: [cloud cluster on GCE cloud provider with VPC peering](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/174/)

### Reminders

### DIFF
--- a/sdcm/utils/resources_cleanup.py
+++ b/sdcm/utils/resources_cleanup.py
@@ -14,6 +14,7 @@
 from __future__ import absolute_import, annotations
 import logging
 import os
+import ipaddress
 from typing import Optional
 from unittest.mock import MagicMock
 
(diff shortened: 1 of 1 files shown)


dimakr:
> Cannot be tested in fully automated way/in CI, for now - when stress thread collects node IPs for stress command it reads them from remote scylla.yaml (when private connection is used for communication with cluster). And due to ssh access approach for scylla cloud backend is not yet implemented, we cannot get those IPs.
> 
> Tested with manual intervention when running test from local sct runner - paused test on a breakpoint and set the IPs manually.
> 
> Working on approach to get IPs of cluster nodes not from scylla.yaml for cloud backend.

Turned out not a big - scylla cloud APIs has a few methods for getting IPS, so re-implementation of `node.cql_address` attribute for scylla cloud node was straightforward.

defaults/cloud_config.yaml
```diff
@@ -5,3 +5,7 @@ n_monitor_nodes: 0
 use_mgmt: false
 
 use_cloud_manager: true
+
+# VPC peering configuration
+xcloud_vpc_peering:
+  enabled: true
```
fruch, line 11: what are the other configuration options needed here ?, that we need this one as dict ?

sdcm/utils/resources_cleanup.py
```diff
@@ -567,6 +568,59 @@ def clean_placement_groups_aws(tags_dict: dict, regions=None, dry_run=False):
                     raise
 
 
+def cleanup_cluster_vpc_peering(api_client: ScyllaCloudAPIClient, account_id: str, cluster_id: int,
+                                cluster_name: str, config: dict) -> None:
+    try:
+        peers = api_client.get_vpc_peers(account_id=account_id, cluster_id=cluster_id)
+    except Exception as e:  # noqa: BLE001
+        if '041104' not in str(e):  # '041104' is the "The cluster does not have VPC peering enabled" Siren error code
```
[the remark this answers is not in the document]
dimakr: Right, this is for convenience, For some reason scylla cloud API for getting peers list is returning this error code 041104, if there are no peering connections (not just empty list, for example). So here we just do nothing, if there are no peers in a cluster (i.e. normal situation) and continue with the cleanup. And if there are some other problems, we log it as something unexpected, but continue with cleanup.

defaults/cloud_config.yaml
dimakr: This is a placeholder for now. If in the future we, for whatever reason, would need some not common ports open, or maybe some additional CIDRs added to the peer (not just the CIDR of the cluster).

dimakr: Or we can drop having dict here for now (in favor of a simple flag). And add more configuration capabilities later, when we really need them.

fruch: yes, even we can put it under the environment configuration, like the api key and rest, and control it be the configuration file we'll use

dimakr:
Added VPC peering management for GCE cloud provider.

Next step is to implement a mechanism to select/use a 'not yet allocated cidr' for peerings, so we can run parallel tests.

sdcm/cluster_cloud.py
```diff
@@ -502,3 +533,165 @@ def _get_cluster_credentials(self):
...
+
+    def setup_vpc_peering(self, dc_id: int) -> None:
+        """Set up VPC peering connection between SCT and Scylla Cloud VPC"""
+        if not self.vpc_peering_enabled:
+            return
+
+        self.log.info("Setting up VPC peering for cluster %s", self._cluster_id)
+        sct_vpc_info = self._get_sct_vpc_info()
+
+        if ipaddress.ip_network(sct_vpc_info['cidr']).overlaps(ipaddress.ip_network(self.cloud_cidr)):
```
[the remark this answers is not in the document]
fruch: it's more or less we we implemented in the siren-test plugin, the fact we splited the whole range into smaller bits.
one more thing is that we can check the route tables for existing ranges, and use that for a lock, i.e. add to the route table, we can write the ranges into the route table as a lock, and later update them to the peering info. 

also we can use cleanups, and clear blackhole routes (i.e. that the VPC peer is canceled)

I don't remember the exact details of GCE, but we probably should be able to achieve something similar 


so yes we should split the range of addresses, and we should use the routing tables as a lock, I think it makes the best sense, any other share resources for lock/sync (i.e. S3 or other) just gonna complicate the situation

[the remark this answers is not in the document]
dimakr: Added commit with cleanup helpers for `blackhole` routes in AWS and `inactive` peerings in GCP

sdcm/cluster_cloud.py
```diff
@@ -200,6 +207,10 @@ def start_coredump_thread(self):
     def is_cloud() -> bool:
         return True
 
+    @cached_property
+    def cql_address(self):
```
fruch, line 211: This gonna be trickery for testing from a local machine, we should test if we can still run from local machine after those chamges

PR #11829, opened 2025-08-25
fix(trigger): scylla version for nemesis test

juliayakovlev:
It was a bug in the performance trigger.
The nemesis test is intentionally defined twice:
- once for releases (with a single sub-test)
- once for master (with three sub-tests). However, when the same job is defined multiple times, the process broke midway and does not to set the correct image_name. As a result, the job is triggered with 'master:latest' instead of using the image name.

### Testing

- [x] [perf-regression-trigger-test](https://github.com/scylladb/scylla-cluster-tests/pull/11829)

### Reminders


fruch (approved):
LGTM

PR #11834, opened 2025-08-27
fix(cluster): cql_connection for exclusive node

juliayakovlev:
The issue has been addressed in [scylla-cluster-tests#11520](https://github.com/scylladb/scylla-cluster-tests/pull/11520).
The problem was that the node IP was omitted when selecting the load balancing policy in the case of an exclusive connection. 
This led to [scylladb#25575,](https://github.com/scylladb/scylladb/issues/25575) which is actually an SCT-related issue, not a Scylla issue.

For exclusive connections, the load balancing policy must use the IP of a specific node. However, due to the bug, a random node was selected instead. This resulted in incorrect peer information being used.
RackAwareRoundRobinPolicy is not applicable for exclusive node connections,
as it operates based on rack and datacenter, not individual nodes.
So revert the change that was presented in https://github.com/scylladb/scylla-cluster-tests/pull/11520 for `cql_connection_exclusive` method

### Testing

- [x] [longevity-100gb-4h-rackaware](https://argus.scylladb.com/tests/scylla-cluster-tests/3834b7f4-05ab-4c79-8e93-c005c4fa1b17)
- [x] [longevity-twcs-3h-test](https://argus.scylladb.com/tests/scylla-cluster-tests/cc63b09c-b5f7-43ce-9768-cb5f8e4d2998)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3691,10 +3691,10 @@ class BaseCluster:
 
         return ScyllaCQLSession(session, cluster_driver, verbose)
 
-    def get_load_balancing_policy(self, node, whitelist_nodes=None):
+    def get_load_balancing_policy(self, whitelist_nodes=None):
         node_ips = self.get_node_cql_ips(nodes=whitelist_nodes if whitelist_nodes else self.nodes)
         wlrr = WhiteListRoundRobinPolicy(node_ips)
-        if self.params.get("rack_aware_loader") and node.parent_cluster.racks_count > 1:
+        if self.params.get("rack_aware_loader") and self.racks_count > 1:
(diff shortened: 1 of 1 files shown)


pehala:
I think https://github.com/scylladb/scylladb/issues/25575#issuecomment-3228130941 had a good point. This seems like critical part of SCT, so I think we should at least try to create unit or integration test for it

mykaul:
This fix does not have any test associated with it. How do we ensure we do not regress in the future? What tests can we make to make SCT more robust?

pehala:
> This fix does not have any test associated with it. How do we ensure we do not regress in the future? What tests can we make to make SCT more robust?

The test is present in the followup PR: https://github.com/scylladb/scylla-cluster-tests/pull/11866

Issue #11850, opened 2025-08-28
FullScanAggregateEvent Severity.ERROR "Keyspace keyspace1 does not exist"


yarongilor:
## Packages

Scylla version: `2025.4.0~dev-20250826.6dc2c42f8bde` with build-id `f105b9ddb8c3baa36fefba1dc44ad1411ac5ae72`

Kernel Version: `6.11.0-1018-azure`

## Issue description
Start time: 2025-08-26 23:20:01

Before nemesis started, full-scan thread got 30 errors of:
```
2025-08-26 23:58:47.791: (FullScanAggregateEvent Severity.ERROR) period_type=end event_id=8319f694-1d98-4409-99c1-d702d657825a duration=0s node=longevity-10gb-3h-master-db-node-6f4e4f94-eastus-4 select_from=keyspace1.standard1 message=InvalidRequest('Error from server: code=2200 [Invalid query] message="Keyspace keyspace1 does not exist"')
```
- [ ] This issue is a regression.
- [ ] It is unknown if this issue is a regression.

*Describe your issue in detail and steps it took to produce it.*

## Impact

*Describe the impact this issue causes to the user.*

## How frequently does it reproduce?

*Describe the frequency with how this issue can be reproduced.*

## Installation details

Cluster size: 6 nodes (Standard_L8s_v3)

Scylla Nodes used in this run:
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-8 (null | 10.0.0.15) (shards: 7)
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-7 (null | 10.0.0.14) (shards: 7)
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-6 (null | 10.0.0.10) (shards: 7)
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-5 (null | 10.0.0.9) (shards: 7)
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-4 (null | 10.0.0.8) (shards: 7)
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-3 (null | 10.0.0.7) (shards: 7)
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-2 (null | 10.0.0.6) (shards: 7)
- longevity-10gb-3h-master-db-node-6f4e4f94-eastus-1 (null | 10.0.0.5) (shards: 7)

[...]

PR #11852, opened 2025-08-28
fix(full scan): wait more time before start the thread

juliayakovlev:
Due to potential issues with Docker Hub, the full scan operation thread may start before the required schema is created. When this occurs, an error message is generated.

To mitigate this, we currently wait for 20 minutes before starting the thread. To further reduce the likelihood of encountering this error, increase the timeout to 40 minutes.

Additionally, an exception should be thrown if the table is not available after waiting.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11850
### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/operations_thread.py
+++ b/sdcm/utils/operations_thread.py
@@ -172,15 +172,15 @@ class OperationThread:
     def join(self, timeout=None):
         return self._thread.join(timeout)
 
-    def _wait_until_user_table_exists(self, timeout_min: int = 20):
+    def _wait_until_user_table_exists(self, timeout_min: int = 40):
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

mykaul:
This is the wrong fix. The correct fix is to pre-download the container if the test needs it. If it fails, no point in running the test.

fruch:
> This is the wrong fix. The correct fix is to pre-download the container if the test needs it. If it fails, no point in running the test.

docker pull didn't failed, it just took longer the usual, this code assumes table is created, and that what it's waiting for, regardless of any docker issues.

and it's now gonna fail clearly starting the table it's expects isn't created, and not silently as before.

that's good enough fix.

PR #11874, opened 2025-09-01
fix(performance): disable parallel node operations

juliayakovlev:
Disable parallel node operations in performance nemesis tests.

### Testing

- [ ]

### Reminders

### DIFF
--- a/test-cases/performance/perf-regression-latency-650gb-with-nemesis.yaml
+++ b/test-cases/performance/perf-regression-latency-650gb-with-nemesis.yaml
@@ -34,6 +34,7 @@ use_hdrhistogram: true
 use_placement_group: true
 use_capacity_reservation: true
 email_subject_postfix: 'latency during operations'
+parallel_node_operations: false  # doing this to align the measurement
 cluster_health_check: false
 stress_image:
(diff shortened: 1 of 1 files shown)


test-cases/performance/perf-regression-latency-650gb-with-nemesis.yaml
```diff
@@ -34,6 +34,7 @@ use_hdrhistogram: true
 use_placement_group: true
 use_capacity_reservation: true
 email_subject_postfix: 'latency during operations'
+parallel_node_operations: false
```
fruch, line 37: let put some comment here, since we are doing this to align the measurement we were doing on vnodes

we should make sure we don't make it false for tablets

juliayakovlev: @fruch 
Why it should be "true" for tablets test? If it is "false"  in v15 branch (https://argus.scylladb.com/tests/scylla-cluster-tests/30709120-a236-4e4a-b720-65561280bf0a)

fruch: let not change the tablets as well, so it can be comparable.

fruch (approved):
LGTM

Issue #11851, opened 2025-08-28
all upgrades base on gce_image are failing on 2025.3 release


fruch:
## Issue description

- [x] This issue is a regression.
- [] It is unknown if this issue is a regression.

seems like since the refactoring of the upgrade pipelines, we are getting those:
```
ValueError: 'scylla_version' can't used together with 'ami_id_db_scylla', 'gce_image_db' or with 'scylla_repo'
```

seems like we are passing both version and image, and the check is failing

need to see why it's not failing on master, and what's broken on the backport (or it's not the backport)

## Impact

all of those jobs can start a test

## How frequently does it reproduce?

on all runs 

## Installation details

Cluster size: Unknown amount of nodes (Unknown instance type)

Scylla Nodes used in this run:
**No resources left at the end of the run**

OS / Image: `No image` (NO RUNNER: NO RUNNER)

Test: `rolling-upgrade-debian11-test`
Test id: `e6777a70-6fd5-48fb-a9c6-c3fceedaef66`
Test name: `scylla-2025.3/rolling-upgrade/rolling-upgrade-debian11-test`

Test config file(s):

- [None](https://github.com/scylladb/scylla-cluster-tests/blob/049fb6e567637f74c0ce2beef0d8abbd3659b819/undefined)

<details>
<summary>
Logs and commands
</summary>

- Restore Monitor Stack command: `$ hydra investigate show-monitor e6777a70-6fd5-48fb-a9c6-c3fceedaef66`
- Restore monitor on AWS instance using [Jenkins job](https://jenkins.scylladb.com/view/QA/job/QA-tools/job/hydra-show-monitor/parambuild/?test_id=e6777a70-6fd5-48fb-a9c6-c3fceedaef66)
- Show all stored logs command: `$ hydra investigate show-logs e6777a70-6fd5-48fb-a9c6-c3fceedaef66`

## Logs:
*No logs captured during this run.*

[Jenkins job URL](https://jenkins.scylladb.com/job/scylla-2025.3/job/rolling-upgrade/job/rolling-upgrade-debian11-test/7/)
[Argus](https://argus.scylladb.com/test/d5a1fa9f-2ae3-48f7-bf9a-3c9975b4cf10/runs?additionalRuns[]=e6777a70-6fd5-48fb-a9c6-c3fceedaef66)
</details>

vponomaryov:
[...]

PR #11880, opened 2025-09-01
fix(ollingUpgradePipeline): remove `scylla_repo` from params

fruch:
since scylla-pkg is passing along both `scylla_repo` and `new_scylla_repo` we end up in situation we are having a conflict of input parameters

this change is removing the `scylla_repo` from params in the rolling upgrade pipelines

Fixes: #11851

### Testing

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu2404-test/5/ (with scylla_repo)
- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu2404-test/6/ (without scylla_repo

### Reminders

### DIFF
--- a/vars/rollingUpgradePipeline.groovy
+++ b/vars/rollingUpgradePipeline.groovy
@@ -204,6 +204,8 @@ def call(Map pipelineParams) {
                             def base_version = version
                             params_mapping[base_version] = params.collectEntries { param -> [param.key, param.value] }
                             params_mapping[base_version].put('scylla_version', base_version)
+                            // since scylla-pkg might pass this one, we are not supporting it here, as we always starts by version
+                            params_mapping[base_version].remove('scylla_repo')
 
                             // those params are not in the job params, so user can`t change them
                             // but they are coming from the pipelineParams, i.e. hardcoded per use case


roydahan:
But how would one set the scylla_repo to install for the base version?
Say I want to install a private build (not official release) and upgrade it.

fruch:
> But how would one set the scylla_repo to install for the base version?
> Say I want to install a private build (not official release) and upgrade it.

It was never supported in this pipeline, this parameter wasn't exposed

vars/rollingUpgradePipeline.groovy
```diff
@@ -204,6 +204,8 @@ def call(Map pipelineParams) {
                             def base_version = version
                             params_mapping[base_version] = params.collectEntries { param -> [param.key, param.value] }
                             params_mapping[base_version].put('scylla_version', base_version)
+                            // since scylla-pkg might pass this one, we are not supporting it here, as we always starts by version
+                            params_mapping[base_version].remove('scylla_repo')
```
vponomaryov, line 208: Will it work if the `scylla_repo` is absent?

fruch: > To remove an entry from the map, use the remove() method. It returns the value that was previously associated with the key passed in, otherwise it returns null if it did not find the given key value in the map to remove.

https://docs.oracle.com/en/cloud/saas/applications-common/25a/cgsac/working-with-maps.html

vponomaryov (approved):
LGTM

Issue #11629, opened 2025-08-05
code should ignore InvalidAllocationID.NotFound when release


timtimb0t:
[Argus](https://argus.scylladb.com/test/cb38c21b-55d1-481e-a808-97cc2695f007/runs?additionalRuns[]=c7a1b827-d3be-4f69-9924-1631143411fc)

Scylla version: `2025.3.0~rc2-20250730.7164f11b997d` with build-id `826d9577efb79327171bd966f9124387510ea5a8`

Kernel Version: `6.14.0-1010-aws`

disrupt_add_remove_dc nemesis failed with error:
```
raceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5634, in wrapper
    result = method(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4833, in disrupt_add_remove_dc
    with ExitStack() as context_manager:
  File "/usr/local/lib/python3.10/contextlib.py", line 576, in __exit__
    raise exc_details[1]
  File "/usr/local/lib/python3.10/contextlib.py", line 561, in __exit__
    if cb(*exc_details):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 4840, in finalizer
    self.cluster.decommission(new_node)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 5220, in decommission
    if tablets_enabled := is_tablets_feature_enabled(node):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/features.py", line 87, in is_tablets_feature_enabled
    with node.remote_scylla_yaml() as scylla_yaml:
  File "/usr/local/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/cluster.py", line 1717, in remote_scylla_yaml
    with self._remote_yaml(path=self.add_install_prefix(abs_path=SCYLLA_YAML_PATH)) as scylla_yaml:
  File "/usr/local/lib/python3.10/contextlib.py", line 135, in __enter__
    return next(self.gen)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/remote/remote_file.py", line 40, in remote_file
    ownership = remoter.sudo(cmd=f'stat -c "%U:%G" {remote_path}').stdout.strip()
[...]

PR #11883, opened 2025-09-02
fix(release_address): ignore invalid allocation id

soyacz:
Sometimes it may happen that address is already released when we try releasing it (e.g due external cleanup script). In that case we should ignore `InvalidAllocationID.NotFound` error.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/11629

### Testing

- [ ] - provision AWS

### Reminders


fruch:
the actual test for this needs to be with multiple networks case (i.e. longevity-50gb-3days-test), the provision test won't be getting into that code path.

maybe asking copilot to devise a unittest for this specific change ?

vponomaryov:
> the actual test for this needs to be with multiple networks case (i.e. longevity-50gb-3days-test), the provision test won't be getting into that code path.
> 
> maybe asking copilot to devise a unittest for this specific change ?

+1

The CI job where the issue happened is just 2 hours long.
So, better to try it out there.

fruch (approved):
LGTM

Issue #9170, opened 2024-10-29
Nemesis `Start` and `end` dates have significantly incorrect values


vponomaryov:
What shows Argus:
![Image](https://github.com/user-attachments/assets/3b228336-4642-4a5e-9573-08fbf74e9a1c)

The same as text:
```
+--------------------------------------+-------------------------------------------------------+
| 1. disrupt_truncate_large_partition  | long-custom-d2-wrkld1-5dc-dev-db-node-454bea32-11     |
|    Status: Succeeded                 | Start: 2024-10-28 23:16:37 | End: 2024-10-28 23:22:24 |
+--------------------------------------+-------------------------------------------------------+
| 2. disrupt_stop_start_scylla_server  | long-custom-d2-wrkld1-5dc-dev-db-node-454bea32-49     |
|    Status: Succeeded                 | Start: 2024-10-28 21:35:58 | End: 2024-10-28 21:36:48 |
+--------------------------------------+------------------------------------------------------+
```

Real time values according to the jenkins logs are following:

```
2024-10-28 17:29:01,408 f:sct_config.py   l:2057 c:sdcm.sct_config      p:INFO  > n_db_nodes: 12 12 12 12 12
2024-10-28 17:29:01,408 f:sct_config.py   l:2057 c:sdcm.sct_config      p:INFO  > n_loaders: 5 1 1 1 1
...
2024-10-28 21:02:28,434 f:tester.py       l:750  c:LongevityTest        p:INFO  > Test start time 2024-10-28 21:02:28, duration is 539 and timeout set to 2024-10-29 06:01:28
...
2024-10-28 21:43:31,370 f:loader_utils.py l:290  c:LongevityTest        p:INFO  > Prepare finished
...
...
2024-10-28 21:10:22,686 f:nemesis.py      l:1897 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.SisyphusMonkey: >>>>>>>>>>>>>Started random_disrupt_method stop_start_scylla_server
...
2024-10-28 22:23:50,978 f:nemesis.py      l:1901 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.SisyphusMonkey: <<<<<<<<<<<<<Finished random_disrupt_method stop_start_scylla_server
...
...
[...]

PR #11903, opened 2025-09-04
improvement(healthcheck): wrap healthcheck with adaptive timeout

soyacz:
In order to track the duration of healthchecks, put adaptive timeout around it.
Because healthcheck is executed right before disruption, removed healthcheck after nemesis run. This way we don't run it twice for each nemesis.

closes: https://github.com/scylladb/scylla-cluster-tests/issues/9170

### Testing

- [x] - https://argus.scylladb.com/tests/scylla-cluster-tests/05f35cf2-8715-4e22-8482-292cc7378a57/results

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4574,8 +4574,11 @@ class BaseScyllaCluster:
             # Don't run health check in case parallel nemesis.
             # TODO: find how to recognize, that nemesis on the node is running
             if self.nemesis_count == 1:
-                for node in self.nodes:
-                    node.check_node_health()
+                node_for_timeout = next((n for n in self.nodes if not n.running_nemesis), self.nodes[0])
+                with adaptive_timeout(Operations.HEALTHCHECK, node=node_for_timeout,
+                                      timeout=len(self.nodes) * 30):
+                    for node in self.nodes:
+                        node.check_node_health()
             else:
                 chc_event.message = "Test runs with parallel nemesis. Nodes health checks are disabled."
                 return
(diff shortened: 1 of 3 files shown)


sdcm/cluster.py
```diff
@@ -4574,8 +4574,10 @@ def check_cluster_health(self):
             # Don't run health check in case parallel nemesis.
             # TODO: find how to recognize, that nemesis on the node is running
             if self.nemesis_count == 1:
-                for node in self.nodes:
-                    node.check_node_health()
+                with adaptive_timeout(Operations.HEALTHCHECK, node=self.nodes[0],
```
vponomaryov, line 4577: Need to use a node without disruption.
For now, if the `self.nodes[0]` is under disruption we will get failure SSHing to it to get scylla.yaml file for is_tablets check.

pehala, line 4577: does `adaptive_timeou`t support multiple nodes? This feels as a hack solution, we are waiting for a single node but actually we need all of the nodes

sdcm/nemesis.py
```diff
@@ -5687,7 +5687,6 @@ def wrapper(*args, **kwargs):  # noqa: PLR0914, PLR0915
                         nodes_set.log_message(
                             "{end_symbol} {msg} {end_symbol}".format(end_symbol='=' * 12, msg=end_msg))
 
-            args[0].cluster.check_cluster_health()
```
pehala, line 5690: when will the health check happen then?

sdcm/cluster.py
soyacz: code wraps around healthcheck of all nodes - so we measure overall time.
Node is needed only to get some perf metrics (we don't show them in Argus).
I think there's nothing more to be done here

fruch (approved):
LGTM

PR #11927, opened 2025-09-09
improvement(actions_log): added more actions log lines

soyacz:
In order to make investigations easier, added more action logs covering more disruptions.
Also simplified the approach by inlining target and metadata to log lines.
 
closes: https://github.com/scylladb/qa-tasks/issues/1952

### Testing

- [x] - [longevity test](https://argus.scylladb.com/tests/scylla-cluster-tests/02537784-665e-4f5b-b930-be8c8f30c7f0)
- [x] - [artifact test](https://argus.scylladb.com/tests/scylla-cluster-tests/4d00490d-c47c-441a-88ed-b989d24e7692)
- [x] - [upgrade test](https://argus.scylladb.com/tests/scylla-cluster-tests/8c5b94c5-8d64-491b-bc14-76a747156b01)

### Reminders

### DIFF
--- a/artifacts_test.py
+++ b/artifacts_test.py
@@ -16,7 +16,7 @@ import re
 from contextlib import contextmanager
 from functools import cached_property
 import json
-from typing import Iterator, Any
+from typing import Iterator
 from unittest import SkipTest
 
 import yaml
@@ -65,9 +65,8 @@ class ArtifactsTest(ClusterTester):
         super().tearDown()
 
     @contextmanager
-    def logged_subtest(self, action: str,
-                       trace_id:  str | None = None, metadata: dict[str, Any] | None = None) -> Iterator[None]:
-        with self.actions_log.action_scope(action, self.node.name, trace_id, metadata):
+    def logged_subtest(self, action: str) -> Iterator[None]:
+        with self.actions_log.action_scope(action):
             with self.subTest(msg=action):
                 yield
 
@@ -80,7 +79,7 @@ class ArtifactsTest(ClusterTester):
         Validate reported version
         prev_id: check if new version is created
         """
-        self.actions_log.info("Validating scylla version in housekeepingdb", target=self.node.name)
+        self.actions_log.info("Validating scylla version in housekeepingdb")
         assert self.node.uuid, "Node UUID wasn't created"
 
         row = self.housekeeping.get_most_recent_record(query=f"SELECT id, version, ip, statuscode "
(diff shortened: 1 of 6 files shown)


soyacz:
@scylladb/qa-maintainers `actions.log` is now more comprehensive, covering all the nemesis and easier to read.
This should bring improvements in tests investigations. I tested it with artifact/upgrade/longevity tests - see attached runs to validate the contents.

Please review.

dimakr (approved):
LGTM

artifacts_test.py
```diff
@@ -65,9 +65,8 @@ def tearDown(self) -> None:
         super().tearDown()
 
     @contextmanager
-    def logged_subtest(self, action: str,
-                       trace_id:  str | None = None, metadata: dict[str, Any] | None = None) -> Iterator[None]:
-        with self.actions_log.action_scope(action, self.node.name, trace_id, metadata):
+    def logged_subtest(self, action: str) -> Iterator[None]:
```
pehala, line 68: Why not include the target and metadata in the message inside this function instead? It would remove a lot of changes and kept backwards compatibility

sdcm/nemesis.py
```diff
@@ -241,7 +240,7 @@ def __init__(self, tester_obj, termination_event, *args, nemesis_selector=None,
         self.cluster: Union[BaseCluster, BaseScyllaCluster] = tester_obj.db_cluster
         self.loaders = tester_obj.loaders
         self.monitoring_set = tester_obj.monitors
-        nemesis_thread_name = f"Nemesis{uuid4()}"
+        nemesis_thread_name = f"Nemesis-{len(getattr(self.cluster, 'nemesis', []))}"
```
pehala, line 243: why?

soyacz: its easier to read, instead nemesis-1235ac we'll get nemesis-0, nemesis-1 so we can easier track which nemesis thread it is

pehala: In that case, I would rather, if the name (or just the suffix) was passed through constructor. This seems clunky, nemesis instance should not care about other nemesis instances

soyacz: ok, I'll change it back - anyway it makes not much difference

artifacts_test.py
soyacz: basically adding curly braces, colons and additional 'target' is redundant. Also I tried to 'compress' it sometimes by changing e.g. `{"duration": "60m"}` to `for 60m` (appended).

pehala: I am fine with removing metadata, those are hard to add so they make sense grammatically, but target seems reasonable
```
if target is not None:
  action = f"{action} on node {target}"
```

soyacz: usually that fits, but not 100% time.
Basically, I already did all the work, and I'm not happy about changing it back.
As to compatibility - I'm not aware any other use than manual or mcp server (which can have backward compatibility there)

fruch (approved):
LGTM

PR #11959, opened 2025-09-14
feature(rolling-upgrades): update the base versions rules

fruch:
now that we have at least 4 short terms versions a year, we should now support upgrade from any of the STS version up to the last LTS version.

this gonna make the matrix more complex as we have more version between every LTS version

### Testing

- [x] 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu2404-test/13/pipeline-overview/ (master)
- [x] 🟢 :   https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ubuntu2404-test/14/pipeline-overview/ (upgrade to release 2025.4)

### Reminders

### DIFF
--- a/unit_tests/test_base_version.py
+++ b/unit_tests/test_base_version.py
@@ -10,10 +10,11 @@
 # See LICENSE for more details.
 #
 # Copyright (c) 2021 ScyllaDB
+from unittest.mock import patch
 
+import pytest
 
 from sdcm.utils.version_utils import ComparableScyllaVersion
-
 from utils.get_supported_scylla_base_versions import UpgradeBaseVersion
 
 """
@@ -22,13 +23,13 @@ the supported Scylla versions for upgrades based on the provided repository, bac
 """
 
 
-def general_test(scylla_repo='', linux_distro='', cloud_provider=None):
+def general_test(scylla_repo='', linux_distro='', cloud_provider=None, base_version_all_sts_versions=False):
     """
     General test function to retrieve the list of supported Scylla versions for upgrade.
     """
     scylla_version = None
 
-    version_detector = UpgradeBaseVersion(scylla_repo, linux_distro, scylla_version)
+    version_detector = UpgradeBaseVersion(scylla_repo, linux_distro, scylla_version, base_version_all_sts_versions)
     version_detector.set_start_support_version(cloud_provider)
     _, version_list = version_detector.get_version_list()
     return version_list
@@ -104,7 +105,7 @@ def test_2025_1_dev():
 
 
 def test_2025_1_release_ubuntu():
-    """ Test that 2025.1 release on ubuntu is returned correct versions """
+    """ Test that 2025.1 release on Ubuntu returns correct versions """
     scylla_repo = url_base + '/branch-2025.1/deb/unified/2025-02-16T22:46:42Z/scylladb-2025.1/scylla.list'
     linux_distro = 'ubuntu-focal'
     version_list = general_test(scylla_repo, linux_distro)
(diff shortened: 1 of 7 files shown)


unit_tests/test_base_version.py
```diff
@@ -117,3 +118,35 @@ def test_2025_1_release_centos():
     linux_distro = 'centos-9'
     version_list = general_test(scylla_repo, linux_distro)
     assert {'6.2', '2024.1', '2024.2'}.issubset(set(version_list))
+
+
+def test_2025_3_release_ubuntu():
+    """ Test that 2025.3 release on ubuntu is returned correct versions """
```
mykaul, line 124:
```suggestion
    """ Test that 2025.3 release on Ubuntu returns correct versions """
```

utils/get_supported_scylla_base_versions.py
```diff
@@ -108,29 +108,52 @@ def get_supported_scylla_base_versions(self, supported_versions: list[str]) -> l
...
-                    source_available_base_version += source_available_release_list[idx - 1:][:2]
-                elif lts_version.search(version):
-                    # for long term version we need to support last version + last LTS version
-                    source_available_base_version += source_available_release_list[idx - 1:idx]
-                    source_available_base_version.append(
-                        next((_version for _version in reversed(source_available_release_list[:idx])
-                             if lts_version.search(_version)), None))
-                else:
-                    LOGGER.warning('version format not the default - %s', version)
+            lts_version = re.compile(r'\d{4}\.1')  # lts = long term support
```
mykaul, line 112: General Q for the code below? Isn't it easier to just build the matrix - initial source, all the way to the latest?
Something like (in the future, showing 2026.1):

2024.1, 2025.1, 2025.2, 2025.3, 2025.4, 2026.1
2024.1, 2025.1, 2026.1

Where the matrix is going over every 'run' above and starts from each slot - all the way to the end?
For example with the above:
2024.1, 2025.1, 2025.2, 2025.3, 2025.4, 2026.1
2025.1, 2025.2, 2025.3, 2025.4, 2026.1
2025.2, 2025.3, 2025.4, 2026.1
2025.3, 2025.4, 2026.1
2025.4, 2026.1

2024.1, 2025.1, 2026.1
2025.1, 2026.1

fruch: I'm not sure I'm following what is the suggestion, the end result is that we should support one last LTS version and all the versions in between.

In 2026.1 we still would want to update from the 4 versions before it ? That would mean 5 variants once the first  release is out ...

fruch: @mykaul you are missing how the tests are working, it's 1 to 1 upgrades, we are not doing continues upgrades like you are assuming... 

target_version=2026.1
bases_vesrions=2025.1,2025.4

means we upgrade from 2025.1 to 2026.1, and in different run from 2025.4 to 2026.1

fruch: how can we list them not based on their suffix ?

fruch: And then we need to keep those update all the time, 

it worked so great in dtest, so we should repeat it here ?

fruch:
@roydahan @mykaul 

to sum it up
* we have a flag
 *  on releases, would make a job upgrade from all STS version up to last LTS  (see the unittest for examples)
  * on master it would update to last LTS and last STS (one or two versions)
 
we gonna enable new flag on two test: 
- AMI based (heavy test 4-5h)
- centos based test (now cent-9, I'll replace it with rocky10 in a different PR)

so now we'll be cover one of our SMI (debian based), and one centos/rhel style

I think that would be enough for initial support, later we can extend it.

roydahan (approved):

PR #11973, opened 2025-09-16
Move alternator perf tests to v17

radoslawcybulski:
Move Alternator performance regression tests to v17.
Temporarily change cadence of the trigger to daily.


vars/perfRegressionParallelPipelinebyRegion.groovy
```diff
@@ -129,7 +129,7 @@ def call(Map pipelineParams) {
                                 region: '',
                                 versions: [],
                                 sub_tests: [],
-                                labels: ['master-weekly']
+                                labels: ['master-daily']
```
fruch, line 132: @juliayakovlev FYI,

@radoslawcybulski as we discussed, this one is one every master build (which is move or less daily cadence)

fruch (approved):
LGTM

PR #12010, opened 2025-09-18
test(Manager Restore): Added object_storage_method (rclone/native)

yarongilor:
test(Manager Restore): Added object_storage_method (rclone/native)

Scylla Manager added a support for selecting the "method" parameter of restoring data from S3.
Added tests for both methods of "Native" and "rClone"
Refs: https://manager.docs.scylladb.com/master/sctool/restore.html#method

### Testing

- [2tb_1t_ics](https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/restore-benchmark-test-1tb-2tables-9nodes/9/)

### Reminders

### DIFF
--- a/configurations/object_storage.yaml
+++ b/configurations/object_storage.yaml
@@ -1,0 +1,7 @@
+append_scylla_yaml:
+  auto_snapshot: false
+  object_storage_endpoints:
+    - name: s3.us-east-1.amazonaws.com
+      port: 443
+      https: true
+      aws_region: us-east-1
--- a/jenkins-pipelines/manager/restore_benchmark/ubuntu22-manager-1tb-1t-3nodes.jenkinsfile
+++ b/jenkins-pipelines/manager/restore_benchmark/ubuntu22-manager-1tb-1t-3nodes.jenkinsfile
@@ -7,8 +7,8 @@ managerPipeline(
     backend: 'aws',
     region: 'us-east-1',
     provision_type: 'on_demand',
-    test_name: 'mgmt_cli_test.ManagerRestoreBenchmarkTests.test_restore_benchmark',
-    test_config: '''["test-cases/manager/manager-backup-restore-set-dataset.yaml", "configurations/manager/1TB_dataset.yaml"]''',
+    test_name: 'mgmt_cli_test.ManagerRestoreBenchmarkTests.test_restore_native_benchmark',
+    test_config: '''["test-cases/manager/manager-backup-restore-set-dataset.yaml", "configurations/manager/1TB_dataset.yaml", "configurations/object_storage.yaml"]''',
     mgmt_reuse_backup_snapshot_name: '1tb_1t_ics',
 
     post_behavior_db_nodes: 'destroy',
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -24,7 +24,7 @@ from sdcm.argus_results import (send_manager_benchmark_results_to_argus, send_ma
 from sdcm.mgmt import ScyllaManagerError, TaskStatus, HostStatus, HostSsl, HostRestStatus
 from sdcm.mgmt.argus_report import report_to_argus, ManagerReportType
 from sdcm.mgmt.cli import RestoreTask
-from sdcm.mgmt.common import reconfigure_scylla_manager, get_persistent_snapshots, get_backup_size
+from sdcm.mgmt.common import reconfigure_scylla_manager, get_persistent_snapshots, get_backup_size, ObjectStorageUploadMode
 from sdcm.provision.helpers.certificate import TLSAssets
 from sdcm.nemesis import MgmtRepair
 from sdcm.utils.adaptive_timeouts import adaptive_timeout, Operations
@@ -1003,7 +1003,7 @@ class ManagerRestoreBenchmarkTests(ManagerTestFunctionsMixIn):
                     resource_to_add=f"arn:aws:s3:::{location.split(':')[-1]}",
                 )
 
-    def test_backup_and_restore_only_data(self):
+    def test_backup_and_restore_only_data(self, object_storage_method: ObjectStorageUploadMode):
         """The test is extensively used for restore benchmarking purposes and consists of the following steps:
         1. Populate the cluster with data (currently operates with datasets of 500GB, 1TB, 2TB, 5TB);
         2. Run the backup task and wait for its completion;
@@ -1031,7 +1031,7 @@ class ManagerRestoreBenchmarkTests(ManagerTestFunctionsMixIn):
 
         extra_params = self.get_restore_extra_parameters()
         task = self.restore_backup_with_task(mgr_cluster=mgr_cluster, snapshot_tag=backup_task.get_snapshot_tag(),
-                                             timeout=110000, restore_data=True, extra_params=extra_params)
+                                             timeout=110000, restore_data=True, extra_params=extra_params, object_storage_method=object_storage_method)
         self.manager_test_metrics.restore_time = task.duration
 
         manager_version_timestamp = manager_tool.sctool.client_version_timestamp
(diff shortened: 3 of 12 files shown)


regevran:
> are there any additional requirements for this testing?

No. But note that for native you might want to apply more changes as [described](https://github.com/scylladb/scylladb/issues/26124) in the issue.

regevran:
> Test should allow changing them (which it does)

For `batch-size` I guess you are right, but `stream_io_throughput_mb_per_sec` should be set separately in each node's `scylla.yaml` file.

pehala (dismissed):

test-cases/manager/manager-backup-restore-set-dataset.yaml
```diff
@@ -10,6 +10,14 @@ instance_type_db: 'i4i.xlarge'
 instance_type_loader: 'c6i.xlarge'
 
 region_name: us-east-1
+append_scylla_yaml:
+  auto_snapshot: false
+  object_storage_endpoints:
+    - name: s3.us-east-1.amazonaws.com
```
fruch, line 16: if this would need to be enabled across longevity, we'll need a better way to configure it based  on `region_name`

yarongilor:
> > > For `batch-size` I guess you are right, but `stream_io_throughput_mb_per_sec` should be set separately in each node's `scylla.yaml` file.
> > 
> > 
> > That should be done automatically through the machine image changes
> 
> [retesting](https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/restore-benchmark-test-1tb-2tables-9nodes/11/) where job rerun manually added extra environment params of:
> 
> ```
> SCT_MGMT_RESTORE_EXTRA_PARAMS=--batch-size 0
> SCT_APPEND_SCYLLA_ARGS=' stream_io_throughput_mb_per_sec = 0 '
> ```
> 
> @pehala , @regevran , this execution option of adding extra env params can cover the gap for testing until we have it configured in scylla SMI. So i think we can merge the PR as is.

@regevran , @pehala , the restore test [ran](https://argus.scylladb.com/tests/scylla-cluster-tests/365996e4-77fb-4b91-b1bd-641ed2d20528) for 12 hours and failed.
Is there any known issue of Native Restore that is not fixed in latest master? should we merge this PR anyway?

test-cases/manager/manager-backup-restore-set-dataset.yaml
```diff
@@ -10,6 +10,14 @@ instance_type_db: 'i4i.xlarge'
 instance_type_loader: 'c6i.xlarge'
 
 region_name: us-east-1
+append_scylla_yaml:
```
mikliapko, line 13: What do you think about introducing `configurations/object_storage.yaml` config file and putting this object storage configuration there?
Thus, we will have more flexibility - it would allow to use the same test case with different configurations.

sdcm/mgmt/cli.py
```diff
@@ -603,7 +603,7 @@ def set_cluster_id(self, value: str):
         self.id = value
 
     def create_restore_task(self, restore_schema=False, restore_data=False, location_list=None, snapshot_tag=None,
-                            dc_mapping=None, extra_params=None):
+                            dc_mapping=None, object_storage_method=None, extra_params=None):
```
mikliapko, line 606: nit: Let's keep extra_params the last one.
I'd suggest to put object_storage_method before it.

mgmt_cli_test.py
```diff
@@ -1070,18 +1071,33 @@ def test_restore_from_precreated_backup(self, snapshot_name: str, restore_outsid
...
+        The test suggests two flows - populate the cluster with data, create the backup, and then restore it or
+        restore from a pre-created backup.
+        """
+        if reuse_snapshot_name := self.params.get('mgmt_reuse_backup_snapshot_name'):
+            self.log.info("Executing test_restore_from_precreated_backup with method rClone...")
+            self.test_restore_from_precreated_backup(
+                reuse_snapshot_name, object_storage_method=ObjectStorageUploadMode.RCLONE)
+        else:
+            self.log.info("Executing test_backup_and_restore_only_data with method rClone...")
+            self.test_backup_and_restore_only_data(object_storage_method=ObjectStorageUploadMode.RCLONE)
```
mikliapko, line 1100: Just an idea to discuss...
First of all, I don't really like the code duplication here - we have two test instances here which differs by backup/restore method configuration only, the rest is the same. 
Secondly, IMO backup/restore method is something we should configure using test configuration like we did with tablets, for example, since the expected test behavior is the same for both scenarios, just configuration is different.

Some time ago I worked on the changes for Manager to support both backup methods but abandon it 
(https://github.com/scylladb/scylla-cluster-tests/pull/11345/files) in favor of your previous backup related PR.
What do you think about the approach implemented there? 
Would it be more suitable to implement the same here considering the fact we will need to support the same across all Manager tests and methods?

jenkins-pipelines/manager/restore_benchmark/ubuntu22-manager-1tb-1t-3nodes.jenkinsfile
```diff
@@ -7,7 +7,7 @@ managerPipeline(
     backend: 'aws',
     region: 'us-east-1',
     provision_type: 'on_demand',
-    test_name: 'mgmt_cli_test.ManagerRestoreBenchmarkTests.test_restore_benchmark',
+    test_name: 'mgmt_cli_test.ManagerRestoreBenchmarkTests.test_restore_native_benchmark',
```
mikliapko, line 10: What if someone would like to execute the previous rclone test?

mikliapko (commented):
Please take a look at the comments above and let me know what do you think

sdcm/mgmt/operations.py
```diff
@@ -757,12 +757,12 @@ def restore_backup_without_manager(self, mgr_cluster, snapshot_tag, ks_tables_li
                     self.log.info(f"[Node {index}][{keyspace}.{table}] Nodetool refresh took {timer.duration}")
 
     def restore_backup_with_task(self, mgr_cluster, snapshot_tag, timeout, restore_schema=False, restore_data=False,
-                                 location_list=None, extra_params=None):
+                                 location_list=None, extra_params=None, object_storage_method=None):
```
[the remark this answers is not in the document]
yarongilor: fixed

mgmt_cli_test.py
```diff
@@ -1052,11 +1048,16 @@ def test_restore_from_precreated_backup(self, snapshot_name: str, restore_outsid
                 )
             restore_time = timer.duration
         else:
-            self.log.info("Restoring the data with standard L&S approach")
+            self.log.info("Restoring the schema")
+            self.restore_backup_with_task(mgr_cluster=mgr_cluster, snapshot_tag=snapshot_data.tag, timeout=600,
+                                          restore_schema=True, location_list=locations)
```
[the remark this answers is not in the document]
yarongilor: well i guess because the terms are confusing - `restore_outside_manager` actually uses manager..
i rolled it back and it will use the default "method" of cli restore command.
@mikliapko , is the test of `test_restore_data_without_manager` still relevant after Native support?

test-cases/manager/manager-backup-restore-set-dataset.yaml
yarongilor: This configuration is now split to a separate yaml.
So in future we can simply add entries of regions to the list of `object_storage_endpoints`

yarongilor:
Converted to draft since we agreed to roll back the review comments refactoring enhancements.

configurations/object_storage.yaml
mikliapko: I wonder if we should include `aws` into the config file name foreseeing adding GCP in the future?

jenkins-pipelines/manager/restore_benchmark/ubuntu22-manager-1tb-1t-3nodes.jenkinsfile
yarongilor: @mikliapko , the part that is rolled back is the wider refactor that introduced `object_storage_method`.
All the implementation concerns can be addressed in the future PR of the wider refactor.

configurations/object_storage.yaml
fruch: all of is should have been set by scylla by default

we have code in scylla already that can read instance metadata and get this information.

yes user might want to override some of the values

it's a bad UX for  the feature that can easily just work out of the box... 

the minimum to raise an issue for it.

regevran:
@yarongilor  - we are using this test as the "acceptance test" for native restore - I prefer having it being stable in master.
What is the expectation for merging?

pehala:
> @yarongilor - we are using this test as the "acceptance test" for native restore - I prefer having it being stable in master.
> What is the expectation for merging?

People who made comments should approve, @mikliapko and @fruch at least. I believe everything is now resolved

pehala (approved):

PR #12016, opened 2025-09-18
[Backport branch-perf-v17] Cherry pick missing commits

radoslawcybulski:
Those commits are needed by alternator regression performance tests. They exist on master and have been tested with alternator regression code. They affect only ycsb code path.

### DIFF
--- a/sdcm/stress/latte_thread.py
+++ b/sdcm/stress/latte_thread.py
@@ -34,7 +34,7 @@ from sdcm.utils.common import get_sct_root_path
 from sdcm.utils.docker_remote import RemoteDocker
 from sdcm.utils.remote_logger import HDRHistogramFileLogger
 
-LATTE_FN_NAME_RE = '(?:-f|--function)[ =]([\w\s\d:,]+)|--functions[ =]([\w\s\d:,]+)'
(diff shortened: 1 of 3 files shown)


fruch:
@radoslawcybulski  can you explain what's this PR is ? which commit you have missing, and in which branch ? v17 ?

fruch:
> I've sent you a priv message, we are missing two commits on v17 perf branch. Except i've messed up creating PR for it.
> 
> EDIT: finally i've found an option to fix base of the PR. Should be ok - cherry-picked two missing commits.

in general the best approach is to mark the original PRs as backport with the relevant labels, and the PR would be automaticlly opened,
and then we can manually amend them as needed.

radoslawcybulski:
> > I've sent you a priv message, we are missing two commits on v17 perf branch. Except i've messed up creating PR for it.
> > EDIT: finally i've found an option to fix base of the PR. Should be ok - cherry-picked two missing commits.
> 
> in general the best approach is to mark the original PRs as backport with the relevant labels, and the PR would be automaticlly opened, and then we can manually amend them as needed.

Sure, gimme a second i will find them and mark them with labels.

radoslawcybulski:
Too late. :)

PR #12024, opened 2025-09-18
feature(copilot): added AGENTS.md

soyacz:
For better Copilot/LLM Agents experience, adding AGENTS.md file with codebase basic description and common development commands.

### Testing

- [ ]

### Reminders

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,279 @@
+# AGENTS.md
+
+This file provides guidance to LLM Agents when working with code in this repository.
+
+## Repository Overview
+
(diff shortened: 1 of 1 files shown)


AGENTS.md
```diff
@@ -0,0 +1,279 @@
...
+# All unit tests
+uv run sct.py unit-tests
+
+# Specific unit test
+uv run sct.py unit-tests -t test_config.py
+```
+
+#### Run integration tests
+```bash
+uv run sct.py integration-tests
```
fruch, line 41: That would be a problem for copilot, it won't have access to AWS to run most of the tests

fruch (approved):

AGENTS.md
soyacz: but it will work when developing locally.
We can add remarks when to use it if we see it's problematic.

PR #12025, opened 2025-09-19
fix(twcs): Use immediate tombstone_gc mode

pehala:
While using TWCS it is recommended to use this mode. Reference: https://github.com/scylladb/scylladb/issues/25540

### Testing

- Longevities:
   - [ ] TWCS 3h
   - [ ] TWCS 48h 

### DIFF
--- a/test-cases/longevity/longevity-twcs-3h.yaml
+++ b/test-cases/longevity/longevity-twcs-3h.yaml
@@ -23,4 +23,4 @@ space_node_threshold: 64424
 
 user_prefix: 'longevity-twcs-3h'
 
-post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test with gc_grace_seconds = 300;"
+post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test WITH tombstone_gc = {'mode':'immediate'};"
--- a/test-cases/longevity/longevity-twcs-48h.yaml
+++ b/test-cases/longevity/longevity-twcs-48h.yaml
@@ -23,7 +23,7 @@ space_node_threshold: 64424
 
 user_prefix: 'longevity-twcs-48h'
 
-post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test with gc_grace_seconds = 12000 and default_time_to_live = 10800 and compaction = {'class':'TimeWindowCompactionStrategy', 'compaction_window_size': 60, 'compaction_window_unit': 'MINUTES'};"
+post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test and default_time_to_live = 10800 and compaction = {'class':'TimeWindowCompactionStrategy', 'compaction_window_size': 60, 'compaction_window_unit': 'MINUTES'} and WITH tombstone_gc = {'mode':'immediate'};"
 
 
 round_robin: true


raphaelsc:
is this the same longevity we were running in https://github.com/scylladb/scylladb/issues/25540?

looks like so.

juliayakovlev:
> is this the same longevity we were running in [scylladb/scylladb#25540](https://github.com/scylladb/scylladb/issues/25540)?
> 
> looks like so.

@raphaelsc 
The issue described in [scylladb/scylladb#25540] occurred during a longevity test that based on longevity TWCS 3h. 
The key difference between the tests is the number of loaders: the original TWCS 3h longevity test used 3 loaders, whereas the rack-aware version used only one.

pehala:
> * [ ]  [TWCS 3h](https://argus.scylladb.com/tests/scylla-cluster-tests/25367578-6779-4945-b1c9-e3c1d7aa65be)

Failed with this error, happened a lot during this test (900x :D). Doubt it is tied to the change here, but needs to be investigated and fixed anyway
```
2025-09-21 13:22:12.693: (ScyllaBenchLogEvent Severity.ERROR) period_type=one-time event_id=bb4c5b35-b550-4afb-a251-e3589e1402c9 during_nemesis=GrowShrinkCluster: type=ConsistencyError regex=received only line_number=35263 node=Node longevity-twcs-3h-twcs-tom-loader-node-1d179fbe-1 [34.247.191.104 | 10.4.0.63] (Type: c6i.xlarge) (rack: RACK0)
[2025-09-21 13:22:12.617] 2025/09/21 13:22:12 [query statement="INSERT INTO scylla_bench.test (pk, ck, v) VALUES (?, ?, ?)" values=[4458176053253 -1758458461213058431 <200-bytes-value>] consistency=QUORUM] || ERROR: Operation timed out for scylla_bench.test - received only 1 responses from 2 CL=QUORUM. (potentially executed: false) attempts applied: 30
```

test-cases/longevity/longevity-twcs-48h.yaml
```diff
@@ -23,7 +23,7 @@ space_node_threshold: 64424
 
 user_prefix: 'longevity-twcs-48h'
 
-post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test with gc_grace_seconds = 12000 and default_time_to_live = 10800 and compaction = {'class':'TimeWindowCompactionStrategy', 'compaction_window_size': 60, 'compaction_window_unit': 'MINUTES'};"
+post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test and default_time_to_live = 10800 and compaction = {'class':'TimeWindowCompactionStrategy', 'compaction_window_size': 60, 'compaction_window_unit': 'MINUTES'} and WITH tombstone_gc = {'mode':'immediate'};"
```
yarongilor: incorrect syntax.
should be like:
```
post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test 
WITH default_time_to_live = 10800 
AND compaction = {'class':'TimeWindowCompactionStrategy', 'compaction_window_size': 60, 'compaction_window_unit': 'MINUTES'} 
AND tombstone_gc = {'mode':'immediate'};"
```

test-cases/longevity/longevity-twcs-3h.yaml
```diff
@@ -23,4 +23,4 @@ space_node_threshold: 64424
 
 user_prefix: 'longevity-twcs-3h'
 
-post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test with gc_grace_seconds = 300;"
+post_prepare_cql_cmds: "ALTER TABLE scylla_bench.test WITH tombstone_gc = {'mode':'immediate'};"
```
yarongilor: @raphaelsc can you please advise - 
i'm not sure why setting mode 'immediate' in this non-TTL test.
Isn't this mode mainly relevant to twcs + TTL? wouldn't it be too-data-resurrection-risky user scenario?

pehala: That is on me, I switched all tests, not sure if they are both relevant.

yarongilor:
> > The [48 hours test](https://argus.scylladb.com/tests/scylla-cluster-tests/af6d62f6-f0b2-4a97-8f6f-51a31a03c795) passed ok
> 
> The [3 hours test](https://argus.scylladb.com/tests/scylla-cluster-tests/1e32e05d-6e27-4a44-95c7-32756a94444d) passed ok as well, tested with 2025.3.
> 
> So that could quite indicate the previous issues (with master latest) are unrelated to this fix.

The 3 hours test is removed from this PR due to https://github.com/scylladb/scylla-cluster-tests/pull/12025#discussion_r2524466183

pehala:
> > > The 3 hours test is removed from this PR due to [#12025 (comment)](https://github.com/scylladb/scylla-cluster-tests/pull/12025#discussion_r2524466183)
> > 
> > 
> > The original issue used this exact test yaml right? So why it doesnt matter there?
> 
> It doesn't matter, or meaningless, to set "immediate" mode for a test that doesn't have any tombstones. After testing it and verifying it passes ok with 2025.3, we can conclude that previously found [issues](https://argus.scylladb.com/tests/scylla-cluster-tests/b40a0d91-9a8e-4fa5-a471-392a5900d00e) are only related to 2026.1, not to this PR change.

https://github.com/scylladb/scylladb/issues/25540#issuecomment-3223053157 What about this? I do believe some TTL setting might be happening the test code itself, instead of only in testconfig

Issue #12032, opened 2025-09-22
Integration tests that use `docker_scylla` fixture are flaky after switching unit test to parallel execution


dimakr:
After we switched unit testing to parallel execution, some integration tests ()those that use `docker_scylla` fixture) periodically fail, e.g.:
```
15:00:45  FAILED unit_tests/test_cassandra_stress_thread.py::test_03_cassandra_stress_client_encrypt - IndexError: list index out of range
15:00:45  FAILED unit_tests/test_cql_stress_cassandra_stress_thread.py::test_02_cql_stress_cassandra_stress_multi_region - IndexError: list index out of range
15:00:45  ERROR unit_tests/test_vector_store.py::test_vector_search_error_handling - RuntimeError: Vector Store service on vs-node-None-0 failed to start
```

There is a suspicion that after switching to parallel execution we occasionally have conflicts, when some parallel integration tests try to bind to the same port when starting scylla docker container.

PR #12045, opened 2025-09-24
fix(integration tests): use docker network isoaltion for parallel tests

dimakr:
After unit tests were upgraded to run in parallel, integration tests that use docker_scylla fixture became flaky, due to parallel tests attempting to bind to the same port when starting Scylla docker container.

This change adds logic of creating dedicated docker networks in the docker_scylla fixture. This isolates docker networks during parallel tests execution, eliminating the possibility of port conflicts.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/12032

### Testing

- [ ]

### Reminders


soyacz (dismissed):
LGTM

unit_tests/conftest.py
```diff
@@ -95,30 +97,51 @@ def prom_address():
...
     # since we are going to mount it in a DinD (docker-inside-docker) setup
     base_dir = os.environ.get("_SCT_BASE_DIR", None)
     entryfile_path = Path(base_dir) if base_dir else Path(__file__).parent.parent
     entryfile_path = entryfile_path / 'docker' / 'scylla-sct' / ('entry_ssl.sh' if ssl else 'entry.sh')
 
+    docker_network = docker_scylla_args.get('docker_network') or params.get('docker_network')
+    test_uuid = None
+    if not docker_network:
+        test_uuid = str(uuid.uuid4())[:8]
+        docker_network = f"sct_test_network_{test_uuid}"
```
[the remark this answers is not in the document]
dimakr: We discussed this offline, that maybe switching to isolated docker networks is not needed here (docker containers have their own addresses and container ports are not bind to host ports). Also such change might cause adverse effects later (e.g. when we want to turn the kafka tests back).

Making step back to revert the changes around docker networks. And will address exact tests that are flaky when accessing shared resources.

unit_tests/conftest.py
```diff
@@ -129,16 +146,14 @@ def configure_scylla_node(docker_scylla_args: dict, params):  # noqa: PLR0914
...
-        finally:
-            os.chdir(curr_dir)
+        create_certificate(ssl_dir / TLSAssets.DB_CLIENT_FACING_CERT, ssl_dir / TLSAssets.DB_CLIENT_FACING_KEY,
+                           cname="scylladb", ca_cert_file=CA_CERT_FILE, ca_key_file=CA_KEY_FILE,
+                           ip_addresses=[scylla.internal_ip_address], dns_names=[scylla.public_dns_name])
+        create_certificate(ssl_dir / TLSAssets.CLIENT_CERT, ssl_dir / TLSAssets.CLIENT_KEY,
+                           cname="scylladb", ca_cert_file=CA_CERT_FILE, ca_key_file=CA_KEY_FILE)
+
+        request.node.stash[ssl_dir_key] = ssl_dir
+        scylla.__class__.ssl_conf_dir = property(lambda self: ssl_dir)
```
pehala, line 156: Do we really have no better way of adding it here? AFAIK, this is our object, cannot we pass it through a method or a constructor?

unit_tests/conftest.py
```diff
@@ -129,16 +146,14 @@ def configure_scylla_node(docker_scylla_args: dict, params):  # noqa: PLR0914
...
-                               ip_addresses=[scylla.ip_address], dns_names=[scylla.public_dns_name])
-            create_certificate(CLIENT_CERT_FILE, CLIENT_KEY_FILE, cname="scylladb",
-                               ca_cert_file=CA_CERT_FILE, ca_key_file=CA_KEY_FILE)
-        finally:
-            os.chdir(curr_dir)
+        create_certificate(ssl_dir / TLSAssets.DB_CLIENT_FACING_CERT, ssl_dir / TLSAssets.DB_CLIENT_FACING_KEY,
+                           cname="scylladb", ca_cert_file=CA_CERT_FILE, ca_key_file=CA_KEY_FILE,
+                           ip_addresses=[scylla.internal_ip_address], dns_names=[scylla.public_dns_name])
+        create_certificate(ssl_dir / TLSAssets.CLIENT_CERT, ssl_dir / TLSAssets.CLIENT_KEY,
+                           cname="scylladb", ca_cert_file=CA_CERT_FILE, ca_key_file=CA_KEY_FILE)
```
pehala, line 153: Why do we have to create SSL certs only after creating scylla?

unit_tests/test_latte_thread.py
```diff
@@ -120,6 +121,11 @@ def test_04_latte_run_client_encrypt(request, docker_scylla, params):
 
     loader_set = LocalLoaderSetDummy(params=params)
 
+    # dedicated SSL certs directory for the test, to avoid conflicts during parallel tests execution
+    if ssl_dir := request.node.stash.get(ssl_dir_key, None):
+        for loader_node in loader_set.nodes:
+            loader_node.__class__.ssl_conf_dir = property(lambda self: ssl_dir)
```
pehala, line 127: Cannot we pass it like a parameter to LocalLoaderSetDummy? It is a testing object, so we can easily change it

unit_tests/conftest.py
```diff
@@ -127,18 +150,17 @@ def configure_scylla_node(docker_scylla_args: dict, params):  # noqa: PLR0914
...
+        if test_uuid:
+            scylla.test_ssl_dir = ssl_dir
```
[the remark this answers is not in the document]
pehala: I believe there is a way how to make it into a fixture (or more precise only function here) and not require any stash. I created PoC https://github.com/pehala/scylla-cluster-tests/commit/d3cdcd2ffeb3abded6644e3d5d18e08e3a0d5458 to demonstrate this.

I believe this way is cleaner, though it can be much more clean.

soyacz:
@dimakr can you rebase to see all integration tests passing (one failing you've already fixed on master)

unit_tests/conftest.py
pehala: I think this is still valid. Its our class, we can add the option to pass ssl_conf_dir in a nicer way

PR #12066, opened 2025-09-29
Increate future timeout

radoslawcybulski:
Increase future's timeout, as current is too low for alternator's performance testing.

YCSB produces mutliple files (each with it's own tag), while hdr histogram processing in sct does scan directory for all hdr files and run a different future over each file x tag combination. As a result each hdr histogram from ycsb will be processed 6 times (one for each tag).
This change is temporary.


juliayakovlev:
> > @radoslawcybulski if there are change you want todo, please do them on master, and mark it for backport, don't do thing straight into v17...
> 
> My bad, updated. `backport/perf-v17` asks for backporting into `v17`?

yes

fruch (approved):
LGTM

PR #12091, opened 2025-10-01
chore(argus): Update client to 0.15.6

k0machi:
This is an update to the client that brings in new event system.


soyacz (approved):
LGTM

soyacz:
@fruch we already test events based on this PR and all seems to work

Issue #12092, opened 2025-10-02
Monitoring stack were not collected, Monitoring version was not found


cezarmoise:
Example: https://jenkins.scylladb.com/job/scylla-master/job/tier1/job/longevity-50gb-3days-test/94/console 

```
06:55:34  Start collect logs for cluster monitor-set
06:55:34  Collecting logs on host: longevity-tls-50gb-3d-master-monitor-node-5b06eee6-1
06:55:39  /home/ubuntu/scylla-monitoring-data/snapshots/20250928T035536Z-67df295f3adee6f6
06:55:39  Detected Linux distribution: UBUNTU22
06:55:47  /home/ubuntu/scylla-monitoring-data/snapshots/prometheus_data_20250928_034359.tar.zst
06:55:49  scylla-grafana-monitoring-scylla-monitoring, None, None
06:55:49  Monitoring stack were not collected
06:55:49  Monitoring version was not found
06:55:50  Uploading 'monitor-set-5b06eee6.tar.zst' to https://cloudius-jenkins-test.s3.amazonaws.com/5b06eee6-efed-402a-bb0c-9996d07469f7/20250928_034359/monitor-set-5b06eee6.tar.zst
06:55:59  Uploaded to https://cloudius-jenkins-test.s3.amazonaws.com/5b06eee6-efed-402a-bb0c-9996d07469f7/20250928_034359/monitor-set-5b06eee6.tar.zst
06:55:59  collected data for monitor-set
06:55:59  ['https://cloudius-jenkins-test.s3.amazonaws.com/5b06eee6-efed-402a-bb0c-9996d07469f7/20250928_034359/monitor-set-5b06eee6.tar.zst']
```

And the resulting archinve does not contain the monitoring stack

<img width="317" height="181" alt="Image" src="https://github.com/user-attachments/assets/ace46893-df40-4cd3-86b9-0131e0df71ac" />

PR #12093, opened 2025-10-02
fix(logcollector): fix getting monitoring version

cezarmoise:
Sometimes, the contents of the `monitor_version` file can be something like `None:master:latest`, which causes the split to fail. Since that value is only used in the archive file name, it is safe to ignore the last part.

closes: #12092

```
ubuntu@balancer-test-feature--monitor-node-72f249d1-1:~$ cat scylla-grafana-monitoring-scylla-monitoring/monitor_version
None:master
```
vs
```
ubuntu@elasticity-test-feature--monitor-node-f2335597-1:~$ cat scylla-grafana-monitoring-scylla-monitoring/monitor_version
None:master:latest
```

### Testing

- [x] N/A

### Reminders

### DIFF
--- a/sdcm/logcollector.py
+++ b/sdcm/logcollector.py
@@ -197,7 +197,7 @@ class BaseMonitoringEntity(BaseLogEntity):
             return None, None, None
 
         try:
-            monitor_version, scylla_version = result.stdout.strip().split(':')
+            monitor_version, scylla_version = result.stdout.strip().split(':')[:2]
         except ValueError:
             monitor_version = None
             scylla_version = None


pehala (approved):

cezarmoise:
This code is weird because it returns
```python
return name, monitor_version, scylla_version
```
but when it is used
```python
monitor_install_dir_name, monitor_branch, monitor_version = self.get_monitoring_version(node)
```
so I would expect if to be `latest:master` instead of `None:master:latest`

vponomaryov (approved):
LGTM

cezarmoise:
> LGTM, but do we know when this file can have such contents?

I noticed that when that happens, there was an extra `master:latest` in Grafana, that does not have contents.

image

PR #12121, opened 2025-10-09
fix(sdcm/utils/common.py): Avoid permission denied on strict umask environment

### LINKED ISSUE #12120: In a strict umask environment, `tar xvfz /tmp/keyspace1.standard1.tar.gz` results in Permission denied
While I'm testing strict umask patch on machine-image (scylladb/scylla-machine-image#712), I have found that longevity test failed with following error, which never happens without this patch:
```
20:24:39  E           Command: 'sudo -u scylla tar xvfz /tmp/keyspace1.standard1.tar.gz -C /var/lib/scylla/data/keyspace1/standard1-f8462f509d0211f09d5fcaf55ccce0c7/upload/'
20:24:39  E           Exit code: 2
20:24:39  E           Stdout:
20:24:39  E           Stderr:
20:24:39  E           tar (child): /tmp/keyspace1.standard1.tar.gz: Cannot open: Permission denied
20:24:39  E           tar (child): Error is not recoverable: exiting now
20:24:39  E           tar: Child returned status 2
20:24:39  E           tar: Error is not recoverable: exiting now
```
https://jenkins.scylladb.com/view/master/job/scylla-master/job/releng-testing/job/longevity/job/longevity-100gb-4h-test/91/consoleFull#-339195974fcc21424-66d2-4bd8-8e0d-9746405e5b16

This is likely happened because `upload_sstables()` [downloads the tar file using curl command,](https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/utils/common.py#L134) and it creates new file in default permission.
With strict umask 027, the permission becomes 640.
Since the permission does not have other-read bit, tar command won't be able to read the file because it running as scylla user (`sudo -u scylla`).

It looks like curl command itself does not have a parameter to set permission on output file, we need to run `chmod 644 <output file>` after curl command executed.

syuu1228:
In a strict umask environment (umask 027), the curl command executed by remote_get_file() sets the permissions of the downloaded file to 640. This causes Permission denied error on upload_sstables(), since it tries to extract downloaded tar file in scylla user, but the file does not have other-read bit on its permission.

To fix the problem, we need to set the file permission to 644 just after curl command executed.

Fixes #12120
Related https://github.com/scylladb/scylla-machine-image/pull/712

### Testing

Tested, it still fails with another error but verified Permission denied error disappear:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/takuya/job/longevity/job/longevity-100gb-4h-test/2/

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -134,6 +134,7 @@ def _remote_get_file(remoter, src, dst, user_agent=None):
     cmd = 'curl -L {} -o {}'.format(src, dst)
     if user_agent:
         cmd += ' --user-agent %s' % user_agent
+    cmd += f' && chmod 644 {dst}'
     return remoter.run(cmd, ignore_status=True)
 
 


soyacz (approved):

dimakr (approved):
LGTM

PR #12163, opened 2025-10-13
Vector Search in Cloud testing

mikliapko:
### Add tests for Vector Search in Cloud

The change introduces a few tests for Vector Search verification in Scylla Cloud. The tests are based on old siren-tests + SCT integration and expect Cloud cluster to be prepared in advance by siren-tests.

New tests include:
- sanity checks for VS (add some Vector data into cluster, build index and run some ANN search request for verification)
- replace node scenarios to make sure VS node can be replaced without any lose of availability for healthy node and with some loses for broken node

sct_config.py extended with two new parameters which comes from siren-tests and can be retrieved in SCT currently:
- `vector_search_node_ids` - IDs of deployed VS nodes
- `vector_search_node_type_id` - ID of deployed VS nodes

In addition to tests, a new email reporter added to serve VS tests.

### Extend schema parsing and handling helpers
- Added new function that extracts schema creation statements from file
- Added new function to substitute datacenter name in keyspace statement to match the current cluster's DC name
- Since parse schema entries functionality is used in two function now, it's moved to a separate function `_parse_schema_entries`

### Changes

| File | Description |
| ---- | ----------- |
| test_vector_search_in_cloud.py | Main test implementation with vector search testing logic and recall validation |
| sdcm/send_email.py | Email reporter class for vector search test results |
| sdcm/sct_config.py | Configuration options for vector search node management |
| sdcm/report_templates/results_vector_search_in_cloud.html | HTML template for vector search test reporting |
| sdcm/mgmt/helpers.py | Schema management utilities with DC name substitution support |
| docs/configuration_options.md | Documentation for new vector search configuration parameters |

### Testing

- [ ]

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -2490,6 +2490,24 @@ Time in seconds it took Siren to bootstrap 1-1-restore cluster
 **type:** int
 
 
+## **vector_search_node_ids** / SCT_VECTOR_SEARCH_NODE_IDS
+
+Vector Search node ids
+
+**default:** N/A
+
+**type:** list
+
+
+## **vector_search_node_type_id** / SCT_VECTOR_SEARCH_TYPE_IDS
+
+Vector Search nodes type id
+
+**default:** N/A
+
+**type:** int
+
+
 ## **stress_cmd_w** / SCT_STRESS_CMD_W
 
 cassandra-stress commands.<br>You can specify everything but the -node parameter, which is going to<br>be provided by the test suite infrastructure.<br>multiple commands can passed as a list
--- a/test_vector_search_in_cloud.py
+++ b/test_vector_search_in_cloud.py
@@ -1,0 +1,274 @@
+import ast
+import random
+import time
+from functools import cached_property
+from threading import Thread
+from typing import Optional
+
+import cassandra
+
+from sdcm.mgmt.helpers import get_schema_create_statements_from_file, substitute_dc_name_in_ks_statements_if_different
+from sdcm.tester import ClusterTester
+from sdcm.utils.common import S3Storage
+from sdcm.utils.file import File
+
+# Vector Search test data files
+BUCKET_NAME = "vector-store-in-cloud"
+
+TEST_DATA_FILENAME = "test_data.txt"
+GROUND_TRUTH_FILENAME = "ground_truth.txt"
+DATASET_FILENAME = "dataset.txt"
+SCHEMA_FILENAME = "schema.json"
+
+DATA_DIR_PATH = "data_dir/vector_search/"
+TEST_DATA_FILE_PATH = DATA_DIR_PATH + TEST_DATA_FILENAME
+GROUND_TRUTH_FILE_PATH = DATA_DIR_PATH + GROUND_TRUTH_FILENAME
+DATASET_FILE_PATH = DATA_DIR_PATH + DATASET_FILENAME
+SCHEMA_FILE_PATH = DATA_DIR_PATH + SCHEMA_FILENAME
+
+ANN_SEARCH_CMD_TEMPLATE = "SELECT id FROM vdb_bench.vdb_bench_collection ORDER BY vector ANN OF {vector} LIMIT 10;"
+
+VECTOR_SEARCH_INSTANCE_TYPES = {
+    "aws": {"t4g.small": 175, "t4g.medium": 176, "r7g.medium": 177},
+    "gce": {"e2-small": 178, "e2-medium": 179, "n4-highmem-2": 180}
+}
+
+
+class VectorSearchInCloudBase(ClusterTester):
+    average_recall: float = 0.0
+
+    def get_email_data(self):
+        self.log.info("Prepare data for email")
+
+        email_data = self._get_common_email_data()
+        email_data.update({"average_recall": self.average_recall})
+
+        return email_data
+
+    @staticmethod
+    def download_vector_search_test_data_from_s3():
+        s3_storage = S3Storage(bucket=BUCKET_NAME)
+
+        base_url = f"https://{BUCKET_NAME}.s3.amazonaws.com"
+        s3_storage.download_file(link=f"{base_url}/{TEST_DATA_FILENAME}", dst_dir=DATA_DIR_PATH)
+        s3_storage.download_file(link=f"{base_url}/{GROUND_TRUTH_FILENAME}", dst_dir=DATA_DIR_PATH)
+
+        s3_storage.download_file(link=f"{base_url}/{SCHEMA_FILENAME}", dst_dir=DATA_DIR_PATH)
+        s3_storage.download_file(link=f"{base_url}/{DATASET_FILENAME}", dst_dir=DATA_DIR_PATH)
+
+    @cached_property
+    def test_data(self):
+        return File(TEST_DATA_FILE_PATH).readlines()
+
+    @cached_property
+    def ground_truth_data(self):
+        return File(GROUND_TRUTH_FILE_PATH).readlines()
+
+    def prepare_vector_search_index(self):
+        self.log.info("Download Vector Search test data from S3")
+        self.download_vector_search_test_data_from_s3()
+
+        self.log.info("Create schema applying CQL statements")
+        ks_statements, other_statements = get_schema_create_statements_from_file(schema_file_path=SCHEMA_FILE_PATH)
+
+        # DC name written in the schema file may differ from the one used in the cluster under test
+        # if run in region or cloud provider different from ones where backup was taken
+        dc_under_test_name = next(iter(self.db_cluster.get_nodetool_status()))
+        ks_statements = substitute_dc_name_in_ks_statements_if_different(ks_statements=ks_statements,
+                                                                         current_dc_name=dc_under_test_name)
+
+        with self.db_cluster.cql_connection_patient(node=self.db_cluster.nodes[0], verbose=False) as session:
+            for cql_stmt in ks_statements + other_statements:
+                session.execute(cql_stmt)
+
+            self.log.info("Insert sample data into the table")
+            insert_cmd_template = "INSERT INTO vdb_bench.vdb_bench_collection (id, vector) VALUES ({id}, {vector});"
+            with open(DATASET_FILE_PATH, encoding="utf-8") as dataset_file:
+                for line in dataset_file:
+                    id_str, vector_str = line.strip().split(",", maxsplit=1)
+                    insert_cmd = insert_cmd_template.format(id=int(id_str), vector=vector_str)
+                    session.execute(insert_cmd)
+
+        self.log.info("Create Vector Index")
+        with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
+            query = ("CREATE CUSTOM INDEX vdb_bench_collection_vector_idx ON vdb_bench.vdb_bench_collection(vector) "
+                     "USING 'vector_index'")
+            session.execute(query)
+
+        # TODO: rework when it'll be possible to check index readiness status
+        self.log.info("Sleep 200 seconds for indexes to be built")
+        time.sleep(200)
+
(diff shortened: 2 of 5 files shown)


test_vector_search_in_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+    def test_vector_search_in_cloud_sanity(self):
+        self.log.info("Prepare Vector Search data")
+        self.prepare_vector_search_index()
+
+        self.log.info("Run Vector Search verification queries")
+        self.test_vector_search_recall(queries_num=100)
+
+
+class VectorSearchInCloudReplaceNode(VectorSearchInCloudBase):
+    def _test_replace_healthy_node(self,
```
fruch: Why not use existing nemesis for those things ? why there a need to reimplment those ?

test_vector_search_in_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+    def get_email_data(self):
+        self.log.info("Prepare data for email")
+
+        email_data = self._get_common_email_data()
+        email_data.update({"average_recall": self.average_recall})
+
+        return email_data
+
+    @staticmethod
+    def download_vector_search_test_data_from_s3():
```
fruch: I would recommend putting this into the same place we have other sstable, we already have the code for downloading the files, and putting loading them into scylla, is should be duplicated

test_vector_search_in_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+
+        with self.db_cluster.cql_connection_patient(node=self.db_cluster.nodes[0], verbose=False) as session:
+            for cql_stmt in ks_statements + other_statements:
+                session.execute(cql_stmt)
+
+            self.log.info("Insert sample data into the table")
+            insert_cmd_template = "INSERT INTO vdb_bench.vdb_bench_collection (id, vector) VALUES ({id}, {vector});"
+            with open(DATASET_FILE_PATH, encoding="utf-8") as dataset_file:
+                for line in dataset_file:
+                    id_str, vector_str = line.strip().split(",", maxsplit=1)
```
fruch: there better way to read CSV files..

also we should consider using cqlsh for such a thing. (as we instruct costumers)

fruch:
@mikliapko 

is there some design or test plan to point to, since this PR doesn't have much of the reasoning for those tests

docs/configuration_options.md
```diff
@@ -2490,6 +2490,24 @@ Time in seconds it took Siren to bootstrap 1-1-restore cluster
...
+## **vector_search_node_ids** / SCT_VECTOR_SEARCH_NODE_IDS
+
+Vector Search node ids
+
+**default:** N/A
+
+**type:** list
+
+
+## **vector_search_node_type_id** / SCT_VECTOR_SEARCH_TYPE_IDS
```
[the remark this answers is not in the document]
mikliapko: I'll check

test_vector_search_in_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+                    insert_cmd = insert_cmd_template.format(id=int(id_str), vector=vector_str)
+                    session.execute(insert_cmd)
+
+        self.log.info("Create Vector Index")
+        with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
+            query = ("CREATE CUSTOM INDEX vdb_bench_collection_vector_idx ON vdb_bench.vdb_bench_collection(vector) "
+                     "USING 'vector_index'")
+            session.execute(query)
+
+        # TODO: rework when it'll be possible to check index readiness status
```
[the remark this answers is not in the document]
mikliapko: We don't have it exposed for the Cloud yet

docs/configuration_options.md
```diff
@@ -2490,6 +2490,24 @@ Time in seconds it took Siren to bootstrap 1-1-restore cluster
 **type:** int
 
 
+## **vector_search_node_ids** / SCT_VECTOR_SEARCH_NODE_IDS
```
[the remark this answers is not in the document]
fruch: @dimakr has a task for doing this integration, sync with him about it

test_vector_search_in_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+DATA_DIR_PATH = "data_dir/vector_search/"
+TEST_DATA_FILE_PATH = DATA_DIR_PATH + TEST_DATA_FILENAME
+GROUND_TRUTH_FILE_PATH = DATA_DIR_PATH + GROUND_TRUTH_FILENAME
+DATASET_FILE_PATH = DATA_DIR_PATH + DATASET_FILENAME
+SCHEMA_FILE_PATH = DATA_DIR_PATH + SCHEMA_FILENAME
+
+ANN_SEARCH_CMD_TEMPLATE = "SELECT id FROM vdb_bench.vdb_bench_collection ORDER BY vector ANN OF {vector} LIMIT 10;"
+
+VECTOR_SEARCH_INSTANCE_TYPES = {
+    "aws": {"t4g.small": 175, "t4g.medium": 176, "r7g.medium": 177},
```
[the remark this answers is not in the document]
fruch: why do we need those instances ?

test_vector_search_in_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+DATASET_FILENAME = "dataset.txt"
+SCHEMA_FILENAME = "schema.json"
+
+DATA_DIR_PATH = "data_dir/vector_search/"
+TEST_DATA_FILE_PATH = DATA_DIR_PATH + TEST_DATA_FILENAME
+GROUND_TRUTH_FILE_PATH = DATA_DIR_PATH + GROUND_TRUTH_FILENAME
+DATASET_FILE_PATH = DATA_DIR_PATH + DATASET_FILENAME
+SCHEMA_FILE_PATH = DATA_DIR_PATH + SCHEMA_FILENAME
+
+ANN_SEARCH_CMD_TEMPLATE = "SELECT id FROM vdb_bench.vdb_bench_collection ORDER BY vector ANN OF {vector} LIMIT 10;"
```
fruch: @mikliapko I would like to see all of the logic and CQL commands needed for validation, backed into latte rune script.

so we can scale it up, and measure pref, and build substantial  cases, for 5min cases  we can spin up dtest environment, of vector store own integration tests can cover that , we can do some basic functional test in SCT, but it's far from being the main goal

test_vector_search_in_cloud.py
```diff
@@ -0,0 +1,274 @@
...
+    @cached_property
+    def ground_truth_data(self):
+        return File(GROUND_TRUTH_FILE_PATH).readlines()
+
+    def prepare_vector_search_index(self):
+        self.log.info("Download Vector Search test data from S3")
+        self.download_vector_search_test_data_from_s3()
+
+        self.log.info("Create schema applying CQL statements")
+        ks_statements, other_statements = get_schema_create_statements_from_file(schema_file_path=SCHEMA_FILE_PATH)
```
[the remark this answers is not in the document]
mikliapko: I think it makes sense, will rework

test_vector_search_in_cloud.py
fruch: https://docs.scylladb.com/manual/stable/cql/cqlsh.html#copy-from

docs/configuration_options.md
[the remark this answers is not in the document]
mikliapko: I wonder if we can switch to SCT xcloud clusters here instead of utilization of old siren-tests + SCT integration.
The one limitation I see is utilization of siren CLI for VS node replacement. 
AFAIK siren CLI is out of scope for xcloud mode and won't be integrated into SCT in any way.

test_vector_search_in_cloud.py
mikliapko: @soyacz Could you please share the plan you've been working on?

mikliapko:
Just wanna sum up on all the comments left above and define the next steps on how to introduce this tests into SCT.

First of all, good news - I was able to run `test_vector_search_in_cloud_sanity` test using XCloud cluster. 
We can get rid of old siren-tests + SCT integration while testing with Cloud.

Secondly, latte utilization - haven't tried it yet, but it sounds reasonable and I'll check how to implement all CQL logic using this tool.

The third major thing mentioned during review is how to organize these tests - should it be part of nemesis or a separate test, etc.
Our initial goal here was to have a pretty simple E2E test that would allow to verify Vector Search in Cloud, to make sure all components are integrated and work as expected. There was no goal to run some complex nemesis or performance tests for the first VS milestone in Cloud.

So, here I still don't have a full picture and understanding for to proceed.

@fruch I'd appreciate your feedback here.
@ewienik I know you are working on VS-related SCT tests for core. Could you please share your thoughts?

P.S. I don't touch minor things highlighted during review here. They are not significant for general approach and would be easily addressed.

test_vector_search_in_cloud.py
fruch: https://github.com/scylladb/scylla-cluster-tests/blob/c8c8487db5ec5adc77e2bc5b2859566d5a7d7ede/sdcm/utils/sstable/load_utils.py#L19

fruch:
> Just wanna sum up on all the comments left above and define the next steps on how to introduce this tests into SCT.
> 
> First of all, good news - I was able to run `test_vector_search_in_cloud_sanity` test using XCloud cluster. 
> We can get rid of old siren-tests + SCT integration while testing with Cloud.

Very good to see it fit into place, according to what we planned it for.

> 
> Secondly, latte utilization - haven't tried it yet, but it sounds reasonable and I'll check how to implement all CQL logic using this tool.

Get in touch with @vponomaryov ,he can help you sketch some initial example that might be suitable.
(And also discuss how we can utilize some static data from files, we didn't used it at all so far)

> 
> The third major thing mentioned during review is how to organize these tests - should it be part of nemesis or a separate test, etc.
> Our initial goal here was to have a pretty simple E2E test that would allow to verify Vector Search in Cloud, to make sure all components are integrated and work as expected. There was no goal to run some complex nemesis or performance tests for the first VS milestone in Cloud.
> 
> So, here I still don't have a full picture and understanding for to proceed.
> 
> @fruch I'd appreciate your feedback here.
> @ewienik I know you are working on VS-related SCT tests for core. Could you please share your thoughts

This is exactly the concern I have,
I'm o.k. with basic sanity tests as a first step, I'm less o.k. with tests that are tied to a specific "backend", i.e. scylla-cloud specific in our case.

I think we will need to test VS with and without scylla-cloud

As for nemesis based, we can build small cases using nemesis, that can act as sanity case. The key element that if we build it like that, we then can reuse the new nemesis (like restart of VS service / node) all over in other more complex use cases, not that we should do the complex cases right now.

Given all that, those are my recommendations, if we would know roughly what  the VS team is going to test with SCT (and when), we can define certain stages that things can be refactored. for example (made up one, cause I didn't see any specific plan for those), converting sanity the restart VS node, in a nemesis based to a nemesis based case in preparation for longevity or performance

> 
> P.S. I don't touch minor things highlighted during review here. They are not significant for general approach and would be easily addressed.

ewienik:
> > Given all that, those are my recommendations, if we would know roughly what the VS team is going to test with SCT (and when), we can define certain stages that things can be refactored. for example (made up one, cause I didn't see any specific plan for those), converting sanity the restart VS node, in a nemesis based to a nemesis based case in preparation for longevity or performance
> 
> @ewienik Could you please share your thoughts on that?

Finally I described an initial version of our goals for SCT in https://scylladb.atlassian.net/wiki/spaces/RND/pages/39518215/Vector+Search+Test+Plan+Document. IIUC we need a longevity tests in some basic scenarios: building index and ann searching. The system should correctly behaves when a scylla node or vector-store node disconnects/reconnects. During these nemesis we should be able to measure vector index build time, qps, latency and recall. There should be also test for ann search with background traffic in scylladb.

fruch:
> > You can do some right now, assuming you have ssh access to the VS node, like restart of the service.
> 
> Does XCloud approach already support ssh connection to DB and Vector nodes?

I think @dimakr did it from the beginning, but no code path was using it. cross check with him.

One caveat, ssh is enabled only on staging, we need to finalize more things to have it available on dev and production

Issue #12105, opened 2025-10-07
MgmtCorruptThenRepair:  Materialized views and secondary indexes are not supported on base tables with tablets


Lakshmipathi:
Latest run failed during `MgmtCorruptThenRepair` with error like.

```
2025-10-07 01:25:34.129: (DisruptionEvent Severity.ERROR) period_type=end event_id=9e438dc6-1482-4839-ab7a-0af70fe113cb duration=8s: nemesis_name=MgmtCorruptThenRepair target_node=Node longevity-100gb-4h-master-db-node-da623308-6 [98.89.28.95 | 10.12.8.31] (Type: i4i.4xlarge) (rack: RACK2) errors=Error from server: code=2200 [Invalid query] message="Materialized views and secondary indexes are not supported on base tables with tablets. To be able to use them, enable the configuration option `rf_rack_valid_keyspaces` and make sure that the cluster feature `VIEWS_WITH_TABLETS` is enabled."
Traceback (most recent call last):
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5826, in wrapper
    result = method(*args, **kwargs)
  File "/home/ubuntu/scylla-cluster-tests/sdcm/nemesis.py", line 5675, in disrupt_kill_mv_building_coordinator
    create_materialized_view_for_random_column(session, ks_name, base_table_name, view_name)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/ubuntu/scylla-cluster-tests/sdcm/utils/nemesis_utils/indexes.py", line 222, in create_materialized_view_for_random_column
    create_materialized_view(session, keyspace_name, base_table_name, view_name, [column],
    ~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
                             primary_key_columns,
                             ^^^^^^^^^^^^^^^^^^^^
                             mv_columns=[column] + primary_key_columns)
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #12164, opened 2025-10-13
fix(config): Add rf_rack_valid_keyspaces config option

Lakshmipathi:
Fixes https://github.com/scylladb/scylla-cluster-tests/issues/12105
### Testing

- [x ] https://argus.scylladb.com/tests/scylla-cluster-tests/d1261c3a-ce39-4af8-b20a-6cf778cd44ca

- [x ] I added the relevant `backport` labels
- [ x] I didn't leave commented-out/debugging code

### Reminders

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -964,7 +964,7 @@ More arguments to append to oracle command line
 
 More configuration to append to /etc/scylla/scylla.yaml
 
-**default:** N/A
+**default:** {'rf_rack_valid_keyspaces': True}
 
 **type:** dict_or_str
 
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -17,6 +17,8 @@ round_robin: false
 append_scylla_args: '--blocked-reactor-notify-ms 25 --abort-on-lsa-bad-alloc 1 --abort-on-seastar-bad-alloc --abort-on-internal-error 1 --abort-on-ebadf 1 --enable-sstable-key-validation 1'
 append_scylla_setup_args: ''
 append_scylla_node_exporter_args: ''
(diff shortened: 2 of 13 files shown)


docs/configuration_options.md
```diff
@@ -3489,6 +3489,15 @@ Forces GossipingPropertyFileSnitch (regardless `endpoint_snitch`) to simulate ra
 **type:** int
 
 
+## **rf_rack_valid_keyspaces** / SCT_RF_RACK_VALID_KEYSPACES
```
fruch: Why do we need to introduce SCT configuration for a property in scylla yaml ?

Won't just changing the default scylla_yaml_append suffice?

I don't like the direction of adding so much for a single scylla configuration

pehala:
> @pehala I'm a bit scared of the effect this one would might have, not sure if we configured racks correctly yet in all tests...

It is required for any MV testing, which is crucial for release testing and on master as it is a new feature. If other tests start failing, we will need to fix them (since we are in release period, we are running more of them). I do have hope that it will minority of them though.

docs/configuration_options.md
pehala: Right, what will hapeen with entry like this, will it be merged or overriden? 

```
 hinted_handoff: 'enabled'
 parallel_node_operations: true

append_scylla_yaml:
  yet_another_option: false
```

Lakshmipathi: Checked now, the `yet_another_option` gets appended to yaml along with `rf_rack_valid_keyspaces: true` when we use

```
 hinted_handoff: 'enabled'
 parallel_node_operations: true

append_scylla_yaml:
  yet_another_option: false
```

PR #12179, opened 2025-10-15
fix(ci): remove creation of MVs in test config used for release testing

vponomaryov:
The MV feature for tablets is immature in `2025.3` and should not be used.
Hence, disable it in the test config used for patch versions testing:
- `test-cases/longevity/longevity-large-partition-3h.yaml`

### Testing

- [ ]

### Reminders


vponomaryov:
> how about testing vnodes?

This test case runs with tablets only.
Without this change automatic trigger will waste resources and I will need to re-trigger it manually setting the `SCT_POST_PREPARE_CQL_CMDS=`  extra environment variable.

PR #12180, opened 2025-10-15
feature(log_transport): default to `vector`

fruch:
we have it working for tier1 for couple of months now, without any issues what so ever.

so now in this commit we make it the default log transport (and remove all the tier1 overiddes)

### Testing

- [x] provision tests

### Reminders

### DIFF
--- a/sdcm/sct_provision/user_data_objects/vector_dev.py
+++ b/sdcm/sct_provision/user_data_objects/vector_dev.py
@@ -12,7 +12,7 @@
 # Copyright (c) 2025 ScyllaDB
 from dataclasses import dataclass
 
-from sdcm.provision.common.utils import configure_vector_target_script, install_vector_service
+from sdcm.provision.common.utils import configure_vector_target_script, install_vector_service, configure_backoff_timeout
 from sdcm.sct_provision.user_data_objects import SctUserDataObject
 
 
@@ -25,7 +25,8 @@ class VectorDevUserDataObject(SctUserDataObject):
 
     @property
     def script_to_run(self) -> str:
-        script = install_vector_service()
+        script = configure_backoff_timeout()
+        script += install_vector_service()
         host, port = self.test_config.get_logging_service_host_port()
         script += configure_vector_target_script(host=host, port=port)
         return script


vponomaryov (dismissed):
LGTM

soyacz (dismissed):

fruch:
@soyacz seems like the Azure provision test is breaking badly, can you take a look ? 

**EDIT:** I'm not merging this, since cloud-init is failing, might be related

soyacz:
> @soyacz seems like the Azure provision test is breaking badly, can you take a look ?
> 
> **EDIT:** I'm not merging this, since cloud-init is failing, might be related

fails on 'could-init --version' command with segfault - doesn't look related.
But I took a look, and cloud-init logs show errors: `var/lib/sct/cloud-init/1_VectorDevUserDataObject.sh: line 6: backoff: command not found `
UserDataObject requires appending `configure_backoff_timeout` script

fruch:
> > > @soyacz seems like the Azure provision test is breaking badly, can you take a look ?
> > > **EDIT:** I'm not merging this, since cloud-init is failing, might be related
> > 
> > 
> > fails on 'could-init --version' command with segfault - doesn't look related. But I took a look, and cloud-init logs show errors: `var/lib/sct/cloud-init/1_VectorDevUserDataObject.sh: line 6: backoff: command not found ` UserDataObject requires appending `configure_backoff_timeout` script
> 
> so we have bug we didn't noticed, when not using provisioning stage...

but now it did pass... I'm a bit puzzled, maybe it being called only on error, and we hit the errors ?

fruch:
> > > > > @soyacz seems like the Azure provision test is breaking badly, can you take a look ?
> > > > > **EDIT:** I'm not merging this, since cloud-init is failing, might be related
> > > > 
> > > > 
> > > > fails on 'could-init --version' command with segfault - doesn't look related. But I took a look, and cloud-init logs show errors: `var/lib/sct/cloud-init/1_VectorDevUserDataObject.sh: line 6: backoff: command not found ` UserDataObject requires appending `configure_backoff_timeout` script
> > > 
> > > 
> > > so we have bug we didn't noticed, when not using provisioning stage...
> > 
> > 
> > but now it did pass... I'm a bit puzzled, maybe it being called only on error, and we hit the errors ?
> 
> probably issue with 'backoff' happens only on some error when it is actually called.

yes what I assumed as well, now it's fixed, I'm merging this one

PR #12062, opened 2025-09-26
fix(tester.py): improve checking kms enabling conditions

cezarmoise:
Improve the decision logic so that it is clear that kms configuration is done when none of the skip conditions are present, and that the config option `scylla_encryption_options` only overwrites the default values.

- added debug logging to explain decision
- added early return for each check
- added error handling: if parsing version fails, raise an error event

closes: #12030

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/c0305504-cdc0-4040-b2e6-9b9805a8bdc1
  with a ami previously built by BYO, it send an error event, but does not crash
  > 2025-09-26 12:56:23.500: (InfoEvent Severity.ERROR) period_type=not-set event_id=9cbed718-a368-4c49-8138-88f19e000080: message=Wanted to enable AWS KMS, but version check raised an error: Cannot parse provided 'null-null' scylla_version for the comparison. Transformed scylla_version: null-null
- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/b2206a10-335b-4047-93a9-f90df45d203a
  with an ami from an official build, but with `configurations/disable_kms.yaml` used
  >   < t:2025-09-26 12:53:31,735 f:tester.py       l:839  c:root                 p:DEBUG > Skip configuring AWS KMS, `enterprise_disable_kms` is set in the config
- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/16631a72-ba09-458f-9d3a-c158b50ef23e
  with an ami from an official build, kms gets enabled
  > < t:2025-09-26 13:45:57,092 f:tester.py       l:853  c:root                 p:DEBUG > `scylla_encryption_options` is not set in the config, using default values
< t:2025-09-26 13:45:57,225 f:aws_kms.py      l:128  c:sdcm.utils.aws_kms   p:INFO  > Creating 'alias/testid-16631a72-ba09-458f-9d3a-c158b50ef23e' alias for the '170e5241-37b2-4a6d-9ee8-049548cd517b' KMS key in the 'eu-west-1' region
  

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #12186, opened 2025-10-16
fix(kms): make KMS checks work correctly when relevant option is empty

vponomaryov:
Ref: https://github.com/scylladb/scylla-cluster-tests/pull/12062

### Testing

- [ ]

### Reminders


vponomaryov:
The approach is the same as it was before the merge of https://github.com/scylladb/scylla-cluster-tests/pull/12062/files

fruch:
> The approach is the same as it was before the merge of https://github.com/scylladb/scylla-cluster-tests/pull/12062/files

isn't reverting would be better ?

fruch:
> > The approach is the same as it was before the merge of https://github.com/scylladb/scylla-cluster-tests/pull/12062/files
> 
> isn't reverting would be better ?

now I see the original PR, it's more then just this change...

PR #12243, opened 2025-10-20
fix(pipelines): replace aws_region with region

cezarmoise:
`region` is the correct name for that parameter, not `aws_region`.
No errors happen currently because all params go into a map, and `aws_region` is just never used.

It has no real consequences, most tests are triggered and given a region.
But in some cases, i.e. native backup, the region must match the bucket defined in the config.

### Reminders


fruch:
@cezarmoise where ever the test that needs a specific region is trigger would need to keep that promise.


Why is holding us from duplicating the needed data into more regions ?

We started using multiple regions for performance tests, to avoid the problem of resources (and try to minimize noisy neighbors issues), and the ability to shift those tests into other regions might become important

Issue #12184, opened 2025-10-15
2025.1 - Offline installer most likely failing due to scylla-doctor


lsfreitas:
These are the two jobs failing:
https://jenkins.scylladb.com/job/scylla-2025.1/job/artifacts-offline-install/job/artifacts-debian12-nonroot-test/21/console
https://jenkins.scylladb.com/job/scylla-2025.1/job/artifacts-offline-install/job/artifacts-debian11-nonroot-test/21/console

Causing offline installer to fail: https://jenkins.scylladb.com/job/scylla-2025.1/job/offline-installer/20/console

Apparently it was recently introduced scylla-doctor check for nonroot runs and it can be related.

```
ERROR: test_scylla_service (artifacts_test.ArtifactsTest)
08:49:56  ----------------------------------------------------------------------
08:49:56  sdcm.sct_events.system.TestResultEvent: ================================= TEST RESULTS =================================
08:49:56  
08:49:56  ----- LAST ERROR EVENT -------------------------------------------------------
08:49:56  2025-10-10 07:49:45.126: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=4b96b526-fd8a-4aba-b5bd-fce8188a5a5f, source=ArtifactsTest.test_scylla_service (artifacts_test.ArtifactsTest) [check scylla_doctor results]() message=Traceback (most recent call last):
08:49:56    File "/tmp/workspace/scylla-2025.1/artifacts-offline-install/artifacts-debian12-nonroot-test/scylla-cluster-tests/artifacts_test.py", line 72, in logged_subtest
08:49:56      yield
08:49:56    File "/tmp/workspace/scylla-2025.1/artifacts-offline-install/artifacts-debian12-nonroot-test/scylla-cluster-tests/artifacts_test.py", line 402, in test_scylla_service
08:49:56      self.run_scylla_doctor()
08:49:56    File "/tmp/workspace/scylla-2025.1/artifacts-offline-install/artifacts-debian12-nonroot-test/scylla-cluster-tests/artifacts_test.py", line 517, in run_scylla_doctor
08:49:56      scylla_doctor.analyze_and_verify_results()
08:49:56    File "/tmp/workspace/scylla-2025.1/artifacts-offline-install/artifacts-debian12-nonroot-test/scylla-cluster-tests/utils/scylla_doctor.py", line 220, in analyze_and_verify_results
[...]

PR #12277, opened 2025-10-22
fix(artifact_test): fix nonroot debian scylla-doctor

soyacz:
In non-root debian11/12 test there's an issue with missing `mdstat` and `sysctl`.
Disabling these checks for this test.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/12184

### Testing

- [x] - [debian non root artifact test](https://argus.scylladb.com/tests/scylla-cluster-tests/c8a47210-741f-4aad-98dc-9fc32224cef8)

### Reminders

### DIFF
--- a/utils/scylla_doctor.py
+++ b/utils/scylla_doctor.py
@@ -198,6 +198,11 @@ class ScyllaDoctor:
                 collector in ["StorageConfigurationCollector", "PerftuneSystemConfigurationCollector"]):
             return True
 
+        if (self.node.distro.is_debian and self.offline_install and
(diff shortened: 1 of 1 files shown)


soyacz:
> @soyacz it was talked/raised with scylla-doctor ?

I wrote an email with the question.

fruch:
> > @soyacz it was talked/raised with scylla-doctor ?
> 
> I wrote an email with the question.

I think we should have bit more breadcrumbs for it, once you'll get an answer I would recommend opening issue

PR #12288, opened 2025-10-22
improvement(upgrade): allow to skip truncate checks in upgrade tests

vponomaryov:
Those checks may be done on cheap upgrade setups.
And those checks may timeout on huge expensive setups.
Also, it will take additional time on expensive setups.

So, add possibility to skip truncate checks for such cases.

### Testing

- [scylla-staging/valerii/vp-rolling-upgrade-custom-d2-w1-latency-regression-v2#7](https://argus.scylladb.com/tests/scylla-cluster-tests/f146079b-90c6-445f-963b-433bb17bc13c)

### Reminders

### DIFF
--- a/upgrade_test.py
+++ b/upgrade_test.py
@@ -72,28 +72,31 @@ NUMBER_OF_ROWS_FOR_TRUNCATE_TEST = 10
 def truncate_entries(func):
     @wraps(func)
     def inner(self, *args, **kwargs):
-        node = args[0]
-        with self.db_cluster.cql_connection_patient(node, keyspace='truncate_ks', connect_timeout=600) as session:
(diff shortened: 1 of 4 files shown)


upgrade_test.py
```diff
@@ -141,7 +144,7 @@ class UpgradeTest(FillDatabaseData, loader_utils.LoaderUtilsMixin):
 
     def setUp(self):
         super().setUp()
-
+        self.do_truncates = True
```
fruch: Let's do that at configuration level

It's something useful to be able to control on all upgrade tests

vponomaryov: Done

Issue #11581, opened 2025-08-01
`Waiting for SSH to be up: timeout - 500 seconds` error spawning multi-dc DB cluster


vponomaryov:
[Argus](https://argus.scylladb.com/test/a9a41918-48d4-4453-a4d7-2fede371c1ff/runs?additionalRuns[]=bfab2af8-f8e7-4fdd-a703-6b13eea87b82)

<details>
<summary>
Extra information
</summary>

## Installation details

Cluster size: 6 nodes (i4i.2xlarge)

Scylla Nodes used in this run:
- multi-dc-rolling-upgrade--ubuntu-fo-db-node-bfab2af8-5 (18.171.190.31 | 10.3.1.195) (shards: -1)
- multi-dc-rolling-upgrade--ubuntu-fo-db-node-bfab2af8-4 (18.171.178.129 | 10.3.3.154) (shards: -1)
- multi-dc-rolling-upgrade--ubuntu-fo-db-node-bfab2af8-3 (34.247.183.115 | 10.4.3.151) (shards: -1)
- multi-dc-rolling-upgrade--ubuntu-fo-db-node-bfab2af8-2 (63.33.202.175 | 10.4.2.190) (shards: -1)
- multi-dc-rolling-upgrade--ubuntu-fo-db-node-bfab2af8-1 (52.214.179.114 | 10.4.0.54) (shards: -1)

OS / Image: `ami-01b649b48804460e8 ami-0047ffd1f9bea49b6` (aws: undefined_region)

Test: `rolling-upgrade-multidc-test`
Test id: `bfab2af8-f8e7-4fdd-a703-6b13eea87b82`
Test name: `scylla-2025.3/rolling-upgrade/rolling-upgrade-multidc-test`
Test method: `upgrade_test.UpgradeTest.test_generic_cluster_upgrade`
Test config file(s):

- [multi-dc-rolling-upgrade.yaml](https://github.com/scylladb/scylla-cluster-tests/blob/650ff82ba883155a2328b9d4030f491c85277014/test-cases/upgrades/multi-dc-rolling-upgrade.yaml)

## Logs:
- **[db-cluster-bfab2af8.tar.zst](https://argus.scylladb.com/api/v1/tests/scylla-cluster-tests/bfab2af8-f8e7-4fdd-a703-6b13eea87b82/log/db-cluster-bfab2af8.tar.zst/download)**
[...]

PR #12328, opened 2025-10-27
fix(wait_ssh_up): use adaptive_timeout and triple the hard timeout

fruch:
so we can track the issue we are seeing with ssh connectivity to nodes we are extending the timeout we wait for them x3, in case is some slowness during cloud-init.

also we are using `adaptive_timeout` so we track the cases it's taking more then original timeout we had.

Ref: #11581

### Testing

- [x] aws provision test
- [x] short longevity - 🟢 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/longevity-100gb-4h-test/160/

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -1314,7 +1314,11 @@ class BaseNode(AutoSshContainerMixin):
         text = None
         if verbose:
             text = '%s: Waiting for SSH to be up' % self.name
-        wait.wait_for(func=self.remoter.is_up, step=10, text=text, timeout=timeout, throw_exc=True)
+
+        ssh_timeout_multiplier = 3
+        with adaptive_timeout(Operations.SSH_CONNECTIVITY, node=self, timeout=timeout, node_available=False):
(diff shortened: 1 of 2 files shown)


sdcm/cluster.py
```diff
@@ -1314,7 +1314,11 @@ def wait_ssh_up(self, verbose=True, timeout=500):
         text = None
         if verbose:
             text = '%s: Waiting for SSH to be up' % self.name
-        wait.wait_for(func=self.remoter.is_up, step=10, text=text, timeout=timeout, throw_exc=True)
+
+        ssh_timeout_multiplier = 3
+        with adaptive_timeout(Operations.SSH_CONNECTIVITY, node=self, timeout=timeout, node_available=False):
+            wait.wait_for(func=self.remoter.is_up, step=10 * ssh_timeout_multiplier,
```
soyacz: I wouldn't change the step - it will prolong test by 30 s if we have bad luck

PR #12334, opened 2025-10-27
feature(xcloud): add 2 scylla cloud tests for xcloud cluster type

MykytaChepurko:
### Testing

- [ ]

### Reminders

### DIFF
--- a/xcloud_test.py
+++ b/xcloud_test.py
@@ -21,7 +21,7 @@ from longevity_test import LongevityTest
 class XcloudScaleOutTest(LongevityTest):
 
     # The expected max time for scale out successful completion is 40 minutes
-    @retrying(n=8, sleep_time=60*5, allowed_exceptions=(Retry,))
+    @retrying(n=8, sleep_time=60*5, allowed_exceptions=(Retry, AssertionError))
     def verify_cluster_scaled_out(self, nodes):
         self.log.info(f"Verify the number of nodes after scale out is 6 and all of them are 'UP'")
         assert len(nodes) == 6, "Initial number of nodes is not 3"


sdcm/sct_config.py
```diff
@@ -3153,19 +3182,21 @@ def _validate_cloud_backend_parameters(self):
...
-                f"Supported instance types: {', '.join(supported_instances)}")
-
-        rf = self.get('xcloud_replication_factor')
-        n_nodes = self.get('n_db_nodes')
-        if rf is None:
-            self['xcloud_replication_factor'] = n_nodes
-        elif rf > n_nodes:
-            raise ValueError(f"xcloud_replication_factor ({rf}) cannot be greater than n_db_nodes ({n_nodes})")
+        if not self.get("scaling"):
+            # setting db_instance_type, rf and n_nodes only for non-xcloud clusters
```
fruch: I don't see a need to automatically set db_instance_type... SCT shouldn't be having this logic...

docs/configuration_options.md
```diff
@@ -3775,3 +3775,11 @@ Vector Store indexing threads (if not set, defaults to number of CPU cores on VS
 **default:** N/A
 
 **type:** int
+
+## **scaling** / SCT_CLUSTER_SCALING
```
fruch: this is auto generated docs, those kind of things should be added to `sct_config.py`, and that it should be generated based on that

xcloud_test.py
```diff
@@ -0,0 +1,48 @@
...
+    def test_xcloud_scale_out(self):
+        """
+        Test Xcloud cluster automatic scale out triggering upon reaching specified threshold
+        """
+        self.log.info("Starting Xcloud scale out test")
+
+        initial_nodes = self.db_cluster.get_nodetool_status()
+
+        self.log.info(f"Verify initial number of nodes -3 and all of them are 'UP'")
+        assert len(initial_nodes) == 3, "Initial number of nodes is not 3"
```
fruch: not sure what's the point of this validation, you won't start the test if the nodes are not up, and if you don't have exactly what defined in the test configuration

xcloud_test.py
```diff
@@ -0,0 +1,48 @@
...
+from sdcm.utils.decorators import retrying, Retry
+from sdcm.mgmt.operations import ManagerTestFunctionsMixIn
+
+
+class XcloudScaleOutTest(ManagerTestFunctionsMixIn):
+
+    # The expected max time for scale out successful completion is 40 minutes
+    @retrying(n=8, sleep_time=5, allowed_exceptions=(Retry,))
+    def verify_cluster_scaled_out(self, nodes):
+        self.log.info(f"Verify the number of nodes after scale out is 6 and all of them are 'UP'")
```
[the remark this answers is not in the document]
fruch: doing that, one can use longevity test, and configure it to do as such, and not need for a new test class

fruch:
couple of general notes:
* I would recommend sticking to existing test as much as possible, so one can later build on the whole possible features of SCT, i.e. nemesis and such.
* we should think about how/when SCT gonna sync with the new topology - other wise during more complex case we are gonna point to non existing nodes, and we won't be able to do much within the test.

sdcm/sct_config.py
```diff
@@ -1857,19 +1857,48 @@ class SCTConfiguration(dict):
...
+         """),
+
     ]
 
     required_params = ['cluster_backend', 'test_duration', 'n_db_nodes', 'n_loaders', 'use_preinstalled_scylla',
                        'user_credentials_path', 'root_disk_size_db', "root_disk_size_monitor", 'root_disk_size_loader']
 
     # those can be added to a json scheme to validate / or write the validation code for it to be a bit clearer output
     backend_required_params = {
-        'aws': ['user_prefix', "instance_type_loader", "instance_type_monitor", "instance_type_db",
```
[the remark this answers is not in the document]
rayakurl: @fruch even if this will be the default, the backend official name is scyllaDB cloud, not xcloud. Xcloud is only a type of cluster. This is important as we possibly will have other tests to run on free trial, snadbox and the standartd, which is still not the default.  
Can you please elaborate more about why you named it xcloud on purpose?

MykytaChepurko:
> @MykytaChepurko - what's the status here? Is this being worked on?

@mykaul I’m not using this PR branch for running xCloud tests right now because I need to make frequent small config changes/adding new tests. Waiting for each change to be merged would slow things down significantly, so I’m working from a separate branch for faster iteration.

Current working branch:
https://github.com/MykytaChepurko/scylla-cluster-tests/tree/feature/xcloud-tests-support-debug

Once the main testing phase is done and the configs stabilize, I’ll move all the changes back into the PR branch.

PR #12487, opened 2025-11-09
fix(get_specific_tag_of_docker_image): fetch image by architecture

fruch:
seems like recently the `00-Build.txt` format slightly changed and we need to adapt the code, to append the architecture to the get to get the correct image

for now we more or less hardcode for x86, and that kind of surface we aren't testing the arm docker variant in any artifact test

### Testing

- [x] integration tests

### Reminders

### DIFF
--- a/sdcm/utils/version_utils.py
+++ b/sdcm/utils/version_utils.py
@@ -584,7 +584,16 @@ def resolve_latest_repo_symlink(url: str) -> str:
     return resolved_url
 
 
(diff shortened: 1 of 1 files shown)


fruch:
> for now we more or less hardcode for x86, and that kind of surface we aren't testing the arm docker variant in any artifact test

FYI @yaronkaikov @roydahan

fruch:
@yaronkaikov @Annamikhlin this in key name, was backported to any branch ?

PR #12530, opened 2025-11-12
fix(docker): fix all deprecated references to IPAddress

fruch:
seems like the API for docker instance networking was changed and some parts were deprecated while we didn't noticed and now we did,

and this change replace all the places to look within bridged network

Ref: https://docs.docker.com/engine/deprecated/#top-level-network-properties-in-networksettings

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/docker_utils.py
+++ b/sdcm/utils/docker_utils.py
@@ -312,7 +312,7 @@ class ContainerManager:
             ip_address = cls.get_container(
                 instance, name).attrs["NetworkSettings"]["Networks"][docker_network]["IPAddress"]
         else:
-            ip_address = cls.get_container(instance, name).attrs["NetworkSettings"]["IPAddress"]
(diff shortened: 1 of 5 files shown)


sdcm/utils/docker_utils.py
```diff
@@ -312,7 +312,7 @@ def get_ip_address(cls, instance: INodeWithContainerManager, name: str) -> str:
             ip_address = cls.get_container(
                 instance, name).attrs["NetworkSettings"]["Networks"][docker_network]["IPAddress"]
         else:
-            ip_address = cls.get_container(instance, name).attrs["NetworkSettings"]["IPAddress"]
+            ip_address = cls.get_container(instance, name).attrs["NetworkSettings"]['Networks']['bridge']['IPAddress']
```
dimakr: Dockek provisioning test failed. Seems like that not all containers managed by SCT will be on `bridge` network.

fruch:
we did update SCT runners last week, so you might still see those issues on Jenkins as well.

We need to fix both ends (both repos)

PR #12550, opened 2025-11-14
improvement(perf): add support for gradual thrpt grow using latte

vponomaryov:
We have a test called `PerformanceRegressionPredefinedStepsTest`
which allows to test throughput performance using gradually growing stress values.

And it uses several substitution vars among which there is '$throttle' one.

And, as of now, it is hardcoded to support cassandra-stress only.
So, as a first step for latte support in this test,
add support for the `$throttle` substitution var for latte.

### Testing

- All test refs in here https://scylladb.atlassian.net/wiki/spaces/RND/pages/175538178/Latte+Custom+schema+Throughput+measurement

### Reminders

### DIFF
--- a/performance_regression_gradual_grow_throughput.py
+++ b/performance_regression_gradual_grow_throughput.py
@@ -39,6 +39,10 @@ class Workload:
             self.num_threads = [self.num_threads]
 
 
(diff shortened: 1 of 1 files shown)


fruch (approved):
LGTM

PR #12551, opened 2025-11-14
fix(get_datacenter_name_per_region): use first alive nodes to get datacenter name

aleksbykov:
method get_datacenter_name_per_region use the nodes[0] to get datacenter name.
But it could be down by nemesis as in job example:
https://argus.scylladb.com/tests/scylla-cluster-tests/3214440e-ead0-4eea-b880-a63938368b8e
In the job nemesis RestartElectedTopologyCoordinator stops nodes[0]:
first node in list(because it was topology coordinator) and then try to
run `nodetool status` on node that is down.
```
 if status := nodes[0].get_nodes_status():
```
this cause to nemesis thread to stuck.

Fix get_datacenter_name_per_region to get first alive node
in sorted by dc nodes to get real name of datacente
### Testing
- [Failed job before fix](https://argus.scylladb.com/tests/scylla-cluster-tests/3214440e-ead0-4eea-b880-a63938368b8e) 
- [Passed job after fix](https://argus.scylladb.com/tests/scylla-cluster-tests/aef508b5-2cc9-43dc-8d84-4161cb677765)

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -3456,12 +3456,16 @@ class BaseCluster:
     def get_datacenter_name_per_region(self, db_nodes=None):
         datacenter_name_per_region = {}
         for region, nodes in self.nodes_by_region(nodes=db_nodes).items():
-            if status := nodes[0].get_nodes_status():
+            node = next((node for node in nodes if node.db_up()), None)
+            if node is None:
+                LOGGER.error("No DB up node found in region %s to get datacenter name", region)
+                continue
+            if status := node.get_nodes_status():
                 # If `nodetool status` failed to get status for the node
-                if dc_name := status.get(nodes[0], {}).get('dc'):
+                if dc_name := status.get(node, {}).get('dc'):
                     datacenter_name_per_region[region] = dc_name
             else:
-                LOGGER.error("Failed to get nodes status from node %s", nodes[0].name)
+                LOGGER.error("Failed to get nodes status from node %s", node.name)
 
         return datacenter_name_per_region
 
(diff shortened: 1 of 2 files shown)


fruch:
@juliayakovlev since this is touching the rackawre logic, please take a look at this one

sdcm/cluster.py
```diff
@@ -3748,7 +3748,8 @@ def get_load_balancing_policy(self, whitelist_nodes: Optional[list[BaseNode]] =
             loader = self.nodes[0].test_config.tester_obj().loaders.nodes[0]
             loader_rack = loader.node_rack
             loader_dc = loader.datacenter
-            datacenter_name_per_region = self.get_datacenter_name_per_region(db_nodes=self.nodes)
+            nodes = whitelist_nodes or self.nodes  # need to get datacenter names from alive nodes if whitelist_nodes passed
```
juliayakovlev: Does your test runs with `self.params.get("rack_aware_loader") and self.racks_count > 1`?
It appears that this code has no impact on your test. 
Since the rackaware test isn’t expected to trigger this type of nemesis, I believe this change is unnecessary.

juliayakovlev: So this should be fixed - set `rack_aware_loader` to false in not relevant for rackaware test

aleksbykov: Without fix any test could be failed when appropriate conditions happened. It should don't affect current tests, because whitelist should be passed explicitly, and this done for now only in 2 nemesis.

aleksbykov: @fruch , fixed `get_datacenter_name_per_region` according to your comment.

PR #12556, opened 2025-11-14
feat: Add Python driver to the cache-issues workflow

pehala:
We want to reference python driver issues in dtest (ref: https://github.com/scylladb/scylla-dtest/pull/6416, https://github.com/scylladb/scylladb/issues/16709)
### Testing

- [x] locally


fruch (approved):
LGTM

PR #12604, opened 2025-11-18
feature(test_read_disk_only_gradual_increase_load): Add new performance test

juliayakovlev:
Add new subtest of latency and throughput to test a workload that reads only from the disk. The workload has 100% miss rate (or 0% hit rate from the cache).

Based on https://github.com/roydahan/scylla-cluster-tests/commit/dd7c0fdaa10ee344f422d5e12de44775b76381ce

### Testing

- [x] [read_disk_only test](https://argus.scylladb.com/tests/scylla-cluster-tests/2a4e611e-30a6-472a-80f5-ef38ae79fd05)
- [x] [run before "read_disk_only" step name change](https://argus.scylladb.com/tests/scylla-cluster-tests/490f520e-a43f-43c7-a37f-5070bad04f5d)

### TODOs
- [x] make sure `read_disk_only` is reported on it's own name to argus, (and not as read)
- [x] add `read_disk_only` to the triggers, should be run on the weekly/3wk vnodes/tablets, and in releases.

### Reminders

### DIFF
--- a/vars/perfRegressionParallelPipelinebyRegion.groovy
+++ b/vars/perfRegressionParallelPipelinebyRegion.groovy
@@ -53,7 +53,7 @@ def call(Map pipelineParams) {
                                 region: 'us-east-1',
                                 versions: ['2024.1', '2024.2', '2025.1', '2025.2', '2025.3', '2025.4', 'master'],
                                 pre_release: [], // Example: ['rc1', 'rc3']
-                                sub_tests: ['"test_read_gradual_increase_load"', '"test_mixed_gradual_increase_load"'],
+                                sub_tests: ['"test_read_gradual_increase_load"', '"test_mixed_gradual_increase_load"', '"test_read_disk_only_gradual_increase_load"'],
                                 labels: ['master-weekly']
                             ],
                             [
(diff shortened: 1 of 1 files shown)


fruch:
> > @juliayakovlev lets add `read_disk_only` also to the perf trigger (and also to the list we did what runs where)
> 
> @fruch I am thinking - may be it should not be separate job. We can add it to existing throughput job. WDYT?

yes we should keep it in the same job, but for that we need to make sure it's not mixing with the other results
regardless we need to define the workload, since we selectively did different for master and different for releases

configurations/performance/cassandra_stress_gradual_load_reduced_steps_number.yaml
```diff
@@ -4,5 +4,5 @@
 ##   - a single-element list or integer (applied to all throttle steps)
 ##   - a list with the same length as perf_gradual_throttle_steps[load] (one thread count per step).
 perf_gradual_threads: {"read": 620, "write": 400, "mixed": 1900}
-perf_gradual_throttle_steps: {"read": ['450000', '700000', 'unthrottled'], "mixed": ['300000', '450000', 'unthrottled'], "write": ['200000', 'unthrottled']}  # where every value is in ops
+perf_gradual_throttle_steps: {"read": ['450000', '700000', 'unthrottled'], "mixed": ['300000', '450000', 'unthrottled'], "write": ['200000', 'unthrottled'], "read_disk_only": ['250000', '300000', 'unthrottled']}  # where every value is in ops
 perf_gradual_step_duration: {"read": '20m', "write": None, "mixed": '20m'}
```
roydahan: don't you need here also read_disk_only ?

roydahan (approved):
Other than my comment on he reduced steps number, it looks good to me.
Probably because I "coded" it :)

configurations/performance/cassandra_stress_gradual_load_reduced_steps_number.yaml
juliayakovlev: fixed

PR #12666, opened 2025-11-21
Update major k8s dependencies used in EKS, GKE tests

abramche:
Current component versions are terribly out of touch with reality. Cert-manager 1.8.0 is unsupported for two years and does not support even the lowest actual Operator k8s version (1.31). This PR addresses the dependency update.

### Testing
All of the included changes were tested before and during the Operator 1.19 release:
https://jenkins.scylladb.com/view/scylla-operator/job/scylla-operator/job/operator-1.19/job/upgrade/

### DIFF
--- a/defaults/k8s_eks_config.yaml
+++ b/defaults/k8s_eks_config.yaml
@@ -9,15 +9,15 @@ availability_zone: 'a,b'
 instance_provision: 'on_demand'
 instance_type_db: 'i4i.4xlarge'
 
-eks_cluster_version: '1.27'
+eks_cluster_version: '1.32'
 eks_role_arn: 'arn:aws:iam::797456418907:role/eksServicePolicy'
 eks_service_ipv4_cidr: '172.20.0.0/16'
 eks_nodegroup_role_arn: 'arn:aws:iam::797456418907:role/helm-test-worker-nodes-NodeInstanceRole-6ACHDYEKNN3I'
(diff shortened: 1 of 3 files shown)


abramche:
> Are we should it's not depend on any of those other changes in those branches ?

> right now any change that might be breaking k8s backends, isn't getting tested on master, since it doesn't work. I would love to see all changes/fixes going back into SCT master.

Both valid points - I merge master into these branches on a weekly basis. It should be good (for now).

I agree though everything would eventually need to be merged to master as it's also harder for me to work in a fork. This is a legacy of https://github.com/scylladb/scylla-cluster-tests/pull/10890

fruch:
> > Are we should it's not depend on any of those other changes in those branches ?
> 
> > right now any change that might be breaking k8s backends, isn't getting tested on master, since it doesn't work. I would love to see all changes/fixes going back into SCT master.
> 
> Both valid points - I merge master into these branches on a weekly basis. It should be good (for now).
> 
> I agree though everything would eventually need to be merged to master as it's also harder for me to work in a fork. This is a legacy of #10890

I would recommend pulling them in in smaller bites, so they can be reviewed and merged in.
we had similar problem with the integration with siren-tests, hopfully now we are heading in a bit of different direction with the scylla-cloud integration with SCT.

with k8s, we had it working out of master for quite some time, and we still have people that was working with it, and can review or pitch in as needed

PR #12722, opened 2025-11-26
ci(longevity): migrating large partition GCE run to new instance

timtimb0t:
Changing instance for large partiton GCE run from n2-highmem instance
 type to the new z3-highmem-8-highlssd instance type to standardize
 infrastructure across all tests.

### Testing

- [ ]

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
+++ b/test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml
@@ -45,7 +45,7 @@ n_loaders: 5 # Each loader will have 1 scylla-bench process at every step of the
 round_robin: true
 
 instance_type_db: 'i3en.2xlarge'
-gce_instance_type_db: 'n2-highmem-16'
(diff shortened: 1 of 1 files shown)


timtimb0t:
Hi @fruch , I saw your PR where you changed pipelines to migrate GCE runs but it seems for its enough to change one value for this test, please correct me if I'm wrong 

CC @pehala

timtimb0t:
> > @fruch offered to migrate GCE run to new instance types for consistent performance, if I'm correct
> 
> if the problem is due to performance, it is fine, but it is not what the cover letter says. It would be also good to reference issue where we saw such performance problems.

Here is the initial commit what I was referred to: https://github.com/scylladb/scylla-cluster-tests/pull/12571

fruch:
My change was mostly cause of instances used were over provisioned for the tests.
it was mostly for cost saving

the is some reasons to start utilizing z3, but moving from 16 cpu to 8 cpu, doesn't sound like something I would do without clear indication this case can manage 50% drop in CPUs

Issue #12366, opened 2025-10-28
Log collection failed with error "rsync: connection unexpectedly closed (0 bytes received so far) [sender]"


timtimb0t:
Runs:
https://argus.scylladb.com/tests/scylla-cluster-tests/8e7470be-1550-4634-967b-5defb51e16cb
https://argus.scylladb.com/tests/scylla-cluster-tests/0f79b774-5640-4e30-ae6c-c6258604d6ab

Logs collection failed with the same symptoms.
```
12:05:12  Syncing /tmp/jenkins/workspace/scylla-2025.4/tier1/longevity-twcs-48h-test/scylla-cluster-tests to the SCT runner instance...
12:05:12  # Host 54.243.24.255 found: line 4
12:05:12  # Host 54.243.24.255 found: line 5
12:05:12  # Host 54.243.24.255 found: line 6
12:05:12  /home/jenkins/.ssh/known_hosts updated.
12:05:12  Original contents retained as /home/jenkins/.ssh/known_hosts.old
12:07:29  ssh: connect to host 54.243.24.255 port 22: Connection timed out
12:07:29  rsync: connection unexpectedly closed (0 bytes received so far) [sender]
12:07:29  rsync error: unexplained error (code 255) at io.c(231) [sender=3.2.7]
12:07:29  Cleaning SSH agent
12:07:29  Agent pid 16173 killed
```

 Also the test runs itself finished unexpectedly with the same errors:
```
11:28:00  < t:2025-10-27 10:27:48,249 f:node_allocator.py l:188  c:sdcm.utils.nemesis_utils.node_allocator p:INFO  > KMS encryption check-ceaa8996: unset running nemesis for node longevity-tls-50gb-3d-2025-4-db-node-8e7470be-3.
12:05:05  time="2025-10-27T11:04:54Z" level=error msg="error waiting for container: command [ssh -l ubuntu -o ConnectTimeout=30 -T -- 98.90.175.245 docker system dial-stdio] has exited with exit status 255, make sure the URL is valid, and Docker 18.09 or later is installed on the remote host: stderr=Connection to 98.90.175.245 closed by remote host.\r\n"
12:05:05  Cleaning SSH agent
12:05:05  Agent pid 3258 killed
``` 

Migrated to Jira: [SCT-36](https://scylladb.atlassian.net/browse/SCT-36)

[...]

PR #12724, opened 2025-11-26
improvement(vector_logging): filter unnecessary logs on client side

dimakr:
Configure vector logging to filter out unnecessary logs on the client side. This includes:
- compaction/repair/streaming logs from scylla, and logs from certain system services - the logs that were previously filter out on SCT runner in DbLogReader thread
- error logs that are ignored through default filters, where we change severity level from error to warning

Filtering logs on the client side helps reduce memory pressure on both, the SCT runner instance and DB nodes. Vector will no longer read certain messages from journald (and buffer them in memory) on the client side; and SCT runner will not be busy processing them at all. Additionally, this reduces network bandwidth usage as logs_to_be_dropped will no longer be sent from DB nodes to SCT.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/12366

### Testing

- [x] executed `longevity-5gb-1h-nemesis.yaml` test, on SCT master and on this change branch:
monitoring shows that on SCT master the `interface RX packets` value on runner instance is 2.1-2.2 kbps on average during most the test (https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/61/)
test-net-without-change

on the branch with this fix the `interface RX packets` value on runner instance is only 50 bps on average during most the test (https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/60/)
test-net-with-change

So before the change we are getting like 40-45 times more log related traffic, from vector on DB nodes to SCT runner. And then this amount of logs is also being processed on the runner side.

### Reminders

### DIFF
--- a/sdcm/provision/common/utils.py
+++ b/sdcm/provision/common/utils.py
@@ -49,6 +49,20 @@ def configure_syslogng_target_script(hostname: str = "") -> str:
 
 
 def configure_vector_target_script(host: str, port: int) -> str:
+    """Prepare vector configuration script with client-side log filtering.
+
+    Configures vector to filter verbose logs before sending them to SCT, reducing memory pressure
+    on database nodes and network resources usage.
+
+    Filter Pipeline:
+        journald > filter_audit > filter_system_services > filter_verbose_scylla > filter_suppress_warnings > sct-runner
+
+    Filters:
+        - filter_audit: remove audit logs
+        - filter_system_services: remove unnecessary system services logs
+        - filter_verbose_scylla: remove compaction/repair/streaming scylla logs
+        - filter_suppress_warnings: remove Severity.SUPPRESS events
+    """
     return dedent("""
         echo "
         sources:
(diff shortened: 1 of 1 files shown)


dimakr:
@fruch we discussed that we can apply this approach of filtering logs on the client side, to drop those logs that we use now in EventsSeverityChanger filters.
I added these filters to vector config, only those from `sdcm/sct_events/setup.py` - i.e. the default ones that we configure for the whole run.
But I wonder if it is OK to drop these logs altogether, i.e. that we won't have warning events for them, like we used to.

sdcm/provision/common/utils.py
```diff
@@ -64,11 +80,66 @@ def configure_vector_target_script(host: str, port: int) -> str:
...
+                    message = to_string(.message) ?? \\"\\"
+                    !(match(message, r'^WARNING.*\\[shard.*\\]') || match(message, r'^!.*WARNING.*\\[shard.*\\]'))
+
+            filter_default_severity_changer_candidates:
+                inputs:
+                    - filter_suppress_warnings
+                type: filter
+                condition: |
+                    message = to_string(.message) ?? \\"\\"
+                    !contains(message, \\"workload prioritization - update_service_levels_from_distributed_data\\") &&
```
fruch: I'm missing comments with references to the issue for each one

fruch:
> @fruch we discussed that we can apply this approach of filtering logs on the client side, to drop those logs that we use now in EventsSeverityChanger filters. I added these filters to vector config, only those from `sdcm/sct_events/setup.py` - i.e. the default ones that we configure for the whole run. But I wonder if it is OK to drop these logs altogether, i.e. that we won't have warning events for them, like we used to.

`filter_default_severity_changer_candidates` are not happening that often, so maybe we can take those out.

as for warnings, we collect them, but almost never do anything with them, we introduced filtering just for the sake of not raising any error event from them, not for keeping them.

dimakr:
> `filter_default_severity_changer_candidates` are not happening that often, so maybe we can take those out.

was thinking the same. I'll remove that filter and re-run the test, to see if anything changes in terms of traffic to SCT runner

fruch (approved):
LGTM

Issue #12734, opened 2025-11-27
latte_thread.py: build_stress_cmd() is failing for scripts if there is a sibling directory


enaydanov:
https://github.com/scylladb/scylla-cluster-tests/commit/050488dd5439473947a5e632c6c0db4c59535349 brought the `vector_search/` subdirectory to `data_dir/latte/` and, in result, tests which use scripts from `data_dir/latte/` itself start to fail with errors like:

```
Command: 'rsync -L   --timeout=300 --rsh=\'/usr/bin/ssh  -a -x  -o StrictHostKeyChecking=no -o UserKnownHostsFile=/tmp/tmpxxyd0fvs -o BatchMode=yes -o ConnectTimeout=300 -o ServerAliveInterval=300 -l ubuntu -p 22  -i /home/ubuntu/.ssh/scylla_test_id_ed25519\' -az /home/ubuntu/scylla-cluster-tests/data_dir/latte/vector_search ubuntu@[10.4.2.72]:"/tmp/tmp.IZtBy0OIIb"'

Exit code: 3

Stdout:



Stderr:

ERROR: destination must be a directory when copying more than 1 file
rsync error: errors selecting input/output files, dirs (code 3) at main.c(760) [Receiver=3.2.7]
```

PR #12737, opened 2025-11-27
fix(latte_thread): do not copy non-file items in build_stress_cmd

mikliapko:
Fixes https://github.com/scylladb/scylla-cluster-tests/issues/12734

Added a check to make sure that the `src_file` instance returned by `iterdir()` is file indeed. If it's a directory, the copy operation that happens next fails.

### Testing

- [x] [vector-search-in-cloud-aws](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/vector-search-in-cloud-aws/32/)

### DIFF
--- a/sdcm/stress/latte_thread.py
+++ b/sdcm/stress/latte_thread.py
@@ -107,7 +107,8 @@ class LatteStressThread(DockerBasedStressThread):
             self.SCHEMA_CMD_CALL_COUNTER[script_name] = 0
 
         for src_file in (Path(get_sct_root_path()) / script_name).parent.iterdir():
-            cmd_runner.send_files(str(src_file), str(Path(script_name).parent / src_file.name))
+            if src_file.is_file():  # exclude non-file items from being copied
+                cmd_runner.send_files(str(src_file), str(Path(script_name).parent / src_file.name))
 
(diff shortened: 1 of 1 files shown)


sdcm/stress/latte_thread.py
```diff
@@ -107,7 +107,8 @@ def build_stress_cmd(self, cmd_runner, loader, hosts):
             self.SCHEMA_CMD_CALL_COUNTER[script_name] = 0
 
         for src_file in (Path(get_sct_root_path()) / script_name).parent.iterdir():
-            cmd_runner.send_files(str(src_file), str(Path(script_name).parent / src_file.name))
+            if src_file.is_file():  # exclude non-file items from being copied
```
vponomaryov: Not critical, but would be good to avoid duplication of copying.
We have test configs with dozens of latte commands for just 2 loaders...

mikliapko: I will check today-tomorrow, somehow lost track of this PR...

mikliapko: sorry, yep, we have very important Vector Search launch now. If you could take it over, I'd be very grateful...

PR #12747, opened 2025-11-30
Revert "improvement(vector_logging): filter unnecessary logs on client side"

fruch:
Seems like this is breaking all of the artifacts tests, so we reverting it first 
Reverts scylladb/scylla-cluster-tests#12724


fruch:
we'll need to reapply the fix, with this solution

PR #12779, opened 2025-12-02
fix(tier1): 1TB-5days-authorization-and-tls-ssl test

temichus:
Increase instance types to prevent cluster overloading

testing: https://argus.scylladb.com/tests/scylla-cluster-tests/fe40920f-1e68-4271-b511-b65222049eb8
image

fixes: https://github.com/scylladb/scylladb/issues/26270


temichus:
> @temichus can you put link to the runs you did ?

added

PR #12842, opened 2025-12-05
[Branch-2025.4] improvement(treewide): Use ruff format instead of autopep8

pehala:
Same as https://github.com/scylladb/scylla-cluster-tests/pull/11291 (The format commit is not a cherrypick)
### Testing

- [x] None


pehala:
Hydra on macos is broken on older versions.

PR #12845, opened 2025-12-05
[Branch-2025.2] improvement(treewide): Use ruff format instead of autopep8

pehala:
Same as https://github.com/scylladb/scylla-cluster-tests/pull/11291 (The format commit is not a cherrypick)

There is one instance, where I had to do `# fmt: skip`, due to single quotes beeing neded on Python 3.10

### Testing

- [x] None


pehala:
`10:54:45  error pulling image configuration: download failed after attempts=6: received unexpected HTTP status: 500 Internal Server Error`
Will retry later

PR #12863, opened 2025-12-08
improvement(build-docker-image): Run regardless of paths

pehala:
Filtering for paths is redundant since we filter by the label. If someone wants to regenerate hydra, he should have the ability to do so. It is a reason why the hydra was not ran in https://github.com/scylladb/scylla-cluster-tests/pull/12845

Adding backports to all used releases, some of those will have conflicts as this file might not exists there. It should be anywhere as it is critical piece of automation IMO and I will deal with confclits on those backports.

### Testing

- [x] None needed

### DIFF
--- a/.github/workflows/build-docker-image.yaml
+++ b/.github/workflows/build-docker-image.yaml
@@ -3,11 +3,6 @@ name: Build hydra image
 on:
   pull_request_target:
     types: [labeled]
-    paths:
-      - 'Dockerfile'
-      - 'docker/env/build_n_push.sh'
-      - 'uv.lock'
-      - 'pyproject.toml'
 
 permissions:
   actions: write


pehala:
> I don't think we should be removing this, if someone want to force it, this machinery isn't for him.

I see it as too restrictive, and could help cases like https://github.com/scylladb/scylla-cluster-tests/pull/12845, where you need to do it to fix issues that was not cause by that PR. In those "I need to fix" cases I would rather "misuse" this automation than for someone to do it manually.

If you dont like this and you are fine with manually fixing these rare occassions, I am fine with closing this. It is just a proposal

fruch:
> > I don't think we should be removing this, if someone want to force it, this machinery isn't for him.
> 
> I see it as too restrictive, and could help cases like https://github.com/scylladb/scylla-cluster-tests/pull/12845, where you need to do it to fix issues that was not cause by that PR. In those "I need to fix" cases I would rather "misuse" this automation than for someone to do it manually.
> 
> If you dont like this and you are fine with manually fixing these rare occassions, I am fine with closing this. It is just a proposal

It's the maintainer's responsibility to make sure those are build correctly, labels are main for the rest of contributers out there. (Including renovate), I rather have it restrictive.

PR #12874, opened 2025-12-08
improvement(treewide): Check for jump-statement-in-finally

pehala:
* In python 3.14, this is not Syntax warning (SyntaxWarning: 'return' in a 'finally' block)

### Testing

- [ ] Integration test
   - I do think it affects the functionality, the code should be equal

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -85,7 +85,7 @@ lint.select = [
     "YTT",
     "F541",
     "PIE",
-    "B006",
+    "B006", "B012",
     "PLW", # mainly for PLW0211
 ]
 lint.ignore = ["E501", "PLR2004"]
--- a/sdcm/scan_operation_thread.py
+++ b/sdcm/scan_operation_thread.py
@@ -176,7 +176,7 @@ class FullscanOperationBase:
                     # success is True if there were no exceptions
                     self.current_operation_stat.success = not bool(self.current_operation_stat.exceptions)
                     self.update_stats(self.current_operation_stat)
-                    return self.current_operation_stat
+                return self.current_operation_stat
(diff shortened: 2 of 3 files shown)


enaydanov (approved):

pehala:
> I don't think we have any tests (unit/integration) covering those modules

Not sure, I am fairly confident in those changes, but I can run e.g. one SLA test to fully verify if needed. I will leave it or @enaydanov and @temichus to decide what should be run with regard to that

> @pehala those autofixers ?


Nope, I had to move the return manually.

fruch:
> > I don't think we have any tests (unit/integration) covering those modules
> 
> Not sure, I am fairly confident in those changes, but I can run e.g. one SLA test to fully verify if needed. I will leave it or @enaydanov and @temichus to decide what should be run with regard to that

> 
> > @pehala those autofixers ?
> 
> Nope, I had to move the return manually.



> > I don't think we have any tests (unit/integration) covering those modules
> 
> Not sure, I am fairly confident in those changes, but I can run e.g. one SLA test to fully verify if needed. I will leave it or @enaydanov and @temichus to decide what should be run with regard to that

@enaydanov you gave a thumb up... but we'll wait for one of you confirm before merging

> 
> > @pehala those autofixers ?
> 
> Nope, I had to move the return manually.

enaydanov:
@pehala The problem here that `return` in `finally` block implicitly suppresses any exception and moving it out of `try` can cause failures because of unhandled exceptions. Personally, I think it's a good change, but if you want to make bug-to-bug change you need to add bare `except:` block with no-op (and don't need `finally` at all, actually.)

PR #12875, opened 2025-12-08
fix(deps): update confluent-kafka to 2.12.1

cezarmoise:
This version adds wheels for python 3.14, removing the need to build it.

### Testing

- [ ]

### Reminders


pehala (dismissed):

Dockerfile
```diff
@@ -37,10 +29,7 @@ RUN apt-get install -y --no-install-recommends \
     cmake \
     libssl-dev \
     zlib1g-dev \
-    libffi-dev \
-    librdkafka-dev \
-    libev4 \
```
fruch: I prefer we leave the libev4 and libev-dev in, they are not kafka related, they are introduce cause of python-driver, and we want to keep the ability of picking it even if there no version wheels

cezarmoise: re-added them

fruch (approved):
LGTM

Issue #6935, opened 2023-12-12
`disrupt_create_index` sets timeout for the operation not considering the node count


vponomaryov:
Running the `disrupt_create_index` nemesis on the multiDC cluster consisting of the `12` nodes the `CREATE INDEX` operation takes more time than it is set in the nemesis:

```
operation 'CREATE_INDEX' took 7692.512597799301s and soft-timeout was set to 7200s
...
operation 'CREATE_INDEX' took 7482.790784358978s and soft-timeout was set to 7200s
```

In current case we have not standard, for test, number of nodes - `12` and those have big load - `50-100%`.

So, we should consider changing the `soft-timeout` to a bigger value.
Probably depending on the node count and number of regions.

## Installation details

Kernel Version: 5.15.0-1050-aws
Scylla version (or git commit hash): `5.5.0~dev-20231205.4505a86f46df` with build-id `2355da584169ed4dedccb225347f30c5896efe06`

Cluster size: 12 nodes (i3en.2xlarge)

Scylla Nodes used in this run:
- parallel-topology-schema-changes-mu-db-node-fd2981d7-9 (18.132.211.226 | 10.3.10.214) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-8 (3.11.8.196 | 10.3.11.25) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-7 (18.170.49.184 | 10.3.9.2) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-6 (3.250.115.191 | 10.4.11.202) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-5 (63.32.53.133 | 10.4.9.201) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-4 (34.245.14.246 | 10.4.8.120) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-30 (18.133.183.109 | 10.3.9.37) (shards: -1)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-3 (63.33.51.242 | 10.4.8.31) (shards: 7)
- parallel-topology-schema-changes-mu-db-node-fd2981d7-29 (63.33.192.144 | 10.4.8.79) (shards: 7)
[...]

PR #12881, opened 2025-12-09
fix(nemesis): disrupt_create_index timeout increased x2

timtimb0t:
timeout been increased to avoid timeout events during testing
fixes: https://github.com/scylladb/scylla-cluster-tests/issues/6935

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -5207,7 +5207,8 @@ class Nemesis(NemesisFlags):
                 raise UnsupportedNemesis(
                     "Tried to create already existing index. See log for details")
             try:
-                with adaptive_timeout(operation=Operations.CREATE_INDEX, node=self.target_node, timeout=14400) as timeout:
+                # timeout changed from 14400 to 28800 to avoid softimeout errors
+                with adaptive_timeout(operation=Operations.CREATE_INDEX, node=self.target_node, timeout=28800) as timeout:
                     with self.action_log_scope("Wait for index to be built"):
                         wait_for_index_to_be_built(self.target_node, ks, index_name, timeout=timeout * 2)
                 verify_query_by_index_works(session, ks, cf, column)


pehala:
That seems too much, do we data on how long it usually takes?

fruch:
> > I'm not sure about it, and not sure if it's possible to estimate the time, there are too many variables: number of nodes, table size, current load, instance types, background Scylla operations. We can track the build status by polling Scylla, but that's almost useless — we only get "in progress" or "done".
> 
> I would like to see some number before just doubling timeout, given it is way too long already IMO. @soyacz @fruch Do we have something regarding adaptive timeout in argus?

what's something ? i'm not sure I understand.

soft time out is what you pass to adaptive_timeout, the hard timeout is what you pass to the actual commands (i.e. nodetool or `wait_for_index_to_be_built` in this case)

regardless the bug report is from 2023 (and the timeout was 2h), after that beginning 2024, I committed 3cd8d055bf348e8b7014ac9c436e191a92bcfd79

I don't think extending it more would help anyone, without any fresh reports of issues.

pehala:
> what's something ? i'm not sure I understand.

I wanted to see something like average time it takes for a job to build this. I got my answer on Slack, that you can check the actual time individually, which should be enough to decide what is the correct timeout for this operation.



> soft time out is what you pass to adaptive_timeout, the hard timeout is what you pass to the actual commands (i.e. nodetool or `wait_for_index_to_be_built` in this case)
> 
> regardless the bug report is from 2023 (and the timeout was 2h), after that beginning 2024, I committed [3cd8d05](https://github.com/scylladb/scylla-cluster-tests/commit/3cd8d055bf348e8b7014ac9c436e191a92bcfd79)
> 
> I don't think extending it more would help anyone, without any fresh reports of issues.

I agree

PR #12962, opened 2025-12-17
sct argus email reports

k0machi:
- **chore(argus): Update argus to 0.15.9**
  

- **feature(sct.py): Argus-based email reports**
  This commit enables new Argus functionality that allows sending email
  reports from SCT using Argus instead of collected `email_data.json`.
  This allows SCT to be able to send email as long as it at least managed
  to send off basic argus report (so, just createArgusTestRun is enough to
  be able to send email).
  

- [x] Email title `[qa] [FAILED] scylla-staging/alexey/alexey-argus-testing#158: 17/12/2025 12:32:54`
- [x] [Jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/alexey-argus-testing/158/)

### DIFF
--- a/vars/artifactsPipeline.groovy
+++ b/vars/artifactsPipeline.groovy
@@ -10,6 +10,7 @@ def call(Map pipelineParams) {
             AWS_ACCESS_KEY_ID     = credentials('qa-aws-secret-key-id')
             AWS_SECRET_ACCESS_KEY = credentials('qa-aws-secret-access-key')
             SCT_GCE_PROJECT = "${params.gce_project}"
+            SCT_ENABLE_ARGUS_REPORT = "1"
         }
         parameters {
             separator(name: 'CLOUD_PROVIDER', sectionHeader: 'Cloud Provider Configuration')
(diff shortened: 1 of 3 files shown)


soyacz (commented):
please also attach example from artifact tests

soyacz:
> @aleksbykov @soyacz

@k0machi 
> 
> 1. what about rolling upgrade ? I think it should be the basic as well ? would be followup ?

We miss info of base/upgraded version I suppose.

> 2. we should record a demo, show it in the weekly dev meeting

I think we should wait until we have full feature set (including results). So people can craft own email reports fully.

k0machi:
> > 1. what about rolling upgrade ? I think it should be the basic as well ? would be followup ?
> 
> We miss info of base/upgraded version I suppose.

Technically we don't - it's just not shown yet in the "Overview" block, packages view would show target/upgrade/base versions however.

> 
> > 2. we should record a demo, show it in the weekly dev meeting
> 
> I think we should wait until we have full feature set (including results). So people can craft own email reports fully.

I plan on adding perf (and other) results support soon - I suppose we could wait until then

soyacz:
> > Small issue - `sendEmail` stage is currently before Finish Argus Test Run stage, which leads to incorrect status being reported if `RunSctStages` crashes. Is it safe to move it after the FinishTestRun stage?
> > @soyacz @fruch
> 
> I'm o.k. with that, but maybe it's deem its own commit

Shall I do it or we wait for @k0machi?
> 
> Also we should list we need todo it on all the followup (maybe we can do right away?

I've created subtasks for the epic: https://github.com/scylladb/argus/issues/650

PR #12970, opened 2025-12-18
fix(get_scylla_docker_repo_from_version): handle dev and user input

fruch:
Since #12953 changd the logic a bit on we automaticlly find the docker repo

two cases wasn't working, and fixed in this commit:
* handeling the dev versions, they are always coming fomr nightly repo
* respect user input, i.e. if user set specific repo, we shouldn't never override it

### Testing

- [ ] unit tests

### Reminders

### DIFF
--- a/unit_tests/test_version_utils.py
+++ b/unit_tests/test_version_utils.py
@@ -761,27 +761,30 @@ def test_get_branched_repo(scylla_version, distro, expected_repo):
 
 
 @pytest.mark.parametrize(
-    "version, expected_repo",
+    "version, expected_repo, input_repo",
     (
-        ("6.2.2", "scylladb/scylla"),
-        ("6.2.3", "scylladb/scylla"),
-        ("6.2.4", "scylladb/scylla"),
-        ("6.2.66", "scylladb/scylla"),
-        ("2024.1.1", "scylladb/scylla-enterprise"),
-        ("2024.2.13", "scylladb/scylla-enterprise"),
-        ("2024.2.14", "scylladb/scylla-enterprise"),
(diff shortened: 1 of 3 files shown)


unit_tests/test_version_utils.py
```diff
@@ -761,27 +761,30 @@ def test_get_branched_repo(scylla_version, distro, expected_repo):
...
+        ("2025.1.0", "scylladb/scylla", None),
+        ("2025.2.99", "scylladb/scylla", None),
+        ("2025.4.0", "scylladb/scylla", None),
+        ("branch-2025.1:latest", "scylladb/scylla", None),
+        ("branch-2025.2:latest", "scylladb/scylla", None),
+        ("branch-2025.3:latest", "scylladb/scylla", None),
+        ("branch-2025.4:latest", "scylladb/scylla", None),
+        ("master:latest", "scylladb/scylla", None),
+        ("2026.1.0-dev-0.20251217.55f4a2b75472", "scylladb/scylla-nightly", None),
+        ("2026.1.0-dev-0.20251217.55f4a2b75472", "scylladb/scylla-nightly", "scylladb/scylla-nightly"),
```
abramche: I would introduce a helper property to `ComparableScyllaVersion` i.e. `isNightlyBuild()` - could be as easy as asserting that `len(parsed_version) > 3` and in this case add the first if clause to the semver block of `get_scylla_docker_repo_from_version()`

```
    try:
        comparable_version = ComparableScyllaVersion(scylla_version)
        if comparable_version.isNightlyBuild():
            return "scylladb/scylla-nighly"
```

and that's it - no other changes needed (except adding a few more test cases to unit tests to make sure it works)

PR #12974, opened 2025-12-18
feat(treewide): Annotate cases and files owned by Storage QA

pehala:
### Testing

- [ ] None - Only doc change

### DIFF
--- a/test-cases/features/out-of-space-prevention/longevity-out-of-space-prevention-secondary-index.yaml
+++ b/test-cases/features/out-of-space-prevention/longevity-out-of-space-prevention-secondary-index.yaml
@@ -1,3 +1,5 @@
+# Maintained by Storage QA
+
 test_duration: 1440
 # This will fill a 3 x i4i.large cluster to 90% disk usage
 prepare_write_cmd: [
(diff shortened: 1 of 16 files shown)


pehala:
@cezarmoise @yarongilor @jsmolar anymore cases/files that are considered ours?

test-cases/features/out-of-space-prevention/longevity-out-of-space-prevention-secondary-index.yaml
```diff
@@ -1,3 +1,5 @@
+# Maintained by Storage QA (Cezar Moise)
```
cezarmoise: Why do these have my name on it?

fruch:
> > Why not add the feature folders to the codeowners file?
> 
> Could be as well, but the CODEOWNERS is too long and right now does not cover testcase so I did not add it here. @fruch Would you like the test_case folder and/or pipelines also covered in codeowners?

it's gonna be long anyhow, but I don't mind add test case, or folders of tests to it.
maybe later it won't be the source of truth, and we'll generate the test case part.

maybe it's worth building github team for this, since it less individual

yarongilor (approved):
sstable does sound related to storage , but it's fine to be taken by other team as well.
LGTM

PR #12978, opened 2025-12-18
fix(vector-search): update Scylla version used for VS cluster in Cloud

mikliapko:
Add `2025.4.0-rc7` to the list of supported by Cloud. 

VS in Cloud is officially released with this version of Scylla. 
The previous 2025.4.0-rc7 removed from the list since VS in Cloud did not make it with RC6, but with RC7

### Testing

- [x] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/mikita/job/vector-search-pre-ga-testing/job/sirenada-stg-vector-search-average-recall-aws/7/

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -4226,9 +4226,9 @@ class SCTConfiguration(dict):
 
         # validate if selected Scylla version is supported
         supported_versions = [v["version"] for v in cloud_api_client.get_scylla_versions()["scyllaVersions"]]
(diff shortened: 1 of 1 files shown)


dimakr:
> It's a one by one deployment

I see. That is the problem - the current check assumes that all VS nodes deployment (as per the number of AZs) start at once. I'll create a task to improve that `wait_for_vs_nodes` waiter.
Interesting that it never happened before, that SCT missed waiting for one of VS nodes, at least I see this incident for the 1st time.

Issue #7287, opened 2024-03-20
When encryption is enabled nemesis fails to start on Docker backend with `cp: cannot create regular file '/home/scylla-test/.cassandra/cqlshrc': Permission denied` error


dimakr:
### Issue description

If encryption is enabled in test configuration, any nemesis fails on the step of client certificate installation:
```
Traceback (most recent call last):
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/cluster.py", line 3717, in node_setup
    cl_inst.node_setup(_node, **setup_kwargs)
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/cluster_docker.py", line 311, in node_setup
    node.config_setup(append_scylla_args=self.get_scylla_args())
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/cluster.py", line 1591, in config_setup
    self.proposed_scylla_yaml
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/cluster.py", line 430, in proposed_scylla_yaml
    scylla_yml = ScyllaYaml(**self._proposed_scylla_yaml_properties)
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/cluster.py", line 422, in _proposed_scylla_yaml_properties
    return node_params_builder.dict(exclude_none=True) | certificate_params_builder.dict(exclude_none=True)
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/provision/common/builders.py", line 69, in dict
    prop_value = getattr(self, prop)
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/provision/scylla_yaml/certificate_builder.py", line 43, in client_encryption_options
    certificate=os.path.join(self._ssl_files_path, 'client', os.path.basename(CLIENT_CERTFILE)),
  File "/home/dmitriy/.pyenv/versions/3.10.0/lib/python3.10/functools.py", line 970, in __get__
    val = self.func(instance)
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/provision/scylla_yaml/certificate_builder.py", line 34, in _ssl_files_path
    install_client_certificate(self.node.remoter)
  File "/home/dmitriy/Work/Scylla/scylla-cluster-tests/sdcm/provision/helpers/certificate.py", line 35, in install_client_certificate
[...]

PR #13065, opened 2025-12-29
fix(docker): fix ssl creation and sync to docker

fruch:
this add into place the creation of ssl for the docker backend

Fix: #7287

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/remote/docker_cmd_runner.py
+++ b/sdcm/remote/docker_cmd_runner.py
@@ -5,7 +5,6 @@ from typing import TYPE_CHECKING
 from shlex import quote
 from pathlib import Path
 import shutil
-import logging
 
 from invoke.runners import Result
 from invoke.exceptions import UnexpectedExit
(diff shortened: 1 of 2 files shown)


fruch:
@dimakr 

I need help with this one

I got to SSL files are created and copied

but to the wrong place:
```
bash-5.1$ ls /etc/scylla/ssl_conf/172.17.0.5/
ca.pem             client-facing.crt  client-facing.key  truststore.jks
```

I think the docker base send_files, has a bug, and it not copying the content of the src folder, but the folder itself.

fruch:
> > I think the docker base send_files, has a bug, and it not copying the content of the src folder, but the folder itself.
> 
> Right, the contract for `sdcm.remote.remote_base.RemoteCmdRunnerBase.send_files`
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/1b9c582e09769f10c0a0ca39dd69d764512a0e31/sdcm/remote/remote_base.py#L332
> 
> (i.e. copy files only, when src directory ends with slash)
> But seems like docker_cmd_runner is not following it due to
> 
> https://github.com/scylladb/scylla-cluster-tests/blob/1b9c582e09769f10c0a0ca39dd69d764512a0e31/sdcm/remote/docker_cmd_runner.py#L163
> 
> - when archive name is calculated it includes the directory holding files as well. So in tar the arcnames are like `172.17.0.5/ca.pem`, `172.17.0.5/client-facing.crt` etc.
> I'll add a fix for this

see ee405b5c87da1ffd82bd6c8cdfa40934b0b7c5b1 I think it's a fix, maybe not a full one
but it a starting point with tests

PR #13073, opened 2025-12-30
chore(CODEOWNERS): update k8s, operator codeowners

abramche:
add ownership missed in https://github.com/scylladb/scylla-cluster-tests/pull/12640

### Testing
- [x] not needed

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -32,6 +32,18 @@
 /sdcm/cloud_api_client.py   @dimakr
 /sdcm/cloud_api_utils.py    @dimakr
 
+# k8s
+/sdcm/cluster_k8s/          @abramche
+/sdcm/k8s_configs/          @abramche
+/sdcm/utils/k8s/            @abramche
+/defaults/k8s*.yaml         @abramche
+
+# Operator functionality
+/jenkins-pipelines/operator/                @abramche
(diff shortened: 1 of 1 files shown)


.github/CODEOWNERS
```diff
@@ -32,6 +32,17 @@
...
+# k8s
+/sdcm/cluster_k8s/          @abramche
+/sdcm/k8s_configs/          @abramche
+/defaults/k8s*.py           @abramche
+
+# Operator functionality
+/jenkins-pipelines/operator/                @abramche
+/test-cases/scylla-operator/                @abramche
+sdcm/fill_db_data.py                        @abramche
+upgrade_test.py                             @abramche
```
pehala: are these two really run only on k8s? based on their name I would expect them to be used for normal scylla as well

abramche: not only, but they are not in the k8s section. I can move these two elsewhere (but where? we do not seem to have a test structure for these)

`fill_db_data.py` is only used by `upgrade_test.py`

fruch: it's both, add my name there as well, as the owner of scylla upgrades

if you think a refactor is needed, got right ahead, please don't  refactor only cause of aligning the code ownership.

fruch (approved):
LGTM

vponomaryov (approved):
LGTM

PR #13084, opened 2025-12-31
feature(oci): add OCI SDK dependency and utilities for sct-runner support

fruch:
Add initial Oracle Cloud Infrastructure (OCI) support for SCT runner:

- Add oci==2.164.2 dependency to pyproject.toml
- Add get_oci_credentials() and get_oci_ssh_key_pair() to keystore.py with support for both S3-stored oci.json and local ~/.oci/config
- Create sdcm/utils/oci_utils.py with OCI utility functions:
  - Client factories: get_oci_compute_client, get_oci_identity_client, get_oci_network_client
  - Instance operations: list_instances_oci, filter_oci_by_tags, oci_public_addresses, oci_private_addresses, wait_for_instance_state
  - Image operations: get_ubuntu_image_ocid, wait_for_image_state
  - AD helpers: get_availability_domains, resolve_availability_domain
  - OciService singleton class for convenient access
- Add comprehensive unit tests in unit_tests/test_oci_utils.py
- Add integration tests in unit_tests/test_oci_utils_integration.py

This is step 1-3 of the OciSctRunner implementation plan.

### Testing

- [x] 🟢   builder and sct-runner working: https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/perf-regression-predefined-throughput-steps-sanity-vnodes-baremetal-test/22

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -126,6 +126,7 @@ from sdcm.utils.aws_builder import AwsCiBuilder, AwsBuilder
 from sdcm.utils.gce_region import GceRegion
 from sdcm.utils.gce_builder import GceBuilder
 from sdcm.utils.oci_region import OciRegion
+from sdcm.utils.oci_builder import OciBuilder
 from sdcm.utils.aws_peering import AwsVpcPeering
 from sdcm.utils.get_username import get_username
 from sdcm.utils.sct_cmd_helpers import add_file_logger, CloudRegion, get_test_config, get_all_regions
@@ -2210,6 +2211,8 @@ def configure_jenkins_builders(cloud_provider, regions):
             AwsBuilder.configure_in_all_region(regions=regions)
         case "gce":
             GceBuilder.configure_in_all_region(regions=regions)
+        case "oci":
+            OciBuilder.configure_in_all_region(regions=regions)
         case "azure":
             raise NotImplementedError("configure_jenkins_builders doesn't support Azure yet")
 
--- a/sdcm/utils/oci_builder.py
+++ b/sdcm/utils/oci_builder.py
@@ -1,0 +1,271 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2025 ScyllaDB
+
+import logging
+from functools import cached_property
+from typing import Optional
+from string import Template
+from ipaddress import ip_network
+from pathlib import Path
+
+import click
+import requests
+
+from sdcm.utils.oci_region import OciRegion
+from sdcm.sct_runner import OciSctRunner
+from sdcm.keystore import KeyStore
+from sdcm.utils.sct_cmd_helpers import get_all_regions
+
+LOGGER = logging.getLogger(__name__)
+
+OCI_JENKINS_CONFIG_TEMPLATE = Template(
+    (Path(__file__).parent / "builder_setup_groovy/oci_jenkind_plugin_config.groovy.tmpl").read_text()
+)
+
+
+class OciBuilder:
+    """Configure OCI compute instances as Jenkins builders.
+
+    This class creates instance configurations based on OCI compute resources
+    and adds the configuration to Jenkins using the OCI Compute Cloud plugin.
+    Following patterns from AwsBuilder and GceBuilder.
+    """
+
+    NUM_CPUS = 2
+    NUM_EXECUTORS = 1
+    VERSION = "v1"
+
+    def __init__(
+        self,
+        region: OciRegion,
+        params: Optional[dict] = None,
+        number: int = 1,
+    ):
+        """Initialize OCI Builder.
+
+        Args:
+            region: OciRegion instance for region-specific configuration
+            params: Optional parameters for builder configuration
+            number: Builder instance number (for multiple builders in same region)
+        """
+        self.region = region
+        self.number = number
+        self.params = params or {}
+        self.jenkins_info = KeyStore().get_json("jenkins.json")
+
+        # Use the first availability domain from the region
+        first_ad = region.availability_domains[0] if region.availability_domains else "a"
+        self.runner = OciSctRunner(region_name=self.region.region_name, availability_zone=first_ad, params=None)
+
+    @cached_property
+    def name(self) -> str:
+        """Generate builder name.
+
+        Example: oci-us-ashburn-1-qa-builder-v1-1
+        """
+        return f"oci-{self.region.region_name}-qa-builder-{self.VERSION}-{self.number}"
+
+    @cached_property
+    def jenkins_labels(self) -> str:
+        """Generate Jenkins labels for job routing."""
+        return f"oci-sct-builders-{self.region.region_name}-{self.VERSION}"
+
+    def _create_instance_configuration(self) -> dict:
+        """Create OCI instance configuration for Jenkins.
+
+        Returns:
+            Dictionary with instance configuration parameters
+        """
+        click.secho(f"{self.region.region_name}: creating instance configuration")
+
+        # Ensure SCT runner image exists
+        if not self.runner.image:
+            click.secho(f"{self.region.region_name}: building SCT runner image")
+            self.runner.create_image()
+
+        image_id = self.runner.image.id if self.runner.image else None
+        if not image_id:
+            raise ValueError(f"Unable to determine image ID for region {self.region.region_name}")
+
+        # Ensure VCN and internet gateway exist
+        _ = self.region.vcn
+        _ = self.region.internet_gateway
+
+        # Use runner's full availability domain for subnet lookup
+        full_ad = self.runner._full_availability_domain
+
+        # Get or create subnet
+        subnet = self.region.subnet(ad=full_ad)
+        if not subnet:
+            click.secho(f"{self.region.region_name}: creating subnet for {full_ad}")
+
+            region_index = self.region._region_index()
+            vcn_cidr = ip_network(self.region.SCT_VCN_CIDR_TMPL.format(region_index))
+            ad_index = self.region.availability_domains.index(full_ad)
+            subnet_cidr = list(vcn_cidr.subnets(prefixlen_diff=8))[ad_index]
+            self.region.create_subnet(ad=full_ad, ipv4_cidr=subnet_cidr)
+            subnet = self.region.subnet(ad=full_ad)
+
+        if not subnet:
+            raise ValueError(f"Unable to create or find subnet in {full_ad}")
+
+        return {
+            "image_id": image_id,
+            "shape": self.params.get("oci_shape", "VM.Standard.E4.Flex"),
+            "vcn_id": self.region.vcn.id,
+            "subnet_id": subnet.id,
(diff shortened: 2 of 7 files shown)


defaults/oci_config.yaml
```diff
@@ -0,0 +1,45 @@
...
+oci_image_db: ''
+oci_image_loader: ''
+oci_image_monitor: ''
+
+# Image Username
+oci_image_username: 'ubuntu'
+
+# Root Disk Size (GB)
+root_disk_size_db: 50
+root_disk_size_loader: 50
```
soyacz: 30GB for db and 20GB for loaders should be enough (taken from AWS)

longevity_operator_multi_tenant_test.py
```diff
@@ -15,7 +15,7 @@
 
 from longevity_test import LongevityTest
 from sdcm.utils.parallel_object import ParallelObject
-from sdcm.utils.operator.multitenant_common import MultiTenantTestMixin
+from sdcm.utils.k8s_operator.multitenant_common import MultiTenantTestMixin
```
pehala: This change does not belong in this PR, renaming should be done separately

.pre-commit-config.yaml
```diff
@@ -41,7 +41,11 @@ repos:
         args: [--allow-missing-credentials]
     -   id: detect-private-key
         language: system  # Override the language
-        exclude: ^data_dir/ssl_conf/.*$
+        exclude: >
+             (?x)^(
+                  unit_tests/test_oci_utils.py|
```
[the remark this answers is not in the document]
fruch: I favored having a realistic unittest, then this check, ontop of it we new have those checks in github, that we can assign and wavier.

longevity_operator_multi_tenant_test.py
fruch: and in this case it's builtins that are failing... i.e. logging.
```
  File "/home/fruch/projects/scylla-cluster-tests/sdcm/utils/oci_region.py", line 24, in <module>
    import logging
  File "/home/fruch/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/logging/__init__.py", line 26, in <module>
    import sys, os, time, io, re, traceback, warnings, weakref, collections.abc
  File "/home/fruch/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/re/__init__.py", line 127, in <module>
    import functools
  File "/home/fruch/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/functools.py", line 18, in <module>
    from collections import namedtuple
  File "/home/fruch/.local/share/uv/python/cpython-3.14.0-linux-x86_64-gnu/lib/python3.14/collections/__init__.py", line 39, in <module>
    from operator import eq as _eq
ImportError: cannot import name 'eq' from 'operator' (consider renaming '/home/fruch/projects/scylla-cluster-tests/sdcm/utils/operator/__init__.py' since it has the same name as the standard library module named 'operator' and prevents importing that standard library module)
```

pehala (commented):
> Lets organize code better this time, I suggest putting OCI into its own folder so like `sdcm/oci` or `sdcm/provisioning/oci`. It will make it much easier to maintain

I would still like this to be addressed

sdcm/sct_runner.py
```diff
@@ -1255,6 +1313,525 @@ def set_tags(sct_runner_info: SctRunnerInfo, tags: dict):
...
+    def list_sct_runners(cls, verbose: bool = True) -> list[SctRunnerInfo]:
+        sct_runners = []
+        instances = list_instances_oci(
+            tags_dict={"NodeType": cls.NODE_TYPE},
+            verbose=verbose,
+        )
+
+        for instance in instances:
+            tags = oci_tags_to_dict(instance.freeform_tags)
+            region = instance.availability_domain.rsplit("-", 1)[0]  # Extract region from AD
```
soyacz: according to AI, this will result in e.g. `Uocm:US-ASHBURN-AD` which is not a valid region name, breaking some stuff. Are you sure this works?

sdcm/sct_runner.py
```diff
@@ -1255,6 +1313,525 @@ def set_tags(sct_runner_info: SctRunnerInfo, tags: dict):
...
+            timeout=600,
+        )
+
+        LOGGER.info("Instance %s is running. Waiting for public IP...", instance.id)
+        self.instance = instance
+
+        public_ips = oci_public_addresses(
+            instance=instance,
+            compute_client=compute_client,
+            network_client=self.network_client,
```
soyacz: network_client should be selected alongside compute_client based on the same condition I presume

sdcm/sct_runner.py
```diff
@@ -1255,6 +1313,525 @@ def set_tags(sct_runner_info: SctRunnerInfo, tags: dict):
...
+        )
+
+        # For flexible instances, we need to specify ShapeConfig
+        # Standard.E4.Flex can have 1-32 OCPUs and 1-192 GB memory
+        # Determine OCPU and memory based on instance type
+        if instance_type == self.IMAGE_BUILDER_INSTANCE_TYPE:
+            # Builder instance: 2 OCPUs, 8 GB memory
+            ocpus = 2.0
+            memory_in_gbs = 8.0
+        elif instance_type == self.REGULAR_TEST_INSTANCE_TYPE:
```
fruch: this code is problematic, since it's the same instance types,
we should add the ocpu and memory into the instance types, or change it to a tuple lilke: 
` REGULAR_TEST_INSTANCE_TYPE = ("VM.Standard.E4.Flex", 2, 8 ) # 2 vcpus, 8G`

sdcm/utils/oci_region.py
```diff
@@ -0,0 +1,363 @@
...
+            cidr_block=str(ipv4_cidr),
+            compartment_id=self.compartment_id,
+            vcn_id=self.vcn.id,
+            display_name=name,
+            prohibit_public_ip_on_vnic=False,
+            security_list_ids=[self.security_list.id],
+            route_table_id=self.vcn.default_route_table_id,
+            dhcp_options_id=self.vcn.default_dhcp_options_id,
+            freeform_tags=self._base_tags(name),
+            ipv6_cidr_block=str(ipv6_cidr),
```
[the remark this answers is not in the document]
fruch: change the input param for it to be mandatory

scripts/setup_oci_sct_user.py
```diff
@@ -0,0 +1,756 @@
+#!/usr/bin/env python3
```
vponomaryov: For such stuff we have `utils` directory in the root of the project.

sdcm/keystore.py
```diff
@@ -144,6 +202,9 @@ def get_argus_rest_credentials_per_provider(self, cloud_provider: str | None = N
         return self.get_json("argus_rest_credentials.json")
 
     def get_baremetal_config(self, config_name: str):
+        local_config_path = Path(f"{config_name}.json")
```
vponomaryov: Why is it changed in scope of this PR?

sdcm/utils/oci_builder.py
```diff
@@ -0,0 +1,496 @@
...
+
+from sdcm.utils.oci_region import OciRegion
+from sdcm.sct_runner import OciSctRunner
+from sdcm.keystore import KeyStore
+from sdcm.utils.sct_cmd_helpers import get_all_regions
+
+LOGGER = logging.getLogger(__name__)
+
+OCI_JENKINS_CONFIG_TEMPLATE = Template(r"""
+// OCI Compute Cloud configuration for Jenkins
```
vponomaryov: I think better to have such a file separately and then just read its contents when needed...
Or copy it when needed...

Doing so we will have code linting and all that stuff...

scripts/setup_oci_sct_user.py
```diff
@@ -0,0 +1,756 @@
...
+    # Create group
+    group_details = CreateGroupDetails(
+        compartment_id=compartment_id,
+        name=group_name,
+        description="Group for SCT builders - created by setup_oci_sct_user.py",
+    )
+
+    try:
+        group = identity_client.create_group(group_details).data
+        print("✓ Group created successfully")
```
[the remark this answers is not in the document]
fruch: removed

vars/getJenkinsLabels.groovy
```diff
@@ -55,17 +57,19 @@ def call(String backend, String region=null, String datacenter=null, String loca
         cloud_provider = params.xcloud_provider?.trim()?.toLowerCase()
     }
 
-    if ((cloud_provider == 'aws' && region) || (cloud_provider == 'gce' && datacenter) || (cloud_provider == 'azure' && location) || (cloud_provider == 'aws-fibs' && region)) {
+    if ((cloud_provider == 'aws' && region) || (cloud_provider == 'gce' && datacenter) || (cloud_provider == 'azure' && location) || (cloud_provider == 'aws-fibs' && region) || (cloud_provider == 'oci' && region)) {
```
vponomaryov: Such an approach means we expect the region value for the Oracle cloud be specified in the AWS-specific field called `region`...
GCE has its own and Azure has its own...

So, why Oracle reuses the one from AWS?
Moreover everything else in the python code assumes we expect a dedicated value with the `oci_region_name` name...

Anyway, there are much more needed updates for the groovy files, so this may stay as is for now.

soyacz (dismissed):
LGTM

sct.py
```diff
@@ -124,6 +125,9 @@
 from sdcm.utils.aws_builder import AwsCiBuilder, AwsBuilder
 from sdcm.utils.gce_region import GceRegion
 from sdcm.utils.gce_builder import GceBuilder
+from sdcm.utils.azure_region import AzureRegion
```
vponomaryov: This one looks like a bugfix for the existing problem.

defaults/oci_config.yaml
```diff
@@ -0,0 +1,45 @@
+# OCI Configuration Defaults
+# This file contains default settings for OCI backend deployments
+
+instance_provision: 'on_demand'
+user_credentials_path: '~/.ssh/scylla_test_id_ed25519'
+
+# OCI Region Configuration
+oci_region_name:
```
vponomaryov: Why adding nonexitent yet options?

vars/getJenkinsLabels.groovy
fruch: it's probably a mistake, cause we didn't yet added the pipeline parameters needed 
I'm trying to use it with baremetal in other branch, and I changed it to oci_region... 

I'll bring my initial findings back here, but they would be partial, and we'll need to adapt the pipeline, one by one, as we gonna use them, with all of the needed parameters

sct.py
fruch: actually it just a mistake, and it's already imported 

removed

vars/getJenkinsLabels.groovy
fruch: now fixed this specific one

defaults/oci_config.yaml
fruch: remove all but `user_credentials_path`

sdcm/sct_runner.py
```diff
@@ -230,10 +263,34 @@ def install_prereqs(self, public_ip: str, connect_timeout: Optional[int] = None)
...
+                sudo pkill -9 apt || true
+                sudo pkill -9 apt-get || true
+                sleep 2
+            fi
+
             apt-get -qq clean
             apt-get -qq update
             apt-get -qq install --no-install-recommends python3-pip htop screen tree systemd-coredump rng-tools
             pip3 install awscli
+            pip3 install oci-cli
```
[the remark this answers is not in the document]
fruch: moved into cloud-init specific for OCI

(also, show me the code that uses awscli of the builder, and not from within hydra, I think you won't find it, 
in gce it's the cli of it comes install without us doing anything)

PR #13094, opened 2026-01-02
fix(latte): fix `schema_scale.rn` rune script

vponomaryov:
Cosmetic changes:
- Fix typo in the word `achieve`.
- Fix one of the assertion messages to provide proper param name.
- Make latte command fail sooner than later if we set inconsistent values
  for the `gauss_mean` and the `gauss_stddev` params. 
  Before it would try to run stress and fail at some step
  and now it will fail right away if values are inconsistent for the normal distribution.

Significant changes:
- Handle negative values from `normal()` distribution by clamping to 0
  Without it data validation may fail trying to read nonexistent rows.
  Probability grows with the number of unique seeds to the `normal` fn.
- Provide values for the normal distribution not limited by the number of table rows.

Details for the significant changes:
```
  We populate data using only non-negative row indexes.
  But in the main mixed stress we use normal distribution
  with `mean=100_000` and `stddev=20_000`.
  It means that running a test with `9_000_000_000` operations
  having the `table_group_size=100`
  we must get some negative values out of the `normal` distribution.
  And it should have made the data validation fail by not finding rows
  which are absent.
  But test worked.
  And the reason for working test was capping the input values
  to the `normal` fn by the number of rows in each table.
  It may be called a significant simplification
  where we get a huge chance to not hit any of the negative values
  just because we repeated the 200_000 seed values all the time.
  So, now it is fixed, our distribution is more unique and handles
  boundary cases correctly.
```

### Testing

- [scylla-staging/valerii/vp-longevity-5000-tables-latte#23](https://argus.scylladb.com/tests/scylla-cluster-tests/ed7b7497-1713-40b2-afcc-f26f4d51e03d)

### Reminders

### DIFF
--- a/data_dir/latte/schema_scale.rn
+++ b/data_dir/latte/schema_scale.rn
@@ -28,7 +28,7 @@ const GAUSS_STDDEV = latte::param!("gauss_stddev", 0);
 // NOTE: 'table_group_size', 'table_shift_count' and 'table_shift_interval_s' work together.
 // When we have a big amount of tables
 // we want to have a smooth workload all over them with adequate cache hit/miss ratio.
-// So, we achive it by making the load walk over a group/subset of tables ('table_group_size')
+// So, we achieve it by making the load walk over a group/subset of tables ('table_group_size')
 // for some time ('table_shift_interval_s') then we shift the table window by the
 // 'table_shift_count' tables which effectively make the queries use a changed subset of tables
 // for queries. And it is done in a loop based on the 'table_shift_interval_s'.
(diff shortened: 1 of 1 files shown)


vponomaryov:
This PR substitutes the following one:
- https://github.com/scylladb/scylla-cluster-tests/pull/13066

This PR is tested here:
- [scylla-staging/valerii/vp-longevity-5000-tables-latte#23](https://argus.scylladb.com/tests/scylla-cluster-tests/ed7b7497-1713-40b2-afcc-f26f4d51e03d)

fruch (approved):

Issue #13093, opened 2026-01-02
Drop is_enterprise check and assume all Scylla versions to have enterprise feature set


abramche:
SCT mixes up enterprise feature set with enterprise repository location in various places. @vponomaryov requests [in this PR](https://github.com/scylladb/scylla-cluster-tests/pull/12973#discussion_r2658103336) that the function responsible for enterprise feature discovery (`sdcm.utils.version_utils.is_enterprise()` ) to be removed.

Draft scope of work:
- remove all usages of `sdcm.utils.version_utils.is_enterprise()` and the function itself
- assume enterprise feature set in all places where it was used
- update and delete redundant tests

PR #13117, opened 2026-01-06
feat(test_config_get_version_based_on_conf): Simplify tests

pehala:
With pytest.param there is no need to bundle values into expected_outcome
It improves readability of the tests and makes it easier to understand what the parameters mean
Also changing assert to be straightforward

Includes fix to a integration test failing as well, can be extracted into separate PR if needed
2025.1 is considered opensource with recent changes to the mechanisms.
Issue gets much deeper and surfaced only lately. Given the culprit patch never made it to 2025.1
I think it is safe to remove the test, until https://github.com/scylladb/scylla-cluster-tests/issues/13093

### Testing

- [x] Locally

### DIFF
--- a/unit_tests/test_config_get_version_based_on_conf.py
+++ b/unit_tests/test_config_get_version_based_on_conf.py
@@ -37,19 +37,19 @@ def function_setup(monkeypatch):
 
 
 @pytest.mark.parametrize(
-    argnames="scylla_version, expected_docker_image, expected_outcome",
+    argnames="scylla_version, expected_docker_image, expected_version, expected_is_enterprise",
     argvalues=[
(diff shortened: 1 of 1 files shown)


abramche (commented):
tbh I would not spend time prettifying the test (not that it's not needed...), but it will need to be refactored completely once https://github.com/scylladb/scylla-cluster-tests/issues/13093 is done - consider that `is_enterprise` is going away completely - so `expected_is_enterprise` is going too.

Good job getting rid of `expected_outcome` though.. Also it feels just wrong to have both `scylla_version` and `expected_version`. This makes me want to split the test in two.

imo it's fine to only comment out the offending test failure until proper fix in #13093 is coming. This test case is bad anyway and not testing anything. The rest of the changes can come along with the is_enterprise refactoring

fruch (approved):
LGTM

Issue #13070, opened 2025-12-30
Data population fails with CDC queries being incompatible with tablets


abramche:
[Some Operator upgrade tests](https://argus.scylladb.com/tests/scylla-cluster-tests/f73bc6c1-ae87-4f61-9dfc-c8c33dd91494/events) fail with CDC queries being incompatible with tablets. Some of the queries (including the failing one) are ancient, dating back to 2017 and are possibly irrelevant.

```
2025-12-17 18:44:37.829: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=6a63e67e-3900-48b5-b5da-3651dc7cfcd2, source=UpgradeTest.test_kubernetes_operator_upgrade()
exception=self = <upgrade_test.UpgradeTest testMethod=test_kubernetes_operator_upgrade>
    def test_kubernetes_operator_upgrade(self):
        self.k8s_cluster.check_scylla_cluster_sa_annotations()
        InfoEvent(message="Step1 - Populate DB with data").publish()
>       self.prepare_keyspaces_and_tables()
upgrade_test.py:1308:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
sdcm/fill_db_data.py:3788: in prepare_keyspaces_and_tables
    self.cql_create_tables(session)
sdcm/fill_db_data.py:3544: in cql_create_tables
    session.execute(create_table)
cassandra/cluster.py:2679: in cassandra.cluster.Session.execute
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
>   ???
E   cassandra.InvalidRequest: Error from server: code=2200 [Invalid query] message="Cannot create CDC log for a table keyspace_fill_db_data.order_by_with_in_test, because keyspace uses tablets. See issue #16317."
cassandra/cluster.py:5054: InvalidRequest
```

Possible ways to resolve:
- fix/improve the skip logic in the queries to turn off this particular query ([seems](https://github.com/scylladb/scylla-cluster-tests/blob/master/sdcm/fill_db_data.py#L3526) to be non-functional now) - there is no time for this right now though
[...]

PR #13146, opened 2026-01-07
chore(operator upgrade pipelines): downgrade scylla versions to 2024

abramche:
I require that scylla versions operator pipelines to use 2024 as a base as the operator team agreed to use it as the base version. 

Eventual introduction of 2025.x is tracked by https://github.com/scylladb/scylla-cluster-tests/issues/13151

### Testing
- [x] not needed, the tests do not work yet (although with this change I ran 8/10 green)


vponomaryov:
> 2025 proved to be a harder effort and there is no time right now to deal with this immediately. It will be a task for the next two months to gradually re-enable it.

Issue reference?
Some details in here?
Right now it is unclear what the `2025 proved to be a harder effort` is based on...

vponomaryov:
> > > 2025 proved to be a harder effort and there is no time right now to deal with this immediately. It will be a task for the next two months to gradually re-enable it.
> > 
> > 
> > Issue reference? Some details in here? Right now it is unclear what the `2025 proved to be a harder effort` is based on...
> 
> Tablets are on the top of it as this is what immediately causes tests to fail (see #13070). "Disabling" tablets is also less trivial than I thought and there's much more SCT legacy to fight.
> 
> tl;dr: re-enabling 2025 has to be done in a separate branch with separate fixes. Scylla 2024 needs to be running now.

If we follow this way then need to write in the commit and PR descriptions that automatic enablement of tablets is the only reason for the switch.
Also, would be useful to see a ref to some issue describing why the tablets is an issue for the k8s scylla-operator...

Then, have you tried to update the test config the following way:
```
enable_tablets_on_upgrade: false
append_scylla_yaml:                                                                                    
  force_gossip_topology_changes: true
  enable_tablets: false
  tablets_mode_for_new_keyspaces: 'disabled'
```
?
If yes, and faced some issues, please share the test results.
If not, try it out.
I expect above config just work for you using vnodes and tablets specifics must be avoided in this case.
And if it works for you and you choose to use this config addition then also need to add an inline comment describing the reason.

abramche:
> If we follow this way then need to write in the commit and PR descriptions that automatic enablement of tablets is the only reason for the switch. 

I do not know that. I suspect there are more issues. Tablets are just the evident top of the iceberg. SCT has a lot of unobvious legacy, like [this sneaky bit](https://github.com/scylladb/scylla-cluster-tests/pull/13143).

> Then, have you tried to update the test config the following way:
> 
> ```
> enable_tablets_on_upgrade: false
> append_scylla_yaml:                                                                                    
>   force_gossip_topology_changes: true
>   enable_tablets: false
>   tablets_mode_for_new_keyspaces: 'disabled'
> ```
> 
> ? If yes, and faced some issues, please share the test results. If not, try it out.

I tried the [partial update](https://github.com/scylladb/scylla-cluster-tests/commit/024ef7abbf6f0f997a52ac3d5d8f13085fb3f713) (does not include everything from above), it did [not work](https://argus.scylladb.com/tests/scylla-cluster-tests/8cd168b1-b678-47ac-8f50-b3b0bcef3050/events), it was evident that Scylla still used tablets.

And anyway I do not want tablets to be disabled, I want them to be enabled in my tests. Thanks for the suggestion though, I will use it as an incremental 2025 update later.

I appreciate you helping in this matter, but right now, I need 2024 to be the base version. I managed to achieve 8/10 pass rate with this before. 2025 never passes.

vponomaryov:
> I tried the [partial update](https://github.com/scylladb/scylla-cluster-tests/commit/024ef7abbf6f0f997a52ac3d5d8f13085fb3f713) (does not include everything from above), it did [not work](https://argus.scylladb.com/tests/scylla-cluster-tests/8cd168b1-b678-47ac-8f50-b3b0bcef3050/events), it was evident that Scylla still used tablets.

Jenkins log:
```
23:41:37  < t:2026-01-06 21:41:36,220 f:sct_config.py   l:3438 c:sdcm.sct_config      p:INFO  > config_files:
23:41:37  < t:2026-01-06 21:41:36,220 f:sct_config.py   l:3438 c:sdcm.sct_config      p:INFO  > - test-cases/scylla-operator/kubernetes-operator-upgrade.yaml
23:41:37  < t:2026-01-06 21:41:36,220 f:sct_config.py   l:3438 c:sdcm.sct_config      p:INFO  > cs_duration: 
```

So, your commit at https://github.com/scylladb/scylla-cluster-tests/commit/024ef7abbf6f0f997a52ac3d5d8f13085fb3f713 where you add one more test config (`configurations/tablets_disabled.yaml`) was not applied in the test run.
See https://jenkins.scylladb.com/job/scylla-staging/job/eugene/job/upgrade/job/upgrade-operator-k8s-gke-test/9/parameters/ there you have explicit just one main config file.

So, my request is still actual, need to really apply those config options and see the results on the CI jobs you have problems with tablets (upgrades only?).

fruch (approved):
LGTM

if the operator team want to first test upgrades from 2024.1, and only later upgrades from newer versions, it's their call.

PR #13167, opened 2026-01-09
feature(c-s): Enable c-s java 4x driver support

soyacz:
Updated cassandra-stress to the latest (3.20.0) version that supports java 4.x driver.
Java 4.x driver is turned on when `-mode` includes `4x` param.
For that, new `c_s_driver` param were added. It supports '3x', '4x'
(when driver 4.x should be used) and 'random' (default) when upon test
startup one value is set and used along the entire test.
closes: https://github.com/scylladb/cassandra-stress/issues/101

### Testing

- [ ] - [longevity 4x](https://argus.scylladb.com/tests/scylla-cluster-tests/0e540003-8769-4817-8a7c-d7a01c2d4849) 
- [ ] - [upgrade test 4x](https://argus.scylladb.com/tests/scylla-cluster-tests/c2ff41e6-b308-42df-b79e-671eac8fac79)
- [ ] - [upgrade test 3x](https://argus.scylladb.com/tests/scylla-cluster-tests/1889cecb-df79-4f3f-9f40-1d1a0ca1afdb)

### Reminders

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -349,3 +349,5 @@ agent:
   binary_url: ""
   max_concurrent_jobs: 10
   log_level: "info"
+
+c_s_driver_version: '3'
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -16,6 +16,7 @@ Handling Scylla-cluster-test configuration loading
 """
 
 import os
+import random
 import re
 import ast
 import json
@@ -2746,6 +2747,13 @@ class SCTConfiguration(dict):
             type=bool,
             help="Whether or not to send email using argus instead of SCT.",
         ),
+        dict(
+            name="c_s_driver_version",
+            env="SCT_C_S_DRIVER_VERSION",
+            type=str,
+            choices=("3", "4", "random"),
+            help="cassandra-stress driver version to use: 3|4|random",
+        ),
     ]
 
     required_params = [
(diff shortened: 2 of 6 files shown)


defaults/test_default.yaml
```diff
@@ -349,3 +349,5 @@ agent:
   binary_url: ""
   max_concurrent_jobs: 10
   log_level: "info"
+
+c_s_driver: 'random'
```
pehala: Do we have visibility in Argus about this variable? C-S version can easily cause issues and having to dig deep to understand the whole config (with Random you would have to look into logs probably?) seems cumbersome.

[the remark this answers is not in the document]
vponomaryov: I agree that explicit 4x for master and status-quo 3x for existing releases is the way.

pehala:
> We need to discuss what should be default and 'randomness'. Introducing it will give us better coverage (no need to duplicate tests to verify both drivers) and will introduce more 'unpredictability'.

We do not need to duplicate tests eitherway IMO.
I do not think we should care about 3.x coverage, I consider it as old stable option, but I do not think customers should use it.
Even if we want 3.x coverage in regular tests we can just choose one with 4.x and one with 3.x I believe we have enough C-S tests to allow this.

Rest can use default

roydahan:
> I've switched the default to 3x - we could merge this PR with this default, backport it and after figuring out c-s issue switch default to random/4x (to be decided later).

I don't want random, at least not for now.
I also don't think we should change the default, at least not for now.
It will take us time to clear all issues and test its performance.

sdcm/sct_config.py
```diff
@@ -2746,6 +2747,13 @@ class SCTConfiguration(dict):
             type=bool,
             help="Whether or not to send email using argus instead of SCT.",
         ),
+        dict(
+            name="c_s_driver_version",
+            env="SCT_C_S_DRIVER_VERSION",
+            type=str,
+            choices=("3", "4", "random"),
+            help="cassandra-stress driver version to use: 3|4|random (default: random)",
```
fruch: remove the default from the help, it's auto generate with the defaults ... (and it's incorrect now)

soyacz:
I see we didn't backport previous versions of c-s to supported branches.
Do we want to backport this?

PR #13193, opened 2026-01-12
improvement(utils): use CQL driver instead of cqlsh for getting tables

dimakr:
Replaced cqlsh CLI-based approach with direct CQL driver queries in get_db_tables function.
This improves reliability by avoiding shell command execution and allows to avoid exhausting system resources of SCT runner instance, when the whole DB schema (with history of changes) is continuously retrieved during tests into SCT log, which eventually could result in SCT runner being OOM-killed.

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/12366

### Testing

- [x] :yellow_circle: [longevity-schema-topology-changes-12h-test](https://jenkins.scylladb.com/job/scylla-staging/job/dimakr/job/longevity-schema-topology-changes-12h-test/3/)
The test run shows that now we don't have millions of `ALTER TABLE` log lines caused by querying `DESC SCHEMA` with all history of changes:
```
❯ grep 'ALTER TABLE' ../sct.log.14012026 | wc -l
10065
❯ grep 'DESC SCHEMA' ../sct.log.14012026 | wc -l
0
```
This particular test config run still failed on SCT runner log collection (SCT runner was OOM-killed), but it is due to wider problem described in https://github.com/scylladb/scylla-cluster-tests/issues/12366#issuecomment-3754166902. We have too aggressive logs streaming from DB nodes to runner (particularly in long running scenarios like this one - 24 hours). The problem should be addressed separately by identifying which logs could be dropped on the client side of vector.dev service and not streamed to SCT runner.

### Reminders

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -52,7 +52,6 @@ from collections import OrderedDict
 import requests
 import boto3
 from botocore.exceptions import ClientError
-from invoke import UnexpectedExit
 from mypy_boto3_s3 import S3Client, S3ServiceResource
 from mypy_boto3_ec2 import EC2Client, EC2ServiceResource
 from mypy_boto3_ec2.service_resource import Image as EC2Image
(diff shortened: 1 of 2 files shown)


yarongilor:
why not completely removing `run_cqlsh` usage across sct?

fruch:
@dimakr what does 🔵 means ?

fruch:
> > @dimakr what does 🔵 means ?
> 
> ah, my mistake. Wanted to show that test is running, and used blue color, which saw in Argus. But blue is for scheduled tests if I understand, for running it is yellow.

We never defined it, hence I'm asking.

I think yellow is warning, i.e. it run and there something to notice, but not failed in the context of the PR.

Personally for running I use any of  clock option.

Regardless I would want this fix in

fruch (approved):
LGTM

PR #13215, opened 2026-01-14
format: Add import rules for ruff

pehala:
Only non automated change was:
* Move all content from monitorstack.__init__.py into service.py to prevent circular imports
   * Not ideal solution, but it does the job. In general import heavy logic should not be in __init__.py if there is also a essential classes that are used everywhere. 

Part of https://github.com/scylladb/scylla-cluster-tests/issues/7860

### Testing

- [x] CI

### Reminders


pehala:
> I think we should backport the rule, and apply on branches as well similar to what we did for `ruff format`
> 
> otherwise it would be painful to backport (this time we don't need new hydra images)

If we want to do this, lets do a bigger PR with more rules added (I planned on having one rule, one PR)

abramche (commented):
I'm fine with this as long as integration tests pass (this is a "massive non-change")

cezarmoise:
Don't forget to update `.git-blame-ignore-revs` after this is merged.

pehala:
Reviving this PR. Redid the change using AI (only 1 manual edit due to splitting of line and moving the ignore). CI failure can be safely ignore, the git-blame-ignore-revs is updated correctly.

Reminder that formatting PRs need to be merged with merge commit so commit SHAs are preserved

@scylladb/qa-maintainers Please take a quick look and merge if possible

PR #13247, opened 2026-01-18
fix(pipeline): pass billing_project parameter to performance tests

juliayakovlev:
The recently added  parameter caused the performance test trigger to fail with:
```
java.lang.NullPointerException: Cannot invoke method get() on null object
```

This occurred because  returns null when the parameter is not set.

As a temporary workaround to restore the trigger functionality:
1. Set a default empty value for the  parameter.
2. Pass the  parameter from the trigger to the performance pipeline.

### Testing

- [ ]

### Reminders


fruch (approved):
LGTM

Issue #13039, opened 2025-12-29
Resolve Nemesis Classes with arguments


pehala:
We have NemesisClasses that try to run disrupt methods with specific argument (https://github.com/scylladb/scylla-cluster-tests/blob/fe93bb81834f257099ee4d804310261406969bfe/sdcm/nemesis.py#L7234). However, due to the machinery behind Nemesis this is never done and default arguments are run instead.

We need to either fix or remove such cases. Fixing such cases would be considered as introducing a new nemesis so it needs enough testing.

PR #13287, opened 2026-01-21
fix(nemesis): Fix wrong nemesis usage

pehala:
Nemesis execution is not intuitive and the actual execution mechanism does not execute code inside `disrupt` method in nemesis classes, it is only used for pattern matching and then the same method is executed with default arguments.
This PR fixes it and as a consequence we will now test more nemesis, so needs to carefully tested.

ManagerBackup nemesis are in COMPLEX_NEMESIS and as such are not executed by Sisyphus, so for them this PR is not critical but it improves them regardless and brings them closer to how nemesis should look like

* Simplifies manager backup nemesis logic
   * Split into two disrupt methods
   * Remove single use methods
* Fix ClusterRollingRestartRandomOrder by adding new disrupt method   

Fixes https://github.com/scylladb/scylla-cluster-tests/issues/13039

### Testing

- [x] Nemesis runs
    * [x] [ClusterRollingRestartRandomOrder](https://argus.scylladb.com/tests/scylla-cluster-tests/a277e63e-80d7-4a28-8831-17ebf7b77c29/nemesis)

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -864,7 +864,6 @@ class NemesisRunner:
         with self.action_log_scope(f"Rolling restart cluster. random order: {random_order}"):
             self.cluster.restart_scylla(random_order=random_order)
 
-    @target_all_nodes
     def disrupt_rolling_restart_cluster_random(self):
         self.disrupt_rolling_restart_cluster(random_order=True)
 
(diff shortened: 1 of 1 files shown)


yarongilor (changes requested):
the manager backup nemesis works fine in performance test in:
scylla-enterprise-perf-manager-native-backup-nemesis.jenkinsfile
It is not used by Sisyphus, no need to modify it.

pehala:
> the manager backup nemesis works fine in performance test in:
> scylla-enterprise-perf-manager-native-backup-nemesis.jenkinsfile
> It is not used by Sisyphus, no need to modify it.

Why it is not used by Sisyphus? It has class, it is in the yamls, so it should be used

pehala:
> Why it is not used by Sisyphus? It has class, it is in the yamls, so it should be used

Discussed offline: 
Yaron is correct, it is in COMPLEX_NEMESIS and as such is not run. This patch brings it closer to running as proper nemesis, more PRs are needed but this is still improvement and a pushes it in a good direction. Followups will enable it so it runs regularly in supported environments

Issue #13173, opened 2026-01-09
`ModifyTableMonkey` nemesis randomly chooses a sub-method which may be disruptive, but it is marked as `disruptive = False`


vponomaryov:
Test run: [scylla-2025.3/tier1/longevity-multidc-schema-topology-changes-12h-test#24](https://argus.scylladb.com/tests/scylla-cluster-tests/8c7ea92b-04f0-43ac-ac62-48c75da5f117)

This test runs 2 `SisyphusMonkey` nemesis threads in parallel:
```
2026-01-08 10:44:41,301 f:sct_config.py   l:3201 c:sdcm.sct_config      p:INFO  > nemesis_class_name: SisyphusMonkey:1 SisyphusMonkey:1
...
2026-01-08 10:44:41,301 f:sct_config.py   l:3201 c:sdcm.sct_config      p:INFO  > nemesis_selector:
2026-01-08 10:44:41,301 f:sct_config.py   l:3201 c:sdcm.sct_config      p:INFO  > - topology_changes
2026-01-08 10:44:41,301 f:sct_config.py   l:3201 c:sdcm.sct_config      p:INFO  > - schema_changes and not disruptive
```

And it happened that one of them (`disrupt_bootstrap_streaming_error`) started decommission process of one node and another (`disrupt_modify_table` + `modify_table_twcs_window_size` method) triggered stopping of another one.
The time diff is just `1.5s`:
```
2026-01-08 20:59:14,381 f:nemesis.py      l:5834 c:sdcm.nemesis         p:INFO  > sdcm.nemesis.SisyphusMonkey: \
    Decommission added node
...
2026-01-08 20:59:15,660 f:decorators.py   l:127  c:sdcm.utils.decorators p:DEBUG > BEGIN: stop_scylla <AWSNode>
```

And it led to the expected decommission error:
```
error executing POST request to http://localhost:10000/storage_service/decommission with parameters {}: remote replied with status code 500 Internal Server Error:
std::runtime_error (Decommission failed. See earlier errors (Canceled. Dead nodes: {c24b2e38-31da-4eec-8d34-34ab6266a5d2}). Request ID: e74ceafa-ecd4-11f0-a16d-1008bd374722)
```

So, decommission of the `node-32` failed just because `node-10` was stopped concurrently.

Node-10:
```
Jan 08 20:59:15.924191 parallel-topology-schema-changes-mu-db-node-8c7ea92b-10 systemd[1]: \
[...]

PR #13288, opened 2026-01-21
feat(nemesis): Split disrupt_modify_table into separate nemeses

pehala:
Mark `ModifyTableTwcsWindowSizeMonkey` as disruptive
Fixes https://github.com/scylladb/scylla-cluster-tests/issues/13173

### Testing

- [x] None
    - Given the code did not change, I do not think we need to run them. If needed for sanity reason, I can choose one to run 


sdcm/nemesis.py
```diff
@@ -7454,6 +7437,37 @@ def disrupt(self):
         self.disrupt_manager_backup(object_storage_upload_mode=ObjectStorageUploadMode.NATIVE, label="native_backup")
 
 
+class ModifyTableMonkey(Nemesis):
+    """
+    Selected number of nemesis that is focused on scylla-operator functionality
```
vponomaryov: Copy-paste left-over

configurations/nemesis/ModifyTableMonkey.yaml
```diff
@@ -1,2 +0,0 @@
-nemesis_class_name: 'ModifyTableMonkey'
```
vponomaryov: Why is it deleted?

It must be used in the following files you updated below:
- jenkins-pipelines/oss/nemesis/longevity-5gb-1h-ModifyTableCachingMonkey-aws.jenkinsfile
- jenkins-pipelines/oss/nemesis/longevity-5gb-1h-ModifyTableCachingMonkey-azure.jenkinsfile
- jenkins-pipelines/oss/nemesis/longevity-5gb-1h-ModifyTableCachingMonkey-docker.jenkinsfile
- jenkins-pipelines/oss/nemesis/longevity-5gb-1h-ModifyTableCachingMonkey-gce.jenkinsfile

Which, afterwards, must be renamed back.

vponomaryov (changes requested):

pehala:
> I do agree with @vponomaryov first response, we had in mind just splitting the problematic part out this is is much bigger change, and for example a bit more problematic to backport to releases, which are suffering from it.

To which release do we want to backport?
The code I am changing is old and does not rely on any recent patches/features so it should be safe to backport. Only thing that might be problematic might be the pipelines, which I can solve manually in each conflict (i.e. regenerate it per backport), so not a big deal IMO. I will split commit in this PR into three commit (split the disruptive nemesis, refactor, regenerate pipelines) so it is easier backported and if you still deem it too big of a change to backport we can cherry pick the first one into separate PR

> 
> if we are doing this kind of refactoring, can you improve the docstrings of each one as part of such change ?

For sure, I am open to any kind of improvements to these nemesis while I am at it

PR #13298, opened 2026-01-22
chore: Remove redundant nemesis configs/pipelines

pehala:
Run create-nemesis-pipelines on a empty directories to identify redundant ones.
Currently the nemesis generating code does not care about pipelines which are not supposed to be there, it is rare we delete any nemesis so this should not be common occurance

### Testing

- [x] None

### Reminders


dimakr (approved):
LGTM

Issue #13396, opened 2026-01-29
Scyllabench failed to run with counter mode: `Unknown scylla-bench mode: ScyllaBenchModes.COUNTER_UPDATE`


aleksbykov:
Failed to run Job with counters and scyllabench stress tool

Error:
```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _
self = <sdcm.scylla_bench_thread.ScyllaBenchThread object at 0x7704115996a0>
    def set_hdr_tags(self) -> list:
        if self.sb_mode.lower() == "write":
            return ["co-fixed"]
        elif self.sb_mode.lower() == "mixed":
            return ["co-fixed-write", "co-fixed-read"]
        elif self.sb_mode.lower() == "read":
            return ["co-fixed"]
        else:
>           raise ValueError(f"Unknown scylla-bench mode: {self.sb_mode}")
E           ValueError: Unknown scylla-bench mode: ScyllaBenchModes.COUNTER_UPDATE
sdcm/scylla_bench_thread.py:182: ValueError
```

[Argus](https://argus.scylladb.com/test/2ae264f5-ebfd-489c-b506-5c8bd7856cc9/runs?additionalRuns[]=813e9c6b-3b16-4af3-8997-bf513e4f6a31)

Scylla version: `2026.1.0~rc0-20260125.f94296e0ae43` with build-id `9680213fda6f301234c43da8ca27e47953987cd8`

Kernel Version: `6.14.0-1018-aws`
## Installation details

Cluster size: 9 nodes (i4i.2xlarge)

Scylla Nodes used in this run:

    - longevity-counters-multidc-2026-1-db-node-813e9c6b-1 (18.202.219.172 | 10.4.1.211) (shards: 7)

    - longevity-counters-multidc-2026-1-db-node-813e9c6b-2 (34.254.65.134 | 10.4.6.103) (shards: 7)

    - longevity-counters-multidc-2026-1-db-node-813e9c6b-3 (108.130.190.99 | 10.4.10.7) (shards: 7)

    - longevity-counters-multidc-2026-1-db-node-813e9c6b-4 (16.145.87.234 | 10.15.1.34) (shards: 7)

    - longevity-counters-multidc-2026-1-db-node-813e9c6b-5 (35.162.27.8 | 10.15.7.69) (shards: 7)

[...]

PR #13399, opened 2026-01-29
fix(hdr): fix s-b missing support for counter mode

soyacz:
Not all s-b modes are supported by hdr tags.

Added proper cases for all.

fixes: https://github.com/scylladb/scylla-cluster-tests/issues/13396

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/scylla_bench_thread.py
+++ b/sdcm/scylla_bench_thread.py
@@ -172,11 +172,11 @@ class ScyllaBenchThread(DockerBasedStressThread):
         self.stop_test_on_failure = stop_test_on_failure
 
     def set_hdr_tags(self) -> list:
-        if self.sb_mode.lower() == "write":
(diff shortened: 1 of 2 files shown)


sdcm/scylla_bench_thread.py
```diff
@@ -172,11 +172,11 @@ def __init__(
         self.stop_test_on_failure = stop_test_on_failure
 
     def set_hdr_tags(self) -> list:
-        if self.sb_mode.lower() == "write":
+        if self.sb_mode.lower() in ("write", "counter_update"):
```
vponomaryov: Looks like it can be simplified to the following:
```
        if self.sb_mode.lower() == "mixed":
            return ["co-fixed-write", "co-fixed-read"]
        else:
            return ["co-fixed"]
```

The same in the second file.

PR #13445, opened 2026-02-02
feature(platform-migration): add test for x86 to ARM cluster migration

dimakr:
Add platform migration test using add-and-replace node cycling approach. The test adds target platform nodes, then decommissions source nodes while running continuous stress workload. Data integrity is validated after migration is finished.

The following migration scenarios are covered:
  - sequential node-by-node migration
  - rack-aware batched migration (add new nodes per rack, decommission one node per rack at a time) - similar to sequential node-by-node migration with the difference that new nodes are not added all at once, but in batches per rack, to avoid doubling the cluster size at once
  - parallel decommission across DCs (one node of each DC at a time)

Test following test config and its overlays are added:
- single dc platform migration with 50% utilization - default config
- single dc with 10% utilization - overlay
- single dc for nodes - overlay
- batched-per-rack migration - batch specific scenario overlay
- multidc migration - multidc specific scenario overlay (2 DCs, 3 nodes each)

Refs: QAINFRA-42

### Testing

- [x] :green_circle: [single dc test config](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/177/)
- [x] :green_circle: [single dc batch per rack test config](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/206/)
- [x] :green_circle: [multidc test config](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/225/)
- [x] :green_circle: [single dc 10% utilization test config](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/215/)
- [x] :green_circle: [single dc vnodes 10% utilization test config](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/216/)
- [x] check during local run of the test that platform is changing on the cluster nodes:
```
>>> for n in self.db_cluster.nodes:
    n.remoter.run('uname -m').stdout
'x86_64\n'
'x86_64\n'
'x86_64\n'
...

for n in self.db_cluster.nodes:
    n.remoter.run('uname -m').stdout 
'aarch64\n'
'aarch64\n'
'aarch64\n'
```

### Reminders

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -1234,6 +1234,16 @@ AWS image type of the oracle node
 * appendable
 
 
+## **instance_type_db_target** / SCT_INSTANCE_TYPE_DB_TARGET
+
+Target AWS instance type for platform migration (e.g., i8g.2xlarge for ARM)
+
+**default:** N/A
+
+**type:** str
+* appendable
+
+
 ## **instance_type_runner** / SCT_INSTANCE_TYPE_RUNNER
 
 instance type of the sct-runner node
@@ -3014,6 +3024,26 @@ cassandra-stress commands. You can specify everything but the -node parameter, w
 
 **type:** str | list[str]
 * supports k8s multitenancy - see [multitenancy docs](k8s-multitenancy.md)
+
+
+## **stress_before_migration** / SCT_STRESS_BEFORE_MIGRATION
+
+Stress command to write data for post-migration validation
+
+**default:** N/A
+
+**type:** str
+* appendable
+
+
+## **verify_stress_after_migration** / SCT_VERIFY_STRESS_AFTER_MIGRATION
+
+Stress command to verify data after migration
+
+**default:** N/A
+
+**type:** str
+* appendable
 
 
 ## **stress_cmd_no_mv** / SCT_STRESS_CMD_NO_MV
--- a/platform_migration_test.py
+++ b/platform_migration_test.py
@@ -1,0 +1,448 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2026 ScyllaDB
+
+from collections import defaultdict
+
+from longevity_test import LongevityTest
+from sdcm.sct_events.system import InfoEvent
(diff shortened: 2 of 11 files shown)


docs/configuration_options.md
```diff
@@ -1275,6 +1284,15 @@ AMS AMI id to use for scylla db node
 **type:** str (appendable)
 
 
+## **ami_id_db_scylla_target** / SCT_AMI_ID_DB_SCYLLA_TARGET
```
fruch: if we are using the same exact version, maybe we should just look it up ? this complicate the way to trigger this on a regular cadence

platform_migration_test.py
```diff
@@ -0,0 +1,204 @@
...
+            self.log.info("Updated seed provider on all nodes")
+
+    def decommission_source_platform_nodes(self) -> None:
+        """Decommission all source platform nodes one by one"""
+        source_nodes = self.get_nodes_by_instance_type(self.source_instance_type)
+        if not source_nodes:
+            self.log.info("No source platform nodes to decommission")
+            return
+
+        for i, node in enumerate(source_nodes, 1):
```
fruch: see `Nemesis.decommission_nodes` we should strive to do it in parallel
even if scylla right no doesn't (it would on the next release), but from our end we should align as it supposes to be

test-cases/platform-migration/x86-to-arm-tablets.yaml
```diff
@@ -0,0 +1,18 @@
+test_duration: 60
+
+instance_type_db: 'i7i.2xlarge'
+instance_type_db_target: 'i8g.2xlarge'
+n_db_nodes: 3
+
+prepare_write_cmd: "cassandra-stress write cl=QUORUM n=1000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)' -mode cql3 native -rate threads=50 -pop seq=1..1000000"
```
fruch: we need to size it a bit more

disk of those is 1,875 GB, lets aim at two variants 10%, and 50% (later someone can add 90%)

so we might need more loaders to fill faster

fruch: some would take more then 45min :), one we would be run once in a release.
not on a regular cadence

sdcm/sct_config.py
```diff
@@ -1164,6 +1170,18 @@ class SCTConfiguration(dict):
...
+            env="SCT_STRESS_BEFORE_MIGRATION",
+            type=str,
+            help="Stress command to write data for post-migration validation",
+        ),
+        dict(
+            name="verify_stress_after_migration",
+            env="SCT_VERIFY_STRESS_AFTER_MIGRATION",
+            type=str,
+            help="Stress command to verify data after migration",
+        ),
```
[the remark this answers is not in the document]
dimakr: we don't define this for specific scenarios (like rolling upgrades, etc.) in test_defaults

fruch (dismissed):
LGTM

PR #13585, opened 2026-02-12
fix(integration-tests): fix vector-store integration tests

dimakr:
Fix vector-store service integration tests:
- at some point there were breaking changes in the index count API (was removed actually). The info provided by the removed API was moved to /status endpoint. Updated VS client in SCT to use the status API
- hot reload support was introduced for vector-store URI parameters in scylla config. Switched restarting scylla container to hot realoding the config
- updated Scylla image in vector-store tests to 2025.4.3, as `vector_primary_uri` scylla config parameter was changed `vector_store_primary_uri` in recent Scylla releases
- tablets is hard requirement for VS now, so added a check to ensure they are enabled in VS integration tests

Fixes: SCT-65

### Testing

- [x] integration tests

### Reminders

### DIFF
--- a/docker/scylla-sct/entry.sh
+++ b/docker/scylla-sct/entry.sh
@@ -16,7 +16,10 @@ authorizer: 'CassandraAuthorizer'
 EOM
 fi
 
+# vector indexes require tablets
+if [[ "${VECTOR_SEARCH_TEST:-}" != "true" ]]; then
 sed -e '/enable_tablets:.*/s/true/false/g' -i /etc/scylla/scylla.yaml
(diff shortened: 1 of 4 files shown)


dimakr:
i think a combination of the API endpoint and the fact that old scylla image was used, and newer releases of Scylla had vector-store related config parameter name changed. 
And I was under impression that test was suffering from race condition issue in the VS itself (it was actually, but only for some time..)
@pehala

fruch (approved):
LGTM

pehala (approved):

PR #13601, opened 2026-02-12
fix(nemesis): suppress raft topology errors in network block nemesis

yarongilor:
Wrapped network blocking operations in disrupt_network_block and _disrupt_network_block_k8s with ignore_raft_topology_cmd_failing() context manager to suppress expected raft topology connection closed errors during network disruption. This fixes DatabaseLogEvent ERROR: raft_topology - topology change coordinator fiber got error with connection is closed exceptions that occur when network connectivity is blocked during nemesis execution.

The network block nemesis intentionally disrupts network connectivity, which causes Raft consensus protocol to report connection failures. These errors are expected and correct behavior when a node is network- isolated, so they should be suppressed during the nemesis operation.

Fixes the error of:
```
2026-02-11 11:11:41.300 <2026-02-11 11:11:40.247>: (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=4f80e089-50e6-4f08-bf84-9f65307ea1da: type=RUNTIME_ERROR regex=std::runtime_error line_number=18953 node=longevity-200gb-48h-network-2026-1-db-node-76155e7b-1
2026-02-11T11:11:40.247 longevity-200gb-48h-network-2026-1-db-node-76155e7b-1  !ERR | scylla[6512]  [shard 0: gms] raft_topology - topology change coordinator fiber got error std::runtime_error (raft topology: exec_global_command(barrier) failed with seastar::rpc::closed_error (got error from node cbbe2fe6-1044-4894-919b-a13720bc3619/10.4.12.221: connection is closed))
```
in https://argus.scylladb.com/tests/scylla-cluster-tests/76155e7b-de05-4cdb-b819-c8e1b0bd5f64
### Testing

- [network-block-nemesis](https://argus.scylladb.com/tests/scylla-cluster-tests/54f3b13b-1982-462c-ac2c-b4f98ccfa533)

### Reminders

### DIFF
--- a/sdcm/nemesis.py
+++ b/sdcm/nemesis.py
@@ -3940,16 +3940,17 @@ class NemesisRunner:
     def _disrupt_network_block_k8s(self, list_of_timeout_options):
         duration = f"{random.choice(list_of_timeout_options)}s"
         experiment = NetworkPacketLossExperiment(self.target_node, duration, probability=100)
-        with DbNodeLogger(
-            self.cluster.nodes,
-            "block network traffic",
-            target_node=self.target_node,
-            additional_info=f"for {duration}sec",
-        ):
-            experiment.start()
-        experiment.wait_until_finished()
-        time.sleep(15)
-        self.cluster.wait_all_nodes_un()
+        with ignore_raft_topology_cmd_failing():
+            with DbNodeLogger(
+                self.cluster.nodes,
+                "block network traffic",
+                target_node=self.target_node,
+                additional_info=f"for {duration}sec",
+            ):
+                experiment.start()
+            experiment.wait_until_finished()
+            time.sleep(15)
+            self.cluster.wait_all_nodes_un()
 
     def disrupt_network_block(self):
         list_of_timeout_options = [10, 60, 120, 300, 500]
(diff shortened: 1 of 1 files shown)


timtimb0t (approved):

yarongilor:
> > How did we not encounter it before? I do not remember these nemesis failing constantly.
> > If we block the network these errors should be 100% right?
> 
> while barrier execution if I understand it correctly



@pehala , this nemesis never ran in a formal pipeline the last year:
image

pehala:
> @pehala , this nemesis never ran in a formal pipeline the last year:

How?

timtimb0t:
> > right, it's actually the opposite, i'm trying to see when the c-s load started failing and just opened this one on the way.
> 
> Right, but my question still remains, if ti started failing only now, I would like to know why and if it cannot be a bug.
> 
> The network block nemesis intentionally disrupts network connectivity, which causes Raft consensus protocol to report connection failures. These errors are expected and correct behavior when a node is network- isolated, so they should be suppressed during the nemesis operation. -> If this is correct, it should be previously as well, so what changed?

here is the issue where this error became to occur more often https://github.com/scylladb/scylladb/issues/28016
maybe barriers became execute more often, not sure. But this error MAY happen while node is down, but its not guaranteed

timtimb0t:
> > maybe barriers became execute more often, not sure. But this error MAY happen while node is down, but its not guaranteed
> 
> So nothing connected to network nemesis specifically, but given they block communiction they are treated like down and thus this can happen same as other cases?

This is definitely connected to network blocking. Whenever we block a node, reboot it, or shut it down, this error MAY occur. It happens when the coordinator has active work in progress at the moment a node becomes unavailable (i.e., it is not idle). If I understand correctly, this corresponds to barrier execution at that time.

pehala (approved):

PR #13604, opened 2026-02-13
[XCloud-VS] Verify Vector Search functionality during cluster scaling

mikliapko:
~~Proper testing blocked by siren issue https://github.com/scylladb/siren/issues/15463~~

Closes https://scylladb.atlassian.net/browse/VECTOR-525

## Description

This PR adds XCloud coverage for vector search during automatic cluster scale-out.

- Adds a new end-to-end test in `scylla_cloud_e2e_test.py` that:
  - creates vector-search schema and index,
  - drives disk usage to trigger XCloud scale-out,
  - runs continuous vector insert + ANN validation while resize is in progress,
  - fails the test if vector index validation fails during scaling.
- Adds test configuration in `test-cases/vector-search/vector-search-xcloud-scale-out.yaml` for the scale-out scenario.
- Adds Jenkins pipeline `jenkins-pipelines/vector-search/vector-search-in-xcloud-scale-out-aws.jenkinsfile` to run this scenario on XCloud AWS staging.

### Testing
- [x] [vector-search-in-xcloud-scale-out-aws](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/vector-search-in-xcloud-scale-out-aws/3)

### DIFF
--- a/scylla_cloud_e2e_test.py
+++ b/scylla_cloud_e2e_test.py
@@ -1,0 +1,330 @@
+#!/usr/bin/env python
+
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2026 ScyllaDB
+
+import random
+import threading
+import time
+import uuid
+from functools import cached_property
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from cassandra.cluster import Session
+
+from cassandra.query import BatchStatement
+
+from sdcm.cloud_api_client import ScyllaCloudAPIClient
+from sdcm.rest.remote_curl_client import RemoteCurlClient
+from sdcm.tester import ClusterTester
+from sdcm.utils.decorators import retrying, Retry
+from sdcm.utils.loader_utils import LoaderUtilsMixin
+
+
+class ScaleOutFailedException(Exception):
+    pass
+
+
+class _LoadThread(threading.Thread):
+    """Background thread that loads the cluster to trigger auto-scaling."""
+
+    def __init__(self, test: "ScyllaCloudVectorSearchXCloudTest"):
+        super().__init__()
+        self._test = test
+
+    def run(self):
+        stress_queue = []
+        self._test.assemble_and_run_all_stress_cmd(
+            stress_queue=stress_queue,
+            stress_cmd=self._test.params.get("stress_cmd"),
+            keyspace_num=1,
+        )
+        for queue in stress_queue:
+            self._test.verify_stress_thread(queue)
+
+        self._test.log.info("Write stress completed, starting background read stress")
+        stress_read_cmd = self._test.params.get("stress_read_cmd")
+        if stress_read_cmd:
+            # Fire-and-forget: keep read load running in the background to stress the cluster.
+            # We don't verify results here - any failures will surface as CassandraStressEvent
+            self._test.assemble_and_run_all_stress_cmd(
+                stress_queue=[],
+                stress_cmd=stress_read_cmd,
+                keyspace_num=1,
+            )
+
+
+class _VectorSearchThread(threading.Thread):
+    """Background thread that continuously inserts vectors and validates index updates."""
+
+    def __init__(self, test: "ScyllaCloudVectorSearchXCloudTest"):
+        super().__init__()
+        self._test = test
+        self._stop_event = threading.Event()
+        self._validation_failures = 0
+
+    @property
+    def validation_failures(self) -> int:
+        return self._validation_failures
+
+    @cached_property
+    def _vs_remote_curl_client(self) -> RemoteCurlClient:
+        """Get RemoteCurlClient that runs curl on the first VS node against its local VS API."""
+        vs_cluster = self._test.db_cluster.vector_store_cluster
+        assert vs_cluster and vs_cluster.nodes, "Vector Store cluster has no nodes"
+        vs_node = vs_cluster.nodes[0]
+        port = self._test.params.get("vector_store_port")
+        return RemoteCurlClient(host=f"localhost:{port}", endpoint="", node=vs_node)
+
+    def _log_index_status(self) -> None:
+        """Log the current status of the vector index via the VS node API."""
+        try:
+            result = self._vs_remote_curl_client.run_remoter_curl(
+                method="GET",
+                path=f"api/v1/indexes/{self._test.KEYSPACE_NAME}/{self._test.INDEX_NAME}/status",
+                params=None,
+            )
+            self._test.log.info(
+                "Vector index status for %s.%s: %s",
+                self._test.KEYSPACE_NAME,
+                self._test.INDEX_NAME,
+                result.stdout.strip(),
+            )
+        except Exception as exc:  # noqa: BLE001
+            self._test.log.warning("Failed to get vector index status: %s", exc)
+
+    @retrying(n=10, sleep_time=10, allowed_exceptions=(Retry,))
+    def _validate_index_update(self, vector_id: uuid.UUID, test_vector: list[float], session: "Session") -> None:
+        """Validate that a recently inserted vector is found in the ANN index.
+
+        Retries up to 10 times with 10s delay to allow for index propagation.
+        Raises Retry if the vector is not yet found.
+        """
+        self._test.log.debug("Checking the recently inserted vector made it into the index...")
+        self._log_index_status()
+        query_cql = f"""
+            SELECT id, commenter FROM {self._test.KEYSPACE_NAME}.{self._test.TABLE_NAME}
+            ORDER BY comment_vector ANN OF {test_vector} LIMIT 20
+        """
+        rows = session.execute(query_cql)
+        found = any(str(row.id) == str(vector_id) for row in rows)
+        if found:
+            self._test.log.debug("Successfully found vector %s in ANN query results", vector_id)
+            return
+        self._test.log.warning("Vector %s not found in ANN query results", vector_id)
+        raise Retry(f"Vector {vector_id} not found in ANN query results")
+
+    def stop_and_join(self, timeout: int = 300) -> None:
+        """Signal the thread to stop and wait for it to finish."""
+        if not self.is_alive():
+            return
+        self._test.log.info("Waiting for vector validation thread to complete...")
+        self._stop_event.set()
+        self.join(timeout=timeout)
+        if self.is_alive():
+            self._test.log.warning("Vector validation thread did not complete within timeout")
+
+    def run(self):
+        operation_number = 0
+
+        with self._test.db_cluster.cql_connection_patient(self._test.db_cluster.nodes[0]) as session:
+            insert_prepared = session.prepare(f"""
+                INSERT INTO {self._test.KEYSPACE_NAME}.{self._test.TABLE_NAME}
+                    (id, commenter, comment_vector, created_at)
+                VALUES (?, ?, ?, toTimestamp(now()))
+            """)
+
+            while not self._stop_event.is_set():
+                operation_number += 1
+                self._test.log.info("Vector search operation #%d", operation_number)
+
+                batch = BatchStatement()
+                vectors = []
+                for i in range(self._test.VECTORS_PER_BATCH):
+                    vector_id = uuid.uuid4()
+                    test_vector = [round(random.uniform(0.01, 0.99), 2) for _ in range(self._test.VECTOR_DIMENSION)]
+                    commenter = f"test_user_{operation_number}_{i}"
+                    batch.add(insert_prepared, (vector_id, commenter, test_vector))
+                    vectors.append((vector_id, test_vector))
+
+                sample_vector_id, sample_test_vector = random.choice(vectors)
+                try:
+                    session.execute(batch)
+                    self._test.log.info("Inserted batch of %d vectors", self._test.VECTORS_PER_BATCH)
+
+                    self._test.log.info("Validating randomly selected vector %s", sample_vector_id)
+                    self._validate_index_update(
+                        vector_id=sample_vector_id,
+                        test_vector=sample_test_vector,
+                        session=session,
+                    )
+
+                except Exception as exc:  # noqa: BLE001
+                    self._test.log.error("Error during vector operation #%d: %s", operation_number, exc)
+                    self._validation_failures += 1
+
+        self._test.log.info("Vector search operations stopped after %d operations", operation_number)
+
+
+class ScyllaCloudTestBase(ClusterTester, LoaderUtilsMixin):
+    """Base class for Scylla Cloud E2E tests, providing cloud API helpers."""
+
+    @cached_property
+    def cloud_api_client(self):
+        return ScyllaCloudAPIClient(
+            api_url=self.params.cloud_env_credentials["base_url"],
+            auth_token=self.params.cloud_env_credentials["api_token"],
+        )
+
+    @cached_property
+    def account_id(self):
+        return self.cloud_api_client.get_current_account_id()
+
+    @cached_property
+    def cluster_id(self):
+        return self.cloud_api_client.get_cluster_id_by_name(
+            account_id=self.account_id,
+            cluster_name=self.db_cluster.name,
+        )
+
+    def get_email_data(self):
+        self.log.info("Prepare data for email")
+        email_data = self._get_common_email_data()
+        return email_data
+
+
+class ScyllaCloudVectorSearchXCloudTest(ScyllaCloudTestBase):
+    """Verify that vector search operations remain functional during XCloud cluster auto-scaling."""
+
+    KEYSPACE_NAME = "vector_search_ks"
+    TABLE_NAME = "vector_search_table"
+    INDEX_NAME = "vector_search_index"
+    VECTOR_DIMENSION = 64
+    VECTORS_PER_BATCH = 100
+
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.load_thread = None
+        self.vector_thread = None
+
+    def tearDown(self) -> None:
+        """Clean up background threads."""
+        if self.load_thread and self.load_thread.is_alive():
+            self.log.info("Waiting for load thread to complete...")
+            self.load_thread.join(timeout=60)
+            if self.load_thread.is_alive():
+                self.log.warning("C-S load thread did not complete within timeout")
+
+        if self.vector_thread:
+            self.vector_thread.stop_and_join()
+
+        super().tearDown()
+
+    def prepare_vs_keyspace(self) -> None:
+        """Prepare vector search keyspace and table with index."""
+        self.create_keyspace(keyspace_name=self.KEYSPACE_NAME, replication_factor=3)
+
+        with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
+            create_table_cql = f"""
+                CREATE TABLE IF NOT EXISTS {self.KEYSPACE_NAME}.{self.TABLE_NAME} (
+                    id uuid,
+                    commenter text,
+                    comment_vector vector<float, {self.VECTOR_DIMENSION}>,
+                    created_at timestamp,
+                    PRIMARY KEY (id, created_at)
+                )
+            """
+            session.execute(create_table_cql)
+
+            create_index_cql = f"""
+                CREATE CUSTOM INDEX IF NOT EXISTS {self.INDEX_NAME}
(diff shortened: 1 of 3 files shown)


scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,250 @@
...
+        if self.load_thread and self.load_thread.is_alive():
+            self.log.info("Waiting for load thread to complete...")
+            self.load_thread.join()
+            if self.load_thread.is_alive():
+                self.log.warning("C-S load thread did not complete within timeout")
+
+        if self.vector_thread and self.vector_thread.is_alive():
+            self.log.info("Waiting for vector validation thread to complete...")
+            self.vector_stop_event.set()
+            self.vector_thread.join()
```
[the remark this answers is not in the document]
mikliapko: fixed

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,250 @@
...
+        self, target_utilization: int, check_interval: int = 30, wait_timeout: int = 3600
+    ) -> None:
+        """Wait until cluster disk utilization reaches the target percentage."""
+        cluster_info = self.cloud_api_client.get_cluster_details(
+            account_id=self.account_id,
+            cluster_id=self.cluster_id,
+            enriched=True,
+        )
+        # DB nodes are missing nodeType field, thus, can be filtered out this way
+        cluster_nodes = [node for node in cluster_info["nodes"] if not node.get("nodeType")]
```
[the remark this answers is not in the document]
mikliapko: it's already done this way, the comment is irrelevant

soyacz (commented):
besides the vs activity, would be worth to add some background stress with c-s as cluster is almost silent.
Or do the latte VS validation to have more load - currently at 120 reads.
Btw. monitoring shows no inserts after initial c-s. Bug in monitoring or we don't insert in this test.

mikliapko:
> Or do the latte VS validation to have more load - currently at 120 reads.

Could you please elaborate on this part?

soyacz:
> > Or do the latte VS validation to have more load - currently at 120 reads.
> 
> Could you please elaborate on this part?

I mean the load for this test is very low 120ops (reads). When we use SCT, it's better to stress Scylla more to uncover more issues

mikliapko:
> > > Or do the latte VS validation to have more load - currently at 120 reads.
> > 
> > 
> > Could you please elaborate on this part?
> 
> I mean the load for this test is very low 120ops (reads). When we use SCT, it's better to stress Scylla more to uncover more issues

Originally, my idea for this test is rather functional verification of VS in XCloud cluster while the cluster is scaling (the index is still being updated throughout the whole scaling procedure - once per ~40 seconds as CDC timeout rules it), not stress or performance testing. And we don't have a better place where to put such a test than SCT (to catch Scylla ERROR events, to have Argus reports, etc). 

To implement the same test logic in latte script seemed pretty complex to me - that's why I went with more classic approach where every validation is done as a test step.

At the same time, I understand your concern about low load. I could improve it and extend the test adding some read load during the test. I could also consider checking the vector index size instead of querying for making sure the new vector data got into index.

soyacz:
> > > > Or do the latte VS validation to have more load - currently at 120 reads.
> > > 
> > > 
> > > Could you please elaborate on this part?
> > 
> > 
> > I mean the load for this test is very low 120ops (reads). When we use SCT, it's better to stress Scylla more to uncover more issues
> 
> Originally, my idea for this test is rather functional verification of VS in XCloud cluster while the cluster is scaling (the index is still being updated throughout the whole scaling procedure - once per ~40 seconds as CDC timeout rules it), not stress or performance testing. And we don't have a better place where to put such a test than SCT (to catch Scylla ERROR events, to have Argus reports, etc).
> 

Ok, in that case, reduce loader size - no need for `c6i.2xlarge`
Also look at my previous comment about 'connection per each loop'

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,283 @@
...
+            while not stop_event.is_set():
+                operation_number += 1
+                self.log.info(f"Vector search operation #{operation_number}")
+
+                test_vector = [round(random.uniform(0.0, 1.0), 2) for _ in range(self.VECTOR_DIMENSION)]
+                vector_id = uuid.uuid4()
+                commenter = f"test_user_{operation_number}"
+
+                try:
+                    with self.db_cluster.cql_connection_patient(self.db_cluster.nodes[0]) as session:
```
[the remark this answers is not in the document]
mikliapko: Fixed

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+        self.log.info("Prepare Vector Search test data")
+        self.prepare_vs_keyspace()
+
+        self.log.info("Start loading cluster to scaling threshold in background")
+        self.load_thread = self.load_cluster_to_scaling_threshold_in_background()
+
+        self.log.info("Wait for 60% of disk utilization to be reached before starting VS operation")
+        self.wait_for_cluster_disk_utilization(target_utilization=60)
+
+        self.log.info("Start running Vector Search operations in background while cluster is scaling")
```
pehala, line 329: You do not verify that the cluster started scaling, only that the utilization was reached.

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+            cluster_name=self.db_cluster.name,
+        )
+
+    def get_email_data(self):
+        self.log.info("Prepare data for email")
+        email_data = self._get_common_email_data()
+        return email_data
+
+
+class ScyllaCloudVectorSearchXCloudTest(ScyllaCloudTestsMixin):
```
pehala, line 64: Please add documentation to every class/module you add. It will make ti easier in the future for others

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+from sdcm.tester import ClusterTester
+from sdcm.utils.decorators import retrying, Retry
+from sdcm.utils.loader_utils import LoaderUtilsMixin
+
+
+class ScaleOutFailedException(Exception):
+    pass
+
+
+class ScyllaCloudTestsMixin(ClusterTester, LoaderUtilsMixin):
```
pehala, line 39: Why do we need a separate mixin? Do we intent to have more tests?

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+                self.log.info("Stop event received during indexing wait")
+                return None
+
+            self.log.debug("Checking the recently inserted vector made it into the index...")
+            self.log_index_status()
+            query_cql = f"""
+                SELECT id, commenter FROM {self.KEYSPACE_NAME}.{self.TABLE_NAME}
+                ORDER BY comment_vector ANN OF {test_vector} LIMIT 20
+            """
+            rows = list(session.execute(query_cql))
```
pehala, line 215: nit: do we need a list here? cannot we iterate with the type that the query returns?

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+                            self.log.error("Index update validation failed for vector %s", sample_vector_id)
+                            self.vector_validation_failures_count += 1
+
+                    except Exception as exc:  # noqa: BLE001
+                        self.log.error("Error during vector operation #%d: %s", operation_number, exc)
+                        self.vector_validation_failures_count += 1
+
+            self.log.info("Vector search operations stopped after %d operations", operation_number)
+
+        vector_thread = threading.Thread(target=_run_vector_ops_in_background)
```
pehala, line 273: The thread code (=inner function) is longer thna the execution of it, maybe it is worht extracting into a Thread subclass, so it can be reused?

mikliapko:
> > > > > Or do the latte VS validation to have more load - currently at 120 reads.
> > > > 
> > > > 
> > > > Could you please elaborate on this part?
> > > 
> > > 
> > > I mean the load for this test is very low 120ops (reads). When we use SCT, it's better to stress Scylla more to uncover more issues
> > 
> > 
> > Originally, my idea for this test is rather functional verification of VS in XCloud cluster while the cluster is scaling (the index is still being updated throughout the whole scaling procedure - once per ~40 seconds as CDC timeout rules it), not stress or performance testing. And we don't have a better place where to put such a test than SCT (to catch Scylla ERROR events, to have Argus reports, etc).
> 
> Ok, in that case, reduce loader size - no need for `c6i.2xlarge` Also look at my previous comment about 'connection per each loop'

Fixed:
- reduced loader size to default loader type `c6i.xlarge`
- fixed connection per each loop issue - now it's a one session across the whole test
- added some read stress happening during scale out process
- added vector index status logging for informativeness and better debugging

Retesting: 
- [ ] https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/vector-search-in-xcloud-scale-out-aws/11/

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+
+            current_disk_utilization = (storage_used / total_storage) * 100
+            if current_disk_utilization >= target_utilization:
+                self.log.info("Target disk utilization of %d%% reached", target_utilization)
+                return
+
+            self.log.debug("Current disk utilization: %.2f%%, waiting...", current_disk_utilization)
+            time.sleep(check_interval)
+
+        raise TimeoutError(f"Disk utilization did not reach {target_utilization}% within {wait_timeout}s")
```
pehala, line 171: doesnt the retrying decorator do exactly just that? Do we need a separate timeout loop?

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+        or until the vector is found in the ANN query results.
+
+        Returns:
+            True if the vector was found, False if not found after all retries,
+            None if the stop event was received and the operation was aborted.
+        """
+        retries: int = 5
+        delay: int = 45
+
+        for attempt in range(1, retries + 1):
```
pehala, line 203: Could retrying be used here? it seems  to be doing similar thing

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+
+    @retrying(n=30, sleep_time=60, allowed_exceptions=(Retry,))
+    def wait_for_resize_request(self) -> int:
+        requests = self.cloud_api_client.get_cluster_requests(account_id=self.account_id, cluster_id=self.cluster_id)
+        resize_request = next((r for r in requests if r["requestType"] == "RESIZE_CLUSTER_V3"), None)
+        if resize_request and resize_request["status"] == "IN_PROGRESS":
+            return resize_request["id"]
+        raise Retry("Resize request not found")
+
+    @retrying(n=200, sleep_time=60, allowed_exceptions=(AssertionError,))
```
pehala, line 287: Thats a long timeout for a scale out (200mins) do we need it that big?

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+
+            self.log.info("Vector search operations stopped after %d operations", operation_number)
+
+        vector_thread = threading.Thread(target=_run_vector_ops_in_background)
+        vector_thread.start()
+        self.log.info("Vector search operations started in background thread")
+
+        return vector_thread
+
+    @retrying(n=30, sleep_time=60, allowed_exceptions=(Retry,))
```
pehala, line 279: Iisnt 30mins too much for a timeout for a resize request?

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+                            stop_event=stop_event,
+                            session=session,
+                        )
+                        if validation_result is False:
+                            self.log.error("Index update validation failed for vector %s", sample_vector_id)
+                            self.vector_validation_failures_count += 1
+
+                    except Exception as exc:  # noqa: BLE001
+                        self.log.error("Error during vector operation #%d: %s", operation_number, exc)
+                        self.vector_validation_failures_count += 1
```
pehala, line 269: I think this is wrong way how to do it IMO. If thread needs to report soething, I think there should be a better mechanism than using inner thread code + object variables. How about doing a subclass of Thread and adding method for getting the failures, which we can use afterwards to get the count. IT will mean everything will be local to the Thread and we can query the result. I think it would be much cleaner

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def tearDown(self) -> None:
+        """Clean up background threads and stop events."""
+        if self.load_thread and self.load_thread.is_alive():
+            self.log.info("Waiting for load thread to complete...")
+            self.load_thread.join(timeout=60)
+            if self.load_thread.is_alive():
+                self.log.warning("C-S load thread did not complete within timeout")
+
+        self.stop_and_join_vector_thread()
```
pehala, line 314: you are doing it in the test, why do you do it in the teardown as well?.

Also the event is created in the test, but used in teardown. I think cleaner way si to create the event in setup if it needs to be useed in teardown

scylla_cloud_e2e_test.py
mikliapko: We should start running VS operations a bit in advance before the scaling process to kick-in to fully cover the scaling process in a test.
Status request is checked in the next steps ("Wait for cluster to scale out")

scylla_cloud_e2e_test.py
mikliapko: My intention is that we may add more tests for Cloud clusters implemented in the same functional manner lately

scylla_cloud_e2e_test.py
mikliapko: we have the stop_event integration here what makes a custom loop necessary

scylla_cloud_e2e_test.py
mikliapko: So, based on all my previous runs of this test scale-out took ~2 hours. 
200 mins set here to be on safer side. We could safely reduce it to 180 mins but won't go further. Moreover, the test main purpose is to validate VS for XCloud clusters, rather than scaling process.

scylla_cloud_e2e_test.py
mikliapko: We start waiting for resize at 60% of disk utilization while actual scale out starts at 80% (threshold set in configuration) + 5% (as per design in Cloud)

scylla_cloud_e2e_test.py
mikliapko: if the test fails before reaching `wait_for_resize_request times out`, `tearDown` ensures the background thread is cleaned up. The guard `if not self.vector_thread or not self.vector_thread.is_alive(): return` makes the second call a no-op in the happy path. The duplication here is intentional safety.

scylla_cloud_e2e_test.py
pehala: Why not start waiting later then?

scylla_cloud_e2e_test.py
pehala: This seems fragile to me. Stop should be performed always once, we should not have two stop paths. Cannot we have it only in the teardown then? the check after the stopping in the test is simple so it wont keep the thread alive much longer.

Also we can then remove the defensivness from the function

scylla_cloud_e2e_test.py
pehala: Ok then it makes sense. Might I suggest creating a module in the future? It will make it easier to organize all Scylla cloud tests and put all of the util logic into __init__.py or other shared infrastructure file, which will make the files cleaner and hierarchy much simpler. I would not like this to end up like manager tests, which are 6+ files on the same level with intertwined imports due to some util classes living in one of the files.

It is not what we are currently doing for the tests, which is why this is just a suggestion, you can clearly ignore, but I think ti is worth thinking about when starting new set of tests, as it is superior IMO.

scylla_cloud_e2e_test.py
mikliapko: In this case I'd need to move this peace of code into a separate function adding to complexity without improving readability in any way. I'd like to keep the current approach.

scylla_cloud_e2e_test.py
pehala: That is interesting, I do not see why this would have to be there. I think the fact this is actuallyfunction that should live in the custom  thread class (which I suggested in a few places already), it makes it hard to understand.

scylla_cloud_e2e_test.py
pehala: I think one of the things this test can verify is the time it takes for scaling request to appear. I think that is a valid testcase and could be easily made part of this sanity test

scylla_cloud_e2e_test.py
pehala: `<opinion>`
In general, I think we also not be overly defensive for no clear reason. It might hide issues and produces more complex code. We wrote the code so we should be certain what needs to happen where, if we are not, the code is probably wrong.
Defensiveness should come only after we know there a potential problem and we need to ensure it works.
Also I think, if you find yourself to have to write too much of defensive code, it indicates the rest of the code is not up to snuff.
`</opinion>`

scylla_cloud_e2e_test.py
mikliapko: I want to have some vectors be included into the index at the moment when scale out kicks in.
I could start waiting at 70% of utilization - that's also fine IMO, but not more

scylla_cloud_e2e_test.py
mikliapko: make sense, I'll try to address it

scylla_cloud_e2e_test.py
mikliapko: okay, having the vector thread as a separate object would allow us to check if there are any validation issues in the test body while moving thread stop into teardown

scylla_cloud_e2e_test.py
mikliapko: fixed

scylla_cloud_e2e_test.py
mikliapko: fixed

scylla_cloud_e2e_test.py
mikliapko: fixed

scylla_cloud_e2e_test.py
mikliapko: fixed

scylla_cloud_e2e_test.py
mikliapko: fixed

scylla_cloud_e2e_test.py
mikliapko: fixed

scylla_cloud_e2e_test.py
mikliapko: fixed

mikliapko:
@pehala @soyacz I've implemented the changes you requested. 
Could you please revisit the PR?

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,330 @@
...
+            cluster_name=self.db_cluster.name,
+        )
+
+    def get_email_data(self):
+        self.log.info("Prepare data for email")
+        email_data = self._get_common_email_data()
+        return email_data
+
+
+class ScyllaCloudVectorSearchXCloudTest(ScyllaCloudTestBase):
```
pehala, line 207: nit: Mouthful, how about `VectorSearchXCloudTest` instead? Or `XCloudVectorSearchTest`

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,330 @@
...
+
+
+class _VectorSearchThread(threading.Thread):
+    """Background thread that continuously inserts vectors and validates index updates."""
+
+    def __init__(self, test: "ScyllaCloudVectorSearchXCloudTest"):
+        super().__init__()
+        self._test = test
+        self._stop_event = threading.Event()
+        self._validation_failures = 0
```
pehala, line 75: nit: I do not think we need to do protected variables here, it just makes reading harder

scylla_cloud_e2e_test.py
```diff
@@ -0,0 +1,330 @@
...
+            # Fire-and-forget: keep read load running in the background to stress the cluster.
+            # We don't verify results here - any failures will surface as CassandraStressEvent
+            self._test.assemble_and_run_all_stress_cmd(
+                stress_queue=[],
+                stress_cmd=stress_read_cmd,
+                keyspace_num=1,
+            )
+
+
+class _VectorSearchThread(threading.Thread):
```
pehala, line 68: nit: I do not think we need to have protected classes. We actually want for these to be used in the future if needed to be

pehala (approved):
Just small naming complaints, otherwise LGTM

soyacz (approved):
LGTM

scylla_cloud_e2e_test.py
mikliapko: Actually, the idea was to emphasis Scylla Cloud deployment type, but I'd rather agree with you that name simplification is needed here. I can fix it as follow-up PR since this one is already merged.

scylla_cloud_e2e_test.py
mikliapko: I'd agree about `self._test` but the rest is a reasonable convention here since `_stop_event` shouldn't be accessed directly but via `stop_and_join()` as well as `_validation_failures` has a public property

scylla_cloud_e2e_test.py
mikliapko: make sense

PR #13662, opened 2026-02-18
fix(remoter): prevent stale SSH connection reuse due to Python id() recycling

juliayakovlev:
Fixed a race condition where a new remoter could inherit a stale SSH connection from a previously garbage-collected remoter. This caused SSH commands intended for one node to be executed on a different node during parallel node creation.

Root cause: connection_thread_map uses str(id(self)) as key to store SSH connections. When a remoter is garbage collected, Python may reuse its id() for a new object. Without cleanup, the new remoter would find and reuse the stale connection from the previous remoter.

Changes:
- Modified stop() to remove connection from connection_thread_map
- Added _remove_connection_from_thread_map() method for cleanup
- Fixed _close_connection() to not trigger connection recreation
- Added refresh of cached network_configuration in AWS node
- Added unit tests to verify the connection cleanup behavior

Fixes: https://scylladb.atlassian.net/browse/SCT-97

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/remote/remote_base.py
+++ b/sdcm/remote/remote_base.py
@@ -165,10 +165,23 @@ class RemoteCmdRunnerBase(CommandRunner, RetryMixin):
 
     def stop(self):
         self._close_connection()
+        # Clean up connection from thread map to prevent id() reuse issues
+        # When a remoter is garbage collected, Python may reuse its id() for a new object.
+        # Without cleanup, the new object would inherit the stale connection.
+        self._remove_connection_from_thread_map()
+
+    def _remove_connection_from_thread_map(self):
+        """Remove connection from thread map to prevent stale connection reuse when id() is recycled."""
+        key = str(id(self))
+        if hasattr(self.connection_thread_map, key):
+            delattr(self.connection_thread_map, key)
 
     def _close_connection(self):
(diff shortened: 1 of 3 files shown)


sdcm/remote/remote_base.py
```diff
@@ -165,10 +165,23 @@ def _is_connection_generation_ok(self, connection: object):
 
     def stop(self):
         self._close_connection()
+        # Clean up connection from thread map to prevent id() reuse issues
```
fruch: This part sounds relevant, but not something we can easily reproduce.

So it's hard to test

cezarmoise (approved):

fruch:
@juliayakovlev I've put some of the provision test to try it out, also do 4-5h longevity with this change to be on the safe side.

fruch:
this thing actually explain multiple bugs we've seen during the years of using this implementation (and also when we used fabric we have something similar)

fruch:
Provision tests passed 

Since it quite rare, not gonna backport it for now

soyacz:
> Provision tests passed
> 
> Since it quite rare, not gonna backport it for now

I'd consider backporting to 2026.1 which will be with us for next 2 y

PR #13663, opened 2026-02-18
fix(auto-ssh): update autossh image and environment variables

fruch:
Updated autossh container configuration to use version 2.1.0 and adapted environment variables to match new naming convention. Added support for localhost connections via host.docker.internal gateway and configured SSH options for host key checking and authentication.

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/auto_ssh.py
+++ b/sdcm/utils/auto_ssh.py
@@ -15,7 +15,7 @@ import os
 from functools import cached_property
 
 
-AUTO_SSH_IMAGE = "jnovack/autossh:1.2.2"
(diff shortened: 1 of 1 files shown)


dimakr:
> @dimakr this is an update we needed long ago, and I think it help us to try doing no auth, that we need.
> 
> I didn't yet test it with no auth, if it's working as expected

actually we are supporting already streaming logs from scylla cloud node, for some time, for cases with VPC peering (it was working since i added static port for vector.dev..)
We only need to find solution for local dev env.

cezarmoise (approved):

PR #13664, opened 2026-02-18
fix(vector.dev): use dynamic port range for vector logging

dimakr:
Replace the static vector.dev container port 15000 with a range 15000-15099. This helps to prevent conflicts when multiple tests share a Jenkins worker.

Fixes: SCT-99

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/provision/security.py
+++ b/sdcm/provision/security.py
@@ -33,5 +33,5 @@ class ScyllaOpenPorts(Enum):
     MANAGER_HTTPS = 5443  # Scylla Manager HTTPS API
     MANAGER_AGENT_PROMETHEUS = 5090  # Scylla Manager Agent Prometheus API
     MANAGER_DEBUG = 5112  # Scylla Manager pprof Debug
-    VECTOR = 15000  # Vector log transport
+    VECTOR = "15000-15099"  # Vector log transport port range
     SCT_AGENT = 16000  # SCT agent HTTP API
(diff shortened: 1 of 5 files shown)


fruch:
@dimakr so to recap, we hardcoded this port, to enable local development for xcloud, right ? (cause we couldn't use autossh), or I missing/forgot something ?

so before we do this one, I prefer we try fixing the autossh, and then revert the hard code port, and not do this range thing, which won't be enough.

regardless we should still handle better test cleanups on builders

dimakr:
> @dimakr so to recap, we hardcoded this port, to enable local development for xcloud, right ? (cause we couldn't use autossh), or I missing/forgot something ?
> so before we do this one, I prefer we try fixing the autossh, and then revert the hard code port, and not do this range thing, which won't be enough.
> regardless we should still handle better test cleanups on builders

This was a prep for enabling logs streaming from cloud nodes, particularly for the case when running on local dev machine as SCT runner. But after our discussion I asked around and search through SDM docs again - on admin guide they explicitly say that remote forwarding is not supported:
`At this time remote forwarding is not supported` (https://docs.strongdm.com/admin/resources/servers/port-forwarding). 
So at the end we probably do not need these changes around vector.dev port - making it static (or known from range, in advance) - as with VPC peering all ports are open in the SCT test sec. group anyway (just need to check this on real run)
@fruch

dimakr:
SO we don't need this change and can revert PR that set static port for vector.dev.

For local dev flow we cannot use reverse tunneling anyway (SDM rejects it), so we can probably adjust the xcloud backend to use ssh logs_transport in that specific case (public communication is enabled and VPC peering is disabled).

PR #13982, opened 2026-03-11
docs(plans): resilient test config dependencies

fruch:
## Summary

- Implementation plan for preventing silent config dependency failures in SCT test configurations
- Triggered by commit `866d18e70` which extracted auth settings from 53 YAML configs into `configurations/auth_cassandra.yaml`, breaking `rolling-upgrade-with-sla-no-shares` due to Jenkins parameter caching
- Proposes 5 options with execution plans:
  - **Option A**: Use anyconfig's built-in Jinja2 `{% include %}` (near-zero code change — already verified working)
  - **Option B**: Hardcode `test_config` in Jenkins environment
  - **Option C**: Config validation rules (fail fast at load time)
  - **Option D**: `requires` directive in YAML configs
  - **Option E**: Auto-include based on config content
- **Recommendation**: Option A + C (self-contained configs + defense-in-depth validation)

## Test plan

- [ ] Review plan for completeness and feasibility
- [ ] Decide on preferred option(s)
- [ ] Create follow-up implementation PRs

### DIFF
--- a/docs/plans/resilient-test-config-dependencies.md
+++ b/docs/plans/resilient-test-config-dependencies.md
@@ -41,6 +41,100 @@ Tested against SCT configs:
 - `defaults/test_default.yaml` (184 keys) — works
 - **No existing SCT config files contain `{{` or `{%` syntax** that would conflict with Jinja2 template processing
 
+### Real-World Composition Patterns in SCT Jenkinsfiles
+
+SCT Jenkins pipelines compose configs in two fundamentally different ways. Option A (includes) only fits one of them.
+
+#### Pattern 1: Hard Dependencies — Include Fits
+
+Some configs are always paired across every Jenkinsfile that uses a given test case. These are true dependencies — the test cannot run without them.
+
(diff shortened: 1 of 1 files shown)


pehala (commented):
Could we analyze and include other similar dependencies as part of this plan or will it be a followup? auth_cassandra is for sure not the only one

fruch:
> Could we analyze and include other similar dependencies as part of this plan or will it be a followup? auth_cassandra is for sure not the only one

probably not the only case, but that that case that drove me to try scoping it out.

fruch:
> > Could we analyze and include other similar dependencies as part of this plan or will it be a followup? auth_cassandra is for sure not the only one
> 
> probably not the only case, but that that case that drove me to try scoping it out.

now it covers all other findings that might benefit from such options

docs/plans/resilient-test-config-dependencies.md
```diff
@@ -0,0 +1,862 @@
+---
+status: draft
+domain: config
+created: 2026-03-11
+last_updated: 2026-04-12
+owner: null
```
fruch, line 6: needs to put me as owner

PR #14110, opened 2026-03-18
feat(unit_tests): Introduce fake_events.py module

pehala:
Resubmit of https://github.com/scylladb/scylla-cluster-tests/pull/13966
### Overview

The existing events system (events_lib) is slow, it sets up threads and processes and has sleeps (~10sec per module). It is needlessly complex for majority of the tests. Introduce a new events module which can be used instead and is much simpler and instantaneous.

### Testing

* [x] locally
* [ ] unit tests
* [ ] integration tests


fruch (approved):
LGTM

PR #14165, opened 2026-03-22
feature(cassandra): add Cassandra cluster support for Docker and AWS backends

fruch:
## Summary

Stacked on #14164. Implements Phase 2+3 of the [Cassandra Cluster Support plan](docs/plans/infrastructure/cassandra-cluster-support.md).

### Core framework (`sdcm/cluster_cassandra.py`)
- `BaseCassandraCluster` mixin — backend-agnostic Cassandra cluster logic: JDK install, apt repo setup, `cassandra.yaml` config, JVM heap sizing, seed management, NVMe data device setup
- `CassandraNodeMixin` — overrides Scylla-specific `BaseNode` properties (`smp`, `cpuset`, `nodetool`, `raft`, etc.) with Cassandra-appropriate behavior
- `compute_jvm_heap_mb()` — JVM heap calculator with 31 GB compressed-oops cap

### Backend integration
- **Docker**: `CassandraDockerCluster`/`CassandraDockerNode` inherit from `BaseCassandraCluster` mixin (refactored from `BaseScyllaCluster`)
- **AWS**: New `CassandraAWSCluster`/`CassandraAWSNode` with apt-based install on Ubuntu AMI, NVMe instance storage mount, and proper node health checks
- **Tester routing**: `db_type: "mixed_cassandra"` in `get_cluster_aws()` for Gemini oracle testing

### Monitoring
- `CassandraExporterSetup` — installs Criteo cassandra_exporter on nodes
- Prometheus scrape targets for Cassandra exporter
- SCT Cassandra Overview Grafana dashboard (latency, throughput, JVM heap, compaction, cache, GC)

### Log collection
- `CassandraLogCollector` — collects Cassandra-specific logs (`system.log`, `debug.log`, `gc.log`, `cassandra.yaml`, nodetool diagnostics)
- Integrated into both tester teardown and `sct.py collect-logs` CLI

### Configuration
| Parameter | Type | Default | Description |
|-----------|------|---------|-------------|
| `cassandra_oracle_version` | str | `"5.0"` | Cassandra version for oracle cluster |
| `ami_id_db_cassandra_oracle` | str | `""` | AMI ID for Cassandra oracle nodes |
| `install_cassandra_exporter` | bool | `true` | Install Prometheus exporter on Cassandra nodes |

### Other changes
- Widen `GeminiStressThread.oracle_cluster` type to `BaseCluster` (removes AWS-specific imports)
- Update all 3 KMS skip checks for `mixed_cassandra`
- Explicit MRO conflict resolution on Cassandra node/cluster classes

## Test plan

- [x] Unit tests: `compute_jvm_heap_mb`, `BaseCassandraCluster`, `CassandraDockerCluster` env vars (pytest-style)
- [x] Existing Phase 1 Docker test still passes after mixin refactor
- [x] Pre-commit passes (1576 tests, 0 failures)
- [ ] Run Gemini test with `mixed_cassandra` on Docker backend
- [ ] Run Gemini test with `mixed_cassandra` on AWS backend (validates CassandraAWSCluster node_setup)
- [ ] Verify legacy `db_type: cassandra` and `db_type: mixed` still work on AWS
- [ ] 🕥 https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/cassandra-preload-500gb-latte-test/18/

### DIFF
--- a/sdcm/cluster_aws.py
+++ b/sdcm/cluster_aws.py
@@ -13,12 +13,9 @@
 
 import json
 import logging
-import os
 import random
 import re
-import tempfile
 import time
-import uuid
 from collections.abc import Callable
 from contextlib import ExitStack
 from datetime import datetime
@@ -33,6 +30,7 @@ from mypy_boto3_ec2 import EC2Client
 from mypy_boto3_ec2.service_resource import EC2ServiceResource
 
 from sdcm import ec2_client, cluster, wait
+from sdcm.cluster_cassandra import BaseCassandraCluster, CassandraNodeMixin
 from sdcm.ec2_client import CreateSpotInstancesError
 from sdcm.provision.aws.utils import configure_set_preserve_hostname_script
 from sdcm.provision.common.utils import configure_hosts_set_hostname_script
@@ -1207,7 +1205,37 @@ class ScyllaAWSCluster(cluster.BaseScyllaCluster, AWSCluster):
         super().destroy()
 
 
-class CassandraAWSCluster(ScyllaAWSCluster):
+class CassandraAWSNode(CassandraNodeMixin, AWSNode):
+    """AWS node running Cassandra instead of Scylla.
+
+    MRO puts CassandraNodeMixin before AWSNode/BaseNode, so the mixin's
+    overrides win.  The explicit forwarding below makes that choice visible
+    to static-analysis tools and prevents silent regressions if the
+    inheritance order ever changes.
+    """
+
+    # --- Explicit MRO forwarding to CassandraNodeMixin ---
+    remote_scylla_yaml = CassandraNodeMixin.remote_scylla_yaml
+    smp = CassandraNodeMixin.smp
+    cpuset = CassandraNodeMixin.cpuset
+    _gen_nodetool_cmd = CassandraNodeMixin._gen_nodetool_cmd
+    extract_seeds_from_scylla_yaml = CassandraNodeMixin.extract_seeds_from_scylla_yaml
+    is_server_encrypt = CassandraNodeMixin.is_server_encrypt
+    raft = CassandraNodeMixin.raft
+    cql_address = CassandraNodeMixin.cql_address
+    wait_native_transport = CassandraNodeMixin.wait_native_transport
+    db_up = CassandraNodeMixin.db_up
+    config_setup = CassandraNodeMixin.config_setup
+    get_scylla_binary_version = CassandraNodeMixin.get_scylla_binary_version
+
+
+class CassandraAWSCluster(BaseCassandraCluster, AWSCluster):
+    """Cassandra cluster provisioned on AWS EC2.
+
+    Installs Cassandra during node_setup() via the official apt repository.
+    Uses a standard Ubuntu base AMI (no custom Cassandra-specific AMI needed).
+    """
+
     def __init__(
         self,
         ec2_ami_id,
@@ -1221,17 +1249,11 @@ class CassandraAWSCluster(ScyllaAWSCluster):
         user_prefix=None,
         n_nodes=3,
         params=None,
+        node_type: str = "cs-db",
     ):
-        if ec2_block_device_mappings is None:
-            ec2_block_device_mappings = []
-        # We have to pass the cluster name in advance in user_data
-        cluster_uuid = uuid.uuid4()
+        cluster_uuid = self.test_config.test_id()
         cluster_prefix = cluster.prepend_user_prefix(user_prefix, "cs-db-cluster")
         node_prefix = cluster.prepend_user_prefix(user_prefix, "cs-db-node")
-        node_type = "cs-db"
-        shortid = str(cluster_uuid)[:8]
-        name = "%s-%s" % (cluster_prefix, shortid)
-        user_data = "--clustername %s --totalnodes %s --version community --release 2.1.15" % (name, sum(n_nodes))
 
         super().__init__(
             ec2_ami_id=ec2_ami_id,
@@ -1239,7 +1261,6 @@ class CassandraAWSCluster(ScyllaAWSCluster):
             ec2_security_group_ids=ec2_security_group_ids,
             ec2_instance_type=ec2_instance_type,
             ec2_ami_username=ec2_ami_username,
-            ec2_user_data=user_data,
             ec2_block_device_mappings=ec2_block_device_mappings,
             cluster_uuid=cluster_uuid,
             services=services,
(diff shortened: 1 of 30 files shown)


sdcm/cluster_aws.py
```diff
@@ -1197,7 +1195,17 @@ def destroy(self):
         super().destroy()
 
 
-class CassandraAWSCluster(ScyllaAWSCluster):
+class CassandraAWSNode(CassandraNodeMixin, AWSNode):
+    """AWS node running Cassandra instead of Scylla."""
+
+
+class CassandraAWSCluster(BaseCassandraCluster, AWSCluster):
```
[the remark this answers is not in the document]
fruch: Fixed: Added explicit `node_setup`, `node_startup`, and `get_node_ips_param` forwarding methods to `CassandraAWSCluster` that delegate to `super()` (resolves to `BaseCassandraCluster` via MRO).

test-cases/cassandra/cassandra-preload-1tb-latte.yaml
```diff
@@ -0,0 +1,51 @@
+test_duration: 960
+n_db_nodes: 3
+n_loaders: 2
+instance_type_db: 'i4i.4xlarge'
+root_disk_size_db: 100
+instance_type_loader: 'c6i.4xlarge'
+
+db_type: cassandra
+cassandra_version: '4.1'
+ami_id_db_cassandra: 'resolve:ssm:/aws/service/canonical/ubuntu/server/22.04/stable/current/amd64/hvm/ebs-gp2/ami-id'
```
fruch, line 10: we should as followup move to 24.04, this is a mistake

fruch: actully not a mistake,

```
check_node_db_up was stuck permanently returning False — not because Cassandra was unhealthy, but because the CQL readiness check used cqlsh, which crashes on Python 3.12 (Ubuntu 24.04):
```

Cassandra cqlsh doesn't work with python 3.12 out of the box (maybe 5.0 would, but not sure)

sdcm/sct_provision/aws/cluster.py
```diff
@@ -379,7 +390,7 @@ def _user_data(self) -> str:
         return ScyllaUserDataBuilder(
             params=self.params,
             cluster_name=self.cluster_name,
-            user_data_format_version=self.params.get("oracle_user_data_format_version"),
+            user_data_format_version=self.params.get("oracle_user_data_format_version") or "3",
```
dimakr: This change would affect the regular Scylla cluster path, not only CS. Is it intentional?

sdcm/cluster_cassandra.py
```diff
@@ -0,0 +1,692 @@
...
+    def _configure_cassandra_yaml(self, node):
+        """Render and upload cassandra.yaml for the given node."""
+        seeds = ",".join(n.ip_address for n in self.nodes if n.is_seed) or node.ip_address
+        num_tokens = self.params.get("cassandra_num_tokens") or 16
+
+        yaml_updates = {
+            "cluster_name": self.name,
+            "num_tokens": num_tokens,
+            "seed_provider": [
+                {
```
dimakr, line 573: This value is overridden by sed anyway later in this method. Should we set any value in the list here?

sdcm/cluster_cassandra.py
```diff
@@ -0,0 +1,692 @@
...
+    Returns:
+        Tuple of (max_heap_mb, heap_new_mb).
+    """
+    if total_ram_mb < 2048:
+        max_heap = min(total_ram_mb // 2, 1024)
+    else:
+        max_heap = min(total_ram_mb // 4, 8192)
+
+    # Hard cap at 31 GB (compressed oops threshold)
+    max_heap = min(max_heap, 31 * 1024)
```
dimakr: This seems to be a dead cap. The max fro above is `min(total_ram_mb // 4, 8192)` i.e. max_heap value in the `min(max_heap, 31 * 1024)` cap will never be bigger than 8GB.

sdcm/sct_provision/aws/cluster.py
fruch: we should revert this change, and set it in the configuration as we need

sdcm/cluster_cassandra.py
fruch: those are the updated rules:
https://cassandra.apache.org/doc/stable/cassandra/managing/operating/hardware.html#memory

I this is based on older version of their docs

I'll update this

sdcm/cluster_cassandra.py
fruch: the sed is using this structure

since we don't have the equivalent of ScyllaYaml

not sure it work investing into building something similar,
but maybe it's not that complicated

test-cases/cassandra/cassandra-preload-1tb-latte.yaml
fruch: we now have a patch, but it probably safer to leave it as is for now

sdcm/cluster_aws.py
```diff
@@ -1251,63 +1272,48 @@ def __init__(
...
-                ec2_user_data = "--clustername %s --totalnodes %s --seeds %s --version community --release 2.1.15" % (
-                    self.name,
-                    count,
-                    seeds,
-                )
-        ec2_user_data = self.update_bootstrap(ec2_user_data, enable_auto_bootstrap)
-        added_nodes = super().add_nodes(
-            count=count,
-            ec2_user_data=ec2_user_data,
+    def _create_node(self, instance, ami_username, node_prefix, node_index, base_logdir, dc_idx, rack):
```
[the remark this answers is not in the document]
fruch: we need to rebase and handle this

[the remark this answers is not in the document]
fruch: Fixed — added `after_config=None` parameter to `CassandraAWSCluster._create_node` and forwarded it to the `CassandraAWSNode` constructor.

sdcm/cluster_cassandra.py
fruch: Fixed — removed the dead `min(max_heap, 31 * 1024)` cap. The heap is now simply 50% of RAM (min 2 GB when system ≥ 4 GB, 256 MB floor for tiny containers).

sdcm/sct_provision/aws/cluster.py
fruch: Reverted — this line is back to the upstream version (`self.params.get("oracle_user_data_format_version")` without the `or "3"` fallback). The default should be set in the configuration instead.

dimakr (approved):
LGTM

PR #14322, opened 2026-04-12
plan(provision): AWS capacity validation and automatic AZ/region fallback

fruch:
## Summary

- Implementation plan for upfront AZ filtering and automatic AZ/region fallback when AWS provisioning fails with `InsufficientInstanceCapacity`
- Covers both legacy (`tester.py`) and modern (`sct_provision/`) provisioning code paths
- Extends AZ fallback to ALL test types including multi-node (not just artifact tests)
- Complements PR #13317 (multi-cloud provisioning resilience for artifact tests) by focusing on AWS robustness first

**Plan location:** `docs/plans/infrastructure/aws-capacity-az-fallback.md`

### Key design decisions

1. **Upfront AZ filtering** (Phase 1-2): Before provisioning, use `describe_instance_type_offerings` to remove AZs that don't even offer the required instance types
2. **AZ fallback on capacity errors** (Phase 3-4): When provisioning fails, automatically try next AZ -- all cluster types (db, loader, monitor) move together
3. **Backend-agnostic config** (Phase 3): Rename `aws_fallback_to_next_availability_zone` to `fallback_to_next_availability_zone`, default to `true`
4. **Region fallback** (Phase 5, future): Optional, single-node only, when all AZs exhausted

### Phases

| Phase | Description | Importance |
|-------|-------------|------------|
| 1 | Centralize AZ discovery and capacity error detection | Critical |
| 2 | Upfront AZ filtering (pre-provisioning validation) | Critical |
| 3 | AZ fallback on capacity errors -- modern path | Critical |
| 4 | AZ fallback on capacity errors -- legacy path | High |
| 5 | Region fallback (future, single-node only) | Low |

## Test plan

- [x] Unit tests with mocked `describe_instance_type_offerings` for AZ filtering
- [x] Unit tests with mocked `ClientError` for capacity error detection
- [x] Unit tests for fallback candidate generation (single-AZ and multi-AZ)
- [x] Planning review -- no code changes to test

### DIFF
--- a/docs/plans/MASTER.md
+++ b/docs/plans/MASTER.md
@@ -31,6 +31,7 @@ For plan writing guidelines, see [INSTRUCTIONS.md](INSTRUCTIONS.md).
 | Cassandra Cluster Support | `draft` | [cassandra-cluster-support.md](infrastructure/cassandra-cluster-support.md) |
 | SSH Key Decoupling | `draft` | [ssh-key-decoupling.md](infrastructure/ssh-key-decoupling.md) |
 | Multi-Cloud Provisioning Resilience | `draft` | [multi-cloud-provisioning-resilience.md](infrastructure/multi-cloud-provisioning-resilience.md) |
+| AWS Capacity AZ Fallback | `draft` | [aws-capacity-az-fallback.md](infrastructure/aws-capacity-az-fallback.md) |
 
 ### Nemesis — Chaos engineering, disruptors
 
--- a/docs/plans/MASTER.md
+++ b/docs/plans/MASTER.md
@@ -30,6 +30,7 @@ For plan writing guidelines, see [INSTRUCTIONS.md](INSTRUCTIONS.md).
 | Source-Destination Clusters | `pending_pr` | [#13908](https://github.com/scylladb/scylla-cluster-tests/pull/13908) |
 | Cassandra Cluster Support | `draft` | [cassandra-cluster-support.md](infrastructure/cassandra-cluster-support.md) |
 | SSH Key Decoupling | `draft` | [ssh-key-decoupling.md](infrastructure/ssh-key-decoupling.md) |
+| AWS Capacity AZ Fallback | `draft` | [aws-capacity-az-fallback.md](infrastructure/aws-capacity-az-fallback.md) |
 
 ### Nemesis — Chaos engineering, disruptors
 
--- a/docs/plans/assets/progress-roadmap.svg
+++ b/docs/plans/assets/progress-roadmap.svg
@@ -1,10 +1,10 @@
 <svg xmlns="http://www.w3.org/2000/svg" width="840" height="884" viewBox="0 0 840 884">
 <rect width="840" height="884" fill="#0D1B2A" rx="8"/>
 <text x="420" y="60" text-anchor="middle" font-family="system-ui, -apple-system, sans-serif" font-size="22" font-weight="bold" fill="#00BCD4">SCT Implementation Plans</text>
-<text x="420" y="84" text-anchor="middle" font-family="system-ui, -apple-system, sans-serif" font-size="13" fill="#8899AA">27 plans across 8 domains</text>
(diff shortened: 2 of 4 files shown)


soyacz (commented):
another approach could be using Capacity Reservations - we already have it implemented.
Still it suffers from cross az (or in future, dc) communication costs between sct-runner and instances.
Besides, we need to make sure that node numbers also reset when provisioning fails.
And would be also good to provision db/loaders/monitor in parallel (if not already done).

fruch:
> another approach could be using Capacity Reservations - we already have it implemented. Still it suffers from cross az (or in future, dc) communication costs between sct-runner and instances. Besides, we need to make sure that node numbers also reset when provisioning fails. And would be also good to provision db/loaders/monitor in parallel (if not already done).

how it would work with cases we don't know how to calculate the extra nodes during test ? and cost might be problematic as well (we need to pay upfront for all of the resources we'll add during the test)

also it doesn't work for spot, so the shot cases we want to benefit from this, would need to be switch to on_demand...

soyacz:
> > another approach could be using Capacity Reservations - we already have it implemented. Still it suffers from cross az (or in future, dc) communication costs between sct-runner and instances. Besides, we need to make sure that node numbers also reset when provisioning fails. And would be also good to provision db/loaders/monitor in parallel (if not already done).
> 
> how it would work with cases we don't know how to calculate the extra nodes during test ? and cost might be problematic as well (we need to pay upfront for all of the resources we'll add during the test)

I was thinking about asking only for initial setup - not reserving capacity for resources added in the test.
We don't pay upfront, after cancellation we are not charged. I think we must pay for at least 1minute.

> 
> also it doesn't work for spot, so the shot cases we want to benefit from this, would need to be switch to on_demand...

Ok, so idea is to also fallback spots to multiple AZ's ?

fruch:
> > > another approach could be using Capacity Reservations - we already have it implemented. Still it suffers from cross az (or in future, dc) communication costs between sct-runner and instances. Besides, we need to make sure that node numbers also reset when provisioning fails. And would be also good to provision db/loaders/monitor in parallel (if not already done).
> > 
> > 
> > how it would work with cases we don't know how to calculate the extra nodes during test ? and cost might be problematic as well (we need to pay upfront for all of the resources we'll add during the test)
> 
> I was thinking about asking only for initial setup - not reserving capacity for resources added in the test. We don't pay upfront, after cancellation we are not charged. I think we must pay for at least 1minute.

I'm not sure doing capacity reservation and canceling doesn't guarantee you'll get resources on regular flow, we have zero idea how those things works behind the scenes   


> 
> > also it doesn't work for spot, so the shot cases we want to benefit from this, would need to be switch to on_demand...
> 
> Ok, so idea is to also fallback spots to multiple AZ's ?

this direction is gonna work for both, i.e. we provision until failure, and move to new AZ, and then to new Region

pehala:
> this direction is gonna work for both, i.e. we provision until failure, and move to new AZ, and then to new Region

I agree in general, I would avoid move to new Region, as it will bring complexity if direct AMI is provided. I would rather keep it as move to new AZ only

fruch:
> > this direction is gonna work for both, i.e. we provision until failure, and move to new AZ, and then to new Region
> 
> I agree in general, I would avoid move to new Region, as it will bring complexity if direct AMI is provided. I would rather keep it as move to new AZ only

it's correct doing region complicate things, but we do have way to find the correct AMIs (even check that upfront), anyhow that the last optional stage of this plan.

fruch:
> > > this direction is gonna work for both, i.e. we provision until failure, and move to new AZ, and then to new Region
> > 
> > 
> > I agree in general, I would avoid move to new Region, as it will bring complexity if direct AMI is provided. I would rather keep it as move to new AZ only
> 
> it's correct doing region complicate things, but we do have way to find the correct AMIs (even check that upfront), anyhow that the last optional stage of this plan.

ontop of it the logic and flow would match in other cloud providers, regardless of the reversed capacity features in each one

PR #14341, opened 2026-04-13
fix(events): add missing yield to ignore_aborted_snapshot_upload_storage_io_errors

mikliapko:
Relates to https://github.com/scylladb/scylla-cluster-tests/pull/14341

## Summary

- `ignore_aborted_snapshot_upload_storage_io_errors` is decorated with `@contextmanager` but was missing a `yield` statement
- Without `yield`, the generator never produces a value; `contextlib._GeneratorContextManager.__enter__` calls `next()` on `None` and raises `TypeError: 'NoneType' object is not an iterator`
- Added `yield` inside the `ExitStack` block so registered severity filters stay active for the caller's `with`-block body and are cleaned up on exit

## Root Cause

```python
@contextmanager
def ignore_aborted_snapshot_upload_storage_io_errors():
    with ExitStack() as stack:
        if SkipPerIssues(...):
            stack.enter_context(...)
            stack.enter_context(...)
        # <-- missing yield here caused the crash
```

## Reproduction

Seen in `mgmt_upgrade_test.py` `test_upgrade` when `manager_backup_restore_method == "native"`:

```
TypeError: 'NoneType' object is not an iterator
/usr/local/lib/python3.14/contextlib.py:141: TypeError
```

## Testing

- [x] [ubuntu24-upgrade-test](https://jenkins.scylladb.com/job/manager-3.10/job/ubuntu24-upgrade-test/4/)

### DIFF
--- a/mgmt_upgrade_test.py
+++ b/mgmt_upgrade_test.py
@@ -144,8 +144,8 @@ class ManagerUpgradeTest(ManagerTestFunctionsMixIn, ClusterTester):
                 method=self.backup_method,
             )
             backup_task_current_details = get_task_run_details(backup_task)
-            # backup_task_snapshot = backup_task.get_snapshot_tag()
-            # pre_upgrade_backup_task_files = mgr_cluster.get_backup_files_dict(backup_task_snapshot)
+            backup_task_snapshot = backup_task.get_snapshot_tag()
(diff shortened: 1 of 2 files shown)


pehala (approved):

dimakr (approved):
LGTM

PR #14359, opened 2026-04-15
refactor(multidc-parallel-topology-schema-changes): increase data size

cezarmoise:
Currently the test has 54GB disk usage (1.1%).
Increase data written in both keyspaces.
Replace instance type with a smaller one, whitout extended storage.

Fixes: SCYLLADB-1440

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/c62a3b3b-a964-421d-8166-5dc1a06da95a
image
image

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
+++ b/test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
@@ -39,18 +39,18 @@ pre_create_keyspace: ["CREATE KEYSPACE IF NOT EXISTS keyspace1 WITH replication 
                       "ATTACH SERVICE_LEVEL sl_test_3 TO stress_user3;",
                       ]
 
-prepare_write_cmd:  ["cassandra-stress write cl=LOCAL_QUORUM n=20971520 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native connectionsPerHost=2000 -rate threads=80 -pop seq=1..20971520 -col 'n=FIXED(5) size=FIXED(16)' -log interval=5",
+prepare_write_cmd:  ["cassandra-stress write cl=LOCAL_QUORUM n=20971520 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native connectionsPerHost=2000 -rate threads=80 -pop seq=1..20971520 -col 'n=FIXED(5) size=FIXED(1024)' -log interval=5",
                      "cassandra-stress user profile=/tmp/cs_multidc_si_lcs.yaml ops'(insert=1)' cl=LOCAL_QUORUM n=3000000 -mode cql3 native connectionsPerHost=2000 -rate threads=40 -log interval=5"
                     ]
 
-stress_cmd: ["cassandra-stress write cl=LOCAL_QUORUM duration=720m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  user=stress_user1 password=stress_user1  connectionsPerHost=400 -rate threads=13 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(5) size=FIXED(16)' -log interval=5",
-             "cassandra-stress read  cl=LOCAL_QUORUM duration=720m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  user=stress_user1 password=stress_user1  connectionsPerHost=400 -rate threads=13 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(5) size=FIXED(16)' -log interval=5",
+stress_cmd: ["cassandra-stress write cl=LOCAL_QUORUM duration=720m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  user=stress_user1 password=stress_user1  connectionsPerHost=400 -rate threads=13 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(5) size=FIXED(1024)' -log interval=5",
+             "cassandra-stress read  cl=LOCAL_QUORUM duration=720m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  user=stress_user1 password=stress_user1  connectionsPerHost=400 -rate threads=13 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(5) size=FIXED(1024)' -log interval=5",
              "cassandra-stress user profile=/tmp/cs_multidc_si_lcs.yaml ops'(insert=1,read=1,si_read1=1,si_read2=1)' cl=LOCAL_QUORUM duration=700m -mode cql3 native  user=stress_user1 password=stress_user1  connectionsPerHost=400 -rate threads=13 -log interval=5",
-             "cassandra-stress write cl=LOCAL_QUORUM duration=720m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  user=stress_user2 password=stress_user2  connectionsPerHost=400 -rate threads=13 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(5) size=FIXED(16)' -log interval=5",
-             "cassandra-stress read  cl=LOCAL_QUORUM duration=720m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  user=stress_user2 password=stress_user2  connectionsPerHost=400 -rate threads=13 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(5) size=FIXED(16)' -log interval=5",
+             "cassandra-stress write cl=LOCAL_QUORUM duration=720m -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=SizeTieredCompactionStrategy)' -mode cql3 native  user=stress_user2 password=stress_user2  connectionsPerHost=400 -rate threads=13 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(5) size=FIXED(1024)' -log interval=5",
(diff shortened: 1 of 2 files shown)


pehala:
Please update the commit message to better indicate what you are changing.
scope "multidc" is not correct as you are changing single  test only, not all multidc configurations

pehala (dismissed):

jsmolar (dismissed):

jsmolar:
One comment. 
Can we also decrease the node from `2xlarge` to `xlarge`?  It would save some money, and I don't see the reason why `2xlarge` is needed here

pehala:
It seems that it didnt work as expected, disk usage is still very low :/

cezarmoise:
> It seems that it didnt work as expected, disk usage is still very low :/

Before
image
After
image

Went from 1.10% to 1.30%.

I will increase the size further from `size=FIXED(128)` to `size=FIXED(1024)`
And in `cs_multidc_si_lcs.yaml` I will double the size too.
Also Jakub's idea to use smaller instances.

in progress: https://argus.scylladb.com/tests/scylla-cluster-tests/4403748f-ceb2-441e-9fd1-85206a8e7d2c

pehala:
> I will increase the size further from `size=FIXED(128)` to `size=FIXED(1024)`

I would increase row size instead, that would be a big column IMO

cezarmoise:
With last update

image

9% usage
image

https://github.com/scylladb/scylla-cluster-tests/commit/5489561b132db7214ae0db318e0cd48f795adf32: just update the commit message

pehala:
The disk usage should be around 50%, 9% is still low.

cezarmoise:
Old keyspace1.standard1: approx 19GB
Old write: 20_971_520 Rows x 5 columns x 16 bytes
New write: 600_000_000 x 8 columns x 128 bytes
ratio: 366
New keyspace1.standard1: approx 7TB (2.3 per node)

But I can't get any runs going, getting capacity issues on every run.
image

fruch:
> Old keyspace1.standard1: approx 19GB Old write: 20_971_520 Rows x 5 columns x 16 bytes New write: 600_000_000 x 8 columns x 128 bytes ratio: 366 New keyspace1.standard1: approx 7TB (2.3 per node)
> 
> But I can't get any runs going, getting capacity issues on every run. image

on all regions ?

cezarmoise:
> > Old keyspace1.standard1: approx 19GB Old write: 20_971_520 Rows x 5 columns x 16 bytes New write: 600_000_000 x 8 columns x 128 bytes ratio: 366 New keyspace1.standard1: approx 7TB (2.3 per node)
> > But I can't get any runs going, getting capacity issues on every run. image
> 
> on all regions ?

It's multidc, so it has eu-west-1, eu-west-2
on other tests, that run on a single region, there are no issues, same instance family

fruch:
> > > Old keyspace1.standard1: approx 19GB Old write: 20_971_520 Rows x 5 columns x 16 bytes New write: 600_000_000 x 8 columns x 128 bytes ratio: 366 New keyspace1.standard1: approx 7TB (2.3 per node)
> > > But I can't get any runs going, getting capacity issues on every run. image
> > 
> > 
> > on all regions ?
> 
> It's multidc, so it has eu-west-1, eu-west-2 on other tests, that run on a single region, there are no issues, same instance family

you can still change AZs, and also regions

there's nothing special on eu-west-1 and eu-west-2

cezarmoise:
With latest changes:

| | before| after|
|--------|--------|--------|
| instance_type_db| i7ie.2xlarge|  i7i.large |
| standanrd1 schema | n=FIXED(5) size=FIXED(16)| n=FIXED(8) size=FIXED(128)|
| standanrd1 rows | 20.9M| 300M | 
| standanrd2 rows | 3M | 10M |


https://argus.scylladb.com/tests/scylla-cluster-tests/3338f43b-ce50-4ba9-86f6-17215a735f5c
DC1:
image
DC2:
image

Load is 100% during prepare, but drops to reasonable during nemesis
image

pehala (dismissed):

cezarmoise:
@scylladb/qa-maintainers 
Please review
This change also updates the instance type from i7ie to i7i. I'm seeing a lot of provisioning errors in the triggers due to it.

e.g. https://argus.scylladb.com/tests/scylla-cluster-tests/71dd327b-b79a-4f5e-90a3-270617b4c54e
image

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
```diff
@@ -61,7 +64,7 @@ n_loaders: '2 2'
 region_aware_loader: true
 simulated_racks: 0
 
-instance_type_db: 'i7ie.2xlarge'
+instance_type_db: 'i7i.large'
```
roydahan: The only problem I have is that "large" has only two shards which may be too little to detect real issues with cross-shards on multiDC scenarios.

roydahan:
@fruch what are we missing in OCI backend in order to support a multiDC case like this?
Switching to running it in OCI will allow us to save most if not all networking costs + use larger instances at the same cost of i7i.large.

pehala:
> @fruch what are we missing in OCI backend in order to support a multiDC case like this?
> Switching to running it in OCI will allow us to save most if not all networking costs + use larger instances at the same cost of i7i.large.

I would consider this for a followup, this is improvement regardless IMO, and it is already done

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
pehala: I agree, but I think the case overall is much better now than it was even if this one aspect might be a bit worse. Considering we need to save money, smaller instance seems like a good compromise here for 12 nodes

fruch:
> @fruch what are we missing in OCI backend in order to support a multiDC case like this? Switching to running it in OCI will allow us to save most if not all networking costs + use larger instances at the same cost of i7i.large.

@vponomaryov just finalize the support for it, not long ago.

we can switch this case to it but it would need to adapt again, cause it would have much bigger disk

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
```diff
@@ -70,12 +73,10 @@ nemesis_interval: 10
 nemesis_during_prepare: false
 
 seeds_num: 3
-# disabled the round robin, to each command would run on every loader.
-# round_robin: true
+# disabled the round robin, so each command would run on every loader.
+round_robin: false
```
roydahan, line 77: Are you sure about this change?
Doesn't it run now all commands on all locaders?

fruch:
> > @fruch what are we missing in OCI backend in order to support a multiDC case like this?
> > Switching to running it in OCI will allow us to save most if not all networking costs + use larger instances at the same cost of i7i.large.
> 
> I would consider this for a followup, this is improvement regardless IMO, and it is already done

```
availability_zone: 'a,b,c'
```

is what probably kills the availability, if we won't be configuring it to other region, it would be much help, cause it rarely have enough capacity

we are working on having ability to fallback to more AZs or regions, it would take some time to get it correct

I do recommend to switch this case into other regions which might be less used

fruch:
@temichus please review this as well, if I'm not mistaken you team work on maximize the number of connection,  and this PR is taking this number down, please make sure it's get reviewed by whom was working on before.

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
cezarmoise: This is not a change.
The old config had `round_robig: true` commented, which means the default of `false` of used.

I just updated it so that is shows what the value is instead of what the value isn't.

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
roydahan: maybe a first compromise would be i7i.xlarge and later when migrating to OCI it can grow back to more shards if we want.

vponomaryov:
> @fruch what are we missing in OCI backend in order to support a multiDC case like this? 

Nothing.
OCI already supports multi-dc tests after merging the following PR:
- https://github.com/scylladb/scylla-cluster-tests/pull/14376

> Switching to running it in OCI will allow us to save most if not all networking costs + use larger instances at the same cost of i7i.large.

Just to clarify, in OCI free of charge exactly inter-AZ traffic, not inter-DC.
But, the inter-DC traffic still must be cheaper than in AWS.

roydahan:
> Just to clarify, in OCI free of charge exactly inter-AZ traffic, not inter-DC.
> But, the inter-DC traffic still must be cheaper than in AWS.

Yes, this is correct.

roydahan:
> I would consider this for a followup, this is improvement regardless IMO, and it is already done

Yes, I don't mind it as a followup.

However, I think that the dataset size adjusted the dynamically (appx) based on the instance size would be better approach.
(Unless we just wanted to increase the size and it doesn't really matter if it's 50% or 70%).

pehala:
> However, I think that the dataset size adjusted the dynamically (appx) based on the instance size would be better approach.

It would be, but functionality for that is not available in SCT AFAIK.

fruch:
> > However, I think that the dataset size adjusted the dynamically (appx) based on the instance size would be better approach.
> 
> It would be, but functionality for that is not available in SCT AFAIK.

It's a nice idea, but it's not currently planned by anyone.

If one want to champion it, raise a Jira for it

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
pehala: twice the instance means twice the data to get to the desired storage utilization, meaning more cost for data transfer as well. If we want to keep reasonable spending I would keep it as it is.

roydahan: I look at it differently.
Cost saving by reducing from i7ie.2xlarge to i7i.xlarge (60% cost reduction) while keeping the test good enough.
Further cost saving would be when moving to OCI.

pehala: > I look at it differently.
> Cost saving by reducing from i7ie.2xlarge to i7i.xlarge (60% cost reduction) while keeping the test good enough.
> Further cost saving would be when moving to OCI.

We can retune it, but we will have to rerun the test in staging and it might again trigger price warnings, if you are fine with additional costs per tuning (hopefully should be like 2 runs) we can adjust.

roydahan: why isn't it just to double the size of storage and threads?
I think it's good enough.

pehala: > why isn't it just to double the size of storage and threads?
> I think it's good enough.

I would hope so, but from experience I put one additional run as expectation. Regardless it will incur more costs, hopefully not substantial, but I would still like your approval before we do that

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/8a1e0c2a91b9231ddcbe6258a522245c22a288fc:
- updated to i7i.xlarge + increased data and threads

https://argus.scylladb.com/tests/scylla-cluster-tests/c62a3b3b-a964-421d-8166-5dc1a06da95a
image
image

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
roydahan: I don't think it's necessary but It's fine.

pehala (approved):

fruch (approved):

PR #13948, opened 2026-03-09
Add Jenkins pipeline linter to validate test configurations

fruch:
## Overview
Introduces a new `lint-pipelines` command to validate SCT test configurations extracted from Jenkins pipeline files. This replaces the deprecated `lint-yamls` command with a more comprehensive solution that:

1. **Parses Jenkins pipeline files** (`*.jenkinsfile`) to extract pipeline function calls and their parameters
2. **Builds SCT environment variables** from pipeline parameters, mirroring the logic in `vars/runSctTest.groovy`
3. **Validates configurations** by instantiating `SCTConfiguration` with the derived environment
4. **Runs in parallel** using a process pool for efficiency across ~1000 pipeline files

### `sdcm/utils/lint/jenkins_parser.py`
- `parse_jenkinsfile()`: Regex-based parser for `.jenkinsfile` files
- `discover_pipeline_files()`: Recursively finds all pipeline files
- `PipelineConfig`: Dataclass representing parsed pipeline configuration
- Handles multiple quoting styles (single, double, triple quotes) and parameter formats (strings, lists, booleans)
- Supports 8 lintable pipeline functions (longevityPipeline, managerPipeline, etc.)

### `sdcm/utils/lint/env_builder.py`
- `build_env()`: Converts `PipelineConfig` to SCT environment variables
- `PARAM_TO_ENV`: Comprehensive mapping of 70+ pipeline parameters to SCT_* env vars
- Handles special cases: k8s_version → both EKS and GKE versions, extra_environment_variables parsing
- Generates placeholder images/AMIs for backends that require them at init time
- Respects non-env parameters (test_name, timeout, sub_tests, etc.)

### `sdcm/utils/lint/validator.py`
- `validate_pipeline()`: Instantiates `SCTConfiguration` with pipeline-derived environment
- Runs `verify_configuration()` and `check_required_files()` checks
- Isolates environment to prevent cross-contamination in parallel execution

### `sct.py` changes
- New `lint-pipelines` command with options for:
  - `--pipeline-dir`: Root directory of pipeline files (default: `jenkins-pipelines`)
  - `--pipeline-file`: Single file validation (ad-hoc mode)
  - `--workers`: Parallel worker count (default: CPU count)
  - `-i/--include`, `-e/--exclude`: Regex filters for pipeline selection
- Deprecation warning added to `lint-yamls` command
- Parallel validation using `ProcessPoolExecutor`

## Testing
- **413 unit tests** covering:
[...]

PR #14370, opened 2026-04-16
refactor(sct_config): Enforce strict types in config fields

cezarmoise:
**TL;DR**:  Enforce strict types in configs, disallow space separated values, add unit tests for types

## Summary

Continuation of #14256. Enforces strict types for `IntOrList` and `BooleanOrList` config fields — rejecting the old space-separated string format — and extends the same strictness to `StringOrList`. Adds a smoke-test that loads every test-case YAML through `SCTConfiguration` to catch regressions early.

## Problem

After the pydantic config migration, the three validator functions (`str_or_list_or_eval`, `int_or_list_or_eval`, `bool_or_list_or_eval`) had inconsistent behaviour and silent failure modes:

- **`IntOrList`** accepted space-separated strings (`"3 3"`) and silently coerced list elements from strings (`["3", "1"]` → `[3, 1]`), masking misconfigured YAML.
- **`BooleanOrList`** similarly accepted space-separated strings (`"true false"`).
- **`StringOrList`** did not validate list element types — `['eu-west-1', 1]` would pass with the integer silently included.

### `sdcm/sct_config.py` — validator rewrites

All three validators now follow a unified contract:

| Input | `str_or_list_or_eval` | `int_or_list_or_eval` | `bool_or_list_or_eval` |
|---|---|---|---|
| `None` | `None` | `None` | `None` |
| Bare scalar of correct type | `[value]` | scalar | scalar |
| Single-element list | `[el]` | `[el]` | unwrapped scalar |
| Multi-element list | `list[str]` | `list[int]` | `list[bool]` |
| String list literal | parsed + validated | parsed + validated | parsed + validated |
| Space-separated string | `[whole string]` | `ValueError` | `ValueError` |
| Wrong element type in list | `ValueError` | `ValueError` | `ValueError` |
| String that parses to non-list | `ValueError` | n/a (scalar ok) | n/a (scalar ok) |

Key behavioural changes:
- **`int_or_list_or_eval`**: list elements must already be `int` — string elements like `["3", "1"]` are now rejected; removed `[`-prefix gate and space-check, replaced with `ast.literal_eval`-first approach.
- **`bool_or_list_or_eval`**: removed `[`-prefix gate and space-check; `ast.literal_eval` + `json.loads` dual-parser used on any string (handles both `[True, False]` and `[true, false]`).
- **`str_or_list_or_eval`**: list elements must be `str` — non-string elements raise `ValueError`; a string that `ast.literal_eval` parses to a non-list value (e.g. `"42"`, `"{'k': 'v'}"`) now raises instead of silently wrapping.
- Removed `StrictIntOrList` / `strict_int_or_list_ints` (dead code).

### Dead code removed from consumers

- `sdcm/mgmt/operations.py` — `.split()` fallbacks on `n_loaders`
- `sdcm/sct_provision/aws/cluster.py` — `isinstance(x, str)` + `.split()` branches
- `sdcm/provision/aws/capacity_reservation.py` — `str().isdigit()` → `isinstance(n_db_nodes, int)`
- `sdcm/tester.py` — `int(n_loaders)` crash on list; `isinstance(seeds, str)` branch
- `sdcm/utils/operations_thread.py` — redundant `int(nemesis_seed)` cast

### YAML files updated (~90 files)

Multi-DC configs using space-separated `n_db_nodes`/`n_loaders` converted from `"3 3"` to `[3, 3]` YAML list syntax.

### Unit tests (`unit_tests/unit/test_config.py`)

Added 51 parametrized test cases covering all three validators:
- Valid inputs for both config path (YAML-native types) and env path (raw strings)
- Invalid inputs with expected error fragments
- Single-element list unwrap for  bool`
- Element-type rejection for all three types

---

### Testing

- [ ]

### Reminders


sdcm/mgmt/operations.py
```diff
@@ -591,13 +591,13 @@ def create_missing_rows_in_cluster(
 
 class StressLoadOperations(ClusterTester, LoaderUtilsMixin):
     def _get_total_loaders(self) -> int:
```
pehala, line 593: This is great PR to add global verification, as you touch all of the files regardless

pehala: Alternatively, you should start with global verification PR first, which would centralize the code and make this PR much smaller

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/ef8ebfb14fa88d9b60f68670b9715d3a0a9f8d08: The CI machine does not have `~/.ssh/scylla_test_id_ed25519`, so it fails at validation. 
Added a monkey patch for the `_file` validator.

unit_tests/test_test_cases_config_load.py
```diff
@@ -0,0 +1,84 @@
...
+        patch("sdcm.sct_config.get_branched_ami", return_value=[_FakeAmi()]),
+        patch("sdcm.sct_config.get_scylla_gce_images_versions", return_value=[_FakeGceImage()]),
+        patch("sdcm.sct_config.get_branched_gce_images", return_value=[_FakeGceImage()]),
+        # Skip filesystem existence checks for ExistingFile fields (e.g. user_credentials_path)
+        # so the test passes on CI machines that don't have the referenced files.
+        patch("sdcm.sct_config._file", side_effect=lambda v: v),
+    ):
+        # Instantiating SCTConfiguration merges defaults → backend defaults → YAML
+        # → env-vars.  Any unknown key or type mismatch raises during __init__.
+        sct_config.SCTConfiguration()
```
fruch, line 88: this is exactly the test case linter, how it's different from https://github.com/scylladb/scylla-cluster-tests/pull/13948 ?

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/043cfe1a8fc082f0e3f5b8b32b420615836f82a1: Add monkey patch for `convert_name_to_ami_if_needed`

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/e6fe0adb599ab6216ff5c864ec43cc4a538ce91f: monkey patch `get_arch_from_instance_type`

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/8b844e0b1c1c3d7366777e4165caaef609b6cae6:
- split config changes into per config value type
- removed smoke test, it is covered by `./utils/lint_test_cases.sh`

unit_tests/test_test_cases_config_load.py
cezarmoise: Yes. Removed the test. It's currently covered by `./utils/lint_test_cases.sh` and later by PR above.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/53c24c63f7061638f9e17ddf0fb3696491b6c368:
- fixed conflicts
- updated some docs that were missed initially.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/2e394bd2739672c5f146eabecf4dd0b10f87786a: rebase to remove some tests that were removed by #14436, but became part of my commit during an earlier rebase.

sdcm/sct_config.py
```diff
@@ -220,23 +244,39 @@ def boolean_or_space_separated_booleans(value: bool | list[bool] | str | None) -
                     result.append(v)
                 else:
                     result.append(bool(strtobool(str(v))))
-            return result
+            return result[0] if len(result) == 1 else result
         except (ValueError, TypeError) as exc:
-            raise ValueError(f"{value} isn't a list of booleans") from exc
+            raise ValueError(f"{value!r} isn't a list of booleans") from exc
```
pehala: why is the !r added everywhere?

sdcm/sct_config.py
```diff
@@ -220,23 +244,39 @@ def boolean_or_space_separated_booleans(value: bool | list[bool] | str | None) -
                     result.append(v)
                 else:
                     result.append(bool(strtobool(str(v))))
-            return result
+            return result[0] if len(result) == 1 else result
```
pehala: Why is this needed?

sdcm/sct_config.py
```diff
@@ -220,23 +244,39 @@ def boolean_or_space_separated_booleans(value: bool | list[bool] | str | None) -
...
         except (ValueError, TypeError) as exc:
-            raise ValueError(f"{value} isn't a list of booleans") from exc
+            raise ValueError(f"{value!r} isn't a list of booleans") from exc
 
     if isinstance(value, str):
+        # Try literal_eval first — handles "[True, False]", "[true]", "True", etc.
+        # ast.literal_eval handles Python-style: [True, False]
+        # json.loads handles JSON/YAML-style: [true, false]
```
pehala, line 282: How did we hanedl it before? Do we have to support both json and python style? Given you are also removing the string support, I think we can unify on one here as well

sdcm/sct_config.py
```diff
@@ -125,78 +125,109 @@ def _check_file_exists(value: str) -> None:
...
     - bool -> bool
-    - list of bools -> list of bools
-    - list of strings (true/false/yes/no/1/0) -> list of bools
-    - space-separated string of boolean values -> list of bools
+    - list[bool | str] -> bool (single element) or list[bool] (multiple elements)
+    - str containing a single boolean value (e.g. from an env var) -> bool
+    - str containing a Python-style list literal (e.g. "[True, False]") -> list[bool]
+    - str containing a JSON/YAML-style list literal (e.g. "[true, false]") -> list[bool]
+      Single-element list literals (e.g. "[True]") are unwrapped to bool.
+
```
pehala: you seems to be changing the scope of this function more than it is expected, it seems to be rewrite instead of removing one supported type

sdcm/sct_config.py
```diff
@@ -125,78 +125,109 @@ def _check_file_exists(value: str) -> None:
...
     str | list[str], BeforeValidator(str_or_list_or_eval), Field(json_schema_extra={"appendable": True})
 ]
 
 
-def int_or_space_separated_ints(value: str | int | list[int]) -> int | list[int]:
+def int_or_list_or_eval(value: str | int | list[int]) -> int | list[int] | None:
+    """Coerce value to a single int or a list of ints.
+
+    Accepts:
+    - None -> None
```
pehala, line 205: None is not int or list, why is it allowed?

sdcm/sct_config.py
cezarmoise: This is existing behaviour
```python
        if len(value) == 1:
            # Single item list, return just the bool
            if isinstance(value[0], bool):
                return value[0]
            if isinstance(value[0], str):
                return bool(strtobool(value[0]))
```

sdcm/sct_config.py
cezarmoise: The issue is that when writing in the yaml config, you need to write `true` or `false`.
But if you want to pass it via SCT_ env variable, it will not be parsed by `ast.literal_eval`, which wants the python version of `True`.

The old code did not support passing a list of bools vis SCT_ env variable, it only allowed a space separated string.

String support is not removed. '1', 'yes', 'no' still work

sdcm/sct_config.py
cezarmoise: Existing code also does it. It's needed because list is never actually used anywhere if I saw correctly

sdcm/sct_config.py
pehala: Ok, so this is a new format supported.

sdcm/sct_config.py
pehala: I see that we allowed None before as well, even though it breaks the expected type (there is no optional there). Since you are rewriting it regardless, do you think it we should keep None as supproted?

sdcm/sct_config.py
fruch: I agree those can be removed, the automatic `str()` on the values would be helpful enough for us

sdcm/sct_config.py
cezarmoise: `None` happens when the option does not exist in the current config and also does not have a default.
According to `docs/configuration_options.md` 372 options do not have a default

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/44613a2c2c6b0d0fb821ee28cdf42d977ec31205:
- rebased
- fixed conflicts
- split the last commit into 3, one for each type

sdcm/sct_config.py
cezarmoise: In this example, both `nemesis_sequence_sleep_between_ops` and `nemesis_grow_shrink_instance_type` can be `None`

```python
        sleep_time_between_ops = self.cluster.params.get("nemesis_sequence_sleep_between_ops")
        if not self.has_steady_run and sleep_time_between_ops:
            self.steady_state_latency()
            self.has_steady_run = True
        new_nodes = self._grow_cluster(rack=None)

        # pass on the exact nodes only if we have specific types for them
        new_nodes = new_nodes if self.tester.params.get("nemesis_grow_shrink_instance_type") else None
        if duration := self.tester.params.get("nemesis_double_load_during_grow_shrink_duration"):
            with self.action_log_scope("Double load after grow cluster"):
                self._double_cluster_load(duration)
        self._shrink_cluster(rack=None, new_nodes=new_nodes)
```

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/ea231115f44e346cf9aaef119af296f5a29111cf:
- rebased
- fixed conflicts

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/1be37be988e59c35efd4c2aa393a75eeef6431e5:
- added docstrings to test functions per coderabbit review

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/a9ed8999a067139825151d801648696e6ddf7919: rebased and fixed conflicts

cezarmoise:
ping @scylladb/qa-maintainers

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/f8937e39e0cd3325fdbdcf8ec64c39431bc6ff91: small unit test fix, introduced by rebase.

sdcm/sct_config.py
```diff
@@ -126,78 +126,108 @@ def _check_file_exists(value: str) -> None:
...
-            except Exception:  # noqa: BLE001
-                ret_values += [str(val)]
-        return ret_values
+        for el in value:
+            if not isinstance(el, str):
+                raise ValueError(f"List element {el} isn't a string in {value}")
+        return value
 
-    raise ValueError(f"{value} isn't a string or a list")
+    raise ValueError(f"{value} (type {type(value).__name__}) isn't a valid string or list[str]")
```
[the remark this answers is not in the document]
cezarmoise: The path mentioned in the unit test goes through `ValueError(f"List element {el} isn't a string in {value}")` not this one.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/232d34e40d3089818b886d51ee1ecb34be27783f: fix docstrings per github-actions review

pehala:
@claude review, focus on potential breakages

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/65ab85163dd19a88eb55b034add14840b22b5372: rebase, fix conflicts after #14805

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/f76d42069a2f907d659e612dbe9905184029d17c: moved tests to `unit_tests/unit/test_sct_config_validators.py`, since it already existed, but now they have expanded coverage.

cezarmoise:
@claude review, focus on potential breakages

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/92e01b9314ef99a5cd82a46740b72113a77b5086: fixed some claude findings:
- [x] 🟡 bool_or_list_or_eval docstring contradicts tests
- [x] 🟡 Stale defensive guards (minor, not a breakage)
- [ ] 🔴 Design gap: BooleanOrList type is defined but never applied to any field
  I noticed this before, should we just remove it?
- [ ] 🔴 Pre-existing (not introduced by this PR): env var value leak
  not in scope

pehala:
> [ ] 🔴 Design gap: BooleanOrList type is defined but never applied to any field
> I noticed this before, should we just remove it?

Keep for this PR, remove in the next if not needed.

> [ ] 🔴 Pre-existing (not introduced by this PR): env var value leak
> not in scope

nice bug, lets do a followup

pehala (dismissed):

cezarmoise:
> > [ ] 🔴 Pre-existing (not introduced by this PR): env var value leak
> > not in scope
> 
> nice bug, lets do a followup

I think claude is just wrong here. The `SCT_` env vars are just config parameters, which are not secrets. And the config is printed in full anyway.

fruch:
> > [ ] 🔴 Design gap: BooleanOrList type is defined but never applied to any field
> > I noticed this before, should we just remove it?
> 
> Keep for this PR, remove in the next if not needed.
> 
> > [ ] 🔴 Pre-existing (not introduced by this PR): env var value leak
> > not in scope
> 
> nice bug, lets do a followup

if any of the sct configuration options is holding a secret, that's a bug. no secret/key should find it self in clear text files in SCT code base, or in scylla.yaml.

docs/artifacts_test.md
```diff
@@ -34,7 +34,7 @@ To verify Scylla Manager package you need to provide `scylla_repo` and `scylla_m
 
 Alternatively, and even recommended, one can use the `manager_version` parameter to just choose the desired version.
 
-Optionally, you can override default cloud instance type by providing `instance_type` parameter. This parameter can be a space-separated list of instance types.
+Optionally, you can override default cloud instance type by providing `instance_type` parameter.
```
fruch: this change is wrong, this talks about the pipeline, and the pipeline does support passing a list of instances

fruch (changes requested):
drop the doc change from artifacts_test.md

since this it changing all of the test configuration, this one has high risk of breaking all of the current PRs that introduce a new test config, and high risk to leak into other backports, we'll need close attention to anything backporting test case updates (hopefully test case linting might warn us, while it's not the same in each branch)

cezarmoise:
> drop the doc change from artifacts_test.md

done https://github.com/scylladb/scylla-cluster-tests/commit/ba3bc1814163b231fecc15392e4d0b33c14a5af1

> since this it changing all of the test configuration, this one has high risk of breaking all of the current PRs that introduce a new test config, and high risk to leak into other backports, we'll need close attention to anything backporting test case updates (hopefully test case linting might warn us, while it's not the same in each branch)

it won't. 
current PRs will get a nice error message in lint-pipelines
backporting to a version like 2025.1 will have no problems. changes like `n_db_nodes: [3, 3]`  are supported everywhere, just not enforced

pehala:
> current PRs will get a nice error message in lint-pipelines

they wont until they rebase, if they dont rebase they wiull be green and only fail in master

PR #14371, opened 2026-04-16
fix(scylla-doctor): use Scylla-bundled python3

juliayakovlev:
Run scylla-doctor with Scylla-bundled python3 which ships with all
required dependencies.

For nonroot installs _find_scylla_bundled_python3 searches broadly
under user_home to find python3 in versioned directories like
scylla-VERSION/scylla-python3/bin/python3. For root installs it
uses remoter.sudo to search /opt/scylladb.

### Testing

- [x] [artifacts-rocky8-test + full scylla-doctor](https://argus.scylladb.com/tests/scylla-cluster-tests/f9bebc22-80a3-4c4a-ad8b-3dca756fa4ea)
- [x] [artifacts-rocky8-test + full scylla-doctor + nonroot](https://argus.scylladb.com/tests/scylla-cluster-tests/e0db624f-e3a4-41d3-b202-d80b0e3ff652)
- [x] [artifacts-rocky8-test + short SD](https://argus.scylladb.com/tests/scylla-cluster-tests/b28c3ae8-0fb5-4291-91b6-fc1f2befac65)
- [x] [artifacts-rocky8-test + short SD + nonroot](https://argus.scylladb.com/tests/scylla-cluster-tests/2f3d2ffc-cd1e-4daa-8eca-8cc376ffad68)
- [x] [artifacts-centos9-test + full scylla-doctor](https://argus.scylladb.com/tests/scylla-cluster-tests/ade4d207-fd12-452f-b6cb-e0812e5ed581)
- [x] [artifacts-centos9-test + full scylla-doctor + nonroot](https://argus.scylladb.com/tests/scylla-cluster-tests/5188157b-0c5b-44e0-ad1e-e1a173d6a709)
- [x] [artifacts-ami-test + full scylla-doctor](https://argus.scylladb.com/tests/scylla-cluster-tests/f9aa2564-34f1-494c-8123-219f14dd8cd6)
- [x] [artifacts-rocky9-test + full SD](https://argus.scylladb.com/tests/scylla-cluster-tests/b1554952-325f-4607-9daa-9d011c4b77d7)
- [x] [artifacts-debian12-test + full SD + nonroot](https://argus.scylladb.com/tests/scylla-cluster-tests/5932fec6-35b3-46bd-9103-af9a4864eac2)

### Reminders

### DIFF
--- a/unit_tests/test_scylla_doctor.py
+++ b/unit_tests/test_scylla_doctor.py
@@ -426,3 +426,179 @@ def test_full_edition_downloaded_flag_set_on_success(mock_node, mock_test_config
         doc.download_scylla_doctor()
 
     assert doc._full_edition_downloaded is True
+
+
+# --- _check_system_python3 tests ---
+
+
+def test_check_system_python3_suitable_version():
+    """When system python3 is >= 3.10, return its path."""
+    node = FakeNode()
+    version_result = MagicMock(ok=True, stdout="3 11\n")
+    which_result = MagicMock(ok=True, stdout="/usr/bin/python3\n")
+    node.remoter.run.side_effect = [version_result, which_result]
+
+    assert ScyllaDoctor._check_system_python3(node) == "/usr/bin/python3"
+
+
+def test_check_system_python3_too_old():
+    """When system python3 is <= 3.9, return None."""
+    node = FakeNode()
+    version_result = MagicMock(ok=True, stdout="3 9\n")
+    node.remoter.run.side_effect = [version_result]
+
+    assert ScyllaDoctor._check_system_python3(node) is None
+
+
+def test_check_system_python3_not_available():
+    """When system python3 is not installed, return None."""
+    node = FakeNode()
+    version_result = MagicMock(ok=False, stdout="")
+    node.remoter.run.side_effect = [version_result]
+
+    assert ScyllaDoctor._check_system_python3(node) is None
+
+
+def test_check_system_python3_unparseable_output():
+    """When system python3 returns unexpected output, return None."""
+    node = FakeNode()
+    version_result = MagicMock(ok=True, stdout="garbage\n")
+    node.remoter.run.side_effect = [version_result]
+
+    assert ScyllaDoctor._check_system_python3(node) is None
+
+
+def test_check_system_python3_exact_38():
+    """Python 3.8 (minor <= 9) should be rejected."""
+    node = FakeNode()
+    version_result = MagicMock(ok=True, stdout="3 8\n")
+    node.remoter.run.side_effect = [version_result]
+
+    assert ScyllaDoctor._check_system_python3(node) is None
+
+
+def test_check_system_python3_exact_310():
+    """Python 3.10 (minor > 9) should be accepted."""
+    node = FakeNode()
+    version_result = MagicMock(ok=True, stdout="3 10\n")
+    which_result = MagicMock(ok=True, stdout="/usr/local/bin/python3\n")
+    node.remoter.run.side_effect = [version_result, which_result]
+
+    assert ScyllaDoctor._check_system_python3(node) == "/usr/local/bin/python3"
+
+
+# --- _find_scylla_bundled_python3 tests ---
+
+
+def test_find_scylla_bundled_python3_nonroot_direct_path(doctor):
+    """Nonroot: when the standard bundled path exists and version > 3.9, return it."""
+    doctor.node.is_nonroot_install = True
+    doctor.node.remoter.run.side_effect = [
+        MagicMock(stdout="/home/user/scylladb/python3/bin/python3\n"),  # ls check
+        MagicMock(ok=True, stdout="3 12\n"),  # version check
+    ]
+
+    assert doctor._find_scylla_bundled_python3("/home/user") == "/home/user/scylladb/python3/bin/python3"
+
+
+def test_find_scylla_bundled_python3_nonroot_versioned_dir(doctor):
+    """Nonroot: when python3 is under a versioned dir like scylla-VERSION/scylla-python3/bin/python3."""
+    doctor.node.is_nonroot_install = True
+    doctor.node.remoter.run.side_effect = [
+        MagicMock(stdout=""),  # standard path doesn't exist
+        MagicMock(stdout="/home/ubuntu/scylla-2026.2.0~dev/scylla-python3/bin/python3\n"),  # broad find
+        MagicMock(ok=True, stdout="3 14\n"),  # version check
+    ]
+
+    result = doctor._find_scylla_bundled_python3("/home/ubuntu")
+    assert result == "/home/ubuntu/scylla-2026.2.0~dev/scylla-python3/bin/python3"
+
+
+def test_find_scylla_bundled_python3_nonroot_not_found(doctor):
+    """Nonroot: when no bundled python3 is found, return None."""
+    doctor.node.is_nonroot_install = True
+    doctor.node.remoter.run.side_effect = [
+        MagicMock(stdout=""),  # standard path doesn't exist
+        MagicMock(stdout=""),  # broad find returns nothing
+    ]
+
+    assert doctor._find_scylla_bundled_python3("/home/user") is None
+
+
+def test_find_scylla_bundled_python3_root_found(doctor):
+    """Root install: find python3 via package manager and verify version."""
+    doctor.node.is_nonroot_install = False
+    doctor.node.remoter.sudo.side_effect = [
+        MagicMock(stdout="/opt/scylladb/python3/bin/python3\n"),  # rpm -ql
+        MagicMock(ok=True, stdout="3 14\n"),  # version check
+    ]
+
+    result = doctor._find_scylla_bundled_python3("/home/user")
+    assert result == "/opt/scylladb/python3/bin/python3"
+
+
+def test_find_scylla_bundled_python3_root_not_found(doctor):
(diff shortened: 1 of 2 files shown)


pehala:
> Raise the minimum python3 version from > 3.7 to > 3.9 so that
> system Python 3.9 on CentOS 9 (which lacks PyYAML) is rejected.
> This forces scylla-doctor to use the Scylla-bundled python3
> (3.10+) which ships with all required dependencies including
> PyYAML.

This doesnt sound like a good solution. If the goal is to use always bundles python3, why dont simply use it, instead on relying on centos 9 python version not satisfying some criteria. Especially since the reason why we cannot use it has nothing to do with the version itself.

juliayakovlev:
> > Raise the minimum python3 version from > 3.7 to > 3.9 so that
> > system Python 3.9 on CentOS 9 (which lacks PyYAML) is rejected.
> > This forces scylla-doctor to use the Scylla-bundled python3
> > (3.10+) which ships with all required dependencies including
> > PyYAML.
> 
> This doesnt sound like a good solution. If the goal is to use always bundles python3, why dont simply use it, instead on relying on centos 9 python version not satisfying some criteria. Especially since the reason why we cannot use it has nothing to do with the version itself.

@pehala 
The system python3 is the primary interpreter. Added a fallback mechanism to the bundled Python distribution for cases where the host version is too old.
I fixed commit message

utils/scylla_doctor.py
```diff
@@ -368,17 +370,150 @@ def argus_collect_sd_package(self):
...
+            try:
+                major, minor = result.stdout.strip().split()
+                if int(major) >= 3 and int(minor) > 9:
+                    LOGGER.info("Using Scylla-bundled python3 (%s.%s) at %s", major, minor, python3_path)
+                    return python3_path
+                LOGGER.warning("Scylla-bundled python3 at %s is %s.%s (<= 3.9), skipping", python3_path, major, minor)
+            except (ValueError, IndexError):
+                LOGGER.debug("Could not parse version from bundled python3 at %s", python3_path)
+        else:
+            LOGGER.debug("Could not check version of bundled python3 at %s", python3_path)
```
fruch: this part isn't needed, if you aren't sure check 2025.1 version, and that the lowest we should support

utils/scylla_doctor.py
```diff
@@ -368,17 +370,150 @@ def argus_collect_sd_package(self):
...
+                python3_path = node.remoter.run("which python3", verbose=False, ignore_status=True).stdout.strip()
+                if python3_path:
+                    LOGGER.info("Using system python3 (%s.%s) at %s", major, minor, python3_path)
+                    return python3_path
+        except (ValueError, IndexError):
+            LOGGER.debug("Could not parse system python3 version from: %s", result.stdout.strip())
+
+        return None
+
+    def _find_python3_via_package_manager(self) -> str | None:
```
fruch: I don't think this is need, scylla one should always be available.

did you seen situation you are reaching this ?

fruch (commented):
Looks good, but I think we can drop some parts

utils/scylla_doctor.py
juliayakovlev: `Auto-discovered Scylla versions: 2025.4.6, 2026.1.1`
So looks like it's redundant

unit_tests/test_scylla_doctor.py
```diff
@@ -426,3 +426,179 @@ def set_flag():
...
+
+
+def test_check_system_python3_exact_310():
+    """Python 3.10 (minor > 9) should be accepted."""
+    node = FakeNode()
+    version_result = MagicMock(ok=True, stdout="3 10\n")
+    which_result = MagicMock(ok=True, stdout="/usr/local/bin/python3\n")
+    node.remoter.run.side_effect = [version_result, which_result]
+
+    assert ScyllaDoctor._check_system_python3(node) == "/usr/local/bin/python3"
```
pehala: These could be parametrized

unit_tests/test_scylla_doctor.py
```diff
@@ -426,3 +426,179 @@ def set_flag():
...
+    version_result = MagicMock(ok=True, stdout="3 6\n")
+    doctor.node.remoter.run.side_effect = [
+        version_result,  # system python3 too old
+        MagicMock(stdout=""),  # standard bundled path doesn't exist
+        MagicMock(stdout="/home/ubuntu/scylla-2026.2.0~dev/scylla-python3/bin/python3\n"),  # broad find
+        MagicMock(ok=True, stdout="3 14\n"),  # version check
+    ]
+
+    result = doctor.find_local_python3_binary("/home/ubuntu")
+    assert result == "/home/ubuntu/scylla-2026.2.0~dev/scylla-python3/bin/python3"
```
pehala: These could be parametrized as well IMO.

utils/scylla_doctor.py
```diff
@@ -368,17 +370,150 @@ def argus_collect_sd_package(self):
...
+            verbose=False,
+            ignore_status=True,
+        )
+        if not result.ok:
+            LOGGER.debug("System python3 not available")
+            return None
+
+        try:
+            major, minor = result.stdout.strip().split()
+            if int(major) >= 3 and int(minor) > 9:
```
pehala, line 391: Why is this hardcoded? I suspect the scyllaDoctor might increase minimal python version i nthe future and then we would have to change this code again (and on multiple places as the version is hardcoded in messages as well)

pehala (commented):
> The system python3 is the primary interpreter. Added a fallback mechanism to the bundled Python distribution for cases where the host version is too old.

Cannot we ALWAYS use the bundled python distribution instead? Is there a need for using system interpreter at all?

unit_tests/test_scylla_doctor.py
juliayakovlev: done

unit_tests/test_scylla_doctor.py
juliayakovlev: done

utils/scylla_doctor.py
juliayakovlev: removed

utils/scylla_doctor.py
juliayakovlev: removed

utils/scylla_doctor.py
juliayakovlev: it would be best solution. But I’d prefer not to be dependent on this.

pehala: > it would be best solution. But I’d prefer not to be dependent on this.

I dont follow, what do you mean by `not to be dependent on this`? Like do this in a followup?

juliayakovlev: I'd like to avoid a direct dependency on Scylla Doctor's environment (python version in this case)

pehala: If we are running scylla doctor, we should use what scylla doctor bundles right? It bundles it for a reason and it needs to contain everything.

Alternatively, if scylla doctor inclusion of python is accidental, then we should not rely on it at all

fruch: scylla is bundling python, for it's own reasons. scylla-doctor doesn't bundle anything, and not even documented that it can use the one bundled in scylla.

we should do what documented in scylla-doctor first, then fallback into other things.

I suggested we could use the one bundled in scylla.

regardless with are working to build SD a testing suite to be used before it's release new versions, if it would change the dependency of python for any reason, it would be found  at that point, so I don't think it's too much to be concerned at.

on a side node, we should get rid of the scylla-python, and replace the setup scripts with something more portable (bash even)

pehala: So we are using bundled python for Scylla? If so, that is also not a good solution because we cannot be sure it works. It currently works, but it is never ensured. It chance of it small, but it is inherently fragile code we are introducing and it is not a small fix (200 lines of new logic and util methods).
I would like to explore if we cannot find simpler option that will not be fragile. If we aim to replace scripts with bash in general, we can do that in a followup there.

Alternatively, I think we should choose one python source and instead ensure it has correct packages and version, that we can easily change if we need to.

fruch: what ever would build, would be in sync with SD, since they are not supplying that information or instructions 

https://docs.scylladb.com/manual/stable/troubleshooting/report-scylla-problem.html#scylladb-doctor
https://www.notion.so/Using-Scylla-Doctor-Toolkit-c30c3f9ac3764a4eb6a5174ef0dc1c10?source=copy_link

so we are testing how it's work across different distros, and in some it's doesn't work out of the box.

we can raise it as issue for scylla-doctor, and not do all of those fallbacks

we use is a user that would follow the documentation, if  do something is just a temporary work around

pehala (dismissed):

utils/scylla_doctor.py
pehala: I think it would be worth discussing this with SD team or whoever maintains it. I think the "what python should I use" is valid question user might have and should be clarified.

In the meantime, I approved this PR.

juliayakovlev:
Changed the code always to run with Scylla bundled python

@fruch @pehala please, review

pehala (approved):

fruch:
> Changed the code always to run with Scylla bundled python
> 
> @fruch @pehala please, review 

Update PR and commit titles and description 

And then we can merge

juliayakovlev:
> > Changed the code always to run with Scylla bundled python
> > @fruch @pehala please, review
> 
> Update PR and commit titles and description
> 
> And then we can merge

done

fruch (approved):

PR #14374, opened 2026-04-17
docs(plans): add config type normalization plan

pehala:
Currently, the composite config types (`StringOrList`, `IntOrList`, `BooleanOrList`, and `DictOrStr`) can return multiple types and they have to be unified in EVERY consumer by an `if isinstance()` guard. This creates a lot of duplicate code, make it easy to miss some consumers when changing types (happened during pydantic migration and we had to do multiple followup fixes) or when adding a new consumer. It is error prone.

Plan to fix BeforeValidator functions for composite config types  to always return canonical types. Phases organized per-type, each including validator fix, consumer fixes, isinstance guard removal, and unit tests.

Example https://github.com/scylladb/scylla-cluster-tests/blob/f92d0f851503130531c41e6e3fc1538b3aedfd09/sdcm/tester.py#L1546

### DIFF
--- a/docs/plans/MASTER.md
+++ b/docs/plans/MASTER.md
@@ -62,6 +62,7 @@ For plan writing guidelines, see [INSTRUCTIONS.md](INSTRUCTIONS.md).
 | Typed Config Access Migration | `pending_pr` | [#13878](https://github.com/scylladb/scylla-cluster-tests/pull/13878) |
 | SCT Config Validation and Lazy Images | `draft` | [sct-config-validation-and-lazy-images.md](sct-config-validation-and-lazy-images.md) |
 | SCT Config Follow-up Refactoring | `pending_pr` | [#13845](https://github.com/scylladb/scylla-cluster-tests/pull/13845) |
+| Config Type Normalization | `draft` | [config-type-normalization.md](config/config-type-normalization.md) |
 
 ### K8s — Kubernetes operator, K8s backends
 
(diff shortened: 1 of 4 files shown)


pehala:
The fix is now very easy, we alrady have validators for those types, we can return always a list/dict there in the branches which do not. From config side it will be an easy change.

Bigger issue will be the implicit int calls with stuff like n_db_nodes, but even that change shouldnt be that hard. It will also force us to think about multidc at all cases, because it will be always a list, which is a good thing IMO

soyacz (approved):
I think we should do it.
LGTM
P.S. progress-roadmap.svg generation is hard to follow because when reading column by column it changes the order. Order is kept from left to right - which is not I think people read like.

fruch (approved):

fruch:
Just one note, we should be touching the inputs or any type of support from the inputs (as part of this one)

pehala:
> Just one note, we should be touching the inputs or any type of support from the inputs (as part of this one)

I do not know what inputs are in this context, does it mean env var or jenkins pipelines?

fruch:
> > Just one note, we should be touching the inputs or any type of support from the inputs (as part of this one)
> 
> I do not know what inputs are in this context, does it mean env var or jenkins pipelines?

the play mentioned configuration output gonna be aligned.

so yet the inputs and how they are being handle shouldn't be changing, it might be break env vars, so I recommend not touching any of the inputs (including configuration yamls)

pehala:
> so yet the inputs and how they are being handle shouldn't be changing, it might be break env vars, so I recommend not touching any of the inputs (including configuration yamls)

Agree, any changes with regards to input will be a separate project (e.g. what Cezar is doing in https://github.com/scylladb/scylla-cluster-tests/pull/14370)

PR #14376, opened 2026-04-18
feature(oci): support OCI multi-dc setups

vponomaryov:
Code changes consist of the following parts:
- Add `OciVcnPeering` class.
  It creates Dynamic Routing Gateways (DRG), DRG-VCN attachments, RPCs, route rule and security list ingress between all region pairs.
- Add Hydra command "configure-oci-peering".
- Extend OciRegion with DRG/RPC lifecycle methods.
- Fix bunch of multi-region bugs
- Update groovy files to properly handle the OCI multi-dc setups

### Testing

- [scylla-staging/valerii/vp-oci-multidc#15](https://argus.scylladb.com/tests/scylla-cluster-tests/2b1df1b9-0e8d-4f66-a432-1129521c1ab1)
- [scylla-staging/valerii/vp-oci-multidc#16](https://argus.scylladb.com/tests/scylla-cluster-tests/1f9fb87d-766a-45ae-9525-bdea53e4a690)

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -128,6 +128,7 @@ from sdcm.utils.gce_builder import GceBuilder
 from sdcm.utils.oci_region import OciRegion
 from sdcm.utils.oci_builder import OciBuilder
 from sdcm.utils.aws_peering import AwsVpcPeering
+from sdcm.utils.oci_peering import OciVcnPeering
 from sdcm.utils.get_username import get_username
 from sdcm.utils.sct_cmd_helpers import add_file_logger, CloudRegion, get_test_config, get_all_regions
 from sdcm.utils.aws_utils import AwsArchType
@@ -2126,6 +2127,14 @@ def prepare_regions(cloud_provider, regions):
 def configure_aws_peering(regions):
     add_file_logger()
     peering = AwsVpcPeering(regions)
+    peering.configure()
+
+
+@cli.command("configure-oci-peering", help="Configure OCI cross-region VCN peering for multi-dc SCT runs")
+@click.option("-r", "--regions", type=CloudRegion(cloud_provider="oci"), default=[], help="OCI regions", multiple=True)
+def configure_oci_peering(regions):
+    add_file_logger()
+    peering = OciVcnPeering(regions or None)
     peering.configure()
 
 
--- a/sdcm/utils/oci_peering.py
+++ b/sdcm/utils/oci_peering.py
@@ -1,0 +1,128 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2026 ScyllaDB
+
+import itertools
+import logging
+
+import oci
+from oci.core.models import ConnectRemotePeeringConnectionsDetails
+
+from sdcm.utils.oci_region import OciRegion
+from sdcm.utils.oci_utils import SUPPORTED_REGIONS
+
+LOG = logging.getLogger(__name__)
+
+
+class OciVcnPeering:
+    """Configure cross-region VCN peering for OCI using DRG + Remote Peering Connections."""
+
+    def __init__(self, regions=None):
+        regions = regions or SUPPORTED_REGIONS
+        self.regions = [OciRegion(r) for r in regions]
+
+    @staticmethod
+    def peering_connection(origin_region: OciRegion, target_region: OciRegion):
+        """Establish a Remote Peering Connection between two OCI regions.
+
+        Creates DRG + DRG-VCN attachment in each region if they don't exist,
+        then creates RPCs and connects them.
+
+        Returns (origin_rpc, target_rpc).
+        """
+        # Ensure DRG infrastructure exists in both regions
+        origin_region.get_or_create_drg_attachment()
+        target_region.get_or_create_drg_attachment()
+
+        origin_rpc = origin_region.get_or_create_rpc(target_region.region_name)
+        target_rpc = target_region.get_or_create_rpc(origin_region.region_name)
+
+        if origin_rpc.peering_status == "PEERED":
+            LOG.info(
+                "RPC between %s and %s is already PEERED",
+                origin_region.region_name,
+                target_region.region_name,
+            )
+            return origin_rpc, target_rpc
+
+        LOG.info(
+            "Connecting RPC %s (%s) -> %s (%s)",
+            origin_rpc.id,
+            origin_region.region_name,
+            target_rpc.id,
+            target_region.region_name,
+        )
+        try:
+            origin_region.network_client.connect_remote_peering_connections(
+                remote_peering_connection_id=origin_rpc.id,
+                connect_remote_peering_connections_details=ConnectRemotePeeringConnectionsDetails(
+                    peer_id=target_rpc.id,
+                    peer_region_name=target_region.region_name,
+                ),
+            )
+        except oci.exceptions.ServiceError as exc:
+            if "InvalidState" in str(exc) or "already peered" in str(exc).lower():
+                LOG.info(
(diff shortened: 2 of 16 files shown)


vars/createSctRunner.groovy
```diff
@@ -27,8 +27,6 @@ def call(Map params, Integer test_duration, String region) {
 
     if ( params.backend.equals("azure") ) {
         region_zone_arg = "--region " + params.azure_region_name
-    } else if ( params.backend.equals("oci") )  {
-        region_zone_arg = "--region " + params.oci_region_name
```
fruch, line 31: why removing this one ?

sct.py
```diff
@@ -2129,6 +2130,14 @@ def configure_aws_peering(regions):
     peering.configure()
 
 
+@cli.command("configure-oci-peering", help="Configure OCI cross-region VCN peering for multi-dc SCT runs")
+@click.option("-r", "--regions", type=CloudRegion(cloud_provider="oci"), default=[], help="OCI regions", multiple=True)
+def configure_oci_peering(regions):
```
fruch, line 2135: nitpick or followup: maybe consolidating it with the AWS command,

vars/createSctRunner.groovy
vponomaryov: Because the `else` block does proper behavior for both - multidc and singledc.
This removed block will provide the list of multiple region-strings as-is.
So, we must remove it to get always a string with the first region from a list if it was a list.

sct.py
vponomaryov: That one has backend-specific name.
We can have backend-agnostic command, yes.
But it is out of scope for PR, because this PR follows the current approach - backend-specific commands.

sdcm/utils/oci_peering.py
```diff
@@ -0,0 +1,128 @@
...
+        """Establish a Remote Peering Connection between two OCI regions.
+
+        Creates DRG + DRG-VCN attachment in each region if they don't exist,
+        then creates RPCs and connects them.
+
+        Returns (origin_rpc, target_rpc).
+        """
+        # Ensure DRG infrastructure exists in both regions
+        _ = origin_region.drg_attachment
+        _ = target_region.drg_attachment
```
pehala: I do not think we should rely on properties to create something. This construction is misleading IMO, we shouldnt use getters if we dont care about returned value but want to execute magic inside

I would change to
```suggestion
        # Ensure DRG infrastructure exists in both regions
        origin_region.create_drg_if_doesnt_exist()
        target_region.create_drg_if_doesnt_exist()
```

sdcm/utils/oci_peering.py
```diff
@@ -0,0 +1,128 @@
...
+        try:
+            origin_region.network_client.connect_remote_peering_connections(
+                remote_peering_connection_id=origin_rpc.id,
+                connect_remote_peering_connections_details=ConnectRemotePeeringConnectionsDetails(
+                    peer_id=target_rpc.id,
+                    peer_region_name=target_region.region_name,
+                ),
+            )
+        except oci.exceptions.ServiceError as exc:
+            if "InvalidState" in str(exc) or "already peered" in str(exc).lower():
```
pehala, line 73: How can this happen, if you check above?

sdcm/utils/oci_peering.py
```diff
@@ -0,0 +1,128 @@
...
+        origin_region.add_drg_route_to_tables(peer_vcn_cidr=str(target_region.vcn_ipv4_cidr))
+
+    @staticmethod
+    def open_security_list(origin_region: OciRegion, target_region: OciRegion):
+        """Allow all traffic from target_region's VCN CIDR in origin_region's security list."""
+        origin_region.allow_cross_vcn_traffic(peer_vcn_cidr=str(target_region.vcn_ipv4_cidr))
+
+    def configure(self):
+        """Configure full cross-region peering for all region pairs."""
+        if len(self.regions) < 2:
```
pehala, line 105: nit:
```suggestion
        if len(self.regions) == 1:
```
This is clearer IMO

sdcm/utils/oci_region.py
```diff
@@ -767,7 +936,7 @@ def _subnet_dns_label(self, public: bool) -> str:
         label_source = f"{self.region_name}-{visibility}"
         checksum = hashlib.sha1(label_source.encode("utf-8")).hexdigest()[:7]
         base = self._dns_label_from_name(visibility)
-        return f"{base[:7]}-{checksum}"
+        return f"{base[:7]}{checksum}"
```
pehala, line 942: why is this needed?

sdcm/utils/oci_region.py
```diff
@@ -719,6 +727,167 @@ def _get_home_region(self) -> str:
...
+            if already_routed:
+                LOGGER.info("Route to %s via DRG already exists in %s route table", peer_vcn_cidr, rt_name_suffix)
+                continue
+            new_rule = RouteRule(
+                destination=str(peer_vcn_cidr),
+                destination_type="CIDR_BLOCK",
+                network_entity_id=drg_id,
+                description=f"Route to peer VCN {peer_vcn_cidr} via DRG",
+            )
+            updated_rules = list(existing_rules) + [new_rule]
```
pehala: nit: This will remove two unnecessary list creations
```suggestion
            updated_rules = [*existing_rules, new_rule]
```

unit_tests/unit/test_oci_peering.py
```diff
@@ -0,0 +1,487 @@
...
+    ),
+)
+@patch(
+    "sdcm.utils.oci_region.OciRegion.drg",
+    new_callable=PropertyMock,
+    return_value=MagicMock(id="ocid1.drg.oc1..123"),
+)
+@patch("sdcm.utils.oci_utils.OciService.get_network_client", return_value=MagicMock())
+@patch("sdcm.utils.oci_utils.OciService.get_identity_client", return_value=MagicMock())
+@patch("sdcm.utils.oci_utils.OciService.get_compute_client", return_value=MagicMock())
```
pehala: These pathes are shared among many tests, might make sense extracting into a fixture

sdcm/utils/oci_peering.py
pehala, line 1: The amount of patches on top of each test is too big IMO, it made it hard for me to decypher which of them are part of test setup and which are for not calling OCI. I would extract patches for not calling OCI into a general sutouse fixture, so the tests are annotated only with stuff relevant for that specific test. It will make it easier to understand

pehala (commented):
The patch look alright, it follows what we are doing with other backend. But I think the fact we have to have soo much code in so many files for multidc support for one backend signals that our general design of backends is not good enough IMO.

This is, of course, not an issue of this PR, but a food for thought for future improvement for SCT infrastructure in general.

sdcm/utils/oci_peering.py
vponomaryov: Multiple unrelated actors.
You, me, any other ScyllaDB employee may run this code in parallel...

It has no harm and has more safety.
I prefer to keep it as-is.

sdcm/utils/oci_peering.py
vponomaryov: Ok, will change.

sdcm/utils/oci_peering.py
vponomaryov: We should skip not only for `1`, but also for `0`.

sdcm/utils/oci_region.py
vponomaryov: Because it is bug.
First region was working ok all that time just because labels already existed.
Second region didn't have labels and was failing on the disallowed symbol which is removed here.

sdcm/utils/oci_region.py
vponomaryov: ok, will do

unit_tests/unit/test_oci_peering.py
vponomaryov: Actually these `compute` and `identity` mocks are redundant in multiple tests.
Will just remove it.

sdcm/utils/oci_peering.py
vponomaryov: I removed redundant mocks

vponomaryov:
Latest PR state is verified here: [scylla-staging/valerii/vp-oci-multidc#16](https://argus.scylladb.com/tests/scylla-cluster-tests/1f9fb87d-766a-45ae-9525-bdea53e4a690)

fruch (approved):

PR #14383, opened 2026-04-20
fix(gce-provision): fail fast on ZONE_RESOURCE_POOL_EXHAUSTED

soyacz:
## Summary
- GCE provisioning previously retried 3x with 15-min sleep on any `GoogleAPIError`, including `ZONE_RESOURCE_POOL_EXHAUSTED`. Retrying the same exhausted zone cannot recover; the 45-min retry cycle collided with Jenkins pipeline timeouts, aborting tests mid-retry in the provision stage.
- Introduce `ProvisionUnrecoverableError` (sibling of `ProvisionError`) and `ZoneResourcesExhaustedError`, raised from the GCE instance provider when the API reports `ZONE_RESOURCE_POOL_EXHAUSTED`.
- Upstream `provision_with_retry` retries only `ProvisionError`, so the new exception propagates immediately — test fails fast instead of hanging.

Fixes SCT-234

## Test plan
- [ ] Unit: mocked `GoogleAPIError` with `ZONE_RESOURCE_POOL_EXHAUSTED` message propagates `ZoneResourcesExhaustedError` without retry sleep
- [ ] Unit: non-zone-exhausted `GoogleAPIError` still follows existing retry path
- [ ] Dry-run longevity GCE job in an exhausted zone — confirm fast-fail (<5 min) instead of 45-min hang

### DIFF
--- a/sdcm/provision/gce/instance_provider.py
+++ b/sdcm/provision/gce/instance_provider.py
@@ -20,7 +20,12 @@ from typing import List, Dict, Optional
 import google.api_core.exceptions
 from google.cloud import compute_v1
 
-from sdcm.provision.provisioner import InstanceDefinition, PricingModel, ProvisionError
+from sdcm.provision.provisioner import (
+    InstanceDefinition,
+    PricingModel,
+    ProvisionError,
+    ZoneResourcesExhaustedError,
+)
 from sdcm.provision.gce.disk_provider import DiskProvider
 from sdcm.provision.gce.network_provider import NetworkProvider
 from sdcm.provision.gce.constants import DISK_TYPE_PD_STANDARD, DISK_TYPE_LOCAL_SSD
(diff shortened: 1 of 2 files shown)


fruch:
@soyacz this is cosmetic, we'll need a fallback to other AZs like we have for AWS artifacts ...

soyacz:
> @soyacz this is cosmetic, we'll need a fallback to other AZs like we have for AWS artifacts ...

yes, just to fix current issue. We didn't have it in GCE in the past and now we could add it (along with Azure).
Do we have task to implement it in GCE/Azure (now both will be updated in one change)?

fruch (approved):

fruch:
> > @soyacz this is cosmetic, we'll need a fallback to other AZs like we have for AWS artifacts ...
> 
> yes, just to fix current issue. We didn't have it in GCE in the past and now we could add it (along with Azure). Do we have task to implement it in GCE/Azure (now both will be updated in one change)?

we don't have a task yet, I just had a plan for doing fallbacks for all cases on all backended (which is a super-set of that), if we have issues now with GCE, we can do just that first for the artifacts tests

PR #12805, opened 2025-12-03
fix(nemesis): fix _refuse_connection_from_banned_node behaviour according to scylla changes

timtimb0t:
Scylla commit https://github.com/scylladb/scylladb/pull/26943 changed the
behavior of banned nodes: once they receive NOTIFY_BANNED they log a
message and terminate instead of staying up with blocked ports.

Update the nemesis to align with this behavior:
- stop expecting the banned node to come back up and listen on blocked ports
- attempt CQL queries from the banned node and ensure they do not succeed
- wait for the banned node to terminate
- verify NOTIFY_BANNED / terminating messages in the node logs
- ensure no data from the banned node appears in the cluster

### Testing
https://jenkins.scylladb.com/job/scylla-staging/job/eugene_test_folder/job/banned_node/11/
https://argus.scylladb.com/tests/scylla-cluster-tests/100f9764-1a92-4e18-b93b-a6e44b6fc1ba

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #14391, opened 2026-04-20
fix(nemesis): fix log follow pattern

cezarmoise:
The log was changed in scylladb/scylladb@22949bae520b1fc482235193f98715de6285bfa2

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis/monkey/abort_decommission.py
+++ b/sdcm/nemesis/monkey/abort_decommission.py
@@ -53,7 +53,7 @@ class AbortDecommissionMonkey(NemesisBaseClass):
         # indicating that SSTable streaming has finished for at least one sstable.
         # Once the `log_follower` yields a line, we know streaming has started.
         log_follower = self.runner.target_node.follow_system_log(
-            patterns=[r"stream_blob - stream_sstables\[[0-9a-f-]+\] Finished sending sstable_nr"]
+            patterns=[r"stream_blob - stream_sstables\[[0-9a-f-]+\] Finished sending files_nr"]
         )
 
         _streaming_started = lambda: len(list(log_follower)) > 0


pehala (approved):

timtimb0t (approved):

vponomaryov:
@cezarmoise , @pehala 

This change affects all who use nemesis, not just your team

And, IIUC, now it will trigger `false negative result/error` using non-latest dev ScyllaDB version...
Please, don't merge that fast - in 33 minutes, and make the code support both approaches - old and new one.

We have had the ScyllaDB version-based conditions in SCT for long time.
And this is the right case to apply it.

pehala:
> This change affects all who use nemesis, not just your team

This was approved by the people who encounter it the most early: Tier 1 and Sanities, so I think this is fine. Do you feel like every change to any nemesis needs to approved by a wider audience, if so, what do you propose?

> And, IIUC, now it will trigger `false negative result/error` using non-latest dev ScyllaDB version...

Master should be up to date on what is scylla master AFAIK right?

> We have had the ScyllaDB version-based conditions in SCT for long time.

Did we ever use it for development versions? I am aware of using this only for released versions, where it makes more sense. Not sure what version would be even check, all of the versions since 2026.1 will be 2026.2.0dev, so it will be hard to discern the specific one we need. Followup can be easily submitted if you have idea how to gate this properly

> Please, don't merge that fast - in 33 minutes, and make the code support both approaches - old and new one.

 It got two approves, it got merged. One maintainer approved as well. The problem is not speed of merge but the fact it got approved by two people without this being raised, which is valid concern but a different one

EDIT: If you need to test older development versions, you can use SCT master from the same date. I do not think SCT master should support older development versions

pehala:
Also if you think this patch in its current state shouldnt have been merged, feel free to submit revert and we can discuss it there.

vponomaryov:
@pehala

SCT master branch must not depend only on the latest dev version of ScyllaDB - it is violated here.
Running tests using SCT master with ScyllaDB GA releases for comparison happens periodically in some of ours activities.
Also, developing new features for SCT I base it on master branch and use GA ScyllaDB versions if available...
And not only I do so. That's why I mentioned that the referenced ScyllaDB PR was not backported to other ScyllaDB branches.

Such backwards incompatible changes make me workaround it for totally ok scenarios like I described above.

Proposing fixing of it myself is nice, then, should I do it each time backward incompatible change gets introduced with resistance to others working routines?
It doesn't sound productive to me.

pehala:
> SCT master branch must not depend only on the latest dev version of ScyllaDB - it is violated here.
> Running tests using SCT master with ScyllaDB GA releases for comparison happens periodically in some of ours activities.
> Also, developing new features for SCT I base it on master branch and use GA ScyllaDB versions if available...
> And not only I do so. That's why I mentioned that the referenced ScyllaDB PR was not backported to other ScyllaDB branches.

Maybe I am wrong here, but I dont think this was ever the case. When something changed in Scylla we immediately changed it in SCT (e.g. https://github.com/scylladb/scylla-cluster-tests/pull/12805, https://github.com/scylladb/scylla-cluster-tests/pull/13586, https://github.com/scylladb/scylla-cluster-tests/pull/14131). These backwards incompatible changes are not frequent.

If SCT master needs to work on GA, then that would be a new requirement for me, I might have just not been aware, but regardless we should declare it somewhere explicitly, because it is not clear.
I dont think this should be the case, IMO SCT master should track Scylla master and I do not want to enforce additional work to verify that the PR works on latest GA. I think support for these rare cases should be best effort basis.

> Such backwards incompatible changes make me workaround it for totally ok scenarios like I described above.

I do not object to a version guard here, like I said feel free to either revert and/or submit a fix with a guard that works here.
Quick search revealed we do not have many version guards and none of them is date based, which is what would be neccessary to do here. I see RC guards but these would not be helpful here

Edit: If you feel like process failed (e.g. the merge too fast), please also suggest process changes

fruch:
> > SCT master branch must not depend only on the latest dev version of ScyllaDB - it is violated here.
> > Running tests using SCT master with ScyllaDB GA releases for comparison happens periodically in some of ours activities.
> > Also, developing new features for SCT I base it on master branch and use GA ScyllaDB versions if available...
> > And not only I do so. That's why I mentioned that the referenced ScyllaDB PR was not backported to other ScyllaDB branches.
> 
> Maybe I am wrong here, but I dont think this was ever the case. When something changed in Scylla we immediately changed it in SCT (e.g. #12805, #13586, #14131). These backwards incompatible changes are not frequent.
> 
> If SCT master needs to work on GA, then that would be a new requirement for me, I might have just not been aware, but regardless we should declare it somewhere explicitly, because it is not clear. I dont think this should be the case, IMO SCT master should track Scylla master and I do not want to enforce additional work to verify that the PR works on latest GA. I think support for these rare cases should be best effort basis.
> 
> > Such backwards incompatible changes make me workaround it for totally ok scenarios like I described above.
> 
> I do not object to a version guard here, like I said feel free to either revert and/or submit a fix with a guard that works here. Quick search revealed we do not have many version guards and none of them is date based, which is what would be neccessary to do here. I see RC guards but these would not be helpful here
> 
> Edit: If you feel like process failed (e.g. the merge too fast), please also suggest process changes

process didn't fail

but we should have backward compatibility in mind 
1) cause we are using master for all kind of testing, for example someone working on performance tests (some of them have nemesis involved)
2) those can easily split into backport,with other fixes that we do want backport, and it happens from time to time.

anyhow in case of regex, it's quite simple to have one that support both cases.
also it's quite common we do that a lot in functional tests (like in dtest), to avoid disruption to everyone 
and can even do those upfront (which is not the case, cause no one is running SCT cases on PRs)

pehala:
Makes sense. I would still mention the expected compatibility of master somewhere (maybe the SCT maintainer handbook?).

PR #13016, opened 2025-12-24
refactor(nemesis): remove run_with_gemini parameter from SCT

yarongilor:
The run_with_gemini parameter is no longer needed as nemesis operations are compatible with all test types following the recent gemini refactor. Removed the flag from NemesisFlags class, all nemesis classes, nemesis.yml configuration, and updated gemini test cases to remove run_with_gemini from their nemesis_selector expressions. 
Fixes: #13003

### Testing

- [gemini-1tb-10h-test](https://argus.scylladb.com/tests/scylla-cluster-tests/1d477c36-a689-45cd-a21a-3f18b7a50bc5)

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #14393, opened 2026-04-20
fix(gemini-test): exclude node isolation nemesis from gemini test cases

yarongilor:
When gemini is running and node isolation nemesis (IsolateNodeWithIptableRuleNemesis or IsolateNodeWithProcessSignalNemesis) triggers, the isolated node cannot receive changes from other nodes, causing gemini validation to fail.

Exclude both IsolateNodeWithIptableRuleNemesis and IsolateNodeWithProcessSignalNemesis from the nemesis_selector in 4 gemini test-case YAML files that use negative selection filters.

Ref: https://scylladb.atlassian.net/browse/SCT-195

### Testing

- [ ]

### Reminders

### DIFF
--- a/test-cases/gemini/gemini-1tb-10h.yaml
+++ b/test-cases/gemini/gemini-1tb-10h.yaml
@@ -8,7 +8,7 @@ instance_type_loader: 'm6i.xlarge'
 user_prefix: 'gemini-1tb-10h'
 
 nemesis_class_name: 'SisyphusMonkey'
-nemesis_selector: 'not enospc'
+nemesis_selector: 'not enospc and not IsolateNodeWithIptableRuleNemesis and not IsolateNodeWithProcessSignalNemesis'
 nemesis_seed: '041'
 
 gemini_cmd: |
--- a/test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
+++ b/test-cases/gemini/gemini-3h-ics-cdc-with-nemesis.yaml
@@ -8,7 +8,7 @@ user_prefix: 'ics-cdc-gemini-basic-3h'
 ami_db_scylla_user: 'centos'
 
 nemesis_class_name: 'SisyphusMonkey'
-nemesis_selector: "not enospc"
+nemesis_selector: "not enospc and not IsolateNodeWithIptableRuleNemesis and not IsolateNodeWithProcessSignalNemesis"
 
 gemini_cmd: |
   --duration 3h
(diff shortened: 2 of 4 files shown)


pehala:
I am ok with it, but didnt we have gemini flag before? Shouldnt we resurrect it, even if it wont be massivelly used, for these exact cases where we temporarily want to remove some nemesis from gemini runs? It would also make it easier to change (only at one place) and we could add docstring to each explicit removal mentioning why

CodeLieutenant (approved):

CodeLieutenant:
> I am ok with it, but didnt we have gemini flag before? Shouldnt we resurrect it, even if it wont be massivelly used, for these exact cases where we temporarily want to remove some nemesis from gemini runs? It would also make it easier to change (only at one place) and we could add docstring to each explicit removal mentioning why

What flag, I don't recall any flag?

pehala:
> What flag, I don't recall any flag?

I might be misremembering. In that case I propose a Nemesis Flag, which would be True by default and we would filter by flag in the testconfigs. For disabling a nemesis we set it to False with a comment  and link to issue why it doesnt work. IF we want to reenable we just remove the flag from specific nemesis

yarongilor:
> > What flag, I don't recall any flag?
> 
> I might be misremembering. In that case I propose a Nemesis Flag, which would be True by default and we would filter by flag in the testconfigs. For disabling a nemesis we set it to False with a comment and link to issue why it doesnt work. IF we want to reenable we just remove the flag from specific nemesis

@pehala , correct, that was the `run_with_gemini` label we removed in https://github.com/scylladb/scylla-cluster-tests/pull/13016.
It could indeed be an alternative to resurrect this label in a possible simpler way. does it sound better?

pehala:
> > > What flag, I don't recall any flag?
> > 
> > I might be misremembering. In that case I propose a Nemesis Flag, which would be True by default and we would filter by flag in the testconfigs. For disabling a nemesis we set it to False with a comment and link to issue why it doesnt work. IF we want to reenable we just remove the flag from specific nemesis
> 
> @pehala , correct, that was the `run_with_gemini` label we removed in https://github.com/scylladb/scylla-cluster-tests/pull/13016.
> It could indeed be an alternative to resurrect this label in a possible simpler way. does it sound better?
> 

Yes it does

fruch:
@pehala if the repetition is the problem, put it in one yaml file and share it with all of the jobs (it's just 3), I don't think a new SCT configuration is need for this case

pehala:
> @pehala if the repetition is the problem, put it in one yaml file and share it with all of the jobs (it's just 3), I don't think a new SCT configuration is need for this case
> 
> 

I did not want another SCT configurstion I wanted just a flag. We will discuss it internally and maybe submit followup

PR #14395, opened 2026-04-21
fix(nemesis): don't abort test on teardown-killed stress

dimakr:
Stress failures with exist status 137 (SIGKILL from test teardown) are flagged as Severity.WARNING, but their messages end up in the errors dict passed to
`sdcm.nemesis.NemesisRunner._nemesis_stress_failure_handler`, causing a false-negative NemesisStressFailure.

The change refactors base stress thread `parse_results` method to return errors with their severities. Additionally, the `verify_stress_thread` in tester.py is updated to
only log the warning severity stress thread entries, thus avoding false-negative stress failures.

Fixes: SCT-183

### Testing

- [x] :green_circle: [test from original issue - longevity-scylla-cloud-5gb-1h](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/262/)
- [x] :green_circle:  [longevity-10gb-3h test on aws](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/263/)

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -3162,28 +3162,40 @@ class ClusterTester(unittest.TestCase):
         :return: bool, indication whether all stress threads have completed successfully
         """
         results, errors = thread_pool.parse_results()
-        if results:
-            pass
-        else:
+        if not results:
             self.log.warning("There is no stress results, probably stress thread has failed.")
 
+        # split errors by severity so WARNING-severity errors do not trigger test failure
+        real_errors, warnings = {}, {}
+        if isinstance(errors, dict):
+            for loader_name, errs in errors.items():
+                for severity, msg in errs:
+                    target = warnings if severity == Severity.WARNING else real_errors
+                    target.setdefault(loader_name, []).append(msg)
+        else:
+            real_errors = errors
+
+        if warnings:
+            warnings_msg = "\n".join(f"{node}:\n  " + "\n  ".join(msgs[-5:]) for node, msgs in warnings.items() if msgs)
+            self.log.warning("stress warnings on nodes:\n%s", warnings_msg)
+
         if error_handler:
-            error_handler(thread_pool, errors)
-        if errors:
+            error_handler(thread_pool, real_errors)
+        if real_errors:
             # Sometimes, we might have an epic error messages list
             # that will make small machines driving the test
             # to run out of memory when writing the XML report. Since
             # the error message is merely informational, let's simply
             # use the last 5 lines for the final error message.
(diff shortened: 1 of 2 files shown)


sdcm/kafka/kafka_consumer.py
```diff
@@ -114,6 +114,6 @@ def stop(self):
     def kill(self):
         self.stop()
 
-    def parse_results(self) -> (list[dict | None], list[str | None]):
+    def parse_results(self) -> tuple[list, list, dict]:
```
pehala: nit: it would be better if you defined the types more. Like dict[str, str]. Given this would be long you can create type alias and use it for all results

pehala:
Do we need the warnings in general?

dimakr:
> Do we need the warnings in general?

Hard to answer.
But it's good to some have additional info, if RCA is needed.

sdcm/tester.py
```diff
@@ -3161,12 +3161,16 @@ def verify_stress_thread(self, thread_pool: list, error_handler: callable = None
...
         :return: bool, indication whether all stress threads have completed successfully
         """
-        results, errors = thread_pool.parse_results()
+        results, errors, warnings = thread_pool.parse_results()
         if results:
             pass
         else:
             self.log.warning("There is no stress results, probably stress thread has failed.")
 
+        if warnings:
```
pehala, line 3178: cannot you filter events here? It would make the PR simpler, as no changes to the stress tool would be needed.

sdcm/kafka/kafka_consumer.py
dimakr: updated

sdcm/tester.py
dimakr: previous version already merged error and warning severities into  errors dict returned above by `thread_pool.parse_results()`.
Hence we would need to bring some logic out of `sdcm.stress.base.DockerBasedStressThread.parse_results` here, for additional filtering, coupling tester.py to stress thread internals

dimakr: As was discussed  - added returning severity along with errors from `sdcm.stress.base.DockerBasedStressThread.parse_results`. This way the returning value of this method doesn't change (tuple of 2 items) and existing usages of the method don't require updates. The consumers of `parse_results` can process the returning value themselves if needed, like now the tester.py `verify_stress_thread` does

pehala (approved):
LGTM

fruch (approved):

PR #14430, opened 2026-04-26
fix(upgrade-test): exclude .tmp files from sstabledump verification

fruch:
## Summary
- Exclude `.tmp` files from the `find` command in `test_rolling_upgrade` Step 7 (sstabledump verification), preventing failures when `upgradesstables` compaction is still active

Fixes: SCT-264

### Testing

- [x] 🟢 [rolling-upgrade-ami-test #43](https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/rolling-upgrade-ami-test/43/)


roydahan:
There is a fix pending in Scylla, would it still be needed here:

https://github.com/scylladb/scylladb/pull/29620

fruch:
> There is a fix pending in Scylla, would it still be needed here:
> 
> [scylladb/scylladb#29620](https://github.com/scylladb/scylladb/pull/29620)

seems like tests are passing with this small patch, with current scylla.

I'll merge it anyhow, as a safety

PR #14442, opened 2026-04-27
chore(azure): improve azure  disk performance to avoid c-s failures

timtimb0t:
Increase the Azure instance size to avoid  disk bottleneck that
 caused test failures

The workload uses a relatively small dataset (~10M keys / ~10GB logical data, RF=3), so it should be treated as a mixed IO / resilience / tail-latency test rather than a large-capacity test. During the add/remove index operation, several Azure nodes show clear localized disk starvation: compaction, memtable, sl:default, and main I/O queues spike, with sl:default disk starvation reaching ~5.5s on one node and ~4.1s on another.

This change moves the affected test to a stronger Azure storage-optimized instance to reduce disk bottlenecks and improve test stability.

Disk performance comparison: Standard_L8s_v3 vs Standard_L8s_v4

Standard_L8s_v3: ~400k IOPS (only aggregated info)
Standard_L8s_v4: ~550k read IOPS

Increase: +150k IOPS
Percentage increase: +37.5%

Fixes: https://scylladb.atlassian.net/browse/SCYLLADB-1729

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-10gb-3h.yaml
+++ b/test-cases/longevity/longevity-10gb-3h.yaml
@@ -8,7 +8,7 @@ n_loaders: 2
 instance_type_db: 'i4i.2xlarge'
 gce_instance_type_db: 'n2-highmem-16'
 gce_instance_type_loader: 'e2-standard-4'
-azure_instance_type_db: 'Standard_L8s_v3'
+azure_instance_type_db: 'Standard_L8s_v4'
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "keyspace1.standard1", "interval": 10}']
 nemesis_class_name: 'SisyphusMonkey'
 nemesis_seed: '111'


pehala:
I am missing explanation of the problem you are facing and explanation of how this will help.
`improve azure disk performance to avoid c-s failures` What does that mean? Do stronger instances have better perfoming disks? I am aware of more size but generally performance is similar for the entire family no?

Also no issue or run that failed

fruch:
> I am missing explanation of the problem you are facing and explanation of how this will help. `improve azure disk performance to avoid c-s failures` What does that mean? Do stronger instances have better perfoming disks? I am aware of more size but generally performance is similar for the entire family no?
> 
> Also no issue or run that failed

it a new family v4, and yes it's disks is much better.

after @timtimb0t was hitting those disk related slowdowns, in this specific case, I suggested that we can't leave it as is, and we need to improve something, my suggestion was a newer family (that we tested it's performance few months ago, and should be using it anyhow)

see this, for the full discussion:
https://argus.scylladb.com/tests/scylla-cluster-tests/31354cf8-3130-4793-9697-d0a2a07065a7/discuss

pehala:
> > I am missing explanation of the problem you are facing and explanation of how this will help. `improve azure disk performance to avoid c-s failures` What does that mean? Do stronger instances have better perfoming disks? I am aware of more size but generally performance is similar for the entire family no?
> > Also no issue or run that failed
> 
> it a new family v4, and yes it's disks is much better.
> 
> after @timtimb0t was hitting those disk related slowdowns, in this specific case, I suggested that we can't leave it as is, and we need to improve something, my suggestion was a newer family (that we tested it's performance few months ago, and should be using it anyhow)
> 
> see this, for the full discussion: https://argus.scylladb.com/tests/scylla-cluster-tests/31354cf8-3130-4793-9697-d0a2a07065a7/discuss

Thanks, makes sense.

@timtimb0t Thanks for providing better description now as well, but please have it there from the start next time.

change looks alright, but I would like to see run with the new instance first, to make sure it fixes the issue

timtimb0t:
> > > I am missing explanation of the problem you are facing and explanation of how this will help. `improve azure disk performance to avoid c-s failures` What does that mean? Do stronger instances have better perfoming disks? I am aware of more size but generally performance is similar for the entire family no?
> > > Also no issue or run that failed
> > 
> > 
> > it a new family v4, and yes it's disks is much better.
> > after @timtimb0t was hitting those disk related slowdowns, in this specific case, I suggested that we can't leave it as is, and we need to improve something, my suggestion was a newer family (that we tested it's performance few months ago, and should be using it anyhow)
> > see this, for the full discussion: https://argus.scylladb.com/tests/scylla-cluster-tests/31354cf8-3130-4793-9697-d0a2a07065a7/discuss
> 
> Thanks, makes sense.
> 
> @timtimb0t Thanks for providing better description now as well, but please have it there from the start next time.
> 
> change looks alright, but I would like to see run with the new instance first, to make sure it fixes the issue

https://argus.scylladb.com/tests/scylla-cluster-tests/ba454c9d-c981-4374-9505-3637fdcd9d30
here is the successful run

The c-s failure is regular, but still random in nature. I’ll rerun the test again.
I’m not running a single nemesis because it fails due to a combination of factors, and the system needs to be under load with at least a minimal dataset.

pehala (approved):

timtimb0t:
@fruch , ping

PR #14450, opened 2026-04-27
feature(nemesis): introduce protected_db_nodes

fruch:
## Summary

Introduces `protected_db_nodes` — a list of DB node indexes that will never be selected as the target of any nemesis disruption.

```yaml
n_db_nodes: 6
protected_db_nodes: [1, 4]   # nodes at index 1 and 4 are always skipped by nemesis
```

Also adds a `full_cluster_restart` flag to every nemesis entry in `nemesis.yml`, marking the nemeses (rolling restarts, KMS encryption enable/disable, etc.) that restart every node and therefore cannot honour the protected-node contract. Tests that use `protected_db_nodes` should pair this with `nemesis_selector: 'not full_cluster_restart'` to skip those nemeses automatically.

## Changes

| File | Change |
|---|---|
| `sdcm/cluster.py` | `BaseNode.is_protected` property |
| `sdcm/nemesis/__init__.py` | `NemesisFlags.full_cluster_restart`; skip protected nodes in node selection |
| `sdcm/utils/nemesis_utils/node_allocator.py` | Skip protected nodes in allocation |
| `sdcm/sct_config.py` | `protected_db_nodes` (list[int]) config field |
| `data_dir/nemesis.yml` | `full_cluster_restart: true/false` on every entry |
| `unit_tests/unit/nemesis/test_node_allocator.py` | Tests for protected node exclusion |
| `unit_tests/unit/test_config.py` | Test for `protected_db_nodes` config parsing |
| `test-cases/longevity/longevity-50GB-3days-authorization-and-tls-ssl.yaml` | Example usage |

## Notes

- Split from #11060. Memory analysis (anomaly detection, teardown validator) is deferred.
- k8s backends explicitly reject `protected_db_nodes` (not supported).

Ref: [SCT-20](https://scylladb.atlassian.net/browse/SCT-20)

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

[SCT-20]: https://scylladb.atlassian.net/browse/SCT-20?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -151,6 +151,11 @@ def str_or_list_or_eval(value: Union[str, List[str], None]) -> List[str] | None:
     raise ValueError(f"{value} isn't a string or a list")
 
 
+def int_list_or_eval(value: Union[str, List[int]]) -> List[int]:
+    values = str_or_list_or_eval(value)
+    return [int(v) for v in values]
+
+
 StringOrList = Annotated[
     str | list[str], BeforeValidator(str_or_list_or_eval), Field(json_schema_extra={"appendable": True})
 ]
@@ -530,6 +535,9 @@ class SCTConfiguration(BaseModel):
     )
     n_monitor_nodes: IntOrList = SctField(
         description="Number list of monitor nodes in multiple data centers",
+    )
+    protected_db_nodes: Annotated[list[int] | None, BeforeValidator(int_list_or_eval)] = SctField(
+        description="List of protected db nodes indexes, that won't be select by nemesis",
     )
     intra_node_comm_public: Boolean = SctField(
         description="If True, all communication between nodes are via public addresses",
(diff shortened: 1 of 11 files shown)


sdcm/sct_config.py
```diff
@@ -531,6 +536,9 @@ class SCTConfiguration(BaseModel):
     n_monitor_nodes: IntOrList = SctField(
         description="Number list of monitor nodes in multiple data centers",
     )
+    protected_db_nodes: Annotated[list[int] | None, BeforeValidator(int_list_or_eval)] = SctField(
```
fruch, line 539: need to use IntOrList, not point of introducing a new type here

test-cases/longevity/longevity-50GB-3days-authorization-and-tls-ssl.yaml
```diff
@@ -11,12 +11,15 @@ availability_zone: 'a,b,c'
 n_db_nodes: 6
 n_loaders: 3
 
+protected_db_nodes: [1,]
```
fruch, line 14: the test configuration change should be on it's own PR, at the end

PR #14451, opened 2026-04-27
feature(logcollector): configurable Grafana panel screenshot capture

fruch:
## Summary

Lets an individual test capture specific Grafana **panels** at log-collection time — on top
of the dashboards SCT always screenshots — configured from the test YAML, with no Python
change needed to add one.

```yaml
grafana_screenshot_panels:
  - dashboard_title: "Detailed"
    panel_title: "LSA total memory"
  - dashboard_title: "Overview"
    panel_title: "Total Disk Usage"
    resolution: [1920, 800]
```

Both titles are substring-matched, and `resolution` defaults to `(1920, 4000)`.

## How it works

1. **`sdcm/sct_config/types.py`** — `GrafanaPanelScreenshot`, a pydantic model with
   `dashboard_title`, `panel_title` and `resolution`. Its `title`/`name` properties make it
   duck-type as a `Dashboard`, so the collector treats it like any other entry in its list.

2. **`sdcm/sct_config/mixins/monitoring.py`** — the new `grafana_screenshot_panels` option.

3. **`sdcm/monitorstack/ui.py`** — `Dashboard` gains `panel_title: str | None = None`, so the
   existing dashboards declare "whole dashboard" explicitly.

4. **`sdcm/logcollector.py`**
   - `MonitoringStack.get_panel_by_title()` — fetches the dashboard JSON and returns the
     numeric panel id whose title contains the configured one, recursing through nested row
     panels (a panel inside a collapsed row is only reachable that way).
   - `GrafanaScreenShot.dashboards` — the static dashboards plus whatever the config adds.
   - `GrafanaScreenShot.get_grafana_screenshot()` — a configured panel renders through
     `d-solo/` with `panelId` and its own height; a whole dashboard keeps rendering at
     `height=-1` as before. A panel that resolves to nothing is logged and skipped, leaving
     the rest of the collection intact.

## Testing

`unit_tests/integration/test_grafana_panel_screenshot.py` runs the collector against a real
scylla-monitoring stack: the `monitor_branch` sources from `defaults/test_default.yaml`,
brought up by their own `start-all.sh` with the Grafana image renderer enabled — the same
stack a monitor node runs. Two panels are configured from a test YAML, one at the top level
of the `Detailed` dashboard and one inside a collapsed row, and the tests assert that each
screenshot comes back as a PNG at the resolution its config asked for, that the default
dashboards are still captured alongside, and that an unresolvable panel title does not take
the rest of the collection down with it.

`unit_tests/unit/test_grafana_panel_screenshot.py` covers the model's validation and slug
building, and the panel lookup's traversal, without a server.

## Split from #11060

Extracts the panel-capture infrastructure from #11060 and generalises it, so any test can
use it and not just the `protected_db_nodes` memory validator.

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -352,6 +352,33 @@ class AdaptiveTimeoutMultipliers(RootModel):
         return float(self.root.get(operation_key, 1.0))
 
 
+class GrafanaPanelScreenshot(BaseModel):
+    """Capture a specific Grafana panel screenshot — no Python code changes needed.
+
+    YAML example::
+
+        grafana_screenshot_panels:
+          - dashboard_title: "Detailed"
+            panel_title: "LSA total memory"
+          - dashboard_title: "Overview"
+            panel_title: "Total Disk Usage"
+            resolution: [1920, 800]
+    """
+
+    dashboard_title: str = Field(description="Grafana dashboard title (substring match used for search)")
+    panel_title: str = Field(description="Panel title within the dashboard (substring match)")
+    resolution: tuple[int, int] = Field(default=(1920, 4000), description="Screenshot resolution (width, height)")
+
+    @property
+    def title(self) -> str:
+        return self.dashboard_title
+
+    @property
+    def name(self) -> str:
+        slug = f"{self.dashboard_title}-{self.panel_title}".lower()
+        return re.sub(r"[^\w\-]", "_", slug)
+
+
 def dict_or_str_or_pydantic(value: dict | str | BaseModel | None) -> dict | BaseModel | None:
     if value is None:
         return None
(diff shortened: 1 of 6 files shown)


sdcm/sct_config.py
```diff
@@ -1893,6 +1893,14 @@ class SCTConfiguration(BaseModel):
     monitor_swap_size: int = SctField(
         description="The size of the swap file for the monitors. Its size in bytes calculated by x * 1MB",
     )
+    grafana_screenshot_panels: StringOrList = SctField(
+        description="""List of Dashboard class names from sdcm.monitorstack.ui to capture at log-collection time,
```
fruch: it should need python code change, all of the params from within Dashboard calls should be exposed in the configuration via pydantic model or something equivlent, use should be able to select to capture any  panel from any dashboard with a specific test

fruch:
@scylladbbot rerun

fruch:
@k0machi ping for review 🙏

Rebased onto current master, no conflicts. Two commits:

- **`feature(logcollector)`** — `grafana_screenshot_panels` lets a test YAML name extra Grafana
  panels to screenshot at log-collection time, on top of the dashboards SCT always captures.
  Dashboard and panel titles are substring-matched, resolution is per-panel.
- **`test(logcollector)`** — integration test that runs the collector against a real
  scylla-monitoring stack in docker (the `monitor_branch` from `defaults/test_default.yaml`,
  its own `start-all.sh`, image renderer enabled) and asserts each panel comes back as a PNG at
  its configured size.

Two things worth a reviewer's eye:

1. Whole-dashboard screenshots deliberately keep `height=-1` (full height) — only a configured
   panel uses its own height. The earlier revision of this branch had changed both, which would
   have clipped the 15000px-tall nemesis dashboard.
2. The integration fixture removes the Grafana + renderer images it pulled (~4GB) on teardown.
   Without that the SCT runner hits ENOSPC and takes the kafka/ycsb tests down with it — that
   was observed and fixed on this branch, see build #10 vs #11.

Last full green run before the rebase: unit 6616 passed, integration 264 passed / 0 failed.

PR #14453, opened 2026-04-27
perf(sct_runner): parallelize clean-runner-instances loop

soyacz:
## Summary

- Extract per-runner logic into `_process_sct_runner()` helper
- Dispatch all runners concurrently via `ThreadPoolExecutor` instead of the previous sequential `for` loop
- Fix a pre-existing `timeout_flag` variable leak across loop iterations (it was function-scoped and could carry a stale value from an AWS runner into a subsequent GCE runner)

## Motivation

The hydra `clean-runner-instances` job processes each SCT runner one-by-one. Each iteration blocks on an SSH round-trip to check for `TestTimeoutEvent` (observed: 10 s – 120 s per runner). With enough runners the job hits its time limit before finishing.

With this change all runners are processed concurrently; total wall-clock time drops from O(N × slowest SSH) to O(slowest single runner).

fixes: QAINFRA-100

### DIFF
--- a/sdcm/sct_runner.py
+++ b/sdcm/sct_runner.py
@@ -21,6 +21,7 @@ import logging
 import random
 import string
 import tempfile
+from concurrent.futures import ThreadPoolExecutor, as_completed
 from contextlib import suppress
 from enum import Enum
 from functools import cached_property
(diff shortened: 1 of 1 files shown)


soyacz:
test run: https://jenkins.scylladb.com/view/QA/job/QA-tools/job/hydra-clean-runner-instances/12714/

fruch (dismissed):

soyacz:
I've added one more fix to this code (to not exceed 32 parallel threads)

fruch:
@soyacz unittest gonna fail, my fault, fixing it

soyacz:
> @soyacz unittest gonna fail, my fault, fixing it

ok, you can merge this one, or I can wait for fix and rebase

PR #14463, opened 2026-04-28
fix(backtrace-decoding): use backtraces.scylladb.com to avoid monitor OOM [SCT-275]

fruch:
## Summary
- Add mini-plan for using `backtraces.scylladb.com` external API to decode backtraces instead of running `addr2line` locally on the monitor node
- Addresses SCT-275: `addr2line` OOM-kills the 8 GB monitor VM during longevity tests (6.6 GB RSS on large debug binaries)
- **Approach:** Try external service first → automatic fallback to local `addr2line` when service unavailable

## Plan Details
See `docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md`

### Key Changes (Planned)
1. New `_decode_via_external_service()` method — POST to `api.backtrace.scylladb.com/api/backtrace` with `build_id` + raw addresses
2. Modify `decode_backtrace()` — try external first, skip debug file copy + local `addr2line` on success
3. Fallback to existing local path on external service failure
4. Unit tests for success/fallback/timeout scenarios

### Files to Modify
- `sdcm/cluster.py` — external service integration + fallback logic
- `unit_tests/unit/test_decode_backtrace.py` — new test cases

**This is a draft PR with the plan only. Implementation to follow.**

### DIFF
--- a/docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
+++ b/docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
@@ -1,0 +1,81 @@
+# Mini-Plan: Use backtraces.scylladb.com for Backtrace Decoding (SCT-275)
+
+**Date:** 2026-04-28
+**Estimated LOC:** ~250
+**Related Issue:** SCT-275
+
+## Problem
+
+During longevity tests (e.g. `longevity-10gb-3h-azure-test`), the `addr2line` process on the
+monitor node consumes ~6.6 GB RSS when decoding Scylla backtraces from 1+ GB debug binaries.
+This OOM-kills the monitor node (8 GB VM), loses monitoring connectivity, and fails the test.
+
+The root cause is `decode_raw_backtrace()` in `sdcm/cluster.py` (line ~2066) which runs
+`addr2line -Cpife <debug_file> <addresses>` locally on the monitor node, loading full DWARF
+debug info into memory for every backtrace.
+
+## Approach
+
+**Strategy:** Try the external `backtraces.scylladb.com` service first, fall back to local
+`addr2line` only when the external service fails.
+
+### Step 1: Add external service helper method
+
+Add `_decode_via_external_service(self, build_id, raw_backtrace)` to the monitor node class
+in `sdcm/cluster.py`:
+- POST to `https://api.backtrace.scylladb.com/api/backtrace` with JSON body
+  `{"build_id": "<hex>", "input": "Backtrace:\n<addresses>"}`
+- Timeout: 120 seconds (decoding large binaries can take 30+ seconds on cache miss)
+- Return `output.stdout` on success (`response["success"] == True`)
+- Raise on failure (404 build not found, network error, `success == false`)
+
+### Step 2: Modify `decode_backtrace()` to try external first
+
+In the `decode_backtrace()` method (line ~1966), which already has `build_id` available from
+`obj["build_id"]`:
+1. If `build_id` is available, try `_decode_via_external_service(build_id, raw_backtrace)`
+2. On success → use the result, **skip** `copy_scylla_debug_info()` and local `addr2line` entirely
+3. On failure → log warning, fall back to existing flow (copy debug file + local `addr2line`)
+4. If `build_id` is `None` → go directly to local fallback
+
+This avoids copying the 1+ GB debug file to the monitor node when external decoding succeeds,
+which is the majority case for official Scylla builds.
+
+### Step 3: Refactor `decode_raw_backtrace()` as local-only fallback
+
+Rename or keep `decode_raw_backtrace()` as the local-only path. The external service path
+bypasses it entirely — no signature change needed for the local method.
+
+### Step 4: Add unit tests
+
+Test the new external service path with mocked HTTP responses:
+- External success → no local `addr2line` called
+- External failure (404) → falls back to local `addr2line`
+- External timeout → falls back to local `addr2line`
+- No `build_id` → goes directly to local `addr2line`
+
+### Design Notes
+
+- **Build ID source:** Already available in the decoding queue object (`obj["build_id"]`),
+  extracted from logs by `DbLogReader` (see `sdcm/db_log_reader.py` line ~157).
+- **Input format:** The API requires `"Backtrace:\n"` prefix before raw addresses.
+  Addresses can be raw hex (`0x12f34`) or path+offset format.
+- **Response format:** `{"success": true, "stdout": "<decoded lines>", "stderr": "", "duration": 5.2}`
+- **Existing API usage:** `sdcm/utils/version_utils.py` already calls the same service
+  (`/api/search/build_id`) — reuse similar HTTP patterns.
+- **No new config params needed initially.** The external service is always tried first;
+  local fallback is automatic. Can add a config toggle later if needed.
+
+## Files to Modify
+
+- `sdcm/cluster.py` — Add `_decode_via_external_service()`, modify `decode_backtrace()` to try external first
+- `unit_tests/unit/test_decode_backtrace.py` — Add tests for external service path and fallback logic
+
+## Verification
+
+- [ ] Unit tests pass: `uv run python -m pytest unit_tests/unit/test_decode_backtrace.py -v`
+- [ ] External service success path skips `copy_scylla_debug_info` and local `addr2line`
+- [ ] Fallback to local `addr2line` works when external service returns 404
+- [ ] Fallback to local `addr2line` works when external service times out
+- [ ] No `build_id` case goes directly to local path without attempting external call
+- [ ] `uv run sct.py pre-commit` passes
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -2024,7 +2024,7 @@ class BaseNode(AutoSshContainerMixin):
 
                 if decoded is None:
                     scylla_debug_file = self.copy_scylla_debug_info(obj["node"], build_id)
-                    decoded = self.decode_raw_backtrace(scylla_debug_file, raw_backtrace_oneline).stdout
+                    decoded = self.decode_backtrace_local(scylla_debug_file, raw_backtrace_oneline).stdout
 
                 event.backtrace = decoded
                 the_map = FindIssuePerBacktrace()
@@ -2116,7 +2116,7 @@ class BaseNode(AutoSshContainerMixin):
         raise Exception("Couldn't find scylla debug information")
 
     @lru_cache(maxsize=None)
-    def decode_raw_backtrace(self, scylla_debug_file, raw_backtrace):
+    def decode_backtrace_local(self, scylla_debug_file, raw_backtrace):
         """run decode backtrace on monitor node
 
         Decode backtrace on monitor node
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -1974,6 +1974,33 @@ class BaseNode(AutoSshContainerMixin):
         self._decoding_backtraces_thread.daemon = True
         self._decoding_backtraces_thread.start()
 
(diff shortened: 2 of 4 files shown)


sdcm/cluster.py
```diff
@@ -1974,6 +1974,33 @@ def start_decode_on_monitor_node_thread(self):
         self._decoding_backtraces_thread.daemon = True
         self._decoding_backtraces_thread.start()
 
+    def _decode_via_external_service(self, build_id: str, raw_backtrace: str) -> str:
```
fruch, line 1978: we should have integration test for this function, to check the actual service is working

sdcm/cluster.py
```diff
@@ -1974,6 +1974,33 @@ def start_decode_on_monitor_node_thread(self):
...
+        response = requests.post(
+            "https://api.backtrace.scylladb.com/api/backtrace",
+            json={"build_id": build_id, "input": "Backtrace:\n" + raw_backtrace},
+            timeout=120,
+        )
+        response.raise_for_status()
+        result = response.json()
+        if not result.get("success"):
+            raise ValueError(f"External backtrace service failed: {result.get('stderr', 'unknown error')}")
+        return result["stdout"]
```
fruch, line 2003: we need to cache this heavily, to avoid repetition

sdcm/cluster.py
```diff
@@ -1983,9 +2010,22 @@ def decode_backtrace(self):
...
+                if build_id:
+                    try:
+                        decoded = self._decode_via_external_service(build_id, event.raw_backtrace)
+                        self.log.debug("Decoded backtrace via external service for build_id=%s", build_id)
+                    except Exception as exc:  # noqa: BLE001
+                        self.log.warning("External backtrace service failed (%s), falling back to local addr2line", exc)
+
+                if decoded is None:
+                    scylla_debug_file = self.copy_scylla_debug_info(obj["node"], build_id)
+                    decoded = self.decode_raw_backtrace(scylla_debug_file, raw_backtrace_oneline).stdout
```
fruch: if we are it decode_raw_backtrace should be cached heavily as well (if it's not already)

sdcm/cluster.py
fruch: Added integration test in `unit_tests/integration/test_external_backtrace_service.py` (commit 0f97fd4f3).

Tests the live `api.backtrace.scylladb.com` service with a known build ID from SCYLLADB-403 (`527bc25417c81c8106b33400a4372616c7c90894`). Verifies:
1. Service returns decoded symbols for known build + addresses
2. Service handles unknown build IDs gracefully

Both tests pass (1.9s total).

fruch:
@pehala one open question, would `scylla_s3_reloc_server` will be  o.k. with the extra load ?

pehala:
> @pehala one open question, would `scylla_s3_reloc_server` will be o.k. with the extra load ?

I would like it to be, but it wasnt tested yet, so I would give it a try and if we overload it we can fix it.

docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
```diff
@@ -0,0 +1,81 @@
...
+**Related Issue:** SCT-275
+
+## Problem
+
+During longevity tests (e.g. `longevity-10gb-3h-azure-test`), the `addr2line` process on the
+monitor node consumes ~6.6 GB RSS when decoding Scylla backtraces from 1+ GB debug binaries.
+This OOM-kills the monitor node (8 GB VM), loses monitoring connectivity, and fails the test.
+
+The root cause is `decode_raw_backtrace()` in `sdcm/cluster.py` (line ~2066) which runs
+`addr2line -Cpife <debug_file> <addresses>` locally on the monitor node, loading full DWARF
```
pehala, line 14: Scylla should be using llvm addr2line, maybe that is the cause of difference between backtrace and SCT.

docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
```diff
@@ -0,0 +1,81 @@
...
+2. On success → use the result, **skip** `copy_scylla_debug_info()` and local `addr2line` entirely
+3. On failure → log warning, fall back to existing flow (copy debug file + local `addr2line`)
+4. If `build_id` is `None` → go directly to local fallback
+
+This avoids copying the 1+ GB debug file to the monitor node when external decoding succeeds,
+which is the majority case for official Scylla builds.
+
+### Step 3: Refactor `decode_raw_backtrace()` as local-only fallback
+
+Rename or keep `decode_raw_backtrace()` as the local-only path. The external service path
```
pehala, line 46: I would rename to `decode_backtrace_local()`

docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
```diff
@@ -0,0 +1,81 @@
...
+- External failure (404) → falls back to local `addr2line`
+- External timeout → falls back to local `addr2line`
+- No `build_id` → goes directly to local `addr2line`
+
+### Design Notes
+
+- **Build ID source:** Already available in the decoding queue object (`obj["build_id"]`),
+  extracted from logs by `DbLogReader` (see `sdcm/db_log_reader.py` line ~157).
+- **Input format:** The API requires `"Backtrace:\n"` prefix before raw addresses.
+  Addresses can be raw hex (`0x12f34`) or path+offset format.
```
pehala, line 62: It can contain rest of the log line as well, if you just take the entire log line, it should correctly decode it.

+ It can decode multiple log lines at once

sdcm/cluster.py
```diff
@@ -1974,6 +1974,34 @@ def start_decode_on_monitor_node_thread(self):
...
+        Returns:
+            decoded backtrace string (stdout from the service)
+
+        Raises:
+            requests.RequestException: on network/HTTP error
+            ValueError: if the service returns success=false
+        """
+        response = requests.post(
+            "https://api.backtrace.scylladb.com/api/backtrace",
+            json={"build_id": build_id, "input": "Backtrace:\n" + raw_backtrace},
```
pehala, line 1996: I would simply put the entire log line, it should be in the necessary format. If it isnt, we can extend backtrace to support them

unit_tests/integration/test_external_backtrace_service.py
```diff
@@ -0,0 +1,55 @@
...
+@pytest.fixture(name="node")
+def node_fixture(tmp_path):
+    node = DummyNode(name="test_node", parent_cluster=None, base_logdir=tmp_path)
+    node.remoter = DummyRemote()
+    return node
+
+
+def test_decode_via_external_service_returns_symbols(node):
+    """Verify _decode_via_external_service returns decoded symbols for a known build."""
+    result = node._decode_via_external_service(KNOWN_BUILD_ID, SAMPLE_BACKTRACE_ADDRESSES)
```
pehala, line 42: Lets not call decode in our tests, this is too heavy for something executed on each PR. Also unnecessary because just symbols dont tell you anything.

unit_tests/unit/test_decode_backtrace.py
```diff
@@ -219,3 +223,74 @@ def test_backtrace_decoding_configuration(
...
+
+    config.DECODING_QUEUE.put({"event": event, "node": "test_node", "build_id": build_id})
+    config.DECODING_QUEUE.put(None)
+
+    monitor_node.test_config = config
+
+    t = threading.Thread(target=monitor_node.decode_backtrace)
+    t.start()
+    t.join(timeout=10)
+    return event
```
pehala: Why do we need the thread here?

unit_tests/unit/test_decode_backtrace.py
```diff
@@ -219,3 +223,74 @@ def test_backtrace_decoding_configuration(
...
+    assert "addr2line" in event.backtrace
+
+
+def test_external_service_timeout_falls_back_to_local(monitor_node):
+    """When external service times out, fall back to local addr2line."""
+    with patch("sdcm.cluster.requests.post", side_effect=requests.Timeout("timed out")):
+        event = _run_decode_with_queue_item(monitor_node, "abc123", "0x1234\n0x5678")
+
+    assert event.backtrace is not None
+    assert "addr2line" in event.backtrace
```
pehala: parametrize

unit_tests/unit/test_decode_backtrace.py
pehala, line 1: I am missing unit test for successful decode here. I think we should test what POST request it sent

unit_tests/integration/test_external_backtrace_service.py
fruch: it's integration tests

unit_tests/unit/test_decode_backtrace.py
fruch: it's a full copy of how the system works. yes, it's not needed for a unit test

docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
fruch: I doesn't have any added value for us to send the whole log line, 
we already extracted the data needed to be decoded

sdcm/cluster.py
fruch: I doesn't have any added value for us to send the whole log line,
we already extracted the data needed to be decoded

fruch:
> > @pehala one open question, would `scylla_s3_reloc_server` will be o.k. with the extra load ?
> 
> I would like it to be, but it wasnt tested yet, so I would give it a try and if we overload it we can fix it.

any particular way you think to test it upfront, or we'll merge it and we'll see how it works ?

docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
fruch: this how it was set years ago... it worth trying, but not sure we should mix it here..

docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
fruch: Done — renamed to `decode_backtrace_local()` in commit db1cc01.

unit_tests/unit/test_decode_backtrace.py
fruch: Done — removed the thread, calling `decode_backtrace()` directly in the test helper.

unit_tests/unit/test_decode_backtrace.py
fruch: Done — parametrized into a single test covering HTTP errors, timeouts, and connection errors.

unit_tests/unit/test_decode_backtrace.py
fruch: Done — added `test_external_service_post_request_payload` that asserts the exact URL, JSON body (build_id + input format), and timeout.

pehala:
> any particular way you think to test it upfront, or we'll merge it and we'll see how it works ?

merge it and see

fruch:
> > any particular way you think to test it upfront, or we'll merge it and we'll see how it works ?
> 
> merge it and see

Give us a long is it's killing the service, and will revert right away.

Any other comments need clearing up ?

sdcm/cluster.py
pehala: I think if you could send the entire log line it could simplify the procedure, but it is not neccessary

docs/plans/mini-plans/2026-04-28-sct-275-external-backtrace-decoding.md
pehala: No action required. When reading it I just realized it might be connected to https://scylladb.atlassian.net/browse/SCT-191, which is going to be fixed by this patch as well

pehala (approved):

PR #14477, opened 2026-04-29
feat(pre-commit): add lint-pipelines hook for pipeline and config changes

pehala:
Runs lint-pipelines automatically when jenkinsfiles, test-cases, configurations, or defaults YAML files are modified. We run it on CI, we should run it locally

### Testing

- [ ] local

### Reminders

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -98,6 +98,13 @@ repos:
         pass_filenames: false
         files: ^pyproject.toml$
 
+    -   id: lint-pipelines
+        name: lint-pipelines
+        entry: ./sct.py lint-pipelines
+        language: system
+        pass_filenames: false
+        files: (?x)(^jenkins-pipelines/.*\.jenkinsfile$|^test-cases/.*\.yaml$|^configurations/.*\.yaml$|^defaults/.*\.yaml$)
+
 - repo: https://github.com/alessandrojcm/commitlint-pre-commit-hook
   rev: v9.22.0
   hooks:


pehala:
Problems:
*  qa-internal repository is needed, and it failed to fetch automatically for me.
    * As I suggested before, I think long term yamls/pipelines which require internal data should be all in scylla-qa-internal and should not leak here - this is not view that many share, so wont be done
    * For local check I think we can exclude all pipelines which fail due to scylla-qa-internal and check these only on CI, @fruch WDYT?

fruch:
> Problems:
> 
> * qa-internal repository is needed, and it failed to fetch automatically for me.
>   
>   * As I suggested before, I think long term yamls/pipelines which require internal data should be all in scylla-qa-internal and should not leak here - this is not view that many share, so wont be done
>   * For local check I think we can exclude all pipelines which fail due to scylla-qa-internal and check these only on CI, @fruch WDYT?

why not fixing you environment that fetch of it would work ?

fruch:
> > Problems:
> > 
> > * qa-internal repository is needed, and it failed to fetch automatically for me.
> >   
> >   * As I suggested before, I think long term yamls/pipelines which require internal data should be all in scylla-qa-internal and should not leak here - this is not view that many share, so wont be done
> >   * For local check I think we can exclude all pipelines which fail due to scylla-qa-internal and check these only on CI, @fruch WDYT?
> 
> why not fixing you environment that fetch of it would work ?

in CI, just need to copy/move `checkoutQaInternal(params)` line into pre-commit stage

pehala:
> why not fixing you environment that fetch of it would work ?

This is a public repository, so the general precommit shouldnt rely on private scylla-qa-internal. Yes, I can fix the local fetch, but conceptually it would be still wrong. This is why I prefer moving the pipelines away from this repository if they rely on private configs. This will also solve the problem of double repository i.e. by removing a config in scylla-qa-internal, all checks pass there (as there arent any), but it will fail all the jobs on master.

fruch:
> > why not fixing you environment that fetch of it would work ?
> 
> This is a public repository, so the general precommit shouldnt rely on private scylla-qa-internal. Yes, I can fix the local fetch, but conceptually it would be still wrong. This is why I prefer moving the pipelines away from this repository if they rely on private configs. This will also solve the problem of double repository i.e. by removing a config in scylla-qa-internal, all checks pass there (as there arent any), but it will fail all the jobs on master.

so don't put it into pre-commit.
I don't agree that SCT pipeline should move out of the is repo, it was done once, and it was a mistake.

pehala:
> so don't put it into pre-commit.
> I don't agree that SCT pipeline should move out of the is repo, it was done once, and it was a mistake.

That makes it that you no longer can run all of the checks locally, which makes precommit less useful. We shouldnt rely on CI, thats what precommit is for.

I am interested why it caused problems in the past, maybe there is a good reason to not do it, that I am just not seeing

fruch:
> > so don't put it into pre-commit.
> > I don't agree that SCT pipeline should move out of the is repo, it was done once, and it was a mistake.
> 
> That makes it that you no longer can run all of the checks locally, which makes precommit less useful. We shouldnt rely on CI, thats what precommit is for.
> 
> I am interested why it caused problems in the past, maybe there is a good reason to not do it, that I am just not seeing

splitting the code, even if it's groovy code across repos, is repice for trouble, and complicate any changes one might need to to make to pipelines or configuration files.

cezarmoise:
> Problems:
> 
> * qa-internal repository is needed, and it failed to fetch automatically for me.
>   

This is because it tries to delete the existing clone with sudo, even if not necessary.
fix: https://github.com/scylladb/scylla-cluster-tests/pull/14676

cezarmoise:
How about a compromise solution:
- if the qa-internal folder exists, then check it
- if it doesn't, assume files with that path exist

This way, the complete check will always be run on CI checks, and users don't get nagged every time. But those who change related pipelines and configs will most likely have the repo cloned.

I don't agree with splitting pipeline stuff across repos.

PR #14485, opened 2026-05-03
Add investigating-perf-regression skill

juliayakovlev:
## Summary

- Add a new AI agent skill for systematically investigating SCT performance regression test failures
- Covers the full investigation workflow: Argus CLI for metadata/logs, Prometheus queries for metrics, HDR histogram analysis, and root cause documentation
- Includes reference docs for Prometheus query syntax and Argus CLI commands

## Skill Structure

```
skills/investigating-perf-regression/
├── SKILL.md                                    # Main skill definition
├── references/
│   ├── prometheus-queries.md                   # Exact PromQL queries for key metrics
│   └── argus-commands.md                       # Argus CLI command reference
└── workflows/
    └── investigate-perf-regression.md           # 6-phase investigation process
```

## Motivation

This skill codifies the investigation process used to diagnose a real performance regression failure (P99 read latency spike from 5ms to 1754ms at 1.2M op/s). The root cause was traced to read queuing on a single node caused by memory pressure and cache evictions, identified through systematic Prometheus metric comparison and Scylla log analysis.

The skill enables AI agents to reproduce this investigation process for future performance regression failures.

### DIFF
--- a/skills/investigating-perf-regression/references/argus-commands.md
+++ b/skills/investigating-perf-regression/references/argus-commands.md
@@ -1,0 +1,162 @@
+# Argus CLI Commands for Performance Investigation
+
+Commands for retrieving SCT test run metadata, events, and logs using the Argus CLI.
+
+## Prerequisites
+
+The `argus` CLI must be installed and authenticated:
+
+```bash
+which argus          # verify installation
+argus auth           # authenticate via Cloudflare Access (if needed)
+argus auth-token     # save token to keyring (alternative)
+```
+
+## Run Metadata
+
+### Get run details
+
+Returns JSON with run ID, region, instance type, Scylla version, SCM revision, AMI, status, Jenkins URL, and events summary.
+
+```bash
+argus run get --run-id <UUID>
+```
+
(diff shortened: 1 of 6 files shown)


skills/investigating-perf-regression/references/argus-commands.md
```diff
@@ -0,0 +1,162 @@
+# Argus CLI Commands for Performance Investigation
+
+Commands for retrieving SCT test run metadata, events, and logs using the Argus CLI.
+
+## Prerequisites
+
+The `argus` CLI must be installed and authenticated:
```
pehala, line 7: How do you do install it? I dont think we have it in setup anywhere

fruch: this is targeted later for zeus, which has access to those.
meanwhile one can get it and build from directly from argus code base. @soyacz team would have it documented / released under argus, at some point.

also, this PR a draft for a reason, not point reviewing it at this point.

juliayakovlev:
This skill is under developing in Zeus. Closing

PR #14486, opened 2026-05-03
fix(gce-provision): align cloud-init SSH timeout to 300s

fruch:
## Summary
- Add `remoter.is_up(timeout=300)` before the `command -v cloud-init` guard in both `cluster_gce.py` and `cluster_aws.py`
- Aligns the SSH timeout with what `wait_cloud_init_completes()` uses internally
- Fixes intermittent `SSHConnectTimeoutError` during GCE provisioning when metadata service is slow (observed 273s on affected nodes)

Fixes: SCT-282


fruch:
recheck

juliayakovlev (approved):

PR #14490, opened 2026-05-03
docs(cross-cloud-sizing): cross-cloud instance sizing abstraction plan

fruch:
## Summary

Introduces a **T-shirt sizing abstraction** for cross-cloud instance types in SCT. Instead of specifying cloud-specific instance types for every backend in every test YAML, test authors can now write a single abstract size that resolves automatically per backend.

### Before (6+ lines per test config)
```yaml
gce_instance_type_db: 'z3-highmem-16'
gce_n_local_ssd_disk_db: 4
azure_instance_type_db: 'Standard_L16s_v4'
oci_instance_type_db: 'DenseIO.E5.Flex:8:128'
```

### After (1 line, all clouds resolved automatically)
```yaml
db_instance_type: '2xlarge'
```

Cloud-specific overrides still take full precedence — zero breaking changes.

---

### Phase 1 — `sdcm/utils/cloud_sizes.py` (commit `dd2e5c4330`)
- New module with `InstanceSpec`, `SizeMapping`, `resolve_size()`, `identify_size()`, `get_cloud_params()`
- Full mapping tables: DB (6 sizes × 4 clouds), Loader (4 sizes × 4 clouds), Monitor (3 sizes × 4 clouds)
- AWS DB defaults to `i8g` (Graviton/ARM); auto-switches to `i4i` for x86
- GCE DB: `z3-highmem` family with local NVMe; Azure: `Standard_L*s_v4`; OCI: `DenseIO.E5`
- 110 parametrized unit tests in `unit_tests/unit/test_cloud_sizes.py`

### Phase 2 — Config integration in `sdcm/sct_config.py` (commit `3df56fb701`)
- Three new `SctField` params: `db_instance_type`, `loader_instance_type`, `monitor_instance_type`
- New `_resolve_instance_sizes()` method called early in `__init__` (before AWS arch/AMI logic)
- `_user_provided_keys` set tracks explicitly-set keys so defaults don't block abstract resolution
- GCE resolution also populates `gce_n_local_ssd_disk_db` and `gce_root_disk_type_db`
- 3 new unit tests in `unit_tests/unit/test_config.py`; all 64 existing tests still pass
- `docs/configuration_options.md` auto-updated by pre-commit hook

### Phase 3 — CLI helpers in `sct.py` (commit `891f6e8492`)
- `sct.py show-sizes [--role db|loader|monitor] [--cloud aws|gce|azure|oci]` — print mapping table
- `sct.py translate-size <instance_type>` — show cross-cloud equivalents for a concrete type
- `sct.py translate-size --config <yaml> [--backend <cloud>]` — expand abstract sizes in a YAML
- Test fixture `unit_tests/test_data/test_abstract_sizing.yaml`

### Phase 4 — Migrate longevity configs (commit `d791aba046`)
- 12 longevity configs in `test-cases/longevity/` migrated to abstract sizes
- Configs using `4xlarge`, `8xlarge`, `16xlarge` DB sizes; `medium`/`large` loader/monitor sizes
- All migrated configs verified to resolve correctly on AWS and GCE

### Phase 5 — Documentation (commit `afb4e505bb`)
- `docs/cross-cloud-sizing.md`: full usage guide, mapping table, how to add new sizes
- `AGENTS.md`: cross-cloud sizing reference added to configuration section

### Fix — Test isolation (commit `e8be7768d9`)
- Fixed module-scoped env leak in `test_config.py`: `fixture_conf` sets `SCT_INSTANCE_TYPE_DB=i4i.large` for the whole module; new sizing tests now explicitly clear it via `monkeypatch.delenv`

---

## Testing

```bash
# All unit tests pass
uv run python -m pytest unit_tests/unit/test_cloud_sizes.py unit_tests/unit/test_config.py -q
# 174 passed

# CLI smoke tests
uv run sct.py show-sizes --role db --cloud aws
uv run sct.py translate-size i8g.2xlarge
uv run sct.py translate-size --config unit_tests/test_data/test_abstract_sizing.yaml --backend gce
```

## Manual Testing Required

- [ ] Run a longevity test on GCE with a migrated config to confirm resolved instance type is correct
- [ ] Verify `sct.py conf -b gce test-cases/longevity/longevity-100gb-4h.yaml` resolves `gce_instance_type_db` correctly (requires network access to SCT config resolution)

## Backends Affected
- [x] AWS — `instance_type_db` resolution
- [x] GCE — `gce_instance_type_db` + disk params resolution
- [x] Azure — `azure_instance_type_db` resolution
- [x] OCI — `oci_instance_type_db` resolution

### DIFF
--- a/.sisyphus/plans/constraint-based-sizing.md
+++ b/.sisyphus/plans/constraint-based-sizing.md
@@ -1,0 +1,877 @@
+# Constraint-Based Instance Sizing
+
+## TL;DR
+
+> **Quick Summary**: Replace T-shirt sizing (`db_instance_type: '2xlarge'`) with constraint-based selection (`db_instance_type: {vcpu: 16, memory: ">20gb"}`). An instance catalog per cloud (populated from APIs + curated) is matched against user constraints, selecting the best instance from preferred families by price.
+>
+> **Deliverables**:
+> - Instance catalog module with per-cloud type specs (vcpu/mem/disk/price)
+> - Catalog generation script querying AWS/GCE/Azure/OCI pricing APIs
+> - Constraint parser supporting exact, range, and comparison operators
+> - Matcher selecting best instance per role/cloud from catalog
+> - Updated `sct_config.py` wiring (backward-compat: literal strings still passthrough)
+> - CLI: `sct.py show-prices`, updated `show-sizes`/`translate-size`
+> - Migration of existing test configs
+> - User-facing documentation
+>
+> **Estimated Effort**: Large
+> **Parallel Execution**: YES — 4 waves
+> **Critical Path**: Task 1 → Task 3 → Task 6 → Task 9 → Task 11 → Final
+
+---
+
+## Context
+
+### Original Request
+Replace the T-shirt sizing system (roydahan/pehala review feedback on PR #14490) with a constraint-based selector where users specify requirements (vcpu, memory, disk) and the system resolves to the best matching instance type per cloud provider.
+
+### Interview Summary
+**Key Discussions**:
+- T-shirt sizes removed entirely; constraint-based only
+- YAML dict syntax in configs, flat string for env vars (`SCT_DB_INSTANCE_TYPE='vcpu=16,memory=>20gb'`)
+- Catalog: generated from live cloud APIs + manual curations, current families only
+- Selection: preferred family per role (price-based), with CLI price display
+- Constraints: exact (`vcpu: 16`), ranges (`memory: "10gb-30gb"`), comparisons (`memory: ">20gb"`), arch (`arch: "arm64"` or `arch: "x86_64"` — optional, overrides cloud default)
+
+**Research Findings**:
+- Existing `InstanceSpec` already has vcpus, memory_gb, local_ssd_count
+- `_resolve_instance_sizes()` in sct_config.py handles resolution at config-load time
+- PR #14490 has 3 commits implementing T-shirt system (to be replaced)
+- AWS Pricing API accessible from us-east-1; GCE uses `gcloud compute machine-types list`; Azure has Retail Prices REST API
+
+### Metis Review
+**Identified Gaps** (addressed):
+- Backward compat for literal strings: added as passthrough detection
+- Docker/baremetal backends: silently skip constraint resolution
+- No-match error handling: explicit error message naming unsatisfied constraint
+- Multi-role (oracle/zero_token): all roles get constraint syntax
+- Determinism: same constraints + catalog → always same instance (no randomness)
+- Local disk as implicit constraint for db role: YES, db role implies `local_disk: true`
+- Tie-breaking: preferred family first, then cheapest within family
+
+---
+
+## Work Objectives
+
+### Core Objective
+Enable cloud-agnostic instance selection via hardware constraints (vcpu, memory, disk) instead of arbitrary T-shirt size names, with automated catalog generation and price-based selection.
+
+### Concrete Deliverables
+- `sdcm/utils/instance_catalog.py` — catalog data model + loader
+- `sdcm/utils/instance_matcher.py` — constraint parser + matching engine
+- `sdcm/utils/catalog_generator.py` — script to populate catalog from cloud APIs
+- `data/instance_catalog/` — generated + curated catalog files (YAML per cloud)
+- Updated `sdcm/sct_config.py` — wiring constraint resolution
+- Updated `sct.py` — CLI commands
+- Updated test configs — migrated from T-shirt to constraints
+- `docs/cross-cloud-sizing.md` — updated user guide
+
+### Definition of Done
+- [ ] `{vcpu: 8, memory: ">16gb"}` on AWS resolves to `i8g.2xlarge` (ARM preferred) and all downstream works
+- [ ] `SCT_DB_INSTANCE_TYPE='vcpu=16,memory=>20gb'` env var parses correctly
+- [ ] `SCT_DB_INSTANCE_TYPE='vcpu=8,arch=x86_64'` forces x86 on AWS (overrides arm64 default)
+- [ ] `instance_type_db: 'i4i.large'` still works (literal passthrough)
+- [ ] `sct.py show-prices --cloud aws --role db` shows pricing table
+- [ ] 100% unit test coverage on matcher logic
+- [ ] Integration tests verify API-based catalog generation
+
+### Must Have
+- Backward compatibility: literal instance type strings still work as passthrough
+- All 5 roles supported: db, db_oracle, zero_token, loader, monitor
+- Constraint types: exact value, min (>), max (<), range (min-max), arch selector (arm64|x86_64)
+- Deterministic selection (same input → same output)
+- Docker/baremetal backends silently skip resolution
+- Clear error messages when no instance matches
+
+### Must NOT Have (Guardrails)
+- No spot/preemptible pricing — on-demand only
+- No cross-region auto-selection — region is fixed by other config
+- No GPU/accelerator constraints
+- No auto-detection of instance availability in region (future work)
+- No new cloud API dependencies in test runtime hot path — catalog loaded at startup only
+- No changes to downstream provisioning code — output is still a string instance type
+
+---
+
+## Verification Strategy
+
+> **ZERO HUMAN INTERVENTION** — ALL verification is agent-executed.
+
+### Test Decision
+- **Infrastructure exists**: YES (pytest + unit_tests/)
+- **Automated tests**: TDD
+- **Framework**: pytest (existing)
+- **Split**: Unit tests for parsing/matching logic, integration tests for API catalog generation
+
+### QA Policy
+Every task includes agent-executed QA scenarios.
+Evidence saved to `.sisyphus/evidence/task-{N}-{scenario-slug}.{ext}`.
+
+- **Library/Module**: Use Bash (python REPL) — import, call functions, compare output
+- **CLI**: Use interactive_bash (tmux) — run command, validate output
+- **API/Backend**: Use Bash (curl/python) — call cloud APIs, verify catalog generation
+
+---
+
+## Execution Strategy
+
+### Parallel Execution Waves
+
+```
+Wave 1 (Foundation — no dependencies):
+├── Task 1: Instance catalog data model [quick]
+├── Task 2: Constraint parser module [deep]
+├── Task 3: Catalog file format + loader [quick]
+├── Task 4: Preferred families config [quick]
+
+Wave 2 (Core logic — depends on Wave 1):
+├── Task 5: Matching engine (depends: 1, 2, 4) [deep]
+├── Task 6: Catalog generator — AWS (depends: 1, 3) [unspecified-high]
+├── Task 7: Catalog generator — GCE (depends: 1, 3) [unspecified-high]
+├── Task 8: Catalog generator — Azure + OCI (depends: 1, 3) [unspecified-high]
+
+Wave 3 (Integration — depends on Wave 2):
+├── Task 9: Wire into sct_config.py (depends: 5) [deep]
+├── Task 10: CLI commands (depends: 5, 6-8) [unspecified-high]
+├── Task 11: Migration of test configs (depends: 9) [quick]
+
+Wave 4 (Docs + cleanup):
+├── Task 12: User documentation (depends: 9, 10) [writing]
+├── Task 13: Remove old T-shirt sizing code (depends: 11) [quick]
+
+Wave FINAL (4 parallel reviews):
+├── F1: Plan compliance audit (oracle)
+├── F2: Code quality review (unspecified-high)
+├── F3: Real QA execution (unspecified-high)
+├── F4: Scope fidelity check (deep)
+→ Present results → Get explicit user okay
+```
+
+### Dependency Matrix
+
+| Task | Depends On | Blocks | Wave |
+|------|-----------|--------|------|
+| 1 | — | 5, 6, 7, 8 | 1 |
+| 2 | — | 5 | 1 |
+| 3 | — | 6, 7, 8 | 1 |
+| 4 | — | 5 | 1 |
+| 5 | 1, 2, 4 | 9, 10 | 2 |
+| 6 | 1, 3 | 10 | 2 |
+| 7 | 1, 3 | 10 | 2 |
+| 8 | 1, 3 | 10 | 2 |
+| 9 | 5 | 11 | 3 |
+| 10 | 5, 6-8 | 12 | 3 |
+| 11 | 9 | 12, 13 | 3 |
+| 12 | 9, 10 | — | 4 |
+| 13 | 11 | — | 4 |
+
+### Agent Dispatch Summary
+
+- **Wave 1**: 4 tasks — T1,T3,T4 → `quick`, T2 → `deep`
+- **Wave 2**: 4 tasks — T5 → `deep`, T6-T8 → `unspecified-high`
+- **Wave 3**: 3 tasks — T9 → `deep`, T10 → `unspecified-high`, T11 → `quick`
+- **Wave 4**: 2 tasks — T12 → `writing`, T13 → `quick`
+- **FINAL**: 4 tasks — F1 → `oracle`, F2-F3 → `unspecified-high`, F4 → `deep`
+
+---
+
+## TODOs
+
+- [ ] 1. Instance catalog data model
+
+  **What to do**:
+  - Create `sdcm/utils/instance_catalog.py`
+  - Define `@dataclass InstanceTypeInfo`: instance_type (str), cloud (str), family (str), vcpus (int), memory_gb (float), local_disk_gb (float), local_disk_count (int), arch (str: "x86_64"|"arm64"), price_per_hour (float|None)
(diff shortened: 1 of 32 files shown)


.sisyphus/plans/cross-cloud-sizing.md
fruch, line 1: this should be only a local symlink, not to take out of the PR

.sisyphus/plans/cross-cloud-sizing.md
```diff
@@ -0,0 +1,644 @@
...
+   replaces that approach entirely.
+
+### Demo: How It Works
+
+A typical longevity test YAML today (6+ lines of cloud-specific sizing):
+
+```yaml
+# Current: must specify every cloud explicitly
+instance_type_db: 'i4i.2xlarge'              # AWS
+gce_instance_type_db: 'z3-highmem-16'        # GCE
```
fruch: across the board should use `z3-highmem-16-highlssd`, only the highlssd ones

I think when using z3 it should be
```
gce_n_local_ssd_disk_db: 0
```

fruch:
## Code Review — Cross-Cloud Instance Sizing Abstraction

Reviewed commits: `b8ac5d7402`..`74644e855c` (6 commits, plan squashed, gce-provision removed, `.sisyphus/` excluded).

---

### Summary

The overall approach is sound and well-structured. The T-shirt sizing model is a genuine improvement over the current per-cloud duplication. The mapping module is clean, the resolution logic is correct in principle, and the CLI tools are useful. A few issues need attention before merge.

---

### 🔴 Bugs / Must Fix

**1. `_resolve_instance_sizes()` always uses ARM arch — x86 AMIs will get wrong instance type**

`sdcm/sct_config.py` L3087:
```python
spec = resolve_size(role, abstract_size, cloud)  # arch defaults to "arm"
```

The `arch` parameter is never passed. For AWS, this always resolves to `i8g` (Graviton). If a test uses an x86 AMI (e.g. `ami_id_db_scylla` pointing to an x86 image), the resolved `instance_type_db` will be `i8g.*` — an ARM instance — which will fail to launch.

The existing arch detection logic is in `sct_config.py` around L3351 (`get_arch_from_instance_type` / `_get_normalized_arch`). The fix is to detect arch from the AMI/image config before calling `resolve_size`, or default to `"x86"` for safety (since most existing AMIs are x86). The plan explicitly called for this wiring:

> *"The caller (`_resolve_instance_sizes` in `sct_config.py`) determines arch from the AMI/image configuration using the existing `get_arch_from_instance_type()` / `_get_normalized_arch()` logic"*

This is the most critical bug — it will silently produce wrong instance types for x86 workloads.

---

**2. Silent `except (KeyError, ValueError): continue` swallows invalid size names**

`sdcm/sct_config.py` L3088-3089:
```python
except (KeyError, ValueError):
    continue
```

If a user writes `db_instance_type: '3xlarge'` (a typo — not a valid size), the resolution silently does nothing and the config falls through to whatever default is set. The user gets no feedback that their abstract size was ignored.

Should at minimum log a warning:
```python
except (KeyError, ValueError) as exc:
    LOGGER.warning("Cannot resolve abstract size %r for role %r on %r: %s", abstract_size, role, cloud, exc)
    continue
```

---

### 🟡 Issues / Should Fix

**3. Loader `xlarge` has a 4× vCPU jump from `large`**

`sdcm/utils/cloud_sizes.py` L158-164:
```
loader large  → c6i.4xlarge  (16 vCPU)
loader xlarge → c6i.16xlarge (64 vCPU)   ← 4× jump, no 8xlarge in between
```

The gap between `large` (16 vCPU) and `xlarge` (64 vCPU) is unusually large. Consider adding `2xlarge` → `c6i.8xlarge` (32 vCPU) to make the progression consistent with the DB role. The GCE side also jumps from `e2-standard-16` to `e2-highcpu-32` which changes the memory ratio.

**4. Azure `8xlarge` and `16xlarge` vCPU counts don't match AWS**

`sdcm/utils/cloud_sizes.py` L116-129:
```
8xlarge:  AWS i8g.8xlarge = 32 vCPU,  Azure Standard_L64s_v4 = 64 vCPU  ← 2× mismatch
16xlarge: AWS i8g.16xlarge = 64 vCPU, Azure Standard_L80s_v4 = 80 vCPU  ← mismatch
```

The Azure `Standard_L*s_v4` family doesn't have a 32-vCPU variant — the closest is `Standard_L32s_v4` (32 vCPU) for `4xlarge` and `Standard_L64s_v4` (64 vCPU) for `8xlarge`. This means `8xlarge` on Azure gives 2× the compute of `8xlarge` on AWS. This may be intentional (best available equivalent) but should be documented in the mapping table comments.

**5. `db_instance_type` abstract param is not validated at config load time**

There is no `verify_configuration()` check that `db_instance_type` contains a valid size name. A user who writes `db_instance_type: 'huge'` will get silent no-op (see bug #2). Consider adding validation in `_instance_type_validation()` or a dedicated check.

**6. `loader_instance_type` and `monitor_instance_type` have no defaults in `defaults/test_default.yaml`**

The three new `SctField` params have no default values set. Per `AGENTS.md`:
> *"Every configuration option in `sct_config.py` MUST have a corresponding default value"*

An empty string default is fine, but it should be explicit in `defaults/test_default.yaml` to follow the project convention.

---

### 🟢 Observations / Minor

**7. `get_cloud_params` uses `instance_type_{role}` for AWS but the existing param is `instance_type_db` (not `instance_type_db`)**

`sdcm/utils/cloud_sizes.py` L310:
```python
params[f"instance_type_{role}"] = spec.instance_type
```

For `role="db"` this produces `instance_type_db` ✓. For `role="loader"` → `instance_type_loader` ✓. For `role="monitor"` → `instance_type_monitor` ✓. This is correct — just worth noting the naming is consistent with existing SCT params.

**8. `translate-size --config` mode removes abstract params from output**

`sct.py` L~3870: `expanded.pop(abstract_param, None)` removes `db_instance_type` from the output YAML. This is the right behaviour for "show me the fully-expanded config", but it means the output can't be round-tripped back through the abstract sizing system. Consider adding a `--keep-abstract` flag or a comment in the output noting the original abstract size.

**9. `docs/plans/config/cross-cloud-sizing.md` vs `docs/cross-cloud-sizing.md`**

There are now two docs files:
- `docs/plans/config/cross-cloud-sizing.md` — the implementation plan (correct location)
- `docs/cross-cloud-sizing.md` — the usage guide (also correct)

Both are useful and serve different purposes. Just make sure `docs/cross-cloud-sizing.md` is linked from the main `README.md` or `docs/sct-configuration.md` so it's discoverable.

**10. `identify_size` returns `None` for unknown cloud instead of raising**

`sdcm/utils/cloud_sizes.py` L274-275:
```python
else:
    return None
```

An unknown cloud silently returns `None` instead of raising `ValueError` like `resolve_size` does. This inconsistency could hide bugs. Consider raising `ValueError` for unknown clouds here too.

---

### ✅ What's Good

- Clean separation of concerns: mapping data (`cloud_sizes.py`) vs resolution logic (`sct_config.py`) vs CLI (`sct.py`)
- `_user_provided_keys` tracking is the right approach — avoids overriding explicit user config
- 110 parametrized unit tests with full forward + reverse coverage
- GCE disk params (`gce_n_local_ssd_disk_db`, `gce_root_disk_type_db`) are correctly co-resolved with instance type
- `k8s-eks` → `aws` and `k8s-gke` → `gce` backend mapping is correct
- The plan commit is clean and the `.sisyphus/` local file is no longer in the PR

---

### Required Before Merge

1. Wire `arch` detection into `_resolve_instance_sizes()` (bug #1)
2. Add warning log for unresolvable abstract sizes (bug #2)
3. Add explicit empty-string defaults to `defaults/test_default.yaml` for the 3 new params (issue #6)

sdcm/utils/cloud_sizes.py
```diff
@@ -0,0 +1,328 @@
...
+    # Loader role — compute-optimised instances for stress tools.
+    # AWS c6i family has no ARM variant; aws_arm == aws_x86.
+    # OCI: VM.Standard3.Flex — notation encodes :vcpus:memory_gb.
+    # ------------------------------------------------------------------
+    "loader": {
+        "small": SizeMapping(
+            aws_arm=InstanceSpec("c6i.xlarge", 4, 8.0),
+            aws_x86=InstanceSpec("c6i.xlarge", 4, 8.0),
+            gce=InstanceSpec("e2-standard-4", 4, 16.0),
+            azure=InstanceSpec("Standard_F4s_v2", 4, 8.0),
```
fruch, line 158: for loaders wev'e  seen that https://learn.microsoft.com/en-us/azure/virtual-machines/sizes/compute-optimized/fasv6-series?tabs=sizebasic are a bit more capable, and we used them in pref tests
need to consider which one to use as default  (they are cost 66% more in avg. https://cloudprice.net/compare/Standard_F32s_v2-vs-Standard_F32as_v6)

fruch:
## Fixes Applied (commit `6e5c1a0791`)

Addressed the 3 required items from the review:

**Bug #1 — Arch detection wired into `_resolve_instance_sizes()`**

`_resolve_instance_sizes()` now detects the AWS architecture before calling `resolve_size()`:
- If `instance_type_db` is already set in the config (e.g. from `defaults/aws_config.yaml`), it calls `_get_normalized_arch()` on that value to determine `"x86"` vs `"arm"`
- Falls back to `"arm"` (Graviton/i8g) when no existing instance type is set or no region is available
- Arch detection only applies to the `db` role — loaders and monitors have no ARM variants in the mapping

**Bug #2 — Warning log for unresolvable abstract sizes**

Silent `except (KeyError, ValueError): continue` now logs:
```
WARNING: Cannot resolve abstract size '3xlarge' for role 'db' on cloud 'aws': ... — skipping
```

**Issue #6 — Explicit defaults in `defaults/test_default.yaml`**

Added:
```yaml
db_instance_type: ''
loader_instance_type: ''
monitor_instance_type: ''
```

**Tests:** 175 pass (added `test_resolve_instance_sizes_aws_x86` to cover the x86 arch path)

docs/plans/config/cross-cloud-sizing.md
```diff
@@ -0,0 +1,644 @@
...
+### No Existing Abstraction
+
+There is no T-shirt size concept, no cross-cloud mapping, and no translation layer anywhere in
+the codebase. PR #14439 proposes a static mapping document (`docs/gce-instance-type-mapping.md`)
+but not a programmatic translation.
+
+## 3. Goals
+
+1. Define a **T-shirt sizing vocabulary** using AWS-style size suffixes (`large`, `xlarge`,
+   `2xlarge`, `4xlarge`, `8xlarge`, `16xlarge`) as values for per-role config parameters.
```
roydahan, line 88: I wouldn't stick with the AWS type of sizing for few reasons:
1. It may confuse us, we may translate it automatically to AWS while it's not running there.
2. The mapping may need need to be more flexible, even in AWS we have the 3xlarge and 6xlarge, it will just bring more confusion.
3. There could be some misalignments between cloud where a more "native" language could help us.

My proposal, define the number of cores per node and if necessary RAM size (per core or per node).
e.g. `instance_db_cores: 16`
(also can be vcpus / shards instead of cores).

We could also use it if we want in a more flexible ways like `instance_db_cores: <16 && >32`
Or if we want a mixed cluster we can set something like: `instance_db_cores: [16,8,4]`

The RAM size can be autocalculated with a default of min RAM size per core or something that we know much all of our supported instance types.

We will also mix it with `Arch` which can be `aarch64 or arm`.

docs/plans/config/cross-cloud-sizing.md
```diff
@@ -0,0 +1,644 @@
...
+
+2. Create a **bidirectional mapping module** (`sdcm/utils/cloud_sizes.py`) that translates:
+   - Size + role + cloud -> concrete cloud instance type (+ disk params where needed)
+   - Concrete cloud instance type -> size (for reporting / display)
+
+3. Each cloud provider has a **default instance family** for DB nodes:
+   - AWS: `i8g` family (Graviton/ARM, default) and `i4i` family (x86 fallback)
```
roydahan, line 95: why i4i and not i7i?

docs/plans/config/cross-cloud-sizing.md
fruch: naming the core complicates the situation, also mixing the arch with it, when you want to default to i8g, and we don't have arm types for any cloud but AWS.

I don't see how 3xlarge and 6xlarge are of a concern here, we are refering to the default, we don't aim to map any possible type to any possible type on other provider, I don't see the sense in it.

docs/plans/config/cross-cloud-sizing.md
fruch: calling fallback it's actually wrong in this context, we shouldn't define fallbacks in this plan/feature.
for each cloud provider we should set one default family, if other family is needed we will define it directly

defaults/azure_config.yaml
```diff
@@ -6,8 +6,6 @@ azure_region_name:
 user_credentials_path: '~/.ssh/scylla_test_id_ed25519'
 
 azure_instance_type_db: 'Standard_L8s_v3'
```
fruch, line 8: we should drop the default of DB as well, like in other providers

defaults/test_default.yaml
```diff
@@ -366,3 +364,13 @@ agent:
   tls: false
 
 c_s_driver_version: '3'
+
+db_instance_type: ''
```
fruch, line 368: should reconsider if we want a default here

docs/plans/config/cross-cloud-sizing.md
pehala: I share the concern about naming. I find the value arbitrary with no meaning, what does 2xlarge mean, how do I find out? If we want to define cloud agnostic details for a test, we should  define exactly what these mean: How many CPUs? Whats the memory for each? How much disk does this have?

These are the questions I will ask to understand the scale of the test and these are invisible if we use arbitary value like "xlarge". Solution, in my opinion, would be to assing exact parameters (not just instance mappings, as those can change) to the values in the documentation

pehala:
I would prefer discussion over the plan (+ user facing doc) before implementation is submitted for this one, it makes it harder to review IMO

roydahan:
> I would prefer discussion over the plan (+ user facing doc) before implementation is submitted for this one, it makes it harder to review IMO

I agree, let's have a discussion about it.
The plan is already very long to read and comment and there are implementations even before we agreed on some basic things.

docs/plans/config/cross-cloud-sizing.md
roydahan: > naming the core complicates the situation, also mixing the arch with it, when you want to default to i8g, and we don't have arm types for any cloud but AWS.
> 
> I don't see how 3xlarge and 6xlarge are of a concern here, we are refering to the default, we don't aim to map any possible type to any possible type on other provider, I don't see the sense in it.

We're mapping the defaults we know now, but in less than one year we will have other instance types as well.
We may need to be able to cross test on multiple instance types again and we will need some flexability.

fruch:
> > I would prefer discussion over the plan (+ user facing doc) before implementation is submitted for this one, it makes it harder to review IMO
> 
> I agree, let's have a discussion about it. The plan is already very long to read and comment and there are implementations even before we agreed on some basic this

I don't think the implementation is in anyone way to review the plan.
and I don't  see a need for a meeting before collecting information  and comments

I don't see a need for a meeting, hence I'm no gonna schedule one

docs/plans/config/cross-cloud-sizing.md
fruch: we can have more flexibility as it would be needed, I don't see how this is limiting ?

naming it based on vcpu, would still be the same type of limitation

.sisyphus/plans/constraint-based-sizing.md
```diff
@@ -0,0 +1,877 @@
...
+- `sdcm/utils/catalog_generator.py` — script to populate catalog from cloud APIs
+- `data/instance_catalog/` — generated + curated catalog files (YAML per cloud)
+- Updated `sdcm/sct_config.py` — wiring constraint resolution
+- Updated `sct.py` — CLI commands
+- Updated test configs — migrated from T-shirt to constraints
+- `docs/cross-cloud-sizing.md` — updated user guide
+
+### Definition of Done
+- [ ] `{vcpu: 8, memory: ">16gb"}` on AWS resolves to `i8g.2xlarge` (ARM preferred) and all downstream works
+- [ ] `SCT_DB_INSTANCE_TYPE='vcpu=16,memory=>20gb'` env var parses correctly
```
pehala, line 71: I dont think this is correct way how to configure. This is not a singular value and as such shouldnt have single option, it should be a composite config option instead. By doing it like this we are saving ourselves pains with parsing, we can extract the logic into a new file (for the composite type) and we can have easier validation
```
db_instance_type:
  arch: x86
  vcpu: 16
  memory: >20GB
```

fruch: environment variables are used when we want to override configuration
```bash
# one option we use today for dicts like objects:
SCT_DB_INSTANCE_TYPE='{"arch":  "x86", "vcpu":  16, "memory": ">20GB"}'`

# this might also gonna work (works for stress tools docker images)
SCT_DB_INSTANCE_TYPE.arch=x86
SCT_DB_INSTANCE_TYPE.vcpv=16
SCT_DB_INSTANCE_TYPE.memory=">20GB"
```

fruch:
continuing in https://github.com/scylladb/scylla-cluster-tests/pull/14517

PR #14492, opened 2026-05-04
fix(install): handle nonroot case in node-exporter fallback setup

yaronkaikov:
On Rocky 9 with SELinux enforcing, the node_exporter binary downloaded from GitHub to `/usr/local/bin` lacks proper SELinux context, preventing it from binding to port 9100 when run as a systemd service.

Fix: add `restorecon` after installing the binary to restore the correct file context (`bin_t`). Debian is unaffected as it doesn't use SELinux, which explains why deb tests pass without this fix.

Fixes artifacts-rocky9-nonroot-test and artifacts-rocky9-test failures.

### Testing

- [x] https://jenkins.scylladb.com/job/releng-testing/job/artifacts-offline-install/job/artifacts-rocky9-test/908/
- [x] https://jenkins.scylladb.com/job/releng-testing/job/artifacts-offline-install/job/artifacts-rocky9-nonroot-test/910/

### Reminders

### DIFF
--- a/sdcm/node_exporter_setup.py
+++ b/sdcm/node_exporter_setup.py
@@ -18,6 +18,10 @@ class NodeExporterSetup:
             curl -L --retry 5 --retry-max-time 300 -O https://github.com/prometheus/node_exporter/releases/download/v{NODE_EXPORTER_VERSION}/node_exporter-{NODE_EXPORTER_VERSION}.linux-amd64.tar.gz
             tar -xzvf node_exporter-{NODE_EXPORTER_VERSION}.linux-amd64.tar.gz
             mv node_exporter-{NODE_EXPORTER_VERSION}.linux-amd64/node_exporter /usr/local/bin
+            # Restore SELinux context so the binary can be executed as a service on RHEL-based systems
+            if command -v restorecon > /dev/null 2>&1; then
+                restorecon -v /usr/local/bin/node_exporter
+            fi
 
             if [ -e /etc/systemd/system/node_exporter.service ]; then
                 rm /etc/systemd/system/node_exporter.service


yaronkaikov:
@fruch This is the fix for the Rocky9 root and nonroot

sdcm/cluster.py
```diff
@@ -6114,13 +6114,19 @@ def _scylla_install(self, node):
...
         # The unified package may not include scylla-node-exporter, so install standalone node_exporter
         # as a fallback when the scylla-node-exporter service doesn't exist or cannot be started.
+        # Note: node_exporter runs as a system service even for nonroot scylla installs,
+        # so we always use sudo systemctl here.
         node.remoter.sudo("systemctl daemon-reload", ignore_status=True)
         node_exporter_setup = NodeExporterSetup()
-        node_exporter_service_exists = node.is_service_exists("scylla-node-exporter")
+        # Check both system and user service locations
+        node_exporter_service_exists = node.remoter.run(
+            "sudo systemctl list-unit-files | grep scylla-node-exporter.service",
```
fruch: sure you don't need to pass --user here ?

fruch: i.e. that's exactly what the code is doing:
```
    def wrap_cmd_with_permission(self, cmd):
        """Generate commandline prefix according to the privilege"""
        if self.is_nonroot_install:
            return f"{cmd} --user"
        else:
            return f"sudo {cmd}"
```

fruch: now I get it, it's now depends on the system one, hence this code doesn't work.

fruch: but I still don't understand why it's failing in looking for `scylla-node-exporter` that would always match how scylla is installed, wouldn't it ?

yaronkaikov: I admit that i don't understand it either ;-)

yaronkaikov: oh, because node-exporter doesn't include in the unified package anymore :-)

yaronkaikov: The `is_service_exists` logic was actually correct — for nonroot it uses `systemctl --user` which doesn't find `scylla-node-exporter` (because it's no longer in the unified package per RELENG-495), and correctly falls through to `NodeExporterSetup.install()` which uses `sudo` to install a standalone node_exporter as a system service.

The actual failure was **SELinux on Rocky 9**: the node_exporter binary downloaded from GitHub to `/usr/local/bin` had the wrong SELinux context (likely `user_home_t` or `default_t` instead of `bin_t`). Even though `systemctl start` may have returned, the process was blocked from binding to port 9100. That's what the `restorecon` fix addresses — it restores the proper SELinux file context after moving the binary.

Debian doesn't use SELinux, which is why deb tests pass without this fix.

That said, I changed the service detection to explicitly use `sudo systemctl` (instead of `node.is_service_exists()` which routes through `systemctl --user` for nonroot) to make the intent clearer — node_exporter is always a system-level service regardless of scylla's install mode.

fruch: still the comment "# Check both system and user service locations" is misleading

if the code was logically and correctly working before, let's drop the unneeded parts from this fix.

yaronkaikov: Reverted unneed changes - and verified with https://jenkins.scylladb.com/job/releng-testing/job/artifacts-offline-install/job/artifacts-rocky9-test/909/

yaronkaikov:
@fruch please consider merging

fruch (approved):

PR #14505, opened 2026-05-04
docs(plans): extend AWS capacity plan with mid-test add_nodes fallback

dimakr:
Add Phases 6-8 for items not covered by initial plan:
- Phase 6: mid-test add_nodes AZ fallback (gated by allow_cross_az_rack)
- Phase 7: extend SCTCapacityReservation to include target/oracle types
- Phase 8: optional pre-flight capacity probe

Extend Phase 2 AZResolver to include instance_type_db_target and nemesis_grow_shrink_instance_type.

Refs: https://github.com/scylladb/scylla-cluster-tests/pull/14322

### Reminders

### DIFF
--- a/docs/plans/infrastructure/aws-capacity-az-fallback.md
+++ b/docs/plans/infrastructure/aws-capacity-az-fallback.md
@@ -2,7 +2,7 @@
 status: draft
 domain: cluster
 created: 2026-04-12
-last_updated: 2026-04-12
+last_updated: 2026-05-04
 owner: fruch
 ---
 # AWS Capacity Validation and Automatic AZ/Region Fallback
@@ -25,6 +25,7 @@ Availability Zone you requested (eu-west-1a).
 3. **AZ fallback only for artifact tests**: The existing `aws_fallback_to_next_availability_zone` feature only works in the legacy provisioning path and only wraps DB cluster creation -- loaders and monitors are not covered
 4. **No cross-AZ fallback in modern path**: The modern provisioning path (`sdcm/sct_provision/`) has zero AZ fallback logic -- any capacity error is a hard failure
 5. **Multi-node tests cannot use AZ fallback**: The existing fallback is effectively limited to single-node artifact tests
+6. **Dynamically-added instance types not considered upfront**: `instance_type_db_target` (platform-migration) and `nemesis_grow_shrink_instance_type` are not part of the upfront AZ filter or capacity-reservation builder, so the chosen AZ may end up being one that does not offer the type the test will try to launch later. Mid-test recovery via cross-AZ retry is unsafe (rack/AZ alignment, tablet placement) and often futile (region-wide pressure), so mitigation is prevention-side only: extend upfront filter, extend capacity reservation, optional pre-flight probe. When prevention fails, the test aborts as today
 
 ### Relationship to PR #13317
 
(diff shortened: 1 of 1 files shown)


docs/plans/infrastructure/aws-capacity-az-fallback.md
```diff
@@ -251,15 +264,116 @@ When all AZs in a region are exhausted, try a different region. Scoped to single
 
 ---
 
+### Phase 6: Mid-test `add_nodes` Capacity-Error Retry
+
+**Importance: Critical** | **Scope: ~150 LOC, 1 PR** | **Dependencies: Phase 1, Phase 2**
+
+Critical for migration / grow-shrink / refresh-with-replace tests. Phases 3 and 4 wrap initial provisioning only -- they do not help when a capacity error hits during `BaseCluster.add_nodes()` mid-test. Real-world example: platform-migration test #4 ran 1h+ of stress, then failed at 00:20 UTC during `migrate_nodes_batch_by_rack` -> `add_nodes` because `i8g.2xlarge` capacity ran out in `eu-west-1a`. Plan as merged would not catch this -- the cluster was already provisioned successfully an hour earlier.
```
fruch: not sure if it would work, when we don't have capacity it doesn't get solved in matter of minutes

1) we need to try other racks, but that might break our test configuration
2) we can skip the nemesis, but that might leave us in problematic condition

so we need to stop the test if the cluster configuration isn't as expected

again, not sure retries might not work for this case.

pehala: Agree, I think this wont help much but will introduce unnecessary complexity. I think stopping the test is still the best option here

pehala:
Shouldnt we first implement the original plan and then create a new miniplan to enhance it further? I dislike updating existing agreed upon plan before it is done.

fruch:
> Shouldnt we first implement the original plan and then create a new miniplan to enhance it further? I dislike updating existing agreed upon plan before it is done.

I don't see why it's a problem, it's not that it goes in without discussion, like we are doing now
it's just we run in more problem around this, that might be taking into consideration when implementing.

pehala:
> I don't see why it's a problem, it's not that it goes in without discussion, like we are doing now
> it's just we run in more problem around this, that might be taking into consideration when implementing.

I think extending already big plan will lead to a bigger implementation PR, which will be harder to review, but given it is just adding additional phases, it shouldnt matter.

docs/plans/infrastructure/aws-capacity-az-fallback.md
dimakr: Dropped the proposed parts related to mid-test mitigation.

fruch (approved):

dimakr:
@pehala @fruch if this one is ok, maybe we can push it forward? I'm starting the implementation of the main plan from https://github.com/scylladb/scylla-cluster-tests/pull/13317, and it would be good to have everything in place

pehala (approved):

fruch:
> @pehala @fruch if this one is ok, maybe we can push it forward? I'm starting the implementation of the main plan from #13317, and it would be good to have everything in place

it's o.k. but I'll leave the capacity reservation to the end.

fruch (approved):

PR #14507, opened 2026-05-04
ci(workflows): replace ISSUE_ASSIGNMENT_TO_PROJECT_TOKEN with AUTO_BACKPORT_TOKEN

fruch:
## Summary
- Replace `ISSUE_ASSIGNMENT_TO_PROJECT_TOKEN` secret with `AUTO_BACKPORT_TOKEN` in the `cache-issues` workflow
- Consolidates token usage since `AUTO_BACKPORT_TOKEN` already has the required cross-repo read permissions

### DIFF
--- a/.github/workflows/cache-issues.yaml
+++ b/.github/workflows/cache-issues.yaml
@@ -19,7 +19,7 @@ jobs:
             gh pr list --state all --json number,state,labels,title --limit 30000 --template '{{range .}}{{.number}},{{.state}},{{range .labels}}{{.name}}|{{end}},{{.title}}{{println ""}}{{end}}' --repo scylladb/$repo > issues/pull-requests/scylladb_$repo.csv
           done
         env:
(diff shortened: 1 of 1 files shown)


pehala (approved):

vponomaryov (approved):
LGTM

roydahan:
@fruch I think it'll be good to improve the caching mechanism by checking its age and don't use it if it's too old.

fruch:
> @fruch I think it'll be good to improve the caching mechanism by checking its age and don't use it if it's too old.

We need other way to inform us it's broken, not by breaking the whole world and rate limit the same account we use for multiple things.

fruch:
> @fruch I think it'll be good to improve the caching mechanism by checking its age and don't use it if it's too old.

Also we need to retire this method of skipping, it's have too many rough edges.
And it's doing more harm then good at this point of dtest life, as it sunsetting.

PR #14508, opened 2026-05-04
fix(gemini): respect stress_duration override and fix seed reporting in Argus

CodeLieutenant:
Three bugs fixed:

1. run_gemini() ignored stress_duration when gemini_cmd used YAML block scalar format (|). The --duration flag appears at the start of the string with no leading space, so the old ' --duration' substring check always evaluated to False, falling through to get_timeout_from_stress_cmd() which parsed the original --duration value unchanged. Fix: drop the leading-space requirement and use a (^|\s) anchor in the replacement regex; use \S+ to match any time-unit format (h/m/s).

2. The --warmup substitution (stress_duration * 0.2) had no basis; removed.

3. Gemini seed was always reported as -1 in Argus. The generated command uses --seed=VALUE (equals sign) but the regex looked for '--seed VALUE ' (space-delimited with trailing space), which never matched. Fix: use r'--seed[= ](\d+)' for both --seed and the new --schema-seed.

Also adds gemini_schema_seed to GeminiResultsRequest and reports it separately from gemini_seed so runs with different run/schema seeds are represented correctly in Argus.

### Testing

- [ ]

### Reminders

Fixes https://scylladb.atlassian.net/browse/SCT-196

### DIFF
--- a/unit_tests/unit/test_stress_thread_functions.py
+++ b/unit_tests/unit/test_stress_thread_functions.py
@@ -65,7 +65,10 @@ def test_get_timeout_from_stress_cmd(stress_cmd, timeout):
 
 def _apply_gemini_stress_duration(cmd: str, stress_duration: int) -> str:
     """Mirror of the substitution logic in ClusterTester.run_gemini()."""
-    cmd = re.sub(r"(^|\s)--duration\s+\S+", f"\\1--duration {stress_duration}m", cmd)
+    if "--duration" in cmd:
+        cmd = re.sub(r"(^|\s)--duration\s+\S+", f"\\1--duration {stress_duration}m", cmd)
+    else:
+        cmd = cmd + f" --duration {stress_duration}m"
     return cmd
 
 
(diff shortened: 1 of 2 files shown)


pehala:
You fixed 3 bugs in a single commit, I would prefer if we kept it to 1 commit per bug, which makes it easier to review individual fixes

unit_tests/unit/test_stress_thread_functions.py
```diff
@@ -59,3 +61,68 @@ def test_duration_str_to_seconds_function(duration, seconds):
 )
 def test_get_timeout_from_stress_cmd(stress_cmd, timeout):
     assert get_timeout_from_stress_cmd(stress_cmd) == timeout
+
+
+def _apply_gemini_stress_duration(cmd: str, stress_duration: int) -> str:
```
pehala, line 66: This is wrong, we need to be testing real code, otherwise this test proves nothing

argus/common/sct_types.py
```diff
@@ -38,6 +38,7 @@ class GeminiResultsRequest(TypedDict):
     gemini_version: str
     gemini_status: str
     gemini_seed: str
+    gemini_schema_seed: str
```
pehala: I believe argus folder should no be updated right? IT is autogenerated

CodeLieutenant: oh, yes, this is argus client, this would not work.

unit_tests/unit/test_stress_thread_functions.py
fruch: should work a bit like the test above `test_get_timeout_from_stress_cmd`, just call one function or method and test that, not a copy of implementation

soyacz:
@CodeLieutenant still a draft?

CodeLieutenant:
> @CodeLieutenant still a draft?

will be ready today, im having a test with it

soyacz:
@CodeLieutenant what happened?

CodeLieutenant:
> @CodeLieutenant what happened?

I deleted my fork, i had problems with it, so will reapply it again

PR #14511, opened 2026-05-04
feat(loaders): detect dead loader nodes and fail test early

cezarmoise:
Add a background SSH health-check thread to BaseLoaderSet that periodically pings every loader node. After LOADER_HEALTH_CHECK_RETRIES (3) consecutive failures a LoaderNodeDownEvent with CRITICAL severity is published, which triggers EventsAnalyzer.kill_test via SIGUSR2.

Fixes: SCT-288

Previously, when a loader node died the test continued silently doing nothing until the full stress timeout (potentially hours) expired.

Key design decisions:
- Check stop event inside per-node sweep so teardown is responsive
- join() timeout uses LOADER_HEALTH_CHECK_TIMEOUT (SSH bound), not the sleep interval; thread ref kept intact if join times out to prevent a second thread spawning alongside a stuck one
- Health check stopped before loader remoter teardown to avoid false CRITICAL events during intentional shutdown
- Health check started after all setUp futures complete to avoid leaking a background thread if later setup work fails

### Testing

- [ ]

### Reminders

### DIFF
--- a/defaults/severities.yaml
+++ b/defaults/severities.yaml
@@ -153,3 +153,4 @@ ScrubValidationErrorEvent: ERROR
 PartitionRowsValidationEvent: CRITICAL
 FailedResultEvent: ERROR
 HDRFileMissed: ERROR
+LoaderNodeDownEvent: CRITICAL
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -171,6 +171,7 @@ from sdcm.sct_events.database import (
 from sdcm.sct_events.nodetool import NodetoolEvent
 from sdcm.sct_events.decorators import raise_event_on_failure
 from sdcm.sct_events.filters import EventsSeverityChangerFilter
+from sdcm.sct_events.loaders import LoaderNodeDownEvent
 from sdcm.utils.auto_ssh import AutoSshContainerMixin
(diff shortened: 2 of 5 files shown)


pehala:
> Key design decisions:
> 
>     * Check stop event inside per-node sweep so teardown is responsive
> 
>     * join() timeout uses LOADER_HEALTH_CHECK_TIMEOUT (SSH bound), not the sleep interval; thread ref kept intact if join times out to prevent a second thread spawning alongside a stuck one
> 
>     * Health check stopped before loader remoter teardown to avoid false CRITICAL events during intentional shutdown
> 
>     * Health check started after all setUp futures complete to avoid leaking a background thread if later setup work fails

I think this does deserve a miniplan, it will make it easier to review the concept

fruch:
this kind of aligns with the work in https://github.com/scylladb/scylla-cluster-tests/pull/13354

when we loose connection with loader or monitor, it's mostly cause of OOMs or kernel panics
if we gonna have a monitoring thread for each node, I would suggest it to be one per node that take care of all of the options. too many threads per node, is gonna limit drastically our scaling tests.

regardless it would be more helpful to have the exact reason, not just vague we lost connection

fruch:
@cezarmoise isn't https://github.com/scylladb/scylla-cluster-tests/pull/13354 covering this ? i.e. can we close this one ?

Issue #833, opened 2019-03-13
Replace all places we use private/public IPs by default


roydahan:
In continue to other issues me and @fruch found where using ip_address incorrectly.
We need to replace all places in the code that are specifically mentioning private_ip_address to ip_address that is decided in one place (public / private).

For example, longevity with multiDC fails during TerminateAndReplace nemesis because of this:
https://github.com/scylladb/scylla-cluster-tests/blob/b64989be567e7023e3861402de31a66158b0aeb0/sdcm/nemesis.py#L349

roydahan:
@fruch already have some changes for this in PR #828

fruch:
@roydahan, closing this one, since we seen the multi region use case is working.
We'll fix specific cases if we found ones

PR #14512, opened 2026-05-04
feature(cluster): use control-connection query fallback for public address connections

fruch:
## Summary

When `IP_SSH_CONNECTIONS` is set to `public`, the CQL driver cannot establish node pools because nodes are accessed via their public IPs, which differ from the broadcast RPC addresses advertised by the cluster. This causes `NoHostAvailable` errors.

This PR uses the new `ControlConnectionQueryFallback.SkipPoolCreation` option (added in [python-driver PR #878](https://github.com/scylladb/python-driver/pull/878), released in scylla-driver 3.29.10) to fall back to the control connection for queries when public addresses are in use.

### What changed

- **`sdcm/cluster.py`**: In `_create_session`, when `IP_SSH_CONNECTIONS == "public"`, pass `allow_control_connection_query_fallback=ControlConnectionQueryFallback.SkipPoolCreation` to the `ClusterDriver` constructor. This tells the driver to skip node-pool creation entirely and route all application queries through the control connection.
- **`pyproject.toml`**: Update `scylla-driver` from `3.29.9` to `3.29.10` (the first release containing `ControlConnectionQueryFallback`).

### Previous approach (dropped)

The original implementation used `DynamicWhiteListRoundRobinPolicy` from an unmerged python-driver branch ([PR #833](https://github.com/scylladb/python-driver/pull/833)), which required pinning `scylla-driver` to a git ref. That PR was not accepted. This rework uses the officially released and merged approach instead.

## Backends Affected
- [x] AWS - public IP connections
- [x] GCE - public IP connections
- [x] Azure - public IP connections

## Manual Testing Required

- [x] **Public IP CQL connections**: Requires cloud backend with `IP_SSH_CONNECTIONS=public`
  - Test: `uv run sct.py run-test artifacts_test.ArtifactsTest.test_scylla_service --backend aws --config test-cases/artifacts/ami.yaml --config configurations/network_config/test_communication_public.yaml`
  - Expected: CQL connections succeed via control-connection fallback


soyacz:
[flaky gemini integration test issue](https://scylladb.atlassian.net/browse/SCT-351)

soyacz (commented):
maybe we should use it in the artifact test (we use it in Azure tests: jenkins-pipelines/oss/artifacts/artifacts-azure-image.jenkinsfile)?

fruch:
> maybe we should use it in the artifact test (we use it in Azure tests: jenkins-pipelines/oss/artifacts/artifacts-azure-image.jenkinsfile)?

I've tested this locally, using public communication with AWS, should be the same.

soyacz (approved):
LGTM

PR #14517, opened 2026-05-05
docs(sizing): constraint-based instance sizing plan

fruch:
## Summary

Implementation plan and user-facing documentation for constraint-based instance sizing.

Replaces T-shirt sizing (`db_type: local_ssd`) with constraint dictionaries:
```yaml
db_instance_type:
  vcpu: 16
  memory: ">20gb"
  arch: arm64
```

The resolver picks the cheapest matching instance per cloud, using preferred families (i8g for AWS, z3 for GCE, v4 for Azure, DenseIO.E5 for OCI).

## Contents

- `docs/plans/constraint-based-sizing.md` — full implementation plan (4 waves, 13 tasks)
- `docs/cross-cloud-sizing.md` — user guide: constraint syntax, selection algorithm, CLI tools, migration

## Reference

- #14490 — prior T-shirt sizing PR (kept as reference, not superseded)

### DIFF
--- a/.sisyphus/plans/constraint-based-sizing.md
+++ b/.sisyphus/plans/constraint-based-sizing.md
@@ -1,0 +1,1 @@
+../../../docs/plans/constraint-based-sizing.md
--- a/.sisyphus/plans/constraint-based-sizing.md
+++ b/.sisyphus/plans/constraint-based-sizing.md
@@ -1,1 +1,0 @@
-../../../docs/plans/constraint-based-sizing.md
--- a/docs/cross-cloud-sizing.md
+++ b/docs/cross-cloud-sizing.md
@@ -10,7 +10,7 @@ SCT uses constraint-based instance selection to automatically resolve hardware r
 # Specify hardware requirements — resolved per cloud at runtime
 db_instance_type:
   vcpu: 8
-  memory: ">16gb"
+  memory: 32
 
 loader_instance_type:
   vcpu: 4
@@ -29,14 +29,20 @@ The same config with `--backend gce` resolves to:
 - **loader**: `e2-standard-4` (4 vCPU, 16 GB)
 - **monitor**: `n2-highmem-4` (4 vCPU, 32 GB)
 
-### Via environment variable
+### Via environment variable (dot-notation)
+
+SCT's existing dot-notation for nested config values works out of the box:
 
 ```bash
-# Flat string format for env vars
-export SCT_DB_INSTANCE_TYPE='vcpu=16,memory=>32gb'
-export SCT_LOADER_INSTANCE_TYPE='vcpu=4'
-export SCT_MONITOR_INSTANCE_TYPE='vcpu=2'
+export SCT_DB_INSTANCE_TYPE.vcpu=16
+export SCT_DB_INSTANCE_TYPE.memory=32
+export SCT_DB_INSTANCE_TYPE.arch=arm64
+
+export SCT_LOADER_INSTANCE_TYPE.vcpu=4
+export SCT_MONITOR_INSTANCE_TYPE.vcpu=2
 ```
+
+This builds the equivalent dict `{vcpu: 16, memory: 32, arch: "arm64"}` using the same mechanism as other nested SCT config parameters (e.g. `append_scylla_yaml`).
 
 ### Literal instance type (backward compatible)
 
@@ -52,38 +58,45 @@ db_instance_type: 'i4i.2xlarge'
 
 ```yaml
 db_instance_type:
-  vcpu: 8              # exact match
-  memory: ">16gb"      # minimum 16 GB
-  disk: "500gb-2tb"    # range: 500 GB to 2 TB local disk
-  arch: "arm64"        # optional: force architecture
+  vcpu: 8              # exact match (required)
+  memory: 32           # minimum 32 GB (plain int = minimum, unit always GB)
+  disk: 500            # minimum 500 GB local disk (plain int = minimum, unit always GB)
+  arch: arm64          # optional: force architecture (accepts: arm64, arm, x86_64, x86)
 ```
 
-### Flat string format (for env vars)
+### Advanced operators (optional)
 
-```bash
-SCT_DB_INSTANCE_TYPE='vcpu=8,memory=>16gb,disk=500gb-2tb,arch=arm64'
+For power users who need more precise control:
+
+```yaml
+db_instance_type:
+  vcpu: "8-16"         # range: 8 to 16 vCPUs
+  memory: ">64"        # more than 64 GB
+  disk: "500-2048"     # range: 500 GB to 2 TB local disk
 ```
-
-Both formats are equivalent and produce identical results.
 
 ## Constraint Reference
 
-| Field | Description | Operators | Units | Default if omitted |
-|-------|-------------|-----------|-------|--------------------|
-| `vcpu` | Number of virtual CPUs | exact, `>`, `<`, `>=`, range | plain int | No filter |
-| `memory` | RAM size | exact, `>`, `<`, `>=`, range | `gb`, `tb` | No filter |
-| `disk` | Local disk size (per disk) | exact, `>`, `<`, `>=`, range | `gb`, `tb` | db role: must have local disk |
-| `arch` | CPU architecture | exact only | `arm64`, `x86_64` | aws=`arm64`, others=`x86_64` |
+| Field | Description | Operators | Units | Required? |
+|-------|-------------|-----------|-------|-----------|
+| `vcpu` | Number of virtual CPUs | exact, range (`"8-16"`) | plain int | **Yes** |
+| `memory` | RAM size (minimum) | plain int (min), `>`, `<`, `>=`, range | GB (always) | No |
+| `disk` | Local disk size per disk (minimum) | plain int (min), `>`, `<`, `>=`, range | GB (always) | No (db role: must have local disk) |
+| `arch` | CPU architecture | exact only | `arm64`/`arm`, `x86_64`/`x86` | No (aws=`arm64`, others=`x86_64`) |
+
+**Plain integers**: For `memory` and `disk`, a plain integer always means "minimum GB". `memory: 32` is equivalent to `memory: ">=32"`.
+
+**Architecture shorthands**: `arm` is shorthand for `arm64`, `x86` is shorthand for `x86_64`.
 
 ### Operator examples
 
 | Syntax | Meaning |
 |--------|---------|
 | `vcpu: 16` | Exactly 16 vCPUs |
-| `memory: ">32gb"` | More than 32 GB |
-| `memory: ">=32gb"` | 32 GB or more |
-| `memory: "<64gb"` | Less than 64 GB |
-| `disk: "500gb-2tb"` | Between 500 GB and 2 TB |
+| `vcpu: "8-16"` | Between 8 and 16 vCPUs |
+| `memory: 32` | At least 32 GB (plain int = minimum) |
+| `memory: ">64"` | More than 64 GB |
+| `disk: "500-2048"` | Between 500 GB and 2 TB |
 
 ## Selection Algorithm
 
@@ -100,14 +113,14 @@ When you specify constraints, the system:
 
 | Role | AWS | GCE | Azure | OCI |
 |------|-----|-----|-------|-----|
-| db | i8g (ARM), i4i (x86 fallback) | z3-highmem | Standard_L*s_v4 | DenseIO.E5.Flex |
+| db | i8g (ARM), i7i, i4i (x86 fallback) | z3-highmem | Standard_L*s_v4 | DenseIO.E5.Flex |
 | loader | c6i | e2-standard, e2-highcpu | Standard_F*s_v2 | VM.Standard3.Flex |
 | monitor | t3, m6i | n2-highmem | Standard_D*_v4 | VM.Standard.E4.Flex |
 
 ### Implicit constraints
 
 - **db role**: Automatically requires `local_disk_count > 0` (NVMe/local SSD) unless explicitly overridden
-- **AWS db**: Defaults to ARM (`i8g`) architecture; use `arch: "x86_64"` to get `i4i` family
+- **AWS db**: Defaults to ARM (`i8g`) architecture; use `arch: x86_64` to get `i7i`/`i4i` family
 
 ## Examples
 
@@ -116,7 +129,7 @@ When you specify constraints, the system:
 ```yaml
 db_instance_type:
   vcpu: 16
-  memory: ">64gb"
+  memory: 64
 
 loader_instance_type:
   vcpu: 16
@@ -132,10 +145,21 @@ Resolves on AWS to: `i8g.4xlarge` / `c6i.4xlarge` / `m6i.xlarge`
 ```yaml
 db_instance_type:
   vcpu: 8
-  arch: "x86_64"
+  arch: x86_64
 ```
 
-Resolves to `i4i.2xlarge` instead of the default `i8g.2xlarge`.
+Resolves to `i7i.2xlarge` instead of the default `i8g.2xlarge`.
+
+### Flexible vCPU for cross-cloud compatibility
+
+```yaml
+# OCI DenseIO starts at 16 vCPU; use range to allow flexibility
+db_instance_type:
+  vcpu: "8-16"
+  memory: 32
+```
+
+Resolves to 8-vCPU instances on AWS/GCE/Azure, 16-vCPU on OCI (smallest available).
 
 ### Minimal CI test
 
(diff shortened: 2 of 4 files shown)


.sisyphus/plans/constraint-based-sizing.md
fruch, line 1: we should git ignore this folder

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+The same config with `--backend gce` resolves to:
+- **db**: `z3-highmem-8` (8 vCPU, 64 GB, local SSD)
+- **loader**: `e2-standard-4` (4 vCPU, 16 GB)
+- **monitor**: `n2-highmem-4` (4 vCPU, 32 GB)
+
+### Via environment variable
+
+```bash
+# Flat string format for env vars
+export SCT_DB_INSTANCE_TYPE='vcpu=16,memory=>32gb'
```
fruch: wondering if using `:` would work better, and might not need quotes (even that `<` would still need

```
export SCT_DB_INSTANCE_TYPE=vcpu:16,memory:>32gb
```

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+db_instance_type: 'i4i.2xlarge'
+```
+
+## Constraint Syntax
+
+### YAML dict format (in config files)
+
+```yaml
+db_instance_type:
+  vcpu: 8              # exact match
```
fruch: should we support range in vcpu ?

fruch: since some (like OCI) wouldn't have a match for everything

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+Both formats are equivalent and produce identical results.
+
+## Constraint Reference
+
+| Field | Description | Operators | Units | Default if omitted |
+|-------|-------------|-----------|-------|--------------------|
+| `vcpu` | Number of virtual CPUs | exact, `>`, `<`, `>=`, range | plain int | No filter |
+| `memory` | RAM size | exact, `>`, `<`, `>=`, range | `gb`, `tb` | No filter |
+| `disk` | Local disk size (per disk) | exact, `>`, `<`, `>=`, range | `gb`, `tb` | db role: must have local disk |
+| `arch` | CPU architecture | exact only | `arm64`, `x86_64` | aws=`arm64`, others=`x86_64` |
```
fruch: maybe should also support shorthands: 'arm', 'x86'

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Backends
+
+| Backend | Constraint resolution |
+|---------|----------------------|
+| `aws` | Full resolution from catalog |
+| `gce` | Full resolution from catalog |
+| `azure` | Full resolution from catalog |
+| `oci` | Full resolution from catalog |
+| `docker` | Silently skipped (no instance types) |
```
fruch: skips should't be silent

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+| `baremetal` | Silently skipped (pre-existing cluster) |
+| `k8s-*` | Silently skipped (pod resources managed separately) |
```
fruch: instance types might be still relevant for GKE/EKS

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+```
+
+## Constraint Syntax
+
+### YAML dict format (in config files)
+
+```yaml
+db_instance_type:
+  vcpu: 8              # exact match
+  memory: ">16gb"      # minimum 16 GB
```
roydahan: When we set our defaults, we should test to make sure that we match all the relevant instances.
For example, here it wouldn't work for GCP instances like Z3 and Z4 because they have 126GB and 63GB RAM.

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+3. **Filters by preferred family** — each role has preferred instance families per cloud (see below)
+4. **Applies constraints** — removes instances that don't satisfy vcpu/memory/disk requirements
+5. **Sorts by**: family preference order → price (cheapest first) → vCPU count (ascending)
+6. **Returns first match** — deterministic, always the same result for the same input
+
+### Preferred Families
+
+| Role | AWS | GCE | Azure | OCI |
+|------|-----|-----|-------|-----|
+| db | i8g (ARM), i4i (x86 fallback) | z3-highmem | Standard_L*s_v4 | DenseIO.E5.Flex |
```
roydahan: i7i should be the first fallback before i4i.

roydahan: we want also to list also the densed options and possible have that into the calculation.
(Some cases that require i8ge / i7ieq/ i3en)

docs/cross-cloud-sizing.md
pehala: With the new composite yaml format, the submissions here is going to be a dict always no? There wont be any additional parsing needed

docs/cross-cloud-sizing.md
pehala: I think it would be fine to add range later, we can adjust once we get some experience with it

docs/cross-cloud-sizing.md
pehala: I dont think it makes any difference, it is a static string regardless and having two concrete values is easy enough

docs/cross-cloud-sizing.md
pehala: They do, but I would postpone this work. Given the current usage patterns for k8s/eks this is not needed there

docs/cross-cloud-sizing.md
pehala: Densed options are for storage only, so unless we add disk storage into the chooses as well (I do not think it is currently there) they are not needed

fruch: cases that need specific instances for what ever reason, can still specify them, and not use this feature.

docs/cross-cloud-sizing.md
fruch: if one need to type them it's relevant

docs/cross-cloud-sizing.md
fruch: it mean we need to define it as string from beginning

pehala: Fair, in that case I would define it as string from the beginning so we can support ranges in the future. If it is easy change, we can tackle it here as well, it is just not neccessary from my poitn of view

docs/cross-cloud-sizing.md
fruch: I don't know what is the "new composite yaml format"

parsing would be needed to convert from env var to dict/model, we just need to define it json parser or other parsers. since yaml supports json, json can be used out of the box.
for complex nest strucaures  the json isn't something easily written

docs/cross-cloud-sizing.md
fruch: exactly the reason I preferred direct mapping up front

regardless we gonna have a cli to be able to check the configures

docs/cross-cloud-sizing.md
fruch: it would probably be working out of the box, since I don't think they are using different configuration parameters for the instance types, but the exact same ones

I think it would be just give it a quick test, the the configuration is expended correctly. and that's it.

docs/cross-cloud-sizing.md
fruch: it might, for example the once we are using in OCI starts from 16 vcpu, and taking a case running 2xlarge with 8 vcpu, wouldn't map to anything.

one would say it's fine, and shouldn't work, and it should be set manually if needed.
one would say we still should map it as best we can.

fruch: or we can accept it up front: `vcpu: "8-16"` ?

docs/cross-cloud-sizing.md
pehala: > I don't know what is the "new composite yaml format"
```
db_instance_type:
  vcpu: 16
  memory: ">20gb"
  arch: arm64
```
I meant this (composite meaning, it is not single value but dict/obj of values), which could (should IMO) be implemented as Pydantic model, as it has static fields

> 
> parsing would be needed to convert from env var to dict/model, we just need to define it json parser or other parsers. since yaml supports json, json can be used out of the box. for complex nest strucaures the json isn't something easily written

`append_scylla_yaml` already does something like this and we have `DictOrStrOrPydantic`, which should fit the case nicely, without us having to define anything special for this case

docs/cross-cloud-sizing.md
pehala: For simplicity reasons, I would support only one name for each arch, but I dont oppose having shorthands

docs/cross-cloud-sizing.md
fruch: I dare to try override some value for `append_scylla_yaml` from the env var

it might be doable, I would take me 3-4 trials before getting it correct
see https://github.com/scylladb/scylla-cluster-tests/blob/master/docs/faq.md#how-to-define-parameters-for-performance-throughput-test-

soyacz: but can't we just `db_instance_type.vcpu=8` IIRC we do this kind for overriding stress tools already

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Quick Start
+
+### In a test config YAML
+
+```yaml
+# Specify hardware requirements — resolved per cloud at runtime
+db_instance_type:
+  vcpu: 8
+  memory: ">16gb"
```
soyacz: I think in the config, we could assume memory and disks to be int's instead of `>16gb` and it would always mean min value. Also unit could be in gb always.
This way we could have `memory: 16` or `memory_gb: 16`

soyacz: the same for the disk

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+### update-catalog
+
+Refresh the instance catalog from live cloud APIs:
+
+```bash
+# Update all clouds
+uv run sct.py update-catalog
+
+# Update specific cloud
+uv run sct.py update-catalog --cloud aws
```
soyacz: I think cloud providers catalog is too big.
we should add params for instance families like `uv run sct.py update-catalog --cloud aws --families i8g,i7i,c7i` etc.

docs/cross-cloud-sizing.md
```diff
@@ -0,0 +1,349 @@
...
+
+## Selection Algorithm
+
+When you specify constraints, the system:
+
+1. **Filters by cloud** — only considers instances for the active backend
+2. **Filters by architecture** — uses `arch` constraint if specified, otherwise cloud default (aws=arm64, others=x86_64)
+3. **Filters by preferred family** — each role has preferred instance families per cloud (see below)
+4. **Applies constraints** — removes instances that don't satisfy vcpu/memory/disk requirements
+5. **Sorts by**: family preference order → price (cheapest first) → vCPU count (ascending)
```
soyacz: I see issue here: imagine we want test with high disk (i3en instance), but preferred one is i4i before i3en. Then i4i with disk that matches requirement will be chosen first, despite it has to use many more cpu's than i3en, and will be much more expensive.
IMHO, we should have vCPU count as the first order (the closest one), then preferred family, then the price.
we can't put preferred family at the end as i4i price will be always lower than i7i (so i7i would not be ever used).

fruch: 1) cpu should be mandatory to set , disk / mem would be optional, 16 vcpu + big enough disk that fit i3en family would work
2) if a case is designed for i3en, is can specify it with using this feature.
3) preference means defaults here, and would be i8g, and i7i if x86 was defined

this is not meant to be a way to generally select all possible types, there is a way to select specific families already, this is only for the common cases that we can easily define the requirements, but we **don't** have  specific instances to work with.

from time to time we might update those defaults

.sisyphus/plans/constraint-based-sizing.md
fruch: Done. Added `.sisyphus/` to `.gitignore` and removed the tracked file.

docs/cross-cloud-sizing.md
fruch: Dropped the flat string format entirely. As soyacz suggested and pehala's `DictOrStrOrPydantic` hint confirmed — SCT's existing dot-notation handles this natively:

```bash
export SCT_DB_INSTANCE_TYPE.vcpu=16
export SCT_DB_INSTANCE_TYPE.memory=32
export SCT_DB_INSTANCE_TYPE.arch=arm64
```

This uses the same mechanism as `append_scylla_yaml` overrides (lines 3195-3209 in sct_config.py). No custom parser needed.

docs/cross-cloud-sizing.md
fruch: Done. vcpu is now mandatory and supports range syntax: `vcpu: "8-16"`.

This addresses the OCI gap (DenseIO starts at 16 vCPU) — a config with `vcpu: "8-16"` will pick 8-vCPU on AWS/GCE/Azure and 16-vCPU on OCI (smallest available).

vcpu is defined as `int | str` from the start so we don't need a breaking change later.

docs/cross-cloud-sizing.md
fruch: Done. Added shorthands: `arm` → `arm64`, `x86` → `x86_64`. Both full and short forms accepted.

docs/cross-cloud-sizing.md
fruch: Done. Changed to info log: `Skipping instance resolution for backend=docker`. Updated both docs and plan.

docs/cross-cloud-sizing.md
fruch: Agreed with pehala to postpone dedicated k8s work. Added a note to the docs: k8s backends use the same `*_instance_type` params for worker node pools, so constraint resolution should work out of the box once tested — just needs validation in a follow-up.

docs/cross-cloud-sizing.md
fruch: Good catch. The approach now is: keep constraints loose — `vcpu: 8` is enough for db role since preferred families already filter to the right instances with correct memory. Task 11 (migrate test configs) includes QA that validates every default resolves correctly across all clouds before shipping. The `translate-size` CLI makes this easy to verify.

docs/cross-cloud-sizing.md
fruch: Done. Updated preferred families: `db/aws: ["i8g", "i7i", "i4i"]`. i7i is now the first x86 fallback.

For dense options (i8ge, i7ieq, i3en) — agreed with fruch and pehala: cases needing specific dense instances can use literal strings. Keeping preferred families focused on the common path.

docs/cross-cloud-sizing.md
fruch: Done. Plain int now always means minimum GB: `memory: 32` equals `memory >= 32 GB`. Same for disk. No units needed — GB is the universal unit. Advanced operators (`>`, ranges) still available for power users who need them.

fruch:
**Note:** We should consider adding n2 family with local disk attached as an option for GCE — it may offer better pricing than z3-highmem instances for equivalent specs.

docs/cross-cloud-sizing.md
fruch: the catalog is not much small and limited as this comment suggests

docs/cross-cloud-sizing.md
fruch: now as @roydahan suggested we will have a configuration yaml to define all the family and order how the logic would work.

later we'll even put it in keystore so we can change the configuration across all branch immediately (like dropping some families)

pehala: I dont think think keystore is a good idea. It wont be transparent when we change it. Changing it should be a concious and visibile decision, so PR with backports make most sense to me, it should be pretty easy to do as well. Keeping it in a configuration yaml is great

fruch: also if we introduce new logic, we'll need to backport that...

fruch: BTW the reason is for ability to immediately control some part of it.
(for example it's in the source, it won't be affecting people development branches and no much affect on what's running in scylla-staging)

fruch: but that would be for later stages

pehala: > also if we introduce new logic, we'll need to backport that...

We will, but we will have to do that regardless of where the config will live. Additional backport from adding keystore support should have minimal conflicts

> BTW the reason is for ability to immediately control some part of it.
> (for example it's in the source, it won't be affecting people development branches and no much affect on what's running in scylla-staging)

We can merge small config changes into SCT including backports pretty quickly, so not seeing how the keystore is better. 

regardless, I would not put keystore into this plan yet, lets implement it with the config yaml and then do the keystore as a followup

PR #14522, opened 2026-05-06
fix(jenkins-pipeline): add missing SCT_BILLING_PROJECT to perfRegressionParallelPipeline

pehala:
## Summary
- `perfRegressionParallelPipeline.groovy` declared a `billing_project` parameter but never mapped it to `SCT_BILLING_PROJECT` in the environment block, causing billing project selection to be silently ignored for perf regression jobs.
- All other pipelines (longevity, rolling upgrade, manager, etc.) correctly set this env var.


pehala:
Too slow, https://github.com/scylladb/scylla-cluster-tests/pull/14516 already exists and I missed it

PR #14529, opened 2026-05-06
fix(gce): normalize GCE version tag for Argus consistency

pehala:
## Summary

- Fix GCE version tag normalization for Argus: after `.replace("-", ".")` converts the GCE tag to dot-separated format, apply `re.sub(r"\.(?=rc\d|dev|alpha|beta)", "~", ...)` before submitting to Argus so pre-release versions (e.g., `2025.4.0.rc5`) are stored as the canonical `2025.4.0~rc5`
- The `.replace("-", ".")` transform itself is preserved — it produces the GCE dot-separated format that `ComparableScyllaVersion` requires for parsing dev builds with date/commit metadata
- Add unit tests verifying Argus receives consistent version strings across all 4 cloud backends (AWS, GCE, Azure, OCI)
- Clarify AGENTS.md unit test guidelines to discourage all class-based test patterns

## Root Cause

GCE image labels forbid `~`, so they store versions with `-` (e.g., `2025.4.0-rc5`). The blanket `.replace("-", ".")` converted this to `2025.4.0.rc5` for `ComparableScyllaVersion` compatibility. This dot-separated version was then submitted directly to Argus, while running nodes report `2025.4.0~rc5` via dpkg. The Argus dashboard displayed both as separate entries in different widgets.

## Fix

Normalize the version string only for Argus submission (not for `artifact_scylla_version` which feeds `ComparableScyllaVersion`). A targeted regex replaces `.` with `~` only before known pre-release identifiers (`rc`, `dev`, `alpha`, `beta`), leaving dev builds with date/commit metadata untouched.

## Testing

- 18 parametrized unit tests covering RC, dev, GA, and enterprise versions across all backends, plus GCE-specific dev builds with date/commit metadata
- Tests exercise the real `update_argus_with_version` code path (ARGUS_VERSION_RE parsing, Package creation) with only the Argus HTTP client mocked
- All pass locally

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -3958,7 +3958,11 @@ class SCTConfiguration(BaseModel):
             _is_enterprise = is_enterprise(scylla_version)
         self.artifact_scylla_version = scylla_version
         self.is_enterprise = _is_enterprise
-        self.update_argus_with_version(scylla_version, "scylla-server-target")
+        # For Argus, normalize the GCE dot-separated format back to '~' for pre-release
+        # identifiers so it matches what the running node reports via package version.
+        # GCE tags use '-' (converted to '.') because '~' is not allowed in GCE labels.
+        argus_version = re.sub(r"\.(?=rc\d|dev|alpha|beta)", "~", scylla_version)
+        self.update_argus_with_version(argus_version, "scylla-server-target")
 
         return scylla_version, _is_enterprise
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -225,13 +225,13 @@ High-risk methods with 5+ overrides: `add_nodes`, `_create_instances`, `wait_for
 
 ### Unit Testing Guidelines
 
-**IMPORTANT: Use pytest style, NOT unittest.TestCase**
+**IMPORTANT: Use pytest style, NOT class-based tests**
 
 All unit tests in `unit_tests/` should follow pytest conventions:
 
-**❌ DON'T:** `class TestMyFeature(unittest.TestCase)` with `setUp()` and `self.assertEqual()`
+**❌ DON'T:** Use `class Test*` groupings — whether `unittest.TestCase` subclasses or plain pytest classes. Avoid `setUp()`, `self.assertEqual()`, and class methods for shared logic.
 
-**✅ DO:** Use pytest functions, fixtures, and simple `assert` statements
+**✅ DO:** Use top-level `def test_*` functions, fixtures, and simple `assert` statements. Use module-level helper functions for shared setup logic.
 
 **Key Principles:**
 
--- a/unit_tests/unit/test_version_consistency_across_backends.py
+++ b/unit_tests/unit/test_version_consistency_across_backends.py
@@ -1,0 +1,143 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2025 ScyllaDB
+
+"""
+Verify that the version submitted to Argus is consistent across all cloud backends.
+
+GCE image labels cannot contain '~', so they use '-' as separator (e.g., '2025.4.0-rc5').
+After .replace("-", ".") this becomes '2025.4.0.rc5'. The Argus submission must normalize
+this back to '~' format so it matches what nodes report via package versions.
+"""
+
+import unittest.mock
+
+import pytest
+
+from sdcm import sct_config
+
+
+@pytest.fixture(autouse=True)
+def minimal_config(monkeypatch):
+    monkeypatch.setenv("SCT_CONFIG_FILES", "unit_tests/test_configs/minimal_test_case.yaml")
+
+
+@pytest.fixture()
+def mock_argus_client():
+    """Mock the Argus client, letting update_argus_with_version run its real logic.
+
+    Lets update_argus_with_version run its ARGUS_VERSION_RE parsing logic,
+    but mocks the TestConfig/argus client to capture the final version string.
+    """
+    client = unittest.mock.MagicMock()
+    mock_test_config = unittest.mock.MagicMock()
+    mock_test_config.argus_client.return_value = client
+
+    with unittest.mock.patch("sdcm.sct_config.TestConfig", return_value=mock_test_config):
+        yield client
+
+
+@pytest.fixture()
+def resolve_version(monkeypatch):
+    """Set up backend config and resolve version."""
+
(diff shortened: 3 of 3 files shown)


pehala:
@k0machi I traced the issue of duplicate tags to GCE backend, so fixing it here should fix it in Argus as well

sdcm/sct_config.py
```diff
@@ -3958,7 +3958,11 @@ def get_version_based_on_conf(self):
             _is_enterprise = is_enterprise(scylla_version)
         self.artifact_scylla_version = scylla_version
         self.is_enterprise = _is_enterprise
-        self.update_argus_with_version(scylla_version, "scylla-server-target")
+        # For Argus, normalize the GCE dot-separated format back to '~' for pre-release
+        # identifiers so it matches what the running node reports via package version.
+        # GCE tags use '-' (converted to '.') because '~' is not allowed in GCE labels.
+        argus_version = re.sub(r"\.(?=rc\d|dev|alpha|beta)", "~", scylla_version)
```
fruch, line 3964: I'll put this in version_utils.py, not here

fruch:
> @k0machi I traced the issue of duplicate tags to GCE backend, so fixing it here should fix it in Argus as well

please point to the argus issue you are trying to solve

it not sure it would solve the thing you want it to solve.

we have duplication by design, version taken from images and such, and version from the running scylla process

pehala:
> > @k0machi I traced the issue of duplicate tags to GCE backend, so fixing it here should fix it in Argus as well
> 
> please point to the argus issue you are trying to solve
> 
> it not sure it would solve the thing you want it to solve.
> 

Different versions can be seen here: https://argus.scylladb.com/view/scylla-2025-4-2025-4-0. You can see that the version list contains both 2025.4.0.rc5 and 2025.4.0~rc5. The 2025.4.0.rc5 version is only for GCE backend test, rest use `2025.4.0~rc5`. Given Argus doesnt do any validation on what it receives, the reason for 2025.4.0.rc5 is in SCT. Given we care about it only for argus submit (it didnt cause issue elsewhere yet) I put it there only

fruch:
> > > @k0machi I traced the issue of duplicate tags to GCE backend, so fixing it here should fix it in Argus as well
> > 
> > 
> > please point to the argus issue you are trying to solve
> > it not sure it would solve the thing you want it to solve.
> 
> Different versions can be seen here: [argus.scylladb.com/view/scylla-2025-4-2025-4-0](https://argus.scylladb.com/view/scylla-2025-4-2025-4-0). You can see that the version list contains both 2025.4.0.rc5 and 2025.4.0~rc5. The 2025.4.0.rc5 version is only for GCE backend test, rest use `2025.4.0~rc5`. Given Argus doesnt do any validation on what it receives, the reason for 2025.4.0.rc5 is in SCT. Given we care about it only for argus submit (it didnt cause issue elsewhere yet) I put it there only


argus has both versions:
https://argus.scylladb.com/tests/scylla-cluster-tests/f291d0d1-0112-47f0-9455-932d20ddfd49/packages

the fact we are passing on labels version is for a fallback, for cases test didn't really manager to boot the instances 
argus shouldn't be using it for those filters, or should align it for it's usage.

pehala:
> argus has both versions: https://argus.scylladb.com/tests/scylla-cluster-tests/f291d0d1-0112-47f0-9455-932d20ddfd49/packages
> 
> the fact we are passing on labels version is for a fallback, for cases test didn't really manager to boot the instances argus shouldn't be using it for those filters, or should align it for it's usage.

It has both in packages, but there is a clear inconsistency between GCE and rest of the clouds:
GCE (2026.3.0.dev.0.20260505.c00fee0316c8): https://argus.scylladb.com/tests/scylla-cluster-tests/77e8c9e1-e578-4a79-a7ea-bc3065b9c20a/packages
Azure (2026.3.0~dev): https://argus.scylladb.com/tests/scylla-cluster-tests/87adee8c-53b4-4940-a695-9bef8f00b4d6/packages

This means we submit things differently from SCT, maybe the fix should be targeted elsewhere though

fruch:
> > argus has both versions: https://argus.scylladb.com/tests/scylla-cluster-tests/f291d0d1-0112-47f0-9455-932d20ddfd49/packages
> > the fact we are passing on labels version is for a fallback, for cases test didn't really manager to boot the instances argus shouldn't be using it for those filters, or should align it for it's usage.
> 
> It has both in packages, but there is a clear inconsistency between GCE and rest of the clouds: GCE (2026.3.0.dev.0.20260505.c00fee0316c8): https://argus.scylladb.com/tests/scylla-cluster-tests/77e8c9e1-e578-4a79-a7ea-bc3065b9c20a/packages Azure (2026.3.0~dev): https://argus.scylladb.com/tests/scylla-cluster-tests/87adee8c-53b4-4940-a695-9bef8f00b4d6/packages
> 
> This means we submit things differently from SCT, maybe the fix should be targeted elsewhere though

maybe at the source, in SMI. but it might need to add it as metadata and not only label, and then align it in SCT.
SCT is just passing from image / artifact, to argus.

I still think Argus should be able to normalize it, to fit the way it's using it.

pehala:
> maybe at the source, in SMI. but it might need to add it as metadata and not only label, and then align it in SCT. SCT is just passing from image / artifact, to argus.
> 
> I still think Argus should be able to normalize it, to fit the way it's using it.

I think the inconsistency is wrong regardless of Argus. I do agree that argus should normalize it regardless but I think we should fix this issue at the source as well

fruch:
> > maybe at the source, in SMI. but it might need to add it as metadata and not only label, and then align it in SCT. SCT is just passing from image / artifact, to argus.
> > I still think Argus should be able to normalize it, to fit the way it's using it.
> 
> I think the inconsistency is wrong regardless of Argus. I do agree that argus should normalize it regardless but I think we should fix this issue at the source as well

original label looks like:
```
scylla_version: 2025-4-0-rc0-0-20251001-6969918d3151
```

SCT fixing it with a replace, so far:
```
            scylla_version = self._require_scylla_version_tag(
                tags=tags,
                resource_label="GCE image",
                resource_id=images[0],
                tag_keys=("scylla_version",),
                resource_type="image",
                resource_id_label="image name",
            ).replace("-", ".")
```

in SMI we add it to the description, in the original format
and SCT can extract it from the description of the image 

ugly, but SCT won't need to "fix" anything.

fruch:
> > > maybe at the source, in SMI. but it might need to add it as metadata and not only label, and then align it in SCT. SCT is just passing from image / artifact, to argus.
> > > I still think Argus should be able to normalize it, to fit the way it's using it.
> > 
> > 
> > I think the inconsistency is wrong regardless of Argus. I do agree that argus should normalize it regardless but I think we should fix this issue at the source as well
> 
> original label looks like:
> 
> ```
> scylla_version: 2025-4-0-rc0-0-20251001-6969918d3151
> ```
> 
> SCT fixing it with a replace, so far:
> 
> ```
>             scylla_version = self._require_scylla_version_tag(
>                 tags=tags,
>                 resource_label="GCE image",
>                 resource_id=images[0],
>                 tag_keys=("scylla_version",),
>                 resource_type="image",
>                 resource_id_label="image name",
>             ).replace("-", ".")
> ```
> 
> in SMI we add it to the description, in the original format and SCT can extract it from the description of the image
> 
> ugly, but SCT won't need to "fix" anything.

something like: https://github.com/scylladb/scylla-machine-image/pull/949

PR #14537, opened 2026-05-06
fix(apt-keys): add fallback keyservers and HTTPS fallback to fetch_apt_keys

fruch:
## Summary

Fixes SCT-320. \`fetch_apt_keys()\` was using a single HKP keyserver (\`hkp://keyserver.ubuntu.com:80\`) with no fallback, causing transient failures when that keyserver was unreachable.

## Changes

- **\`sdcm/cluster.py\`** — \`fetch_apt_keys()\` now tries 3 HKP keyservers in order:
  1. \`hkp://keyserver.ubuntu.com:80\`
  2. \`hkp://keys.openpgp.org\`
  3. \`hkp://pgp.mit.edu\`

  If all HKP attempts fail, falls back to HTTPS (\`https://keyserver.ubuntu.com/pks/lookup?...\`). Raises \`NodeSetupFailed\` only if every source fails.

- **\`unit_tests/unit/test_fetch_apt_keys.py\`** — 4 new pytest tests covering:
  - First keyserver succeeds
  - First fails, second succeeds
  - All HKP fail, HTTPS fallback succeeds
  - All sources fail → \`NodeSetupFailed\` raised

## Testing

\`\`\`
uv run sct.py unit-tests -t test_fetch_apt_keys.py
\`\`\`

All 4 tests pass. Pre-commit checks clean.

### DIFF
--- a/unit_tests/unit/test_fetch_apt_keys.py
+++ b/unit_tests/unit/test_fetch_apt_keys.py
@@ -1,0 +1,161 @@
+from unittest import mock
+
+import pytest
+from invoke.runners import Result
+
+from sdcm.cluster import BaseNode, NodeSetupFailed
+from sdcm.remote import shell_script_cmd
+
+
+APT_KEY = "AABBCCDD11223344"
+HKP_KEYSERVERS = [
+    "hkp://keyserver.ubuntu.com:80",
+    "hkp://keys.openpgp.org",
+    "hkp://pgp.mit.edu",
+]
+
+
+@pytest.fixture()
+def node():
+    n = mock.MagicMock()
+    n.parent_cluster.params.get.return_value = [APT_KEY]
+    return n
+
+
+def _make_result(ok):
+    r = mock.MagicMock(spec=Result)
+    r.ok = ok
+    return r
+
+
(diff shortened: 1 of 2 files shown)


unit_tests/unit/test_fetch_apt_keys.py
```diff
@@ -0,0 +1,127 @@
...
+        None,  # mkdir
+        success_result,  # gpg recv-keys from first keyserver
+        None,  # export
+        None,  # cleanup
+    ]
+
+    BaseNode.fetch_apt_keys(node)
+
+    # Verify sudo was called 4 times (mkdir, recv-keys, export, cleanup)
+    assert node.remoter.sudo.call_count == 4
```
pehala: I would rather assert the exact calls in all tests if possible

pehala (dismissed):

fruch:
Addressed review: rewrote all tests to assert exact calls using `assert_has_calls` with full `mock.call(...)` objects. Also mocked `uuid.uuid4` to make the temp keyring path deterministic.

pehala (approved):

vponomaryov (approved):
LGTM

PR #14538, opened 2026-05-06
fix(apt-install): add Acquire::http::Timeout and Acquire::Retries to apt-get calls

fruch:
## Summary

- Add `-o Acquire::http::Timeout=60 -o Acquire::Retries=3` to all `apt-get update` and `apt-get install` calls in `sdcm/cluster.py`
- Prevents tests from hanging indefinitely (30+ hours observed) when package repos or keyservers are unreachable
- Adds `package-installation` skill documenting mandatory options for apt-get, yum, and zypper commands

## Root Cause

SCT test `longevity-twcs-48h-master` hung for 30+ hours during scylla-manager installation on the monitor node. The GPG keyserver returned "No data" and `apt-get` had no network timeout configured, causing an indefinite hang.

## Changes

| File | Change |
|------|--------|
| `sdcm/cluster.py` | Added `Acquire::http::Timeout=60` and `Acquire::Retries=3` to all apt-get calls |
| `skills/package-installation/SKILL.md` | New skill documenting package installer best practices |
| `AGENTS.md` | Registered new skill |

## Ref

SCT-290

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -96,6 +96,7 @@ from sdcm.remote.remote_long_running import run_long_running_cmd
 from sdcm.remote.remote_file import remote_file, yaml_file_to_dict, dict_to_yaml_file
 from sdcm import wait, mgmt
 from sdcm.sct_config import SCTConfiguration
+from sdcm.utils.apt import apt_cmd
 from sdcm.sct_events.continuous_event import ContinuousEventsRegistry
 from sdcm.sct_events.system import AwsKmsEvent
 from sdcm.snitch_configuration import SnitchConfig
@@ -2360,7 +2361,7 @@ class BaseNode(AutoSshContainerMixin):
 
         if self.distro.is_debian_like:
             self.fetch_apt_keys()
-            self.remoter.sudo("apt-get update", ignore_status=True)
+            self.remoter.sudo(apt_cmd("update"), ignore_status=True)
 
     def fetch_apt_keys(self):
         """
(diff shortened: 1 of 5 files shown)


sdcm/cluster.py
```diff
@@ -2360,7 +2360,7 @@ def download_scylla_manager_repo(self, scylla_repo: str) -> None:
 
         if self.distro.is_debian_like:
             self.fetch_apt_keys()
-            self.remoter.sudo("apt-get update", ignore_status=True)
+            self.remoter.sudo("apt-get -o Acquire::http::Timeout=60 -o Acquire::Retries=3 update", ignore_status=True)
```
pehala: Could we set it up as a default somwhere, so we dont have to update every usage of it?

fruch:
Addressed @pehala's feedback: created `sdcm/utils/apt.py` with an `apt_cmd()` utility function that centralizes all apt-get defaults (timeouts, retries, dpkg lock, confold/confdef). All apt-get calls in `cluster.py` now use it.

Usage:
```python
from sdcm.utils.apt import apt_cmd

# Full command:
self.remoter.sudo(apt_cmd("update"), ignore_status=True)
self.remoter.sudo(apt_cmd("install -y scylla-manager-agent"))

# Override specific option:
apt_cmd("install -y pkg", options={"DPkg::Lock::Timeout": "300"})

# As prefix (used by install_package):
pkg_cmd = apt_cmd()  # returns 'apt-get -o ... -o ...'
self.remoter.sudo(f"{pkg_cmd} install -y {package_name}")
```

CI is green: all unit tests, precommit, provision_aws, and provision_docker pass.

pehala (approved):

PR #14541, opened 2026-05-07
fix(cql): use AsyncioConnection to prevent [Errno 9] Bad file descriptor

jsmolar:
SCT creates many short-lived ClusterDriver instances during setUp. The default libev reactor is shared across all Cluster objects in the process, causing stale file descriptor watchers to accumulate across shutdown/reconnect cycles. When the OS reuses a closed FD, the stale watcher fires on the wrong socket producing NoHostAvailable: ConnectionShutdown('[Errno 9] Bad file descriptor').

Switching to AsyncioConnection gives each Cluster its own isolated event loop, eliminating the shared state. Workaround for scylladb/python-driver#614.

This issue prevented reliable spin-up of 12-node GCP clusters https://argus.scylladb.com/tests/scylla-cluster-tests/ffb076b5-dd08-4efd-ba4b-db2594aeb7af; this workaround fixes that.

### Testing

- [ ] Test is running

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -47,6 +47,7 @@ import packaging.version
 
 import yaml
 import requests
+from cassandra.io.asyncioreactor import AsyncioConnection
 from paramiko import SSHException
 from tenacity import RetryError
 from invoke.exceptions import UnexpectedExit, Failure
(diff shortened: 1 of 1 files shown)


fruch:
@jsmolar  don't it's gonna cause more harm, the Asyncio isn't production ready... at all. it marked as experimental for very good reasons. i.e. it's not working with SSL as far as we know.

i.e. its a waste of your time

jsmolar:
> @jsmolar don't it's gonna cause more harm, the Asyncio isn't production ready... at all. it marked as experimental for very good reasons. i.e. it's not working with SSL as far as we know.
> 
> i.e. its a waste of your time

ok, closing this one.

PR #14551, opened 2026-05-07
ci(gce-trigger): drop hardcoded availability_zone=a from tier1-gce-custom-time-trigger

fruch:
## Summary
- Remove hardcoded `availability_zone=a` from the tier1-gce-custom-time-trigger job
- Allows GCE to auto-select availability zone, improving flexibility and avoiding zone-a capacity constraints (the default region we use, doesn't have zone a !!!)


fruch:
in 2026.2 for now fixed it manually (probably what happened in 2026.1...)

vponomaryov (approved):
LGTM

PR #14559, opened 2026-05-08
feature(monkey): Update AddRemoveDCNemesis

cezarmoise:
Part of the RF Changes feature.
RF can now be changed from 0 to N in a single step.
- move nemesis in its own module
- add 2 nodes (instead of 1 to the new DC), se the feature can be tested
- write some data before adding the new dc, so there is something to migrate
- clean up and refactor the code for clarity, it has multiple helper functions

Fixes: SCYLLADB-526

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/eb1371ae-a8f0-4c03-97b5-f58988326ba4/nemesis
  Needs `nemesis_during_prepare=false`, otherwise stress commands of the longevity might end up on the new DC.
  [Old version](https://argus.scylladb.com/tests/scylla-cluster-tests/a62d992a-d095-43bb-b568-a97d5853c9a4) had same issue.

### Reminders

### DIFF
--- a/sdcm/nemesis/monkey/add_remove_dc.py
+++ b/sdcm/nemesis/monkey/add_remove_dc.py
@@ -1,28 +1,127 @@
-from contextlib import ExitStack
-from typing import List
+from collections.abc import Iterator
+from contextlib import ExitStack, contextmanager
 
 from sdcm.cluster import BaseNode
 from sdcm.exceptions import KillNemesis, UnsupportedNemesis
 from sdcm.nemesis import NemesisBaseClass
 from sdcm.provision.scylla_yaml import SeedProvider
-from sdcm.sct_events.system import InfoEvent
-from sdcm.utils.decorators import skip_on_capacity_issues
+from sdcm.utils.decorators import retrying, skip_on_capacity_issues
+from sdcm.utils.features import is_tablets_feature_enabled
 from sdcm.utils.replication_strategy_utils import (
     NetworkTopologyReplicationStrategy,
     ReplicationStrategy,
-    SimpleReplicationStrategy,
     temporary_replication_strategy_setter,
 )
 from sdcm.wait import wait_for_log_lines
 
 
 class AddRemoveDcNemesis(NemesisBaseClass):
+    """
+    https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-dc-to-existing-dc.html
+    https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/decommissioning-data-center.html
+
+    1. Create and populate a new keyspace
+    2. Add some new nodes in a new datacenter
+    3. Ensure relevant keyspaces are using NetworkTopologyStrategy
+    4. Alter the keyspaces to add the new datacenter with the appropriate replication factor
+    5. Run rebuild on the new datacenter nodes, if vnodes
+    6. Run a full cluster repair
+    7. Run some writes and reads to verify the new datacenter is working
+    8. Alter the keyspaces to set replication factor to 0 for the new datacenter
+    9. Decommission the new datacenter nodes
+    10. Verify the keyspace is healthy after decommissioning the new datacenter
+    """
+
     disruptive = True
     limited = True
     topology_changes = True
 
-    def _configure_new_dc(self, node: BaseNode):
-        """Configure node for new datacenter before Scylla starts."""
+    def __init__(self, *args, **kwargs):
+        """Initialize new-datacenter state and keyspace defaults."""
+        super().__init__(*args, **kwargs)
+        self.new_nodes: list[BaseNode] = []
+        self.initial_dc_name: str = self.datacenters[0]
+        self.new_ks_name: str = "keyspace_new_dc"
+        self.new_ks_rf: int = len(self.runner.cluster.racks)
+
+    @property
+    def num_nodes_in_new_dc(self) -> int:
+        """Return new datacenter size based on supported RF change size."""
+        if self.runner.cluster.is_features_enabled_on_node(
+            node=self.runner.target_node, feature_list=["KEYSPACE_MULTI_RF_CHANGE"]
+        ):
+            return 2
+        return 1
+
+    @property
+    def system_keyspaces(self) -> list[str]:
+        """Return system keyspaces"""
+        system_keyspaces = ["system_distributed", "system_traces"]
+        if (
+            not self.runner.target_node.raft.is_consistent_topology_changes_enabled
+        ):  # auth-v2 is used when consistent topology is enabled
+            system_keyspaces.append("system_auth")
+        return system_keyspaces
+
+    @contextmanager
+    def temporary_system_keyspaces_network_topology_strategy(self) -> Iterator[None]:
+        """Temporarily switch system keyspaces to NetworkTopologyStrategy."""
+        with (
+            temporary_replication_strategy_setter(self.runner.target_node) as ntrs_setter,
+            self.runner.action_log_scope("Temporarily switch system keyspaces to NetworkTopologyStrategy"),
+        ):
+            for keyspace in self.system_keyspaces:
+                old_rs = ReplicationStrategy.get(self.runner.target_node, keyspace)
+                if not isinstance(old_rs, NetworkTopologyReplicationStrategy):
+                    new_rs = NetworkTopologyReplicationStrategy(**{self.initial_dc_name: old_rs.replication_factors[0]})
+                    ntrs_setter(**{keyspace: new_rs})
+            yield
+
+    @contextmanager
+    def temporary_new_dc_replication_factors(self) -> Iterator[None]:
+        """Temporarily update replication factors for the new datacenter."""
+        with (
+            temporary_replication_strategy_setter(self.runner.target_node) as new_dc_setter,
+            self.runner.action_log_scope("Temporarily update replication factors for new datacenter"),
+        ):
+            for keyspace in self.system_keyspaces + [self.new_ks_name]:
+                strategy = ReplicationStrategy.get(self.runner.target_node, keyspace)
+                strategy.replication_factors_per_dc.update({self.new_dc_name: self.num_nodes_in_new_dc})
+                new_dc_setter(**{keyspace: strategy})
+
+            # when context exits, set RF=0 for the new DC
+            for _, preserved_strategy in new_dc_setter.preserved.items():
+                preserved_strategy.replication_factors_per_dc[self.new_dc_name] = 0
+            yield
+
+    @property
+    def datacenters(self) -> list[str]:
+        """Return datacenter names currently reported by nodetool status."""
+        return list(self.runner.tester.db_cluster.get_nodetool_status().keys())
+
+    @property
+    def new_dc_name(self) -> str | None:
+        """Return the temporary nemesis datacenter name if it is registered."""
+        if has_suffix := [dc for dc in self.datacenters if dc.endswith("_nemesis_dc")]:
+            return has_suffix[0]
+        return None
+
+    @retrying(n=3, sleep_time=30, allowed_exceptions=(AssertionError,))
+    def assert_new_dc_registered(self) -> None:
+        """Assert that the temporary datacenter is visible in cluster status."""
+        assert self.new_dc_name, "new datacenter was not registered"
+
+    @retrying(n=3, sleep_time=30, allowed_exceptions=(AssertionError,))
+    def assert_new_dc_unregistered(self) -> None:
+        """Assert that the temporary datacenter is no longer visible in cluster status."""
+        assert not self.new_dc_name, "new datacenter was not unregistered"
+
+    def configure_new_dc(self, node: BaseNode) -> None:
+        """Configure node for new datacenter before Scylla starts.
+
+        Args:
+            node: New datacenter node to configure.
+        """
         with node.remote_scylla_yaml() as scylla_yml:
             scylla_yml.rpc_address = node.ip_address
             scylla_yml.seed_provider = [
@@ -41,31 +140,55 @@ class AddRemoveDcNemesis(NemesisBaseClass):
         with node.remote_cassandra_rackdc_properties() as properties_file:
             properties_file.update(**rackdc_value)
 
-    def _add_new_node_in_new_dc(self, is_zero_node=False) -> BaseNode:
+    def add_nodes_in_new_dc(self) -> None:
+        """Add and initialize temporary nodes configured to join a new datacenter."""
         add_node_func_args = {
-            "count": 2,
+            "count": self.num_nodes_in_new_dc,
             "dc_idx": 0,
             "rack": None,
             "enable_auto_bootstrap": True,
             "disruption_name": self.runner.current_disruption,
-            "after_config": self._configure_new_dc,
-            **({"is_zero_node": is_zero_node} if is_zero_node else {}),
+            "after_config": self.configure_new_dc,
         }
-        new_nodes = skip_on_capacity_issues(db_cluster=self.runner.tester.db_cluster)(self.runner.cluster.add_nodes)(
-            **add_node_func_args
-        )
+        with self.runner.action_log_scope("Add nodes in new DC"):
+            self.new_nodes = skip_on_capacity_issues(db_cluster=self.runner.tester.db_cluster)(
+                self.runner.cluster.add_nodes
+            )(**add_node_func_args)
         # wait_for_init() will call node_setup(), which executes the callback after config_setup()
-        self.runner.cluster.wait_for_init(node_list=new_nodes, timeout=900, check_node_health=False)
-        for new_node in new_nodes:
+        self.runner.cluster.wait_for_init(node_list=self.new_nodes, check_node_health=False)
+        for new_node in self.new_nodes:
             new_node.wait_node_fully_start()
         self.runner.monitoring_set.reconfigure_scylla_monitoring()
-        return new_nodes
 
-    def _create_new_keyspace(self, dc_rf3: str) -> None:
-        InfoEvent(message="Writing to new keyspace").publish()
+    def rebuild_node(self, new_node: BaseNode) -> None:
+        """Run nodetool rebuild on a node in the new datacenter.
+
+        Args:
+            new_node: New datacenter node to rebuild from the initial datacenter.
+        """
+        cmd = f"rebuild -- {self.initial_dc_name}"
+        with (
+            wait_for_log_lines(
+                node=new_node,
+                start_line_patterns=["rebuild.*started with keyspaces=", "Rebuild starts"],
+                end_line_patterns=["rebuild.*finished with keyspaces=", "Rebuild succeeded"],
+                start_timeout=60,
+                end_timeout=600,
+            ),
+            self.runner.action_log_scope(f"Run rebuild on {new_node.name} with cmd: {cmd}"),
+        ):
+            new_node.run_nodetool(sub_cmd=cmd, long_running=True, retry=0)
+
+    def rebuild_nodes_in_new_dc(self) -> None:
+        """Run nodetool rebuild on all new datacenter nodes."""
+        for new_node in self.new_nodes:
+            self.rebuild_node(new_node)
+
+    def create_new_dc_keyspace(self) -> None:
+        """Create and populate the keyspace used to validate add/remove DC flow."""
         write_cmd = (
-            f"cassandra-stress write no-warmup cl=ALL n=100000 -schema 'keyspace=keyspace_new_dc "
-            f"replication(strategy=NetworkTopologyStrategy,{dc_rf3}=3) "
+            f"cassandra-stress write no-warmup cl=ALL n=100000 -schema 'keyspace={self.new_ks_name} "
+            f"replication(strategy=NetworkTopologyStrategy,{self.initial_dc_name}={self.new_ks_rf}) "
             f"compression=LZ4Compressor compaction(strategy=SizeTieredCompactionStrategy)' "
             f"-mode cql3 native compression=lz4 -rate threads=5 -pop seq=1..100000 -log interval=5"
         )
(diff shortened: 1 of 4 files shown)


sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -1,5 +1,4 @@
 from contextlib import ExitStack
-from typing import List
```
Deexie: Please split the commit so that refactor is separate from other changes. Currently, it is hard to determine what's meaningful and what is a part of the cleanup

bhalevy:
@cezarmoise let's drive this PR to completion

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: Added https://github.com/scylladb/scylla-cluster-tests/pull/14559/commits/f0141e31bde3b8a858c1c1afdbcffb6dabc24faa that just updated the RF of the new DC from 1 to 2 (also adds 2 nodes instead of one).

Then https://github.com/scylladb/scylla-cluster-tests/pull/14559/commits/3d366453146c36dfa8fb1165495483b40606ba6e is just the refactor + unit tests

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,207 @@
...
+            with self.runner.action_log_scope("Add new keyspace"):
+                self.runner.tester.create_keyspace(
+                    self.new_ks_name, replication_factor={self.initial_dc_name: self.new_ks_rf}
+                )
+
+            with self.runner.action_log_scope("Add nodes in new DC"):
+                self.add_nodes_in_new_dc()
+            assert self.new_dc_name, "new datacenter was not registered"
+
+            # switch system keyspaces to NetworkTopologyStrategy
```
pehala: Why is this needed?

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -43,8 +43,9 @@ def _configure_new_dc(self, node: BaseNode):
 
     def _add_new_node_in_new_dc(self, is_zero_node=False) -> BaseNode:
         add_node_func_args = {
-            "count": 1,
+            "count": 2,
             "dc_idx": 0,
+            "rack": None,
```
Deexie, line 148: What happens when `"rack": None`? Is every node added to a different rack?

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: Just following the [procedure ](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-dc-to-existing-dc.html#add-new-dc)to the letter.
To ensure the system keyspaces are in `NetworkTopologyReplicationStrategy` for the next step.
The new ks already is because it is created like that.

sdcm/nemesis/monkey/add_remove_dc.py
Deexie: In line 58 you init one node only

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -156,7 +157,7 @@ def finalizer(exc_type, *_):
                     assert isinstance(strategy, NetworkTopologyReplicationStrategy), (
                         "Should have been already switched to NetworkStrategy"
                     )
-                    strategy.replication_factors_per_dc.update({new_dc_name: 1})
+                    strategy.replication_factors_per_dc.update({new_dc_name: 2})
```
Deexie: Is there any data in the keyspace at this point?

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: > What happens when "rack": None? Is every node added to a different rack?

yes, it does round robin over each rack.

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: > Is there any data in the keyspace at this point?

No. I will update so there is some data before altering.

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: > In line 58 you init one node only

Will fix.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/pull/14559/commits/b298478853d1328ff32ea9917a37a7faed7ee99e:
- fixed missing code for handling new nodes 
- write some data before adding the new DC

sdcm/nemesis/monkey/add_remove_dc.py
Deexie: > yes, it does round robin over each rack.

What does `each rack` mean? Do we create many racks somewhere? It is very important to put each node in different rack

sdcm/nemesis/monkey/add_remove_dc.py
Deexie: Does `2` get extended to a rack list? Is `enforce_rack_list` option set to true?

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/f0e7b7e307506b299c0583f2f3ada18a28db70c0:
- fixed copilot comments (typo, type annotation)
- fixed decommission being outside the context for updating system keyspaces strategies

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: Yes, that's what it does. When rack=None, it puts one node in the 1st rack, then 1 node in the 2nd rack, and so on, and when it reaches the last rack it circles back to the 1st rack.
So in our case, with 2 nodes, one will be in RACK0, and one in RACK1. 

See https://argus.scylladb.com/tests/scylla-cluster-tests/cb64b7ec-104b-4dc0-9ac7-a0d274c6a2f2/resources
image
each pair of node added is one in RACK0 and one in RACK1

Deexie: Great!

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: > Does 2 get extended to a rack list? 

Not sure.
When it's updated, it's set like this
`{'class': 'NetworkTopologyStrategy', 'eu-west-1': ['RACK1', 'RACK2', 'RACK0'], 'add_remove_nemesis_dc': 2}"}`

> Is enforce_rack_list option set to true?

The legacy option `rf_rack_valid_keyspaces=true`.
`enforce_rack_list` is not used in SCT yet from what I can see.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/ca6602bf6882f53f08ece8bc9fe922f8e4e49dba:
- add info event with replication strategy of the new keyspace
- simplify the new keyspace's initial RF to be the number or racks in the cluster instead of node math.

sdcm/nemesis/monkey/add_remove_dc.py
Deexie: Fine, `rf_rack_valid_keyspaces` would be enough for this to work

cezarmoise: Update: It gets expanded
> (InfoEvent Severity.NORMAL) period_type=not-set event_id=894d8284-748f-4b7b-89df-2e4fe170a83e: message=Replication strategy for keyspace_new_dc is {'class': 'NetworkTopologyStrategy', 'add_remove_nemesis_dc': ['RACK0', 'RACK1'], 'eu-west-1': ['RACK0', 'RACK2', 'RACK1']}

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/a35cd75c655df7ace900037042c5b95e9a121452: unit test fix

Deexie (approved):
Approve to `0->N->0`-related changes

sdcm/nemesis/monkey/add_remove_dc.py
fruch: where is the documentation (docstring) of this nemesis, that should be pointing to this offical guide ?

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,226 @@
...
+        write_thread = self.runner.tester.run_stress_thread(
+            stress_cmd=write_cmd, round_robin=True, stop_test_on_failure=False
+        )
+        self.runner.tester.verify_stress_thread(write_thread, error_handler=self.runner._nemesis_stress_failure_handler)
+        # flush data to ensure it is seen in monitoring
+        for node in self.runner.cluster.nodes:
+            with self.runner.action_log_scope(f"Flush data in {self.new_ks_name} on {node.name} node"):
+                node.run_nodetool(f"flush {self.new_ks_name}")
+
+    def write_to_multi_dc_keyspace(self) -> None:
```
fruch, line 204: nitpick: why in some function there a return signature annotions ? and in some there isn't ?

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,226 @@
...
+                start_line_patterns=["rebuild.*started with keyspaces=", "Rebuild starts"],
+                end_line_patterns=["rebuild.*finished with keyspaces=", "Rebuild succeeded"],
+                start_timeout=60,
+                end_timeout=600,
+            ),
+            self.runner.action_log_scope(f"Run rebuild on {new_node.name} with cmd: {cmd}"),
+        ):
+            new_node.run_nodetool(sub_cmd=cmd, long_running=True, retry=0)
+
+    def create_new_dc_keyspace(self):
```
fruch: where is the call to this one ?

fruch:
## Code Review — Comparison with Official Add-DC Procedure

I reviewed this PR against the [official ScyllaDB documentation for adding a new DC](https://docs.scylladb.com/manual/stable/operating-scylla/procedures/cluster-management/add-dc-to-existing-dc.html).

---

### Critical Issues

**1. `create_new_dc_keyspace()` is dead code — no pre-migration data written**

The PR description says *"write some data before adding the new dc, so there is something to migrate"*, but `disrupt()` never calls `create_new_dc_keyspace()`. It only calls `self.runner.tester.create_keyspace(...)` (empty keyspace). The method that writes 100K rows is orphaned — the nemesis doesn't actually test data migration via rebuild.

**2. Migrated data is never verified after rebuild**

Even if the prewrite is restored, `verify_multi_dc_keyspace()` reads `seq=100001..200000` (post-add writes), not `seq=1..100000` (pre-existing data). There's no proof that rebuild actually streamed old data to the new DC.

---

### Important Issues

**3. `new_ks_rf = len(self.runner.cluster.racks)` — semantic regression**

Previous code used `min(3, len(self.cluster.data_nodes))`. The new formula can collapse RF to **1** on single-rack clusters, making consistency-level checks trivial and rebuild nearly a no-op.

**4. Unconditional `nodetool rebuild` — not tablet-safe**

SCT already has `rebuild_or_repair()` specifically because rebuild isn't supported for tablet keyspaces. This PR always runs `nodetool rebuild`, which will fail or produce incorrect results if the keyspace has tablets enabled.

**5. Immediate `assert self.new_dc_name` after `add_nodes_in_new_dc()` is flaky**

`get_nodetool_status()` can lag behind actual topology changes. A brief wait/retry would prevent intermittent failures.

**6. System keyspaces switched to NTS *after* nodes join — out of official order**

The docs say: *"Make sure **all** your keyspaces are using NetworkTopologyStrategy"* **before** adding the new DC. This PR adds nodes first, then alters system keyspaces.

---

### Suggestions

**7.** Official docs mention updating `audit` keyspace if enabled — not addressed here.

**8.** The nested `temporary_replication_strategy_setter` rollback-by-mutation pattern (`preserved_strategy.replication_factors_per_dc[new_dc_name] = 0`) works but is fragile. An explicit ALTER to RF=0 would be clearer.

**9.** Unit tests: `cluster.racks` is a MagicMock so `new_ks_rf` resolves to `0`; tests validate call ordering but not semantic correctness (ranges written/read, tablet vs vnode path).

---

### Comparison vs Official Procedure

| Official Step | PR Status |
|---|---|
| Ensure all KS use NTS | ⚠️ Only system KS + new temp KS |
| Configure new nodes (seeds, snitch, rackdc) | ✅ |
| Start new nodes, verify join | ✅ |
| ALTER KS to include new DC | ✅ |
| `nodetool rebuild` on each new node | ⚠️ Not tablet-safe |
| Full cluster repair | ✅ |
| Update monitoring | ✅ |
| Write pre-existing data to migrate | ❌ Dead code |
| Verify migrated data after rebuild | ❌ Only verifies new writes |
| Handle tablet KS incrementally | ❌ Single-step (intentional per SCYLLADB-526, but no guard) |

---

### Strengths

- Clean module extraction with auto-discovery preserved
- Multi-node (2) new DC is more realistic than original single-node
- Proper `ExitStack` + finalizer pattern for cleanup
- Unit tests cover key paths (multi-region rejection, finalizer, system_auth conditional)

sdcm/nemesis/monkey/add_remove_dc.py
cezarmoise: Got lost in a rebase. Will fix.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/25fd5729b85c82a5dd2c8dc991d5468b9bb16260:
- added missing type annotations
- added missing call to `create_new_dc_keyspace`
- changed verify to read the entire data range
- only run rebuild with vnodes
- Added check for KEYSPACE_MULTI_RF_CHANGE feature

Official Step | PR Status | Response
-- | -- | --
Ensure all KS use NTS | ⚠️ Only system KS + new temp KS | Intentional, only those that will be altered. Migration other test KS is not wanted
Configure new nodes (seeds, snitch, rackdc) | ✅
Start new nodes, verify join | ✅
ALTER KS to include new DC | ✅
nodetool rebuild on each new node | ⚠️ Not tablet-safe | Fixed
Full cluster repair | ✅
Update monitoring | ✅
Write pre-existing data to migrate | ❌ Dead code | Fixed
Verify migrated data after rebuild | ❌ Only verifies new writes | Fixed
Handle tablet KS incrementally | ❌ Single-step (intentional per SCYLLADB-526, but no guard) | Added check for `KEYSPACE_MULTI_RF_CHANGE` feature

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/0cc969fefdde9c4e11d0ca4a98a580fc38a7f22d:
- fix `is_features_enabled_on_node` param order

cezarmoise:
Nemesis success with latest commit: https://argus.scylladb.com/tests/scylla-cluster-tests/eb1371ae-a8f0-4c03-97b5-f58988326ba4/nemesis

And with older scylla versions, the nemesis is skipped https://argus.scylladb.com/tests/scylla-cluster-tests/bd0ad8fc-1b67-4936-9223-55cd86534176/nemesis

sdcm/nemesis/monkey/add_remove_dc.py
fruch: see `disrupt_remove_node_then_add_node` as reference

@pehala where is the agreed location for the docs ? on the class ? on the disrupt function ?

cezarmoise: Added docs. https://github.com/scylladb/scylla-cluster-tests/commit/9203a2c6b6e3bf456b388482ed4a2b4e6e70de3f
On the class, like `modify_table.py`

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/9203a2c6b6e3bf456b388482ed4a2b4e6e70de3f:
- add docstring
- move the call to `verify_multi_dc_keyspace` from inside of `write_to_multi_dc_keyspace` to outside of it, so it's clear it happens. And the names of the methods no longer lie to you.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/382bb7f41feadd1364b39530552a3d7c24760c9f:
- fixes for coderabbit:
  - add docstrings to everything
  - instead of bare assert, use a method with `@retrying`, in case `nodetool status` does not update fast enough to get datacenter list

cezarmoise:
Checks failed with:
```
error: Claude Code returned an error result: Credit balance is too low
      at readMessages (/home/runner/work/_actions/anthropics/claude-code-action/787c5a0ce96a9a6cfb050ea0c8f4c05f2447c251/node_modules/@anthropic-ai/claude-agent-sdk/sdk.mjs:60:19184)
Log saved to /home/runner/work/_temp/claude-execution-output.json
Error: Action failed with error: SDK execution error: Error: Claude Code returned an error result: Credit balance is too low
Internal error: directory mismatch for directory "/home/runner/work/_actions/anthropics/claude-code-action/787c5a0ce96a9a6cfb050ea0c8f4c05f2447c251/tsconfig.json", fd 4. You don't need to do anything, but this indicates a bug.
Error: Process completed with exit code 1.
```

pehala (dismissed):

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,289 @@
...
+            "rack": None,
+            "enable_auto_bootstrap": True,
+            "disruption_name": self.runner.current_disruption,
+            "after_config": self.configure_new_dc,
+        }
+        self.new_nodes = skip_on_capacity_issues(db_cluster=self.runner.tester.db_cluster)(
+            self.runner.cluster.add_nodes
+        )(**add_node_func_args)
+        # wait_for_init() will call node_setup(), which executes the callback after config_setup()
+        self.runner.cluster.wait_for_init(node_list=self.new_nodes, timeout=900, check_node_health=False)
```
yarongilor: i'm not sure about this hard coded timeout value, where now we have multiple nodes.

cezarmoise: Agree, I other places `wait_for_init` is not given a timeout, so the default `1200` is used. Will remove it here as well.
900 was the value that the existing test had.

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,289 @@
...
+            "count": self.num_nodes_in_new_dc,
+            "dc_idx": 0,
+            "rack": None,
+            "enable_auto_bootstrap": True,
+            "disruption_name": self.runner.current_disruption,
+            "after_config": self.configure_new_dc,
+        }
+        self.new_nodes = skip_on_capacity_issues(db_cluster=self.runner.tester.db_cluster)(
+            self.runner.cluster.add_nodes
+        )(**add_node_func_args)
```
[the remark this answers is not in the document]
cezarmoise: This is wrong. No need to allocate since `add_nodes` is decorated with `mark_new_nodes_as_running_nemesis`

pehala (dismissed):

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,289 @@
...
+        """Initialize new-datacenter state and keyspace defaults."""
+        super().__init__(*args, **kwargs)
+        self.new_nodes: list[BaseNode] = []
+        self.num_nodes_in_new_dc: int = 2
+        self.initial_dc_name: str = self.datacenters[0]
+        self.new_ks_name: str = "keyspace_new_dc"
+        self.new_ks_rf: int = len(self.runner.cluster.racks)
+
+    @property
+    def system_keyspaces(self) -> list[str]:
```
jsmolar, line 57: I feel like `system_keyspaces` and `datacenters` properties should be in the common nemesis module. But that is for another PR perhaps.

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,289 @@
...
+        """Return system keyspaces that need temporary NetworkTopologyStrategy replication."""
```
jsmolar: " that need temporary NetworkTopologyStrategy replication." - This part is not neccesary here, I would rather add "auth-v2 is used when consistent topology is enabled" to the doc string.

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,289 @@
...
+        self.new_nodes = skip_on_capacity_issues(db_cluster=self.runner.tester.db_cluster)(
+            self.runner.cluster.add_nodes
+        )(**add_node_func_args)
+        # wait_for_init() will call node_setup(), which executes the callback after config_setup()
+        self.runner.cluster.wait_for_init(node_list=self.new_nodes, check_node_health=False)
+        for new_node in self.new_nodes:
+            new_node.wait_node_fully_start()
+        self.runner.monitoring_set.reconfigure_scylla_monitoring()
+
+    def rebuild_new_dc(self, new_node: BaseNode) -> None:
```
jsmolar: `rebuild_new_dc` -> `rebuild_node`
We are running a rebuild on the node, not the whole DC.

fruch (dismissed):
LGTM

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -0,0 +1,289 @@
...
+        ):
+            raise UnsupportedNemesis(
+                "Altering RF by more than 1 requires KEYSPACE_MULTI_RF_CHANGE feature to be enabled"
+            )
+
+        InfoEvent(message="Starting New DC Nemesis").publish()
+        with ExitStack() as context_manager:
+            context_manager.push(self.finalizer)
+
+            with self.runner.action_log_scope("Add new keyspace"):
```
jsmolar: Move `with self.runner.action_log_scope("Add new keyspace"):` into the function.

The disruption function is similar to a test function, so I prefer it to be brief and easy to read. This line does not improve readability.

Also apply for other similar cases.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/d2b302b7fdc1181c4f70c578f60d0bc4cc76cac5
- Cosmetic changes:
  - clean up the disrupt function
  - move action log code inside the methods
  - removed some unnecessary InfoEvents, most commands already log what they do
  - extracted the temporary replication strategy code into separate methods
-  Instead of skipping when `KEYSPACE_MULTI_RF_CHANGE`, use just 1 node in the new dc, i.e. the old version of the nemesis

Testing: 
- [x] `master:latest` -> should use 2 nodes in new DC | [Argus](https://argus.scylladb.com/tests/scylla-cluster-tests/1e751504-7ce7-431e-8fca-1d92d7b02dfd)
- [x] `2025.4` -> should use 1 node in the new DC | [Argus](https://argus.scylladb.com/tests/scylla-cluster-tests/74260816-a7e1-4f33-bb45-c32c239faca3)

pehala (approved):

cezarmoise:
ping @scylladb/storage-qa

jsmolar (approved):

PR #14576, opened 2026-05-10
feature(sizing): constraint-based cross-cloud instance sizing

fruch:
## Summary

Replace per-cloud `instance_type` hardcoding with constraint-based selection. Define hardware needs once, get the cheapest matching instance per cloud automatically.

```yaml
instance_type_db:
  vcpu: 16
  memory: 64
```

→ `i8g.4xlarge` (AWS), `z3-highmem-16` (GCE), `Standard_L16s_v4` (Azure), `DenseIO.E5.Flex` (OCI)

Existing literal strings (`instance_type_db: 'i4i.4xlarge'`) continue working unchanged.

## What's New

| Component | Path | Purpose |
|-----------|------|---------|
| Constraint parser + matcher | `sdcm/utils/instance_matcher.py` | Parse `vcpu`, `memory`, `disk`, `arch` constraints; match against catalog |
| Instance catalog model | `sdcm/utils/instance_catalog.py` | Load YAML specs + pricing per cloud |
| Catalog generators | `sdcm/utils/catalog_generator.py` | Fetch live pricing from AWS/GCE/Azure/OCI APIs |
| Config integration | `sdcm/sct_config.py` | `_resolve_instance_sizes()` — resolves dicts → strings before validation |
| CLI | `sct_sizing.py` | `sizing preview`, `resolve`, `catalog`, `update-catalog` |
| Cloud config | `data/instance_catalog/sizing_config.yaml` | Preferred families, role constraints, sort order |
| Tier1 migration | 12 test-case YAMLs | Converted to constraint syntax |
| Documentation | `docs/cross-cloud-sizing.md` + README link | User guide with migration instructions |

## Commits

1. **`feature(sizing): add constraint-based cross-cloud instance sizing`** — core engine, catalogs, config wiring, tests, docs
2. **`refactor(sizing): extract sizing CLI commands to sct_sizing.py`** — CLI module, externalized cloud config, agnostic defaults
3. **`config(sizing): migrate tier1 test configs to constraint syntax`** — tier1 YAML migration (separate commit for easy revert)
4. **`docs(sizing): add migrate-to-sizing skill`** — AI-assisted migration skill, catalog improvements

### Automated (all passing)
- 54+ unit tests covering parser, matcher, catalog model, config loading
- `unit_tests/unit/test_instance_matcher.py` — 40 tests
- `unit_tests/unit/test_instance_catalog.py` — 14 tests
- `unit_tests/unit/test_instance_sizing_config.py` — config validation

### Suggested Manual Verification

1. **Preview resolution** — verify the resolved instances look correct for a real config:
   ```bash
   uv run sct.py sizing preview test-cases/longevity/longevity-50GB-3days-authorization-and-tls-ssl.yaml
   ```

2. **Dry-run a test with constraint config** — confirm SCT resolves and provisions correctly:
   ```bash
   export SCT_SCYLLA_VERSION=2025.3.0
   uv run sct.py run-test longevity_test.LongevityTest.test_custom_time \
     --backend aws \
     --config test-cases/longevity/longevity-50GB-3days-authorization-and-tls-ssl.yaml
   ```
   Check the log output for `Resolved instance_type_db: ...` lines showing the selected instances.

3. **Fallback behavior** — set an impossible constraint and verify a clear error:
   ```bash
   uv run sct.py sizing resolve --vcpu 1024 --memory 8192 --cloud aws
   ```
   Should raise `ValueError` with "No matching instance found".

4. **Catalog freshness** — regenerate catalogs and diff:
   ```bash
   uv run sct.py sizing update-catalog --cloud all
   git diff data/instance_catalog/
   ```

### Backends Affected
- [x] AWS, GCE, Azure, OCI — instance resolution + catalog
- [ ] Docker — N/A (no instance types)
- [ ] k8s-eks/k8s-gke — uses underlying AWS/GCE resolvers

## Summary by CodeRabbit

* **New Features**
  * Cross-cloud constraint-based instance sizing with new CLI sizing commands (catalog, update-catalog, resolve, preview) and per-cloud instance catalogs.

* **Defaults / Configuration**
  * Replaced many hardcoded instance-type defaults with sizing constraint fields (sizing_db/sizing_loader/sizing_monitor) and a central sizing configuration driving selection.

* **Documentation**
  * Expanded cross-cloud sizing guide, migration workflow, and updated configuration docs/README link.

* **Tests**
  * Added unit and integration tests for catalog generation, matching, pricing, and sizing resolution.

[Review Change Stack](https://app.coderabbit.ai/change-stack/scylladb/scylla-cluster-tests/pull/14576?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


soyacz:
smthing wrong with this one - 52 commits

fruch:
> smthing wrong with this one - 52 commits

I was trying to stack it over the plan, didn't worked exactly as planned
I'll fix it soon

fruch:
## Self-Review Summary (Automated)

Ran 3 review agents (code-reviewer, silent-failure-hunter, pr-test-analyzer) against this PR.

### Results

**Code Reviewer & Test Analyzer**: Could not produce findings (ran against wrong checkout — the feature files live in a worktree).

**Silent Failure Hunter** — found 5 issues in `sdcm/sct_config.py`, **all pre-existing** (not introduced by this PR):

| # | Location | Severity | Issue |
|---|----------|----------|-------|
| 1 | `_get_normalized_arch` (~L3806) | HIGH | Bare `except Exception` returns wrong default arch silently |
| 2 | `_resolve_xcloud_version_tag` (~L3942) | CRITICAL | Unsupported xcloud tag type silently ignored via `pass` |
| 3 | `_resolve_xcloud_version_tag` (~L3931) | HIGH | Non-"latest" release tag value silently ignored |
| 4 | `_check_version_supplied` (~L4092) | MEDIUM | `unified_package` bypass with no validation/logging |
| 5 | AMI/image resolution (~L4442+) | MEDIUM | Overly broad `except Exception` (mitigated by re-raise) |

**Verdict**: No issues introduced by this PR. The `_resolve_instance_sizes()` method properly raises `ValueError` on resolution failure and logs warnings for fallback scenarios.

### Next steps
- Splitting tier1 config migration into its own commit for easy revert if needed

data/instance_catalog/azure.yaml
pehala, line 1: 10K lines for Azure? Do we need that?

fruch: we are doing very naive selection of families, and azure has lot of variants 
```
preferred_families:
  db:
  - Standard_L
  db_oracle:
  - Standard_L
  zero_token:
  - Standard_L
  loader:
  - Standard_F
  - Standard_D
  monitor:
  - Standard_D
```

it's 10K of generated code, mostly a dump from their API, not something you gonna read or edit
just a cache.

fruch: anyhow change it to regex now, and minimized the selection we could have

data/instance_catalog/aws.yaml
```diff
@@ -0,0 +1,749 @@
+cloud: aws
+cloud_defaults:
+  arch: arm64
+preferred_families:
+  db:
+  - i8ge
```
soyacz: if `ge` are preferred, then we'll never select `g`?
I think we should switch the order.

unit_tests/unit/test_instance_catalog.py
```diff
@@ -0,0 +1,222 @@
...
+
+from sdcm.utils.instance_catalog import InstanceCatalog, InstanceTypeInfo
+
+
+AWS_YAML = """\
+cloud: aws
+cloud_defaults:
+  arch: arm64
+preferred_families:
+  db: [i8g, i7i, i4i]
```
soyacz: this test does not include testing prefererence for i8g vs i8ge

fruch: this is just for unittest of the configuration, is working as expected

data/instance_catalog/aws.yaml
fruch: fix it, there was a mix of where exactly was the source of truth about the ordering.

it should be in the configuration file, not in those, that should be just the data extrated from the cloud provider

fruch:
example of how it use in other PR;
https://github.com/scylladb/scylla-cluster-tests/pull/14629/changes#r3234621204

data/instance_catalog/azure.yaml
fruch: now it's 

```yaml 
  azure:
    families:
      - Standard_L
      - Standard_D
      - Standard_E
      - Standard_F
    series_patterns:
      - "^Standard_L\\d+s_v[34]$"
      - "^Standard_D\\d+s_v[45]$"
      - "^Standard_D\\d+as_v[56]$"
      - "^Standard_D\\d+_v[45]$"
      - "^Standard_E\\d+s_v[45]$"
      - "^Standard_E\\d+as_v[56]$"
      - "^Standard_F\\d+s_v2$"
      - "^Standard_F\\d+as_v6$"
 ```

and the file we cache is much smaller

File: data/instance_catalog/gce.yaml
Comment (fruch) [inline data/instance_catalog/gce.yaml:1]: This one is also a bit big... need to think how to minimize it a bit

**Comment by fruch** [comment]
one thing we should consider as followup, is to change the outdated `sdcm/utils/pricing.py` to use those new catalogs introduced here

File: sct_sizing.py [hunk 3]
```diff
@@ -0,0 +1,459 @@
...
+                resolved.extend(pipeline_config.test_config)
+            else:
+                click.echo(f"Warning: no test_config found in {path}", err=True)
+        else:
+            resolved.append(path_str)
+    return resolved
+
+
+@sizing_group.command("preview", help="Preview how a config file resolves instance types across all clouds")
+@click.argument("config_files", nargs=-1, required=True, type=click.Path(exists=True))
```
fruch, line 193: would be nice if we could document clearly how to override part of the configuration with env var or maybe introduce other commandline options for it.

so we can for example evaluate different test duration of different sizing constraints

fruch: this is now fixed, and env var can be used

fruch:
@scylladb/qa-maintainers feed would be helpful, and it might even be useful for some of 2026.2 testing phases

sct_sizing.py
fruch: and also cli for the sizing and types selection

data/instance_catalog/sizing_config.yaml
```diff
@@ -0,0 +1,183 @@
...
+      arch: arm64
+    preferred_families:
+      db:
+        - i8g
+        - i8ge
+        - i7i
+        - i7ie
+        - i4i
+      db_oracle:
+        - i8ge
```
soyacz: I think i8g is enough as preferred one - let's keep the same order like db

data/instance_catalog/sizing_config.yaml
```diff
@@ -0,0 +1,183 @@
...
+        - i8g
+        - i7ie
+        - i7i
+        - i4i
+      zero_token:
+        - i8ge
+        - i8g
```
soyacz, line 98: alo i8g (small disk) is enough for zero_token nodes

data/instance_catalog/sizing_config.yaml
```diff
@@ -0,0 +1,183 @@
...
+        - i7ie
+        - i7i
+        - i4i
+      loader:
+        - c6i
```
soyacz, line 106: why not c7i?

data/instance_catalog/sizing_config.yaml
```diff
@@ -0,0 +1,183 @@
...
+  oci:
+    families:
+      - BM.DenseIO
+      - VM.DenseIO
+      - VM.Standard
+    cloud_defaults:
+      arch: x86_64
+    preferred_families:
+      db:
+        - DenseIO.E5
```
soyacz, line 176: this family is not listed in families - isn't it a problem?

defaults/test_default.yaml
```diff
@@ -125,9 +125,17 @@ authenticator_password: ''
 n_test_oracle_db_nodes: 1
 oracle_scylla_version: '2026.1'
 append_scylla_args_oracle: '--enable-cache false'
-instance_type_db_oracle: 'i8g.2xlarge'
+instance_type_db_oracle: ''
```
soyacz, line 128: we should add instance sizing constraints default for db oracle - removing it will break tests as we don't have it in test configs everywhere

sdcm/utils/catalog_generator.py
```diff
@@ -0,0 +1,1079 @@
...
+    "n2": {"vcpu_hr": 0.031833, "ram_gb_hr": 0.004167},
+    "n2d": {"vcpu_hr": 0.027333, "ram_gb_hr": 0.003667},
+    "c3": {"vcpu_hr": 0.034500, "ram_gb_hr": 0.004000},
+    # z3 is excluded: bundled pricing (includes local SSD) — see _GCE_Z3_BUNDLED_PRICES
+    "t2a": {"vcpu_hr": 0.019250, "ram_gb_hr": 0.004812},  # ARM-based
+}
+
+# z3 bundled on-demand prices (us-east1) — includes compute + local SSD.
+# Source: https://cloud.google.com/products/compute/pricing/storage-optimized
+_GCE_Z3_BUNDLED_PRICES: dict[str, float] = {
```
soyacz: I think such things should be in a  separate config file (yaml/json) not in the code - will be easier to update

sdcm/utils/cloud_catalog/catalog_generator.py
soyacz, line 1: I'd prefer putting these new files under utils/cloud_catalog module
But this is only preference

sdcm/utils/pricing.py
```diff
@@ -2,22 +2,50 @@
...
+        if catalog_dir.exists():
+            _CATALOG = InstanceCatalog.from_directory(catalog_dir)
+        else:
+            _CATALOG = InstanceCatalog()
+    return _CATALOG
+
+
+def _catalog_price(cloud: str, region: str, instance_type: str) -> float | None:
+    catalog = _get_catalog()
+    for inst in catalog.get_instances(cloud):
```
soyacz: we could have another method in catalog for that: `catalog.get_instance(cloud, instance_type)`

test-cases/PR-provision-test.yaml
```diff
@@ -14,7 +14,9 @@ prepare_write_cmd:  ["scylla-bench -workload=sequential -mode=write -max-rate=30
                     ]
 
 n_loaders: 1
-instance_type_db: 'i4i.large'
+sizing_db:
+  vcpu: '>=2'
```
soyacz, line 18: do we support `>=2` for cpu's?

soyacz (commented):
hard to review 8k loc's...

data/instance_catalog/sizing_config.yaml
fruch: we had much more refs to c6i in our configuration then c7i,

defaults/test_default.yaml
fruch: we added sizing for it.

test-cases/PR-provision-test.yaml
fruch: yes

data/instance_catalog/sizing_config.yaml
fruch: but I can add it

data/instance_catalog/gce.yaml
fruch: it's now small enough

sdcm/utils/cloud_catalog/catalog_generator.py
fruch: can be moved

data/instance_catalog/sizing_config.yaml
fruch: Fixed — reordered to match db (i8g first for db_oracle and zero_token).

data/instance_catalog/sizing_config.yaml
fruch: Fixed — i8g first for zero_token too.

data/instance_catalog/sizing_config.yaml
fruch: Good catch — added `DenseIO.E5` to the OCI families list.

sdcm/utils/catalog_generator.py
fruch: Done — moved the section map out of code. The scraping metadata now lives in the config. Also moved all sizing utils into `sdcm/utils/cloud_catalog/` module as you suggested.

sdcm/utils/pricing.py
fruch: Added `catalog.get_instance(cloud, instance_type)` method and updated pricing.py to use it.

sdcm/utils/cloud_catalog/catalog_generator.py
fruch: was moved to it's own module

fruch:
@scylladb/qa-maintainers one last round of reivew

data/instance_catalog/sizing_config.yaml
```diff
@@ -0,0 +1,187 @@
...
+# and cloud-specific defaults used by the instance matcher.
+#
+# This file is loaded alongside per-cloud catalog YAML files and drives
+# the matching logic in sdcm.utils.instance_matcher.select_instance().
+
+# Roles define how instance selection behaves for each cluster component.
+# - category: 'storage' (needs local disk) or 'compute' (no local disk)
+# - implicit_constraints: auto-applied constraints (user doesn't need to specify them)
+# - arch_source: 'cloud_default' (use cloud's default arch) or 'fixed' (use arch below)
+# - arch: architecture when arch_source is 'fixed'
```
soyacz, line 13: I suggest to provide a link to the docs explaining what to do when we want to extend with a new family

soyacz (commented):
I've tried command from the PR desc:
```
uv run sct.py sizing resolve --vcpu 1024 --memory 8192 --cloud aws
logged in as [...]
Usage: hydra sizing resolve [OPTIONS]
Try 'hydra sizing resolve --help' for help.

Error: No such option: --cloud Did you mean --vcpu?
```
but overall looks good

fruch:
> I've tried command from the PR desc:
> 
> ```
> uv run sct.py sizing resolve --vcpu 1024 --memory 8192 --cloud aws
> logged in as [...]
> Usage: hydra sizing resolve [OPTIONS]
> Try 'hydra sizing resolve --help' for help.
> 
> Error: No such option: --cloud Did you mean --vcpu?
> ```
> 
> but overall looks good

where did you picked the --cloud for this one ? it's designed to show you options for all backend 
it might be a bit confusing since you can pass in region for more specific pricing...

but actually the preview command is the improvement one, and much more useful

fruch:
Re: [sizing_config.yaml comment](https://github.com/scylladb/scylla-cluster-tests/pull/14576#discussion_r3301557685) — Added a "Adding a New Instance Family" section to `docs/cross-cloud-sizing.md` with step-by-step instructions (edit sizing_config.yaml → regenerate catalog → verify → commit). Also added a link from the `sizing_config.yaml` header comment.

data/instance_catalog/sizing_config.yaml
fruch: added this section to docs

sct_sizing.py
```diff
@@ -0,0 +1,589 @@
...
+    "--role",
+    type=click.Choice(["db", "loader", "monitor"]),
+    default=None,
+    help="Filter by role (shows preferred families)",
+)
+@click.option("--family", type=str, default=None, help="Filter by instance family")
+@click.pass_context
+def sizing_catalog(ctx, cloud, role, family):
+    from sdcm.utils.cloud_catalog.instance_catalog import InstanceCatalog  # noqa: PLC0415
+
```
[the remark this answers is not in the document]
fruch: Fixed. Hoisted `os`, `yaml`, and `dataclasses.replace` to module scope. The `sdcm.*` imports remain deferred inside click commands intentionally — they pull in heavy cloud SDKs and we want fast CLI startup for all commands.

sct_sizing.py
```diff
@@ -0,0 +1,589 @@
...
+                role_sources[cloud_name] = "env-var"
+            elif param_name in user_config and user_config[param_name]:
+                role_values[cloud_name] = user_config[param_name]
+                role_sources[cloud_name] = "test-config"
+            elif agnostic_param in user_config and isinstance(user_config.get(agnostic_param), dict):
+                role_values[cloud_name] = user_config[agnostic_param]
+                role_sources[cloud_name] = "test-config"
+            elif isinstance(agnostic_value, dict) and param_name != agnostic_param:
+                role_values[cloud_name] = agnostic_value
+                role_sources[cloud_name] = "defaults"
```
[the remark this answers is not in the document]
fruch: Fixed. The Source column now shows the actual config key the value came from (e.g. `sizing_db (defaults)` instead of always showing the cloud-specific `instance_type_db`). Changed `role_sources` to store `(key_name, origin)` tuples.

sdcm/sct_config.py
```diff
@@ -3239,6 +3256,130 @@ def no_op(x):
...
+                            self.log.warning("Agnostic fallback for %s found no match: %s", param_name, exc)
+                            if env is not None:
+                                env.pop(param_name, None)
+                            continue
+                        raise ValueError(f"Cannot resolve {param_name}: {exc}") from exc
+                    except FileNotFoundError as exc:
+                        self.log.warning("Instance catalog not found for %s: %s", param_name, exc)
+                        if env is not None:
+                            env.pop(param_name, None)
+                        continue
```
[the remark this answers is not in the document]
fruch: The code already re-raises for non-fallback cases (line 3364: `raise ValueError`). The `continue` only triggers for `is_fallback=True` — meaning the value came from an agnostic key like `sizing_db` being applied as a fallback to a cloud-specific param. If the cloud-specific param has no match, that's expected (the agnostic constraint may not be satisfiable on all clouds). The `FileNotFoundError` branch (catalog file missing) is also acceptable to swallow — it means catalogs aren't installed yet and sizing should degrade gracefully to manual config.

sdcm/utils/cloud_catalog/catalog_generator.py
```diff
@@ -0,0 +1,1286 @@
...
+
+"""Catalog generator for cloud instance types.
+
+Provides functions to generate InstanceTypeInfo catalogs from live cloud APIs
+for AWS, GCE, Azure, and OCI. Intended as a developer tool to refresh the
+static YAML catalog files in data/instance_catalog/.
+
+Cloud SDK imports are deferred inside each function (try/except at function
+level) because these SDKs are optional dependencies — not all clouds are
+needed in every environment.
```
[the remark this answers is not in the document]
fruch: Fixed. Hoisted `os`, `requests`, and removed duplicate `yaml` import to module scope. The cloud SDK imports (`boto3`, `google.cloud`, `azure.*`) remain intentionally deferred — as documented in the module docstring, these are optional dependencies and each function only imports the SDK for its target cloud.

sdcm/utils/cloud_catalog/catalog_generator.py
```diff
@@ -0,0 +1,1286 @@
...
+
+    return result
+
+
+def generate_gce_catalog(  # noqa: PLR0914
+    families: list[str],
+    project: str = "",
+    zone: str = "us-east1-b",
+    regions: list[str] | None = None,
+) -> list[InstanceTypeInfo]:
```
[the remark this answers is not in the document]
fruch: By design. The `regions` parameter controls which GCE API zones are queried for machine type discovery (availability). The pricing dict (scraped from public pricing pages) is intentionally stored in full — the cached YAML file should contain all-region pricing so that `get_price(region)` works for any region at runtime without re-scraping. This matches AWS/Azure behavior where the catalog YAML contains all-region data.

sdcm/utils/cloud_catalog/catalog_generator.py
```diff
@@ -0,0 +1,1286 @@
...
+            if "Spot" in sku_name or "Low Priority" in sku_name:
+                continue
+            price = item.get("retailPrice")
+            if sku and price is not None and sku not in prices:
+                prices[sku] = float(price)
+
+        url = data.get("NextPageLink")
+        params = {}  # NextPageLink already contains query params
+
+    return prices
```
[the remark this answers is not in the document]
fruch: This is a developer-only catalog generation CLI tool (run manually via `sct.py sizing update-catalog`), not production runtime code. Using `requests.get` with a 30s timeout is acceptable here — the retry-enabled session from `rest_client.py` is designed for runtime Scylla API calls and adds unnecessary complexity for a one-shot data scraping script. If the pricing API is temporarily down, the developer simply retries the command.

sdcm/utils/cloud_catalog/pricing.py
```diff
@@ -0,0 +1,265 @@
...
+def _get_catalog():
+    global _CATALOG  # noqa: PLW0603
+    if _CATALOG is None:
+        from sdcm.utils.cloud_catalog.instance_catalog import InstanceCatalog  # noqa: PLC0415 — avoid circular import
+
+        catalog_dir = Path(__file__).parent.parent.parent / "data" / "instance_catalog"
+        if catalog_dir.exists():
+            _CATALOG = InstanceCatalog.from_directory(catalog_dir)
+        else:
+            _CATALOG = InstanceCatalog()
```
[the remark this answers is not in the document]
fruch: Fixed. The path was resolving to `sdcm/data/instance_catalog` instead of the repo root `data/instance_catalog`. Added one more `.parent` level: `Path(__file__).parent.parent.parent.parent / "data" / "instance_catalog"`.

sdcm/utils/cloud_catalog/pricing.py
```diff
@@ -0,0 +1,265 @@
...
+        else:
+            return 0
+
+    @staticmethod
+    @lru_cache(maxsize=None)
+    def _get_sku_prices(instance_type: str, region):
+        resp = requests.get(
+            f"https://prices.azure.com/api/retail/prices?$filter=serviceName eq 'Virtual Machines' "
+            f"and armSkuName eq '{instance_type}' and armRegionName eq '{region}' and priceType eq 'consumption'"
+        )
```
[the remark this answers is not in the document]
fruch: This is pre-existing code in the `AzurePricing` class (part of the cloud_monitor pricing integration that predates this PR). The retry-session suggestion is valid but out of scope for this PR — it would affect the existing monitoring stack behavior.

unit_tests/integration/test_catalog_generator.py
```diff
@@ -0,0 +1,191 @@
...
+)
+from sdcm.utils.cloud_catalog.instance_catalog import InstanceTypeInfo
+
+
+@pytest.mark.integration
+class TestAWSCatalogGenerator:
+    @pytest.fixture(autouse=True)
+    def _check_aws_credentials(self):
+        try:
+            import boto3  # noqa: PLC0415
```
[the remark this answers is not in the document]
fruch: These imports are inside `try/except ImportError` blocks specifically to detect SDK availability and skip tests when the SDK is missing. This is the standard SCT integration test pattern (see other integration tests in the repo). Using `pytest.importorskip` would be functionally equivalent but would change the skip message style — keeping the current pattern for consistency with the rest of the test suite.

unit_tests/unit/test_catalog_generator.py
```diff
@@ -0,0 +1,130 @@
...
+
+
+def test_generate_oci_catalog_family_filter():
+    result = generate_oci_catalog(["BM.DenseIO"])
+    assert all(r.family == "BM.DenseIO" for r in result)
+
+
+def test_generate_oci_catalog_unknown_family_returns_empty():
+    result = generate_oci_catalog(["NonExistentFamily99"])
+    assert result == []
```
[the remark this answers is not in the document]
fruch: Fixed. Added `_mock_oci_pricing` autouse fixture that monkeypatches `_oci_get_pricing` with static rate data, preventing network calls during unit tests.

unit_tests/unit/test_instance_matcher.py
```diff
@@ -0,0 +1,493 @@
...
+
+
+# ---------------------------------------------------------------------------
+# parse_constraints — unknown key warning
+# ---------------------------------------------------------------------------
+
+
+def test_parse_constraints_unknown_key_logs_warning(caplog):
+    with caplog.at_level(logging.WARNING, logger="sdcm.utils.instance_matcher"):
+        result = parse_constraints({"vcpu": 8, "gpu": 2, "network": "10gbe"})
```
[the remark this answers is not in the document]
fruch: Fixed. The logger name was stale after the module restructure. Updated `caplog.at_level` to use `sdcm.utils.cloud_catalog.instance_matcher` instead of `sdcm.utils.instance_matcher`.

unit_tests/unit/test_instance_matcher.py
```diff
@@ -0,0 +1,493 @@
...
+
+# ---------------------------------------------------------------------------
+# select_instance — fixture
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def test_catalog():
+    from sdcm.utils.cloud_catalog.instance_catalog import InstanceCatalog, InstanceTypeInfo  # noqa: PLC0415
+
```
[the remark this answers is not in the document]
fruch: Fixed. Moved `InstanceCatalog` and `InstanceTypeInfo` imports to module scope.

unit_tests/unit/test_instance_matcher.py
```diff
@@ -0,0 +1,493 @@
...
+    result = select_instance(test_catalog, "monitor", "oci", {"vcpu": 4, "memory": "<=128"})
+    assert result.instance_type == "VM.Standard.E4.Flex:2:16"
+    assert result.memory_gb == 16.0
+
+
+def test_flex_memory_range(test_catalog):
+    result = select_instance(test_catalog, "monitor", "oci", {"vcpu": 4, "memory": "16-64"})
+    assert result.instance_type == "VM.Standard.E4.Flex:2:16"
+    assert result.memory_gb == 16.0
+
```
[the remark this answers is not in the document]
fruch: Intentionally kept separate. Each test exercises a different constraint operator (no constraint, exact, `>=`, `<=`, range) with different expected flex memory resolution behavior. Parametrizing would require complex conditional assertions that hurt readability more than the modest duplication.

unit_tests/unit/test_instance_sizing_config.py
```diff
@@ -0,0 +1,187 @@
...
+import logging
+from pathlib import Path
+from unittest.mock import patch
+
+import pytest
+
+from sdcm import sct_config
+from sdcm.utils.cloud_catalog.instance_catalog import InstanceCatalog, InstanceTypeInfo
+
+logging.basicConfig(level=logging.ERROR)
```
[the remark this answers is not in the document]
fruch: Fixed. Replaced `logging.basicConfig(level=logging.ERROR)` at module scope with an `autouse` fixture that uses `caplog.at_level(logging.ERROR, logger="sdcm")` — scoped per test, no global side effects.

test-cases/longevity/longevity-twcs-48h.yaml
```diff
@@ -13,7 +13,9 @@ stress_read_cmd: [
 n_db_nodes: 6
 n_loaders: 3
 
-instance_type_db: 'i8ge.xlarge'
+sizing_db:  # no OCI match (min 16 vCPUs in OCI catalog)
```
soyacz: if i8ge instance was used, shouldn't we restrict bigger disk?

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
```diff
@@ -64,7 +64,11 @@ n_loaders: '2 2'
 region_aware_loader: true
 simulated_racks: 0
 
-instance_type_db: 'i7i.xlarge'
+sizing_db:  # no Azure/OCI match (Azure L8s_v3 only 1920 GB, OCI min 16 vCPUs)
+  vcpu: 8
+  memory: '>=64'
+  arch: x86_64
+  local_disk_gb: '>=3000'
```
soyacz: i7i.xlarge is not having this size disks - looks like an error

skills/migrate-to-sizing/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+│   └── NO (just happened to be Intel) → still add arch: x86_64 to preserve behavior
+└── NO (i4i, i8g, i8ge, or mixed) → do NOT pin arch
+```
+
+### Decision Tree: When to Pin Disk Size
+
+```
+Does the original instance have notably large disk (>= 2500 GB per disk)?
+├── YES → Is there a smaller-disk instance with same vCPU/memory/arch?
+│   ├── YES → add local_disk_gb: '>=XXXX' to differentiate
```
soyacz: I think we should base on instance family, when i7ie/i8ge is used then we should add disk constraint

skills/migrate-to-sizing/SKILL.md
```diff
@@ -0,0 +1,204 @@
...
+
+# Migrate Test Configs to Constraint-Based Sizing
+
+Convert literal per-cloud instance type parameters to cloud-agnostic sizing constraints.
+
+## Essential Principles
+
+### Map Instance Specs to Constraints
+
+Every literal instance type has specific vCPU, memory, and disk characteristics. The sizing constraint must capture the **intent** (what resources the test needs) not the specific instance. Use the minimum constraints that would resolve to an equivalent or better instance across all clouds.
```
soyacz, line 24: by default we take intent from AWS instances sizes, but if test is designed for GCE it should take from GCE instances sizes.

soyacz (commented):
Several more comments

skills/migrate-to-sizing/SKILL.md
fruch: yes we should base on what ever instance type are available for the test case

skills/migrate-to-sizing/SKILL.md
fruch: yes we should put disk constraint, if it's important for the test case (if we had dense families we can assume it was deliberately )

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
fruch: I think this case was using bigger disk instances, but recently was changed. so we need to align/adapt this one

test-cases/longevity/longevity-twcs-48h.yaml
fruch: yes  we need to fix this one

fruch: You're right. i8ge is in the preferred_families list but wasn't included in the `families` list used to generate the AWS catalog. I'll add i8ge to the AWS families in sizing_config.yaml and regenerate aws.yaml so the constraint can resolve. Adding a disk constraint (`local_disk_gb: >=1200`) as well.

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
fruch: Fixed. The constraint was based on the old i7ie.2xlarge (3750GB) but the test was recently downgraded to i7i.xlarge (937GB). Updated to `vcpu: 4, memory: >=32, arch: x86_64, local_disk_gb: >=900` which correctly resolves to i7i.xlarge on AWS.

skills/migrate-to-sizing/SKILL.md
fruch: Agreed — updated the migration logic. When the original config uses a dense storage family (i8ge, i7ie, i4i, BM.DenseIO, etc.), the migration should include a `local_disk_gb` constraint derived from that instance's disk capacity.

skills/migrate-to-sizing/SKILL.md
fruch: Agreed — the migration should derive intent from whichever instance types are present in the config, not assume AWS. If only `gce_instance_type_db` is set (and no AWS equivalent), the constraints should be based on the GCE instance specs. Will update the SKILL.md guidance.

soyacz: actually, I think we could make it script based and not LLM based: we could read configs and map instances to proper settings

fruch: for now, we won't be scripting it.

defaults/test_default.yaml
```diff
@@ -125,9 +125,17 @@ authenticator_password: ''
...
 run_gemini_in_rolling_upgrade: false
 
+# cloud-agnostic instance sizing constraints (resolved per-cloud by _resolve_instance_sizes)
+sizing_loader:
+  vcpu: 4
+  memory: '>=8'
+sizing_monitor:
+  vcpu: 2
+  memory: '>=8'
+
```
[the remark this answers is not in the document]
fruch: Fixed — added `sizing_db: null` and `sizing_db_oracle: null` defaults.

docs/cross-cloud-sizing.md
```diff
@@ -1,152 +1,333 @@
...
-| memory | memory_gb | int, float, or string | no | any | RAM in GB |
-| disk | local_disk_gb | int, float, or string | no | auto for db roles | Local disk in GB |
-| arch | - | string | no | per-cloud default | x86_64 or arm64 |
+```
+Role: db  Constraints: {vcpu: 8, memory: >=32}
+  AWS:   i8g.2xlarge       (8 vCPU, 64 GB, 1875 GB NVMe, arm64)    $0.62/hr
+  GCE:   z3-highmem-8      (8 vCPU, 64 GB, 750 GB SSD, x86_64)
+  Azure: Standard_L8s_v4   (8 vCPU, 64 GB, 1788 GB NVMe, x86_64)
+  OCI:   DenseIO.E5.Flex   (8 vCPU, 96 GB, NVMe, x86_64)          $0.82/hr
+```
```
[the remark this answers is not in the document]
fruch: Fixed — added `text` language identifier to both bare code fences.

sct_sizing.py
```diff
@@ -0,0 +1,586 @@
...
+    for path_str in config_files:
+        path = Path(path_str)
+        if path.suffix in {".jenkinsfile", ".groovy"}:
+            pipeline_config = parse_jenkinsfile(path)
+            if pipeline_config and pipeline_config.test_config:
+                resolved.extend(pipeline_config.test_config)
+            else:
+                click.echo(f"Warning: no test_config found in {path}", err=True)
+        else:
+            resolved.append(path_str)
```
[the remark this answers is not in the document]
fruch: Fixed — now checks `path.name == "Jenkinsfile"` in addition to suffix matching.

skills/migrate-to-sizing/SKILL.md
```diff
@@ -0,0 +1,214 @@
...
+When the original config uses a dense storage family (i8ge, i7ie, i4i, i3en, BM.DenseIO, VM.DenseIO, L-series), always include a `local_disk_gb` constraint. These families were chosen deliberately for their storage capacity — without a disk constraint, the resolver may pick a smaller-disk instance from the same family.
+
+### DB Role Always Gets Local Disk Implicitly
+
+The sizing system adds `local_disk_count > 0` automatically for db/db_oracle/zero_token roles. Never specify `local_disk_count` in sizing_db — it's redundant. Only specify `local_disk_gb` when you need to differentiate between disk tiers (e.g., i7ie's 5000 GB vs i7i's 1875 GB).
+
+### Pin Architecture Only When Required
+
+Only add `arch: x86_64` when the original config used Intel-only families (i7i, i7ie, i3en) AND the test specifically needs x86. If the original used i4i/i8g/i8ge (which have both arch variants or ARM), don't pin arch — let the resolver pick the best match per cloud.
+
```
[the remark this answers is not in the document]
fruch: Fixed — removed the contradictory branching. Rule is now: Intel-only families → always pin `arch: x86_64` to preserve original behavior. Non-Intel → don't pin.

skills/migrate-to-sizing/SKILL.md
```diff
@@ -0,0 +1,214 @@
...
+### Decision Tree: When to Pin Architecture
+
+```
+Was the original instance Intel-only (i7i, i7ie, i3en)?
+├── YES → Does the test NEED Intel specifically?
+│   ├── YES (e.g., testing Intel-specific features) → add arch: x86_64
+│   └── NO (just happened to be Intel) → still add arch: x86_64 to preserve behavior
+└── NO (i4i, i8g, i8ge, or mixed) → do NOT pin arch
+```
```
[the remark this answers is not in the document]
fruch: Fixed — added `text` language to both decision tree fences.

test-cases/longevity/longevity-parallel-schema-changes-12h-cql-stress.yaml
```diff
@@ -16,7 +16,9 @@ n_db_nodes: 6
 n_loaders: 2
 simulated_racks: 3
 
-instance_type_db: 'i4i.2xlarge'
+sizing_db:  # no OCI match (min 16 vCPUs in OCI catalog)
+  vcpu: 8
+  memory: '>=64'
```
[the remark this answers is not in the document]
fruch: Fixed — clarified to "DenseIO min 16 vCPUs". The catalog has smaller OCI instances but without local disk, which DB role requires automatically.

test-cases/features/elasticity/elasticity-90-percent-with-nemesis.yaml
```diff
@@ -18,9 +18,12 @@ n_db_nodes: 6
...
 
-instance_type_loader: 'c6i.2xlarge'
-instance_type_monitor: 't3.large'
-instance_type_db: 'i8g.large'
+sizing_db:  # no GCE/OCI match (min 16 vCPUs in OCI catalog, min 8 in GCE z3)
+  vcpu: 2
+  memory: '>=16'
+sizing_loader:
+  vcpu: 8
+  memory: '>=8'
```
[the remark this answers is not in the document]
fruch: Fixed — clarified to "DenseIO min 16 vCPUs on OCI, z3 min 8 vCPUs on GCE". The OCI catalog does have smaller instances but they lack local disk (required for DB role).

test-cases/longevity/longevity-parallel-schema-changes-12h-cql-stress.yaml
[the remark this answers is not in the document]
fruch: not relevent we care about specific family

test-cases/features/elasticity/elasticity-90-percent-with-nemesis.yaml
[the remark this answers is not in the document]
fruch: the comment is fine

skills/migrate-to-sizing/SKILL.md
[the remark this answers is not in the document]
fruch: not relevent

soyacz (commented):
Did you test it? 
It affects the defaults, and some tests don't specify instances - IIUC, this sizing will be used.

fruch:
> It affects the defaults, and some tests don't specify instances - IIUC, this sizing will be used.

tested it with provision tests across all backends, and validate the tier1 configuration that gonna use it. 

it does drop the defaults for DB (align all backend with how AWS didn't had default
tests that didn't specify it, I'm find with it to fail until it set a instance or sizing

data/instance_catalog/sizing_config.yaml
```diff
@@ -0,0 +1,190 @@
...
+    arch: x86_64
+
+# Sort order for candidate instances after filtering.
+# Instances are sorted by these fields in order; first match wins.
+# - preferred_family_rank: index in the preferred_families list (lower = better)
+# - vcpus: fewer vCPUs preferred (closest match to request)
+# - price_per_hour: cheaper preferred as tiebreaker
+sort_order:
+  - preferred_family_rank
+  - vcpus
```
roydahan, line 50: @fruch what will it choose in case of say someone asked for 10 cpus and you have:
i7i.4xlarge - 16cpus - cost is 1.51$
i7ie.3xlarge - 12cpus - cost is 1.55$

From the explanation above it seems like it will choose the more expensive one with less cpus.

fruch: The sort order is: **family_rank → vcpus (smallest) → price (cheapest)**.

Since i7i is ranked higher than i7ie in `preferred_families.db` (i7i=rank 2, i7ie=rank 3), **i7i wins regardless of vCPU count or price** — family preference takes priority.

You can verify with the CLI:

```
uv run sct.py sizing resolve --vcpu ">=10" --arch x86_64 --role db

Cloud     Resolved Instance          vCPUs  Memory(GB)  Disk(GB)  Arch      Price/hr
aws       i7i.4xlarge                   16      128.00      3750  x86_64     $1.5101
gce       z3-highmem-16-highlssd        16      128.00     12000  x86_64     $3.0116
azure     Standard_L16s_v3              16      128.00      3840  x86_64     $1.3920
oci       VM.DenseIO.E5.Flex:8          16       96.00      6400  x86_64     $0.8237
```

The design intentionally prefers the "better" family (newer generation, better perf/$ ratio) even if it means a slightly larger instance. vCPU and price only break ties within the same family rank.

fruch: @roydahan is this something you want to change now ?

fruch:
@scylladb/qa-maintainers ready for review — rebased on latest master, all review comments addressed, CI passing.

skills/migrate-to-sizing/SKILL.md
```diff
@@ -0,0 +1,212 @@
...
+
+Add `arch: x86_64` when the original config used Intel-only families (i7i, i7ie, i3en) — this preserves the original instance selection behavior. If the original used i4i/i8g/i8ge (which have ARM variants or are ARM-only), don't pin arch — let the resolver pick the best match per cloud.
+
+### Loader and Monitor Use Defaults Unless Overridden
+
+`sizing_loader` and `sizing_monitor` have defaults in `defaults/test_default.yaml` (loader: vcpu 4, memory >=8; monitor: vcpu 2, memory >=8). Only add sizing_loader/sizing_monitor to a test config if it needs MORE than the default.
+
+### Keep Special Params As-Is
+
+Some params like `nemesis_grow_shrink_instance_type` are AWS-specific operational parameters, not sizing constraints. Leave them as literal instance types.
```
soyacz, line 44: I disagree with this statement, 
things like `instance_type_db_target`, `nemesis_grow_shrink_instance_type` - should be also covered.

fruch: It can be added as a followup, if those are needed to be treated the same.

even if this PR gonna have support for them, the cases using them wouldn't be touch in this PR

fruch: also those are in 5 configuration yamls only... I don't see any urgency in attending to those in this PR

soyacz (dismissed):
LGTM

docs/cross-cloud-sizing.md
```diff
@@ -1,152 +1,333 @@
...
+
+## Environment Variables
+
+Use SCT's dot-notation to override constraints from the command line:
+
+```bash
+export SCT_INSTANCE_TYPE_DB.VCPU=16
+export SCT_INSTANCE_TYPE_DB.MEMORY=64
+export SCT_INSTANCE_TYPE_DB.ARCH=arm64
+```
```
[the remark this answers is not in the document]
fruch: Fixed — replaced invalid dotted env var syntax with the correct SCT pattern: `export SCT_SIZING_DB='{"vcpu": 16, "memory": 64, "arch": "arm64"}'`

sdcm/utils/cloud_catalog/pricing.py
```diff
@@ -279,11 +235,10 @@ def get_instance_price(self, region, instance_type, state, lifecycle):
                     ][0]
                 else:
                     return [price["retailPrice"] for price in prices if "Spot" in price["meterName"]][0]
-            except KeyError:
+            except KeyError, IndexError:
```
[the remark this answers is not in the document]
fruch: Fixed — `except KeyError, IndexError:` → `except (KeyError, IndexError):`

sdcm/sct_config.py
```diff
@@ -3308,6 +3325,130 @@ def no_op(x):
...
+
+        AGNOSTIC_FALLBACK = {
+            "db": "sizing_db",
+            "db_oracle": "sizing_db_oracle",
+            "zero_token": "sizing_db",
+            "loader": "sizing_loader",
+            "monitor": "sizing_monitor",
+        }
+
+        db_type = self.get("db_type") or ""
```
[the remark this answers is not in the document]
fruch: Fixed — now reads `db_type` from env first, falling back to `self`: `db_type = (env.get("db_type") if env else None) or self.get("db_type") or ""`

sct_sizing.py
```diff
@@ -0,0 +1,586 @@
...
+@click.option(
+    "--role",
+    type=click.Choice(["db", "loader", "monitor"]),
+    default=None,
+    help="Filter by role (shows preferred families)",
+)
+@click.option("--family", type=str, default=None, help="Filter by instance family")
+@click.pass_context
+def sizing_catalog(ctx, cloud, role, family):
+    from sdcm.utils.cloud_catalog.instance_catalog import InstanceCatalog  # noqa: PLC0415
```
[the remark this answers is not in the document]
fruch: Fixed — added `# lazy-import: ...` comments explaining why each inline import exists (heavy cloud SDK imports at CLI startup time).

unit_tests/integration/test_catalog_generator.py
```diff
@@ -0,0 +1,191 @@
...
+    generate_aws_catalog,
+    generate_azure_catalog,
+    generate_gce_catalog,
+    generate_oci_catalog,
+)
+from sdcm.utils.cloud_catalog.instance_catalog import InstanceTypeInfo
+
+
+@pytest.mark.integration
+class TestAWSCatalogGenerator:
```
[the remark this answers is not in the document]
fruch: Fixed — flattened all `class Test*` groupings into module-level `test_*` functions with `@pytest.mark.usefixtures` for shared setup.

sdcm/utils/cloud_catalog/instance_matcher.py
```diff
@@ -0,0 +1,581 @@
...
+        raw_str = str(raw)
+    else:
+        raw_str = str(raw).strip()
+
+    # Range: "16-64" or "16gb-64gb"
+    m = _RANGE_RE.match(raw_str)
+    if m:
+        lo_val = float(m.group(1)) * _UNIT_MULTIPLIER.get((m.group(2) or "gb").lower(), 1.0)
+        hi_val = float(m.group(3)) * _UNIT_MULTIPLIER.get((m.group(4) or "gb").lower(), 1.0)
+        return Constraint(field, "range", (lo_val, hi_val))
```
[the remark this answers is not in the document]
fruch: Fixed — added `if lo_val > hi_val: raise ValueError(...)` after parsing the range in `_parse_memory_or_disk`.

sdcm/utils/cloud_catalog/instance_matcher.py
```diff
@@ -0,0 +1,581 @@
...
+        ValueError: If the raw value cannot be parsed.
+    """
+    if isinstance(raw, int):
+        return Constraint("vcpus", "eq", raw)
+
+    if isinstance(raw, str):
+        raw = raw.strip()
+        m = re.match(r"^(\d+)\s*-\s*(\d+)$", raw)
+        if m:
+            return Constraint("vcpus", "range", (int(m.group(1)), int(m.group(2))))
```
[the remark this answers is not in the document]
fruch: Fixed — same validation added in `_parse_vcpu_constraint`: `if lo > hi: raise ValueError(...)`.

docs/cross-cloud-sizing.md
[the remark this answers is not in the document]
soyacz: Are you sure it was invalid?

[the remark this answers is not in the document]
fruch: yes

```bash
❯ export SCT_INSTANCE_TYPE_DB.ARCH=arm64                                                                                                                                                               
bash: export: `SCT_INSTANCE_TYPE_DB.ARCH=arm64': not a valid identifier 
```

it need to be used like:
```bash
env 'SCT_INSTANCE_TYPE_DB.ARCH=arm64'
```

which make this notation not that nice, if you pass it via pipeline overrides it would work (machinery is a bit different)



```
Bash flat-out rejects dots in variable names. There's no workaround with export:
$ export SCT_INSTANCE_TYPE_DB.ARCH=arm64
bash: export: `SCT_INSTANCE_TYPE_DB.ARCH=arm64': not a valid identifier
Your options:
# 1. Pass to a single command via env
env 'SCT_INSTANCE_TYPE_DB.ARCH=arm64' ./run_test.sh ...

# 2. Spawn a subshell with it set
env 'SCT_INSTANCE_TYPE_DB.ARCH=arm64' bash

# 3. If the tool supports it, use a config file or CLI flag instead
There's no way to export a dotted name that persists in the current bash session. env is the only path.
```

the suggest it to replace the dot with __ (double underscore), this is what pydantic does for nesting 

but that's not for this PR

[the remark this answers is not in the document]
fruch: https://scylladb.atlassian.net/browse/SCT-466 for further handling this

PR #14596, opened 2026-05-11
perf(cli): speed up tab completion by short-circuiting heavy imports

fruch:
## Summary

- Detects shell completion mode (`_HYDRA_COMPLETE` env var) at the top of `sct.py` before any heavy imports
- Provides a lightweight CLI stub with just command names for instant tab completion
- Reduces completion time from **several seconds** to **~0.18s**

## Problem

Every tab press on `./sct.py` or `hydra` triggers a full Python process that imports the entire SCT dependency tree (AWS SDK, kubernetes, pytest, azure, google-cloud, paramiko, fabric, etc.) before Click can short-circuit to completion. This makes tab completion painfully slow (3-8 seconds depending on environment).

## Solution

Add an early-exit fast path at the top of `sct.py` that:
1. Checks for `_HYDRA_COMPLETE` or `_SCT_COMPLETE` environment variables (set by Click's bash completion)
2. Imports only `click`, `os`, `sys`, `pathlib` (~0.15s)
3. Builds a minimal CLI group with all command names registered as stubs
4. Lets Click generate completions and exits immediately

The command names are maintained as a static list in the fast-path block. When new commands are added to `sct.py`, they should also be added to this list.

## Trade-offs

- **Maintenance**: New commands must be added to the `_COMMANDS` list in the fast-path block. If forgotten, tab completion won't suggest the new command (but it will still work when typed manually).
- **No option completion**: The fast path only provides command-name completion, not per-command option completion. Full option completion would require importing the actual command implementations.

## Manual Testing Required

- [ ] Test tab completion inside hydra container: `hydra <TAB>`
- [ ] Test tab completion for subcommands: `hydra investigate <TAB>`
- [ ] Verify normal command execution still works: `hydra nemesis-list`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -125,6 +125,12 @@ force-exclude = true
 line-length = 120
 respect-gitignore = true
 
+[tool.ruff.lint.per-file-ignores]
+# sct.py and sct_cmd_helpers.py use function-level imports intentionally
+# to speed up tab completion (~0.3s vs ~8s with top-level imports)
+"sct.py" = ["PLC0415"]
+"sdcm/utils/sct_cmd_helpers.py" = ["PLC0415"]
+
 [tool.ruff.lint.pylint]
 max-args = 12
 max-statements = 100
--- a/sct.py
+++ b/sct.py
@@ -33,121 +33,63 @@ import xml.etree.ElementTree as ET
 from concurrent.futures import ProcessPoolExecutor
 from pathlib import Path
 from functools import partial, reduce
-from typing import List, Literal
+from typing import List, Literal, TYPE_CHECKING
 from uuid import UUID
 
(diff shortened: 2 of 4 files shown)


pehala:
I dont think this is a way to go. It is a temporary workaround that will add work and potential desync everytime we add new command. If we want to fix the perf for the shell commands we should think about a longterm solution as a main thing and then use something like this for the transition period only

fruch:
> I dont think this is a way to go. It is a temporary workaround that will add work and potential desync everytime we add new command. If we want to fix the perf for the shell commands we should think about a longterm solution as a main thing and then use something like this for the transition period only

that was initial quick option, now the PR has a bit more helpful change, that should have a drift issue.

PR #14624, opened 2026-05-13
fix(nemesis): remove table-existence check from _prepare_test_table

jsmolar:
No need to cross-check if a keyspace/table exists — just let cassandra-stress
run and create the schema or overwrite the data. Worse case it overwrites
existing rows, which is fine.

This also fixes a ~9h stall in longevity-5000-tables where the old check
called nodetool cfstats once per table: 5000 tables × 10 iterations × ~0.5s
each adds up fast.

Failed run: https://argus.scylladb.com/tests/scylla-cluster-tests/e5286788-c029-465d-8841-31c9e6c2bf23

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -898,7 +898,7 @@ class NemesisRunner:
         try:
             # Run connect a new session after authenticator switch, and run a short workload
             self.actions_log.info("Run a short workload after Authenticator switch")
-            self._prepare_test_table(ks="keyspace_for_authenticator_switch", table="standard1")
+            self._prepare_test_table(ks="keyspace_for_authenticator_switch")
         finally:
             # Wait 2 mins to let the workloads run with new Authenticator,
             # then switch Authenticator back to original
@@ -1668,7 +1668,7 @@ class NemesisRunner:
         # prepare test tables and fill test data
         self.actions_log.info("Preparing test tables")
         for i in range(10):
-            self._prepare_test_table(ks=f"drop_table_during_repair_ks_{i}", table="standard1")
+            self._prepare_test_table(ks=f"drop_table_during_repair_ks_{i}")
             self.cluster.wait_for_schema_agreement()
 
         self.log.debug("Start repair target_node in background")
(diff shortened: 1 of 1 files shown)


sdcm/nemesis/__init__.py
```diff
@@ -2189,7 +2189,7 @@ def disrupt_nodetool_cleanup(self):
         self.nodetool_cleanup_on_all_nodes_parallel()
 
     def _prepare_test_table(self, ks="keyspace1", table=None):
-        ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=self.target_node)
+        ks_cfs = self.cluster.get_non_system_ks_cf_list(db_node=self.target_node, filter_empty_tables=False)
         table_exist = f"{ks}.{table}" in ks_cfs if table else True
 
         test_keyspaces = [cql_unquote_if_needed(ks) for ks in self.cluster.get_test_keyspaces()]
```
fruch: I'm not sure anymore why we need all of those checks for... why not just let c-s write the data, or overwrite the data ?, not sure we need so much fuss...

jsmolar: I think it's there just to save some time if the table exists, which is 99% of the time.

jsmolar: So the question is, what is better? Rewrite almost every time? Or these checks?

cezarmoise (dismissed):
LGTM

sdcm/nemesis/__init__.py
vponomaryov: @jsmolar 
The `disrupt_no_corrupt_repair` nemesis creates and deletes tables after itself and exactly this nemesis always may skip checks. And, in this case, we may have some of the tables only if we did run this nemesis before and it failed in the middle. It is not 99%.

So, simple solution:
- Add a parameter for the `_prepare_test_table` function with possibility to skip checks which will be `True` for the ^ nemesis.

More complex solution, but technically better:
- Do not use serial approach, use latte to do everything in a single stress command like in the 5k tables test scenario where you observed the problem. It can reuse the existing rune script, just need to update it with something like "custom keyspace/table prefix" - to not overlap with the main stress in 5k scenario.

fruch: checked all of the usages, we don't really care if it's created or not, just run c-s (or latte) let it create the schema, and fill the data.

worse case it would overwrite the 40K rows data, that's fine. no real need to cross check if table or ks exists, it's a waste of our testing time.

pehala (approved):

cezarmoise (approved):
LGTM

Please update PR title and description before this is merged.

jsmolar:
> LGTM
> 
> Please update PR title and description before this is merged.

Done

PR #14145, opened 2026-03-19
Add trigger matrix system for centralized Jenkins job scheduling

fruch:
## Overview
Introduces a new trigger matrix system that consolidates Jenkins job scheduling into YAML-based configuration files. This replaces scattered Groovy trigger definitions with a single source of truth, making it easier to manage, audit, and modify job triggers across multiple backends and versions.

Fixes: QAINFRA-35

### Core Implementation
- **`sdcm/utils/trigger_matrix.py`**: Trigger matrix engine with wait mode support
  - `load_matrix_config()`: Loads and validates YAML matrix files
  - `determine_job_folder()`: Derives Jenkins job folder from version strings
  - `filter_jobs()`: Dual-version filtering — original version for include/exclude, resolved for pre_release
  - `build_job_parameters()`: Merges defaults, job params, and CLI overrides
  - `trigger_jenkins_job()`: Fire-and-forget trigger with retry logic
  - `trigger_jenkins_job_with_queue()`: Gating trigger that polls build status, collects artifacts
  - `trigger_matrix()`: Orchestration with `wait=true` / `fail_on_error` semantics

### Configuration Files
- **`configurations/triggers/tier1.yaml`**: Tier1 jobs (10 jobs across AWS/Azure/GCE)
- **`configurations/triggers/sanity.yaml`**: Sanity jobs
- **`configurations/triggers/perf-regression.yaml`**: Performance regression (~30 jobs, 6 cron schedules)
- **`configurations/triggers/pgo-offline-installer.yaml`**: PGO perf-simple-query (fire-and-forget, replaces Groovy)
- **`configurations/triggers/scylla-doctor-gating.yaml`**: Scylla-doctor analysis gating (wait=true, collects JSON results)

### Pipeline Locations (replacing previous jobs in-place)
| Previous | Replaced by | Notes |
|----------|-------------|-------|
| `master-triggers/sct_triggers/tier1-custom-time-trigger.xml` | `master-triggers/sct_triggers/tier1-custom-time-trigger.jenkinsfile` | Same base name, XML→Pipeline |
| `master-triggers/sct_triggers/perf-regression-trigger.jenkinsfile` | Same file, content replaced | In-place replacement |
| `master-triggers/sct_triggers/performance/perf-simple-query-offline-installer-trigger-arm.jenkinsfile` | Same file, content replaced | In-place replacement |
[...]

PR #14629, opened 2026-05-13
refactor(longevity): move asymmetric cluster test to GCE

jsmolar:
- move backend from AWS to GCE, keep AWS support (n2-highmem-8 / i7i.2xlarge)
- keep 6 db nodes at 8 vCPU, so random shards stay 4-7
- increase dataset to 300M rows, ~60% disk usage
- replace separate write + read stress with one mixed cmd at 45k ops/s
- remove `150GB` from config and jenkinsfile name
- move job from AWS to GCE trigger chain

Fixes https://scylladb.atlassian.net/browse/SCYLLADB-1438

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/5fd01818-1c4c-4d47-afa2-68a1cd71f9d7

### Reminders


test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -1,25 +1,35 @@
...
-n_loaders: 1
+
+n_db_nodes: 12
+n_loaders: 2
 simulated_racks: 3
+round_robin: true
 
-instance_type_db: 'i7i.2xlarge'
-gce_instance_type_db: 'n2-highmem-16'
+gce_instance_type_db: 'n2-highmem-4'
```
fruch: in the sizing PR https://github.com/scylladb/scylla-cluster-tests/pull/14576:

```yaml
sizing_db:  # no OCI match (min 16 vCPUs in OCI catalog)                                                                                                                                                     
  vcpu: 4                                                                                                                                                                                                    
  memory: '>=32'                                                                                                                                                                                             
  arch: x86_64
                                                                                                                                                                                                             
sizing_loader:                                                                                                                                                                                               
  vcpu: 2    
```

```bash

🟢 ❯ uv run ./sct.py sizing preview jenkins-pipelines/oss/tier1/longevity-150gb-asymmetric-cluster-12h.jenkinsfile                                                                                           
logged in as arn:aws:iam::797456418907:user/fruch                                                                                                                                                            
Config files: test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey-cql-stress.yaml, configurations/db-nodes-shards-random.yaml, configurations/auth_cassandra.yaml                            
Test duration: 750 min (12.5 hr)                                                                                                                                                                             
                                                                                                                                                                                                             
Role: db (× 12 nodes)                                                                                                                                                                                        
                                                                                                                                                                                                             
  Cloud     Resolved Instance          vCPUs  Memory(GB)  Disk(GB)  Arch      Price/hr   Est. Cost  Source                                                                                                   
  ────────  ─────────────────────────  ─────  ──────────  ────────  ────────  ────────  ──────────  ────────────────────                                                                                     
  aws       i7i.xlarge                     4       32.00       937  x86_64     $0.3775      $56.62  instance_type_db (defaults) (resolved)                                                                   
  gce       n2-highmem-4                   4       32.00  1500 (4d)  x86_64     $0.2607      $39.10  gce_instance_type_db (defaults) (resolved)                                                              
  azure     Standard_L4s_v4                4       32.00       960  x86_64     $0.3480      $52.20  azure_instance_type_db (defaults) (resolved)                                                             
  oci       No match                                                                                oci_instance_type_db (defaults)                                                                          
                                                                                                                                                                                                             
Role: loader (× 2 nodes)                                                                                                                                                                                     
                                                                                                                                                                                                             
  Cloud     Resolved Instance          vCPUs  Memory(GB)  Disk(GB)  Arch      Price/hr   Est. Cost  Source                                                                                                   
  ────────  ─────────────────────────  ─────  ──────────  ────────  ────────  ────────  ──────────  ────────────────────                                                                                     
  aws       c6i.large                      2        4.00         0  x86_64     $0.0850       $2.12  instance_type_loader (defaults) (resolved)                                                               
  gce       e2-micro                       2        1.00         0  x86_64     $0.0460       $1.15  gce_instance_type_loader (defaults) (resolved)                                                           
  azure     Standard_F2s_v2                2        4.00         0  x86_64     $0.0846       $2.11  azure_instance_type_loader (defaults) (resolved)                                                         
  oci       VM.Standard.E4.Flex:1:8        2        8.00         0  x86_64     $0.0490       $1.23  oci_instance_type_loader (defaults) (resolved)                                                           
                                                                                                                                                                                                             
Role: monitor (× 1 nodes)                                                                                                                                                                                    
                                                                                                                                                                                                             
  Cloud     Resolved Instance          vCPUs  Memory(GB)  Disk(GB)  Arch      Price/hr   Est. Cost  Source                                                                                                   
  ────────  ─────────────────────────  ─────  ──────────  ────────  ────────  ────────  ──────────  ────────────────────                                                                                     
  aws       t3.large                       2        8.00         0  x86_64     $0.0832       $1.04  instance_type_monitor (defaults) (resolved)                                                              
  gce       e2-standard-2                  2        8.00         0  x86_64     $0.0670       $0.84  gce_instance_type_monitor (defaults) (resolved)                                                          
  azure     Standard_D2as_v5               2        8.00         0  x86_64     $0.0860       $1.07  azure_instance_type_monitor (defaults) (resolved)                                                        
  oci       VM.Standard.E4.Flex:1:8        2        8.00         0  x86_64     $0.0490       $0.61  oci_instance_type_monitor (defaults) (resolved) 

Estimated Total Cost                                                                                                                                                                                         
  Cloud          Total                                                                                                                                                                                       
  ────────  ──────────                                                                                                                                                                                       
  aws       $    59.79                                                                                                                                                                                       
  gce       $    41.09                                                                                                                                                                                       
  azure     $    55.39 
```

pehala: I would like to now bind this PR to the sizing one, they should be independent. We cans witch to using dynamic system after it is merged

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -1,25 +1,35 @@
...
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "keyspace1.standard1", "interval": 5}']
-n_db_nodes: 6
-n_loaders: 1
+
+n_db_nodes: 12
+n_loaders: 2
 simulated_racks: 3
+round_robin: true
 
-instance_type_db: 'i7i.2xlarge'
```
pehala: I would leave the AWS variant as well. Decrease it to match the google one (4CPU, 32gb mem), i.e. i7i.xlarge should suffice. I would add note that the tuning was done only for GCE

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
pehala, line 1: Lets add description here with the expected CPU utilization, disk utilization and other parameters we wanted from this test

jsmolar:
Load is between 20% and 70%
image

pehala (dismissed):

fruch:
@jsmolar @pehala  I don't understand, the configuration would support both GCE and AWS, and still gonna default to AWS ? i.e. the pipeline wasn't touched...

i.e. why the word GCE is it the title here, if it's not moving to use it ?

pehala:
@fruch FYI: Coderabbit added unnecessary and wrong lables to this PR, promotedto master and backports done for some reason.

jenkins-pipelines/oss/tier1/longevity-asymmetric-cluster-12h.jenkinsfile
cezarmoise, line 1: Is this job in any triggers that need to be updated?

jsmolar: I didn't move or rename the job, so it isn't needed.

cezarmoise: I meant from aws trigger to gce trigger, since the scylla images are different

pehala: Cezar is correct, we need to move it to the GCP triggers

jsmolar: I removed it from the AWS trigger

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -1,27 +1,42 @@
...
+n_db_nodes: 12
+n_loaders: 2
 simulated_racks: 3
 
 sizing_db:  # no OCI match (DenseIO min 16 vCPUs)
-  vcpu: 8
-  memory: '>=64'
+  vcpu: 4
+  memory: '>=32'
   arch: x86_64
```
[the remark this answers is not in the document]
jsmolar: @gith according to https://github.com/scylladb/scylla-cluster-tests/pull/14576/changes `gce_n_local_ssd_disk_db` should be removed.

jenkins-pipelines/master-triggers/sct_triggers/tier1-custom-time-trigger.xml
```diff
@@ -73,8 +73,7 @@ provision_type=on_demand
 post_behavior_db_nodes=destroy
 post_behavior_monitor_nodes=destroy
 stress_duration=1440
-requested_by_user=pehala
-region=us-east-1</properties>
```
pehala: why? this is not your test

jenkins-pipelines/oss/tier1/longevity-asymmetric-cluster-12h.jenkinsfile
pehala: And you didnt add it to the GCE trigger :) https://github.com/scylladb/scylla-cluster-tests/blob/master/jenkins-pipelines/oss/sct_triggers/tier1-gce-custom-time-trigger.xml

pehala (changes requested):

jenkins-pipelines/master-triggers/sct_triggers/tier1-custom-time-trigger.xml
pehala: It might be, I confused it with the TWCS above

pehala (dismissed):

jenkins-pipelines/oss/tier1/longevity-asymmetric-cluster-12h.jenkinsfile
```diff
@@ -4,8 +4,8 @@
 def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
 
 longevityPipeline(
-    backend: 'aws',
-    region: 'eu-west-1',
+    backend: 'gce',
+    region: 'us-east1',
```
fruch:
```suggestion
    gce_datacenter: "us-east1",
```

fruch (dismissed):
LGTM, on small change in the pipeline

roydahan:
@jsmolar / @pehala what is the motivation behind this PR / issue?
It's not explained anywhere.

pehala:
> @jsmolar / @pehala what is the motivation behind this PR / issue?
> It's not explained anywhere.

As with other Tier 1 changes, increasing storage utilization and making the test more intensive, + move to GCP to save AWS costs

roydahan:
> As with other Tier 1 changes, increasing storage utilization and making the test more intensive, + move to GCP to save AWS costs

Well, I'm not sure that running 66 times the same test worth this goal, we also need to make sure we track the costs of such efforts.

jsmolar:
> Well, I'm not sure that running 66 times the same test worth this goal, we also need to make sure we track the costs of such efforts.

Due to `[Errno 9] Bad file descriptor')` Issue, 80% of these runs lasted ~20 mins, and I tried to use spot as much as possible. I completely agree with you, and I had costs in mind when running this.

pehala:
> Well, I'm not sure that running 66 times the same test worth this goal, we also need to make sure we track the costs of such efforts.

I agree we should track it. All of Tier 1 changes are in correct direction 1) They increase utilization and as such test scylla under much more interesting conditions, making the test better 2) They update instances to either the newest or switch backend to the GCE to save on AWS

So either we get our money worth with a better test, or we save on some costs. Overall cost of running should be easily amortized over the long period and amounts of tests we run on releases, resulting in savings, even if we spent a bit immediately.

If you have problems with suggested changes, feel free to write them on the Jira issue in the epic or reach to me offline, we can discuss them individually before they are even transformed into PRs

roydahan:
In general I don't have any issue with the direction, just tried to understand why simple change took so long and so many runs.
Eventually, we need to find the balance between doing what's needed (like moving to GCP) and doing the nice to have like increasing the utilization of the test while keeping the costs reasonable.

Another approach is to try guesstimate and have a run or two in staging and get this test further stabilized over time in tier1.
Not ideal, but could be a compromise.

pehala:
> Eventually, we need to find the balance between doing what's needed (like moving to GCP) and doing the nice to have like increasing the utilization of the test while keeping the costs reasonable.
> 

Understandable, like Jakub mentioned it was caused by factors we couldn't affect (the big python driver bug) and as such hopefully shouldn't replicate in other tests after the fix is propagated

> Another approach is to try guesstimate and have a run or two in staging and get this test further stabilized over time in tier1. Not ideal, but could be a compromise.

We can do that, but I dont like the idea of merging something we did not run before, we could do it for some incremental changes

vponomaryov:
> > Well, I'm not sure that running 66 times the same test worth this goal, we also need to make sure we track the costs of such efforts.
> 
> Due to `[Errno 9] Bad file descriptor')` Issue, 80% of these runs lasted ~20 mins, and I tried to use spot as much as possible. I completely agree with you, and I had costs in mind when running this.

...
> Understandable, like Jakub mentioned it was caused by factors we couldn't affect (the big python driver bug)

If you refer to the following bug:
- https://github.com/scylladb/python-driver/issues/614
- https://scylladb.atlassian.net/browse/SCT-83

Then it was already fixed and SCT has been updated to the fixed python driver version already since June 16.

So, is the state of this PR still actual in that situation?

jsmolar:
> > > Well, I'm not sure that running 66 times the same test worth this goal, we also need to make sure we track the costs of such efforts.
> > 
> > 
> > Due to `[Errno 9] Bad file descriptor')` Issue, 80% of these runs lasted ~20 mins, and I tried to use spot as much as possible. I completely agree with you, and I had costs in mind when running this.
> 
> ...
> 
> > Understandable, like Jakub mentioned it was caused by factors we couldn't affect (the big python driver bug)
> 
> If you refer to the following bug:
> 
>     * [Driver reported "[Errno 9] Bad file descriptor" python-driver#614](https://github.com/scylladb/python-driver/issues/614)
> 
>     * https://scylladb.atlassian.net/browse/SCT-83
> 
> 
> Then it was already fixed and SCT has been updated to the fixed python driver version already since June 16.
> 
> So, is the state of this PR still actual in that situation?

Yes, thank you! I already successfully ran this after the fix you mentioned.

vponomaryov:
Before the shard count was randomized from 4 to 7 on 6 DB nodes.
Now it is from 2 to 3 on 12 DB nodes.
The potential shard count skew became really different.
So, is it intended way of covering the asymmetric DB setup?

jsmolar:
> Before the shard count was randomized from 4 to 7 on 6 DB nodes. Now it is from 2 to 3 on 12 DB nodes. The potential shard count skew became really different. So, is it intended way of covering the asymmetric DB setup?

Good point, I changed db instances to 8 x vCPU

pehala (dismissed):

cezarmoise (approved):
LGTM, apart from 2 small issues

jenkins-pipelines/oss/sct_triggers/tier1-gce-custom-time-trigger.xml
```diff
@@ -45,7 +45,10 @@ stress_duration=$stress_duration</properties>
               <textParamValueOnNewLine>false</textParamValueOnNewLine>
             </hudson.plugins.parameterizedtrigger.PredefinedBuildParameters>
           </configs>
-          <projects>../tier1/longevity-large-partition-200k-pks-4days-gce-test</projects>
+          <projects>
+              ../tier1/longevity-large-partition-200k-pks-4days-gce-test,
+              ../tier1/longevity-asymmetric-cluster-12h-test,
```
cezarmoise:
```suggestion
              ../tier1/longevity-asymmetric-cluster-12h-test
```
remove trailing comma

test-cases/longevity/longevity-150GB-12h-autorization-LimitedMonkey-cql-stress.yaml
cezarmoise, line 1: There still exists a pipeline that references the old name
`jenkins-pipelines/oss/tier1/longevity-asymmetric-cluster-12h.jenkinsfile`

cezarmoise (changes requested):

cezarmoise (dismissed):

vponomaryov:
@jsmolar 
The main change LGTM

Please, look at the recently merged following PR:
- https://github.com/scylladb/scylla-cluster-tests/pull/14145

And check whether you need to adjust changes in this PR or not.

vponomaryov (dismissed):
LGTM

pehala (dismissed):

vponomaryov:
@fruch this one is waiting for your review

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,66 @@
...
+n_loaders: 2
+simulated_racks: 3
+round_robin: true
+
+instance_type_db: 'i7i.2xlarge'
+instance_type_loader: 'c6i.large'
+
+gce_instance_type_db: 'n2-highmem-8'
+gce_instance_type_loader: 'e2-standard-2'
+gce_n_local_ssd_disk_db: 1
```
roydahan, line 46: Why isn't it following the sizing as it was?

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,66 @@
...
+
+run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "keyspace1.standard1", "interval": 5}']
+
+n_db_nodes: 12
+n_loaders: 2
+simulated_racks: 3
+round_robin: true
+
+instance_type_db: 'i7i.2xlarge'
+instance_type_loader: 'c6i.large'
```
roydahan, line 42: Same, wrong.

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,66 @@
...
+  - "cql-stress-cassandra-stress write cl=ALL n=225000000 -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=500 -col 'size=FIXED(256) n=FIXED(4)' -pop seq=225000001..450000000 -log interval=15"
+
+stress_cmd: ["cql-stress-cassandra-stress write cl=QUORUM duration=11h -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate 'threads=200 throttle=30000/s' -col 'size=FIXED(256) n=FIXED(4)' -pop seq=450000001..600000000 -log interval=15",
+             "cql-stress-cassandra-stress read cl=QUORUM duration=11h -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate 'threads=200 throttle=30000/s' -col 'size=FIXED(256) n=FIXED(4)' -pop seq=1..450000000 -log interval=5"]
+
+run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "keyspace1.standard1", "interval": 5}']
+
+n_db_nodes: 12
```
roydahan: Why is it increased to 12?

roydahan (changes requested):

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
pehala: We were asked to add a test which tests on cluster with more nodes as part of the Tier 1 retune.

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
pehala: Until https://github.com/scylladb/scylla-cluster-tests/pull/15368 is in, we cannot claim 40% data utilization with dynamic sizing. We can add the sizing back for other clouds but not use it for GCE, but thats about it

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
pehala: loaders can use dynamic sizing. Main instance type cannot if we want specific data utilization

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
roydahan: by whom?

pehala: Yaniv

roydahan: Well, I didn't approve it.

pehala: You were part of the discussions, the doc was open, you had a lot of comments there and it was adjusted according to them.

@bhalevy I will leave this to you. This test was meant to be the one in Tier 1 which actually has more nodes to catch more interesting issues and be closer to customer deployment.

roydahan: Not approved.
This is not the case to do it,
It's expensive test and there is no real justification for it.

If we wanted to invest more money, this case would be better candidate for larger nodes than more nodes, because this case randomize the amount of shards each node has. So more shards per node gets us more interesting mix of shards count.

bhalevy: @mykaul my understanding is that we want a tier-1 test for higher scale.
The question is whether to scale nodes or shards (with bigger instances).
Ideally, we should test both, but given the limited budget we should pick one.
What would you prefer?

mykaul: I'm afraid it's not either/or - we have both - we have customer with .8xlarge and we have customers with 10's (or more!) of nodes. I'd focus first on the .8xlarge nodes (so more shards).
I do believe we need to thoroughly ensure the tests are efficient and cover a lot of ground, not just set up a large node and run some workload against it.
I'd also add to the mix - substantial no. of client connections per shard.

roydahan: If we want to extend tier1 for scale (many shards, many clients, many nodes), I think we should introduce another test for it (existing one or a new one) that will run significantly shorter than 12 hours and have its costs known and controlled.
(e.g. I don't think anyone calculated the new costs of this change)

pehala:
@jsmolar lets go back to 6 nodes, there no longer an agreement to expand this test to 12 nodes

jsmolar:
I return `n_db_nodes` to 6 and downscale the load and disk usage:
image
image

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
roydahan: @jsmolar please put back the sizing instead of specific instance types.

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,60 @@
...
+gce_instance_type_db: 'n2-highmem-8'
+gce_instance_type_loader: 'e2-standard-2'
+gce_n_local_ssd_disk_db: 1
+
+nemesis_class_name: 'SisyphusMonkey'
+nemesis_selector: 'limited'
+nemesis_interval: 30
+nemesis_during_prepare: false
+
+user_prefix: 'longevity-6nodes-12h'
```
roydahan: A better prefix would be longevity-assymetric-12h.

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,60 @@
...
+
+instance_type_db: 'i7i.2xlarge'
+instance_type_loader: 'c6i.large'
+
+gce_instance_type_db: 'n2-highmem-8'
+gce_instance_type_loader: 'e2-standard-2'
+gce_n_local_ssd_disk_db: 1
+
+nemesis_class_name: 'SisyphusMonkey'
+nemesis_selector: 'limited'
```
roydahan, line 47: did you intentionally removed the nemesis_seed?

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
```diff
@@ -0,0 +1,60 @@
...
+nemesis_interval: 30
+nemesis_during_prepare: false
+
+user_prefix: 'longevity-6nodes-12h'
+
+space_node_threshold: 322122547
```
roydahan: I don't think it's needed at all.

roydahan (changes requested):

test-cases/longevity/longevity-12h-autorization-LimitedMonkey-cql-stress.yaml
jsmolar: Yes, this should be slowly removed from most of the configs.

pehala (approved):

roydahan (approved):

PR #14633, opened 2026-05-13
fix(workflows): use paginated GitHub REST API for cache-issues

fruch:
## Summary

Fixes the root cause of [DTEST-220](https://scylladb.atlassian.net/browse/DTEST-220): the `cache-issues` workflow was silently dropping issues because `gh issue list --limit 5000` is insufficient for repos with >5000 closed issues (scylladb/scylladb has 12,700+).

**Root cause**: Commit fb63e47d (May 4 2026) reduced `--limit` from 30000 to 5000 per state, causing ~7,600 closed issues to be missing from the S3 cache. This broke `@pytest.mark.require()` markers in scylla-dtest that depend on cached issue state.

**Fix**: Replace `gh issue list --limit` with GitHub GraphQL API using cursor-based pagination. GraphQL returns only requested fields (number, state, labels, title) instead of full issue payloads, making it significantly faster.

## Changes

- `.github/workflows/cache-issues.yaml`: Rewritten to use GraphQL cursor pagination
- `.github/workflows/test_cache_issues_graphql.sh`: Standalone validation script

## Performance Comparison

| Method | Time (scylladb/scylladb closed, 12,700+ issues) |
|--------|--------------------------------------------------|
| `gh issue list --limit 5000` (before) | ~1m but **drops issues silently** |
| `gh api --paginate` (REST) | ~5m |
| **GraphQL (this PR)** | **~1m48s** |

## Validation

```
$ ./test_cache_issues_graphql.sh scylladb closed
=== Testing GraphQL pagination for scylladb/scylladb (state=closed) ===
  page 10: total so far 1000
  page 20: total so far 2000
  ...
  page 120: total so far 12000

=== Results ===
Pages:          127
Issues fetched: 12653

--- Validation ---
PASS: Issue #9216 found (the issue that triggered DTEST-220)
PASS: Got 12653 closed issues (expected 10000+)

=== All checks passed ===

real    1m48.450s
```

**Before** (with `--limit 5000`): Only 5000 closed issues cached, #9216 missing → dtest xdist collection failures.
**After** (with GraphQL): All 12,653 closed issues cached, #9216 present, ~2.7x faster than REST pagination.

## Related

- dtest fix: https://github.com/scylladb/scylla-dtest/pull/6970
- Follow-up improvement: [QAINFRA-104](https://scylladb.atlassian.net/browse/QAINFRA-104) (cache only referenced issues)
- Jira: [DTEST-220](https://scylladb.atlassian.net/browse/DTEST-220)

[DTEST-220]: https://scylladb.atlassian.net/browse/DTEST-220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[QAINFRA-104]: https://scylladb.atlassian.net/browse/QAINFRA-104?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/cache-issues.yaml
+++ b/.github/workflows/cache-issues.yaml
@@ -18,23 +18,21 @@ jobs:
       - run: |
           set -euo pipefail
 
-          BATCH_SIZE=5000
           MAX_RETRIES=3
-          TEMPLATE='{{range .}}{{.number}},{{.state}},{{range .labels}}{{.name}}|{{end}},{{.title}}{{println ""}}{{end}}'
 
           fetch_with_retry() {
-            local cmd="$1"
-            local output_file="$2"
+            local output_file="$1"
+            shift
             local attempt=1
             while [ $attempt -le $MAX_RETRIES ]; do
-              if eval "$cmd" >> "$output_file"; then
+              if "$@" > "$output_file"; then
                 return 0
               fi
-              echo "Attempt $attempt/$MAX_RETRIES failed, retrying in 10s..."
+              echo "  Attempt $attempt/$MAX_RETRIES failed, retrying in 10s..."
               sleep 10
               attempt=$((attempt + 1))
             done
-            echo "ERROR: Failed after $MAX_RETRIES attempts: $cmd"
+            echo "ERROR: Failed after $MAX_RETRIES attempts: $*"
             return 1
           }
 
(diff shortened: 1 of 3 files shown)


.github/workflows/test_cache_issues.sh
```diff
@@ -0,0 +1,73 @@
+#!/bin/bash
```
fruch, line 1: we don't really need it, but we can keep it for manually testing (without any other tricks with github actions)

.github/workflows/cache-issues.yaml
```diff
@@ -45,14 +43,26 @@ jobs:
...
+            echo "  Total issues: $count"
 
             echo "Fetching PRs for scylladb/$repo..."
             > issues/pull-requests/scylladb_$repo.csv
-            for state in open closed merged; do
-              fetch_with_retry "gh pr list --state $state --json number,state,labels,title --limit $BATCH_SIZE --template '$TEMPLATE' --repo scylladb/$repo" "issues/pull-requests/scylladb_$repo.csv"
+            for state in open closed; do
+              fetch_with_retry "issues/pull-requests/scylladb_$repo.csv.tmp" \
+                gh api --paginate "repos/scylladb/$repo/pulls?state=$state&per_page=100&direction=asc" \
+                  --jq '.[] | "\(.number),\(if .merged_at then "MERGED" elif .state == "closed" then "closed" else .state end),\([.labels[].name] | join("|")),\(.title)"'
```
soyacz: did you verify these merged are properly collected?

soyacz (dismissed):
LGTM

.github/workflows/cache-issues.yaml
fruch: yes it's collected correctly, very slow, but correct.

PR #14651, opened 2026-05-14
fix(longevity-50gb): Disable simulated_racks dues to multi-az

cezarmoise:
Since the test has multiple availability zones, disable simulated racks as the interaction produces unexpected results.

Fixes: SCT-84

### Testing

- [ ] Before: https://argus.scylladb.com/tests/scylla-cluster-tests/d3a16144-8b29-4cc8-b158-b1b60a192daa
- [x] After: https://argus.scylladb.com/tests/scylla-cluster-tests/6c62bb86-5e58-4e42-8018-90f55206dff5
  Checked by setting `SCT_SIMULATED_RACKS=0`

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-50GB-3days-authorization-and-tls-ssl.yaml
+++ b/test-cases/longevity/longevity-50GB-3days-authorization-and-tls-ssl.yaml
@@ -8,6 +8,7 @@ stress_read_cmd: ["cassandra-stress read cl=QUORUM duration=4320m -mode cql3 nat
 
 run_fullscan: ['{"mode": "table_and_aggregate", "ks_cf": "keyspace1.standard1", "interval": 60}']
 availability_zone: 'a,b,c'
+simulated_racks: 0
 n_db_nodes: 6
 n_loaders: 3
 


vponomaryov (approved):
LGTM

cezarmoise:
@scylladb/qa-maintainers 

The issue here is that nodes are added like this
```python
for az_index in range(azs):
    rack = None if self.params.get("simulated_racks") else az_index
    self.add_nodes(nodes_per_az[az_index], rack=rack, enable_auto_bootstrap=self.auto_bootstrap)
```

So if we have 3 AZs, 6 nodes, and simulated racks, it will run `add_nodes` 3 times, with 2 nodes each time, rack=None
So we will get
AZ a: RACK0, RACK1
AZ b: RACK0, RACK1
AZ c: RACK0, RACK1

The only way this works if (NUM_NODES / NUM_AZs) % SIMULATED_RACKS == 0.

There are other places where this could happen, but avalilability_zone may be set by the trigger in those cases.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/c7567aa94eaaffd3d8b0239ee043b65b44344afd: just fixed typo in commit message

pehala (approved):

pehala:
> So if we have 3 AZs, 6 nodes, and simulated racks, it will run `add_nodes` 3 times, with 2 nodes each time, rack=None So we will get AZ a: RACK0, RACK1 AZ b: RACK0, RACK1 AZ c: RACK0, RACK1
> 
> The only way this works if (NUM_NODES / NUM_AZs) % SIMULATED_RACKS == 0.
> 
> There are other places where this could happen, but avalilability_zone may be set by the trigger in those cases.

I suspect the invariant hold for majority of the tests, but if you have idea how to do it better, patch is for sure welcome

soyacz:
basically, when using simulated_racks we should not provide availability zones - these two are not compatible (we cant use real racks if we simulate them).
We could think of failing fast when this is set like this (or test lint it), or use thre first az from the list if simulated racks are used.

roydahan:
> @scylladb/qa-maintainers
> 
> The issue here is that nodes are added like this
> 
> ```python
> for az_index in range(azs):
>     rack = None if self.params.get("simulated_racks") else az_index
>     self.add_nodes(nodes_per_az[az_index], rack=rack, enable_auto_bootstrap=self.auto_bootstrap)
> ```
> 
> So if we have 3 AZs, 6 nodes, and simulated racks, it will run `add_nodes` 3 times, with 2 nodes each time, rack=None So we will get AZ a: RACK0, RACK1 AZ b: RACK0, RACK1 AZ c: RACK0, RACK1
> 
> The only way this works if (NUM_NODES / NUM_AZs) % SIMULATED_RACKS == 0.
> 
> There are other places where this could happen, but avalilability_zone may be set by the trigger in those cases.

Something which I don't understand.
In availability_zones here it's also defined as 3 availability zones, so why does setting simulated_racks behaves differently?

cezarmoise:
> > @scylladb/qa-maintainers
> > The issue here is that nodes are added like this
> > ```python
> > for az_index in range(azs):
> >     rack = None if self.params.get("simulated_racks") else az_index
> >     self.add_nodes(nodes_per_az[az_index], rack=rack, enable_auto_bootstrap=self.auto_bootstrap)
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So if we have 3 AZs, 6 nodes, and simulated racks, it will run `add_nodes` 3 times, with 2 nodes each time, rack=None So we will get AZ a: RACK0, RACK1 AZ b: RACK0, RACK1 AZ c: RACK0, RACK1
> > The only way this works if (NUM_NODES / NUM_AZs) % SIMULATED_RACKS == 0.
> > There are other places where this could happen, but avalilability_zone may be set by the trigger in those cases.
> 
> Something which I don't understand. In availability_zones here it's also defined as 3 availability zones, so why does setting simulated_racks behaves differently?

When `simulated_racks` is not 0, I see that all code, like in `AWSCluster.add_nodes` give that priority.

```python
instance_az = 0 if self.params.get("simulated_racks") else rack
...
rack_distribution = [count if i == instance_az else 0 for i in range(self.racks_count)]
```

This will always create a rack_distribution like [2, 0, 0] (where 2 = count = nodes_per_az[az_index])
Which means when calling `_create_or_find_instances`, it will always have `az_idx=0`, so all nodes will be on the same AZ every time when simulated racks, regardless of what `availability_zones`  says

PR #14669, opened 2026-05-14
feature(adaptive-timeouts): per-operation timeout multipliers

aleksbykov:
Implement per-operation multipliers for adaptive timeouts. Each operation
can now have its own independent scaling factor configured via
'adaptive_timeout_multipliers'.

When configured, only the multiplier matching the current operation is
applied. Operations without a configured multiplier default to 1.0.
When not configured (default null), existing behavior is preserved.

Example usage in test-case YAML:
  adaptive_timeout_multipliers:
    decommission: 4
    new_node: 2

This scales decommission timeouts by 4x and new_node by 2x.

Implementation:
- AdaptiveTimeoutMultipliers RootModel with confloat(gt=0) for native
  Pydantic value validation and get_multiplier() method
- Valid keys are derived dynamically from Operations enum (operation.value[0]),
  so any new operation added to the enum is automatically supported
- _get_operation_timeout_factor() resolves multiplier using operation.value[0]
- Config validation rejects unknown keys with clear error messages listing
  all valid operation names
- Bool values are explicitly rejected to prevent accidental misconfiguration

Workaround for https://scylladb.atlassian.net/browse/SCT-356

### Testing

- [job passed](https://argus.scylladb.com/tests/scylla-cluster-tests/83deee7a-04a8-44a4-a44b-1ab89348f6c8)

### Reminders

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -2041,6 +2041,15 @@ Store adaptive timeout metrics in Argus. Disabled for performance tests only.
 **type:** bool
 
 
+## **adaptive_timeout_multipliers** / SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS
+
+Optional dict of adaptive-timeout multipliers keyed by operation name (from Operations enum value[0], e.g. decommission, remove_node, new_node, repair, etc.). If the current operation key is absent, multiplier 1.0 is used.<br>YAML example:<br>adaptive_timeout_multipliers:<br>  decommission: 4<br>  new_node: 2<br>Environment variable examples:<br>SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS="{'decommission': 4, 'new_node': 2}"<br>Or dot-notation: SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS.decommission=4
+
+**default:** N/A
+
+**type:** sdcm.sct_config.AdaptiveTimeoutMultipliers
+
+
 ## **gce_n_local_ssd_disk_monitor** / SCT_GCE_N_LOCAL_SSD_DISK_MONITOR
 
 Number of local SSD disks for monitor nodes in Google Compute Engine
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -35,7 +35,8 @@ from distutils.util import strtobool
 import anyconfig
 from argus.client.sct.types import Package
 from packaging import version
-from pydantic import BaseModel, Field, ConfigDict, fields as pydantic_fields
+from pydantic import BaseModel, Field, ConfigDict, RootModel, fields as pydantic_fields, model_validator
+from pydantic.types import confloat
 from typing_extensions import Annotated
 from pydantic.functional_validators import BeforeValidator
 from pydantic.fields import FieldInfo
@@ -262,6 +263,52 @@ def dict_or_str(value: dict | str | None) -> dict | None:
 
 
 DictOrStr = Annotated[dict | str, BeforeValidator(dict_or_str)]
+
+
+class AdaptiveTimeoutMultipliers(RootModel):
+    """Per-operation multipliers for adaptive timeouts.
+
+    Keys must be valid operation names from Operations enum (operation.value[0]),
+    e.g. decommission, remove_node, new_node, repair, rebuild, etc.
+    Missing keys default to multiplier 1.
+
+    YAML config example::
+
+        adaptive_timeout_multipliers:
+          decommission: 4
+          new_node: 4
+          remove_node: 4
+
+    Environment variable examples:
+
+        SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS="{'decommission': 2, 'new_node': 3}"
+
+    Or using dot-notation (same pattern as SCT_STRESS_IMAGE.*):
+
+        SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS.decommission=4
+        SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS.new_node=3
+    """
+
+    root: dict[str, confloat(gt=0)] = Field(default_factory=dict)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _validate_operations(cls, value):
+        if not isinstance(value, dict):
+            return value
+
+        # cyclic-import: Operations imports from sct_config indirectly via cluster
+        from sdcm.utils.adaptive_timeouts import Operations  # noqa: PLC0415
+
+        valid_keys = {op.value[0] for op in Operations}
+        for key in value.keys():
+            if key not in valid_keys:
+                raise ValueError(f"Unknown operation key '{key}'. Valid keys: {sorted(valid_keys)}")
+        return value
+
+    def get_multiplier(self, operation_key: str) -> float:
+        """Return multiplier for the given operation key, or 1.0 if not configured."""
+        return float(self.root.get(operation_key, 1.0))
 
 
 def dict_or_str_or_pydantic(value: dict | str | BaseModel | None) -> dict | BaseModel | None:
@@ -1262,6 +1309,20 @@ class SCTConfiguration(BaseModel):
     )
     adaptive_timeout_store_metrics: Boolean = SctField(
         description="Store adaptive timeout metrics in Argus. Disabled for performance tests only.",
+    )
+    adaptive_timeout_multipliers: Annotated[AdaptiveTimeoutMultipliers, BeforeValidator(dict_or_str_or_pydantic)] = (
+        SctField(
+            description="Optional dict of adaptive-timeout multipliers keyed by operation name "
+            "(from Operations enum value[0], e.g. decommission, remove_node, new_node, repair, etc.). "
+            "If the current operation key is absent, multiplier 1.0 is used.<br>"
+            "YAML example:<br>"
+            "adaptive_timeout_multipliers:<br>"
+            "  decommission: 4<br>"
+            "  new_node: 2<br>"
+            "Environment variable examples:<br>"
+            "SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS=\"{'decommission': 4, 'new_node': 2}\"<br>"
+            "Or dot-notation: SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS.decommission=4",
+        )
     )
 
     # Google Compute Engine options
--- a/sdcm/utils/adaptive_timeouts/__init__.py
+++ b/sdcm/utils/adaptive_timeouts/__init__.py
@@ -16,6 +16,20 @@ from sdcm.utils.features import is_tablets_feature_enabled
 
 LOGGER = logging.getLogger(__name__)
 
+
+def _get_operation_timeout_factor(params, operation: "Operations") -> float:
+    """Return timeout multiplier for the current operation.
+
+    Looks up operation.value[0] in the configured multipliers dict.
+    Missing config or missing operation key returns 1.0.
+    """
+    multipliers = params.get("adaptive_timeout_multipliers")
+    if multipliers is None:
+        return 1.0
+
+    return multipliers.get_multiplier(operation.value[0])
+
+
 TABLETS_SOFT_TIMEOUT = 1 * 60 * 60
 TABLETS_HARD_TIMEOUT = 3 * 60 * 60
 _STREAMING_OVERHEAD = 600  # add 10 minutes overhead for operations other than streaming
@@ -260,6 +274,11 @@ def adaptive_timeout(  # noqa: PLR0914
         hard_timeout = None
     if store_metrics:
         load_metrics.update(TestInfoServices.get(node))
+
+    # Apply operation-based timeout scaling (opt-in via config)
+    operation_factor = _get_operation_timeout_factor(node.parent_cluster.params, operation)
+    soft_timeout = soft_timeout * operation_factor if soft_timeout else None
+    hard_timeout = hard_timeout * operation_factor if hard_timeout else None
 
     start_time = time.monotonic()
     timeout_occurred = False
--- a/unit_tests/unit/test_adaptive_timeouts.py
+++ b/unit_tests/unit/test_adaptive_timeouts.py
@@ -20,7 +20,7 @@ import pytest
 from invoke import Result
 
 from sdcm.remote import RemoteCmdRunnerBase
-from sdcm.sct_config import SCTConfiguration
+from sdcm.sct_config import AdaptiveTimeoutMultipliers, SCTConfiguration
 from sdcm.utils.adaptive_timeouts.load_info_store import (
     AdaptiveTimeoutStore,
     NodeLoadInfoService,
@@ -28,7 +28,13 @@ from sdcm.utils.adaptive_timeouts.load_info_store import (
     _I4I_LARGE_BASELINE_THROUGHPUT_MB_PER_SEC,
     _I4I_LARGE_SHARD_COUNT,
 )
-from sdcm.utils.adaptive_timeouts import _STREAMING_OVERHEAD, Operations, adaptive_timeout, TABLETS_HARD_TIMEOUT
+from sdcm.utils.adaptive_timeouts import (
+    _STREAMING_OVERHEAD,
+    _get_operation_timeout_factor,
+    Operations,
+    adaptive_timeout,
+    TABLETS_HARD_TIMEOUT,
+)
 from unit_tests.lib.fake_cluster import DummyDbCluster
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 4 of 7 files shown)


unit_tests/unit/test_adaptive_timeouts.py
```diff
@@ -332,3 +338,132 @@ def test_expected_throughput_falls_back_to_estimate_when_zero(fake_node):
...
+        throughput = _I4I_LARGE_BASELINE_THROUGHPUT_MB_PER_SEC / _I4I_LARGE_SHARD_COUNT * 3
+        estimated = int(102400 / throughput)
+        expected_hard = (estimated * 4 + _STREAMING_OVERHEAD) * 2
+        assert timeout == expected_hard
+
+
+# --- Config-level validation tests ---
+
+
+class TestValidateFeatureMultipliers:
```
pehala: no classes....

docs/configuration_options.md
```diff
@@ -2041,6 +2041,15 @@ Store adaptive timeout metrics in Argus. Disabled for performance tests only.
...
 
 
+## **adaptive_timeout_feature_multipliers** / SCT_ADAPTIVE_TIMEOUT_FEATURE_MULTIPLIERS
+
+Optional dict of feature multipliers for adaptive timeouts. When configured, all adaptive timeout values are multiplied by the sum of the declared multiplier values. Supported keys: cdc, lwt, mv, si. Values must be positive numbers. Example: {cdc: 4, mv: 4} produces factor=8. Default: null (no scaling).
+
+**default:** N/A
+
+**type:** dict
+
```
pehala: If is meant to be a dict, replace it with pydantic class.

soyacz:
IMHO, we should not introduce this change. Adding more params complicate things up, people need to know about them to use (and remember).
My proposal would be to increase hard timeout for `_get_decommission_timeout`, or even set it to 2h-3h - so we never critically finish the test and see actual value.
Another thing is to improve `_get_decommission_timeout` to base it on more metrics. E.g. this method is not taking into account `cpu_load_5` which is high when these timeouts happened (9.8 on 16 cpu test and 31 on 32 cpu test).
This could also be an Scylla issue - decommission taking long even while we use <5% of cluster disk capacity should not take long.

sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -261,6 +275,12 @@ def adaptive_timeout(  # noqa: PLR0914
     if store_metrics:
         load_metrics.update(TestInfoServices.get(node))
 
+    # Apply feature-based timeout scaling (opt-in via config)
+    feature_factor = _get_feature_timeout_factor(node.parent_cluster.params)
+    if feature_factor != 1:
```
cezarmoise: Don't really need this `if`, it's just multiplying by 1.

pehala:
> IMHO, we should not introduce this change. Adding more params complicate things up, people need to know about them to use (and remember).
> My proposal would be to increase hard timeout for `_get_decommission_timeout`, or even set it to 2h-3h - so we never critically finish the test and see actual value.

My suggestion would be to do both, I think the increase of timeout for decomission make sense, and at the same time we can add a flat multiplier value to all parameters that can be used as a last resort, not set frequently but can help us overcome situations where we know more tuning of the timeouts is needed

aleksbykov:
@pehala , @soyacz , the main pr is here https://github.com/scylladb/scylla-cluster-tests/pull/14670. it will dynamically detect features ans store appropriate statistics( at least it is an idea.) also it could be update with additional metrics.

unit_tests/unit/test_adaptive_timeouts.py
aleksbykov: fixed.

sdcm/utils/adaptive_timeouts/__init__.py
aleksbykov: fixed

docs/configuration_options.md
aleksbykov: fixed

k0machi (commented):
@aleksbykov  What's the status on this one? I assume we need this to run 2026.2 pipelines, so I would like to get a resolution on whether or not we can run the extended tests.

cc @pehala @soyacz

soyacz:
> > IMHO, we should not introduce this change. Adding more params complicate things up, people need to know about them to use (and remember).
> > My proposal would be to increase hard timeout for `_get_decommission_timeout`, or even set it to 2h-3h - so we never critically finish the test and see actual value.
> 
> My suggestion would be to do both, I think the increase of timeout for decomission make sense, and at the same time we can add a flat multiplier value to all parameters that can be used as a last resort, not set frequently but can help us overcome situations where we know more tuning of the timeouts is needed

Ok, but let's have this multiplier per AdaptiveTimeout, not per feature. eg. `adaptive_timeout_multiplier` with `DECOMMISSION: 3.0`. By default all 1.0
This way it is going to be more general and more tunable per timeout - as some timeouts should not be affected by features.

pehala:
> @aleksbykov What's the status on this one? I assume we need this to run 2026.2 pipelines, so I would like to get a resolution on whether or not we can run the extended tests.

I hope this is not blocking, I believe we should have ways how to remedy this even now, which are sufficient for release testing (this is not first release where this happens AFAIK), but still @aleksbykov needs to confirm

aleksbykov:
> @aleksbykov What's the status on this one? I assume we need this to run 2026.2 pipelines, so I would like to get a resolution on whether or not we can run the extended tests.
> 
> cc @pehala @soyacz

I run tests with this patch in Reproducers folder. Tests mostly passed add issued fixed by this pr is not reproduced. I also update the argus. I know it is not a good practice to use custom branches for release, but those jobs was in reproducers and tests themselves are not yet merged to master, so i reconfigured them and run.

aleksbykov:
> Ok, but let's have this multiplier per AdaptiveTimeout, not per feature. eg. `adaptive_timeout_multiplier` with `DECOMMISSION: 3.0`. By default all 1.0
> This way it is going to be more general and more tunable per timeout - as some timeouts should not be affected by features.

@soyacz , correct me if i am wrong. You suggest to have a dict addaptive_timeouts_multipliers, where keys are operations instead of features:
```yaml
addaptive_timeouts_multipliers:
  decommission: 3.0
  bootstrap: 3.0
  removenode: 4.0
```
and by default it is 1.

aleksbykov:
@soyacz , can you review again

sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -16,6 +16,25 @@
...
+
+    Missing config, missing operation mapping, or missing key all return 1.
+    Operations not listed in _OPERATION_TIMEOUT_KEYS (e.g. TABLET_MIGRATION,
+    SOFT_TIMEOUT) intentionally remain unscaled.
+    """
+    multipliers = params.get("adaptive_timeout_operation_multipliers")
+    if multipliers is None:
+        return 1
+
+    operation_key = _OPERATION_TIMEOUT_KEYS.get(operation)
```
soyacz: I don't see the reason to introduce `_OPERATION_TIMEOUT_KEYS` - `operation.value[0]` should return it.
This way we would not be limited to only these listed in this new const.

sdcm/sct_config.py
```diff
@@ -264,6 +264,31 @@ def dict_or_str(value: dict | str | None) -> dict | None:
...
+
+    Supported keys: decommission, removenode, new_node.
+    Missing keys imply multiplier 1.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    decommission: PositiveFloat | None = None
+    removenode: PositiveFloat | None = None
+    new_node: PositiveFloat | None = None
```
soyacz: instead of listing allowed ones, we could take automatically from `sdcm.utils.adaptive_timeouts.Operations` - just verify if specified multiplier is found in any  Operation value[0].

soyacz (commented):
I think it could be more flexible and not repeating timeouts names in config and const's

sdcm/sct_config.py
aleksbykov: fixed

sdcm/utils/adaptive_timeouts/__init__.py
aleksbykov: fixed

sdcm/sct_config.py
```diff
@@ -1263,6 +1298,12 @@ class SCTConfiguration(BaseModel):
     adaptive_timeout_store_metrics: Boolean = SctField(
         description="Store adaptive timeout metrics in Argus. Disabled for performance tests only.",
     )
+    adaptive_timeout_multipliers: AdaptiveTimeoutMultipliers | None = SctField(
```
pehala: Why it can be a none? empty dict is better than having none no?

sdcm/sct_config.py
```diff
@@ -264,6 +265,40 @@ def dict_or_str(value: dict | str | None) -> dict | None:
 DictOrStr = Annotated[dict | str, BeforeValidator(dict_or_str)]
 
 
+class AdaptiveTimeoutMultipliers(RootModel):
```
pehala, line 268: From this I have no idea how it should look, can you add an example how the config looks here?

sdcm/sct_config.py
```diff
@@ -264,6 +265,40 @@ def dict_or_str(value: dict | str | None) -> dict | None:
...
+            return value
+
+        # cyclic-import: Operations imports from sct_config indirectly via cluster
+        from sdcm.utils.adaptive_timeouts import Operations  # noqa: PLC0415
+
+        valid_keys = {op.value[0] for op in Operations}
+        for key, val in value.items():
+            if key not in valid_keys:
+                raise ValueError(f"Unknown operation key '{key}'. Valid keys: {sorted(valid_keys)}")
+            if isinstance(val, bool):
```
pehala: Why do you verify bool here?

sdcm/sct_config.py
```diff
@@ -264,6 +265,40 @@ def dict_or_str(value: dict | str | None) -> dict | None:
...
+    def _validate_operations(cls, value):
+        if isinstance(value, str):
+            value = dict_or_str(value)
+        if not isinstance(value, dict):
+            return value
+
+        # cyclic-import: Operations imports from sct_config indirectly via cluster
+        from sdcm.utils.adaptive_timeouts import Operations  # noqa: PLC0415
+
+        valid_keys = {op.value[0] for op in Operations}
```
pehala, line 303: @soyacz @aleksbykov What about a reverse ownership instead? Let this class have all of the operations definedhere as static ones and Operations could import from here. Or create a third class which will hold the source of truth.

This has two reasons: 
1) having this be nonstatic is not great as it is not clear what is allowed here
2) Cyclic-import means we are doing dependencies wrong somewhere in the chain

sdcm/sct_config.py
```diff
@@ -264,6 +265,40 @@ def dict_or_str(value: dict | str | None) -> dict | None:
...
+
+    root: dict[str, confloat(gt=0)] = Field(default_factory=dict)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _validate_operations(cls, value):
+        if isinstance(value, str):
+            value = dict_or_str(value)
+        if not isinstance(value, dict):
+            return value
```
pehala: Lets allow only dict, I dont think we need dict_or_str anymore

unit_tests/unit/test_adaptive_timeouts.py
```diff
@@ -332,3 +338,165 @@ def test_expected_throughput_falls_back_to_estimate_when_zero(fake_node):
...
+    fake_node.parent_cluster.params["adaptive_timeout_multipliers"] = None
+
+    with adaptive_timeout(
+        operation=Operations.DECOMMISSION,
+        node=fake_node,
+        stats_storage=adaptive_timeout_store,
+    ) as timeout:
+        assert timeout == 7200  # base decommission timeout unchanged
+
+
```
pehala: These can be parametrized

unit_tests/unit/test_adaptive_timeouts.py
```diff
@@ -332,3 +338,165 @@ def test_expected_throughput_falls_back_to_estimate_when_zero(fake_node):
...
+@pytest.mark.parametrize(
+    "input_val,expected_key,expected_val",
+    [
+        pytest.param({"decommission": 4, "remove_node": 2}, "decommission", 4, id="dict-input"),
+        pytest.param("{'new_node': 3}", "new_node", 3, id="python-string"),
+        pytest.param('{"decommission": 5}', "decommission", 5, id="json-string"),
+    ],
+)
+def test_operation_multiplier_model_validates_input(input_val, expected_key, expected_val):
```
pehala: You have no negative cases, you did nto test it validates input at all. Probably bad naming of the test, but I believe you are doing this exact test somewhere else as well. I would revisit how many tests we actually need and that are unique

unit_tests/unit/test_adaptive_timeouts.py
```diff
@@ -332,3 +338,165 @@ def test_expected_throughput_falls_back_to_estimate_when_zero(fake_node):
...
+    model = AdaptiveTimeoutMultipliers.model_validate(input_val)
+    assert model.root[expected_key] == expected_val
+
+
+def test_operation_multiplier_get_multiplier_returns_1_for_missing_key():
```
pehala: tested by `test_get_operation_timeout_factor` Remove node case, choose which one test to keep

sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -261,6 +275,11 @@ def adaptive_timeout(  # noqa: PLR0914
     if store_metrics:
         load_metrics.update(TestInfoServices.get(node))
 
+    # Apply operation-based timeout scaling (opt-in via config)
+    operation_factor = _get_operation_timeout_factor(node.parent_cluster.params, operation)
+    soft_timeout = soft_timeout * operation_factor if soft_timeout else soft_timeout
```
pehala: Is this the same as `soft_timeout * operation_factor if soft_timeout else None`? Is so I would prefer it, since this is unnecessarily confusing

docs/configuration_options.md
```diff
@@ -2041,6 +2041,15 @@ Store adaptive timeout metrics in Argus. Disabled for performance tests only.
 **type:** bool
 
 
+## **adaptive_timeout_multipliers** / SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS
+
+Optional dict of adaptive-timeout multipliers keyed by operation name (from Operations enum value[0], e.g. decommission, remove_node, new_node, repair, etc.). If the current operation key is absent, multiplier 1.0 is used. Example: {decommission: 4, new_node: 2}.
```
pehala: How would this look in yaml? Is this notation allowed? 
```yaml
adaptive_timeout_multipliers:
  decomission: 4
  new_node: 2
```

sdcm/sct_config.py
soyacz: Could be done, but I'ld prefer third class, that would not require deps - avoiding cycles.
This would require changing `sdcm.utils.adaptive_timeouts.Operations` so we are sure we implement all of them (or redesign it completely).
Maybe in a followup?

pehala: Followup sounds good

sdcm/sct_config.py
aleksbykov: > Why it can be a none? empty dict is better than having none no?

It set as nonw, because be default (test_default.yaml) config parameter adaptive_timeout_multipliers is set to `null`. This is done to preserve old behavior.

pehala: what if you define it as empty dict instead?

sdcm/sct_config.py
aleksbykov: it is some safe from yaml configuration:
adaptive_timeout_multipliers:
  decommission: true
  remove_node: false

because this could be converted by python to 0 and 1.

pehala: I dont think we need it. It would be misuse of the config and we should catch in the review. I also find it unlikely that someone would do this, it seems clearly wrong to me.

sdcm/utils/adaptive_timeouts/__init__.py
aleksbykov: fixed

docs/configuration_options.md
aleksbykov: Both supported.
in yaml files
```yaml
adaptive_timeout_multipliers:
  decomission: 4
  new_node: 2
```
in sct env variables 
` SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS="{'decommission': 2, 'new_node': 3}"`

unit_tests/unit/test_adaptive_timeouts.py
aleksbykov: fixed

unit_tests/unit/test_adaptive_timeouts.py
aleksbykov: Remove duplicating tests

unit_tests/unit/test_adaptive_timeouts.py
aleksbykov: done

sdcm/sct_config.py
aleksbykov: we need str if value passed via SCT env variable

sdcm/sct_config.py
aleksbykov: It could be none, because by default the value is null in test_default.yaml. This will save the old behavior

sdcm/sct_config.py
aleksbykov: added

sdcm/sct_config.py
aleksbykov: if it will be configured with True/False, then the python could convert them to 1/0. so it is just a simple protection

aleksbykov:
@pehala , sorry for network issues, i answered twice on same question.

sdcm/sct_config.py
aleksbykov: removed

sdcm/sct_config.py
aleksbykov: updated

sdcm/sct_config.py
pehala: I think we can supply it via env even if it isnt a string, look at https://github.com/scylladb/scylla-cluster-tests/blob/bbc8359ab1d5a177205e94f023fdda4ac291f04f/sdcm/sct_config.py#L885 which can be passed via https://github.com/scylladb/scylla-cluster-tests/blob/bbc8359ab1d5a177205e94f023fdda4ac291f04f/vars/longevityPipeline.groovy#L213. Also look at the pipeline itself, I believe `SCT_STRESS_IMAGE.cassandra-stress` is exactly the same usecase as this.

sdcm/sct_config.py
soyacz: https://scylladb.atlassian.net/browse/ARGUS-151?atlOrigin=eyJpIjoiNTNmYWIyYTU4ZDE5NDlmMTljNDc2MmE3ZTAxYjJmZjUiLCJwIjoiaiJ9

pehala (dismissed):

soyacz:
@pehala I think we should use `DictOrStrOrPydantic` as you proposed before merging this one

pehala:
> @pehala I think we should use `DictOrStrOrPydantic` as you proposed before merging this one

I think only pydantic is fine as well. I think it will support everything we need it to, but I dont oppose `DictOrStrOrPydantic`

fruch:
> > @pehala I think we should use `DictOrStrOrPydantic` as you proposed before merging this one
> 
> I think only pydantic is fine as well. I think it will support everything we need it to, but I dont oppose `DictOrStrOrPydantic`

if we are sure `SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS` works as expected, no need to change.
but it should be tested checked out

aleksbykov:
@soyacz @pehala can you review?

fruch:
> @soyacz @pehala can you review?

@aleksbykov

small comment like, "Yes i've added more tests to show env variables are working as expected" 

would be really helpful in getting things reviewed/approved

fruch (dismissed):

sdcm/utils/adaptive_timeouts/__init__.py
```diff
@@ -261,6 +275,11 @@ def adaptive_timeout(  # noqa: PLR0914
     if store_metrics:
         load_metrics.update(TestInfoServices.get(node))
 
+    # Apply operation-based timeout scaling (opt-in via config)
+    operation_factor = _get_operation_timeout_factor(node.parent_cluster.params, operation)
+    soft_timeout = soft_timeout * operation_factor if soft_timeout else None
+    hard_timeout = hard_timeout * operation_factor if hard_timeout else None
```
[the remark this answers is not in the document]
soyacz: this looks like a real issue, @aleksbykov can you confirm it's valid?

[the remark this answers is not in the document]
pehala: Do we have cases where hard timeout or soft timeout is zero? I think we use None instead of zero generally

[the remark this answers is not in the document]
aleksbykov: Checking the code, and @pehala is right, that instead of zero hard and soft timeouts use None, when not initialized. Code analysis  show the same.

aleksbykov:
@soyacz @pehala Can you take a look?

soyacz:
@aleksbykov let's use new feature for tests you had issues and wanted that feature

aleksbykov:
> @aleksbykov let's use new feature for tests you had issues and wanted that feature

@soyacz , i added conf file and updated 1 tests which this commit should fix for 2026.2.0. Other tests are not merged yet where i saw this issue , i will update prs accordingly

soyacz (approved):
LGTM

aleksbykov:
@soyacz @fruch @vponomaryov @pehala can we merge it and backport to 2026.2?

vponomaryov (approved):
LGTM

vponomaryov:
@scylladb/qa-maintainers 
Since when it is not allowed to merge PRs having 2 approvals?

cezarmoise (approved):
LGTM

cezarmoise:
>  Since when it is not allowed to merge PRs having 2 approvals?

@vponomaryov There was an old unresolved comment. I think that blocked it.
Hidden in here 
image

vponomaryov:
@aleksbykov 
This PR causes following problem:
```
20:22:18  < t:2026-05-29 17:22:18,502 f:tester.py       l:443  c:LongevityTest        p:INFO  > sct_runner info in Argus TestRun is updated
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR > General Error submitting data to Argus < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR > General Error submitting data to Argus
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR > Traceback (most recent call last):
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >   File "/home/ubuntu/scylla-cluster-tests/sdcm/tester.py", line 444, in init_argus_run
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >     self.test_config.argus_client().sct_submit_config(name="sct_config", content=json.dumps(dict(self.params)))
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >                                                                                  ~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >   File "/usr/local/lib/python3.14/json/__init__.py", line 231, in dumps
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >     return _default_encoder.encode(obj)
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >            ~~~~~~~~~~~~~~~~~~~~~~~^^^^^
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >   File "/usr/local/lib/python3.14/json/encoder.py", line 200, in encode
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >     chunks = self.iterencode(o, _one_shot=True)
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >   File "/usr/local/lib/python3.14/json/encoder.py", line 261, in iterencode
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR >     return _iterencode(o, 0)
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR > ValueError: Circular reference detected
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR > when serializing method object
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR > when serializing sdcm.sct_config.AdaptiveTimeoutMultipliers object
20:22:18  < t:2026-05-29 17:22:18,503 f:tester.py       l:449  c:LongevityTest        p:ERROR > when serializing dict item 'adaptive_timeout_multipliers'
```

pehala:
Lets revert then

aleksbykov:
@vponomaryov , could point the argus job or steps how to reproduce this issue? ( i didn't catch it during testing)
i have a fix and want to check this case before submiting
@pehala what are steps to return this pr back? should i create new one with fix?

pehala:
> @pehala what are steps to return this pr back? should i create new one with fix?

Yes, please mention the bug and what tests you added to ideally catch it from happening again

vponomaryov:
> @vponomaryov , could point the argus job or steps how to reproduce this issue? ( i didn't catch it during testing) i have a fix and want to check this case before submiting @pehala what are steps to return this pr back? should i create new one with fix?

The error is suppressed, because it is Argus-related failure.
You can see it in the next run to the one you added to the PR description:
- [scylla-2026.2/reproducers/lwt-cdc-mv-tablets-spitmerge-test#6](https://argus.scylladb.com/tests/scylla-cluster-tests/646dcfc0-4c77-4f65-8567-e8985e1243cd)

PR #14677, opened 2026-05-15
docs: clarify unit test guidelines to discourage all class-based test patterns

pehala:
Expand the DON'T/DO guidance to explicitly prohibit plain pytest classes (not just unittest.TestCase subclasses), and recommend module-level helper functions instead of class methods for shared setup logic.


fruch (approved):
LGTM

PR #14688, opened 2026-05-18
fix(monkey): Fix cleanup in AbortDecommissionMonkey

cezarmoise:
Ensure that a new node will always be added if aborting fails. 
Only run the nemesis when tablet test keypsaces exist.

Fixes: https://scylladb.atlassian.net/browse/SCT-367

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis/monkey/abort_decommission.py
+++ b/sdcm/nemesis/monkey/abort_decommission.py
@@ -1,15 +1,20 @@
+import logging
+
 from invoke import UnexpectedExit
 
 from sdcm.cluster import NodeCleanedAfterDecommissionAborted, NodeStayInClusterAfterDecommission
-from sdcm.exceptions import UnsupportedNemesis
+from sdcm.exceptions import KillNemesis, UnsupportedNemesis
 from sdcm.nemesis import NemesisBaseClass, target_data_nodes
 from sdcm.remote.libssh2_client.exceptions import UnexpectedExit as Libssh2UnexpectedExit
 from sdcm.sct_events import Severity
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.filters import EventsSeverityChangerFilter
+from sdcm.utils.features import is_tablets_feature_enabled
 from sdcm.utils.tasks import wait_for_tasks
 from sdcm.utils.parallel_object import ParallelObject
 from sdcm.wait import wait_for
+
+logger = logging.getLogger(__name__)
 
 # How long to wait for the node to recover to UN after a successful abort.
 # A node may land in a transitional state (UJ / DN) for up to several minutes
@@ -91,14 +96,26 @@ class AbortDecommissionMonkey(NemesisBaseClass):
                 f"Target node {self.runner.target_node.name} is the only one in rack {self.runner.target_node.rack}, cannot decommission it."
             )
 
+        # Aborting decommission is only supported with tablets, https://github.com/scylladb/scylladb/pull/24129
+        if not is_tablets_feature_enabled(self.runner.target_node):
+            raise UnsupportedNemesis("Aborting decommission is only supported with tablets.")
+
+        if not self.runner.cluster.get_non_system_ks_cf_with_tablets_list(db_node=self.runner.target_node):
+            raise UnsupportedNemesis("No test tables with tablets found.")
+
         # save info of the target node, as it will not be available if decommission succeeds
         target_is_seed = self.runner.target_node.is_seed
         target_name = self.runner.target_node.name
 
-        # Run decommission and abort in parallel, since we want to abort as soon as streaming starts, without waiting for decommission to finish
-        ParallelObject(
-            objects=[self.decommission_target_node, self.abort_decommission_task], timeout=600
-        ).call_objects()
+        try:
+            # Run decommission and abort in parallel, since we want to abort as soon as streaming starts, without waiting for decommission to finish
+            ParallelObject(
+                objects=[self.decommission_target_node, self.abort_decommission_task], timeout=600
+            ).call_objects()
+        except KillNemesis:
+            return
+        except Exception as exc:
+            logger.exception("ParallelObject failed during decommission/abort: %s", exc)
 
         # verify_decommission checks gossip + Raft to determine the actual outcome.
         # It also calls terminate_node() internally when the node has left the ring,
--- a/unit_tests/unit/nemesis/monkey/test_abort_decommission.py
+++ b/unit_tests/unit/nemesis/monkey/test_abort_decommission.py
@@ -66,11 +66,18 @@ def runner(base_runner):
 
 
 @pytest.fixture(autouse=True)
-def sequential_parallel_object():
-    """Replace ParallelObject.call_objects with sequential execution for all tests.
+def mock_tablets():
+    """Patch is_tablets_feature_enabled to return True for all tests."""
+    with patch(f"{_MODULE}.is_tablets_feature_enabled", return_value=True):
+        yield
 
-    Ensures the two callables passed to ParallelObject run deterministically
-    in-process so tests can assert on their side-effects without real threads.
+
+@pytest.fixture(autouse=True)
+def sequential_parallel_object():
+    """Replace ParallelObject.call_objects with deterministic sequential execution.
+
+    Runs the two callables in-process and returns successfully. Individual
+    tests patch this fixture when they need an exceptional ParallelObject path.
     """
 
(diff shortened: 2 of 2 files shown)


sdcm/nemesis/monkey/abort_decommission.py
```diff
@@ -138,3 +108,50 @@ def disrupt(self):
...
+        with ExitStack() as context_manager:
+
+            def finalizer(exc_type, *_):
+                # in case of test end/killed, leave the cleanup alone
+                if exc_type is not KillNemesis:
+                    self.verify(
+                        target_is_seed=self.runner.target_node.is_seed, target_name=self.runner.target_node.name
+                    )
+
+            context_manager.push(finalizer)
```
pehala: I dont understand why we have to use ctx manager here, would try/expect/finally work here as well?

sdcm/nemesis/monkey/abort_decommission.py
```diff
@@ -138,3 +108,50 @@ def disrupt(self):
...
+        - Returns normally: full decommission succeeded despite the abort.
+          verify_decommission already terminated the node.  Add a replacement.
+        """
+        if len([n for n in self.runner.cluster.data_nodes if n.rack == self.runner.target_node.rack]) == 1:
+            raise UnsupportedNemesis(
+                f"Target node {self.runner.target_node.name} is the only one in rack {self.runner.target_node.rack}, cannot decommission it."
+            )
+
+        if not is_tablets_feature_enabled(self.runner.target_node):
+            raise UnsupportedNemesis("Aborting decommission is only supported with tablets.")
```
fruch, line 101: why it's supported only for tablets ? don't see any reference here that explain this

sdcm/nemesis/monkey/abort_decommission.py
fruch: it's kind of align with other nemesis code using this pattern (usually when there more complicated logic, and multiple context managers in use)

pehala: If we dont need it here yet, there is no need to use it. We can always migrate to this pattern when we need it

sdcm/nemesis/monkey/abort_decommission.py
pehala: The decommission abort was implemented in 2026.1 and is specifically targetting tablets, there will never be a vnode implementation.

fruch: sounds kind of similar to `DecommissionStreamingErrMonkey` one... that I remember having for quite some time.

regardless, would be usual to document it, and why it doesn't have or need vnode counterpart.
cause until this PR, it's not mentioned anywhere.

pehala: > sounds kind of similar to `DecommissionStreamingErrMonkey` one... that I remember having for quite some time.

It is not, this is done via new abort command/API, i.e. it is user initiated operation not an error.

> sounds kind of similar to `DecommissionStreamingErrMonkey` one... that I remember having for quite some time.
> 
> regardless, would be usual to document it, and why it doesn't have or need vnode counterpart. cause until this PR, it's not mentioned anywhere.

It is a new feature and was implemented as such in https://github.com/scylladb/scylla-cluster-tests/pull/13333. We can change the text to mention this is Tablets-only feature but I dont think we need to add more doc than this, same as we dont mention it for every new feature we test

fruch: I still argue the "Aborting decommission is only supported with tablets" is misleading text here. if there API it can be used for vnodes.

pehala: > I still argue the "Aborting decommission is only supported with tablets" is misleading text here. if there API it can be used for vnodes.

It cannot be used for vnodes AFAIK, the API is that the task for decomission for tablets is now abortable, I do not think this is case for Vnodes

@cezarmoise Can you confirm/verify?

cezarmoise: Yes. I confirmed with @tgrabiec before, and tried it myself to see it doesn't work.
https://argus.scylladb.com/tests/scylla-cluster-tests/ea1e5e9f-04ab-45f6-a788-a5f0078ac1bf/details

> (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=826025e7-edb4-4261-8cc8-0c72c69e9e92: type=RUNTIME_ERROR regex=std::runtime_error line_number=1221 node=longevity-5gb-1h-AbortDecommissionM-db-node-ea1e5e9f-5
std::runtime_error (request 82ff03c8-5389-11f1-b198-f41bdce7af99 for node c5562bf1-4fa7-407d-8f6b-7558dd6d1a17 is already running and cannot be aborted)

The nodes get decommissioned, it doesn't get aborted

cezarmoise: From design doc:
https://scylladb.atlassian.net/wiki/spaces/RND/pages/67731753/Concurrent+topology+operations#Scope

> The solution applies to clusters where all user data is in tablets, which is the case for the X-Cloud and our long-term goal. 
> We assume that vnode-based keyspaces don’t hold much data, so topology operations which are O(data size in vnodes) are O(1) and it’s ok if bootstrap is blocked behind them. We assume this part is short-enough.

I could link the change (https://github.com/scylladb/scylladb/pull/24129), but it does not explicitly say vnodes are not supported, or that only tablets are supported, it just does not mention vnodes at all.

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/0bf724d28ff8557f6a15d6e60070296d1bb0c663: fixed tests

sdcm/nemesis/monkey/abort_decommission.py
```diff
@@ -91,14 +92,24 @@ def disrupt(self):
...
-        ParallelObject(
-            objects=[self.decommission_target_node, self.abort_decommission_task], timeout=600
-        ).call_objects()
+        try:
+            # Run decommission and abort in parallel, since we want to abort as soon as streaming starts, without waiting for decommission to finish
+            ParallelObject(
+                objects=[self.decommission_target_node, self.abort_decommission_task], timeout=600
+            ).call_objects()
+        except KillNemesis:
+            pass
```
pehala: How does try/except with pass fix the cleanup?

cezarmoise: Nemesis does not exit early now, so `verify_decommission` is always called.

pehala: But that covers only KIllNemesis, was KillNemesis called in https://scylladb.atlassian.net/browse/SCT-367? Since other nemesis continued afterewards I would say that it wasnt

cezarmoise: You are right. The version with ExitStack was correct, but after changing it it's no longer the same!

cezarmoise: Fixed in https://github.com/scylladb/scylla-cluster-tests/commit/2018e45f50722acc735c0ac9f0e65ce019d17926

sdcm/nemesis/monkey/abort_decommission.py
```diff
@@ -91,14 +92,26 @@ def disrupt(self):
...
-        ).call_objects()
+        try:
+            # Run decommission and abort in parallel, since we want to abort as soon as streaming starts, without waiting for decommission to finish
+            ParallelObject(
+                objects=[self.decommission_target_node, self.abort_decommission_task], timeout=600
+            ).call_objects()
+        except KillNemesis:
+            return
+        except:
+            pass
```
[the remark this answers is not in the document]
cezarmoise: Fixed in https://github.com/scylladb/scylla-cluster-tests/commit/e69c69aab46a65e470463413ba56a3a4cfb1587d

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/e69c69aab46a65e470463413ba56a3a4cfb1587d:
- coderabbit findings:
  - fix bare Except
  - add tests for when ParallelObject raises an exception

pehala (dismissed):

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/commit/6f6ff53406b3ec87d0dfa02a791c128247eafd50:
- rebased
- parametrize a test per coderabbit request

pehala (approved):

cezarmoise:
ping @scylladb/storage-qa

jsmolar (approved):

PR #14694, opened 2026-05-18
feature(oci): support OCI backend in the cloud resource usage reports

vponomaryov:
### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/pricing.py
+++ b/sdcm/utils/pricing.py
@@ -299,3 +299,12 @@ class AzurePricing:
             LOGGER.warning("Failed to fetch prices for %s in location: %s", instance_type, region)
             return []
         return [item for item in resp.json()["Items"] if "Windows" not in item["productName"]]
+
+
+class OCIPricing:
+    """OCI pricing - returns 0 as OCI does not have a public retail pricing API like AWS/Azure."""
+
+    def get_instance_price(self, region, instance_type, state, lifecycle):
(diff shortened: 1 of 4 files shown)


vponomaryov:
To try it out yourself run following command:
```
./docker/env/hydra.sh cloud-usage-report --emails <name>@scylladb.com --report-type general
```

sdcm/utils/pricing.py
```diff
@@ -299,3 +299,12 @@ def _get_sku_prices(instance_type: str, region):
             LOGGER.warning("Failed to fetch prices for %s in location: %s", instance_type, region)
             return []
         return [item for item in resp.json()["Items"] if "Windows" not in item["productName"]]
+
+
+class OCIPricing:
```
fruch, line 304: in https://github.com/scylladb/scylla-cluster-tests/pull/14576, pricing would be handle a bit better

fruch: there is an API, but it fetch all the pricing catalog in one call.

we gonna cache the parts we need in SCT code, and update them from time to time.

vponomaryov: Sounds as "not a blocker" for this PR.

fruch: yes, just an ack that we would solve it later with actual data

fruch (approved):

PR #14696, opened 2026-05-19
fix: filter raft topology barrier_and_drain closed_error in node-disrupting nemesis

pehala:
## Summary

Fixes transient `raft topology: exec_global_command(barrier_and_drain) failed with seastar::rpc::closed_error (connection is closed)` false-positive errors that cause test failures in multiple nemesis.

This error is self-resolving (coordinator retries successfully within seconds) and is expected whenever a node is killed/rebooted/gossip-disabled while the topology coordinator holds an RPC connection to it.

### `sdcm/sct_events/setup.py`
- Extend the permanent global `DbEventsFilter` regex to cover both `exec_global_command(barrier)` and `exec_global_command(barrier_and_drain)` variants using `(?:_and_drain)?`

### `sdcm/nemesis/__init__.py`
- **`disrupt_hard_reboot_node`**: upgrade from `ignore_raft_topology_cmd_failing()` to `suppress_expected_unavailability_errors()` for full coverage (also covers raft transport errors, drain rpc, etc.)
- **`disrupt_disable_binary_gossip_execute_major_compaction`**: add `ignore_raft_topology_cmd_failing()` — disabling gossip causes RPC teardown that triggers barrier failures on the topology coordinator (originally reported in SCT-141)
- **`_nodetool_decommission`**: add `ignore_raft_topology_cmd_failing()` — decommission streaming races with `barrier_and_drain` (see SCYLLADB-2099)

## Related

- Closes SCT-141
- Related: SCYLLADB-2099
- Reference fix pattern: https://github.com/scylladb/scylla-cluster-tests/pull/13601

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -805,7 +805,7 @@ class NemesisRunner:
         time.sleep(60)
 
     def disrupt_hard_reboot_node(self):
-        with ignore_raft_topology_cmd_failing():
+        with suppress_expected_unavailability_errors():
             self.reboot_node(target_node=self.target_node, hard=True)
         with self.action_log_scope(f"Wait for {self.target_node.name} node to be fully started"):
             self.target_node.wait_node_fully_start()
@@ -1317,8 +1317,11 @@ class NemesisRunner:
             self.set_target_node(allow_only_last_node_in_rack=True)
 
         target_is_seed = self.target_node.is_seed
-        with self.action_log_scope(
-            f"Decommission {self.target_node.name} node. is_zero_token_node: {self.target_node._is_zero_token_node}"
+        with (
+            suppress_expected_unavailability_errors(),
+            self.action_log_scope(
+                f"Decommission {self.target_node.name} node. is_zero_token_node: {self.target_node._is_zero_token_node}"
+            ),
         ):
             dc_topology_rf_change = self.cluster.decommission(self.target_node)
         new_node = None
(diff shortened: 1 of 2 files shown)


timtimb0t (dismissed):

sdcm/nemesis/__init__.py
```diff
@@ -1317,8 +1317,11 @@ def _nodetool_decommission(self, add_node=True):
             self.set_target_node(allow_only_last_node_in_rack=True)
 
         target_is_seed = self.target_node.is_seed
-        with self.action_log_scope(
-            f"Decommission {self.target_node.name} node. is_zero_token_node: {self.target_node._is_zero_token_node}"
+        with (
+            ignore_raft_topology_cmd_failing(),
```
fruch: on hard reboot `suppress_expected_unavailability_errors` was replacing `ignore_raft_topology_cmd_failing`, so why not use `suppress_expected_unavailability_errors` here ? 
 
isn't is exactly the same cluster of things needs to be ignored ?

pehala: Fixed

sdcm/nemesis/__init__.py
```diff
@@ -5580,7 +5583,10 @@ def disrupt_bootstrap_streaming_error(self):
                     self.cluster.decommission(new_node, timeout=decommission_timeout)
 
     def disrupt_disable_binary_gossip_execute_major_compaction(self):
-        with nodetool_context(node=self.target_node, start_command="disablebinary", end_command="enablebinary"):
+        with (
+            ignore_raft_topology_cmd_failing(),
```
fruch: same question here ? shouldn't we `suppress_expected_unavailability_errors` ?

pehala: I dont think it is needed here, but if you feel like it is, I can add it.

fruch: I don't seen any argument in this commit/PR why this case is different then the others.

if we need to ignore raft issues, cause of unavailable nodes here, it make sense we'll need to ignore all the rest, or I missing what is the definition of `suppress_expected_unavailability_errors` ?

fruch: regardless any of those errors, are warning at best. and I prefer ignoring more of them. then wasting time on log lines development team doesn't care on adjusting.

those log lines won't not gonna surface any impotent bug or issue.

pehala: > regardless any of those errors, are warning at best. and I prefer ignoring more of them. then wasting time on log lines development team doesn't care on adjusting.
> 
> those log lines won't not gonna surface any impotent bug or issue.

I agree, my preference would be to ignore them globally then (in a followup)

fruch (approved):

PR #14697, opened 2026-05-19
fix: remove JsonOutput.toJson() from gce_datacenter and use single-quoted shell exports

juliayakovlev:
JsonOutput.toJson() on a string like '["us-east1","us-east4"]' wraps it in additional JSON string encoding, corrupting the value so Python's ast.literal_eval() cannot parse it as a list. The parameter value is already a valid Python list literal as provided by the user.

Use single-quoted shell exports to preserve the value exactly as-is, preventing both word splitting on spaces and issues with inner double quotes in the list literal.

### Testing

- [x] [muliDC GCE test](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/yulia/job/longevity-cdc-8h-multi-dc-topology-changes-streaming-vnodes-test/). Provisioning failed due to `Machine type with name 'z3-highmem-16-highlssd' does not exist in zone 'us-west1-c'`
- [x] [single DC](https://argus.scylladb.com/tests/scylla-cluster-tests/e755eb69-0e92-4657-ae4f-38a83de2579b) Provisioning failed due to capacity problem
- [x] [artifact test](https://argus.scylladb.com/tests/scylla-cluster-tests/50ceec9f-6b7a-4c36-a53c-7a6f67b5291c)

### Reminders

### DIFF
--- a/vars/artifactsPipeline.groovy
+++ b/vars/artifactsPipeline.groovy
@@ -188,7 +188,7 @@ def call(Map pipelineParams) {
                                                         export SCT_REGION_NAME='${params.region}'
                                                     fi
                                                     if [[ -n "${params.gce_datacenter ? params.gce_datacenter : ''}" ]] ; then
-                                                        export SCT_GCE_DATACENTER=${params.gce_datacenter}
+                                                        export SCT_GCE_DATACENTER='${params.gce_datacenter}'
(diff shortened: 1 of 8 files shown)


fruch (approved):
LGTM

PR #14698, opened 2026-05-19
fix(provision/gce): prevent live migration from disrupting tests

dimakr:
GCE provider default MIGRATE policy allows live-migrating VMs during host maintenance, and it can cause unpredictable latency spikes and performance degradation anomalies that result in test failures.

The fix sets on_host_maintenance=TERMINATE and automatic_restart=False policy values by default when creating GCE instances.
e2 family is kept on MIGRATE (GCE rejects TERMINATE for non-spot e2 VMs).

Fixes: SCT-345

### Testing

- [x] :green_circle: [rolling-upgrade-alternator-test (original issue scenario)](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/upgrades/job/rolling-upgrade-alternator-test/9/)

### Reminders

### DIFF
--- a/sdcm/provision/gce/instance_provider.py
+++ b/sdcm/provision/gce/instance_provider.py
@@ -274,6 +274,13 @@ class VirtualMachineProvider:
             instance.scheduling.on_host_maintenance = "TERMINATE"
             instance.scheduling.provisioning_model = compute_v1.Scheduling.ProvisioningModel.SPOT.name
             instance.scheduling.instance_termination_action = "STOP"
+        elif definition.type.startswith("e2-"):
+            # e2 family supports only on_host_maintenance=MIGRATE for non-spot VMs
+            instance.scheduling.on_host_maintenance = "MIGRATE"
+        else:
+            # avoid live migration and unexpected restarts disrupting tests
+            instance.scheduling.on_host_maintenance = "TERMINATE"
+            instance.scheduling.automatic_restart = False
 
         # Network tags and service accounts
         network_tags = self.network_provider.get_network_tags(allow_public_access=definition.use_public_ip)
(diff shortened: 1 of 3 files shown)


fruch:
@dimakr how about dropping the rolling upgrade specific events filter ? is would be a followup ? or it already done ?

dimakr:
> @dimakr how about dropping the rolling upgrade specific events filter ? is would be a followup ? or it already done ?

yes, wanted to ask about it. But there is a thing - e2 family is kept on MIGRATE (GCE rejects TERMINATE policy for non-spot e2 family VMs), with error:
```
20:02:14  google.api_core.exceptions.BadRequest: 400 POST https://compute.googleapis.com/compute/v1/projects/sct-project-1/zones/us-east1-d/instances: e2 instances do not support onHostMaintenance=TERMINATE unless they are preemptible.
```
And we use the e2-* in our tests. Maybe that filter will be useful for that specific case?

fruch:
> > @dimakr how about dropping the rolling upgrade specific events filter ? is would be a followup ? or it already done ?
> 
> yes, wanted to ask about it. But there is a thing - e2 family is kept on MIGRATE (GCE rejects TERMINATE policy for non-spot e2 family VMs), with error:
> 
> ```
> 20:02:14  google.api_core.exceptions.BadRequest: 400 POST https://compute.googleapis.com/compute/v1/projects/sct-project-1/zones/us-east1-d/instances: e2 instances do not support onHostMaintenance=TERMINATE unless they are preemptible.
> ```
> 
> And we use the e2-* in our tests. Maybe that filter will be useful for that specific case?

still if we leave the critical event, it should be across the board, not just for upgrades.

dimakr:
> still if we leave the critical event, it should be across the board, not just for upgrades.

I will create a task for that, I see that https://scylladb.atlassian.net/browse/SCT-283 and its PR we decided to give filter a try for couple weeks. Now we can follow-up on that.

dimakr:
> I will create a task for thaT

Created https://scylladb.atlassian.net/browse/SCT-385

fruch (approved):

PR #14709, opened 2026-05-20
test(manager): add WORM backup test for GCE

mikliapko:
Closes https://scylladb.atlassian.net/browse/CLOUD-1939

## Summary

Implements SCT test coverage for the immutable/WORM backup feature in Scylla Manager on GCE.

The changes span three commits:

1. **refactor(mgmt/common.py)**: Reorganize file structure for clarity — consolidate constants, group enums/exceptions/data-model classes, add `BackupRetentionLockMode` enum (DISABLED / UNLOCKED / LOCKED).

2. **feature(manager): add lock-mode and retention-days to backup CLI** — Extend `create_backup_task()` in `sdcm/mgmt/cli.py` with `--retention-lock-mode` and `--retention-days` flags required by the WORM backup API.

3. **test(manager): add WORM backup test for GCE** — New `test_worm_backup` sub-test inside `test_backup_feature`:
   - Dynamically creates a GCS bucket with object retention enabled (governance mode)
   - Grants `roles/storage.objectAdmin` to the Manager backup service account
   - Waits for all DB nodes to access the bucket via `check-location` polling
   - Runs a backup task with `retention-lock-mode=unlocked` and `retention-days=1`
   - Verifies that backed-up snapshot files cannot be deleted (`google.api_core.exceptions.Forbidden`)
   - Cleans up the bucket best-effort in a `finally` block

## Test scope

- GCE backend only (guarded by `cluster_backend == "gce"` check in `test_backup_feature`)
- Part of the existing `test_backup_feature` suite

## Testing

- [x] [separate execution of test_worm_backup](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master-clone/job/sct-feature-test-backup-gce/31/)
- [x] [test_worm_backup in scope of test_backup_feature](https://jenkins.scylladb.com/job/scylla-staging/job/mikita/job/manager-master-clone/job/sct-feature-test-backup-gce/36/) (the failure is not related to the added test)

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -20,6 +20,8 @@ import time
 from contextlib import nullcontext
 from datetime import timedelta
 
+from google.api_core.exceptions import Forbidden
+
 from sdcm import mgmt
 from sdcm.argus_results import (
     send_manager_benchmark_results_to_argus,
@@ -34,6 +36,7 @@ from sdcm.mgmt.common import (
     reconfigure_scylla_manager,
     get_persistent_snapshots,
     get_backup_size,
+    BackupRetentionLockMode,
     ObjectStorageUploadMode,
 )
 from sdcm.provision.helpers.certificate import (
@@ -309,6 +312,57 @@ class ManagerBackupTests(ManagerRestoreTests):
                 clean_enospc_on_node(target_node=target_node, sleep_time=30)
         self.log.info("finishing test_enospc_before_restore")
 
+    def test_worm_backup(self):
+        self.log.info("starting test_worm_backup")
+        mgr_cluster = self.db_cluster.get_cluster_manager()
+
+        self.log.info("Create WORM backup bucket")
+        worm_bucket_name = f"manager-worm-backup-test-{self.test_id[:8]}"
+        location = f"gcs:{worm_bucket_name}"
+        worm_bucket = self.create_worm_bucket(region=self.params.gce_datacenters[0], bucket_name=worm_bucket_name)
+
+        try:
+            self.log.info("Wait 30s for backup location accessibility after bucket creation")
+            # We tried to use `scylla-manager-agent check-location` here instead of dummy sleep, but the approach
+            # turned out to be not stable enough - check-location could proceed while the follow-up sctool backup
+            # could fail immediately due to bucket access issue. The reason - check-location command issued in test
+            # initializes rclone from scratch with fresh tokens, while backup runs through the SM agent server may
+            # hold stale tokens. A fixed sleep is the most robust approach here.
+            time.sleep(30)
+
+            self.log.info("Create backup task and wait for its completion")
+            backup_task = mgr_cluster.create_backup_task(
+                location_list=[location],
+                method=self.backup_method,
+                retention_days=1,
+                retention_lock_mode=BackupRetentionLockMode.UNLOCKED,
+            )
+            task_status = backup_task.wait_and_get_final_status(timeout=1500)
+            assert task_status == TaskStatus.DONE, f"Backup task ended in {task_status} instead of {TaskStatus.DONE}"
+
+            snapshot_files = self.get_all_snapshot_files(
+                cluster_id=mgr_cluster.id,
+                bucket_location=worm_bucket_name,
+                only_sstables=False,
+            )
+            assert snapshot_files, "No snapshot files found after backup"
+
+            self.log.info("Try to delete locked backup files")
+            for file_path in snapshot_files:
+                blob = worm_bucket.blob(file_path)
+                try:
+                    blob.delete()
+                    raise AssertionError(f"Deletion of locked file {file_path} unexpectedly succeeded")
+                except Forbidden:
+                    self.log.debug("File %s is properly locked (deletion forbidden)", file_path)
+            self.log.info("All snapshot files are properly protected by object lock")
+
+        finally:
+            self.destroy_worm_bucket(bucket=worm_bucket)
+            mgr_cluster.delete()
+
+        self.log.info("finishing test_worm_backup")
+
     def test_backup_feature(self):
         self.generate_load_and_wait_for_results()
         with self.subTest("Backup Multiple KS' and Tables"):
@@ -319,6 +373,10 @@ class ManagerBackupTests(ManagerRestoreTests):
             self.test_backup_rate_limit()
         with self.subTest("Test Backup Purge Removes Orphans Files"):
             self.test_backup_purge_removes_orphan_files()
+        if self.params.get("cluster_backend") == "gce":
+            # WORM backup feature is currently available for GCP only
+            with self.subTest("Test WORM backup with object lock"):
+                self.test_worm_backup()
         with self.subTest("Test Backup end of space"):  # Preferably at the end
             self.test_enospc_during_backup()
         with self.subTest("Test Restore end of space"):
--- a/sdcm/mgmt/cli.py
+++ b/sdcm/mgmt/cli.py
@@ -720,6 +720,8 @@ class ManagerCluster(ScyllaManagerBase):
         num_retries=None,
         rate_limit_list=None,
         retention=None,
+        retention_days=None,
+        retention_lock_mode=None,
         show_tables=None,
         snapshot_parallel_list=None,
         start_date=None,
@@ -747,6 +749,10 @@ class ManagerCluster(ScyllaManagerBase):
             cmd += " --rate-limit {} ".format(rate_limit_string)
         if retention is not None:
             cmd += " --retention {} ".format(retention)
+        if retention_days is not None:
+            cmd += " --retention-days {} ".format(retention_days)
+        if retention_lock_mode is not None:
+            cmd += " --retention-lock-mode {} ".format(retention_lock_mode)
         if show_tables is not None:
             cmd += " --show-tables {} ".format(show_tables)
         if snapshot_parallel_list is not None:
--- a/sdcm/mgmt/common.py
+++ b/sdcm/mgmt/common.py
@@ -1,7 +1,7 @@
 import logging
 import re
 from datetime import datetime, timedelta
-from enum import Enum
+from enum import Enum, StrEnum
 from typing import Optional, TYPE_CHECKING
 
 import yaml
@@ -52,6 +52,12 @@ class ScyllaManagerError(Exception):
 # ---------------------------------------------------------------------------
 # Enum classes (alphabetical)
 # ---------------------------------------------------------------------------
+
+
+class BackupRetentionLockMode(StrEnum):
+    DISABLED = "disabled"
+    UNLOCKED = "unlocked"
+    LOCKED = "locked"
 
 
 class HostRestStatus(Enum):
--- a/sdcm/mgmt/common.py
+++ b/sdcm/mgmt/common.py
@@ -1,10 +1,10 @@
+import logging
 import re
 from datetime import datetime, timedelta
 from enum import Enum
-import logging
-import yaml
 from typing import Optional, TYPE_CHECKING
 
+import yaml
 from pydantic import BaseModel, ConfigDict
 
 if TYPE_CHECKING:
@@ -12,6 +12,10 @@ if TYPE_CHECKING:
 
 from sdcm.utils.distro import Distro
 from sdcm.utils.version_utils import find_scylla_repo
+
+# ---------------------------------------------------------------------------
+# Constants
+# ---------------------------------------------------------------------------
 
 DEFAULT_TASK_TIMEOUT = 7200  # 2 hours
 LOGGER = logging.getLogger(__name__)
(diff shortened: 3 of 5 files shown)


sdcm/mgmt/operations.py
```diff
@@ -223,6 +225,30 @@ def copy_backup_snapshot_bucket(self, source: str, destination: str, region: str
...
+
+    @staticmethod
+    def destroy_worm_bucket(bucket_name: str) -> None:
+        gce_override_object_retention(bucket_name=bucket_name, path="")
+
+        storage_client, _ = get_gce_storage_client()
+        bucket = storage_client.bucket(bucket_name)
+        blobs = list(storage_client.list_blobs(bucket_name))
+        for blob in blobs:
+            blob.delete()
```
Michal-Leszczynski: nit: is it possible to set `override_unlocked_retention` while deleting the files?
I guess it should be possible to remove the file with `override_unlocked_retention` in a single step instead of first removing the lock with `override_unlocked_retention` and then deleting the file.

mgmt_cli_test.py
```diff
@@ -335,6 +340,63 @@ def test_enospc_before_restore(self):
...
+                )
+                return result.ok and "failed to list: googleapi: Error 403" not in result.stderr.strip()
+
+            for db_node in self.db_cluster.nodes:
+                wait_for(
+                    func=lambda: check_location(db_node),
+                    step=10,
+                    timeout=120,
+                    text=f"Waiting for {db_node.name} to access gs://{worm_bucket_name}",
+                )
```
Michal-Leszczynski: nit: not sure if needed, as the backup task should perform necessary permission checks on its own, but it's fine to do it here as initial test side validation.

mikliapko: The follow-up backup task fails immediately if not waited for access in advance
```
11:57:50  >               raise UnexpectedExit(result)
11:57:50  E               sdcm.remote.libssh2_client.exceptions.UnexpectedExit: Encountered a bad command exit code!
11:57:50  E               
11:57:50  E               Command: 'sudo sctool backup -c fa4f3321-27d8-4984-bda1-3671a0e66ee4 --location gcs:manager-worm-backup-test-ae836549  --retention-days 1  --retention-lock-mode unlocked '
11:57:50  E               
11:57:50  E               Exit code: 1
11:57:50  E               
11:57:50  E               Stdout:
11:57:50  E               
11:57:50  E               
11:57:50  E               
11:57:50  E               Stderr:
11:57:50  E               
11:57:50  E               Error: create backup target: location is not accessible
11:57:50  E                10.138.15.253: agent [HTTP 400] operation put: failed to get bucket: googleapi: Error 403: sct-manager-backup@sct-project-1.iam.gserviceaccount.com does not have storage.objects.list access to the Google Cloud Storage bucket. Permission 'storage.objects.list' denied on resource (or it may not exist)., forbidden - make sure the location is correct and credentials are set, to debug SSH to 10.138.15.253 and run "scylla-manager-agent check-location -L gcs:manager-worm-backup-test-ae836549 --debug"
11:57:50  E               Trace ID: QqBVqeivSeu3sRY-HxAsZw (grep in scylla-manager logs)
```

mgmt_cli_test.py
```diff
@@ -335,6 +340,63 @@ def test_enospc_before_restore(self):
...
+            backup_task = mgr_cluster.create_backup_task(
+                location_list=[location],
+                method=self.backup_method,
+                retention_days=1,
+                retention_lock_mode=BackupRetentionLockMode.UNLOCKED,
+            )
+            task_status = backup_task.wait_and_get_final_status(timeout=1500)
+            assert task_status == TaskStatus.DONE, f"Backup task ended in {task_status} instead of {TaskStatus.DONE}"
+
+            snapshot_files = self.get_all_snapshot_files(cluster_id=mgr_cluster.id, bucket_location=worm_bucket_name)
```
Michal-Leszczynski: nit: IIUC, this gets just the sstables and scylla manifests:
```python
   def _get_all_snapshot_files_gce(cluster_id, bucket_name):
        file_set = set()
        storage_client, _ = get_gce_storage_client()
        blobs = storage_client.list_blobs(bucket_or_name=bucket_name, prefix=f"backup/sst/cluster/{cluster_id}")
        for listing_object in blobs:
            file_set.add(listing_object.name)
        # Unlike S3, if no files match the prefix, no error will occur
        return file_set
```
Retention locks should be applied on all files, so also SM manifest files (`/backup/meta`) and schema files (`/backup/schema`). Would it be easy to verify retention locks for those files as well?

sdcm/utils/gce_utils.py
```diff
@@ -150,6 +150,68 @@ def get_gce_storage_client() -> tuple[storage.Client, dict]:
...
 
+def create_gce_storage_bucket(name: str, region: str, object_lock_enabled: bool = False) -> storage.Bucket:
+    """Create a GCS bucket.
+
+    Args:
+        name: bucket name
+        region: GCS region (e.g., 'us-east1')
+        object_lock_enabled: if True, enables object retention (object lock) on the bucket.
+                             Requires uniform bucket-level access (set automatically).
+                             Must be set at creation time and cannot be changed later.
```
Michal-Leszczynski: question: is a gcp side limitation or test side limitation? I thought that it's possible to enable object locking on an existing bucket in gcp.

mgmt_cli_test.py
mikliapko: Sure, these files would be located in `backup/meta/cluster/{cluster_id}` and `backup/schema/cluster/{cluster_id}` accordingly, right?

sdcm/mgmt/operations.py
mikliapko: Good point, I'll check on that

sdcm/utils/gce_utils.py
mikliapko: I overlooked this one, will remove this sentence from docstring

mgmt_cli_test.py
Michal-Leszczynski: Correct, but I guess to be complete, you can also just list `/backup` entirely - depending on what is better/easier.

mgmt_cli_test.py
Michal-Leszczynski: okay, so it's just to wait for the changes to propagate - makes sense

mgmt_cli_test.py
mikliapko: Make sense. 
We have a dedicated bucket for this test only meaning we won't have any other cluster files there.

Michal-Leszczynski: hmm, so it's up to you - perhaps listing it per cluster ID is indeed less fragile.

sdcm/mgmt/operations.py
mikliapko: So, it looks like GCP REST API actually supports it on delete, but not Python SDK `google-cloud-storage` we use here.
https://docs.cloud.google.com/python/docs/reference/storage/latest/summary_method#google_cloud_storage_blob_Blob_delete_summary

sdcm/mgmt/common.py
```diff
@@ -25,6 +29,183 @@
...
+
+# ---------------------------------------------------------------------------
+# Enum classes (alphabetical)
+# ---------------------------------------------------------------------------
+
+
+class BackupRetentionLockMode(StrEnum):
+    DISABLED = "disabled"
+    UNLOCKED = "unlocked"
+    LOCKED = "locked"
```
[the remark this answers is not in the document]
mikliapko: it's okay, we use the latest 3.14 python

mgmt_cli_test.py
```diff
@@ -309,6 +313,58 @@ def test_enospc_before_restore(self):
...
+                try:
+                    blob.delete()
+                    raise AssertionError(f"Deletion of locked file {file_path} unexpectedly succeeded")
+                except Forbidden:
+                    self.log.debug("File %s is properly locked (deletion forbidden)", file_path)
+            self.log.info("All snapshot files are properly protected by object lock")
+
+        finally:
+            self.destroy_worm_bucket(bucket_name=worm_bucket_name)
+            mgr_cluster.delete()
```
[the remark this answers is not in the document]
mikliapko: In a long term we need a better solution for `mgr_cluster.delete()` across all manager tests, keeping for now since it corresponds the currently existing approach

sdcm/utils/gce_utils.py
mikliapko: fixed

mgmt_cli_test.py
mikliapko: fixed listing per cluster id

mgmt_cli_test.py
mikliapko: replaced with dummy sleep(30) as per discussion with you

sdcm/mgmt/common.py
pehala, line 1: I dont think restructure is correct thing to do here. The unclarity come from us mixing multiple files into one. Enums/Dataclasses should be in __init__, util methods in utils (or there respective util subpackage), finding manager version probably deserves its own small module as well.

This is fine for this PR, but I think what that spoliting this file up is the correct way to approach it in the long term

sdcm/mgmt/cli.py
```diff
@@ -763,6 +767,8 @@ def create_backup_task(  # noqa: PLR0913
             cmd += " --upload-parallel {} ".format(upload_parallel_string)
         if method is not None:
             cmd += " --method {} ".format(method.value)
+        if retention_lock_mode is not None:
```
pehala: nit: lets sort it so all retention params are near each other, same for the parameters

sdcm/mgmt/operations.py
```diff
@@ -223,6 +225,30 @@ def copy_backup_snapshot_bucket(self, source: str, destination: str, region: str
         else:
             raise ValueError(f"Unsupported cluster backend - {cluster_backend}, should be either aws or gce")
 
+    @staticmethod
```
pehala, line 232: No need to use static method, python allows methods in modules without a class as well

sdcm/mgmt/operations.py
```diff
@@ -223,6 +225,30 @@ def copy_backup_snapshot_bucket(self, source: str, destination: str, region: str
...
+        # Grant the access to this bucket for sct-manager-backup service account
+        project_id = KeyStore().get_gcp_credentials()["project_id"]
+        sa_email = f"{GceRegion.SCT_BACKUP_SERVICE_ACCOUNT}@{project_id}.iam.gserviceaccount.com"
+
+        policy = bucket.get_iam_policy(requested_policy_version=3)
+        policy.bindings.append({"role": "roles/storage.objectAdmin", "members": {f"serviceAccount:{sa_email}"}})
+        bucket.set_iam_policy(policy)
+
+    @staticmethod
+    def destroy_worm_bucket(bucket_name: str) -> None:
```
pehala: Could you pass the bucker object directly? I see that in the test you create it regardless

sdcm/mgmt/common.py
mikliapko: Yep, I agree with you, this one is temporal and very quick fix just to make this file usable

sdcm/mgmt/operations.py
mikliapko: I'd suggest to keep it as is. 
The method is on a mixin class and `self.create_worm_bucket()` is used in the test; keeping `@staticmethod` is consistent with the file's pattern.

sdcm/mgmt/operations.py
mikliapko: fixed

sdcm/mgmt/cli.py
mikliapko: fixed

Michal-Leszczynski (approved):

sdcm/mgmt/operations.py
pehala: I think we are overusing static methods here, again, it is more of a suggestion for future than actual request to change it here, as you correctly pointed this is consistent with existing pattern (even if the pattern is not ideal)

sdcm/mgmt/operations.py
```diff
@@ -6,24 +6,30 @@
 from pathlib import Path
 from textwrap import dedent
 from time import sleep
+from typing import TYPE_CHECKING
 
 import boto3
 import yaml
 from invoke import exceptions
 
+if TYPE_CHECKING:
```
pehala: nit: I would do TYPE_CHECKING only after actual imports. It cleaner IMO

sdcm/mgmt/operations.py
```diff
@@ -6,24 +6,30 @@
...
+    from google.cloud import storage as gcs_storage
```
pehala: nit: You actually want to import Bucket only, I see you are only using `gcs_storage.Bucket`

pehala (dismissed):

sdcm/mgmt/operations.py
mikliapko: make sense, fixed

sdcm/mgmt/operations.py
mikliapko: make sense, fixed

pehala (approved):

PR #14718, opened 2026-05-20
fix(ldap): move are_ldap_users_on_scylla runtime state to TestConfig singleton

fruch:
## Summary

Fixes SCT-382: `_init_ldap()` fails with `ValueError: Unknown configuration key 'are_ldap_users_on_scylla'` after Pydantic migration.

## Root Cause

`_init_ldap()` and `_setup_ldap_roles()` in `sdcm/tester.py` were using `self.params["are_ldap_users_on_scylla"]` to track whether LDAP users had been created in Scylla. After the Pydantic migration, `SCTConfiguration.__setitem__` validates key existence and rejects unknown keys — and `are_ldap_users_on_scylla` was removed from the schema.

## Fix

This flag is **runtime state** (has `_setup_ldap_roles()` been called?), not a configuration option. It belongs in the `TestConfig` singleton, not `SCTConfiguration`.

- Add `LDAP_USERS_ON_SCYLLA: bool = False` to `TestConfig`
- Replace `self.params["are_ldap_users_on_scylla"]` writes with `self.test_config.LDAP_USERS_ON_SCYLLA`
- Replace `self.params.get("are_ldap_users_on_scylla")` read in `cluster.py` with `TestConfig().LDAP_USERS_ON_SCYLLA`

## Testing

- [x] 🟡 [rolling-upgrade-ubuntu22.04-ldap-authorization-test #4](https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/oss/job/rolling-upgrade/job/rolling-upgrade-ubuntu22.04-ldap-authorization-test/4/) - now failing on a coredump as the rest of the rolling upgrades cases

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -4504,7 +4504,7 @@ class BaseCluster:
             node.destroy()
 
     def get_db_auth(self):
-        if self.params.get("use_ldap") and self.params.get("are_ldap_users_on_scylla"):
+        if self.params.get("use_ldap") and TestConfig().LDAP_USERS_ON_SCYLLA:
             user = LDAP_USERS[0]
             password = LDAP_PASSWORD
         else:
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -769,7 +769,7 @@ class ClusterTester(unittest.TestCase):
         return None
 
     def _init_ldap(self):
-        self.params["are_ldap_users_on_scylla"] = False
+        self.test_config.LDAP_USERS_ON_SCYLLA = False
 
         match self.params.get("ldap_server_type"):
             case LdapServerType.MS_AD:
@@ -795,7 +795,7 @@ class ClusterTester(unittest.TestCase):
         for user in LDAP_USERS:
             node.run_cqlsh(f"CREATE ROLE '{user}' WITH login=true")
         node.run_cqlsh(f"ALTER ROLE '{LDAP_USERS[0]}' with SUPERUSER=true and password='{LDAP_PASSWORD}'")
-        self.params["are_ldap_users_on_scylla"] = True
+        self.test_config.LDAP_USERS_ON_SCYLLA = True
 
     def configure_ldap(self, node, use_ssl=False):
         self.test_config.configure_ldap(node=node, use_ssl=use_ssl)
--- a/sdcm/test_config.py
+++ b/sdcm/test_config.py
@@ -57,6 +57,7 @@ class TestConfig(metaclass=Singleton):
     SYSLOGNG_ADDRESS = None
     VECTOR_ADDRESS = None
     LDAP_ADDRESS = None
+    LDAP_USERS_ON_SCYLLA: bool = False
     DECODING_QUEUE = None
 
     _test_id = None


sdcm/tester.py
```diff
@@ -3997,7 +3997,7 @@ def tearDown(self):
 
         # NOTE: running on K8S we need to gather logs otherwise a lot of
         # debugging info is lost
-        if self.k8s_clusters:
+        if getattr(self, "k8s_clusters", None):
```
soyacz: what are these unrelated changes?

fruch: opus didn't liked that, also how `()`  are used in some places, I think it's 1 million time I see it change those

fruch: removed them

sdcm/tester.py
```diff
@@ -769,7 +769,7 @@ def _init_data_validation(self):
         return None
 
     def _init_ldap(self):
-        self.params["are_ldap_users_on_scylla"] = False
+        self.test_config.LDAP_USERS_ON_SCYLLA = False
```
pehala, line 772: Why does it have to be global? cannot it be within tester only?

sdcm/cluster.py
```diff
@@ -4504,7 +4504,7 @@ def terminate_node(self, node: BaseNode, scylla_shards: int = 0) -> None:
             node.destroy()
 
     def get_db_auth(self):
-        if self.params.get("use_ldap") and self.params.get("are_ldap_users_on_scylla"):
+        if self.params.get("use_ldap") and TestConfig.LDAP_USERS_ON_SCYLLA:
```
dimakr: Looks like this one will always be `False` as here it reads class attr, which is set sdcm/test_config.py.

The value that is set in test.py (https://github.com/scylladb/scylla-cluster-tests/pull/14718/changes#diff-0a4abb09b1a9c6fe1c09a8b43229bbe1821b5e144b1abb746cce648fe45f7a4aR772) will not be read here as that one is set for instance attr of the tester object, not the `TestConfig.LDAP_USERS_ON_SCYLLA` class attr.

sdcm/tester.py
dimakr: After this is removed the statement `Also guard tearDown() k8s_clusters access with getattr() to prevent AttributeError when setUp() fails before k8s_clusters is initialized (secondary bug from same issue)` in the PR description is stale

sdcm/tester.py
fruch: cause the usage of it isn't in Tester, if you think it's that important go ahead and fix it in a followup.

fruch: or find who should own  the LDAP implementation and clear it with him.

fruch:
Addressed review comments:

- **@dimakr** — Good catch on class vs instance attribute. Fixed: `cluster.py` now uses `TestConfig().LDAP_USERS_ON_SCYLLA` (reads from the singleton instance, same object that tester sets the attribute on).
- **@dimakr** — Removed stale k8s mention from PR description.
- **@soyacz** — Unrelated changes (k8s getattr, exception parentheses) have been removed.

PR #14733, opened 2026-05-21
fix(test-cases): exclude ModifyTable nemesis where the test pins table settings

jsmolar:
Some tests exist to verify a specific table setting, for example TWCS
compaction with a TTL, ZSTD compression, or a given gc_grace_seconds.
ModifyTable* nemeses change exactly these properties on a random table
and never restore them. When one of them runs in such a test, the test
still passes but no longer tests what it was designed for: the TWCS test
ran without TWCS, the multi-DC schema changes test ran without ZSTD, and
the alternator TTL tests validated expiry with a rewritten gc_grace_seconds.

Exclude the conflicting nemeses by name via nemesis_selector in the
affected test cases:

- longevity-twcs-3h, longevity-twcs-48h and the operator twcs overlay:
  compaction, TTL, TWCS window size (and compression on 48h)
- longevity-multidc-parallel-topology-schema-changes-12h: compaction,
  compression
- alternator-ttl sisyphus tests: gc_grace_seconds, TTL
- longevity-100gb-4h-stcs, longevity-50-percent-utilization-4h,
  gemini-3h-ics-cdc-with-nemesis: compaction, TWCS window size

Document in docs/nemesis.md, AGENTS.md and the writing-nemesis skill that
a test config which depends on a specific setting must turn off the
nemeses that change it, so future test authors and AI agents apply the
same rule.
    
Fixes https://scylladb.atlassian.net/browse/SCT-100

### Testing

- [ ] No testing needed

### Reminders

### DIFF
--- a/sdcm/nemesis/monkey/modify_table.py
+++ b/sdcm/nemesis/monkey/modify_table.py
@@ -7,6 +7,7 @@ the abstract base class ``ModifyTableBaseMonkey``.
 
 import abc
 import logging
+import re
 import time
 from typing import Any, Dict
 
@@ -23,6 +24,20 @@ from test_lib.compaction import (
 
 LOGGER = logging.getLogger(__name__)
 
+# Matches CREATE TABLE [IF NOT EXISTS] ks.table — used to identify which CQL
+# statements in pre_create_keyspace
+_RE_CREATE_TABLE = re.compile(
+    r"CREATE\s+TABLE\s+(?:IF\s+NOT\s+EXISTS\s+)?(\w+\.\w+)",
+    re.IGNORECASE,
+)
+
+# Matches ALTER TABLE ks.table — used to identify which CQL statements in
+# post_prepare_cql_cmds set explicit properties.
+_RE_ALTER_TABLE = re.compile(
+    r"ALTER\s+TABLE\s+(\w+\.\w+)",
+    re.IGNORECASE,
+)
+
 
 class ModifyTableBaseMonkey(NemesisBaseClass, abc.ABC):
     """
@@ -32,6 +47,42 @@ class ModifyTableBaseMonkey(NemesisBaseClass, abc.ABC):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.random = self.runner.random
+
+    def _get_tables_with_explicit_property(self, property_name: str) -> set[str]:
+        """Return ks.table names where ``property_name`` is explicitly set via test configuration.
+
+        Scans the following sources:
+        1. ``pre_create_keyspace`` — CREATE TABLE statements with explicit properties.
+        2. ``post_prepare_cql_cmds`` — ALTER TABLE statements that set properties after data load.
+        3. ``compaction_strategy`` param — when set, protects ``keyspace1.standard1`` from
+           compaction changes (this is the table created by ``_pre_create_schema()``).
+        """
+        tables = set()
+        pattern = re.compile(rf"\b{property_name}\s*=", re.IGNORECASE)
+
+        # 1. pre_create_keyspace — CREATE TABLE with explicit properties
+        pre_create = self.runner.cluster.params.get("pre_create_keyspace") or []
+        if isinstance(pre_create, str):
+            pre_create = [pre_create]
+        for cql in pre_create:
+            match = _RE_CREATE_TABLE.search(cql)
+            if match and pattern.search(cql):
+                tables.add(match.group(1))
+
+        # 2. post_prepare_cql_cmds — ALTER TABLE with explicit properties
+        post_cmds = self.runner.cluster.params.get("post_prepare_cql_cmds") or []
+        if isinstance(post_cmds, str):
+            post_cmds = [post_cmds]
+        for cql in post_cmds:
+            match = _RE_ALTER_TABLE.search(cql)
+            if match and pattern.search(cql):
+                tables.add(match.group(1))
+
+        # 3. compaction_strategy param — protects the pre-created keyspace1.standard1 table
+        if property_name == "compaction" and self.runner.cluster.params.get("compaction_strategy"):
+            tables.add("keyspace1.standard1")
+
+        return tables
 
     def modify_table_property(
         self,
@@ -56,12 +107,23 @@ class ModifyTableBaseMonkey(NemesisBaseClass, abc.ABC):
                 filter_out_mv=True,
             )  # not allowed to modify MV
 
-            keyspace_table = self.random.choice(ks_cfs) if ks_cfs else ks_cfs
+            if not ks_cfs:
+                raise UnsupportedNemesis(
+                    "Non-system keyspace and table are not found. ModifyTableProperties nemesis can't be run"
+                )
 
-        if not keyspace_table:
-            raise UnsupportedNemesis(
-                "Non-system keyspace and table are not found. ModifyTableProperties nemesis can't be run"
-            )
+            # Remove tables where this property is explicitly set in test configuration
+            # so that intentional user settings are not silently overwritten.
+            explicit_tables = self._get_tables_with_explicit_property(name)
+            if explicit_tables:
+                ks_cfs = [table for table in ks_cfs if table not in explicit_tables]
+                if not ks_cfs:
+                    raise UnsupportedNemesis(
+                        f"All available tables have '{name}' explicitly set in test configuration. "
+                        f"Skipping to preserve intentional settings."
+                    )
+
+            keyspace_table = self.random.choice(ks_cfs)
 
         cmd = "ALTER TABLE {keyspace_table} WITH {name} = {val};".format(
             keyspace_table=keyspace_table, name=name, val=val
@@ -298,6 +360,17 @@ class ModifyTableDefaultTimeToLiveMonkey(ModifyTableBaseMonkey):
 
         if not ks_cfs:
             raise UnsupportedNemesis("No non-system user tables found")
+
+        # Filter out tables where default_time_to_live is explicitly set in test configuration
+        # so that intentional TTL settings are not silently overwritten.
+        explicit_tables = self._get_tables_with_explicit_property("default_time_to_live")
+        if explicit_tables:
+            ks_cfs = [t for t in ks_cfs if t not in explicit_tables]
+            if not ks_cfs:
+                raise UnsupportedNemesis(
+                    "All available tables have 'default_time_to_live' explicitly set in test configuration. "
+                    "Skipping to preserve intentional settings."
+                )
 
         keyspace_table = self.random.choice(ks_cfs)
         keyspace, table = keyspace_table.split(".")
--- a/unit_tests/unit/nemesis/monkey/test_modify_table.py
+++ b/unit_tests/unit/nemesis/monkey/test_modify_table.py
@@ -13,6 +13,7 @@ from sdcm.exceptions import UnsupportedNemesis
 from sdcm.nemesis.monkey.modify_table import (
     ModifyTableCommentMonkey,
     ModifyTableCompactionMonkey,
+    ModifyTableCompressionMonkey,
     ModifyTableDefaultTimeToLiveMonkey,
     ModifyTableTwcsWindowSizeMonkey,
 )
@@ -31,7 +32,8 @@ pytestmark = pytest.mark.usefixtures("events")
 @pytest.fixture()
 def runner(base_runner):
     """``base_runner`` with a single non-system table for modify-table tests."""
-    base_runner.cluster.get_non_system_ks_cf_list.return_value = ["ks1.tbl1"]
+    base_runner.cluster.get_non_system_ks_cf_list.return_value = ["keyspace1.unprotected"]
+    base_runner.cluster.params.get.return_value = None
     return base_runner
 
 
@@ -40,7 +42,7 @@ def twcs_runner(runner):
     """TestRunner pre-configured for ModifyTableTwcsWindowSizeMonkey tests."""
     runner.cluster.get_all_tables_with_twcs.return_value = [
         {
-            "name": "ks1.tbl1",
+            "name": "keyspace1.unprotected",
             "compaction": {
                 "class": "TimeWindowCompactionStrategy",
                 "compaction_window_unit": "HOURS",
@@ -53,6 +55,24 @@ def twcs_runner(runner):
     runner.random.randint.return_value = 5
     runner.tester.params = {"test_duration": 0}
     return runner
+
+
+@pytest.fixture()
+def compression_monkey(runner):
+    """Generic monkey exercising the shared modify_table_property() path."""
+    return ModifyTableCompressionMonkey(runner)
+
+
+@pytest.fixture()
+def compaction_monkey(runner):
+    """Monkey for testing compaction-specific protection."""
+    return ModifyTableCompactionMonkey(runner)
+
+
+@pytest.fixture()
+def ttl_monkey(runner):
+    """Monkey with its own table selection path."""
+    return ModifyTableDefaultTimeToLiveMonkey(runner)
 
 
 # ---------------------------------------------------------------------------
@@ -100,7 +120,7 @@ def test_comment_monkey(runner):
         monkey.disrupt()
 
     mock_gen.assert_called_once_with(24)
-    assert monkey.runner.executed[-1] == "ALTER TABLE ks1.tbl1 WITH comment = 'abc123';"
+    assert monkey.runner.executed[-1] == "ALTER TABLE keyspace1.unprotected WITH comment = 'abc123';"
 
 
 def test_compaction_twcs_sets_ttl_before_compaction(runner):
@@ -123,7 +143,7 @@ def test_default_ttl_non_twcs(runner):
         monkey = ModifyTableDefaultTimeToLiveMonkey(runner)
         monkey.disrupt()
 
-    assert monkey.runner.executed[-1] == "ALTER TABLE ks1.tbl1 WITH default_time_to_live = 4300000;"
+    assert monkey.runner.executed[-1] == "ALTER TABLE keyspace1.unprotected WITH default_time_to_live = 4300000;"
 
 
 def test_default_ttl_twcs(runner):
@@ -143,7 +163,7 @@ def test_default_ttl_twcs(runner):
         monkey = ModifyTableDefaultTimeToLiveMonkey(runner)
         monkey.disrupt()
 
-    assert monkey.runner.executed[-1] == "ALTER TABLE ks1.tbl1 WITH default_time_to_live = 2000000;"
+    assert monkey.runner.executed[-1] == "ALTER TABLE keyspace1.unprotected WITH default_time_to_live = 2000000;"
     mock_calc_ttl.assert_called_once()
 
 
@@ -181,7 +201,7 @@ def test_twcs_disrupt(twcs_runner):
     monkey.runner.target_node.run_nodetool.assert_called()
     call_args = monkey.runner.target_node.run_nodetool.call_args
     assert call_args[0][0] == "compact"
-    assert call_args[1]["args"] == "ks1 tbl1"
+    assert call_args[1]["args"] == "keyspace1 unprotected"
 
     # waits for schema agreement after ALTER statements
     monkey.runner.cluster.wait_for_schema_agreement.assert_called_once()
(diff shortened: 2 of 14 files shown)


configurations/nemesis/ModifyTableCompressionMonkey.yaml
```diff
@@ -1,3 +1,3 @@
-nemesis_class_name: "SisyphusMonkey"
+nemesis_class_name: ["SisyphusMonkey", "SisyphusMonkey"]
```
pehala: huh, why?

pehala:
@jsmolar This is not a draft, yet you did not ask anyone to review and you have precommit failing

unit_tests/unit/nemesis/monkey/test_modify_table.py
```diff
@@ -277,3 +279,131 @@ def test_twcs_settings(runner, randint_return, unit, initial_size, expected):
...
+    "AND compression = {'chunk_length_in_kb': '64', "
+    "'sstable_compression': 'org.apache.cassandra.io.compress.ZstdCompressor'};"
+)
+
+_GC_CREATE = "CREATE TABLE ks2.tbl2 (id int PRIMARY KEY) WITH gc_grace_seconds = 3600;"
+
+
+def test_protected_tables_empty_when_no_pre_create_keyspace(runner):
+    """No pre_create_keyspace → no protected tables for any property."""
+    runner.cluster.params.get.return_value = None
```
pehala: why?

unit_tests/unit/nemesis/monkey/test_modify_table.py
```diff
@@ -277,3 +279,131 @@ def test_twcs_settings(runner, randint_return, unit, initial_size, expected):
...
+    monkey = ModifyTableCommentMonkey(runner)
```
pehala: Use fixture to set it up if you use this exact line in every test

unit_tests/unit/nemesis/monkey/test_modify_table.py
```diff
@@ -277,3 +279,131 @@ def test_twcs_settings(runner, randint_return, unit, initial_size, expected):
     assert result == expected
     # invariant: gc is always half of dttl
     assert result["gc"] == result["dttl"] // 2
+
+
+# ---------------------------------------------------------------------------
+# Tests for _table_protected_properties and the skip-if-explicitly-configured logic
+# ---------------------------------------------------------------------------
+
+_TTL_CREATE = "CREATE TABLE ks3.tbl3 (id int PRIMARY KEY) WITH default_time_to_live = 86400;"
```
pehala: no need for protected naming here, you can drop the _

sdcm/nemesis/monkey/modify_table.py
```diff
@@ -33,6 +41,21 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.random = self.runner.random
 
+    def _get_protected_tables(self, name: str) -> set[str]:
+        """Return ks.table names where ``name`` is explicitly set in pre_create_keyspace."""
+        pre_create = self.runner.cluster.params.get("pre_create_keyspace") or []
+        if isinstance(pre_create, str):
+            pre_create = [pre_create]
+        pattern = re.compile(rf"\b{name}\s*=", re.IGNORECASE)
```
pehala: I would extract the pattern

sdcm/nemesis/monkey/modify_table.py
```diff
@@ -33,6 +41,21 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.random = self.runner.random
 
+    def _get_protected_tables(self, name: str) -> set[str]:
```
pehala: I would change the naming, protected doesnt fit here IMO. Also lets change variable name to property_name to be more explicit. I confused me when I first read it

sdcm/nemesis/monkey/modify_table.py
jsmolar: This is parametrized re. Can't compile it on the module level.

fruch:
@jsmolar if we have specific test case that we don't want to do changes to the tables, let's ignore those nemesis for those test cases, I rather improve the filtering then add out-of-band new configuration 

**edit:** now I see it's not configuration, but assuming from configuration on how tables are created.
tables can get created in gazillion ways (every stress command can do that, and there more code creating it in all bunch of ways, alternator as example, while this is checking for specific CQL commands)

jsmolar:
> if we have specific test case that we don't want to do changes to the tables, let's ignore those nemesis for those test cases, I rather improve the filtering then add out-of-band new configuration
> 
> **edit:** now I see it's not configuration, but assuming from configuration on how tables are created. tables can get created in gazillion ways (every stress command can do that, and there more code creating it in all bunch of ways, alternator as example, while this is checking for specific CQL commands)

So your take on this is just manually (with AI) identify test cases/configurations where certain `ModifyTable` nemeses should be skipped and exclude them from the run. Am I correct?

sdcm/nemesis/monkey/modify_table.py
```diff
@@ -33,6 +48,42 @@ def __init__(self, *args, **kwargs):
...
+        if isinstance(post_cmds, str):
+            post_cmds = [post_cmds]
+        for cql in post_cmds:
+            match = _RE_ALTER_TABLE.search(cql)
+            if match and pattern.search(cql):
+                tables.add(match.group(1))
+
+        # 3. compaction_strategy param — protects the pre-created keyspace1.standard1 table
+        if property_name == "compaction" and self.runner.cluster.params.get("compaction_strategy"):
+            tables.add("keyspace1.standard1")
```
cezarmoise: When schema is pre-created, if `keyspace_num` is set and larger than `1`, more tables will be created, so this case needs to be accounted for.

e.g. `keyspace_num=3` -> `keyspace1.standard1, keyspace2.standard1, keyspace3.standard1`

sdcm/nemesis/monkey/modify_table.py
```diff
@@ -24,15 +24,109 @@
...
         self.random = self.runner.random
 
+    @property
+    def _table_initial_properties(self) -> TableInitialProperties:
+        """Return the registry shared by all ModifyTable monkeys of this runner, creating it if needed."""
+        registry = getattr(self.runner, "_table_initial_properties", None)
+        if registry is None:
+            registry = TableInitialProperties()
+            self.runner._table_initial_properties = registry
+        return registry
```
pehala: Definitely not like this. there is no reason to have it on runner. Have the initial properties on this basemonkey class instead.

unit_tests/unit/nemesis/monkey/test_modify_table.py
```diff
@@ -279,3 +282,186 @@ def test_twcs_settings(runner, randint_return, unit, initial_size, expected):
...
+    assert registry.is_explicitly_configured("ks1.tbl1", ["gc_grace_seconds", "comment"]) is True
+
+
+def test_registry_unseen_table_is_protected():
+    """A table whose pristine values were never captured is conservatively protected."""
+    session = _FakeSession(tables_rows=[], canary_row=_CANARY_ROW)
+    registry = TableInitialProperties()
+    registry.refresh(_make_cluster(session), MagicMock())
+
+    assert registry.is_explicitly_configured("ks9.unknown", ["comment"]) is True
```
pehala: I would create separate test file for registry

unit_tests/unit/nemesis/__init__.py
```diff
@@ -7,6 +7,26 @@
 from sdcm.nemesis.registry import NemesisRegistry
 
 
+class FakeTableInitialProperties:
```
pehala: You can extend and replace the CQL with a mock. IF you use real class with cluster.cql_connection_patient mocked you should be good

pehala (changes requested):

sdcm/nemesis/monkey/modify_table.py
```diff
@@ -24,15 +24,105 @@
...
+    # live on the class to be shared by all ModifyTable monkeys for the whole run.
+    table_initial_properties = TableInitialProperties()
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.random = self.runner.random
 
+    def _filter_out_explicitly_configured(self, ks_cfs: list[str], properties: Sequence[str]) -> list[str]:
+        """Drop tables where any of ``properties`` was explicitly configured by the test."""
+        self.table_initial_properties.refresh(self.runner.cluster, self.runner.target_node)
```
pehala: If this can run in parallel, wont it be a provlem since the object is shared across instances?

sdcm/nemesis/monkey/modify_table.py
```diff
@@ -24,15 +24,105 @@
...
+                self.defaults = self._fetch_defaults(session)
+            for row in session.execute("SELECT * FROM system_schema.tables"):
+                row_dict = row._asdict()
+                self.initial.setdefault(f"{row_dict['keyspace_name']}.{row_dict['table_name']}", row_dict)
+
+    def _fetch_defaults(self, session) -> dict:
+        """Create a canary table with no explicit options, read its schema row, then delete it."""
+        session.execute(
+            f"CREATE KEYSPACE IF NOT EXISTS {self.CANARY_KEYSPACE} "
+            "WITH replication = {'class': 'NetworkTopologyStrategy', 'replication_factor': 1}"
```
cezarmoise: If `rf_rack_valid_keyspaces` is enabled, this fails.
Should be RF=racks to be safe.

jsmolar: Fixed it.

cezarmoise:
This could be an issue so please get another confirmation first.
I looks like any table created by cassandra-stress that didn't get passed something in `-schema compression(...)`, will end ut with `compression = {}`. But Scylla's default is `{'sstable_compression': 'org.apache.cassandra.io.compress.LZ4Compressor'}`.
So `ModifyTableCompressionMonkey` will never run on anything created by cassandra-stress (so most longevities).

<details><summary>Details</summary>
<p>


## cassandra-stress side: `WITH compression = {}` is emitted unconditionally

`cassandra-stress/src/java/org/apache/cassandra/stress/settings/SettingsSchema.java`, in `createStandard1StatementCQL3()` (lines 176-181):

```java
//Compression
b.append(") WITH compression = {");
if (compression != null)
    b.append("'sstable_compression' : '").append(compression).append("'");

b.append("}");
```

The `WITH compression = {` and `}` are appended outside the `if` — only the *contents* are conditional. So when the user passes no `-schema compression(...)` option, `standard1` is still created with `WITH compression = {}`. The identical pattern is in `createCounter1StatementCQL3()` (lines 212-217) for `counter1`. Compaction, by contrast, is fully inside `if (compactionStrategy != null)` (line 184) and is omitted when unspecified — which is why the heuristic works fine for compaction but not compression.

## ScyllaDB side: `compression = {}` ≠ no compression clause

All in `scylladb/sstables/compressor.cc`:

1. **No compression clause at all** → the default constructor is used, which is LZ4 (lines 419-421):
```cpp
compression_parameters::compression_parameters()
    : compression_parameters(algorithm::lz4)
{}
```

2. **`compression = {}` (empty map)** → the map constructor's fallthrough sets `none`, i.e. compression disabled (lines 490-495):
```cpp
if (auto v = get_option(SSTABLE_COMPRESSION)) {
    _algorithm = name_to_algorithm(*v);
} else if (!options.empty()) {
    throw exceptions::configuration_exception(...);
} else {
    _algorithm = algorithm::none;
}
```

3. **What lands in `system_schema.tables.compression`** — the column the PR compares — is written via `store_map(m, ckey, "compression", timestamp, table->get_compressor_params().get_options())` at `scylladb/db/schema_tables.cc:1577`, and `get_options()` (compressor.cc:578-582) emits `sstable_compression` only when the algorithm is not `none`:
```cpp
if (_algorithm != algorithm::none) {
    opts.emplace(compression_parameters::SSTABLE_COMPRESSION, algorithm_to_qualified_name(_algorithm));
}
```

So the stored rows are:
- canary/default table → `{'sstable_compression': 'org.apache.cassandra.io.compress.LZ4Compressor'}` (qualified name from `algorithm_to_qualified_name`, compressor.cc:462-479)
- c-s `standard1` → `{}`

`initial.get("compression") != defaults.get("compression")` is therefore true for every default cassandra-stress table, and `ModifyTableCompressionMonkey` will treat `keyspace1.standard1` as explicitly configured in every plain c-s longevity.

</p>
</details>

fruch:
> This could be an issue so please get another confirmation first. I looks like any table created by cassandra-stress that didn't get passed something in `-schema compression(...)`, will end ut with `compression = {}`. But Scylla's default is `{'sstable_compression': 'org.apache.cassandra.io.compress.LZ4Compressor'}`. So `ModifyTableCompressionMonkey` will never run on anything created by cassandra-stress (so most longevities).
> 
please raise it in QATOOLS cassandra-stress, so @CodeLieutenant can look into it from c-s code

fruch:
> > if we have specific test case that we don't want to do changes to the tables, let's ignore those nemesis for those test cases, I rather improve the filtering then add out-of-band new configuration
> > **edit:** now I see it's not configuration, but assuming from configuration on how tables are created. tables can get created in gazillion ways (every stress command can do that, and there more code creating it in all bunch of ways, alternator as example, while this is checking for specific CQL commands)
> 
> So your take on this is just manually (with AI) identify test cases/configurations where certain `ModifyTable` nemeses should be skipped and exclude them from the run. Am I correct?

@jsmolar  more or less, if the table configuration is that important for a test and should be change, this nemesis should run 
I see you change the implementation to change it back into original or otherwise skip "unique" configurations,  I still think explicitly skip this one nemesis on a specific set of test where it's impotent is something that should be done, regardless of the suggestion in this PR, 

not all nemesis can be valid for all cases, something we can identify in run time and skip, and something we should skip upfront, we need to balance between complexity and the change we'll something get the selection wrong (which is easily fixed, once identified)

jsmolar:
> > > if we have specific test case that we don't want to do changes to the tables, let's ignore those nemesis for those test cases, I rather improve the filtering then add out-of-band new configuration
> > > **edit:** now I see it's not configuration, but assuming from configuration on how tables are created. tables can get created in gazillion ways (every stress command can do that, and there more code creating it in all bunch of ways, alternator as example, while this is checking for specific CQL commands)
> > 
> > 
> > So your take on this is just manually (with AI) identify test cases/configurations where certain `ModifyTable` nemeses should be skipped and exclude them from the run. Am I correct?
> 
> @jsmolar more or less, if the table configuration is that important for a test and should be change, this nemesis should run I see you change the implementation to change it back into original or otherwise skip "unique" configurations, I still think explicitly skip this one nemesis on a specific set of test where it's impotent is something that should be done, regardless of the suggestion in this PR,
> 
> not all nemesis can be valid for all cases, something we can identify in run time and skip, and something we should skip upfront, we need to balance between complexity and the change we'll something get the selection wrong (which is easily fixed, once identified)

@fruch Yes, my take is that if a table was intentionally created with a specific configuration that is important to the test, we should not modify that configuration. Anything else should still be free to be targeted by the relevant `ModifyTable` nemesis.

Your take is to have a simple flag that skips all `ModifyTable` nemeses (or a specific `ModifyTable` nemesis). If the flag is set, it means that the test is designed not to have its table configuration modified at all (or that a specific configuration should not be modified). Is that correct?

fruch:
> > > > if we have specific test case that we don't want to do changes to the tables, let's ignore those nemesis for those test cases, I rather improve the filtering then add out-of-band new configuration
> > > > **edit:** now I see it's not configuration, but assuming from configuration on how tables are created. tables can get created in gazillion ways (every stress command can do that, and there more code creating it in all bunch of ways, alternator as example, while this is checking for specific CQL commands)
> > > 
> > > 
> > > So your take on this is just manually (with AI) identify test cases/configurations where certain `ModifyTable` nemeses should be skipped and exclude them from the run. Am I correct?
> > 
> > 
> > @jsmolar more or less, if the table configuration is that important for a test and should be change, this nemesis should run I see you change the implementation to change it back into original or otherwise skip "unique" configurations, I still think explicitly skip this one nemesis on a specific set of test where it's impotent is something that should be done, regardless of the suggestion in this PR,
> > not all nemesis can be valid for all cases, something we can identify in run time and skip, and something we should skip upfront, we need to balance between complexity and the change we'll something get the selection wrong (which is easily fixed, once identified)
> 
> @fruch Yes, my take is that if a table was intentionally created with a specific configuration that is important to the test, we should not modify that configuration. Anything else should still be free to be targeted by the relevant `ModifyTable` nemesis.
> 
> Your take is to have a simple flag that skips all `ModifyTable` nemeses (or a specific `ModifyTable` nemesis). If the flag is set, it means that the test is designed not to have its table configuration modified at all (or that a specific configuration should not be modified). Is that correct?

yes if a test is design in a way it never have it's table configuration changed, that specific table modification nemesis should be deselected.

the assumption is that most cases doesn't have such a problem, and only a very small set of configuration would be added with this.

it can happen with any nemesis where some cases wouldn't play along with it, and this is why we have the nemesis filter/selections

it's a simple way forward, yes, it doesn't guarantee it for any new test case, but we can document it in repo, so agents helping people write tests can pick it up (or on the nemesis itself documentation as well)

doing too much in runtime to "magic"lly fixing the issue, has a price and complexity

jsmolar:
> > > > > if we have specific test case that we don't want to do changes to the tables, let's ignore those nemesis for those test cases, I rather improve the filtering then add out-of-band new configuration
> > > > > **edit:** now I see it's not configuration, but assuming from configuration on how tables are created. tables can get created in gazillion ways (every stress command can do that, and there more code creating it in all bunch of ways, alternator as example, while this is checking for specific CQL commands)
> > > > 
> > > > 
> > > > So your take on this is just manually (with AI) identify test cases/configurations where certain `ModifyTable` nemeses should be skipped and exclude them from the run. Am I correct?
> > > 
> > > 
> > > @jsmolar more or less, if the table configuration is that important for a test and should be change, this nemesis should run I see you change the implementation to change it back into original or otherwise skip "unique" configurations, I still think explicitly skip this one nemesis on a specific set of test where it's impotent is something that should be done, regardless of the suggestion in this PR,
> > > not all nemesis can be valid for all cases, something we can identify in run time and skip, and something we should skip upfront, we need to balance between complexity and the change we'll something get the selection wrong (which is easily fixed, once identified)
> > 
> > 
> > @fruch Yes, my take is that if a table was intentionally created with a specific configuration that is important to the test, we should not modify that configuration. Anything else should still be free to be targeted by the relevant `ModifyTable` nemesis.
> > Your take is to have a simple flag that skips all `ModifyTable` nemeses (or a specific `ModifyTable` nemesis). If the flag is set, it means that the test is designed not to have its table configuration modified at all (or that a specific configuration should not be modified). Is that correct?
> 
> yes if a test is design in a way it never have it's table configuration changed, that specific table modification nemesis should be deselected.
> 
> the assumption is that most cases doesn't have such a problem, and only a very small set of configuration would be added with this.
> 
> it can happen with any nemesis where some cases wouldn't play along with it, and this is why we have the nemesis filter/selections
> 
> it's a simple way forward, yes, it doesn't guarantee it for any new test case, but we can document it in repo, so agents helping people write tests can pick it up (or on the nemesis itself documentation as well)
> 
> doing too much in runtime to "magic"lly fixing the issue, has a price and complexity

That means we don't need any code change. All we need is to add `nemesis_selector: "not ModifyTable..."` to relevant test cases (configs) and just add documentation for humans/AI. Am I correct?

fruch:
> > > > > > if we have specific test case that we don't want to do changes to the tables, let's ignore those nemesis for those test cases, I rather improve the filtering then add out-of-band new configuration
> > > > > > **edit:** now I see it's not configuration, but assuming from configuration on how tables are created. tables can get created in gazillion ways (every stress command can do that, and there more code creating it in all bunch of ways, alternator as example, while this is checking for specific CQL commands)
> > > > > 
> > > > > 
> > > > > So your take on this is just manually (with AI) identify test cases/configurations where certain `ModifyTable` nemeses should be skipped and exclude them from the run. Am I correct?
> > > > 
> > > > 
> > > > @jsmolar more or less, if the table configuration is that important for a test and should be change, this nemesis should run I see you change the implementation to change it back into original or otherwise skip "unique" configurations, I still think explicitly skip this one nemesis on a specific set of test where it's impotent is something that should be done, regardless of the suggestion in this PR,
> > > > not all nemesis can be valid for all cases, something we can identify in run time and skip, and something we should skip upfront, we need to balance between complexity and the change we'll something get the selection wrong (which is easily fixed, once identified)
> > > 
> > > 
> > > @fruch Yes, my take is that if a table was intentionally created with a specific configuration that is important to the test, we should not modify that configuration. Anything else should still be free to be targeted by the relevant `ModifyTable` nemesis.
> > > Your take is to have a simple flag that skips all `ModifyTable` nemeses (or a specific `ModifyTable` nemesis). If the flag is set, it means that the test is designed not to have its table configuration modified at all (or that a specific configuration should not be modified). Is that correct?
> > 
> > yes if a test is design in a way it never have it's table configuration changed, that specific table modification nemesis should be deselected.
> > 
> > the assumption is that most cases doesn't have such a problem, and only a very small set of configuration would be added with this.
> > 
> > it can happen with any nemesis where some cases wouldn't play along with it, and this is why we have the nemesis filter/selections
> > 
> > it's a simple way forward, yes, it doesn't guarantee it for any new test case, but we can document it in repo, so agents helping people write tests can pick it up (or on the nemesis itself documentation as well)
> > 
> > doing too much in runtime to "magic"lly fixing the issue, has a price and complexity
> 
> That means we don't need any code change. All we need is to add `nemesis_selector: "not ModifyTable..."` to relevant test cases (configs) and just add documentation for humans/AI. Am I correct?

Yes, I think that would be straight forward, and least amount of surprises (and no new code, and it's complexity)

If we end up with too many of those, we can again reconsider of better machinery to add to nemeis to help, but first let's try the simplest solution

fruch:
Real[]() to

jsmolar:
@fruch I changed this PR according to your commens.

fruch (approved):
LGTM

PR #14735, opened 2026-05-22
fix(azure): delay vm creation after nic creation

soyacz:
In several tests we could observe first db node provision hanged. My suspection is that NIC creation was not fully propagated and VM fails to properly attach it.
Fix is by adding 5s delay (should be enough, as second db node that is triggered 1 s after first one works ok).

fixes: https://scylladb.atlassian.net/browse/SCT-373

### Testing

- [x] - azure provision test

### Reminders

### DIFF
--- a/sdcm/provision/azure/network_interface_provider.py
+++ b/sdcm/provision/azure/network_interface_provider.py
@@ -12,6 +12,7 @@
 # Copyright (c) 2022 ScyllaDB
 
 import logging
+import time
 from dataclasses import dataclass, field
 
 from typing import Dict, List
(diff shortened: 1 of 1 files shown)


sdcm/provision/azure/network_interface_provider.py
```diff
@@ -85,6 +86,7 @@ def get_or_create(
             LOGGER.info("Provisioned nic %s in the %s resource group", nic.name, self._resource_group_name)
             self._cache[nic_name] = nic
             nics.append(nic)
+        time.sleep(5)  # wait for nic to be fully propagated before returning (SCT-373)
```
dimakr: When `pollers` is empty there is probably nothing to wait for? This wait then can be under `if pollers` condition

soyacz: true, will fix

dimakr (approved):
LGTM

vponomaryov (approved):
LGTM

PR #14710, opened 2026-05-20
feature(sct_event): decrease severity of runtime_error when it is ignored

cezarmoise:
During an operation like restart, a node may log that it ignored a runtime_error, but this is treated as an RUNTIME_ERROR by that regex, so it needs to be downgraded in this case.

Example:
> (DatabaseLogEvent Severity.ERROR) period_type=one-time event_id=d34dab97-f082-49da-8a71-99f572a97fa9: type=RUNTIME_ERROR regex=std::runtime_error line_number=34425 node=elasticity-test-nemesis-master-db-node-ad36b2ce-1
2026-05-16T20:02:22.139 elasticity-test-nemesis-master-db-node-ad36b2ce-1  !INFO | scylla[28793]  [shard 0:strm] rpc - client 10.12.11.40:7000: **ignoring error response**: std::runtime_error: Got stream_blob_cmd::error from peer 848946c4-463d-4af8-a305-bb0d129a02ff

Fixes: SCT-365

### Testing

- [ ]

### PR pre-checks (self review)

- [x] I added the relevant `backport` labels
- [x] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #14736, opened 2026-05-22
fix: presenting severity INFO field

timtimb0t:
fix for https://github.com/scylladb/scylla-cluster-tests/pull/14710
Fixes https://scylladb.atlassian.net/browse/SCT-390

### Reminders

### DIFF
--- a/sdcm/sct_events/__init__.py
+++ b/sdcm/sct_events/__init__.py
@@ -25,6 +25,7 @@ class Severity(enum.Enum):
     WARNING = 2
     ERROR = 3
     CRITICAL = 4
+    INFO = 5
 
 
 @runtime_checkable


timtimb0t:
@fruch @pehala @cezarmoise I'm not sure If i got the initial idea correctly, but if i did we need to present new field

sdcm/sct_events/__init__.py
```diff
@@ -25,6 +25,7 @@ class Severity(enum.Enum):
     WARNING = 2
     ERROR = 3
     CRITICAL = 4
+    INFO = 5
```
pehala, line 28: This would make it more important than critical

cezarmoise (changes requested):
I think the event I introduced should just be NORMAL instead of INFO
I put INFO without checking, assuming that was the name.

pehala (changes requested):
I think the correct fix is to change the severity warning to NORMAL instead of INFO

cezarmoise:
https://github.com/scylladb/scylla-cluster-tests/pull/14737
you can close this one

PR #14745, opened 2026-05-23
fix(stress_thread): handle custom user-profile ops in set_hdr_tags

soyacz:
## Summary
- `set_hdr_tags()` crashed with `ValueError` for cassandra-stress user profile commands with custom operations (e.g. `select_base`, `url_column_update`, `row_delete`) because it only recognized `insert=` and `read=`
- Extracted operation classification into `_classify_user_profile_ops()` which parses the `ops(...)` clause and classifies each operation by keyword matching (same keywords used by downstream `_get_workload_type_by_hdr_tag`)
- Unknown operations default to mixed (WRITE+READ) with a warning, rather than crashing
- Also fixed the `raise ValueError("...%s", cmd)` format-string-as-tuple bug

## Testing
- 24 new unit tests covering: standard ops, the exact failing command from the issue, select-only, update/delete-only, LWT ops, CDC ops, unknown ops fallback, legacy patterns, throttled/unthrottled, and the error case

Fixes: https://github.com/scylladb/scylla-cluster-tests/issues/13401

### DIFF
--- a/sdcm/stress_thread.py
+++ b/sdcm/stress_thread.py
@@ -137,26 +137,55 @@ class CassandraStressThread(DockerBasedStressThread):
         params = get_stress_cmd_params(stress_cmd)
         tag_suffix = "rt" if "fixed threads" in params else "st"
         if "user profile=" in stress_cmd:
-            # Examples:
-            # Write: ops(insert=1)
-            # Read: ops(read=2)
-            # Mixed: ops(insert=1,read=2)
-            # Only standard operations (read and insert) are supported per user profile stress command in this implementation
-            if "insert=" in stress_cmd:
+            has_write, has_read = self._classify_user_profile_ops(stress_cmd)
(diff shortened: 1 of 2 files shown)


soyacz:
this is fixing issue we have when running 2026.2 tests: https://argus.scylladb.com/tests/scylla-cluster-tests/94697248-9e5c-4220-82e4-1e87b4c1864a

fruch:
> this is fixing issue we have when running 2026.2 tests: [argus.scylladb.com/tests/scylla-cluster-tests/94697248-9e5c-4220-82e4-1e87b4c1864a](https://argus.scylladb.com/tests/scylla-cluster-tests/94697248-9e5c-4220-82e4-1e87b4c1864a)

if things are blocking release testing, marks them at least as P1.

fruch (approved):

PR #14747, opened 2026-05-24
fix: handle comma-separated availability_zone in GCE provisioner

juliayakovlev:
`GceAZResolver.resolve()` can set `availability_zone` to a comma-separated string like `'a,b,c'`, which means *spread the nodes over those zones, one rack per zone*. The GCE provisioner used the value as-is to build the zone name (`'us-central1-a,b,c'`), which is invalid, and the GCE cluster ignored racks altogether.

GCE now follows the same contract AWS implements with per-AZ subnets: **the rack selects the zone**, and instances are looked up by the zone they actually live in.

Failed test: https://jenkins.scylladb.com/job/scylla-2026.2/job/vnodes/job/tier1/job/longevity-multidc-schema-topology-changes-12h-test/3/

### Provisioner (`sdcm/provision/gce/provisioner.py`)
- The zone of an instance is derived from its **rack index**, falling back to its node index — never from its position in the batch being provisioned. A node lands in the same zone whether the cluster is created in one call or grown a node at a time, and a retry after a partial failure does not move nodes around.
- One VM provider per zone. Every instance operation — terminate, reboot, tagging, cleanup — is issued against the zone the instance lives in. Previously they all went to the first zone, so cleanup of a multi-AZ cluster silently left the nodes of the other zones running.
- An instance sitting outside the configured zones (runtime AZ fallback relocates one on a capacity error) gets a provider for its own zone, with a warning, rather than having the operation sent to the primary zone.
- Instances are tracked under the name GCE gives them (anything outside `[a-z0-9-]` replaced), so a definition whose name gets rewritten — e.g. a `user_prefix` carrying a dot — still matches the instance created for it.
- `get_or_create_instances()` returns instances in the order they were requested, and raises `ProvisionError` if creation comes back short, instead of returning a list that would shift every pairing a caller makes with its definitions.
- Zone letters are parsed defensively: empty tokens and whitespace (`'a, ,b,'`) are dropped; an empty value still picks a random zone.

### Cluster (`sdcm/cluster_gce.py`)
- `add_nodes` splits the requested nodes into per-rack groups and provisions each group in the zone of its rack, mirroring `cluster_aws.add_nodes`.
- `_create_instances` takes the node indexes to create and hands the rack to the provisioner as the definition's rack index, like `cluster_oci`.
- Node rack labels use the same node-index-based rule as zone selection, so rack and zone stay in sync.
- Pre-provisioned instances are bucketed by the zone they really occupy, and rack N addresses the Nth zone of that DC's provisioner whether or not that zone holds anything — compacting the order would hand a rack the instances of another zone. Single-rack clusters keep finding every instance of the DC regardless of zone.
- Reusing a cluster errors when a rack holds fewer instances than requested, instead of quietly adding fewer nodes than asked for.
- Instance lookup by name covers every zone of the region instead of one zone, and instance deletion uses the instance's own zone.
- AZ fallback is skipped for a multi-AZ configuration: it would replace the provisioner with a single-zone one and collapse the cluster into one zone.

### Zone resolver (`sdcm/provision/gce/zone_resolver.py`)
- `resolve()` now raises `NoValidAvailabilityZoneError` when fewer zones are valid than the configuration asks for. `availability_zone: 'b,c,d'` requests three racks; silently returning `'b,c'` would build a two-rack cluster (`racks_count` is derived from this parameter), running a topology the test never asked for. Dropping a rack is the job's decision, not the resolver's.

### Side effect
GCE joins OCI in getting `rack=None` for initial provisioning, so `add_nodes` derives racks from node indexes, matching how the pre-provisioner placed the VMs. As a result GCE loaders and monitors now get real rack labels instead of always `0`, which is what makes `rack_aware_loader` meaningful on GCE.

Single-AZ and empty-AZ (random selection) behavior is preserved.

### Testing

- [ ] [longevity-multidc-schema-topology-changes-12h](https://argus.scylladb.com/tests/scylla-cluster-tests/1b8d727f-1b10-4d44-a614-994b0ba66dd6) - failed with:
```
google.api_core.exceptions.ServiceUnavailable: 503 SERVICE UNAVAILABLE ZONE_RESOURCE_POOL_EXHAUSTED_WITH_DETAILS:
The zone 'projects/sct-project-1/zones/us-central1-b' does not have enough resources available to fulfill the request.  'NULL:0/NULL:0/NULL:0 (state:STOCKOUT, sub-state:STOCKOUT, resource type:compute)'.
```
- Unit tests: `unit_tests/unit/provisioner/test_gce_provisioner_az_distribution.py` (placement stability, lifecycle routing per zone, return ordering, partial-creation failure, cleanup across zones, name normalization, single/empty AZ), `unit_tests/unit/test_cluster_gce.py`, `unit_tests/unit/test_partial_provision_guard.py`, `unit_tests/unit/provisioner/test_gce_zone_resolver.py`

### Reminders

### DIFF
--- a/sdcm/cluster_gce.py
+++ b/sdcm/cluster_gce.py
@@ -38,6 +38,7 @@ from sdcm.utils.gce_utils import (
     SECONDARY_NIC_ROUTING_SERVICE,
     SECONDARY_NIC_ROUTING_SERVICE_UNIT,
     GceLoggingClient,
+    gce_instance_zone,
     get_gce_compute_disks_client,
     get_alternative_zones,
     wait_for_extended_operation,
@@ -287,7 +288,7 @@ class GCENode(cluster.BaseNode):
 
     @property
     def zone(self):
-        return self._instance.zone.split("/")[-1]
+        return gce_instance_zone(self._instance)
 
     def set_hostname(self):
         self.log.debug("Hostname for node %s left as is", self.name)
@@ -433,7 +434,7 @@ class GCECluster(cluster.BaseCluster):
         # Use provisioner's zones to ensure consistency - provisioners already resolved random_zone() if needed
         self._gce_zone_names: list[str] = [provisioner.availability_zone for provisioner in provisioners]
         # Keep this print out for debugging purposes: validate that zones are correctly set
-        LOGGER.debug("GCE zones used: %s", self._gce_zone_names)
+        LOGGER.debug("GCE zones used: %s", [provisioner.availability_zones for provisioner in provisioners])
         self._gce_n_local_ssd = int(gce_n_local_ssd) if gce_n_local_ssd else 0
         self._add_disks = add_disks
         # the full node prefix will contain unique uuid, so use this for search of existing nodes
@@ -476,20 +477,28 @@ class GCECluster(cluster.BaseCluster):
             return bool(value)
         return False
 
-    def _create_instances(self, count, dc_idx=0, instance_type=None):
+    def _create_instances(self, node_indexes: list[int], dc_idx=0, rack=None, instance_type=None):
+        """Provision instances for the given node indexes, placing them in the zone of `rack`.
+
+        `rack` is handed to the provisioner as the definition's rack index, which selects the zone -
+        the GCE equivalent of picking the AZ subnet on AWS. When it is None the provisioner falls
+        back to the node index, so placement stays keyed on the node rather than on the batch.
+        """
+        count = len(node_indexes)
         region = self._definition_builder.regions[dc_idx]
         pricing_model = PricingModel.SPOT if "spot" in self.instance_provision else PricingModel.ON_DEMAND
         definitions = []
-        for node_index in range(self._node_index + 1, self._node_index + count + 1):
-            definitions.append(
-                self._definition_builder.build_instance_definition(
-                    region=region,
-                    node_type=self.node_type,
-                    index=node_index,
-                    dc_idx=dc_idx,
-                    instance_type=instance_type,
-                )
+        for node_index in node_indexes:
+            definition = self._definition_builder.build_instance_definition(
+                region=region,
+                node_type=self.node_type,
+                index=node_index,
+                dc_idx=dc_idx,
+                instance_type=instance_type,
             )
+            if rack is not None:
+                definition.rack_index = rack
+            definitions.append(definition)
         try:
             return provision_instances_with_fallback(
                 self.provisioners[dc_idx],
@@ -501,6 +510,14 @@ class GCECluster(cluster.BaseCluster):
             if isinstance(exc, ProvisionError) and not _is_zone_exhausted(exc):
                 raise
             if not self.is_az_fallback_enabled:
+                raise
+            if len(self.provisioners[dc_idx].availability_zones) > 1:
+                # Replacing a multi-AZ provisioner with a single-zone one would collapse the
+                # cluster into one zone; there is also no way to tell which zone was exhausted.
+                self.log.warning(
+                    "AZ fallback is not supported for a multi-AZ configuration (%s), not retrying",
+                    self.provisioners[dc_idx].availability_zones,
+                )
                 raise
             if count > 1:
                 self.log.warning(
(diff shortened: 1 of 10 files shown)


sdcm/provision/gce/provisioner.py
```diff
@@ -58,6 +59,9 @@ def __init__(self, test_id: str, region: str, availability_zone: str, network_na
         # This matches the logic in GCECluster.__init__
         if not availability_zone:
             availability_zone = random_zone(region)
+        elif "," in availability_zone:
+            # availability_zone can be comma-separated (e.g., "a,b,c") — pick one randomly
+            availability_zone = random.choice([az.strip() for az in availability_zone.split(",")])
```
fruch: why random the order of AZs ? and not use it as is ?

juliayakovlev: @fruch if a few AZ in the list, do we want to choose the same AZ every time?

fruch: if the configuration says: a,b,c. we should used all 3.

once we have fallbacks, those should be the first to be tried out.

when 3 AZs are defined, it doesn't mean we want to randomly pick one of them,

fruch: we need to align it to how it's used in AWS

sdcm/provision/gce/provisioner.py
```diff
@@ -50,19 +50,32 @@ def __init__(self, test_id: str, region: str, availability_zone: str, network_na
         Args:
             test_id: Test ID for tagging resources
             region: GCE region (e.g., 'us-east1')
-            availability_zone: GCE zone letter (e.g., 'b') or None for random
+            availability_zone: GCE zone letter (e.g., 'b'), comma-separated letters
+                (e.g., 'a,b,c') for fallback, or None/empty for random selection
```
fruch: `a,b,c` are not for fallback...

if test configuration set that, it mean we should be splitting the rack on those AZs, not a fallback..

juliayakovlev: changed a comment

sdcm/provision/gce/provisioner.py
juliayakovlev: changed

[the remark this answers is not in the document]
juliayakovlev: changed

sdcm/provision/gce/provisioner.py
```diff
@@ -193,22 +225,40 @@ def get_or_create_instances(
...
-            )
+            gce_instances = self._create_with_fallback(vms_to_create, pricing_model, user_data_list)
             for gce_instance in gce_instances:
                 vm_instance = self._gce_instance_to_vm_instance(gce_instance)
                 self._cache[vm_instance.name] = vm_instance
                 provisioned.append(vm_instance)
 
         return provisioned
 
+    def _create_with_fallback(self, vms_to_create, pricing_model, user_data_list):
```
fruch: @juliayakovlev implementation of fallback is on a different PR, why is it needed here ?

all we need in this PR, is to make sure if the test is passed with multiple AZs, it uses them all to split the racks to each one.

fruch:
@juliayakovlev 

the provision where no matching AZs, is faling back into using only two AZs, and not 3 as instructed
 
```bash
Zones in us-east1 supporting ['n2-highmem-4', 'e2-standard-2', 'n2-highmem-8']: ['us-east1-b', 'us-east1-c', 'us-east1-d']                                                              
Preferred zones ['us-west1-d'] do not support all required machine types ['n2-highmem-4', 'e2-standard-2', 'n2-highmem-8']                                                              
Zones in us-west1 supporting ['n2-highmem-4', 'e2-standard-2', 'n2-highmem-8']: ['us-west1-b', 'us-west1-c', 'us-west1-a']                                                              
GceAZResolver: availability_zone 'b,c,d' does not support all required machine types ['n2-highmem-4', 'e2-standard-2', 'n2-highmem-8'] in regions ['us-east1', 'us-west1']; replacing wi
th 'b,c'                                                                                                                                                                                
Creating 3 instance(s) in zone us-east1-b
```

it should fail, at least until we would define how az fallback are working

juliayakovlev:
@claude review

fruch:
@codetabbitai full review

sdcm/provision/gce/provisioner.py
```diff
@@ -50,46 +50,66 @@ def __init__(self, test_id: str, region: str, availability_zone: str, network_na
         Args:
             test_id: Test ID for tagging resources
             region: GCE region (e.g., 'us-east1')
-            availability_zone: GCE zone letter (e.g., 'b') or None for random
+            availability_zone: GCE zone letter (e.g., 'b'), or comma-separated
+                letters (e.g., 'a,b,c') to distribute nodes across AZs.
+                If None/empty, a random zone is selected.
```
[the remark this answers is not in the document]
juliayakovlev: fixed

sdcm/provision/gce/provisioner.py
```diff
@@ -50,46 +50,66 @@ def __init__(self, test_id: str, region: str, availability_zone: str, network_na
...
+                self.project_id,
+                zone,
+                test_id,
+                disk_provider,
+                self._network_provider,
+            )
+
+        # Primary zone providers for backward compatibility
+        self._disk_provider = DiskProvider(self.project_id, primary_zone)
+        self._vm_provider = self._vm_providers[primary_zone]
```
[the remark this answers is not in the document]
juliayakovlev: fixed

sdcm/provision/gce/provisioner.py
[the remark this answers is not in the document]
juliayakovlev: fixed

unit_tests/unit/provisioner/test_gce_provisioner_az_distribution.py
```diff
@@ -0,0 +1,173 @@
...
+    provisioner = GceProvisioner(
+        test_id="test-123",
+        region="us-east1",
+        availability_zone="a,b,c",
+        network_name="qa-vpc",
+    )
+
+    definitions = [_make_definition(f"node-{i}") for i in range(6)]
+    provisioner.get_or_create_instances(definitions, PricingModel.ON_DEMAND)
+
```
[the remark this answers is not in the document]
juliayakovlev: fixed

juliayakovlev:
All comments are addressed

fruch:
@coderabbitai full review

fruch:
@juliayakovlev the review from coderabbit is just for reference, for me to compare the output of claude review and it.

should fix only if something major would be found

sdcm/cluster_gce.py
```diff
@@ -680,33 +725,27 @@ def add_nodes(
...
+        rack_distribution = self._rack_distribution(count, instance_rack)
+        self.log.info("rack distribution: %s", {rack_idx: len(indexes) for rack_idx, indexes in rack_distribution})
+
+        placed = []
+        for rack_idx, node_indexes in rack_distribution:
+            instances = self._create_or_find_instances(
+                node_indexes=node_indexes, dc_idx=instance_dc, az_idx=rack_idx, instance_type=instance_type
+            )
+            # Requesting N instances may find more than N leftover pre-provisioned ones; take N
+            placed.extend(zip(node_indexes, instances))
```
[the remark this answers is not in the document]
juliayakovlev: changed

juliayakovlev:
@fruch all comment are addressed

fruch (approved):

fruch:
@scylla-zeus-bot 

Please raise SCT Jira for the monitor failing to restore I gce provision 

https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-14747/24/pipeline-overview/

after that raise a PR for the issue (assuming you figured the root cause)

PR #14757, opened 2026-05-25
fix(multidc-parallel-topology-schema-changes): select only nemesis with supports_high_disk_utilization

cezarmoise:
After 84ab956031011f1234d602cd6b90dbd521718c9c, the test has >50% disk utilization, so any nemesis that removes a node will cause the other node in rack to run out of space.

### Testing

- [ ]

### Reminders

### DIFF
--- a/test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
+++ b/test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
@@ -67,7 +67,7 @@ simulated_racks: 0
 instance_type_db: 'i7i.xlarge'
 
 nemesis_class_name: ['SisyphusMonkey', 'SisyphusMonkey']
-nemesis_selector: ["topology_changes", "schema_changes and not disruptive"]
+nemesis_selector: ["topology_changes and supports_high_disk_utilization", "schema_changes and not disruptive and supports_high_disk_utilization"]
 nemesis_seed: [253, 328]
 nemesis_interval: 10
 nemesis_during_prepare: false


pehala:
> After [84ab956](https://github.com/scylladb/scylla-cluster-tests/commit/84ab956031011f1234d602cd6b90dbd521718c9c), the test has >50% disk utilization, so any nemesis that removes a node will cause the other node in rack to run out of space.

Dead nodes can happen in the field and we advertise much greater utilizations for tablets. Not to mention OOS prevention should keep the node available for reads. This needs more investigation of individual issues, not removal of the nemesis IMO

cezarmoise:
50% utilization is the limit because we usually have 2 nodes per rack, and there are multiple nemesis that remove/decommission one node, in which case the other node in the rack does not have enough space to receive that much.

pehala (approved):

roydahan:
What is the list of nemesis that remain for this test after this change?
And what are the nemesis that are going to be excluded.

We need to make sure we don't cripple the test, it's better to reduce the amount of disk utilization than exclude important nemesis which translates into better test coverage.

cezarmoise:
> What is the list of nemesis that remain for this test after this change? And what are the nemesis that are going to be excluded.
> 
> We need to make sure we don't cripple the test, it's better to reduce the amount of disk utilization than exclude important nemesis which translates into better test coverage.

Only 13 are skipped (AddRemoveMvNemesis, CorruptThenScrubMonkey, CreateIndexNemesis, DecommissionMonkey, DecommissionSeedNode, IsolateNodeWithIptableRuleNemesis, IsolateNodeWithProcessSignalNemesis, KillMVBuildingCoordinator, MgmtBackup, MgmtBackupSpecificKeyspaces, MgmtRestore, StartStopValidationCompaction, TerminateAndRemoveNodeMonkey)
There remain 98.

pehala:
> What is the list of nemesis that remain for this test after this change? And what are the nemesis that are going to be excluded.
> 
> We need to make sure we don't cripple the test, it's better to reduce the amount of disk utilization than exclude important nemesis which translates into better test coverage.

I dont think the reduction is significant. The flag is currently on 13 nemesis, which doesn't seem a lot to me. Most of the time it is on nemesis that remove node first and only then add node. Many of these nemesis could be converted to doing this other way around (or introducing duplicates of these nemesis), which would mean they would be reenabled, which we decide to do as a followups

roydahan:
> Only 13 are skipped (AddRemoveMvNemesis, CorruptThenScrubMonkey, CreateIndexNemesis, DecommissionMonkey, DecommissionSeedNode, IsolateNodeWithIptableRuleNemesis, IsolateNodeWithProcessSignalNemesis, KillMVBuildingCoordinator, MgmtBackup, MgmtBackupSpecificKeyspaces, MgmtRestore, StartStopValidationCompaction, TerminateAndRemoveNodeMonkey)
> There remain 98.

Possibly these are the nemesis with the flag, but it's not the exact list of nemesis that are being selected and deselected (this test choose only topology and schema changes).

pehala:
> Possibly these are the nemesis with the flag, but it's not the exact list of nemesis that are being selected and deselected (this test choose only topology and schema changes).

@cezarmoise please quantify how many would be rejected with these filter, to assess the loss of coverage

cezarmoise:
We would be losing from the `toplogy_changes` part:
`DecommissionMonkey`, `DecommissionSeedNode`, `KillMVBuildingCoordinator`, `IsolateNodeWithIptableRuleNemesis`, `IsolateNodeWithProcessSignalNemesis`, `TerminateAndRemoveNodeMonkey`
And `CreateIndexNemesis` from the `schema_changes` part

Of these, `KillMVBuildingCoordinator` and `CreateIndexNemesis` would be unfairly removed, they wouldn't cause out of space at this level of utilization.

cezarmoise:
Closing this and opened https://github.com/scylladb/scylla-cluster-tests/pull/15070 to reduce the disk usage.
For high disk utilization there is a specialized pipeline.

PR #14777, opened 2026-05-26
fix(apt-keys): make gpg --recv-keys recover from keyserver stalls

dimakr:
`gpg --keyserver-options timeout=10` only caps the dirmngr connect, not the overall HKP transaction. A silent stall can block `remoter.sudo` forever, so the HKP/HTTPS fallback chain (added in PR #14537) never kicks in.

Wrap each gpg call with `timeout 30` and pass `timeout=120` to `remoter.sudo` in fetch_apt_keys as an additional SSH-layer guard.

Ref: SCT-290

### Testing

- [x] provision tests

### Reminders

## Summary by CodeRabbit

* **Bug Fixes**
  * Added execution time limits to GPG key retrieval operations to prevent indefinite hangs during package manager setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/scylladb/scylla-cluster-tests/pull/14777?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/unit_tests/unit/test_fetch_apt_keys.py
+++ b/unit_tests/unit/test_fetch_apt_keys.py
@@ -30,9 +30,10 @@ def _make_result(ok):
 
 def _gpg_recv_call(temp_keyring, keyserver, apt_key):
     return mock.call(
-        f"gpg --homedir /tmp --no-default-keyring --keyring {temp_keyring} --keyserver {keyserver} --keyserver-options timeout=10 --recv-keys {apt_key}",
+        f"timeout 30 gpg --homedir /tmp --no-default-keyring --keyring {temp_keyring} --keyserver {keyserver} --keyserver-options timeout=10 --recv-keys {apt_key}",
         retry=1,
         ignore_status=True,
+        timeout=120,
     )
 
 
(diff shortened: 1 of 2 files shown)


unit_tests/unit/test_fetch_apt_keys.py
```diff
@@ -44,6 +45,7 @@ def _https_fallback_call(temp_keyring, apt_key):
         ),
         retry=1,
         ignore_status=True,
+        timeout=120,
     )
```
pehala, line 49: Do we really copy the code in the test instead of testing real code?

dimakr: Probably yes, looks like the test just retypes the commands strings from original _gpg_recv_call / _https_fallback_call  helpers.
But as this is something preexisting, I was not changing it in this PR.

Do you think we should rework the tests here?

pehala: As a followup, but yeah I think we should replace them as they are not useful at all in the current state

pehala (approved):

fruch (approved):

PR #14778, opened 2026-05-26
feat(argus): add always-on request replay log with two-phase write protocol

CodeLieutenant:
Replace MagicMock fallback with a replay-only ArgusSCTClient that records all API calls to JSONL. The replay log is always active — even when Argus is healthy — ensuring every POST is captured for server-side ingest.

Key changes:
- argus/client/replay_log.py: ReplayLog with two-phase write (pre-flight record flushed to disk before HTTP call, outcome written after)
- argus/client/base.py: integrate replay log into post(); add replay-only mode with _ReplayOnlyResponse stub
- sdcm/test_config.py: replace MagicMock with replay-only client
- sdcm/logcollector.py: register argus_replay_log_*.jsonl for S3 upload
- sdcm/tester.py: heartbeat guard uses client.replay_only attribute
- sdcm/utils/resources_cleanup.py: replace MagicMock with replay-only client
- unit_tests/unit/test_argus_replay_log.py: 24 tests covering two-phase write, thread safety, crash resilience, and log collection

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/dusan/job/longevity-10gb-3h-gce-test/4/

### Reminders

Closes https://scylladb.atlassian.net/browse/SCT-225


CodeLieutenant:
🟢 https://jenkins.scylladb.com/job/scylla-staging/job/dusan/job/longevity-10gb-3h-gce-test/4/ (Email Sending Failed) - There is no argus run since Argus is disabled for this run to test when argus is down what would happen. Log is stored and with SCT PR there is command `download-logs`

```bash
./sct.py investigate download-logs 15bb6cad-1d24-4e81-800e-9c45d6ea7d06 --log-type event
```

soyacz (commented):
## Review findings

### 🔴 Critical

**1. `sdcm/utils/argus.py:75` — parens-less `except` breaks Python 3.10–3.13**
```python
except ValueError, AttributeError, TypeError:   # was: except (ValueError, AttributeError, TypeError):
```
This is PEP 758 syntax, valid **only on Python 3.14+**. `pyproject.toml` declares `requires-python = ">=3.10, <3.15"`. On 3.10–3.13 this is a hard `SyntaxError` at import time — and `sdcm/utils/argus.py` is imported nearly everywhere, so the whole framework fails to load. CI passes only because the runner pins 3.14. The change is gratuitous; the original parenthesized form works on every supported version. Please restore the parentheses (unless dropping <3.14 support is an explicit, separate decision).

### 🟡 Moderate

**2. `test_config.py:375` — lazy `argus_client()` has heavy hidden side effects + thread leak**
`argus_client()` now builds a replay-only client when `_argus_client is None`. That construction calls `cls.logdir()`, which lazily **creates a fresh log directory** (`make_new_logdir`, sets `_SCT_TEST_LOGDIR`), spawns a daemon writer thread, registers an `atexit` handler, and opens a JSONL file. The old `MagicMock()` had zero side effects.

Consequence: any path touching `argus_client()` *before* `init_argus_client()` runs will (a) materialize a logdir as a side effect and (b) orphan that client's writer thread + atexit when `init_argus_client()` later overwrites `_argus_client` with the real client **without `close()`ing the old one**. Consider closing an existing client before replacing it, and/or guarding the lazy path.

### 🟢 Minor

- **Private `_replay_log_only` reached across module boundary** (`tester.py`, `sct.py`): `getattr(client, "_replay_log_only", False)`. The PR description calls it `client.replay_only` — but no such public attribute exists. Add a public `replay_only` property on `ArgusClient` and use it; fix the description.
- **`resources_cleanup.py` import grouping**: `from argus.client.sct.client import ArgusSCTClient` was inserted among `sdcm.*` internal imports. The vendored `argus.*` namespace belongs in the third-party group, before the internal block.
- **Temp-dir replay logs never collected**: `get_argus_client` / `resources_cleanup` fall back to `log_dir=tempfile.gettempdir()`. Logcollector only globs the test logdir, so replay logs in `/tmp` (cleanup/standalone contexts) are never uploaded — the "always captured for ingest" guarantee silently doesn't hold there. Acceptable by design, but worth a comment.
- **Title vs. implementation mismatch**: title says "two-phase write protocol"; the actual code (and the tests' own docstrings) is single-write — one record enqueued in `finally`. Stale title/description.

### ✅ Good
- All 4 client-construction sites correctly updated for the new required `log_dir` arg — no missed call sites.
- Heartbeat-guard swap (MagicMock → `_replay_log_only`) is behavior-preserving.
- Solid parametrized pytest coverage for both `download-logs` and the replay log; clean circular-import handling in `test_replay_log_in_log_entities`.
- `run_id` filename sanitization defends against path traversal, with a test.

**Bottom line:** one blocker (#1 — restore the `except` parentheses or the framework won't import on 3.10–3.13). #2 worth fixing before merge. Rest minor.

_Note: argus/* internal client logic (replay_log.py, base.py mechanics) was not audited — vendored from another repo per request._

CodeLieutenant:
@soyacz 

Passing Tests with latest version of the Argus Client:

🟢 https://jenkins.scylladb.com/job/scylla-staging/job/dusan/job/longevity-10gb-3h-gce-test/7 (Argus Disabled)

CodeLieutenant:
> ## Review findings
> ### 🔴 Critical
> **1. `sdcm/utils/argus.py:75` — parens-less `except` breaks Python 3.10–3.13**
> 
> ```python
> except ValueError, AttributeError, TypeError:   # was: except (ValueError, AttributeError, TypeError):
> ```
> 
> This is PEP 758 syntax, valid **only on Python 3.14+**. `pyproject.toml` declares `requires-python = ">=3.10, <3.15"`. On 3.10–3.13 this is a hard `SyntaxError` at import time — and `sdcm/utils/argus.py` is imported nearly everywhere, so the whole framework fails to load. CI passes only because the runner pins 3.14. The change is gratuitous; the original parenthesized form works on every supported version. Please restore the parentheses (unless dropping <3.14 support is an explicit, separate decision).
> 


Ruff removes it in pre-commit lint and it pre-commit lint fails. Since we are on Python 3.14 on master we should keep it, and when back porting, I'll adjust the stuff for 3.10 where needed (branch-2025.1 and branch-2025.4)

soyacz:
@scylladb/qa-maintainers can someone also look at this one?

soyacz:
@k0machi ^^

vponomaryov:
@CodeLieutenant 

AFAIK, the Argus client code gets copied `as-is` from the Argus github project.
Is it the case here?
If not, these changes must go to the https://github.com/scylladb/argus

argus/client/replay_log.py
```diff
@@ -0,0 +1,249 @@
...
+_instance_counter = itertools.count()
+
+LOGGER = logging.getLogger(__name__)
+
+
+class _CloseSentinel:
+    """Singleton signal placed on the queue to stop the writer thread."""
+
+
+_CLOSE = _CloseSentinel()
```
pehala, line 40: Could you epxlain more why this is needed? It seems likea form of a signal

argus/client/replay_log.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def record(self, response: Any) -> None:
+        """Populate ``success`` / ``error`` from a ``requests.Response``.
+
+        Mirrors :meth:`ArgusClient.check_response` discriminator: only a
+        2xx with a JSON body and ``status == "ok"`` counts as success.
+        A 2xx HTML page (auth proxy, gateway error) or
+        ``{"status": "error", ...}`` is a failure.
+        """
+        if not 199 < response.status_code < 300:
```
pehala, line 79: response.ok() wont work here?

argus/client/replay_log.py
```diff
@@ -0,0 +1,249 @@
...
+                # underlying fs (e.g. some network mounts) rejects fsync.
+                LOGGER.exception("argus replay log: durable sync failed")
+
+        try:
+            with open(self._path, "a", encoding="utf-8") as f:
+                while True:
+                    item = self._queue.get()
+                    if item is _CLOSE:
+                        _sync(f)
+                        return
```
pehala, line 188: This seems like it should be a signal

argus/client/replay_log.py
```diff
@@ -0,0 +1,249 @@
...
+                    self._write_line(f, item)
+                    # Drain whatever else is already waiting before syncing,
+                    # to amortize fsync cost across bursts of requests.
+                    for _ in range(63):
```
pehala, line 192: why 63?

argus/client/replay_log.py
```diff
@@ -0,0 +1,249 @@
...
+    def __exit__(self, exc_type, exc, tb) -> None:
+        self.close()
+
+    @staticmethod
+    def _write_line(f: IO[str], item: dict) -> None:
+        # Compact separators (no spaces) -- ~10-15% smaller than the default.
+        f.write(json.dumps(item, default=str, separators=(",", ":"), ensure_ascii=False))
+        f.write("\n")
+
+    def _writer_loop(self) -> None:
```
pehala, line 173: This method seems very complex considering what it does, shouldnt this be a background thread instead?

sdcm/utils/argus.py
```diff
@@ -71,21 +72,27 @@ def is_uuid(uuid) -> bool:
...
+    # The replay log is always created, so a real log_dir is required. Prefer the
+    # SCT test logdir (so logcollector picks the replay log up) and only fall back
+    # to a temp dir for callers that have neither an explicit log_dir nor a logdir.
     argus_client = ArgusSCTClient(
         run_id=run_id,
         auth_token=creds["token"],
         base_url=creds["baseUrl"],
         extra_headers=creds.get("extra_headers"),
         use_tunnel=use_tunnel,
+        log_dir=log_dir or os.environ.get("_SCT_TEST_LOGDIR") or tempfile.gettempdir(),
```
pehala: Why do we have protected named env variable? Shouldnt this be a config option?

sdcm/test_config.py
```diff
@@ -371,31 +371,63 @@ def set_duration(cls, duration):
...
+        all intended Argus submissions are captured for later replay.
+        """
+        run_id = test_id or cls.test_id() or str(uuid4())
+        return ArgusSCTClient(
+            run_id=run_id,
+            auth_token="",
+            base_url="",
+            log_dir=cls.logdir(),
+            replay_log_only=True,
+        )
```
pehala: It seems weird that it is controlled via a variable, wouldnt subclass be cleaner?

sdcm/tester.py
```diff
@@ -487,7 +486,7 @@ def run_nemesis_hooks(self):
 
     def start_argus_heartbeat_thread(self) -> threading.Event:
         def send_argus_heartbeat(client: ArgusSCTClient, stop_signal: threading.Event):
-            if isinstance(client, unittest.mock.MagicMock):
+            if getattr(client, "_replay_log_only", False):
```
pehala: Anothe reason why subclass would be better, we could simply call has_heartbeat method

sdcm/utils/argus.py
k0machi: Agreed, should be documented and promoted to normal variable

sdcm/utils/resources_cleanup.py
```diff
@@ -80,7 +81,15 @@ def init_argus_client(test_id: str, use_tunnel: bool | None = None):
...
         LOGGER.warning("Unable to initialize Argus: %s", exc.message)
-        argus_client = MagicMock()
+        # Fall back to replay-only mode so all intended API calls are still
+        # captured to the replay log instead of being silently mocked away.
+        argus_client = ArgusSCTClient(
+            run_id=test_id,
+            auth_token="",
+            base_url="",
+            log_dir=os.environ.get("_SCT_TEST_LOGDIR") or tempfile.gettempdir(),
+            replay_log_only=True,
```
k0machi: Something for the future - it would be better to have a method to convert an existing client into a replay log client.

sdcm/test_config.py
k0machi: Subclass or method toggle would be good

sct.py
```diff
@@ -2639,7 +2714,16 @@ def create_argus_test_run():
             branch_name=git_status.get("branch.upstream"),
             sct_config=None,
         )
-        LOGGER.info("Initialized Argus TestRun with test id %s", get_test_config().argus_client().run_id)
+        if getattr(argus_client, "_replay_log_only", False):
```
k0machi: Ditto. Method check would be better.

k0machi:
> @CodeLieutenant
> 
> AFAIK, the Argus client code gets copied `as-is` from the Argus github project. Is it the case here? If not, these changes must go to the https://github.com/scylladb/argus

Yes, only the second commit contains actual changes in this PR.

CodeLieutenant:
> @CodeLieutenant
> 
> AFAIK, the Argus client code gets copied `as-is` from the Argus github project. Is it the case here? If not, these changes must go to the https://github.com/scylladb/argus

There is a command to update argus client that does not involve copying manually

`uv run python utils/update_argus_client.py`. It fetches the latest version and updates everything accordingly.

soyacz:
@CodeLieutenant address/reply to comments that aims sct repo, for things in Argus client change, create followup task with relevant findings.

argus/client/replay_log.py
CodeLieutenant: Miss-type, I wanted to put 64, and just put 63 by accident. No reason why 64, only that is power of 2, and I dont want to call fdatasync for everything every time, tradeoff between perf and durability

argus/client/replay_log.py
CodeLieutenant: It's the close signal, for exiting the Background thread that is doing the writes.

argus/client/replay_log.py
CodeLieutenant: Yes it will

CodeLieutenant:
@soyacz @pehala @k0machi Issue tracking the improvements and redesign of the Replay Log inside Argus Client

https://scylladb.atlassian.net/browse/ARGUS-182

sdcm/sct_config.py
```diff
@@ -3876,6 +3875,12 @@ def _get_target_upgrade_version(self):
             self.scylla_version_upgrade_target = scylla_version
             self.update_argus_with_version(scylla_version, "scylla-server-upgrade-target")
 
+    # Internal SCT_ vars that are NOT config params and must be excluded from the
+    # unknown-variable check.  SCT_TEST_LOGDIR is set by TestConfig.logdir() so
+    # that nested cluster/K8S initialisers can discover the current log directory
+    # without an explicit parameter thread-through.
+    _INTERNAL_SCT_ENV_VARS: frozenset[str] = frozenset({"SCT_TEST_LOGDIR"})
+
```
k0machi: Why not just document it in sct_config.py and promote it?

k0machi (changes requested):

sdcm/sct_config.py
k0machi: Also adding it into test_default.yaml

CodeLieutenant: I' thinking of removing this variable from the code, since changing it will make SCT not upload the replay log, so that is maybe the best moving forward

pehala: I think removal of the var is for the best, we didnt have concept of internal sct env vars for now and it wasnt an issue I dont think we need to start

unit_tests/integration/conftest.py
```diff
@@ -252,7 +252,7 @@ def reload_config_for_test():
     docker_scylla.reload_config = reload_config_for_test
 
     cluster = docker_scylla.parent_cluster
-    os.environ.setdefault("_SCT_TEST_LOGDIR", "/tmp/test_vector_search_logs")
+    os.environ.setdefault("SCT_TEST_LOGDIR", "/tmp/test_vector_search_logs")
```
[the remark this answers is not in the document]
soyacz: this one needs fixing

soyacz:
@CodeLieutenant I looked into https://jenkins.scylladb.com/job/scylla-staging/job/dusan/job/longevity-10gb-3h-gce-test/4/ and this job is not visible in Argus. Why?

CodeLieutenant:
> @CodeLieutenant I looked into https://jenkins.scylladb.com/job/scylla-staging/job/dusan/job/longevity-10gb-3h-gce-test/4/ and this job is not visible in Argus. Why?

I've disabled Argus on couple of runs to test replay log, but i could not upload the replay log until we merged the backend part. And now we need to deploy argus in order to have it in argus

soyacz (dismissed):
LGTM
@pehala we agreed to do the client fixes in a separate task, do you think we can merge this one?

sct.py
```diff
@@ -1785,6 +1723,80 @@ def show_log(test_id, output_format, update_argus: bool):
             LOGGER.error("Error updating logs in argus.", exc_info=True)
 
 
+@investigate.command("download-logs", help="Download logs collected for testrun by test-id")
+@click.argument("test_id")
+@click.option("-d", "--dest-dir", type=click.Path(), default=".", help="Destination directory for downloaded logs")
+@click.option("-t", "--log-type", type=str, multiple=True, help="Filter by log type (can be specified multiple times)")
+@click.option("--list-types", is_flag=True, help="List available log types and exit")
+def download_logs(test_id: str, dest_dir: str, log_type: tuple[str, ...], list_types: bool):
```
pehala, line 1816: Followup: We might want to add integration test for this, given ti is directly interacting with S3

pehala (commented):
> @pehala we agreed to do the client fixes in a separate task, do you think we can merge this one?

I am ok with it, I would like the SCT config issue addressed first before approval. I also think it would be better to introduce the new hydra command only in a followup as it doesnt really fit scope of this PR in my opinion, but I will accept it here if deemed necessary

CodeLieutenant:
@soyacz @k0machi I've removed SCT_LOG_DIR env, and now we rely on log dir passed into the arguments.

Everything is rebased, ready for merging

unit_tests/integration/conftest.py
```diff
@@ -252,7 +252,6 @@ def reload_config_for_test():
     docker_scylla.reload_config = reload_config_for_test
 
     cluster = docker_scylla.parent_cluster
-    os.environ.setdefault("_SCT_TEST_LOGDIR", "/tmp/test_vector_search_logs")
```
k0machi, line 255: Why was this removed? This seems unrelated.

k0machi: It's from this commit: https://github.com/CodeLieutenant/scylla-cluster-tests/commit/d2e3dc7a9941b9596ace9649209fe6d8748fc3ce#diff-2d79d500cc6d1bd0410679768806f47a0a042341af43c2b2e96aa5e1dbfdb7ecR245-R246

CodeLieutenant: So we are using _SCT_TEST_LOGDIR? Why should I not then use it in this PR

k0machi: No, I just don't understand why this variable was referenced previously in the conftest. @pehala Is this maybe something for debugging?

k0machi: Okay, I see it now. This is an env var used to hint subprocesses about where the logdir is, so going back to my previous comment of promoting this variable to config I don't see why it was needed to have circumvention logic inside sct_config where previously this worked fine before. Ergo, I'd say we keep this for now, but maybe someone can later look if we can communicate it to subprocesses better or maybe have it always be available for derivation.

CodeLieutenant: Ive removed all my references of this variable

k0machi (approved):

soyacz (approved):

PR #14789, opened 2026-05-26
ci(claude): enable PR review for all scylladb org members including forks

fruch:
## Summary

- Switch `claude-code-review.yml` trigger from `workflow_dispatch` to `pull_request_target` so Claude automatically reviews every PR, including from forks
- Replace `CLAUDE_CODE_OAUTH_TOKEN` with `ANTHROPIC_API_KEY` in both `claude.yml` and `claude-code-review.yml`
- Add a two-layer org membership gate that restricts the review to scylladb org members and collaborators

## Security design

`pull_request_target` runs with base-repo secrets but the workflow code is always from the base branch — forks can't modify the CI. The known "pwn request" attack vector (checking out and executing fork code) is blocked because checkout is pinned to `base.sha`.

**Membership gate** (two layers per [GitHub Security Lab guidance](https://securitylab.github.com/resources/github-actions-preventing-pwn-requests/)):
1. Fast-path: `author_association` in `OWNER | MEMBER | COLLABORATOR`
2. Fallback: live `GET /orgs/scylladb/members/{login}` API call — handles the [known GitHub bug](https://github.com/actions/github-script/issues/643) where org members occasionally receive `CONTRIBUTOR` association

## Setup required

Add `ANTHROPIC_API_KEY` as a repository secret:
```bash
gh secret set ANTHROPIC_API_KEY --repo scylladb/scylla-cluster-tests
```

## Test plan

- [ ] This PR itself triggers `claude-code-review.yml` via `pull_request_target` — check the Actions tab to confirm the review runs and posts a comment
- [ ] Confirm the membership check step shows `✅` for the PR author
- [ ] Verify Claude posts a review comment on this PR
- [ ] Confirm a PR from a non-org member would be skipped (membership step outputs `is_member=false`)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **Chores**
  * Enabled CI workflow to run on pull-request events, added step-level gating that verifies PR author membership before sensitive steps run, removed the previous fork-repo guard, narrowed workflow permissions, and pinned action versions; migrated the automated review to a new credential input.
* **Documentation**
  * Added FAQ explaining how the automated review is triggered, which branch it runs against, membership gating, and how to re-trigger a review.

[Review Change Stack](https://app.coderabbit.ai/change-stack/scylladb/scylla-cluster-tests/pull/14789?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

PR #14800, opened 2026-05-26
ci(claude): smoke test for code review workflow

fruch:
Empty PR to verify the Claude Code Review workflow fires correctly after #14789, #14795, #14797, #14799.


fruch:
@claude just checking if it's working

PR #14805, opened 2026-05-27
refactor(sct-config): normalize union config types to always return concrete types

pehala:
## Summary

SCT config parameters backed by union validators (`IntOrList`, `StringOrList`, `BooleanOrList`, `DictOrStr`) previously returned either a scalar or a list depending on the input — forcing every consumer to do `isinstance` checks before use. This PR removes all that defensive branching by making each validator always return a single concrete type.

A preparatory commit first narrows 8 parameters that are **never** multi-valued to plain scalar types, eliminating unnecessary union complexity for those params entirely.

## Changes

**Phase 0 — narrow scalar-only params**
- 5 `IntOrList` → `int`: `nemesis_interval`, `nemesis_sequence_sleep_between_ops`, `nemesis_add_node_cnt`, `space_node_threshold`, `nemesis_multiply_factor`
- 3 `BooleanOrList` → `Boolean`: `nemesis_during_prepare`, `nemesis_filter_seeds`, `round_robin`

**Phase 1 — `IntOrList` → always `list[int]`**
- Validator now always returns `list[int] | None`
- Consumers use `sum()` in scalar contexts (e.g. `add_nodes`, K8S node pool `num_nodes`), direct list for multi-DC splits
- Removed ~30 `isinstance` guards across `tester.py`, `nemesis/`, `cluster_k8s/`, `provision/`

**Phase 2 — `StringOrList` → always `list[str]`**
- Validator wraps single-string results in `[value]`
- Flattened nested loops in `sct_config.py` stress validation
- Removed `isinstance` guards in `loader_utils.py` and all stress test files

**Phase 3 — `BooleanOrList` → always `list[bool]`**
- Validator always returns `list[bool] | None`

**Phase 4 — `DictOrStr` → always `dict`**
- Validator rejects non-dict results from `ast.literal_eval` / `yaml.safe_load`

## Testing

- 2041 unit tests pass (`uv run python -m pytest unit_tests/unit/ -x -q -n4`)
- New validator unit tests added in `unit_tests/unit/test_config.py` covering all four types


docs/configuration_options.md
```diff
@@ -40,7 +40,7 @@ a list of config files that would be used
 
 **default:** N/A
 
-**type:** str | list[str]
+**type:** list[str]
```
pehala: The accepted input types stayed the same, but our current doc generation algorithm uses output of the type to use Type, which is incorrect now, given the accepted Input is still str | list[str], just the output changed

pehala:
@scylladb/qa-maintainers Do you feel comfortable with size of the PR, or would you rather I split it by phases?

pehala:
> @scylladb/qa-maintainers Do you feel comfortable with size of the PR, or would you rather I split it by phases?

Given I saw no objection, I will continue here

sdcm/utils/operations_thread.py
```diff
@@ -126,9 +126,7 @@ def __init__(self, thread_params: ThreadParams, thread_name: str = ""):
         self.log = logging.getLogger(self.__class__.__name__)
         self.log.debug("Thread params: %s", thread_params)
         nemesis_seed = self.thread_params.db_cluster.params.get("nemesis_seed")
-        if isinstance(nemesis_seed, list):
-            nemesis_seed = nemesis_seed[0]
-        self.generator = random.Random(int(nemesis_seed)) if nemesis_seed is not None else random.Random()
+        self.generator = random.Random(nemesis_seed[0]) if nemesis_seed is not None else random.Random()
```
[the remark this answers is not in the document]
pehala: Unlikely to happen, would have to have set  nemesis_seed: "", which would be user error IMO. Will fix regardless

sdcm/nemesis/__init__.py
```diff
@@ -4445,11 +4445,8 @@ def _shrink_cluster(self, rack=None, new_nodes: list[BaseNode] | None = None):
         if self._is_it_on_kubernetes():
             initial_db_size = self.tester.params.get("k8s_n_scylla_pods_per_cluster") or initial_db_size
 
-        if isinstance(initial_db_size, int):
-            decommission_nodes_number = min(cur_num_nodes_in_dc - initial_db_size, add_nodes_number)
-        else:
-            initial_db_size_in_dc = initial_db_size[self.target_node.dc_idx]
-            decommission_nodes_number = min(cur_num_nodes_in_dc - initial_db_size_in_dc, add_nodes_number)
+        initial_db_size_in_dc = initial_db_size[self.target_node.dc_idx]
```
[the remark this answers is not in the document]
pehala: This is a valid bug, but suggested fixes are wrong

performance_regression_test.py
```diff
@@ -270,9 +266,6 @@ def preload_data(self, compaction_strategy=None):
     def _run_cql_commands(self, cmds, node=None):
         node = node if node else self.db_cluster.nodes[0]
 
-        if not isinstance(cmds, list):
-            cmds = [cmds]
-
         for cmd in cmds:
```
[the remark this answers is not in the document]
pehala: Real bug, partially preexisting

pehala:
v2:
   * Addressed all of the missed call sites. It seems we dont have unit tests robust enough for this
     * As a folllowup I can add them, but I would like to not add morecontent to this PR

pehala:
@claude Review, try to find guards which I missed and might fail, focus on unwanted behaviour changes

sdcm/sct_config.py
```diff
@@ -956,19 +990,19 @@ class SCTConfiguration(BaseModel):
                 longer supported. Use an explicit YAML list instead.
         """,
     )
-    nemesis_interval: IntOrList = SctField(
```
fruch, line 959: @pehala why changing this back to int ? all nemesis parameter should be able to be list, since you can control each nemesis thread parameters 

the design should be that each nemesis thread should have it's own configuration

pehala: It should, but it isn't, never was. nemesis interval is used outside of the nemesis class as well and it definitely isn't written in a way that would work with list. e.g.
https://github.com/scylladb/scylla-cluster-tests/blob/b30872832be602d15255815be5f2d82352db335d/performance_regression_test.py#L301

Currently, making it int only is the correct way to do, in the long run we should make these properties inherent to the nemesis class and thus ensure it can be specified per thread

fruch (approved):

sdcm/sct_config.py
fruch: it's part of the nemesis plans  ?

pehala: Not yet, I only discovered this while doing this PR. I will make a followup task in Jira, but given nobody noticed for so long and nobody really wanted this feature for now, I would postpone it until it is needed or when we have finished other nemesis tasks.

EDIT: https://scylladb.atlassian.net/browse/SCT-517

fruch: given we have only 3 cases doing parallel nemesis (and the one that need it more then others is something that was run for 2 years now), not much surprise.

for sure picking different selectors for each thread  and implemented

dimakr:
Looks like this change broke xcloud backend deployment.

pehala:
> Looks like this change broke xcloud backend deployment.

That was quick :/ can you send me info so I can fix it?

dimakr:
> > Looks like this change broke xcloud backend deployment.
> 
> That was quick :/ can you send me info so I can fix it?

It's only on the step of reading the config before any deployment is started:
```
  sdcm/tester.py:2597:  if monitor_info["n_nodes"] > 0:
  E  TypeError: '>' not supported between instances of 'list' and 'int'
```
I already have a PR in progress for xcloud backed, where the test spotted this after rebase. So I'll add the fix and test it there

pehala:
They are more failures in the similar manner in other provisioning paths :/ I am working on fixes and tests to ensure better coverage. Either the fix gets merged today or we can revert this
```
2026-06-22 10:52:36.058000: (TestFrameworkEvent Severity.ERROR) period_type=one-time event_id=65474f4d-b8ed-4da4-9ae0-56e810c94b8e, source=LongevityTest.SetUp()
exception='>' not supported between instances of 'list' and 'int'
```

PR #14806, opened 2026-05-27
Stop nemesis before teardown diagnostics to prevent hang

timtimb0t:
gather_failure_statistics() can hang indefinitely when nemesis is still running during tearDown - nodetool commands block on nodes being actively disrupted, and stop_nemesis() in stop_resources() is never reached.

Move stop_nemesis() to early in tearDown, before any diagnostic operations

Fixes SCT-391

### Testing
https://jenkins.scylladb.com/job/scylla-staging/job/eugene_test_folder/job/teardown_test/3/

### Reminders

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved test teardown reliability by stopping the external fault injector immediately after the final test snapshot. This ensures diagnostics and final test-state capture happen while the injector is still active.
  * Avoided redundant shutdown during later cleanup so diagnostic collection and failure statistics run consistently and without interference during final analysis.

[Review Change Stack](https://app.coderabbit.ai/change-stack/scylladb/scylla-cluster-tests/pull/14806?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -3728,10 +3728,8 @@ class ClusterTester(unittest.TestCase):
         with silence(parent=self, name="Kill Stress Threads"):
             self.kill_stress_thread()
 
-        # Stopping nemesis, using timeout of 30 minutes, since replace/decommission node can take time
         if self.db_cluster:
             self.get_nemesis_report(self.db_cluster)
-            self.stop_nemesis(self.db_cluster)
             self.stop_resources_stop_tasks_threads(self.db_cluster)
             self.get_backtraces(self.db_cluster)
 
@@ -3980,6 +3978,10 @@ class ClusterTester(unittest.TestCase):
 
     def tearDown(self):
         self.teardown_started = True
+        # Stop nemesis first — if still running, it keeps disrupting nodes and
+        # diagnostic commands (gather_failure_statistics, validators) hang indefinitely.
+        if self.db_cluster:
+            self.stop_nemesis(self.db_cluster)
         with silence(parent=self, name="Enabling teardown filters"):
             enable_teardown_filters()
         with silence(parent=self, name="Sending test end event"):


pehala:
> Move stop_nemesis() to early in tearDown, before any diagnostic operations. The call is idempotent so the later call in stop_resources() becomes a no-op.

If you know you are going to stop them here, no need to stop them afterwards. Doing it like this indicates that we dont have clear path that will execute

> gather_failure_statistics() can hang indefinitely when nemesis is still running during tearDown - nodetool commands block on nodes being actively disrupted, and stop_nemesis() in stop_resources() is never reached.

How come we did not encounter it earlier? This would have been happening for years

fruch:
> > Move stop_nemesis() to early in tearDown, before any diagnostic operations. The call is idempotent so the later call in stop_resources() becomes a no-op.
> 
> If you know you are going to stop them here, no need to stop them afterwards. Doing it like this indicates that we dont have clear path that will execute

I agree there should be one place we stop it

> 
> > gather_failure_statistics() can hang indefinitely when nemesis is still running during tearDown - nodetool commands block on nodes being actively disrupted, and stop_nemesis() in stop_resources() is never reached.
> 
> How come we did not encounter it earlier? This would have been happening for years

destructive nemesis are a a cause of issue all the time, not that this fix can prevent the nemesis thread to continue.

anyhow this change make sense, to stop the nemesis as soon as possible, and not wait for the rest of the teardown to happen, since it's one of the slowest things to stop (if there's a in progress nemesis).

timtimb0t:
> How come we did not encounter it earlier? This would have been happening for years

1. gather_failure_statistics() is called only when the status is FAILED. If the test passes, this function is skipped, and tearDown quickly reaches stop_resources()

2. the race condition window is narrow. The nemesis must be in the middle of a destructive operation, with the node down, exactly when entering tearDown. With nemesis_interval=5 or 10, the chance of hitting this window is lower, i uguess but here interval=2

3. TestTimeoutEvent doesnt show the root cause. We saw test timeout several times and there were several reasons for it but we cant exclude this one too, but it imho

> If you know you are going to stop them here, no need to stop them afterwards. Doing it like this indicates that we dont have clear path that will execute

I wanted to make the change in the safest possible way, but generally agreed

pehala:
> 1. gather_failure_statistics() is called only when the status is FAILED. If the test passes, this function is skipped, and tearDown quickly reaches stop_resources()

This probably means we have to stop the thread before we check if we status is failed. At that point, we want to start ending the test and as such it makes no sense to keep nemesis running.

timtimb0t:
> > 1. gather_failure_statistics() is called only when the status is FAILED. If the test passes, this function is skipped, and tearDown quickly reaches stop_resources()
> 
> This probably means we have to stop the thread before we check if we status is failed. At that point, we want to start ending the test and as such it makes no sense to keep nemesis running.

Yes, exactly. This PR moves stop_nemesis() earlier in tearDown, before we check the final test status and before failure diagnostics are collected

sdcm/tester.py
```diff
@@ -3986,6 +3984,12 @@ def tearDown(self):
             InfoEvent(message="TEST_END").publish()
         self.save_schema()
 
+        # Stopping nemesis, using timeout of 30 minutes, since replace/decommission node can take time.
+        # Must happen before diagnostics (gather_failure_statistics, validators) — otherwise
+        # nemesis keeps disrupting nodes and diagnostic commands hang indefinitely.
+        if self.db_cluster:
+            self.stop_nemesis(self.db_cluster)
```
pehala: maybe we can move right after the self.teardown_started? It will be correctly mark Test End only after nemesis finish running

pehala:
LGTM codewise, I am interested in rerun of the test if it will indeed work. Maybe this change will help us with cleanuip the shutdown process of SCT and simplify in the future

timtimb0t:
@pehala @fruch please take a look at latest test results

fruch (approved):

PR #14813, opened 2026-05-27
fix(claude): checkout PR branch and restrict to read-only tools

fruch:
## Root cause (two issues)

**1. Wrong checkout** — we were checking out `base.sha` (the base branch tip). Claude was reading the base branch files, not the PR's actual changes. The plugin had no diff to review.

**2. Tool permission denials** — the plugin needs `Bash` to run `git diff` and `Read` to access files. Both were being denied (6 denials observed in [run 26528341484](https://github.com/scylladb/scylla-cluster-tests/actions/runs/26528341484)). Claude ran 18 turns and spent $2.59 producing zero comments.

## Fix

- Switch checkout to `head.sha` with `fetch-depth: 0` so Claude reads the actual PR code
- Restrict `Bash` to a read-only allowlist: `git diff/log/show/blame` and `gh pr view/diff` — no `Write`, `Edit`, or arbitrary shell execution

## Security

The `pull_request_target` pwn-request risk applies only when fork code is **executed**. Checking out the fork's code for **reading** is safe as long as Claude cannot run arbitrary shell commands from the working tree — which the restricted `allowedTools` enforces.

`show_full_output: true` is temporarily enabled to confirm the fix is working; can be removed once reviews are confirmed.

### DIFF
--- a/.github/workflows/claude-code-review.yml
+++ b/.github/workflows/claude-code-review.yml
@@ -47,17 +47,17 @@ jobs:
               ;;
           esac
 
-      - name: Checkout base repository
+      - name: Checkout PR branch for review
         if: steps.membership.outputs.is_member == 'true'
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6
         with:
-          # Always checkout the BASE branch, never the fork's HEAD.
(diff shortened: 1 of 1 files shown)


fruch:
Closing — reopening from upstream branch so pull_request trigger fires from the PR branch for testing.

PR #14851, opened 2026-05-29
fix(scylla-doctor): download vitals/logs with sudo on root installs

dimakr:
scylla-doctor runs via sudo on root installs, so its vitals JSON and scylla_logs tarball are root-owned. The failure-stats download useds a plain rsync as the login user and fails with `Permission denied` error.
This change fixes it by using sudo to receive files for cases with root installs.

Ref: SCT-108

### Testing

- [x] :green_circle: [artifacts-gce-image regression check](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts/job/artifacts-gce-image-new/25/)
- [x] :yellow_circle: [pr-provision-test with forced failure](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test-new/304/)
```
17:28:59  < t:2026-05-29 15:28:57,894 f:tester.py       l:3981 c:LongevityTest        p:INFO  > Scylla-doctor completed for node PR-provision-test-fix-sc-d-db-node-4b105a3a-1
...
17:29:15  < t:2026-05-29 15:29:14,537 f:tester.py       l:3981 c:LongevityTest        p:INFO  > Scylla-doctor completed for node PR-provision-test-fix-sc-d-db-node-4b105a3a-2
...
17:29:31  < t:2026-05-29 15:29:30,040 f:tester.py       l:3981 c:LongevityTest        p:INFO  > Scylla-doctor completed for node PR-provision-test-fix-sc-d-db-node-4b105a3a-3
...
```

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -3963,14 +3963,19 @@ class ClusterTester(unittest.TestCase):
                         doctor.install_scylla_doctor()
                         doctor.run_scylla_doctor_and_collect_results()
 
+                        download_with_sudo = not node.is_nonroot_install
                         if doctor.json_result_file:
                             local_json_path = os.path.join(self.logdir, f"scylla_doctor_{node.name}_vitals.json")
-                            node.remoter.receive_files(src=doctor.json_result_file, dst=local_json_path)
+                            node.remoter.receive_files(
+                                src=doctor.json_result_file, dst=local_json_path, sudo=download_with_sudo
+                            )
                             self.log.debug("Downloaded scylla-doctor vitals from %s to %s", node.name, local_json_path)
 
                         if doctor.scylla_logs_file:
                             local_logs_path = os.path.join(self.logdir, f"scylla_doctor_{node.name}_logs.tar.gz")
-                            node.remoter.receive_files(src=doctor.scylla_logs_file, dst=local_logs_path)
+                            node.remoter.receive_files(
+                                src=doctor.scylla_logs_file, dst=local_logs_path, sudo=download_with_sudo
+                            )
                             self.log.debug("Downloaded scylla-doctor logs from %s to %s", node.name, local_logs_path)
 
                         self.log.info("Scylla-doctor completed for node %s", node.name)


vponomaryov (approved):
LGTM

fruch:
@dimakr the issue refer here is not about non root install.

can you have a closer look there what the fail is ? maybe size of things ? or something else ?

i.e. I don't see how this change fixed it for a longevity run

dimakr:
> @dimakr the issue refer here is not about non root install.
> 
> can you have a closer look there what the fail is ? maybe size of things ? or something else ?
> 
> i.e. I don't see how this change fixed it for a longevity run

I disagree.
The SCT-108 issue is exactly about `Permission denied` error as per the title and description. That run is a root install - and it is exactly the problem what the fix targets: download with sudo on root installs; the non-root isn't the subject of the fix, we just ensure that we don't break that case.

The mentioned `No space left on device` is a problem indeed, but a separate one that needs follow-up investigation. It was likely mentioned to point out that we don't have sct.log to investigate the original thing.

fruch:
> > @dimakr the issue refer here is not about non root install.
> > can you have a closer look there what the fail is ? maybe size of things ? or something else ?
> > i.e. I don't see how this change fixed it for a longevity run
> 
> I disagree. The [SCT-108](https://scylladb.atlassian.net/browse/SCT-108) issue is exactly about `Permission denied` error as per the title and description. That run is a root install - and it is exactly the problem what the fix targets: download with sudo on root installs; the non-root isn't the subject of the fix, we just ensure that we don't break that case.
> 
> The mentioned `No space left on device` is a problem indeed, but a separate one that needs follow-up investigation. It was likely mentioned to point out that we don't have sct.log to investigate the original thing.

so it's not fixing SCT-108 (maybe some part of it)
also from this PR I don't understand how it's working in other places this code flow were run 
it's by default running on all tests, isn't it ? it's something unique to GCE ?

[SCT-108]: https://scylladb.atlassian.net/browse/SCT-108?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

dimakr:
As the config in question is a big setup - 6+6 db nodes, 2 loaders, 15h long longevity - we were likely hit by the size of logs being collected.
We probably can increase the disk size for runner instance for this particular scenario (via `root_disk_size_runner`)? Or maybe apply some compression to logs before they are collected from nodes?

fruch:
> As the config in question is a big setup - 6+6 db nodes, 2 loaders, 15h long longevity - we were likely hit by the size of logs being collected. We probably can increase the disk size for runner instance for this particular scenario (via `root_disk_size_runner`)? Or maybe apply some compression to logs before they are collected from nodes?

is that gonna bigger then what we collect during SCT test ? which doesn't go via the sct-runner today.
if it's gonna be that big, we have wrongly assumed what's in those vitals (if it's db logs, we don't need them, if it's system table that is huge, we need to reconsider and maybe skip those problematic ones)

we enable it to get same state as one can get from production system, it it's huge, we weren't expected for something huge

dimakr:
> if it's gonna be that big, we have wrongly assumed what's in those vitals

There is no problem with `no space left on device` when collecting scylla-doctor artifacts. That was the problem when scp-ing logs from one of db nodes.

fruch (approved):

PR #14872, opened 2026-05-31
fix(test): filter GCE region builder test by NodeType tag instead of name substring

fruch:
## Summary

The GCE region builder test was filtering instance definitions by substring match on their name (e.g. `"db" in d.name`). This is fragile because the test ID is embedded in the name — if the short test ID happens to contain the filter string (e.g. `"db"`), unrelated node types match and assertions fail.

## Fix

Filter instance definitions by their `tags["NodeType"]` value (`"scylla-db"`, `"loader"`, `"monitor"`) instead of matching substrings in the instance name. This uses the actual semantic type and is immune to test ID content.

## Test plan

- [x] All 3 parametrized GCE tests pass in isolation
- [x] All 148 provisioner unit tests pass

### DIFF
--- a/unit_tests/unit/provisioner/test_gce_region_definition_builder.py
+++ b/unit_tests/unit/provisioner/test_gce_region_definition_builder.py
@@ -58,23 +58,23 @@ def build_region_defs():
 
 
 @pytest.mark.parametrize(
-    "extra_env,name_filter,expected_disk_type",
+    "extra_env,node_type,expected_disk_type",
     [
         ({"SCT_GCE_ROOT_DISK_TYPE_LOADER": "pd-ssd"}, "loader", "pd-ssd"),
-        ({"SCT_GCE_ROOT_DISK_TYPE_DB": "pd-ssd"}, "db", "pd-ssd"),
+        ({"SCT_GCE_ROOT_DISK_TYPE_DB": "pd-ssd"}, "scylla-db", "pd-ssd"),
         ({"SCT_GCE_ROOT_DISK_TYPE_MONITOR": "pd-standard"}, "monitor", "pd-standard"),
     ],
 )
 def test_root_disk_type_flows_into_instance_definition(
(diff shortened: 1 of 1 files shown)


unit_tests/unit/provisioner/test_gce_region_definition_builder.py
```diff
@@ -33,6 +33,9 @@ def make_config(monkeypatch, defaults_dir):
     }
 
     def _factory(extra_env: dict | None = None):
+        # Reset _test_id so set_test_id_only always wins, regardless of what
+        # other xdist-worker tests have set on this singleton before us.
+        monkeypatch.setattr(TestConfig, "_test_id", None)
```
pehala: This can still race if other tests can access it in parallel ans change it

unit_tests/unit/provisioner/test_gce_region_definition_builder.py
```diff
@@ -33,6 +33,9 @@ def make_config(monkeypatch, defaults_dir):
     }
 
     def _factory(extra_env: dict | None = None):
+        # Reset _test_id so set_test_id_only always wins, regardless of what
```
pehala: I dont think this is correct solution. Correct solution is to ensure this can never race. If the race can happen only within this testfile, solution could be changing the load distribution to loadfile or using groups to ensure these tests are run together.

fruch: test id should be irrelevant to the test , had to read 3 more times to understand the this filter is within the test itself.

pehala (approved):

unit_tests/unit/provisioner/test_gce_region_definition_builder.py
fruch: and this after sitting down with @juliayakovlev for 30min last week looking at this test...

PR #14873, opened 2026-05-31
feature(docker): add Scylla Manager support for Docker backend

fruch:
## Summary

Implement Scylla Manager integration for the Docker backend, enabling manager operations (backup, repair) in local Docker-based test runs.

## Changes

- Remove validation block that prevented `use_mgmt` on Docker backend
- Add `install_scylla_manager` to `ScyllaDockerCluster` (copies agent binary from `scylladb/scylla-manager-agent` image into running Scylla containers)
- Add `MonitorSetDocker.install_scylla_manager` (runs manager server and its backing ScyllaDB as separate Docker containers on the test network)
- Add `SCToolDocker` class that executes sctool via `docker exec` on the manager server container
- Add `ScyllaManagerToolDocker` class that bypasses distro check and uses `SCToolDocker` for all sctool operations
- Wire up `get_scylla_manager_tool` to return `ScyllaManagerToolDocker` when `manager_node.is_docker()` is true
- Manager DB uses `--developer-mode 1` to work around SCYLLADB-2043 (I/O error on non-XFS filesystems with 4096 block size)

The default `use_mgmt=true` now applies to Docker backend as well.

## Backends Affected
- [x] Docker - Modified cluster_docker.py, mgmt/__init__.py, mgmt/cli.py, sct_config.py

## Required Labels
- `test-provision-docker`

### What Copilot Couldn't Test

- [x] **End-to-end manager operations on Docker**: Requires stable multi-node cluster
  - Test: `SCT_USE_MGMT=true SCT_SCYLLA_VERSION=2025.3.0 uv run sct.py run-test longevity_test.LongevityTest.test_custom_time --backend docker --config test-cases/PR-provision-test.yaml`
  - Expected: Manager registers cluster and basic backup/repair operations succeed

### Notes
- Agent installation (image pull + binary copy) verified working
- Manager DB + server container creation verified working
- `get_scylla_manager_tool` dispatch verified working
- Full E2E blocked by pre-existing ZFS issue on local machine (SCT-428 / SCYLLADB-2043)

### DIFF
--- a/sdcm/cluster_docker.py
+++ b/sdcm/cluster_docker.py
@@ -15,6 +15,7 @@ import os
 import re
 import logging
 import shutil
+import time
 from typing import Optional, Union, Dict
 from functools import cached_property
 
@@ -471,6 +472,61 @@ class ScyllaDockerCluster(cluster.BaseScyllaCluster, DockerCluster):
 
         node.config_setup(append_scylla_args=self.get_scylla_args())
         node.restart_scylla(verify_up_before=True)
+
+        if self.params.get("use_mgmt"):
+            self.install_scylla_manager(node)
+
+    def install_scylla_manager(self, node):
+        """Install manager agent into running Scylla Docker container.
+
+        Copies the agent binary from the scylladb/scylla-manager-agent image,
+        writes config with auth_token, and starts the agent process.
+        """
+        mgmt_agent_image = "scylladb/scylla-manager-agent:latest"
+        auth_token = self.test_config.test_id()
+        manager_prometheus_port = self.params.get("manager_prometheus_port") or "56090"
+
+        LOGGER.info("Installing manager agent on Docker node %s", node.name)
+
+        node.remoter.run("sudo mkdir -p /etc/scylla-manager-agent", ignore_status=True)
+        node.remoter.run(
+            f"sudo bash -c 'cat > /etc/scylla-manager-agent/scylla-manager-agent.yaml << EOF\n"
+            f"auth_token: {auth_token}\n"
+            f'prometheus: ":{manager_prometheus_port}"\n'
+            f"EOF'",
+        )
+
+        container = ContainerManager.get_container(node, "node")
+
+        from sdcm.utils.docker_utils import _default_docker_client  # noqa: PLC0415 - avoid circular import
+
+        docker_client = _default_docker_client()
+
+        tmp_container_name = f"mgmt-agent-copy-{node.node_index}"
+        try:
+            docker_client.containers.get(tmp_container_name).remove(force=True)
+        except Exception:  # noqa: BLE001
+            pass
+
+        docker_client.images.pull(*mgmt_agent_image.split(":", maxsplit=1))
+        tmp_container = docker_client.containers.create(
+            mgmt_agent_image,
+            name=tmp_container_name,
+            command="sleep 1",
+        )
+        try:
+            bits, _ = tmp_container.get_archive("/usr/bin/scylla-manager-agent")
+            container.put_archive("/usr/bin/", bits)
+        finally:
+            tmp_container.remove(force=True)
+
+        node.remoter.run("sudo chmod +x /usr/bin/scylla-manager-agent")
+
+        node.remoter.run(
+            "sudo bash -c 'nohup /usr/bin/scylla-manager-agent > /var/log/scylla-manager-agent.log 2>&1 &'",
+        )
+
+        node.wait_manager_agent_up(timeout=60)
 
     def _generate_db_node_certs(self, node):
         """Generate per-node SSL certificates for a Docker DB node."""
@@ -967,6 +1023,9 @@ class MonitorSetDocker(cluster.BaseMonitorSet, DockerCluster):
             n_nodes=n_nodes,
             params=params,
         )
+        self._manager_container = None
+        self._manager_db_container = None
+        self.manager_container_name = None
 
     def _create_node(self, node_index, container=None, after_config=None):
         node = DockerMonitoringNode(
@@ -996,6 +1055,73 @@ class MonitorSetDocker(cluster.BaseMonitorSet, DockerCluster):
     def install_scylla_monitoring_prereqs(node):
         pass  # since running local, don't install anything, just the monitor
 
+    def install_scylla_manager(self, node):
+        """Start manager server + manager-db as Docker containers."""
+        from sdcm.utils.docker_utils import _default_docker_client  # noqa: PLC0415 - avoid circular import
+
+        docker_client = _default_docker_client()
+        docker_network = self.params.get("docker_network")
+        user_prefix = self.params.get("user_prefix") or "sct"
+
+        LOGGER.info("Starting Scylla Manager DB container")
+        mgr_db_name = f"{user_prefix}-manager-db"
+        try:
+            docker_client.containers.get(mgr_db_name).remove(force=True)
+        except Exception:  # noqa: BLE001
+            pass
+
+        mgr_db_image = self.params.get("scylla_docker_image") or "scylladb/scylla:latest"
+        docker_client.images.pull(*mgr_db_image.split(":", maxsplit=1))
+        self._manager_db_container = docker_client.containers.run(
+            image=mgr_db_image,
+            name=mgr_db_name,
+            command="--smp 1 --memory 512M --developer-mode 1",
+            network=docker_network,
+            detach=True,
+        )
+
+        LOGGER.info("Waiting for manager DB to be ready")
+        for _ in range(60):
+            time.sleep(5)
+            exit_code, _ = self._manager_db_container.exec_run("cqlsh -e 'SELECT now() FROM system.local'")
+            if exit_code == 0:
+                break
+        else:
+            raise RuntimeError("Manager DB did not become ready in time")
+
+        LOGGER.info("Starting Scylla Manager server container")
+        mgr_name = f"{user_prefix}-manager-server"
+        try:
+            docker_client.containers.get(mgr_name).remove(force=True)
+        except Exception:  # noqa: BLE001
+            pass
+
+        mgmt_image = self.params.get("mgmt_docker_image") or "scylladb/scylla-manager:latest"
+        docker_client.images.pull(*mgmt_image.split(":", maxsplit=1))
+        self._manager_container = docker_client.containers.run(
+            image=mgmt_image,
+            name=mgr_name,
+            network=docker_network,
+            environment={
+                "SCYLLA_MANAGER_DB_HOSTS": mgr_db_name,
+            },
+            detach=True,
+        )
+
+        LOGGER.info("Waiting for manager server to be ready")
+        for _ in range(60):
+            time.sleep(5)
+            exit_code, output = self._manager_container.exec_run(
+                'curl --silent --output /dev/null --write-out "%{http_code}" http://127.0.0.1:5080/ping'
+            )
+            if exit_code == 0 and b"204" in output:
+                break
+        else:
+            raise RuntimeError("Manager server did not become ready in time")
+
+        LOGGER.info("Scylla Manager is running in container %s", mgr_name)
+        self.manager_container_name = mgr_name
+
     def get_backtraces(self):
         pass
 
@@ -1007,3 +1133,14 @@ class MonitorSetDocker(cluster.BaseMonitorSet, DockerCluster):
             except Exception as exc:  # noqa: BLE001
                 self.log.error(f"Stopping scylla monitoring failed with {exc!s}")
             node.destroy()
+
+        if self._manager_container:
+            try:
+                self._manager_container.remove(force=True)
+            except Exception:  # noqa: BLE001
+                pass
+        if self._manager_db_container:
+            try:
+                self._manager_db_container.remove(force=True)
+            except Exception:  # noqa: BLE001
+                pass
--- a/sdcm/cluster_docker.py
+++ b/sdcm/cluster_docker.py
@@ -31,7 +31,13 @@ from sdcm.remote import LOCALRUNNER
 from sdcm.remote.docker_cmd_runner import DockerCmdRunner
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.filters import DbEventsFilter
-from sdcm.utils.docker_utils import get_docker_bridge_gateway, Container, ContainerManager, DockerException
+from sdcm.utils.docker_utils import (
+    get_docker_bridge_gateway,
+    Container,
+    ContainerManager,
+    DockerException,
+    _default_docker_client,
+)
 from sdcm.utils.health_checker import check_nodes_status
 from sdcm.nemesis.utils.node_allocator import mark_new_nodes_as_running_nemesis
 from sdcm.utils.net import get_my_public_ip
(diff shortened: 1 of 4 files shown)


sdcm/cluster_docker.py
```diff
@@ -946,6 +1005,73 @@ def add_nodes(
...
+            pass
+
+        mgmt_image = self.params.get("mgmt_docker_image") or "scylladb/scylla-manager:latest"
+        docker_client.images.pull(*mgmt_image.split(":", maxsplit=1))
+        self._manager_container = docker_client.containers.run(
+            image=mgmt_image,
+            name=mgr_name,
+            network=docker_network,
+            environment={
+                "SCYLLA_MANAGER_DB_HOSTS": mgr_db_name,
```
dimakr: As per Docker docs `Containers on the default bridge network can only access each other by IP addresses`, and as `docker_network` config parameter is not set by default in the SCT, we will use the default bridge here. So the resolution by `mgr_db_name` value won't work. The server won't be able to reach db mgr node then

sdcm/cluster_docker.py
```diff
@@ -473,6 +474,61 @@ def node_setup(self, node, verbose=False, timeout=3600):
...
+        node.remoter.run(
+            f"sudo bash -c 'cat > /etc/scylla-manager-agent/scylla-manager-agent.yaml << EOF\n"
+            f"auth_token: {auth_token}\n"
+            f'prometheus: ":{manager_prometheus_port}"\n'
+            f"EOF'",
+        )
+
+        container = ContainerManager.get_container(node, "node")
+
+        from sdcm.utils.docker_utils import _default_docker_client  # noqa: PLC0415 - avoid circular import
```
dimakr: There is no circular import - this module already imports from sdcm.utisls.docker_utils at https://github.com/scylladb/scylla-cluster-tests/pull/14873/changes#diff-00ce153d4f59525276fd971dfd2029cd20b921b4646aba953208e77f065c9ea4R35

sdcm/cluster_docker.py
```diff
@@ -946,6 +1005,73 @@ def add_nodes(
     def install_scylla_monitoring_prereqs(node):
         pass  # since running local, don't install anything, just the monitor
 
+    def install_scylla_manager(self, node):
+        """Start manager server + manager-db as Docker containers."""
+        from sdcm.utils.docker_utils import _default_docker_client  # noqa: PLC0415 - avoid circular import
```
dimakr: ditto about no circular imports' problem

sdcm/cluster_docker.py
```diff
@@ -473,6 +474,61 @@ def node_setup(self, node, verbose=False, timeout=3600):
...
+        try:
+            bits, _ = tmp_container.get_archive("/usr/bin/scylla-manager-agent")
+            container.put_archive("/usr/bin/", bits)
+        finally:
+            tmp_container.remove(force=True)
+
+        node.remoter.run("sudo chmod +x /usr/bin/scylla-manager-agent")
+
+        node.remoter.run(
+            "sudo bash -c 'nohup /usr/bin/scylla-manager-agent > /var/log/scylla-manager-agent.log 2>&1 &'",
```
dimakr, line 543: will agent started this way survive destructive nemesis (where container is restarted)?

fruch: probably not, but not sure we have destructive nemesis for manger in our nemesis list

fruch:
Addressed all review comments in b3a7a7a:

1. **`__init__.py:19` — Unprotected attribute chain** ✅ Fixed. Added `getattr` guard with descriptive `RuntimeError`.

2. **`cli.py:1565` — Shell injection in docker exec** — Dismissed. `container_name` is internally generated, `cmd` comes from our own sctool wrappers. Same pattern as all `remoter.run` calls in SCT.

3. **`cluster_docker.py:1023` — Wrong config key** ✅ Fixed. Changed to `docker_image`.

4. **`cluster_docker.py:1056` — Bridge network name resolution** ✅ Fixed. Now uses container IP.

5. **`cluster_docker.py:502,1010` — Inline imports** ✅ Fixed. Moved to top-level.

6. **`cluster_docker.py:527` — Agent survival after nemesis** ✅ Fixed. Added `_restart_manager_agent_if_needed()` hook in `start_scylla_server`.

sdcm/mgmt/__init__.py
```diff
@@ -1,17 +1,22 @@
...
+]
 AnyManagerCluster = Union[OperatorManagerCluster, ManagerCluster]
 
 
 def get_scylla_manager_tool(manager_node, scylla_cluster=None) -> AnyManagerTool:
     if manager_node.is_kubernetes():
         return ScyllaManagerToolOperator(manager_node=manager_node, scylla_cluster=scylla_cluster)
+    if manager_node.is_docker():
+        container_name = manager_node.parent_cluster.manager_container_name
+        return ScyllaManagerToolDocker(manager_node=manager_node, manager_container_name=container_name)
```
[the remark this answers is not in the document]
fruch: Fixed. \`get_scylla_manager_tool()\` now validates both \`parent_cluster\` and \`manager_container_name\` with \`getattr\` and raises a descriptive \`RuntimeError\` pointing at the missing \`install_scylla_manager()\` call, instead of dereferencing the chain blindly (sdcm/mgmt/__init__.py).

sdcm/mgmt/cli.py
```diff
@@ -1545,3 +1545,50 @@ def client_version_timestamp(self) -> int:
...
+        self,
+        cmd,
+        is_verify_errorless_result=False,
+        parse_table_res=True,
+        is_multiple_tables=False,
+        replace_broken_unicode_values=True,
+    ):
+        LOGGER.debug("Issuing: 'sctool %s' in container %s", cmd, self.container_name)
+        try:
+            res = self.manager_node.remoter.run(f"docker exec {self.container_name} sctool {cmd}")
```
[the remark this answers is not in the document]
fruch: Partially applied. The container name is now passed through \`shlex.quote()\` (\`SCToolDocker._run_sctool\`).

\`cmd\` itself is deliberately left unquoted: it is a full sctool command line with flags and arguments (e.g. \`backup -c <id> --location s3:bucket\`), built by our own sctool wrappers in this module, so quoting it as a single token would break every call. This is exactly what the base \`SCTool.run()\` does (\`remoter.sudo(f"sctool {cmd}")\`) — there is no argv form on \`remoter.run\`, and no user-controlled input reaches it.

sdcm/cluster_docker.py
```diff
@@ -946,6 +1005,73 @@ def add_nodes(
...
+        user_prefix = self.params.get("user_prefix") or "sct"
+
+        LOGGER.info("Starting Scylla Manager DB container")
+        mgr_db_name = f"{user_prefix}-manager-db"
+        try:
+            docker_client.containers.get(mgr_db_name).remove(force=True)
+        except Exception:  # noqa: BLE001
+            pass
+
+        mgr_db_image = self.params.get("scylla_docker_image") or "scylladb/scylla:latest"
```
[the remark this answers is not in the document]
fruch: Good catch — fixed. The manager DB container now uses \`self.params.get("docker_image")\`. The dead \`or "scylladb/scylla:latest"\` fallback was dropped too, since \`docker_image\` always has a value for the docker backend.

sdcm/cluster_docker.py
fruch: Right, fixed. \`SCYLLA_MANAGER_DB_HOSTS\` is now set to the manager DB container's IP, read back from \`container.attrs["NetworkSettings"]["Networks"][docker_network or "bridge"]["IPAddress"]\` after a \`reload()\`, instead of relying on name resolution that the default bridge does not provide.

sdcm/cluster_docker.py
fruch: You are right — fixed. \`_default_docker_client\` is now imported at the top of the module together with the other \`sdcm.utils.docker_utils\` names.

sdcm/cluster_docker.py
fruch: Same fix — the second inline import is gone as well, both call sites use the top-level import now.

sdcm/cluster_docker.py
fruch: Addressed. \`DockerNode._ensure_manager_agent_running()\` now runs after both \`start_scylla_server()\` and \`restart_scylla_server()\`, so the agent comes back after a container restart.

It is idempotent: it returns early when the agent binary was never installed on the node, and also when \`pgrep -f /usr/bin/scylla-manager-agent\` finds it still alive, so a no-op \`start_scylla_server()\` on a live container cannot spawn a second agent process. After a relaunch it waits on \`wait_manager_agent_up(timeout=60)\` rather than assuming success.

sdcm/mgmt/__init__.py
```diff
@@ -1,17 +1,29 @@
 from typing import Union
 
 from .common import ScyllaManagerError, TaskStatus, HostStatus, HostSsl, HostRestStatus
-from .cli import ScyllaManagerToolRedhatLike, ScyllaManagerToolNonRedhat, ManagerCluster
+from .cli import ScyllaManagerToolRedhatLike, ScyllaManagerToolNonRedhat, ScyllaManagerToolDocker, ManagerCluster
```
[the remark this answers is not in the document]
fruch: Fixed — the import is now \`from .cli import ManagerCluster, ScyllaManagerToolDocker, ScyllaManagerToolNonRedhat, ScyllaManagerToolRedhatLike\`, which also cleans up the pre-existing ordering violation on that line.

sdcm/cluster_docker.py
```diff
@@ -996,6 +1072,76 @@ def add_nodes(
...
+        LOGGER.info("Waiting for manager server to be ready")
+        for _ in range(60):
+            time.sleep(5)
+            exit_code, output = self._manager_container.exec_run(
+                'curl --silent --output /dev/null --write-out "%{http_code}" http://127.0.0.1:5080/ping'
+            )
+            if exit_code == 0 and b"204" in output:
+                break
+        else:
+            raise RuntimeError("Manager server did not become ready in time")
```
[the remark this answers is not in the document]
fruch: Not applicable — the premise is wrong. `curl` is present in the image:

```
$ docker run --rm --entrypoint sh scylladb/scylla-manager:latest -c 'ls -l /usr/bin/curl'
-rwxr-xr-x 1 root root ... /usr/bin/curl
```

(same for `scylladb/scylla:latest`, which the agent-side `wait_manager_agent_up()` check relies on). Keeping the `/ping` readiness loop as is.

fruch:
Rebased on current `master` and squashed into a single commit. The conflict was `sdcm/sct_config.py`, which became the `sdcm/sct_config/` package on master — the change moved to `sdcm/sct_config/config.py`.

While resolving it I removed `_validate_docker_backend_parameters()` outright instead of leaving an empty `pass` body, and dropped the `patch.object(SCTConfiguration, "_validate_docker_backend_parameters")` line in `sdcm/utils/lint/validator.py` that only existed to work around that check.

Review round:

| Thread | Outcome |
|---|---|
| `mgmt/__init__.py` unprotected attribute chain | fixed — `getattr` guards + descriptive `RuntimeError` |
| `mgmt/cli.py` shell interpolation in `docker exec` | container name now `shlex.quote()`d; `cmd` stays a shell string, same as base `SCTool.run()` |
| `cluster_docker.py` wrong config key for manager DB image | fixed — `docker_image` |
| manager DB unreachable by name on the default bridge | fixed — uses the container IP |
| inline imports of `_default_docker_client` (x2) | fixed — moved to module top |
| agent surviving a container restart | fixed — `_ensure_manager_agent_running()` after start *and* restart, `pgrep`-guarded so it can't double-spawn, waits on `wait_manager_agent_up()` |
| import ordering in `mgmt/__init__.py` | fixed |
| `curl` missing from `scylladb/scylla-manager` | not applicable — `curl` is in the image (verified by running it) |

Also fixed while in here:
- The `ruff` failure that broke the precommit stage (`.format()` → f-string).
- The manager agent image was pinned to `:latest` while the server came from `mgmt_docker_image` (default `3.12.0`) — the agent now tracks `scylla_mgmt_agent_version`, so server and agent versions match.
- `SCToolDocker` duplicated the whole of `SCTool.run()` to change one line. Extracted `SCTool._run_sctool()` as the single override point; `SCToolDocker` is now four lines and can't drift from the base parsing logic.

Two CodeRabbit findings from the 2026-07-07 review (`scylla_version` being a `cached_property` that memoizes a transient `None`, and `client_version_timestamp` using a naive `datetime.timestamp()`) are on pre-existing `master` code that this PR does not touch — out of scope here, worth a separate PR.

fruch:
CI note: every test stage in [build #5](https://jenkins.scylladb.com/job/sct-github-PRs-scan/job/scylla-cluster-tests/job/PR-14873/5/) is green — precommit, unittests, collect_tests, lint_test_cases, provision_docker (`1 passed in 277.08s`, `SUCCESS :)`), and the summary comment above reports `Test Summary: PASSED` (7744 tests, 0 failed).

The red `continuous-integration/jenkins/pr-merge` comes from the post-build *Restore Monitor stack* step:

```
Failed to download monitoring stack archive: Argus communication failed: no redirect occurred.
This may indicate authentication issues or a Cloudflare access problem. Status: 405,
URL: https://argus.scylladb.com/api/v1/tests/.../log/monitor-set-....tar.zst/download
Errors were found when restoring Scylla monitoring stack
hudson.AbortException: script returned exit code 1
```

Not specific to this PR — PR #15681 hits the identical Argus 405 on the same step.

Also worth noting for reviewers: the `provision_docker` stage runs with `use_mgmt: false`, so CI does not exercise the manager-on-Docker path added here. Green CI confirms the rebase and refactors do not regress the existing Docker backend; the manager path still needs the manual run described in the PR body.

fruch:
@dimakr @mikliapko — rebased on current `master` (`5ec69f4`), ready for another look when you have a moment.

Nothing changed in the patch itself: `git patch-id` is identical to the previous head (`9e79389`), only hunk offsets moved. All your review comments from the earlier round are addressed and answered inline above.

The red CI on the previous run was **not** this PR — every test stage passed (`Test Summary: PASSED`, 7744 tests, 0 failed) and the umbrella job died in the post-build *Restore Monitor stack* step on an Argus `405`. That turned out to be Argus dropping `HEAD` support in its Flask→FastAPI migration; it was fixed server-side in scylladb/argus#1091, so this run should come back clean.

Still worth flagging for review: the `provision_docker` CI stage runs with `use_mgmt: false`, so CI does not exercise the manager-on-Docker path this PR adds — that part is covered only by the manual run described in the PR description.

sdcm/cluster_docker.py
```diff
@@ -1029,6 +1114,76 @@ def add_nodes(
...
+        mgr_db_ip = self._manager_db_container.attrs["NetworkSettings"]["Networks"][docker_network or "bridge"][
+            "IPAddress"
+        ]
+
+        self._manager_container = docker_client.containers.run(
+            image=mgmt_image,
+            name=mgr_name,
+            network=docker_network,
+            environment={
+                "SCYLLA_MANAGER_DB_HOSTS": mgr_db_ip,
```
dimakr: Does Scylla Manager read this env var? Seems like it only uses its config file, and the config baked into the image has `database.hosts: [scylla-manager-db]`.

sdcm/mgmt/cli.py
```diff
@@ -1633,3 +1638,27 @@ def client_version_timestamp(self) -> int:
...
+
+    def __init__(self, manager_node, container_name):
+        super().__init__(manager_node=manager_node)
+        self.container_name = container_name
+
+    def _run_sctool(self, cmd):
+        return self.manager_node.remoter.run(f"docker exec {shlex.quote(self.container_name)} sctool {cmd}")
+
+
+class ScyllaManagerToolDocker(ScyllaManagerTool):
```
dimakr: SCToolDocker is only used by this class. ManagerCluster and the task objects (repair, backup, restore) still create a plain SCTool, so they run `sudo sctool` on the host, where it doesn't exist, instead of in the manager container.
So get_cluster() would alway returns None, the cluster get re-added on every call, and any backup or repair fail. They need to use the docker sctool too.

PR #14874, opened 2026-05-31
refactor(ssh-transport): make libssh2 the default SSH transport

fruch:
## Summary

Switch the default SSH transport from Fabric to libssh2 across all SCT components.

This removes the legacy Fabric dependency from the critical path, including the create-runner flow which was failing on OCI due to key compatibility issues with Fabric.

## Changes

- Remove `default=True` from `RemoteCmdRunner` (Fabric-based)
- Add `default=True` to `RemoteLibSSH2CmdRunner` (libssh2-based)
- Remove hardcoded `set_default_ssh_transport("fabric")` in `sct_ssh.py`
- Update `ssh_transport` config description to reflect new default

## Context

The libssh2 transport has been available as an alternative and is lighter weight, using a thin C library wrapper instead of the full Fabric/Paramiko/Invoke stack. The config default was already set to `"libssh2"` but the programmatic default was still Fabric.

Fixes: https://scylladb.atlassian.net/browse/SCT-343

## Backends Affected
- [x] AWS
- [x] GCE
- [x] Azure
- [x] OCI
- [x] Docker

### What Copilot Couldn't Test
- [x] **End-to-end provisioning with libssh2**: Requires cloud backends
- [ ] **OCI key compatibility**: The original issue trigger
- [x] **create-runner-instance flow**: Requires actual runner provisioning

### DIFF
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -780,7 +780,7 @@ Email subject postfix
 
 ## **ssh_transport** / SCT_SSH_TRANSPORT
 
-Set type of ssh library to use. Could be 'fabric' (default) or 'libssh2'
+Set type of ssh library to use. Could be 'libssh2' (default) or 'fabric'
 
 **default:** libssh2
 
--- a/sdcm/remote/remote_cmd_runner.py
+++ b/sdcm/remote/remote_cmd_runner.py
@@ -44,7 +44,7 @@ def open_none_auth(self):
     self.transport = self.client._transport = transport
 
 
-class RemoteCmdRunner(RemoteCmdRunnerBase, ssh_transport="fabric", default=True):
+class RemoteCmdRunner(RemoteCmdRunnerBase, ssh_transport="fabric"):
     connection: Connection
     ssh_config: Config = None
     ssh_is_up: threading.Event = None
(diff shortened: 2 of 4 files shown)


sct_ssh.py
```diff
@@ -674,7 +674,6 @@ def init_nodes(self, instances):
             }
             node = BaseNode(name=get_name(instance), parent_cluster=self, ssh_login_info=login_info)
             node.set_seed_flag(True)
-            RemoteCmdRunnerBase.set_default_ssh_transport("fabric")
```
fruch, line 677: don't drop this, it's probably import for this command

dimakr:
No comments to code itself. Just 2 things:
- PR description is stale - mentions `Remove hardcoded set_default_ssh_transport("fabric") in sct_ssh.py` among changes, but the statement wasn't really removed, just switched to setting libssh2
- the change mentions OCI backend in particular, that it was having with fabric. Probably makes sense to test it on OCI (at least provisioning test)

sct_ssh.py
```diff
@@ -674,7 +674,7 @@ def init_nodes(self, instances):
...
+            RemoteCmdRunnerBase.set_default_ssh_transport("libssh2")
```
fruch: we shouldn't change this line, not without test it carefully that the logic can really work with libssh2

fruch:
OCI provision test passed, merging this

PR #14895, opened 2026-06-02
test: import production code instead of inline copies in unit tests

pehala:
## Summary

- Replace inline copies of production functions in unit tests with direct imports, ensuring tests validate real behavior
- Extract `build_gce_image_filter()` from `get_scylla_gce_images_versions()` to make the filter logic independently testable
- Fix incorrect test assertions in GCE filter tests that had silently drifted from production behavior

## Motivation

Three test files contained duplicated production logic instead of importing and testing the real code. This creates a risk of **silent drift** — if production logic changes, tests still pass because they validate a stale copy. This was already happening: the GCE filter test omitted the `labels.environment eq 'production'` filter for `None`/`"all"` versions, diverging from actual production behavior.

## Performance Impact

| Test file | Before | After | Delta |
|-----------|--------|-------|-------|
| `test_docker_tar_stream.py` | 5.95s | 4.52s | -1.4s |
| `test_gce_image_filter.py` | 4.66s | 4.70s | +0.04s |
| `test_gradual_grow_throughput.py` | 4.63s | 5.55s | +0.9s |

The slight increase in `test_gradual_grow_throughput.py` is due to the transitive import chain of `performance_regression_gradual_grow_throughput.py` (which pulls in `sdcm.utils.common`). This is acceptable given the correctness guarantee.

### DIFF
--- a/sdcm/utils/common.py
+++ b/sdcm/utils/common.py
@@ -1105,15 +1105,15 @@ def get_vector_store_ami_versions(
     )
 
 
-@lru_cache
-def get_scylla_gce_images_versions(
-    project: str = SCYLLA_GCE_IMAGES_PROJECT, version: str = None, arch: VmArch = None
-) -> list[GceImage]:
-    # Server-side resource filtering described in Google SDK reference docs:
-    #   API reference: https://cloud.google.com/compute/docs/reference/rest/v1/images/list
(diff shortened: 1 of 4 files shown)


fruch (approved):

cezarmoise (approved):
LGTM

PR #14902, opened 2026-06-02
fix(nemesis): fix adaptive timeout for replacing a node

cezarmoise:
When replacing a node with RBNO enabled, the TABLET_MIGRATION operation should be used, since it will also stream the data during init.

Fixes: SCT-437

### Testing

- [ ] https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/cezar/job/jira/job/SCT-437-rbno-timeout/3/

### Reminders

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -1240,8 +1240,13 @@ class NemesisRunner:
             new_node.replacement_host_id = host_id
         else:
             new_node.replacement_node_ip = old_node_ip
+        # With RBNO, the new node will also stream data during init
+        if verification_node.get_scylla_config_param("enable_repair_based_node_ops") == "false":
+            operation = Operations.NEW_NODE
+        else:
+            operation = Operations.TABLET_MIGRATION
         try:
-            with adaptive_timeout(Operations.NEW_NODE, node=self.cluster.data_nodes[0], timeout=timeout):
+            with adaptive_timeout(operation, node=self.cluster.data_nodes[0], timeout=timeout):
                 self.cluster.wait_for_init(node_list=[new_node], timeout=timeout, check_node_health=False)
             self.actions_log.info(f"New node initialized: {new_node.name}")
             self.cluster.clean_replacement_node_options(new_node)


pehala:
> node with RBNO enabled,

Only for RBNO? Isnt this the case for all tablets, we wait for tablet migration for bootstrap as well, right?

cezarmoise:
> > node with RBNO enabled,
> 
> Only for RBNO? Isnt this the case for all tablets, we wait for tablet migration for bootstrap as well, right?

Yes, only for replacing a node, it happens in one step. During this time it has status UJ instead of UN.
```
May 30 18:27:28.430079 elasticity-test-nemesis-master-db-node-56a83da5-8 scylla[9200]:  [shard 0:strm] storage_service - entering JOINING mode
...
May 30 20:20:59.556034 elasticity-test-nemesis-master-db-node-56a83da5-8 scylla[9200]:  [shard 0:strm] storage_service - entering NORMAL mode
```
The normal case when adding a node we still wait for tablet migration, but init finishes before.

vponomaryov (approved):
LGTM

yarongilor (commented):
How about considering a clearer definition of this different "REPLACE_NODE" operation as well?
It's not accurate to say it's a "NEW_NODE" and no just "tablets-migration". you can consider adding something like:

```
-            with adaptive_timeout(Operations.NEW_NODE, node=self.cluster.data_nodes[0], timeout=timeout):
+            with adaptive_timeout(Operations.REPLACE_NODE, node=self.cluster.data_nodes[0], timeout=timeout):
                 self.cluster.wait_for_init(node_list=[new_node], timeout=timeout, check_node_health=False)
             self.actions_log.info(f"New node initialized: {new_node.name}")
             self.cluster.clean_replacement_node_options(new_node)
diff --git a/sdcm/utils/adaptive_timeouts/__init__.py b/sdcm/utils/adaptive_timeouts/__init__.py
index 9cc5a1d31..0c0a1c5bc 100644
--- a/sdcm/utils/adaptive_timeouts/__init__.py
+++ b/sdcm/utils/adaptive_timeouts/__init__.py
@@ -136,6 +136,7 @@ class Operations(Enum):
 
     DECOMMISSION = ("decommission", _get_decommission_timeout, ())
     NEW_NODE = ("new_node", _get_new_node_timeout, ("timeout",))
+    REPLACE_NODE = ("replace_node", _get_tablet_migration_timeout, ("timeout",))
     CREATE_INDEX = ("create_index", _get_soft_timeout, ("timeout",))
     START_SCYLLA = ("start_scylla", _get_soft_timeout, ("timeout",))
     RESTART_SCYLLA = ("restart_scylla", _get_soft_timeout, ("timeout",))
@@ -235,7 +236,12 @@ def adaptive_timeout(  # noqa: PLR0914
     Also store node load info and timeout value in AdaptiveTimeoutStore (ES by default) for future reference.
     Use Operation.SOFT_TIMEOUT to set timeout explicitly without calculations.
     """
-    tablet_sensitive_op = operation in {Operations.DECOMMISSION, Operations.NEW_NODE, Operations.TABLET_MIGRATION}
+    tablet_sensitive_op = operation in {
+        Operations.DECOMMISSION,
+        Operations.NEW_NODE,
+        Operations.REPLACE_NODE,
+        Operations.TABLET_MIGRATION,
+    }
```

cezarmoise:
> How about considering a clearer definition of this different "REPLACE_NODE" operation as well?
> It's not accurate to say it's a "NEW_NODE" and no just "tablets-migration". you can consider adding something like:

Not convinced this is clearer. Because there are indeed 2 cases, without RBNO, it's NEW_NODE as before, but with RBNO it's technically NEW_NODE and TABLET_MIGRATION, but the migration part is considerably bigger so only taking that in consideration seems ok to me.

yarongilor:
> > How about considering a clearer definition of this different "REPLACE_NODE" operation as well?
> > It's not accurate to say it's a "NEW_NODE" and no just "tablets-migration". you can consider adding something like:
> 
> Not convinced this is clearer. Because there are indeed 2 cases, without RBNO, it's NEW_NODE as before, but with RBNO it's technically NEW_NODE and TABLET_MIGRATION, but the migration part is considerably bigger so only taking that in consideration seems ok to me.

Well, i think, in principal, we should prefer focusing on RBNO and Tablets if non-RBNO and non-tablets is legacy.
as for TABLET_MIGRATION:

1. It may sound more confusing for the same operation to have 2 different timeouts.
2. TABLET_MIGRATION is a general process that is defined on various operations, in SCT as well, like change-rf.
3. in scylla, a tablet-migration process happens also during repair/rebuild/decommission etc. So "replace-node" is a specific different "full" mode of tablet-migration.
4. If a test will get this "NEW_NODE" soft timeout - how can QA/AI know/tell if it's RBNO/tablets and the consequences of that, comparing to the other mode of "NEW_NODE"?

yarongilor (approved):
LGTM. 
The comment of https://github.com/scylladb/scylla-cluster-tests/pull/14902#issuecomment-4641763144 can be addressed later on as well.

soyacz (commented):
I think we should add a new operation then. Or adjusting timeout by adjusting `adaptive_timeout_multipliers` for this run.

pehala:
> I think we should add a new operation then. Or adjusting timeout by adjusting `adaptive_timeout_multipliers` for this run.

I agree, but currently it is failing Tier 1 so I would like to merge this as a urgent fix and submit a followup where we improve it, if you are ok with it @soyacz

soyacz:
> > I think we should add a new operation then. Or adjusting timeout by adjusting `adaptive_timeout_multipliers` for this run.
> 
> I agree, but currently it is failing Tier 1 so I would like to merge this as a urgent fix and submit a followup where we improve it, if you are ok with it @soyacz

I'm ok with it.

PR #14906, opened 2026-06-02
feature(adaptive-timeouts): per-operation timeout multipliers

aleksbykov:
Implement per-operation multipliers for adaptive timeouts. Each operation can now have its own independent scaling factor configured via 'adaptive_timeout_multipliers'.

When configured, only the multiplier matching the current operation is applied. Operations without a configured multiplier default to 1.0. When not configured (default null), existing behavior is preserved.

Example usage in test-case YAML:
  adaptive_timeout_multipliers:
    decommission: 4
    new_node: 2

This scales decommission timeouts by 4x and new_node by 2x.

Implementation:
- AdaptiveTimeoutMultipliers RootModel with confloat(gt=0) for native Pydantic value validation and get_multiplier() method
- Valid keys are derived dynamically from Operations enum (operation.value[0]), so any new operation added to the enum is automatically supported
- _get_operation_timeout_factor() resolves multiplier using operation.value[0]
- Config validation rejects unknown keys with clear error messages listing all valid operation names
- Bool values are explicitly rejected to prevent accidental misconfiguration

Argus config serialization fix:
- Replaced json.dumps(dict(self.params)) with self.params.model_dump_json() in init_argus_run() (sdcm/tester.py) to prevent ValueError: Circular reference detected when serializing AdaptiveTimeoutMultipliers RootModel
- Root cause: dict() on Pydantic BaseModel produces shallow dict where nested models remain as raw Python objects; stdlib json encoder cannot serialize them (observed on Python 3.14, test run efd56534)
- model_dump_json() uses Pydantic's recursive serializer that properly converts all nested models including RootModel subclasses
- Consistent with self.params.model_dump() already used for submit_sct_run()
- Fields with Field(exclude=True) are omitted from Argus submission: multi_region_params, regions_data, artifact_scylla_version, is_enterprise, scylla_version_upgrade_target, target_db_image_ids

Regression tests added:
- test_sct_submit_config_receives_valid_json: exercises init_argus_run() with mocked Argus client, asserts sct_submit_config receives valid JSON with properly serialized adaptive_timeout_multipliers
- test_sct_submit_config_contains_expected_fields: verifies essential fields present and exclude=True fields absent
- test_sct_config_with_multipliers_is_json_serializable: standalone model-level test for SCTConfiguration.model_dump_json()

Workaround for https://scylladb.atlassian.net/browse/SCT-356

Including fix for Reverted PR: https://github.com/scylladb/scylla-cluster-tests/pull/14669

Fixes: SCYLLADB-2364

### Testing

- [test1](https://argus.scylladb.com/tests/scylla-cluster-tests/5e5e4062-68cb-46c7-b381-4293da31ffd4) - test failed by another reason. because small multipliers were set. Or scylladb have a bug. But multipliers applied and no error from argus.
- [test2](https://argus.scylladb.com/tests/scylla-cluster-tests/5e5e4062-68cb-46c7-b381-4293da31ffd4) test is running. not mentioned errors for now

### Reminders

### DIFF
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -35,7 +35,8 @@ from distutils.util import strtobool
 import anyconfig
 from argus.client.sct.types import Package
 from packaging import version
-from pydantic import BaseModel, Field, ConfigDict, fields as pydantic_fields
+from pydantic import BaseModel, Field, ConfigDict, RootModel, fields as pydantic_fields, model_validator
+from pydantic.types import confloat
 from typing_extensions import Annotated
 from pydantic.functional_validators import BeforeValidator
 from pydantic.fields import FieldInfo
@@ -262,6 +263,52 @@ def dict_or_str(value: dict | str | None) -> dict | None:
 
 
 DictOrStr = Annotated[dict | str, BeforeValidator(dict_or_str)]
+
+
+class AdaptiveTimeoutMultipliers(RootModel):
+    """Per-operation multipliers for adaptive timeouts.
+
+    Keys must be valid operation names from Operations enum (operation.value[0]),
+    e.g. decommission, remove_node, new_node, repair, rebuild, etc.
+    Missing keys default to multiplier 1.
+
+    YAML config example::
+
+        adaptive_timeout_multipliers:
+          decommission: 4
+          new_node: 4
+          remove_node: 4
+
+    Environment variable examples:
+
+        SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS="{'decommission': 2, 'new_node': 3}"
+
+    Or using dot-notation (same pattern as SCT_STRESS_IMAGE.*):
+
+        SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS.decommission=4
+        SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS.new_node=3
+    """
+
+    root: dict[str, confloat(gt=0)] = Field(default_factory=dict)
+
+    @model_validator(mode="before")
+    @classmethod
+    def _validate_operations(cls, value):
+        if not isinstance(value, dict):
+            return value
+
+        # cyclic-import: Operations imports from sct_config indirectly via cluster
+        from sdcm.utils.adaptive_timeouts import Operations  # noqa: PLC0415
+
+        valid_keys = {op.value[0] for op in Operations}
+        for key in value.keys():
+            if key not in valid_keys:
+                raise ValueError(f"Unknown operation key '{key}'. Valid keys: {sorted(valid_keys)}")
+        return value
+
+    def get_multiplier(self, operation_key: str) -> float:
+        """Return multiplier for the given operation key, or 1.0 if not configured."""
+        return float(self.root.get(operation_key, 1.0))
 
 
 def dict_or_str_or_pydantic(value: dict | str | BaseModel | None) -> dict | BaseModel | None:
@@ -1262,6 +1309,20 @@ class SCTConfiguration(BaseModel):
     )
     adaptive_timeout_store_metrics: Boolean = SctField(
         description="Store adaptive timeout metrics in Argus. Disabled for performance tests only.",
+    )
+    adaptive_timeout_multipliers: Annotated[AdaptiveTimeoutMultipliers, BeforeValidator(dict_or_str_or_pydantic)] = (
+        SctField(
+            description="Optional dict of adaptive-timeout multipliers keyed by operation name "
+            "(from Operations enum value[0], e.g. decommission, remove_node, new_node, repair, etc.). "
+            "If the current operation key is absent, multiplier 1.0 is used.<br>"
+            "YAML example:<br>"
+            "adaptive_timeout_multipliers:<br>"
+            "  decommission: 4<br>"
+            "  new_node: 2<br>"
+            "Environment variable examples:<br>"
+            "SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS=\"{'decommission': 4, 'new_node': 2}\"<br>"
+            "Or dot-notation: SCT_ADAPTIVE_TIMEOUT_MULTIPLIERS.decommission=4",
+        )
     )
 
     # Google Compute Engine options
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -34,7 +34,6 @@ from uuid import uuid4
 from functools import partial, wraps, cache
 import threading
 import signal
-import json
 import botocore
 import yaml
 import pytest
@@ -438,7 +437,7 @@ class ClusterTester(unittest.TestCase):
                 backend=self.params.get("cluster_backend"),
             )
             self.log.info("sct_runner info in Argus TestRun is updated")
-            self.test_config.argus_client().sct_submit_config(name="sct_config", content=json.dumps(dict(self.params)))
+            self.test_config.argus_client().sct_submit_config(name="sct_config", content=self.params.model_dump_json())
             self.log.info("Submitted SCTConfiguration to Argus.")
         except ArgusClientError:
             self.log.error("Failed to submit data to Argus", exc_info=True)
--- a/unit_tests/unit/test_tester.py
+++ b/unit_tests/unit/test_tester.py
@@ -10,13 +10,16 @@
 # See LICENSE for more details.
 #
 # Copyright (c) 2020 ScyllaDB
+import json
 import logging
 import time
+import types
 import unittest.mock
 from unittest.mock import MagicMock
 
 import pytest
 
+from sdcm.sct_config import SCTConfiguration
 from sdcm.sct_events import Severity
 from sdcm.sct_events.base import SctEvent
 from sdcm.sct_events.health import ClusterHealthValidatorEvent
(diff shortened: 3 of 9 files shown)


aleksbykov:
@pehala , @vponomaryov , @soyacz can you take a look.

unit_tests/unit/test_tester.py
```diff
@@ -617,3 +594,75 @@ def test_emr_clean_resources_no_emr_cluster(self):
...
+        assert parsed["adaptive_timeout_multipliers"] == {"decommission": 5.0, "new_node": 5.0}
+
+    def test_sct_submit_config_contains_expected_fields(self, tester_with_argus):
+        """Verify that model_dump_json() output contains essential config fields."""
+        tester, mock_argus_client = tester_with_argus
+
+        tester.init_argus_run()
+
+        call_kwargs = mock_argus_client.sct_submit_config.call_args
+        content = call_kwargs.kwargs.get("content") or call_kwargs[1].get("content") or call_kwargs[0][1]
```
pehala: You should always know what was called no?

unit_tests/unit/test_tester.py
```diff
@@ -617,3 +594,75 @@ def test_emr_clean_resources_no_emr_cluster(self):
...
+        # adaptive_timeout_multipliers must be properly serialized
+        assert parsed["adaptive_timeout_multipliers"] == {"decommission": 5.0, "new_node": 5.0}
+
+    def test_sct_submit_config_contains_expected_fields(self, tester_with_argus):
+        """Verify that model_dump_json() output contains essential config fields."""
+        tester, mock_argus_client = tester_with_argus
+
+        tester.init_argus_run()
+
+        call_kwargs = mock_argus_client.sct_submit_config.call_args
```
pehala: Can you assert exact call params instead?

unit_tests/unit/test_tester.py
```diff
@@ -617,3 +594,75 @@ def test_emr_clean_resources_no_emr_cluster(self):
...
+        Previously json.dumps(dict(self.params)) raised ValueError: Circular
+        reference detected when AdaptiveTimeoutMultipliers (a Pydantic RootModel)
+        was in the config.
+        """
+        tester, mock_argus_client = tester_with_argus
+
+        tester.init_argus_run()
+
+        # Verify sct_submit_config was called
+        mock_argus_client.sct_submit_config.assert_called_once()
```
pehala: Can you assert exact call params instead?

unit_tests/unit/test_tester.py
```diff
@@ -617,3 +594,75 @@ def test_emr_clean_resources_no_emr_cluster(self):
         """Test that clean_resources handles None emr_cluster gracefully."""
         tester = self._make_tester(self._make_params("destroy"))
         self._run_clean_resources(tester, critical_events=False)
+
+
+class TestInitArgusRun:
```
pehala: no classes...again

aleksbykov: I used same pattern for the file. So all tests were wriiten in same manner. But ok, i will rewrite then with functions

pehala: Fair enough, but we want to get rid of them regardless. I am surprised they are still used here

unit_tests/unit/test_tester.py
aleksbykov: Yes. Fixed

unit_tests/unit/test_tester.py
aleksbykov: fixed

unit_tests/unit/test_tester.py
aleksbykov: fixed

unit_tests/unit/test_tester.py
aleksbykov: done

aleksbykov:
v2: Change tests to functions. added more mock for untested calls of argus(not a scope of the pr).

sdcm/sct_config.py
```diff
@@ -264,6 +265,52 @@ def dict_or_str(value: dict | str | None) -> dict | None:
...
+        valid_keys = {op.value[0] for op in Operations}
+        for key in value.keys():
+            if key not in valid_keys:
+                raise ValueError(f"Unknown operation key '{key}'. Valid keys: {sorted(valid_keys)}")
+        return value
+
+    def get_multiplier(self, operation_key: str) -> float:
+        """Return multiplier for the given operation key, or 1.0 if not configured."""
+        return float(self.root.get(operation_key, 1.0))
+
```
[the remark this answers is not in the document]
aleksbykov: This value is not expected to be in config

unit_tests/unit/test_tester.py
```diff
@@ -617,3 +594,97 @@ def test_emr_clean_resources_no_emr_cluster(self):
...
+        },
+    )
+    monkeypatch.setattr("sdcm.tester.get_git_commit_id", lambda: "abc123")
+    monkeypatch.setattr("sdcm.tester.get_job_name", lambda: "test-job")
+    monkeypatch.setattr("sdcm.tester.get_job_url", lambda: "http://jenkins/job/1")
+    monkeypatch.setattr("sdcm.tester.get_username", lambda: "test-user")
+    monkeypatch.setattr("sdcm.tester.get_sct_runner_ip", lambda: "1.2.3.4")
+    monkeypatch.setattr("sdcm.tester.get_my_ip", lambda: "10.0.0.1")
+
+    tester.init_argus_run = ClusterTester.init_argus_run.__get__(tester)
```
pehala: nit: Why do you use get metamethod? Wont `ClusterTester.init_argus_run[tester]` work as well?

unit_tests/unit/test_tester.py
```diff
@@ -617,3 +594,97 @@ def test_emr_clean_resources_no_emr_cluster(self):
...
+            "upstream.url": "git@github.com:test",
+            "branch.upstream": "origin/main",
+        },
+    )
+    monkeypatch.setattr("sdcm.tester.get_git_commit_id", lambda: "abc123")
+    monkeypatch.setattr("sdcm.tester.get_job_name", lambda: "test-job")
+    monkeypatch.setattr("sdcm.tester.get_job_url", lambda: "http://jenkins/job/1")
+    monkeypatch.setattr("sdcm.tester.get_username", lambda: "test-user")
+    monkeypatch.setattr("sdcm.tester.get_sct_runner_ip", lambda: "1.2.3.4")
+    monkeypatch.setattr("sdcm.tester.get_my_ip", lambda: "10.0.0.1")
```
pehala, line 648: nit: These methods live in different modules, cannot we mock them in a better way? ideally in a waythat is sharable via fixture and doesnt use setattr on an import?

sdcm/tester.py
```diff
@@ -441,7 +437,7 @@ def init_argus_run(self):
                 backend=self.params.get("cluster_backend"),
             )
             self.log.info("sct_runner info in Argus TestRun is updated")
-            self.test_config.argus_client().sct_submit_config(name="sct_config", content=json.dumps(dict(self.params)))
+            self.test_config.argus_client().sct_submit_config(name="sct_config", content=self.params.model_dump_json())
```
pehala, line 440: nice, this was a latent bug. This is much better. I suspect this was the cause of the original failure right?

pehala (dismissed):
LGTM, small test nits can be addressed separately or not at all

sdcm/tester.py
aleksbykov: yes

unit_tests/unit/test_tester.py
aleksbykov: No, we cannot fully avoid setattr-style patching here -- Python's import semantics require patching at the import site (sdcm.tester) not the source module, and these are free-standing functions (not methods on an object we control). However, we can improve it two ways:
 
1. Extract into a reusable fixture in conftest.py so any test needing these mocks just declares mock_tester_env_calls as a dependency.
2. Use object-based setattr (monkeypatch.setattr(tester_module, "get_git_status_info", ...)) instead of string paths -- this gives immediate AttributeError if the function is ever renamed, rather than silently passing with real I/O.

The only way to eliminate setattr entirely would be refactoring init_argus_run to accept a dependency-injection parameter (e.g., a context object), but that's invasive for a single method and adds test-only complexity to production code.

unit_tests/unit/test_tester.py
aleksbykov: No, the method doesn't have this possibility. I made it a bit more readable.

unit_tests/unit/test_tester.py
aleksbykov:
```python
import sdcm.tester as _tester_module
 

@pytest.fixture()
def mock_tester_env_calls(monkeypatch):
    """Mock I/O utility functions imported by sdcm.tester.

    Patches git, CI, and network helpers that perform real subprocess/network
    calls when invoked from ClusterTester methods (e.g. init_argus_run).
    Uses object-based setattr so that renaming an import raises AttributeError
    immediately rather than silently leaving real I/O unmocked.
    """
    monkeypatch.setattr(
        _tester_module,
        "get_git_status_info",
        lambda: {
            "branch.oid": "abc123",
            "upstream.url": "git@github.com:test",
            "branch.upstream": "origin/main",
        },
    )
    monkeypatch.setattr(_tester_module, "get_git_commit_id", lambda: "abc123")
    monkeypatch.setattr(_tester_module, "get_job_name", lambda: "test-job")
    monkeypatch.setattr(_tester_module, "get_job_url", lambda: "http://jenkins/job/1")
    monkeypatch.setattr(_tester_module, "get_username", lambda: "test-user")
    monkeypatch.setattr(_tester_module, "get_sct_runner_ip", lambda: "1.2.3.4")
    monkeypatch.setattr(_tester_module, "get_my_ip", lambda: "10.0.0.1")
```

this fixture could be added and used

cezarmoise:
Previous PR had a configuration `configurations/adaptive_timeout_multipliers.yaml` and added it to `jenkins-pipelines/oss/tier1/longevity-mv-si-4days-streaming.jenkinsfile`. Why are those changes removed here?

aleksbykov:
> Previous PR had a configuration `configurations/adaptive_timeout_multipliers.yaml` and added it to `jenkins-pipelines/oss/tier1/longevity-mv-si-4days-streaming.jenkinsfile`. Why are those changes removed here?

Thanks. Added them.

cezarmoise (approved):
LGTM

unit_tests/unit/test_tester.py
pehala: It is extremely ugly but lets keep it as is

pehala (approved):

vponomaryov (approved):
LGTM

PR #14931, opened 2026-06-04
improvement(jenkins): add Argus link and stage tracking to byoLongevityPipeline

fruch:
## Summary

Aligned `byoLongevityPipeline.groovy` with other pipelines (longevityPipeline, managerPipeline, jepsenPipeline) by adding:

- **Create Argus Test Run** stage after Checkout — registers the test run and adds the Argus link to the Jenkins build description (sidebar link)
- **Finish Argus Test Run** stage — finalizes the Argus test run status
- **`completed_stages` tracking** on all stages
- **`post { always { ... } }` block** — retries any failed stages (collect_logs, clean_resources, report_to_argus, send_email, clean_sct_runner) ensuring cleanup always happens even on pipeline failure

## Why

The BYO longevity pipeline was the only pipeline missing Argus integration. Without it, there was no Argus link in the Jenkins build sidebar and no guaranteed cleanup on failure.

## Manual Testing Required

- [ ] Run a BYO longevity job and verify the Argus link appears in the Jenkins build description
- [ ] Verify the Finish Argus Test Run stage completes successfully
- [ ] Verify post-failure retry logic works (e.g., abort mid-run and confirm cleanup stages execute)

### DIFF
--- a/vars/byoLongevityPipeline.groovy
+++ b/vars/byoLongevityPipeline.groovy
@@ -1,5 +1,6 @@
 #!groovy
 
+def completed_stages = [:]
 
 def call() {
 
@@ -17,7 +18,7 @@ def call() {
             AWS_ACCESS_KEY_ID     = credentials('qa-aws-secret-key-id')
             AWS_SECRET_ACCESS_KEY = credentials('qa-aws-secret-access-key')
             SCT_TEST_ID = UUID.randomUUID().toString()
-		}
+        }
 
          parameters {
             separator(name: 'SCT_CONFIG', sectionHeader: 'SCT Repository Configuration')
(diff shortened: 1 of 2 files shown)


vars/byoLongevityPipeline.groovy
```diff
@@ -1,5 +1,6 @@
 #!groovy
 
+def completed_stages = [:]
```
vponomaryov, line 3: Just realized that here you update the obsolete groovy file.

The BYO is supported by standard longevity, upgrade and performance pipelines.

fruch: It's not obselate if it's still in the code and people are still using it

I remember I don't like it much, I don't recall we defined it as obsolete at any point (no that I'm against it,  just we need to guide those using it into recreating thier jobs)

vponomaryov: It is technically obsolete, because we can use standard pipelines instead.
This one has no benefits over the standard pipelines...

fruch:
need to deprecate the byo pipeline, hence closing this

PR #14933, opened 2026-06-04
fix(azure-provision): comprehensive diagnostics collection on provisioning failure

fruch:
## Summary

Comprehensive Azure VM diagnostics collection when node provisioning fails (SSH timeout).
Root cause investigation of SCT-434: `longevity-10gb-3h-azure-test` with `Standard_L8s_v4` where db-node-1 SSH never connected.

## Problem

When Azure VM SSH connection fails during `init()`, the node log directory remains completely empty — making post-mortem debugging impossible. Azure v4 instance types need longer NIC propagation time.

## Changes (7 separate commits for easier backporting)

1. **Collect VM diagnostics on SSH failure** — `get_console_output()`, `collect_azure_vm_diagnostics()`, `_collect_diagnostics_on_failure()`, NIC delay 5s→10s
2. **Collect Activity Log** — Query Azure Monitor for errors/warnings in the resource group (quota, throttling, deployment failures)
3. **Collect effective NSG rules** — Get firewall rules applied to the NIC (SSH port blocked?)
4. **Query Resource Health API** — Detect platform-level issues (host crashes, maintenance)
5. **Save boot diagnostics screenshot** — BMP console screenshot for kernel panics / boot hangs
6. **Extend SSH timeout for Azure** — Override `wait_ssh_up` with 12-min minimum timeout for Azure VMs
7. **Verify NIC provisioning state** — Warn if NIC not in "Succeeded" state before VM creation

## Validation Required

⚠️ **This PR needs a few runs with the Azure sanity test to validate all diagnostics methods are working correctly and collecting the expected data.**

## Manual Testing Required

- [ ] Run Azure provision test and verify diagnostics are collected on failure
- [ ] Verify boot diagnostics screenshot is saved to log directory
- [ ] Confirm Activity Log captures errors when quota is exceeded
- [ ] Test with Standard_L8s_v4 instance type to validate extended SSH timeout

## Backends Affected
- [x] Azure

[SCT-434]

[SCT-434]: https://scylladb.atlassian.net/browse/SCT-434?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdcm/cluster_azure.py
+++ b/sdcm/cluster_azure.py
@@ -108,6 +108,9 @@ class AzureNode(cluster.BaseNode):
                 self.log.error(
                     "Azure Activity Log errors/warnings for %s: %s", self.name, json.dumps(activity_log, indent=2)
                 )
+            nsg_rules = self.collect_effective_nsg_rules()
+            if nsg_rules:
+                self.log.error("Azure effective NSG rules for %s: %s", self.name, json.dumps(nsg_rules, indent=2))
         except Exception as exc:  # noqa: BLE001
             self.log.warning("Failed to collect diagnostics on failure for %s: %s", self.name, exc)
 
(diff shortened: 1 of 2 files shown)


soyacz (changes requested):
this will not work - problem is during `wait_cloud_init_completes` which is not during `init` IIRC.
I propose to extend `sdcm.provision.provisioner.VmInstance` with new method `collect_diagnostics` which will call proper method in `sdcm.provision.provisioner.Provisioner` - by default noop, and for `sdcm.provision.azure.provisioner.AzureProvisioner` should implement things  like getting console output  collect_azure_vm_diagnostics etc.

This does not belong to `cluster_azure` - it's provisioning issue and should be handled there. Btw. Later we'll be able to add support for GCE/OCI backends too

fruch:
Closing - diagnostics approach being reworked.

PR #14955, opened 2026-06-07
fix(events): prevent 30+ minute shutdown hangs in post-test cleanup

fruch:
Fixes: [SCT-443](https://scylladb.atlassian.net/browse/SCT-443)

## Problem

During local development, every test run ended with an unnecessary 30+ minute hang (or indefinite block) during teardown and shutdown. This made the develop-run-debug cycle painfully slow — you'd finish a test, see it pass or fail, then wait half an hour before getting your terminal back. The root cause was several interrelated issues in the event system and teardown logic:

1. **EventsDevice.stop()** called `join(timeout)` then `join_thread()` — if the process didn't exit in time, `join_thread()` would block forever waiting for the queue feeder to drain a pipe nobody was reading.

2. **BaseEventsProcess.stop()** used `assert False` when a process was still alive after timeout — crashing the teardown instead of recovering.

3. **EventsDevice.run()** drain loop (`while not queue.empty()`) could spin indefinitely if no ZMQ subscriber was consuming events.

4. **Coredump thread join** waited 3600s (1 hour) — far longer than any real coredump collection.

5. **Node task threads** (DbLogReader, alert manager, journal thread) continued feeding events after consumers stopped, causing QueueFeederThread to block on write to a dead pipe.

These hangs also affect CI — jobs would sit idle during teardown, wasting runner time and delaying pipeline feedback.

## Fixes

Each commit is independently revertable:

- **fix(cluster)**: Reduce coredump thread join from 3600s to 300s with warning if still alive
- **fix(events)**: Replace `assert False` with `LOGGER.error` + kill escalation in BaseEventsProcess.stop()
- **fix(events)**: Bound EventsDevice drain loop to 5 seconds after `_running` is cleared
- **fix(events)**: Use `cancel_join_thread()` instead of `join_thread()` + kill escalation via `super().is_alive()`
- **fix(tester)**: Call `_stop_all_node_task_threads()` in the early-return path when `execute_post_behavior=false`

## Key Design Decisions

- **`super().is_alive()` over `self.is_alive()`**: EventsDevice.is_alive() override always returns False after `_running.clear()`, so we must bypass it to check the actual process state.
- **`cancel_join_thread()` over `join_thread()`**: Prevents pipe-full deadlock when queue has unread data.
- **`kill()` not `terminate()`**: `suppress_interrupt()` context manager catches SIGTERM; SIGKILL bypasses it.

## Testing

- All 39 events unit tests pass
- Ruff clean on all modified files
- Verified Docker teardown completes in <6s (previously 30+ min)
- Zero zombie processes after shutdown

[SCT-443]: https://scylladb.atlassian.net/browse/SCT-443?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdcm/sct_events/events_device.py
+++ b/sdcm/sct_events/events_device.py
@@ -112,7 +112,13 @@ class EventsDevice(multiprocessing.Process):
 
                 time.sleep(self.start_delay)
 
-                while self._running.is_set() or not self._queue.empty():
+                drain_deadline = None
+                while True:
+                    if not self._running.is_set():
+                        if drain_deadline is None:
+                            drain_deadline = time.monotonic() + 5.0
+                        if time.monotonic() >= drain_deadline:
+                            break
                     try:
                         event = self._queue.get(timeout=self.pub_queue_wait_timeout)
                         try:
@@ -128,7 +134,9 @@ class EventsDevice(multiprocessing.Process):
                             LOGGER.error("EventsDevice failed to verify delivery of %s", pickle.loads(event))
                         time.sleep(self.pub_queue_events_rate)
                     except queue.Empty:
-                        pass
+                        if not self._running.is_set():
+                            break
+                        continue
 
     def publish_event(self, event, timeout=PUBLISH_EVENT_TIMEOUT) -> None:
         with verbose_suppress("%s: failed to write %s to %s", self, event, self.raw_events_log):
--- a/sdcm/sct_events/events_device.py
+++ b/sdcm/sct_events/events_device.py
@@ -91,6 +91,12 @@ class EventsDevice(multiprocessing.Process):
     def stop(self, timeout: Optional[float] = None) -> None:
         self._running.clear()
         self.join(timeout)
+        if super().is_alive():
+            LOGGER.warning("EventsDevice did not stop within timeout, killing")
+            self.kill()
+            self.join(5)
+        self._queue.cancel_join_thread()
+        self._queue.close()
 
     @property
     def subscribe_address(self) -> str:
--- a/sdcm/sct_events/events_device.py
+++ b/sdcm/sct_events/events_device.py
@@ -222,7 +222,7 @@ class EventsDevice(multiprocessing.Process):
                 yield obj.base, obj
 
     def is_alive(self) -> bool:
-        return self._running.is_set()
+        return self._running.is_set() and super().is_alive()
 
 
 start_events_main_device = partial(start_events_process, EVENTS_MAIN_DEVICE_ID, EventsDevice)
--- a/sdcm/sct_events/events_device.py
+++ b/sdcm/sct_events/events_device.py
@@ -91,6 +91,8 @@ class EventsDevice(multiprocessing.Process):
     def stop(self, timeout: Optional[float] = None) -> None:
         self._running.clear()
         self.join(timeout)
+        # self.is_alive() returns self._running.is_set(), which is already False here
+        # (cleared above). Call super() to check the actual OS-level process liveness.
         if super().is_alive():
             LOGGER.warning("EventsDevice did not stop within timeout, killing")
             self.kill()
--- a/sdcm/sct_events/events_device.py
+++ b/sdcm/sct_events/events_device.py
@@ -15,6 +15,7 @@ import time
 import queue
 import ctypes
 import pickle
+import signal
 import logging
 import multiprocessing
 from typing import Optional, Generator, Any, Tuple, Callable, cast, Dict
@@ -91,10 +92,12 @@ class EventsDevice(multiprocessing.Process):
     def stop(self, timeout: Optional[float] = None) -> None:
         self._running.clear()
         self.join(timeout)
-        # self.is_alive() returns self._running.is_set(), which is already False here
-        # (cleared above). Call super() to check the actual OS-level process liveness.
         if super().is_alive():
-            LOGGER.warning("EventsDevice did not stop within timeout, killing")
+            LOGGER.warning("EventsDevice did not stop within timeout, sending SIGTERM")
+            self.terminate()
+            self.join(5)
+        if super().is_alive():
+            LOGGER.warning("EventsDevice still alive after SIGTERM, sending SIGKILL")
             self.kill()
             self.join(5)
         self._queue.cancel_join_thread()
@@ -107,7 +110,10 @@ class EventsDevice(multiprocessing.Process):
         raise RuntimeError("EventsDevice is not ready to send events.")
 
     def run(self):
-        with suppress_interrupt(), verbose_suppress("EventsDevice failed"):
+        # SIGTERM handler enables cooperative shutdown: terminate() sends SIGTERM which
+        # triggers _running.clear(), letting the loop drain and exit gracefully.
+        signal.signal(signal.SIGTERM, lambda *_: self._running.clear())
+        with verbose_suppress("EventsDevice failed"):
             with zmq.Context() as ctx, ctx.socket(zmq.PUB) as pub, ctx.socket(zmq.SUB) as sub:
                 self._sub_port.value = pub.bind_to_random_port("tcp://*")
                 self._running.set()
@@ -222,7 +228,14 @@ class EventsDevice(multiprocessing.Process):
                 yield obj.base, obj
 
     def is_alive(self) -> bool:
-        return self._running.is_set() and super().is_alive()
+        if not self._running.is_set():
+            return False
+        try:
+            return super().is_alive()
+        except AssertionError:
+            # CPython's BaseProcess.is_alive() asserts self._parent_pid == os.getpid().
+            # If called from a forked child, fall back to the shared event flag.
+            return True
 
 
 start_events_main_device = partial(start_events_process, EVENTS_MAIN_DEVICE_ID, EventsDevice)
--- a/sdcm/sct_events/events_device.py
+++ b/sdcm/sct_events/events_device.py
@@ -93,11 +93,7 @@ class EventsDevice(multiprocessing.Process):
         self._running.clear()
         self.join(timeout)
         if super().is_alive():
-            LOGGER.warning("EventsDevice did not stop within timeout, sending SIGTERM")
-            self.terminate()
-            self.join(5)
-        if super().is_alive():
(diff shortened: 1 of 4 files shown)


sdcm/sct_events/events_device.py
```diff
@@ -91,6 +91,12 @@ def raw_events_log(self) -> Path:
     def stop(self, timeout: Optional[float] = None) -> None:
         self._running.clear()
         self.join(timeout)
+        if super().is_alive():
```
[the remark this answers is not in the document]
fruch: Added the comment in 1d0f293. Thanks for catching this — the `is_alive()` override trap is exactly the kind of thing that bites future readers.

sdcm/sct_events/events_processes.py
```diff
@@ -100,8 +100,10 @@ def stop(self, timeout: float = None) -> None:
         LOGGER.debug("Waiting for events process %s to stop", events_process_class_name)
         self.join(timeout)
         if self.is_alive():
-            LOGGER.error("Events process %s is still alive after timeout", events_process_class_name)
-            assert False, f"Events process {events_process_class_name} is still alive after timeout"
+            LOGGER.error("Events process %s is still alive after %s timeout", events_process_class_name, timeout)
+            if hasattr(self, "kill"):
+                self.kill()
```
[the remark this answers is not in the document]
fruch: Added the inline comment in 1d0f293. Good call — the Process/Thread mixin duality is non-obvious without it.

sdcm/sct_events/events_device.py
```diff
@@ -91,6 +91,14 @@ def raw_events_log(self) -> Path:
     def stop(self, timeout: Optional[float] = None) -> None:
         self._running.clear()
         self.join(timeout)
+        # self.is_alive() returns self._running.is_set(), which is already False here
+        # (cleared above). Call super() to check the actual OS-level process liveness.
```
pehala: How about clearing it later then? I do not think we should set running to False if we didnt fully stopped the OS thread.

sdcm/sct_events/events_processes.py
```diff
@@ -100,8 +100,10 @@ def stop(self, timeout: float = None) -> None:
         LOGGER.debug("Waiting for events process %s to stop", events_process_class_name)
         self.join(timeout)
         if self.is_alive():
-            LOGGER.error("Events process %s is still alive after timeout", events_process_class_name)
-            assert False, f"Events process {events_process_class_name} is still alive after timeout"
+            LOGGER.error("Events process %s is still alive after %s timeout", events_process_class_name, timeout)
+            if hasattr(self, "kill"):  # threading.Thread has no kill(); only escalate for Process-backed subclasses
+                self.kill()
```
pehala: I think isinstance check would be cleaner. hasttr doesnt discriminate and if we have e.g Custom thread class with field kill it would trigger as well.

sdcm/tester.py
```diff
@@ -3809,10 +3770,24 @@ def show_alive_threads(self):
...
+            if not cluster:
+                continue
+            for node in cluster.nodes:
+                node.stop_task_threads()
+
     @silence()
     def clean_resources(self):
         if not self.params.get("execute_post_behavior"):
             self.log.info("Resources will continue to run")
+            self._stop_all_node_task_threads()
```
pehala: Shouldnt we stop always? I think we want to stop the threads regardless of the execute_post_behaviour as we want to stop the SCT even if the resources stay alive

pehala (commented):
> **`kill()` not `terminate()`**: `suppress_interrupt()` context manager catches SIGTERM; SIGKILL bypasses it.

Seems like we should fix or update suppress_interrupt then or not use it there.

Overall I am not convinced this is correct, I believe it fixes some of the shutdown issues but I worry it might introduce more and/or we are leaving many more at the table. Shutdown issue are pain to get correctly.

Would it be possible to reproduce the shutdown problems in integration test, without mocking everything?

sdcm/sct_events/events_device.py
fruch: it's the signal to stop the inner loop, maybe the fact we override the is_alive function isn't the best thing todo

but we can't move this one

sdcm/sct_events/events_device.py
fruch: maybe we should do `self._running.is_set() and super().is_alive()` ?

sdcm/tester.py
fruch: probably this one assume this is called as part of the other destroy sequences below

sdcm/sct_events/events_device.py
fruch: We can't move `_running.clear()` after `join()` because it's the signal that makes the `run()` loop exit (the loop checks `self._running.is_set()` to decide when to break). Without clearing first, `join()` would block forever.

However, I've addressed the underlying concern by adopting fruch's suggestion: `is_alive()` now returns `self._running.is_set() and super().is_alive()`, so the API is more truthful — it reflects both the logical intent AND the actual OS-level process state. Fixed in 68bd01b.

sdcm/sct_events/events_processes.py
fruch: Good point — switched to `isinstance(self, multiprocessing.Process)` which is explicit and won't trigger on custom attributes. Fixed in 68bd01b.

sdcm/tester.py
fruch: Agreed — moved `_stop_all_node_task_threads()` to run unconditionally at the top of `clean_resources()`, before the `execute_post_behavior` check. Task threads should always be stopped before any cleanup (whether we're destroying resources or keeping them alive) to prevent event pipeline deadlocks. Fixed in 68bd01b.

sdcm/sct_events/events_device.py
fruch: Done — `is_alive()` now returns `self._running.is_set() and super().is_alive()`. This makes it reflect both "we intend it to run" AND "it's actually alive at the OS level". In `stop()` we still need `super().is_alive()` directly for the kill escalation (since `_running` is already cleared at that point), but the override itself is now more correct for external callers. Fixed in 68bd01b.

fruch: https://github.com/scylladb/scylla-cluster-tests/pull/14955#discussion_r3371542367 seems to suggest it wasn't a very good idea

sdcm/sct_events/events_device.py
```diff
@@ -206,7 +222,7 @@ def outbound_events(
                 yield obj.base, obj
 
     def is_alive(self) -> bool:
-        return self._running.is_set()
+        return self._running.is_set() and super().is_alive()
```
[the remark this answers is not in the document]
fruch: Fixed in a68437c. `is_alive()` now wraps `super().is_alive()` in try/except AssertionError and falls back to `self._running.is_set()` (which is `True` at that point). This handles the forked-child case safely without crashing.

sdcm/sct_events/events_device.py
fruch: Addressed in a68437c — took your suggestion and fixed `suppress_interrupt` usage rather than working around it:

**What changed:**
- Removed `suppress_interrupt()` from `EventsDevice.run()` entirely
- Added a SIGTERM signal handler (`signal.signal(SIGTERM, lambda *_: self._running.clear())`) — this makes `terminate()` trigger cooperative shutdown via the existing drain loop
- Updated `stop()` escalation: `_running.clear()` → `join(timeout)` → `terminate()` (SIGTERM) → `join(5)` → `kill()` (SIGKILL) as last resort
- `suppress_interrupt` is kept only in `outbound_events()` where it protects subscriber threads from mid-delivery crashes (different concern)

**Re: integration tests without mocking** — agreed this is important. The existing `test_publish_subscribe` already exercises `stop(timeout=1)` on a real process, and it passes with this change. I will add a focused integration test that:
1. Starts EventsDevice, sends events
2. Calls `stop(timeout=0.1)` to force the terminate→kill escalation
3. Verifies the process actually exits within a bounded time (no hang)

This can run without any external services — just real multiprocessing.Process lifecycle.

sdcm/sct_events/events_device.py
```diff
@@ -91,6 +92,16 @@ def raw_events_log(self) -> Path:
...
+        if super().is_alive():
+            LOGGER.warning("EventsDevice did not stop within timeout, sending SIGTERM")
+            self.terminate()
+            self.join(5)
+        if super().is_alive():
+            LOGGER.warning("EventsDevice still alive after SIGTERM, sending SIGKILL")
+            self.kill()
+            self.join(5)
+        self._queue.cancel_join_thread()
+        self._queue.close()
```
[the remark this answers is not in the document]
fruch: Valid point. `_running.clear()` already triggers the cooperative drain loop exit, so by the time `terminate()` fires, the SIGTERM handler just re-clears an already-cleared event — a no-op. And if the process is blocked in a C-level ZMQ call, the handler can't run anyway.

Will simplify `stop()` to go directly to `kill()` after the cooperative join fails. Keeping the SIGTERM handler in `run()` though — it still enables cooperative shutdown for external `terminate()` calls (e.g. process manager sending SIGTERM directly).

sdcm/tester.py
```diff
@@ -3770,8 +3770,22 @@ def show_alive_threads(self):
...
+
+        Prevents daemon threads from feeding events into a dead queue
+        after event consumers are stopped, which would cause the process
+        to hang on exit (QueueFeederThread blocked on write to dead pipe).
+        """
+        for cluster in [self.db_cluster, self.loaders, self.monitors]:
+            if not cluster:
+                continue
+            for node in cluster.nodes:
+                node.stop_task_threads()
```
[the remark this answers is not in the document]
fruch: Valid observation. The race exists but is already mitigated: `verbose_suppress` in `publish_event` catches the `ValueError: is closed` so teardown doesn't crash — late events are silently dropped (acceptable during shutdown).

A proper fix (joining task threads after signalling) would add teardown latency and change cleanup ordering across all backends. Deferring to a follow-up PR since it's orthogonal to this fix (this PR addresses the EventsDevice process hang, not the producer side).

fruch:
integration test failure is not part of this PR, it caused by recent baremetal PR. gonna fix separately

PR #14960, opened 2026-06-08
feature(scylla-doctor): add analysis phase support for full edition

juliayakovlev:
Add support for running the Scylla Doctor analysis phase when scylla_doctor_edition=full. The analysis phase loads previously collected vitals and runs analyzers to produce findings and recommendations about cluster health.

### Testing

- [x] [sd-release-gating](https://jenkins.scylladb.com/job/scylla-doctor/job/sd-release-gating/14/)
- [x] [GCE - artifacts-debian12-test with Scylla version parameter](https://argus.scylladb.com/tests/scylla-cluster-tests/dda06b89-49ca-4930-a1cc-5c3a24046887)
- [x] [GCE - artifacts-debian12-test with Scylla repo parameter](https://argus.scylladb.com/tests/scylla-cluster-tests/e0d697d8-1b34-41d9-8f9c-5398659a2a5c)
- [x] [AWS - artifacts-amazon2023-arm-test with Scylla repo parameter](https://argus.scylladb.com/tests/scylla-cluster-tests/7bd1281c-a2f0-49d8-9296-8946e99d1a50)
- [x] [Docker - artifacts-docker-test](https://argus.scylladb.com/tests/scylla-cluster-tests/9ed6aef3-a885-4fcc-8552-e107cf93f3c9)

### Reminders

### DIFF
--- a/artifacts_test.py
+++ b/artifacts_test.py
@@ -358,6 +358,13 @@ class ArtifactsTest(ClusterTester):
 
         backend = self.params.get("cluster_backend")
 
+        if self.params.get("run_scylla_doctor_only"):
+            with self.logged_subtest("verify node health"):
+                self.verify_node_health()
+            with self.logged_subtest("check scylla_doctor results"):
+                self.run_scylla_doctor()
+            return
+
         if backend == "aws":
             with self.logged_subtest("check ENA support"):
                 assert self.node.ena_support, "ENA support is not enabled"
--- a/artifacts_test.py
+++ b/artifacts_test.py
@@ -563,6 +563,9 @@ class ArtifactsTest(ClusterTester):
                 scylla_doctor.run_scylla_doctor_and_collect_results()
                 scylla_doctor.analyze_vitals()
                 scylla_doctor.analyze_and_verify_results()
+                if scylla_doctor.is_full_edition:
+                    scylla_doctor.run_analysis_phase()
+                    scylla_doctor.analyze_and_verify_analysis_results()
 
     def get_email_data(self):
         self.log.info("Prepare data for email")
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -1957,6 +1957,15 @@ Scylla Doctor version to use for artifact tests. Set to specific version (e.g., 
 * appendable
 
 
+## **run_scylla_doctor_only** / SCT_RUN_SCYLLA_DOCTOR_ONLY
+
+When true, the artifact test runs only the Scylla Doctor validation<br>(install, collect vitals, analyze, verify) and skips all other artifact checks<br>such as stop/start, cassandra-stress, housekeeping, etc. Useful for fast SD<br>release gating. Implies run_scylla_doctor=true.
+
+**default:** N/A
+
+**type:** bool
+
+
 ## **scylla_doctor_edition** / SCT_SCYLLA_DOCTOR_EDITION
 
 Scylla Doctor edition to use. Allowed values: 'basic', 'full'.<br>'basic' fetches the free/open-source edition via HTTP.<br>'full' fetches the full/enterprise edition from a private S3 bucket.
--- a/docs/configuration_options.md
+++ b/docs/configuration_options.md
@@ -1957,6 +1957,16 @@ Scylla Doctor version to use for artifact tests. Set to specific version (e.g., 
 * appendable
 
 
+## **scylla_doctor_full_tarball_url** / SCT_SCYLLA_DOCTOR_FULL_TARBALL_URL
+
+Direct URL to a full edition Scylla Doctor tarball in S3. When set, bypasses the<br>standard version-based S3 lookup and downloads SD directly from this URL.<br>Use for testing unofficial or pre-release SD versions.<br>Example: 'https://s3.amazonaws.com/my-bucket/scylla-doctor-1.11-rc1.tar.gz'
+
+**default:** N/A
+
+**type:** str
+* appendable
+
+
 ## **run_scylla_doctor_only** / SCT_RUN_SCYLLA_DOCTOR_ONLY
 
 When true, the artifact test runs only the Scylla Doctor validation<br>(install, collect vitals, analyze, verify) and skips all other artifact checks<br>such as stop/start, cassandra-stress, housekeeping, etc. Useful for fast SD<br>release gating. Implies run_scylla_doctor=true.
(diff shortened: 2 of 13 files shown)


vars/sdReleaseGatingPipeline.groovy
pehala, line 1: nit: I would rather have a longer name without abbreveations: `ScyllaDoctorGatingPipeline` I think it is more verbose and easier to understand.

artifacts_test.py
```diff
@@ -358,6 +358,13 @@ def test_scylla_service(self):  # noqa: PLR0915
 
         backend = self.params.get("cluster_backend")
 
+        if self.params.get("run_scylla_doctor_only"):
+            with self.logged_subtest("verify node health"):
+                self.verify_node_health()
+            with self.logged_subtest("check scylla_doctor results"):
+                self.run_scylla_doctor()
+            return
```
pehala, line 366: Wouldnt it be better to have a specific test function for it then? I see no reason why we need to reuse another test with a new flag

unit_tests/test_scylla_doctor.py
```diff
@@ -489,3 +500,81 @@ def test_find_scylla_bundled_python3_raises_when_not_found(doctor, is_nonroot, r
 
     with pytest.raises(ScyllaDoctorException, match="No Scylla-bundled python3"):
         doctor.find_scylla_bundled_python3("/home/user")
+
+
+# --- run_analysis_phase tests ---
```
pehala, line 505: nit: I would prefer these tests in a new file. Overall the tests in this test file seems to be all connected to scylla doctor but to different features, so I think splitting it into module and individual small testfiles would be better.

Given you add only 3 tests, no need to do in this PR (or ever if you dont agree with me)

sdcm/tester.py
```diff
@@ -3975,6 +3975,20 @@ def collect_node_stats(node):
...
                             self.log.debug("Downloaded scylla-doctor logs from %s to %s", node.name, local_logs_path)
 
+                        if doctor.is_full_edition:
+                            self.log.info("Running scylla-doctor analysis phase (full edition) on %s...", node.name)
+                            doctor.run_analysis_phase()
+                            if doctor.analysis_report_file:
+                                local_analysis_path = os.path.join(
+                                    self.logdir, f"scylla_doctor_{node.name}_analysis.json"
+                                )
+                                node.remoter.receive_files(src=doctor.analysis_report_file, dst=local_analysis_path)
```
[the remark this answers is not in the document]
pehala: Ai might have a point here

sdcm/utils/aws_utils.py
```diff
@@ -616,6 +616,17 @@ def get_by_owner_ami(parameter: str, region_name) -> str:
...
+        # List available images from the owner to help debug naming issues
+        all_images = ec2_resource.images.filter(
+            Owners=[owner],
+            Filters=[{"Name": "architecture", "Values": [arch]}],
+        )
+        available_names = sorted([img.name for img in all_images], reverse=True)[:10]
+        raise ValueError(
+            f"No AMI found for owner={owner}, architecture={arch}, name_filter={name_filter} "
+            f"in region {region_name}. Available image names (top 10): {available_names}"
+        )
```
pehala, line 635: This seems unrelated to the scylla doctor

artifacts_test.py
juliayakovlev: This is was a decision - to use the artifact test and do not create a new one (if I understood your comment correctly)

sdcm/utils/aws_utils.py
juliayakovlev: @pehala 
I put it in the separate commit now. Would you prefer it in the separate PR?
It was a problem when I ran the artifact test for this PR in this part, so I fixed it here

artifacts_test.py
pehala: Just the test function, not a new test file. You dont need a new tesst config, you can still reuse artifacts machinery and you are using a new pipeline you chsnging the test function there is also fine

fruch: we don't want to run just that part, no point of premature optimization here.

sdcm/utils/aws_utils.py
pehala: Separate commit is fine, thanks

artifacts_test.py
pehala: > we don't want to run just that part, no point of premature optimization here.

If I understand it correctly, if `run_scylla_doctor_only` is set, then you run only the one if and then you return and not run anything else from the function, why not have a test function for scylla doctor only instead of a config option, what am I missing?

docs/configuration_options.md
```diff
@@ -1957,6 +1957,25 @@ Scylla Doctor version to use for artifact tests. Set to specific version (e.g.,
...
+
+**type:** str
+* appendable
+
+
+## **run_scylla_doctor_only** / SCT_RUN_SCYLLA_DOCTOR_ONLY
+
+When true, the artifact test runs only the Scylla Doctor validation<br>(install, collect vitals, analyze, verify) and skips all other artifact checks<br>such as stop/start, cassandra-stress, housekeeping, etc. Useful for fast SD<br>release gating. Implies run_scylla_doctor=true.
+
+**default:** N/A
```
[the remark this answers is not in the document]
fruch: we need to fix the configuration to put false

artifacts_test.py
fruch: since we don't want to duplicate all of the jobs for this purpose

docs/configuration_options.md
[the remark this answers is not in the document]
pehala: Is it a mistake of the autogenerator or something is missing in this PR? I have seen this happen multiple times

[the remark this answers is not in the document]
fruch: probably there's  something like `'N/A' if not x else: str(x)` somewhere ...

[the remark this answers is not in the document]
fruch: `default_text = default if default else "N/A"` is the exact one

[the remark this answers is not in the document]
pehala: Fixed in https://github.com/scylladb/scylla-cluster-tests/pull/15005

pehala (approved):

fruch:
@juliayakovlev once you run some of the default artifacts tests and they are working, we can merge it

fruch (approved):

PR #14961, opened 2026-06-08
ci(oci): add 'oci' label to the vars/getJenkinsLabels.groovy

vponomaryov:
### Testing

- [ ]

### Reminders

### DIFF
--- a/vars/getJenkinsLabels.groovy
+++ b/vars/getJenkinsLabels.groovy
@@ -37,6 +37,10 @@ def call(String backend, String region=null, String datacenter=null, String loca
     def gcp_project = params.gce_project?.trim() ?: 'gcp-sct-project-1'
     gcp_project = gcp_project == 'gcp' ? 'gcp-skilled-adapter-452' : gcp_project
 
+    // NOTE: 'aws', 'gce' and 'oci' labels (without region) are useful for cases
+    //       when backend-specific region param has no value.
+    //       Even if test-case config has, it won't be accessible from here.
+    //       So, apply default value for such cases.
     def jenkins_labels = ['aws-eu-west-1': 'aws-sct-builders-eu-west-1-v3-asg',
                           'aws-eu-west-2': 'aws-sct-builders-eu-west-2-v3-asg',
                           'aws-eu-north-1': 'aws-sct-builders-eu-north-1-v3-asg',
@@ -54,6 +58,7 @@ def call(String backend, String region=null, String datacenter=null, String loca
                           'aws': 'aws-sct-builders-eu-west-1-v3-asg',
                           'azure-eastus': 'aws-sct-builders-us-east-1-v3-asg',
                           'aws-fips': 'aws-sct-builders-us-east-1-v4-fibs-CI-FIPS',
+                          'oci': 'oci-sct-builders-us-ashburn-1-v1',
                           'oci-us-ashburn-1': 'oci-sct-builders-us-ashburn-1-v1',
                           'oci-us-phoenix-1': 'oci-sct-builders-us-phoenix-1-v1',
                           ]


vponomaryov:
Example of an error WITHOUT this fix: https://jenkins.scylladb.com/view/master/job/scylla-master/job/artifacts/job/artifacts-oel8-test/1030/console

pehala (dismissed):
LGTM, is it caused by the fact we ran OCI on different DC than `us-ashburn` and `us-phoenix` and so it couldnt find the default?

vponomaryov:
> LGTM, is it caused by the fact we ran OCI on different DC than `us-ashburn` and `us-phoenix` and so it couldnt find the default?

No, it was caused by specifics of the Jenkins:

Each particular build of a CI job runs with parameters that were existing in the old state of the jenkinsfile.
And the "oci_region_name" was not set in such a case.
Moreover, Jenkins was not applying new params, for unknown to me reasons, for multiple test runs.
But this change is not about parameters of a CI job, it is about under-the hood stuff which should have workarounded such a limitation.

fruch:
> > LGTM, is it caused by the fact we ran OCI on different DC than `us-ashburn` and `us-phoenix` and so it couldnt find the default?
> 
> No, it was caused by specifics of the Jenkins:
> 
> Each particular build of a CI job runs with parameters that were existing in the old state of the jenkinsfile. And the "oci_region_name" was not set in such a case. Moreover, Jenkins was not applying new params, for unknown to me reasons, for multiple test runs. But this change is not about parameters of a CI job, it is about under-the hood stuff which should have workarounded such a limitation.

it worth putting a comment for those, else someone would be eager to remove those
(I didn't completely remember why we have those fallbacks)

vponomaryov:
> > > LGTM, is it caused by the fact we ran OCI on different DC than `us-ashburn` and `us-phoenix` and so it couldnt find the default?
> > 
> > 
> > No, it was caused by specifics of the Jenkins:
> > Each particular build of a CI job runs with parameters that were existing in the old state of the jenkinsfile. And the "oci_region_name" was not set in such a case. Moreover, Jenkins was not applying new params, for unknown to me reasons, for multiple test runs. But this change is not about parameters of a CI job, it is about under-the hood stuff which should have workarounded such a limitation.
> 
> it worth putting a comment for those, else someone would be eager to remove those (I didn't completely remember why we have those fallbacks)

Which comment?
I described the situation which is kind of one-time situation.
There may be other reasons.
As far as I understand, having these short variants as `aws`, `gce` and `oci` allows us to have an empty value in the appropriate region parameter field.
So, why one would want to remove it?

fruch:
> > > > LGTM, is it caused by the fact we ran OCI on different DC than `us-ashburn` and `us-phoenix` and so it couldnt find the default?
> > > 
> > > 
> > > No, it was caused by specifics of the Jenkins:
> > > Each particular build of a CI job runs with parameters that were existing in the old state of the jenkinsfile. And the "oci_region_name" was not set in such a case. Moreover, Jenkins was not applying new params, for unknown to me reasons, for multiple test runs. But this change is not about parameters of a CI job, it is about under-the hood stuff which should have workarounded such a limitation.
> > 
> > 
> > it worth putting a comment for those, else someone would be eager to remove those (I didn't completely remember why we have those fallbacks)
> 
> Which comment? I described the situation which is kind of one-time situation. There may be other reasons. As far as I understand, having these short variants as `aws`, `gce` and `oci` allows us to have an empty value in the appropriate region parameter field. So, why one would want to remove it?

cause I don't think any remembers this fact (I didn't), i.e. why I suggest a comment next to it.
it's not something anyone can easily assume.

vponomaryov:
> > > > > LGTM, is it caused by the fact we ran OCI on different DC than `us-ashburn` and `us-phoenix` and so it couldnt find the default?
> > > > 
> > > > 
> > > > No, it was caused by specifics of the Jenkins:
> > > > Each particular build of a CI job runs with parameters that were existing in the old state of the jenkinsfile. And the "oci_region_name" was not set in such a case. Moreover, Jenkins was not applying new params, for unknown to me reasons, for multiple test runs. But this change is not about parameters of a CI job, it is about under-the hood stuff which should have workarounded such a limitation.
> > > 
> > > 
> > > it worth putting a comment for those, else someone would be eager to remove those (I didn't completely remember why we have those fallbacks)
> > 
> > 
> > Which comment? I described the situation which is kind of one-time situation. There may be other reasons. As far as I understand, having these short variants as `aws`, `gce` and `oci` allows us to have an empty value in the appropriate region parameter field. So, why one would want to remove it?
> 
> cause I don't think any remembers this fact (I didn't), i.e. why I suggest a comment next to it. it's not something anyone can easily assume.

I still have no idea what comment you expect there.
I don't know your reasons for other such 2 cases and my reason is definitely not the same.

fruch:
> > > > > > LGTM, is it caused by the fact we ran OCI on different DC than `us-ashburn` and `us-phoenix` and so it couldnt find the default?
> > > > > 
> > > > > 
> > > > > No, it was caused by specifics of the Jenkins:
> > > > > Each particular build of a CI job runs with parameters that were existing in the old state of the jenkinsfile. And the "oci_region_name" was not set in such a case. Moreover, Jenkins was not applying new params, for unknown to me reasons, for multiple test runs. But this change is not about parameters of a CI job, it is about under-the hood stuff which should have workarounded such a limitation.
> > > > 
> > > > 
> > > > it worth putting a comment for those, else someone would be eager to remove those (I didn't completely remember why we have those fallbacks)
> > > 
> > > 
> > > Which comment? I described the situation which is kind of one-time situation. There may be other reasons. As far as I understand, having these short variants as `aws`, `gce` and `oci` allows us to have an empty value in the appropriate region parameter field. So, why one would want to remove it?
> > 
> > 
> > cause I don't think any remembers this fact (I didn't), i.e. why I suggest a comment next to it. it's not something anyone can easily assume.
> 
> I still have no idea what comment you expect there. I don't know your reasons for other such 2 cases and my reason is definitely not the same.

you exampled it above:

"this definition without region is needed for some situations when we have new parameters part of this selection logic, and we should still want to run the job (so update to default parameters would still happen)"

without any comment inst' not clear what it's needed for, and there's a risk someone would consider it redundant.
could be there there's more reason, I don't know/remember them, and would forget this one in few days too.

pehala (approved):

fruch (approved):

PR #14966, opened 2026-06-09
fix(upgrade): use java cassandra-stress for azure complex workload [SCT-456]

soyacz:
## What

Revert the four `stress_cmd_complex_*` commands in `configurations/azure/azure_rolling_upgrade.yaml` from `cql-stress-cassandra-stress` back to java `cassandra-stress`.

## Why

The azure rolling-upgrade config overrode the complex workload to `cql-stress-cassandra-stress user no-warmup profile=/tmp/complex_schema.yaml`. cql-stress does **not** support the user-profile mode — [QATOOLS-76](https://scylladb.atlassian.net/browse/QATOOLS-76) is **Won't Fix** — and SCT's `CqlStressCassandraStressThread` never stages the profile YAML onto the loader. The complex steps fail immediately:

```
Failed to parse parameter profile=.
  0: Invalid profile yaml filepath: /tmp/complex_schema.yaml
  1: No such file or directory (os error 2)
```

The bug was latent since `174af7dd5` and surfaced on 2026-06-08 when scylladb/scylladb#22677 closed, un-skipping the complex workload on branch-2026.2 (which has no `sct-2026.2-skip` label).

Java `cassandra-stress` stages the profile (`stress_thread.py` `_run_cs_stress`) and supports user mode, matching the base `rolling-upgrade.yaml`.

## Fix issue

[SCT-456](https://scylladb.atlassian.net/browse/SCT-456)

## Testing

- `grep complex configurations/azure/azure_rolling_upgrade.yaml` — all four commands now use `cassandra-stress`.

Follow-up (separate): migrate the complex data-integrity workload to Latte for proper rust-driver coverage with validation.

[QATOOLS-76]: https://scylladb.atlassian.net/browse/QATOOLS-76?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[SCT-456]: https://scylladb.atlassian.net/browse/SCT-456?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pehala:
> The bug was latent since `174af7dd5` and surfaced on 2026-06-08 when [scylladb/scylladb#22677](https://github.com/scylladb/scylladb/issues/22677) closed, un-skipping the complex workload on branch-2026.2 (which has no `sct-2026.2-skip` label).

I would backport it to also 2026.1 and 2025.4, complex workload makes more sense everywhere and MV are enabled since 2025.4, which is the last supported release.

fruch:
> > The bug was latent since `174af7dd5` and surfaced on 2026-06-08 when [scylladb/scylladb#22677](https://github.com/scylladb/scylladb/issues/22677) closed, un-skipping the complex workload on branch-2026.2 (which has no `sct-2026.2-skip` label).
> 
> I would backport it to also 2026.1 and 2025.4, complex workload makes more sense everywhere and MV are enabled since 2025.4, which is the last supported release.

but those shouldn't be running it, https://github.com/scylladb/scylladb/issues/22677 was fixed on master
it wouldn't be relevant if all of the release would get that fix.

we can backport it, but it won't be used

regardless it's part of the test that was never tested with tablets, I see no point backporting it to the releases, it won't just gonna generate noise

also for upgrade test both ends need to be supported, so if we upgrade from 2025.1 (in 2025.4 and in 2026.1) we can't run it

fruch (approved):

PR #14982, opened 2026-06-09
fix(mgmt-restore): exclude CDC table from restore keyspace list by default

yarongilor:
Added a version-aware keyspace exclusion filter to create_restore_task()
that excludes system_distributed_everywhere.cdc_generation_descriptions_v2
on Scylla >= 2026.2. This table was removed in newer Scylla versions but
may still be present in backup snapshots taken from older versions, causing
Scylla Manager restore to fail. The filter is only applied when the target
cluster runs a version where the table no longer exists. This is a
temporary workaround until Scylla Manager handles CDC table exclusion
natively.

Fixes: https://scylladb.atlassian.net/browse/SCT-363
Refs: https://scylladb.atlassian.net/browse/CLOUD-1519

Refs: https://github.com/scylladb/scylladb/pull/29482

### Testing

- [restore_a_2025_backup](https://argus.scylladb.com/tests/scylla-cluster-tests/ab55fcd3-0af2-4722-86e5-f243ed449127)
--  The test passed the workaround part ok. (Then later it failed for a different non-related SM issue of https://scylladb.atlassian.net/browse/SCYLLADB-2526 )

### Reminders

### DIFF
--- a/sdcm/mgmt/cli.py
+++ b/sdcm/mgmt/cli.py
@@ -11,6 +11,7 @@
 #
 # Copyright (c) 2021 ScyllaDB
 
+from functools import cached_property
 import time
 import logging
 import datetime
@@ -40,6 +41,7 @@ from sdcm.provision.helpers.certificate import TLSAssets
 from sdcm.utils.context_managers import DbNodeLogger
 from sdcm.utils.database_query_utils import is_system_keyspace
 from sdcm.utils.replication_strategy_utils import ReplicationStrategy
+from sdcm.utils.version_utils import ComparableScyllaVersion
 from sdcm.wait import WaitForTimeoutError
 
 LOGGER = logging.getLogger(__name__)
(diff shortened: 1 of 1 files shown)


sdcm/mgmt/cli.py
```diff
@@ -700,6 +707,12 @@ def create_restore_task(
             cmd += f" --dc-mapping {dc_mapping}"
         if manager_backup_restore_method:
             cmd += f" --method {manager_backup_restore_method.value} "
+        if restore_data:
```
mikliapko: We already have `restore_data` branch that extends cmd.
Let's move this code there to avoid duplication

sdcm/mgmt/cli.py
```diff
@@ -674,6 +674,12 @@ def get_cluster_id_by_name(self, cluster_name: str):
     def set_cluster_id(self, value: str):
         self.id = value
 
+    # Temporary workaround for https://scylladb.atlassian.net/browse/SCT-363
+    # Exclude CDC table that no longer exists in newer Scylla versions but may still be present in backup snapshots.
+    # Remove this once Scylla Manager excludes CDC tables from backup/restore automatically
+    # (in https://scylladb.atlassian.net/browse/CLOUD-1519).
+    DEFAULT_RESTORE_KEYSPACE_LIST = ["*", "!system_distributed_everywhere.cdc_generation_descriptions_v2"]
+
```
mikliapko, line 686: I'd suggest to move it directly to `create_restore_task` together with the comment to make this change atomic and easy to remove in the future when  https://scylladb.atlassian.net/browse/CLOUD-1519 issue is solved

mikliapko (commented):
What do you think about applying this change only for specific Scylla versions (AFAIU >= 2026.2)? Thus, we would keep the default behavior for all previous Scylla releases.

pehala:
> What do you think about applying this change only for specific Scylla versions (AFAIU >= 2026.2)? Thus, we would keep the default behavior for all previous Scylla releases.

We dont plan to backport it exactly for that reason

mikliapko:
> > What do you think about applying this change only for specific Scylla versions (AFAIU >= 2026.2)? Thus, we would keep the default behavior for all previous Scylla releases.
> 
> We dont plan to backport it exactly for that reason

Manager tests are run from master branch. 
We don't support several branches there. This approach was dismissed some time ago due to all complexities it was creating (no one bothered about backporting framework core related changes into manager branches).

pehala:
> Manager tests are run from master branch.
> We don't support several branches there. This approach was dismissed some time ago due to all complexities it was creating (no one bothered about backporting framework core related changes into manager branches).

Welp, that is far from ideal. I would say this workaround is harmless even for older oversions and will be removed once manager changes gets in or after we create new backups, however, if you feel like the version if is neccessary we can add it

mikliapko:
> I would say this workaround is harmless even for older oversions and will be removed once manager changes gets in or after we create new backups, however, if you feel like the version if is neccessary we can add it

Please, add it.
It usually takes more time to remove workaround then implement it in first place. 
Moreover, I'd really like to keep the command unchanged for all previous versions. I'm pretty sure that's the way how the majority of users run restore now (without `--keyspace` flag)

sdcm/nemesis/__init__.py
```diff
@@ -3041,8 +3036,18 @@ def _restore_schema(locations: list, cluster_id: str, tag: str) -> None:
             # self.tester.set_ks_strategy_to_network_and_rf_according_to_cluster(
             #    keyspace=chosen_snapshot_info["keyspace_name"], repair_after_alter=False)
         try:
+            # Temporary workaround for https://scylladb.atlassian.net/browse/SCT-363
+            # In Scylla >= 2026.2, cdc_generation_descriptions_v2 was removed but may still exist in backup
+            # snapshots. Remove this once https://scylladb.atlassian.net/browse/CLOUD-1519 is resolved.
+            keyspace_list = None
+            scylla_version = self.cluster.nodes[0].scylla_version
+            if ComparableScyllaVersion(scylla_version) >= "2026.2.0~dev":
```
pehala: This wont help you with tablet aware tests, I dont think we should include this only for nemesis. IF we want this for nemesis only, we wont need the version check at all, as manager doesnt use that AFAIK

sdcm/mgmt/cli.py
yarongilor: fixed

sdcm/nemesis/__init__.py
yarongilor: this is needed in order to fix the nemesis error of https://scylladb.atlassian.net/browse/SCT-363

sdcm/mgmt/cli.py
yarongilor: fixed

yarongilor:
> > I would say this workaround is harmless even for older oversions and will be removed once manager changes gets in or after we create new backups, however, if you feel like the version if is neccessary we can add it
> 
> Please, add it. It usually takes more time to remove workaround then implement it in first place. Moreover, I'd really like to keep the command unchanged for all previous versions. I'm pretty sure that's the way how the majority of users run restore now (without `--keyspace` flag)

@mikliapko , fixed it to be "version aware" as suggested.

sdcm/mgmt/operations.py
```diff
@@ -942,6 +947,21 @@ def restore_with_manager_task(
 
         return restore_task
 
+    def _get_restore_keyspace_filter_for_legacy_tables(self) -> list[str] | None:
```
mikliapko: Why for internal use only?
Do you even need a separate method for that if it's used only once? If yes, is it the right place to put it here in `ManagerTestFunctionsMixIn`?

yarongilor: @mikliapko , @pehala , i updated it to a more minimalistic change now.

sdcm/mgmt/cli.py
```diff
@@ -682,13 +691,23 @@ def create_restore_task(
         snapshot_tag=None,
         dc_mapping=None,
         manager_backup_restore_method=None,
+        keyspace_list=None,
+        scylla_version=None,
```
mikliapko: TBH, I don't like the fact we have to pass scylla_version into `create_restore_task` method.
Considering there is no other way to implement it without wider refactoring, let's keep it for now and hope the original issue will be resolved on Manager side soon and we will remove this workaround.

mikliapko (dismissed):
LGTM

sdcm/mgmt/cli.py
yarongilor: fixed

yarongilor:
@pehala , @mikliapko ,

The fix is updated with:

sdcm/mgmt/cli.py — adds cached_property scylla_version (self-contained, gets version from sctool status) + CDC exclusion logic in create_restore_task

mikliapko (dismissed):
It looks much better now, thanks for addressing it.

LGTM

pehala (dismissed):

pehala:
@yarongilor precommit is failing on both your cases and outside failure, fix your failure, I will submit a fix

EDIT: fix for the unrelated failures has been already merged, so please rebase

pehala (approved):

PR #14993, opened 2026-06-09
fix(nemesis): always add node after removal even if repair fails

cezarmoise:
In `_remove_node_add_node`, wrap the best-effort repair call in a try/except so that `_add_and_init_new_cluster_nodes` is always reached, keeping the cluster at its original node count.

Fixes: SCT-190

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -3780,9 +3780,17 @@ class NemesisRunner:
 
         # full cluster repair
         # Repairing will result in a best effort repair due to the terminated node,
-        # and as a result requires ignoring repair errors
-        with DbEventsFilter(db_event=DatabaseLogEvent.RUNTIME_ERROR, line="failed to repair"):
-            self.run_repair(ignore_down_hosts=True)
+        # and as a result requires ignoring repair errors.
+        # Even if repair fails, we must still add a replacement node so the
+        # cluster is left with the original number of nodes.
+        try:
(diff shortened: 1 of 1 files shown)


sdcm/nemesis/__init__.py
```diff
@@ -3788,13 +3807,28 @@ def remove_node():
...
+        try:
+            with DbEventsFilter(db_event=DatabaseLogEvent.RUNTIME_ERROR, line="failed to repair"):
+                self.run_repair(ignore_down_hosts=True)
+        except WaitForTimeoutError as exc:
+            # See https://scylladb.atlassian.net/browse/SCYLLADB-639
+            # Repair may take longer to complete than expected, but it doesn't fail.
+            InfoEvent(
+                message=f"Repair timed out after removing node {node_to_remove.name}: {exc}",
+                severity=Severity.WARNING,
+            ).publish()
```
pehala: We should check if the repair finished eventually, we shouldnt leave it hanging and accept that it is timed out

cezarmoise: I will handle repair finishing or not in another patch.
Simplified this so that it just ensures that a node is added even if repair fails.

cezarmoise:
Removed the check for `WaitForTimeoutError`. If a repair times out, then there is another problem, because it exists the `run_repair_manager` early and does not call `mgr_task.stop()`. Which causes other repairs to fail because this one is still in progress.
I will handle that in another PR.
This one is just to ensure that the nemesis does not exit early before adding a node back to the cluster.

pehala (approved):

jsmolar (approved):

PR #14994, opened 2026-06-09
ci(oci): wait for rpm lock on rhel-based distros

vponomaryov:
Running Artifacts tests on the OCI backend may lead to the following error:

```
        > Command: 'sudo yum install -y scylla'
        >
        > Exit code: 1
        >
        > Stdout:
            ...
        > Error: GPG check FAILED
```

It is caused by the `rpm.lock` being held by the OCI agent installation
process which starts after the DB node startup.
    
So, tolerate rpm.lock holdings in SCT.
Do it by adding an additional bash checker for that lock before
executing package management commands.

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-artifacts-oel8-oci/12

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -97,6 +97,7 @@ from sdcm.remote.remote_file import remote_file, yaml_file_to_dict, dict_to_yaml
 from sdcm import wait, mgmt
 from sdcm.sct_config import SCTConfiguration
 from sdcm.utils.apt import apt_cmd
+from sdcm.utils.rpm import rpm_cmd
 from sdcm.sct_events.continuous_event import ContinuousEventsRegistry
 from sdcm.sct_events.system import AwsKmsEvent
 from sdcm.snitch_configuration import SnitchConfig
@@ -2263,7 +2264,7 @@ class BaseNode(AutoSshContainerMixin):
 
         if debug_install:
             if self.distro.is_rhel_like:
-                self.remoter.sudo("yum install -y scylla-gdb", verbose=True, ignore_status=True)
+                self.remoter.sudo(rpm_cmd("yum", "install -y scylla-gdb"), verbose=True, ignore_status=True)
             elif self.distro.is_sles:
                 self.remoter.sudo("zypper install -y scylla-gdb", verbose=True, ignore_status=True)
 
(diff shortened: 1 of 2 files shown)


vponomaryov:
Example of an issue which gets solved here:
- [scylla-master/artifacts/artifacts-oel8-test#1036](https://argus.scylladb.com/tests/scylla-cluster-tests/50efc02b-e7af-49b9-8316-5e91d3f4acd6)

sdcm/cluster.py
```diff
@@ -2467,15 +2468,23 @@ def fetch_apt_keys(self):
     )
     def install_package(self, package_name: str, package_version: str = None, ignore_status: bool = False) -> None:
         if self.distro.is_rhel_like:
-            pkg_cmd = next(
+            pkg_mgr = next(
                 mgr
                 for mgr in ["microdnf", "yum", "dnf"]
```
pehala, line 2473: Why does this use all of the three package managers while e.g. in `config_setup` we use only `yum`. Since you are unifying calls, wouldnt it make sense to unify the managers as well? i.e. the unction would choose the best one?

fruch: microdnf is there cause our docker images are ubi9 based

it's not need it `config_setup`, since we are not doing installations of scylla on docker images

it's not related to this PR

fruch:
@vponomaryov can we have PR and commit description for this one ?

sdcm/cluster.py
vponomaryov: @pehala 
As @fruch already mentioned, usage of 3 binaries is not related to this PR.
This PR only adds checker for rpm lock for these binaries which already exist in the code.

pehala (approved):

vponomaryov:
Just updated the commit message as requested

pehala:
> It is caused by the `rpm.lock` being held by the OCI agent installation
> process which starts after the DB node startup.

Why are we running the instalation in paralle? Shouldnt we wait with setup until the OCI agents finishes?

vponomaryov:
> > It is caused by the `rpm.lock` being held by the OCI agent installation
> > process which starts after the DB node startup.
> 
> Why are we running the instalation in paralle? Shouldnt we wait with setup until the OCI agents finishes?

This PR does exactly it - makes SCT wait for the lock to be freed.

PR #15021, opened 2026-06-11
ci(oci): add analog of the tier2 multidc test for the OCI backend

vponomaryov:
### Testing

- [scylla-staging/valerii/vp-longevity-cdc-8h-multi-dc-topology-changes-streaming-oci-test#3](https://argus.scylladb.com/tests/scylla-cluster-tests/6f2c365f-8067-4ba5-8f9b-de0837c11fa4) (failed, but error is unrelated to the PR)

### Reminders

### DIFF
--- a/jenkins-pipelines/oss/tier2/longevity-cdc-8h-multi-dc-topology-changes-streaming-oci.jenkinsfile
+++ b/jenkins-pipelines/oss/tier2/longevity-cdc-8h-multi-dc-topology-changes-streaming-oci.jenkinsfile
@@ -1,0 +1,12 @@
+#!groovy
+
+// trick from https://github.com/jenkinsci/workflow-cps-global-lib-plugin/pull/43
+def lib = library identifier: 'sct@snapshot', retriever: legacySCM(scm)
+
+// Runs LongevityTest for topology changes with streaming instead of RBNO
+longevityPipeline(
+    backend: 'oci',
+    oci_region_name: '''["us-ashburn-1", "us-phoenix-1"]''',
+    test_name: 'longevity_test.LongevityTest.test_custom_time',
+    test_config: '''["test-cases/longevity/longevity-cdc-100gb-8h-multi-dc-topology-changes.yaml", "configurations/disable_rbno.yaml"]''',
+)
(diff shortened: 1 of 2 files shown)


pehala (approved):
To reduce the number of total tests, I think it is totally fine to transform the existing one instead as well

vponomaryov:
> To reduce the number of total tests, I think it is totally fine to transform the existing one instead as well

The `OEL8` and `OEL9` CI jobs, which I assume you refer to, were switched to the OCI (Oracle Cloud) because it has "Oracle Enterprise Linux" images whereas AWS doesn't.
So, in that situation it was choice where we either delete CI jobs or use OCI backend for it.

Here, the situation is different. The base CI job is ok.
The config, which is being added, is the first multi-dc config for CI which we are able use for comparison to some existing one.

I think that, right now, we should not completely switch anything to OCI having available analog with another backend.
Why? Because it is too unstable.

For example, see following issue caused by the Oracle Cloud:
- https://scylladb.atlassian.net/browse/SCT-447

It has been around for about 2 weeks already spoiling some tests by making a newly provisioned VM not have NVMe disks available.

Several months ago, there was about 5-6 days when `us-ashburn-1` region (main one in our account) was just failing resource provisioning.
And Oracle didn't communicate the issue clearly and openly.

So, additional CI job.

fruch:
@vponomaryov 

what about https://jenkins.scylladb.com/job/scylla-master/job/tier1/job/longevity-multidc-schema-topology-changes-12h-test/ from tier1 ?
it something blocking us from moving it to OCI ? (i.e. that is our target to handle once, we which are run regularly)

pehala:
> what about https://jenkins.scylladb.com/job/scylla-master/job/tier1/job/longevity-multidc-schema-topology-changes-12h-test/ from tier1 ?
> it something blocking us from moving it to OCI ? (i.e. that is our target to handle once, we which are run regularly)

If @vponomaryov is not sure of the stability of OCI yet, I would rather not switch it, unless we also change ownership of that test to him.

vponomaryov:
> @vponomaryov
> 
> what about https://jenkins.scylladb.com/job/scylla-master/job/tier1/job/longevity-multidc-schema-topology-changes-12h-test/ from tier1 ? it something blocking us from moving it to OCI ? (i.e. that is our target to handle once, we which are run regularly)

It uses custom network config, which is not yet supported for OCI.
But, even if it was supported, the cloud instability is a concern.
I would not recommend switching some job completely to OCI - only adding an equal option.

vponomaryov:
Anyway, it is simple config addition PR, can we merge it?

pehala (approved):

PR #15061, opened 2026-06-15
fix(monitoring): Backport monitoring 4.15 to 2025.x

k0machi:
Due to some bugs with IPAddress parsing in monitoring scripts, backporting newer versions to fix

Fixes SCT-201

### Testing

- [x] [Argus](https://argus.scylladb.com/workspace?state=WyIxZDY1ZWFhZi1jNDVkLTQ1ODEtYmM2OS04ZTQwNzdmMDVkYTgiXQ)
- [x] [Jenkins](https://jenkins.scylladb.com/job/scylla-staging/job/alexey/job/longevity-1tb-5days-azure-test/)

### DIFF
--- a/configurations/manager/debian12.yaml
+++ b/configurations/manager/debian12.yaml
@@ -1,0 +1,3 @@
+ami_id_monitor: 'resolve:ssm:/aws/service/debian/release/12/latest/amd64'  # Debian Bookworm latest image
+ami_monitor_user: 'admin'
+monitor_branch: 'branch-4.15'  # Pass it as debian image is not the formal monitor image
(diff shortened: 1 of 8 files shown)


pehala:
@fruch @amnonh Could we backport newer monitoring to 2025.1 and 2025.4 instead? it should fix it as well, without doing branch specific removals. Is the new monitoring supported for these versions?

k0machi:
> @fruch @amnonh Could we backport newer monitoring to 2025.1 and 2025.4 instead? it should fix it as well, without doing branch specific removals. Is the new monitoring supported for these versions?

I will try this since after doing some extended investigations I found that this bug is all over the scripts in the current monitoring version.

Update: It works. I think it would be for the best to do it this way.

pehala:
> Update: It works. I think it would be for the best to do it this way.

The test is still running, which is great improvement, but it definitely ignored custom stress duration, it is running for 3+ hours.

I removed the backport label, it wont work on branch specific PRs, you will have to do one manually for 2025.4

fruch:
seems like we need to backport this from more other reasons, the GCP monitoring image is now gone

fruch (approved):

PR #15069, opened 2026-06-15
fix(nemesis): restore original kms_hosts config after encryption nemesis

yarongilor:
The EnableDisableTableEncryptionAwsKmsProviderWithoutRotationMonkey nemesis adds a "kms-host" entry to kms_hosts in scylla.yaml but never restores the original configuration on cleanup. This can cause SSTable loading failures when a subsequent operation (e.g., MgmtRestore) loads SSTables that reference a pre-existing KMS host (like "auto") after a node restart picks up a scylla.yaml that lost the original kms_hosts entries due to the model_dump serialization in remote_scylla_yaml().

Save the original kms_hosts state before modification and restore it in the finally block.

Fixes: https://scylladb.atlassian.net/browse/SCYLLADB-2208

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -17,6 +17,7 @@ Module containing logic for running disruptions on a test cluster
 """
 
 import contextlib
+import copy
 import datetime
 import importlib
 import inspect
@@ -4464,6 +4465,7 @@ class NemesisRunner:
         scylla_encryption_options = {"cipher_algorithm": "AES/ECB/PKCS5Padding", "secret_key_strength": 128}
         scylla_encryption_options |= additional_scylla_encryption_options or {}
         aws_kms, kms_key_alias_name = None, None
+        original_kms_hosts = None  # Track original kms_hosts to restore on cleanup
 
         # Handle AWS KMS specific parts
         if (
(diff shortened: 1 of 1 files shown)


yarongilor:
@vponomaryov , do you think this can fix the issue of https://scylladb.atlassian.net/browse/SCYLLADB-2208?

pehala:
I would consider if we want to run this nemesis in the first place if it can interact like this with other nemesis and overall KMS config.

sdcm/nemesis/__init__.py
pehala, line 4495: We are not overriding the kms_hosts right? we are appending to it only, so everything previously will still be there

sdcm/nemesis/__init__.py
```diff
@@ -4614,6 +4619,14 @@ def run_write_scylla_bench_load(write_cmd):
             with self.cluster.cql_connection_patient(self.target_node, keyspace=keyspace_name) as session:
                 session.execute(f"DROP TABLE {table_name};")
 
+            # Restore original kms_hosts configuration to avoid breaking other operations
+            # (e.g., backup restore) that may rely on the pre-existing KMS host entries
+            if original_kms_hosts is not None:
+                self.actions_log.info("Restoring original kms_hosts configuration on all nodes")
+                for node in self.cluster.nodes:
+                    with node.remote_scylla_yaml() as scylla_yml:
+                        scylla_yml.kms_hosts = copy.deepcopy(original_kms_hosts)
```
vponomaryov, line 4628: This change must be followed by the full cluster restart as it is done above when it gets added.
It is if we want to remove the KMS hosts.

Then,
Full cluster restart is kind of expensive...

So, probably it is easier to configure the KMS host in the SCT beforehand and avoid restarts completely.

sdcm/nemesis/__init__.py
yarongilor: that's how it should be, right, so i guess we can consider shifting the "resposibility" to the nemesis of mgmt restore, to verify/configure a correct kms settings.

yarongilor:
The analysis of https://scylladb.atlassian.net/browse/SCYLLADB-2208 was apparently wrong. The issue should be addressed under the nemesis of MgmtRestore (could create a newer backup as well). closing this one.

PR #15070, opened 2026-06-15
refactor(multidc-parallel-topology-schema-changes): reduce data size

cezarmoise:
Reduce data utilization from approx 60% to approx 30%. Certain nemeses, like `DecommissionMonkey` can cause out of space issues when they remove a node from a rack.
This behavior is true in any disk utilization greater than 50%.

This change will also reduce the time it take for the prepare stress to finish, currently approx 4h.

Fixes: SCYLLADB-2752

### Testing

- [ ]

### Reminders


pehala (approved):

pehala (approved):

yarongilor:
@pehala , i think it should be decided, perhaps in https://github.com/scylladb/scylla-cluster-tests/pull/15147, what is the longevity sub-tests  with capacity > 50% + filtering high-disk-usage (or another laternative) we want to keep .

yarongilor (approved):
LGTM

pehala:
> @pehala , i think it should be decided, perhaps in #15147, what is the longevity sub-tests with capacity > 50% + filtering high-disk-usage (or another laternative) we want to keep .

The plan you mention, is exactly because we dont want to decide which one to keep and we want to be able to keep all of them. For now, I would allow one additional test that is over 50% (your PR), rest downgrade and the same is true for all of future tunings

PR #15075, opened 2026-06-15
feature(build-system): add _symlinks.yaml support for Jenkins pipeline job reuse

fruch:
## Summary

Adds a **symlink mechanism for Jenkins pipelines** that allows creating jobs in one Jenkins folder that reference `.jenkinsfile` files from another location — without duplicating files.

## Problem

Jenkins does not support filesystem symlinks for pipeline definitions. When the same test pipeline needs to appear in multiple Jenkins folders (e.g., scylla-doctor needs copies of artifact tests), the only option today is to duplicate `.jenkinsfile` files. This causes:

- Maintenance burden: changes to the original must be manually propagated
- Drift risk: copies can silently diverge from the original
- File bloat: 35+ artifact pipelines × N copies = unnecessary duplication

## Solution

Introduce `_symlinks.yaml` — a convention file (following the existing patterns of `_display_name`, `_folder_definitions.yaml`, and `_deprecated_jobs.yaml`) that declares virtual pipeline jobs pointing to target `.jenkinsfile` paths.

### Format

```yaml
symlinks:
  - name: "artifacts-docker"
    target: "jenkins-pipelines/oss/artifacts/artifacts-docker.jenkinsfile"
  - name: "artifacts-ami"
    target: "jenkins-pipelines/oss/artifacts/artifacts-ami.jenkinsfile"
    job_name_suffix: "-gating"  # optional, overrides default "-test"
```

### How it works

1. `create_job_tree()` processes `_symlinks.yaml` in each directory during the job creation walk
2. For each symlink entry, it calls `create_pipeline_job()` with:
   - The **target** file as the pipeline script path (so Jenkins runs the original)
   - The **current directory** as the job location (so the job appears in the desired folder)
   - The **name** from the entry as the job name
3. The resulting Jenkins job is identical to one created from the target file directly — same `scriptPath`, same execution — just placed in a different folder

### Example use case

`jenkins-pipelines/scylla-doctor/_symlinks.yaml` creates artifact test jobs in the scylla-doctor folder without duplicating any `.jenkinsfile` files:

```
scylla-doctor/
├── _display_name
├── _symlinks.yaml          ← NEW: declares symlinks
└── sd-release-gating.jenkinsfile
```

The jobs `artifacts-docker-test`, `artifacts-ami-test`, and `artifacts-ami-arm-test` will appear in the scylla-doctor Jenkins folder, all pointing to the original pipelines in `oss/artifacts/`.

## Changes

| File | Change |
|------|--------|
| `utils/build_system/create_test_release_jobs.py` | Added `process_symlinks()` method + integration into `create_job_tree()` |
| `jenkins-pipelines/scylla-doctor/_symlinks.yaml` | Example: scylla-doctor referencing artifact test pipelines |
| `unit_tests/unit/test_pipeline_symlinks.py` | 6 unit tests covering happy path, errors, custom suffix, integration |

## Validation

- [x] All 6 unit tests pass
- [x] ruff lint clean
- [x] Pre-commit hooks pass
- [x] Follows existing SCT patterns (`_deprecated_jobs.yaml`, `_folder_definitions.yaml`)

## Discussion Points

This is a **draft PR for discussion**. Questions for reviewers:

1. Is `_symlinks.yaml` the right naming convention? Alternatives: `_linked_jobs.yaml`, `_job_refs.yaml`
2. Should we support glob patterns for targets (e.g., `target: "jenkins-pipelines/oss/artifacts/artifacts-*.jenkinsfile"`)?
3. Should there be a way to override pipeline parameters for the symlinked job (e.g., different timeout or backend)?


jenkins-pipelines/scylla-doctor/oss/artifacts/_symlinks.yaml
fruch, line 1: need to move this into `oss/artifacts` to match current location

fruch:
need to update the generation pipeline with the need scylla-doctor folder location, otherwise this won't work

PR #15089, opened 2026-06-16
fix(logcollector): pass params to monitoring entities in MonitorSetBase

juliayakovlev:
GrafanaScreenShot and PrometheusSnapshots instances created directly in BaseMonitorSet.get_grafana_screenshots() and download_monitor_data() never had set_params() called, leaving _params as an empty dict.

After commit 2bca67a5b added guards that skip collection when `n_monitor_nodes ` is 0, these direct usages always hit the early return because `{}.get('n_monitor_nodes') ` returns None.

Call `set_params(self.params)`  before `collect()` so the guard sees the actual test configuration.

Fixes: https://scylladb.atlassian.net/browse/SCT-492

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -7670,6 +7670,7 @@ class BaseMonitorSet:
         screenshot_collector = GrafanaScreenShot(
             name="grafana-screenshot", test_start_time=test_start_time, extra_entities=grafana_extra_dashboards
         )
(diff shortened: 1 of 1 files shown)


fruch (approved):

fruch:
@juliayakovlev check why jenkins didn't started

PR #15096, opened 2026-06-16
SCYLLADB-1336: Drop 2023.1 from supported base versions

aleksbykov:
Exclude 2023.1 from the list of Scylla base versions used by rolling upgrade tests, without changing the version indexing used while the base versions are selected.

A new `excluded_base_versions` list holds versions that must not be returned as upgrade base versions. Unlike `unsupported_versions`, excluded versions are kept in the supported list during selection so they do not shift the positional indexing in `ent_release_list` (which previously caused an unrelated STS version, e.g. 2022.2, to slip into the result). The excluded versions are filtered out only at the final stage, right before returning, via a list comprehension.

This preserves the original "two latest" base version selection behavior and removes only 2023.1 from the final result.

Fixes: SCYLLADB-1336

According comments no need to run upgrade from 2023.1: 
 - https://scylladb.atlassian.net/browse/SCYLLADB-1336?focusedCommentId=68294
 - https://scylladb.atlassian.net/browse/SCYLLADB-1336?focusedCommentId=65178

### Testing

- [ ]

### Reminders

### DIFF
--- a/unit_tests/test_base_version.py
+++ b/unit_tests/test_base_version.py
@@ -53,7 +53,10 @@ class TestBaseVersion(unittest.TestCase):
         linux_distro = "ubuntu-jammy"
         cloud_provider = "azure"
         version_list = general_test(scylla_repo, linux_distro, cloud_provider)
-        assert set(version_list) == {"5.2", "2023.1"}
+        # 2023.1 is intentionally excluded from returned upgrade base versions;
+        # see: https://scylladb.atlassian.net/browse/SCT-249
+        assert set(version_list) == {"5.2"}
+        assert "2023.1" not in version_list
 
     def test_2021_1(self):
         scylla_repo = self.url_base + "-enterprise/branch-2021.1/rpm/centos/2021-08-29T00:58:58Z/scylla.repo"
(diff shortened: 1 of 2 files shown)


utils/get_supported_scylla_base_versions.py
```diff
@@ -155,7 +162,13 @@ def get_supported_scylla_base_versions(self, supported_versions) -> list:
...
         ent_base_version = self.filter_rc_only_version(ent_base_version)
         LOGGER.info("Supported scylla base versions for: oss=%s enterprise=%s", oss_base_version, ent_base_version)
-        return list(set(oss_base_version + ent_base_version))
+        # Drop versions that should not be used as upgrade base versions, while keeping the
+        # selection logic above operating on the full supported list (preserves version indexing).
+        base_versions = [v for v in set(oss_base_version + ent_base_version) if v not in excluded_base_versions]
+        LOGGER.info(
+            "Final supported scylla base versions after excluding %s: %s", excluded_base_versions, base_versions
+        )
+        return base_versions
```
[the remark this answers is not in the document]
aleksbykov: Don't need to have sorted list

utils/get_supported_scylla_base_versions.py
```diff
@@ -42,6 +42,13 @@
     "2024.2",
 ]
 
+# Versions that must not be returned as upgrade base versions, but are intentionally kept in the
+# supported list so they don't shift the version indexing used while selecting the base versions.
+# These are filtered out only at the very last stage, right before returning the result.
+excluded_base_versions = [
+    "2023.1",
+]
```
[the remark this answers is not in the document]
aleksbykov: list is enough

fruch (approved):

unit_tests/test_base_version.py
```diff
@@ -53,7 +53,10 @@ def test_ubuntu_22_azure_2023_1(self):
         linux_distro = "ubuntu-jammy"
         cloud_provider = "azure"
         version_list = general_test(scylla_repo, linux_distro, cloud_provider)
-        assert set(version_list) == {"5.2", "2023.1"}
+        # 2023.1 is intentionally excluded from returned upgrade base versions; see:
+        # https://github.com/scylladb/scylla-cluster-tests/pull/15096
```
pehala: I would reference the issue this PR fixes directly

aleksbykov: set issue instead of PR

pehala: minor nit, it would be better to include a full link instead of just a reference as you can click on it directly and look at the issue

aleksbykov: fixed

pehala (approved):

PR #15110, opened 2026-06-17
fix(build-image): skip Docker build steps when image is already built

fruch:
## Problem

The `build-docker-image.yaml` workflow has a step "Check if docker image is already built" that detects when the hydra image has already been built (by checking for a `chore(hydra): create image` commit). However, it only printed a message and `exit 0`, which does **not** prevent subsequent steps from running.

This caused unnecessary work: Set up Python, Install dependencies, Docker login, and even a full Docker build attempt would all run despite the image already existing.

Observed in: https://github.com/scylladb/scylla-cluster-tests/actions/runs/27681450176/job/81869689408?pr=15083

## Fix

- Added `id: check_image` to the check step
- Replaced `exit 0` with a `GITHUB_OUTPUT` variable (`already_built=true/false`)
- Added `if: steps.check_image.outputs.already_built != 'true'` to all subsequent steps

When the image is already built, all build/push steps are now cleanly skipped.

### DIFF
--- a/.github/workflows/build-docker-image.yaml
+++ b/.github/workflows/build-docker-image.yaml
@@ -43,56 +43,66 @@ jobs:
           persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal access token.
 
       - name: Check if docker image is already built
+        id: check_image
         run: |
           git fetch origin master
           commits_headlines=$(git log origin/master..HEAD --pretty=format:"%s")
 
           if [[ "$commits_headlines" == *"chore(hydra): create image"* ]]; then
               echo "Docker image already built"
-              exit 0
+              echo "already_built=true" >> "$GITHUB_OUTPUT"
           else
               echo "Docker image not built yet"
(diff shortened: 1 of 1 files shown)


vponomaryov:
@fruch too many changes in here

fruch:
> @fruch too many changes in here

now fixed

.github/workflows/build-docker-image.yaml
```diff
@@ -43,56 +43,66 @@ jobs:
           persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal access token.
 
       - name: Check if docker image is already built
```
pehala, line 45: Could we extract this into a separate job instead, so we have simpler dependency chain? We could condition the entire build_image with this output. I find it better than having the same condition multiple times

fruch: maybe, not sure it worth while for asking 3 builder and not 2 for this specific things.

pehala (approved):

vponomaryov (approved):
LGTM

PR #15122, opened 2026-06-18
fix(argus): report InvalidInstanceType validation failures as TEST_ERROR

juliayakovlev:
Config validation errors (e.g. invalid instance types) were not reported to Argus because they occurred before Argus was initialized.

Two entry points are fixed:

1. sct.py provision_resources: init_and_verify_sct_config() can fail before the provisioning try/except block. Wrap it in a try/except that reports to Argus via the new _report_provision_error_to_argus helper. Refactor existing provisioning error handler to use the same helper.

2. sdcm/tester.py setUp: teardown_on_exception now attempts to initialize Argus when setUp fails before init_argus_run was called, so the failure event reaches Argus.

Also add regex patterns to _is_test_error() for:
- InvalidInstanceType (boto ClientError)
- 'is not supported/available in region' (validation assert/ValueError)

Add hasattr guard on argus_heartbeat_stop_signal in fixture_argus_finalize.

### Testing

- [x] provision AWS
- [x] provision GCE
- [x]  [scylla-enterprise-perf-regression-predefined-throughput-steps-i8g-tablets failed with correct error and correct ru status](https://argus.scylladb.com/tests/scylla-cluster-tests/0476a274-2bb1-4a35-8689-b15fb9d0f584)
- [x]  [scylla-enterprise-perf-regression-predefined-throughput-steps-i8g-tablets- instances created](https://argus.scylladb.com/tests/scylla-cluster-tests/cbecf0ef-ca3c-401e-846e-a07abfcd82ae)

### Reminders

### DIFF
--- a/sct.py
+++ b/sct.py
@@ -259,6 +259,49 @@ def cli(ctx):
         docker_hub_login(remoter=LOCALRUNNER)
 
 
+def _report_provision_error_to_argus(backend: str, exc: Exception):
+    """Report a provision-stage failure to Argus with TEST_ERROR status."""
+    test_config = get_test_config()
+    try:
+        params = SCTConfiguration()
+        test_id = params.get("test_id")
+        if test_id and test_id != "None":
+            test_config.set_test_id_only(test_id)
+        test_config.init_argus_client(params)
+    except Exception:  # noqa: BLE001
+        LOGGER.warning("Failed to initialize Argus client for error reporting", exc_info=True)
+        return
+
+    error_message = (
+        f"(TestFrameworkEvent Severity.CRITICAL) Failed to provision {backend} resources: {type(exc).__name__}: {exc}"
+    )
+    event_payload: RawEventPayload = {
+        "run_id": str(test_config.test_id()),
+        "severity": "CRITICAL",
+        "ts": time.time(),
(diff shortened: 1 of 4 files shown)


soyacz:
I have proposal (maybe a followup):
Make jenkins always create Argus run - in case we don't have it when finalizing SCT pipeline, create it and post console log to s3 (maybe an sct command) and link to Argus.
This way Zeus would always see argus run for each jenkins job and would have artifact to analyze the failure.

juliayakovlev:
> I have proposal (maybe a followup): Make jenkins always create Argus run - in case we don't have it when finalizing SCT pipeline, create it and post console log to s3 (maybe an sct command) and link to Argus. This way Zeus would always see argus run for each jenkins job and would have artifact to analyze the failure.

It's good idea for follow up task

fruch (approved):

fruch:
> I have proposal (maybe a followup): Make jenkins always create Argus run - in case we don't have it when finalizing SCT pipeline, create it and post console log to s3 (maybe an sct command) and link to Argus. This way Zeus would always see argus run for each jenkins job and would have artifact to analyze the failure.

so you you're suggesting collection of jenkins output on any failure.

PR #15124, opened 2026-06-18
feature(nemesis): Add precheck() hook to NemesisBaseClass 

pehala:
Implement the skip plan docs/plans/nemesis/nemesis-precheck.md. The precheck will be evaluated before execution, nemesis will be removed from the pool and sent as skipped to Argus (maybe we can in the future add a different mark to distinguish them).

Not currently present, but I will migrate some nemesis to use precheck here for verification. Rest is going to be done separately

### Testing

- [x] unittests
- [x] integration tests
- [x] Longevity https://argus.scylladb.com/tests/scylla-cluster-tests/24b0f3a1-b5d8-4fac-90cf-c53d75cf61eb
     - With skipped test at the start: https://argus.scylladb.com/tests/scylla-cluster-tests/b339d915-69eb-490e-a410-b42d496a787a/nemesis

### Reminders

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -254,6 +254,29 @@ class NemesisBaseClass(NemesisFlags, ABC):
         super().__init__()
         self.runner: "NemesisRunner" = runner
 
+    def precheck(self, node: BaseNode) -> str | None:
+        """
+        Static feasibility check evaluated ONCE before the nemesis is scheduled.
+
+        Return None  -> nemesis is runnable and kept in the rotation.
+        Return a str -> human-readable skip reason; nemesis is permanently excluded
+                        from the disruption list and will not run during this test.
+
+        Use ONLY for conditions that do not change during the test run:
+          - Test config, backend, or product edition
+            (e.g. cluster.params.get(...), _is_it_on_kubernetes(), node.is_enterprise)
+          - Scylla version, feature flags, or cluster-uniform node attributes
+            (e.g. ComparableScyllaVersion checks, is_tablets_feature_enabled(), node.distro)
+
+        No target node is selected yet — use the provided representative live
+        node for any cluster-wide probes.
+
+        Do NOT check dynamic state (data presence, live node counts, whether a
+        specific target node is busy or alone in its rack) — those belong in
+        disrupt().
+        """
+        return None
+
     @abstractmethod
     def disrupt(self):
         """
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -97,7 +97,7 @@ from sdcm.sct_events.group_common_events import (
 
 from sdcm.sct_events.loaders import CassandraStressLogEvent, ScyllaBenchEvent
 from sdcm.sct_events.nemesis import DisruptionEvent
-from sdcm.sct_events.system import InfoEvent
+from sdcm.sct_events.system import InfoEvent, TestFrameworkEvent
 from sdcm.sla.sla_tests import SlaTests
 from sdcm.utils.aws_kms import AwsKms
 from sdcm.utils import cdc
@@ -438,12 +438,15 @@ class NemesisRunner:
         excluded = self.precheck_nemesis()
         if excluded and not self.disruptions_list:
             reason_list = "; ".join(f"{name}: {reason}" for name, reason in excluded)
-            InfoEvent(
-                message=(
-                    f"All {len(excluded)} selected nemesis were excluded by precheck and will not run. "
-                    f"Reasons: {reason_list}"
-                ),
+            selector_message = f" Selector: {self.nemesis_selector!r}." if self.nemesis_selector else ""
+            TestFrameworkEvent(
+                source=self.__class__.__name__,
                 severity=Severity.CRITICAL,
+                message=(
+                    f"No runnable nemesis remain after precheck. "
+                    f"All {len(excluded)} selected nemesis were excluded before execution.{selector_message} "
+                    f"Exclusions: {reason_list}"
+                ),
             ).publish()
             return
         try:
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -435,6 +435,17 @@ class NemesisRunner:
         if interval:
             self.interval = interval * 60
         self.log.info("Interval: %s s", self.interval)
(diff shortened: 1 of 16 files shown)


pehala:
@claude review, focus on correctness and edgecases

pehala:
Initial tests seems to be working. I moved tablets enabled switch to precheck in AbortDecomissionMonkey
With tablets: https://argus.scylladb.com/tests/scylla-cluster-tests/fcf63972-a374-4555-9021-0261892e5a81/nemesis
   * The skips experienced in this run are dynamic and are due to running before the table exists
Without tablets: https://argus.scylladb.com/tests/scylla-cluster-tests/e6475a64-ba4d-411d-b065-01cef10a59a0/sct-events
   * Ran only once, returned error (The error needs to be critical and have better type) since all nemesis were skipped

pehala:
@scylladb/qa-maintainers This is ready for review. In Argus all skipped nemesis this way will show first in the list. There is no way currently how to differenciate between runtime skip and precheck skip in Argus. Longevity and nemesis test both work

pehala:
@scylladb/storage-qa Ready for review also from you

jsmolar (dismissed):
Love this feature. Design and implementation are clean.
One minor naming nit: `precheck()` returning `None` to mean "run this nemesis" is unintuitive — `None` feels like "nothing" rather than "proceed."

sdcm/nemesis/__init__.py
```diff
@@ -254,6 +254,29 @@ def __init__(self, runner: "NemesisRunner"):
         super().__init__()
         self.runner: "NemesisRunner" = runner
 
+    def precheck(self, node: BaseNode) -> str | None:
+        """
+        Static feasibility check evaluated ONCE before the nemesis is scheduled.
+
+        Return None  -> nemesis is runnable and kept in the rotation.
```
fruch, line 257: I do agree the action name is precheck, and the truthiness of the response is a bit vague like that (based on the method name only), would False gonna work as None ? 

Also this correlates to the current approach of rising exception to skip, I think extending it to report multiple skip reason could prove useful, maybe too mouthful, but would want to "fix" some reason, and fix the next on on the next run. Not for this PR.

fruch (dismissed):

pehala:
> One minor naming nit: `precheck()` returning `None` to mean "run this nemesis" is unintuitive — `None` feels like "nothing" rather than "proceed."

I agree, I just struggled to find the the right balance. I have to have the string there to have the reason, so standard boolean is out of the questions and as such I am missing the "Falsy" value there to make it look normal. Do you have any suggestions how it should be called?

sdcm/nemesis/__init__.py
pehala: I think False would work as well, I will try it out.



> Also this correlates to the current approach of rising exception to skip, I think extending it to report multiple skip reason could prove useful, maybe too mouthful, but would want to "fix" some reason, and fix the next on on the next run. Not for this PR.

I thought about it, but we dont have that many of those multi skips, so I decided to leave it out. Could be easily added in the followup

sdcm/nemesis/__init__.py
```diff
@@ -2020,6 +2057,63 @@ def build_disruptions_by_selector(self, nemesis_selector: str | None = None):
...
+        returns the list of (name, reason) exclusions. Each class is evaluated and
+        reported at most once — when nemesis_multiply_factor > 1 the disruptions_list
+        may contain multiple instances of the same class, but precheck() is invariant
+        so only the first instance triggers the report; subsequent copies are silently
+        pruned without duplicate Argus calls.
+        """
+        runnable = []
+        excluded: list[tuple[str, str]] = []
+        reported: set[str] = set()  # classes already reported to avoid duplicates
+        start_time = time.time()
```
cezarmoise, line 2041: Just FYI
If you submit all the pre-check skipped nemesis with the same start time, only one will be shown, until https://scylladb.atlassian.net/browse/ARGUS-150 is fixed.

cezarmoise (dismissed):
LGTM.

I'm ok with `None`.
`if skip_reason is None:` reads as "there is no reason to skip".

sdcm/nemesis/__init__.py
pehala: Damn, I cannot workaround it here though, prechecks should be quick so even if I moved the start_time it would be in the same second.

sdcm/nemesis/__init__.py
pehala: False wont work due to 
`if skip_reason is None:`

I think None is better because it is either string or nothing. I think `str | bool` is a type we should avoid and `str | None` is much better

pehala:
v3:
   * rebased
   * Fixed multiple prechecks when using nemesis_multiply_factor > 1
   * Fixed CategoricalNemesis
   * Fixed docstrings/plan

sdcm/nemesis/monkey/runners.py
```diff
@@ -208,7 +208,26 @@ def is_nonnegative_number(val):
...
+        """Prune infeasible members from disruption_distribution too, via disruptions_list.
+
+        CategoricalMonkey keeps its executable candidates in disruption_distribution rather
+        than disruptions_list, so the base precheck alone would leave infeasible weighted
+        candidates selectable by select_next_nemesis(). Mirroring the population into
+        disruptions_list lets the base pruning run once, then this resyncs the weights to
+        match the survivors.
+        """
+        excluded = super().precheck_nemesis()
+        self.disruption_distribution = (
```
fruch, line 226: this seems as abuse of this new introduced machinary, not sure I understand it

fruch (approved):

sdcm/nemesis/monkey/runners.py
fruch: but since no one is using CategoricalNemesis... not sure if it's important or not

pehala: It is a design flaw for sure, if designed correctly with CategoricalMonkey in mind we could support it in a more clean way (e.g. add more abstract method for runner which we could utilize for intercompatibility). I decided to not complicate design for a monkey that noone uses so I made the cleanest fix I could think of without complicating the machinery.

It only redoes self.disruption_distribution based on pruned nemesis (because we removed some of them as part of precheck), which I considered reasonable

cezarmoise (approved):

PR #15125, opened 2026-06-18
fix(2TB_high_disk_used_capacity.yaml): Added a supports_high_disk_utilization nemesis selector

yarongilor:
In order to avoid disk capacity issues for this high-disk-used-capacity configuration, only nemesis having supports_high_disk_utilization should be used. Preious runs encountered disk-usage errors running a nemesis of add-remove-mv, for example.

See a discussion in https://argus.scylladb.com/tests/scylla-cluster-tests/4e40f329-ecc7-4109-93df-a1edf7063719/discuss

### Testing

- [ ]

### Reminders

### DIFF
--- a/configurations/2TB_high_disk_used_capacity.yaml
+++ b/configurations/2TB_high_disk_used_capacity.yaml
@@ -24,3 +24,4 @@ sizing_db:  # no OCI match (min 16 vCPUs in OCI catalog)
   memory: '>=64'
 
 user_prefix: 'longevity-2tb-high-disk-used'
+nemesis_selector: 'supports_high_disk_utilization'


fruch:
@pehala at this rate, all of tier1 would be skipping those nemesis.

Is the issues those are surfacing even on someone's list of items to handle ?

pehala:
> @pehala at this rate, all of tier1 would be skipping those nemesis.
> 

They shouldnt, I think for majority of the cases we will simply put it under 50% so this doesnt happen (same as for the multidc case changed recently)

> Is the issues those are surfacing even on someone's list of items to handle ?

There will be a plan to tackle them.

yarongilor:
> > @pehala at this rate, all of tier1 would be skipping those nemesis.
> 
> They shouldnt, I think for majority of the cases we will simply put it under 50% so this doesnt happen (same as for the multidc case changed recently)
> 
> > Is the issues those are surfacing even on someone's list of items to handle ?
> 
> There will be a plan to tackle them.

A plan is added in https://github.com/scylladb/scylla-cluster-tests/pull/15147, and also a suggestion to resolve the limitation when requested.

pehala (approved):

fruch (approved):

PR #15132, opened 2026-06-18
fix(s3-uploader): parse du output by tab delimiter instead of G suffix

fruch:
Some du implementations (e.g. BusyBox) output size without the G suffix when using -BG flag, producing '1\ttotal' instead of '1G\ttotal'. Split by tab character which is the universal du output delimiter, then strip any trailing G suffix if present.

Fixes: SCT-445

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/utils/s3_remote_uploader.py
+++ b/sdcm/utils/s3_remote_uploader.py
@@ -69,7 +69,8 @@ def upload_remote_files_directly_to_s3(
         if not out:
             raise FileNotFoundError(f"Could not get the size of {files}. Possibly it does not exist.")
         total = out.split(b"\n")[-2]
-        return int(total.split(b"G")[0].decode())
+        size_str = total.split(b"\t")[0].decode().rstrip("G")
+        return int(size_str)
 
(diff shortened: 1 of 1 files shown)


soyacz (commented):
This function has more issues, I think worth to send agent to fix it (including name which says 'kb' while we want 'gb').
E.g. LLM claims this will not work anyway on BusyBox (Alpine)

fruch:
> This function has more issues, I think worth to send agent to fix it (including name which says 'kb' while we want 'gb'). E.g. LLM claims this will not work anyway on BusyBox (Alpine)

LLM wrote this commit :) I've extracted form the branch I was working on minicloud

we don't use alpine to run scylla...

soyacz:
> > This function has more issues, I think worth to send agent to fix it (including name which says 'kb' while we want 'gb'). E.g. LLM claims this will not work anyway on BusyBox (Alpine)
> 
> LLM wrote this commit :) I've extracted form the branch I was working on minicloud
> 
> we don't use alpine to run scylla...

you (or LLM) mentioned 'Some du implementations (e.g. BusyBox) ' :P

soyacz (approved):

PR #15147, opened 2026-06-19
docs(plans): add nemesis high disk utilization support plan

pehala:
Analyze all 15 nemesis marked supports_high_disk_utilization=False and classify each as inherent vs implementation-driven:
- 2 blocked by ScyllaDB bug #22088 (scrub/validation compaction)
- 1 inherent but sizing-tunable (MgmtRestore)
- 12 implementation-fixable: topology-shrink family via grow-before-shrink, MV/SI via dedicated bounded table, Manager backup via free-space guard

Register the plan in MASTER.md and progress.json.

### Testing

- [ ]

### Reminders


yarongilor (commented):
One possible optional way to handle this limitation:
Introduce a new configuration mode of "temporarily 'extend-cluster-on-nemesis' to avoid high-disk-usage limitation".
Thus when this new configuration applies:

- A previous configuration of `nemesis_selector: 'supports_high_disk_utilization'` can be removed from this test configuration.
- On running a nemesis with `supports_high_disk_utilization = true` , it'll first add the required number of new db nodes so that total used capacity is less than 50%. Then, after executing the nemesis, it will remove these extended nodes.

pehala:
> On running a nemesis with `supports_high_disk_utilization = true` , it'll first add the required number of new db nodes so that total used capacity is less than 50%. Then, after executing the nemesis, it will remove these extended nodes.

This beats the purpose of the tests, we want to test on higher utilization levels, because thats what we are advertising to our customers.

docs/plans/nemesis/nemesis-high-disk-utilization.md
```diff
@@ -0,0 +1,178 @@
...
+7. Every flag change is accompanied by regenerated `data_dir/nemesis.yml` + pipelines and a green
+   elasticity-90 validation run.
+
+## 4. Implementation Phases
+
+> Default behavior must be preserved for normal longevity. Where a behavior change (e.g. grow-first)
+> alters coverage, prefer a **dedicated high-disk variant class** or a **runtime disk-usage check**
+> over unconditionally changing the existing nemesis. See Needs-Investigation notes.
+
+### Phase 1 — Grow-before-shrink for the `removenode` family (Importance: High)
```
pehala: Instead we should unify with growshrinkcluster nemesis, we should use this opportunity to unify the decomissions/remove nodes paths and simplify it

docs/plans/nemesis/nemesis-high-disk-utilization.md
```diff
@@ -0,0 +1,178 @@
...
+
+### Phase 2 — Grow-before-shrink for the decommission family (Importance: Medium)
+- Apply the same add-before-shrink ordering to `disrupt_nodetool_decommission` /
+  `disrupt_nodetool_seed_decommission` (they already add a node *after*; reorder/guard for high disk).
+- **Needs Investigation:** overlap with `GrowShrinkClusterNemesis` semantics — confirm value vs. just
+  relying on GrowShrink for topology coverage at high disk. May be dropped if redundant.
+- **Dependencies:** Phase 1 (reuses the grow-first activation model).
+- **DoD:** decision recorded; if implemented, flags flipped + regenerated + validated.
+
+### Phase 3 — Bounded dedicated table for MV/SI nemesis (Importance: High)
```
pehala, line 152: This is something I am not sure of. creating MV/Si on 5% table doesnt seem good enough for most cases, not to mention we will have to fill it ourselves and ideally nemesis shouldnt be doing more stress.

I think this could stay, or we introduce new copies of these which will be weaker but can be ran

docs/plans/nemesis/nemesis-high-disk-utilization.md
```diff
@@ -0,0 +1,178 @@
...
+- Add a helper that creates a small dedicated keyspace/table with a capped row count (sized so the
+  resulting SI/MV fits comfortably in free headroom), used by `disrupt_create_index`,
+  `disrupt_add_remove_mv`, and `disrupt_kill_mv_building_coordinator` when running under high disk;
+  drop both table and view/index at the end. (`KillMVBuildingCoordinator` needs a tablets-enabled
+  dedicated table.)
+- **Dependencies:** none.
+- **DoD:** three nemesis flags → `True`; regenerated; elasticity-90 build+drop cycle stays within free
+  space.
+
+### Phase 4 — Manager backup free-space guard + scoping (Importance: Medium)
```
pehala: Backu on its own shouldnt take space, it hardlinks and prevents space from being freed, so if we backup and then delete, I think it might be fine
@cezarmoise WDYT?

docs/plans/nemesis/nemesis-high-disk-utilization.md
```diff
@@ -0,0 +1,178 @@
...
+- Add a pre-flight node free-space check; backup a bounded keyspace subset; ensure snapshot deletion is
+  prompt (already present in `_delete_existing_backups` / `delete_backup_snapshot`).
+- Enable in order: `MgmtBackupSpecificKeyspaces` → `MgmtBackup` → `ManagerRcloneBackup` /
+  `ManagerNativeBackup` (`runners.py:120/129`).
+- **Dependencies:** none.
+- **DoD:** flags flipped incrementally with a validation run between full-cluster variants.
+
+### Phase 5 — MgmtRestore sizing to free headroom (Importance: Low)
```
pehala, line 193: This one will have to go IMO. restore into live cluster was never officialy supported so this nemesis doesnt test any customer scenario. For the upcoming features, that enable it, this would have to rewritten anywa

docs/plans/nemesis/nemesis-high-disk-utilization.md
```diff
@@ -0,0 +1,178 @@
...
+- Change snapshot selection in `disrupt_mgmt_restore` (`__init__.py:2888`) from "≤5% of total partition"
+  to "≤ a safe fraction of *currently free* space" (use the existing `df` query). If not feasibly safe,
+  document as inherent and keep `False`.
+- **Dependencies:** none.
+- **DoD:** decision + (optional) flag flip + validation.
+
+### Phase 6 — `#22088` gating for scrub/validation (Importance: Medium)
```
pehala: Need to check the validaity of the bug again

yarongilor:
> > On running a nemesis with `supports_high_disk_utilization = true` , it'll first add the required number of new db nodes so that total used capacity is less than 50%. Then, after executing the nemesis, it will remove these extended nodes.
> 
> This beats the purpose of the tests, we want to test on higher utilization levels, because thats what we are advertising to our customers.

if it's interesting to test this list of nemeses, at least once, with a large dataset (let's say 2TB), then we can also consider a new Longevity test with a specific configuration of 4TB disk capacity (so only up to 50% disk used capacity).

pehala:
> if it's interesting to test this list of nemeses, at least once, with a large dataset (let's say 2TB), then we can also consider a new Longevity test with a specific configuration of 4TB disk capacity (so only up to 50% disk used capacity).

I dont follow, could you elaborate more why this would help?

jsmolar (dismissed):

cezarmoise (dismissed):

docs/plans/nemesis/nemesis-high-disk-utilization.md
cezarmoise: This is still open, https://github.com/scylladb/scylladb/issues/24345
But I didn't see it happening during the GCE / OCI tests.

PR #15152, opened 2026-06-21
fix(kernel-panic): suppress false KernelPanicEvent during intentional reboots

fruch:
## Summary

Fixes [SCT-459](https://scylladb.atlassian.net/browse/SCT-459) — False `KernelPanicEvent` on OCI backend after nemesis hard reboot.

## Problem

On OCI, a hard reboot transitions the instance through `STOPPED`/`STOPPING` states. `OCIKernelPanicChecker._check_instance_state()` treats any `STOPPED`/`STOPPING` state as a kernel panic, returning a synthetic message containing "Kernel panic". This causes a false positive `KernelPanicEvent` when nemesis operations (e.g. `RepairStreamingErrMonkey`) intentionally hard-reboot a node.

The test `longevity-100gb-4h-oci-test/51` failed on this false positive — the node rebooted cleanly with only XFS log recovery, but the event analyzer treated the `KernelPanicEvent` as a real failure.

## Fix

Add a `_reboot_in_progress` threading.Event to `BaseKernelPanicChecker` with `suspend_during_reboot()` / `resume_after_reboot()` methods. The checker's monitoring loop skips panic detection while the flag is set. `BaseNode.reboot()` sets the flag before rebooting and clears it in a `finally` block after the reboot completes (including SSH verification).

### Why this approach

- **Backend-agnostic** — any cloud provider where hard reboot causes transient STOPPED state gets the fix for free
- **Thread-safe** — uses `threading.Event` (same pattern as existing `_stop_event`/`_panic_detected`)
- **No string-matching fragility** — does not depend on matching specific cloud provider messages
- **Clean separation** — reboot code signals intent; checker respects it
- **Single integration point** — all `hard_reboot()` callers go through `BaseNode.reboot()`

## Changes

| File | Change |
|------|--------|
| `sdcm/kernel_panic_checker.py` | Add `_reboot_in_progress` event, `suspend_during_reboot()`/`resume_after_reboot()` methods, skip logic in `run()` |
| `sdcm/cluster.py` | Wrap reboot logic in `try/finally` with suspend/resume |
| `unit_tests/unit/test_kernel_panic_checker.py` | 4 new tests covering suppression, resume, and idempotency |

## Testing

- All 25 unit tests in `test_kernel_panic_checker.py` pass
- No LSP diagnostics errors
- Pre-commit hooks pass (ruff, formatting)

## Manual Testing Required

- [x] Run `longevity-100gb-4h-oci-test` with `RepairStreamingErrMonkey` to verify no false `KernelPanicEvent` on hard reboot
- [x] Verify real kernel panics are still detected after reboot completes (tested in unit tests but needs cloud validation)

[SCT-459]: https://scylladb.atlassian.net/browse/SCT-459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -1505,6 +1505,31 @@ class BaseNode(AutoSshContainerMixin):
     def uptime(self):
         return datetime.strptime(self.remoter.run("uptime -s", ignore_status=True).stdout.strip(), "%Y-%m-%d %H:%M:%S")
 
+    def _reboot_inner(self, hard, verify_ssh, uptime_changed):
+        """Execute the reboot sequence: trigger reboot, wait for uptime change, optionally verify SSH.
+
+        Args:
+            hard: If True, perform a hard reboot (power cycle). If False, perform a soft reboot via SSH.
+            verify_ssh: If True, wait for SSH connectivity to be restored after reboot.
+            uptime_changed: Callable returning True once the node's uptime reflects a fresh boot.
+        """
+        if hard:
+            self.log.debug("Hardly rebooting node")
+            with DbNodeLogger([self], "hard reboot node", target_node=self):
+                self.hard_reboot()
+        else:
+            self.log.debug("Softly rebooting node")
+            if not self.remoter.is_up(60):
+                raise RuntimeError("Target host is down")
+            with DbNodeLogger([self], "soft reboot node", target_node=self):
+                self.soft_reboot()
+
+        # wait until the reboot is executed
+        wait.wait_for(func=uptime_changed, step=10, timeout=60 * 45, throw_exc=True)
+
+        if verify_ssh:
+            self.wait_ssh_up()
+
     def reboot(self, hard=True, verify_ssh=True):
         pre_uptime = self.uptime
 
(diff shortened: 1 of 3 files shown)


.github/workflows/claude.yml
fruch, line 119: this change isn't part of this PR

sdcm/cluster.py
```diff
@@ -1533,22 +1533,28 @@ def uptime_changed():
...
-            with DbNodeLogger([self], "hard reboot node", target_node=self):
-                self.hard_reboot()
-        else:
-            self.log.debug("Softly rebooting node")
-            if not self.remoter.is_up(60):
-                raise RuntimeError("Target host is down")
-            with DbNodeLogger([self], "soft reboot node", target_node=self):
-                self.soft_reboot()
+        if self.kernel_panic_checker:
+            self.kernel_panic_checker.suspend_during_reboot()
```
fruch: maybe this should be also context manager, it would minimized the need changes here.

sdcm/kernel_panic_checker.py
```diff
@@ -36,6 +36,7 @@ def __init__(self, node_name: str, host: str | None = None, logdir: str | None =
         self.logdir = logdir
         self._stop_event = threading.Event()
         self._panic_detected = threading.Event()
+        self._reboot_in_progress = threading.Event()
```
fruch: should be named suspense, it might be cause of other reason then reboot of node.

.github/workflows/claude.yml
fruch: fixed

sdcm/kernel_panic_checker.py
```diff
@@ -94,6 +96,13 @@ def _save_console_output(self, output: str):
     def run(self):
         while not self._stop_event.is_set():
             try:
+                # Skip panic detection while suspended — cloud providers
+                # (especially OCI) report the instance as STOPPED during a hard reboot,
+                # which would otherwise trigger a false KernelPanicEvent.
```
pehala, line 101: Could we use event supression instead, same as we do with other errors during unavailabiulity?

sdcm/cluster.py
```diff
@@ -1505,6 +1505,24 @@ def restart_binary_protocol(self, verify_up=True):
     def uptime(self):
         return datetime.strptime(self.remoter.run("uptime -s", ignore_status=True).stdout.strip(), "%Y-%m-%d %H:%M:%S")
 
+    def _reboot_inner(self, hard, verify_ssh, uptime_changed):
```
pehala, line 1508: Could you add docstring for this new method?

sdcm/kernel_panic_checker.py
fruch: I rather have this then a regex based filter, if we can stop it at it's core, it's better the filtering it after the fact.

sdcm/cluster.py
fruch: added one

vponomaryov:
@fruch backport to 2026.2 (OCI has ScyllaDB images for this version)?

fruch:
> @fruch backport to 2026.2 (OCI has ScyllaDB images for this version)?

tried, it didn't reached that branch

PR #15174, opened 2026-06-23
fix(events): ignore audit unavailable errors globally

pehala:
## Summary
- add a global default event filter for audit write unavailable errors
- remove the nemesis-local audit ignore wrapper now that the filter is always on
- extend the events integration test to cover the audit login unavailable case
   - Mock SkipPerIssues to make the deterministic

Fixes https://scylladb.atlassian.net/browse/SCT-518

## Testing
- uv run pytest unit_tests/integration/test_events.py -k test_default_filters

### DIFF
--- a/sdcm/sct_events/setup.py
+++ b/sdcm/sct_events/setup.py
@@ -175,6 +175,17 @@ def enable_default_filters(sct_config: SCTConfiguration):
             r"Cannot achieve consistency level for cl ONE",
         ).publish()
 
+    # Audit writes can briefly fail with unavailable_exception during transient cluster
+    # disruptions. This is expected and should not fail the test.
+    # Decided to do a global ignore instead of ignoring on all of the nemesis
+    # https://scylladb.atlassian.net/browse/SCT-518
+    # https://scylladb.atlassian.net/browse/SCYLLADB-706
+    EventsSeverityChangerFilter(
+        new_severity=Severity.WARNING,
+        event_class=DatabaseLogEvent,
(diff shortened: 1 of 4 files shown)


sdcm/sct_events/setup.py
```diff
@@ -175,6 +175,14 @@ def enable_default_filters(sct_config: SCTConfiguration):
             r"Cannot achieve consistency level for cl ONE",
         ).publish()
 
+    # Audit writes can briefly fail with unavailable_exception during transient cluster
+    # disruptions. This is expected and should not fail the test.
```
fruch, line 179: can we have refs to the issues explaining it ?

pehala: Added

fruch (dismissed):
LGTM

sdcm/nemesis/__init__.py
```diff
@@ -3606,6 +3606,25 @@ def reboot_node(self, target_node, hard=True, verify_ssh=True):
         if self.tester.params.get("print_kernel_callstack"):
             save_kallsyms_map(node=target_node)
 
+    def disrupt_network_start_stop_interface(self):
```
cezarmoise: This shouldn't be here. Methods is not used anywhere and it's identical to `StopStartInterfacesNetworkMonkey`

pehala: Good catch, it comes from bad rebase it seems :/

cezarmoise (approved):
LGTM

PR #115, opened 2016-06-30
Introduce a simple regression test

kongove:
A simple regression test to use cassandra-stress as workloads, it contains write/read/mixed stresses.
We can do result compare and analysis in future patches.

PR #15177, opened 2026-06-23
fix(nemesis): increase CREATE_MV and CREATE_INDEX soft timeouts to 5 hours

yarongilor:
Observed CREATE_MV taking ~15453s and CREATE_INDEX (wait for index build) taking ~15857s, both exceeding the previous 14400s (4h) soft timeout and triggering SoftTimeoutEvent followed by FailedResultEvent from Argus validation.

Increase the soft timeout from 14400s to 18000s (5h) for all three adaptive_timeout calls in disrupt_create_index, disrupt_add_remove_mv, and disrupt_add_drop_mv_with_node_restarts.

Fixes: https://scylladb.atlassian.net/browse/SCT-353

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -4706,7 +4706,7 @@ class NemesisRunner:
                 raise UnsupportedNemesis("Tried to create already existing index. See log for details")
             try:
                 with adaptive_timeout(
-                    operation=Operations.CREATE_INDEX, node=self.target_node, timeout=14400
+                    operation=Operations.CREATE_INDEX, node=self.target_node, timeout=18000
                 ) as timeout:
                     with self.action_log_scope("Wait for index to be built"):
                         wait_for_index_to_be_built(self.target_node, ks, index_name, timeout=timeout * 2)
@@ -4769,7 +4769,7 @@ class NemesisRunner:
                         self.target_node.run_nodetool(sub_cmd="repair -pr")
                     with (
                         adaptive_timeout(
-                            operation=Operations.CREATE_MV, node=self.target_node, timeout=14400
+                            operation=Operations.CREATE_MV, node=self.target_node, timeout=18000
                         ) as timeout,
                         self.action_log_scope(
                             f"Wait for {ks_name}.{view_name} materialized view to be built on "
@@ -5294,7 +5294,7 @@ class NemesisRunner:
                         )
                         break
 
-                with adaptive_timeout(operation=Operations.CREATE_MV, node=working_node, timeout=14400) as timeout:
+                with adaptive_timeout(operation=Operations.CREATE_MV, node=working_node, timeout=18000) as timeout:
                     wait_for_view_to_be_built(working_node, ks_name, view_name, timeout=timeout * 2)
 
                 with self.cluster.cql_connection_patient(node=working_node, connect_timeout=600) as session:


unit_tests/unit/nemesis/monkey/test_network.py
fruch, line 111: how this is related to this PR ?

fruch (changes requested):
https://scylladb.atlassian.net/browse/SCT-353 is a bunch of reports from all kind of causes and release.

I don't see in any of those the analysis that suggests we should wait longer, but we should focus of what is slowing it down (disk/cpu/tombstones), and fix the source of the problem, waiting more is waste of our time and resources.

we doing work to do those wait by factor of something base on feature, we should align those as well to something, it should be increased on every slight issue.

unit_tests/unit/nemesis/monkey/test_network.py
yarongilor: not related. not here anymore.

yarongilor:
> https://scylladb.atlassian.net/browse/SCT-353 is a bunch of reports from all kind of causes and release.
> 
> I don't see in any of those the analysis that suggests we should wait longer, but we should focus of what is slowing it down (disk/cpu/tombstones), and fix the source of the problem, waiting more is waste of our time and resources.
> 
> we doing work to do those wait by factor of something base on feature, we should align those as well to something, it should be increased on every slight issue.

@fruch ,

1. There are some long on-going/stuck discussions about MV performance, like in https://scylladb.atlassian.net/browse/SCYLLADB-2851.
2. The issue is reproduced across multiple different scylla versions, so it's not a new regression issue on master or something.
3. I think at the moment, no point continue getting these errors failing tests since it doesn't have any added value. And affects multiple versions/pipelines.
4. i can open a new task to deeper investigate the add-remove-mv nemesis and it's results to see a clearer pattern of the failure.
5. This PR is not too dramatic, only allow MV building to last a bit more than the current 4 hours (up to 5 hours).
6. Some reproduction reports on https://scylladb.atlassian.net/browse/SCT-353 are not relevant so i added a comment for that as well.
7. We might want to discuss with Dev what would be an agreed way to prove a real MV performance issue. 
8. So we have to decide if it's better to merge this one or put on hold until getting more resolutions.

fruch:
> > https://scylladb.atlassian.net/browse/SCT-353 is a bunch of reports from all kind of causes and release.
> > 
> > I don't see in any of those the analysis that suggests we should wait longer, but we should focus of what is slowing it down (disk/cpu/tombstones), and fix the source of the problem, waiting more is waste of our time and resources.
> > 
> > we doing work to do those wait by factor of something base on feature, we should align those as well to something, it should be increased on every slight issue.
> 
> @fruch ,
> 
> 1. There are some long on-going/stuck discussions about MV performance, like in https://scylladb.atlassian.net/browse/SCYLLADB-2851.
> 2. The issue is reproduced across multiple different scylla versions, so it's not a new regression issue on master or something.
> 3. I think at the moment, no point continue getting these errors failing tests since it doesn't have any added value. And affects multiple versions/pipelines.
> 4. i can open a new task to deeper investigate the add-remove-mv nemesis and it's results to see a clearer pattern of the failure.
> 5. This PR is not too dramatic, only allow MV building to last a bit more than the current 4 hours (up to 5 hours).
> 6. Some reproduction reports on https://scylladb.atlassian.net/browse/SCT-353 are not relevant so i added a comment for that as well.
> 7. We might want to discuss with Dev what would be an agreed way to prove a real MV performance issue. 
> 8. So we have to decide if it's better to merge this one or put on hold until getting more resolutions.

It's continuing now, you are just getting notifications about the soft limit
I.e. it tells exactly how long it took above the soft timeout

Charging this would be to bury the discussions.

It's clearly in some of the reproduce runs, that the case doesn't have enough CPU, and the MV is probably doesn't have enough resources to be able to do it in a timely fashion, we should run longevity with a close to 100% CPU utilization, its one of many things that wouldn't work.

yarongilor:
Findings — SCT-353 reproduction check across scylla-master/tier1 and 2026.3, since 2026-06-20
Short answer: SCT-353's specific signature did not reproduce. I checked every previously-investigated scylla-master/tier1 and scylla-2026.3 (tier1+tier2) run after 2026-06-20 that executed AddRemoveMvNemesis:

AddRemoveMvNemesis ran and completed normally (no soft timeout), in scope:

longevity-twcs-48h-test #130, #132, #133 (scylla-master/tier1) and #4, #5 (scylla-2026.3/vnodes/tier1) — MV build finished in ~2h47m–2h51m, well under the 14400s (4h) soft timeout. These runs instead hit an unrelated CommitLogCheckErrorEvent ("free segments drop to zero") issue during the MV build, tracked under SCYLLADB-2876/SCT-231 — not a timeout.
AddRemoveMvNemesis failed for a different, unrelated bug (not a timeout):

gemini-1tb-10h-test #65 (2026-06-20, shortly after your run) and #72 (2026-08-22) — both failed instantly with Error from server: code=2200 [Invalid query] "Cannot use Duration column '&lt;col&gt;' in PRIMARY KEY of materialized view", tracked under open SCT-393.
One different CREATE_MV soft timeout did occur, but it is not SCT-353:

longevity-mv-si-4days-streaming-test #115 (scylla-master/tier1, nemesis window 2026-08-29T19:06Z–2026-08-30T03:13Z): SoftTimeoutEvent: operation 'CREATE_MV' is finished and took 28807.8s (soft timeout was set to 14400.0s) on sec_index.users_view.
Same numeric symptom class as SCT-353 (CREATE_MV soft timeout), but a different root cause: node logs show view_building_worker stuck in an infinite retry loop on a staging-SSTable filesystem error (link failed: No such file or directory), concurrent with tablet-cleanup compaction on the target view. This was matched Likely Duplicate to open SCYLLADB-4022 (related: SCYLLADB-3372, closed SCYLLADB-2312), already commented/linked by a prior investigation — not filed against SCT-353.
SCT-353 itself is scoped to the large-dataset build-slowness mechanism on scylla_bench.test (longevity-twcs-48h-test); this #115 occurrence built a much smaller MV on a different keyspace and stalled via a distinct filesystem-error/infinite-retry path, not sheer build slowness.
No other jobs I have prior investigation coverage for (scale-schema, elasticity, alternator, rolling-upgrade, longevity-mv-synchronous-updates, longevity-large-collections, etc.) showed AddRemoveMvNemesis reproducing a CREATE_MV soft timeout in this window.

Closing this PR for now.

PR #15180, opened 2026-06-23
fix(seeds): change default seeds_selector from 'all' to 'random' 3 nodes

roydahan:
## Summary

- Changes the default seed selection from all nodes to up to 3 random seed nodes
- Caps `seeds_num` by the current cluster size so 1-node/2-node tests remain valid with the default `seeds_num: 3`
- Makes `set_seeds()` replace the seed set instead of accumulating stale seed flags across reselection
- Adds unit coverage for capped random seed selection and seed flag replacement

## Motivation

According to the [ScyllaDB official docs on seed nodes](https://opensource.docs.scylladb.com/stable/kb/seed-nodes.html):

> Since ScyllaDB OSS 4.3 / Enterprise 2021.1, a seed node is **only used as the initial contact point** for a new node joining the cluster to discover ring topology. Once the nodes have joined the cluster, the seed node has no function.

The previous default of `seeds_selector: "all"` marks **every** node in the cluster as a seed. This is unnecessary for modern ScyllaDB and was explicitly forbidden in older versions, where at least one non-seed node was required.

Using only one seed is minimal, but it is fragile in SCT: several nemesis operations can terminate, decommission, remove, or replace the seed node. If that happens before a new node is added, SCT may no longer have a valid seed node configured for the new node bootstrap.

## Change

| | Before | After |
|---|---|---|
| `seeds_selector` default | `"all"` | `"random"` |
| `seeds_num` default | `1` (ignored when selector is `"all"`) | `3` |

With `seeds_selector: "random"` and `seeds_num: 3`, SCT keeps a bounded set of seed nodes that is more resilient to nemesis operations than a single seed, while avoiding the unnecessary all-nodes-as-seeds configuration.

## Small clusters

`seeds_num` now behaves as a maximum for `first`/`random` selection. If a cluster has fewer than 3 nodes, SCT selects all available nodes instead of failing config validation or `random.sample()`.

## Testing

- `uv run sct.py unit-tests -t unit/test_seed_selector.py`

### DIFF
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -47,8 +47,8 @@ nemesis_add_node_cnt: 3
 
 nemesis_filter_seeds: false
 
-seeds_selector: "first"
-seeds_num: 1
+seeds_selector: "random"
+seeds_num: 3
 
 instance_provision: "spot"
 
--- a/defaults/test_default.yaml
+++ b/defaults/test_default.yaml
@@ -47,7 +47,7 @@ nemesis_add_node_cnt: 3
 
 nemesis_filter_seeds: false
 
-seeds_selector: "all"
+seeds_selector: "first"
 seeds_num: 1
 
 instance_provision: "spot"
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5392,6 +5392,7 @@ class BaseScyllaCluster:
                 throw_exc=True,
             )
         else:
+            seeds_num = min(seeds_num, len(self.nodes))
             if seeds_selector == "random":
                 selected_nodes = random.sample(self.nodes, seeds_num)
             # seeds_selector == 'first'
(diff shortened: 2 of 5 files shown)


roydahan:
Still need to verify that in case the seed node was terminated/removed/decommissioned by a nemesis, we replace the seed.

In the past we had special handling for seed nodes, I don't know if it's still the case.

fruch:
> Still need to verify that in case the seed node was terminated/removed/decommissioned by a nemesis, we replace the seed.
> 
> In the past we had special handling for seed nodes, I don't know if it's still the case.

We don't, but this `first` needs to be reviewed and tested before enabling it by default.

We should add one more option to select 2 or 3, since we can't greentree the one selected will be up (parallel nemesis case, even that we don't do topology changes in parallel), since the safe recommended option would be to have couple of nodes as contact points for fallback if needed (even if it's a short while)

PR #15183, opened 2026-06-23
fix(azure): raise SpotTerminationEvent when spot VM evicted during provisioning

soyacz:
Fixes https://scylladb.atlassian.net/browse/SCT-530

## Problem (SCT-530)

An Azure spot VM can be accepted by ARM, report `Creating` for a few minutes, then be deleted asynchronously by Azure on spot eviction / capacity-allocation failure (`evictionPolicy=Delete`). After that, `virtual_machines.get()` returns `404 ResourceNotFound`.

Previously the 404 was retried until the stuck-VM timeout and then re-raised as an **unclassified `TestFrameworkEvent` ERROR**, hiding the real cause and (with `instance_provision_fallback_on_demand: false`) failing `setUp` with a misleading error.

## Fix

`_wait_until_provisioned`, for spot pricing, now:
- publishes a `SpotTerminationEvent` — already matched by `tester._is_test_error`, so the run classifies as an infra `TEST_ERROR` rather than a Scylla bug;
- raises `OperationPreemptedError`, reusing the existing spot → on-demand fallback path.

On-demand 404s re-raise unchanged.

## Notes

We still poll for the full timeout duration before classifying the 404 as an eviction, for simplicity — this happens usually just once during first cluster setup, so the wasted wait is not worth optimizing. Polling the whole timeout is also robust against a transient propagation 404 right after create: a real eviction is a 404 that persists for the entire timeout, while a too-early 404 resolves on a later poll (and the create poller is already awaited once before the first `get()`, so the VM exists by the time we start polling).

## Tests

- fake Azure service gained `disappear_after_polls` to simulate a VM that vanishes mid-provisioning;
- added a spot-eviction case (asserts `OperationPreemptedError` + one `SpotTerminationEvent`) and an on-demand-404 case (asserts the 404 propagates, no event);
- updated `_redeploy_stuck_node` call assertions for the new `pricing_model` argument.

### DIFF
--- a/sdcm/provision/azure/provisioner.py
+++ b/sdcm/provision/azure/provisioner.py
@@ -213,13 +213,13 @@ class AzureProvisioner(Provisioner):
                         message="recovering stuck VM (redeploy, recreate on failure)",
                         severity=Severity.NORMAL,
                     ).publish_or_dump(default_logger=LOGGER)
-                    if not self._redeploy_stuck_node(vm_name):
+                    if not self._redeploy_stuck_node(vm_name, pricing_model):
                         LOGGER.warning(
                             "Redeploy did not rescue VM %s; deleting it to recreate on fresh capacity", vm_name
                         )
                         self._delete_stuck_node(vm_name)
 
-    def _redeploy_stuck_node(self, name: str) -> bool:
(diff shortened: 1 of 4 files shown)


dimakr (approved):
LGTM

PR #15195, opened 2026-06-24
improvement(test-cases): retune longevity-change-cluster-size-by-2-times

cezarmoise:
This test had 100% load.
image
causing high latencies
image
which let to timeouts
> (CassandraStressLogEvent Severity.CRITICAL) period_type=one-time event_id=a67452dc-8a19-4e5e-a20a-783acf3f0fe4 during_nemesis=GrowShrinkClusterNemesis: type=OperationOnKey regex=Operation x10 on key\(s\) \[ line_number=8711 node=Node raft-double-cluster-size-with-schem-loader-node-a5a85827-0-1 [35.245.79.88 | 10.150.15.211] (Type: e2-standard-2) (rack: b)
[2026-05-19 13:52:18.210] java.io.IOException: Operation x10 on key(s) [3930344b4f4f4b4c3331]: Error executing: (WriteTimeoutException): Cassandra timeout during SIMPLE write query at consistency QUORUM (2 replica were required but only 1 acknowledged the write)

- reduce the load by reducing the number of threads
- use smaller instances to reduce costs
  - also don't use i8ge family, which caused the GrowShrinkNemesis to be skipped a lot due to capacity issues
  - don't use arm instances, as they may run into capacity issues (e.g. https://argus.scylladb.com/tests/scylla-cluster-tests/dbebd513-240a-46c1-b191-d1b803cc6e8b/nemesis)
- re-write the stress commands using multi-line strings for clarity (and better diffs in the future).

Closes: SCT-380
Closes: SCT-497

### Testing

- [x] https://argus.scylladb.com/tests/scylla-cluster-tests/a25db33d-8623-4bf1-bf0a-ddafae25d444

Load
image

Disk Usage
image

Latencies
Screenshot From 2026-07-23 13-40-31

### Reminders

### DIFF
--- a/data_dir/cs_mv_si_keyspace1.yaml
+++ b/data_dir/cs_mv_si_keyspace1.yaml
@@ -29,25 +29,25 @@ extra_definitions:
 columnspec:
   - name: key
     size: fixed(10)
-    population: uniform(1..10M)
+    population: uniform(1..2M)
 
   - name: key2
     cluster: uniform(1..100)
 
   - name: c0
-    size: fixed(1024)
+    size: fixed(128)
 
   - name: c1
-    size: fixed(1024)
+    size: fixed(128)
 
   - name: c2
-    size: fixed(1024)
+    size: fixed(128)
 
   - name: c3
-    size: fixed(1024)
+    size: fixed(128)
 
   - name: c4
-    size: fixed(1024)
+    size: fixed(128)
 
 
 ### Batch Ratio Distribution Specifications ###
--- a/test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
+++ b/test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
@@ -1,22 +1,65 @@
 test_duration: 900
 
-prepare_write_cmd:  ["cassandra-stress write cl=QUORUM n=20971520 no-warmup -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=80 -pop seq=1..20971520 -col 'n=FIXED(10) size=FIXED(512)' -log interval=5",
-                     "cassandra-stress user profile=/tmp/cs_mv_si_keyspace1.yaml n=3000000 no-warmup ops'(insert=1)' -mode cql3 native -rate threads=40 -log interval=5"
-                    ]
+prepare_write_cmd:
+  - >-
+    cassandra-stress write cl=QUORUM no-warmup
+    n=2M
+    -pop seq=1..2M
+    -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)'
+    -col 'n=FIXED(5) size=FIXED(128)'
+    -mode cql3 native
+    -log interval=5
+    -rate threads=2
+  - >-
+    cassandra-stress user profile=/tmp/cs_mv_si_keyspace1.yaml no-warmup ops'(insert=1)'
+    n=2M
+    -mode cql3 native
+    -log interval=5
+    -rate threads=2
 
-stress_cmd: ["cassandra-stress write cl=QUORUM duration=720m no-warmup -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=40 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(10) size=FIXED(512)' -log interval=5",
-             "cassandra-stress read  cl=QUORUM duration=720m no-warmup -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)' -mode cql3 native -rate threads=40 -pop 'dist=uniform(1..20971520)' -col 'n=FIXED(10) size=FIXED(512)' -log interval=5",
-             "cassandra-stress user profile=/tmp/cs_mv_si_keyspace1.yaml duration=700m no-warmup ops'(insert=1,read=1,mv_read1=1,mv_read2=1)' -mode cql3 native -rate threads=40 -log interval=5",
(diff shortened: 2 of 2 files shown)


test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
```diff
@@ -1,21 +1,67 @@
...
-# Example: i8ge.2xlarge is currently supported on us-east-1b,us-east-1c,us-east-1d. And not on us-east-1a.
-instance_type_db: 'i8ge.2xlarge'
+sizing_db:
+  # Use >=4 (not exact 4) so OCI and GCE can match:
+  # OCI DenseIO starts at 16 vCPU .
+  # GCE z3-highmem starts at 8 vCPUs.
+  # AWS/Azure still resolve to their 4-vCPU instances (smallest fitting).
+  vcpu: ">=4"
+  memory: ">=32"
+  # Force x86_64 arch to avoid AWS i8g family, which may run into capacity issues
```
pehala: I would not do anything to discourage i8g, it is after all, our most propagated instance type

cezarmoise: image

cezarmoise: @pehala btw, shouldn't it kill the test after 3+ skips?

pehala: It should, no idea why it doesnt in this case

pehala: My point stands though, ofc it didnt get any nodes. If you try every 5 seconds for half an hour it wont be better

cezarmoise:
ping @scylladb/storage-qa

test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
pehala: This is not resolved. What you provided is not proof i8g is probglematic, given you tried every few seconds for up to an hour

cezarmoise: That's exactly why it's problematic. This test has 2 nemesis threads. One of them runs GrowShrinkCluster. If there is no capacity, then it does nothing, and the `parallel` part of the test is moot.
It tries so often by design, when a nemesis is skipped, the `nemesis_interval` is ignored.

pehala: This can happen with every instance type, right now it might be i8g, later it will be something else, will you continue to create patches to fixate the test to one instance which at the moment doesnt fail as often?

It seems like a more structural problem, not something we should hotfix

cezarmoise: removed

pehala (dismissed):

data_dir/cs_mv_si_keyspace1.yaml
pehala, line 1: Is this used somewhere else as well? e.g. the Tier 1 test?

test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
```diff
@@ -1,22 +1,65 @@
...
+prepare_write_cmd:
+  - >-
+    cassandra-stress write cl=QUORUM no-warmup
+    n=2M
+    -pop seq=1..2M
+    -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)'
+    -col 'n=FIXED(5) size=FIXED(128)'
+    -mode cql3 native
+    -log interval=5
+    -rate threads=2
```
yarongilor, line 12: why decreasing from a total of 80 to 2 threads?

data_dir/cs_mv_si_keyspace1.yaml
```diff
@@ -29,25 +29,25 @@ extra_definitions:
 columnspec:
   - name: key
     size: fixed(10)
-    population: uniform(1..10M)
+    population: uniform(1..2M)
```
yarongilor, line 32: does this test passes ok in previous scylla versions? is it a regression related to tablets? is it tested with vnodes and passed ok (or is it less interesting)?
i see it also used in jenkins-pipelines/oss/vnodes/tier2/longevity-double-cluster-with-schema-changes-vnodes.jenkinsfile

test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
```diff
@@ -1,22 +1,65 @@
...
+
+stress_cmd:
+  - >-
+    cassandra-stress write cl=QUORUM no-warmup
+    duration=720m
+    -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3) compaction(strategy=LeveledCompactionStrategy)'
+    -col 'n=FIXED(5) size=FIXED(128)'
+    -mode cql3 native
+    -log interval=5
+    -rate threads=1
```
yarongilor, line 28: why decreasing from 40 to 1 thread?

data_dir/cs_mv_si_keyspace1.yaml
yarongilor: it is also used in jenkins-pipelines/oss/vnodes/tier2/longevity-double-cluster-with-schema-changes-vnodes.jenkinsfile

test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
cezarmoise: That was what was needed to reduce the load. 
MV/SI take up a lot of processing power, so this was needed to get reasonable write latencies.

test-cases/longevity/longevity-change-cluster-size-by-2-times.yaml
cezarmoise: That was what was needed to reduce the load. 
MV/SI take up a lot of processing power, so this was needed to get reasonable write latencies.

Issue #7969, opened 2024-07-14
adjust disrupt_bootstrap_streaming_error nemesis  to check on each node that node is removed from all node


temichus:
ScyllaDB issue - https://github.com/scylladb/scylladb/issues/19421

resolution of the original issue: update the nemesis, ScyllaDB works as expected here.

this is still a P3 Scylla issue - so this adjustment would be easy to remove when/if the issue is fixed

PR #15207, opened 2026-06-25
fix(nemesis): BootstrapStreamingErrorNemesis rebootstrap race condition - don't run removenode in bootstrapping state

yarongilor:
Fixes: https://scylladb.atlassian.net/browse/SCT-223

The rebootstrap after a mid-bootstrap kill was racing with the raft topology cleanup. When a node is killed mid-bootstrap, its entry remains in BOOTSTRAPPING state in the raft topology coordinator. The previous code issued removenode immediately (ineffective against a BOOTSTRAPPING node) and started the rebootstrap without verifying the stale entry was actually purged — causing the coordinator to reject the fresh join with 'A node with address X already exists'.

Fix by adding two strict synchronization points:

1. _wait_for_host_id_not_bootstrapping(): called before each removenode, blocks until the coordinator transitions the stale entry out of BOOTSTRAPPING state (up to 180s, throw_exc=True). This ensures the subsequent removenode is effective.

2. _verify_node_removed_from_topology(): called after prepare_node_for_rebootstrap() and before the rebootstrap starts, blocks until the bootstrap node's IP is confirmed gone from system.cluster_status (up to 300s, throw_exc=True). This is the final safety gate preventing a doomed rebootstrap attempt.

Both waits use throw_exc=True to fail fast with a clear error message instead of silently proceeding into a guaranteed-to-fail rebootstrap.

### Testing

- https://argus.scylladb.com/tests/scylla-cluster-tests/321cffb1-da7e-442e-80f8-139698cc443b

### Reminders

### DIFF
--- a/sdcm/utils/raft/common.py
+++ b/sdcm/utils/raft/common.py
@@ -15,7 +15,7 @@ from sdcm.wait import wait_for
 
 from sdcm.sct_events.group_common_events import decorate_with_context, ignore_ycsb_connection_refused
 from sdcm.utils.parallel_object import ParallelObject
-from sdcm.utils.raft import get_node_status_from_system_by
+from sdcm.utils.raft import get_node_status_from_system_by, NodeState
 from sdcm.cluster import BaseMonitorSet, NodeSetupFailed, BaseScyllaCluster, BaseNode
 from sdcm.exceptions import RaftTopologyCoordinatorNotFound
 from sdcm.rest.storage_service_client import StorageServiceClient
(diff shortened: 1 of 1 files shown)


pehala:
Can you elaborate more on why this is not in fact a scylla issue? From the PR description is sounds more like an workaround on our side, not a bug

yarongilor:
> Can you elaborate more on why this is not in fact a scylla issue? From the PR description is sounds more like an workaround on our side, not a bug

@pehala , running a `removenode` command for a node in a state of `bootstrapping` is not valid.
See https://docs.scylladb.com/manual/stable/operating-scylla/nodetool-commands/removenode.html:
`This command allows you to remove a node from the cluster when the status of the node is Down Normal (DN) `.

So this test issue has to be addressed.
It could be also there's some indirect scylla P3 issue as in https://github.com/scylladb/scylladb/issues/19421 and https://github.com/scylladb/scylla-cluster-tests/issues/7969

yarongilor:
The nemesis passed ok in https://argus.scylladb.com/tests/scylla-cluster-tests/321cffb1-da7e-442e-80f8-139698cc443b

pehala (approved):

cezarmoise (approved):
LGTM

PR #15208, opened 2026-06-25
fix(tier1): Restore tuned tier1 instances

pehala:
For tests that were tuned to specific disk usage, restore the hardcoded instances so the tuning is adhered to now and in the future. I do not want these tests changing without us knowing. Leave the sizing info so it can be still used cross cloud, just the tuning probably wont be exact, but that is expected.

To make these cases use dynamic sizing, we need to parametrize command based on disk size, which is possible but will require additional effort (plan in progress)

### Testing

- [x] none

### DIFF
--- a/configurations/2TB_high_disk_used_capacity.yaml
+++ b/configurations/2TB_high_disk_used_capacity.yaml
@@ -18,6 +18,7 @@ stress_cmd: ["cassandra-stress mixed cl=QUORUM duration=6800m -schema 'replicati
              "cassandra-stress user profile=/tmp/cs_mv_profile.yaml ops'(insert=3,read1=1,read2=1,read3=1)' cl=QUORUM duration=6800m -mode cql3 native -rate threads=10",
              "cassandra-stress read cl=QUORUM duration=6800m -mode cql3 native  -rate threads=10 -pop seq=1..1840000000  -log interval=5 -col 'size=FIXED(200) n=FIXED(5)'" ]
 
+instance_type_db: 'i8g.2xlarge'
 # i8g.2xlarge: 8 vCPU, 64 GB RAM, 1x1,875 GB NVMe SSD (~1,750 GiB formatted)
 sizing_db:  # no OCI match (min 16 vCPUs in OCI catalog)
   vcpu: 8
--- a/test-cases/features/elasticity/elasticity-90-percent-with-nemesis.yaml
+++ b/test-cases/features/elasticity/elasticity-90-percent-with-nemesis.yaml
@@ -17,13 +17,7 @@ pre_create_keyspace: "CREATE KEYSPACE IF NOT EXISTS keyspace1 WITH replication =
 n_db_nodes: 6
 n_loaders: 4
 nemesis_grow_shrink_instance_type: 'i8g.2xlarge'
-
-sizing_db:  # no GCE/OCI match (DenseIO min 16 vCPUs on OCI, z3 min 8 vCPUs on GCE)
-  vcpu: 2
-  memory: '>=16'
-sizing_loader:
-  vcpu: 8
-  memory: '>=8'
+instance_type_db: 'i8g.large'
 
 nemesis_class_name: 'SisyphusMonkey'
 nemesis_selector: 'supports_high_disk_utilization'


cezarmoise (dismissed):

configurations/2TB_high_disk_used_capacity.yaml
```diff
@@ -18,6 +18,7 @@ stress_cmd: ["cassandra-stress mixed cl=QUORUM duration=6800m -schema 'replicati
              "cassandra-stress user profile=/tmp/cs_mv_profile.yaml ops'(insert=3,read1=1,read2=1,read3=1)' cl=QUORUM duration=6800m -mode cql3 native -rate threads=10",
              "cassandra-stress read cl=QUORUM duration=6800m -mode cql3 native  -rate threads=10 -pop seq=1..1840000000  -log interval=5 -col 'size=FIXED(200) n=FIXED(5)'" ]
 
+instance_type_db: 'i8g.2xlarge'
```
yarongilor, line 21: perhaps the naming of `2TB_high_disk_used_capacity.yaml` is not ideal, since it really just defines a 2TB dataset, that can possibly be used for multiple setups.

perhaps it would make more sense to hardcode the instance type in `test-cases/longevity/longevity-1TB-5days-authorization-and-tls-ssl.yaml` that is the base-configuration of this "extention" file.

and also , it's less meaningful to set instance where no `n_db_nodes` parameter is found and is taken from another file.

test-cases/longevity/longevity-twcs-48h.yaml
```diff
@@ -38,6 +38,7 @@ stress_read_cmd: [
 
 n_db_nodes: 6
 n_loaders: 3
+instance_type_db: 'i8ge.xlarge'
```
roydahan: Why is it needed?
What's tuned in this PR?

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
```diff
@@ -63,6 +63,7 @@ n_loaders: '2 2'
 
 region_aware_loader: true
 simulated_racks: 0
+instance_type_db: 'i7i.xlarge'
```
roydahan: I don't think it should be set, it doesn't really significant if the size is bigger.

configurations/2TB_high_disk_used_capacity.yaml
roydahan, line 1: Where is this configuration being used?
What longevities? which tests?

roydahan (dismissed):

configurations/2TB_high_disk_used_capacity.yaml
pehala: Then separate this.

configurations/2TB_high_disk_used_capacity.yaml
pehala: https://github.com/scylladb/scylla-cluster-tests/blob/master/jenkins-pipelines/oss/tier1/longevity-2tb-5days.jenkinsfile

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
pehala: @cezarmoise As a owner of the testcase, do you agree?

test-cases/longevity/longevity-twcs-48h.yaml
pehala: @cezarmoise As a owner of the testcase, can you elaborate more?

configurations/2TB_high_disk_used_capacity.yaml
roydahan: Why is it overriding a test that is called 1TB with a 2TB configuration?
Also, I don't understand why we need to pin it to a specific type.

fruch: 1) it was set to i7i as we include i7i support to test it more regularly
2) someone scale up the test with more data to get to more disk utilization (new pipeline)
3) we introduced the sizing option to all tier1 cases, base on the instance value it was (including x86)

@pehala suggest pinning it cause he's scared the sizing would end up selecting other types or sized that would break the assumption about disk usage

one can argue if it's important or not, but the main goal of the sizing is preserved, i.e. other backend would be picking good enough instances, even if disk usage pattern would be a different.

so it's o.k. to use sizing and fixed in tandam (it's per it's design), the only risk it might drift when doing changes to either one, and if it's critical enough to keep the disk usage ratio ontop of the flexibility of change defaults of the sizing feature.

roydahan: 1. I don't think it's critical to keep the disk ratio as long as it's not too small.
2. I rather not mixing up the two approaches and stick only with the new approach, avoiding as much as possible from fixing a test to instance type.

pehala: > I don't think it's critical to keep the disk ratio as long as it's not too small.

I do, cluster has to work harder if it has more data and thus the testcase is better, besides we want to decrease TCO and thus we size customers clusters to have much more data -> so we should test more data as well.

> I rather not mixing up the two approaches and stick only with the new approach, avoiding as much as possible from fixing a test to instance type.

There are fully compatible and use advantages of both and this is still only short term measure, we want to parametrize CS commands to fill it based on available disk size in the long term.

TLDR: There is no harm done with this PR, Tier 1 will be still working on ALL clouds

roydahan: There should be (and I believe there are) scenarios that are size specific.

We also aim for instance type agnostic, so we can switch instance types as we want and/or based on availability.

pehala: > There should be (and I believe there are) scenarios that are size specific.
> 
> We also aim for instance type agnostic, so we can switch instance types as we want and/or based on availability.

Yes, but currently instance type agnostic is incompatible with size specific, which why this patch exists in the first place. I want to keep Tier 1 tests as size specific and add a feature which will make them type agnostic later. It is still agnostic on other clouds and it can run there, so again no harm done

fruch: @pehala name the jira for the followup item on calibration of stress commands, and have note about it in next those hardcoded value

fruch:
@cezarmoise you input was asked in couple of threads here.

configurations/2TB_high_disk_used_capacity.yaml
fruch: let's double check those we hard code, are one that was really optimized for specific disk usage target

test-cases/longevity/longevity-twcs-48h.yaml
roydahan: The actual change shouldn't be this, it should be setting the local_disk_gb to be greater than X and smaller than Y.

cezarmoise: Agree with Roy.

test-cases/longevity/longevity-multidc-parallel-topology-schema-changes-12h.yaml
cezarmoise: Yes, the only thing that matters is for utilization to be under 50%, se we don't run into issues when decommissioning or removing nodes via nemesis.

cezarmoise:
The only one that needs fixed instance type is the 90% test, because those stress commands are tuned to write just the right ammount of data for an AWS instance. For example they are tuned for `iXX.large` instances, which have 468 GB disk size. By coincidence, the test also works on AZURE, `Standard_L2s_v4` instance types has 480 GB disk size.

For the rest of the tests, the only limitation should be a lower limit to disk size, to ensure disk utilization stays under 50%, because at (usually) 2 nodes per rack, a nemesis that removes or decommissions a node would fail due to lack of disk space on that rack.

pehala:
> The only one that needs fixed instance type is the 90% test, because those stress commands are tuned to write just the right ammount of data for an AWS instance. For example they are tuned for `iXX.large` instances, which have 468 GB disk size. By coincidence, the test also works on AZURE, `Standard_L2s_v4` instance types has 480 GB disk size.
> 
> For the rest of the tests, the only limitation should be a lower limit to disk size, to ensure disk utilization stays under 50%, because at (usually) 2 nodes per rack, a nemesis that removes or decommissions a node would fail due to lack of disk space on that rack.

Ok I will change it

fruch:
> > The only one that needs fixed instance type is the 90% test, because those stress commands are tuned to write just the right ammount of data for an AWS instance. For example they are tuned for `iXX.large` instances, which have 468 GB disk size. By coincidence, the test also works on AZURE, `Standard_L2s_v4` instance types has 480 GB disk size.
> > For the rest of the tests, the only limitation should be a lower limit to disk size, to ensure disk utilization stays under 50%, because at (usually) 2 nodes per rack, a nemesis that removes or decommissions a node would fail due to lack of disk space on that rack.
> 
> Ok I will change it

I would recommend dropping the sizing for now, and leaving a note on those 90% disk usage cases
so it would cause any confusion when anyone might try it on other clouds
(at least until the test be full agnostic - stress command and anything else)

configurations/2TB_high_disk_used_capacity.yaml
pehala: @yarongilor please issue for yourself to separate the config

pehala:
v2: Only two changes present
   * elasticity test -> Remove sizing as the config has `nemesis_grow_shrink_instance_type: 'i8g.2xlarge'` which is not sizing compatible
   * 2TB test as that is really tuned to 40% on specific instance

configurations/2TB_high_disk_used_capacity.yaml
fruch: can't we just put disk limit >=1875, and not pin the specific instance, as in the rest of the tier 1 ?

pehala: We would have to put disk limit with both upper and lower bounds to make sure the instance is within the expected storage utilization and if we had that it would most likely fail to resolve unless different clouds have similar cpu to disk ratio, plus it could stop working randomly because we switched instances around or something new was introduced which has different cpu to disk ratio and doesnt fit within the bounds.

fruch: how it's different the other ones that were dropped from this PR ? 

also "it could stop working randomly" changes to sizing logic are not  happening at random.
pinning it to AWS type, wouldn't help when running on other clouds, which is the whole goal of the sizing, so we can switch easily

pehala: > how it's different the other ones that were dropped from this PR ?

TWCS was dropped because the storage utilization is not that high, we had to downscope it due to the TWCS itself (we dont insert quickly enough to keep the windows filled and we want short windows for testing the feature). MultiDC was dropped because the owner of the testcase requested it.

> also "it could stop working randomly" changes to sizing logic are not happening at random.
> pinning it to AWS type, wouldn't help when running on other clouds, which is the whole goal of the sizing, so we can switch easily

This does not remove sizing from the underlying config, so state with this PR is:
AWS -> Static instance, ensured data utilization percentage
Other clouds -> Work, but the data utilization percentage is not ensured.

It is best of both worlds, other clouds works but only AWS is fixed and has ensured data utilization percentage. For this testcase nothing harmful was done with respect to sizing.

Besides, they are testcases managed by our team so we should have free hand how we want to run them as part of Tier1.

cezarmoise (approved):

PR #15219, opened 2026-06-26
feat(docker): inject simulated rack via --rack entrypoint arg

aleksbykov:
Configure the simulated rack through the native `--rack/--dc` Docker entrypoint arguments (available since Scylla 2026.1) instead of reconfiguring the node after it has started.

Why we need this
----------------
Rack awareness in SCT is gated on the cluster reporting more than one rack, so none of it can be exercised on a single-rack cluster:

* `sdcm/cluster.py`: the CQL connection only switches to `RackAwareRoundRobinPolicy` when `rack_aware_loader` is set and `racks_count > 1`, otherwise it stays on `WhiteListRoundRobinPolicy`.
* `sdcm/tester.py`: `is_rack_aware_validation_enabled` returns False unless `racks_count > 1`, which silently disables rack-aware validation.
* `sdcm/nemesis/`: the rack-aware nemesis paths are guarded by the same `racks_count > 1` condition.
* A node only reports its rack in its log/dc_info when `racks_count > 1`.

Until now the Docker backend was excluded from rack simulation -- the snitch auto-resolution in `sct_config.py` carried `and cluster_backend != docker` -- so `simulated_racks` was silently ignored there. Every rack-aware path above could therefore only be covered on a cloud backend with real availability zones: no local reproduction of a rack-aware bug, no cheap CI coverage, and any rack regression is only caught by a slow and costly cloud job.

Why through the entrypoint rather than after boot: since Scylla 2026.1 the image entrypoint accepts `--dc/--rack` and writes `cassandra-rackdc.properties` before the first boot. Doing it after the node is up means rewriting that file, wiping the data directory and restarting Scylla behind an event filter -- racy, slow, and the filter hides genuine errors that happen in the same window. Passing the rack at container creation removes all of it.

When simulated racks take effect
--------------------------------
Racks need more than one rack AND more than one DB node to spread over. A single-node cluster stays in one rack whatever `simulated_racks` says, because the snitch auto-resolution already leaves `endpoint_snitch` alone below two nodes, so nothing reads the rack. Three call sites have to agree on that condition, and they now share one predicate, `sct_config.simulated_racks_enabled()`:

* the Docker version check below,
* the snitch auto-resolution in `SCTConfiguration.__init__` (section 15),
* the `--dc/--rack` injection in `NodeContainerMixin.node_container_run_args()`.

Keeping them in sync is what makes a single-node Docker cluster run on any Scylla version: it never gets a rack argument it cannot use. It also holds while a cluster grows -- `n_db_nodes` is the configured topology and is never mutated at runtime, so a nemesis adding a node to a one-node cluster does not suddenly build `--rack=RACK1` and kill that container.

Racks that do take effect require Scylla >= 2026.1, and `sct_config` raises when the image is older, before any container is created:

```
ValueError: simulated_racks=3 is not supported on Scylla 2025.1.0 (docker
backend): the --dc/--rack entrypoint arguments were added in 2026.1.0-dev.
Use a 2026.1+ image or set simulated_racks: 1.
```

Failing at configuration time rather than at container startup is deliberate: an older entrypoint forwards the arguments to the Scylla binary, which rejects them and exits, leaving a dead container and no explanation. master runs the latest Scylla and this change is backported to 2026.1 and 2026.2, so racks are available wherever it lands; a Docker test-case that must run on older images sets `simulated_racks: 1` itself. Branch versions such as `master:latest` are not comparable and are assumed new enough.

Note that `simulated_racks` keeps the global default of 3 on Docker as on every other backend -- there is no backend override -- so a multi-node Docker test that never mentions racks does get them on a 2026.1+ image. That is how the rack-aware code paths get cheap coverage.

Production changes
------------------
* `sct_config.py`: add `simulated_racks_enabled()`; use it for the snitch auto-resolution (replacing its inline duplicate) and for the new Docker version check, which raises `ValueError` from `SCTConfiguration.__init__`. `DOCKER_RACK_ARG_MIN_VERSION` pins the requirement at 2026.1.0-dev.
* `cluster_docker.py`: `NodeContainerMixin.node_container_run_args()` appends `--dc=datacenter1 --rack=RACK{n}` for db nodes when racks are in effect, so the entrypoint writes `cassandra-rackdc.properties` before the first boot. The snitch itself is applied later by `config_setup()`, from the `endpoint_snitch` `sct_config` resolves.
* Add a trailing `rack` parameter to `DockerNode`, `VectorStoreDockerNode` and `DockerMonitoringNode` `__init__` and to every `_create_node` in `cluster_docker.py`, so existing positional calls are unaffected. Loaders accept and ignore it.
* Implement round-robin rack distribution in `_create_nodes` (`rack=None` means "spread across racks", matching aws/gce/azure/oci).
* Add a REUSE_CLUSTER early-return guard to `ScyllaDockerCluster.node_setup()`.

Documentation
-------------
* `docs/docker-backend-overview.md`: new section on how the rack reaches the container, the version requirement and the error it produces, and the single-node case that is accepted on any version.
* Extend the `simulated_racks` option description with the node-count condition and the Docker version requirement; regenerate `docs/configuration_options.md`.

Test changes
------------
* `configure_scylla_node()` (integration conftest): accept optional rack/dc in `docker_scylla_args` and append `--rack/--dc` to the container `command_line`.
* `test_docker_simulated_racks.py`: boots two containers with `--rack/--dc` and asserts the racks in `system.local`; also covers the REUSE_CLUSTER path. Tracks the rolling `scylladb/scylla:2026.1` tag, since the patch level is irrelevant to what it asserts.
* `DummyScyllaDockerCluster`: `ScyllaDockerCluster` test double in `unit_tests/lib/` that bypasses only the constructor, so `node_setup` and `_create_nodes` under test are the real implementations.
* `unit_tests/unit/docker/`: new package grouping the Docker unit tests. `test_docker_rack_assignment.py` covers round-robin distribution and rack injection, including that a single-node cluster gets no rack argument; `test_docker_reuse_cluster.py` covers the REUSE_CLUSTER guard.
* `test_config.py`: covers `simulated_racks_enabled()` directly -- including the multi-DC boundary where `[1, 1]` enables racks and `[1, 0]` does not -- plus the snitch resolution and the version check accepting and rejecting per image.
* Pre-existing tests that pinned an old Scylla on the docker backend for unrelated reasons (image lookup, instance sizing, stress-command validation) now pin 2026.1.0 instead, so they exercise the version SCT targets rather than opting out of the requirement. `test_env_var_whitespace_is_stripped` keeps `simulated_racks: 1`, because the shape of its old `'5.4.0'` input is the point of the test.

Fixes: SCT-270

### Testing

- [ Artifact tests - Passed ](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/aleksbykov/job/artifact_tests/job/artifacts-docker-test/5/)

### Reminders

### DIFF
--- a/sdcm/cluster_docker.py
+++ b/sdcm/cluster_docker.py
@@ -28,6 +28,7 @@ from sdcm.provision.helpers.certificate import (
 )
 from sdcm.remote import LOCALRUNNER
 from sdcm.remote.docker_cmd_runner import DockerCmdRunner
+from sdcm.sct_config import simulated_racks_enabled
 from sdcm.sct_events.database import DatabaseLogEvent
 from sdcm.sct_events.filters import DbEventsFilter
 from sdcm.utils.docker_utils import get_docker_bridge_gateway, Container, ContainerManager, DockerException
@@ -75,10 +76,16 @@ class NodeContainerMixin:
             smp_match = re.search(r"--smp\s(\d+)", scylla_args)
             smp = int(smp_match.group(1)) if smp_match else 1
 
+        command_parts = []
+        if seed_ip:
+            command_parts.append(f'--seeds="{seed_ip}"')
+        if self.node_type == "db" and simulated_racks_enabled(self.parent_cluster.params):
+            command_parts.extend(["--dc=datacenter1", f"--rack=RACK{self.rack}"])
+
         return dict(
             name=self.name,
             image=self.node_container_image_tag,
-            command=f'--seeds="{seed_ip}"' if seed_ip else None,
+            command=" ".join(command_parts) if command_parts else None,
             volumes=volumes,
             network=self.parent_cluster.params.get("docker_network"),
             nano_cpus=smp * 10**9,
@@ -95,6 +102,7 @@ class DockerNode(cluster.BaseNode, NodeContainerMixin):
         ssh_login_info: Optional[dict] = None,
         node_index: int = 1,
         after_config=None,
+        rack: int = 0,
     ) -> None:
         super().__init__(
             name=f"{node_prefix}-{node_index}",
@@ -102,6 +110,7 @@ class DockerNode(cluster.BaseNode, NodeContainerMixin):
             ssh_login_info=ssh_login_info,
             base_logdir=base_logdir,
             node_prefix=node_prefix,
+            rack=rack,
             after_config=after_config,
         )
         self.node_index = node_index
@@ -308,6 +317,7 @@ class VectorStoreDockerNode(VectorStoreNodeMixin, DockerNode):
         ssh_login_info: Optional[dict] = None,
         node_index: int = 1,
         after_config=None,
+        rack: int = 0,
     ) -> None:
         super().__init__(
             parent_cluster=parent_cluster,
@@ -317,6 +327,7 @@ class VectorStoreDockerNode(VectorStoreNodeMixin, DockerNode):
             ssh_login_info=ssh_login_info,
             node_index=node_index,
             after_config=after_config,
+            rack=rack,
         )
 
     def node_container_run_args(self, seed_ip=None):
@@ -376,7 +387,7 @@ class DockerCluster(cluster.BaseCluster):
             os.path.dirname(__file__), "../docker/scylla-sct", self.params.get("scylla_linux_distro").split("-")[0]
         )
 
-    def _create_node(self, node_index, container=None, after_config=None):
+    def _create_node(self, node_index, container=None, after_config=None, rack=0):
         node = DockerNode(
             parent_cluster=self,
             container=container,
@@ -384,6 +395,7 @@ class DockerCluster(cluster.BaseCluster):
             base_logdir=self.logdir,
             node_prefix=self.node_prefix,
             node_index=node_index,
+            rack=rack,
             after_config=after_config,
         )
 
@@ -402,10 +414,11 @@ class DockerCluster(cluster.BaseCluster):
 
         return sorted(set(range(len(self.nodes) + count)) - set(node.node_index for node in self.nodes))
 
-    def _create_nodes(self, count, enable_auto_bootstrap=False):
+    def _create_nodes(self, count, enable_auto_bootstrap=False, after_config=None, rack=None):
         new_nodes = []
         for node_index in self._get_new_node_indexes(count):
-            node = self._create_node(node_index)
+            node_rack = node_index % self.racks_count if rack is None else rack
+            node = self._create_node(node_index, rack=node_rack, after_config=after_config)
             node.enable_auto_bootstrap = enable_auto_bootstrap
             self.nodes.append(node)
             new_nodes.append(node)
@@ -415,7 +428,8 @@ class DockerCluster(cluster.BaseCluster):
         containers = ContainerManager.get_containers_by_prefix(self.node_prefix)
         for node_index, container in sorted(((int(c.labels["NodeIndex"]), c) for c in containers), key=lambda x: x[0]):
             LOGGER.debug("Found container %s with name `%s' and index=%d", container, container.name, node_index)
-            node = self._create_node(node_index, container)
+            node_rack = node_index % self.racks_count
+            node = self._create_node(node_index, container, rack=node_rack)
             self.nodes.append(node)
         return self.nodes
 
@@ -431,7 +445,11 @@ class DockerCluster(cluster.BaseCluster):
         after_config=None,
     ):
         assert instance_type is None, "docker can't provision different instance types"
-        return self._get_nodes() if self.test_config.REUSE_CLUSTER else self._create_nodes(count, enable_auto_bootstrap)
+        return (
+            self._get_nodes()
+            if self.test_config.REUSE_CLUSTER
+            else self._create_nodes(count, enable_auto_bootstrap, after_config, rack=rack)
+        )
 
 
 class ScyllaDockerCluster(cluster.BaseScyllaCluster, DockerCluster):
@@ -460,6 +478,16 @@ class ScyllaDockerCluster(cluster.BaseScyllaCluster, DockerCluster):
         self.vector_store_cluster = None
 
     def node_setup(self, node, verbose=False, timeout=3600):
+        if self.test_config.REUSE_CLUSTER:
+            self._reuse_cluster_setup(node)
+            if (
+                any([self.params.get("server_encrypt"), self.params.get("client_encrypt")])
+                and not (node.ssl_conf_dir / TLSAssets.DB_CERT).exists()
+            ):
+                self._generate_db_node_certs(node)
+                install_client_certificate(node.remoter, node.ip_address, force=True)
+            return
+
         node.is_scylla_installed(raise_if_not_installed=True)
         self.check_aio_max_nr(node)
         if self.test_config.BACKTRACE_DECODING:
@@ -597,7 +625,7 @@ class CassandraDockerCluster(BaseCassandraCluster, DockerCluster):
             params=params,
         )
 
-    def _create_node(self, node_index, container=None, after_config=None):
+    def _create_node(self, node_index, container=None, after_config=None, rack=0):
         node = CassandraDockerNode(
             parent_cluster=self,
             container=container,
@@ -605,6 +633,7 @@ class CassandraDockerCluster(BaseCassandraCluster, DockerCluster):
             base_logdir=self.logdir,
             node_prefix=self.node_prefix,
             node_index=node_index,
+            rack=rack,
             after_config=after_config,
         )
 
@@ -697,13 +726,14 @@ class VectorStoreSetDocker(VectorStoreClusterMixin, DockerCluster):
                 self.log.error("Failed to reconfigure container %s: %s", node.name, e)
                 raise
 
-    def _create_node(self, node_index, container=None, after_config=None):
+    def _create_node(self, node_index, container=None, after_config=None, rack=0):
         node = VectorStoreDockerNode(
             parent_cluster=self,
             container=container,
             node_prefix=self.node_prefix,
             base_logdir=self.logdir,
             node_index=node_index,
+            rack=rack,
         )
 
         if container is None:
@@ -825,7 +855,9 @@ class LoaderSetDocker(cluster.BaseLoaderSet, DockerCluster):
             params=params,
         )
 
-    def _create_node(self, node_index, container=None, after_config=None):
+    def _create_node(self, node_index, container=None, after_config=None, rack=0):
+        # Loaders run directly on the host and have no rack semantics; the rack
+        # argument is accepted only to match the base _create_nodes() contract.
         node = DockerLoaderNode(
             parent_cluster=self,
             base_logdir=self.logdir,
@@ -880,6 +912,7 @@ class DockerMonitoringNode(cluster.BaseNode):
         node_index: int = 1,
         ssh_login_info: Optional[dict] = None,
         after_config=None,
+        rack: int = 0,
     ) -> None:
         super().__init__(
             name=f"{node_prefix}-{node_index}",
@@ -888,6 +921,7 @@ class DockerMonitoringNode(cluster.BaseNode):
             node_prefix=node_prefix,
             ssh_login_info=ssh_login_info,
             after_config=after_config,
+            rack=rack,
         )
         self.node_index = node_index
 
@@ -968,13 +1002,14 @@ class MonitorSetDocker(cluster.BaseMonitorSet, DockerCluster):
             params=params,
         )
 
-    def _create_node(self, node_index, container=None, after_config=None):
+    def _create_node(self, node_index, container=None, after_config=None, rack=0):
         node = DockerMonitoringNode(
             parent_cluster=self,
             base_logdir=self.logdir,
             node_prefix=self.node_prefix,
             node_index=node_index,
             ssh_login_info=None,
+            rack=rack,
         )
         node.init()
         return node
@@ -990,7 +1025,7 @@ class MonitorSetDocker(cluster.BaseMonitorSet, DockerCluster):
         after_config=None,
     ):
         assert instance_type is None, "docker can provision different instance types"
-        return self._create_nodes(count, enable_auto_bootstrap)
+        return self._create_nodes(count, enable_auto_bootstrap, rack=rack)
 
     @staticmethod
     def install_scylla_monitoring_prereqs(node):
--- a/sdcm/sct_config.py
+++ b/sdcm/sct_config.py
@@ -597,6 +597,27 @@ def _load_docker_images_defaults_cached():
     return None
 
 
+#: First Scylla release whose Docker image entrypoint accepts the `--dc`/`--rack` arguments.
+DOCKER_RACK_ARG_MIN_VERSION = "2026.1.0-dev"
+
+
+def simulated_racks_enabled(params) -> bool:
+    """True when simulated racks actually take effect.
+
+    Racks need both more than one rack and more than one DB node to spread over.  A single-node
+    cluster stays in one rack whatever `simulated_racks` says: the snitch auto-resolution leaves
+    `endpoint_snitch` alone, so nothing ever reads the rack.  It must therefore not pay any of the
+    cost of racks either -- in particular the Scylla >= 2026.1 requirement of the Docker
+    `--dc`/`--rack` entrypoint arguments, which older images reject outright.
+
+    Kept in one place because three call sites have to agree on it: the Docker version check and
+    the snitch auto-resolution in `SCTConfiguration.__init__`, and the `--dc`/`--rack` injection in
+    `NodeContainerMixin.node_container_run_args`.  `n_db_nodes` is the configured topology and is
+    never mutated at runtime, so growing a cluster does not change the answer mid-test.
+    """
+    return (params.get("simulated_racks") or 0) > 1 and sum(params.get("n_db_nodes") or []) > 1
+
+
 class SCTConfiguration(BaseModel):
     """
     Class the hold the SCT configuration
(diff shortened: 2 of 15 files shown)


sdcm/cluster_docker.py
```diff
@@ -171,6 +179,17 @@ def host_public_ip_address(self) -> str:
     def is_running(self):
         return ContainerManager.is_running(self, "node")
 
+    def is_docker_rack_arg_supported(self) -> bool:
```
pehala: Could this change for a cluster? If not, we can make it cached

unit_tests/unit/docker/test_docker_rack_assignment.py
pehala, line 1: Since you are adding two new docker tests, I would create docker package and put them together

unit_tests/lib/fake_docker_cluster.py
```diff
@@ -0,0 +1,117 @@
...
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2025 ScyllaDB
+
+"""Lightweight ScyllaDockerCluster test double for unit and integration tests.
```
pehala, line 14: Mocking should not be present in integration tests

unit_tests/lib/fake_docker_cluster.py
```diff
@@ -0,0 +1,117 @@
...
+``_create_nodes``, and ``_get_nodes``.
+"""
+
+import logging
+from types import SimpleNamespace
+
+from sdcm.cluster_docker import ScyllaDockerCluster
+
+
+class DummyScyllaDockerCluster(ScyllaDockerCluster):
```
pehala, line 28: There is already a docker cluster we use for testing, could that be resued/unified?

unit_tests/integration/test_docker_simulated_racks.py
```diff
@@ -0,0 +1,128 @@
...
+import pytest
+
+from unit_tests.integration.conftest import configure_scylla_node
+from unit_tests.lib.fake_docker_cluster import DummyDockerNode, DummyScyllaDockerCluster
+
+pytestmark = [pytest.mark.integration]
+
+
+# Scylla >= 2026.1 image that supports --rack/--dc entrypoint arguments.
+_RACK_CAPABLE_IMAGE = "docker.io/scylladb/scylla-nightly:latest"
```
pehala: Isnt this what we use always? If we use <2026.1 in tests that would be a bug IMO

unit_tests/integration/test_docker_simulated_racks.py
```diff
@@ -0,0 +1,128 @@
...
+        DummyDockerNode(scylla2, rack=1, node_index=1),
+    ]
+
+    yield nodes[0], nodes[1]
+
+    scylla2.kill()
+    scylla1.kill()
+
+
+@pytest.mark.xdist_group("docker-racks")
```
pehala: why is this needed?

unit_tests/integration/test_docker_simulated_racks.py
```diff
@@ -0,0 +1,128 @@
...
+    result = node1.run_cqlsh("SELECT rack FROM system.local")
+    assert "RACK0" in result.stdout
+
+    result = node2.run_cqlsh("SELECT rack FROM system.local")
+    assert "RACK1" in result.stdout
+
+    # Simulate a REUSE_CLUSTER run: create a new cluster object with
+    # reuse_cluster=True and call node_setup on the same nodes.
+    reuse_cluster = DummyScyllaDockerCluster(
+        params={"simulated_racks": 2, "docker_image": "scylladb/scylla:latest"},
```
pehala: Why are you forcing specific image, sholdnt the test machinery take care of that?

sdcm/cluster_docker.py
```diff
@@ -847,7 +904,7 @@ def add_nodes(
         after_config=None,
     ):
         assert instance_type is None, "docker can provision different instance types"
-        return self._create_nodes(count, enable_auto_bootstrap)
+        return self._create_nodes(count, enable_auto_bootstrap=enable_auto_bootstrap)
```
pehala: Dont understand why this is needed

unit_tests/integration/conftest.py
```diff
@@ -128,21 +128,30 @@ def configure_scylla_node(docker_scylla_args: dict, params, ssl_dir: Path | None
     ssl_mount = f" -v {ssl_dir}:{SCYLLA_SSL_CONF_DIR}:z" if ssl else ""
 
     env_vars = "-e VECTOR_SEARCH_TEST=true" if docker_scylla_args.get("scylla_docker_image") else ""
+    # native_entrypoint=True bypasses the SCT entry.sh wrapper (which enables
+    # PasswordAuthenticator) and lets the image's own entrypoint run, so the node
+    # boots without authentication. Used by rack tests against Scylla >= 2026.1.
+    if docker_scylla_args.get("native_entrypoint"):
```
pehala, line 134: this is really starting to be messy, we dont have any documentation for docker_scylla_args and any checking of the arguments is impossible. In the future we should refactor this

unit_tests/integration/conftest.py
```diff
@@ -128,21 +128,30 @@ def configure_scylla_node(docker_scylla_args: dict, params, ssl_dir: Path | None
     ssl_mount = f" -v {ssl_dir}:{SCYLLA_SSL_CONF_DIR}:z" if ssl else ""
 
     env_vars = "-e VECTOR_SEARCH_TEST=true" if docker_scylla_args.get("scylla_docker_image") else ""
+    # native_entrypoint=True bypasses the SCT entry.sh wrapper (which enables
+    # PasswordAuthenticator) and lets the image's own entrypoint run, so the node
+    # boots without authentication. Used by rack tests against Scylla >= 2026.1.
```
pehala, line 133: Why cannot you have passwordauthenticator in rack tests?

sdcm/cluster_docker.py
```diff
@@ -460,6 +493,19 @@ def __init__(
...
+            self._reuse_cluster_setup(node)
+            if (
+                any([self.params.get("server_encrypt"), self.params.get("client_encrypt")])
+                and not (node.ssl_conf_dir / TLSAssets.DB_CERT).exists()
+            ):
+                self._generate_db_node_certs(node)
+                install_client_certificate(node.remoter, node.ip_address, force=True)
+            # warm up raft cached_property per node to avoid rapid TLS session cycling
+            # during validate_raft_on_nodes (python-driver#614 workaround)
+            _ = node.raft
```
pehala: If this is bad file descriptor, this is already fixed

pehala:
> Replace the post-start rack configuration (writing cassandra-rackdc.properties, wiping data, restarting with DbEventsFilter) with the native --rack/--dc Docker entrypoint arguments (available since Scylla 2026.1).

Just a note, the entire cover letter is missing the most important part: **Why we need this**

fruch (dismissed):
check why the monitor stack on the provision docker test was failing

unit_tests/lib/fake_docker_cluster.py
fruch: I don't think we have any such rule anywhere...

unit_tests/lib/fake_docker_cluster.py
fruch: they are not sharing the same code, and don't operate in the same way at all...
I would try to mesh them up in such a PR

unit_tests/integration/test_docker_simulated_racks.py
fruch: this is new test, it doesn't have any test machinery fitting.

regardless it would be good if the version would be hardcoded, so it would be surprises on the followup release

fruch:
> check why the monitor stack on the provision docker test was failing

I'm guessing it was a bug @dimakr fixed already cause by new rsync versions, would recommend rebase and retest

sdcm/cluster_docker.py
fruch: LLMs keep thinking they are good boy-scouts

defaults/docker_config.yaml
```diff
@@ -4,3 +4,7 @@ use_preinstalled_scylla: true
 docker_image: ''
 scylla_linux_distro: 'centos'
 print_kernel_callstack: false
+# Simulated racks on Docker require the --rack/--dc entrypoint args (Scylla >= 2026.1),
+# so keep them off by default here instead of inheriting the global default of 3.
+# A test-case that wants racks opts in explicitly by setting simulated_racks itself.
+simulated_racks: 1
```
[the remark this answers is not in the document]
aleksbykov: comment updated

unit_tests/unit/test_config.py
```diff
@@ -831,3 +831,37 @@ def test_env_var_whitespace_is_stripped(monkeypatch, raw_value, expected):
...
+    """
+    monkeypatch.setenv("SCT_CLUSTER_BACKEND", "docker")
+    monkeypatch.setenv("SCT_SIMULATED_RACKS", "1")
+    monkeypatch.setenv("SCT_N_DB_NODES", "3")
+    monkeypatch.setenv("SCT_USE_MGMT", "false")
+
+    conf = sct_config.SCTConfiguration()
+    conf.verify_configuration()
+
+    assert conf.get("endpoint_snitch") is None
```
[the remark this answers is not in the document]
aleksbykov: fixed

sdcm/cluster_docker.py
aleksbykov: fixed

unit_tests/lib/fake_docker_cluster.py
aleksbykov: This isn't a mock — it's a thin test double that subclasses the real ScyllaDockerCluster so the genuine node_setup/_create_nodes/_get_nodes logic stays under test; only the heavy constructor chain is bypassed. It sits in shared unit_tests/lib/ next to fake_cluster.py/dummy_remote.py so unit and integration suites share it. I did remove RackAwareDummyDockerNode from this file — it was unused after the fixture rewrite.

unit_tests/unit/docker/test_docker_rack_assignment.py
aleksbykov: moved both new test files there. other test could be moved in another PR

unit_tests/lib/fake_docker_cluster.py
aleksbykov: They don't overlap: LocalScyllaClusterDummy is a minimal stand-in that never runs cluster logic, whereas DummyScyllaDockerCluster keeps the real methods so the REUSE_CLUSTER branch and the version guard are actually exercised. Merging them means reshaping one to the other's contract and touching every existing consumer — better as a dedicated refactor than widening this PR

unit_tests/integration/test_docker_simulated_racks.py
aleksbykov: set scylla:2026.1.9

sdcm/cluster_docker.py
aleksbykov: I set rack to the last parameter everywhere, so this change is not needed anymore.

unit_tests/integration/conftest.py
aleksbykov: Agree, but better do it in another PR

unit_tests/integration/conftest.py
aleksbykov: Not a choice about racks — an image constraint. These tests need Scylla ≥ 2026.1 for --rack, and 2026.1 removed the authenticator_user/authenticator_password scylla.yaml options that SCT's entry.sh writes to enable PasswordAuthenticator, so the node fails to start on the unknown config. native_entrypoint=True skips entry.sh so the node boots auth-free and run_cqlsh needs no -u/-p. Once entry.sh handles 2026.1 config these tests can use auth again.

unit_tests/integration/test_docker_simulated_racks.py
aleksbykov: removed

sdcm/cluster_docker.py
aleksbykov: removed

unit_tests/lib/fake_docker_cluster.py
```diff
@@ -0,0 +1,89 @@
...
+    - ``_create_nodes`` / ``_get_nodes`` (need ``racks_count``, ``nodes``,
+      ``node_prefix``)
+    """
+
+    # noinspection PyMissingConstructor
+    def __init__(
+        self,
+        params: dict,
+        logdir: str = "/tmp",
+        racks_count: int = 0,
```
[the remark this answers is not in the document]
aleksbykov: fixed

sdcm/cluster_docker.py
```diff
@@ -415,7 +442,8 @@ def _get_nodes(self):
         containers = ContainerManager.get_containers_by_prefix(self.node_prefix)
         for node_index, container in sorted(((int(c.labels["NodeIndex"]), c) for c in containers), key=lambda x: x[0]):
             LOGGER.debug("Found container %s with name `%s' and index=%d", container, container.name, node_index)
-            node = self._create_node(node_index, container)
+            node_rack = node_index % self.racks_count
+            node = self._create_node(node_index, container, rack=node_rack)
```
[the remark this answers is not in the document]
aleksbykov: Could be fixed in another pr

sdcm/cluster_docker.py
```diff
@@ -469,6 +511,13 @@ def node_setup(self, node, verbose=False, timeout=3600):
             self._generate_db_node_certs(node)
             install_client_certificate(node.remoter, node.ip_address, force=True)
 
+        simulated_racks = (self.params.get("simulated_racks") or 0) > 1
+        if simulated_racks and not node.is_docker_rack_arg_supported:
```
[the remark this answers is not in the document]
fruch: it has a point that we should strive those kind of check as part of sct_config.py,

fruch (dismissed):

fruch:
I would want to see a full cycle of the docker provision test working, to make sure no regression.
and it can be merged

defaults/docker_config.yaml
```diff
@@ -4,3 +4,14 @@ use_preinstalled_scylla: true
...
+# are injected as --dc/--rack entrypoint arguments at container creation, which only Scylla
+# >= 2026.1 understands. Inheriting 3 here would silently turn racks on for every Docker test
+# with more than one DB node and break each one running an older image, without any test
+# asking for racks.
+#
+# Racks on Docker are therefore opt-in: a test-case that wants them sets simulated_racks > 1
+# explicitly (and must run a 2026.1+ image, otherwise node_setup raises
+# ScyllaDockerRequirementError). The multi-rack path is covered by
+# unit_tests/integration/test_docker_simulated_racks.py.
+simulated_racks: 1
```
pehala: I think 2026.1+ shoud be the main usecase, with warning if you try simualted racks > 1 on older versions.

Only older supported version is 2025.1, which is not ran often.

sdcm/cluster_docker.py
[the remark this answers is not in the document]
pehala: I agree, it should be part of sct_config. Fail fast is always better

unit_tests/integration/test_docker_simulated_racks.py
```diff
@@ -0,0 +1,131 @@
...
+pytestmark = [
+    pytest.mark.integration,
+    pytest.mark.xdist_group("docker_heavy"),
+]
+
+
+# Scylla >= 2026.1 image that supports --rack/--dc entrypoint arguments.
+# Pinned to a released 2026.1.x tag (not :latest) so a future release cannot
+# silently change the image under this test.
+_RACK_CAPABLE_IMAGE = "docker.io/scylladb/scylla:2026.1.9"
```
pehala: Do we have generic 2026.1 tag, which we can use? I dont think we care about specific patch version used

aleksbykov: set generic image

defaults/docker_config.yaml
aleksbykov: remove it. sct_config checks fix this gap

sdcm/cluster_docker.py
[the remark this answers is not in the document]
aleksbykov: moved to sct_config

sdcm/sct_config.py
```diff
@@ -4173,6 +4085,37 @@ def _validate_perf_gradual_throttle_steps(self):
...
+            image_version = ComparableScyllaVersion(scylla_version)
+        except ValueError:
+            # Branched versions ('master:latest' and friends) are not comparable and are always new enough.
+            self.log.warning(
+                "Cannot determine whether Scylla docker image '%s' supports the --rack entrypoint argument "
+                "(scylla_version=%r); assuming it does.",
+                self.get("docker_image"),
+                scylla_version,
+            )
+            return
```
pehala: nit: I would degrade this to info or debug, it is not important message IMO

unit_tests/unit/test_instance_sizing_config.py
```diff
@@ -102,6 +102,8 @@ def test_docker_backend_skips_resolution(monkeypatch):
     monkeypatch.setenv("SCT_CLUSTER_BACKEND", "docker")
     monkeypatch.setenv("SCT_USE_MGMT", "false")
     monkeypatch.setenv("SCT_SCYLLA_VERSION", "2025.1.0")
+    # Racks on Docker need the --dc/--rack entrypoint args (Scylla 2026.1); unrelated here.
+    monkeypatch.setenv("SCT_SIMULATED_RACKS", "1")
```
pehala: nit: how about upgrading tests to use 2026.1+ instead? Can be followup as well.

pehala (dismissed):

sdcm/sct_config.py
aleksbykov: Done

unit_tests/unit/test_instance_sizing_config.py
aleksbykov: Done

aleksbykov:
Waiting results of jobs

aleksbykov:
Artifacts passed: https://jenkins.scylladb.com/job/scylla-staging/job/aleksbykov/job/artifact_tests/job/artifacts-docker-test/10/

dimakr (approved):
LGTM

but PR description is stale. It has pieces of the previous design
@aleksbykov

aleksbykov:
@dimakr , thanks. i fixed description

PR #15222, opened 2026-06-28
fix(gce_builder): make Jenkins builder agents single-use to avoid stuck launching

yaronkaikov:
## Summary
Dozens of SCT Jenkins `builders-*` agents were stuck in the `launching` state (RELENG-686).

The stuck nodes are **GCP Compute Engine** agents (`com.google.jenkins.plugins.computeengine.ComputeEngineComputer`), provisioned by the `ComputeEngineCloud` / `InstanceConfiguration` defined in `sdcm/utils/gce_builder.py` (`setNamePrefix("builders")`).

### Root cause
The Google Compute Engine plugin **does not terminate an instance when its SSH/agent launch fails or times out**. In `ComputeEngineComputerLauncher.launch()`, the outer SSH-connection path logs the exception but never calls `terminateNode()`. With reusable agents (`oneShot=false`), every instance whose SSH launch timed out (`launchTimeoutSeconds=300`) or failed was left:
- as a Jenkins node stuck in `connecting`/`launching` (`offline=true, connecting=true, offlineCause=null`), and
- as a running GCP VM.

Because the per-attempt `connectTime` resets on each reconnect, the idle-retention strategy never reaped them, so they accumulated into the "dozens" seen in the ticket. The config has been unchanged since 2023 — this is latent behavior that surfaced at scale, not a regression.

## Fix
Set `instanceConfiguration.setOneShot(true)` so each builder VM is **single-use**. One-shot agents are reaped aggressively and never reused into a stuck state, which prevents the zombie pile-up. Trade-off: a fresh VM per build (higher instance churn), acceptable for short CI builder jobs.

## Backport
**Not required (`backport/none`).** The Jenkins builder cloud is a single global configuration applied via the manually-run `sct.py configure-jenkins-builders` hydra command, executed from `master`. Release branches carry the identical config line but do not independently provision the builder cloud, so a backport would have no operational effect.

## Verification
Config-only change to the GCE cloud `InstanceConfiguration`. No automated SCT test covers Jenkins cloud provisioning; effect is validated operationally after re-running `configure-jenkins-builders` and confirming that failed/never-connecting builder VMs no longer linger in `launching`.

> Existing stuck nodes were cleared separately via a one-time Jenkins Script Console cleanup (terminating the orphaned `ComputeEngineInstance` nodes). This PR prevents recurrence.

Fixes: RELENG-686 (https://scylladb.atlassian.net/browse/RELENG-686)

### DIFF
--- a/sdcm/utils/gce_builder.py
+++ b/sdcm/utils/gce_builder.py
@@ -54,7 +54,7 @@ gcpCloud.setNoDelayProvisioning(true)
 
 InstanceConfiguration instanceConfiguration = new InstanceConfiguration()
 instanceConfiguration.setNumExecutorsStr("1")
-instanceConfiguration.setOneShot(false)
+instanceConfiguration.setOneShot(true)
 instanceConfiguration.setNamePrefix("builders")
 instanceConfiguration.setDescription("SCT Jenkins builders")
 instanceConfiguration.setRegion("https://www.googleapis.com/compute/v1/projects/{project_id}/regions/{region_name}")


fruch:
@yaronkaikov the builder-* are from GCE, this change wouldn't be doing anything about.

yaronkaikov:
Closing: the stuck `builders-*` nodes are GCP Compute Engine agents (ComputeEngineComputer), not AWS EC2 Fleet. This AWS change does not fix RELENG-686. Replaced by a GCE-targeted fix in gce_builder.py.

yaronkaikov:
Reopened and repurposed this PR for the correct fix.

Diagnostics on the Jenkins master showed the stuck `builders-*` nodes are **GCP Compute Engine** agents (`ComputeEngineComputer`), not AWS EC2 Fleet — so the original `aws_builder.py` change was the wrong cloud and has been dropped. The branch now carries the GCE fix in `sdcm/utils/gce_builder.py` (`setOneShot(true)`), and the title/description are updated with the verified root cause (the GCE plugin not terminating instances whose SSH launch fails).

yaronkaikov:
@fruch Any idea why CI here is failing ?

fruch:
> @fruch Any idea why CI here is failing ?

it's started failing after recent restart:
https://scylla-db.slack.com/archives/C08FTDLKD2Q/p1782631549820959?thread_ts=1782625793.062189&cid=C08FTDLKD2Q

yaronkaikov:
> > @fruch Any idea why CI here is failing ?
> 
> it's started failing after recent restart: https://scylla-db.slack.com/archives/C08FTDLKD2Q/p1782631549820959?thread_ts=1782625793.062189&cid=C08FTDLKD2Q

It's back :-)

yaronkaikov:
@fruch All passed, can we get it in?

fruch:
> @fruch All passed, can we get it in?

those kind of changes needs to be tested first, since it immediately affects multiple branches

fruch:
> > @fruch All passed, can we get it in?
> 
> those kind of changes needs to be tested first, since it immediately affects multiple branches

testing it now here:
https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-gce-image-test/36/

yaronkaikov:
> > > @fruch All passed, can we get it in?
> > 
> > 
> > those kind of changes needs to be tested first, since it immediately affects multiple branches
> 
> testing it now here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-gce-image-test/36/

Seems the branch was mis-configure - re-run now https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-gce-image-test/37/

yaronkaikov:
> > > > @fruch All passed, can we get it in?
> > > 
> > > 
> > > those kind of changes needs to be tested first, since it immediately affects multiple branches
> > 
> > 
> > testing it now here: https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-gce-image-test/36/
> 
> Seems the branch was mis-configure - re-run now https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/fruch/job/artifacts-gce-image-test/37/

@fruch job passed

PR #15223, opened 2026-06-28
feature(lint): add cyclic import detection 

fruch:
## Summary

Adds a pre-commit hook that detects cyclic imports using Python's `ast` module and Tarjan's SCC algorithm. Blocks new cycles from being introduced while grandfathering existing ones via an edge-level allowlist.

## Motivation

Cyclic imports in Python cause subtle runtime errors and import failures. This tooling prevents new ones from being introduced, with a path to fixing existing ones incrementally.

## Implementation

- **`scripts/detect_cyclic_imports.py`**: Standalone detector using Python `ast` module + Tarjan SCC
  - TYPE_CHECKING-guarded imports excluded (avoids false positives)
  - Function-level (deferred) imports excluded — they cannot cause an import-time cycle
  - Relative imports resolved via file path context
  - `from pkg import mod` resolves to the submodule `pkg.mod`, not just the package
  - Edges from a module up to its own ancestor packages are dropped (structural, not cyclic)
  - An edge counts only when both ends are in the *same* SCC, so a one-way bridge between two disjoint cycles is not reported
  - A file that fails to parse is an error, not a silent skip — the graph would be incomplete
  - Only project-internal edges analyzed (stdlib + third-party ignored)
  - Runtime: ~1s for a full scan
  - CLI: `--update-allowlist`, `--verbose`, `--include-deferred`
- **`data/known_cyclic_imports.yml`**: allowlist — **currently empty**. Once deferred (function-level) imports are excluded, the repo has no import-time cycles left to grandfather. Deferred-inclusive mode still reports 657 edges; fixing those is follow-up work and deliberately out of scope here.
- **`unit_tests/unit/test_detect_cyclic_imports.py`**: 21 pytest tests (extraction, TYPE_CHECKING exclusion, relative + package-style imports, ancestor edges, cross-SCC bridges, cycle detection, allowlist, edge cases)
- **`docs/cyclic-import-detection.md`**: usage, cycle-breaking strategies, and the parent/child scope limitation
- **`.pre-commit-config.yaml`**: New `cyclic-imports` hook (`language: system`)

## Design Decisions

- **Python `ast` over ast-grep**: Natively handles TYPE_CHECKING guards, relative imports, try/except — ast-grep cannot distinguish these
- **No new dependencies**: Tarjan SCC is ~40 lines of Python. The first iteration pulled in networkx, which was overkill. PyYAML is the only import beyond the stdlib and is already a hard dependency of SCT
- **Edge-level allowlist**: More stable across refactors than cycle-path-level; pairs of (source, target) modules
- **Whole-repo scan, not `sdcm/`-only**: cycles between top-level test modules and `sdcm/` matter too
- **Import-time cycles first**: deferred imports are a much larger, lower-priority backlog; surfacing the ones that actually break at import time is the point

## Usage

```bash
# Check for new cycles (used by pre-commit):
uv run python scripts/detect_cyclic_imports.py

# Regenerate allowlist after intentionally adding a cycle:
uv run python scripts/detect_cyclic_imports.py --update-allowlist

# Show all known cycles:
uv run python scripts/detect_cyclic_imports.py --verbose

# Diagnostics: include function-level imports (not used by pre-commit):
uv run python scripts/detect_cyclic_imports.py --verbose --include-deferred
```

Ref: SCT-531

### Testing
- [x] 21 unit tests pass
- [x] Pre-commit hook passes on current codebase
- [x] Synthetic cycle injection correctly detected and blocked (both `from pkg import mod` and `from . import mod` forms)

### DIFF
--- a/data/known_cyclic_imports.yml
+++ b/data/known_cyclic_imports.yml
@@ -1,0 +1,5 @@
+# Auto-generated allowlist of known cyclic import edges.
+# Each entry is a directed edge (source -> target) that forms part of a cycle.
+# To add: run `python scripts/detect_cyclic_imports.py --update-allowlist`
+# To fix: resolve the cycle and remove both directional edges.
+edges: []
--- a/data/known_cyclic_imports.yml
+++ b/data/known_cyclic_imports.yml
@@ -1,0 +1,7 @@
+# Auto-generated allowlist of known cyclic import edges.
+# Each entry is a directed edge (source -> target) that forms part of a cycle.
+# An empty list means the repository has no import-time cycles left to grandfather;
+# deferred (function-level) imports are out of scope, see --include-deferred.
+# To add: run `python scripts/detect_cyclic_imports.py --update-allowlist`
+# To fix: resolve the cycle and remove both directional edges.
+edges: []
--- a/scripts/detect_cyclic_imports.py
+++ b/scripts/detect_cyclic_imports.py
@@ -1,0 +1,382 @@
+#!/usr/bin/env python3
+"""Detect cyclic imports among project Python modules.
+
+Statically parses Python files with ast, builds a directed graph of
+project-internal imports, and reports import edges that belong to
+strongly connected components.
+"""
+
+from __future__ import annotations
+
+import argparse
+import ast
+import os
+import sys
+from collections.abc import Iterator
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import yaml
+
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+DEFAULT_ALLOWLIST = REPO_ROOT / "data" / "known_cyclic_imports.yml"
+ALLOWLIST_HEADER = """\
+# Auto-generated allowlist of known cyclic import edges.
+# Each entry is a directed edge (source -> target) that forms part of a cycle.
+# To add: run `python scripts/detect_cyclic_imports.py --update-allowlist`
+# To fix: resolve the cycle and remove both directional edges.
+"""
+
+EXCLUDED_DIRS = frozenset(
+    {
+        "__pycache__",
+        "node_modules",
+        "docker",
+        "docs",
+        "jenkins-pipelines",
+        "test-cases",
+        "configurations",
+        "defaults",
+        "data",
+        "data_dir",
+        "templates",
+        "argus_report_templates",
+        "rule-tests",
+        "rules",
+        "skills",
+        "tree-sitter-groovy",
+        "kafka-stack-docker-compose",
+        "jupyter",
+        "examples",
+        "scylla-qa-internal",
+    }
+)
+
+
+@dataclass
+class ImportEdge:
+    source: str
+    target: str
+    file: str
+    line: int
+
+
+@dataclass
+class ModuleImports:
+    targets: dict[str, list[int]] = field(default_factory=dict)
+
+    def add(self, target: str, line: int) -> None:
+        self.targets.setdefault(target, []).append(line)
+
+
+class ImportVisitor(ast.NodeVisitor):
+    """Collect import targets with line numbers.
+
+    Excludes:
+    - Imports inside TYPE_CHECKING guards (not runtime imports)
+    - Imports inside function/method bodies (deferred, not import-time cycles)
+      unless include_deferred=True
+    """
+
+    def __init__(self, source: str, package: str, *, include_deferred: bool = False) -> None:
+        self.source = source
+        self.package = package
+        self.imports = ModuleImports()
+        self._include_deferred = include_deferred
+
+    def visit_Import(self, node: ast.Import) -> None:  # noqa: N802
+        for alias in node.names:
+            self._add_target(alias.name, node.lineno)
+
+    def visit_ImportFrom(self, node: ast.ImportFrom) -> None:  # noqa: N802
+        target = self._resolve_import_from(node)
+        if target:
+            self._add_target(target, node.lineno)
+
+    def visit_If(self, node: ast.If) -> None:  # noqa: N802
+        if _is_type_checking_guard(node.test):
+            for item in node.orelse:
+                self.visit(item)
+            return
+        self.generic_visit(node)
+
+    def visit_FunctionDef(self, node: ast.FunctionDef) -> None:  # noqa: N802
+        if self._include_deferred:
+            self.generic_visit(node)
+
+    def visit_AsyncFunctionDef(self, node: ast.AsyncFunctionDef) -> None:  # noqa: N802
+        if self._include_deferred:
+            self.generic_visit(node)
+
+    def _add_target(self, target: str, line: int) -> None:
+        if target != self.source:
+            self.imports.add(target, line)
+
+    def _resolve_import_from(self, node: ast.ImportFrom) -> str | None:
+        module = node.module or ""
+        if node.level == 0:
+            return module
+
+        package_parts = self.package.split(".")
+        if node.level > len(package_parts):
+            return None
+
+        base_parts = package_parts[: len(package_parts) - node.level + 1]
+        if module:
+            base_parts.extend(module.split("."))
+        return ".".join(base_parts)
+
+
+def extract_imports(
+    project_root: Path,
+    *,
+    include_deferred: bool = False,
+) -> tuple[dict[str, set[str]], list[ImportEdge]]:
+    """Walk project Python files, return (adjacency dict, list of edges with locations)."""
+    graph: dict[str, set[str]] = {}
+    edges: list[ImportEdge] = []
+    project_root = project_root.resolve()
+
+    internal_modules = _discover_internal_modules(project_root)
+
+    for path in sorted(_iter_python_files(project_root)):
+        source = _module_name(path, project_root)
+        package = source if path.name == "__init__.py" else source.rsplit(".", 1)[0] if "." in source else ""
+        graph.setdefault(source, set())
+
+        try:
+            tree = ast.parse(path.read_text(encoding="utf-8"), filename=str(path))
+        except SyntaxError as exc:
+            print(f"WARNING: failed to parse {path}: {exc}", file=sys.stderr)
+            continue
+
+        visitor = ImportVisitor(source=source, package=package, include_deferred=include_deferred)
+        visitor.visit(tree)
+
+        rel_path = str(path.relative_to(project_root))
+        for target, lines in visitor.imports.targets.items():
+            if target in internal_modules and target != source:
+                graph[source].add(target)
+                for line in lines:
+                    edges.append(ImportEdge(source=source, target=target, file=rel_path, line=line))
+
+    for targets in list(graph.values()):
+        for target in targets:
+            graph.setdefault(target, set())
+
+    return graph, edges
+
+
+def find_cycles(graph: dict[str, set[str]]) -> list[tuple[str, str]]:
+    """Return list of (source, target) edges that participate in any cycle."""
+    components = _strongly_connected_components(graph)
+    cyclic_nodes = {node for component in components if len(component) > 1 for node in component}
+    cyclic_edges = [
+        (source, target)
+        for source, targets in graph.items()
+        if source in cyclic_nodes
+        for target in targets
+        if target in cyclic_nodes
+    ]
+    return sorted(cyclic_edges)
+
+
+def load_allowlist(path: Path) -> set[tuple[str, str]]:
+    """Load known_cyclic_imports.yml, return set of (source, target) pairs."""
+    if not path.exists():
+        return set()
+
+    data = yaml.safe_load(path.read_text(encoding="utf-8")) or {}
+    allowlist_edges = data.get("edges") or []
+    return {
+        (edge["source"], edge["target"])
+        for edge in allowlist_edges
+        if isinstance(edge, dict) and edge.get("source") and edge.get("target")
+    }
+
+
+def main() -> int:
+    """CLI entry point. Returns exit code."""
+    parser = argparse.ArgumentParser(description="Detect cyclic imports in Python project.")
+    parser.add_argument("--update-allowlist", action="store_true", help="write current cyclic edges to allowlist")
+    parser.add_argument("--verbose", action="store_true", help="print all cyclic edges with file locations")
+    parser.add_argument(
+        "--include-deferred",
+        action="store_true",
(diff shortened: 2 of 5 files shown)


scripts/detect_cyclic_imports.py
```diff
@@ -0,0 +1,382 @@
...
+import sys
+from collections.abc import Iterator
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import yaml
+
+
+REPO_ROOT = Path(__file__).resolve().parent.parent
+DEFAULT_ALLOWLIST = REPO_ROOT / "data" / "known_cyclic_imports.yml"
```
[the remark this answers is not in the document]
fruch: this is what is supposed todo test all

fruch:
@claude review

pehala:
How long does it take?

fruch:
> How long does it take?


~1.5s 

much better then picking up pylint for it.


also we can flush out the inline imports that was done to avoid cycles:
```python
    parser.add_argument(
        "--include-deferred",
        action="store_true",
        help="also count imports inside functions (for diagnostics, not used by pre-commit)",
    )
```

data/known_cyclic_imports.yml
```diff
@@ -0,0 +1,5 @@
+# Auto-generated allowlist of known cyclic import edges.
+# Each entry is a directed edge (source -> target) that forms part of a cycle.
+# To add: run `python scripts/detect_cyclic_imports.py --update-allowlist`
+# To fix: resolve the cycle and remove both directional edges.
+edges: []
```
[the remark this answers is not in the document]
fruch: this is by design

cezarmoise, line 7: Commit mentions `638 grandfathered cycle edges`.

scripts/detect_cyclic_imports.py
```diff
@@ -0,0 +1,382 @@
...
+
+import argparse
+import ast
+import os
+import sys
+from collections.abc import Iterator
+from dataclasses import dataclass, field
+from pathlib import Path
+
+import yaml
```
cezarmoise, line 19: Commit says `stdlib-only`, but it imports `yaml`

data/known_cyclic_imports.yml
fruch: Initially it counted the inline imports as well, decided to exclude them.

The command can fetch them on demand but handling those would be a followup (since those are so many of those)

The idea is to flush out early to one that really gonna cause issues and gonna braek the code (the inline are 2nd priority)

scripts/detect_cyclic_imports.py
fruch: It meabing no extra dependcy is needed, in the first iteration it pulled in networkx to do the computation, it was overkill.

scripts/detect_cyclic_imports.py
```diff
@@ -0,0 +1,382 @@
...
+        self._include_deferred = include_deferred
+
+    def visit_Import(self, node: ast.Import) -> None:  # noqa: N802
+        for alias in node.names:
+            self._add_target(alias.name, node.lineno)
+
+    def visit_ImportFrom(self, node: ast.ImportFrom) -> None:  # noqa: N802
+        target = self._resolve_import_from(node)
+        if target:
+            self._add_target(target, node.lineno)
```
cezarmoise:
```suggestion
            self._add_target(target, node.lineno)
            for alias in node.names:
                if alias.name != "*":
                    self._add_target(f"{target}.{alias.name}", node.lineno)
```

This needs to be enhanced to handle package-style imports.
For example, `from sdcm.utils.alternator import api` is recorded only as an edge to `sdcm.utils.alternator`, not to `sdcm.utils.alternator.api`.

Example reproducer.
```diff
diff --git a/sdcm/cyclic_import_repro_pkg/a.py b/sdcm/cyclic_import_repro_pkg/a.py
new file mode 100644
index 000000000..c65973eb0
--- /dev/null
+++ b/sdcm/cyclic_import_repro_pkg/a.py
@@ -0,0 +1,5 @@
+"""Module A: imports module B via package-style import."""
+
+from sdcm.cyclic_import_repro_pkg import b
+
+VALUE_A = b.VALUE_B + 1
diff --git a/sdcm/cyclic_import_repro_pkg/b.py b/sdcm/cyclic_import_repro_pkg/b.py
new file mode 100644
index 000000000..946e89ba3
--- /dev/null
+++ b/sdcm/cyclic_import_repro_pkg/b.py
@@ -0,0 +1,5 @@
+"""Module B: imports module A via package-style import."""
+
+from sdcm.cyclic_import_repro_pkg import a
+
+VALUE_B = a.VALUE_A + 1
```


```bash
➜ python -c "import sdcm.cyclic_import_repro_pkg.a"
  File "<string>", line 1, in <module>
    import sdcm.cyclic_import_repro_pkg.a
  File "/home/cezar/github/scylla-cluster-tests/sdcm/cyclic_import_repro_pkg/a.py", line 3, in <module>
    from sdcm.cyclic_import_repro_pkg import b
  File "/home/cezar/github/scylla-cluster-tests/sdcm/cyclic_import_repro_pkg/b.py", line 5, in <module>
    VALUE_B = a.VALUE_A + 1
              ^^^^^^^^^
AttributeError: partially initialized module 'sdcm.cyclic_import_repro_pkg.a' from '/home/cezar/github/scylla-cluster-tests/sdcm/cyclic_import_repro_pkg/a.py' has no attribute 'VALUE_A' (most likely due to a circular import)
```

with fix it detects it in the script too
```bash
➜ python scripts/detect_cyclic_imports.py
ERROR: New cyclic import edge(s) detected:

  sdcm.cyclic_import_repro_pkg.a -> sdcm.cyclic_import_repro_pkg.b  (sdcm/cyclic_import_repro_pkg/a.py:3)
  sdcm.cyclic_import_repro_pkg.b -> sdcm.cyclic_import_repro_pkg.a  (sdcm/cyclic_import_repro_pkg/b.py:3)

Cycle path: sdcm.cyclic_import_repro_pkg.a -> sdcm.cyclic_import_repro_pkg.b -> sdcm.cyclic_import_repro_pkg.a

To add to allowlist: uv run python scripts/detect_cyclic_imports.py --update-allowlist
To fix: break the cycle by making one import lazy or moving it under TYPE_CHECKING.
```

cezarmoise: And some alternator one
```
➜ python scripts/detect_cyclic_imports.py
ERROR: New cyclic import edge(s) detected:

  sdcm.utils.alternator -> sdcm.utils.alternator.api  (sdcm/utils/alternator/__init__.py:1)
  sdcm.utils.alternator -> sdcm.utils.alternator.schemas  (sdcm/utils/alternator/__init__.py:4)
  sdcm.utils.alternator.api -> sdcm.utils.alternator  (sdcm/utils/alternator/api.py:15)
  sdcm.utils.alternator.api -> sdcm.utils.alternator.schemas  (sdcm/utils/alternator/api.py:15)
  sdcm.utils.alternator.schemas -> sdcm.utils.alternator  (sdcm/utils/alternator/schemas.py:1)
  sdcm.utils.alternator.schemas -> sdcm.utils.alternator  (sdcm/utils/alternator/schemas.py:2)
  ```

Comment (fruch) [reply scripts/detect_cyclic_imports.py]: Fixed in 80c7b22.

`visit_ImportFrom()` now records the submodule target as well as the package:

```python
self._add_target(target, node.lineno)
for alias in node.names:
    if alias.name != "*":
        self._add_target(f"{target}.{alias.name}", node.lineno)
```

Names that are not submodules (classes, constants) resolve to module names that do not exist and are dropped by the internal-module filter in `extract_imports()`, so `from sdcm.bar import SomeClass` still yields only `sdcm.foo -> sdcm.bar`.

Your reproducer is now detected, and so is the relative form (`from . import b`) — both covered by new tests `test_package_style_cycle_detected` and `test_relative_package_style_import_creates_submodule_edge`.

fruch: Those alternator ones were a false positive that the naive fix introduced, so I did not allowlist them — I made the resolution more precise instead.

The noise came from the `submodule -> own ancestor package` edges (`sdcm.utils.alternator.api -> sdcm.utils.alternator` at api.py:15). Importing `sdcm.utils.alternator.api` always initialises `sdcm.utils.alternator` first, so that edge is structural, never an extra constraint. Counting it turns *every* re-exporting `__init__.py` into a cycle, which would have made the hook useless.

`_add_target()` now drops edges to the source module`s own ancestors:

```python
if target == self.source or self.source.startswith(f"{target}."):
    return
```

Result: alternator is clean again, while your `cyclic_import_repro_pkg` reproducer still fails the hook — verified both ways locally. Covered by `test_ancestor_package_edge_excluded` and `test_import_of_own_ancestor_excluded`.

data/known_cyclic_imports.yml
fruch: Good catch — the commit message was stale. Reworded in 80c7b22: it now says the allowlist is currently empty and why (no import-time cycles left once deferred/function-level imports are excluded), instead of the old 638 figure from the first iteration.

`data/known_cyclic_imports.yml` also carries that note in its auto-generated header now, so the empty list does not read as a bug.

For reference, `--verbose --include-deferred` still reports 732 edges — that is the follow-up backlog.

scripts/detect_cyclic_imports.py
fruch: Reworded in 80c7b22 — the bullet now reads "no new dependencies beyond PyYAML, already required by SCT" instead of "stdlib-only". The point was that the first iteration pulled in networkx and this one does not; PyYAML is already a hard dependency of the repo.

scripts/detect_cyclic_imports.py
```diff
@@ -0,0 +1,268 @@
...
+        self.targets: set[str] = set()
+
+    def visit_Import(self, node: ast.Import) -> None:  # noqa: N802
+        for alias in node.names:
+            self._add_target(alias.name)
+
+    def visit_ImportFrom(self, node: ast.ImportFrom) -> None:  # noqa: N802
+        target = self._resolve_import_from(node)
+        if target:
+            self._add_target(target)
```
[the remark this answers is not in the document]
fruch: Fixed in 80c7b22, together with the package-style variant @cezarmoise raised below. `from . import foo` in `sdcm.pkg` now resolves per-alias to `sdcm.pkg.foo`, not just `sdcm.pkg`.

New test: `test_relative_package_style_import_creates_submodule_edge`.

scripts/detect_cyclic_imports.py
```diff
@@ -0,0 +1,268 @@
...
+        if _is_type_checking_guard(node.test):
+            for item in node.orelse:
+                self.visit(item)
+            return
+        self.generic_visit(node)
+
+    def _add_target(self, target: str) -> None:
+        if target.startswith("sdcm.") and target != self.source:
+            self.targets.add(target)
+
```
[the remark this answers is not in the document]
fruch: Already handled — `visit_FunctionDef` / `visit_AsyncFunctionDef` stop the traversal unless `--include-deferred` is passed, so function-local imports are not graph edges. Covered by `test_function_level_imports_excluded` and `test_method_level_imports_excluded`.

scripts/detect_cyclic_imports.py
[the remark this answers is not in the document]
fruch: Reply-only — scanning the whole repo is intentional (cycles between top-level test modules and `sdcm/` matter too). The docs were the thing that was wrong: they claimed `sdcm.*`-only. Corrected in 80c7b22 to say project-internal modules.

data/known_cyclic_imports.yml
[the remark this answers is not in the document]
fruch: Reply-only, as noted above — the empty list is correct, not a rollout gap. Once deferred (function-level) imports are excluded there are no import-time cycles left to grandfather. The file header now says so explicitly, and the commit message no longer claims 638 edges.

scripts/detect_cyclic_imports.py
fruch: Follow-up on the same thread: I audited whether dropping the ancestor edges can hide a *real* cycle. Restoring only the 87 `from <own ancestor> import <attribute>` edges (i.e. the ones that are not submodule imports) and re-running Tarjan gives **0 cycles** across the repo — so nothing is masked today.

Longer term this is a genuine scope boundary rather than a bug, and I documented it in `docs/cyclic-import-detection.md` under "Scope limitation: parent/child ordering": whether

```python
# sdcm/pkg/__init__.py
from sdcm.pkg import sub
helper = ...
# sdcm/pkg/sub.py
from sdcm.pkg import helper
```

blows up depends on whether `helper` is defined before or after the `import sub` line — an ordering question, not a graph question. No SCC-based detector can answer it, and flagging the shape would flag every package in the repo.

fruch:
Pushed `fb836a5` — review comments addressed, plus a self-review pass that turned up a few more.

**From the review threads**
- `from pkg import mod` now records the submodule edge (@cezarmoise's reproducer), and the relative `from . import mod` form with it.
- The alternator fallout that fix produced was a false positive, not something to allowlist: `submodule -> own ancestor package` is structural. Dropped those edges instead; audited that restoring the 87 attribute-from-ancestor edges yields 0 cycles, so nothing real is masked.
- Commit message and PR description corrected (no more "638 grandfathered edges", no more "stdlib-only").

**Found while re-reviewing the result**
- `find_cycles()` tested edges against the *union* of all SCCs, so a one-way bridge between two disjoint cycles was reported as cyclic — 49 of 706 edges in deferred mode were bogus. Now scoped per component (706 -> 657). This mattered: the first real cycle anyone landed would have pointed them at innocent imports, and `--update-allowlist` would have grandfathered those permanently.
- A file that failed to parse was skipped silently and the hook still exited 0 — a cycle through it was invisible. Parse failures are now a hard error.
- `--update-allowlist --include-deferred` wrote the deferred-inclusive edge set into the allowlist the non-deferred check reads, silently disabling the hook repo-wide. The flags are now mutually exclusive.
- Directory exclusions matched basenames at any depth, so `unit_tests/unit/docker/` was silently unscanned. Anchored to the repo root (844 -> 847 files).
- The "Picking Low-Hanging Fruit" recipe in the docs could never match (it did not strip the `(file:line)` suffix), so it always printed `Found 0`. Fixed — it now reports 11 mutual cycles.
- Dropped the dead `files: \.py$` from the hook config (no effect under `always_run`).

21 unit tests, `pre-commit run --all-files` clean.

cezarmoise (approved):
LGTM

PR #15253, opened 2026-06-29
fix(nemesis): use dedicated keyspace for refresh/load-and-stream to avoid stress conflicts

fruch:
## Fixes SCT-555

Refresh and load-and-stream nemesis operations use `keyspace1.standard1` which is the same table used by the main stress command (`cassandra-stress`). This causes conflicts — the nemesis operations interfere with ongoing stress writes, leading to failures.

## What

Move refresh/load-and-stream nemesis to use a dedicated `keyspace_refresh.standard1` table instead of `keyspace1.standard1`, so these nemesis operations no longer conflict with the main stress workload.

## Changes

- **`sdcm/nemesis/__init__.py`** — switch `disrupt_nodetool_refresh` and `disrupt_load_and_stream` from `keyspace1` to `keyspace_refresh`; pass explicit `keyspace` to `_prepare_test_table` and `SstableLoadUtils` calls
- **`sdcm/utils/sstable/load_utils.py`** — update default `keyspace_name` parameter from `keyspace1` to `keyspace_refresh` across `calculate_columns_count_in_table`, `upload_sstables`, `load_and_stream`, and `run_refresh`
- **`sdcm/remote/local_cmd_runner.py`** — add missing `user` parameter to `run` method signature
- **`unit_tests/integration/test_nemesis_refresh.py`** — new integration test (`test_refresh_monkey_flow`) that validates the refresh flow end-to-end using Docker
- **`unit_tests/lib/fake_cluster.py`** — test fixture update
- **`unit_tests/test_data/load_and_stream.log`** — updated test data to match new keyspace name

## Testing

```bash
uv run python -m pytest unit_tests/integration/test_nemesis_refresh.py -x -q --no-header --tb=short -n0
# 1 passed in 35.30s
```

### Testing

- [x] 🟢 [oss/nemesis/longevity-5gb-1h-RefreshMonkey-aws-test #3](https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/oss/job/nemesis/job/longevity-5gb-1h-RefreshMonkey-aws-test/3/)

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -1674,10 +1674,12 @@ class NemesisRunner:
         self._major_compaction()
 
     def disrupt_load_and_stream(self):
-        # Checking the columns number of keyspace1.standard1
-        self.log.debug("Prepare keyspace1.standard1 if it does not exist")
-        self._prepare_test_table(ks="keyspace1")
-        column_num = SstableLoadUtils.calculate_columns_count_in_table(self.target_node)
+        # Checking the columns number of keyspace_refresh.standard1
+        self.log.debug("Prepare keyspace_refresh.standard1 if it does not exist")
+        self._prepare_test_table(ks="keyspace_refresh")
+        column_num = SstableLoadUtils.calculate_columns_count_in_table(
+            self.target_node, keyspace_name="keyspace_refresh"
+        )
 
         # Run load-and-stream test on regular standard1 table of cassandra-stress.
         if column_num < 5:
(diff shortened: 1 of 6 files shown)


fruch:
@claude review

sdcm/nemesis/__init__.py
```diff
@@ -1674,37 +1674,43 @@ def disrupt_major_compaction(self):
...
                 self.actions_log.info(f"Uploading sstables to {load_on_node.name}")
-                SstableLoadUtils.upload_sstables(load_on_node, test_data=sstables_info, table_name="standard1")
+                SstableLoadUtils.upload_sstables(
+                    load_on_node, test_data=sstables_info, keyspace_name="keyspace_refresh", table_name="standard1"
+                )
                 # NOTE: on K8S logs may appear with a delay, so add a bigger timeout for it.
                 #       See https://github.com/scylladb/scylla-cluster-tests/issues/6314
                 kwargs = {"start_timeout": 1800, "end_timeout": 1800} if self._is_it_on_kubernetes() else {}
                 with self.action_log_scope(f"Loading and streaming sstables on {load_on_node.name} node"):
                     SstableLoadUtils.run_load_and_stream(load_on_node, **kwargs)
```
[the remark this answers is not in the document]
fruch: Fixed. Both `disrupt_load_and_stream` and `disrupt_nodetool_refresh` now raise `UnsupportedNemesis` when `cfstats` returns non-zero, and the subsequent logic is un-indented out of the conditional.

sdcm/utils/sstable/load_utils.py
```diff
@@ -245,7 +250,9 @@ def create_table_for_load(cls, node, schema_file_and_path: str, session):
         session.execute(schema.replace("\n", ""))
 
     @classmethod
-    def validate_data_count_after_upload(cls, node, keyspace_name: str = "keyspace1", table_name: str = "standard1"):
+    def validate_data_count_after_upload(
+        cls, node, keyspace_name: str = "keyspace_refresh", table_name: str = "standard2"
+    ):
```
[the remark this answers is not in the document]
fruch: Fixed. Reverted `validate_data_count_after_upload` default back to `keyspace1` — this method is only called by the K8S sstable upload flow which operates on `keyspace1.standard2`, not the refresh flow.

unit_tests/integration/test_nemesis_refresh.py
```diff
@@ -0,0 +1,73 @@
...
+            SstableLoadUtils.upload_sstables(
+                node,
+                test_data=test_data[0],
+                table_name="standard1",
+                is_cloud_cluster=False,
+            )
+            SstableLoadUtils.run_refresh(node, test_data=test_data[0])
+        # Verify that the special key is loaded by SELECT query
+        result = docker_scylla.run_cqlsh(query_verify)
+        assert "(1 rows)" in result.stdout, f"The key {key} is not loaded by `nodetool refresh`"
```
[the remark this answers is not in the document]
fruch: Fixed. Replaced the `if result ... == 0:` conditional with `assert result is not None and result.exit_status == 0`, so a broken setup now fails immediately instead of silently passing.

unit_tests/integration/test_nemesis_refresh.py
```diff
@@ -0,0 +1,73 @@
...
+    # Check one special key before refresh, we will verify refresh by query in the end
+    # Note: we can't DELETE the key before refresh, otherwise the old sstable won't be loaded
+    #       TRUNCATE can be used the clean the table, but we can't do it for keyspace_refresh.standard1
+    query_verify = f"SELECT * FROM keyspace_refresh.standard1 WHERE key={key}"
+    result = docker_scylla.run_cqlsh(query_verify)
+    if "(0 rows)" in result.stdout:
+        logging.debug("Key %s does not exist before refresh", key)
+    else:
+        logging.debug("Key %s already exists before refresh", key)
+
```
pehala: What is this for? you never assert it anywhere

unit_tests/integration/test_nemesis_refresh.py
```diff
@@ -0,0 +1,73 @@
...
+pytestmark = [
+    pytest.mark.usefixtures("events"),
+    pytest.mark.integration,
+    pytest.mark.xdist_group("docker_heavy"),
+]
+
+
+@pytest.mark.integration
+def test_refresh_monkey_flow(docker_scylla, params, events, request):
```
pehala: I dont really understand what this does. It doesn't test the monkey so it is just testing copy of the code, which I think brings no value.

unit_tests/integration/test_nemesis_refresh.py
```diff
@@ -0,0 +1,73 @@
...
+    #       TRUNCATE can be used the clean the table, but we can't do it for keyspace_refresh.standard1
+    query_verify = f"SELECT * FROM keyspace_refresh.standard1 WHERE key={key}"
+    result = docker_scylla.run_cqlsh(query_verify)
+    if "(0 rows)" in result.stdout:
+        logging.debug("Key %s does not exist before refresh", key)
+    else:
+        logging.debug("Key %s already exists before refresh", key)
+
+    # Executing rolling refresh one by one
+    for node in [docker_scylla]:
```
pehala: Why iterate over single value, wont node simply be docker_scylla?

pehala (commented):
Change to the nemesis LGTM, but I dont understand the point of the test, what it is testing and why

unit_tests/integration/test_nemesis_refresh.py
fruch: It help be test and validate the code flow, of the pulling the data from s3 and doing the refresh with actual scylla node, with messing about everything around the nemesis.

i.e. this is integration test I needed.

unit_tests/integration/test_nemesis_refresh.py
fruch: that what search and replace did

unit_tests/integration/test_nemesis_refresh.py
fruch: informational as it in the nemesis

unit_tests/integration/test_nemesis_refresh.py
fruch: if you more of those to test nemesis code as integration tests, you can do it as a followup.

here it was to validate the changes, so I would need to debug it on AWS.

I think it's useful one for the sstable utils part.

pehala: I think it served its purpose as a one time test for manual testing, but I am not sure if we need it like this for master. 

It doesnt verify the nemesis will work in the future as it is totally disconnected. I think it could be useful if we simply verify the refresh from data S3, without any links to the nemesis itself (i.e. remove all that is in the nemesis and doesnt directly relate to this)

fruch: you are saying it serve it's one off propose,  
and then if we remove the nemesis from it's name, now it's could be useful ? 


I'm gonna rename he file and test name, and that's it.

pehala: > and then if we remove the nemesis from it's name, now it's could be useful ?

No, I am saying trim it down to remove all nemesis references and uneeded stuff like, logging messages for no reason etc

pehala (approved):

timtimb0t:
The PR has backport labels, but it looks like it wasn’t actually backported to 2026.1.

PR #15274, opened 2026-06-30
fix(gemini): custom command inherits --duration from configured gemini_cmd

CodeLieutenant:
When a custom gemini command omits --duration, run_gemini now extracts the duration from the configured gemini_cmd param (the previous command) and appends it, preventing silent fallback to the Gemini CLI default of 30s.

Also fixes the explicit duration= override branch to inject --duration into the command string, not just set the timeout.

Closes SCT-556

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -171,6 +171,7 @@ from sdcm.stress_thread import (
     apply_gemini_stress_duration,
     extract_gemini_seed,
     get_timeout_from_stress_cmd,
+    stress_cmd_get_duration_pattern,
 )
 from sdcm.gemini_thread import GeminiStressThread
 from sdcm.utils.log_time_consistency import DbLogTimeConsistencyAnalyzer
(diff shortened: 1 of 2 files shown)


sdcm/tester.py
```diff
@@ -3129,10 +3119,18 @@ def run_cdclog_reader_thread(
...
             timeout = self.get_duration(self._stress_duration)
             cmd = apply_gemini_stress_duration(cmd, self._stress_duration)
         else:
+            if "--duration" not in cmd:
+                # When the custom command has no --duration, extract it from the configured
+                # gemini_cmd (the "previous command") so the test doesn't silently run for
+                # the Gemini CLI default of 30s.
+                default_cmd = self.params.get("gemini_cmd") or ""
+                if duration_match := stress_cmd_get_duration_pattern.search(default_cmd):
+                    cmd += f" --duration {duration_match.group(1)}"
```
soyacz, line 3133: when users override, they override `gemini_cmd` param - so this is no-op.
I think we should raise error during sct_config verification - gemini_cmd without duration should fail the test fast

PR #15275, opened 2026-06-30
fix(monitor): show non-DB node metrics on xcloud dashboards

dimakr:
On xcloud backend DB host metrics come from the Scylla Cloud promproxy labeled as `cluster=#<id>`. SCT non-DB nodes (loader, monitor, sct-runner) have no cluster label, so the host dashboards (where filter by `cluster` is applied) hides them.
The fix is dropping `cluster` label from host metrics so xcloud matches every other backend in SCT.

Fixes: SCT-244

### Testing

- [x] :green_circle: [xcloud backend cluster on AWS provider](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/xcloud-test/13/)
The `OS metrics` dashboard in Grafana now has metrics for all node - DB nodes + lodaer + monitor + sct-runner. Previously it was displaying only DB nodes metrics as the others were filtered out by cluster ID.
Screenshot from 2026-06-30 16-30-04

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -7249,6 +7249,24 @@ class BaseMonitorSet:
         """)
         node.remoter.run(f"bash -ce '{script}'")
 
+    @staticmethod
+    def _build_xcloud_scrape_config(promproxy_config):
+        """Build the scrape job from the Scylla Cloud promproxy config."""
+        scrape_configs = (yaml.safe_load(promproxy_config) or {}).get("scrape_configs") or []
+        return scrape_configs[0] if scrape_configs else None
+
+    @staticmethod
+    def _build_xcloud_cluster_relabel_config(xcloud_scrape_config):
+        """Create a relabel rule that adds the Scylla Cloud cluster label to SCT-scraped metrics."""
+        if match := re.search(r"/cluster/(\d+)/", xcloud_scrape_config.get("metrics_path") or ""):
+            return {
+                "source_labels": ["cluster"],
+                "regex": "^$",
+                "target_label": "cluster",
+                "replacement": f"#{match[1]}",
+            }
+        return None
+
     def configure_scylla_monitoring(self, node, sct_metrics=True, alert_manager=True):  # noqa: PLR0914
         cloud_prom_bearer_token = self.params.get("cloud_prom_bearer_token")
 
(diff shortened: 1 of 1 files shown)


fruch:
@amnonh what you think about this one, is dropping the cluster gonna break any monitor features ? (Beside supporting multiple cluster at the same time, which we don't need right now)

amnonh:
@fruch wouldn't the easy and correct way be to add a cluster label instead?

dimakr:
> @fruch wouldn't the easy and correct way be to add a cluster label instead?

@fruch checked this option. It would require somewhat more changes - 3 places (label loader, sct runner, monitor nodes) instead of 1 place with drop-the-label approach.
Also the label drop touches only `node_*` metrics (`scylla_*` ones are not touched). So it touches only OS Metrics dashboard, there is no other consumers of the `node_*`. With regards to alerts, the OS specific ones do not use `cluster` label.

So on the first glance, there should be no degradation if we drop label in SCT for `node_*` metrics

amnonh:
@dimakr I'm against removing the cluster label, it's safer when and if we combine data from different sources or use one dashboard for multiple clusters.
Adding a cluster label to the target file, couldn't be too complicated

amnonh:
I didn't mean to close it; my browser was doing something funky

sdcm/cluster.py
```diff
@@ -7249,6 +7249,24 @@ def find_overview_row(row):
...
+        if not scrape_configs:
+            return None
+
+        scrape_config = scrape_configs[0]
+        scrape_config.setdefault("metric_relabel_configs", []).append(
+            {
+                "source_labels": ["__name__"],
+                "regex": "node_.*",
+                "target_label": "cluster",
+                "replacement": "",
```
fruch: @dimakr isn't this enough ? assuming you know the cluster id upfront ? 

```yaml
    metric_relabel_configs:
      - source_labels: [cluster]
        regex: '^$'
        target_label: cluster
        replacement: actual_cluster_id
        action: replace
```

amnonh: Why do you need the relabel part? You can simply add a cluster label to the target file (or maybe I'm missing something)

fruch: Care to demonstrate? cause we never added anything to it but the addresses...

fruch:
@dimakr just confirm it was working as you expect with this last change, and we'll merge it

fruch (approved):

sdcm/cluster.py
amnonh: This is how you should configure it:
https://monitoring.docs.scylladb.com/stable/install/monitoring-stack.html#configure-scylla-monitoring-stack

```
- targets:
      - 172.17.0.2
      - 172.17.0.3
  labels:
      cluster: cluster1
      dc: dc1
```

dimakr:
> @dimakr just confirm it was working as you expect with this last change, and we'll merge it

yes, re-executed the test https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/xcloud-test/14/. Metrics for non-db nodes include the cluster id label now and are visible on e.g.  OS Metrics dashboard
@fruch

amnonh:
@dimakr @fruch I saw you went with relabel_config, it's not the end of the world, but less efficient than adding labels while scraping.

PR #15276, opened 2026-06-30
feature(mgmt): report actual restore method to Argus benchmark results

yarongilor:
Add a 'method' column to ManagerRestoreBenchmarkResult to report the actual Scylla Manager restore mode (Tablet/Native/Rclone) used during restore benchmark tests.

The detection reads the locally-collected scylla_manager.log using the existing File utility, looking for 'tablet aware restore' log entries that indicate SM internally used tablet-aware restore.

This is needed because SM does not yet expose the actual restore method through its CLI or progress output. When 'native' is requested via --method, SM may internally override it to use tablet-aware restore on clusters that support tablets. This override currently happens in private SM builds as a temporary workaround and is not reflected in the sctool restore progress output or task properties. Therefore, the only reliable way to determine the real method is by inspecting the SM log for tablet_restore activity.

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/yarongilor/job/restore-benchmark-test-tablets-aware/93/

### Reminders

### DIFF
--- a/mgmt_cli_test.py
+++ b/mgmt_cli_test.py
@@ -13,7 +13,9 @@
 #
 # Copyright (c) 2016 ScyllaDB
 
+import os
 import random
+import re
 import shutil
 import threading
 import time
@@ -53,6 +55,7 @@ from sdcm.utils.alternator.table_setup import alternator_backuped_tables
 from sdcm.utils.aws_utils import AwsIAM
 from sdcm.utils.features import is_tablets_feature_enabled
 from sdcm.utils.common import reach_enospc_on_node, clean_enospc_on_node
+from sdcm.utils.file import File
 from sdcm.utils.time_utils import ExecutionTimer
 from sdcm.mgmt.operations import ManagerTestFunctionsMixIn, SnapshotData
 from sdcm.sct_events.system import InfoEvent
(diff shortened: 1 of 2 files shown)


yarongilor:
@claude , review

mgmt_cli_test.py
```diff
@@ -1199,6 +1202,37 @@ def get_restore_extra_parameters(self) -> str:
...
 
+    def _detect_actual_restore_method(self, manager_backup_restore_method=None) -> str:
+        """Detect the actual restore method used by Scylla Manager from its log.
+
+        The requested method (e.g. 'native') can be internally overridden by SM
+        (e.g. to 'tablet' when the cluster supports tablets). This method reads the
+        locally-collected SM log to determine the real mode used.
+
+        Returns:
+            One of "Tablet", "Native", or "Rclone".
```
pehala, line 1222: Should return Enum :) The same Enum that we used to choose which method to use

"Tablet" is also not descriptive enough. Tablet-Aware  would be better

mgmt_cli_test.py
```diff
@@ -1199,6 +1202,37 @@ def get_restore_extra_parameters(self) -> str:
...
+            sm_log = File(sm_log_path)
+            tablet_lines = list(sm_log.read_lines_filtered(re.compile(r"tablet aware restore")))
+        except (OSError, FileNotFoundError) as err:
+            self.log.warning("Could not read SM log at %s: %s", sm_log_path, err)
+            tablet_lines = []
+
+        if tablet_lines:
+            return "Tablet"
+
+        if manager_backup_restore_method and hasattr(manager_backup_restore_method, "value"):
```
pehala: hasattr is a code smell why do you need to use it here?

yarongilor:
A [test](https://argus.scylladb.com/tests/scylla-cluster-tests/0041dcd2-1863-4cd9-a10a-a54751d4cb97) passed ok like:
image

mgmt_cli_test.py
```diff
@@ -1199,6 +1202,43 @@ def get_restore_extra_parameters(self) -> str:
...
+            tablet_lines = []
+
+        if tablet_lines:
+            return "Tablet"
+
+        method_value = manager_backup_restore_method.value if manager_backup_restore_method else None
+
+        if method_value == "rclone":
+            return "Rclone"
+        return "Native"
```
cezarmoise, line 1240: This logic is unclear.
`ObjectStorageUploadMode` has 3 modes: `AUTO`, `RCLONE`, `NATIVE`. And we have the 4th case, were `manager_backup_restore_method` was not set in the config, so `manager_backup_restore_method=None`.

I would like all 4 cases being treated individually.

Something like
```python
    match manager_backup_restore_method:
        case None:
            # Explain why None means Native
            return "Native"
        case ObjectStorageUploadMode.AUTO:
            # AUTO defaults to Native
            return "Native"
        case ObjectStorageUploadMode.NATIVE:
            return "Native"
        case ObjectStorageUploadMode.RCLONE:
            return "Rclone"
```

yarongilor: The `Auto` mode is less relevant for this performance testing, i don't think it's needed at all. It just tells SM to select it's most suitable default from the other modes.

the "real interesting" methods are rclone, native and tablets.

pehala: The enum should mirror what the actual manager has, so if it has auto we should be able to select it and use it. We should also not do any calls on the default when AUTO is used, we should report auto, or gather from manager what was actually used. By hardcoding this value it will grow stale

pehala:
Comments are not resolved, moving to draft

PR #15307, opened 2026-07-02
fix(kms): properly filter DB nodes in the `_rotate_kms_key`

vponomaryov:
The `_rotate_kms_key` method may pick up a DB node that is still being provisioned.
So, fix by filtering DB nodes out before picking up.

Fixes: SCT-451

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -5874,9 +5874,13 @@ class BaseScyllaCluster:
                         traceback=traceback.format_exc(),
                     ).publish()
                 try:
+                    ready_nodes = [node for node in db_cluster.data_nodes if node.remoter and node.db_up()]
+                    if not ready_nodes:
+                        self.log.warning("No ready nodes available for KMS encryption check, skipping this iteration")
+                        continue
                     nemesis_node_allocator = self.test_config.tester_obj().nemesis_allocator
                     with nemesis_node_allocator.run_nemesis(
-                        nemesis_label="KMS encryption check", node_list=db_cluster.data_nodes
+                        nemesis_label="KMS encryption check", node_list=ready_nodes
                     ) as target_node:
                         self.log.debug("Target node for 'rotate_kms_key' is %s", target_node.name)
 


cezarmoise (dismissed):
LGTM

sdcm/cluster.py
```diff
@@ -5874,9 +5874,17 @@ def _rotate_kms_key(kms_key_alias_name, kms_key_rotation_interval, db_cluster):
                         traceback=traceback.format_exc(),
                     ).publish()
                 try:
+                    ready_nodes = [
+                        node
+                        for node in db_cluster.data_nodes
+                        if not node.running_nemesis and node.remoter and node.db_up()
+                    ]
```
fruch: other places using the `run_nemesis` are using `self.cluster.get_nodes_up_and_normal()`

@dimakr I wondering maybe we should do this as default, and not just current running nemesis

i.e. for being on the safe side across all usages, and not doing this logic again and again, in multiple places

fruch: some of the usages are using `cluster.nodes` and not data nodes (which might be wrong for some cases...)

sdcm/cluster.py
```diff
@@ -5874,9 +5874,17 @@ def _rotate_kms_key(kms_key_alias_name, kms_key_rotation_interval, db_cluster):
                         traceback=traceback.format_exc(),
                     ).publish()
                 try:
+                    ready_nodes = [
+                        node
+                        for node in db_cluster.data_nodes
+                        if not node.running_nemesis and node.remoter and node.db_up()
```
dimakr: maybe `not node.running_nemesis` is not needed here? (`select_target_node` should already exclude nodes in `active_nemesis_on_nodes`)

sdcm/cluster.py
dimakr: > @dimakr I wondering maybe we should do this as default, and not just current running nemesis

agree, the logic is likely duplicated
But maybe should be done in a separate PR, to implement this properly (need to check if some nemesis are expecting down nodes on purpose)

sdcm/cluster.py
vponomaryov: The `select_target_node` changes state setting the `running_nemesis` attr.
So, we cannot use it here.

sdcm/cluster.py
vponomaryov: > some of the usages are using cluster.nodes and not data nodes (which might be wrong for some cases...)

In current case exactly data nodes must be the target.
Zero nodes have no data to encrypt.

sdcm/cluster.py
fruch: the selection logic is setting and working based on `running_nemesis`, user doesn't need to do anything about it. hence other places are passing the output of `get_nodes_up_and_normal()` without any other filter

vponomaryov:
@fruch , @dimakr 
So, I am not sure what is the expectation here.
Should I change something?

fruch:
> @fruch , @dimakr 
> So, I am not sure what is the expectation here.
> Should I change something?

Drop the logic regarding running_nemesis, that should be handled by run_nemesis logic


As a followup we should consider inlineing the up and normal logic, which is almost a must for user of this functionalty

sdcm/cluster.py
vponomaryov: Done

fruch (approved):

PR #15323, opened 2026-07-05
fix(trigger-matrix): resolve {branch} placeholder in scylla_version parameter

fruch:
The parameterized cron triggers pass `scylla_version="{branch}:latest"` as a literal string (from `configurations/triggers/perf-regression.yaml` and `tier1.yaml`). The `triggerMatrixPipeline` input sanitization rejects curly braces, causing all timer-triggered perf-regression builds to fail with:

```
ERROR: Invalid characters in parameter 'scylla_version': {branch}:latest
```

**Root cause:** The `{branch}` placeholder is embedded in the Jenkinsfile cron spec by the generator script. When Jenkins fires the timer, it passes the value literally. The `safePattern` regex (`^[a-zA-Z0-9_.:\-\/,@\s]*$`) does not allow `{` or `}`.

**Fix:** Resolve `{branch}` from the `job_folder` parameter (e.g. `scylla-master` → `master`) or from `JOB_NAME` before validation runs.

**Failing build:** https://jenkins.scylladb.com/job/scylla-master/job/sct_triggers/job/perf-regression-trigger-test/471/

### What could not be tested locally

- [x] **Timer-triggered perf-regression build**: Requires Jenkins timer to fire with `{branch}:latest` in parameters
  - Verify: Retrigger build 471 or wait for next weekly schedule
  - Expected: `{branch}` resolves to `master`, validation passes, downstream jobs triggered
  - https://jenkins.scylladb.com/job/scylla-master/job/sct_triggers/job/perf-regression-trigger-test/477/

### Suggested manual verification

```bash
# Build the job manually with the same params the timer sends:
# scylla_version={branch}:latest, job_folder=scylla-master, labels_selector=master-weekly
```

### DIFF
--- a/utils/build_system/create_test_release_jobs.py
+++ b/utils/build_system/create_test_release_jobs.py
@@ -385,10 +385,11 @@ class JenkinsPipelines:
             for job_file in job_files:
                 job_file = Path(root) / job_file  # noqa: PLW2901
                 if (job_file.suffix == ".jenkinsfile") and create_pipelines_jobs:
+                    suffix = "" if job_file.stem.endswith("-trigger") else job_name_suffix
                     self.create_pipeline_job(
                         job_file,
                         group_name=jenkins_path,
-                        job_name_suffix=job_name_suffix,
+                        job_name_suffix=suffix,
                         defines={**defines, **self.locate_job_overrides(job_file.stem, job_overrides)},
                     )
                 if (job_file.suffix == ".xml") and create_freestyle_jobs:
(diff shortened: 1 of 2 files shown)


utils/build_system/create_test_release_jobs.py
```diff
@@ -385,10 +385,11 @@ def create_job_tree(
             for job_file in job_files:
                 job_file = Path(root) / job_file  # noqa: PLW2901
                 if (job_file.suffix == ".jenkinsfile") and create_pipelines_jobs:
+                    suffix = "" if job_file.stem.endswith("-trigger") else job_name_suffix
```
fruch, line 388: @juliayakovlev we'll need to rebuild, and then we'll have the new trigger job

fruch: and we can ditch the old one

PR #15325, opened 2026-07-05
fix(nemesis): add timeout to repair nodetool commands to prevent indefinite hang

yarongilor:
fix(nemesis): enforce repair adaptive timeout to prevent indefinite hang

nodetool repair could hang indefinitely (24h+ test runs) because no
timeout was passed down to the repair command.

Use the value yielded by adaptive_timeout(Operations.REPAIR) as the
command timeout, as-is: the established 3h default, scaled per test by
the existing adaptive_timeout_multipliers config param ({repair: N}).
No hidden slack factors; tests whose repairs legitimately take longer
tune the multiplier in their yaml.

Run repair via long_running=True so SSH disconnects don't fail a
multi-hour repair, and fix run_long_running_cmd so the timeout is
actually enforced: the /proc poll used retry=10, which retries on any
exception - including CommandTimedOut - so a hung command would only
fail after ~10x the requested timeout. The poll is now deadline-based:
an SSH disconnect resumes watching the same pid with the remaining time
budget, while a timeout is terminal and never retried.

Fixes: https://scylladb.atlassian.net/browse/SCT-575
Fixes: https://scylladb.atlassian.net/browse/SCT-976
Refs: https://scylladb.atlassian.net/browse/SCT-964

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -1661,7 +1661,7 @@ class NemesisRunner:
                             SimpleStatement(f"DROP TABLE drop_table_during_repair_ks_{i}.standard1"), timeout=300
                         )
             finally:
-                thread.result()
+                thread.result(timeout=HOUR_IN_SEC * 8)
 
     def _major_compaction(self):
         with (
@@ -2086,21 +2086,37 @@ class NemesisRunner:
     def run_repair_nodetool(self, nodes: list, publish_event=True, timeout=HOUR_IN_SEC * 3):
         """
         Execute a repair using Nodetool, which runs both vnode repair (repair) and tablet repairs (cluster repair)
+
+        The adaptive_timeout yields the soft_timeout value. We use 2.5x that as the actual command
+        timeout so the SoftTimeoutEvent fires as a non-disruptive early warning well before the
+        command is killed.
         """
         for node in nodes:
             with (
-                adaptive_timeout(Operations.REPAIR, node, timeout=timeout),
+                adaptive_timeout(Operations.REPAIR, node, timeout=timeout) as soft_timeout,
                 self.action_log_scope(f"nodetool repair on {node.name} node"),
             ):
-                node.run_nodetool(sub_cmd="repair", publish_event=publish_event)
+                node.run_nodetool(
+                    sub_cmd="repair",
+                    publish_event=publish_event,
+                    timeout=soft_timeout * 2.5,
+                    long_running=True,
+                    retry=0,
+                )
 
         target_node = nodes[0]
         if is_tablets_feature_enabled(target_node):
             with (
-                adaptive_timeout(Operations.REPAIR, target_node, timeout=timeout),
+                adaptive_timeout(Operations.REPAIR, target_node, timeout=timeout) as soft_timeout,
                 self.action_log_scope("nodetool cluster repair", target=target_node.name),
             ):
-                target_node.run_nodetool(sub_cmd="cluster repair", publish_event=publish_event)
+                target_node.run_nodetool(
+                    sub_cmd="cluster repair",
+                    publish_event=publish_event,
+                    timeout=soft_timeout * 2.5,
+                    long_running=True,
+                    retry=0,
+                )
 
     @latency_calculator_decorator(legend="Run repair process through Scylla manager", cycle_name="_mgmt_repair_cli")
     def run_repair_manager(self, ignore_down_hosts: bool = False, timeout=HOUR_IN_SEC * 3):
--- a/sdcm/nemesis/__init__.py
+++ b/sdcm/nemesis/__init__.py
@@ -2141,21 +2141,37 @@ class NemesisRunner:
     def run_repair_nodetool(self, nodes: list, publish_event=True, timeout=HOUR_IN_SEC * 3):
         """
         Execute a repair using Nodetool, which runs both vnode repair (repair) and tablet repairs (cluster repair)
+
+        The command timeout is the repair adaptive timeout: the given `timeout` scaled by the
+        `adaptive_timeout_multipliers` config param (`{repair: N}`), so tests with longer repairs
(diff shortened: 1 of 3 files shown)


yarongilor:
@claude , review

sdcm/nemesis/__init__.py
```diff
@@ -1661,7 +1661,7 @@ def disrupt_no_corrupt_repair(self):
                             SimpleStatement(f"DROP TABLE drop_table_during_repair_ks_{i}.standard1"), timeout=300
                         )
             finally:
-                thread.result()
+                thread.result(timeout=HOUR_IN_SEC * 8)
```
fruch, line 1664: why 8h, when the default for the repair is timeout=HOUR_IN_SEC * 3 ?

sdcm/nemesis/__init__.py
```diff
@@ -2086,21 +2086,37 @@ def call_next_nemesis(self):
...
             with (
-                adaptive_timeout(Operations.REPAIR, target_node, timeout=timeout),
+                adaptive_timeout(Operations.REPAIR, target_node, timeout=timeout) as soft_timeout,
                 self.action_log_scope("nodetool cluster repair", target=target_node.name),
             ):
-                target_node.run_nodetool(sub_cmd="cluster repair", publish_event=publish_event)
+                target_node.run_nodetool(
+                    sub_cmd="cluster repair",
+                    publish_event=publish_event,
+                    timeout=soft_timeout * 2.5,
```
fruch, line 2116: why 2.5 ? where this figure is coming from ? why no doing the calculation within `adaptive_timeout` base on something specific, like data size or other metrics ?

is always repair takes up to 8h ?

sdcm/nemesis/__init__.py
yarongilor: The 8-hours is a hard timeout with a purpose to avoid wasting time and money for 25 hours, as in the original issue.
If there's any bug in scylla repair or in test-configuration, so we allow waiting up to this arbitrary limit of 8 hours before stopping the test. This allows telling if repair unexpectedly lasted 4 or 5 or 6 or 8 hours and keeping logs for it as well.
If you think a different value could be better then we can modify it as well.

sdcm/nemesis/__init__.py
yarongilor: This assumes the calculations is already done for `soft_timeout` (=`adaptive_timeout`) so no need for more calculations.

The `* 2.5` is a hard timeout with a purpose to avoid wasting time and money for 25 hours, as in the original issue.
If there's any bug in scylla repair or in test-configuration, so we allow waiting up to this arbitrary limit of soft-timeout*2.5  before stopping the test. This allows telling if repair unexpectedly lasted up to *2.5 and keeping logs for it as well. There will be not much added-value waiting for more than this rate.

fruch:
```
REPAIR = ("repair", _get_soft_timeout, ("timeout",))
```

_get_soft_timeout is passing the timeout as is, with no computation.
so assumption on existing calculation is kind of empty.

I wasn't talking about more, I was talking about less.
the hard coded HOUR_IN_SEC * 3 is coming from when we didn't collected the data, about how log things took, the questions is how can we tell it's the correct amount

sdcm/nemesis/__init__.py
fruch: where the definition of it's o.k. to give it more 3 hours before we'll fail is coming from ? 

who said it's o.k. to waste 3 hours on something that is stuck ?

yarongilor: reverted. Addressed in a different approach now, not changing any original timeout threshold.

sdcm/nemesis/__init__.py
yarongilor: reverted. using the original `repair_timeout` now.

yarongilor:
@scylla-zeus-bot , following the comment of https://github.com/scylladb/scylla-cluster-tests/pull/15681#issuecomment-5588892168.
can you compare this PR to PR 15681 and recommend if this one is a useful needed fix.

fruch:
@scylla-zeus-bot there multiple different fixes in this PR, each fix should go into its own commit, in case we'll need to revert, or backport separately

fruch:
@yarongilor I thought Zeus can do it.

Please split each fix into its own commit, so this can be merged

yarongilor:
> @yarongilor I thought Zeus can do it.
> 
> Please split each fix into its own commit, so this can be merged

@fruch , it's now split into 2 commits:

1. fix(nemesis): enforce repair adaptive timeout to prevent indefinite hang
2. fix(remoter): enforce run_long_running_cmd timeout as a real deadline

yarongilor:
Still have to run a short test to verify it i guess.

dimakr:
Should we backport this to release branches?
`retry=10` inside run_long_running_cmd is something from a 2024 commit, so it is in many release branches now.

dimakr:
`Drop the thread.result() timeout in disrupt_no_corrupt_repair: ... ` part of the PR description is not in the change, so that paragraph can be removed?

fruch:
> Should we backport this to release branches? `retry=10` inside run_long_running_cmd is something from a 2024 commit, so it is in many release branches now.

@dimakr now you'll need to help with those backports :)

yarongilor:
> `Drop the thread.result() timeout in disrupt_no_corrupt_repair: ... ` part of the PR description is not in the change, so that paragraph can be removed?

removed.

yarongilor:
@fruch , @dimakr , the nemesis passed ok in https://argus.scylladb.com/tests/scylla-cluster-tests/bf3ce566-aaa0-4dbb-af86-79e9915d34ef.

dimakr (approved):

yarongilor:
@fruch , can you please review/merge?

fruch (approved):

PR #15333, opened 2026-07-06
refactor(longevity-large-partition-200k-pks): Retune to 40% utilization and add LZ4 compression

yarongilor:
refactor(longevity-large-partition-200k-pks): Retune to ~40% utilization and add LZ4

Add a high-capacity override config for the large-partition 200K-PKs 4-day
longevity test, targeting ~40% disk utilization (after MV build) with LZ4
compression, and append it to the GCE tier1 pipeline's test_config.

The base test-case is left untouched, so the other five pipelines using it
(AWS longevity, AWS raft, AWS ICS 24h, EKS operator, vnodes GCE) keep their
current dataset, stress profile and instance types. Following the pattern of
configurations/2TB_high_disk_used_capacity.yaml, the override is named for the
capacity it targets rather than for a cloud.

Deltas vs the base test-case:
- Increase prepare from 1250 to 2500 partitions (5x500) to fill ~40% of a
  ~3.2 TB NVMe disk after MV build (z3-highmem-8-highlssd)
- Add 1500 new stress-write partitions (2501-4000) proportionally
- Reduce stress concurrency to ~25% of the base
  (write: 10->2, read: 50/200/150->12/50/37) to target ~50% CPU
- Add LZ4Compressor via ALTER TABLE in post_prepare_cql_cmds
- Set disk sizing constraint (local_disk_gb >= 3000 GB)
- Add adaptive_timeout_multipliers (decommission/new_node/remove_node/
  repair/mgmt_repair: 5) to accommodate longer operations with the
  larger dataset

Fixes: SCYLLADB-1439

### Testing

- https://argus.scylladb.com/tests/scylla-cluster-tests/2bd8be7b-ad77-43ad-88c6-8c2ea9ebcb69

### Reminders

### DIFF
--- a/jenkins-pipelines/oss/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
+++ b/jenkins-pipelines/oss/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
@@ -7,6 +7,6 @@ longevityPipeline(
     backend: 'gce',
     region: 'us-east1',
     test_name: 'longevity_large_partition_test.LargePartitionLongevityTest.test_large_partition_longevity',
-    test_config: 'test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml'
+    test_config: 'test-cases/longevity/longevity-large-partition-200k_pks-4days-gce.yaml'
 
 )
--- a/jenkins-pipelines/oss/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
+++ b/jenkins-pipelines/oss/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
@@ -7,6 +7,6 @@ longevityPipeline(
     backend: 'gce',
     region: 'us-east1',
     test_name: 'longevity_large_partition_test.LargePartitionLongevityTest.test_large_partition_longevity',
-    test_config: '''["test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml", "configurations/auto-repair-1h.yaml"]'''
+    test_config: '''["test-cases/longevity/longevity-large-partition-200k_pks-4days-high-disk-capacity.yaml", "configurations/auto-repair-1h.yaml"]'''
 
 )
--- a/test-cases/longevity/longevity-large-partition-200k_pks-4days-gce.yaml
+++ b/test-cases/longevity/longevity-large-partition-200k_pks-4days-gce.yaml
@@ -1,0 +1,112 @@
+# GCE-tuned longevity test: large partitions (200K PKs) — 4 days with LZ4 compression
+#
+# Tuned from the base longevity-large-partition-200k_pks-4days.yaml to:
+#   - Increase partition count to 2500 (5x500) targeting ~40% disk after MV build
+#   - Add LZ4 compression via post_prepare_cql_cmds
+#   - Reduce stress concurrency to ~25% of original to target ~50% CPU utilization
+#   - Set disk sizing constraint for GCE (z3-highmem-8-highlssd, ~3.2 TB NVMe)
+#   - Extend adaptive timeout multipliers for operations affected by larger dataset
+#
+# Disk utilization (measured on z3-highmem-8-highlssd, RF=3, 6 nodes, ~3.2 TB NVMe per node):
+#   - Prepare (base table + LZ4): ~24% at prepare end
+#   - After MV build completes:   ~40% (~1.29 TB / 3.22 TB per node)
+#   - Stress adds more data on top over the 4-day duration
+
+test_duration: 6480
+
+# Writing 2500 partitions, 200k clustering rows each (1 write command per loader).
+prepare_write_cmd: [
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=1    -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data",
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=501  -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data",
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=1001 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data",
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=1501 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data",
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=2001 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data"
+]
+
+# Not verifying the entire range, only some partitions: 501-550, 1001-1050, 1501-1550, 2001-2050, 2401-2450
+prepare_verify_cmd: [
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=50 -partition-offset=501  -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=50 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations 1 -validate-data",
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=50 -partition-offset=1001 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=50 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations 1 -validate-data",
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=50 -partition-offset=1501 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=50 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations 1 -validate-data",
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=50 -partition-offset=2001 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=50 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations 1 -validate-data",
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=50 -partition-offset=2401 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=50 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations 1 -validate-data"
+]
+
+stress_cmd: [
+    # Write additional 1500 partitions with 200k rows each (on top of prepare) - test total partitions is 4000
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=1500 -partition-offset=2501 -clustering-row-count=200000 -clustering-row-size=uniform:100..8192 -concurrency=2 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations=0 -duration=6420m -retry-number=30 -retry-interval=500ms,5s",
+    # Rewrite partitions 501-1500 to compensate for delete nemesis. Range 1501-2500 remains as tombstones.
+    # First 500 partitions (data_validation range) won't be deleted, so no re-write needed.
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=1000 -partition-offset=501 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=2 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations=0 -duration=6420m -retry-number=30 -retry-interval=500ms,5s"
+]
+
+stress_read_cmd: [
+    # Read the first 500 partitions with validate-data flag (never deleted during test)
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=500 -partition-offset=1 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=12 -connection-count=100 -consistency-level=quorum -rows-per-request=10 -timeout=90s -iterations=0 -duration=6420m -validate-data -retry-number=30 -retry-interval=500ms,5s",
+    # Read partitions 501-2500 (2000 partitions)
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=2000 -partition-offset=501 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -rows-per-request=10 -consistency-level=quorum -timeout=90s -concurrency=50 -connection-count=100 -iterations=0 -duration=6420m -retry-number=30 -retry-interval=500ms,5s",
+    # Read partitions 2501-4000 (1500 partitions) written during test (different row size)
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1500 -partition-offset=2501 -clustering-row-count=200000 -clustering-row-size=uniform:100..8192 -rows-per-request=10 -consistency-level=quorum -timeout=90s -concurrency=37 -connection-count=100 -iterations=0 -duration=6420m -retry-number=30 -retry-interval=500ms,5s"
+]
+
+# Create MV, set tombstone_gc, and apply LZ4 compression to base table (MV inherits)
+post_prepare_cql_cmds: [
+    "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where ck IS NOT NULL AND v is not null PRIMARY KEY (v, pk, ck) with comment = 'TEST VIEW'",
+    "ALTER TABLE scylla_bench.test with tombstone_gc = {'mode': 'repair', 'propagation_delay_in_seconds':'300'};",
+    "ALTER TABLE scylla_bench.test WITH compression = {'sstable_compression': 'LZ4Compressor'};"
+]
+
+n_db_nodes: 6
+n_loaders: 5 # Each loader will have 1 scylla-bench process at every step of the test (prepare, verify, stress & stress_read)
+
+round_robin: true
+
+sizing_db:
+  vcpu: 8
+  memory: '>=64'
+  local_disk_gb: '>=3000'
+sizing_loader:
+  vcpu: 16
+  memory: '>=16'
+
+nemesis_class_name: 'SisyphusMonkey'
+nemesis_seed: '016'
+nemesis_interval: 30
+nemesis_during_prepare: false
+
+# Adaptive timeout multipliers for operations that take longer with the larger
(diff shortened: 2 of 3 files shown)


configurations/large-partition-200k-pk-high-capacity.yaml
pehala, line 1: Create copy of the testcase or update the existing one, instead of adding overlay, it is extremely confusing.

configurations/large-partition-200k-pk-high-capacity.yaml
```diff
@@ -0,0 +1,73 @@
...
+    # Read partitions 2501-4000 (1500 partitions) written during test (different row size)
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1500 -partition-offset=2501 -clustering-row-count=200000 -clustering-row-size=uniform:100..8192 -rows-per-request=10 -consistency-level=quorum -timeout=90s -concurrency=37 -connection-count=100 -iterations=0 -duration=6420m -retry-number=30 -retry-interval=500ms,5s"
+]
+
+# Create MV, set tombstone_gc, and apply LZ4 compression to base table (MV inherits)
+post_prepare_cql_cmds: [
+    "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where ck IS NOT NULL AND v is not null PRIMARY KEY (v, pk, ck) with comment = 'TEST VIEW'",
+    "ALTER TABLE scylla_bench.test with tombstone_gc = {'mode': 'repair', 'propagation_delay_in_seconds':'300'};",
+    "ALTER TABLE scylla_bench.test WITH compression = {'sstable_compression': 'LZ4Compressor'};"
```
pehala: LZ4withdict is default now, no need to do anything

pehala (changes requested):

configurations/large-partition-200k-pk-high-capacity.yaml
yarongilor: created a copy.

configurations/large-partition-200k-pk-high-capacity.yaml
yarongilor: @pehala ,
scylla_bench.test never falls under that default at all

scylla-bench's own code hardcodes an explicit, empty compression clause when creating the table (confirmed directly in main.go, scylladb/scylla-bench):

Query(session, "CREATE TABLE IF NOT EXISTS "+keyspaceName+"."+tableName+
    " (pk bigint, ck bigint, v blob, PRIMARY KEY(pk, ck)) WITH compression = { }")
tableName defaults to "test".

In CQL/ScyllaDB, WITH compression = { } is not the same as omitting the clause:

cf_prop_defs.cc's get_compression_options() returns a value (even if empty) whenever the clause is present, vs. nullopt only when the clause is absent.
sstables/compressor.cc then interprets an empty options map as algorithm::none — i.e., compression explicitly disabled, not "inherit server default."
So scylla_bench.test is created uncompressed, regardless of any server-side default (old or new).

We can consider filing an issue for s-b to change this query and remove the empty compression.

test-cases/longevity/longevity-large-partition-200k_pks-4days-gce.yaml
```diff
@@ -0,0 +1,112 @@
+# GCE-tuned longevity test: large partitions (200K PKs) — 4 days with LZ4 compression
+#
+# Tuned from the base longevity-large-partition-200k_pks-4days.yaml to:
+#   - Increase partition count to 2500 (5x500) targeting ~40% disk after MV build
```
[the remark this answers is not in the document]
pehala: @yarongilor Could you please update the description to match the actual changes?

test-cases/longevity/longevity-large-partition-200k_pks-4days-gce.yaml
```diff
@@ -0,0 +1,112 @@
...
+    # Read partitions 501-2500 (2000 partitions)
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=2000 -partition-offset=501 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -rows-per-request=10 -consistency-level=quorum -timeout=90s -concurrency=50 -connection-count=100 -iterations=0 -duration=6420m -retry-number=30 -retry-interval=500ms,5s",
+    # Read partitions 2501-4000 (1500 partitions) written during test (different row size)
+    "scylla-bench -workload=sequential -mode=read -replication-factor=3 -partition-count=1500 -partition-offset=2501 -clustering-row-count=200000 -clustering-row-size=uniform:100..8192 -rows-per-request=10 -consistency-level=quorum -timeout=90s -concurrency=37 -connection-count=100 -iterations=0 -duration=6420m -retry-number=30 -retry-interval=500ms,5s"
+]
+
+# Create MV, set tombstone_gc, and apply LZ4 compression to base table (MV inherits)
+post_prepare_cql_cmds: [
+    "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where ck IS NOT NULL AND v is not null PRIMARY KEY (v, pk, ck) with comment = 'TEST VIEW'",
```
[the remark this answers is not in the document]
yarongilor: we can leave it like that and might consider updating MV-create-command on all longevities in a separate future PR.

test-cases/longevity/longevity-large-partition-200k_pks-4days-gce.yaml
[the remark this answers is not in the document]
yarongilor: @pehala , fixed.

configurations/large-partition-200k-pk-high-capacity.yaml
pehala: Yeah, that is a bug for SB for sure

pehala (dismissed):

yarongilor:
@scylladb/qa-maintainers , please review.

roydahan:
Where is the reason of why do we need it?

pehala:
> Where is the reason of why do we need it?

It is the same as with all tier 1 changes: Have a test with bigger storage utilization, bigger, more realistic load and/or decreasing instance size.

jenkins-pipelines/oss/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
```diff
@@ -7,6 +7,6 @@ longevityPipeline(
     backend: 'gce',
     region: 'us-east1',
     test_name: 'longevity_large_partition_test.LargePartitionLongevityTest.test_large_partition_longevity',
-    test_config: 'test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml'
+    test_config: 'test-cases/longevity/longevity-large-partition-200k_pks-4days-gce.yaml'
```
fruch: why doing gce specific configuration ? I would assume it's temporary until one can work with in the test based on disk sizes, and make it work on any cloud in the same ratios 

and what about `longevity-large-partition-200k_pks-4days.yaml` ? isn't not gonna be used anymore, please remove it, if it's dangling configuration file.

roydahan: No, just don't do any GCE or any cloud specific configuration.

yarongilor: The task of https://scylladb.atlassian.net/browse/SCYLLADB-1439 instructs: Retune longevity-large-partition-200k-pks-**gce**-test and add LZ4.

There are 5 pipelines using the original yaml of `test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml`:

```
jenkins-pipelines/oss/raft/longevity-large-partition-200k-pks-4days.jenkinsfile
jenkins-pipelines/oss/longevity/longevity-large-partition-200k-pks-4days.jenkinsfile
jenkins-pipelines/oss/features/ICS/ics-longevity-toggle-strategy-large-partitions-24h.jenkinsfile
jenkins-pipelines/operator/eks/longevity-scylla-operator-large-partition-200k-pks-4days.jenkinsfile
jenkins-pipelines/oss/vnodes/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
```
So we probably don't want to double the instance type for all of them but rather to the gce alone.
The file already contains sizing, so it's cloud-agnostic like:
```
sizing_db:
  vcpu: 8
  memory: '>=64'
sizing_loader:
  vcpu: 16
  memory: '>=16'
```
only the `local_disk_gb` will be added in a new file, like:
`  local_disk_gb: '>=3000' `

added configurations/large-partition-200k-pk-high-capacity.yaml — an override holding only the retune deltas (2500-partition prepare, 1500 extra stress partitions, reduced concurrency, LZ4, local_disk_gb >= 3000, adaptive timeout multipliers), appended to the GCE tier1 pipeline's test_config. Same pattern as configurations/2TB_high_disk_used_capacity.yaml, and named for the capacity it targets, not for a cloud. The base test-case is untouched, so the other 5 pipelines run exactly what they run today.

pehala: I actually liked the new file better, I think the only problem there was that it contained GCE in the name while it was already cloud agnostic (only pipeline should contain the name I guess)

Overriding stress command and other essential configs in overlay does not seem correct to me. This is IMO separate testcase now, and we cannot delete the original since it is used everywhere.

@roydahan  @fruch Did you have any concrete concerns about how it was before except for the "gce" in the testcase name?

roydahan: Yes, having new file is better and without gce.
I see that @yarongilor just did it.

pehala: > I see that @yarongilor just did it.

No he did not, he created overlay, i.e. that file requires the original testcase to be added in the pipeline as well. The testcase file should be complete here.

roydahan: Yes, I didn't notice it. I also don't like it.

configurations/large-partition-200k-pk-high-capacity.yaml
```diff
@@ -0,0 +1,84 @@
...
+# LZ4 compression of the base table (the MV inherits it).
+post_prepare_cql_cmds: [
+    "CREATE MATERIALIZED VIEW scylla_bench.view_test AS SELECT * FROM scylla_bench.test where ck IS NOT NULL AND v is not null PRIMARY KEY (v, pk, ck) with comment = 'TEST VIEW'",
+    "ALTER TABLE scylla_bench.test with tombstone_gc = {'mode': 'repair', 'propagation_delay_in_seconds':'300'};",
+    "ALTER TABLE scylla_bench.test WITH compression = {'sstable_compression': 'LZ4Compressor'};"
+]
+
+# Deep-merged onto the base test-case's sizing_db (vcpu: 8, memory: '>=64').
+sizing_db:  # no azure/oci match with the disk constraint
+  local_disk_gb: '>=3000'
```
roydahan: This doesn't make sense, if you tune it to be 40% of disk capacity.

configurations/large-partition-200k-pk-high-capacity.yaml
```diff
@@ -0,0 +1,84 @@
...
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=1001 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data",
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=1501 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data",
+    "scylla-bench -workload=sequential -mode=write -replication-factor=3 -partition-count=500 -partition-offset=2001 -clustering-row-count=200000 -clustering-row-size=uniform:512..5120 -concurrency=250 -connection-count=10 -consistency-level=quorum -rows-per-request=10 -timeout=90s -validate-data"
+]
```
roydahan: What is the raw dataset size?

jenkins-pipelines/oss/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
```diff
@@ -7,6 +7,6 @@ longevityPipeline(
     backend: 'gce',
     region: 'us-east1',
     test_name: 'longevity_large_partition_test.LargePartitionLongevityTest.test_large_partition_longevity',
-    test_config: '''["test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml", "configurations/auto-repair-1h.yaml"]'''
+    test_config: '''["test-cases/longevity/longevity-large-partition-200k_pks-4days.yaml", "configurations/auto-repair-1h.yaml", "configurations/large-partition-200k-pk-high-capacity.yaml"]'''
```
roydahan: Assuming we really want this task, it should edit the original yaml and not add new configuration file.
In addition, It should be tested and verified both on AWS and GCE.

I would be really careful considering the necessity of this task based on its costs.

pehala: I think new testcase file where we explicitly say we tested on GCE is cheapest and safest. We dont need to do additional testing for it, given it uses sizing risk on AWS is small

roydahan: Fastest and cheapest for this task is not doing it.
Investing the time and budget on real customer workload with latte would be more beneficial.

pehala: > Fastest and cheapest for this task is not doing it.

Thats why I want to merge whats already done, just with a proper structured files. I dont want more reruns of this

pehala (dismissed):

configurations/large-partition-200k-pk-high-capacity.yaml
pehala: Still not addressed, but I am ok with it as we can start using templated stress in a followup to target data utilization regardless of the instance

configurations/large-partition-200k-pk-high-capacity.yaml
yarongilor: https://github.com/yarongilor/scylla-cluster-tests/blob/a4bae58a557e682b4ae12c8c97ee76a60c02f985/test-cases/longevity/longevity-large-partition-200k_pks-4days-high-disk-capacity.yaml#L17:

Raw dataset size after prepare (logical payload, before compression):
  500M rows (2500 partitions x 200K rows) x ~2816 B mean row  = ~1.41 TB
  plus the materialized view, which duplicates every row      = ~2.82 TB
  at RF=3                                                     = ~8.45 TB cluster-wide
  over 6 nodes                                                = ~1.41 TB per node

configurations/large-partition-200k-pk-high-capacity.yaml
yarongilor: there's nothing to address in here. It is currently an SCT limitation, and it's stated in the file as well -
https://github.com/yarongilor/scylla-cluster-tests/blob/a4bae58a557e682b4ae12c8c97ee76a60c02f985/test-cases/longevity/longevity-large-partition-200k_pks-4days-high-disk-capacity.yaml#L32.
So this appears to be the best available solution currently, and it was tested and work as expected.

jenkins-pipelines/oss/tier1/longevity-large-partition-200k-pks-4days-gce.jenkinsfile
yarongilor: fixed

configurations/large-partition-200k-pk-high-capacity.yaml
yarongilor: @roydahan , @pehala , not sure i understand this correctly but it might be a good catch of another SCT bug in GCE instance catalog.

the calculation as proven on test is:
```
it isn't an attempt to express "40% of disk". The dataset is a fixed size, and >=3000 is only the floor that picks a machine big enough to hold it:

Raw dataset after prepare: ~1.41 TB logical per node — 500M rows (2500 partitions × 200K) × ~2816 B mean row = ~1.41 TB, ×2 for the MV, ×RF3, ÷ 6 nodes. On disk that measures 1.29 TB; LZ4 only buys ~8% because the scylla-bench payload is pseudo-random.
z3-highmem-8-highlssd has 3,000 GiB (3.22 TB) of usable NVMe, so that comes to ~40% once the MV build completes, and ~46% now that the stress phase is writing partitions 2501-4000.
```
the sct catalog bug is:
```
instance catalog lists z3-highmem-8-highlssd as 6000 GB / 2 local SSD partitions, but the GCE docs say 1 partition of 3,000 GiB, and the run's Grafana agrees (1.48 TB used at 46% → 3.22 TB total). _Z3_PARTITION_COUNTS looks 2× off for every z3 highlssd shape
```

z3 highlssd local-SSD capacity is 2× over-reported in the instance catalog.

_Z3_PARTITION_COUNTS lists double the real partition count for every z3 highlssd shape. Since catalog_generator.py:556 computes local_disk_gb = partition_count × 3000, every one of these instances shows twice its actual disk:

machine type	SCT	GCE docs
z3-highmem-8-highlssd	2	1
z3-highmem-16-highlssd	4	2
z3-highmem-22-highlssd	6	3
z3-highmem-32-highlssd	8	4
z3-highmem-44-highlssd	12	6
z3-highmem-88-highlssd	24	12

As for the inaccuracy of 40%, It is currently an SCT limitation, and it's stated in the file as well -
https://github.com/yarongilor/scylla-cluster-tests/blob/a4bae58a557e682b4ae12c8c97ee76a60c02f985/test-cases/longevity/longevity-large-partition-200k_pks-4days-high-disk-capacity.yaml#L32.

yarongilor: opened https://scylladb.atlassian.net/browse/SCT-879

yarongilor: @roydahan , the value is changed to `local_disk_gb: '3000-4000'`

roydahan: Did you verify it finds the correct instance types in AWS and GCP?

yarongilor: @roydahan , the test fails as expected for both of them - 
GCP - fails provisioning because of https://scylladb.atlassian.net/browse/SCT-879.
AWS - fails provisioning because no such disk-size range available instance.

The GCP will probably run ok if it's tested with the 879 fix of https://github.com/scylladb/scylla-cluster-tests/pull/15858

yarongilor: The provisioning passed ok with the above 879 fix in https://argus.scylladb.com/tests/scylla-cluster-tests/7f2d4f0d-4ab6-4541-b42f-882dae453b9f.

roydahan: Please edit the range so it will also match an AWS instance. even if it's not exactly 40%, it's ok if it'll be close enough.

pehala: > Please edit the range so it will also match an AWS instance. even if it's not exactly 40%, it's ok if it'll be close enough.

This will be done as a followup, there is a new feature which will help with that. Changing the range to match AWS will require a rerun and we dont want to spend more money here. Lets not delay this by unnecessarily adding stuff.

For Tier 1 we will be running on GCE so the tuning will hold.

roydahan: It doesn't require a rerun, change it to 3000-5000, it should match i8ge.2xlarge.
I don't want anyone later to reverse engineer this later to be able to run on AWS.

pehala: > It doesn't require a rerun, change it to 3000-5000, it should match i8ge.2xlarge.
> I don't want anyone later to reverse engineer this later to be able to run on AWS.

Ok, if the problem is just a sizing definition then yeah, it makes sense to do here.

yarongilor: fixed to `local_disk_gb: '3000-5000'`

roydahan (approved):
Approved assuming that if we start bumping into many non-scylla issues in tier1 we may choose to go back to the older test config file.

yarongilor:
@scylladb/qa-maintainers , please approve/merge.

PR #15216, opened 2026-06-25
ci(sct-runner): bump base Ubuntu image to 26.04

vponomaryov:
List of changes:
- AWS:
  - Switch to the Ubuntu 26.04
  - Use image resolver instead of the hardcoded AMI
  - Autodetect arch instead of hardcoding it
  - Use modern 'signed-by' approach instead of the 'apt-key add'
- GCE: switch to the Ubuntu 26.04
- Azure:
  - Switch to the Ubuntu 26.04
  - Do several fixing to the SCT runner VM creation process
- OCI:
  - Switch to the Ubuntu 26.04 using custom non-official image
  - Do several fixing to the SCT runner VM creation process
    
Fixes: https://scylladb.atlassian.net/browse/SCT-306

### Testing

- [ ]

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #15334, opened 2026-07-06
fix(remote): fix stress metrics not published on Ubuntu 26 / Python 3.14

vponomaryov:
Root cause: a race condition between stress command completion and the
LatteExporter (FileFollowerThread) being stopped. When latte's stdout is
pipe-buffered through docker exec, all sampling data arrives at once when
the process exits. The cmd_runner.run() returns, the `with` block exits,
and LatteExporter.__exit__() signals stop — before the file-following
thread reads the freshly-written data.

On Ubuntu 24 this race rarely triggered because either the pipe flushed
incrementally or scheduling gave the reader thread time. On Ubuntu 26
(kernel 6.14, Python 3.14), the tighter scheduling consistently loses
the race.

Three fixes applied:

1. FileFollowerIterator: remove select.poll() (unreliable on regular
   files with kernel 6.14) and add a drain loop that reads all remaining
   data after stop is signaled.

2. StressExporter.run(): remove `if self.stopped(): break` inside the
   for-loop over follow_file(). This was preventing the drain from
   working — the iterator yields remaining lines but the consumer broke
   out immediately.

3. LogWriteWatcher/OutputWatcher: replace shared self.len counter with
   per-thread position tracking. Invoke calls submit() from separate
   stdout/stderr threads with independent buffers, corrupting the shared
   counter under aggressive thread scheduling.

Closes: #SCT-583

### Testing

- [ ]

### Reminders

### DIFF
--- a/sdcm/loader.py
+++ b/sdcm/loader.py
@@ -117,9 +117,6 @@ class StressExporter(FileFollowerThread, metaclass=ABCMeta):
                 continue
 
             for line in self.follow_file(self.stress_log_filename):
-                if self.stopped():
-                    break
-
                 if self.skip_line(line=line):
                     continue
 
--- a/sdcm/remote/base.py
+++ b/sdcm/remote/base.py
@@ -19,6 +19,7 @@ import logging
 import re
 import os
 import subprocess
+import threading
 from textwrap import dedent
 
 from invoke.watchers import StreamWatcher, Responder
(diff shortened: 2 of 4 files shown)


vponomaryov:
It must go to any branch where the following one went in:
- https://github.com/scylladb/scylla-cluster-tests/pull/15216

fruch:
@vponomaryov locally it doesn't reproduce ?

fruch:
@vponomaryov the information on the description isn't completely accurate

the kernel is 7.0  for 26.04
we are using debian python within the hydra image, the host is what changed.

 I would recommend spinning sct-runner in AWS and running the specific test with hydra --execute-on-runner option

vponomaryov:
> @vponomaryov locally it doesn't reproduce ?

It works on SCT runner triggering manually using "execute on runner" option from laptop...

So, it is some race condition, maybe related to running all tests, not just 4 latte ones...

vponomaryov:
For now I can tell that the failed test, in different test runs, runs for 0-7 seconds whereas the latte command must run for 10 seconds.
So, latte command fails for some reason and we have no info about it...

fruch:
> For now I can tell that the failed test, in different test runs, runs for 0-7 seconds whereas the latte command must run for 10 seconds. So, latte command fails for some reason and we have no info about it...

unfortunately we don't collect the system message of the sct runner being used to run the tests...

vponomaryov:
Created another oneliner here:
- https://github.com/scylladb/scylla-cluster-tests/pull/15343

I expect it to work because it is what passed in this draft PR

vponomaryov:
Closing in favor of the following PR that proved to be working:
- https://github.com/scylladb/scylla-cluster-tests/pull/15343

PR #14573, opened 2026-05-10
feature(perf-regression): add 2.5TB elasticity latency regression tests

juliayakovlev:
Added new performance regression configuration and Jenkins pipelines for 2.5TB elasticity latency tests. The configuration prepares 2.6B rows across 4 loaders and defines write, read, and mixed stress workloads with fixed throughput rates. The cluster will be grown from 3 to 12 nodes during the test (nemesis_add_node_cnt=9). Jenkins pipelines target i4i and i8g ARM instance types on AWS for enterprise perf-v17 branch.

Task: https://scylladb.atlassian.net/browse/QAINFRA-101

### Testing

- [ ]

### PR pre-checks (self review)

- [ ] I added the relevant `backport` labels
- [ ] I didn't leave commented-out/debugging code

### Reminders

- Add New configuration option and document them (in `sdcm/sct_config.py`)
- Add unit tests to cover my changes (under `unit-test/` folder)
- Update the Readme/doc folder relevant to this change (if needed)

PR #15344, opened 2026-07-07
fix(nemesis): resolve timeout and node allocation failures in parallel grow-shrink

juliayakovlev:
Fix two issues observed during GrowShrinkClusterNemesis on large (2.5TB) clusters with parallel decommission:
1. Tablet migration timeout: increase quiesce_topology timeout from 3600s to 7200s in add_new_nodes, matching the ParallelObject timeout. With large datasets and many new nodes, tablet rebalancing routinely exceeds one hour.
2. Node allocation conflict in verify_decommission: add retry with backoff (up to 10 attempts) when all candidate nodes are temporarily busy. During parallel decommission of 9 nodes, multiple verify_decommission calls competed for the same 3 remaining nodes, causing AllNodesRunNemesisError.

### Testing

- [x] It was tested while https://github.com/scylladb/scylla-cluster-tests/pull/14573

### Reminders

### DIFF
--- a/sdcm/cluster.py
+++ b/sdcm/cluster.py
@@ -6469,9 +6469,25 @@ class BaseScyllaCluster:
         target_node_ip = node.ip_address
         node_allocator = self.test_config.tester_obj().nemesis_allocator
         verification_candidates = [n for n in self.data_nodes if n != node]
-        with node_allocator.run_nemesis(
-            nemesis_label="verify decommission", node_list=verification_candidates
-        ) as verification_node:
+
+        # Retry node allocation in case all candidates are temporarily busy
+        # (e.g. during parallel decommissions where multiple verify_decommission calls compete)
+        from sdcm.nemesis.utils.node_allocator import AllNodesRunNemesisError  # noqa: PLC0415
+
+        max_attempts = 10
+        for attempt in range(max_attempts):
+            try:
+                allocator_context = node_allocator.run_nemesis(
+                    nemesis_label="verify decommission", node_list=verification_candidates
+                )
+                verification_node = allocator_context.__enter__()
+                break
+            except AllNodesRunNemesisError as err:
+                if attempt == max_attempts - 1:
+                    raise AllNodesRunNemesisError("Failed to allocate node for verify_decommission") from err
+                time.sleep(5 * (attempt + 1))
+
+        try:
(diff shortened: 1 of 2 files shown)


sdcm/cluster.py
```diff
@@ -6465,9 +6465,25 @@ def get_node_ip_list(verification_node):
...
         verification_candidates = [n for n in self.data_nodes if n != node]
-        with node_allocator.run_nemesis(
-            nemesis_label="verify decommission", node_list=verification_candidates
-        ) as verification_node:
+
+        # Retry node allocation in case all candidates are temporarily busy
+        # (e.g. during parallel decommissions where multiple verify_decommission calls compete)
+        from sdcm.nemesis.utils.node_allocator import AllNodesRunNemesisError  # noqa: PLC0415
+
+        max_attempts = 10
```
fruch, line 6477: this sounds like something `node_allocator.run_nemesis()` should have out of the box, like extra params for retry and backoff

it would alos solve the inline import


even that it's a bit rare situation, I think it should be part of the allocation logic itself 

@dimakr what do you think ?

juliayakovlev:
updated the branch

sdcm/nemesis/__init__.py
```diff
@@ -3872,7 +3872,14 @@ def disrupt_corrupt_then_scrub(self):
...
         self.actions_log.info(f"New nodes added: {', '.join(node.name for node in nodes)}")
-        ParallelObject(objects=self.cluster.data_nodes, timeout=7200).run(wait_no_tablets_migration_running)
+        tablet_migration_timeout = 7200
+
+        def _wait_no_tablets_migration(node):
+            wait_no_tablets_migration_running(node, timeout=tablet_migration_timeout)
+
+        ParallelObject(objects=self.cluster.data_nodes, timeout=tablet_migration_timeout + 120).run(
+            _wait_no_tablets_migration
+        )
```
cezarmoise, line 3882: This change is not needed.
The parallel call will be removed (see https://github.com/scylladb/scylla-cluster-tests/pull/15367)
And `TABLET_MIGRATION` timeout is dynamically calculated, the timeout passed to the method is just the fallback value, won't be used.

juliayakovlev:
This part was changed by https://github.com/scylladb/scylla-cluster-tests/pull/15367. Closing

PR #15345, opened 2026-07-07
fix(latency-decorator): preserve exceptions through latency collection

juliayakovlev:
The latency_calculator_decorator now catches exceptions from the decorated function, still collects and reports latency metrics to Argus, then re-raises. Previously, any exception during the decorated operation caused the entire step's latency data to be silently dropped.

The re-raise is guarded on all exit paths (two early returns and the main path) to ensure the original exception always propagates to the caller regardless of monitoring configuration or workload detection.

### Testing

- [x] It was tested while https://github.com/scylladb/scylla-cluster-tests/pull/14573

### Reminders

### DIFF
--- a/sdcm/utils/decorators.py
+++ b/sdcm/utils/decorators.py
@@ -185,7 +185,7 @@ def _find_hdr_tags(*args):
     raise ValueError("Failed to find 'hdr_tags'")
 
 
-def latency_calculator_decorator(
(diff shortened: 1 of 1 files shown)


fruch:
@juliayakovlev anything keeping this draft ?

juliayakovlev:
Updated the branch

fruch (approved):

PR #15346, opened 2026-07-07
fix(argus-events): drain queued events before terminating postman

soyacz:
## What / Why

`ArgusEventPostman.terminate()` only set `stop_event`, which halts the aggregator's `outbound_events` generator (`while not stop_event.is_set()`). Any events still sitting in the aggregator's `outbound_queue` at teardown were therefore **abandoned instead of posted** — a teardown-time `ERROR` event could silently vanish from Argus.

This makes the postman **drain the queue before terminating**, while still **quitting after a timeout (60s by default)**.

## Changes

- **`stop()` override** on `ArgusEventPostman`: `terminate()` → drain → bounded `join()`. Same `(self, timeout=None)` signature as `BaseEventsProcess.stop`, so `stop_events_device`'s polymorphic call picks it up with no caller changes.
- **`_drain_outbound(deadline)`**: reads the aggregator's `outbound_queue` **directly** (not via `inbound_events()`, which stops yielding the instant `terminate()` fires) and posts queued events **ERROR/CRITICAL-first**, so the most important events go out if the budget runs out mid-drain.
- **`ARGUS_POSTMAN_DRAIN_TIMEOUT = 60`** (module constant): the drain budget, **independent of the caller's `timeout`**. Needed because the production teardown path calls `stop(timeout=EVENTS_PROCESS_STOP_TIMEOUT=10)` (`setup.py:99`), which would otherwise cap the drain to 10s.
- The whole drain runs in a **daemon thread joined with the 60s deadline**, so a single hung `submit_event` POST (an uninterruptible C-level socket read) cannot make `stop()` exceed the budget.
- Extracted `_submit_event()` so `run()` and the drain share one `verbose_suppress`-wrapped submit path; wrapped `run()`'s consume loop in `verbose_suppress` to avoid a teardown-race `AttributeError` when the upstream is already gone.

## Testing

New pytest-style unit tests in `unit_tests/unit/test_argus_postman_stop.py`:
- `test_stop_drains_pending_events_error_first` — all queued events posted, highest severity first.
- `test_stop_is_bounded_when_submit_hangs` — `stop()` returns within budget even when `submit_event` blocks forever.
- `test_stop_without_queued_events_returns_quickly` — no-op fast path.

Verified: 3 new tests pass under `-n2 --random-order` (×3, no flakiness); existing `test_sct_events_{grafana,events_processes,events_analyzer}` still green; `pre-commit` (ruff + ruff-format) passes on both files.

## Notes

- No shared base-class signature changed (override on one subclass only), so the SCT override-audit rule does not apply.
- The base's `isinstance(self, multiprocessing.Process)` kill branch was always dead code for this `threading.Thread`; the override intentionally drops it.
- No backend-provisioning code touched (no provision label needed).

### DIFF
--- a/sdcm/sct_events/argus.py
+++ b/sdcm/sct_events/argus.py
@@ -34,6 +34,7 @@ from sdcm.utils.argus import Argus
 
 
 ARGUS_EVENT_AGGREGATOR_TIME_WINDOW: float = 90  # seconds
+
 LOGGER = logging.getLogger(__name__)
 
 
@@ -109,14 +110,20 @@ class ArgusEventPostman(BaseEventsProcess[SCTArgusEvent, None], threading.Thread
     def run(self) -> None:
         self.enabled.wait()
 
-        for event in self.inbound_events():  # events from ArgusAggregator
-            with verbose_suppress(
-                "ArgusEventPostman failed to post an event to '%s' endpoint.\nEvent: %s",
-                self._argus_client.Routes.SUBMIT_EVENT,
-                event,
-            ):
-                if self._argus_client:
-                    self._argus_client.submit_event(event)
+        with verbose_suppress("ArgusEventPostman failed while consuming inbound events"):
+            for event in self.inbound_events():  # events from ArgusAggregator
+                self._submit_event(event)
+
+    def _submit_event(self, event: SCTArgusEvent) -> None:
+        """Post a single event to Argus, swallowing and logging any failure (see verbose_suppress)."""
+        if not self._argus_client:
+            return
+        with verbose_suppress(
+            "ArgusEventPostman failed to post an event to '%s' endpoint.\nEvent: %s",
+            self._argus_client.Routes.SUBMIT_EVENT,
+            event,
+        ):
+            self._argus_client.submit_event(event)
 
     def enable_argus_posting(self) -> None:
         self._argus_client = Argus.get().client
--- a/sdcm/sct_events/argus.py
+++ b/sdcm/sct_events/argus.py
@@ -34,7 +34,6 @@ from sdcm.utils.argus import Argus
 
 
 ARGUS_EVENT_AGGREGATOR_TIME_WINDOW: float = 90  # seconds
-
 LOGGER = logging.getLogger(__name__)
 
 
@@ -110,9 +109,8 @@ class ArgusEventPostman(BaseEventsProcess[SCTArgusEvent, None], threading.Thread
     def run(self) -> None:
         self.enabled.wait()
 
-        with verbose_suppress("ArgusEventPostman failed while consuming inbound events"):
-            for event in self.inbound_events():  # events from ArgusAggregator
-                self._submit_event(event)
+        for event in self.inbound_events():  # events from ArgusAggregator
+            self._submit_event(event)
 
     def _submit_event(self, event: SCTArgusEvent) -> None:
         """Post a single event to Argus, swallowing and logging any failure (see verbose_suppress)."""
--- a/unit_tests/unit/test_argus_postman_stop.py
+++ b/unit_tests/unit/test_argus_postman_stop.py
@@ -94,7 +94,10 @@ def test_stop_is_bounded_when_submit_hangs(argus_pipeline):
     elapsed = time.monotonic() - start
 
     try:
-        assert elapsed < 3
+        # Generous margin over the timeout=1 above: real thread scheduling under a loaded,
+        # parallelized CI run can add noticeable jitter; this only needs to catch stop()
+        # ignoring the timeout altogether, not verify it precisely.
+        assert elapsed < 10
         postman._argus_client.submit_event.assert_called_once()  # drain was attempted
         assert postman.is_alive()  # the hung submit is still stuck past the timeout
     finally:
--- a/unit_tests/unit/test_argus_postman_stop.py
+++ b/unit_tests/unit/test_argus_postman_stop.py
@@ -1,0 +1,114 @@
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU Affero General Public License as published by
+# the Free Software Foundation; either version 3 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
+#
+# See LICENSE for more details.
+#
+# Copyright (c) 2025 ScyllaDB
+
+"""Tests for ArgusEventPostman draining its queue on stop() and staying time-bounded."""
+
+import time
+import shutil
+import tempfile
+import threading
+from unittest.mock import MagicMock
+
+import pytest
+
+from sdcm.sct_events import Severity
+from sdcm.sct_events.setup import EVENTS_DEVICE_START_DELAY, stop_events_device
+from sdcm.sct_events.argus import SCTArgusEvent, get_argus_postman, start_argus_pipeline
+from sdcm.sct_events.events_device import start_events_main_device
+from sdcm.sct_events.events_processes import EVENTS_ARGUS_AGGREGATOR_ID, EventsProcessesRegistry, get_events_process
+
+
+def _argus_event(severity: Severity, message: str) -> SCTArgusEvent:
+    """Build a minimal Argus event payload as produced by ArgusEventCollector."""
+    return SCTArgusEvent(
+        {
+            "run_id": "test-run",
+            "severity": severity.name,
+            "ts": time.time(),
+            "event_type": "TestEvent",
+            "message": message,
+        }
+    )
+
+
+@pytest.fixture
+def argus_pipeline():
+    """Start an isolated Argus pipeline (collector/aggregator/postman) on its own registry.
+
+    The postman is intentionally left not-enabled, so its run() loop parks and only the
+    stop()/drain path touches the aggregator's outbound_queue - making the tests deterministic.
+    """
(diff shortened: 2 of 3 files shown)


fruch:
@claude review

fruch (dismissed):

pehala:
> `ArgusEventPostman.terminate()` only set `stop_event`, which halts the aggregator's `outbound_events` generator (`while not stop_event.is_set()`). Any events still sitting in the aggregator's `outbound_queue` at teardown were therefore **abandoned instead of posted** — a teardown-time `ERROR` event could silently vanish from Argus.

Why not add post step for the generator, so after the loop it sends the requests?

soyacz:
@pehala good call, done — pushed. Added the drain as a post-loop step directly in `EventsProcessPipe.outbound_events()`: once `stop_event` is set, it keeps yielding whatever's left via `get_nowait()` until the queue is empty, instead of a separate drainer thread reaching into the aggregator's internals. The postman's normal `run()` loop then flushes the backlog through the standard per-event path, and `ArgusEventPostman` goes back to only overriding `terminate()` — `stop()` uses the base class's `terminate()+join(timeout)` like every other events process. Bonus: this fixes the same drop-on-shutdown gap in the Grafana pipeline for free.

unit_tests/unit/test_argus_postman_stop.py
```diff
@@ -0,0 +1,114 @@
...
+    time.sleep(EVENTS_DEVICE_START_DELAY)
+    start_argus_pipeline(_registry=registry)
+    postman = get_argus_postman(_registry=registry)
+    aggregator = get_events_process(EVENTS_ARGUS_AGGREGATOR_ID, _registry=registry)
+    postman._argus_client = MagicMock()
+    try:
+        yield postman, aggregator
+    finally:
+        stop_events_device(_registry=registry)
+        shutil.rmtree(temp_dir, ignore_errors=True)
```
pehala: we have events setup in tests already, no need to reinvent the wheel, as a bonus, it is in memory only for unit tests so no need for any cleanup.

unit_tests/unit/test_argus_postman_stop.py
```diff
@@ -0,0 +1,114 @@
...
+
+    aggregator.outbound_queue.put(_argus_event(Severity.ERROR, "err"))
+
+    start = time.monotonic()
+    postman.stop(timeout=1)
+    elapsed = time.monotonic() - start
+
+    try:
+        assert elapsed < 3
+        assert postman._argus_client.submit_event.called  # drain was attempted
```
pehala: We sahould know exactly what submit even was called and we should assert it with called_once

unit_tests/unit/test_argus_postman_stop.py
```diff
@@ -0,0 +1,114 @@
...
+    postman._argus_client.submit_event.side_effect = lambda event: release.wait()
+
+    aggregator.outbound_queue.put(_argus_event(Severity.ERROR, "err"))
+
+    start = time.monotonic()
+    postman.stop(timeout=1)
+    elapsed = time.monotonic() - start
+
+    try:
+        assert elapsed < 3
```
pehala: What is the reason for checking time in these tests? It seems unnecessary to me, if it gets stuck we will never get here regardless

unit_tests/unit/test_argus_postman_stop.py
```diff
@@ -0,0 +1,114 @@
...
+    postman._argus_client.submit_event.side_effect = lambda event: submitted.append(event["severity"])
+
+    aggregator.outbound_queue.put(_argus_event(Severity.WARNING, "warn"))
+    aggregator.outbound_queue.put(_argus_event(Severity.ERROR, "err"))
+    aggregator.outbound_queue.put(_argus_event(Severity.NORMAL, "info"))
+
+    postman.stop(timeout=5)
+
+    assert not postman.is_alive()
+    assert submitted == [Severity.WARNING.name, Severity.ERROR.name, Severity.NORMAL.name]
```
pehala: Why not assert entire events instead?

unit_tests/unit/test_argus_postman_stop.py
```diff
@@ -0,0 +1,114 @@
...
+import pytest
+
+from sdcm.sct_events import Severity
+from sdcm.sct_events.setup import EVENTS_DEVICE_START_DELAY, stop_events_device
+from sdcm.sct_events.argus import SCTArgusEvent, get_argus_postman, start_argus_pipeline
+from sdcm.sct_events.events_device import start_events_main_device
+from sdcm.sct_events.events_processes import EVENTS_ARGUS_AGGREGATOR_ID, EventsProcessesRegistry, get_events_process
+
+
+def _argus_event(severity: Severity, message: str) -> SCTArgusEvent:
```
pehala: nit: I would drop protected naming here, its only effect is that is hurts readability

sdcm/sct_events/argus.py
```diff
@@ -34,6 +34,7 @@
 
 
 ARGUS_EVENT_AGGREGATOR_TIME_WINDOW: float = 90  # seconds
+
```
pehala: nit: unnecessary formatting change

sdcm/sct_events/argus.py
```diff
@@ -109,14 +110,20 @@ def __init__(self, _registry: EventsProcessesRegistry):
...
-                    self._argus_client.submit_event(event)
+        with verbose_suppress("ArgusEventPostman failed while consuming inbound events"):
+            for event in self.inbound_events():  # events from ArgusAggregator
+                self._submit_event(event)
+
+    def _submit_event(self, event: SCTArgusEvent) -> None:
+        """Post a single event to Argus, swallowing and logging any failure (see verbose_suppress)."""
+        if not self._argus_client:
+            return
+        with verbose_suppress(
```
pehala, line 119: we have two nested verbose_supress here, chaining with one at line 113 in run

pehala (commented):
I like the simplified change, it looks much better

unit_tests/unit/test_argus_postman_stop.py
soyacz: Fixed

unit_tests/unit/test_argus_postman_stop.py
soyacz: Fixed

unit_tests/unit/test_argus_postman_stop.py
soyacz: Fixed

sdcm/sct_events/argus.py
soyacz: Fixed

sdcm/sct_events/argus.py
soyacz: Fixed

unit_tests/unit/test_argus_postman_stop.py
soyacz: Partially fixed: dropped `tempfile.mkdtemp()`/`shutil.rmtree` in favor of pytest's `tmp_path` (auto-cleaned), so no manual cleanup remains. Kept the real collector/aggregator/postman pipeline though — the `events`/`events_function_scope` fixture is backed by `FakeEventsDevice`, which calls `publish_event` directly and bypasses the queue/thread machinery entirely, so it can't exercise the actual drain-on-stop() behavior this test is checking.

unit_tests/unit/test_argus_postman_stop.py
soyacz: Not changed: this isn't just checking that `stop()` returns eventually — it's checking that it respects the given timeout. Since `release` is never set, if the timeout were ignored or miscomputed but the thread still returned via some other path, elapsed would reveal a bound that's wrong-but-finite (e.g. multiplied/misapplied timeout). A true infinite hang would indeed just hang the test itself, but that's a different failure mode than what this assertion targets.

unit_tests/unit/test_argus_postman_stop.py
pehala: What about integration level events fixtures we have? Those are actually thread based

soyacz: That's `EventsUtilsMixin` (e.g. `TestGrafana` in `test_sct_events_grafana.py`, or `RealEventsTest`) — it predates AGENTS.md's current unit-test convention ("Use pytest style, NOT class-based tests": no `class Test*`, no `setup_class`/`teardown_class`, no class-method shared setup), which is explicitly marked IMPORTANT and enforced for new tests. Since the mixin's API is classmethod-based, reusing it here would mean reintroducing that pattern for this one file. Our fixture already gets the same real thread/queue behavior as a plain `@pytest.fixture` (with `tmp_path`, per your earlier comment) without the class wrapper.

unit_tests/unit/test_argus_postman_stop.py
soyacz: Reconsidered — you're right that this is a legitimate flakiness risk: the 1s timeout vs 3s bound is only a 2x margin, and this file has real threads with no mocked clock, running under pytest-xdist parallel workers. Widened to 10s; the check now only needs to catch stop() ignoring its timeout altogether, not verify it precisely.

unit_tests/unit/test_argus_postman_stop.py
pehala: There was a fixture I believe that could be used (which just called the mixin), but maybe it is lost as only integration class  based test use real events for now. Anyway problem for refactor later

unit_tests/unit/test_argus_postman_stop.py
pehala: I still think any time measuring here is unnecessary, but lets see how flaky it will be in the future

pehala (approved):

PR #15366, opened 2026-07-09
fix(oci): properly set rack and dc using `add_remove_nemesis_dc` nemesis

vponomaryov:
Running tests on OCI backend we must forcefully update
datacenter and rack info in the `cassandra-rackdc.properties` file.
It is what gets done in the `sdcm/snitch_configuration.py` file.

And it must be done in any place where we create new DB node on OCI.

So, fix the `add_remove_nemesis_dc` (`AddRemoveDcNemesis`) by doing identical changes.

Closes: #SCT-96

### Testing

- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-100gb-4h-oci-test/44
- https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-100gb-4h-test/19

### Reminders

### DIFF
--- a/sdcm/nemesis/monkey/add_remove_dc.py
+++ b/sdcm/nemesis/monkey/add_remove_dc.py
@@ -5,6 +5,7 @@ from sdcm.cluster import BaseNode
 from sdcm.exceptions import KillNemesis, UnsupportedNemesis
 from sdcm.nemesis import NemesisBaseClass
 from sdcm.provision.scylla_yaml import SeedProvider
+from sdcm.snitch_configuration import SnitchConfig
 from sdcm.utils.decorators import retrying, skip_on_capacity_issues
 from sdcm.utils.features import is_tablets_feature_enabled
 from sdcm.utils.replication_strategy_utils import (
@@ -137,13 +138,16 @@ class AddRemoveDcNemesis(NemesisBaseClass):
             ]
 
         endpoint_snitch = self.runner.cluster.params.get("endpoint_snitch") or ""
-        if endpoint_snitch.endswith("GossipingPropertyFileSnitch"):
-            rackdc_value = {"dc": "add_remove_nemesis_dc"}
-        else:
-            rackdc_value = {"dc_suffix": "_nemesis_dc"}
+        dc_suffix = "_nemesis_dc"
+        dc_name = f"add_remove{dc_suffix}"
 
-        with node.remote_cassandra_rackdc_properties() as properties_file:
-            properties_file.update(**rackdc_value)
+        SnitchConfig(node=node, datacenters=[dc_name]).apply(force=True)
+
+        if not endpoint_snitch.endswith("GossipingPropertyFileSnitch"):
+            # Non-GPFS snitches (Ec2Snitch, GoogleCloudSnitch, etc.) derive dc from cloud metadata
+            # and only read dc_suffix from the properties file to form the final dc name.
+            with node.remote_cassandra_rackdc_properties() as properties_file:
+                properties_file["dc_suffix"] = dc_suffix
 
     def add_nodes_in_new_dc(self) -> None:
         """Add and initialize temporary nodes configured to join a new datacenter."""
--- a/sdcm/snitch_configuration.py
+++ b/sdcm/snitch_configuration.py
@@ -43,11 +43,12 @@ class SnitchConfig:
         else:
             return ""
 
-    def apply(self) -> None:
+    def apply(self, force: bool = False) -> None:
         """
         Apply snitch configuration to the node (only in cassandra-rackdc.properties, scylla.yaml must be updated separately.).
 
-        Returns `True` if require Scylla restart to make the effect.
+        Args:
+            force: If True, forcibly overwrite existing properties (used by AddRemoveDc nemesis).
         """
         properties = {
             "dc": self._datacenter,
@@ -59,7 +60,7 @@ class SnitchConfig:
 
         with self._node.remote_cassandra_rackdc_properties() as properties_file:
             for key, value in properties.items():
-                if hasattr(self._node, "query_oci_metadata"):
+                if force or hasattr(self._node, "query_oci_metadata"):
                     # NOTE: OCI images have predefined properties file with real data for dc and rack.
                     #       So, set it forcibly because defaulting will not have an effect.
                     properties_file[key] = value


sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -137,13 +137,23 @@ def configure_new_dc(self, node: BaseNode) -> None:
...
         else:
-            rackdc_value = {"dc_suffix": "_nemesis_dc"}
+            rackdc_value = {"dc_suffix": dc_suffix}
 
         with node.remote_cassandra_rackdc_properties() as properties_file:
-            properties_file.update(**rackdc_value)
+            if hasattr(node, "query_oci_metadata"):
+                # NOTE: OCI images have predefined properties file with real data for dc and rack.
+                #       So, set it forcibly because defaulting will not have an effect.
+                #       Ref: sdcm/snitch_configuration.py
```
pehala: Why are OCI images different from other clouds?

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -137,13 +137,23 @@ def configure_new_dc(self, node: BaseNode) -> None:
...
         if endpoint_snitch.endswith("GossipingPropertyFileSnitch"):
-            rackdc_value = {"dc": "add_remove_nemesis_dc"}
+            rackdc_value = {"dc": dc_name}
         else:
-            rackdc_value = {"dc_suffix": "_nemesis_dc"}
+            rackdc_value = {"dc_suffix": dc_suffix}
 
         with node.remote_cassandra_rackdc_properties() as properties_file:
-            properties_file.update(**rackdc_value)
+            if hasattr(node, "query_oci_metadata"):
```
pehala: I would prefer a check for node being on OCI instead of hasattr, I hope we can do it a way that wont use isinstance but I am ok with that as as well.

By checking for OCI directly we are saying that every OCI node needs to use this. `query_oci_metadata` is internal value which should not be used in comparison, if we ever changed internals or rename this variable this would silently stop working.

sdcm/nemesis/monkey/add_remove_dc.py
vponomaryov: `AWS` uses `Ec2Snitch/Ec2MultiRegionSnitch`.
`GCE` uses `GoogleCloudSnitch`.
`Azure` uses `AzureSnitch`.
`OCI`? There is no `OciSnitch`...

So, `OCI` uses non-backend specific `GossipingPropertyFileSnitch` and it reads directly from the `cassandra-rackdc.properties` file on disk.
Obviously, for production needs there must be automatic info for DC and rack.

Then, look at the `sdcm/snitch_configuration.py` as comment says above.
You will see there that SCT has special behavior by setting "defaults" only if value is absent.
In case of OCI, due to the usage of the `GossipingPropertyFileSnitch`, it is always set and default always don't get applied.
So, we must set SCT-specific DC and rack forcibly all the time for OCI DB nodes.

fruch: cause it's using GossipingPropertyFileSnitch by default, OCI don't have it's own snitch as the others.

I've asked for it and it was declined
https://github.com/scylladb/scylladb/issues/27901

sdcm/nemesis/monkey/add_remove_dc.py
vponomaryov: It is identical approach to the logic in the `sdcm/snitch_configuration.py` file.

I will keep it as-is.

You may create an issue/task to update/refactor both later.

This PR fixes the nemesis.

sdcm/nemesis/monkey/add_remove_dc.py
pehala: Thanks for explanation

sdcm/nemesis/monkey/add_remove_dc.py
```diff
@@ -137,13 +137,23 @@ def configure_new_dc(self, node: BaseNode) -> None:
             ]
 
         endpoint_snitch = self.runner.cluster.params.get("endpoint_snitch") or ""
+        dc_suffix = "_nemesis_dc"
+        dc_name = f"add_remove{dc_suffix}"
         if endpoint_snitch.endswith("GossipingPropertyFileSnitch"):
```
fruch: shouldn't we default this for OCI ? and that would be enough ?

sdcm/nemesis/monkey/add_remove_dc.py
pehala: Fact we did it incorrectly elsewhere in completely different part of the code is not a good argument for not doing it in a better way here. They are separate parts of the code.

My proposal is trivial to apply so I dont see reason why not include it here.

pehala: Besides, this nemesis is already migrated and was recently refactored so I would like to not introduce flawed code here

vponomaryov: Because this PR is about fixing the nemesis, not debatting about the refactor of an existent approach.

The `query_oci_metadata` attr is OCI specific, other backends will not have it.
It is defined in SCT, not some dependency lib, so, it must stay all the time.
If one removes any attr (s)he must update all the places with its usages - a must in professional software dev.

So, that's why I consider the current approach as `totally ok` to be part of the `fix` for the nemesis.

sdcm/nemesis/monkey/add_remove_dc.py
vponomaryov: I don't understand the question.

This particular part of the `code changes` only removes code unsafety (`multiple identical literals`) and adds 2 reusable vars.

sdcm/nemesis/monkey/add_remove_dc.py
fruch: @vponomaryov can't we use SnitchConfig here ? 

or better optionally pass in the snitch config we need via add_node explicitly ? (so we won't need this trick of assuming what we can or can't override ?)

sdcm/nemesis/monkey/add_remove_dc.py
pehala: > Because this PR is about fixing the nemesis, not debatting about the refactor of an existent approach.

I do not require changes to other parts of code that use it, I don't want to introduce bad approach into components that we maintain, where it wasn't before.

> The `query_oci_metadata` attr is OCI specific, other backends will not have it.
> It is defined in SCT, not some dependency lib.
> So, it must stay all the time.
> If one removes any attr (s)he must update all the places with its usages.

Yes, they should. That doesn't change the fact it can be missed.
It is error prone, it is harder to read/understand than direct OCI comparison and it is wrong on conceptual level, because what you actually want to guard is if the node/cluster is on OCI not that it contain the variable as you dont use it in the if branch, so by definition you are checking wrong thing, even if it effectively equals OCI check.

Is it a big issue? no, but it is completely unnecessary and better solution is instant to implement

Edit: Also suggesting a one line PR "refactor" followup is much greater overhead than doing it here

vponomaryov: `It is error prone,`
The thing we fully control in our code base is `not` error prone.
So, false statement.

`Yes, they should. That doesn't change the fact it can be missed.`
It is general possibility to break any code, it is applicable to everything.
Not an argument.

`so by definition you are checking wrong thing, even if it effectively equals OCI check.`

Which definition? Definition of what?
There are multiple ways to solve some task.
The current one is 100% working one, because it depends on the SCT-controlled attr.

So, whole this discussion is not about some possible instability fix, it is about `I want another approach for solving it`.

`I do not require changes to other parts of code that use it,`

So, you propose to have 2 completely different solutions for the same thing in the code base - exactly such a thing is error prone.

I agree on a separate task for update of both places at once keeping a single solution in each point of time.

sdcm/nemesis/monkey/add_remove_dc.py
vponomaryov: @fruch 
Yes, I think it should be ok to use `SnitchConfig` class here directly.

sdcm/nemesis/monkey/add_remove_dc.py
vponomaryov: Reusing the SnitchConfig.
This discussion is not actual anymore.

sdcm/nemesis/monkey/add_remove_dc.py
vponomaryov: Applied the SnitchConfig in the nemesis.

vponomaryov:
Current state of the code requires retesting with 2 different backends (OCI and another one).

sdcm/nemesis/monkey/add_remove_dc.py
pehala: I wont continue this argument anymore, it is clear we wont reach consensus and arguing over this here is waste of time for both of us. If you want to continue discussion on this topic reach out to me offline, I will gladly continue it there.

I require single line change to this PR to make it approvable by me. Reason why I care about the oneline is not necessarily the one line itself, but the fact it is completely pointless to introduce it here. It is not better in any way, and we only used it once in completely unrelated part of the code.

It is a codebase and nemesis which my team takes ownership (marked nowhere officially for now) and changed recently (due to RF changes).

I will approve on one of three scenarios:
1) Fix is done here - quick, we can merge almost immediately
2) Fix is done by you in a followup (can be for both usages but I dont care about that part of the code) - quick, we can merge almost immediately
3) We continue this discussion offline and you convince me this solution is good enough - Continue academic discussion.

If you dont care about my approve, then just resolve and ignore this comment

pehala: Well, it was solved by a better solution which doesnt require this line at all :) Thats also good, maybe even better

sdcm/nemesis/monkey/add_remove_dc.py
fruch: using SnitchConfig make sense, I think the implementation within it, can be improved to remove the duality about different snitches, but that for different PR

sdcm/nemesis/monkey/add_remove_dc.py
fruch: I've meant that endpoint_snitch == GossipingPropertyFileSnitch should have been enough for the logic needed (and maybe defaulting it for OCI), and not base it on some specific property of OCI node.

but now that's more or less what the PR is doing

fruch (dismissed):

fruch:
> Current state of the code requires retesting with 2 different backends (OCI and another one).

I see job OCI mentioned, but not the other...

vponomaryov:
Requires update, AWS failed on this nemesis with the current state of the PR

vponomaryov:
> > Current state of the code requires retesting with 2 different backends (OCI and another one).
> 
> I see job OCI mentioned, but not the other...

Updated links in the PR description.
Now both refer to the running CI jobs with the latest PR state

pehala (approved):

vponomaryov:
> ### Testing
> * https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-100gb-4h-oci-test/44
> * https://jenkins.scylladb.com/job/scylla-staging/job/valerii/job/vp-longevity-100gb-4h-test/19

Both passed.

It is ready for merge

PR #15367, opened 2026-07-09
feat: Change wait_not_tablet_migration to wait_tablet_balanced

pehala:
* quiesce_topology endpoint which is used has changed the behaviour since this was introduced. It now also waits for tablet balancing and doesn't need to be called on all nodes
* Remove 60sec wait
  * Confirmed to not be needed 

### Testing

- [x] Longevity - https://argus.scylladb.com/tests/scylla-cluster-tests/164ed230-55ab-46fd-9879-25a6d94b1a8c/sct-events

### Reminders

### DIFF
--- a/sdcm/utils/tablets/common.py
+++ b/sdcm/utils/tablets/common.py
@@ -1,5 +1,4 @@
 import logging
-import time
 from dataclasses import dataclass
 from typing import Optional
 
@@ -26,31 +25,23 @@ class TabletsConfiguration:
         return "{" + ", ".join(items) + "}"
 
 
-def wait_no_tablets_migration_running(node, timeout: int = 3600):
+def wait_tablets_balanced(node, timeout: int = 3600):
     """
-    Waiting for having no ongoing tablets topology operations using REST API.
-    !!! It does not guarantee that tablets are balanced !!!
-    Keep in mind that it is good only to find when ongoing tablets topology operations is done.
-    Very next second another topology operation can be started.
-
-    doing it several times as there's a risk of:
-    "currently a small time window after adding nodes and before load balancing starts during which
-    topology may appear as quiesced because the state machine goes through an idle state before it enters load balancing state"
-    """
+    Wait for tablets to be balanced, no more pending splits/merges and no ongoing tablets topology operations using REST API.
+    A single request is enough as it is submitted as a global topology request and completes only after fresh tablet load stats produce an empty balance plan and tablets are idle."""
     if not is_tablets_feature_enabled(node):
         LOGGER.info("Tablets are disabled, skipping wait for balance")
         return
-    time.sleep(60)  # one minute gap before checking, just to give some time to the state machine
     client = RemoteCurlClient(host="127.0.0.1:10000", endpoint="", node=node)
-    LOGGER.info("Waiting for having no ongoing tablets topology operations")
+    LOGGER.info("Waiting for tablets balancing (no pending splits/merges, no ongoing topology operations)")
     try:
         with adaptive_timeout(Operations.TABLET_MIGRATION, node, timeout=timeout) as adaptive_timeout_value:
             client.run_remoter_curl(
                 method="POST", path="storage_service/quiesce_topology", params={}, timeout=adaptive_timeout_value
             )
-        LOGGER.info("All ongoing tablets topology operations are done")
+        LOGGER.info("Tablets are balanced")
     except Exception as exc:  # noqa: BLE001
         InfoEvent(
-            f"Failed to wait for having no ongoing tablets topology operations. Exception: {exc.__repr__()}",
+            f"Failed to wait for tablets to be balanced. Exception: {exc.__repr__()}",
             severity=Severity.ERROR,
         ).publish()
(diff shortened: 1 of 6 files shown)


pehala:
@claude review

pehala:
v2:
   * Addressed Claude nits: Changed log messages

paszkow (commented):
The code change looks consistent with the enhanced `storage_service/quiesce_topology` behavior. The only part I would still polish is the helper wording: the current docstring says the node gathers information from all nodes, but the important API contract is that a single request is coordinated through topology and, when the enhanced feature is enabled cluster-wide, it waits until there is no pending tablet balance plan and tablets are idle.

Suggested wording:

```python
"""
Wait for tablet topology to quiesce via `storage_service/quiesce_topology`.

With `QUIESCE_TOPOLOGY_ENHANCED` enabled cluster-wide, a single request is enough: it is submitted as a global topology request and completes only after fresh tablet load stats produce an empty balance plan and tablets are idle. In mixed-version clusters Scylla falls back to the old "topology not busy" behavior, so callers that rely on full balancing should only use this on versions with the enhanced endpoint.
"""
```

I would also align the success log with the state being waited for, e.g. `Tablets are balanced`, rather than `Tablets finished balancing`.

sdcm/utils/tablets/common.py
```diff
@@ -26,31 +25,24 @@ def __str__(self):
...
-    Waiting for having no ongoing tablets topology operations using REST API.
-    !!! It does not guarantee that tablets are balanced !!!
-    Keep in mind that it is good only to find when ongoing tablets topology operations is done.
-    Very next second another topology operation can be started.
-
-    doing it several times as there's a risk of:
-    "currently a small time window after adding nodes and before load balancing starts during which
-    topology may appear as quiesced because the state machine goes through an idle state before it enters load balancing state"
+    Waiting for tablets being balanced, no more pending splits/merges and no ongoing tablets topology operations using REST API.
+    This should be called on a single node, as the node will then be a coordinator for this request and gather the information from all nodes in the cluster.
```
paszkow: I would change the second line to:

A single request is enough as it is submitted as a global topology request and completes only after fresh tablet load stats produce an empty balance plan and tablets are idle.

sdcm/utils/tablets/common.py
```diff
@@ -26,31 +25,24 @@ def __str__(self):
...
     """
-    Waiting for having no ongoing tablets topology operations using REST API.
-    !!! It does not guarantee that tablets are balanced !!!
-    Keep in mind that it is good only to find when ongoing tablets topology operations is done.
-    Very next second another topology operation can be started.
-
-    doing it several times as there's a risk of:
-    "currently a small time window after adding nodes and before load balancing starts during which
-    topology may appear as quiesced because the state machine goes through an idle state before it enters load balancing state"
+    Waiting for tablets being balanced, no more pending splits/merges and no ongoing tablets topology operations using REST API.
```
paszkow: Wait for tablets to be balanced, ...

paszkow (commented):
A few typos in the commit message:

- quience_topology -> quiesce_topology
- neither of the functions has `_for_` in its name

Similarly with PR title and cover letter.

pehala:
> With `QUIESCE_TOPOLOGY_ENHANCED` enabled cluster-wide, a single request is enough: it is submitted as a global topology request and completes only after fresh tablet load stats produce an empty balance plan and tablets are idle. In mixed-version clusters Scylla falls back to the old "topology not busy" behavior, so callers that rely on full balancing should only use this on versions with the enhanced endpoint.

Is QUIESCE_TOPOLOGY_ENHANCED enabled by default? If so, since when?

paszkow:
> > With `QUIESCE_TOPOLOGY_ENHANCED` enabled cluster-wide, a single request is enough: it is submitted as a global topology request and completes only after fresh tablet load stats produce an empty balance plan and tablets are idle. In mixed-version clusters Scylla falls back to the old "topology not busy" behavior, so callers that rely on full balancing should only use this on versions with the enhanced endpoint.
> 
> Is QUIESCE_TOPOLOGY_ENHANCED enabled by default? If so, since when?

It was a copy-paste. I guess you can drop this `With `QUIESCE_TOPOLOGY_ENHANCED` enabled cluster-wide,`. It's a feature flag provided as part of the PR to handle rolling upgrade.

paszkow (approved):

fruch:
> > > With `QUIESCE_TOPOLOGY_ENHANCED` enabled cluster-wide, a single request is enough: it is submitted as a global topology request and completes only after fresh tablet load stats produce an empty balance plan and tablets are idle. In mixed-version clusters Scylla falls back to the old "topology not busy" behavior, so callers that rely on full balancing should only use this on versions with the enhanced endpoint.
> > 
> > 
> > Is QUIESCE_TOPOLOGY_ENHANCED enabled by default? If so, since when?
> 
> It was a copy-paste. I guess you can drop this `With `QUIESCE_TOPOLOGY_ENHANCED` enabled cluster-wide,`. It's a feature flag provided as part of the PR to handle rolling upgrade.

still, since which scylla version is this new behavior ?, so we'll know where to backport this change.

and if it's save enough to take it to the performance branch, that would run things up to 2025.1 brach

cezarmoise (approved):
LGTM

pehala:
> still, since which scylla version is this new behavior ?, so we'll know where to backport this change.
> 
> and if it's save enough to take it to the performance branch, that would run things up to 2025.1 brach

It is from 2026.3

fruch (approved):

PR #15368, opened 2026-07-09
docs(plans): add strict jinja stress templating mini-plan

pehala:
Longevity test configs hardcode row counts in stress commands, so dataset size must be recalculated manually whenever disk size, instance type, or expected compression changes. Add runtime-only strict Jinja rendering for longevity stress commands so configs can calculate row counts from available `/var/lib/scylla` capacity through an `effective_disk_size_bytes` template variable.

This will enable us to create tests with e.g. 40% of disk utilization regardless of instance and cloud.

### Testing

- [ ]

### Reminders

### DIFF
--- a/docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
+++ b/docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
@@ -6,58 +6,84 @@
 
 ## Problem
 
-Longevity test configs hardcode row counts in stress commands, so dataset size must be recalculated manually whenever disk size, instance type, or expected compression changes. Add runtime-only strict Jinja rendering for longevity stress commands so configs can calculate row counts from available `/var/lib/scylla` capacity through an `effective_disk_size_bytes` template variable.
+Longevity test configs hardcode row counts in stress commands, so dataset size must be recalculated manually whenever disk size, instance type, or expected compression changes. The implementation now supports runtime-only strict Jinja rendering for stress commands so configs can calculate row counts from available `/var/lib/scylla` capacity through an `effective_disk_size_bytes` template variable, while keeping config loading and validation unaware of Jinja syntax.
 
 ## Approach
 
-- Add `effective_compression_ratio` to SCT config with default `1.0` and validation `0 < value <= 1`. Treat it as `on_disk_bytes / logical_uncompressed_bytes`, so `0.68` means generated logical data is expected to occupy 68% of its uncompressed size on disk.
-- Add `effective_compression_ratio: 1.0` to defaults and regenerate config docs.
-- Do not render or validate Jinja templates during config load, `verify_configuration()`, or `check_required_files()`. Existing config validation should keep seeing the original literal strings.
-- Render only at stress execution time in `LoaderUtilsMixin`, immediately before `_run_all_stress_cmds()` inspects command prefixes or dispatches to `run_stress_thread()`.
-- Use a strict Jinja environment with `StrictUndefined`; fail fast before starting stress if a command references an unknown variable or has invalid syntax.
-- Render only commands containing Jinja delimiters: `{{`, `{%`, or `{#`. Non-templated commands must remain byte-for-byte unchanged.
-- Add a cached property for `effective_disk_size_bytes` because capacity is expected to be stable for this purpose.
-- Calculate raw available capacity from DB nodes with `df -B1` against `/var/lib/scylla`, use the average available bytes across nodes because Scylla load-balances data across nodes, then export `floor(average_available_bytes / effective_compression_ratio)`.
-- Keep RF and node count outside the exported variable. Test configs should account for RF and node count directly in the Jinja formula.
-- Document a compact example that hardcodes formula constants in the command:
+- Added `effective_compression_ratio` to SCT config with default `1.0` and validation `0 < value <= 1`. It is defined as `on_disk_bytes / logical_uncompressed_bytes`, so `0.68` means the logical dataset is expected to occupy about 68% of its uncompressed size on disk.
+- Added `effective_compression_ratio: 1.0` to defaults and updated generated config docs.
+- Added `stress_template_context` to SCT config so tests can define shared derived Jinja variables once and reuse them across multiple stress commands.
+- Kept config load, `verify_configuration()`, and `check_required_files()` free of Jinja rendering. Existing validation still sees the original literal stress command strings.
+- Implemented runtime rendering in `LoaderUtilsMixin` by rendering every stress command immediately before `_run_all_stress_cmds()` performs command-specific handling and dispatches to `run_stress_thread()`.
+- Implemented a strict Jinja environment with `StrictUndefined` in `LoaderUtilsMixin.jinja_env`, plus `MixinLazyContext` so template variables are resolved lazily from the mixin owner object only when referenced.
+- Added a `stress_template_whitelist` hook to `LoaderUtilsMixin` and overrode it in `LongevityTest`, so all longevity-derived tests automatically restrict owner-backed template variables to `effective_disk_size_bytes` and `db_node_count_per_dc`.
+- Implemented ordered `stress_template_context` evaluation in `LoaderUtilsMixin`, with YAML scalar coercion and conflict detection for context keys that shadow built-in owner-backed variables.
+- Failed fast on unknown variables by converting `UndefinedError` into a `RuntimeError` that includes the original command. Syntax errors still propagate as Jinja `TemplateSyntaxError` before any stress thread starts.
+- Did not add a delimiter pre-scan optimization. Instead, all stress commands go through `render_stress_cmd()`, and literal commands remain unchanged because rendering a template with no Jinja syntax returns the original string.
+- Kept `effective_disk_size_bytes` as a cached property on `LongevityTest`, not on `LoaderUtilsMixin`. The exported value is still `floor(average_available_bytes / effective_compression_ratio)`.
+- Calculated raw available capacity from DB nodes with `df -B1 /var/lib/scylla`, using the average available bytes across nodes because Scylla load-balances data across nodes.
+- Kept RF outside the exported variables. Test configs can use `db_node_count_per_dc` directly in the Jinja formula when all DCs are expected to have the same DB node count.
+- The implementation also normalizes rendered multi-line YAML block scalars by collapsing whitespace to a single line before dispatch.
+- Added user-facing documentation in `docs/loader_usage.md` with a compact example, runtime behavior notes, the longevity-only whitelist rule, and the shared `stress_template_context` mechanism.
+
+- Example formula supported by the implementation:
 
 ```yaml
 effective_compression_ratio: 0.68
 
+stress_template_context:
+  total_rows: "{{ ((effective_disk_size_bytes * db_node_count_per_dc) // 3 // (200 * 5)) | int }}"
+  rows_per_cmd: "{{ total_rows // 4 }}"
+
 prepare_write_cmd:
   - >-
-    {% set rows = ((effective_disk_size_bytes * 3) // 3 // (200 * 5)) | int %}
-    cassandra-stress write cl=QUORUM n={{ rows }}
+    cassandra-stress write cl=QUORUM n={{ rows_per_cmd }}
     -schema 'replication(strategy=NetworkTopologyStrategy,replication_factor=3)'
     -mode cql3 native
     -rate threads=150
     -col 'size=FIXED(200) n=FIXED(5)'
-    -pop seq=1..{{ rows }}
+    -pop seq=1..{{ rows_per_cmd }}
 ```
 
-- Existing delimiter scan found no Jinja delimiters in relevant SCT test/config YAML directories: `test-cases/`, `configurations/`, `defaults/`, and `unit_tests/test_configs/`. Existing repository YAML delimiter usage is unrelated GitHub Actions `${{ ... }}` and Helm/minio templates under `sdcm/k8s_configs/minio/templates/`, which are unaffected because only stress command values are rendered.
+- Existing repository YAML delimiter usage outside stress command values remains unaffected because rendering happens only when stress commands are executed.
 
 ## Files to Modify
 
-- `sdcm/sct_config.py` -- Add `effective_compression_ratio` field, type constraints, and config help text.
-- `defaults/test_default.yaml` -- Add default `effective_compression_ratio: 1.0`.
-- `sdcm/utils/loader_utils.py` -- Add cached effective disk capacity calculation and strict runtime Jinja rendering before stress command dispatch.
-- `docs/configuration_options.md` -- Regenerate config documentation after adding the new option.
-- `docs/sct-configuration.md` -- Add a short usage note and example for templated longevity stress commands.
-- `unit_tests/unit/test_sct_config_validators.py` -- Add validation tests for accepted and rejected `effective_compression_ratio` values.
-- `unit_tests/unit/test_longevity.py` -- Add runtime rendering tests for successful formulas, strict undefined failures, syntax failures, unchanged literal commands, and cached effective disk size behavior.
+- `sdcm/sct_config.py` -- Added `effective_compression_ratio` field, type constraints, and config help text.
+- `sdcm/sct_config.py` -- Added `stress_template_context` field for shared runtime stress command variables.
+- `defaults/test_default.yaml` -- Added default `effective_compression_ratio: 1.0`.
+- `defaults/test_default.yaml` -- Added default `stress_template_context: {}`.
+- `longevity_test.py` -- Added cached `effective_disk_size_bytes`, scalar `db_node_count_per_dc`, and longevity template whitelist entries on `LongevityTest`.
+- `sdcm/utils/loader_utils.py` -- Added `MixinLazyContext`, strict Jinja environment setup, shared context evaluation, runtime stress command rendering before stress dispatch, and the whitelist hook used by longevity tests.
+- `docs/configuration_options.md` -- Updated generated config documentation for the new option.
+- `docs/loader_usage.md` -- Added feature documentation and an example for templated longevity stress commands.
+- `unit_tests/unit/test_sct_config_validators.py` -- Added validation tests for accepted and rejected `effective_compression_ratio` values.
+- `unit_tests/unit/test_longevity.py` -- Added tests for `effective_disk_size_bytes` calculation, scalar `db_node_count_per_dc` behavior, longevity whitelist enforcement, and context usage from whitelisted built-ins.
+- `unit_tests/unit/test_render_stress_cmd.py` -- Added renderer tests for literal commands, successful templates, shared context evaluation, unknown variables, syntax errors, multiline command normalization, and unrestricted generic mixin behavior.
+- `unit_tests/unit/test_mixin_lazy_context.py` -- Added focused tests for lazy owner attribute resolution and whitelist behavior in the custom Jinja context.
+- `test-cases/features/elasticity/elasticity-90-percent-with-nemesis.yaml` -- Migrated the Tier 1 90%% elasticity testcase to shared `stress_template_context` variables.
 
 ## Verification
 
-- [ ] `effective_compression_ratio=1.0` exports average raw available bytes unchanged.
-- [ ] `effective_compression_ratio=0.68` exports `floor(average_available_bytes / 0.68)`.
-- [ ] Effective disk size calculation averages available capacity across DB nodes instead of using the lowest node value.
-- [ ] Ratios `0`, negative values, and values above `1` fail config validation.
-- [ ] `sct.py conf` succeeds for a config containing Jinja in stress commands without requiring DB nodes or disk capacity.
-- [ ] A command without Jinja delimiters is passed to stress dispatch unchanged.
-- [ ] A command with `{{ effective_disk_size_bytes }}` renders to a concrete command before stress dispatch.
-- [ ] A command with an unknown template variable fails before any stress thread starts.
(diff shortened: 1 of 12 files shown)


pehala:
@fruch This is the idea I had for the disk agnostic stress commands, it should be quite simple and backwards compatible change, only a few configs can use it. Please let me know what you think, before I start implementing

@scylladb/storage-qa This might be interesting to you as well

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
```diff
@@ -0,0 +1,63 @@
+# Mini-Plan: Strict Jinja Stress Command Templating
+
+**Date:** 2026-07-09
+**Estimated LOC:** 250
+**Related PR:** none
+
+## Problem
+
+Longevity test configs hardcode row counts in stress commands, so dataset size must be recalculated manually whenever disk size, instance type, or expected compression changes. Add runtime-only strict Jinja rendering for longevity stress commands so configs can calculate row counts from available `/var/lib/scylla` capacity through an `effective_disk_size_bytes` template variable.
```
fruch: I think it can be more general then just disk size, see https://scylladb.atlassian.net/browse/SCT-631

also other places we have stress command templated today, is in SLA tests, putting the SLAs into the command,

and in performance  tests when we have the thrrorle/limit templated into the commands

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
```diff
@@ -0,0 +1,63 @@
...
+
+## Approach
+
+- Add `effective_compression_ratio` to SCT config with default `1.0` and validation `0 < value <= 1`. Treat it as `on_disk_bytes / logical_uncompressed_bytes`, so `0.68` means generated logical data is expected to occupy 68% of its uncompressed size on disk.
```
fruch: I didn't understood where would `effective_compression_ratio` is coming, and how people would know how to set it

pehala: I suspect it wont have to tempered with much. Goal here is to account for compression and given we use mostly similar workloads, one value in defaults will suffice and we will update it only when neccessary

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
pehala: I think this method can be extended in the future, I think it is more universal and transparent than currently used methods, but it is out of scope to update these for this PR. I would like to keep this one lean and continue building on it with futher followups

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
fruch: Isn't that connect with what actual compression is being used (not that many tests are changing from defaults)

pehala: It is, but it is not the only one. Compression effectivity and thus the final saved storage space depends on the schema and data. In general, we cannot know beforehand how much space we are going to save with compression. Thats why this exists: I can gather single value from existing test and put it into test_defaults and if there is a test where it is significantly off, we can change it on per testcase bases, if needed.

fruch: We need to clearify and write this down, that this feature is dependent on such and such meuserment and how on should do it and calculate it. just say this procees should be documented (and that we shouldn't end up with a magic number that no one knows how to recalculate it, and would need to start guessing, we have enough of those)

pehala: Fair enough, I plan on gathering it from metrics, so will document exactly where there to find it

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
```diff
@@ -0,0 +1,63 @@
...
+    -col 'size=FIXED(200) n=FIXED(5)'
+    -pop seq=1..{{ rows }}
+```
+
+- Existing delimiter scan found no Jinja delimiters in relevant SCT test/config YAML directories: `test-cases/`, `configurations/`, `defaults/`, and `unit_tests/test_configs/`. Existing repository YAML delimiter usage is unrelated GitHub Actions `${{ ... }}` and Helm/minio templates under `sdcm/k8s_configs/minio/templates/`, which are unaffected because only stress command values are rendered.
+
+## Files to Modify
+
+- `sdcm/sct_config.py` -- Add `effective_compression_ratio` field, type constraints, and config help text.
+- `defaults/test_default.yaml` -- Add default `effective_compression_ratio: 1.0`.
```
pehala: this should be changed for an actual compression_ratio based on metrics from recent Tier 1/Sanity runs.

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
yarongilor: Not sure i follow this parameter - is it meant to set a specific ratio value for a stress compression? how could it be done/supported? there are fixes load patterns and fixed compression methods?

yarongilor:
Not sure the `effective_disk_size_bytes` can be isolated from multiple other factors dependency.
The stress tools have different patterns with a different compression potential. 
The load could also have deletions that impact the dataset size.
A fixed yaml configuration of MV for the test can also impact and double the used-disk size.
There could be tests configured with multiple db instance types as well.

In i8g family, for example there's also an extended mode of i8ge with a larger disk that changes the ration between disk-size and cpu/memory size. This can make it more complicated and it's also possible the adjusted stress-command will end up too slow , exceeding the prepare timeout threshold.

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
pehala: > Not sure i follow this parameter - is it meant to set a specific ratio value for a stress compression? how could it be done/supported? there are fixes load patterns and fixed compression methods?

First this parameters doesnt need to precise, I suspect 0.6 vs 0.7 wont make major difference and at worst you would end up in 35-45%, instead of 40%, not that big deal IMO. 

Second we use very similar workloads and generally dont change compression

In conclusion, this parameter will be there, but should be used rarily. With compression we CANNOT predict it based on other parameters, so this parameter only serves as a knob one can turn in case the default value is not good enough (again should be rare, but it is better to have the knob than just to make this a constant)

pehala:
> Not sure the `effective_disk_size_bytes` can be isolated from multiple other factors dependency.
> The stress tools have different patterns with a different compression potential.

I dont think that is the case, majority of test tools are CS-like with very simple schemas. (We will see based on the metrics)

> The load could also have deletions that impact the dataset size.
> A fixed yaml configuration of MV for the test can also impact and double the used-disk size.
> There could be tests configured with multiple db instance types as well.
>
> In i8g family, for example there's also an extended mode of i8ge with a larger disk that changes the ration between disk-size and cpu/memory size. This can make it more complicated and it's also possible the adjusted stress-command will end up too slow , exceeding the prepare timeout threshold.

This was never meant to be perfect or ideal. It is the first step. Currently we have absolutely nothing in this regard, now we have something, test which has setup 40% and ends up with 50% is better than test with fixated instance type.

i8g vs i8ge wont be ever solved by `effective_disk_size_bytes`, but if we will require interchangeability between vastly different cpu/disk ratio we can variable based on CPU which you can use for the throttle. But that is way out of scope of this plan. For now we can simply limit the disk size in the sizing to make sure we have instances in similar category.

It is a first feature in a series of improvements (most likely) which we can develop (quickly) and utilize to make things better, it is not the penultimate thing that will solve all of our problems. It is also extendible, so we can add more variables etc as needed

Will the single knob be enough? we shall see, we can always change it or add more knobs if needed, but I dont want to prematurely add complexity.

fruch:
> > Not sure the `effective_disk_size_bytes` can be isolated from multiple other factors dependency.
> > The stress tools have different patterns with a different compression potential.
> 
> I dont think that is the case, majority of test tools are CS-like with very simple schemas. (We will see based on the metrics)
> 
> > The load could also have deletions that impact the dataset size.
> > A fixed yaml configuration of MV for the test can also impact and double the used-disk size.
> > There could be tests configured with multiple db instance types as well.
> > In i8g family, for example there's also an extended mode of i8ge with a larger disk that changes the ration between disk-size and cpu/memory size. This can make it more complicated and it's also possible the adjusted stress-command will end up too slow , exceeding the prepare timeout threshold.
> 
> This was never meant to be perfect or ideal. It is the first step. Currently we have absolutely nothing in this regard, now we have something, test which has setup 40% and ends up with 50% is better than test with fixated instance type.
> 
> i8g vs i8ge wont be ever solved by `effective_disk_size_bytes`, but if we will require interchangeability between vastly different cpu/disk ratio we can variable based on CPU which you can use for the throttle. But that is way out of scope of this plan. For now we can simply limit the disk size in the sizing to make sure we have instances in similar category.
> 
> It is a first feature in a series of improvements (most likely) which we can develop (quickly) and utilize to make things better, it is not the penultimate thing that will solve all of our problems. It is also extendible, so we can add more variables etc as needed
> 
> Will the single knob be enough? we shall see, we can always change it or add more knobs if needed, but I dont want to prematurely add complexity.

I think exposing cpu/memory/disk for instance selected the template, one can do any calculation it might needed, later can expose more things

thinking of it if we can expose the SCT configuration itself, might be also useful, like for performance tests

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
cezarmoise: The one downside with jinja templating it that you cannot have values in one yaml affect other yamls, like we currently have with performance tests. The throttle steps are usually in another config than the stress commands.

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
cezarmoise: Technically, we have access to actual compression ratio, but only after we start writing

image

But it is a big deal, because if you want 40% and miss the compression ratio, you could reach 50% at which decommisision / removenode fail

pehala: > But it is a big deal, because if you want 40% and miss the compression ratio, you could reach 50% at which decommisision / removenode fail

I dont think we will miss by 10%, but we will see in practice. We can  specifiy higher default compression:ratio which will push more down instead of up.



> Technically, we have access to actual compression ratio, but only after we start writing

Yes too late

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
pehala: > The one downside with jinja templating it that you cannot have values in one yaml affect other yamls, like we currently have with performance tests. The throttle steps are usually in another config than the stress commands.

We can cross this bridge when we get there in a followup, I think we will come up with better solution with Jinja then what we have now, we dont have to decide here. Point is, it is more extendible than our current solution

fruch: It depends when you are doing the resulation 

Cause later will need to resolve based on more context (SCT configuration, phase of the tests or other), I'll recommend well resolve as late as possible, and also keep unresolved fields, for when we need to resolve it multiple times with different values (i.e. perf tests, SLA)

pehala:
I did some investigations and found out, our tests are not being compressed at all. 90% elasticity test had 0,998 compression ratio, while rest of the tests I checked had 1.0, meaning there was no compression used.

I suspect it is due to usage of blobs (maybe it is the reason blobs are generally used). The effective_disk_size_bytes is now basically not needed, but I will keep it as an option and have 1.0 as default value

fruch:
> I did some investigations and found out, our tests are not being compressed at all. 90% elasticity test had 0,998 compression ratio, while rest of the tests I checked had 1.0, meaning there was no compression used.
> 
> I suspect it is due to usage of blobs (maybe it is the reason blobs are generally used). The effective_disk_size_bytes is now basically not needed, but I will keep it as an option and have 1.0 as default value

in most tests we don't pass `compression=` in the c-s command, so I probably default to none... 
and yes the default blobs c-s is using probably aren't gonna be compressed very well.
once can build relevant to compression setup with latte, but no on did so far.

pehala:
> in most tests we don't pass `compression=` in the c-s command, so I probably default to none...

Default in scylla should be LZ4WithDict I believe, so we should be using it, right?

pehala:
v2: Testing is now needed to verify if everything works
   * Add documentation to the plan
   * Add `stress_template_context` as well
      * Without it we ended up duplicates formulas across commands
   * Implement the plan   
   * Convert 90% testcase

test-cases/features/elasticity/elasticity-90-percent-with-nemesis.yaml
```diff
@@ -2,15 +2,52 @@
 test_duration: 3000
 prepare_stress_duration: 720
 
+stress_template_context:
+  fill_percent: 90
+  replication_factor: 3
+  row_size_bytes: 1024
```
cezarmoise:
```suggestion
  row_size_bytes: 1034
```
Add 10 bytes for the key column.

sdcm/utils/loader_utils.py
```diff
@@ -96,6 +136,68 @@ def tablets_enabled(self):
...
+        raw_context = self.params.get("stress_template_context") or {}
+        resolved_context = {}
+
+        for key, value in raw_context.items():
+            if self.stress_template_whitelist is not None and key in self.stress_template_whitelist:
+                raise RuntimeError(
+                    f"stress_template_context key '{key}' conflicts with a built-in stress template variable"
+                )
+
+            if isinstance(value, str):
```
cezarmoise:
```suggestion
            if isinstance(value, str) and ("{{" in value or "{%" in value or "{#" in value):
```
So only template strings go down this branch.

For example, putting `duration=30m` in there would fail.

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
pehala: resolution is currently done close to running the stress, so this shouldnt be an issue. For different phases you would have different values if the test object changed.

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
yarongilor: It could be that compression deserves a specific SCT test of it's own (or is there any exists?), so it can assist this new configuration in the future.  Thus it can be only a basic scope for this PR that could be adjusted more accurately in the future.

yarongilor:
> > I did some investigations and found out, our tests are not being compressed at all. 90% elasticity test had 0,998 compression ratio, while rest of the tests I checked had 1.0, meaning there was no compression used.
> > I suspect it is due to usage of blobs (maybe it is the reason blobs are generally used). The effective_disk_size_bytes is now basically not needed, but I will keep it as an option and have 1.0 as default value
> 
> in most tests we don't pass `compression=` in the c-s command, so I probably default to none... and yes the default blobs c-s is using probably aren't gonna be compressed very well. once can build relevant to compression setup with latte, but no on did so far.

right, it sounds like an SCT testing gap. It probably wasn't prioritized because there's not much of compression code enhancements or customer new requests. Or there's also limited coverage in unit tests.

Side Note: one testing aspect point could be about saving costs: 
If it's possible to enable and compress data in longevities, then it means we can potentially replace existing instances with less expensive ones. Or, for example, use i8g instead of i8ge.

docs/plans/mini-plans/2026-07-09-strict-jinja-stress-templating.md
```diff
@@ -0,0 +1,89 @@
...
+- Added user-facing documentation in `docs/loader_usage.md` with a compact example, runtime behavior notes, the longevity-only whitelist rule, and the shared `stress_template_context` mechanism.
+
+- Example formula supported by the implementation:
+
+```yaml
+effective_compression_ratio: 0.68
+
+stress_template_context:
+  total_rows: "{{ ((effective_disk_size_bytes * db_node_count_per_dc) // 3 // (200 * 5)) | int }}"
+  rows_per_cmd: "{{ total_rows // 4 }}"
```
yarongilor, line 36: Is this determined by the number of loaders? if so it could be a bit cleared to add "loader" to it.

pehala: This is determined by the stress command layout used by this testcase. I am not sure if it is dependant on number of loaders. I would say it is simply matter of having 4 write stress commands

pehala:
v3:
   * Based on experiments I added 10% overhead constant. This is to account for disk overhead of metadata for each row, for at least the 90% test. It is not exactly precise but it is enough for this type of PR.
   * With the constant I was able to almost match what currently was in elasticity 90% test case. 720M orignally, 738M with this PR
       * Test run was taking too long so I aborted and investigated using AI https://argus.scylladb.com/tests/scylla-cluster-tests/c3558ccd-36a2-4424-837c-97983202ff98/sct-events, there were some errors but I dont that were caused this PR.

pehala:
@scylladb/storage-qa ping for review

pehala:
> @scylladb/storage-qa ping for review

Still valid, please review

jsmolar (dismissed):

sdcm/utils/loader_utils.py
```diff
@@ -31,9 +36,44 @@
...
     """This mixin can be added to any class that inherits the 'ClusterTester' one"""
 
+    @property
+    def stress_template_whitelist(self) -> set[str] | None:
+        """Return the allowed owner-backed Jinja variables for stress command rendering.
+
+        ``None`` means owner attributes are not restricted. Callers that want a strict
+        template surface can override this property and return a concrete set.
+        """
+        return None
```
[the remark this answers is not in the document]
pehala: I dont think this is a problem. Only test that uses it for now, has a whitelist

sdcm/utils/loader_utils.py
pehala: Fixed

sdcm/sct_config.py
```diff
@@ -1850,6 +1850,28 @@ class SCTConfiguration(BaseModel):
...
+    )
+    stress_template_context: DictOrStr = SctField(
+        description=(
+            "Shared runtime-only Jinja variables for stress command templating. "
+            "Entries are resolved in declaration order and may reference earlier context entries as well as "
+            "built-in stress template variables such as effective_disk_size_bytes and db_node_count_per_dc. "
+            "These values are available to stress commands rendered by SCT, but are not evaluated during config "
+            "load or validation."
+        ),
+    )
```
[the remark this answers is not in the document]
pehala: It is documented

sdcm/utils/loader_utils.py
```diff
@@ -96,6 +136,68 @@ def tablets_enabled(self):
...
+        """
+        stress_template_context = self.build_stress_template_context()
+        try:
+            rendered = self.jinja_env.from_string(stress_cmd).render(**stress_template_context)
+        except UndefinedError as exc:
+            raise RuntimeError(
+                f"Stress command references an unknown Jinja variable: {exc}\n\nCommand: {stress_cmd}"
+            ) from exc
+        # Collapse any whitespace introduced by multi-line YAML block scalars.
+        return " ".join(rendered.split())
```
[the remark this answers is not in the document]
pehala: Valid, fixed

pehala:
v4:
   * Rebase
   * Fix coercion on non-templated variables
   * Fix whitespace handling

cezarmoise (dismissed):
LGTM

jsmolar (dismissed):

pehala:
v5:
  * Fix the unit tests

fruch (approved):

yarongilor:
> > Not sure the `effective_disk_size_bytes` can be isolated from multiple other factors dependency.
> > The stress tools have different patterns with a different compression potential.
> 
> I dont think that is the case, majority of test tools are CS-like with very simple schemas. (We will see based on the metrics)
> 
> > The load could also have deletions that impact the dataset size.
> > A fixed yaml configuration of MV for the test can also impact and double the used-disk size.
> > There could be tests configured with multiple db instance types as well.
> > In i8g family, for example there's also an extended mode of i8ge with a larger disk that changes the ration between disk-size and cpu/memory size. This can make it more complicated and it's also possible the adjusted stress-command will end up too slow , exceeding the prepare timeout threshold.
> 
> This was never meant to be perfect or ideal. It is the first step. Currently we have absolutely nothing in this regard, now we have something, test which has setup 40% and ends up with 50% is better than test with fixated instance type.
> 
> i8g vs i8ge wont be ever solved by `effective_disk_size_bytes`, but if we will require interchangeability between vastly different cpu/disk ratio we can variable based on CPU which you can use for the throttle. But that is way out of scope of this plan. For now we can simply limit the disk size in the sizing to make sure we have instances in similar category.
> 
> It is a first feature in a series of improvements (most likely) which we can develop (quickly) and utilize to make things better, it is not the penultimate thing that will solve all of our problems. It is also extendible, so we can add more variables etc as needed
> 
> Will the single knob be enough? we shall see, we can always change it or add more knobs if needed, but I dont want to prematurely add complexity.

fine, well, it can be addressed in a followup PR taking into account an MV doubles the used disk size.

pehala:
> fine, well, it can be addressed in a followup PR taking into account an MV doubles the used disk size.

It can be easily accounted for individually for the testcase that have significant MV sizes. Code I introduced doesnt need to account for MV (it doesnt even have a way how)

PR #15382, opened 2026-07-10
test(events): group event unit tests

pehala:
## Why
If multiple tests share the same name prefix, it indicates they should be grouped together into common module.
Also these tests used classes a lot, so this also converts them

## Summary
- move dedicated event unit tests under unit_tests/unit/events
- convert event test classes to pytest functions and fixtures
- parametrize repeated pure event-object checks

## Testing
- uv run python -m pytest unit_tests/unit/events -q


pehala:
@claude review

pehala:
@claude review

pehala:
> 1. `ContinuousEventsRegistry` is a Singleton; `registry` fixture doesn't reset it
> 2. Loader tests call `begin_event()` without registry isolation

  unit_tests/conftest.py already cleans it up globally
  

> #### 3. `real_events` fixture scope vs. stateful assertions (minor)
> 
> **`unit_tests/unit/events/test_real_events.py:37`**

  log_content_before = real_events.get_event_log_file("events.log")

pehala:
@scylladb/storage-qa Please review

unit_tests/unit/events/conftest.py
```diff
@@ -0,0 +1,40 @@
...
+    """Keep event process state isolated from other tests using EventsUtilsMixin.
+
+    EventsUtilsMixin stores its runtime state on class attributes, so this dedicated
+    subclass gives the fixture its own namespace instead of sharing mutable state
+    with other helpers such as RealEventsTest.
+    """
+
+
+@pytest.fixture(autouse=True)
+def tester_obj():
```
jsmolar, line 19: Only `test_events_handler.py` is using `tester_object` we can remove autouse and call it directly in the test (or remove it from conftest).

PR #15388, opened 2026-07-12
fix(trigger-matrix): use Groovy scriptText API to prevent duplicate job triggers

fruch:
## Problem

The [tier1-custom-time-trigger build #10](https://jenkins.scylladb.com/job/scylla-master/job/sct_triggers/job/tier1-custom-time-trigger/10/console) triggered several jobs **twice**:
- `longevity-twcs-48h-test`
- `longevity-2tb-5days-test`
- `elasticity-90-percent-with-nemesis-test`

### Root Cause

**Two issues combined:**

1. **Jenkins REST API "Location header" bug** ([gojenkins#248](https://github.com/bndr/gojenkins/issues/248)): `client.build_job()` triggers a job, but Jenkins sometimes omits the `Location` header from its 201 response. The python-jenkins library raises `EmptyResponseException`, our retry logic retries, and the job gets triggered **again** — even though it was already queued.

2. **Label filtering produces duplicates**: `tier1.yaml` has entries for both x86 and aarch64 variants with the same `job_name`. With `labels_selector=weekly`, both pass the filter (`{"weekly"} ⊆ {"weekly", "aarch64"}`), causing the same Jenkins job path to be triggered twice.

## Fix

1. **Replace `client.build_job()` with `client.run_script()`** — Uses Groovy `scheduleBuild2()` via the `/scriptText` API. Returns HTTP 200 with script stdout (no Location header dependency). Parses `TRIGGERED:` or `ERROR:` prefix to determine success.

2. **Inject `UpstreamCause`** — Jenkins now shows "Started by upstream project X build Y" with clickable links, making trigger chains fully traceable.

3. **Deduplicate by resolved job path** — `triggered_paths: set[str]` prevents the same job from being triggered twice even when multiple YAML entries match the same filter.

## Demo (staging)

Validated on staging Jenkins — both `RemoteCause` and `UpstreamCause` work correctly:
- [Build #8](https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/master-triggers/job/sct_triggers/job/tier1-custom-time-trigger-test/8/) — RemoteCause
- [Build #9](https://jenkins.scylladb.com/job/scylla-staging/job/fruch/job/master-triggers/job/sct_triggers/job/tier1-custom-time-trigger-test/9/) — UpstreamCause from build #7 (shows clickable link in console)

## Tests

All 115 trigger_matrix tests pass, including a new `test_duplicate_job_names_deduplicated` test and updated mocks for the `run_script` approach.

### DIFF
--- a/sdcm/utils/trigger_matrix.py
+++ b/sdcm/utils/trigger_matrix.py
@@ -717,8 +717,105 @@ def resolve_job_path(job_name: str, job_folder: str) -> str:
     return f"{job_folder}/{job_name}"
 
 
+def _escape_groovy_string(value: str) -> str:
+    """Escape a value for embedding in a Groovy single-quoted string literal."""
+    return value.replace("\\", "\\\\").replace("'", "\\'")
+
+
+def _build_trigger_groovy(job_name: str, parameters: dict, return_queue_url: bool = False) -> str:
+    """Build Groovy script that triggers a Jenkins job with UpstreamCause.
+
+    Uses Jenkins Script Console to schedule a build, bypassing the REST API
+    'Location' header issue (https://github.com/bndr/gojenkins/issues/248).
+    Injects UpstreamCause so Jenkins UI shows 'Started by upstream project X build Y'.
+
+    Args:
+        job_name: Full Jenkins job path (slash-separated).
+        parameters: Parameters to pass to the job.
+        return_queue_url: If True, return queue item URL (for wait-mode polling).
+            If False, return the job URL (fire-and-forget).
+
+    Returns:
+        Groovy script string.
+    """
+    escaped_job_name = _escape_groovy_string(job_name)
+    params_groovy = ", ".join(
+        f"new StringParameterValue('{_escape_groovy_string(k)}', '{_escape_groovy_string(str(v))}')"
+        for k, v in parameters.items()
+    )
+
+    upstream_job = os.environ.get("JOB_NAME", "")
+    upstream_build = os.environ.get("BUILD_NUMBER", "")
+
+    if upstream_job and upstream_build:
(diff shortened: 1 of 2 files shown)


unit_tests/unit/test_argus_postman_stop.py
fruch, line 97: this changes isn't related to the PR

fruch:
@claude review

fruch:
All review feedback addressed in the force-push (`e67d065`):

**🔴 Critical — Wait-mode queue URL**: Fixed. `_build_trigger_groovy()` now takes a `return_queue_url` parameter. When `True` (used by `trigger_jenkins_job_with_queue`), the Groovy script looks up the queue item via `Jenkins.instance.queue.getItems().find { it.task == targetJob }` and returns `{rootUrl}queue/item/{id}/` — suitable for `_wait_for_build_start()` polling. Falls back to job URL if the item already left the queue.

**🟠 Major — Missing retry loop**: Fixed. `trigger_jenkins_job_with_queue` now has the same `MAX_TRIGGER_RETRIES` retry loop with exponential backoff as `trigger_jenkins_job`. `ERROR:` prefix exits immediately (non-retryable), `JenkinsException` retries.

**🟡 Minor — Groovy injection**: Fixed. Added `_escape_groovy_string()` helper (escapes `\` and `'`) applied to all interpolated strings: `job_name`, `upstream_job`, parameter keys and values.

**🔵 Unrelated change**: Dropped `test_argus_postman_stop.py` and a stray `snitch_configuration.py` change. Diff is now only the 2 relevant files.

unit_tests/unit/test_argus_postman_stop.py
fruch: Dropped in `e67d065` — no longer in the diff.

PR #15404, opened 2026-07-13
improvement(tester): collect scylla shards from all live nodes in parallel

roydahan:
## Summary

`SCTTester.all_nodes_scylla_shards()` collects per-node shard counts for the end-of-test email report. Each `node.scylla_shards` access runs a remote SSH command (a `grep` on `SCYLLA_ARGS` in the scylla-server sysconfig), which can take up to a few minutes per node. Iterating serially over the nodes made teardown of large clusters (hundreds of nodes) prohibitively slow.

This change collects the data from all live nodes **concurrently** using `ParallelObject`.

## Details

- Extracted the per-node logic into a helper and run it across all nodes with a **capped worker pool** (`min(len(nodes), 100)`) to avoid spawning hundreds of threads and simultaneous SSH sessions.
- `ignore_exceptions=True` — a single slow/unreachable node no longer discards the shard data of every other node. Failed nodes are logged and still listed in the report with `"N/A"` shards.
- Generous `600s` timeout. `ParallelObject` shares one timeout budget across futures collected in submission order, and once one future times out it marks all remaining uncollected futures as timed-out too — so a tight timeout would cascade into false failures for nodes that actually succeeded.

## Testing

- `pre-commit` (ruff / ruff-format) passes on the changed file.

Fixes QAINFRA-125

### DIFF
--- a/sdcm/tester.py
+++ b/sdcm/tester.py
@@ -4836,15 +4836,42 @@ class ClusterTester(unittest.TestCase):
 
     def all_nodes_scylla_shards(self):
         all_nodes_shards = defaultdict(list)
-        for node in self.db_cluster.nodes:
+
+        def get_node_shards(node):
             ipv6 = node.ipv6_ip_address if node.ip_address == node.ipv6_ip_address else ""
-            all_nodes_shards["live_nodes"].append(
-                {
-                    "name": node.name,
-                    "ip": f"{node.public_ip_address} | {node.private_ip_address}{f' | {ipv6}' if ipv6 else ''}",
-                    "shards": node.scylla_shards,
-                }
+            return {
+                "name": node.name,
+                "ip": f"{node.public_ip_address} | {node.private_ip_address}{f' | {ipv6}' if ipv6 else ''}",
+                "shards": node.scylla_shards,
+            }
+
+        nodes = list(self.db_cluster.nodes)
+        if nodes:
+            # Collect shards data from all live nodes in parallel. `scylla_shards` runs remote SSH
+            # commands that may take up to a few minutes per node, so a single serial pass over
+            # hundreds of nodes would be prohibitively slow.
+            # ignore_exceptions=True is used deliberately: this data feeds the end-of-test email
+            # report, so one unreachable/slow node must not discard the results of all the others.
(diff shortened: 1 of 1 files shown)


fruch:
@claude review

soyacz:
@fruch do we still need that? I thought we dropped email data from SCT - currently Argus sends that.

roydahan:
So, can we merge?

soyacz:
LLM says:
## Summary

`email_data.json` is **not used anywhere** in the current codebase for sending emails or anything else.

- It was a legacy artifact written by `performance_regression_test.py` (an SMTP/Jinja2-based email pipeline via `sdcm/send_email.py`).
- It was fully removed in commit `5103941ff` — *"refactor(email): remove sdcm/send_email.py and use only Argus email reporting"* (2026-03-17, Israel Fruchter). That commit deleted `sdcm/send_email.py`, all report templates, and every `email_data.json` reference from perf tests and the log collector.
- Current email sending goes through `hydra.sh send-email` → `sct.py:send_email()` → `client.send_email(...)` (Argus client), which pulls data from Argus (`client.get_run()`) and a YAML template in `argus_report_templates/`, not from any JSON file.
- The only remaining mention of `email_data.json` in your working tree is in an untracked scratch notebook, `perf_report.ipynb`, referencing a `read_email_data_from_file` helper that no longer exists in `sdcm/` — that notebook is stale/leftover from before the refactor.

The `get_email_data()`/`save_email_data()` methods still in `sdcm/tester.py` build an in-memory dict, but it now feeds Argus submissions (screenshots, relocatable package info), not a JSON file.

I propose, for safety, just make this function no-op (or return dummy data) - later we'll handle full removal.

fruch:
yes, we can throw everything out, and do 2 calls to Argus directly  

so this one is just wasting our time, and better of dropping completely.

roydahan:
ok, since I don't know where to test it and how.
@soyacz can you please take over and add to this PR or replace it with new PR?

fruch:
> ok, since I don't know where to test it and how. @soyacz can you please take over and add to this PR or replace it with new PR?

moved the issue to https://scylladb.atlassian.net/browse/SCT-662, and gave it to @soyacz 
closing this PR for now

PR #15412, opened 2026-07-13
fix(curl): retry connection reset errors

dimakr:
Downloads through curl_with_retry() occasionally fail on TCP connection resets (curl exit 35/56), which plain `--retry` option does not retry.

The fix adds an opt-in `retry_all_errors` flag to curl_with_retry() that appends `--retry-all-errors` option to curl command to be executed. Additionally, ClusterTester.tearDown() is updated to check for a missing self.params, so an early setUp failure throws real error instead of an AttributeError.

Fixes: SCT-622

### Testing

- [x] :green_circle: [artifacts-ubuntu2604-test (offline-install)](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts/job/artifacts-ubuntu2604-test/3/)
- [x] :green_circle: [artifacts-rocky8-test (old curl path)](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts/job/artifacts-rocky8-test/2/)
- [x] :green_circle: [artifacts-ubuntu2204-test (old curl path)](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/artifacts/job/artifacts-ubuntu2204-test/2/)
- [x] :green_circle: [pr-provision-test (userdata script path)](https://jenkins.scylladb.com/view/staging/job/scylla-staging/job/dimakr/job/pr-provision-test/320/)
- [x] :green_circle: tested sct-runner image build:
```
❯ uv run ./sct.py create-runner-image -c aws -r eu-west-2 -z a
...
Got public IP: 13.40.65.48
Connecting instance...
Installing required packages...
All packages successfully installed.
Stopping the SCT Image Builder instance...
SCT Image Builder instance stopped.
Creating image...
Waiting for image 'ami-005c8a42816ece0f3' to become available...
Image 'ami-005c8a42816ece0f3' is available. Tagging...
Tagging completed.
```

### Reminders

### DIFF
--- a/sdcm/cluster_k8s/mini_k8s.py
+++ b/sdcm/cluster_k8s/mini_k8s.py
@@ -43,6 +43,7 @@ from sdcm.cluster_k8s import (
     SCYLLA_MANAGER_AGENT_RESOURCES,
     SCYLLA_MANAGER_AGENT_VERSION_IN_SCYLLA_MANAGER,
 )
+from sdcm.utils.curl import curl_with_retry
 from sdcm.utils.k8s import TokenUpdateThread, HelmValues
 from sdcm.utils.k8s.chaos_mesh import ChaosMesh
 from sdcm.utils.decorators import retrying
@@ -131,11 +132,13 @@ class MinimalK8SOps:
             sysctl --system
             """)
         node.remoter.sudo(f'bash -cxe "{script}"')
+        helm_curl = curl_with_retry(
+            f'"https://get.helm.sh/helm-{HELM_VERSION}-linux-amd64.tar.gz"',
+            silent=True,
+            follow_redirects=True,
+        )
         node.remoter.sudo(
-            'bash -cxe "helm version'
-            f' || (curl --silent --location "https://get.helm.sh/helm-{HELM_VERSION}-linux-amd64.tar.gz"'
-            "  | tar xz -C /tmp && mv /tmp/linux-amd64/helm /usr/local/bin)"
-            '"'
+            f'bash -cxe "helm version || ({helm_curl}  | tar xz -C /tmp && mv /tmp/linux-amd64/helm /usr/local/bin)"'
         )
 
         # NOTE: if running in Hydra then it must have '/dev' mount from host as 'rw'
@@ -153,9 +156,16 @@ class MinimalK8SOps:
 
     @staticmethod
     def setup_docker_ubuntu(node: cluster.BaseNode, target_user: str = None) -> None:
+        docker_gpg_curl = curl_with_retry(
+            "https://download.docker.com/linux/ubuntu/gpg",
+            silent=True,
+            follow_redirects=True,
+            fail_early=True,
+            extra_flags="-S",
+        )
         script = dedent(f"""
             # Install and configure Docker.
-            curl -fsSL https://download.docker.com/linux/ubuntu/gpg | apt-key add -
+            {docker_gpg_curl} | apt-key add -
             add-apt-repository \\"deb [arch=amd64] https://download.docker.com/linux/ubuntu $(lsb_release -cs) stable\\"
             apt-get -qq install --no-install-recommends docker-ce docker-ce-cli containerd.io
             {f"usermod -a -G docker {target_user}" if target_user else ""}
@@ -171,10 +181,17 @@ class MinimalK8SOps:
 
     @staticmethod
     def setup_kubectl_ubuntu(node: cluster.BaseNode, kubectl_version: str) -> None:
+        kubectl_curl = curl_with_retry(
+            f"https://storage.googleapis.com/kubernetes-release/release/v{kubectl_version}/bin/linux/amd64/kubectl",
+            silent=True,
+            follow_redirects=True,
+            fail_early=True,
+            output="/usr/local/bin/kubectl",
+            extra_flags="-S",
+        )
         script = dedent(f"""
             # Download kubectl binary.
-            curl -fsSLo /usr/local/bin/kubectl \
-                https://storage.googleapis.com/kubernetes-release/release/v{kubectl_version}/bin/linux/amd64/kubectl
+            {kubectl_curl}
             chmod +x /usr/local/bin/kubectl
             """)
         node.remoter.sudo(f'bash -cxe "{script}"')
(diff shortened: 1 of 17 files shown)


sdcm/utils/curl.py
```diff
@@ -10,6 +10,7 @@ def curl_with_retry(
     *,
     retry: int = 5,
     retry_max_time: int = 300,
+    retry_all_errors: bool = False,
```
fruch: Maybe we should make it default true ?

Which is the point in this one, to apply what we assume is best practice for resilience

fruch:
@claude review

What are the downsides of applying retry on all error for curl, based on the usage of it in SCT, what should be paying attention to ?

fruch:
@claude 

I don't mind waiting for extra time on 404/403, we will fail the test in all of those case if we can download what we need

Cross check suggestion with https://blog.gregor.com/curl-considered-harmful-for-automation-without-these-options-6b89f3e4f50a

Doing version/distro seems a must for now (and will enable us to backport this fix)

fruch:
@dimakr 

clause suggested something in:
https://github.com/scylladb/scylla-cluster-tests/commit/199e316cd9647da9ece97c24d1f1dcdcc698960b

maybe it would be better todo base on SCT distros ? 

anyhow if we don't go with default, we should followup his review of usage, and enable where it's clear it's supported

dimakr:
> @dimakr
> 
> clause suggested something in: [199e316](https://github.com/scylladb/scylla-cluster-tests/commit/199e316cd9647da9ece97c24d1f1dcdcc698960b)
> 
> maybe it would be better todo base on SCT distros ?
> 
> anyhow if we don't go with default, we should followup his review of usage, and enable where it's clear it's supported

the suggested regex approach could probably break in some cases, as sometimes the command will be used/wrapped inside single quotes

dimakr:
@claude I assume the suggested regex approach might be problematic in some call sites where it will used inside single quotes.
The current implementation has been reworked to cover old curl distros. Re-review it

sdcm/cluster_k8s/mini_k8s.py
```diff
@@ -133,7 +134,7 @@ def setup_prerequisites_ubuntu(node: cluster.BaseNode) -> None:
         node.remoter.sudo(f'bash -cxe "{script}"')
         node.remoter.sudo(
             'bash -cxe "helm version'
-            f' || (curl --silent --location "https://get.helm.sh/helm-{HELM_VERSION}-linux-amd64.tar.gz"'
+            f' || (curl --silent --location --retry 5 --retry-max-time 300 {RETRY_ALL_ERRORS_PROBE} "https://get.helm.sh/helm-{HELM_VERSION}-linux-amd64.tar.gz"'
```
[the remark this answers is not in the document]
dimakr: updated

dimakr:
@fruch reworked this, changing the latest suggested approach to a probe based obe

sdcm/provision/common/utils.py
```diff
@@ -262,7 +262,7 @@ def install_vector_service():
     return dedent("""\
         # install repo
         for n in 2 4 6 8 10 10 10 10; do # cloud-init is running it with set +o braceexpand
-            if bash -c "$(curl -L --retry 5 --retry-max-time 300 https://setup.vector.dev)"; then
+            if bash -c "$(curl -L --retry 5 --retry-max-time 300 $(curl --retry-all-errors --version >/dev/null 2>&1 && echo --retry-all-errors) https://setup.vector.dev)"; then
```
fruch: why can't we use curl_with_retry here ?

sdcm/provision/common/utils.py
```diff
@@ -311,7 +311,7 @@ def install_vector_service():
 
 def install_syslogng_exporter():
     return dedent("""\
-    curl -L -O https://github.com/brandond/syslog_ng_exporter/releases/download/0.1.0/syslog_ng_exporter
+    curl -L -O --retry 5 --retry-max-time 300 $(curl --retry-all-errors --version >/dev/null 2>&1 && echo --retry-all-errors) https://github.com/brandond/syslog_ng_exporter/releases/download/0.1.0/syslog_ng_exporter
```
fruch: why can't we use curl_with_retry here ?

sdcm/utils/sct_agent_installer.py
```diff
@@ -186,7 +188,8 @@ def install_agent_script(
     return f"""echo "Installing SCT agent..."
 
 echo "Downloading agent binary from {agent_binary_url}..."
-curl -fsSL -o {binary_path} {agent_binary_url}
+# probe emits --retry-all-errors only when node curl supports it, curl >= 7.71
+curl -fsSL -o {binary_path} --retry 5 --retry-max-time 300 {RETRY_ALL_ERRORS_PROBE} {agent_binary_url}
```
fruch: why can't we use curl_with_retry here ?

fruch (dismissed):
LGTM

would be nice it we can make `curl_with_retry` used across the board, so we would need to tailor the retry and probe in multiple places

fruch:
@claude review

sdcm/provision/common/utils.py
```diff
@@ -427,11 +435,19 @@ def install_vector_from_local_pkg(pkg_path: str) -> str:
...
+        fail_early=True,
+        output="get-docker.sh",
+        extra_flags="-S",
+    )
+    return dedent(f"""\
         # Install Docker
 
         for n in 1 2 3; do
-            if bash -c "$(curl -fsSL get.docker.com --retry 5 --retry-max-time 300 -o get-docker.sh)"; then
+            if bash -c "$({get_docker_curl})"; then
```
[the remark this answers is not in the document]
dimakr: predates this PR, but fixed

sdcm/provision/common/utils.py
```diff
@@ -259,10 +261,11 @@ def install_syslogng_service():
 
 
 def install_vector_service():
-    return dedent("""\
+    vector_setup_curl = curl_with_retry("https://setup.vector.dev", follow_redirects=True)
+    return dedent(f"""\
         # install repo
         for n in 2 4 6 8 10 10 10 10; do # cloud-init is running it with set +o braceexpand
-            if bash -c "$(curl -L --retry 5 --retry-max-time 300 https://setup.vector.dev)"; then
+            if bash -c "$({vector_setup_curl})"; then
```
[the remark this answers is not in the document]
dimakr: added flags

sdcm/kafka/kafka_cluster.py
```diff
@@ -87,14 +88,26 @@ def install_connector_from_hub(self, connector_version: str = "scylladb/scylla-c
...
             if connector_url.endswith(".jar"):
                 self.remoter.run(
-                    f"curl --show-error --fail -L --create-dirs -O --output-dir {self.docker_compose_path / 'connectors'} {connector_url} "
+                    curl_with_retry(
+                        connector_url,
+                        follow_redirects=True,
+                        fail_early=True,
+                        extra_flags=f"--show-error --create-dirs -O --output-dir {connectors_dir}",
+                    )
                 )
```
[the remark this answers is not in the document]
dimakr: no change - runs on localrunner, so has has modern curl

sdcm/provision/common/utils.py
dimakr: updated, here and other occurrences

fruch:
@dimakr Claude is doing lots of work to dismiss what I was saying :)

Still consolidation of those calls is important.

fruch (approved):

